Issue #3130, opened 2024-02-14
FFI: BackupDownloadStrategy cannot be set in ClientBuilder


kegsay:
The `client_builder` used to set the `BackupDownloadStrategy` to `AfterDecryptionFailure` which was a sensible default. In the refactoring in https://github.com/matrix-org/matrix-rust-sdk/pull/3115 this changed, and the default now fell to the value in the `enum` which is `Manual`. I don't think this is a sensible default.

This is made worse by the fact that it's impossible to change this value now, because FFI bindings aren't created for `pub(crate) fn with_encryption_settings(...)` due to visibility rules.

The net result is that key backup downloading is broken on latest main EDIT: if you use the authentication service then the right defaults are set. This only applies when you use client_builder directly.

bnjbvr:
Reopening: #3131 was a quick workaround, but it's still not possible to set it.

PR #3326, opened 2024-04-11
ffi: Expose encryption settings via FFI

andybalaam:
Fixes https://github.com/matrix-org/matrix-rust-sdk/issues/3130

This allows providing encryption settings while constructing a client via FFI, instead of hard-coding specific values.

We do it by holding on to an `EncryptionSettings` inside `ClientBuilder` and updating it when certain methods on the builder are called. During `build_inner()` we provide those `EncryptionSettings` by calling `with_encryption_settings` on `inner_builder`.

I added an enum `BackupDownloadStrategy` to be passed as an argument to `backup_download_strategy`. Let me know if there's a better way.

### DIFF
--- a/bindings/matrix-sdk-ffi/src/client_builder.rs
+++ b/bindings/matrix-sdk-ffi/src/client_builder.rs
@@ -76,6 +76,7 @@ pub struct ClientBuilder {
     cross_process_refresh_lock_id: Option<String>,
     session_delegate: Option<Arc<dyn ClientSessionDelegate>>,
     additional_root_certificates: Vec<Vec<u8>>,
+    encryption_settings: EncryptionSettings,
 }
 
 #[uniffi::export(async_runtime = "tokio")]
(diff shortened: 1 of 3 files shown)


bnjbvr (approved):

bindings/matrix-sdk-ffi/src/client_builder.rs
```diff
@@ -203,21 +244,41 @@ impl ClientBuilder {
...
+        let mut builder = unwrap_or_clone_arc(self);
+        builder.encryption_settings.auto_enable_cross_signing = auto_enable_cross_signing;
+        Arc::new(builder)
     }
-}
 
-impl ClientBuilder {
-    pub(crate) fn with_encryption_settings(
+    /// Select a strategy to download room keys from the backup, by default room
+    /// keys won't be downloaded from the backup automatically.
```
bnjbvr, line 257: Can we describe what the default behavior is, instead of what it is not?

bindings/matrix-sdk-ffi/src/client_builder.rs
```diff
@@ -27,6 +27,44 @@ enum HomeserverConfig {
...
+    /// Attempt to download a single room key if an event fails to be decrypted.
+    AfterDecryptionFailure,
+
+    /// Don't download any room keys automatically. The user can manually
+    /// download room keys using the [`Backups::download_room_key()`] methods.
+    Manual,
+}
+
+impl From<BackupDownloadStrategy> for matrix_sdk::encryption::BackupDownloadStrategy {
+    fn from(strat: BackupDownloadStrategy) -> Self {
```
bnjbvr, line 53: nit: can you spell it `strategy`?

bindings/matrix-sdk-ffi/src/client_builder.rs
```diff
@@ -27,6 +27,44 @@ enum HomeserverConfig {
     ServerNameOrUrl(String),
 }
 
+#[derive(uniffi::Enum)]
+pub enum BackupDownloadStrategy {
```
bnjbvr, line 31: nit: instead of copying the enum, can you add the derive with a `feature = 'uniffi'` guard to the original enum?

bnjbvr (approved):

bindings/matrix-sdk-ffi/src/client_builder.rs
andybalaam: Thanks, I don't know what came over me! Fixed in cf35c38dad

bindings/matrix-sdk-ffi/src/client_builder.rs
andybalaam: Much better, done in 71c6878697 !

Hywan (approved):

PR #3328, opened 2024-04-12
sliding sync: don't exclude room membership events sent by other users

### LINKED ISSUE #3315: Kick or Ban action is not updating the RoomInfo.membership
When the current user is kicked or banned from a room, his `roomInfo.membership` is not updated to `LEFT` but stays `JOINED`.

If I restart the app, it does get updated, so might be something wrong with the sliding sync?

Note from @kegsay : that's by design currently,so you know you have been kicked/banned. So rust sdk needs to process the membership event on his own.

bnjbvr:
A kick/ban event has the sender key and the state key defined to different values. The banner/kicker is the actual sender. Sliding sync considered they had to be the same for the event to be legit, but it's not the case. Thanks @ganfra for showing me actual events sent from the sliding sync proxy, which made the bug in the SDK obvious.

Contains a regression test.

Fixes #3315.

### DIFF
--- a/crates/matrix-sdk-base/src/sliding_sync.rs
+++ b/crates/matrix-sdk-base/src/sliding_sync.rs
@@ -897,6 +897,43 @@ mod tests {
     }
 
     #[async_test]
+    async fn test_kick_or_ban_updates_room_to_left() {
+        for membership in [MembershipState::Leave, MembershipState::Ban] {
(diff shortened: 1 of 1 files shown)


andybalaam (approved):
Looks good, thanks!

Hywan:
Well done!

PR #3329, opened 2024-04-15
timeline: add event focus mode for permalinks

### LINKED ISSUE #3234: Pin the timeline to a given event and paginate all the directions (permalink)
We want to support permalinks in the (UI) timeline of the Rust SDK, that is, the ability to click a link to an event, and navigate around it (in the past and future, i.e. backwards- and forward- paginate).

We do have some support for backwards pagination already in the event cache, so there's some previous work here. Here's how I see the next steps:
  
- [x] add support for arbitrary pagination in the event cache, backwards- or forward-
  - [x] retrieve the initial data, given an initial event id, using the `/context` endpoint
  - [x] then allow stateful pagination using a paginator kind of API. Paginating forward in the timeline would return whether we've reached the end of the timeline, at which point the caller may change the focus back to `Live`.
  - [x] (edge case) handle forged event ids not resolving to an actual event (transform 404 into a proper error)
  - https://github.com/matrix-org/matrix-rust-sdk/pull/3309
- [x] add a "focus" mode to a timeline: either pinned to an event (for permalinks), or live (i.e. listening to sync).
  - [x] make it possible to switch focus
  - [x] (edge case) handle event ids pointing to non-visible (filtered out) events
  - https://github.com/matrix-org/matrix-rust-sdk/pull/3329
- [x] (bugfix + nice to have) fix date separator insertion by using a post-processing pass instead of inserting in an ad-hoc manner. This may avoid future headaches when working on forward pagination, or insertion at random spots in the timeline.
  - https://github.com/matrix-org/matrix-rust-sdk/pull/3254
  - https://github.com/matrix-org/matrix-rust-sdk/pull/3288
  - https://github.com/matrix-org/matrix-rust-sdk/pull/3273
  - https://github.com/matrix-org/matrix-rust-sdk/pull/3268
  - https://github.com/matrix-org/matrix-rust-sdk/pull/3267

Part of https://github.com/matrix-org/matrix-rust-sdk/issues/3058

## Nice to have (deferred)

- [ ] (nice to have) when clicking a permalink to an event that's "close by" (TBD), don't switch focus, but prepend events (until the focused event - N events)
- [ ] (nice to have) another focus mode that's "latest unread" (= read marker)
- [ ] (nice to have) make the paginator stateful object able to read events from the cache before spawning server queries
- [ ] (nice to have, but not mandatory for this to happen) reconcile events received from a pagination with cached events.
  - Subsets of the problem to consider: no overlap (easy; throw away/keep in memory until there's *some* overlap after keeping on paginating), complete overlap (easy; do nothing — maybe replace pagination tokens), partial overlaps (intractable in the worst cases: what if we know A/B/C and receive A/D/C, as there's no way to figure the relative position of B and D — easy if there's linear overlap, aka A/B/C is known and we receive A/B/C/D or D/A/B/C).

bnjbvr:
Fixes #3234.

See the commit message of the first commit for more details. This is sufficient to implement pagination in the timeline.

### DIFF
--- a/bindings/matrix-sdk-ffi/src/timeline/mod.rs
+++ b/bindings/matrix-sdk-ffi/src/timeline/mod.rs
@@ -22,7 +22,7 @@ use matrix_sdk::attachment::{
     AttachmentConfig, AttachmentInfo, BaseAudioInfo, BaseFileInfo, BaseImageInfo,
     BaseThumbnailInfo, BaseVideoInfo, Thumbnail,
 };
-use matrix_sdk_ui::timeline::{BackPaginationStatus, EventItemOrigin, Profile, TimelineDetails};
+use matrix_sdk_ui::timeline::{EventItemOrigin, PaginationStatus, Profile, TimelineDetails};
 use mime::Mime;
 use ruma::{
     events::{
@@ -162,7 +162,7 @@ impl Timeline {
 
     pub fn subscribe_to_back_pagination_status(
         &self,
-        listener: Box<dyn BackPaginationStatusListener>,
+        listener: Box<dyn PaginationStatusListener>,
     ) -> Result<Arc<TaskHandle>, ClientError> {
         let mut subscriber = self.inner.back_pagination_status();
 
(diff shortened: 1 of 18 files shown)


crates/matrix-sdk-ui/src/timeline/pagination.rs
```diff
@@ -14,26 +14,67 @@
...
+
+    /// Paginate backwards in live mode.
+    ///
+    /// This can only be called when the timeline is in live mode, not focused
+    /// on a specific event.
+    ///
+    /// Returns whether we hit the start of the timeline.
     #[instrument(skip_all, fields(room_id = ?self.room().room_id(), ?options))]
-    pub async fn paginate_backwards(
+    pub async fn live_paginate_backwards(
```
Hywan, line 64: I wonder if this renaming is necessary.

bindings/matrix-sdk-ffi/src/timeline/mod.rs
```diff
@@ -176,11 +176,18 @@ impl Timeline {
...
+    /// Returns whether we hit the end of the timeline or not.
+    pub async fn paginate_backwards(&self, num_events: u16) -> Result<bool, ClientError> {
+        Ok(self.inner.paginate_backwards(num_events).await?)
+    }
+
+    /// Paginate forwards, when in focused mode.
+    ///
+    /// Returns whether we hit the end of the timeline or not.
+    pub async fn paginate_forwards(&self, num_events: u16) -> Result<bool, ClientError> {
+        Ok(self.inner.focused_paginate_forwards(num_events).await?)
```
Hywan, line 190: Should it be simply `paginate_forwards`?

Hywan (changes requested):
Excellent work, well done!

Hywan:
Actually, there is a problem with the pagination code:

1. it is duplicated with the code inside `event_cache::paginator`, or at least partially
2. it creates another duplication to handle the pagination status.

Finally, there is another problem:

3. if the pagination is cancelled, the status is not reset, thus forbidding it to re-run properly; for example, if the status is `Paginating`, the pagination won't run; problem is: the status is reset after the pagination is done, it's set to `Idle`, but if the method is cancelled before reaching the `Idle` status, it will in `Paginating` forever.

Hywan (approved):

PR #3331, opened 2024-04-15
fix(sdk-common, wasm): implment JoinHandle correctly

maan2003:
fixes task getting cancelled on drop of join handle

RemoteHandle already aborts on drop, we don't need abort handle


maan2003:
looks like I broke some stuff, my work based on old commit. will fix

maan2003:
closing in favor of https://github.com/matrix-org/matrix-rust-sdk/pull/3333

PR #3332, opened 2024-04-15
fix(wasm): avoid std::time and tokio::time

maan2003:
Signed-off-by: Manmeet Singh <manmeetmann2003@gmail.com>


maan2003:
will part of a bigger PR that actually compiles matrix sdk ui on wasm

Issue #3310, opened 2024-04-09
Regression: The script to build the crypto ffi bindings is not working anymore


BillCarsonFr:
I think related to that https://github.com/matrix-org/matrix-rust-sdk/pull/2997 that did remove the UDL file

Now the script is failing https://github.com/matrix-org/matrix-rust-sdk/blob/main/bindings/apple/build_crypto_xcframework.sh

```
Caused by:
    0: failed to open file `../../bindings/matrix-sdk-crypto-ffi/src/olm.udl`
    1: No such file or directory (os error 2)
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

BillCarsonFr:
This is blocking update of the legacy apps that would benefit from recent fixes on the crypto sdk

PR #3334, opened 2024-04-16
crypto: Apple Crypto Bindings | Fix crypto xcframework script

BillCarsonFr:
Fixes https://github.com/matrix-org/matrix-rust-sdk/issues/3310

The script to build the crypto bindings was not working anymore since the move from UDL to procedural macro.

Modified as per the following recommendations https://mozilla.github.io/uniffi-rs/tutorial/foreign_language_bindings.html#running-uniffi-bindgen-using-a-library-file-recommended

There is some changes in the generated framework, there are now two swift files and two modulemap files that have to be merged together in the final xcframework.

Previous structure was
image

The new one is 
image

Signed-off-by:

### DIFF
--- a/.github/workflows/bindings_ci.yml
+++ b/.github/workflows/bindings_ci.yml
@@ -140,7 +140,7 @@ jobs:
           rustup target add aarch64-apple-ios
           rustup target add aarch64-apple-darwin
           rustup target add x86_64-apple-darwin
-          rustup target add iaarch64-apple-ios-sim
+          rustup target add aarch64-apple-ios-sim
           rustup target add x86_64-apple-ios
 
       # Cargo config can screw with caching and is only used for alias config
--- a/.github/workflows/bindings_ci.yml
+++ b/.github/workflows/bindings_ci.yml
@@ -154,7 +154,7 @@ jobs:
           save-if: ${{ github.ref == 'refs/heads/main' }}
 
       - name: Run the Build Framework script
-        run: ./bindings/apple/build_crypto_xcframework.sh
+        run: ./bindings/apple/build_crypto_xcframework.sh -i
 
       - name: Is XCFramework generated?
         if: ${{ hashFiles('generated/MatrixSDKCryptoFFI.zip') != '' }}
--- a/.github/workflows/bindings_ci.yml
+++ b/.github/workflows/bindings_ci.yml
@@ -138,10 +138,6 @@ jobs:
       - name: Add rust targets
         run: |
           rustup target add aarch64-apple-ios
-          rustup target add aarch64-apple-darwin
-          rustup target add x86_64-apple-darwin
-          rustup target add aarch64-apple-ios-sim
-          rustup target add x86_64-apple-ios
 
       # Cargo config can screw with caching and is only used for alias config
       # and extra lints, which we don't care about here
--- a/.github/workflows/bindings_ci.yml
+++ b/.github/workflows/bindings_ci.yml
@@ -116,3 +116,46 @@ jobs:
     with:
         use_rust_sdk: "." # use local checkout
         use_complement_crypto: "MATCHING_BRANCH"
+
+  test-crypto-apple-framework-generation:
+    name: Generate Crypto FFI Apple XCFramework
+    runs-on: macos-12
+    if: github.event_name == 'push' || !github.event.pull_request.draft
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+
+      # install protoc in case we end up rebuilding opentelemetry-proto
+      - name: Install protoc
+        uses: taiki-e/install-action@v2
+        with:
+          tool: protoc@3.20.3
+
+      - name: Install Rust
+        uses: dtolnay/rust-toolchain@stable
+
+      - name: Add rust targets
+        run: |
+          rustup target add aarch64-apple-ios
+          rustup target add aarch64-apple-darwin
+          rustup target add x86_64-apple-darwin
+          rustup target add iaarch64-apple-ios-sim
+          rustup target add x86_64-apple-ios
+
+      # Cargo config can screw with caching and is only used for alias config
+      # and extra lints, which we don't care about here
+      - name: Delete cargo config
+        run: rm .cargo/config.toml
+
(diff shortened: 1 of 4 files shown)


bindings/apple/build_crypto_xcframework.sh
```diff
@@ -54,20 +54,19 @@ lipo -create \
...
 
 # Move headers to the right place
 HEADERS_DIR=${GENERATED_DIR}/headers
 mkdir -p ${HEADERS_DIR}
 mv ${GENERATED_DIR}/*.h ${HEADERS_DIR}
 
-# Rename and move modulemap to the right place
-mv ${GENERATED_DIR}/*.modulemap ${HEADERS_DIR}/module.modulemap
+# Rename and merge the modulemap files into a single file to the right place
+for f in ${GENERATED_DIR}/*.modulemap; do cat $f; echo; done > ${HEADERS_DIR}/module.modulemap
```
Hywan, line 68: Why do you need an `echo` here? To have a newline?

Hywan (commented):
Thanks for the fix!

It looks OK to me. Please address the feedback from @bnjbvr and rebase your commits. Once it's done, I can approve this PR :-).

It's also surprising that it's not detected by the CI. Do you know if the CI is tested this script? I was sure we were testing it previously, but running a `git rev-list --all | xargs git grep …` returns nothing valuable. Maybe let's add this check on the CI? I can take a while though… cc @poljar

Hywan:
Can you update `.github/workflows/ci.yml` to test this script please?

bindings/apple/build_crypto_xcframework.sh
BillCarsonFr: Yes for the new line. What would be the best way?

poljar:
> Maybe let's add this check on the CI? I can take a while though… cc @poljar

Yeah, if possible we should ensure that this works using CI.

bindings/apple/build_crypto_xcframework.sh
Hywan: I've looked for but found nothing. Let's consider it's fine.

bnjbvr (changes requested):
The new CI step takes 38 minutes, which is longer than the whole pipeline, so we agreed to try to make it shorter first.

BillCarsonFr:
> The new CI step takes 38 minutes, which is longer than the whole pipeline, so we agreed to try to make it shorter first.

Now `Successful in 9:58mn` https://github.com/matrix-org/matrix-rust-sdk/actions/runs/9079468164/job/24948734659?pr=3334

bindings/apple/build_crypto_xcframework.sh
```diff
@@ -22,52 +38,68 @@ TARGET_CRATE=matrix-sdk-crypto-ffi
...
+    --language swift \
+    --library "${TARGET_DIR}/aarch64-apple-ios/${REL_TYPE_DIR}/libmatrix_sdk_crypto_ffi.a" \
+    --out-dir ${GENERATED_DIR}
+else
+  # Generate uniffi files
+  cd ../matrix-sdk-crypto-ffi && cargo run --bin matrix_sdk_crypto_ffi generate \
+    --language swift \
+    --library "${TARGET_DIR}/aarch64-apple-ios-sim/${REL_TYPE_DIR}/libmatrix_sdk_crypto_ffi.a" \
+    --out-dir ${GENERATED_DIR}
+fi
```
Velin92, line 93: just not super sure about this part, we are not creating ffi files for Mac in te non iOS case?

BillCarsonFr: The previous script was only doing that by default
```
  # Generate uniffi files
  cd ../matrix-sdk-crypto-ffi && cargo run --bin matrix_sdk_crypto_ffi generate \
    --language swift \
    --library "${TARGET_DIR}/aarch64-apple-ios-sim/${REL_TYPE_DIR}/libmatrix_sdk_crypto_ffi.a" \
    --out-dir ${GENERATED_DIR}
```

Do you know what it should be doing?

BillCarsonFr: As discussed with Doug, the `if` is not needed. We can just use aarch64-apple-ios for uniffi (built in the 2 options), it will be the same output for all targets.

bindings/apple/build_crypto_xcframework.sh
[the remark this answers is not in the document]
BillCarsonFr: done

.github/workflows/bindings_ci.yml
```diff
@@ -116,3 +116,46 @@ jobs:
...
+      - name: Delete cargo config
+        run: rm .cargo/config.toml
+
+      - name: Load cache
+        uses: Swatinem/rust-cache@v2
+        with:
+          save-if: ${{ github.ref == 'refs/heads/main' }}
+
+      - name: Run the Build Framework script
+        run: ./bindings/apple/build_crypto_xcframework.sh -i
```
Hywan, line 157: The `-i` option describes the fact you want to test for iOS only. Then, this is not necessary to add the following targets with `rustup` hereinabove (in the “Add rust targets” step):
- `aarch64-apple-darwin`
- `x86_64-apple-darwin`
- `aarch64-apple-ios-sim`
- `x86_64-apple-ios`.

Only the `aarch64-apple-ios` target can be kept, according to https://github.com/matrix-org/matrix-rust-sdk/pull/3334/files#diff-3bbe5c410acf8c549fb356c29229df7b45a37e8db0932b773ce6294cc3935262R44.

Hywan (changes requested):
A tiny change and we are good! You don't need to add all these targets for Rust with `rustup add target …` because you run `build_crypto_xcframework.sh` with `-i`, so only for iOS.

.github/workflows/bindings_ci.yml
BillCarsonFr: Done https://github.com/matrix-org/matrix-rust-sdk/pull/3334/commits/b430d95c0aab174a471577f262e91e5a38defa77

Hywan (approved):
LGTM, kudos!

PR #3337, opened 2024-04-16
crypto: UtdCause enum in reporting hooks and encryption event

andybalaam:
We want to decorate messages we can't decrypt based on _why_ we can't decrypt them. This PR detects the `membership` info provided if the server implements [MSC4115](https://github.com/matrix-org/matrix-spec-proposals/pull/4115) and provides that feedback to the client via the `UtdCause` enum, which we expect to expand in the future.


Hywan (approved):
I have nothing to say about this patch; it is just great. Nice code, nice commits, nice tests, thanks!

Issue #3110, opened 2024-02-08
Olm sessions can wedge when using multiple Clients (e.g with NSE process)


kegsay:
I've been looking into bug reports on Element X, and this appears on iOS: (snipped a lot of data)
```
App:
2024-02-06T07:23:56.293963Z DEBUG matrix_sdk_crypto::machine: Received a to-device event | crates/matrix-sdk-crypto/src/machine.rs:1117 | 
2024-02-06T07:23:56.296398Z DEBUG matrix_sdk_crypto::machine: Received a decrypted to-device event | crates/matrix-sdk-crypto/src/machine.rs:975 | 
2024-02-06T07:23:56.298524Z INFO matrix_sdk_crypto::machine: Received a new megolm room key | crates/matrix-sdk-crypto/src/machine.rs:755 | 
...
2024-02-06T07:23:56.300586Z DEBUG matrix_sdk_crypto::machine: Received a to-device event | crates/matrix-sdk-crypto/src/machine.rs:1117 | 
2024-02-06T07:23:56.301713Z DEBUG matrix_sdk_crypto::machine: Received a decrypted to-device event | crates/matrix-sdk-crypto/src/machine.rs:975 | 
2024-02-06T07:23:56.302390Z INFO matrix_sdk_crypto::machine: Received a new megolm room key | crates/matrix-sdk-crypto/src/machine.rs:755 | 

NSE:
2024-02-06T07:46:24.132801Z DEBUG matrix_sdk_crypto::machine: Received a to-device event | crates/matrix-sdk-crypto/src/machine.rs:1117 | 
2024-02-06T07:46:24.133377Z DEBUG matrix_sdk_crypto::olm::account: Creating a new Olm session from a pre-key message | crates/matrix-sdk-crypto/src/olm/account.rs:953 | 
2024-02-06T07:46:24.133655Z WARN matrix_sdk_crypto::olm::account: Failed to create a new Olm session from a pre-key message: InboundCreation(MissingOneTimeKey("curve25519:3ZuWFzabqiB1bYH1iE/dys5sCxSh8/8KKgo1iNZxsE8")) | crates/matrix-sdk-crypto/src/olm/account.rs:1220 | 
[...]

PR #3338, opened 2024-04-18
bugfix: ensure the SessionStore is cleared when regenerating the OlmMachine

kegsay:
This fixes https://github.com/matrix-org/matrix-rust-sdk/issues/3110

### DIFF
--- a/crates/matrix-sdk-base/src/client.rs
+++ b/crates/matrix-sdk-base/src/client.rs
@@ -1697,6 +1697,36 @@ mod tests {
         client.get_room(room_id).expect("Just-created room not found!")
     }
 
+    #[cfg(feature = "e2e-encryption")]
+    #[async_test]
+    async fn test_regerating_olm_clears_store_caches() {
+        // See https://github.com/matrix-org/matrix-rust-sdk/issues/3110
+        // We must clear the store cache when we regenerate the OlmMachine
+        // to ensure we really get the new state.
+
+        use ruma::{owned_device_id, owned_user_id};
+
+        use crate::store::StoreConfig;
+
+        // Given a client using a fake store
+        let user_id = owned_user_id!("@u:m.o");
+        let device_id = owned_device_id!("DEVICE");
+        let fake_store = fake_crypto_store::FakeCryptoStore::default();
+        let store_config = StoreConfig::new().crypto_store(fake_store.clone());
+        let client = BaseClient::with_store_config(store_config);
+        client.set_session_meta(SessionMeta { user_id, device_id }).await.unwrap();
+        fake_store.clear_method_calls();
+
+        // When we regenerate the OlmMachine
+        client.regenerate_olm().await.expect("Failed to regenerate olm");
+
+        // Then we cleared the store cache
+        assert!(
+            fake_store.method_calls().contains(&"clear_caches".to_owned()),
+            "No clear_caches call!"
+        );
+    }
+
     #[async_test]
     async fn test_deserialization_failure() {
         let user_id = user_id!("@alice:example.org");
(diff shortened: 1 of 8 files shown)


andybalaam:
@poljar I approve this code but can't review as I am a co-author

crates/matrix-sdk-indexeddb/src/crypto_store/mod.rs
```diff
@@ -1273,6 +1273,12 @@ impl_crypto_store! {
             }
         }
     }
+
+    async fn clear_caches(&self) {
+        self.session_cache.clear()
+        // We don't need to clear static_account as it only contains immutable data
```
Hywan, line 1279:
```suggestion
        // We don't need to clear `static_account` as it only contains immutable data
```

crates/matrix-sdk-sqlite/src/crypto_store.rs
```diff
@@ -682,6 +682,13 @@ impl SqliteObjectCryptoStoreExt for deadpool_sqlite::Object {}
 impl CryptoStore for SqliteCryptoStore {
     type Error = Error;
 
+    async fn clear_caches(&self) {
+        self.session_cache.clear()
+        // We don't need to clear static_account as it only contains immutable
```
Hywan, line 687:
```suggestion
        // We don't need to clear `static_account` as it only contains immutable
```

poljar (changes requested):
All of this looks good, but surely we can add some tests for this.

Hywan (commented):
Thanks for the PR!

The code looks good. How hard is it to add a test to ensure the caches are cleared as expected?

kegsay:
The current test for this is https://github.com/matrix-org/complement-crypto/pull/52/files#diff-fd8835211753c17a95b8c01292df184e7a252c2544f0535baf79ea9db1d56d7cR243 which is what we used to verify the fix. We can probably write a unit test though for clear_caches?

andybalaam:
I am working on a couple of tests.

crates/matrix-sdk-indexeddb/src/crypto_store/mod.rs
bnjbvr: no, the static account only contains immutable data about the account. The parts of the account changing are in the store cache, that's correctly reset upon regeneration of the olm machine.

andybalaam:
Testing the `clear_caches()` implementation on the sqlite and indexeddb stores is difficult since the cache is not observable via external interfaces, and we need a real DB to run this code. I don't want to add to the `CryptoStore` interface just to test this.

I've added a test that `regenerate_olm()` calls `clear_caches()`.

A real test will bring up two store instances against the same backing store and check that they keep up to date with each other.

I HEREBY PINKIE PROMISE TO WRITE A GENUINE REGRESSION TEST FOR [#3110](https://github.com/matrix-org/matrix-rust-sdk/issues/3110) OR FAIL TRYING.

But that will be a later PR.

poljar (changes requested):
I mean this is a test, so you got me there...

Sorry but can we have a test that puts a session into the store and cache and then when we call the method we check that the cache is empty?

As we agreed, porting the complement test isn't necessary but please at least something that doesn't just check that a method gets called.

poljar:
> Testing the clear_caches() implementation on the sqlite and indexeddb stores is difficult since the cache is not observable via external interfaces, and we need a real DB to run this code. I don't want to add to the CryptoStore interface just to test this.

To help you out, you don't need a new interface, the field is available inside the store specific tests:

```diff
diff --git a/crates/matrix-sdk-sqlite/src/crypto_store.rs b/crates/matrix-sdk-sqlite/src/crypto_store.rs
index 612165662..80a5fec40 100644
--- a/crates/matrix-sdk-sqlite/src/crypto_store.rs
+++ b/crates/matrix-sdk-sqlite/src/crypto_store.rs
@@ -1304,7 +1304,11 @@ mod tests {
 
 #[cfg(test)]
 mod encrypted_tests {
-    use matrix_sdk_crypto::{cryptostore_integration_tests, cryptostore_integration_tests_time};
+    use matrix_sdk_crypto::{
+        cryptostore_integration_tests, cryptostore_integration_tests_time,
+        store::{Changes, CryptoStore as _, PendingChanges},
+    };
+    use matrix_sdk_test::async_test;
     use once_cell::sync::Lazy;
     use tempfile::{tempdir, TempDir};
 
@@ -1321,6 +1325,26 @@ mod encrypted_tests {
             .expect("Can't create a passphrase protected store")
     }
 
+    #[async_test]
+    async fn cache_cleared() {
+        let store = get_store("cache_cleaerd", None).await;
+        // Given we created a session and saved it in the store
+        let (account, session) = cryptostore_integration_tests::get_account_and_session().await;
+        let sender_key = session.sender_key.to_base64();
+
+        store
+            .save_pending_changes(PendingChanges { account: Some(account.deep_clone()) })
+            .await
+            .expect("Can't save account");
+
+        let changes = Changes { sessions: vec![session.clone()], ..Default::default() };
+        store.save_changes(changes).await.unwrap();
+
+        store.session_cache.get(&sender_key).expect("We should have a session");
+
+        // TODO: Clear the cache
+    }
+
     cryptostore_integration_tests!();
     cryptostore_integration_tests_time!();
 }
```

BillCarsonFr (approved):
LGTM

poljar (approved):
I find the other test a bit pointless, but ok if we manage to port the complement test, then we can remove it.

crates/matrix-sdk-base/src/client.rs
```diff
@@ -1695,6 +1697,36 @@ mod tests {
         client.get_room(room_id).expect("Just-created room not found!")
     }
 
+    #[cfg(feature = "e2e-encryption")]
+    #[async_test]
+    async fn test_regerating_olm_clears_store_caches() {
```
bnjbvr, line 1702: Not sure this test is really useful? This is using a fake crypto store impl, and just checking that a method has been called, which is better tested in the other test. How do y'all feel about removing this test instead and all the bootstrapping that it requires, which is a bit noisey to me?

bnjbvr (approved):

bnjbvr:
E_TOO_SLOW

crates/matrix-sdk-base/src/client.rs
bnjbvr: https://github.com/matrix-org/matrix-rust-sdk/pull/3342

poljar: I agreed as much in my final review, it is a bit useful since it's the only place which checks that the method gets called.

Porting the complement crypto test is still in the pinkie promise so that might get rid of this noisy code.

PR #3339, opened 2024-04-18
room preview: use the room summary MSC3266 endpoint

bnjbvr:
This adds support for MSC3266 when fetching a room preview, for servers that support it. If the server doesn't support it, we resort to using the room state event endpoint.

Pretty sure the integration testing in CI will hate this, since I'm unit-testing the `from_room_summary` method, which requires an experimental feature to be enabled on the synapse service. If that's the case, I'll disable this test.

Using a fork of Ruma to cut corners, since the review of the feature there is getting complicated. Maybe https://github.com/ruma/ruma/pull/1776 will get approved some day soon :smiling_face_with_tear:

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -39,15 +39,16 @@ futures-util = { version = "0.3.26", default-features = false, features = [
 http = "0.2.6"
 imbl = "2.0.0"
 itertools = "0.12.0"
-ruma = { git = "https://github.com/ruma/ruma", rev = "4c00bd010dbdca6005bd599b52e90a0b7015d056", features = [
+ruma = { git = "https://github.com/ruma/ruma", rev = "21b644ac6ae1c7d4a4f7e98a6481a3318f2deeaa", features = [
     "client-api-c",
     "compat-upload-signatures",
     "compat-user-id",
     "compat-arbitrary-length-ids",
     "compat-tag-info",
     "unstable-msc3401",
+    "unstable-msc3266",
 ] }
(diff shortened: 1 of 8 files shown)


bnjbvr:
> Pretty sure the integration testing in CI will hate this, since I'm unit-testing the from_room_summary method, which requires an experimental feature to be enabled on the synapse service. If that's the case, I'll disable this test.

Bingo, so disabling the test on CI.

bnjbvr:
@Hywan Can haz review, plz? Okthxbye :pleading_face:

Hywan:
https://github.com/ruma/ruma/pull/1776 has been merged 😉.

Cargo.toml
```diff
@@ -39,15 +39,16 @@ futures-util = { version = "0.3.26", default-features = false, features = [
 http = "0.2.6"
 imbl = "2.0.0"
 itertools = "0.12.0"
-ruma = { git = "https://github.com/ruma/ruma", rev = "4c00bd010dbdca6005bd599b52e90a0b7015d056", features = [
+ruma = { git = "https://github.com/ruma/ruma", rev = "21b644ac6ae1c7d4a4f7e98a6481a3318f2deeaa", features = [
```
Hywan, line 42: Please use the original repository since your PR has been merged.

Hywan (approved):
Looks good to me. Please fix the repository of Ruma before merging this PR.

Cargo.toml
bnjbvr: Agreed to use the original repo, which I've done in a commit this morning,… so there's nothing to change, right? :grin:

PR #3341, opened 2024-04-18
fix(ui): `Timeline::send_reply` correctly sets up `m.mentions`

### LINKED ISSUE #3314: When user reply to their message, the reply mentions the current user
Originally opened on Ruma project: https://github.com/ruma/ruma/issues/1770 (see details there), but the fix seems to be located in the Rust project.

Hywan:
In https://github.com/matrix-org/matrix-rust-sdk/pull/2691, I suppose
the way `add_mentions` is computed is… wrong. [`AddMentions`](https://docs.rs/ruma/latest/ruma/events/room/message/enum.AddMentions.html) is used to
automatically infer the [`m.mentions`](https://spec.matrix.org/v1.10/client-server-api/#definition-mmentions) of the reply event based on the
replied event. The way it was computed was based on the reply event
`mentions`, which seems wrong: if the reply contains mentions, then the
sender should be part of it? Nah. That's a bug. We want the reply event
to automatically mention the sender of the replied event if and only
if it's not the same as the current user, i.e. the sender of the reply
event.

This patch fixes the `add_mentions` calculation. This patch also updates
a test and adds another test to ensure that `m.mentions` is correctly
defined when replying to an event.

---

* Fix #3314 
* Address https://github.com/element-hq/element-x-android/issues/2671

### DIFF
--- a/crates/matrix-sdk-ui/src/timeline/mod.rs
+++ b/crates/matrix-sdk-ui/src/timeline/mod.rs
@@ -281,14 +281,13 @@ impl Timeline {
 
     /// Send a reply to the given event.
     ///
-    /// Currently only supports events events with an event ID and JSON being
+    /// Currently it only supports events with an event ID and JSON being
     /// available (which can be removed by local redactions). This is subject to
     /// change. Please check [`EventTimelineItem::can_be_replied_to`] to decide
     /// whether to render a reply button.
     ///
-    /// If the `content.mentions` is `Some(_)`, the sender of `reply_item` will
-    /// be added to the mentions of the reply. If `content.mentions` is `None`,
-    /// it will be kept as-is.
+    /// The sender of `reply_item` will be added to the mentions of the reply if
+    /// and only if `reply_item` has not been written by the sender.
     ///
     /// # Arguments
     ///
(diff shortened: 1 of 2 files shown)


crates/matrix-sdk-ui/src/timeline/mod.rs
```diff
@@ -312,8 +311,24 @@ impl Timeline {
...
+        // replied event (`reply_item`)…
+        let mut mention_the_sender_of_the_replied_event = AddMentions::Yes;
+
+        // … but if `reply_item` has been written by the current user, let's toggle
+        // to `AddMentions::No`.
+        if let Some(current_user_id) = client.user_id() {
+            if current_user_id == reply_item.sender {
+                mention_the_sender_of_the_replied_event = AddMentions::No;
+            }
+        }
```
bnjbvr, line 331: nit: Maybe we can avoid a bit of verbosity, the `client` local variable, and the `mut` by having something like:

```suggestion
        let mention_sender = if self.room().own_user_id() == reply_item.sender { AddMentions::No } else { AddMentions::Yes };
```

crates/matrix-sdk-ui/src/timeline/mod.rs
```diff
@@ -325,7 +340,11 @@ impl Timeline {
                     content: msg.to_content(),
                     unsigned: Default::default(),
                 };
-                content.make_reply_to(&event, forward_thread, add_mentions)
+                content.make_reply_to(
+                    &event,
+                    forward_thread,
+                    mention_the_sender_of_the_replied_event,
```
bnjbvr, line 346:
```suggestion
                    mention_sender,
```

crates/matrix-sdk-ui/src/timeline/mod.rs
```diff
@@ -337,7 +356,7 @@ impl Timeline {
                     event_id.to_owned(),
                     self.room().room_id(),
                     forward_thread,
-                    add_mentions,
+                    mention_the_sender_of_the_replied_event,
```
bnjbvr, line 359:
```suggestion
                    mention_sender,
```

bnjbvr (approved):
Thanks, that's a sweet test!

crates/matrix-sdk-ui/src/timeline/mod.rs
Hywan: Indeed, that's better, thanks!

PR #3342, opened 2024-04-19
crypto: remove the fake crypto store test for olm session wedging

bnjbvr:
Two reviewers were suspicious the test wasn't very useful, so I hereby propose to remove it.


poljar:
Let's see if we get the comlement test ported, the test has some use before that's achieved.

bnjbvr:
Oh, I thought the other test did test the same effect properly, by adding a session, regenerating the olm machine and checking that the session was then missing from the cache?

poljar:
> Oh, I thought the other test did test the same effect properly, by adding a session, regenerating the olm machine and checking that the session was then missing from the cache?

Well kinda, the other test ensures that the `SqliteCryptoStore` method does the correct thing, this test ensures that the `regenerate_olm()` method will call the method to clear the cache.

bnjbvr:
Ah, so you're saying the `SqliteCryptoStore` could do the correct thing, without the method to clear the cache getting called? Then agreed, we can merge this later.

poljar:
> Ah, so you're saying the `SqliteCryptoStore` could do the correct thing, without the method to clear the cache getting called? Then agreed, we can merge this later.

Yeah, we might accidentally remove the call to this method and the dumb test ensures that we don't do so.

poljar:
This depends on #3343.

bnjbvr:
@poljar now that #3343 has been merged, can haz review plz? :pleading_face:

poljar (approved):
Very elegant removal, more of this please.

PR #3343, opened 2024-04-19
crypto: Add a standalone integration test for the NSE race

andybalaam:
Adds a test that fails before https://github.com/matrix-org/matrix-rust-sdk/pull/3338 and passes afterwards.

Tests the fix for https://github.com/matrix-org/matrix-rust-sdk/issues/3110

The test sets up a cross-process lock and 2 clients using it, then makes sure they both have Olm sessions with a 3rd client, before sending from one of them and checking that the other can still receive. This validates that it flushed and replaced its Olm session with the 3rd client.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -399,7 +399,7 @@ jobs:
 
       - name: Test
         env:
-          RUST_LOG: "hyper=trace"
+          RUST_LOG: "info,matrix_sdk=trace"
           HOMESERVER_URL: "http://localhost:8008"
           HOMESERVER_DOMAIN: "synapse"
           SLIDING_SYNC_PROXY_URL: "http://localhost:8118"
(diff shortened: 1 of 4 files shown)


andybalaam:
Undrafted to get integration test

Hywan:
Do you need help?

andybalaam:
> Do you need help?

Thank you, no I think I'm nearly there - Rich and I made a breakthough yesterday.

andybalaam:
Requesting review from @Hywan since he was initially chosen.

This is (finally) ready for a first look from you, thanks!

Hywan (approved):

crates/matrix-sdk-crypto/src/olm/account.rs
```diff
@@ -1277,7 +1277,7 @@ impl Account {
                         // session in lieu of an OTK.)
 
                         warn!(
-                            session_id = session.session_id(),
+                            ?session,
```
bnjbvr, line 1280: Do you want to keep this logging change?

bnjbvr (commented):
(small driveby, sorry)

crates/matrix-sdk-crypto/src/olm/account.rs
andybalaam: Thank you! Corrected and re-pushed along with a compile fix.

PR #3347, opened 2024-04-22
room-list: exclude non-joined rooms when filtering by rooms, people, unreads and favourites

stefanceriu:
With the addition of `experimental-room-list-with-unified-invites` we now need to be more strict with what types of rooms we allow in the room list when applying filters.

### DIFF
--- a/crates/matrix-sdk-ui/src/room_list_service/filters/category.rs
+++ b/crates/matrix-sdk-ui/src/room_list_service/filters/category.rs
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 use matrix_sdk::{Client, RoomListEntry};
+use matrix_sdk_base::RoomState;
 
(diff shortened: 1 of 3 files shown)


Hywan (changes requested):
Thanks for you work! I am however wondering if it's not better to add a `join` filter, just like the `invite` filter, and then use logic combinators like `all(join, unread)`, it's seem more modular this way, thoughts?

Why am I proposing this? Because with this PR, we cannot use the `unread`, `category` and the `favorite` filter with `invite` (or later `left` maybe) anymore. It doesn't make a lot of sense for `unread` and `favorite`, but it makes sense for `category`.

stefanceriu:
> it's seem more modular this way, thoughts?

Yeah, that sounds like a great idea, don't know why I didn't think about it myself 🤦

PR #3349, opened 2024-04-22
paginator: select how many events to retrieve during initial /context and pagination queries

bnjbvr:
Does what it says on the tin. Mostly test changes.

Supports https://github.com/matrix-org/matrix-rust-sdk/issues/3234


Hywan (approved):
LGTM!

PR #3350, opened 2024-04-22
timeline: include the remote event's origin in the timeline position/end

bnjbvr:
And don't assume inserting to the end means it's coming from sync — as it won't be true for forward pagination anymore.

Supports https://github.com/matrix-org/matrix-rust-sdk/issues/3234, extracted from https://github.com/matrix-org/matrix-rust-sdk/pull/3329.


andybalaam (approved):
Thanks!

PR #3351, opened 2024-04-22
timeline: stringify the internal ID and allow setting a prefix to the internal ID

bnjbvr:
This will be handy to distinguish IDs coming from different timelines, and this will be useful for the permalink PR, where the "main"/live timeline may not have a prefix, but detached timelines may have one. With something like that, the apps will be able to transition from a detached timeline to the new one, and having different internal IDs for the events mean that their UI frameworks will be able to notice the change.

Supports #3234.


poljar (approved):
I just hope that this won't be abused somehow to include metadata.

Makes sense.

PR #3352, opened 2024-04-22
timeline: prevent deadlock in `replace_with_initial_events`

bnjbvr:
:warning: This is preventing timelines to be built with read marker support, if there was a fully read marker event stored in the database.

The `state` lock was taken at the top level of this function, and indirectly implicitly in the `set_fully_read_event` function. This fixes it, and adds a regression test.


Hywan (approved):
Thanks for the test!

PR #3353, opened 2024-04-23
timeline: prevent deadlock in `populate_initial_user_receipt`

bnjbvr:
Follow-up to 13cc7962. Same thing would happen when populating the initial user receipts, with the `populate_initial_user_receipt` re-attempting to take the state lock while it was already taken.

Includes a regression test.


andybalaam (approved):
Looks good, thanks!

Issue #3255, opened 2024-03-20
Upgrade reqwest to 0.12


zecakeh:
This new version just came out and uses the new major version 1.x of the http crate.

Dependencies still depending on older versions of reqwest/http:

- [ ] mas-oidc-client (for the `experimental-oidc` feature): https://github.com/matrix-org/matrix-authentication-service/issues/2530
- [x] ruma: https://github.com/ruma/ruma/issues/1748

zecakeh:
This was done in #3362. Should we keep this open to remember to track the status of http in matrix-authentication-service?

bnjbvr:
We can close this one, and get back to the matrix-authentication-service one later; I suspect that as soon as they've bumped the http crate, we'll eventually get a (failing) PR from dependabot.

Thanks for your work here Kevin!

PR #3357, opened 2024-04-27
sdk: Replace hyper with axum for SSO login

zecakeh:
`hyper::Server` was dropped in hyper 1, without a replacement. If we want to have all crates depending on http 1 in the end, we need to replace it.

Related to #3255.


Hywan (approved):
Looks good to me, thanks!

PR #3358, opened 2024-04-29
FFI: Allow the creation of a matrix.to link for any room alias.

pixlwave:
Whilst the FFI currently supports creating a matrix.to room alias link directly from a `Room`, there isn't a way to do so when you already have one as a string. This PR fixes that by adding a global `matrix_to_room_alias_permalink` method similar to the existing `matrix_to_user_permalink` method. (This is only useful against an alias, as a plain room ID doesn't provide enough info to construct the URL with the necessary via parameters).

### DIFF
--- a/bindings/matrix-sdk-ffi/src/room.rs
+++ b/bindings/matrix-sdk-ffi/src/room.rs
@@ -18,7 +18,7 @@ use ruma::{
         },
         TimelineEventType,
     },
-    EventId, Int, UserId,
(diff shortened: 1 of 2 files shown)


andybalaam (approved):
Looks in line with `matrix_to_user_permalink` so seems good to me. Thanks!

pixlwave:
Thanks for the super quick review 🙌

Issue #3355, opened 2024-04-25
fix(ui): Pagination clean up in the timeline


Hywan:
In https://github.com/matrix-org/matrix-rust-sdk/pull/3329, there is 2 small problems that must be addressed:

* When a pagination is cancelled, the status isn't reset. Typically, it starts with `Idle`, then a pagination is fired, the status becomes `Paginating`, then it's cancelled, the status stays `Paginating`. When restarting the pagination, it returns immediately because the system believes it's already paginating. We must address that.
* The pagination code seems partially duplicated between `timeline::pagination` and `event_cache::paginator`. We must reconcile both.

bnjbvr:
Thanks for opening the follow up.

PR #3359, opened 2024-04-29
paginator: reset the paginator's state if the task is cancelled

bnjbvr:
This makes all the requests (/context and /messages) cancellation-safe by making two changes:

- first, use sync locks instead of async locks for the prev/next batch tokens. This ensures that we can't get cancelled after receiving a response and managing internal state, i.e. we keep run-to-completion semantics until the end of the function, once we got a response.
- second, introduce a RAII guard that will reset the state to a given value when dropped. Then use this to reset the state to Initial (resp. Idle) in /context (resp. /messages) when the async call is aborted. We use `mem::forget` once the response has been returned, so as to not call the `Drop` implementation of the guard later on.

A regression test has also been introduced.

Part of https://github.com/matrix-org/matrix-rust-sdk/issues/3355.

### DIFF
--- a/crates/matrix-sdk/src/event_cache/paginator.rs
+++ b/crates/matrix-sdk/src/event_cache/paginator.rs
@@ -17,7 +17,7 @@
 //! makes it possible to paginate forward or backward, from that event, until
 //! one end of the timeline (front or back) is reached.
 
-use std::{mem, sync::Mutex};
+use std::sync::Mutex;
 
 use eyeball::{SharedObservable, Subscriber};
 use matrix_sdk_base::{deserialized_responses::TimelineEvent, SendOutsideWasm, SyncOutsideWasm};
@@ -138,13 +138,27 @@ pub struct StartFromResult {
 
 /// Reset the state to a given target on drop.
 struct ResetStateGuard {
-    target: PaginatorState,
+    target: Option<PaginatorState>,
     state: SharedObservable<PaginatorState>,
+}
+
+impl ResetStateGuard {
+    /// Create a new reset state guard.
+    fn new(target: PaginatorState, state: SharedObservable<PaginatorState>) -> Self {
+        Self { target: Some(target), state }
+    }
+
+    /// Render the guard effectless, and consume it.
+    fn disarm(mut self) {
+        self.target = None;
+    }
 }
 
 impl Drop for ResetStateGuard {
     fn drop(&mut self) {
-        self.state.set_if_not_eq(self.target);
+        if let Some(target) = self.target.take() {
+            self.state.set_if_not_eq(target);
+        }
     }
 }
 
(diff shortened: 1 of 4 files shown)


crates/matrix-sdk/src/event_cache/paginator.rs
```diff
@@ -267,25 +294,28 @@ impl Paginator {
...
+        // don't want the task to be interrupted anymore, or the internal state
+        // may be incorrect.
 
         let hit_end_of_timeline = response.end.is_none();
-        *token_lock.lock().await = response.end;
+        *token_lock.lock().unwrap() = response.end;
 
         // TODO: what to do with state events?
 
+        // Forget the reset state guard, so its Drop method is not called.
```
poljar, line 316: It sounds like you'd want to use `ManuallyDrop` instead? I'm not sure, but it's not out of the question that you have some heap allocated data inside the guard and `forget` will not deallocate those:

> Takes ownership and “forgets” about the value without running its destructor.
> 
> Any resources the value manages, such as heap memory or a file handle, will linger forever in an unreachable state. However, it does not guarantee that pointers to this memory will remain valid.

crates/matrix-sdk/src/event_cache/paginator.rs
```diff
@@ -179,17 +196,27 @@ impl Paginator {
...
+        // NOTE: it's super important to not have any `await` after this point, since we
+        // don't want the task to be interrupted anymore, or the internal state
+        // may become incorrect.
+
         let has_prev = response.prev_batch_token.is_some();
         let has_next = response.next_batch_token.is_some();
-        *self.prev_batch_token.lock().await = response.prev_batch_token;
-        *self.next_batch_token.lock().await = response.next_batch_token;
+        *self.prev_batch_token.lock().unwrap() = response.prev_batch_token;
+        *self.next_batch_token.lock().unwrap() = response.next_batch_token;
```
poljar, line 215: Do we want these two to perhaps share the lock?

crates/matrix-sdk/src/event_cache/paginator.rs
```diff
@@ -179,17 +196,27 @@ impl Paginator {
...
 
+        // Forget the reset state guard, so its Drop method is not called.
+        mem::forget(reset_state_guard);
```
poljar, line 218: Like in the other comment, I don't think we should use `forget()` here. Or at all in the crate for that matter.

poljar (changes requested):

crates/matrix-sdk/src/event_cache/paginator.rs
bnjbvr: Hmm, we're very close to being able to paginate backwards *while* we paginate forwards, by using two locks… I'd like to keep it that way and implement that later on.

bnjbvr: I realize it's probably premature optimization, but also it's nice for the `paginate()` internal helper to have two locks. Considering they're all fine-grained and shouldn't be contended in general, seems OK to keep them both?

poljar: Sure, I just wasn't sure why we'd not have two locks considering the above code snippet.

poljar (approved):

PR #3362, opened 2024-04-30
Upgrade http, ruma, reqwest and wiremock dependencies

zecakeh:
They need to be updated together because the latters depend on the former.

matrix-authentication-service is still using http 0.2 so we need to add a conversion layer between both major versions for OIDC requests for now.

We need to update vodozemac too because of a dependency resolution issue.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -11,7 +11,7 @@ dependencies = [
  "macroific",
  "proc-macro2",
  "quote",
- "syn 2.0.48",
+ "syn 2.0.60",
 ]
 
 [[package]]
@@ -57,7 +57,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "77c3a9648d43b9cd48db467b3f87fdd6e146bcc88ab0180006cef2179fe11d01"
 dependencies = [
  "cfg-if",
- "getrandom 0.2.12",
+ "getrandom",
  "once_cell",
  "version_check",
  "zerocopy",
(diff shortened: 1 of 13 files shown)


bnjbvr (commented):
I'd prefer that we'd wait for the matrix-authentication-service crate to update its HTTP crate instead of adding all the HttpOld/HttpNew conversation stuff.

Out of curiosity: are you making this PR because you require it for some reasons, or is it only careful maintenance? (in any case, thanks for opening such a PR!)

bnjbvr (approved):
Actually it's a nice step forward as it removes a bunch of duplicated crates in the dependency tree, and we can easily get rid of the http_old conversion code later on. Thanks!

poljar:
> Actually it's a nice step forward as it removes a bunch of duplicated crates in the dependency tree, and we can easily get rid of the http_old conversion code later on. Thanks!

For reference, I looked into updating MAS itself, but it's blocked by https://github.com/open-telemetry/opentelemetry-rust/pull/1674 getting merged and released.

zecakeh:
> Out of curiosity: are you making this PR because you require it for some reasons, or is it only careful maintenance? (in any case, thanks for opening such a PR!)

No we don't require it for the moment, I just knew it was probably going to be a big change, so I wanted to get a head start.

> For reference, I looked into updating MAS itself, but it's blocked by https://github.com/open-telemetry/opentelemetry-rust/pull/1674 getting merged and released.

Yeah we have been waiting a bit to update http in Ruma so we could avoid the conversion layer for the SDK, but we would like to make a new release soon and we wanted to upgrade http before it.

zecakeh:
Merged main branch to resolve conflict

PR #3305, opened 2024-04-05
sdk: Use the `GET /auth_issuer` endpoint for OIDC

zecakeh:
The well-known method is deprecated.

This removes some tests on methods that were previously only getters but now make a request to the auth issuer.

Tested the example CLI to work with the OIDC playground. Element X should probably be tested with other deployments to confirm this. There should be no code changes needed with the FFI bindings.

PR #3363, opened 2024-04-30
 oidc: Use the correct types to compare the status codes in the oidc example

poljar:
Something went wrong while #3305 was merged. This fixes the issue introduced by the github rebase.

### DIFF
--- a/examples/oidc_cli/src/main.rs
+++ b/examples/oidc_cli/src/main.rs
@@ -23,9 +23,12 @@ use std::{
 };
 
 use anyhow::{anyhow, bail};
-use axum::{response::IntoResponse, routing::any_service};
+use axum::{
+    http::{Method, StatusCode},
+    response::IntoResponse,
+    routing::any_service,
+};
 use futures_util::StreamExt;
-use http::{Method, StatusCode};
 use matrix_sdk::{
     config::SyncSettings,
     oidc::{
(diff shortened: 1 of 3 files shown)


zecakeh:
I believe this should not be necessary if #3362 is merged.

zecakeh:
Actually, we could fully qualify `StatusCode` in this case with `matrix_sdk::reqwest::StatusCode`.

poljar:
> I believe this should not be necessary if #3362 is merged.

Hah yeah, though I think it makes sense to fix the bodged merge of #3305 with the smallest possible patch.

poljar:
> Actually, we could fully qualify `StatusCode` in this case with `matrix_sdk::reqwest::StatusCode`.

You mean post merge or in this PR?

zecakeh:
> You mean post merge or in this PR?

In this PR, that should work at any time.

 Actually the best option would probably be to not depend directly on http, but use `axum::http` types for interactions with axum types, and `matrix::reqwest` types for interacting with the SDK's types.

bnjbvr (approved):

poljar:
> > You mean post merge or in this PR?
> 
> In this PR, that should work at any time.
> 
> Actually the best option would probably be to not depend directly on http, but use `axum::http` types for interactions with axum types, and `matrix::reqwest` types for interacting with the SDK's types.

Right: https://github.com/matrix-org/matrix-rust-sdk/pull/3363/commits/344c77bcec6c16b2e48c62fbb11d9a30db20be7d

examples/oidc_cli/src/main.rs
```diff
@@ -23,9 +23,12 @@ use std::{
 };
 
 use anyhow::{anyhow, bail};
-use axum::{response::IntoResponse, routing::any_service};
+use axum::{
+    http::{Method, StatusCode},
```
zecakeh, line 27: The http dependency needs to be removed from the `Cargo.toml`.

PR #3364, opened 2024-04-30
ci: Upgrade most actions

zecakeh:
Should get rid of most warnings in CI.

I took the liberty to replace 2 unmaintained actions, these commits can be reverted if that's not wanted.

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -71,7 +71,7 @@ jobs:
 
     steps:
       - name: Checkout repository
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
         with:
           ref: ${{ github.event.pull_request.head.sha }}
 
@@ -94,7 +94,7 @@ jobs:
           tool: cargo-tarpaulin
 
       # set up backend for integration tests
-      - uses: actions/setup-python@v4
+      - uses: actions/setup-python@v5
         with:
           python-version: 3.8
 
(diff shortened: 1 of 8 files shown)


.github/workflows/coverage.yml
```diff
@@ -111,7 +111,7 @@ jobs:
           SLIDING_SYNC_PROXY_URL: "http://localhost:8118"
 
       - name: Upload to codecov.io
-        uses: codecov/codecov-action@v4
+         codecov/codecov-action@v4
```
poljar, line 114: v4 doesn't support tokenless uploads anymore: https://github.com/codecov/codecov-action?tab=readme-ov-file#breaking-changes.

I added the token using the `CODECOV_TOKEN` name as described in: https://docs.codecov.com/docs/adding-the-codecov-token#github-actions.

Could you please add it as described over here: https://github.com/codecov/codecov-action?tab=readme-ov-file#usage.

poljar (changes requested):

.github/workflows/coverage.yml
zecakeh: Wait I didn't actually change the action version. But strangely the `uses:` was removed here.

zecakeh: Fixed

poljar: Ah indeed.

zecakeh:
Rebased on main for CI

poljar (approved):

PR #3365, opened 2024-04-30
timeline: reset pagination status if a live back-pagination is aborted

bnjbvr:
Repetition of https://github.com/matrix-org/matrix-rust-sdk/pull/3359 but at a different layer: the timeline has its own observable state for a back-pagination status, and we should *also* reset it when using back-pagination on a live timeline.

Instead of doing this, we should probably have the status + the pagination being handled in the event cache, using the new Paginator API, but that's a bigger change that I'm working on, and it changes many many public APIs related to the event cache, so I'm slightly postponing that and instead I propose to get this simpler fix merged.

Part of #3355.

### DIFF
--- a/crates/matrix-sdk-ui/src/timeline/pagination.rs
+++ b/crates/matrix-sdk-ui/src/timeline/pagination.rs
@@ -14,12 +14,35 @@
 
 use std::{fmt, ops::ControlFlow, sync::Arc, time::Duration};
 
-use eyeball::Subscriber;
+use eyeball::{SharedObservable, Subscriber};
 use matrix_sdk::event_cache::{self, BackPaginationOutcome};
 use tracing::{instrument, trace, warn};
 
 use super::Error;
 use crate::timeline::{event_item::RemoteEventOrigin, inner::TimelineEnd};
+
+struct ResetStatusGuard {
+    status: SharedObservable<PaginationStatus>,
+    target: Option<PaginationStatus>,
+}
+
+impl ResetStatusGuard {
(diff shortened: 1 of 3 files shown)


crates/matrix-sdk-ui/src/timeline/pagination.rs
```diff
@@ -14,13 +14,36 @@
...
 use super::Error;
 use crate::timeline::{event_item::RemoteEventOrigin, inner::TimelineEnd};
 
+struct ResetStatusGuard {
+    status: SharedObservable<PaginationStatus>,
+    target: Option<PaginationStatus>,
+}
+
+impl ResetStatusGuard {
+    fn new(status: SharedObservable<PaginationStatus>, target: PaginationStatus) -> Self {
```
andybalaam, line 30: Any reason why target is the first argument here: https://github.com/matrix-org/matrix-rust-sdk/pull/3359/files#diff-a6ab59f890b91b82d1648c9c8bbd78c6387ece6d89f8a30683fe941be504628eR147 and the second in this code?

If not, I'd recommend keeping the order consistent.

andybalaam (approved):
Looks good, one suggestion. Thanks!

crates/matrix-sdk-ui/src/timeline/pagination.rs
bnjbvr: There's no reason, so you're right, we should make it consistent. I slightly prefer state first, then target, so I'll switch the other one around. Thanks!

PR #3367, opened 2024-04-30
chore: Upgrade dependencies

zecakeh:
base64 now matches the version used in ruma and vodozemac.

I saw other possible upgrades but they created duplicate entries in `Cargo.lock`, so I tried to be conservative, and ended up only upgrading ~~this one~~ these.

Edit: It turns out I missed the update of async-channel which allows to update event-listener too.


Hywan (approved):
Looks good to me, thank you very much for handling this!

PR #3368, opened 2024-04-30
ffi: remove usage of `block_on` in the FFI layer

bnjbvr:
Does what it says on the tin. Let's see how it goes!


Hywan (approved):
Oh yes, that's so good! Thank you so much for doing this, that's something I want to do since a long time but failed to because of lack of time.

pixlwave:
Please could we keep roomItem `roomListItem.name`, `room.isDirect` `room.isEncrypted` as sync methods somehow?

PR #3369, opened 2024-04-30
notification client: get rid of builder

bnjbvr:
The builder had only one meaningful method, `filter_by_push_rules`, which was always called by the applications — and in fact should always be true. It was designed as an extra method because it was experimental at the time, but it's stabilized sufficiently that we can enable this behavior by default now, considering that a notification that is not wanted by the user shouldn't be kept, to respect their intent. (This is in the UI crate, which is opinionated, so it's fine to assume such intents by design.)


Hywan (approved):
I agree with these changes, thanks for the cleanup!

PR #3370, opened 2024-05-01
ffi: revert some functions back to sync

bnjbvr:
Reverts a few changes of https://github.com/matrix-org/matrix-rust-sdk/pull/3368, and also simplify / add more comments / etc.

One notable change: in `Room`, `RoomInfo` and `RoomListItem`, the `name()` method will now always return the computed display name only, and not alternate between the computed display name and the raw name from the room state event. Let me know if that's something we don't want.

### DIFF
--- a/bindings/matrix-sdk-ffi/src/room.rs
+++ b/bindings/matrix-sdk-ffi/src/room.rs
@@ -210,8 +210,11 @@ impl Room {
         Ok(Timeline::new(timeline))
     }
 
-    pub async fn display_name(&self) -> Result<String, ClientError> {
-        Ok(self.inner.display_name().await?.to_string())
+    /// Returns the room's name from the state event if available, otherwise
+    /// compute a room name based on the room's nature (DM or not) and number of
+    /// members.
+    pub fn computed_display_name(&self) -> Result<String, ClientError> {
+        Ok(RUNTIME.block_on(self.inner.computed_display_name())?.to_string())
     }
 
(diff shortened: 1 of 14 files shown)


Hywan (approved):
Looks sensible. Thanks for the doc improvements!

bindings/matrix-sdk-ffi/src/room.rs
```diff
@@ -77,8 +77,11 @@ impl Room {
         self.inner.room_id().to_string()
     }
 
-    pub fn name(&self) -> Option<String> {
-        self.inner.name()
+    /// Returns the room's name from the state event if available, otherwise
+    /// compute a room name based on the room's nature (DM or not) and number of
+    /// members.
+    pub fn name(&self) -> Result<String, ClientError> {
+        Ok(RUNTIME.block_on(self.inner.computed_display_name())?.to_string())
```
bmarty, line 84: Would it be possible to have both `name` and `display_name` API?

There is a case when we want to use the name from the state event: editing the room name.

Currently we use the computed name, which is not correct in this case:

image

So from the POV of a developer who use the SDK with some MAtrix knowledge background, having two API like:
- `name(): String?` the raw name, i.e. the name from the `m.room.name` state event, can be null
- `displayName(): String`, equals to the raw name if available, or a fallback using heroes
would be awesome.

bnjbvr: Yep, I've added `raw_name()` in the latest iteration. Thanks!

PR #3373, opened 2024-05-01
event cache/timeline: reuse the `Paginator` when running back-paginations

### LINKED ISSUE #3355: fix(ui): Pagination clean up in the timeline

bnjbvr:
## First commit

Sorry, I got a bit ahead of myself and did multiple interesting changes in that large commit:

- In contrary to the pagination happening for focused-mode timelines, the back-pagination happening from the event cache doesn't require focusing on the target event first (which causes the /context query). So the `Paginator` must now be instantiable from that state, with a previous-batch token that's passed by the caller. This justified introducing `Paginator::set_idle_state(prev_batch_token)`.
- Then, since it's possible to give a `prev_batch_token` set to `None` to that function (meaning, we don't have any prev-batch-token from sync for that room), we now need to distinguish `prev_batch_token is set to None by choice` from `prev_batch_token is set to None because we hit the end of the timeline`. This justified the introduction of `paginator::PaginationToken`, which reflects those three states (the third one being `Some(token)`).
- From that, it was possible to reuse the `Paginator` for the room back-paginations. The `RoomEventCache` now also takes care of the pagination token, instead of being handed one, removing some failure modes (e.g. passing an incorrect back-pagination token internally). If it didn't get a proper token from a sync, it'll be in charge of waiting for it (for 3 seconds), so the timeline doesn't have to do that itself anymore.
- The timeline's back-pagination subscriber is now just the one from the room's paginator status, which is nice, because it means it's now shared among all the timeline instances too. (What might be slightly less nice is that it may toggle between Paginating and Idle, if the caller in timeline calls the function multiple times. It seemed OK to me, especially since the timeline will silent multiple requests to back-paginate concurrently.)

## Other commits

The other commits are slightly more atomic and should be easier to digest one by one, by looking at the commit messages.

## TL;DR

This makes it possible for the event cache to back-paginate in the background, even when no timelines are shown, which should be handy for some use cases (esp. around read-receipts).

Fixes #3355.


bindings/matrix-sdk-ffi/src/timeline/mod.rs
```diff
@@ -160,7 +163,7 @@ impl Timeline {
         self.inner.fetch_members().await
     }
 
-    pub fn subscribe_to_back_pagination_status(
+    pub async fn subscribe_to_back_pagination_status(
```
Hywan, line 166: It doesn't seem necessary to have this method `async`. Am I missing something?

crates/matrix-sdk-ui/src/timeline/pagination.rs
```diff
@@ -12,38 +12,19 @@
...
+    self,
+    paginator::{PaginatorError, PaginatorState},
+    BackPaginationOutcome, EventCacheError,
+};
 use tracing::{instrument, trace, warn};
 
 use super::Error;
 use crate::timeline::{event_item::RemoteEventOrigin, inner::TimelineEnd};
 
-struct ResetStatusGuard {
```
Hywan, line 24: Haaaaaa.

crates/matrix-sdk-ui/src/timeline/pagination.rs
```diff
@@ -163,22 +132,20 @@ impl super::Timeline {
                     }
                 }
 
-                // Retrieve the next earliest back-pagination token.
-                token = self.event_cache.oldest_backpagination_token(None).await?;
```
Hywan, line 167: 🫨

crates/matrix-sdk-ui/src/timeline/pagination.rs
```diff
@@ -163,22 +132,20 @@ impl super::Timeline {
...
-        // The status is automatically reset to idle by `reset_status_guard`.
-
         Ok(false)
     }
 
     /// Subscribe to the back-pagination status of the timeline.
-    pub fn back_pagination_status(&self) -> Subscriber<PaginationStatus> {
-        self.back_pagination_status.subscribe()
+    ///
+    /// Note: this may send multiple Paginating/Idle sequences during a single
```
Hywan, line 114: Why? Do we want to dedup them? Maybe with [`async_rx::StreamExt::dedup`](https://docs.rs/async-rx/latest/async_rx/trait.StreamExt.html#method.dedup) if we can provide a `Stream`.

crates/matrix-sdk-ui/src/timeline/pagination.rs
```diff
@@ -330,19 +297,6 @@ pub struct PaginationOutcome {
     pub total_items_updated: u64,
 }
 
-/// The status of a pagination.
-#[derive(Clone, Copy, Debug, PartialEq, Eq)]
-#[cfg_attr(feature = "uniffi", derive(uniffi::Enum))]
-pub enum PaginationStatus {
```
Hywan, line 336: 🎉

crates/matrix-sdk/src/event_cache/mod.rs
```diff
@@ -409,52 +415,65 @@ impl RoomEventCache {
...
 }
 
+#[derive(Debug)]
+struct RoomPaginationData {
+    /// The stateful paginator instance used for the integrated pagination.
+    paginator: Paginator,
+
+    /// Have we ever waited for a previous-batch-token from the sync? We do this
+    /// at most once per room, and we reset that upon clearing the timeline
+    /// events.
```
Hywan: If you can clarify to explain why and when this can happen, that would be lovely!

crates/matrix-sdk/src/event_cache/mod.rs
```diff
@@ -409,52 +415,65 @@ impl RoomEventCache {
...
     /// The events of the room.
     events: RwLock<RoomEvents>,
 
     /// A notifier that we received a new pagination token.
     pagination_token_notifier: Notify,
 
-    /// A lock that ensures we don't run multiple pagination queries at the same
-    /// time.
-    pagination_lock: Mutex<()>,
+    /// A paginator instance, that's configured to run back-pagination on our
```
Hywan, line 431: It doesn't handle front-pagination? Why?

crates/matrix-sdk/src/event_cache/mod.rs
```diff
@@ -464,14 +483,23 @@ impl RoomEventCacheInner {
...
-            pagination_lock: Default::default(),
+            pagination: RoomPaginationData {
+                paginator: Paginator::new(Box::new(room)),
+                waited_for_initial_prev_token: Mutex::new(false),
+            },
             pagination_token_notifier: Default::default(),
         }
     }
 
+    async fn clear(&self, room_events: &mut RwLockWriteGuard<'_, RoomEvents>) {
```
Hywan, line 456: Why passing the `room_events`? What's the benefits of this new approach?

crates/matrix-sdk/src/event_cache/mod.rs
```diff
@@ -658,34 +686,29 @@ impl RoomEventCacheInner {
...
+            self.pagination.paginator.paginate_backward(batch_size.into()).await?;
 
         // Make sure the `RoomEvents` isn't updated while we are saving events from
         // backpagination.
         let mut room_events = self.events.write().await;
 
-        // Check that the `token` exists if any.
-        let gap_identifier = if let Some(token) = token.as_ref() {
+        // Check that the previous token still exists; otherwise it's a sign that the
+        // timeline has been cleared..
```
Hywan:
```diff
- // timeline has been cleared..
+ // timeline has been cleared.
```

Also, it may not be appropriate to talk about “timeline” here, thoughts?

crates/matrix-sdk/src/event_cache/paginator.rs
```diff
@@ -66,6 +67,22 @@ pub enum PaginatorError {
     SdkError(#[source] crate::Error),
 }
 
+#[derive(Clone, Debug)]
+enum PaginationToken {
+    None,
+    HasMore(String),
+    HitEnd,
+}
```
Hywan, line 75: Can you document this type please?

crates/matrix-sdk/src/event_cache/pagination.rs
Hywan, line 1: Ahh great, I have written a comment asking for a dedicated module, neat!

crates/matrix-sdk/src/event_cache/pagination.rs
```diff
@@ -66,9 +66,9 @@ impl RoomPagination {
     /// that case, it's desirable to call the method again.
     // TODO(bnjbvr): that's bad API, this API should restart for us!
     #[instrument(skip(self))]
-    pub async fn paginate_backwards(&self, batch_size: u16) -> Result<BackPaginationOutcome> {
+    pub async fn run_backwards(&self, batch_size: u16) -> Result<BackPaginationOutcome> {
```
Hywan, line 68:
```diff
- run_backwards
+ backwards
```

Thoughts?

Hywan (commented):
Excellent improvements. It really improves the code, thank you!

I'm a bit skeptical about the last commit that removes the `custom` and `until_num_items` API for the pagination, see my comment.

crates/matrix-sdk-ui/src/timeline/pagination.rs
```diff
@@ -150,216 +117,3 @@ impl super::Timeline {
...
-    /// `ControlFlow::Break(())`).
-    pub fn custom(
-        initial_event_limit: u16,
-        pagination_strategy: impl Fn(PaginationOutcome) -> ControlFlow<(), u16> + Send + Sync + 'a,
-    ) -> Self {
-        Self::new(PaginationOptionsInner::Custom {
-            event_limit_if_first: Some(initial_event_limit),
-            strategy: Arc::new(pagination_strategy),
-        })
-    }
```
Hywan, line 231: My comment has been lost by Github, so here it is again.

I clearly understand why you want to drop this code. However, it addresses a useful and pragmatic feature of Element X and other app client. Being able to predictably paginate $n$ timeline event is a useful feature.

What if we keep the `Self::custom` API inside the `EventCache` and use it inside the `Timeline`?

bindings/matrix-sdk-ffi/src/timeline/mod.rs
bnjbvr: Indeed, that's a remnant of a previous impl. Thanks!

crates/matrix-sdk-ui/src/timeline/pagination.rs
bnjbvr: It's an implementation detail of `eyeball` that we're running into. If you have an emitter E and an observer O. Say E emits very quickly e1, e2, e3; then if O doesn't react fast enough, it may only see e1, handle it, and then miss e2, and only see e3.

Before this PR, the exposed pagination status was controlled at the timeline level, and the timeline would make sure to reset it only when a full pagination loop is over, after we obtained e.g. N events.

After this PR, the pagination status is that of the underlying paginator. The timeline may continue to paginate after the paginator has run one pagination query. In that case, the underlying status of the paginator will be:

- running (because it's busy doing the network request)
- idle (because it's done processing the response)
- (the timeline sees it hasn't received as many events as it wanted, so it retriggers the paginator) the paginator quickly re-runs into the running state

So the quick `running -> idle -> running` transition might make it so that an observer could see `running -> running`, because they didn't react fast enough to see the `idle` state in the middle.

I've added a `dedup()` call to address this, thanks.

Now, if the caller *is* actually fast enough, they may see a few quick transitions from `running -> idle -> running`; I wonder if that may cause some UI flickering (e.g. the spinner quickly disappears and reappears). At this point, both EX apps doesn't use the back-pagination status anymore, so it's a theoretical issue.

crates/matrix-sdk/src/event_cache/mod.rs
bnjbvr: Expanded the comment, but TLDR; this is only for back-paginations that interact with the event cache. Forward paginations don't, at the moment (i.e. events obtained in a response of forward pagination will NOT be saved into the cache — such paginations are "out of band")

crates/matrix-sdk/src/event_cache/mod.rs
bnjbvr: Expanded the comment a bit.

crates/matrix-sdk/src/event_cache/mod.rs
bnjbvr: Initially, I took the `room_events` lock within this function; but when this is called in `replace_all_events_by`, we're already holding onto the lock, and we *really* want to keep it throughout the full function's body, to avoid a bad race when two callers racily call `replace_all_events_by`.

crates/matrix-sdk/src/event_cache/mod.rs
bnjbvr: (thanks re: typo!)

> Also, it may not be appropriate to talk about “timeline” here, thoughts?

But it's the most precise word; what would you use?

crates/matrix-sdk/src/event_cache/pagination.rs
bnjbvr: I like my effectful functions to start with verbs :eyes: Is that ok to keep it like that?

crates/matrix-sdk-ui/src/timeline/pagination.rs
bnjbvr: So for what it's worth: we chatted about this offline, and it turns out the EX apps (our main "users") aren't using those capabilities, plus these capabilities were annoying to keep working. I've taken the liberty to remove those, but we could reintroduce them later, if there's a strong use case for them.

Hywan (approved):
Nothing to add, good job, thanks for addressing the feedback!

crates/matrix-sdk/src/event_cache/pagination.rs
Hywan: I can live with that 😉

PR #3375, opened 2024-05-02
chore(cargo) Update dependencies

Hywan:
As the title says!

This PR updates the dependencies. It also removes `cfg_vis` from `matrix-rust-sdk` because that's pretty useless if I understand what it was supposed to do. It was also introducing more proc-macros and a duplicated dep with Ruma.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -11,7 +11,7 @@ dependencies = [
  "macroific",
  "proc-macro2",
  "quote",
- "syn 2.0.60",
+ "syn 2.0.63",
 ]
 
 [[package]]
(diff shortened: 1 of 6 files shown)


poljar (approved):
Looks fine, but what was `cfg_vis` supposed to do? Hide the method on WASM, why?

Hywan:
I don't know. The default is `pub` so we cannot have any warning or whatever.

poljar:
This should now be fine to merge since #3380 landed?

zecakeh:
You'll need to update the version of nightly used in CI if you want to upgrade axum.

PR #3377, opened 2024-05-02
base: sort the heroes in the computed display name alphabetically

bnjbvr:
This ensures the room name is deterministic, when it's computed client-side. Includes regression tests too, where the members or heroes might not be in the expected order.

### DIFF
--- a/crates/matrix-sdk-base/src/rooms/normal.rs
+++ b/crates/matrix-sdk-base/src/rooms/normal.rs
@@ -148,6 +148,14 @@ impl From<&MembershipState> for RoomState {
     }
 }
 
+/// The number of heroes chosen to compute a room's name, if the room didn't
+/// have a name set by the users themselves.
+///
+/// A server must return at most 5 heroes, according to the paragraph below
+/// https://spec.matrix.org/v1.10/client-server-api/#get_matrixclientv3sync (grep for "heroes"). We
+/// try to behave similarly here.
+const NUM_HEROES: usize = 5;
+
 impl Room {
     /// The size of the latest_encrypted_events RingBuffer
     // SAFETY: `new_unchecked` is safe because 10 is not zero.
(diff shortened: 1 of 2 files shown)


andybalaam (approved):
Makes sense, thanks!

crates/matrix-sdk-base/src/rooms/normal.rs
```diff
@@ -596,24 +596,39 @@ impl Room {
...
+            // Make the ordering deterministic.
+            members.sort_unstable_by(|lhs, rhs| lhs.name().cmp(rhs.name()));
+
+            // We can make a good prediction of the total number of members here. This might
+            // be incorrect if the database info is outdated.
+            let guessed_num_members = Some(members.len());
+
+            // The server returns at most 5 heroes; behave similarly here.
+            (
+                members.into_iter().take(5).filter(|u| u.user_id() != own_user_id).collect(),
```
Hywan, line 622: I would love to see a constant instead of having `5` here, thoughts?

crates/matrix-sdk-base/src/rooms/normal.rs
```diff
@@ -596,24 +596,39 @@ impl Room {
...
+            let mut members = self.members(RoomMemberships::ACTIVE).await?;
+
+            // Make the ordering deterministic.
+            members.sort_unstable_by(|lhs, rhs| lhs.name().cmp(rhs.name()));
+
+            // We can make a good prediction of the total number of members here. This might
+            // be incorrect if the database info is outdated.
+            let guessed_num_members = Some(members.len());
+
+            // The server returns at most 5 heroes; behave similarly here.
```
Hywan, line 620: Do we have a reference to it, a proof, something 🙂?

crates/matrix-sdk-base/src/rooms/normal.rs
```diff
@@ -596,24 +596,39 @@ impl Room {
...
+            (
+                members.into_iter().take(5).filter(|u| u.user_id() != own_user_id).collect(),
+                guessed_num_members,
+            )
         } else {
-            let members: Vec<_> = stream::iter(summary.heroes.iter())
+            let mut heroes = summary.heroes;
+
+            // Make the ordering deterministic.
+            heroes.sort();
```
Hywan, line 629: We are using `sort_unstable` above, and here we are using `sort`, is that fine?

crates/matrix-sdk-base/src/rooms/normal.rs
```diff
@@ -1919,6 +1939,110 @@ mod tests {
...
+    #[async_test]
+    async fn test_display_name_deterministic() {
+        let (store, room) = make_room(RoomState::Joined);
+
+        let alice = user_id!("@alice:example.org");
+        let bob = user_id!("@bob:example.org");
+        let carol = user_id!("@carol:example.org");
+        let denis = user_id!("@denis:example.org");
+        let erica = user_id!("@erica:example.org");
+        let fred = user_id!("@fred:example.org");
```
Hywan, line 1951: Nice game 😄.

Hywan (changes requested):
(wrongly send the review before)

crates/matrix-sdk-base/src/rooms/normal.rs
Hywan: Are you comfortable by quoting (a part of) the spec with a link to it? We already do that in the code when it needs specific clarifications.

poljar: Sure, but the *should* isn't that confidence inspiring.

Hywan: Indeed… at least we can consider it's a safe number.

crates/matrix-sdk-base/src/rooms/normal.rs
bnjbvr: Yeah, this code path could be used without the other code path being called, so we need to sort; at most there are 5 items in that list, so any sorting works. Or are you requesting to use `sort_unstable` instead of `sort` for consistency? (Wouldn't make a big difference, but happy to make the change)

Hywan: The latter. I think it's better for the sake of consistency.

Hywan (approved):
Approving ahead of time to avoid back and forth during the review process.

PR #3092, opened 2024-02-05
feat(sdk): Remove `SlidingSyncRoom::timeline_queue` and `prev_batch`

Hywan:
These patch removes `SlidingSyncRoom::timeline_queue`, `SlidingSyncRoom::prev_batch`, `FrozenSlidingSyncRoom::timeline` and `FrozenSlidingSyncRoom::prev_batch`.

The removal of `timeline_queue` has escalated to updating `read_receipts::compute_unread_counts` to be called with events from the Event Cache instead of `timeline_queue`. The code has been revisited a little bit around the sliding sync workflow, but nothing fancy or deep.

---

* Address https://github.com/matrix-org/matrix-rust-sdk/issues/3079

Issue #3058, opened 2024-01-26
[meta] Event Cache API 🕸


bnjbvr:
## Context

After working for a bit on the Rust SDK, I've observed it seems there's a missing layer between the high-level Timeline in the UI crate, and the Rust SDK itself. In addition to rendering the events themselves, the Timeline is involved in a lot of functionality:

- it computes the read receipts displayed on each event in a room,
- it runs backpagination, keeping track of the `prev_batch` tokens, re-spawning requests to `/messages` until it reaches a desired amount of messages, or the end of a timeline,
- it retries decrypting events after new keys have been received, be it from the encryption sync loop, or from the backup storage

These extra responsibilities make the Timeline more a view-and-controller bundle, as opposed to a pure view over the events it receives. This tight coupling comes as an advantage (the code freely does what it needs when it needs to), but also it comes with a few drawbacks: it's harder to test and fix a logic bug in isolation without _rendering_ a timeline, and all that functionality is not reusable by other clients who wouldn't want to render as a timeline.

Since rendering a timeline requires subscribing to the room it belongs to, and we have requirements for other features that don't involve a specific room, but the entire room list, some of that functionality is re-implemented elsewhere:

- renewed attempts to decrypt still-encrypted events happen when computing the `latest_event` for a room, so as we have something to display in the room list,
- the room unread markers (namely, the unread counts, # of new messages/notifications/mentions) are also needed for all the rooms, not only a room we'd be visiting (or the whole point of the unread marker would be lost)
- in addition to that, while the latest events and unread counts are technically information *derived* (computed) from other events, it's serialized and stored into the database as part of the `RoomInfo` struct, into the state store.
- the scattering of unread counts and read receipts has caused some impedance mismatch, exemplified as https://github.com/matrix-org/matrix-rust-sdk/pull/3054 (the unread counts computation expected read markers even for our own events, while the timeline code handling read receipts took into account implicit read receipts)

## The Proposal

Overall, my take is that all this kind of functionality is really nice to have, when writing *any* Matrix client. While it's possible that we could continue like we've done so far, I think that having a new abstraction to provide all this functionality, independently of a given observed room, would be super nice and useful. I call this new abstraction the Event Graph (API, if you insist).

The goals are the following:

- make the Timeline (mostly) a view over the events observed via the Event Graph, for a clearer and cleaner separation of responsibilities
- make the code more modular and easier to test in isolation, to improve maintainability
[...]

PR #3378, opened 2024-05-02
event cache: handle blocking correctly

bnjbvr:
This should "fix" ignoring/unignoring users, which had regressed likely since the introduction of the event cache.

This is still not perfect: in theory the sliding sync room cache should *also* be emptied every time we get an ignored/unignored change, but better to wait for https://github.com/matrix-org/matrix-rust-sdk/pull/3092 to land, instead of trying to work around it somewhere else.

Part of #3058, should help with https://github.com/element-hq/element-meta/issues/2382 too.

### DIFF
--- a/crates/matrix-sdk/tests/integration/event_cache.rs
+++ b/crates/matrix-sdk/tests/integration/event_cache.rs
@@ -5,7 +5,9 @@ use matrix_sdk::{
     event_cache::{BackPaginationOutcome, EventCacheError, RoomEventCacheUpdate},
     test_utils::{assert_event_matches_msg, events::EventFactory, logged_in_client_with_server},
 };
-use matrix_sdk_test::{async_test, EventBuilder, JoinedRoomBuilder, SyncResponseBuilder};
+use matrix_sdk_test::{
+    async_test, EventBuilder, GlobalAccountDataTestEvent, JoinedRoomBuilder, SyncResponseBuilder,
+};
 use ruma::{
     event_id,
     events::{room::message::RoomMessageEventContent, AnyTimelineEvent},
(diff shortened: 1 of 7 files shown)


crates/matrix-sdk/tests/integration/event_cache.rs
```diff
@@ -136,6 +138,117 @@ async fn test_add_initial_events() {
...
+    sync_builder
+        .add_joined_room(JoinedRoomBuilder::new(room_id))
+        .add_joined_room(JoinedRoomBuilder::new(other_room_id));
+
+    mock_sync(&server, sync_builder.build_json_sync_response(), None).await;
+    client.sync_once(Default::default()).await.unwrap();
+    server.reset().await;
+
+    let dexter = user_id!("@dexter:lab.org");
+    let ivan = user_id!("@ivan:lab.ch");
```
Hywan, line 162: 😉

crates/matrix-sdk/tests/integration/event_cache.rs
```diff
@@ -136,6 +138,117 @@ async fn test_add_initial_events() {
...
+        .expect("timeout after receiving a sync update")
+        .expect("should've received a room event cache update");
+
+    // Which notifies about the clear.
+    assert_matches!(update, RoomEventCacheUpdate::Clear);
+
+    // Receiving new events still works.
+    sync_builder.add_joined_room(
+        JoinedRoomBuilder::new(room_id)
+            .add_timeline_event(ev_factory.text_msg("i don't like this dexter").sender(ivan)),
```
Hywan, line 224: Bim, takes that Dexter!

Hywan (approved):
I like how you insert jokes with my name in the tests so that you're checking whether I've reviewed them or not, that's smart 😏.

crates/matrix-sdk/tests/integration/event_cache.rs
bnjbvr: could be any Ivan in Switzerland :see_no_evil:

Hywan: Suuuuuuure

PR #3380, opened 2024-05-02
chore(ffi): Remove support for opentelemetry

Hywan:
This patch removes support for OpenTelemetry in `matrix-sdk-ffi` because it's not used anymore by anybody. It also adds multiple duplicated dependencies (like `reqwest`). Anyway. Farewell.

Removing opentelemetry removes axum 0.6.20, hyper-timeout, reqwest 0.11.20, tokio-timeout, tonic and winreg.

### DIFF
--- a/bindings/matrix-sdk-ffi/src/platform.rs
+++ b/bindings/matrix-sdk-ffi/src/platform.rs
@@ -1,11 +1,3 @@
-use std::{collections::HashMap, fmt::Debug, pin::Pin};
-
-use base64::{engine::general_purpose::STANDARD, Engine};
-use futures_core::future::BoxFuture;
-use opentelemetry::KeyValue;
-use opentelemetry_otlp::{Protocol, WithExportConfig};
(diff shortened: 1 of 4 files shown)


bindings/matrix-sdk-ffi/src/platform.rs
```diff
@@ -291,18 +211,12 @@ pub fn setup_otlp_tracing(config: OtlpTracingConfiguration) {
     #[cfg(target_os = "android")]
     log_panics();
 
-    let otlp_tracer =
```
poljar, line 294: What is the point of this method if it doesn't do anymore what it's advertised to do? The `OtlpTracingConfiguration` also seems to be pointless?

poljar (changes requested):
I think there's more to remove.

bindings/matrix-sdk-ffi/src/platform.rs
Hywan: Indeed, you're right!

Hywan:
More code removed cap'tain 🫡!

poljar (approved):

PR #3382, opened 2024-05-03
Use a released version of uniffi

poljar:


andybalaam (approved):

Hywan (approved):

poljar:
@jmartinesp, @Velin92 care to test this using EX? It shouldn't break anything as far as I could tell, but perhaps it's wise to double check.

jmartinesp:
I'm getting:

```
java.lang.RuntimeException: Unable to create application io.element.android.x.ElementXApplication: androidx.startup.StartupException: java.lang.RuntimeException: UniFFI API checksum mismatch: try cleaning and rebuilding your project
```

I'm trying to solve it, I'll update the status once it's fixed.

Hywan:
You may want to run `cargo clean`.

jmartinesp:
> You may want to run `cargo clean`.

I've cleaned cargo and gradle several times in both the [Rust Kotlin components](https://github.com/matrix-org/matrix-rust-components-kotlin) project and EXA, but it doesn't seem to help.

jmartinesp:
Ok, so UniFFi is now generating 2 `matrix_sdk_ffi.kt` files, one with the `org.matrix.rustcomponents.sdk` package and another one with the `uniffi.matrix_sdk_ffi` one. The contents are quite similar except for the API checksums and some function definitions, so I guess it *is* using the cache somehow to generate one of them (I'm guessing the `org.matrix.rustcomponents.sdk` one).

jmartinesp:
Found and fixed the issue, but it forces us to change *all* imports from `org.matrix.rustcomponents.*` to `uniffi.matrix_sdk_ffi.*`, which is not ideal... Is there some way to keep the package name here?

Velin92:
Seems to build and run fine on iOS without any issue

poljar:
> Found and fixed the issue, but it forces us to change _all_ imports from `org.matrix.rustcomponents.*` to `uniffi.matrix_sdk_ffi.*`, which is not ideal... Is there some way to keep the package name here?

There is https://mozilla.github.io/uniffi-rs/kotlin/configuration.html, but we should be using that already: https://github.com/matrix-org/matrix-rust-sdk/blob/60017241b28da8feedf79f8f2908b68507c58ceb/bindings/matrix-sdk-ffi/uniffi.toml#L1-L4

poljar:
> Ok, so UniFFi is now generating 2 `matrix_sdk_ffi.kt` files, one with the `org.matrix.rustcomponents.sdk` package and another one with the `uniffi.matrix_sdk_ffi` one. The contents are quite similar except for the API checksums and some function definitions, so I guess it _is_ using the cache somehow to generate one of them (I'm guessing the `org.matrix.rustcomponents.sdk` one).

I tried to reproduce this, and indeed this is happening:

image

But I reproduced the same behavior on main, to be exact on commit 60017241b28da8feedf79f8f2908b68507c58ceb. So is main as well broken or is this continuing to work as expected?

poljar:
Alright, after a lot of back and forth and countless compilation cycles me and @jmartinesp found out that this patch https://github.com/matrix-org/matrix-rust-sdk/pull/3382/commits/49f5cf236b31f46af0b303893da5c689c322da06 restores the old behavior and ensures that the `package_name` in the `uniffi.toml` file is respected again.

No mass rename for the imports is needed.

Hywan:
Well done!

PR #3387, opened 2024-05-08
Parsing of QR codes from MSC4108

poljar:
This is the first part of [MSC4108] support, it just adds the basic parsing and encoding parts for the data which will be present in QR codes for [MSC4108].

[MSC4108]: https://github.com/matrix-org/matrix-spec-proposals/pull/4108


Hywan (approved):
LGTM

PR #3389, opened 2024-05-08
chore(crypto): Refactor the cross-signing key wrappers

poljar:
Since the master/self-signing/user-signing public key types are used for public user identities as well as for the private key type we have, and we'd like to sign the public key types it makes sense that the types itself aren't using an Arc.

Let's instead put the Arc inside the user identity structs.

This will allow us later on to more easily sign the public key types.


andybalaam (approved):
Yep, thanks!

Issue #3280, opened 2024-03-27
[meta] `EventCache` storage


Hywan:
![output](https://github.com/matrix-org/matrix-rust-sdk/assets/946104/d98fedf7-5c3b-4e5c-b21b-0a2278cbb09b)

---

`matrix_sdk::EventCache` is a new module that has been introduced with #3058. The idea is to manage all events inside the Rust SDK in a single place. An important user of `EventCache` is the `matrix_sdk_ui::Timeline`.

`EventCache` uses a new data structure to organize all events, `LinkedChunk`, since:

* #3166
* #3230

This issue describes the plan to get a persistent storage for `EventCache`, along with the ability to make it reactive.

# Persistent storage

In order to create a persistent storage for `EventCache`, we need a mechanism that listens to `LinkedChunk` updates, and map these updates to database operations (like `INSERT`, `DELETE` and so on).

We initially went to using a reactive mechanism (like `ObservableLinkedChunk`), but like any reactive mechanisms, we need to handle the _lag_.

> [!NOTE]
> What is a lag? When new updates are generated, they are usually accumulated in a buffer. This buffer is drained by subscribers of the observed value. When the buffer is full, which can happen because subscribers _lag_ to consume updates, the buffer is reset.

The lag, in this case, is pretty problematic. In case of a lag, the database will be out of sync —data will be missing—, and there is no easy way to get them again. In case of a lag, we might imagine to reset the database and to rewrite everything again, but not all events are loaded in memory. Alternatively, we may want to recompute all the differences between what is in memory and what is inside the database, but again, it's not an easy problem. Anyway, it implies more guards and more complexity.

Instead, we've decided to define a `LinkedChunkListener` trait, used by `LinkedChunk`, onto which methods will be called on some particular operations, like `insert_chunk`, `remove_chunk`, `insert_events` and `remove_events`, that's probably all we need.

The cons:

* It's a new API to manage,
* It cannot be re-used by something else. It exists for one usage only.

The pros:

* It's _immediate_: it removes the problem of the lag introduced by the initial reactive approach,
* It's easily unit-testable!
* The API is quite small and clearly not complex at all, it's straighforward.

## Tasks

* [x] Create `LinkedChunkListener`: https://github.com/matrix-org/matrix-rust-sdk/pull/3281
* [x] Create a persistent storage for `EventCache`; write mode: from memory to persistent storage:
  * [x] https://github.com/matrix-org/matrix-rust-sdk/pull/4192
[...]

PR #3392, opened 2024-05-08
feat(sdk): Add `LinkedChunk::subscribe_as_vector`

Hywan:
This patch implements `LinkedChunk::<Item, Gap>::as_vector`,
which returns a `AsVector`.

`AsVector` is a type that transforms a
`Vec<Update<Item, Gap>>` into a `Vec<VectorDiff<Item>>` —this type—. Basically, it
helps to consume a `LinkedChunk<CAP, Item, Gap>` as if it was an
`eyeball::ObservableVector<Item>`.

How this type transforms `Update` into `VectorDiff`? There is no internal
buffer of kind `eyeball_im::ObservableVector<Item>`, which could have been
used to generate the `VectorDiff`s. They are computed manually.

The only buffered data is pairs of `ChunkIdentifier` and `ChunkLength`.
The following rules must be respected:

* A chunk of kind `ChunkContent::Gap` has a length of 0,
* A chunk of kind `ChunkContent::Items` has a length equals to its number
  of items,
* The pairs must be ordered exactly like the chunks in `LinkedChunk`, i.e.
  the first pair must represent the first chunk, the last pair must
  represent the last chunk.

The only thing this algorithm does is maintaining the pairs:

* `Update::NewItemsChunk` and `Update::NewGapChunk` are inserting a new
  pair with a chunk length of 0 at the appropriate index,
* `Update::RemoveChunk` is removing a pair,
* `Update::PushItems` is increasing the length of the appropriate pair by
  the number of new items, and is potentially emitting `VectorDiff`,
* `Update::DetachLastItems` is decreasing the length of the appropriate pair
  by the number of items to be detached; no `VectorDiff` is emitted,
* `Update::StartReattachItems` and `Update::EndReattachItems` are
  respectively muting or unmuting the emission of `VectorDiff` by
  `Update::PushItems`.

The only `VectorDiff` that are emitted are `VectorDiff::Insert` or
`VectorDiff::Append` because a `LinkedChunk` is append-only.

`VectorDiff::Append` is an optimisation when numerous `VectorDiff::Insert`s
have to be emitted at the last position.

`VectorDiff::Insert` need an index. To compute this index, the algorithm
will iterate over all pairs to accumulate each chunk length until it finds
the appropriate pair (given by `Update::PushItems::position_hint`). This
is _the offset_. To this offset, the algorithm adds the position's index of
the new items (still given by `Update::PushItems::position_hint`). This is
_the index_. This logic works for all cases as long as pairs are maintained
according to the rules hereinabove.

That's a pretty memory compact and computation efficient way to map a
`Vec<Update<Item, Gap>>` into a `Vec<VectorDiff<Item>>`. The larger
the `LinkedChunk` capacity is, the
fewer pairs the algorithm will have to handle, e.g. for 1'000 items and a
`LinkedChunk` capacity of 128, it's only 8 pairs, be 256 bytes.

Most of the patch is documentation and tests, don't be impressed 🙂.

---

### DIFF
--- a/crates/matrix-sdk/src/event_cache/linked_chunk/as_vector.rs
+++ b/crates/matrix-sdk/src/event_cache/linked_chunk/as_vector.rs
@@ -62,9 +62,8 @@ impl<Item, Gap> AsVector<Item, Gap> {
         Item: Clone,
     {
         let mut updates = self.updates.write().unwrap();
-        let updates = updates.take_with_token(self.token);
 
-        self.mapper.map(updates)
+        self.mapper.map(updates.take_with_token(self.token))
     }
 }
 
--- a/crates/matrix-sdk/src/event_cache/linked_chunk/as_vector.rs
+++ b/crates/matrix-sdk/src/event_cache/linked_chunk/as_vector.rs
@@ -117,7 +117,7 @@ impl UpdateToVectorDiff {
     /// * [`Update::DetachLastItems`] is decreasing the length of the
     ///   appropriate pair by the number of items to be detached; no
     ///   [`VectorDiff`] is emitted,
-    /// * [`Update::ReattachItems`] and [`Update::ReattachItemsDone`] are
+    /// * [`Update::StartReattachItems`] and [`Update::EndReattachItems`] are
     ///   respectively muting or unmuting the emission of [`VectorDiff`] by
     ///   [`Update::PushItems`].
     ///
@@ -203,7 +203,7 @@ impl UpdateToVectorDiff {
         // Step 3, reattaching detached items:
         //
         // ```
-        // Update::ReattachItems
+        // Update::StartReattachItems
         // Update::PushItems {
         //     position_hint: Position(ChunkIdentifier(2), 2),
         //     items: vec!['b']
@@ -217,7 +217,7 @@ impl UpdateToVectorDiff {
         //     position_hint: Position(ChunkIdentifier(3), 0),
         //     items: vec!['c'],
         // }
-        // Update::ReattachItemsDone
+        // Update::EndReattachItems
         // ```
         //
         // To ensure an optimised behaviour of this algorithm:
@@ -225,8 +225,8 @@ impl UpdateToVectorDiff {
         // * `Update::DetachLastItems` must not emit `VectorDiff::Remove`,
         //
         // * `Update::PushItems` must not emit `VectorDiff::Insert`s or
-        //   `VectorDiff::Append`s if it happens after `ReattachItems` and before
-        //   `ReattachItemsDone`. However, `Self::chunks` must always be updated.
+        //   `VectorDiff::Append`s if it happens after `StartReattachItems` and before
+        //   `EndReattachItems`. However, `Self::chunks` must always be updated.
         //
         // From the `VectorDiff` “point of view”, this optimisation aims at avoiding
         // removing items to push them again later.
@@ -369,12 +369,12 @@ impl UpdateToVectorDiff {
                     *length = new_length;
                 }
 
-                Update::ReattachItems => {
+                Update::StartReattachItems => {
                     // Entering the `reattaching` mode.
                     mute_push_items = true;
                 }
 
-                Update::ReattachItemsDone => {
+                Update::EndReattachItems => {
                     // Exiting the `reattaching` mode.
                     mute_push_items = false;
                 }
--- a/crates/matrix-sdk/src/event_cache/linked_chunk/as_vector.rs
+++ b/crates/matrix-sdk/src/event_cache/linked_chunk/as_vector.rs
@@ -148,7 +148,10 @@ where
                     }
                 }
 
-                Update::TruncateItems { chunk: expected_chunk_identifier, length: new_length } => {
+                Update::DetachLastItems { at } => {
+                    let expected_chunk_identifier = at.chunk_identifier();
+                    let new_length = at.index();
+
                     let length = this
                         .chunks
                         .iter_mut()
--- a/crates/matrix-sdk/src/event_cache/linked_chunk/as_vector.rs
+++ b/crates/matrix-sdk/src/event_cache/linked_chunk/as_vector.rs
@@ -354,7 +354,7 @@ impl UpdateToVectorDiff {
                     }
                     // No optimisation: let's emit `VectorDiff::Insert`.
                     else {
-                        diffs.extend(items.into_iter().enumerate().map(|(nth, item)| {
+                        diffs.extend(items.iter().enumerate().map(|(nth, item)| {
                             VectorDiff::Insert { index: offset + nth, value: item.clone() }
                         }));
                     }
--- a/crates/matrix-sdk/src/event_cache/linked_chunk/as_vector.rs
+++ b/crates/matrix-sdk/src/event_cache/linked_chunk/as_vector.rs
@@ -1,0 +1,239 @@
+// Copyright 2024 The Matrix.org Foundation C.I.C.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use std::{
+    collections::VecDeque,
+    ops::ControlFlow,
+    pin::Pin,
+    task::{ready, Context, Poll},
+};
+
+use eyeball_im::VectorDiff;
+use futures_core::Stream;
+use pin_project_lite::pin_project;
+
+use super::{
+    updates::{Update, UpdatesSubscriber},
+    ChunkIdentifier, ChunkIdentifierGenerator,
+};
+
+type Offset = usize;
+type ChunkLength = usize;
+
+pin_project! {
+    pub struct AsVectorSubscriber<Item, Gap> {
+        #[pin]
+        updates_subscriber: UpdatesSubscriber<Item, Gap>,
+
+        chunks: VecDeque<(ChunkIdentifier, ChunkLength)>,
+    }
+}
+
+impl<Item, Gap> AsVectorSubscriber<Item, Gap> {
+    pub(super) fn new(updates_subscriber: UpdatesSubscriber<Item, Gap>) -> Self {
+        Self {
+            updates_subscriber,
+            chunks: {
+                let mut chunks = VecDeque::new();
+                chunks.insert(0, (ChunkIdentifierGenerator::FIRST_IDENTIFIER, 0));
+
+                chunks
+            },
+        }
+    }
+}
+
+impl<Item, Gap> Stream for AsVectorSubscriber<Item, Gap>
+where
+    Item: Clone + std::fmt::Debug,
+    Gap: Clone + std::fmt::Debug,
+{
+    type Item = Vec<VectorDiff<Item>>;
+
+    fn poll_next(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
+        let mut this = self.project();
+
+        let Some(updates) = ready!(this.updates_subscriber.as_mut().poll_next(cx)) else {
+            return Poll::Ready(None);
+        };
+
+        let mut diffs = Vec::with_capacity(updates.len());
+
+        for update in updates {
+            match update {
+                Update::NewItemsChunk { previous, new, next }
+                | Update::NewGapChunk { previous, new, next, .. } => {
+                    match (previous, next) {
+                        // New chunk at the end.
+                        (Some(_previous), None) => {
+                            // TODO: chec `previous` is correct
+                            this.chunks.push_back((new, 0));
+                        }
+
+                        // New chunk at the beginning.
+                        (None, Some(_next)) => {
+                            // TODO: check `next` is correct
+                            this.chunks.push_front((new, 0));
+                        }
+
+                        // New chunk is inserted between 2 chunks.
+                        (Some(_previous), Some(next)) => {
+                            // TODO: check `previous` is correct
+                            let next_chunk_index = this
+                                .chunks
+                                .iter()
+                                .position(|(chunk_identifier, _)| *chunk_identifier == next)
+                                .expect("next chunk not found");
+
+                            this.chunks.insert(next_chunk_index, (new, 0));
+                        }
+
+                        (None, None) => {
+                            unreachable!("?");
+                        }
+                    }
+                }
+
+                Update::RemoveChunk(expected_chunk_identifier) => {
+                    let chunk_index = this
+                        .chunks
+                        .iter()
+                        .position(|(chunk_identifier, _)| {
+                            *chunk_identifier == expected_chunk_identifier
+                        })
+                        .expect("oops 4");
+
+                    this.chunks.remove(chunk_index).unwrap();
+                }
+
+                Update::InsertItems { at: position, items } => {
+                    let expected_chunk_identifier = position.chunk_identifier();
+
+                    let (chunk_index, offset, chunk_length) = {
+                        let control_flow = this.chunks.iter_mut().enumerate().try_fold(
+                            position.index(),
+                            |offset, (chunk_index, (chunk_identifier, chunk_length))| {
+                                if chunk_identifier == &expected_chunk_identifier {
+                                    ControlFlow::Break((chunk_index, offset, chunk_length))
+                                } else {
+                                    ControlFlow::Continue(offset + *chunk_length)
+                                }
+                            },
+                        );
+
+                        match control_flow {
+                            ControlFlow::Break(value) => Some(value),
+                            ControlFlow::Continue(..) => None,
+                        }
+                    }
+                    .expect("`ChunkIdentifier` must exist");
+
+                    *chunk_length += items.len();
+
+                    // Optimisation: we can emit a `VectorDiff::Append` in this particular case.
+                    if chunk_index == this.chunks.len() - 1 {
+                        diffs.push(VectorDiff::Append { values: items.into() });
+                    }
+                    // No optimisation: let's emit `VectorDiff::Insert`.
+                    else {
+                        diffs.extend(items.into_iter().enumerate().map(|(nth, item)| {
+                            VectorDiff::Insert { index: offset + nth, value: item }
+                        }));
+                    }
+                }
+
+                Update::TruncateItems { chunk: expected_chunk_identifier, length: new_length } => {
+                    let length = this
+                        .chunks
+                        .iter_mut()
+                        .find_map(|(chunk_identifier, length)| {
+                            (*chunk_identifier == expected_chunk_identifier).then(|| length)
+                        })
+                        .expect("oops 3");
+
+                    let old_length = *length;
+                    *length = new_length;
+
+                    diffs.extend(
+                        (new_length..old_length)
+                            .into_iter()
+                            .map(|index| VectorDiff::Remove { index }),
+                    );
+                }
+            }
+        }
+
+        Poll::Ready(Some(diffs))
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use futures_util::pin_mut;
+    use imbl::vector;
+    use stream_assert::{assert_closed, assert_next_eq, assert_pending};
+
+    use super::{super::LinkedChunk, VectorDiff};
+
+    #[test]
+    fn test_as_vector() {
+        let mut linked_chunk = LinkedChunk::<3, char, ()>::new_with_update_history();
+        let as_vector = linked_chunk.subscribe_as_vector().unwrap();
+        pin_mut!(as_vector);
+
+        assert_pending!(as_vector);
+
+        linked_chunk.push_items_back(['a', 'b']);
+        linked_chunk.push_items_back(['c', 'd', 'e']);
+
+        assert_next_eq!(
+            as_vector,
+            &[
+                VectorDiff::Append { values: vector!['a', 'b'] },
+                VectorDiff::Append { values: vector!['c'] },
+                VectorDiff::Append { values: vector!['d', 'e'] },
+            ]
+        );
+
+        linked_chunk
+            .insert_items_at(['f', 'g'], linked_chunk.item_position(|item| *item == 'b').unwrap())
+            .unwrap();
+
+        assert_next_eq!(
+            as_vector,
+            &[
+                VectorDiff::Remove { index: 1 },
+                VectorDiff::Remove { index: 2 },
+                VectorDiff::Insert { index: 1, value: 'f' },
+                VectorDiff::Insert { index: 2, value: 'g' },
+                VectorDiff::Insert { index: 3, value: 'b' },
+                VectorDiff::Insert { index: 4, value: 'c' },
+            ]
+        );
+
+        linked_chunk.push_gap_back(());
+        linked_chunk.push_items_back(['h', 'i']);
+
+        assert_next_eq!(as_vector, &[VectorDiff::Append { values: vector!['h', 'i'] }]);
+
+        linked_chunk
+            .replace_gap_at(['j'], linked_chunk.chunk_identifier(|chunk| chunk.is_gap()).unwrap())
+            .unwrap();
+
+        assert_next_eq!(as_vector, &[VectorDiff::Insert { index: 7, value: 'j' }]);
+
+        drop(linked_chunk);
+        assert_closed!(as_vector);
+    }
+}
--- a/crates/matrix-sdk/src/event_cache/linked_chunk/as_vector.rs
+++ b/crates/matrix-sdk/src/event_cache/linked_chunk/as_vector.rs
@@ -342,7 +342,7 @@ where
                         .chunks
                         .iter_mut()
                         .find_map(|(chunk_identifier, length)| {
-                            (*chunk_identifier == expected_chunk_identifier).then(|| length)
+                            (*chunk_identifier == expected_chunk_identifier).then_some(length)
                         })
                         // SAFETY: Assuming `LinkedChunk` and `Updates` are not buggy, and assuming
                         // `Self::chunks` is correctly initialized, it is not possible to detach
--- a/crates/matrix-sdk/src/event_cache/linked_chunk/as_vector.rs
+++ b/crates/matrix-sdk/src/event_cache/linked_chunk/as_vector.rs
@@ -28,20 +28,38 @@ use super::{
     ChunkIdentifier,
 };
 
-type Offset = usize;
+/// A type alias to represent a chunk's length. This is purely for commodity.
 type ChunkLength = usize;
 
 pin_project! {
+    /// A type used to transform a `Stream<Item = Vec<Update<Item, Gap>>>` into
+    /// a `Stream<Item = Vec<VectorDiff<Item>>>`. Basically, it helps to consume
+    // a [`LinkedChunk<CAP, Item, Gap>`] as if it was an [`ObservableVector<Item>`].
     pub struct AsVectorSubscriber<Item, Gap> {
+        // The inner `UpdatesSubscriber`.
         #[pin]
         updates_subscriber: UpdatesSubscriber<Item, Gap>,
 
+        // Pairs of all known chunks and their respective length. This is the only
+        // required data for this algorithm.
         chunks: VecDeque<(ChunkIdentifier, ChunkLength)>,
     }
 }
 
 impl<Item, Gap> AsVectorSubscriber<Item, Gap> {
-    pub(super) fn new(
+    /// Create a new [`Self`].
+    ///
+    /// `updates_subscriber` is simply `UpdatesSubscriber`.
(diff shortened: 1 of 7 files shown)


crates/matrix-sdk/src/event_cache/linked_chunk/as_vector.rs
```diff
@@ -0,0 +1,529 @@
...
+        let mut updates = self.updates.write().unwrap();
+        let updates = updates.take_with_token(self.token);
+
+        self.mapper.map(updates)
+    }
+}
+
+/// Internal type that converts [`Update`] into [`VectorDiff`].
+///
+/// This type maintains its own state.
```
bnjbvr, line 87: This should be true of any data structure?

```suggestion
```

crates/matrix-sdk/src/event_cache/linked_chunk/as_vector.rs
```diff
@@ -0,0 +1,529 @@
...
+struct UpdateToVectorDiff {
+    /// Pairs of all known chunks and their respective length. This is the only
+    /// required data for this algorithm.
+    chunks: VecDeque<(ChunkIdentifier, ChunkLength)>,
+}
+
+impl UpdateToVectorDiff {
+    /// Map several [`Update`] into [`VectorDiff`].
+    ///
+    /// How this type transforms `Update` into `VectorDiff`? There is no
```
bnjbvr, line 97:
```suggestion
    /// How does this type transform `Update` into `VectorDiff`? There is no
```

crates/matrix-sdk/src/event_cache/linked_chunk/as_vector.rs
```diff
@@ -0,0 +1,529 @@
...
+    ///   [`VectorDiff`],
+    /// * [`Update::DetachLastItems`] is decreasing the length of the
+    ///   appropriate pair by the number of items to be detached; no
+    ///   [`VectorDiff`] is emitted,
+    /// * [`Update::ReattachItems`] and [`Update::ReattachItemsDone`] are
+    ///   respectively muting or unmuting the emission of [`VectorDiff`] by
+    ///   [`Update::PushItems`].
+    ///
+    /// The only `VectorDiff` that are emitted are [`VectorDiff::Insert`] or
+    /// [`VectorDiff::Append`] because a [`LinkedChunk`] is append-only.
```
bnjbvr, line 128: We should also be able to `Clear` a full linked chunk, in case we're seeing a limited sync (until we have a fancier diff/reconciliation algorithm)?

crates/matrix-sdk/src/event_cache/linked_chunk/as_vector.rs
```diff
@@ -0,0 +1,529 @@
...
+    /// finds the appropriate pair (given by
+    /// [`Update::PushItems::position_hint`]). This is _the offset_. To this
+    /// offset, the algorithm adds the position's index of the new items
+    /// (still given by [`Update::PushItems::position_hint`]). This is
+    /// _the index_. This logic works for all cases as long as pairs are
+    /// maintained according to the rules hereinabove.
+    ///
+    /// That's a pretty memory compact and computation efficient way to map a
+    /// `Stream<Item = Vec<Update<Item, Gap>>>` into a `Stream<Item =
+    /// Vec<VectorDiff<Item>>>`. The larger the `LinkedChunk` capacity is, the
```
bnjbvr, line 144: nit: destination type not a `Stream` anymore

crates/matrix-sdk/src/event_cache/linked_chunk/as_vector.rs
```diff
@@ -0,0 +1,529 @@
...
+    /// fewer pairs the algorithm will have to handle, e.g. for 1'000 items
+    /// and a `LinkedChunk` capacity of 128, it's only 8 pairs, be 256
+    /// bytes.
```
bnjbvr, line 147: what do you mean by `be` here? `i.e.` / `that is`? (if so let's use one of these two instead)

crates/matrix-sdk/src/event_cache/linked_chunk/updates.rs
```diff
@@ -61,23 +61,31 @@ pub enum Update<Item, Gap> {
...
-    /// A chunk of kind Items has been truncated.
-    TruncateItems {
-        /// The identifier of the chunk.
-        chunk: ChunkIdentifier,
-
-        /// The new length of the chunk.
-        length: usize,
+    /// The last items of a chunk have been detached, i.e. the chunk has been
+    /// truncated.
+    DetachLastItems {
```
bnjbvr, line 81: Not sure this name is very telling, after reading it in context. Could we rename this to `SplitOff`, maybe?

crates/matrix-sdk/src/event_cache/linked_chunk/as_vector.rs
```diff
@@ -0,0 +1,529 @@
...
+                                .position(|(chunk_identifier, _)| chunk_identifier == next)
+                                // SAFETY: Assuming `LinkedChunk` and `Updates` are not buggy, and
+                                // assuming `Self::chunks` is correctly initialized, it is not
+                                // possible to insert a chunk between two chunks where one does not
+                                // exist. If this predicate fails, it means `LinkedChunk` or
+                                // `Updates` contain a bug.
+                                .expect("Inserting new chunk: The chunk is not found");
+
+                            debug_assert!(
+                                matches!(self.chunks.get(next_chunk_index - 1), Some((p, _)) if p == previous),
```
bnjbvr, line 278: `next_chunk_index` should always be >0 too, right? Just making sure this subtract is safe…

crates/matrix-sdk/src/event_cache/linked_chunk/as_vector.rs
```diff
@@ -0,0 +1,529 @@
...
+                        match control_flow {
+                            ControlFlow::Break(value) => Some(value),
+                            ControlFlow::Continue(..) => None,
+                        }
+                    }
+                    // SAFETY: Assuming `LinkedChunk` and `Updates` are not buggy, and assuming
+                    // `Self::chunks` is correctly initialized, it is not possible to push items on
+                    // a chunk that does not exist. If this predicate fails, it means `LinkedChunk`
+                    // or `Updates` contain a bug.
+                    .expect("Pushing items: The chunk is not found");
```
bnjbvr, line 336: Could this be replaced with a loop? I find it hard to follow:

- we introduce an option at the end, that we try to unwrap immediately thereafter
- ending with matching that control_flow is `Continue(..)` means that we never found the item, which could be express much more explicitly (e.g. by keeping `let mut found = false` flag set to `true` when found, with an explicit `break` statement in the loop)

crates/matrix-sdk/src/event_cache/linked_chunk/as_vector.rs
```diff
@@ -0,0 +1,529 @@
...
+
+                    *chunk_length += items.len();
+
+                    // See `mute_push_items` to learn more.
+                    if mute_push_items {
+                        continue;
+                    }
+
+                    // Optimisation: we can emit a `VectorDiff::Append` in this particular case.
+                    if chunk_index == self.chunks.len() - 1 {
```
bnjbvr, line 346: Just to make sure we don't underflow:

```suggestion
                    if chunk_index + 1 == self.chunks.len() {
```

crates/matrix-sdk/src/event_cache/linked_chunk/as_vector.rs
```diff
@@ -0,0 +1,529 @@
...
+        // ^^^^^^^^^^^^^^^^
+        // |
+        // new
+        assert_eq!(
+            as_vector.take(),
+            &[
+                VectorDiff::Append { values: vector!['a', 'b', 'c'] },
+                VectorDiff::Append { values: vector!['d'] },
+            ]
+        );
```
bnjbvr, line 424: Nice! Alternatively, or in addition to those checks, could we also have the corresponding vector that we're building, and applying the diffs then checking its content? Then the comments don't really need to be put there, we can "just" have the assertions on the entire vectors:

```
assert_eq!(reconstructed_vector, vector!['a', 'b', 'c', 'd']);
```

crates/matrix-sdk/src/event_cache/linked_chunk/updates.rs
```diff
@@ -61,23 +61,31 @@ pub enum Update<Item, Gap> {
...
+        /// The split position. Before this position, items are kept, after this
+        /// position, items are detached.
+        at: Position,
     },
+
+    ReattachItems,
+    ReattachItemsDone,
 }
 
 impl<Item, Gap> Clone for Update<Item, Gap>
```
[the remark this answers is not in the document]
Hywan: `Update<Item, Gap` implements `Clone` if and only if `Item: Clone` and `Gap: Clone`.

crates/matrix-sdk/src/event_cache/linked_chunk/mod.rs
```diff
@@ -611,9 +729,33 @@ impl<const CAP: usize, Item, Gap> LinkedChunk<CAP, Item, Gap> {
...
     /// `Option::take()` directly but rather [`Updates::take`] for example.
+    ///
+    /// It returns `None` if updates are disabled, i.e. if this linked chunk has
+    /// been constructed with [`Self::new`], otherwise, if it's been constructed
+    /// with [`Self::new_with_update_history`], it returns `Some(…)`.
     pub fn updates(&mut self) -> Option<&mut Updates<Item, Gap>> {
         self.updates.as_mut()
     }
+
+    pub fn as_vector(&mut self) -> Option<AsVector<Item, Gap>> {
```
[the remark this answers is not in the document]
Hywan: How did I forget that… thanks!

crates/matrix-sdk/src/event_cache/linked_chunk/updates.rs
```diff
@@ -61,23 +61,31 @@ pub enum Update<Item, Gap> {
...
-        /// The identifier of the chunk.
-        chunk: ChunkIdentifier,
-
-        /// The new length of the chunk.
-        length: usize,
+    /// The last items of a chunk have been detached, i.e. the chunk has been
+    /// truncated.
+    DetachLastItems {
+        /// The split position. Before this position, items are kept, after this
+        /// position, items are detached.
```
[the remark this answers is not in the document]
Hywan: :+1:

crates/matrix-sdk/src/event_cache/linked_chunk/as_vector.rs
```diff
@@ -0,0 +1,529 @@
...
+    updates::{ReaderToken, Update, UpdatesInner},
+    ChunkIdentifier,
+};
+
+/// A type alias to represent a chunk's length. This is purely for commodity.
+pub(super) type ChunkLength = usize;
+
+/// A type that transforms a `Vec<Update<Item, Gap>>` —given by
+/// [`Updates`]— into a `Vec<VectorDiff<Item>>` —this
+/// type—. Basically, it helps to consume a [`LinkedChunk<CAP, Item, Gap>`] as
```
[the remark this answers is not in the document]
Hywan: :+1:

crates/matrix-sdk/src/event_cache/linked_chunk/as_vector.rs
```diff
@@ -0,0 +1,529 @@
...
+    token: ReaderToken,
+
+    /// Mapper from `Update` to `VectorDiff`.
+    mapper: UpdateToVectorDiff,
+}
+
+impl<Item, Gap> AsVector<Item, Gap> {
+    /// Create a new [`Self`].
+    ///
+    /// `updates_subscriber` is simply `UpdatesSubscriber`.
```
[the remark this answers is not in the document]
Hywan: :+1:

crates/matrix-sdk/src/event_cache/linked_chunk/as_vector.rs
Hywan: Indeed.

crates/matrix-sdk/src/event_cache/linked_chunk/as_vector.rs
Hywan: :+1:

crates/matrix-sdk/src/event_cache/linked_chunk/as_vector.rs
Hywan: That's something I want to discuss with you. `RoomEventCacheUpdate` has a `Clear` variant already. Clearing a `LinkedChunk` means removing data from the memory and probably the database. Clearing a `RoomEventCache` means removing data from the memory only as far as I know. The meaning and the usage aren't exactly the same. Let's discuss about that and maybe let's open a new PR if necessary.

crates/matrix-sdk/src/event_cache/linked_chunk/updates.rs
Hywan: `Detach…` was playing well with `Reattach…` 😇.

crates/matrix-sdk/src/event_cache/linked_chunk/as_vector.rs
Hywan: Yeah correct, if `next` is at position 0, then `previous` must be at position -1 at best, which doesn't exist. Or, alternatively, `Update::NewItemsChunk` or `Update::NewGapChunk` is totally broken, but in this case, we would have other problems 😉.

This subtract is safe.

crates/matrix-sdk/src/event_cache/linked_chunk/updates.rs
```diff
@@ -61,23 +61,31 @@ pub enum Update<Item, Gap> {
...
+    /// Items are pushed inside a chunk of kind Items.
+    PushItems {
+        /// The [`Position`] of the items.
+        ///
+        /// This value is given to prevent the need for position computations by
+        /// the update readers. Items are pushed, so the positions should be
+        /// incrementally computed from the previous items, which requires the
+        /// reading of the last previous item. With `position_hint`, the update
+        /// readers no longer need to do so.
+        position_hint: Position,
```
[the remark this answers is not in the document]
Hywan: I've had a hard time finding the best naming here. `at` could imply it can be _at any position_ within the chunk, so possibility before or in the middle of existing items, while it's impossible and will never happen. It's a _push_. The `Position` is not required. It's here only to save computations from the consumers. For example, if we need to insert in a database, we need to first read the items, sort them, and find the last items; from here we can compute the position for the newly pushed items. This reading + sorting can be costly, hence `position_hint`, it's a pre-computed position.

Naming it `at` sounds bad to me. Naming it `position` creates a feeling of inconsistency with the rest of the code. That's why I've decided to use `position_hint`. It may not be the best idea I had, but in my mental scheme, it works 😛.

crates/matrix-sdk/src/event_cache/linked_chunk/updates.rs
```diff
@@ -61,23 +61,31 @@ pub enum Update<Item, Gap> {
...
+    /// The last items of a chunk have been detached, i.e. the chunk has been
+    /// truncated.
+    DetachLastItems {
+        /// The split position. Before this position, items are kept, after this
+        /// position, items are detached.
+        at: Position,
     },
+
+    ReattachItems,
+    ReattachItemsDone,
```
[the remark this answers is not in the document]
Hywan: Yup, good idea.

crates/matrix-sdk/src/event_cache/linked_chunk/updates.rs
```diff
@@ -258,25 +268,37 @@ impl<Item, Gap> Updates<Item, Gap> {
...
 
-        UpdatesSubscriber { updates: Arc::downgrade(&self.inner), token }
+        AsVector::new(Arc::clone(&self.inner), token, initial_chunk_lengths)
+    }
+
+    pub(super) fn new_reader_token(&mut self) -> ReaderToken {
+        let mut inner = self.inner.write().unwrap();
+        inner.last_token += 1;
+
+        let last_token = inner.last_token;
```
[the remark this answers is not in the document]
Hywan: Good idea.

crates/matrix-sdk/src/event_cache/linked_chunk/as_vector.rs
```diff
@@ -0,0 +1,529 @@
...
+    /// `initial_chunk_lengths` must be pairs of all chunk identifiers with the
+    /// associated chunk length. The pairs must be in the exact same order than
+    /// in [`LinkedChunk`].
+    ///
+    /// # Safety
+    ///
+    /// This method is marked as unsafe only to attract the caller's attention
+    /// on the order of pairs. If the order of pairs are incorrect, the entire
+    /// algorithm will not work properly and is very likely to panic.
```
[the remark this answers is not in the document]
Hywan: Indeed, it simplifies the code greatly. `initial_chunk_lengths` is now handled by `UpdateToVectorDiff::new` (🆕).

crates/matrix-sdk/src/event_cache/linked_chunk/as_vector.rs
```diff
@@ -0,0 +1,529 @@
...
+    /// Take the new updates as [`VectorDiff`].
+    ///
+    /// It returns an empty `Vec` if there is no new `VectorDiff` for the
+    /// moment.
+    pub(super) fn take(&mut self) -> Vec<VectorDiff<Item>>
+    where
+        Item: Clone,
+    {
+        let mut updates = self.updates.write().unwrap();
+        let updates = updates.take_with_token(self.token);
```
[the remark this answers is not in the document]
Hywan: There is no different concept, I don't understand what you're referring to. I've changed the highlighted code a little bit to avoid having 2 variables with the `updates` name though.

crates/matrix-sdk/src/event_cache/linked_chunk/as_vector.rs
Hywan: I've removed the `Option`, indeed it's pointless. I wish [`ControlFlow::break_value`](https://doc.rust-lang.org/std/ops/enum.ControlFlow.html#method.break_value) lands soon in stable.

I like the `try_fold` approach better than a `for` loop though.

crates/matrix-sdk/src/event_cache/linked_chunk/as_vector.rs
Hywan: Good eye, done!

crates/matrix-sdk/src/event_cache/linked_chunk/as_vector.rs
Hywan: Done.

crates/matrix-sdk/src/event_cache/linked_chunk/as_vector.rs
Hywan: Also, to have a `VectorDiff::Clear` here, we must first have an `Update::Clear`, which isn't the case so far. Let's handle that as a separate feature patch.

crates/matrix-sdk/src/event_cache/linked_chunk/as_vector.rs
```diff
@@ -0,0 +1,573 @@
...
+                        .chunks
+                        .iter()
+                        .position(|(chunk_identifier, _)| {
+                            chunk_identifier == expected_chunk_identifier
+                        })
+                        // SAFETY: Assuming `LinkedChunk` and `Updates` are not buggy, and assuming
+                        // `Self::chunks` is correctly initialized, it is not possible to remove a
+                        // chunk that does not exist. If this predicate fails, it means
+                        // `LinkedChunk` or `Updates` contain a bug.
+                        .expect("Removing a chunk: The chunk is not found");
```
poljar, line 305: Why is it better to crash here instead of turning this into a no-op?

poljar (approved):
This mostly makes sense to me, I'm not that familiar with what you're trying to achieve and don't have time to do a deep dive as of now.

I left some nits, and it would probably be wise to get approval from @bnjbvr due to the above, but feel free to move forward if you disagree.

crates/matrix-sdk/src/event_cache/linked_chunk/updates.rs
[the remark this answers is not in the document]
Hywan: Fixed.

crates/matrix-sdk/src/event_cache/linked_chunk/as_vector.rs
```diff
@@ -0,0 +1,573 @@
...
+use super::{
+    updates::{ReaderToken, Update, UpdatesInner},
+    ChunkContent, ChunkIdentifier, Iter,
+};
+
+/// A type alias to represent a chunk's length. This is purely for commodity.
+type ChunkLength = usize;
+
+/// A type that transforms a `Vec<Update<Item, Gap>>` (given by
+/// [`Updates`](super::Update)) into a `Vec<VectorDiff<Item>>` (this
```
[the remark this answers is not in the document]
Hywan: Oh, this doc is wrong. Fixed. Thanks!

crates/matrix-sdk/src/event_cache/linked_chunk/as_vector.rs
```diff
@@ -0,0 +1,573 @@
...
+/// type). Basically, it helps to consume a [`LinkedChunk<CAP, Item, Gap>`] as
+/// if it was an [`eyeball::ObservableVector<Item>`].
```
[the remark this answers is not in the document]
Hywan: :+1:

crates/matrix-sdk/src/event_cache/linked_chunk/updates.rs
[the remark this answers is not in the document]
bnjbvr: I don't get it; if the `position_hint` is always used, and we always insert at the `position_hint` position, how is that conceptually different from the position `at` which we want to insert?

crates/matrix-sdk/src/event_cache/linked_chunk/as_vector.rs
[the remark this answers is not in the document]
bnjbvr: `self.updates` is rather some kind of `UpdateManager` / `UpdateDispatcher` / `ObservableUpdates` type, but no worries, I'll make a proposal in form of a PR later :)

crates/matrix-sdk/src/event_cache/linked_chunk/as_vector.rs
```diff
@@ -0,0 +1,573 @@
...
+
+    /// The token to read the updates.
+    token: ReaderToken,
+
+    /// Mapper from `Update` to `VectorDiff`.
+    mapper: UpdateToVectorDiff,
+}
+
+impl<Item, Gap> AsVector<Item, Gap> {
+    /// Create a new [`Self`].
```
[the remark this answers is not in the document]
Hywan: Indeed, solved and improved.

crates/matrix-sdk/src/event_cache/linked_chunk/as_vector.rs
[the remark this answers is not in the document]
Hywan: Renamed to `ObservableUpdates` 👍.

crates/matrix-sdk/src/event_cache/linked_chunk/as_vector.rs
poljar: Doesn't need to be silent, a log line isn't for example. But sure, if it really can't happen.

crates/matrix-sdk/src/event_cache/linked_chunk/updates.rs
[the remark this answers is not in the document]
Hywan: Renamed to `at`.

crates/matrix-sdk/src/event_cache/linked_chunk/updates.rs
```diff
@@ -61,75 +61,110 @@ pub enum Update<Item, Gap> {
...
-        /// The identifier of the chunk.
-        chunk: ChunkIdentifier,
-
-        /// The new length of the chunk.
-        length: usize,
+    /// The last items of a chunk have been detached, i.e. the chunk has been
+    /// truncated.
+    DetachLastItems {
+        /// The split position. Before this position (`[..position]`), items are
+        /// kept, from this position (`[position..]`), items are
```
bnjbvr, line 83: Are the examples in the parens supposed to be using range notation? I found the `[]` confusing, and thought it was maths intervals, and I was *just* about to request either `[..position[` or `]position..]` to make this clearer. Maybe we could remove the brackets?

crates/matrix-sdk/src/event_cache/linked_chunk/mod.rs
```diff
@@ -24,22 +102,39 @@ use std::{
...
 pub enum Error {
+    /// A chunk identifier is invalid.
     #[error("The chunk identifier is invalid: `{identifier:?}`")]
-    InvalidChunkIdentifier { identifier: ChunkIdentifier },
+    InvalidChunkIdentifier {
+        /// The chunk identifier.
+        identifier: ChunkIdentifier,
+    },
 
+    /// A chunk is a gap, and it was expected to be an items.
```
bnjbvr, line 118:
```suggestion
    /// A chunk is a gap, and it was expected to be an items chunk.
```

crates/matrix-sdk/src/event_cache/linked_chunk/mod.rs
```diff
@@ -24,22 +102,39 @@ use std::{
...
     #[error("The chunk is a gap: `{identifier:?}`")]
-    ChunkIsAGap { identifier: ChunkIdentifier },
+    ChunkIsAGap {
+        /// The chunk identifier.
+        identifier: ChunkIdentifier,
+    },
 
+    /// A chunk is an items, and it was expected to be a gap.
```
bnjbvr, line 125:
```suggestion
    /// A chunk is an items chunk, and it was expected to be a gap.
```

crates/matrix-sdk/src/event_cache/linked_chunk/as_vector.rs
```diff
@@ -0,0 +1,584 @@
...
+        drop(linked_chunk);
+        assert!(as_vector.take().is_empty());
+
+        // Finally, ensure the “reconstitued” vector is the one expected.
+        assert_eq!(
+            accumulator,
+            vector![
+                'm', 'a', 'w', 'x', 'y', 'z', 'b', 'c', 'd', 'i', 'j', 'k', 'l', 'e', 'f', 'g', 'h'
+            ]
+        );
```
bnjbvr, line 582: Instead of doing this assert once at the end, could we do it in `apply_and_assert_eq`? Iterate on all the items of the linked chunk to reconstruct a `Vector` from it, and compare it against the accumulator \o/

crates/matrix-sdk/src/event_cache/linked_chunk/as_vector.rs
```diff
@@ -0,0 +1,584 @@
...
+
+                    // See `mute_push_items` to learn more.
+                    if mute_push_items {
+                        continue;
+                    }
+
+                    // Optimisation: we can emit a `VectorDiff::Append` in this particular case.
+                    if chunk_index + 1 == self.chunks.len() {
+                        diffs.push(VectorDiff::Append { values: items.into() });
+                    }
```
bnjbvr, line 365: Should we only emit append if `items.len() > 1`, and emit `PushBack` otherwise?

crates/matrix-sdk/src/event_cache/linked_chunk/as_vector.rs
```diff
@@ -0,0 +1,584 @@
...
+                            // Chunk has not been found.
+                            ControlFlow::Continue(..) => {
+                                // SAFETY: Assuming `LinkedChunk` and `ObservableUpdates` are not
+                                // buggy, and assuming `Self::chunks` is correctly initialized, it
+                                // is not possible to push items on a chunk that does not exist. If
+                                // this predicate fails, it means `LinkedChunk` or
+                                // `ObservableUpdates` contain a bug.
+                                panic!("Pushing items: The chunk is not found");
+                            }
+                        }
```
bnjbvr, line 352: I still find this a bit obtuse and harder to read for newcomers to this code, would really prefer a plain loop with a `let mut found = false` flag :pensive:

bnjbvr (approved):
Thanks! An amazing chunk of work :drum:

crates/matrix-sdk/src/event_cache/linked_chunk/updates.rs
Hywan: :+1:

crates/matrix-sdk/src/event_cache/linked_chunk/mod.rs
Hywan: :+1:

crates/matrix-sdk/src/event_cache/linked_chunk/mod.rs
Hywan: :+1:

crates/matrix-sdk/src/event_cache/linked_chunk/as_vector.rs
Hywan: I think it's robust enough right now. The idea of comparing a rebuilt vector is to ensure the `VectorDiff`s we expect are the correct ones. Having this once at the end seems fine to me.

crates/matrix-sdk/src/event_cache/linked_chunk/as_vector.rs
Hywan: We can have multiple optimisations here. I would let that for another PR, this one is large enough :-D.

PR #3393, opened 2024-05-09
FFI: Support the use of via parameters when looking up and joining rooms.

pixlwave:
This PR makes the following changes:
- When resolving a room alias, also return the server names so they can be used later if needed.
- Expose `join_room_by_alias_or_id` which allows passing the server names from the via parameters in a permalink.

It doesn't help with getting the room preview which also needs to support the via parameters, but I haven't figured out how that one works yet.

### DIFF
--- a/bindings/matrix-sdk-ffi/src/client.rs
+++ b/bindings/matrix-sdk-ffi/src/client.rs
@@ -49,7 +49,7 @@ use ruma::{
         room::power_levels::RoomPowerLevelsEventContent, GlobalAccountDataEventType,
     },
     push::{HttpPusherData as RumaHttpPusherData, PushFormat as RumaPushFormat},
-    RoomAliasId,
+    OwnedServerName, RoomAliasId, RoomOrAliasId,
 };
 use serde::{Deserialize, Serialize};
 use serde_json::Value;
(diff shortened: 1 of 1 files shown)


bindings/matrix-sdk-ffi/src/client.rs
```diff
@@ -725,6 +725,21 @@ impl Client {
         Ok(Arc::new(Room::new(room)))
     }
```
pixlwave, line 726: I'm not sure if there's any benefit of leaving this `join_room_by_id` method here or not. It seems redundant to me having both, but then the SDK supports both so 🤷‍♂️

poljar: Well the spec supports both so we have it in the SDK.

bindings/matrix-sdk-ffi/src/client.rs
```diff
@@ -725,6 +725,21 @@ impl Client {
...
 
+    pub async fn join_room_by_id_or_alias(
```
poljar, line 728: Don't you want to add some docs, now that uniffi forwards them through the FFI?

bindings/matrix-sdk-ffi/src/client.rs
```diff
@@ -789,6 +807,21 @@ impl From<NotificationProcessSetup> for MatrixNotificationProcessSetup {
     }
 }
 
+#[derive(uniffi::Record)]
+pub struct ResolvedRoomAlias {
```
poljar, line 811: Seems sensible, but some docs would be nice as well.

poljar (changes requested):
Makes sense, but we might want to start requiring docs in the bindings.

pixlwave:
Oh yes, I forgot we had docs now, thanks for the reminder! Added.

bindings/matrix-sdk-ffi/src/client.rs
pixlwave: Sorry yes, I'm just not sure if there's an advantage of using this one through the FFI and was thinking "out loud". Will leave them both - the docs will help clarify.

poljar: Yeah, I mean somebody else might be using the bindings and might find it useful. So I think it makes sense to just leave it there.

poljar (approved):

PR #3396, opened 2024-05-10
ci: Bump version of rust nightly

zecakeh:
Includes fixes for:

- Warnings when using `default_features` instead of `default-features` in `Cargo.toml`: Cargo says that it will not be supported anymore in the upcoming 2024 edition
- Errors with `cfg` options: Cargo now [checks cfgs at compile-time](https://blog.rust-lang.org/2024/05/06/check-cfg.html), which means that custom options need to be declared in build scripts. It also detects cargo features used in code that do not exist in the manifest.


andybalaam (approved):
Great, thanks!

PR #3398, opened 2024-05-11
ui: Expose message mentions

zecakeh:
Our main use case for this in Fractal is to check if there is a intentional room mention in the message to see if we need to try to look for `@room` in the text to replace it with a visual aid.

This has another benefit for edits: the users that were mentioned in the original event are not supposed to be mentioned again when sending an edit. Ruma takes care of that, but only if the original mentions are provided, which is now the case.

I didn't add the mentions in `Message::to_content()` because in both cases where it the method is called it would be unused.


poljar (approved):
Makes sense.

PR #3406, opened 2024-05-13
ui: Make Timeline::edit take a RoomMessageEventContentWithoutRelation

zecakeh:
That way the API is closer to that of `send_reply`.


Hywan (approved):
Ah great, that was something I wanted to improve! Thank you very much!

PR #3407, opened 2024-05-13
room preview: allow passing through a list of servers to discover a room with MSC3266

### LINKED ISSUE #3395: Room previews using the room summary API from MSC3266 fail for remote rooms
I noticed this when trying the new room permalink support in Element X iOS.

When requesting the room summary for a room that isn't in the room directory, the SDK tries the MSC3266 room summary endpoint next.
For example, when opening the URL `https://matrix.to/#/#matrix:matrix.org`, the following request reaches my server:
```
GET /_matrix/client/unstable/im.nheko.summary/rooms/!OGEhHVWSdvArJzumhm:matrix.org/summary
```
Notice the absence of `via` URL parameters. This results in an error on the client and synapse complaining in the logs:
```
SynapseError: 400 - Missing via to query remote room
```
and returning:
```json
{"errcode":"M_UNKNOWN","error":"Missing via to query remote room"}
```
The SDK then falls back to `/state` and `/joined_members` which fail because my server doesn't know the room.

This results in unavailable room previews for rooms that my server doesn't know about, aka remote rooms. For a single-user homeserver, this affects virtually all rooms since, if I'm not already in the room, my server likely doesn't know about it.

matrix.to uses the same API, however when telling it to use my server the summary call works and I can view room details. The API request differs in two interesting ways:
```
GET /_matrix/client/unstable/im.nheko.summary/rooms/%23matrix%3Amatrix.org/summary?via=matrix.org&_cacheBuster=1365069511896160
```
1. The summary API is called with the room alias from the permalink instead of the room id, and
2. `via` parameters are added, in this case from the server name in the alias.

Either of which is apparently enough to get the expected result from the API (based on manual testing).

A similar problem happens for permalinks already containing `via` parameters, where the parameters are not passed on. For example, `https://matrix.to/#/!OGEhHVWSdvArJzumhm:matrix.org?via=matrix.org&via=wolfo.tech&via=blob.cat` results in the same API request and the same synapse error.

The relevant code was introduced in #3339, which only passes the room id to ruma: https://github.com/matrix-org/matrix-rust-sdk/blob/74b79d8212f1f4f161625bad34acfb4acf8ff65f/crates/matrix-sdk/src/room_preview.rs#L135-L139

bnjbvr:
Fixes #3395.

### DIFF
--- a/bindings/matrix-sdk-ffi/src/client.rs
+++ b/bindings/matrix-sdk-ffi/src/client.rs
@@ -771,36 +771,48 @@ impl Client {
         Ok(response.into())
     }
 
-    /// Get the preview of a room, to interact with it.
+    /// Given a room id, get the preview of a room, to interact with it.
     ///
-    /// The (optional) list of `via_servers` must be a list of servers that know
+    /// The list of `via_servers` must be a list of servers that know
     /// about the room and can resolve it, and that may appear as a `via`
-    /// parameter in e.g. a permalink URL. It can be empty.
-    pub async fn get_room_preview(
+    /// parameter in e.g. a permalink URL. This list can be empty.
+    pub async fn get_room_preview_from_room_id(
         &self,
-        room_id_or_alias: String,
+        room_id: String,
         via_servers: Vec<String>,
     ) -> Result<RoomPreview, ClientError> {
-        let room_id = if let Ok(parsed_room_id) = RoomId::parse(&room_id_or_alias) {
-            parsed_room_id
-        } else {
-            // Try to resolve it as an alias.
-            let Ok(room_alias) = RoomAliasId::parse(&room_id_or_alias) else {
-                return Err(anyhow!("room_id_or_alias is neither a room id or an alias").into());
-            };
-            let response = self.inner.resolve_room_alias(&room_alias).await?;
-            response.room_id
-        };
+        let room_id = RoomId::parse(&room_id).context("room_id is not a valid room id")?;
 
         let via_servers = via_servers
             .into_iter()
             .map(ServerName::parse)
             .collect::<Result<Vec<_>, _>>()
-            .context("invalid via server name")?;
+            .context("at least one `via` server name is invalid")?;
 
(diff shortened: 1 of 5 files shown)


bindings/matrix-sdk-ffi/src/client.rs
```diff
@@ -772,9 +772,14 @@ impl Client {
     }
 
     /// Get the preview of a room, to interact with it.
+    ///
+    /// The (optional) list of `via_servers` must be a list of servers that know
```
Hywan, line 776: The list isn't optional as in `Option`. Do you mean it can be an empty vector?

Hywan (approved):

bindings/matrix-sdk-ffi/src/client.rs
bnjbvr: Yes, the `(optional)` might be redundant because I say it can be empty in the next sentence.

(Thanks for the quick review!)

bnjbvr:
Note to self: rework so that there are two methods:

- `get_room_preview_with_id(room_id, via)`
- `get_room_preview_with_alias(room_alias)`

And ditto for the `join_room` methods.

bnjbvr:
Requesting another round of review because there's been some dramatic changes:

- passing a `via` list of servers to help resolving is only useful when there's a `RoomId`, not a `RoomAliasId`; as a result, the FFI method has been split into two methods.
- I suspect MSC3266's implementation will work better if we pass directly a room alias id, instead of 1. resolving the room alias to a room id, 2. and then pass the room alias id. The reasoning is the same as above: if we passed it only a room id, then it would fail because the `via` list would be empty; if we pass it the `RoomAliasId` directly, it might resolve it server-side. Your guess as good as mine.
  - this required a bit of a rewrite of the `get_room_preview` so it takes a `RoomOrAliasId` parameter, and a few matches here and there to make it work as intended. I didn't add a changelog entry because there hasn't been any release since the introduction of `get_room_preview`, as far as I can tell.

Hywan (commented):
I like the split and I believe it brings clarity, thanks!

Hywan (approved):
approving

PR #3408, opened 2024-05-13
xtask: Allow passing multiple targets to Swift's build-framework.

pixlwave:
Currently `cargo xtask swift build-framework` lets us either build a single target, or all supported targets. This PR allows to pass `--target triple-1 --target triple2` to build a subset of the supported targets.


bnjbvr (approved):

stefanceriu (commented):
🥳

PR #3413, opened 2024-05-15
ui: Take a `PathBuf` to send an attachment

zecakeh:
I found the previous API confusing. Firstly, the parameter name was "filename" when we want a file path. Secondly, we take a `String` when we want a `Path` (or `PathBuf` to get an owned value here).

### DIFF
--- a/crates/matrix-sdk-ui/CHANGELOG.md
+++ b/crates/matrix-sdk-ui/CHANGELOG.md
@@ -1,0 +1,10 @@
+# unreleased
+
+Breaking changes:
+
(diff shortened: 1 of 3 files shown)


poljar (approved):
Indeed, this makes more sense.

Hywan (approved):
Good to me too. Can you rebase on top of `main` please, before merging?

zecakeh:
I merge it actually. Do you prefer a rebase?

zecakeh:
For info, the auto-merge here has failed.

PR #3415, opened 2024-05-15
sdk: Improvements around `generate_image_thumbnail`

zecakeh:
The main improvement here is to be able to choose the format of the generated thumbnail to avoid errors because the image crate cannot encode to the same format as the original.

We also allow to send the attachment even if the generation fails, because thumbnails are optional after all.

See commits messages for details.


crates/matrix-sdk/src/attachment.rs
```diff
@@ -397,12 +394,14 @@ pub fn generate_image_thumbnail<R: BufRead + Seek>(
...
-    Ok((
-        data,
-        BaseThumbnailInfo {
-            width: Some(thumbnail_width.into()),
-            height: Some(thumbnail_height.into()),
-            size: Some(data_size.into()),
-        },
-    ))
+    let content_type = mime::Mime::from_str(image_format.to_mime_type())
+        .expect("image should give a valid mimetype");
```
bnjbvr: Can we avoid the `expect` and return an error instead?

crates/matrix-sdk/src/room/futures.rs
```diff
@@ -298,14 +293,7 @@ impl<'a> IntoFuture for SendAttachment<'a> {
                     let (data, res) = make_thumbnail(data);
 
                     let thumbnail = match res {
-                        Ok((thumbnail_data, thumbnail_info)) => {
-                            data_slot = thumbnail_data;
-                            Some(Thumbnail {
-                                data: data_slot,
-                                content_type: mime::IMAGE_JPEG,
```
bnjbvr, line 305: Nice catch. Is it worth adding a regression test, or too much hassle?

crates/matrix-sdk/CHANGELOG.md
```diff
@@ -12,6 +12,9 @@ Breaking changes:
 - A custom sliding sync proxy set with `ClientBuilder::sliding_sync_proxy` now takes precedence over a discovered proxy.
 - `Client::get_profile` was moved to `Account` and renamed to `Account::fetch_user_profile_of`. `Account::get_profile` was renamed to `Account::fetch_user_profile`.
 - `generate_image_thumbnail` now returns a `Thumbnail`.
+- It is know possible to select the format of a generated thumbnail.
```
bnjbvr, line 15: nit: now, not `know`

crates/matrix-sdk/src/attachment.rs
```diff
@@ -390,11 +376,17 @@ pub fn generate_image_thumbnail<R: BufRead + Seek>(
     let thumbnail = image.thumbnail(width, height);
     let (thumbnail_width, thumbnail_height) = thumbnail.dimensions();
 
+    let thumbnail_format = match format {
+        ThumbnailFormat::Always(format) => format,
+        ThumbnailFormat::Fallback(format) if !image_format.writing_enabled() => format,
+        _ => image_format,
```
bnjbvr, line 385: Can you use the explicit last variant instead of `_` here, so that if we add a new variant to `ThumbnailFormat`, the compiler warns us here, and we can then decide what's the best thing to do?

crates/matrix-sdk/src/room/futures.rs
```diff
@@ -299,7 +299,10 @@ impl<'a> IntoFuture for SendAttachment<'a> {
                             ImageError::ThumbnailBiggerThanOriginal
                             | ImageError::FormatNotSupported,
                         ) => None,
-                        Err(error) => return Err(error.into()),
+                        Err(error) => {
+                            tracing::error!("Failed to generate thumbnail: {error}");
+                            None
+                        }
```
bnjbvr, line 305: I like this; can we remove the previous match arm that handles two `ImageError` variants, so they're also logged?

Also, can we use `warn!` instead of `error!`, since these are non-critical?

zecakeh: I am not sure that `ThumbnailBiggerThanOriginal` should be logged, or at least not at the `warn` level.

It could allow the users to be lazy and always use `generate_thumbnail` without having to check the image's size. It is more an optimization of the code than a real error.

bnjbvr: Ok. Would it make sense to log the other error at the warn level, though?

zecakeh: Yes, definitely.

crates/matrix-sdk/src/attachment.rs
zecakeh: Changed

crates/matrix-sdk/src/room/futures.rs
zecakeh: I added tests that check this.

crates/matrix-sdk/CHANGELOG.md
zecakeh: Right, thanks

crates/matrix-sdk/src/attachment.rs
zecakeh: Changed

crates/matrix-sdk/src/room/futures.rs
zecakeh: Changed

zecakeh:
I took care of the reviews, merged `main` to solve conflicts, and added integration tests.

For the integration tests I needed a real image so I took the avatar from #matrix-rust-sdk:matrix.org.

poljar:
> For the integration tests I needed a real image so I took the avatar from #matrix-rust-sdk:matrix.org.

Nice :crab:.

crates/matrix-sdk/tests/integration/room/attachment/mod.rs
```diff
@@ -290,3 +292,230 @@ async fn test_room_attachment_send_mentions() {
...
+    let room = client.get_room(&DEFAULT_TEST_ROOM_ID).unwrap();
+
+    let config = AttachmentConfig::new().generate_thumbnail(None, ThumbnailFormat::Original);
+
+    let response = room
+        .send_attachment("image", &mime::IMAGE_JPEG, IMAGE_BYTES.to_vec(), config)
+        .await
+        .unwrap();
+
+    assert_eq!(event_id!("$h29iv0s8:example.com"), response.event_id)
```
bnjbvr, line 350: Here and in the 3 other tests: can we replace the raw `event_id` with `test_json::EVENT_ID` too?

testing/matrix-sdk-test/src/lib.rs
```diff
@@ -93,7 +93,7 @@ macro_rules! init_tracing_for_tests {
                     // Since tracing_subscriber does prefix matching, the `matrix_sdk=` directive
                     // takes effect for all the main crates (`matrix_sdk_base`, `matrix_sdk_crypto`
                     // and so on).
-                    "info,matrix_sdk=debug".into()
+                    "info,wiremock=debug,matrix_sdk=debug".into()
```
bnjbvr, line 96: Can we revert this change? wiremock can be verbose in the output…

bnjbvr (approved):
Thanks for the tests!

testing/matrix-sdk-test/src/lib.rs
zecakeh: Ah right sorry. I had enabled it locally to understand why an event wasn't matching for an endpoint.

crates/matrix-sdk/tests/integration/room/attachment/mod.rs
zecakeh: `test_json::EVENT_ID` is a `serde_json::Value` containing a JSON object, so we can't use it for asserting the event ID, afaiu.

We would probably need a `matrix_sdk_test::DEFAULT_TEST_EVENT_ID`, like we have for the room ID for that.

PR #3418, opened 2024-05-15
sdk: Allow to send mentions with attachments

zecakeh:
Particularly useful for captions


Hywan (approved):
Looks good to me, thanks for the improvement!

PR #3420, opened 2024-05-16
FFI: Expose ed25519/curve25519 keys in bindings

BillCarsonFr:
The Element-X rageshakes are lacking device keys info, see https://github.com/element-hq/element-x-ios/issues/2550

This change exposes these so that it can be used by the app.
The `ed25519_key()` was already present in the sdk but not exposed in bindings. I added a second one for `curve25519_key()` in both sdk and bindings.

Can be accessed via `client.encryption().curve25519Key()` in the aps

Signed-off-by:

### DIFF
--- a/bindings/matrix-sdk-ffi/src/encryption.rs
+++ b/bindings/matrix-sdk-ffi/src/encryption.rs
@@ -213,20 +213,17 @@ impl From<encryption::VerificationState> for VerificationState {
 
 #[uniffi::export(async_runtime = "tokio")]
 impl Encryption {
-
-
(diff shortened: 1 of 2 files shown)


bnjbvr (changes requested):
Please follow the contribution guidelines when it comes to PR names.

poljar:
Could you make CI happy? I think that this is should be a pretty trivial change, but CI failures need to be addressed.

BillCarsonFr:
> Could you make CI happy? I think that this is should be a pretty trivial change, but CI failures need to be addressed.

Oh sry, I missed the failing checks

poljar (approved):

PR #3423, opened 2024-05-16
indexeddb: expose new method `IndexeddbCryptoStore::open_with_key`

richvdh:
Part of a resolution to https://github.com/element-hq/element-web/issues/26821: allow applications to skip the PBKDF2 operation if they already have a cryptographically secure key.

In order to maintain compatibility for existing element-web sessions, if we discover that we have an existing store that was encrypted with a key derived from PBKDF2, then we reconstruct what element-web used to do: specifically, we base64-encode the key to obtain the "passphrase" that was previously passed in. If that matches, we know we've got the right key, and can update the meta store accordingly.

### DIFF
--- a/crates/matrix-sdk-indexeddb/Cargo.toml
+++ b/crates/matrix-sdk-indexeddb/Cargo.toml
@@ -50,6 +50,7 @@ matrix-sdk-base = { workspace = true, features = ["testing"] }
 matrix-sdk-common = { workspace = true, features = ["js"] }
 matrix-sdk-crypto = { workspace = true, features = ["js", "testing"] }
 matrix-sdk-test = { workspace = true }
+rand = { workspace = true }
 tracing-subscriber = { version = "0.3.18", default-features = false, features = ["registry", "tracing-log"] }
 uuid = "1.3.0"
 wasm-bindgen-test = "0.3.33"
--- a/crates/matrix-sdk-indexeddb/Cargo.toml
+++ b/crates/matrix-sdk-indexeddb/Cargo.toml
@@ -38,6 +38,9 @@ tokio = { workspace = true }
 tracing = { workspace = true }
 wasm-bindgen = "0.2.83"
 web-sys = { version = "0.3.57", features = ["IdbKeyRange"] }
+hkdf = "0.12.4"
+zeroize.workspace = true
+sha2.workspace = true
 
 [target.'cfg(target_arch = "wasm32")'.dependencies]
 # for wasm32 we need to activate this
--- a/crates/matrix-sdk-indexeddb/Cargo.toml
+++ b/crates/matrix-sdk-indexeddb/Cargo.toml
@@ -39,8 +39,8 @@ tracing = { workspace = true }
 wasm-bindgen = "0.2.83"
 web-sys = { version = "0.3.57", features = ["IdbKeyRange"] }
 hkdf = "0.12.4"
-zeroize.workspace = true
-sha2.workspace = true
+zeroize = { workspace = true }
+sha2 = { workspace = true }
 
 [target.'cfg(target_arch = "wasm32")'.dependencies]
 # for wasm32 we need to activate this
--- a/crates/matrix-sdk-indexeddb/src/crypto_store/mod.rs
+++ b/crates/matrix-sdk-indexeddb/src/crypto_store/mod.rs
@@ -272,7 +272,22 @@ impl IndexeddbCryptoStore {
         IndexeddbCryptoStore::open_with_store_cipher("crypto", None).await
     }
 
-    /// Open a new `IndexeddbCryptoStore` with given name and passphrase
+    /// Open an `IndexeddbCryptoStore` with given name and passphrase.
+    ///
+    /// If the store previously existed, the encryption cipher is initialised
+    /// using the given passphrase and the details from the meta store. If the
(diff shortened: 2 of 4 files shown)


crates/matrix-sdk-indexeddb/src/crypto_store/mod.rs
```diff
@@ -1289,6 +1313,103 @@ impl Drop for IndexeddbCryptoStore {
...
+
+/// Given a serialised store cipher, try importing with the given key.
+///
+/// This is a helper for [`IndexeddbCryptoStore::open_with_key`].
+async fn import_store_cipher_with_key(
+    key: &[u8; 32],
+    serialised_cipher: &[u8],
+    db: &IdbDatabase,
+) -> Result<StoreCipher, IndexeddbCryptoStoreError> {
+    let cipher = match StoreCipher::import_with_key(key, serialised_cipher) {
```
richvdh, line 1389: I'm a bit uneasy about using the key directly here, not least because the same key (after an HKDF) is used to derive an AES key to encrypt things in the react sdk's store.

Should we add another pair of methods to `StoreCipher` which will use an HKDF on the provided key, rather than using it directly as a ChaCha20-Poly1305 key?

(Note that we can't do the HKDF at a higher level, because that would break the fallback to passphrase below.)

poljar: > Should we add another pair of methods to StoreCipher which will use an HKDF on the provided key, rather than using it directly as a ChaCha20-Poly1305 key?

Hmm, can we do this here locally inside the IndexedDB store instead? This would avoid the need to expand the API surface of the `StoreCipher`.

richvdh: The thing is that, if we use an HKDF, really we would ideally use a salt, so we need somewhere to store the salt. We could store it separately of course, but it would go very nicely as another entry in the `KdfInfo` enum.

poljar: > The thing is that, if we use an HKDF, really we would ideally use a salt

Yes, if you're not sure in the randomness of your input key material, but you said that you are using 256 bits of uniformly random data, just that it's base64 encoded.

Is that not the case?

richvdh: As used by the react-sdk, yes it's 256 bits of uniformly random data, and I don't think either an HKDF or a salt are particularly necessary. It's just that the same 256 bits are used to derive a separate key elsewhere; reusing the same cryptographic material for multiple keys is bad practice. And this is going to be annoying to change later.

I'm not that bothered. LMK

poljar: Well yes, I'm not against the HKDF step, I'm questioning the necessity of the salt, which then requires you to store the salt, which requires to extend the `KdfInfo` enum.

You can provide the domain separation you're seeking by providing the `info` parameter to the HDKF function.

richvdh: ok, I'll do that.

richvdh: Now done.

crates/matrix-sdk-indexeddb/Cargo.toml
```diff
@@ -38,6 +38,9 @@ tokio = { workspace = true }
 tracing = { workspace = true }
 wasm-bindgen = "0.2.83"
 web-sys = { version = "0.3.57", features = ["IdbKeyRange"] }
+hkdf = "0.12.4"
+zeroize.workspace = true
```
poljar, line 42: Can we stick to one style for defining workspace deps? The one with the braces was used until now exclusively.

```suggestion
zeroize = { workspace = true }
```

crates/matrix-sdk-indexeddb/src/crypto_store/mod.rs
```diff
@@ -315,17 +309,58 @@ impl IndexeddbCryptoStore {
...
+    /// did not previously exist, a new encryption cipher is derived from
+    /// the passphrase, and the details are stored to the metastore.
+    ///
+    /// The store is then opened, or a new one created, using the encryption
+    /// cipher.
+    ///
+    /// # Arguments
+    ///
+    /// * `prefix` - Common prefix for the names of the two IndexedDB stores.
+    /// * `key` - key with which to encrypt the key which is used to encrypt the
```
poljar, line 339:
```suggestion
    /// * `key` - Key with which to encrypt the key which is used to encrypt the
```

poljar (approved):
Left a couple of nits, looks good otherwise.

crates/matrix-sdk-indexeddb/Cargo.toml
richvdh: sure, sorry, that was just what `cargo add` gave me.

crates/matrix-sdk-indexeddb/Cargo.toml
```diff
@@ -38,6 +38,9 @@ tokio = { workspace = true }
...
+sha2.workspace = true
```
richvdh, line 43:
```suggestion
sha2 = { workspace = true }
```

PR #3424, opened 2024-05-16
room list service: always request the room name event

bnjbvr:
And don't rely on the `name` field in the response as provided by the server. The name displayed by the server was before saved as the `m.room.name` event for that room, which is not the right semantics when the name has been computed by the server. The computation of a name should happen client-side, so as to be able to translate it according to the user's language.

Instead, this PR makes it so that we ignore the `name` field from the sliding sync response, and instead request the `m.room.name` event for all the rooms in the room list service.

Conversely, the `m.room.avatar` is always set to the `avatar` field from a room sub-part of a response, and it's likely more precise than the actual content of the `m.room.avatar` (it will user the other user avatar for a DM room). We could also compute that client-side, but since the server does it for us, and its default choice is Good™, we use that instead. As a result, the state event was requested but always unused, so we can stop requesting it explicitly in the room list / notification services.

### DIFF
--- a/crates/matrix-sdk-base/src/sliding_sync.rs
+++ b/crates/matrix-sdk-base/src/sliding_sync.rs
@@ -678,16 +678,11 @@ async fn cache_latest_events(
 }
 
 fn process_room_properties(room_data: &v4::SlidingSyncRoom, room_info: &mut RoomInfo) {
-    // Handle the room's name.
-    if let Some(name) = &room_data.name {
-        room_info.update_name(name.to_owned());
-    }
-
     // Handle the room's avatar.
     //
-    // It can be updated via the state events, or via the `SlidingSyncRoom::avatar`
-    // field. This part of the code handles the latter case. The former case is
-    // handled by [`BaseClient::handle_state`].
+    // It can be updated via the state events, or via the
+    // [`v4::SlidingSyncRoom::avatar`] field. This part of the code handles the
+    // latter case. The former case is handled by [`BaseClient::handle_state`].
     match &room_data.avatar {
         // A new avatar!
         JsOption::Some(avatar_uri) => room_info.update_avatar(Some(avatar_uri.to_owned())),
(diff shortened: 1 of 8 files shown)


crates/matrix-sdk-base/src/sliding_sync.rs
```diff
@@ -1,5 +1,3 @@
-// Copyright 2023 The Matrix.org Foundation C.I.C.
-//
```
bmarty, line 2: Is the copyright removed on purpose?

crates/matrix-sdk-base/src/sliding_sync.rs
```diff
@@ -1194,7 +1194,7 @@ mod tests {
 
         // When I send sliding sync response containing an invited room
         let mut room = v4::SlidingSyncRoom::new();
-        set_room_invited(&mut room, user_id);
+        set_room_invited(&mut room, user_id, user_id);
```
bmarty, line 1197: This look like a bit weird, `inviter` and `invitee` are distinct users, no? (repeated 3 times, line 1221 and 1244). So maybe create a distinct user?

bnjbvr: Oh this `set_room_invited` is a test function, and I've checked that all those tests don't actually care about the inviter/invitee distinction, since they observe other side-effects.

crates/matrix-sdk-base/src/sliding_sync.rs
bnjbvr: Nope, thanks 😭

andybalaam (approved):
Looks good, thanks!

bmarty:
I confirm that `rawName` is not populated anymore with the computed name, and `displayName` still contains a computed name.

Also avatar of room where user is invited are correctly rendered.

crates/matrix-sdk-base/src/sliding_sync.rs
bmarty: OK, thanks for checking!

Issue #3361, opened 2024-04-29
timeline: introduce a Send Queue API


bnjbvr:
- [x] Make it so that the sending queue is shared by all the timeline instances of the same room (live or detached), so that local echoes appear on all the live/detached timelines, if need be
- [x] Allow retry sending
- [x] Allow enabling/disabling the sending queue globally
- [x] Change course to have a per-room sending queue enablement flag
- [x] Expose whenever an event couldn't be sent because of an API error (status code 4XX or 5XX)
- [x] Mark such events from the queue as wedged, and don't block the queue for them.
- [x] Store and manage unsent events on disk
  - [x] make all the ruma `*EventContent` types serializable and deserializable with a type ([PR](https://github.com/ruma/ruma/pull/1850))
  - [x] [implement state store backends to save events on disk](https://github.com/matrix-org/matrix-rust-sdk/pull/3592)
  - [x] reload all the rooms that have at least one pending event when the Client restarts
  - [x] timeline: mark wedged local echoes with the proper failure send state
- [x] Allow redacting local echoes in a timeline
- [x] Allow editing local echoes in a timeline
- [x] Allow sending raw events
  - blocked by https://github.com/ruma/ruma/issues/1865

---

* Edit by @Hywan: Fix https://github.com/element-hq/element-x-ios/issues/1758

maan2003:
related https://github.com/matrix-org/matrix-rust-sdk/issues/3152

bnjbvr:
Ah thanks, this issue subsumes that one 👍

PR #3433, opened 2024-05-17
timeline queue: refactor the code a bit

bnjbvr:
Part of https://github.com/matrix-org/matrix-rust-sdk/issues/3361. This is me understanding the code and starting to refactor it a bit.

### DIFF
--- a/crates/matrix-sdk-ui/src/timeline/inner/mod.rs
+++ b/crates/matrix-sdk-ui/src/timeline/inner/mod.rs
@@ -645,7 +645,8 @@ impl<P: RoomDataProvider> TimelineInner<P> {
 
     /// Update the send state of a local event represented by a transaction ID.
     ///
-    /// If no local event is found, a warning is raised.
+    /// If the corresponding local timeline item is missing, a warning is
+    /// raised.
     #[instrument(skip_all, fields(txn_id))]
     pub(super) async fn update_event_send_state(
         &self,
@@ -718,9 +719,11 @@ impl<P: RoomDataProvider> TimelineInner<P> {
         txn.items.set(idx, new_item);
 
         if is_error {
-            // When there is an error, sending further messages is paused. This
-            // should be reflected in the timeline, so we set all other pending
-            // events to cancelled.
+            // When there is an error, sending further messages is paused. This should be
+            // reflected in the timeline, so we set all other pending events to
+            // cancelled.
+            //
+            // TODO(bnjbvr): spooky action at a distance here^
             let items = &mut txn.items;
             let num_items = items.len();
             for idx in 0..num_items {
@@ -797,6 +800,10 @@ impl<P: RoomDataProvider> TimelineInner<P> {
         Ok(follow_up_action)
     }
 
+    /// Attempts to find a local echo item for some event to be sent.
+    ///
+    /// If it's found, it's moved back to the end of the timeline, then its
+    /// content is returned by this function.
     pub(super) async fn prepare_retry(
         &self,
         txn_id: &TransactionId,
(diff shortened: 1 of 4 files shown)


crates/matrix-sdk-ui/src/timeline/inner/mod.rs
```diff
@@ -718,9 +719,11 @@ impl<P: RoomDataProvider> TimelineInner<P> {
...
 
         if is_error {
-            // When there is an error, sending further messages is paused. This
-            // should be reflected in the timeline, so we set all other pending
-            // events to cancelled.
+            // When there is an error, sending further messages is paused. This should be
+            // reflected in the timeline, so we set all other pending events to
+            // cancelled.
+            //
+            // TODO(bnjbvr): spooky action at a distance here^
```
poljar, line 726: While I like the physics reference and don't want you to remove it, maybe it would make sense what the physicist in you meant here? I can guess that it's probably about one item affecting/cancelling other items, it would be nice to write it down.

crates/matrix-sdk-ui/src/timeline/mod.rs
```diff
@@ -571,33 +571,32 @@ impl Timeline {
...
+            TimelineItemContent::Poll(poll_state) => AnyMessageLikeEventContent::UnstablePollStart(
+                UnstablePollStartEventContent::New(poll_state.into()),
+            ),
             TimelineItemContent::MembershipChange(_)
             | TimelineItemContent::ProfileChange(_)
-            | TimelineItemContent::OtherState(_) => {
-                error_return!("Retrying state events is not currently supported");
+            | TimelineItemContent::OtherState(_)
+            | TimelineItemContent::CallInvite => {
+                error_return!("Retrying state events/call invite is not currently supported");
```
poljar, line 589:
```suggestion
                error_return!("Retrying state events/call invites is not currently supported");
```

crates/matrix-sdk-ui/src/timeline/queue.rs
```diff
@@ -96,78 +101,104 @@ pub(super) async fn send_queued_messages(
...
+        timeline
+            .update_event_send_state(
+                &msg.txn_id,
+                EventSendState::SendingFailed {
+                    // FIXME: Probably not exactly right
+                    error: Arc::new(matrix_sdk::Error::InconsistentState),
+                },
+            )
+            .await;
+        return;
```
poljar, line 129: Why would you do a `if/return/implicit return` instead of a `if/else` here? I understand the above one to avoid a nested if, this one seems to go a bit too far with the removal of the nesting.

crates/matrix-sdk-ui/src/timeline/queue.rs
```diff
@@ -96,78 +101,104 @@ pub(super) async fn send_queued_messages(
...
+
+            // See above comment in the `SendingFailed` arm.
+            queue.clear();
         }
     }
 }
 
 /// Result of [`SendMessageTask`].
 enum SendMessageResult {
-    /// The message was sent successfully, and the local echo was updated to
```
poljar, line 158: Hmm, why is the additional context about the local echo not useful? Is it untrue or what?

poljar (approved):
Left a couple of nits, but looks good.

crates/matrix-sdk-ui/src/timeline/queue.rs
bnjbvr: Indeed, it's not updated anymore. Now the local echo is updated in `handle_send_result` for all the match variants, which seemed clearer to me.

crates/matrix-sdk-ui/src/timeline/queue.rs
bnjbvr: I do like that the main operation is not under a conditional, and just lives at the end in the main path, but I can revert this one change here.

crates/matrix-sdk-ui/src/timeline/inner/mod.rs
bnjbvr: Hah fun, I for one didn't know it was a physics reference xD

https://en.wikipedia.org/wiki/Action_at_a_distance_(computer_programming)

In this case, there's tight coupling between this loop and the behavior of the sending queue, which is just working by coincidence (both paths know they interact together that way), and it's not good. Will update the comment.

PR #3434, opened 2024-05-18
feat(calls): add support for sending Matrix RTC call notifications (MSC4075)

stefanceriu:
This PR adds 2 new methods exposed all the way up to the ffi level that allow the client application to send call notifications when entering an Element call.
* `send_call_notification_if_needed` will check whether there's already an active call before chosing the notify type base don whether the room is a DM or not
* `send_call_notification` is supposed to be used for very custom use cases, logic living on the client
* it's also exposing a new `CallNotify` event type for the client to use when receiving a push notification
* part of https://github.com/matrix-org/matrix-spec-proposals/blob/toger5/matrixrtc-call-ringing/proposals/4075-call-notify-event.md

NB I had a serious go at writing a test for this but setting it up is beyond my rust sdk knowledge. If you really think it's necessary please provide tips, tricks and send thoughts and prayers.


andybalaam:
The code looks sensible to me but I have asked for help from other team members with more context since I'm not confident I understand the implications of this.

andybalaam:
In terms of testing, it looks like you might be able to do it using a `MockServer` like `test_cache_invalidation_while_encrypt` does in `crates/matrix-sdk/src/room/mod.rs`

stefanceriu:
> In terms of testing, it looks like you might be able to do it using a `MockServer` like `test_cache_invalidation_while_encrypt` does in `crates/matrix-sdk/src/room/mod.rs`

Thanks for the review. I already looked around but honestly all of this feels ridiculously complicated just to check a trivial piece of logic i.e. send a ring if it's a dm. Isn't there a simpler way to mock rooms in the rust sdk?

andybalaam:
> Thanks for the review. I already looked around but honestly all of this feels ridiculously complicated just to check a trivial piece of logic i.e. send a ring if it's a dm. Isn't there a simpler way to mock rooms in the rust sdk?

It does seem unfortunate that `Room` holds a concrete `Client` reference rather than some kind of trait object or type param. I'll ask the team in the [room](https://matrix.to/#/#matrix-rust-sdk-dev:flipdot.org) what they think.

stefanceriu:
Alrighty, I managed to make something happen, please have another look

bindings/matrix-sdk-ffi/src/ruma.rs
```diff
@@ -375,6 +376,31 @@ impl From<RumaMessageType> for MessageType {
...
+    Ring,
+    Notify,
+}
+
+impl From<RumaNotifyType> for NotifyType {
+    fn from(val: RumaNotifyType) -> Self {
+        match val {
+            RumaNotifyType::Notify => Self::Notify,
+            RumaNotifyType::Ring => Self::Ring,
+            _ => Self::Notify,
```
Hywan, line 390:
```suggestion
            RumaNotifyType::Ring => Self::Ring,
            RumaNotifyType::Notify | _ => Self::Notify,
```

or:

```suggestion
            RumaNotifyType::Ring => Self::Ring,
            _ => Self::Notify,
```

andybalaam (approved):
Looks great, thanks!

PR #3437, opened 2024-05-21
day dividers: resolve a few extra bugs

bnjbvr:
This checks in a few tests found thanks to some local poorman's fuzzing. They were failing locally and have been fixed since, see each commit that contains one failing test and its fix. These are likely showing real-world bugs, where in the worst case, the last message could disappear because it's been incorrectly removed as if it were a duplicate day divider.

### DIFF
--- a/crates/matrix-sdk-ui/src/timeline/day_dividers.rs
+++ b/crates/matrix-sdk-ui/src/timeline/day_dividers.rs
@@ -142,7 +142,14 @@ impl DayDividerAdjuster {
         for (i, item) in items.iter().enumerate().rev() {
             if item.is_day_divider() {
                 // The item is a trailing day divider: remove it.
-                self.ops.push(DayDividerOperation::Remove(i));
+                if !self
+                    .ops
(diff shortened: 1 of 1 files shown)


crates/matrix-sdk-ui/src/timeline/day_dividers.rs
```diff
@@ -369,10 +369,16 @@ impl DayDividerAdjuster {
         };
 
         // Assert invariants.
-        // 1. The timeline starts with a date separator.
+        // 1. The timeline starts with a day divider.
         if let Some(item) = items.get(0) {
-            if !item.is_day_divider() {
-                report.errors.push(DayDividerInsertError::FirstItemNotDayDivider)
+            if item.is_read_marker() {
+                if let Some(next_item) = items.get(1) {
```
andybalaam, line 383: I assume there can only be 1 read marker? Is it just for this user?

andybalaam (approved):
Great, thanks! Just one question.

PR #3439, opened 2024-05-21
feat: store and restore composer drafts

### LINKED ISSUE #3430: Store drafts for each room so that clients can restore them at a later time
We should store message drafts in the local encrypted store for security reason, so that clients can save them and restore them at a later time.
We should store drafts in a dictionary using the room IDs to which the draft belongs as a key.
The draft should consist in a simple struct that sores the plain text, an optional html text in case the message is using rich text formatting, and if the draft is a normal message, a reply or an edit (the latter two should also store the event ID they are replying/editing)
We should then expose bindings to save and restore the draft given the room ID.

Velin92:
Exposed a function in both the FFI room binding and the sdk room, to allow to store and restore the composer draft for that specific room in the persistent storage.

fixes #3430


crates/matrix-sdk-base/src/store/traits.rs
```diff
@@ -808,6 +809,31 @@ pub enum StateStoreDataValue {
...
+
+#[derive(Debug, Clone, Serialize, Deserialize, uniffi::Enum)]
+pub enum DraftType {
+    /// The draft is a new message.
+    NewMessage,
+    /// The draft is a reply to an event.
+    Reply { event_id: String },
+    /// The draft is an edit to an event.
+    Edit { event_id: String },
 }
```
Velin92, line 837: not sure what is the best place where to put this code.

crates/matrix-sdk-base/src/store/traits.rs
```diff
@@ -808,6 +809,31 @@ pub enum StateStoreDataValue {
...
+    /// The draft content in plain  text.
+    plain_text: String,
+    /// If the message is formatted in HTML, the HTML representation of the
+    /// message.
+    html_text: Option<String>,
+    /// The type of draft.
+    draft_type: DraftType,
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize, uniffi::Enum)]
```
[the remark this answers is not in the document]
bnjbvr: Yep: when you're working locally, most Cargo features get enabled, thanks to [Cargo feature unification](https://doc.rust-lang.org/cargo/reference/features.html#feature-unification) across all the crates. But when build in isolation, it won't be the case.

Here, I think that means that you're using the `uniffi::Enum` derive macro, while the `uniffi` cargo feature might be disabled. Hence, you need to conditionally derive this, something like:

```suggestion
#[derive(Debug, Clone, Serialize, Deserialize)]
#[cfg_attr(feature = "uniffi", derive(uniffi::Enum))]
```

bindings/matrix-sdk-ffi/src/room.rs
```diff
@@ -644,6 +648,35 @@ impl Room {
...
+        self.inner.clear_composer_draft().await?;
+        Ok(())
+    }
+
+    pub async fn get_loaded_reply_details(
+        &self,
+        event_id: String,
+    ) -> Result<InReplyToDetails, ClientError> {
+        let event_id = EventId::parse(event_id)?;
+        let res = match self.inner.event(&event_id).await {
```
Hywan, line 672:
```diff
- let res
+ let details
```

bindings/matrix-sdk-ffi/src/timeline/mod.rs
```diff
@@ -68,7 +68,7 @@ use crate::{
     RUNTIME,
 };
 
-mod content;
+pub mod content;
```
Hywan, line 71: Is it a necessary change?

crates/matrix-sdk-base/src/store/traits.rs
```diff
@@ -808,6 +809,31 @@ pub enum StateStoreDataValue {
 
     /// A list of recently visited room identifiers for the current user
     RecentlyVisitedRooms(Vec<String>),
+
+    /// A composer draft for the room
+    ComposerDraft(ComposerDraft),
+}
+
+/// Struct that represents the current draft of the composer for the room
```
Hywan, line 817:
```diff
- for the room
+ for the room.
```

crates/matrix-sdk-base/src/store/traits.rs
```diff
@@ -808,6 +809,31 @@ pub enum StateStoreDataValue {
...
+#[derive(Debug, Clone, Serialize, Deserialize, uniffi::Record)]
+pub struct ComposerDraft {
+    /// The draft content in plain  text.
```
Hywan, line 820:
```diff
- plain  text.
+ plain text.
```

crates/matrix-sdk-base/src/store/traits.rs
```diff
@@ -808,6 +809,31 @@ pub enum StateStoreDataValue {
...
+    draft_type: DraftType,
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize, uniffi::Enum)]
+pub enum DraftType {
+    /// The draft is a new message.
+    NewMessage,
+    /// The draft is a reply to an event.
+    Reply { event_id: String },
+    /// The draft is an edit to an event.
```
Hywan, line 835:
```suggestion
    /// The draft is an edit of an event.
```

crates/matrix-sdk-base/src/store/traits.rs
```diff
@@ -808,6 +809,31 @@ pub enum StateStoreDataValue {
...
+    plain_text: String,
+    /// If the message is formatted in HTML, the HTML representation of the
+    /// message.
+    html_text: Option<String>,
+    /// The type of draft.
+    draft_type: DraftType,
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize, uniffi::Enum)]
+pub enum DraftType {
```
Hywan, line 830:
```suggestion
pub enum ComposerDraftType {
```

so that we are sure it goes hand in hand with `ComposerDraft`?

crates/matrix-sdk-base/src/store/traits.rs
```diff
@@ -846,6 +877,9 @@ pub enum StateStoreDataKey<'a> {
 
     /// Recently visited room identifiers
     RecentlyVisitedRooms(&'a UserId),
+
+    /// Composer draft for the room
```
Hywan, line 881: Can you elaborate what it is? Imagine you don't know what the editor is, what the composer is, what the draft is, and you've to explain that here :-). Also, you can create links to `ComposerDraft` and `DraftType` maybe, like so:

```rust
/// To learn more, see [`ComposerDraft`] and [`DraftType`].
///
/// [`ComposerDraft`]: path::to::ComposerDraft
/// [`DrafType`]: path::to::DraftType
```

crates/matrix-sdk-ui/src/timeline/event_item/content/message.rs
```diff
@@ -237,6 +237,13 @@ impl InReplyToDetails {
 
         InReplyToDetails { event_id, event: TimelineDetails::from_initial_value(event) }
     }
+
+    pub fn new_from_event(
```
Hywan, line 241: Please add missing documentation.

Also, what about this naming:

```suggestion
    pub fn new_with_timeline_details(
```

crates/matrix-sdk-ui/src/timeline/event_item/content/message.rs
```diff
@@ -237,6 +237,13 @@ impl InReplyToDetails {
...
+        event_id: OwnedEventId,
+        event: TimelineDetails<Box<RepliedToEvent>>,
+    ) -> InReplyToDetails {
```
Hywan, line 244:
```suggestion
    ) -> Self {
```

You can also modify the return type of the `new` method.

crates/matrix-sdk-ui/src/timeline/event_item/content/message.rs
```diff
@@ -279,7 +286,7 @@ impl RepliedToEvent {
         }
     }
 
-    pub(in crate::timeline) async fn try_from_timeline_event<P: RoomDataProvider>(
+    pub async fn try_from_timeline_event<P: RoomDataProvider>(
```
Hywan, line 289: Are we sure about this change?
If it goes public, it must have a documentation at least.

Hywan (dismissed):
Thanks for the PR. Almost there! A couple of small feedback, but the overall code looks nice, thanks!

bindings/matrix-sdk-ffi/src/timeline/mod.rs
Velin92: yeah I need for  `get_loaded_reply_details`

crates/matrix-sdk-base/src/store/traits.rs
Velin92: i don't see any other use case in which it is not tied to it, at least for now

crates/matrix-sdk-ui/src/timeline/mod.rs
```diff
@@ -220,8 +224,53 @@ impl Timeline {
...
+        }
+        .into();
+
+        Some(EventTimelineItem::new(
+            sender,
+            TimelineDetails::Unavailable,
+            timestamp,
+            item_content,
+            event_kind,
+        ))
```
Velin92, line 273: So this is an experimental solution that I tried to address the issue that I referenced in the meta issue @bnjbvr @Hywan 

Essentially this tries synthesises a timeline item if it is not found directly in the timeline, to address the issue of needing a timeline item for the edit/reply APIs, that would otherwise fail in case the timeline item has not been loaded/paginated yet.

Ofc the code is rough and I am probably doing something wrong (for example I am building the content as a `latest_event`)but I wanted to che check if this would work, since the only thing I can work with is the `event_id`

This makes replies work well! For edits instead what happens is that I am able to send the edit event, and the message appears as edited but its content does not result as changed at all, but I think is because I am not generating the original content properly. Any ideas on how to improve upon this, or if there is a better solution to address the issue?

bnjbvr: I think the proper way would be to change the API to send_reply or edit an event, so that instead of a `EventTimelineItem` it gets passed another parameter shape that is all the fields that are required to reply/edit the target.

Then, you introduce two new APIs:

- given an `EventTimelineItem`, build this edit/reply parameter from it. This would be used every time there's a corresponding timeline item already existing in the timeline
- given an `EventId`, fetch this data and gather it inside an edit/reply parameter. This woudl be used every time the timeline item couldn't be found from the item in the timeline.

(Alternatively: you could also trigger back-paginations until you find the event in the timeline?)

crates/matrix-sdk-ui/src/timeline/event_item/content/message.rs
Velin92: Made this function not public, and instead created a wrapper function around it that hides the generic usage (since RoomDataProvider would be more private than the actual function in that case) for hide this one. Also wrote a small piece of documentation for it.

crates/matrix-sdk-ui/src/timeline/mod.rs
Velin92: Okay for now I have removed this code, and see if I can directly change the behaviour of the edit and the reply functions

bnjbvr (changes requested):
Hey! sorry for the drive by but this requires tests, since it introduces new APIs, and there's not a single test as far as I can see. Thanks in advance for making our SDK safe and more robust :innocent:

Velin92:
> Hey! sorry for the drive by but this requires tests, since it introduces new APIs, and there's not a single test as far as I can see. Thanks in advance for making our SDK safe and more robust 😇

Yeah I'll now focus on writing some tests for the new APIs
Also seems that the issue we saw with the edit, is an existing issue tied to how the timeline updates when paginating, since I was able to reproduce it on both the old and the new API, and also on Android, also wrote an integration test for it and it does pass the test without any issue.

More info here:
https://github.com/element-hq/element-x-ios/issues/2884

crates/matrix-sdk-ui/src/timeline/mod.rs
```diff
@@ -293,6 +314,68 @@ impl Timeline {
...
+        };
+
+        let raw_sync_event: Raw<AnySyncTimelineEvent> = event.event.cast();
+        let Ok(sync_event) = raw_sync_event.deserialize_as::<AnySyncTimelineEvent>() else {
+            return Err(UnsupportedReplyItem::DESERIALIZATION_ERROR);
+        };
+
+        // Likely needs to be changed or renamed?
+        let Some(timeline_item_content) =
+            TimelineItemContent::from_latest_event_content(sync_event.clone())
```
Velin92, line 359: I think we can reuse this function, but maybe is worth changing it's name to `from_any_sync_event` ?

crates/matrix-sdk-ui/src/timeline/mod.rs
```diff
@@ -293,6 +314,68 @@ impl Timeline {
...
+    /// # Arguments
+    ///
+    /// * `content` - The content of the reply
+    ///
+    /// * `event_id` - The event id of the item you want to reply to
+    ///
+    /// * `forward_thread` - Usually `Yes`, unless you explicitly want to the
+    ///   reply to show up in the main timeline even though the `reply_item` is
+    ///   part of a thread
+    pub async fn send_reply_with_event(
```
bnjbvr, line 336: Same high-level comment for this function and for the edit part.

Instead of having two methods:

- one to send using an event id
- one to send using an event timeline item

Rather, I was thinking about the following:

- one method to retrieve the `ReplyInfo` from an event id
- one method to retrieve the `ReplyInfo` from an event timeline item (would return an Option, if it's been found)
- and the `send_reply` method you already have

This would make the API leaner, and then only the FFI layer could use the two first methods in combination to get a `ReplyInfo`, and then use it in `send_reply`.

And same for the edit.

crates/matrix-sdk-ui/tests/integration/timeline/edit.rs
```diff
@@ -224,6 +227,95 @@ async fn test_send_edit() {
...
+
+    // Since we assume we can't use the timeline item directly in this use case, the
+    // API will fetch the event from the server directly so we need to mock the
+    // response.
+    Mock::given(method("GET"))
+        .and(path_regex(r"^/_matrix/client/r0/rooms/.*/event/"))
+        .and(header("authorization", "Bearer 1234"))
+        .respond_with(
+            ResponseTemplate::new(200)
+                .set_body_json(hello_world_item.latest_json().unwrap().json()),
```
bnjbvr, line 291: Can you put the raw JSON as a raw value instead of reusing something we don't see / have control over?

crates/matrix-sdk-ui/tests/integration/timeline/replies.rs
```diff
@@ -354,6 +354,127 @@ async fn test_send_reply() {
...
+        .await;
+
+    // Since we assume we can't use the timeline item directly in this use case, the
+    // API will fetch the event from the server directly so we need to mock the
+    // response.
+    Mock::given(method("GET"))
+        .and(path_regex(r"^/_matrix/client/r0/rooms/.*/event/"))
+        .and(header("authorization", "Bearer 1234"))
+        .respond_with(
+            ResponseTemplate::new(200).set_body_json(event_from_bob.latest_json().unwrap().json()),
```
bnjbvr, line 430: here too

testing/matrix-sdk-integration-testing/src/tests/sliding_sync/room.rs
```diff
@@ -1126,6 +1128,207 @@ async fn test_room_preview() -> Result<()> {
     Ok(())
 }
 
+#[tokio::test]
+async fn test_edit_unpaginated_item_after_resetting_timeline() -> Result<()> {
```
bnjbvr, line 1132: Can you extract out those two integration tests, and post a new PR with those, as they seem unrelated to the feature at hand, please?

Velin92: well they are actually also a way to test the feature, since they can address both the issue we currently have and an integration test implementation of editing an item that is not in the timeline

crates/matrix-sdk-ui/src/timeline/mod.rs
Velin92: Okay I see this means I need to expose `ReplyInfo` and `EditInfo` also through FFI right?

bnjbvr: `ReplyInfo` and `EditInfo` would need to be exposed outside of the SDK UI crate, but then the FFI could make use of those internally to provide the final functionality the apps require.

testing/matrix-sdk-integration-testing/src/tests/sliding_sync/room.rs
bnjbvr: If an integration test doesn't trigger a failure you observe with new functions introduced, let's try to keep the PR lean and remove it from this PR and open another instead; it'll make this PR easier to review, and investigating why the test doesn't pass easier, without blocking one with the other.

crates/matrix-sdk-ui/src/timeline/mod.rs
Velin92: @bnjbvr what about this?

bnjbvr (changes requested):

crates/matrix-sdk-ui/tests/integration/timeline/edit.rs
```diff
@@ -224,6 +225,100 @@ async fn test_send_edit() {
...
+        timeline.get_edit_info_from_event_id(hello_world_item.event_id().unwrap()).await.unwrap();
+    timeline
+        .edit(RoomMessageEventContentWithoutRelation::text_plain("Hello, Room!"), edit_info)
+        .await
+        .unwrap();
+
+    // Since verifying the content would mean mocking the sliding sync response with
+    // what we are already expecting, because this test would require to paginate
+    // again the timeline, testing the content change would not be meaningful.
+    // Use an integration test for the full case.
```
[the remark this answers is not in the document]
Velin92: well it's at least checking for the correct API to be called thanks to the verify in the mock, so at least we know that is doing what is supposed to do, which is probably better than no test at all

crates/matrix-sdk-ui/src/timeline/mod.rs
```diff
@@ -300,73 +326,133 @@ impl Timeline {
...
+        let Ok(sync_event) = raw_sync_event.deserialize_as::<AnySyncTimelineEvent>() else {
+            return Err(UnsupportedReplyItem::DESERIALIZATION_ERROR);
+        };
+
+        // Likely needs to be changed or renamed?
+        let Some(timeline_item_content) =
+            TimelineItemContent::from_latest_event_content(sync_event.clone())
+        else {
+            return Err(UnsupportedReplyItem::MISSING_CONTENT);
+        };
```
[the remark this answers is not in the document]
Velin92: there is any other way to get the timeline content from here?

crates/matrix-sdk-base/src/store/traits.rs
```diff
@@ -808,6 +809,37 @@ pub enum StateStoreDataValue {
 
     /// A list of recently visited room identifiers for the current user
     RecentlyVisitedRooms(Vec<String>),
+
+    /// A composer draft for the room.
+    /// /// To learn more, see [`ComposerDraft`].
+    ///
+    /// [`ComposerDraft`]: self::ComposerDraft
```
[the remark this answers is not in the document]
Velin92: yeah it did not complain, but I'll change it!

crates/matrix-sdk-ui/src/timeline/event_item/content/message.rs
```diff
@@ -237,6 +237,13 @@ impl InReplyToDetails {
 
         InReplyToDetails { event_id, event: TimelineDetails::from_initial_value(event) }
     }
+
+    pub fn new_with_timeline_details(
+        event_id: OwnedEventId,
+        event: TimelineDetails<Box<RepliedToEvent>>,
```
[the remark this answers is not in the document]
Velin92: there is a chance it might also be an error in case there is a failure.

bindings/matrix-sdk-ffi/src/timeline/mod.rs
Velin92: I need it to be able to be able to import it here:
https://github.com/matrix-org/matrix-rust-sdk/blob/4b88fe1078f6905cb9b67c8851478b035510804f/bindings/matrix-sdk-ffi/src/room.rs#L38-L40

bnjbvr:
Hey @Velin92, the PR has merge conflicts, can you solve them before I do a re-review, please? Thanks!

bnjbvr (changes requested):
(see above comment)

bnjbvr (changes requested):
This PR is gigantic, and it's still broken; see my last comment. I'm pretty sure I've requested *not* using this particular function multiple times, and you've ignored my feedback.

Also this PR is too big, and the commit history doesn't make sense, and it doesn't help making a proper review. Let's break this PR into smaller chunks, it's still doing invalid stuff, and I'd like to see some bits merged first.

Can you split the `matrix-sdk-base` changes into their own PR, with their own testing (even if the feature isn't used at the timeline layer yet), please?

crates/matrix-sdk-ui/src/timeline/event_item/content/mod.rs
```diff
@@ -117,12 +117,10 @@ pub enum TimelineItemContent {
...
-    /// message preview, extract its contents and wrap it as a
-    /// `TimelineItemContent`.
-    pub(crate) fn from_latest_event_content(
-        event: AnySyncTimelineEvent,
-    ) -> Option<TimelineItemContent> {
+    /// If the supplied event is suitable to be used as a message preview
+    /// like reply previews or latest event items in the room list,
+    /// extract its contents and wrap it as a `TimelineItemContent`.
+    pub(crate) fn from_any_sync_event(event: AnySyncTimelineEvent) -> Option<TimelineItemContent> {
         match is_suitable_for_latest_event(&event) {
```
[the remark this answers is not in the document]
Velin92: Async in our DMs we agreed that it was possible to reuse the `from_latest_event` function to synthethise the event, in fact you even said it was okay to do the renaming. I even asked if in case it was not the correct way to address this function here: https://github.com/matrix-org/matrix-rust-sdk/pull/3439#discussion_r1622395128 what was the correct way to solve this issue, since I myself was unsure if that was okay,
Your suggestion now makes sense, but would have been nice to have this been pointed way before without the need of saying that I have ignored feedback, since I actually asked for help, and flagged my concern about reusing this function

PR #3442, opened 2024-05-22
crypto: emit no `identities_stream` items on no-op changes

richvdh:
Fixes: https://github.com/element-hq/element-web/issues/27165.

Currently, whenever we receive a `/keys/query` response, we emit items from `identities_stream_raw` and `devices_stream` for any users listed in the response, even if there was no actual change to the devices. This PR changes behaviour so that items are only emitted if there are actual changes.


Hywan (approved):
I like the commits and the new code. Thanks for your work!

Issue #3447, opened 2024-05-23
When using the in-memory crypto store, keys restored from backup are re-uploaded to backup


richvdh:
`BackupMachine::backup()` returns a list of message keys which need backing up to the server.

When we restore a key from backup, it is imported via `BackupMachine::import_backed_up_room_keys()`. The intention is that such keys are *not* returned by a later call to `BackupMachine::backup()` (since, obviously, they are already backed up).

However, this has broken, at least for `MemoryStore`; the regression happened in #3320. The problem is that, as of that PR, `MemoryStore` keeps a separate record of "which backup version was this session backed up to", which it uses to populate the results for `BackupMachine::backup()`; and `BackupMachine::import_backed_up_room_keys()` does not populate that field.

PR #3448, opened 2024-05-23
Fix backed-up keys being re-backed-up

richvdh:
This PR aims to fix https://github.com/matrix-org/matrix-rust-sdk/issues/3447.

The first thing to do is to make sure that a backup version is stored for any keys that are imported from backup. As a quick fix, which doesn't break backwards compatibility, we therefore update `BackupMachine::import_backed_up_room_keys` to pass through the current backup version to the store.

However, that is very much a bodge: we have no guarantee that is the correct version. A *better* fix is to have the application pass through the version that it got the keys from. We therefore deprecate the inherently-broken `BackupMachine::import_backed_up_room_keys`, and chase the changes through to the SDK and FFI layers.


poljar:
> Should be reviewable commit-by-commit. [I'd also be open to splitting this into two PRs: one for a quick fix, and one for the deprecation stuff. LMK what you think]

No need, I think it's manageable.

crates/matrix-sdk-crypto/src/backups/mod.rs
```diff
@@ -619,7 +619,15 @@ impl BackupMachine {
             }
         }
 
-        self.store.import_room_keys(decrypted_room_keys, true, progress_listener).await
+        // This method is a bit flawed: we have no real idea which backup version these
```
poljar, line 623: Could you annotate this with a `TODO: ` or `FIXME: `, so we can more easily find this fact.

crates/matrix-sdk-crypto/src/machine.rs
```diff
@@ -1823,7 +1823,11 @@ impl OlmMachine {
         from_backup: bool,
         progress_listener: impl Fn(usize, usize),
     ) -> StoreResult<RoomKeyImportResult> {
-        self.store().import_room_keys(exported_keys, from_backup, progress_listener).await
+        let backup_version =
```
poljar, line 1826: I think we can just remove this method. We had a `0.7.x` release.

poljar (approved):
Makes sense, I disagree a bit on the deprecations but don't care enough to argue.

I left some small nits, feel free to merge without further review.

crates/matrix-sdk-crypto/src/backups/mod.rs
richvdh: I can, but this is the main reason I deprecated the method. The real fix is to remove it.

crates/matrix-sdk-crypto/src/machine.rs
richvdh: Even better! Will remove it.

PR #3449, opened 2024-05-23
Don't return a Results if some functions can't fail

poljar:
The function to run the E2EE initialization tasks and to enable the cross-process lock in a deferred manner can't actually fail, they still return a `Result` type.

This PR makes removes the `Result` from those functions. I also refactored the enabling of the cross-process lock into a separate method.

No functional changes due to these patches.

### DIFF
--- a/crates/matrix-sdk/src/oidc/mod.rs
+++ b/crates/matrix-sdk/src/oidc/mod.rs
@@ -300,17 +300,17 @@ impl Oidc {
     /// olm machine has been initialized.
     ///
     /// Must be called after `set_session_meta`.
-    async fn deferred_enable_cross_process_refresh_lock(&self) -> Result<()> {
+    async fn deferred_enable_cross_process_refresh_lock(&self) {
         let deferred_init_lock = self.ctx().deferred_cross_process_lock_init.lock().await;
 
         // Don't `take()` the value, so that subsequent calls to
         // `enable_cross_process_refresh_lock` will keep on failing if we've enabled the
         // lock at least once.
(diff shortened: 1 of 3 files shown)


crates/matrix-sdk/src/oidc/mod.rs
```diff
@@ -300,17 +300,17 @@ impl Oidc {
...
         // Don't `take()` the value, so that subsequent calls to
         // `enable_cross_process_refresh_lock` will keep on failing if we've enabled the
         // lock at least once.
         let Some(lock_value) = deferred_init_lock.as_ref() else {
-            return Ok(());
+            return;
         };
 
-        // FIXME We shouldn't be using the crypto store for that! see also https://github.com/matrix-org/matrix-rust-sdk/issues/2472
+        // FIXME: We shouldn't be using the crypto store for that! see also https://github.com/matrix-org/matrix-rust-sdk/issues/2472
```
bnjbvr, line 313: :monocle_face:

bnjbvr (approved):
Love it.

crates/matrix-sdk/src/oidc/mod.rs
poljar: You're still not using [trouble](https://github.com/folke/trouble.nvim) to highlight and quickly list those?

image

bnjbvr: uhuh i'm getting into trouble right now!

poljar: :drum:

PR #3452, opened 2024-05-23
timeline: indicate if we've hit the start of the timeline in the live back-pagination status

bnjbvr:
This should help with the integration of the back-pagination status, when we have automatic back-pagination in the background.


poljar (approved):
Makes sense and looks good.

PR #3453, opened 2024-05-23
feat(crypto): If available, sign the new device keys using the cross-signing keys

poljar:


andybalaam (approved):
This is great, thanks!

PR #2825, opened 2023-11-07
Add get_element_call_required_permissions

toger5:
Adding a function to get the required `WidgetCapabilites`

This is supposed to be used with the PermissionProvider.

Its still a draft since its not yet tested.

Signed-off-by:

PR #3455, opened 2024-05-23
ffi: Fix `get_element_call_required_permissions`

stefanceriu:
-fix what permissions `get_element_call_required_permissions` returns and have them match what Element Call actually expects
- this got implemented in https://github.com/matrix-org/matrix-rust-sdk/pull/2825 but never actually worked in the app, resulting in ElementCall getting stuck on a `Loading` screen
- tweaked the permissions list by running ElementCall and comparing it to the capabilities returned in `acquireCapabilities(capabilities:)`
- check all of this with @toger5, which also tells me that `org.matrix.msc3401.call` is deprecated and should not be required although ElementCall doesn't work without it.  He will clean up the web side and then come back to remove this as well in a later PR.

### DIFF
--- a/bindings/matrix-sdk-ffi/src/widget.rs
+++ b/bindings/matrix-sdk-ffi/src/widget.rs
@@ -260,11 +260,12 @@ pub fn new_virtual_element_call_widget(
 /// but should only be done as temporal workarounds until this function is
 /// adjusted
 #[uniffi::export]
-pub fn get_element_call_required_permissions() -> WidgetCapabilities {
+pub fn get_element_call_required_permissions(own_user_id: String) -> WidgetCapabilities {
     use ruma::events::StateEventType;
 
     WidgetCapabilities {
         read: vec![
+            WidgetEventFilter::StateWithType { event_type: "org.matrix.msc3401.call".to_owned() },
             WidgetEventFilter::StateWithType { event_type: StateEventType::CallMember.to_string() },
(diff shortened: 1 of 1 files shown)


bnjbvr (approved):
How can we test this, other than running it on the apps? This seems quite frail; it would make sense in my opinion to take all the ElementCall related call, put it under a new crate umbrella so we can test it properly, and then expose that to the FFI, maybe?

stefanceriu:
> How can we test this, other than running it on the apps? 

I'm not entirely sure to be honest as in the end you need to run the actual ElementCall web app to validate things actually work. I reckon an integration UI test in the final client would be the simplest way.

toger5:
I think this particular case is very much a Element Call testing case and not rust sdk.
The original implementation is technically the more correct one (it leaves out a capability which is not needed by element call in theory but element call apparently also does not laod when it is missing.
So this is very much a stop gap until we found the issue on ElementCall. It is a security related issue/change.

The best test approach imo is to strictly define the permissions we want the widget to have and then use test that both projects are compatible with our defined capabilities.
We would need a test in element call that it successfully loads with the correct capabilities. (The rust sdk just needs a unit test that it only allows the defined capabilites.)

Going forward we need to eventually use resources on the voip team to investigate why ElementCall wont load if the capability is missing. then we can remove the `org.matrix.msc3401.call` part of the pr again.

stefanceriu:
Alrighty, I think you can go ahead and merge this @bnjbvr. Thaaank you!

PR #3457, opened 2024-05-23
feat(crypto): Allow the explicit upload of the device keys

poljar:

### DIFF
--- a/crates/matrix-sdk-base/src/client.rs
+++ b/crates/matrix-sdk-base/src/client.rs
@@ -226,6 +226,7 @@ impl BaseClient {
             &session_meta.user_id,
             &session_meta.device_id,
             self.crypto_store.clone(),
+            None,
         )
         .await
         .map_err(OlmError::from)?;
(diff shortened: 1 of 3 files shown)


bnjbvr (approved):
lgtm, is it intended that the check for `account.shared()` has been removed? if so, can you comment here on github why, please?

poljar:
> lgtm, is it intended that the check for `account.shared()` has been removed? if so, can you comment here on github why, please?

It is, `keys_for_upload()` does a similar check, it returns `None` if we don't have any keys, device keys or one-time keys, to upload.

It doesn't hurt to send out the request even if the device keys are already uploaded, it will just upload some one-time keys.

Furthermore, later on I think I'll want to remove the device key uploading from the cross-signing bootstrap method and tell users to use the explicit device key uploading method if they need this.

In the main crate on the other hand, I plan to go away from the current flow we have, where device keys are uploaded once the outgoing crypto requests are sent, instead I want the device key uploading to happen at login time.

This will already be happening for the QR login case, and I think it makes great sense for the other login flows.

PR #3459, opened 2024-05-23
ci: Split codecov uploads into a separate workflow

jplatte:
… which runs in the context of the main repo even for PRs from forks, and can be retried individually without rerunning coverage collection.

I saw you're still suffering from codecov's problems, so I decided to take the time required to look up the most recent suggested workaround (most up-to-date upstream issue discussing this stuff is https://github.com/codecov/feedback/issues/301 and contains links to the workflows I copied here).

Hope it works. If not, maybe you can get it to work. Can't promise that I'll put more time into this.

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -1,4 +1,4 @@
-name: Code coverage
+name: Code Coverage
 
 on:
   push:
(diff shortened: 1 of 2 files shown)


jplatte:
Hm, new workflow file doesn't trigger, prob. will only start doing that after it's part of `main`?

bnjbvr:
I'm curious as to why it didn't trigger in this branch; shouldn't it? I'm pretty sure I've seen in the past that new CI steps would run in the PR that introduced them…

bnjbvr (approved):
Let's try it out!

PR #3460, opened 2024-05-24
feat(sdk): Mutation-based testing and property-based testing for `LinkedChunk`

Hywan:
This patch is the result of:

* Running [`cargo mutants`](https://mutants.rs/) on `matrix_sdk::event_cache::linked_chunk`,
* Having fun with [`cargo fuzz`](https://rust-fuzz.github.io/book/cargo-fuzz.html), then replacing it with [`proptest`](https://proptest-rs.github.io/proptest/proptest/index.html).

Mutation-based testing has revealed that one test was missing, it's been addressed by https://github.com/matrix-org/matrix-rust-sdk/commit/50f9cd6b4536389d65d7b51fb6e5a940c0346eb1.

Property-based testing has revealed nothing except that the test I wrote was raising a bug, addressed by https://github.com/matrix-org/matrix-rust-sdk/pull/3460/commits/ed41d49fe14350edb746278d52a39a884debb29b.

The property-based test is present in this patch too.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -554,6 +554,21 @@ dependencies = [
 ]
 
 [[package]]
+name = "bit-set"
+version = "0.5.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "0700ddab506f33b20a03b13996eccd309a48e5ff77d0d95926aa0210fb4e95f1"
+dependencies = [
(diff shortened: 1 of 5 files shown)


poljar (approved):
Eyy, very nice proptest you got there going.

PR #3441, opened 2024-05-22
feat(base+ffi+ui): `SlidingSync` and `RoomList` support heroes

Hywan:
This patch does the following things:

1. It updates Ruma to the latest revision at the time of writing, it adds `include_heroes` on `RoomSubscription`
2. It updates `matrix-sdk-ffi` to provide the `RoomSubscription::include_heroes` field,
3. It updates `matrix-sdk-base` so that SlidingSync consumes `heroes` from the response and update the `RoomSummary` accordingly.
4. It updates Ruma once again, this time on my own fork (waiting for the PR to be merged), which includes a bug fix (a hero has a `displayname`, not a `name` :-/), and adds `include_heroes` on the `SlidingSyncList`
5. It updates `matrix-sdk` to provide `include_heroes` on `SlidingSyncListBuilder` and inside the sticky parameters of a `SlidingSyncList`
6. It updates `matrix-sdk-ui` to enable `include_heroes` by default for `all_rooms` and `visible_rooms`.

---

* Close https://github.com/matrix-org/matrix-rust-sdk/issues/2702

Issue #3272, opened 2024-03-24
[EPIC] Spring cleaning


poljar:
# Spring cleaning

![image](https://github.com/matrix-org/matrix-rust-sdk/assets/552026/913143d2-80fc-4e0a-beef-031ec7a15c90)

## Motivation

As the seasons change and the warmth of the sun begins to illuminate our surroundings, we're often confronted with the less appealing aspects of the thaw: dust accumulating, cobwebs in every nook, neglected electronic parts under desks, and remnants of recent woodworking projects scattered about. It's a scenario many of us are familiar with. Similarly, our SDK has been the focus of relentless feature development, but now it's crucial to take a step back and see which puzzle pieces don't quite fit.

## The list

This list is unordered and contains short descriptions of issues that we should tackle. Some of the items on the list are known bugs and have a separate issue already, some of them do not. If you want to work on one of the items, and there isn't a separate issue already, please open a separate issue and link it in the list.

If it's unclear what one of the items means, please just ask and we'll reword it, or open a dedicated issue for the item.

### General
- [ ] Bring back git-cliff with a vengeance, it probably makes sense to test this out in vodozemac.
- [ ] Determine a release schedule with a 1-2 week feature freeze period.
- [ ] Switch to using [similar_asserts](https://github.com/mitsuhiko/similar-asserts), it provides much nicer diffs if an assert fails.
- [ ] https://github.com/matrix-org/matrix-rust-sdk/issues/2937
- [ ] Get rid of the testing feature

### Main crate
- [ ] The calculated name is stored in the `m.room.name` field of the `RoomInfo`
- [ ] The avatar URL for DMs is calculated by the SS proxy and put into the `m.room.avatar` field of the `RoomInfo`
- [ ] AuthenticationService duplicates the ClientBuilder API without a builder
- [ ] The `BaseRoomInfo` since some time uses something called `MinimalStateEvent`,
   this event isn't that minimal since it's keeping the whole content. Since it
   is keeping the whole content, and additionally is using Ruma types for the
   content, those may fail to be reserialized.
- [ ] People hack around having no `MemberList`, an equivalent to the `RoomList` but for members.
- [ ] Deprecate the base crate and merge the `BaseClient` into the main crate and `Client`.
- [ ] Check that we don't handle state events in the timeline portion of the sliding sync responses.
- [ ] https://github.com/matrix-org/matrix-rust-sdk/issues/1254

### Crypto crate
[...]

PR #3461, opened 2024-05-24
room names: better make use of the hero names for locally computing a room display name

bnjbvr:
This is #3441 but since I've added many commits while taking it over, reopening a new PR for it.

This makes it so that we prefer using the hero names if we receive those from the sliding sync, rather than retrieving the names from the room members that are hero, or retrieving the room members from the first N members.

I've also updated the very unfortunate edge case where we have 0 heroes to display "n people" instead of ", and n others.".

### DIFF
--- a/crates/matrix-sdk-base/src/rooms/normal.rs
+++ b/crates/matrix-sdk-base/src/rooms/normal.rs
@@ -116,8 +116,11 @@ pub struct RoomSummary {
     ///
     /// This was called `heroes` and contained raw `String`s of the `UserId`
     /// before; changing the field's name helped with avoiding a migration.
-    #[serde(default)]
+    #[serde(default, skip_serializing_if = "Vec::is_empty")]
     pub(crate) heroes_user_ids: Vec<OwnedUserId>,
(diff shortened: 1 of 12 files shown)


crates/matrix-sdk-base/src/rooms/normal.rs
```diff
@@ -628,46 +627,44 @@ impl Room {
...
-                            return None;
-                        }
-                        self.get_member(&user_id).await.transpose()
-                    })
-                    .collect()
-                    .await;
-
-                let members: StoreResult<Vec<_>> = members.into_iter().collect();
+        let (heroes, guessed_num_members): (Vec<String>, _) = if !summary.heroes_names.is_empty() {
+            // Straight-forward path: pass through the heroes names, don't give a guess of
```
Hywan, line 631:
```diff
- Straight-forward
+ Straightforward
```

Hywan (approved):
I think we are good with these patches. Thanks for digging into this!

Hywan:
Address part of https://github.com/matrix-org/matrix-rust-sdk/issues/3272.

Issue #3132, opened 2024-02-15
The SQLite store serializes user identities and devices using rmp_serde instead of serde_json


poljar:
User identities and devices are internally using a `serde_json::Value`, this means that, if the `Value` isn't empty, `rmp_serde` will fail to with a cryptic error message.

The interesting part is, as far as my memory serves me, that serialization will succeed, but deserialization will fail. This will lead to UTDs for such devices and incorrect trust chain calculation for the user identities.

Currently this isn't a problem since the `Value`, which is used to gather unknown additional data, is always empty.

The offending lines can be found here:

https://github.com/matrix-org/matrix-rust-sdk/blob/315a29f5684cea68cef61775e27127fc7387cc2b/crates/matrix-sdk-sqlite/src/crypto_store.rs#L804

https://github.com/matrix-org/matrix-rust-sdk/blob/315a29f5684cea68cef61775e27127fc7387cc2b/crates/matrix-sdk-sqlite/src/crypto_store.rs#L816

This issue is similar to this commit c73aeef2.

While the fix is trivial, the problematic part here will be the data migration.

poljar:
Hmm, I don't think it is. Only `Option` to handle the custom stuff that might appear in the `DeviceKeys` so signature verification continues to work.

uhoreg:
I modified a test to check if this is working with this patch:

```diff
diff --git a/crates/matrix-sdk-crypto/src/store/integration_tests.rs b/crates/matrix-sdk-crypto/src/store/integration_tests.rs
index 9373300ef..6caa3c2cd 100644
--- a/crates/matrix-sdk-crypto/src/store/integration_tests.rs
+++ b/crates/matrix-sdk-crypto/src/store/integration_tests.rs
@@ -18,6 +18,7 @@ macro_rules! cryptostore_integration_tests {
                 user_id, DeviceId, JsOption, OwnedDeviceId, OwnedUserId, RoomId, TransactionId, UserId
             };
             use serde_json::value::to_raw_value;
+            use serde_json::json;
             use $crate::{
                 olm::{
                     Curve25519PublicKey, InboundGroupSession, OlmMessageHash,
@@ -40,9 +41,10 @@ macro_rules! cryptostore_integration_tests {
                         secret_send::SecretSendContent,
                         ToDeviceEvent,
                     },
+                    DeviceKeys,
                     EventEncryptionAlgorithm,
                 },
[...]

PR #3464, opened 2024-05-24
Add tests for unknown properties in device keys

uhoreg:
Add tests to ensure that unknown properties in device keys are handled correctly.

One test was previously discussed at https://github.com/matrix-org/matrix-rust-sdk/issues/3132#issuecomment-2002060930  The other test ensures that the /keys/query response handler doesn't freak out about unknown properties.


poljar (approved):
Thanks, indeed it works.

PR #3466, opened 2024-05-24
Support to log in using a QR code

poljar:
This implements part of [MSC4108](https://github.com/matrix-org/matrix-spec-proposals/pull/4108). We're only supporting the case where new device scans the QR code and gets logged in by the existing device.

Opening as a draft as I want to reword the commit messages a bit, add some more documentation (mainly to insert some jokes about rendezvous), and possibly one or two more tests.

Best reviewed commit by commit, although there is one giant commit. Reviewer please have mercy on me.

### DIFF
--- a/crates/matrix-sdk/Cargo.toml
+++ b/crates/matrix-sdk/Cargo.toml
@@ -45,6 +45,7 @@ uniffi = ["dep:uniffi", "matrix-sdk-base/uniffi"]
 
 experimental-oidc = [
     "ruma/unstable-msc2967",
+    "ruma/unstable-msc4108",
     "dep:chrono",
     "dep:http_old",
     "dep:language-tags",
@@ -72,6 +73,7 @@ async-channel = "2.2.1"
 async-stream = { workspace = true }
 async-trait = { workspace = true }
 axum = { version = "0.7.4", optional = true }
+base64 = { workspace = true }
 bytes = "1.1.0"
 bytesize = "1.1"
 chrono = { version = "0.4.23", optional = true }
@@ -108,9 +110,10 @@ tokio-stream = { workspace = true, features = ["sync"] }
 tower = { version = "0.4.13", features = ["make"], optional = true }
 tracing = { workspace = true, features = ["attributes"] }
 uniffi = { workspace = true, optional = true }
-url = { workspace = true }
+url = { workspace = true, features = ["serde"] }
 urlencoding = "2.1.3"
 uuid = { version = "1.4.1", features = ["serde", "v4"], optional = true }
+vodozemac = { workspace = true }
 zeroize = { workspace = true }
 
 [dependencies.image]
@@ -126,6 +129,7 @@ tokio = { workspace = true, features = ["macros"] }
 
 [target.'cfg(not(target_arch = "wasm32"))'.dependencies]
 backoff = { version = "0.4.0", features = ["tokio"] }
+openidconnect = { git = "https://github.com/poljar/openidconnect-rs", rev = "c7e1dc31b83dd7559125984bfd36b9c0f191585e" }
 # only activate reqwest's stream feature on non-wasm, the wasm part seems to not
 # support *sending* streams, which makes it useless for us.
 reqwest = { workspace = true, features = ["stream"] }
@@ -144,6 +148,7 @@ matrix-sdk-base = { workspace = true, features = ["testing"] }
 matrix-sdk-test = { workspace = true }
 once_cell = { workspace = true }
 serde_urlencoded = "0.7.1"
+similar-asserts = "1.5.0"
 stream_assert = { workspace = true }
 tracing-subscriber = { version = "0.3.11", features = ["env-filter"] }
 
--- a/crates/matrix-sdk/src/authentication/qrcode/login.rs
+++ b/crates/matrix-sdk/src/authentication/qrcode/login.rs
@@ -1,0 +1,925 @@
+// Copyright 2024 The Matrix.org Foundation C.I.C.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use std::future::IntoFuture;
+
+use eyeball::SharedObservable;
+use futures_core::Stream;
+use mas_oidc_client::types::{
+    client_credentials::ClientCredentials, registration::VerifiedClientMetadata,
+};
+use matrix_sdk_base::{
+    boxed_into_future,
+    crypto::types::qr_login::{QrCodeData, QrCodeMode},
+    SessionMeta,
+};
+use openidconnect::DeviceCodeErrorResponseType;
+use ruma::OwnedDeviceId;
+use tracing::trace;
+use vodozemac::ecies::CheckCode;
+
+use super::{
+    messages::LoginFailureReason, oidc_client::OidcClient, DeviceAuhorizationOidcError,
+    SecureChannelError,
+};
+#[cfg(doc)]
+use crate::oidc::Oidc;
+use crate::{
+    authentication::qrcode::{
+        messages::QrAuthMessage, secure_channel::EstablishedSecureChannel, QRCodeLoginError,
+    },
+    Client,
+};
+
+async fn send_unexpected_message_error(
+    channel: &mut EstablishedSecureChannel,
+) -> Result<(), SecureChannelError> {
+    channel
+        .send_json(QrAuthMessage::LoginFailure {
+            reason: LoginFailureReason::UnexpectedMessageReceived,
+            homeserver: None,
+        })
+        .await
+}
+
+/// Type telling us about the progress of the QR code login.
+#[derive(Clone, Debug, Default)]
+pub enum LoginProgress {
+    /// We're just starting up, this is the default and initial state.
+    #[default]
+    Starting,
+    /// We have established the secure channel, but we need to let the other
+    /// side know about the [`CheckCode`] so they can verify that the secure
+    /// channel is indeed secure.
+    EstablishingSecureChannel {
+        /// The check code we need to, out of band, send to the other device.
+        check_code: CheckCode,
+    },
+    /// We're waiting for the OIDC provider to give us the access token. This
+    /// will only happen if the other device allows the OIDC provider to so.
+    WaitingForToken {
+        /// The user code the OIDC provider has given us, the OIDC provider
+        /// might ask the other device to enter this code.
+        user_code: String,
+    },
+    /// The login process has completed.
+    Done,
+}
+
+/// Named future for the [`Oidc::login_with_qr_code()`] method.
+#[derive(Debug)]
+pub struct LoginWithQrCode<'a> {
+    client: &'a Client,
+    client_metadata: VerifiedClientMetadata,
+    qr_code_data: &'a QrCodeData,
+    state: SharedObservable<LoginProgress>,
+}
+
+impl<'a> LoginWithQrCode<'a> {
+    /// Subscribe to the progress of QR code login.
+    ///
+    /// It's usually necessary to subscribe to this to let the existing device
+    /// know about the [`CheckCode`] which is used to verify that the two
+    /// devices are communicating in a secure manner.
+    pub fn subscribe_to_progress(&self) -> impl Stream<Item = LoginProgress> {
+        self.state.subscribe()
+    }
+}
+
+impl<'a> IntoFuture for LoginWithQrCode<'a> {
+    type Output = Result<(), QRCodeLoginError>;
+    boxed_into_future!(extra_bounds: 'a);
+
+    fn into_future(self) -> Self::IntoFuture {
+        Box::pin(async move {
+            // First things first, establish the secure channel. Since we're the one that
+            // scanned the QR code, we're certain that the secure channel is
+            // secure, under the assumption that we didn't scan the wrong QR code.
+            let mut channel = self.establish_secure_channel().await?;
+
+            trace!("Established the secure channel.");
+
+            // The other side isn't yet sure that it's talking to the right device, show
+            // a check code so they can confirm.
+            let check_code = channel.check_code().to_owned();
+            self.state.set(LoginProgress::EstablishingSecureChannel { check_code });
+
+            // Register the client with the OIDC provider.
+            trace!("Registering the client with the OIDC provider.");
+            let oidc_client = self.register_client().await?;
+
+            // We want to use the Curve25519 public key for the device ID, so let's generate
+            // a new vodozemac `Account` now.
+            let account = vodozemac::olm::Account::new();
+            let public_key = account.identity_keys().curve25519;
+            let device_id = public_key;
+
+            // Let's tell the OIDC provider that we want to log in using the device
+            // authorization grant described in [RFC8628](https://datatracker.ietf.org/doc/html/rfc8628).
+            trace!("Requesting device authorization.");
+            let auth_grant_response = oidc_client.request_device_authorization(device_id).await?;
+
+            // Now we need to inform the other device of the login protocols we picked and
+            // the URL they should use to log us in.
+            trace!("Letting the existing device know about the device authorization grant.");
+            let message = QrAuthMessage::authorization_grant_login_protocol(
+                (&auth_grant_response).into(),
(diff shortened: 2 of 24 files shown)


crates/matrix-sdk/Cargo.toml
```diff
@@ -96,6 +98,7 @@ matrix-sdk-sqlite = { workspace = true, optional = true }
 matrix-sdk-test = { workspace = true, optional = true }
 mime = "0.3.16"
 mime2ext = "0.1.52"
+openidconnect = { git = "https://github.com/poljar/openidconnect-rs", rev = "c7e1dc31b83dd7559125984bfd36b9c0f191585e" }
```
Hywan, line 101: Do you have a PR for your fork https://github.com/poljar/openidconnect-rs?

Do you have a PR for your fork https://github.com/poljar/oauth2-rs?

crates/matrix-sdk/src/authentication/qrcode/login.rs
```diff
@@ -0,0 +1,740 @@
...
+    qr_code_data: &'a QrCodeData,
+    state: SharedObservable<LoginProgress>,
+}
+
+impl<'a> LoginWithQrCode<'a> {
+    /// Subscribe to the progress of QR code login.
+    ///
+    /// It's usually necessary to subscribe to this to let the existing device
+    /// know about the [`CheckCode`] which is used to verify that the two
+    /// devices are communicating in a secure manner.
```
Hywan, line 94: When the `Stream` ends, it's because an error happens? How do we handle errors from this `Stream`? I think it's important to explain that in the doc.

crates/matrix-sdk/src/authentication/qrcode/login.rs
```diff
@@ -0,0 +1,740 @@
...
+    pub fn subscribe_to_progress(&self) -> impl Stream<Item = LoginProgress> {
+        self.state.subscribe()
+    }
+}
+
+impl<'a> IntoFuture for LoginWithQrCode<'a> {
+    type Output = Result<(), QRCodeLoginError>;
+    boxed_into_future!(extra_bounds: 'a);
+
+    fn into_future(self) -> Self::IntoFuture {
```
Hywan, line 104: Really clever to use `IntoFuture` here.

crates/matrix-sdk/src/authentication/qrcode/login.rs
Hywan: Read the code further. Errors will be seen in `IntoFuture::into_future`. We get progress from `subscribe_to_progress` but errors from `into_future`, we need to look at 2 “channels”. Can we add a `Error(…)` variant in `LoginProgress`? It means that `into_future` should catch the errors, update the `LoginProgress` accordingly, and “rethrow” the error. That would be lovely for the consumers of `LoginProgress` as they would have only one thing to listen to.

crates/matrix-sdk/src/authentication/qrcode/messages.rs
```diff
@@ -0,0 +1,326 @@
...
+        #[serde(
+            deserialize_with = "deserialize_curve_key",
+            serialize_with = "serialize_curve_key"
+        )]
+        /// The device ID the new device will be using.
+        device_id: Curve25519PublicKey,
+    },
+
+    /// Message declaring that the protocol in the previous `m.login.protocol`
+    /// message was accepted. Sent by the extising device.
```
Hywan, line 61:
```diff
- extising device
+ existing device
```

crates/matrix-sdk/src/authentication/qrcode/messages.rs
```diff
@@ -0,0 +1,326 @@
...
+    #[serde(rename = "m.login.protocol_accepted")]
+    LoginProtocolAccepted {},
```
Hywan, line 63:
```diff
- LoginProtocolAccepted {},
+ LoginProtocolAccepted,
```

?

crates/matrix-sdk/src/authentication/qrcode/messages.rs
```diff
@@ -0,0 +1,326 @@
...
+
+    /// Message that informs the existing device that it successfully obtained
+    /// an access token from the OIDC provider. Sent by the new device.
+    #[serde(rename = "m.login.success")]
+    LoginSuccess {},
```
Hywan, line 68:
```diff
- LoginSuccess {},
+ LoginSuccess,
```

?

(and so on for the other variants of this enum)

crates/matrix-sdk/src/authentication/qrcode/messages.rs
```diff
@@ -0,0 +1,326 @@
...
+        assert_let!(
+            QrAuthMessage::LoginSecrets(SecretsBundle { cross_signing, backup }) = &message
+        );
+        assert_eq!(cross_signing.master_key, "rTtSv67XGS6k/rg6/yTG/m573cyFTPFRqluFhQY+hSw");
+        assert_eq!(cross_signing.self_signing_key, "4jbPt7jh5D2iyM4U+3IDa+WthgJB87IQN1ATdkau+xk");
+        assert_eq!(cross_signing.user_signing_key, "YkFKtkjcsTxF6UAzIIG/l6Nog/G2RigCRfWj3cjNWeM");
+
+        assert_let!(Some(BackupSecrets::MegolmBackupV1Curve25519AesSha2(backup)) = backup);
+        assert_eq!(backup.backup_version, "2");
+        assert_eq!(&backup.key.to_base64(), "AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA");
```
Hywan, line 321: AAAAAAAaaaaaaa… *boom*”

— “Are you alright?”

crates/matrix-sdk/src/authentication/qrcode/mod.rs
```diff
@@ -0,0 +1,216 @@
...
+/// code.
+#[derive(Debug, Error)]
+#[cfg_attr(feature = "uniffi", derive(uniffi::Error), uniffi(flat_error))]
+pub enum QRCodeLoginError {
+    /// An error happened while we were communicating with the OIDC provider.
+    #[error(transparent)]
+    Oidc(#[from] DeviceAuhorizationOidcError),
+
+    /// The other device has signaled to us that the login has failed.
+    #[error("The login failed, reason: {reason}.")]
```
Hywan, line 64: I think it's better to avoid `.` at the end of the error message. Especially if the reason already contains a `.`.

Alternatively, you can put the reason between quotes, like so:

```diff
- "The login failed, reason: {reason}."
+ "The login failed, reason: “{reason}”."
```

crates/matrix-sdk/src/authentication/qrcode/rendezvous_channel.rs
```diff
@@ -0,0 +1,433 @@
...
+    error::{FromHttpResponseError, HeaderDeserializationError, IntoHttpError, MatrixError},
+    EndpointError,
+};
+use tracing::{debug, instrument, trace, warn};
+use url::Url;
+
+use crate::{http_client::HttpClient, HttpError, RumaApiError};
+
+const TEXT_PLAIN_CONTENT_TYPE: &str = "text/plain";
+const RECEIVE_TIMEOUT: Duration = Duration::from_secs(1);
```
Hywan, line 32: 1s seems short for a low-end network. Do we want to increase it to 3s maybe?

crates/matrix-sdk/src/authentication/qrcode/rendezvous_channel.rs
```diff
@@ -0,0 +1,433 @@
...
+    ///
+    /// This is currently unused, but left in for completeness sake.
+    #[allow(dead_code)]
+    pub initial_message: Vec<u8>,
+}
+
+struct RendezvousGetResponse {
+    pub status_code: StatusCode,
+    pub etag: String,
+    // TODO: This is currently unused, but will be required once we implement the reciprocation of
```
Hywan, line 65: Can you create an issue in github.com/matrix-org/matrix-rust-sdk/ so that we don't forget about it please?

crates/matrix-sdk/src/authentication/qrcode/rendezvous_channel.rs
```diff
@@ -0,0 +1,433 @@
...
+            .into();
+
+            Err(error)
+        }
+    }
+
+    /// Attempt to receive a message from the [`RendezvousChannel`] from the
+    /// other device.
+    ///
+    /// The should be of the `text/plain` content type but the parsing and
```
Hywan, line 190:
```diff
- The should be of the `text/plain` content type
+ It should have the `text/plain` content type
```

?

crates/matrix-sdk/src/authentication/qrcode/rendezvous_channel.rs
Hywan: OK, after further read the code, I think we should rename this constant to `TIME_BEFORE_RETRY` or something like this?

crates/matrix-sdk/src/authentication/qrcode/rendezvous_channel.rs
```diff
@@ -0,0 +1,433 @@
...
+                }
+                Err(e) => {
+                    warn!("Error while receiving data from the rendezvous channel {e:?}");
+
+                    // If it's a permanent error, return an error.
+                    if let Some(kind) = e.client_api_error_kind() {
+                        if *kind == ErrorKind::NotFound {
+                            return Err(e.into());
+                        }
+                    }
```
Hywan, line 219: Can we add a maximum number of “retries” here so that we are sure we will leave this `loop` whatever happens? For example, if the server is busy, it will return a 5xx error, which is not a 404 as expected here, in which case we will never escape this loop.

Hywan (commented):
Looking really good! Congrats for the quality of the code and the commits, thank you.

I've left a couple of feedback, but nothing too difficult to fix.

crates/matrix-sdk/Cargo.toml
poljar: No, not yet. The fork only adds the `Send` trait bound on one of the things in those crates. We don't need this ourselves, but `uniffi` requires futures to be `Send` for them to be exposed over the FFI. So only the bindings require this.

crates/matrix-sdk/src/authentication/qrcode/login.rs
poljar: Hmm, that doesn't quite work with the bindings, there we have a method which returns this `Result<Client, QrLoginError>`, we would have had this function signature in the main crate but the bindings really want to test if SS is supported before logging in...

So the login had to become a two step process in the main crate. The EX guys told me that they prefer the progress to not contain the error, since they need to get the `Client` object from the future.

Perhaps we can move this function into the `ClientBuilder` and return `Result<Client, QrLoginError>` once the bindings don't need to check for SS anymore, and leave this as is for now?

crates/matrix-sdk/src/authentication/qrcode/messages.rs
poljar: It's just 32 bytes of zeroes. Not the best key, but usable for a test and produces a funny and well recognized base64 representation.

crates/matrix-sdk/src/authentication/qrcode/rendezvous_channel.rs
poljar: Would `POLL_TIMEOUT` work? The MSC uses this term:

image

crates/matrix-sdk/src/authentication/qrcode/rendezvous_channel.rs
poljar: The server will expire the rendezvous session, and if it doesn't do so we should probably expect the user to cancel things.

But you are right, this is a bit too optimistic, we should forward other error types to the user as well.

Fixed and added some more test for this in: 74883c816e129fa2f643bf74c9e470a0a13491fb.

crates/matrix-sdk/src/authentication/qrcode/rendezvous_channel.rs
poljar: Fixed in https://github.com/matrix-org/matrix-rust-sdk/commit/74883c816e129fa2f643bf74c9e470a0a13491fb as well.

crates/matrix-sdk/src/authentication/qrcode/mod.rs
poljar: Removed them dots in 6a66cdf420f2324d40bc6adc240c9e6a11778869.

crates/matrix-sdk/src/authentication/qrcode/messages.rs
poljar: Duh: 191afaec22d059a94cc7ae54ed58f788a28cb198.

crates/matrix-sdk/src/authentication/qrcode/messages.rs
poljar: Don't ask me what I was thinking... 191afaec22d059a94cc7ae54ed58f788a28cb198

crates/matrix-sdk/src/authentication/qrcode/messages.rs
poljar: Found another instance of this: efb2da6c7f42ef6743a0ed57b4d157eae4d8f090.

crates/matrix-sdk/src/authentication/qrcode/rendezvous_channel.rs
poljar: Set it to `POLL_TIMEOUT` and set it to 1ms when run under tests, so our tests don't take seconds to run.

crates/matrix-sdk/src/matrix_auth/mod.rs
```diff
@@ -862,7 +862,7 @@ impl MatrixAuth {
         #[cfg(feature = "e2e-encryption")] login_info: Option<login::v3::LoginInfo>,
     ) -> Result<()> {
         self.set_session_tokens(session.tokens);
-        self.client.set_session_meta(session.meta).await?;
+        self.client.set_session_meta(session.meta, None).await?;
```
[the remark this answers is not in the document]
poljar: You are indeed correct: deb88003ca6e8209d1bc20e6fa2a7c0fe6f9151b.

crates/matrix-sdk/src/oidc/mod.rs
```diff
@@ -705,7 +705,7 @@ impl Oidc {
             authorization_data: Default::default(),
         };
 
-        self.client.set_session_meta(meta).await?;
+        self.client.set_session_meta(meta, None).await?;
```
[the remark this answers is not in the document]
poljar: Yep, true as well: deb88003ca6e8209d1bc20e6fa2a7c0fe6f9151b.

crates/matrix-sdk/src/authentication/qrcode/rendezvous_channel.rs
poljar: I usually avoid doing this because [trouble](https://github.com/folke/trouble.nvim) is much better at showing/searching TODO items in the codebase than github is.

Another point is that some customer already requested this, so it is somewhere already present and will turn up on our plate.

That being said, I wouldn't mind adding it if the above didn't convince you that it isn't necessary.

crates/matrix-sdk/src/encryption/mod.rs
```diff
@@ -1407,6 +1407,22 @@ impl Encryption {
         }
     }
 
+    #[cfg(feature = "experimental-oidc")]
+    pub(crate) async fn ensure_device_keys_upload(&self) -> Result<()> {
```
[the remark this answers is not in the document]
poljar: Added: 008a33d7b85acb946c9fd612ef63b920abc64690.

crates/matrix-sdk/src/authentication/qrcode/login.rs
Hywan: Hmm OK. Let's keep things are they are now.

crates/matrix-sdk/src/authentication/qrcode/rendezvous_channel.rs
Hywan: Fiiiine 😉.

crates/matrix-sdk/src/authentication/qrcode/rendezvous_channel.rs
```diff
@@ -171,56 +186,43 @@ impl RendezvousChannel {
...
+                && message.body.len() != 0
+            {
+                return Ok(message.body);
+            } else if message.status_code == StatusCode::NOT_MODIFIED {
+                tokio::time::sleep(POLL_TIMEOUT).await;
+                continue;
+            } else {
+                let error = response_to_error(message.status_code, message.body);
+
+                return Err(error);
```
Hywan, line 224: :+1:

Hywan (approved):
Thanks for addressing my feedback!

poljar:
> Thanks for addressing my feedback!

No no, thanks for the swift review.

PR #3468, opened 2024-05-25
crypto: Decrypt events bundled in unsigned object of another event

### LINKED ISSUE #2976: `Room::decrypt_event` does not decrypt event bundled in `m.relations`
In the case of edits (`m.replace` relation type), the latest edit is bundled under `m.relations` on the event. It means that, in encrypted rooms, we have two encrypted events in the raw JSON.

Currently, when we pass the raw JSON to `Room::decrypt_event()`, only the main event is decrypted, leaving the bundled one encrypted. So when we deserialize the decrypted JSON, the bundled event is mostly lost because the Ruma type expects an edit of the same type as the main event, but the bundled event is still of type `m.room.encrypted`. It still keeps track that there was an "invalid" replacement, but we lost the content.

That leads to issues like #2848.

My proposal would be to change `Room::decrypt_event()` to also decrypt the bundled event. That has the benefit of being simpler for the users, but has the downside that it will result in breaking changes to the return type, since it should know reflect that we tried to decrypt several events. It should probably not fail if the bundled event failed to be decrypted though, but keep it around so decrypting only the bundled event can be retried later.

It is still possible to solve this issue without the proposed solution by extracting the bundled event in the raw JSON and sending it to `Room::decrypt_event` manually. That has the benefit of not changing the API, but that is something that users will need to do manually. Besides some APIs are already broken because of it, and we could solve them globally instead of one by one. For example, `Room::messages()` will always lose the bundled event in encrypted rooms.

zecakeh:
Allows to potentially have a fully decrypted event.

Fixes #2976.

### DIFF
--- a/crates/matrix-sdk-base/src/read_receipts.rs
+++ b/crates/matrix-sdk-base/src/read_receipts.rs
@@ -772,6 +772,7 @@ mod tests {
                 }),
                 encryption_info: None,
                 push_actions,
+                unsigned_encryption_info: None,
             }
         }
 
@@ -857,6 +858,7 @@ mod tests {
                 }),
                 encryption_info: None,
                 push_actions: Vec::new(),
+                unsigned_encryption_info: None,
             }
         }
 
--- a/crates/matrix-sdk-common/src/deserialized_responses.rs
+++ b/crates/matrix-sdk-common/src/deserialized_responses.rs
@@ -17,7 +17,7 @@ use std::{collections::BTreeMap, fmt};
 use ruma::{
     events::{AnySyncTimelineEvent, AnyTimelineEvent},
     push::Action,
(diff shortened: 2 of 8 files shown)


poljar (approved):
Ok, this looks good, sorry for taking so long to review this.

poljar:
I rerun the CI just in case, seems that you'll need to rebase to make the complement crypto tests happy.

zecakeh:
Ok, I rebased on main.

richvdh:
Would be good to include an entry in `matrix-sdk-crypto/CHANGELOG.md` for changes like this in future.

poljar:
Gah, yes: https://github.com/matrix-org/matrix-rust-sdk/pull/3588.

PR #3469, opened 2024-05-26
sdk: Add method to request user identity from homeserver

### LINKED ISSUE #2924: Add an API to know if another user has set up cross-signing
We want to be able to know if we can verify another user with cross-signing so we can update the interface accordingly in our client.

My understanding is that currently there is only one API in the `matrix-sdk` crate to get the other user's cross-signing identity, which is `Encryption::get_user_identity`, that returns only local cross-signing identities. However there are a few cases where this will return `None`:

1. The user is not tracked, i.e. we don't already share an encrypted room with them: can be solved by creating an encrypted DM, which is required for verification anyway.
2. We just started tracking the user and we haven't received the keys yet: can be solved by waiting.
3. The user has not set up cross-signing: if I understand the spec correctly, the only way to know that is to request their keys with the `/keys/query` endpoint and get an empty response.

If all that is correct I see two ways to solve that:

1. Add an extra public method that requests the keys for the user (or waits for an ongoing request that does that). It seems to have been added in `matrix-sdk-crypto` to solve #2263.
2. Make `Encryption::get_user_identity` try a request if the identity is not found locally.

What is the preferred way to go?

zecakeh:
Fixes #2924.

### DIFF
--- a/crates/matrix-sdk/src/encryption/mod.rs
+++ b/crates/matrix-sdk/src/encryption/mod.rs
@@ -849,8 +849,9 @@ impl Encryption {
     ///
     /// * `user_id` - The ID of the user that the identity belongs to.
     ///
-    /// Returns a `UserIdentity` if one is found. Returns an error if there was
-    /// an issue with the crypto store or with the request to the homeserver.
+    /// Returns a [`UserIdentity`] if one is found. Returns an error if there
+    /// was an issue with the crypto store or with the request to the
+    /// homeserver.
     ///
     /// This will always return `None` if the client hasn't been logged in.
     ///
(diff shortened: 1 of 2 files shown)


crates/matrix-sdk/src/encryption/mod.rs
```diff
@@ -825,16 +828,61 @@ impl Encryption {
...
+    /// up cross-signing.
+    ///
+    /// If you want the E2EE identity of a user without making a request to the
+    /// homeserver, use [`Encryption::get_user_identity()`] instead.
+    ///
+    /// # Arguments
+    ///
+    /// * `user_id` - The ID of the user that the identity belongs to.
+    ///
+    /// Returns a `UserIdentity` if one is found. Returns an error if there was
```
poljar, line 852: Let's linkify this one.

```suggestion
    /// Returns a [`UserIdentity`] if one is found. Returns an error if there was
```

poljar (changes requested):
Looks good, left a tiny nit.

crates/matrix-sdk/src/encryption/mod.rs
zecakeh: Applied.

poljar (approved):

PR #3470, opened 2024-05-26
crypto: Add `SasState::Created` variant

### LINKED ISSUE #2979: `SasState::Started` does not allow to know who started the SAS verification
The [`SasState::Started`](https://matrix-org.github.io/matrix-rust-sdk/matrix_sdk/encryption/verification/enum.SasState.html#variant.Started) state occurs when either party started the SAS verification. However it is important to know whether we started it, because, when we use `SasVerification::changes`, if we receive the state and the other party started it, we need to accept it so the verification can continue.

I thought that's what `SasVerification::we_started` was for, but it's actually the same value as `VerificationRequest::we_started`, which is the party that sent the verification request, not the party that started the SAS verification. That makes sense if the use case is to drop `VerificationRequest` when the `SasVerification` starts and we want to keep the same API.

I am wondering which solution is more acceptable:
1. Keep a boolean internally and expose it in a method, e.g. `SasVerification::we_started_sas()`.
2. Use different variants according to whether we or the other party started it. This is the solution in other cases were it matters who sent what: `VerificationRequestState::Created` vs `VerificationRequestState::Requested` and `QrVerificationState::Reciprocated` vs `QrVerificationState::Scanned`.

zecakeh:
To differentiate the SAS state between the party that sent the verification start and the party that received it.

Fixes #2979.

### DIFF
--- a/bindings/matrix-sdk-crypto-ffi/src/verification.rs
+++ b/bindings/matrix-sdk-crypto-ffi/src/verification.rs
@@ -28,8 +28,11 @@ pub trait SasListener: Send {
 /// An Enum describing the state the SAS verification is in.
 #[derive(uniffi::Enum)]
 pub enum SasState {
-    /// The verification has been started, the protocols that should be used
-    /// have been proposed and can be accepted.
+    /// The verification has been created, the protocols that should be used
+    /// have been proposed to the other party.
+    Created,
+    /// The verification has been started, the other party proposed the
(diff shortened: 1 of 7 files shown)


Hywan (approved):
It looks good to me. I think having a second pair of eyes is not negligible (cc @andybalaam or @poljar).

Thanks!

poljar (approved):
Yeah, this looks fine.

PR #3471, opened 2024-05-27
feat(sdk): Improve `RoomEventCacheUpdate`

Hywan:
This patch improves/refactors `RoomEventCacheUpdate` a little bit. That's preliminary work to support `VectorDiff` from the `EventCache` into the `Timeline`.

---

### DIFF
--- a/crates/matrix-sdk/src/event_cache/mod.rs
+++ b/crates/matrix-sdk/src/event_cache/mod.rs
@@ -44,7 +44,6 @@
 use std::{
     collections::BTreeMap,
     fmt::Debug,
-    result::Result as StdResult,
     sync::{Arc, OnceLock},
 };
 
@@ -60,10 +59,7 @@ use ruma::{
     OwnedEventId, OwnedRoomId, RoomId,
 };
 use tokio::sync::{
-    broadcast::{
-        error::{RecvError, SendError},
-        Receiver, Sender,
-    },
+    broadcast::{error::RecvError, Receiver, Sender},
     Mutex, RwLock, RwLockWriteGuard,
 };
 use tracing::{error, info_span, instrument, trace, warn, Instrument as _, Span};
@@ -676,7 +672,13 @@ impl RoomEventCacheInner {
         }
 
         // The ordering is important here.
-        let _ = self.sender.send(RoomEventCacheUpdate::SyncEvents { timeline, ephemeral })?;
+        let _ = self.sender.send(RoomEventCacheUpdate::AddTimelineEvents {
+            events: sync_timeline_events,
+            origin: EventsOrigin::Sync,
+        });
+        let _ = self
+            .sender
+            .send(RoomEventCacheUpdate::AddEphemeralEvents { events: sync_ephemeral_events });
         let _ = self.sender.send(RoomEventCacheUpdate::UpdateMembers { ambiguity_changes });
 
         Ok(())
@@ -721,14 +723,28 @@ pub enum RoomEventCacheUpdate {
         ambiguity_changes: BTreeMap<OwnedEventId, AmbiguityChange>,
     },
 
-    /// The room has new events.
-    SyncEvents {
+    /// The room has received new timeline events.
+    AddTimelineEvents {
         /// All the new events that have been added to the room's timeline.
-        timeline: Vec<SyncTimelineEvent>,
+        events: Vec<SyncTimelineEvent>,
+
+        /// Where the events are coming from.
+        origin: EventsOrigin,
+    },
+
+    /// The room has received new ephemeral events.
+    AddEphemeralEvents {
         /// XXX: this is temporary, until read receipts are handled in the event
         /// cache
-        ephemeral: Vec<Raw<AnySyncEphemeralRoomEvent>>,
+        events: Vec<Raw<AnySyncEphemeralRoomEvent>>,
     },
+}
+
+/// Indicate where events are coming from.
+#[derive(Debug, Clone)]
+pub enum EventsOrigin {
+    /// Events are coming from a sync.
+    Sync,
 }
 
 #[cfg(test)]
--- a/crates/matrix-sdk/src/event_cache/mod.rs
+++ b/crates/matrix-sdk/src/event_cache/mod.rs
@@ -44,6 +44,7 @@
 use std::{
     collections::BTreeMap,
     fmt::Debug,
+    result::Result as StdResult,
     sync::{Arc, OnceLock},
 };
 
@@ -59,7 +60,10 @@ use ruma::{
     OwnedEventId, OwnedRoomId, RoomId,
 };
 use tokio::sync::{
-    broadcast::{error::RecvError, Receiver, Sender},
+    broadcast::{
+        error::{RecvError, SendError},
+        Receiver, Sender,
+    },
     Mutex, RwLock, RwLockWriteGuard,
 };
 use tracing::{error, info_span, instrument, trace, warn, Instrument as _, Span};
@@ -671,8 +675,19 @@ impl RoomEventCacheInner {
             self.pagination.token_notifier.notify_one();
         }
 
-        let _ =
-            self.sender.send(RoomEventCacheUpdate::Append { events, ephemeral, ambiguity_changes });
+        let _ = self.send_grouped_updates_for_events(events, ephemeral, ambiguity_changes);
+
+        Ok(())
+    }
+
+    fn send_grouped_updates_for_events(
+        &self,
+        events: Vec<SyncTimelineEvent>,
+        ephemeral: Vec<Raw<AnySyncEphemeralRoomEvent>>,
+        ambiguity_changes: BTreeMap<OwnedEventId, AmbiguityChange>,
+    ) -> StdResult<(), SendError<RoomEventCacheUpdate>> {
+        self.sender.send(RoomEventCacheUpdate::Append { events, ephemeral })?;
+        self.sender.send(RoomEventCacheUpdate::Members { ambiguity_changes })?;
 
         Ok(())
     }
@@ -707,6 +722,15 @@ pub enum RoomEventCacheUpdate {
         move_to: OwnedEventId,
     },
 
+    /// The members have changed.
+    Members {
+        /// Collection of ambiguity changes that room member events trigger.
+        ///
+        /// This is a map of event ID of the `m.room.member` event to the
+        /// details of the ambiguity change.
+        ambiguity_changes: BTreeMap<OwnedEventId, AmbiguityChange>,
+    },
+
     /// The room has new events.
     Append {
         /// All the new events that have been added to the room's timeline.
@@ -714,11 +738,6 @@ pub enum RoomEventCacheUpdate {
         /// XXX: this is temporary, until read receipts are handled in the event
         /// cache
         ephemeral: Vec<Raw<AnySyncEphemeralRoomEvent>>,
-        /// Collection of ambiguity changes that room member events trigger.
-        ///
-        /// This is a map of event ID of the `m.room.member` event to the
-        /// details of the ambiguity change.
-        ambiguity_changes: BTreeMap<OwnedEventId, AmbiguityChange>,
     },
 }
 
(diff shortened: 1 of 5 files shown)


crates/matrix-sdk/src/event_cache/mod.rs
```diff
@@ -702,7 +702,7 @@ pub enum RoomEventCacheUpdate {
     Clear,
 
     /// The fully read marker has moved to a different event.
-    UpdateReadMarker {
+    ReadMarker {
```
bnjbvr, line 724: I disagree with this change: this breaks symmetry with the other two variants which start with a verb, and since it's a deferred command it makes sense to start with a verb. Can you revert this please?

crates/matrix-sdk/src/event_cache/mod.rs
```diff
@@ -704,7 +704,7 @@ pub enum RoomEventCacheUpdate {
     /// The fully read marker has moved to a different event.
     ReadMarker {
         /// Event at which the read marker is now pointing.
-        event_id: OwnedEventId,
+        move_to: OwnedEventId,
```
bnjbvr, line 726: Sorry, but not sure I like this change either; maybe the read marker was already on that event in which case nothing moves. It's really "at which event id is the read marker pointing to right now".

crates/matrix-sdk/src/event_cache/mod.rs
```diff
@@ -671,8 +675,19 @@ impl RoomEventCacheInner {
...
+    }
+
+    fn send_grouped_updates_for_events(
+        &self,
+        events: Vec<SyncTimelineEvent>,
+        ephemeral: Vec<Raw<AnySyncEphemeralRoomEvent>>,
+        ambiguity_changes: BTreeMap<OwnedEventId, AmbiguityChange>,
+    ) -> StdResult<(), SendError<RoomEventCacheUpdate>> {
+        self.sender.send(RoomEventCacheUpdate::Append { events, ephemeral })?;
+        self.sender.send(RoomEventCacheUpdate::Members { ambiguity_changes })?;
```
bnjbvr: I'm not sure introducing a new function is worth it to do so little, especially as it's happening only once. But if you care strongly about having a new function, can you document that it exists to ensure the ordering invariant, please? (as a code comment above the first `send()` maybe?)

crates/matrix-sdk/src/event_cache/mod.rs
```diff
@@ -707,18 +722,22 @@ pub enum RoomEventCacheUpdate {
         move_to: OwnedEventId,
     },
 
+    /// The members have changed.
+    Members {
```
bnjbvr, line 730: Can you follow the naming convention `verb + action` here too?

crates/matrix-sdk/src/event_cache/mod.rs
```diff
@@ -732,9 +736,9 @@ pub enum RoomEventCacheUpdate {
     },
 
     /// The room has new events.
-    Append {
+    SyncEvents {
         /// All the new events that have been added to the room's timeline.
-        events: Vec<SyncTimelineEvent>,
+        timeline: Vec<SyncTimelineEvent>,
```
bnjbvr, line 741: Please please please no, absolutely not. Timeline is already way too overloaded (as in "collection of events" or "the timeline API object from the UI crate), and it's much clearer to call this a collection of events, as one could imagine it's the full timeline from looking at the term here.

crates/matrix-sdk/src/event_cache/mod.rs
```diff
@@ -646,14 +646,14 @@ impl RoomEventCacheInner {
     async fn append_events_locked_impl(
         &self,
         mut room_events: RwLockWriteGuard<'_, RoomEvents>,
-        events: Vec<SyncTimelineEvent>,
+        timeline_events: Vec<SyncTimelineEvent>,
```
bnjbvr, line 649: I don't like this renaming either, there's something called `TimelineEvent` and the name suggests it's a collection of such `TimelineEvent`. Please keep it as `events` or use the full type in the name, aka `sync_timeline_events`.

crates/matrix-sdk/src/event_cache/mod.rs
```diff
@@ -623,16 +623,16 @@ impl RoomEventCacheInner {
     /// observers.
     async fn append_new_events(
         &self,
-        events: Vec<SyncTimelineEvent>,
+        timeline_events: Vec<SyncTimelineEvent>,
```
bnjbvr, line 626: ditto

crates/matrix-sdk/src/event_cache/mod.rs
```diff
@@ -594,9 +594,9 @@ impl RoomEventCacheInner {
     /// storage, notifying observers.
     async fn replace_all_events_by(
         &self,
-        events: Vec<SyncTimelineEvent>,
+        timeline_events: Vec<SyncTimelineEvent>,
```
bnjbvr, line 597: ditto

bnjbvr (changes requested):
Sorry, I think most of these renamings lower clarity, introduce confusion and/or break preexisting conventions around the event cache. I'm fine with `SyncEvents`, but the rest should be re-discussed and made consistent, at the very least.

crates/matrix-sdk/src/event_cache/mod.rs
Hywan: `…Update::Update…`, that's what I dislike. What do you think about `MoveReadMarker`?

crates/matrix-sdk/src/event_cache/mod.rs
Hywan: `sync_timeline_events` is better then.

crates/matrix-sdk/src/event_cache/mod.rs
bnjbvr: Sounds good!

crates/matrix-sdk/src/event_cache/mod.rs
Hywan: I've renamed `Sync` to `ReceiveEventsFromSync` and the `timeline` field to `timeline_events`, it reads as `ReceiveEventsFromSync::timeline_events`, I believe there is no ambiguity here.

crates/matrix-sdk/src/event_cache/mod.rs
Hywan: Inlined the function body.

crates/matrix-sdk/src/event_cache/mod.rs
Hywan: The comment literally says “The fully read marker has moved to a different event”, I was assuming this renaming was alright.

I've renamed to `MoveReadMarkerTo::event_id`.

crates/matrix-sdk/src/event_cache/mod.rs
```diff
@@ -590,9 +590,9 @@ impl RoomEventCacheInner {
     /// storage, notifying observers.
     async fn replace_all_events_by(
         &self,
-        events: Vec<SyncTimelineEvent>,
+        sync_timeline_events: Vec<SyncTimelineEvent>,
         prev_batch: Option<String>,
-        ephemeral: Vec<Raw<AnySyncEphemeralRoomEvent>>,
+        sync_ephemeral_events: Vec<Raw<AnySyncEphemeralRoomEvent>>,
```
bnjbvr, line 595: No need for the `sync_` prefix for ephemeral events: as opposed to the `sync_timeline_events` which make a distinction between `TimelineEvent` and `SyncTimelineEvent` (two different types that exist in the Rust SDK), there's no `SyncEphemeralEvent` in the code base, and the fully qualified name with the type in it would be `any_sync_ephemeral_room_event` which we don't really care about, so let's keep it short :)

```suggestion
        ephemeral_events: Vec<Raw<AnySyncEphemeralRoomEvent>>,
```

crates/matrix-sdk/src/event_cache/mod.rs
```diff
@@ -607,9 +607,9 @@ impl RoomEventCacheInner {
         // Push the new events.
         self.append_events_locked_impl(
             room_events,
-            events,
+            sync_timeline_events,
             prev_batch,
-            ephemeral,
+            sync_ephemeral_events,
```
bnjbvr, line 612:
```suggestion
            ephemeral_events,
```

crates/matrix-sdk/src/event_cache/mod.rs
```diff
@@ -619,16 +619,16 @@ impl RoomEventCacheInner {
     /// observers.
     async fn append_new_events(
         &self,
-        events: Vec<SyncTimelineEvent>,
+        sync_timeline_events: Vec<SyncTimelineEvent>,
         prev_batch: Option<String>,
-        ephemeral: Vec<Raw<AnySyncEphemeralRoomEvent>>,
+        sync_ephemeral_events: Vec<Raw<AnySyncEphemeralRoomEvent>>,
```
bnjbvr, line 624:
```suggestion
        ephemeral_events: Vec<Raw<AnySyncEphemeralRoomEvent>>,
```

crates/matrix-sdk/src/event_cache/mod.rs
```diff
@@ -619,16 +619,16 @@ impl RoomEventCacheInner {
...
         ambiguity_changes: BTreeMap<OwnedEventId, AmbiguityChange>,
     ) -> Result<()> {
         self.append_events_locked_impl(
             self.events.write().await,
-            events,
+            sync_timeline_events,
             prev_batch,
-            ephemeral,
+            sync_ephemeral_events,
```
bnjbvr, line 631:
```suggestion
            ephemeral_events,
```

crates/matrix-sdk/src/event_cache/mod.rs
```diff
@@ -642,14 +642,14 @@ impl RoomEventCacheInner {
     async fn append_events_locked_impl(
         &self,
         mut room_events: RwLockWriteGuard<'_, RoomEvents>,
-        events: Vec<SyncTimelineEvent>,
+        sync_timeline_events: Vec<SyncTimelineEvent>,
         prev_batch: Option<String>,
-        ephemeral: Vec<Raw<AnySyncEphemeralRoomEvent>>,
+        sync_ephemeral_events: Vec<Raw<AnySyncEphemeralRoomEvent>>,
```
bnjbvr, line 647:
```suggestion
        ephemeral_events: Vec<Raw<AnySyncEphemeralRoomEvent>>,
```

crates/matrix-sdk/src/event_cache/mod.rs
```diff
@@ -642,14 +642,14 @@ impl RoomEventCacheInner {
...
         ambiguity_changes: BTreeMap<OwnedEventId, AmbiguityChange>,
     ) -> Result<()> {
-        if events.is_empty()
+        if sync_timeline_events.is_empty()
             && prev_batch.is_none()
-            && ephemeral.is_empty()
+            && sync_ephemeral_events.is_empty()
```
bnjbvr, line 652:
```suggestion
            && ephemeral_events.is_empty()
```

crates/matrix-sdk/src/event_cache/mod.rs
```diff
@@ -671,8 +671,25 @@ impl RoomEventCacheInner {
...
+        // The order of `RoomEventCacheUpdate`s is **really** important here.
+        {
+            if !sync_timeline_events.is_empty() {
+                let _ = self.sender.send(RoomEventCacheUpdate::AddTimelineEvents {
+                    events: sync_timeline_events,
+                    origin: EventsOrigin::Sync,
+                });
+            }
+
+            if !sync_ephemeral_events.is_empty() {
```
bnjbvr, line 683:
```suggestion
            if !ephemeral_events.is_empty() {
```

crates/matrix-sdk/src/event_cache/mod.rs
```diff
@@ -671,8 +671,25 @@ impl RoomEventCacheInner {
...
+                let _ = self.sender.send(RoomEventCacheUpdate::AddEphemeralEvents {
+                    events: sync_ephemeral_events,
```
bnjbvr, line 685:
```suggestion
                    events: ephemeral_events,
```

bnjbvr (approved):

PR #3472, opened 2024-05-28
feat(sdk): Add `RoomPagination::run_backwards(…, until)`

Hywan:
This patch adds a new argument to `RoomPagination::run_backwards`: `until`. It becomes:

```rust
pub async fn run_backwards<Until, Break, UntilFuture>(
    &self,
    batch_size: u16,
    mut until: Until,
) -> Result<Break>
where
    Until: FnMut(BackPaginationOutcome, TimelineHasBeenResetWhilePaginating) -> UntilFuture,
    UntilFuture: Future<Output = ControlFlow<Break, ()>>,
```

The idea behind `until` is to run pagination _until_ `until` returns `ControlFlow::Break`, otherwise it continues paginating.

This is useful is many scenearii (cf. the documentation). This is also and primarily the first step to stop adding events directly from the pagination, and starts adding events only and strictly only from `event_cache::RoomEventCacheUpdate` (again, see the `TODO` in the documentation). This is not done in this patch for the sake of ease of review.

---

### DIFF
--- a/crates/matrix-sdk/src/event_cache/pagination.rs
+++ b/crates/matrix-sdk/src/event_cache/pagination.rs
@@ -115,7 +115,7 @@ impl RoomPagination {
             if let Some(outcome) = self.run_backwards_impl(batch_size).await? {
                 match until(outcome, timeline_has_been_reset).await {
                     ControlFlow::Continue(()) => {
-                        debug!("back-pagination continues");
+                        trace!("back-pagination continues");
 
                         timeline_has_been_reset = TimelineHasBeenResetWhilePaginating::No;
 
--- a/crates/matrix-sdk/tests/integration/event_cache.rs
+++ b/crates/matrix-sdk/tests/integration/event_cache.rs
@@ -435,6 +435,7 @@ async fn test_backpaginate_many_times_with_many_iterations() {
     }
 
     let mut num_iterations = 0;
+    let mut num_paginations = 0;
     let mut global_events = Vec::new();
     let mut global_reached_start = false;
 
@@ -459,19 +460,29 @@ async fn test_backpaginate_many_times_with_many_iterations() {
     // Then if I backpaginate in a loop,
     let pagination = room_event_cache.pagination();
     while pagination.get_or_wait_for_token().await.is_some() {
-        let BackPaginationOutcome { reached_start, events } =
-            pagination.run_backwards(20, once).await.unwrap();
+        pagination
+            .run_backwards(20, |outcome, timeline_has_been_reset| {
+                num_paginations += 1;
 
-        if !global_reached_start {
-            global_reached_start = reached_start;
-        }
-        global_events.extend(events);
+                assert_matches!(timeline_has_been_reset, TimelineHasBeenResetWhilePaginating::No);
+
+                if !global_reached_start {
+                    global_reached_start = outcome.reached_start;
+                }
+
+                global_events.extend(outcome.events);
+
+                ready(ControlFlow::Break(()))
+            })
+            .await
+            .unwrap();
 
         num_iterations += 1;
     }
 
     // I'll get all the previous events,
-    assert_eq!(num_iterations, 2);
+    assert_eq!(num_iterations, 2); // in two iterations…
+    assert_eq!(num_paginations, 2); // … we get two paginations.
     assert!(global_reached_start);
 
     assert_event_matches_msg(&global_events[0], "world");
@@ -540,6 +551,7 @@ async fn test_backpaginate_many_times_with_one_iteration() {
     }
 
     let mut num_iterations = 0;
+    let mut num_paginations = 0;
     let mut global_events = Vec::new();
     let mut global_reached_start = false;
 
@@ -565,7 +577,11 @@ async fn test_backpaginate_many_times_with_one_iteration() {
     let pagination = room_event_cache.pagination();
     while pagination.get_or_wait_for_token().await.is_some() {
         pagination
-            .run_backwards(20, |outcome, _timeline_has_been_reset| {
+            .run_backwards(20, |outcome, timeline_has_been_reset| {
+                num_paginations += 1;
+
+                assert_matches!(timeline_has_been_reset, TimelineHasBeenResetWhilePaginating::No);
+
                 if !global_reached_start {
                     global_reached_start = outcome.reached_start;
                 }
@@ -585,7 +601,8 @@ async fn test_backpaginate_many_times_with_one_iteration() {
     }
 
     // I'll get all the previous events,
-    assert_eq!(num_iterations, 1); // in one iteration!
+    assert_eq!(num_iterations, 1); // in one iteration…
+    assert_eq!(num_paginations, 2); // … we get two paginations!
     assert!(global_reached_start);
 
     assert_event_matches_msg(&global_events[0], "world");


crates/matrix-sdk-ui/src/timeline/pagination.rs
```diff
@@ -67,49 +69,45 @@ impl super::Timeline {
...
-            let num_events = events.len();
-            trace!("Back-pagination succeeded with {num_events} events");
-
-            self.inner
-                .add_events_at(events, TimelineEnd::Front, RemoteEventOrigin::Pagination)
-                .await;
-
-            if reached_start {
-                return Ok(true);
+        let reached_start = pagination
```
bnjbvr, line 72: nit: can you keep the previous name `result` for this variable please?

crates/matrix-sdk-ui/src/timeline/pagination.rs
```diff
@@ -67,49 +69,45 @@ impl super::Timeline {
...
-                // As an exceptional contract: if there were no events in the response,
-                // and we've not hit the start of the timeline, retry until we get
-                // some events or reach the start of the timeline.
-                continue;
-            }
-
-            // Exit the inner loop, and ask for another limit.
-            break;
+            Err(err) => Err(err),
         }
```
bnjbvr, line 110: Looks like this match could be simplified to explicitly handle the intricate `Err(...)` case, and have a `result => result` nistead of `Ok(reached_start) => Ok(reached_start)` and `Err(err) => Err(err)`?

crates/matrix-sdk/src/event_cache/pagination.rs
```diff
@@ -59,17 +59,60 @@ impl RoomPagination {
...
+    ///         // Do something with `events` and `reached_start` maybe?
+    ///         let _ = events;
+    ///         let _ = reached_start;
+    ///
+    ///         ControlFlow::Break(())
+    ///     }
+    /// ).await;
+    /// # }
+    #[instrument(skip(self, until))]
+    pub async fn run_backwards<F, B, Fut>(&self, batch_size: u16, mut until: F) -> Result<B>
```
bnjbvr, line 98: Can you give better names for the generics `F` and `B`?

crates/matrix-sdk/src/event_cache/pagination.rs
```diff
@@ -59,17 +59,60 @@ impl RoomPagination {
...
+    where
+        F: FnMut(BackPaginationOutcome) -> Fut,
+        Fut: Future<Output = ControlFlow<B, ()>>,
+    {
         loop {
-            if let Some(result) = self.run_backwards_impl(batch_size).await? {
-                return Ok(result);
+            if let Some(outcome) = self.run_backwards_impl(batch_size).await? {
+                match until(outcome).await {
```
bnjbvr, line 105: Unfortunately, this is not the right thing to do, because of restarts. Imagine the following situation:

- a caller would like to continue until it got 1000 events (but it uses a smaller batch size, say 100).
- the first back-pagination works and returns 100 events, the `until()` method provided will record that it's still waiting for 900 events
- a restart happens because of a timeline reset (limited flag is set)
- another back-pagination works, starting from the end of the timeline, and returns 100 events.
- the `until()` method records that we have received 100 events.

At this point, the timeline only contains 100 events, but the until() method wasn't aware of the restart so it thought that it included 200 events already.

One way to fix this would be to include a boolean parameter in the `until()` parameters' list, that indicates a timeline reset, so the caller may do appropriate action whenever that happens.

Can you include a test for this, please?

crates/matrix-sdk/tests/integration/event_cache.rs
```diff
@@ -24,6 +24,10 @@ use wiremock::{
 
 use crate::mock_sync;
 
+pub async fn once(outcome: BackPaginationOutcome) -> ControlFlow<BackPaginationOutcome, ()> {
```
bnjbvr, line 27:
```suggestion
async fn once(outcome: BackPaginationOutcome) -> ControlFlow<BackPaginationOutcome, ()> {
```

crates/matrix-sdk/tests/integration/event_cache.rs
```diff
@@ -656,7 +660,7 @@ async fn test_backpaginating_without_token() {
     // If we try to back-paginate with a token, it will hit the end of the timeline
     // and give us the resulting event.
     let BackPaginationOutcome { events, reached_start } =
-        pagination.run_backwards(20).await.unwrap();
+        pagination.run_backwards(20, once).await.unwrap();
```
bnjbvr, line 663: Can you add tests where the `until()` method is *not* once?

crates/matrix-sdk-ui/src/timeline/pagination.rs
bnjbvr, line 1: Nice, less indent and a more straightforward control-flow in this file \o/

bnjbvr (changes requested):
Cheers! This may require a bit more testing…

crates/matrix-sdk/tests/integration/event_cache.rs
Hywan: Done!

crates/matrix-sdk/src/event_cache/pagination.rs
Hywan: Good eye, thanks. I've added a `TimelineHasBeenResetWhilePaginating` enum type (to avoid having a bool…), and the tests have been added/updated accordingly.

crates/matrix-sdk/src/event_cache/pagination.rs
Hywan: Done.

crates/matrix-sdk/src/event_cache/pagination.rs
```diff
@@ -59,17 +59,75 @@ impl RoomPagination {
...
+    {
+        let mut timeline_has_been_reset = TimelineHasBeenResetWhilePaginating::No;
+
         loop {
-            if let Some(result) = self.run_backwards_impl(batch_size).await? {
-                return Ok(result);
+            if let Some(outcome) = self.run_backwards_impl(batch_size).await? {
+                match until(outcome, timeline_has_been_reset).await {
+                    ControlFlow::Continue(()) => {
+                        debug!("back-pagination continues");
```
bnjbvr, line 118: Not sure this is useful to have, since there's a `debug!()` line already when the more exceptional timeline reset happens. If you really want to keep it, can we make it a trace log instead?

```suggestion
                        trace!("back-pagination continues");
```

crates/matrix-sdk/src/event_cache/pagination.rs
```diff
@@ -259,6 +317,16 @@ impl RoomPagination {
     }
 }
 
+/// A type representing whether the timeline has been reset.
+#[derive(Debug)]
+pub enum TimelineHasBeenResetWhilePaginating {
```
bnjbvr, line 322: Roses are red,
Violets are blue,
Unnamed booleans suck,
All hail enums.

:heart_hands: 

The name is a bit of a mouthful though, can you think of a shorter one please?

crates/matrix-sdk/tests/integration/event_cache.rs
```diff
@@ -481,6 +491,120 @@ async fn test_backpaginate_multiple_iterations() {
...
+        None,
+        non_sync_events!(event_builder, [ (room_id, "$4": "oh well"), ]),
+    )
+    .await;
+
+    // Then if I backpaginate in a loop,
+    let pagination = room_event_cache.pagination();
+    while pagination.get_or_wait_for_token().await.is_some() {
+        pagination
+            .run_backwards(20, |outcome, _timeline_has_been_reset| {
```
bnjbvr, line 568: Can we also assert `timeline_has_been_reset` is false?

crates/matrix-sdk/tests/integration/event_cache.rs
```diff
@@ -481,6 +491,120 @@ async fn test_backpaginate_multiple_iterations() {
...
+                })
+            })
+            .await
+            .unwrap();
+
+        num_iterations += 1;
+    }
+
+    // I'll get all the previous events,
+    assert_eq!(num_iterations, 1); // in one iteration!
```
bnjbvr, line 588: Uh? wasn't there two calls to the endpoint? (Probably above `num_iterations += 1` should be inside the continue handler?)

bnjbvr (approved):
yay

crates/matrix-sdk/src/event_cache/pagination.rs
Hywan: Let's remove it.

crates/matrix-sdk/src/event_cache/pagination.rs
Hywan: I thought about `TimelineHasBeenReset` but I'm afraid it can create confusion if used somewhere else. Let's keep the name as is, and let's refactor if needed.

crates/matrix-sdk/tests/integration/event_cache.rs
Hywan: Done.

bnjbvr:
This didn't pass tests on CI, and it's been force-merged. Will revert because it was incorrectly merged.

PR #3473, opened 2024-05-28
Bindings for the QR code login

poljar:


bindings/matrix-sdk-ffi/src/client_builder.rs
```diff
@@ -27,6 +33,140 @@ enum HomeserverConfig {
...
+
+#[derive(Debug, thiserror::Error, uniffi::Error)]
+pub enum HumanQrLoginError {
+    #[error("Linking with this device is not supported.")]
+    LinkingNotSupported,
+    #[error("The sign in was cancelled.")]
+    Cancelled,
+    #[error("The sign in was not completed in the required time.")]
+    Expired,
+    #[error("A secure connection could not have been established between the two device.")]
```
bnjbvr, line 75:
```suggestion
    #[error("A secure connection could not have been established between the two devices.")]
```

bindings/matrix-sdk-ffi/src/client_builder.rs
```diff
@@ -27,6 +33,140 @@ enum HomeserverConfig {
...
+pub enum QrCodeDecodeError {
+    #[error("Error decoding QR code: {error:?}")]
+    Crypto {
+        #[from]
+        error: LoginQrCodeDecodeError,
+    },
+}
+
+#[derive(Debug, thiserror::Error, uniffi::Error)]
+pub enum HumanQrLoginError {
```
bnjbvr, line 68: Is the intent to display those errors to the end-user? If so, I think it's not our task, since these error messages have to be localized, and it's likely we're not the best equipped for crafting the most understandable error messages ("what's a sliding sync? what's a homeserver?"), although I appreciate the intent to have precise errors that are nicely documented :+1: 

(No special action required; maybe "downgrade" the `error()` to doc comments, but maybe the apps may use these error messages directly, not sure.)

bindings/matrix-sdk-ffi/src/client_builder.rs
```diff
@@ -27,6 +33,140 @@ enum HomeserverConfig {
...
+            },
+            QRCodeLoginError::Oidc(e) => {
+                if let Some(e) = e.as_request_token_error() {
+                    match e {
+                        DeviceCodeErrorResponseType::AccessDenied => HumanQrLoginError::Declined,
+                        DeviceCodeErrorResponseType::ExpiredToken => HumanQrLoginError::Expired,
+                        _ => HumanQrLoginError::Unknown,
+                    }
+                } else {
+                    HumanQrLoginError::Unknown
```
bnjbvr, line 108: Are these unknown errors logged some place else in the code base?

bindings/matrix-sdk-ffi/src/client_builder.rs
```diff
@@ -27,6 +33,140 @@ enum HomeserverConfig {
...
+    Starting,
+    /// We established a secure channel with the other device.
+    EstablishingSecureChannel {
+        /// The check code that the device should display so the other device
+        /// can confirm that the channel is secure as well.
+        check_code: u8,
+        /// The string representation of the check code, will be guaranteed to
+        /// be 2 characters long, preserving the leading zero if the
+        /// first digit is a zero.
+        check_code_string: String,
```
bnjbvr, line 136: Could this be `[char; 2]`, or is that too much for uniffi?

bindings/matrix-sdk-ffi/src/client_builder.rs
```diff
@@ -238,6 +378,109 @@ impl ClientBuilder {
...
+    /// [`CheckCode`] to the existing client.
+    ///
+    /// [MSC4108]: https://github.com/matrix-org/matrix-spec-proposals/pull/4108
+    pub async fn build_with_qr_code(
+        self: Arc<Self>,
+        qr_code_data: &QrCodeData,
+        oidc_configuration: &OidcConfiguration,
+        progress_listener: Box<dyn QrLoginProgressListener>,
+    ) -> Result<Arc<Client>, HumanQrLoginError> {
+        if let QrCodeModeData::Reciprocate { homeserver_url } = &qr_code_data.inner.mode_data {
```
bnjbvr, line 400: style proposal to spare one full level of indent:

 `let QrCodeModeData::Reciprocate { homeserver_url } = &qr_code_data.inner.mode_data else { return Err(HumanQrLoginError::InvalidQrCode); }`

bindings/matrix-sdk-ffi/src/client_builder.rs
```diff
@@ -238,6 +378,109 @@ impl ClientBuilder {
...
+            let client_metadata = oidc_configuration
+                .try_into()
+                .map_err(|_| HumanQrLoginError::OidcMetadataInvalid)?;
+
+            let oidc = client.inner.oidc();
+            let login = oidc.login_with_qr_code(&qr_code_data.inner, client_metadata);
+
+            let mut progress = login.subscribe_to_progress();
+
+            let _progress_task = TaskHandle::new(RUNTIME.spawn(async move {
```
bnjbvr, line 442: a `TaskHandle::drop` will abort the task, so surely this `_progress_task` must either be returned to the caller, or stashed somewhere else?

bindings/matrix-sdk-ffi/src/client_builder.rs
```diff
@@ -265,11 +508,16 @@ impl ClientBuilder {
...
             let data_path = PathBuf::from(base_path).join(sanitize(username));
             fs::create_dir_all(&data_path)?;
 
+            inner_builder = inner_builder.sqlite_store(&data_path, builder.passphrase.as_deref());
+        } else if let Some(base_path) = &builder.base_path {
+            let data_path = PathBuf::from(base_path);
+            fs::create_dir_all(&data_path)?;
+
             inner_builder = inner_builder.sqlite_store(&data_path, builder.passphrase.as_deref());
         }
```
bnjbvr, line 522: This control flow could be simplified to something like:

```
if let Some(base_path) = ... {
  let data_path = if let Some(username) = ... { ... } else { ... };
 ...
}
```

bnjbvr (changes requested):
I think there's a bug with the `TaskHandle`, but the rest looks good. Excited to ship it soon!

bindings/matrix-sdk-ffi/src/client_builder.rs
poljar: We used to have two error types, the full Rust one and this one, and a method to convert from one to the other, but uniffi wouldn't comply as of now with that strategy. We might want to try to bring the full error type back once we add the other QR login flows.

The applications will not use the error messages, though each of those variants will create a different error screen.

That being said, I can imagine some other application which would not care about localizations to use the error message directly.

bindings/matrix-sdk-ffi/src/client_builder.rs
poljar: Very unlikely, at least not until complement bumps its uniffi version, I had to remove `&[u8]` from one function because of that, I don't see any arrays being supported: https://mozilla.github.io/uniffi-rs/udl/builtin_types.html

bindings/matrix-sdk-ffi/src/client_builder.rs
poljar: Why? No further progress will be made once the function completes.

bindings/matrix-sdk-ffi/src/client_builder.rs
poljar: Ah indeed: https://github.com/matrix-org/matrix-rust-sdk/pull/3473/commits/18c8696883e3a76b63730a53175b565677f5f647.

Thanks.

bindings/matrix-sdk-ffi/src/client_builder.rs
poljar: Yeah, the main crate has a bunch of logs.

bindings/matrix-sdk-ffi/src/client_builder.rs
poljar: I actually prefer it this way, as long as it's only one indent.

bindings/matrix-sdk-ffi/src/client_builder.rs
bnjbvr: Oh, I missed that the following code could fail thereafter, so it's just a small API crime in that we're abusing a public API to abort on drop, correct? If so, I *think* we had some `abort_on_drop()` extension method on a trait that achieves the same. But at this point, it's ok to keep `TaskHandle`, maybe add a comment explaining why it's fine to not use the task handle, please? since another reader may go through the same steps as I did (I know future-me would surely appreciate this).

poljar: I added a comment, though is it an API crime to use a public API? We just want a task that will not detach itself implicitly. Other executors might create tasks which require explicit detachment and the default is the cancellation.

I don't think that this is terribly important since the stream should end, once the future is dropped the `SharedObservable` is dropped, but the cancellation shouldn't hurt either as a precaution.

PR #3478, opened 2024-05-28
room: cache the return of a previous call to `computed_display_name`  and use that in search filters

bnjbvr:
It's not possible to use async code from a search filter, so we can't just plain call `Room::computed_display_name`. Also, it would be too expensive to recompute it every single time we call that function, hence the idea of caching the result.

Ideally, we'd have something like a "get_cached_or_recompute()` method, but it would need to be async as well in the case we have to actually compute the underlying `computed_display_name`. Instead, this commit proposes a stopgap that will store the outcome of
`computed_display_name` in a local internal cache, and make it available in a sync method, so that the search filters may use it.

In theory, the computed display name could then be empty, but! the Room list service that offers the filters is powered by a sliding sync that publishes `RoomInfo` updates for rooms, and `RoomInfo` will internally call `computed_display_name` too.

All in all, it'd be better if there was a better proper caching, but this can act as a stopgap solution in the meanwhile.

### DIFF
--- a/crates/matrix-sdk-base/src/rooms/normal.rs
+++ b/crates/matrix-sdk-base/src/rooms/normal.rs
@@ -93,6 +93,9 @@ pub struct Room {
     roominfo_update_sender: broadcast::Sender<RoomInfoUpdate>,
     store: Arc<DynStateStore>,
 
+    /// Cached computed display name.
+    computed_display_name: Arc<SyncRwLock<Option<DisplayName>>>,
+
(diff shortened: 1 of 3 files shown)


bnjbvr:
Hmm this approach is really wrong. Caching could happen independently, but we'd need to either have a sync version of `computed_display_name()` or allow filters to be async :thinking:

bnjbvr:
I don't think that's a good approach, we'd need something like async filters instead.

PR #3479, opened 2024-05-29
bindings(qr): Add an error variant for when the other device is not signed in

poljar:


hughns (approved):
The new value and revised mapping look good to me.

PR #3444, opened 2024-05-23
Expose new sticker source field (and drop url field)

surakin:
Depends on https://github.com/ruma/ruma/pull/1820

The StickerEventContent struct changed (url is gone, source was added), this change exposes the new field.

- [ ] Public API changes documented in changelogs (optional)

Signed-off-by: Marco Alvarez <surakin@gmail.com>

PR #3482, opened 2024-05-30
chore: Go back to depending on Ruma from the Ruma repo

poljar:
Bumping Ruma further would require #3444.


bnjbvr (approved):
Sweet! Looking forward to using a released version…

PR #3484, opened 2024-05-30
Fix trigger condition for new upload_coverage workflow

jplatte:
Because GitHub actions has no IDs for workflows, you refer to them by name. What could possibly go wrong?


bnjbvr (approved):
Can we also rename this flow to "Code Coverage Upload", while we're around here, please?

PR #3485, opened 2024-05-30
ci: Fix coverage uploads in separate workflow

jplatte:
Should start working after this merges. Second-to-last commit is the main fix, everything else is related refactorings.


bnjbvr (approved):
Thanks for spending time to fix this! Let's try it out.

PR #3475, opened 2024-05-28
timeline: make use of the sending queue

bnjbvr:
This is part 2 of https://github.com/matrix-org/matrix-rust-sdk/pull/3465, extracted to make review easier. This integrates the new sending queue with the timeline, exposes FFI methods, and adds some way to quickly test the new queuing mechanism in `multiverse`.

Note to embedders: to allow cancelling a local echo, embedders should use `Timeline::redact()` instead of `Room::redact()`. This requires an `EventTimelineItem`, which can be retrieved using `get_event_timeline_item_with_event_id` (~for remote echoes) or `get_event_timeline_item_with_transaction_id` (for local echoes).

Part of #3361.

PR #3488, opened 2024-05-31
timeline/sendq: add a `Timeline::redact()` method and tweak local echoes behavior

bnjbvr:
Moar changes requested by @stefanceriu, please test :)

- Add a way to redact an event directly from the timeline. It's more general than `Room::redact` since it can have an effect on local echoes too.
- Gets rid of the EventSendState::Cancelled variant, since it's a white lie now that the sending queue automatically retries in the background.
- Mark local echoes (that have not been sent yet) as not editable, and test this functionality (!).

Part of #3361.


bnjbvr:
(Closing, #3475 contains all the changes now.)

PR #3497, opened 2024-06-03
clippy: fix extra errors and enforce `target-applies-to-host` in xtask

bnjbvr:
This enforces the env variable to set the `--target-applies-to-host` setting when running in xtask. I'm not sure why it's required, but without it, the CI step didn't find all the issues that I could find locally. @jplatte if you have any idea about that, I'd be curious; I've mostly cargo culted the existing workaround used when building the doc. It looks like another instance of https://github.com/rust-lang/cargo/issues/10744 in a slightly different context.

The second commit fixes the issues found by clippy.


Hywan (approved):
It looks sensible to me, thanks for your investigation.

PR #3498, opened 2024-06-03
ci: attempt to reproduce the clippy failures in CI step

bnjbvr:
(Don't review yet)


bnjbvr:
Yup, that was it :partying_face:

PR #3501, opened 2024-06-04
pagination: don't hold onto the waited_for_initial_prev_token lock for long

bnjbvr:
While this mutex is taken, in `oldest_token()` we can get a hold of the RoomEvent mutex, making it so that the `waited_...` token covers the critical region of room events.

Unfortunately, in `clear()`, we're taking these two locks in the opposite order, implying that the room events critical region now overlaps that of the `waited_...` lock.

The way to break the cycle is to keep the `waited_` token for as short as possible.


ganfra:
Tested ok on EXA!

andybalaam (approved):
Looks good, thanks!

Hywan:
Well done! 👏

PR #3503, opened 2024-06-04
ffi: Expose room heroes (adding support for avatars).

### LINKED ISSUE #2702: Add room heroes to `RoomSubscription`
Now that https://github.com/matrix-org/sliding-sync/pull/304 has been merged, there's a new parameter that can be included on `RoomSubscription` - [`include_heroes: Bool`](https://github.com/matrix-org/sliding-sync/pull/304/files#diff-6c24ba2ad5df22c2d7ddc50e12fc731b05b2aaa115e0993220bd41ad8ed69714R572) which will in turn include the [heroes](https://github.com/matrix-org/sliding-sync/blob/20c361ffc57e99e2b02ce73964c4eb2c8dbf71a9/internal/roomname.go#L157-L161) in the sync response.

pixlwave:
This PR makes 2 changes:
- Updates the storage of room heroes to be a single array containing the user's complete profile.
- Exposes these to the FFI so that client apps can use these for avatar colours/clusters.

I'm not 100% sure what to do about the difference between syncv2 and syncv3 returning different things, adding some thoughts inline.

Closes #2702 (again, now with avatars 🖼️)

### DIFF
--- a/crates/matrix-sdk-base/src/rooms/normal.rs
+++ b/crates/matrix-sdk-base/src/rooms/normal.rs
@@ -1148,7 +1148,6 @@ impl RoomInfo {
 
         if !summary.is_empty() {
             if !summary.heroes.is_empty() {
-                // Parse the user IDs from Ruma.
                 self.summary.room_heroes = summary
                     .heroes
                     .iter()
--- a/crates/matrix-sdk-base/src/rooms/normal.rs
+++ b/crates/matrix-sdk-base/src/rooms/normal.rs
@@ -1150,7 +1150,7 @@ impl RoomInfo {
             if !summary.heroes.is_empty() {
                 // Parse the user IDs from Ruma.
                 self.summary.room_heroes = summary
-                    .heroes // TODO: Load the profiles?
+                    .heroes
                     .iter()
                     .map(|hero_id| RoomHero {
                         user_id: hero_id.to_owned(),
--- a/crates/matrix-sdk-base/src/rooms/normal.rs
+++ b/crates/matrix-sdk-base/src/rooms/normal.rs
@@ -110,26 +110,36 @@ pub struct Room {
 /// calculate the room display name.
 #[derive(Clone, Debug, Default, Serialize, Deserialize)]
 pub struct RoomSummary {
-    /// The heroes of the room, members that should be used for the room display
-    /// name.
+    /// The heroes of the room, members that can be used as a fallback for the
+    /// room's display name or avatar if these haven't been set.
     ///
     /// This was called `heroes` and contained raw `String`s of the `UserId`
-    /// before; changing the field's name helped with avoiding a migration.
+    /// before. Following this it was called `heroes_user_ids` and a
+    /// complimentary `heroes_names` existed too; changing the field's name
+    /// helped with avoiding a migration.
     #[serde(default, skip_serializing_if = "Vec::is_empty")]
-    pub(crate) heroes_user_ids: Vec<OwnedUserId>,
-    /// The heroes names, as returned by a server, if available.
-    #[serde(default, skip_serializing_if = "Vec::is_empty")]
-    pub(crate) heroes_names: Vec<String>,
+    pub(crate) room_heroes: Vec<RoomHero>,
     /// The number of members that are considered to be joined to the room.
     pub(crate) joined_member_count: u64,
     /// The number of members that are considered to be invited to the room.
     pub(crate) invited_member_count: u64,
 }
 
+/// Information about a member considered to be a room hero.
+#[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
+pub struct RoomHero {
+    /// The user id of the hero.
+    pub user_id: OwnedUserId,
+    /// The display name of the hero.
+    pub display_name: Option<String>,
+    /// The avatar url of the hero.
+    pub avatar_url: Option<OwnedMxcUri>,
+}
+
 #[cfg(test)]
 impl RoomSummary {
-    pub(crate) fn heroes(&self) -> &[OwnedUserId] {
-        &self.heroes_user_ids
+    pub(crate) fn heroes(&self) -> &[RoomHero] {
+        &self.room_heroes
     }
 }
 
@@ -517,21 +527,26 @@ impl Room {
         // From here, use some heroes to compute the room's name.
         let own_user_id = self.own_user_id().as_str();
 
-        let (heroes, num_joined_guess): (Vec<String>, _) = if !summary.heroes_names.is_empty() {
-            // Straightforward path: pass through the heroes names, don't give a guess of
-            // the number of members.
-            (summary.heroes_names, None)
-        } else if !summary.heroes_user_ids.is_empty() {
-            // Use the heroes, if available.
-            let heroes = summary.heroes_user_ids;
-
-            let mut names = Vec::with_capacity(heroes.len());
-            for user_id in heroes {
-                if user_id == own_user_id {
+        let (heroes, num_joined_guess): (Vec<String>, _) = if !summary.room_heroes.is_empty() {
+            let mut names = Vec::with_capacity(summary.room_heroes.len());
+            for hero in &summary.room_heroes {
+                if hero.user_id == own_user_id {
                     continue;
                 }
-                if let Some(member) = self.get_member(&user_id).await? {
-                    names.push(member.name().to_owned());
+                if let Some(display_name) = &hero.display_name {
+                    names.push(display_name.clone());
+                    continue;
+                }
+                match self.get_member(&hero.user_id).await {
+                    Ok(Some(member)) => {
+                        names.push(member.name().to_owned());
+                    }
+                    Ok(None) => {
+                        warn!("Ignoring hero, no member info for {}", hero.user_id);
+                    }
+                    Err(error) => {
+                        warn!("Ignoring hero, error getting member: {}", error);
+                    }
                 }
             }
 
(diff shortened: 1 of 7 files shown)


crates/matrix-sdk-base/src/rooms/normal.rs
```diff
@@ -1099,11 +1122,13 @@ impl RoomInfo {
         if !summary.is_empty() {
             if !summary.heroes.is_empty() {
                 // Parse the user IDs from Ruma.
-                self.summary.heroes_user_ids = summary
-                    .heroes
+                self.summary.room_heroes = summary
+                    .heroes // TODO: Load the profiles?
```
pixlwave, line 1126: From what I can tell (please correct me if I'm wrong), because we're in the `RoomInfo` here, it isn't possible to load profiles for these users before storing them?

Maybe it would make sense to make `RoomHero` an enum in this instance with `user_id` and `user_profile` being the 2 fields so we can tell whether or not to load the profiles when computing the room name?

crates/matrix-sdk-base/src/rooms/normal.rs
```diff
@@ -627,25 +643,32 @@ impl Room {
...
-            let mut names = Vec::with_capacity(heroes.len());
-            for user_id in heroes {
-                if user_id == own_user_id {
-                    continue;
-                }
-                if let Some(member) = self.get_member(&user_id).await? {
-                    names.push(member.name().to_owned());
+        let (heroes, guessed_num_members): (Vec<String>, _) = if !summary.room_heroes.is_empty() {
+            // TODO: Fallback to the local part of the user_id if the display name is
+            // missing?
```
pixlwave, line 648: This seems to make sense to me, so long as we know for sure that the hero has a loaded profile.

crates/matrix-sdk-base/src/rooms/normal.rs
pixlwave: Alternatively, perhaps we could add another field `needs_profile` or similar to the struct and use that when computing the name, that might work better actually 🤔

crates/matrix-sdk-base/src/rooms/normal.rs
bnjbvr: How about something simpler like:

- iterate over all the heroes
- if hero == own user id, ignore it
- if hero has a name, use the name
- otherwise try to get a room member for the hero's user_id, and call `display_name()` (which will do the right thing: either use the display name from the member event, or use the localpart)

This way we can avoid the inner `if !hero_names.is_empty()` extra case handling below.

crates/matrix-sdk-base/src/rooms/normal.rs
bnjbvr: Why is not sufficient to have a `RoomHero` with only the user id set?

crates/matrix-sdk-base/src/rooms/normal.rs
```diff
@@ -1140,11 +1165,15 @@ impl RoomInfo {
...
-    pub(crate) fn update_heroes(&mut self, heroes: Vec<OwnedUserId>, names: Vec<String>) {
-        self.summary.heroes_user_ids = heroes;
-        self.summary.heroes_names = names;
+    pub(crate) fn update_heroes(&mut self, heroes: Vec<RoomHero>) {
+        self.summary.room_heroes = heroes;
+    }
+
+    /// The heroes for this room.
+    pub fn heroes(&self) -> &[RoomHero] {
+        self.summary.room_heroes.deref()
```
bnjbvr, line 1176: this smells like AI-generated code now :stuck_out_tongue: 

```suggestion
        &self.summary.room_heroes
```

crates/matrix-sdk-base/src/rooms/normal.rs
```diff
@@ -578,6 +589,11 @@ impl Room {
         Ok(members)
     }
 
+    /// Get the heroes for this room.
+    pub fn heroes(&self) -> Vec<RoomHero> {
+        self.inner.read().heroes().to_owned()
```
bnjbvr, line 594: Why `.to_owned()` here? If it is because the underlying return type is `&[...]`, can you use `to_vec()` to make this more obvious?

bnjbvr (changes requested):
You're on the right track! I've proposed a change for the room name computation, but this code was to change anyways because of the search filter issue…

crates/matrix-sdk-base/src/rooms/normal.rs
pixlwave: Nothing other than it seems weird to me that there's no obvious explanation for the storage containing
- Heroes with their profile when using Sliding Sync
- Heroes without their profile when using non-sliding sync.

I.e. when making a call to  `room.heroes()` you can't trust the name/avatar values and have to decide whether to load the profiles based upon if you're using sliding sync or not.

crates/matrix-sdk-base/src/rooms/normal.rs
pixlwave: > which will do the right thing

Ah, ha! That sounds like a good course of action for computing the room name.

crates/matrix-sdk-base/src/rooms/normal.rs
pixlwave: Ha, although whilst there's definitely an healthy amount of Copilot assistance in my PR, I'm pretty sure I wrote this one myself - I very much forget to `&thing` now and then look for a method to get the ref 🙈

crates/matrix-sdk-base/src/rooms/normal.rs
pixlwave: Oh, so unlike `&str` where `to_owned` is preferred over `to_string()` for arrays `to_vec` is preferred over `to_owned`?

bnjbvr: Hmm good question — I don't know about Clippy (who may disagree with me), but having `to_vec()` makes it more visible that it was a slice, because that's one of the few structs that has a `to_vec()` on it.

crates/matrix-sdk-base/src/rooms/normal.rs
bnjbvr: This only reflects a difference between sliding sync (will return avatar and display names for heroes by default) and non-sliding sync (will only return user ids). We could try to erase the difference here, but that would probably have to be happening on the receiving side, that is, when calling `Room::heroes()`, we oucld try to fill in the display name and the avatar URL only if it's missing.

pixlwave: Yeah that does make sense, although I was very much hoping to keep `Room::heroes` sync.

bnjbvr:
Hey @pixlwave, happy to do another review, but can you solve the merge conflicts first? Thanks!

crates/matrix-sdk-base/src/rooms/normal.rs
```diff
@@ -627,24 +642,28 @@ impl Room {
...
+                if let Some(display_name) = &hero.display_name {
+                    names.push(display_name.clone());
+                    continue;
+                }
+                match self.get_member(&hero.user_id).await {
+                    Ok(Some(member)) => {
+                        names.push(member.name().to_owned());
+                    }
+                    Ok(None) => {
+                        warn!("Ignoring hero, member not: {}", hero.user_id);
```
bnjbvr, line 660:
```suggestion
                        warn!("Ignoring hero, no member info for {}", hero.user_id);
```

crates/matrix-sdk-base/src/rooms/normal.rs
bnjbvr: What do you plan to do with this TODO?

bnjbvr (changes requested):
Actually, not that many changes, so we're close to getting this merged! I'd like to hear about your plans for the remaining TODO in the code.

crates/matrix-sdk-base/src/rooms/normal.rs
pixlwave: You can tell the exact point where I got distracted by something else here…

crates/matrix-sdk-base/src/rooms/normal.rs
pixlwave: I'm happy to simply remove it given it shouldn't affect us on EX (although that might not be the answer you were looking for). Longer term, filling in the missing profiles would work, so long as we could keep some form of sync access (I keep wondering if we could store `RoomInfo` on `Room` and have it automatically update it properties when anything changes).

pixlwave: I've rebased and removed it.

crates/matrix-sdk-base/src/rooms/normal.rs
```diff
@@ -1128,7 +1148,17 @@ impl RoomInfo {
 
         if !summary.is_empty() {
             if !summary.heroes.is_empty() {
-                self.summary.heroes_user_ids = summary.heroes.clone();
+                // Parse the user IDs from Ruma.
```
bnjbvr, line 1151: I think this comment is outdated now?

bnjbvr (approved):

crates/matrix-sdk-base/src/rooms/normal.rs
bnjbvr:
```suggestion
```

pixlwave: My bad, I missed that in the rebase 👍

pixlwave: Oh thanks for the update, much appreciated 🙏

PR #3499, opened 2024-06-03
Investigate the test_room_preview intermittent failure

bnjbvr:
(Don't review yet)

PR #3504, opened 2024-06-04
integration test: sync longer in `test_room_preview` before considering it stable

### LINKED ISSUE #3483: ci: `test_room_preview` intermittently fails
Full name: `tests::sliding_sync::room::test_room_preview`

See for instance: https://github.com/matrix-org/matrix-rust-sdk/actions/runs/9301293644/job/25599110490?pr=3482

bnjbvr:
It turns out that the failure came when using the known room path in the room preview: Alice knows about the room, but for some reason the client didn't retrieve all the state events from the sliding sync proxy yet.

Before, the sync would be considered stable after 2 seconds. This is too little, considering that events come from the proxy that listens to events from synapse. Raising this threshold to 15 seconds should help getting all the room information from the proxy, and thus get all the information we expected in the client.

Fixes #3483.

### DIFF
--- a/testing/matrix-sdk-integration-testing/src/tests/sliding_sync/room.rs
+++ b/testing/matrix-sdk-integration-testing/src/tests/sliding_sync/room.rs
@@ -1056,8 +1056,8 @@ async fn test_room_preview() -> Result<()> {
     let sliding_alice = alice
         .sliding_sync("main")?
         .with_all_extensions()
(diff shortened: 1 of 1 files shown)


andybalaam:
> Fixes #3499.

Wrong issue number?

andybalaam (approved):
Looks reasonable, thanks.

PR #3508, opened 2024-06-04
room preview integration test: explicitly request some state events to sync

bnjbvr:
Another attempt at fixing #3483.


andybalaam (approved):
Extremely plausible.

PR #3509, opened 2024-06-04
Configure unexpected-cfgs via Cargo.toml instead of build scripts

jplatte:
See https://blog.rust-lang.org/2024/05/06/check-cfg.html#expecting-custom-cfgs for docs.


Hywan (approved):
Great, thank you!

PR #3512, opened 2024-06-05
feat(ui): Propagate `RoomEventCache`'s `VectorDiff` to `Timeline`

Hywan:
WIP

---

* Address #3280

### DIFF
--- a/crates/matrix-sdk-ui/src/timeline/inner/mod.rs
+++ b/crates/matrix-sdk-ui/src/timeline/inner/mod.rs
@@ -73,8 +73,8 @@ use crate::{
 mod state;
 
 pub(super) use self::state::{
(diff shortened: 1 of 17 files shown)


crates/matrix-sdk-ui/src/timeline/inner/mod.rs
```diff
@@ -1244,6 +1315,17 @@ pub(super) struct HandleManyEventsResult {
     pub items_updated: u64,
 }
 
+impl Add for HandleManyEventsResult {
```
bnjbvr, line 1318: Please no. Explicit is better than implicit. This is not C++, let's not override operators when we don't have to.

Hywan: Where's the fun then? 😉

Hywan:
Will be reopened soon 😃

PR #3513, opened 2024-06-05
clippy: disallow useless `async`s

bnjbvr:
Using async when not required will increase compile times, and propagate async-ness to the callers, transitively.

See also the [lint description](https://rust-lang.github.io/rust-clippy/master/#/unused_async).

Since we only had a few false positives, I've enabled it by default. This PR is supposed to be squashed-and-merged, there's no point in keeping the individual commits, but these should help with the review.

(Note there's a renaming in the middle too.)


poljar (approved):
Good idea. Looks good.

Hywan:
Oh that's a nice lintç

PR #3514, opened 2024-06-05
FFI: Replace `base_path + user_id` with a `session_path`.

pixlwave:
This PR removes the concept of a base_path from the FFI. Instead we do the following:

1. Provide a direct path to the OIDC registrations.json file instead of building it - this was weird using a base_path directly in the main SDK crate.
2. Remove all the ClientBuilder logic that creates the store path from the base_path + user_id and instead have the caller generate one (using a UUID for new logins and saving it for restoration) themselves.
3. Remove the in-memory client from the authentication service, as now we don't need to worry about the whole "we don't know the final path until the login is complete".

Have tested with password, OIDC and QR logins.

### DIFF
--- a/bindings/matrix-sdk-ffi/src/authentication_service.rs
+++ b/bindings/matrix-sdk-ffi/src/authentication_service.rs
@@ -19,12 +19,9 @@ use matrix_sdk::{
         AuthorizationResponse, Oidc, OidcError,
     },
     reqwest::StatusCode,
-    AuthSession, ClientBuildError as MatrixClientBuildError, HttpError, RumaApiError,
+    ClientBuildError as MatrixClientBuildError, HttpError, RumaApiError,
 };
-use ruma::{
-    api::error::{DeserializationError, FromHttpResponseError},
-    OwnedUserId,
-};
+use ruma::api::error::{DeserializationError, FromHttpResponseError};
 use tokio::sync::RwLock as AsyncRwLock;
 use url::Url;
 use zeroize::Zeroize;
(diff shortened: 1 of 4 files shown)


Hywan (commented):
I'm uncomfortable that we have no test :-/. Can you try to add even basic ones?

poljar:
> I'm uncomfortable that we have no test :-/. Can you try to add even basic ones?

We discussed this in our Rust meeting and decided that it's fine to merge this without the requirement for additional tests since:

1. It removes a bunch of code and brings us closer to a 1:1 mapping of the SDK API.
2. It makes it much easier to move or even remove the `AuthenticationService` completely.

I think the only missing thing would be a bit of a higher level and observable OIDC login method so the application doesn't need to call 3-4 methods to complete an OIDC login. The QR login method is a good example of such an API.

pixlwave:
> 1. It removes a bunch of code and brings us closer to a 1:1 mapping of the SDK API.
> 2. It makes it much easier to move or even remove the AuthenticationService completely.

Yes, sorry I should have written in the PR description, that this was basically my plan. Step 1 is done in this PR, step 2 is to move the OIDC stuff into the SDK client so it can be tested and then (probably in a follow-up step 3) remove the authentication service.

pixlwave:
I can confirm, I've also successfully tested this for QR login too 👍

poljar:
> I can confirm, I've also successfully tested this for QR login too 👍

Nice, can we perhaps also get @jmartinesp or @bmarty to confirm that they are fine with this change, or even test it.

jmartinesp:
It seems to be working fine on Android after some extra code for the migration ✅ .

poljar:
> It seems to be working fine on Android after some extra code for the migration ✅ .

Nice, thanks for testing.

poljar (approved):
Lets move this along so we can eventually get rid of the `AuthService.

PR #3516, opened 2024-06-06
base: cache the computed display name and reuse that in the search filters

bnjbvr:
Could be reviewed commit by commit. It gets messy only at the end. GLHF.

The computed display name is filled in two cases:
- when creating a room for the first time,
- when getting a sync (from sync v3 or sliding sync) update for a room, in all cases.

This means that the first time we see a room during a sync, we'll compute the display name twice :shrug: 

Recomputing on every sync might be a bit wasteful, but getting all the inputs right when computing the display name is a tad complicated, since this depends on the room summary, the members, the heroes, the room name event and the canonical alias event (and maybe other things I'm forgetting about right now). Alternatives involved: using a dirty flag (bleh), also keeping all the inputs along the cached output (so we could early return when the inputs haven't changed in `compute_display_name` — could work, but seems frail), making the search filters async (nopenopenope).

### DIFF
--- a/crates/matrix-sdk-base/src/rooms/normal.rs
+++ b/crates/matrix-sdk-base/src/rooms/normal.rs
@@ -1658,7 +1658,7 @@ mod tests {
             .unwrap();
 
         let room_id = room_id!("!test:localhost");
-        let room = client.get_or_create_room(room_id, RoomState::Joined);
+        let room = client.get_or_create_room(room_id, RoomState::Joined).await;
 
         // Sanity checks to ensure the room isn't marked as favourite.
         assert!(room.is_favourite().not());
@@ -1732,7 +1732,7 @@ mod tests {
             .unwrap();
 
         let room_id = room_id!("!test:localhost");
-        let room = client.get_or_create_room(room_id, RoomState::Joined);
+        let room = client.get_or_create_room(room_id, RoomState::Joined).await;
 
         // Sanity checks to ensure the room isn't marked as low priority.
         assert!(!room.is_low_priority());
@@ -2179,7 +2179,7 @@ mod tests {
             .unwrap();
 
         let room_id = room_id!("!test:localhost");
-        let room = client.get_or_create_room(room_id, RoomState::Joined);
+        let room = client.get_or_create_room(room_id, RoomState::Joined).await;
 
         // That has an encrypted event,
         add_encrypted_event(&room, "$A");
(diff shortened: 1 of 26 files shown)


bnjbvr:
That's new, the CI task is not looking at the right version of the code…

https://github.com/matrix-org/matrix-rust-sdk/actions/runs/9399452164/job/25886897317?pr=3516

poljar:
> That's new, the CI task is not looking at the right version of the code…
> 
> https://github.com/matrix-org/matrix-rust-sdk/actions/runs/9399452164/job/25886897317?pr=3516

I think we had it a couple of times, it might happen if the base branch of your PR is too far off from main on the repo.

Try a rebase?

bnjbvr:
@poljar This is now ready for review \o/

crates/matrix-sdk/src/sync.rs
```diff
@@ -154,6 +154,19 @@ impl Client {
     ) -> Result<()> {
         let BaseSyncResponse { rooms, presence, account_data, to_device, notifications } = response;
 
+        {
```
poljar, line 157: I feel like this is a bit of the wrong place for this? Perhaps a separate method would be in place, something like `update_cached_values()` which we're doing to call before calling this handlers method.

crates/matrix-sdk-base/src/rooms/normal.rs
```diff
@@ -202,7 +206,13 @@ impl Room {
                 Self::MAX_ENCRYPTED_EVENTS,
             ))),
             roominfo_update_sender,
-        }
+            cached_display_name: Arc::new(SyncRwLock::new(None)),
+        };
+
+        // Refill the display name cache.
+        let _ = room.compute_display_name().await;
```
poljar, line 213: Since we restore all rooms at startup, this has the potential to slow down startup quite badly. I guess it won't in the nominal case where most of your rooms have names but still quite worrisome.

That's the primary reason why I suggested the `RoomInfo` for this cache instead.

poljar (commented):
Left some comments we should discuss.

crates/matrix-sdk/src/sync.rs
bnjbvr: You're right, I've spotted the first common place to both syncs, but this should rather be somewhere in the base crate, for it to make sense.

crates/matrix-sdk-base/src/rooms/normal.rs
bnjbvr: I think it's a code smell to persist derived/computed state, but oh well.

poljar: Hmm, but why? How would you otherwise avoid the slowdown? Perhaps a benchmark telling us that the slowdown isn't important would settle the discussion either way?

The only other idea would be a deferred filling out of the computed state. But that sounds worse to me.

bnjbvr: I've put it in the `RoomInfo` now.

poljar (approved):
Alright, I don't think I have any more complaints.

PR #3517, opened 2024-06-06
Embed device keys in Olm-encrypted messages

uhoreg:

Signed-off-by:


uhoreg:
A couple issues with the current approach:

- we don't seem to have the cross-signed device key available, so we need to re-cross-sign it every time, but that's an async operation, so it makes a bunch of things async.  Is there a better way?
- the function for cross-signing is only available within the crypto crate, but I'd rather do the signing in the storage crates so that we only need to do the signing once (and avoid some complexity), but I'm not sure what are the implications of changing the function from `pub(crate)`  to `pub`

poljar:
> * we don't seem to have the cross-signed device key available, so we need to re-cross-sign it every time, but that's an async operation, so it makes a bunch of things async.  Is there a better way?

We do, well we do if a `/keys/query` has been done. It's part of the `Device` in the store. You'll have the same async problem though, since the `Device` isn't cached. I think re-cross-signing the device keys is not a good idea. We should use this signature iff it has been published to the server and it has to be the exact same signature that was published to the server, otherwise we'll just confuse ourselves. We'll have multiple valid, albeit different signatures for the same device keys.

> * the function for cross-signing is only available within the crypto crate, but I'd rather do the signing in the storage crates so that we only need to do the signing once (and avoid some complexity), but I'm not sure what are the implications of changing the function from `pub(crate)`  to `pub`

I think that letting stores cross-sign should not happen at all, we don't control all store implementations and this is an obvious shift of responsiblity where store implementations need to care about cryptography to get things working correctly.

uhoreg:
I think that this is ready to be reviewed now.

crates/matrix-sdk-crypto/src/olm/account.rs
```diff
@@ -908,19 +908,18 @@ impl Account {
         identity_key: Curve25519PublicKey,
         one_time_key: Curve25519PublicKey,
         fallback_used: bool,
+        our_device_keys: DeviceKeys,
```
andybalaam, line 911: Please can we have an entry in the above doc comment for this param?

crates/matrix-sdk-crypto/src/olm/account.rs
```diff
@@ -980,6 +979,7 @@ impl Account {
         &self,
         device: &ReadOnlyDevice,
         key_map: &BTreeMap<OwnedDeviceKeyId, Raw<ruma::encryption::OneTimeKey>>,
+        our_device_keys: DeviceKeys,
```
andybalaam, line 982: Describe in the doc comment please.

crates/matrix-sdk-crypto/src/olm/session.rs
```diff
@@ -237,23 +233,17 @@ impl Session {
...
     /// * `pickle_mode` - The mode that was used to pickle the session, either
     /// an unencrypted mode or an encrypted using passphrase.
-    pub fn from_pickle(
-        user_id: OwnedUserId,
-        device_id: OwnedDeviceId,
-        our_identity_keys: Arc<IdentityKeys>,
-        pickle: PickledSession,
-    ) -> Self {
+    pub fn from_pickle(device_keys: DeviceKeys, pickle: PickledSession) -> Self {
+        // FIXME: assert that device_keys has curve25519 and ed25519 keys
```
andybalaam, line 237: Is this a FIXME we should do before merging?

crates/matrix-sdk-crypto/src/olm/session.rs
```diff
@@ -285,3 +275,70 @@ pub struct PickledSession {
...
+        let alice_device = ReadOnlyDevice::from_account(&alice);
+
+        let message = alice_session
+            .encrypt(&alice_device, "m.dummy", ToDeviceDummyEventContent::new(), None)
+            .await
+            .unwrap()
+            .deserialize()
+            .unwrap();
+
+        let content = if let ToDeviceEncryptedEventContent::OlmV1Curve25519AesSha2(c) = message {
```
andybalaam, line 320: This would be slightly easier to read with the `let ... else` pattern, which was recently added https://doc.rust-lang.org/beta/rust-by-example/flow_control/let_else.html

crates/matrix-sdk-crypto/src/olm/session.rs
```diff
@@ -285,3 +275,70 @@ pub struct PickledSession {
...
+    use serde_json::{self, Value};
+    use vodozemac::olm::{OlmMessage, SessionConfig};
+
+    use crate::{
+        identities::ReadOnlyDevice, olm::Account,
+        types::events::room::encrypted::ToDeviceEncryptedEventContent,
+    };
+
+    #[async_test]
+    async fn test_encryption_and_decryption() {
```
andybalaam, line 293: I'm finding it hard to work out what this test does. Could it have a more specific name, and/or could you add some comments? I often add "given, when, then" comments to explain what I'm doing, but here what I'm really looking for is which line of code is the actual thing we're testing?

andybalaam: Is it "Given some stuff, when we create an inbound session, then it contains the signed device keys"?

crates/matrix-sdk-crypto/src/session_manager/sessions.rs
```diff
@@ -514,6 +514,10 @@ impl SessionManager {
         let mut new_sessions: BTreeMap<&UserId, BTreeMap<&DeviceId, SessionInfo>> = BTreeMap::new();
         let mut store_transaction = self.store.transaction().await;
 
+        // Cache of our own device keys, so we only need to fetch it once, if
+        // we need it at all.
+        let mut our_device_keys: Option<DeviceKeys> = None;
```
andybalaam, line 519: Could we actually fetch the device keys up front to avoid starting off with this empty and filling it in later? It might make this easier to read but I assume there's a good reason we're doing it lazily?

crates/matrix-sdk-indexeddb/src/crypto_store/mod.rs
```diff
@@ -447,7 +447,10 @@ impl IndexeddbCryptoStore {
 
     /// Process all the changes and do all encryption/serialization before the
     /// actual transaction.
-    async fn prepare_for_transaction(&self, changes: &Changes) -> Result<PendingIndexeddbChanges> {
+    async fn prepare_for_transaction(
+        &self,
+        changes: &Changes,
+    ) -> Result<(PendingIndexeddbChanges, bool)> {
```
andybalaam, line 453: We need some some documentation of the return type here.

crates/matrix-sdk-indexeddb/src/crypto_store/mod.rs
```diff
@@ -534,7 +537,17 @@ impl IndexeddbCryptoStore {
...
         let mut device_store = indexeddb_changes.get(keys::DEVICES);
 
+        let account_info = self.get_static_account();
+        let mut clear_caches = false;
         for device in device_changes.new.iter().chain(&device_changes.changed) {
+            // If our own device key changes, we need to clear the session
+            // cache because the sessions contain a copy of our device key, and
+            // we want the sessions to use the new version.
+            if account_info.clone().is_some_and(|info| {
+                info.user_id == device.user_id() && info.device_id == device.device_id()
```
andybalaam, line 547: Should this be `!=`? I don't fully understand the code, but if we're looking for a changed device ID I'd expect to be looking for a different value here.

crates/matrix-sdk-indexeddb/src/crypto_store/mod.rs
```diff
@@ -534,7 +537,17 @@ impl IndexeddbCryptoStore {
 
         let mut device_store = indexeddb_changes.get(keys::DEVICES);
 
+        let account_info = self.get_static_account();
+        let mut clear_caches = false;
         for device in device_changes.new.iter().chain(&device_changes.changed) {
+            // If our own device key changes, we need to clear the session
+            // cache because the sessions contain a copy of our device key, and
+            // we want the sessions to use the new version.
+            if account_info.clone().is_some_and(|info| {
```
andybalaam, line 546: Do we need the `clone` here? I'd expect to be able to just inspect the value so could do it by reference.

crates/matrix-sdk-indexeddb/src/crypto_store/mod.rs
```diff
@@ -713,9 +726,13 @@ impl_crypto_store! {
 
         tx.await.into_result()?;
 
-        // all good, let's update our caches:indexeddb
-        for session in changes.sessions {
-            self.session_cache.add(session).await;
+        if clear_caches {
+            self.clear_caches().await;
+        } else {
```
andybalaam, line 731: Should this be inside an `else`? I'd expect to want to add the sessions to the cache even if we've just cleared the cache. I expect I'm misunderstanding what this is doing.

crates/matrix-sdk-sqlite/src/crypto_store.rs
```diff
@@ -804,7 +806,16 @@ impl CryptoStore for SqliteCryptoStore {
                     txn.set_kv("backup_version_v1", &serialized_backup_version)?;
                 }
 
+                let account_info = this.get_static_account();
                 for device in changes.devices.new.iter().chain(&changes.devices.changed) {
+                    // If our own device key changes, we need to clear the
+                    // session cache because the sessions contain a copy of our
+                    // device key.
+                    if account_info.clone().is_some_and(|info| {
+                        info.user_id == device.user_id() && info.device_id == device.device_id()
```
andybalaam, line 815: Same question about `!=`

crates/matrix-sdk-indexeddb/src/crypto_store/mod.rs
andybalaam: Ah, no - it's a new _key_ for the _same_ device, right?

crates/matrix-sdk-sqlite/src/crypto_store.rs
```diff
@@ -1377,6 +1409,42 @@ mod encrypted_tests {
         );
     }
 
+    #[async_test]
+    async fn cache_cleared_after_device_update() {
```
andybalaam, line 1413: Is there some test we can add to `cryptostore_integration_tests` for this? Then it would cover indexeddb, which I think is not covered so far in this PR.

andybalaam (commented):
This looks good, with some minor comments and questions scattered around.

My main question is: if we don't find the device keys in the store, we generate them. Shouldn't we save them to the store at that point, so we don't end up generating them multiple times and risking inconsistency as @poljar mentioned in his previous comment?

I don't feel confident to give this full approval so will ask @poljar to take a look.

poljar:
> My main question is: if we don't find the device keys in the store, we generate them.

Have not looked at the whole PR yet, but please take a look at our constructor:

https://github.com/matrix-org/matrix-rust-sdk/blob/abda95979691c4d51d6b7fc06a9777d9e32a795d/crates/matrix-sdk-crypto/src/machine.rs#L297-L311

We are guaranteed to have our own device keys and device in the store. We can add a method to the store called `get_own_device()` which will always return a device to make this clearer. Of course, this does not mean that we're guaranteed to have the cross-signing keys attached to our device.

crates/matrix-sdk-indexeddb/src/crypto_store/mod.rs
uhoreg: We need the `clone` because `is_some_and` takes over the `account_info`, and we're in a `for` loop. Though, taking another look at it, replacing `clone` with `as_ref` seems to make the borrow checker happy, and looks like a better thing to do.

crates/matrix-sdk-crypto/src/session_manager/sessions.rs
uhoreg: If there are no sessions to save, then we don't need to fetch the device keys.  So it saves us from doing work if it isn't needed.  I don't know how much work it is to fetch the device keys, so I don't know if it makes a difference.

crates/matrix-sdk-indexeddb/src/crypto_store/mod.rs
uhoreg: Yup, we only need to clear the cache if we get new device keys for our own device (e.g. if we get new cross-signing signatures).

crates/matrix-sdk-indexeddb/src/crypto_store/mod.rs
uhoreg: We don't know if the sessions were created using the old or the new device keys, so we don't want to cache them.  (In fact, they're probably created using the old device keys, since we hadn't saved the new device keys yet).

uhoreg:
@andybalaam I think that I've addressed your comments.  There's some CI issues that I still need to look into, but it should be reviewable otherwise.

crates/matrix-sdk-crypto/src/olm/session.rs
uhoreg: Right, I meant to put a comment about that in the PR.  I'm wondering if we should change the return value to a `Result`, or if we should just do an `unwrap` or `expect` and panic if the keys aren't there.  We should be pretty safe for the time being, but I'm a bit worried that if we change crypto systems in the future, this will come back and bite us.  I would lean towards just doing an `unwrap`/`expect`, but I wanted a second opinion.

andybalaam: When in doubt, prefer `Result`, I'd say.

crates/matrix-sdk-indexeddb/src/crypto_store/mod.rs
andybalaam: Makes sense - please could you add a comment explaining this?

crates/matrix-sdk-crypto/src/olm/session.rs
```diff
@@ -181,7 +181,7 @@ impl Session {
             #[cfg(feature = "experimental-algorithms")]
             EventEncryptionAlgorithm::OlmV2Curve25519AesSha2 => OlmV2Curve25519AesSha2Content {
                 ciphertext,
-                sender_key: self.our_identity_keys.curve25519,
+                sender_key: self.device_keys.curve25519_key().unwrap(),
```
andybalaam, line 184: Is this `unwrap` OK? If so, please can we have a comment explaining why?

andybalaam (commented):
Looks good from my point of view, when the not generating device key part is done. Thanks!

crates/matrix-sdk-sqlite/src/crypto_store.rs
uhoreg: I don't think we can add this to `cryptostore_integration_tests` because `memorystore` works differently, and always keeps the sessions in cache (the cache can't be cleared). (But yes, I should add a test for indexeddb -- once I figure out how to run the indexeddb tests locally)

crates/matrix-sdk-crypto/src/session_manager/sessions.rs
uhoreg: With the new function for getting our device from storage, I decided to just fetch the keys every time -- we're already fetching the account every time.

crates/matrix-sdk-crypto/src/olm/session.rs
uhoreg: It's safe, if [the FIXME](https://github.com/matrix-org/matrix-rust-sdk/pull/3517#discussion_r1636123693) gets fixed.  I'll wait for @poljar's review before changing anything here.

uhoreg:
@poljar I think this is ready for review from you now. Some notes

- It currently just uses the device keys from storage, and assumes that it will get updated after cross-signing is done, to include the cross-signing signatures. Is this true, or is there anything else that needs to be done? Is there a test that I should write to make sure that it gets updated?
- Can you give your opinion on https://github.com/matrix-org/matrix-rust-sdk/pull/3517#discussion_r1636123693 ?
- I'm currently not refreshing the sessions in memorystore with the device keys. Do I need to? Is memorystore used for any real purposes, or is it just for testing.

crates/matrix-sdk-sqlite/src/crypto_store.rs
andybalaam: Here is now I ran the tests locally here:

```
cd crates/matrix-sdk-indexeddb/
WASM_BINDGEN_TEST_TIMEOUT=2400 CARGO_TARGET_DIR=/tmp/andyb-rust-target wasm-pack test --firefox --headless -- --no-default-features --features e2e-encryption
```

poljar (changes requested):
I think the general direction is good, left some questions/nits/suggestions.

crates/matrix-sdk-crypto/src/olm/account.rs
```diff
@@ -1286,20 +1301,35 @@ impl Account {
...
 
                 // We didn't find a matching session; try to create a new session.
-                let result = match self.create_inbound_session(sender_key, prekey_message) {
-                    Ok(r) => r,
-                    Err(e) => {
-                        warn!("Failed to create a new Olm session from a pre-key message: {e:?}");
-                        return Err(OlmError::SessionWedged(sender.to_owned(), sender_key));
-                    }
+                // Try to get our own stored device keys.
+                let device_keys = store
```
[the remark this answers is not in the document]
uhoreg: Forgot to change this place to use it.

crates/matrix-sdk-sqlite/src/crypto_store.rs
```diff
@@ -905,9 +920,9 @@ impl CryptoStore for SqliteCryptoStore {
     }
 
     async fn get_sessions(&self, sender_key: &str) -> Result<Option<Arc<Mutex<Vec<Session>>>>> {
-        let account_info = self.get_static_account().ok_or(Error::AccountUnset)?;
-
         if self.session_cache.get(sender_key).is_none() {
+            let device_keys = self.get_own_device().await?.as_device_keys().clone();
```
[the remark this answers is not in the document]
uhoreg: I don't understand what you mean by "modify the SQLite query to load both pieces of data at the same time".  Which two pieces of data are you talking about?

Alternatively, what do you think of having the cryptostore cache our own device keys, so it doesn't need to fetch it from storage each time?  It's a fairly small bit of data, so should be fine to cache.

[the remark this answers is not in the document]
poljar: > I don't understand what you mean by "modify the SQLite query to load both pieces of data at the same time". Which two pieces of data are you talking about?

Ah sorry, I meant `SQL` query. The two pieces of data are our own device keys and the session pickle. The piece of code I'm commenting on is inside the `SQLite` store. If you look at what `get_sessions_for_sender_key()` does, it executes a SQL query to fetch the serialized and possibly encrypted binary data of the Olm sessions from the database:

https://github.com/matrix-org/matrix-rust-sdk/blob/6701fd0685f59cdb86a9e11cd4aa88328ab11f49/crates/matrix-sdk-sqlite/src/crypto_store.rs#L450-L456

If you follow down the stack what `get_device()` does, you'll end up here:

https://github.com/matrix-org/matrix-rust-sdk/blob/6701fd0685f59cdb86a9e11cd4aa88328ab11f49/crates/matrix-sdk-sqlite/src/crypto_store.rs#L545-L554

So I'm wondering if we could combine those two SQL queries into one.

> Alternatively, what do you think of having the cryptostore cache our own device keys, so it doesn't need to fetch it from storage each time? It's a fairly small bit of data, so should be fine to cache.

I'm in the process of removing caches from the individual cryptostore implementations, but perhaps that will be the only way to preserve the performance. But let's do that in a separate PR if it comes to that.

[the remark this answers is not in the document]
uhoreg: I've tried to fetch both things in one SQL query: https://github.com/matrix-org/matrix-rust-sdk/pull/3517/commits/2d660419e5280ccc791206ce59d1b4e79d34cb9b  Is this how you were thinking of doing it?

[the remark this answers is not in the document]
poljar: Yeah, though benchmarking this, sadly it doesn't seem to make a difference. Let's revert it then and keep the simpler implementation. If we want our performance back we'll have to introduce a cache for our own device, though let's not bother with that in this PR.

poljar (approved):
Left some final nits, we should revert the SQLite query patch since it doesn't help.

Approving ahead of time, feel free to merge once the final nits have been resolved.

crates/matrix-sdk-sqlite/src/crypto_store.rs
uhoreg: I tried that, but it gave me an error saying:

```
Try find `webdriver.json` for configure browser's capabilities:
Not found
driver status: signal: 9 (SIGKILL)
```

I wrote a test anyways and it seems to pass CI, so I'll ignore it for now, and try to figure it out later if needed.

uhoreg:
I think I've addressed all outstanding comments.  I don't have permissions to hit the merge button.  Can someone do so for me?

crates/matrix-sdk-crypto/src/olm/session.rs
```diff
@@ -156,11 +151,12 @@ impl Session {
...
             let payload = json!({
-                "sender": &self.user_id,
-                "sender_device": &self.device_id,
+                "sender": &self.our_device_keys.user_id,
+                "sender_device": &self.our_device_keys.device_id,
                 "keys": {
-                    "ed25519": self.our_identity_keys.ed25519.to_base64(),
+                    "ed25519": self.our_device_keys.ed25519_key().expect("Device doesn't have ed25519 key").to_base64(),
                 },
+                "device_keys": self.our_device_keys,
```
poljar, line 159: Doesn't this property need a different name:

>Until this MSC is accepted, the new property should be named org.matrix.msc4147.device_keys.

We can do this in a follow up, so we don't risk more merge conflicts.

poljar: https://github.com/matrix-org/matrix-rust-sdk/pull/3618

richvdh:
For some reason it doesn't seem to be mentioned in the PR :angry: but I believe this is implementing MSC4147.

Issue #3374, opened 2024-05-02
Posthog: stop double reporting UTD events when app is relaunched


richvdh:
Per https://github.com/element-hq/element-meta/issues/2333

richvdh:
This follows up on code added in https://github.com/matrix-org/matrix-rust-sdk/pull/3206. I think we want to add logic to `matrix-sdk-ui::unable_to_decrypt_hook::UtdHookManager` to filter out UTD errors that have already been reported the "parent" UTD hook; and we need to persist that data. As with https://github.com/element-hq/element-web/issues/27421, we probably want to use a Bloom filter to avoid the data getting too huge.

It's not quite the same as `known_utds`, because that is updated when we first see a UTD for a given event. At that point, the report has not yet been sent. If the application is restarted before the `max_delay` elapses, the task to actually send the report will be killed. In short, we shouldn't update the list of reported UTDs until the report is actually sent.

The main difficulty, I think, is going to be actually connecting `UtdHookManager` up to some persistent storage.

PR #3519, opened 2024-06-06
UTD hook: stop re-reporting of UTDs on app relaunch

richvdh:
This is a fix to #3374, and a re-do of #3490.

The principle remains to use a Bloom filter to keep track of which events we have already reported to the parent UTD hook. We load the data from database on startup; we now flush it out immediately on every update.


crates/matrix-sdk-ui/CHANGELOG.md
```diff
@@ -9,7 +9,14 @@ Breaking changes:
 Bug fixes:
 
 - `UtdHookManager` no longer re-reports UTD events as late decryptions.
-  ([#3840](https://github.com/matrix-org/matrix-rust-sdk/pull/3840))
+  ([#3480](https://github.com/matrix-org/matrix-rust-sdk/pull/3480))
```
richvdh, line 12: fixing a typo while I'm here :innocent:

BillCarsonFr:
FYI I tested this PR with Element X. No source breaking changes.

As required I tested logout, and it's not crashing.

Also after puting the app in background, killing and re-open it it's not reporting again `Error` events for UTD

image

bnjbvr (approved):
Very nice commits, lgtm, thanks

PR #3520, opened 2024-06-06
send queue: try to avoid a race when disabling then aborting a send

bnjbvr:
@ganfra reported that he got a "disabled" global status notification first, and that he couldn't cancel sending of the event at hand. In theory, one would need to be very lucky to be in that situation, but if the tokio runtime is very busy, it could be that the next `.await` at line 347 (before) is super slow. 

I couldn't trial this in a regression test, so I just introduced a test that mimics the steps to reproduce. It was passing before this PR, but probably would intermittently fail with a very low rate.


andybalaam (approved):
Looks good as far as I understand it :-)

bnjbvr:
(I introduced merge conflicts with myself \o/)

PR #3523, opened 2024-06-06
send queue: use an enablement status per room

bnjbvr:
This makes it so that getting any error (network or API error) will only disable the send queue for a single room. Callers are expected to `SendQueue::subscribe_errors()` to see whenever they get an error:

- if they get an error, and network is disconnected: disable the send queue
- if they get an error, but network is still here: either re-enable the send queue, or display the error somewhere to the user. (It's also reflected live in the timeline.)

Based on top of #3522 to avoid conflicts.

### DIFF
--- a/crates/matrix-sdk/src/send_queue.rs
+++ b/crates/matrix-sdk/src/send_queue.rs
@@ -22,7 +22,6 @@ use std::{
     },
 };
 
-use eyeball::{SharedObservable, Subscriber};
 use matrix_sdk_base::RoomState;
 use matrix_sdk_common::executor::{spawn, JoinHandle};
 use ruma::{
@@ -51,8 +50,13 @@ impl SendQueue {
         Self { client }
     }
 
+    #[inline(always)]
+    fn data(&self) -> &SendQueueData {
+        &self.client.inner.send_queue_data
+    }
+
     fn for_room(&self, room: Room) -> RoomSendQueue {
-        let data = &self.client.inner.send_queue_data;
+        let data = self.data();
 
         let mut map = data.rooms.write().unwrap();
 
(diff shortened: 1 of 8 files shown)


andybalaam (approved):
Looks good, thanks!

crates/matrix-sdk/src/send_queue.rs
```diff
@@ -72,51 +77,49 @@ impl SendQueue {
...
-            debug!("globally enabling send queue");
-            let rooms = self.client.inner.send_queue_data.rooms.read().unwrap();
-            // Wake up the rooms, in case events have been queued in the meanwhile.
-            for room in rooms.values() {
-                room.inner.notifier.notify_one();
-            }
-        }
-    }
+    pub fn set_enabled(&self, enabled: bool) {
+        debug!("globally {}abling send queue", if enabled { "en" } else { "dis" });
```
Hywan, line 90: Nitpicking: It creates a string allocation every time `debug!` is called. Changing for:

```rust
if enabled {
    debug!("globally enabling send queue");
} else {
    debug!("globally disabling send queue");
}
```

would create zero string allocation, and will use a static string. I think it's preferable :-].

Hywan (approved):
I think the code is great. One nitpicking though, otherwise it looks good.

crates/matrix-sdk/src/send_queue.rs
bnjbvr: I don't think it allocates, there's no reason the strings in the conditionals' bodies can't be static too.

bnjbvr: Replaced with         debug!(?enabled, "setting global send queue enablement");

PR #3525, opened 2024-06-10
chore(ui): Remove a useless clone

Hywan:
This patch removes a useless clone of `TimelineInner::items`. This clone was technically cheap because it's a `Vector` behind the scene, which is cheap to clone, but still, it's not necessary at all to clone it.

---

* Extracted from https://github.com/matrix-org/matrix-rust-sdk/pull/3512.


bnjbvr (approved):
Wasn't there a useless-clone lint in clippy…

PR #3527, opened 2024-06-10
chore(ui): Rewrite a bit `TimelineEventHandler::add`

Hywan:
This patch renames `TimelineEventHandler::add` to `add_item`.

This patch also removes the `should_add` argument. The `add_item` is called conditionally everytime. I believe this is much cleaner. Otherwise the method should have been called `maybe_add_item` with a return type or something that indicates whether the item is added. This patch makes it clear and remove one possible state in `add_item`.

---

* Extracted from https://github.com/matrix-org/matrix-rust-sdk/pull/3512.

### DIFF
--- a/crates/matrix-sdk-ui/src/timeline/event_handler.rs
+++ b/crates/matrix-sdk-ui/src/timeline/event_handler.rs
@@ -354,12 +354,14 @@ impl<'a, 'o> TimelineEventHandler<'a, 'o> {
                     self.handle_room_message_edit(re);
                 }
                 AnyMessageLikeEventContent::RoomMessage(c) => {
-                    self.add(should_add, TimelineItemContent::message(c, relations, self.items));
+                    if should_add {
+                        self.add_item(TimelineItemContent::message(c, relations, self.items));
+                    }
                 }
                 AnyMessageLikeEventContent::RoomEncrypted(c) => {
                     // TODO: Handle replacements if the replaced event is also UTD
                     let cause = UtdCause::determine(raw_event);
-                    self.add(true, TimelineItemContent::unable_to_decrypt(c, cause));
+                    self.add_item(TimelineItemContent::unable_to_decrypt(c, cause));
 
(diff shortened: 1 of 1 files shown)


crates/matrix-sdk-ui/src/timeline/event_handler.rs
```diff
@@ -846,11 +866,7 @@ impl<'a, 'o> TimelineEventHandler<'a, 'o> {
     }
 
     /// Add a new event item in the timeline.
-    fn add(&mut self, should_add: bool, content: TimelineItemContent) {
-        if !should_add {
-            return;
-        }
-
+    fn add_new_item(&mut self, content: TimelineItemContent) {
```
bnjbvr, line 869: Can we call this `add_item`? The `new` part sounds just a bit redundant, since we're `add`ing something…

crates/matrix-sdk-ui/src/timeline/event_handler.rs
```diff
@@ -846,11 +866,7 @@ impl<'a, 'o> TimelineEventHandler<'a, 'o> {
     }
 
     /// Add a new event item in the timeline.
-    fn add(&mut self, should_add: bool, content: TimelineItemContent) {
-        if !should_add {
-            return;
-        }
```
bnjbvr, line 852: I seem to recall I had done the same change at some point and was requested changes, but that makes two of us agreeing with it now, so let's do it :muscle:

bnjbvr (approved):

crates/matrix-sdk-ui/src/timeline/event_handler.rs
Hywan: :+1:

crates/matrix-sdk-ui/src/timeline/event_handler.rs
Hywan: \o/

PR #3531, opened 2024-06-11
feat (bindings): added APIs to save and load composer drafts

Velin92:
This PR is the first piece of the breakdown of the following PR: https://github.com/matrix-org/matrix-rust-sdk/pull/3439 where only the core functionalities of the feature are implemented, without addressing the issue of editing and replying to events that have not yet been paginated.

### DIFF
--- a/bindings/matrix-sdk-ffi/src/room.rs
+++ b/bindings/matrix-sdk-ffi/src/room.rs
@@ -4,7 +4,7 @@ use anyhow::{Context, Result};
 use matrix_sdk::{
     event_cache::paginator::PaginatorError,
     room::{power_levels::RoomPowerLevelChanges, Room as SdkRoom, RoomMemberRole},
-    RoomMemberships, RoomState,
+    ComposerDraft, RoomMemberships, RoomState,
 };
 use matrix_sdk_ui::timeline::{PaginationError, RoomExt, TimelineFocus};
 use mime::Mime;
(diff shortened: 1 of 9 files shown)


bindings/matrix-sdk-ffi/src/room.rs
```diff
@@ -698,6 +701,55 @@ impl Room {
...
+    /// Remove the `ComposerDraft` stored in the state store for this room.
+    pub async fn clear_composer_draft(&self) -> Result<(), ClientError> {
+        Ok(self.inner.clear_composer_draft().await?)
+    }
+
+    /// Load the reply details for the given event id.
+    ///
+    /// This will return an `InReplyToDetails` object that contains the details
+    /// which will either be ready or an error.
+    pub async fn load_reply_details(
```
bnjbvr, line 725: Can we remove this one from this PR? I'd like to sit down with you to assess if we can do this differently, because I'm not quite satisfied even with the changes I added the other day…

bnjbvr (changes requested):
looks good on principle, thanks for splitting the PR. Can we include some smoke tests for the new (sdk) `Room` functions, please? A single test would be sufficient, save/load/clear/re-load. Thanks!

bindings/matrix-sdk-ffi/src/room.rs
Velin92: done! now the pr only contains draft saving , clearing and loading

Issue #3529, opened 2024-06-11
crypto: events can be sent without an up-to-date member list causing UTDs


kegsay:
I've seen two bug reports from the same user on EXA which shows the following:
 - Get invited to a DM over federation
 - Accept the invite
 - Send some messages
 - Those messages are UTD.
 - ~45s later send more messages,  they are not UTD.

The SDK is _supposed_ to hit `/members` prior to sending encrypted events to ensure they have an up-to-date member list. This is done in https://github.com/matrix-org/matrix-rust-sdk/blob/0c97c85aea952887b3d7677932b96af492808a42/crates/matrix-sdk/src/room/futures.rs#L166-L168

I see no evidence of a `/members` hit when sending the event.

However, this is complicated by the several confounding factors:
 - the app itself also asks for `/members` and then immediately cancels the job, causing `kotlinx.coroutines.JobCancellationException: Job was cancelled; job=SupervisorJobImpl{Cancelling}@6f96c4a` - this is normal behaviour when you back out of a room quickly before it can fully load.
 - `sync_members` will not actually sync the members if the user is invited _and_ history visibility is `HistoryVisibility::WorldReadable | HistoryVisibility::Invited`
 - the SS request doesn't return the member event of the inviter in this scenario because it's before the join event and the client doesn't `required_state: [[m.room.member, *]]`.
 - the client is stuck unable to upload OTKs due to https://github.com/matrix-org/matrix-rust-sdk/issues/1415 which is causing `/keys/upload` to HTTP 400.

These factors mean there's several things which could be going wrong here:
 - Can calling [Room.members()](https://github.com/matrix-org/matrix-rust-sdk/blob/1510f87fa07ac0dd9ae8d229ef32cbc246cd5161/crates/matrix-sdk/src/room/mod.rs#L645) break the state machine such that it doesn't call it again when you send an event?
 - Is there a race condition where you can send an event after you join but before the rust SDK layer believes you are joined, so `sync_members` doesn't actually sync members? 
 - The SDK does eventually learn that the inviter is in the room, but how? SS isn't telling the SDK directly, so what mechanism is being used here? `/messages`? That doesn't make sense as it's historical state.
 - Does the inability to upload OTKs cause the state machine to terminate early or something?

There's a lot of moving parts here but I've done due diligence:
 - haproxy logs on the server show that `/members` is only hit one time, just after the `/join` (this was triggered by EXA)
[...]

PR #3533, opened 2024-06-11
deduplicating handler: handle cancellation by repeating the query, if needs be

bnjbvr:
Solves one potential issue spotted in https://github.com/matrix-org/matrix-rust-sdk/issues/3529. If the following sequence happened:

- request A for key K starts
- request B for key K wants to start, but it's deduplicated, so it waits for request A to finish
- request A gets aborted

Then, in the previous version of the code, request B would finish and assume that A successfully completed, while this is not the case.

This new version handles that by repeating the deduplicated query. **This assumes that it's fine to repeat the entire query**, which is not a light assumption, but looking at the uses of the deduplicating handler (/members, sending a read receipt for a room, reading the encryption state for a room), this should be fine.

### DIFF
--- a/crates/matrix-sdk/src/deduplicating_handler.rs
+++ b/crates/matrix-sdk/src/deduplicating_handler.rs
@@ -24,7 +24,7 @@ use tokio::sync::Mutex;
 
 use crate::{Error, Result};
 
-type DeduplicatedRequestMap<Key> = Mutex<BTreeMap<Key, Arc<Mutex<Result<(), ()>>>>>;
+type DeduplicatedRequestMap<Key> = Mutex<BTreeMap<Key, Arc<Mutex<Option<Result<(), ()>>>>>>;
 
 /// Handler that properly deduplicates function calls given a key uniquely
 /// identifying the call kind, and will properly report error upwards in case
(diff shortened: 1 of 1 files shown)


crates/matrix-sdk/src/deduplicating_handler.rs
```diff
@@ -24,7 +24,20 @@ use tokio::sync::Mutex;
...
 use crate::{Error, Result};
 
-type DeduplicatedRequestMap<Key> = Mutex<BTreeMap<Key, Arc<Mutex<Result<(), ()>>>>>;
+/// State machine for the state of a query deduplicated by the
+/// [`DeduplicatingHandler`].
+enum QueryState {
+    /// The query hasn't completed yet. This doesn't mean it hasn't *started*
+    /// yet, but rather that it couldn't get to completion: some
+    /// intermediate steps might have run.
+    NotFinishedYet,
```
poljar, line 33: `NotFinishedYet` sounds like we should wait and see if it'll get finished next time, if we're sure that this only happens only if we cancel the future, perhaps `Cancelled` would be a better name.

crates/matrix-sdk/src/deduplicating_handler.rs
```diff
@@ -55,31 +68,67 @@ impl<Key: Clone + Ord + std::hash::Hash> DeduplicatingHandler<Key> {
...
+                    // most one holder at the same time.
+                    self.run_code(key, code, &mut request_guard).await
+                }
+            };
         }
 
-        // Assume a successful request; we'll modify the result in case of failures
-        // later.
-        let request_mutex = Arc::new(Mutex::new(Ok(())));
+        // Start at the `None` state to indicate we haven't completed the request yet.
```
poljar, line 102: That's not the `None` state :sweat_smile:. I see now why you picked `NotFinishedYet`, though it still seems that we could say something like:

> Let's assume the cancelled state, if we succeed or fail we'll modify the result.

poljar (approved):
Left some minor comments, feel free to ignore if you disagree. Looks good. Thanks for tackling this so swiftly :muscle:.

crates/matrix-sdk/src/deduplicating_handler.rs
bnjbvr: Haha, remnant from a previous push. Thanks, I've used Cancelled in lieu of NotFinishedYet, and tweaked the comment accordingly.

(Also added a note about the `code` future which must be idempotent if run multiple times.)

PR #3534, opened 2024-06-11
feat (bindings): added an API to retrieve a loaded reply

Velin92:
second part of the #3439 breakdown

The context for this API, is for the composer preview an reply without the need of it being actively in the timeline.

### DIFF
--- a/bindings/matrix-sdk-ffi/src/timeline/content.rs
+++ b/bindings/matrix-sdk-ffi/src/timeline/content.rs
@@ -189,8 +189,8 @@ impl Message {
 
 #[derive(uniffi::Record)]
 pub struct InReplyToDetails {
-    event_id: String,
-    event: RepliedToEventDetails,
+    pub(crate) event_id: String,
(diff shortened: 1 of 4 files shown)


bindings/matrix-sdk-ffi/src/timeline/content.rs
```diff
@@ -189,8 +189,8 @@ impl Message {
 
 #[derive(uniffi::Record)]
 pub struct InReplyToDetails {
-    event_id: String,
-    event: RepliedToEventDetails,
+    pub(crate) event_id: String,
+    pub(crate) event: RepliedToEventDetails,
```
bnjbvr, line 193: Can it be a small constructor instead of two `pub(crate)` fields, please?

bnjbvr (approved):
Thanks! One small comment, and this is good to go!

bindings/matrix-sdk-ffi/src/timeline/content.rs
Velin92: sure! done!

PR #3535, opened 2024-06-11
ffi: call the initial events in the same task that listens to updates

bnjbvr:
This avoids a race condition where the caller hasn't set up the initial items or the listener, and the listener is called *before* the initial items have been used.

cc @kegsay

### DIFF
--- a/bindings/matrix-sdk-ffi/src/timeline/mod.rs
+++ b/bindings/matrix-sdk-ffi/src/timeline/mod.rs
@@ -132,24 +132,28 @@ impl Timeline {
 
 #[uniffi::export(async_runtime = "tokio")]
 impl Timeline {
-    pub async fn add_listener(
-        &self,
-        listener: Box<dyn TimelineListener>,
-    ) -> RoomTimelineListenerResult {
+    pub async fn add_listener(&self, listener: Box<dyn TimelineListener>) -> Arc<TaskHandle> {
         let (timeline_items, timeline_stream) = self.inner.subscribe_batched().await;
-        let timeline_stream = TaskHandle::new(RUNTIME.spawn(async move {
+
+        Arc::new(TaskHandle::new(RUNTIME.spawn(async move {
             pin_mut!(timeline_stream);
 
+            // It's important that the initial items are passed *before* we forward the
+            // stream updates, with a guaranteed ordering. Otherwise, it could
+            // be that the listener be called before the initial items have been
+            // handled by the caller. See #3535 for details.
+
+            // First, pass all the items as a reset update.
+            listener.on_update(vec![Arc::new(TimelineDiff::new(VectorDiff::Reset {
+                values: timeline_items,
+            }))]);
+
+            // Then forward new items.
             while let Some(diffs) = timeline_stream.next().await {
                 listener
                     .on_update(diffs.into_iter().map(|d| Arc::new(TimelineDiff::new(d))).collect());
             }
-        }));
-
-        RoomTimelineListenerResult {
-            items: timeline_items.into_iter().map(TimelineItem::from_arc).collect(),
-            items_stream: Arc::new(timeline_stream),
-        }
+        })))
     }
 
     pub fn retry_decryption(self: Arc<Self>, session_ids: Vec<String>) {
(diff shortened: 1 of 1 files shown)


bindings/matrix-sdk-ffi/src/timeline/mod.rs
```diff
@@ -132,24 +132,20 @@ impl Timeline {
...
-    ) -> RoomTimelineListenerResult {
+    pub async fn add_listener(&self, listener: Box<dyn TimelineListener>) -> Arc<TaskHandle> {
         let (timeline_items, timeline_stream) = self.inner.subscribe_batched().await;
-        let timeline_stream = TaskHandle::new(RUNTIME.spawn(async move {
+
+        Arc::new(TaskHandle::new(RUNTIME.spawn(async move {
             pin_mut!(timeline_stream);
 
+            let initial_items = timeline_items.into_iter().map(TimelineItem::from_arc).collect();
+            listener.on_initial_items(initial_items);
```
Hywan, line 142: Simply call on_update here would be similar, and it removes the race condition on the consumer side. Otherwise it's still possible to get on_update actually called before on_initial_items on the host/consumer side depending on how busy the runtime is I guess. By only calling/using on_update, we remove a lot of complexity on the host side, and fix the race condition.

Hywan (changes requested):

bindings/matrix-sdk-ffi/src/timeline/mod.rs
bnjbvr: How would that work? The two delegate methods have different signatures 🤔

kegsay: > it's still possible to get on_update actually called before on_initial_items on the host/consumer side depending on how busy the runtime is

Is this true? I'd have thought calling `on_initial_items` here would block until the consumer code had finished running?

kegsay (commented):
This should fix the race condition, but it does so in a rather subtle way at first glance.

This helps the rust side because before it was possible for `on_update` to be called _before_ the call to `timeline_items.into_iter().map(TimelineItem::from_arc).collect(),`. With this PR, that is no longer possible.

I had written:

> Because the consumer is constructing the timeline from these listener functions, this does not help the consumer side because the race can still happen

..but it can't because `on_initial_items` and `on_update` are called in the same task, so the FFI layer guarantees that `on_initial_items` happens before `on_update`. As a result, so long as the consumer is synchronously constructing the timeline, it's not possible to get into the race condition mentioned previously:

> I assumed that I should be using those items when they are returned, and then the listener will be called for any new ones going forwards, right? But in this test it looks like the listener fires with alice's join event before AddListener returns, and the Items don't contain alice's join event. At best this _should_ just cause alice's join to be at `timeline[0]` (the wrong location) but I replace the timeline based on `Items`, so I lose alice's join event.

bindings/matrix-sdk-ffi/src/timeline/mod.rs
Hywan: > How would that work? The two delegate methods have different signatures 🤔

You can wrap the initial values in a `TimelineDiff::Reset` or `TimelineDiff::Append`. Thoughts?

bnjbvr: > but on the consumer side

I don't buy this argument. This PR is fixing a real issue that's caused in the code we control. Fixing issues on the consumer side is outside our control, and while we can do our best to *limit* or *mitigate* damage by providing good APIs in the first place, this argument doesn't apply to the problem at hand here, IMO; if a consumer goes async-wild on their side, I'd rather consider it's on them and they might be shooting _themselves_ in the feet.

> You can wrap the initial values in a TimelineDiff::Reset or TimelineDiff::Append. Thoughts?

Sounds like a bit of API abuse to me, but I'll happily get rid of one new method to limit the API churn, so I'll consider this, thanks.

bnjbvr: It's actually great, thanks for the suggestion.

Hywan (approved):

Issue #3528, opened 2024-06-10
Use the same library name in the XCFramework / Swift bindings for all platforms


Johennes:
The static libraries inside the XCFramework are currently using a [different name](https://github.com/matrix-org/matrix-rust-sdk/blob/main/xtask/src/swift.rs#L103) for each platform. This appears to be problematic when using the XCFramework inside a Cocoa Pod[^1].

```
[!] Unable to install vendored xcframework `MatrixSDKFFI` for Pod `react-native-matrix-sdk` because it contains static libraries
with differing binary names: libmatrix_sdk_ffi, libmatrix_sdk_ffi_iossimulator, and libmatrix_sdk_ffi_macos.
```

Would it be possible to use the same filename for all `.a` files inside the XCFramework? I suspect the specific names have been set up for a reason but given that the files end up in different folders anyway I'm not sure where they would clash?

```
MatrixSDKFFI.xcframework
├── Info.plist
├── ios-arm64
│   ├── Headers
│   │   ├── ...
│   └── libmatrix_sdk_ffi.a
├── ios-arm64_x86_64-simulator
│   ├── Headers
│   │   ├── ...
│   └── libmatrix_sdk_ffi_iossimulator.a
└── macos-arm64_x86_64
    ├── Headers
    │   ├── ...
    └── libmatrix_sdk_ffi_macos.a
```

[^1]: I'm working on a React Native module where using Cocoapods, unfortunately, appears to be mandatory

stefanceriu:
Hi @Johennes, first of all, I'm sorry 😅

I don't think there's any particular reason for them to be named like that. That's what [cargo names them](https://github.com/matrix-org/matrix-rust-sdk/blob/main/xtask/src/swift.rs#L280) based on the triples we pass in and then that gets preserved into the final xcframework. 

We have no interest in providing CocoaPods support but we wouldn't be introducing an extra step to the xtask to name them properly, either by seprating them into tripple folders or as a post processing step on the resulting xcframework.

PR #3536, opened 2024-06-11
fix(bindings): use the same library name for all platforms in the xcframework

Johennes:
Fixes: #3528

### DIFF
--- a/xtask/src/swift.rs
+++ b/xtask/src/swift.rs
@@ -98,13 +98,13 @@ impl Platform {
         }
     }
 
-    /// The name of the library for the platform once all architectures are
-    /// lipo'd together.
-    fn lib_name(&self) -> &str {
+    /// The name of the subfolder in which to place the library for the platform
+    /// once all architectures are lipo'd together.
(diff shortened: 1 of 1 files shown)


poljar (commented):
This looks fine, but hard to double check without a Apple machine.

@stefanceriu can you double check if this is fine for you guys?

stefanceriu:
Yup yup, leave it with me, it might just take me a bit to get to it 👍

stefanceriu (approved):
Alrighty, I had a go at building the whole thing and it not only looks great but it also works 😁 

🚀 it and thank you for handling it yourself!

Johennes:
Thanks for reviewing it so quickly! 🙂

stefanceriu:
@Johennes made a new release here https://github.com/matrix-org/matrix-rust-components-swift/releases/tag/v1.1.68 if you want to try it out. Seems fine on our side fwiw.

Johennes:
Tested and works in RN as well. 🎉 

Thanks again!

PR #3537, opened 2024-06-12
crypto: Use the server name in the QR code login data

poljar:
Using the resolved homeserver URL causes problems if we need to inspect the well-known configuration of the homeserver, for example, if the server name is matrix.org, but the homeserver URL is server.matrix.org, the well-known might be only available for the former.

This is why we also need to receive the former, i.e. the server name in the QR code data.

### DIFF
--- a/crates/matrix-sdk-crypto/src/types/qr_login.rs
+++ b/crates/matrix-sdk-crypto/src/types/qr_login.rs
@@ -84,7 +84,7 @@ pub enum QrCodeModeData {
     Reciprocate {
         /// The homeserver the existing device is using. This will let the new
         /// device know which homeserver it should use as well.
-        homeserver_url: Url,
+        server_name: String,
     },
 }
 
@@ -203,14 +203,14 @@ impl QrCodeData {
                 QrCodeMode::Reciprocate => {
                     // 7. If the mode is 0x04, we attempt to read the two bytes for the length of
                     //    the homeserver URL.
-                    let homeserver_url_len = reader.read_u16::<BigEndian>()?;
+                    let server_name_len = reader.read_u16::<BigEndian>()?;
 
                     // 8. We read and parse the homeserver URL.
-                    let mut homeserver_url = vec![0u8; homeserver_url_len.into()];
-                    reader.read_exact(&mut homeserver_url)?;
-                    let homeserver_url = Url::parse(str::from_utf8(&homeserver_url)?)?;
+                    let mut server_name = vec![0u8; server_name_len.into()];
+                    reader.read_exact(&mut server_name)?;
+                    let server_name = String::from_utf8(server_name).map_err(|e| e.utf8_error())?;
 
-                    QrCodeModeData::Reciprocate { homeserver_url }
+                    QrCodeModeData::Reciprocate { server_name }
                 }
             };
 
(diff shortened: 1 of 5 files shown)


crates/matrix-sdk-crypto/src/types/qr_login.rs
```diff
@@ -327,6 +341,46 @@ mod test {
...
+
+        assert_eq!(
+            expected_rendezvous, data.rendezvous_url,
+            "The parsed rendezvous URL should match to the expected one",
+        );
+
+        assert_eq!(
+            data.mode(),
+            QrCodeMode::Reciprocate,
+            "The mode in the test bytes vector should be Login"
```
bnjbvr, line 370: Is this a bad copypasto and this should read `should be Reciprocate`?

crates/matrix-sdk-crypto/src/types/qr_login.rs
```diff
@@ -327,6 +341,46 @@ mod test {
...
+        );
+
+        assert_let!(
+            QrCodeModeData::Reciprocate { server_name } = data.mode_data,
+            "The parsed QR code mode should match to the expected one",
```
bnjbvr, line 375:
```suggestion
            "The parsed QR code mode should match the expected one",
```

examples/qr-login/src/main.rs
```diff
@@ -112,7 +112,7 @@ async fn login(proxy: Option<Url>) -> Result<()> {
 
     let data = QrCodeData::from_base64(input).context("Couldn't parse the base64 QR code data")?;
 
-    let QrCodeModeData::Reciprocate { homeserver_url } = &data.mode_data else {
+    let QrCodeModeData::Reciprocate { server_name: homeserver_url } = &data.mode_data else {
```
bnjbvr, line 115: nit: missed renaming here for the right-hand side

crates/matrix-sdk/src/authentication/qrcode/secure_channel.rs
```diff
@@ -50,7 +50,7 @@ impl SecureChannel {
     pub(super) async fn new(http_client: HttpClient, homeserver_url: &Url) -> Result<Self, Error> {
         let channel = RendezvousChannel::create_outbound(http_client, homeserver_url).await?;
         let rendezvous_url = channel.rendezvous_url().to_owned();
-        let mode_data = QrCodeModeData::Reciprocate { homeserver_url: homeserver_url.clone() };
+        let mode_data = QrCodeModeData::Reciprocate { server_name: homeserver_url.to_string() };
```
bnjbvr, line 53: Maybe a missed renaming for the `homeserver_url` parameter?

bnjbvr (approved):
Mostly nits, so approving. The code makes sense, I'm not too sure about the why of the change, but trust you here. (Alternatively, one could ask review from somebody who's more knowledgeable about the QR code login process.)

poljar:
> Mostly nits, so approving. The code makes sense, I'm not too sure about the why of the change, but trust you here. (Alternatively, one could ask review from somebody who's more knowledgeable about the QR code login process.)

The MSC got changed: https://github.com/matrix-org/matrix-spec-proposals/pull/4108/commits/87f8317a902cd7bc5c2d2d225f71021b3a509e2d. The commit attempts to clarify why as well, you need to be able to do the well-known discovery, which is guaranteed to only work if you get the server name.

crates/matrix-sdk/src/authentication/qrcode/secure_channel.rs
poljar: This one actually isn't, we pass in the resolved homeserver URL for this test method. Kinda abusing the API, but fine for now.

poljar: Added a comment explaining this.

crates/matrix-sdk-crypto/src/types/qr_login.rs
```diff
@@ -84,7 +84,7 @@ pub enum QrCodeModeData {
     Reciprocate {
         /// The homeserver the existing device is using. This will let the new
         /// device know which homeserver it should use as well.
-        homeserver_url: Url,
+        server_name: String,
```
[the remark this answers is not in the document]
poljar: I think we might want that but for now we'll keep the weak type to support both the server name and the homeserver URL the initial version used.

crates/matrix-sdk-crypto/src/types/qr_login.rs
poljar: Bad copypasta indeed :pinched_fingers:.

PR #3539, opened 2024-06-12
send queue: flag unrecoverable errors, and don't block a room sending queue on those

bnjbvr:
Part of #3361.

This attempts to flag requests that are transient failures (recoverable) vs permanent failures (unrecoverable), using the same logic we have in the HTTP client for retrying an HTTP request.

The rules in the send queue have changed to the following:

- if an error was transient, but we still failed multiple times, this is likely a sign that something's wrong in the environment (network has disappeared, server is failing, etc.), so this disables the sending queue, to avoid wasting the user's battery to death, assuming subsequent requests would fail in the same way.
- if an error was permanent, the event queued in the send queue is marked as wedged, and it is ignored for now on (it doesn't move position in the send queue, so the corresponding timeline items won't move either). The send queue still processes subsequent queued events in the order they appear, and is not disabled by such an event.


bnjbvr:
@stefanceriu @ganfra here's a new version of the PR to try out, please! In particular, I'm curious if network errors are properly flagged as recoverable by the flagging mechanism (and thus disable the send queue), which is a bit opaque at the moment as it tries to look through a cascade of error types.

ganfra:
@bnjbvr so I'm afraid network errors are marked as unrecoverable :/

bnjbvr:
Latest commit seem to better handle unreachable networks (tested in multiverse, in an integration test where the server is unreachable, and on EXA thanks to @ganfra), so this is ready for review :partying_face:

crates/matrix-sdk/src/send_queue.rs
```diff
@@ -353,30 +366,68 @@ impl RoomSendQueue {
...
-
-                    // Let observers know about a failure *after* we've marked the item as not
-                    // being sent anymore. Otherwise, there's a possible race where a caller might
-                    // try to remove an item, while it's still marked as being sent, resulting in a
-                    // cancellation failure.
-
-                    // Disable the queue for this room after an error.
-                    locally_enabled.store(false, Ordering::SeqCst);
+                    let (err, is_recoverable) = if let crate::Error::Http(http_err) = err {
+                        // A small trick to confuse borrowck (and humans too).
```
poljar, line 370: Hmm, this smells. Aren't you in both cases where `characterize_retry_kind()` is used (which is clearly a contstructor for `RetryKind`, not sure why it's not under an `impl` block for `RetryKind`) immediately deconstructing or taking the error out of `RetryKind`?

Why does `RetryKind` then take ownership of the error in the first place? Furthermore, if we don't take ownership wouldn't something like this work much better:

```rust
pub(crate) enum RetryKind {
    Transient {
        err: HttpError,
        // This is used only for attempts to retry, so on non-wasm32 code (in the `native` module).
        #[cfg_attr(target_arch = "wasm32", allow(dead_code))]
        retry_after: Option<Duration>,
    },
    Permanent(HttpError),
}

impl HttpError {
    pub(crate) fn retry_kind(&self) -> RetryKind {
        ... body of `characterize_retry_kind()`
    }
}
```

Granted, `RetryKind` starts to look suspiciously similar to `RetryError` but perhaps we just don't want to include that type in our public API.

crates/matrix-sdk/src/send_queue.rs
```diff
@@ -353,30 +366,68 @@ impl RoomSendQueue {
...
+                            is_recoverable =
+                                matches!(http_err, crate::error::HttpError::Reqwest(..));
+                        }
+
+                        (crate::Error::Http(http_err), is_recoverable)
+                    } else {
+                        (err, false)
+                    };
+
+                    if is_recoverable {
```
poljar, line 395: Can't we move the whole `is_recoverable` logic under the `HttpError`? This way this becomes:

```rust
if http_err.is_recoverable() {
```

Which probably makes it easier to define what a recoverable error is and test as well.

poljar (changes requested):
Left a bit of a bigger comment, but I think it makes sense.

crates/matrix-sdk/src/send_queue.rs
poljar: Edited my comment, I forgot to remove the error field from `RetryKind`. Should now be possible.

bnjbvr: Great idea, thanks, done.

crates/matrix-sdk/src/send_queue.rs
bnjbvr: I didn't introduce a new method as now checking for a recoverable error really is checking that the RetryKind is Transient :smiling_face_with_three_hearts:

crates/matrix-sdk/src/send_queue.rs
```diff
@@ -353,30 +362,49 @@ impl RoomSendQueue {
...
-                    locally_enabled.store(false, Ordering::SeqCst);
+                    let is_recoverable = if let crate::Error::Http(ref http_err) = err {
+                        // All transient errors are recoverable.
+                        matches!(http_err.retry_kind(), RetryKind::Transient { .. })
+                    } else {
+                        false
+                    };
+
+                    if is_recoverable {
+                        warn!(txn_id = %queued_event.transaction_id, error = ?err, "recoverable error when sending event: {err}, disabling send queue");
```
poljar, line 373:
```suggestion
                        warn!(txn_id = %queued_event.transaction_id, error = ?err, "Recoverable error when sending event: {err}, disabling send queue");
```

crates/matrix-sdk/src/send_queue.rs
```diff
@@ -353,30 +362,49 @@ impl RoomSendQueue {
...
+                        // might try to remove an item, while it's still
+                        // marked as being sent, resulting in a cancellation
+                        // failure.
+
+                        // Disable the queue for this room after a recoverable error happened. This
+                        // should be the sign that this error is temporary (maybe network
+                        // disconnected, maybe the server had a hiccup).
+                        locally_enabled.store(false, Ordering::SeqCst);
+                    } else {
+                        warn!(txn_id = %queued_event.transaction_id, error = ?err, "unrecoverable error when sending event: {err}");
```
poljar, line 390:
```suggestion
                        warn!(txn_id = %queued_event.transaction_id, error = ?err, "Unrecoverable error when sending event: {err}");
```

poljar (approved):
Left some two final nits, looks good otherwise.

Issue #3079, opened 2024-02-01
[Meta] `SlidingSyncRoom` tech debt


Hywan:
We have a non-neligeable tech debt about `matrix_sdk::sliding_sync::SlidingSyncRoom`.

It holds data that are not shared with the `matrix_sdk::Room`. Ideally, we want `SlidingSyncRoom` to _not_ exist.

The idea is to clean this as much as possible.

---

Blocked by:

* We need the event cache (#3058) to have a storage, so that we can get rid off `SlidingSyncRoom::timeline_queue`,
* [x] https://github.com/matrix-org/matrix-rust-sdk/pull/3166
* [x] https://github.com/matrix-org/matrix-rust-sdk/pull/3512

PR #3540, opened 2024-06-13
Fix UI roomlist latest event

Hywan:
This patch is twofold:

* It updates `matrix_sdk_ui::room_list_service::Room::latest_event` to fetch the latest event from `matrix_sdk::Room` instead of `matrix_sdk::sliding_sync::SlidingSyncRoom`. It removes one dependency to `SlidingSyncRoom` and it simplifies the code.
* It removes the `matrix_sdk_ui::timeline::SlidingSyncRoomExt` trait 🎉.

---

* Address https://github.com/matrix-org/matrix-rust-sdk/issues/3079


crates/matrix-sdk-ui/src/room_list_service/room.rs
```diff
@@ -154,22 +154,32 @@ impl Room {
     /// contains a local event. Otherwise, it comes from the cache. This method
     /// does not fetch any events or calculate anything — if it's not
     /// already available, we return `None`.
+    ///
+    /// Reminder: this method also returns `None` is the latest event is not
+    /// suitable for use in a message preview.
     pub async fn latest_event(&self) -> Option<EventTimelineItem> {
         // Look for a local event in the `Timeline`.
```
bnjbvr, line 161: nit: this comment needs a change too, now it's local or remote.

crates/matrix-sdk-ui/src/room_list_service/room.rs
```diff
@@ -154,22 +154,32 @@ impl Room {
     /// contains a local event. Otherwise, it comes from the cache. This method
```
bnjbvr, line 154: nit: comment needs an update here too.

We could also improve the comment by saying that the timeline may contain more info than the cache, since it may have ran some back-pagination (as I was wondering why we would use the latest event from the timeline rather than the cache).

bnjbvr (approved):
Fare thee well `SlidingSyncRoomExt`! This refactoring makes me happy, thanks.

crates/matrix-sdk-ui/src/room_list_service/room.rs
Hywan: Indeed, thanks.

crates/matrix-sdk-ui/src/room_list_service/room.rs
Hywan: Yup!

PR #3541, opened 2024-06-13
chore(ffi): Remove `RoomInfo::latest_event`

Hywan:
This patch removes `RoomInfo::latest_event`. To get the latest event,
one has to use `RoomListItem::latest_event` because it supports
local events and remote events. It was supposed to be the case of
`Room::room_info` too, but only when the method was called: Once the
`RoomInfo` was created, the latest event inside `RoomInfo`
was static, not dynamic. Also, this code wasn't tested
contrary to `RoomListItem::latest_event` which uses
`matrix_sdk_ui::room_list_service::Room::latest_event` which is itself
tested.

---

* Based on https://github.com/matrix-org/matrix-rust-sdk/pull/3540. It must be merged after #3540 is merged. Please review only the last commit.

### DIFF
--- a/bindings/matrix-sdk-ffi/src/room.rs
+++ b/bindings/matrix-sdk-ffi/src/room.rs
@@ -32,7 +32,7 @@ use crate::{
     room_info::RoomInfo,
     room_member::RoomMember,
     ruma::{ImageInfo, Mentions, NotifyType},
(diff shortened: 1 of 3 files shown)


bnjbvr (approved):
yay! (Please make sure to address the review comments on the base PR.)

PR #3551, opened 2024-06-13
feat(ui): `RoomListService::room` is no longer async!

Hywan:
This patch is twofold:

* it updates `room_list_service::Room::new` to take a `&Client` and a `&RoomId` instead of a `SlidingSyncRoom`. The `SlidingSyncRoom` is only used in `Room::default_room_timeline_builder` and is fetched there from the `SlidingSync` instance lazily. It confines the `SlidingSyncRoom` to one single method for `Room` now.
* it makes `RoomListService::room` synchronous. It no longer reads a `SlidingSyncRoom` from `SlidingSync`, then it not needs to be async anymore. This patch replaces the `RwLock` of `RoomListService::rooms` from `tokio::sync` to `std::sync`.

--- 

* Address https://github.com/matrix-org/matrix-rust-sdk/issues/3079

### DIFF
--- a/crates/matrix-sdk-ui/src/room_list_service/mod.rs
+++ b/crates/matrix-sdk-ui/src/room_list_service/mod.rs
@@ -69,7 +69,7 @@ mod state;
 use std::{
     future::ready,
     num::NonZeroUsize,
-    sync::{Arc, RwLock},
+    sync::{Arc, Mutex as StdMutex},
     time::Duration,
 };
 
@@ -114,7 +114,7 @@ pub struct RoomListService {
     state: SharedObservable<State>,
 
     /// Room cache, to avoid recreating `Room`s every time users fetch them.
-    rooms: Arc<RwLock<RingBuffer<Room>>>,
+    rooms: Arc<StdMutex<RingBuffer<Room>>>,
 
     /// The current viewport ranges.
     ///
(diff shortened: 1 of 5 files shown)


crates/matrix-sdk-ui/src/room_list_service/mod.rs
```diff
@@ -435,7 +437,7 @@ impl RoomListService {
 
         let room = Room::new(&self.client, room_id, &self.sliding_sync)?;
 
-        self.rooms.write().await.push(room.clone());
+        self.rooms.write().unwrap().push(room.clone());
```
bnjbvr, line 440: Preexisting: this has one subtle race conditions, when two callers call around the same time and there's no cache hit in `self.rooms`. After taking the write-lock, you need to check for the room being there again, probably before calling `Room::new()`.

An alternative is also to take the lock as writable from the beginning, and keep it for the entire function's body.

bnjbvr (approved):
IT'S HAPPENING :partying_face: 

I spotted a potential race condition that we could fix while we're touching this code, can you fix it before merging please?

crates/matrix-sdk-ui/src/room_list_service/mod.rs
Hywan: You're right! Let's change the `RwLock` for a `Mutex` then. It's now cheap to create a `Room`, which wasn't the case before because of the `.await` point.

PR #3068, opened 2024-01-29
feat(base): Implement `Client::rooms_stream`

Hywan:
This patch defines the basis to get client-side sorting inside the `RoomList`. The first step is to get a `Stream` of `Room` from a `Client`, and this is what this patch does! It's only for non-wasm targets.

It's built on top of https://github.com/matrix-org/matrix-rust-sdk/pull/3561 to avoid Clippy warnings.

---

Blocked by:

* https://github.com/jplatte/eyeball/pull/43
* ~https://github.com/matrix-org/matrix-rust-sdk/issues/3079~

PR #3552, opened 2024-06-13
feat(base): Make `Store::get_rooms` wayyy faster

Hywan:
`Store::get_rooms` worked as follows: For each room ID, call

* Take the read lock for `rooms`,
* For each entry in `rooms`, take the room ID (the key), then call `Store::get_room`, which…
* Take the read lock for `rooms`,
* Based on the room ID (the key), read the room (the value) from `rooms`.

So calling `get_rooms` calls `get_room` for each room, which takes the lock every time.

This patch modifies that by fetching the values (the rooms) directly from `rooms`, without calling `get_room`.

In my benchmark, it took 1.2ms to read 10'000 rooms. Now it takes 542µs. Performance time has improved by -55.8%.

---

* Extracted from https://github.com/matrix-org/matrix-rust-sdk/pull/3068

### DIFF
--- a/crates/matrix-sdk-base/src/store/mod.rs
+++ b/crates/matrix-sdk-base/src/store/mod.rs
@@ -142,6 +142,7 @@ pub(crate) struct Store {
     session_meta: Arc<OnceCell<SessionMeta>>,
     /// The current sync token that should be used for the next sync call.
     pub(super) sync_token: Arc<RwLock<Option<String>>>,
+    /// All rooms the store knows about.
     rooms: Arc<StdRwLock<BTreeMap<OwnedRoomId, Room>>>,
     /// A lock to synchronize access to the store, such that data by the sync is
     /// never overwritten.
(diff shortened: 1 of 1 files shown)


bnjbvr (approved):
Nice! (Would be nice to have a formal benchmark for this, that we can run with `cargo-bench`, but the change is so obvious and trivial…)

Can you also fix `get_rooms_filtered`, while you're at it, please? It does something similar.

Hywan:
`get_rooms_filtered` is updated as part of #3068 (this commit is extracted from #3068) with a `filter_map` and so on, it's a bit different. Let's focus on `get_rooms` for the moment.

About the benchmark, I've sadly lost it! Never run `git stash clear` without being careful… I'll re-add it later with #3068 I believe.

Hywan:
`rooms_filtered ` is fixed in https://github.com/matrix-org/matrix-rust-sdk/pull/3554

PR #3553, opened 2024-06-13
Allow edit and reply to work also for events that have not yet been paginated

### LINKED ISSUE #3430: Store drafts for each room so that clients can restore them at a later time

### LINKED ISSUE #3538: Sending a reply or an edit event for an item that has not been paginated yet fails.
This is a limitation that has been found with the following issue: https://github.com/matrix-org/matrix-rust-sdk/issues/3430 and may impact the UX of the draft, we already have a PR that contains a possible fix but needs to be iterated upon.

Velin92:
fixes #3538 

The current implementation for send_reply and edit only work with timeline items that have already been paginated.
However given the fact that by restoring drafts, we may restore a reply to an event for timeline where such event has not been paginated, sending such reply would fail (same for the edit event).

So I reworked a bit the code here to use. only the event id, and reuse the existing timeline if available, otherwise we can fetch the event and synthethise the content and still be able to successfully send the event.

This is the third part of the breakdown of the following PR: https://github.com/matrix-org/matrix-rust-sdk/pull/3439


Hywan:
@Velin92 Can you explain in the PR description what the code does, and why please?

Hywan:
> The current implementation for send_reply and edit only work with timeline items that have already been paginated.

Do you mean it's been loaded via sync or via the pagination?

Are you trying to edit or to reply to an event that (i) is a remote event (not a local event, (ii) can be absent from the `Timeline`?

Velin92:
> > The current implementation for send_reply and edit only work with timeline items that have already been paginated.
> 
> Do you mean it's been loaded via sync or via the pagination?
> 
> Are you trying to edit or to reply to an event that (i) is a remote event (not a local event, (ii) can be absent from the `Timeline`?

not been loaded yet in the pagination, so yeah it's absent from the timeline, and yeah is a remote event also.

bnjbvr (changes requested):

crates/matrix-sdk-ui/src/timeline/mod.rs
```diff
@@ -309,76 +341,110 @@ impl Timeline {
...
+        let Some(timeline_item_content) =
+            TimelineItemContent::from_latest_event_content(sync_event.clone())
+        else {
+            return Err(UnsupportedReplyItem::MISSING_CONTENT);
+        };
+
+        let reply_content = match timeline_item_content {
+            TimelineItemContent::Message(msg) => ReplyContent::Message(msg.to_owned()),
+            _ => ReplyContent::Raw(raw_sync_event),
+        };
```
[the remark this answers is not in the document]
Velin92: Okay I think a found a better solution, can you check the code again?

crates/matrix-sdk-ui/src/timeline/mod.rs
```diff
@@ -309,76 +341,120 @@ impl Timeline {
...
     ///   part of a thread
-    #[instrument(skip(self, content, reply_item))]
+    #[instrument(skip(self, content, replied_to_info))]
     pub async fn send_reply(
         &self,
         content: RoomMessageEventContentWithoutRelation,
-        reply_item: &EventTimelineItem,
+        replied_to_info: RepliedToInfo,
         forward_thread: ForwardThread,
     ) -> Result<(), SendEventError> {
```
bnjbvr, line 363: Note: it looks like this could now return only `Result<(), RoomSendQueueError>`; let's open a follow-up PR to determine if that's true.

crates/matrix-sdk-ui/src/timeline/event_item/mod.rs
```diff
@@ -421,6 +422,47 @@ impl EventTimelineItem {
             kind,
         }
     }
+
+    /// Gives the information needed to reply to the event of the item.
+    pub fn get_replied_to_info(&self) -> Result<RepliedToInfo, UnsupportedReplyItem> {
```
bnjbvr, line 427: nit: let's get rid of the `get_` prefix everywhere (it's not very idiomatic in Rust)

crates/matrix-sdk-ui/src/timeline/event_item/mod.rs
```diff
@@ -421,6 +422,47 @@ impl EventTimelineItem {
...
+        Ok(RepliedToInfo {
+            event_id: event_id.to_owned(),
+            sender: self.sender().to_owned(),
+            timestamp: self.timestamp(),
+            content: reply_content,
+        })
+    }
+
+    /// Gives the information needed to edit the event of the item.
+    pub fn get_edit_info(&self) -> Result<EditInfo, UnsupportedEditItem> {
```
bnjbvr, line 452: ditto

crates/matrix-sdk-ui/src/timeline/mod.rs
```diff
@@ -309,76 +341,120 @@ impl Timeline {
...
+
+        let event = match self.room().event(event_id).await {
+            Ok(event) => event,
+            Err(error) => {
+                error!("Failed to fetch event with ID {event_id} with error: {error}");
+                return Err(UnsupportedReplyItem::MISSING_EVENT);
+            }
+        };
+
+        // We need to synthetise the TimelineitemContent and we can do that by casting
```
bnjbvr, line 422: Outdated comment here.

crates/matrix-sdk-ui/src/timeline/mod.rs
```diff
@@ -437,6 +505,58 @@ impl Timeline {
...
+
+        let event = match self.room().event(event_id).await {
+            Ok(event) => event,
+            Err(error) => {
+                error!("Failed to fetch event with ID {event_id} with error: {error}");
+                return Err(UnsupportedEditItem::MISSING_EVENT);
+            }
+        };
+
+        // We need to synthetise the TimelineitemContent and we can do that by casting
```
bnjbvr, line 525: here too?

crates/matrix-sdk-ui/src/timeline/mod.rs
```diff
@@ -309,76 +341,120 @@ impl Timeline {
...
+                return Err(UnsupportedReplyItem::MISSING_EVENT);
+            }
+        };
+
+        // We need to synthetise the TimelineitemContent and we can do that by casting
+        // the event as a `AnySyncTimelineEvent` which is the same as a
+        // `AnyTimelineEvent`, but without the `room_id` field.
+        // The cast is valid because we are just losing track of such field.
+        let raw_sync_event: Raw<AnySyncTimelineEvent> = event.event.cast();
+        let Ok(sync_event) = raw_sync_event.deserialize_as::<AnySyncTimelineEvent>() else {
```
bnjbvr, line 427: nit: instead of using `deserialize_as` with the explicit generic type, you can use `deserialize()` here.

crates/matrix-sdk-ui/src/timeline/mod.rs
```diff
@@ -309,76 +341,120 @@ impl Timeline {
...
+            return Err(UnsupportedReplyItem::FAILED_TO_DESERIALIZE_EVENT);
```
bnjbvr, line 428: Can you log the deserialization error message here, too, please?

crates/matrix-sdk-ui/src/timeline/mod.rs
```diff
@@ -309,76 +341,120 @@ impl Timeline {
...
+        };
+
+        let reply_content = match &sync_event {
+            AnySyncTimelineEvent::MessageLike(message_like_event) => {
```
bnjbvr, line 432: You can probably match on `AnySyncTimelineEvent::MessageLike(AnySyncMessageLikeEvent::RoomMessage(SyncRoomMessageEvent::Original(original_message))) =>` for the full message, and then use a `_` => at the end with the `ReplyContent::Raw` arm.

crates/matrix-sdk-ui/src/timeline/mod.rs
```diff
@@ -388,24 +464,16 @@ impl Timeline {
...
-    /// * `edit_item` - The event item you want to edit
+    /// * `edit_info` - A wrapper that contains the event ID and the content of
+    ///  the event to edit
     #[instrument(skip(self, new_content))]
     pub async fn edit(
         &self,
         new_content: RoomMessageEventContentWithoutRelation,
-        edit_item: &EventTimelineItem,
+        edit_info: EditInfo,
     ) -> Result<(), SendEventError> {
```
bnjbvr, line 474: This is not fallible anymore \o/ Can you get rid of the `Result`?

crates/matrix-sdk-ui/src/timeline/mod.rs
```diff
@@ -437,6 +505,58 @@ impl Timeline {
         Ok(())
     }
 
+    /// Gives the information needed to edit an event from an event id.
```
bnjbvr, line 508:
```suggestion
    /// Give the information needed to edit an event from an event id.
```

crates/matrix-sdk-ui/src/timeline/mod.rs
```diff
@@ -437,6 +505,58 @@ impl Timeline {
...
+    pub async fn get_edit_info_from_event_id(
```
bnjbvr, line 509: Can we get rid of the get_ prefix here too?

crates/matrix-sdk-ui/src/timeline/mod.rs
```diff
@@ -309,76 +341,120 @@ impl Timeline {
...
+            ),
         };
 
         self.send(content.into()).await?;
 
         Ok(())
     }
 
+    /// Gives the information needed to reply to an event from an event id.
+    pub async fn get_replied_to_info_from_event_id(
```
bnjbvr, line 406: and here too?

crates/matrix-sdk-ui/tests/integration/timeline/replies.rs
```diff
@@ -557,3 +561,227 @@ async fn test_send_reply_to_threaded() {
...
+    let reply_message = reply_item_remote_echo.content().as_message().unwrap();
+    assert_eq!(reply_message.body(), "Replying to Bob");
+    let in_reply_to = reply_message.in_reply_to().unwrap();
+    assert_eq!(in_reply_to.event_id, event_id_from_bob);
+
+    server.verify().await;
+}
+
+#[async_test]
+async fn test_send_reply_with_event_id_that_is_redacted() {
```
bnjbvr, line 680: 1. This is mostly a copy-pasta from the previous code, can you factor out some of the common parts?
2. Can you add a small comment at the top explaining what this is testing, under which conditions, please?

bnjbvr (changes requested):
Nice, we're getting there! A few small changes to tidy up the code would be nice to have.

crates/matrix-sdk-ui/src/timeline/mod.rs
Velin92: I need the `message_like_event` from the first match to get the relations

crates/matrix-sdk-ui/src/timeline/mod.rs
Velin92: we still need it because we use `self.send(content.into()).await?;`

bnjbvr: Oops you're right! So like the other function, this can be a plain `RoomSendQueueError` now.

crates/matrix-sdk-ui/tests/integration/timeline/replies.rs
Velin92: 1. Is it really worth for such a test? In case of change of the conditions of the test may make the common code more of a nuisance that useful.
2. Done!

bnjbvr (approved):
Yay, let's ship it! Thanks for getting it to the finish line!

PR #3555, opened 2024-06-13
sdk: tame `HttpError` just a bit

bnjbvr:
These are tiny changes to the `HttpError`, each nicely described in a small commit, to enhance my own understanding of those errors. No changes in behavior.

Part of #3361.


andybalaam (approved):
Great, nothing controversial to me: explict > implicit.

PR #3558, opened 2024-06-16
ffi: expose methods for SSO login

Johennes:
This exposes methods for logging in via the (legacy) SSO flow in the FFI bindings. I still need to verify that this actually works by calling it from an app. However, I'd first like to get feedback on whether adding this to the bindings would be acceptable in general.


bnjbvr (commented):
Out of curiosity, are you making use of the SSO login method in an app? I thought it was to be deprecated, and as such I would encourage to migrate to the new login methods instead, using OIDC.

(Moreover, I wouldn't piggyback on the OIDC login methods for new login ways, because the code there is messy; rather, we should have some kind of stateful object that remembers the login steps, or any kind of better, stateful API like that.)

Johennes:
> Out of curiosity, are you making use of the SSO login method in an app? I thought it was to be deprecated, and as such I would encourage to migrate to the new login methods instead, using OIDC.

Yes, indeed. I'm hoping to use these functions in https://github.com/unomed-dev/react-native-matrix-sdk for an app that connects to a server that doesn't currently use OIDC (which I think is still unstable / at MSC-stage?).

> (Moreover, I wouldn't piggyback on the OIDC login methods for new login ways, because the code there is messy; rather, we should have some kind of stateful object that remembers the login steps, or any kind of better, stateful API like that.)

Oh, hm. I might need some guidance on how to properly do that. 🙈

Currently I'm not using the `state` parameter on the URL because the SSO functions in the SDK didn't seem to support that. So I think other than the redirect URL (which needs to be passed in again anyway) there might not actually be any state to track on the bindings side?

bindings/matrix-sdk-ffi/Cargo.toml
```diff
@@ -46,6 +46,7 @@ url = { workspace = true }
 zeroize = { workspace = true }
 uuid = { version = "1.4.1", features = ["v4"] }
 language-tags = "0.3.2"
+serde_html_form.workspace = true
```
Hywan, line 49: Please be consistent with the rest of the `Cargo.toml` file 🙂:

```suggestion
serde_html_form = { workspace = true }
```

bindings/matrix-sdk-ffi/src/authentication_service.rs
```diff
@@ -91,6 +92,13 @@ pub enum AuthenticationError {
     #[error("An error occurred with OIDC: {message}")]
     OidcError { message: String },
 
+    #[error("The supplied callback URL used to complete SSO is invalid.")]
+    SsoCallbackUrlInvalid,
+    #[error("The supplied callback URL used to complete SSO does not contain a login token.")]
+    SsoCallbackUrlMissingLoginToken,
+    #[error("Logging in with the token from the supplied callback URL failed.")]
+    SsoLoginWithTokenFailed,
```
Hywan, line 100:
```suggestion
    #[error("The supplied callback URL used to complete SSO is invalid.")]
    SsoCallbackUrlInvalid,

    #[error("The supplied callback URL used to complete SSO does not contain a login token.")]
    SsoCallbackUrlMissingLoginToken,

    #[error("Logging in with the token from the supplied callback URL failed.")]
    SsoLoginWithTokenFailed,
```

Hywan (commented):
Do you think it is possible to write tests for these new methods?

Hywan:
Oh wait, I didn't see the comments… @bnjbvr has already replied to your PR.

bindings/matrix-sdk-ffi/Cargo.toml
Johennes: Oh, sorry, I didn't realize the format was different elsewhere. I blindly ran `cargo add`. 🙈

bindings/matrix-sdk-ffi/src/authentication_service.rs
Johennes: I personally don't mind the format but the OIDC errors above seem to not use a separating blank line either. Should I change it there as well?

Johennes:
> Do you think it is possible to write tests for these new methods?

I can probably do that, yes. However, @bnjbvr suggested that the architecture should maybe be changed to something akin to a state machine. It would be helpful if we could clarify this before adding the tests.

bnjbvr:
To clarify, the state machine I was talking about was to return a small object that starts the SSO login, with a single method "finish()" that received the right arguments, instead of adding two loose methods which don't look connected in any way. Just changing the API shape, but not necessarily how the functions work :)

Johennes:
I'm struggling a little to make this work because I'm not sure how to properly move the things needed to complete the login from the service into the returned object. 😕

I can do something like in the snippet below but this consumes the service which means it cannot be used again if `finish` fails. I'm not sure if that's a good design?

```rust
impl AuthenticationService {
    ...

    pub async fn start_sso_login(
        self,
        redirect_url: String,
        idp_id: Option<String>,
    ) -> Result<SsoHandler, AuthenticationError> {
        let client_guard = self.client.read().await;
        let Some(client) = client_guard.as_ref() else {
            return Err(AuthenticationError::ClientMissing);
        };

        let auth = client.inner.matrix_auth();
        let url = auth.get_sso_login_url(redirect_url.as_str(), idp_id.as_deref()).await.map_err(|e| AuthenticationError::Generic { message: e.to_string() })?;

        drop(client_guard);

        Ok(SsoHandler { service: self, url: url })
    }
}

#[derive(uniffi::Object)]
pub struct SsoHandler {
    /// The wrapped authentication service
    service: AuthenticationService,

    /// The underlying URL for authentication.
    pub url: String
}

#[uniffi::export]
impl SsoHandler {
    /// Completes the SSO login process.
    pub async fn finish(&self, callback_url: String) -> Result<Arc<Client>, AuthenticationError> {
        let client_guard = self.service.client.read().await;
        let Some(client) = client_guard.as_ref() else {
            return Err(AuthenticationError::ClientMissing);
        };

        let auth = client.inner.matrix_auth();

        let url =
            Url::parse(&callback_url).map_err(|_| AuthenticationError::SsoCallbackUrlInvalid)?;

        #[derive(Deserialize)]
        struct QueryParameters {
            #[serde(rename = "loginToken")]
            login_token: Option<String>,
        }

        let query_string = url.query().unwrap_or("");
        let query: QueryParameters = serde_html_form::from_str(query_string)
            .map_err(|_| AuthenticationError::SsoCallbackUrlInvalid)?;
        let token =
            query.login_token.ok_or(AuthenticationError::SsoCallbackUrlMissingLoginToken)?;

        auth.login_token(token.as_str())
            .await
            .map_err(|_| AuthenticationError::SsoLoginWithTokenFailed)?;

        drop(client_guard);

        let Some(client) = self.service.client.write().await.take() else {
            return Err(AuthenticationError::ClientMissing);
        };

        Ok(Arc::new(client))
    }
}
```

Johennes:
@bnjbvr do you have any thoughts on how to best proceed here?

bnjbvr:
> I can do something like in the snippet below but this consumes the service which means it cannot be used again if finish fails. I'm not sure if that's a good design?

Since there can be an error during the process, it'd be better to not consume the `AuthenticationService` (i.e. take its ownership) since we might need to retry, and we'd need to make sure it will properly error out (likely because a client might be absent) next time it's called after a successful login. It's really the best we can have here.

Johennes:
Ok, I think I _might_ have it. Adding tests for this part turned out somewhat tedious. So I only added one to verify the login URL so far.

Johennes:
I everything over to `Client` due to https://github.com/matrix-org/matrix-rust-sdk/issues/3579#issuecomment-2181435689.

I think the clippy error might only be fixable once `AuthenticationService` has been removed and `Client` made public. So this probably still needs to wait for @pixlwave's PR. Feedback on the rest of the code is still appreciated though. 🙂

pixlwave:
@Johennes The PR removing the service is here. There's still an `authentication.rs` file which might be a nice place to keep any types you've added. It hasn't been reviewed yet but I don't think there's anything controversial in there.

Johennes:
@Hywan this is ready for review now.

CC @pixlwave in case you'd like to take a look as well.

bindings/matrix-sdk-ffi/src/client.rs
```diff
@@ -172,6 +173,54 @@ impl From<matrix_sdk::TransmissionProgress> for TransmissionProgress {
     }
 }
 
+/// An object encapsulating the SSO login flow
```
pixlwave, line 176: Might be nice to have this in `authentication.rs` too perhaps?

bindings/matrix-sdk-ffi/src/client.rs
```diff
@@ -1527,3 +1590,100 @@ impl MediaFileHandle {
         )
     }
 }
+
+#[cfg(test)]
```
pixlwave, line 1594: More of a question for the Rust team, but I thought we didn't want to test anything in the FFI and if any logic needed testing it should instead live in the main SDK or UI crate, to keep the FFI down to lightweight wrappers.

pixlwave (commented):
Looking through out of curiosity I've added a couple of comments, but probably best to defer to the Rust team on these :)

bindings/matrix-sdk-ffi/src/client.rs
Johennes: Oh, I've only added these tests because of @Hywan's comment [above](https://github.com/matrix-org/matrix-rust-sdk/pull/3558#pullrequestreview-2122770915).

Hywan: Ideally if something needs tests, it should live in a `crates/` crate, not in FFI. FFI bindings should be as thin as possible with no logic except mapping data from the foreign language to Rust and vice versa.

bindings/matrix-sdk-ffi/src/client.rs
Johennes: Have moved it in b8f69b0ed98942a4b7ca09a38f97d0d4a157e584.

bindings/matrix-sdk-ffi/src/client.rs
Johennes: Have pushed the callback URL parsing and tests down into the SDK with b8f69b0ed98942a4b7ca09a38f97d0d4a157e584.

Johennes:
@Hywan is there anything else you'd like me to change before this can be reviewed?

crates/matrix-sdk/src/matrix_auth/mod.rs
```diff
@@ -292,6 +302,24 @@ impl MatrixAuth {
...
+    /// the user has logged in through a webview. This method should be used
+    /// in tandem with [`MatrixAuth::get_sso_login_url`].
+    pub fn login_with_sso_callback(&self, callback_url: Url) -> Result<LoginBuilder, SsoError> {
+        #[derive(Deserialize)]
+        struct QueryParameters {
+            #[serde(rename = "loginToken")]
+            login_token: Option<String>,
+        }
+
+        let query_string = callback_url.query().unwrap_or("");
```
Hywan, line 315: I think I would prefer `unwrap_or_default()` here, which avoid creating an empty `str` (the `Default` implementation always has an empty `str`, let's reuse this one instead).

bindings/matrix-sdk-ffi/src/authentication.rs
```diff
@@ -47,6 +53,54 @@ impl HomeserverLoginDetails {
...
+        let builder =
+            auth.login_with_sso_callback(url).map_err(|_| SsoError::CallbackUrlInvalid)?;
+        builder.await.map_err(|_| SsoError::LoginWithTokenFailed)?;
+        Ok(())
+    }
+}
+
+impl Debug for SsoHandler {
+    fn fmt(&self, fmt: &mut fmt::Formatter<'_>) -> Result<(), fmt::Error> {
+        write!(fmt, "SsoHandler")
```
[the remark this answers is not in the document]
Johennes: Yes, good idea. Have made it so.

bindings/matrix-sdk-ffi/Cargo.toml
```diff
@@ -46,6 +46,10 @@ url = { workspace = true }
 zeroize = { workspace = true }
 uuid = { version = "1.4.1", features = ["v4"] }
 language-tags = "0.3.2"
+serde_html_form = { workspace = true }
```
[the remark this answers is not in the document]
Johennes: Oh, great catch. I forgot removing these after moving the tests into the main SDK. 🤦‍♂️

Hywan (approved):
LGTM. Thanks!

Issue #3446, opened 2024-05-23
Room keys are not downloaded from the backup when it's ratcheted too much


kegsay:
From an end-user perspective on EX, this causes UTDs on login even if you have key backup set up correctly. This happens because:
 - rust SDK _has the key_ for that session ID,
 - ...but the key it has is ratcheted too much (since it's a new login)
 - ...and rust SDK doesn't try to query the backup for the same key in an earlier ratcheted state.

From a dev perspective, this means any time we see `2024-05-23T07:35:10.351132Z  WARN matrix_sdk_crypto::machine: Failed to decrypt a room event: The message was encrypted using an unknown message index, first known index 11, index of the message 7` - this will not self-heal even if the keys are being backed up correctly and it can be decrypted.

[Rageshake](https://github.com/element-hq/element-x-ios-rageshakes/issues/1835)

richvdh:
> 12 issues on EXI, 4 on EXA.

@kegsay presumably there is no way to be sure that the backup would actually have contained the key in question, even if we'd bothered to look it up? Indeed, given these are mobile clients (which tend to be quite long-lived), I'd say the balance of probability is that it did not.

... which is not to say we can't improve this, just that these statistics showing it as the highest-impact of all UTD bugs might well be bogus.

kegsay:
Fair point, the grep for logs don't show that those keys are in the backup, so it should be 0~12 EXI, 0~4 EXA.

richvdh:
For the record: the check in question is at https://github.com/matrix-org/matrix-rust-sdk/blob/b1c09ed/crates/matrix-sdk/src/encryption/tasks.rs#L142

PR #3560, opened 2024-06-17
sdk: Refactor key-backup integration tests

richvdh:
Some groundwork for fixing https://github.com/matrix-org/matrix-rust-sdk/issues/3446.

~~Almost all the changes are in the integration tests, but there is a tiny change to the `crypto` crate to expose `MegolmV1BackupKey::encrypt` outside the crate.~~

Edit: now exclusively updates in the integ tests.

### DIFF
--- a/crates/matrix-sdk/tests/integration/encryption/backups.rs
+++ b/crates/matrix-sdk/tests/integration/encryption/backups.rs
@@ -21,6 +21,7 @@ use matrix_sdk::{
     config::RequestConfig,
     encryption::{
         backups::{futures::SteadyStateError, BackupState, UploadState},
+        secret_storage::SecretStore,
         BackupDownloadStrategy, EncryptionSettings,
     },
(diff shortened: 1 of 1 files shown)


poljar (changes requested):
> Hardcoding the keys makes maintaining the tests difficult. In particular, trivial changes to the test can result in a lot of work, because you have to set up an entire test environment and recapture the entire flow.

Well yes, that's what test vectors are, if we remove all the test vectors then we'll have no tests that confirm that our implementation is compliant with whatever is out there.

 I'm fine with adding `mock_download_session_from_key_backup()` and adding more tests that will use this method, even though it is quite stretching the definition of what an integration test for the `matrix-sdk` crate should be, but please don't remove test vectors which I have taken the time to capture and add.

richvdh:
Hrm. Ok, I'll take that bit out.

poljar (approved):
Thanks, clippy still needs a bit more convincing, but looks good otherwise.

PR #3568, opened 2024-06-17
sdk: Fix key backup download for ratcheted sessions

richvdh:
Fixes https://github.com/matrix-org/matrix-rust-sdk/issues/3446 by checking that we have the key for the event in question, not just the key for the session.

~~This PR is based on #3567.~~

### DIFF
--- a/crates/matrix-sdk/tests/integration/encryption/backups.rs
+++ b/crates/matrix-sdk/tests/integration/encryption/backups.rs
@@ -12,13 +12,18 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use std::{fs::File, io::Write};
+use std::{fs::File, io::Write, sync::Arc};
 
 use anyhow::Result;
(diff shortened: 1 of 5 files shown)


crates/matrix-sdk/tests/integration/encryption/backups.rs
```diff
@@ -1361,6 +1372,131 @@ async fn enable_from_secret_storage_and_download_after_utd() {
...
+
+    // Now, import the megolm session into the client's store, at ratchet index 1.
+    {
+        let inbound_group_session = inbound_session_from_outbound_session(
+            sender_identity_keys.ed25519,
+            room_id,
+            &outbound_group_session,
+        )
+        .await
+        .unwrap();
```
poljar, line 1452: Maybe an assert that checks that the message index is indeed `1` wouldn't hurt here.

poljar (approved):
Left a small suggestion, looks good.

Issue #3029, opened 2024-01-18
Move the authentication service into the main SDK.


pixlwave:
The authentication service grew too big for the FFI a long time ago. It needs moving into the main SDK and properly covering with tests.

PR #3569, opened 2024-06-18
Client Builder: Add support for requiring Sliding Sync.

pixlwave:
This PR is a small part of #3029 that moves (and tests) the check for mandatory sliding sync from the FFI into the SDK with the use of a `ClientBuilder::requires_sliding_sync` method.


poljar (approved):
Looks good, thanks for writing this and doubly thanks for writing detailed tests for this.

PR #3570, opened 2024-06-18
timeline: when deduplicating event meta, keep the most recent instead of the oldest

bnjbvr:
Not doing this leads to an invalid ordering of events, as shown in the test: if we increase the timeline limit of a room using sliding sync, we'll receive a duplicate event, and if we ignore it, it'll be in an invalid position. The solution is to keep the most recent event (and remove the old one from event_meta).


andybalaam (approved):
Does what it says it does, and sounds like it is needed to fix this problem, but (I think?) exposes the fact that we need some kind of wider understanding of the order of an event, independent of how we received it. I thought the timeline included some of that, but maybe I'm wrong.

PR #3572, opened 2024-06-18
room: rename the `encrypted` log field to `encrypted_room`

bnjbvr:
This adjusts the control flow so we start with the simplest cases (do not encrypt) first, then the encryption path. This then properly records that a reaction is *not* sent encrypted. Also lowers the level of logging from debug to trace for those log events, since they're not particularly interesting.

### DIFF
--- a/crates/matrix-sdk/src/room/futures.rs
+++ b/crates/matrix-sdk/src/room/futures.rs
@@ -32,7 +32,9 @@ use ruma::{
     serde::Raw,
     OwnedTransactionId, TransactionId,
 };
-use tracing::{debug, info, Instrument, Span};
+#[cfg(feature = "image-proc")]
+use tracing::debug;
+use tracing::{info, trace, Instrument, Span};
 
 use super::Room;
 #[cfg(feature = "image-proc")]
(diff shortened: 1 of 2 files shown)


poljar:
> This then properly records that a reaction is not sent encrypted

I believe the `encrypted` flag we recorded told us if the room is encrypted or not.

bnjbvr:
Ah, I didn't know about this; in this case the current code might be sufficient, and maybe deserves just a comment about this `encrypted` for reactions.

poljar:
> Ah, I didn't know about this; in this case the current code might be sufficient, and maybe deserves just a comment about this `encrypted` for reactions.

Maybe we can rename the field to make it more explicit, sadly we can't add it to the span in the `instrument` attribute because the function to fetch the field is async.

crates/matrix-sdk/src/room/futures.rs
```diff
@@ -172,17 +174,17 @@ impl<'a> IntoFuture for SendRawMessageLikeEvent<'a> {
             Span::current().record("transaction_id", tracing::field::debug(&txn_id));
 
             #[cfg(not(feature = "e2e-encryption"))]
-            debug!("Sending plaintext event to room because we don't have encryption support.");
+            trace!("Sending plaintext event to room because we don't have encryption support.");
 
             #[cfg(feature = "e2e-encryption")]
             if room.is_encrypted().await? {
-                Span::current().record("encrypted", true);
+                Span::current().record("encrypted_room", true);
```
poljar, line 181: You can't just rename the field here, this will not record anything now, the span fields need to be renamed here: https://github.com/matrix-org/matrix-rust-sdk/blob/565f97409ed181c713d0c358b10d2d6085ca4404/crates/matrix-sdk/src/room/mod.rs#L1688

Not really a requirement, but I think `is_room_encrypted` would be even more clear, feel free to disagree.

poljar (approved):
Still one thing missing, but approving under the assumption you'll rename the field.

crates/matrix-sdk/src/room/futures.rs
bnjbvr: Ah, I was looking where the field is, thanks.

PR #3578, opened 2024-06-19
fix(sdk): Fix a substract with overflow in `LinkedChunk`

Hywan:
This patch fixes a panic in `LinkedChunk` where we substract with overflow when fetching the latest position of an item in an empty `Chunk`.


poljar (approved):
Makes sense.

PR #3580, opened 2024-06-19
feat(sdk): Implement `event_cache::Deduplicator`

Hywan:
This is WIP of an event deduplicator implemented on top of a Bloom filter. It works great, but it doesn't solve all issues we have right now. Right now, it filters duplicated events; instead, it should collect duplicated events with their positions, so that they can be removed from the `LinkedChunk` before being re-inserted. It's the basis to implement a naive reconciliation in the `EventCache`.

---

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -640,16 +640,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "bloomfilter"
-version = "1.0.13"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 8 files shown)


Hywan:
I have notice we have a growable bloom filter in our direct deps already, we should use it instead of introducing a new one.

Hywan:
It works. End of the experiment. This PR has been split into https://github.com/matrix-org/matrix-rust-sdk/pull/4172, https://github.com/matrix-org/matrix-rust-sdk/pull/4173 and https://github.com/matrix-org/matrix-rust-sdk/pull/4174.

PR #3581, opened 2024-06-19
timeline: small refactorings after #3553

bnjbvr:
These are small refactorings I spotted during the review of #3553, and I didn't want to add more back-and-forth to that review.


Hywan (approved):
Triple yes. Thanks 🙏 !

PR #3582, opened 2024-06-19
timeline: merge `edit_poll` and `edit`

bnjbvr:
So this is an attempt to merge `Timeline::edit` with `Timeline::edit_poll`, since it'd be nice to have a unique method to edit events in general.

Right now it's possible to dynamically provide an invalid combination of parameters (new content is a message while the edited event is a poll), but I think we can avoid that by introducing traits and more static typing.


bnjbvr:
I've got a better direction for doing this as part of #3663.

PR #3584, opened 2024-06-19
examples: Add a recovery command to the oidc-cli example

poljar:
This adds support to input your recovery key to the OIDC example which will allow the OIDC example client to be verified and have access to all the secrets (cross-signing keys and the backup recovery key).

Not particularly useful right now, but once the OIDC example is able to log in other devices via a QR code it becomes necessary to have access to all the secrets.


andybalaam (approved):
Looks good, thanks!

PR #3585, opened 2024-06-19
feat(ui): Migrate `RoomList` to manipulate `Room` instead of `RoomListEntry`

Hywan:
WIP


bindings/matrix-sdk-ffi/src/room_list.rs
```diff
@@ -479,6 +516,12 @@ pub struct RoomListItem {
     utd_hook: Option<Arc<UtdHookManager>>,
 }
 
+impl From<matrix_sdk_ui::room_list_service::Room> for RoomListItem {
+    fn from(value: matrix_sdk_ui::room_list_service::Room) -> Self {
+        Self { inner: Arc::new(value), utd_hook: None }
```
bnjbvr, line 521: It's probably wrong to set `utd_hook` to `None`: if the `RoomListItem` is then used to create a timeline with `init_timeline`, then we'll have lost the `utd_hook`. Instead, we need to pass down the `utd_hook` properly from the `RoomListService`.

bindings/matrix-sdk-ffi/src/room_list.rs
```diff
@@ -192,62 +161,137 @@ impl RoomList {
...
+        // SAFETY: `ptr` is correctly aligned.
+        unsafe {
+            addr_of_mut!((*ptr).entries_stream).write(entries_stream);
+        }
+
+        // The result is complete, let's return it!
+        //
+        // SAFETY: `result` is fully initialized, all its fields have received a valid
+        // value.
+        Arc::new(unsafe { result.assume_init() })
```
bnjbvr, line 271: nit: use a single `unsafe` block?

bindings/matrix-sdk-ffi/src/room_list.rs
```diff
@@ -192,62 +161,137 @@ impl RoomList {
...
     }
 }
 
-#[derive(uniffi::Record)]
-pub struct RoomListEntriesResult {
-    pub entries: Vec<RoomListEntry>,
-    pub entries_stream: Arc<TaskHandle>,
+#[derive(uniffi::Object)]
+pub struct RoomListEntriesWithDynamicAdaptersResult {
+    this: Arc<RoomList>,
```
bnjbvr, line 281: Should this be `Pin`ned, considering the requirement for the reference to never be invalidated?

bnjbvr (changes requested):
Great piece! I *think* there's a bug when it comes to the UTD hook, so would like to see just the minimal changes needed to make this work again. To make the next round of review super fast, I would propose to have `fixup!` commits so I can quickly check these, and then you could autosquash and we could merge later. How does that sound?

testing/matrix-sdk-integration-testing/src/tests/sliding_sync/room.rs
```diff
@@ -1000,26 +1012,15 @@ async fn test_roominfo_update_deduplication() -> Result<()> {
     // Stream has the room again, but no second event
     // TODO: Synapse sometimes sends the same event two times. This is the
     // workaround:
```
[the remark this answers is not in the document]
Hywan: :+1:

crates/matrix-sdk-ui/src/room_list_service/mod.rs
```diff
@@ -393,35 +392,11 @@ impl RoomListService {
...
-            .await
-            .ok_or_else(|| Error::InputCannotBeApplied(Input::Viewport(ranges.clone())))?;
-
-        *viewport_ranges = ranges;
-
-        Ok(InputResult::Applied)
+    ///
+    /// After some updates, these is no more `Input` but this API is kept in
+    /// case of a future necessity.
+    pub async fn apply_input(&self, _input: Input) -> Result<InputResult, Error> {
```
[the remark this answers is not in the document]
Hywan: Not a big deal to remove it. Bye bye `apply_input`.

crates/matrix-sdk-base/src/latest_event.rs
```diff
@@ -129,6 +129,13 @@ pub struct LatestEvent {
     /// The actual event.
     event: SyncTimelineEvent,
 
+    /// A copy of `Self::event::event::origin_server_ts`.
+    ///
+    /// This information can potentially be requested multiple times. Keeping a
+    /// copy here avoid to parse `Self::event::event` which is a `Raw<_>`.
+    #[serde(skip_serializing)]
```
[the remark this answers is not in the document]
Hywan: Exactly. I'm updating the comment.

crates/matrix-sdk-ui/src/room_list_service/sorters/name.rs
```diff
@@ -0,0 +1,129 @@
...
+    use matrix_sdk_test::async_test;
+    use ruma::room_id;
+
+    use super::{
+        super::super::filters::{client_and_server_prelude, new_rooms},
+        *,
+    };
+
+    #[async_test]
+    async fn test_with_two_latest_events() {
```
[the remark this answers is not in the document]
Hywan: Ahhh, bad copy-paste. Fixing it!

crates/matrix-sdk-ui/src/room_list_service/sorters/or.rs
```diff
@@ -0,0 +1,104 @@
...
+use super::{BoxedSorterFn, Sorter};
+
+/// Create a new sorter that will run multiple sorters. When the nth sorter
+/// returns [`Ordering::Equal`], the next sorter is called. It stops at soon as
+/// a sorter return [`Ordering::Greater`] or [`Ordering::Less`].
+pub fn new_sorter(sorters: Vec<BoxedSorterFn>) -> impl Sorter {
+    move |left, right| -> Ordering {
+        for sorter in &sorters {
+            match sorter(left, right) {
+                result @ Ordering::Greater | result @ Ordering::Less => return result,
```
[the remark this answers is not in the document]
Hywan: 🙂

crates/matrix-sdk-ui/src/room_list_service/sorters/mod.rs
```diff
@@ -0,0 +1,38 @@
...
+//! A collection of room sorters.
+
+mod name;
+mod or;
+mod recency;
+
+use std::cmp::Ordering;
+
+pub use name::new_sorter as new_sorter_name;
+pub use or::new_sorter as new_sorter_or;
```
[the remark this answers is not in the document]
Hywan: Ah, you've found the definition I was looking for, thanks!

crates/matrix-sdk-base/src/sliding_sync.rs
```diff
@@ -292,7 +292,7 @@ impl BaseClient {
 
         trace!("ready to submit changes to store");
         store.save_changes(&changes).await?;
-        self.apply_changes(&changes, false);
+        self.apply_changes(&changes, true);
```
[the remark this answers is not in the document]
Hywan: The room stream won't fire if the room already exists. A room info update is likely to trigger a room list update if the latest event changes for example. Hence the `true` here. I'm adding a comment.

crates/matrix-sdk-ui/tests/integration/room_list_service.rs
```diff
@@ -1597,95 +1684,49 @@ async fn test_dynamic_entries_stream_manual_update() -> Result<(), Error> {
...
+        // Receive the next values.
+        append [
+            "!r3:bar.org",
+            "!r2:bar.org",
+            "!r1:bar.org",
+        ];
         end;
     };
-
     assert_pending!(dynamic_entries_stream);
```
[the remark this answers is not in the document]
Hywan: What do you mean?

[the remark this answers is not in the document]
bnjbvr: Ah, I meant to select the few lines above too. I think you deleted some test code that is now untested, and that would be nice to keep as some sort of unit testing for the manual update functionality.

bindings/matrix-sdk-ffi/src/room_list.rs
Hywan: Yup, thanks for the heads up. Done.

bindings/matrix-sdk-ffi/src/room_list.rs
Hywan: This is not a self-referential structure. We just need `this` to live as long as `entries_stream`, but `entries_stream` doesn't reference `this` (“pointer speaking”).

Hywan:
@bnjbvr I've rebased on top of `main` but I've marked all the new commits as `!fixup`, except github.com/matrix-org/matrix-rust-sdk/pull/3585/commits/98aae99674def3e38505f71c464846de850237ba, https://github.com/matrix-org/matrix-rust-sdk/pull/3585/commits/50c804b25308fdadd7dfe7299eeaf9365ba0e8a3, https://github.com/matrix-org/matrix-rust-sdk/pull/3585/commits/235297849e3c086353354ca35ef84e962a608d5c, https://github.com/matrix-org/matrix-rust-sdk/pull/3585/commits/6132841f5c29e5984b663c46f7eac94cbb8f2a26, https://github.com/matrix-org/matrix-rust-sdk/pull/3585/commits/7bd042796647b080408d404dfebf09716c3e4d93, and https://github.com/matrix-org/matrix-rust-sdk/pull/3585/commits/824cb0682a072ed73a2c8f1c7d1b92b83ddf34c2.

crates/matrix-sdk-ui/tests/integration/room_list_service.rs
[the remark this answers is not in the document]
Hywan: I think it deserves way more tests than only two tests. I'll add a couple more tests with another PR, as I also think we should revisit this room info update thingy.

crates/matrix-sdk-base/src/sliding_sync.rs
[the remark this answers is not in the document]
bnjbvr: I think this would lead to spurious updates, because we're already saving room infos explicitly here. (The room state management is a bit of a mess, we should likely add this to the spring cleaning list). Putting `true` will force a spurious update here, which might not be an issue in testing, so let's see if we can keep it.

crates/matrix-sdk-ui/src/room_list_service/room_list.rs
```diff
@@ -195,34 +195,37 @@ fn merge_stream_and_receiver(
...
-                biased; // Prefer manual updates for easier test code
+                // We want to give priority on updates from `raw_stream` as it will necessarily trigger a “refresh” of the rooms.
+                biased;
 
-                Ok(update) = roominfo_update_recv.recv() => {
-                    if !update.trigger_room_list_update {
-                        continue;
-                    }
+                diffs = raw_stream.next() => {
+                    tracing::error!("merge streams: receive diff");
```
bnjbvr: nit: debug printing here. If you want to keep it, lower the log level to trace.

crates/matrix-sdk-ui/src/room_list_service/room_list.rs
```diff
@@ -195,34 +195,37 @@ fn merge_stream_and_receiver(
...
-                        yield v;
+                        yield diffs;
                     } else {
                         // Restart immediately, don't keep on waiting for the receiver
                         break;
                     }
                 }
+
+                Ok(update) = roominfo_update_recv.recv() => {
+                    tracing::error!(trigger_update = ?update.trigger_room_list_update, "merge streams: receive roominfo update");
```
bnjbvr: ditto here

crates/matrix-sdk-ui/src/room_list_service/room_list.rs
```diff
@@ -195,34 +195,37 @@ fn merge_stream_and_receiver(
 
         loop {
             select! {
-                biased; // Prefer manual updates for easier test code
+                // We want to give priority on updates from `raw_stream` as it will necessarily trigger a “refresh” of the rooms.
+                biased;
```
bnjbvr, line 199: Any way to add a regression test for the issue fixed by 98aae99674def3e38505f71c464846de850237ba?

crates/matrix-sdk-ui/src/room_list_service/sorters/recency.rs
```diff
@@ -30,6 +30,29 @@ where
     F: Fn(&Room, &Room) -> (Option<LatestEvent>, Option<LatestEvent>),
 {
     fn matches(&self, left: &Room, right: &Room) -> Ordering {
+        if left.id() == right.id() {
+            tracing::error!("recency, with the same room");
```
bnjbvr: nit: debug printing again

crates/matrix-sdk-ui/src/room_list_service/sorters/recency.rs
```diff
@@ -30,6 +30,29 @@ where
...
+            // `left` and `right` are the same room. We are comparing the same
+            // `LatestEvent`!
+            //
+            // The way our `Room` types are implemented makes it so they are sharing the
+            // same data, because they are all built from the same store. They can be seen
+            // as shallow clones of each others. In practise it's really great: a `Room` can
```
bnjbvr, line 39: nit: practice

crates/matrix-sdk-ui/src/room_list_service/sorters/recency.rs
```diff
@@ -30,6 +30,29 @@ where
...
+            // search algorithm. `left` and `right` will have the exact same `LatestEvent`,
+            // so `left` and `right` will always be `Ordering::Equal`. This is wrong: if
+            // `left` is compared with `right` and if they are both the same room, it means
+            // that one of them (either `left`, or `right`, it's not important) has received
+            // an update. The room position is very likely to change. But if they compare to
+            // `Equal`, the position may not change. It actually depends of the search
+            // algorithm used by [`eyeball_im_util::SortBy`].
+            //
+            // Since this sorter is sorting rooms from the most recent latest event to the
+            // oldest, we should return `Ordering::Greatest`: the room is greatest to its
```
bnjbvr, line 50: nit (twice in that sentence): Greater, not Greatest

crates/matrix-sdk-ui/src/room_list_service/sorters/recency.rs
```diff
@@ -30,6 +30,29 @@ where
...
+            // previous version.
```
bnjbvr, line 51: Ok, I'm convinced by this argument, but can we tweak the wording to something like `Since this room received an update, it is more recent than the previous one we matched against, so return Ordering::Greater`.

crates/matrix-sdk-ui/src/room_list_service/room_list.rs
```diff
@@ -199,8 +199,6 @@ fn merge_stream_and_receiver(
                 biased;
 
                 diffs = raw_stream.next() => {
-                    tracing::error!("merge streams: receive diff");
```
bnjbvr: Please fold this commit (6132841f5c29e5984b663c46f7eac94cbb8f2a26) with the one that introduced the logs. There's no point in polluting our commit history by adding code in commits that's reverted in commits in the same PR (unless you plan to squash everything into a single commit at the end).

crates/matrix-sdk-ui/tests/integration/room_list_service.rs
```diff
@@ -1476,6 +1476,16 @@ async fn test_dynamic_entries_stream() -> Result<(), Error> {
         push front [ "!r4:bar.org" ];
         end;
     };
+    assert_entries_batch! {
```
bnjbvr, line 1479: This commit (824cb0682a072ed73a2c8f1c7d1b92b83ddf34c2) likely tweaks some test expectations after a change in the code, can you fold it into the same commit that introduced the change in test expectations, instead?

bnjbvr (approved):
Sweet! The commit history is messy and I would like you to tidy it up before merging, please, to avoid intermediate commits that should've been folded into other commits in the first place.

Otherwise, the new commits look good; I haven't looked again at the previous ones, since you've addressed all the comments there.

Thank you very much, and great work here :partying_face:

crates/matrix-sdk-ui/src/room_list_service/room_list.rs
Hywan: Removed.

crates/matrix-sdk-ui/src/room_list_service/room_list.rs
Hywan: Removed.

crates/matrix-sdk-ui/src/room_list_service/room_list.rs
Hywan: In a next PR yes. I would like to clearly tests the room info update in a dedicated test suite.

crates/matrix-sdk-ui/src/room_list_service/sorters/recency.rs
Hywan: Sure.

crates/matrix-sdk-ui/src/room_list_service/room_list.rs
bnjbvr: Not sure how it's fixing anything though; there's no await in the select's body, so no place for tokio to interrupt a branch or the other. I really would like to understand it better, and a regression test would be an irrefutable proof.

crates/matrix-sdk-base/src/sliding_sync.rs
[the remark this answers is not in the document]
Hywan: Ideally, we want to detect whether an update is necessary instead of always coercing to `true`. But let's do that in another PR.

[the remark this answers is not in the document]
bnjbvr: The way it worked before: changing the room's data meant the vectordiff would include a change for that room (even if it's not moved position in the list: just to notify that there's a change in the room, so the subscribers can be notified). Hence, we didn't need to pass `true` here, since the change would happen automatically.

Has this changed? Is this the actual reason why you need to pass `true` here?

Hywan:
I've changed the code to use the `timestamp` from the sliding sync room response instead of the `LatestEvent` origin server timestamp. Why? Because if we don't have a `LatestEvent`, then we can't sort the rooms. On a fresh start, it happens pretty easily, and the room list feels pretty broken. With this new strategy, everything looks like before the PR, which is what we are trying to achieve so far, except things are computed client-side instead of server-side.

@bnjbvr Can you review https://github.com/matrix-org/matrix-rust-sdk/pull/3585/commits/9744ad410c90d7a96c9250ab2dd56df3affe10f9, https://github.com/matrix-org/matrix-rust-sdk/pull/3585/commits/22294896901f0e6385802a66719d206257a09500 and https://github.com/matrix-org/matrix-rust-sdk/pull/3585/commits/063fa52d8e0eaf70ad90ec270e676f4d27021003 please? @jmartinesp confirms it now work similarly to what we have previously, and I confirm that too with multiverse.

crates/matrix-sdk-base/src/rooms/normal.rs
```diff
@@ -946,6 +954,15 @@ pub struct RoomInfo {
...
     pub(crate) cached_display_name: Option<DisplayName>,
+
+    /// The recency timestamp of this room.
+    ///
+    /// It's not to be confused with `origin_server_ts` of the latest event.
+    /// Sliding Sync might "ignore” some events when computing the recency
+    /// timestamp of the room. Thus, using this `recency_timestamp` value is
+    /// more accurate than relying on the latest event.
+    #[cfg(feature = "experimental-sliding-sync")]
+    pub(crate) recency_timestamp: Option<MilliSecondsSinceUnixEpoch>,
```
bnjbvr, line 965: nit: add a `#[serde(default)]` annotation, otherwise `RoomInfo`s serialized before this patch will fail to deserialize, resulting in a global error that may lead to user logouts on EX apps.

crates/matrix-sdk-base/src/sliding_sync.rs
```diff
@@ -1736,6 +1741,92 @@ mod tests {
...
+            let room = {
+                let mut room = v4::SlidingSyncRoom::new();
+                room.timestamp = Some(MilliSecondsSinceUnixEpoch(42u32.into()));
+
+                room
+            };
+            let response = response_with_room(room_id, room);
+            client.process_sliding_sync(&response, &()).await.expect("Failed to process sync");
+
+            // Then the room in the client has the avatar
```
Hywan, line 1782: s/avatar/recency timestamp/

Spotted and will be fixed.

crates/matrix-sdk-base/src/sliding_sync.rs
```diff
@@ -1736,6 +1741,92 @@ mod tests {
...
+            let room = {
+                let mut room = v4::SlidingSyncRoom::new();
+                room.timestamp = Some(MilliSecondsSinceUnixEpoch(153u32.into()));
+
+                room
+            };
+            let response = response_with_room(room_id, room);
+            client.process_sliding_sync(&response, &()).await.expect("Failed to process sync");
+
+            // Then the room in the client has the avatar
```
Hywan, line 1821: s/avatar/recency timestamp/

Spotted and will be fixed.

crates/matrix-sdk-base/src/sliding_sync.rs
```diff
@@ -1736,6 +1741,92 @@ mod tests {
...
+        let client = logged_in_base_client(None).await;
+        let room_id = room_id!("!r:e.uk");
+
+        // When I send sliding sync response containing a room with a recency timestamp
+        let room = {
+            let mut room = v4::SlidingSyncRoom::new();
+            room.timestamp = Some(MilliSecondsSinceUnixEpoch(42u32.into()));
+
+            room
+        };
```
bnjbvr, line 1756: nit: here and below a few times, we could use `assign!()` to avoid these 6 lines :eyes:

crates/matrix-sdk-base/src/store/migration_helpers.rs
```diff
@@ -80,6 +80,8 @@ pub struct RoomInfoV1 {
     #[cfg(feature = "experimental-sliding-sync")]
     latest_event: Option<SyncTimelineEvent>,
     base_info: BaseRoomInfoV1,
+    #[cfg(feature = "experimental-sliding-sync")]
+    recency_timestamp: Option<MilliSecondsSinceUnixEpoch>,
```
bnjbvr, line 84: This is not correct: this struct was used for migrating previous data with a structural change, so instead, you should remove this field here, and fill `recency_timestamp` to `None` in the below constructor of `RoomInfo`.

bnjbvr (changes requested):

crates/matrix-sdk-base/src/rooms/normal.rs
Hywan: Oh, nice, thanks!

crates/matrix-sdk-base/src/store/migration_helpers.rs
Hywan: :+1: Done

crates/matrix-sdk-base/src/sliding_sync.rs
Hywan: Done.

Hywan:
@bnjbvr One more time with https://github.com/matrix-org/matrix-rust-sdk/pull/3585/commits/1f1ff3dea1f332fa4777247c80768b9a7f019223 please 🥺?

bnjbvr (approved):
yay

Hywan:
Complement crypto is expected but I've disabled it. Force merging the PR as all other CI jobs are fine.

PR #3586, opened 2024-06-19
feat(sdk): `room_list_service::Room::new` is now infallible

Hywan:
This patch makes `Room::new` infallible, i.e. it no longer returns a `Result<Self, _>` but `Self` directly.

---

* Extracted from https://github.com/matrix-org/matrix-rust-sdk/pull/3585.


bnjbvr (approved):
Sure.

PR #3587, opened 2024-06-19
sdk: Move the OIDC helper methods from the FFI's `AuthenticationService` into `Client`

pixlwave:
This PR moves the remaining logic from the `AuthenticationService` into the SDK. For now this doesn't expose anything new in the FFI and simply calls the inner methods. I'll leave that and deleting the service for a follow-up PR.

Another part of #3029.

### DIFF
--- a/bindings/matrix-sdk-ffi/src/client.rs
+++ b/bindings/matrix-sdk-ffi/src/client.rs
@@ -13,15 +13,12 @@ use matrix_sdk::{
         requests::account_management::AccountManagementActionFull,
         types::{
             client_credentials::ClientCredentials,
-            errors::ClientErrorCode::AccessDenied,
             registration::{
                 ClientMetadata, ClientMetadataVerificationError, VerifiedClientMetadata,
             },
-            requests::Prompt,
         },
-        AuthorizationResponse, Oidc, OidcSession,
+        OidcAuthorizationData, OidcError, OidcSession,
     },
-    reqwest::StatusCode,
     ruma::{
         api::client::{
             media::get_content_thumbnail::v3::Method,
@@ -39,7 +36,7 @@ use matrix_sdk::{
         serde::Raw,
         EventEncryptionAlgorithm, RoomId, TransactionId, UInt, UserId,
     },
-    AuthApi, AuthSession, Client as MatrixClient, SessionChange, SessionTokens,
+    AuthApi, AuthSession, Client as MatrixClient, Error, SessionChange, SessionTokens,
 };
 use matrix_sdk_ui::notification_client::{
     NotificationClient as MatrixNotificationClient,
@@ -63,7 +60,7 @@ use url::Url;
 
 use super::{room::Room, session_verification::SessionVerificationController, RUNTIME};
 use crate::{
-    authentication_service::{AuthenticationError, OidcAuthenticationData, OidcConfiguration},
+    authentication_service::{AuthenticationError, OidcConfiguration},
     client,
     encryption::Encryption,
     notification::NotificationClient,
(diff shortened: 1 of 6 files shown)


bindings/matrix-sdk-ffi/src/client.rs
```diff
@@ -350,6 +353,68 @@ impl Client {
...
+    /// Completes the OIDC login process.
+    pub(crate) async fn login_with_oidc_callback(
+        &self,
+        authorization_data: Arc<OidcAuthorizationData>,
+        callback_url: String,
+    ) -> Result<(), AuthenticationError> {
+        let url = Url::parse(&callback_url).or(Err(AuthenticationError::OidcCallbackUrlInvalid))?;
+
+        self.inner.oidc().login_with_oidc_callback(&authorization_data, url).await.map_err(
+            |e| match e {
```
andybalaam, line 403: If you `impl Into<AuthenticationError> for Error` then you could replace this whole `map_err` with just `?`, which might be easier to read.

bindings/matrix-sdk-ffi/src/client.rs
```diff
@@ -350,6 +353,68 @@ impl Client {
...
+            })
+            .collect::<HashMap<_, _>>();
+        let registrations = OidcRegistrations::new(
+            registrations_file,
+            oidc_metadata.clone(),
+            static_registrations,
+        )?;
+
+        let data =
+            self.inner.oidc().url_for_oidc_login(oidc_metadata, registrations).await.map_err(
```
andybalaam, line 383: It might be nice to `impl Into<AuthenticationError> for OidcError` to replace this whole thing with `?`

crates/matrix-sdk/src/oidc/mod.rs
```diff
@@ -434,6 +437,157 @@ impl Oidc {
...
+                if err.error.error == AccessDenied {
+                    // The user cancelled the login in the web view.
+                    return Err(OidcError::CancelledAuthorization.into());
+                }
+                return Err(OidcError::Authorization(err).into());
+            }
+        };
+
+        // Given the InvalidState error already exists, I wonder if this was ever
+        // necessary to manually check?
```
andybalaam, line 502: Should this be marked with `TODO` or similar?

crates/matrix-sdk/src/oidc/mod.rs
```diff
@@ -434,6 +437,157 @@ impl Oidc {
...
+
+        registrations
+            .set_and_write_client_id(ClientId(client_id), issuer)
+            .map_err(|e| OidcError::UnknownError(Box::new(e)))?;
+
+        Ok(())
+    }
+
+    /// Attempts to load an existing OIDC dynamic client registration for a
+    /// given issuer.
```
andybalaam, line 567: Please can you document the return value? (If there were a meaningful way to return `Result<(), SomeError>` that would be even better...

andybalaam (approved):
Thanks for the tests! Some optional comments, but looking good.

crates/matrix-sdk/src/oidc/mod.rs
pixlwave: Oh I forgot about that. Well now that there's a test for this it should be pretty easy to answer the question! 👀

bindings/matrix-sdk-ffi/src/client.rs
pixlwave: Oh yeah that would be nice. I've added a TODO for this as in the next PR I'll kill the `AuthenticationService` and simplify the error into an FFI `OidcError` so I'd rather not introduce breaking changes right now.

bindings/matrix-sdk-ffi/src/client.rs
pixlwave: Ditto

crates/matrix-sdk/src/oidc/mod.rs
pixlwave: Turns out it was properly handled *if* you hadn't cancelled a previous request. I've clarified in the comment and prepared to expose an extra method for apps to do the right thing.

crates/matrix-sdk/src/oidc/mod.rs
pixlwave: Have added the documentation - it is perfectly valid to not find an existing registration so a `bool` seemed more appropriate to me than creating an error in this instance.

andybalaam: Great, thanks!

andybalaam (approved):
Great, thanks!

PR #3592, opened 2024-06-20
send queue: add state store backend for persisting pending events to disk

bnjbvr:
Ruma needs https://github.com/ruma/ruma/pull/1850. I need an Advil.

Part of #3361. The remaining bit will be to, at start, reload all the rooms' send queues that had _some_ pending event (right now, pending events will be automatically re-sent as soon as a timeline is opened for that room).

### DIFF
--- a/crates/matrix-sdk-base/src/store/traits.rs
+++ b/crates/matrix-sdk-base/src/store/traits.rs
@@ -27,13 +27,14 @@ use ruma::{
     events::{
         presence::PresenceEvent,
         receipt::{Receipt, ReceiptThread, ReceiptType},
-        AnyGlobalAccountDataEvent, AnyRoomAccountDataEvent, EmptyStateKey, GlobalAccountDataEvent,
-        GlobalAccountDataEventContent, GlobalAccountDataEventType, RedactContent,
-        RedactedStateEventContent, RoomAccountDataEvent, RoomAccountDataEventContent,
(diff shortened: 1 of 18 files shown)


bnjbvr:
I'm really not a big fan of the suggestions made by clippy here: https://github.com/matrix-org/matrix-rust-sdk/actions/runs/9601700440/job/26480909434?pr=3592#step:7:40

I imagine clippy's suggestion may avoid a few bounds check, but since this is testing code, why would we care?

crates/matrix-sdk-base/src/store/traits.rs
```diff
@@ -369,27 +370,64 @@ pub trait StateStore: AsyncTraitDeps {
...
     ///
     /// * `room_id` - The `RoomId` of the room to delete.
     async fn remove_room(&self, room_id: &RoomId) -> Result<(), Self::Error>;
+
+    /// Save an event to be sent by a send queue later.
+    ///
+    /// # Arguments
+    ///
+    /// * `room_id` - The `RoomId` of the send queue's room.
+    /// * `transaction_id` - The unique key identifying the event to be sent
```
poljar, line 400: Let us mention that the transaction ID must be random.

poljar (approved):
Left a small comment, I think this looks good.

bnjbvr:
Thanks for the review meeting party :partying_face:

PR #3598, opened 2024-06-23
sdk: Add support for authenticated media requests

zecakeh:
This follows the [implementation guidance](https://matrix.org/blog/2024/06/20/matrix-v1.11-release/#client-implementation-guidance) to only use the authenticated endpoints when the server advertises support for Matrix 1.11.


crates/matrix-sdk/src/media.rs
```diff
@@ -269,10 +272,19 @@ impl Media {
             }
         };
 
+        let server_versions = self.client.server_versions().await?;
```
poljar, line 275: This allows the homeserver to downgrade us towards unauthenticated media uploads. While it isn't the end of the world, I think we should provide a `ClientBuilder` option that allows users to always use the authenticated endpoints, despite what the homeserver suggests.

This doesn't need to happen as part of this PR, but please leave a TODO item here.

poljar (changes requested):
Left a small nit, please get rid of the merge conflicts and add the TODO item and we can merge.

crates/matrix-sdk/src/media.rs
zecakeh: Todo added

zecakeh:
I merged main and added the todo (with a small refactoring).

poljar (approved):
Looks good. Thanks.

PR #3602, opened 2024-06-24
fix(sdk): when an encryption event is received, mark the room encryption as set

jmartinesp:
`RoomInfo::handle_state_event(...)` will check this condition so we don't have to later ask the HS whether the room is encrypted or not through `room::is_encrypted()`.

Also, as we need some way to get this info in the room list in our clients, I also added 2 ways of doing this the FFI crate, one through `RoomInfo` and another one through `RoomListItem`. If we don't want the duplication I can get rid of one of them.

This bug was found by @poljar, thanks for pointing it out!

Signed-off-by:


bindings/matrix-sdk-ffi/src/room_list.rs
```diff
@@ -566,6 +566,10 @@ impl RoomListItem {
         self.inner.init_timeline_with_builder(timeline_builder).map_err(RoomListError::from).await
     }
 
+    async fn is_encrypted(&self) -> bool {
```
poljar, line 569: Please add a doc comment, documentation is nowadays visible over the FFI, also please document that the correctness of this might depend on specifying `m.room.encryption` as required state.

poljar (changes requested):
Looks good, left a nit about the documentation.

bindings/matrix-sdk-ffi/src/room_list.rs
jmartinesp: Let me know if https://github.com/matrix-org/matrix-rust-sdk/pull/3602/commits/0162f6ba49e92264416d45dc6a99f77d7d62f0d3 works for you.

poljar (approved):
Thanks, looks good.

PR #3603, opened 2024-06-24
send queue: mark reloaded local echoes as wedged or not

bnjbvr:
Based on top of https://github.com/matrix-org/matrix-rust-sdk/pull/3601.

This adds an `is_wedged` boolean to `LocalEcho`, so it's possible to know if a local echo was wedged or not, and so the timeline can update its event send state as soon as possible when it's reloading persisted local echoes.

Spotted by @stefanceriu (thanks!). Includes a regression test.

Note: draft only because this is waiting for the Ruma fix, otherwise the code is ready for review.

### DIFF
--- a/crates/matrix-sdk-ui/src/timeline/builder.rs
+++ b/crates/matrix-sdk-ui/src/timeline/builder.rs
@@ -273,21 +273,7 @@ impl TimelineBuilder {
             Some(spawn({
                 // Handles existing local echoes first.
                 for echo in local_echoes {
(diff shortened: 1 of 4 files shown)


bnjbvr:
This is now rebased and ready for review :partying_face:

poljar (approved):
Makes sense, I slightly preferred the local echo handling which destructured the `LocalEcho` struct, but only because I always forget to destructure things when it makes sense myself :sweat_smile:.

PR #3604, opened 2024-06-25
Reshare Megolm session after the other party unwedges

uhoreg:
Fixes https://github.com/element-hq/element-meta/issues/2389

For each recipient device, we keep an "Olm wedging index" that indicates (approximately) how many times they tried to unwedge the Olm session.  When we share a Megolm session, we store the Olm wedging index.  This allows us to tell whether the Olm session was unwedged since we shared the Megolm session, so that we know if we should re-share it.

We detect an attempt to unwedge the Olm session by checking if the Olm message that we decrypted is from a brand new Olm session.  This is not entirely accurate, since this could happen, for example, if Alice and Bob create Olm sessions at the same time.  This may result in some Megolm sessions being re-shared unnecessarily, but should not make a huge difference.


crates/matrix-sdk-crypto/src/olm/group_sessions/outbound.rs
```diff
@@ -69,7 +69,7 @@ const ROTATION_MESSAGES: u64 = 100;
 pub(crate) enum ShareState {
     NotShared,
     SharedButChangedSenderKey,
-    Shared(u32),
+    Shared(u32, u32),
```
poljar, line 72: This could take some docs, a single u32 could be figured out relatively easily, two of them double the confusion. It probably makes sense to convert the enum variant from a tuple variant to one with named fields, i.e.:

```rust
Shared {
    megolm_index: u32,
    wedging_foo: u32,
}
```

crates/matrix-sdk-crypto/src/olm/account.rs
```diff
@@ -1305,12 +1305,22 @@ impl Account {
...
                 // TODO: separate the session cache from the storage so we only add
                 // it to the cache but don't store it.
-                store
-                    .save_changes(Changes {
-                        sessions: vec![result.session.clone()],
-                        ..Default::default()
-                    })
-                    .await?;
+                let mut changes =
+                    Changes { sessions: vec![result.session.clone()], ..Default::default() };
```
poljar, line 1309: A bit whitespace wouldn't hurt here.
```suggestion
                    Changes { sessions: vec![result.session.clone()], ..Default::default() };

```

crates/matrix-sdk-crypto/src/olm/account.rs
```diff
@@ -1305,12 +1305,22 @@ impl Account {
...
+                // Any new Olm session will bump the Olm wedging index for the
+                // sender's device, if we have their device, which will cause
+                // us to re-send existing Megolm sessions to them the next time
+                // we use the session.  If we don't have their device, this
+                // means that we haven't tried to send them any Megolm sessions
+                // yet, so we don't need to worry about it.
+                if let Some(device) = store.get_device_from_curve_key(sender, sender_key).await? {
```
poljar, line 1316: This section is clustered together as well, please add some whitespace to make this easier to read.

crates/matrix-sdk-crypto/src/session_manager/group_sessions.rs
```diff
@@ -797,8 +797,19 @@ impl GroupSessionManager {
         let devices: Vec<_> = devices
             .into_iter()
             .flat_map(|(_, d)| {
-                d.into_iter()
-                    .filter(|d| matches!(outbound.is_shared_with(d), ShareState::NotShared))
+                d.into_iter().filter(|d| match outbound.is_shared_with(d) {
+                    ShareState::NotShared => true,
+                    ShareState::Shared(_msg_index, olm_wedging_index) => {
+                        // If the recipient device's Olm wedging index is
```
poljar, line 803: This seems to be wrapped at a really small width, we wrap at 80 chars, `cargo fmt` will do this automatically for you if the lines are too long.

crates/matrix-sdk-crypto/src/olm/account.rs
```diff
@@ -1305,12 +1305,22 @@ impl Account {
...
                 // it to the cache but don't store it.
-                store
-                    .save_changes(Changes {
-                        sessions: vec![result.session.clone()],
-                        ..Default::default()
-                    })
-                    .await?;
+                let mut changes =
+                    Changes { sessions: vec![result.session.clone()], ..Default::default() };
+                // Any new Olm session will bump the Olm wedging index for the
```
poljar, line 1310: This seems to be also wrapped manually, though it doesn't look as comical as the other doc comment, wouldn't hurt to let `cargo fmt` format this as well.

crates/matrix-sdk-crypto/src/olm/account.rs
```diff
@@ -1305,12 +1305,22 @@ impl Account {
...
+                // sender's device, if we have their device, which will cause
+                // us to re-send existing Megolm sessions to them the next time
+                // we use the session.  If we don't have their device, this
+                // means that we haven't tried to send them any Megolm sessions
+                // yet, so we don't need to worry about it.
+                if let Some(device) = store.get_device_from_curve_key(sender, sender_key).await? {
+                    let mut read_only_device = device.inner;
+                    read_only_device.olm_wedging_index += 1;
```
poljar, line 1318: Technically this can panic, what do you think about using our `SequenceNumber` type for this. Not that it's terribly important but since we have such a type, seems like we could just use it for correctness sake:
https://github.com/matrix-org/matrix-rust-sdk/blob/3be84a5a3014abc2bdd3018392d6a43c92c01b1d/crates/matrix-sdk-crypto/src/store/caches.rs#L191-L229

You would need to derive `Serialize` and `Deserialize` for it with the [`transparent`](https://serde.rs/container-attrs.html#transparent) attribute, but that's about it.

poljar (changes requested):
Quite the clever implementation, nice. I left some smaller nits we should address before we merge.

crates/matrix-sdk-crypto/src/olm/account.rs
uhoreg: I'm not seeing any difference between your suggestion and what's there.

crates/matrix-sdk-crypto/src/store/caches.rs
```diff
@@ -197,8 +198,9 @@ impl DeviceStore {
 /// subtraction. For example, suppose we've just overflowed from i64::MAX to
 /// i64::MIN. (i64::MAX.wrapping_sub(i64::MIN)) is -1, which tells us that
 /// i64::MAX comes before i64::MIN in the sequence.
-#[derive(Clone, Copy, Debug, Default, PartialEq, Eq)]
-pub(crate) struct SequenceNumber(i64);
+#[derive(Clone, Copy, Debug, Default, PartialEq, Eq, Deserialize, Serialize)]
+#[serde(transparent)]
+pub struct SequenceNumber(i64);
```
uhoreg, line 203: I had to make `SeuenceNumber` public, since it's used by other things that are public (https://github.com/matrix-org/matrix-rust-sdk/actions/runs/9685106739/job/26724426654)  Though I have a suspicion that those things aren't used outside of this crate, so maybe it's better to make those things `pub(crate)` instead?

uhoreg:
I think I've addressed all the review comments

crates/matrix-sdk-crypto/src/store/caches.rs
poljar: I think they are pub because store implementations might use them to serialize things more efficiently. I'm sure there are things that are needlessly public but this might not be such a situation.

crates/matrix-sdk-crypto/src/olm/group_sessions/outbound.rs
```diff
@@ -70,7 +71,7 @@ const ROTATION_MESSAGES: u64 = 100;
 pub(crate) enum ShareState {
     NotShared,
     SharedButChangedSenderKey,
-    Shared(u32),
+    Shared { message_index: u32, olm_wedging_index: SequenceNumber },
```
poljar, line 74: This still isn't documented.

poljar (changes requested):

crates/matrix-sdk-crypto/src/olm/group_sessions/outbound.rs
uhoreg: Sorry, I misunderstood your original comment and thought that switching to named fields would avoid the need for documentation.  I've added docs, and while I was there, added docs for the rest of the enum too.

poljar (approved):
Nice, thanks for adding the docs.

PR #3606, opened 2024-06-25
sdk: retry all requests which previous response was a plain 429 without an errcode

bnjbvr:
This can happen if there's a load-balancer or any modification of the response by a reverse proxy (e.g. rewrite 5XX errors into 429, to not let a reverse proxy mark the upstream server as being down, as Cloudflare seems to do).

As a result, such requests will be retried in multiple places, including when sending something with the send queue. Also, the send queue will mark these errors as recoverable instead of unrecoverable.

No test, because the change really is trivial and a regression test didn't seem worth it, for once.

Part of #3361.

### DIFF
--- a/crates/matrix-sdk/src/error.rs
+++ b/crates/matrix-sdk/src/error.rs
@@ -175,12 +175,15 @@ impl HttpError {
                     }
                     _ => Some(e.status_code),
                 },
-                RumaApiError::Uiaa(_) => None,
                 RumaApiError::Other(e) => Some(e.status_code),
+                RumaApiError::Uiaa(_) => None,
             };
 
             if let Some(status_code) = status_code {
-                if status_code.is_server_error() {
+                // If the status code is 429, this is requesting a retry in HTTP, without the
+                // custom `errcode`. Treat that as a retriable request with no
+                // specified retry_after delay.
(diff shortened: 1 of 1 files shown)


crates/matrix-sdk/src/error.rs
```diff
@@ -175,12 +175,15 @@ impl HttpError {
...
                     _ => Some(e.status_code),
                 },
-                RumaApiError::Uiaa(_) => None,
                 RumaApiError::Other(e) => Some(e.status_code),
+                RumaApiError::Uiaa(_) => None,
             };
 
             if let Some(status_code) = status_code {
-                if status_code.is_server_error() {
+                // If the status code is 429, this is requesting a retry in HTTP, without the
```
poljar, line 183: MDN tells us that the 429 might contain a `retry-after` value in the HTTP header[^1], are we sure that we don't want to attempt to fish that out of the headers?

[^1]: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/429

poljar: Ah right, I had trouble with this for the QR login as well. We probably should do something about this, one of these days... :smiling_face_with_tear:

poljar (approved):

Issue #3563, opened 2024-06-17
InvisibleCrypto | Share Room Keys: new RoomKeyShareStrategy based on user identities instead of per device


BillCarsonFr:
Follow up of #3562
Part of https://github.com/matrix-org/matrix-spec-proposals/pull/4153.

Introduce a new strategy to collect recipients for room_keys.
New strategy will consider user identities and only distribute to devices signed by their owner

Notes from https://github.com/element-hq/crypto-internal/issues/307:

> Introduce a new implementation of the new `RoomKeyShareStrategy` trait based on Indentities instead of the existing one based on devices. + tests

PR #3607, opened 2024-06-25
feat(sdk-crypto): Add Identity based room key sharing strategy

BillCarsonFr:
~~Draft because based on https://github.com/matrix-org/matrix-rust-sdk/pull/3605~~
Fixes https://github.com/matrix-org/matrix-rust-sdk/issues/3563

This PR just introduce a new way to distribute room keys based on identities, by only sharing with devices signed by their owner.
This is a first PR to introduce this mode, but it will be really usable once we introduce some support for identity TOFU.

Signed-off-by:

### DIFF
--- a/crates/matrix-sdk-crypto/src/identities/device.rs
+++ b/crates/matrix-sdk-crypto/src/identities/device.rs
@@ -762,7 +762,11 @@ impl ReadOnlyDevice {
         device_owner_identity: &ReadOnlyUserIdentities,
     ) -> bool {
         match device_owner_identity {
+            // If it's one of our own devices, just check that
+            // we signed the device.
             ReadOnlyUserIdentities::Own(identity) => identity.is_device_signed(self).is_ok(),
+            // If it's a device from someone else, check
+            // if the other user has signed this device.
             ReadOnlyUserIdentities::Other(device_identity) => {
                 device_identity.is_device_signed(self).is_ok()
             }
--- a/crates/matrix-sdk-crypto/src/identities/device.rs
+++ b/crates/matrix-sdk-crypto/src/identities/device.rs
@@ -764,6 +764,18 @@ impl ReadOnlyDevice {
         )
     }
 
+    pub(crate) fn is_cross_signed_by_owner(
+        &self,
+        device_owner_identity: &ReadOnlyUserIdentities,
+    ) -> bool {
+        match device_owner_identity {
+            ReadOnlyUserIdentities::Own(identity) => identity.is_device_signed(self).is_ok(),
+            ReadOnlyUserIdentities::Other(device_identity) => {
+                device_identity.is_device_signed(self).is_ok()
+            }
+        }
+    }
+
     /// Encrypt the given content for this device.
     ///
     /// # Arguments
--- a/crates/matrix-sdk-crypto/src/identities/device.rs
+++ b/crates/matrix-sdk-crypto/src/identities/device.rs
@@ -273,16 +273,9 @@ impl Device {
 
     /// Is this device cross signed by its owner?
     pub fn is_cross_signed_by_owner(&self) -> bool {
-        self.device_owner_identity.as_ref().is_some_and(|device_identity| match device_identity {
-            // If it's one of our own devices, just check that
-            // we signed the device.
-            ReadOnlyUserIdentities::Own(identity) => identity.is_device_signed(&self.inner).is_ok(),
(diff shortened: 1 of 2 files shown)


crates/matrix-sdk-crypto/src/session_manager/group_sessions/share_strategy.rs
```diff
@@ -225,6 +241,41 @@ fn split_recipients_withhelds_for_user(
...
+    device_owner_identity: &Option<ReadOnlyUserIdentities>,
+) -> RecipientDevices {
+    let mut allowed_devices: Vec<ReadOnlyDevice> = Default::default();
+    let mut denied_devices_with_code: Vec<(ReadOnlyDevice, WithheldCode)> = Default::default();
+
+    match device_owner_identity {
+        None => {
+            // withheld all the users devices, we need to have an identity for this
+            // distribution mode
+            denied_devices_with_code
```
poljar, line 255: Let's return in each branch of that match a `RecipientDevices` struct, makes it a bit clearer that we're not really mutating multiple times the above vectors and might get rid of a couple of allocations.

```suggestion
            RecipientDevices {
                denied_devices_with_code: user_devices
                    .into_values()
                    .map(|d| (d, WithheldCode::Unauthorised)),
                ..Default::default()
            }
```
(You won't be able to apply that suggestion because I can't modify multiple lines and you need further modifications in the rest of the function).

crates/matrix-sdk-crypto/src/session_manager/group_sessions/share_strategy.rs
```diff
@@ -225,6 +241,41 @@ fn split_recipients_withhelds_for_user(
     RecipientDevices { allowed_devices: recipients, denied_devices_with_code: withheld_recipients }
 }
 
+fn split_recipients_withhelds_for_user_based_on_identity(
```
poljar, line 244: This function duplicates a lot of code from the other split function, though since we're ignoring blacklisted devices it isn't quite easy to not make a mess out of a function that merges both strategies.

I guess that's fine.

crates/matrix-sdk-crypto/src/session_manager/group_sessions/share_strategy.rs
```diff
@@ -402,6 +453,81 @@ mod tests {
...
+        );
+
+        // Check withhelds for others
+        let (_, code) = share_result
+            .withheld_devices
+            .iter()
+            .find(|(d, _)| d.device_id() == KeyDistributionTestData::dan_unsigned_device_id())
+            .expect("This dan's device should receive a withheld code");
+
+        assert_eq!(code.as_str(), WithheldCode::Unauthorised.as_str());
```
poljar, line 519: Why the `as_str()` on both sides? This should be the same type, no?

poljar (changes requested):
Much better PR, thanks for taking the time to make this reviewable. I left some nits.

crates/matrix-sdk-crypto/src/identities/device.rs
```diff
@@ -764,6 +764,18 @@ impl ReadOnlyDevice {
         )
     }
 
+    pub(crate) fn is_cross_signed_by_owner(
```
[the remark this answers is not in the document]
BillCarsonFr: done

crates/matrix-sdk-crypto/src/session_manager/group_sessions/share_strategy.rs
BillCarsonFr: updated

crates/matrix-sdk-crypto/src/identities/device.rs
```diff
@@ -764,6 +757,18 @@ impl ReadOnlyDevice {
...
+        &self,
+        device_owner_identity: &ReadOnlyUserIdentities,
+    ) -> bool {
+        match device_owner_identity {
+            ReadOnlyUserIdentities::Own(identity) => identity.is_device_signed(self).is_ok(),
```
poljar, line 765: Could you please preserve the comments from the other body of this function?

poljar (changes requested):

crates/matrix-sdk-crypto/src/identities/device.rs
BillCarsonFr: done

poljar (approved):

PR #3608, opened 2024-06-25
chore: Fix more nightly warnings and bump nightly in CI

zecakeh:
This should be reviewed commit by commit.

I could not find a satisfying solution for the 3rd commit. [`dependency_on_unit_never_type_fallback`](https://doc.rust-lang.org/nightly/rustc/lints/listing/warn-by-default.html#dependency-on-unit-never-type-fallback) is a warning to prepare for the 2024 edition but it applies to the [never](https://doc.rust-lang.org/std/primitive.never.html) type that is still unstable.

It seems that [another temporary solution](https://github.com/rust-lang/rust/issues/126466#issuecomment-2171025116) is to add `#![feature(never_type, never_type_fallback)]` in the crate root. The proper solution might be to implement `EventHandlerResult` for `!`, if I understand the problem correctly.

This bumps Rust nightly in CI so the lint from the first commit is applied to future PRs.

### DIFF
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -17,7 +17,7 @@ jobs:
     - name: Install Rust
       uses: dtolnay/rust-toolchain@master
       with:
-        toolchain: nightly-2024-05-19
+        toolchain: nightly-2024-06-24
         components: rustfmt
 
     - name: Run Benchmarks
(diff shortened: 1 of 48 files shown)


Hywan (approved):
LGTM, thanks for your work!

Hywan:
About [`dependency_on_unit_never_type_fallback`](https://doc.rust-lang.org/nightly/rustc/lints/listing/warn-by-default.html#dependency-on-unit-never-type-fallback), I propose to see how rust nightly will evolve in the next weeks and adapt the code accordingly.

Hywan:
@zecakeh Can you please rebase your patch to solve the merge conflict?

zecakeh:
Rebased and fixed new rustfmt errors.

PR #3609, opened 2024-06-25
send queue: allow to respawn tasks for send queues which had unsent events

bnjbvr:
This changes the database scheme for the send queue: to be able to list all the rooms (room_id) which have unsent events, the `room_id`s need to be saved as *values* (i.e. encrypted/decrypted when the db is encrypted), and not just as *keys* (i.e. one-way hashed when the db is encrypted). To make this work, I've added another `room_id` field in the sqlite/indexeddb database, and introduced a sqlite migration (hoping no one used the indexeddb backend with the new feature since two days ago :sweat_smile:) that drops and recreates the table with the new format.

Then, we can implement a query to retrieve all the rooms which have unsent events. With that, we can add a new public function to the `SendQueue` object, which allows to automatically respawn the background room send queue tasks for each room that had unsent events, and that can be useful when restarting a Client.

This new function is then called, in the FFI layer, at the same time we set the listener for client-wide errors; this means that there's unsent events ending up in the error state would be immediately observable by a client.
 
Review can be done commit by commit.

Part of #3361.

### DIFF
--- a/bindings/matrix-sdk-ffi/src/client.rs
+++ b/bindings/matrix-sdk-ffi/src/client.rs
@@ -388,8 +388,8 @@ impl Client {
     ///
     /// This can be controlled for individual rooms, using
     /// [`Room::enable_send_queue`].
-    pub fn enable_all_send_queues(&self, enable: bool) {
-        self.inner.send_queue().set_enabled(enable);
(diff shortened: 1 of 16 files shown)


andybalaam (approved):
Looks good, thanks!

PR #3610, opened 2024-06-26
feat(sdk-crypto): Add Identity local tofu trust flag.

### LINKED ISSUE #3564: InvisibleCrypto | Share Room Keys: Add a new `tofu` flag on `ReadOnlyUserIdentity`
Part of https://github.com/element-hq/crypto-internal/issues/282 (Invisible crypto).

As per MSC4153

> Matrix client should intialy blindly trust the identity, and display non-blocking warnings when identity changes.

In order to know if an identity change has been shown to the user, we need a flag on the identity to store it.

- Update identity struct to store that flag
- Ensure migration by just marking existing identities in cache as tofu trusted
- The first time an identity is seen for a user, mark it as tofu trusted
- If an identity change is detected, store the identity and mark the flag as false
- Add a new API to mark the identity change as seen by the user (should be set by the app)
- Test

Notes from  https://github.com/element-hq/crypto-internal/issues/307:

> Add a new flag on read only identities `tofu_trusted`, that is set to true when it's the first time an identity is seen for a user. It shoud be set to false when a new identity is detected. This flag can be set back to true via API. As per tofu implementation client should ensure that this flag is only updated to true when the identity change notice has been shown to the user.

BillCarsonFr:
part of invisible crypto, follow up of https://github.com/matrix-org/matrix-rust-sdk/pull/3607
Fixes #3564

> Add a new flag on read only identities tofu_trusted, that is set to true when it's the first time an identity is seen for a user. It shoud be set to false when a new identity is detected. This flag can be set back to true via API. As per tofu implementation client should ensure that this flag is only updated to true when the identity change notice has been shown to the user.

Signed-off-by:

### DIFF
--- a/crates/matrix-sdk-crypto/src/identities/user.rs
+++ b/crates/matrix-sdk-crypto/src/identities/user.rs
@@ -375,6 +375,24 @@ pub struct ReadOnlyUserIdentity {
     user_id: OwnedUserId,
     pub(crate) master_key: Arc<MasterPubkey>,
     self_signing_key: Arc<SelfSigningPubkey>,
+    /// The first time an identity is seen for a given user it will be marked as
+    /// tofu trusted. If a new identity is detected for that user, we must
+    /// ensure that this change has been shown and validated by the user. If
+    /// this boolean is set to `false`, sharing room keys to this user might
+    /// fail depending on the room key sharing strategy.
+    #[serde(
+        default = "tofu_trusted_default",
+        serialize_with = "atomic_bool_serializer",
+        deserialize_with = "atomic_bool_deserializer"
+    )]
+    tofu_trusted: Arc<AtomicBool>,
+}
+
+// The tofu_trusted field introduced a new field in the database
+// schema, for backwards compatibility we assume that if the identity is
+// already in the database we considered it as tofu trusted.
+fn tofu_trusted_default() -> Arc<AtomicBool> {
+    Arc::new(true.into())
 }
 
 impl PartialEq for ReadOnlyUserIdentity {
@@ -395,6 +413,7 @@ impl PartialEq for ReadOnlyUserIdentity {
             && self.master_key == other.master_key
             && self.self_signing_key == other.self_signing_key
             && self.master_key.signatures() == other.master_key.signatures()
+            && self.is_tofu_trusted() == other.is_tofu_trusted()
     }
 }
 
@@ -406,12 +425,14 @@ impl ReadOnlyUserIdentity {
     /// * `master_key` - The master key of the user identity.
     ///
     /// * `self signing key` - The self signing key of user identity.
+    /// * `tofu_trusted` - Is this identity tofu trusted.
     ///
     /// Returns a `SignatureError` if the self signing key fails to be correctly
     /// verified by the given master key.
     pub(crate) fn new(
         master_key: MasterPubkey,
         self_signing_key: SelfSigningPubkey,
+        tofu_trusted: bool,
     ) -> Result<Self, SignatureError> {
         master_key.verify_subkey(&self_signing_key)?;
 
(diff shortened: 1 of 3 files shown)


crates/matrix-sdk-crypto/src/identities/user.rs
```diff
@@ -375,6 +375,24 @@ pub struct ReadOnlyUserIdentity {
...
+    /// tofu trusted. If a new identity is detected for that user, we must
+    /// ensure that this change has been shown and validated by the user. If
+    /// this boolean is set to `false`, sharing room keys to this user might
+    /// fail depending on the room key sharing strategy.
+    #[serde(
+        default = "tofu_trusted_default",
+        serialize_with = "atomic_bool_serializer",
+        deserialize_with = "atomic_bool_deserializer"
+    )]
+    tofu_trusted: Arc<AtomicBool>,
```
richvdh, line 388: `tofu_trusted` doesn't seem like a great name for for a flag that means *either* "the first identity we saw for this user" *or* "not the first identity, but the user acknowledged it". TBH, I'm wondering if we should actually split it into two different flags.

crates/matrix-sdk-crypto/src/identities/user.rs
```diff
@@ -447,6 +474,21 @@ impl ReadOnlyUserIdentity {
         &self.self_signing_key
     }
 
+    /// Check if our identity is verified.
```
richvdh, line 477: This certainly doesn't mean that it is "verified", which would imply we have done the cross-signing dance.

crates/matrix-sdk-crypto/src/identities/user.rs
```diff
@@ -406,19 +425,22 @@ impl ReadOnlyUserIdentity {
     /// * `master_key` - The master key of the user identity.
     ///
     /// * `self signing key` - The self signing key of user identity.
+    /// * `tofu_trusted` - Is this identity tofu trusted.
```
richvdh, line 428: If we're sticking with a single flag, we will need to explain what "tofu trusted" actually means each time we reference it.

crates/matrix-sdk-crypto/src/identities/user.rs
```diff
@@ -375,6 +375,24 @@ pub struct ReadOnlyUserIdentity {
...
+    #[serde(
+        default = "tofu_trusted_default",
+        serialize_with = "atomic_bool_serializer",
+        deserialize_with = "atomic_bool_deserializer"
+    )]
+    tofu_trusted: Arc<AtomicBool>,
+}
+
+// The tofu_trusted field introduced a new field in the database
+// schema, for backwards compatibility we assume that if the identity is
```
richvdh, line 392:
```suggestion
// schema. For backwards compatibility, we assume that if the identity is
```

crates/matrix-sdk-crypto/src/identities/user.rs
dkasak: Or an enum, given that the two flags would be mutually exclusive?

richvdh: > Or an enum, given that the two flags would be mutually exclusive?

I did think about that, but we still have the problem of what to call the damn thing

dkasak: I discussed this with @BillCarsonFr and semantics of the flag are roughly whether the identity is pinned or not, so perhaps the name could be `pinned`.

However, the way it's currently implemented—where we immediately replace the stored identity as soon as a new one is returned from the server, but with this flag set to `false`—means that a server can delete your pin by simply presenting a new identity and then switching back to the old one, resulting in the flag switching from `true` to `false`.

Valere is going to experiment with an alternative approach, where we

- Remember the old pinned identity as long as a new identity doesn't become pinned, replacing the old one
- Using the logic `current != pinned` to control whether the user should be notified / prompted to confirm

BillCarsonFr:
Here is the alternative proposal https://github.com/matrix-org/matrix-rust-sdk/pull/3639

crates/matrix-sdk-crypto/src/identities/user.rs
BillCarsonFr: @dkasak Alternative for MSK pinning https://github.com/matrix-org/matrix-rust-sdk/pull/3639

PR #3613, opened 2024-06-26
ui: Make public all timeline module errors

zecakeh:
A few were missing in the public API like `SendEventError` and `RedactEventError` and their dependencies.

This uses a wildcard because it should be rare to not want to expose an error type.


bnjbvr (approved):
Cheers!

PR #3614, opened 2024-06-26
ui: Reexport eyeball-im

zecakeh:
A user of the library needs to add this crate as a dependency just to be able to match on `VectorDiff`.


andybalaam (approved):
Thanks!

PR #3615, opened 2024-06-26
chore(base): Use constructors of `SyncTimelineEvent` to simplify code

Hywan:
Extracted from https://github.com/matrix-org/matrix-rust-sdk/pull/3585. Just a simple refactoring.


bnjbvr (approved):
I think we should replace uses of `sync_timeline_event!()` by using the `EventFactory` directly, since it usually reads better IMO at call sites, but this works too :)

PR #3616, opened 2024-06-26
send queue: allow editing local echoes

### LINKED ISSUE #3361: timeline: introduce a Send Queue API

bnjbvr:
Final step of #3361, as a bonus.

Fixes #3361.

### DIFF
--- a/bindings/matrix-sdk-ffi/src/timeline/mod.rs
+++ b/bindings/matrix-sdk-ffi/src/timeline/mod.rs
@@ -485,13 +485,9 @@ impl Timeline {
         item: Arc<EventTimelineItem>,
         new_content: Arc<RoomMessageEventContentWithoutRelation>,
     ) -> Result<bool, ClientError> {
-        let edit_info = item.0.edit_info().map_err(|err| anyhow::anyhow!(err))?;
+        let edit_info = item.0.edit_info().map_err(ClientError::from)?;
 
-        Ok(self
-            .inner
-            .edit((*new_content).clone(), edit_info)
-            .await
-            .map_err(|err| anyhow::anyhow!(err))?)
+        self.inner.edit((*new_content).clone(), edit_info).await.map_err(ClientError::from)
     }
 
     /// Edit an event given its event id. Useful when we're not sure a remote
@@ -502,17 +498,10 @@ impl Timeline {
         new_content: Arc<RoomMessageEventContentWithoutRelation>,
     ) -> Result<(), ClientError> {
         let event_id = EventId::parse(event_id)?;
-        let edit_info = self
-            .inner
-            .edit_info_from_event_id(&event_id)
-            .await
-            .map_err(|err| anyhow::anyhow!(err))?;
+        let edit_info =
+            self.inner.edit_info_from_event_id(&event_id).await.map_err(ClientError::from)?;
 
-        self.inner
-            .edit((*new_content).clone(), edit_info)
-            .await
-            .map_err(|err| anyhow::anyhow!(err))?;
-
+        self.inner.edit((*new_content).clone(), edit_info).await.map_err(ClientError::from)?;
         Ok(())
     }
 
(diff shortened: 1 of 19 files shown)


crates/matrix-sdk/src/send_queue.rs
```diff
@@ -698,6 +698,32 @@ impl QueueStorage {
...
+        if self.being_sent.read().await.contains(transaction_id) {
+            return Ok(false);
+        }
+
+        self.client
+            .get()
+            .ok_or(RoomSendQueueStorageError::ClientShuttingDown)?
+            .store()
+            .update_send_queue_event(&self.room_id, transaction_id, serializable)
+            .await?;
```
Hywan, line 722: Can we have a race condition here? I think no, but maybe `being_sent` can be modified before `update_send_queue_event` is called?

crates/matrix-sdk-ui/src/timeline/inner/mod.rs
```diff
@@ -805,6 +805,53 @@ impl<P: RoomDataProvider> TimelineInner<P> {
...
 
+    pub(super) async fn replace_local_echo(
+        &self,
+        txn_id: &TransactionId,
+        content: AnyMessageLikeEventContent,
+    ) -> bool {
+        let AnyMessageLikeEventContent::RoomMessage(content) = content else {
+            // Ideally, we'd support replacing local echoes for a reaction, etc., but
+            // handling RoomMessage should be sufficient in most cases. Worst
+            // case, the local echo will be sent Soon™ and we'll get another chance at
```
Hywan, line 816: “Soon™” 😄  

By the way, I agree with this. It means someone reacts to its own local event, and wants to change that. That's clearly an edge case.

bindings/matrix-sdk-ffi/src/timeline/mod.rs
```diff
@@ -470,10 +470,36 @@ impl Timeline {
...
+    /// redacted by sending an edit request to the server.
+    ///
+    /// Returns whether the edit did happen. It can only return false for
+    /// local events that are being processed.
     pub async fn edit(
         &self,
+        item: Arc<EventTimelineItem>,
         new_content: Arc<RoomMessageEventContentWithoutRelation>,
+    ) -> Result<bool, ClientError> {
+        let edit_info = item.0.edit_info().map_err(|err| anyhow::anyhow!(err))?;
```
Hywan, line 488: `ClientError::from(err)` doesn't work (to replace `anyhow!`)?

Hywan (approved):
It looks generally good to me. The fact the patches modify both the `SendQueue`, the database, and the `Timeline` make it a bit difficult to follow, but thanks to your atomic commits, it wasn't that much complicated.

Please rebase your `fixup!` patches before merging your PR.

crates/matrix-sdk/src/send_queue.rs
bnjbvr: No indeed: there's only one possible caller, the sender task. Will add a small comment.

bindings/matrix-sdk-ffi/src/timeline/mod.rs
bnjbvr: Yes, with a new conversion method. I'll add it. We should do a pass at making better errors at the FFI layer some day.

PR #3617, opened 2024-06-27
Element specific well-known support

stefanceriu:
Handles part of element-hq/element-meta/issues/2441

* exposes a generic method to make GET requests throgh the SDKs internal HTTP client. This avoids various configuration differences between the SDK stack and that of the client
* exposes a way to read the server name part of the current user's identifier (already built into ruma, make sense to use it across clients)
* adds an Element specific well-known struct that the FFI clients can use to deserialize a response JSON

Powers https://github.com/element-hq/element-x-ios/pull/2971

### DIFF
--- a/bindings/matrix-sdk-ffi/src/element.rs
+++ b/bindings/matrix-sdk-ffi/src/element.rs
@@ -1,0 +1,21 @@
+use serde::Deserialize;
+
+use crate::ClientError;
+
+/// Well-known settings specific to ElementCall
+#[derive(Deserialize, uniffi::Record)]
+pub struct ElementCallWellKnown {
+    widget_url: String,
+}
+
+/// Element specific well-known settings
+#[derive(Deserialize, uniffi::Record)]
+pub struct ElementWellKnown {
+    call: ElementCallWellKnown,
+}
+
+/// Helper function to parse a string into a ElementWellKnown struct
+#[uniffi::export]
+pub fn make_element_well_known(string: String) -> Result<ElementWellKnown, ClientError> {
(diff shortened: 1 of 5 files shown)


bnjbvr:
> exposes a generic method to make GET requests throgh the SDKs internal HTTP client. This avoids various configuration differences between the SDK stack and that of the client

I'm a bit curious, what kind of requests are you doing outside the SDK? :eyes:

stefanceriu:
> I'm a bit curious, what kind of requests are you doing outside the SDK? 👀

Have a look at the linked issue and ElementX PR 
tl;dr; fetching Element specific configs from a file

bindings/matrix-sdk-ffi/src/element.rs
```diff
@@ -0,0 +1,21 @@
...
+
+use crate::ClientError;
+
+/// Well-known settings specific to ElementCall
+#[derive(Deserialize, uniffi::Record)]
+pub struct ElementCallWellKnown {
+    widget_url: String,
+}
+
+/// Element specific well-known settings
```
poljar, line 11: Do we have some docs where this is defined? Could you please add a pointer to the docs if so.

poljar (approved):
I think that this is simple enough and fine for now, though we really need to start thinking about a Element specific crate more seriously.

Left a small nit.

bindings/matrix-sdk-ffi/src/element.rs
poljar: According insider sources, it seems that this well-known file is super secret stuff and no public documentation exists.

poljar (approved):

PR #3619, opened 2024-06-27
send queue: mark `ConcurrentRequestFailed` as recoverable

bnjbvr:
This will still disable the room's send queue, but the embedder may then decide whether to re-enable the queue or not, based on network connectivity. Ideally, we'd implement some retry mechanism here too, but since we're at a different layer than the HTTP one, we can't get that "for free", so let the embedder decide what to do here.

---

No test because the only way to trigger this is likely by having an encrypted room, start a /member query for that room, then try to send an encrypted message into that room. Sending encrypted messages require lots of endpoints, and I'm not sure whether we have all the testing setup for that; I'll ask on the Matrix room. But since this is a tiny change in behavior (and we're likely to have other tweaks like that related to error handling, I don't think each new error type marked as recoverable would need to require its own test).

---

Part of #3361.

### DIFF
--- a/crates/matrix-sdk/src/error.rs
+++ b/crates/matrix-sdk/src/error.rs
@@ -320,8 +320,9 @@ pub enum Error {
     #[error("a concurrent request failed; see logs for details")]
     ConcurrentRequestFailed,
 
-    /// An other error was raised
-    /// this might happen because encryption was enabled on the base-crate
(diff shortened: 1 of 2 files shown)


poljar (approved):
Makes sense, a bit annoying that we don't have access to all the erorr data we want.

PR #3621, opened 2024-06-28
sdk: implement MSC3489 live location sharing

torrybr:
Add support for MSC3489 live location sharing. Requesting an initial review to clarify some technical questions and help get this over the finish line.

Building on this [Ruma PR](https://github.com/ruma/ruma/pull/1722)

Signed-off-by:

### DIFF
--- a/bindings/matrix-sdk-ffi/src/event.rs
+++ b/bindings/matrix-sdk-ffi/src/event.rs
@@ -119,6 +119,7 @@ impl TryFrom<AnySyncStateEvent> for StateEventContent {
 
 #[derive(uniffi::Enum)]
 pub enum MessageLikeEventContent {
+    Beacon,
     CallAnswer,
     CallInvite,
     CallNotify { notify_type: NotifyType },
@@ -144,6 +145,7 @@ impl TryFrom<AnySyncMessageLikeEvent> for MessageLikeEventContent {
 
     fn try_from(value: AnySyncMessageLikeEvent) -> anyhow::Result<Self> {
         let content = match value {
+            AnySyncMessageLikeEvent::Beacon(_) => MessageLikeEventContent::Beacon,
             AnySyncMessageLikeEvent::CallAnswer(_) => MessageLikeEventContent::CallAnswer,
             AnySyncMessageLikeEvent::CallInvite(_) => MessageLikeEventContent::CallInvite,
             AnySyncMessageLikeEvent::CallNotify(content) => {
@@ -230,6 +232,7 @@ where
 
 #[derive(Clone, uniffi::Enum)]
 pub enum StateEventType {
+    BeaconInfo,
     CallMember,
     PolicyRuleRoom,
     PolicyRuleServer,
@@ -257,6 +260,7 @@ pub enum StateEventType {
 impl From<StateEventType> for ruma::events::StateEventType {
     fn from(val: StateEventType) -> Self {
         match val {
+            StateEventType::BeaconInfo => Self::BeaconInfo,
             StateEventType::CallMember => Self::CallMember,
             StateEventType::PolicyRuleRoom => Self::PolicyRuleRoom,
             StateEventType::PolicyRuleServer => Self::PolicyRuleServer,
@@ -285,6 +289,7 @@ impl From<StateEventType> for ruma::events::StateEventType {
 
 #[derive(Clone, uniffi::Enum)]
 pub enum MessageLikeEventType {
+    Beacon,
     CallAnswer,
     CallCandidates,
     CallHangup,
@@ -313,6 +318,7 @@ pub enum MessageLikeEventType {
 impl From<MessageLikeEventType> for ruma::events::MessageLikeEventType {
     fn from(val: MessageLikeEventType) -> Self {
         match val {
+            MessageLikeEventType::Beacon => Self::Beacon,
             MessageLikeEventType::CallAnswer => Self::CallAnswer,
             MessageLikeEventType::CallInvite => Self::CallInvite,
             MessageLikeEventType::CallNotify => Self::CallNotify,
--- a/bindings/matrix-sdk-ffi/src/room.rs
+++ b/bindings/matrix-sdk-ffi/src/room.rs
@@ -160,6 +160,20 @@ impl Room {
         }
     }
 
+    pub async fn send_beacon_info(&self, duration_millis: u64) -> Result<(), ClientError> {
+        RUNTIME.block_on(async move {
+            self.inner.send_beacon_info(duration_millis).await?;
+            Ok(())
+        })
+    }
+
+    pub async fn stop_beacon_info(&self) -> Result<(), ClientError> {
+        RUNTIME.block_on(async move {
+            self.inner.stop_beacon_info().await?;
+            Ok(())
+        })
+    }
+
     /// Forces the currently active room key, which is used to encrypt messages,
     /// to be rotated.
     ///
--- a/bindings/matrix-sdk-ffi/src/timeline/content.rs
+++ b/bindings/matrix-sdk-ffi/src/timeline/content.rs
@@ -23,7 +23,7 @@ use ruma::events::{
 use tracing::warn;
 
 use super::ProfileDetails;
-use crate::ruma::{ImageInfo, MessageType, PollKind};
+use crate::ruma::{ImageInfo, LocationContent, MessageType, PollKind};
 
 #[derive(Clone, uniffi::Object)]
 pub struct TimelineItemContent(pub(crate) matrix_sdk_ui::timeline::TimelineItemContent);
@@ -42,6 +42,29 @@ impl TimelineItemContent {
                     body: content.body.clone(),
                     info: (&content.info).into(),
                     source: Arc::new(MediaSource::from(content.source.clone())),
+                }
+            }
+            Content::BeaconInfoState(beacon_state) => {
+                let Some(location) = beacon_state.last_location() else {
+                    return TimelineItemContentKind::FailedToParseMessageLike {
+                        event_type: "org.matrix.msc3672.beacon".to_string(),
+                        error: "Could not find beacon last location content".to_string(),
+                    };
+                };
+
+                let body = location.description.unwrap_or_else(|| "Location".to_string());
+
+                let location = LocationContent {
+                    body,
+                    geo_uri: location.uri,
+                    description: None,
+                    zoom_level: None,
+                    asset: None,
+                };
+
+                TimelineItemContentKind::BeaconInfoState {
+                    location,
+                    user_id: String::from(beacon_state.user_id()),
                 }
             }
             Content::Poll(poll_state) => TimelineItemContentKind::from(poll_state.results()),
@@ -110,6 +133,10 @@ impl TimelineItemContent {
 
 #[derive(uniffi::Enum)]
 pub enum TimelineItemContentKind {
+    BeaconInfoState {
+        location: LocationContent,
+        user_id: String,
+    },
     Message,
     RedactedMessage,
     Sticker {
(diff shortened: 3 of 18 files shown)


crates/matrix-sdk-ui/src/timeline/beacons.rs
```diff
@@ -0,0 +1,95 @@
...
+        user_id: OwnedUserId,
+    ) -> Self {
+        match &content {
+            FullStateEventContent::Original { content, .. } => BeaconState {
+                beacon_info_event_content: content.clone(),
+                last_location: None,
+                user_id,
+            },
+            FullStateEventContent::Redacted(_) => {
+                todo!("How should this be handled?")
```
torrybr, line 36: Could use some direction on how to handle this case?

bindings/matrix-sdk-ffi/src/timeline/mod.rs
```diff
@@ -521,6 +523,49 @@ impl Timeline {
...
+            .get_state_event_static_for_key::<BeaconInfoEventContent, _>(
+                self.inner.room().own_user_id(),
+            )
+            .await?
+        else {
+            todo!("How to handle case of missing beacon event for state key?")
+        };
+
+        match raw_event.deserialize() {
+            Ok(SyncOrStrippedState::Sync(SyncStateEvent::Original(beacon_info))) => {
```
torrybr, line 549: I believe the only `SyncState` that I care about is `SyncStateEvent::Original``?

crates/matrix-sdk-ui/src/timeline/event_handler.rs
```diff
@@ -655,6 +685,47 @@ impl<'a, 'o> TimelineEventHandler<'a, 'o> {
...
+    /// TODO: Handle case of `m.beacon_info` event that is identical to the first, except with live: false.
+    /// This stops the live location share.
+    ///
+    /// TODO: How do you search for an existing beacon_info event in the timeline by state key?
+    fn handle_beacon_info(
+        &mut self,
+        c: FullStateEventContent<BeaconInfoEventContent>,
+        user_id: OwnedUserId,
+    ) {
+        let mut beacon_state = BeaconState::new(c, user_id.clone());
```
torrybr, line 699: Use case of an m.beacon_info event that is identical to the first, except with live: false..

How do you search for an existing `beacon_info` event in the timeline by `state_key` in order to update the `beacon_info_state`?

bindings/matrix-sdk-ffi/src/event.rs
```diff
@@ -119,6 +119,7 @@ impl TryFrom<AnySyncStateEvent> for StateEventContent {
 
 #[derive(uniffi::Enum)]
 pub enum MessageLikeEventContent {
+    Beacon,
```
bnjbvr, line 122: Shouldn't this contain the location too?

bindings/matrix-sdk-ffi/src/room.rs
```diff
@@ -160,6 +160,20 @@ impl Room {
...
+            self.inner.send_beacon_info(duration_millis).await?;
+            Ok(())
+        })
+    }
+
+    pub async fn stop_beacon_info(&self) -> Result<(), ClientError> {
+        RUNTIME.block_on(async move {
+            self.inner.stop_beacon_info().await?;
+            Ok(())
+        })
```
bnjbvr, line 174: In both cases: no need for `RUNTIME.block_on` since the methods are async.

crates/matrix-sdk-ui/src/timeline/event_handler.rs
```diff
@@ -655,6 +682,53 @@ impl<'a, 'o> TimelineEventHandler<'a, 'o> {
         }
     }
 
+    /// Handle a beacon_info event and update the `BeaconInfoState`.
+    ///
+    /// TODO: Handle case of `m.beacon_info` event that is identical to the
+    /// first, except with live: false. This stops the live location share.
+    ///
+    /// TODO: How do you search for an existing beacon_info event in the
+    /// timeline by state key?
```
bnjbvr, line 691: If you have to do something like this, then the Timeline isn't the right place to do it; the timeline should eventually just be considered a "view" over events, not a model / source of truth in any ways, except when it comes to the view model itself.

crates/matrix-sdk-ui/src/timeline/event_handler.rs
```diff
@@ -655,6 +682,53 @@ impl<'a, 'o> TimelineEventHandler<'a, 'o> {
...
+    fn handle_beacon_info(
+        &mut self,
+        c: FullStateEventContent<BeaconInfoEventContent>,
+        should_add: bool,
+        user_id: OwnedUserId,
+    ) {
+        let mut beacon_state = BeaconState::new(c, user_id.clone());
+        if let Flow::Remote { event_id, .. } = self.ctx.flow.clone() {
```
bnjbvr, line 699: nit: use `&self.ctx.flow` instead of cloning

crates/matrix-sdk-ui/src/timeline/event_handler.rs
```diff
@@ -655,6 +682,53 @@ impl<'a, 'o> TimelineEventHandler<'a, 'o> {
...
+            let beacon_state =
+                as_variant!(event_item.content(), TimelineItemContent::BeaconInfoState)?;
+            Some(event_item.with_content(
+                TimelineItemContent::BeaconInfoState(beacon_state.update_beacon(&c)),
+                None,
+            ))
+        });
+
+        if !found {
+            warn!("Did not find beacon_info event in timeline items. Adding to pending list");
```
bnjbvr, line 727: `warn` should be kept for non-fatal errors.

```suggestion
            debug!("Did not find beacon_info event in timeline items. Adding to pending list");
```

bindings/matrix-sdk-ffi/src/timeline/content.rs
```diff
@@ -44,6 +44,29 @@ impl TimelineItemContent {
                     source: Arc::new(MediaSource::from(content.source.clone())),
                 }
             }
+            Content::BeaconInfoState(beacon_state) => {
+                let Some(location) = beacon_state.last_location() else {
+                    return TimelineItemContentKind::FailedToParseMessageLike {
+                        event_type: "org.matrix.msc3672.beacon".to_string(),
+                        error: "Could not find beacon last location content".to_string(),
+                    };
+                };
```
bnjbvr, line 53: If the last location was not set, then I think the TimelineItemContent should not exist in the first place.

bindings/matrix-sdk-ffi/src/timeline/mod.rs
```diff
@@ -521,6 +526,51 @@ impl Timeline {
...
+    /// Retrieves the last beacon_info from the room state and sends a beacon
+    /// with the geo_uri. Since only one active beacon_info per user per
+    /// room is allowed, we can grab the currently live beacon_info event
+    /// from the room state.
+    ///
+    /// TODO: Does the logic belong in self.inner.room() or here?
+    pub async fn send_user_location_beacon(
+        self: Arc<Self>,
+        geo_uri: String,
+    ) -> Result<(), ClientError> {
```
bnjbvr, line 541: There's way too much logic in this function; it should not live at the FFI layer, otherwise it's ~impossible to test.

crates/matrix-sdk-ui/src/timeline/beacons.rs
```diff
@@ -0,0 +1,97 @@
+//! This module handles rendering of MSC3489 live location sharing in the
+//! timeline.
```
bnjbvr, line 2: I don't think having this in the timeline is the proper cutting point. Instead, this would be more useful to have at the SDK layer, and maybe even with a slightly different API.

bnjbvr (changes requested):
Thanks for the PR! supporting live location sharing in the SDK will be grand!

So, I appreciate the amount of code you've written here, and that it might be sufficient to have the _basic_ feature working. However, I think the work here could be split (to make reviews faster) and layered in different ways:

- support for the base crate and SDK could go in a separate PR
- sending the beacon events themselves could happen in the SDK main crate too (I haven't seen any code for this, I think?)
- better handling of beacon events coming before their associated beacon info event could be another PR.
- then support for the timeline / FFI bindings

Each PR would also have tests at its own level, to make sure that it's behaving as intended. We're requiring tests for major functionality like this, otherwise it's too big a risk to discover later that it may be broken in subtle ways.

In fact, I think there's also potential for a more ergonomic, higher level API for beacons:

- provide a stream to listen to new live location shares (client wide)
- for each live location share, a stream to listen to new beacon events, and updates about the beacon info itself (stopped/redacted/etc.)

Then, the timeline could spawn a task, and provide UI updates based on both stream updates.

How does that sound?

bnjbvr:
Also, we'd need some way to implement redaction of all the beacon events related to a beacon info event, as [indicated by the MSC](https://github.com/matrix-org/matrix-spec-proposals/blob/matthew/location-streaming/proposals/3489-location-streaming.md#redacting-shared-locations).

torrybr:
@bnjbvr  Thanks for taking time to provide such detailed feedback! I agree with all your points and can work on splitting this up into simpler PRs and layer it as described. The stream API is good approach and would make life a lot easier 😃

crates/matrix-sdk/src/room/mod.rs
```diff
@@ -2499,6 +2497,59 @@ impl Room {
...
+    /// Find existing beacon_info by state key and set the `live` field to
+    /// `false` in the existing beacon_info event.
+    ///
+    /// TODO: Update the return type to be specific.
+    pub async fn stop_beacon_info(&self) -> Result<send_state_event::v3::Response> {
+        let Some(raw_event) = self
+            .get_state_event_static_for_key::<BeaconInfoEventContent, _>(self.own_user_id())
+            .await?
+        else {
+            todo!("How to handle case of missing beacon event for state key?")
```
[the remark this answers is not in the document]
torrybr: How does this look? 

```
/// Errors that can happen when interacting with the beacon API.
#[derive(Debug, Error)]
pub enum BeaconError {
    #[error("Network error: {0}")]
    Network(#[from] HttpError),

    #[error("Existing beacon information not found.")]
    NotFound,

    #[error("Beacon event is redacted and cannot be processed.")]
    Redacted,

    #[error("Must join the room to access beacon information.")]
    Stripped,
}

// Not sure on the conversion from Error to BeaconError
impl From<Error> for BeaconError {
    fn from(_: Error) -> Self {
        BeaconError::NotFound
    }
}
```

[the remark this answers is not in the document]
torrybr: Great Idea! I will incorporate your feedback into a new smaller PR which focuses on mainly on `matrix-sdk-base` and `matrix-sdk`. Thanks again for the direction!

torrybr:
> In fact, I think there's also potential for a more ergonomic, higher level API for beacons:
> - provide a stream to listen to new live location shares (client wide)
> - for each live location share, a stream to listen to new beacon events, and updates about the beacon info itself 
> (stopped/redacted/etc.)

@bnjbvr Do you have an example of streams being used in the codebase? This should help me get some more direction once I get to this.

bnjbvr:
> @bnjbvr Do you have an example of streams being used in the codebase? This should help me get some more direction once I get to this.

You can look for functions which names end up with `_stream` :)

torrybr:
@bnjbvr I'm working on handling beacon events and associating them with their beacon info events, as mentioned in 

> better handling of beacon events coming before their associated beacon info event could be another PR

Any guidance on where to place this logic specifically where I should be grabbing the beacon events from?

bnjbvr:
> @bnjbvr I'm working on handling beacon events and associating them with their beacon info events, as mentioned in
> 
> > better handling of beacon events coming before their associated beacon info event could be another PR
> 
> Any guidance on where to place this logic specifically where I should be grabbing the beacon events from?

Sweet! I suppose that could go in a subsequent step, when it's time to display/render the beacons themselves, in the timeline. As far as I understand, the support should be pretty minimal/straightforward in the SDK, and ordering is only important as soon as you want to display something, right? Happy to chat about this in the Matrix SDK dev room, if needs be.

Issue #3445, opened 2024-05-23
Receiving backup keys from secret send doesn't update the recovery state


kegsay:
From an end-user perspective on EX this means:
 - they login
 - they are forced to verify
 - they do so
 - they still see the prompt for their recovery key

<img width="249" alt="Screenshot 2024-05-23 at 08 37 17" src="https://github.com/matrix-org/matrix-rust-sdk/assets/7190048/00e990fb-bcd2-475a-bd2b-05ca28aa0f03">

This should not be the case because, post verification, keys are gossiped.

This fixes itself if the app is force quit and restarted.

[Rageshake](https://github.com/element-hq/element-x-ios-rageshakes/issues/1834#issuecomment-2126466456)

poljar:
This is a bit of a tricky one to reproduce. We do listen for `m.secret.send` events to be received to do two things:

1. Enable the backup if it's a backup recovery/decryption key that was recevied.
2. Update the recovery state, in case any of the cross-signing keys *or* the backup recovery/decryption key were received.

So we do update the recovery state after we receive stuff over secret send, but crucially the enabling of the backup and the updating of the recovery state are in a bit of a race. If we update the recovery state before we enable the backup then we're left in the `Incomplete` state until something else triggers and update.

Relevant code snippet is found here: https://github.com/matrix-org/matrix-rust-sdk/blob/9b05d0d822b9f93adc01c6d54f41f7b0fc8b9cf8/crates/matrix-sdk/src/encryption/recovery/mod.rs#L502-L505

PR #3583, opened 2024-06-19
 test: Add a test to verify that interactive self-verification works

poljar:
This also checks that secrets are gossiped from one device to the other and that the recovery and backup states are correctly updated.

PR #3623, opened 2024-06-28
recovery: Fix the recovery state not being updated after backups are enabled via secret send

poljar:
A review commit by commit is the best way to go through this PR, please take a look at the commit message of the last commit for a full explanation on how the fix works.

A test for this was written in #3583, but sadly it doesn't reproduce the issue since the issue is only triggered if event handlers are run in the wrong order. This PR doesn't rely on the event handler for the backup state, so the order doesn't matter anymore.

This closes: #3445.

### DIFF
--- a/crates/matrix-sdk/src/encryption/backups/mod.rs
+++ b/crates/matrix-sdk/src/encryption/backups/mod.rs
@@ -926,7 +926,6 @@ impl Backups {
     /// removed on the homeserver.
     async fn handle_deleted_backup_version(&self, olm_machine: &OlmMachine) -> Result<(), Error> {
         olm_machine.backup_machine().disable_backup().await?;
-        self.client.encryption().recovery().update_state_after_backup_disabling().await;
         self.set_state(BackupState::Unknown);
 
         Ok(())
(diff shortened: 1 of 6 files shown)


bnjbvr (commented):
Looks good at first sight, but:

1. the commit message and the PR description mention an event handler being removed, and no event handler seems to be removed here. Should the commit message be updated?
2. Can you explain the changes in the test, and confirm they're testing the new change in behavior, and if so, how/why, please?

poljar:
> Looks good at first sight, but:
> 
>     1. the commit message and the PR description mention an event handler being removed, and no event handler seems to be removed here. Should the commit message be updated?

Oh, removal might be the bad word for this. Previously we relied on the `m.secret.send` event handler to update our backup state, but that turned out to be flaky because of the ordering issue explained in the commit. We still need the event handler for other secret types, i.e. the cross-signing keys, we can't remove it. What changed is, we don't rely on event handler for the backup state update. 

>     2. Can you explain the changes in the test, and confirm they're testing the new change in behavior, and if so, how/why, please?

The tests now allow some requests to be done more times, i.e. now we're updating the state when:

1. The secret send event handler is run, due to point no. 1
2. The backup state actually changes

bnjbvr (approved):
Thanks; please tweak the commit message to make it clear the event handler is not removed, but that we stop relying on it to update the recovery state. Then let's use a PR name according to our guidelines, and we're good to go :+1:

PR #3624, opened 2024-06-28
ci: Add a github workflow to fail when fixup commits exist

poljar:
We tend to use fixup commits quite a bit, and in the heat of the moment we sometimes forget to squash them before the final merge.

This should prevent us from doing so.


bnjbvr (approved):
Sweet!

PR #3625, opened 2024-06-28
sdk: Allow to limit the number of concurrent network requests

gnunicorn:
Add a new `max_concurrent_requests` parameter in the `RequestConfig` limits the number of http(s) requests the internal sdk client issues concurrently (if > 0). The default behavior is the same as before: there is no limit on concurrent requests issued.

This is especially useful for resource constrained platforms (e.g. mobile platforms), and if your pattern might lead to issuing many requests at the same time (like downloading and caching all avatars at startup).

Signed-off-by: Benjamin Kampmann <ben.kampmann@gmail.com>

### DIFF
--- a/crates/matrix-sdk/src/config/request.rs
+++ b/crates/matrix-sdk/src/config/request.rs
@@ -44,18 +44,21 @@ pub struct RequestConfig {
     pub(crate) timeout: Duration,
     pub(crate) retry_limit: Option<u64>,
     pub(crate) retry_timeout: Option<Duration>,
+    pub(crate) max_concurrent_requests: usize,
     pub(crate) force_auth: bool,
 }
 
 #[cfg(not(tarpaulin_include))]
 impl Debug for RequestConfig {
     fn fmt(&self, fmt: &mut fmt::Formatter<'_>) -> fmt::Result {
-        let Self { timeout, retry_limit, retry_timeout, force_auth } = self;
+        let Self { timeout, retry_limit, retry_timeout, force_auth, max_concurrent_requests } =
+            self;
 
         let mut res = fmt.debug_struct("RequestConfig");
         res.field("timeout", timeout)
             .maybe_field("retry_limit", retry_limit)
-            .maybe_field("retry_timeout", retry_timeout);
+            .maybe_field("retry_timeout", retry_timeout)
+            .field("max_concurrent_requests", max_concurrent_requests);
 
         if *force_auth {
             res.field("force_auth", &true);
@@ -71,6 +74,7 @@ impl Default for RequestConfig {
             timeout: DEFAULT_REQUEST_TIMEOUT,
             retry_limit: Default::default(),
             retry_timeout: Default::default(),
+            max_concurrent_requests: 0,
             force_auth: false,
         }
     }
@@ -106,6 +110,22 @@ impl RequestConfig {
         self
     }
 
+    /// The total limit of request that are pending or run concurrently.
+    /// Any additional request beyond that number will be waiting until another
+    /// concurrent requests finished. Requests are queued fairly.
+    #[must_use]
+    pub fn max_concurrent_requests(mut self, limit: usize) -> Self {
+        self.max_concurrent_requests = limit;
+        self
+    }
+
+    /// Disable the limit of concurrent requests. Setting the limit to 0
+    /// has the same effect.
+    #[must_use]
+    pub fn disable_max_concurrent_requests(mut self) -> Self {
+        self.max_concurrent_requests = 0;
+        self
+    }
     /// Set the timeout duration for all HTTP requests.
     #[must_use]
     pub fn timeout(mut self, timeout: Duration) -> Self {
--- a/crates/matrix-sdk/src/config/request.rs
+++ b/crates/matrix-sdk/src/config/request.rs
@@ -14,6 +14,7 @@
 
 use std::{
     fmt::{self, Debug},
+    num::NonZeroUsize,
     time::Duration,
 };
 
@@ -44,7 +45,7 @@ pub struct RequestConfig {
     pub(crate) timeout: Duration,
     pub(crate) retry_limit: Option<u64>,
     pub(crate) retry_timeout: Option<Duration>,
-    pub(crate) max_concurrent_requests: usize,
+    pub(crate) max_concurrent_requests: Option<NonZeroUsize>,
     pub(crate) force_auth: bool,
 }
 
@@ -58,7 +59,7 @@ impl Debug for RequestConfig {
         res.field("timeout", timeout)
             .maybe_field("retry_limit", retry_limit)
             .maybe_field("retry_timeout", retry_timeout)
-            .field("max_concurrent_requests", max_concurrent_requests);
+            .maybe_field("max_concurrent_requests", max_concurrent_requests);
 
         if *force_auth {
             res.field("force_auth", &true);
@@ -74,7 +75,7 @@ impl Default for RequestConfig {
             timeout: DEFAULT_REQUEST_TIMEOUT,
             retry_limit: Default::default(),
             retry_timeout: Default::default(),
-            max_concurrent_requests: 0,
+            max_concurrent_requests: Default::default(),
             force_auth: false,
         }
     }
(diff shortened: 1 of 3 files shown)


crates/matrix-sdk/src/config/request.rs
```diff
@@ -44,18 +44,21 @@ pub struct RequestConfig {
     pub(crate) timeout: Duration,
     pub(crate) retry_limit: Option<u64>,
     pub(crate) retry_timeout: Option<Duration>,
+    pub(crate) max_concurrent_requests: usize,
```
bnjbvr, line 47: Instead of a `usize` that has a sentinel value (0), could we use `Option<NonZero...` here (everywhere in this PR)?

crates/matrix-sdk/src/http_client/mod.rs
```diff
@@ -48,16 +49,45 @@ pub(crate) use native::HttpSettings;
...
+            None => MaybeSemaphorePermit(None),
+        }
+    }
+}
+
 #[derive(Clone, Debug)]
 pub(crate) struct HttpClient {
     pub(crate) inner: reqwest::Client,
     pub(crate) request_config: RequestConfig,
+    queue: MaybeSemaphore,
```
bnjbvr, line 79: nit: can we name this `concurrent_request_semaphore`, instead? `queue` evokes an array/deque of things to me, which isn't the case here

bnjbvr (changes requested):
I like the implementation; just a few nits here and there. Thanks!

bnjbvr:
(Also would be nice to test, but likely it's a lot of bother for little value, so not mandatory. For this one time :))

crates/matrix-sdk/src/config/request.rs
gnunicorn: ha. had that before but saw that the other values in the config are documented as "0 means the same" and adapted to that. Happy to change it to that.

crates/matrix-sdk/src/http_client/mod.rs
```diff
@@ -48,16 +50,47 @@ pub(crate) use native::HttpSettings;
...
+impl MaybeSemaphore {
+    fn new(max: Option<NonZeroUsize>) -> Self {
+        let inner = max.map(|i| Semaphore::new(i.into()));
+        MaybeSemaphore(Arc::new(inner))
+    }
+
+    async fn acquire(&self) -> MaybeSemaphorePermit<'_> {
+        match self.0.as_ref() {
+            Some(inner) => {
+                // ignoring errors as we never close this
```
bnjbvr, line 68: What kind of errors do you mean here?

crates/matrix-sdk/src/http_client/mod.rs
```diff
@@ -259,3 +295,107 @@ impl tower::Service<http_old::Request<Bytes>> for HttpClient {
...
+        let inner_counter = counter.clone();
+
+        Mock::given(method("GET"))
+            .respond_with(move |_req: &Request| {
+                inner_counter.fetch_add(1, Ordering::SeqCst);
+                // we stall the requests
+                ResponseTemplate::new(200).set_delay(Duration::from_secs(60))
+            })
+            .mount(&server)
+            .await;
```
bnjbvr, line 380: The test is failing, likely because the client may do another `GET` request on start, or something like this; using the actual `whoami` endpoint URL should suffice to avoid this.

crates/matrix-sdk/src/http_client/mod.rs
```diff
@@ -259,3 +295,107 @@ impl tower::Service<http_old::Request<Bytes>> for HttpClient {
...
+        set_client_session(&mut client).await;
+
+        let counter = Arc::new(AtomicU8::new(0));
+        let inner_counter = counter.clone();
+
+        Mock::given(method("GET"))
+            .respond_with(move |_req: &Request| {
+                inner_counter.fetch_add(1, Ordering::SeqCst);
+                // we stall the requests
+                ResponseTemplate::new(200).set_delay(Duration::from_secs(60))
```
bnjbvr, line 334: Here and in the other test: does it mean the test needs approximately 60 seconds to run? If yes, can we lower the 60 to 3 here, instead?

crates/matrix-sdk/src/http_client/mod.rs
```diff
@@ -259,3 +295,107 @@ impl tower::Service<http_old::Request<Bytes>> for HttpClient {
...
+
+            for _n in 0..10 {
+                pollers.push(client.whoami());
+            }
+            // issue parallel execution
+            futures_util::future::join_all(pollers).await
+        });
+
+        // give it a moment to issue the requests
+        tokio::time::sleep(Duration::from_secs(2)).await;
```
bnjbvr, line 350: Can this be smaller, like 1 sec, or even 300ms?

crates/matrix-sdk/src/http_client/mod.rs
```diff
@@ -259,3 +295,107 @@ impl tower::Service<http_old::Request<Bytes>> for HttpClient {
...
+            .await;
+
+        let _bg_task = tokio::spawn(async move {
+            let mut pollers = Vec::new();
+
+            for _n in 0..10 {
+                pollers.push(client.whoami());
+            }
+            // issue parallel execution
+            futures_util::future::join_all(pollers).await
```
bnjbvr, line 346: This might be slightly more concise:

```suggestion
            futures_util::future::join_all((0..10).map(|_| client.whoami()).await
```

bnjbvr (changes requested):
Thanks for adding a test, this is on the right track! Might need a tweak to the test, a bit of Clippy and `cargo xtask fixup style`, and that should be good to go.

crates/matrix-sdk/src/http_client/mod.rs
gnunicorn: > If the semaphore has been closed, this returns an [AcquireError](https://docs.rs/tokio/latest/tokio/sync/struct.AcquireError.html). 

This only ever happens when there the Semphore was explicitly closed, which this `MaybeSemaphore` doesn't ever do. So the error can never occur and can be ignored.

crates/matrix-sdk/src/http_client/mod.rs
gnunicorn: Yes, with the spawn-move this might work now (had this before but had Lifetime issue for client in that circumstance).

crates/matrix-sdk/src/http_client/mod.rs
gnunicorn: really? didn't here locally ... I'll add that

crates/matrix-sdk/src/http_client/mod.rs
gnunicorn: No. We never await for this to actually finish. It takes only the delay time of time.

crates/matrix-sdk/src/http_client/mod.rs
gnunicorn: yes.

crates/matrix-sdk/src/http_client/mod.rs
gnunicorn: I added an `abort` to make that more explicit.

bnjbvr (changes requested):
New test failure when the test is running in the super-slow code coverage run: https://github.com/matrix-org/matrix-rust-sdk/actions/runs/9760610496/job/26939845059?pr=3625#step:9:1663

crates/matrix-sdk/src/http_client/mod.rs
Hywan: Can you please update the comment to make this clear?

crates/matrix-sdk/src/http_client/mod.rs
```diff
@@ -48,16 +50,47 @@ pub(crate) use native::HttpSettings;
 
 pub(crate) const DEFAULT_REQUEST_TIMEOUT: Duration = Duration::from_secs(30);
 
+#[derive(Clone, Debug)]
+struct MaybeSemaphore(Arc<Option<Semaphore>>);
+
+#[allow(dead_code)] // holding this until drop is all we are doing
```
Hywan, line 56: Is it still dead code?

Hywan (dismissed):

gnunicorn:
@bnjbvr yeah, looks like that 300ms wasn't long enough. I suppose the only way to fix it is to increase it to 1s again for that second test :( .

```
 assertion `left == right` failed: Not all requests passed through
  left: 163
 right: 254
```

crates/matrix-sdk/src/http_client/mod.rs
gnunicorn: well, yes, the linter is complaining about it because the inner object is never used ... which is fine we only want to keep it for the drop ... thus it is a false-positive lint error. I left the comment but maybe that doesn't explain it well enough then?

Hywan: The comment needs some improvement indeed 😀

bnjbvr (approved):
Sweet, thanks!

PR #3628, opened 2024-06-30
sdk: Use strongly-typed strings where it makes sense

zecakeh:
Where a string is clearly documented as a room ID, event ID or mxc URI, use `OwnedRoomId`, `OwnedEventId` and `OwnedMxcURI`, respectively.

Leave the conversion to the bindings.


poljar (approved):
Thanks, looks good.

PR #3630, opened 2024-07-01
send queue: add useful logs when aborting/editing a local echo

bnjbvr:
These logs may help with debugging issues around editing of local echoes.

Part of #3361.

### DIFF
--- a/crates/matrix-sdk/src/send_queue.rs
+++ b/crates/matrix-sdk/src/send_queue.rs
@@ -863,14 +863,21 @@ impl SendHandle {
     ///
     /// Returns true if the sending could be aborted, false if not (i.e. the
     /// event had already been sent).
+    #[instrument(skip(self), fields(room_id = %self.room.inner.room.room_id(), txn_id = %self.transaction_id))]
     pub async fn abort(self) -> Result<bool, RoomSendQueueStorageError> {
+        trace!("received an abort request");
+
         if self.room.inner.queue.cancel(&self.transaction_id).await? {
+            trace!("successful abort");
+
             // Propagate a cancelled update too.
             let _ = self.room.inner.updates.send(RoomSendQueueUpdate::CancelledLocalEvent {
                 transaction_id: self.transaction_id.clone(),
             });
+
             Ok(true)
         } else {
+            debug!("event was being sent, can't abort");
             Ok(false)
         }
     }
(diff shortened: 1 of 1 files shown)


crates/matrix-sdk/src/send_queue.rs
```diff
@@ -863,14 +863,21 @@ impl SendHandle {
     ///
     /// Returns true if the sending could be aborted, false if not (i.e. the
     /// event had already been sent).
+    #[instrument]
     pub async fn abort(self) -> Result<bool, RoomSendQueueStorageError> {
+        trace!(tnx_id = %self.transaction_id, "received an abort request");
```
poljar, line 868: That's already part of the span because `instrument` will attach whatever is in `self` to the span by default.

crates/matrix-sdk/src/send_queue.rs
```diff
@@ -879,21 +886,28 @@ impl SendHandle {
     ///
     /// Returns true if the event to be sent was replaced, false if not (i.e.
     /// the event had already been sent).
+    #[instrument(skip(new_content))]
     pub async fn edit_raw(
         &self,
         new_content: Raw<AnyMessageLikeEventContent>,
         event_type: String,
     ) -> Result<bool, RoomSendQueueStorageError> {
+        trace!(tnx_id = %self.transaction_id, "received an edit request");
```
poljar, line 895: Again, `self`, `event_type` are already part of the span.

poljar (changes requested):

crates/matrix-sdk/src/send_queue.rs
bnjbvr: Thanks! Will probably skip `self`, and manually record the fields I'm interested in at the `instrument` derive level.

crates/matrix-sdk/src/send_queue.rs
```diff
@@ -879,21 +886,28 @@ impl SendHandle {
     ///
     /// Returns true if the event to be sent was replaced, false if not (i.e.
     /// the event had already been sent).
+    #[instrument(skip(self, new_content), fields(room_id = %self.room.inner.room.room_id(), txn_id = %self.transaction_id))]
```
poljar, line 889: That's a bit of a long line, sadly rustfmt doesn't want to touch macros.

poljar (approved):

PR #3631, opened 2024-07-01
send queue: wake up the sending task after editing an event

bnjbvr:
It could be that the last event in a room's send queue has been marked as wedged. In that case, the task will sleep until it's notified again. If the event is being edited, then nothing would wake up the task; a manual wakeup might be required in that case.

The new integration test shows the issue; the last `assert_update` would fail with a timeout before this patch.

Part of #3361.


Hywan (approved):

bnjbvr:
Thanks for the quick review! (The force-push was for edits to the comments in the test code.)

PR #3632, opened 2024-07-01
send queue: make `SendHandle::abort()`/`update()` more precise

bnjbvr:
Using `SendHandle::abort()` after the event has been sent would look like a successful abort of the event, while it's not the case; this fixes this by having the state store backends return whether they've touched an entry in the database.

Found thanks to `multiverse`.

Part of #3361.

### DIFF
--- a/crates/matrix-sdk/src/send_queue.rs
+++ b/crates/matrix-sdk/src/send_queue.rs
@@ -674,13 +674,19 @@ impl QueueStorage {
     ) -> Result<(), RoomSendQueueStorageError> {
         self.mark_as_not_being_sent(transaction_id).await;
 
-        Ok(self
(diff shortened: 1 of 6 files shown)


crates/matrix-sdk/src/send_queue.rs
```diff
@@ -693,20 +708,22 @@ impl QueueStorage {
         &self,
         transaction_id: &TransactionId,
     ) -> Result<bool, RoomSendQueueStorageError> {
-        // Note: since there's a single caller (the room sending task, which processes
-        // events to send linearly), there's no risk for race conditions here.
```
Hywan, line 697: Trolololol

Hywan (approved):

PR #3634, opened 2024-07-01
sdk: don't clobber the DM list if we failed to deserialize the previous `m.direct` event

### LINKED ISSUE #3410: Element X can overwrite user's DM rooms
We've had multiple bug reports on element-web from different users since ~Sept 2023 complaining that their DM rooms are being wiped. The meta issue for this is https://github.com/element-hq/element-meta/issues/2035

From one of the more recent rageshakes, I have dug into the server-side logs to see which user agent did the PUT to `/account_data/m.direct` and it implicated `"Element X/0.4.9 (Google Pixel 7; Android 14; AP1A.240405.002; Sdk 742700b7f)" [0 dbevts]`.

In order to set DM rooms correctly, the API is horrible. The SDK MUST ensure that it first does a GET to fetch the latest `m.direct` event, then adds to the JSON map, then does a PUT of the whole event again. If it can't do the GET, it MUST NOT do the PUT or else it will overwrite the DM rooms on the account.

bnjbvr:
Fixes #3410.

I don't think there's a point in waiting for the `m.direct` data to come through sync, as 1. it would require a causality link (channel) between getting the m.direct event and this method — not impossible but tedious, and 2. marking a room as DM doesn't occur that often, so there's no need to optimize for it that much.

### DIFF
--- a/crates/matrix-sdk/tests/integration/client.rs
+++ b/crates/matrix-sdk/tests/integration/client.rs
@@ -1,6 +1,6 @@
 use std::{collections::BTreeMap, time::Duration};
 
-use assert_matches2::assert_let;
+use assert_matches2::{assert_let, assert_matches};
 use eyeball_im::VectorDiff;
 use futures_util::FutureExt;
 use matrix_sdk::{
@@ -442,7 +442,7 @@ async fn test_marking_room_as_dm() {
     client
         .sync_once(SyncSettings::default())
         .await
-        .expect("We should be able to performa an initial sync");
+        .expect("We should be able to perform an initial sync");
 
     let account_data = client
         .account()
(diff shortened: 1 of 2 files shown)


crates/matrix-sdk/tests/integration/client.rs
```diff
@@ -504,6 +504,48 @@ async fn test_marking_room_as_dm() {
...
+
+    let bob = user_id!("@bob:example.com");
+    let users = vec![bob.to_owned()];
+
+    // The response must be valid JSON, but not a valid `DirectEventContent`
+    // representation.
+    Mock::given(method("GET"))
+        .and(path("_matrix/client/r0/user/@example:localhost/account_data/m.direct"))
+        .and(header("authorization", "Bearer 1234"))
+        .respond_with(ResponseTemplate::new(200).set_body_json(json!(["hey", null, true, 42])))
```
Hywan, line 536: That's my new catchy phrase: _hey null true 42_!

Hywan (approved):
It looks good to me. Kudos for the test! Thanks for the fix!

PR #3635, opened 2024-07-02
ffi: refactor `ClientBuilder`

bnjbvr:
No functional changes here, just inlining methods called by a single caller, and simplifying things that could be simplified.


poljar (approved):
Neat.

PR #3636, opened 2024-07-02
sdk-store-encryption: use `ZeroizeOnDrop` instead of `[zeroize(drop)]`

bnjbvr:
One item of https://github.com/matrix-org/matrix-rust-sdk/issues/3272.

### DIFF
--- a/crates/matrix-sdk-store-encryption/src/lib.rs
+++ b/crates/matrix-sdk-store-encryption/src/lib.rs
@@ -33,7 +33,7 @@ use pbkdf2::pbkdf2;
 use rand::{thread_rng, Error as RandomError, Fill};
 use serde::{de::DeserializeOwned, Deserialize, Serialize};
 use sha2::Sha256;
-use zeroize::Zeroize;
+use zeroize::{Zeroize, ZeroizeOnDrop};
 
 const VERSION: u8 = 1;
 const KDF_SALT_SIZE: usize = 32;
@@ -774,8 +774,7 @@ impl StoreCipher {
     }
 }
 
-#[derive(Zeroize)]
-#[zeroize(drop)]
+#[derive(ZeroizeOnDrop)]
 struct MacKey(Box<[u8; 32]>);
 
 impl MacKey {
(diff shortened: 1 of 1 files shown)


crates/matrix-sdk-store-encryption/src/lib.rs
```diff
@@ -774,8 +774,7 @@ impl StoreCipher {
     }
 }
 
-#[derive(Zeroize)]
-#[zeroize(drop)]
+#[derive(ZeroizeOnDrop)]
```
poljar, line 777: I'm not quite certain, but I think we want `#[derive(Zeroize, ZeroizeOnDrop)]`: https://docs.rs/zeroize/latest/zeroize/index.html#custom-derive-support.

Not exactly sure what only deriving `ZeroizeOnDrop` does, I would need to let Rust expand the macro: https://github.com/RustCrypto/utils/blob/3867a73955c0c35bd3bd066d086c07c5e43a78c8/zeroize_derive/src/lib.rs#L99-L120.

poljar (changes requested):

crates/matrix-sdk-store-encryption/src/lib.rs
bnjbvr: This is `Zeroize`:

```
// ========================================
// Recursive expansion of the Zeroize macro
// ========================================

impl ::zeroize::Zeroize for MacKey {
    fn zeroize(&mut self) {
        match self {
            #[allow(unused_variables)]
            MacKey(__zeroize_field_0) => __zeroize_field_0.zeroize(),
            _ => {}
        }
    }
}
```

vs `ZeroizeOnDrop`:

```
// ==============================================
// Recursive expansion of the ZeroizeOnDrop macro
// ==============================================

impl Drop for MacKey {
    fn drop(&mut self) {
        use ::zeroize::__internal::AssertZeroize;
        use ::zeroize::__internal::AssertZeroizeOnDrop;
        match self {
            #[allow(unused_variables)]
            MacKey(__zeroize_field_0) => __zeroize_field_0.zeroize_or_on_drop(),
            _ => {}
        }
    }
}
#[doc(hidden)]
impl ::zeroize::ZeroizeOnDrop for MacKey {}
```

so I think this is doing the right thing?

poljar: Alright, though I can't think of a reason why we wouldn't want `Zeroize` and the `zeroize()` method to be available.

poljar (approved):

PR #3637, opened 2024-07-02
ui: Expose identifier of `RepliedToInfo` and `EditInfo`

zecakeh:
That way, users of the crate don't need to keep track of the ID separately.


bnjbvr:
Thanks for the PR; why do you need this, though? This ID is only internal as far as I can tell.

zecakeh:
I realized, I also need `RepliedToInfo` to implement `Clone`, I'll add a commit if you don't mind.

bnjbvr (approved):
Thanks.

PR #3086, opened 2024-02-02
Deprecate methods in the store-encryption crate that use serde_json

poljar:
These methods should not be used anymore, people should pick their own
serialization format when writing a store.

Issue #718, opened 2022-05-27
Remove the default serializer in the matrix-sdk-store-encryption crate


poljar:
The default serialization format for encrypted values in the store-encryption crate is JSON using serde_json. Since the ciphertext and nonce are byte vectors and JSON a textual encoding this ends up using more storage than what would be needed.

People can already use any other serialization format that is supported by Serde so let's just remove the default one.

poljar:
This PR will deprecate the methods in question: https://github.com/matrix-org/matrix-rust-sdk/pull/3086.

richvdh:
It's worth noting that there are two serialization steps in encryption, and this issue isn't entirely clear whether we're talking about both, or just one (and if so, which).

First of all we have to serialize the input data from rust. `encrypt_value` , `encrypt_value_typed`, and `encrypt_value_base64_typed` serialize the data as JSON. `encrypt_value_base64_data` and `encrypt_value_data`  take the serialized data.

Then we do the encryption, which gives us an `EncryptedValue` struct.

Then, possibly, we, do a second round of serialization, on the `EncryptedValue`. Here, `encrypt_value` serializes the `EncryptedValue` as JSON. `encrypt_value_typed` and `encrypt_value_data`  return the raw `EncryptedValue` for serialization by the caller. `encrypt_value_base64_typed` and `encrypt_value_base64_data` base64-encode the values inside the `EncryptedValue` to give an `EncryptedValueBase64`. (The latter is not strictly a serialization, but it's convenient to consider together here.)

In summary:

| Method | Pre-encryption serialization | Post-encryption conversion |
|---|---|---|
| `encrypt_value` | JSON | JSON |
| `encrypt_value_typed` | JSON | None |
| `encrypt_value_base64_typed`  | JSON | `EncryptedValueBase64` |
| `encrypt_value_base64_data` | None | `EncryptedValueBase64` |
| `encrypt_value_data` | None | None |

So, I'm not sure if this issue is talking about just removing `encrypt_value` (and its counterpart `decrypt_value`), or also the `*_typed` methods?

poljar:
> So, I'm not sure if this issue is talking about just removing encrypt_value (and its counterpart decrypt_value), or also the *_typed methods?

Ideally, the `matrix-sdk-store-encryption` crate would not depend on the `serde_json` crate anymore. This would mean that only `encrypt_value_data()` would be left, though that could be a bit hard to pull off due to backwards compatibility needs.

[...]

PR #688, opened 2022-05-18
store-encryption: Use bincode for encoding

DarkKirb:
This commit changes the default ciphertext serialization format for matrix-sdk-store-encryption to bincode, as the current json serialization is incurring a large space overhead.

Currently, the following plaintext encrypts to ~237 bytes. With this change, it only takes 58 bytes (75% reduction in size):

```
{"test":"test"}
```

Existing ciphertexts are unaffected, and still supported

PR #3638, opened 2024-07-02
Stores: Remove `StoreCipher::{en,de}crypt_value_[base64_]typed`

richvdh:
`StoreCipher`'s interface has grown *baroque* over the years. These four functions aren't giving a whole lot of value: they are only used in a couple of places in the codebase, and are pretty simple.

This PR is a mission to clean that up.

Review commit-by-commit.

### DIFF
--- a/crates/matrix-sdk-store-encryption/src/lib.rs
+++ b/crates/matrix-sdk-store-encryption/src/lib.rs
@@ -418,45 +418,8 @@ impl StoreCipher {
     /// # anyhow::Ok(()) };
     /// ```
     pub fn encrypt_value(&self, value: &impl Serialize) -> Result<Vec<u8>, Error> {
-        Ok(serde_json::to_vec(&self.encrypt_value_typed(value)?)?)
-    }
-
-    /// Encrypt a value before it is inserted into the key/value store.
-    ///
-    /// A value can be decrypted using the
-    /// [`StoreCipher::decrypt_value_typed()`] method. This is the lower
-    /// level function to `encrypt_value`, but returns the
-    /// full `EncryptdValue`-type
-    ///
-    /// # Arguments
-    ///
-    /// * `value` - A value that should be encrypted, any value that implements
-    ///   `Serialize` can be given to this method. The value will be serialized
-    ///   as json before it is encrypted.
-    ///
-    ///
-    /// # Examples
-    ///
-    /// ```no_run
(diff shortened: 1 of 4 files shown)


crates/matrix-sdk-store-encryption/src/lib.rs
```diff
@@ -603,89 +525,12 @@ impl StoreCipher {
...
-    ///     "some": "data",
-    /// });
-    ///
-    /// let encrypted = store_cipher.encrypt_value_typed(&value)?;
-    /// let decrypted: Value = store_cipher.decrypt_value_typed(encrypted)?;
-    ///
-    /// assert_eq!(value, decrypted);
-    /// # anyhow::Ok(()) };
-    /// ```
-    pub fn decrypt_value_typed<T: DeserializeOwned>(
```
poljar, line 640: Removing this removes the abililty to use the `StoreCipher` with more efficient serialization formats, which at least one user used in the past: https://github.com/matrix-org/matrix-rust-sdk/pull/688.

Also, we wanted to remove the other, JSON methods: https://github.com/matrix-org/matrix-rust-sdk/issues/718.

While I understand the baroqueness argument and the poor naming of the methods I don't think I'd want to remove this functionality.

poljar (changes requested):

crates/matrix-sdk-store-encryption/src/lib.rs
richvdh: > Removing this removes the ability to use the `StoreCipher` with more efficient serialization formats, which at least one user used in the past: #688.

Well, you can use `encrypt_value_data` instead?

> Also, we wanted to remove the other, the JSON based, methods: #718.
> 
> While I understand the baroqueness argument and the poor naming of the methods I don't think I'd want to remove this functionality.

Well, I'd quite like to get rid of `encrypt_value` and `decrypt_value` as well; I just didn't get that far.

Per my note on #718, I'm slightly confused about what it's asking for. In particular I note that https://github.com/matrix-org/matrix-rust-sdk/pull/3086 deprecated `decrypt_value_typed`. Has your thinking changed here?

poljar: > Well, you can use encrypt_value_data instead?

Ah indeed, we still have that one. Then I don't have any objections.

poljar (approved):

PR #3639, opened 2024-07-02
feature(crypto): Add support for master key local pinning

BillCarsonFr:
part of invisible crypto, follow up of https://github.com/matrix-org/matrix-rust-sdk/pull/3607
Alternative fix for  https://github.com/matrix-org/matrix-rust-sdk/issues/3564
Alternative PR https://github.com/matrix-org/matrix-rust-sdk/pull/3610

Add the capability to locally pin a public MSK for a `ReadOnlyUserIdentity`. 
The first time an identity is seen for a user, the msk is pinned. Pin violation will be reported when an identity is rotated.

This PR only adds support for pinning, support for serialization/migration, persistance. 
This could be used later by other PRs to report specific errors or show pinning violation to users.

Note about verification and pinning.
As part of this PR, if a new identity is detected it will still be seen as a pinning violation even if the new identity is signed by our usk. But the `UserIdentity::has_identity_mismatch()` will be ok. There is a pinning violation, but the new identity is verified and verification has priority. 
That is to say that there is no auto-pinning if verified so far. To be discussed if we want it later

Signed-off-by:

### DIFF
--- a/crates/matrix-sdk-crypto/src/identities/manager.rs
+++ b/crates/matrix-sdk-crypto/src/identities/manager.rs
@@ -1969,7 +1969,7 @@ pub(crate) mod tests {
         assert!(!first_device.is_cross_signed_by_owner(&identity));
 
         let remember_previous_identity = other_identity.clone();
-        // We receive a new keys update for that user, with no identity anymore
+        // We receive updated keys for that user, with no identity anymore.
         // Notice that there is no server API to delete identity, but we want to test
         // here that a home server cannot clear the identity and serve a new one
         // after that would get automatically approved.
--- a/crates/matrix-sdk-crypto/src/identities/manager.rs
+++ b/crates/matrix-sdk-crypto/src/identities/manager.rs
@@ -1922,8 +1922,8 @@ pub(crate) mod tests {
         let identity = manager.store.get_user_identity(other_user).await.unwrap().unwrap();
         let other_identity = identity.other().unwrap();
 
-        // There should be now an identity and no pin violation (pinned msk is the
-        // current one)
+        // We should now have an identity for the user but no pin violation
+        // (pinned master key is the current one)
         assert!(!other_identity.has_pin_violation());
         let first_device = manager
             .store
@@ -1970,9 +1970,9 @@ pub(crate) mod tests {
 
         let remember_previous_identity = other_identity.clone();
         // We receive updated keys for that user, with no identity anymore.
-        // Notice that there is no server API to delete identity, but we want to test
-        // here that a home server cannot clear the identity and serve a new one
-        // after that would get automatically approved.
+        // Notice that there is no server API to delete identity, but we want to
+        // test here that a home server cannot clear the identity and
+        // subsequently serve a new one which would get automatically approved.
         manager
             .receive_keys_query_response(
                 &TransactionId::new(),
@@ -1989,7 +1989,7 @@ pub(crate) mod tests {
     }
 
     #[async_test]
-    async fn test_manager_resolve_identity_mismatch() {
+    async fn test_manager_resolve_identity_pin_violation() {
         use test_json::keys_query_sets::IdentityChangeDataSet as DataSet;
 
         let manager = manager_test_helper(user_id(), device_id()).await;
@@ -2018,7 +2018,7 @@ pub(crate) mod tests {
         // We have a new identity now, so there should be a pin violation
         assert!(other_identity.has_pin_violation());
 
-        // Resolve the misatch by pinning the new identity
+        // Resolve the violation by pinning the new identity
         other_identity.pin();
 
         assert!(!other_identity.has_pin_violation());
--- a/crates/matrix-sdk-crypto/src/identities/manager.rs
+++ b/crates/matrix-sdk-crypto/src/identities/manager.rs
@@ -512,6 +512,7 @@ impl IdentityManager {
             *changed_private_identity = self.check_private_identity(&identity).await;
             Ok(identity.into())
         } else {
+            // First time seen, create the identity. The current MSK will be pinned.
             let identity = OtherUserIdentityData::new(master_key, self_signing)?;
             Ok(identity.into())
         }
@@ -1338,7 +1339,7 @@ pub(crate) mod tests {
     use std::ops::Deref;
 
     use futures_util::pin_mut;
-    use matrix_sdk_test::{async_test, response_from_file};
+    use matrix_sdk_test::{async_test, response_from_file, test_json};
     use ruma::{
         api::{client::keys::get_keys::v3::Response as KeysQueryResponse, IncomingResponse},
         device_id, user_id, TransactionId,
@@ -1896,5 +1897,130 @@ pub(crate) mod tests {
         assert_eq!(devices.devices().count(), 1);
 
         manager.store.get_device_data(other_user, device_id!("OBEBOSKTBE")).await.unwrap().unwrap();
+    }
+
+    #[async_test]
+    async fn test_manager_identity_updates() {
+        use test_json::keys_query_sets::IdentityChangeDataSet as DataSet;
+
+        let manager = manager_test_helper(user_id(), device_id()).await;
+        let other_user = DataSet::user_id();
+        let devices = manager.store.get_user_devices(other_user).await.unwrap();
+        assert_eq!(devices.devices().count(), 0);
+
+        let identity = manager.store.get_user_identity(other_user).await.unwrap();
+        assert!(identity.is_none());
+
+        manager
+            .receive_keys_query_response(
+                &TransactionId::new(),
+                &DataSet::key_query_with_identity_a(),
+            )
+            .await
+            .unwrap();
+
+        let identity = manager.store.get_user_identity(other_user).await.unwrap().unwrap();
+        let other_identity = identity.other().unwrap();
+
+        // There should be now an identity and no pin violation (pinned msk is the
+        // current one)
+        assert!(!other_identity.has_pin_violation());
+        let first_device = manager
+            .store
+            .get_device_data(other_user, DataSet::first_device_id())
+            .await
+            .unwrap()
+            .unwrap();
+        assert!(first_device.is_cross_signed_by_owner(&identity));
+
+        // We receive a new keys update for that user, with a new identity
+        manager
+            .receive_keys_query_response(
+                &TransactionId::new(),
+                &DataSet::key_query_with_identity_b(),
+            )
+            .await
+            .unwrap();
+
+        let identity = manager.store.get_user_identity(other_user).await.unwrap().unwrap();
+        let other_identity = identity.other().unwrap();
+
+        // The previous known identity has been replaced, there should be a pin
+        // violation
+        assert!(other_identity.has_pin_violation());
+
+        let second_device = manager
+            .store
+            .get_device_data(other_user, DataSet::second_device_id())
+            .await
+            .unwrap()
+            .unwrap();
+
+        // There is a new device signed by the new identity
+        assert!(second_device.is_cross_signed_by_owner(&identity));
+
+        // The first device should not be signed by the new identity
+        let first_device = manager
+            .store
+            .get_device_data(other_user, DataSet::first_device_id())
+            .await
+            .unwrap()
+            .unwrap();
+        assert!(!first_device.is_cross_signed_by_owner(&identity));
+
+        let remember_previous_identity = other_identity.clone();
+        // We receive a new keys update for that user, with no identity anymore
+        // Notice that there is no server API to delete identity, but we want to test
+        // here that a home server cannot clear the identity and serve a new one
+        // after that would get automatically approved.
+        manager
+            .receive_keys_query_response(
+                &TransactionId::new(),
+                &DataSet::key_query_with_identity_no_identity(),
+            )
+            .await
+            .unwrap();
+
+        let identity = manager.store.get_user_identity(other_user).await.unwrap().unwrap();
+        let other_identity = identity.other().unwrap();
+
+        assert_eq!(other_identity, &remember_previous_identity);
+        assert!(other_identity.has_pin_violation());
+    }
+
+    #[async_test]
+    async fn test_manager_resolve_identity_mismatch() {
+        use test_json::keys_query_sets::IdentityChangeDataSet as DataSet;
+
+        let manager = manager_test_helper(user_id(), device_id()).await;
+        let other_user = DataSet::user_id();
+
+        manager
+            .receive_keys_query_response(
+                &TransactionId::new(),
+                &DataSet::key_query_with_identity_a(),
+            )
+            .await
+            .unwrap();
+
+        // We receive a new keys update for that user, with a new identity
+        manager
+            .receive_keys_query_response(
+                &TransactionId::new(),
+                &DataSet::key_query_with_identity_b(),
+            )
+            .await
+            .unwrap();
+
+        let identity = manager.store.get_user_identity(other_user).await.unwrap().unwrap();
+        let other_identity = identity.other().unwrap();
+
+        // We have a new identity now, so there should be a pin violation
+        assert!(other_identity.has_pin_violation());
+
+        // Resolve the misatch by pinning the new identity
+        other_identity.pin();
+
+        assert!(!other_identity.has_pin_violation());
     }
 }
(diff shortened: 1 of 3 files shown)


crates/matrix-sdk-crypto/src/identities/user.rs
```diff
@@ -371,10 +402,87 @@ impl ReadOnlyUserIdentities {
 /// only contain a master key and a self signing key, meaning that only device
 /// signatures can be checked with this identity.
 #[derive(Debug, Clone, Deserialize, Serialize)]
+#[serde(from = "ReadOnlyUserIdentitySerializer", into = "ReadOnlyUserIdentitySerializer")]
```
richvdh, line 405: It might be cleaner to write custom `Serialize` and `Deserialize` implementations, but I'm not sure.

crates/matrix-sdk-crypto/src/identities/user.rs
```diff
@@ -371,10 +402,87 @@ impl ReadOnlyUserIdentities {
...
     user_id: OwnedUserId,
     pub(crate) master_key: Arc<MasterPubkey>,
     self_signing_key: Arc<SelfSigningPubkey>,
+    /// The first time a cryptographic identity is seen for a given user, it
+    /// will be associated to that user (i.e pinned). Future interaction
+    /// will expect this user crypto identity to stay the same,
+    /// this will help prevent some MITM attacks.
+    /// In case of identity change, it will be possible to pin the new identity
+    /// is the user wants.
+    pinned_msk: Arc<RwLock<MasterPubkey>>,
```
richvdh, line 416: Worth noting the presence of this field means that the documentation on this class ("other users will
only contain a master key and a self signing key"), as well as its name, are no longer correct. We should at least fix the doc; fixing the name would be good too but might be a bit invasive (plus, we'd need to think of good names).

richvdh (commented):
This looks broadly sensible to me, though I'd defer to @poljar on this.

There are a bunch of documentation nits I'd like to see cleaned up, but let's agree on the basic structure first

crates/matrix-sdk-crypto/src/identities/user.rs
BillCarsonFr: I tried and for it to work properly I had to put a new field `version: Option<u8>` in the `ReadOnlyUserIdentity` model. 
I tried then with the helper pattern seen in other part of the code and it looks a bit better I think as it delegates serialization to a specific object and just needs standard `From` impl.

crates/matrix-sdk-crypto/src/identities/user.rs
```diff
@@ -447,6 +560,24 @@ impl ReadOnlyUserIdentity {
...
+        *m = self.master_key.as_ref().clone()
+    }
+
+    /// Key pinning acts as a trust on first use mechanism, the first time an
+    /// identity is known for a user it will be pinned.
+    /// For future interaction with a user, the identity is expected to be the
+    /// one that was pinned. In case of identity change the UI client should
+    /// receive reports of pinning violation and decide to act accordingly;
+    /// that is accept and pin the new identity, perform a verification or
+    /// stop communications.
```
poljar, line 575: The first line should be a short explanation of the method kept under some 80-100 characters, then we can expand on the first line in the next paragraph.

Could you try to reword this to conform this style?

crates/matrix-sdk-crypto/src/identities/user.rs
```diff
@@ -291,6 +292,36 @@ impl UserIdentity {
...
+
+    /// An identity mismatch is detected when there is a trust problem with the
+    /// user identity. There is an identity mismatch if the current identity
+    /// is not verified and there is a pinning violation. An identity
+    /// mismatch must be reported to the user, and can be resolved by:
+    /// - Verifying the new identity (see
+    ///   [`UserIdentity::request_verification`])
+    /// - Or by updating the pinned key
+    ///   ([`UserIdentity::pin_current_master_key`]).
+    pub fn has_identity_mismatch(&self) -> bool {
```
poljar, line 316: Similar to the `has_pin_violation()` method, the first line should be something akin to "Has the identity changed since we last pinned it?"

I think this could also could be perhaps better named, identity mismatch makes me think of the identity not being what we expect. I guess you could argue that this is exactly the case here, but perhaps we could name it to be something along the lines that the identity has changed without our approval instead. Feel free to ignore this part if you disagree.

crates/matrix-sdk-crypto/src/identities/user.rs
```diff
@@ -1015,4 +1212,74 @@ pub(crate) mod tests {
             [second_device_id]
         );
     }
+
+    #[async_test]
+    async fn resolve_identity_mismacth_with_verification() {
```
poljar, line 1217:
```suggestion
    async fn resolve_identity_mismatch_with_verification() {
```

testing/matrix-sdk-test/src/test_json/keys_query_sets.rs
```diff
@@ -457,3 +457,215 @@ impl KeyDistributionTestData {
...
+            "failures": {},
+            "master_keys": Self::msk_b(),
+            "self_signing_keys": Self::ssk_b(),
+        }));
+        KeyQueryResponse::try_from_http_response(data)
+            .expect("Can't parse the `/keys/upload` response")
+    }
+
+    /// A key query with a new identity (Ib) and a new device `ATWKQFSFRN`.
+    /// `ATWKQFSFRN` is signed with the new identity but
```
poljar, line 637: But? But what?!? This cliff-hanger produces anxiety.

testing/matrix-sdk-test/src/test_json/keys_query_sets.rs
```diff
@@ -457,3 +457,215 @@ impl KeyDistributionTestData {
...
+                "@bob:localhost": {
+                    "ed25519:ATWKQFSFRN": "BQ9Gp0p+6srF+c8OyruqKKd9R4yaub3THYAyyBB/7X/rG8BwcAqFynzl1aGyFYun4Q+087a5OSiglCXI+/kQAA",
+                    "ed25519:At1ai1VUZrCncCI7V7fEAJmBShfpqZ30xRzqcEjTjdc": "TWmDPaG7t0rZ6luauonELD3dmBDTIRryqXhgsIQRiGint2rJdic8RVyZ6a61bgu6mtBjfvU3prqMNp6sVi16Cg"
+                }
+            },
+            "user_id": "@bob:localhost",
+        })
+    }
+    /// A key query with a new identity (Ib) and a new device `ATWKQFSFRN`.
+    /// `ATWKQFSFRN` is signed with the new identity but
```
poljar, line 619: More butts.

crates/matrix-sdk-crypto/src/identities/user.rs
```diff
@@ -465,7 +596,15 @@ impl ReadOnlyUserIdentity {
     ) -> Result<bool, SignatureError> {
         master_key.verify_subkey(&self_signing_key)?;
 
-        let new = Self::new(master_key, self_signing_key)?;
+        // The pin is maintained.
```
poljar, line 599: Perhaps we can expand this a bit, to explain the why. I know the why right now but another reader might not know it.

crates/matrix-sdk-crypto/src/identities/manager.rs
```diff
@@ -1893,4 +1894,129 @@ pub(crate) mod tests {
...
+        let first_device = manager
+            .store
+            .get_readonly_device(other_user, DataSet::first_device_id())
+            .await
+            .unwrap()
+            .unwrap();
+        assert!(!first_device.is_cross_signed_by_owner(&identity));
+
+        let remember_previous_identity = other_identity.clone();
+        // We receive a new keys update for that user, with no identity anymore
```
poljar, line 1968:
```suggestion
        // We receive updated keys for that user, with no identity anymore
```

poljar (changes requested):
I think the approach looks good, I left a bunch of nits though. I'll let @dkasak double check things.

crates/matrix-sdk-crypto/src/identities/user.rs
BillCarsonFr: I am not sure what would be the best name for that. I took identity mismatch to use the same term that is used in this [doc](https://github.com/element-hq/element-meta/blob/develop/docs/crypto/trust_v2.md#behavior-for-identity-mismatch).

Maybe @dkasak have an idea? Maybe `is_identity_trustworthy`? or more based on the use of this method `identity_needs_user_approval` / `validation` / `review` ?

crates/matrix-sdk-crypto/src/identities/user.rs
```diff
@@ -471,7 +612,18 @@ impl OtherUserIdentityData {
     ) -> Result<bool, SignatureError> {
         master_key.verify_subkey(&self_signing_key)?;
 
-        let new = Self::new(master_key, self_signing_key)?;
+        // We update the identity with the new master and self signing key, but we keep
+        // the previous pinned master key.
+        // This identity will have a pin violation until the new master key is pinned
+        // (see [`has_pin_violation`]).
```
poljar, line 618: This won't produce a link since it isn't a doc comment, at least it doesn't for me, for you it does?
```suggestion
        // (see `has_pin_violation()`).
```

poljar (approved):
Left a tiny nit, but I think that this is fine.

crates/matrix-sdk-crypto/src/identities/user.rs
dkasak: I would avoid the `mismatch` and `trustworthy` wording. `identity_needs_user_approval` sounds good to me.

crates/matrix-sdk-crypto/src/identities/manager.rs
```diff
@@ -1897,4 +1898,129 @@ pub(crate) mod tests {
...
+                &DataSet::key_query_with_identity_a(),
+            )
+            .await
+            .unwrap();
+
+        let identity = manager.store.get_user_identity(other_user).await.unwrap().unwrap();
+        let other_identity = identity.other().unwrap();
+
+        // There should be now an identity and no pin violation (pinned msk is the
+        // current one)
```
dkasak, line 1926:
```suggestion
        // We should now have an identity for the user but no pin violation
        // (pinned master key is the current one)
```

crates/matrix-sdk-crypto/src/identities/manager.rs
```diff
@@ -1897,4 +1898,129 @@ pub(crate) mod tests {
...
+            .await
+            .unwrap()
+            .unwrap();
+        assert!(!first_device.is_cross_signed_by_owner(&identity));
+
+        let remember_previous_identity = other_identity.clone();
+        // We receive updated keys for that user, with no identity anymore.
+        // Notice that there is no server API to delete identity, but we want to test
+        // here that a home server cannot clear the identity and serve a new one
+        // after that would get automatically approved.
```
dkasak, line 1975:
```suggestion
        // We receive updated keys for that user, with no identity anymore.
        // Notice that there is no server API to delete identity, but we want to
        // test here that a home server cannot clear the identity and
        // subsequently serve a new one which would get automatically approved.
```

crates/matrix-sdk-crypto/src/identities/user.rs
```diff
@@ -295,6 +296,39 @@ impl UserIdentity {
             methods,
         )
     }
+
+    /// Pin the current identity (Master public key).
```
dkasak, line 300:
```suggestion
    /// Pin the current identity (public part of the master signing key).
```

crates/matrix-sdk-crypto/src/identities/user.rs
```diff
@@ -295,6 +296,39 @@ impl UserIdentity {
...
+        let changes = Changes {
+            identities: IdentityChanges { changed: vec![to_save], ..Default::default() },
+            ..Default::default()
+        };
+        self.verification_machine.store.inner().save_changes(changes).await?;
+        Ok(())
+    }
+
+    /// Returns true if the identity is not verified and did change since the
+    /// last time we pinned it.
```
dkasak, line 313:
```suggestion
    /// Did the identity change after an initial observation in a way that
    /// requires approval from the user?
```

crates/matrix-sdk-crypto/src/identities/user.rs
```diff
@@ -295,6 +296,39 @@ impl UserIdentity {
...
+    ///
+    /// An identity mismatch is detected when there is a trust problem with the
+    /// user identity. There is an identity mismatch if the current identity
+    /// is not verified and there is a pinning violation. An identity
+    /// mismatch must be reported to the user, and can be resolved by:
```
dkasak, line 318:
```suggestion
    /// A user identity needs approval if it changed after the crypto machine
    /// has already observed ("pinned") a different identity for that user *and*
    /// it is not an explicitly verified identity (using for example interactive
    /// verification).
    ///
    /// Such a change is to be considered a pinning violation which the
    /// application should report to the local user, and can be resolved by:
    ///
```

crates/matrix-sdk-crypto/src/identities/user.rs
```diff
@@ -295,6 +296,39 @@ impl UserIdentity {
...
+    /// - Verifying the new identity (see
+    ///   [`UserIdentity::request_verification`])
+    /// - Or by updating the pinned key (see
```
dkasak, line 321:
```suggestion
    /// - Or by updating the pin to the new identity
```

crates/matrix-sdk-crypto/src/identities/user.rs
```diff
@@ -376,11 +410,93 @@ impl UserIdentityData {
...
 /// only contain a master key and a self signing key, meaning that only device
 /// signatures can be checked with this identity.
+///
+/// This struct also contains the currently pinned master key for that user.
+/// The first time a cryptographic identity is seen for a given user, it
+/// will be associated to that user (i.e. pinned). Future interactions
+/// will expect this user crypto identity to stay the same,
+/// this will help prevent some MITM attacks.
+/// In case of identity change, it will be possible to pin the new identity
+/// is the user wants.
```
dkasak, line 420:
```suggestion
/// This struct also contains the currently pinned user identity (master signing
/// key) for that user.
///
/// The first time a cryptographic user identity is seen for a given user, it
/// will be associated with that user ("pinned"). Future interactions
/// will expect this identity to stay the same, to avoid MITM attacks from the
/// homeserver.
///
/// The user can explicitly pin the new identity to allow for legitimate
/// identity changes (for example, in case of key material or device loss).
```

crates/matrix-sdk-crypto/src/identities/user.rs
```diff
@@ -1027,86 +1232,73 @@ pub(crate) mod tests {
...
-            bob.get_identity(alice.user_id(), None).await.unwrap().unwrap().other().unwrap();
+        println!("USK ID: {}", usk_key_id);
+        let keys_query = DataSet::key_query_with_identity_a();
+        let txn_id = TransactionId::new();
+        machine.mark_request_as_sent(&txn_id, &keys_query).await.unwrap();
 
-        assert!(bobs_own_identity.is_verified(), "Bob's identity should be verified.");
-        assert!(bobs_alice_identity.is_verified(), "Alice's identity should be verified as well.");
-    }
+        // Simulate an identity hange
```
dkasak, line 1251:
```suggestion
        // Simulate an identity change
```

crates/matrix-sdk-crypto/src/identities/manager.rs
```diff
@@ -1897,4 +1898,129 @@ pub(crate) mod tests {
...
+            .await
+            .unwrap();
+
+        let identity = manager.store.get_user_identity(other_user).await.unwrap().unwrap();
+        let other_identity = identity.other().unwrap();
+
+        // We have a new identity now, so there should be a pin violation
+        assert!(other_identity.has_pin_violation());
+
+        // Resolve the misatch by pinning the new identity
```
dkasak, line 2021:
```suggestion
        // Resolve the mismatch by pinning the new identity
```

crates/matrix-sdk-crypto/src/identities/user.rs
```diff
@@ -376,11 +410,93 @@ impl UserIdentityData {
...
 pub struct OtherUserIdentityData {
     user_id: OwnedUserId,
     pub(crate) master_key: Arc<MasterPubkey>,
     self_signing_key: Arc<SelfSigningPubkey>,
+    pinned_master_key: Arc<RwLock<MasterPubkey>>,
+}
+
+/// Intermediate struct to help serialize OtherUserIdentityData and support
+/// versioning and migration.
+/// Version v1 is adding support for identity pinning (`pinned_master_key`), as
```
dkasak, line 432:
```suggestion
/// versioning and migration.
///
/// Version v1 is adding support for identity pinning (`pinned_master_key`), as
```

dkasak (approved):
I've left a bunch of wording suggestions and would advise changing the `has_identity_mismatch` method name to `identity_needs_user_approval` as discussed, but otherwise this seems good to me, so leaving an `Approve` for when suggestions are handled.

PR #3640, opened 2024-07-02
timeline: when trying to edit/abort on a stale local echo, retry on matching remote echo

bnjbvr:
It's quite easy for consumers of the UI crate to get a stale `EditInfo` or `EventTimelineItem`, when trying to abort/edit an event:

1. embedder gets an `EditInfo` or `EventTimelineItem` from the local echo
2. timeline updates the item to a remote echo
3. editing or aborting using the `EditInfo` / `EventTimelineItem` obtained in step 1 would fail, before this patch.

So this is some kind of [time-of-check/time-of-use](https://en.wikipedia.org/wiki/Time-of-check_to_time-of-use) bug, with no security implications.

After this patch, we use better strategies, both for edits and redactions:

- for edits, we try to find an updated local item in the timeline `items` that would now have an event id (== it's been sent), or a matching remote echo (that does have an event id). This covers all the transitions (local echo => local echo with an event id => remote echo with an event id).
- for redactions, we use the more precise `TimelineItem.identifier()` method, which will get an event id if the local echo has been sent. There's no TOCTOU issue, because the `redact()` method takes the event timeline item directly, and not an `EditInfo`.

Part of #3361.


crates/matrix-sdk-ui/src/timeline/mod.rs
```diff
@@ -503,19 +518,38 @@ impl Timeline {
...
 
-                warn!("No handle for a local echo; should only happen in testing situations");
-                return Ok(false);
+                // We end up here in two cases: either there wasn't a local echo with this
+                // transaction id, or the send queue refused to edit the local echo (likely
+                // because it's sent).
+                //
+                // Try to find a matching local echo that now has an event id (it's been sent),
+                // or a remote echo with a matching transaction id, so as to
+                // send an actual redaction.
```
poljar, line 542: A redaction? We aren't redacting an event here, did you mean edit or replacement event?

poljar (approved):
As far as I understood this PR, I think it looks good. Left a small doc nit.

crates/matrix-sdk-ui/src/timeline/mod.rs
bnjbvr: Indeedy, thanks! (wow you actually read the code comments :exploding_head:)

poljar: There have been plenty times where I did not and have missed things, so I try to read them nowadays.

PR #3642, opened 2024-07-02
base: Add trace logging to memory store

kegsay:
To help debug https://github.com/matrix-org/complement-crypto/issues/77

### DIFF
--- a/crates/matrix-sdk-base/src/store/memory_store.rs
+++ b/crates/matrix-sdk-base/src/store/memory_store.rs
@@ -34,7 +34,7 @@ use ruma::{
     CanonicalJsonObject, EventId, MxcUri, OwnedEventId, OwnedMxcUri, OwnedRoomId,
     OwnedTransactionId, OwnedUserId, RoomId, RoomVersionId, TransactionId, UserId,
 };
-use tracing::{debug, warn};
+use tracing::{debug, trace, warn};
 
 use super::{
     traits::{ComposerDraft, QueuedEvent, SerializableEventContent},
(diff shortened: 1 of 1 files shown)


crates/matrix-sdk-base/src/store/memory_store.rs
```diff
@@ -284,11 +284,15 @@ impl StateStore for MemoryStore {
 
     async fn save_changes(&self, changes: &StateChanges) -> Result<()> {
```
bnjbvr, line 285: Instead of the manual `save changes:` prefix in every debug/trace statement, you can add an instrument derive here that will have the same effect:

```suggestion
    #[instrument(skip(self, changes))]
    async fn save_changes(&self, changes: &StateChanges) -> Result<()> {
```

If you remove "self" or "changes" from the `skip` list, they will be logged on every debug/trace statement too.

bnjbvr (changes requested):
Does it need to be checked-in? I managed to use a local fork of the Rust SDK repository while running Complement Crypto, so I imagined it's possible for others to do that too?

kegsay:
It doesn't happen frequently, so if it's checked in it will be quicker to identify which lock is deadlocking. I have a matching PR at https://github.com/matrix-org/complement-crypto/pull/99 which I'm hitting retry job on endlessly.

bnjbvr:
Alright then, please address the above review comment and then I'm happy to approve this :+1:

crates/matrix-sdk-base/src/store/memory_store.rs
kegsay: Done, I think.

bnjbvr (approved):

PR #3644, opened 2024-07-03
Crypto store: clear db before integ tests

richvdh:
It's currently possible for integ test results to leak from one test run to the next (for example, the indexeddb stores hang around in the browser), causing bad test results.

Extend the test setup routine to clear out the store before the test starts.


poljar (approved):

richvdh:
@poljar thanks for the quick review! I had to make a couple of other changes, but I don't think they are material, so will go ahead and land this.

PR #3645, opened 2024-07-03
Indexeddb: more efficient serialization format

richvdh:
The current serialization format that we use for most objects in the indexeddb crypto store is extremely space-inefficient, leading to poor performance and errors (cf https://github.com/element-hq/element-web/issues/27658).

The idea here is to switch to the more efficient format used by `maybe_encrypt_value`. To avoid a huge migration step, we make the deserialization routine handle either format.


andybalaam (approved):
Looks good. Hoping for big performance improvements.

poljar (approved):

PR #3646, opened 2024-07-03
feat(ui): Trigger room list update only when necessary

Hywan:
This patch revisits the need to trigger a room list update for all changes of `RoomInfo`. For the moment, it reduces the scope to `recency_timestamp` update.

This patch comes with a test to ensure things work as expected.

### DIFF
--- a/crates/matrix-sdk-ui/src/room_list_service/room.rs
+++ b/crates/matrix-sdk-ui/src/room_list_service/room.rs
@@ -14,6 +14,7 @@
 
 //! The `Room` type.
 
+use core::fmt;
 use std::{ops::Deref, sync::Arc};
 
 use async_once_cell::OnceCell as AsyncOnceCell;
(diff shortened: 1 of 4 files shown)


crates/matrix-sdk-ui/src/room_list_service/room.rs
```diff
@@ -29,12 +30,17 @@ use crate::{
...
-#[derive(Clone, Debug)]
+#[derive(Clone)]
 pub struct Room {
     inner: Arc<RoomInner>,
 }
 
-#[derive(Debug)]
+impl fmt::Debug for Room {
+    fn fmt(&self, formatter: &mut fmt::Formatter<'_>) -> fmt::Result {
+        formatter.debug_tuple("Room").field(&self.id().to_owned()).finish()
```
bnjbvr, line 40: nit: do we need `&{something}.to_owned()`?

```suggestion
        formatter.debug_tuple("Room").field(self.id()).finish()
```

testing/matrix-sdk-integration-testing/src/tests/sliding_sync/room.rs
```diff
@@ -982,7 +974,6 @@ async fn test_roominfo_update_deduplication() -> Result<()> {
             }
         );
     }
-    assert_pending!(stream);
```
bnjbvr, line 985: Can you keep this, or add an assertion on the expected next value, then `assert_pending` again?

bnjbvr (approved):
Cool

crates/matrix-sdk-ui/src/room_list_service/room.rs
Hywan: Yeah… otherwise `&RoomId` doesn't have a size known at compile-time.

testing/matrix-sdk-integration-testing/src/tests/sliding_sync/room.rs
Hywan: Sure.

PR #3648, opened 2024-07-03
timeline: use the previous content's membership info when it's missing from the current membership event

bnjbvr:
Synapse returns a bare `{ "membership": "leave" }` as the content of a room membership event (for leave membership changes and likely others). In this case, it'd still be nice to have some kind of display name/avatar URL to show in UIs; it's possible to reuse information from the previous member event, if available.

### DIFF
--- a/crates/matrix-sdk-ui/src/timeline/tests/basic.rs
+++ b/crates/matrix-sdk-ui/src/timeline/tests/basic.rs
@@ -31,7 +31,7 @@ use ruma::{
         },
         FullStateEventContent,
     },
-    owned_event_id, MilliSecondsSinceUnixEpoch,
+    owned_event_id, owned_mxc_uri, MilliSecondsSinceUnixEpoch,
 };
 use stream_assert::assert_next_matches;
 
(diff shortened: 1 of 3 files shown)


crates/matrix-sdk-ui/src/timeline/tests/basic.rs
```diff
@@ -201,6 +201,48 @@ async fn test_room_member() {
     assert_matches!(profile.displayname_change(), Some(_));
     assert_matches!(profile.avatar_url_change(), None);
 
+    let mut fourth_room_member_content = RoomMemberEventContent::new(MembershipState::Join);
+    fourth_room_member_content.displayname = Some("Alice In Wonderland".to_owned());
```
Hywan, line 205: 🐰

Hywan (approved):
It looks good to me. Thanks for the clean up too.
I like the fact we have less logics in FFI.

jmartinesp:
This requires adding an exception client-side (at least for EX Android) but I confirm it works as expected. Thanks!

bnjbvr:
Thanks for testing! will land the first part, and see how we can improve the disambiguation mappings.

PR #3652, opened 2024-07-03
 feat(base): Revisit the `roominfo_update` API

Hywan:
This patch updates several parts of the code but it's basically just
renamings.

First off, the patch renames `RoomInfoUpdate` to
`RoomInfoNotableUpdate`. The functions, methods or variables
whose names start with `roominfo_update(.*)` are renamed
`room_info_notable_update$1`.

Second, this patch removes the
`RoomInfoNotableUpdate::trigger_room_list_update` field. It is replaced
by a `reasons: RoomInfoNotableUpdateReasons` 8-bit unsigned integer.
It addresses the following issues:

1. When a subscriber receives a `RoomInfoNotableUpdate`, they have no
   idea what has triggered this update.
2. In
   `matrix_sdk_base::sliding_sync::BaseClient::process_sliding_sync_e2ee`,
   we were triggering an update even if the latest event wasn't
   modified: it is a false-positive, it was a bug and a waste of
   resources. Now it's more refined, see the why below.

Third, this patch removes the second `trigger_room_list_update` argument
of `matrix_sdk_base::BaseClient::apply_changes`. This method now knows
where to find the reasons for the room info notable updates, see next
point.

Fourth, this patch adds a new
`matrix_sdk::StateChanges::room_info_notable_updates` field which is a
B-tree map between an `OwnedRoomId` and a
`RoomInfoNotableUpdateReasons`. The idea is that all places that receive
a `StateChanges` can also create a room info notable update with a
specific reason. This is a finer grained mechanism, and it avoids to
pass new arguments everywhere. It's cleaner.

Finally, it's easier than ever to add a new reason and to propagate it
to subscribers.

---

Another patch is being added straight afterward to avoid emitting a
`RoomInfoNotableUpdateReasons::RECENCY_TIMESTAMP` for rooms that are new.
Otherwise the entries in `matrix_sdk_ui::room_list_service::RoomList`
receive a `VectorDiff` because a new room is inserted, and then a
`VectorDiff` because the recency timestamp is updated. The second
`VectorDiff` is useless in this case.

---

I think we can improve the `RoomInfo` _updates_ in FFI by adding more reasons.
We need to ask the Element's teams, or Fractal, for example.

### DIFF
--- a/crates/matrix-sdk-base/src/sliding_sync.rs
+++ b/crates/matrix-sdk-base/src/sliding_sync.rs
@@ -16,6 +16,7 @@ use std::collections::BTreeMap;
 #[cfg(feature = "e2e-encryption")]
 use std::ops::Deref;
 
+#[cfg(feature = "e2e-encryption")]
 use matrix_sdk_common::deserialized_responses::SyncTimelineEvent;
 #[cfg(feature = "e2e-encryption")]
 use ruma::events::AnyToDeviceEvent;
--- a/crates/matrix-sdk-base/src/sliding_sync.rs
+++ b/crates/matrix-sdk-base/src/sliding_sync.rs
@@ -12,9 +12,9 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use std::collections::BTreeMap;
 #[cfg(feature = "e2e-encryption")]
 use std::ops::Deref;
-use std::{collections::BTreeMap, ops::Not};
 
 #[cfg(feature = "e2e-encryption")]
 use matrix_sdk_common::deserialized_responses::SyncTimelineEvent;
@@ -76,6 +76,8 @@ impl BaseClient {
         );
 
         let mut changes = StateChanges::default();
+        let mut room_info_notable_updates =
+            BTreeMap::<OwnedRoomId, RoomInfoNotableUpdateReasons>::new();
 
         // Process the to-device events and other related e2ee data. This returns a list
         // of all the to-device events that were passed in but encrypted ones
@@ -95,12 +97,13 @@ impl BaseClient {
                         .map(|to_device| to_device.next_batch.clone()),
                 },
                 &mut changes,
+                &mut room_info_notable_updates,
             )
             .await?;
 
         trace!("ready to submit changes to store");
         self.store.save_changes(&changes).await?;
-        self.apply_changes(&changes);
+        self.apply_changes(&changes, room_info_notable_updates);
         trace!("applied changes");
 
         Ok(to_device)
@@ -144,6 +147,8 @@ impl BaseClient {
         };
 
         let mut changes = StateChanges::default();
+        let mut room_info_notable_updates =
+            BTreeMap::<OwnedRoomId, RoomInfoNotableUpdateReasons>::new();
 
         let store = self.store.clone();
         let mut ambiguity_cache = AmbiguityCache::new(store.inner.clone());
@@ -165,6 +170,7 @@ impl BaseClient {
                     &mut rooms_account_data,
                     &store,
                     &mut changes,
+                    &mut room_info_notable_updates,
                     &mut notifications,
                     &mut ambiguity_cache,
                 )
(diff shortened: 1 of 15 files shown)


bnjbvr (changes requested):
Can you put the renaming into a separate commit please?

Hywan:
It's done Sir.

crates/matrix-sdk-base/src/store/mod.rs
```diff
@@ -313,8 +313,21 @@ pub struct StateChanges {
...
     pub room_account_data:
         BTreeMap<OwnedRoomId, BTreeMap<RoomAccountDataEventType, Raw<AnyRoomAccountDataEvent>>>,
-    /// A map of `RoomId` to `RoomInfo`.
+
+    /// A map of `OwnedRoomId` to `RoomInfo`.
     pub room_infos: BTreeMap<OwnedRoomId, RoomInfo>,
+
+    /// A map of `OwnedRoomId` to `RoomInfoNotableUpdateReason`.
+    ///
+    /// A room info notable update is an update that can be interested for other
```
bnjbvr, line 328: interesting, not interested?

crates/matrix-sdk-base/src/store/mod.rs
```diff
@@ -313,8 +313,21 @@ pub struct StateChanges {
...
+    /// parts of the code. This mechanism is used in coordination with
+    /// [`BaseClient::room_info_notable_update_receiver`][baseclient] where
+    /// `RoomInfo` can be observed and some of its updates can be spread to
+    /// listeners.
+    ///
+    /// [baseclient]: crate::BaseClient::room_info_notable_update_receiver
+    pub room_info_notable_updates: BTreeMap<OwnedRoomId, RoomInfoNotableUpdateReasons>,
```
bnjbvr, line 335: I think putting the update reasons as a field in `StateStore` is a regression in terms of API: it's unused in the deepest layers (that do the actual storage), it's only used temporarily as a side parameter for `apply_changes`. Can you remove it from this struct please, and keep on using a plain simple parameter instead, please?

crates/matrix-sdk-base/src/sliding_sync.rs
```diff
@@ -741,11 +744,16 @@ fn process_room_properties(
...
 
-        changes
-            .room_info_notable_updates
-            .entry(room_id.to_owned())
-            .or_default()
-            .insert(RoomInfoNotableUpdateReasons::RECENCY_TIMESTAMP);
+        // If it's not a new room, let's emit a `RECENCY_TIMESTAMP` update.
+        // For a new room, the room will appear as new, so we don't care about this
+        // update.
+        if is_a_new_room.not() {
```
bnjbvr, line 750: nit: i know you love `.not()`, but how about using a plain `!` here :)

crates/matrix-sdk-base/src/sliding_sync.rs
```diff
@@ -341,6 +341,8 @@ impl BaseClient {
         };
 
         // Find or create the room in the store
+        let is_a_new_room = store.room_exists(room_id).not();
```
bnjbvr, line 344: nit: here and in the parameters list, rename `is_new_room` maybe?

Also let's not abuse of the `not()` combinator and use a plain `!` please.

testing/matrix-sdk-integration-testing/src/tests/sliding_sync/room.rs
```diff
@@ -848,18 +848,6 @@ async fn test_delayed_decryption_latest_event() -> Result<()> {
         }
     );
 
-    // Sometimes Synapse sends the same message twice. Let's consume useless `Set`…
-    // if they arrived before 3s.
```
bnjbvr, line 852: mwahahahahahah

bnjbvr (changes requested):
looks good! I'd like to revert the update reason to a parameter, since it's only used like that, and it'd be clearer at the call sites :+1:

crates/matrix-sdk-base/src/store/mod.rs
Hywan: Done.

crates/matrix-sdk-base/src/sliding_sync.rs
Hywan: Done. I really dislike `!`. My fight isn't over 😉.

crates/matrix-sdk-base/src/sliding_sync.rs
Hywan: Done.

bnjbvr (approved):
Thank you good Sir!

PR #3655, opened 2024-07-04
ffi: remove `is_encrypted` field from `RoomInfo`

jmartinesp:
It turns out this will cause a network request if the encryption info hasn't been loaded before, which is the case for opening a client in offline mode. It will slow down displaying the room list or loading the room info in general.

Partially reverts https://github.com/matrix-org/matrix-rust-sdk/pull/3602.

Signed-off-by:


Hywan (approved):
LGTM; thanks!

Issue #3589, opened 2024-06-20
ffi: `Timeline::load_reply_details()` could try to fetch the reply details from the event if it's in the timeline locally


bnjbvr:
Right now, we'll always try to send a request to /event to get the room event, but we may have it already (either in the event cache, or in the timeline too). In this case, we might want to prefer fetching it locally first, then try the remote request.

PR #3656, opened 2024-07-04
ffi: Timeline::load_reply_details checks if the event exists locally

jmartinesp:
Before it would go fetch the event from the server using a network request.

Fixes https://github.com/matrix-org/matrix-rust-sdk/issues/3589

Signed-off-by:

### DIFF
--- a/bindings/matrix-sdk-ffi/src/timeline/mod.rs
+++ b/bindings/matrix-sdk-ffi/src/timeline/mod.rs
@@ -19,9 +19,12 @@ use as_variant::as_variant;
 use content::{InReplyToDetails, RepliedToEventDetails};
 use eyeball_im::VectorDiff;
 use futures_util::{pin_mut, StreamExt as _};
-use matrix_sdk::attachment::{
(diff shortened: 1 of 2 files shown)


bindings/matrix-sdk-ffi/src/timeline/mod.rs
```diff
@@ -636,23 +639,30 @@ impl Timeline {
...
-                        sender_profile: replied_to.sender_profile().into(),
-                    },
-                ))
-            }
+        let replied_to: Result<RepliedToEvent, Error> =
+            if let Some(event) = self.inner.item_by_event_id(&event_id).await {
+                Ok(RepliedToEvent::from_timeline_item(&event))
+            } else {
+                let ret = self.inner.room().event(&event_id).await;
+                match ret {
```
bnjbvr, line 647: nit: can we get rid of the `ret` temporary variable please?

bnjbvr (approved):
Looks good, thanks!

PR #3658, opened 2024-07-04
sdk: cache server versions and unstable features

### LINKED ISSUE #2372: Cache server versions
In order to improve launch times we're currently [relying](https://github.com/vector-im/element-x-ios/blob/develop/ElementX/Sources/Services/UserSession/UserSessionStore.swift#L118) on the `serverVersions` client builer method introduced in https://github.com/matrix-org/matrix-rust-sdk/pull/1376.

Needless to say that's not great and moreso android isn't even using it, presumably because of lack of discoverability.

We need to start handling this on the Rust side and make it transparent to the final client.

bnjbvr:
This series of commits makes it possible to cache the server versions and unstable features, with a cache duration of one day (can be tweaked easily!). Caching works this way:

- if we have already filled the in-memory cache, use that
- instead, try to reload the disk caches,
- if they have expired, do a network request to get the latest server version (and cache the results in both memory and disk)

Fixes #2372.

### DIFF
--- a/crates/matrix-sdk/src/client/mod.rs
+++ b/crates/matrix-sdk/src/client/mod.rs
@@ -1444,8 +1444,10 @@ impl Client {
             .send(SessionChange::UnknownToken { soft_logout: *soft_logout });
     }
 
-    async fn request_server_versions(&self) -> HttpResult<Box<[MatrixVersion]>> {
-        let server_versions: Box<[MatrixVersion]> = self
+    async fn request_server_capabilities(
+        &self,
+    ) -> HttpResult<(Box<[MatrixVersion]>, BTreeMap<String, bool>)> {
+        let resp = self
             .inner
             .http_client
             .send(
@@ -1456,47 +1458,33 @@ impl Client {
                 &[MatrixVersion::V1_0],
                 Default::default(),
             )
-            .await?
-            .known_versions()
-            .collect();
+            .await?;
 
-        if server_versions.is_empty() {
-            Ok(vec![MatrixVersion::V1_0].into())
-        } else {
-            Ok(server_versions)
+        // Fill both unstable features and server versions at once.
+        let mut versions: Box<[MatrixVersion]> = resp.known_versions().collect();
+        if versions.is_empty() {
+            versions = vec![MatrixVersion::V1_0].into();
         }
+
+        Ok((versions, resp.unstable_features))
     }
 
     pub(crate) async fn server_versions(&self) -> HttpResult<&[MatrixVersion]> {
-        let server_versions = self
+        Ok(self
             .inner
             .server_versions
-            .get_or_try_init(|| Box::pin(self.request_server_versions()))
-            .await?;
+            .get_or_try_init(|| async {
+                let (versions, unstable_features) = self.request_server_capabilities().await?;
 
-        Ok(server_versions)
+                // Fill the unstable features, while we're at it.
+                self.inner.unstable_features.get_or_init(|| async { unstable_features }).await;
+
+                HttpResult::Ok(versions)
+            })
+            .await?)
     }
 
-    /// Fetch unstable_features from homeserver
-    async fn request_unstable_features(&self) -> HttpResult<BTreeMap<String, bool>> {
-        let unstable_features: BTreeMap<String, bool> = self
-            .inner
-            .http_client
-            .send(
-                get_supported_versions::Request::new(),
-                None,
-                self.homeserver().to_string(),
-                None,
-                &[MatrixVersion::V1_0],
-                Default::default(),
-            )
-            .await?
-            .unstable_features;
-
-        Ok(unstable_features)
-    }
-
-    /// Get unstable features from `request_unstable_features` or cache
+    /// Get unstable features from by fetching from the server or the cache.
     ///
     /// # Examples
     ///
@@ -1511,13 +1499,18 @@ impl Client {
     /// # anyhow::Ok(()) };
     /// ```
     pub async fn unstable_features(&self) -> HttpResult<&BTreeMap<String, bool>> {
-        let unstable_features = self
+        Ok(self
             .inner
             .unstable_features
-            .get_or_try_init(|| self.request_unstable_features())
-            .await?;
+            .get_or_try_init(|| async {
+                let (versions, unstable_features) = self.request_server_capabilities().await?;
 
-        Ok(unstable_features)
+                // Fill the versions, while we're at it.
+                self.inner.server_versions.get_or_init(|| async { versions }).await;
+
+                HttpResult::Ok(unstable_features)
+            })
+            .await?)
     }
 
     /// Check whether MSC 4028 is enabled on the homeserver.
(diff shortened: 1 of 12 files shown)


pixlwave:
Can confirm this works on EXI. First start-up I see `/versions` being hit, and then it's never used on subsequent launches.

Out of interest, what's the definition of `expired`, from a quick skim?

crates/matrix-sdk/src/client/mod.rs
```diff
@@ -1456,47 +1458,79 @@ impl Client {
...
+        match self.store().get_kv_data(StateStoreDataKey::ServerCapabilities).await {
+            Ok(Some(stored)) => {
+                if let Some((versions, unstable_features)) =
+                    stored.into_server_capabilities().and_then(|cap| cap.maybe_decode())
+                {
+                    return Ok((versions.into(), unstable_features));
+                }
+            }
+            Ok(None) => {
+                // fallthrough: cache is empty
```
poljar, line 1501: It's OK to have a call to `self.fetch_and_cache_server...` twice, I would say that this is more reasonable than having an early return and two empty clauses explaining why they are empty.

crates/matrix-sdk/src/client/mod.rs
```diff
@@ -239,10 +239,10 @@ pub(crate) struct ClientInner {
     base_client: BaseClient,
 
     /// The Matrix versions the server supports (well-known ones only)
-    server_versions: OnceCell<Box<[MatrixVersion]>>,
+    server_versions: RwLock<Option<Box<[MatrixVersion]>>>,
```
poljar, line 242: I think we should put those behind the same lock, as we saw in the memory store, it's tricky to remember how to avoid a deadlock when two pieces of information are accessed or updated in the same method but are living behind separate locks.

Further bellow we do acquire the locks in the same method.

crates/matrix-sdk/src/client/mod.rs
```diff
@@ -1510,16 +1544,41 @@ impl Client {
...
     /// # anyhow::Ok(()) };
     /// ```
-    pub async fn unstable_features(&self) -> HttpResult<&BTreeMap<String, bool>> {
-        let unstable_features = self
-            .inner
-            .unstable_features
-            .get_or_try_init(|| self.request_unstable_features())
-            .await?;
+    pub async fn unstable_features(&self) -> HttpResult<BTreeMap<String, bool>> {
+        if let Some(unstable_features) = self.inner.unstable_features.read().await.as_ref() {
```
poljar, line 1548: This is almost the same as `server_version()` I think we can remove a bunch of code here, especially if those two pieces of data live behind the same lock.

crates/matrix-sdk/src/client/mod.rs
```diff
@@ -1510,16 +1544,41 @@ impl Client {
...
     }
 
+    /// Empty the server version and unstable features cache.
+    ///
+    /// Since the SDK caches server capabilities (versions and unstable
+    /// features), it's possible to have a stale entry in the cache. This
+    /// functions makes it possible to force reset it.
+    pub async fn reset_server_capabilities(&self) -> Result<()> {
+        // Empty the in-memory caches.
+        *self.inner.server_versions.write().await = None;
```
poljar, line 1575: I think that this is fine, as you eagerly call drop in the previous methods, so there's no deadlock. But modifying any of the above methods could easily introduce one. I don't see a reason for this not being behind the same lock.

poljar (changes requested):
Looking good, left some nits and some preference about the locking situation.

crates/matrix-sdk/src/client/mod.rs
bnjbvr: *narrator's voice* there are in fact not two, but three fallthrough paths, if the `Ok(Some())`'s `if` is not taken. I don't quite see the issue with the comments?

poljar: Alright.

crates/matrix-sdk/src/client/mod.rs
```diff
@@ -1444,6 +1440,7 @@ impl Client {
             .send(SessionChange::UnknownToken { soft_logout: *soft_logout });
     }
 
+    /// Fetches server capabilities from network; no caching.
```
poljar, line 1443: Err, that's a very confusing doc for a method called `fetch_and_cache_...`.

crates/matrix-sdk/src/client/mod.rs
```diff
@@ -1466,26 +1463,11 @@ impl Client {
             versions.push(MatrixVersion::V1_0);
         }
 
-        let unstable_features = resp.unstable_features;
```
poljar: Why is it better to inline this in the caller method?

crates/matrix-sdk/src/client/mod.rs
```diff
@@ -1466,26 +1463,11 @@ impl Client {
...
-                    StateStoreDataValue::ServerCapabilities(encoded),
-                )
-                .await
-            {
-                warn!("error when caching server capabilities: {err}");
-            }
-        }
-
-        Ok((versions.into(), unstable_features))
+        Ok((versions.into(), resp.unstable_features))
```
poljar, line 1466: Might be nicer to return the `ServerCapabilities` struct.

poljar (approved):

crates/matrix-sdk/src/client/mod.rs
bnjbvr: yep lol

crates/matrix-sdk/src/client/mod.rs
bnjbvr: I don't understand this comment?

crates/matrix-sdk/src/client/mod.rs
bnjbvr: Hmm, it makes things a bit awkward:

- if we return `ServerCapabilities` (which holds the stringified Matrix versions), then we need to immediately deserialize it, and handle a potential error that never happens,
- if we return `ClientServerCapabilities` (which has both fields be `Option`al), then we need to immediately unwrap those after calling this method.

I'll keep the current form, as a result, unless you have an idea how to handle this better.

crates/matrix-sdk/src/client/mod.rs
poljar: Err, I missed a bit, I wanted to point at this piece of code: https://github.com/matrix-org/matrix-rust-sdk/pull/3658/commits/7917bbb7c6ded8b1ad7b402438fb2aa94560b4e1#diff-6d2ff673b14b2a563388b2d177b0c4e7ff2c16714054858fdec9bfd01858ba96L1471-L1484.

The caching of the result was moved out of this method, and I was wondering why that is considered to be better.

bnjbvr: I found that this layered responsibilities slightly better:

- one method is only concerned about network
- one method is only concerned about storage, and will resort to network if needs be, then cache in storage
- one method is only concerned about memory caching, and will resort to storage etc.

PR #3659, opened 2024-07-05
base: adjust trace logging in memory store

kegsay:
Now we have more information on which locks are implicated, refine the logs to identify the exact lock.

Follow up to https://github.com/matrix-org/matrix-rust-sdk/pull/3642

### DIFF
--- a/crates/matrix-sdk-base/src/store/memory_store.rs
+++ b/crates/matrix-sdk-base/src/store/memory_store.rs
@@ -290,10 +290,8 @@ impl StateStore for MemoryStore {
 
         if let Some(s) = &changes.sync_token {
             *self.sync_token.write().unwrap() = Some(s.to_owned());
-            trace!("assigned sync token");
         }
 
-        trace!("profiles");
         {
             let mut profiles = self.profiles.write().unwrap();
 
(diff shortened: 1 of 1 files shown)


crates/matrix-sdk-base/src/store/memory_store.rs
```diff
@@ -290,10 +290,8 @@ impl StateStore for MemoryStore {
 
         if let Some(s) = &changes.sync_token {
             *self.sync_token.write().unwrap() = Some(s.to_owned());
-            trace!("assigned sync token");
```
poljar, line 293: I wouldn't mind having those left in, after all they are trace logs. Feel free to disagree if you find them spammy.

poljar (approved):

crates/matrix-sdk-base/src/store/memory_store.rs
kegsay: I don't mind really. They aren't sufficiently spammy that it'd be a problem to keep them in.

kegsay: Added them back.

PR #3660, opened 2024-07-05
crypto: Expose new stream about room_key withheld messages

richvdh:
Part of the fix to https://github.com/element-hq/element-web/issues/27653.

[Please ignore the branch name; I appear to have fat-fingered]

### DIFF
--- a/crates/matrix-sdk-crypto/src/machine.rs
+++ b/crates/matrix-sdk-crypto/src/machine.rs
@@ -2372,7 +2372,7 @@ pub(crate) mod tests {
     };
 
     use assert_matches2::{assert_let, assert_matches};
-    use futures_util::{FutureExt, StreamExt};
+    use futures_util::{pin_mut, FutureExt, StreamExt};
     use itertools::Itertools;
     use matrix_sdk_common::deserialized_responses::{
         DeviceLinkProblem, ShieldState, UnableToDecryptInfo, UnsignedDecryptionResult,
@@ -3243,8 +3243,8 @@ pub(crate) mod tests {
             get_machine_pair_with_setup_sessions_test_helper(alice_id(), user_id(), false).await;
         let room_id = room_id!("!test:example.org");
 
-        let mut room_keys_withheld_received_stream =
-            Box::pin(bob.store().room_keys_withheld_received_stream());
+        let room_keys_withheld_received_stream = bob.store().room_keys_withheld_received_stream();
+        pin_mut!(room_keys_withheld_received_stream);
 
         let encryption_settings = EncryptionSettings::default();
         let encryption_settings = EncryptionSettings {
--- a/crates/matrix-sdk-crypto/src/machine.rs
+++ b/crates/matrix-sdk-crypto/src/machine.rs
@@ -2428,7 +2428,9 @@ pub(crate) mod tests {
         types::{
             events::{
                 room::encrypted::{EncryptedToDeviceEvent, ToDeviceEncryptedEventContent},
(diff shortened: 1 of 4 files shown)


crates/matrix-sdk-crypto/src/machine.rs
```diff
@@ -3241,6 +3243,9 @@ pub(crate) mod tests {
             get_machine_pair_with_setup_sessions_test_helper(alice_id(), user_id(), false).await;
         let room_id = room_id!("!test:example.org");
 
+        let mut room_keys_withheld_received_stream =
+            Box::pin(bob.store().room_keys_withheld_received_stream());
```
Hywan, line 3247: You may want to use [`pin_mut!`](https://docs.rs/futures-util/latest/futures_util/macro.pin_mut.html) here, as we do elsewhere in the code. It does a similar thing but without boxing the value.

crates/matrix-sdk-crypto/src/machine.rs
```diff
@@ -3280,6 +3285,21 @@ pub(crate) mod tests {
         .await
         .unwrap();
 
+        // We should receive a notification on the room_keys_withheld_received_stream
+        let withheld_received = room_keys_withheld_received_stream
+            .next()
+            .now_or_never()
```
Hywan, line 3291: To use `now_or_never`, you must be sure that the value will be immediately available in the `Stream`, otherwise the test will be flaky.

Is it a problem to use `next().await` here instead? The risk is that the test might be pending forever if something changes in the future, but it's still an indication that a regression has been detected. The pros, however, is that we are sure the test can never be flaky.

crates/matrix-sdk-crypto/src/store/crypto_store_wrapper.rs
```diff
@@ -68,6 +75,13 @@ impl CryptoStoreWrapper {
         let room_key_updates: Vec<_> =
             changes.inbound_group_sessions.iter().map(RoomKeyInfo::from).collect();
 
+        let withheld_session_updates: Vec<_> = changes
+            .withheld_session_info
+            .values()
+            .flat_map(|session_map| session_map.values())
+            .map(Clone::clone)
```
Hywan, line 82: Probably more Rust idiomatic to do as such:

```suggestion
            .flat_map(|session_map| session_map.values().cloned())
```

Hywan (commented):
I can only review the code, not the feature. A second pair of eyes from @poljar might be useful.

crates/matrix-sdk-crypto/src/machine.rs
richvdh: > Is it a problem to use `next().await` here instead?

I guess not. TBH I pretty much cargo-culted this from `test_megolm_encryption` above, which afaik has never shown any sign of being flaky. Flipping it around: is there any reason to believe that the value would *not* already be available in the Stream?

poljar: The stream is created before we receive the sync changes, the sync changes trigger an update, this means that the `now_or_never()` should never fail here, unless we introduce a bug.

There are no hidden tasks/threads or requests being sent out in `receive_sync_changes()` it's all a pretty dumb, for better or worse, state machine.

`now_or_never()` is the right thing to use, otherwise a regression will produce a hang instead of a loud test failure.

poljar (approved):
I think that this looks fine. I would probably add a doc example but not required since it's a simple stream and we have plenty examples for streams.

poljar:
Oh and thanks for taking the time to refactor the filter we had for each stream.

PR #3661, opened 2024-07-08
bench: Add the `room_list` bench

Hywan:
Example of the output:

```
room_list/sync/10       time:   [3.5673 ms 3.5899 ms 3.6181 ms]
                        thrpt:  [2.7639 Kelem/s 2.7856 Kelem/s 2.8032 Kelem/s]
Found 12 outliers among 100 measurements (12.00%)
  1 (1.00%) low severe
  3 (3.00%) high mild
  8 (8.00%) high severe
room_list/sync/100      time:   [3.6342 ms 3.6667 ms 3.7037 ms]
                        thrpt:  [27.000 Kelem/s 27.273 Kelem/s 27.516 Kelem/s]
Found 12 outliers among 100 measurements (12.00%)
  5 (5.00%) high mild
  7 (7.00%) high severe
room_list/sync/1000     time:   [30.426 ms 30.611 ms 30.810 ms]
                        thrpt:  [32.457 Kelem/s 32.668 Kelem/s 32.867 Kelem/s]
Found 20 outliers among 100 measurements (20.00%)
  13 (13.00%) high mild
  7 (7.00%) high severe
Benchmarking room_list/sync/10000: Warming up for 3.0000 s
Warning: Unable to complete 100 samples in 5.0s. You may wish to increase target time to 35.7s, or reduce sample count to 10.
room_list/sync/10000    time:   [353.56 ms 355.14 ms 356.97 ms]
                        thrpt:  [28.014 Kelem/s 28.158 Kelem/s 28.284 Kelem/s]
Found 5 outliers among 100 measurements (5.00%)
  3 (3.00%) high mild
  2 (2.00%) high severe
```

### DIFF
--- a/benchmarks/benches/room_list_bench.rs
+++ b/benchmarks/benches/room_list_bench.rs
@@ -1,4 +1,4 @@
-use criterion::{criterion_group, criterion_main, BenchmarkId, Criterion, Throughput};
+use criterion::{criterion_group, criterion_main, BatchSize, BenchmarkId, Criterion, Throughput};
 use futures_util::{pin_mut, StreamExt};
 use matrix_sdk::test_utils::logged_in_client_with_server;
 use matrix_sdk_ui::{room_list_service::filters::new_filter_non_left, RoomListService};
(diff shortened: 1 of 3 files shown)


benchmarks/benches/room_list_bench.rs
```diff
@@ -0,0 +1,176 @@
...
+
+    let mut benchmark = c.benchmark_group("room_list");
+
+    for number_of_rooms in [10, 100, 1000, 10_000] {
+        benchmark.throughput(Throughput::Elements(number_of_rooms));
+        benchmark.bench_with_input(
+            BenchmarkId::new("sync", number_of_rooms),
+            &number_of_rooms,
+            |benchmark, maximum_number_of_rooms| {
+                benchmark.to_async(&runtime).iter(|| async {
```
poljar, line 52: There are a lot of line in this benchmark that should be considered to be part of the setup. I don't think we want to time `wiremock` starting up.

Since there are so many lines that are timed that shouldn't be, I think it's quite hard to get the benchmark to settle down. I.e. after some 10 runs I finally got it to settle down, but still it's quite the noisy benchmark:

```
    time:   [-2.6332% +7.1045% +17.983%] (p = 0.18 > 0.05)
    thrpt:  [-15.242% -6.6332% +2.7044%]
    No change in performance detected.
```

Please take a look at the [`Bencher`](https://docs.rs/criterion/latest/criterion/struct.Bencher.html) documentation and more specifically at the [`iter_batched()`](https://docs.rs/criterion/latest/criterion/struct.AsyncBencher.html#method.iter_batched) method:

> If your routine requires some per-iteration setup that shouldn’t be timed, use iter_batched or iter_batched_ref.

Having that many lines as part of the benchmark also makes it quite hard to see what exactly we're measuring. So what are we intending to measure? The sync response processing, the room list creation?

poljar (changes requested):

poljar:
We're not particularly interested in this as of now and this is bitrotting, so closing now. If we ever decide to benchmark this properly let's reopen.

PR #3665, opened 2024-07-08
timeline: adjust day divider and read marker items after a local echo has been redacted/edited

bnjbvr:
Part of #3361 polish.


Hywan (approved):
A test would be lovely, otherwise it looks good

Hywan (approved):
The commit c0590de7b3ab3fecca9e5279531a952bed6039ae has the following message: `timelien: remove…`. Can you rename it to `timeline: remove…` please?

Otherwise, it's good to me! Thanks for the clean up and the regression test.

PR #3668, opened 2024-07-09
memory-store: release locks earlier to avoid deadlocks

kegsay:
I've been debugging a cause of flakey complement-crypto tests for about a month now. I was pretty convinced it was deadlocking somewhere in the memory store `save_changes` code. With additional logging, it's now clear that the there is an ABBA style deadlock when `save_changes` is called at the same time as `get_state_events`.

I've also adjusted code for `get_user_ids` as it has a very similar pattern and also acquires locks in reverse order to `save_changes`, so is potentially vulnerable to this.


poljar (approved):
A regression test would be nice, but I guess we can cheat a bit and use complement for this.

PR #3670, opened 2024-07-09
sdk: Add a higher level method to reset the cross-signing keys

poljar:
This method supports OIDC as well as the usual UIAA method of cross-signing reset, it attempts to guide the user more easily towards a successful reset.

The cross-signing bootstrap example was modified to use the new method and the functionality was exposed in the oidc-cli example as well.

### DIFF
--- a/crates/matrix-sdk/src/encryption/mod.rs
+++ b/crates/matrix-sdk/src/encryption/mod.rs
@@ -36,14 +36,16 @@ use matrix_sdk_base::crypto::{
 use matrix_sdk_common::executor::spawn;
 use ruma::{
     api::client::{
+        error::{ErrorBody, ErrorKind},
         keys::{
-            get_keys, upload_keys, upload_signing_keys::v3::Request as UploadSigningKeysRequest,
+            get_keys, upload_keys, upload_signatures::v3::Request as UploadSignaturesRequest,
+            upload_signing_keys::v3::Request as UploadSigningKeysRequest,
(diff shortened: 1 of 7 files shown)


crates/matrix-sdk/src/encryption/mod.rs
```diff
@@ -227,6 +230,121 @@ impl CrossProcessLockStoreGuardWithGeneration {
...
+    }
+
+    /// Continue the cross-signing reset by either waiting for the
+    /// authentication to be done on the side of the OIDC issuer or by
+    /// providing additional [`AuthData`] the homeserver requires.
+    pub async fn auth(&self, auth: Option<AuthData>) -> Result<()> {
+        let mut upload_request = self.upload_request.clone();
+        upload_request.auth = auth;
+
+        // TODO: Do we want to put a limit on this infinite loop? 🤷
```
Hywan, line 265: Yes 🙂

Hywan (approved):
LGTM for me. I would like to see to todo about the potential infinite loo being addressed if possible, but it's probably not a blocker for the moment.

crates/matrix-sdk/src/encryption/mod.rs
poljar: Which kind of limit, time-based or attempt based? The auth team argued that there's no need for a limit, which I can see since the user can just cancel the request/future if they would like to abort this.

PR #3671, opened 2024-07-09
docs: Fix the OIDC QR code login example in the docs

poljar:

Signed-off-by:


bnjbvr (approved):
Would be sweet to have our doc-tests actually run and test things :)

poljar:
> Would be sweet to have our doc-tests actually run and test things :)

Our doc-tests actually run and test things, but they don't enable the `experimental-oidc` feature flag. Instead of adding the feature, perhaps we should consider removing the feature flag?

bnjbvr:
> perhaps we should consider removing the feature flag?

Sounds good to me. At some point we also discussed having a `new-features` feature flag for both sliding sync and OIDC stuff that was not stable yet. Not sure what the stability level of OIDC is, at this point.

PR #3672, opened 2024-07-09
fix(sdk): force room member reload after inviting a user

### LINKED ISSUE #3622: Delayed membership responses in /sync cause UTDs
This outlines a race condition in the CSAPI which can cause UTDs.

Consider:
- Alice is in a E2EE room and invites Bob. The request 200 OKs but has yet to come down /sync.
 - Alice tries to send a message in the room.
 - Alice should encrypt for Bob.

In practice, clients will not encrypt for Bob, causing a UTD if you very quickly send an encrypted message after inviting a user. This can happen due to:
 - A bot is using the SDK so message sending is very quick after an invite
 - DMs: for better UX a room is only made / user invited / message sent, when someone _sends_ a message in a DM room rather than just selects the user's room. This means the message is already ready to go at the same time as invite time.

Rust SDK is slightly better than JS SDK on this, because Rust SDK will hit `/members` for the first message it sends, so it isn't reliant on `/sync` for that one request. In practice, this means it handles the DM case.

To fix this, we should be remembering that we, the client, have modified the membership state of the room, and invalidate the room member list (so we hit /members again). We can't assume that a 200 OK to /invite will guarantee that the user is in an invited state, so we still need to defer to the server.

A test for this is in https://github.com/matrix-org/complement-crypto/pull/98

jmartinesp:
This is needed to prevent the race condition where the invite request finished, the `/sync` one didn't fetch the new membership event yet and we send a message in the room. This message won't be encrypted for the newly invited user and will result in an UTD.

I added a new integration test and I can confirm this [complement-crypto test](https://github.com/matrix-org/complement-crypto/pull/98) now passes instead of being skipped.

Fixes #3622.

Signed-off-by:

### DIFF
--- a/testing/matrix-sdk-integration-testing/src/tests/sliding_sync/room.rs
+++ b/testing/matrix-sdk-integration-testing/src/tests/sliding_sync/room.rs
@@ -907,6 +907,18 @@ async fn test_delayed_invite_response_and_sent_message_decryption() -> Result<()
     alice_room.invite_user_by_id(bob.user_id().unwrap()).await.unwrap();
     alice_room.send(RoomMessageEventContent::text_plain("hello world")).await?;
 
+    // Wait until Bob receives the invite
+    let bob_room_stream = bob.sync_stream(SyncSettings::new()).await;
+    pin_mut!(bob_room_stream);
+
+    while let Some(Ok(response)) =
+        timeout(Duration::from_secs(3), bob_room_stream.next()).await.expect("Room sync timed out")
+    {
+        if response.rooms.invite.contains_key(alice_room.room_id()) {
+            break;
+        }
+    }
+
     // Join the room from Bob's client
     let bob_room = bob.get_room(alice_room.room_id()).unwrap();
     bob_room.join().await?;
--- a/testing/matrix-sdk-integration-testing/src/tests/sliding_sync/room.rs
+++ b/testing/matrix-sdk-integration-testing/src/tests/sliding_sync/room.rs
@@ -881,38 +881,16 @@ async fn test_delayed_invite_response_and_sent_message_decryption() -> Result<()
         .send(RoomMessageEventContent::text_plain("dummy message to make members call"))
         .await?;
 
-    // Set up sliding sync for bob.
-    let sliding_bob = bob
-        .sliding_sync("main")?
-        .with_all_extensions()
-        .poll_timeout(Duration::from_secs(2))
-        .network_timeout(Duration::from_secs(2))
-        .add_list(
-            SlidingSyncList::builder("all")
-                .sync_mode(SlidingSyncMode::new_selective().add_range(0..=20)),
-        )
-        .build()
-        .await?;
-
-    let s = sliding_bob.clone();
-    let sliding_bob_handle = spawn(async move {
-        let stream = s.sync();
-        pin_mut!(stream);
-        while let Some(up) = stream.next().await {
-            warn!("bob received update: {up:?}");
-        }
-    });
-
     // Send the invite to Bob and a message to reproduce the edge case
     alice_room.invite_user_by_id(bob.user_id().unwrap()).await.unwrap();
     alice_room.send(RoomMessageEventContent::text_plain("hello world")).await?;
 
     // Wait until Bob receives the invite
-    let bob_room_stream = bob.sync_stream(SyncSettings::new()).await;
-    pin_mut!(bob_room_stream);
+    let bob_sync_stream = bob.sync_stream(SyncSettings::new()).await;
+    pin_mut!(bob_sync_stream);
 
     while let Some(Ok(response)) =
-        timeout(Duration::from_secs(3), bob_room_stream.next()).await.expect("Room sync timed out")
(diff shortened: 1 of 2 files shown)


testing/matrix-sdk-integration-testing/src/tests/sliding_sync/room.rs
```diff
@@ -853,6 +854,97 @@ async fn test_delayed_decryption_latest_event() -> Result<()> {
...
+        .network_timeout(Duration::from_secs(2))
+        .add_list(
+            SlidingSyncList::builder("all")
+                .sync_mode(SlidingSyncMode::new_selective().add_range(0..=20)),
+        )
+        .build()
+        .await?;
+
+    let s = sliding_bob.clone();
+    spawn(async move {
```
bnjbvr, line 898: nit: in general it's a better practice to `spawn` a task and keep the `join_handle` around, so that we can `.abort()` or `.await` on it later. In this case, we should abort at the end of the test.

testing/matrix-sdk-integration-testing/src/tests/sliding_sync/room.rs
```diff
@@ -853,6 +854,97 @@ async fn test_delayed_decryption_latest_event() -> Result<()> {
...
+            warn!("bob received update: {up:?}");
+        }
+    });
+
+    // Send the invite to Bob and a message to reproduce the edge case
+    alice_room.invite_user_by_id(bob.user_id().unwrap()).await.unwrap();
+    alice_room.send(RoomMessageEventContent::text_plain("hello world")).await?;
+
+    // Join the room from Bob's client
+    let bob_room = bob.get_room(alice_room.room_id()).unwrap();
```
bnjbvr, line 911: The test is inherently racy because of this line: bob may not have received the room info from its sliding sync at this point. In fact, the test isn't passing on my machine, because it's failing at this point (before and after the fix).

There are a few ways to handle this:

- introduce a large sleep()
- loop until the `get_room()` returns Some, with some exponential backoff (sleep between attempts)
- there's likely a stream of room changes that we could listen to, and wait (with a timeout) until we get the room with the room id created by alice.

testing/matrix-sdk-integration-testing/src/tests/sliding_sync/room.rs
```diff
@@ -853,6 +854,97 @@ async fn test_delayed_decryption_latest_event() -> Result<()> {
...
+            .expect("Timed out. Neither an UTD or the sent message were found")
+            .unwrap();
+        if let VectorDiff::PushFront { value } = diff {
+            if let Some(event) = value.as_event() {
+                if let Some(message) = event.content().as_message() {
+                    if message.body() == "hello world" {
+                        return Ok(());
+                    } else {
+                        panic!("Unexpected message event found");
+                    }
```
bnjbvr, line 939: nit: else after return.

```suggestion
                    if message.body() == "hello world" {
                        return Ok(());
                    }
                    panic!("Unexpected message event found");
```

bnjbvr (changes requested):
The fix's code looks good to me, thanks for fixing!

And welcome to the realm of shady intermittently failing integration tests. There's no better baptism of fire… Good luck! :crossed_fingers:

testing/matrix-sdk-integration-testing/src/tests/sliding_sync/room.rs
jmartinesp: https://github.com/matrix-org/matrix-rust-sdk/pull/3672/commits/75dfe35c127d4f4bb768126a4adef21878649c4a

testing/matrix-sdk-integration-testing/src/tests/sliding_sync/room.rs
jmartinesp: https://github.com/matrix-org/matrix-rust-sdk/pull/3672/commits/f6a38134d60e241d04b0335d46cf45dbbe641554

testing/matrix-sdk-integration-testing/src/tests/sliding_sync/room.rs
jmartinesp: https://github.com/matrix-org/matrix-rust-sdk/pull/3672/commits/66c286879f6a1b19eb2a1bdfd0e8b7a7f630edad

testing/matrix-sdk-integration-testing/src/tests/sliding_sync/room.rs
```diff
@@ -853,6 +854,110 @@ async fn test_delayed_decryption_latest_event() -> Result<()> {
...
+        .await?;
+
+    let s = sliding_bob.clone();
+    let sliding_bob_handle = spawn(async move {
+        let stream = s.sync();
+        pin_mut!(stream);
+        while let Some(up) = stream.next().await {
+            warn!("bob received update: {up:?}");
+        }
+    });
```
bnjbvr, line 904: Is this needed? Below we're calling `sync_stream()`, which will call `sync_once()` repeatedly, and that uses sync v2 as far as I can tell.

bnjbvr (approved):
Thank you :partying_face:

PR #3673, opened 2024-07-09
tests(timeline): make use of the `EventFactory` more

bnjbvr:
The `TestTimeline` had lots and lots of custom methods for creating test events and handling them immediately. Since the `EventFactory` helper makes it possible to easily create events, there's no need to repeat those responsibilities in the timeline testing code itself, and we can make more use of the `EventFactory` instead.

There's much more to do there; ideally, we'd get rid of most of `TestTimeline::handle_` methods (we'd keep `handle_live_events`), as well as `EventBuilder` entirely, which API isn't flexible at all. Could be some nice and fun good-first-bugs :sunglasses:


andybalaam (approved):
Great, thanks!

PR #3674, opened 2024-07-09
crypto: Pass room id and session id to `room_keys_withheld_received_stream`

richvdh:
Followup on #3660: I realised I didn't want to have to parse the room_id and session_id out of the messages again.

### DIFF
--- a/crates/matrix-sdk-crypto/src/store/mod.rs
+++ b/crates/matrix-sdk-crypto/src/store/mod.rs
@@ -913,6 +913,20 @@ impl From<&InboundGroupSession> for RoomKeyInfo {
     }
 }
 
+/// Information on a room key that has been withheld
+#[derive(Clone, Debug, Deserialize, Serialize)]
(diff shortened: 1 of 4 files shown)


crates/matrix-sdk-crypto/src/store/mod.rs
```diff
@@ -913,6 +913,20 @@ impl From<&InboundGroupSession> for RoomKeyInfo {
     }
 }
 
+/// Information on a room key that has been withheld
+#[derive(Clone, Debug, Deserialize, Serialize)]
+pub struct RoomKeyWithheldInfo {
```
poljar, line 918: Might make sense to use this in the store as well, maybe? But probably not worth to do a migration for that.

poljar (approved):

crates/matrix-sdk-crypto/src/store/mod.rs
richvdh: Yeahhh. Maybe some other time.

Issue #3647, opened 2024-07-03
[meta] SSS & PPS


Hywan:
_Here be dragons._

(that's a draft)

This is a meta issue to track the work about _Simplified Sliding Sync_ (SSS) and _Pragmatic Pragmatic Sync_ (PPS)

# Simplified Sliding Sync

Synapse is implementing Sliding Sync natively. Since [the Proxy implementation](https://github.com/matrix-org/sliding-sync) was experimental and was exposing too many features we now know we don't need, we call this project Simplified Sliding Sync.

## Tasks

- [x] Replace the sync `ops` by client-side sorting
- [ ] To Be Defined

# Pragmatic Pragmatic Sync

Pragmatic Pragmatic Sync is another effort to extract some features of Sliding Sync into their own HTTP endpoints.

## Tasks

- [ ] Replace Sliding Sync by `/sync/e2ee`
- [ ] Replace Sliding Sync by the improved `/context`

# Release and Clean Up

Now it's time to clean up the old `matrix_sdk::sliding_sync` implementation.

## Tasks

- [ ] Remove old features of Sliding Sync
- [ ] Remove potential intermediate feature flags
- [ ] Remove the `experimental-sliding-sync` feature flag

PR #3676, opened 2024-07-10
feat: Migrate from Sliding Sync to Simplified Sliding Sync

Hywan:
We want to support the Simplified MSC3575 (temporary name).

This PR should be reviewed patch-by-patch. It's not super complex, but you need to understand the approach:

1. Create a `matrix_sdk_base::sliding_sync::http` module, which provides all the HTTP types for MSC3575,
2. Migrate the entire codebase to use this module instead of `ruma`: it's just type aliases at this point, but it's important for the next steps; tl;dr: `s/ruma::…::sync::v4/matrix_sdk_base::sliding_sync::http:ss`,
3. This `http` module then provides HTTP types for Simplified MSC3575 (!) + conversions from MSC3575 to Simplified MSC3575, and exposes Simplfied MSC3575 as the default types,
4. Switch the entire codebase to use Simplified MSC3575; tl;dr: `s/http:ss/http` + remove features from MSC3575 that are no longer used in Simplified MSC3575 (like `sort`, `bump_event_types`, `unsubscribe_rooms`) or rename some features (like `timestamp` to `bump_stamp`),
5. Finally, since now the entire codebase uses Simplified MSC3575, we need a mechanism to convert the Simplified MSC3575 `Request` into a MSC3575 `Request` and the opposite for the `Response`. It's done with a `experimental-not-simplified-sliding-sync` flag.
6. … Bonus: `experimental-not-simplified-sliding-sync` turned to be difficult to use esp. when testing the whole project with `--workspace`. Instead, we went a bit further and replace this compiler feature flag by a runtime configuration on the `Client` with `Client::is_simplified_sliding_sync_enabled`. That way we can decide at runtime if we want Simplified MSC3575 or the ol' MSC3575.

Why this approach?

- The day we fully sunset MSC3575, we simply remove `Client::is_simplified_sliding_sync_enabled`
- Since we can decide at runtime to enable Simplified or regular MSC3575, it's easier for the apps (like Element X, Fractal etc.) to play with it.

---

* Address https://github.com/matrix-org/matrix-rust-sdk/issues/3647


bnjbvr (changes requested):
Please request another round of review when the PR is stable and there are no new commits while I'm reviewing :)

bnjbvr (dismissed):
Cool to see SSS coming to life, I appreciate the amount of work on this and done in Ruma under the hood :clap:

There are a few strong no-noes in that PR that make it impossible to land in the current shape.

https://github.com/matrix-org/matrix-rust-sdk/pull/3676/commits/9961db5037de762d9e7cc077c4cc7dbfb4a3ee38 should be in Ruma, not in our code. Too many code smells, likely too hard to maintain in the long term.

a9f22d1feae76ffd91f88b5e4861d7e269a6aaba was super hard to review. From a glance, it would have seemed much much simpler to review if it were split into smaller commits:

- get rid of the delta token (to fuse with the latest commit)
- get rid of the unsubscribe mechanism (why can we do that? we still have to support SS as it is right now)
- remove the `::ss` namespacing in most of the code
- rename the `recency_timestamp` to `recency_stamp` and associated code

I understand that all these changes must come before the actual migration from SS to SSS, but it would've been much much simpler to review if those changes were smaller (usually when i see a commit titled "rename X to Y", if i agree with the renaming i'll skim the code real-quick, and that takes me much less time, than trying to disentangle many changes that live in the same commit).

Pretty sure there's a big regression for SS, since we get rid of the `unsubscribe()` mechanism a bit too early in the patch set.

Also: do we _need_ a8a1315cb7cc6a8d58cf1fad56269772630fb304? I thought we said that having static configuration was sufficient…

crates/matrix-sdk-ui/src/room_list_service/room.rs
```diff
@@ -110,13 +110,6 @@ impl Room {
             .subscribe_to_room(self.inner.room.room_id().to_owned(), Some(settings))
     }
 
-    /// Unsubscribe to this room.
-    ///
-    /// It's the opposite method of [Self::subscribe`].
-    pub fn unsubscribe(&self) {
-        self.inner.sliding_sync.unsubscribe_from_room(self.inner.room.room_id().to_owned())
-    }
```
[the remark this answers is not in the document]
Hywan: Because SSS doesn't have room unsubscriptions and will never have it.

crates/matrix-sdk/src/sliding_sync/README.md
```diff
@@ -1,4 +1,4 @@
-Sliding Sync Client implementation of [MSC3575][MSC] & extensions
+simplified_Sliding Sync Client implementation of [MSC3575][MSC] & extensions
```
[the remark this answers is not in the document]
Hywan: Oh, well spotted, thanks!

bindings/matrix-sdk-ffi/Cargo.toml
```diff
@@ -57,6 +57,7 @@ features = [
     "e2e-encryption",
     "experimental-oidc",
     "experimental-sliding-sync",
+    "experimental-not-simplified-sliding-sync", # we don't want simplified MSC3575 yet!
```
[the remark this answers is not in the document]
Hywan: It's fine, we've discussed about that with the team, and it's OK to not being able to unsubscribe until SSS is deployed. Plus, SSS will not have room unsubscription, which guides everyone to adopt the new API.

crates/matrix-sdk-base/src/sliding_sync/http.rs
```diff
@@ -12,8 +12,59 @@
...
+
+impl<K, T, U> From<BTreeMap<K, T>> for BTreeMap<K, U>
+where
+    K: Ord,
+    U: From<T>,
+{
+    fn from(value: BTreeMap<K, T>) -> Self {
+        value.into_iter().map(|(key, value)| (key, From::from(value))).collect()
+    }
+}
```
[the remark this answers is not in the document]
Hywan: I agree. Conversions between `v4` and `v5` has been moved inside `ruma`, making our code much simpler and avoiding the use of specific `From` and `Into` traits with their blanket impls.

crates/matrix-sdk-base/src/sliding_sync/http.rs
```diff
@@ -12,8 +12,59 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-pub mod ss {
-    use ruma::api::client::sync::sync_events::v4;
+use std::collections::BTreeMap;
+
+pub trait From<T>: Sized {
+    fn from(value: T) -> Self;
+}
```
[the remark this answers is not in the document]
Hywan: I agree. Conversions between `v4` and `v5` has been moved inside `ruma`, making our code much simpler and avoiding the use of specific `From` and `Into` traits.

crates/matrix-sdk/src/sliding_sync/cache.rs
```diff
@@ -509,7 +496,6 @@ mod tests {
             .expect("must have restored sliding sync fields");
 
         // After restoring, the delta token and to-device token could be read.
```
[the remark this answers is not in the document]
Hywan: Good eye, fixed.

crates/matrix-sdk/src/sliding_sync/cache.rs
```diff
@@ -547,7 +532,6 @@ mod tests {
 
         // After restoring, the delta token, the to-device since token, stream
```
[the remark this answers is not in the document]
Hywan: Eagle eye. Fixed!

crates/matrix-sdk-base/src/sliding_sync/http.rs
```diff
@@ -12,9 +12,21 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+//! HTTP types for (Simplified) MSC3575.
+//!
+//! This module provides a unified namings for types from MSC3575 and
```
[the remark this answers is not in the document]
Hywan: Indeed, thanks.

testing/matrix-sdk-integration-testing/src/helpers.rs
```diff
@@ -85,6 +85,9 @@ impl TestClientBuilder {
             .user_agent("matrix-sdk-integration-tests")
             .homeserver_url(homeserver_url)
             .sliding_sync_proxy(sliding_sync_proxy_url)
+            // Disable Simplified MSC3575 for the integration tests as, at the time of writing, we
```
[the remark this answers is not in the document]
Hywan: 😅 indeed!

I've opened the issue, good idea: https://github.com/matrix-org/matrix-rust-sdk/issues/3698. It's also linked in https://github.com/matrix-org/matrix-rust-sdk/issues/3647 for maximum visibility.

crates/matrix-sdk-base/src/sliding_sync/http.rs
```diff
@@ -266,6 +266,12 @@ pub mod simplified_msc3575 {
 
     use super::{msc3575 as ss, From};
 
+    impl From<Response> for Response {
+        fn from(value: Response) -> Self {
+            value
+        }
+    }
+
```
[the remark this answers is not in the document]
Hywan: Yeah, it was required to make `http::Into::<http::simplified_msc3575>::into` to work on both `Response` types (from `simplified_msc3575` and `msc3575`). That's to mimic [the reflexive implementation of `std::convert::From`](https://doc.rust-lang.org/std/convert/trait.From.html#generic-implementations). Note that this has been removed since we do the conversion between these types directly in `ruma` with `std::convert::From` and `std::convert::Into`.

crates/matrix-sdk/src/sliding_sync/mod.rs
```diff
@@ -536,127 +539,152 @@ impl SlidingSync {
...
+        };
 
-        // Send the request and get a response with end-to-end encryption support.
-        //
-        // Sending the `/sync` request out when end-to-end encryption is enabled means
-        // that we need to also send out any outgoing e2ee related request out
-        // coming from the `OlmMachine::outgoing_requests()` method.
+        // The sending logic. It takes a generic `Request` because it can be a
+        // Simplified MSC3575 or a MSC3575 `Request`.
+        async fn send<Request>(
```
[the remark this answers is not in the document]
Hywan: Yup, good idea!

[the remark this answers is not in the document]
bnjbvr: Sounds good!

crates/matrix-sdk-base/src/sliding_sync/http.rs
```diff
@@ -0,0 +1,45 @@
...
+            ReceiptsConfig as RequestReceipts, ToDeviceConfig as RequestToDevice,
+            TypingConfig as RequestTyping,
+        };
+
+        // Response types.
+        pub use super::v4::{
+            AccountData as ResponseAccountData, Receipts as ResponseReceipts,
+            ToDevice as ResponseToDevice, Typing as ResponseTyping, E2EE as ResponseE2EE,
+        };
+    }
```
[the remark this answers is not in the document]
Hywan: I like the proposal with `request::` and `response::` modules. I believe it brings enough decoupling to justify removing the `extensions::` module. We can keep `Request` and `Response` at the “root”. Thus we would have `http::Request`, `http::Response`, `http::request::List`, `http::request::E2EE`, `http::response::E2EE` etc.

crates/matrix-sdk-base/src/rooms/normal.rs
```diff
@@ -989,15 +989,15 @@ pub struct RoomInfo {
...
     ///
     /// It's not to be confused with `origin_server_ts` of the latest event.
     /// Sliding Sync might "ignore” some events when computing the recency
-    /// timestamp of the room. Thus, using this `recency_timestamp` value is
+    /// stamp of the room. Thus, using this `recency_stamp` value is
     /// more accurate than relying on the latest event.
     #[cfg(feature = "experimental-sliding-sync")]
     #[serde(default)]
-    pub(crate) recency_timestamp: Option<MilliSecondsSinceUnixEpoch>,
+    pub(crate) recency_stamp: Option<u64>,
```
poljar, line 1000: I guess we don't need a serde `alias`here because the types have different semantic meaning?

crates/matrix-sdk-base/src/rooms/normal.rs
```diff
@@ -923,12 +923,12 @@ impl Room {
         self.inner.read().base_info.is_marked_unread
     }
 
-    /// Returns the recency timestamp of the room.
+    /// Returns the recency stamp of the room.
     ///
-    /// Please read `RoomInfo::recency_timestamp` to learn more.
+    /// Please read `RoomInfo::recency_stamp` to learn more.
```
poljar, line 928: You should link to this if you'd like to reference docs on another struct. That being said, I don't think that you can sensibly link to the `recency_stamp` field since it's `pub(crate)` and not `pub` like this method.

```suggestion
    /// Please read [`RoomInfo::recency_stamp`] to learn more.
```

crates/matrix-sdk-base/src/sliding_sync/http.rs
```diff
@@ -0,0 +1,50 @@
+// Copyright 2023 The Matrix.org Foundation C.I.C.
```
poljar, line 1:
```suggestion
// Copyright 2024 The Matrix.org Foundation C.I.C.
```

testing/matrix-sdk-integration-testing/Cargo.toml
```diff
@@ -16,7 +16,8 @@ futures = { version = "0.3.29", features = ["executor"] }
 futures-core = { workspace = true }
 futures-util = { workspace = true }
 http = { workspace = true }
-matrix-sdk = { workspace = true, default-features = true, features = ["testing", "qrcode"] }
+matrix-sdk-base = { workspace = true, default-features = true, features = ["testing", "qrcode"] }
```
poljar, line 19: Might be nice to preserve the sort order, base comes after the one without a suffix.

crates/matrix-sdk-ui/src/room_list_service/sorters/recency.rs
```diff
@@ -66,11 +64,11 @@ where
 }
 
 /// Create a new sorter that will sort two [`Room`] by recency, i.e. by
-/// comparing their [`matrix_sdk_base::RoomInfo::recency_timestamp`] value. The
-/// `Room` with the newest recency timestamp comes first, i.e. newest < oldest.
+/// comparing their [`matrix_sdk_base::RoomInfo::recency_stamp`] value. The
```
poljar, line 67: An import behind a `#[cfg(doc)]` is usually much nicer.

```suggestion
/// comparing their [`RoomInfo::recency_stamp`] value. The
```

crates/matrix-sdk/src/client/builder.rs
```diff
@@ -107,6 +109,9 @@ impl ClientBuilder {
             requires_sliding_sync: false,
             #[cfg(feature = "experimental-sliding-sync")]
             sliding_sync_proxy: None,
+            // Simplified MSC3575 is turned on by default for the SDK.
+            #[cfg(feature = "experimental-sliding-sync")]
+            is_simplified_sliding_sync_enabled: true,
```
poljar, line 114: Hmm, what happens if the server doesn't support triple S?

crates/matrix-sdk/src/sliding_sync/cache.rs
```diff
@@ -88,10 +88,7 @@ pub(super) async fn store_sliding_sync_state(
     storage
         .set_custom_value(
             instance_storage_key.as_bytes(),
-            serde_json::to_vec(&FrozenSlidingSync::new(
-                position,
-                &*sliding_sync.inner.rooms.read().await,
-            ))?,
+            serde_json::to_vec(&FrozenSlidingSync::new(&*sliding_sync.inner.rooms.read().await))?,
```
poljar, line 91: Oh nou, this frozen thing still exists :smiling_face_with_tear:

poljar (approved):
I think that this looks fine, the PR is a bit chunky but most of it are type swaps.

I left some small nits, feel free to merge once addressed. Bonus points if you get the history cleaned up as well.

crates/matrix-sdk-base/src/rooms/normal.rs
Hywan: Exactly

Cargo.toml
```diff
@@ -47,7 +47,7 @@ once_cell = "1.16.0"
 pin-project-lite = "0.2.9"
 rand = "0.8.5"
 reqwest = { version = "0.12.4", default-features = false }
-ruma = { git = "https://github.com/ruma/ruma", rev = "c37843e9be619ffac8c4d33ad3a6a175cc32610c", features = [
+ruma = { git = "https://github.com/matrix-org/ruma", branch = "feat-sss", features = [
```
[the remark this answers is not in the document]
Hywan: Sure

crates/matrix-sdk-base/src/rooms/normal.rs
Hywan: Exactly, the field is not public so rustdoc would complain.

crates/matrix-sdk-base/src/sliding_sync/http.rs
Hywan: Done

testing/matrix-sdk-integration-testing/Cargo.toml
Hywan: Done, thanks

crates/matrix-sdk-ui/src/room_list_service/sorters/recency.rs
Hywan: If you prefer a shorter link, I prefer to use the following format:

```md
[`RoomInfo::recency_stamp`] blah blah blah

[`RoomInfo::recency_stamp`]: matrix_sdk_base::RoomInfo::recency_stamp
```

crates/matrix-sdk/src/client/builder.rs
Hywan: The SDK has to be configured to choose between SS or SSS. I believe the same question holds if the server doesn't support the SS proxy? If we decide to choose `false` for the default option, it means we have to update the tests in the opposite direction (when mocking the response from the server).

crates/matrix-sdk/src/sliding_sync/cache.rs
Hywan: Yeah but its days are counted.

crates/matrix-sdk-ui/src/room_list_service/sorters/recency.rs
poljar: I think that this format won't autodetect dead links. So leave it as is if you disagree with the import and `#[cfg(doc)]`.

crates/matrix-sdk/src/client/builder.rs
poljar: Alright, feel free to leave it as is.

crates/matrix-sdk-base/src/rooms/normal.rs
poljar: Well yes, but how would the user find the docs as well, since you are referencing docs for a `pub(crate)` field?

Issue #3543, opened 2024-06-13
Sender Data: Fetch and store sender data when we receive a new megolm session


andybalaam:
**Update:** the algorithm below describes background tasks but after [discussion](https://github.com/matrix-org/matrix-rust-sdk/pull/3590#discussion_r1671860791) on a PR for this, we think we can avoid that, and instead trigger a retry when `receive_keys_query_response` runs. I updated https://github.com/matrix-org/matrix-rust-sdk/issues/3546 to specify that we should do this during `receive_keys_query_response` as well as on a schedule.

When we receive a to-device message containing a megolm session, fetch and store the sender data for it.

Part of https://github.com/matrix-org/matrix-rust-sdk/issues/3544 which is part of [Invisible Crypto](https://github.com/matrix-org/matrix-spec-proposals/pull/4153). Depends on https://github.com/matrix-org/matrix-rust-sdk/issues/3542 because it holds the information in the data structures created there.

This task includes adding a `SessionManager` to prevent clashes between concurrent tasks.

## Preventing clashes

Cross-process, we are protected by the cross-process lock - the entire OlmMachine will be reloaded if some other process takes the lock. But we need a way to prevent 2 tasks both updating a session at the same time.

Possibly something like this (but could do with some validation from the Rust team):

```
struct SessionManager {
    sessions_being_processed: HashSet<OwnedSessionId>
}

impl SessionManager {
    fn try_lock(session_id: &SessionId) -> Option<SessionGuard>;  // If None, give up on this session
}
```

Pass the `SessionGuard` in to any async tasks you spawn. I.e. you keep hold of the lock even across the async boundary.

## Algorithm

When we receive a to-device message establishing a megolm session:

### A (start)

* [take the lock] TODO: if we fail to get it, can we just bail out here, dropping the information in this to-device message?  What if it contained device info that we need? How will this work if someone maliciously sent us a duplicate of someone else's session id?

* Does the to-device message contain the device_keys property from [MSC4147](https://github.com/matrix-org/matrix-spec-proposals/pull/4147)? Yes->**D** No->**B**

### B (no device info in to-device message)

We need to find the device details. If we have them in the store, we should use them immediately (rather than waiting for a background task to pick up the session for further processing).

* Does the locally-cached (in the store) devices list contain a device with the curve key of the sender of the to-device message? Yes->**D** No->**C**

### C (no device info locally)

[...]

PR #3677, opened 2024-07-10
crypto: Rename msk to master_key for consistency with the wider codebase

andybalaam:
As suggested in https://github.com/matrix-org/matrix-rust-sdk/pull/3590#discussion_r1671945866

Part of https://github.com/matrix-org/matrix-rust-sdk/issues/3543

### DIFF
--- a/crates/matrix-sdk-crypto/src/olm/group_sessions/sender_data.rs
+++ b/crates/matrix-sdk-crypto/src/olm/group_sessions/sender_data.rs
@@ -60,13 +60,13 @@ pub enum SenderData {
         user_id: OwnedUserId,
 
         /// The cross-signing key of the user who established this session.
-        msk: Ed25519PublicKey,
+        master_key: Ed25519PublicKey,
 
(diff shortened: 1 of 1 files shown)


bnjbvr (approved):
Looks like this wasn't tested at all :eyes: Bonus points if you add a test that reads and checks those values.

poljar (approved):
Thank you.

poljar:
> Looks like this wasn't tested at all 👀 Bonus points if you add a test that reads and checks those values.

It's unused as of now, it will become used as part of #3590, where it is also tested. It was spit out for easier review.

andybalaam:
> Looks like this wasn't tested at all 👀 Bonus points if you add a test that reads and checks those values.

Thank you for your diligence and ordinarily I would enthusiastically agree but see @poljar 's answer :-)

It is unfortunate that this change is so big that I needed to break it up in annoying ways. Maybe we could have found a better way to break it up that formed individual functional changes, but I couldn't find one when I tried.

Issue #3667, opened 2024-07-09
Expose Message `ShieldState` in UI crate for `TimelineEventItemId` and in ffi Bindings


BillCarsonFr:
Expose API in order to get the shield of a timeline item if needed.

https://github.com/matrix-org/matrix-rust-sdk/blob/02dddb47c9bd448fee6dc3d6877b321035de029f/crates/matrix-sdk-common/src/deserialized_responses.rs#L177-L189

Should be available for [EventTimelineItem](https://github.com/matrix-org/matrix-rust-sdk/blob/02dddb47c9bd448fee6dc3d6877b321035de029f/crates/matrix-sdk-ui/src/timeline/event_item/mod.rs#L87)

PR #3679, opened 2024-07-10
feat(timeline) - Expose shield state for EventTimelineItem

BillCarsonFr:
Fixes https://github.com/matrix-org/matrix-rust-sdk/issues/3667

Signed-off-by:

### DIFF
--- a/bindings/matrix-sdk-ffi/src/timeline/mod.rs
+++ b/bindings/matrix-sdk-ffi/src/timeline/mod.rs
@@ -24,6 +24,7 @@ use matrix_sdk::{
         AttachmentConfig, AttachmentInfo, BaseAudioInfo, BaseFileInfo, BaseImageInfo,
         BaseThumbnailInfo, BaseVideoInfo, Thumbnail,
     },
+    deserialized_responses::ShieldState as RustShieldState,
     Error,
 };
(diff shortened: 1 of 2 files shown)


bindings/matrix-sdk-ffi/src/timeline/mod.rs
```diff
@@ -1008,6 +1033,10 @@ impl EventTimelineItem {
     pub fn can_be_replied_to(&self) -> bool {
         self.0.can_be_replied_to()
     }
+
```
poljar, line 1036: Please add docs, I know it's annoying to duplicate them but the docs now are visible in Kotlin/Swift land so they do have a use.

crates/matrix-sdk-ui/src/timeline/event_item/mod.rs
```diff
@@ -335,6 +338,18 @@ impl EventTimelineItem {
         }
     }
 
+    /// Gets the `ShieldState` which can be used to decorate messages in the
```
poljar, line 341: Auto-links are nice and cheap.

```suggestion
    /// Gets the [`ShieldState`] which can be used to decorate messages in the
```

poljar (changes requested):

poljar (approved):

PR #3680, opened 2024-07-10
feat(base): New `RoomInfoNotableUpdateReasons::READ_RECEIPT`

Hywan:
Please review this PR patch-by-patch.

tl;dr: this PR adds the `READ_RECEIPT` reason for the room info notable update, in order to refresh the room list for this particular reason. This PR also adds a missing test for the `LATEST_EVENT` reason (it was tested in `matrix-sdk-ui` but it's better to also unit test it in `matrix-sdk-base` directly).

### DIFF
--- a/crates/matrix-sdk-base/src/sliding_sync.rs
+++ b/crates/matrix-sdk-base/src/sliding_sync.rs
@@ -1866,7 +1866,8 @@ mod tests {
         let response = response_with_room(room_id, room);
         client.process_sliding_sync(&response, &()).await.expect("Failed to process sync");
 
-        // Then a room info notable update is NOT received, because it's the first time the room is seen.
+        // Then a room info notable update is NOT received, because it's the first time
+        // the room is seen.
         assert_matches!(
             room_info_notable_update_stream.recv().await,
             Ok(RoomInfoNotableUpdate { room_id: received_room_id, reasons: received_reasons }) => {
@@ -1913,7 +1914,8 @@ mod tests {
             }
         );
 
-        // When I send sliding sync response containing a couple of events with no read receipt.
+        // When I send sliding sync response containing a couple of events with no read
+        // receipt.
         let room_id = room_id!("!r:e.uk");
         let events = vec![
             make_raw_event("m.room.message", "$3"),
--- a/crates/matrix-sdk-base/src/sliding_sync.rs
+++ b/crates/matrix-sdk-base/src/sliding_sync.rs
@@ -275,6 +275,11 @@ impl BaseClient {
                 );
 
                 if prev_read_receipts != room_info.read_receipts {
+                    room_info_notable_updates
+                        .entry(room_id.clone())
+                        .or_default()
+                        .insert(RoomInfoNotableUpdateReasons::READ_RECEIPT);
+
                     changes.add_room(room_info);
                 }
             }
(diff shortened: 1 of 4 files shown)


crates/matrix-sdk-base/src/sliding_sync.rs
```diff
@@ -1847,6 +1847,46 @@ mod tests {
...
+        });
+        let response = response_with_room(room_id, room);
+        client.process_sliding_sync(&response, &()).await.expect("Failed to process sync");
+
+        // Then a room info notable update is NOT received, because it's the first time the room is seen.
+        assert_matches!(
+            room_info_notable_update_stream.recv().await,
+            Ok(RoomInfoNotableUpdate { room_id: received_room_id, reasons: received_reasons }) => {
+                assert_eq!(received_room_id, room_id);
+                assert!(!received_reasons.contains(RoomInfoNotableUpdateReasons::RECENCY_TIMESTAMP));
```
bnjbvr, line 1875: Does this pass?

```
assert!(received_reasons.is_empty());
```

crates/matrix-sdk-base/src/sliding_sync.rs
```diff
@@ -1887,6 +1892,51 @@ mod tests {
...
+        let room = v4::SlidingSyncRoom::new();
+        let response = response_with_room(room_id, room);
+        client.process_sliding_sync(&response, &()).await.expect("Failed to process sync");
+
+        // Then a room info notable update is NOT received.
+        assert_matches!(
+            room_info_notable_update_stream.recv().await,
+            Ok(RoomInfoNotableUpdate { room_id: received_room_id, reasons: received_reasons }) => {
+                assert_eq!(received_room_id, room_id);
+                dbg!(&received_reasons);
```
bnjbvr: nit: remove dbg!() statement

crates/matrix-sdk-base/src/sliding_sync.rs
```diff
@@ -1887,6 +1892,51 @@ mod tests {
...
+                assert!(!received_reasons.contains(RoomInfoNotableUpdateReasons::READ_RECEIPT));
```
bnjbvr, line 1913: Same as before, can this be a check that reasons is empty?

crates/matrix-sdk-base/src/sliding_sync.rs
```diff
@@ -1887,6 +1892,51 @@ mod tests {
...
+            }
+        );
+
+        // When I send sliding sync response containing a couple of events with no read receipt.
+        let room_id = room_id!("!r:e.uk");
+        let events = vec![
+            make_raw_event("m.room.message", "$3"),
```
bnjbvr, line 1921: `make_raw_event` and `make_event` should be replaced with the `EventFactory` :innocent:

crates/matrix-sdk-base/src/sliding_sync.rs
```diff
@@ -1887,6 +1892,51 @@ mod tests {
...
+            make_raw_event("m.room.message", "$4"),
+            make_raw_event("m.read", "$5"),
```
bnjbvr, line 1923: Technically this is wrong, as this will create a `m.read` event with the content of a `m.room.message` event; it seems `compute_unread_counts()` still does something with it (store it, maybe?), but this could be a bit better…

crates/matrix-sdk-base/src/sliding_sync.rs
bnjbvr, line 1: Please fold 0e73620027e5348747ae7fe4c69bcdcb2da2998e into the commits it belongs to (consider git-absorb next time!)

bnjbvr (approved):
Thanks!

crates/matrix-sdk-base/src/sliding_sync.rs
Hywan: I prefer to test the reason is _not_ present, in case other reasons may exist. We don't care about the other reasons at this step. Tests are less likely to break in the future if these kind of assertions.

crates/matrix-sdk-base/src/sliding_sync.rs
Hywan: This entire test suite should be revisited. Later. For another time. But I agree!

crates/matrix-sdk-base/src/sliding_sync.rs
Hywan: There is no `RoomReadReceipt` before, and now there is one, so the previous and the new one are different, which triggers something. Do you have a better test in mind?

bnjbvr: All I'm saying is that the `m.read` event should be created in a better way (with a proper `content` property), because it doesn't make sense now :)

Hywan: Since you've approved the PR, I understand it's fine for this particular test. The read receipt mechanism is already pretty well test. Here we only want to test the notable update reason is triggered, which is fine for the moment.

PR #3681, opened 2024-07-10
timeline: move `TimelineEventContext::encryption_info` to the remote flow

bnjbvr:
Since it's only used for remote events, and a local echo couldn't figure that out anyways (since it's not sent yet, and that information makes sense after sending).


andybalaam (approved):
Looks sensible to me.

Issue #3663, opened 2024-07-08
send queue: dependent actions system


frebib:
Under less-than-optimal network situations, EX fails to send reactions and "eats" them. It'd be cool if the send queue (+retry mechanisms) was used for reaction sending. Benji suggests that it would probably improve in less duplicate code https://matrix.to/#/!iYnZafYUoXkeVPOSQh:matrix.org/$hGjNwhDvoNeK15O6h3DhmO7wlVrzNsdDBBssJ9o4iYQ?via=matrix.org&via=element.io&via=envs.net

bnjbvr:
Ok, so I'm stealing this issue because I've found about one extra use case that is quite similar and would be handy.

Right now, there's a possible race condition in the timeline: if you try to edit/abort an event that's being sent, then the send queue refuses to edit/abort it (because it's being sent), and the timeline will assume it's *been* sent, so will look for a corresponding event that has a matching transaction id. In that case, it results into a no-op, because there's no remote echo for that event yet.

Since applying a reaction requires the event to be sent, an extension of that is to consider edits/aborts as pending operations too, and have the send queue remember them. So the send queue would now allow edit/abort items that are *being* sent, and would apply pending operations later:

- if the event could be successfully sent, then send a redaction/edit event later
- otherwise, apply the redaction/edit on the local echo itself

This would work also for reactions, which would only take the success branch and send an event after the event's been sent.

Now, when editing something in the timeline:
- either it's a local echo at the time the edit operation is done;
  - if the send queue still knows about the local echo (it was being sent, or not): then the send queue either uses a local echo edit, or a pending operation if it's being sent
- or the local echo isn't known by the timeline anymore: the timeline tries to find a matching remote event_id for that transaction id

I think that will cover all the cases, and avoid all race conditions.

As a result, there's a bit of code duplication that needs to be handled: editing/aborting could happen either at the timeline level (second top-level bullet in the above list), or at the send queue level (first top-level bullet).

Some high-level items needed for this work:

- [ ] migration for a new table for the pending operations in the state storage
- [ ] new APIs to insert/edit/remove in that pending operations table
- [ ] adapt the send queue to make use of this
- [ ] extract the edit code from the timeline and put it in the SDK, so it can be reused everywhere
- [ ] (performance bonus) have the event cache provide an API to retrieve a known event by ID, to avoid events network lookups in the above `edit()` code

frebib:
🎉  Thanks 😍

PR #3682, opened 2024-07-10
event cache: add a way to get an event by ID

bnjbvr:
"dude… what if the event cache… allowed to retrieve events by id… i know, it's crazy right…"

This is going to be useful for #3663, for edits we'll want to retrieve the original event by id, to see if it was a reply, and some other things like this.

Part of #3058 too.

### DIFF
--- a/crates/matrix-sdk/src/event_cache/mod.rs
+++ b/crates/matrix-sdk/src/event_cache/mod.rs
@@ -56,7 +56,7 @@ use matrix_sdk_common::executor::{spawn, JoinHandle};
 use ruma::{
     events::{AnyRoomAccountDataEvent, AnySyncEphemeralRoomEvent},
     serde::Raw,
-    OwnedEventId, OwnedRoomId, RoomId,
+    EventId, OwnedEventId, OwnedRoomId, RoomId,
 };
 use tokio::sync::{
     broadcast::{error::RecvError, Receiver, Sender},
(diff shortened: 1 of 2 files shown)


crates/matrix-sdk/src/event_cache/mod.rs
```diff
@@ -196,6 +196,24 @@ impl EventCache {
         Ok(())
     }
 
+    /// Try to find an event by id in all the rooms.
```
poljar, line 199: Perhaps we should clarify by which ID, if you don't like this form perhaps spelling out the term "event ID" would be better?

```suggestion
    /// Try to find an event by its ID in all the rooms.
```

crates/matrix-sdk/src/event_cache/mod.rs
```diff
@@ -196,6 +196,24 @@ impl EventCache {
...
+    ///
+    /// Note: this does a linear scan, so it could be slow. If performance
+    /// requires it, using a direct mapping of event id -> event might be
```
poljar, line 202: Does this tell the user that they should create such a mapping or that we, the developers, will provide it if this method turns out to be slow?

poljar (approved):

crates/matrix-sdk/src/event_cache/mod.rs
bnjbvr: You're right, it's unclear. We should do that, if needs be; i'll convert this into a code non-doc comment.

PR #3684, opened 2024-07-10
clippy: Enforce the `redundant_clone` lint rule

### LINKED ISSUE #3683: code base: enforce the `redundant_clone` Clippy lint
This would catch a few useless `clone()` calls in the code base.

To work on this:

- add `-Wclippy::redundant_clone` to [this list](https://github.com/matrix-org/matrix-rust-sdk/blob/9e9df163b903e2a5ec415b1a708a85cf64e48d4b/.cargo/config.toml#L41).
- compile with Clippy, using `cargo xtask ci clippy` and fix all the issues it finds.

almet:
Fixes #3683

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -39,6 +39,7 @@ rustflags = [
     "-Wclippy::str_to_string",
     "-Wclippy::todo",
     "-Wclippy::unused_async",
+    "-Wclippy::redundant_clone",
 ]
 
 [target.'cfg(target_arch = "wasm32")']
--- a/bindings/matrix-sdk-crypto-ffi/src/lib.rs
+++ b/bindings/matrix-sdk-crypto-ffi/src/lib.rs
@@ -434,7 +434,7 @@ fn collect_sessions(
     // the session cache after migration) so we don't need to worry about
     // signatures.
     let device_keys = DeviceKeys::new(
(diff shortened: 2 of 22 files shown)


almet:
I don't understand why clippy is failing on the CI, especially as the lines it mention doesn't exist locally for me, and I don't see the same errors when running the command.

Any idea why that is?

bnjbvr:
It looks like the first attempt at this CI task checked out the wrong commit, which is surprising; I retried and now we're running into caching issues. Investigating with other repo maintainers, I wonder if we could empty the caches or cause a new run on a new commit (folding the fixup commit into the other one and force-pushing might do the trick!).

almet:
> (folding the fixup commit into the other one and force-pushing might do the trick!).

Just did, let's see :-)

bnjbvr:
Hmm, it didn't work. Can you try rebasing over main, please? (this might introduce new redundant clones, if new PRs introduced some)

bnjbvr:
Ah, turns out I can now repro the local Clippy failures when running with `cargo xtask ci clippy`, so you should be able to as well.

almet:
Thanks for your support. We should be good to go.

bnjbvr (approved):
Thanks very much :partying_face: And congrats on your first PR!

PR #3685, opened 2024-07-10
sdk: provide a way to create edit events

bnjbvr:
Part of my quest to move functionality from the `Timeline` to the SDK, this makes it possible to create a replacement event for another one (edit), for all the `m.room.message`s to start with. The timeline can now use this, instead of trying to look into its items for the data it needs for the edit to work correctly; this helps get rid of `EditInfo`, which was an oversight from yours truly.

Part of #3663 too.

### DIFF
--- a/crates/matrix-sdk-ui/src/timeline/mod.rs
+++ b/crates/matrix-sdk-ui/src/timeline/mod.rs
@@ -26,7 +26,7 @@ use matrix_sdk::{
     event_cache::{EventCacheDropHandles, RoomEventCache},
     event_handler::EventHandlerHandle,
     executor::JoinHandle,
-    room::{Receipts, Room},
+    room::{edit::EditedContent, Receipts, Room},
     send_queue::{RoomSendQueueError, SendHandle},
     Client, Result,
 };
@@ -45,8 +45,8 @@ use ruma::{
         relation::Annotation,
         room::{
             message::{
-                AddMentions, ForwardThread, OriginalRoomMessageEvent, ReplacementMetadata,
-                RoomMessageEventContent, RoomMessageEventContentWithoutRelation,
+                AddMentions, ForwardThread, OriginalRoomMessageEvent, RoomMessageEventContent,
+                RoomMessageEventContentWithoutRelation,
             },
             redaction::RoomRedactionEventContent,
         },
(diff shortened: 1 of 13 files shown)


crates/matrix-sdk/src/room/edit.rs
```diff
@@ -0,0 +1,187 @@
...
+        /// The type of the target event.
+        target: &'static str,
+        /// The type of the new content.
+        new_content: &'static str,
+    },
+
+    /// While it's technically possible to edit this kind of events, this is not
+    /// yet supported by the SDK.
+    #[error("Edits for events of type {0} aren't supported yet")]
+    UnsupportedEventType(String),
```
zecakeh, line 87: It seems like this error should already be catched by the `IncompatibleEditType` variant since all supported types have a variant in `EditedContent`.

bnjbvr: Good point! By rejiggering the order of these match statements, we could probably spare one error type indeed :+1:

crates/matrix-sdk/src/room/edit.rs
```diff
@@ -0,0 +1,385 @@
...
+    async fn test_edit_state_event() {
+        let event_id = event_id!("$1");
+        let own_user_id = user_id!("@me:saucisse.bzh");
+
+        let mut cache = TestEventCache::default();
+
+        cache.events.insert(
+            event_id.to_owned(),
+            // TODO: use the EventFactory for state events too.
+            SyncTimelineEvent::new(
```
Hywan, line 239: Do you want to address this `TODO` now?

crates/matrix-sdk/src/room/edit.rs
```diff
@@ -0,0 +1,385 @@
...
+                        "origin_server_ts": 1,
+                        "type": "m.room.name",
+                    })
+                    .to_string(),
+                )
+                .unwrap(),
+            ),
+        );
+
+        let room_id = room_id!("!galette:saucisse.bzh");
```
Hywan, line 257: 🏴󠁦󠁲󠁢󠁲󠁥󠁿

crates/matrix-sdk-ui/src/timeline/mod.rs
```diff
@@ -106,27 +106,6 @@ use self::{
     util::{rfind_event_by_id, rfind_event_item},
 };
 
-/// Information needed to edit an event.
-#[derive(Debug, Clone)]
-pub struct EditInfo {
```
Hywan, line 111: Hurray!

Hywan (approved):
I've enjoyed reading this PR. The commits are clear, and they all make sense. I'm happy to say _farewell_ to `EditInfo`. The code is now much cleaner. Thanks!

crates/matrix-sdk/src/room/edit.rs
bnjbvr: No, that would be some future fun :)

PR #3690, opened 2024-07-12
crypto: Rename the ReadOnlyFoo types

poljar:
A bit of renaming, these types were a sore to the eyes for quite some time.

I added a snapshot of a crypto database and added a test, to double check that we're not introducing a regression here, but we shouldn't anyways since the type name isn't used to serialize things the field name is.

I also renamed one field name, but added a serde alias declaration with the old name, so all should be fine™.


andybalaam (approved):
Much better, thanks!

PR #3691, opened 2024-07-12
test(integration-testing): Remove a flaky and useless test

Hywan:
This patch removes the `test_room_info_notable_update_deduplication` test. First off, it's flaky because sometimes Synapse lags, or sends another events, which makes the test to fail. Second, the same feature is tested inside the `matrix_sdk_ui::room_list_service` test suite, with `test_room_sorting` and `test_room_latest_event`, and inside the `matrix_sdk_base::sliding_sync` test suite, with a better granularity. And lastly, this test doesn't test what it says: there is no room info notable update deduplication whatsoever. I personally don't believe it has ever existed. This test isn't necessary.


poljar (approved):
Alright, I believe you, the coverage doesn't seem to be impacted either.

bnjbvr:
The test was incorrectly written and spawning multiple sync loops per user, and maybe that's why you thought we had duplicate events. We should reinstate and rewrite this test to perform correctly, it's testing something useful.

PR #3692, opened 2024-07-13
base-sdk: Properly update direct targets of rooms with m.direct event

zecakeh:
The code used to only add new targets to rooms but never remove the ones that are not in the event anymore.

### DIFF
--- a/crates/matrix-sdk-base/src/client.rs
+++ b/crates/matrix-sdk-base/src/client.rs
@@ -14,7 +14,7 @@
 // limitations under the License.
 
 use std::{
-    collections::{BTreeMap, BTreeSet},
+    collections::{BTreeMap, BTreeSet, HashMap, HashSet},
     fmt, iter,
 };
 #[cfg(feature = "e2e-encryption")]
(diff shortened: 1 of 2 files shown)


crates/matrix-sdk-base/src/client.rs
```diff
@@ -643,22 +643,55 @@ impl BaseClient {
...
-                        if let Some(room) = changes.room_infos.get_mut(room_id) {
-                            room.base_info.dm_targets.insert(user_id.clone());
-                        } else if let Some(room) = self.store.room(room_id) {
-                            let mut info = room.clone_info();
-                            if info.base_info.dm_targets.insert(user_id.clone()) {
-                                changes.add_room(info);
-                            }
+                let rooms = self.store.rooms();
+                let mut old_dms = rooms
+                    .iter()
```
Hywan, line 655: I feel like for such iterations over rooms, we need a `Store::rooms_iter()` API. Otherwise it's clearly a waste of resources. I am doing it now.

Hywan: Hmm, it's harder than expected to do that wit rustc stable. Not a blocker here.

zecakeh:

poljar (approved):
Makes sense to me. Thanks for providing a test and sorry for taking this long people went on holidays.

PR #3696, opened 2024-07-15
test(sdk): Rewrite `test_delayed_decryption_latest_event`

Hywan:
This patch rewrites the `test_delayed_decryption_latest_event` test a little bit. It does exactly the same things, but in a simpler way: it removes multiple `sleep` and remove 2 sliding sync loops.

First off, the `SyncService` already starts the `RoomListService` and the
 `EncryptionSync` service. Both of them have their own sliding sync
loop. The test doesn't need other sliding sync loops in their own tasks, this is not necessary at all: it's just pretty confusing and doesn't reflect the reality, i.e. how these API are supposed to be used.

Second, it also tests the room for Bob is seen as encrypted.

Third, the `VectorDiff::Reset` is tested before the event from Bob is sent. It's not only for clarity: it makes the test more robust for future modifications.

Fourth, instead of waiting with a `sleep` for the event from Bob to be received by Alice, we instead wait on the room list's stream of Alice to receive an update. It's more robust this way and reflects the real usage of this API. It also helps to remove an intermediate `assert_pending!` that is no longer necessary because we are waiting on the stream just after.

Finally, just like for the previous modification, this patch removes another `sleep` for the to-device event from Bob to be received by Alice, and instead wait on the room list's stream to receive an update. It's again more robust and reflects the real usage of this API. Plus, it makes the last `assert_pending!` macro to not be flaky.

### DIFF
--- a/testing/matrix-sdk-integration-testing/src/tests/sliding_sync/room.rs
+++ b/testing/matrix-sdk-integration-testing/src/tests/sliding_sync/room.rs
@@ -740,15 +740,40 @@ async fn test_delayed_decryption_latest_event() -> Result<()> {
     // Room is created by Alice. Let's enable encryption.
     room.enable_encryption().await.unwrap();
 
-    // Wait for Alice to see its new room, and for Bob to receive the invitation.
-    sleep(Duration::from_secs(2)).await;
+    // Wait for Alice to see its new room…
+    let mut i = 1;
+    let alice_room = loop {
+        // Linear backoff: wait at most (1+2+3+4)*200ms = 2s for the response to come
+        // back from the server.
+        if let Some(room) = alice.get_room(room.room_id()) {
+            break room;
+        };
+        if i == 4 {
+            panic!("alice can't see the room");
+        }
+        warn!("attempt #{i}, alice couldn't see the room, retrying in a few…");
+        i += 1;
+        sleep(Duration::from_millis(i * 200)).await;
+    };
 
-    let alice_room = alice.get_room(room.room_id()).unwrap();
-    let bob_room = bob.get_room(alice_room.room_id()).unwrap();
+    // …and for Bob to receive the invitation.
+    let mut i = 1;
+    let bob_room = loop {
+        // Linear backoff: wait at most (1+2+3+4)*200ms = 2s for the response to come
+        // back from the server.
+        if let Some(room) = bob.get_room(room.room_id()) {
+            break room;
+        };
+        if i == 4 {
+            panic!("boob can't see the room");
+        }
+        warn!("attempt #{i}, bob couldn't see the room, retrying in a few…");
+        i += 1;
+        sleep(Duration::from_millis(i * 200)).await;
+    };
+
+    // Bob accepts the invitation by joining the room.
     bob_room.join().await.unwrap();
-
-    // Wait for Alice to see Bob in the room.
-    sleep(Duration::from_secs(2)).await;
 
     assert_eq!(alice_room.state(), RoomState::Joined);
     assert!(alice_room.is_encrypted().await.unwrap());
@@ -764,7 +789,9 @@ async fn test_delayed_decryption_latest_event() -> Result<()> {
     pin_mut!(alice_room_list_stream);
 
     // The room list receives its initial rooms.
-    assert_let!(Some(diffs) = alice_room_list_stream.next().await);
+    assert_let!(
+        Ok(Some(diffs)) = timeout(Duration::from_secs(5), alice_room_list_stream.next()).await
+    );
     assert_eq!(diffs.len(), 1);
     assert_matches!(
         &diffs[0],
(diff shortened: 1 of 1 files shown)


testing/matrix-sdk-integration-testing/src/tests/sliding_sync/room.rs
```diff
@@ -724,119 +724,85 @@ async fn test_delayed_decryption_latest_event() -> Result<()> {
...
 
-    sleep(Duration::from_secs(1)).await;
-    let alice_room = alice.get_room(alice_room.room_id()).unwrap();
+    // Room is created by Alice. Let's enable encryption.
+    room.enable_encryption().await.unwrap();
+
+    // Wait for Alice to see its new room, and for Bob to receive the invitation.
+    sleep(Duration::from_secs(2)).await;
+
+    let alice_room = alice.get_room(room.room_id()).unwrap();
```
bnjbvr, line 746: Could be better to do a few rounds of waiting with exponential backoff here

testing/matrix-sdk-integration-testing/src/tests/sliding_sync/room.rs
```diff
@@ -724,119 +724,85 @@ async fn test_delayed_decryption_latest_event() -> Result<()> {
...
     let bob_room = bob.get_room(alice_room.room_id()).unwrap();
```
bnjbvr, line 747: ditto

testing/matrix-sdk-integration-testing/src/tests/sliding_sync/room.rs
```diff
@@ -724,119 +724,85 @@ async fn test_delayed_decryption_latest_event() -> Result<()> {
...
     bob_room.join().await.unwrap();
 
-    sleep(Duration::from_secs(1)).await;
+    // Wait for Alice to see Bob in the room.
```
bnjbvr, line 750: comment is incorrect, we're not checking that alice "sees" bob in the room

bnjbvr (approved):
I'm taking over this PR to make sure we don't have this intermittent failure anymore.

testing/matrix-sdk-integration-testing/src/tests/sliding_sync/room.rs
```diff
@@ -724,119 +724,85 @@ async fn test_delayed_decryption_latest_event() -> Result<()> {
...
     let alice_all_rooms = alice_sync_service.room_list_service().all_rooms().await.unwrap();
-    let (stream, entries) = alice_all_rooms
+    let (alice_room_list_stream, entries) = alice_all_rooms
         .entries_with_dynamic_adapters(10, alice.room_info_notable_update_receiver());
     entries.set_filter(Box::new(new_filter_all(vec![])));
-    pin_mut!(stream);
+    pin_mut!(alice_room_list_stream);
+
+    // The room list receives its initial rooms.
+    assert_let!(Some(diffs) = alice_room_list_stream.next().await);
```
bnjbvr, line 767: general nit: prefer adding a timeout do those waits, to make the tests fail fast in real-world conditions

testing/matrix-sdk-integration-testing/src/tests/sliding_sync/room.rs
```diff
@@ -724,119 +724,85 @@ async fn test_delayed_decryption_latest_event() -> Result<()> {
...
     let event = bob_room.send(RoomMessageEventContent::text_plain("hello world")).await?;
 
-    // Wait shortly so the manual roominfo update is triggered before we load the
-    // stream.
-    sleep(Duration::from_secs(1)).await;
-
-    // Stream only has the initial Reset entry.
-    assert_let!(Some(diffs) = stream.next().await);
+    // Wait the message to be received by Alice.
+    assert_let!(Some(diffs) = alice_room_list_stream.next().await);
```
bnjbvr, line 782: here too

testing/matrix-sdk-integration-testing/src/tests/sliding_sync/room.rs
```diff
@@ -724,119 +724,85 @@ async fn test_delayed_decryption_latest_event() -> Result<()> {
...
 
-    // Wait for next sync
-    sleep(Duration::from_secs(3)).await;
-
-    // Latest event is set now
-    assert_eq!(alice_room.latest_event().unwrap().event_id(), Some(event.event_id));
-
-    // The stream has a single update
-    assert_let!(Some(diffs) = stream.next().await);
+    assert_let!(Some(diffs) = alice_room_list_stream.next().await);
```
bnjbvr, line 795: and here too

testing/matrix-sdk-integration-testing/src/tests/sliding_sync/room.rs
```diff
@@ -724,119 +724,85 @@ async fn test_delayed_decryption_latest_event() -> Result<()> {
...
         &diffs[0],
         VectorDiff::Set { index: 0, value: room } => {
-            assert_eq!(room.room_id(), alice_room.room_id());
+            // The latest event is now decrypted!
+            assert_eq!(room.latest_event().await.unwrap().event_id().unwrap(), event.event_id);
         }
     );
 
-    assert_pending!(stream);
+    assert_pending!(alice_room_list_stream);
```
bnjbvr, line 805: add a small wait before this assert_pending, just in case the final stream update is delayed.

Issue #3544, opened 2024-06-13
Store sender data for Megolm sessions as per Invisible crypto


andybalaam:
Part of [Invisible Crypto](https://github.com/matrix-org/matrix-spec-proposals/pull/4153).

Currently when we store an Inbound group session, we associate some [SessionCreatorInfo](https://github.com/matrix-org/matrix-rust-sdk/blob/1f524f2decdb83ac88c392eb4af1957b3127b165/crates/matrix-sdk-crypto/src/olm/group_sessions/inbound.rs#L63-L80) to it. Currently this info mainly contains the `curve25519` identity key of the session creator.

We need now to add more info:
- `user_id` - the associated mxid of the session creator (rooted in Master Signing Key authenticity).
- `master_key` - the Master Signing Key signing (via the SSK) this device at time of reception.
- `master_key_verified` - true if at time of reception the user was verified (see [here](https://github.com/element-hq/crypto-internal/issues/301))

We are currently computing the "authenticity" of a megolm session but we are doing it at decryption time.
The info we get at decryption time are `sender: OwnedUserId` and `sender_device: OwnedDeviceId`.
Both related to a `VerificationState`. This owner info is untrusted (claimed), unless the VerificationState is Trusted.

The VerificationState variants are:
- Verified
- Unverified::UnverifiedIdentity
- Unverified::UnsignedDevice
- Unverified::MissingDevice
- Unverified::InsecureSource

We are basically trying to get this same info but at reception time.

## Three places we populate SenderData in an InboundGroupSession

1. When we receive the room keys via a to-device message. (This is already done via SenderDataFinder.)

2. When we get new or updated device info from /keys/query. To do this we need to be able to look up InboundGroupSessions that don't have SenderData by device curve key. This will require a new index on the inboundgroupsessions table. Question: if there are lots of sessions for a particular device, will we break everything by working through batches of them to update them?

3. When we decrypt a message for a session. In this case we have the session and device already, so it's just a case of persisting (into inboundgroupsessions) the VerificationState that we already look up at this moment.

## The plan

In order to complete this we need to do these tasks:

* https://github.com/matrix-org/matrix-rust-sdk/issues/3542
* https://github.com/matrix-org/matrix-rust-sdk/issues/3543
* https://github.com/matrix-org/matrix-rust-sdk/issues/3751
* Populate missing SenderData when we receive device info:
    * https://github.com/matrix-org/matrix-rust-sdk/issues/3752
[...]

PR #3701, opened 2024-07-15
Configurable sender trust checking on decrypting

uhoreg:

Signed-off-by:


crates/matrix-sdk-crypto/src/olm/group_sessions/inbound.rs
```diff
@@ -55,6 +55,26 @@ use crate::{
...
+    CrossSignedOrLegacy,
+    /// Only decrypt events from cross-signed devices
+    CrossSigned,
+    /// Only decrypt events from cross-signed devices where we have verified the user
+    VerifiedUser,
+}
+
+/// Settings for decrypting messages
+#[derive(Clone, Debug, Deserialize, Serialize)]
+pub struct DecryptionSettings {
```
uhoreg, line 73: I initially put this in this file for symmetry with `EncryptionSettings`, which is in `olm/group_sessions/outbound.rs`, but it only gets used in `machine.rs`.  So I suspect that I should put this in `machine.rs` instead?  Or is there a better place to put it?

uhoreg:
I'm not sure how to handle the higher levels.  It seems like we should be storing the setting in some struct so that, e.g. messages that come down /sync get decrypted with the right setting, but I'm not sure where it should go.  I don't see a relevant struct that has settings, so there isn't a clear place to put the setting.

jmartinesp:
Sorry, I don't have much experience with the SDK and much less with the crypto part of it. Maybe someone else from the team can answer your questions? @poljar , @andybalaam?

crates/matrix-sdk-crypto/src/olm/group_sessions/inbound.rs
```diff
@@ -55,6 +55,26 @@ use crate::{
     },
 };
 
+/// The trust level required to decrypt an event
+#[derive(Clone, Copy, Debug, Deserialize, Serialize)]
+pub enum TrustRequirement {
```
BillCarsonFr, line 60: I expected 2 options here? 
- Legacy: Like it's doing now
- InvisibleCrypto: The new mode

And a 3rd temporary one for Invisible transition mode?

What do you think, can we find a way to make the encryption mode a bit symetrical to the decryption mode?

uhoreg: I just implemented what was given by the "Quarantine" section of https://github.com/matrix-org/matrix-rust-sdk/issues/3544 But I think that "Legacy" would be "Untrusted", and "InvisibleCrypto" would be "CrossSigned" or "CrossSignedOrLegacy". transition would be "CrossSignedOrLegacy".

andybalaam: Sorry, I moved that section into a comment here: https://github.com/matrix-org/matrix-rust-sdk/issues/3544#issuecomment-2245005736

crates/matrix-sdk-crypto/src/olm/group_sessions/inbound.rs
poljar: Probably into the `types/` submodule.

crates/matrix-sdk-crypto/src/olm/group_sessions/inbound.rs
```diff
@@ -55,6 +55,26 @@ use crate::{
...
+    /// Decrypt events from everyone regardless of trust
+    Untrusted,
+    /// Only decrypt events from cross-signed or legacy devices
```
poljar, line 63: People won't know what legacy means, myself included :sweat_smile:. Could you define that.

crates/matrix-sdk-crypto/src/olm/group_sessions/inbound.rs
```diff
@@ -55,6 +55,26 @@ use crate::{
     },
 };
 
+/// The trust level required to decrypt an event
+#[derive(Clone, Copy, Debug, Deserialize, Serialize)]
+pub enum TrustRequirement {
+    /// Decrypt events from everyone regardless of trust
```
poljar, line 61:
```suggestion
    /// Decrypt events from everyone regardless of trust.
```

Same for the rest.

crates/matrix-sdk-crypto/src/olm/group_sessions/inbound.rs
```diff
@@ -55,6 +55,26 @@ use crate::{
...
+    Untrusted,
+    /// Only decrypt events from cross-signed or legacy devices
+    CrossSignedOrLegacy,
+    /// Only decrypt events from cross-signed devices
+    CrossSigned,
+    /// Only decrypt events from cross-signed devices where we have verified the user
+    VerifiedUser,
```
poljar, line 68: Not sure, but perhaps it makes sense to call it user identity instead of user, since it might confuse people that those are the same concept.
```suggestion
    VerifiedUserIdentity,
```

crates/matrix-sdk-crypto/src/error.rs
```diff
@@ -141,6 +141,19 @@ pub enum MegolmError {
     /// The storage layer returned an error.
     #[error(transparent)]
     Store(#[from] CryptoStoreError),
+
```
poljar, line 144: It really feels like we should create a separate and more specific error type instead of adding ever-more variants here. Not needed for this PR but please put something on the board.

crates/matrix-sdk-crypto/src/olm/group_sessions/inbound.rs
poljar: What is the result of this conversation? Is this resolved or do we need to remove some variants?

crates/matrix-sdk-crypto/src/machine.rs
```diff
@@ -1552,15 +1548,86 @@ impl OlmMachine {
...
         &self,
         room_id: &RoomId,
         event: &EncryptedEvent,
         content: &SupportedEventEncryptionSchemes<'_>,
+        decryption_settings: &DecryptionSettings,
     ) -> MegolmResult<(JsonObject, EncryptionInfo)> {
         let session =
             self.get_inbound_group_session_or_error(room_id, content.session_id()).await?;
 
+        self.check_sender_trust_requirement(&session, &event.sender, decryption_settings).await?;
```
poljar, line 1629: Shouldn't this use the `get_encryption_info()` method or be incorporated into it? This way we're now fetching the device twice, once for this method and another time for the `get_encryption_info()` method.

It's also fetching the `sender_data` twice, the `get_encryption_info()` method fetches and updates this piece of information as well, sure it's turning it into a `VerificationState` in the end but we can modify the place where we do the conversion.

crates/matrix-sdk-crypto/src/machine.rs
```diff
@@ -4947,4 +5066,349 @@ pub(crate) mod tests {
             .unwrap();
         assert_matches!(thread_encryption_result, UnsignedDecryptionResult::Decrypted(_));
     }
+
+    #[async_test]
+    async fn test_decryption_trust_requirement() {
+        let (alice, bob) =
```
poljar, line 5072: This is now one of those unfocused tests we mentioned yesterday, especially since this whole setup section is repeated in the test bellow.

Could you turn this into a helper that gets called by both tests, there are other common places you could turn into helpers. That would also make it easier to make smaller, more focused tests.

poljar (changes requested):
The direction makes sense for what it is trying to achieve but we need to incorporate it a bit better into the existing code.

crates/matrix-sdk-crypto/src/olm/group_sessions/inbound.rs
uhoreg: That sounds reasonable.  I don't really see an existing file in there that looks like a compelling place to put it.  Should I just put it in `types/mod.rs`, or create a new file, like `types/decryption.rs`?

poljar: A new file sounds fine.

crates/matrix-sdk-crypto/src/olm/group_sessions/inbound.rs
uhoreg: I've dropped the `VerifiedUserIdentity` option, so we now have three options, which correspond do legacy, invisible crypto, and transition

crates/matrix-sdk-crypto/src/error.rs
uhoreg: While I was investigating https://github.com/matrix-org/matrix-rust-sdk/pull/3701#discussion_r1701500823 I noticed that `matrix_sdk_common::deserialized_responses::VerificationLevel` maps pretty well to what we want here, except that it's missing a level, so I added the level and referenced it here.

crates/matrix-sdk-crypto/src/machine.rs
uhoreg: Done

crates/matrix-sdk-crypto/src/machine.rs
uhoreg: Added helpers.  It seems like a bunch of tests could be changed to use the `encrypt_message` helper, but I haven't searched to see which ones.  I can do that in this PR if you want.

crates/matrix-sdk-crypto/src/olm/group_sessions/sender_data.rs
```diff
@@ -74,6 +74,9 @@ pub enum SenderData {
         /// If false, we had simply accepted the key as this user's latest
         /// key.
         master_key_verified: bool,
+
+        #[serde(default)]
+        identity_needs_user_approval: bool,
```
uhoreg, line 79: I added this so that session_data_finder can find this information, so that the trust level can be calculated.  But I think that it would be better to combine this with the `master_key_verified` field, and make them into a new `master_key_status` field that is an enum with `Unverified`, `NeedsApproval`, and `Verified` options.  Maybe we should even add another option for the case where the user's previous identity was verified, but they rotated their identity.  Maybe something like `PreviousIdentityVerified`.

What do you think?

(If you think it's better to keep it as-is, then I'll write documentation for this field, but I didn't want to write it if I was going to remove it right away.)

uhoreg:
I believe the tests fail because `get_encryption_info` now updates the `sender_data`, so I'll need to poke more things into the database so that it updates to the right thing.  Aside from that, I think that this is ready for re-review.

richvdh:
Discussion: should we throw an error when an identity has been rotated?
 * For a TOFU user, we should definitely just decrypt as normal (for symmetry with encryption)
 * For a verified user, we will throw an error on decryption if the identity is replaced (an expected UTD). The application can try to re-decrypt after the user re-verifies/reverts to unverified.

richvdh:
@uhoreg will update this PR a bit and then get @richvdh to review

uhoreg:
@richvdh I think this is ready for review.  I do have [one question](https://github.com/matrix-org/matrix-rust-sdk/pull/3701#discussion_r1710536276) that I'd like an opinion on.

The commit history is a big mess, partially due to being based on a branch that got force-pushed at some point, so this PR should probably get squash-merged.

crates/matrix-sdk-crypto/src/olm/group_sessions/sender_data.rs
```diff
@@ -74,6 +74,14 @@ pub enum SenderData {
...
         /// key.
         master_key_verified: bool,
+
+        #[serde(default)]
+        /// Whether the user's master key was previously verified, but the
+        /// current master key is not.
+        ///
+        /// TODO: should this be merged with the master_key_verified field and
+        /// turned into an enum?
+        previously_verified: bool,
```
richvdh, line 84: hrm, this feels wrong. 

"Previously-verified-but-no-longer" is surely a property of the sender, not the session? Not least because all received messages are going to get stuck in that state even once we verify the new identity.

But perhaps that's behaviour we want?

crates/matrix-sdk-crypto/src/olm/group_sessions/sender_data.rs
```diff
@@ -74,6 +74,14 @@ pub enum SenderData {
         /// If false, we had simply accepted the key as this user's latest
         /// key.
         master_key_verified: bool,
+
+        #[serde(default)]
+        /// Whether the user's master key was previously verified, but the
+        /// current master key is not.
+        ///
+        /// TODO: should this be merged with the master_key_verified field and
+        /// turned into an enum?
```
richvdh, line 83: Short answer: yes, this really feels like it ought to be a tri-state.

(https://github.com/matrix-org/matrix-rust-sdk/pull/3846 added an `OwnUserIdentityVerifiedState` enum for a similar reason; not sure we should reuse that, but I think it's a pattern we should follow)

A downside is that it breaks backwards-compatibility for the crypto store :/

crates/matrix-sdk-crypto/src/machine/mod.rs
```diff
@@ -1570,11 +1567,69 @@ impl OlmMachine {
         self.get_encryption_info(&session, &event.sender).await
     }
 
+    /// Check whether the sender of a Megolm session is trusted, based on the
+    /// verification state.
+    ///
+    /// This is used by `check_sender_trust_requirement`, and ensures that the
+    /// sending device is cross-signed.
+    fn check_sender_trusted(&self, encryption_info: &EncryptionInfo) -> MegolmResult<()> {
```
richvdh, line 1575: per a discussion today in `#matrix-rust-sdk-dev`, helper functions should probably come after the functions they help.

crates/matrix-sdk-crypto/src/machine/tests/decryption_verification_state.rs
```diff
@@ -373,3 +380,251 @@ async fn test_verification_states_multiple_device() {
 
     assert_eq!(VerificationState::Unverified(VerificationLevel::UnverifiedIdentity), state);
 }
+
+// Helper function that encrypts a message and shares the Megolm session
+// with a recipient
```
richvdh, line 385: helpers after the functions they help, please

crates/matrix-sdk-crypto/src/olm/group_sessions/sender_data.rs
```diff
@@ -74,6 +74,14 @@ pub enum SenderData {
         /// If false, we had simply accepted the key as this user's latest
         /// key.
         master_key_verified: bool,
+
+        #[serde(default)]
+        /// Whether the user's master key was previously verified, but the
+        /// current master key is not.
```
richvdh, line 80: if this is the actual behaviour (and I'm not sure that it is), it needs a name that reflects it.

crates/matrix-sdk-crypto/src/olm/group_sessions/sender_data.rs
richvdh: You've asked me for my opinion on this question, but this change doesn't seem to appear anywhere in the PR, so I don't think I have enough context on the change to form a sensible opinion.

crates/matrix-sdk-crypto/CHANGELOG.md
```diff
@@ -32,6 +32,9 @@ Changes:
 
 Breaking changes:
 
+- `OlmMachine::decrypt_room_event` now takes an `EncryptionSettings` argument.
```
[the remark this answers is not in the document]
richvdh: we should probably also document some of the changes to publicly-visible types (`MegolmError`, `VerificationLevel`, etc)

crates/matrix-sdk-crypto/src/error.rs
```diff
@@ -115,6 +116,10 @@ pub enum MegolmError {
...
+    /// The sender's cross-signing identity isn't trusted
```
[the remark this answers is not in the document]
richvdh: Actually, isn't it more "because the message could not be linked to a verified device" than specific to the sender's identity?

richvdh (changes requested):
Some initial thoughts.

I'm really struggling to follow this as one big diff. +600 lines isn't *huge*, but there is actually quite a lot changing here, and following the logic though everywhere is hard.

I think it might be better to start again on the commit history, and break it into smaller changes. For example - let's separate the changes to `SenderData` from the changes to decryption. Then add the `DecryptionSettings` parameter without actually implementing it, so that we can see the real changes without all the noise from that.

crates/matrix-sdk-crypto/src/olm/group_sessions/sender_data.rs
richvdh: We need to store `previously_verified` (and, for that matter, `master_key_verified`, and the whole of `SenderKnown`-ness) in SenderData rather than in the user's identity because the session data needs to latch towards is most-trusted state. 

Imagine we see *3* different identities for Alice. Initially she has identity A, which is verified. We then see identity B; while that is active, she sends a message. Finally, we see a third identity, C.

For that message sent under identity B, we must remember whether the identity was verified or not, as well as the fact that it followed an identity A that was previously verified.

We cannot simply rely on the verification properties of the *current* identity C.

richvdh: Next question: is it really a property of a SenderKnown session? Arguably the sender isn't really known, because it's a bad state; so we should have a separate enum variant for this case. Also, it would keep clean the interface to CryptoStore for the update logic on /keys/query (
https://github.com/matrix-org/matrix-rust-sdk/issues/3753#issuecomment-2263389209).

So we need a new enum variant called something like `SenderKnownButUnverified`.

Question: should it be `SenderKnownButUnverified` or `SenderKnownButUnverifiedAndPreviouslyVerified` (ie, where do sessions from tofu-trusted identities end up, and which of the variants has a `master_key_verified` field)

Arguably some of the logic is easier with `SenderKnownButUnverified` - and we get a bonus of updating tofu-trusted sessions to verified sessions. But it conflates known-bad state (no longer verified) with ok state (tofu-verified). Also updating all the tofu-trusted sessions could be prohibitively expensive.

What if we had *five* variants? It leaves the update algorithm open for later choice.

... We agree we want five variants. @uhoreg will make a separate PR to create new variants.

richvdh: > We need to store `previously_verified` (and, for that matter, `master_key_verified`, and the whole of `SenderKnown`-ness) in SenderData rather than in the user's identity because the session data needs to latch towards is most-trusted state.
> 
> Imagine we see _3_ different identities for Alice. Initially she has identity A, which is verified. We then see identity B; while that is active, she sends a message. Finally, we see a third identity, C.
> 
> For that message sent under identity B, we must remember whether the identity was verified or not, as well as the fact that it followed an identity A that was previously verified.
> 
> We cannot simply rely on the verification properties of the _current_ identity C.

It would be good to try to distill this train-of-thought into a useful comment in the code.

uhoreg:
replaced by https://github.com/matrix-org/matrix-rust-sdk/pull/3899

PR #3702, opened 2024-07-16
fix(sdk): Sliding sync rooms aren't restored from the cache

Hywan:
This patch fixes a bug where rooms stored in the sliding sync cache aren't restored by the `SlidingSyncBuilder`.

This patch also improves a test to ensure we don't break this in the future again.


bnjbvr (approved):
Makes sense, thanks for the regression test.

PR #3704, opened 2024-07-16
integration tests: reinstate no spurious room update test

bnjbvr:
This reverts the removal of the test done in https://github.com/matrix-org/matrix-rust-sdk/pull/3691, since the test is actually useful (it makes sure the apps don't get spurious updates about the room list), and was indeed badly written (two syncs running at once for one client — good catch on that for the other test @Hywan!).

This rewrites it in such a way that it should fail fast; it's been pretty solid for me locally, let's see how CI likes it…

### DIFF
--- a/crates/matrix-sdk-base/src/sliding_sync.rs
+++ b/crates/matrix-sdk-base/src/sliding_sync.rs
@@ -760,16 +760,18 @@ fn process_room_properties(
     }
 
     if let Some(recency_timestamp) = &room_data.timestamp {
-        room_info.update_recency_timestamp(*recency_timestamp);
+        if room_info.recency_timestamp.as_ref() != Some(recency_timestamp) {
+            room_info.update_recency_timestamp(*recency_timestamp);
 
-        // If it's not a new room, let's emit a `RECENCY_TIMESTAMP` update.
-        // For a new room, the room will appear as new, so we don't care about this
-        // update.
-        if !is_new_room {
-            room_info_notable_updates
-                .entry(room_id.to_owned())
-                .or_default()
-                .insert(RoomInfoNotableUpdateReasons::RECENCY_TIMESTAMP);
+            // If it's not a new room, let's emit a `RECENCY_TIMESTAMP` update.
+            // For a new room, the room will appear as new, so we don't care about this
+            // update.
+            if !is_new_room {
+                room_info_notable_updates
+                    .entry(room_id.to_owned())
+                    .or_default()
+                    .insert(RoomInfoNotableUpdateReasons::RECENCY_TIMESTAMP);
+            }
         }
     }
 }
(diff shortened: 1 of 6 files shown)


bnjbvr:
> it's been pretty solid for me locally, let's see how CI likes it…

Always Famous Last Words. So at first I could reproduce some of the failures locally, but now they don't repro locally, so I'll be chasing them from looking at CI logs. The new logs for notable updates may be helpful, for that matter.

What I *think* may be happening in the test is that a room.member event causes the proxy server to *sometimes* send an update about the membership. I'm double-checking that with the sliding sync proxy folks.

Locally, I've had another intermittent failure with the timeline stale local echo test, where the room is marked as *not* encrypted while it should be. That sounds like a different race condition, and somehow it's not triggering on CI, so I'll open a separate issue/PR for that with its own test for that.

bnjbvr:
Alright, if CI doesn't fail this time, it's ready for review!

testing/matrix-sdk-integration-testing/src/tests/timeline.rs
```diff
@@ -287,7 +287,12 @@ async fn test_stale_local_echo_time_abort_edit() {
     let vector_diff = timeout(Duration::from_secs(5), stream.next()).await.unwrap().unwrap();
     let local_echo = assert_matches!(vector_diff, VectorDiff::PushBack { value } => value);
 
-    assert!(local_echo.is_local_echo());
+    if !local_echo.is_local_echo() {
+        // If the server raced and we've already received the remote echo, then this
+        // test is meaningless, so short-circuit and leave it.
+        return;
+    }
```
Hywan, line 301: THANKS!

crates/matrix-sdk-base/src/sliding_sync.rs
```diff
@@ -760,16 +760,18 @@ fn process_room_properties(
...
-
-        // If it's not a new room, let's emit a `RECENCY_TIMESTAMP` update.
-        // For a new room, the room will appear as new, so we don't care about this
-        // update.
-        if !is_new_room {
-            room_info_notable_updates
-                .entry(room_id.to_owned())
-                .or_default()
-                .insert(RoomInfoNotableUpdateReasons::RECENCY_TIMESTAMP);
+        if room_info.recency_timestamp.as_ref() != Some(recency_timestamp) {
```
Hywan, line 763: I thought it wasn't possible for the server to send the same timestamp twice as any new event would change this value. Should we raise a bug onto the proxy?

Hywan (approved):

crates/matrix-sdk-base/src/sliding_sync.rs
bnjbvr: I think it might rather be that there's *another change elsewhere* in the room response subpart, but the timestamp might be the same, or something. I don't think we can blame the proxy for certainty here, maybe it's an expected behavior.

PR #3705, opened 2024-07-16
notable room updates: have a change to the unread-marker event cause a notable update

bnjbvr:
Also log out why an account data event couldn't be deserialized.


pixlwave:
Tested on iOS and everything is working again 🙇‍♂️

Hywan (approved):
LGTM!

PR #3709, opened 2024-07-16
logging: Extract debug log code into a separate function.

andybalaam:
This reduces the work we do to calculate changed devices etc. when DEBUG logging is not enabled, but more importantly (to me) it makes clear that this code is only used for logging.


poljar (approved):
Ey, that's nice.

Bonus points if you check if the `unknown_lints` thing is still needed.

PR #3710, opened 2024-07-16
sdk: Allow to request animated thumbnails

zecakeh:
New feature from Matrix 1.11.

### DIFF
--- a/crates/matrix-sdk-base/src/media.rs
+++ b/crates/matrix-sdk-base/src/media.rs
@@ -31,14 +31,14 @@ pub enum MediaFormat {
     File,
 
     /// A thumbnail of the file that was uploaded.
-    Thumbnail(MediaThumbnailSize),
+    Thumbnail(MediaThumbnailSettings),
 }
 
 impl UniqueKey for MediaFormat {
     fn unique_key(&self) -> String {
         match self {
             Self::File => "file".into(),
-            Self::Thumbnail(size) => size.unique_key(),
+            Self::Thumbnail(settings) => settings.unique_key(),
         }
     }
 }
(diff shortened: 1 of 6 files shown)


crates/matrix-sdk-base/src/media.rs
```diff
@@ -64,6 +64,46 @@ impl UniqueKey for MediaThumbnailSize {
...
+    /// The desired size of the thumbnail.
+    pub size: MediaThumbnailSize,
+
+    /// If we want to request an animated thumbnail from the homeserver.
+    ///
+    /// If it is `Some(true)`, the server should return an animated thumbnail if
+    /// the media supports it. If it is `Some(false)`, the server must never
+    /// return an animated thumbnail. If it is `None`, the server should not
+    /// return an animated thumbnail.
+    pub animated: Option<bool>,
```
bnjbvr, line 79: It seems there's no difference between `Some(false)` and `None`; do you confirm this, or is the difference subtle (in which case the code comment should elaborate on that)? If there's no difference, can we use a plain `bool` here instead, set to false by default?

bnjbvr (approved):
Thanks. Small API question below.

crates/matrix-sdk-base/src/media.rs
zecakeh: The subtle difference is that in one case it's a MUST NOT for the homeserver and in the other case it's a SHOULD NOT. I am not sure if that difference makes sense for clients, it might make more sense to always send `true` or `false`.

zecakeh: In theory homeservers should ignore unknown query parameters.

zecakeh: Done

bnjbvr (approved):
Sweet, thanks!

PR #3711, opened 2024-07-16
xtask (Swift): Put the headers in a module subdirectory.

pixlwave:
This change fixes a conflict with any other UniFFI library built the same way. This is the SDK equivalent of https://github.com/matrix-org/matrix-rich-text-editor/pull/1018


stefanceriu (approved):
🙌

PR #3713, opened 2024-07-16
room: mark encryption state as missing if a room thinks it's not encrypted after enabling it

bnjbvr:
After enabling encryption, previous code would wait for a sync, assuming (probably?) that the result of encrypting would come from the next sync result.

Unfortunately, this is not sufficient: it could be that the sync would only contain previous events that were unrelated to the enabling of encryption, or that there's no sync in the next 3 seconds. As a result, checking that the room was encrypted after calling `enable_encryption()` doesn't guarantee it always is, and since there's a cache preventing another direct lookup from the state event, it's even worse: the room would consider itself to be not encrypted, with absolute certainty (until the remote echo for the encryption state event came back from the sync). Sending a room message event, in that case, would assume the room is unencrypted, while it is; not sure how receiving clients would handle that, if they know the room's been encrypted since then.

The test introduced in this PR acts as a regression test, where the last assertion would fail without the fix in `enable_encryption()`.

This patch updates it so that, if the room is not encrypted after waiting for the sync, we manually mark the room state as unknown, so it's manually fetched from the state events end point next time. We could even get rid of the waiting on the sync beat, and let an impatient client run the state event request again, if it wants to send an event immediately after enabling encryption.

### DIFF
--- a/crates/matrix-sdk/src/room/mod.rs
+++ b/crates/matrix-sdk/src/room/mod.rs
@@ -1575,6 +1575,25 @@ impl Room {
             // could be quite useful if someone wants to enable encryption and
             // send a message right after it's enabled.
             _ = timeout(self.client.inner.sync_beat.listen(), SYNC_WAIT_TIME).await;
+
+            // If after waiting for a sync, we don't have the encryption state we expect,
+            // assume the local encryption state is incorrect; this will cause
+            // the SDK to re-request it later for confirmation, instead of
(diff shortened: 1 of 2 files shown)


Hywan (commented):
I think the patch is a good idea, but I would prefer to see tests here, esp. because of the use of `sync_lock`.

bnjbvr:
@Hywan now with a test, so re-ready for review \o/

andybalaam (approved):

PR #3715, opened 2024-07-16
matrix_sdk_base: make sticker events suitable as latest_event

surakin:
Currently on Element X if you receive a sticker in a room, the room list will show the room as updated but it will show the latest event that is not a sticker. This change fixes that.

Signed-off-by: 
Marco Antonio Alvarez <surakin@gmail.com>

### DIFF
--- a/crates/matrix-sdk-base/src/latest_event.rs
+++ b/crates/matrix-sdk-base/src/latest_event.rs
@@ -444,6 +444,7 @@ mod tests {
             PossibleLatestEvent::NoUnsupportedMessageLikeType
         );
     }
+
     #[test]
     fn test_redacted_messages_are_suitable() {
         // Ruma does not allow constructing UnsignedRoomRedactionEvent instances.
--- a/crates/matrix-sdk-base/src/latest_event.rs
+++ b/crates/matrix-sdk-base/src/latest_event.rs
@@ -12,7 +12,7 @@ use ruma::events::{
     room::message::SyncRoomMessageEvent,
     AnySyncMessageLikeEvent, AnySyncTimelineEvent,
 };
-use ruma::{MxcUri, OwnedEventId};
+use ruma::{events::sticker::SyncStickerEvent, MxcUri, OwnedEventId};
 use serde::{Deserialize, Serialize};
 
 use crate::MinimalRoomMemberEvent;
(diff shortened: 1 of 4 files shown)


crates/matrix-sdk-base/src/latest_event.rs
```diff
@@ -391,7 +397,7 @@ mod tests {
     }
 
     #[test]
-    fn test_different_types_of_messagelike_are_unsuitable() {
+    fn test_stickers_are_suitable() {
```
bnjbvr, line 400: Thanks; can you add another test for an unsuitable messagelike event?

bnjbvr (changes requested):
Sweet! A small testing request and this would be good to go. Thanks!

crates/matrix-sdk-base/src/latest_event.rs
surakin: done!

crates/matrix-sdk-base/src/latest_event.rs
```diff
@@ -408,10 +406,30 @@ mod tests {
...
+                sender: owned_user_id!("@a:b.c"),
+                origin_server_ts: MilliSecondsSinceUnixEpoch(UInt::new(2123).unwrap()),
+                unsigned: MessageLikeUnsigned::new(),
+            })));
+
+        assert_matches!(
+            is_suitable_for_latest_event(&event),
+            PossibleLatestEvent::NoUnsupportedMessageLikeType
+        );
+    }
```
bnjbvr, line 432: ubernit, sorry :D (I'll apply and squash it with your PR

```suggestion
    }

```

bnjbvr (approved):
Thank you!

bnjbvr:
Hmm Clippy's complaining about some issues. You can run it locally with `cargo xtask ci clippy`, and fix them as they show up.

surakin:
yes I forgot :P doing that now

surakin:
all green :D

bnjbvr (approved):
Thanks!

bnjbvr:
Hmm sorry I realize I might need some input from our product folks on this. Technically the PR is sane and perfect as is, but it's not clear to me that we support images/stickers as the latest visible event in a room, in EX.

surakin:
Sure no problem.
Right now if the latest event is an image it displays 'Image' in the room summary, but if it's a sticker, it shows the previous valid latest event (even when it still pushes the room to the top). So it looks a bit broken :P

bnjbvr:
Yep, that's also what others at Element think, so let's merge it. Thanks again for your contribution!

PR #3717, opened 2024-07-17
sdk-base: Add `RoomInfoNotableUpdateReasons::MEMBERSHIP`

jmartinesp:
This fixes left rooms not disappearing from the room list.

Signed-off-by:

### DIFF
--- a/crates/matrix-sdk-base/src/sliding_sync/mod.rs
+++ b/crates/matrix-sdk-base/src/sliding_sync/mod.rs
@@ -623,14 +623,18 @@ impl BaseClient {
                 // If this event updates the current user's membership, record that in the
                 // room_info.
                 if member.state_key() == meta.user_id.as_str() {
-                    room_info.set_state(member.membership().into());
+                    let new_state: RoomState = member.membership().into();
+                    if new_state != room_info.state() {
+                        room_info.set_state(new_state);
+                        room_info_notable_updates.insert(
+                            room_info.room_id.to_owned(),
+                            RoomInfoNotableUpdateReasons::MEMBERSHIP,
+                        );
+                    }
                     break;
                 }
             }
         }
-
-        room_info_notable_updates
-            .insert(room_info.room_id.to_owned(), RoomInfoNotableUpdateReasons::MEMBERSHIP);
     }
 
     pub(crate) fn deserialize_state_events_from_timeline(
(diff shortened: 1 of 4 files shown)


jmartinesp:
@Hywan just double checking, do we still want to use the `RoomInfoNotableUpdateReasons` APIs? If so I can rebase and fix the conflicts.

crates/matrix-sdk-base/src/sliding_sync/mod.rs
```diff
@@ -564,7 +575,14 @@ impl BaseClient {
...
                 // room_info.
                 if member.state_key() == meta.user_id.as_str() {
-                    room_info.set_state(member.membership().into());
+                    let new_state: RoomState = member.membership().into();
+                    if new_state != room_info.state() {
+                        room_info.set_state(new_state);
+                        room_info_notable_updates.insert(
+                            room_info.room_id.to_owned(),
+                            RoomInfoNotableUpdateReasons::MEMBERSHIP,
+                        );
```
Hywan, line 584: Be careful that it will override an existing entry. Instead, I recommend to use the following code:

```suggestion
room_info_notable_updates
    .entry(room_id.clone()) // get the entry
    .or_default() // if it doesn't exist, create a default one
    .insert(RoomInfoNotableUpdateReasons::MEMBERSHIP); // insert `MEMBERSHIP`
```

Hywan (changes requested):
Looks almost good! Thanks for working on this!

crates/matrix-sdk-base/src/sliding_sync/mod.rs
jmartinesp: Should by fixed by https://github.com/matrix-org/matrix-rust-sdk/pull/3717/commits/ea949a4408d48d82a8ee34afd999825c63f7ce64

Hywan:
> @Hywan just double checking, do we still want to use the `RoomInfoNotableUpdateReasons` APIs? If so I can rebase and fix the conflicts.

Yes! The idea is to restore this API, as mentioned in https://github.com/matrix-org/matrix-rust-sdk/issues/3802.

jmartinesp:
So... does this need any more changes?

Hywan:
I don't think so.

jmartinesp:
Then should we approve it so it can be squashed and merged 😅 ?

Hywan (approved):
LGTM, thanks!

PR #3718, opened 2024-07-17
Room: Add access to the device ID in the same fashion as the user ID.

pixlwave:
Small follow-on from https://github.com/matrix-org/matrix-rust-sdk/pull/3706, this PR adds access to the user's device ID from the `Room` in the same way the user's matrix ID is available. I'm not 100% sure if this is the right pattern but looking at the call site that was broken, it seemed like following the existing pattern would be acceptable.

### DIFF
--- a/bindings/matrix-sdk-ffi/src/room.rs
+++ b/bindings/matrix-sdk-ffi/src/room.rs
@@ -534,6 +534,10 @@ impl Room {
         self.inner.own_user_id().to_string()
     }
 
+    pub fn own_device_id(&self) -> String {
+        self.inner.own_device_id().to_string()
+    }
+
     pub async fn typing_notice(&self, is_typing: bool) -> Result<(), ClientError> {
         Ok(self.inner.typing_notice(is_typing).await?)
     }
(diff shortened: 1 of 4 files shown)


bnjbvr (changes requested):
Can you not use `Client::device_id()` instead? It's exposed in the FFI layer, so there are chances this PR doesn't need to exist in the first place…

pixlwave:
Sure, we just need to rework the service as it doesn't have access to the client, only the room but that's not a problem. Just wanted to explore this route first as tbh, I don't expect `Room` to have `ownUserID` on it so thought it might be how we do things on the SDK.

bnjbvr:
Ok. At the extreme, we could have something like:

```
impl Room {
  pub fn own_device_id(&self) -> &DeviceId { self.client.device_id().expect("if you're a room you should have a device id") }
}
```

but the current `Room::own_user_id()` is a bit of a shame, and undocumented why we store it along the room details in the first place :thinking:

poljar:
> if you're a room you should have a device id

And don't call me a room.

PR #3723, opened 2024-07-17
FFI: Allow restoring an existing session with SSS enabled.

pixlwave:
Small follow-up to #3676 that makes 2 small changes:
- Ignore a `Session`'s sliding sync proxy when restoring a client in the FFI so we don't have to do it in both EXI and EXA.
- Ignore a custom sliding sync proxy when building a client with support for SSS (not tested that one specifically but seemed worth adding whilst writing the other one).


Hywan (approved):
Looks sane to me. Thanks!

PR #3728, opened 2024-07-18
fix(ui): `RoomList` reacts to all `RoomInfoNotableUpdate`

Hywan:
The `RoomList` provides a `Stream<Item = Vec<VectorDiff<Room>>>`. This `Stream` receives updates from 2 sources: `RoomList::entries`, and `Receiver<RoomInfoNotableUpdate>`. When a `RoomInfo` is updated, a notable update is emitted and broadcasted. The `RoomList` was filtering these notable updates by _reasons_ (namely `RoomInfoNotableUpdateReasons`).

This is great and it's a good idea since we can filter which `RoomInfoNotableUpdate` will trigger a `RoomList` update. However, too many _reasons_ were hidden/implicit, and it creates several regressions because (i) these _reasons_ were implicit, (ii) since the business rules are not defined, there is no tests for that (not in this SDK, not in apps like ElementX). It means we discover missing _reasons_ bug after bug. It's not pleasant.

The reality is: we are in the middle of big changes, mostly with room list client-side sorting and simplified sliding sync. We want to relax a little bit. This patch then disable the feature _filter updates by reasons_. The `RoomList` will update to all `RoomInfoNotableUpdate` for the moment. We will get back to this optimisation later.

### DIFF
--- a/crates/matrix-sdk-ui/src/room_list_service/room_list.rs
+++ b/crates/matrix-sdk-ui/src/room_list_service/room_list.rs
@@ -25,7 +25,7 @@ use matrix_sdk::{
     executor::{spawn, JoinHandle},
     Client, SlidingSync, SlidingSyncList,
 };
-use matrix_sdk_base::{RoomInfoNotableUpdate, RoomInfoNotableUpdateReasons};
+use matrix_sdk_base::RoomInfoNotableUpdate;
 use tokio::{select, sync::broadcast};
 use tracing::trace;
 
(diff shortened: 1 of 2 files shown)


bnjbvr (approved):
Yep, thanks! It'd be nice to get some testing from the EX app devs first before merging this :+1:

Hywan:
@jmartinesp confirmed to me that it fixes several regressions we saw recently. I think we are good.

jmartinesp:
> @jmartinesp confirmed to me that it fixes several regressions we saw recently. I think we are good.

I also found some exciting new (old) ones!

> invites work ok, except decline somehow got stuck when declining, but I'm not sure it's related, but a newly created room with no messages still displays at the bottom 🫤

> also, the room name doesn't seem to load for empty rooms sometimes (I created an empty room from EW, it displays 'No room name' on EXA in the same account)

> and leaving rooms doesn't make them disappear from the room list either

Hywan:
Yes, those aren't related to room info notable update I think. Let's see that…

jmartinesp:
I mean, some of those were fixed by adding notable updates.

PR #3730, opened 2024-07-18
timeline: edit/abort cleanups

bnjbvr:
A review commit by commit should be possible there. This makes the code less verbose, in preparation for sending reactions with the send queue, and reacting to local echoes.

### DIFF
--- a/crates/matrix-sdk-ui/src/timeline/mod.rs
+++ b/crates/matrix-sdk-ui/src/timeline/mod.rs
@@ -459,10 +459,7 @@ impl Timeline {
 
     /// Returns a local or remote timeline item identified by this transaction
     /// id.
-    async fn any_timeline_item_by_txn_id(
-        &self,
-        txn_id: &TransactionId,
-    ) -> Option<EventTimelineItem> {
+    async fn item_by_transaction_id(&self, txn_id: &TransactionId) -> Option<EventTimelineItem> {
         let items = self.inner.items().await;
 
         let (_, found) = rfind_event_item(&items, |item| match &item.kind {
@@ -496,7 +493,7 @@ impl Timeline {
         let event_id = match item.identifier() {
             TimelineEventItemId::TransactionId(txn_id) => {
                 // See if we have an up-to-date timeline item with that transaction id.
-                if let Some(item) = self.any_timeline_item_by_txn_id(&txn_id).await {
+                if let Some(item) = self.item_by_transaction_id(&txn_id).await {
                     match item.handle() {
                         TimelineItemHandle::Remote(event_id) => event_id.to_owned(),
                         TimelineItemHandle::Local(handle) => {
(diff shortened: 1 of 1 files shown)


crates/matrix-sdk-ui/src/timeline/event_item/mod.rs
```diff
@@ -474,6 +485,24 @@ impl EventTimelineItem {
...
+        match &self.kind {
+            EventTimelineItemKind::Local(local) => {
+                if let Some(event_id) = local.event_id() {
+                    TimelineItemHandle::Remote(event_id)
+                } else {
+                    TimelineItemHandle::Local(
+                        local
+                            .send_handle
+                            .as_ref()
+                            .expect("EventTimelineItem::handle used in a unit test"),
```
andybalaam, line 499: This message is a bit confusing to me. Does it mean that we don't expect to find a missing `send_handle` except in a unit test? Can we guarantee that? Maybe something simpler like "Unexpected missing send_handle" would work?

crates/matrix-sdk-ui/src/timeline/mod.rs
```diff
@@ -455,20 +456,22 @@ impl Timeline {
         })
     }
 
-    /// Given a transaction id, try to find a remote echo that used this
-    /// transaction id upon sending.
-    async fn find_remote_by_transaction_id(&self, txn_id: &TransactionId) -> Option<OwnedEventId> {
+    /// Returns a local or remote timeline item identified by this transaction
+    /// id.
+    async fn any_timeline_item_by_txn_id(
```
andybalaam, line 461: `find_item_by_transaction_id` might be a clearer name. The "any" part doesn't seem to add anything, although I understand it makes sense in the context that this method used to find only remote items.

andybalaam (approved):
Looks good! A couple of tiny suggestions.

crates/matrix-sdk-ui/src/timeline/event_item/mod.rs
bnjbvr: > Does it mean that we don't expect to find a missing send_handle except in a unit test?
Yes

> Can we guarantee that?
We may try to sparkle generics there to mean presence or absence of the send handle, but it's overkill IMO.

> Unexpected missing send_handle
Sounds good!

crates/matrix-sdk-ui/src/timeline/mod.rs
bnjbvr: There's already `item_by_transaction_id` which returns only local echoes, so I wanted something to distinguish it. But maybe the right move would be to rename `item_by_transaction_id` to something else :thinking:

andybalaam: Yeah, I would go for `local_item_by_transation_id` and `item_by_transaction_id` I guess.

PR #3732, opened 2024-07-18
fix(base): Fix a bug when an invite has no timestamp

Hywan:
The sliding sync proxy has a bug: despite the presence of
`m.room.create` in `bump_event_types`, an invite with an `m.room.create`
event will not have a `timestamp`. Thus, such a room cannot be sorted
reliably.

This patch fixes this problem with a terrible hack, where it tries to
find an `origin_server_ts` value from within the `invite_state` state
events.

Please read the comment to learn more.


jmartinesp (approved):
I can confirm this fixes ordering for invites for me.

poljar (approved):
Ouch but OK, let's merge this.

andybalaam (approved):

PR #3734, opened 2024-07-19
sqlite: Don't depend on vodozemac directly

poljar:
There's no reason to depend on vodozemac if you depend on the matrix-sdk-crypto crate, we have an re-export which ensures that the same version is always used.


poljar:
That ain't going to work without making the error non-exhaustive.

PR #3740, opened 2024-07-19
Add set_account_data to room, too

gnunicorn:
As it says on the tin. Needed the functions but they were missing. They are analogous to the GlobalAccountData setters.

Signed-off-by: Benjamin Kampmann <ben@acter.global>


crates/matrix-sdk/src/room/mod.rs
```diff
@@ -952,6 +952,43 @@ impl Room {
         Ok(true)
     }
 
+    /// Set the given account data event for this room.
```
poljar, line 955: That's a bit terse, could you expand this with an example?

Likewise for the other method.

poljar (changes requested):
Sorry but we'll need tests for this.

poljar (approved):

PR #3743, opened 2024-07-19
SDK: Ignore the discovered/custom sliding sync proxy when using SSS.

pixlwave:
Oversight on my part in #3723, this code needed to handle the discovered proxy as well as the custom one.

### DIFF
--- a/bindings/matrix-sdk-ffi/src/client_builder.rs
+++ b/bindings/matrix-sdk-ffi/src/client_builder.rs
@@ -12,7 +12,6 @@ use matrix_sdk::{
 };
 use ruma::api::error::{DeserializationError, FromHttpResponseError};
 use tracing::{debug, error};
-use url::Url;
 use zeroize::Zeroizing;
 
 use super::{client::Client, RUNTIME};
@@ -527,6 +526,10 @@ impl ClientBuilder {
 
         inner_builder = inner_builder.with_encryption_settings(builder.encryption_settings);
 
+        if let Some(sliding_sync_proxy) = builder.sliding_sync_proxy {
+            inner_builder = inner_builder.sliding_sync_proxy(sliding_sync_proxy);
+        }
+
         inner_builder =
             inner_builder.simplified_sliding_sync(builder.is_simplified_sliding_sync_enabled);
 
@@ -535,25 +538,6 @@ impl ClientBuilder {
         }
 
         let sdk_client = inner_builder.build().await?;
-
-        // At this point, `sdk_client` might contain a `sliding_sync_proxy` that has
-        // been configured by the homeserver (if it's a `ServerName` and the
-        // `.well-known` file is filled as expected).
-        //
-        // If `builder.sliding_sync_proxy` contains `Some(_)`, it means one wants to
-        // overwrite this value. It would be an error to call
-        // `sdk_client.set_sliding_sync_proxy()` with `None`, as it would erase the
-        // `sliding_sync_proxy` if any, and it's not the intended behavior.
-        //
-        // So let's call `sdk_client.set_sliding_sync_proxy()` if and only if there is
-        // `Some(_)` value in `builder.sliding_sync_proxy`. That's really important: It
-        // might not break an existing app session, but it is likely to break a new
-        // session, which not immediate to detect if there is no test.
-        if !builder.is_simplified_sliding_sync_enabled {
-            if let Some(sliding_sync_proxy) = builder.sliding_sync_proxy {
-                sdk_client.set_sliding_sync_proxy(Some(Url::parse(&sliding_sync_proxy)?));
-            }
-        }
 
         Ok(Arc::new(
             Client::new(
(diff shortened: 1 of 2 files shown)


bnjbvr (changes requested):
Sorry, this fix should not happen only for FFI users, but for all users of the SDK (and then we can even include a test \o/ a very simple one). It should happen around [here](https://github.com/matrix-org/matrix-rust-sdk/blob/d65e33ca6af52bbdb55e77f84bf650788486eb28/crates/matrix-sdk/src/client/builder.rs#L471-L472), where we'd set the proxy to `None` if the boolean is set. Can you implement that and add a test please?

bindings/matrix-sdk-ffi/src/client_builder.rs
pixlwave, line 1: Seems we weren't using the SDK logic for this, so now we are!

crates/matrix-sdk/src/client/builder.rs
```diff
@@ -450,6 +450,7 @@ impl ClientBuilder {
 
         let http_client = HttpClient::new(inner_http_client.clone(), self.request_config);
 
+        #[allow(unused_variables)]
```
pixlwave, line 453: This is moved up from the old `if let` (`well_known` is only used when `feature = "experimental-sliding-sync"` and Clippy wants to rename it).

crates/matrix-sdk/src/client/builder.rs
```diff
@@ -471,19 +472,29 @@ impl ClientBuilder {
...
         #[cfg(feature = "experimental-sliding-sync")]
-        if self.requires_sliding_sync && sliding_sync_proxy.is_none() {
-            // In the future we will need to check for native support on the homeserver too.
-            return Err(ClientBuildError::SlidingSyncNotAvailable);
+        if self.requires_sliding_sync {
+            if self.is_simplified_sliding_sync_enabled {
+                // TODO: The server doesn't yet expose the availability of SSS.
+            } else if sliding_sync_proxy.is_none() {
+                return Err(ClientBuildError::SlidingSyncNotAvailable);
+            }
```
bnjbvr, line 497: I would coalesce all these checks together, since it's not clear that we'd keep `requires_sliding_sync` for `is_simplified_sliding_sync` (at least we'd have some method named after it). Instead, we could even throw an error when `requires_sliding_sync` and `is_simplified_sliding_sync` are set at the same time, because it sounds like a misconfigure.

crates/matrix-sdk/src/client/builder.rs
```diff
@@ -1174,4 +1213,18 @@ pub(crate) mod tests {
             object
         })
     }
+
+    /// The FFI's build has SSS disabled by default, but the SDK has it enabled
+    /// by default. The tests are aligned with the FFI for now (until we remove
+    /// regular sliding sync).
+    fn make_ffi_builder() -> ClientBuilder {
```
bnjbvr, line 1220: I wouldn't mention FFI here, because it's not clear for a usual reader what FFI we're referring to, and we need to keep a clear separation of concerns. Maybe rename it `make_non_sss_client_builder`?

crates/matrix-sdk/src/client/builder.rs
```diff
@@ -1061,6 +1072,34 @@ pub(crate) mod tests {
...
+            .mount(&server)
+            .await;
+
+        // When building a client for simplified sliding sync with the base server.
+        builder = builder.simplified_sliding_sync(true);
+        builder = builder.server_name_or_homeserver_url(server.uri());
+        let _client = builder.build().await.unwrap();
+
+        // Then a client should not use the discovered sliding sync proxy.
+        #[cfg(feature = "experimental-sliding-sync")]
```
bnjbvr, line 1099: nit: don't need this cfg because the test is guarded against the same cfg guard

Then we could rename `_client` to `client` since it'd be always used.

bnjbvr (changes requested):
Thanks for writing a test!

crates/matrix-sdk/src/client/builder.rs
pixlwave: Ah ok, I imagined that `requires_sliding_sync` would make sure you build a client with support for the type of sliding sync you configured it with but I see the confusion there. For now I'll revert this change and we can think about it when SSS is a bit more mature.

crates/matrix-sdk/src/client/builder.rs
```diff
@@ -471,12 +472,19 @@ impl ClientBuilder {
...
-                    well_known.sliding_sync_proxy.and_then(|p| Url::parse(&p.url).ok())
+        #[cfg(feature = "experimental-sliding-sync")]
+        if self.is_simplified_sliding_sync_enabled {
+            // When using Simplified MSC3575, don't use a sliding sync proxy, allow the
+            // requests to be sent directly to the homeserver.
+            tracing::info!("Simplified MSC3575 is enabled, ignoring any sliding sync proxy.");
+            sliding_sync_proxy = None;
+        } else {
+            // Otherwise, if a proxy wasn't set, use the one discovered from the well-known.
+            if let Some(well_known) = well_known {
```
bnjbvr, line 483:
```suggestion
        } else if let Some(well_known) = well_known {
            // Otherwise, if a proxy wasn't set, use the one discovered from the well-known.
```

bnjbvr (approved):
lgtm, thanks

PR #3746, opened 2024-07-22
workspace: update dependencies

### LINKED ISSUE #3744: RUSTSEC-2024-0357: `MemBio::get_buf` has undefined behavior with empty buffers
| Details |  |
| --- | --- |
| Package | `openssl` |
| Version | `0.10.64` |
| URL | https://github.com/sfackler/rust-openssl/pull/2266 |
| Patched Versions | >=0.10.66 |

Previously, `MemBio::get_buf` called `slice::from_raw_parts` with a null-pointer, which violates the functions invariants, leading to undefined behavior. In debug builds this would produce an assertion failure. This is now fixed.

bnjbvr:
Fixes #3744.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,14 +4,14 @@ version = 3
 
 [[package]]
 name = "accessory"
-version = "1.3.0"
+version = "1.3.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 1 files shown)


andybalaam (approved):
Confirmed that this includes an update to openssl 10.66 which is the required version mentioned in https://github.com/matrix-org/matrix-rust-sdk/issues/3744

Issue #3726, opened 2024-07-18
Live pinned events list/timeline of a a room


Velin92:
All pinned events of a room should be tracked live, so that clients are able to display a list of pinned events at any time, and any change made on such events can result on a live update client side.
Should also be possible to operate actions on such events, just like in a normal timeline.

jmartinesp:
I think I'd split this into 2 parts:

1. Getting the updated `RoomPinnedEventsEventContent` from `/sync`, storing the contained ids and making them accessible to the rest of the SDK.
2. Actually loading them using a focused timeline or some other implementation, and making this info available for the clients through FFI.

Issue #3716, opened 2024-07-17
testing: use more the `EventFactory` to get rid of `sync_timeline_event!` and `EventBuilder`


bnjbvr:
The `EventBuilder` struct/methods and `sync_timeline_event!` macros are used all over the place, but they're quite verbose and not the easier to use, when usually tests want simple things ("a plain text message with that content sent from this user") and don't care about low-level Matrix details.

So as to improve this, I've added an `EventFactory` struct that makes it easier to write such events, by chaining calls, and taking care of the fields that always need to be set (if necessary).

It would be great to fully get rid of `EventBuilder` and `sync_timeline_event!`, so we don't have three ways of defining test events, but a single one. I've started migrating in a few PRs, but I think this is a great way to get accointed with the code base by making relatively simple changes:

- look at a use of `EventBuilder` or `sync_timeline_event!`
- consider if it can be replaced by using an `EventFactory` instead
  - if yes, do the replacement
    - some tests do already have an `EventFactory`, often called `factory`, or a plain `f` in the code; reuse those if present, or follow that same naming convention for new uses.
  - if no: consider adding new methods/helpers to the `EventFactory`. This might be a bit more involved, so if you're chasing the good first issue part of this issue, then I'd recommend doing the plain replacements instead, aka the previous bullet item.

yousefmoazzam:
Hi @bnjbvr! New matrix user here, and I enjoy working with rust, so I thought it may be fun to try and contribute 🙂 The details in the OP were really useful for this one, it helped me start taking a look at what I could do; as a newbie that's much appreciated!

I spotted a small use of `sync_timeline_event!` in the integration tests for the notification functionality: https://github.com/matrix-org/matrix-rust-sdk/blob/3b84b2c5e757860e60961b8361a3b62005e4831c/crates/matrix-sdk/tests/integration/notification.rs#L68-L83

and I thought that this may be a relatively isolated place to tentatively try and replace the `sync_timeline_event!` with an `EventFactory`.

Would you mind if I gave that a go?

Hywan:
@yousefmoazzam Feel free to submit any patches that remove the `sync_timeline_event!` macro by using the `EventFactory` 🙂.

PR #3747, opened 2024-07-22
sdk-base: add `Room::pinned_events(&self)` 

jmartinesp:
This can be used to subscribe to pinned event ids in a room.

Implements the first part of https://github.com/matrix-org/matrix-rust-sdk/issues/3726.

Signed-off-by:

### DIFF
--- a/crates/matrix-sdk-base/src/rooms/mod.rs
+++ b/crates/matrix-sdk-base/src/rooms/mod.rs
@@ -30,6 +30,7 @@ use ruma::{
             join_rules::RoomJoinRulesEventContent,
             member::MembershipState,
             name::RoomNameEventContent,
+            pinned_events::RoomPinnedEventsEventContent,
             tombstone::RoomTombstoneEventContent,
             topic::RoomTopicEventContent,
         },
@@ -117,6 +118,8 @@ pub struct BaseRoomInfo {
     /// others, and this field collects them.
     #[serde(skip_serializing_if = "RoomNotableTags::is_empty", default)]
     pub(crate) notable_tags: RoomNotableTags,
+    /// The `m.room.pinned_events` of this room.
+    pub(crate) pinned_events: Option<RoomPinnedEventsEventContent>,
 }
 
 impl BaseRoomInfo {
@@ -194,6 +197,9 @@ impl BaseRoomInfo {
                     ev.as_original().is_some_and(|o| !o.content.active_memberships(None).is_empty())
                 });
             }
+            AnySyncStateEvent::RoomPinnedEvents(p) => {
+                self.pinned_events = p.as_original().map(|p| p.content.clone());
+            }
             _ => return false,
         }
 
@@ -252,6 +258,11 @@ impl BaseRoomInfo {
                 // Ignore stripped call state events. Rooms that are not in Joined or Left state
                 // wont have call information.
                 return false;
+            }
+            AnyStrippedStateEvent::RoomPinnedEvents(p) => {
+                if let Some(pinned) = p.content.pinned.clone() {
+                    self.pinned_events = Some(RoomPinnedEventsEventContent::new(pinned));
+                }
             }
             _ => return false,
         }
@@ -349,6 +360,7 @@ impl Default for BaseRoomInfo {
             rtc_member: BTreeMap::new(),
             is_marked_unread: false,
             notable_tags: RoomNotableTags::empty(),
+            pinned_events: None,
         }
     }
 }
--- a/crates/matrix-sdk-base/src/rooms/normal.rs
+++ b/crates/matrix-sdk-base/src/rooms/normal.rs
@@ -947,6 +947,11 @@ impl Room {
     pub fn recency_stamp(&self) -> Option<u64> {
         self.inner.read().recency_stamp
     }
+
+    /// Get the list of event ids for pinned events in this room.
+    pub fn pinned_events(&self) -> Vec<OwnedEventId> {
+        self.inner.get().base_info.pinned_events.map(|content| content.pinned).unwrap_or_default()
+    }
 }
 
 /// The underlying pure data structure for joined and left rooms.
@@ -1615,10 +1620,11 @@ mod tests {
                     SyncRoomMemberEvent,
                 },
                 name::RoomNameEventContent,
+                pinned_events::RoomPinnedEventsEventContent,
             },
             AnySyncStateEvent, EmptyStateKey, StateEventType, StateUnsigned, SyncStateEvent,
         },
-        room_alias_id, room_id,
+        owned_event_id, room_alias_id, room_id,
         serde::Raw,
         user_id, EventEncryptionAlgorithm, MilliSecondsSinceUnixEpoch, OwnedEventId, OwnedUserId,
         UserId,
@@ -1671,7 +1677,9 @@ mod tests {
             latest_event: Some(Box::new(LatestEvent::new(
                 Raw::from_json_string(json!({"sender": "@u:i.uk"}).to_string()).unwrap().into(),
             ))),
-            base_info: Box::new(BaseRoomInfo::new()),
+            base_info: Box::new(
+                assign!(BaseRoomInfo::new(), { pinned_events: Some(RoomPinnedEventsEventContent::new(vec![owned_event_id!("$a")])) }),
+            ),
             read_receipts: Default::default(),
             warned_about_unknown_room_version: Arc::new(false.into()),
             cached_display_name: None,
@@ -1720,6 +1728,9 @@ mod tests {
                 "name": null,
                 "tombstone": null,
                 "topic": null,
+                "pinned_events": {
+                    "pinned": ["$a"]
+                },
             },
             "read_receipts": {
                 "num_unread": 0,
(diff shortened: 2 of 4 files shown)


crates/matrix-sdk-base/src/store/mod.rs
```diff
@@ -340,6 +340,9 @@ pub struct StateChanges {
     /// A map from room id to a map of a display name and a set of user ids that
     /// share that display name in the given room.
     pub ambiguity_maps: BTreeMap<OwnedRoomId, BTreeMap<String, BTreeSet<OwnedUserId>>>,
+
+    /// A map of `RoomId` to pinned `OwnedEventId`s for each room.
+    pub pinned_events: BTreeMap<OwnedRoomId, Vec<OwnedEventId>>,
```
bnjbvr, line 345: I don't think this should be here: this is not stored, and this struct is meant to include things that are to be stored. Instead, you probably want a local variable in the one function that makes use of this `pinned_events ` field.

crates/matrix-sdk-base/src/rooms/normal.rs
```diff
@@ -132,6 +132,10 @@ pub struct Room {
     /// to disk but held in memory.
     #[cfg(all(feature = "e2e-encryption", feature = "experimental-sliding-sync"))]
     pub latest_encrypted_events: Arc<SyncRwLock<RingBuffer<Raw<AnySyncTimelineEvent>>>>,
+
+    /// The list of pinned event ids in this room in an observable format so
+    /// they can be subscribed to.
+    pub pinned_event_ids: SharedObservable<Vec<OwnedEventId>>,
```
bnjbvr, line 138: nit: not `pub`, since there's a getter for a subscriber

Also, would it make sense to use `ObservableVector` here? If we did that, and rejigger the API so that we only push new pinned event / remove pinned event, we could have a finer-grained API, but maybe it's not worth it?

crates/matrix-sdk-base/src/client.rs
```diff
@@ -1173,6 +1180,15 @@ impl BaseClient {
             }
         }
 
+        for (room_id, pinned_event_ids) in changes.pinned_events.clone() {
+            if let Some(room) = self.store.room(&room_id) {
+                room.pinned_event_ids.update(|ids| {
+                    ids.clear();
+                    ids.append(&mut pinned_event_ids.clone());
```
bnjbvr, line 1187: :crab: :teacher: 

- [Vec::extend](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.extend) can take an iterator of the element type (that's `ids`)
- and a `Vec` also has an `IntoIter` implementation, so it can be used as the parameter for `extend` (here I'm talking about `pinned_event_ids`).

Combining the two, that gives you a way to avoid the cloning:

```suggestion
                    ids.extend(pinned_event_ids);
```

crates/matrix-sdk-base/src/client.rs
```diff
@@ -1173,6 +1180,15 @@ impl BaseClient {
             }
         }
 
+        for (room_id, pinned_event_ids) in changes.pinned_events.clone() {
```
bnjbvr, line 1183: nit: you shouldn't need the `clone()` here, and as said before `changes.pinned_events` should be a local variable

crates/matrix-sdk-base/src/sliding_sync/mod.rs
```diff
@@ -2178,6 +2181,59 @@ mod tests {
...
+        assert!(pinned_events_subscriber.get().is_empty());
+
+        // Load new pinned event id
+        let mut room_response = http::response::Room::new();
+        room_response.required_state.push(make_state_event(
+            user_a_id,
+            "",
+            RoomPinnedEventsEventContent::new(vec![pinned_event_id.clone()]),
+            None,
+        ));
```
bnjbvr, line 2209: https://github.com/matrix-org/matrix-rust-sdk/issues/3716 :eyes:

crates/matrix-sdk-base/src/rooms/normal.rs
jmartinesp: `ObservableVector` doesn't implement `Clone` so it can't be used in `Room`, which does.

crates/matrix-sdk-base/src/store/mod.rs
jmartinesp: Removed this in https://github.com/matrix-org/matrix-rust-sdk/pull/3747/commits/eafed32f64566eedd2c0edb0fe200fd3b7c1f672

crates/matrix-sdk-base/src/client.rs
jmartinesp: I had to use `Vector::extend_from_slice` in some places instead, but it's done in https://github.com/matrix-org/matrix-rust-sdk/pull/3747/commits/8c43c7523b37a595cc9b6cddcf22da78d560c46b

crates/matrix-sdk-base/src/client.rs
jmartinesp: Removed in https://github.com/matrix-org/matrix-rust-sdk/pull/3747/commits/8c43c7523b37a595cc9b6cddcf22da78d560c46b

crates/matrix-sdk-base/src/sliding_sync/mod.rs
jmartinesp: I think I'll attach this in some other PR.

jmartinesp:
I had to add an extra check so the pinned event ids are only emitted if they are different from the previous list.

jmartinesp:
> > I had to add an extra check so the pinned event ids are only emitted if they are different from the previous list.
> 
> Sweet, did you add a test for that too? 🥺

The test for this is the previously added one not failing anymore 😅 .

jmartinesp:
I actually ended up adding an extra test check and improving how the new values are set.

crates/matrix-sdk-base/src/rooms/normal.rs
```diff
@@ -132,6 +132,10 @@ pub struct Room {
     /// to disk but held in memory.
     #[cfg(all(feature = "e2e-encryption", feature = "experimental-sliding-sync"))]
     pub latest_encrypted_events: Arc<SyncRwLock<RingBuffer<Raw<AnySyncTimelineEvent>>>>,
+
+    /// The list of pinned event ids in this room in an observable format so
+    /// they can be subscribed to.
+    pub(crate) pinned_event_ids: SharedObservable<Vec<OwnedEventId>>,
```
bnjbvr, line 138: So the reason why we usually don't put things in this struct is that they're not persisted after a reload; if you reload a room, you'd have to wait for the next sync to arrive to give you the list of events, and if you happen to be offline, that means that now you don't have access to pinned events.

From here, we have a few possibilities:

- include the `Vec<OwnedEventId>` in `RoomInfo`,
- include the pinned event in the `BaseRoomInfo`, and have a getter to read the pinned event ids from there
- have a new (de)serializable field in `Room` that is filled at startup, and updated as the events come. You've done half of the work here by having it be updated as the events come, and we'd need some way to serialize/deserialize it too later.

My preference would likely go to the first or second option (second option might be faster to ser/de), to benefit from persistence. Can you also add a test that would've failed without this change, please?

bnjbvr (changes requested):

crates/matrix-sdk-base/src/rooms/normal.rs
jmartinesp: Ok, I tried fixing this in https://github.com/matrix-org/matrix-rust-sdk/pull/3747/commits/4aa51cada2be754cddbd6f88f1b2fe68c5a5ef7c using the 2nd strategy 🤞 .

crates/matrix-sdk-base/src/rooms/mod.rs
```diff
@@ -253,6 +259,11 @@ impl BaseRoomInfo {
                 // wont have call information.
                 return false;
             }
+            AnyStrippedStateEvent::RoomPinnedEvents(p) => {
+                if let Some(pinned) = p.content.pinned.clone() {
+                    self.pinned_events = Some(RoomPinnedEventsEventContent::new(pinned));
+                }
+            }
```
jmartinesp, line 266: I'm not entirely sure about this part.

bnjbvr: Stripped events = for a room in which we're invited. I think your code is fine, we could also decide to store `None` there, but this shouldn't matter much :+1: I don't expect one to be able to look at pinned events in an invited room, at least for a first pass.

crates/matrix-sdk-base/src/rooms/mod.rs
```diff
@@ -117,6 +118,8 @@ pub struct BaseRoomInfo {
     /// others, and this field collects them.
     #[serde(skip_serializing_if = "RoomNotableTags::is_empty", default)]
     pub(crate) notable_tags: RoomNotableTags,
+    /// The `m.room.pinned_events` of this room.
+    pub(crate) pinned_events: Option<RoomPinnedEventsEventContent>,
```
bnjbvr, line 122: Note to self: i think we could have used `MinimalStateEvent` here, but it's unclear what's the value of it, so no need to change anything here.

bnjbvr (approved):
Thanks!

crates/matrix-sdk-base/src/rooms/mod.rs
jmartinesp: I can try changing it so it matches the other fields if it's easy to do (I'm not sure how I missed it 🫤 ).

jmartinesp: Actually, it does make everything way more complex just to store a possible event id which isn't used anywhere, so it's probably not worth to add it.

PR #3750, opened 2024-07-23
sdk & ffi: add `pin_event` and `unpin_event` fns, as well as power level checks

jmartinesp:
- Adds `fn Timeline::pin_event`, `fn Timeline::unpin_event` and exposes them through the FFI.
- Adds `fn Room::can_user_pin_unpin` to make sure the user can perform the action.
- Exposes the currently pinned event ids using `RoomInfo`.

Depends on https://github.com/matrix-org/matrix-rust-sdk/pull/3747

Signed-off-by:

### DIFF
--- a/crates/matrix-sdk-ui/src/timeline/mod.rs
+++ b/crates/matrix-sdk-ui/src/timeline/mod.rs
@@ -50,6 +50,7 @@ use ruma::{
                 AddMentions, ForwardThread, OriginalRoomMessageEvent, RoomMessageEventContent,
                 RoomMessageEventContentWithoutRelation,
             },
+            pinned_events::RoomPinnedEventsEventContent,
             redaction::RoomRedactionEventContent,
         },
         AnyMessageLikeEventContent, AnySyncMessageLikeEvent, AnySyncTimelineEvent,
@@ -854,6 +855,42 @@ impl Timeline {
             self.send_single_receipt(receipt_type, ReceiptThread::Unthreaded, event_id).await
         } else {
             trace!("can't mark room as read because there's no latest event id");
+            Ok(false)
+        }
+    }
+
+    /// Adds a new pinned event by sending an updated `m.room.pinned_events`
+    /// event containing the new event id.
+    ///
+    /// Returns `true` if we sent the request, `false` if the event was already
+    /// pinned.
+    pub async fn pin_event(&self, event_id: &EventId) -> Result<bool> {
+        let mut pinned_events = self.room().pinned_events();
+        let event_id = event_id.to_owned();
+        if pinned_events.contains(&event_id) {
+            Ok(false)
+        } else {
+            pinned_events.push(event_id);
+            let content = RoomPinnedEventsEventContent::new(pinned_events);
+            self.room().send_state_event(content).await?;
+            Ok(true)
+        }
+    }
+
+    /// Adds a new pinned event by sending an updated `m.room.pinned_events`
+    /// event without the event id we want to remove.
+    ///
+    /// Returns `true` if we sent the request, `false` if the event wasn't
+    /// pinned.
+    pub async fn unpin_event(&self, event_id: &EventId) -> Result<bool> {
+        let mut pinned_events = self.room().pinned_events();
+        let event_id = event_id.to_owned();
+        if let Some(idx) = pinned_events.iter().position(|e| *e == *event_id) {
+            pinned_events.remove(idx);
+            let content = RoomPinnedEventsEventContent::new(pinned_events);
+            self.room().send_state_event(content).await?;
+            Ok(true)
+        } else {
             Ok(false)
         }
     }
--- a/crates/matrix-sdk-ui/tests/integration/timeline/mod.rs
+++ b/crates/matrix-sdk-ui/tests/integration/timeline/mod.rs
@@ -507,3 +507,169 @@ async fn test_duplicate_maintains_correct_order() {
     let content = items[3].as_event().unwrap().content().as_message().unwrap().body();
     assert_eq!(content, "C");
 }
+
+#[async_test]
+async fn test_pin_event_is_sent_successfully() {
+    let mut setup = PinningTestSetup::new().await;
+    let timeline = setup.timeline().await;
+
+    setup.mock_sync(false).await;
+    assert!(!timeline.items().await.is_empty());
+
+    // Pinning a remote event succeeds.
+    setup
+        .mock_response(ResponseTemplate::new(200).set_body_json(json!({
+            "event_id": "$42"
+        })))
+        .await;
+
+    let event_id = setup.event_id();
+    assert!(timeline.pin_event(event_id).await.unwrap());
+
+    setup.reset_server().await;
+}
+
+#[async_test]
+async fn test_pin_event_is_returning_false_because_is_already_pinned() {
+    let mut setup = PinningTestSetup::new().await;
+    let timeline = setup.timeline().await;
+
+    setup.mock_sync(true).await;
+    assert!(!timeline.items().await.is_empty());
+
+    let event_id = setup.event_id();
+    assert!(!timeline.pin_event(event_id).await.unwrap());
+
+    setup.reset_server().await;
+}
+
+#[async_test]
+async fn test_pin_event_is_returning_an_error() {
+    let mut setup = PinningTestSetup::new().await;
+    let timeline = setup.timeline().await;
+
+    setup.mock_sync(false).await;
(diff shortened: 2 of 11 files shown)


crates/matrix-sdk-ui/src/timeline/mod.rs
```diff
@@ -870,6 +871,30 @@ impl Timeline {
             Ok(false)
         }
     }
+
+    pub async fn pin_event(&self, event_id: &EventId) -> Result<bool> {
```
andybalaam, line 879: A doc comment explaining the return value would be great.

crates/matrix-sdk-ui/src/timeline/mod.rs
```diff
@@ -870,6 +871,30 @@ impl Timeline {
...
+        if pinned_events.contains(&event_id) {
+            Ok(false)
+        } else {
+            pinned_events.push(event_id);
+            let content = RoomPinnedEventsEventContent::new(pinned_events);
+            self.room().send_state_event(content).await.map(|_| true)
+        }
+    }
+
+    pub async fn unpin_event(&self, event_id: &EventId) -> Result<bool> {
```
andybalaam, line 895: Also doc comment, please.

crates/matrix-sdk-ui/src/timeline/mod.rs
```diff
@@ -870,6 +871,30 @@ impl Timeline {
             Ok(false)
         }
     }
+
+    pub async fn pin_event(&self, event_id: &EventId) -> Result<bool> {
+        let mut pinned_events = self.room().pinned_events().get();
+        let event_id = event_id.to_owned();
```
andybalaam, line 881: I think you can do the `to_owned()` inside the `else` here, saving a copy.

crates/matrix-sdk-ui/src/timeline/mod.rs
```diff
@@ -870,6 +871,30 @@ impl Timeline {
...
+        if pinned_events.contains(&event_id) {
+            Ok(false)
+        } else {
+            pinned_events.push(event_id);
+            let content = RoomPinnedEventsEventContent::new(pinned_events);
+            self.room().send_state_event(content).await.map(|_| true)
```
andybalaam, line 887: I would prefer `?; Ok(true)` instead of `.map(|_| true)` but happy to leave it if you prefer it the way it is.

crates/matrix-sdk-ui/src/timeline/mod.rs
```diff
@@ -870,6 +871,30 @@ impl Timeline {
...
+        }
+    }
+
+    pub async fn unpin_event(&self, event_id: &EventId) -> Result<bool> {
+        let mut pinned_events = self.room().pinned_events().get();
+        let event_id = event_id.to_owned();
```
andybalaam, line 897: Same idea about moving `to_owned()` into `else`.

crates/matrix-sdk-ui/src/timeline/mod.rs
```diff
@@ -870,6 +871,30 @@ impl Timeline {
...
+        if let Some(idx) = pinned_events.iter().position(|e| *e == *event_id) {
+            pinned_events.remove(idx);
+            let content = RoomPinnedEventsEventContent::new(pinned_events);
+            self.room().send_state_event(content).await.map(|_| true)
```
andybalaam, line 901: Same comment about preferring `?; Ok(true)`

crates/matrix-sdk-ui/tests/integration/timeline/mod.rs
```diff
@@ -507,3 +507,98 @@ async fn test_duplicate_maintains_correct_order() {
     let content = items[3].as_event().unwrap().content().as_message().unwrap().body();
     assert_eq!(content, "C");
 }
+
+#[async_test]
+async fn test_pin_event() {
```
andybalaam, line 512: If I understand correctly, this test checks 2 things:

1. When I try to unpin something that is not pinned I receive `false`
2. When I pin something I am allowed to pin I receive `true`

I would prefer a few separate tests, preferably each with very simple Given-When-Then structure (i.e. set up, do something, check results). I often actually write sentences starting with "Given", "When" and "Then" as comments in my test. I find it makes it much easier to read the test and find the part that is actually being tested.

I'd also like to check a little more about what actually happened.

So I might write tests like:

* `pinning_an_event_makes_it_appear_in_the_pinned_list`
* `unpinning_an_event_makes_it_disappear_from_the_pinned_list`
* `trying_to_unpin_a_nonexistent_event_returns_false`
* `trying_to_unpin_an_unpinned_event_returns_false`
* `trying_to_pin_an_event_without_permission_is_an_error`
* `trying_to_unpin_an_event_without_permission_is_an_error`

To avoid repeating code in the tests, I would make the setup code use a function or even a `TestSetup` struct if needed.

I know this is my very opinionated way of writing tests :-) I will defend it enthusiastically at the pub, but will not insist you do it this way in this PR if you are not persuaded.

crates/matrix-sdk-ui/src/timeline/mod.rs
andybalaam: Ah, done later, thanks!

crates/matrix-sdk-ui/src/timeline/mod.rs
```diff
@@ -872,8 +872,12 @@ impl Timeline {
         }
     }
 
+    /// Adds a new pinned event by sending an updated `m.room.pinned_events`
+    /// event containing the new event id.
+    ///
+    /// Returns a boolean indicating if we sent the request or not.
```
andybalaam, line 878: I suggest:

```
/// Returns `true` if we sent the request, `false` otherwise.
```

crates/matrix-sdk-ui/src/timeline/mod.rs
andybalaam: Done later, thanks.

crates/matrix-sdk-ui/src/timeline/mod.rs
```diff
@@ -884,8 +888,12 @@ impl Timeline {
         }
     }
 
+    /// Adds a new pinned event by sending an updated `m.room.pinned_events`
+    /// event without the event id we want to remove.
+    ///
+    /// Returns a boolean indicating if we sent the request or not.
```
andybalaam, line 894: I suggest: 

```
/// Returns `true` if we sent the request, `false` otherwise.
```

andybalaam (changes requested):
Looking really good. A few nits, and a recommendation about tests that you can argue against if you don't like it :-)

crates/matrix-sdk-ui/src/timeline/mod.rs
Velin92: I tried to do that but since `pinned_events.contains` requires a `OwnedEventId`, I would need to do `to_owned()` also in the `contains` which would make for two copies

crates/matrix-sdk-ui/src/timeline/mod.rs
Velin92: same comment as before

crates/matrix-sdk-ui/src/timeline/mod.rs
andybalaam: Cool, ignore me then!

crates/matrix-sdk-ui/tests/integration/timeline/mod.rs
```diff
@@ -510,235 +510,166 @@ async fn test_duplicate_maintains_correct_order() {
...
+
+        Self { event_id, room_id, client, server, sync_settings, sync_builder }
+    }
+
+    async fn timeline(&self) -> matrix_sdk_ui::Timeline {
+        let room = self.client.get_room(self.room_id).unwrap();
+        room.timeline().await.unwrap()
+    }
+
+    async fn reset_server(&self) {
```
andybalaam, line 646: This could go inside an `impl Drop for PinningTestSetup` (I think), which would mean you did not have to remember to call it in the test.

andybalaam (approved):
All looking good: one optional suggestion, then let me know when it's ready to merge!

crates/matrix-sdk-ui/tests/integration/timeline/mod.rs
Velin92: That would be cool, but drop is not async sadly

andybalaam: Ha, of course!

PR #3756, opened 2024-07-23
crypto: Re-use existing get_device_from_curve_key method

andybalaam:
These calls are equivalent because the old code called `self.get_user_devices` with a `timeout` of `None`, which meant the call to `wait_if_user_pending` inside was a no-op.


stefanceriu (approved):
lgtm 👍

PR #3759, opened 2024-07-24
ffi: add high level method for resetting the user's identity and deleting all associated secrets

stefanceriu:
This PR adds support for completely resetting the user's identity and all their secrets. It will go through the following steps:

1. Disable backing up room keys and delete the active backup
2. Disable recovery (We can't delete account data events)
3. Remove previously known secrets by creating a new store (?)
4. Go through the cross-signing key reset flow
5. Finally, re-enable key backups only if they were enabled before

- [x] Expose through FFI
- [x] ~~See if creating a new secrets store is necessary~~ Not needed
- [x] ~~Check coverage to see if we need an OIDC specific test as well or whether the existing cross-singing tests suffice~~ Looks good

### DIFF
--- a/bindings/matrix-sdk-ffi/src/encryption.rs
+++ b/bindings/matrix-sdk-ffi/src/encryption.rs
@@ -9,7 +9,7 @@ use thiserror::Error;
 use zeroize::Zeroize;
 
 use super::RUNTIME;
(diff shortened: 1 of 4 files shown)


poljar (changes requested):

poljar (approved):
Alright, this looks good. Feel free to get rid of the fixup commits and merge.

PR #3760, opened 2024-07-24
dependent events: followup

bnjbvr:
I'm tired. Here are some improvements/bugfixes for the dependent event mechanism.


crates/matrix-sdk-base/src/store/traits.rs
```diff
@@ -1254,6 +1255,38 @@ pub enum DependentQueuedEventKind {
     Redact,
 }
 
+/// A transaction id identifying a [`DependentQueuedEvent`] rather than its
+/// parent [`QueuedEvent`].
+///
+/// This thins wrapper adds some safety to some APIs, making it possible to
```
andybalaam, line 1261: thins -> thin

crates/matrix-sdk/src/send_queue.rs
```diff
@@ -1145,24 +1145,41 @@ impl SendHandle {
...
+
+        if prevs.iter().any(|prev| matches!(prev.kind, DependentQueuedEventKind::Redact)) {
+            // The event has already been flagged for redaction, don't consider the other
+            // dependent events.
+            continue;
+        }
+
         match &d.kind {
-            DependentQueuedEventKind::Edit { .. } => latest_edit = Some(d),
+            DependentQueuedEventKind::Edit { .. } => {
```
andybalaam, line 1174: Maybe a comment here like

`// Replace any previous edit with this one`

crates/matrix-sdk/src/send_queue.rs
```diff
@@ -1145,24 +1145,41 @@ impl SendHandle {
 /// meaningful, i.e. the ones that wouldn't be overridden after applying the
 /// others.
 fn canonicalize_dependent_events(dependent: &[DependentQueuedEvent]) -> Vec<DependentQueuedEvent> {
```
andybalaam, line 1161: So I think previously we assumed everything in `dependent` would be for the same parent and now we don't? Is this fixing an existing bug?

crates/matrix-sdk/src/send_queue.rs
```diff
@@ -924,6 +924,20 @@ impl QueueStorage {
         let num_initial_dependent_events = dependent_events.len();
 
         let canonicalized_dependent_events = canonicalize_dependent_events(&dependent_events);
+
+        // Get rid of the all non-canonical dependent events.
```
andybalaam, line 928: Does this fix an existing bug?

andybalaam (approved):
Looks good, thank you; some questions.

crates/matrix-sdk/src/send_queue.rs
bnjbvr: Yes indeed; this function was already handling all the dependent events for all the queued events, so this was a bug. The new test was failing on main.

crates/matrix-sdk/src/send_queue.rs
bnjbvr: Yes indeed; didn't add a test for it, because it's a bit far fetched and shouldn't happen in theory that we don't have canonicalized events, and would require lots of setup etc. so little value vs what it's testing.

crates/matrix-sdk/src/send_queue.rs
andybalaam: I assume there was not a github issue for it though? It feels weird to fix a real thing and have no audit trail, but I guess creating a whole issue just to close it seems a bit heavy.

crates/matrix-sdk/src/send_queue.rs
andybalaam: No, happy to follow your instincts.

crates/matrix-sdk/src/send_queue.rs
bnjbvr: No github issue, I've found out about the issues while addressing review comments in the PR that introduced the feature, so I figured I could address those in a followup, instead of sneaking in the changes in the approved PR.

bnjbvr:
@andybalaam For what it's worth, I've added a small commit that's a big "duh" from me: the previous PR introduced a transaction id for the dependent event, but forgot to make use of it. This isn't a big deal for edits/redacts, but it would have been a massive failure for reactions. I've addressed it here for edits/redacts, which isn't a big change, but wanted to let you know before I autosquash and merge later.

andybalaam:
All looks good to me, thanks!

PR #3761, opened 2024-07-25
timeline: refactor reaction handling

bnjbvr:
This moves fields around, renames stuff, and changes the way we store reactions:

- instead of relying on the `Reactions::map` *and* storing reactions in `EventTimelineItem` as two sources of truth, reactions are only stored in `EventTimelineItem` now. There's still the need to map from a reaction txn-id | event-id to the thing it's reacting to (e.g. when redacting a reaction), so `map` has been adapted for that usage.
- pending reactions now store all the content they need, without the need for an indirection with the above `Reactions::map`.
- reactions are now keyed by `annotation key` then by `sender id`, instead of `annotation key` then `txn id | event id`. This has the benefit of making it impossible to have two reactions from the same sender. I'm wondering if that could cause issue with local vs remote reactions from the current user, so will ask for further testing.

This should be reviewable commit by commit.


stefanceriu (approved):
Code looks great to me (as much as i can tell) and it also works great, tested it on EX iOS.
Lovely cleanup! 👌

Issue #3742, opened 2024-07-19
Create a release process


poljar:
It's time, I don't think we can realistically afford to not have a proper release process anymore or invent ever more complicated schemes how EW and EX can use the SDK without the SDK having proper releases.

## Tasks

* [x] Get rid of Git dependencies.
* [x]  Automated changelog handling.
* [x] Create release automation.
* [x] Write a release playbook.

## Get rid of Git dependencies

This one is self-explanatory, we use Git dependencies sometimes because we have to, but we'll need to put an end of this practice if we want to stick to release schedule.

We don't have many of them but some have been here for quite some time:

* [x] ruma: https://github.com/matrix-org/matrix-rust-sdk/blob/d65e33ca6af52bbdb55e77f84bf650788486eb28/Cargo.toml#L50
* [x] ruma-common: https://github.com/matrix-org/matrix-rust-sdk/blob/d65e33ca6af52bbdb55e77f84bf650788486eb28/Cargo.toml#L62
* [x] openidconnect: https://github.com/matrix-org/matrix-rust-sdk/blob/d65e33ca6af52bbdb55e77f84bf650788486eb28/crates/matrix-sdk/Cargo.toml#L130
  * [x] Needs https://github.com/poljar/openidconnect-rs/commit/c7e1dc31b83dd7559125984bfd36b9c0f191585e upstreamed
  * [x] Needs https://github.com/poljar/oauth2-rs/commit/f8e28ce5a7f3278ac85b8593ecdd86f2cf51fa2e upstreamed
    * https://github.com/ramosbugs/oauth2-rs/pull/280

The following list is used by the bindings and thus doesn't necessarily, need to be solved since we don't plan on releasing the bindings as a crate. The `openidconnect` crate could be moved to the bindings as well, since it solves a bindings-specific `Send` isssue:

* [ ] async-compat: https://github.com/matrix-org/matrix-rust-sdk/blob/d65e33ca6af52bbdb55e77f84bf650788486eb28/Cargo.toml#L121
* [ ] const_panic: https://github.com/matrix-org/matrix-rust-sdk/blob/d65e33ca6af52bbdb55e77f84bf650788486eb28/Cargo.toml#L122

## Automated changelog handling

For changelogs I suggest we adopt [git-cliff](https://git-cliff.org/). It has support for [monorepos](https://git-cliff.org/docs/usage/monorepos/).

[...]

PR #3762, opened 2024-07-25
dependencies: get rid of custom fork of openidconnect-rs

bnjbvr:
There's a published 4.0.0-alpha.2 version that compiles and doesn't require the custom changes we needed.

Part of #3742.


poljar (approved):
Oh nice, under the assumption that them bindings compile.

PR #3764, opened 2024-07-25
latest event: allow selecting an edit to the previous latest event as a viable latest event

bnjbvr:
It won't work in many cases (e.g. restart the whole SDK, receive many events from sync), but it should handle the simple case where we edit an event and get back to the room list.

Related https://github.com/element-hq/element-x-ios/issues/2683

### DIFF
--- a/crates/matrix-sdk-base/src/latest_event.rs
+++ b/crates/matrix-sdk-base/src/latest_event.rs
@@ -8,10 +8,11 @@ use matrix_sdk_common::deserialized_responses::SyncTimelineEvent;
 use ruma::events::{
     call::{invite::SyncCallInviteEvent, notify::SyncCallNotifyEvent},
     poll::unstable_start::SyncUnstablePollStartEvent,
-    relation::RelationType,
     room::message::SyncRoomMessageEvent,
     AnySyncMessageLikeEvent, AnySyncTimelineEvent,
 };
+#[cfg(feature = "e2e-encryption")]
+use ruma::EventId;
 use ruma::{events::sticker::SyncStickerEvent, MxcUri, OwnedEventId};
 use serde::{Deserialize, Serialize};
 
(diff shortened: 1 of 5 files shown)


crates/matrix-sdk-base/src/latest_event.rs
```diff
@@ -50,26 +51,33 @@ pub enum PossibleLatestEvent<'a> {
...
+pub fn is_suitable_for_latest_event<'a>(
+    event: &'a AnySyncTimelineEvent,
+    prev_latest: Option<&EventId>,
+) -> PossibleLatestEvent<'a> {
+    use ruma::events::room::message::Relation;
+
     match event {
         // Suitable - we have an m.room.message that was not redacted or edited
         AnySyncTimelineEvent::MessageLike(AnySyncMessageLikeEvent::RoomMessage(message)) => {
             // Check if this is a replacement for another message. If it is, ignore it
```
andybalaam, line 63: This comment needs updating (right?)

crates/matrix-sdk-base/src/latest_event.rs
```diff
@@ -50,26 +51,33 @@ pub enum PossibleLatestEvent<'a> {
...
             if let Some(original_message) = message.as_original() {
-                let is_replacement =
-                    original_message.content.relates_to.as_ref().map_or(false, |relates_to| {
-                        if let Some(relation_type) = relates_to.rel_type() {
-                            relation_type == RelationType::Replacement
+                // We include messages that are not replacements; but we allow a replacement if
+                // that's the replacement to the previous known latest event.
+                let is_suitable =
+                    original_message.content.relates_to.as_ref().map_or(true, |relates_to| {
```
andybalaam, line 68: If I read this right, it is equivalent to:

```
let relates_to = original_message.content.relates_to().as_ref();
let is_suitable = if let Some(relates_to) = relates_to {
    // This is a relation: is it replacement of the previous latest event?
    if let Relation::Replacement(c) = &relates_to {
        // It is a replacement: suitable if it replaces the latest event
        prev_latest == Some(&c.event_id)
    else {
        // Some other type of relation: not suitable
        false
    }
} else {
    // Not a relation: definitely suitable
    true
};
```

I find the original version really dense - what do you think about using a more expanded version like the above? I might even make it a separate function...

crates/matrix-sdk-base/src/latest_event.rs
```diff
@@ -531,7 +542,63 @@ mod tests {
...
+
+        assert_let!(
+            PossibleLatestEvent::YesRoomMessage(sync_message_event) =
+                is_suitable_for_latest_event(&event, Some(&prev_latest_event_id))
+        );
+        assert_eq!(sync_message_event.as_original().unwrap().content.body(), "*Bye bye, world!");
+    }
+
+    #[test]
+    fn test_replacement_events_is_suitable_if_doesnt_replace_previous_latest_event() {
```
andybalaam, line 581: is_suitable -> is_not_suitable

crates/matrix-sdk-ui/src/timeline/event_item/content/mod.rs
```diff
@@ -158,29 +159,30 @@ impl TimelineItemContent {
...
+    ) -> TimelineItemContent {
         match event {
             SyncRoomMessageEvent::Original(event) => {
-                // Grab the content of this event
+                // Grab the content of this event.
                 let event_content = event.content.clone();
 
-                // We don't have access to any relations via the `AnySyncTimelineEvent` (I think
-                // - andyb) so we pretend there are none. This might be OK for
-                // the message preview use case.
```
andybalaam, line 175: Woohoo, glad to see this go!

crates/matrix-sdk-ui/src/timeline/event_item/content/mod.rs
```diff
@@ -163,57 +163,46 @@ impl TimelineItemContent {
...
-                // be found inside the timeline_items.
-                let timeline_items = Vector::new();
-
-                TimelineItemContent::Message(Message::from_event(
-                    event_content,
-                    relations,
-                    &timeline_items,
-                ))
-            }
+        let Some(event) = event.as_original() else {
```
andybalaam: Nice change! Note that you could use just `if let` here if you wanted. That would make the whole function one logical line of code, which some people like, but at the cost of one level of indentation.

andybalaam (approved):
Good stuff - a couple of nits and questions. Thanks!

stefanceriu:
Just tested this on iOS: it looks like it should work but in practice those edit just show up as empty `*`. 
Everything else is like the descriptions says.

bnjbvr:
Thanks for the review! This will have to wait for a bit, I need to add more testing because something is super weird there…

bnjbvr:
I'm not satisfied with this technical solution, it was a workaround attempt until we do The Right Thing™ which is to handle the latest event in the event cache, and it's not working as intended, as demonstrated by the latest comment, so I'm closing this for now.

Issue #3754, opened 2024-07-23
Sender Data: Validate SenderData using `Device::is_owner_of_session`


andybalaam:
Part of https://github.com/matrix-org/matrix-rust-sdk/issues/3544

Create a prototype `InboundGroupSession` and pass it in to `SenderDataFinder` so that we can check it is valid via `Device::is_owner_of_session`. Reject the session with an error if it is not.

PR #3765, opened 2024-07-25
crypto: Check the device owns the session in SenderDataFinder

andybalaam:
Fixes https://github.com/matrix-org/matrix-rust-sdk/issues/3754

When we are verifying who sent a session, make sure we do the "backwards" check that the device we have found is actual the owner of the session.

Review commit by commit

### DIFF
--- a/crates/matrix-sdk-crypto/src/olm/group_sessions/inbound.rs
+++ b/crates/matrix-sdk-crypto/src/olm/group_sessions/inbound.rs
@@ -699,10 +699,12 @@ mod tests {
         // And we populated the InboundGroupSession's sender_data with a default value,
         // with legacy_session set to true.
         assert_let!(
-            SenderData::UnknownDevice { retry_details, legacy_session } = unpickled.sender_data
+            SenderData::UnknownDevice { retry_details, legacy_session, owner_check_failed } =
+                unpickled.sender_data
         );
         assert_eq!(retry_details.retry_count, 0);
         assert!(legacy_session);
+        assert!(!owner_check_failed);
     }
 
     #[async_test]
@@ -832,13 +834,14 @@ mod tests {
 
         // And we populated the InboundGroupSession's sender_data with the provided
         // values
-        let SenderData::UnknownDevice { retry_details, legacy_session } = unpickled.sender_data
-        else {
-            panic!("Expected sender_data to be UnknownDevice!");
-        };
+        assert_let!(
+            SenderData::UnknownDevice { retry_details, legacy_session, owner_check_failed } =
+                unpickled.sender_data
+        );
         assert_eq!(retry_details.retry_count, 0);
         assert_eq!(retry_details.next_retry_time_ms.0, UInt::new(98765).unwrap());
         assert!(!legacy_session);
+        assert!(!owner_check_failed);
     }
 
     #[async_test]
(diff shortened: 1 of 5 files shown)


crates/matrix-sdk-crypto/src/olm/group_sessions/sender_data_finder.rs
```diff
@@ -205,13 +217,26 @@ impl<'a> SenderDataFinder<'a> {
     }
 
     /// Step D (we have a device)
+    ///
+    /// Returns Err if the device does not own the session.
     fn have_device(&self, sender_device: Device) -> SenderData {
+        // Is the session owned by the device?
+        let device_is_owner = sender_device.is_owner_of_session(self.session);
```
andybalaam, line 224: This line of code is what this entire PR is about.

crates/matrix-sdk-crypto/src/olm/group_sessions/inbound.rs
```diff
@@ -832,13 +834,15 @@ mod tests {
 
         // And we populated the InboundGroupSession's sender_data with the provided
         // values
-        let SenderData::UnknownDevice { retry_details, legacy_session } = unpickled.sender_data
+        let SenderData::UnknownDevice { retry_details, legacy_session, owner_check_failed } =
+            unpickled.sender_data
         else {
             panic!("Expected sender_data to be UnknownDevice!");
         };
```
poljar, line 841: This one could have used `assert_let` as well.

crates/matrix-sdk-crypto/src/olm/group_sessions/sender_data_finder.rs
```diff
@@ -202,13 +217,26 @@ impl<'a> SenderDataFinder<'a> {
...
+        if let Ok(false) | Err(_) = device_is_owner {
```
poljar, line 225: Can we perhaps not use an early return here? I vastly prefer having each case listed in the same tree instead of having an early return I have to keep in mind. You can either match on both `device_is_owner` and `device_is_cross_signed()` or add an if clause to one of the match cases like so:

```rust
match device_is_owner {
    Ok(true) if sender_device.is_cross_signed_by_owner() => {
        // Is the device cross-signed?
        // Does the cross-signing key match that used to sign the device?
        // And is the signature in the device valid?
        self.device_is_cross_signed_by_sender(sender_device)
    }
    Ok(true) => {
        // No: F (we have device keys, but they are not signed by the sender)
        SenderData::DeviceInfo {
            device_keys: sender_device.as_device_keys().clone(),
            retry_details: SenderDataRetryDetails::retry_soon(),
            legacy_session: true, 
        }
    }
    Ok(false) | Err(_) => {
        // Step E (the device does not own the session)
        // Give up: something is wrong with the session.
        SenderData::UnknownDevice {
            retry_details: SenderDataRetryDetails::retry_soon(),
            legacy_session: true, // TODO: change to false when all SenderData work is done
            owner_check_failed: true,
        }
    }
}
```

poljar (approved):
Left some comment, hopefully you don't mind too much, but I'm allergic to early returns, especially in crypto related code.

Feel free to merge once addressed.

crates/matrix-sdk-crypto/src/olm/group_sessions/inbound.rs
andybalaam: Thanks, no idea what happened to my brain here :-) Fixed in f05f624521

crates/matrix-sdk-crypto/src/olm/group_sessions/sender_data_finder.rs
andybalaam: Good idea, fixed in db784d45d2

Issue #3751, opened 2024-07-23
Sender Data: When we decrypt a message in an InboundGroupSession that is missing SenderData, populate it with what we know


andybalaam:
Part of https://github.com/matrix-org/matrix-rust-sdk/issues/3544

The plan is to have `OlmMachine::get_verification_state` return `SenderData` instead of `VerificationState`.

If we already have `SenderKnown` state, we can just return it instead of calculating each time. If not, we use `FinderSenderData` to try to find a better state. If we find one, save it into the InboundGroupSession and return it.

In order for `FindSenderData` to be equivalent to the existing code, we need to do https://github.com/matrix-org/matrix-rust-sdk/issues/3754 first, so shuffling that one ahead of this.

PR #3769, opened 2024-07-26
crypto: Store device_id in SenderData

andybalaam:
This means we have all the information inside SenderData to populate VerificationStatus and DeviceId for EncryptionInfo, so we can share the code between SenderDataFinder and get_verification_state.

### DIFF
--- a/crates/matrix-sdk-crypto/src/olm/group_sessions/sender_data.rs
+++ b/crates/matrix-sdk-crypto/src/olm/group_sessions/sender_data.rs
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use ruma::{MilliSecondsSinceUnixEpoch, OwnedUserId};
+use ruma::{MilliSecondsSinceUnixEpoch, OwnedDeviceId, OwnedUserId};
 use serde::{Deserialize, Serialize};
 use vodozemac::Ed25519PublicKey;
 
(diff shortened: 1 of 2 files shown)


poljar (approved):

crates/matrix-sdk-crypto/src/olm/group_sessions/sender_data.rs
```diff
@@ -66,6 +66,11 @@ pub enum SenderData {
         /// The user ID of the user who established this session.
         user_id: OwnedUserId,
 
+        /// The device ID of the device that send the session.
+        /// This is an `Option` for backwards compatibility, but we should
+        /// always populate it on creation.
+        device_id: Option<OwnedDeviceId>,
```
andybalaam, line 72: @poljar Clippy is warning that `SenderData` now has `large size difference between variants` because `the largest variant contains at least 225 bytes` since I added the `device_id`.

I think this might be fine, since the `SenderKnown` variant is the "normal" one, so I am inclined to ignore this. What do you think? It would mean a user with lots of legacy sessions or unverified friends would be wasting memory.

poljar: Can't you just box the type?

andybalaam: Yes, do you think it's worth doing?

poljar: Usually I blindly trust clippy, and boxing it is cheap, it's slightly more ugly than not boxing but I think we can live with that.

andybalaam: Done in 7d3e49d798

PR #3771, opened 2024-07-26
sdk: basic support for sending location beacons

torrybr:
This MR introduces basic functionality for sending beacons associated with a live location share, as outlined in the https://github.com/matrix-org/matrix-spec-proposals/pull/3489. This foundational support allows users to send their live location beacon.

**Key Changes**
- `send_location_beacon`: simple support for sending a `beacon` message event with related `beacon_info` event id.

These changes were adapted from a previous MR discussion, available for reference [here](https://github.com/matrix-org/matrix-rust-sdk/pull/3621).

**Follow-up**

This does not save any `beacon` message events with its associated `beacon_info`. This MR simply introduces `beacon` sending support with the intention of follow on MRs to build on this basic functionality.

Signed-off-by:

### DIFF
--- a/crates/matrix-sdk/src/error.rs
+++ b/crates/matrix-sdk/src/error.rs
@@ -521,8 +521,13 @@ pub enum BeaconError {
     #[error("Must join the room to access beacon information.")]
     Stripped,
 
+    // The beacon event could not be deserialized.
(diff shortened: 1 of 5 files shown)


jmartinesp (approved):
LGTM! Thanks for moving the tests to a new file.

PR #3772, opened 2024-07-29
crypto: Replace the implementation of get_verification_state with SenderDataFinder

andybalaam:
Replace the implementation of `get_verification_state` with `SenderDataFinder`, since they are now equivalent.

As part of developing this PR, I created (and then removed) this commit: 282320f79ed2447d2d39f4ca7f60df02dbfb5a2f which proves that for every case, converting from `SenderDataFinder`'s result into a `VerificationState` and an `Option<DeviceId>` produces the same result as the old `get_verification_state` code, so we can safely replace it.

### DIFF
--- a/crates/matrix-sdk-crypto/src/machine.rs
+++ b/crates/matrix-sdk-crypto/src/machine.rs
@@ -1397,60 +1397,39 @@ impl OlmMachine {
         session: &InboundGroupSession,
         sender: &UserId,
     ) -> MegolmResult<(VerificationState, Option<OwnedDeviceId>)> {
-        let claimed_device =
-            self.store().get_device_from_curve_key(sender, session.sender_key()).await?;
+        let sender_data = SenderDataFinder::find_using_curve_key(
+            self.store(),
+            session.sender_key(),
+            sender,
+            session,
+        )
+        .await?;
 
-        Ok(match claimed_device {
-            None => {
-                // We didn't find a device, no way to know if we should trust the
-                // `InboundGroupSession` or not.
-
-                let link_problem = if session.has_been_imported() {
+        Ok(match sender_data {
+            SenderData::UnknownDevice { owner_check_failed: false, .. } => {
+                let device_link_problem = if session.has_been_imported() {
                     DeviceLinkProblem::InsecureSource
                 } else {
                     DeviceLinkProblem::MissingDevice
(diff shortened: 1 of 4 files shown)


andybalaam:
Pasting the patch for 0d9cf722fb in case it ever disappears.

This code was temporary code to validate the `SenderDataFinder` and `get_verification_state` were equivalent.

[validate-that-SenderData-produces-the-same-resul.patch.txt](https://github.com/user-attachments/files/16411271/validate-that-SenderData-produces-the-same-resul.patch.txt)

andybalaam:
Pasting the patch for 282320f79ed2447d2d39f4ca7f60df02dbfb5a2f in case it ever disappears.

This code was temporary code to validate the `SenderDataFinder` and `get_verification_state` were equivalent.

[validate-that-SenderData-produces-the-same-resul.patch.txt](https://github.com/user-attachments/files/16416187/validate-that-SenderData-produces-the-same-resul.patch.txt)

jmartinesp (approved):
I'm not sure I follow every change, but it overall LGTM.

crates/matrix-sdk-crypto/src/machine.rs
```diff
@@ -1397,60 +1397,39 @@ impl OlmMachine {
...
-                        }
-                    } else {
-                        // The device owner hasn't self-verified its device.
-                        (
-                            VerificationState::Unverified(VerificationLevel::UnsignedDevice),
-                            Some(device_id),
-                        )
-                    }
-                }
+            SenderData::UnknownDevice { owner_check_failed: true, .. } => (
```
poljar, line 1418: Shame that we can't put this into a `From` implementation, would cement down the fact that these two are parts of the same coin.

poljar (approved):

crates/matrix-sdk-crypto/src/machine.rs
andybalaam: Yeah I considered that, and in a future PR I moved it into a function, but it's `(SenderData, is_imported) -> (VerificationState, DeviceId)` so it doesn't quite fit.

poljar: Yeah, I saw that `is_imported` messes everything up :cry:.

PR #3773, opened 2024-07-29
sdk-ui: Add `TimelineFocus::PinnedEvents`

jmartinesp:
## Changes

- Create a new type of `TimelineFocus`, `TimelineFocus::PinnedEvents` that will use the pinned event ids in `RoomInfo` to load the associated events, cache them and sort them chronologically.
- Create `PinnedEventsRoom` trait to be able to provide the necessary data to the new timeline type.
- Change the `is_live` parameter used in the `Timeline` code for a `LiveTimelineUpdatesAllowed` enum to distinguish between timelines that need to process every update (live timeline), those who don't need to process any (the focused one) and those which may need to process *some* items (pinned events one).
- Add tests and a benchmark for the concurrent pinned event loading.

Signed-off-by:

### DIFF
--- a/benchmarks/benches/room_bench.rs
+++ b/benchmarks/benches/room_bench.rs
@@ -1,20 +1,32 @@
+use std::time::Duration;
+
 use criterion::{criterion_group, criterion_main, BenchmarkId, Criterion, Throughput};
-use matrix_sdk::utils::IntoRawStateEventContent;
+use matrix_sdk::{
+    config::SyncSettings,
+    test_utils::{events::EventFactory, logged_in_client_with_server},
+    utils::IntoRawStateEventContent,
+};
 use matrix_sdk_base::{
     store::StoreConfig, BaseClient, RoomInfo, RoomState, SessionMeta, StateChanges, StateStore,
 };
 use matrix_sdk_sqlite::SqliteStateStore;
-use matrix_sdk_test::EventBuilder;
+use matrix_sdk_test::{EventBuilder, JoinedRoomBuilder, StateTestEvent, SyncResponseBuilder};
+use matrix_sdk_ui::{timeline::TimelineFocus, Timeline};
 use ruma::{
     api::client::membership::get_member_events,
     device_id,
     events::room::member::{RoomMemberEvent, RoomMemberEventContent},
-    owned_room_id,
+    owned_room_id, owned_user_id,
     serde::Raw,
-    user_id, OwnedUserId,
+    user_id, EventId, MilliSecondsSinceUnixEpoch, OwnedEventId, OwnedUserId,
 };
+use serde::Serialize;
 use serde_json::json;
 use tokio::runtime::Builder;
+use wiremock::{
+    matchers::{header, method, path, path_regex, query_param, query_param_is_missing},
+    Mock, MockServer, Request, ResponseTemplate,
+};
 
 pub fn receive_all_members_benchmark(c: &mut Criterion) {
     const MEMBERS_IN_ROOM: usize = 100000;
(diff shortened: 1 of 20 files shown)


crates/matrix-sdk-ui/src/timeline/builder.rs
```diff
@@ -175,6 +176,27 @@ impl TimelineBuilder {
         let room = inner.room();
         let client = room.client();
 
+        let mut pinned_event_ids_stream = room.pinned_event_ids_stream();
```
andybalaam, line 179: Can this line move inside the `if`?

crates/matrix-sdk-ui/src/timeline/builder.rs
```diff
@@ -239,13 +263,25 @@ impl TimelineBuilder {
                         RoomEventCacheUpdate::AddTimelineEvents { events, origin } => {
                             trace!("Received new timeline events.");
 
-                            inner.add_events_at(
-                                events,
-                                TimelineEnd::Back,
-                                match origin {
-                                    EventsOrigin::Sync => RemoteEventOrigin::Sync,
+                            // Special case for pinned events: when we receive new events what we'll do is
+                            // updated the cache for those events that are pinned and reload the
```
andybalaam, line 267: updated -> update

crates/matrix-sdk-ui/src/timeline/builder.rs
```diff
@@ -239,13 +263,25 @@ impl TimelineBuilder {
...
+                            // list.
+                            match &*focus.clone() {
+                                TimelineFocus::PinnedEvents { .. } => {
```
andybalaam, line 270: So we ignore the events that appeared via `AddTimelineEvents` and just refresh the whole list? Might be good to make this clear in the comment as I didn't quite get it.

crates/matrix-sdk-ui/src/timeline/builder.rs
```diff
@@ -239,13 +263,25 @@ impl TimelineBuilder {
...
 
-                            inner.add_events_at(
-                                events,
-                                TimelineEnd::Back,
-                                match origin {
-                                    EventsOrigin::Sync => RemoteEventOrigin::Sync,
+                            // Special case for pinned events: when we receive new events what we'll do is
+                            // updated the cache for those events that are pinned and reload the
+                            // list.
+                            match &*focus.clone() {
```
andybalaam, line 269: If you can, I would prefer an `if let` here instead of `match`, because the pinned case is currently making the normal case hard to see.

crates/matrix-sdk-ui/src/timeline/pinned_events_loader.rs
```diff
@@ -0,0 +1,221 @@
...
+        for ev_id in pinned_event_ids {
+            if let Some(ev) = self.cache.get(&ev_id).await {
+                loaded_events.push(ev.clone());
+            } else {
+                event_ids_to_request.push(ev_id);
+            }
+        }
+
+        if !event_ids_to_request.is_empty() {
+            let semaphore = Arc::new(Semaphore::new(self.max_concurrent_requests));
```
andybalaam, line 60: Can you extract the logic for performing-things-concurrently-with-a-limit into a separate function, and then pass in the list of inputs (event IDs) and a closure for the thing to do? Don't worry if not.

crates/matrix-sdk-ui/src/timeline/pinned_events_loader.rs
```diff
@@ -0,0 +1,221 @@
...
+#[cfg_attr(not(target_arch = "wasm32"), async_trait::async_trait)]
+pub trait PinnedEventsRoom: SendOutsideWasm + SyncOutsideWasm {
+    /// Load a single room event.
+    async fn room_event(&self, event_id: &EventId) -> Result<SyncTimelineEvent, PaginatorError>;
+
+    /// Get the pinned event ids for a room.
+    fn room_pinned_event_ids(&self) -> Vec<OwnedEventId>;
+
+    /// Checks whether an event id is pinned in this room.
+    fn room_is_pinned_event_id(&self, event_id: &EventId) -> bool;
```
andybalaam, line 146: I think just `is_pinned_event_id` or `is_pinned_event`, or even `is_pinned`.

andybalaam: (Basically I think `room` is not helpful here.

crates/matrix-sdk-ui/src/timeline/pinned_events_loader.rs
```diff
@@ -0,0 +1,221 @@
...
+}
+
+#[cfg_attr(target_arch = "wasm32", async_trait::async_trait(?Send))]
+#[cfg_attr(not(target_arch = "wasm32"), async_trait::async_trait)]
+pub trait PinnedEventsRoom: SendOutsideWasm + SyncOutsideWasm {
+    /// Load a single room event.
+    async fn room_event(&self, event_id: &EventId) -> Result<SyncTimelineEvent, PaginatorError>;
+
+    /// Get the pinned event ids for a room.
+    fn room_pinned_event_ids(&self) -> Vec<OwnedEventId>;
```
andybalaam, line 143: I suggest `pinned_event_ids`

benchmarks/benches/room_bench.rs
```diff
@@ -99,6 +111,120 @@ pub fn receive_all_members_benchmark(c: &mut Criterion) {
...
+        sync_response_builder.add_joined_room(joined_room_builder).build_json_sync_response();
+    runtime.block_on(mock_sync(&server, response_json, None));
+
+    let sync_settings = SyncSettings::default();
+    runtime.block_on(client.sync_once(sync_settings)).expect("Could not sync");
+    runtime.block_on(server.reset());
+
+    runtime.block_on(
+        Mock::given(method("GET"))
+            .and(path_regex(r"/_matrix/client/r0/rooms/.*/event/.*"))
```
andybalaam, line 155: Is it possible to make the mocked response slow? If so I think it would be helpful, to demonstrate that we are fast because we wait for the responses in parallel.

andybalaam (approved):
Looks good! Some questions.

crates/matrix-sdk-ui/src/timeline/builder.rs
jmartinesp: Ok, I'll try to come up with a better comment to explain this.

crates/matrix-sdk-ui/src/timeline/pinned_events_loader.rs
jmartinesp: It seems like I can't because I'd have to make the closure async and that's unstable.

crates/matrix-sdk-ui/src/timeline/pinned_events_loader.rs
jmartinesp: The problem with these is there already are some existing methods in `Room` that have these names and they access private properties, and then I had to create new methods to expose them in the `PinnedEventsRoom` trait and use them in this file. 

I can't just replace the existing methods because they need access to those private properties to work and they can't have the same name because the compiler will mix them up. Unless there's some way to distinguish between those?

A good example of this is `PinnedEventsRoom::room_pinned_event_ids` and `Room::pinned_event_ids`.

benchmarks/benches/room_bench.rs
jmartinesp: The mocked response already has a delay of `50ms`. When tested agains the sequential approach, the concurrent one was 0.55s vs 5.3s.

crates/matrix-sdk-ui/src/timeline/pinned_events_loader.rs
jmartinesp: Ok, I think I got it with https://github.com/matrix-org/matrix-rust-sdk/pull/3773/commits/237ee9ae0e3bef053cedd621b8a83fd2e9a29e7d.

crates/matrix-sdk-ui/src/timeline/pinned_events_loader.rs
jmartinesp: Same as above, https://github.com/matrix-org/matrix-rust-sdk/pull/3773/commits/237ee9ae0e3bef053cedd621b8a83fd2e9a29e7d.

crates/matrix-sdk-ui/src/timeline/pinned_events_loader.rs
andybalaam: Normally when you want an async closure you can get away with a closure that returns a Future by doing:

```
|args| async { body }
```

Not sure here.

crates/matrix-sdk-ui/src/timeline/pinned_events_loader.rs
andybalaam: Looks good.

benchmarks/benches/room_bench.rs
andybalaam: Perfect, thanks!

crates/matrix-sdk-ui/src/timeline/pinned_events_loader.rs
jmartinesp: I think I'll try doing this in a separate PR, since the team implementing the feature is a bit blocked without it merged.

andybalaam: Yep, fine - thanks.

PR #3763, opened 2024-07-25
WidgetDriver: Add authenticated future update action

toger5:
This PR adds a new widget action to the widget api: `org.matrix.msc4157.update.delayed_event`
This needs a capability and be handled properly by the widget api.
This also adapts to the new wording `future event` -> `delayed event` where ruma uses the new wording.
This PR does not rename its own fields (do we want them to all change? or is it okay if the rust sdk keeps using the delayed event name?

Depends on: https://github.com/ruma/ruma/pull/1883

- [x] Public API changes documented in changelogs (optional)

Signed-off-by:

PR #3778, opened 2024-07-30
chore: Bump the Ruma revision

poljar:


toger5:
Ci fails. This makes sense since the ruma change contains breaking changes for the widget api (in the rust sdk)
So those issues will go away once we merge the related rust-sdk PR: https://github.com/matrix-org/matrix-rust-sdk/pull/3763

poljar:
> Ci fails. This makes sense since the ruma change contains breaking changes for the widget api (in the rust sdk) So those issues will go away once we merge the related rust-sdk PR: #3763

Oh, in that case, the Ruma bump needs to be part of you PR.

toger5:
So I just copy the ruma hash you included here.

poljar:
> So I just copy the ruma hash you included here.

Yup, don't forget to run `cargo check` so the lockfile gets updated as well. You can probably just cherry-pick the patch as well.

PR #3779, opened 2024-07-30
crypto: Store the updated SenderData in the store when we calculate it

andybalaam:
This is the key PR that implements the significant part of https://github.com/matrix-org/matrix-rust-sdk/issues/3751 - lazily updating SenderData inside an InboundGroupSession whenever we decrypt a message.


jmartinesp (approved):
LGTM unless someone with more crypto knowledge wants to double check.

poljar (approved):
Looks good, thanks.

PR #3782, opened 2024-07-31
crypto: Use most-trusted SenderData available when decrypting

andybalaam:
Fixes https://github.com/matrix-org/matrix-rust-sdk/issues/3751

Before this PR: every time we decrypt a message we calculate the VerificationState of the sender (by calculating the SenderData and converting).

After this PR: while decrypting, we check whether there is already SenderData linked to the InboundGroupSession. If there is, and it's "fully trusted" i.e. we have already manually verified the sender, we just use it and don't calculate. If it's not fully trusted, we recalculate SenderData and, if it's more trusted, store it and return it.

When we are de we have already stored SenderData with an InboundGroupSession

### DIFF
--- a/crates/matrix-sdk-crypto/src/olm/group_sessions/sender_data.rs
+++ b/crates/matrix-sdk-crypto/src/olm/group_sessions/sender_data.rs
@@ -12,7 +12,9 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use ruma::{OwnedDeviceId, OwnedUserId};
+use std::cmp::Ordering;
+
+use ruma::{DeviceId, OwnedDeviceId, OwnedUserId, UserId};
 use serde::{Deserialize, Serialize};
 use vodozemac::Ed25519PublicKey;
 
@@ -76,8 +78,7 @@ pub enum SenderData {
 }
 
 impl SenderData {
-    /// Create a [`SenderData`] which contains no device info and will be
-    /// retried soon.
+    /// Create a [`SenderData`] which contains no device info.
     pub fn unknown() -> Self {
         Self::UnknownDevice {
             // TODO: when we have implemented all of SenderDataFinder,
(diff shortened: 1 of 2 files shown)


crates/matrix-sdk-crypto/src/olm/group_sessions/sender_data.rs
```diff
@@ -89,14 +88,68 @@ impl SenderData {
...
+        user_id: &UserId,
+        device_id: &DeviceId,
+        master_key: Ed25519PublicKey,
+        master_key_verified: bool,
+    ) -> Self {
+        Self::SenderKnown {
+            user_id: user_id.to_owned(),
+            device_id: Some(device_id.to_owned()),
+            master_key: Box::new(master_key),
+            master_key_verified,
```
poljar, line 114: I'm starting to ponder if this is not quite as good of a name as we could have. We're calling this `master_key_verified` which would imply that two things need to be true for this field to be true:

1. The master key of the user identity has been signed by our own identity.
2. Our own identity is trusted.

But the finder thing also checks if the device is signed by the users identity as well, so it checks the whole chain `own identity -> other identity -> device` and only if the whole chain is valid and trusted then we set this to true.

crates/matrix-sdk-crypto/src/olm/group_sessions/sender_data.rs
```diff
@@ -89,14 +88,68 @@ impl SenderData {
...
+    /// Returns true if this SenderData represents a greater level of trust than
+    /// the supplied one.
+    ///
+    /// So calling this method on a SenderKnown or DeviceInfo SenderData would
+    /// return true if passed an UnknownDevice as its argument, and a
+    /// SenderKnown with `master_key_verified == true` would return true if
+    /// passed a SenderKnown with `master_key_verified == false`.
+    ///
+    /// Returns false if the two SenderDatas have the same trust level.
+    pub(crate) fn is_more_trusted_than(&self, other: &Self) -> bool {
```
poljar, line 141: Probably overly pendantic, but do you mind of introducing a type for the result of the method instead of using a boolean, this would require the function to be renamed, but that way we don't fold the `==` case into false. We have something similar for the comparison of the quality of inbound group sessions:

https://github.com/matrix-org/matrix-rust-sdk/blob/21efd60dbab1c16f9cec3a77d88429601ef645d7/crates/matrix-sdk-crypto/src/store/mod.rs#L1023-L1038

poljar (commented):
Looks fine, I left a suggestions that I think makes things a bit more explicit.

crates/matrix-sdk-crypto/src/olm/group_sessions/sender_data.rs
andybalaam: Any suggestions for a better name? @BillCarsonFr what do you think?

crates/matrix-sdk-crypto/src/olm/group_sessions/sender_data.rs
andybalaam: Changed to return an `Ordering` in be16e70391a4e2d3665bcb095147637c2b6f19c8 - I think that makes sense here... I created a custom enum and its names were almost identical to `Ordering`...

crates/matrix-sdk-crypto/src/olm/group_sessions/sender_data.rs
```diff
@@ -129,26 +131,29 @@ impl SenderData {
...
+    /// would return `Greater` if passed a `SenderKnown` with
+    /// `master_key_verified == false`.
+    pub(crate) fn compare_trust_level(&self, other: &Self) -> Ordering {
+        self.trust_number().cmp(&other.trust_number())
+    }
+
+    /// Internal function to give a numeric value of how much trust this
+    /// `SenderData` represents. Used to make the implementation of
+    /// compare_trust_level simpler.
+    fn trust_number(&self) -> u8 {
```
poljar, line 150: Nice trick, makes things really obvious.

poljar (approved):

PR #3783, opened 2024-07-31
UI Timeline: Build a ShieldState for an unencrypted event in an encrypted room.

pixlwave:
This PR makes the following changes:
- Add a property to the timeline so that it knows whether the room is encrypted or not.
- Pass this value when building timeline items and use it inside `get_shield` to add a new ShieldState.
- Add tests for the ShieldState of an unencrypted item in both and encrypted and unencrypted timeline.

### DIFF
--- a/crates/matrix-sdk-common/src/deserialized_responses.rs
+++ b/crates/matrix-sdk-common/src/deserialized_responses.rs
@@ -29,6 +29,7 @@ const AUTHENTICITY_NOT_GUARANTEED: &str =
 const UNVERIFIED_IDENTITY: &str = "Encrypted by an unverified user.";
 const UNSIGNED_DEVICE: &str = "Encrypted by a device not verified by its owner.";
 const UNKNOWN_DEVICE: &str = "Encrypted by an unknown or deleted device.";
+pub const SENT_IN_CLEAR: &str = "Sent in clear.";
 
 /// Represents the state of verification for a decrypted message sent by a
 /// device.
--- a/crates/matrix-sdk-ui/src/timeline/builder.rs
+++ b/crates/matrix-sdk-ui/src/timeline/builder.rs
@@ -158,9 +158,17 @@ impl TimelineBuilder {
         let (_, mut event_subscriber) = room_event_cache.subscribe().await?;
 
         let is_live = matches!(focus, TimelineFocus::Live);
+        let is_room_encrypted =
+            room.is_encrypted().await.map_err(|_| Error::UnknownEncryptionState)?;
 
-        let inner = TimelineInner::new(room, focus, internal_id_prefix, unable_to_decrypt_hook)
-            .with_settings(settings);
+        let inner = TimelineInner::new(
+            room,
+            focus,
+            internal_id_prefix,
+            unable_to_decrypt_hook,
+            is_room_encrypted,
+        )
+        .with_settings(settings);
 
         let has_events = inner.init_focus(&room_event_cache).await?;
 
--- a/crates/matrix-sdk-ui/src/timeline/event_item/mod.rs
+++ b/crates/matrix-sdk-ui/src/timeline/event_item/mod.rs
@@ -119,11 +119,20 @@ impl EventTimelineItem {
 
     /// If the supplied low-level `SyncTimelineEvent` is suitable for use as the
     /// `latest_event` in a message preview, wrap it as an `EventTimelineItem`.
+    ///
+    /// **Note:** Timeline items created via this constructor do **not** produce
+    /// the correct ShieldState when calling
+    /// [`get_shield`][EventTimelineItem::get_shield]. This is because they are
+    /// intended for display in the room list which a) is unlikely to show
+    /// shields and b) would incur a significant performance overhead.
     pub async fn from_latest_event(
         client: Client,
         room_id: &RoomId,
         latest_event: LatestEvent,
     ) -> Option<EventTimelineItem> {
+        // TODO: We shouldn't be returning an EventTimelineItem here because we're
+        // starting to diverge on what kind of data we need. The note above is a
+        // potential footgun which could one day turn into a security issue.
         use super::traits::RoomDataProvider;
 
         let SyncTimelineEvent { event: raw_sync_event, encryption_info, .. } =
--- a/crates/matrix-sdk-ui/src/timeline/event_item/mod.rs
+++ b/crates/matrix-sdk-ui/src/timeline/event_item/mod.rs
@@ -21,7 +21,10 @@ use matrix_sdk::{
     send_queue::SendHandle,
     Client, Error,
 };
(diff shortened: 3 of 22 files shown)


crates/matrix-sdk-ui/src/timeline/event_item/mod.rs
```diff
@@ -165,20 +168,32 @@ impl EventTimelineItem {
...
-            profile.map(TimelineDetails::Ready).unwrap_or(TimelineDetails::Unavailable)
+            let sender_profile =
+                profile.map(TimelineDetails::Ready).unwrap_or(TimelineDetails::Unavailable);
+
+            let is_room_encrypted = room.is_encrypted().await.unwrap_or(false);
+
+            (sender_profile, is_room_encrypted)
         } else {
-            TimelineDetails::Unavailable
+            (TimelineDetails::Unavailable, false)
```
pixlwave, line 186: (Same here).

crates/matrix-sdk-ui/src/timeline/builder.rs
```diff
@@ -158,9 +158,17 @@ impl TimelineBuilder {
         let (_, mut event_subscriber) = room_event_cache.subscribe().await?;
 
         let is_live = matches!(focus, TimelineFocus::Live);
-
-        let inner = TimelineInner::new(room, focus, internal_id_prefix, unable_to_decrypt_hook)
-            .with_settings(settings);
+        let is_room_encrypted =
+            room.is_encrypted().await.map_err(|_| Error::UnknownEncryptionState)?;
```
stefanceriu, line 162: I wonder if this is going to slow down the timeline or create weird problems if it fails. I imagine that at this point `is_encrypted` is already fetched and cached so I don't think so.

crates/matrix-sdk-ui/src/timeline/event_item/mod.rs
```diff
@@ -350,13 +368,20 @@ impl EventTimelineItem {
     /// Gets the [`ShieldState`] which can be used to decorate messages in the
     /// recommended way.
     pub fn get_shield(&self, strict: bool) -> Option<ShieldState> {
-        self.encryption_info().map(|info| {
-            if strict {
-                info.verification_state.to_shield_state_strict()
-            } else {
-                info.verification_state.to_shield_state_lax()
+        if self.is_room_encrypted {
```
stefanceriu, line 371: Can we maybe bail out early here and reduce the indendation level if not encrypted?

crates/matrix-sdk-common/src/deserialized_responses.rs
```diff
@@ -29,6 +29,7 @@ const AUTHENTICITY_NOT_GUARANTEED: &str =
 const UNVERIFIED_IDENTITY: &str = "Encrypted by an unverified user.";
 const UNSIGNED_DEVICE: &str = "Encrypted by a device not verified by its owner.";
 const UNKNOWN_DEVICE: &str = "Encrypted by an unknown or deleted device.";
+pub const SENT_IN_CLEAR: &str = "Sent in clear.";
```
stefanceriu, line 32: Not a fan of these and I know you aren't either. Hope we'll find a bit of time to migrate away from them soon.

crates/matrix-sdk-ui/src/timeline/event_item/mod.rs
```diff
@@ -165,20 +168,32 @@ impl EventTimelineItem {
...
             // Fallback to the slow path.
             if profile.is_none() {
                 profile = room.profile_from_user_id(&sender).await;
             }
 
-            profile.map(TimelineDetails::Ready).unwrap_or(TimelineDetails::Unavailable)
+            let sender_profile =
+                profile.map(TimelineDetails::Ready).unwrap_or(TimelineDetails::Unavailable);
+
+            let is_room_encrypted = room.is_encrypted().await.unwrap_or(false);
```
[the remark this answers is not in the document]
stefanceriu: I agree but I wonder if we should even try to fetch it in this case. I reckon there's a higher chance it's not cached at this point and if would slow down rendering the latest event.

crates/matrix-sdk-ui/src/timeline/builder.rs
pixlwave: Yeah this one should be fine from my testing.

crates/matrix-sdk-ui/src/timeline/event_item/mod.rs
```diff
@@ -66,6 +69,11 @@ pub struct EventTimelineItem {
     pub(super) content: TimelineItemContent,
     /// The kind of event timeline item, local or remote.
     pub(super) kind: EventTimelineItemKind,
+    /// Whether or not the event belongs to an encrypted room.
+    ///
+    /// When `None` it is unknown if the room is encrypted and the item won't
+    /// return a ShieldState.
+    pub(super) is_room_encrypted: Option<bool>,
```
poljar, line 76: Why is this an optional? The constructor takes a `bool`, when will this be `None`?

pixlwave: When creating an item further in the `from_latest_event` constructor(?) it is set to `None` to avoid requesting the encryption state from the room list.

pixlwave: Yeah it's the latter, it results in no shield info being available (which is fine at least for Element as we don't show it in the room list).

I would imagine this will need to change for Invisible Crypto but I'd consider that a conversation for another day?

poljar: I certainly hope that we're not showing shields on rooms ever, as that is computationally not feasible to do fast, i.e. when the verification state of your own identity changes then the verification state and thus shield state of all identities changes.

This conversation leads me to believe that the latest event, i.e. the event we're using in the room list should be of a different type from what we use in the timeline, i.e. the room list shouldn't use `EventTimelineItem` because we're starting to diverge on what kind of data we need.

Could you add two things on the constructor `from_latest_event()`:
1.  Some docs that we will never produce a correct shield state if this constructor is used
2. A `TODO: ` item explaining that, due to a), we should move this functionality to another type, so we don't have such a footgun which might turn into a security issue.

pixlwave: Fair, done in https://github.com/matrix-org/matrix-rust-sdk/pull/3783/commits/de502f42c61990e07ed450f84612789e3b8a934d

> I certainly hope that we're not showing shields on rooms ever

I wasn't specifically thinking shields here, but filtering out events that aren't trusted. If we ever start to do this in the timeline we would presumably also need to do the same in the room list. Although maybe by that point we'll be able to drive it from a local event cache or something :)

poljar (approved):
Thanks. Looking good.

PR #3784, opened 2024-07-31
crypto: Don't recalculate SenderData if the sender is known but not verified

andybalaam:
After discussion with @richvdh we realised that the policy we agreed earlier for initial creation of SenderData applies here too: if the sender is known, we stick with that and don't recalculate to find out whether they are verified.

Applies on top of https://github.com/matrix-org/matrix-rust-sdk/pull/3782


poljar (commented):
Doesn't that mean that a unverified message will never turn into a verified one? Are we sure we're fine with that?

poljar (approved):
Alright, looks good then.

PR #3785, opened 2024-07-31
crypto: Mark all new SenderData info as non-legacy

andybalaam:
Since we now have a clear idea of the structure, and anything we create now should be usable in future.

Fixes https://github.com/matrix-org/matrix-rust-sdk/issues/3544

Explanation:

This whole feature (SenderData) is about allowing us to know whether a session was created by a trusted person.

Sessions that we created in the past before this feature existed need to be treated differently: we can't just consider them all untrusted.

The `legacy_session` flag is our way of knowing that this session was created before we introduced the feature, or that we have imported the session from somewhere that does not support storing this information (at time of writing, neither key storage nor import/export persist the SenderData info).

This means that we can choose to hide messages that we don't trust, but still display messages from legacy sessions (if we want to).

Up until now, we have been marking all sessions as legacy, but since the feature is now complete, we can start saying new sessions are not legacy.


richvdh (approved):
LGTM

jmartinesp (approved):

PR #3786, opened 2024-07-31
Shields: Add a code field for easier mapping of the state.

pixlwave:
Following on from the conversation over on https://github.com/element-hq/element-x-ios/pull/3051#discussion_r1680872736 this PR adds a `code` field to the `ShieldState` so that it can be easily mapped by apps that need to provide a translated description to the user.

Note: This PR also adds uniffi to the matrix_sdk_common crate to avoid a duplicate type in the FFI for the code.

### DIFF
--- a/crates/matrix-sdk-common/src/deserialized_responses.rs
+++ b/crates/matrix-sdk-common/src/deserialized_responses.rs
@@ -87,22 +87,24 @@ impl VerificationState {
     pub fn to_shield_state_strict(&self) -> ShieldState {
         match self {
             VerificationState::Verified => ShieldState::None,
-            VerificationState::Unverified(level) => {
-                let red = ShieldStateColor::Red;
-                match level {
-                    VerificationLevel::UnverifiedIdentity | VerificationLevel::UnsignedDevice => {
-                        ShieldState::UnverifiedIdentity { color: red }
+            VerificationState::Unverified(level) => match level {
+                VerificationLevel::UnverifiedIdentity | VerificationLevel::UnsignedDevice => {
+                    ShieldState::Red {
+                        code: ShieldStateCode::UnverifiedIdentity,
+                        message: UNVERIFIED_IDENTITY,
                     }
-                    VerificationLevel::None(link) => match link {
-                        DeviceLinkProblem::MissingDevice => {
-                            ShieldState::UnknownDevice { color: red }
-                        }
-                        DeviceLinkProblem::InsecureSource => {
-                            ShieldState::AuthenticityNotGuaranteed { color: red }
-                        }
-                    },
                 }
-            }
+                VerificationLevel::None(link) => match link {
+                    DeviceLinkProblem::MissingDevice => ShieldState::Red {
+                        code: ShieldStateCode::UnknownDevice,
+                        message: UNKNOWN_DEVICE,
+                    },
+                    DeviceLinkProblem::InsecureSource => ShieldState::Red {
+                        code: ShieldStateCode::AuthenticityNotGuaranteed,
+                        message: AUTHENTICITY_NOT_GUARANTEED,
+                    },
+                },
+            },
         }
     }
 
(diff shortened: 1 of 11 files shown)


BillCarsonFr:
> Draft for now more for discussion, this is an exploration following the conversation over on [element-hq/element-x-ios#3051 (comment)](https://github.com/element-hq/element-x-ios/pull/3051#discussion_r1680872736)
> 
> It makes the `ShieldState` strongly typed so clients don't need to switch over raw strings until a decision is made about how i18n should be handled. The string is still available via `ShieldState::message` (or `message_for_shield_state` in the FFI) so we're not losing anything here, plus the tests now assert that the right message is being chosen.
> 
> If this is considered a valid direction, then I would also like to suggest the idea of renaming `ShieldState` to something like `EncryptionAuthenticity`, considering shields are basically a UI element and it doesn't sound like we'll be using a shield icon for these warnings in Element.
> 
> Note: This PR also adds uniffi to the matrix_sdk_common crate to remove the duplicate mapping in the FFI.

As an alternative I was thinking about having the shield state having two fields, a `code`(machine readable) and a description `human_readable` string.
Like that you could do i18n on the code and if code is unknown just use the raw human_readable.
Could even be done in a non API breaking way for existing clients?

jmartinesp (approved):
LGTM, thanks!

jmartinesp:
@BillCarsonFr does the new implementation work for you?

pixlwave:
Force-pushed to rebase on main and add a missing `code` field in the test from #3788.

crates/matrix-sdk-ui/src/timeline/event_item/mod.rs
```diff
@@ -382,7 +382,10 @@ impl EventTimelineItem {
                     Some(info.verification_state.to_shield_state_lax())
                 }
             }
-            None => Some(ShieldState::Grey { message: SENT_IN_CLEAR }),
+            None => Some(ShieldState::Grey {
```
BillCarsonFr, line 385: Sent in clear in an encrypted room is high level warning, should be red

BillCarsonFr (commented):
So yes, I think I prefer the human/machine readable approach.
I just let a comment regarding the sent in clear

crates/matrix-sdk-ui/src/timeline/event_item/mod.rs
pixlwave: Ah my bad, updated 👍

crates/matrix-sdk-common/src/deserialized_responses.rs
```diff
@@ -181,14 +195,40 @@ pub enum DeviceLinkProblem {
...
+        message: &'static str,
+    },
     /// No shield should be presented.
     None,
 }
 
+/// A machine-readable representation of the authenticity for a `ShieldState`.
+#[derive(Clone, Debug, Deserialize, Serialize, Eq, PartialEq)]
+#[cfg_attr(feature = "uniffi", derive(uniffi::Enum))]
+pub enum ShieldStateCode {
```
BillCarsonFr, line 219: If some client app persist that and then later we add a new code, how will this behave? When deserializing an old version of this enum?

pixlwave: I'm no Rust expert but the below seems to indicate that it will be fine if serialised to JSON:

<details>
<summary>Example written with serde</summary>
```
use serde::{Serialize, Deserialize};
use serde_json;

#[derive(Serialize, Deserialize, Debug)]
enum Color {
    Red,
}


#[derive(Serialize, Deserialize, Debug)]
enum ColorPop {
    Red,
    Green,
}


fn main() {
    let color = Color::Red;
    let json_string = serde_json::to_string(&color).unwrap();
    println!("Serialized color: {}", json_string); // Serialized color: "Red"
    
    let color: ColorPop = serde_json::from_str(&json_string).unwrap();
    println!("Deserialized color: {:?}", color); // Deserialized color: Red
}
```
</details>

How come the ShieldState is serialisable in the first place? Shouldn't the app always ask the SDK for the state to be up to date?

poljar: It might be for the WASM use-case where we sometimes, instead of adding a new type in the bindings which the SDK type gets converted into, we return a JSON object. This is using [serde-wasm-bindgen](https://docs.rs/serde-wasm-bindgen/latest/serde_wasm_bindgen/).

If that's not the case, and nobody needs the implementations, perhaps we should just remove the `Serialize` and `Deserialize` implementations since they might have been cargo culted.

pixlwave: Got it, I didn't know that about WASM 👍

PR #3787, opened 2024-07-31
timeline: re-export `ReactionsByKeyBySender` and `ReactionInfo` types

kevinaboos:
This tiny change allows one to easily name the return type of `EventTimelineItem::reactions()` such that you can use it in a function signature. Same goes for the `ReactionInfo` type.

I think this was likely just a small oversight in recent changes to the reactions API, so hopefully it's not controversial.

Without this, it's impossible to write a function that uses `ReactionsByKeyBySender` or `ReactionInfo` in the signature.

Signed-off-by: Kevin Boos <kevinaboos@gmail.com>


andybalaam (approved):
Great, thanks!

PR #3788, opened 2024-08-01
Timeline: Fix a bug where local echoes were showing unencrypted shields.

pixlwave:
Small fix following on from #3783 as local echoes were always showing as `SentInClear` due to the `encryption_info` not being [available](https://github.com/matrix-org/matrix-rust-sdk/blob/7c5c5a8f307a29ac03d7f6de4b1ed41e526a7e7b/crates/matrix-sdk-ui/src/timeline/event_item/mod.rs#L365).

Will need a rebase after #3786 to add the `code` field (or vice versa).

### DIFF
--- a/crates/matrix-sdk-ui/src/timeline/tests/shields.rs
+++ b/crates/matrix-sdk-ui/src/timeline/tests/shields.rs
@@ -1,10 +1,14 @@
+use assert_matches::assert_matches;
 use eyeball_im::VectorDiff;
 use matrix_sdk_base::deserialized_responses::ShieldState;
-use matrix_sdk_test::{async_test, ALICE};
-use ruma::events::room::message::RoomMessageEventContent;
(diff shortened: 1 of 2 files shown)


crates/matrix-sdk-ui/src/timeline/tests/shields.rs
```diff
@@ -39,3 +43,73 @@ async fn test_sent_in_clear_shield() {
     let shield = item.as_event().unwrap().get_shield(false);
     assert_eq!(shield, Some(ShieldState::Grey { message: "Sent in clear." }));
 }
+
+#[async_test]
+/// Note: This is a regression test for a bug where the SentInClear shield was
```
stefanceriu, line 48: Very nice! 👌

stefanceriu (approved):

PR #3790, opened 2024-08-01
Identity reset backup fixes

stefanceriu:
This PR handles 2 different issues:
1) The normal `backups().disable` method does not delete the remote backup if the backups haven't been setup yet locally
2) and it doesn't throw an error or reset the backup state when that happens

We've introduced a new `delete_backup` method what **will** delete remote backups and that can be used from the identity reset flows to correctly setup the fresh session.

### DIFF
--- a/crates/matrix-sdk/src/encryption/backups/mod.rs
+++ b/crates/matrix-sdk/src/encryption/backups/mod.rs
@@ -203,11 +203,12 @@ impl Backups {
                 self.set_state(BackupState::Unknown);
 
                 info!("Backup successfully disabled and deleted");
+
+                Ok(())
             } else {
                 info!("Backup is not enabled, can't disable it");
+                Err(Error::BackupNotEnabled)
             }
-
-            Ok(())
         };
 
         let result = future.await;
--- a/crates/matrix-sdk/src/encryption/backups/mod.rs
+++ b/crates/matrix-sdk/src/encryption/backups/mod.rs
@@ -163,7 +163,11 @@ impl Backups {
         result
     }
 
-    /// Disable and delete the currently active backup.
+    /// Disable and delete the currently active backup only if previously
+    /// enabled before, otherwise an error will be returned.
+    ///
+    /// For a more aggressive variant see [`Backups::disable_and_delete`] which
+    /// will delete the remote backup without checking the local state.
     ///
     /// # Examples
     ///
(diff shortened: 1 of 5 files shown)


crates/matrix-sdk/src/encryption/backups/mod.rs
```diff
@@ -423,6 +423,39 @@ impl Backups {
         Ok(())
     }
 
+    /// Deletes the current backup version from the server
```
poljar, line 426: I think that this need a bit more of a warning, feel free to add some ⚠️ symbols. We should probably also link to the safer, `disable()` mehod from, and to, this one.

As usual, an doc example wouldn't hurt either.

crates/matrix-sdk/src/encryption/backups/mod.rs
```diff
@@ -423,6 +423,39 @@ impl Backups {
...
+            if let Some(response) = response {
+                self.delete_backup_from_server(response.version).await?;
+            }
+
+            let olm_machine = self.client.olm_machine().await;
+            let olm_machine = olm_machine.as_ref().ok_or(Error::NoOlmMachine)?;
+
+            olm_machine.backup_machine().disable_backup().await?;
+
+            self.set_state(BackupState::Unknown);
```
poljar, line 445: Both the OK and error case go into the `Unknown` state, so you can just have one `set_state()` in the main part of the function, just before you return the `result`.

crates/matrix-sdk/tests/integration/encryption/recovery.rs
```diff
@@ -780,11 +780,15 @@ async fn test_reset_identity() {
     assert_eq!(client.encryption().recovery().state(), RecoveryState::Enabled);
 
     // Disabling backups
-    Mock::given(method("DELETE"))
-        .and(path("_matrix/client/r0/room_keys/version/1"))
+    Mock::given(method("GET"))
```
poljar, line 783: You are not quite testing the happy path if you never need to mock the `DELETE` endpoint. You're returning a 404 on the GET version call, so the SDK concludes that a backup doesn't exist and doesn't need to be deleted.

A valid test, but if we had a test for the happy path, which does need to delete the backup as well.

poljar (changes requested):
Looking good but needs a bit more polish.

poljar (approved):

PR #3794, opened 2024-08-01
sdk: refactor beacon_info tests into their own folder

torrybr:
Move the `beacon_info` tests to their own folder to accommodate the addition of several more future test cases.

Signed-off-by: Torry Brelsford <tb@brehl.dev>


stefanceriu (commented):
Very nice, looks good to me but I do wonder if a new `beacon_info` module was necessary on this level as opposed to bundling it with the existing `beacon` one (which is also using `beacon_info`). 

Also, please fix the sign off.

torrybr:
Your choice on the preferred approach 👍 I can refactor to combine under `beacon` if you want but they are 2 separate event types in the spec if that makes any difference

stefanceriu:
Okay, fair point, let's keep them separate then 👍

stefanceriu (approved):

Issue #1129, opened 2022-10-20
Track whether we've ever verified another user, and later signal when that user becomes unverified for any reason


BillCarsonFr:
We should warn a user when the identity of a previously trusted user is rotated. 

On android we expect to be able to map that in MXCrossSIgningInfo#wasTrustedOnce
On pre-element-R web, this is implemented at https://github.com/matrix-org/matrix-js-sdk/blob/0f08c00c07b9495dd9b8b11f6bc2a8cd0ca105e9/src/crypto/CrossSigning.ts#L62-L67

BillCarsonFr:
See https://github.com/element-hq/element-android/issues/6702

richvdh:
I believe this is part of https://github.com/element-hq/element-meta/issues/2488 ?

richvdh:
In the weekly meeting, @BillCarsonFr said that this is in progress. Updating its state on the sprint board.

PR #3795, opened 2024-08-02
Crypto | Add support for verified identities change detection.

BillCarsonFr:
Fixes https://github.com/matrix-org/matrix-rust-sdk/issues/1129

Signed-off-by:

### DIFF
--- a/crates/matrix-sdk-crypto/src/identities/manager.rs
+++ b/crates/matrix-sdk-crypto/src/identities/manager.rs
@@ -39,7 +39,7 @@ use crate::{
         Result as StoreResult, Store, StoreCache, UserKeyQueryResult,
     },
     types::{CrossSigningKey, DeviceKeys, MasterPubkey, SelfSigningPubkey, UserSigningPubkey},
-    CryptoStoreError, LocalTrust, SignatureError,
+    CryptoStoreError, LocalTrust, OwnUserIdentity, SignatureError, UserIdentities,
 };
 
 enum DeviceChange {
@@ -83,6 +83,13 @@ struct KeysQueryRequestDetails {
     /// more actual KeysQueryRequests, each with their own request id. We
     /// record the outstanding request ids here.
     request_ids: HashSet<OwnedTransactionId>,
+}
+
+// Helper type to handle key query response
+struct KeySetInfo {
+    user_id: OwnedUserId,
+    master_key: MasterPubkey,
+    self_signing: SelfSigningPubkey,
 }
 
 impl IdentityManager {
@@ -444,6 +451,7 @@ impl IdentityManager {
     async fn handle_changed_identity(
         &self,
         response: &KeysQueryResponse,
+        maybe_verified_own_identity: Option<&OwnUserIdentity>,
         master_key: MasterPubkey,
         self_signing: SelfSigningPubkey,
         i: UserIdentityData,
@@ -461,7 +469,12 @@ impl IdentityManager {
                 }
             }
             UserIdentityData::Other(mut identity) => {
-                let has_changed = identity.update(master_key, self_signing)?;
+                let has_changed = identity.update(
+                    master_key,
+                    self_signing,
+                    maybe_verified_own_identity.map(|o| o.user_signing_key()),
+                )?;
+
                 if has_changed {
                     Ok(IdentityUpdateResult::Updated(identity.into()))
                 } else {
@@ -502,11 +515,13 @@ impl IdentityManager {
     async fn handle_new_identity(
         &self,
         response: &KeysQueryResponse,
+        maybe_verified_own_identity: Option<&OwnUserIdentity>,
         master_key: MasterPubkey,
         self_signing: SelfSigningPubkey,
         changed_private_identity: &mut Option<PrivateCrossSigningIdentity>,
     ) -> Result<UserIdentityData, SignatureError> {
         if master_key.user_id() == self.user_id() {
+            // Own identity
             let user_signing = self.get_user_signing_key_from_response(response)?;
             let identity = OwnUserIdentityData::new(master_key, self_signing, user_signing)?;
             *changed_private_identity = self.check_private_identity(&identity).await;
@@ -514,6 +529,13 @@ impl IdentityManager {
         } else {
             // First time seen, create the identity. The current MSK will be pinned.
             let identity = OtherUserIdentityData::new(master_key, self_signing)?;
+            let is_verified = maybe_verified_own_identity.map_or(false, |own_user_identity| {
+                own_user_identity.is_identity_signed(&identity).is_ok()
+            });
+            if is_verified {
+                identity.mark_as_verified_once();
+            }
+
             Ok(identity.into())
         }
     }
@@ -616,10 +638,9 @@ impl IdentityManager {
     /// * `changed_identity` - Output parameter: Unchanged if the identity is
     ///   that of another user. If it is our own, set to `None` or `Some`
     ///   depending on whether our stored private identity needs updating.
-    /// * `user_id` - The user id of the user whose identity is being processed.
-    /// * `master_key` - The public master cross-signing key for this user from
-    ///   the `/keys/query` response.
-    /// * `self_signing` - The public self-signing key from the `/keys/query`
+    /// * `maybe_verified_own_identity` - Own verified identity if any to check
+    ///   verification status of updated identity.
+    /// * `key_set_info` - The identity info as returned by the `/keys/query`
     ///   response.
     #[instrument(skip_all, fields(user_id))]
     async fn update_or_create_identity(
@@ -627,17 +648,18 @@ impl IdentityManager {
         response: &KeysQueryResponse,
         changes: &mut IdentityChanges,
         changed_private_identity: &mut Option<PrivateCrossSigningIdentity>,
-        user_id: &UserId,
-        master_key: MasterPubkey,
-        self_signing: SelfSigningPubkey,
+        maybe_verified_own_identity: Option<&OwnUserIdentity>,
+        key_set_info: KeySetInfo,
     ) -> StoreResult<()> {
+        let KeySetInfo { user_id, master_key, self_signing } = key_set_info;
         if master_key.user_id() != user_id || self_signing.user_id() != user_id {
             warn!(?user_id, "User ID mismatch in one of the cross signing keys");
-        } else if let Some(i) = self.store.get_user_identity(user_id).await? {
+        } else if let Some(i) = self.store.get_user_identity(&user_id).await? {
             // an identity we knew about before, which is being updated
             match self
                 .handle_changed_identity(
                     response,
+                    maybe_verified_own_identity,
                     master_key,
                     self_signing,
                     i,
@@ -660,7 +682,13 @@ impl IdentityManager {
         } else {
             // an identity we did not know about before
             match self
-                .handle_new_identity(response, master_key, self_signing, changed_private_identity)
+                .handle_new_identity(
+                    response,
+                    maybe_verified_own_identity,
+                    master_key,
+                    self_signing,
+                    changed_private_identity,
+                )
                 .await
             {
                 Ok(identity) => {
@@ -698,6 +726,15 @@ impl IdentityManager {
         let mut changes = IdentityChanges::default();
         let mut changed_identity = None;
 
+        // We want to check if the updated/new other identities are trusted by us or
+        // not. This is based on the current verified state of the own identity.
+        let maybe_own_verified_identity = self
+            .store
+            .get_identity(self.user_id())
+            .await?
+            .and_then(UserIdentities::own)
+            .filter(|own| own.is_verified());
+
         for (user_id, master_key) in &response.master_keys {
             // Get the master and self-signing key for each identity; those are required for
             // every user identity type. If we don't have those we skip over.
@@ -707,13 +744,14 @@ impl IdentityManager {
                 continue;
             };
 
+            let key_set_info = KeySetInfo { user_id: user_id.clone(), master_key, self_signing };
+
             self.update_or_create_identity(
                 response,
                 &mut changes,
                 &mut changed_identity,
-                user_id,
-                master_key,
-                self_signing,
+                maybe_own_verified_identity.as_ref(),
+                key_set_info,
             )
             .await?;
         }
@@ -1353,6 +1391,7 @@ pub(crate) mod tests {
     use crate::{
         identities::manager::testing::{other_key_query_cross_signed, own_key_query},
         olm::PrivateCrossSigningIdentity,
+        CrossSigningKeyExport, OlmMachine,
     };
 
     fn key_query_with_failures() -> KeysQueryResponse {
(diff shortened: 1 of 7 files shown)


crates/matrix-sdk-crypto/src/identities/user.rs
```diff
@@ -424,13 +425,22 @@ impl UserIdentityData {
...
+/// the user should be notified with a blocking warning.
 #[derive(Debug, Clone, Deserialize, Serialize)]
 #[serde(try_from = "OtherUserIdentityDataSerializer", into = "OtherUserIdentityDataSerializer")]
 pub struct OtherUserIdentityData {
     user_id: OwnedUserId,
     pub(crate) master_key: Arc<MasterPubkey>,
     self_signing_key: Arc<SelfSigningPubkey>,
     pinned_master_key: Arc<RwLock<MasterPubkey>>,
+    /// This tracks whether we've ever verified this user with any identity at
+    /// any point in time on this device. To use it in the future to detect
```
andybalaam, line 441: "at any point in time" ... won't this get reset back to `false` when we "revoke verification", so actually this is not at any point in time - it's more like "we currently think they are verified", right?

crates/matrix-sdk-crypto/src/identities/user.rs
```diff
@@ -424,13 +425,22 @@ impl UserIdentityData {
...
+    /// cases where the user has become unverified for any reason.
+    verified_latch: Arc<AtomicBool>,
```
andybalaam, line 443: Why `_latch`? Does it have a specific meaning here that is helpful?

I would have expected just `verified` maybe.

crates/matrix-sdk-crypto/src/identities/user.rs
```diff
@@ -587,6 +624,29 @@ impl OtherUserIdentityData {
         *m = self.master_key.as_ref().clone()
     }
 
+    /// Remember that this identity used to be verified at some point.
+    pub(crate) fn mark_as_verified_once(&self) {
```
andybalaam, line 627: `mark_as_verified`?

crates/matrix-sdk-crypto/src/identities/user.rs
```diff
@@ -587,6 +624,29 @@ impl OtherUserIdentityData {
...
+        self.verified_latch.store(true, Ordering::SeqCst)
+    }
+
+    /// True if we ever verified this identity (with any own identity, at any
```
andybalaam, line 631: Remove the word "ever"? (similar to other comments)

crates/matrix-sdk-crypto/src/identities/user.rs
```diff
@@ -587,6 +624,29 @@ impl OtherUserIdentityData {
...
+    /// point).
+    ///
+    /// To pass this latch back to false, one must call
+    /// [`OtherUserIdentityData::withdraw_verification()`].
+    pub fn is_verified_latch_set(&self) -> bool {
```
andybalaam, line 636: Maybe just `is_verified` ?

crates/matrix-sdk-crypto/src/identities/manager.rs
```diff
@@ -2023,4 +2023,269 @@ pub(crate) mod tests {
```
andybalaam, line 2062: This commit: "crypto: Verified identity changes TDD" - I would say that the tests for something could be in the same commit as the code itself, so no need for a separate commit here.

If you want a separate commit I suggest "crypto: Tests for verified_latch" or similar.

crates/matrix-sdk-crypto/src/identities/user.rs
andybalaam: Hmm, reading later commits, maybe `was_verified` would be a good name?

crates/matrix-sdk-crypto/src/identities/user.rs
andybalaam: Or, reading later commits, maybe `was_verified`?

crates/matrix-sdk-crypto/src/identities/manager.rs
```diff
@@ -2023,4 +2023,269 @@ pub(crate) mod tests {
...
+
+        let bob_identity =
+            machine.get_identity(DataSet::bob_id(), None).await.unwrap().unwrap().other().unwrap();
+        // The verified latch should be true
+        assert!(bob_identity.is_verified_latch_set());
+        // And bob is verified
+        assert!(bob_identity.is_verified());
+
+        // ######
+        // Second test: Assert that the local latch stays on if the identity is rotated
```
andybalaam, line 2113: I would prefer 2 separate tests but I won't insist here :-)

crates/matrix-sdk-crypto/src/identities/manager.rs
```diff
@@ -2023,4 +2023,269 @@ pub(crate) mod tests {
...
+            .unwrap()
+            .other()
+            .unwrap();
+        // The identity is not verified
+        assert!(!carol_identity.is_verified());
+        // The verified latch is off
+        assert!(!carol_identity.is_verified_latch_set());
+
+        // Carol is verified, likely from another session. Ensure the latch is updated
+        // when the key query response is processes
```
andybalaam, line 2169: processes -> processed

testing/matrix-sdk-test/src/test_json/keys_query_sets.rs
```diff
@@ -667,3 +667,424 @@ impl IdentityChangeDataSet {
             .expect("Can't parse the `/keys/upload` response")
     }
 }
+
+pub struct VerificationLatchTestData {}
+
+#[allow(dead_code)]
```
andybalaam, line 673: Would `#[cfg(test)]` work? If so, it's better than allowing dead code.

crates/matrix-sdk-crypto/src/identities/manager.rs
```diff
@@ -502,18 +515,27 @@ impl IdentityManager {
...
             *changed_private_identity = self.check_private_identity(&identity).await;
             Ok(identity.into())
         } else {
             // First time seen, create the identity. The current MSK will be pinned.
             let identity = OtherUserIdentityData::new(master_key, self_signing)?;
+            let is_verified = maybe_verified_own_identity.map_or(false, |own_user_identity| {
+                own_user_identity.is_identity_signed(&identity).is_ok()
+            });
+            if is_verified {
+                identity.mark_as_verified_once();
```
andybalaam, line 536: Wait, do the tests in the previous commit fail without this? I would like those tests to be in the same commit that makes them pass, or a later one.

andybalaam: Now I see why you called the commit "TDD" - because the tests were written before the code! That is awesome during development, but we don't want commits in the history that fail tests because it is much harder to `git bisect` to find what caused a problem.

andybalaam (dismissed):
This looks good for the parts I understand, but I need some support from someone with more crypto knowledge, especially for 53b04844d4c5be2ac338ebc83a825e591a49e254 . @poljar sorry to ask, but can you take a look?

crates/matrix-sdk-crypto/src/identities/user.rs
BillCarsonFr: Yes it will revert only if you call `withdraw_verification`. I will clarify. 
I'll try to add also that it's "local", i.e not shared across devices

crates/matrix-sdk-crypto/src/identities/user.rs
BillCarsonFr: I found the `latch`(electronic) interesting because it convey that it stays like that until you actively reset it.
In opposition to, only the previous identity was verified. So maybe `was_verified_once`?

andybalaam: `previously_verified` is the closest I can get. If you like `latch` I can live with it.

poljar: Err, let's not use latch, nobody will understand this without reading the doc comment, and then they'll be decently confused why not a more straight forward name was picked.

crates/matrix-sdk-crypto/src/identities/user.rs
BillCarsonFr: If we call it `mark_as_verified` it would be confusing I think. This is not marking the identity as verified (like it can be trusted from this point to send key). It's saying that this identity is expected to be verified starting this point.

Not sure what would be a good name :/ 
I thought of `pin_verification`, but can be `should_be_verified`?

BillCarsonFr: To be consistent with the `previously_verified` suggestion, I'll rename it to `mark_as_previously_verified`

crates/matrix-sdk-crypto/src/identities/user.rs
BillCarsonFr: Will use `was_previously_verified`

testing/matrix-sdk-test/src/test_json/keys_query_sets.rs
BillCarsonFr: No it's not, if I add it then the crypto crate won't compile. If I understand well the `test` config doesn't propagate to other crates, so the annotated code here won't be accessible. I thinking that's why there is the `testing` cfg? Not sure how to make it work

crates/matrix-sdk-crypto/src/identities/user.rs
BillCarsonFr: Using `previously_verified` as per suggestion

crates/matrix-sdk-crypto/src/identities/user.rs
BillCarsonFr: Updated

crates/matrix-sdk-crypto/src/identities/manager.rs
BillCarsonFr: So now, as I can't force push, I have to wait the end of the review, then reset all to re-order the changes and then force push?
Won't all of this squashed at the end?

andybalaam: No, we don't squash when we merge: after review, we force push a sensible set of commits, then merge with "Rebase and merge".

testing/matrix-sdk-test/src/test_json/keys_query_sets.rs
andybalaam: Yes, that is why the `testing` feature exists. It's painful, but better than `#[allow(dead_code)]`. There is an example in `crates/matrix-sdk-crypto/src/lib.rs` in the module called `testing`.

crates/matrix-sdk-crypto/src/store/crypto_store_wrapper.rs
```diff
@@ -109,10 +120,71 @@ impl CryptoStoreWrapper {
             // Mapping the devices and user identities from the read-only variant to one's
             // that contain side-effects requires our own identity. This is
             // guaranteed to be up-to-date since we just persisted it.
-            let own_identity =
+            let maybe_own_identity =
                 self.store.get_user_identity(&self.user_id).await?.and_then(|i| i.into_own());
 
-            let _ = self.identities_broadcaster.send((own_identity, identities, devices));
+            // If our identity was not verified before the change and is now, that means
```
poljar, line 126: This piece of code very much feels like a hack, why aren't we doing the same thing we're doing for other user's identities?

crates/matrix-sdk-crypto/src/identities/manager.rs
```diff
@@ -502,18 +515,27 @@ impl IdentityManager {
...
         if master_key.user_id() == self.user_id() {
+            // Own identity
             let user_signing = self.get_user_signing_key_from_response(response)?;
             let identity = OwnUserIdentityData::new(master_key, self_signing, user_signing)?;
             *changed_private_identity = self.check_private_identity(&identity).await;
             Ok(identity.into())
         } else {
             // First time seen, create the identity. The current MSK will be pinned.
             let identity = OtherUserIdentityData::new(master_key, self_signing)?;
+            let is_verified = maybe_verified_own_identity.map_or(false, |own_user_identity| {
```
poljar, line 532: This isn't checking if the other identity is verified, it's checking if it's signed. If we don't need to check if we're verified then please rename the variables.

crates/matrix-sdk-crypto/src/store/crypto_store_wrapper.rs
```diff
@@ -109,10 +120,71 @@ impl CryptoStoreWrapper {
...
+            // this could impact the verification chain of other known
+            // identities.
+            if let Some(own_identity_after) = maybe_own_identity.as_ref() {
+                // Only do this if our identity is passing from not verified to verified,
+                // the previously_verified can only change in that case.
+                if !own_identity_was_verified_before_change && own_identity_after.is_verified() {
+                    debug!("Own identity is now verified, check all known identities for verification status changes");
+                    // We need to review all the other identities to see if they are verified now
+                    // and mark them as such
+                    self.check_all_identities_and_update_was_previousy_verified_flag_if_needed(
```
poljar, line 136: Thank god for `typos`.
```suggestion
                    self.check_all_identities_and_update_was_previously_verified_flag_if_needed(
```

poljar (changes requested):
Left some questions we should discuss.

crates/matrix-sdk-crypto/src/identities/manager.rs
BillCarsonFr: As per the name `maybe_verified_own_identity` our identity at this point (if any) is verified. So by checking if it is signed I actually check that it is verified

crates/matrix-sdk-crypto/src/store/crypto_store_wrapper.rs
BillCarsonFr: I am not sure to get it. For other identities if there is an update we just need to check the updated identity. But when my own identity trust state change then all the other identities signed by it would be affected. Is this what you mean? 
Also I want only to detect the precise moment the own identity goes from unverified to verified, to only do the checking of all other identities when actually needed.

poljar (approved):
Alright, I still feel like this might produce performance issues and the whole thing might not work out as we expect it, but I don't have concrete advice which will solve this.

testing/matrix-sdk-test/src/test_json/keys_query_sets.rs
BillCarsonFr: There are other #[allow(dead_code)] in this file unrelated to the current changes. Will do a follow-up PR to remove them

poljar:
> comments addressed, I hope

Err, at least this wasn't addressed: https://github.com/matrix-org/matrix-rust-sdk/pull/3795#discussion_r1701725883

richvdh:
Argh. @poljar, @andybalaam, @BillCarsonFr: please remember to document changes in the changelog, *especially* if they are going to cause breaking changes to the cryptostore such that old apps won't run with new stores.

PR #3796, opened 2024-08-02
sdk: add global `PinnedEventCache` instead of having a single instance per timeline

jmartinesp:
## Changes

- Moves the `PinnedEventCache` from `sdk-ui` to `sdk` so it can work in the same way as `EventCache`, attached to a `Client`.
- Remove the `PinnedEventCache` included inside `PinnedEventsLoader`, now you need to provide a cache for the loading event methods.
- Add a way to clear the cached pinned events for a room.
- Add a test to verify this change works as expected.

## Why

Loading pinned events is expensive and slow-ish for rooms with lots of them since we have no bulk `/event` or `/context` endpoint to load a set of events given their event ids, and we need to perform those requests concurrently. To avoid this, a cache was added, but it only worked for events you had already loaded in the same `Timeline` instance so if you exited the room and entered it again you'd still have to reload every single pinned event again.

With the new cache implementation pinned events are kept in memory so reloading them when returning to an already visited room is almost instant. However, this means the events are kept in memory indefinitely so we may want to either:

- Move this cache to a disk-based one. Which will probably be quite complex although probably the most correct way to do this. I have no idea how to do this at the moment, and I'd rather wait until someone with more knowledge about the SDK can tell me if the idea makes any sense.
- Add a way to clear the cache for a room at will: for this PR I've implemented it in `ffi::Room::clear_pinned_events_cache`.

Signed-off-by:

### DIFF
--- a/crates/matrix-sdk-ui/src/timeline/inner/mod.rs
+++ b/crates/matrix-sdk-ui/src/timeline/inner/mod.rs
@@ -354,6 +354,20 @@ impl<P: RoomDataProvider> TimelineInner<P> {
         }
     }
 
+    pub(crate) async fn pinned_events_update(
+        &self,
+        events: Vec<SyncTimelineEvent>,
+        pinned_event_cache: &PinnedEventCache,
+    ) -> Option<Result<Vec<SyncTimelineEvent>, PinnedEventsLoaderError>> {
+        let focus_guard = self.focus.read().await;
+
+        if let TimelineFocusData::PinnedEvents { loader } = &*focus_guard {
+            loader.update_if_needed(events, pinned_event_cache).await
+        } else {
+            Some(Err(PinnedEventsLoaderError::TimelineFocusNotPinnedEvents))
+        }
+    }
+
     /// Run a backward pagination (in focused mode) and append the results to
     /// the timeline.
     ///
--- a/crates/matrix-sdk-ui/src/timeline/inner/mod.rs
+++ b/crates/matrix-sdk-ui/src/timeline/inner/mod.rs
@@ -26,6 +26,7 @@ use matrix_sdk::crypto::OlmMachine;
 use matrix_sdk::{
     deserialized_responses::SyncTimelineEvent,
     event_cache::{paginator::Paginator, RoomEventCache},
+    pinned_events_cache::PinnedEventCache,
     send_queue::{LocalEcho, RoomSendQueueUpdate, SendHandle},
     Result, Room,
 };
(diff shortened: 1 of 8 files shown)


crates/matrix-sdk-ui/src/timeline/pinned_events_loader.rs
```diff
@@ -1,31 +1,27 @@
...
     pub fn new(room: Box<dyn PinnedEventsRoom>, max_events_to_load: usize) -> Self {
-        Self {
-            room: Arc::new(room),
-            max_events_to_load,
-            max_concurrent_requests: 10,
-            cache: PinnedEventCache {
-                inner: Arc::new(InnerPinnedEventCache { events: Default::default() }),
-            },
-        }
+        Self { room: Arc::new(room), max_events_to_load, max_concurrent_requests: 10 }
```
stefanceriu, line 24: Probably best we put these values in separate`const`s

crates/matrix-sdk-ui/src/timeline/inner/mod.rs
```diff
@@ -319,7 +321,10 @@ impl<P: RoomDataProvider> TimelineInner<P> {
             }
 
             TimelineFocusData::PinnedEvents { loader } => {
-                let loaded_events = loader.load_events().await.map_err(Error::PinnedEventsError)?;
+                let loaded_events = loader
+                    .load_events(pinned_event_cache)
```
stefanceriu, line 325: Unsure if this is doable but can't we pass the cache directly to the loader when constructing it?

crates/matrix-sdk/src/pinned_events_cache.rs
```diff
@@ -0,0 +1,59 @@
+use std::{collections::BTreeMap, sync::Arc};
+
+use matrix_sdk_common::deserialized_responses::SyncTimelineEvent;
+use ruma::{EventId, OwnedEventId};
+use tokio::sync::RwLock;
+
+/// Cache used to store the events associated with pinned event ids.
+///
+/// Cloning is shallow, and thus is cheap to do.
```
stefanceriu, line 9: Interesting! 👌

stefanceriu (approved):
Just a couple of small remarks but lgtm! 🚀

crates/matrix-sdk-ui/src/timeline/inner/mod.rs
jmartinesp: Maybe we could make a clone since it'll anyway reuse the `Arc` inside it, but I wanted to keep this working with similar APIs to those of `EventCache`,  and that one is passed around.

crates/matrix-sdk-ui/src/timeline/pinned_events_loader.rs
jmartinesp: https://github.com/matrix-org/matrix-rust-sdk/pull/3796/commits/60a86d7eb4970d84f1d9e0b994c68ecaaa2e9644

jmartinesp: `max_events_to_load` is an argument so clients can provide the max value themselves instead of being limited by the SDK.

Issue #2624, opened 2023-09-26
Implement finer-grained in-memory caching for the crypto store


bnjbvr:
Currently, our cross-process lock blows up the entire crypto state (`OlmMachine`) whenever we notice another process has written to underlying storage. This was implemented because we really wanted a working implementation for decrypting notifications on iOS, but this really is a shortcut. In particular, short-lived crypto procedures (like verification etc) could be interrupted if another process holds the lock in the middle; or we don't currently have any locking when writing a new message (and this might happen in another process too, see also #1960). 

This issue is about fixing and implementing that properly, by having some sort of "cross-process lock guard" representing all the crypto-store data that can be put in a cache. Acquiring this lock may invalidate the cache if another process has written to the underlying database, transparently for the users of the lock. For the point of view of any user of that cache, it's "just" another `async` fallible operation.

---

The idea is to introduce a new `CryptoStoreCache` data structure, which maintains two properties:

- it can only be read by a single process at a time (locked behind the cross-process lock),
- if the lock observed a write to the DB by another process, then the whole cache is invalidated (flushed) since we can't rely on it anymore

The overall "migration" process is thus the following:

- consider each data type that's cached in memory throughout the entire crypto codebase,
- for each of these data types, put it into the `CryptoStoreCache`
- refactor usage so as to limit the number of places that will try to read it from the database (e.g. identify static data that won't change over the course of the app's lifetime, put it aside so that it doesn't need to be invalidated/reloaded)

Alternatively, we could also drop in-memory caching support for some of these fields, if the performance impact is deemed low enough.

Once that's done for all those fields, the final step would be to move back the crypto-store lock into the `OlmMachine` and simplify public APIs again.

A good testing strategy should also be established to make sure we can invalidate the cache at any time, and that it would not fiddle with any of the function calls that may be running in the background.

manuroe:
Just discussed it with @BillCarsonFr. In the quest of killing UTDs, the crypto team is mostly interested by the `Move Olm Sessions into the cache` migration. This is a task they may attack a Q1 but they still want a proof that one problem is coming from here.
We just released UTDs analytics on EX. We need to wait a bit more get more data.

PR #3799, opened 2024-08-02
Remove the Session cache from the cryptostore implementations

poljar:
This PR is another step in the https://github.com/matrix-org/matrix-rust-sdk/issues/2624 epic. It moves the Olm `Session` cache from each individual cryptostore implementation to the `CryptoStoreWrapper`.

While the wrapper shouldn't be the final destination for this cache it does help with our current multi-process support which relied on the `CryptoStore::clear_caches()` method. This method is now gone and no further caches should be added to the store implementations.

Review commit by commit isn't advised as the API changed and changed back when the cache was removed and re-added.

### DIFF
--- a/bindings/matrix-sdk-crypto-ffi/src/lib.rs
+++ b/bindings/matrix-sdk-crypto-ffi/src/lib.rs
@@ -338,10 +338,6 @@ async fn save_changes(
     processed_steps += 1;
     listener(processed_steps, total_steps);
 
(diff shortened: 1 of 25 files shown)


jmartinesp (approved):
Having no clue about the underlying crypto logic the code makes sense to me. LGTM but maybe we want someone else's review here too.

andybalaam (approved):
Looks good.

PR #3800, opened 2024-08-05
crypto: crypto store wrapper migration framework

BillCarsonFr:
The rust-sdk offers an abstraction of the persistence layer via the `CryptoStore`.
There are several implementation, with each of them having their own migration framework to manage their internal data format.
There is some basic migration support at the OlmMachine layer via serde serialization/deserialization. Like for example the use of `#[serde(default)]` that allows to add a default static value for a field added in a json model, or [Serialization helpers](https://github.com/matrix-org/matrix-rust-sdk/blob/fba61751d57f0161b9f2f8f7dec025945edf886f/crates/matrix-sdk-crypto/src/identities/user.rs#L442) like for identity migration.

But this only allows static/basic migration with no access to other existing data.
There is a need to be able to do more in a migration, like seen in [this PR](https://github.com/matrix-org/matrix-rust-sdk/pull/3795), were we are adding a new flag on identities that needs to access the current identity as well as the identity currently migrated. This can be solved by marking all users as dirty, and we have no way to do that currently (appart from doing it in each store implementation?)

This PR adds basic support for such migration.
It impacts a lots of files because the `CryptoStoreWrapper::new` function needs now to be async

Signed-off-by:

### DIFF
--- a/crates/matrix-sdk-crypto/src/dehydrated_devices.rs
+++ b/crates/matrix-sdk-crypto/src/dehydrated_devices.rs
@@ -96,7 +96,7 @@ impl DehydratedDevices {
         let user_identity = self.inner.store().private_identity();
 
         let account = Account::new_dehydrated(user_id);
-        let store = Arc::new(CryptoStoreWrapper::new(user_id, MemoryStore::new()));
+        let store = Arc::new(CryptoStoreWrapper::new(user_id, MemoryStore::new()).await.unwrap());
 
         let verification_machine = VerificationMachine::new(
             account.static_data().clone(),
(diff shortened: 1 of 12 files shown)


bnjbvr:
@BillCarsonFr What's the status of this PR? It's been opened as a draft for a long time now, and it has many conflicts with the existing code. We'd rather not keep PRs as draft for too long, so curious if you have any update on the status here?

BillCarsonFr:
> @BillCarsonFr What's the status of this PR? It's been opened as a draft for a long time now, and it has many conflicts with the existing code. We'd rather not keep PRs as draft for too long, so curious if you have any update on the status here?

WIll close for now. Tracked here https://github.com/matrix-org/matrix-rust-sdk/issues/3811

PR #3803, opened 2024-08-05
Fixes for the `TimelineFocus::PinnedEvents` timeline

jmartinesp:
## Changes

- Use reversed chronological order for the events, as the rest of the timelines do.
- Add `fn sdk::Room::clear_pinned_events_cache`, previously this was only in the FFI layer.
- Fix benchmark and tests after the changes above.

Signed-off-by:

### DIFF
--- a/crates/matrix-sdk-ui/src/timeline/pinned_events_loader.rs
+++ b/crates/matrix-sdk-ui/src/timeline/pinned_events_loader.rs
@@ -36,8 +36,8 @@ impl PinnedEventsLoader {
     /// `max_concurrent_requests` allows, to avoid overwhelming the server.
     ///
     /// It returns a `Result` with either a
-    /// chronologically sorted list of retrieved `SyncTimelineEvent`s  or a
-    /// `PinnedEventsLoaderError`.
+    /// reversed chronologically sorted list of retrieved `SyncTimelineEvent`s
+    /// or a `PinnedEventsLoaderError`.
     pub async fn load_events(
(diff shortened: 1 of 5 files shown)


crates/matrix-sdk-ui/src/timeline/pinned_events_loader.rs
```diff
@@ -104,9 +104,10 @@ impl PinnedEventsLoader {
                 .unwrap_or_else(|_| MilliSecondsSinceUnixEpoch::now())
         }
 
+        // Sort using reversed chronological ordering (newest -> oldest)
         let sorted_events = loaded_events
             .into_iter()
-            .sorted_by(|e1, e2| timestamp(e1).cmp(&timestamp(e2)))
+            .sorted_by(|e1, e2| timestamp(e1).cmp(&timestamp(e2)).reverse())
```
stefanceriu, line 110: Isn't it simpler to do a `.rev()` after sorting all of them the normal way? 
I initially wanted to comment that the `reverse()` should go on the next line and then I realised what it's doing.

stefanceriu (approved):
lgtm 👍

crates/matrix-sdk-ui/src/timeline/pinned_events_loader.rs
Hywan: It doesn't really matter here. You're reversing the `Ord`, the number of sorting operations will be the same.

PR #3804, opened 2024-08-05
sdk-ui: add `PinnedEventsLoaderError::TimelineReloadFailed`

jmartinesp:
This error will be returned when the room has pinned event ids but the timeline couldn't load any of them. 

We need to make this distinction instead of just failing as soon as the first request fails because we don't want an incorrect pinned event id to break loading any other existing events, or have a single failed request with poor connectivity to invalidate the rest of them.

Also fix and add new tests.

Signed-off-by:


andybalaam (approved):
Looks good, thanks!

Issue #3753, opened 2024-07-23
Sender Data: Populate missing SenderData when we receive device info via `/keys/query`


andybalaam:
Part of https://github.com/matrix-org/matrix-rust-sdk/issues/3544

richvdh:
Referring to the flowchart at https://github.com/matrix-org/matrix-rust-sdk/blob/7c5c5a8f307a29ac03d7f6de4b1ed41e526a7e7b/crates/matrix-sdk-crypto/src/olm/group_sessions/sender_data_finder.rs#L36-L124

As I understand things, we need to do the following when handling each device in a `/keys/query` response:

1. Look for `InboundGroupSession`s from the device whose `sender_data` is `UnknownDevice` (ie, state `C` in the flow-chart). For such sessions, we now have the device, and can resume the flowchart at state `D`.
    * XXX: should we exclude sessions where `SenderData::UnknownDevice::owner_check_failed` is true (ie, state `E`)?

      I think it is *possible* for the owner check to pass after it has previously failed (notably, when the device has rotated its curve25519 key, and we receive the megolm key from an Olm session created by the new key before we receive a `/keys/query` response in which the curve25519 key is signed by the ed25519 key), so in theory we should re-check this.

      On the other hand, it's very much an unlikely edge-case, and since there could be a large number of sessions to process, it could be a DoS risk to reprocess all those sessions on each `/keys/query` response for the affected device.
     
      I'm mostly inclined to do whatever is easiest in terms of the database query.

2. If, and only if, the device is now correctly cross-signed (ie, `Device.is_cross_signed_by_owner` is true, and we have the master cross-signing key for the owner), look for `InboundGroupSession`s from the device whose `sender_data` is `DeviceInfo`. We can resume the flowchart at H.

richvdh:
So, in short, I think we need two new methods on cryptostore:
 * `async fn get_unknown_device_inbound_group_sessions_for_device(&self, device_key: Curve25519PublicKey) -> InboundGroupSessionStream`
 * `async fn get_deviceinfo_inbound_group_sessions_for_device(&self, device_key: Curve25519PublicKey) -> InboundGroupSessionStream`

`InboundGroupSessionStream` would, of course, be a type which implements `Stream<Type = InboundGroupSession>`.

We'd need to add indices on (curve_key, sender_data_type) to both the sqlite and indexeddb table implementations to implement this.

poljar:
[...]

PR #3806, opened 2024-08-06
crypto: Allow querying InboundGroupSessions using curve key

### LINKED ISSUE #3752: Allow looking up InboundGroupSessions with missing SenderData by device curve key
Part of https://github.com/matrix-org/matrix-rust-sdk/issues/3544

andybalaam:
This PR adds a new method `CryptoStore::get_inbound_group_sessions_for_device_batch`, which is required for us to be able to update the information we hold about inbound group sessions when we get a `/keys/query` response (https://github.com/matrix-org/matrix-rust-sdk/issues/3753).

Strongly suggest reviewing commit-by-commit!

Fixes #3752

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2542,9 +2542,9 @@ checksum = "ce23b50ad8242c51a442f3ff322d56b02f08852c77e4c0b4d3fd684abc89c683"
 
 [[package]]
 name = "indexed_db_futures"
-version = "0.4.1"
+version = "0.5.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "6cc2083760572ee02385ab8b7c02c20925d2dd1f97a1a25a8737a238608f1152"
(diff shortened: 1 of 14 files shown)


richvdh:
Hrm, with the exception of 2d9fe4dd6e0441b6056907810e22a908b565f004, which I appear to have forgotten about, I think all of this has either landed in the base branch, or been superceded by #3849

richvdh:
Seems like the coverage data is wrong here. It seems like it's treating the text of the SQL statement at https://app.codecov.io/gh/matrix-org/matrix-rust-sdk/pull/3806/blob/crates/matrix-sdk-sqlite/src/crypto_store.rs#L517 as a coverage miss... but clearly that method is being executed, so the SQL statement must be being used.  I wonder if it would help to pull the text out to a separate `const`

richvdh (approved):
@andybalaam I co-wrote this and have co-reviewed it. We're both now happy with the changes.

PR #3807, opened 2024-08-06
ffi: allow the clients to provide a custom `RequestConfig`

jmartinesp:
Without this, the default request config would always be used. This meant having unlimited retries for almost every request, even with no network connection.

Signed-off-by:


stefanceriu (approved):
👌

PR #3808, opened 2024-08-06
sdk-ui: reverse (again) the order in the pinned events timeline

jmartinesp:
This way it matches the rest of timelines in the SDK, I reversed it here because I didn't realise most clients just do this reversal of ordering themselves. As they do, they need the same order for this timeline too to be able to reuse their existing logic.

Signed-off-by:


ganfra (approved):
Tested ok on EXA

PR #3809, opened 2024-08-06
sdk-ui: add `fn Room::event_with_config`

jmartinesp:
This method works the same as `Room::event` but you can provide a custom `RequestConfig` to it.

It's especially useful for the pinned events timeline, since we need a max number of retries and a max number of concurrent requests. With this we can remove some unnecessary complexity.

Related but not depending on https://github.com/matrix-org/matrix-rust-sdk/pull/3807.

Signed-off-by:


stefanceriu (approved):
Very nice, lgtm 👍

Issue #3792, opened 2024-08-01
Element X: Return an error when sharing a room key to a verified user, who has an unverified device 


richvdh:
Part of https://github.com/element-hq/element-meta/issues/2488, and a follow-up to https://github.com/matrix-org/matrix-rust-sdk/issues/1129.

We need to prevent people who have explicitly verified someone accidentally sending messages. `OlmMachine::share_room_key` should check whether any of the target users have previously been verified; if so, and those users have unverified devices, `share_room_key` should return an error code.

We should ensure the error code is propagated up through the ui crate.

This behaviour needs to be optional, so should be configured via `EncryptionSettings` (could be the same setting as #3793)

BillCarsonFr:
As discussed an alternative to that would be to simply ignore unsigned devices of verified users (and send a `m.unverified` withheld code so they know why)?
PRO: Less disruptive than fail to send.
CON: Will silently break communication, without the sender beeing aware of it. Notice that If the recipient properly supports withheld codes he would be able to spot the problem and verify the device.

@pmaier1 any preference?

pmaier1:
I generally like it and think it's better than failing to send. Still, might be very disruptive. But if there's no way to understand why some devices get the message and some don't, then this is not gonna fly. Also, if there's no way to understand for the user on the unverified device, this is a problem and won't work. we need to wait for the full switch to Invisible Crypto then.

If there's no simple solution to this challenge, we have to warn when sending to unverified devices of a verified user, I'm afraid.

richvdh:
See https://github.com/element-hq/element-meta/issues/2488#issuecomment-2271227159 for designs

pixlwave:
For my reference:

> For unverified device: In order to send anyhow, you have to blacklist or whitelist the device using Device::set_local_trust (you have all the problematic devices listed in the error) then just resend⁦

Issue #3793, opened 2024-08-01
Element X: Return an error when sharing a room key to a verified user, who is no longer verified


richvdh:
Part of https://github.com/element-hq/element-meta/issues/2488, and a follow-up to https://github.com/matrix-org/matrix-rust-sdk/issues/1129 and https://github.com/matrix-org/matrix-rust-sdk/issues/3792.

We need to prevent people who have explicitly verified someone accidentally sending messages. `OlmMachine::share_room_key` should check whether any of the target users have previously been verified; if so, and those users are no longer verified, `share_room_key` should return an error code.

We should ensure the error code is propagated up through the ui crate.

This behaviour needs to be optional, so should be configured via `EncryptionSettings` (could be the same setting as #3792).

richvdh:
See https://github.com/element-hq/element-meta/issues/2488#issuecomment-2271227159 for designs

pixlwave:
For my reference:

> For "verification violation" there is a new `UserIdentity::withdraw_verification` that you can call, then retry sending. (you can also blacklist if you prefer via `Device::set_local_trust`)
> ⁦And same the error gaves you all the problematic `user_id`, so you can just iterate, withdraw_verification then resend

PR #3810, opened 2024-08-07
Crypto: Return an error when sharing a room key to a verified user, who has an unverified device (optional encryption setting)

BillCarsonFr:
Fixes https://github.com/matrix-org/matrix-rust-sdk/issues/3792

Suggest review commit-by-commit.

### DIFF
--- a/bindings/matrix-sdk-crypto-ffi/src/lib.rs
+++ b/bindings/matrix-sdk-crypto-ffi/src/lib.rs
@@ -677,7 +677,9 @@ impl From<EncryptionSettings> for RustEncryptionSettings {
             rotation_period: Duration::from_secs(v.rotation_period),
             rotation_period_msgs: v.rotation_period_msgs,
             history_visibility: v.history_visibility.into(),
-            sharing_strategy: CollectStrategy::new_device_based(v.only_allow_trusted_devices),
+            sharing_strategy: CollectStrategy::DeviceBasedStrategy {
+                only_allow_trusted_devices: v.only_allow_trusted_devices,
+            },
         }
     }
 }
--- a/bindings/matrix-sdk-crypto-ffi/src/lib.rs
+++ b/bindings/matrix-sdk-crypto-ffi/src/lib.rs
@@ -668,6 +668,8 @@ pub struct EncryptionSettings {
     /// Should untrusted devices receive the room key, or should they be
     /// excluded from the conversation.
     pub only_allow_trusted_devices: bool,
+    /// Should fail to send when a verified user has unverified devices.
+    pub error_on_verified_user_problem: bool,
 }
 
 impl From<EncryptionSettings> for RustEncryptionSettings {
@@ -679,6 +681,7 @@ impl From<EncryptionSettings> for RustEncryptionSettings {
             history_visibility: v.history_visibility.into(),
             sharing_strategy: CollectStrategy::DeviceBasedStrategy {
                 only_allow_trusted_devices: v.only_allow_trusted_devices,
+                error_on_verified_user_problem: v.error_on_verified_user_problem,
             },
         }
     }
--- a/crates/matrix-sdk-base/src/client.rs
+++ b/crates/matrix-sdk-base/src/client.rs
@@ -28,8 +28,8 @@ use futures_util::Stream;
 use matrix_sdk_common::instant::Instant;
 #[cfg(feature = "e2e-encryption")]
 use matrix_sdk_crypto::{
-    store::DynCryptoStore, EncryptionSettings, EncryptionSyncChanges, OlmError, OlmMachine,
-    ToDeviceRequest,
+    store::DynCryptoStore, CollectStrategy, EncryptionSettings, EncryptionSyncChanges, OlmError,
+    OlmMachine, ToDeviceRequest,
 };
 #[cfg(feature = "e2e-encryption")]
 use ruma::events::{
@@ -1391,7 +1391,13 @@ impl BaseClient {
                 let members = self.store.get_user_ids(room_id, filter).await?;
 
                 let settings = settings.ok_or(Error::EncryptionNotEnabled)?;
-                let settings = EncryptionSettings::new(settings, history_visibility, false);
+                let settings = EncryptionSettings::new(
+                    settings,
+                    history_visibility,
+                    CollectStrategy::DeviceBasedStrategy {
+                        only_allow_trusted_devices: false,
+                    },
+                );
 
                 Ok(o.share_room_key(room_id, members.iter().map(Deref::deref), settings).await?)
             }
--- a/crates/matrix-sdk-base/src/client.rs
+++ b/crates/matrix-sdk-base/src/client.rs
@@ -1396,6 +1396,7 @@ impl BaseClient {
                     history_visibility,
                     CollectStrategy::DeviceBasedStrategy {
                         only_allow_trusted_devices: false,
+                        error_on_verified_user_problem: false,
                     },
                 );
 
--- a/crates/matrix-sdk-crypto/src/error.rs
+++ b/crates/matrix-sdk-crypto/src/error.rs
@@ -12,6 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use std::collections::BTreeMap;
+
 use ruma::{CanonicalJsonError, IdParseError, OwnedDeviceId, OwnedRoomId, OwnedUserId};
 use serde::{ser::SerializeMap, Serializer};
 use serde_json::Error as SerdeError;
@@ -23,6 +25,8 @@ use crate::{
     olm::SessionExportError,
     types::{events::room_key_withheld::WithheldCode, SignedKey},
 };
+#[cfg(doc)]
+use crate::{CollectStrategy, Device, LocalTrust};
 
 pub type OlmResult<T> = Result<T, OlmError>;
 pub type MegolmResult<T> = Result<T, MegolmError>;
@@ -70,6 +74,10 @@ pub enum OlmError {
             have a valid Olm session with us"
     )]
     MissingSession,
+
+    /// Encryption failed due to an error collecting the recipient devices.
+    #[error("encryption failed due to an error collecting the recipient devices: {0}")]
+    SessionRecipientCollectionError(SessionRecipientCollectionError),
 }
 
 /// Error representing a failure during a group encryption operation.
@@ -359,4 +367,22 @@ pub enum SetRoomSettingsError {
     /// The store ran into an error.
     #[error(transparent)]
     Store(#[from] CryptoStoreError),
+}
+
+/// Error representing a problem when collecting the recipient devices for the
+/// room key, during an encryption operation.
+#[derive(Error, Debug)]
+pub enum SessionRecipientCollectionError {
+    /// One or more verified users has one or more unsigned devices.
+    ///
+    /// Happens only with [`CollectStrategy::DeviceBasedStrategy`] when
+    /// [`error_on_verified_user_problem`](`CollectStrategy::DeviceBasedStrategy::error_on_verified_user_problem`)
+    /// is true.
+    ///
+    /// In order to resolve this, the caller can set the trust level of the
+    /// affected devices to [`LocalTrust::Ignored`] or
+    /// [`LocalTrust::BlackListed`] (see [`Device::set_local_trust`]), and
+    /// then retry the encryption operation.
+    #[error("one or more verified users have unsigned devices")]
+    VerifiedUserHasUnsignedDevice(BTreeMap<OwnedUserId, Vec<OwnedDeviceId>>),
 }
--- a/crates/matrix-sdk-crypto/src/identities/device.rs
+++ b/crates/matrix-sdk-crypto/src/identities/device.rs
@@ -297,9 +297,7 @@ impl Device {
         self.device_owner_identity.as_ref().is_some_and(|id| match id {
             UserIdentityData::Own(own_identity) => own_identity.is_verified(),
             UserIdentityData::Other(other_identity) => {
-                self.own_identity.as_ref().is_some_and(|oi| {
-                    oi.is_verified() && oi.is_identity_signed(other_identity).is_ok()
-                })
+                self.own_identity.as_ref().is_some_and(|oi| oi.is_identity_verified(other_identity))
             }
         })
     }
@@ -744,21 +742,19 @@ impl DeviceData {
     ) -> bool {
         own_identity.as_ref().zip(device_owner.as_ref()).is_some_and(
             |(own_identity, device_identity)| {
-                // Our own identity needs to be marked as verified.
-                own_identity.is_verified()
-                    && match device_identity {
-                        // If it's one of our own devices, just check that
-                        // we signed the device.
-                        UserIdentityData::Own(_) => own_identity.is_device_signed(self).is_ok(),
-
-                        // If it's a device from someone else, first check
-                        // that our user has signed the other user and then
-                        // check if the other user has signed this device.
-                        UserIdentityData::Other(device_identity) => {
-                            own_identity.is_identity_signed(device_identity).is_ok()
-                                && device_identity.is_device_signed(self).is_ok()
-                        }
+                match device_identity {
+                    UserIdentityData::Own(_) => {
+                        own_identity.is_verified() && own_identity.is_device_signed(self).is_ok()
                     }
+
+                    // If it's a device from someone else, first check
+                    // that our user has verified the other user and then
+                    // check if the other user has signed this device.
+                    UserIdentityData::Other(device_identity) => {
+                        own_identity.is_identity_verified(device_identity)
+                            && device_identity.is_device_signed(self).is_ok()
+                    }
+                }
             },
         )
     }
--- a/crates/matrix-sdk-crypto/src/machine/tests/mod.rs
+++ b/crates/matrix-sdk-crypto/src/machine/tests/mod.rs
@@ -587,7 +587,10 @@ async fn test_withheld_unverified() {
 
     let encryption_settings = EncryptionSettings::default();
     let encryption_settings = EncryptionSettings {
-        sharing_strategy: CollectStrategy::DeviceBasedStrategy { only_allow_trusted_devices: true },
+        sharing_strategy: CollectStrategy::DeviceBasedStrategy {
+            only_allow_trusted_devices: true,
+            error_on_verified_user_problem: false,
+        },
         ..encryption_settings
     };
 
--- a/crates/matrix-sdk-crypto/src/machine/tests/mod.rs
+++ b/crates/matrix-sdk-crypto/src/machine/tests/mod.rs
@@ -587,7 +587,7 @@ async fn test_withheld_unverified() {
 
     let encryption_settings = EncryptionSettings::default();
     let encryption_settings = EncryptionSettings {
-        sharing_strategy: CollectStrategy::new_device_based(true),
+        sharing_strategy: CollectStrategy::DeviceBasedStrategy { only_allow_trusted_devices: true },
         ..encryption_settings
     };
 
--- a/crates/matrix-sdk-crypto/src/session_manager/group_sessions/share_strategy.rs
+++ b/crates/matrix-sdk-crypto/src/session_manager/group_sessions/share_strategy.rs
@@ -14,6 +14,7 @@
 
 use std::{
     collections::{BTreeMap, BTreeSet, HashMap},
+    default::Default,
     ops::Deref,
 };
 
@@ -236,6 +237,7 @@ fn is_session_overshared_for_user(
     should_rotate
 }
 
+#[derive(Default)]
 struct RecipientDevices {
     allowed_devices: Vec<DeviceData>,
     denied_devices_with_code: Vec<(DeviceData, WithheldCode)>,
@@ -248,23 +250,21 @@ fn split_recipients_withhelds_for_user(
     only_allow_trusted_devices: bool,
 ) -> RecipientDevices {
     // From all the devices a user has, we're splitting them into two
-    // buckets, a bucket of devices that should receive the
-    // room key and a bucket of devices that should receive
+    // buckets: a bucket of devices that should receive the
+    // room key, and a bucket of devices that should receive
     // a withheld code.
-    let (recipients, withheld_recipients): (Vec<DeviceData>, Vec<(DeviceData, WithheldCode)>) =
-        user_devices.into_values().partition_map(|d| {
-            if d.is_blacklisted() {
-                Either::Right((d, WithheldCode::Blacklisted))
-            } else if only_allow_trusted_devices
-                && !d.is_verified(own_identity, device_owner_identity)
-            {
-                Either::Right((d, WithheldCode::Unverified))
-            } else {
-                Either::Left(d)
-            }
-        });
-
-    RecipientDevices { allowed_devices: recipients, denied_devices_with_code: withheld_recipients }
+    let mut recipient_devices: RecipientDevices = Default::default();
+    for d in user_devices.into_values() {
+        if d.is_blacklisted() {
+            recipient_devices.denied_devices_with_code.push((d, WithheldCode::Blacklisted));
+        } else if only_allow_trusted_devices && !d.is_verified(own_identity, device_owner_identity)
+        {
+            recipient_devices.denied_devices_with_code.push((d, WithheldCode::Unverified));
+        } else {
+            recipient_devices.allowed_devices.push(d);
+        }
+    }
+    recipient_devices
 }
 
 fn split_recipients_withhelds_for_user_based_on_identity(
--- a/crates/matrix-sdk-crypto/src/session_manager/group_sessions/share_strategy.rs
+++ b/crates/matrix-sdk-crypto/src/session_manager/group_sessions/share_strategy.rs
@@ -207,32 +207,33 @@ fn is_session_overshared_for_user(
     let recipient_device_ids: BTreeSet<&DeviceId> =
         recipient_devices.iter().map(|d| d.device_id()).collect();
 
-    if let Some(shared) = outbound_session.shared_with_set.read().unwrap().get(user_id) {
-        // Devices that received this session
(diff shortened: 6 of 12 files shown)


crates/matrix-sdk-crypto/src/error.rs
```diff
@@ -70,8 +72,26 @@ pub enum OlmError {
...
+#[derive(Error, Debug)]
+pub enum DeviceCollectError {
+    /// When a verified user has some unsigned devices.
+    /// Happens only with `CollectStrategy::DeviceBasedStrategy` when
+    /// `error_on_unsigned_device_of_verified_users` is true.
+    ///
+    /// In order to resolve this the caller has to ignore or blacklist the given
+    /// devices (see [`Device::set_local_trust`]).
+    #[error("Encryption failed because one or several verified users have unsigned devices")]
+    DeviceBasedVerifiedUserHasUnsignedDevice(BTreeMap<OwnedUserId, Vec<OwnedDeviceId>>),
```
BillCarsonFr, line 93: I am not very good at naming. This error is specific to the DeviceBased strategy so added it has prefix.
The error have all the needed info to resolve the problem, the caller can iterate through the map and blacklist or whitelist the problematic devices

crates/matrix-sdk-crypto/src/identities/device.rs
```diff
@@ -642,6 +642,14 @@ impl DeviceData {
         self.local_trust_state() == LocalTrust::BlackListed
     }
 
+    /// Is the device locally marked as whitelisted.
+    ///
+    /// Whitelisted devices will receive group sessions regardless of their
+    /// verification status.
+    pub fn is_whitelisted(&self) -> bool {
+        self.local_trust_state() == LocalTrust::Ignored
```
BillCarsonFr, line 650: The LocalTrust::Ignored actually is really Whitelist, as it says that the trust state should be ignored. Should it be renamed to Whitelisted? to be consistent with LocalTrust::BlackListed?

crates/matrix-sdk-crypto/src/session_manager/group_sessions/share_strategy.rs
```diff
@@ -209,17 +252,59 @@ pub(crate) async fn collect_session_recipients(
...
+        should_rotate
+    } else {
+        false
+    }
+}
+
 struct RecipientDevices {
     allowed_devices: Vec<DeviceData>,
     denied_devices_with_code: Vec<(DeviceData, WithheldCode)>,
+    unsigned_of_verified_user: Option<Vec<OwnedDeviceId>>,
```
BillCarsonFr, line 297: Initially the 2 sharing strategies were using the same structure to return what devices should or should not receive the key for a given user.
But now with the new option we are adding, the strategies differs on what they do. So this `unsigned_of_verified_user` is specific to device based strategy but this struct is used by both strategy. I put it as an option for now.
I try another alternative that is to return a `Result<RecipientDevices>` instead, then each strategy could do their custom errors.
But then the enclosing code had to aggregate the Error per user into a more global error (to gather all the errors). It seemed a bit of an overkill? so for now going for that. But open to suggestions (given it's going to get worse with more options)

crates/matrix-sdk-crypto/src/session_manager/group_sessions/share_strategy.rs
```diff
@@ -228,16 +313,61 @@ fn split_recipients_withhelds_for_user(
         user_devices.into_values().partition_map(|d| {
             if d.is_blacklisted() {
```
BillCarsonFr, line 314: This if/else sequence is getting harder and harder to read with all the possible settings/outcome.
We might want to get rid of the partition/Either system.
Also the set of settings can create invalid configuration, like `only_allow_trusted_devices`  + `error_on_unsigned_device_of_verified_users` doesn't really make sense.
Will get worse with more options

testing/matrix-sdk-test/src/test_json/keys_query_sets.rs
```diff
@@ -690,9 +690,63 @@ impl PreviouslyVerifiedTestData {
...
-    /// Current user keys query response containing the cross-signing keys
+    pub fn own_unsigned_device_id() -> &'static DeviceId {
+        device_id!("AHIVRZICJK")
+    }
+    pub fn own_signed_device_id() -> &'static DeviceId {
+        device_id!("LCNRWQAVWK")
+    }
+
+    /// Current user keys query response containing the cross-signing keys.
+    /// Contains also two additional devices, one unsigned and one signed
```
richvdh, line 701:
```suggestion
    /// Current user keys query response containing the cross-signing keys.
    /// Contains also two additional devices, one unsigned (with device id 
    /// [`own_unsigned_device_id`]) and one signed (with device id [`own_signed_device_id`]).
```

testing/matrix-sdk-test/src/test_json/keys_query_sets.rs
```diff
@@ -690,9 +690,63 @@ impl PreviouslyVerifiedTestData {
         user_id!("@carol:localhost")
     }
 
-    /// Current user keys query response containing the cross-signing keys
+    pub fn own_unsigned_device_id() -> &'static DeviceId {
+        device_id!("AHIVRZICJK")
+    }
+    pub fn own_signed_device_id() -> &'static DeviceId {
+        device_id!("LCNRWQAVWK")
+    }
```
richvdh, line 698: Could you add some doc-comments to these, explaining what they are, and linking to `own_keys_query_response_1`?

testing/matrix-sdk-test/src/test_json/keys_query_sets.rs
```diff
@@ -946,6 +994,10 @@ impl PreviouslyVerifiedTestData {
             .expect("Can't parse the `/keys/upload` response")
     }
 
+    pub fn carol_unsigned_device_id() -> &'static DeviceId {
+        device_id!("BAZAPVEHGA")
+    }
```
richvdh, line 999: again, could do with some documentation here. It wasn't clear to me that this was related to `device_1_keys_payload_carol`.

richvdh (commented):
Looks broadly good to me. Some initial thoughts, but nothing major -- except my question about checking  `own_identity.is_verified()`.

I'll take a closer look tomorrow.

crates/matrix-sdk-crypto/src/session_manager/group_sessions/share_strategy.rs
```diff
@@ -261,16 +313,61 @@ fn split_recipients_withhelds_for_user(
...
+
+    // If the device owner doesn't have an identity he can't be verified so return
+    // false
+    let owner_identity = match device_owner_identity {
+        Some(o) => o,
+        _ => return false,
+    };
+
+    let is_owner_verified = match owner_identity {
+        UserIdentityData::Other(other_data) => own_identity.is_identity_signed(other_data).is_ok(),
```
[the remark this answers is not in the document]
BillCarsonFr: Good catch, I thought i had a test for that. Let me check my stashes

crates/matrix-sdk-crypto/src/session_manager/group_sessions/share_strategy.rs
```diff
@@ -261,16 +313,61 @@ fn split_recipients_withhelds_for_user(
...
                 && !d.is_verified(own_identity, device_owner_identity)
             {
                 Either::Right((d, WithheldCode::Unverified))
             } else {
+                // track and collect unsigned devices of verified users
+                if error_on_unsigned_device_of_verified_users
+                    && is_unsigned_device_of_verified_user(own_identity, device_owner_identity, &d)
+                {
+                    unsigned_of_verified.push(d.device_id().to_owned())
+                }
```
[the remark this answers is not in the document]
BillCarsonFr: I agree. I tried it but reverted because It was making harder to spot change of behaviors when doing the review.

crates/matrix-sdk-crypto/src/session_manager/group_sessions/share_strategy.rs
```diff
@@ -486,6 +589,425 @@ mod tests {
         assert_eq!(code, &WithheldCode::Unverified);
     }
 
+    #[async_test]
+    async fn test_share_with_per_device_strategy_only_trusted_error_on_unsigned_of_verified() {
```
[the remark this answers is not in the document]
BillCarsonFr: Yes, updated. It's testing there is no unwanted interactions between `only_allow_trusted_devices` and `error_on_unsigned_device_of_verified_users`. Given that we only distribute to trusted devices there is no point in throwing errors for untrusted devices (of verified users).

crates/matrix-sdk-crypto/src/session_manager/group_sessions/share_strategy.rs
```diff
@@ -486,6 +589,425 @@ mod tests {
...
+        let (_, code) = share_result
+            .withheld_devices
+            .iter()
+            .find(|(d, _)| d.device_id() == KeyDistributionTestData::dave_device_id())
+            .expect("This daves's device should receive a withheld code");
+
+        assert_eq!(code, &WithheldCode::Unverified);
+    }
+
+    async fn error_on_unsigned_of_verified_machine_setup() -> OlmMachine {
```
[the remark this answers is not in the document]
BillCarsonFr: Updated https://github.com/matrix-org/matrix-rust-sdk/pull/3810/commits/3fa6891ad8125254d600381000ba914c6136a548

testing/matrix-sdk-test/src/test_json/keys_query_sets.rs
BillCarsonFr: Updated https://github.com/matrix-org/matrix-rust-sdk/pull/3810/commits/3fa6891ad8125254d600381000ba914c6136a548

testing/matrix-sdk-test/src/test_json/keys_query_sets.rs
BillCarsonFr: I added some more comments https://github.com/matrix-org/matrix-rust-sdk/pull/3810/commits/3fa6891ad8125254d600381000ba914c6136a548

There is probably a better way to organize these keys/query response. Would need to put some more thinking on it. But basically there could be a full template response, and one could just add/remove devices as needed, add/remove signatures as needed.
That would allow for more flexible test configuration.

testing/matrix-sdk-test/src/test_json/keys_query_sets.rs
BillCarsonFr: Added some more comments

crates/matrix-sdk-crypto/src/session_manager/group_sessions/share_strategy.rs
```diff
@@ -127,73 +152,91 @@ pub(crate) async fn collect_session_recipients(
...
+                            .or_default()
+                            .extend(blockers)
+                    }
+                }
             }
-            CollectStrategy::IdentityBasedStrategy => {
+
+            if error_on_unsigned_device_of_verified_users
+                && !unsigned_devices_of_verified_users.is_empty()
+            {
```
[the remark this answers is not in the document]
BillCarsonFr: Yes it is. I am hesitant on it. 
By putting it, it makes it very clear for the reader that we only throw if this settings is on.
The fact that in `split_recipients_withhelds_for_user` we only put the device in the `unsigned_devices_of_verified_users`  is more an optimization to avoid checking signatures when it's not needed.

There is another similar optimization thing in this code, like for the `device_owner_identity` we only access the store if `only_allow_trusted_devices || error_on_unsigned_device_of_verified_users` so depending on the setting the identity could be some or None.

So yeah not sure what is best :/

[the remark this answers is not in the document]
richvdh: >  By putting it, it makes it very clear for the reader that we only throw if this settings is on.

I think we can convey this in a comment.

```suggestion
            // `unsigned_devices_of_verified_users` will only be populated if `error_on_unsigned_device_of_verified_users` was set.
            // We should return an error if there are any entries.
            if !unsigned_devices_of_verified_users.is_empty() {
```

For me, it's actually less about optimisation (the compiler can do a better job of optimizing this than I can anyway), and more about emphasising the fact that we don't need to worry about what it means if `unsigned_devices_of_verified_users` has entries but `error_on_unsigned_device_of_verified_users` is false.

testing/matrix-sdk-test/src/test_json/keys_query_sets.rs
richvdh: > There is probably a better way to organize these keys/query response.

Yes, I don't love this system. But we don't need to solve that right now.

crates/matrix-sdk-crypto/src/error.rs
```diff
@@ -70,8 +72,26 @@ pub enum OlmError {
             have a valid Olm session with us"
     )]
     MissingSession,
+
+    #[error(transparent)]
```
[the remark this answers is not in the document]
richvdh: Though in this case, I'm not convinced we want `#[error(transparent)]`, and we probably want 

```
  #[error("encryption failed due to an error collecting the recipients: {0}")]
```

(we should then update the error text on `DeviceBasedVerifiedUserHasUnsignedDevice` to omit the "encryption failed" bit).

crates/matrix-sdk-crypto/src/error.rs
```diff
@@ -70,8 +72,26 @@ pub enum OlmError {
...
+/// room key could fail.
+#[derive(Error, Debug)]
+pub enum DeviceCollectError {
+    /// When a verified user has some unsigned devices.
+    /// Happens only with `CollectStrategy::DeviceBasedStrategy` when
+    /// `error_on_unsigned_device_of_verified_users` is true.
+    ///
+    /// In order to resolve this the caller has to ignore or blacklist the given
+    /// devices (see [`Device::set_local_trust`]).
+    #[error("Encryption failed because one or several verified users have unsigned devices")]
```
richvdh, line 92: For the record: error messages start with a lower-case letter, so that they can be more easily composed.

crates/matrix-sdk-crypto/src/error.rs
```diff
@@ -70,8 +72,26 @@ pub enum OlmError {
...
+    #[error(transparent)]
+    /// The room key that was to be shared was not shared because the sharing
+    /// strategy could not be fulfilled.
+    RoomKeySharingStrategyError(DeviceCollectError),
 }
 
+/// Depending on the sharing strategy for room keys, the distribution of the
+/// room key could fail.
+#[derive(Error, Debug)]
+pub enum DeviceCollectError {
```
richvdh, line 85: I think it would make more sense to rename this, and the enum entry in `OlmError`, to `SessionRecipientCollectionError`, to make it represent the sorts of errors that can happen when finding recipients.

Ultimately we could have `collect_session_recipients` return a `SessionRecipientCollectionError` rather than a generic `OlmError`, but that is a bigger job that should happen another day.

crates/matrix-sdk-crypto/src/error.rs
```diff
@@ -70,8 +72,26 @@ pub enum OlmError {
...
+    /// When a verified user has some unsigned devices.
```
richvdh, line 86:
```suggestion
    /// A verified user has some unsigned devices.
    ///
```

crates/matrix-sdk-crypto/src/session_manager/group_sessions/share_strategy.rs
```diff
@@ -42,6 +46,21 @@ pub enum CollectStrategy {
...
+        /// device:
+        /// - [`Device::set_local_trust`] using [`LocalTrust::Ignored`], this
+        ///   will ignore the warning for this device, and send it the key.
+        /// - [`Device::set_local_trust`] using [`LocalTrust::BlackListed`],
+        ///   this will not distribute the key to this device.
+        ///
+        /// Once the problematic devices are blacklisted or whitelisted the
+        /// caller can retry to share a second time. This has to be done
+        /// once per problematic device.
+        #[serde(default)]
```
richvdh, line 62: is this persisted somewhere, that makes the `default` necessary?

crates/matrix-sdk-crypto/src/session_manager/group_sessions/share_strategy.rs
```diff
@@ -42,6 +46,21 @@ pub enum CollectStrategy {
         ///       trusted via interactive verification.
         ///     - It is the current own device of the user.
         only_allow_trusted_devices: bool,
+        /// If true, when a verified user has an unsigned device the key sharing
+        /// will fail with an error. If false the key will be
+        /// distributed to that unsigned device. In order to resolve
```
richvdh, line 51:
```suggestion
        /// If `true`, and a verified user has an unsigned device, key sharing
        /// will fail with an error. Otherwise, keys are shared with
        /// unsigned devices as normal. 
        ///
        /// In order to resolve
```

crates/matrix-sdk-crypto/src/identities/device.rs
richvdh: yeahh, this feels pretty confusing to me, and could do with clean up. 

That said: "whitelisted"/"blacklisted" are somewhat problematic terms, and we might take the opportunity to rename both.

crates/matrix-sdk-crypto/src/session_manager/group_sessions/share_strategy.rs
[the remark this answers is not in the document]
richvdh: the solution to that is to refactor it in one commit, then change the behaviour in another...

richvdh (commented):
A few more, more nitpicky comments. I still need to take a closer look at the tests.

crates/matrix-sdk-crypto/src/session_manager/group_sessions/share_strategy.rs
BillCarsonFr: `EncryptionSettings` is persisted in the OutboundSession at least, and maybe some clients could persist the encryption settings? All the things we put as serializable could be persisted... (are we serializing things for other reasons? like for bindings to pass things back and forth?)
We really need to improve that part of the sdk, it's too easy to make a change that would break applications.  
I thought about having tests with json exports of all the serializable things that are persisted, that just tries to load them back and see if it breaks. This would be minimum. Or we need to think about a specific trait for things that are expected to be serialized persisted, with a version of some form

richvdh: > `EncryptionSettings` is persisted in the OutboundSession at least

Hrm... does that means that, once we enable the new behaviour, it won't work until megolm sessions are replaced?

BillCarsonFr: I don't think so. But there is something odd with the settings how they are now.

The settings persisted is used mainly (i think) in the code to detect if the session should be rotated.
I think that only part of the `EncryptionSetting` should be persisted (rotation_period, rotation_period_msgs, and history_visibility?).
The old `only_allow_trusted_devices` that was in the EncryptionSetting (replaced now by CollectStrategy) shouldn't be persisted.  Because anyhow if a change on this setting change the set of devices that should receive it, then `is_session_overshared_for_user` will detect it and rotate the session.

`history_visibility` is different because you want to rotate if you decide that this session could be shared now as part of history sharing for example, event if it doesn't change right now who get the key

richvdh: I'm not really following all of this. It sounds like we shouldn't be persisting `EncryptionSettings` at all to me.

BillCarsonFr: If you change the `rotation_period` you would want the session to be rotated immediatly instead of having to wait for the old one to be rotated based on the old rotation period. Same for `history_visibility`, you want to make history sharable from the point you change to that setting and not make early history available?

richvdh: Both of those things sound like we should be taking notice of the *current* `EncryptionSettings`, not whatever was persisted to the store.

bindings/matrix-sdk-crypto-ffi/src/lib.rs
```diff
@@ -668,6 +668,8 @@ pub struct EncryptionSettings {
     /// Should untrusted devices receive the room key, or should they be
     /// excluded from the conversation.
     pub only_allow_trusted_devices: bool,
+    /// Should fail to send when a verified user has unverified devices.
+    pub error_on_verified_user_problem: bool,
```
richvdh, line 672: This is a slightly rubbish name; ultimately I want to use it for both unverified devices (#3792) and verified users that become unverified (#3793).

crates/matrix-sdk-crypto/src/session_manager/group_sessions/share_strategy.rs
```diff
@@ -271,26 +381,56 @@ fn split_recipients_withhelds_for_user_based_on_identity(
...
                 denied_devices_with_code: withheld_recipients,
+                unsigned_of_verified_user: Vec::default(),
             }
         }
     }
 }
 
+fn is_unsigned_device_of_verified_user(
+    own_identity: &Option<OwnUserIdentityData>,
+    device_owner_identity: &Option<UserIdentityData>,
```
bnjbvr, line 392: :crab: nit: usually (and I know there's a clippy lint for this, not enabled in our code base), it's preferable to use `Option<&T>` instead of `&Option<T>` in parameters lists, because that's how we're going to use it later. (Proof in case, you're calling `as_ref()` in the bodies, and following my suggestion it would be called at the call-sites instead.)

crates/matrix-sdk-crypto/src/session_manager/group_sessions/share_strategy.rs
```diff
@@ -271,26 +381,56 @@ fn split_recipients_withhelds_for_user_based_on_identity(
...
+) -> bool {
+    device_owner_identity.as_ref().is_some_and(|device_owner_identity| {
+        is_user_verified(own_identity, device_owner_identity)
+            && !device_data.is_cross_signed_by_owner(device_owner_identity)
+    })
+}
+
+fn is_user_verified(
+    own_identity: &Option<OwnUserIdentityData>,
+    user_identity: &UserIdentityData,
```
bnjbvr, line 403: (here too)

bnjbvr (changes requested):
On request, I've just skimmed the PR and looked at Rust idioms. I've got some small suggestions to improve readability of the code, which I think should be addressed; confusing a bool for another could be particularly bad in the context of crypto, IMO. LGTM otherwise!

crates/matrix-sdk-crypto/src/session_manager/group_sessions/share_strategy.rs
```diff
@@ -209,35 +241,112 @@ pub(crate) async fn collect_session_recipients(
...
+/// Partition the list of a user's devices according to whether they should
+/// receive the key, for [`CollectStrategy::DeviceBasedStrategy`].
+///
+/// We split the list into three buckets:
+///
+///  * the devices that should receive the room key.
+///
+///  * the devices that should receive a withheld code.
+///
+///  * If `error_on_verified_user_problem` is set, the devices that should cause
```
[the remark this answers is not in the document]
richvdh: I don't think that's right: markdown allows blank lines between list items.

I ran it through `cargo doc --document-private-items`, and got HTML that looks like:

image

crates/matrix-sdk-base/src/client.rs
```diff
@@ -1391,7 +1391,7 @@ impl BaseClient {
                 let members = self.store.get_user_ids(room_id, filter).await?;
 
                 let settings = settings.ok_or(Error::EncryptionNotEnabled)?;
-                let settings = EncryptionSettings::new(settings, history_visibility, false);
+                let settings = EncryptionSettings::new(settings, history_visibility, false, false);
```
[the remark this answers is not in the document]
richvdh: Indeed. Given that all we're doing here is populating a struct with a pair of boolean fields, I think we might as well just pass in a pre-built struct, which means that the fields can be explicitly named.

Fixed in ec0df35dee63f8ebb3af12164e9990d09145ae29

crates/matrix-sdk-crypto/src/identities/device.rs
```diff
@@ -744,21 +742,19 @@ impl DeviceData {
...
-                        // If it's a device from someone else, first check
-                        // that our user has signed the other user and then
-                        // check if the other user has signed this device.
-                        UserIdentityData::Other(device_identity) => {
-                            own_identity.is_identity_signed(device_identity).is_ok()
-                                && device_identity.is_device_signed(self).is_ok()
-                        }
+                match device_identity {
+                    UserIdentityData::Own(_) => {
+                        own_identity.is_verified() && own_identity.is_device_signed(self).is_ok()
```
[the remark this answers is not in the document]
richvdh: Agreed. Let's fix that in another PR.

crates/matrix-sdk-crypto/src/machine/tests/mod.rs
```diff
@@ -587,7 +587,7 @@ async fn test_withheld_unverified() {
 
     let encryption_settings = EncryptionSettings::default();
     let encryption_settings = EncryptionSettings {
-        sharing_strategy: CollectStrategy::new_device_based(true),
+        sharing_strategy: CollectStrategy::new_device_based(true, false),
```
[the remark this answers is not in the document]
richvdh: in this case, let's just get rid of the `CollectStrategy::new_device_based` "helper", which is more of an obfuscator.

Fixed in 1a8f0f29c

crates/matrix-sdk-crypto/src/session_manager/group_sessions/share_strategy.rs
richvdh: Yup. Done in b341e19e0.

crates/matrix-sdk-crypto/src/session_manager/group_sessions/share_strategy.rs
richvdh: also done in b341e19e0

bnjbvr (approved):
Super nice, thanks!

PR #3812, opened 2024-08-07
indexeddb: accept any db schema version up to 99

richvdh:
Currently, each time we make an indexeddb schema change, we break backwards compatibility with any existing applications. In other words, if someone updates their Element install, then decides to roll back, all existing sessions are broken because the old application can't cope with the new schema.

We can do better than this by being more tolerant in the database schema versions we accept. If we promise not to break backwards compat without bumping the schema version to ~~20~~ 100, then we can accept all schema versions up to ~~19~~ 99.

(Also contains a warmup commit which fixes something that becomes a problem in my new test.)

### DIFF
--- a/crates/matrix-sdk-indexeddb/src/crypto_store/migrations/mod.rs
+++ b/crates/matrix-sdk-indexeddb/src/crypto_store/migrations/mod.rs
@@ -62,6 +62,40 @@ impl Drop for MigrationDb {
     }
 }
 
+/// The latest version of the schema we can support. If we encounter a database
+/// version with a higher schema version, we will return an error.
+///
+/// A note on how this works.
+///
+/// Normally, when you open an indexeddb database, you tell it the "schema
+/// version" that you know about. If the existing database is older than
+/// that, it lets you run a migration. If the existing database is newer, then
+/// it assumes that there have been incompatible schema changes and complains
+/// with an error ("The requested version (10) is less than the existing version
+/// (11)").
+///
+/// The problem with this is that, if someone upgrades their installed
+/// application, then realises it was a terrible mistake and tries to roll
+/// back, then suddenly every user's session is completely hosed. (They see
+/// an "unable to restore session" dialog.) Often, schema updates aren't
+/// actually backwards-incompatible — for example, existing code will work just
+/// fine if someone adds a new store or a new index — so this approach is too
+/// heavy-handed.
+///
+/// The solution we take here is to say "any schema changes up to
+/// [`MAX_SUPPORTED_SCHEMA_VERSION`] will be backwards-compatible". If, at some
+/// point, we do make a breaking change, we will give that schema version a
+/// higher number. Then, rather than using the implicit version check that comes
+/// with `indexedDB.open(name, version)`, we explicitly check the version
+/// ourselves.
+///
+/// It is expected that we will use version numbers that are multiples of 100 to
+/// represent breaking changes — for example, version 100 is a breaking change,
+/// as is version 200, but versions 101-199 are all backwards compatible with
+/// version 100. In other words, if you divide by 100, you get something
+/// approaching semver: version 200 is major version 2, minor version 0.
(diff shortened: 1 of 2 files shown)


crates/matrix-sdk-indexeddb/src/crypto_store/migrations/mod.rs
```diff
@@ -62,6 +62,40 @@ impl Drop for MigrationDb {
...
+/// heavy-handed.
+///
+/// The solution we take here is to say "any schema changes up to
+/// [`MAX_SUPPORTED_SCHEMA_VERSION`] will be backwards-compatible". If, at some
+/// point, we do make a breaking change, we will give that schema version a
+/// higher number. Then, rather than using the implicit version check that comes
+/// with `indexedDB.open(name, version)`, we explicitly check the version
+/// ourselves.
+///
+/// It is expected that we will use version numbers that are multiples of 10 to
```
Hywan, line 92: I appreciate the proposal. Thanks for working on this!

I wonder if 10 versions before a breaking change is enough. May 100 be a safer interval? It doesn't hurt, but it gives more rooms for backward-compatible new versions.

What do you think?

crates/matrix-sdk-indexeddb/src/crypto_store/migrations/mod.rs
```diff
@@ -62,6 +62,40 @@ impl Drop for MigrationDb {
...
+/// represent breaking changes — for example, version 20 is a breaking change,
+/// as is version 30, but versions 21-29 are all backwards compatible with
+/// version 20. In other words, if you divide by 10, you get something
+/// approaching semver: version 20 is major version 2, minor version 0.
+const MAX_SUPPORTED_SCHEMA_VERSION: u32 = 19;
```
Hywan, line 97: To avoid being confused by the interval (10 —as you suggest— or 100 —as I suggest—) minus 1 and so on, we may want to have to `const` expression here.

```rust
use std::num::NonZeroU32;

// The value we must update.
// Note: this is `NonZeroU32` for better safety, otherwise we might have 0 in `MAX_…`.
const NUMBER_OF_BREAKING_CHANGES: NonZeroU32 = unsafe { NonZeroU32::new_unchecked(2) };

// The value we use in the code.
// Note: the `saturating_sub` should not be necessary because of `NonZeroU32` but we never know!
const MAX_SUPPORTED_SCHEMA_VERSION: u32 = (NUMBER_OF_BREAKING_CHANGES.get() * 100).saturating_sub(1);
```

crates/matrix-sdk-indexeddb/src/crypto_store/migrations/mod.rs
richvdh: My main thinking here was that, if we find ourselves getting close to 20 without a breaking change, we can always bump `MAX_SUPPORTED_SCHEMA_VERSION` to 29, provided we do so before we actually *hit* 20.

You might be right though. Version numbers are cheap (we've got 4 billion of them to play with...)

richvdh: I've made this change in [41dfa88](https://github.com/matrix-org/matrix-rust-sdk/pull/3812/commits/41dfa888ac742bb264ab4186036bc676c58bd1b1)

crates/matrix-sdk-indexeddb/src/crypto_store/migrations/mod.rs
Hywan: We can still improve this in the future if we observe difficulties to understand or to update `MAX_SUPPORTED_SCHEMA_VERSION`.

Hywan (approved):
Please rebase your commits before merging. Otherwise it looks good, thanks!

richvdh:
I'm not sure this is actually going to help very much in practice, because changes to the format of the serialized objects (such as that in https://github.com/matrix-org/matrix-rust-sdk/pull/3795) are also backwards-incompatible :(.

PR #3814, opened 2024-08-07
FFI Element Call: Fix the permission list returned by the `get_element_call_required_permissions` ffi function.

toger5:
Before we set them for reading events which is not required since:
```rust
read: vec![
  WidgetEventFilter::StateWithType { event_type: StateEventType::CallMember.to_string() },
]
```
already covers all keys. And an additional `@` was used resulting in:
`@@user_id:demain.org_DEVICEID` and `_@@user_id:demain.org_DEVICEID`.

This also made it obvious that we need unit tests for this ffi method!

(We currently use the tauri widget wrapper to do manual end to end tests for the widget driver. But the ffi layer is not part of the tauri app. The testing cycle for EX is cumbersome without merging the changes.
Offtopic: any ideas or recommendations how to mitigate this in the future are welcome)

Signed-off-by:

### DIFF
--- a/bindings/matrix-sdk-ffi/src/widget.rs
+++ b/bindings/matrix-sdk-ffi/src/widget.rs
@@ -304,14 +304,14 @@ pub fn get_element_call_required_permissions(
             // [MSC3779](https://github.com/matrix-org/matrix-spec-proposals/pull/3779), with no leading underscore.
             WidgetEventFilter::StateWithTypeAndStateKey {
                 event_type: StateEventType::CallMember.to_string(),
-                state_key: format!("{}_{}", own_user_id, own_device_id),
+                state_key: format!("{own_user_id}_{own_device_id}"),
             },
             // The same as above but with an underscore.
             // To work around the issue that state events starting with `@` have to be matrix id's
             // but we use mxId+deviceId.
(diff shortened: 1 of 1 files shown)


bindings/matrix-sdk-ffi/src/widget.rs
```diff
@@ -303,10 +292,26 @@ pub fn get_element_call_required_permissions(
...
+            WidgetEventFilter::StateWithTypeAndStateKey {
+                event_type: StateEventType::CallMember.to_string(),
+                state_key: own_user_id.clone(),
+            },
+            // `delayed_event`` version for session memberhips
+            // [MSC3779](https://github.com/matrix-org/matrix-spec-proposals/pull/3779), with no leading underscore.
             WidgetEventFilter::StateWithTypeAndStateKey {
                 event_type: StateEventType::CallMember.to_string(),
-                state_key: own_user_id,
+                state_key: format!("{}_{}", own_user_id, own_device_id),
```
Hywan, line 307: Could be:

```suggestion
                state_key: format!("{own_user_id}_{own_device_id}"),
```

Probably more Rust idiomatic.

Hywan (approved):
@toger5 is on holidays. I'm taking over this PR.

PR #3816, opened 2024-08-08
Crypto: return error when sharing keys with a user that was previously verified but is not any more

BillCarsonFr:
Fixes https://github.com/matrix-org/matrix-rust-sdk/issues/3793

~~(currently based on https://github.com/matrix-org/matrix-rust-sdk/pull/3847)~~


andybalaam (approved):
Looks good but I think it needs rebasing on top of a [PR](https://github.com/matrix-org/matrix-rust-sdk/pull/3847) I reviewed last week.

PR #3820, opened 2024-08-08
feat(sdk): Subscribe to many rooms only via Sliding Sync

Hywan:
This patch changes the `SlidingSync::subscribe_to_room` method to `subscribe_to_rooms`. Note the plural form. It's now mandatory to subscribe to a set of rooms. The idea is to avoid calling this method repeatedly. Why? Because each time the method is called, it sends a `SlidingSyncInternalMessage` of kind `SyncLoopSkipOverCurrentIteration`, i.e. it cancels the in-flight sliding sync request, to start over with a new one (with the new room subscription). A problem arises when the async runtime (here, Tokio) is busy: in this case, the internal message channel can be filled pretty easily because its size is 8. Messages are not consumed as fast as they are inserted. By changing this API: subscribing to multiple rooms will result in a single internal message, instead of one per room.

Consequently, the rest of the patch moves the `subscribe` method of `room_list_service::Room` to `room_list_service::RoomListService` because it now concerns multiple rooms instead of a single one.

### DIFF
--- a/bindings/matrix-sdk-ffi/src/room_list.rs
+++ b/bindings/matrix-sdk-ffi/src/room_list.rs
@@ -132,6 +132,26 @@ impl RoomListService {
             }
         })))
     }
+
+    fn subscribe_to_rooms(
+        &self,
+        room_ids: Vec<String>,
+        settings: Option<RoomSubscription>,
+    ) -> Result<(), RoomListError> {
+        let room_ids = room_ids
+            .into_iter()
+            .map(|room_id| {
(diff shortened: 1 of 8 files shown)


jmartinesp (approved):
It's hard to confirm since Android didn't have an issue with the previous approach, but everything seems to work fine AFAICT, and we no longer have long lists of 'subscribing to...' logs, which is a plus since it reduces the noise quite a bit.

stefanceriu (approved):
looks and works great on iOS too, no more panics either 👍

PR #3821, opened 2024-08-09
ffi: provide manual cancellation mechanism for identity/cross-signing reset handles

stefanceriu:
Works around swift issue where nil-ing the handle is not enough for it to get cancelled.


jmartinesp (approved):
Thanks, I can confirm this still works fine on Android with just cancelling the coroutine that runs the code, but also calling the `cancel` method works.

PR #3822, opened 2024-08-09
Sync: Don't spuriously show spinner

erikjohnston:
We should only show the spinner if the *first* sliding sync request is taking a while. If we have received some data and the second request takes a while, that is OK.

For the state transition of `Init -> SettingUp` this is handled correctly, however for `Terminated -> Recovering -> Running` we waited until the second request returned before hiding the sync spinner. This meant that if the first request returned quickly the app would show new data and *then* the sync spinner would show (if the second request took time).

This situation occurs frequently with the new SSS API, where if all the new data was returned in the first sync then the second sync would block waiting for new data, triggering the sync spinner.

--- 

Not sure if this needs a changelog entry, as it really only is noticeable with SSS.

### DIFF
--- a/crates/matrix-sdk-ui/src/room_list_service/mod.rs
+++ b/crates/matrix-sdk-ui/src/room_list_service/mod.rs
@@ -301,11 +301,11 @@ impl RoomListService {
 
             loop {
                 let (sync_indicator, yield_delay) = match current_state {
-                    State::Init | State::Recovering | State::Error { .. } => {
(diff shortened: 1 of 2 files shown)


stefanceriu (commented):
This all sounds and looks good to me but I would like Ivan's opinion before we merge it

Hywan (approved):
It looks good to me, it makes sense. Thanks for having updated the tests too!

PR #3823, opened 2024-08-09
testing: create `ruma_response_from_json`

richvdh:
We had *two* copies of `response_from_file`, and all calls to them were always immediately followed by an operation to parse the response as a Ruma response object.

We can save a whole lot of boilerplate with a generic function that wraps the json into an HTTP response *and* parses it into a Ruma object.

### DIFF
--- a/crates/matrix-sdk-crypto/Cargo.toml
+++ b/crates/matrix-sdk-crypto/Cargo.toml
@@ -27,7 +27,7 @@ _disable-minimum-rotation-period-ms = []
 message-ids = []
 
 # Testing helpers for implementations based upon this
-testing = ["dep:http"]
+testing = ["matrix-sdk-test"]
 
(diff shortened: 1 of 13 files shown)


crates/matrix-sdk-crypto/Cargo.toml
```diff
@@ -26,8 +26,7 @@ _disable-minimum-rotation-period-ms = []
 # "message-ids" feature doesn't do anything and is deprecated.
 message-ids = []
 
-# Testing helpers for implementations based upon this
-testing = ["dep:http"]
+testing = []
```
richvdh, line 29: since `matrix_sdk_crypto::machine::testing::response_from_file` no longer exists, we no longer need `http`.

bnjbvr (approved):
Nice!!

PR #3824, opened 2024-08-09
Remove redundant `testing` feature from matrix-sdk-crypto

richvdh:
Nobody actually needs any of this, so we may as well kill it off.


richvdh:
This turns out to be misguided...

Issue #3720, opened 2024-07-17
in-memory state store: use a single lock for all the data instead of one lock per field


bnjbvr:
The memory store is not intended to be used for high-performance use cases, so it should be fine to use a single lock instead of one lock per field [there](https://github.com/matrix-org/matrix-rust-sdk/blob/1cc82920348e5a240db283ea356e25bc5bea2e3a/crates/matrix-sdk-base/src/store/memory_store.rs#L54).

We could have something like this:

```rust
struct MemoryStore(Mutex<MemoryStoreInner>);

struct MemoryStoreInner {
  recently_visited_rooms: HashMap<OwnedUserId, Vec<OwnedRoomId>>,
  // etc.
}
```

and adapt all the uses to take the lock at most once per method.

Once that's done and it compiles, you'd need to make sure all the tests keep on passing; perhaps some tests will show that we sometimes attempt to take the lock recursively, in which case we'd need to split into two (or more) locked sets of fields.

PR #3825, opened 2024-08-09
#3720 issue : Updated memory_store.rs

AllMightLegend:
used a single lock for all the data instead of one lock per field

Signed-off-by:


bnjbvr (changes requested):
Thanks for trying to contribute to the Rust SDK! The checked-in `.history` files that are unnecessary, and the code changes are unlikely to be sufficient. Have you tried to compile the SDK and run tests?

bnjbvr:
Do you have some basic Rust experience? If not, I'd suggest putting this aside and getting better with Rust in the first place, using resources available [here](https://www.rust-lang.org/learn).

If you do have some basic Rust experience: I'd suggest compiling the code, and fix the compilation issues first.

AllMightLegend:
okay sir will do it

PR #3826, opened 2024-08-12
chore(ui): Add logs inside `RoomListService::sync`

Hywan:
This patch adds logs inside the `RoomListService::sync` method to know what are the current states.


stefanceriu (approved):
Looks good to me 👍

PR #3828, opened 2024-08-12
refactor(sdk): merge `Room::event` and `Room::event_with_config`

bnjbvr:
It's better to have fewer public APIs, especially when there's little annoyance to have it. We could use a request builder that converts into a Future, too, but considering there's only a single optional parameter, it's fine to include it in the function's signature.


Hywan (approved):
Looks sane to me, thanks!

PR #3831, opened 2024-08-12
pinned events: a collection of small refactorings

bnjbvr:
To make the code more idiomatic, avoid having too many concepts, rename a few things, etc.

Next I'll merge the pinned event cache into the event cache, because there's no reason to have two different event caches.

### DIFF
--- a/crates/matrix-sdk-ui/src/timeline/builder.rs
+++ b/crates/matrix-sdk-ui/src/timeline/builder.rs
@@ -22,7 +22,7 @@ use matrix_sdk::{
 };
 use ruma::{events::AnySyncTimelineEvent, RoomVersionId};
 use tokio::sync::broadcast::error::RecvError;
-use tracing::{error, info, info_span, trace, warn, Instrument, Span};
+use tracing::{info, info_span, trace, warn, Instrument, Span};
 
 #[cfg(feature = "e2e-encryption")]
 use super::to_device::{handle_forwarded_room_key_event, handle_room_key_event};
@@ -207,8 +207,6 @@ impl TimelineBuilder {
                 info_span!(parent: Span::none(), "room_update_handler", room_id = ?room.room_id());
             span.follows_from(Span::current());
 
-            let focus = Arc::new(focus);
-
             async move {
                 trace!("Spawned the event subscriber task.");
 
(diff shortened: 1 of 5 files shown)


crates/matrix-sdk-ui/src/timeline/pinned_events_loader.rs
```diff
@@ -64,31 +65,28 @@ impl PinnedEventsLoader {
...
 
-            for id in event_ids_to_request {
-                handles.push(tokio::spawn({
-                    let provider = Arc::clone(&provider);
-                    async move {
-                        provider
-                            .event_with_config(&id, config)
-                            .await
-                            .map_err(|_| PinnedEventsLoaderError::EventNotFound(id.to_owned()))
+            let new_events = join_all(event_ids_to_request.into_iter().map(|event_id| {
```
jmartinesp, line 87: I think `join_all` will fail if any of the futures fails, right? In this case we want to fail only if no event could be retrieved, either from the HS or the cache.

crates/matrix-sdk-ui/src/timeline/builder.rs
```diff
@@ -265,11 +265,16 @@ impl TimelineBuilder {
...
 
-                            // Special case for pinned events: when we receive new events what we'll do is, instead of adding the
-                            // events, update the pinned events cache with them, reload the list of pinned event ids and reload
-                            // the list of pinned events with this info.
+                            // Special case for pinned events: when we receive new events what
+                            // we'll do is, instead of adding the events, update the pinned events
+                            // cache with them, reload the list of pinned event ids and reload the
+                            // list of pinned events with this info.
+
+                            // TODO(bnjbvr): why is this done? the timeline already listens to
```
jmartinesp: I think I did it to simplify the flow and ensure the chronological sorting was enforced. It may not be needed.

crates/matrix-sdk-ui/src/timeline/builder.rs
```diff
@@ -265,22 +263,9 @@ impl TimelineBuilder {
...
-                            if let TimelineFocus::PinnedEvents { .. } = &*focus.clone() {
-                                if let Some(ret) = inner.update_pinned_events_if_needed(events, &pinned_event_cache).await {
-                                    match ret {
-                                        Ok(events) => inner.replace_with_initial_remote_events(events, RemoteEventOrigin::Sync).await,
-                                        Err(err) => error!("Couldn't update pinned events with incoming timeline events: {err}"),
-                                    }
-                                }
-                            } else {
+                            // Note: we deliberately choose to not handle
+                            // updates/reactions/redactions for pinned events.
```
jmartinesp, line 267: This is temporary, right? We do want to know about updates/redactions. Or does it just mean we don't handle them here?

jmartinesp (commented):
The changes LGTM in general, but I have a few questions.

crates/matrix-sdk-ui/src/timeline/builder.rs
bnjbvr: This was the previous behavior: the new events were not added, so I assumed this was expected. (At least it wasn't tested since no test failed from this change.)

crates/matrix-sdk-ui/src/timeline/pinned_events_loader.rs
bnjbvr: `join_all` takes a future which output is not a result, so it can't fail. If you look at the `async` block, it doesn log errors. Using `try_join_all` would have caused the issue you describe here.

crates/matrix-sdk-ui/src/timeline/builder.rs
jmartinesp: Unless I missed something [the very first test](https://github.com/matrix-org/matrix-rust-sdk/blob/1c4f035c99284e6c245843ec0066a4e786726a37/crates/matrix-sdk-ui/tests/integration/timeline/pinned_event.rs#L22) should handle that, I'm not sure why there was no behaviour change 🫤 .

jmartinesp: Ok, so this was a misunderstanding on my side of how the Matrix spec works. I thought for updates and redaction I'd get an event with the same event id but different contents. Like this, it would be filtered, I'd update the cache with the new contents, then reload the list taking this value from the cache.

But if it doesn't work like that, then I wasn't really doing anything.

PR #3832, opened 2024-08-12
feat(sdk): `compute_limited` is only useful for the SS proxy

Hywan:
This patch restricts the call to `compute_limited` to the sliding sync proxy implementation (aka MCS3575). It is not necessary for the sliding sync native implementation (aka Simplified MSC3575). The proxy doesn't implement the `limited` flag, contrary to Synapse. Let's not run workarounds when we don't need them.


stefanceriu (approved):
👍

bnjbvr (approved):
:+1:

PR #2395, opened 2023-08-09
feat(sdk): Skip Sliding Sync `Response` if it's been received already

Hywan:
A Sliding Sync `v4::Response` contains a `pos` value. It helps to identify progress during several requests/responses. If two requests with the same `pos` are sent, the server **must** reply with the same response, as defined in the specification.

The corollary is: If a response contains a `pos` that has already been received, the client **must** ignore it, otherwise it can create duplications. For example, let a response containing sync operations, like `DELETE` or `INSERT`, with indexes: if this pair of operations are repeated twice with the same index, it creates a broken state.

To avoid this behaviour, this patch stores the last 20 position markers received from the server. If a response contains a `pos` that has already been received, a (new) error is returned. As with all the other errors, the sync-loop is stopped, and must restarted. The past positions survive to this restart, as for the rest of the state.

When the server replies with a `M_UNKNOWN_POS`, the `pos` and the `past_positions` are all cleared.

⚠️ Note to reviewer:

* **What I am sure about**. Checking that a `pos` has not been received before running `handle_response` must be done at the beginning of the method.
* **What I am not sure about**. Right now, the `pos` and `past_positions` are saved when `handle_response` starts to execute. The rest of the code in this method can fail. If an error is raised _after_ the `pos` has been saved, the next request will pretend everything was alright, and the server will reply with a different response. It can be a problem. For example, if a list has not been updated because an error happened before (e.g. with encryption handling, or with room response handling), the next list update will be broken because the previous one didn't happen.
I was thinking of saving `pos` at the end of the method, so that we are sure that everything was done correctly, _however_ it creates new problem: if the error was raised by the update of the list, the next request will use the same `pos`, the server will reply with the same response, and list updates is likely to be broken if it has partly been updated.
👉 Actually, the place where `pos` and `past_positions` must be placed is unclear to me yet:
    * Should we consider that errors emitted during `handle_response` (apart from the `pos` check) should lead to closing the sliding sync session?
    * Should we make `list.update` infallible? Making `handle_encryption`, `handle_room_response`, and `process_and_take_response` the only fallible places (with `pos` and `past_positions` to be saved _after_ those calls)?

PR #3833, opened 2024-08-12
fix(sdk): Remove `SlidingSyncInner::past_positions`

Hywan:
The patch https://github.com/matrix-org/matrix-rust-sdk/pull/2395 has introduced `SlidingSyncInner::past_positions` as a mechanism to filter duplicated responses. It was a problem because the sliding sync `ops` could easily create corrupted states if they were applied more than once.

Since https://github.com/matrix-org/matrix-rust-sdk/pull/3664/, `ops` are ignored.

Now, `past_positions` create a problem with the sliding sync native implementation inside Synapse because `pos` can stay the same between multiple responses.

While `past_positions` was helpful to fix bugs in the past, it's no longer necessary today. Moreover, it breaks an invariant about `pos`: we must consider it as a blackbox. It means we must ignore if a `pos` value has been received in the past or not. This invariant has been broken for good reasons, but it now creates new issues.

This patch removes `past_positions`, along with the associated code (like `Error::ResponseAlreadyReceived` for example).

### DIFF
--- a/crates/matrix-sdk/src/sliding_sync/builder.rs
+++ b/crates/matrix-sdk/src/sliding_sync/builder.rs
@@ -1,13 +1,12 @@
 use std::{
     collections::BTreeMap,
     fmt::Debug,
-    num::NonZeroUsize,
     sync::{Arc, RwLock as StdRwLock},
     time::Duration,
 };
 
 use matrix_sdk_base::sliding_sync::http;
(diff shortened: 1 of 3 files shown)


stefanceriu (approved):
Reasoning sounds good to me, code looks fine too 👍

PR #3834, opened 2024-08-12
pinned events: remove the `PinnedEventCache` and use the `EventCache` instead

bnjbvr:
The first commit introduces new capabilities in the `EventCache` so we can indeed get rid of the `PinnedEventsCache`. Mostly, the one interesting feature is to introduce a new side cache for events (independent of events stored in the linked chunks; note that in a future with a storage, the events will be stored only once, and those queries will be different SQL queries on the same entries). Then, we can have `Room::event()` save the events in the event cache, if it can (i.e. the event has an id). Finally, the `EventCache::event(event_id)` and `RoomEventCache::event(event_id)` methods can use this to retrieve an event by ID.

This is sufficient to remove the different event cache that was used for pinned events, aka `PinnedEventsCache`.


jmartinesp (approved):
Sorry for the delay in the review, I wanted to test it in a real client and I had some issues doing that for unrelated reasons. LGTM!

PR #3835, opened 2024-08-12
Test: improve documentation of `PreviouslyVerifiedTestData`

richvdh:
`PreviouslyVerifiedTestData` is currently an unfortunate code dump. Attempt to make it a bit less write-only with some explanation of wtf the methods do.

### DIFF
--- a/testing/matrix-sdk-test/src/test_json/keys_query_sets.rs
+++ b/testing/matrix-sdk-test/src/test_json/keys_query_sets.rs
@@ -650,29 +650,74 @@ impl IdentityChangeDataSet {
     }
 }
 
+/// A set of `/keys/query` responses that were initially created to simulate
+/// when a user that was verified reset his keys and became unverified.
+///
+/// The local user (as returned by [`PreviouslyVerifiedTestData::own_id`]) is
+/// `@alice:localhost`. There are 2 other users: `@bob:localhost` (returned by
+/// [`PreviouslyVerifiedTestData::bob_id`]), and `@carol:localhost` (returned by
+/// [`PreviouslyVerifiedTestData::carol_id`]).
+///
+/// We provide two `/keys/query` responses for each of Bob and Carol: one signed
+/// by Alice, and one not signed.
+///
+/// Bob and Carol each have 2 devices, one signed by the owning user, and
+/// another one not cross-signed.
+///
+/// The `/keys/query` responses were generated using a local synapse.
 pub struct PreviouslyVerifiedTestData {}
 
 #[allow(dead_code)]
 impl PreviouslyVerifiedTestData {
+    /// Secret part of Alice's master cross-signing key.
+    ///
+    /// Exported from Element-Web with the following console snippet:
+    ///
+    /// ```javascript
+    /// (await mxMatrixClientPeg.get().getCrypto().olmMachine.exportCrossSigningKeys()).masterKey
+    /// ```
     pub const MASTER_KEY_PRIVATE_EXPORT: &'static str =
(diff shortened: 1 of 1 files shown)


testing/matrix-sdk-test/src/test_json/keys_query_sets.rs
```diff
@@ -915,6 +915,24 @@ impl PreviouslyVerifiedTestData {
...
+
+    /// Device ID of Carol's unsigned device.
+    ///
+    /// The device is returned as part of
+    /// [`Self::carol_keys_query_response_signed`] and
+    /// [`Self::carol_keys_query_response_unsigned`].
+    pub fn carol_unsigned_device_id() -> &'static DeviceId {
+        device_id!("BAZAPVEHGA")
+    }
+
```
richvdh, line 992: arguably this doesn't match the subject of the PR, but the main reason I'm adding them now is so that I can refer back to them in documentation that I'm going to add in the next commit...

testing/matrix-sdk-test/src/test_json/keys_query_sets.rs
```diff
@@ -650,29 +650,74 @@ impl IdentityChangeDataSet {
...
 #[allow(dead_code)]
 impl PreviouslyVerifiedTestData {
+    /// Secret part of Alice's master cross-signing key.
+    ///
+    /// Exported from Element-Web with the following console snippet:
+    ///
+    /// ```javascript
+    /// (await mxMatrixClientPeg.get().getCrypto().olmMachine.exportCrossSigningKeys()).masterKey
+    /// ```
     pub const MASTER_KEY_PRIVATE_EXPORT: &'static str =
```
stefanceriu, line 679: Would it be worth it to prepend `ALICE` to these key constants? Or maybe reuse the same `own` naming convention

testing/matrix-sdk-test/src/test_json/keys_query_sets.rs
```diff
@@ -919,9 +981,30 @@ impl PreviouslyVerifiedTestData {
         ruma_response_from_json(&data)
     }
 
-    pub fn device_1_keys_payload_carol() -> Value {
+    /// Device ID of Carol's signed device.
+    ///
+    /// The device is returned as part of
+    /// [`Self::carol_keys_query_response_signed`] and
+    /// [`Self::carol_keys_query_response_unsigned`].
+    pub fn carol_signed_device_id() -> &'static DeviceId {
```
stefanceriu, line 989: Seeing as the users have 2 devices each, would it be better to always call them `_signed`/`_unsigned` and drop the `_1` numbering scheme used above (`Self::bob_device_1_id`)?

stefanceriu (approved):
Nice cleanup, thanks!

testing/matrix-sdk-test/src/test_json/keys_query_sets.rs
richvdh: probably. I was trying to avoid getting into doing much beyond documenting the current behaviour here.

testing/matrix-sdk-test/src/test_json/keys_query_sets.rs
richvdh: The problem is that which of bob's devices are signed depends on whether you use `bob_keys_query_response_signed` or `bob_keys_query_response_rotated`.

richvdh:
@stefanceriu thanks for the review. I'm going to assume that neither of your comments are blocking concerns and go ahead and merge.

PR #3836, opened 2024-08-13
rtc: Handle non-MXID call member event state keys

AndrewFerr:
Update Ruma dependency to expect call membership state events with state keys that are arbitrary strings, not just pure MXIDs.

When a call membership state key does not exactly match the format of an MXID, treat it as a valid state key if it starts with an MXID followed by an underscore, with that MXID designating the owner of the event.

(The state key may also be optionally prefixed with an underscore, which is permitted as a way to bypass pre-MSC3757 authorization rules against sending state events with state keys that do not exactly match the sender's MXID.)

Signed-off-by: Andrew Ferrazzutti <andrewf@element.io>

### DIFF
--- a/crates/matrix-sdk-base/src/rooms/mod.rs
+++ b/crates/matrix-sdk-base/src/rooms/mod.rs
@@ -327,30 +327,26 @@ impl BaseRoomInfo {
 /// - `<user ID>_<string>`
 /// - `_<user ID>_<string>`
 fn get_user_id_for_state_key(state_key: &str) -> Option<OwnedUserId> {
-    match UserId::parse(state_key) {
-        Ok(user_id) => Some(user_id),
-        Err(_) => {
-            // Ignore leading underscore if present
-            // (used for avoiding auth rules on @-prefixed state keys)
-            let state_key = state_key.strip_prefix('_').unwrap_or(state_key);
-            if state_key.starts_with('@') {
-                if let Some(colon_idx) = state_key.find(':') {
-                    let state_key_user_id = match state_key[colon_idx + 1..].find('_') {
-                        None => state_key,
-                        Some(suffix_idx) => &state_key[..colon_idx + 1 + suffix_idx],
-                    };
-                    match UserId::parse(state_key_user_id) {
-                        Ok(user_id) => Some(user_id),
-                        Err(_) => None,
-                    }
-                } else {
-                    None
-                }
-            } else {
-                None
+    if let Ok(user_id) = UserId::parse(state_key) {
+        return Some(user_id);
+    }
+
+    // Ignore leading underscore if present
+    // (used for avoiding auth rules on @-prefixed state keys)
+    let state_key = state_key.strip_prefix('_').unwrap_or(state_key);
+    if state_key.starts_with('@') {
+        if let Some(colon_idx) = state_key.find(':') {
+            let state_key_user_id = match state_key[colon_idx + 1..].find('_') {
+                None => state_key,
+                Some(suffix_idx) => &state_key[..colon_idx + 1 + suffix_idx],
+            };
+            if let Ok(user_id) = UserId::parse(state_key_user_id) {
+                return Some(user_id);
             }
         }
     }
+
+    None
 }
 
 bitflags! {
--- a/crates/matrix-sdk-base/src/rooms/mod.rs
+++ b/crates/matrix-sdk-base/src/rooms/mod.rs
@@ -197,7 +197,7 @@ impl BaseRoomInfo {
 
                 let state_key = m.state_key();
                 let owned_user_id = match UserId::parse(state_key) {
-                    Ok(u) => u,
+                    Ok(user_id) => user_id,
                     Err(_) => {
                         // Ignore leading underscore if present
                         // (used for avoiding auth rules on @-prefixed state keys)
@@ -209,7 +209,7 @@ impl BaseRoomInfo {
                                     Some(suffix_idx) => &state_key[..colon_idx + 1 + suffix_idx],
                                 };
                                 match UserId::parse(state_key_user_id) {
-                                    Ok(u) => u,
+                                    Ok(user_id) => user_id,
                                     Err(_) => return false,
                                 }
                             } else {
--- a/crates/matrix-sdk-base/src/rooms/mod.rs
+++ b/crates/matrix-sdk-base/src/rooms/mod.rs
@@ -195,31 +195,10 @@ impl BaseRoomInfo {
                 let mut o_ev = o_ev.clone();
                 o_ev.content.set_created_ts_if_none(o_ev.origin_server_ts);
 
-                let state_key = m.state_key();
-                let owned_user_id = match UserId::parse(state_key) {
-                    Ok(user_id) => user_id,
-                    Err(_) => {
-                        // Ignore leading underscore if present
-                        // (used for avoiding auth rules on @-prefixed state keys)
-                        let state_key = state_key.strip_prefix('_').unwrap_or(state_key);
-                        if state_key.starts_with('@') {
-                            if let Some(colon_idx) = state_key.find(':') {
-                                let state_key_user_id = match state_key[colon_idx + 1..].find('_') {
-                                    None => state_key,
-                                    Some(suffix_idx) => &state_key[..colon_idx + 1 + suffix_idx],
-                                };
-                                match UserId::parse(state_key_user_id) {
-                                    Ok(user_id) => user_id,
-                                    Err(_) => return false,
-                                }
-                            } else {
-                                return false;
-                            }
-                        } else {
-                            return false;
-                        }
-                    }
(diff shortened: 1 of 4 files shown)


crates/matrix-sdk-base/src/rooms/mod.rs
```diff
@@ -195,9 +195,36 @@ impl BaseRoomInfo {
                 let mut o_ev = o_ev.clone();
                 o_ev.content.set_created_ts_if_none(o_ev.origin_server_ts);
 
+                let state_key = m.state_key();
+                let owned_user_id = match UserId::parse(state_key) {
```
bnjbvr, line 199: 1. Can you extract this into a helper function please?
2. Can you add unit tests for this function please?

crates/matrix-sdk-base/src/rooms/mod.rs
```diff
@@ -195,9 +195,36 @@ impl BaseRoomInfo {
...
+                    Ok(u) => u,
+                    Err(_) => {
+                        // Ignore leading underscore if present
+                        // (used for avoiding auth rules on @-prefixed state keys)
+                        let state_key = match state_key.strip_prefix('_') {
+                            None => state_key,
+                            Some(stripped_state_key) => stripped_state_key,
+                        };
```
bnjbvr, line 207: nit: `let state_key = state_key.strip_prefix('_').unwrap_or(state_key);`

bnjbvr (changes requested):
Thanks, looks sane but it'd be nice to have some basic unit tests (which would also show what behavior is expected and what isn't).

crates/matrix-sdk-base/src/rooms/mod.rs
```diff
@@ -195,9 +195,36 @@ impl BaseRoomInfo {
                 let mut o_ev = o_ev.clone();
                 o_ev.content.set_created_ts_if_none(o_ev.origin_server_ts);
```
stefanceriu, line 196: Can we please not use abbreviations, presumably `o_ev` is `original_event`. `Ok(u)` can easily be `Ok(user_id)`. I see no benefit in shortening them this way, especially as it's not consistent with the rest of the scope.

stefanceriu (commented):
> I hadn't seen you removed my review request

Oh no worries, I only removed you because I can test it end to end, but it doesn't really matter.

It seems to work just fine on both `call.element.io` and `call.element.dev` so from my point of view you can merge it after adressing benji's concerns. Nicely done! 👍

crates/matrix-sdk-base/src/rooms/mod.rs
AndrewFerr: `o_ev` was there before this PR, so I'll leave the task of changing it to a later PR to keep this one focused.  But I will rename the `u` variables that were added in this PR.

stefanceriu: Fair 👍

crates/matrix-sdk-base/src/rooms/mod.rs
AndrewFerr: [`20b5dba`](https://github.com/matrix-org/matrix-rust-sdk/pull/3836/commits/20b5dba3e0b19d312c8df4e6987c46cdf742be5a) & [`cfbe204`](https://github.com/matrix-org/matrix-rust-sdk/pull/3836/commits/cfbe204313f4760f8375efaf8bdb1bb2b3cda4d5)

crates/matrix-sdk-base/src/rooms/mod.rs
AndrewFerr: [`96bba75`](https://github.com/matrix-org/matrix-rust-sdk/pull/3836/commits/96bba7523e431d54a4f3ce823ef2c0cc350ffb9a)

crates/matrix-sdk-base/src/rooms/mod.rs
AndrewFerr: [`8932070`](https://github.com/matrix-org/matrix-rust-sdk/pull/3836/commits/89320701a401558d327693da16e9ae09c5093af8)

crates/matrix-sdk-base/src/rooms/mod.rs
```diff
@@ -319,6 +322,37 @@ impl BaseRoomInfo {
...
+                        Err(_) => None,
+                    }
+                } else {
+                    None
+                }
+            } else {
+                None
+            }
+        }
+    }
```
bnjbvr, line 353: Let's rewrite this code to avoid all the indentation levels and the many empty branches with `None`:

```suggestion
    if let Ok(user_id) = UserId::parse(state_key) {
        return Some(user_id);
    }

    // Ignore leading underscore if present
    // (used for avoiding auth rules on @-prefixed state keys)
    let state_key = state_key.strip_prefix('_').unwrap_or(state_key);
    if state_key.starts_with('@') {
        if let Some(colon_idx) = state_key.find(':') {
            let state_key_user_id = match state_key[colon_idx + 1..].find('_') {
                None => state_key,
                Some(suffix_idx) => &state_key[..colon_idx + 1 + suffix_idx],
            };
            if let Ok(user_id) = UserId::parse(state_key_user_id) {
                return Some(user_id);
            }
        }
    }

    None
```

crates/matrix-sdk-base/src/rooms/mod.rs
```diff
@@ -569,4 +604,19 @@ mod tests {
...
         assert!(base_room_info.notable_tags.contains(RoomNotableTags::LOW_PRIORITY).not());
     }
+
+    #[test]
+    fn test_get_user_id_from_state_key() {
+        assert_matches!(get_user_id_for_state_key(""), None);
+        assert_matches!(get_user_id_for_state_key("abc"), None);
+        assert_matches!(get_user_id_for_state_key("username:example.org"), None);
+
+        assert_matches!(get_user_id_for_state_key("@username:example.org"), Some(_));
```
bnjbvr, line 614: Can you explicit, in all of the `Some()` tests, what the expected value is? e.g.

```suggestion
        assert_matches!(get_user_id_for_state_key("@username:example.org"), Some(user_id!("@username:example.org"));
```

bnjbvr (changes requested):
Getting closer, thanks for making a separate function and writing tests!

crates/matrix-sdk-base/src/rooms/mod.rs
AndrewFerr: [`699369a`](https://github.com/matrix-org/matrix-rust-sdk/pull/3836/commits/699369aa2876c98144ac220227e68e8e12094834) -- let me know if this too verbose.

crates/matrix-sdk-base/src/rooms/mod.rs
```diff
@@ -569,4 +603,37 @@ mod tests {
...
+        assert_matches!(get_user_id_for_state_key("abc"), None);
+        assert_matches!(get_user_id_for_state_key("@nocolon"), None);
+        assert_matches!(get_user_id_for_state_key("@noserverpart:"), None);
+        assert_matches!(get_user_id_for_state_key("@noserverpart:_suffix"), None);
+
+        let user_id = user_id!("@username:example.org");
+
+        assert_matches!(
+            get_user_id_for_state_key(user_id.as_str()),
+            Some(captured_user_id) => assert_eq!(captured_user_id, user_id));
```
bnjbvr, line 619: Does this not work? (and a few times below)

```suggestion
        assert_eq!(
            get_user_id_for_state_key(user_id.as_str()),
            Some(user_id)
        );
```

crates/matrix-sdk-base/src/rooms/mod.rs
```diff
@@ -569,4 +603,37 @@ mod tests {
...
         assert!(base_room_info.notable_tags.contains(RoomNotableTags::LOW_PRIORITY).not());
     }
+
+    #[test]
+    fn test_get_user_id_for_state_key() {
+        assert_matches!(get_user_id_for_state_key(""), None);
+        assert_matches!(get_user_id_for_state_key("abc"), None);
+        assert_matches!(get_user_id_for_state_key("@nocolon"), None);
+        assert_matches!(get_user_id_for_state_key("@noserverpart:"), None);
+        assert_matches!(get_user_id_for_state_key("@noserverpart:_suffix"), None);
```
bnjbvr, line 613: All those can be `assert!(get_....().is_none())`, which is simpler to read.

bnjbvr (approved):
Looks good to me, modulo test nits, thanks! Please push another commit with the test changes, and then no need to ask for another review, I'll merge it :+1:

crates/matrix-sdk-base/src/rooms/mod.rs
bnjbvr: This required `.as_deref()` to the LHS, because `get_user_id_for_state_key` returns an `OwnedUserId` while the `user_id` was a `&UserId`.

bnjbvr:
I've addressed the nits for you actually, so we can merge :)

PR #3840, opened 2024-08-14
feat(sdk): Live updates for pinned events timeline with replacements and redactions

jmartinesp:
This is a follow-up PR for https://github.com/matrix-org/matrix-rust-sdk/pull/3870.

## Changes

- Make `TimelineEventBuilder` process synced events for `TimelineFocus::PinnedEvents` too.
- Use the relationship cache added in https://github.com/matrix-org/matrix-rust-sdk/pull/3870 to load pinned events and other events related to them so their final timeline event results contain the changes those would add (edits, redactions, reactions, etc.).
- Make `Room::event_with_context` also save its results to the cache.
- Add new tests for pinned events and related events.
- Fix the pinned events benchmark in `room_bench`.

The end result of this is the pinned events timeline now reacts to edits and redactions, and those values are kept even when the event list reloads.

Signed-off-by:

### DIFF
--- a/benchmarks/benches/room_bench.rs
+++ b/benchmarks/benches/room_bench.rs
@@ -1,4 +1,4 @@
-use std::time::Duration;
+use std::{sync::Arc, time::Duration};
 
 use criterion::{criterion_group, criterion_main, BenchmarkId, Criterion, Throughput};
 use matrix_sdk::{
@@ -157,7 +157,6 @@ pub fn load_pinned_events_benchmark(c: &mut Criterion) {
             .respond_with(move |r: &Request| {
                 let segments: Vec<&str> = r.url.path_segments().expect("Invalid path").collect();
                 let event_id_str = segments[6];
-                // let f = EventFactory::new().room(&room_id)
                 let event_id = EventId::parse(event_id_str).expect("Invalid event id in response");
                 let event = f
                     .text_msg(format!("Message {event_id_str}"))
@@ -170,9 +169,6 @@ pub fn load_pinned_events_benchmark(c: &mut Criterion) {
             })
             .mount(&server),
     );
-    // runtime.block_on(server.reset());
-
-    client.event_cache().subscribe().unwrap();
 
     let room = client.get_room(&room_id).expect("Room not found");
     assert!(!room.pinned_event_ids().is_empty());
@@ -183,6 +179,15 @@ pub fn load_pinned_events_benchmark(c: &mut Criterion) {
     let mut group = c.benchmark_group("Test");
     group.throughput(Throughput::Elements(count as u64));
     group.sample_size(10);
+
+    let client = Arc::new(client);
+
+    {
+        let client = client.clone();
+        runtime.spawn_blocking(move || {
+            client.event_cache().subscribe().unwrap();
+        });
+    }
 
     group.bench_function(BenchmarkId::new("load_pinned_events", name), |b| {
         b.to_async(&runtime).iter(|| async {
@@ -207,7 +212,6 @@ pub fn load_pinned_events_benchmark(c: &mut Criterion) {
     {
         let _guard = runtime.enter();
         runtime.block_on(server.reset());
-        drop(client);
         drop(server);
     }
 
--- a/crates/matrix-sdk-ui/src/timeline/pinned_events_loader.rs
+++ b/crates/matrix-sdk-ui/src/timeline/pinned_events_loader.rs
@@ -17,12 +17,12 @@ use std::{fmt::Formatter, num::NonZeroUsize, sync::Arc};
 use futures_util::{future::join_all, FutureExt as _};
 use matrix_sdk::{
     config::RequestConfig, event_cache::paginator::PaginatorError, BoxFuture, Room,
-    SendOutsideWasm, SyncOutsideWasm,
 };
 use matrix_sdk_base::deserialized_responses::SyncTimelineEvent;
 use ruma::{EventId, MilliSecondsSinceUnixEpoch, OwnedEventId};
 use thiserror::Error;
 use tracing::{debug, warn};
+
 use crate::timeline::event_handler::TimelineEventKind;
 
 const MAX_CONCURRENT_REQUESTS: usize = 10;
@@ -74,8 +74,12 @@ impl PinnedEventsLoader {
         let new_events = join_all(pinned_event_ids.into_iter().map(|event_id| {
             let provider = self.room.clone();
             async move {
-                match provider.load_event(&event_id, request_config).await {
-                    Ok(event) => Some(event),
+                match provider.load_event_with_relations(&event_id, request_config).await {
+                    Ok((event, related_events)) => {
+                        let mut events = vec![event];
+                        events.extend(related_events);
+                        Some(events)
+                    }
                     Err(err) => {
                         warn!("error when loading pinned event: {err}");
                         None
@@ -85,7 +89,13 @@ impl PinnedEventsLoader {
         }))
         .await;
 
-        let mut loaded_events = new_events.into_iter().flatten().collect::<Vec<_>>();
+        let mut loaded_events = new_events
+            .into_iter()
+            // Get only the `Some<Vec<_>>` results
+            .flatten()
+            // Flatten the `Vec`s into a single one containing all their items
+            .flatten()
+            .collect::<Vec<SyncTimelineEvent>>();
         if loaded_events.is_empty() {
             return Err(PinnedEventsLoaderError::TimelineReloadFailed);
         }
@@ -102,13 +112,14 @@ impl PinnedEventsLoader {
     }
 }
 
-pub trait PinnedEventsRoom: SendOutsideWasm + SyncOutsideWasm {
-    /// Load a single room event using the cache or network.
-    fn load_event<'a>(
+pub trait PinnedEventsRoom {
+    /// Load a single room event using the cache or network and any events
+    /// related to it, if they are cached.
+    async fn load_event_with_relations<'a>(
         &'a self,
         event_id: &'a EventId,
         request_config: Option<RequestConfig>,
-    ) -> BoxFuture<'a, Result<SyncTimelineEvent, PaginatorError>>;
+    ) -> BoxFuture<'a, Result<(SyncTimelineEvent, Vec<SyncTimelineEvent>), PaginatorError>>;
 
     /// Get the pinned event ids for a room.
     fn pinned_event_ids(&self) -> Vec<OwnedEventId>;
@@ -137,23 +148,23 @@ pub trait PinnedEventsRoom: SendOutsideWasm + SyncOutsideWasm {
 }
 
 impl PinnedEventsRoom for Room {
-    fn load_event<'a>(
+    async fn load_event_with_relations<'a>(
         &'a self,
         event_id: &'a EventId,
         request_config: Option<RequestConfig>,
-    ) -> BoxFuture<'a, Result<SyncTimelineEvent, PaginatorError>> {
+    ) -> BoxFuture<'a, Result<(SyncTimelineEvent, Vec<SyncTimelineEvent>), PaginatorError>> {
         async move {
             if let Ok((cache, _handles)) = self.event_cache().await {
-                if let Some(event) = cache.event(event_id).await {
-                    debug!("Loaded pinned event {event_id} from cache");
-                    return Ok(event);
+                if let Some(ret) = cache.event_with_relations(event_id).await {
+                    debug!("Loaded pinned event {event_id} and related events from cache");
+                    return Ok(ret);
                 }
             }
 
             debug!("Loading pinned event {event_id} from HS");
             self.event(event_id, request_config)
                 .await
-                .map(|e| e.into())
+                .map(|e| (e.into(), Vec::new()))
                 .map_err(|err| PaginatorError::SdkError(Box::new(err)))
         }
         .boxed()
--- a/crates/matrix-sdk-ui/src/timeline/pinned_events_loader.rs
+++ b/crates/matrix-sdk-ui/src/timeline/pinned_events_loader.rs
@@ -17,6 +17,7 @@ use std::{fmt::Formatter, num::NonZeroUsize, sync::Arc};
 use futures_util::{future::join_all, FutureExt as _};
 use matrix_sdk::{
     config::RequestConfig, event_cache::paginator::PaginatorError, BoxFuture, Room,
+    SendOutsideWasm, SyncOutsideWasm,
 };
 use matrix_sdk_base::deserialized_responses::SyncTimelineEvent;
 use ruma::{EventId, MilliSecondsSinceUnixEpoch, OwnedEventId};
@@ -112,10 +113,10 @@ impl PinnedEventsLoader {
     }
 }
 
-pub trait PinnedEventsRoom {
+pub trait PinnedEventsRoom: SendOutsideWasm + SyncOutsideWasm {
     /// Load a single room event using the cache or network and any events
     /// related to it, if they are cached.
-    async fn load_event_with_relations<'a>(
+    fn load_event_with_relations<'a>(
         &'a self,
         event_id: &'a EventId,
         request_config: Option<RequestConfig>,
@@ -148,7 +149,7 @@ pub trait PinnedEventsRoom {
 }
 
 impl PinnedEventsRoom for Room {
-    async fn load_event_with_relations<'a>(
+    fn load_event_with_relations<'a>(
         &'a self,
         event_id: &'a EventId,
         request_config: Option<RequestConfig>,
--- a/crates/matrix-sdk-ui/src/timeline/pinned_events_loader.rs
+++ b/crates/matrix-sdk-ui/src/timeline/pinned_events_loader.rs
@@ -23,6 +23,7 @@ use matrix_sdk_base::deserialized_responses::SyncTimelineEvent;
 use ruma::{EventId, MilliSecondsSinceUnixEpoch, OwnedEventId};
 use thiserror::Error;
 use tracing::{debug, warn};
+use crate::timeline::event_handler::TimelineEventKind;
 
 const MAX_CONCURRENT_REQUESTS: usize = 10;
 
@@ -117,6 +118,22 @@ pub trait PinnedEventsRoom: SendOutsideWasm + SyncOutsideWasm {
     /// It avoids having to clone the whole list of event ids to check a single
     /// value.
     fn is_pinned_event(&self, event_id: &EventId) -> bool;
+
+    /// Returns whether the passed `event_kind` is either an edit or a redaction
+    /// of a pinned event.
+    fn is_replacement_for_pinned_event(&self, event_kind: &TimelineEventKind) -> bool {
+        match event_kind {
+            TimelineEventKind::Message { relations, .. } => {
+                if let Some(orig_ev) = &relations.replace {
+                    self.is_pinned_event(orig_ev.event_id())
+                } else {
+                    false
+                }
+            }
+            TimelineEventKind::Redaction { redacts } => self.is_pinned_event(redacts),
+            _ => false,
+        }
+    }
 }
 
 impl PinnedEventsRoom for Room {
--- a/crates/matrix-sdk-ui/tests/integration/timeline/pinned_event.rs
+++ b/crates/matrix-sdk-ui/tests/integration/timeline/pinned_event.rs
@@ -15,7 +15,10 @@ use matrix_sdk_ui::{
     timeline::{TimelineFocus, TimelineItemContent},
     Timeline,
 };
-use ruma::{event_id, owned_room_id, MilliSecondsSinceUnixEpoch, OwnedRoomId};
+use ruma::{
+    event_id, events::room::message::RoomMessageEventContentWithoutRelation, owned_room_id,
+    MilliSecondsSinceUnixEpoch, OwnedRoomId,
+};
 use serde_json::json;
 use stream_assert::assert_pending;
 use wiremock::MockServer;
@@ -402,7 +405,11 @@ async fn test_edited_events_are_reflected_in_sync() {
     test_helper.server.reset().await;
 
     let edited_event = f
-        .replacement_msg("edited message!", event_id!("$1"))
+        .text_msg("edited message!")
+        .edit(
+            event_id!("$1"),
+            RoomMessageEventContentWithoutRelation::text_plain("* edited message!"),
+        )
         .event_id(event_id!("$2"))
         .server_ts(MilliSecondsSinceUnixEpoch::now())
         .into_timeline();
@@ -425,7 +432,7 @@ async fn test_edited_events_are_reflected_in_sync() {
         assert_eq!(index, 1);
         match value.as_event().unwrap().content() {
             TimelineItemContent::Message(m) => {
-                assert_eq!(m.body(), "edited message!")
+                assert_eq!(m.body(), "* edited message!")
             }
             _ => panic!("Should be a message event"),
         }
@@ -547,7 +554,11 @@ async fn test_edited_events_survive_pinned_event_ids_change() {
     test_helper.server.reset().await;
 
     let edited_pinned_event = f
-        .replacement_msg("edited message!", event_id!("$1"))
+        .text_msg("* edited message!")
+        .edit(
+            event_id!("$1"),
+            RoomMessageEventContentWithoutRelation::text_plain("edited message!"),
+        )
         .event_id(event_id!("$2"))
         .server_ts(MilliSecondsSinceUnixEpoch::now())
         .into_timeline();
--- a/crates/matrix-sdk-ui/tests/integration/timeline/pinned_event.rs
+++ b/crates/matrix-sdk-ui/tests/integration/timeline/pinned_event.rs
@@ -2,16 +2,20 @@ use std::time::Duration;
 
 use assert_matches::assert_matches;
 use eyeball_im::VectorDiff;
-use futures_util::StreamExt;
 use matrix_sdk::{
+    assert_next_matches_with_timeout,
     config::SyncSettings,
     sync::SyncResponse,
     test_utils::{events::EventFactory, logged_in_client_with_server},
     Client,
 };
+use matrix_sdk_base::deserialized_responses::TimelineEvent;
 use matrix_sdk_test::{async_test, JoinedRoomBuilder, StateTestEvent, SyncResponseBuilder, BOB};
-use matrix_sdk_ui::{timeline::TimelineFocus, Timeline};
-use ruma::{event_id, owned_room_id, OwnedEventId, OwnedRoomId};
+use matrix_sdk_ui::{
+    timeline::{TimelineFocus, TimelineItemContent},
+    Timeline,
+};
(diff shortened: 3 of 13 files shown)


crates/matrix-sdk-ui/src/timeline/pinned_events_loader.rs
```diff
@@ -63,20 +63,24 @@ impl PinnedEventsLoader {
...
 
         let new_events = join_all(pinned_event_ids.into_iter().map(|event_id| {
+            let semaphore = semaphore.clone();
             let provider = self.room.clone();
             async move {
-                match provider.load_event(&event_id, request_config).await {
+                if let Err(err) = semaphore.acquire().await {
+                    warn!("error when loading pinned event: {err}");
+                    return None;
+                }
```
bnjbvr, line 83: Pretty sure this is a no-op when the semaphore could be acquired, because the guard (in the `Ok()` value) is not bound, so it's immediately dropped. Can you add a test for this? (I would even recommend making a separate PR for this feature, since it seems unrelated to the current PR's purpose, and could land separately.)

crates/matrix-sdk/src/event_cache/mod.rs
```diff
@@ -396,6 +404,49 @@ impl EventCacheInner {
                 continue;
             };
 
+            // Handle and cache events and relations
+            let mut relations = self.relations.write().await;
```
bnjbvr: Unfortunately, this is not the right place to add this; I suppose this should be in the same place where I inserted events into `Self::events`. Follow the breadcrumbs :)

crates/matrix-sdk/src/event_cache/mod.rs
```diff
@@ -396,6 +404,49 @@ impl EventCacheInner {
...
+            }
+            drop(relations);
+
+            let mut events = self.events.write().await;
+            for ev in &joined_room_update.timeline.events {
+                if let Some(ev_id) = ev.event_id() {
+                    events.insert(ev_id, (room_id.clone(), ev.clone()));
+                }
+            }
+            drop(events);
```
bnjbvr: This can entirely be removed, because it's already done a few layers below.

crates/matrix-sdk/src/event_cache/mod.rs
```diff
@@ -501,6 +558,29 @@ impl RoomEventCache {
         None
     }
 
+    /// Try to find an event by id in this room, along with all relations.
+    pub async fn event_with_relations(&self, event_id: &EventId) -> Vec<SyncTimelineEvent> {
```
bnjbvr, line 583: Can you return an `(SyncTimelineEvent, Vec<SyncTimelineEvent>)` instead, so it's clear where the target event is? (Right now there's nothing either in the signature or the doc-comment that makes it obvious it's the first one.)

crates/matrix-sdk/src/event_cache/mod.rs
```diff
@@ -501,6 +558,29 @@ impl RoomEventCache {
...
+        let relations = self.inner.relations.read().await;
+        let relation_ids = relations.get(event_id);
```
bnjbvr, line 585: Can you inline this where it's used, instead? It's not useful to preload it, if the event couldn't be found.

crates/matrix-sdk/src/event_cache/mod.rs
```diff
@@ -312,6 +316,7 @@ impl EventCache {
 }
 
 type AllEventsMap = Arc<RwLock<BTreeMap<OwnedEventId, (OwnedRoomId, SyncTimelineEvent)>>>;
+type RelationsMap = Arc<RwLock<BTreeMap<OwnedEventId, BTreeSet<OwnedEventId>>>>;
```
bnjbvr, line 319: There are two locks for data that's touched around the same time; I'd suggest grouping the two fields inside a data structure, and using a single lock, to avoid having to think about lock ordering.

crates/matrix-sdk/src/event_cache/mod.rs
```diff
@@ -396,6 +404,49 @@ impl EventCacheInner {
                 continue;
             };
 
+            // Handle and cache events and relations
+            let mut relations = self.relations.write().await;
+            for ev in &joined_room_update.timeline.events {
+                if let Ok(AnySyncTimelineEvent::MessageLike(ev)) = ev.event.deserialize() {
+                    if let Some(AnyMessageLikeEventContent::RoomMessage(c)) = ev.original_content()
```
bnjbvr, line 411: Can you use a `match` statement so we get a single call to `ev.original_content()`?

Also, do you want to keep reactions too?

crates/matrix-sdk/src/event_cache/mod.rs
```diff
@@ -396,6 +404,49 @@ impl EventCacheInner {
...
+                    {
+                        if let Some(Relation::Replacement(relation)) = c.relates_to {
+                            if let Some(replacements) = relations.get_mut(&relation.event_id) {
+                                replacements.insert(ev.event_id().to_owned());
+                            } else {
+                                relations.insert(
+                                    relation.event_id,
+                                    BTreeSet::from_iter(vec![ev.event_id().to_owned()]),
+                                );
+                            }
```
bnjbvr: You can probably use `relations.entry(relation.event_id.clone()).or_default().insert(…)` to make this a bit shorter to read.

crates/matrix-sdk/src/event_cache/mod.rs
```diff
@@ -396,6 +404,49 @@ impl EventCacheInner {
...
+                        if let Some(redacted_ev_id) = c.redacts {
+                            let original_ev_id = redacted_ev_id;
+                            if let Some(replacements) = relations.get_mut(&original_ev_id) {
+                                replacements.insert(ev.event_id().to_owned());
+                            } else {
+                                relations.insert(
+                                    original_ev_id,
+                                    BTreeSet::from_iter(vec![ev.event_id().to_owned()]),
+                                );
+                            }
```
bnjbvr: Same remark for `.entry()`.

crates/matrix-sdk/src/room/mod.rs
```diff
@@ -445,6 +445,21 @@ impl Room {
         )
         .await?;
 
+        // Save the loaded events into the event cache, if it's set up.
```
bnjbvr, line 449: Nice! Can you test this please? :)

Also consider adding a `RoomEventCache::save_events` method, so that we can save all the events at once without giving and releasing and giving and releasing etc. the lock.

crates/matrix-sdk-ui/src/timeline/pinned_events_loader.rs
```diff
@@ -74,6 +77,7 @@ impl PinnedEventsLoader {
             let provider = self.room.clone();
             async move {
                 match provider.load_event(&event_id, request_config).await {
+                match provider.load_event_with_relations(&event_id, request_config).await {
```
bnjbvr: uh?

crates/matrix-sdk-ui/src/timeline/pinned_events_loader.rs
```diff
@@ -124,22 +130,23 @@ pub trait PinnedEventsRoom: SendOutsideWasm + SyncOutsideWasm {
...
+                return Ok(cached_event_with_relations);
             }
         }
 
         debug!("Loading pinned event {event_id} from HS");
-        self.event(event_id, request_config)
+        self.event_with_context(event_id, true, uint!(0), request_config)
             .await
-            .map(|e| e.into())
+            .map(|e| if let Some(ev) = e.event { vec![ev.into()] } else { Vec::new() })
```
bnjbvr, line 169: nit: use a better name of `e`, and what's happening to related events here?

crates/matrix-sdk/src/event_cache/mod.rs
```diff
@@ -396,6 +404,49 @@ impl EventCacheInner {
...
+                                relations.insert(
+                                    relation.event_id,
+                                    BTreeSet::from_iter(vec![ev.event_id().to_owned()]),
+                                );
+                            }
+                        }
+                    } else if let Some(AnyMessageLikeEventContent::RoomRedaction(c)) =
+                        ev.original_content()
+                    {
+                        if let Some(redacted_ev_id) = c.redacts {
```
bnjbvr, line 443: It's optional because in previous versions than v11, the redacts field was somewhere else, if I remember correctly (in the event itself, not the content). Can you handle that case too?

crates/matrix-sdk-ui/tests/integration/main.rs
```diff
@@ -84,7 +84,7 @@ async fn mock_context(
 }
 
 /// Mocks the /event endpoint
-#[allow(clippy::too_many_arguments)] // clippy you've got such a fixed mindset
+#[allow(clippy::too_many_arguments, dead_code)] // clippy you've got such a fixed mindset
```
bnjbvr, line 87: Why the `dead_code`? If it's unused, let's get rid of it.

crates/matrix-sdk-ui/tests/integration/timeline/pinned_event.rs
```diff
@@ -355,20 +599,56 @@ impl TestHelper {
...
         let mut joined_room_builder = JoinedRoomBuilder::new(&self.room_id);
-        for (id, txt, add_to_timeline) in text_messages {
-            let event_id: OwnedEventId = id.try_into().unwrap();
-            let f = EventFactory::new().room(&self.room_id);
-            let event_builder = f.text_msg(txt).event_id(&event_id).sender(*BOB);
-            mock_event(&self.server, &self.room_id, &event_id, event_builder.into_timeline()).await;
+        for (message_type, add_to_timeline) in text_messages {
+            let (event_id, timeline_item) = match message_type {
+                MessageType::New { id, txt } => {
+                    let f = EventFactory::new().room(&self.room_id);
```
bnjbvr, line 609: nit: reuse the same `EventFactory` instead of creating a new one each time.

crates/matrix-sdk-ui/tests/integration/timeline/pinned_event.rs
```diff
@@ -355,20 +599,56 @@ impl TestHelper {
...
+                    let event_builder = f
+                        .text_msg(txt)
+                        .event_id(&id)
+                        .sender(*BOB)
```
bnjbvr, line 613: Since the sender is always `BOB`, you can call `f.sender(*BOB)` on the `EventFactory` itself.

crates/matrix-sdk/src/test_utils/events.rs
```diff
@@ -213,6 +219,17 @@ impl EventFactory {
...
+    /// Create a replacement message for an existing one given its id.
+    pub fn replacement_msg(
+        &self,
+        new_content: impl Into<String>,
+        original_event_id: &EventId,
+    ) -> EventBuilder<RoomMessageEventContent> {
+        let msg_content = RoomMessageEventContent::text_plain(new_content.into());
+        let builder = self.event(msg_content.clone());
+        builder.replaces(original_event_id, msg_content)
+    }
```
bnjbvr, line 231: Is this function needed? It's just a shortcut to other functions which can be used directly, and it would avoid adding too much abstraction in the tests themselves.

crates/matrix-sdk/src/test_utils/events.rs
```diff
@@ -115,6 +115,12 @@ impl EventBuilder<RoomMessageEventContent> {
             Some(Relation::Reply { in_reply_to: InReplyTo::new(event_id.to_owned()) });
         self
     }
+
+    pub fn replaces(mut self, event_id: &EventId, new_content: RoomMessageEventContent) -> Self {
```
bnjbvr, line 119: Can you add a doc-comment here please?

crates/matrix-sdk-ui/tests/integration/timeline/pinned_event.rs
```diff
@@ -402,3 +682,28 @@ impl TestHelper {
         self.client.sync_once(self.sync_settings.clone()).await
     }
 }
+
+enum MessageType {
+    New { id: OwnedEventId, txt: String },
+    Edit { id: OwnedEventId, old_id: OwnedEventId, txt: String },
+    Redaction { id: OwnedEventId, old_id: OwnedEventId },
+}
```
bnjbvr, line 690: Instead of introducing all of this, could the caller make use of an `EventFactory` immediately, and not pass a `MessageType` etc.? This would make the tests much more straightforward to read…

crates/matrix-sdk-ui/tests/integration/timeline/pinned_event.rs
```diff
@@ -355,20 +599,56 @@ impl TestHelper {
...
         pinned_event_ids: Option<Vec<&str>>,
     ) -> Result<SyncResponse, matrix_sdk::Error> {
         let mut joined_room_builder = JoinedRoomBuilder::new(&self.room_id);
-        for (id, txt, add_to_timeline) in text_messages {
-            let event_id: OwnedEventId = id.try_into().unwrap();
-            let f = EventFactory::new().room(&self.room_id);
-            let event_builder = f.text_msg(txt).event_id(&event_id).sender(*BOB);
-            mock_event(&self.server, &self.room_id, &event_id, event_builder.into_timeline()).await;
+        for (message_type, add_to_timeline) in text_messages {
+            let (event_id, timeline_item) = match message_type {
```
bnjbvr, line 607: nit: not `timeline_item`, but `timeline_event`, I think; item should be reserved to the UI Timeline object :)

crates/matrix-sdk-ui/tests/integration/timeline/pinned_event.rs
```diff
@@ -319,6 +329,240 @@ async fn test_pinned_timeline_with_no_pinned_event_ids_is_just_empty() {
...
+    let (items, mut timeline_stream) = timeline.subscribe().await;
+
+    assert_eq!(items.len(), 1 + 1); // event item + a day divider
+    assert!(items[0].is_day_divider());
+    assert_eq!(items[1].as_event().unwrap().content().as_message().unwrap().body(), "in the end");
+    assert_pending!(timeline_stream);
+    test_helper.server.reset().await;
+
+    // Load new pinned event contents from sync, where $2 is and edit on $1
+    let _ = Box::pin(
```
bnjbvr, line 370: Why is the `Box::pin` needed? Can you do without?

crates/matrix-sdk-ui/tests/integration/timeline/pinned_event.rs
```diff
@@ -319,6 +329,240 @@ async fn test_pinned_timeline_with_no_pinned_event_ids_is_just_empty() {
...
+        test_helper.setup_sync_response(vec![(edit("$2", "$1", "edited message!"), true)], None),
+    )
+    .await;
+
+    // The list is reloaded, so it's reset
+    assert_matches!(timeline_stream.next().await.unwrap(), VectorDiff::Clear);
```
bnjbvr, line 376: I think there's an `assert_next` that exists and could replace most of these `assert_matches` invokes here and below, and in the other test(s) too?

crates/matrix-sdk/src/event_cache/mod.rs
```diff
@@ -496,6 +504,18 @@ impl EventCacheInner {
...
+    async fn insert_or_update_relationship(&self, orig_ev_id: &EventId, related_ev_id: &EventId) {
+        let mut relations = self.relations.write().await;
+
+        if let Some(replacements) = relations.get_mut(orig_ev_id) {
+            replacements.insert(related_ev_id.to_owned());
+        } else {
+            relations
+                .insert(orig_ev_id.to_owned(), BTreeSet::from_iter(vec![related_ev_id.to_owned()]));
+        }
+        drop(relations);
```
bnjbvr: nit: no need to drop explicitly, it'll happen automatically at the end of the function's body.

crates/matrix-sdk/src/event_cache/mod.rs
```diff
@@ -496,6 +504,18 @@ impl EventCacheInner {
             }
         }
     }
+
+    async fn insert_or_update_relationship(&self, orig_ev_id: &EventId, related_ev_id: &EventId) {
```
bnjbvr, line 509: nit: please fully spell out `event_id` for consistency, and `original` too

(I for one would call these two parameters `original` and `related`, because including the parameter's type is spurious, since the parameter type is next to the variable name, but that's my preference :))

benchmarks/benches/room_bench.rs
```diff
@@ -186,6 +187,8 @@ pub fn load_pinned_events_benchmark(c: &mut Criterion) {
 
     group.bench_function(BenchmarkId::new("load_pinned_events", name), |b| {
         b.to_async(&runtime).iter(|| async {
+            client.event_cache().subscribe().unwrap();
```
bnjbvr, line 190: Is there really a need to do it for *each* iteration of the benchmark? I don't think so, it should be subscribed to once and for all.

bnjbvr (changes requested):
Looks like you're on the right track! I'd like a few test changes, and the main logic in the `EventCache` itself isn't tested independently. Please add some tests for that too, since it's a different layer than the timeline.

Also, please reword the PR title to follow our conventional commit guidelines :)
Thanks!

crates/matrix-sdk-ui/src/timeline/pinned_events_loader.rs
jmartinesp: I have no idea how to test that, to be honest. Since there is no timeout I think the execution will just get stuck there until a new permit is available.

crates/matrix-sdk/src/event_cache/mod.rs
jmartinesp: I think I'll have to do that in a separate method then? I don't think I can have 2 `RwLockWriteGuards` at the same time, one for `all_events` and another one for `relations`, so I don't think I'll be able to add this on `append_events_locked_impl`.

jmartinesp: Oh, I just saw https://github.com/matrix-org/matrix-rust-sdk/pull/3840#discussion_r1716841925.

crates/matrix-sdk-ui/src/timeline/pinned_events_loader.rs
jmartinesp: I think it was a rebase hiccup, the code is not there later 😅 .

crates/matrix-sdk-ui/tests/integration/main.rs
jmartinesp: It's a helper for mocking data, so I expected it to be useful in the future. I can remove it though.

crates/matrix-sdk/src/test_utils/events.rs
jmartinesp: It's a helper, so I expect it to contain boilerplate to avoid repetitive code. Should I remove the `redaction` fn above too?

crates/matrix-sdk/src/test_utils/events.rs
jmartinesp: Oh, sure. I though I had handled all the pub fns but I missed this one at least.

crates/matrix-sdk-ui/tests/integration/timeline/pinned_event.rs
jmartinesp: That means inlining the whole `setup_sync_response` code for every sync. I honestly think that's way less readable, and it's the main reason why I abstracted it inside `TestHelper` in a previous PR.

crates/matrix-sdk-ui/tests/integration/timeline/pinned_event.rs
jmartinesp: Oops, leftover from one of my tests to check why the tests were stuck.

crates/matrix-sdk/src/event_cache/mod.rs
jmartinesp: Yep, copy&paste issue. It's gone now.

crates/matrix-sdk/src/event_cache/mod.rs
jmartinesp: I'm removing this code so it shouldn't be needed anymore.

crates/matrix-sdk-ui/tests/integration/timeline/pinned_event.rs
jmartinesp: For some reason changing `assert_matches!(timeline_stream.next().await.unwrap(), *)` to `assert_next_matches!(timeline_stream, *)` makes one of the tests fail with:

> assertion failed: stream is not ready, it's pending

Maybe I need to add a delay before? It's quite weird.

jmartinesp: Oh, it's because under the hood it calls `Stream::next_now_or_never`... That seems dangerous. I don't see any similar function for doing it while awaiting for the next item.

benchmarks/benches/room_bench.rs
jmartinesp: Maybe I can wrap `Client` in an `Arc` and use `runtime.spawn_blocking` to run that code? Is there a better solution?

crates/matrix-sdk/src/event_cache/mod.rs
jmartinesp: Reactions weren't planned for the pined events feature, but I can try to add them too.

crates/matrix-sdk/src/event_cache/mod.rs
jmartinesp: So... this seems way more difficult than expected. In the `EventCache` I have no way to access the `RoomBaseInfo` to check the room version.

jmartinesp: Maybe I can try not relying on the version id and just checking the event: if the redaction is in the event itself, use it, otherwise, check for the redaction inside the contents.

crates/matrix-sdk-ui/tests/integration/timeline/pinned_event.rs
bnjbvr: Can you pass the `SyncTimelineEvent`s or `TimelineEvent`s to `setup_sync_responses`?

crates/matrix-sdk-ui/tests/integration/timeline/pinned_event.rs
bnjbvr: We need to add a timeout then, otherwise if the stream item doesn't come immediately, then the test will be timing out with `cargo-nextest`'s delay, which is a minute or longer.

crates/matrix-sdk-ui/src/timeline/pinned_events_loader.rs
jmartinesp: I replaced this with a buffered stream now. It seems to work just as fine and there is no need for the semaphore.

crates/matrix-sdk/src/event_cache/mod.rs
jmartinesp: Added reactions and polls.

crates/matrix-sdk-ui/tests/integration/timeline/pinned_event.rs
jmartinesp: It's done that way now.

crates/matrix-sdk-ui/tests/integration/timeline/pinned_event.rs
jmartinesp: Added a new `assert_next_matches_with_timeout!` macro for this. It's based on the `assert_next_matches`, but instead of checking the immediate result of the stream, it adds a small timeout which can be customised.

jmartinesp:
So, funny thing: it seems like when the devs reported `/event` didn't return the sender info it was a mistake. So I'm wondering if I should go back to using `/event` for loading the pinned events since it *was* faster...

benchmarks/benches/room_bench.rs
jmartinesp: Tried fixing this in https://github.com/matrix-org/matrix-rust-sdk/pull/3840/commits/807c87525854af90989e9920f5d5ed07254ed60a, please let me know if there's a better solution.

crates/matrix-sdk/src/room/mod.rs
jmartinesp: Done in https://github.com/matrix-org/matrix-rust-sdk/pull/3840/commits/919b5458fe886ec96cab4de9fe94b66bd50c404a, added some tests.

jmartinesp:
Ok, so I re-used this PR for holding only the part of the code related to the live updates to the pinned events timeline. It depends on a previous PR being merged and it'll have a following PR containing the semaphore/buffered futures changes.

jmartinesp:
There is something weird happening with the CI actions, only a couple of them are running 🫤 .

jmartinesp:
The PR this one depended on has been merged and this one has been rebased, so it's ready to review again.

jmartinesp:
Well, apparently https://github.com/matrix-org/matrix-rust-sdk/pull/3879 just broke this PR, another rebase will be needed.

crates/matrix-sdk/src/event_cache/mod.rs
```diff
@@ -582,6 +582,23 @@ impl RoomEventCache {
             warn!("couldn't save event without event id in the event cache");
         }
     }
+
+    /// Save some events in the event cache, for further retrieval with
+    /// [`Self::event`]. This function will save them using a single lock,
+    /// as opposed to [`Self::save_event`].
+    // This doesn't insert the event into the linked chunk. In the future there'll
+    // be no distinction between the linked chunk and the separate cache.
```
Hywan, line 590: Why?
Can we add a `TODO` and open an issue to track this please?

Hywan (changes requested):
We are almost good!

crates/matrix-sdk/src/event_cache/mod.rs
jmartinesp: The comment is just a copy of the one above it written by @bnjbvr since it follows the same logic, but I think it's probably because we lack the context of where the `LinkedChunk` containing events to save would go, or if it even makes sense to add them to one (these events can come from very different sources and may not be related enough to be added as a chunk of a timeline, if I understood how `LinkedChunk` works properly).

I guess the 'in the future' part means once we have a persistent event cache, but I'm not 100% sure.

Anyway,  can create an issue for that.

jmartinesp: @Hywan I added the TODO [here](https://github.com/matrix-org/matrix-rust-sdk/pull/3840/commits/4f57f54b7b7dfba0894aa16c5b7547a104155771) and opened the issue [here](https://github.com/matrix-org/matrix-rust-sdk/issues/3886).

benchmarks/benches/room_bench.rs
Hywan: Excellent, thanks!

crates/matrix-sdk/src/event_cache/mod.rs
Hywan: Excellent, thanks!

Hywan (approved):
Good to me, thanks! Please rebase and you can merge.

PR #3843, opened 2024-08-14
feat(sdk): Remove `NotificationSettings::subscribe_to_changes`

Hywan:
This patch removes `NotificationSettings::subscribe_to_changes` because it's not used anywhere in our code except in tests. It is indeed part of the public API but I'm not aware of anyone using it for the moment. It only adds complexity in the code.


andybalaam (approved):
Yay for deleting unused code!

PR #3846, opened 2024-08-15
crypto: Remember whether our own identity has been verified

### LINKED ISSUE #3845: `UserIdentity::has_verification_violation` does not work for our own identity
https://github.com/matrix-org/matrix-rust-sdk/pull/3795 added a new method `UserIdentity::has_verification_violation`.

However, it appears to have omitted support for tracking verification of the user's *own* identity. This means that if our own identity was rotated, and a new device added, signed by the new identity, we would share messages to that device despite there being no evidence it is not controlled by an attacker.

This will become less of a problem in future, when we refuse to do any operations without our own identity being verified, but in the meantime it is a security hole, and, for as long as matrix-sdk-crypto has support for sending messages without verifying our identity, it will remain a footgun. I think it therefore needs fixing as a prerequisite to https://github.com/matrix-org/matrix-rust-sdk/issues/3793.

richvdh:
A follow-up to #3795, which extends the same principle to `OwnUserIdentity`.

Fixes #3845.

### DIFF
--- a/crates/matrix-sdk-crypto/src/identities/user.rs
+++ b/crates/matrix-sdk-crypto/src/identities/user.rs
@@ -29,11 +29,10 @@ use ruma::{
     },
     DeviceId, EventId, OwnedDeviceId, OwnedUserId, RoomId, UserId,
 };
-use serde::{Deserialize, Serialize};
+use serde::{Deserialize, Deserializer, Serialize};
 use serde_json::Value;
 use tracing::error;
 
-use super::{atomic_bool_deserializer, atomic_bool_serializer};
 use crate::{
     error::SignatureError,
     store::{Changes, IdentityChanges, Store},
@@ -786,11 +785,18 @@ pub struct OwnUserIdentityData {
     master_key: Arc<MasterPubkey>,
     self_signing_key: Arc<SelfSigningPubkey>,
     user_signing_key: Arc<UserSigningPubkey>,
-    #[serde(
-        serialize_with = "atomic_bool_serializer",
-        deserialize_with = "atomic_bool_deserializer"
-    )]
-    verified: Arc<AtomicBool>,
+    #[serde(deserialize_with = "deserialize_own_user_identity_data_verified")]
+    verified: Arc<RwLock<OwnUserIdentityVerifiedState>>,
+}
+
+#[derive(Default, Debug, Serialize, Deserialize, PartialEq, Eq)]
+enum OwnUserIdentityVerifiedState {
+    /// We have never verified our own identity
+    #[default]
+    NeverVerified,
+
+    /// We have verified the current identity.
+    Verified,
 }
 
 impl PartialEq for OwnUserIdentityData {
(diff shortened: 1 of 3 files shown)


crates/matrix-sdk-crypto/src/identities/user.rs
```diff
@@ -940,17 +961,42 @@ impl OwnUserIdentityData {
...
-        self.verified.load(Ordering::SeqCst)
+        *self.verified.read().unwrap() == OwnUserIdentityVerifiedState::Verified
+    }
+
+    /// True if we verified our own identity at some point in the past.
+    ///
+    /// To reset this latch back to `false`, one must call
+    /// [`OwnUserIdentityData::withdraw_verification()`].
+    pub fn was_previously_verified(&self) -> bool {
+        let verified_state = (*self.verified.read().unwrap()).clone();
```
andybalaam, line 985: Do we need to clone here?

crates/matrix-sdk-crypto/src/identities/user.rs
```diff
@@ -940,17 +961,42 @@ impl OwnUserIdentityData {
...
+        verified_state == OwnUserIdentityVerifiedState::Verified
+            || verified_state == OwnUserIdentityVerifiedState::PreviouslyVerifiedButNoLonger
```
andybalaam, line 987: Could use the `matches!` macro with `|`, but I'm not sure it would be better.

crates/matrix-sdk-crypto/src/identities/user.rs
```diff
@@ -1004,6 +1050,44 @@ impl OwnUserIdentityData {
...
+    ///
+    /// Such a violation should be reported to the local user by the
+    /// application, and resolved by
+    ///
+    /// - Verifying the new identity with
+    ///   [`OwnUserIdentity::request_verification`]
+    /// - Or by withdrawing the verification requirement
+    ///   [`OwnUserIdentity::withdraw_verification`].
+    pub fn has_verification_violation(&self) -> bool {
+        self.was_previously_verified() && !self.is_verified()
```
andybalaam, line 1064: A `matches!` macro here would be minutely more efficient, and maybe also easier to understand?

andybalaam (approved):
Looks good; a couple of minor suggestions.

crates/matrix-sdk-crypto/src/identities/user.rs
richvdh: The problem is that the stored verification state is behind a mutex, so we can't just take a reference to it without holding a separate lock guard. So I think our options are
 * clone, like this
 * use a separate variable to hold a lock guard while we do the comparison
 * `matches!`, as you suggest below

`OwnUserIdentityData` is a unit enum, so cloning it is just copying the enum discriminator -- in other words, loading it into a register (which we'd have to do anyway).

In terms of readability, I'm not sure that any of the others are any better. YMMV?

crates/matrix-sdk-crypto/src/identities/user.rs
richvdh: so it would. Thanks.

crates/matrix-sdk-crypto/src/identities/user.rs
richvdh: fair enough. will make it so.

PR #3847, opened 2024-08-15
Crypto: add helper methods to `UserIdentities`

richvdh:
We have a bunch of methods which are the same in both `OtherUserIdentity` and `OwnUserIdentity`, so add some convenience methods to access them.

### DIFF
--- a/crates/matrix-sdk-crypto/src/identities/manager.rs
+++ b/crates/matrix-sdk-crypto/src/identities/manager.rs
@@ -2137,13 +2137,8 @@ pub(crate) mod tests {
         let txn_id = TransactionId::new();
         machine.mark_request_as_sent(&txn_id, &keys_query).await.unwrap();
 
-        let carol_identity = machine
-            .get_identity(DataSet::carol_id(), None)
-            .await
-            .unwrap()
-            .unwrap()
-            .other()
-            .unwrap();
+        let carol_identity =
+            machine.get_identity(DataSet::carol_id(), None).await.unwrap().unwrap();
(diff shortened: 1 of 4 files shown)


crates/matrix-sdk-crypto/src/identities/manager.rs
```diff
@@ -2213,43 +2208,30 @@ pub(crate) mod tests {
         let own_identity =
             machine.get_identity(DataSet::own_id(), None).await.unwrap().unwrap().own().unwrap();
```
andybalaam, line 2209: This line looks similar: can it be simplified too?

crates/matrix-sdk-crypto/src/identities/manager.rs
```diff
@@ -2294,19 +2276,13 @@ pub(crate) mod tests {
...
```
andybalaam, line 2276: Also this one?

andybalaam (approved):
Looks good; 2 questions.

crates/matrix-sdk-crypto/src/identities/manager.rs
richvdh: no. We need to call `own_identity.is_identity_signed()`; `is_identity_signed` is specific to `OwnUserIdentity`.

crates/matrix-sdk-crypto/src/identities/manager.rs
richvdh: yeah, fair. Done in 67b221ab4

PR #3849, opened 2024-08-16
WIP: Update SenderData on `/keys/query` response

richvdh:


crates/matrix-sdk-crypto/src/identities/manager.rs
```diff
@@ -162,6 +162,27 @@ impl IdentityManager {
...
+        // user identities from the store, so this has to happen *after* the
+        // changes from `handle_cross_signing_keys` are saved.
+        //
+        // FIXME: This thing is racy. If a session is created at the same time as a
+        //   `/keys/query` response is being processed, it could be saved
+        //   without up-to-date sender data, but it might be saved too late for
+        //   it to be picked up by `update_sender_data_from_device_changes`.
+        //
+        // FIXME: Come to that, what if this races against some other operation on the
+        //   session, such as marking it as needing backup? It's all a bit terrifying.
```
andybalaam, line 178: Can we be any clearer what will happen if one of these races happens? I think we will lose one of the writes, so e.g. forget that a session needs backup, or forget that we trust the sender?

crates/matrix-sdk-crypto/src/identities/manager.rs
```diff
@@ -1019,6 +1040,126 @@ impl IdentityManager {
...
+                    .update_sender_data_for_sessions_for_device(device, SenderDataType::DeviceInfo)
+                    .await?;
+                result.append(&mut result1);
+            }
+        }
+
+        Ok(result)
+    }
+
+    /// Given a list of changed devices, look for [`InboundGroupSession`]s
```
andybalaam, line 1091: "Given a device that has changed"?

crates/matrix-sdk-crypto/src/identities/manager.rs
```diff
@@ -1019,6 +1040,126 @@ impl IdentityManager {
...
+        device: &DeviceData,
+        sender_data_type: SenderDataType,
+    ) -> Result<Vec<InboundGroupSession>, CryptoStoreError> {
+        const IGS_BATCH_SIZE: usize = 50;
+
+        let mut result = Vec::new();
+        let Some(curve_key) = device.curve25519_key() else { return Ok(result) };
+
+        let mut last_session_id: Option<String> = None;
+        loop {
```
andybalaam, line 1108: This uses an unbounded amount of memory if there are lots of sessions for this device. Can we loop at a higher level to avoid this? We'll need to do the save_changes inside the loop I guess.

crates/matrix-sdk-crypto/src/identities/manager.rs
```diff
@@ -1019,6 +1040,126 @@ impl IdentityManager {
...
+        for device in device_changes.new.iter().chain(device_changes.changed.iter()) {
+            // 1. Look for InboundGroupSessions from the device whose sender_data is
+            //    UnknownDevice. For such sessions, we now have the device, and can update
+            //    the sender_data accordingly.
+            //
+            // In theory, we only need to do this for new devices. In practice, I'm a bit
+            // worried about races leading us to getting stuck in the
+            // UnknownDevice state, so we'll paper over that by doing this check
+            // on device updates too.
+            let mut result1 = self
```
andybalaam, line 1065: `result1` -> `sessions_unknown` or similar?

crates/matrix-sdk-crypto/src/identities/manager.rs
```diff
@@ -1019,6 +1040,126 @@ impl IdentityManager {
...
+            //    cross-signing key for the owner), look for InboundGroupSessions from the
+            //    device whose sender_data is DeviceInfo. We can also update the sender_data
+            //    for these sessions.
+            //
+            // In theory, we can skip a couple of steps of the SenderDataFinder algorithm,
+            // because we're doing the cross-signing check here. In practice,
+            // it's *way* easier just to use the same logic.
+            let device_owner_identity = self.store.get_user_identity(device.user_id()).await?;
+            if device_owner_identity.is_some_and(|id| device.is_cross_signed_by_owner(&id)) {
+                let mut result1 = self
```
andybalaam, line 1081: `result1` -> `sessions_device_info` or similar

crates/matrix-sdk-crypto/src/identities/manager.rs
```diff
@@ -2286,4 +2427,81 @@ pub(crate) mod tests {
...
+                .update_sender_data_from_device_changes(&DeviceChanges {
+                    changed: vec![device_data],
+                    ..Default::default()
+                })
+                .await
+                .unwrap();
+
+            // Then those sessions should be updated
+            assert_eq!(
+                updated_sessions.iter().map(|s| s.session_id()).collect::<HashSet<_>>(),
```
andybalaam, line 2484: A little helper function called something like "session_ids" would make this much easier for me to read, if you can be bothered.

crates/matrix-sdk-crypto/src/machine/tests/megolm_sender_data.rs
```diff
@@ -105,6 +108,105 @@ async fn test_receive_megolm_session_from_known_device() {
     );
 }
 
+/// If we have a megolm session from an unknown device, test what happens when
+/// we get a /keys/query response that includes that device.
+#[async_test]
+async fn test_update_unknown_device_senderdata_on_keys_query() {
+    // GIVEN we have a megolm session from an unknown device
```
andybalaam, line 115: GIVEN -> Given (etc below)

andybalaam (changes requested):
One conversation to be had about unbounded memory usage; all good otherwise!

crates/matrix-sdk-crypto/src/identities/manager.rs
richvdh: 8766b2334

crates/matrix-sdk-crypto/src/identities/manager.rs
richvdh: Yeah. 74e63219d changes this so that we save each session separately. I'm not entirely convinced by that -- presumably there is a reason that we attempt to batch changes together into a single `save_changes`. But a middle ground is hard to achieve

crates/matrix-sdk-crypto/src/identities/manager.rs
richvdh: removed as a side-effect of 74e63219d

crates/matrix-sdk-crypto/src/identities/manager.rs
richvdh: removed as a side-effect of 74e63219d

crates/matrix-sdk-crypto/src/identities/manager.rs
richvdh: removed as a side-effect of 74e63219d

crates/matrix-sdk-crypto/src/machine/tests/megolm_sender_data.rs
richvdh: 6666eb817

andybalaam (approved):
Great, thanks!

PR #3851, opened 2024-08-19
fix(ffi): New fields formatter to remove duplicated span fields

### LINKED ISSUE #3494: All span fields in logs are appearing twice
Watching the logs, I see something like:

```
next_sync_with_lock{store_generation=60 store_generation=60} > sync_once{pos="1" pos="1"} > send{server_versions=[V1_0, V1_1, V1_2, V1_3, V1_4, V1_5] config=RequestConfig { timeout: 60s } config=RequestConfig { timeout: 60s } request_id="REQ-4" request_id="REQ-4" method=POST method=POST uri="https://slidingsync.lab.matrix.org/_matrix/client/unstable/org.matrix.msc3575/sync" uri="https://slidingsync.lab.matrix.org/_matrix/client/unstable/org.matrix.msc3575/sync" request_size="72 B" request_size="72 B" request_body=b"{\"conn_id\":\"encryption\",\"extensions\":{\"to_device\":{\"since\":\"59340139\"}}}" request_body=b"{\"conn_id\":\"encryption\",\"extensions\":{\"to_device\":{\"since\":\"59340139\"}}}" status=200 status=200 response_size="180 B" response_size="180 B"}
```

where most fields (like `store_generation`, `pos`, `request_id`, `method`, `uri`, `request_size`, `request_body`, `status` and `response_size`) are appearing twice. They must appear only once.

 Note that the `server_version` field appears only once strangely.

Hywan:
This patch fixes a bug in the tracing system. It introduces one
fields formatter _per layer_ to force the fields to be recorded in
different span extensions, and thus to remove the duplicated fields in
`FormattedFields`.

The patch contains links to the bug report in `tokio-rs/tracing`. This
patch is a workaround.

---

* Fix #3494

### DIFF
--- a/bindings/matrix-sdk-ffi/src/platform.rs
+++ b/bindings/matrix-sdk-ffi/src/platform.rs
@@ -97,17 +97,19 @@ where
 
             if let Some(scope) = ctx.event_scope() {
                 writer.write_str(" | spans: ")?;
+
                 let mut first = true;
 
                 for span in scope.from_root() {
                     if !first {
                         writer.write_str(" > ")?;
                     }
-                    first = false;
-                    write!(writer, "{}", span.metadata().name())?;
 
-                    let ext = span.extensions();
-                    if let Some(fields) = &ext.get::<FormattedFields<N>>() {
+                    first = false;
+
+                    write!(writer, "{}", span.name())?;
+
+                    if let Some(fields) = &span.extensions().get::<FormattedFields<N>>() {
                         if !fields.is_empty() {
                             write!(writer, "{{{fields}}}")?;
                         }
--- a/bindings/matrix-sdk-ffi/src/platform.rs
+++ b/bindings/matrix-sdk-ffi/src/platform.rs
@@ -1,7 +1,13 @@
 use tracing_appender::rolling::{RollingFileAppender, Rotation};
 use tracing_core::Subscriber;
 use tracing_subscriber::{
-    fmt::{self, time::FormatTime, FormatEvent, FormatFields, FormattedFields},
+    field::RecordFields,
+    fmt::{
+        self,
+        format::{DefaultFields, Writer},
+        time::FormatTime,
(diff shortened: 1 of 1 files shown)


bindings/matrix-sdk-ffi/src/platform.rs
```diff
@@ -145,15 +171,34 @@ where
...
+
+            impl<'writer> FormatFields<'writer> for FieldsFormatterFormStdoutOrSystem {
+                fn format_fields<R: RecordFields>(
+                    &self,
+                    writer: Writer<'writer>,
+                    fields: R,
+                ) -> std::fmt::Result {
+                    self.0.format_fields(writer, fields)
+                }
+            }
```
bnjbvr, line 189: nit: we don't need to duplicate the struct definition, if we hoist it outside the functions and reuse it in both places. A better name would need to be found then.

bnjbvr (changes requested):
Sorry for the drive-by :innocent:

bindings/matrix-sdk-ffi/src/platform.rs
Hywan: indeed!

bnjbvr (approved):
\o/

stefanceriu:
Tested it out on ElementX iOS, works on `10ab0face912e5e352f0bef757e9f280da23ba3d` but unfortunately no longer works on `10ab0face912e5e352f0bef757e9f280da23ba3d` ¯\_(ツ)_/¯

Hywan:
Last commit has been removed. Ready to merge.

bindings/matrix-sdk-ffi/src/platform.rs
Hywan: As tested by @stefanceriu in https://github.com/matrix-org/matrix-rust-sdk/pull/3851#issuecomment-2296453939, using the same type sadly doesn't work. Let's use 2 types then.

PR #3852, opened 2024-08-19
timeline: use the `EncryptionInfo` provided by the replacement event when processing edits

stefanceriu:
- this prevents issues where spoofing the sender field is enough to spoof and edit and display wrong decorations in the app
- fixes matrix-org/internal-config/issues/1549

### DIFF
--- a/crates/matrix-sdk-ui/src/timeline/event_handler.rs
+++ b/crates/matrix-sdk-ui/src/timeline/event_handler.rs
@@ -543,7 +543,21 @@ impl<'a, 'o> TimelineEventHandler<'a, 'o> {
             };
 
             trace!("Applying edit");
-            Some(event_item.with_content(new_content, edit_json))
+
+            if let EventTimelineItemKind::Remote(remote_event) = &event_item.kind {
+                let new_encryption_info = match &this.ctx.flow {
+                    Flow::Local { .. } => None,
+                    Flow::Remote { encryption_info, .. } => encryption_info.clone(),
+                };
+
+                Some(event_item.with_content(new_content, edit_json).with_kind(
+                    EventTimelineItemKind::Remote(
+                        remote_event.with_encryption_info(new_encryption_info),
+                    ),
+                ))
+            } else {
+                Some(event_item.with_content(new_content, edit_json))
+            }
         });
 
         if !found {
--- a/crates/matrix-sdk-ui/src/timeline/tests/edit.rs
+++ b/crates/matrix-sdk-ui/src/timeline/tests/edit.rs
@@ -12,10 +12,16 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use std::collections::BTreeMap;
+
 use assert_matches2::assert_let;
 use eyeball_im::VectorDiff;
+use matrix_sdk::deserialized_responses::{
+    AlgorithmInfo, EncryptionInfo, VerificationLevel, VerificationState,
+};
 use matrix_sdk_test::{async_test, sync_timeline_event, ALICE};
 use ruma::{
+    event_id,
     events::room::message::{MessageType, RedactedRoomMessageEventContent},
     server_name, EventId,
 };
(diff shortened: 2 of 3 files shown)


crates/matrix-sdk-ui/src/timeline/event_handler.rs
```diff
@@ -542,8 +542,17 @@ impl<'a, 'o> TimelineEventHandler<'a, 'o> {
...
+                Flow::Local { .. } => None,
+                Flow::Remote { encryption_info, .. } => encryption_info.clone(),
+            };
+
             trace!("Applying edit");
-            Some(event_item.with_content(new_content, edit_json))
+            Some(
+                event_item
+                    .with_content(new_content, edit_json)
+                    .with_encryption_info(encryption_info),
```
bnjbvr, line 554: Do I understand correctly that this is the fix? If so, then we should rather:

1. extract the previous `kind`
2. replace the encryption info in there
3. use `with_kind` here to replace with the updated kind

crates/matrix-sdk-ui/src/timeline/event_item/mod.rs
```diff
@@ -67,6 +67,8 @@ pub struct EventTimelineItem {
     pub(super) timestamp: MilliSecondsSinceUnixEpoch,
     /// The content of the event.
     pub(super) content: TimelineItemContent,
+    // The event's encryption information, if any
+    pub(super) encryption_info: Option<EncryptionInfo>,
```
bnjbvr, line 71: Adding this new field is wrong:

- it makes the `encryption_info` stored in the `RemoteEventTimelineItem` redundant, since it's now in two places
- this field only makes sense in the context of remote events, so it should remain in the `RemoteEventTimelineItem`, and should not be added here.

As a matter of fact, I think that most changes in this PR (except for the test, and the other place I've commented) could/should be reverted, unless I missed other calls to `with_encryption_info`.

crates/matrix-sdk-ui/src/timeline/tests/edit.rs
```diff
@@ -163,3 +168,108 @@ async fn test_aggregated_sanitized() {
...
+                        "msgtype": "m.text"
+                    },
+                    "event_id": EventId::new(server_name!("dummy.server")),
+                    "origin_server_ts": timeline.event_builder.next_server_ts(),
+                    "sender": *ALICE,
+                    "type": "m.room.message",
+                }
+            }
+        }
+    });
```
bnjbvr, line 255: Any chance you can use the fancy `EventFactory` instead? There's a team-wide effort to get rid of `sync_timeline_event` so as to get to more readable code (https://github.com/matrix-org/matrix-rust-sdk/issues/3716). Looks like the way to create edits I've added in #3854 would be sufficient to create the edit event.

crates/matrix-sdk-ui/src/timeline/tests/edit.rs
```diff
@@ -163,3 +168,108 @@ async fn test_aggregated_sanitized() {
...
+            "formatted_body": "<strong>original</strong> message",
+            "format": "org.matrix.custom.html",
+            "body": "**original** message",
+            "msgtype": "m.text"
+        },
+        "event_id": &original_event_id,
+        "origin_server_ts": timeline.event_builder.next_server_ts(),
+        "sender": *ALICE,
+        "type": "m.room.message",
+    });
```
bnjbvr, line 188: We should use the `EventFactory` here too.

bnjbvr (changes requested):

crates/matrix-sdk-ui/src/timeline/event_handler.rs
stefanceriu: Well.. it seems I made a right mess out of all of this: I completely missunderstood what the `Kind` does. I thought it's supposed to hold the *original* event data with mutations coming in as changes to locally stored properties e.g. content. 

I went and reimplemented everything. Please have another look, hope it makes more sense now 🙏

crates/matrix-sdk-ui/src/timeline/tests/edit.rs
stefanceriu: Whoa, that's sooo much nicer in comparison. I can go ahead and rebase this on top of your 3845 and switch to it if you agree.

bnjbvr (approved):
Sweet, thanks!

PR #3853, opened 2024-08-19
 feat(sdk): Sliding sync has a timeout if all lists require a timeout

Hywan:
This patch updates when sliding sync requests have a `timeout`.

Prior to this patch, all requests had a `timeout` query, set to the
`poll_timeout` duration value. However it means: if there is no data
to return, wait `timeout` milliseconds for new data before returning.
This definition is correct. Problem: if the current range of a list
has no data, the server will wait! It means that, in a situation where
there is no update at all, but the client is fetching all rooms batch by
batch, it will wait `poll_timeout` for each batch!

The behaviour described above is absolutely correct. Some server
implementations are less strict though, and we didn't realise our code
was doing that, because the server had some optimisations to ignore the
timeout if the range wasn't covering all the rooms. Nonetheless, a new
server implementation (namely Synapse) is strict, and it confirms we
have a bug here.

This patch then configures a `timeout` if all lists require a `timeout`,
otherwise there is no `timeout`, which is equivalent to `timeout=0`.

---

* Fixes https://github.com/element-hq/element-x-ios/issues/3165#issuecomment-2295700461

### DIFF
--- a/crates/matrix-sdk/src/sliding_sync/list/mod.rs
+++ b/crates/matrix-sdk/src/sliding_sync/list/mod.rs
@@ -1032,7 +1032,7 @@ mod tests {
 
     #[async_test]
     #[allow(clippy::await_holding_lock)]
-    async fn test_sliding_sync_inner_update_maximum_number_of_rooms() {
+    async fn test_inner_update_maximum_number_of_rooms() {
         let (sender, _receiver) = channel(1);
 
         let mut list = SlidingSyncList::builder("foo")
@@ -1085,7 +1085,7 @@ mod tests {
     }
 
     #[test]
-    fn test_sliding_sync_state_serialization() {
+    fn test_sliding_sync_list_loading_state_serialization() {
         assert_json_roundtrip!(from SlidingSyncListLoadingState: SlidingSyncListLoadingState::NotLoaded => json!("NotLoaded"));
         assert_json_roundtrip!(from SlidingSyncListLoadingState: SlidingSyncListLoadingState::Preloaded => json!("Preloaded"));
         assert_json_roundtrip!(from SlidingSyncListLoadingState: SlidingSyncListLoadingState::PartiallyLoaded => json!("PartiallyLoaded"));
--- a/crates/matrix-sdk/src/sliding_sync/list/mod.rs
+++ b/crates/matrix-sdk/src/sliding_sync/list/mod.rs
@@ -94,6 +94,23 @@ impl SlidingSyncList {
         self.inner.state.read().unwrap().clone()
     }
 
+    /// Check whether this list requires a [`http::Request::timeout`] value.
+    ///
+    /// A list requires a `timeout` query if and only if we want the server to
+    /// wait on new updates, i.e. to do a long-polling. If the list has a
+    /// selective sync mode ([`SlidingSyncMode::Selective`]), we expect the
+    /// server to always wait for new updates as the list ranges are always
+    /// the same. Otherwise, if the list is fully loaded, it means the list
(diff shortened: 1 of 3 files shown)


crates/matrix-sdk/src/sliding_sync/mod.rs
```diff
@@ -415,13 +415,18 @@ impl SlidingSync {
     ) -> Result<(http::Request, RequestConfig, OwnedMutexGuard<SlidingSyncPositionMarkers>)> {
         // Collect requests for lists.
         let mut requests_lists = BTreeMap::new();
+        let all_lists_are_fully_loaded;
```
bnjbvr: Having auxiliary verbs in variable names always feels super weird. How about `fully_loaded_all_lists`?

crates/matrix-sdk/src/sliding_sync/mod.rs
```diff
@@ -466,10 +471,13 @@ impl SlidingSync {
 
         Span::current().record("pos", &pos);
 
+        // Configure the timeout.
+        let timeout = if all_lists_are_fully_loaded { Some(self.inner.poll_timeout) } else { None };
```
bnjbvr: Can you expand on that comment, to explain why we don't set the `poll_timeout` sometimes? (basically repeating the commit's message in fewer words, ideally)

crates/matrix-sdk/src/sliding_sync/list/mod.rs
```diff
@@ -94,6 +94,12 @@ impl SlidingSyncList {
         self.inner.state.read().unwrap().clone()
     }
 
+    /// Check whether this list requires a [`http::Request::timeout`] value.
```
bnjbvr, line 97: Maybe expand on that comment too, or link to the other comment where we explain why we set the timeout value or not?

crates/matrix-sdk/src/sliding_sync/mod.rs
```diff
@@ -415,18 +415,18 @@ impl SlidingSync {
     ) -> Result<(http::Request, RequestConfig, OwnedMutexGuard<SlidingSyncPositionMarkers>)> {
         // Collect requests for lists.
         let mut requests_lists = BTreeMap::new();
-        let all_lists_are_fully_loaded;
+        let set_timeout;
 
         {
             let lists = self.inner.lists.read().await;
-            let mut fully_loaded = true; // Start at `true` in case there is zero list.
+            let mut require_timeout = true; // Start at `true` in case there is zero list.
```
bnjbvr, line 422: Why using a temporary variable, btw? Now that `set_timeout` is a short variable name, we could override it in the loop directly, avoiding two lines of code and a cognitive overhead of one (1) temporary binding :grin:

bnjbvr (approved):
I'd have wished the last 3 commits would be squashed into 59324a63bdb8b48514c0f35f50a30d7ffe9a9814, since they all update the logic introduced in that commit, and having them as separate commits makes the review somewhat harder. LGTM otherwise, assuming my comments will be addressed.

crates/matrix-sdk/src/sliding_sync/list/mod.rs
Hywan: Yup, indeed!

crates/matrix-sdk/src/sliding_sync/mod.rs
Hywan: Yes, you're right. I didn't want to have the outer `set_timeout` to be mutable, but I've refactored this a little bit. Thanks for the notice.

PR #3854, opened 2024-08-19
test: get rid of `TestTimeline::handle_live_message_event` in favor of `handle_live_event`

bnjbvr:
By using the `EventFactory` a bit more.

Part of #3716.


andybalaam (approved):
Looks good, thanks!

bnjbvr:
Thanks! I've noticed a few opportunities to avoid creating a new `EventFactory` and reuse that present in the `TestTimeline` (I introduced it along the way of this patch), so I'll update the tests accordingly (this small change doesn't, IMO, require a new round of review, since it's mostly trivial).

PR #3856, opened 2024-08-19
Add variants to SenderData to represent different verification states

uhoreg:
as [previously discussed](https://github.com/matrix-org/matrix-rust-sdk/pull/3701/files#r1721932382)

Signed-off-by:

### DIFF
--- a/crates/matrix-sdk-crypto/src/olm/group_sessions/sender_data.rs
+++ b/crates/matrix-sdk-crypto/src/olm/group_sessions/sender_data.rs
@@ -166,13 +166,6 @@ impl SenderData {
         Self::UnknownDevice { legacy_session: true, owner_check_failed: false }
     }
 
-    /// Returns true if this is SenderKnown and `master_key_verified` is true.
-    pub(crate) fn is_known(&self) -> bool {
-        matches!(self, SenderData::SenderUnverifiedButPreviouslyVerified { .. })
-            || matches!(self, SenderData::SenderUnverified { .. })
-            || matches!(self, SenderData::SenderVerified { .. })
-    }
-
     /// Returns `Greater` if this `SenderData` represents a greater level of
     /// trust than the supplied one, `Equal` if they have the same level, and
     /// `Less` if the supplied one has a greater level of trust.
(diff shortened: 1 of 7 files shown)


uhoreg:
One test is known to fail because I need to work on migrating data. And I probably need to write a changelog. But otherwise this PR should be reviewable, and I wanted to get it out so that I'm not holding up Andy's work.

crates/matrix-sdk-crypto/src/olm/group_sessions/sender_data.rs
```diff
@@ -20,11 +20,30 @@ use vodozemac::Ed25519PublicKey;
 
 use crate::types::DeviceKeys;
 
+/// Information about the sender of a megolm session where we know their
```
andybalaam, line 23: their->the?

crates/matrix-sdk-crypto/src/olm/group_sessions/sender_data.rs
```diff
@@ -55,26 +74,21 @@ pub enum SenderData {
         legacy_session: bool,
     },
 
+    /// We have found proof that this user, with this cross-signing key, send
```
andybalaam, line 77: send->sent

crates/matrix-sdk-crypto/src/olm/group_sessions/sender_data.rs
```diff
@@ -128,7 +167,9 @@ impl SenderData {
 
     /// Returns true if this is SenderKnown and `master_key_verified` is true.
```
andybalaam, line 168: It looks like this comment was wrong before, but definitely needs updating now.

andybalaam (approved):
Looks good to me. My only question is whether the already-incorrect doc comment on `is_known` was actually describing the behaviour some other code expected.

We also need to think about migration, as you mention.

crates/matrix-sdk-crypto/src/olm/group_sessions/sender_data.rs
uhoreg: Looking at this function, it looks like it's only used in OlmMachine to determine whether it should recalculate the sender_data.  So I definitely don't want it to return `true` if it's `SenderUnverifiedButPreviouslyVerified`.  But I'm not sure that it should return `true` if it's `SenderUnverified` either, although that would match the previous behaviour, since `sender_data_to_verification_state` returns different values if the sender is verified or not.  What do you think?  What was your rationale for having it return true when the sender was unverified?

andybalaam: I'm happy for you to merge this as-is or with one of the changes I mention above. If you want to change the behaviour I'll want to discuss it further.

uhoreg: > Maybe this function should be moved inside `OlmMachine` and be named `should_recalculate`?

Done.  It's ended up adding a bunch of negations, which I'm not really a fan of, but to avoid the negations, I'd either need to rename to `should_not_recalculate`, or switch the logic in the place where it's called, both of which seem at least as bad.

If you're OK with the way it is, can you hit the merge button?  I don't have permissions.

andybalaam:
I wanted to make some very minor changes, so closing in favour of https://github.com/matrix-org/matrix-rust-sdk/pull/3877

PR #3858, opened 2024-08-20
base: Move media cache to new `EventCacheStore` trait

### LINKED ISSUE #1810: Split media caching from the StateStore trait
The media cache should be disposable easily to free space on a device. Some OSes (usually on mobile devices) offer to empty the app cache to free space. Even on Linux, the convention is to store persistent data and cached data in different locations. This is currently not possible because the data is mixed with the rest of the state store. 

The default implementation could do nothing, like the memory store currently. It should also be possible to provide a media cache based solely on files in a directory. The cache should only handle encoding/decoding those files. I don't know if it would be enough for iOS/Android, or if they need to use the OS native methods to write files. It could also be possible to store the files unencrypted in the media repositories of the device, if the user wants to, like some apps do.

Maybe it also makes sense to have a SQLite implementation? The [docs](https://sqlite.org/fasterthanfs.html) suggest that reads and writes are more efficient than directly with the FS, especially for files up to 100KB.

The new trait could be improved later to limit the size the cache or clean up unused files.

zecakeh:
Allows to save media in a different path than the state store.

This adds a "last_access" field to the SQLite implementation, to prepare for future work on a media retention policy.

This removes the IndexedDB media cache implementation, because as far as I know it is currently unused, and I have no idea how to implement efficiently the planned media retention policy with a key-value store.

Closes #1810.


jmartinesp (commented):
The code LGTM in general, but since I'm not familiar with this part of the SDK maybe it would be better to have an extra pair of eyes on this? @bnjbvr , @Hywan maybe?

bnjbvr:
Thanks, I'll take over the review.

crates/matrix-sdk-base/src/event_cache_store/integration_tests.rs
```diff
@@ -0,0 +1,193 @@
...
+///
+///     event_cache_store_integration_tests!();
+/// }
+/// ```
+#[allow(unused_macros, unused_extern_crates)]
+#[macro_export]
+macro_rules! event_cache_store_integration_tests {
+    () => {
+        mod event_cache_store_integration_tests {
+            $crate::event_cache_store_integration_tests!(@inner);
```
bnjbvr, line 176: Does having an `@inner` macro buy us anything, or could we inline the below macro-expanded block here instead? (and have a single rule for this macro)

crates/matrix-sdk-base/src/store/integration_tests.rs
```diff
@@ -1611,18 +1503,6 @@ impl StateStoreIntegrationTests for DynStateStore {
 #[allow(unused_macros, unused_extern_crates)]
 #[macro_export]
 macro_rules! statestore_integration_tests {
-    (with_media_tests) => {
```
bnjbvr, line 1614: Same here, since there's no more macro rule for `with_media_tests`, we can probably inline the `(@inner)` expanded block into the `()` rule.

crates/matrix-sdk-sqlite/src/event_cache_store.rs
```diff
@@ -0,0 +1,385 @@
...
+
+#[async_trait]
+impl EventCacheStore for SqliteEventCacheStore {
+    type Error = Error;
+
+    async fn add_media_content(&self, request: &MediaRequest, content: Vec<u8>) -> Result<()> {
+        let uri = self.encode_key(keys::MEDIA, request.source.unique_key());
+        let format = self.encode_key(keys::MEDIA, request.format.unique_key());
+        let data = self.encode_value(content)?;
+        self.acquire().await?.set_media(uri, format, data).await
```
bnjbvr, line 225: Out of curiosity, what's the reasoning behind having multiple layers to perform SQL queries? (i.e. having `set_media` somewhere else, and not inlined here)

In my recent works on the send queue, I've put all the code at the same layer, and it feels simpler to understand, and with fewer indirections.

crates/matrix-sdk-sqlite/src/event_cache_store.rs
```diff
@@ -0,0 +1,385 @@
...
+        sqlite_db
+            .prepare("SELECT data FROM media ORDER BY last_access DESC", |mut stmt| {
+                stmt.query(())?.mapped(|row| row.get(0)).collect()
+            })
+            .await
+            .expect("querying media cache content by last access failed")
+    }
+
+    #[async_test]
+    async fn test_last_access() {
```
bnjbvr, line 294: Maybe we can introduce this feature in another PR, because I'd really like us to have it for *all* the `EventCacheStore` implementations, if we decide to go further down that road, and not only for a single backend, otherwise it may show at the use sites of the (abstract) event cache store.

Also I think for simplifying testing purposes and consistency of the backends, the timestamps should be provided by the caller; this will make it trivial to emulate time has spent without relying on `tokio::time` in tests, and will make it possible to implement this feature on the memory backend trivially too, I think.

crates/matrix-sdk-sqlite/src/event_cache_store.rs
```diff
@@ -0,0 +1,385 @@
...
+    static TMP_DIR: Lazy<TempDir> = Lazy::new(|| tempdir().unwrap());
+    static NUM: AtomicU32 = AtomicU32::new(0);
+
+    async fn get_event_cache_store() -> Result<SqliteEventCacheStore, EventCacheStoreError> {
+        let name = NUM.fetch_add(1, SeqCst).to_string();
+        let tmpdir_path = TMP_DIR.path().join(name);
+
+        tracing::info!("using event cache store @ {}", tmpdir_path.to_str().unwrap());
+
+        Ok(SqliteEventCacheStore::open(tmpdir_path.to_str().unwrap(), None).await.unwrap())
```
bnjbvr, line 276: It seems sqlite supports in-memory backends too, and our code half supports it (the path is optional); is this something we could use in testing contexts, instead of having to create temporary directories and counting the number of temporaries we've created?

crates/matrix-sdk-sqlite/src/event_cache_store.rs
```diff
@@ -0,0 +1,385 @@
...
+
+        assert_eq!(contents.len(), 2, "media cache contents length is wrong");
+        assert_eq!(contents[0], thumbnail_content, "thumbnail is not last access");
+        assert_eq!(contents[1], content, "file is not second-to-last access");
+
+        // Since the precision of the timestamp is in seconds, wait so the timestamps
+        // differ.
+        tokio::time::sleep(Duration::from_secs(3)).await;
+
+        // Access the file so it's last access is more recent.
```
bnjbvr, line 338:
```suggestion
        // Access the file so its last access is more recent.
```

crates/matrix-sdk-sqlite/migrations/event_cache_store/001_init.sql
```diff
@@ -0,0 +1,14 @@
+-- basic kv data like the database version and store cipher
```
bnjbvr, line 1:
```suggestion
-- basic kv metadata like the database version and store cipher
```

bnjbvr (changes requested):
Thanks! Would you be ok implementing the last access feature as a separate PR? I don't think it should be implemented only for the sqlite backend, and my proposal below should make it possible to implement for other backends too by tweaking the APIs.

crates/matrix-sdk-sqlite/src/event_cache_store.rs
zecakeh: I am not sure anymore. I mostly copy-pasted here what was in the `StateStore` implementation so we can definitely reduce the number of layers.

crates/matrix-sdk-base/src/event_cache_store/integration_tests.rs
zecakeh: I don't think so, I'll remove it.

crates/matrix-sdk-sqlite/src/event_cache_store.rs
zecakeh: It is planned to be for all the implementations in the next PR. The reasons I introduced it like that for the SQLite store is that:

1. This is a brand new database and I know we'll need this column so it seemed like a waste to init the table without this only to have to write a migration in the next PR.
2. Even if we only add the column without more logic, if a client uses that commit before the next PR is merged, the data will be wrong unless we decide to write a migration to clear the database.

So this is basically a placeholder implementation while we wait for the next PR.

Of course I didn't do this for the memory store because it's not persistent so we can change anything without migration.

crates/matrix-sdk-sqlite/src/event_cache_store.rs
zecakeh: Where do you see the path as optional in our code?

We could use in-memory stores, but currently we rely heavily on `deadpool-sqlite` that requires a file. It's fine if we want to use in-memory store but we have to rewrite this to be able to interact with both `deadpool-sqlite` and `rusqlite` directly. In that case the layers we currently have (or something similar) might make more sense.

crates/matrix-sdk-sqlite/src/event_cache_store.rs
zecakeh: Fixed

crates/matrix-sdk-sqlite/migrations/event_cache_store/001_init.sql
zecakeh: Changed

crates/matrix-sdk-base/src/event_cache_store/integration_tests.rs
zecakeh: Removed

crates/matrix-sdk-sqlite/src/event_cache_store.rs
bnjbvr: Good, we can do this in a follow-up PR!

crates/matrix-sdk-sqlite/src/event_cache_store.rs
bnjbvr: Sounds good!

crates/matrix-sdk-sqlite/src/event_cache_store.rs
bnjbvr: > Where do you see the path as optional in our code?

`path` in the `SqliteEventCacheStore` impl; but it's probably a copy-pasto from the state store, which… never fills this path, and only uses it for debug purposes :facepalm:

bnjbvr (approved):
Thank you!

PR #3860, opened 2024-08-20
crypto: add a note about libolm only being used in tests

richvdh:
It seems that the fact that matrix-rust-sdk contains `olm-rs` in its `Cargo.lock` sparked panic, so let's attempt to fend off future concerns by adding a comment.


andybalaam (approved):
Looks good, thanks!

PR #3861, opened 2024-08-20
Add a tip about using RustRover

richvdh:
... because it confused me.

Ref: https://youtrack.jetbrains.com/issue/RUST-15472


andybalaam (approved):
Great, thanks!

PR #3863, opened 2024-08-20
crypto: In sqlite, use the SQL column value for backed_up everywhere

andybalaam:
Most times we pulled an InboundGroupSession from the sqlite DB, we were overriding whatever value for `backed_up` was stored inside the pickled value, and using the value stored in the SQL column.
    
But when we pulled a single InboundGroupSession from the DB by ID, we did not override it.
    
I am fairly sure this was an accidental oversight, so this change corrects it, and unifies the code with other places we create these objects.

Part of https://github.com/matrix-org/matrix-rust-sdk/issues/3544
Applies on top of https://github.com/matrix-org/matrix-rust-sdk/pull/3862

### DIFF
--- a/crates/matrix-sdk-sqlite/src/crypto_store.rs
+++ b/crates/matrix-sdk-sqlite/src/crypto_store.rs
@@ -454,12 +454,12 @@ trait SqliteObjectCryptoStoreExt: SqliteObjectExt {
     async fn get_inbound_group_session(
         &self,
         session_id: Key,
-    ) -> Result<Option<(Vec<u8>, Vec<u8>)>> {
+    ) -> Result<Option<(Vec<u8>, Vec<u8>, bool)>> {
         Ok(self
             .query_row(
-                "SELECT room_id, data FROM inbound_group_session WHERE session_id = ?",
+                "SELECT room_id, data, backed_up FROM inbound_group_session WHERE session_id = ?",
                 (session_id,),
-                |row| Ok((row.get(0)?, row.get(1)?)),
+                |row| Ok((row.get(0)?, row.get(1)?, row.get(2)?)),
             )
             .await
             .optional()?)
@@ -921,7 +921,7 @@ impl CryptoStore for SqliteCryptoStore {
         session_id: &str,
     ) -> Result<Option<InboundGroupSession>> {
         let session_id = self.encode_key("inbound_group_session", session_id);
-        let Some((room_id_from_db, value)) =
+        let Some((room_id_from_db, value, backed_up)) =
             self.acquire().await?.get_inbound_group_session(session_id).await?
         else {
             return Ok(None);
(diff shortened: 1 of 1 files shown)


richvdh:
Hrm. I support the principle of being consistent, but (other than the comment asserting as much) why do we consider the sql column to be the primary source of truth? 

* It's not what the indexeddb impl does
* It requires additional work to SELECT the column and overwrite the value we have in the pickle
* I think the column was added later, so there is probably some migration code that reads from the pickle and populates the column. At the very least, that makes the runtime code inconsistent with the migration code.

andybalaam:
> Hrm. I support the principle of being consistent, but (other than the comment asserting as much) why do we consider the sql column to be the primary source of truth?

When we mark the sessions as backed up, [we only update the SQL column](https://github.com/matrix-org/matrix-rust-sdk/blob/main/crates/matrix-sdk-sqlite/src/crypto_store.rs#L502), not the pickled value.

>     * It's not what the indexeddb impl does

Yeah, it [modifies the serialized values](https://github.com/matrix-org/matrix-rust-sdk/blob/main/crates/matrix-sdk-indexeddb/src/crypto_store/mod.rs#L1004).

>     * It requires additional work to SELECT the column and overwrite the value we have in the pickle

A very tiny amount of work, yes.

>     * I think the column was added later, so there is probably some migration code that reads from the pickle and populates the column. At the very least, that makes the runtime code inconsistent with the migration code.

This is a one-time operation though, right? Then as soon as the column exists, it is the source of truth.

richvdh:
(in other words, I'd change 

> When we mark the sessions as backed up, [we only update the SQL column](https://github.com/matrix-org/matrix-rust-sdk/blob/main/crates/matrix-sdk-sqlite/src/crypto_store.rs#L502), not the pickled value.

Right. This is the clincher. Sorry, I'd forgotten that.

crates/matrix-sdk-sqlite/src/crypto_store.rs
```diff
@@ -151,16 +151,18 @@ impl SqliteCryptoStore {
...
-        value: &[u8],
+        value: Vec<u8>,
         backed_up: bool,
-    ) -> Result<PickledInboundGroupSession> {
-        let mut pickle: PickledInboundGroupSession = self.deserialize_value(value)?;
+    ) -> Result<InboundGroupSession> {
+        let mut pickle: PickledInboundGroupSession = self.deserialize_value(&value)?;
+
         // backed_up SQL column is source of truth, backed_up field in pickle
         // needed for other stores though
```
richvdh, line 162: It would be nice if this comment mentioned that the pickle is not updated when the key is backed up, which is *why* the column is the source of truth.

richvdh (approved):

crates/matrix-sdk-sqlite/src/crypto_store.rs
andybalaam: Done in [dec768c](https://github.com/matrix-org/matrix-rust-sdk/pull/3863/commits/dec768c4fd7e44a9030226eb7fb47e3a67572e05)

PR #3864, opened 2024-08-20
Expose `room_key_recipient_strategy` through higher-level SDK crates

richvdh:
https://github.com/matrix-org/matrix-rust-sdk/pull/3810 added a new option to the `CollectStrategy` used when collecting recipients for encrypted room keys. We need to expose that to applications via the FFI bindings.

Part of https://github.com/element-hq/element-internal/issues/614


andybalaam (approved):
LGTM

PR #3865, opened 2024-08-20
timeline: only use the send queue mechanism to send reactions

bnjbvr:
Independently of #3749, we can, as of today, start using the send queue only to send reactions, instead of having the logic of sending reactions in the timeline itself

There was a mechanism to queue adding/removing a reaction, with the possibility to get observable updates for each add/remove that occurred. In some contexts, if you quickly tapped a reaction button on ElementX apps, you could observe it: the reaction would be added, then removed, then added back, etc. until it stabilized.

I don't see the user value for this: if I quickly tap the reaction button, I'd expect the sending of the reaction to be aborted, if it can, instead of delayed into a redaction event that's subsequently sent. The sending queue allows doing this for free, so making use of that.

I've also removed tests which value was also unclear or suspect.

### DIFF
--- a/crates/matrix-sdk-ui/src/timeline/event_item/mod.rs
+++ b/crates/matrix-sdk-ui/src/timeline/event_item/mod.rs
@@ -12,7 +12,10 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use std::sync::Arc;
+use std::{
+    ops::{Deref, DerefMut},
+    sync::Arc,
+};
 
 use as_variant::as_variant;
 use indexmap::IndexMap;
@@ -154,7 +157,7 @@ impl EventTimelineItem {
 
         // We don't currently bundle any reactions with the main event. This could
         // conceivably be wanted in the message preview in future.
-        let reactions = IndexMap::new();
+        let reactions = ReactionsByKeyBySender::default();
 
         // The message preview probably never needs read receipts.
         let read_receipts = IndexMap::new();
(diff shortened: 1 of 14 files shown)


crates/matrix-sdk-ui/src/timeline/event_item/mod.rs
```diff
@@ -609,19 +612,69 @@ pub enum EventItemOrigin {
...
 
+/// What's the status of a reaction?
+#[derive(Clone, Debug)]
+pub enum ReactionStatus {
+    /// It's a local reaction to a remote event.
+    ///
+    /// The handle is missing only in testing contexts.
+    LocalToRemote(Option<SendHandle>),
+    /// It's a remote reaction to a remote event.
+    Remote(OwnedEventId),
```
Hywan, line 623: I would rename it `RemoteToRemote` to get the symmetry with `LocalToRemote`, but it's a very personal taste.

crates/matrix-sdk-ui/src/timeline/inner/mod.rs
```diff
@@ -739,7 +712,32 @@ impl<P: RoomDataProvider> TimelineInner<P> {
         });
 
         let Some((idx, item)) = result else {
-            // Event isn't found at all.
+            // Event isn't found as a proper item. Try to find it as a reaction?
```
Hywan, line 715: Eh look, a new feature.

crates/matrix-sdk-ui/src/timeline/inner/mod.rs
```diff
@@ -844,11 +780,42 @@ impl<P: RoomDataProvider> TimelineInner<P> {
...
+        }
+
+        // Look if this was a local reaction echo.
+        if let Some(full_key) =
+            state.meta.reactions.map.remove(&TimelineEventItemId::TransactionId(txn_id.to_owned()))
+        {
+            let item = match &full_key.item {
+                TimelineEventItemId::TransactionId(_) => {
+                    // TODO(bnjbvr): reactions on local echoes
+                    unimplemented!("no reactions on local echoes yet")
```
Hywan, line 793: Prefer a `error!` instead of a panic?

crates/matrix-sdk-ui/src/timeline/traits.rs
```diff
@@ -90,6 +91,17 @@ pub(super) trait RoomDataProvider:
...
+    /// Send an event to the fake room.
+    async fn send(&self, content: AnyMessageLikeEventContent) -> Result<(), super::Error>;
+
+    /// Redact an event in the fake room.
+    async fn redact(
+        &self,
+        event_id: &EventId,
+        reason: Option<&str>,
+        transaction_id: Option<OwnedTransactionId>,
+    ) -> Result<(), super::Error>;
```
Hywan, line 104: Why “fake room”?

Could this change be part of a single commit?

Hywan (approved):
I like the way it goes, and it's clearly an improvement. Less code, more re-used code, well done!

I wonder if the patch could be split into multiple smaller ones though, it's not easy to review, but generally the changes look good! I'm confident with the tests.

crates/matrix-sdk-ui/src/timeline/inner/mod.rs
bnjbvr: This code path is unreachable, and transient until we get https://github.com/matrix-org/matrix-rust-sdk/pull/3749, but I'll put an `error!` just in case.

crates/matrix-sdk-ui/src/timeline/traits.rs
bnjbvr: Oops, `fake room` is from a previous version of the code where this was on the `TestTimeline` impl. Will see how hard it is to isolate it into its own commit first :pray:

PR #3867, opened 2024-08-20
MatrixRTC: fix different devices from the same user overwriting the room info state event.

toger5:
With the new event format we have one event per device (not one event per user). Consequently it does not make a lot of sense to store the state events in the room info using the user id as the keys for the Map of state events.

Instead a tuple of userId and deviceId is used.

Signed-off-by:


toger5:
superseded by: https://github.com/matrix-org/matrix-rust-sdk/pull/3972

PR #3868, opened 2024-08-21
ffi: add missing info for `MessageLikeEventContent::RoomRedaction`

### LINKED ISSUE #3859: Add members to `MessageLikeEventContent.RoomRedaction`
In the context of https://github.com/element-hq/element-x-android/issues/3310, the application needs to know which Event have been redacted to be able to cleanup any displayed notification.

Adding the `roomId` and the `eventId` as fields of `MessageLikeEventContent.RoomRedaction` should be enough for the application to be able to remove the notification for the redacted event.

Happy to test a work in progress branch using EXA.

jmartinesp:
This was missing and was needed to be able to properly use the timeline item for custom logic in the clients, such as removing notifications for redacted events.

Closes #3859.

Signed-off-by:

### DIFF
--- a/bindings/matrix-sdk-ffi/src/event.rs
+++ b/bindings/matrix-sdk-ffi/src/event.rs
@@ -1,7 +1,8 @@
 use anyhow::{bail, Context};
 use ruma::events::{
-    room::message::Relation, AnySyncMessageLikeEvent, AnySyncStateEvent, AnySyncTimelineEvent,
-    AnyTimelineEvent, MessageLikeEventContent as RumaMessageLikeEventContent, RedactContent,
+    room::{message::Relation, redaction::SyncRoomRedactionEvent},
+    AnySyncMessageLikeEvent, AnySyncStateEvent, AnySyncTimelineEvent, AnyTimelineEvent,
+    MessageLikeEventContent as RumaMessageLikeEventContent, RedactContent,
     RedactedStateEventContent, StaticStateEventContent, SyncMessageLikeEvent, SyncStateEvent,
 };
 
@@ -135,7 +136,7 @@ pub enum MessageLikeEventContent {
     ReactionContent { related_event_id: String },
     RoomEncrypted,
     RoomMessage { message_type: MessageType, in_reply_to_event_id: Option<String> },
-    RoomRedaction,
+    RoomRedaction { redacted_event_id: Option<String>, reason: Option<String> },
     Sticker,
 }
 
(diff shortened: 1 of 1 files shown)


bindings/matrix-sdk-ffi/src/event.rs
```diff
@@ -200,7 +200,24 @@ impl TryFrom<AnySyncMessageLikeEvent> for MessageLikeEventContent {
                     in_reply_to_event_id,
                 }
             }
-            AnySyncMessageLikeEvent::RoomRedaction(_) => MessageLikeEventContent::RoomRedaction,
+            AnySyncMessageLikeEvent::RoomRedaction(c) => {
+                let (redacted_event_id, reason) = match c {
```
jmartinesp, line 204: Maybe using `c.redacts(&RoomVersionId)` would be better, but I don't see a way to get this version id from the existing data.

bnjbvr: Ideally, we'd use this method, but we're not quite guaranteed that we have the data; maybe it's not been requested by the sync, or it's not been received yet.

So I would slightly change the implementation to start with the `content` field, if it's there (should be present for room >= v11), and then use the `redacts` field otherwise (should be present for previous room versions).

bindings/matrix-sdk-ffi/src/event.rs
```diff
@@ -200,7 +201,21 @@ impl TryFrom<AnySyncMessageLikeEvent> for MessageLikeEventContent {
...
+                    SyncRoomRedactionEvent::Original(o) => {
+                        if let Some(id) = o.redacts {
+                            (Some(id.to_string()), None)
+                        } else if let Some(id) = o.content.redacts {
+                            (Some(id.to_string()), o.content.reason)
+                        } else {
+                            (None, None)
+                        }
```
bnjbvr, line 213: You can pass the reason as `o.content.reason` all the time, it's optional but it's in the same place for all versions of the room (if I read Ruma's code correctly).

bnjbvr (changes requested):
Sweet!

bindings/matrix-sdk-ffi/src/event.rs
```diff
@@ -200,7 +201,23 @@ impl TryFrom<AnySyncMessageLikeEvent> for MessageLikeEventContent {
...
+                    SyncRoomRedactionEvent::Original(o) => {
+                        let reason = o.content.reason;
+                        let id = if o.content.redacts.is_some() {
+                            o.content.redacts
+                        } else if o.redacts.is_some() {
+                            o.redacts
+                        } else {
+                            None
+                        }.map(|id| id.to_string());
+                        (id, reason)
```
bnjbvr, line 215:
```suggestion
                        let id = if o.content.redacts.is_some() {
                            o.content.redacts
                        } else {
                            o.redacts
                        };
                        (id.map(|id| id.to_string(), o.content.reason)
```

bnjbvr (approved):
LGTM with my suggestion applied. Thanks for doing this!

Issue #3850, opened 2024-08-19
timeline: `is_room_encrypted` should be part of the `RoomProvider` trait


bnjbvr:
Otherwise, the timeline won't correctly handle a transition from an unencrypted room to an encrypted one, which will be visible in embedders as shields not appearing in this case. Not only there's a correctness issue, but also for consistency with the rest of the `RoomProvider` trait's methods, which all return dynamic data about the `Room`.

cc @pixlwave

pixlwave:
My bad, I didn't consider that a room could be upgraded to use encryption 🙈

jmartinesp:
After discussing this we agreed a a potential fix for this could be:

1. Have a `RwLock<Option<bool>> is_encrypted` in the state, that's touched by the different sources, populated be either the previous event or `room.is_encrypted().await.ok()`, then updated via state events from sync.
2. When creating timeline items, use it to get the encryption state that's passed to the timeline items.
3. If no encryption state is found, assume it's unencrypted there and not display any shields.

Issue #3848, opened 2024-08-16
Fetching an invited room through the RoomListService fails


stefanceriu:
Using the `RoomListService` to fetch a room by id, where that room is in the invited state has regresed and is now failing with 
```
RoomListError
  ▿ InitializingTimeline : 1 element
    - error : "The room\'s encryption state is unknown."
```

This was introduced in https://github.com/matrix-org/matrix-rust-sdk/commit/bae0f304f7efc3c6a34d2ca805eb2d4b27189ce9#diff-dff33932423bfedc588739ae79f87f64f063a52aada9483069fb11bc5487c3b7R161-R162

stefanceriu:
Turns out this is not necessarily a rust bug. The client asks to initialise the timeline when fetching a room and actually what's wrong over here. We'll switch away from that and not change anything on the rust side.

stefanceriu:
Reopening this as it can't be solved on the client side only. Using `full_room` on a roomListItem on which the timeline hasn't been initialized failes as per https://github.com/matrix-org/matrix-rust-sdk/blob/main/bindings/matrix-sdk-ffi/src/room_list.rs#L616

The "full_room" is used to accept or decline and invite.

PR #3869, opened 2024-08-21
ffi: expose room membership through the RoomListItem and allow invited rooms to be build differently than "full" ones

stefanceriu:
We're finding ourselves in the situation in which we can't interact with invites through normal Room APIs as `full_room`s can't be build from the RoomListItem. Full rooms require the timeline to be configured before use and the timeline can't be configured because encryption cannot be fetched for invited rooms on homeservers that have previews disabled (see #3848 and #3850)

In response we now expose the room's membership directly from the `RoomListItem` so that the final client can chose which of the 2 rooms types (invited or full) to ask for before using aforementioned APIs.

Powers https://github.com/element-hq/element-x-ios/pull/3189

### DIFF
--- a/bindings/matrix-sdk-ffi/src/room_list.rs
+++ b/bindings/matrix-sdk-ffi/src/room_list.rs
@@ -23,7 +23,7 @@ use tokio::sync::RwLock;
 
 use crate::{
     error::ClientError,
-    room::Room,
+    room::{Membership, Room},
     room_info::RoomInfo,
     timeline::{EventTimelineItem, Timeline},
     timeline_event_filter::TimelineEventTypeFilter,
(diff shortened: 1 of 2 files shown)


bindings/matrix-sdk-ffi/src/room_list.rs
```diff
@@ -603,10 +605,35 @@ impl RoomListItem {
         Ok(RoomInfo::new(self.inner.inner_room()).await?)
     }
 
+    /// The room's current membership state
```
bnjbvr, line 608:
```suggestion
    /// The room's current membership state.
```

bindings/matrix-sdk-ffi/src/room_list.rs
```diff
@@ -50,6 +50,8 @@ pub enum RoomListError {
     InitializingTimeline { error: String },
     #[error("Event cache ran into an error: {error}")]
     EventCache { error: String },
+    #[error("The requested room doesn't match the membership requirements")]
+    IncorrectRoomMembership,
```
bnjbvr, line 54: bonus points if you include two fields `expected` / `observed` that help matching the requirements, but not mandatory for merging :)

bnjbvr (approved):
LGTM, thanks!

bindings/matrix-sdk-ffi/src/room_list.rs
stefanceriu: Noice, bonus points for me it is then! 😁

PR #3870, opened 2024-08-21
sdk: add relationships cache to `EventCache`

jmartinesp:
## Changes

- Creates a separate `AllEventsCache` struct holding both the actual all events cache map and the relationship one. This new cache wrapper also holds a single `RwLock` to both inner caches, as they are independent from each other.
- When a new event is saved either from `/sync` or another HS request, it'll be saved to both the 'all events' cache and the relationship map.
- Add tests for the relationship cache.

Signed-off-by:

### DIFF
--- a/crates/matrix-sdk/src/event_cache/mod.rs
+++ b/crates/matrix-sdk/src/event_cache/mod.rs
@@ -64,7 +64,7 @@ use ruma::{
 };
 use tokio::sync::{
     broadcast::{error::RecvError, Receiver, Sender},
-    Mutex, RwLock, RwLockWriteGuard,
+    Mutex, RwLock, RwLockReadGuard, RwLockWriteGuard,
 };
 use tracing::{error, info_span, instrument, trace, warn, Instrument as _, Span};
 
@@ -324,9 +324,14 @@ impl EventCache {
 type AllEventsMap = BTreeMap<OwnedEventId, (OwnedRoomId, SyncTimelineEvent)>;
 type RelationsMap = BTreeMap<OwnedEventId, BTreeSet<OwnedEventId>>;
 
+/// Cache wrapper containing both copies of received events and lists of event
+/// ids related to them.
 #[derive(Default, Clone)]
 struct AllEventsCache {
+    /// A cache of received events mapped by their event id.
     events: AllEventsMap,
+    /// A cache of related event ids for an event id. The key is the original
+    /// event id and the value a list of event ids related to it.
     relations: RelationsMap,
 }
 
@@ -354,7 +359,6 @@ struct EventCacheInner {
     ///
     /// This is shared between the [`EventCache`] singleton and all
     /// [`RoomEventCache`] instances.
-    // If necessary, we can add a RoomId in the value later.
     all_events: Arc<RwLock<AllEventsCache>>,
 
     /// Handles to keep alive the task listening to updates.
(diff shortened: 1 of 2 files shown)


crates/matrix-sdk/src/test_utils/events.rs
```diff
@@ -238,6 +250,72 @@ impl EventFactory {
...
+
+    /// Create a poll start event given a text, the question and the possible
+    /// answers.
+    pub fn poll_start(
+        &self,
+        content: impl Into<String>,
+        poll_question: impl Into<String>,
+        answers: Vec<impl Into<String>>,
+    ) -> EventBuilder<PollStartEventContent> {
+        // PollAnswers 'constructor' is not public, so we need to deserialize them
```
bnjbvr, line 272: It seems you can convert a `Vec<PollAnswer>` `into()` a `PollAnswers`; can you do this instead of the json hack please?

crates/matrix-sdk/src/event_cache/mod.rs
```diff
@@ -338,7 +355,7 @@ struct EventCacheInner {
     /// This is shared between the [`EventCache`] singleton and all
     /// [`RoomEventCache`] instances.
     // If necessary, we can add a RoomId in the value later.
```
bnjbvr, line 357: pre-existing, but can you remove this line of the doc-comment please? I forgot.

crates/matrix-sdk/src/event_cache/mod.rs
```diff
@@ -501,17 +524,40 @@ impl RoomEventCache {
         None
     }
 
+    /// Try to find an event by id in this room, along with all relations.
+    pub async fn event_with_relations(
```
bnjbvr, line 528: So here's a design question: a related event could also have related events (e.g. an edit could be redacted). Should the function return the transitive closure of relations (i.e. all the related events, and all their related events recursively)? I think so, in which case we should likely implement some kind of fixed point algorithm to get all the related events.

If you agree with this and implement it, can you add a test exhibiting this use case, please?

crates/matrix-sdk/src/event_cache/mod.rs
```diff
@@ -712,6 +762,113 @@ impl RoomEventCacheInner {
         .await
     }
 
+    /// If the event is related to another one, it's id is added to the
```
bnjbvr, line 765:
```suggestion
    /// If the event is related to another one, its id is added to the
```

crates/matrix-sdk/src/event_cache/mod.rs
```diff
@@ -712,6 +762,113 @@ impl RoomEventCacheInner {
...
+                    }
+                    SyncRoomRedactionEvent::Redacted(ev) => {
+                        if let Some(redacted_event_id) = &ev.content.redacts {
+                            cache
+                                .relations
+                                .entry(redacted_event_id.to_owned())
+                                .or_default()
+                                .insert(ev.event_id.to_owned());
+                        }
+                    }
```
bnjbvr, line 799: Wait, do we care about redacted redaction events now?

crates/matrix-sdk/src/event_cache/mod.rs
```diff
@@ -712,6 +762,113 @@ impl RoomEventCacheInner {
...
+                                .entry(redacted_event_id.to_owned())
+                                .or_default()
+                                .insert(ev.event_id.to_owned());
+                        } else if let Some(redacted_event_id) = &ev.redacts {
+                            cache
+                                .relations
+                                .entry(redacted_event_id.to_owned())
+                                .or_default()
+                                .insert(ev.event_id.to_owned());
+                        }
```
bnjbvr, line 789: nit: let's fold these two branches:

```suggestion
                        if let Some(redacted_event_id) = ev.content.redacts.as_ref().or_else(|| ev.redacts.as_ref()) {
                            cache
                                .relations
                                .entry(redacted_event_id.to_owned())
                                .or_default()
                                .insert(ev.event_id.to_owned());
                        }
```

crates/matrix-sdk/src/event_cache/mod.rs
```diff
@@ -712,6 +762,113 @@ impl RoomEventCacheInner {
...
+                    }
+                }
+            }
+            match ev.original_content() {
```
bnjbvr, line 802: nit: in this entire match, can you collect an `Option<&EventId>` first, and then add it to the `cache.relations` mapping if it exists? It will avoid duplicating the insertion into the `cache.relations` for every single arm.

bnjbvr (changes requested):
Thanks for splitting this from the other large PR, it makes the review definitely easier!

I've written lots of nit comments (hopefully should be fast to address! don't hesitate to address many of them in a single commit), and only one major comment about functionality, so I'd like to take another look, but this looks like this is on the right track!

crates/matrix-sdk/src/test_utils/events.rs
```diff
@@ -238,6 +250,72 @@ impl EventFactory {
...
+    /// Create a replacement message for an existing one given its id.
+    pub fn replacement_msg(
+        &self,
+        new_content: impl Into<String>,
+        original_event_id: &EventId,
+    ) -> EventBuilder<RoomMessageEventContent> {
+        let msg_content = RoomMessageEventContent::text_plain(new_content.into());
+        let builder = self.event(msg_content.clone());
+        builder.edit(original_event_id, msg_content.into())
+    }
```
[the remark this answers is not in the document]
jmartinesp: Ok, although we'll have to pass the message content twice then and clone it, I think.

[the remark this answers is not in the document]
jmartinesp: Since we're already providing a `content` value when we get the `EventBuilder`, should we remove the `new_content` param from `EventBuilder::edit` instead and use this content?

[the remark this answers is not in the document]
jmartinesp: Ah, I see it's done that way so we can add the `* ` prefix separately...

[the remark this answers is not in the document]
bnjbvr: No, it's different: the parameter to `text_msg` for the edit event is the fallback text, while the new content for clients that support edits will be extracted from `new_content`.

crates/matrix-sdk/src/event_cache/mod.rs
```diff
@@ -1018,4 +1178,159 @@ mod tests {
...
+
+        let event_cache = client.event_cache();
+        event_cache.subscribe().unwrap();
+
+        client.base_client().get_or_create_room(room_id, matrix_sdk_base::RoomState::Joined);
+        let room = client.get_room(room_id).unwrap();
+
+        let (room_event_cache, _drop_handles) = room.event_cache().await.unwrap();
+
+        // Save the original event
```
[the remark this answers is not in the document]
jmartinesp: Ok, although technically that's a line comment, not a doc one 🤷 .

crates/matrix-sdk/src/test_utils/events.rs
jmartinesp:
```
error[E0277]: the trait bound `PollAnswers: From<Vec<PollAnswer>>` is not satisfied
   --> crates/matrix-sdk/src/test_utils/events.rs:267:32
    |
267 |         }).collect::<Vec<_>>().into();
    |                                ^^^^ the trait `From<Vec<PollAnswer>>` is not implemented for `PollAnswers`, which is required by `Vec<PollAnswer>: Into<_>`
    |
    = note: required for `Vec<PollAnswer>` to implement `Into<PollAnswers>`
```

crates/matrix-sdk/src/event_cache/mod.rs
jmartinesp: I think it makes sense, yes. Let's see how hard to implement this is.

jmartinesp: Just to confirm, there's no way relationships can be circular, right?

crates/matrix-sdk/src/event_cache/mod.rs
jmartinesp: Oops, no, I don't think so.

crates/matrix-sdk/src/event_cache/mod.rs
jmartinesp: This was handled in https://github.com/matrix-org/matrix-rust-sdk/pull/3870/commits/748625a8600d16d246a94958539a1602d451393c, I added a check to try to avoid circular relationships in case there could be some.

jmartinesp:
I think everything mentioned in the nits is fixed in https://github.com/matrix-org/matrix-rust-sdk/pull/3870/commits/c5f6c444db83df78abc590ac8d0818d5f7331fff, the recursive loading of related events is done in a separate commit (https://github.com/matrix-org/matrix-rust-sdk/pull/3870/commits/748625a8600d16d246a94958539a1602d451393c).

crates/matrix-sdk/src/test_utils/events.rs
jmartinesp: So it seems like this was fixed in Ruma, but our fork doesn't contain the change 🫤 .

jmartinesp: Ok, I just saw it does implement `TryFrom` in the commit hash we use 🤦 . Fixed in https://github.com/matrix-org/matrix-rust-sdk/pull/3870/commits/cc3c20ed27e08d3c0209e073d20865e25455b2fa.

crates/matrix-sdk/src/event_cache/mod.rs
```diff
@@ -501,17 +528,60 @@ impl RoomEventCache {
...
+            Some((event.clone(), relation_events))
+        } else {
+            None
+        }
+    }
+
+    /// Looks for related event ids for the passed event id, and appends them to
+    /// the `results` parameter. Then it'll recursively get the related
+    /// event ids for those too.
+    fn related_event_ids_for_event(
```
bnjbvr, line 550: (nit: can we name this `collect_related_events`?)

crates/matrix-sdk/src/event_cache/mod.rs
```diff
@@ -501,17 +528,60 @@ impl RoomEventCache {
...
+        cache: &RwLockReadGuard<'_, AllEventsCache>,
+        event_id: &EventId,
+        results: &mut Vec<SyncTimelineEvent>,
```
bnjbvr, line 553: Do you think it would be ok to have `results` be a `HashSet` or `BTreeSet` instead? (and return that from the `event_with_relations` function)

It would avoid the linear search below, and pretty sure consumers don't care about the order; after all we're returning an arbitrary order of related events here.

bnjbvr (changes requested):
Almost there! Last round :)

crates/matrix-sdk/src/event_cache/mod.rs
jmartinesp: A couple of questions about that: 

- Wouldn't that mean we'd have to implement either `Hash` or `Ord` for `SyncTimelineEvent`? And if so, what would the hash value or the ordering be based on?
- Are we sure ordering doesn't matter for the `TimelineEventHandler`? `rfind_event_by_id` is used in some places to look for related events, and that will only work if the event was already added to `self.items`, right?

bnjbvr: - You're right about `Hash` for both data structures, if we only use `HashSet` we don't need `Ord` (we'd need it for `BTreeSet` though). If it's not implemented already, let's forget about that idea.
- The "only" thing that matters with respect to ordering is that the related-to event is added before the relation events, if I understand correctly; otherwise you're right it would be problematic.

jmartinesp: We already discussed this in person, but just so we track the conclusions:
- We don't have `Hash` implemented and while we could generate it from the contents, it's probably not worth it.
- Using a `Set` could be a problem because with this code we'd also add transitively related events to it (a related event for a related event) and that may break the current code for handling timeline events since the original event for a related event must be already added to the timeline and we need the topological order to ensure that.

bnjbvr (approved):
Thank you!

PR #3871, opened 2024-08-21
fix(sdk): Clear all sliding sync room subscriptions when session expires

Hywan:
This patch clears all sliding sync room subscriptions when a session expires. Indeed, we might not want to request all room subscriptions when the session restarts. Imagine if the client has subscribed to 400 rooms and the session expires: once the session restarts, it will ask for 400 room subscriptions, which is a lot and will result in a quite slow response.

---

* Address https://github.com/matrix-org/matrix-rust-sdk/issues/3647.

### DIFF
--- a/crates/matrix-sdk/src/sliding_sync/mod.rs
+++ b/crates/matrix-sdk/src/sliding_sync/mod.rs
@@ -756,7 +756,7 @@ impl SlidingSync {
         Ok(self.inner.internal_channel_send(SlidingSyncInternalMessage::SyncLoopStop)?)
     }
 
-    /// Expire the current Sliding Sync session.
+    /// Expire the current Sliding Sync session on the client-side.
     ///
     /// Expiring a Sliding Sync session means: resetting `pos`. It also resets
     /// sticky parameters.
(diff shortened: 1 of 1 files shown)


crates/matrix-sdk/src/sliding_sync/mod.rs
```diff
@@ -783,7 +783,14 @@ impl SlidingSync {
...
         // Force invalidation of all the sticky parameters.
-        let _ = self.inner.sticky.write().unwrap().data_mut();
+        {
+            let mut sticky = self.inner.sticky.write().unwrap();
+            let data = sticky.data_mut();
+
+            // Clear all room subscriptions: we don't want to resend all room subscriptions
+            // when the session will restart.
+            data.room_subscriptions.clear();
+        }
```
bnjbvr, line 793: Can you get rid of all the temporaries, and remove the comment about forcing invalidation? Calling `data_mut()` invalidates sticky parameters anyways.

crates/matrix-sdk/src/sliding_sync/mod.rs
```diff
@@ -1151,6 +1158,65 @@ mod tests {
...
+        let room_id_2 = room_id!("!r2:bar.org");
+
+        // Subscribe to two rooms.
+        sliding_sync.subscribe_to_rooms(&[room_id_0, room_id_1], None);
+
+        {
+            let sticky = sliding_sync.inner.sticky.read().unwrap();
+            let room_subscriptions = &sticky.data().room_subscriptions;
+
+            assert!(room_subscriptions.contains_key(&room_id_0.to_owned()));
```
bnjbvr, line 1178: Is any of the `.to_owned()` necessary in that test? I thought the hashmap's key could be a `&RoomId` since it can be borrowed from an `OwnedRoomId` (the real key type).

bnjbvr (approved):
LGTM with comments addressed, thanks

crates/matrix-sdk/src/sliding_sync/mod.rs
Hywan: Good to me.

PR #3873, opened 2024-08-21
testing: add a mocks module in matrix-sdk-test to reuse mock endpoints across test suites

bnjbvr:
What it says on the tin.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3503,6 +3503,7 @@ dependencies = [
  "tokio",
  "tracing-subscriber",
  "wasm-bindgen-test",
+ "wiremock",
 ]
 
 [[package]]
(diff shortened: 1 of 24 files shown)


jmartinesp (approved):
LGTM, although there are some other mocked endpoints in [sdk-ui](https://github.com/matrix-org/matrix-rust-sdk/blob/ab12f2f7cac59bd9c6604e8cbd7ac7b6122e7c8e/crates/matrix-sdk-ui/tests/integration/main.rs) and other places, should we try extracting those too?

bnjbvr:
Thanks for the review!

> should we try extracting those too?

Yep, this is a bit tedious so I haven't done all of them yet, but we can coordinate this as a team effort.

PR #3876, opened 2024-08-22
sqlite: Cleanup

zecakeh:
See each commit description for details.

### DIFF
--- a/crates/matrix-sdk-sqlite/src/crypto_store.rs
+++ b/crates/matrix-sdk-sqlite/src/crypto_store.rs
@@ -16,7 +16,7 @@ use std::{
     borrow::Cow,
     collections::HashMap,
     fmt,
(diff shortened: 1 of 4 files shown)


bnjbvr (commented):
Thanks. I'm a bit ambivalent about the `Object` renaming: object seems to refer, in deadpool, to the "object that's being managed", and that's one of the most overloaded and least descriptive terms in computer science. I'd rather have it the other way around, and rename all of it "cursor" or "connection" — or something else, really, since I'm not clear what the fundamental meaning of this thing is.

zecakeh:
Okay, so I rewrote the first commit to use `SqliteConn` instead of `deadpool_sqlite::Object` everywhere then.

bnjbvr (approved):
Thanks!

PR #3877, opened 2024-08-22
crypto: Add variants to SenderData to represent different verification states

andybalaam:
(Slightly-modified version of https://github.com/matrix-org/matrix-rust-sdk/pull/3856/)

As [previously discussed](https://github.com/matrix-org/matrix-rust-sdk/pull/3701/files#r1721932382)

### DIFF
--- a/crates/matrix-sdk-crypto/CHANGELOG.md
+++ b/crates/matrix-sdk-crypto/CHANGELOG.md
@@ -32,15 +32,24 @@ Changes:
 
 Breaking changes:
 
+  **NOTE**: this version causes changes to the format of the serialised data in
+  the CryptoStore, meaning that, once upgraded, it will not be possible to roll
+  back applications to earlier versions without breaking user sessions.
+
+- Change the structure of the `SenderData` enum to separate variants for
+  previously-verified, unverified and verified.
+  ([#3877](https://github.com/matrix-org/matrix-rust-sdk/pull/3877))
+
+- Where `EncryptionInfo` is returned it may include the new `PreviouslyVerified`
+  variant of `VerificationLevel` to indicate that the user was previously
+  verified and is no longer verified.
+  ([#3877](https://github.com/matrix-org/matrix-rust-sdk/pull/3877))
+
 - Expose new methods `OwnUserIdentity::was_previously_verified`,
   `OwnUserIdentity::withdraw_verification`, and
   `OwnUserIdentity::has_verification_violation`, which track whether our own
   identity was previously verified.
   ([#3846](https://github.com/matrix-org/matrix-rust-sdk/pull/3846))
-
-  **NOTE**: this causes changes to the format of the serialised data in the
-  CryptoStore, meaning that, once upgraded, it will not be possible to roll
-  back applications to earlier versions without breaking user sessions.
 
 - Add a new `error_on_verified_user_problem` property to
   `CollectStrategy::DeviceBasedStrategy`, which, when set, causes
--- a/crates/matrix-sdk-crypto/src/machine/mod.rs
+++ b/crates/matrix-sdk-crypto/src/machine/mod.rs
@@ -66,8 +66,8 @@ use crate::{
     identities::{user::UserIdentities, Device, IdentityManager, UserDevices},
     olm::{
         Account, CrossSigningStatus, EncryptionSettings, IdentityKeys, InboundGroupSession,
-        OlmDecryptionInfo, PrivateCrossSigningIdentity, SenderData, SenderDataFinder, SessionType,
-        StaticAccountData,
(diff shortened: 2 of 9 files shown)


crates/matrix-sdk-crypto/src/machine/mod.rs
```diff
@@ -1465,11 +1462,32 @@ impl OlmMachine {
...
             }
+        } else {
+            session.sender_data.clone()
         };
 
         Ok(sender_data_to_verification_state(sender_data, session.has_been_imported()))
     }
 
+    /// Whether we should recalculate the Megolm sender's data, given the
+    /// current sender data. We only want to recalculate if it might increase
```
richvdh, line 1473: I feel like this is missing context. It surely can't be the case that we *always* want to recalculate senderdata -- there must be some event that prompts it? In other words, when is this method called?

crates/matrix-sdk-crypto/src/machine/mod.rs
```diff
@@ -1465,11 +1462,32 @@ impl OlmMachine {
...
+    /// trust and allow us to decrypt messages that we otherwise might
+    /// refuse to decrypt.
+    ///
+    /// We recalculate for all states except:
+    ///
+    /// - SenderUnverified: the sender is trusted enough that we will decrypt
+    ///   their messages in all cases, or
+    /// - SenderVerified: the sender is the most trusted they can be.
+    fn should_recalculate_sender_data(sender_data: &SenderData) -> bool {
```
richvdh, line 1482: I don't really get why you want to move this to `OlmMachine`. `OlmMachine` is way too big as it is, and keeping it as a method on `SenderData` (with another name, maybe) seems preferable.

richvdh (changes requested):

crates/matrix-sdk-crypto/src/machine/mod.rs
andybalaam: When it was on `SenderData` it appeared to be a general-purpose method. What we discovered as we discussed it was that the logic here is tightly coupled to when we want to recalculate sender data when we decrypt a message. It will change if we change our minds (e.g. to recalculate when we have an unverified sender in case they became verified).

So it feels like an implementation detail of `get_or_update_verification_state`. I'd be happy to move it literally inside `get_or_update_verification_state` if you like.

crates/matrix-sdk-crypto/src/machine/mod.rs
andybalaam: This is used when we decrypt a message, and its logic is tied to those circumstances. A free function or a function definition inside `get_or_update_verification_state` might be better?

crates/matrix-sdk-crypto/src/machine/mod.rs
richvdh: ok possibly this is a sign that the whole of `get_or_update_verification_state`  needs to move somewhere else.

Please either make it an inner function or give it a name and doc-comment which identify it is part of `get_or_update_verification_state`. I don't have a strong feeling which.

crates/matrix-sdk-crypto/CHANGELOG.md
```diff
@@ -32,16 +32,20 @@ Changes:
 
 Breaking changes:
 
+  **NOTE**: this version causes changes to the format of the serialised data in
+  the CryptoStore, meaning that, once upgraded, it will not be possible to roll
+  back applications to earlier versions without breaking user sessions.
+
+- Change the structure of the `SenderData` enum. The new `SenderDataReader`
+  demonstrates how to migrate from the previous form.
+  ([#3877](https://github.com/matrix-org/matrix-rust-sdk/pull/3877))
```
richvdh, line 41: We need to actually document the behaviour change, as well as the interface change. As I understand it, `OlmMachine.get_encryption_info` will now return a different `VerificationState` if a user who was previously verified changes their identity?

richvdh (changes requested):
Sorry I'm still nit-picking about the changelog.

crates/matrix-sdk-crypto/src/machine/mod.rs
andybalaam: Moved to an inner function in 3a84196573

crates/matrix-sdk-crypto/src/machine/mod.rs
andybalaam: Moved to an inner function in 3a84196573

crates/matrix-sdk-crypto/CHANGELOG.md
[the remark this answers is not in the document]
andybalaam: Fixed in 146656105d

andybalaam: Fixed in 146656105d

richvdh (approved):

PR #3878, opened 2024-08-22
pinned events: limit the max amount of concurrent requests to load pinned events

jmartinesp:
Depends on https://github.com/matrix-org/matrix-rust-sdk/pull/3840.

`RequestConfig::max_concurrent_requests` won't work for requests to a same endpoint, since it's a global value that can only be passed to `HttpClient` when created.

To replace it I added a buffered stream that can run at most `MAX_PINNED_EVENTS_CONCURRENT_REQUESTS` (10) requests concurrently.

There is also a test to make sure this max concurrency mechanism works.

Signed-off-by:

### DIFF
--- a/crates/matrix-sdk-ui/src/timeline/pinned_events_loader.rs
+++ b/crates/matrix-sdk-ui/src/timeline/pinned_events_loader.rs
@@ -12,9 +12,9 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use std::{fmt::Formatter, num::NonZeroUsize, sync::Arc};
+use std::{fmt::Formatter, sync::Arc};
 
-use futures_util::{future::join_all, FutureExt as _};
+use futures_util::{stream, FutureExt as _, StreamExt};
 use matrix_sdk::{
     config::RequestConfig, event_cache::paginator::PaginatorError, BoxFuture, Room,
     SendOutsideWasm, SyncOutsideWasm,
@@ -24,8 +24,6 @@ use ruma::{EventId, MilliSecondsSinceUnixEpoch, OwnedEventId};
 use thiserror::Error;
 use tracing::{debug, warn};
 
-const MAX_CONCURRENT_REQUESTS: usize = 10;
-
 /// Utility to load the pinned events in a room.
 pub struct PinnedEventsLoader {
     /// Backend to load pinned events.
@@ -34,12 +32,21 @@ pub struct PinnedEventsLoader {
     /// Maximum number of pinned events to load (either from network or the
     /// cache).
     max_events_to_load: usize,
+
+    /// Number of requests to load pinned events that can run concurrently. This
+    /// is used to avoid overwhelming a home server with dozens or hundreds
+    /// of concurrent requests.
+    max_concurrent_requests: usize,
 }
 
 impl PinnedEventsLoader {
     /// Creates a new `PinnedEventsLoader` instance.
-    pub fn new(room: Arc<dyn PinnedEventsRoom>, max_events_to_load: usize) -> Self {
-        Self { room, max_events_to_load }
+    pub fn new(
+        room: Arc<dyn PinnedEventsRoom>,
+        max_events_to_load: usize,
+        max_concurrent_requests: usize,
+    ) -> Self {
+        Self { room, max_events_to_load, max_concurrent_requests }
     }
 
     /// Loads the pinned events in this room, using the cache first and then
@@ -64,37 +71,33 @@ impl PinnedEventsLoader {
             return Ok(Vec::new());
         }
 
-        let request_config = Some(
-            RequestConfig::default()
-                .retry_limit(3)
-                .max_concurrent_requests(NonZeroUsize::new(MAX_CONCURRENT_REQUESTS)),
-        );
+        let request_config = Some(RequestConfig::default().retry_limit(3));
 
-        let new_events = join_all(pinned_event_ids.into_iter().map(|event_id| {
-            let provider = self.room.clone();
-            async move {
-                match provider.load_event_with_relations(&event_id, request_config).await {
-                    Ok((event, related_events)) => {
-                        let mut events = vec![event];
-                        events.extend(related_events);
-                        Some(events)
-                    }
-                    Err(err) => {
-                        warn!("error when loading pinned event: {err}");
-                        None
+        let mut loaded_events: Vec<SyncTimelineEvent> =
+            stream::iter(pinned_event_ids.into_iter().map(|event_id| {
+                let provider = self.room.clone();
+                async move {
+                    match provider.load_event_with_relations(&event_id, request_config).await {
+                        Ok((event, related_events)) => {
+                            let mut events = vec![event];
+                            events.extend(related_events);
+                            Some(events)
+                        }
+                        Err(err) => {
+                            warn!("error when loading pinned event: {err}");
+                            None
+                        }
                     }
                 }
-            }
-        }))
-        .await;
-
-        let mut loaded_events = new_events
-            .into_iter()
+            }))
+            .buffer_unordered(self.max_concurrent_requests)
             // Get only the `Some<Vec<_>>` results
-            .flatten()
+            .flat_map(stream::iter)
             // Flatten the `Vec`s into a single one containing all their items
-            .flatten()
-            .collect::<Vec<SyncTimelineEvent>>();
+            .flat_map(stream::iter)
+            .collect()
+            .await;
+
         if loaded_events.is_empty() {
             return Err(PinnedEventsLoaderError::TimelineReloadFailed);
         }
--- a/crates/matrix-sdk-ui/tests/integration/timeline/pinned_event.rs
+++ b/crates/matrix-sdk-ui/tests/integration/timeline/pinned_event.rs
@@ -12,7 +12,7 @@ use matrix_sdk::{
 use matrix_sdk_base::deserialized_responses::TimelineEvent;
 use matrix_sdk_test::{async_test, JoinedRoomBuilder, StateTestEvent, SyncResponseBuilder, BOB};
 use matrix_sdk_ui::{
-    timeline::{TimelineFocus, TimelineItemContent},
+    timeline::{RoomExt, TimelineFocus, TimelineItemContent},
     Timeline,
 };
 use ruma::{
@@ -21,7 +21,11 @@ use ruma::{
 };
 use serde_json::json;
 use stream_assert::assert_pending;
-use wiremock::MockServer;
+use tokio::time::sleep;
+use wiremock::{
+    matchers::{header, method, path_regex},
+    Mock, MockServer, ResponseTemplate,
+};
 
 use crate::{mock_event, mock_sync};
 
@@ -46,11 +50,8 @@ async fn test_new_pinned_events_are_added_on_sync() {
     let _ = test_helper.setup_sync_response(vec![(event_1, false)], Some(vec!["$1", "$2"])).await;
 
     let room = test_helper.client.get_room(&room_id).unwrap();
-    let timeline = Timeline::builder(&room)
-        .with_focus(TimelineFocus::PinnedEvents { max_events_to_load: 100 })
-        .build()
-        .await
-        .unwrap();
+    let timeline =
+        Timeline::builder(&room).with_focus(pinned_events_focus(100)).build().await.unwrap();
     test_helper.server.reset().await;
 
     assert!(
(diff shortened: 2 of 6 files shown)


crates/matrix-sdk-ui/tests/integration/timeline/pinned_event.rs
```diff
@@ -624,6 +630,78 @@ async fn test_edited_events_survive_pinned_event_ids_change() {
...
+            .with_focus(TimelineFocus::PinnedEvents { max_events_to_load: 100 });
+        async {
+            let _ = timeline_builder.build().await;
+        }
+    });
+
+    // Give it time to load events. As each request takes `request_delay`, we should
+    // have exactly `MAX_PINNED_EVENTS_CONCURRENT_REQUESTS` if the max
+    // concurrency setting is honoured.
+    sleep(Duration::from_millis(request_delay)).await;
```
andybalaam, line 699: Would it be safer to sleep for `1.5 * request_delay` just to make sure some other delay doesn't cause us to just miss it?

andybalaam (approved):
Looks good, one question.

crates/matrix-sdk-ui/tests/integration/timeline/pinned_event.rs
andybalaam: Then `0.5 * request_delay`?

jmartinesp: Ok, sounds good. I wanted to use the full delay to prove this worked as expected, but it's true it might result in flaky tests 🫤 .

crates/matrix-sdk-ui/src/timeline/pinned_events_loader.rs
```diff
@@ -26,7 +26,7 @@ use tracing::{debug, warn};
 
 use crate::timeline::event_handler::TimelineEventKind;
 
-const MAX_CONCURRENT_REQUESTS: usize = 10;
+pub const MAX_PINNED_EVENTS_CONCURRENT_REQUESTS: usize = 10;
```
Hywan, line 29: If it's a `pub` item, please add a documentation to it.

If it's a `pub` only for testing purposes, maybe adding a `#[cfg(test)]` would be nice?

Hywan (changes requested):

crates/matrix-sdk-ui/src/timeline/pinned_events_loader.rs
jmartinesp: > If it's a pub item, please add a documentation to it.

True, my bad.

> If it's a pub only for testing purposes, maybe adding a #[cfg(test)] would be nice?

Well, it just needs to be `pub` for the tests, but it needs to exist in general since it's used in the code below it. I don't know if there's some way to do that or it's better to just keep it public.

Hywan: That's one limitation of Rust where we can add a `#[cfg]` for a visiblity only. There is a crate to handle that, https://crates.io/crates/cfg-vis, but [we have removed it recently](https://github.com/matrix-org/matrix-rust-sdk/pull/3375).

You can write this as follows:

```rust
#[cfg(test)]
pub const … = …;

#[cfg(not(test))]
const … = …;
```

if you don't want to repeat the constant value, you can have a private constant just for that:

```rust
const VALUE: usize = 10;

#[cfg(test)]
pub const …: usize = VALUE;

#[cfg(not(test))]
const …: usize = VALUE;
```

of course, with better names that `…` and `VALUE` :-D.

jmartinesp: It seems like this won't work as the target test is an integration one.

crates/matrix-sdk-ui/src/timeline/pinned_events_loader.rs
```diff
@@ -66,37 +68,33 @@ impl PinnedEventsLoader {
...
-                        events.extend(related_events);
-                        Some(events)
-                    }
-                    Err(err) => {
-                        warn!("error when loading pinned event: {err}");
-                        None
+        let request_config = Some(RequestConfig::default().retry_limit(3));
+
+        let mut loaded_events: Vec<SyncTimelineEvent> =
+            stream::iter(pinned_event_ids.into_iter().map(|event_id| {
```
bnjbvr, line 74: Wonder if we could use a plain Semaphore instead of a stream here, it's a bit overkill to use the stream machinery for this IMO…

jmartinesp: We could and did, but the code is more readable like this (IMO) and there's no performance penalty as far as I could tell by the results of the benchmark. Also, it's used in a very similar way [in some other parts of the SDK](https://github.com/matrix-org/matrix-rust-sdk/blob/224292ab3e3425600a6ee25aa5df17eb197c3831/crates/matrix-sdk/src/encryption/mod.rs#L723).

If we wanted to use a `Semaphore` instead, what would the `semaphore.acquire()` code look like if the result contains an error? I remember we discussed it in a previous PR but I'm not sure if we ever make a decision.

bnjbvr: > the code is more readable like this (IMO)

I find the code much worse, because it involves other async concepts that one must be aware of (streams), and higher-level typing that's hard to grasp just from reading the code (stream flattening). But let's keep it like this now, and we can change that later.

crates/matrix-sdk-ui/src/timeline/pinned_events_loader.rs
jmartinesp: Can I have another review, pretty please 🙏 ?

bnjbvr:
No, PRs need to be approved before getting merged.

jmartinesp:
Then we're lucky we're not in a rush, I guess.

crates/matrix-sdk-ui/src/timeline/pinned_events_loader.rs
bnjbvr: > It seems like this won't work as the target test is an integration one.

Why? Does `#[cfg(any(feature = "testing", test))]` not work?

jmartinesp: I didn't try that exact configuration, let me try with that and report back.

jmartinesp: It doesn't seem to work:

```rust
#[cfg(any(feature = "testing", test))]
pub const MAX_CONCURRENT_REQUESTS: usize = 10;

... // then

#[cfg(any(feature = "testing", test))]
pub use self::pinned_events_loader::MAX_CONCURRENT_REQUESTS;
```

Ends up with this error, I tried several variations:

```
error[E0432]: unresolved import `matrix_sdk_ui::timeline::MAX_CONCURRENT_REQUESTS`
   --> crates/matrix-sdk-ui/tests/integration/timeline/pinned_event.rs:35:5
    |
35  | use matrix_sdk_ui::timeline::MAX_CONCURRENT_REQUESTS;
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ no `MAX_CONCURRENT_REQUESTS` in `timeline`
    |
note: found an item that was configured out
   --> /Users/jorge/Developer/Element/matrix-rust-sdk/crates/matrix-sdk-ui/src/timeline/mod.rs:110:37
    |
110 | pub use self::pinned_events_loader::MAX_CONCURRENT_REQUESTS;
    |                                     ^^^^^^^^^^^^^^^^^^^^^^^
```

bnjbvr: looks like you need a bit more cfg guards then :thinking:

jmartinesp: Any idea what those might be? Are the tests in `tests/integration` running in a different way than the rest?

jmartinesp: From the [Rust docs](https://doc.rust-lang.org/book/ch11-03-test-organization.html#integration-tests):

> In Rust, integration tests are entirely external to your library. They use your library in the same way any other code would, which means they can only call functions that are part of your library’s public API. Their purpose is to test whether many parts of your library work together correctly. Units of code that work correctly on their own could have problems when integrated, so test coverage of the integrated code is important as well. To create integration tests, you first need a tests directory.

So I guess the test and testing cfgs are not used and we're back to "It seems like this won't work as the target test is an integration one." 🫤 .

jmartinesp: Since this apparaently can't be done and the doc comment was added as @Hywan requested, can we have another review so we can at least make some progress?

bnjbvr: I've found a better idea, but it involves a bit more work (sowwy). Since the `PinnedEventLoader` can use an arbitrary `PinnedEventsRoom` impl, can you make the test a non-integration test, at the bottom of this file, using a test impl for `PinnedEventsRoom`? Maybe running the assertions will be even simpler with that, as we don't have to rely on wiremock anymore?

jmartinesp: Honestly, wiremock is quite the help here... I'd rather not change everything for just a `pub` modifier. Maybe I can add this to be part of the `TimelineFocus::PinnedEvents` instead, like `max_events_to_load`?

bnjbvr: It would also make the testing more targeted by not involved network, and make it much faster to run, but I understand this is a big ask.

> Maybe I can add this to be part of the TimelineFocus::PinnedEvents instead, like max_events_to_load?

Yep, sounds great, thanks!

crates/matrix-sdk-ui/src/timeline/pinned_events_loader.rs
```diff
@@ -64,37 +71,34 @@ impl PinnedEventsLoader {
...
-                    Ok((event, related_events)) => {
-                        let mut events = vec![event];
-                        events.extend(related_events);
-                        Some(events)
-                    }
-                    Err(err) => {
-                        warn!("error when loading pinned event: {err}");
-                        None
+        let request_config = Some(RequestConfig::default().retry_limit(3));
+        let max_concurrent_requests = self.max_concurrent_requests;
```
bnjbvr, line 75: Can you inline this variable, as it's only used once?

crates/matrix-sdk-ui/tests/integration/timeline/pinned_event.rs
```diff
@@ -132,7 +139,10 @@ async fn test_new_pinned_event_ids_reload_the_timeline() {
 
     let room = test_helper.client.get_room(&room_id).unwrap();
     let timeline = Timeline::builder(&room)
-        .with_focus(TimelineFocus::PinnedEvents { max_events_to_load: 100 })
+        .with_focus(TimelineFocus::PinnedEvents {
+            max_events_to_load: 100,
+            max_concurrent_requests: 10,
+        })
```
bnjbvr, line 145: Can you make a tiny test helper that creates a pinned events focused timeline, instead? that will avoid repeating the new argument everywhere

crates/matrix-sdk-ui/tests/integration/timeline/pinned_event.rs
```diff
@@ -635,6 +672,81 @@ async fn test_edited_events_survive_pinned_event_ids_change() {
...
+                "state_key": "",
+                "type": "m.room.pinned_events",
+                "unsigned": {
+                    "age": 703422
+                }
+            }
+        )));
+
+    // Amount of time to delay the response of an /event mock request, in ms.
+    let request_delay = 50;
```
bnjbvr, line 705: Is 50 milliseconds enough for the code cov task? That would be fine to use something greater up to 1 second, fwiw.

bnjbvr (approved):
Thanks, approving assuming my comments will be addressed.

crates/matrix-sdk-ui/tests/integration/timeline/pinned_event.rs
jmartinesp: From the codecov report it seems like those lines are counted as covered too, so maybe there's no need to increase this.

Hywan (approved):

PR #3879, opened 2024-08-22
sdk: lower compile times

bnjbvr:
This series of patches, investigated using Fasterthanlime's [good advice](https://fasterthanli.me/articles/why-is-my-rust-build-so-slow#rustc-self-profiling), helped lowering the compiles times from 44 seconds to 4 on my machine, for the `matrix-sdk-ui` crate.

Compile times for the `matrix-sdk` crate have also been lowered from 33 to 12 seconds.

Overall, for a plain `cargo build`:

- before: 113 seconds
- after: 54 seconds

Each commit can be reviewed independently.

See also: https://github.com/rust-lang/rust/issues/87012

### DIFF
--- a/crates/matrix-sdk-ui/Cargo.toml
+++ b/crates/matrix-sdk-ui/Cargo.toml
@@ -26,7 +26,6 @@ async_cell = "0.2.2"
 async-once-cell = "0.5.2"
 async-rx = { workspace = true }
 async-stream = { workspace = true }
-async-trait = { workspace = true }
 chrono = "0.4.23"
 eyeball = { workspace = true }
 eyeball-im = { workspace = true }
--- a/crates/matrix-sdk-ui/src/timeline/event_handler.rs
+++ b/crates/matrix-sdk-ui/src/timeline/event_handler.rs
@@ -73,7 +73,6 @@ use crate::{
 };
 
 /// When adding an event, useful information related to the source of the event.
-#[derive(Clone)]
 pub(super) enum Flow {
     /// The event was locally created.
     Local {
@@ -103,7 +102,6 @@ pub(super) enum Flow {
     },
 }
 
-#[derive(Clone)]
 pub(super) struct TimelineEventContext {
     pub(super) sender: OwnedUserId,
     pub(super) sender_profile: Option<Profile>,
@@ -693,10 +691,10 @@ impl<'a, 'o> TimelineEventHandler<'a, 'o> {
 
     fn handle_poll_start(&mut self, c: NewUnstablePollStartEventContent, should_add: bool) {
         let mut poll_state = PollState::new(c);
-        if let Flow::Remote { event_id, .. } = self.ctx.flow.clone() {
+        if let Flow::Remote { event_id, .. } = &self.ctx.flow {
             // Applying the cache to remote events only because local echoes
             // don't have an event ID that could be referenced by responses yet.
-            self.meta.poll_pending_events.apply(&event_id, &mut poll_state);
+            self.meta.poll_pending_events.apply(event_id, &mut poll_state);
         }
 
         if should_add {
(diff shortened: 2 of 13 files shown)


poljar:
Not a review, but thank you for your service :saluting_face:

crates/matrix-sdk-ui/src/timeline/event_handler.rs
```diff
@@ -506,61 +504,57 @@ impl<'a, 'o> TimelineEventHandler<'a, 'o> {
...
+        };
+
+        let mut new_item = item.with_content(new_content, edit_json);
+
+        if let EventTimelineItemKind::Remote(remote_event) = &item.kind {
+            if let Flow::Remote { encryption_info, .. } = &self.ctx.flow {
+                new_item = new_item.with_kind(EventTimelineItemKind::Remote(
+                    remote_event.with_encryption_info(encryption_info.clone()),
+                ));
+            }
```
bnjbvr, line 552: A small note: this is slightly changing behavior, because I thought the previous behavior was buggy.

Previous behavior did this: if the item comes from a local echo, use an `encryption_info` set to `None`, and replace it in the original timeline item.

With the new behavior, we override the `encryption_info` only if the new information comes from a remote echo.

Incidentally, @pixlwave opened an issue which *I think* describes that, today: #3882. Happy to move the change to a separate PR if it's preferred; since it was only a 3 lines change, I deemed it not useful. But it's worth highlighting for the review here, since this is the only functional change of this overall PR.

pixlwave: Yep, I can confirm that this fixes the issue 👍

crates/matrix-sdk-ui/src/timeline/event_handler.rs
```diff
@@ -103,7 +102,6 @@ pub(super) enum Flow {
     },
 }
 
-#[derive(Clone)]
```
Hywan, line 106: \o/

crates/matrix-sdk-ui/src/timeline/event_handler.rs
```diff
@@ -73,7 +73,6 @@ use crate::{
 };
 
 /// When adding an event, useful information related to the source of the event.
-#[derive(Clone)]
```
Hywan, line 76: \o/

crates/matrix-sdk-ui/src/timeline/event_handler.rs
Hywan: I'm fine keeping the change here.

crates/matrix-sdk-ui/src/timeline/traits.rs
```diff
@@ -43,7 +44,7 @@ pub trait RoomExt {
     /// independent events.
     ///
     /// This is the same as using `room.timeline_builder().build()`.
-    async fn timeline(&self) -> Result<Timeline, timeline::Error>;
+    fn timeline(&self) -> impl Future<Output = Result<Timeline, timeline::Error>> + Send;
```
Hywan, line 47: I'm even wondering if we can use `async fn` in traits since we are targeting a recent Rust version. I'm lazy checking though. I'm fine with `-> impl Future`, which is exactly the same.

Hywan: I was coming back here to ensure the `+ Send` bound is present. It is present. Well done :-].

crates/matrix-sdk-ui/Cargo.toml
```diff
@@ -26,7 +26,6 @@ async_cell = "0.2.2"
 async-once-cell = "0.5.2"
 async-rx = { workspace = true }
 async-stream = { workspace = true }
-async-trait = { workspace = true }
```
Hywan, line 29: \o/

Hywan (approved):
I love what I've read. Excellent work!

crates/matrix-sdk-ui/src/timeline/traits.rs
bnjbvr: Thanks for the review!

The compiler warns if we use `async fn`, and we have some builds with warn-as-errors, so it fails those builds. The reasoning is that using an `async fn` in trait

PR #3880, opened 2024-08-22
sdk: lower compile times too by using `async_trait` less

bnjbvr:
This one is slightly more controversial than #3879, since it makes the code slightly less legible. But it's pretty sweet in that it also divides the compiles times of matrix-sdk by more than two on my machine, from 33 seconds to 12 seconds.

With this PR and #3879, the overall build time for multiverse goes from 113 seconds (cold build cache) down to 65 seconds (cold build cache too).

See also: https://github.com/rust-lang/rust/issues/87012


bnjbvr:
Will fold with the other PR.

PR #3662, opened 2024-07-08
crypto(feat): Key distribution errors for pin violations

BillCarsonFr:
Part of invisible crypto, follow up of https://github.com/matrix-org/matrix-rust-sdk/pull/3639

Fixes #3565

Following support for key pinning, we have now to manage pinning violation when encrypting olm messages (room key distribution).

The new encryption errors:
```rust
/// Depending on the sharing strategy for room keys, the distribution of the
/// room key could fail.
#[derive(Error, Debug)]
pub enum RoomKeyDistributionError {
    /// When encrypting using the IdentityBased strategy.
    /// Will be thrown when sharing room keys when there is a new identity for a
    /// user that has not been confirmed by the user.
    /// Application should display identity changes to the user as soon as
    /// possible to avoid hitting this case. If it happens the app might
    /// just retry automatically after the identity change has been
    /// notified, or offer option to cancel.
    #[error("Encryption failed because there are key pinning violation, please re-pin or verify the problematic users")]
    KeyPinningViolation(Vec<OwnedUserId>),

    /// Cross-signing is required for encryption with invisible crypto
    #[error("Encryption failed: Setup cross-signing on your account")]
    CrossSigningNotSetup,
    /// The current device needs to be verified when encrypting using the
    /// IdentityBased strategy. Apps should prevent sending in the UI to
    /// avoid hitting this case.
    #[error("Encryption failed: Verify your device to send encrypted messages")]
    SendingFromUnverifiedDevice,
}
```

- [ ] Public API changes documented in changelogs (optional)

Signed-off-by:

PR #3884, opened 2024-08-22
refactor collect_session_recipients in advance of some behavioural changes

uhoreg:
This is part of https://github.com/matrix-org/matrix-rust-sdk/pull/3662, pulled out to into a separate PR.  Recent changes in `main` made it pretty much impossible to merge this section of code from `main` into that PR, and Rich wanted to see the refactoring bits separate from the behavioural changes.  So I've re-written the refactoring.

Pulls the `match` on `sharing_strategy` outside of the `for` loop, and moves any code that is specific to one strategy into the appropriate branch.

### DIFF
--- a/crates/matrix-sdk-crypto/src/session_manager/group_sessions/share_strategy.rs
+++ b/crates/matrix-sdk-crypto/src/session_manager/group_sessions/share_strategy.rs
@@ -120,9 +120,6 @@ pub(crate) async fn collect_session_recipients(
     let users: BTreeSet<&UserId> = users.collect();
     let mut devices: BTreeMap<OwnedUserId, Vec<DeviceData>> = Default::default();
     let mut withheld_devices: Vec<(DeviceData, WithheldCode)> = Default::default();
-    let mut unsigned_devices_of_verified_users: BTreeMap<OwnedUserId, Vec<OwnedDeviceId>> =
-        Default::default();
-    let mut verified_users_with_new_identities: Vec<OwnedUserId> = Default::default();
 
     trace!(?users, ?settings, "Calculating group session recipients");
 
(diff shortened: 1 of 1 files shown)


crates/matrix-sdk-crypto/src/session_manager/group_sessions/share_strategy.rs
```diff
@@ -179,73 +182,91 @@ pub(crate) async fn collect_session_recipients(
...
+                let recipient_devices = split_recipients_withhelds_for_user_based_on_identity(
+                    user_devices,
+                    &device_owner_identity,
+                );
 
-        devices.entry(user_id.to_owned()).or_default().extend(recipients);
-        withheld_devices.extend(withheld_recipients);
-    }
+                let recipients = recipient_devices.allowed_devices;
+                let withheld_recipients = recipient_devices.denied_devices_with_code;
```
richvdh, line 256: consider inlining `withheld_recipients`

richvdh (commented):
This feels a bit odd because now `recipient_devices.unsigned_of_verified_user` etc are completely unused in the IdentityBasedStrategy case. We should probably use a different return type for `split_recipients_withhelds_for_user_based_on_identity` if we're going to do this.

---

I'd previously seen another PR from Valere making this change, and at the time dismissed it, because it felt like it was adding duplication (eg, we have to do the should_rotate logic twice) without a clear benefit. I appreciate I'm lacking the context from the later PR, so could you give me a one-liner on why you feel that duplicating that logic is a worthwhile tradeoff?

crates/matrix-sdk-crypto/src/session_manager/group_sessions/share_strategy.rs
uhoreg: Indeed, this was done in https://github.com/matrix-org/matrix-rust-sdk/pull/3662 (along with inlining `recipients`) so I've added that here now.

uhoreg:
Initially, I was just trying to make #3662 easier to review, so I wasn't looking much into why the refactor was made, but IMHO, it isn't much duplication, and the code looks easier to follow.  For example, in this version, it's clearer that `unsigned_devices_of_verified_users` and `verified_users_with_new_identities` are only used in the device-based strategy, rather than having to rely on comments to document that.

uhoreg:
Actually, looking at #3662 again, it basically also adds the `verified_users_with_new_identities` check (and we should probably combine them).  But I still think the refactor is useful because the check within the loop will still need to be duplicated (because the device-based check only runs if `error_on_verified_user_problem` is `true`, while the identity-based check runs all the time), and it will make it clear that the `verified_users_with_new_identities` check applies to both cases, whereas the `unsigned_devices_of_verified_users` check only applies to the device-based case.

uhoreg:
And, having mostly re-implemented #3662 on top of this PR, another reason for the refactor is that identity-based collection needs to [ensure that the user has cross-signing set up](https://github.com/matrix-org/matrix-rust-sdk/pull/3662/files#diff-8d716e55b5bccd9bb93db0bab9b6ea1813a8ab16eda1a98b43e767a158c38e74R174-R186), so we already need to have a check for the collection strategy outside of the `for` loop, and it may be cleaner to just have the one check in the function, rather than one outside the `for` loop and one inside.

> Actually, looking at #3662 again, it basically also adds the `verified_users_with_new_identities` check (and we should probably combine them). ...

Which leads to the somewhat awkward result that I need to move some things such as variable definitions back out to their original positions.  So I'm somewhat torn as to whether it's better to actually move those things here, with the rationale that the moves make sense for what this PR does, or if it's better to leave those in place, with the rationale that it makes more sense for the end result, even though this PR gets more "trust me, it'll make sense later" thrown into it.

richvdh (approved):
The change looks good to me, so if you're happy that this is progress in the right general direction, suggest you go ahead and land it.

uhoreg:
> The change looks good to me, so if you're happy that this is progress in the right general direction, suggest you go ahead and land it.

I'm happy with this, but I don't have permissions to merge.  Can you hit the button for me?

BTW, the next bit that builds on top of this PR is https://github.com/matrix-org/matrix-rust-sdk/pull/3896  I have some compilation issues to fix and a changelog to write, but it should be ready for review soon.

PR #3885, opened 2024-08-23
chore: Use `ruma::time` instead of instant

### LINKED ISSUE #2937: Switch to another crate instead of `instant`
We use `instant` to provide an implementation of `std::time::Instant` on wasm.

However, as https://github.com/sebcrozet/instant/issues/52 notes, that crate isn't very well maintained. Consider the use of `web-time` instead.

zecakeh:
This is a reexport of web-time.

Closes #2937.


jmartinesp (approved):
The changes make sense to me, thanks!

andybalaam (approved):
Looks good, thanks!

PR #3887, opened 2024-08-23
FFI: Expose the new cache store path to the bindings.

pixlwave:
This PR is a follow-up to https://github.com/matrix-org/matrix-rust-sdk/pull/3858, exposing the ability to set a custom cache path in the FFI.

PR to fix the complement tests is [here](https://github.com/matrix-org/complement-crypto/pull/127) - will need merging after this one for CI to continue to be happy on `main`.

### DIFF
--- a/bindings/matrix-sdk-ffi/src/client_builder.rs
+++ b/bindings/matrix-sdk-ffi/src/client_builder.rs
@@ -246,7 +246,7 @@ impl From<ClientError> for ClientBuildError {
 
 #[derive(Clone, uniffi::Object)]
 pub struct ClientBuilder {
-    session_path: Option<String>,
+    session_paths: Option<SessionPaths>,
     username: Option<String>,
     homeserver_cfg: Option<HomeserverConfig>,
     passphrase: Zeroizing<Option<String>>,
@@ -271,7 +271,7 @@ impl ClientBuilder {
     #[uniffi::constructor]
     pub fn new() -> Arc<Self> {
         Arc::new(Self {
-            session_path: None,
+            session_paths: None,
             username: None,
             homeserver_cfg: None,
             passphrase: Zeroizing::new(None),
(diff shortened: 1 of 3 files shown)


bindings/matrix-sdk-ffi/src/client_builder.rs
```diff
@@ -648,6 +669,15 @@ impl ClientBuilder {
     }
 }
 
+#[derive(Clone)]
+/// The store paths the client will use when built.
+struct SessionPaths {
+    /// The path that the client will use to store its data.
+    data_path: String,
+    /// The path that the client will use to store its caches.
+    cache_path: Option<String>,
```
Hywan, line 678: Why is it optional? If that's necessary, can you update the doc to clarify please?

crates/matrix-sdk/src/client/builder.rs
```diff
@@ -233,13 +233,22 @@ impl ClientBuilder {
...
+        self.sqlite_store_with_cache_path(path, None::<&str>, passphrase)
+    }
+
+    /// Set up the store configuration for a SQLite store with an optional path
+    /// to separate cached data from state/crypto data.
+    #[cfg(feature = "sqlite")]
+    pub fn sqlite_store_with_cache_path(
         mut self,
         path: impl AsRef<std::path::Path>,
+        cache_path: Option<impl AsRef<std::path::Path>>,
```
Hywan, line 246: `Option<impl Into<PathBuf>>` seems a simpler approach, it removes the need to call `as_ref().to_owned()`, you just need to call `into()`. And it probably removes the need to get `None::<&str>`, simply having `None` would be fine.

Hywan (changes requested):
I believe it's a nice improvement, thanks for working on this!

bindings/matrix-sdk-ffi/src/client_builder.rs
pixlwave: It probably isn't necessary tbf. I just followed the situation in the main SDK, but it would be cleaner in the FFI to force users to set this.

crates/matrix-sdk/src/client/builder.rs
pixlwave: Ok so making the cache_path above non-optional all the way through to the new method on the SDK crate means that I didn't need to do this any more.

pixlwave: (At least I think, its not the same as `path` which seems better for consistency?

Hywan: I think so.

Hywan (approved):
Let's go with this version!

PR #3888, opened 2024-08-23
common: generate wasm bindings for ShieldStateCode on wasm architectures

richvdh:
If we're building for the wasm architecture, jump through the hoops to tell
wasm_bindgen about `ShieldStateCode`. This solves the need to declare an
identical copy of `ShieldStateCode` in the wasm bindings.


bnjbvr (approved):
lgtm! needs a reformat, and that could land :+1:

PR #3892, opened 2024-08-25
sqlite: Use a transaction for chunked queries

zecakeh:
Allows the operation to be atomic, and to chunk part of a transaction.

I believe that this change makes sense on its own, but for context I will need to be able to chunk part of a transaction in an upcoming PR.


Hywan (approved):
I believe it makes sense to run the queries on a transaction. The modification makes sense and the code is clear. I don't believe it should impact the performance a lot. It's however a nice security net in case one query fails for a reason or another.

Thanks, good job!

PR #3893, opened 2024-08-26
sqlite: Rename types and traits and refactoring

zecakeh:
To make it easier to understand what the types and traits do, rename:
 - `SqliteConn` to `SqliteAsyncConn`,
 - `SqliteObjectExt` to `SqliteAsyncConnExt`
 - `SqliteConnectionExt` to `SqliteKeyValueStoreConnectionExt`
 - `SqliteObjectStoreExt` to `SqliteKeyValueStoreAsyncConnExt`.

And refactor the code to put more methods in `SqliteKeyValueStoreAsyncConnExt`, to avoid free functions that take an `&SqliteAsyncConn` and reduce code duplication.

### DIFF
--- a/crates/matrix-sdk-sqlite/src/utils.rs
+++ b/crates/matrix-sdk-sqlite/src/utils.rs
@@ -18,6 +18,7 @@ use std::{borrow::Borrow, cmp::min, iter, ops::Deref};
 use async_trait::async_trait;
 use deadpool_sqlite::Object as SqliteAsyncConn;
 use itertools::Itertools;
+use matrix_sdk_store_encryption::StoreCipher;
 use rusqlite::{limits::Limit, OptionalExtension, Params, Row, Statement, Transaction};
 
 use crate::{
@@ -280,6 +281,16 @@ impl SqliteKeyValueStoreConnExt for rusqlite::Connection {
 /// ```
 #[async_trait]
 pub(crate) trait SqliteKeyValueStoreAsyncConnExt: SqliteAsyncConnExt {
+    /// Whether the `kv` table exists in this database.
+    async fn kv_table_exists(&self) -> rusqlite::Result<bool> {
+        self.query_row(
+            "SELECT EXISTS (SELECT 1 FROM sqlite_master WHERE type = 'table' AND name = 'kv')",
+            (),
+            |row| row.get(0),
+        )
+        .await
+    }
+
     /// Get the stored value for the given key.
     async fn get_kv(&self, key: &str) -> rusqlite::Result<Option<Vec<u8>>> {
         let key = key.to_owned();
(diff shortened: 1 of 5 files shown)


crates/matrix-sdk-sqlite/src/utils.rs
```diff
@@ -255,6 +241,32 @@ impl<'a> SqliteTransactionExt for Transaction<'a> {
...
+///
+/// The table should be created like this:
+///
+/// ```sql
+/// CREATE TABLE "kv" (
+///     "key" TEXT PRIMARY KEY NOT NULL,
+///     "value" BLOB NOT NULL
+/// );
+/// ```
+pub(crate) trait SqliteKeyValueStoreConnectionExt {
```
Hywan, line 256: I would rename to `SqliteKeyValueStoreConnExt`, so that we use `Conn` everywhere in then names and it feels more consistent. What do you think?

Hywan: It also looks very similar to `SqliteKeyValueStoreAsyncConnExt` this way, though I don't know if there are supposed to represent the same thing.

crates/matrix-sdk-sqlite/src/utils.rs
```diff
@@ -280,6 +281,19 @@ impl SqliteKeyValueStoreConnectionExt for rusqlite::Connection {
 /// ```
 #[async_trait]
 pub(crate) trait SqliteKeyValueStoreAsyncConnExt: SqliteAsyncConnExt {
+    /// Whether the `kv` table exists in this database.
+    async fn kv_table_exists(&self) -> rusqlite::Result<bool> {
+        let count = self
+            .query_row(
+                "SELECT count(*) FROM sqlite_master WHERE type = 'table' AND name = 'kv'",
```
Hywan, line 288: I believe `SELECT 1` would be faster than `SELECT count(*)` here. What do you think?

crates/matrix-sdk-sqlite/src/utils.rs
```diff
@@ -290,6 +304,48 @@ pub(crate) trait SqliteKeyValueStoreAsyncConnExt: SqliteAsyncConnExt {
 
     /// Store the given value for the given key.
     async fn set_kv(&self, key: &str, value: Vec<u8>) -> rusqlite::Result<()>;
+
+    /// Get the version of the database.
+    async fn db_version(&self) -> Result<u8, OpenStoreError> {
+        let kv_exists = self.kv_table_exists().await.map_err(OpenStoreError::LoadVersion)?;
+
+        if kv_exists {
+            match self.get_kv("version").await.map_err(OpenStoreError::LoadVersion)?.as_deref() {
```
Hywan, line 313: Do we need to check if the table exists before running the query? If the query fails, can't we assume it's because the table doesn't exist (maybe by inferring the `Error` type?), and return `Ok(0)` in this case?

The idea behind this proposal is to limit the number of queries. I am aware you've mostly moved code here, so feel free to ignore my comments if you believe it's too much or too risky changes.

Hywan (changes requested):
Some feedback but overall it looks good, thanks!

crates/matrix-sdk-sqlite/src/utils.rs
zecakeh: They do basically represent the same thing, a trait for methods to interact with the `kv` store. But one is a trait with blocking methods for `rusqlite::Connection`, and the other is a trait with async methods for `SqliteAsyncConn`, which is why it's the only part of the name that changes.

crates/matrix-sdk-sqlite/src/utils.rs
zecakeh: So the issue here is that SQLite returns a generic `ErrorCode::Unknown` in this case. The only way to clearly identify that the error is a missing table would be to match the beginning of the error message, which doesn't sound great.

crates/matrix-sdk-sqlite/src/utils.rs
Hywan: ok so can you rename to `SqliteKeyValueStoreConnExt` (`Connection` becomes `Conn`)? What do you think?

crates/matrix-sdk-sqlite/src/utils.rs
Hywan: Indeed. Let's not change that then.

crates/matrix-sdk-sqlite/src/utils.rs
zecakeh: Sure

crates/matrix-sdk-sqlite/src/utils.rs
zecakeh: In your case that means that we should check whether it returns a row or not, right?

I have been wondering if we should do this:

```sql
SELECT EXISTS (SELECT 1 FROM sqlite_master WHERE type = 'table' AND name = 'kv')
```

Which will basically return a boolean.

Hywan: Yeah, that's another approach and I like it too.

zecakeh:
I made fixup commits on purpose so you can validate the changes. I'll clean up the commit history when this is accepted.

crates/matrix-sdk-sqlite/src/utils.rs
zecakeh: Done

crates/matrix-sdk-sqlite/src/utils.rs
zecakeh: Done

Hywan:
@zecakeh Ping me when it's ready to be merged.

Issue #3565, opened 2024-06-17
Invisible Crypto | Share Room Keys: Update the Identity Based sharing strategy to use the tofu_trusted flag


BillCarsonFr:
Part of https://github.com/element-hq/crypto-internal/issues/282 (Invisible Crypto).

#3563 added a new `IdentityBasedStrategy` room-key sharing strategy. #3564 added a "tofu" flag to `ReadOnlyUserIdentity`. We now need to update `share_room_keys` to check the tofu flag.

Add new cases to [`OlmError`](https://github.com/matrix-org/matrix-rust-sdk/blob/8e8d793f0d0c10f9c8a21071c3bae80ce66bb48d/crates/matrix-sdk-crypto/src/error.rs#L32-L72), to be returned by `share_room_keys`:
  * ~~`OlmError::UnconfirmedIdentityChange(Vec<OwnedUserId>)` if the user has not yet acknowledged an identity change for the recipient; and:~~ Update 2024-08-08: we no longer want this behaviour.
  * `OlmError::SendingFromUnverifiedDevice` if our current device has not yet been verified.

richvdh:
@BillCarsonFr:

> Add new errors that share_room_keys will throw based on `UnconfirmedIdentityChange(Vec<OwnedUserId>)` `SendingFromUnverifiedDevice`.

I don't understand this. What does "based on `UnconfirmedIdentityChange(Vec<OwnedUserId>)` `SendingFromUnverifiedDevice`" mean?

richvdh:
Thanks. I've updated the description. Can you confirm?

andybalaam:
@uhoreg will shepherd this one.

richvdh:
See also https://github.com/matrix-org/matrix-rust-sdk/issues/3793 which will update the sharing code to throw an error if the other user's identity was previously verified, and changes.

richvdh:
Per recent discussions, we no longer want to throw an error on message send if a TOFU-trusted user rotates their identity. In short, if identity trust matters to you so much that you are worried about the race between sending a message and seeing an identity change notification, then you should have verified.

BillCarsonFr:
> Per recent discussions, we no longer want to throw an error on message send if a TOFU-trusted user rotates their identity. In short, if identity trust matters to you so much that you are worried about the race between sending a message and seeing an identity change notification, then you should have verified.

[...]

PR #3896, opened 2024-08-27
crypto: Error when sending keys to previously-verified users with identity-based strategy

uhoreg:
This is the rest of https://github.com/matrix-org/matrix-rust-sdk/pull/3662, on top of https://github.com/matrix-org/matrix-rust-sdk/pull/3884.

3662 says:

> Part of invisible crypto, follow up of https://github.com/matrix-org/matrix-rust-sdk/pull/3639
>
> Fixes https://github.com/matrix-org/matrix-rust-sdk/issues/3565

Signed-off-by:

### DIFF
--- a/bindings/matrix-sdk-ffi/CHANGELOG.md
+++ b/bindings/matrix-sdk-ffi/CHANGELOG.md
@@ -2,14 +2,20 @@
 
 Breaking changes:
 
-- `EventSendState` now has four additional variants: `VerifiedUserHasUnsignedDevice`,
-  `VerifiedUserChangedIdentity`, `CrossSigningNotSetup`, and
-  `SendingFromUnverifiedDevice`. The first two reflect problems with verified users in
-  the room and as such can only be returned when the room key recipient strategy has
-  `error_on_verified_user_problem` set, or when using the identity-based strategy. The
-  last two indicate that your own device is not properly cross-signed, which is a
-  requirement when using the identity-based strategy, and can only be returned when
-  using the identity-based strategy.
+- `EventSendState` now has two additional variants: `CrossSigningNotSetup` and
+  `SendingFromUnverifiedDevice`. These indicate that your own device is not
+  properly cross-signed, which is a requirement when using the identity-based
+  strategy, and can only be returned when using the identity-based strategy.
+
+  In addition, the `VerifiedUserHasUnsignedDevice` and
+  `VerifiedUserChangedIdentity` variants can be returned when using the
+  identity-based strategy, in addition to when using the device-based strategy
+  with `error_on_verified_user_problem` is set.
+
+- `EventSendState` now has two additional variants: `VerifiedUserHasUnsignedDevice` and
+  `VerifiedUserChangedIdentity`. These reflect problems with verified users in the room
+  and as such can only be returned when the room key recipient strategy has
+  `error_on_verified_user_problem` set.
 
 - The `AuthenticationService` has been removed:
     - Instead of calling `configure_homeserver`, build your own client with the `serverNameOrHomeserverUrl` builder
--- a/bindings/matrix-sdk-ffi/CHANGELOG.md
+++ b/bindings/matrix-sdk-ffi/CHANGELOG.md
@@ -2,10 +2,14 @@
 
 Breaking changes:
 
-- `EventSendState` now has two additional variants: `VerifiedUserHasUnsignedDevice` and
-  `VerifiedUserChangedIdentity`. These reflect problems with verified users in the room
-  and as such can only be returned when the room key recipient strategy has
-  `error_on_verified_user_problem` set.
+- `EventSendState` now has four additional variants: `VerifiedUserHasUnsignedDevice`,
+  `VerifiedUserChangedIdentity`, `CrossSigningNotSetup`, and
+  `SendingFromUnverifiedDevice`. The first two reflect problems with verified users in
+  the room and as such can only be returned when the room key recipient strategy has
+  `error_on_verified_user_problem` set, or when using the identity-based strategy. The
+  last two indicate that your own device is not properly cross-signed, which is a
+  requirement when using the identity-based strategy, and can only be returned when
+  using the identity-based strategy.
 
 - The `AuthenticationService` has been removed:
     - Instead of calling `configure_homeserver`, build your own client with the `serverNameOrHomeserverUrl` builder
--- a/crates/matrix-sdk-crypto/src/session_manager/group_sessions/share_strategy.rs
+++ b/crates/matrix-sdk-crypto/src/session_manager/group_sessions/share_strategy.rs
@@ -1322,8 +1322,6 @@ mod tests {
             .await
             .unwrap()
             .unwrap()
-            .other()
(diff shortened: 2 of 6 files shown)


uhoreg:
Compared with 3662, this uses `has_identity_verification_violation` instead of its own calculation for whether the sender was previously verified, with the end result that it no longer looks at the pinned master signing key.  I think that this is fine (as a result of [a decision that we change the behaviour](https://github.com/matrix-org/matrix-rust-sdk/issues/3565#issuecomment-2275908743)), but I think it also means that the pinning code is now unused.  I think we need to double-check whether that's the case, and if so, possibly remove it.

crates/matrix-sdk-crypto/src/session_manager/group_sessions/share_strategy.rs
```diff
@@ -1122,6 +1150,187 @@ mod tests {
         assert_eq!(code, &WithheldCode::Unauthorised);
     }
 
+    #[async_test]
+    async fn test_share_identity_strategy_no_cross_signing() {
+        // Test key sharing with the identity-based strategy with different
+        // states of our own verification.
```
richvdh, line 1156: make this a doc-comment for the test function, please

crates/matrix-sdk-crypto/src/session_manager/group_sessions/share_strategy.rs
```diff
@@ -1122,6 +1150,187 @@ mod tests {
...
+
+        // Starting off, we have not yet set up our own cross-signing, so
```
richvdh, line 1158: > set up

:tada:   :smile:

crates/matrix-sdk-crypto/src/session_manager/group_sessions/share_strategy.rs
```diff
@@ -1122,6 +1150,187 @@ mod tests {
...
+
+        // Dan has two devices, but only one is cross-signed, so there should
+        // only be one key share.
+        assert_eq!(requests.len(), 1);
+    }
+
+    #[async_test]
+    async fn test_share_identity_strategy_report_pinning_violation() {
+        // Test that identity-based key sharing gives an error when a verified
+        // user changes their identity.
```
richvdh, line 1244: doc-comment please

crates/matrix-sdk-crypto/src/session_manager/group_sessions/share_strategy.rs
```diff
@@ -1122,6 +1150,187 @@ mod tests {
...
+        )
+        .await;
+
+        machine.bootstrap_cross_signing(false).await.unwrap();
+
+        let keys_query = IdentityChangeDataSet::key_query_with_identity_a();
+        machine.mark_request_as_sent(&TransactionId::new(), &keys_query).await.unwrap();
+
+        let keys_query = MaloIdentityChangeDataSet::initial_key_query();
+        machine.mark_request_as_sent(&TransactionId::new(), &keys_query).await.unwrap();
```
richvdh, line 1257: looks like we're mixing up data from two different test sets. That's okay, if that test data is the most appropriate for this test, but could you add a comment in here to explain what's going on? Which users' keys are we importing here?

crates/matrix-sdk-crypto/src/session_manager/group_sessions/share_strategy.rs
```diff
@@ -1122,6 +1150,187 @@ mod tests {
...
+        machine
+            .get_identity(MaloIdentityChangeDataSet::user_id(), None)
+            .await
+            .unwrap()
+            .unwrap()
+            .other()
+            .unwrap()
+            .mark_as_previously_verified()
+            .await
+            .unwrap();
```
richvdh, line 1284: we're changing our own identity and that of someone else?

bindings/matrix-sdk-ffi/CHANGELOG.md
```diff
@@ -2,10 +2,14 @@
...
-  and as such can only be returned when the room key recipient strategy has
-  `error_on_verified_user_problem` set.
+- `EventSendState` now has four additional variants: `VerifiedUserHasUnsignedDevice`,
+  `VerifiedUserChangedIdentity`, `CrossSigningNotSetup`, and
+  `SendingFromUnverifiedDevice`. The first two reflect problems with verified users in
+  the room and as such can only be returned when the room key recipient strategy has
+  `error_on_verified_user_problem` set, or when using the identity-based strategy. The
+  last two indicate that your own device is not properly cross-signed, which is a
+  requirement when using the identity-based strategy, and can only be returned when
+  using the identity-based strategy.
```
richvdh, line 12: erm, we need to make new changelog entries, not just modify old changelogs that apply to old versions that are already built into applications.

richvdh (changes requested):

bindings/matrix-sdk-ffi/CHANGELOG.md
uhoreg: OK, I wasn't sure about this, because it's under the "unreleased" section, so wasn't sure if it was better to modify the existing entry, or have two separate entries that are related to the same area.

richvdh: unfortunately everything is "unreleased" at the moment :(

crates/matrix-sdk-crypto/src/session_manager/group_sessions/share_strategy.rs
uhoreg: We're changing two other users' identities.  Our own user is `KeyDistributionTestData::me_id()`

crates/matrix-sdk-crypto/src/session_manager/group_sessions/share_strategy.rs
uhoreg: Added some comments.  I also noticed that we were basically just doing the same thing to both users, so I changed it so that we withdraw verification on one user, and re-verify the second user, so that it tests both ways of resolving the issue.

uhoreg:
I think that I've addressed all of the comments. I can re-write history if/when needed.

crates/matrix-sdk-crypto/src/session_manager/group_sessions/share_strategy.rs
```diff
@@ -1293,38 +1301,101 @@ mod tests {
...
-                .await
-                .unwrap()
+        // We resolve this for user1 by withdrawing their verification.
+        machine
+            .get_identity(user1, None)
+            .await
+            .unwrap()
+            .unwrap()
+            .other()
+            .unwrap()
```
richvdh, line 1326: think this is redundant. (And likewise below)

richvdh (approved):
lgtm otherwise

BillCarsonFr:
> but I think it also means that the pinning code is now unused. I think we need to double-check whether that's the case, and if so, possibly remove it.

It will be needed at least to display the warning notice that an identity has changed https://github.com/element-hq/element-meta/issues/2513

uhoreg:
@richvdh can you hit the merge button for me?

PR #3899, opened 2024-08-27
crypto: Configurable sender trust checking on decrypting

uhoreg:
Replaces https://github.com/matrix-org/matrix-rust-sdk/pull/3701.  Part of invisible crypto.

The room event decryption function now takes a `DecryptionSetting` which allows you to require that event senders have a certain trust level based on their cross-signing status.  Events that do not meet the required trust level will fail to decrypt with an error.

Signed-off-by:

### DIFF
--- a/crates/matrix-sdk-common/src/deserialized_responses.rs
+++ b/crates/matrix-sdk-common/src/deserialized_responses.rs
@@ -189,6 +189,22 @@ pub enum VerificationLevel {
     None(DeviceLinkProblem),
 }
 
+impl fmt::Display for VerificationLevel {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> Result<(), fmt::Error> {
+        let display = match self {
+            VerificationLevel::UnverifiedIdentity => "The sender's identity was not verified",
+            VerificationLevel::PreviouslyVerified => {
+                "The sender's identity was previously verified but has changed"
+            }
+            VerificationLevel::UnsignedDevice => {
+                "The sending device was not signed by the user's identity"
+            }
+            VerificationLevel::None(..) => "The sending device is not known",
+        };
+        write!(f, "{}", display)
+    }
+}
+
 /// The sub-enum containing detailed information on why we were not able to link
 /// a message back to a device.
 #[derive(Clone, Debug, Deserialize, Serialize, PartialEq, Eq)]
--- a/crates/matrix-sdk-crypto/src/lib.rs
+++ b/crates/matrix-sdk-crypto/src/lib.rs
@@ -91,6 +91,7 @@ pub use requests::{
     IncomingResponse, KeysBackupRequest, KeysQueryRequest, OutgoingRequest, OutgoingRequests,
     OutgoingVerificationRequest, RoomMessageRequest, ToDeviceRequest, UploadSigningKeysRequest,
 };
+use serde::{Deserialize, Serialize};
 pub use session_manager::CollectStrategy;
 pub use store::{
     CrossSigningKeyExport, CryptoStoreError, SecretImportError, SecretInfo, TrackedUser,
@@ -112,3 +113,25 @@ matrix_sdk_test::init_tracing_for_tests!();
 
 #[cfg(feature = "uniffi")]
 uniffi::setup_scaffolding!();
+
+/// The trust level in the sender's device that is required to decrypt an
+/// event.
+#[derive(Clone, Copy, Debug, Deserialize, Serialize)]
+pub enum TrustRequirement {
+    /// Decrypt events from everyone regardless of trust.
+    Untrusted,
+    /// Only decrypt events from cross-signed devices or legacy sessions (Megolm
+    /// sessions created before we started collecting trust information).
+    CrossSignedOrLegacy,
+    /// Only decrypt events from cross-signed devices.
+    CrossSigned,
+}
+
+/// Settings for decrypting messages
+#[derive(Clone, Debug, Deserialize, Serialize)]
+pub struct DecryptionSettings {
+    /// The trust level in the sender's device that is required to decrypt the
+    /// event. If the sender's device is not sufficiently trusted,
+    /// [`MegolmError::SenderIdentityNotTrusted`] will be returned.
+    pub sender_device_trust_requirement: TrustRequirement,
+}
--- a/crates/matrix-sdk-crypto/src/machine/mod.rs
+++ b/crates/matrix-sdk-crypto/src/machine/mod.rs
@@ -934,8 +934,6 @@ impl OlmMachine {
         &self,
         room_id: &RoomId,
     ) -> OlmResult<()> {
-        use crate::olm::SenderData;
-
         let (_, session) = self
             .inner
             .group_session_manager
@@ -957,8 +955,6 @@ impl OlmMachine {
         &self,
         room_id: &RoomId,
     ) -> OlmResult<InboundGroupSession> {
-        use crate::olm::SenderData;
-
         let (_, session) = self
             .inner
             .group_session_manager
@@ -1609,6 +1605,13 @@ impl OlmMachine {
         match result {
             Ok((decrypted_event, _)) => {
                 let encryption_info = self.get_encryption_info(&session, &event.sender).await?;
+
+                self.check_sender_trust_requirement(
+                    &session,
+                    &encryption_info,
+                    &decryption_settings.sender_device_trust_requirement,
+                )?;
+
                 Ok((decrypted_event, encryption_info))
             }
             Err(error) => Err(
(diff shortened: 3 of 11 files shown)


richvdh:
@uhoreg does this relate to a specific issue? Please link it if so. If not, please give more context about what we're trying to achieve in this PR.

bnjbvr:
(Also please follow the [contribution guidelines](https://github.com/matrix-org/matrix-rust-sdk/blob/main/CONTRIBUTING.md#commit-messages-and-pr-title-guidelines) for naming PRs and commits! It helps figuring out which PR relates to what :))

crates/matrix-sdk-crypto/src/machine/mod.rs
```diff
@@ -1630,6 +1637,55 @@ impl OlmMachine {
...
+        trust_requirement: &TrustRequirement,
+    ) -> MegolmResult<()> {
+        /// Get the error from the encryption information.
+        fn encryption_info_to_error(encryption_info: &EncryptionInfo) -> MegolmResult<()> {
+            // When this is called, the verification state *must* be unverified,
+            // otherwise the sender_data would have been SenderVerified
+            let VerificationState::Unverified(verification_level) =
+                &encryption_info.verification_state
+            else {
+                panic!("inconsistent verification state");
```
richvdh, line 1655: could maybe use `unreachable!` here

crates/matrix-sdk-crypto/src/machine/tests/decryption_verification_state.rs
```diff
@@ -380,3 +381,253 @@ async fn test_verification_states_multiple_device() {
...
+            identities: IdentityChanges {
+                new: vec![OtherUserIdentityData::new(alice_msk.clone(), alice_ssk.clone())
+                    .unwrap()
+                    .into()],
+                ..Default::default()
+            },
+            ..Default::default()
+        })
+        .await
+        .unwrap();
```
richvdh, line 463: looks like we have two copies of this code? factor it out, maybe?

crates/matrix-sdk-crypto/src/machine/tests/decryption_verification_state.rs
```diff
@@ -380,3 +381,253 @@ async fn test_verification_states_multiple_device() {
...
+        tests::alice_id(),
+        tests::user_id(),
+        false,
+    )
+    .await;
+    let room_id = room_id!("!test:example.org");
+    let (event, session_id) = encrypt_message(&alice, room_id, &bob, "Secret message").await;
+
+    // Set the sender data to various values, and test that we can or can't
+    // decrypt, depending on what the trust requirement is.
```
richvdh, line 397: It's not obvious how this comment relates to the code immediately after it. Or does it apply to the whole test? it should maybe be a doc-comment on the function if so.

richvdh: In fact, the test needs a doc-comment describing what it is testing in any case.

crates/matrix-sdk-crypto/src/machine/tests/decryption_verification_state.rs
```diff
@@ -373,3 +381,254 @@ async fn test_verification_states_multiple_device() {
...
+
+    let mut session =
+        bob.store().get_inbound_group_session(room_id, &session_id).await.unwrap().unwrap();
+
+    // A session where we have the device keys but no cross-signing information
+    // should be decryptable only when the trust requirement allows untrusted or
+    // legacy sessions
```
richvdh, line 404:
```suggestion
    // Set the SenderData on the megolm session used to encrypt `event` to `DeviceInfo`
    // (ie,  we have the device keys but no cross-signing information).
    // Events sent on such a session should be decryptable only when the trust 
    // requirement allows untrusted or legacy sessions.
    let mut session =
        bob.store().get_inbound_group_session(room_id, &session_id).await.unwrap().unwrap();
```

crates/matrix-sdk-crypto/src/machine/tests/decryption_verification_state.rs
```diff
@@ -373,3 +381,254 @@ async fn test_verification_states_multiple_device() {
...
+        tests::user_id(),
+        false,
+    )
+    .await;
+    bob.bootstrap_cross_signing(false).await.unwrap();
+    let room_id = room_id!("!test:example.org");
+    let (event, session_id) = encrypt_message(&alice, room_id, &bob, "Secret message").await;
+
+    // Simulate Alice's cross-signing key changing after having been verified by
+    // setting the `previously_verified` flag
```
richvdh, line 495: we don't actually seem to set the `previously_verified` flag until about 50 lines later?

richvdh (changes requested):

crates/matrix-sdk-crypto/src/types/decryption.rs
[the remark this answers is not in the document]
uhoreg: Right, I meant to comment on this.  I put it in `types` because [Damir suggested it](https://github.com/matrix-org/matrix-rust-sdk/pull/3701#discussion_r1688648114).  I don't mind changing the location, but I'm not sure where the right place is.  I had initially put it in `olm::group_sessions::inbound`, for symmetry with `EncryptionSettings` in `olm::group_sessions::outbound`, but it isn't used in that module (unlike `EncryptionSettings` in `...::outbound`).  It's only used in `OlmMachine`, so maybe it belongs in that module? But there isn't anything else like it in that module.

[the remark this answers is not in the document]
richvdh: oh, hrm.

[`matrix_sdk_crypto::olm`](https://matrix-org.github.io/matrix-rust-sdk/matrix_sdk_crypto/olm/index.html) says "Note: You’ll only be interested in these if you are implementing a custom CryptoStore", which doesn't sound right either.

Honestly I'd just put it in the top-level module for now (`matrix_sdk_crypto` itself). That namespace re-exports all the useful types currently (including `EncryptionSettings`) so I don't think it clutters the public API too badly to do so. If we want to move its definition somewhere else later on, we can do so.

crates/matrix-sdk-common/src/deserialized_responses.rs
```diff
@@ -166,24 +167,28 @@ impl VerificationState {
 
 /// The sub-enum containing detailed information on why a message is considered
 /// to be unverified.
-#[derive(Clone, Debug, Deserialize, Serialize, PartialEq, Eq)]
+#[derive(Clone, Debug, Deserialize, Error, Serialize, PartialEq, Eq)]
```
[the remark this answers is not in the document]
uhoreg: Yes, the `Error`-ness itself isn't used, but it does allow me to define the `Display` trait by using the `#[error(..)]` attribute so that `[MegolmError](SenderIdentityNotTrusted)`.  It's slightly abusing the macro, but I like the `#[error(..)]` attribute because it seems cleaner to me than creating an `impl` block for it.  Alternatively, I could use something like the [`derive_more`](https://docs.rs/derive_more/latest/derive_more/derive.Display.html) crate, though I'm not a fan of adding extra dependencies.

uhoreg:
I think that I've addressed all the review comments

bnjbvr:
Can you follow the contribution guidelines for the naming of this PR and the commits, please?

crates/matrix-sdk-common/src/deserialized_responses.rs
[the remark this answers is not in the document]
richvdh: if all you need is an `impl Display`, can't you just write one? Should be pretty simple.

crates/matrix-sdk-crypto/src/lib.rs
```diff
@@ -112,3 +113,25 @@ matrix_sdk_test::init_tracing_for_tests!();
 
 #[cfg(feature = "uniffi")]
 uniffi::setup_scaffolding!();
+
+/// The trust level in the sender's device that is required to decrypt an
+/// event. If the sender's device is not sufficiently trusted,
+/// [`MegolmError::SenderIdentityNotTrusted`] will be returned.
```
richvdh, line 119: Nit: this second sentence maybe applies more to `DecryptionSrettings::sender_device_trust_requirement` than the type.

crates/matrix-sdk-crypto/src/machine/tests/decryption_verification_state.rs
```diff
@@ -393,15 +397,13 @@ async fn test_decryption_trust_requirement() {
...
 
-    // Set the sender data to various values, and test that we can or can't
-    // decrypt, depending on what the trust requirement is.
-
+    // Set the SenderData on the megolm session used to encrypt `event` to
+    // `DeviceInfo` (ie,  we have the device keys but no cross-signing
+    // information). Events sent on such a session should be decryptable only
+    // when the trust requirement allows untrusted or legacy sessions.
     let mut session =
         bob.store().get_inbound_group_session(room_id, &session_id).await.unwrap().unwrap();
```
richvdh, line 405: nit: this empty line feels a bit unhelpful, given the comment at lines 400-403 applies as much to the bit below the blank line as it does to lines 404-405.

richvdh (changes requested):
Looks good to me now except:
* Per https://github.com/matrix-org/matrix-rust-sdk/pull/3899#discussion_r1734646756, `#[derive(Error)]` seems like an unclear way to implement `Display`.
* Per https://github.com/matrix-org/matrix-rust-sdk/pull/3899#issuecomment-2315227590, please update the description of the PR to describe what it is we're fixing.
* Per benji's requests, please give the PR a name that reflects the area of the code that is affected (`crypto:`, I guess)
* Also please rebase your commits to remove the "minor fixes" and "address review", and rename them so that they match the contributing guide. If you pick the right point on `main` to rebase on top of (468ee536444dd), then github will show the force-push with no changes, which means we keep an audit trail that the code has been reviewed.

richvdh (approved):
:tada: :ship:

PR #3900, opened 2024-08-27
indexeddb: Reinstate crypto store integration tests

richvdh:
We are supposed to run the cryptostore integ tests twice for indexeddb: once with encryption enabled, and once without.

We are no longer running the tests without encryption. I think this was accidentally removed in [#3799](https://github.com/matrix-org/matrix-rust-sdk/pull/3799/commits/3dae56d34338086aa681e3a5a932a9f4c9d86570#diff-a6cc12d0312753050af771b6007b12f02f0997e8578eeb0a11a17df87f79817aL1804).


bnjbvr (approved):
Oopsy, thanks!

PR #3902, opened 2024-08-28
sqlite: Make migrations atomic

zecakeh:
Setting the version number only when all migrations are done means that the version of the db will be wrong if a migration fails.

So this sets the version of the database for each migration transaction.

This includes also a related second commit that merges `init` and `run_migrations` for `SqliteEventCacheStore`, since they don't need to be separate like in `SqliteStateStore`.


stefanceriu (approved):
Makes perfect sense to me, thank you!

PR #3908, opened 2024-08-28
feat(sdk): Add `Client::server`

Hywan:
This patch adds `Client::server` as an `Option<Url>`. This should not be confused with the `homeserver`. For example, in the case of Matrix, `matrix.org` is the server, and `matrix-client.matrix.org` is the homeserver.

This patch should be reviewed commit-by-commit for the sake of simplicity. Code is moved into their own module for clarity.

### DIFF
--- a/crates/matrix-sdk/src/client/builder/homeserver_config.rs
+++ b/crates/matrix-sdk/src/client/builder/homeserver_config.rs
@@ -36,8 +36,8 @@ pub(super) enum HomeserverConfig {
     /// A host/port pair representing a server URL.
     ServerName { server: OwnedServerName, protocol: UrlScheme },
 
-    /// First attempts to build as a server name, then falls back to a URL,
-    /// failing if no valid homeserver is found.
+    /// First attempts to build as a server URL, then falls back to a
+    /// homeserver URL, failing if no valid homeserver is found.
     ServerNameOrUrl(String),
 }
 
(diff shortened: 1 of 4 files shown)


crates/matrix-sdk/src/client/builder/homeserver_config.rs
```diff
@@ -30,14 +30,14 @@ use crate::{
 /// Configuration for the homeserver.
 #[derive(Clone, Debug)]
 pub(super) enum HomeserverConfig {
-    /// A precise URL, including the protocol.
+    /// A server name URL, including the protocol.
```
jmartinesp, line 33: Should this be `a homeserver URL`?

jmartinesp (commented):
The changes look good, although the distinction between 'server' and 'home server' is a bit difficult to grasp at first. I might have found an issue with the docs though, if I understood the API correctly.

crates/matrix-sdk/src/client/builder/homeserver_config.rs
Hywan: Indeed! I want to rename all these variants in another PR because it is so confusing.

Hywan: Fixed and rebased.

jmartinesp (approved):
Thanks for the changes!

PR #3866, opened 2024-08-20
ffi: expose new `SessionRecipientCollectionError`s to application

richvdh:
Fixes https://github.com/matrix-org/matrix-rust-sdk/issues/3842

Part of https://github.com/element-hq/element-internal/issues/614

PR #3909, opened 2024-08-29
send queue: allow unwedging and retry after recipient trust changes

### LINKED ISSUE #3837: [Task] Restore the method to retry sending of failed messages
Part of the crypto work from https://github.com/element-hq/element-meta/issues/2488

Whilst most send failures can be resolved automatically (e.g. network failures, invalid access token etc) or are deemed unrecoverable (e.g. invalid data, too large etc) there are some instances where a message can fail to send but may work later:

- The other user was verified but has become unverified (will succeed when they're verified again).
- The other user is verified but has an unverified device (will succeed when the device has been verified).

Having a method to retry sending of an item will be useful for this.

### LINKED ISSUE #3838: [Task] Add a method to resolve send failures due to unverified devices and retry sending.
Part of https://github.com/element-hq/element-meta/issues/2501 and likely requires https://github.com/matrix-org/matrix-rust-sdk/issues/3837

Once the error is surface from https://github.com/matrix-org/matrix-rust-sdk/issues/3792 we need a way to allow a user to resolve the error and retry sending. This can be done in the following way:

> For unverified device: In order to send anyhow, you have to blacklist or whitelist the device using `Device::set_local_trust` (you have all the problematic devices listed in the error) then just resend⁦

### LINKED ISSUE #3839: [Task] Add a method to resolve send failures due to an identity change and retry sending.
Part of https://github.com/element-hq/element-meta/issues/2500 and likely requires https://github.com/matrix-org/matrix-rust-sdk/issues/3837

Once the error is surface from https://github.com/matrix-org/matrix-rust-sdk/issues/3793 we need a way to allow a user to resolve the error and retry sending. This can be done in the following way:

> For "verification violation" there is a new `UserIdentity::withdraw_verification` that you can call, then retry sending. (you can also blacklist if you prefer via `Device::set_local_trust`)
> ⁦And same the error gaves you all the problematic `user_id`, so you can just iterate, withdraw_verification then resend

stefanceriu:
* add a generic mechanism for unwedging and resending requests based on their transaction identifier
* expose methods for manually withdrawing certain users' verification or trusting their devices and resending failed messages

This will be used in conjunction with the send errors introduced in https://github.com/matrix-org/matrix-rust-sdk/pull/3866 to allow the user to force resend failed messages after manually resolving trust issues.

Fixes #3839, fixes #3838, fixes #3837.

### DIFF
--- a/bindings/matrix-sdk-ffi/src/error.rs
+++ b/bindings/matrix-sdk-ffi/src/error.rs
@@ -134,14 +134,14 @@ impl From<EventCacheError> for ClientError {
     }
 }
 
-impl From<RoomSendQueueError> for ClientError {
-    fn from(e: RoomSendQueueError) -> Self {
+impl From<EditError> for ClientError {
+    fn from(e: EditError) -> Self {
         Self::new(e)
     }
 }
 
-impl From<EditError> for ClientError {
-    fn from(e: EditError) -> Self {
+impl From<RoomSendQueueError> for ClientError {
+    fn from(e: RoomSendQueueError) -> Self {
         Self::new(e)
     }
 }
--- a/bindings/matrix-sdk-ffi/src/room.rs
+++ b/bindings/matrix-sdk-ffi/src/room.rs
@@ -1,7 +1,8 @@
-use std::sync::Arc;
+use std::{collections::HashMap, sync::Arc};
 
 use anyhow::{Context, Result};
 use matrix_sdk::{
+    crypto::LocalTrust,
     event_cache::paginator::PaginatorError,
     room::{
         edit::EditedContent, power_levels::RoomPowerLevelChanges, Room as SdkRoom, RoomMemberRole,
@@ -23,7 +24,7 @@ use ruma::{
         },
         TimelineEventType,
     },
-    EventId, Int, RoomAliasId, UserId,
+    EventId, Int, OwnedDeviceId, OwnedTransactionId, OwnedUserId, RoomAliasId, UserId,
 };
 use tokio::sync::RwLock;
 use tracing::error;
@@ -735,6 +736,75 @@ impl Room {
             .await?;
 
         self.inner.send_queue().send(replacement_event).await?;
+        Ok(())
+    }
+
+    /// Remove verification requirements for the given users and
+    /// resend messages that failed to send because their identities were no
+    /// longer verified (in response to
+    /// `SessionRecipientCollectionError::VerifiedUserChangedIdentity`)
+    ///
+    /// # Arguments
+    ///
+    /// * `user_ids` - The list of users identifiers received in the error
+    /// * `transaction_id` - The send queue transaction identifier of the local
+    ///   echo the send error applies to
+    pub async fn withdraw_verification_and_resend(
+        &self,
+        user_ids: Vec<String>,
+        transaction_id: String,
+    ) -> Result<(), ClientError> {
+        let transaction_id: OwnedTransactionId = transaction_id.into();
+
+        let user_ids: Vec<OwnedUserId> =
+            user_ids.iter().map(UserId::parse).collect::<Result<_, _>>()?;
+
+        let encryption = self.inner.client().encryption();
+
+        for user_id in user_ids {
+            if let Some(user_identity) = encryption.get_user_identity(&user_id).await? {
+                user_identity.withdraw_verification().await?;
+            }
+        }
+
+        self.inner.send_queue().unwedge(&transaction_id).await?;
+
+        Ok(())
+    }
+
+    /// Set the local trust for the given devices to `LocalTrust::Ignored`
+    /// and resend messages that failed to send because said devices are
+    /// unverified (in response to
+    /// `SessionRecipientCollectionError::VerifiedUserHasUnsignedDevice`).
+    /// # Arguments
+    ///
+    /// * `devices` - The map of users identifiers to device identifiers
+    ///   received in the error
+    /// * `transaction_id` - The send queue transaction identifier of the local
+    ///   echo the send error applies to
+    pub async fn ignore_device_trust_and_resend(
+        &self,
+        devices: HashMap<String, Vec<String>>,
+        transaction_id: String,
+    ) -> Result<(), ClientError> {
+        let transaction_id: OwnedTransactionId = transaction_id.into();
+
+        let encryption = self.inner.client().encryption();
+
+        for (user_id, device_ids) in devices.iter() {
+            let user_id = UserId::parse(user_id)?;
+
+            for device_id in device_ids {
+                let device_id: OwnedDeviceId = device_id.as_str().into();
+
+                if let Some(device) = encryption.get_device(&user_id, &device_id).await? {
+                    device.set_local_trust(LocalTrust::Ignored).await?;
+                }
+            }
+        }
+
+        self.inner.send_queue().unwedge(&transaction_id).await?;
+
         Ok(())
     }
 }
--- a/crates/matrix-sdk/src/send_queue.rs
+++ b/crates/matrix-sdk/src/send_queue.rs
@@ -556,6 +556,20 @@ impl RoomSendQueue {
             self.inner.notifier.notify_one();
         }
     }
+
+    /// Unwedge a local echo identified by its transaction identifier and try to
+    /// resend it.
+    pub async fn unwedge(&self, transaction_id: &TransactionId) -> Result<(), RoomSendQueueError> {
+        self.inner
+            .queue
+            .mark_as_unwedged(transaction_id)
+            .await
+            .map_err(RoomSendQueueError::StorageError)?;
+
+        self.inner.notifier.notify_one();
+
+        Ok(())
+    }
 }
 
 struct RoomSendQueueInner {
@@ -670,6 +684,19 @@ impl QueueStorage {
             .client()?
             .store()
             .update_send_queue_event_status(&self.room_id, transaction_id, true)
+            .await?)
+    }
+
+    /// Marks an event identified with the given transaction id as being now
+    /// unwedged and adds it back to the queue
+    async fn mark_as_unwedged(
+        &self,
+        transaction_id: &TransactionId,
+    ) -> Result<(), RoomSendQueueStorageError> {
+        Ok(self
+            .client()?
+            .store()
+            .update_send_queue_event_status(&self.room_id, transaction_id, false)
             .await?)
     }
 
--- a/crates/matrix-sdk/src/send_queue.rs
+++ b/crates/matrix-sdk/src/send_queue.rs
@@ -568,6 +568,11 @@ impl RoomSendQueue {
 
         self.inner.notifier.notify_one();
 
+        let _ = self
+            .inner
+            .updates
+            .send(RoomSendQueueUpdate::RetryEvent { transaction_id: transaction_id.to_owned() });
+
         Ok(())
     }
 }
@@ -1192,6 +1197,12 @@ pub enum RoomSendQueueUpdate {
         /// while an unrecoverable error will be parked, until the user
         /// decides to cancel sending it.
         is_recoverable: bool,
+    },
+
+    /// The event has been unwedged and sending is now being retried.
+    RetryEvent {
+        /// Transaction id used to identify this event.
+        transaction_id: OwnedTransactionId,
     },
 
     /// The event has been sent to the server, and the query returned
--- a/crates/matrix-sdk/src/send_queue.rs
+++ b/crates/matrix-sdk/src/send_queue.rs
@@ -566,6 +566,7 @@ impl RoomSendQueue {
             .await
             .map_err(RoomSendQueueError::StorageError)?;
 
+        // Wake up the queue, in case the room was asleep before unwedging the event.
         self.inner.notifier.notify_one();
 
         let _ = self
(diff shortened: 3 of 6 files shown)


crates/matrix-sdk/src/room/mod.rs
```diff
@@ -2957,6 +2962,54 @@ impl Room {
...
+        &self,
+        devices: HashMap<OwnedUserId, Vec<OwnedDeviceId>>,
+        transaction_id: &TransactionId,
+    ) -> Result<()> {
+        let encryption = self.client().encryption();
+
+        for user_id in devices.keys() {
+            for device_id in &devices[user_id] {
+                if let Some(device) = encryption.get_device(user_id, device_id).await? {
+                    device.set_local_trust(LocalTrust::Verified).await?;
```
pixlwave, line 3004: Just tested this and turns out we need to set it to `LocalTrust::Ignored` here.

crates/matrix-sdk/src/room/mod.rs
```diff
@@ -2957,6 +2962,54 @@ impl Room {
...
+
+        Ok(())
+    }
+
+    /// Set the local trust for the given devices to `LocalTrust::Verified`
+    /// and resend messages that failed to send because said devices are
+    /// unverified (in response to
+    /// `SessionRecipientCollectionError::VerifiedUserHasUnsignedDevice`).
+    #[cfg(feature = "e2e-encryption")]
+    pub async fn trust_devices_and_resend(
```
pixlwave, line 2994: I guess this should probably then become something like `ignore_device_trust_and_resend`

crates/matrix-sdk/src/send_queue.rs
```diff
@@ -556,6 +556,20 @@ impl RoomSendQueue {
...
+    /// Unwedge a local echo identified by its transaction identifier and try to
+    /// resend it.
+    pub async fn unwedge(&self, transaction_id: &TransactionId) -> Result<(), RoomSendQueueError> {
+        self.inner
+            .queue
+            .mark_as_unwedged(transaction_id)
+            .await
+            .map_err(RoomSendQueueError::StorageError)?;
+
+        self.inner.notifier.notify_one();
```
bnjbvr, line 569: Can you add a comment here similar to others doing the same thing, explaining why we're notifying?

crates/matrix-sdk/src/send_queue.rs
```diff
@@ -673,6 +687,19 @@ impl QueueStorage {
             .await?)
     }
 
+    /// Marks an event identified with the given transaction id as being now
+    /// unwedged and adds it back to the queue
```
bnjbvr, line 696: uber-nit: final dot at the end for consistency with the rest of this file :)

crates/matrix-sdk/tests/integration/send_queue.rs
```diff
@@ -1398,6 +1398,94 @@ async fn test_unrecoverable_errors() {
...
+    // The room updates will report the error for the first message as unrecoverable
+    // too.
+    assert_update!(watch => error { recoverable=false, txn=txn1 });
+
+    // No queue is being disabled, because the error was unrecoverable.
+    assert!(room.send_queue().is_enabled());
+    assert!(client.send_queue().is_enabled());
+
+    // Unwedge the previously failed message and try sending it again
+    let _ = q.unwedge(&txn1).await;
```
bnjbvr, line 1493: instead of ignoring the result, please `.unwrap()` at the end to make sure it's an `Ok()` result.

crates/matrix-sdk/src/room/mod.rs
```diff
@@ -2957,6 +2962,54 @@ impl Room {
...
+    ) -> Result<()> {
+        let encryption = self.client().encryption();
+
+        for user_id in user_ids {
+            if let Some(user_identity) = encryption.get_user_identity(&user_id).await? {
+                user_identity.withdraw_verification().await?;
+            }
+        }
+
+        self.send_queue().unwedge(transaction_id).await?;
```
bnjbvr, line 2984: The API here is weird, because it's undocumented anywhere what the transaction_id is for, and that's it's one managed by the send queue. I'd remove this line from this function and the other, and instead have the callers at the FFI layer call the `unwedge` function and document the transaction id properties.

crates/matrix-sdk/src/room/mod.rs
```diff
@@ -2957,6 +2962,54 @@ impl Room {
...
+
+        for user_id in devices.keys() {
+            for device_id in &devices[user_id] {
+                if let Some(device) = encryption.get_device(user_id, device_id).await? {
+                    device.set_local_trust(LocalTrust::Verified).await?;
+                }
+            }
+        }
+
+        self.send_queue().unwedge(transaction_id).await?;
```
bnjbvr, line 3009: here too

crates/matrix-sdk/src/room/mod.rs
```diff
@@ -2957,6 +2962,54 @@ impl Room {
             .await?;
         Ok(())
     }
+
+    /// Remove verification requirements for the given users and
+    /// resend messages that failed to send because their identities were no
+    /// longer verified (in response to
+    /// `SessionRecipientCollectionError::VerifiedUserChangedIdentity`)
+    #[cfg(feature = "e2e-encryption")]
+    pub async fn withdraw_verification_and_resend(
```
bnjbvr, line 2971: Should these two methods live in the `Encryption` API object instead? They don't make any usage of the room, except for getting the send queue, which I argued below isn't the right place to do it.

bnjbvr: Also these two functions would need individual tests please 😁

crates/matrix-sdk/src/error.rs
```diff
@@ -379,6 +381,10 @@ pub enum Error {
     /// Backups are not enabled
     #[error("backups are not enabled")]
     BackupNotEnabled,
+
+    /// An error coming from the send queue subsystem.
+    #[error(transparent)]
+    SendQueueError(#[from] RoomSendQueueError),
```
bnjbvr, line 387: With my suggestion below, you can likely remove this…

bnjbvr (changes requested):
Good job on the send queue! I disagree with the other methods' API shape a little though, so let's rethink it through.

crates/matrix-sdk/src/room/mod.rs
stefanceriu: Just to make sure I get it straight: 
1) We would move `withdraw_verification` and `trust_devices` to Encryption
2) We would expose unwedge by txnId on the Room level
3) The final client would first trust via Client and then resend via Room

Does that sounds about right?

crates/matrix-sdk/src/room/mod.rs
stefanceriu: > these two functions would need individual tests

I reckon that if we split them up from the resending they would just proxy to the crypto crate where they're already tested.

crates/matrix-sdk/src/room/mod.rs
bnjbvr: > We would expose unwedge by txnId on the Room level

Do we need to, or can the caller themselves do `room.send_queue().unwedge(transaction_id)`? The idea here is that we should make it clear why it's ok to pass a transcation id to the send queue; it's because it's originating from the send queue too, and this information is super hidden at the moment, with the current API shape.

To be honest, I'm thinking all of this, the loop included above, could happen at the FFI layer, if it's just calling into other existing APIs without adding other value.

Consider the semantics carefully though: if any of the verification withdrawing fails, is it ok that 1. others have been successful, 2. we don't unwedge in that case?

stefanceriu: > Do we need to, or can the caller themselves do room.send_queue().unwedge(transaction_id)

Oh I see, I'm thinking of it from the FFI layer and you're talking SDK. Yeah that would be totally okay

> all of this, the loop included above, could happen at the FFI layer

Right yeah, I agree!

> Consider the semantics carefully though

I have, I think that bit's fine. Retrying to send it while some of the devices/users from the original failure haven't been resolved will result in the same error.

bindings/matrix-sdk-ffi/src/room.rs
```diff
@@ -737,6 +738,68 @@ impl Room {
         self.inner.send_queue().send(replacement_event).await?;
         Ok(())
     }
+
+    /// Remove verification requirements for the given users and
+    /// resend messages that failed to send because their identities were no
+    /// longer verified (in response to
+    /// `SessionRecipientCollectionError::VerifiedUserChangedIdentity`)
```
bnjbvr, line 745: please document that the `transaction_id` is one coming from the send queue

bindings/matrix-sdk-ffi/src/room.rs
```diff
@@ -737,6 +738,68 @@ impl Room {
...
+    pub async fn withdraw_verification_and_resend(
+        &self,
+        user_ids: Vec<String>,
+        transaction_id: String,
+    ) -> Result<(), ClientError> {
+        let transaction_id: OwnedTransactionId = transaction_id.into();
+
+        let parsing_result: Result<Vec<OwnedUserId>, _> =
+            user_ids.iter().map(UserId::parse).collect();
```
bnjbvr, line 754: One of these should work:

```suggestion
        let user_ids: Vec<OwnedUserId> = user_ids.iter().map(UserId::parse).collect()?;
```


```suggestion
        let user_ids: Vec<OwnedUserId> = user_ids.iter().map(UserId::parse).collect::<Result<_, _>>()?;
```

bindings/matrix-sdk-ffi/src/room.rs
```diff
@@ -737,6 +738,68 @@ impl Room {
...
+            Err(error) => return Err(ClientError::new(error)),
+        };
+
+        Ok(())
+    }
+
+    /// Set the local trust for the given devices to `LocalTrust::Ignored`
+    /// and resend messages that failed to send because said devices are
+    /// unverified (in response to
+    /// `SessionRecipientCollectionError::VerifiedUserHasUnsignedDevice`).
```
bnjbvr, line 777: Same comment about the transaction id here.

crates/matrix-sdk/src/error.rs
bnjbvr: Can this not be removed, now?

crates/matrix-sdk/src/send_queue.rs
```diff
@@ -673,6 +693,19 @@ impl QueueStorage {
...
+    /// Marks an event identified with the given transaction id as being now
+    /// unwedged and adds it back to the queue.
+    async fn mark_as_unwedged(
+        &self,
+        transaction_id: &TransactionId,
+    ) -> Result<(), RoomSendQueueStorageError> {
+        Ok(self
+            .client()?
+            .store()
+            .update_send_queue_event_status(&self.room_id, transaction_id, false)
```
bnjbvr, line 705: Hmm I'm wondering if this function should return a boolean, to indicate if the underlying item with such a transaction id was found, otherwise this might give a false sense that something happened, while it's not the case… (if one passes an invalid transaction id, for instance)

We don't have to do it in this PR, though.

bnjbvr (changes requested):
Getting closer!

crates/matrix-sdk/src/error.rs
stefanceriu: Sorry, I completely missed this in the previous pass, was literally working on removing it 👍

bindings/matrix-sdk-ffi/src/room.rs
stefanceriu: Oh sweet, the second one does! So I can specialise collect to avoid inference issues, that's the bit I was missing when I first tried to write it that way. Really cool 👍

crates/matrix-sdk/src/send_queue.rs
stefanceriu: Oh yeah, that would make sense but why not return an error directly?

bindings/matrix-sdk-ffi/src/error.rs
```diff
@@ -134,14 +134,14 @@ impl From<EventCacheError> for ClientError {
...
+    fn from(e: EditError) -> Self {
         Self::new(e)
     }
 }
 
-impl From<EditError> for ClientError {
-    fn from(e: EditError) -> Self {
+impl From<RoomSendQueueError> for ClientError {
+    fn from(e: RoomSendQueueError) -> Self {
         Self::new(e)
```
bnjbvr, line 145: :eyes: something fun is going on here…

crates/matrix-sdk/src/send_queue.rs
bnjbvr: It's a philosophical question: is it an error to misuse the API with a transaction id that's not handled by the send queue, or should we signal it nicely so that the user can try something else with that transaction id?

bnjbvr (approved):
lgtm, thanks

bindings/matrix-sdk-ffi/src/error.rs
stefanceriu: Yeah, I don't understand what its problem is.

crates/matrix-sdk/src/send_queue.rs
stefanceriu: I would passing in an invalid/unknown transaction id should be a developer error, api missuse. Giving back a boolean might be unclear to what actually went wrong.

PR #3911, opened 2024-08-29
timeline: allow all foci to handle local echoes

bnjbvr:
There's already the right logic to not *add* new timeline items to a non-live timeline, so adding a message etc. will not show it in a pinned / permalinked timeline. On the other hand, *updating* a timeline item (reaction/redaction/edit) will affect it, no matter the type. So we can handle local echoes of those types for those timeline items, such that they're displayed in real-time on all foci.

The added test would fail before the patch.


stefanceriu (approved):
Very nice! 👌

PR #3914, opened 2024-08-29
timeline: rename some concepts

bnjbvr:
`Inner` doesn't make sense. In this PR built on top of #3913, I propose the following:

- remove `Inner` from all the names
- rename `TimelineInner` to `TimelineController`. It's vague too, but if you squint your eyes the `room_data_provider` is actually a model, so that kind of fits the whole MVC story.

No functional changes.

### DIFF
--- a/crates/matrix-sdk-ui/src/timeline/mod.rs
+++ b/crates/matrix-sdk-ui/src/timeline/mod.rs
@@ -171,14 +171,10 @@ pub enum TimelineFocus {
     Live,
 
     /// Focus on a specific event, e.g. after clicking a permalink.
-    Event {
-        target: OwnedEventId,
-        num_context_events: u16,
-    },
+    Event { target: OwnedEventId, num_context_events: u16 },
 
-    PinnedEvents {
-        max_events_to_load: u16,
-    },
+    /// Only show pinned events.
+    PinnedEvents { max_events_to_load: u16 },
 }
 
 impl Timeline {


crates/matrix-sdk-ui/src/timeline/controller/mod.rs
```diff
@@ -107,7 +107,7 @@ enum TimelineFocusData<P: RoomDataProvider> {
 }
 
 #[derive(Clone, Debug)]
-pub(super) struct TimelineInner<P: RoomDataProvider = Room> {
+pub(super) struct TimelineController<P: RoomDataProvider = Room> {
```
Hywan, line 110: I don't like this naming, as it doesn't feel consistent with the rest of the SDK. `Controller` is really too generic and has no particular meaning. However, in this particular context, the `TimelineController` does a lot, it's really the `Timeline` implementation. I imagine the meaning for you is that it _controls_ the timeline. Well, all types control what it represents… I don't find a better name and I clearly understand why you don't like the `Inner` suffix.

I'm going to accept it because I don't have anything better to propose.

Hywan (approved):

crates/matrix-sdk-ui/src/timeline/controller/mod.rs
bnjbvr: Yeah, I think of it as the Controller in a MVC model — but even that is kind of wrong, as it should be more of a View Controller, really — and the model is scattered from multiple async data sources, the room provider, etc. so it's a bit unclear that there's a single model, and that MVC is a suitable metaphor for all of it.

Thanks, I'm happy to tweak the name later if we find a better one.

PR #3915, opened 2024-08-29
testing: enforce a `test_` prefix for tests

bnjbvr:
This will only apply to `async_test` functions, but I think this is a win:

1. for consistency within the codebase, since I've started doing so in many places,
2. because these function names will clearly identify these functions as tests, in the call tree interfaces, when rendered using the LSP show-callers/show-callees functionality.

### DIFF
--- a/testing/matrix-sdk-test-macros/src/lib.rs
+++ b/testing/matrix-sdk-test-macros/src/lib.rs
@@ -7,6 +7,11 @@ use syn::parse_macro_input;
 #[proc_macro_attribute]
 pub fn async_test(_attr: TokenStream, item: TokenStream) -> TokenStream {
     let fun = parse_macro_input!(item as syn::ItemFn);
+
+    if !fun.sig.ident.to_string().starts_with("test_") {
+        panic!("test function names must start with test_");
+    }
+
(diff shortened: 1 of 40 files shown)


testing/matrix-sdk-test-macros/src/lib.rs
```diff
@@ -7,6 +7,11 @@ use syn::parse_macro_input;
 #[proc_macro_attribute]
 pub fn async_test(_attr: TokenStream, item: TokenStream) -> TokenStream {
     let fun = parse_macro_input!(item as syn::ItemFn);
+
+    if !fun.sig.ident.to_string().starts_with("test_") {
```
stefanceriu, line 11: I will never get used to this twitchy convulsive way of reading abbreviations. Presumably it's `function.signature.identifier(?)`? who knows..

stefanceriu (approved):
🙌

testing/matrix-sdk-test-macros/src/lib.rs
bnjbvr: Yep, that's it! These are unfortunately quite common in the proc-macro's `syn` library…

PR #3918, opened 2024-08-30
base: Add media retention policy to EventCacheStore

zecakeh:
Allows to prevent the media cache from growing indefinitely by setting different limitations like size or last access time. I tried to select sensible defaults for those values.

Currently cleanups must be triggered manually, a future PR will add the possibility to run a regular cleanup task so they happen automatically.

### DIFF
--- a/crates/matrix-sdk-base/src/event_cache_store/mod.rs
+++ b/crates/matrix-sdk-base/src/event_cache_store/mod.rs
@@ -19,21 +19,25 @@
 //! into the event cache for the actual storage. By default this brings an
 //! in-memory store.
 
-use std::str::Utf8Error;
+use std::{str::Utf8Error, time::Duration};
 
 #[cfg(any(test, feature = "testing"))]
 #[macro_use]
 pub mod integration_tests;
 mod memory_store;
 mod traits;
+mod wrapper;
 
 pub use matrix_sdk_store_encryption::Error as StoreEncryptionError;
+use ruma::time::SystemTime;
+use serde::{Deserialize, Serialize};
 
 #[cfg(any(test, feature = "testing"))]
 pub use self::integration_tests::EventCacheStoreIntegrationTests;
 pub use self::{
     memory_store::MemoryStore,
     traits::{DynEventCacheStore, EventCacheStore, IntoEventCacheStore},
+    wrapper::EventCacheStoreWrapper,
 };
 
 /// Event cache store specific error type.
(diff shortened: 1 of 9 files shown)


stefanceriu:
Sounds awesome but it's a biggie so I think it's safer to pass it over to somebody more familiar with the SDK, congrats Benji! 😁

zecakeh:
Rebased on main to solve a conflict.

bnjbvr:
Hi, thanks for the PR, since this is a large commit it might take a bit of time to get to review it. I'll try my best.

bnjbvr (changes requested):
My main feedback is that I wouldn't want to have external users have to deal with a `EventCacheStoreWrapper`, but rather with the `EventCacheStore`; so `EventCacheStore` (the trait) would likely need to be renamed into something else, maybe `EventCacheStoreBackend` or `EventCacheStoreImpl`, or something else that makes this clear enough?

I haven't looked at the code in details, so forwarding review to somebody else.

zecakeh:
I know this is a big commit, so for reviewers I am going to reiterate what I said in the Matrix room to help reviewing part by part:

I would recommend starting in `matrix-sdk-base/event_cache_store` with `mod.rs` and `traits.rs`. This is where the new API is actually defined with docs to explain the available settings.

Then in any order:
- The (currently named) `EventCacheStoreWrapper` exposes a simpler public API to users of the SDK.
- The implementations of the memory store and the SQLite store can be reviewed separately.
- Check the integration tests (this is actually the biggest part by far!).

Hywan:
Before reviewing this PR, I'm wondering: don't we have an MSC about media retention policy? I see there is [MSC1763](https://github.com/matrix-org/matrix-spec-proposals/blob/matthew/msc1763/proposals/1763-configurable-retention-periods.md). I don't know the state of this MSC, but I think we should link both (a media is an event after all). I don't know if the `/sync` can tell a client that a message is too old and should be dropped (that's possibly not the case, I didn't read the MSC). Anyway, I think we should think more globally about data retention. Why? Because with this PR, this is client-specific. A user with 2 devices would see this information “synced” between its devices. Moreover, the media can be removed from the client, but would continue to live on the server: it's pretty easy to understand this as a _lie_ from the client. Worst: if a user has 2 devices with 2 different configuration, the user can be pretty confused to see the media removed on one device, but not on another one.

What do you think about this?

zecakeh:
This is on purpose a device-specific setting. The sole goal here is for the client or the user (if the client offers a setting for it) to be allowed to limit the size of the cache on the device.

This kind of setting can usually change from one device to the other, e.g. the limitations would be more strict on an phone where disk space is limited than on a computer with a lot of free space.

crates/matrix-sdk-base/src/event_cache_store/mod.rs
```diff
@@ -83,3 +87,162 @@ impl EventCacheStoreError {
...
+    ///
+    /// If it is set, media content bigger than the maximum size will not be
+    /// cached. If the maximum size changed after media content that exceeds the
+    /// new value was cached, the corresponding content will be removed
+    /// during a cleanup.
+    ///
+    /// Defaults to 20 MiB.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub max_file_size: Option<usize>,
+    /// The duration after which unaccessed media content is considered
```
Hywan, line 127:
```suggestion
    pub max_file_size: Option<usize>,

    /// The duration after which unaccessed media content is considered
```

crates/matrix-sdk-base/src/event_cache_store/mod.rs
```diff
@@ -83,3 +87,162 @@ impl EventCacheStoreError {
...
+    /// expired.
+    ///
+    /// If this is set, media content whose last access is older than this
+    /// duration will be removed from the media cache during a cleanup.
+    ///
+    /// Defaults to 30 days.
```
Hywan, line 133: I think 60 days is theoretically a better default for this value.

crates/matrix-sdk-base/src/event_cache_store/mod.rs
```diff
@@ -83,3 +87,162 @@ impl EventCacheStoreError {
...
+    /// bytes.
+    ///
+    /// This is the lowest value between `max_cache_size` and `max_file_size`.
+    pub fn computed_max_file_size(&self) -> Option<usize> {
+        match (self.max_cache_size, self.max_file_size) {
+            (None, None) => None,
+            (None, Some(size)) => Some(size),
+            (Some(size), None) => Some(size),
+            (Some(max_cache_size), Some(max_file_size)) => Some(max_cache_size.min(max_file_size)),
+        }
```
Hywan, line 201:
```suggestion
        self.max_cache_size.unwrap_or(usize::MAX).min(self.max_file_size.unwrap_or(usize::MAX))
        }
```

would allow to return a `usize` instead of `Option<usize>`, thoughts?

crates/matrix-sdk-base/src/event_cache_store/mod.rs
```diff
@@ -83,3 +87,162 @@ impl EventCacheStoreError {
 
 /// An `EventCacheStore` specific result type.
 pub type Result<T, E = EventCacheStoreError> = std::result::Result<T, E>;
+
+/// The retention policy for media content used by the `EventCacheStore`.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Serialize, Deserialize)]
+pub struct MediaRetentionPolicy {
```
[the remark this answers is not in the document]
Hywan: Also, this type and its implementation is quite long. It must be inside its own module.

Hywan, line 93: Can you add unit tests for this type please?

crates/matrix-sdk-base/src/event_cache_store/traits.rs
```diff
@@ -29,43 +29,81 @@ pub trait EventCacheStore: AsyncTraitDeps {
     /// The error type used by this event cache store.
     type Error: fmt::Debug + Into<EventCacheStoreError>;
 
-    /// Add a media file's content in the media store.
+    /// The retention policy set to cleanup the media cache.
+    async fn media_retention_policy(&self) -> Result<Option<MediaRetentionPolicy>, Self::Error>;
```
Hywan, line 33: This addition can go in a second commit:

- First commit: add `MediaRetentionPolicy`
- Second commit: use `MediaRetentionPolicy` along with tests.

crates/matrix-sdk-base/src/event_cache_store/wrapper.rs
```diff
@@ -0,0 +1,130 @@
...
+
+use super::{DynEventCacheStore, MediaRetentionPolicy, Result};
+use crate::media::MediaRequest;
+
+/// A wrapper around an [`EventCacheStore`] implementation to abstract common
+/// operations.
+///
+/// [`EventCacheStore`]: super::EventCacheStore
+#[derive(Debug, Clone)]
+pub struct EventCacheStoreWrapper {
```
Hywan, line 28: I understand why you've chosen this design.

I'm not comfortable with this though for several reasons, the major one being it introduces another type indirection for the `EventCacheStore` itself. To reach the `EventCacheStore`, the entry point is `EventCacheStoreWrapper`, then `DynEventCacheStore`, then `EventCacheStore`. If we include the `EventCacheStoreLock`, it introduces _another_ level of indirection.

Note that this is only for media management for the moment.

Possible solutions:

- Use standalone functions to be used by the multiple `EventCacheStore` implementations, liek we did for the cross-process lock and `MemoryStore`: https://github.com/matrix-org/matrix-rust-sdk/commit/94bd421a8db199fe6aedd7e12b6a1788059bda32
- Use a standalone type to manage media, like `MediaService`, which takes a `MediaRetentionPolicy`. The `EventCacheStore` would delegate the handling of media to `MediaService`, such as:

  ```rust
  pub async fn add_media_content(&self, request: &MediaRequest, content: Vec<u8>) -> Result<()> {
      self.media_service.add_media_content(&self, request).await
  }
  ```

That way we can test `MediaService` alone, it's a self-contained API, and it removes the need for another `EventCacheWrapper` type + type indirection.

Bonus: `MediaService` can be introduced inside its own commit, along with its tests. Another commit can use `MediaService` inside the `EventCacheStore` implementations.

What do you think?

crates/matrix-sdk-sqlite/src/utils.rs
```diff
@@ -187,27 +188,27 @@ impl SqliteAsyncConnExt for SqliteAsyncConn {
 }
 
 pub(crate) trait SqliteTransactionExt {
-    fn chunk_large_query_over<Query, Res>(
+    fn chunk_large_query_over<K, Query, Res>(
```
Hywan, line 191:
```suggestion
    fn chunk_large_query_over<Key, Query, Res>(
```

crates/matrix-sdk-sqlite/src/utils.rs
```diff
@@ -187,27 +188,27 @@ impl SqliteAsyncConnExt for SqliteAsyncConn {
...
     ) -> Result<Vec<Res>>
     where
         Res: Send + 'static,
-        Query: Fn(&Transaction<'_>, Vec<Key>) -> Result<Vec<Res>> + Send + 'static;
+        Query: Fn(&Transaction<'_>, Vec<K>) -> Result<Vec<Res>> + Send + 'static;
 }
 
 impl<'a> SqliteTransactionExt for Transaction<'a> {
-    fn chunk_large_query_over<Query, Res>(
+    fn chunk_large_query_over<K, Query, Res>(
```
Hywan, line 203:
```suggestion
    fn chunk_large_query_over<Key, Query, Res>(
```

Hywan (changes requested):
I like the general idea and the `MediaPolicy` type. I'm suggesting to rewrite the history to make it clearer. I'm also suggesting to introduce a new `MediaService` API instead of adding a new event cache indirection with `EventCacheStoreWrapper`. What do you think?

Sorry for the late review.

crates/matrix-sdk-base/src/event_cache_store/mod.rs
zecakeh: If this is `None`, there is no max file size, it's different from returning `usize::MAX`.

zecakeh: What I mean is that I don't see what we gain from returning a `usize` instead of an `Option<usize>`. If that's what you prefer, I'll change it.

crates/matrix-sdk-base/src/event_cache_store/wrapper.rs
zecakeh: How does the `MediaService` access the data from the backend? Are we supposed to add a new trait for that?

crates/matrix-sdk-base/src/event_cache_store/mod.rs
Hywan: I did miss the possibility to return “no max file size”, let's keep it as is.

crates/matrix-sdk-base/src/event_cache_store/wrapper.rs
zecakeh: I'm sorry but this is still confusing to me. From what you are saying, my understanding is that an `EventCacheStore` (or `EventCacheStoreMedia` with what you just said) implementation would look like:

```rust
impl EventCacheStore for MemoryStore {
  async fn add_media_content(&self, request: &MediaRequest, content: Vec<u8>) -> Result<()> {
    self.media_service.add_media_content(&self, request, content).await
  }

  // other methods…
}
```

Which means the `MediaService` would look like this:

```rust
pub struct MediaService {
  media_retention_policy: Arc<StdMutex<Option<MediaRetentionPolicy>>>,
  media_cleanup_guard: Arc<AsyncMutex<()>>,
}

impl MediaService {
  // other methods…

  async fn add_media_content(&self, cache: &impl InnerMediaApi, request: &MediaRequest, content: Vec<u8>) -> Result<()> {
    let media_retention_policy = self.media_retention_policy();

    // maybe do some pre-checks with the policy…

    cache.add_media_content_inner(media_retention_policy, request, content).await
  }
}
```

We need a trait for a private API that is only accessed by the `MediaService`, right? The `MediaService` cannot call the methods from the `EventCacheStore` trait because they are calling it?

Hywan: What I propose is:

```rust
trait EventCacheStoreMedia {
    fn add_media_content(…);
    fn remove_media_content(…);
}

trait EventCacheStore: EventCacheStoreMedia { … }

impl EventCacheStoreMedia for MediaService { … }

impl EventCacheStore for MemoryStore {
    fn add_media_content(…) {
        self.media_service.add_media_content(…);
    }

    …
}
```

something roughly like that.

zecakeh: But what API does the `MediaService` call in its implementation of `add_media_content`? Since we are storing/removing data from the backend, the `MediaService` needs a generic API to access it somehow.

Hywan: Yes. `MediaService<Store: EventCacheStore>`, something like this. Thoughts?

zecakeh:
Superceded by #4571

PR #3919, opened 2024-08-30
crypto: Log the received device keys on an encrypted olm message

richvdh:
MSC4147 adds a `device_keys` property to the plaintext content of encrypted to-device events, and support for picking this out was added in #3556/#3633.

Turns out it's quite handy to log this information.

The output of this is something like:

```
2024-08-30T14:40:27.061681Z DEBUG receive_sync_changes:receive_to_device_event{sender="@alice:example.org" event_type="m.room.encrypted" message_id="01J6HVNZXBPF860R0TBH3H75KZ"}:handle_decrypted_to_device_event{sender_key="curve25519:CpEg1cKt2OiASokDD23L9/buqT7LAkcUEVaFXaBSCFU" event_type="m.room_key"}: matrix_sdk_crypto::machine: Received a decrypted to-device event sender_device_keys=(Some("curve25519:CpEg1cKt2OiASokDD23L9/buqT7LAkcUEVaFXaBSCFU"), Some("ed25519:aEMaB1+jf3tzIRdyNzI514m1czCra9W23JwYXho27iw"))
```

### DIFF
--- a/crates/matrix-sdk-crypto/src/machine/mod.rs
+++ b/crates/matrix-sdk-crypto/src/machine/mod.rs
@@ -1109,7 +1109,11 @@ impl OlmMachine {
         decrypted: &mut OlmDecryptionInfo,
         changes: &mut Changes,
     ) -> OlmResult<()> {
-        debug!("Received a decrypted to-device event");
+        debug!(
+            sender_device_keys =
+                ?decrypted.result.event.sender_device_keys().map(|k| (k.curve25519_key(), k.ed25519_key())).unwrap_or((None, None)),
+            "Received a decrypted to-device event",
+        );
 
         match &*decrypted.result.event {
             AnyDecryptedOlmEvent::RoomKey(e) => {
--- a/crates/matrix-sdk-crypto/src/types/events/olm_v1.rs
+++ b/crates/matrix-sdk-crypto/src/types/events/olm_v1.rs
@@ -148,6 +148,17 @@ impl AnyDecryptedOlmEvent {
             AnyDecryptedOlmEvent::Dummy(e) => e.content.event_type(),
         }
     }
+
+    /// The sender's device keys, if supplied in the message as per MSC4147
+    pub fn sender_device_keys(&self) -> Option<&DeviceKeys> {
+        match self {
+            AnyDecryptedOlmEvent::Custom(_) => None,
+            AnyDecryptedOlmEvent::RoomKey(e) => e.device_keys.as_ref(),
+            AnyDecryptedOlmEvent::ForwardedRoomKey(e) => e.device_keys.as_ref(),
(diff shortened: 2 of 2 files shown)


crates/matrix-sdk-crypto/src/machine/mod.rs
```diff
@@ -1109,7 +1109,11 @@ impl OlmMachine {
         decrypted: &mut OlmDecryptionInfo,
         changes: &mut Changes,
     ) -> OlmResult<()> {
-        debug!("Received a decrypted to-device event");
+        debug!(
+            sender_device_keys =
+                ?decrypted.result.event.sender_device_keys().map(|k| (k.curve25519_key(), k.ed25519_key())).unwrap_or((None, None)),
```
bnjbvr, line 1114: nit: you can probably [`Option::unzip`](https://doc.rust-lang.org/std/option/enum.Option.html#method.unzip) somewhere around here

Maybe:
```suggestion
                ?decrypted.result.event.sender_device_keys().map(|k| (k.curve25519_key(), k.ed25519_key())).unzip(),
```

bnjbvr (approved):
Approving for the Rust side, still needs some crypto review to validate it's OK to include this info in rageshakes/logs.

crates/matrix-sdk-crypto/src/machine/mod.rs
richvdh: Erm, not sure that works.

The problem is that `k.curve25519_key()` and `k.ed25519_key()` both return an Option, so the result of my `map` is  `Option<(Option<_>, Option<_>)>`. Assuming the keys are populated, when I `unzip` that, I get `(Some(Some(k1)), Some(Some(k2))`.

bnjbvr: I see, let's keep it like this then!

poljar: Probably meant:

```rust
decrypted.result.event
    .sender_device_keys()
    .and_then(|k| k.curve25519_key().zip(k.ed25519_key()))
```

richvdh: > Probably meant:
> 
> ```rust
> decrypted.result.event
>     .sender_device_keys()
>     .and_then(|k| k.curve25519_key().zip(k.ed25519_key()))
> ```

well, that might be nice, but AFAICT it doesn't handle having one key and not the other? So it would either be `Some(k1, k2)` or `None`?

I guess it's very unlikely to have just one key, but I think I'd rather allow for it

andybalaam:
These keys are safe to log because they are public keys: anyone who shares a room with the user can find them proactively by making /keys/query request.

This seems potentially dangerous because they are included in the encrypted content of the event, but the MSC mentions that they are public information and could just as easily have been included in the cleartext.

andybalaam (approved):

PR #3921, opened 2024-08-30
crypto: expose `SenderDataFinder::find_using_device_data`

richvdh:
Turns out that most of the places we call `find_using_device_keys`, we already have a DeviceData. So we might as well pass that in directly, rather than extracting the device keys and then rebuilding a DeviceData.


andybalaam (approved):
Great, thanks!

PR #3923, opened 2024-09-02
fix(sdk): Don't subscribe to already subscribed rooms in Sliding Sync

Hywan:
When subscribing to an already subscribed room, the subscription state was reset, the subscription was resent by cancelling the in-flight request. All this is useless and can create a feeling of lag.

This patch checks if a room is subscribed first. If it is, nothing happens. If it is not, the room subscription is created, and the in-flight request is cancelled.

Tests are updated to reflect this new change.

Note: room subscription settings are not taken into account in this “presence” pattern. It means that if we subscribe to an already subscribed room but with different settings, nothing will happen. The server will ignore the new settings anyway for the moment. It's because room subscriptions are sticky. It's a normal behaviour here.

### DIFF
--- a/crates/matrix-sdk/src/sliding_sync/mod.rs
+++ b/crates/matrix-sdk/src/sliding_sync/mod.rs
@@ -25,7 +25,7 @@ mod sticky_parameters;
 mod utils;
 
 use std::{
-    collections::{BTreeMap, HashSet},
+    collections::{btree_map::Entry, BTreeMap, HashSet},
     fmt::Debug,
     future::Future,
     sync::{Arc, RwLock as StdRwLock},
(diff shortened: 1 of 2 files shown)


crates/matrix-sdk/src/sliding_sync/mod.rs
```diff
@@ -154,20 +154,25 @@ impl SlidingSync {
         let mut sticky = self.inner.sticky.write().unwrap();
         let room_subscriptions = &mut sticky.data_mut().room_subscriptions;
 
+        let mut skip_sync_loop = false;
+
         for room_id in room_ids {
-            if let Some(room) = self.inner.client.get_room(room_id) {
-                room.mark_members_missing();
+            if let Entry::Vacant(entry) = room_subscriptions.entry((*room_id).to_owned()) {
```
poljar, line 160: I would probably add a comment explaining what is going on here, from the PR description it is clear, but that might be hard to find again as time goes on.

poljar (approved):
Makes sense, left a nit.

crates/matrix-sdk/src/sliding_sync/mod.rs
Hywan: Indeed. I've improved to public doc, I've added an inline doc, and I've rebased the entire change.

PR #3925, opened 2024-09-02
fix(sdk): Sliding sync discovers the proxy on the server, not the homeserver

Hywan:
The `.well-known` file is located on the server, not the homeserver. This patch fixes that along with the associated tests.

### DIFF
--- a/crates/matrix-sdk/src/client/mod.rs
+++ b/crates/matrix-sdk/src/client/mod.rs
@@ -474,7 +474,7 @@ impl Client {
     /// The server used by the client.
     ///
     /// See `Self::server` to learn more.
-    fn server(&self) -> Option<&Url> {
+    pub fn server(&self) -> Option<&Url> {
         self.inner.server.as_ref()
     }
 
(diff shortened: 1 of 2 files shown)


crates/matrix-sdk/src/client/mod.rs
```diff
@@ -474,7 +474,7 @@ impl Client {
     /// The server used by the client.
     ///
     /// See `Self::server` to learn more.
-    fn server(&self) -> Option<&Url> {
+    pub fn server(&self) -> Option<&Url> {
```
bnjbvr, line 477: Is it intended to make it `pub`? Looks like `pub(crate)` would have been sufficient, but I can see the appeal of making this public for users too.

bnjbvr (approved):
lgtm

crates/matrix-sdk/src/client/mod.rs
Hywan: `pub(crate)` would be fine for this patch, but I believe it can be useful for someone else.

PR #3926, opened 2024-09-02
ffi: provide the `ClientDelegate` when creating a `ClientBuilder`

bnjbvr:
So it is always there, and set at the same time as the `Client`.

This would avoid a few race conditions app-side, with respect to getting updates about sessions: it's not possible anymore to get an update while there wasn't any client delegate set yet.

It might fix a few OIDC issues when creating a Client.


Hywan (approved):
LGTM

bnjbvr:
Closing, @stefanceriu and @jmartinesp said this would require more changes, and this was just an experiment to see if this could work simply. Let's put it aside and get back to it later.

PR #3928, opened 2024-09-02
sqlite: Update last access time first to force write transaction

zecakeh:
Avoids errors when the read transaction tries to upgrade to a write transaction (see [SQLite docs](https://sqlite.org/lang_transaction.html#read_transactions_versus_write_transactions) for details).

### DIFF
--- a/crates/matrix-sdk-sqlite/src/event_cache_store.rs
+++ b/crates/matrix-sdk-sqlite/src/event_cache_store.rs
@@ -162,25 +162,20 @@ impl EventCacheStore for SqliteEventCacheStore {
         let conn = self.acquire().await?;
         let data = conn
             .with_transaction::<_, rusqlite::Error, _>(move |txn| {
-                let Some(media) = txn
-                    .query_row::<Vec<u8>, _, _>(
-                        "SELECT data FROM media WHERE uri = ? AND format = ?",
-                        (&uri, &format),
-                        |row| row.get(0),
-                    )
-                    .optional()?
-                else {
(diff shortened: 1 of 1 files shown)


crates/matrix-sdk-sqlite/src/event_cache_store.rs
```diff
@@ -162,25 +162,20 @@ impl EventCacheStore for SqliteEventCacheStore {
...
-                        (&uri, &format),
-                        |row| row.get(0),
-                    )
-                    .optional()?
-                else {
-                    return Ok(None);
-                };
-
                 // Update the last access.
+                // We need to do this first so the transaction is in write mode right away.
```
Hywan, line 166: I think it's worst mentionning https://sqlite.org/lang_transaction.html#read_transactions_versus_write_transactions here, what do you think?

Hywan (commented):
I don't know how easy it is to test such a feature but the patch looks correct according the SQLite doc.

crates/matrix-sdk-sqlite/src/event_cache_store.rs
zecakeh: Sure

zecakeh: Added.

zecakeh:
I'd say it is not easy, it should only happen when the database is really busy.

Hywan (approved):
@stefanceriu has confirmed the problem is fixed. It follows our expectations. Thanks!

PR #3929, opened 2024-09-02
ffi: Add Python bindings UniFFI config

MatMaul:
Those changes allow to build functional Python wheel of the crypto FFI layer using [maturin](https://github.com/PyO3/maturin).
Something is still not right for the full SDK FFI, I am looking into it but it should be a change on maturin side.

Currently depends on a [change](https://github.com/PyO3/maturin/pull/2208) in maturin.

Signed-off-by: Mathieu Velten <mathieu@velten.xyz>

### DIFF
--- a/bindings/matrix-sdk-crypto-ffi/Cargo.toml
+++ b/bindings/matrix-sdk-crypto-ffi/Cargo.toml
@@ -10,10 +10,11 @@ license = "Apache-2.0"
 publish = false
 
 [lib]
+name = "matrix_sdk_crypto_ffi"
 crate-type = ["cdylib", "staticlib"]
 
 [[bin]]
-name = "matrix_sdk_crypto_ffi"
+name = "uniffi-bindgen"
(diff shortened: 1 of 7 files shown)


Hywan:
Hello and thanks for your contribution.

While I acknowledge it can be useful to support Python, we don't want to support it here as it means we will also be responsible to publish wheels/packages to pypi etc., which we won't.

I'm sorry but we don't want to provide official Python support for the moment.

MatMaul:
I understand you don't want too much support burden and I am not asking to provide official support or wheels, but can we still work on basic support like this PR ? By adding an explicit disclaimer (no official support, no build, experimental) in the README for example.

FYI I am trying to solve [libolm deprecation for the content scanner.](https://github.com/element-hq/matrix-content-scanner-python/issues/64)

MatMaul:
Let's close for now, thanks for the feedback.

PR #3933, opened 2024-09-03
crypto: fix OIDC cross-signing reset flows after backend authorization failure response change

stefanceriu:
What it says on the tin.

### DIFF
--- a/crates/matrix-sdk/src/encryption/mod.rs
+++ b/crates/matrix-sdk/src/encryption/mod.rs
@@ -36,7 +36,6 @@ use matrix_sdk_base::crypto::{
 use matrix_sdk_common::executor::spawn;
 use ruma::{
     api::client::{
-        error::{ErrorBody, ErrorKind},
         keys::{
             get_keys, upload_keys, upload_signatures::v3::Request as UploadSignaturesRequest,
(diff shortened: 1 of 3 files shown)


crates/matrix-sdk/src/encryption/mod.rs
```diff
@@ -330,45 +337,43 @@ impl CrossSigningResetAuthType {
...
-                .oidc()
-                .account_management_url(Some(AccountManagementActionFull::CrossSigningReset))
-                .await?;
+        Ok(OidcCrossSigningResetInfo { approval_url: parameters.reset.url })
+    }
+}
 
-            #[cfg(not(feature = "experimental-oidc"))]
-            let approval_url = None;
+#[cfg(feature = "experimental-oidc")]
```
poljar, line 360: Docs go above the attributes.

poljar (approved):
Left some nits, are you absolutely sure that we don't need to support the old way as well?

Anyways looks good and can confirm that it works.

crates/matrix-sdk/src/encryption/mod.rs
stefanceriu: Upsie, my bad, thank you 👍

PR #3934, opened 2024-09-03
chore(ui): Rename variables

Hywan:
This is not a timestamp but a regular stamp. Make it clear with the variable names.


bnjbvr (changes requested):
Can you break this commit down into smaller chunks, please? it's so hard to review.

bnjbvr (approved):
j/k

PR #3939, opened 2024-09-04
sqlite: Bump sqlite crates

zecakeh:


zecakeh:
Not really, no. I just noticed these versions have been out for a while.

PR #3940, opened 2024-09-04
sdk: tiny refactorings

bnjbvr:
These were found while looking at rageshakes involving missing initial local echoes, but I didn't get my head around it this time. Oh well.

### DIFF
--- a/crates/matrix-sdk/src/client/builder/mod.rs
+++ b/crates/matrix-sdk/src/client/builder/mod.rs
@@ -443,7 +443,9 @@ impl ClientBuilder {
             let mut client =
                 BaseClient::with_store_config(build_store_config(self.store_config).await?);
             #[cfg(feature = "e2e-encryption")]
-            (client.room_key_recipient_strategy = self.room_key_recipient_strategy.clone());
+            {
+                client.room_key_recipient_strategy = self.room_key_recipient_strategy;
+            }
             client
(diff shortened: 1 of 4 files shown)


crates/matrix-sdk/src/client/builder/mod.rs
```diff
@@ -443,7 +443,9 @@ impl ClientBuilder {
             let mut client =
                 BaseClient::with_store_config(build_store_config(self.store_config).await?);
             #[cfg(feature = "e2e-encryption")]
-            (client.room_key_recipient_strategy = self.room_key_recipient_strategy.clone());
+            {
```
poljar, line 446: Are any of those parenthesis needed? I think no? Or are you keeping them for future e2e-enrytpion specific additions?

poljar (approved):

crates/matrix-sdk/src/client/builder/mod.rs
bnjbvr: I did remove the unneeded parenthesis indeed. They were used as a hack because a cfg guard cannot be applied to a single statement, it has to apply to an expression; I found using braces clearer to express this intent, instead of parenthesis.

poljar: Oh, TIL.

PR #3942, opened 2024-09-04
WIP: make `EventTimelineItem` a record type for UniFFI

jmartinesp:

Signed-off-by: 
…nt-ffi-object-to-record…nt-ffi-object-to-record


jmartinesp:
I'm not sure if we should have both `edit` and `edit_by_id` or just replace everything with the `_by_id` variants.

As for the `TimelineItemEventId` use instead of the `UniqueId`, I already tried using `UniqueId` for everything and run quite quickly into issues when working with drafts, so I'm keeping this for now, although if we end up fixing the draft usage issues we could replace one with the other quite easily.

jmartinesp:
On another topic, how should we handle complement crypto (expectedly) failing after these FFI changes?

bnjbvr:
> On another topic, how should we handle complement crypto (expectedly) failing after these FFI changes?

The way to go would be to open a PR against complement-crypto that updates the API uses, and if you name the PR branch on the complement-crypto side the same as the branch here, then CI is going to pick it up automatically, and then we need to coordinate merging both (if I remember correctly).

crates/matrix-sdk-ui/src/timeline/event_item/content/message.rs
```diff
@@ -139,6 +144,11 @@ impl Message {
         self.thread_root.is_some()
     }
 
+    /// Get the [`OwnedEventId`] of the root event of a thread if it exists.
+    pub fn thread_root(&self) -> Option<OwnedEventId> {
+        self.thread_root.clone()
```
bnjbvr, line 149: A slightly better API would consist in returning `Option<&EventId>` and let the caller decide whether they want to clone or not.

crates/matrix-sdk-ui/src/timeline/event_item/content/message.rs
```diff
@@ -117,6 +117,11 @@ impl Message {
         Self { msgtype, in_reply_to, thread_root, edited, mentions }
     }
 
+    /// Create a forwarded message from a [`MessageType`].
+    pub fn forwarded(msgtype: MessageType) -> Self {
+        Self { msgtype, in_reply_to: None, thread_root: None, edited: false, mentions: None }
+    }
```
bnjbvr, line 123: What's a "forwarded message"? Why does it lose all the information about the message? Can you clarify all of this in the docs please?

Also, since this looks super specific to ElementX, would it make sense to do this processing there instead?

bindings/matrix-sdk-ffi/src/timeline/mod.rs
```diff
@@ -600,16 +608,26 @@ impl Timeline {
...
+                .inner
+                .redact(&event, reason.as_deref())
+                .await
+                .map_err(|err| anyhow::anyhow!(err))?;
 
-        Ok(removed)
+            Ok(removed)
+        } else {
+            Ok(false)
+        }
```
bnjbvr: Why the changes here? What's an "id"? Is it an event id? a transaction id? a timeline item id?

bindings/matrix-sdk-ffi/src/timeline/mod.rs
```diff
@@ -486,13 +488,19 @@ impl Timeline {
...
+            let transaction_id: OwnedTransactionId = id.into();
+            self.inner.local_item_by_transaction_id(&transaction_id).await
+        };
+        if let Some(event) = event {
+            let new_content: SdkEditedContent = new_content.try_into()?;
+            self.inner.edit(&event, new_content).await.map_err(ClientError::from)
+        } else {
+            Ok(false)
+        }
     }
```
bnjbvr: same questions apply here

bindings/matrix-sdk-ffi/src/timeline/content.rs
```diff
@@ -74,45 +69,74 @@ impl TimelineItemContent {
...
+pub struct MessageContent {
+    pub msg_type: MessageType,
+    pub body: String,
+    pub in_reply_to: Option<Arc<InReplyToDetails>>,
+    pub thread_root: Option<String>,
+    pub is_edited: bool,
+    pub mentions: Option<Mentions>,
+}
+
+impl From<&matrix_sdk_ui::timeline::Message> for MessageContent {
```
bnjbvr, line 113: out of curiosity, could these be `From<Message> for MessageContent` instead? I'd suspect we could avoid lots of cloning in that case, and that might be sweet :)

bindings/matrix-sdk-ffi/src/timeline/content.rs
```diff
@@ -16,45 +16,40 @@ use std::{collections::HashMap, sync::Arc};
...
-use crate::ruma::{ImageInfo, MessageType, PollKind};
+use crate::ruma::{ImageInfo, Mentions, MessageType, PollKind};
 
-#[derive(Clone, uniffi::Object)]
-pub struct TimelineItemContent(pub(crate) matrix_sdk_ui::timeline::TimelineItemContent);
-
-#[uniffi::export]
-impl TimelineItemContent {
-    pub fn kind(&self) -> TimelineItemContentKind {
+impl From<&matrix_sdk_ui::timeline::TimelineItemContent> for TimelineItemContent {
```
bnjbvr, line 24: ditto here

bindings/matrix-sdk-ffi/src/lib.rs
```diff
@@ -1,6 +1,6 @@
 // TODO: target-os conditional would be good.
 
-#![allow(unused_qualifications, clippy::new_without_default)]
+#![allow(unused_qualifications, clippy::new_without_default, unused_macros)]
```
bnjbvr, line 3: Why is this needed?

bnjbvr (commented):
Just looked at the first commit between two talks, sorry ^^

bindings/matrix-sdk-ffi/src/lib.rs
jmartinesp: The macro isn't used anymore. We can remove it, but it's something Jonas added long ago and I wasn't sure if it could be useful in the future.

bindings/matrix-sdk-ffi/src/timeline/mod.rs
jmartinesp: > Why the changes here?

The previous `ffi::EventTimelineItem` implementation used to have a `ui::EventTimelineItem` inside it which we could use for stuff like cloning the contents, getting a send handle, etc. But that's no longer the case since `EventTimelineItem` is now a record type. What we use instead is an id.

> What's an "id"? Is it an event id? a transaction id?

In this case, it can be any of those two (and I need to add docs for that). Since we don't really have a wrapper for the event/transaction/unique ids in the FFI layer (for performance reasons?), I can think of several options when we can to be able to pass both event ids and transaction ids:

- Use this strategy, parse the id once received, I chose this one..
- Create those wrappers, although I don't know how that could affect both performance and the existing APIs.
- Create separate `redact_event_by_event_id`, `redact_event_by_transaction_id` fns.
- Use the `UniqueId`: I already tried this in the past, but it has some issues with the current APIs.

bindings/matrix-sdk-ffi/src/timeline/content.rs
jmartinesp: I don't think so unless I missed something important: either we clone a couple of values here or we need to clone the whole parent struct, or modify some other `From` and `TryFrom` implementations (like `impl TryFrom<MessageType> for RumaMessageType`) and clone other values there.

crates/matrix-sdk-ui/src/timeline/event_item/content/message.rs
jmartinesp: I think I implemented this a some kind of workaround and it's no longer needed, I don't even see it referenced in my test branch for EXA. I'm deleting it.

bindings/matrix-sdk-ffi/src/lib.rs
jmartinesp: https://github.com/matrix-org/matrix-rust-sdk/pull/3942/commits/7711a85e04bf598330587bc21b6379aeadc7c0cb

crates/matrix-sdk-ui/src/timeline/event_item/content/message.rs
jmartinesp: https://github.com/matrix-org/matrix-rust-sdk/pull/3942/commits/4b08774cb0142ea3553a7b1ab523ea1ade8eaa07

crates/matrix-sdk-ui/src/timeline/event_item/content/message.rs
jmartinesp: https://github.com/matrix-org/matrix-rust-sdk/pull/3942/commits/4b08774cb0142ea3553a7b1ab523ea1ade8eaa07

poljar: Was this perhaps the feature where you re-send a message in another room, just like copy pasting would do.

crates/matrix-sdk-ui/src/timeline/event_item/mod.rs
```diff
@@ -99,6 +99,22 @@ pub enum TimelineEventItemId {
...
+
+impl From<&str> for TimelineEventItemId {
+    fn from(value: &str) -> Self {
+        if let Ok(event_id) = EventId::parse(value) {
+            TimelineEventItemId::EventId(event_id)
+        } else {
+            TimelineEventItemId::TransactionId(value.into())
+        }
+    }
+}
```
Hywan, line 116: :+1:

I think it should be in a separate PR though.

crates/matrix-sdk-ui/src/timeline/mod.rs
```diff
@@ -455,6 +466,21 @@ impl Timeline {
         Some(found.clone())
     }
 
+    /// Edit an event given its [`TimelineEventItemId`] and some new content.
+    ///
+    /// See [`Self::edit`] for more information.
+    pub async fn edit_by_id(
```
Hywan, line 472: Can we extract this new method in a separate PR?

crates/matrix-sdk-ui/src/timeline/mod.rs
```diff
@@ -568,6 +594,38 @@ impl Timeline {
         SendAttachment::new(self, path.into(), mime_type, config)
     }
 
+    /// Redact an event given its [`TimelineEventItemId`] and an optional
+    /// reason.
+    ///
+    /// See [`Self::redact`] for more info.
+    pub async fn redact_by_id(
```
Hywan, line 601: Can we extract this new method in a separate PR?

Hywan (commented):
That's great, well done! I think we can extract some features as separate PRs though, in order to clarify the patches and makes the reviews easier. Overall, it's well done, thank you.

bindings/matrix-sdk-ffi/src/timeline/mod.rs
```diff
@@ -596,20 +593,13 @@ impl Timeline {
     /// being sent already. If the event was a remote event, then it will be
     /// redacted by sending a redaction request to the server.
     ///
-    /// Returns whether the redaction did happen. It can only return false for
-    /// local events that are being processed.
+    /// Will return an error if the event couldn't be redacted.
     pub async fn redact_event(
         &self,
-        item: Arc<EventTimelineItem>,
+        id: String,
```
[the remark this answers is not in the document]
jmartinesp: See https://github.com/matrix-org/matrix-rust-sdk/pull/3942#discussion_r1768852442.

It can be both the event id or a transaction id. Maybe I should rename the function to `redact_by_event_id` and `redact_by_transaction_id`, or use some wrapper. I listed the possible implementations I thought about in that comment, but I don't really know which one is best.

bindings/matrix-sdk-ffi/src/timeline/mod.rs
```diff
@@ -1021,100 +1019,73 @@ impl From<SdkShieldState> for ShieldState {
...
-    pub fn debug_info(&self) -> EventTimelineItemDebugInfo {
-        EventTimelineItemDebugInfo {
-            model: format!("{:#?}", self.0),
-            original_json: self.0.original_json().map(|raw| raw.json().get().to_owned()),
-            latest_edit_json: self.0.latest_edit_json().map(|raw| raw.json().get().to_owned()),
+            .collect();
+        let debug_info = EventTimelineItemDebugInfo {
+            model: format!("{:#?}", value),
+            original_json: value.original_json().map(|raw| raw.json().get().to_owned()),
+            latest_edit_json: value.latest_edit_json().map(|raw| raw.json().get().to_owned()),
```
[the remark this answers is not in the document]
jmartinesp: I *think* I may be able to create some `DebugInfoProvider` FFI object and add it as part of the record type, so you'd do `event.debugInfoProvider.get()` instead.

As for benchmarking, not with a real benchmark, I just made some measurements with the same timeline events with the code in `main` and this one, and even when having to parse the extra debug info the parsing was always faster with this branch than with `main`.

crates/matrix-sdk-ui/src/timeline/mod.rs
jmartinesp: Sure, I just added it because I wanted to move the logic down a layer where it could be tested, as I think @bnjbvr suggested in a chat room.

bindings/matrix-sdk-ffi/src/timeline/content.rs
jmartinesp: I must be doing something wrong, because I end up with more cloning by removing the references in the `From` implementations, not less (see abe00c30709d294b1c0851f69d9e272f35a23a1f): most of these clones like `value.msgtype().clone().into()` happen because the functions have a signature like: `fn msgtype() -> &MessageType`, and I can't just use `(*value.msgtype()).into()` because I'd move the `value` variable.

crates/matrix-sdk-ui/src/timeline/mod.rs
jmartinesp: I'm not sure about this one though, because at the FFI layer we no longer have a `sdk_ui::EventTimelineItem` we can pass to the existing `fn redact`. 

We *could* get it by passing an `ffi::EventTimelineItem`, checking its event and transaction ids, using them to look for the event in the timeline and then passing it down to the existing function. If we're alright with that implementation this one can be delayed into a next one, but it means we'll have to adapt the client's API twice instead of just once.

If we don't do this separate change we'd end up doing this:

1. In the client, call `val event = timeline.getTimelineItemByEventId(eventId)` or `val event = timeline.getTimelineItemByTransactionId(transactionId)`.
2. This looks for the event in the timeline, transforms it ands passes it back through the FFI layer.
3. We then call `timeline.redact(event)` in the client.
4. This will get the event id or transaction id from the FFI record, search for the sdk_ui event again, and *then* do the operation.

This is not wrong, and it would still use the existing API, but it's definitely not a good implementation IMO.

bindings/matrix-sdk-ffi/src/timeline/mod.rs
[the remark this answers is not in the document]
jmartinesp: https://github.com/matrix-org/matrix-rust-sdk/pull/3942/commits/4344be6076d04f4e6aae008ee68fd2605be9e1ac

bindings/matrix-sdk-ffi/src/room_list.rs
```diff
@@ -702,8 +702,8 @@ impl RoomListItem {
         self.inner.is_encrypted().await.unwrap_or(false)
     }
 
-    async fn latest_event(&self) -> Option<Arc<EventTimelineItem>> {
-        self.inner.latest_event().await.map(EventTimelineItem).map(Arc::new)
+    async fn latest_event(&self) -> Option<EventTimelineItem> {
+        self.inner.latest_event().await.map(|e| e.into())
```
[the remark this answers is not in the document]
jmartinesp: Done in https://github.com/matrix-org/matrix-rust-sdk/pull/3942/commits/4344be6076d04f4e6aae008ee68fd2605be9e1ac

bindings/matrix-sdk-ffi/src/timeline/mod.rs
Hywan: Or name the argument `event_or_transaction_id`?

jmartinesp: That I can do!

crates/matrix-sdk-ui/src/timeline/mod.rs
poljar: I think he meant that we would first add this method in a separate PR, without the removal of the stuff from the FFI layer, then this PR would depend on the PR introducing this method.

But since this PR has been open for quite a while and this is a fairly minor point, let's not split it out at this point in time anymore.

crates/matrix-sdk-ui/src/timeline/mod.rs
```diff
@@ -568,6 +594,38 @@ impl Timeline {
...
+        &self,
+        id: &TimelineEventItemId,
+        reason: Option<&str>,
+    ) -> Result<(), Error> {
+        match id {
+            TimelineEventItemId::TransactionId(transaction_id) => {
+                let Some(event) = self.item_by_transaction_id(transaction_id).await else {
```
poljar, line 608: I think that this reads easier, if we use a `match` and list all the possible branches using the match:

```rust
TimelineEventItemId::TransactionId(transaction_id) => {
    let item = self.item_by_transaction_id(transaction_id).await;
    
    match item.as_ref().map(|i| i.handle()) {
        Some(TimelineItemHandle::Local(handle)) => {
            handle.abort().await.map_err(RoomSendQueueError::StorageError)?;
                Ok(())
            }
            Some(TimelineItemHandle::Remote(_)) => {
                // Log an error here instead of a panic?
                panic!("If the item is local, this should never happen");
            }
            None => Err(Error::RedactError(RedactError::LocalEventNotFound(
                transaction_id.to_owned(),
            ))),
        }
    }
TimelineEventItemId::EventId(event_id) => {
    self.room()
        .redact(event_id, reason, None)
        .await
        .map_err(|e| Error::RedactError(RedactError::HttpError(e)))?;
    
    Ok(())
}
```

Also, that panic scares me a bit, though I understand that it can't currently happen.

poljar (approved):
Left a couple of nits and questions, but I think that this can go in after addressing those.

bindings/matrix-sdk-ffi/src/timeline/mod.rs
```diff
@@ -1021,100 +1029,69 @@ impl From<SdkShieldState> for ShieldState {
...
-    }
-
-    pub fn timestamp(&self) -> u64 {
-        self.0.timestamp().0.into()
-    }
+#[derive(Clone, uniffi::Record)]
+pub struct EventTimelineItem {
+    is_local: bool,
+    is_remote: bool,
+    transaction_id: Option<String>,
```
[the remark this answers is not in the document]
jmartinesp: I actually had implemented a `EventOrTransactionId` enum before, but I ended up removing it when trying `UniqueId` instead. I'll add it back.

bindings/matrix-sdk-ffi/src/timeline/mod.rs
```diff
@@ -1021,100 +1029,69 @@ impl From<SdkShieldState> for ShieldState {
...
+            is_editable: value.is_editable(),
+            content: value.content().clone().into(),
+            timestamp: value.timestamp().0.into(),
+            reactions,
+            debug_info_provider,
+            local_send_state: value.send_state().map(|s| s.into()),
+            read_receipts,
+            origin: value.origin(),
+            can_be_replied_to: value.can_be_replied_to(),
+            message_shield: value.get_shield(false).map(Into::into),
```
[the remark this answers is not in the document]
jmartinesp: Maybe I can create another provider here, as I did for the debug info. I wonder if it's better to wrap the ui `EventTimelineItem` into an `Arc` instead of cloning them so we can have it in both structs.

crates/matrix-sdk-ui/src/timeline/mod.rs
jmartinesp: I can't just log that, I'd have to return an error.

poljar: Or an `Ok(())`, no?

PR #3946, opened 2024-09-04
timeline: don't add a read marker after one's own events

bnjbvr:
Pet peeve of mine.

### DIFF
--- a/crates/matrix-sdk-ui/src/timeline/tests/virt.rs
+++ b/crates/matrix-sdk-ui/src/timeline/tests/virt.rs
@@ -25,7 +25,7 @@ use ruma::{
 use stream_assert::assert_next_matches;
 
 use super::TestTimeline;
-use crate::timeline::{traits::RoomDataProvider as _, TimelineItemKind, VirtualTimelineItem};
+use crate::timeline::{traits::RoomDataProvider as _, VirtualTimelineItem};
 
 #[async_test]
 async fn test_day_divider() {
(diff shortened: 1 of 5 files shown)


crates/matrix-sdk-ui/src/timeline/tests/virt.rs
```diff
@@ -83,68 +84,107 @@ async fn test_day_divider() {
...
+    item.as_event().unwrap();
+
+    assert!(stream.next().now_or_never().is_none());
+
+    // But when it's another user who sent the event, then we get a read marker for
+    // their message.
+    timeline.handle_live_event(f.text_msg("H").sender(&BOB)).await;
+    let item = assert_next_matches!(stream, VectorDiff::PushBack { value } => value);
+    item.as_event().unwrap();
+
```
stefanceriu, line 187: Bear with me, just trying to understand the logic here:
* we left the read marker at `event_id4`
* we send 2 messages of our own and 1 from bob
* we get a read marker for `event_id8`?

Shouldn't the market have stayed at `event_id4` seeing how we never `handle_fully_read_marker`-ed it again?

bnjbvr: If the read marker was at `event_id4`, then it would look like the two events we sent ourselves are *before* the read marker, and this PR makes it so that events we sent ourselves are considered read, so should be *after* the read marker. I can beef up the code with comments showing what's displayed on the timeline at any interesting statement, if you think that can help with understanding the test code?

bnjbvr: Expanded the comments to show this in an hopefully clearer way.

stefanceriu: > I can beef up the code with comments

Thank you, it's super super clear now! 🙌

PR #3947, opened 2024-09-04
timeline: small cleanups (doc polls + use event factory more)

bnjbvr:
What it says on the tin. I was looking how poll events stash pending related events, will probably do something similar for edits later.


Hywan (approved):
What do we want? MORE CLEAN UPS.
When do we want them? NOW.

Thanks!

PR #3949, opened 2024-09-04
crypto: logging updates for verification

richvdh:
A collection of random logging updates, that would have made my life easier when trying to debug a verification failure this week.

### DIFF
--- a/crates/matrix-sdk-crypto/src/store/crypto_store_wrapper.rs
+++ b/crates/matrix-sdk-crypto/src/store/crypto_store_wrapper.rs
@@ -126,6 +126,26 @@ impl CryptoStoreWrapper {
 
         self.store.save_changes(changes).await?;
 
+        // If we updated our own public identity, log it for debugging purposes
+        if tracing::level_enabled!(tracing::Level::DEBUG) {
+            for updated_identity in
+                identities.new.iter().chain(identities.changed.iter()).filter_map(|id| id.own())
+            {
+                let master_key = updated_identity.master_key().get_first_key();
+                let user_signing_key = updated_identity.user_signing_key().get_first_key();
+                let self_signing_key = updated_identity.self_signing_key().get_first_key();
+
+                debug!(
+                    ?master_key,
+                    ?user_signing_key,
+                    ?self_signing_key,
+                    previously_verified = updated_identity.was_previously_verified(),
+                    verified = updated_identity.is_verified(),
+                    "Stored our own identity"
+                );
+            }
+        }
+
         if !room_key_updates.is_empty() {
             // Ignore the result. It can only fail if there are no listeners.
             let _ = self.room_keys_received_sender.send(room_key_updates);
--- a/crates/matrix-sdk-crypto/src/types/cross_signing/self_signing.rs
+++ b/crates/matrix-sdk-crypto/src/types/cross_signing/self_signing.rs
@@ -2,6 +2,7 @@ use std::collections::btree_map::Iter;
 
 use ruma::{encryption::KeyUsage, OwnedDeviceKeyId, UserId};
 use serde::{Deserialize, Serialize};
+use vodozemac::Ed25519PublicKey;
 
 use super::{CrossSigningKey, SigningKey};
 use crate::{
@@ -31,6 +32,14 @@ impl SelfSigningPubkey {
     /// Get the list of `KeyUsage` that is set for this key.
     pub fn usage(&self) -> &[KeyUsage] {
         &self.0.usage
+    }
+
+    /// Get the first available self signing key.
+    ///
+    /// There's usually only a single key so this will usually fetch the
+    /// only key.
+    pub fn get_first_key(&self) -> Option<Ed25519PublicKey> {
+        self.0.get_first_key_and_id().map(|(_, k)| k)
     }
 
     /// Verify that the [`DeviceKeys`] have a valid signature from this
--- a/crates/matrix-sdk-crypto/src/verification/cache.rs
+++ b/crates/matrix-sdk-crypto/src/verification/cache.rs
@@ -19,6 +19,8 @@ use std::{
 
 use as_variant::as_variant;
 use ruma::{DeviceId, OwnedTransactionId, OwnedUserId, TransactionId, UserId};
+#[cfg(feature = "qrcode")]
+use tracing::debug;
 use tracing::{trace, warn};
 
 use super::{event_enums::OutgoingContent, FlowId, Sas, Verification};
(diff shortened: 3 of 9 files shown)


crates/matrix-sdk-crypto/src/store/crypto_store_wrapper.rs
```diff
@@ -126,6 +126,24 @@ impl CryptoStoreWrapper {
 
         self.store.save_changes(changes).await?;
 
+        // If we updated our own public identity, log it for debugging purposes
+        for updated_identity in
```
bnjbvr, line 130: Can you wrap the loop in a `if tracing::level_enabled!(tracing::level::Debug)`?

crates/matrix-sdk-crypto/src/verification/cache.rs
```diff
@@ -105,11 +105,21 @@ impl VerificationCache {
 
     #[cfg(feature = "qrcode")]
     pub fn insert_qr(&self, qr: QrVerification) {
+        tracing::debug!(
```
bnjbvr, line 108: nit: don't use a fully qualified path but instead `use tracing::debug` at the top.

crates/matrix-sdk-crypto/src/verification/mod.rs
```diff
@@ -634,9 +634,10 @@ impl IdentitiesBeingVerified {
                 if verified_identities.is_some_and(|i| {
                     i.iter().any(|verified| verified.user_id() == identity.user_id())
                 }) {
-                    trace!(
+                    info!(
```
bnjbvr, line 637: Does this have to be an `info!` level? I suppose it's fine because it shouldn't happen that often?

crates/matrix-sdk-crypto/src/verification/qrcode.rs
```diff
@@ -328,6 +328,7 @@ impl QrVerification {
 
     /// Confirm that the other side has scanned our QR code.
     pub fn confirm_scanning(&self) -> Option<OutgoingVerificationRequest> {
+        info!("User confirmed other side scanned our QR code");
```
bnjbvr, line 331: Can this be a debug or trace instead? Otherwise the verification will cause many many info-level statements

bnjbvr (approved):
I'm worried that there are many high level log statements (info/debug), but if these are for events that don't happen that often I'm fine with the changes.

Still needs a review from some crypto folks to validate this is safe to publish this data.

richvdh:
>  I'm worried that there are many high level log statements (info/debug), but if these are for events that don't happen that often I'm fine with the changes.

Yes, these should be pretty rare: either a one-off at startup, or in response to a user-interactive verification, which itself happens very rarely.

crates/matrix-sdk-crypto/src/verification/mod.rs
richvdh: I think it's kinda important -- about the same as other stuff that is logged at info -- and it should only happen once per verification.

crates/matrix-sdk-crypto/src/verification/qrcode.rs
richvdh: yeah will make this one debug.

crates/matrix-sdk-crypto/src/store/crypto_store_wrapper.rs
richvdh: 063c62d6047db500f40479a1f17aa554cdef16dd

crates/matrix-sdk-crypto/src/verification/cache.rs
richvdh: ff484509331180c2be6114b32404485e948db2cf

crates/matrix-sdk-crypto/src/verification/qrcode.rs
richvdh: d7f7ad528f0180646473ab477cf7d75af764ce4a

crates/matrix-sdk-crypto/src/store/crypto_store_wrapper.rs
```diff
@@ -126,6 +126,26 @@ impl CryptoStoreWrapper {
...
+                    ?master_key,
+                    ?user_signing_key,
+                    ?self_signing_key,
+                    previously_verified = updated_identity.was_previously_verified(),
+                    verified = updated_identity.is_verified(),
+                    "Stored our own identity"
+                );
+            }
+        }
+
```
andybalaam, line 148: From a DM with Rich:

> Andy: You state here that these are all part of our public identity
> 
> Rich:
> 
> identities.new and identities.changed are both Vec<UserIdentityData>. UserIdentityData::own is Option<OwnUserIdentityData>, which is:

```
pub struct OwnUserIdentityData {
    user_id: OwnedUserId,
    master_key: Arc<MasterPubkey>,
    user_signing_key: Arc<UserSigningPubkey>,
    #[serde(deserialize_with = "deserialize_own_user_identity_data_verified")]
    verified: Arc<RwLock<OwnUserIdentityVerifiedState>>,
}
```

> ... and note the keys in there are all PubKey.

This convinced me :-)

andybalaam (approved):
All the additional crypto content logged is part of a user's public identity so OK to log.

crates/matrix-sdk-crypto/src/store/crypto_store_wrapper.rs
```diff
@@ -126,6 +126,26 @@ impl CryptoStoreWrapper {
 
         self.store.save_changes(changes).await?;
 
+        // If we updated our own public identity, log it for debugging purposes
+        if tracing::level_enabled!(tracing::Level::DEBUG) {
+            for updated_identity in
+                identities.new.iter().chain(identities.changed.iter()).filter_map(|id| id.own())
```
BillCarsonFr, line 132: Took me sometime to see why this was just own identity. I sort of expected a have a filter on `user_id() ==` but I guess it's fine

crates/matrix-sdk-crypto/src/types/cross_signing/self_signing.rs
```diff
@@ -33,6 +34,14 @@ impl SelfSigningPubkey {
         &self.0.usage
     }
 
+    /// Get the first available self signing key.
```
BillCarsonFr, line 37: This comment is a bit strange. The first available? It's just getting the ed25519 key and there is only one?

BillCarsonFr (approved):
Just minor nits, approved

crates/matrix-sdk-crypto/src/types/cross_signing/self_signing.rs
richvdh: Well, the line just below does explain that there is probably only one key.

Honestly I'm not a big fan of this API either, but it's just cargo-culted from the `MasterPubkey` implementation.

PR #3951, opened 2024-09-05
timeline: unify `edit` and `edit_poll` functions

jmartinesp:
## Changes

Takes care of [this TODO](https://github.com/matrix-org/matrix-rust-sdk/blob/9df1c480795c42afcee39e4c7e553d5a927a2680/crates/matrix-sdk-ui/src/timeline/mod.rs#L520).

- sdk & sdk-ui: unify `Timeline::edit` and `Timeline::edit_polls`, the new fn takes an `EditedContent` parameter now, which includes a `PollStart` case too.
- ffi: also unify the FFI fns there,  using `PollStart` and a new `EditContent` enum that must be passed from the clients like:

```kotlin
val messageContent = MessageEventContent.from(...)
timeline.edit(event, EditedContent.RoomMessage(messageContent))
```

Since the is mainly about changing the fns signatures I've reused the existing tests, including one that used `edit_poll` that now uses the new fn.

Signed-off-by:

### DIFF
--- a/crates/matrix-sdk-ui/src/timeline/mod.rs
+++ b/crates/matrix-sdk-ui/src/timeline/mod.rs
@@ -490,12 +490,13 @@ impl Timeline {
                                         return Ok(false);
                                     }
                                 }
-                                EditedContent::PollStart(_, poll_start) => {
+                                EditedContent::PollStart { new_content, .. } => {
                                     if matches!(item.content, TimelineItemContent::Poll(_)) {
-                                        let content = UnstablePollStartEventContent::New(
-                                            NewUnstablePollStartEventContent::new(poll_start),
-                                        );
-                                        AnyMessageLikeEventContent::UnstablePollStart(content)
+                                        AnyMessageLikeEventContent::UnstablePollStart(
+                                            UnstablePollStartEventContent::New(
+                                                NewUnstablePollStartEventContent::new(new_content),
+                                            ),
+                                        )
                                     } else {
                                         warn!("New content (poll start) doesn't match previous event content.");
                                         return Ok(false);
--- a/crates/matrix-sdk-ui/src/timeline/mod.rs
+++ b/crates/matrix-sdk-ui/src/timeline/mod.rs
@@ -483,13 +483,23 @@ impl Timeline {
                             // Relations are filled by the editing code itself.
                             let new_content: AnyMessageLikeEventContent = match new_content {
                                 EditedContent::RoomMessage(message) => {
-                                    AnyMessageLikeEventContent::RoomMessage(message.into())
+                                    if matches!(item.content, TimelineItemContent::Message(_)) {
+                                        AnyMessageLikeEventContent::RoomMessage(message.into())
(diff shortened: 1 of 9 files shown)


crates/matrix-sdk-ui/src/timeline/mod.rs
```diff
@@ -477,7 +474,7 @@ impl Timeline {
     pub async fn edit(
         &self,
         item: &EventTimelineItem,
-        new_content: RoomMessageEventContentWithoutRelation,
+        new_content: EditedContent,
```
bnjbvr, line 477: I do realize this function is missing some checks, and that predates your PR: what if we're trying to edit a poll with a room message? We'd need to check that the item content kind matches the `EditedContent` variant too.

Can you add a test for this please? :pleading_face:

bnjbvr (changes requested):
Wow, sweet!

1. You'd also need to change the code in the send queue here: https://github.com/matrix-org/matrix-rust-sdk/blob/32049537389c99eaa8d7062030a862bff48041fb/crates/matrix-sdk/src/send_queue.rs#L920
2. Can you add a test for the above, please? :pleading_face:

crates/matrix-sdk-ui/src/timeline/mod.rs
jmartinesp: That's already done [for edits to remote events](https://github.com/matrix-org/matrix-rust-sdk/blob/58ab52dc5f9853974658026bb4ffc70c922b2d43/crates/matrix-sdk/src/room/edit.rs#L158), although I'm not sure it's tested. Should I do that too for local echoes?

bnjbvr: Ah, you're right that the `make_edit` function a few layers below will fail if the timeline item doesn't contain something matching the event type. Yes, having something like this for 1. polls at the `make_edit` function level (I think it was there, IIRC), 2. local echoes for poll edits, will be useful.

jmartinesp:
> Wow, sweet!
> 
> 1. You'd also need to change the code in the send queue here: https://github.com/matrix-org/matrix-rust-sdk/blob/32049537389c99eaa8d7062030a862bff48041fb/crates/matrix-sdk/src/send_queue.rs#L920
> 2. Can you add a test for the above, please? 🥺

Done in https://github.com/matrix-org/matrix-rust-sdk/pull/3951/commits/fa0563e2872f922fda9b8cfb2716b3e55e01665b.

crates/matrix-sdk-ui/src/timeline/mod.rs
jmartinesp: https://github.com/matrix-org/matrix-rust-sdk/pull/3951/commits/26c2da4dd6cd967f66afa599b43b58866bd934db should handle this.

bnjbvr (changes requested):
Thanks, I feel like coding now so I'll handle my remaining review comments as a separate commit and will merge later :+1:

PR #3952, opened 2024-09-05
timeline: check that unique IDs are indeed unique

bnjbvr:
And log an error in production builds if that's not the case.

### DIFF
--- a/crates/matrix-sdk-ui/src/timeline/controller/state.rs
+++ b/crates/matrix-sdk-ui/src/timeline/controller/state.rs
@@ -15,6 +15,7 @@
 use std::{collections::VecDeque, future::Future, sync::Arc};
 
 use eyeball_im::{ObservableVector, ObservableVectorTransaction, ObservableVectorTransactionEntry};
+use itertools::Itertools as _;
 use matrix_sdk::{deserialized_responses::SyncTimelineEvent, send_queue::SendHandle};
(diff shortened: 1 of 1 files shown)


crates/matrix-sdk-ui/src/timeline/controller/state.rs
```diff
@@ -342,9 +346,33 @@ impl TimelineStateTransaction<'_> {
 
         self.adjust_day_dividers(day_divider_adjuster);
 
+        self.check_no_unused_unique_ids();
         total
     }
 
+    fn check_no_unused_unique_ids(&self) {
+        let mut known_unique_ids = HashSet::new();
```
poljar, line 354: We depend on `itertools` so we can do this instead:

```rust
let duplicates: Vec<_> = self.items.iter().duplicates_by(|item| item.unique_id()).collect();
```

More info: https://docs.rs/itertools/latest/itertools/trait.Itertools.html#method.duplicates_by.

poljar (approved):

crates/matrix-sdk-ui/src/timeline/controller/state.rs
bnjbvr: Indeed. `duplicates_by` returns the `Arc<TimelineItem>`, so I added a `map` to get the unique id itself instead.

Issue #3719, opened 2024-07-17
test: `encryption::backups::incremental_upload_of_keys` is failing intermittently


bnjbvr:
Example of a failed run [here](https://github.com/matrix-org/matrix-rust-sdk/actions/runs/9972852971/job/27556981541?pr=3695#step:9:1296).

Supposedly, an e2ee request has failed *before* the sync response has been received, because the `/keys/upload` request was created before receiving the response to sync. Not sure about that, would need more logs to confirm.

cc @BillCarsonFr

bnjbvr:
Oops, sounds like this should've been closed at the time.

PR #2939, opened 2023-12-13
Maybe trigger backup at end of sync

BillCarsonFr:
Fixes https://github.com/matrix-org/matrix-rust-sdk/issues/2938

There might be other way to do that. On some other client there is signaling when a new key is imported to trigger that.
But doing it after the sync is probably reasonable.

- [ ] Public API changes documented in changelogs (optional)

Signed-off-by:

PR #3953, opened 2024-09-05
tests: try to address intermittent failures of incremental backup tests

bnjbvr:
My theory is that the intermittent failure depends on the ordering of the requests, and if the /keys/upload request happened before the key backup request, then after failing the next key backup request wouldn't run.

This is likely a small typo that the key upload returns a 404 error instead of a 200, let's see if this improves the situation.

Should help with https://github.com/matrix-org/matrix-rust-sdk/issues/3719.

### DIFF
--- a/crates/matrix-sdk/tests/integration/encryption/backups.rs
+++ b/crates/matrix-sdk/tests/integration/encryption/backups.rs
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use std::{fs::File, io::Write, sync::Arc};
+use std::{fs::File, io::Write, sync::Arc, time::Duration};
 
 use anyhow::Result;
 use assert_matches::assert_matches;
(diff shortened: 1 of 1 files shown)


crates/matrix-sdk/tests/integration/encryption/backups.rs
```diff
@@ -574,7 +574,7 @@ async fn setup_create_room_and_send_message_mocks(server: &wiremock::MockServer)
     Mock::given(method("POST"))
         .and(path("/_matrix/client/r0/keys/upload"))
         .and(header("authorization", "Bearer 1234"))
-        .respond_with(ResponseTemplate::new(404).set_body_json(json!({
+        .respond_with(ResponseTemplate::new(200).set_body_json(json!({
```
Hywan, line 577: Reported here, https://github.com/matrix-org/matrix-rust-sdk/pull/2939/files#r1749634419. Reading the diffs, it looks like a copy-pasta error. I think it's fine to return a 200 here.

I am however really surprised it didn't crash more often. Returning a 200 instead of a 404 has also really few impacts. We may consider rewriting these tests but probably not today.

Hywan (approved):
Thanks for digging into this.

PR #3957, opened 2024-09-06
ui: Remove the e2e-encryption feature from the matrix-sdk-ui crate

poljar:
It does not make much sense to create an UI client that does not support end-to-end encryption, besides disabling the feature was broken for quite some time.

### DIFF
--- a/bindings/matrix-sdk-ffi/Cargo.toml
+++ b/bindings/matrix-sdk-ffi/Cargo.toml
@@ -28,7 +28,7 @@ eyeball-im = { workspace = true }
 extension-trait = "1.0.1"
 futures-util = { workspace = true }
 log-panics = { version = "2", features = ["with-backtrace"] }
-matrix-sdk-ui = { workspace = true, features = ["e2e-encryption", "uniffi"] }
+matrix-sdk-ui = { workspace = true, features = ["uniffi"] }
 mime = "0.3.16"
 once_cell = { workspace = true }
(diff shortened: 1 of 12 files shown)


poljar:
Kicking the CI.

crates/matrix-sdk/src/event_cache/linked_chunk/mod.rs
```diff
@@ -888,11 +888,7 @@ impl<'a, const CAP: usize, Item, Gap> Iterator for IterBackward<'a, CAP, Item, G
     type Item = &'a Chunk<CAP, Item, Gap>;
 
     fn next(&mut self) -> Option<Self::Item> {
-        self.chunk.map(|chunk| {
-            self.chunk = chunk.previous();
-
-            chunk
-        })
+        self.chunk.inspect(|chunk| self.chunk = chunk.previous())
```
andybalaam, line 891: Changes in this file seem unrelated?

andybalaam (changes requested):
One extra file seems to have crept in to the PR.

Looks good otherwise, and yay for simplification.

crates/matrix-sdk/src/event_cache/linked_chunk/mod.rs
poljar: That commit was part of #3958, why didn't github notice :shrug:, it did change the base from that branch to main automagically.

poljar: Rebased locally and force pushed to get rid of the extra commit.

andybalaam (approved):
Great, thanks!

PR #3961, opened 2024-09-08
sdk: Add support for authenticated media stable feature

zecakeh:
Was added post-merge to the MSC in https://github.com/matrix-org/matrix-spec-proposals/pull/4180 for servers that support authenticated media but do not support all of Matrix 1.11 yet.

Requires a kind of hack to force Ruma to select the stable endpoint by ignoring the real versions advertised by the homeserver for this particular endpoint.

### DIFF
--- a/crates/matrix-sdk/src/config/request.rs
+++ b/crates/matrix-sdk/src/config/request.rs
@@ -19,6 +19,7 @@ use std::{
 };
 
 use matrix_sdk_common::debug::DebugStructExt;
+use ruma::api::MatrixVersion;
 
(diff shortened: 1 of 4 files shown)


Hywan (approved):
It looks good to me, thanks!

Hywan (changes requested):
Is it possible to assert in the tests that we are actually forcing the matrix version?

zecakeh:
@Hywan, maybe you can clarify what you meant, if it is blocking the review?

Hywan:
Hmm, it seems fine actually, I don't remember exactly what I was referring to. I'm approving and merging this PR.

Hywan (approved):

Issue #3698, opened 2024-07-15
test: Use Synapse with Simplified Sliding Sync in our integration tests


Hywan:
In https://github.com/matrix-org/matrix-rust-sdk/pull/3676, we are migrating the SDK to use Simplified Sliding Sync instead of Sliding Sync itself.

When the PR was written, the version of Synapse used in our integration tests didn't have Simplified Sliding Sync, but was still using the Sliding Sync Proxy. As rightly suggested by @bnjbvr in https://github.com/matrix-org/matrix-rust-sdk/pull/3676#discussion_r1677576900, we should not forget to move away from the Proxy once everything is ready inside Synapse.

## Step 1

Run tests with Synapse instead of the sliding sync proxy:

* [x] https://github.com/matrix-org/matrix-rust-sdk/pull/3983

## Step 2

We now need to migrate the following tests (that are failing since the migration to Synapse):

* [x] `sliding_sync::room::test_room_avatar_group_conversation`
  * [x] https://github.com/matrix-org/matrix-rust-sdk/pull/4032
* [x] `sliding_sync::room::test_room_notification_count`
  * [x] https://github.com/ruma/ruma/pull/1914
  * [x] https://github.com/matrix-org/matrix-rust-sdk/pull/4031
* [ ] `sliding_sync::room::test_delayed_decryption_latest_event`
* [ ] `sliding_sync::room::test_room_info_notable_update_deduplication`
* [x] `sliding_sync::room::test_room_preview`
  * [x] https://github.com/matrix-org/matrix-rust-sdk/pull/4032
* [x] `sliding_sync::notification_client::test_notification`
  * [x] Blocked by https://github.com/element-hq/synapse/issues/17754
  * [x] The fix https://github.com/matrix-org/matrix-rust-sdk/pull/4029
  * [x] https://github.com/ruma/ruma/pull/1913

Some tests seem to be fixed with Synapse 1.115:

* [x] https://github.com/matrix-org/matrix-rust-sdk/pull/4030

---

* Address https://github.com/matrix-org/matrix-rust-sdk/issues/3647

poljar:
I believe this was closed by https://github.com/matrix-org/matrix-rust-sdk/pull/3983.

Hywan:
Not really. `test_delayed_decryption_latest_event` and `test_room_info_notable_update_deduplication` are still marked as `#[ignore]` 🫤.

PR #3963, opened 2024-09-09
chore(sdk): Replace “simplified sliding sync” by “MSC4186”

Hywan:
Simplified sliding sync now has an MSC number: 4186. Let's use it to clarify our code. I've done the same on Ruma side, and I've updated our `matrix-org/ruma` fork with its `feat-sss` branch.

### DIFF
--- a/crates/matrix-sdk-base/src/sliding_sync/mod.rs
+++ b/crates/matrix-sdk-base/src/sliding_sync/mod.rs
@@ -119,14 +119,14 @@ impl BaseClient {
     ///   sync.
     /// * `previous_events_provider` - Timeline events prior to the current
     ///   sync.
-    /// * `from_msc4186` - Whether the `response` comes from simplified sliding
+    /// * `with_msc4186` - Whether the `response` comes from simplified sliding
     ///   sync (MSC4186) or sliding sync (MSC3575).
     #[instrument(skip_all, level = "trace")]
     pub async fn process_sliding_sync<PEP: PreviousEventsProvider>(
         &self,
         response: &http::Response,
         previous_events_provider: &PEP,
-        from_msc4186: bool,
+        with_msc4186: bool,
     ) -> Result<SyncResponse> {
         let http::Response {
             // FIXME not yet supported by sliding sync. see
@@ -180,7 +180,7 @@ impl BaseClient {
                     &mut room_info_notable_updates,
                     &mut notifications,
                     &mut ambiguity_cache,
-                    from_msc4186,
+                    with_msc4186,
                 )
                 .await?;
 
@@ -352,7 +352,7 @@ impl BaseClient {
         room_info_notable_updates: &mut BTreeMap<OwnedRoomId, RoomInfoNotableUpdateReasons>,
         notifications: &mut BTreeMap<OwnedRoomId, Vec<Notification>>,
         ambiguity_cache: &mut AmbiguityCache,
-        from_msc4186: bool,
+        with_msc4186: bool,
     ) -> Result<(RoomInfo, Option<JoinedRoomUpdate>, Option<LeftRoomUpdate>, Option<InvitedRoom>)>
     {
         // This method may change `room_data` (see the terrible hack describes below)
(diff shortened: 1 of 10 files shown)


crates/matrix-sdk-base/src/sliding_sync/mod.rs
```diff
@@ -119,15 +119,14 @@ impl BaseClient {
...
-    ///   (MSC3575).
+    /// * `from_msc4186` - Whether the `response` comes from simplified sliding
+    ///   sync (MSC4186) or sliding sync (MSC3575).
     #[instrument(skip_all, level = "trace")]
     pub async fn process_sliding_sync<PEP: PreviousEventsProvider>(
         &self,
         response: &http::Response,
         previous_events_provider: &PEP,
-        from_simplified_sliding_sync: bool,
+        from_msc4186: bool,
```
bnjbvr, line 129: uber-nit: `with_msc4186`? The msc itself isn't a source of sync data, `from_..._sync` made sense up to now, but not the case with the msc number.

crates/matrix-sdk-base/src/sliding_sync/mod.rs
```diff
@@ -353,7 +352,7 @@ impl BaseClient {
         room_info_notable_updates: &mut BTreeMap<OwnedRoomId, RoomInfoNotableUpdateReasons>,
         notifications: &mut BTreeMap<OwnedRoomId, Vec<Notification>>,
         ambiguity_cache: &mut AmbiguityCache,
-        from_simplified_sliding_sync: bool,
+        from_msc4186: bool,
```
bnjbvr, line 355: and ditto in a few more places

crates/matrix-sdk/src/sliding_sync/mod.rs
```diff
@@ -531,7 +531,7 @@ impl SlidingSync {
     /// Send a sliding sync request.
     ///
     /// This method contains the sending logic. It takes a generic `Request`
-    /// because it can be a Simplified MSC3575 or a MSC3575 `Request`.
+    /// because it can be an MSC4186 or an MSC3575 `Request`.
```
bnjbvr, line 534: I'm not a native English person, so bear with me, but is it correct to say `an MSC request` instead of `a MSC request`?

crates/matrix-sdk/src/sliding_sync/mod.rs
```diff
@@ -682,10 +681,9 @@ impl SlidingSync {
         let (request, request_config, position_guard) =
             self.generate_sync_request(&mut LazyTransactionId::new()).await?;
 
-        // The code manipulates `Request` and `Response` from Simplified MSC3575
-        // because it's the future standard. If
-        // `Client::is_simplified_sliding_sync_enabled` is turned off, the
-        // Simplified MSC3575 `Request` must be transformed into a MSC3575 `Request`.
+        // The code manipulates `Request` and `Response` from MSC4186 because it's
+        // the future standard. Let's check if the generated request must be
```
bnjbvr, line 685: preexisting: if you say "future", you need to precise what is "now" by including a date :)

testing/matrix-sdk-integration-testing/src/helpers.rs
```diff
@@ -84,8 +84,8 @@ impl TestClientBuilder {
         let mut client_builder = Client::builder()
             .user_agent("matrix-sdk-integration-tests")
             .homeserver_url(homeserver_url)
-            // Disable Simplified MSC3575 for the integration tests as, at the time of writing
-            // (2024-07-15), we use a Synapse version that doesn't support Simplified MSC3575.
+            // Disable MSC4186 for the integration tests as, at the time of writing
+            // (2024-07-15), we use a Synapse version that doesn't support MSC4186.
```
bnjbvr, line 88: (Not for this PR, but) we should try to move forward on this maybe?

Hywan: For sure, but not today (maybe tomorrow?).

crates/matrix-sdk-base/src/sliding_sync/mod.rs
Hywan: I agree it creates confusion. Not a big fan of `with_…` but I can't come up with a better naming. Thanks for the proposal!

crates/matrix-sdk/src/sliding_sync/mod.rs
Hywan: Yup

testing/matrix-sdk-integration-testing/src/helpers.rs
Hywan: For the record, it's already tracked here https://github.com/matrix-org/matrix-rust-sdk/issues/3698

crates/matrix-sdk/src/sliding_sync/mod.rs
andybalaam: "an MSC request" when you are expecting (as I am) that the reader will pronounce it "em ess see" when they read.

Issue #3959, opened 2024-09-06
Sliding sync: Clear the device list cache when the connection is reset.


erikjohnston:
There is a change in behaviour with device lists between the sliding sync proxy and simplified sliding sync. When the client "resets" the encryption sync loop (i.e. sends a request with no `since` token):
- The proxy returns all device lists updates since the last request from the device that asked for device lists (this works similarly to to-device message handling). 
- Synapse returns no device lists updates, as it only returns changes since the provided `token` (which is null). This is in line with sync v2.

The effect of the above is that the rust sdk can easily miss device list updates that happen between the previous encryption sync request and sending an initial encryption sync request, when using SSS. Given the encryption loop "resets" relatively frequently (on one account it happened ~8 times in 12 hours), this will cause quite a few UTDs. 

The Rust SDK should therefore ensure whenever the connection is reset the device list cache is cleared. To mitigate the performance impact of this change, the SDK may want to try and reduce the number of times it proactively resets the connection (I think this happens whenever a request fails).

The alternative would be to change how Synapse behaves to match the proxy. However, this would be quite a big undertaking with unknown costs (device list management is already incredibly expensive in Synapse, converting it to track what updates we have sent to each device would likely make it even worse).

erikjohnston:
Given there are already locks to ensure that the encryption loop doesn't run at the same time as the other process, and correctly persists the received updates, I'm surprised it doesn't reuse the `pos`. 

> The major changes here afaict are

Those things won't actually fix this bug though, as you'd *still* need to invalidate the device list cache if you reset the connection. Those changes help by reducing the number of times the SDK does do that.

A few other things to note:
- The `e2ee` extension isn't needed to decrypt messages, so we could move the `e2ee`  extension out of the current encryption loop and have it just run on the main process.
- Doing a one-off encryption sync with no `since` token will return to-device messages that were already received. Not sure if that matters much.
- Doing an `encryption` sync without setting a `since` will invalidate the `pos` on the main process, meaning that the server will respond with an `M_UNKNOWN_POS` if the old `pos` was used again (though doesn't look like we do that).

---

In general I'd reiterate that the general architecture of sync (both sliding and v2) is to assume the client is starting from a clean slate if it receives an initial sync. Often this doesn't matter, but the SDK does have to be a little bit careful, e.g. (slightly different but still) the SDK does not clear the cached `required_state` when it receives a room entry with `initial=true` and instead treats the `required_state` as a delta.

PR #3965, opened 2024-09-09
fix(sdk): Mark tracked users as dirty when the SS connection is reset. 

Hywan:
Closes https://github.com/matrix-org/matrix-rust-sdk/issues/3959.

There is a non-negligible difference [MSC3575](https://github.com/matrix-org/matrix-spec-proposals/pull/3575) and [MSC4186](https://github.com/matrix-org/matrix-spec-proposals/pull/4186) in how the
`e2ee` extension works. When the client sends a request with no `pos`:

* [MSC3575](https://github.com/matrix-org/matrix-spec-proposals/pull/3575) returns all device lists updates since the last request
  from the device that asked for device lists (this works similarly to
  to-device message handling),

* [MSC4186](https://github.com/matrix-org/matrix-spec-proposals/pull/4186) returns no device lists updates, as it only returns changes
  since the provided `pos` (which is `null` in this case); this is in
  line with sync v2.

Therefore, with [MSC4186](https://github.com/matrix-org/matrix-spec-proposals/pull/4186), the device list cache must be marked as to be
re-downloaded if the `since` token is `None`, otherwise it's easy to
miss device lists updates that happened between the previous request and
the new “initial” request.

This patch first off implements `OlmMachine::mark_all_tracked_users_as_dirty`.
Next, this patch uses this method inside sliding sync when `pos` is `None` and
`e2ee` extension is enabled.

### DIFF
--- a/crates/matrix-sdk-crypto/src/machine/mod.rs
+++ b/crates/matrix-sdk-crypto/src/machine/mod.rs
@@ -185,29 +185,42 @@ impl OlmMachine {
             })
             .await?;
 
+        let (verification_machine, store, identity_manager) =
+            Self::new_helper_prelude(store, static_account, self.store().private_identity());
+
         Ok(Self::new_helper(
             device_id,
             store,
-            static_account,
+            verification_machine,
+            identity_manager,
             self.store().private_identity(),
             None,
         ))
     }
 
+    fn new_helper_prelude(
+        store_wrapper: Arc<CryptoStoreWrapper>,
+        account: StaticAccountData,
+        user_identity: Arc<Mutex<PrivateCrossSigningIdentity>>,
+    ) -> (VerificationMachine, Store, IdentityManager) {
+        let verification_machine =
+            VerificationMachine::new(account.clone(), user_identity.clone(), store_wrapper.clone());
+        let store = Store::new(account, user_identity, store_wrapper, verification_machine.clone());
+
+        let identity_manager = IdentityManager::new(store.clone());
+
+        (verification_machine, store, identity_manager)
+    }
+
     fn new_helper(
         device_id: &DeviceId,
-        store: Arc<CryptoStoreWrapper>,
-        account: StaticAccountData,
+        store: Store,
+        verification_machine: VerificationMachine,
+        identity_manager: IdentityManager,
         user_identity: Arc<Mutex<PrivateCrossSigningIdentity>>,
         maybe_backup_key: Option<MegolmV1BackupKey>,
     ) -> Self {
-        let verification_machine =
-            VerificationMachine::new(account.clone(), user_identity.clone(), store.clone());
-        let store = Store::new(account, user_identity.clone(), store, verification_machine.clone());
-
         let group_session_manager = GroupSessionManager::new(store.clone());
-
-        let identity_manager = IdentityManager::new(store.clone());
 
         let users_for_key_claim = Arc::new(StdRwLock::new(BTreeMap::new()));
         let key_request_machine = GossipMachine::new(
@@ -360,11 +373,21 @@ impl OlmMachine {
         let identity = Arc::new(Mutex::new(identity));
         let store = Arc::new(CryptoStoreWrapper::new(user_id, device_id, store));
 
+        let (verification_machine, store, identity_manager) =
+            Self::new_helper_prelude(store, static_account, identity.clone());
+
         // FIXME: We might want in the future a more generic high-level data migration
         // mechanism (at the store wrapper layer).
-        Self::migration_post_verified_latch_support(&store).await?;
+        Self::migration_post_verified_latch_support(&store, &identity_manager).await?;
 
-        Ok(OlmMachine::new_helper(device_id, store, static_account, identity, maybe_backup_key))
+        Ok(Self::new_helper(
+            device_id,
+            store,
+            verification_machine,
+            identity_manager,
+            identity,
+            maybe_backup_key,
+        ))
     }
 
     // The sdk now support verified identity change detection.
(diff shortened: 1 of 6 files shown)


crates/matrix-sdk-crypto/src/machine/mod.rs
```diff
@@ -1992,6 +2011,17 @@ impl OlmMachine {
         self.inner.identity_manager.update_tracked_users(users).await
     }
 
+    /// Mark all tracked users as dirty.
```
poljar, line 2014: `SyncedKeyQueryManager` is not `pub` it's `pub(crate)` so people can't look at the docs for this.

Shame that we can't put this method anywhere else since it needs the `IdentityManager`. Later on we should probably expose the `IdentityManager` publicly and move all the tracked user methods under that struct.

crates/matrix-sdk-crypto/src/machine/mod.rs
```diff
@@ -2457,6 +2499,29 @@ fn sender_data_to_verification_state(
     }
 }
 
+/// Real implementation for `OlmMachine::mark_all_tracked_users_as_dirty`.
```
poljar, line 2502: Let's at least put this method under the `IdentityManager`.

crates/matrix-sdk-crypto/src/machine/tests/mod.rs
```diff
@@ -1528,6 +1528,44 @@ async fn test_unsigned_decryption() {
     assert_matches!(thread_encryption_result, UnsignedDecryptionResult::Decrypted(_));
 }
 
+#[async_test]
+async fn test_mark_all_tracked_users_as_dirty() {
+    let store = MemoryStore::new();
+    let account = vodozemac::olm::Account::new();
+
+    // Put some tracked users
+    let damir = user_id!("@damir:localhost");
```
poljar, line 1537: :heart:

crates/matrix-sdk/src/sliding_sync/error.rs
```diff
@@ -53,4 +54,14 @@ pub enum Error {
...
+
+    /// No Olm machine.
+    #[cfg(feature = "e2e-encryption")]
+    #[error("The Olm machine is missing")]
+    NoOlmMachine,
+
+    /// An error occurred during a E2EE operation.
+    #[cfg(feature = "e2e-encryption")]
+    #[error(transparent)]
+    OlmError(#[from] OlmError),
```
poljar, line 66: I think you only need the `CryptoStoreError`?

crates/matrix-sdk/src/sliding_sync/mod.rs
```diff
@@ -513,6 +513,27 @@ impl SlidingSync {
...
+            // no `since` token:
+            //
+            // * MSC3575 returns all device lists updates since the last
+            //   request from the device that asked for device lists (this works
+            //   similarly to to-device message handling),
+            // * MSC4186 returns no device lists updates, as it only returns
+            //   changes since the provided `since` token (which is `null` in
+            //   this case); this is in line with sync v2.
+            //
+            // Therefore, with MSC4186, the device list cache must be reset if
```
poljar, line 528: We're not really clearing the cache, we're marking it that it needs to be re-downloaded. Not sure if you'd like to clarify that a bit.

poljar (changes requested):
Left some nits, I think it generally does what it should do.

crates/matrix-sdk/src/sliding_sync/mod.rs
Hywan: Indeed. Gonna modify.

crates/matrix-sdk-crypto/src/machine/mod.rs
Hywan: Done.

crates/matrix-sdk/src/sliding_sync/error.rs
Hywan: Yup. Done.

kegsay:
It's worth emphasising here that this will negatively impact performance and bandwidth, as the rust SDK will be doing many more `/keys/query` requests now because the client itself resets the E2EE connection more frequently due to not remembering `?pos=`. The SDK needs to remember the `?pos=` in the same way it remembers `to_device.since` to not be hit with this performance penalty.

crates/matrix-sdk/src/sliding_sync/mod.rs
```diff
@@ -491,6 +491,28 @@ impl SlidingSync {
...
+        //   sync v2.
+        //
+        // Therefore, with MSC4186, the device list cache must be marked as to be
+        // re-downloaded if the `since` token is `None`, otherwise it's easy to miss
+        // device lists updates that happened between the previous request and the new
+        // “initial” request.
+        #[cfg(feature = "e2e-encryption")]
+        if pos.is_none() && self.inner.version.is_native() && self.is_e2ee_enabled() {
+            let olm_machine = self.inner.client.olm_machine().await;
+            let olm_machine = olm_machine.as_ref().ok_or(Error::NoOlmMachine)?;
```
poljar, line 512: Perhaps we should add a log that this is happening?

crates/matrix-sdk-crypto/src/machine/mod.rs
```diff
@@ -1992,6 +2011,17 @@ impl OlmMachine {
         self.inner.identity_manager.update_tracked_users(users).await
     }
 
+    /// Mark all tracked users as dirty.
+    ///
+    /// See `IdentityManager::mark_all_tracked_users_as_dirty()` to learn
```
poljar, line 2016: IdentityManager isn't `pub`, besides the docs on `IdentityManager::mark_all_tracked_users_as_dirty()` also points you to another non-pub struct.

I'm afraid you'll have to document this without indirection right here.

poljar (approved):
There are still some nits left, but they are trivial so gonna approve ahead of time.

Thanks for taking the time to write a test for this.

crates/matrix-sdk/src/sliding_sync/mod.rs
Hywan: Good idea, done.

crates/matrix-sdk-crypto/src/machine/mod.rs
Hywan: Done!

PR #3966, opened 2024-09-09
FFI: Expose RoomSendQueue::unwedge to allow resending.

pixlwave:
This PR adds the `_and_resend` part of #3909 as it's own method so a user can re-verify on a different device that supports it, and then come back to the device using the FFI and send the message that failed to be sent.

### DIFF
--- a/bindings/matrix-sdk-ffi/src/room.rs
+++ b/bindings/matrix-sdk-ffi/src/room.rs
@@ -24,7 +24,8 @@ use ruma::{
         },
         TimelineEventType,
     },
-    EventId, Int, OwnedDeviceId, OwnedTransactionId, OwnedUserId, RoomAliasId, UserId,
+    EventId, Int, OwnedDeviceId, OwnedTransactionId, OwnedUserId, RoomAliasId, TransactionId,
+    UserId,
 };
 use tokio::sync::RwLock;
 use tracing::error;
@@ -808,6 +809,24 @@ impl Room {
 
         self.inner.send_queue().unwedge(&transaction_id).await?;
 
+        Ok(())
+    }
+
+    /// Attempt to manually resend messages that failed to send due to issues
+    /// that should now have been fixed.
+    ///
+    /// This is useful for example, when there's a
+    /// `SessionRecipientCollectionError::VerifiedUserChangedIdentity` error;
+    /// the user may have re-verified on a different device and would now
+    /// like to send the failed message that's waiting on this device.
+    ///
+    /// # Arguments
+    ///
+    /// * `transaction_id` - The send queue transaction identifier of the local
+    ///   echo that should be unwedged.
+    pub async fn try_resend(&self, transaction_id: String) -> Result<(), ClientError> {
+        let transaction_id: &TransactionId = transaction_id.as_str().into();
+        self.inner.send_queue().unwedge(transaction_id).await?;
         Ok(())
     }
 }


bindings/matrix-sdk-ffi/src/room.rs
```diff
@@ -810,6 +810,23 @@ impl Room {
...
+    /// the user may have re-verified on a different device and would now
+    /// like to send the failed message that's waiting on this device.
+    ///
+    /// # Arguments
+    ///
+    /// * `transaction_id` - The send queue transaction identifier of the local
+    ///  echo the send error applies to.
+    pub async fn resend(&self, transaction_id: String) -> Result<(), ClientError> {
+        let transaction_id: OwnedTransactionId = transaction_id.into();
+        self.inner.send_queue().unwedge(&transaction_id).await?;
```
bnjbvr, line 827:
```suggestion
        let transaction_id: &TransactionId = transaction_id.into();
        self.inner.send_queue().unwedge(transaction_id).await?;
```

bindings/matrix-sdk-ffi/src/room.rs
```diff
@@ -810,6 +810,23 @@ impl Room {
 
         Ok(())
     }
+
+    /// Attempt to manually resend messages that failed to send due to issues
+    /// that should now have been fixed. This is useful for example, when
+    /// there's a
```
bnjbvr, line 816: nit: let's add some breathing space:

```suggestion
    /// that should now have been fixed.
    ///
    /// This is useful for example, when there's a
```

bindings/matrix-sdk-ffi/src/room.rs
```diff
@@ -810,6 +810,23 @@ impl Room {
...
+    /// `SessionRecipientCollectionError::VerifiedUserChangedIdentity` error;
+    /// the user may have re-verified on a different device and would now
+    /// like to send the failed message that's waiting on this device.
+    ///
+    /// # Arguments
+    ///
+    /// * `transaction_id` - The send queue transaction identifier of the local
+    ///  echo the send error applies to.
+    pub async fn resend(&self, transaction_id: String) -> Result<(), ClientError> {
```
bnjbvr, line 825: Could we call this `try_resend()` instead?

bindings/matrix-sdk-ffi/src/room.rs
```diff
@@ -810,6 +810,23 @@ impl Room {
...
+    /// that should now have been fixed. This is useful for example, when
+    /// there's a
+    /// `SessionRecipientCollectionError::VerifiedUserChangedIdentity` error;
+    /// the user may have re-verified on a different device and would now
+    /// like to send the failed message that's waiting on this device.
+    ///
+    /// # Arguments
+    ///
+    /// * `transaction_id` - The send queue transaction identifier of the local
+    ///  echo the send error applies to.
```
bnjbvr, line 824: Bad copy pasto, nowhere else we do mention a "send error".

bindings/matrix-sdk-ffi/src/room.rs
pixlwave: Ah nice, didn't know you could `into` a borrowed type.

bindings/matrix-sdk-ffi/src/room.rs
pixlwave: Sure :)

bindings/matrix-sdk-ffi/src/room.rs
pixlwave: 🍝

bindings/matrix-sdk-ffi/src/room.rs
pixlwave: I had to add an `as_str()` to this to make the compiler happy.

bnjbvr (approved):

PR #3967, opened 2024-09-10
sdk-ui: add `TimelineState::replace_all` to batch clear + append

jmartinesp:
This combines `clear` and `add_remote_events_at` in the same transaction, preventing some flickering from happening in the clients when we removed every single item in the timelines and a few ms later we added them back again.

Signed-off-by:


crates/matrix-sdk-ui/src/timeline/controller/mod.rs
```diff
@@ -637,8 +637,6 @@ impl<P: RoomDataProvider> TimelineController<P> {
     ) {
         let mut state = self.state.write().await;
 
-        state.clear();
```
jmartinesp, line 640: This is now done in `state.replace_all(...)`.

crates/matrix-sdk-ui/src/timeline/controller/mod.rs
```diff
@@ -647,9 +645,9 @@ impl<P: RoomDataProvider> TimelineController<P> {
                 .await;
         }
 
-        if !events.is_empty() {
+        if !state.items.is_empty() || !events.is_empty() {
```
jmartinesp, line 648: The previous check made sense because we used `clear` before, but now we may want to replace a populated timeline with an empty one.

crates/matrix-sdk-ui/src/timeline/controller/state.rs
```diff
@@ -265,6 +265,22 @@ impl TimelineState {
         txn.commit();
     }
 
+    pub(super) async fn replace_all<P: RoomDataProvider>(
+        &mut self,
+        events: Vec<impl Into<SyncTimelineEvent>>,
```
bnjbvr, line 270: nit: no generics here please

crates/matrix-sdk-ui/src/timeline/controller/state.rs
```diff
@@ -265,6 +265,22 @@ impl TimelineState {
         txn.commit();
     }
 
+    pub(super) async fn replace_all<P: RoomDataProvider>(
```
bnjbvr, line 268: nit: rename `replace_with_remote_events`

bnjbvr (approved):

crates/matrix-sdk-ui/src/timeline/controller/mod.rs
bnjbvr: Add this in a comment please.

PR #3968, opened 2024-09-10
feat(sdk): Use the user ID to discover the sliding sync proxy

Hywan:
This patch improves `Client::available_sliding_sync_versions` when trying to detect the sliding sync proxy. Previously, we were relying on the `Client::server` to send the `discover_homeserver::Request`. Sadly, this value is an `Option<_>`, meaning it's not always defined (it depends how the `Client` has been built with `HomeserverConfig`: sometimes the homeserver URL is passed directly, so the server cannot be known).

This patch tries to find to discover the homeserver by using `Client::server` if it exists, like before, but it also tries by using `Client::user_id`. Another problem arises then: the user ID indeed contains a server name, but we don't know whether it's behind HTTPS or HTTP. Thus, this patch tries both: it starts by testing with `https://` and then fallbacks to `http://`.

A test has been added accordingly.

### DIFF
--- a/crates/matrix-sdk/src/sliding_sync/client.rs
+++ b/crates/matrix-sdk/src/sliding_sync/client.rs
@@ -16,7 +16,7 @@ use tracing::error;
 use url::Url;
 
 use super::{SlidingSync, SlidingSyncBuilder};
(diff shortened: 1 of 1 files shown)


stefanceriu (approved):
Works great, thank you!

PR #3970, opened 2024-09-10
crypto: Bump the vodozemac version and remove the PkEncryption compat module

poljar:
The PkEncryption support now lives inside of vodozemac so no need to keep our own copy around.


andybalaam (approved):
Looks good, thanks!

PR #3972, opened 2024-09-10
MatrixRTC: Fix different devices from the same user overwriting the room info state event.

toger5:
This PR is best reviewed with the background from: https://github.com/matrix-org/matrix-rust-sdk/pull/3998
The main change in this PR is the transition from using a UserId as the index for call member events in the room info to using the newly introduced `CallMemberStateKey` that is a combination of UserId and device id.

This is important since otherwise the room state could only store one event per user (but a user can join with multiple devices)

It also adds tests for session membership events (which is the new state event format for call member state events we need for reliable state events)

Signed-off-by:

### DIFF
--- a/crates/matrix-sdk-base/src/rooms/normal.rs
+++ b/crates/matrix-sdk-base/src/rooms/normal.rs
@@ -30,7 +30,7 @@ use ruma::events::AnySyncTimelineEvent;
 use ruma::{
     api::client::sync::sync_events::v3::RoomSummary as RumaSummary,
     events::{
-        call::member::MembershipData,
+        call::member::{CallMemberStateKey, MembershipData},
         ignored_user_list::IgnoredUserListEventContent,
         receipt::{Receipt, ReceiptThread, ReceiptType},
         room::{
@@ -1485,10 +1485,10 @@ impl RoomInfo {
     /// associated UserId's in this room.
     ///
     /// The vector is ordered by oldest membership to newest.
-    fn active_matrix_rtc_memberships(&self) -> Vec<(OwnedUserId, MembershipData<'_>)> {
+    fn active_matrix_rtc_memberships(&self) -> Vec<(CallMemberStateKey, MembershipData<'_>)> {
         let mut v = self
             .base_info
-            .rtc_member
+            .rtc_member_events
             .iter()
             .filter_map(|(user_id, ev)| {
                 ev.as_original().map(|ev| {
@@ -1509,7 +1509,7 @@ impl RoomInfo {
     /// returns Memberships with application "m.call" and scope "m.room".
     ///
     /// The vector is ordered by oldest membership user to newest.
-    fn active_room_call_memberships(&self) -> Vec<(OwnedUserId, MembershipData<'_>)> {
+    fn active_room_call_memberships(&self) -> Vec<(CallMemberStateKey, MembershipData<'_>)> {
         self.active_matrix_rtc_memberships()
             .into_iter()
             .filter(|(_user_id, m)| m.is_room_call())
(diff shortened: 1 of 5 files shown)


toger5:
The commit history does not conform to guidelines. I suspect a rebase will be required anyways which contains the switch to the ruma/ruma.

toger5:
Tarpaulin is failing. `test_stale_local_echo_time_abort_edit` (fails)

Is this sth that I can fix?

bnjbvr:
> Tarpaulin is failing. `test_stale_local_echo_time_abort_edit` (fails)
> 
> Is this sth that I can fix?

No, intermittent failure. Do you have the permissions to restart it by yourself in the CI interface? (I'll restart it now for you)

crates/matrix-sdk-base/src/rooms/normal.rs
```diff
@@ -2729,7 +2732,25 @@ mod tests {
         .expect("date out of range")
     }
 
-    fn call_member_state_event(
+    fn legacy_membership_for_my_call(
+        device_id: &str,
```
Hywan, line 2736: Can't you use a `DeviceId` type from Ruma?

crates/matrix-sdk-base/src/rooms/normal.rs
```diff
@@ -2743,37 +2764,68 @@ mod tests {
             // we can simply use now here since this will be dropped when using a MinimalStateEvent
             // in the roomInfo
             origin_server_ts: timestamp(0),
-            state_key: CallMemberStateKey::from_str(user_id.as_str())
-                // SAFETY: `user_id` is a valid `UserId` and cannot fail to be transformed into a
-                // `CallMemberStateKey`.
-                .expect("Failed to transform a `UserId` into a `CallMemberStateKey`"),
+            state_key: CallMemberStateKey::new(user_id.to_owned(), None, false),
```
Hywan, line 2767: 👍

crates/matrix-sdk-base/src/rooms/normal.rs
```diff
@@ -2743,37 +2764,68 @@ mod tests {
...
 
-    fn membership_for_my_call(
-        device_id: &str,
-        membership_id: &str,
+    struct InitData<'a> {
+        device_id: &'a DeviceId,
         minutes_ago: u32,
-    ) -> LegacyMembershipData {
+    }
+    fn session_member_state_event(
```
Hywan, line 2776: Please add an empty line between the structure and the function.

crates/matrix-sdk-base/src/rooms/normal.rs
```diff
@@ -2743,37 +2764,68 @@ mod tests {
...
+        ev_id: &str,
```
Hywan, line 2777: Can't we use an `EventId` type from Ruma?

Hywan (changes requested):
It looks good, thanks for these changes!

Tiny feedback.

crates/matrix-sdk-base/src/rooms/normal.rs
toger5: `LegacyMembershipDataInit` should then definitly also require a DeviceId. This cannot be done in this PR but I will add the change to another PR.

toger5: I added into this PR: https://github.com/ruma/ruma/pull/1908
It seems fitting to use this as the ruma cleanup PR.

toger5:
> LegacyMembershipDataInit should then definitly also require a DeviceId. This cannot be done in this PR but I will add the change to another PR.
(https://github.com/matrix-org/matrix-rust-sdk/pull/3972#discussion_r1761392170)
The ruma PR that changes this got merged, so I updated this PR to use the updated `LegacyMembershipDataInit`.

PR #3974, opened 2024-09-10
event cache: reset paginator state when receiving a limited timeline

bnjbvr:
Before this patch, we did reset the paginator state just a bit too late, that is: at the beginning of running the next back-pagination. Resetting the state would trigger an update of the paginator status, which would reflect at the timeline paginator status level too.

But we have a bit of a conundrum here. Imagine the following sequence of events:

- back-pagination => at some point the paginator from the event cache hits the start of the timeline
- it reflects that at the timeline level, setting `hit_start_of_timeline` to true
- we received a limited/gappy sync via (simplified?) sliding sync, which resets the internal back-pagination token to None in the room event cache (not the associated paginator)
- this clears the timeline

Now, calling the paginator's `run_backwards()` method (or equivalently, the timeline's `live_paginate_backwards` method) would reset the pagination token as intended. But an API consumer would not see a status update there, so they're not aware that they can run back-pagination, because the previous status indicated that pagination hit the start of the timeline (and it hasn't changed since then, b/o absence of status update). Kaboom, we're stuck.

The resolution consists in resetting the paginator's state a bit earlier, i.e. just after receiving a limited timeline. In that case, we can even forward the pagination token a bit ahead of time, if it was already available.

Includes a regression test, that failed when checking `!hit_start_of_timeline()` after the timeline reset, before the patch.

Should fix https://github.com/element-hq/element-x-ios/issues/2024, @pixlwave can you confirm please?

### DIFF
--- a/crates/matrix-sdk/src/event_cache/paginator.rs
+++ b/crates/matrix-sdk/src/event_cache/paginator.rs
@@ -219,10 +219,18 @@ impl<PR: PaginableRoom> Paginator<PR> {
     /// (running /context or /messages).
     pub(super) fn set_idle_state(
         &self,
+        next_state: PaginatorState,
         prev_batch_token: Option<String>,
         next_batch_token: Option<String>,
     ) -> Result<(), PaginatorError> {
         let prev_state = self.state.get();
+
+        match next_state {
+            PaginatorState::Initial | PaginatorState::Idle => {}
+            PaginatorState::FetchingTargetEvent | PaginatorState::Paginating => {
+                panic!("internal error: set_idle_state only accept Initial|Idle next states");
+            }
+        }
 
         match prev_state {
             PaginatorState::Initial | PaginatorState::Idle => {}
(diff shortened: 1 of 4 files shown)


jmartinesp (approved):
Thanks, the code changes LGTM. Now we wait for @pixlwave 's confirmation that it fixes the issue.

crates/matrix-sdk/src/event_cache/paginator.rs
```diff
@@ -219,11 +219,19 @@ impl<PR: PaginableRoom> Paginator<PR> {
     /// (running /context or /messages).
     pub(super) fn set_idle_state(
         &self,
+        next_state: PaginatorState,
```
bnjbvr, line 222: Thanks for the review!

Small API question here: I could've kept the previous code, *but*! if I did that, there would not be any state notification, because the code below uses `self.state.set_if_not_eq(Idle)`, and the previous value was still `Idle`. Using `Initial` is fine because it's also the initial set you get when creating an empty paginator. Another way to fix that would've been to replace `set_if_not_eq` with `set`, which forces a notification (and maybe check in the timeline that we're not deduplicating such updates. Any preference over here?

jmartinesp: I think using `Initial` makes sense and as you said prevents the clients from messing up the integration by deduplicating the state updates, which is something we tend to do on other places.

pixlwave:
Thanks @bnjbvr, I can confirm, this fixes the issue 🎉

crates/matrix-sdk/src/event_cache/paginator.rs
bnjbvr: Good point, thanks!

PR #3976, opened 2024-09-11
sdk: notify when a sync response is received by SlidingSync

jmartinesp:
Previously this was only done in `Client::sync_once`, which made `ClientInner::sync_beat` not that useful.

Signed-off-by:

### DIFF
--- a/crates/matrix-sdk/src/sliding_sync/mod.rs
+++ b/crates/matrix-sdk/src/sliding_sync/mod.rs
@@ -710,7 +710,7 @@ impl SlidingSync {
         // The code manipulates `Request` and `Response` from MSC4186 because it's
         // the future standard (at the time of writing: 2024-09-09). Let's check if
         // the generated request must be transformed into an MSC3575 `Request`.
-        if !self.inner.version.is_native() {
+        let result = if !self.inner.version.is_native() {
             self.send_sync_request(
                 Into::<http::msc3575::Request>::into(request),
                 request_config,
@@ -719,7 +719,12 @@ impl SlidingSync {
             .await
         } else {
             self.send_sync_request(request, request_config, position_guard).await
-        }
+        };
+
+        // Notify a new sync was received
+        self.inner.client.inner.sync_beat.notify(usize::MAX);
+
+        result
     }
 
     /// Create a _new_ Sliding Sync sync loop.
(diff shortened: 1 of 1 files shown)


crates/matrix-sdk/src/sliding_sync/mod.rs
```diff
@@ -3073,4 +3078,50 @@ mod tests {
...
+        let sliding_sync = Arc::new(sliding_sync);
+
+        assert!(!client.inner.sync_beat.is_notified());
+
+        // Create the listener and perform a sync request
+        let sync_beat_listener = client.inner.sync_beat.listen();
+        sliding_sync.sync_once().await?;
+
+        // The sync beat listener should be notified shortly after
+        timeout(sync_beat_listener, Duration::from_millis(20))
```
bnjbvr, line 3119:
```suggestion
        timeout(sync_beat_listener, Duration::from_secs(1))
```

crates/matrix-sdk/src/sliding_sync/mod.rs
```diff
@@ -3073,4 +3078,50 @@ mod tests {
...
+    #[async_test]
+    async fn test_sync_beat_is_notified_on_sync_response() -> Result<()> {
+        let server = MockServer::start().await;
+        let client = logged_in_client(Some(server.uri())).await;
+
+        let pos = Arc::new(Mutex::new(0));
+        let _mock_guard = Mock::given(SlidingSyncMatcher)
+            .respond_with(move |_: &Request| {
+                let mut pos = pos.lock().unwrap();
+                *pos += 1;
```
bnjbvr, line 3091: You don't need the whole `pos` stuff, since it's only returning once, you can return an hardcoded pos of `1`?

crates/matrix-sdk/src/sliding_sync/mod.rs
```diff
@@ -719,7 +719,12 @@ impl SlidingSync {
             .await
         } else {
             self.send_sync_request(request, request_config, position_guard).await
-        }
+        };
+
+        // Notify a new sync was received
+        self.inner.client.inner.sync_beat.notify(usize::MAX);
+
+        result
```
bnjbvr, line 727: This isn't the same behavior as sync v3: the sync beat only beats if when the sync is successful there. Can you fix the behavior here, and add a corresponding test for this please?

bnjbvr (changes requested):

bnjbvr (approved):
Thanks! I'm afraid the timeout values are too low to be of any help, so let's modify these and merge!

PR #3977, opened 2024-09-11
feat(ffi): Add `RoomInfo::creator`.

Hywan:
This patch adds the `matrix_sdk_ffi::RoomInfo::creator` field
as an `Option<String>`. This value comes from the new method
`matrix_sdk_base::Room::creator()`, which simply forwards the value from the
existing `matrix_sdk_base::RoomInfo::creator()`.

---

* Addresses https://github.com/element-hq/customer-success/issues/347#issuecomment-2312062942


poljar (approved):

bmarty (approved):
I have tested the fix and this is working well! thanks!

PR #3979, opened 2024-09-11
feat: add fn to await a synced room from `Client`

jmartinesp:
## Changes

- Add `fn sdk_base::Room::is_state_partially_or_fully_synced()`. I was planning to reuse `Room::is_state_fully_synced()`, but this apparently doesn't work with SSS.
- Add `fn Client::await_room_remote_echo(&RoomId)`, which will loop indefinitely until it finds a room with the provided room id, using `ClientInner::sync_beat` to delay the checks. This is the core of the changes, and an FFI fn has been created to access it.
- `SlidingSyncRoom` is no longer needed in `fn RoomListItem::default_room_timeline_builder()`. It was mandatory, and used only to pre-fill the timeline with initial items. After discussing this behavior we decided we can make it non-mandatory so timelines can be created for rooms that haven't been synced yet, which was a source of pain in the clients.
- Fix dumb typo in `TimelineState`.

## Motivation

After using `Client::create_room` in the clients we found some cases where we were trying to create a timeline for the just created room while it still hadn't been synced and this resulted in an error and the full room not being returned to the clients. The new `await_room_remote_echo` fn will let us wait until the room has been synced, fixing this issue.

Also, there may be some cases where a room is in the local DB but it hasn't been synced yet. The behaviour change in `default_room_timeline_builder` should fix that.

Signed-off-by:

### DIFF
--- a/bindings/matrix-sdk-ffi/src/client.rs
+++ b/bindings/matrix-sdk-ffi/src/client.rs
@@ -1002,6 +1002,16 @@ impl Client {
 
         Ok(RoomPreview::from_sdk(sdk_room_preview))
     }
+
+    /// Waits until an at least partially synced room is received, and returns
+    /// it.
+    ///
+    /// **Note: this function will loop endlessly until either it finds the room
+    /// or an externally set timeout happens.**
+    pub async fn await_room_remote_echo(&self, room_id: String) -> Result<Arc<Room>, ClientError> {
+        let room_id = RoomId::parse(room_id)?;
+        Ok(Arc::new(Room::new(self.inner.await_room_remote_echo(&room_id).await)))
+    }
 }
 
 #[uniffi::export(callback_interface)]
--- a/crates/matrix-sdk-ui/src/room_list_service/room.rs
+++ b/crates/matrix-sdk-ui/src/room_list_service/room.rs
@@ -20,6 +20,7 @@ use std::{ops::Deref, sync::Arc};
 use async_once_cell::OnceCell as AsyncOnceCell;
 use matrix_sdk::SlidingSync;
 use ruma::RoomId;
+use tracing::info;
 
 use super::Error;
 use crate::{
@@ -150,27 +151,32 @@ impl Room {
     }
 
     /// Create a new [`TimelineBuilder`] with the default configuration.
+    ///
+    /// If the room was synced before some initial events will be added to the
+    /// [`TimelineBuilder`].
     pub async fn default_room_timeline_builder(&self) -> Result<TimelineBuilder, Error> {
         // TODO we can remove this once the event cache handles his own cache.
 
-        let sliding_sync_room =
-            self.inner
-                .sliding_sync
-                .get_room(self.inner.room.room_id())
-                .await
-                .ok_or_else(|| Error::RoomNotFound(self.inner.room.room_id().to_owned()))?;
+        let sliding_sync_room = self.inner.sliding_sync.get_room(self.inner.room.room_id()).await;
 
-        self.inner
-            .room
-            .client()
-            .event_cache()
-            .add_initial_events(
-                self.inner.room.room_id(),
-                sliding_sync_room.timeline_queue().iter().cloned().collect(),
-                sliding_sync_room.prev_batch(),
-            )
-            .await
-            .map_err(Error::EventCache)?;
+        if let Some(sliding_sync_room) = sliding_sync_room {
+            self.inner
+                .room
+                .client()
+                .event_cache()
+                .add_initial_events(
+                    self.inner.room.room_id(),
+                    sliding_sync_room.timeline_queue().iter().cloned().collect(),
+                    sliding_sync_room.prev_batch(),
+                )
+                .await
+                .map_err(Error::EventCache)?;
+        } else {
+            info!(
+                "No cached sliding sync room found for `{}`, the timeline will be empty.",
+                self.room_id()
+            );
+        }
 
         Ok(Timeline::builder(&self.inner.room).track_read_marker_and_receipts())
     }
--- a/crates/matrix-sdk/src/client/mod.rs
+++ b/crates/matrix-sdk/src/client/mod.rs
@@ -2252,14 +2252,12 @@ impl Client {
                 if room.is_state_partially_or_fully_synced() {
                     debug!("Found just created room!");
                     return room;
-                } else {
-                    warn!("Room wasn't partially synced, waiting for sync beat to try again");
                 }
+                debug!("Room wasn't partially synced, waiting for sync beat to try again");
             } else {
-                warn!("Room wasn't found, waiting for sync beat to try again");
+                debug!("Room wasn't found, waiting for sync beat to try again");
             }
             self.inner.sync_beat.listen().await;
-            debug!("New sync beat found");
         }
     }
 }
@@ -2310,6 +2308,7 @@ pub(crate) mod tests {
     use std::{sync::Arc, time::Duration};
 
     use assert_matches::assert_matches;
+    use futures_util::FutureExt;
     use matrix_sdk_base::{
         store::{MemoryStore, StoreConfig},
         RoomState,
@@ -2328,6 +2327,10 @@ pub(crate) mod tests {
         owned_room_id, room_id, RoomId, ServerName, UserId,
     };
     use serde_json::json;
+    use tokio::{
+        spawn,
+        time::{sleep, timeout},
+    };
     use url::Url;
     use wiremock::{
         matchers::{body_json, header, method, path, query_param_is_missing},
@@ -2668,7 +2671,7 @@ pub(crate) mod tests {
         let client = logged_in_client(None).await;
 
         // Wait for the init tasks to die.
-        tokio::time::sleep(Duration::from_secs(1)).await;
+        sleep(Duration::from_secs(1)).await;
 
         let weak_client = WeakClient::from_client(&client);
         assert_eq!(weak_client.strong_count(), 1);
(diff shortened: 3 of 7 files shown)


crates/matrix-sdk/src/client/mod.rs
```diff
@@ -2240,6 +2240,28 @@ impl Client {
...
+    /// or an externally set timeout happens.**
+    pub async fn await_room_remote_echo(&self, room_id: &RoomId) -> Room {
+        loop {
+            if let Some(room) = self.get_room(room_id) {
+                if room.is_state_partially_or_fully_synced() {
+                    debug!("Found just created room!");
+                    return room;
+                } else {
+                    warn!("Room wasn't partially synced, waiting for sync beat to try again");
+                }
```
bnjbvr, line 2257: nit: no else after return please :)

Also I think the warn can just be a debug, it's not high priority enough that it should jump out to our eyes when looking at rageshakes IMO.

crates/matrix-sdk/src/client/mod.rs
```diff
@@ -2240,6 +2240,28 @@ impl Client {
...
+            } else {
+                warn!("Room wasn't found, waiting for sync beat to try again");
```
bnjbvr, line 2259: ditto, debug here

crates/matrix-sdk/src/client/mod.rs
```diff
@@ -2240,6 +2240,28 @@ impl Client {
...
+            }
+            self.inner.sync_beat.listen().await;
+            debug!("New sync beat found");
```
bnjbvr, line 2262: and this one we can remove, since the next loop iteration will immediately log something or return.

crates/matrix-sdk/src/client/mod.rs
```diff
@@ -2773,4 +2799,126 @@ pub(crate) mod tests {
...
+            .respond_with(ResponseTemplate::new(200).set_body_json(response_body))
+            .mount(&server)
+            .await;
+
+        client.sync_once(SyncSettings::default()).await.unwrap();
+
+        let room =
+            tokio::time::timeout(Duration::from_secs(1), client.await_room_remote_echo(room_id))
+                .await
+                .unwrap();
```
bnjbvr, line 2830: We can do slightly better here, and not involve time at all:

```rust
let room = client.await_room_remote_echo(room_id).now_or_never().unwrap();
```

(`now_or_never()` requires a trait to be in scope)

crates/matrix-sdk/src/client/mod.rs
```diff
@@ -2773,4 +2799,126 @@ pub(crate) mod tests {
...
+        builder
+            .respond_with(ResponseTemplate::new(200).set_body_json(response_body))
+            .mount(&server)
+            .await;
+
+        let client = Arc::new(client);
+
+        // Perform the /sync request with a delay so it starts after the
+        // `await_room_remote_echo` call has happened
+        tokio::spawn({
```
bnjbvr, line 2860: nit, here and below for the `tokio::` usage, please add `use tokio::` statements at the top of this file, so it's shorter to read here

crates/matrix-sdk/src/client/mod.rs
```diff
@@ -2773,4 +2799,126 @@ pub(crate) mod tests {
...
+        let client = client_builder.request_config(RequestConfig::new()).build().await.unwrap();
+        set_client_session(&client).await;
+
+        let room_id = room_id!("!room:example.org");
+        // Room is not present so the client won't be able to find it. The call will
+        // timeout.
+        let err =
+            tokio::time::timeout(Duration::from_secs(1), client.await_room_remote_echo(room_id))
+                .await
+                .err();
```
bnjbvr, line 2887: can `unwrap_err()` and remove the `let err =` and the assertion below

crates/matrix-sdk/src/client/mod.rs
```diff
@@ -2773,4 +2799,126 @@ pub(crate) mod tests {
...
+            .await
+            .unwrap();
+
+        // Room is locally present, but not synced, the call will timeout
+        let err = tokio::time::timeout(
+            Duration::from_secs(1),
+            client.await_room_remote_echo(room.room_id()),
+        )
+        .await
+        .err();
```
bnjbvr, line 2921: ditto here (nice tests!)

bindings/matrix-sdk-ffi/src/client.rs
```diff
@@ -1002,6 +1002,16 @@ impl Client {
 
         Ok(RoomPreview::from_sdk(sdk_room_preview))
     }
+
+    /// Waits until an at least partially synced room is received, and returns
+    /// it.
+    ///
+    /// **Note: this function will loop endlessly until either it finds the room
+    /// or an externally set timeout happens.**
```
bnjbvr, line 1010: i'd do a slightly different API at the FFI level: add an `Option<Duration>` as a parameter, and add a timeout here, so the consumer doesn't have to. Does it make sense? (You get to decide if it's a good idea, based on your usage of the API in the app :))

crates/matrix-sdk-ui/src/room_list_service/room.rs
```diff
@@ -150,27 +150,32 @@ impl Room {
...
 
-        let sliding_sync_room =
+        let sliding_sync_room = self
+            .inner
+            .sliding_sync
+            .get_room(self.inner.room.room_id())
+            .await
+            .ok_or_else(|| Error::RoomNotFound(self.inner.room.room_id().to_owned()));
+
+        if let Some(sliding_sync_room) = sliding_sync_room {
```
bnjbvr, line 161: Is this code correct? I'd think that `sliding_sync_room` is a `Result` here, not an `Option`…

crates/matrix-sdk-ui/src/room_list_service/room.rs
```diff
@@ -150,27 +150,32 @@ impl Room {
...
-            .event_cache()
-            .add_initial_events(
-                self.inner.room.room_id(),
-                sliding_sync_room.timeline_queue().iter().cloned().collect(),
-                sliding_sync_room.prev_batch(),
-            )
-            .await
-            .map_err(Error::EventCache)?;
+                .map_err(Error::EventCache)?;
+        }
```
bnjbvr, line 173: Can you add some info! statement in the else case, to indicate that we didn't have a timeline cache for this room? That will help debunk issues when people are in offline mode.

crates/matrix-sdk-ui/src/room_list_service/room.rs
```diff
@@ -150,27 +150,32 @@ impl Room {
     }
 
     /// Create a new [`TimelineBuilder`] with the default configuration.
+    ///
+    /// If the room was synced before some initial events will be added to the
+    /// [`TimelineBuilder`].
```
bnjbvr, line 155: Can you add some tests checking for this new behavior?

crates/matrix-sdk-ui/src/room_list_service/room.rs
```diff
@@ -156,12 +156,7 @@ impl Room {
     pub async fn default_room_timeline_builder(&self) -> Result<TimelineBuilder, Error> {
         // TODO we can remove this once the event cache handles his own cache.
 
-        let sliding_sync_room = self
-            .inner
-            .sliding_sync
-            .get_room(self.inner.room.room_id())
-            .await
-            .ok_or_else(|| Error::RoomNotFound(self.inner.room.room_id().to_owned()));
+        let sliding_sync_room = self.inner.sliding_sync.get_room(self.inner.room.room_id()).await;
```
bnjbvr, line 159: ah, here you go \o/

bnjbvr (changes requested):
Just an extra test please, LGTM otherwise!

crates/matrix-sdk-ui/src/room_list_service/room.rs
jmartinesp: I think this is already tested [here](https://github.com/matrix-org/matrix-rust-sdk/blob/5e662e855db925b380517bd94ef2b3622804a421/crates/matrix-sdk-ui/tests/integration/room_list_service.rs#L2382), it checks the previously synced item is already there: I can create another test if you want to, but it seems kind of redundant. 

Do you want me to test the opposite, which happens when no `SlidingSyncRoom` is found?

bnjbvr: > Do you want me to test the opposite, which happens when no SlidingSyncRoom is found?

Yes, indeed, since that's the new behavior we're introducing in this PR.

jmartinesp: Ok, the doc comment mentioned the opposite, so I was a bit confused 😅 . I'll add one, thanks!

jmartinesp: https://github.com/matrix-org/matrix-rust-sdk/pull/3979/commits/b472a87ba09723b1b2a65d7b8a66a867f685b686 added a test for this.

bnjbvr: Thanks!

bnjbvr (changes requested):
Thanks for the test. A few comments still unaddressed, so requesting changes again.

jmartinesp:
My bad, I have addressed the rest in https://github.com/matrix-org/matrix-rust-sdk/pull/3979/commits/606c6b9ca022939eba87344cd9e4f17da448f5ee, I thought I had already uploaded it.

bindings/matrix-sdk-ffi/src/client.rs
jmartinesp: I asked @stefanceriu and we decided to have the client implement the timeout.

bnjbvr (approved):
Thanks!

Issue #3978, opened 2024-09-11
sliding_sync::client::tests::test_available_sliding_sync_versions_proxy test failure


poljar:
This is failing on an unrelated PR: https://github.com/matrix-org/matrix-rust-sdk/actions/runs/10810424442/job/29988925650?pr=3970#step:9:1098

poljar:
This has been fixed, more info at: https://github.com/matrix-org/matrix-rust-sdk/pull/3980#issuecomment-2343551302.

PR #3980, opened 2024-09-11
test: Don't use the matrix.org server for the sliding sync version discovery tests

poljar:
Since the well-known lookup will modify the homeserver you are using, we need to set the homeserver URL to our own mocked server as well.


poljar:
This should have fixed https://github.com/matrix-org/matrix-rust-sdk/issues/3978, but a fix has been silently snuck into commit a6f84d8513cb5008c07225280f93f5f6919a188a as well.

PR #3985, opened 2024-09-11
crypto: change withheld code for IdentityBased share strategy

BillCarsonFr:
Change the withheld code for the `IdentityBasedStrategy` from `WithheldCode::Unauthorised`
  to `WithheldCode::Unverified`.

Signed-off-by:

### DIFF
--- a/crates/matrix-sdk-crypto/CHANGELOG.md
+++ b/crates/matrix-sdk-crypto/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 Changes:
 
+- Change the withheld code for keys not shared due to the `IdentityBasedStrategy`, from `m.unauthorised`
+  to `m.unverified`.
+  ([#3985](https://github.com/matrix-org/matrix-rust-sdk/pull/3985))
+
 - Improve logging for undecryptable Megolm events.
   ([#3989](https://github.com/matrix-org/matrix-rust-sdk/pull/3989))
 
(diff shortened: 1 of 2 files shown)


stefanceriu (approved):
Looks good to me but can we have an explanation as to why this happened?

BillCarsonFr:
> Looks good to me but can we have an explanation as to why this happened?

You can find more here https://github.com/element-hq/element-web-rageshakes/issues/27055#issuecomment-2344092110

TL;DR It's for the new invisible crypto, and I was not sure of what withheld code would be the best to return (unauthorised, unverified, a new one?). After some discussion we decided to use `m.unverified` for now (also there is proper UX for that code)

richvdh:
Even more context:

`m.unverified` could be considered inappropriate because that code means that the sender is only sending to devices/users that the *sender* has verified (eg via emojis). That is not, strictly speaking, the problem here: the sender does not care that they have not done an interactive verification; they do care that the device *owner* has not done a verification.

Nevertheless, it feels less bad than co-opting `m.unauthorised`. The spec isn't very explicit, but the intention with `m.unauthorised` is that it be sent in response to a key-share request.

Ideally, we would have a different withheld code, more relevant to this case, but doing so would require spec changes and a bunch of work on the recipient side to handle the new code.

Using `m.unverified` feels like a pragmatic solution.

crates/matrix-sdk-crypto/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 Changes:
 
+- Change the withheld code for the `IdentityBasedStrategy` from `WithheldCode::Unauthorised`
+  to `WithheldCode::Unverified`
```
richvdh, line 6:
```suggestion
- Change the withheld code for keys not shared due to the `IdentityBasedStrategy`, from `m.unauthorised`
  to `m.unverified`.
```

richvdh (approved):
LGTM otherwise

PR #3987, opened 2024-09-11
WidgetDriver: Handle widget send action that is a redact.

toger5:

Signed-off-by:

### DIFF
--- a/crates/matrix-sdk/src/widget/matrix.rs
+++ b/crates/matrix-sdk/src/widget/matrix.rs
@@ -112,7 +112,6 @@ impl MatrixDriver {
     /// This method allows the widget machine to handle widget requests by
     /// providing a unified, high-level widget-specific API for sending events
     /// to the room.
-
     pub(crate) async fn send(
         &self,
         event_type: TimelineEventType,
--- a/crates/matrix-sdk/src/widget/matrix.rs
+++ b/crates/matrix-sdk/src/widget/matrix.rs
@@ -29,10 +29,10 @@ use ruma::{
         AnyMessageLikeEventContent, AnyStateEventContent, AnySyncTimelineEvent, AnyTimelineEvent,
         MessageLikeEventType, StateEventType, TimelineEventType,
     },
-    serde::Raw,
-    RoomId, TransactionId,
+    serde::{from_raw_json_value, Raw},
+    EventId, RoomId, TransactionId,
 };
-use serde_json::value::RawValue as RawJsonValue;
+use serde_json::{value::RawValue as RawJsonValue, Value};
 use tokio::sync::mpsc::{unbounded_channel, UnboundedReceiver};
 use tracing::error;
 
@@ -107,7 +107,12 @@ impl MatrixDriver {
         Ok(events)
     }
 
-    /// Sends a given `event` to the room.
+    /// Sends the given `event` to the room.
+    ///
+    /// This method allows the widget machine to handle widget requests by
+    /// providing a unified, high-level widget-specific API for sending events
+    /// to the room.
+
     pub(crate) async fn send(
         &self,
         event_type: TimelineEventType,
@@ -116,6 +121,15 @@ impl MatrixDriver {
         delayed_event_parameters: Option<delayed_events::DelayParameters>,
     ) -> Result<SendEventResponse> {
         let type_str = event_type.to_string();
+
+        if let Some(redacts) = from_raw_json_value::<Value, serde_json::Error>(&content)
+            .ok()
+            .and_then(|b| b["redacts"].as_str().and_then(|s| EventId::parse(s).ok()))
+        {
+            return Ok(SendEventResponse::from_event_id(
+                self.room.redact(&redacts, None, None).await?.event_id,
+            ));
+        }
         Ok(match (state_key, delayed_event_parameters) {
             (None, None) => SendEventResponse::from_event_id(
                 self.room.send_raw(&type_str, content).await?.event_id,
--- a/crates/matrix-sdk/tests/integration/widget.rs
+++ b/crates/matrix-sdk/tests/integration/widget.rs
@@ -19,6 +19,7 @@ use async_trait::async_trait;
 use futures_util::FutureExt;
 use matrix_sdk::{
     config::SyncSettings,
+    test_utils::mocks::MatrixMockServer,
     widget::{
         Capabilities, CapabilitiesProvider, WidgetDriver, WidgetDriverHandle, WidgetSettings,
     },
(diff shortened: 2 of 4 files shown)


bnjbvr:
Hi @toger5, we're triaging the SDK PRs a bit, what's the status of this one? Do you need help, a review, or should it be closed until it's ready/because it's not needed eventually?

toger5:
@bnjbvr This is a pr we definitely want to get merged since it enables the raise hand/reactions feature on EX.
The only reason for it to stale is bandwidth.
The only missing bit is adding a test case that hits:
```
match (state_key, delayed_event_parameters, redacts) {

...
        // this case
        (None, None, Some(redacts)) => SendEventResponse::from_event_id(
            self.room.redact(&redacts, None, None).await?.event_id,
        ),
        
...
}
```

crates/matrix-sdk/src/widget/matrix.rs
```diff
@@ -116,14 +121,20 @@ impl MatrixDriver {
         delayed_event_parameters: Option<delayed_events::DelayParameters>,
     ) -> Result<SendEventResponse> {
         let type_str = event_type.to_string();
-        Ok(match (state_key, delayed_event_parameters) {
-            (None, None) => SendEventResponse::from_event_id(
+
+        let redacts =
+            from_raw_json_value::<RedactBody, serde_json::Error>(&content).ok().map(|b| b.redacts);
```
bnjbvr, line 126: Instead of adding `redacts` to the match below, can we do

```rust
if let Some(redacts) = ... {
  return Ok(self.room.redact(&redacts, None, None).await?.event_id);
}
```

crates/matrix-sdk/src/widget/matrix.rs
```diff
@@ -47,6 +47,11 @@ pub(crate) struct MatrixDriver {
     room: Room,
 }
 
+#[derive(serde::Deserialize)]
+struct RedactBody {
```
bnjbvr, line 51: nit: add a doc comment explaining what this is, how it's used, where it comes from (I suppose this is part of the widget protocol? is it an input or an output?)

crates/matrix-sdk/tests/integration/widget.rs
```diff
@@ -827,6 +827,55 @@ async fn test_try_update_delayed_event_without_permission_negotiate() {
...
+    assert_eq!(msg["api"], "fromWidget");
+    assert_eq!(msg["action"], "send_event");
+    let redact_event_id = msg["response"]["event_id"].as_str().unwrap();
+    let redact_room_id = msg["response"]["room_id"].as_str().unwrap();
+
+    assert_eq!(redact_event_id, "$redact_event_id");
+    assert_eq!(redact_room_id, "!a98sd12bjh:example.org");
+
+    // Make sure the event-sending endpoint was hit exactly once
+    mock_server.verify().await;
```
bnjbvr, line 876: Fwiw, you don't need this, verify() is called when the `MockServer` is dropped.

bnjbvr (changes requested):
Thanks. Would it make sense to have another method for redacting, since most of the parameters aren't used when redacting?

Bonus points if you add nice comments to `send()`, since as a newcomer I have no idea what calls it, etc.

crates/matrix-sdk/src/widget/matrix.rs
toger5: Yea that also work. I was torn between this which keeps the match expressions small and the readibility by having all possible matrix driver sends in one single match statement.
Thanks for providing you opinion!

crates/matrix-sdk/src/widget/matrix.rs
toger5: I think I actually prefer to make the parsing from RawValue to the radacts index explicitly.

crates/matrix-sdk/src/widget/matrix.rs
```diff
@@ -116,6 +120,16 @@ impl MatrixDriver {
         delayed_event_parameters: Option<delayed_events::DelayParameters>,
     ) -> Result<SendEventResponse> {
         let type_str = event_type.to_string();
+
+        if let Some(redacts) = from_raw_json_value::<Value, serde_json::Error>(&content)
+            .ok()
+            .and_then(|b| b["redacts"].as_str().map(|s| EventId::parse(s).ok()))
+            .flatten()
```
bnjbvr, line 127: Does this work?

```suggestion
            .and_then(|b| b["redacts"].as_str())
            .and_then(|s| EventId::parse(s).ok())
```

bnjbvr (changes requested):

crates/matrix-sdk/src/widget/matrix.rs
toger5: I dont think so since b is only referenced and does not like leaving the `|b| {}` context.

toger5: we can do it with one additional clone (to_owned)?

toger5: I ended up to at least replace the map+flatten with a and_then.

toger5:
```
from_raw_json_value::<Value, serde_json::Error>(&content)
            .ok()
            .and_then(|b| b["redacts"].as_str().and_then(|s| EventId::parse(s).ok()))
            ```

File: crates/matrix-sdk/src/widget/matrix.rs [hunk 5]
```diff
@@ -108,6 +108,11 @@ impl MatrixDriver {
     }
 
     /// Sends a given `event` to the room.
```
poljar, line 110:
```suggestion
    /// Sends the given `event` to the room.
```

crates/matrix-sdk/src/widget/matrix.rs
```diff
@@ -108,6 +108,11 @@ impl MatrixDriver {
...
+    ///
+    /// This is the method that is used by the widget machine to execute the
```
poljar, line 112: This whole sentence is a bit clumsy, perhaps something like:

> This method allows the widget machine to handle widget requests by providing a unified, high-level widget-specific API for sending events to the room.

poljar (approved):
Left a couple of nits for the docs, otherwise looks good and I think all the previous review requests were addressed.

Thanks for switching to the new `MatrixMockServer`.

Feel free to address the doc nits and at the same time squash/clean up the history.

After that we can merge.

crates/matrix-sdk/src/widget/matrix.rs
```diff
@@ -107,7 +107,12 @@ impl MatrixDriver {
         Ok(events)
     }
 
-    /// Sends a given `event` to the room.
+    /// Sends the given `event` to the room.
+    ///
+    /// This method allows the widget machine to handle widget requests by
+    /// providing a unified, high-level widget-specific API for sending events
+    /// to the room.
+
```
poljar, line 115:
```suggestion
```

PR #3989, opened 2024-09-12
crypto: Include megolm ratchet index in logging span fields

richvdh:
This field is helpful as it tells us the sequence number of the message in the megolm session, which gives us a clue about how long it will have been since the session should have been shared with us.

[Question: an alternative would be just to attach the whole of `content` to the span, instead of separate fields for `sender_key`, `session_id`, `message_index`. The Debug defintion for `MegolmV1AesSha2Content` looks like it contains the right things, so the code might be simpler. On the other hand, I think the output would be less readable. Thoughts?]

### DIFF
--- a/crates/matrix-sdk-crypto/CHANGELOG.md
+++ b/crates/matrix-sdk-crypto/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 Changes:
 
+- Improve logging for undecryptable Megolm events.
+  ([#3989](https://github.com/matrix-org/matrix-rust-sdk/pull/3989))
+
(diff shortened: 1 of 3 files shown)


poljar:
> [Question: an alternative would be just to attach the whole of `content` to the span, instead of separate fields for `sender_key`, `session_id`, `message_index`. The Debug defintion for `MegolmV1AesSha2Content` looks like it contains the right things, so the code might be simpler. On the other hand, I think the output would be less readable. Thoughts?]

I think I  prefer separate fields, since the content might contain unknown fields somebody might include something they might not want to end up in logs or might contain spam.

PR #3990, opened 2024-09-13
sdk-base: use updated account data for processing direct rooms

jmartinesp:
When processing a sliding sync response, if no account_data for direct rooms is found in the current response, use the cached value in the store.

This fixes an issue in the EX clients where only very few DMs were visible under the `People` filter since they weren't considered DMs, until a full restart of the app.

Signed-off-by:

### DIFF
--- a/crates/matrix-sdk-base/src/client.rs
+++ b/crates/matrix-sdk-base/src/client.rs
@@ -1169,11 +1169,19 @@ impl BaseClient {
         // because we want to have the push rules in place before we process
         // rooms and their events, but we want to create the rooms before we
         // process the `m.direct` account data event.
-        if let Ok(Some(direct_account_data)) =
+        let has_new_direct_room_data = response.account_data.events.iter().any(|raw_event| {
+            raw_event
+                .deserialize()
+                .map(|event| event.event_type() == GlobalAccountDataEventType::Direct)
+                .unwrap_or_default()
+        });
+        if has_new_direct_room_data {
+            self.handle_account_data(&response.account_data.events, &mut changes).await;
+        } else if let Ok(Some(direct_account_data)) =
             self.store.get_account_data_event(GlobalAccountDataEventType::Direct).await
         {
             debug!("Found direct room data in the Store, applying it");
-            self.handle_account_data(&vec![direct_account_data], &mut changes).await;
+            self.handle_account_data(&[direct_account_data], &mut changes).await;
         }
 
         changes.presence = response
--- a/crates/matrix-sdk-base/src/client.rs
+++ b/crates/matrix-sdk-base/src/client.rs
@@ -1169,7 +1169,12 @@ impl BaseClient {
         // because we want to have the push rules in place before we process
         // rooms and their events, but we want to create the rooms before we
         // process the `m.direct` account data event.
-        self.handle_account_data(&response.account_data.events, &mut changes).await;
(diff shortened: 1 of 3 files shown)


crates/matrix-sdk-base/src/client.rs
```diff
@@ -1169,7 +1169,12 @@ impl BaseClient {
         // because we want to have the push rules in place before we process
         // rooms and their events, but we want to create the rooms before we
         // process the `m.direct` account data event.
-        self.handle_account_data(&response.account_data.events, &mut changes).await;
+        if let Ok(Some(direct_account_data)) =
```
poljar, line 1172: I don't think that this is the correct place to do this. Do note that the currently received account data will only be persisted after the following code snippet further bellow:

https://github.com/matrix-org/matrix-rust-sdk/blob/5827bb7ab33b275835f16eacb4c0fecfa79f4a2b/crates/matrix-sdk-base/src/client.rs#L1187-L1188

If the current sync has some new account data, this `handle_account_data()` call will overwrite it with the old one found in the store.

I think that the most correct way would be to apply the direct targets when we initially create the `Room` object, though that might mean that we load the account data from the store multiple times. I can't come up with a more performant way for this right now.

poljar (changes requested):
I think that this introduces a subtle bug where we might go back in time with the account data.

crates/matrix-sdk-base/src/client.rs
```diff
@@ -1169,7 +1169,14 @@ impl BaseClient {
         // because we want to have the push rules in place before we process
         // rooms and their events, but we want to create the rooms before we
         // process the `m.direct` account data event.
-        self.handle_account_data(&response.account_data.events, &mut changes).await;
+        if !response.account_data.is_empty() {
```
poljar, line 1172: Still not correct, consider the following scenario:

We get some updated account data events, but the `m.direct` event was not updated. So `response.account_data` will not be empty, but it will also not contain the `m.direct` event. Thus rooms will remain without their direct target.

poljar (changes requested):

crates/matrix-sdk-base/src/client.rs
jmartinesp: I hope https://github.com/matrix-org/matrix-rust-sdk/pull/3990/commits/b53fa163922365d0ec021c4b2a767b7ff3d6e02b handles that in a better way 🤞 .

poljar (approved):
Ok, this can go in as a fix. If you have the time and motivation it would be good to get rid of this TODO item above as well.

For this we would split the `handle_account_data()` method into two methods, one that handles all account data except for `m.direct` and one that handles just the `m.direct` events.

I think then this PR also becomes less of a hack.

jmartinesp:
Since we want to merge this ASAP, do you mind if we merge this as is, then I immediately start working in a PR with the changes you propose? That way we have the fix in now since we need, and we can have the right structure soon-ish (today, tomorrow?).

poljar:
> Since we want to merge this ASAP, do you mind if we merge this as is, then I immediately start working in a PR with the changes you propose? That way we have the fix in now since we need it, and we can have the right structure soon-ish (today, tomorrow?).

I don't mind this going in, no. After all I approved the PR and said as much in my review:

> Ok, this can go in as a fix.

Issue #1415, opened 2023-01-27
Lost OTK, leading to "OneTime key already exists" error and later UTDs


BillCarsonFr:
From time to time we are getting this error on various clients.

Element R iOS shows it as a modal:

<img width="292" alt="image" src="https://user-images.githubusercontent.com/9841565/215055401-9a9754d6-198f-46f9-81dd-f182c5369d70.png">

Most other clients (including Element Web R and Element X) just log the error without showing it to the user.

This happens ([among other reasons](https://github.com/element-hq/element-meta/issues/2406#issuecomment-2079953181)) when a client creates a one-time keypair, uploads the public part to the server, and then forgets that it has done so (eg due to unsafe concurrent access to the crypto store). The client is then unable to upload further OTKs until the forgotten one is claimed by another device; and once it *is* claimed by another device, the resultant messages will be undecryptable.

Anderas:
Root cause on iOS: `recieve_sync_changes` is not multiprocess safe, and one process may be writing new values of account to the store whilst another retains stale cache that may eventually override the store with outdated values. When this happens, `next_key_id` for one time keys is dropped to some lower value, and this then triggers duplicate OTK ids.

Steps to reproduce:
1. create two `OlmMachine`s, `foreground` and background
2. simulate few sync loops on `foreground` by
    - calling `receive_sync_changes` with `key_counts` lower than 50
    - `mark_request_as_sent` for `keysUpload` setting it back to 50
    - => as a result `foreground` will keep creating new OTKs and incrementing `next_key_id`
4. call single `receive_sync_changes` on `background`
=> result: background has just overriden account in store with `next_key_id` that is much lower than background

Anderas:
Some rageshakes
- https://github.com/matrix-org/element-ios-rageshakes/issues/22849
- https://github.com/matrix-org/element-ios-rageshakes/issues/22863
- https://github.com/matrix-org/element-ios-rageshakes/issues/22936

Anderas:
Lastly we also need a recovery method, given that OTK upload failure will wedge the machine and never mark OTKs as sent. Possible solution

1. Pretend that the upload succeeded, setting a safe default value for the uploaded one-time key count on the Account
2. Claim all your keys on the server, you don't know which one-time key has lost his private parts
3. Re-upload a new fresh set of one-time keys

richvdh:
[...]

PR #3992, opened 2024-09-13
crypto: Don't deep copy the OlmMachine when creating a notification client

poljar:
Opening as a draft since I will want to make test for this. This closes: #1415.


crates/matrix-sdk-ui/tests/integration/encryption_sync_service.rs
```diff
@@ -320,3 +337,231 @@ async fn test_encryption_sync_always_reloads_todevice_token() -> anyhow::Result<
...
+        .unwrap();
+
+    let session = MatrixSession {
+        meta: SessionMeta { user_id: user_id.into(), device_id: device_id!("DEVICEID").to_owned() },
+        tokens: MatrixSessionTokens { access_token: "1234".to_owned(), refresh_token: None },
+    };
+
+    client.restore_session(session.to_owned()).await.unwrap();
+
+    tracing::info!("Creating the notification client");
```
bnjbvr, line 363: nit, here and below: feel free to `use tracing::{info, warn, etc};`

crates/matrix-sdk-ui/tests/integration/encryption_sync_service.rs
```diff
@@ -320,3 +337,231 @@ async fn test_encryption_sync_always_reloads_todevice_token() -> anyhow::Result<
...
+
+                let duplicates: HashSet<_> = uploaded_key_ids.intersection(&new_key_ids).collect();
+
+                if let Some(duplicate) = duplicates.into_iter().next() {
+                    tracing::error!("Duplicate one-time keys were uploaded.");
+
+                    found_duplicate.store(true, Ordering::SeqCst);
+
+                    ResponseTemplate::new(400).set_body_json(json!({
+                        "errcode": "M_WAT",
```
bnjbvr, line 442: nice ^^ This results in an unparseable http error result, likely, though, right? I mean, at this point the test is to fail, but will it fail in a way we can understand?

crates/matrix-sdk-ui/tests/integration/encryption_sync_service.rs
```diff
@@ -320,3 +337,231 @@ async fn test_encryption_sync_always_reloads_todevice_token() -> anyhow::Result<
...
+                        }
+                    }))
+                }
+            }
+        })
+        .expect(4)
+        .mount(&server)
+        .await;
+
+    tracing::info!("Main sync not gets told that a one-time key has been used up.");
```
bnjbvr, line 461: nit: s/not gets/doesn't get/

crates/matrix-sdk-base/src/client.rs
```diff
@@ -162,13 +162,30 @@ impl BaseClient {
...
+        let (room_info_notable_update_sender, _room_info_notable_update_receiver) =
+            broadcast::channel(100);
+
+        let copy = Self {
+            store: Store::new(config.state_store),
+            event_cache_store: config.event_cache_store,
+            crypto_store: self.crypto_store.clone(),
+            olm_machine: self.olm_machine.clone(),
+            ignore_user_list_changes: Default::default(),
+            room_info_notable_update_sender,
```
bnjbvr, line 178: It doesn't matter much since we're not using it in the notification client, but for correctness it should be the same one as in `self`, I suppose.

crates/matrix-sdk-base/src/client.rs
```diff
@@ -162,13 +162,30 @@ impl BaseClient {
...
-        result.room_key_recipient_strategy = self.room_key_recipient_strategy.clone();
-        result
+        let (room_info_notable_update_sender, _room_info_notable_update_receiver) =
+            broadcast::channel(100);
+
+        let copy = Self {
+            store: Store::new(config.state_store),
+            event_cache_store: config.event_cache_store,
+            crypto_store: self.crypto_store.clone(),
+            olm_machine: self.olm_machine.clone(),
```
bnjbvr, line 176: Can you add a comment here, basically stating that it's fine to clone the current `olm_machine` which embeds the `self.crypto_store` because it's the same `crypto_store` in both? (I wonder if implementing the proper cross-process lock mechanism could allow us to get rid of `BaseClient::crypto_store` — i fathom it's there because we sometimes re-create the `crypto_store`, but that's bad state duplication :fearful:)

bnjbvr (approved):
Amazing catch, thanks for writing a test too. I'll send you my best baguette over mail.

crates/matrix-sdk-ui/tests/integration/encryption_sync_service.rs
poljar: It does get parsed as it has the correct Matrix error json thing. No, the sync loop won't abort at this point, nor will it return an error. The test checks the `found_duplicate` boolean for that.

crates/matrix-sdk-ui/tests/integration/encryption_sync_service.rs
poljar: s/not/now rather.

crates/matrix-sdk-base/src/client.rs
poljar: It's rather there because we can create a `OlmMachine` only once we're logged in, i.e. we require a `UserId` and `DeviceId`, you could convert this into a `UnitializedOlmMachine` instead of having two optionals, this would make things clearer I would say.

PR #3995, opened 2024-09-13
Get back to Recovering syncing when we haven't sync for a while

### LINKED ISSUE #3935: SSS: Enter "recovering" mode if the app has been offline for "a while"
c.f. https://github.com/element-hq/element-x-ios/issues/3223#issuecomment-2326348104, cc. @Hywan 

This helps ensure that the next sliding sync request is responded to fast by the server. If we use a `pos` token from a long time ago with the full range, the server will need to respond with potentially lots and lots of updates, which might take a while

It'd be good to reduce the range back down to `[0-19]` after some time of inactivity (but *not* reset the connection, like we do when we see a connection error/timeout). This will then allow the server to (hopefully) respond quickly to the first request and for the client to fetch the rest of the updates in when the list grows. I don't think it really matters too much if we reduce to `[0,19]` relatively quickly, so I'd probably suggest a 1hr timer would be a good first try.

MatMaul:
Fixes #3935.

Untested in the apps for now, but integration test is included and working.

Signed-off-by: Mathieu Velten <mathieu.velten@beta.gouv.fr>

### DIFF
--- a/crates/matrix-sdk-ui/src/room_list_service/mod.rs
+++ b/crates/matrix-sdk-ui/src/room_list_service/mod.rs
@@ -60,7 +60,7 @@ mod state;
 use std::{sync::Arc, time::Duration};
 
 use async_stream::stream;
-use eyeball::{SharedObservable, Subscriber};
+use eyeball::Subscriber;
 use futures_util::{pin_mut, Stream, StreamExt};
 use matrix_sdk::{
     event_cache::EventCacheError, Client, Error as SlidingSyncError, SlidingSync, SlidingSyncList,
@@ -89,7 +89,7 @@ pub struct RoomListService {
     /// The current state of the `RoomListService`.
     ///
     /// `RoomListService` is a simple state-machine.
-    state: SharedObservable<State>,
+    state_machine: StateMachine,
 }
 
 impl RoomListService {
@@ -172,7 +172,7 @@ impl RoomListService {
         // Eagerly subscribe the event cache to sync responses.
         client.event_cache().subscribe()?;
 
-        Ok(Self { client, sliding_sync, state: SharedObservable::new(State::Init) })
+        Ok(Self { client, sliding_sync, state_machine: StateMachine::new() })
     }
 
     /// Start to sync the room list.
@@ -208,7 +208,7 @@ impl RoomListService {
                 debug!("Run a sync iteration");
 
                 // Calculate the next state, and run the associated actions.
-                let next_state = self.state.get().next(&self.sliding_sync).await?;
+                let next_state = self.state_machine.next(&self.sliding_sync).await?;
 
                 // Do the sync.
                 match sync.next().await {
@@ -217,7 +217,7 @@ impl RoomListService {
                         debug!(state = ?next_state, "New state");
 
                         // Update the state.
-                        self.state.set(next_state);
+                        self.state_machine.set(next_state);
 
                         yield Ok(());
                     }
@@ -227,7 +227,7 @@ impl RoomListService {
                         debug!(expected_state = ?next_state, "New state is an error");
 
                         let next_state = State::Error { from: Box::new(next_state) };
-                        self.state.set(next_state);
+                        self.state_machine.set(next_state);
 
                         yield Err(Error::SlidingSync(error));
 
@@ -239,7 +239,7 @@ impl RoomListService {
                         debug!(expected_state = ?next_state, "New state is a termination");
 
                         let next_state = State::Terminated { from: Box::new(next_state) };
-                        self.state.set(next_state);
+                        self.state_machine.set(next_state);
 
                         break;
                     }
@@ -286,8 +286,8 @@ impl RoomListService {
         // when the session is forced to expire, the state remains `Terminated`, thus
         // the actions aren't executed as expected. Consequently, let's update the
         // state.
-        if let State::Terminated { from } = self.state.get() {
-            self.state.set(State::Error { from });
+        if let State::Terminated { from } = self.state_machine.get() {
+            self.state_machine.set(State::Error { from });
         }
     }
 
@@ -341,7 +341,7 @@ impl RoomListService {
                     // Update the `current_state`.
                     current_state = next_state;
                 } else {
-                    // Something is broken with `self.state`. Let's stop this stream too.
+                    // Something is broken with the state. Let's stop this stream too.
                     break;
                 }
             }
@@ -355,7 +355,7 @@ impl RoomListService {
 
     /// Get a subscriber to the state.
     pub fn state(&self) -> Subscriber<State> {
-        self.state.subscribe()
+        self.state_machine.subscribe()
     }
 
     async fn list_for(&self, sliding_sync_list_name: &str) -> Result<RoomList, Error> {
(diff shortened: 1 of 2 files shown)


crates/matrix-sdk-ui/src/room_list_service/state.rs
```diff
@@ -57,10 +61,18 @@ impl State {
...
             SettingUp | Recovering => {
                 set_all_rooms_to_growing_sync_mode(sliding_sync).await?;
-                Running
+                Running { last_time: Instant::now() }
             }
 
-            Running => Running,
+            Running { last_time } => {
+                // We haven't sync for a while so we should go back to recovering
+                if last_time.elapsed().as_secs() > 1800 {
```
Hywan, line 69: Ideally, we would like to make this (`1800`) configurable because it can depend on the homeserver or how the user wants to use this API. Thoughts?

Hywan: If this is configurable, then we don't need `mock_instant`, nor the `testing` feature (we want to remove them, and this patch is adding them). If this is configurable then, we can simply set it to 1 sec, and have a `sleep` of 2 secs in the test. It can sound like a bit hacky but it's probably easier to maintain than having a new dependency and a new feature.

Hywan (commented):
Thank you for working on this. I appreciate.

I'm a bit annoyed by the introduction of the new `testing` feature: we are trying to remove then, and this PR adds a new one. I'm proposing a solution in my comment. Feel free to question my comment.

crates/matrix-sdk-ui/src/room_list_service/state.rs
MatMaul: I am not sure it will be used much but it doesn't cost much to make it a parameter, and if it helps removing some maintenance burden in the tests it's probably worth it.

Yeah what you propose for the test is a bit hacky but it works :tm: , and it feels fine at the integration level.

On it.

MatMaul: Some docs are still missing, I'll clean up later on in the week, probably from Berlin ;)

crates/matrix-sdk-ui/src/room_list_service/mod.rs
```diff
@@ -100,19 +102,29 @@ impl RoomListService {
     ///
     /// This won't start an encryption sync, and it's the user's responsibility
     /// to create one in this case using `EncryptionSync`.
-    pub async fn new(client: Client) -> Result<Self, Error> {
-        Self::new_internal(client, false).await
+    pub async fn new(
+        client: Client,
+        delay_before_recover: Option<Duration>,
```
Hywan, line 107: I'm not comfortable having `delay_before_recover` as an argument of the constructor. To me it feels like we need a builder.

Ideally, we would want to hardcode the value, and only expose a setter to change it when `#[cfg(test)]`. But since we want to change that via integration test, we would need a `testing` feature, which is something we don't want 🔁.

So, I _think_ we should move the specific tests (about this delay) inside this crate: they become a unit tests and we can use the `#[cfg(test)] fn set_delay_before_recover(…)` approach.

What do you think? It would remove the need for this argument in the constructor, it keeps the code as simple as before, we don't need to make it dynamic, it's a hardcoded value that can be overwritten for tests only. Thoughts?

MatMaul: That is a lot simpler indeed :) 

Done !

Hywan (commented):
Thanks! It's indeed much simpler now!

Can you rebase all your patches now? I think we should be good.

crates/matrix-sdk-ui/src/room_list_service/state.rs
```diff
@@ -46,21 +49,49 @@ pub enum State {
     Terminated { from: Box<State> },
 }
 
-impl State {
+const DEFAULT_DELAY_BEFORE_RECOVER: Duration = Duration::from_secs(1800);
+
+/// The state machine used to transition between the [`State`]s.
+#[derive(Clone, Debug, PartialEq)]
+pub struct StateMachine {
```
Hywan, line 56: Can you move `State` inside `StateMachine` please?

Hywan (commented):
Last bit and we are good!

crates/matrix-sdk-ui/src/room_list_service/state.rs
MatMaul: So that was my first try, and then I got lost in how to deal with observing the `State` and ownership trouble. I can try again but the change will probably be quite more invasive.

Hywan: Do you allow me to try please?

MatMaul: Let me push my current state. I am not far, but I broke a sync indicator test so I think I messed up something regarding observable.

MatMaul: Done !

MatMaul: Oupsss it seems to break a lot of complement crypto tests.

MatMaul: I think it was not the case before my (bad ^^) last refactor.

crates/matrix-sdk-ui/src/room_list_service/state.rs
```diff
@@ -81,9 +125,9 @@ impl State {
                     state => state.to_owned(),
                 }
             }
-        };
+        });
 
-        Ok(next_state)
+        Ok(self.get())
```
Hywan, line 130: This can be simplified as:

```rust
let next_state = match self.get() { … };

self.set(next_state);

Ok(next_state)
```

Hywan (commented):
We are close to something I really like, thanks! What's left on your side?

MatMaul:
> We are close to something I really like, thanks! What's left on your side?

Mainly that last refactor introduces an error somewhere because this change seems to break almost all complement crypto tests :sweat_smile: 

https://github.com/matrix-org/matrix-rust-sdk/actions/runs/11217272471/job/31178453216?pr=3995

I just try a revert to test and tests are back to green (outside of coverage).

I can't really figure out why this is happening so if you can have a look that would be neat.

crates/matrix-sdk-ui/src/room_list_service/state.rs
```diff
@@ -46,21 +50,61 @@ pub enum State {
...
-impl State {
+const DEFAULT_DELAY_BEFORE_RECOVER: Duration = Duration::from_secs(1800);
+
+/// The state machine used to transition between the [`State`]s.
+#[derive(Clone, Debug)]
+pub struct StateMachine {
+    /// The current state of the `RoomListService`.
+    current: SharedObservable<State>,
+
+    last_sync_date: Instant,
```
Hywan, line 61: This value is never updated. I'm working on it :-).

crates/matrix-sdk-ui/src/room_list_service/state.rs
```diff
@@ -83,6 +127,8 @@ impl State {
             }
         };
 
+        self.set(next_state.clone());
```
Hywan, line 130: This is the error that makes `test_sliding_sync_indicator` to fail, and possibly other tests. This method caculates the next state, but the state must be updated if and only if a sync has run. It's `RoomListService::sync` that is responsible to update the state.

Hywan:
I've pushed https://github.com/tchapgouv/matrix-rust-sdk/pull/2, it's a patch on your branch. If you merge it, I think this patch should pass all the tests.

Hywan (approved):
Now we are good! Thanks for having merged my patches and rebased everything!

PR #3998, opened 2024-09-16
chore: Update to `ruma/ruma`

Hywan:
It does what's on the tin.

### DIFF
--- a/crates/matrix-sdk-base/src/rooms/mod.rs
+++ b/crates/matrix-sdk-base/src/rooms/mod.rs
@@ -40,7 +40,7 @@ use ruma::{
         RedactedStateEventContent, StaticStateEventContent, SyncStateEvent,
     },
     room::RoomType,
-    EventId, OwnedUserId, RoomVersionId, UserId,
+    EventId, OwnedUserId, RoomVersionId,
 };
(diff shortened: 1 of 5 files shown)


crates/matrix-sdk-base/src/rooms/mod.rs
```diff
@@ -195,12 +195,11 @@ impl BaseRoomInfo {
                 let mut o_ev = o_ev.clone();
                 o_ev.content.set_created_ts_if_none(o_ev.origin_server_ts);
 
-                let Some(owned_user_id) = get_user_id_for_state_key(m.state_key()) else {
```
bnjbvr, line 198: this changes behavior, are you sure this new behavior is correct? does Ruma do the complicated parsing now? cc @toger5

Hywan: Yes it does, see https://github.com/ruma/ruma/pull/1889.

bnjbvr: Cool cool

bnjbvr (approved):
lgtm

toger5:
Thanks for the work on this!

PR #3999, opened 2024-09-16
chore(ui,ffi): Remove the `RoomList::entries` method

Hywan:
This method is now private inside `matrix_sdk_ui` and removed
from `matrix_sdk_ffi`. This method is returning a stream of rooms,
but updates on rooms won't update the stream (only new rooms
will be seen on the stream). Nobody uses it as far as I know, and
`entries_with_dynamic_adapters` is the real only API we want people
to use.


bnjbvr (approved):
lgtm (doc's CI step's failing)

PR #3983, opened 2024-09-11
test: Bye bye SS proxy, hello Synapse \o/

Hywan:
This patch removes the sliding sync proxy, and makes the `matrix-sdk-integration-testing` tests to run against Synapse with MSC4186 enabled.

The following tests from `matrix-sdk-integration-testing` have been disabled because they are not working with Synapse for the moment:

* `sliding_sync::room::test_room_avatar_group_conversation`
* `sliding_sync::room::test_room_notification_count`
* `sliding_sync::room::test_delayed_decryption_latest_event`
* `sliding_sync::room::test_room_info_notable_update_deduplication`
* `sliding_sync::room::test_room_preview`
* `sliding_sync::notification_client::test_notification`

---

* Addresses https://github.com/matrix-org/matrix-rust-sdk/issues/3698

PR #4000, opened 2024-09-16
notification client: use the membership state to match an invite

bnjbvr:
The notification client was relying on the SS proxy including the `event_id` of a stripped room membership event, to correctly match an invite. This is not the case anymore with SSS (in the spec, stripped events don't include an event_id, although I don't see any good reason why), so we have to resort to a different mechanism:

- stash an invite event for the current user, if found.
- use it, if and only if, the room was inferred by the notification client to be in the invited state, based on the sync.

With this change, `matrix-sdk-integration-testing tests::sliding_sync::notification_client::test_notification` integration test from #3983 passes correctly.

### DIFF
--- a/crates/matrix-sdk-ui/src/notification_client.rs
+++ b/crates/matrix-sdk-ui/src/notification_client.rs
@@ -29,7 +29,10 @@ use ruma::{
     assign,
     directory::RoomTypeFilter,
     events::{
-        room::{member::StrippedRoomMemberEvent, message::SyncRoomMessageEvent},
+        room::{
+            member::{MembershipState, StrippedRoomMemberEvent},
+            message::SyncRoomMessageEvent,
+        },
         AnyFullStateEventContent, AnyStateEvent, AnySyncMessageLikeEvent, AnySyncTimelineEvent,
         FullStateEventContent, StateEventType, TimelineEventType,
     },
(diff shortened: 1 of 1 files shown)


crates/matrix-sdk-ui/src/notification_client.rs
```diff
@@ -322,16 +325,32 @@ impl NotificationClient {
...
 
-        let cloned_notif = notification.clone();
+        let raw_invite = Arc::new(Mutex::new(None));
+
         let target_event_id = event_id.to_owned();
+        let user_id = self.client.user_id().unwrap().to_owned();
+        let handler_raw_invite = raw_invite.clone();
         let stripped_member_handler =
             self.client.add_event_handler(move |raw: Raw<StrippedRoomMemberEvent>| async move {
+                warn!("DBG found a raw stripped room member event");
```
Hywan: May I recommend `debug!` instead of `warn!("DBG…"`?

crates/matrix-sdk-ui/src/notification_client.rs
```diff
@@ -322,25 +337,57 @@ impl NotificationClient {
...
                     }
                     Err(err) => {
-                        warn!("a room member event id couldn't be decoded: {err}");
+                        debug!("a room member event id couldn't be decoded: {err}");
                     }
                 }
+
+                // Try to match the event by membership and state_key for the current user.
+                if deserialized.content.membership == MembershipState::Invite
+                    && deserialized.state_key == user_id
```
Hywan, line 381: I'm not sure it's gonna work with the latest version of Ruma. Either you transform `user_id` into the type of `state_key`, or you extract the user ID from `state_key` with `state_key.user_id()`.

bnjbvr: This is rebased on main, so should be fine.

bnjbvr: (It's not a call room member event, but a general room member event.)

Hywan: Alright.

Hywan (approved):

PR #4002, opened 2024-09-16
ffi: expose method to allow user account deactivate for `m.login.password` based accounts

stefanceriu:
Expose the Account method but restrict it to the types we were already using for user interactive authorization on the identity reset flows.

### DIFF
--- a/crates/matrix-sdk/src/account.rs
+++ b/crates/matrix-sdk/src/account.rs
@@ -360,6 +360,9 @@ impl Account {
     ///   information for the interactive auth and the same request needs to be
     ///   made but this time with some `auth_data` provided.
     ///
+    /// * `erase` - Whether the user would like their content to be erased as
+    ///   much as possible from the server.
+    ///
     /// # Examples
     ///
     /// ```no_run
@@ -376,7 +379,7 @@ impl Account {
     /// # let homeserver = Url::parse("http://localhost:8080")?;
     /// # let client = Client::new(homeserver).await?;
     /// # let account = client.account();
-    /// let response = account.deactivate(None, None).await;
+    /// let response = account.deactivate(None, None, false).await;
     ///
     /// // Proceed with UIAA.
     /// # anyhow::Ok(()) };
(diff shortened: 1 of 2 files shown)


crates/matrix-sdk/src/account.rs
```diff
@@ -388,10 +391,12 @@ impl Account {
         &self,
         id_server: Option<&str>,
         auth_data: Option<AuthData>,
+        erase_data: bool,
     ) -> Result<deactivate::v3::Response> {
         let request = assign!(deactivate::v3::Request::new(), {
             id_server: id_server.map(ToOwned::to_owned),
             auth: auth_data,
+            erase: erase_data,
```
Hywan, line 399: I'm surprised there is no test for this method? Can you check please?

Hywan (commented):
Looks good, although I would prefer to have a test with the `erase_data`, it seems pretty important to test :-].

crates/matrix-sdk/src/account.rs
stefanceriu: There's isn't one, no, but it's all just a network request.

stefanceriu: added tests in https://github.com/matrix-org/matrix-rust-sdk/pull/4002/commits/e414149d452b9496dec2a539365c723986b39cdc

crates/matrix-sdk/tests/integration/account.rs
```diff
@@ -0,0 +1,61 @@
...
+                }))
+            })
+            .expect(1)
+            .mount_as_scoped(&server)
+            .await;
+
+        assert_matches!(
+            client.account().deactivate(Some("FirstIdentityServer"), None, false).await,
+            Ok(..)
+        );
```
Hywan, line 40: You can replace it by:

```rust
assert!(client.account().deactivate(Some(…), None, false).await.is_ok())
```

(the trick is `.is_ok()`)

stefanceriu: Oh, yeah, that does make more sense 😁

Issue #3931, opened 2024-09-03
tests::timeline::test_stale_local_echo_time_abort_edit intermittently fails


bnjbvr:
e.g. https://github.com/matrix-org/matrix-rust-sdk/actions/runs/10670911032/job/29575806644

bnjbvr:
Failed on main today: https://github.com/matrix-org/matrix-rust-sdk/actions/runs/10881840181/job/30191636360

PR #4005, opened 2024-09-16
tests: increase timeout duration when awaiting a timeline update

bnjbvr:
Fixes https://github.com/matrix-org/matrix-rust-sdk/issues/3931, or so I think: the test waits for a stable state where there aren't any new updates to the timeline, before trying to edit a timeline item, and observe a precise sequence of timeline updates after the item has been edited.

The failures showed that the stable state wasn't reached, because in some occasions, we'd receive the original text as a new timeline update — which could also be the sign of a failing server / duplicated event or other things, but I believe it is just that 1 second timeouts ain't enough for code running under CodeCov.


andybalaam (approved):
I agree: we could easily see some kind of system slowdown that takes longer than 1 second, so this change is good even if it doesn't fix the problem.

PR #4013, opened 2024-09-17
feat(base): Increase the `room_info_notable_update_sender` capacity

Hywan:
This broadcast channel can easily be overflowed if more than 100 updates arrive
at the time. This patch extends the capacity to 2^16 - 1.

This patch also adds more logs to detect lagged on this channel from the room
list point of view.

### DIFF
--- a/crates/matrix-sdk-base/src/client.rs
+++ b/crates/matrix-sdk-base/src/client.rs
@@ -143,7 +143,7 @@ impl BaseClient {
     ///   previous login call.
     pub fn with_store_config(config: StoreConfig) -> Self {
         let (room_info_notable_update_sender, _room_info_notable_update_receiver) =
-            broadcast::channel(100);
+            broadcast::channel(u16::MAX as usize);
 
         BaseClient {
             store: Store::new(config.state_store),
--- a/crates/matrix-sdk-ui/src/room_list_service/room_list.rs
+++ b/crates/matrix-sdk-ui/src/room_list_service/room_list.rs
@@ -26,8 +26,11 @@ use matrix_sdk::{
     Client, SlidingSync, SlidingSyncList,
 };
 use matrix_sdk_base::RoomInfoNotableUpdate;
-use tokio::{select, sync::broadcast};
-use tracing::trace;
+use tokio::{
+    select,
+    sync::broadcast::{self, error::RecvError},
+};
+use tracing::{error, trace};
 
 use super::{
     filters::BoxedFilterFn,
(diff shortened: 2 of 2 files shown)


jmartinesp (approved):
Nice catch! LGTM except for the typo 😉 .

crates/matrix-sdk-base/src/client.rs
```diff
@@ -143,7 +143,7 @@ impl BaseClient {
     ///   previous login call.
     pub fn with_store_config(config: StoreConfig) -> Self {
         let (room_info_notable_update_sender, _room_info_notable_update_receiver) =
-            broadcast::channel(100);
+            broadcast::channel((u16::MAX - 1) as usize);
```
bnjbvr, line 146: nit: `u16::MAX` is already 2**16 - 1, so no need to subtract one :nerd_face:

bnjbvr (approved):
Thanks for adding logs too.

(Should we remove the code that's commented out? Would be fairly easy to add it back later anyways.)

crates/matrix-sdk-base/src/client.rs
poljar, line 146: > This patch extends the capacity to 2^16 - 1.

That's 2¹⁶ - 2 though: 

> The largest value that can be represented by this integer type (2¹⁶ − 1). [^1]

[^1]: https://doc.rust-lang.org/std/primitive.u16.html#associatedconstant.MAX

poljar (approved):
:stuck_out_tongue_closed_eyes:

crates/matrix-sdk-base/src/client.rs
Hywan: Done.

Hywan:
I've removed the comments as suggested by @bnjbvr and I've used `u16::MAX` directly, without substrating 1.

stefanceriu (approved):

crates/matrix-sdk-ui/src/room_list_service/room_list.rs
```diff
@@ -215,32 +218,25 @@ fn merge_stream_and_receiver(
...
+                            if let Some(index) = raw_current_values.iter().position(|room| room.room_id() == update.room_id) {
+                                let room = &raw_current_values[index];
+                                let update = VectorDiff::Set { index, value: room.clone() };
+                                yield vec![update];
+                            }
+                        }
+
+                        Err(RecvError::Closed) => {
+                            error!("Cannot receive room info notable updates because the sender has been closed");
+                        }
```
bnjbvr, line 234: hacktualley: if this stream has closed, then this infinite loop will iloop on this case, right? Maybe we should return at this point?

Hywan: I'm adding a little `break` then. Good catch. It would have been caught previously, but not anymore.

PR #4014, opened 2024-09-17
ui: Allow to subscribe to read receipt changes in timeline metadata

zecakeh:
This is to fix a bug caused by a race condition in Fractal. We used an event handler to detect changes in our own read receipts, then we called methods like [`Timeline::latest_user_read_receipt_timeline_event_id()`](https://matrix-org.github.io/matrix-rust-sdk/matrix_sdk_ui/timeline/struct.Timeline.html#method.latest_user_read_receipt_timeline_event_id). The problem is that there is no guarantee that the metadata of the timeline has already been updated when we call the method since they happen in separate threads. So instead we use this stream to react when the metadata is updated.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -69,7 +69,7 @@ sha2 = "0.10.8"
 similar-asserts = "1.5.0"
 stream_assert = "0.1.1"
 thiserror = "1.0.38"
-tokio = { version = "1.30.0", default-features = false, features = ["sync"] }
+tokio = { version = "1.39.1", default-features = false, features = ["sync"] }
 tokio-stream = "0.1.14"
 tracing = { version = "0.1.40", default-features = false, features = ["std"] }
 tracing-core = "0.1.32"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1150,7 +1150,7 @@ dependencies = [
  "bitflags 2.6.0",
  "crossterm_winapi",
  "libc",
- "mio",
+ "mio 0.8.11",
  "parking_lot",
  "signal-hook",
  "signal-hook-mio",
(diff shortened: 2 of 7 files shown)


Cargo.toml
```diff
@@ -69,7 +69,7 @@ sha2 = "0.10.8"
 similar-asserts = "1.5.0"
 stream_assert = "0.1.1"
 thiserror = "1.0.38"
-tokio = { version = "1.30.0", default-features = false, features = ["sync"] }
+tokio = { version = "1.39.1", default-features = false, features = ["sync"] }
```
zecakeh, line 72: This change is necessary, otherwise `tokio::sync::watch::Sender` does not implement `Default`.

bnjbvr:
> This is to fix a bug caused by a race condition in Fractal.

So should this be a fix in Fractal, instead? I'm not sure we want to pay the price of supporting a new API just for a bug that's happening in Fractal, especially since 1. the read receipts code is already quite complicated in the timeline API, 2. it should be moved over to the event cache eventually, so it'd be ideal to not support app-specific features in there. If you can make the argument why this feature would be useful in general, and not only for Fractal, then I'm all for it, otherwise I'd rather not add an extra maintenance burden onto ourselves.

zecakeh:
We use it to compute if there are unread messages in the room, because `Room::num_unread_messages` doesn't work with `/v2/sync` so it might only be needed temporarily.

On the other end, I don't see methods like `Timeline::latest_user_read_receipt()` and `Timeline::latest_user_read_receipt_timeline_event_id()` being useful without this API, because imo it makes sense to be able to call these functions after the read receipts changed. There is nothing to "fix" in Fractal, the current bug we have is because of a shortcoming in the current API.

Usually a "bug in Fractal" is a bug that can be encountered by a client that uses the current stable Matrix APIs.

poljar (approved):
I think that this is sufficiently simple that we can take it, besides as far as I understood this is needed for clients using sync v2.

PR #4016, opened 2024-09-17
client builder: return a `ClientBuildError` when failing to build, instead of filtering out unexpected errors

reivilibre:
This old method was checking invariants that were
spooky-action-at-a-distance: these invariants have changed since then,
so this would panic instead of returning a proper error to the caller.

Signed-off-by: oliverw@element.io

### DIFF
--- a/crates/matrix-sdk/src/client/builder/mod.rs
+++ b/crates/matrix-sdk/src/client/builder/mod.rs
@@ -709,19 +709,6 @@ pub enum ClientBuildError {
     SqliteStore(#[from] matrix_sdk_sqlite::OpenStoreError),
 }
 
-impl ClientBuildError {
-    /// Assert that a valid homeserver URL was given to the builder and no other
-    /// invalid options were specified, which means the only possible error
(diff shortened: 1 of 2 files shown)


bnjbvr (dismissed):
Thanks for the PR. It's bad that you can hit this panic, I suppose it was put here as an internal-API error, and never supposed to be triggerable by users of the SDK. I think the code has evolved sufficiently that it's a bad idea for us to use `assert_valid_builder_args` at all: the error could also be a store error, or a discovery error, or, or, or…

So instead, I think we should:

- get rid of `assert_valid_builder_args`
- have `Client::new` return a `Result<Self, ClientBuildError>`.

Are you willing to make the changes here? Otherwise one of us can probably get to it at some point.

reivilibre:
I can have a go at that soon, will have to happen after the mad rush to get this other stuff done that led to this issue.

BTW it was a `SqliteStore(MissingVersion)`.... :shrug:

stefanceriu (approved):
Makes sense to me 👍

Issue #4009, opened 2024-09-16
FFI bindings crash on iOS in RoomListItem.initTimeline when built in debug mode


Johennes:
I'm hitting a weird `EXC_BAD_ACCESS` with the FFI bindings on iOS. I'm using the following example code:

```
import UIKit
import AuthenticationServices

var client: Client! = nil
var ssoHandler: SsoHandler! = nil
var syncService: SyncService! = nil
var listener: AllRoomsListener! = nil
var roomListService: RoomListService! = nil
var handle: TaskHandle! = nil

var asSession: ASWebAuthenticationSession! = nil

class ViewController: UIViewController {

    override func viewDidLoad() {
        super.viewDidLoad()
    
        Task {
            try Thread.sleep(forTimeInterval: 2)
            
            client = try! await ClientBuilder()
                .serverNameOrHomeserverUrl(serverNameOrUrl: "https://chat.dev.unomed.ch")
                .sessionPaths(dataPath: URL.applicationSupportDirectory.path(percentEncoded: false).appending("/").appending(UUID().uuidString),
                              cachePath: URL.cachesDirectory.path(percentEncoded: false).appending("/").appending(UUID().uuidString))
                .slidingSyncVersionBuilder(versionBuilder: .proxy(url: "https://ssync.dev.unomed.ch"))
                .build()
            
            ssoHandler = try! await client.startSsoLogin(redirectUrl: "test.matrix://Main", idpId: nil)
            
            let url = URL(string: ssoHandler.url())!
            asSession = ASWebAuthenticationSession(url: url, callbackURLScheme: "test.matrix") { callbackURL, error in
                Task {
                    try! await ssoHandler.finish(callbackUrl: callbackURL!.absoluteString)
                    
                    let session = try client.session()
                    print(session)
        
                    // Create a sync service which controls the sync loop.
                    syncService = try await client.syncService().finish()
        
                    // Listen to room list updates.
                    listener = AllRoomsListener()
                    roomListService = syncService.roomListService()
                    handle = try await roomListService.allRooms().entries(listener: listener)
        
                    // Start the sync loop.
                    await syncService.start()
                }
            }
            asSession.presentationContextProvider = self
            asSession.start()
        }
    }
}

extension ViewController: ASWebAuthenticationPresentationContextProviding {
    func presentationAnchor(for session: ASWebAuthenticationSession) -> ASPresentationAnchor {
        return view.window!
    }
}

[...]

PR #4020, opened 2024-09-18
ffi: default to reldbg when building iOS bindings

Johennes:
Relates to: #4009


jmartinesp (approved):
Given @stefanceriu 's comments I think it make sense, but I'm assigning him as a reviewer too to double check.

stefanceriu (approved):
Yup yup, thanks Johannes! 🚀

PR #4021, opened 2024-09-19
sdk-ui: make room encryption optional to create a timeline

jmartinesp:
Instead of forcing the room encryption to be known when the timeline is created and failing if it's not known, take the latest room encryption info as a base value and update it when processing timeline events.

At the time of writing this PR, the encryption info is only used to decide whether shields should be calculated for timeline items or not.

Fixes https://github.com/matrix-org/matrix-rust-sdk/issues/3850 (I hope).

Signed-off-by:

### DIFF
--- a/crates/matrix-sdk-ui/src/timeline/controller/mod.rs
+++ b/crates/matrix-sdk-ui/src/timeline/controller/mod.rs
@@ -240,7 +240,7 @@ impl<P: RoomDataProvider> TimelineController<P> {
         focus: TimelineFocus,
         internal_id_prefix: Option<String>,
         unable_to_decrypt_hook: Option<Arc<UtdHookManager>>,
-        is_room_encrypted: bool,
+        is_room_encrypted: Option<bool>,
     ) -> Self {
         let (focus_data, focus_kind) = match focus {
             TimelineFocus::Live => (TimelineFocusData::Live, TimelineFocusKind::Live),
@@ -341,6 +341,34 @@ impl<P: RoomDataProvider> TimelineController<P> {
                 .await;
 
                 Ok(has_events)
+            }
+        }
+    }
+
+    /// Listens to encryption state changes for the room in
+    /// [`matrix_sdk_base::RoomInfo`] and applies the new value to the
+    /// existing timeline items. This will then cause a refresh of those
+    /// timeline items.
+    pub async fn handle_encryption_state_changes(&self) {
+        let mut room_info = self.room_data_provider.room_info();
+
+        while let Some(info) = room_info.next().await {
+            let changed = {
+                let state = self.state.read().await;
+                let mut old_is_room_encrypted = state.meta.is_room_encrypted.write().unwrap();
+                let is_encrypted_now = info.is_encrypted();
+
+                if *old_is_room_encrypted != Some(is_encrypted_now) {
+                    *old_is_room_encrypted = Some(is_encrypted_now);
+                    true
+                } else {
+                    false
+                }
+            };
+
+            if changed {
+                let mut state = self.state.write().await;
+                state.update_all_events_is_room_encrypted();
             }
         }
     }
--- a/crates/matrix-sdk-ui/src/timeline/controller/state.rs
+++ b/crates/matrix-sdk-ui/src/timeline/controller/state.rs
@@ -12,7 +12,12 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use std::{collections::VecDeque, future::Future, num::NonZeroUsize, sync::Arc};
+use std::{
+    collections::VecDeque,
+    future::Future,
+    num::NonZeroUsize,
+    sync::{Arc, RwLock},
+};
 
 use eyeball_im::{ObservableVector, ObservableVectorTransaction, ObservableVectorTransactionEntry};
 use itertools::Itertools as _;
@@ -83,7 +88,7 @@ impl TimelineState {
         room_version: RoomVersionId,
         internal_id_prefix: Option<String>,
         unable_to_decrypt_hook: Option<Arc<UtdHookManager>>,
-        is_room_encrypted: bool,
+        is_room_encrypted: Option<bool>,
     ) -> Self {
         Self {
             // Upstream default capacity is currently 16, which is making
@@ -293,6 +298,16 @@ impl TimelineState {
             txn.add_remote_events_at(events, position, origin, room_data_provider, settings).await;
         txn.commit();
         result
+    }
+
+    pub(super) fn update_all_events_is_room_encrypted(&mut self) {
+        let is_room_encrypted = *self.meta.is_room_encrypted.read().unwrap();
+
+        // When this transaction finishes, all items in the timeline will be emitted
+        // again with the updated encryption value
+        let mut txn = self.transaction();
+        txn.update_all_events_is_room_encrypted(is_room_encrypted);
+        txn.commit();
     }
 
     pub(super) fn transaction(&mut self) -> TimelineStateTransaction<'_> {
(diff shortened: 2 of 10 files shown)


crates/matrix-sdk-ui/src/timeline/controller/mod.rs
```diff
@@ -345,6 +345,29 @@ impl<P: RoomDataProvider> TimelineController<P> {
...
+                self.state.read().await.meta.is_room_encrypted.write()
+            {
+                let is_encrypted_now = info.is_encrypted();
+                if is_room_encrypted.is_none() || is_room_encrypted.unwrap() != is_encrypted_now {
+                    changed = true;
+                    *is_room_encrypted = Some(is_encrypted_now);
+                }
+            }
+            if changed {
+                let mut state = self.state.write().await;
```
poljar, line 365: Won't this deadlock? `self.state.read()`, while it doesn't have a binding, will only be dropped after the end of the block. Trying to acquire the write lock won't eve succeed since a read lock is held till the end of the block.

crates/matrix-sdk-ui/src/timeline/controller/mod.rs
```diff
@@ -345,6 +345,29 @@ impl<P: RoomDataProvider> TimelineController<P> {
...
+    /// existing timeline items.
+    pub async fn handle_encryption_state_changes(&self) {
+        let mut room_info = self.room_data_provider.room_info();
+        while let Some(info) = room_info.next().await {
+            let mut changed = false;
+            if let Ok(mut is_room_encrypted) =
+                self.state.read().await.meta.is_room_encrypted.write()
+            {
+                let is_encrypted_now = info.is_encrypted();
+                if is_room_encrypted.is_none() || is_room_encrypted.unwrap() != is_encrypted_now {
```
poljar, line 359:
```suggestion
                if *is_room_encrypted != Some(is_encrypted_now) {
```

crates/matrix-sdk-ui/src/timeline/controller/state.rs
```diff
@@ -295,6 +300,19 @@ impl TimelineState {
         result
     }
 
+    pub(super) fn reload_shields(&mut self) {
+        let Ok(is_room_encrypted) = self.meta.is_room_encrypted.read() else {
```
poljar, line 304: We generally `unwrap()` if a lock is poisoned, it's the only place where we allow unwraps.

crates/matrix-sdk-ui/src/timeline/controller/state.rs
```diff
@@ -295,6 +300,19 @@ impl TimelineState {
...
+            return;
+        };
+
+        if let Some(is_encrypted) = *is_room_encrypted {
```
poljar, line 308: Don't we need to do this only if it's `Some()` as well as `true`? On the other hand, the only place where we call this is in a place where we know the value of this and if it'll be true, why don't we just unconditionally call `replace_all_events_encryption()`?

crates/matrix-sdk-ui/src/timeline/controller/state.rs
```diff
@@ -720,6 +738,18 @@ impl TimelineStateTransaction<'_> {
     fn adjust_day_dividers(&mut self, mut adjuster: DayDividerAdjuster) {
         adjuster.run(&mut self.items, &mut self.meta);
     }
+
+    fn replace_all_events_encryption(&mut self, is_encrypted: bool) {
+        for idx in 0..self.items.len() {
+            let item = &self.items[idx];
```
poljar, line 744:
```suggestion
            let item = &self.items[idx];
            
```

crates/matrix-sdk-ui/src/timeline/controller/state.rs
```diff
@@ -720,6 +738,18 @@ impl TimelineStateTransaction<'_> {
     fn adjust_day_dividers(&mut self, mut adjuster: DayDividerAdjuster) {
         adjuster.run(&mut self.items, &mut self.meta);
     }
+
+    fn replace_all_events_encryption(&mut self, is_encrypted: bool) {
```
poljar, line 742: That's probably a really poor name for this method, at least I can't really explain from the name what this method does.

Nor is it quite clear that this method does enough, it's just setting a flag on the event that the room is encrypted? How will this trigger shield replacements?

crates/matrix-sdk-ui/src/timeline/controller/state.rs
```diff
@@ -720,6 +738,18 @@ impl TimelineStateTransaction<'_> {
...
+        for idx in 0..self.items.len() {
+            let item = &self.items[idx];
+            if let Some(event) = item.as_event() {
+                let mut cloned_event = event.clone();
+                cloned_event.is_room_encrypted = Some(is_encrypted);
```
poljar, line 747:
```suggestion
                cloned_event.is_room_encrypted = Some(is_encrypted);
                
```

crates/matrix-sdk-ui/src/timeline/controller/mod.rs
```diff
@@ -345,6 +345,29 @@ impl<P: RoomDataProvider> TimelineController<P> {
         }
     }
 
+    /// Listens to encryption state changes for the room in
+    /// [`matrix_sdk_base::RoomInfo`] and applies the new value to the
```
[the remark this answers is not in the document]
jmartinesp: Technically this function isn't recalculating the shield states: that is created on the fly every time we call the `fn EventTimelineItem::get_shield`, based on both `is_room_encrypted` and the remote event's encryption info.

crates/matrix-sdk-ui/src/timeline/controller/mod.rs
jmartinesp: The tests pass and they do run this code, so it must be working, although I don't really like how this code looks and I've been trying to improve it, but I'm not sure how to achieve that. Should I keep a reference, then manually drop it, like I did somewhere else?

crates/matrix-sdk-ui/src/timeline/controller/state.rs
jmartinesp: I can add some docs for it and maybe find a new name, but it's kind of difficult, since all it does is replace the `is_room_encrypted` value for each timeline item and create a `VectorDiff::Set` for each item so this update is emitted.

crates/matrix-sdk-ui/src/timeline/controller/mod.rs
jmartinesp: Ok, I think I found a better way to handle this.

jmartinesp:
I tried following your suggestions in https://github.com/matrix-org/matrix-rust-sdk/pull/4021/commits/71a8e24536674b4af7aa8730d895c557c6e28b23 🤞 .

crates/matrix-sdk-ui/src/timeline/controller/mod.rs
poljar: The way you handled it would have been my recommendation as well.

crates/matrix-sdk-ui/src/timeline/controller/mod.rs
```diff
@@ -345,6 +345,33 @@ impl<P: RoomDataProvider> TimelineController<P> {
...
+    /// existing timeline items. This will then cause a refresh of those
+    /// timeline items.
+    pub async fn handle_encryption_state_changes(&self) {
+        let mut room_info = self.room_data_provider.room_info();
+
+        while let Some(info) = room_info.next().await {
+            let changed = {
+                let state = self.state.read().await;
+                let mut old_is_room_encrypted = state.meta.is_room_encrypted.write().unwrap();
+                let is_encrypted_now = info.is_encrypted();
```
poljar, line 359:
```suggestion
                let is_encrypted_now = info.is_encrypted();
                
```

poljar (approved):
Left a small nit, I think that this is fine now.

PR #4023, opened 2024-09-20
Add nix flake to build "examples/command_bot"

6543:
this makes it easy to make the build work on all linux/unix systems that have nix installed.


andybalaam:
Hi, thanks for the contribution, but we are not able to accept it here, because we don't have the expertise or time to support Nix packaging. We suggest you start a separate repo for this.

Thanks again!

PR #4026, opened 2024-09-23
chore(ruma): Update Ruma to support negative `bump_stamp`

Hywan:
Blocked by https://github.com/ruma/ruma/pull/1912.

This is a draft patch to check we can support negative `bump_stamp` from Synapse.


Hywan:
I believe the homeserver must fix the issue instead :-).

PR #4031, opened 2024-09-25
test(integration): Enable `test_room_notification_count`

Hywan:
This test was failing since the migration from the sliding sync proxy
to Synapse.

This patch fixes the test. The failing parts were:

1. The `timeline_limit` wasn't set, so Synapse was returning an error,
2. The `unread_notifications` was set to 0 and could not be set to 1
   because that's an encrypted room.

The fact `timeline_limit` is now mandatory has been mentioned in the MSC:
https://github.com/matrix-org/matrix-spec-proposals/pull/4186/files#r1775138458
A patch in Ruma has been created. The previous patch in this repository
also contains the fix for the SDK side.

The assertions around `unread_notifications` have been removed. We no
longer use this API anymore (and it should be deprecated by the way).

### DIFF
--- a/crates/matrix-sdk/src/sliding_sync/list/builder.rs
+++ b/crates/matrix-sdk/src/sliding_sync/list/builder.rs
@@ -37,7 +37,7 @@ pub struct SlidingSyncListBuilder {
     required_state: Vec<(StateEventType, String)>,
     include_heroes: Option<bool>,
     filters: Option<http::request::ListFilters>,
-    timeline_limit: Option<Bound>,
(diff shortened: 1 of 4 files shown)


stefanceriu (approved):
Makes sense to me 👍

jmartinesp (approved):
> The assertions around unread_notifications have been removed. We no
longer use this API anymore (and it should be deprecated by the way).

LGTM! I was going to ask about this, but it's ok if we're just going to remove it in the near future.

PR #4032, opened 2024-09-25
test(integration): Enable `test_room_preview` and `test_room_avatar_group_conversation`

Hywan:
These tests were failing since the migration from the sliding sync proxy to Synapse.

Since the previous fixes to re-enable other tests, these 2 passes for free.


stefanceriu (approved):
Yey, noice! 👏

PR #4038, opened 2024-09-26
ffi: use fork of `tracing` crate with a fix for Android logs rotation

jmartinesp:
The `tracing` library has [an issue](https://github.com/tokio-rs/tracing/issues/2937) that prevents the configured log rotation from working on Android. While there has been [a proposed fix](https://github.com/tokio-rs/tracing/pull/3000) for some time, it hasn't been merged yet.

We recently discovered some cases of users who had hundreds of log files, several times our current max limit, and today I confirmed old logs weren't being deleted.

My proposed solution is to use [a forked tracing library](https://github.com/element-hq/tracing) in the meantime, along with [a forked paranoid-android one](https://github.com/element-hq/paranoid-android), which needs to use this former fork.

I tested this on an Android client and I saw the old logs being removed, as expected.

Signed-off-by:

### DIFF
--- a/bindings/matrix-sdk-ffi/Cargo.toml
+++ b/bindings/matrix-sdk-ffi/Cargo.toml
@@ -82,3 +82,10 @@ features = [
 
 [lints]
 workspace = true
+
+[patch.crates-io]
+tracing = { git = "https://github.com/element-hq/tracing.git", rev = "ca9431f74d37c9d3b5e6a9f35b2c706711dab7dd", default-features = false, features = ["std"] }
+tracing-core = { git = "https://github.com/element-hq/tracing.git", rev = "ca9431f74d37c9d3b5e6a9f35b2c706711dab7dd" }
+tracing-subscriber = { git = "https://github.com/element-hq/tracing.git", rev = "ca9431f74d37c9d3b5e6a9f35b2c706711dab7dd" }
+tracing-appender = { git = "https://github.com/element-hq/tracing.git", rev = "ca9431f74d37c9d3b5e6a9f35b2c706711dab7dd" }
+paranoid-android = { git = "https://github.com/element-hq/paranoid-android.git", rev = "69388ac5b4afeed7be4401c70ce17f6d9a2cf19b" }


poljar (changes requested):
Please use the [patch](https://doc.rust-lang.org/cargo/reference/overriding-dependencies.html#the-patch-section) section for this like we already do for some other things: https://github.com/matrix-org/matrix-rust-sdk/blob/322c5b3f830ae4967ff07b0e475b37fd4ab5f963/Cargo.toml#L122-L124.

Since this is only needed for the bindings, there's no need to change the dependency definition. This way we'll be able to publish releases, while Android and the bindings, which don't need to be published on crates.io, can use the fork.

jmartinesp:
I think it should be ok after https://github.com/matrix-org/matrix-rust-sdk/pull/4038/commits/e7f51352d7d32b8b7ce1c09d9d09caa6aefefb73

bindings/matrix-sdk-ffi/Cargo.toml
```diff
@@ -82,3 +82,10 @@ features = [
 
 [lints]
 workspace = true
+
+[patch.crates-io]
+tracing = { git = "https://github.com/element-hq/tracing.git", rev = "ca9431f74d37c9d3b5e6a9f35b2c706711dab7dd", default-features = false, features = ["std"] }
+tracing-core = { git = "https://github.com/element-hq/tracing.git", rev = "ca9431f74d37c9d3b5e6a9f35b2c706711dab7dd" }
+tracing-subscriber = { git = "https://github.com/element-hq/tracing.git", rev = "ca9431f74d37c9d3b5e6a9f35b2c706711dab7dd" }
+tracing-appender = { git = "https://github.com/element-hq/tracing.git", rev = "ca9431f74d37c9d3b5e6a9f35b2c706711dab7dd" }
+paranoid-android = { git = "https://github.com/element-hq/paranoid-android.git", rev = "69388ac5b4afeed7be4401c70ce17f6d9a2cf19b" }
```
poljar, line 91: There's a newline missing here at the end.

poljar (approved):

PR #4047, opened 2024-09-30
base: Expose room avatar info from RoomInfo

zecakeh:
In Fractal, we try to optimize our image requests according to the image info that was provided, e.g. we don't request thumbnails but the original image for SVGs or if the original is small. Currently it means we need to request the `m.room.avatar` event from the store to get this info although it is already available in the `RoomInfo` for `/v3/sync`. This exposes that data from `RoomInfo`.


stefanceriu (approved):
Makes sense to me 👌

PR #4049, opened 2024-09-30
misc: Remove deprecated apis

zecakeh:
The APIs have been deprecated for a year and a half, and the deprecations were released in 0.7.0.

### DIFF
--- a/crates/matrix-sdk-base/src/store/integration_tests.rs
+++ b/crates/matrix-sdk-base/src/store/integration_tests.rs
@@ -71,8 +71,6 @@ pub trait StateStoreIntegrationTests {
     async fn test_receipts_saving(&self);
     /// Test custom storage.
     async fn test_custom_storage(&self) -> Result<()>;
-    /// Test invited room saving.
-    async fn test_persist_invited_room(&self) -> Result<()>;
     /// Test stripped and non-stripped room member saving.
     async fn test_stripped_non_stripped(&self) -> Result<()>;
     /// Test room removal.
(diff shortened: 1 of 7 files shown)


jmartinesp (approved):
Thanks for the cleanup!

crates/matrix-sdk-base/src/store/integration_tests.rs
```diff
@@ -918,25 +913,12 @@ impl StateStoreIntegrationTests for DynStateStore {
         Ok(())
     }
 
-    async fn test_persist_invited_room(&self) -> Result<()> {
```
bnjbvr, line 921: Are we not persisting invited rooms anymore? :sweat_smile: 
Can you use the non-deprecated method in this test, instead (or explain why it's totally fine to get rid of this test — because this is tested elsewhere)?

crates/matrix-sdk-base/src/store/integration_tests.rs
```diff
@@ -918,25 +913,12 @@ impl StateStoreIntegrationTests for DynStateStore {
...
-    }
-
     async fn test_stripped_non_stripped(&self) -> Result<()> {
         let room_id = room_id!("!test_stripped_non_stripped:localhost");
         let user_id = user_id();
 
         assert!(self.get_member_event(room_id, user_id).await.unwrap().is_none());
         assert_eq!(self.get_room_infos().await.unwrap().len(), 0);
-        #[allow(deprecated)]
-        let stripped_rooms = self.get_stripped_room_infos().await?;
```
bnjbvr, line 938: ditto for this function call, and the other below

bnjbvr (changes requested):

crates/matrix-sdk-base/src/store/integration_tests.rs
zecakeh: To be honest, I am not sure what was the purpose of this test. It just calls `populate` then checks the number of stripped room infos, and we already did that in `test_populate_store`.

zecakeh: And the reason why its fine to remove the calls to `get_stripped_room_infos` is that stripped room are included with other results in `get_room_infos`. It used to not be the case a long time ago, but when it was changed this method was deprecated and unused but we kept the tests for it.

bnjbvr (approved):
Cool, thanks for explaining!

PR #4052, opened 2024-09-30
base: Apply `RoomInfo` migrations for `notable_tags` and `pinned_events`

zecakeh:
Because those fields were added in the last months, a user that was already logged-in will have invalid data for those fields unless we fetch it from the state store.


poljar (approved):
Looks good, thanks.

PR #4060, opened 2024-10-01
fix: deadlock between `bootstrap_cross_signing` and `sync`

boxdot:
`bootstrap_cross_signing` holds a lock on the private identity. In case a new identity is created, it will try to acquire a lock on `account`. The latter is locked by `sync`, which tries to acquire a lock on the private identity.

Note that the `bootstrap_cross_signing` call is spawned in a separate task e.g. in `restore_session`. In particular, this task and `sync` both race to acquire locks described above.

### DIFF
--- a/crates/matrix-sdk-crypto/src/machine/mod.rs
+++ b/crates/matrix-sdk-crypto/src/machine/mod.rs
@@ -642,19 +642,20 @@ impl OlmMachine {
         &self,
         reset: bool,
     ) -> StoreResult<CrossSigningBootstrapRequests> {
-        let mut identity = self.inner.user_identity.lock().await;
+        // Don't hold the lock, otherwise we might deadlock in `bootstrap_cross_signing()` on
+        // `account` if a sync task is already running (which locks `account`), or we will deadlock
+        // in `upload_device_keys()` which lock private identity again.
+        let identity = self.inner.user_identity.lock().await.clone();
 
         let (upload_signing_keys_req, upload_signatures_req) = if reset || identity.is_empty().await
         {
             info!("Creating new cross signing identity");
 
-            let (new_identity, upload_signing_keys_req, upload_signatures_req) = {
+            let (identity, upload_signing_keys_req, upload_signatures_req) = {
                 let cache = self.inner.store.cache().await?;
                 let account = cache.account().await?;
                 account.bootstrap_cross_signing().await
             };
-
-            *identity = new_identity;
 
             let public = identity.to_public_identity().await.expect(
                 "Couldn't create a public version of the identity from a new private identity",
@@ -681,11 +682,6 @@ impl OlmMachine {
 
             (upload_signing_keys_req, upload_signatures_req)
         };
-
-        // `upload_device_keys()` will attempt to sign the device keys using this
-        // `identity`, it will attempt to acquire the lock, so we need to drop
-        // here to avoid a deadlock.
-        drop(identity);
 
         // If there are any *device* keys to upload (i.e. the account isn't shared),
         // upload them before we upload the signatures, since the signatures may


andybalaam:
@poljar I'm out of depth here: please could you review?

crates/matrix-sdk-crypto/src/machine/mod.rs
```diff
@@ -642,28 +642,32 @@ impl OlmMachine {
         &self,
         reset: bool,
     ) -> StoreResult<CrossSigningBootstrapRequests> {
-        let mut identity = self.inner.user_identity.lock().await;
+        // Don't hold the lock, otherwise we might deadlock in
```
poljar, line 645: I'm not quite sure I understand what's happening here.

We drop the lock on the identity before we call `upload_device_keys()`, so it can't be that we deadlock because of that.

Is it perhaps because we can't acquire the lock on the account to call `bootstrap_cross_singing()`?

If so, then we should, instead of not holding the lock here, acquire the account lock first. The lock on the identity also servers to make this method idempotent. Which kinda was ruined by the `drop()` call bellow, but this would make it even worse.

poljar (changes requested):
I think an alternative approach might be better here, let me know if that solves the deadlock as well.

crates/matrix-sdk-crypto/src/machine/mod.rs
boxdot: Here are the backtraces from two places locking the account.

```
  <Holding accounts lock from sync>
  0: matrix_sdk_crypto::store::StoreCache::account::{{closure}}
            at src/matrix-rust-sdk/crates/matrix-sdk-crypto/src/store/mod.rs:371:25
  1: matrix_sdk_crypto::store::StoreCacheGuard::account::{{closure}}
            at src/matrix-rust-sdk/crates/matrix-sdk-crypto/src/store/mod.rs:409:30
  2: matrix_sdk_crypto::machine::OlmMachine::outgoing_requests::{{closure}}
            at src/matrix-rust-sdk/crates/matrix-sdk-crypto/src/machine/mod.rs:513:49
  3: matrix_sdk::encryption::<impl matrix_sdk::client::Client>::send_outgoing_requests::{{closure}}
            at src/matrix-rust-sdk/crates/matrix-sdk/src/encryption/mod.rs:723:18
  4: matrix_sdk::client::Client::sync_once::{{closure}}::{{closure}}
            at src/matrix-rust-sdk/crates/matrix-sdk/src/client/mod.rs:1821:55
  5: <tracing::instrument::Instrumented<T> as core::future::future::Future>::poll
            at .cargo/registry/src/index.crates.io-6f17d22bba15001f/tracing-0.1.40/src/instrument.rs:321:9
  6: matrix_sdk::client::Client::sync_once::{{closure}}
            at src/matrix-rust-sdk/crates/matrix-sdk/src/client/mod.rs:1809:5
  7: matrix_sdk::sync::<impl matrix_sdk::client::Client>::sync_loop_helper::{{closure}}
            at src/matrix-rust-sdk/crates/matrix-sdk/src/sync.rs:283:62
  8: matrix_sdk::client::Client::sync_with_result_callback::{{closure}}::{{closure}}
            at src/matrix-rust-sdk/crates/matrix-sdk/src/client/mod.rs:2066:68

<Holding account lock from bootstrap_cross_signing>
   0: matrix_sdk_crypto::store::StoreCache::account::{{closure}}
             at src/matrix-rust-sdk/crates/matrix-sdk-crypto/src/store/mod.rs:371:25
   1: matrix_sdk_crypto::store::StoreCacheGuard::account::{{closure}}
             at src/matrix-rust-sdk/crates/matrix-sdk-crypto/src/store/mod.rs:409:30
   2: matrix_sdk_crypto::machine::OlmMachine::bootstrap_cross_signing::{{closure}}
             at src/matrix-rust-sdk/crates/matrix-sdk-crypto/src/machine/mod.rs:654:47
   3: matrix_sdk::encryption::Encryption::bootstrap_cross_signing::{{closure}}
             at src/matrix-rust-sdk/crates/matrix-sdk/src/encryption/mod.rs:1185:48
   4: matrix_sdk::encryption::Encryption::bootstrap_cross_signing_if_needed::{{closure}}
             at src/matrix-rust-sdk/crates/matrix-sdk/src/encryption/mod.rs:1353:53
   5: matrix_sdk::encryption::Encryption::spawn_initialization_task::{{closure}}
             at src/matrix-rust-sdk/crates/matrix-sdk/src/encryption/mod.rs:1668:83
```

Technically, we need the other two stack trace acquiring `private_identity`, but we can just follow the stack down and track the execution. Indeed, in the second backtrace the `private_identity` is already locked (this is the lock this patch is removing). In the first, trace continuing the executing `private_identity` is locked in `keys_for_upload`.

As proposed, we could try to acquire the lock on account first. Before actually locking the private identity. I am not sure it is ok to hold two locks for such a big chunk of execution. It also creates a dependency between locks. WDYT?

poljar: > As proposed, we could try to acquire the lock on account first. Before actually locking the private identity. I am not sure it is ok to hold two locks for such a big chunk of execution. It also creates a dependency between locks. WDYT?

I think that this would be the way to go. Eventually we'll want to move the private identity behind the store cache lock, and then we would only have a single lock for this and no deadlock because of interdependent locks could happen.

I don't think that this is quite as a big chunk of execution, we're just generating some random bytes and producing a signature, so things should™ be fine.

boxdot: Locking account on top of `bootstrap_cross_signing` now deadlocks on it. But only for the first two initial logins. First time (no account yet) it just deadlocks, second time the bootstrap finishes with `Own identity is now verified, check all known identities for verification status changes` but now sync is deadlocked. Third time all starts working. I can't tell why it is like that without digging deeper. Maybe sync hold a lock to it for some reason.

boxdot: Maybe I could take a look at this as a follow up?

poljar (approved):

PR #4061, opened 2024-10-01
base: tidy up processing of global events in sync responses

bnjbvr:
This tidies up the code by making the data dependencies very explicit, instead of stashing data into `StateChanges` and then relying on it later on during computation. It reduces the amount of duplicated code, streamlines the processing of global account data events, and makes it easier to reason. Also would make it easier to test in isolation, would we need to. Overall, I think this is an improvement, and generalizing this would help commonize more code between sync-v2 and SSS processing.

### DIFF
--- a/crates/matrix-sdk-base/src/client.rs
+++ b/crates/matrix-sdk-base/src/client.rs
@@ -789,17 +789,8 @@ impl BaseClient {
     pub(crate) async fn preprocess_to_device_events(
         &self,
         encryption_sync_changes: EncryptionSyncChanges<'_>,
-        #[cfg(feature = "experimental-sliding-sync")] changes: &mut StateChanges,
-        #[cfg(not(feature = "experimental-sliding-sync"))] _changes: &mut StateChanges,
-        #[cfg(feature = "experimental-sliding-sync")] room_info_notable_updates: &mut BTreeMap<
-            OwnedRoomId,
-            RoomInfoNotableUpdateReasons,
-        >,
-        #[cfg(not(feature = "experimental-sliding-sync"))]
-        _room_info_notable_updates: &mut BTreeMap<
-            OwnedRoomId,
-            RoomInfoNotableUpdateReasons,
-        >,
+        changes: &mut StateChanges,
+        room_info_notable_updates: &mut BTreeMap<OwnedRoomId, RoomInfoNotableUpdateReasons>,
     ) -> Result<Vec<Raw<ruma::events::AnyToDeviceEvent>>> {
         if let Some(o) = self.olm_machine().await.as_ref() {
             // Let the crypto machine handle the sync response, this
@@ -815,8 +806,9 @@ impl BaseClient {
                     self.decrypt_latest_events(&room, changes, room_info_notable_updates).await;
                 }
             }
-            #[cfg(not(feature = "experimental-sliding-sync"))]
-            drop(room_key_updates); // Silence unused variable warning
+
+            #[cfg(not(feature = "experimental-sliding-sync"))] // Silence unused variable warnings.
+            let _ = (room_key_updates, changes, room_info_notable_updates);
 
             Ok(events)
         } else {
--- a/crates/matrix-sdk-base/src/client.rs
+++ b/crates/matrix-sdk-base/src/client.rs
@@ -16,7 +16,7 @@
 #[cfg(feature = "e2e-encryption")]
 use std::ops::Deref;
 use std::{
-    collections::{BTreeMap, BTreeSet, HashMap, HashSet},
+    collections::{BTreeMap, BTreeSet},
     fmt, iter,
     sync::Arc,
 };
@@ -49,10 +49,9 @@ use ruma::{
                 RoomPowerLevelsEvent, RoomPowerLevelsEventContent, StrippedRoomPowerLevelsEvent,
             },
         },
-        AnyGlobalAccountDataEvent, AnyRoomAccountDataEvent, AnyStrippedStateEvent,
-        AnySyncEphemeralRoomEvent, AnySyncMessageLikeEvent, AnySyncStateEvent,
-        AnySyncTimelineEvent, GlobalAccountDataEventType, StateEvent, StateEventType,
-        SyncStateEvent,
+        AnyRoomAccountDataEvent, AnyStrippedStateEvent, AnySyncEphemeralRoomEvent,
+        AnySyncMessageLikeEvent, AnySyncStateEvent, AnySyncTimelineEvent,
+        GlobalAccountDataEventType, StateEvent, StateEventType, SyncStateEvent,
     },
     push::{Action, PushConditionRoomCtx, Ruleset},
     serde::Raw,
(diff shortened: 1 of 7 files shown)


crates/matrix-sdk-base/src/client.rs
```diff
@@ -1503,14 +1377,15 @@ impl BaseClient {
...
-        if let Some(event) = changes
-            .account_data
-            .get(&GlobalAccountDataEventType::PushRules)
-            .and_then(|ev| ev.deserialize_as::<PushRulesEvent>().ok())
+    /// Gets the push rules previously processed, otherwise get them from the
+    /// store. As a fallback, uses [`Ruleset::server_default`] if the user
+    /// is logged in.
+    pub(crate) async fn get_push_rules(
+        &self,
+        account_data: &AccountDataProcessor,
```
stefanceriu, line 1385:
```suggestion
        account_data_processor: &AccountDataProcessor,
```

crates/matrix-sdk-base/src/client.rs
```diff
@@ -1270,20 +1147,17 @@ impl BaseClient {
...
-                    Ok(event) => {
-                        let user_ids: Vec<String> =
-                            event.content.ignored_users.keys().map(|id| id.to_string()).collect();
-
-                        self.ignore_user_list_changes.set(user_ids);
-                    }
-                    Err(error) => {
-                        warn!("Failed to deserialize ignored user list event: {error}")
-                    }
+        if let Some(event) = changes.account_data.get(&GlobalAccountDataEventType::IgnoredUserList)
```
stefanceriu, line 1150: Asking to use the full processor name everywhere because I got confused here. Started looking for a get method in the processor. It's just simpler to not have that mental overload to being with.

crates/matrix-sdk-base/src/client.rs
```diff
@@ -1270,20 +1147,17 @@ impl BaseClient {
...
+        {
+            match event.deserialize_as::<IgnoredUserListEvent>() {
+                Ok(event) => {
+                    let user_ids: Vec<String> =
+                        event.content.ignored_users.keys().map(|id| id.to_string()).collect();
+
+                    self.ignore_user_list_changes.set(user_ids);
+                }
+                Err(error) => {
+                    warn!("Failed to deserialize ignored user list event: {error}")
```
stefanceriu, line 1160: Should this be logged as an actual error?

crates/matrix-sdk-base/src/client.rs
```diff
@@ -1206,25 +1101,7 @@ impl BaseClient {
...
-            self.store.get_account_data_event(GlobalAccountDataEventType::Direct).await
-        {
-            debug!("Found direct room data in the Store, applying it");
-            if let Ok(direct_account_data) = direct_account_data.deserialize() {
-                self.process_direct_rooms(&[direct_account_data], &mut changes).await;
-            } else {
-                warn!("Failed to deserialize direct room account data");
-            }
-        }
+        account_data.apply(&mut changes, &self.store).await;
```
stefanceriu, line 1104: Interface wise it's a bit weird for a processor to be stateful. I would've imagined it a bit more functional with the `Processor` only having static methods on itself and producing various versions of `AccountData` on successive calls
e.g. `Processor::process(Response) -> AccountData`; `Processor::process(StateChanges, AccountData) -> AccountData`

crates/matrix-sdk-base/src/sliding_sync/mod.rs
```diff
@@ -163,16 +161,11 @@ impl BaseClient {
         let store = self.store.clone();
         let mut ambiguity_cache = AmbiguityCache::new(store.inner.clone());
 
-        let account_data = &extensions.account_data;
-        let global_account_data_events = if !account_data.is_empty() {
-            self.handle_account_data(&account_data.global, &mut changes).await
-        } else {
-            Vec::new()
-        };
+        let account_data = AccountDataProcessor::process(&extensions.account_data.global);
```
stefanceriu, line 164:
```suggestion
        let account_data_processor = AccountDataProcessor::process(&extensions.account_data.global);
```

crates/matrix-sdk-base/src/sliding_sync/mod.rs
```diff
@@ -163,16 +161,11 @@ impl BaseClient {
...
 
         let mut new_rooms = RoomUpdates::default();
         let mut notifications = Default::default();
-        let mut rooms_account_data = account_data.rooms.clone();
+        let mut rooms_account_data = extensions.account_data.rooms.clone();
```
stefanceriu, line 168: Presumably this will go away too eventually and only the Processor/AccountData(?) will be used?

stefanceriu (commented):
Good stuff, definitely a huge improvement! 👏

crates/matrix-sdk-base/src/client.rs
jmartinesp: I'm not sure I follow, isn't `account_data` just a `BTreeMap`?

crates/matrix-sdk-base/src/sliding_sync/mod.rs
jmartinesp: I guess so, either we add room account data to the processor too or we create a different processor for this data, right?

crates/matrix-sdk-base/src/client.rs
stefanceriu: Yeah, that's my point: we use the same `account_data` name for 2 different concepts in this file.

crates/matrix-sdk-base/src/client.rs
poljar: To expand on this. We introduced a `Processor` concept for sliding sync responses as well, but it also gets mutated and it needs to be mutated in the correct order, otherwise it panics:

https://github.com/matrix-org/matrix-rust-sdk/blob/59d3608c32c94b7bf210a9bee1ba2733d37b2cec/crates/matrix-sdk/src/sliding_sync/client.rs#L288-L290

A good way to handle this sort of requirements might be to introduce typestates, i.e. like @stefanceriu suggest, but each `process()` function accepts a certain part of the response and returns a different processor.

For the `SlidingSyncResponseProcessor` this would look somewhat like this:


```rust

struct SlidingSyncEncryptionProcessor {
...
}

impl SlidingSyncEncryptionProcessor {
    fn process(self, extensions: &Extensions) -> SlidingSyncRoomProcessor {
    }
}

struct SlidingSyncRoomProcessor {
    ...
}

...
```

This avoids the need for any sort of asserts to ensure things happen in the right order, because the right order is encoded in the types.

crates/matrix-sdk/src/sliding_sync/client.rs
```diff
@@ -289,11 +289,19 @@ impl<'a> SlidingSyncResponseProcessor<'a> {
...
+            .client
+            .base_client()
+            .process_sliding_sync_e2ee(extensions.to_device.as_ref(), &extensions.e2ee)
+            .await?
+        {
+            // Some new keys might have been received, so trigger a backup if needed.
+            self.client.encryption().backups().maybe_trigger_backup();
 
-        // Some new keys might have been received, so trigger a backup if needed.
-        self.client.encryption().backups().maybe_trigger_backup();
```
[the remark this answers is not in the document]
bnjbvr: @poljar ought to be able to confirm, but: if you haven't received any e2ee/to-device data, I suppose there's no new key coming from another way, right?

crates/matrix-sdk-base/src/client.rs
```diff
@@ -990,10 +886,9 @@ impl BaseClient {
 
         let mut ambiguity_cache = AmbiguityCache::new(self.store.inner.clone());
 
-        let global_account_data_events =
-            self.handle_account_data(&response.account_data.events, &mut changes).await;
+        let account_data = AccountDataProcessor::process(&response.account_data.events);
```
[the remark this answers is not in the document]
bnjbvr: Sure! I like short names, but this one is a bit weird indeed.

crates/matrix-sdk-base/src/client.rs
bnjbvr: Indeed, here it's not the processor per se, it's the `account_data` from `StateChanges`.

crates/matrix-sdk-base/src/client.rs
bnjbvr: sure, why not

crates/matrix-sdk-base/src/sliding_sync/mod.rs
bnjbvr: Yep, it would be another processor indeed.

crates/matrix-sdk/src/sliding_sync/client.rs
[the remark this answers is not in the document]
poljar: There are a bunch of places where room keys can come from, but in a `/sync` response, only the to-device part will carry any room keys.

This would be another place and we call the `maybe_trigger_backup()` method as well:

https://github.com/matrix-org/matrix-rust-sdk/blob/59d3608c32c94b7bf210a9bee1ba2733d37b2cec/crates/matrix-sdk/src/encryption/mod.rs#L1490-L1492

crates/matrix-sdk-base/src/client.rs
stefanceriu: Yeah, I can definitely live with that, it's already a great step forward. 
Another quick win might be to assert that it's consumed once and only once internally after moving the data, or perhaps even renaming `apply` to `transfer_into` or something.

stefanceriu (approved):

crates/matrix-sdk-base/src/client.rs
bnjbvr: For the sake of completeness, forgot to respond here, so responded in private after a ping from stefan:

> this doesn't buy us anything because the type is must_use (so must be consumed at least once) and since apply(self) consumes self, it's consumed at most once
> static typing ftw!

Issue #4048, opened 2024-09-30
Add support in timeline for the new MegolmError `SenderIdentityNotTrusted` when cross-signed `TrustRequirement` is enabled


BillCarsonFr:
Part of https://github.com/element-hq/crypto-internal/issues/308.

Modify TimelineEvent to expose New Megolm errors when TrustRequirement is CrossSigning*, add new cases to UtdCause.

See https://github.com/element-hq/crypto-internal/issues/353 for the new error to show in timeline

richvdh:
Having discussed this a bit with the rust team, the concern was raised that the types proposed allow for a conflict in state between `encryption_state` and `event`.  (For example: what would it mean if `encryption_state` indicated a UTD, but `event` was a successful decryption?)

Instead: it was proposed to replace `event` and `encryption_info` in `TimelineEvent` and `SyncTimelineEvent` with a new enum:

```rust
enum TimelineFoo {
    Utd {
        event: Raw<EncryptedEvent>,
        cause: UtdCause,
    },
    PlainText {
        event: Raw<T>,
    }
    Decrypted {
        event: Raw<T>,
        info: EncryptionInfo,
    }
}
```

(better name needed, obvs)

We should also update `OlmMachine::decrypt_room_event` to return a `TimelineEvent` with `TimelineFoo::Utd` instead of an `Err`, on UTD errors.

richvdh:
@BillCarsonFr the rust bits of this are now done - can you take this over to add the UI layers?

BillCarsonFr:
> [@BillCarsonFr](https://github.com/BillCarsonFr) the rust bits of this are now done - can you take this over to add the UI layers?

Working on it, I think this issue can be closed. The remaining work is tracked in other issues

richvdh:
Fixed by https://github.com/matrix-org/matrix-rust-sdk/pull/4070

PR #4064, opened 2024-10-03
crypto: New return type for `decrypt_room_event`

richvdh:
I need to do a bit of a refactoring on `TimelineEvent`, so let's start by giving `decrypt_room_event` its own return type.

(some groundwork for https://github.com/matrix-org/matrix-rust-sdk/issues/4048)

### DIFF
--- a/crates/matrix-sdk-common/src/deserialized_responses.rs
+++ b/crates/matrix-sdk-common/src/deserialized_responses.rs
@@ -15,7 +15,7 @@
 use std::{collections::BTreeMap, fmt};
 
 use ruma::{
-    events::{AnySyncTimelineEvent, AnyTimelineEvent},
+    events::{AnyMessageLikeEvent, AnySyncTimelineEvent, AnyTimelineEvent},
     push::Action,
     serde::{JsonObject, Raw},
     DeviceKeyAlgorithm, OwnedDeviceId, OwnedEventId, OwnedUserId,
@@ -380,6 +380,13 @@ impl From<TimelineEvent> for SyncTimelineEvent {
     }
 }
 
+impl From<DecryptedRoomEvent> for SyncTimelineEvent {
+    fn from(decrypted: DecryptedRoomEvent) -> Self {
+        let timeline_event: TimelineEvent = decrypted.into();
+        timeline_event.into()
+    }
+}
+
 #[derive(Clone)]
 pub struct TimelineEvent {
     /// The actual event.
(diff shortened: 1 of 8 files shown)


crates/matrix-sdk-common/src/deserialized_responses.rs
```diff
@@ -423,6 +444,35 @@ impl fmt::Debug for TimelineEvent {
     }
 }
 
+#[derive(Clone, Serialize, Deserialize)]
+/// A successfully-decrypted encrypted event.
```
richvdh, line 448: The casual reader might wonder why this is here, rather than in the crypto crate. The answer is that, in the fullness of time, I want to make `TimelineEvent` and `SyncTimelineEvent` wrap a `DecryptedRoomEvent` rather than duplicating the fields.

crates/matrix-sdk-crypto/src/machine/mod.rs
```diff
@@ -1818,14 +1818,9 @@ impl OlmMachine {
                 .await;
         }
 
-        let event = serde_json::from_value::<Raw<AnyTimelineEvent>>(decrypted_event.into())?;
+        let event = serde_json::from_value::<Raw<AnyMessageLikeEvent>>(decrypted_event.into())?;
```
andybalaam, line 1821: I know state events aren't currently encrypted, but does the change from `AnyTimelineEvent` to `AnyMessageLikeEvent` here mean that we are assuming this is not a state event? Are we OK with that?

andybalaam (approved):
One question - looks good.

crates/matrix-sdk-crypto/src/machine/mod.rs
richvdh: Good question.

Yes I am assuming this is not a state event, and that our lack of encryption for state events is not going to change any time soon. It felt like using the more specific type might be more useful for callers. It's also worth noting that situation can't change without changes elsewhere in the crypto crate (specifically we construct the JSON for the decrypted event in `InboundGroupSession::decrypt`).

On the other hand: maybe it's better we encourage callers to be ready for state events by continuing to return `Raw<AnyTimelineEvent>`.

I'd be interested in @poljar's thoughts here.

poljar: I really don't expect state events to be able to use the same mechanisms for encryption as we use for non-state events, perhaps I'm mistaken here but I would expect that they would be different enough to warrant a separate method.

richvdh: >  I would expect that they would be different enough to warrant a separate method.

I guess the point is that it's not just about a separate method: the intention is to embed `DecryptedRoomEvent` inside `TimelineEvent` and `SyncTimelineEvent` -- so we'll be hardcoding this knowledge there too.

That said: on the grand scale of changes that will be needed to support encrypted state events, I don't think changing this later is going to be that significant.

PR #4065, opened 2024-10-03
Do not log full keys query response which can be very large

erikjohnston:
Noticed this while looking at the logs, where I saw a log line of ~1.26MB

### DIFF
--- a/crates/matrix-sdk/src/encryption/recovery/mod.rs
+++ b/crates/matrix-sdk/src/encryption/recovery/mod.rs
@@ -629,7 +629,7 @@ impl Recovery {
         }
     }
 
-    #[instrument(skip(self, response))]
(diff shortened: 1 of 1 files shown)


crates/matrix-sdk/src/encryption/recovery/mod.rs
```diff
@@ -629,7 +629,7 @@ impl Recovery {
         }
     }
 
-    #[instrument]
+    #[instrument(skip(self, response))]
```
poljar, line 632:
```suggestion
    #[instrument(skip_all)]
```

https://docs.rs/tracing/latest/tracing/attr.instrument.html#skipping-fields

poljar (approved):

Issue #3905, opened 2024-08-28
timeline: improve handling of edits


bnjbvr:
- If an edit fails, find Some Way™ to reflect that in the timeline
  - revert to the unedited version of the item?
  - add a way to re-send the edit?
- If an edit succeeds, mark the item as such, and don't replace it with the previous unedited version (if it comes late from sync), see also https://github.com/element-hq/element-x-ios-rageshakes/issues/2126#issuecomment-2315708428
  - but handle that another edit can come from another client through sync :eyes:

PR #4066, opened 2024-10-03
timeline: prefer a bundled edit over a pending one

### LINKED ISSUE #2848: Sometimes EventTimelineItem::latest_edit_json() returns None for encrypted room
For unencrypted room, `EventTimelineItem::latest_edit_json()` works well.
For encrypted room, `EventTimelineItem::latest_edit_json()` works well about msg that was modified during app's current launch.
But for encrypted room, `EventTimelineItem::latest_edit_json()` returns `None` about msg that was modified during app's previous launch (when app is restarted).
`EventTimelineItem::is_edited()` works well for both unencrypted room and encrypted room.

bnjbvr:
When adding a new event, we might also receive a bundled edit as part of `m.relations`; this is guaranteed to be the most up-to-date (if received from sync), so we should prefer this bundled edit over any pending edit we might have found.

Consider the following scenario:

- received pending edit $2 for event $1
- another edit $3 happens to $1
- when receiving $1, we may receive $3 as a bundled edit, but might have seen $2 before. In that case, it's preferable to use $3 rather than $2.

It's not clear it's the absolute fix, since we don't know the relative ordering of $2 and $3: if getting the event from cache, it might have bundled relations that are now outdated. I wonder if we should get rid of the bundled relations when storing them in the event cache :eyes: A question for another day.

This also fixes an issue where we'd not include the `latest_edit_json` for bundled edits of a room message. This is now included and properly tested. As such, this fixes #2848.

This contains a few extra refactorings, tests, etc. In particular, it also includes bundled relations when considering a latest event preview, and as such it might help displaying the edit rather than the original in more cases.

Part of #3905.

### DIFF
--- a/crates/matrix-sdk-ui/src/timeline/event_handler.rs
+++ b/crates/matrix-sdk-ui/src/timeline/event_handler.rs
@@ -51,7 +51,7 @@ use super::{
     controller::{TimelineMetadata, TimelineStateTransaction},
     day_dividers::DayDividerAdjuster,
     event_item::{
-        AnyOtherFullStateEventContent, EventSendState, EventTimelineItemKind,
+        extract_edit_content, AnyOtherFullStateEventContent, EventSendState, EventTimelineItemKind,
         LocalEventTimelineItem, Profile, ReactionsByKeyBySender, RemoteEventOrigin,
         RemoteEventTimelineItem, TimelineEventItemId,
     },
@@ -328,7 +328,33 @@ impl<'a, 'o> TimelineEventHandler<'a, 'o> {
 
                 AnyMessageLikeEventContent::RoomMessage(c) => {
                     if should_add {
-                        self.add_item(TimelineItemContent::message(c, relations, self.items));
+                        // Always remove the pending edit, if there's any. The reason is that if
+                        // there's an edit in the relations mapping, we want to prefer it over any
+                        // other pending edit, since it's more likely to be up to date, and we
+                        // don't want to apply another pending edit on top of it.
+                        let pending_edit = if let Flow::Remote { event_id, .. } = &self.ctx.flow {
+                            let edits = &mut self.meta.pending_edits;
+                            edits
+                                .iter()
+                                .position(|(prev_event_id, _)| prev_event_id == event_id)
+                                .into_iter()
+                                .filter_map(|pos| {
+                                    Some(
+                                        as_variant!(
+                                            edits.remove(pos).unwrap().1,
+                                            PendingEdit::RoomMessage
+                                        )?
+                                        .new_content,
+                                    )
+                                })
+                                .next()
+                        } else {
+                            None
+                        };
+
+                        let edit = extract_edit_content(relations).or(pending_edit);
+
+                        self.add_item(TimelineItemContent::message(c, edit, self.items));
                     }
                 }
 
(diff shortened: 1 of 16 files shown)


crates/matrix-sdk-ui/src/timeline/event_handler.rs
```diff
@@ -339,20 +339,40 @@ impl<'a, 'o> TimelineEventHandler<'a, 'o> {
...
+                                    }
+                                    _ => None,
                                 });
 
-                        let edit = extract_edit_content(relations).or(pending_edit);
+                        let (edit_json, edit_content) = extract_edit_content(relations)
+                            .map(|content| {
+                                let edit_json = as_variant!(&self.ctx.flow, Flow::Remote { raw_event, ..} => raw_event).and_then(|raw| {
+                                    // Kids, don't do this at home. We're extracting the edit event
+                                    // from the `unsigned`.m.relations`.`m.replace` path.
```
andybalaam: Can you explain why we can't do this a nicer way? Thanks!

crates/matrix-sdk-ui/src/timeline/event_handler.rs
```diff
@@ -553,14 +577,16 @@ impl<'a, 'o> TimelineEventHandler<'a, 'o> {
             TimelineItemContent::Message(..) => {
                 let pending =
                     Self::find_and_remove_pending(&mut self.meta.pending_edits, event_id)?;
-                let edit = as_variant!(pending, PendingEdit::RoomMessage)?;
-                self.apply_msg_edit(item, edit)
+                let edit = as_variant!(pending.kind, PendingEditKind::RoomMessage)?;
+                //let json = as_variant!(&self.ctx.flow, Flow::Remote { raw_event, .. } =>
+                // raw_event.clone()); self.apply_msg_edit(item, edit, json)
```
andybalaam: Commented code?

crates/matrix-sdk-ui/src/timeline/event_handler.rs
```diff
@@ -553,15 +553,15 @@ impl<'a, 'o> TimelineEventHandler<'a, 'o> {
...
+                let _ = Self::maybe_unstash_pending_edit(edits, &replaced_event_id);
                 edits.push(replacement);
                 debug!("Timeline item not found, stashing edit");
             }
         }
     }
 
-    /// TODO rename to maybe_unstash_pending_edit?
-    fn find_and_remove_pending(
+    /// Look for a pending edit for the given event, and returns it, if found.
```
andybalaam, line 575: And removes it from the `edits` list, right?

crates/matrix-sdk-ui/src/timeline/event_handler.rs
```diff
@@ -99,6 +99,18 @@ pub(super) enum Flow {
     },
 }
 
+impl Flow {
+    /// If the flow is remote, returns the associated event id.
+    pub(crate) fn event_id(&self) -> Option<&EventId> {
```
andybalaam, line 104: I'm a big fan of these methods!

crates/matrix-sdk/src/test_utils/events.rs
```diff
@@ -122,9 +138,23 @@ where
         if let Some(redacts) = self.redacts {
             map.insert("redacts".to_owned(), json!(redacts));
         }
+
         if let Some(unsigned) = self.unsigned {
-            map.insert("unsigned".to_owned(), json!(unsigned));
+            let mut unsigned_json = json!({});
+
+            // We can't plain serialize `Unsigned`, otherwise this would result in some
+            // `null` fields when options are set to none, which Ruma rejects.
```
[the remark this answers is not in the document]
bnjbvr: Ah sweet, I always forget about this. Thanks!

PR #4067, opened 2024-10-03
crypto: Rename PreviouslyVerified to VerificationViolation

andybalaam:
Part of https://github.com/element-hq/element-meta/issues/2526

For consistency with other places, we have now settled on `VerificationViolation` as the best way to express this situation.

### DIFF
--- a/crates/matrix-sdk-crypto/CHANGELOG.md
+++ b/crates/matrix-sdk-crypto/CHANGELOG.md
@@ -57,6 +57,9 @@ Breaking changes:
   the CryptoStore, meaning that, once upgraded, it will not be possible to roll
   back applications to earlier versions without breaking user sessions.
 
+- Renamed `VerificationLevel::PreviouslyVerified` to
+  `VerificationLevel::VerificationViolation`.
+
 - `OlmMachine::decrypt_room_event` now takes a `DecryptionSettings` argument,
   which includes a `TrustRequirement` indicating the required trust level for
   the sending device.  When it is called with `TrustRequirement` other than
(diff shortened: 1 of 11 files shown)


crates/matrix-sdk-crypto/src/olm/group_sessions/sender_data.rs
```diff
@@ -273,8 +269,9 @@ pub enum SenderDataType {
     UnknownDevice = 1,
     /// The [`SenderData`] is of type `DeviceInfo`.
     DeviceInfo = 2,
-    /// The [`SenderData`] is of type `SenderUnverifiedButPreviouslyVerified`.
-    SenderUnverifiedButPreviouslyVerified = 3,
+    /// The [`SenderData`] is of type
+    /// `VerificationViolation`.
```
bnjbvr, line 273: nit: doesn't this fit on a single line?

bnjbvr (approved):
LGTM. Note that `previously_verified` returns quite a few matches with a plain grep, so consider whether you want to rename those also or not (and no need to ask for another review from me IMO, unless you're unsure/unclear about one thing in particular).

crates/matrix-sdk-crypto/CHANGELOG.md
```diff
@@ -69,7 +69,7 @@ Breaking changes:
   previously-verified, unverified and verified.
   ([#3877](https://github.com/matrix-org/matrix-rust-sdk/pull/3877))
 
-- Where `EncryptionInfo` is returned it may include the new `PreviouslyVerified`
+- Where `EncryptionInfo` is returned it may include the new `VerificationViolation`
```
richvdh, line 72: driveby comment: given we've already released this stuff, can you put a separate changelog entry higher up, please?

crates/matrix-sdk-crypto/src/olm/group_sessions/sender_data.rs
andybalaam: Fixed in b1e946c0fc

crates/matrix-sdk-crypto/CHANGELOG.md
andybalaam: Thanks, I dithered about this and chose the wrong choice!

andybalaam: Fixed in a3da650b06

PR #4068, opened 2024-10-03
crypto: Expose a method to pin a user's identity

andybalaam:
Part of https://github.com/element-hq/element-meta/issues/2526

### DIFF
--- a/bindings/matrix-sdk-ffi/CHANGELOG.md
+++ b/bindings/matrix-sdk-ffi/CHANGELOG.md
@@ -31,4 +31,5 @@ Breaking changes:
 
 Additions:
 
+- Add `Encryption::get_user_identity`
 - Add `ClientBuilder::room_key_recipient_strategy`
--- a/bindings/matrix-sdk-ffi/CHANGELOG.md
+++ b/bindings/matrix-sdk-ffi/CHANGELOG.md
@@ -31,5 +31,5 @@ Breaking changes:
 
 Additions:
 
-- Add `Encryption::get_user_identity`
+- Add `Encryption::get_user_identity` which returns `UserIdentity`
 - Add `ClientBuilder::room_key_recipient_strategy`
--- a/bindings/matrix-sdk-ffi/src/encryption.rs
+++ b/bindings/matrix-sdk-ffi/src/encryption.rs
@@ -413,19 +413,16 @@ impl Encryption {
 
     /// Get the E2EE identity of a user.
     ///
-    /// Returns an error if this user does not exist, if there is an error
-    /// contacting the crypto store, or if our client is not logged in.
+    /// Returns Ok(None) if this user does not exist.
+    ///
+    /// Returns an error if there was a problem contacting the crypto store, or
+    /// if our client is not logged in.
     pub async fn get_user_identity(
         &self,
         user_id: String,
-    ) -> Result<Arc<UserIdentity>, ClientError> {
-        Ok(Arc::new(UserIdentity {
-            inner: self
-                .inner
-                .get_user_identity(user_id.as_str().try_into()?)
-                .await?
-                .ok_or(ClientError::new("User not found"))?,
-        }))
+    ) -> Result<Option<Arc<UserIdentity>>, ClientError> {
+        let identity = self.inner.get_user_identity(user_id.as_str().try_into()?).await?;
+        Ok(identity.map(|i| Arc::new(UserIdentity { inner: i })))
     }
 }
 
@@ -453,6 +450,16 @@ impl UserIdentity {
     /// which is not verified and is in pin violation.
     pub(crate) async fn pin(&self) -> Result<(), ClientError> {
         Ok(self.inner.pin().await?)
+    }
+
+    /// Get the public part of the Master key of this user identity.
+    ///
+    /// The public part of the Master key is usually used to uniquely identify
+    /// the identity.
+    ///
+    /// Returns None if the master key does not actually contain any keys.
+    pub(crate) fn master_key(&self) -> Option<String> {
+        self.inner.master_key().get_first_key().map(|k| k.to_base64())
     }
 }
 
--- a/bindings/matrix-sdk-ffi/src/encryption.rs
+++ b/bindings/matrix-sdk-ffi/src/encryption.rs
@@ -410,6 +410,50 @@ impl Encryption {
     pub async fn wait_for_e2ee_initialization_tasks(&self) {
         self.inner.wait_for_e2ee_initialization_tasks().await;
     }
+
+    /// Get the E2EE identity of a user.
+    ///
+    /// Returns an error if this user does not exist, if there is an error
+    /// contacting the crypto store, or if our client is not logged in.
+    pub async fn get_user_identity(
+        &self,
+        user_id: String,
+    ) -> Result<Arc<UserIdentity>, ClientError> {
+        Ok(Arc::new(UserIdentity {
+            inner: self
+                .inner
+                .get_user_identity(user_id.as_str().try_into()?)
+                .await?
+                .ok_or(ClientError::new("User not found"))?,
+        }))
+    }
+}
+
+/// The E2EE identity of a user.
+#[derive(uniffi::Object)]
+pub struct UserIdentity {
+    inner: matrix_sdk::encryption::identities::UserIdentity,
+}
+
+#[uniffi::export]
+impl UserIdentity {
+    /// Remember this identity, ensuring it does not result in a pin violation.
+    ///
+    /// When we first see a user, we assume their cryptographic identity has not
+    /// been tampered with by the homeserver or another entity with
+    /// man-in-the-middle capabilities. We remember this identity and call this
+    /// action "pinning".
+    ///
+    /// If the identity presented for the user changes later on, the newly
+    /// presented identity is considered to be in "pin violation". This
+    /// method explicitly accepts the new identity, allowing it to replace
+    /// the previously pinned one and bringing it out of pin violation.
+    ///
+    /// UIs should display a warning to the user when encountering an identity
+    /// which is not verified and is in pin violation.
+    pub(crate) async fn pin(&self) -> Result<(), ClientError> {
+        Ok(self.inner.pin().await?)
+    }
 }
 
 #[derive(uniffi::Object)]
--- a/crates/matrix-sdk-crypto/src/identities/user.rs
+++ b/crates/matrix-sdk-crypto/src/identities/user.rs
@@ -125,6 +125,32 @@ impl UserIdentity {
         }
     }
 
+    /// Remember this identity, ensuring it does not result in a pin violation.
+    ///
+    /// When we first see a user, we assume their cryptographic identity has not
+    /// been tampered with by the homeserver or another entity with
+    /// man-in-the-middle capabilities. We remember this identity and call this
+    /// action "pinning".
+    ///
+    /// If the identity presented for the user changes later on, the newly
+    /// presented identity is considered to be in "pin violation". This
+    /// method explicitly accepts the new identity, allowing it to replace
+    /// the previously pinned one and bringing it out of pin violation.
+    ///
+    /// UIs should display a warning to the user when encountering an identity
+    /// which is not verified and is in pin violation. See
+    /// [`OtherUserIdentity::identity_needs_user_approval`].
+    pub async fn pin(&self) -> Result<(), CryptoStoreError> {
+        match self {
+            UserIdentity::Own(_) => {
+                // Nothing to be done for our own identity: we already
+                // consider it trusted in this sense.
+                Ok(())
+            }
+            UserIdentity::Other(u) => u.pin_current_master_key().await,
+        }
+    }
+
     /// Was this identity previously verified, and is no longer?
     pub fn has_verification_violation(&self) -> bool {
         match self {
@@ -737,7 +763,21 @@ impl OtherUserIdentityData {
         &self.self_signing_key
     }
 
-    /// Pin the current identity
+    /// Remember this identity, ensuring it does not result in a pin violation.
+    ///
+    /// When we first see a user, we assume their cryptographic identity has not
+    /// been tampered with by the homeserver or another entity with
+    /// man-in-the-middle capabilities. We remember this identity and call this
+    /// action "pinning".
+    ///
+    /// If the identity presented for the user changes later on, the newly
+    /// presented identity is considered to be in "pin violation". This
+    /// method explicitly accepts the new identity, allowing it to replace
+    /// the previously pinned one and bringing it out of pin violation.
+    ///
+    /// UIs should display a warning to the user when encountering an identity
+    /// which is not verified and is in pin violation. See
+    /// [`OtherUserIdentity::identity_needs_user_approval`].
     pub(crate) fn pin(&self) {
         let mut m = self.pinned_master_key.write().unwrap();
         *m = self.master_key.as_ref().clone()
--- a/crates/matrix-sdk/CHANGELOG.md
+++ b/crates/matrix-sdk/CHANGELOG.md
@@ -28,6 +28,7 @@ Breaking changes:
 
 Additions:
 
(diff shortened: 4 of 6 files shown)


crates/matrix-sdk/src/room/mod.rs
```diff
@@ -403,6 +403,18 @@ impl Room {
...
+    /// identity before.
+    #[cfg(feature = "e2e-encryption")]
+    pub async fn pin_user_identity(&self, user_id: &UserId) -> Result<()> {
+        let x = self.client.encryption().get_user_identity(user_id).await?;
+
+        match x {
+            Some(x) => Ok(x.pin().await?),
+            None => Err(Error::UserNotFound),
+        }
+    }
```
richvdh, line 416: this doesn't seem specific to rooms... it seems a bit odd to expose it as part of `Room`?

crates/matrix-sdk/src/room/mod.rs
```diff
@@ -403,6 +403,18 @@ impl Room {
         IdentityStatusChanges::create_stream(self.clone()).await
     }
 
+    /// Mark this user as "pinned" meaning that we have seen their cryptographic
+    /// identity before.
+    #[cfg(feature = "e2e-encryption")]
+    pub async fn pin_user_identity(&self, user_id: &UserId) -> Result<()> {
+        let x = self.client.encryption().get_user_identity(user_id).await?;
+
+        match x {
```
richvdh, line 412: nit: either inline `x`, or give it a better name?

bindings/matrix-sdk-ffi/src/room.rs
```diff
@@ -609,6 +609,10 @@ impl Room {
         })))
     }
 
+    pub fn pin_user_identity(&self, user_id: &str) -> Result<(), ClientError> {
```
richvdh, line 612: could do with some documentation?

crates/matrix-sdk/src/room/mod.rs
andybalaam: Thanks, sorry I realised this the same time as you did and force-pushed an update.

bindings/matrix-sdk-ffi/src/room.rs
andybalaam: We don't document in the FFI layer normally - is there some reason why this one requires it?

crates/matrix-sdk/src/room/mod.rs
andybalaam: This was particularly requested by @stefanceriu - I guess it's for convenience because they have a room already available. I guess we could put it on `Client`. Any comment @stefanceriu ?

dkasak: Agreed this doesn't make sense to be on `Room`. Please let's not make the API mess solely for the purpose of convenience. `Client` does make sense.

stefanceriu: Yes, that was the reason. Also it makes it consistent with the other verification withdrawal methods. Otherwise you could even put it in the Client or even Encryption. Up to you really ¯\_(ツ)_/¯

crates/matrix-sdk/src/room/mod.rs
```diff
@@ -403,6 +403,20 @@ impl Room {
         IdentityStatusChanges::create_stream(self.clone()).await
     }
 
+    /// Mark this user as "pinned" meaning that we have seen their cryptographic
+    /// identity before.
+    #[cfg(feature = "e2e-encryption")]
+    pub async fn pin_user_identity(&self, user_id: &UserId) -> Result<()> {
```
dkasak, line 409: We're not saying *which* identity to pin, only which user's identity should be pinned, so this feels a bit smelly. I'm not sure how we will be using this API, but is there potential for a switcheroo/race here?

We observe one identity for some user U, which we want to pin. But then it switches under us, which we don't notice because the API is simply "pin *some* identity for user U, whichever one you find".

bindings/matrix-sdk-ffi/src/room.rs
stefanceriu: Actually we should (and have) because they get properly forwarded to the bindings.

crates/matrix-sdk-crypto/src/identities/user.rs
```diff
@@ -125,6 +125,22 @@ impl UserIdentity {
...
+    ///
+    /// Resets this identity to the same "pinned" state assigned the first time
+    /// a user is encountered, where we follow "trust on first use" and
+    /// consider it trusted.
+    pub async fn pin(&self) -> Result<(), CryptoStoreError> {
+        match self {
+            UserIdentity::Own(_) => {
+                // Nothing to be done for our own identity: we already
+                // consider it trusted in this sense.
+                Ok(())
```
BillCarsonFr, line 138: Right, it's more that we consider that you should always verifiy your own identity. We need to clarify this issue with own identity as it is surfacing in several places of the API
We could also error? or maybe one should do `identity.other().map(.. pin..)`?

BillCarsonFr (approved):
A bit original to have the API on room, but looks like it's convenient for apps

bindings/matrix-sdk-ffi/CHANGELOG.md
```diff
@@ -31,4 +31,5 @@ Breaking changes:
 
 Additions:
 
+- Add `Room::pin_user_identity`
```
poljar, line 34: I'm sorry, but this can't be exposed in this manner. Please expose the `UserIdentity` and the `pin()` method the same way we have it in the main crate.

crates/matrix-sdk/src/room/mod.rs
poljar: Even if we are sure that a switcheroo can't happen in a dangerous manner at this point in time, and in the way EX wants to use it right now, things might change in the future and we have other users of the bindings. We can't guarantee that you can't shoot yourself in the foot this way.

There is a reason why this isn't exposed on the room in the main crate and the API for this should be kept 1:1.

poljar (changes requested):

crates/matrix-sdk-crypto/src/identities/user.rs
```diff
@@ -125,6 +125,22 @@ impl UserIdentity {
         }
     }
 
+    /// Consider this identity to be trusted.
+    ///
+    /// Resets this identity to the same "pinned" state assigned the first time
+    /// a user is encountered, where we follow "trust on first use" and
+    /// consider it trusted.
```
dkasak, line 132: I must admit I don't understand this explanation at all.

The first thing that throws me is the word "trusted". Do we still retain the ability to differentiate between identities which are simply pinned versus properly verified?

Furthermore, the word "resets" implies to me that we will force the "pinned" state, replacing any previous state, even if it was "verified". Is this the case?

crates/matrix-sdk/src/room/mod.rs
dkasak: Sorry, unresolving, since I didn't see poljar's comment before hitting resolve (the UI didn't update).

bindings/matrix-sdk-ffi/src/room.rs
andybalaam: OK, update force-pushed with docs included

crates/matrix-sdk/src/room/mod.rs
andybalaam: Hmm, good point. I've re-structured the interface and force-pushed. Now we can get the identity first, and pin it later.

We don't currently have a good way of actually identifying which identity we have, but I think that is outside the scope of this change. (I suspect we should expose the master key?)

bindings/matrix-sdk-ffi/CHANGELOG.md
andybalaam: Done in the force-pushed version I just did

crates/matrix-sdk/src/room/mod.rs
andybalaam: Changed to `Encryption::get_user_identity` then `UserIdentity::pin` in the force-pushed update I just did.

crates/matrix-sdk-crypto/src/identities/user.rs
andybalaam: Do you think we should error? I think probably not but am persuadable.

crates/matrix-sdk-crypto/src/identities/user.rs
andybalaam: Yes, this is very confusing at the moment! How about this wording:

```
/// "Pin" this identity, meaning it is not in "pin violation".
///
/// When we first see a user, we assume their cryptographic identity is correct.
/// We call this "pinning" the identity. If they later change their identity, then they are
/// in "pin violation": we require user interaction before they are pinned again.
///
/// Note: this is not related to verification of users, which is a more secure way of
/// confirming that the identity you are communicating with is really the person you
/// think it is.
```

andybalaam: (Applied in 2d89f6cab4 but definitely up for further discussion)

crates/matrix-sdk-crypto/src/identities/user.rs
BillCarsonFr: I don't know it makes me wonder if we should pin our own identity also :/ for the transition mode. Anyhow not blocking.

crates/matrix-sdk/src/room/mod.rs
BillCarsonFr: Sry @poljar I am a bit lost. Why would keeping the API 1:1 will help for the "switcheroo" thing?
If I understand well, what would fix it is that the `pin` method takes a "master_key" as parameter and that the room identity state changes stream gives you the `master_key` to confirm that you can then later pass back as a parameter.

But even that doesn't really help the race scenario in case of 2 consecutive changes? Not sure the current pinning mecanism works like that.

poljar: > If I understand well, what would fix it is that the pin method takes a "master_key" as parameter 

The `pin()` method takes the master key as an argument, it's just that the master key is inside of `self.inner.master_key`.

So yes, you are correct, the master key needs to be part of the function call.

> But even that doesn't really help the race scenario in case of 2 consecutive changes? Not sure the current pinning mecanism works like that.

The changes can happen at any point in time, but you have a copy of the identity which will not change under your feet while you decide if you want to pin it or not.

crates/matrix-sdk-crypto/src/identities/user.rs
dkasak:
```rust
/// Remember this identity, ensuring it does not result in a pin violation.
///
/// When we first see a user, we assume their cryptographic identity has not
/// been tampered with by the homeserver or another entity with
/// man-in-the-middle capabilities. We remember this identity and call this
/// action "pinning".
///
/// If the identity presented for the user changes later on, the newly presented
/// identity is considered to be in "pin violation". This method explicitly
/// accepts the new identity, allowing it to replace the previously pinned one
/// and bringing it out of pin violation.
///
/// An identity which is not verified and is in pin violation will require user
/// approval before it can be used. See
/// [`OtherUserIdentity::identity_needs_user_approval`].
```

How about this instead?

I aimed to avoid glossing over verification as an entirely separate matter, instead linking pinning and verification to show how they jointly impact the system. I also spell out what "correct" means when talking about an identity.

crates/matrix-sdk/src/room/mod.rs
andybalaam: @poljar does the new version match what you expected?

bindings/matrix-sdk-ffi/src/encryption.rs
```diff
@@ -410,6 +410,50 @@ impl Encryption {
...
+    /// If the identity presented for the user changes later on, the newly
+    /// presented identity is considered to be in "pin violation". This
+    /// method explicitly accepts the new identity, allowing it to replace
+    /// the previously pinned one and bringing it out of pin violation.
+    ///
+    /// An identity which is not verified and is in pin violation will require
+    /// user approval before it can be used.
+    pub(crate) async fn pin(&self) -> Result<(), ClientError> {
+        Ok(self.inner.pin().await?)
+    }
```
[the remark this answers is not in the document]
andybalaam: Done in 7355cb46da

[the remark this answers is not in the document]
andybalaam: @poljar sorry, but I'd like you to take one last look at that please -  I called `get_first_key` and then `to_base64` and am not sure either of those is correct.

crates/matrix-sdk-crypto/src/identities/user.rs
```diff
@@ -737,7 +763,21 @@ impl OtherUserIdentityData {
...
+    /// man-in-the-middle capabilities. We remember this identity and call this
+    /// action "pinning".
+    ///
+    /// If the identity presented for the user changes later on, the newly
+    /// presented identity is considered to be in "pin violation". This
+    /// method explicitly accepts the new identity, allowing it to replace
+    /// the previously pinned one and bringing it out of pin violation.
+    ///
+    /// An identity which is not verified and is in pin violation will require
+    /// user approval before it can be used. See
```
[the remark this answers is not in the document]
dkasak: Yep, we should, though I'm out of my depth here since I've not been keeping this close of an eye on the new trust requirement code.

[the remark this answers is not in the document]
BillCarsonFr: We don't require the violation to be resolved before using the identity anymore (using here is for sharing room key or trust requirement for decryption). It was considered but it is per design a non blocking warning, so the sdk remembers if the warning was displayed to the user, until resolved the warning should be shown by UI.
So just outdated doc?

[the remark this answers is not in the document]
dkasak: Hmm, I have a number of questions now.

- So what exactly is the effect of `identity_needs_user_approval() == true`? It simply displays a warning with no other effect?
- Do we differentiate between a `identity_needs_user_approval() == true` state created due to a verified identity mismatch vs a pin violation? Or is the effect exactly the same in both cases?
- Is any of this going to change after the "transition period"?

Also, can you suggest wording that would make it accurate?

bindings/matrix-sdk-ffi/src/encryption.rs
```diff
@@ -410,6 +410,63 @@ impl Encryption {
...
+
+    /// Get the public part of the Master key of this user identity.
+    ///
+    /// The public part of the Master key is usually used to uniquely identify
+    /// the identity.
+    pub(crate) fn master_key(&self) -> Result<String, ClientError> {
```
poljar, line 462: This is fine from my point of view, though usually we return an `Result<Option<T>>` for these cases since it's easier to distinguish the "Missing master key" case from other error types. This is especially true for `ClientError` which has only a single variant and people would need to match on the string.

The same applies to the `get_user_identity()` method.

crates/matrix-sdk-crypto/src/identities/user.rs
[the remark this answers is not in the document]
andybalaam: > What exactly is the effect of identity_needs_user_approval() == true? It simply displays a warning with no other effect?

Yes. (This is for _unverified_ identities.)

> So what is the behaviour when a verified identity is mismatched?

In legacy/transition mode (they are the same) we will block sending and decrypting messages to/from an identity with a verification violation. I think it will be the same in future too.

> what is the name of the method which enforces this case?

`withdraw_verification` changes a (potentially-violating) identity back into a pinned identity. `has_verification_violation` asks whether this identity has a verification violation.

I find these methods very confusing and am trying to think about how to structure them better.

> Is any of this going to change after the "transition period"?

I don't think so.

> Also, can you suggest wording that would make it accurate?

I went for "UIs should display a warning to the user when encountering an identity which is not verified and is in pin violation."

bindings/matrix-sdk-ffi/src/encryption.rs
andybalaam: Fixed in 562316c4e5

PR #4070, opened 2024-10-04
Support for UTD error codes in `[Sync]TimelineEvent`

richvdh:
This PR takes the `TimelineEventKind` that was added in https://github.com/matrix-org/matrix-rust-sdk/pull/4082, and adds a third enum variant which is specific to unable-to-decrypt events. We update the two methods which are responsible for decrypting encrypted events that have been received from the network (`decrypt_sync_room_event` and `decrypt_room_event`), and make them return a `[Sync]TimelineEvent` of the relevant kind instead of a megolm error.

Since pretty much everything downstream accesses UTD events via `TimelineEventKind::raw()`, we don't have to change much else to keep things working. 

Part of #4048.

### DIFF
--- a/crates/matrix-sdk-ui/src/notification_client.rs
+++ b/crates/matrix-sdk-ui/src/notification_client.rs
@@ -20,10 +20,7 @@ use std::{
 use futures_util::{pin_mut, StreamExt as _};
 use matrix_sdk::{room::Room, Client, ClientBuildError, SlidingSyncList, SlidingSyncMode};
 use matrix_sdk_base::{
-    crypto::{vodozemac, MegolmError},
-    deserialized_responses::TimelineEvent,
-    sliding_sync::http,
-    RoomState, StoreError,
+    deserialized_responses::TimelineEvent, sliding_sync::http, RoomState, StoreError,
 };
 use ruma::{
     assign,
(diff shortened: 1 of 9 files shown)


crates/matrix-sdk-ui/src/timeline/tests/edit.rs
```diff
@@ -165,13 +166,15 @@ async fn test_edit_updates_encryption_info() {
...
     let original_event_id = event_id!("$original_event");
 
-    let mut original_event = event_factory
+    let original_event = event_factory
         .text_msg("**original** message")
         .sender(*ALICE)
         .event_id(original_event_id)
-        .into_sync();
+        .room(room_id)
+        .into_raw_timeline();
```
richvdh, line 177: since we can no longer poke values directly into `original_event.encryption_info` below, the easiest way to do what this test wants is to build an `AnyTimelineEvent` instead of an `AnySyncTimelineEvent`, wrap that into a `DecryptedRoomEvent`, and then use `.into()` to turn *that* into a `SyncTimelineEvent`.

crates/matrix-sdk-ui/src/notification_client.rs
```diff
@@ -216,24 +213,24 @@ impl NotificationClient {
...
+
+                        match new_event.kind {
+                            matrix_sdk::deserialized_responses::TimelineEventKind::UnableToDecrypt {
+                                utd_info, ..} => {
+                                if utd_info.reason.is_missing_room_key() {
+                                    // Decryption error that could be caused by a missing room
+                                    // key; retry in a few.
+                                    wait *= 2;
+                                } else {
+                                    trace!("Event could not be decrypted, but waiting longer is unlikely to help: {:?}", utd_info.reason);
```
stefanceriu, line 226: This feels useful on the nse level, should we maybe promote it to at least info?

crates/matrix-sdk-ui/src/notification_client.rs
```diff
@@ -259,10 +256,21 @@ impl NotificationClient {
...
                 Ok(()) => match room.decrypt_event(raw_event.cast_ref()).await {
-                    Ok(new_event) => {
-                        trace!("Encryption sync managed to decrypt the event.");
-                        Ok(Some(new_event))
-                    }
+                    Ok(new_event) => match new_event.kind {
+                        matrix_sdk::deserialized_responses::TimelineEventKind::UnableToDecrypt {
+                            utd_info, ..
+                        } => {
+                            trace!(
```
stefanceriu, line 263: Maybe here too?

stefanceriu (approved):
Nice, looks good to me 👍

crates/matrix-sdk-ui/src/notification_client.rs
richvdh: Again, the actual error will be logged at warning, and in this case it's just replicating the previous behaviour (note the `trace!("Encryption sync failed to decrypt the event: {err}");` just below).

I can change this if you feel strongly but it's not obvious to me that we should.

stefanceriu: No no, that's quite okay, I didn't check the lower levels to see if we print it anywhere else.

PR #4074, opened 2024-10-04
base: Do not warn when room in m.direct account data is missing

zecakeh:
It can occur that the data is partly outdated and contains rooms that were forgotten. Given that we now update that data after every sync, that creates a lot of noise in the logs for something that is not really a problem.

I demoted it to the DEBUG level, in case someone needs it for debugging purposes, but it would also be fine not to log anything, imo.

### DIFF
--- a/crates/matrix-sdk-base/src/response_processors.rs
+++ b/crates/matrix-sdk-base/src/response_processors.rs
@@ -41,7 +41,7 @@ fn map_info<F: FnOnce(&mut RoomInfo)>(
         f(&mut info);
         changes.add_room(info);
     } else {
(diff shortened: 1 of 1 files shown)


bnjbvr (approved):
Hmm I find it weird that it occurs frequently: that would mean a server is sending us data about rooms we don't know at all (i.e. not from a previous sync, not from the previous sync), but lowering should be fine if we need to investigate.

zecakeh:
The thing is that according to the spec, the `m.direct` event is only changed by clients. So if a DM is forgotten, it's the client's job to remove it from the `m.direct` event and it doesn't seem like we are doing that.

PR #4075, opened 2024-10-04
ui: Do not warn when no reaction to redact was found

zecakeh:
The `handle_reaction_redaction` method is called by `handle_redaction` for every single redaction event that we receive as a first step to check if the redaction matches a reaction.
It means that not finding a reaction to redact is perfectly fine and is not worthy of a warning.


stefanceriu (approved):
Good point 👍

PR #4076, opened 2024-10-06
ReadReceipts: Disable display name change `Notify` events to effect `num_notifications`.

toger5:

Signed-off-by:

### DIFF
--- a/crates/matrix-sdk-base/src/read_receipts.rs
+++ b/crates/matrix-sdk-base/src/read_receipts.rs
@@ -129,8 +129,8 @@ use ruma::{
         poll::{start::PollStartEventContent, unstable_start::UnstablePollStartEventContent},
         receipt::{ReceiptEventContent, ReceiptThread, ReceiptType},
         room::message::Relation,
-        AnySyncMessageLikeEvent, AnySyncTimelineEvent, OriginalSyncMessageLikeEvent,
-        SyncMessageLikeEvent,
+        AnySyncMessageLikeEvent, AnySyncStateEvent, AnySyncTimelineEvent,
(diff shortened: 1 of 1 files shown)


bnjbvr (changes requested):
This is incorrect; `num_notifications` and `num_mentions` are the number of notifications (resp. mentions) as computed by the push rules, with no extra "smart" filtering. Any other kind of "smart" filtering should go into `marks_as_unread` instead.

Please provide a PR description explaining why you think this change is worthy, too; it's really hard to make sense of a text-less PR like this otherwise.

toger5:
@bnjbvr thanks for looking through the draft.

Sorry for the complete absence of context (Timo edits the inital PR message)

bnjbvr:
Ah thank you for the explanation. In this case I think that EXA is wrong in that it should use `num_unread` and not `num_notifications` / `num_mentions`.

toger5:
> Any other kind of "smart" filtering should go into marks_as_unread instead.

Filtering in unread has no impact on the green dot in EX however. So the push rules override any of this filtering.
It is also listed in the unread section of EX (filters at the top of the room list)
So the linked issue, that a user who is updating the displayname, sets `num_notifications` to 1 in all shared rooms (and as a consequence makes them seem as unread to the user) is not fixable with `marks_as_unread`.

I think the best solution would still be to update the default push rules to not include display name changes

bnjbvr:
It sounds like modifying the default push rules would be the way to go, then indeed: it would make all clients behave better, and not have to implement custom workarounds like this one IMO.

toger5:
I added the according EW issue and will close this PR draft. It is liked as a reference to what would be needed if we dont patch EW.

PR #4080, opened 2024-10-07
ffi & sdk: add room knocking to `Client`

jmartinesp:
## Changes

- Add `Client::knock` to both the binding and the SDK.
- Add `Client::room_knocked` to mark a room as knocked. I more or less copied the implementation from `room_joined`, so I'm not sure if some of the internal operations are not needed.
- Added `RoomState::Knocked` and `RoomStateFilter::KNOCKED` enum cases.

Signed-off-by:


andybalaam (approved):
Looks good, thanks!

PR #4039, opened 2024-09-27
chore(cargo): Update `ruma`

Hywan:
This patch updates `ruma` to include:

* https://github.com/ruma/ruma/pull/1913
* https://github.com/ruma/ruma/pull/1914

PR #4081, opened 2024-10-07
Media: Add support for media captions.

pixlwave:
This PR makes 2 changes:

- Update Ruma to include https://github.com/ruma/ruma/pull/1919 and https://github.com/ruma/ruma/pull/1921 so that computed filenames/captions are available for audio, file, image and video messages.
- Update the FFI mappings to include captions and use the computed filename.

### DIFF
--- a/bindings/matrix-sdk-ffi/src/ruma.rs
+++ b/bindings/matrix-sdk-ffi/src/ruma.rs
@@ -277,7 +277,7 @@ impl TryFrom<MessageType> for RumaMessageType {
                     RumaImageMessageEventContent::new(content.body, (*content.source).clone())
                         .info(content.info.map(Into::into).map(Box::new));
                 event_content.formatted = content.formatted.map(Into::into);
-                event_content.filename = content.filename;
+                event_content.filename = content.raw_filename;
                 Self::Image(event_content)
             }
             MessageType::Audio { content } => {
@@ -285,7 +285,7 @@ impl TryFrom<MessageType> for RumaMessageType {
                     RumaAudioMessageEventContent::new(content.body, (*content.source).clone())
                         .info(content.info.map(Into::into).map(Box::new));
                 event_content.formatted = content.formatted.map(Into::into);
-                event_content.filename = content.filename;
+                event_content.filename = content.raw_filename;
                 Self::Audio(event_content)
             }
             MessageType::Video { content } => {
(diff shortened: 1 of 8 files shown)


bindings/matrix-sdk-ffi/src/ruma.rs
```diff
@@ -277,31 +277,31 @@ impl TryFrom<MessageType> for RumaMessageType {
                     RumaImageMessageEventContent::new(content.body, (*content.source).clone())
                         .info(content.info.map(Into::into).map(Box::new));
                 event_content.formatted = content.formatted.map(Into::into);
-                event_content.filename = content.filename;
+                event_content.filename = Some(content.filename);
```
pixlwave, line 280: I'm not 100% about these mappings going back to Ruma. Is it important to store the non-computed filename in the FFI type so that we can pass that back here, or is that an unnecessary step?

jmartinesp: I think this is technically only used for edits and media events can't have them AFAICT, so it might not be needed. However, I don't see the harm in passing it around in case we might need it in the future.

jmartinesp (approved):
LGTM, thanks!

jmartinesp:
The tests need some love though 😅 .

bindings/matrix-sdk-ffi/src/ruma.rs
pixlwave: I've added it back as `raw_filename`.

pixlwave:
Just updated adding in the `formatted_caption` field too which I forgot about. Let's see if the tests are more happy now 🤞

crates/matrix-sdk-ui/tests/integration/room_list_service.rs
```diff
@@ -537,6 +543,7 @@ async fn test_sync_resumes_from_previous_state() -> Result<(), Error> {
                 "lists": {
                     ALL_ROOMS: {
                         "ranges": [[0, 9]],
+                        "timeline_limit": 0,
```
Hywan, line 546: That is a bug, fixed in https://github.com/matrix-org/matrix-rust-sdk/pull/4039 :-).

pixlwave: Ooops, I thought it was mocks not being fully fleshed out 🙈

PR #4082, opened 2024-10-07
timeline: factor out inner part of `[Sync]TimelineEvent`

richvdh:
As [previously discussed](https://github.com/matrix-org/matrix-rust-sdk/pull/4053#issuecomment-2385782287), we'd like to replace the innards of `SyncTimelineEvent` and `TimelineEvent` with a new enum, so that we can accurately represent the three possible states of (Unencrypted, Decrypted, UnableToDecrypt).

This PR lays the groundwork for that change by pulling out the enum. For now it only supports the Decrypted and PlainText cases: the UTD case is still mapped onto PlainText, and will change in a followup PR. It is essentially a reworking of https://github.com/matrix-org/matrix-rust-sdk/issues/4053.

One potential point of contention is that I've used the same inner type for both `TimelineEvent` and `SyncTimelineEvent`: it seems to me that we never rely on the different type information, so having two implementations felt redundant. (It does mean that `TimelineEvent` and `SyncTimelineEvent` become essentially identical, but replacing one with the other felt like a job for another day.) Anyway, I'm happy to revisit this and add a separate `SyncTimelineEventKind` or somesuch instead.

~~Speaking of which, I'm also not entirely happy with the name `TimelineEventInner`. When we discussed this, poljar spoke of `TimelineFoo` and I've not felt much more inspired since. If we could think of a decent name, I think the struct could usefully be exposed as a public type which might clean up some other bits of code. Suggestions welcome.~~ Now renamed to `TimelineEventKind`.

Unfortunately, this PR is necessarily somewhat large. I've tried to structure it in a way to make it easily reviewable, but let me know if I can help with review.

Part of #4048.

### DIFF
--- a/crates/matrix-sdk-common/src/deserialized_responses.rs
+++ b/crates/matrix-sdk-common/src/deserialized_responses.rs
@@ -537,14 +537,19 @@ pub struct UnableToDecryptInfo {
 
 #[cfg(test)]
 mod tests {
+    use assert_matches::assert_matches;
     use ruma::{
+        event_id,
         events::{room::message::RoomMessageEventContent, AnySyncTimelineEvent},
         serde::Raw,
+        user_id,
     };
     use serde::Deserialize;
     use serde_json::json;
 
-    use super::{SyncTimelineEvent, TimelineEvent, VerificationState};
+    use super::{
+        AlgorithmInfo, EncryptionInfo, SyncTimelineEvent, TimelineEvent, VerificationState,
+    };
     use crate::deserialized_responses::{DeviceLinkProblem, VerificationLevel};
 
     fn example_event() -> serde_json::Value {
@@ -618,5 +623,59 @@ mod tests {
             deserialized.state,
             VerificationState::Unverified(VerificationLevel::UnsignedDevice)
         );
+    }
+
+    #[test]
+    fn sync_timeline_event_serialisation() {
+        let room_event = SyncTimelineEvent {
+            event: Raw::new(&example_event()).unwrap().cast(),
+            encryption_info: Some(EncryptionInfo {
+                sender: user_id!("@sender:example.com").to_owned(),
+                sender_device: None,
+                algorithm_info: AlgorithmInfo::MegolmV1AesSha2 {
+                    curve25519_key: "xxx".to_owned(),
+                    sender_claimed_keys: Default::default(),
+                },
+                verification_state: VerificationState::Verified,
+            }),
+            push_actions: Default::default(),
+            unsigned_encryption_info: None,
+        };
+
+        let serialized = serde_json::to_value(&room_event).unwrap();
+
+        // Test that the serialization is as expected
+        assert_eq!(
+            serialized,
+            json!({
+                "event": {
+                    "content": {"body": "secret", "msgtype": "m.text"},
+                    "event_id": "$xxxxx:example.org",
+                    "origin_server_ts": 2189,
+                    "room_id": "!someroom:example.com",
+                    "sender": "@carl:example.com",
+                    "type": "m.room.message",
+                },
+                "encryption_info": {
+                    "sender": "@sender:example.com",
+                    "sender_device": null,
+                    "algorithm_info": {
+                        "MegolmV1AesSha2": {
+                            "curve25519_key": "xxx",
+                            "sender_claimed_keys": {}
+                        }
+                    },
+                    "verification_state": "Verified",
+                },
+            })
+        );
+
+        // And it can be properly deserialized from the new format.
+        let event: SyncTimelineEvent = serde_json::from_value(serialized).unwrap();
+        assert_eq!(event.event_id(), Some(event_id!("$xxxxx:example.org").to_owned()));
+        assert_matches!(
+            event.encryption_info.unwrap().algorithm_info,
+            AlgorithmInfo::MegolmV1AesSha2 { .. }
+        )
     }
 }
--- a/crates/matrix-sdk-common/src/deserialized_responses.rs
+++ b/crates/matrix-sdk-common/src/deserialized_responses.rs
@@ -305,10 +305,14 @@ pub struct EncryptionInfo {
 #[derive(Clone, Deserialize, Serialize)]
 pub struct SyncTimelineEvent {
     /// The actual event.
-    pub event: Raw<AnySyncTimelineEvent>,
+    #[serde(rename = "event")]
+    inner_event: Raw<AnySyncTimelineEvent>,
+
     /// The encryption info about the event. Will be `None` if the event was not
     /// encrypted.
-    pub encryption_info: Option<EncryptionInfo>,
+    #[serde(rename = "encryption_info")]
+    inner_encryption_info: Option<EncryptionInfo>,
+
     /// The push actions associated with this event.
     #[serde(default, skip_serializing_if = "Vec::is_empty")]
     pub push_actions: Vec<Action>,
@@ -325,7 +329,12 @@ impl SyncTimelineEvent {
     /// This is a convenience constructor for when you don't need to set
     /// `encryption_info` or `push_action`, for example inside a test.
     pub fn new(event: Raw<AnySyncTimelineEvent>) -> Self {
-        Self { event, encryption_info: None, push_actions: vec![], unsigned_encryption_info: None }
+        Self {
+            inner_event: event,
+            inner_encryption_info: None,
+            push_actions: vec![],
+            unsigned_encryption_info: None,
+        }
     }
 
     /// Create a new `SyncTimelineEvent` from the given raw event and push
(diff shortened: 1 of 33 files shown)


crates/matrix-sdk-common/src/deserialized_responses.rs
```diff
@@ -305,10 +305,14 @@ pub struct EncryptionInfo {
 #[derive(Clone, Deserialize, Serialize)]
 pub struct SyncTimelineEvent {
     /// The actual event.
-    pub event: Raw<AnySyncTimelineEvent>,
+    #[serde(rename = "event")]
+    inner_event: Raw<AnySyncTimelineEvent>,
```
bnjbvr: Naming is a bit weird here: why the `inner_` prefix? Could we name this `raw`?

crates/matrix-sdk-common/src/deserialized_responses.rs
```diff
@@ -305,10 +305,14 @@ pub struct EncryptionInfo {
...
+
     /// The encryption info about the event. Will be `None` if the event was not
     /// encrypted.
-    pub encryption_info: Option<EncryptionInfo>,
+    #[serde(rename = "encryption_info")]
+    inner_encryption_info: Option<EncryptionInfo>,
```
bnjbvr: Ditto, not sure about the `inner_` prefix here, could we remove it?

crates/matrix-sdk-common/src/deserialized_responses.rs
```diff
@@ -337,24 +346,56 @@ impl SyncTimelineEvent {
...
+    /// its encryption info. Otherwise, `None`.
+    pub fn encryption_info(&self) -> Option<&EncryptionInfo> {
+        self.inner_encryption_info.as_ref()
+    }
+
+    /// Takes ownership of this `TimelineEvent`, returning the (potentially
+    /// decrypted) Matrix event within.
+    pub fn into_event(self) -> Raw<AnySyncTimelineEvent> {
+        self.inner_event
+    }
```
bnjbvr: This should be renamed `into_inner` for consistency with the other method below, or `into_raw` if you follow my above suggestion.

crates/matrix-sdk-common/src/deserialized_responses.rs
```diff
@@ -337,24 +346,56 @@ impl SyncTimelineEvent {
...
+
+    /// Replace the Matrix event within this event. Used to handle redaction.
+    pub fn replace_inner(&mut self, event: Raw<AnySyncTimelineEvent>) {
```
bnjbvr, line 362: I'd call this plain `set_inner` (or `set_raw`).

crates/matrix-sdk-common/src/deserialized_responses.rs
```diff
@@ -464,7 +464,30 @@ impl TimelineEvent {
...
+    /// If the event was a decrypted event that was successfully decrypted, get
+    /// its encryption info. Otherwise, `None`.
+    pub fn encryption_info(&self) -> Option<&EncryptionInfo> {
+        self.inner_encryption_info.as_ref()
+    }
+
+    /// Takes ownership of this `TimelineEvent`, returning the (potentially
+    /// decrypted) Matrix event within.
+    pub fn into_event(self) -> Raw<AnyTimelineEvent> {
+        self.inner_event
```
bnjbvr: Same comments apply here.

crates/matrix-sdk-common/src/deserialized_responses.rs
```diff
@@ -432,103 +396,195 @@ impl From<DecryptedRoomEvent> for SyncTimelineEvent {
...
         }
-        s.maybe_field("unsigned_encryption_info", unsigned_encryption_info);
         s.finish()
     }
 }
 
+/// The event within a [`TimelineEvent`] or [`SyncTimelineEvent`], together with
+/// encryption data.
+#[derive(Clone, Serialize, Deserialize)]
+pub enum TimelineEventInner {
```
bnjbvr, line 520: While I'm not a super fan of the `Kind` name scheme, I do prefer it over `Inner`, and think that an enum called `TimelineEventKind` would be better than `Inner` here (especially in a public API). Other ideas that make it clearer that it's about the kind of encryption may work nicely too.

crates/matrix-sdk-common/src/deserialized_responses.rs
```diff
@@ -300,106 +300,79 @@ pub struct EncryptionInfo {
...
 #[cfg(not(tarpaulin_include))]
 impl fmt::Debug for SyncTimelineEvent {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        let SyncTimelineEvent {
-            inner_event,
-            inner_encryption_info,
-            push_actions,
-            unsigned_encryption_info,
-        } = self;
+        let SyncTimelineEvent { inner, push_actions } = self;
```
bnjbvr, line 370: Maybe this custom `Debug` ain't needed anymore, and we can derive Debug on `SyncTimelineEvent` now instead?

crates/matrix-sdk-common/src/deserialized_responses.rs
```diff
@@ -432,103 +396,195 @@ impl From<DecryptedRoomEvent> for SyncTimelineEvent {
...
 #[cfg(not(tarpaulin_include))]
 impl fmt::Debug for TimelineEvent {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        let TimelineEvent {
-            inner_event,
-            inner_encryption_info,
-            push_actions,
-            unsigned_encryption_info,
-        } = self;
+        let TimelineEvent { inner, push_actions } = self;
```
bnjbvr, line 505: ditto, could we just derive `Debug` on `TimelineEvent` now?

crates/matrix-sdk-common/src/deserialized_responses.rs
```diff
@@ -300,106 +300,79 @@ pub struct EncryptionInfo {
...
         event: Raw<AnySyncTimelineEvent>,
         push_actions: Vec<Action>,
     ) -> Self {
-        Self {
-            inner_event: event,
-            inner_encryption_info: None,
-            push_actions,
-            unsigned_encryption_info: None,
-        }
+        Self { inner: TimelineEventInner::PlainText { event }, push_actions }
```
bnjbvr, line 334: Are we sure it's only used for plain text events, even in testing?

bnjbvr (changes requested):
Looks good; I'm rather unsure about some of the implicit conversion to a plain text event. Good job on the custom deserializer, too bad we couldn't use an untagged enum (I seem to recall this worked in the past…).

crates/matrix-sdk-common/src/deserialized_responses.rs
richvdh: I agree it's a bit of a mess in this transitional state, but since it's all being changed again in [a later commit](https://github.com/matrix-org/matrix-rust-sdk/pull/4082/commits/fde43c93dd402fc0fe243614c62d48cd9168382c#diff-f34dc4bbd38062a031d0c33f7164f950b1735b9b8580d8a6efcb046a9626978cL307-R309), I didn't worry too much about the names here.

crates/matrix-sdk-common/src/deserialized_responses.rs
bnjbvr: No, the custom Debug implementation was there for adding the debug wrapper for the event, I don't think anyone cared about the custom formatting for `push_actions` that much.

crates/matrix-sdk-common/src/deserialized_responses.rs
```diff
@@ -300,106 +300,79 @@ pub struct EncryptionInfo {
...
     /// decrypted) Matrix event within.
     pub fn into_event(self) -> Raw<AnySyncTimelineEvent> {
-        self.inner_event
+        self.inner.into_event()
     }
 
     /// Replace the Matrix event within this event. Used to handle redaction.
     pub fn replace_inner(&mut self, event: Raw<AnySyncTimelineEvent>) {
-        self.inner_event = event;
+        self.inner = TimelineEventInner::PlainText { event };
```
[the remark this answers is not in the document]
richvdh: I think it is correct, yes.

This case is only used in one very specific case: updating an existing event (which may or may not have been encrypted), when handling a redaction event. As you know, redaction removes all the content of an event, which means that, after redaction, it cannot possibly be successfully decrypted.

We could debate whether a redacted encrypted event is better classified as plaintext, unable-to-decrypt, or something else entirely, but this change preserves the existing functionality which is what I'm seeking to do in this PR.

crates/matrix-sdk-common/src/deserialized_responses.rs
richvdh: Well, the fact is that the old code was setting `inner_encryption_info` (previously `encryption_info`) to `None`, which is pretty much the definition of a plaintext event, or at least one we're choosing to treat as such (which currently includes events we failed to decrypt).

Part of our problem here is that, for now, we're using `PlainText` to represent both actual plaintext events, and also events that we failed to decrypt. I think that's a fair representation of the way the code currently works, but does lead to a slightly confusing situation.

I'm working on a followup PR which will add a third entry to the `TimelineEventKind` enum, and we might need to figure out which calls to these methods actually want `PlainText`, and which want `UnableToDecrypt` then.

crates/matrix-sdk-common/src/deserialized_responses.rs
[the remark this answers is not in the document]
bnjbvr: Wouldn't that result in messages being redacted appearing as insecure in an e2ee chat? (think it's already the case, fwiw, so if you're fine with this status quo for this PR and intend to change it later, let's do that)

I'm not satisfied with this invariant that it's only used for redaction, appearing only in the doc comment. Can we do one of these two things:

1. rename the function to `replace_as_redacted` and do something super specific? (maybe we don't even need the raw event anymore in this case?)
2. instead of this method, we provide `raw_mut(&mut self) -> &mut Raw<AnySyncTimelineEvent>` and the callers do whatever they want with the mutable ref? (including: replacing the inner event; but that's an explicit decision done at the caller's call-site)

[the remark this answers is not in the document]
richvdh: > Wouldn't that result in messages being redacted appearing as insecure in an e2ee chat? (think it's already the case, fwiw, so if you're fine with this status quo for this PR and intend to change it later, let's do that)

Yes.

> I'm not satisfied with this invariant that it's only used for redaction, appearing only in the doc comment. 

Fair. Having renamed `TimelineEventInner` to `TimelineEventKind`, we can just make `kind` public, and then the caller can replace `kind` directly, which makes the whole problem go away.

crates/matrix-sdk-common/src/deserialized_responses.rs
richvdh: oh, I'm an idiot, and forgot that Rust allowed you to have fields and methods with the same name, hence the awkward name for this and `inner_encryption_info`. I think this is moot because the field goes away later anyway, but sorry for the unnecessary churn.

crates/matrix-sdk-common/src/deserialized_responses.rs
richvdh: As above: yes, `encryption_info` would have been a better name, but it's going away again anyway.

crates/matrix-sdk-common/src/deserialized_responses.rs
richvdh: 3a0673f69 renames to `into_raw`

crates/matrix-sdk-common/src/deserialized_responses.rs
richvdh: ba762a895 removes this altogether

crates/matrix-sdk-common/src/deserialized_responses.rs
richvdh: 7f08892b0 renames to `into_raw`

crates/matrix-sdk-common/src/deserialized_responses.rs
richvdh: 35e1bf46e renames to TimelineEventKind

crates/matrix-sdk-common/src/deserialized_responses.rs
richvdh: 6e40211af removes the explicit `Debug` impl

crates/matrix-sdk-common/src/deserialized_responses.rs
richvdh: fixed in 6e40211af

crates/matrix-sdk-common/src/deserialized_responses.rs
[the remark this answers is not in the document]
richvdh: https://github.com/matrix-org/matrix-rust-sdk/commit/ba762a895e284d03877b306b45d713dfdb3dd105 removes this, which solves the problem I think.

richvdh:
@bnjbvr please take another look?

bnjbvr (approved):
Thanks, LGTM! Maybe worth a note in a CHANGELOG, or a `Changelog` trailer in the commits (see also our [future awesome release process](https://github.com/matrix-org/matrix-rust-sdk/pull/4097) for details).

crates/matrix-sdk-common/src/deserialized_responses.rs
Hywan: You're not an idiot.

PR #4086, opened 2024-10-07
ci: speed up iOS bindings tests by building them on the `dev` profile

stefanceriu:
- speed regression introduced when switching the default bindings profile to `reldbg` in #4020


bnjbvr (approved):
Thanks! Let's see if this helps… (likely not on the first commit, but might on subsequent incremental rebuilds)

PR #4087, opened 2024-10-07
ci: use macos-14 runners for the swift related tasks

bnjbvr:
macos-12 runners are going [to be deprecated](https://github.com/actions/runner-images/issues/10721), so we should try to upgrade any time soon.

cc @stefanceriu @pixlwave


stefanceriu (approved):
🚀 it! (but maybe get rid of my commit first 😁)

PR #4090, opened 2024-10-08
sdk-base: add `RoomInfo::prev_room_state` property

jmartinesp:
This is useful for the knocking feature since we'll be able to differentiate between rooms that you were just invited from rooms that you knocked and then were granted access, or rooms that you left and rooms where your knocking attempt was rejected.

The `mark_room_as_*` functions have been updated so they reuse the same `set_state` function underneath, which only updates the previous state if the new one doesn't match.

Needs https://github.com/matrix-org/matrix-rust-sdk/pull/4080.

Signed-off-by:

### DIFF
--- a/crates/matrix-sdk-base/src/rooms/normal.rs
+++ b/crates/matrix-sdk-base/src/rooms/normal.rs
@@ -182,7 +182,7 @@ impl RoomSummary {
 }
 
 /// Enum keeping track in which state the room is, e.g. if our own user is
-/// joined, invited, or has left the room.
+/// joined, RoomState::Invited, or has left the room.
 #[derive(Clone, Copy, Debug, Eq, PartialEq, Serialize, Deserialize)]
 pub enum RoomState {
     /// The room is in a joined state.
@@ -273,6 +273,11 @@ impl Room {
     /// Get the state of the room.
     pub fn state(&self) -> RoomState {
         self.inner.read().room_state
+    }
+
+    /// Get the previous state of the room, if it had any.
+    pub fn prev_state(&self) -> Option<RoomState> {
+        self.inner.read().prev_room_state
     }
 
     /// Whether this room's [`RoomType`] is `m.space`.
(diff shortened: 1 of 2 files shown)


crates/matrix-sdk-base/src/rooms/normal.rs
```diff
@@ -2061,7 +2078,8 @@ mod tests {
         let info: RoomInfo = serde_json::from_value(info_json).unwrap();
 
         assert_eq!(info.room_id, room_id!("!gda78o:server.tld"));
-        assert_eq!(info.room_state, RoomState::Invited);
```
andybalaam, line 2064: I'm not a big fan of importing enum variants like this. Is this something you've seen elsewhere in the SDK? If not, please can you change these back to e.g. `RoomState::Joined` ?

crates/matrix-sdk-base/src/rooms/normal.rs
```diff
@@ -415,9 +420,7 @@ impl Room {
     #[instrument(skip_all, fields(room_id = ?self.room_id))]
     pub async fn is_direct(&self) -> StoreResult<bool> {
         match self.state() {
-            RoomState::Joined | RoomState::Left => {
```
andybalaam, line 418: Looks like nothing changed here - did you mean to do this, or did your editor get clever on you?

andybalaam (approved):
A couple of minor comments, all good though!

crates/matrix-sdk-base/src/rooms/normal.rs
jmartinesp: Maybe, but I think I just run the fixup style task and it re-formatted the code here after removing the `RoomState::` prefix, then adding it back.

crates/matrix-sdk-base/src/rooms/normal.rs
jmartinesp: Also, in these test `Invited` was already imported directly without the `RoomState` prefix so I just followed the convention here. I'll revert it though, being more explicit seems less prone to errors.

PR #4091, opened 2024-10-08
SDK: Pass the filename to get_media_file instead of the body.

pixlwave:
In https://github.com/element-hq/element-x-ios/pull/3375 I noticed that we request a file using the event's `body` which doesn't seem to make sense to me now that #4081 (and https://github.com/ruma/ruma/pull/1921) is available.

This PR updates the `get_media_file` methods in the SDK and FFI to both take the filename instead.

### DIFF
--- a/crates/matrix-sdk/src/media.rs
+++ b/crates/matrix-sdk/src/media.rs
@@ -176,8 +176,13 @@ impl Media {
     ///
     /// * `request` - The `MediaRequest` of the content.
     ///
+    /// * `filename` - The filename specified in the event. It is suggested to
+    ///   use the `filename()` method on the event's content instead of using
+    ///   the `filename` field directly. If not provided, a random name will be
+    ///   generated.
+    ///
     /// * `content_type` - The type of the media, this will be used to set the
-    ///   temporary file's extension.
+    ///   temporary file's extension when one isn't included in the filename.
     ///
     /// * `use_cache` - If we should use the media cache for this request.
     ///
@@ -189,7 +194,7 @@ impl Media {
     pub async fn get_media_file(
         &self,
         request: &MediaRequest,
-        body: Option<String>,
+        filename: Option<String>,
         content_type: &Mime,
         use_cache: bool,
         temp_dir: Option<String>,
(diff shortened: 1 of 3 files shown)


crates/matrix-sdk/src/media.rs
```diff
@@ -198,49 +203,51 @@ impl Media {
 
         let inferred_extension = mime2ext::mime2ext(content_type);
 
-        let body_path = body.as_ref().map(Path::new);
-        let filename = body_path.and_then(|f| f.file_name().and_then(|f| f.to_str()));
-        let filename_with_extension = body_path.and_then(|f| {
+        let filename_as_path = filename.as_ref().map(Path::new);
+        let filename_without_extension =
+            filename_as_path.and_then(|f| f.file_name().and_then(|f| f.to_str()));
```
jmartinesp, line 208: Maybe rename this first `|f|` to `|path|` so it's clear we get just the file name and not the extension because we're asking that from the  `Path`?

crates/matrix-sdk/src/media.rs
```diff
@@ -198,49 +203,51 @@ impl Media {
...
+        let filename_with_extension = filename_as_path.and_then(|f| {
```
jmartinesp, line 209: Same here.

jmartinesp (approved):
LGTM, a couple of naming suggestions.

crates/matrix-sdk/src/media.rs
bnjbvr: drive-by nit: use an external `.and_then` instead of putting it inside the first `and_then`:

```suggestion
            filename_as_path.and_then(|f| f.file_name()).and_then(|f| f.to_str());
```

crates/matrix-sdk/src/media.rs
```diff
@@ -198,49 +203,51 @@ impl Media {
...
             if f.extension().is_some() {
                 f.file_name().and_then(|f| f.to_str())
```
bnjbvr, line 211: ditto: same here

Actually, since you have two code paths that deal with `if the filename as path exists`, use `if let Some(filename_path)` instead, and return a pair tuple?

crates/matrix-sdk/src/media.rs
pixlwave: Ah yep makes sense. Also looking at this code, renaming `filename` to `filename_without_extension` is wrong. Both filenames are the same if there's an extension and if there isn't then `filename_with_extension` is `None`.

I've updated the code to use a bool `filename_has_extension` and not a second copy of the string.

crates/matrix-sdk/src/media.rs
pixlwave: Ah yeah, that sounds more sane, will update 👍

pixlwave: Updated, thanks that's much clearer.

PR #4092, opened 2024-10-08
sdk-ui: allow sent local events to be redacted in `redact_by_id`

jmartinesp:
When writing `Timeline::redact_by_id` I thought the case where a local event had a remote handle was invalid, but it's a valid case when the local event was sent to the HS and now has an event id. Since the code I added returned an error when you tried to redact one of these events, we need to fix it.

I also added some tests for this use case.

Signed-off-by:


crates/matrix-sdk-ui/tests/integration/timeline/mod.rs
```diff
@@ -307,6 +307,83 @@ async fn test_redact_message() {
...
+
+    // Check the new item is sent
+    assert_let!(Some(VectorDiff::Set { index, value: second }) = timeline_stream.next().await);
+    assert_eq!(index, 2);
+
+    let second = second.as_event().unwrap();
+    assert_matches!(second.send_state(), Some(EventSendState::Sent { .. }));
+
+    // Redacting a remote event works.
+    mock_redaction(second.event_id().unwrap()).mount(&server).await;
```
bnjbvr, line 381: Add an `expect(1)` here, to make sure it's called in the call below.

crates/matrix-sdk-ui/tests/integration/timeline/mod.rs
```diff
@@ -381,6 +458,83 @@ async fn test_redact_by_id_message() {
...
```
bnjbvr, line 532: ditto

bnjbvr (changes requested):
Can you add comments in the test of the form `When I do X, then Y, then Z` etc., or at least add a high-level description of the test as an initial comment at the top? It's a bit hard to read right now…

Also we should make use of the `with_timeout` variants of the macro more; all these `next().await` will cause a 60s time out at the CI step level otherwise.

jmartinesp:
Simplified and added some more clarifications in the tests in https://github.com/matrix-org/matrix-rust-sdk/pull/4092/commits/ef5283edfb400b33710eeeebe42fad9d280f67a5, as well as adding the `.expect(1)` check for the mocked endpoints.

crates/matrix-sdk-ui/tests/integration/timeline/mod.rs
```diff
@@ -307,6 +308,76 @@ async fn test_redact_message() {
...
+        Some(VectorDiff::PushFront { value: day_divider }) = timeline_stream.next()
+    );
+    assert!(day_divider.is_day_divider());
+
+    // Let the send queue send the event so it's now a local event with a remote
+    // handle (event_id).
+    yield_now();
+
+    // We receive an update in the timeline from the send queue.
+    assert_let_timeout!(Some(VectorDiff::Set { index, value: item }) = timeline_stream.next());
```
bnjbvr, line 365: Do we need the yield_now now that we're using `assert_let_timeout`?

crates/matrix-sdk-ui/tests/integration/timeline/mod.rs
```diff
@@ -381,6 +452,76 @@ async fn test_redact_by_id_message() {
...
+
+    // As well as a day divider.
+    assert_let_timeout!(
+        Some(VectorDiff::PushFront { value: day_divider }) = timeline_stream.next()
+    );
+    assert!(day_divider.is_day_divider());
+
+    // Let the send queue send the event so it's now a local event with a remote
+    // handle (event_id).
+    yield_now();
```
bnjbvr, line 506: (same here)

bnjbvr (approved):
Thanks for the test code comments, it's a joy to review :heart:

crates/matrix-sdk-ui/tests/integration/timeline/mod.rs
jmartinesp: Oh, it seems like we don't. I'm removing them, thanks!

PR #4094, opened 2024-10-08
ffi: implement our own `export` macro that always sets up the correct async runtime

bnjbvr:
The first commit implements new macros:

- `matrix_sdk_ffi_macros::export` must be used for all *sync* functions and impl blocks. It will complain if anything is async (the exported function or any function in the exported impl block).
- `matrix_sdk_ffi_macros::export_async` must be used for async functions and impl blocks. Currently, uniffi already complains if the exported item doesn't involve async, so no code had to be added for that.

Screenshot From 2024-10-08 14-59-14
Screenshot From 2024-10-08 14-57-20

The second commit prevents use of the `uniffi::export` macro by making it into a clippy unauthorized macro, but then the local `#[allow()]` statement doesn't seem to be taken into account when using the `uniffi::export` macro internally :( I've opened a thread on the Rust Zulip, to ask if I've missed something by chance or not.

## Ways to move forward

1. only take the first commit
2. wait for a conclusion over the Clippy side
3. replace the second commit by a manual check in xtask, using e.g. grep :nerd_face:

### DIFF
--- a/testing/matrix-sdk-ffi-macros/Cargo.toml
+++ b/testing/matrix-sdk-ffi-macros/Cargo.toml
@@ -1,0 +1,24 @@
+[package]
+description = "Helper macros to write FFI bindings"
+edition = "2021"
+homepage = "https://github.com/matrix-org/matrix-rust-sdk"
+keywords = ["matrix", "chat", "messaging", "ruma"]
+license = "Apache-2.0"
+name = "matrix-sdk-ffi-macros"
+readme = "README.md"
+repository = "https://github.com/matrix-org/matrix-rust-sdk"
+rust-version = { workspace = true }
+version = "0.7.0"
+
+[lib]
+proc-macro = true
+test = false
+doctest = false
+
+[dependencies]
+proc-macro2 = "1.0.86"
+quote = "1.0.18"
+syn = { version = "2.0.43", features = ["full", "extra-traits"] }
+
+[lints]
+workspace = true
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3136,6 +3136,7 @@ dependencies = [
  "mas-oidc-client",
  "matrix-sdk-base",
  "matrix-sdk-common",
+ "matrix-sdk-ffi-macros",
  "matrix-sdk-indexeddb",
  "matrix-sdk-sqlite",
  "matrix-sdk-test",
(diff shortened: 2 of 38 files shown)


poljar (approved):
I think we can wait a bit for the second commit, otherwise I would prefer to merge only the first commit. It's unlikely that anybody will use the wrong macro, people just copy paste things.

testing/matrix-sdk-ffi-macros/Cargo.toml
```diff
@@ -0,0 +1,24 @@
...
+description = "Helper macros to write FFI bindings"
+edition = "2021"
+homepage = "https://github.com/matrix-org/matrix-rust-sdk"
+keywords = ["matrix", "chat", "messaging", "ruma"]
+license = "Apache-2.0"
+name = "matrix-sdk-ffi-macros"
+readme = "README.md"
+repository = "https://github.com/matrix-org/matrix-rust-sdk"
+rust-version = { workspace = true }
+version = "0.7.0"
```
Hywan, line 11: authot is missing :-).

Hywan (approved):
I like the first commit, and I think it's fine, we may not need the second if it's too much work. Well done, thanks!

testing/matrix-sdk-ffi-macros/Cargo.toml
bnjbvr: We don't usually set it, or we set it to Damir. Would need to do something more general for all the crates we have :)

bnjbvr: like setting all to "The Rust Matrix SDK developers" or something like this

bnjbvr:
Dropped the second commit (since it's a [bug in Clippy](https://github.com/rust-lang/rust-clippy/issues/13521)) and enabled auto-merge. Thanks for the quick reviews!

testing/matrix-sdk-ffi-macros/Cargo.toml
```diff
@@ -0,0 +1,24 @@
+[package]
+description = "Helper macros to write FFI bindings"
+edition = "2021"
```
bmarty, line 3: 2024?

bnjbvr: image

Hywan: @bmarty You can learn what a Rust edition is by reading https://doc.rust-lang.org/edition-guide/editions/index.html :-).

PR #4095, opened 2024-10-08
fix(base): Fix state events handling in sliding sync

Hywan:
Sliding sync expects all state events to be in `required_state`. State
events in `timeline` **must be ignored**. However, in sync v2, state
events in `timeline` **must be handled**.

Two patches:

* From sliding sync response handler, we don't gather state events
from `timeline` anymore,
  
* In the same handler, both sliding sync and sync v2 uses the same
`handle_timeline` method. I add an argument to ignore state
events. This is not ideal, but it's a temporary solution as a first
step. The next step is to refactor this code, but let's start easy.

Other patches updates the tests.

Next step: clean this flow/handler just like https://github.com/matrix-org/matrix-rust-sdk/pull/4061 did.

### DIFF
--- a/crates/matrix-sdk-base/src/client.rs
+++ b/crates/matrix-sdk-base/src/client.rs
@@ -384,6 +384,7 @@ impl BaseClient {
         room: &Room,
         limited: bool,
         events: Vec<Raw<AnySyncTimelineEvent>>,
+        ignore_state_events: bool,
         prev_batch: Option<String>,
         push_rules: &Ruleset,
         user_ids: &mut BTreeSet<OwnedUserId>,
@@ -402,7 +403,7 @@ impl BaseClient {
                 Ok(e) => {
                     #[allow(clippy::single_match)]
                     match &e {
-                        AnySyncTimelineEvent::State(s) => {
+                        AnySyncTimelineEvent::State(s) if !ignore_state_events => {
                             match s {
                                 AnySyncStateEvent::RoomMember(member) => {
                                     Box::pin(ambiguity_cache.handle_event(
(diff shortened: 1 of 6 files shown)


bnjbvr:
For what it's worth: this patch breaks support for the sliding sync proxy, which behaves differently and expected that we did handle those timeline events, as far as I recall (hence the change in the test expectation you've tweaked). It might not be the first patch that did break the sliding sync proxy, but it's at least one. Should we merge it only when we officially deprecate support for the sliding sync proxy?

poljar:
> For what it's worth: this patch breaks support for the sliding sync proxy, which behaves differently and expected that we did handle those timeline events, as far as I recall (hence the change in the test expectation you've tweaked). It might not be the first patch that did break the sliding sync proxy, but it's at least one. Should we merge it only when we officially deprecate support for the sliding sync proxy?

Hmm, AFAIR we shouldn't handle state events for the proxy case either. I think this was raised multiple times, but we failed to do something about it. @kegsay could you please confirm or deny this?

Hywan:
[Kegan has confirmed](https://matrix.to/#/!cnVVNLKqgUzNTOFQkz:matrix.org/$CHuXOjP1E-sfdQKWvC-AXvkFE9mI5sg-_IlWIYpxYsY?via=element.io&via=matrix.org&via=jki.re) the proxy also requires to look inside the `required_state` only.

crates/matrix-sdk-base/src/client.rs
```diff
@@ -384,6 +384,7 @@ impl BaseClient {
         room: &Room,
         limited: bool,
         events: Vec<Raw<AnySyncTimelineEvent>>,
+        ignore_state_events: bool,
```
poljar, line 387: Not a fan of this boolean, but ok it's not as complex as I imagined it would be.

Hywan: Me neither! I'm gonna refactor this in a next PR (see the PR description where I already mention this).

PR #4098, opened 2024-10-09
sdk: Fixes to `Room::forget`

### LINKED ISSUE #4078: Calling `Room::forget` should remove the room from `m.direct` account data.
This was brought up in https://github.com/matrix-org/matrix-rust-sdk/pull/4074#issuecomment-2396156907. According to the spec, the homeserver has no obligation to keep the `m.direct` account data up-to-date. That means that we can have the `m.direct` account data with rooms that are unknown because they were left then forgotten.

To fix that we should remove the room from the `m.direct` account data when `Room::forget` is called.

### LINKED ISSUE #4079: Calling `Room::forget` should remove room from in-memory map in `Store`
`Room::forget` calls `store.remove_room()`, but since this is the method from the `StateStore` trait, the room is only removed from the database, not from the in-memory `Store::rooms` map.

There should maybe be a wrapper method on `Store` that removes it both from the database and the in-memory map.

zecakeh:
When forgetting rooms:
- Remove direct rooms from the `m.direct` account data,
- Remove rooms from the in-memory list.

Fixes #4078.\
Fixes #4079.

### DIFF
--- a/crates/matrix-sdk-base/src/client.rs
+++ b/crates/matrix-sdk-base/src/client.rs
@@ -1403,6 +1403,17 @@ impl BaseClient {
         self.store.room(room_id)
     }
 
+    /// Forget the room with the given room ID.
+    ///
+    /// The room will be dropped from the room list and the store.
+    ///
+    /// # Arguments
+    ///
+    /// * `room_id` - The id of the room that should be forgotten.
+    pub async fn forget_room(&self, room_id: &RoomId) -> StoreResult<()> {
+        self.store.forget_room(room_id).await
+    }
+
     /// Get the olm machine.
     #[cfg(feature = "e2e-encryption")]
     pub async fn olm_machine(&self) -> RwLockReadGuard<'_, Option<OlmMachine>> {
--- a/crates/matrix-sdk-base/src/store/observable_map.rs
+++ b/crates/matrix-sdk-base/src/store/observable_map.rs
@@ -130,6 +130,18 @@ mod impl_non_wasm32 {
         pub(crate) fn stream(&self) -> (Vector<V>, impl Stream<Item = Vec<VectorDiff<V>>>) {
             self.values.subscribe().into_values_and_batched_stream()
         }
+
+        /// Remove a `V` value based on their ID, if it exists.
+        ///
+        /// Returns the removed value.
+        pub(crate) fn remove<L>(&mut self, key: &L) -> Option<V>
+        where
+            K: Borrow<L>,
+            L: Hash + Eq + ?Sized,
+        {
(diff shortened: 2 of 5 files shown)


crates/matrix-sdk-base/src/client.rs
```diff
@@ -1403,6 +1403,15 @@ impl BaseClient {
         self.store.room(room_id)
     }
 
+    /// Forget the room with the given room ID.
```
Hywan, line 1406: Might be useful to explain how it differs from leaving a room?

crates/matrix-sdk-base/src/store/observable_map.rs
```diff
@@ -249,6 +272,30 @@ mod tests {
...
+        assert_eq!(map.get(&'a'), Some(&'e'));
+        assert_eq!(map.get(&'b'), Some(&'f'));
+        assert!(map.get(&'c').is_none());
+
+        // remove one item
+        map.remove(&'b');
+
+        assert_eq!(map.get(&'a'), Some(&'e'));
+        assert_eq!(map.get(&'b'), None);
+        assert_eq!(map.get(&'c'), None);
```
Hywan, line 296: Can you also `map.remove(&'d')` —which doesn't exist— to assert it returns `None` please?

Hywan (changes requested):
It looks good, as always, thanks!

A small feedback about tests, otherwise it's good.

crates/matrix-sdk-base/src/store/observable_map.rs
zecakeh: Ah good catch!

crates/matrix-sdk-base/src/client.rs
zecakeh: Is what I added okay?

Hywan: But it's sending particular request. I think a user may not want to read the spec to understand the difference between `room.leave()` and `room.forget()`. Can we clarify that please?

crates/matrix-sdk-base/src/store/observable_map.rs
Hywan: Nice, thanks!

crates/matrix-sdk-base/src/client.rs
zecakeh: This is the base SDK, there is not request.

poljar (approved):
I think that this looks fine, can you just get rid of the fixup commit?

zecakeh:
Done.

PR #4100, opened 2024-10-09
refactor(timeline): fuse `edit_by_id` (resp `redact_by_id`) into their ID-less counterparts

### LINKED ISSUE #4093: Use `UniqueId` as a parameter for timeline APIs
I'm opening this issue so we can discuss about what we want to do with the `Timeline` API and `UniqueId` as a replacement for event/transaction ids.

## Context

After #3942 a couple of new `Timeline::edit_by_id` and `Timeline::redact_by_id` functions were added to the codebase, which take a `TimelineEventItemId` (either an `EventId` or a `TransactionId`) as the event's identifier since we can't pass back to the FFI layer the original Rust item and we need to use something to search for the expected item in the timeline.

It seems like this is not the API we want, we'd rather use `UniqueId` instead for the timeline APIs since it's a unique identifier for each timeline item and that way we don't have to track if the event has an id or a transaction id if it's local.

## Proposal

What we want to do, as far as I can tell, is:

- Have most/all `Timeline` functions that modify events in the timeline to use `UniqueId` instead.
- Use this unique id to search for the items in the timeline.
- Remove the duplicate `_by_id` functions.

## Possible problems

- **What happens to functions that already accept only `EventId` or only `TransactionId` parameters, like `Timeline::replied_to_info_from_event_id`?** Should we also make them use `UniqueId` and return an error if the items is not local/remote as expected?
- **Some drafts-related code in the mobile EX clients rely on event or transaction ids to work**, and although most of them can be moved to use unique ids instead, there are some cases, like editing or replying to an event, where this event may not come from the timeline and not have a unique id. In those cases we need some way to ask the client to use `Room::edit` and `Room::reply` instead. This will also be painful for the clients to implement, but there's not much we can do about that.
- Editing *every Timeline test* that uses their APIs to use `UniqueId` is going to be a fun ride.

bnjbvr:
This PR sits on top of #4111 and #4127, and allows us to get rid of the `edit_by_id` / `redact_by_id` variants.

I've added changelog entries \o/

Fixes #4093.

### DIFF
--- a/bindings/matrix-sdk-ffi/src/timeline/mod.rs
+++ b/bindings/matrix-sdk-ffi/src/timeline/mod.rs
@@ -31,7 +31,7 @@ use matrix_sdk::{
     Error,
 };
 use matrix_sdk_ui::timeline::{
-    EventItemOrigin, LiveBackPaginationStatus, Profile, RepliedToEvent, TimelineDetails,
+    self, EventItemOrigin, LiveBackPaginationStatus, Profile, RepliedToEvent, TimelineDetails,
     TimelineUniqueId as SdkTimelineUniqueId,
 };
 use mime::Mime;
@@ -494,11 +494,30 @@ impl Timeline {
         &self,
         event_or_transaction_id: EventOrTransactionId,
         new_content: EditedContent,
-    ) -> Result<bool, ClientError> {
-        self.inner
-            .edit(&event_or_transaction_id.try_into()?, new_content.try_into()?)
+    ) -> Result<(), ClientError> {
+        match self
+            .inner
+            .edit(&event_or_transaction_id.clone().try_into()?, new_content.clone().try_into()?)
             .await
-            .map_err(Into::into)
+        {
+            Ok(true) => Ok(()),
+            Ok(false) | Err(timeline::Error::EventNotInTimeline(_)) => {
+                // If we couldn't edit, assume it was an (remote) event that wasn't in the
+                // timeline, and try to edit it via the room itself.
+                let event_id = match event_or_transaction_id {
+                    EventOrTransactionId::EventId { event_id } => EventId::parse(event_id)?,
+                    EventOrTransactionId::TransactionId { .. } => {
+                        warn!("trying to apply an edit to a local echo that doesn't exist in this timeline, aborting");
+                        return Ok(());
+                    }
+                };
+                let room = self.inner.room();
+                let edit_event = room.make_edit_event(&event_id, new_content.try_into()?).await?;
+                room.send_queue().send(edit_event).await?;
+                Ok(())
+            }
+            Err(err) => Err(err)?,
+        }
     }
 
     pub async fn send_location(
--- a/bindings/matrix-sdk-ffi/src/timeline/mod.rs
+++ b/bindings/matrix-sdk-ffi/src/timeline/mod.rs
@@ -496,7 +496,7 @@ impl Timeline {
         new_content: EditedContent,
     ) -> Result<bool, ClientError> {
         self.inner
-            .edit_by_id(&event_or_transaction_id.try_into()?, new_content.try_into()?)
+            .edit(&event_or_transaction_id.try_into()?, new_content.try_into()?)
             .await
             .map_err(Into::into)
     }
--- a/bindings/matrix-sdk-ffi/src/timeline/mod.rs
+++ b/bindings/matrix-sdk-ffi/src/timeline/mod.rs
@@ -629,11 +629,7 @@ impl Timeline {
         event_or_transaction_id: EventOrTransactionId,
         reason: Option<String>,
     ) -> Result<(), ClientError> {
-        if !self.inner.redact(&(event_or_transaction_id.try_into()?), reason.as_deref()).await? {
-            // TODO make it a hard error instead
-            warn!("Couldn't redact item");
-        }
-        Ok(())
+        Ok(self.inner.redact(&(event_or_transaction_id.try_into()?), reason.as_deref()).await?)
     }
 
     /// Load the reply details for the given event id.
--- a/bindings/matrix-sdk-ffi/src/timeline/mod.rs
+++ b/bindings/matrix-sdk-ffi/src/timeline/mod.rs
@@ -500,8 +500,8 @@ impl Timeline {
             .edit(&event_or_transaction_id.clone().try_into()?, new_content.clone().try_into()?)
             .await
         {
-            Ok(true) => Ok(()),
-            Ok(false) | Err(timeline::Error::EventNotInTimeline(_)) => {
+            Ok(()) => Ok(()),
+            Err(timeline::Error::EventNotInTimeline(_)) => {
                 // If we couldn't edit, assume it was an (remote) event that wasn't in the
                 // timeline, and try to edit it via the room itself.
                 let event_id = match event_or_transaction_id {
--- a/bindings/matrix-sdk-ffi/src/timeline/mod.rs
+++ b/bindings/matrix-sdk-ffi/src/timeline/mod.rs
@@ -610,10 +610,11 @@ impl Timeline {
         event_or_transaction_id: EventOrTransactionId,
         reason: Option<String>,
     ) -> Result<(), ClientError> {
-        self.inner
-            .redact_by_id(&(event_or_transaction_id.try_into()?), reason.as_deref())
-            .await
-            .map_err(Into::into)
+        if !self.inner.redact(&(event_or_transaction_id.try_into()?), reason.as_deref()).await? {
+            // TODO make it a hard error instead
+            warn!("Couldn't redact item");
+        }
+        Ok(())
     }
 
     /// Load the reply details for the given event id.
--- a/crates/matrix-sdk-ui/src/timeline/mod.rs
+++ b/crates/matrix-sdk-ui/src/timeline/mod.rs
@@ -18,7 +18,7 @@
 
 use std::{path::PathBuf, pin::Pin, sync::Arc, task::Poll};
 
-use event_item::{extract_room_msg_edit_content, EventTimelineItemKind, TimelineItemHandle};
+use event_item::{extract_room_msg_edit_content, TimelineItemHandle};
 use eyeball_im::VectorDiff;
 use futures_core::Stream;
 use imbl::Vector;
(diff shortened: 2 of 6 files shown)


crates/matrix-sdk-ui/src/timeline/item.rs
```diff
@@ -18,6 +18,27 @@ use as_variant::as_variant;
...
 use super::{EventTimelineItem, VirtualTimelineItem};
 
+/// Opaque unique identifier for a timeline item.
+///
+/// It is transferred whenever a timeline item is updated. This can be used as a
+/// stable identifier for UI purposes, as well as operations on the event
+/// represented by the item.
+#[derive(Clone, Debug, PartialEq, Eq, Hash)]
+#[cfg_attr(feature = "uniffi", derive(uniffi::Object))]
+pub struct TimelineUniqueId(pub(crate) String);
```
jmartinesp, line 28: I guess this is created as an `Object` instead of `Record` so it can't be instantiated from the clients? I'd rather have a plain type instead of passing a pointer as it would be better for debugging purposes though 🫤 .

crates/matrix-sdk-ui/src/timeline/mod.rs
```diff
@@ -609,16 +558,21 @@ impl Timeline {
...
     pub async fn redact(
         &self,
-        event: &EventTimelineItem,
+        unique_id: &TimelineUniqueId,
         reason: Option<&str>,
     ) -> Result<bool, Error> {
-        match event.handle() {
+        let items = &self.controller.items().await;
+        let Some((_item_pos, item)) = rfind_event_by_uid(&items, unique_id) else {
+            warn!("couldn't find timeline item identified with id");
```
jmartinesp, line 542: Having a `bool` indicating whether the event was redacted or not seemed weird when I was looking at this code, same as when editing: if I ask theSDK to do an operation and it can't, I'd expect the issue to be raised as an error instead of a `false` value, but maybe that's because I'm used to the java way of catching exceptions for unexpected behavior.

bindings/matrix-sdk-ffi/src/timeline/mod.rs
```diff
@@ -1069,27 +1026,26 @@ impl From<matrix_sdk_ui::timeline::EventTimelineItem> for EventTimelineItem {
...
+        let item = Arc::new(item);
+        let debug_info_provider = Arc::new(EventTimelineItemDebugInfoProvider(item.clone()));
+        let shields_provider = Arc::new(EventShieldsProvider(item.clone()));
         let read_receipts =
-            value.read_receipts().iter().map(|(k, v)| (k.to_string(), v.clone().into())).collect();
+            item.read_receipts().iter().map(|(k, v)| (k.to_string(), v.clone().into())).collect();
         Self {
-            is_local: value.is_local_echo(),
-            is_remote: !value.is_local_echo(),
-            event_or_transaction_id: value.identifier().into(),
```
jmartinesp, line 1077: How do we get a unique id without this? Did I miss where we return it?

jmartinesp (commented):
Thanks for the work on this! I have a few questions, especially about how we return this unique id. 

I'll also try to build a version and test it against Android to see how it works.

bindings/matrix-sdk-ffi/src/timeline/mod.rs
jmartinesp: Ah, I see, it's [here](https://github.com/matrix-org/matrix-rust-sdk/pull/4100/files#diff-585addf5b1e9b7be30f6de5cf83d63d84d4f901b98ef4524c703e3ebf14392f1R857).

crates/matrix-sdk-ui/src/timeline/item.rs
bnjbvr: I suppose `Record` means the inner string would be available to all, and the intent here is that the identifier is an opaque thingy. Would you like a "gimme a debug string" helper for debugging purposes?

crates/matrix-sdk-ui/src/timeline/mod.rs
bnjbvr: I'm not closed to changing the API here; I'd rather have it be a different PR though, ideally with an issue explaining the motivation.

crates/matrix-sdk-ui/src/timeline/item.rs
jmartinesp: I think that would be enough, yes. Without an event id or transaction id it could be a bit difficult to keep track of an item if it changed while debugging, with the event id back this should be better and having a way to manually check the unique id would be perfect. Thanks!

bindings/matrix-sdk-ffi/src/timeline/mod.rs
bnjbvr: Added it back in a fixup commit autosquashed.

jmartinesp:
Some things I found while trying to integrate the changes:

- Sometimes we add virtual events of our own to the timeline, so those need a 'fake' timeline id too to reuse the same models. Since we already wrap the `TimelineUniqueId` in a wrapper of our own to be able to create fakes for testing too, this is not that bad for Android, but I don't know how difficult it would be for iOS (@stefanceriu ?).
- For the forward action in the clients (as in, taking the content of an event and sending it in another event, maybe in a different room) we need to get this message like content first, and until now we were just passing the associated event id, fetching the content using `timeline.getEventTimelineItemByEventId` [at the last moment](https://github.com/element-hq/element-x-android/blob/366db4791b1febfed836c88d16a8ffa1f27b2c57/libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/room/RoomContentForwarder.kt#L45), then sending a new message with the contents using the new timeline. However, there is no `timeline.getTimelineItemByUniqueId` exposed so the only way we could do this now is by getting the content at the point where the action starts and passing it around several screens, which is not impossible on Android, but it is difficult. Having the `timeline.getTimelineItemByUniqueId` method would be super helpful.
- This forces us to remove some APIs from our `Room` object wrappers, but I think that's fine on most cases since I don't think they were being used anymore and also they relied on the internal timeline implementation, so we should have used the timeline version anyway.

I'm still trying to adapt our current code to the new APIs, I'll post any other issues I run into.

stefanceriu (commented):
Took me 2 hours to get the app building again but I finally did it. You can see what it took here: https://github.com/element-hq/element-x-ios/compare/develop...stefan/TimelineUniqueId

I found problems in:
- methods for retring using txnids (room.tryResend, room.ignoreDeviceTrustAndResend, room.withdrawVerificationAndResend)
- message drafts
- voice message audio players: we need to keep their states in between timeline resets/updates. The IDs need to be stable.
- sending read receipts
- editing polls
- ending polls
- selecting poll options
- copying permalinks (tell if it's been sent, use its event id)
- replying
- pinning, unpinning message and viewing them in the timeline. Checking if a particular timeline item is pinned.
- reporting content
- composer states (editing, replying)
- focusing a timeline on a particular event id
- fetching event details through `fetchDetailsForEvent(eventId)`

While some of them can probably be rewritten some other way even now I don't think this PR considers *all* of them.

I'm all up for improving APIs but introducing so many changes at the same time is a sure recipe for regressions.

bnjbvr:
@stefanceriu Thanks for trying it out. I'm not sure how I'm supposed to guess what you need based on skimming the list of "problems", since they don't mention any particular APIs or precise issues, except for the voice message item? (The first bullet item should've been addressed, considering I've added back event_id/transaction_id in the EventTimelineItem.)

bindings/matrix-sdk-ffi/src/timeline/mod.rs
```diff
@@ -1032,27 +1018,31 @@ impl From<SdkShieldState> for ShieldState {
...
-    is_local: bool,
+    /// Indicates that an event is remote.
     is_remote: bool,
-    event_or_transaction_id: EventOrTransactionId,
+    /// An event id, for events that have been remote-echoed.
+    event_id: Option<String>,
+    /// A transaction id, for events waiting in the send queue.
+    ///
+    /// Note: not set for remote-echoes.
+    transaction_id: Option<String>,
```
jmartinesp, line 1028: This sounds a lot like `EventOrTransactionId` to me.

bnjbvr: We're already losing much static typing safety by using Strings everywhere, maybe we can afford two types to help distinguish one from the other 🥲 Also, one can't use an `EventOrTransactionId` in any API, it's either an event id OR a transaction id, so using any of those APIs would require passing the event|transaction id and pray your best luck/favorite god that it doesn't fail because it wasn't the expected type.

poljar:
> @stefanceriu Thanks for trying it out. I'm not sure how I'm supposed to guess what you need based on skimming the list of "problems", since they don't mention any particular APIs or precise issues, except for the voice message item? (The first bullet item should've been addressed, considering I've added back event_id/transaction_id in the EventTimelineItem.)

Can we perhaps split this PR out into smaller chunks, that would perhaps make it easier to see what introduces certain regressions.

bnjbvr:
We've discussed about it offline, and we think using the `event or transaction id` instead of the `TransactionUniqueId` in more places is the right thing to do:

- transaction unique IDs aren't stable across timeline resets (when receiving a gappy timeline from the server)
- transaction unique IDs aren't stable across timeline instances (e.g. when there's a permalink-focused and a live timelines living at the same time)

So the plan is:

- to use transaction-or-event-id in more places (i.e. not use `EventTimelineItem` or `TransactionUniqueId` in APIs for edit/react/redact)
- get rid of the duplicated APIs at the SDK level
- ???
- fun and profits

bindings/matrix-sdk-ffi/src/timeline/mod.rs
```diff
@@ -610,10 +627,11 @@ impl Timeline {
         event_or_transaction_id: EventOrTransactionId,
         reason: Option<String>,
     ) -> Result<(), ClientError> {
-        self.inner
-            .redact_by_id(&(event_or_transaction_id.try_into()?), reason.as_deref())
-            .await
-            .map_err(Into::into)
+        if !self.inner.redact(&(event_or_transaction_id.try_into()?), reason.as_deref()).await? {
+            // TODO make it a hard error instead
```
stefanceriu, line 631: Can we please change the underlying API to return `Result<bool, Error>` at this point please? We already have a `RedactError`, it shouldn't be too much of a stretch to convert into it from the underlying handle abortion boolean.

stefanceriu (changes requested):
With the other 2 PRs now in place this is looking very nice! 👏  (bar those inline comment of mine).

bindings/matrix-sdk-ffi/src/timeline/mod.rs
```diff
@@ -494,11 +494,28 @@ impl Timeline {
...
-            .map_err(Into::into)
+    ) -> Result<(), ClientError> {
+        let edited = self
+            .inner
+            .edit(&event_or_transaction_id.clone().try_into()?, new_content.clone().try_into()?)
+            .await?;
+
+        if !edited {
+            // If we couldn't edit, assume it was an (remote) event that wasn't in the
+            // timeline, and try to edit it via the room itself.
```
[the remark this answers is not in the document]
bnjbvr: Alright, this convinced me that your next comment about not returning bool anymore was the right thing :) Thanks for pushing it.

stefanceriu (approved):
Love it! 👏

PR #4102, opened 2024-10-09
feat(sdk): Dropping an `UpdatesSubscriber` release its reader token for the garbage collector

Hywan:
The event cache stores its events in a linked chunk. The linked chunk
supports updates (`ObservableUpdates`) via `LinkedChunk::updates()`.
This `ObservableUpdates` receives all updates that are happening inside
the `LinkedChunk`. An `ObservableUpdates` wraps `UpdatesInner`, which
is the real logic to handle multiple update readers. Each reader has a
unique `ReaderToken`. `UpdatesInner` has a garbage collector that drops
all updates that are read by all readers. And here comes the problem.

A category of readers are `UpdatesSubscriber`, returned by
`ObservableUpdates::subscribe()`. When an `UpdatesSubscriber` is
dropped, its reader token was still alive, thus preventing the garbage
collector to clear all its pending updates: they were kept in memory
for the eternity.

This patch implements `Drop` for `UpdatesSubscriber` to correctly remove
its `ReaderToken` from `UpdatesInner`. This patch also adds a test that
runs multiple subscribers, and when one is dropped, its pending updates
are collected by the garbage collector.

---

### DIFF
--- a/crates/matrix-sdk/src/event_cache/linked_chunk/updates.rs
+++ b/crates/matrix-sdk/src/event_cache/linked_chunk/updates.rs
@@ -495,7 +495,7 @@ mod tests {
         // | d | e | f | g | h | i |
         // +---+---+---+---+---+---+
         //
-        // “main” will have its index updated from 3 to 0.
+        // “main” will have its index updated from 0 to 3.
         // “other” will have its index updated from 6 to 3.
         {
             let updates = linked_chunk.updates().unwrap();
--- a/crates/matrix-sdk/src/event_cache/linked_chunk/updates.rs
+++ b/crates/matrix-sdk/src/event_cache/linked_chunk/updates.rs
@@ -322,6 +322,22 @@ where
     }
 }
 
+impl<Item, Gap> Drop for UpdatesSubscriber<Item, Gap> {
+    fn drop(&mut self) {
+        // Remove `Self::token` from `UpdatesInner::last_index_per_reader`.
+        // This is important so that the garbage collector can do its jobs correctly
+        // without a dead dangling reader token.
+        if let Some(updates) = self.updates.upgrade() {
+            let mut updates = updates.write().unwrap();
+
+            // Remove the reader token from `UpdatesInner`.
+            // It's safe to ignore the result of `remove` here: `None` means the token was
+            // already removed (note: it should be unreachable).
+            let _ = updates.last_index_per_reader.remove(&self.token);
+        }
+    }
+}
+
 #[cfg(test)]
 mod tests {
     use std::{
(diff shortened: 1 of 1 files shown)


crates/matrix-sdk/src/event_cache/linked_chunk/updates.rs
```diff
@@ -646,4 +662,161 @@ mod tests {
...
+        let counter_waker1 = Arc::new(CounterWaker { number_of_wakeup: Mutex::new(0) });
+        let counter_waker2 = Arc::new(CounterWaker { number_of_wakeup: Mutex::new(0) });
+
+        let waker1 = counter_waker1.clone().into();
+        let waker2 = counter_waker2.clone().into();
+
+        let mut context1 = Context::from_waker(&waker1);
+        let mut context2 = Context::from_waker(&waker2);
+
+        let mut linked_chunk = LinkedChunk::<3, char, ()>::new_with_update_history();
```
bnjbvr, line 679: <3

crates/matrix-sdk/src/event_cache/linked_chunk/updates.rs
```diff
@@ -646,4 +662,161 @@ mod tests {
...
+                    );
+                }
+            );
+            assert_matches!(updates_subscriber2.as_mut().poll_next(&mut context2), Poll::Pending);
+
+            // Let's generate two other updates.
+            linked_chunk.push_items_back(['b']);
+            linked_chunk.push_items_back(['c']);
+
+            // The waker must have been called only once for the two updates.
```
bnjbvr, line 728: I think I understand why, but can you elaborate in the comment? (It's because we only add a waker whenever we explicitly start polling, which we did once before for each stream.)

crates/matrix-sdk/src/event_cache/linked_chunk/updates.rs
```diff
@@ -646,4 +662,161 @@ mod tests {
...
+                            PushItems { at: Position(ChunkIdentifier(0), 2), items: vec!['c'] },
+                        ]
+                    );
+                }
+            );
+            assert_matches!(updates_subscriber1.as_mut().poll_next(&mut context1), Poll::Pending);
+
+            // For the sake of this test, we also need to advance the main reader token.
+            let _ = linked_chunk.updates().unwrap().take();
+            let _ = linked_chunk.updates().unwrap().take();
```
bnjbvr, line 749: Uh? Why do we need to do it twice?

bnjbvr (approved):
LGTM, thanks. Need a few clarifications in the test IMO, but good to merge after comments have been addressed :+1:

crates/matrix-sdk/src/event_cache/linked_chunk/updates.rs
Hywan: A waker is consumed after it's been called. A waker is registered when a `Future` or a `Stream` is polled. I'm updating the comment.

crates/matrix-sdk/src/event_cache/linked_chunk/updates.rs
Hywan: That's something I want to change. The garbage collector is run _before_ `take_with_token` is called (which is called by `take`). I think it's better to run the garbage collector _after_ `take_with_token` is called: “I read new items, I clean read items” vs. “I clean previously read items, I read new ones”. Right now, we are in the second pattern, but I believe the first is preferable. I'll address that in the next pull request.

PR #4107, opened 2024-10-10
ci: detect unused dependencies

bnjbvr:
I've noticed there were a few unused dependencies; this removes them, and introduces a new check that will fail if there are any unused dependencies detected with `cargo-machete`. Also marks one false positive that the tool can't detect.


poljar (approved):
Nice, though not sure if the author of this tool can be trusted :eyes:.

PR #4110, opened 2024-10-10
refactor(ffi): remove two fields in `EventTimelineItem`

bnjbvr:
- `is_local` is the opposite of `is_remote`. I hope all foreign languages have a `!` operator :clown_face: 
- the two lazy providers can be unified into a single one, avoiding a bit of boilerplate.

Split from #4100.


jmartinesp:
> is_local is the opposite of is_remote. I hope all foreign languages have a ! operator 🤡

This will do!

```kotlin
val isLocal: Boolean = is_remote.toInt() - 1 == 0
```

jmartinesp (approved):
LGTM, thanks!

I liked the previous  `*Provider` way of splitting responsibilities better, but we'll probably add wrappers for them once they're in Kotlin/Swift anyway, so we can split them there if we want to, so it's no big deal.

PR #4111, opened 2024-10-10
feat(timeline): introduce the `TimelineUniqueId`

bnjbvr:
See commit by commit description.

Extracted from #4100.


stefanceriu (changes requested):
This looks good to me. There's a bunch of changes we will need to make on the iOS side but I think it's about time, they're welcome.

bnjbvr:
See https://github.com/matrix-org/matrix-rust-sdk/pull/4100#issuecomment-2413529652. From this PR we can keep the first commit (don't try to reuse unique IDs) + the opaque type, and we'll get rid of the third one likely. It might be better to *not* use the transaction unique id in the react function first.

bnjbvr:
Ready for another round of review, after/on-top-of https://github.com/matrix-org/matrix-rust-sdk/pull/4127 (but since you're the reviewer, you choose to decide how you review all this :)). Thanks!

stefanceriu (approved):
Just tried it out in [on this branch and on top of 4127](https://github.com/element-hq/element-x-ios/tree/stefan/rust4127) and everything looks good to me. Nicely done 👍

PR #4115, opened 2024-10-10
refactor(event cache): simplify the locking situation

### LINKED ISSUE #3502: event cache: rethink locking of internal room data
Followup to #3501, where two mutexes were covering each other's critical regions in different places.

As the event cache grows, we need to have more precise rules for using locking:

- either all the locks must be as short-lived as possible, with the introduction of potential races if things happen in parallel (the fix above may introduce a benign race where multiple pagination wait for the initial token multiple times)
- or all the data should live under the same lock umbrella

I think the latter is better. We'll need to improve the situation before moving on and adding more data in a room event cache.

cc @Hywan

bnjbvr:
This simplifies the locking situation, in particular it would've made the bug fixed by https://github.com/matrix-org/matrix-rust-sdk/pull/3501 impossible to happen in the first place (or at least, way more obvious).

Fixes #3502. Thanks @Hywan for the technical discussion we had about it.

### DIFF
--- a/crates/matrix-sdk/src/event_cache/paginator.rs
+++ b/crates/matrix-sdk/src/event_cache/paginator.rs
@@ -90,6 +90,15 @@ impl From<Option<String>> for PaginationToken {
     }
 }
 
+/// Paginations tokens used for backward and forward pagination.
+#[derive(Debug)]
+struct PaginationTokens {
+    /// Pagination token used for backward pagination.
+    previous: PaginationToken,
+    /// Pagination token used for forward pagination.
+    next: PaginationToken,
+}
+
 /// A stateful object to reach to an event, and then paginate backward and
 /// forward from it.
 ///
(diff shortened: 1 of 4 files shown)


crates/matrix-sdk/src/event_cache/paginator.rs
```diff
@@ -296,14 +302,17 @@ impl<PR: PaginableRoom> Paginator<PR> {
...
-        *self.prev_batch_token.lock().unwrap() = match response.prev_batch_token {
-            Some(token) => PaginationToken::HasMore(token),
-            None => PaginationToken::HitEnd,
-        };
-        *self.next_batch_token.lock().unwrap() = match response.next_batch_token {
-            Some(token) => PaginationToken::HasMore(token),
-            None => PaginationToken::HitEnd,
-        };
+        {
+            let mut tokens = self.tokens.lock().unwrap();
```
poljar, line 306: I have a deja-vu, didn't I suggest a single lock for this a month or two ago? Anyways, nice that we remove some locks.

bnjbvr: Yes, that rings a bell, sorry I hadn't listened more :sweat_smile:

PR #4116, opened 2024-10-10
crypto: expose new method `OlmMachine::try_decrypt_room_event`

richvdh:
... which returns an enum which encodes decryption failure information, rather than returning `MegolmError`

To make this work, we also extend the existing `UnableToDecryptInfo` to include a reason code for the failure.

Part of https://github.com/matrix-org/matrix-rust-sdk/issues/4048

### DIFF
--- a/crates/matrix-sdk-common/src/deserialized_responses.rs
+++ b/crates/matrix-sdk-common/src/deserialized_responses.rs
@@ -705,6 +705,8 @@ impl From<SyncTimelineEventDeserializationHelperV0> for SyncTimelineEvent {
 
 #[cfg(test)]
 mod tests {
+    use std::collections::BTreeMap;
+
     use assert_matches::assert_matches;
     use ruma::{
         event_id,
@@ -717,7 +719,8 @@ mod tests {
 
     use super::{
         AlgorithmInfo, DecryptedRoomEvent, EncryptionInfo, SyncTimelineEvent, TimelineEvent,
-        TimelineEventKind, VerificationState,
+        TimelineEventKind, UnableToDecryptInfo, UnsignedDecryptionResult, UnsignedEventLocation,
+        VerificationState,
     };
     use crate::deserialized_responses::{DeviceLinkProblem, VerificationLevel};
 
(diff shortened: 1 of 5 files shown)


crates/matrix-sdk-common/src/deserialized_responses.rs
```diff
@@ -628,6 +628,59 @@ pub struct UnableToDecryptInfo {
...
+
+    /// Decryption failed because, while we have the megolm session that was
+    /// used to encrypt the message, it is ratcheted too far forward.
+    UnknownMegolmMessageIndex,
+
+    /// We found the Megolm session, but were unable to decrypt the event using
+    /// that session for some reason (e.g. incorrect MAC).
+    ///
+    /// This represents all `vodozemac::megolm::DecryptionError`s, with the
+    /// exception of `UnknownMessageIndex`, which is represented as
```
andybalaam, line 668: I found this difficult to parse because of the special meaning of "exception" in my head. Would "with the exception of" -> "except" be better?

andybalaam (approved):
Looks good to me.

crates/matrix-sdk-common/src/deserialized_responses.rs
```diff
@@ -628,6 +628,59 @@ pub struct UnableToDecryptInfo {
...
     pub session_id: Option<String>,
+
+    /// Reason code for the decryption failure
+    #[serde(default = "unknown_utd_reason")]
+    pub reason: UnableToDecryptReason,
+}
+
+fn unknown_utd_reason() -> UnableToDecryptReason {
+    UnableToDecryptReason::Unknown
+}
```
bnjbvr, line 639: For what it's worth, you can instead `derive(Default)` to `UnableToDecryptReason`, and then mark one variant as the default:

```rust
#[default]
#[doc(hidden)]
Unknown,
```

bnjbvr (approved):
LGTM, thanks for breaking it into simple changes :)

crates/matrix-sdk-common/src/deserialized_responses.rs
```diff
@@ -628,6 +628,59 @@ pub struct UnableToDecryptInfo {
...
+    /// We found the Megolm session, but were unable to decrypt the event using
+    /// that session for some reason (e.g. incorrect MAC).
+    ///
+    /// This represents all `vodozemac::megolm::DecryptionError`s, with the
+    /// exception of `UnknownMessageIndex`, which is represented as
+    /// `UnknownMegolmMessageIndex`.
+    MegolmDecryptionFailure,
+
+    /// The event could not be deserialized after decryption.
+    PayloadDeserializationFailure,
```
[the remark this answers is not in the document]
richvdh: I don't really agree. If I compare with something like `MissingMegolmSession`: a missing megolm session is a sensible reason for being unable to decrypt an event. So is a "megolm decryption failure". But just `MegolmDecryption` on its own doesn't really make sense to me.

I'm prepared to believe there could be better names for these things (`UnableToDecryptMegolmMessage`? `VodozemacDecryptionError`?), but I don't think `MegolmDecryption` would be a good one.

PR #4118, opened 2024-10-10
chore(ffi): Merge export and export_async attribute macros

jplatte:
I decided to make this forwards-compatible with `async fn` in callback interfaces / traits, bc. why not?


bnjbvr (approved):
Good idea, thanks!

PR #4119, opened 2024-10-10
chore: Move lint configuration out of .cargo/config.toml

jplatte:
This allows removing a lot of hacks to avoid spurious rebuilds.


poljar (approved):
Nice, thanks.

PR #4126, opened 2024-10-15
Timeline: Extend UtdCause with new reason codes

richvdh:
This PR is the final part of the epic of making unable-to-decrypt information available in the timeline. As of #4070, the UTD info is now available in the `[Sync]TimelineEvent` created by the network code: now, we just need to copy that information into the items that are added to the timeline itself.

To do this, we add a new variant to `matrix_sdk_ui::event_handler::TimelineEventKind` (not to be confused with `matrix_sdk_common::deserialized_responses::TimelineEventKind` which was recently added in #4082) to hold the relevant data, and then add a bunch of new reason codes to the `UtdCause` which is derived from the decryption failure.

Part of https://github.com/matrix-org/matrix-rust-sdk/issues/4048.
~~Based on https://github.com/matrix-org/matrix-rust-sdk/pull/4070.~~

### DIFF
--- a/crates/matrix-sdk-crypto/src/types/events/utd_cause.rs
+++ b/crates/matrix-sdk-crypto/src/types/events/utd_cause.rs
@@ -117,7 +117,7 @@ mod tests {
     use crate::types::events::UtdCause;
 
     #[test]
-    fn a_missing_raw_event_means_we_guess_unknown() {
+    fn test_a_missing_raw_event_means_we_guess_unknown() {
         // When we don't provide any JSON to check for membership, then we guess the UTD
         // is unknown.
         assert_eq!(
@@ -133,7 +133,7 @@ mod tests {
     }
 
     #[test]
-    fn if_there_is_no_membership_info_we_guess_unknown() {
+    fn test_if_there_is_no_membership_info_we_guess_unknown() {
         // If our JSON contains no membership info, then we guess the UTD is unknown.
         assert_eq!(
             UtdCause::determine(
@@ -148,7 +148,7 @@ mod tests {
     }
 
     #[test]
-    fn if_membership_info_cant_be_parsed_we_guess_unknown() {
+    fn test_if_membership_info_cant_be_parsed_we_guess_unknown() {
         // If our JSON contains a membership property but not the JSON we expected, then
         // we guess the UTD is unknown.
         assert_eq!(
@@ -164,7 +164,7 @@ mod tests {
     }
 
     #[test]
-    fn if_membership_is_invite_we_guess_unknown() {
+    fn test_if_membership_is_invite_we_guess_unknown() {
         // If membership=invite then we expected to be sent the keys so the cause of the
         // UTD is unknown.
         assert_eq!(
@@ -180,7 +180,7 @@ mod tests {
     }
 
     #[test]
-    fn if_membership_is_join_we_guess_unknown() {
+    fn test_if_membership_is_join_we_guess_unknown() {
         // If membership=join then we expected to be sent the keys so the cause of the
         // UTD is unknown.
         assert_eq!(
@@ -196,7 +196,7 @@ mod tests {
     }
 
     #[test]
-    fn if_membership_is_leave_we_guess_membership() {
+    fn test_if_membership_is_leave_we_guess_membership() {
         // If membership=leave then we have an explanation for why we can't decrypt,
         // until we have MSC3061.
         assert_eq!(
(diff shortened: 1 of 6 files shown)


crates/matrix-sdk-crypto/src/types/events/utd_cause.rs
```diff
@@ -180,6 +212,24 @@ mod tests {
         );
     }
 
+    #[test]
+    fn if_reason_is_not_missing_key_we_guess_unknown_even_if_membership_is_leave_we_guess_membership(
```
andybalaam, line 216: We have a convention (which I personally don't like, but do comply with) of test names starting with `test_`

andybalaam (approved):
One suggestion, but good to go, thanks!

crates/matrix-sdk-crypto/src/types/events/utd_cause.rs
richvdh: For me, consistency within a file is more important than the convention for new tests

crates/matrix-sdk-crypto/src/types/events/utd_cause.rs
```diff
@@ -24,16 +27,27 @@ pub enum UtdCause {
...
+    /// The message was sent by a user identity we have not verified, but the
+    /// user was previously verified.
+    VerificationViolation = 2,
+
+    /// The [`crate::TrustRequirement`] requires that the sending device be
+    /// signed by its owner, and it was not.
+    UnsignedDevice = 3,
+
+    /// The [`crate::TrustRequirement`] requires that the sending device be
+    /// signed by its owner, and we were unable to securely  find the device.
```
richvdh, line 43:
```suggestion
    /// signed by its owner, and we were unable to securely find the device.
```

crates/matrix-sdk-crypto/src/types/events/utd_cause.rs
```diff
@@ -24,16 +27,27 @@ pub enum UtdCause {
...
+    ///
+    /// This could be because the device has since been deleted, because we
+    /// haven't yet downloaded it from the server, or because the session
+    /// data was obtained from an insecure source (imported from a file,
+    /// obtained from a legacy (asymmetric) backup, unsafe key forward, etc.
+    /// )
```
richvdh, line 49:
```suggestion
    /// obtained from a legacy (asymmetric) backup, unsafe key forward, etc.)
```

crates/matrix-sdk-crypto/src/types/events/utd_cause.rs
bnjbvr: Then please prefix all the tests names with `test_`. This makes it easier to distinguish test functions from non-test functions in LSP searches.

crates/matrix-sdk-ui/src/timeline/event_handler.rs
```diff
@@ -182,6 +195,22 @@ impl TimelineEventKind {
...
+                        // `SyncTimelineEvent` containing an `m.room.encrypted` event
+                        // without decrypting it. Possibly this means that encryption has not been
+                        // configured.
+                        // We treat it the same as any other message-like event.
+                        Self::Message {
+                            content: AnyMessageLikeEventContent::RoomEncrypted(content),
+                            relations: ev.relations(),
+                        }
+                    }
+                }
```
bnjbvr, line 213: Why do we handle those two cases separately? The comment doesn't really explain it to me. I'm not quite sure about why we need a new `TimelineEventKind` variant just for this…

richvdh: > I'm not quite sure about why we need a new `TimelineEventKind` variant just for this…

Well, the objective here is to convey the `unable_to_decrypt_info` that is in the `SyncTimelineEvent` or `TimelineEvent` to `TimelineEventHandler::handle_event`. It seems wrong to just shove another field onto `TimelineEventKind::Message`, hence adding a new variant.

> Why do we handle those two cases separately? The comment doesn't really explain it to me

If the lower layers have done their jobs correctly, then encrypted events will appear either as:
 * a `[Sync]TimelineEvent` with `kind: Decrypted`, in which case `ev.original_content()` here will be the decrypted event, and we hit the `Some(content) case on line 214
 * a `[Sync]TimelineEvent` with `kind: UnableToDecrypt`, in which case `ev.original_content()` here will be the *encrypted* event and `unable_to_decrypt_info` will be populated, and we hit the case at line 200.

However, the object model still allows for a `[Sync]TimelineEvent` with `kind: PlainText` whose content is actually a `AnyMessageLikeEventContent::RoomEncrypted`, and the second case (lines 203-211) is to deal with that situation. One way that I think it will happen is if the OlmMachine is not correctly set up for some reason (`BaseClient::olm_machine` is an `Option` which has to be loaded somewhere.)  And that's what the comment is trying to say.

PR #4128, opened 2024-10-15
feat(knocking): process knocked rooms separately during sync

jmartinesp:
## Changes

This PR contains 2 commits:

1. The first one processes the membership state events in `process_sliding_sync_room_membership`, deserializes them and decides whether the resulting room should be a knock or invite one default to invite if not sure. Then it propagates the changes up the flow.
2. Since we already have `handle_stripped_state` to parse the contents of the `invite_state` stripped state events it seemed better to reuse this and deserialize the events only once, then mark the room as invited or knocked in this method. Ideally this should be part of the inner `fn BaseRoomInfo::handle_stripped_state_event`, but that doesn't have access to the `Room::state` property.

Both solutions work AFAICT, the 1st one centralises the membership processing in a single point but it's more inefficient, the 2nd one does only one round of deserialization but hides the membership change a bit lower in the sliding sync processing flow. Let me know if you have any preferences and I'll use one of those versions, or propose a better alternative.

Existing tests have been extended to check the invite vs knock room state processing.

Signed-off-by:

### DIFF
--- a/crates/matrix-sdk-base/src/client.rs
+++ b/crates/matrix-sdk-base/src/client.rs
@@ -523,11 +523,22 @@ impl BaseClient {
         Ok(timeline)
     }
 
+    /// Handles the stripped state events in `invite_state`, modifying the
+    /// room's info and posting notifications as needed.
+    ///
+    /// * `room` - The [`Room`] to modify.
+    /// * `events` - The contents of `invite_state` in the form of list of pairs
+    ///   of raw stripped state events with their deserialized counterpart.
+    /// * `push_rules` - The push rules for this room.
+    /// * `room_info` - The current room's info.
+    /// * `changes` - The accumulated list of changes to apply once the
+    ///   processing is finished.
+    /// * `notifications` - Notifications to post for the current room.
     #[instrument(skip_all, fields(room_id = ?room_info.room_id))]
     pub(crate) async fn handle_invited_state(
         &self,
         room: &Room,
-        events: &[Raw<AnyStrippedStateEvent>],
+        events: &[(Raw<AnyStrippedStateEvent>, AnyStrippedStateEvent)],
         push_rules: &Ruleset,
         room_info: &mut RoomInfo,
         changes: &mut StateChanges,
@@ -535,22 +546,12 @@ impl BaseClient {
     ) -> Result<()> {
         let mut state_events = BTreeMap::new();
 
-        for raw_event in events {
-            match raw_event.deserialize() {
-                Ok(e) => {
-                    room_info.handle_stripped_state_event(&e);
-                    state_events
-                        .entry(e.event_type())
-                        .or_insert_with(BTreeMap::new)
-                        .insert(e.state_key().to_owned(), raw_event.clone());
-                }
-                Err(err) => {
-                    warn!(
-                        room_id = ?room_info.room_id,
-                        "Couldn't deserialize stripped state event: {err:?}",
-                    );
-                }
-            }
+        for (raw_event, event) in events {
+            room_info.handle_stripped_state_event(event);
+            state_events
+                .entry(event.event_type())
+                .or_insert_with(BTreeMap::new)
+                .insert(event.state_key().to_owned(), raw_event.clone());
         }
 
         changes.stripped_state.insert(room_info.room_id().to_owned(), state_events.clone());
@@ -1121,13 +1122,16 @@ impl BaseClient {
                 self.room_info_notable_update_sender.clone(),
             );
 
+            let invite_state =
+                Self::deserialize_stripped_state_events(&new_info.invite_state.events);
+
             let mut room_info = room.clone_info();
             room_info.mark_as_invited();
             room_info.mark_state_fully_synced();
 
             self.handle_invited_state(
                 &room,
-                &new_info.invite_state.events,
+                &invite_state,
                 &push_rules,
                 &mut room_info,
                 &mut changes,
@@ -1138,6 +1142,34 @@ impl BaseClient {
             changes.add_room(room_info);
 
             new_rooms.invite.insert(room_id, new_info);
+        }
+
+        for (room_id, new_info) in response.rooms.knock {
+            let room = self.store.get_or_create_room(
+                &room_id,
+                RoomState::Knocked,
+                self.room_info_notable_update_sender.clone(),
+            );
+
+            let knock_state = Self::deserialize_stripped_state_events(&new_info.knock_state.events);
+
+            let mut room_info = room.clone_info();
+            room_info.mark_as_knocked();
+            room_info.mark_state_fully_synced();
+
+            self.handle_invited_state(
+                &room,
+                &knock_state,
+                &push_rules,
+                &mut room_info,
+                &mut changes,
+                &mut notifications,
+            )
+            .await?;
+
+            changes.add_room(room_info);
+
+            new_rooms.knocked.insert(room_id, new_info);
         }
 
         account_data_processor.apply(&mut changes, &self.store).await;
@@ -1576,6 +1608,21 @@ impl BaseClient {
                 Ok(event) => Some((raw_event.clone(), event)),
                 Err(e) => {
                     warn!("Couldn't deserialize state event: {e}");
+                    None
+                }
+            })
+            .collect()
+    }
+
+    pub(crate) fn deserialize_stripped_state_events(
+        raw_events: &[Raw<AnyStrippedStateEvent>],
+    ) -> Vec<(Raw<AnyStrippedStateEvent>, AnyStrippedStateEvent)> {
+        raw_events
+            .iter()
+            .filter_map(|raw_event| match raw_event.deserialize() {
+                Ok(event) => Some((raw_event.clone(), event)),
+                Err(e) => {
+                    warn!("Couldn't deserialize stripped state event: {e}");
                     None
                 }
             })
--- a/crates/matrix-sdk-base/src/sliding_sync/mod.rs
+++ b/crates/matrix-sdk-base/src/sliding_sync/mod.rs
@@ -27,10 +27,13 @@ use ruma::api::client::sync::sync_events::v5;
 #[cfg(feature = "e2e-encryption")]
 use ruma::events::AnyToDeviceEvent;
 use ruma::{
-    api::client::sync::sync_events::v3::{self, InvitedRoom},
-    events::{AnyRoomAccountDataEvent, AnySyncStateEvent},
+    api::client::sync::sync_events::v3::{self, InvitedRoom, KnockedRoom},
+    events::{
+        room::member::MembershipState, AnyRoomAccountDataEvent, AnyStrippedStateEvent,
+        AnySyncStateEvent,
+    },
     serde::Raw,
-    JsOption, OwnedRoomId, RoomId, UInt,
+    JsOption, OwnedRoomId, RoomId, UInt, UserId,
 };
 use tracing::{debug, error, instrument, trace, warn};
 
@@ -47,6 +50,7 @@ use crate::{
         normal::{RoomHero, RoomInfoNotableUpdateReasons},
         RoomState,
     },
+    ruma::assign,
     store::{ambiguity_map::AmbiguityCache, StateChanges, Store},
     sync::{JoinedRoomUpdate, LeftRoomUpdate, Notification, RoomUpdates, SyncResponse},
     Room, RoomInfo,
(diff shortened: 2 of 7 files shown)


jmartinesp:
I just rebased the changes to avoid a `use ...` conflict, there aren't any other changes.

crates/matrix-sdk-base/src/sliding_sync/mod.rs
```diff
@@ -555,20 +576,35 @@ impl BaseClient {
...
-            // invite_state, but we default to considering this room invited
-            // simply because invite_state exists. This is needed in the normal
-            // case, because the sliding sync server tries to send minimal state,
-            // meaning that we normally actually just receive {"type": "m.room.member"} with
-            // no content at all.
-            room_info.mark_as_invited();
+            // We need to find the membership event since it could be for either an invited
+            // or knocked room
+            let membership_event_content = invite_state.iter().find_map(|raw| {
+                raw.deserialize().ok().iter().find_map(|e| {
```
bnjbvr: So I'm not sure we're looking in the right place here: the `required_state` events have already been deserialized and passed as a function parameter called `state_events`. Might be worth checking against a real-world Synapse, but I'd expect the membership event to be there instead. If that's true, then we don't need the deserialization step here, and this solution feels better than the second one IMO (with the previous sync protocol, knocked rooms would appear in the knocked subset of the response, which is nice and simple — I'd rather not change the room state based on extra other information).

bnjbvr (commented):
Looks promising. I think I prefer the way it's handled in the first commit, especially since there might be a nice simplification there too.

crates/matrix-sdk-base/src/sliding_sync/mod.rs
jmartinesp: Invite and Knock rooms use `invite_state` instead of `required_state`, I suspect their required_state would be empty, but I can double check.

crates/matrix-sdk-base/src/sliding_sync/mod.rs
```diff
@@ -555,20 +576,35 @@ impl BaseClient {
...
+                    if let AnyStrippedStateEvent::RoomMember(membership_event) = e {
+                        if membership_event.sender == user_id {
```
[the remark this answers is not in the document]
jmartinesp: Yes, I think this is an auto-completion typo 😬 . Nice catch!

crates/matrix-sdk-base/src/sliding_sync/mod.rs
jmartinesp: Yes, there's no `required_state` for knocked rooms:

```json
"!luXevcpKTZauFIpPnL:matrix.org": {
  "bump_stamp": 21538303,
  "notification_count": 0,
  "highlight_count": 0,
  "initial": true,
  "invite_state": [
    {
      "content": {
        "creator": "@jorgem_test:matrix.org",
        "room_version": "7"
      },
      "sender": "@jorgem_test:matrix.org",
      "state_key": "",
      "type": "m.room.create"
    },
    {
      "content": {
        "join_rule": "knock"
      },
      "sender": "@jorgem_test:matrix.org",
      "state_key": "",
      "type": "m.room.join_rules"
    },
    {
      "content": {
        "name": "Knocking room test"
      },
      "sender": "@jorgem_test:matrix.org",
      "state_key": "",
      "type": "m.room.name"
    },
    {
      "content": {
        "algorithm": "m.megolm.v1.aes-sha2"
      },
      "sender": "@jorgem_test:matrix.org",
      "state_key": "",
      "type": "m.room.encryption"
    },
    {
      "type": "m.room.member",
      "state_key": "@jorgem:element.io",
      "content": {
        "avatar_url": "mxc://element.io/75d9a551895233d284aff390d2086814a2f7e875",
        "displayname": "Jorge",
        "membership": "knock"
      },
      "sender": "@jorgem:element.io"
    }
  ]
}
```

jmartinesp: Another option would be to handle the stripped state events in `invite_state` as part of the membership in this same method, but I think that's not possible because we need the `push_rules` too which are processed after this.

jmartinesp: We could also deserialize the stripped state events in `process_sliding_sync_room` and pass it down both functions, this one and `handle_invited_state`, so they'd be deserialized just once.

crates/matrix-sdk-base/src/client.rs
```diff
@@ -1140,6 +1140,32 @@ impl BaseClient {
...
 
+        for (room_id, new_info) in response.rooms.knock {
+            let room = self.store.get_or_create_room(
+                &room_id,
+                RoomState::Knocked,
+                self.room_info_notable_update_sender.clone(),
+            );
+
+            let mut room_info = room.clone_info();
+            room_info.mark_as_invited();
```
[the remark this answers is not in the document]
jmartinesp: https://github.com/matrix-org/matrix-rust-sdk/pull/4128/commits/01b3bc953c832a3c3698045866a98c942138730f

crates/matrix-sdk-base/src/sliding_sync/mod.rs
```diff
@@ -546,7 +562,12 @@ impl BaseClient {
         store: &Store,
         room_id: &RoomId,
         room_info_notable_updates: &mut BTreeMap<OwnedRoomId, RoomInfoNotableUpdateReasons>,
-    ) -> (Room, RoomInfo, Option<InvitedRoom>) {
+    ) -> (Room, RoomInfo, Option<InvitedRoom>, Option<KnockedRoom>) {
+        let user_id = self
+            .session_meta()
+            .expect("Sliding sync shouldn't run without an authenticated user.")
+            .user_id
+            .to_owned();
```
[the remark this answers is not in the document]
jmartinesp: https://github.com/matrix-org/matrix-rust-sdk/pull/4128/commits/01b3bc953c832a3c3698045866a98c942138730f should take care of this

crates/matrix-sdk-base/src/sliding_sync/mod.rs
jmartinesp: I tried this approach ⬆️  in https://github.com/matrix-org/matrix-rust-sdk/pull/4128/commits/0b1ef5b3fe107e219fb93278b973affdb94b949c

crates/matrix-sdk-base/src/sliding_sync/mod.rs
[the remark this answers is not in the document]
jmartinesp: https://github.com/matrix-org/matrix-rust-sdk/pull/4128/commits/01b3bc953c832a3c3698045866a98c942138730f should take care of the typo, https://github.com/matrix-org/matrix-rust-sdk/pull/4128/commits/50f7fb06a41832f3e813b6709701b1f61f629b55 adds more tests for corner cases with wrong membership events.

crates/matrix-sdk-base/src/client.rs
```diff
@@ -527,30 +527,20 @@ impl BaseClient {
     pub(crate) async fn handle_invited_state(
         &self,
         room: &Room,
-        events: &[Raw<AnyStrippedStateEvent>],
+        events: &[(Raw<AnyStrippedStateEvent>, AnyStrippedStateEvent)],
```
andybalaam, line 530: Some documentation of the meaning of this parameter would be appreciated. It's a bit cryptic at the moment: a slice of pairs of things that look the same at first glance.

crates/matrix-sdk-base/src/sliding_sync/mod.rs
```diff
@@ -1042,6 +1090,76 @@ mod tests {
         assert_eq!(client_room.compute_display_name().await.unwrap().to_string(), "The Name");
     }
 
+    #[async_test]
+    async fn test_knocked_room_membership_event() {
```
andybalaam, line 1094: `test_receiving_a_knocked_room_membership_event_creates_a_knocked_room` ?

crates/matrix-sdk-base/src/sliding_sync/mod.rs
```diff
@@ -1042,6 +1090,76 @@ mod tests {
...
+        let response = response_with_room(room_id, room);
+        client.process_sliding_sync(&response, &(), true).await.expect("Failed to process sync");
+
+        // The room is knocked.
+        let client_room = client.get_room(room_id).expect("No room found");
+        assert_eq!(client_room.state(), RoomState::Knocked);
+    }
+
+    #[async_test]
+    async fn test_knocked_room_membership_event_with_wrong_state_key() {
```
andybalaam, line 1114: `test_receiving_a_knocked_room_membership_event_with_wrong_state_key_creates_an_invited_room` ?

crates/matrix-sdk-base/src/sliding_sync/mod.rs
```diff
@@ -1042,6 +1090,76 @@ mod tests {
...
+        client.process_sliding_sync(&response, &(), true).await.expect("Failed to process sync");
+
+        // The room is invited since the membership event doesn't belong to the current
+        // user.
+        let client_room = client.get_room(room_id).expect("No room found");
+        assert_eq!(client_room.state(), RoomState::Invited);
+    }
+
+    #[async_test]
+    async fn test_unknown_room_membership_event_in_invite_state() {
```
andybalaam, line 1134: `test_receiving_an_unknown_room_membership_event_in_invite_state_creates_an_invited_room` ?

andybalaam (approved):
Some suggestions, all optional. Looks good to me.

crates/matrix-sdk-base/src/client.rs
```diff
@@ -1582,6 +1614,21 @@ impl BaseClient {
...
 
+    pub(crate) fn deserialize_stripped_state_events(
+        raw_events: &[Raw<AnyStrippedStateEvent>],
+    ) -> Vec<(Raw<AnyStrippedStateEvent>, AnyStrippedStateEvent)> {
+        raw_events
+            .iter()
+            .filter_map(|raw_event| match raw_event.deserialize() {
+                Ok(event) => Some((raw_event.clone(), event)),
+                Err(e) => {
+                    warn!("Couldn't deserialize state event: {e}");
```
bnjbvr, line 1625:
```suggestion
                    warn!("Couldn't deserialize stripped state event: {e}");
```

crates/matrix-sdk-base/src/sliding_sync/mod.rs
```diff
@@ -341,14 +356,20 @@ impl BaseClient {
...
         ambiguity_cache: &mut AmbiguityCache,
         with_msc4186: bool,
-    ) -> Result<(RoomInfo, Option<JoinedRoomUpdate>, Option<LeftRoomUpdate>, Option<InvitedRoom>)>
-    {
+    ) -> Result<(
+        RoomInfo,
+        Option<JoinedRoomUpdate>,
+        Option<LeftRoomUpdate>,
+        Option<InvitedRoom>,
+        Option<KnockedRoom>,
```
bnjbvr, line 371: (god what have we done here :fearful:)(no need to change anything right now, unless you see an obvious better way here)

crates/matrix-sdk-base/src/sliding_sync/mod.rs
```diff
@@ -428,7 +457,8 @@ impl BaseClient {
 
         let push_rules = self.get_push_rules(account_data_processor).await?;
 
-        if let Some(invite_state) = &room_data.invite_state {
+        // This will be used for both invited and knocked rooms
```
bnjbvr, line 460: uber-nit:

```suggestion
        // This will be used for both invited and knocked rooms.
```

crates/matrix-sdk-base/src/sliding_sync/mod.rs
```diff
@@ -541,34 +572,49 @@ impl BaseClient {
     /// otherwise. https://github.com/matrix-org/matrix-spec-proposals/blob/kegan/sync-v3/proposals/3575-sync.md#room-list-parameters
     fn process_sliding_sync_room_membership(
         &self,
-        room_data: &http::response::Room,
         state_events: &[AnySyncStateEvent],
+        invite_state: &Option<Vec<(Raw<AnyStrippedStateEvent>, AnyStrippedStateEvent)>>,
```
bnjbvr, line 576: nit: in Rust it's more idiomatic to pass `Option<&T>` rather than `&Option<T>`, by calling `as_ref()` at the caller's call-site.

Alternatively, you could make this non `Option`al and have the caller `.unwrap_or_default()`.

bnjbvr: Could we name this `stripped_state`, though, since it applies to both invites and knocked rooms?

testing/matrix-sdk-test/src/test_json/sync.rs
```diff
@@ -1357,6 +1360,30 @@ pub static MIXED_SYNC: Lazy<JsonValue> = Lazy::new(|| {
...
+                        "type": "m.room.name",
+                        "state_key": "",
+                        "content": {
+                          "name": "My Room Name"
+                        }
+                      },
+                      {
+                        "sender": "@alice:example.com",
+                        "type": "m.room.member",
+                        "state_key": "@bob:example.com",
```
bnjbvr, line 1378: Can someone knock on behalf of somebody else? :thinking: Should the sender/state_key be the same?

bnjbvr (approved):
Thanks, LGTM! only minor nits, so unless something big comes up, feel free to address these (or not) and merge. Good job!

testing/matrix-sdk-test/src/test_json/sync.rs
jmartinesp: Ooops. Copy&paste issue 😅 .

PR #4129, opened 2024-10-15
crypto: Extract a test helper function for simulating verification

andybalaam:
I'd really like some help on how to name the function, the internal variables and what the comments should say, please!

### DIFF
--- a/crates/matrix-sdk-crypto/src/identities/user.rs
+++ b/crates/matrix-sdk-crypto/src/identities/user.rs
@@ -1189,7 +1189,15 @@ where
 #[cfg(any(test, feature = "testing"))]
 #[allow(dead_code)]
 pub(crate) mod testing {
-    use ruma::{api::client::keys::get_keys::v3::Response as KeyQueryResponse, user_id};
+    use matrix_sdk_test::ruma_response_from_json;
+    use ruma::{
+        api::client::keys::{
+            get_keys::v3::Response as KeyQueryResponse,
+            upload_signatures::v3::Request as SignatureUploadRequest,
+        },
+        user_id, UserId,
+    };
+    use serde_json::json;
 
     use super::{OtherUserIdentityData, OwnUserIdentity, OwnUserIdentityData};
     #[cfg(test)]
@@ -1266,6 +1274,117 @@ pub(crate) mod testing {
         OtherUserIdentityData::new(master_key.try_into().unwrap(), self_signing.try_into().unwrap())
             .unwrap()
     }
+
+    /// When we want to test identities that are verified, we need to simulate
+    /// the verification process. This function supports that by simulating
+    /// what happens when a successful verification dance happens and
+    /// providing the /keys/query response we would get when that happened.
+    ///
+    /// signature_upload_request will be the result of calling
+    /// [`super::OtherUserIdentity::verify`].
+    ///
+    /// # Example
+    ///
+    /// ```ignore
+    /// let signature_upload_request = their_identity.verify().await.unwrap();
+    ///
+    /// let msk_json = json!({
+    ///     "their_user_id": {
+    ///         "keys": { "ed25519:blah": "blah" }
+    ///         "signatures": {
+    ///             "their_user_id": { "ed25519:blah": "blah", ... }
+    ///         }
+    ///         "usage": [ "master" ],
+    ///         "user_id": "their_user_id"
+    ///     }
+    /// });
+    ///
+    /// let ssk_json = json!({
+    ///     "their_user_id": {
+    ///         "keys": { "ed25519:blah": "blah" },
+    ///         "signatures": {
+    ///             "their_user_id": { "ed25519:blah": "blah" }
+    ///         },
+    ///         "usage": [ "self_signing" ],
+    ///         "user_id": "their_user_id"
+    ///     }
+    /// })
+    ///
+    /// let response = simulate_key_query_response_for_verification(
+    ///     signature_upload_request,
+    ///     my_identity,
+    ///     my_user_id,
+    ///     their_user_id,
+    ///     msk_json,
+    ///     ssk_json
+    /// ).await;
+    ///
+    /// olm_machine
+    ///     .mark_request_as_sent(
+    ///         &TransactionId::new(),
+    ///         crate::IncomingResponse::KeysQuery(&kq_response),
+    ///     )
+    ///     .await
+    ///     .unwrap();
+    /// ```
+    pub fn simulate_key_query_response_for_verification(
+        signature_upload_request: SignatureUploadRequest,
+        my_identity: OwnUserIdentity,
+        my_user_id: &UserId,
+        their_user_id: &UserId,
+        msk_json: serde_json::Value,
+        ssk_json: serde_json::Value,
+    ) -> KeyQueryResponse {
+        // Find the signed key inside the SignatureUploadRequest
+        let cross_signing_key: CrossSigningKey = serde_json::from_str(
+            signature_upload_request
+                .signed_keys
+                .get(their_user_id)
+                .expect("Signature upload request should contain a key for their user ID")
+                .iter()
+                .next()
+                .expect("There should be a key in the signature upload request")
+                .1
+                .get(),
+        )
+        .expect("Should not fail to deserialize the key");
+
+        // Find their master key that we want to update inside their msk JSON
+        let mut their_msk: CrossSigningKey = serde_json::from_value(
+            msk_json.get(their_user_id.as_str()).expect("msk should contain their user ID").clone(),
+        )
+        .expect("Should not fail to deserialize msk");
+
+        // Find our own user signing key
+        let my_user_signing_key_id = my_identity
+            .user_signing_key()
+            .keys()
+            .iter()
+            .next()
+            .expect("There should be a user signing key")
+            .0;
+
+        // Add the signature from the SignatureUploadRequest to their master key, under
+        // our user ID
+        their_msk.signatures.add_signature(
+            my_user_id.to_owned(),
+            my_user_signing_key_id.to_owned(),
+            cross_signing_key
+                .signatures
+                .get_signature(my_user_id, my_user_signing_key_id)
+                .expect("There should be a signature for our user"),
+        );
+
+        // Create a JSON response as if the verification has happened
+        ruma_response_from_json(&json!({
+            "device_keys": {}, // Don't need devices here, even though they would exist
+            "failures": {},
+            "master_keys": {
+                their_user_id: their_msk,
+            },
+            "self_signing_keys": ssk_json,
+        }))
+    }
 }
 
 #[cfg(test)]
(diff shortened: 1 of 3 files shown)


crates/matrix-sdk-crypto/src/identities/user.rs
```diff
@@ -1266,18 +1274,103 @@ pub(crate) mod testing {
...
+
+    /// When we want to test identities that are verified, we need to simulate
+    /// the verification process. This function supports that by simulating
+    /// what happens when a successful verification dance happens and
+    /// providing the /keys/query response we would get when that happened.
+    ///
+    /// signature_upload_request will be the result of calling
+    /// [`super::OtherUserIdentity::verify`].
+    ///
+    /// msk_json should be JSON like: {
```
poljar, line 1286: Can we convert this into code and put this into a code fence, that way this will get rendered a bit nicer in various code editors.

Something like:

```
let msk_json = json!({
   ...
```

andybalaam (commented):
TODO: use conventional commit message

poljar (approved):
I think the name is fine, or at least I can't come up with a vastly better one.

crates/matrix-sdk-crypto/src/identities/user.rs
andybalaam: Done in 6ad623f795

crates/matrix-sdk-crypto/src/identities/user.rs
```diff
@@ -1266,18 +1274,126 @@ pub(crate) mod testing {
...
+
+        // Add the signature from the SignatureUploadRequest to their master key, under
+        // our user ID
+        their_msk.signatures.add_signature(
+            my_user_id.to_owned(),
+            my_user_signing_key_id.to_owned(),
+            cross_signing_key
+                .signatures
+                .get_signature(my_user_id, my_user_signing_key_id)
+                .expect("There should be a signature for our user"),
```
richvdh, line 1375: suggest moving this up to next to where we construct `cross_signing_key`. It's only needed for this one operation and conceptually they fit together.

crates/matrix-sdk-crypto/src/identities/user.rs
```diff
@@ -1266,18 +1274,126 @@ pub(crate) mod testing {
...
+                .next()
+                .expect("There should be a key in the signature upload request")
+                .1
+                .get(),
+        )
+        .expect("Should not fail to deserialize the key");
+
+        // Find their master key that we want to update inside their msk JSON
+        let mut their_msk: CrossSigningKey = serde_json::from_value(
+            msk_json.get(their_user_id.as_str()).expect("msk should contain their user ID").clone(),
```
richvdh, line 1354:
```suggestion
            msk_json.get(their_user_id.as_str()).expect("msk JSON should contain their user ID").clone(),
```

crates/matrix-sdk-crypto/src/identities/user.rs
```diff
@@ -1266,18 +1274,126 @@ pub(crate) mod testing {
...
+    pub fn simulate_key_query_response_for_verification(
+        signature_upload_request: SignatureUploadRequest,
+        my_identity: OwnUserIdentity,
+        my_user_id: &UserId,
+        their_user_id: &UserId,
+        msk_json: serde_json::Value,
+        ssk_json: serde_json::Value,
+    ) -> KeyQueryResponse {
+        // Find the signed key inside the SignatureUploadRequest
+        let cross_signing_key: CrossSigningKey = serde_json::from_str(
```
richvdh, line 1339: maybe `signed_cross_signing_key`

crates/matrix-sdk-crypto/src/identities/user.rs
```diff
@@ -1266,18 +1274,126 @@ pub(crate) mod testing {
...
+                .get(),
+        )
+        .expect("Should not fail to deserialize the key");
+
+        // Find their master key that we want to update inside their msk JSON
+        let mut their_msk: CrossSigningKey = serde_json::from_value(
+            msk_json.get(their_user_id.as_str()).expect("msk should contain their user ID").clone(),
+        )
+        .expect("Should not fail to deserialize msk");
+
```
richvdh, line 1357: It is important for this method to work correctly that the signed key in `signature_upload_request`  is the same as that in `msk_json`. We could maybe assert as much.

something like:

```rust
        let (their_msk_id, _) = their_msk.get_first_key_and_id().expect("MSK JSON should contain at least one key");
        let (signed_msk_id, _) = signed_cross_signing_key..get_first_key_and_id().expect("signature data should contain at least one key");
        assert_eq!(signed_msk_id, their_msk_id, "MSK data did not match the key that was signed in signature_upload_request");
```

crates/matrix-sdk-crypto/src/identities/user.rs
```diff
@@ -1266,18 +1274,126 @@ pub(crate) mod testing {
...
+        // Find our own user signing key
+        let my_user_signing_key_id = my_identity
+            .user_signing_key()
+            .keys()
+            .iter()
+            .next()
+            .expect("There should be a user signing key")
+            .0;
```
richvdh, line 1365: we could probably manage without figuring out our own USK id: we could just pull out the first (and only signature) in `signed_cross_signing_key`. Would mean we don't need to provide `my_identity`, which would be a win.

crates/matrix-sdk-crypto/src/identities/user.rs
```diff
@@ -1266,18 +1274,126 @@ pub(crate) mod testing {
...
+    /// ```
+    pub fn simulate_key_query_response_for_verification(
+        signature_upload_request: SignatureUploadRequest,
+        my_identity: OwnUserIdentity,
+        my_user_id: &UserId,
+        their_user_id: &UserId,
+        msk_json: serde_json::Value,
+        ssk_json: serde_json::Value,
+    ) -> KeyQueryResponse {
+        // Find the signed key inside the SignatureUploadRequest
```
richvdh, line 1338:
```suggestion
        // Extract our signature of the other user's MSK from the SignatureUploadRequest
```

crates/matrix-sdk-crypto/src/identities/user.rs
```diff
@@ -1266,18 +1274,126 @@ pub(crate) mod testing {
...
+    ///     .await
+    ///     .unwrap();
+    /// ```
+    pub fn simulate_key_query_response_for_verification(
+        signature_upload_request: SignatureUploadRequest,
+        my_identity: OwnUserIdentity,
+        my_user_id: &UserId,
+        their_user_id: &UserId,
+        msk_json: serde_json::Value,
+        ssk_json: serde_json::Value,
```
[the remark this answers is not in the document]
richvdh: We could do with better documentation on these parameters. Maybe use a bullet list under a `# Arguments` section in the doc-comment.

In particular, it's important to note that msk_json and ssk_json are the other user's cross-signing keys as we would expect to see them in a `/keys/query` response: the MSK is self-signed, and the SSK is signed by the MSK.

crates/matrix-sdk-crypto/src/identities/user.rs
richvdh: > It is important for this method to work correctly that the signed key in signature_upload_request is the same as that in msk_json

oh also, put something about that in the docs

crates/matrix-sdk-crypto/src/identities/user.rs
```diff
@@ -1266,18 +1274,126 @@ pub(crate) mod testing {
...
+        // Add the signature from the SignatureUploadRequest to their master key, under
+        // our user ID
+        their_msk.signatures.add_signature(
+            my_user_id.to_owned(),
+            my_user_signing_key_id.to_owned(),
+            cross_signing_key
+                .signatures
+                .get_signature(my_user_id, my_user_signing_key_id)
+                .expect("There should be a signature for our user"),
+        );
```
richvdh, line 1376: Another thought on this: a much more intuitive way to generate this signature is to build it ourselves, rather than abusing `OtherUserIdentity::verify` (which of course requires the offending identity to have been imported into the olm machine).

Building the signature ourselves requires getting hold of our own private USK, but that's not so bad:

```rust
    async fn get_private_usk(machine: &OlmMachine) -> Ed25519SecretKey {
        Ed25519SecretKey::from_base64(
            machine
                .export_cross_signing_keys()
                .await
                .unwrap()
                .unwrap()
                .user_signing_key
                .as_ref()
                .unwrap(),
        )
        .unwrap()
    }
```

And then signing the MSK is something like

```
        use crate::olm::utility::SignJson;
        let signature = my_usk.sign_json(serde_json::to_value(&msk_to_update).unwrap()).unwrap(),
        let my_user_signing_key_id = DeviceKeyId::from_parts(
                DeviceKeyAlgorithm::Ed25519,
                my_usk.public_key().to_base64().as_str().into(),
        );

        msk_to_update.signatures.add_signature(
            my_user_id.to_owned(),
            my_user_signing_key_id,
            signature
        );
```

The problem with that is that `crate::olm::utility::SignJson` is `pub(crate)`, but maybe we could add a `#[cfg(test)]`?

poljar: > The problem with that is that crate::olm::utility::SignJson is pub(crate), but maybe we could add a #[cfg(test)]?

Why is that a problem this PR is completely inside the crypto crate?

richvdh: > Why is that a problem this PR is completely inside the crypto crate?

Sorry, I'm confused. It's private, not pub(crate).

`utility::{SignedJsonObject, VerifyJson}` are exported from `olm` at pub(crate); maybe `SignJson` should be too.

andybalaam: This seems like a good plan but I don't have time to work on it now. Shall we make an issue for it, or just leave the info here?

poljar: > Sorry, I'm confused. It's private, not pub(crate).

I think I left that private for a reason and implemented higher level methods like `sign_device()` on things that can sign devices. It is supposed to encode in our type system which things are allowed to sign which objects.

Exposing the general signing mechanism and re-implementing `sign_user()` in a test helper doesn't sound that great to me. Refactoring `sign_user()` into  two methods, one which returns a request, and another one which returns a `CrossSigningKey` sounds better to me.

richvdh: > This seems like a good plan but I don't have time to work on it now. Shall we make an issue for it, or just leave the info here?

I can't see that an issue is likely to be very useful. It will just clutter things up and we won't remember to look for it when we come to do the work.

poljar: > I don't really agree with this: I think that having to construct the higher-level objects makes testing unnecessarily difficult, and makes tests fragile, complicated, and generally hard to maintain.

That's a fair argument, but the example Rust code you posted requires the same higher-level objects to be created as my suggestion would.

PR #4130, opened 2024-10-15
refactor(ffi): get rid of the `get_event_timeline_item` by id methods

bnjbvr:
This PR would be the cherry on cake on top of https://github.com/matrix-org/matrix-rust-sdk/pull/4100. Drafting for now, to let a bit of time for other PRs to be reviewed and merged first.

### DIFF
--- a/bindings/matrix-sdk-ffi/src/timeline/mod.rs
+++ b/bindings/matrix-sdk-ffi/src/timeline/mod.rs
@@ -53,7 +53,7 @@ use ruma::{
         },
         AnyMessageLikeEventContent,
     },
-    EventId, OwnedTransactionId,
+    EventId,
 };
 use tokio::{
     sync::Mutex,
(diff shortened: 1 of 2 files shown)


bnjbvr:
@jmartinesp @stefanceriu curious about your take on this one! it attempts to remove APIs that I'm not sure we'd need anymore, but if we did, I'd be curious to learn how/why!

stefanceriu:
We use it in one single place, [when editing a poll](https://github.com/element-hq/element-x-ios/blob/develop/ElementX/Sources/Services/Timeline/TimelineProxy.swift#L461) but that feels like a mistake now seeing that we have the original eventID string and we can eaisly construct an `EventOrTransactionId` from it 🤔

jmartinesp:
We do need it at the moment because we map the records Rust returns into our own Kotlin object representations and we don't have a way to map it back to a Rust object. We *could* maybe pass it to these representations and override the equals implementation so it check everything except for this object, or maybe add some kind of cache for our Timeline object that holds onto this original content.

Another option would be a function that creates a `MessageContent` from just a `MessageType`, which we can probably create from the timeline event.

bnjbvr:
Ok, I've changed it to the "simpler" version where only `getEventTimelineItemByTransactionId` is removed, since it's unused in both EX apps as far as I can tell. The change for `byEventId` can be done later, if we have some time on the EXA side for that refactoring too.

jmartinesp (approved):
LGTM! Sorry for the delay:

1. I just realised I'm the reviewer of this PR and not just mentioned in it 😅 .
2. I wanted to double check we're not using this API, I just confirmed we're not.

PR #4131, opened 2024-10-16
refactor(oidc): allow passing a `Prompt` to get an OIDC url

jmartinesp:
This allows clients to directly open the web page they want: the login one, the registration one, consent, etc. It should improve the UX in the registration flow since we can now skip the login one.

Requested in https://github.com/element-hq/element-x-android/pull/3635.

Signed-off-by:

### DIFF
--- a/bindings/matrix-sdk-ffi/src/client.rs
+++ b/bindings/matrix-sdk-ffi/src/client.rs
@@ -19,6 +19,7 @@ use matrix_sdk::{
             registration::{
                 ClientMetadata, ClientMetadataVerificationError, VerifiedClientMetadata,
             },
+            requests::Prompt as SdkOidcPrompt,
         },
         OidcAuthorizationData, OidcSession,
     },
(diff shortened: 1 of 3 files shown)


stefanceriu (approved):
Very nice! 👏

bindings/matrix-sdk-ffi/src/client.rs
```diff
@@ -1731,3 +1734,48 @@ impl TryFrom<SlidingSyncVersion> for SdkSlidingSyncVersion {
...
+    /// account.
+    ///
+    /// Defined in [Initiating User Registration via OpenID Connect](https://openid.net/specs/openid-connect-prompt-create-1_0.html).
+    Create,
+
+    /// An unknown value.
+    Unknown { value: String },
+}
+
+impl From<OidcPrompt> for SdkOidcPrompt {
```
bnjbvr, line 1770: Consider annotating the `SdkOidcPrompt` with the uniffi derive in place of adding the wrapper, since the wrapper seems completely equivalent?

bnjbvr (approved):

bindings/matrix-sdk-ffi/src/client.rs
jmartinesp: `Prompt` comes from [an external crate](https://docs.rs/oauth2-types/0.11.0/oauth2_types/requests/enum.Prompt.html), so I think there's no way to export it? Also, the enum uses the `Case(value)` form, which is not allowed by uniffi 🫤 .

jmartinesp: It's true the name makes you think the type comes from the SDK... I can rename it to something better in another PR, but nothing came to mind. `OAuth2OidcPrompt`? Should I just use the full `oauth2types::requests::Prompt` form?

bnjbvr: Do not fret too much about it, sorry, this was more a light remark at the existing code. But yeah using the fully-qualified type would look great here!

PR #4132, opened 2024-10-16
Remove `image-proc` feature and functions to generate a thumbnail

zecakeh:
Since no one seems to use it, according to our discussion [in the Dev room](https://matrix.to/#/!qSsPTKDfMGYqhgiLPJ:flipdot.org/$Aot5AwX56U1mPA8-24hgNhL-Di1joPxyDpoESJ3qbEQ?via=matrix.org&via=element.io&via=acter.global).

The second commit, on top of the first, allows to remove a bunch of dependencies in `Cargo.lock`.


jmartinesp (approved):
Thanks!

PR #4133, opened 2024-10-16
doc(contributing): Recommend --interactive for git rebase --autosquash

andybalaam:
Older versions of Git require --interactive when we supply --autosquash, and it's also probably a good idea generally.

See https://stackoverflow.com/a/77663575/22610 for more info.


poljar (approved):
So that's why it autosquash didn't work for some people without `--interactive` mode.

People don't update their software :cry:

PR #4134, opened 2024-10-16
fix(crypto): Don't warn about verified users when subscribing to identity updates

andybalaam:


andybalaam:
Sorry for force-pushing - github wasn't triggering the checks.

testing/matrix-sdk-test/src/test_json/keys_query_sets.rs
```diff
@@ -486,7 +486,7 @@ impl IdentityChangeDataSet {
         })
     }
 
-    fn msk_a() -> Value {
+    pub fn msk_a() -> Value {
```
poljar, line 489: Not your fault, but we avoid the MSK abbreviation in the SDK. No need to rename this but I wanted to remind people.

crates/matrix-sdk/src/room/identity_status_changes.rs
```diff
@@ -110,10 +116,25 @@ impl IdentityStatusChanges {
...
+    mut input: Vec<IdentityStatusChange>,
+    own_user_id: &UserId,
+) -> Vec<IdentityStatusChange> {
+    // We are never interested in changes to our own identity, and also for initial
+    // updates, we are only interested in "bad" states where we need to
+    // notify the user, so we can remove Verified states (Pinned states are
+    // already missing, because Pinned is considered the default).
+    input.retain(|change| {
+        change.user_id != own_user_id && change.changed_to != IdentityState::Verified
+    });
```
poljar, line 129:
```suggestion
    });
    
```

poljar (approved):
Makes sense. Thanks.

testing/matrix-sdk-test/src/test_json/keys_query_sets.rs
andybalaam: :+1:  Will update in a later PR.

crates/matrix-sdk/src/room/identity_status_changes.rs
andybalaam: Will fix in a final rebase.

PR #4135, opened 2024-10-16
feat(multiverse): add support to toggle a reaction on the last message

bnjbvr:
Does what it says on the tin. I was trying to reproduce weird logs from #4127, with no success.


bnjbvr:
Wait, this is using the new API for `toggle_reaction`, so this can't be landed as a separate PR :facepalm:

Issue #3973, opened 2024-09-10
send_queue: persist send errors


pixlwave:
When a message fails to send with one of the `EventSendState` cases below, the next time the timeline is opened it send state has become `SendingFailed`:

- `VerifiedUserHasUnsignedDevice`
- `VerifiedUserChangedIdentity`
- `CrossSigningNotSetup`
- `SendingFromUnverifiedDevice`

These cases don't exist in the UI crate as they're mapped [directly in the FFI](https://github.com/matrix-org/matrix-rust-sdk/blob/1dd8c908c590ac1dc78815869d18ea79c3cd14ba/bindings/matrix-sdk-ffi/src/timeline/mod.rs#L983-L1004), and so they aren't being persisted to the store.

In practice this looks like so Element X, meaning the user can't resolve the send failure once they've closed the room:

https://github.com/user-attachments/assets/c4a9c4fa-efeb-403b-863c-2626410730cc

bnjbvr:
Those errors would need to be saved in the send queue on-disk backend:

- change `QueuedEvent` in the base state store trait so it contains the error: let's replace `is_wedged: bool` by an `error: Option<YourErrorEnumHere>`, so `is_wedged` become equivalent to `error.is_some()`
- update databases/backends so they include this error, with a minimal migration: is_wedged = true => the error is set to Some(dummy)
- in [`QueueStorage::mark_as_wedged`](https://github.com/matrix-org/matrix-rust-sdk/blob/ef6237045e999f15a6cf652296e69c0ac6bc33d3/crates/matrix-sdk/src/send_queue.rs#L681), include the observed error when saving it
- change a `LocalEchoContent::Event` so it includes the optional error instead of [`is_wedged`](https://github.com/matrix-org/matrix-rust-sdk/blob/ef6237045e999f15a6cf652296e69c0ac6bc33d3/crates/matrix-sdk/src/send_queue.rs#L1137)
- use this error when handling a local echo in the timeline, [instead of using a dummy error](https://github.com/matrix-org/matrix-rust-sdk/blob/ef6237045e999f15a6cf652296e69c0ac6bc33d3/crates/matrix-sdk-ui/src/timeline/controller/mod.rs#L1224-L1237)
- also, don't forget to test this :clown_face: 

Wanted to say good luck, we're all counting on you.

PR #4137, opened 2024-10-16
feat(send_queue): Persist failed to send errors

BillCarsonFr:
Modify the SendQueue in order to persist the error that cause the event to fail to send as a `QueueWedgeError`. The `QueueWedgeError` is not a 1:1 mapping for all kinds of errors, but holds variant and information that the client can react to in order to propose "quick fixes"/solution before retrying to send.

Fixes https://github.com/matrix-org/matrix-rust-sdk/issues/3973 
Also fixes https://github.com/element-hq/element-x-ios/issues/3287 because when a timeline reset occurs the fail to send reason is also lost.

This PR starts with a refactoring commit https://github.com/matrix-org/matrix-rust-sdk/commit/e7696003e846b64c41761109f326fd37c4506040 to introduce the new `QueueWedgedError` and move the logic that was in the ffi layer to convert api errors to SendState error variant. This `QueueWedgedError` can be directly use in the `SendingFailed` variant and expose to ffi.

Second commit https://github.com/matrix-org/matrix-rust-sdk/commit/109c1337465ba7965825fec858fd2a4b8b954611 adds the persistence, `QueuedEvent` now have an optional error field instead of a `is_weged` boolean. Same for LocalEchoContent::Event.
Adds also Migration for sqlite and indexeddb

Signed-off-by:

### DIFF
--- a/crates/matrix-sdk-base/src/store/traits.rs
+++ b/crates/matrix-sdk-base/src/store/traits.rs
@@ -394,6 +394,8 @@ pub trait StateStore: AsyncTraitDeps {
 
     /// Updates the send queue error status (wedge) for a given send queue
     /// event.
+    /// Set `error` to None if the problem has been resolved and the event was
+    /// finally sent.
     async fn update_send_queue_event_status(
         &self,
         room_id: &RoomId,
@@ -1154,16 +1156,14 @@ pub struct QueuedEvent {
     /// Unique transaction id for the queued event, acting as a key.
     pub transaction_id: OwnedTransactionId,
 
-    /// If the event couldn't be sent because of an API error, it's marked as
-    /// wedged, and won't ever be peeked for sending. The only option is to
-    /// remove it.
+    /// Set when the event couldn't be sent because of an unrecoverable API
+    /// error. `None` if the event is in queue for being sent.
     pub error: Option<QueueWedgeError>,
 }
 
 impl QueuedEvent {
-    /// If the event couldn't be sent because of an API error, it's marked as
-    /// wedged, and won't ever be peeked for sending. The only option is to
-    /// remove it.
+    /// True if the event couldn't be sent because of an unrecoverable API
+    /// error. See `error` for more details on the reason.
     pub fn is_wedged(&self) -> bool {
         self.error.is_some()
     }
--- a/crates/matrix-sdk-base/src/store/traits.rs
+++ b/crates/matrix-sdk-base/src/store/traits.rs
@@ -1163,7 +1163,7 @@ pub struct QueuedEvent {
 
 impl QueuedEvent {
     /// True if the event couldn't be sent because of an unrecoverable API
-    /// error. See `error` for more details on the reason.
+    /// error. See [`Self::error`] for more details on the reason.
     pub fn is_wedged(&self) -> bool {
         self.error.is_some()
     }
@@ -1185,6 +1185,7 @@ pub enum QueueWedgeError {
         /// The insecure devices as a Map of userID to deviceID.
         user_device_map: BTreeMap<OwnedUserId, Vec<OwnedDeviceId>>,
     },
+
     /// This error occurs when a previously verified user is not anymore, and
     /// the current encryption setting prohibits sharing when it happens.
     #[error("Some users that were previously verified are not anymore")]
@@ -1192,10 +1193,12 @@ pub enum QueueWedgeError {
         /// The users that are expected to be verified but are not.
         users: Vec<OwnedUserId>,
     },
+
     /// It is required to set up cross-signing and properly verify the current
     /// session before sending.
     #[error("Own verification is required")]
     CrossVerificationRequired,
+
     /// Other errors.
     #[error("Other unrecoverable error: {msg}")]
     GenericApiError {
--- a/crates/matrix-sdk-base/src/store/traits.rs
+++ b/crates/matrix-sdk-base/src/store/traits.rs
@@ -23,7 +23,7 @@ use std::{
 use as_variant::as_variant;
 use async_trait::async_trait;
 use growable_bloom_filter::GrowableBloom;
-use matrix_sdk_common::{deserialized_responses::QueueWedgeError, AsyncTraitDeps};
+use matrix_sdk_common::AsyncTraitDeps;
 use ruma::{
     api::MatrixVersion,
     events::{
@@ -37,8 +37,8 @@ use ruma::{
     },
     serde::Raw,
     time::SystemTime,
-    EventId, OwnedEventId, OwnedMxcUri, OwnedRoomId, OwnedTransactionId, OwnedUserId, RoomId,
-    TransactionId, UserId,
+    EventId, OwnedDeviceId, OwnedEventId, OwnedMxcUri, OwnedRoomId, OwnedTransactionId,
+    OwnedUserId, RoomId, TransactionId, UserId,
 };
 use serde::{Deserialize, Serialize};
 
@@ -1167,6 +1167,39 @@ impl QueuedEvent {
     pub fn is_wedged(&self) -> bool {
         self.error.is_some()
     }
+}
+
+/// Represent a failed to send unrecoverable error of an event sent via the
+/// send_queue. It is a serializable representation of a client error, see
+/// `From` implementation for more details. These errors can not be
+/// automatically retried, but yet some manual action can be taken before retry
+/// sending. If not the only solution is to delete the local event.
+#[derive(Clone, Debug, Serialize, Deserialize, thiserror::Error)]
+pub enum QueueWedgeError {
+    /// This error occurs when there are some insecure devices in the room, and
+    /// the current encryption setting prohibit sharing with them.
+    #[error("There are insecure devices in the room")]
+    InsecureDevices {
+        /// The insecure devices as a Map of userID to deviceID.
+        user_device_map: BTreeMap<OwnedUserId, Vec<OwnedDeviceId>>,
+    },
+    /// This error occurs when a previously verified user is not anymore, and
+    /// the current encryption setting prohibit sharing when it happens.
+    #[error("Some users that were previously verified are not anymore")]
+    IdentityViolations {
+        /// The users that are expected to be verified but are not.
+        users: Vec<OwnedUserId>,
+    },
+    /// It is required to set up cross-signing and properly erify the current
+    /// session before sending.
+    #[error("Own verification is required")]
+    CrossVerificationRequired,
+    /// Other errors.
+    #[error("Other unrecoverable error: {msg}")]
+    GenericApiError {
+        /// Description of the error.
+        msg: String,
+    },
 }
 
 /// The specific user intent that characterizes a [`DependentQueuedEvent`].
--- a/crates/matrix-sdk-base/src/store/traits.rs
+++ b/crates/matrix-sdk-base/src/store/traits.rs
@@ -23,7 +23,7 @@ use std::{
 use as_variant::as_variant;
 use async_trait::async_trait;
 use growable_bloom_filter::GrowableBloom;
-use matrix_sdk_common::AsyncTraitDeps;
+use matrix_sdk_common::{deserialized_responses::QueueWedgeError, AsyncTraitDeps};
 use ruma::{
     api::MatrixVersion,
     events::{
@@ -392,12 +392,13 @@ pub trait StateStore: AsyncTraitDeps {
         room_id: &RoomId,
     ) -> Result<Vec<QueuedEvent>, Self::Error>;
 
-    /// Updates the send queue wedged status for a given send queue event.
+    /// Updates the send queue error status (wedge) for a given send queue
+    /// event.
     async fn update_send_queue_event_status(
         &self,
         room_id: &RoomId,
         transaction_id: &TransactionId,
-        wedged: bool,
+        error: Option<QueueWedgeError>,
     ) -> Result<(), Self::Error>;
 
     /// Loads all the rooms which have any pending events in their send queue.
@@ -662,10 +663,10 @@ impl<T: StateStore> StateStore for EraseStateStoreError<T> {
         &self,
         room_id: &RoomId,
         transaction_id: &TransactionId,
-        wedged: bool,
+        error: Option<QueueWedgeError>,
     ) -> Result<(), Self::Error> {
         self.0
-            .update_send_queue_event_status(room_id, transaction_id, wedged)
+            .update_send_queue_event_status(room_id, transaction_id, error)
             .await
             .map_err(Into::into)
     }
@@ -1156,7 +1157,16 @@ pub struct QueuedEvent {
     /// If the event couldn't be sent because of an API error, it's marked as
     /// wedged, and won't ever be peeked for sending. The only option is to
     /// remove it.
-    pub is_wedged: bool,
+    pub error: Option<QueueWedgeError>,
+}
+
+impl QueuedEvent {
+    /// If the event couldn't be sent because of an API error, it's marked as
+    /// wedged, and won't ever be peeked for sending. The only option is to
+    /// remove it.
+    pub fn is_wedged(&self) -> bool {
+        self.error.is_some()
+    }
 }
 
 /// The specific user intent that characterizes a [`DependentQueuedEvent`].
@@ -1254,7 +1264,7 @@ impl fmt::Debug for QueuedEvent {
         // Hide the content from the debug log.
         f.debug_struct("QueuedEvent")
             .field("transaction_id", &self.transaction_id)
-            .field("is_wedged", &self.is_wedged)
+            .field("is_wedged", &self.is_wedged())
             .finish_non_exhaustive()
     }
 }
--- a/crates/matrix-sdk-base/src/store/traits.rs
+++ b/crates/matrix-sdk-base/src/store/traits.rs
@@ -1169,28 +1169,30 @@ impl QueuedEvent {
     }
 }
 
-/// Represent a failed to send unrecoverable error of an event sent via the
-/// send_queue. It is a serializable representation of a client error, see
+/// Represents a failed to send unrecoverable error of an event sent via the
+/// send_queue.
+///
+/// It is a serializable representation of a client error, see
 /// `From` implementation for more details. These errors can not be
 /// automatically retried, but yet some manual action can be taken before retry
 /// sending. If not the only solution is to delete the local event.
 #[derive(Clone, Debug, Serialize, Deserialize, thiserror::Error)]
 pub enum QueueWedgeError {
     /// This error occurs when there are some insecure devices in the room, and
-    /// the current encryption setting prohibit sharing with them.
+    /// the current encryption setting prohibits sharing with them.
     #[error("There are insecure devices in the room")]
     InsecureDevices {
         /// The insecure devices as a Map of userID to deviceID.
         user_device_map: BTreeMap<OwnedUserId, Vec<OwnedDeviceId>>,
     },
     /// This error occurs when a previously verified user is not anymore, and
-    /// the current encryption setting prohibit sharing when it happens.
+    /// the current encryption setting prohibits sharing when it happens.
     #[error("Some users that were previously verified are not anymore")]
     IdentityViolations {
         /// The users that are expected to be verified but are not.
         users: Vec<OwnedUserId>,
     },
-    /// It is required to set up cross-signing and properly erify the current
+    /// It is required to set up cross-signing and properly verify the current
     /// session before sending.
     #[error("Own verification is required")]
     CrossVerificationRequired,
--- a/crates/matrix-sdk-common/src/deserialized_responses.rs
+++ b/crates/matrix-sdk-common/src/deserialized_responses.rs
@@ -12,7 +12,10 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use std::{collections::BTreeMap, fmt};
+use std::{
+    collections::{BTreeMap, HashMap},
+    fmt,
+};
 
 use ruma::{
     events::{AnyMessageLikeEvent, AnySyncTimelineEvent, AnyTimelineEvent},
@@ -700,6 +703,50 @@ impl From<SyncTimelineEventDeserializationHelperV0> for SyncTimelineEvent {
         };
 
         SyncTimelineEvent { kind, push_actions }
+    }
+}
+
+/// Represent a failed to send error of an event sent via the send_queue.
+/// These errors are not automatically retrievable, but yet some manual action
+/// can be taken before retry sending.
+#[derive(Clone, Debug, Serialize, Deserialize)]
+#[cfg_attr(feature = "uniffi", derive(uniffi::Enum))]
+pub enum QueueWedgeError {
+    /// This error occurs when there are some insecure devices in the room, and
+    /// the current encryption setting prohibit sharing with them.
+    InsecureDevices {
+        /// The insecure devices as a Map of userID to deviceID.
+        user_device_map: HashMap<String, Vec<String>>,
+    },
+    /// This error occurs when a previously verified user is not anymore, and
+    /// the current encryption setting prohibit sharing when it happens.
+    IdentityViolations {
+        /// The users that are expected to be verified but are not.
+        users: Vec<String>,
+    },
+    /// It is required to set up cross-signing and properly verify the current
+    /// session before sending.
+    CrossVerificationRequired,
+    /// Other errors.
+    GenericApiError { msg: String },
+}
+
+/// Simple display implementation that strips out userIds/DeviceIds to avoid
+/// accidental logging.
+impl fmt::Display for QueueWedgeError {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            QueueWedgeError::InsecureDevices { .. } => {
+                f.write_str("There are insecure devices in the room")
+            }
+            QueueWedgeError::IdentityViolations { .. } => {
+                f.write_str("Some users that were previously verified are not anymore")
+            }
+            QueueWedgeError::CrossVerificationRequired => {
+                f.write_str("Own verification is required")
+            }
+            QueueWedgeError::GenericApiError { msg } => f.write_str(msg),
+        }
     }
 }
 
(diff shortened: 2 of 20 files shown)


crates/matrix-sdk-base/src/store/traits.rs
```diff
@@ -1156,7 +1157,16 @@ pub struct QueuedEvent {
     /// If the event couldn't be sent because of an API error, it's marked as
     /// wedged, and won't ever be peeked for sending. The only option is to
     /// remove it.
-    pub is_wedged: bool,
+    pub error: Option<QueueWedgeError>,
```
jmartinesp, line 1160: Should we update this doc comment, since an identical one was added below to `is_wedged()`?

crates/matrix-sdk-sqlite/migrations/state_store/007_a_send_queue_wedge_reason.sql
```diff
@@ -0,0 +1,4 @@
+-- New send queue events, now persists the type of error causing it to be wedged
+ALTER TABLE "send_queue_events"
+    -- Used as a value, thus encrypted/decrypted
+    ADD COLUMN "wedge_reason" BLOB;
```
jmartinesp, line 4: I don't know how this works in Rust, would this still work if `QueueWedgeError` adds/removes some case?

jmartinesp (commented):
LGTM in general, but some clarification about the serialization/deserialization of the BLOB in the DB when there are new/removed enum cases would be good.

Also, since this is related to the send queue, maybe a 2nd pair of eyes from @bnjbvr wouldn't hurt either.

crates/matrix-sdk-ui/src/timeline/controller/mod.rs
```diff
@@ -1521,9 +1522,8 @@ impl TimelineController {
     }
 }
 
-#[derive(Debug, Error)]
-#[error("local echo failed to send in a previous session")]
-struct MissingLocalEchoFailError;
+const MISSING_LOCAL_ECHO_FAIL_ERROR: &'static str =
```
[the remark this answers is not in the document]
bnjbvr: +1, let's inline this.

crates/matrix-sdk-ui/src/timeline/controller/mod.rs
```diff
@@ -1262,9 +1263,9 @@ impl<P: RoomDataProvider> TimelineController<P> {
                         EventSendState::SendingFailed {
                             // Put a dummy error in this case, since we're not persisting the errors
                             // that occurred in previous sessions.
-                            error: Arc::new(matrix_sdk::Error::UnknownError(Box::new(
-                                MissingLocalEchoFailError,
-                            ))),
+                            error: QueueWedgeError::GenericApiError {
+                                msg: MISSING_LOCAL_ECHO_FAIL_ERROR.into(),
```
bnjbvr: nit: for `&str` to `String` conversions, we prefer `.to_owned()`.

crates/matrix-sdk-common/src/deserialized_responses.rs
```diff
@@ -703,6 +706,50 @@ impl From<SyncTimelineEventDeserializationHelperV0> for SyncTimelineEvent {
...
+/// These errors are not automatically retrievable, but yet some manual action
+/// can be taken before retry sending.
+#[derive(Clone, Debug, Serialize, Deserialize)]
+#[cfg_attr(feature = "uniffi", derive(uniffi::Enum))]
+pub enum QueueWedgeError {
+    /// This error occurs when there are some insecure devices in the room, and
+    /// the current encryption setting prohibit sharing with them.
+    InsecureDevices {
+        /// The insecure devices as a Map of userID to deviceID.
+        user_device_map: HashMap<String, Vec<String>>,
```
bnjbvr, line 719: Outside the FFI layer, we're not using stringly-typing, so this would become a `HashMap<OwnedUserId, Vec<OwnedDeviceId>>`, and similarly below.

crates/matrix-sdk-common/src/deserialized_responses.rs
```diff
@@ -703,6 +706,50 @@ impl From<SyncTimelineEventDeserializationHelperV0> for SyncTimelineEvent {
     }
 }
 
+/// Represent a failed to send error of an event sent via the send_queue.
+/// These errors are not automatically retrievable, but yet some manual action
+/// can be taken before retry sending.
+#[derive(Clone, Debug, Serialize, Deserialize)]
+#[cfg_attr(feature = "uniffi", derive(uniffi::Enum))]
+pub enum QueueWedgeError {
```
bnjbvr, line 714: Since this is related to the send queue, can we put it in the send queue file? Or is there a reason I'm missing to put it into deserialized_responses?

Also, should this use our common pattern of deriving `thiserror::Error`, to make it clear it's an error type, and would avoid the manual display impl? (uniffi also has an `uniffi::Error` derive which could be convenient here)

crates/matrix-sdk-common/src/deserialized_responses.rs
```diff
@@ -703,6 +706,50 @@ impl From<SyncTimelineEventDeserializationHelperV0> for SyncTimelineEvent {
...
+    /// the current encryption setting prohibit sharing with them.
+    InsecureDevices {
+        /// The insecure devices as a Map of userID to deviceID.
+        user_device_map: HashMap<String, Vec<String>>,
+    },
+    /// This error occurs when a previously verified user is not anymore, and
+    /// the current encryption setting prohibit sharing when it happens.
+    IdentityViolations {
+        /// The users that are expected to be verified but are not.
+        users: Vec<String>,
```
bnjbvr, line 725: `Vec<OwnedUserId>`

crates/matrix-sdk-common/src/deserialized_responses.rs
```diff
@@ -703,6 +706,50 @@ impl From<SyncTimelineEventDeserializationHelperV0> for SyncTimelineEvent {
...
+    },
+    /// It is required to set up cross-signing and properly verify the current
+    /// session before sending.
+    CrossVerificationRequired,
+    /// Other errors.
+    GenericApiError { msg: String },
```
bnjbvr, line 731: Could it be a `Box<dyn Error>` or something similar? It's a bit weird to store the stringified error, since API users can't match on it, unless they precisely know what string to expect.

crates/matrix-sdk/tests/integration/send_queue.rs
```diff
@@ -471,8 +472,7 @@ async fn test_error_then_locally_reenabling() {
     // seconds).
     // It's the same transaction id that's used to signal the send error.
     let error = assert_update!(watch => error { recoverable=true, txn=txn1 });
-    let error = error.as_client_api_error().unwrap();
-    assert_eq!(error.status_code, 500);
+    assert_matches!(error, QueueWedgeError::GenericApiError { .. });
```
bnjbvr, line 475: This change justifies why we need to store the error itself, in the case of `GenericApiError`: otherwise here we're loosing information that it could be a client API error. Maybe `GenericApiError` could contain a (boxed or not) `matrix_sdk::Error`, as this is the most general error type we have in this crate.

crates/matrix-sdk-base/src/store/traits.rs
```diff
@@ -392,12 +392,13 @@ pub trait StateStore: AsyncTraitDeps {
         room_id: &RoomId,
     ) -> Result<Vec<QueuedEvent>, Self::Error>;
 
-    /// Updates the send queue wedged status for a given send queue event.
+    /// Updates the send queue error status (wedge) for a given send queue
+    /// event.
```
bnjbvr, line 396: Can you clarify the contract around what it means when `error` is not set? (I suppose that means "unwedge", but it's not specified here, and not obvious at first sight)

crates/matrix-sdk-base/src/store/traits.rs
```diff
@@ -1156,7 +1157,16 @@ pub struct QueuedEvent {
     /// If the event couldn't be sent because of an API error, it's marked as
     /// wedged, and won't ever be peeked for sending. The only option is to
     /// remove it.
-    pub is_wedged: bool,
+    pub error: Option<QueueWedgeError>,
+}
+
+impl QueuedEvent {
+    /// If the event couldn't be sent because of an API error, it's marked as
```
bnjbvr, line 1164: nit: this is context around what it means to have an error; I'd rather put that near the `error` field itself, and have the doc comment here say something like "returns whether the queued event is wedged (i.e. ran into an unrecoverable error) or not"

crates/matrix-sdk-indexeddb/src/state_store/mod.rs
```diff
@@ -1432,7 +1440,19 @@ impl_state_store!({
...
+                        if legacy_wedged {
+                            // migrate a generic error
+                            Some(QueueWedgeError::GenericApiError {
+                                msg: "local echo failed to send in a previous session".into(),
+                            })
+                        } else {
+                            None
+                        }
+                    }
+                    None => item.error,
```
bnjbvr, line 1454: This doesn't correctly handle the case where legacy is set to false, but there's still a new-form error. I think you want something like:

```rust
Some(true) => // migrate,
_ => item.error
```

crates/matrix-sdk-indexeddb/src/state_store/mod.rs
```diff
@@ -432,7 +433,12 @@ struct PersistedQueuedEvent {
     // All these fields are the same as in [`QueuedEvent`].
     event: SerializableEventContent,
     transaction_id: OwnedTransactionId,
-    is_wedged: bool,
+
+    // Deprecated (from old format), now replaced with error field.
+    // Kept here for migration
+    is_wedged: Option<bool>,
```
bnjbvr, line 439: Does the conversion from bool to `Option<bool>` work? (i.e. is there a test for that?)

I think it may work because we're using JSON as the serialization format, but it's not guaranteed it would work in general with any serialization format, so I'd rather have a test for this please.

crates/matrix-sdk-sqlite/migrations/state_store/007_a_send_queue_wedge_reason.sql
bnjbvr: Yeah, the value is serialized (I suspect to JSON bytes) before being put into the table, and it has to be a blob because of the encryption case.

crates/matrix-sdk-sqlite/src/state_store.rs
```diff
@@ -261,6 +261,41 @@ impl SqliteStateStore {
...
+
+                    let (room_id, transaction_id) = wedged_entries?;
+
+                    txn.prepare_cached("UPDATE send_queue_events SET wedge_reason = ? WHERE room_id = ? AND transaction_id = ?")?
+                        .execute((default_err.clone(), room_id, transaction_id))?;
+                }
+
+
+                // Clean up the table now that data is migrated
+                txn.execute_batch(include_str!("../migrations/state_store/007_b_send_queue_clean.sql"))?;
```
bnjbvr, line 293: Nice! Good job on the migration.

crates/matrix-sdk-sqlite/migrations/state_store/007_a_send_queue_wedge_reason.sql
bnjbvr: Is there no need to annotate this to mark it nullable? and that the default value is null when not set? Maybe it's the default with sqlite, and I don't know, but might be worth indicating in the comment then?

crates/matrix-sdk-sqlite/src/state_store.rs
```diff
@@ -1744,17 +1779,17 @@ impl StateStore for SqliteStateStore {
...
         transaction_id: &TransactionId,
-        wedged: bool,
+        error: Option<QueueWedgeError>,
     ) -> Result<(), Self::Error> {
         let room_id = self.encode_key(keys::SEND_QUEUE, room_id);
 
         // See comment in `save_send_queue_event`.
         let transaction_id = transaction_id.to_string();
-
+        let error_value = error.map(|e| self.serialize_value(&e)).transpose()?;
```
bnjbvr, line 1788: uber-nit: since the transpose is not trivial, can you add a newline before and after, so we notice there's something going on here, please? (and it'll make the code breathe :))

crates/matrix-sdk-sqlite/src/state_store.rs
```diff
@@ -2213,4 +2256,76 @@ mod migration_tests {
...
+        let room_a_id = room_id!("!room_a:dummy.local");
+        let wedged_event_transaction_id = TransactionId::new();
+        let local_event_transaction_id = TransactionId::new();
+
+        // Create and populate db.
+        {
+            let db = create_fake_db(&path, 7).await.unwrap();
+            let conn = db.pool.get().await.unwrap();
+
+            let this = db.clone();
```
bnjbvr, line 2273: I think we can get rid of the local binding and clone here.

crates/matrix-sdk/src/send_queue.rs
```diff
@@ -601,7 +604,7 @@ fn convert_error_to_wedged_reason(value: &crate::Error) -> QueueWedgeError {
             }
             SessionRecipientCollectionError::VerifiedUserChangedIdentity(users) => {
                 QueueWedgeError::IdentityViolations {
-                    users: users.iter().map(OwnedUserId::to_string).collect(),
+                    users: users.iter().map(ruma::OwnedUserId::to_string).collect(),
```
bnjbvr, line 607: nit: unnecessary qualification here

crates/matrix-sdk/src/send_queue.rs
```diff
@@ -1171,8 +1175,8 @@ pub enum LocalEchoContent {
         /// A handle to manipulate the sending of the associated event.
         send_handle: SendHandle,
         /// Whether trying to send this local echo failed in the past with an
-        /// unrecoverable error (see [`SendQueueRoomError::is_recoverable`]).
-        is_wedged: bool,
+        /// error (see [`SendQueueRoomError::is_recoverable`]).
```
bnjbvr, line 1178: Let's keep `unrecoverable` here. It has a precise meaning in the context of this file, i.e. retrying would lead to the same error again and again (vs network errors that may be transient).

bnjbvr (changes requested):
Good job overall! I have issues with the first commit which loses us a bit of static typing and checkability (because of the use of stringly typed values and errors), so I'd like to take another look, otherwise the code looks great and most of my comments should be minor. Thanks!

BillCarsonFr:
> Good job overall! I have issues with the first commit which loses us a bit of static typing and checkability (because of the use of stringly typed values and errors), so I'd like to take another look, otherwise the code looks great and most of my comments should be minor. Thanks!

Yeah, I started by doing this refactoring because persisting all the different kind of [errors](https://github.com/matrix-org/matrix-rust-sdk/blob/a901506a536bf12b081eea6d6c70777d0fdcc565/crates/matrix-sdk/src/error.rs#L257) required to make a lot of things serializable, and at the end this is not used at all by the application layer (that was doing a mapping of the error to usable SendingFailed types).
What the app was using was more the thing that we really interesting into persisting. 
It's not really the error that the app layer really need, I think of it as a suggestion a compiler could give you when compilation fails. Lik "Re-verify this user or whitelist the problematic device" then retry.

I could have done without the `stringly typed values` and just put a `GenericUnrecoverableError`, the only suggestion is to delete the message. The reason why doesn't matter much as it cannot be presented to the user, it will be an english only technical string.

bnjbvr:
Good point about needing to ser/de all the error types. In that case, can we keep the String for the generic unrecoverable error case, and use proper types for the other error variants?

BillCarsonFr:
> Good point about needing to ser/de all the error types. In that case, can we keep the String for the generic unrecoverable error case, and use proper types for the other error variants?

Hello @bnjbvr, What do you think about the following changes:
So we are looking at all the errors that could happen, and we are doing 2 things:
 - Signalling the errors via `RoomSendQueueUpdate` and `SendQueueRoomError` and telling if is recoverable or not.
 - Persisting the errors that are unrecoverable in the queue.
 
 So my idea is that we keep the global sdk `Error` type for signalling. But for unrecoverable errors that we want to persist we use the new `QueueWedgeError` (that will stringify some types not really useful).
 
 See this commit for this change: https://github.com/matrix-org/matrix-rust-sdk/pull/4137/commits/e824fd24d273319ecb15116bec7384b6b8dbc842
 
 As requested I also put back the ruma OwnedUserId/DeviveId inside the QueueWedgeError variant instead of raw string. I did that initially so that I can directly create ffi bindings. But I see the point, and now I separated 2 versions of the error with a string one for bindings.
 
 https://github.com/matrix-org/matrix-rust-sdk/pull/4137/commits/3e657538a61fd298255504f75b9b2c20fdd75b82

crates/matrix-sdk-sqlite/migrations/state_store/007_a_send_queue_wedge_reason.sql
BillCarsonFr: Added some comments. And yes the default in NULLABLE and NULL.

crates/matrix-sdk-common/src/deserialized_responses.rs
BillCarsonFr: see https://github.com/matrix-org/matrix-rust-sdk/pull/4137/commits/3e657538a61fd298255504f75b9b2c20fdd75b82

crates/matrix-sdk-common/src/deserialized_responses.rs
BillCarsonFr: Send Queue is in `matrix-sdk` and this error is persisted in the state storage `matrix-sdk-base`. 
So it cannot be in the send_queue.rs file.

crates/matrix-sdk-common/src/deserialized_responses.rs
BillCarsonFr: see here https://github.com/matrix-org/matrix-rust-sdk/pull/4137#issuecomment-2425914825

crates/matrix-sdk-common/src/deserialized_responses.rs
BillCarsonFr: see here https://github.com/matrix-org/matrix-rust-sdk/pull/4137#issuecomment-2425914825

crates/matrix-sdk/tests/integration/send_queue.rs
BillCarsonFr: Fixed now, see here details https://github.com/matrix-org/matrix-rust-sdk/pull/4137#issuecomment-2425914825

crates/matrix-sdk/src/send_queue.rs
BillCarsonFr: Outdated now, but it was previously because of the `#[cfg(feature = "e2e-encryption")]`. Other option could be to add the feature requirement at the import level (if not would be unused import)

crates/matrix-sdk-common/src/deserialized_responses.rs
```diff
@@ -703,6 +703,36 @@ impl From<SyncTimelineEventDeserializationHelperV0> for SyncTimelineEvent {
...
+/// Represent a failed to send unrecoverable error of an event sent via the
+/// send_queue. It is a serializable representation of a client error, see
+/// `From` implementation for more details. These errors can not be
+/// automatically retried, but yet some manual action can be taken before retry
+/// sending. If not the only solution is to delete the local event.
+#[derive(Clone, Debug, Serialize, Deserialize, thiserror::Error)]
+pub enum QueueWedgeError {
+    /// This error occurs when there are some insecure devices in the room, and
+    /// the current encryption setting prohibit sharing with them.
+    #[error("There are insecure devices in the room")]
```
bnjbvr, line 715: With this error message, you won't see the `user_device_map` in the display impl. Saying here in case it's not intended; to display them, you'd need to add something like `: {user_device_map:?}` to the string above, and it will include the debug display of the btreemap.

crates/matrix-sdk-common/src/deserialized_responses.rs
```diff
@@ -703,6 +703,36 @@ impl From<SyncTimelineEventDeserializationHelperV0> for SyncTimelineEvent {
...
+    InsecureDevices {
+        /// The insecure devices as a Map of userID to deviceID.
+        user_device_map: BTreeMap<OwnedUserId, Vec<OwnedDeviceId>>,
+    },
+    /// This error occurs when a previously verified user is not anymore, and
+    /// the current encryption setting prohibit sharing when it happens.
```
bnjbvr, line 721:
```suggestion
    /// the current encryption setting prohibits sharing when it happens.
```

crates/matrix-sdk-common/src/deserialized_responses.rs
```diff
@@ -703,6 +703,36 @@ impl From<SyncTimelineEventDeserializationHelperV0> for SyncTimelineEvent {
...
+    #[error("Some users that were previously verified are not anymore")]
+    IdentityViolations {
+        /// The users that are expected to be verified but are not.
+        users: Vec<OwnedUserId>,
+    },
+    /// It is required to set up cross-signing and properly erify the current
```
bnjbvr, line 727:
```suggestion
    /// It is required to set up cross-signing and properly verify the current
```

crates/matrix-sdk/src/send_queue.rs
```diff
@@ -577,6 +589,28 @@ impl RoomSendQueue {
     }
 }
 
+impl From<&crate::Error> for QueueWedgeError {
```
bnjbvr, line 592: nit: using a `From` impl makes it part of the public API of `QueueWedgeError`, which we probably don't want to do, since it's an internal conversion. Can you use a plain function instead, please?

Also, take the `crate::Error`  by ownership (i.e. no ref `&`), since this could avoid some cloning if I'm reading the code correctly.

crates/matrix-sdk/src/send_queue.rs
```diff
@@ -1136,7 +1171,7 @@ pub enum LocalEchoContent {
         send_handle: SendHandle,
         /// Whether trying to send this local echo failed in the past with an
         /// unrecoverable error (see [`SendQueueRoomError::is_recoverable`]).
-        is_wedged: bool,
+        send_error: Option<QueueWedgeError>,
```
bnjbvr, line 1174: I might have misunderstood your intent, I thought the goal was to use a `matrix_sdk::Error` here, since it's the most generic type for signalling the error to the caller. Can we do that?

crates/matrix-sdk-ui/tests/integration/timeline/queue.rs
```diff
@@ -277,9 +278,13 @@ async fn test_reloaded_failed_local_echoes_are_marked_as_failed() {
...
-    assert_matches!(
-        error.as_client_api_error().unwrap().error_kind(),
-        Some(ruma::api::client::error::ErrorKind::TooLarge)
+    let msg = assert_matches!(
+        error,
+        QueueWedgeError::GenericApiError { msg } => msg
+    );
+    assert_eq!(
+        msg,
+        "the server returned an error: [413 / M_TOO_LARGE] Sounds like you have a lot to say!"
```
bnjbvr, line 287: (with respect to the previous comment, this would mean undoing this, since we could still pattern-match the right client API error here)

crates/matrix-sdk-ui/src/timeline/tests/echo.rs
```diff
@@ -66,15 +65,12 @@ async fn test_remote_echo_full_trip() {
     // Scenario 2: The local event has not been sent to the server successfully, it
     // has failed. In this case, there is no event ID.
     {
-        let some_io_error = Error::Io(io::Error::new(io::ErrorKind::Other, "this is a test"));
+        let some_io_error = QueueWedgeError::GenericApiError { msg: "this is a test".to_owned() };
```
bnjbvr, line 68: uber-nit: let's rename `some_io_error` to `some_error` :nerd_face:

bnjbvr (changes requested):
Thanks for the changes, this is on the right track. I might have misunderstood the intent to store the most generic error type when signalling the error to callers, and it seems we can achieve this; can you try it out please? :)

crates/matrix-sdk-indexeddb/src/state_store/mod.rs
BillCarsonFr: added here https://github.com/matrix-org/matrix-rust-sdk/pull/4137/commits/93c452872cae5915fa97b7f8eed9c55150d7adcf

crates/matrix-sdk/src/send_queue.rs
bnjbvr: Why not? After deserializing a `QueueWedgeError`, isn't there a way to turn it into a `matrix_sdk::Error`, by the way of conversion?

bnjbvr: Answering myself: no, and prior to the patch we only had the bool "is wedged or not", so this is already quite an improvement. And indeed in other contexts that we care more about, we can keep the `crate::Error`, so the current code is fine as is. Resolving this comment, thanks for the ping :)

crates/matrix-sdk/src/send_queue.rs
BillCarsonFr: It is used in the timeline controller to update the in memory local event EventSendState. Why is it a issue that it is a public API?

crates/matrix-sdk-ui/tests/integration/timeline/queue.rs
BillCarsonFr: Not for this one see https://github.com/matrix-org/matrix-rust-sdk/pull/4137#discussion_r1808911747
EventSendState is using `QueueWedgeError` and not the full `crate::Error`. Notice that for this particular error we propbably will want to add a proper variant for `QueueWedgeError` to have a proper media local echo feedback

bnjbvr:
So as to help landing this PR, I've pushed a small commit that keeps the whole `matrix_sdk::Error` for the timeline, so that we don't lose any information that we shouldn't have to lose. At the FFI layer it's always converted into a `QueueWedgeError`, which makes it easy to pattern-match on the other side of the fence, so it's functionally equivalent. With this patch, I don't have more objections to landing this, but before giving my final approval, I'd like to make sure there's no changes required on our side to address the complement crypto bindings failure.

bnjbvr (approved):
(Approving assuming no dramatic changes for complement crypto)

Thanks for the PR, notably good job getting through all the layers!

PR #4138, opened 2024-10-16
chore(timeline): instrument timeline tasks with their focus and internal id prefix

bnjbvr:
This would have avoided a few hours of debugging where we thought there was an issue with multiple timelines spawned at the same time, and then realized it was expected because of the existence of the pinned timeline in EX apps.


stefanceriu (approved):
Yey, super nice! 👏

PR #4140, opened 2024-10-16
refactor(media): some preparatory refactorings

bnjbvr:
aka "Benjamin tries to make sense of the media code. Oh, nice: it makes sense."

### DIFF
--- a/crates/matrix-sdk/src/media.rs
+++ b/crates/matrix-sdk/src/media.rs
@@ -33,11 +33,10 @@ use ruma::{
     assign,
     events::room::{
         message::{
-            AudioInfo, AudioMessageEventContent, FileInfo, FileMessageEventContent,
-            ImageMessageEventContent, MessageType, UnstableAudioDetailsContentBlock,
-            UnstableVoiceContentBlock, VideoInfo, VideoMessageEventContent,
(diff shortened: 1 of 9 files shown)


crates/matrix-sdk/src/media.rs
```diff
@@ -542,45 +542,41 @@ impl Media {
...
                     thumbnail_info,
                 });
-                let mut image_message_event_content =
-                    ImageMessageEventContent::plain(body, url).info(Box::new(info));
-                image_message_event_content.filename = filename;
-                image_message_event_content.formatted = config.formatted_caption;
-                MessageType::Image(image_message_event_content)
+                let mut content = ImageMessageEventContent::plain(body, url).info(Box::new(info));
+                content.filename = filename;
+                content.formatted = config.formatted_caption;
```
jplatte: Could also use the `assign!` macro here (I think I saw it still being used in a bunch of other places). Same for the other changes in this commit.

poljar (approved):
Nicely done.

crates/matrix-sdk/src/media.rs
bnjbvr: Yep, `make_attachment` in the latest version does maximally use `assign!`, IMO :+1:

PR #4142, opened 2024-10-16
FFI: Expose is_verified UserIdentity method.

pixlwave:
Nothing exciting here, just exposing another method.

### DIFF
--- a/bindings/matrix-sdk-ffi/src/encryption.rs
+++ b/bindings/matrix-sdk-ffi/src/encryption.rs
@@ -461,6 +461,14 @@ impl UserIdentity {
     pub(crate) fn master_key(&self) -> Option<String> {
         self.inner.master_key().get_first_key().map(|k| k.to_base64())
     }
+
+    /// Is the user identity considered to be verified.
+    ///
+    /// If the identity belongs to another user, our own user identity needs to
+    /// be verified as well for the identity to be considered to be verified.
+    pub fn is_verified(&self) -> bool {
+        self.inner.is_verified()
+    }
 }
 
 #[derive(uniffi::Object)]


bindings/matrix-sdk-ffi/src/encryption.rs
```diff
@@ -461,6 +461,14 @@ impl UserIdentity {
     pub(crate) fn master_key(&self) -> Option<String> {
         self.inner.master_key().get_first_key().map(|k| k.to_base64())
     }
+
+    /// Is the user identity considered to be verified.
+    ///
+    /// If the identity belongs to another user, our own user identity needs to
+    /// be verified as well for the identity to be considered to be verified.
```
stefanceriu, line 468: I wonder if we should just copy paste the underlying documentation instead of rolling a new one 🤔

bnjbvr (approved):
Code looks good, I'm fine with either this code comment or one copied from the SDK. Thanks.

bindings/matrix-sdk-ffi/src/encryption.rs
stefanceriu: Ah, I see, I was looking at the one inside the crypto crate which seemed just perfect. I'm just worried something will change on one level or another and then the documentation will be out of date.

bindings/matrix-sdk-ffi/src/encryption.rs
```diff
@@ -461,6 +461,14 @@ impl UserIdentity {
...
+    pub fn is_verified(&self) -> bool {
```
bmarty, line 469: I am wondering if this data should be live? Maybe not for now but in the future the app could be more reactive if we had a way to get a live `UserIdentity`.

BillCarsonFr:
I am not sure of the full context here but if you expose `is_verified()` you probably want also to expose `was_previously_verified()` / `has_verification_violation()`.

pixlwave:
> I am not sure of the full context here but if you expose `is_verified()` you probably want also to expose `was_previously_verified()` / `has_verification_violation()`.

This is being addd to show the green badge on the user profile screen (https://github.com/element-hq/element-meta/issues/2569).

I'm unsure now, are you saying that it is possible for `is_verified` and `has_verification_violation` to both be true at the same time and we need to think about these together, or just that the other 2 might be useful?

bindings/matrix-sdk-ffi/src/encryption.rs
pixlwave: Yeah, I'd really like this to be a property on `RoomMember` that would update automatically through the member's stream if I'm honest, but I don't have a clue how to do that as the members live separately to the encryption 🫤

bmarty:
(I confirm that the API works as expected on a EXA WIP branch)

BillCarsonFr:
> > I am not sure of the full context here but if you expose `is_verified()` you probably want also to expose `was_previously_verified()` / `has_verification_violation()`.
> 
> This is being added to show the green badge on the user profile screen ([element-hq/element-meta#2569](https://github.com/element-hq/element-meta/issues/2569)).
> 
> I'm unsure now, are you saying that it is possible for `is_verified` and `has_verification_violation` to both be true at the same time and we need to think about these together, or just that the other 2 might be useful?

No `has_verification_violation` means that the user used to be verified, but is not anymore. (so should have a red badge, whereas a user that was never verified and is not should have no badge)

pixlwave:
For now this is everything we need.

pixlwave:
@bnjbvr I exposed 1 more method to allow requesting identities for unknown users from the homeserver, otherwise I'm happy to have this merged if you're ok with it 🙂

bindings/matrix-sdk-ffi/src/encryption.rs
```diff
@@ -411,19 +411,55 @@ impl Encryption {
...
+    /// Get the E2EE identity of a user from the crypto store.
     ///
-    /// Returns Ok(None) if this user does not exist.
+    /// Usually, we only have the E2EE identity of a user locally if the user
+    /// is tracked, meaning that we are both members of the same encrypted room.
     ///
-    /// Returns an error if there was a problem contacting the crypto store, or
-    /// if our client is not logged in.
+    /// To get the E2EE identity of a user even if it is not available locally
+    /// use `request_user_identity()`.
```
bnjbvr, line 420: How does the API work overall: there's both `get_user_identity()` (which may return `None`) and `request_user_identity()` (which is guaranteed to return something for rooms where the two users are present). How is the caller supposed to know which one should be called? Call `get_user_identity()` first, and if it returns nothing, try `request_user_identity()`?

If so:

- can we either document it clearly here?
- or provide a unique method that first fetches from the local crypto store, then does the remote query if not found?

bnjbvr (commented):
API question below. Looks good otherwise :)

bindings/matrix-sdk-ffi/src/encryption.rs
bnjbvr: Both work for me, but if you want to keep the current version, I'd like a comment precising this on `get_user_identity()` at least please :)

bindings/matrix-sdk-ffi/src/encryption.rs
```diff
@@ -411,52 +412,51 @@ impl Encryption {
...
+    /// This method always tries to fetch the identity from the store, which we
+    /// only have if the user is tracked, meaning that we are both members
+    /// of the same encrypted room. Optionally it can request the user from the
+    /// homeserver if not found locally.
     ///
     /// # Arguments
     ///
     /// * `user_id` - The ID of the user that the identity belongs to.
     ///
+    /// * `request_if_needed` - If true, the identity will be requested from the
```
bnjbvr, line 426: nit: doesn't match the parameter name :stuck_out_tongue:

bindings/matrix-sdk-ffi/src/encryption.rs
```diff
@@ -411,52 +412,51 @@ impl Encryption {
...
+            Ok(Some(identity)) => {
+                return Ok(Some(Arc::new(UserIdentity { inner: identity })));
+            }
+            Ok(None) => {
+                info!("No identity found in the store.");
+                if !request_from_homeserver_if_needed {
+                    return Ok(None);
+                }
+            }
+            Err(e) => {
```
bnjbvr, line 450: nit: please use `err` or `error` for this variable's name, one-letter named variables are bleh :P

bindings/matrix-sdk-ffi/src/encryption.rs
pixlwave: Oops, refactor didn't get the docs.

bindings/matrix-sdk-ffi/src/encryption.rs
pixlwave: https://github.com/matrix-org/matrix-rust-sdk/commit/dce7297fcc5eef8f9c2a97d7549e12f4aba483ed

bindings/matrix-sdk-ffi/src/encryption.rs
```diff
@@ -413,16 +414,51 @@ impl Encryption {
...
+    ///
+    /// Returns a `UserIdentity` if one is found. Returns an error if there
+    /// was an issue with the crypto store or with the request to the
+    /// homeserver.
+    ///
+    /// This will always return `None` if the client hasn't been logged in.
+    pub async fn user_identity(
         &self,
         user_id: String,
+        request_from_homeserver_if_needed: bool,
```
bmarty, line 438: I am wondering in which case(s) an sdk user could set the parameter `request_from_homeserver_if_needed` to `false`.

bindings/matrix-sdk-ffi/src/encryption.rs
```diff
@@ -413,16 +414,51 @@ impl Encryption {
...
-    /// Returns an error if there was a problem contacting the crypto store, or
-    /// if our client is not logged in.
-    pub async fn get_user_identity(
+    /// # Arguments
+    ///
+    /// * `user_id` - The ID of the user that the identity belongs to.
+    ///
+    /// * `request_from_homeserver_if_needed` - If true, the identity will be
+    ///   requested from the homeserver if it is not found in the crypto store.
+    ///   When `true` the identity returned is guaranteed to be up-to-date.
```
bmarty, line 428: > When `true` the identity returned is guaranteed to be up-to-date.

I would claim that if when `true` there is **always** a request sent to the server, no? I mean in case identity is know locally, and has changed without the SDK knowing (no network, etc.), this assumption will not be true.

bnjbvr: How would the app / the SDK know whether an identity was known locally but we should trigger the request because it was updated? In that case, we might as well spawn the request all the time, since we can't really know when it's been updated, right?

pixlwave: https://github.com/matrix-org/matrix-rust-sdk/commit/195fed7f5c1f60dc8ccc50ef92e7ccca6189f446

Issue #4051, opened 2024-09-30
RoomListItem.roomInfo().displayName() often null


ganfra:
When building the room summaries, we got a lot of `RoomListItem.roomInfo().displayName()` being `null`.
Then after some time, they get updated, but only in memory. 
Indeed, when launching the app in airplane mode, we got `null` values again for most of the rooms, even if the app had all the names before killing the app. 

I've checked the code, and the persistence to disk is only made during a next sync in the room only if there is new event in the room. So if you never get updates for a room, his display name will never be saved to disk.

It'd be great to have the `displayName` computed synchronously and not async, so we don't get null display name while we have avatar, and so it'd be saved to disk directly.

bnjbvr:
Thanks for opening an issue, good catch. Computing the `displayName` synchronously would require a deep overhaul of the compute-room-name logic, which relies on events stored in the database at the moment; we should likely make it more generic and rely on database-or-sync, and save it immediately (even if the computing is async).

Hywan:
I can replicate the bug. I believe it's the root cause for https://github.com/element-hq/element-x-ios/issues/3365 and https://github.com/element-hq/customer-success/issues/405. I'm working on it.

Hywan:
The problem is the following:

https://github.com/matrix-org/matrix-rust-sdk/blob/cabb345a1c5b2b1ae4936eb2b3383af0c312d0f3/crates/matrix-sdk-base/src/client.rs#L1287-L1300

For some room, they receive a single update, then their `RoomInfo` are persisted, and finally their names are computed.
If another update is triggered, the cached name will be persisted, otherwise it won't.

That's why, in offline, some rooms have nothing, because the name wasn't persisted.

bnjbvr:
Yes, the solution always has been to use little processors for the sync response. Quoting past self:

> we should likely make it more generic and rely on database-or-sync, and save it immediately (even if the computing is async).

PR #4143, opened 2024-10-16
Process name changes that come back on the sync api

zzorba:
In debugging code, we found that the 'name' that comes back on Room responses in sliding sync was seemingly not being processed.  This caused name changes to not be reflected in the client.

Signed-off-by: Daniel Salinas

### DIFF
--- a/crates/matrix-sdk-base/src/rooms/normal.rs
+++ b/crates/matrix-sdk-base/src/rooms/normal.rs
@@ -42,6 +42,7 @@ use ruma::{
             history_visibility::HistoryVisibility,
             join_rules::JoinRule,
             member::{MembershipState, RoomMemberEventContent},
+            name::RoomNameEventContent,
             pinned_events::RoomPinnedEventsEventContent,
             redaction::SyncRoomRedactionEvent,
             tombstone::RoomTombstoneEventContent,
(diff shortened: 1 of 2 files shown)


bnjbvr (commented):
Thanks for the PR, but we've namely deviated from this behavior (that we implemented in the past), because this doesn't work for unnamed DM rooms, where the server would compute a room name, based on the room members ("A and B and C"), which means the server-computed name is always in English, and doesn't fare well with i10n. Accepting this PR would be a regression in terms of that behavior; I suppose you're opening this because you ran into https://github.com/matrix-org/matrix-rust-sdk/issues/4051 maybe?

zzorba:
Yes, we were seeing newly created rooms not receive their name, as well as any room whose name is changed after the initial sync not being updated.

Wouldn't it be possible to change the logic for updating the room name to consider if it is_direct and ignore the stored name in that case in favor of the dynamic localized one? Or store two cached names (server state and local based on heroes) and then prefer the one that applies when accessed?

It seems like the root of the problem here is that it is deciding which name to use during the stream processing (when state might be incomplete), rather than at display time (eventually complete).

zzorba:
Turned out this was an omission of the sliding-sync proxy go system, exposed by the recent removal of processing `timeline` events.  Closing.

PR #4144, opened 2024-10-17
fix(ffi): don't panic when running into an unknown membership state

### LINKED ISSUE #1254: Handle the _Custom(PrivOwnedStr) MembershipState case in the FFI.
#1245 exposed `MembershipState` to the FFI but left ` _Custom(PrivOwnedStr)` for later. This will need to be implemented.

bnjbvr:
Fixes #1254. This adds a bandaid in the form of converting `Into` impl to `TryInto`, and then will happily ignore errors in multiple places; unknown states are logged in warnings, so we can still get some information about it, would they become an issue in the future.

### DIFF
--- a/bindings/matrix-sdk-ffi/src/room_member.rs
+++ b/bindings/matrix-sdk-ffi/src/room_member.rs
@@ -1,7 +1,7 @@
 use matrix_sdk::room::{RoomMember as SdkRoomMember, RoomMemberRole};
 use ruma::UserId;
 
-use crate::error::ClientError;
+use crate::error::{ClientError, NotYetImplemented};
 
 #[derive(Clone, uniffi::Enum)]
 pub enum MembershipState {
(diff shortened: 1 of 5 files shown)


poljar (approved):
I suppose it's better than a panic.

bnjbvr:
There's no way to expose the private custom variant's content anyways, right?

bnjbvr:
It was actually quite simple to handle custom memberships; not sure how the apps will like it? cc @stefanceriu @jmartinesp 

Re-requesting a review, since there's a non-trivial change in there.

stefanceriu:
Turns out you need to name that field but otherwise it shouldn't be a problem for the app
`error: UniFFI only supports enum variants with named fields (or no fields at all)`

bindings/matrix-sdk-ffi/src/room_member.rs
```diff
@@ -19,25 +19,40 @@ pub enum MembershipState {
...
-            matrix_sdk::ruma::events::room::member::MembershipState::Join => MembershipState::Join,
             matrix_sdk::ruma::events::room::member::MembershipState::Knock => {
-                MembershipState::Knock
+                Ok(MembershipState::Knock)
             }
             matrix_sdk::ruma::events::room::member::MembershipState::Leave => {
-                MembershipState::Leave
+                Ok(MembershipState::Leave)
+            }
+            matrix_sdk::ruma::events::room::member::MembershipState::_Custom(_) => {
```
poljar, line 49: I think that this is not the intended way, as the `_Custom` variant is hidden and instead you're supposed to use the catch all, but then other known states won't be handled (there are none other known states but the enum is non-exhaustive).

I think we can live with that despite the intentions of Ruma.

poljar (approved):

PR #4145, opened 2024-10-17
feature(ffi): Add optional `CreateRoomParameters::join_rule_override`

jmartinesp:
This allows clients to set custom join rules for a room, as would be needed for the knock-only rooms, or restricted rooms (those that can only be joined if the user is part of some other room or space).

The mapping from `CreateRoomParameters` to the `create_room::v3::Request` is now using a `TryFrom` since some parsing inside can fail and it seems better to explicitly make the action fail so it can be retried than creating a room with the wrong parameters.

Signed-off-by:

### DIFF
--- a/bindings/matrix-sdk-ffi/src/client.rs
+++ b/bindings/matrix-sdk-ffi/src/client.rs
@@ -55,7 +55,8 @@ use ruma::{
     },
     events::{
         ignored_user_list::IgnoredUserListEventContent,
-        room::power_levels::RoomPowerLevelsEventContent, GlobalAccountDataEventType,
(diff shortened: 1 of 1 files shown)


bindings/matrix-sdk-ffi/src/client.rs
```diff
@@ -1779,3 +1792,88 @@ impl From<OidcPrompt> for SdkOidcPrompt {
         }
     }
 }
+
+#[derive(uniffi::Enum)]
```
poljar, line 1796: The enum should also have some docs, not just its variants.

poljar (approved):
Left a nit, seems sensible, nicely done.

PR #4146, opened 2024-10-17
refactor(base): Remove `impl From` for `SyncTimelineEvent`

richvdh:
I feel like the ability to convert straight from a `Raw<AnySyncTimelineEvent>>` into a `SyncTimelineEvent` is somewhat over-simplified: the two are only occasionally equivalent, and it's better to be explicit.

### DIFF
--- a/crates/matrix-sdk-ui/src/timeline/event_item/mod.rs
+++ b/crates/matrix-sdk-ui/src/timeline/event_item/mod.rs
@@ -1023,7 +1023,7 @@ mod tests {
         formatted_body: &str,
         ts: u64,
     ) -> SyncTimelineEvent {
-        sync_timeline_event!({
+        SyncTimelineEvent::new(sync_timeline_event!({
             "event_id": "$eventid6",
             "sender": user_id,
             "origin_server_ts": ts,
@@ -1035,7 +1035,6 @@ mod tests {
                 "formatted_body": formatted_body,
                 "msgtype": "m.text"
             },
-        })
-        .into()
+        }))
     }
 }
--- a/crates/matrix-sdk-base/src/client.rs
+++ b/crates/matrix-sdk-base/src/client.rs
@@ -396,8 +396,10 @@ impl BaseClient {
         let mut timeline = Timeline::new(limited, prev_batch);
         let mut push_context = self.get_push_room_context(room, room_info, changes).await?;
 
-        for event in events {
-            let mut event: SyncTimelineEvent = event.into();
+        for raw_event in events {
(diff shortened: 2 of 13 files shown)


crates/matrix-sdk-ui/src/timeline/event_item/mod.rs
```diff
@@ -1023,7 +1023,7 @@ mod tests {
         formatted_body: &str,
         ts: u64,
     ) -> SyncTimelineEvent {
-        sync_timeline_event!({
+        SyncTimelineEvent::new(sync_timeline_event!({
```
bnjbvr, line 1026: Drive-by:constructing the `SyncTimelineEvent` should be part of the `sync_timeline_event!` macro

richvdh: should it? I think there are cases where we actually want a `Raw<AnySyncTimelineEvent>`

bnjbvr: It should. It's named `sync_timeline_event!` after all :)

(Long term, better, bnj-preferred fix, would be to [stop using `sync_timeline_event` and use the event factory more](https://github.com/matrix-org/matrix-rust-sdk/issues/3716), so as to consolidate testing utilities.)

richvdh: Ok so a couple of problems with this.

First, `sync_timeline_event` is in `matrix-sdk-test`, and is used by `EventBuilder`, which is also in that crate, whereas `SyncTimelineEvent` is in `matrix-sdk-common`. We're therefore in danger of circular dependencies.

I guess we could unroll the invocations of `sync_timeline_event` which do want a `Raw<AnySyncTimelineEvent>` (such as those in `EventBuilder`) to sidestep this, but it feels inelegant.

It would also mean that `sync_timeline_event` would have quite different behaviour to `timeline_event`, `sync_state_event`, and other similar macros. Possibly we could come up with a way to change all of them, but it feels like a lot of work, and as #3716 notes, we probably want to phase out these macros than refactor half the codebase to improve them.

Maybe we instead want a *different* macro that constructs a `SyncTimelineEvent` ? I'm not sure it's worth it just to avoid the `SyncTimelineEvent::new` invocation, though.

bnjbvr: oof, my bad, I thought that `sync_timeline_event!` did create a `SyncTimelineEvent` before this patch, and that this behavior had changed; please nevermind me.

bnjbvr (approved):
Thanks!

Issue #1732, opened 2023-04-03
[Media+Timeline] Local echo support for media attachments


stefanceriu:
Sending media attachments through the Timeline should make use of local echoes for a seamless experience.

---

## Sub project: handle media upload in the send queue

- ~~break apart uploading in a small state machine that can be reused in the send queue~~
  - Turns out this was not strictly required, because the code in the send queue and out are so different.
- [x] modify the send queue storage format so it takes care of "requests", not only sending "event content"
- [x] add a new request type for uploading a media
- [x] (draw the rest of the freaking howl)
- [x] handle medias local echoes in the timeline

## Tasks

- https://github.com/matrix-org/matrix-rust-sdk/pull/4140
- https://github.com/matrix-org/matrix-rust-sdk/pull/4149 (cache media and thumbnail)

## Related

- https://github.com/matrix-org/matrix-rust-sdk/issues/1968
- [element-meta task](https://github.com/element-hq/element-meta/issues/2549)

jplatte:
Synapse implements asyncronous uploads from [v1.97.0](https://github.com/matrix-org/synapse/releases/tag/v1.97.0) onwards, with which this should be easier to implement (can generate a transaction ID / get an event ID for it in parallel with the upload). Ruma also supports [the new endpoint](https://docs.rs/ruma/latest/ruma/api/client/media/create_content_async/index.html).

PR #4149, opened 2024-10-17
media: cache media's content after uploading it

bnjbvr:
This adds an optional way to cache a media's content after uploading it successfully. In a followup PR, I'll also add caching of the thumbnail, in a slightly smarter way than we do right now (always return a cache hit independently of the requested thumbnail features).

Part of #1732.

### DIFF
--- a/bindings/matrix-sdk-ffi/src/timeline/mod.rs
+++ b/bindings/matrix-sdk-ffi/src/timeline/mod.rs
@@ -104,12 +104,18 @@ impl Timeline {
         mime_type: Option<String>,
         attachment_config: AttachmentConfig,
         progress_watcher: Option<Box<dyn ProgressWatcher>>,
+        store_in_cache: bool,
     ) -> Result<(), RoomError> {
         let mime_str = mime_type.as_ref().ok_or(RoomError::InvalidAttachmentMimeType)?;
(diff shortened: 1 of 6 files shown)


andybalaam (approved):
Looks good, thanks.

bnjbvr:
I'm hesitant to merge this, because I think that for the local echo for media upload, we'll need a strategy where we'll first store the media in the cache, with a temporary MXC url, and then after it's been uploaded, replace the underlying temporary MXC url with the one returned from the server. We could use the same strategy here to avoid cloning the data we want to store in the cache, and that will be aligned with the local echo feature.

bnjbvr:
Re-requesting review because of non-trivial changes: I made it so that one can take the bytes by reference instead of by ownership, which required tweaking the `SendRequest` `IntoFuture` impl too.

bnjbvr:
We've discussed this in the Rust chapter, and we're going to keep it super simple (so, remove the last commit), and add caching of thumbnails as is. Drafting.

bnjbvr:
Done. Would still appreciate a re-review, since the thumbnail is now being cached too :) Thanks in advance!

andybalaam (approved):
Great, thanks

PR #4151, opened 2024-10-18
chore(ffi): Add optional `canonical_alias` field to `CreateRoomParameters`

jmartinesp:
We need this value to provide a room alias for public rooms.

Signed-off-by:


andybalaam (approved):
Yup, I like it :-)

PR #4152, opened 2024-10-18
chore(room_preview): add `RoomListItem::preview_room`

jmartinesp:
This method will return a `RoomPreview` for the provided room id. 

Also added `fn RoomPreview::leave()` action to be able to decline invites or cancel knocks, since there wasn't a `Client::leave_room_by_id` counterpart as there is for  join.

The PR also deprecates `RoomListItem::invited_room`, since we have a better alternative now.

Signed-off-by:

### DIFF
--- a/bindings/matrix-sdk-ffi/src/client.rs
+++ b/bindings/matrix-sdk-ffi/src/client.rs
@@ -2,7 +2,6 @@ use std::{
     collections::HashMap,
     fmt::Debug,
     mem::ManuallyDrop,
-    ops::Deref,
     path::Path,
     sync::{Arc, RwLock},
 };
@@ -1028,7 +1027,7 @@ impl Client {
         &self,
         room_id: String,
         via_servers: Vec<String>,
-    ) -> Result<RoomPreview, ClientError> {
+    ) -> Result<Arc<RoomPreview>, ClientError> {
         let room_id = RoomId::parse(&room_id).context("room_id is not a valid room id")?;
 
         let via_servers = via_servers
@@ -1042,16 +1041,15 @@ impl Client {
         let room_id: &RoomId = &room_id;
 
         let sdk_room_preview = self.inner.get_room_preview(room_id.into(), via_servers).await?;
-        let client = (*self.inner.deref()).clone();
 
-        RoomPreview::try_from_sdk(sdk_room_preview, client)
+        Ok(Arc::new(RoomPreview::from_sdk(sdk_room_preview)))
     }
 
     /// Given a room alias, get the preview of a room, to interact with it.
     pub async fn get_room_preview_from_room_alias(
         &self,
         room_alias: String,
-    ) -> Result<RoomPreview, ClientError> {
+    ) -> Result<Arc<RoomPreview>, ClientError> {
         let room_alias =
             RoomAliasId::parse(&room_alias).context("room_alias is not a valid room alias")?;
 
@@ -1060,9 +1058,8 @@ impl Client {
         let room_alias: &RoomAliasId = &room_alias;
 
         let sdk_room_preview = self.inner.get_room_preview(room_alias.into(), Vec::new()).await?;
-        let client = (*self.inner.deref()).clone();
 
-        RoomPreview::try_from_sdk(sdk_room_preview, client)
+        Ok(Arc::new(RoomPreview::from_sdk(sdk_room_preview)))
     }
 
     /// Waits until an at least partially synced room is received, and returns
@@ -1807,7 +1804,7 @@ impl From<OidcPrompt> for SdkOidcPrompt {
 }
 
 /// The rule used for users wishing to join this room.
-#[derive(uniffi::Enum)]
+#[derive(Debug, Clone, uniffi::Enum)]
 pub enum JoinRule {
     /// Anyone can join the room without any prior action.
     Public,
@@ -1836,7 +1833,7 @@ pub enum JoinRule {
 }
 
 /// An allow rule which defines a condition that allows joining a room.
-#[derive(uniffi::Enum)]
+#[derive(Debug, Clone, uniffi::Enum)]
 pub enum AllowRule {
     /// Only a member of the `room_id` Room can join the one this rule is used
     /// in.
--- a/bindings/matrix-sdk-ffi/src/client.rs
+++ b/bindings/matrix-sdk-ffi/src/client.rs
@@ -1040,9 +1040,9 @@ impl Client {
         // rustc win that one fight.
         let room_id: &RoomId = &room_id;
 
-        let sdk_room_preview = self.inner.get_room_preview(room_id.into(), via_servers).await?;
+        let room_preview = self.inner.get_room_preview(room_id.into(), via_servers).await?;
 
-        Ok(Arc::new(RoomPreview::from_sdk(sdk_room_preview)))
+        Ok(Arc::new(RoomPreview::new(self.inner.clone(), room_preview)))
     }
 
     /// Given a room alias, get the preview of a room, to interact with it.
@@ -1057,9 +1057,9 @@ impl Client {
         // rustc win that one fight.
         let room_alias: &RoomAliasId = &room_alias;
 
-        let sdk_room_preview = self.inner.get_room_preview(room_alias.into(), Vec::new()).await?;
+        let room_preview = self.inner.get_room_preview(room_alias.into(), Vec::new()).await?;
 
-        Ok(Arc::new(RoomPreview::from_sdk(sdk_room_preview)))
+        Ok(Arc::new(RoomPreview::new(self.inner.clone(), room_preview)))
     }
 
     /// Waits until an at least partially synced room is received, and returns
@@ -1830,6 +1830,12 @@ pub enum JoinRule {
     /// conditions described in a set of [`AllowRule`]s, or they can request
     /// an invite to the room.
     KnockRestricted { rules: Vec<AllowRule> },
+
+    /// A custom join rule, up for interpretation by the consumer.
+    Custom {
+        /// The string representation for this custom rule.
+        repr: String,
+    },
 }
 
 /// An allow rule which defines a condition that allows joining a room.
@@ -1857,6 +1863,7 @@ impl TryFrom<JoinRule> for ruma::events::room::join_rules::JoinRule {
                 let rules = allow_rules_from(rules)?;
                 Ok(Self::KnockRestricted(ruma::events::room::join_rules::Restricted::new(rules)))
             }
+            JoinRule::Custom { repr } => Ok(serde_json::from_str(&repr)?),
         }
     }
 }
--- a/bindings/matrix-sdk-ffi/src/client.rs
+++ b/bindings/matrix-sdk-ffi/src/client.rs
@@ -2,6 +2,7 @@ use std::{
     collections::HashMap,
     fmt::Debug,
     mem::ManuallyDrop,
+    ops::Deref,
     path::Path,
     sync::{Arc, RwLock},
 };
@@ -1041,8 +1042,9 @@ impl Client {
         let room_id: &RoomId = &room_id;
 
         let sdk_room_preview = self.inner.get_room_preview(room_id.into(), via_servers).await?;
+        let client = (*self.inner.deref()).clone();
 
-        Ok(RoomPreview::from_sdk(sdk_room_preview))
+        RoomPreview::try_from_sdk(sdk_room_preview, client)
     }
 
     /// Given a room alias, get the preview of a room, to interact with it.
@@ -1058,8 +1060,9 @@ impl Client {
         let room_alias: &RoomAliasId = &room_alias;
 
         let sdk_room_preview = self.inner.get_room_preview(room_alias.into(), Vec::new()).await?;
+        let client = (*self.inner.deref()).clone();
 
-        Ok(RoomPreview::from_sdk(sdk_room_preview))
+        RoomPreview::try_from_sdk(sdk_room_preview, client)
     }
 
     /// Waits until an at least partially synced room is received, and returns
--- a/bindings/matrix-sdk-ffi/src/lib.rs
+++ b/bindings/matrix-sdk-ffi/src/lib.rs
@@ -18,6 +18,8 @@ mod platform;
 mod room;
 mod room_directory_search;
 mod room_info;
+
+#[allow(deprecated)]
 mod room_list;
 mod room_member;
 mod room_preview;
--- a/bindings/matrix-sdk-ffi/src/lib.rs
+++ b/bindings/matrix-sdk-ffi/src/lib.rs
@@ -18,8 +18,6 @@ mod platform;
 mod room;
 mod room_directory_search;
 mod room_info;
-
-#[allow(deprecated)]
 mod room_list;
 mod room_member;
 mod room_preview;
--- a/bindings/matrix-sdk-ffi/src/room_list.rs
+++ b/bindings/matrix-sdk-ffi/src/room_list.rs
@@ -48,7 +48,7 @@ pub enum RoomListError {
     #[error("Event cache ran into an error: {error}")]
     EventCache { error: String },
     #[error("The requested room doesn't match the membership requirements {expected:?}, observed {actual:?}")]
-    IncorrectRoomMembership { expected: Membership, actual: Membership },
+    IncorrectRoomMembership { expected: Vec<Membership>, actual: Membership },
 }
 
 impl From<matrix_sdk_ui::room_list_service::Error> for RoomListError {
@@ -574,16 +574,32 @@ impl RoomListItem {
     }
 
     /// Builds a `Room` FFI from an invited room without initializing its
-    /// internal timeline
+    /// internal timeline.
     ///
-    /// An error will be returned if the room is a state different than invited
+    /// An error will be returned if the room is a state different than invited.
     ///
     /// ⚠️ Holding on to this room instance after it has been joined is not
-    /// safe. Use `full_room` instead
+    /// safe. Use `full_room` instead.
+    #[deprecated(note = "Please use `room_without_timeline` instead.")]
     fn invited_room(&self) -> Result<Arc<Room>, RoomListError> {
-        if !matches!(self.membership(), Membership::Invited) {
+        self.room_without_timeline()
+    }
+
+    /// Builds a `Room` FFI from a room without initializing its internal
+    /// timeline.
+    ///
+    /// An error will be returned if the room is a state other than invited
+    /// or knocked, the 2 states that would match this use case.
+    ///
+    /// ⚠️ Holding on to this room instance after it has been joined is not
+    /// safe. Use `full_room` instead.
+    fn room_without_timeline(&self) -> Result<Arc<Room>, RoomListError> {
+        let membership = self.membership();
+        if !matches!(membership, Membership::Invited)
+            && !matches!(self.membership(), Membership::Knocked)
+        {
             return Err(RoomListError::IncorrectRoomMembership {
-                expected: Membership::Invited,
+                expected: vec![Membership::Invited, Membership::Knocked],
                 actual: self.membership(),
             });
         }
@@ -598,7 +614,7 @@ impl RoomListItem {
     fn full_room(&self) -> Result<Arc<Room>, RoomListError> {
         if !matches!(self.membership(), Membership::Joined) {
             return Err(RoomListError::IncorrectRoomMembership {
-                expected: Membership::Joined,
+                expected: vec![Membership::Joined],
                 actual: self.membership(),
             });
         }
--- a/bindings/matrix-sdk-ffi/src/room_list.rs
+++ b/bindings/matrix-sdk-ffi/src/room_list.rs
@@ -1,4 +1,12 @@
-use std::{fmt::Debug, mem::MaybeUninit, ptr::addr_of_mut, sync::Arc, time::Duration};
+#![allow(deprecated)]
+
+use std::{
+    fmt::Debug,
+    mem::{ManuallyDrop, MaybeUninit},
+    ptr::addr_of_mut,
+    sync::Arc,
+    time::Duration,
+};
 
 use eyeball_im::VectorDiff;
 use futures_util::{pin_mut, StreamExt, TryFutureExt};
@@ -599,6 +607,13 @@ impl RoomListItem {
     /// An error will be returned if the room is in a state other than invited
     /// or knocked.
     async fn preview_room(&self, via: Vec<String>) -> Result<Arc<RoomPreview>, ClientError> {
+        // Validate parameters first.
+        let server_names: Vec<OwnedServerName> = via
+            .into_iter()
+            .map(|server| ServerName::parse(server).map_err(ClientError::from))
+            .collect::<Result<_, ClientError>>()?;
+
+        // Validate internal room state.
         let membership = self.membership();
         if !matches!(membership, Membership::Invited | Membership::Knocked) {
             return Err(RoomListError::IncorrectRoomMembership {
@@ -608,21 +623,19 @@ impl RoomListItem {
             .into());
         }
 
+        // Do the thing.
         let client = self.inner.client();
         let (room_or_alias_id, server_names) = if let Some(alias) = self.inner.canonical_alias() {
             let room_or_alias_id: OwnedRoomOrAliasId = alias.into();
             (room_or_alias_id, Vec::new())
         } else {
             let room_or_alias_id: OwnedRoomOrAliasId = self.inner.id().to_owned().into();
-            let server_names: Vec<OwnedServerName> = via
-                .into_iter()
-                .map(|server| ServerName::parse(server).map_err(ClientError::from))
-                .collect::<Result<_, ClientError>>()?;
             (room_or_alias_id, server_names)
         };
 
         let room_preview = client.get_room_preview(&room_or_alias_id, server_names).await?;
-        Ok(Arc::new(RoomPreview::from_sdk(room_preview)))
+
+        Ok(Arc::new(RoomPreview::new(ManuallyDrop::new(client), room_preview)))
     }
 
     /// Build a full `Room` FFI object, filling its associated timeline.
--- a/bindings/matrix-sdk-ffi/src/room_list.rs
+++ b/bindings/matrix-sdk-ffi/src/room_list.rs
@@ -16,12 +16,16 @@ use matrix_sdk_ui::{
     timeline::default_event_filter,
     unable_to_decrypt_hook::UtdHookManager,
 };
+use ruma::{
+    IdParseError, OwnedRoomOrAliasId, OwnedServerName, RoomAliasId, RoomOrAliasId, ServerName,
+};
 use tokio::sync::RwLock;
 
 use crate::{
     error::ClientError,
     room::{Membership, Room},
     room_info::RoomInfo,
+    room_preview::RoomPreview,
     timeline::{EventTimelineItem, Timeline},
     timeline_event_filter::TimelineEventTypeFilter,
     TaskHandle, RUNTIME,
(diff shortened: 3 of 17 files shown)


bindings/matrix-sdk-ffi/src/room_list.rs
```diff
@@ -584,16 +584,32 @@ impl RoomListItem {
...
+    }
+
+    /// Builds a `Room` FFI from a room without initializing its internal
+    /// timeline.
+    ///
+    /// An error will be returned if the room is a state other than invited
+    /// or knocked, the 2 states that would match this use case.
+    ///
+    /// ⚠️ Holding on to this room instance after it has been joined is not
+    /// safe. Use `full_room` instead.
```
bnjbvr, line 605: What do you mean by "not safe" here? This has quite a precise meaning in Rust, and I'm not sure if it's that, and I'm curious :)

bindings/matrix-sdk-ffi/src/room_list.rs
```diff
@@ -584,16 +584,32 @@ impl RoomListItem {
...
+    fn room_without_timeline(&self) -> Result<Arc<Room>, RoomListError> {
```
bnjbvr, line 606: Really not fond of this name, I see one of three ways to go from here (there can be more!):

1. Either rename this `preview_room()` or `stripped_room()`, as it's really a very basic stripped view of a room.
2. Rename `fn room()` to `fn room_with_timeline()`, and this one can be `fn room()` later.
3. (is kind of orthogonal, and could be in addition to (1) or (2)) It might be nice to return a `StrippedRoom` or `PreviewRoom` type instead of `Room`, from this function, to make it clear which functions are available and which are not. In fact, this might even get rid of the unused internal timeline instance.

What do you think?

bindings/matrix-sdk-ffi/src/lib.rs
```diff
@@ -18,6 +18,8 @@ mod platform;
 mod room;
 mod room_directory_search;
 mod room_info;
+
+#[allow(deprecated)]
 mod room_list;
```
bnjbvr, line 23: How many uses of the deprecated functions are there in this module? Can we add the guard on them one by one instead, or would that be too many? If too many, can we put the allow guard in the module itself, instead of here?

bnjbvr (changes requested):

bindings/matrix-sdk-ffi/src/room_list.rs
jmartinesp: This is copied from the `invited_room` fn, so we'd probably better ask @stefanceriu , but I think it means 'not safe' as in 'anything related to the timeline will fail'.

bindings/matrix-sdk-ffi/src/lib.rs
jmartinesp: When I tried using the allow guard in the function it didn't take effect (I still had the same compiler error message), I had to place it in the module for it to work. I suspect it has something to do with the FFI generated code, but I'm not sure.

bindings/matrix-sdk-ffi/src/room_list.rs
jmartinesp: `preview_room` sounds fine to me. As for having a separate struct backing those, maybe we can create one, once we know all the values we need to expose through the FFI. I can try to get this working.

bindings/matrix-sdk-ffi/src/lib.rs
bnjbvr: Ah yeah, expanding the macro could tell us whether it's possible to put it on a slightly more specialized location, e.g. the `impl` block itself, or if it's using it in code generated elsewhere.

bindings/matrix-sdk-ffi/src/room_list.rs
```diff
@@ -584,16 +584,32 @@ impl RoomListItem {
...
+        let membership = self.membership();
+        if !matches!(membership, Membership::Invited)
+            && !matches!(self.membership(), Membership::Knocked)
```
Hywan, line 609: It can be a single `matches!`.

```suggestion
        if !matches(membership, Membership::Invited | Membership::Knocked)
```

bindings/matrix-sdk-ffi/src/lib.rs
jmartinesp: Adding it to the impl block fails too. I've only been able to fix it by adding it in the `mod` declaration.

jmartinesp:
I ended up creating a `RoomPreviewActions` struct in the FFI layer so when you request a `RoomListItem::preview_room` you get a `RoomPreview` record that contains `RoomPreviewInfo` (also a record) and `RoomPreviewActions`, which are either `join`, `leave` or `knock` and check the membership state of the room preview before running. So from a client, you'd do something like:

```kotlin
val preview = roomListItem.previewRoom()
if (preview.info.membership == Membership.Invite) {
    preview.actions.join()
}
```

bindings/matrix-sdk-ffi/src/lib.rs
jmartinesp: When I add it to the function, the expanded macro looks like this:

```rust
#[doc = " Builds a `Room` FFI from an invited room without initializing its"]
#[doc = " internal timeline."]
#[doc = ""]
#[doc = " An error will be returned if the room is a state different than invited."]
#[doc = ""]
#[doc = " ⚠\u{65039} Holding on to this room instance after it has been joined is not"]
#[doc = " safe. Use `full_room` instead."]
#[allow(deprecated)]
#[deprecated(note = "Please use `preview_room` instead.")]
fn invited_room(&self) -> Result<Arc<Room>, RoomListError> {
    if !match (self.membership()) {
        Membership::Invited => true,
        _ => false
    } {
        return Err(RoomListError::IncorrectRoomMembership {
            expected: (<[_]>::into_vec(
                #[rustc_box]
                ::alloc::boxed::Box::new([Membership::Invited])
            )),
            actual: self.membership(),
        });
    }
    Ok(Arc::new(Room::new(self.inner.inner_room().clone())))
}
```

And if I add it to the impl block it looks like:

```rust
#[uniffi::export(async_runtime = "tokio", )]
#[allow(deprecated)]
impl RoomListItem {
    ...
```

But clippy still complains in both cases.

crates/matrix-sdk/src/room_preview.rs
```diff
@@ -219,6 +250,105 @@ impl RoomPreview {
 
         let state = client.get_room(room_id).map(|room| room.state());
 
-        Ok(Self::from_room_info(room_info, num_joined_members, state))
+        Ok(Self::from_room_info(client, room_info, num_joined_members, state).await)
     }
 }
+
+pub enum RoomPreviewActions {
```
bnjbvr: In this entire file you need to comment each struct, enum, variant, field, function. I'll enable a new lint for this :smiling_imp:

crates/matrix-sdk/src/room_preview.rs
```diff
@@ -219,6 +250,105 @@ impl RoomPreview {
...
+#[derive(thiserror::Error, Debug)]
+pub enum RoomPreviewError {
+    /// Attempted to call a method on a room preview that requires the user to
+    /// have a specific membership state, but the membership state is
+    /// different.
+    #[error("wrong room state: {0}")]
+    WrongRoomPreviewState(WrongRoomPreviewState),
+
+    #[error(transparent)]
+    Generic(#[from] Box<dyn std::error::Error + Send + Sync>)
```
bnjbvr: I really don't like this; having an exceptional generic error type like that should be limited at the strict minimum, i.e. `matrix_sdk::Error` only. Let's avoid it entirely by not making a custom error, but instead use an `matrix_sdk::Error` instead of this error type, there's already `WrongRoomState` which describes the same error.

bindings/matrix-sdk-ffi/src/room_list.rs
```diff
@@ -580,31 +584,47 @@ impl RoomListItem {
...
+                actual: self.membership(),
+            });
+        }
+        Ok(Arc::new(Room::new(self.inner.inner_room().clone())))
     }
 
-    /// Builds a `Room` FFI from a room without initializing its internal
-    /// timeline.
+    /// Builds a `RoomPreview` from a room list item. This is intended for rooms
+    /// with [`Membership::Invite`] or [`Membership::Knocked`],
```
bnjbvr, line 597: nit: this sentence ends with a comma, and it doesn't really translate to proper English IMO: maybe `This is intended for invited or knocked rooms.`?

bindings/matrix-sdk-ffi/src/room_preview.rs
```diff
@@ -1,9 +1,87 @@
...
+        match actions {
+            RoomPreviewActions::Invited { join, leave } => Self::Invited { join, leave },
+            RoomPreviewActions::Knocked { leave } => Self::Knocked { leave },
+        }
+    }
+}
+
+struct
+
+impl RoomPreview {
```
bnjbvr: Seems like this is not an atomic commit… (It's fine but next time to make the review easier, feel free to squash them altogether, since I know that another commit will now change these three lines.)

bindings/matrix-sdk-ffi/src/room_preview.rs
```diff
@@ -1,9 +1,87 @@
...
+    Client, RoomState,
+};
 use ruma::space::SpaceRoomJoinRule;
 
-/// The preview of a room, be it invited/joined/left, or not.
+use crate::{client::JoinRule, error::ClientError, room::Membership};
+
 #[derive(uniffi::Record)]
 pub struct RoomPreview {
+    pub info: RoomPreviewInfo,
```
bnjbvr, line 15: Why do we need the intermediary `RoomPreviewInfo` data structure?

bindings/matrix-sdk-ffi/src/room_preview.rs
```diff
@@ -1,9 +1,87 @@
...
+
+#[derive(uniffi::Object)]
+pub struct JoinRoomPreviewAction {
+    inner: SdkJoinRoomPreviewAction,
+}
+
+#[matrix_sdk_ffi_macros::export]
+impl JoinRoomPreviewAction {
+    async fn run(&self) {
+        self.inner.run().unwrap()
```
bnjbvr: Let's not `unwrap()` at the FFI layer, this will cause unnecessary panics that we need to handle at the last minute.

bindings/matrix-sdk-ffi/src/room_preview.rs
```diff
@@ -1,9 +1,87 @@
...
+        let info = RoomPreviewInfo {
+            room_id: info.room_id.to_string(),
+            canonical_alias: info.canonical_alias.map(|alias| alias.to_string()),
+            name: info.name,
+            topic: info.topic,
+            avatar_url: info.avatar_url.map(|url| url.to_string()),
+            num_joined_members: info.num_joined_members,
+            room_type: info.room_type.map(|room_type| room_type.to_string()),
+            is_history_world_readable: info.is_world_readable,
+            membership: info.state.map(|state| state.into()).unwrap_or_else(|| Membership::Left),
```
bnjbvr: nit: `unwrap_or` is sufficient; `unwrap_or_else` is to be used when some computation needs to happen in the closure.

Why can we assume `Left` in the first place, though?

crates/matrix-sdk/src/room_preview.rs
```diff
@@ -69,18 +77,27 @@ pub struct RoomPreview {
...
 }
 
 impl RoomPreview {
     /// Constructs a [`RoomPreview`] from the associated room info.
     ///
     /// Note: not using the room info's state/count of joined members, because
     /// we can do better than that.
-    fn from_room_info(
+    async fn from_room_info(
+        client: &Client,
```
bnjbvr, line 87: nit: since both callers have a handle to the `room` from the Client, can you pass `is_direct: bool` here instead?

bindings/matrix-sdk-ffi/src/room_list.rs
```diff
@@ -624,7 +622,7 @@ impl RoomListItem {
         };
 
         let room_preview = client.get_room_preview(&room_or_alias_id, server_names).await?;
-        Ok(Arc::new(RoomPreview::try_from_sdk(room_preview, client)))
+        RoomPreview::try_from_sdk(room_preview, client)
```
bnjbvr, line 625: This fixes stuff from the previous commit, right? :pensive:

crates/matrix-sdk/src/room_preview.rs
```diff
@@ -254,101 +250,181 @@ impl RoomPreview {
...
-impl JoinRoomPreviewAction {
-    pub fn new(client: Client) -> Self {
-        Self { client }
+impl RoomPreviewActions {
+    /// Create a new instance given a [`Client`] to perform the actions and a
+    /// [`RoomPreview`] to get the needed data.
+    pub fn new(client: Client, room_preview: RoomPreview) -> Self {
+        let room_id = room_preview.room_id;
+        let room_state = room_preview.state;
+        let is_direct = room_preview.is_direct.unwrap_or_default();
```
bnjbvr, line 280: That's a lie, let's keep an `Option<bool>` for `is_direct` ;)

crates/matrix-sdk/src/room_preview.rs
```diff
@@ -254,101 +250,181 @@ impl RoomPreview {
     }
 }
 
-pub enum RoomPreviewActions {
-    Invited { join: JoinRoomPreviewAction, leave: LeaveRoomPreviewAction },
-    Knocked { leave: LeaveRoomPreviewAction },
+/// A set of actions that can be performed in a [`RoomPreview`].
+#[derive(Debug)]
+pub struct RoomPreviewActions {
```
bnjbvr, line 255: Overall, I don't think this data structure and the other actions are required at the SDK level; maybe they're required at the FFI layer because we need some kind of "lazy provider"-like object for the Record requirements, but I don't think they make sense for the `RoomPreview` of the SDK, considering there are many other methods that already exist providing the same functionality, so removing it would avoid blowing up the number of abstractions, code duplication, etc.

bnjbvr (changes requested):
I think we don't need all the preview actions data structure at the SDK level; these need to exist at the FFI layer because of… reasons (that I'm detailed below; let me know if that's wrong). In the current shape I am strongly not in favor of merging this, let's identify what's the minimal set of changes that we could have to unblock us, and maybe consider doing some refactoring later.

crates/matrix-sdk/src/room_preview.rs
jmartinesp: Sure, but I'd still have to call `room.is_direct().await.unwrap_or_default()`, is it fine to assume it's not direct at this point?

bindings/matrix-sdk-ffi/src/room_preview.rs
```diff
@@ -21,33 +99,22 @@ pub struct RoomPreview {
...
+    fn from(join_rule: SpaceRoomJoinRule) -> Self {
+        match join_rule {
+            SpaceRoomJoinRule::Public => JoinRule::Public,
+            SpaceRoomJoinRule::Private => JoinRule::Private,
+            SpaceRoomJoinRule::Invite => JoinRule::Invite,
+            SpaceRoomJoinRule::Knock => JoinRule::Knock,
+            SpaceRoomJoinRule::KnockRestricted => JoinRule::KnockRestricted { rules: Vec::new() },
+            SpaceRoomJoinRule::Restricted => JoinRule::Restricted { rules: Vec::new() },
+            // For the `_Custom` case, assume it's private
+            _ => JoinRule::Private,
```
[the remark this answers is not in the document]
jmartinesp: I'll add a `Custom(String)` variant, although I guess in the clients we'll just need to assume it's private anyway.

bindings/matrix-sdk-ffi/src/room_preview.rs
jmartinesp: That's what `RoomPreview` does at the SDK level [here](https://github.com/matrix-org/matrix-rust-sdk/blob/b6e9d4f6bdb35b4f5c5320c63133b0a33d278c21/crates/matrix-sdk/src/room_preview.rs#L114) for join rules, so I assumed we could do the same. I'll change it to `Option<Membership>` instead.

bindings/matrix-sdk-ffi/src/room_preview.rs
jmartinesp: So we didn't need to move everything inside `RoomPreview` into a `Record` if we made it an `Object` (as is the case now since I'm removing the separate `actions` part).

bindings/matrix-sdk-ffi/src/room_preview.rs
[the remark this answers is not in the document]
jmartinesp: We don't have a way to map the `PrivOwnedStr` in `_Custom` to a `String` AFAICT, or to create a `PrivOwnedStr` from a String either. I think this is why the mappers just assumed the `Private` usage.

crates/matrix-sdk/src/room_preview.rs
jmartinesp: I'd still get an `Option<bool>` in some cases. Maybe I'll just pass the nullable `is_direct` field to the clients and let them decide what to do with it.

crates/matrix-sdk/src/room_preview.rs
jmartinesp: I added it to the SDK so they could be tested and didn't have to duplicate code, i.e. the `Room::join` contents mentioned below, which could be just put into one of these actions and then reused inside `Room::join` instead:

```rust
impl Room {
    async fn join(&self) -> Result<_, _> {
        self.actions.join()
   }
```

But if that's not something we want, we can just remove the actions, leaving a needed `RoomPreview::leave` method that's needed (`Client` has code for joining and knocking, but it doesn't have one for leaving a room given its id, so a `RoomPreview` with invite/knock state can never be left like that).

bindings/matrix-sdk-ffi/src/room_list.rs
jmartinesp: Resolving this as the doc comment doesn't have the 'not safe' part anymore.

crates/matrix-sdk/src/room_preview.rs
jmartinesp: This was fixed in the following commit if I'm not mistaken.

crates/matrix-sdk/src/room_preview.rs
jmartinesp: I'll still need to pass a `Client` around since I need it to be able to call `RoomPreview::leave()`, as there's no other way to leave an invited/knocked room if we change those to return a `RoomPreview` instead. Should I still calculate the `is_direct` value outside, if the client can do it inside with the latest info?

jmartinesp:
- I removed the `Room*Actions` in https://github.com/matrix-org/matrix-rust-sdk/pull/4152/commits/0fcfd84311e2047825763aeae41cc210d3007041.
- I fixed the `RoomPreview` contents and added a `fn RoomPreview::leave()` that was needed in https://github.com/matrix-org/matrix-rust-sdk/pull/4152/commits/7bfb1bd1c4d3bd65f15b132330f3dbc9189063f0.
- The doc comment using direct references to types was fixed in https://github.com/matrix-org/matrix-rust-sdk/pull/4152/commits/f8110d24e28b12e6f7a31fe14e1cad2551da6165.
- I tried to explain why the `JoinRule::_Custom` case can't be handled and we default to `Private` in https://github.com/matrix-org/matrix-rust-sdk/pull/4152/commits/69ee5c2e4defb4ad3b6400f2fa7246d6ae4ad14d.

crates/matrix-sdk/src/room_preview.rs
jmartinesp: Removed in https://github.com/matrix-org/matrix-rust-sdk/commit/0fcfd84311e2047825763aeae41cc210d3007041.

bindings/matrix-sdk-ffi/src/room_preview.rs
jmartinesp: Removed in https://github.com/matrix-org/matrix-rust-sdk/pull/4152/commits/0fcfd84311e2047825763aeae41cc210d3007041.

bindings/matrix-sdk-ffi/src/room_preview.rs
jmartinesp: Handled in https://github.com/matrix-org/matrix-rust-sdk/commit/7bfb1bd1c4d3bd65f15b132330f3dbc9189063f0.

crates/matrix-sdk/src/room_preview.rs
jmartinesp: This code was removed.

crates/matrix-sdk/src/client/mod.rs
```diff
@@ -1208,6 +1208,13 @@ impl Client {
         Ok(Room::new(self.clone(), base_room))
     }
 
+    /// Leave a room given its `RoomId`.
```
bnjbvr, line 1211: We don't need this method, it's the same as `Room::leave()`, and we can only leave a room we do know about (i.e. we can get it with `Client::get_room` first).

crates/matrix-sdk/src/room_preview.rs
```diff
@@ -69,19 +69,46 @@ pub struct RoomPreview {
...
+        match self.state {
+            Some(RoomState::Left) | None => {
+                Err(Error::WrongRoomPreviewState(WrongRoomPreviewState {
+                    expected: "Invited, Knocked or Joined",
+                    got: self.state,
+                }))
+            }
+            Some(_) => self.client.leave_room(&self.room_id).await.map_err(Into::into),
+        }
+    }
```
bnjbvr, line 93: Sorry, I still think that `RoomPreview` ought to be a purely descriptive data record (i.e. no action methods); can the caller use the room id that's always around, and call `leave_room` by themselves? Then, can we get rid of the `Client` stored and address the other comment I had on `from_room_info`?

bindings/matrix-sdk-ffi/src/room_preview.rs
```diff
@@ -20,34 +65,28 @@ pub struct RoomPreview {
...
+            SpaceRoomJoinRule::Private => JoinRule::Private,
+            SpaceRoomJoinRule::Restricted => JoinRule::Restricted { rules: Vec::new() },
+            SpaceRoomJoinRule::KnockRestricted => JoinRule::KnockRestricted { rules: Vec::new() },
+            SpaceRoomJoinRule::Public => JoinRule::Public,
+            _ => {
+                // Since we have no way to get the _Custom contents, assume it's private.
+                // Warning! If new join rules are introduced we may be mapping wrong values
+                // here, but there's no way to match
+                // `SpaceRoomJoinRule::_Custom(_)` and have an exhaustive match.
+                JoinRule::Private
```
bnjbvr, line 88: I'd do it differently:

- for custom, add a `JoinRule::Custom(String)` variant that's built with `join_rule.to_string()` (`SpaceRoomJoinRule` implements `StringEnum` which derives a `Display` impl)?
- for `_`, log a warning and return `None`. Yes, let's turn this into `TryInto` :innocent:

bindings/matrix-sdk-ffi/src/client.rs
```diff
@@ -1027,7 +1027,7 @@ impl Client {
         &self,
         room_id: String,
         via_servers: Vec<String>,
-    ) -> Result<RoomPreview, ClientError> {
+    ) -> Result<Arc<RoomPreview>, ClientError> {
```
bnjbvr, line 1030: Do we still need all the `Arc`s here? (I'm not very sure about it myself, but I suspect we don't since `RoomPreview` is now a uniffi object)

bindings/matrix-sdk-ffi/src/room_list.rs
```diff
@@ -574,31 +576,63 @@ impl RoomListItem {
...
+        let client = self.inner.client();
+        let (room_or_alias_id, server_names) = if let Some(alias) = self.inner.canonical_alias() {
+            let room_or_alias_id: OwnedRoomOrAliasId = alias.into();
+            (room_or_alias_id, Vec::new())
+        } else {
+            let room_or_alias_id: OwnedRoomOrAliasId = self.inner.id().to_owned().into();
+            let server_names: Vec<OwnedServerName> = via
+                .into_iter()
+                .map(|server| ServerName::parse(server).map_err(ClientError::from))
+                .collect::<Result<_, ClientError>>()?;
```
bnjbvr, line 620: It's a bit weird that if there's a canonical alias, the `via` routes will not be validated :thinking: Should we validate them ahead, even if we're discarding them before?

crates/matrix-sdk/tests/integration/room_preview.rs
```diff
@@ -0,0 +1,139 @@
...
+    let room_preview = client.get_room_preview(room_id.into(), Vec::new()).await.unwrap();
+    assert_eq!(room_preview.state.unwrap(), RoomState::Invited);
+
+    room_preview.leave().await.unwrap();
+
+    assert_eq!(client.get_room(room_id).unwrap().state(), RoomState::Left);
+}
+
+#[async_test]
+async fn test_room_preview_leave_knocked() {
```
bnjbvr, line 39: With the suggestions I've made above, I think that most of these tests could go away. Unless we didn't have tests to leave a knocked room, maybe?

bindings/matrix-sdk-ffi/src/lib.rs
bnjbvr: Yeah, it's the call sites that need the `allow(deprecated)` guard, not the function itself; but it might be that it's in code generated by uniffi, that's outside our control, so we might need to resort to the mod-level guard indeed.

bnjbvr (changes requested):
Sorry, it's still too many unneeded changes for me. Let me try to hack on this, see if I can get it into the shape I'm expecting.

crates/matrix-sdk/src/client/mod.rs
jmartinesp: I think we're going in circles... We don't have anything in the `ffi::Client` to get a `Room`, and even if we had it, it would be a 'not safe to use' Room as we had in `RoomListItem::invited_room` (see https://github.com/matrix-org/matrix-rust-sdk/pull/4152#discussion_r1808445972). That's the main reason I split Room and RoomPreview here, and I think it's what we agreed to do.

So now I'm not sure what to do anymore:
- If we're ok with returning a Room with no timeline as we did before then I'll just remove everything from `RoomPreview` and just return that Room in `RoomListItem::preview_room` instead.
- If we're not ok with that, then we need a way to be able to leave that room (aka decline the invite, cancel the knock).

bindings/matrix-sdk-ffi/src/room_preview.rs
```diff
@@ -20,34 +89,32 @@ pub struct RoomPreview {
...
+            SpaceRoomJoinRule::Private => JoinRule::Private,
+            SpaceRoomJoinRule::Restricted => JoinRule::Restricted { rules: Vec::new() },
+            SpaceRoomJoinRule::KnockRestricted => JoinRule::KnockRestricted { rules: Vec::new() },
+            SpaceRoomJoinRule::Public => JoinRule::Public,
+            SpaceRoomJoinRule::_Custom(_) => JoinRule::Custom { repr: join_rule.to_string() },
+            _ => {
+                // Since we have no way to get the _Custom contents, assume it's private.
+                // Warning! If new join rules are introduced we may be mapping wrong values
+                // here, but there's no way to match
+                // `SpaceRoomJoinRule::_Custom(_)` and have an exhaustive match.
```
jmartinesp, line 114: Maybe remove/fix this comment?

bnjbvr (approved):

bindings/matrix-sdk-ffi/src/lib.rs
bnjbvr: Forgot to comment here: it's indeed an unmangled free-function generated in the same module, that's calling the `invited_room` method, so welp :shrug: Module-level guard it is. I wondered if we could get rid of the entire `invited_room` method altogether, but I don't want to add more API changes to the other side of this PR.

bnjbvr:
Jorge confirmed this was good enough for the EX apps, so merging. Thanks @jmartinesp for all the work done here, and sorry about the few back-and-forths.

PR #4153, opened 2024-10-18
Session verification incoming request support

### LINKED ISSUE #3595: Receiving a verification request does not work
As per https://github.com/matrix-org/matrix-rust-sdk/blob/main/bindings/matrix-sdk-ffi/src/session_verification.rs#L153 - you cannot be on the receiving end of a verification request currently.

This means you can't verify your EX device if you only have EX clients.

stefanceriu:
This PR introduced FFI level API support for interacting with incoming session verification requests. Best reviewed commit by commit
- it exposes a new `did_receive_verification_request` `SessionVerificationController` delegate method that includes information about the incoming request/device/sender
- the client can `acknowledge_verification_request` to get informed about it getting accepted somewhere else or cancelled
- the client can also `accept_verification_request` to start interacting with it

This PR also internally switches to the `VerificationRequest::changes` publisher, fixes a couple of invalid VerificationMachine tests and starts storing requesting `DeviceData` directly in `VerificationRequestState. Requested`

Relates to element-hq/element-meta/issues/2464, fixes #3595.

### DIFF
--- a/bindings/matrix-sdk-ffi/src/session_verification.rs
+++ b/bindings/matrix-sdk-ffi/src/session_verification.rs
@@ -1,6 +1,5 @@
 use std::sync::{Arc, RwLock};
 
-use anyhow::Context as _;
 use futures_util::StreamExt;
 use matrix_sdk::{
     encryption::{
@@ -10,6 +9,8 @@ use matrix_sdk::{
     },
     ruma::events::{key::verification::VerificationMethod, AnyToDeviceEvent},
 };
+use ruma::UserId;
+use tracing::{error, info};
 
 use super::RUNTIME;
 use crate::error::ClientError;
@@ -39,6 +40,7 @@ pub enum SessionVerificationData {
 
 #[matrix_sdk_ffi_macros::export(callback_interface)]
 pub trait SessionVerificationControllerDelegate: Sync + Send {
+    fn did_receive_verification_request(&self, sender_id: String, flow_id: String);
     fn did_accept_verification_request(&self);
     fn did_start_sas_verification(&self);
     fn did_receive_verification_data(&self, data: SessionVerificationData);
@@ -60,17 +62,43 @@ pub struct SessionVerificationController {
 
 #[matrix_sdk_ffi_macros::export]
 impl SessionVerificationController {
-    pub async fn is_verified(&self) -> Result<bool, ClientError> {
-        let device =
-            self.encryption.get_own_device().await?.context("Our own device is missing")?;
-
-        Ok(device.is_cross_signed_by_owner())
-    }
-
     pub fn set_delegate(&self, delegate: Option<Box<dyn SessionVerificationControllerDelegate>>) {
         *self.delegate.write().unwrap() = delegate;
     }
 
+    pub async fn accept_verification_request(
+        &self,
+        sender_id: String,
+        flow_id: String,
+    ) -> Result<(), ClientError> {
+        let sender_id = UserId::parse(sender_id.clone())?;
+
+        let verification_request = self
+            .encryption
+            .get_verification_request(&sender_id, flow_id)
+            .await
+            .ok_or(ClientError::new("Unknown session verification request"))?;
+
+        verification_request.accept().await?;
+
+        *self.verification_request.write().unwrap() = Some(verification_request);
+
+        Ok(())
+    }
+
+    /// Accept the previously acknowledged verification request
+    pub async fn accept_verification_request(&self) -> Result<(), ClientError> {
+        let verification_request = self.verification_request.read().unwrap().clone();
+
+        if let Some(verification_request) = verification_request {
+            let methods = vec![VerificationMethod::SasV1];
+            verification_request.accept_with_methods(methods).await?;
+        }
+
+        Ok(())
+    }
+
+    /// Request verification for the current device
     pub async fn request_verification(&self) -> Result<(), ClientError> {
         let methods = vec![VerificationMethod::SasV1];
         let verification_request = self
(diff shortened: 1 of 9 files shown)


bindings/matrix-sdk-ffi/src/session_verification.rs
```diff
@@ -60,80 +73,128 @@ pub struct SessionVerificationController {
 
 #[matrix_sdk_ffi_macros::export]
 impl SessionVerificationController {
-    pub async fn is_verified(&self) -> Result<bool, ClientError> {
-        let device =
-            self.encryption.get_own_device().await?.context("Our own device is missing")?;
+    pub fn set_delegate(&self, delegate: Option<Box<dyn SessionVerificationControllerDelegate>>) {
```
stefanceriu, line 76: I considered switching away from a delegate and towards a listener but the public api changes are so small that I don't really thing it's worth it at this point. Let's reassess when we start thinking about user verification.

crates/matrix-sdk-crypto/src/verification/requests.rs
```diff
@@ -78,9 +78,9 @@ pub enum VerificationRequestState {
         /// The verification methods supported by the sender.
         their_methods: Vec<VerificationMethod>,
 
-        /// The device ID of the device that responded to the verification
+        /// The device data of the device that responded to the verification
         /// request.
-        other_device_id: OwnedDeviceId,
+        other_device_data: DeviceData,
```
stefanceriu, line 83: @poljar We initially agreed to pass this `DeviceData` down into the `Transitioned` state but I couldn't identify were that would be useful 🤔

stefanceriu:
Please double check the commit messages, I tried making them fit the new contributing guidelines but not sure I nailed it.

bindings/matrix-sdk-ffi/src/session_verification.rs
```diff
@@ -60,80 +73,130 @@ pub struct SessionVerificationController {
...
+    }
+
+    /// Set this particular request as the currently active one and register for
+    /// events pertaining it.
+    /// * `sender_id` - The user requesting verification.
+    /// * `flow_id` - - The ID that uniquely identifies the verification flow.
+    pub async fn acknowledge_verification_request(
+        &self,
+        sender_id: String,
+        flow_id: String,
```
bmarty, line 87: Just a quick question regarding API usage: the `flow_id` is not enough here? We need to also provide the `sender_id`? (I believe that those data are coming from `SessionVerificationRequestDetails`).

stefanceriu: That's a good question, one I had myself. It's based on the underlying `OlmMachine::get_verification_request` which requires both, so there must be a good reason or it might be just because (maybe indexing by user ID so we can cancel existing requests for the same user https://github.com/matrix-org/matrix-rust-sdk/blob/main/crates/matrix-sdk-crypto/src/verification/machine.rs#L168-L169)

bmarty: OK, thanks. That's fine, I was just wondering :)

bindings/matrix-sdk-ffi/src/session_verification.rs
```diff
@@ -60,80 +73,130 @@ pub struct SessionVerificationController {
 
 #[matrix_sdk_ffi_macros::export]
 impl SessionVerificationController {
-    pub async fn is_verified(&self) -> Result<bool, ClientError> {
```
bmarty, line 63: Was this method removed by mistake @stefanceriu ? We are using it in EXA.

stefanceriu: Oh, it was intentional, you should be using `client.encryption().verificationState()` 🤔

bmarty: OK, that's what I have done, thanks for confirming!

crates/matrix-sdk-crypto/src/verification/requests.rs
poljar: Well that's because, when we go from the `Requested` state into the `Ready` state you throw away the `DeviceData` and the `Ready` state again only remembers the device ID.

You should propagate the `DeviceData` to the `Ready` state here:

https://github.com/matrix-org/matrix-rust-sdk/blob/1d4b34f7c92d389d9ac5ed59377822603848c764/crates/matrix-sdk-crypto/src/verification/requests.rs#L915-L916

And then further to the `Transitioned` state, this would allow you to get rid of the second `get_device_call()` we have:

https://github.com/matrix-org/matrix-rust-sdk/blob/1d4b34f7c92d389d9ac5ed59377822603848c764/crates/matrix-sdk-crypto/src/verification/requests.rs#L1354-L1362

poljar (changes requested):
Most of this look great, thanks for handling this. It is a bit unfinished though.

crates/matrix-sdk-crypto/src/verification/requests.rs
stefanceriu: Nice, thank you for the very clear explanation, it should be good to go now! I've rebased the whole PR to help app developers but only the last 3 commits are new.

richvdh:
Taking myself off the reviewer list here because it looks like @poljar already has context on it and will do a better job than me anyway

poljar (approved):
Very nice. Thanks for taking care of this.

poljar:
Oh about the commits, some of them you got right some of them not.

`fix(crypto): ...` is correct, `crypto(verification)` is not. In the second case you should have used `fix(verification)` or something along those lines.

The format is documented here: https://www.conventionalcommits.org/en/v1.0.0/.

PR #4155, opened 2024-10-21
[WIP] WidgetDriver: add to_device support

toger5:
This is very WIP and should not yet be reviewed. Opening the PR to track it related EPICs.

Signed-off-by:


toger5:
There is also a strategic shift that we will focus on shipping a new version of Element Call with all the stability fixes before we tackle to device key sending for element call. (Where we then need this change)

poljar:
> There is also a strategic shift that we will focus on shipping a new version of Element Call with all the stability fixes before we tackle to device key sending for element call. (Where we then need this change)

Alright, can we close this in the meantime then?

poljar:
Closing due to inactivity.

PR #4158, opened 2024-10-22
Tidy the IdentityStatusChanges test data

andybalaam:
A bit less repeating ourselves in this test data. Ulterior motive: if we ever want to add more users it will be easier now we can see exactly what information is needed to create a user.…ally

### DIFF
--- a/testing/matrix-sdk-test/src/test_json/keys_query_sets.rs
+++ b/testing/matrix-sdk-test/src/test_json/keys_query_sets.rs
@@ -4,7 +4,11 @@ use ruma::{
 };
 use serde_json::{json, Value};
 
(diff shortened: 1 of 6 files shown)


testing/matrix-sdk-test/src/test_json/keys_query_sets.rs
```diff
@@ -472,11 +472,11 @@ impl IdentityChangeDataSet {
         KeysQueryUser::bob_c().device_id
     }
 
-    pub fn msk_a() -> Value {
+    pub fn master_signing_keys_a() -> Value {
```
Hywan, line 475: :+1:

Hywan (approved):
LGTM, thanks!

PR #4159, opened 2024-10-22
feat(ui): Revisit `required_state` for `RoomListService`

Hywan:
This patch updates `matrix_sdk_ui::room_list_service`, and is twofold:

First off, it adds `m.call.member`, `m.room.topic` and `m.room.pinned_events` to
`required_state` of the `all_rooms` sliding sync list.

Second, this `required_state` sliding sync list value is easily out-of-sync with
the `required_state` sliding sync room subscription value. Thus, this patch
removes the `settings` argument of `RoomListService::subscribe_to_rooms` and
computes its own `RoomSubscription` instance. For `required_state`, it uses the
same `required_state` used by the `all_rooms` list, plus the `m.room.create`
state event. For `timeline_limit`, it uses 20 by default (this is not the
default value we were using in our tests, but that's the default value most
Matrix clients use).

### DIFF
--- a/crates/matrix-sdk-ui/src/room_list_service/mod.rs
+++ b/crates/matrix-sdk-ui/src/room_list_service/mod.rs
@@ -57,7 +57,7 @@ mod room_list;
 pub mod sorters;
 mod state;
 
-use std::{iter::once, sync::Arc, time::Duration};
+use std::{sync::Arc, time::Duration};
 
 use async_stream::stream;
 use eyeball::Subscriber;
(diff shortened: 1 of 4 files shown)


crates/matrix-sdk-ui/src/room_list_service/mod.rs
```diff
@@ -151,8 +151,8 @@ impl RoomListService {
                         (StateEventType::RoomMember, "$ME".to_owned()),
                         (StateEventType::RoomName, "".to_owned()),
                         (StateEventType::RoomCanonicalAlias, "".to_owned()),
-                        (StateEventType::RoomAvatar, "".to_owned()),
```
stefanceriu, line 154: Best leave an explanation for this

stefanceriu (approved):
Talked about most of this offline, looks good to me and works well in EXI

crates/matrix-sdk-ui/src/room_list_service/mod.rs
Hywan: I cannot add a comment in the code, but I can comment here.

Sliding sync already supplies a `response.avatar` value, so we don't need to request the `m.room.avatar` state event.

PR #4160, opened 2024-10-22
chore(media): use the same media method when caching a thumbnail as the FFI

bnjbvr:
The FFI will request a scaled version of the thumbnail by default; let's use the same cache key when caching the thumbnail after an upload.

Thanks @zecakeh for flagging the issue.

Part of #1732.


Hywan (approved):
LGTM!

PR #4166, opened 2024-10-24
sliding_sync(state): use `*` for `m.call.member` when requesting state through sliding sync

stefanceriu:
- introduced in https://github.com/matrix-org/matrix-rust-sdk/pull/4159 with an empty string
- call members use custom `state_key`s and as such not specifying the sentinel won't match them and state won't be returned


Hywan (approved):
LGTM!

PR #4168, opened 2024-10-25
chore(ffi): introduce `AsyncRuntimeDropped` helper

bnjbvr:
This avoids proliferation of `ManuallyDrop` in the code base, by having
a single type that's used for dropping under an async runtime.

---

I actually saw three uses of `ManuallyDrop` on the FFI layer, but failed to notice that two of them were in the crypto FFI crate. Should I introduce a new FFI helper crate just for sharing this code across the two other crates?

### DIFF
--- a/bindings/matrix-sdk-ffi/src/utils.rs
+++ b/bindings/matrix-sdk-ffi/src/utils.rs
@@ -12,6 +12,9 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use std::{mem::ManuallyDrop, ops::Deref};
+
+use async_compat::TOKIO1 as RUNTIME;
 use ruma::UInt;
 use tracing::warn;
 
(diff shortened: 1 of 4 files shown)


poljar (approved):
No need to share the code I would say.

bindings/matrix-sdk-ffi/src/utils.rs
```diff
@@ -21,3 +24,45 @@ pub(crate) fn u64_to_uint(u: u64) -> UInt {
...
     })
 }
+
+/// Tiny wrappers for data types that must be dropped in the context of an async
+/// runtime.
+///
+/// This is useful whenever such a data type may transitively call some
+/// runtime's `block_on` function in their `Drop` impl (since we lack async drop
+/// at the moment), like done in some `deadpool` drop impls.
+pub(crate) struct AsyncRuntimeDropped<T>(ManuallyDrop<T>);
```
Hywan, line 34: Yes please!

Hywan (approved):

PR #4169, opened 2024-10-25
FFI: Expose support for registration through OIDC in the login details.

pixlwave:
This allows clients to decide whether or not passing `Create` to `url_for_oidc` is allowed.

### DIFF
--- a/bindings/matrix-sdk-ffi/src/client.rs
+++ b/bindings/matrix-sdk-ffi/src/client.rs
@@ -266,7 +266,31 @@ impl Client {
 impl Client {
     /// Information about login options for the client's homeserver.
     pub async fn homeserver_login_details(&self) -> Arc<HomeserverLoginDetails> {
-        let supports_oidc_login = self.inner.oidc().fetch_authentication_issuer().await.is_ok();
+        let oidc = self.inner.oidc();
+        let (supports_oidc_login, supported_oidc_prompts) = match oidc
+            .fetch_authentication_issuer()
+            .await
+        {
+            Ok(issuer) => match &oidc.given_provider_metadata(&issuer).await {
+                Ok(metadata) => {
+                    let prompts = metadata
+                        .prompt_values_supported
+                        .as_ref()
+                        .map_or_else(Vec::new, |prompts| prompts.iter().map(Into::into).collect());
+
+                    (true, prompts)
+                }
+                Err(error) => {
+                    error!("Failed to fetch OIDC provider metadata: {error}");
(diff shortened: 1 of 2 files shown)


bnjbvr (changes requested):
Too much logic at the FFI layer, please create a method in the SDK with tests :)

pixlwave:
> Too much logic at the FFI layer, please create a method in the SDK with tests :)

Haha, a response faster than Clippy! Ok will do next week 🙂

bindings/matrix-sdk-ffi/src/client.rs
```diff
@@ -266,14 +266,29 @@ impl Client {
 impl Client {
     /// Information about login options for the client's homeserver.
     pub async fn homeserver_login_details(&self) -> Arc<HomeserverLoginDetails> {
-        let supports_oidc_login = self.inner.oidc().fetch_authentication_issuer().await.is_ok();
+        let oidc = self.inner.oidc();
+        let (supports_oidc_login, supports_oidc_registration) =
+            if let Ok(issuer) = oidc.fetch_authentication_issuer().await {
+                if let Ok(metadata) = &oidc.given_provider_metadata(&issuer).await {
+                    metadata
+                        .prompt_values_supported
```
bmarty, line 274: Wouldn't be more efficient / generic to expose the `prompt_values_supported` list as a list of `OidcPrompt`?

pixlwave: Sure, I wasn't sure if the concept of `OidcPrompt`s was a bit too much info but if we're happy with that I'll update and at which point, there wouldn't be any custom FFI logic so I would leave all this in the FFI.

pixlwave: Hmmm, the amount of code didn't really change that much, it's basically just the use of `Into` instead of `contains`. WDYT @bnjbvr, I wouldn't really consider there to be any custom FFI logic here now as it is just mapping SDK values to FFI values.

bnjbvr: If that's sufficient for your needs, that's fine with me. (Please make sure to log errors that are silently swallowed here, for both `if let Ok(...)` branches.)

pixlwave: Done in https://github.com/matrix-org/matrix-rust-sdk/commit/4829f48ce609f1809143217f5cddb7172378933d

bnjbvr (approved):
lgtm, thanks!

PR #4170, opened 2024-10-25
chore(room_list): allow knock state event as `latest_event`

jmartinesp:
This allows clients to display pending knocking requests in the room list items (needed by EXI and EXA).

Signed-off-by:


stefanceriu (approved):
Very nice, lgtm 👌

jmartinesp:
@stefanceriu I added some logic changes in https://github.com/matrix-org/matrix-rust-sdk/pull/4170/commits/c79eaf0374325d24a4df47b1ba707bd640d096ee in case you want to re-review. @bnjbvr let me know if you want to take a look too.

crates/matrix-sdk-base/src/rooms/normal.rs
```diff
@@ -508,6 +509,17 @@ impl Room {
...
+    pub async fn power_levels(&self) -> Option<RoomPowerLevels> {
+        self.store
+            .get_state_event_static::<RoomPowerLevelsEventContent>(&self.room_id)
+            .await
+            .ok()
+            .flatten()
+            .and_then(|raw| raw.deserialize().ok())
+            .map(|ev| ev.power_levels())
+    }
+
```
jmartinesp, line 522: We already have `matrix_sdk::Room::room_power_levels`, but we needed something at the `matrix_sdk_base` crate too. Since `sdk::Room` already has access to the fns in `sdk_base::Room`, should I just remove the one in sdk and use this one instead?

crates/matrix-sdk-base/src/latest_event.rs
```diff
@@ -102,8 +118,29 @@ pub fn is_suitable_for_latest_event(event: &AnySyncTimelineEvent) -> PossibleLat
...
 
-        // We don't currently support state events
-        AnySyncTimelineEvent::State(_) => PossibleLatestEvent::NoUnsupportedEventType,
+        // We don't currently support most state events
+        AnySyncTimelineEvent::State(state) => {
+            // But we make an exception for knocked state events *if* the current user
+            // can either accept or decline them
+            if let AnySyncStateEvent::RoomMember(member) = state {
+                if matches!(member.membership(), MembershipState::Knock) {
+                    let can_accept_or_decline_knocks = match power_levels_info {
```
stefanceriu, line 127: Would it make sense to move this logic to within the `RoomPowerLevels`, similar to the other `can_foo`s?

crates/matrix-sdk-base/src/sliding_sync/mod.rs
```diff
@@ -698,14 +698,34 @@ async fn cache_latest_events(
     let mut encrypted_events =
         Vec::with_capacity(room.latest_encrypted_events.read().unwrap().capacity());
 
+    // Try to get room power levels from the current changes
```
stefanceriu, line 701: This feels like a lot of work to do just in case you get a knocked state event, which I presume doesn't happen very often.

crates/matrix-sdk-ui/src/timeline/event_item/mod.rs
```diff
@@ -158,9 +158,21 @@ impl EventTimelineItem {
         let event_id = event.event_id().to_owned();
         let is_own = client.user_id().map(|uid| uid == sender).unwrap_or(false);
 
+        // Get the room's power levels for calculating the latest event
```
stefanceriu, line 161: Same here I guess but I'm not sure how heavy it actually is. Does fetching a room and power levels have any potential of introducing slowdowns? 

Passing the client into PossibleLatestEvent would be weird and it wouldn't improve the sliding sync part 🤔

bnjbvr:
I'll do a lightweight review (I promise :innocent:)

crates/matrix-sdk-base/src/latest_event.rs
jmartinesp: I'm not sure it's worth it, since it'll only be used here (I hope).

crates/matrix-sdk-base/src/sliding_sync/mod.rs
jmartinesp: I agree, but I didn't find any alternatives other than not implementing this part of the functionality 🫤 . We'd have to change the sync process so we can process these latest events *after* the client has processed the sync request and saved all the data to the store and I think that would be quite complex.

crates/matrix-sdk-ui/src/timeline/event_item/mod.rs
jmartinesp: I *think* getting both the room and the state event should be quite lightweight, but I'm not sure how much it could affect performance. The alternative could be to only load them if we reach a knock state event, maybe, although I'd have to change some data structures.

crates/matrix-sdk-base/src/latest_event.rs
```diff
@@ -37,6 +47,9 @@ pub enum PossibleLatestEvent<'a> {
     /// This message is suitable - it's a call notification
     YesCallNotify(&'a SyncCallNotifyEvent),
 
+    /// This state event is suitable - it's a knock membership change
```
bnjbvr, line 50:
```suggestion
    /// This state event is suitable - it's a knock membership change
    /// that can be handled by the current user.
```

crates/matrix-sdk-base/src/rooms/normal.rs
bnjbvr: Indeed, they do the same thing. Can we have this function here return a `Result` instead, to not silently swallow errors, please? (and we can keep the shorter `power_levels` name, I like it better :+1:)

crates/matrix-sdk-base/src/sliding_sync/mod.rs
```diff
@@ -698,14 +698,34 @@ async fn cache_latest_events(
...
+            AnySyncStateEvent::RoomPowerLevels(ev) => Some(ev.power_levels()),
+            _ => None,
+        }
+    };
+
+    // If we didn't get any info, try getting it from local data
+    let power_levels = match power_levels_from_changes() {
+        Some(power_levels) => Some(power_levels),
+        None => room.power_levels().await,
+    };
```
bnjbvr, line 716: Beautiful :star_struck:

crates/matrix-sdk-ui/src/timeline/event_item/mod.rs
```diff
@@ -158,9 +158,21 @@ impl EventTimelineItem {
...
+        let power_levels = if let Some(room) = client.get_room(room_id) {
+            room.power_levels().await
+        } else {
+            None
+        };
+        let room_power_levels_info = match (client.user_id(), power_levels.as_ref()) {
+            (Some(user_id), Some(power_levels)) => Some((user_id, power_levels)),
+            _ => None,
+        };
```
bnjbvr, line 170: GOOD NEWS EVERYONE! There's [`Option::zip()`](https://doc.rust-lang.org/std/option/enum.Option.html#method.zip) that does exactly the same thing :)

```suggestion
        let room_power_levels_info = client.user_id().zip(power_levels.as_ref()));
```

bnjbvr (approved):
LGTM, thanks!

PR #4171, opened 2024-10-27
refactor!: Upgrade Ruma to 0.11.0

zecakeh:
The version that was just published on crates.io.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -45,7 +45,7 @@ once_cell = "1.16.0"
 pin-project-lite = "0.2.9"
 rand = "0.8.5"
 reqwest = { version = "0.12.4", default-features = false }
-ruma = { git = "https://github.com/ruma/ruma", rev = "26165b23fc2ae9928c5497a21db3d31f4b44cc2a", features = [
+ruma = { version = "0.11.0", features = [
     "client-api-c",
     "compat-upload-signatures",
     "compat-user-id",
@@ -59,7 +59,7 @@ ruma = { git = "https://github.com/ruma/ruma", rev = "26165b23fc2ae9928c5497a21d
     "unstable-msc4075",
     "unstable-msc4140",
 ] }
-ruma-common = { git = "https://github.com/ruma/ruma", rev = "26165b23fc2ae9928c5497a21db3d31f4b44cc2a" }
+ruma-common = "0.14.0"
 serde = "1.0.151"
 serde_html_form = "0.2.0"
 serde_json = "1.0.91"
--- a/crates/matrix-sdk-base/src/client.rs
+++ b/crates/matrix-sdk-base/src/client.rs
@@ -43,6 +43,7 @@ use ruma::{
     api::client as api,
     events::{
         ignored_user_list::IgnoredUserListEvent,
+        marked_unread::MarkedUnreadEventContent,
         push_rules::{PushRulesEvent, PushRulesEventContent},
         room::{
             member::{MembershipState, RoomMemberEventContent, SyncRoomMemberEvent},
@@ -683,6 +684,25 @@ impl BaseClient {
             }
         }
 
+        // Helper to update the unread marker for stable and unstable prefixes.
+        fn on_unread_marker(
+            room_id: &RoomId,
+            content: &MarkedUnreadEventContent,
+            room_info: &mut RoomInfo,
+            room_info_notable_updates: &mut BTreeMap<OwnedRoomId, RoomInfoNotableUpdateReasons>,
+        ) {
+            if room_info.base_info.is_marked_unread != content.unread {
+                // Notify the room list about a manual read marker change if the
+                // value's changed.
+                room_info_notable_updates
+                    .entry(room_id.to_owned())
+                    .or_default()
+                    .insert(RoomInfoNotableUpdateReasons::UNREAD_MARKER);
+            }
+
+            room_info.base_info.is_marked_unread = content.unread;
+        }
+
         // Handle new events.
         for raw_event in events {
             match raw_event.deserialize() {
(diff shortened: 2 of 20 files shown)


crates/matrix-sdk-base/src/client.rs
```diff
@@ -704,7 +704,20 @@ impl BaseClient {
                                 room_info.base_info.is_marked_unread = event.content.unread;
                             });
                         }
+                        AnyRoomAccountDataEvent::UnstableMarkedUnread(event) => {
```
zecakeh, line 707: We handle both the stable and unstable version to keep the old behavior.

crates/matrix-sdk-crypto/src/types/one_time_keys.rs
```diff
@@ -110,24 +110,28 @@ pub enum OneTimeKey {
...
-                ))
-            }
-            DeviceKeyAlgorithm::SignedCurve25519 => {
+            OneTimeKeyAlgorithm::SignedCurve25519 => {
                 let key: SignedKey = key.deserialize_as()?;
                 Ok(OneTimeKey::SignedKey(key))
             }
-            _ => Err(serde::de::Error::custom(format!("Unsupported key algorithm {algorithm}"))),
+            _ => match algorithm.as_str() {
+                "curve25519" => {
```
zecakeh, line 125: The Matrix spec only defines `signed_curve25519` for one-time keys so we don't have other variants currently in `OneTimeKeyAlgorithm` and we need to fallback to string matching if we want to keep this branch.

crates/matrix-sdk/src/room/mod.rs
```diff
@@ -2941,7 +2941,7 @@ impl Room {
     pub async fn set_unread_flag(&self, unread: bool) -> Result<()> {
         let user_id = self.client.user_id().ok_or(Error::AuthenticationRequired)?;
 
-        let content = MarkedUnreadEventContent::new(unread);
+        let content = UnstableMarkedUnreadEventContent::from(MarkedUnreadEventContent::new(unread));
```
zecakeh, line 2944: To keep the same behavior as before, we need to use the type with the unstable prefix. A migration might need to be written to switch to the stable version.

Cargo.toml
```diff
@@ -59,7 +59,7 @@ ruma = { git = "https://github.com/ruma/ruma", rev = "26165b23fc2ae9928c5497a21d
     "unstable-msc4075",
     "unstable-msc4140",
 ] }
-ruma-common = { git = "https://github.com/ruma/ruma", rev = "26165b23fc2ae9928c5497a21db3d31f4b44cc2a" }
+ruma-common = { version = "0.14.0" }
```
jplatte, line 62:
```suggestion
ruma-common = "0.14.0"
```

crates/matrix-sdk-base/src/client.rs
jmartinesp: Maybe we could extract this to a function, so we don't forget to update one branch in the future?

crates/matrix-sdk-crypto/src/types/one_time_keys.rs
jmartinesp: Do we actually need to keep it for backwards compatibility? Does someone from the crypto team know?

crates/matrix-sdk-base/src/client.rs
zecakeh: Done

Cargo.toml
zecakeh: Applied.

andybalaam:
Nominate @poljar - I don't feel qualified to review this. Damir, feel free to send me some other review in exchange :-)

crates/matrix-sdk-crypto/src/types/one_time_keys.rs
poljar: The `Key` variant is actually unused, so let's just remove it.

poljar (changes requested):
Thanks, looks good, let's just remove the unsigned Curve25519 key handling, we don't use it.

crates/matrix-sdk-crypto/src/types/one_time_keys.rs
zecakeh: Maybe this can be done in a separate PR? Because this is a breaking change for the crypto crate since the enum is public.

crates/matrix-sdk/src/notification_settings/rule_commands.rs
```diff
@@ -510,8 +493,7 @@ mod tests {
         #[allow(deprecated)]
```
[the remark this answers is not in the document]
zecakeh: No the scope was never deprecated. It's the rule ID that is deprecated here.

zecakeh:
I realized that since the Ruma types are reexported, this is actually a breaking so the conventional commit should reflect that.

crates/matrix-sdk-crypto/src/types/one_time_keys.rs
poljar: I guess we can do that as well.

crates/matrix-sdk-crypto/src/olm/account.rs
```diff
@@ -1532,9 +1529,9 @@ mod tests {
         let (_, second_one_time_keys, _) = account.keys_for_upload();
         assert!(!second_one_time_keys.is_empty());
 
-        let device_key_ids: BTreeSet<&DeviceKeyId> =
+        let device_key_ids: BTreeSet<&OneTimeKeyId> =
```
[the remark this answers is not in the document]
zecakeh: Done

poljar:
> I realized that since the Ruma types are reexported, this is actually a breaking change so the conventional commit should reflect that.

Alright, feel free to rewrite the history. The PR looks good.

poljar (approved):

PR #4172, opened 2024-10-28
test(sdk): Improve the test suite of `RoomEvents`

Hywan:
This patch should be reviewed commit-by-commit. The idea is to improve the test suite of `RoomEvents`.

### DIFF
--- a/crates/matrix-sdk/src/event_cache/linked_chunk/mod.rs
+++ b/crates/matrix-sdk/src/event_cache/linked_chunk/mod.rs
@@ -852,6 +852,13 @@ impl ChunkIdentifierGenerator {
 #[repr(transparent)]
 pub struct ChunkIdentifier(u64);
 
+#[cfg(test)]
+impl PartialEq<u64> for ChunkIdentifier {
+    fn eq(&self, other: &u64) -> bool {
+        self.0 == *other
+    }
+}
+
 /// The position of something inside a [`Chunk`].
 ///
 /// It's a pair of a chunk position and an item index.
--- a/crates/matrix-sdk/src/event_cache/store.rs
+++ b/crates/matrix-sdk/src/event_cache/store.rs
@@ -137,3 +137,386 @@ impl fmt::Debug for RoomEvents {
         formatter.debug_struct("RoomEvents").field("chunk", &self.chunks).finish()
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use assert_matches2::assert_let;
+    use matrix_sdk_test::{EventBuilder, ALICE};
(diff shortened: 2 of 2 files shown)


crates/matrix-sdk/src/event_cache/store.rs
```diff
@@ -18,6 +18,9 @@ use matrix_sdk_common::deserialized_responses::SyncTimelineEvent;
 
 use super::linked_chunk::{Chunk, ChunkIdentifier, Error, Iter, LinkedChunk, Position};
 
+/// An alias for the real event type.
+pub type Event = SyncTimelineEvent;
```
poljar, line 22: I wonder if that won't confuse people more than it will help.

crates/matrix-sdk/src/event_cache/store.rs
```diff
@@ -38,6 +43,7 @@ impl Default for RoomEvents {
 }
 
 impl RoomEvents {
+    /// Build a new `Self` with zero events.
```
poljar, line 46: Addressing the current type as `Self` in documentation continues to be quite jarring.

```suggestion
    /// Build a new [`RoomEvents`] struct with zero events.
```

crates/matrix-sdk/src/event_cache/linked_chunk/mod.rs
```diff
@@ -852,6 +852,12 @@ impl ChunkIdentifierGenerator {
 #[repr(transparent)]
 pub struct ChunkIdentifier(u64);
 
+impl PartialEq<u64> for ChunkIdentifier {
```
poljar, line 855: Do we want to have this implemented generally? I understand that this is useful for tests, but perhaps we'll want to guard using a `#[cfg(test)]` attribute?

poljar (approved):

crates/matrix-sdk/src/event_cache/linked_chunk/mod.rs
Hywan: Indeed, good point!

crates/matrix-sdk/src/event_cache/store.rs
poljar: No idea if that will help, people complained about too many similarly named types in the internal as well as in the external API.

Hywan: I prefer to keep this, and wait for people to complain… It's a rather isolated API.

Hywan: Yup, it's `pub(crate)` now.

PR #4173, opened 2024-10-28
feat(sdk): Add `LinkedChunk::remove_item_at`

Hywan:
This patch should be reviewed commit-by-commit.

The idea is to implement `LinkedChunk::remove_item_at`, which removes an item at a specific position. This new feature propagates to `Update` with `Update::RemoveItem` and to `UpdateToVectorDiff` which can emit a `VectorDiff::Remove`.

One bug has been found and fixed in `UpdateToVectorDiff`, which no impact on the SDK so far since this API is not yet used.

---

### DIFF
--- a/crates/matrix-sdk/src/event_cache/linked_chunk/mod.rs
+++ b/crates/matrix-sdk/src/event_cache/linked_chunk/mod.rs
@@ -406,6 +406,79 @@ impl<const CAP: usize, Item, Gap> LinkedChunk<CAP, Item, Gap> {
         Ok(())
     }
 
+    /// Remove item at a specified position in the [`LinkedChunk`].
+    ///
+    /// Because the `position` can be invalid, this method returns a
+    /// `Result`.
+    pub fn remove_item_at(&mut self, position: Position) -> Result<Item, Error> {
+        let chunk_identifier = position.chunk_identifier();
+        let item_index = position.index();
+
+        let mut chunk_ptr = None;
+        let removed_item;
+
+        {
+            let chunk = self
+                .links
+                .chunk_mut(chunk_identifier)
+                .ok_or(Error::InvalidChunkIdentifier { identifier: chunk_identifier })?;
+
+            let can_unlink_chunk = match &mut chunk.content {
+                ChunkContent::Gap(..) => {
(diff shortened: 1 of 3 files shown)


crates/matrix-sdk/src/event_cache/linked_chunk/mod.rs
```diff
@@ -406,6 +406,79 @@ impl<const CAP: usize, Item, Gap> LinkedChunk<CAP, Item, Gap> {
...
+
+                    current_items.is_empty()
+                }
+            };
+
+            // If the `chunk` can be unlinked, and if the `chunk` is not the first one, we
+            // can remove it.
+            if can_unlink_chunk && chunk.is_first_chunk().not() {
+                // Unlink `chunk`.
+                chunk.unlink(&mut self.updates);
```
andybalaam, line 453: Looks like we should be able to share this logic with `LinkedChunk::replace_gap_at` maybe?

andybalaam (approved):
Looks good, thanks!

crates/matrix-sdk/src/event_cache/linked_chunk/mod.rs
Hywan: Okay, I note that on my todo list.

crates/matrix-sdk/src/event_cache/linked_chunk/mod.rs
```diff
@@ -1852,6 +1925,206 @@ mod tests {
...
+                    RemoveItem { at: Position(ChunkIdentifier(1), 2) },
+                    RemoveItem { at: Position(ChunkIdentifier(1), 1) },
+                    RemoveItem { at: Position(ChunkIdentifier(1), 0) },
+                    RemoveChunk(ChunkIdentifier(1)),
+                ]
+            );
+        }
+
+        // Remove the first item of the first chunk, 3 times. The chunk is empty after
+        // that. The chunk is NOT removed because it's the first chunk.
```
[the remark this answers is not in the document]
Hywan: I was hesitant on this one, the balance is more code vs. more tests. I usually prefer small unit tests though. I note that on my to do list. Thanks for the feedback.

PR #4174, opened 2024-10-28
feat(sdk): Introduce `event_cache::Deduplicator`

Hywan:
This patch is based on https://github.com/matrix-org/matrix-rust-sdk/pull/4173 (it must be merged before this one).

This patch should be reviewed commit-by-commit. Only the last two commits are new (the others are part of #4173):

* [2900f69](https://github.com/matrix-org/matrix-rust-sdk/commit/2900f694c833053a0fcbb67b3359d98c4ccc19aa) aims at implementing `Deduplicator`, an efficient type to detect duplicated event. It uses a bloom filter (one we had from our dependencies already), and decorates all events with a `Decoration::Unique` if the event is unique, `Decoration::Duplicated` if it's duplicated, or `Decoration::Invalid` if it's invalid (if it has no `event_id`).
* [8c41d63](https://github.com/matrix-org/matrix-rust-sdk/commit/8c41d63ab92cf4f95ce9f92a442c36f1eac6bc15) uses `Deduplicator` and `LinkedChunk::remove_item_at` (from #4173!) to remove duplicated items from the event cache/the linked chunk.

In the presence of two duplicated events, the older one is removed. It mimics the current behaviour from the `Timeline` API.


Hywan:
Rebased on `main` since #4173 is now merged.

crates/matrix-sdk/src/event_cache/deduplicator.rs
```diff
@@ -0,0 +1,266 @@
...
+
+impl Deduplicator {
+    const APPROXIMATED_MAXIMUM_NUMBER_OF_EVENTS: usize = 800_000;
+    const DESIRED_FALSE_POSITIVE_RATE: f64 = 0.001;
+
+    /// Create a new `Deduplicator`.
+    pub fn new() -> Self {
+        Self {
+            bloom_filter: Mutex::new(
+                GrowableBloomBuilder::new()
```
bnjbvr, line 45: Sorry to be the knight who says [YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it), but… is it so much faster, considering we do have a btreeset check in both branches after the `check_and_set` below (equivalent to inserting into the btreeset)? Do you have data supporting this (a benchmark)?

crates/matrix-sdk/src/event_cache/deduplicator.rs
```diff
@@ -0,0 +1,266 @@
...
+    {
+        let mut already_seen = BTreeSet::new();
+
+        events_to_scan.map(move |event| {
+            let Some(event_id) = event.event_id() else {
+                // The event has no `event_id`.
+                return Decoration::Invalid(event);
+            };
+
+            if self.bloom_filter.lock().unwrap().check_and_set(&event_id) {
```
bnjbvr, line 79: nit: this lock is taken N times where N is the number of events to scan that have event IDs (should be all of them). Can we take the lock above the `.map()` call?

crates/matrix-sdk/src/event_cache/deduplicator.rs
```diff
@@ -0,0 +1,266 @@
...
+                // Bloom filter has false positives. We are NOT sure the event is NOT present.
+                // Even if the false positive rate is low, we need to iterate over all events to
+                // ensure it isn't present.
+
+                // But first, let's ensure `event` is not a duplicate from `events`, i.e. if the
```
bnjbvr, line 84:
```suggestion
                // But first, let's ensure `event` is not a duplicate from `events_to_scan`, i.e. if the
```

crates/matrix-sdk/src/event_cache/linked_chunk/mod.rs
```diff
@@ -948,6 +948,18 @@ impl Position {
     pub fn index(&self) -> usize {
         self.1
     }
+
+    /// Move the index part (see [`Self::index`]) to the left, i.e. subtract 1.
+    ///
+    /// # Panic
+    ///
+    /// This method will panic if it will overflow, i.e. if the index is 0.
```
bnjbvr, line 956: uber-nit

```suggestion
    /// This method will panic if it will underflow, i.e. if the index is 0.
```

crates/matrix-sdk/src/event_cache/store.rs
```diff
@@ -130,6 +166,126 @@ impl RoomEvents {
...
+    /// associated event (if it exists) from `Self::chunks`.
+    ///
+    /// This is used to remove duplicated events, see
+    /// [`Self::filter_duplicated_events`].
+    fn remove_events(&mut self, event_ids: Vec<OwnedEventId>) {
+        for event_id in event_ids {
+            let Some(event_position) = self.revents().find_map(|(position, event)| {
+                (event.event_id().as_ref() == Some(&event_id)).then_some(position)
+            }) else {
+                error!(?event_id, "An event has been detected, but it's position seems unknown");
```
bnjbvr, line 219: The message is slightly too coupled to the deduplication use case IMO, how about something more generic?

```suggestion
                error!(?event_id, "Couldn't find event to remove");
```

crates/matrix-sdk/src/event_cache/store.rs
```diff
@@ -130,6 +166,126 @@ impl RoomEvents {
...
+    {
+        let mut duplicated_event_ids = Vec::new();
+
+        let deduplicated_events = self
+            .deduplicator
+            .scan_and_learn(events, self)
+            .filter_map(|decorated_event| match decorated_event {
+                Decoration::Unique(event) => Some(event),
+                Decoration::Duplicated(event) => {
+                    error!(event_id = ?event.event_id(), "Found a duplicated event");
```
bnjbvr, line 186: not an error, it's pretty benign, so `trace` or `debug` please :)

crates/matrix-sdk/src/event_cache/store.rs
```diff
@@ -130,6 +166,126 @@ impl RoomEvents {
...
+
+                    duplicated_event_ids.push(event.event_id().expect("The event has no ID"));
```
bnjbvr, line 188: nit: please add safety comment that if it could be marked as invalidated, it must have had an event id in the first place.

crates/matrix-sdk/src/event_cache/store.rs
```diff
@@ -130,6 +166,126 @@ impl RoomEvents {
...
+
+                    // Keep the new event!
+                    Some(event)
+                }
+                Decoration::Invalid(event) => {
+                    error!(?event, "Found an invalid event");
```
bnjbvr, line 194:
```suggestion
                    warn!(?event, "Found an event with no event id");
```

crates/matrix-sdk/src/event_cache/store.rs
```diff
@@ -130,6 +166,126 @@ impl RoomEvents {
...
+                .expect("Failed to remove an event we have just found");
+        }
+    }
+
+    /// Remove all events from `Self::chunks` and update a fix [`Position`].
+    ///
+    /// This method iterates over all event IDs in `event_ids` and removes the
+    /// associated event (if it exists) from `Self::chunks`, exactly like
+    /// [`Self::remove_events`]. The difference is that it will update a
+    /// [`Position`] according to the removals.
```
bnjbvr, line 235: Reading this doc comment, I have no ideas what is `Position` about, how it's been selected, what it represents, sorry. Can you rephrase the doc comment?

crates/matrix-sdk/src/event_cache/store.rs
```diff
@@ -130,6 +166,126 @@ impl RoomEvents {
...
+    fn remove_events_and_update_position(
+        &mut self,
+        event_ids: Vec<OwnedEventId>,
+        position: &mut Position,
+    ) {
+        for event_id in event_ids {
+            let Some(event_position) = self.revents().find_map(|(position, event)| {
+                (event.event_id().as_ref() == Some(&event_id)).then_some(position)
+            }) else {
+                error!(?event_id, "An event has been detected, but it's position seems unknown");
```
bnjbvr, line 248:
```suggestion
                error!(?event_id, "Couldn't find event to remove");
```

crates/matrix-sdk/src/event_cache/store.rs
```diff
@@ -130,6 +166,126 @@ impl RoomEvents {
...
+                .remove_item_at(event_position)
+                .expect("Failed to remove an event we have just found");
+
+            // A `Position` is composed of a `ChunkIdentifier` and an index.
+            // The `ChunkIdentifier` is stable, i.e. it won't change if an
+            // event is removed in another chunk. It means we only need to
+            // update `position` if the removal happened in **the same
+            // chunk**.
+            if event_position.chunk_identifier() == position.chunk_identifier() {
+                // Now we can compare the the position indices.
```
bnjbvr, line 263:
```suggestion
                // Now we can compare the position indices.
```

crates/matrix-sdk/src/event_cache/store.rs
```diff
@@ -130,6 +166,126 @@ impl RoomEvents {
...
+                        // An event has been removed at the same position of
+                        // the new events: `position` does _NOT_ need tp be
+                        // modified.
+                    }
+
+                    // `event_position`'s index > `position`'s index
+                    Ordering::Greater => {
+                        // An event has been removed _after_ the new events:
+                        // `position` does _NOT_ need to be modified.
+                    }
```
bnjbvr, line 284:
```suggestion
                    // `event_position`'s index >= `position`'s index
                    Ordering::Equal | Ordering::Greater => {
                        // An event has been removed at the _same_ or _after_ the new events:
                        // `position` does _NOT_ need to be modified.
                    }
```

crates/matrix-sdk/src/event_cache/store.rs
```diff
@@ -352,8 +528,38 @@ mod tests {
...
+            // The first `event_id_0` has been removed.
+            assert_let!(Some((position, event)) = events.next());
+            assert_eq!(position.chunk_identifier(), 0);
+            assert_eq!(position.index(), 0);
             assert_eq!(event.event_id().unwrap(), event_id_1);
 
+            assert_let!(Some((position, event)) = events.next());
+            assert_eq!(position.chunk_identifier(), 0);
+            assert_eq!(position.index(), 1);
+            assert_eq!(event.event_id().unwrap(), event_id_0);
```
bnjbvr, line 551: Could we have a small assert helper function (if not a `macro_rules!`) that checks chunk_identifier, index and event id?

crates/matrix-sdk/src/event_cache/store.rs
```diff
@@ -519,4 +744,257 @@ mod tests {
...
+
+                assert_let!(Some((position, event)) = events.next());
+                assert_eq!(position.chunk_identifier(), 2);
+                assert_eq!(position.index(), 0);
+                assert_eq!(event.event_id().unwrap(), event_id_7);
+
+                assert_let!(Some((position, event)) = events.next());
+                assert_eq!(position.chunk_identifier(), 2);
+                assert_eq!(position.index(), 1);
+                assert_eq!(event.event_id().unwrap(), event_id_8);
```
bnjbvr, line 901: ouch yeah, let's introduce a test helper for that

crates/matrix-sdk/src/event_cache/store.rs
```diff
@@ -130,6 +166,126 @@ impl RoomEvents {
...
+    /// Remove all events from `Self::chunks` and update a fix [`Position`].
+    ///
+    /// This method iterates over all event IDs in `event_ids` and removes the
+    /// associated event (if it exists) from `Self::chunks`, exactly like
+    /// [`Self::remove_events`]. The difference is that it will update a
+    /// [`Position`] according to the removals.
+    ///
+    /// This is used to remove duplicated events, see
+    /// [`Self::filter_duplicated_events`].
+    fn remove_events_and_update_position(
```
bnjbvr, line 239: Now that I understand what it does: how about renaming `remove_events_and_update_insert_position`?

crates/matrix-sdk/src/event_cache/deduplicator.rs
```diff
@@ -0,0 +1,266 @@
...
+                // iterator itself contains duplicated events! We use a `BTreetSet`, otherwise
```
bnjbvr, line 85: `BTroutSet` :fish: 

```suggestion
                // iterator itself contains duplicated events! We use a `BTreeSet`, otherwise
```

crates/matrix-sdk/src/event_cache/deduplicator.rs
```diff
@@ -0,0 +1,266 @@
...
+    /// Each scanned event will update `Self`'s internal state.
+    ///
+    /// `existing_events` represents all events of a room that already exist.
+    pub fn scan_and_learn<'a, I>(
+        &'a self,
+        events_to_scan: I,
+        existing_events: &'a RoomEvents,
+    ) -> impl Iterator<Item = Decoration<I::Item>> + 'a
+    where
+        I: Iterator<Item = Event> + 'a,
```
bnjbvr, line 69: I'm not sure about this API's signature: instead of taking an iterator and returning one, could we have a function that operates on a single item at a time, returning a decoration for it? `already_seen` could be a member of this struct (which would be more optimized too), and with respect to the YAGNI about the bloom filter, we could likely only use `already_seen`…

crates/matrix-sdk/src/event_cache/deduplicator.rs
```diff
@@ -0,0 +1,266 @@
...
+        SyncTimelineEvent::new(event_builder.make_sync_message_event_with_id(
+            *ALICE,
+            event_id,
+            RoomMessageEventContent::text_plain("foo"),
+        ))
+    }
+
+    #[test]
+    fn test_filter_no_duplicate() {
+        let event_builder = EventBuilder::new();
```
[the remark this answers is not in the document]
Hywan: Good catch. Done!

crates/matrix-sdk/src/event_cache/deduplicator.rs
Hywan: The `BTreeSet` is used only to check the duplicates in the _new_ events that we are checking. The bloom filter is to check the duplicates in _past_ events. For example:

* The bloom filter contains `[]`. I check `["$ev0", "$ev1", "$ev2]`. No duplicates.
* The bloom filter now contains `["$ev0", "$ev1", "$ev2"]`. I check `["$ev3"]`. No duplicates.
* The bloom filter now contains `["$ev0", "$ev1", "$ev2", "$ev3"]`. I check `["$ev0"]`. There is a duplicate.
* The bloom filter now contains `["$ev0", "$ev1", "$ev2", "$ev3"]`. I check `["$ev4", "$ev4"]`. There is a duplicate but in the _new_ events! That's detected by the use of `BTreeSet`.
* The bloom filter now contains `["$ev0", "$ev1", "$ev2", "$ev3", "$ev4"]` now.

This `BTreeSet` is necessary to efficiently detect duplicates in the _new_ scanned events. If we rely on the Bloom filter only, we cannot check the false positives, because the _new_ events are not yet added into the `RoomEvents`.

You're right the `BTreeSet` is hit in the 2 paths inside `check_and_set` but the bloom filter is still more memory-space efficient than if we were using `BTreeSet` for _new_ (currently scanned) **and** _past_ (already scanned) events. That's the only reason I use a bloom filter here: to save memory.

crates/matrix-sdk/src/event_cache/deduplicator.rs
Hywan: Sadly no because the iterator is returned, so it would outlive the lock…

crates/matrix-sdk/src/event_cache/deduplicator.rs
Hywan: It's indeed a typo, but it's `existing_events`, not `events_to_scan`.

Hywan: Thanks!

crates/matrix-sdk/src/event_cache/store.rs
Hywan: I agree. I was hesitant keeping this message. Updating it!

crates/matrix-sdk/src/event_cache/store.rs
Hywan: Went for a `debug!`.

crates/matrix-sdk/src/event_cache/store.rs
Hywan: Okay.

crates/matrix-sdk/src/event_cache/store.rs
Hywan: Better :+1:.

crates/matrix-sdk/src/event_cache/store.rs
Hywan: Done.

crates/matrix-sdk/src/event_cache/store.rs
Hywan: Good idea.

crates/matrix-sdk/src/event_cache/store.rs
Hywan: Sir, yes Sir!

```diff
-            {
-                let mut events = room_events.events();
-
-                assert_let!(Some((position, event)) = events.next());
-                assert_eq!(position.chunk_identifier(), 0);
-                assert_eq!(position.index(), 0);
-                assert_eq!(event.event_id().unwrap(), event_id_1);
-
-                assert_let!(Some((position, event)) = events.next());
-                assert_eq!(position.chunk_identifier(), 0);
-                assert_eq!(position.index(), 1);
-                assert_eq!(event.event_id().unwrap(), event_id_2);
-
-                assert_let!(Some((position, event)) = events.next());
-                assert_eq!(position.chunk_identifier(), 0);
-                assert_eq!(position.index(), 2);
-                assert_eq!(event.event_id().unwrap(), event_id_3);
-
-                assert_let!(Some((position, event)) = events.next());
-                assert_eq!(position.chunk_identifier(), 0);
-                assert_eq!(position.index(), 3);
-                assert_eq!(event.event_id().unwrap(), event_id_6);
-
-                assert_let!(Some((position, event)) = events.next());
-                assert_eq!(position.chunk_identifier(), 2);
-                assert_eq!(position.index(), 0);
-                assert_eq!(event.event_id().unwrap(), event_id_7);
-
-                assert_let!(Some((position, event)) = events.next());
-                assert_eq!(position.chunk_identifier(), 2);
-                assert_eq!(position.index(), 1);
-                assert_eq!(event.event_id().unwrap(), event_id_8);
-
-                assert!(events.next().is_none());
-            }
+            assert_events_eq!(
+                room_events.events(),
+                [
+                    (event_id_1 at (0, 0)),
+                    (event_id_2 at (0, 1)),
+                    (event_id_3 at (0, 2)),
+                    (event_id_6 at (0, 3)),
+                    (event_id_7 at (2, 0)),
+                    (event_id_8 at (2, 1)),
+                ]
+            );
```

It's pretty useful :-).

crates/matrix-sdk/src/event_cache/deduplicator.rs
Hywan: <3

crates/matrix-sdk/src/event_cache/store.rs
Hywan: Okay with the name.

crates/matrix-sdk/src/event_cache/deduplicator.rs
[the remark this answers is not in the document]
bnjbvr: …Not done?

crates/matrix-sdk/src/event_cache/deduplicator.rs
```diff
@@ -0,0 +1,266 @@
...
+
+    #[test]
+    fn test_filter_no_duplicate() {
+        let event_id_0 = owned_event_id!("$ev0");
+        let event_id_1 = owned_event_id!("$ev1");
+        let event_id_2 = owned_event_id!("$ev2");
+
+        let event_0 = sync_timeline_event(&event_id_0);
+        let event_1 = sync_timeline_event(&event_id_1);
+        let event_2 = sync_timeline_event(&event_id_2);
```
bnjbvr, line 160: Time for an `EventFactory 101` :nerd_face: You can curry some parameters at the factory level, to avoid dull repetition, which would justify *not* having to have the `sync_timeline_event` helper:

```suggestion
        let f = EventFactory::new().sender(user_id!("@mnt_io:matrix.org"));
        let event_0 = f.text_msg("").event_id(&event_id_0).into_sync();
        let event_1 = f.text_msg("").event_id(&event_id_1).into_sync();
        let event_2 = f.text_msg("").event_id(&event_id_2).into_sync();
```

crates/matrix-sdk/src/event_cache/room/events.rs
```diff
@@ -129,24 +167,182 @@ impl RoomEvents {
...
+    fn remove_events_and_update_insert_position(
+        &mut self,
+        event_ids: Vec<OwnedEventId>,
+        position: &mut Position,
+    ) {
+        for event_id in event_ids {
+            let Some(event_position) = self.revents().find_map(|(position, event)| {
+                (event.event_id().as_ref() == Some(&event_id)).then_some(position)
+            }) else {
+                error!(?event_id, "Cannot find te event to remove");
```
bnjbvr, line 277:
```suggestion
                error!(?event_id, "Cannot find the event to remove");
```

crates/matrix-sdk/src/event_cache/room/events.rs
```diff
@@ -129,24 +167,182 @@ impl RoomEvents {
...
+                        // events: `position` needs to be shifted to the
+                        // left by 1.
+                        position.move_index_to_the_left();
+                    }
+
+                    // `event_position`'s index >= `position`'s index
+                    Ordering::Equal | Ordering::Greater => {
+                        // An event has been removed at the _same_ position of
+                        // or _after_ the new events: `position` does _NOT_ need
+                        // tp be modified.
```
bnjbvr, line 306:
```suggestion
                        // to be modified.
```

crates/matrix-sdk/src/event_cache/room/events.rs
```diff
@@ -129,24 +167,182 @@ impl RoomEvents {
...
+            // chunk**.
+            if event_position.chunk_identifier() == position.chunk_identifier() {
+                // Now we can compare the position indices.
+                match event_position.index().cmp(&position.index()) {
+                    // `event_position`'s index < `position`'s index
+                    Ordering::Less => {
+                        // An event has been removed _before_ the new
+                        // events: `position` needs to be shifted to the
+                        // left by 1.
+                        position.move_index_to_the_left();
```
bnjbvr, line 299: Could `move_index_to_the_left` be called `decrement_index`?

crates/matrix-sdk/src/event_cache/room/events.rs
```diff
@@ -129,24 +167,182 @@ impl RoomEvents {
...
+                let mut events = $events_iterator;
+
+                $(
+                    assert_let!(Some((position, event)) = events.next());
+                    assert_eq!(position.chunk_identifier(), $chunk_identifier );
+                    assert_eq!(position.index(), $index );
+                    assert_eq!(event.event_id().unwrap(), $event_id );
+                )*
+
+                assert!(events.next().is_none(), "No more event is expected");
```
bnjbvr, line 334: Zero is plural in English :nerd_face: 

```suggestion
                assert!(events.next().is_none(), "No more events are expected");
```

crates/matrix-sdk/src/event_cache/room/events.rs
```diff
@@ -129,24 +167,182 @@ impl RoomEvents {
...
 mod tests {
     use assert_matches2::assert_let;
-    use matrix_sdk_test::{EventBuilder, ALICE};
-    use ruma::{events::room::message::RoomMessageEventContent, EventId, OwnedEventId};
+    use ruma::{user_id, EventId, OwnedEventId};
 
     use super::*;
+    use crate::test_utils::events::EventFactory;
+
+    macro_rules! assert_events_eq {
```
bnjbvr, line 322: So much nicer with this macro :star_struck:

crates/matrix-sdk/src/event_cache/deduplicator.rs
Hywan: Thanks for the tips.

I agree it's nice but in this case it adds a non-negligible cognitive effort to read `.text_msg("")` and `.into_sync();` here. I think it's fine to keep `sync_timeline_event` as it makes the test code easier to understand, and we can focus on the real things we want to test rather than the prelude.

crates/matrix-sdk/src/event_cache/deduplicator.rs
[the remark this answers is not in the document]
Hywan: I don't use `EventBuilder` but I still prefer to use a helper to make the prelude of the test easier simpler. See my comment here https://github.com/matrix-org/matrix-rust-sdk/pull/4174#discussion_r1820902651.

crates/matrix-sdk/src/event_cache/room/events.rs
Hywan: 🤦  thanks!

crates/matrix-sdk/src/event_cache/room/events.rs
Hywan: Sure.

crates/matrix-sdk/src/event_cache/room/events.rs
Hywan: 🤦

crates/matrix-sdk/src/event_cache/room/events.rs
Hywan: It is!

crates/matrix-sdk/src/event_cache/room/events.rs
Hywan: TIL

crates/matrix-sdk/src/event_cache/deduplicator.rs
```diff
@@ -0,0 +1,266 @@
...
+    /// `existing_events` represents all events of a room that already exist.
+    pub fn scan_and_learn<'a, I>(
+        &'a self,
+        events_to_scan: I,
+        existing_events: &'a RoomEvents,
+    ) -> impl Iterator<Item = Decoration<I::Item>> + 'a
+    where
+        I: Iterator<Item = Event> + 'a,
+    {
+        let mut already_seen = BTreeSet::new();
```
bnjbvr, line 77: 1. Can you comment that this is a temporary variable (and not stored in `Deduplicator`) because we don't want an unbounded growth for this set? (and that it justifies the shape of the API, taking an in-parameter that's an iterator)
2. Can you also comment what its purpose is, here (i.e. find duplicates in the incoming iterator), please?

bnjbvr (approved):

Issue #4162, opened 2024-10-22
send queue: handle redactions and get local echoes for redactions


bnjbvr:
For regular events and dependent events, the send queue can abort sending. If the underlying event was not being sent at the time of the abort request, then it's removed from the send queue's storage; otherwise, it results in a redaction event being sent to the server.

However, this is not infallible if a user tries to do it the other way, i.e. there was a reaction to an event, and they try to redact it and add it again. In this case, since the redaction isn't handled by the send queue, it may happen before or after the reaction has been added back, resulting in weird errors and situations (i.e. after removing and re-adding the reaction, it's not clear what's the final state of the reaction is in the room — and the timeline may show something that's temporarily incorrect, as a result).

Reusing the send queue for redactions would generally solve the problem of ordering here, and make it possible to ensure a total ordering for sending and redacting events in general. It could also mean that we may have local echo for redactions; this would require remembering what the redacted event was, for a short period of time, in case the redaction itself fails, but that's not intractable.

PR #4176, opened 2024-10-28
refactor(send queue): refactor in preparation for media local echoes

bnjbvr:
This refactors the send queue so that it can apply sending more abstract requests, instead of only sending plain events. This will generalize to media uploads soon, but in the future might also generalize to [sending redactions](https://github.com/matrix-org/matrix-rust-sdk/issues/4162), etc. So we now queue requests, update requests, and so on; which leads to some fun renaming in a massive patch, but no changes in functionality. Those would come in subsequent pull requests.

Indexeddb has a database format change, that is gently handled the same way we did for is_wedged vs wedged errors, that is: the migration applies at the reading boundary (we can't use the `#[serde(untagged)]` enum trick because we're storing a raw JSON value, and this is known to be [buggy](https://github.com/serde-rs/json/issues/497) in serde). Sqlite hasn't been modified yet.

Part of #1732.

### DIFF
--- a/bindings/matrix-sdk-ffi/src/client.rs
+++ b/bindings/matrix-sdk-ffi/src/client.rs
@@ -503,7 +503,7 @@ impl Client {
         Arc::new(TaskHandle::new(RUNTIME.spawn(async move {
             // Respawn tasks for rooms that had unsent events. At this point we've just
             // created the subscriber, so it'll be notified about errors.
-            q.respawn_tasks_for_rooms_with_unsent_events().await;
+            q.respawn_tasks_for_rooms_with_unsent_requests().await;
 
             loop {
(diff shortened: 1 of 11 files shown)


bnjbvr:
Wait, a rename of serialized JSON field names is a breaking change, since the field name is encoded next to the value, IIRC. Will have to do the sqlite migration here too \o/

bnjbvr:
> Wait, a rename of serialized JSON field names is a breaking change, since the field name is encoded next to the value, IIRC. Will have to do the sqlite migration here too \o/

[nope](https://serde.rs/variant-attrs.html#rename)

andybalaam (approved):
Looks good, and thank you for breaking into consumable parts.

PR #4177, opened 2024-10-28
feat(event cache store): add ability to rename a media cache key

bnjbvr:
What it says on the tin. Will allow caching medias before they're sent to a remote, when used with the send queue.

Part of #1732.


jmartinesp (approved):
LGTM! Very clean PR, thanks.

PR #4178, opened 2024-10-28
chore(sdk): A bit of code moving in `event_cache`

Hywan:
This patch renames `event_cache/store.rs` to `event_cache/room/events.rs`, and extract types about `RoomEventCache` from `event_cache/mod.rs` to `event_cache/room/mod.rs`. I hope it clarifies things.

It's purely code moves, no renaming, no other changes.


bnjbvr (changes requested):
Thanks! There are some unit tests that would require some code motion too, please :-)

bnjbvr (approved):

PR #4187, opened 2024-10-30
feat(ffi): add `Client::public_rooms` function

jmartinesp:
This is needed to get rooms from the public rooms directory. Also add `ffi::PublicRoomsChunk` structure to map the resulting data.

A customer couldn't find a way to use this feature in their fork because the method wasn't in the FFI layer, so I added it here.

Signed-off-by:


pixlwave:
Shouldn't this be managed through the existing `RoomDirectorySearch` type, rather than exposing the raw API?

jmartinesp:
> Shouldn't this be managed through the existing `RoomDirectorySearch` type, rather than exposing the raw API?

Ah, I didn't realise this already existed since I couldn't find the raw API call being used in the project. Closing this, I'll modify the directory search API then (if possible).

PR #4188, opened 2024-10-30
task(room_directory_search): add 'server' parameter to the room directory search

jmartinesp:
What the title says. This was previously missing.

Signed-off-by:

### DIFF
--- a/bindings/matrix-sdk-ffi/src/room_directory_search.rs
+++ b/bindings/matrix-sdk-ffi/src/room_directory_search.rs
@@ -18,6 +18,7 @@ use std::{fmt::Debug, sync::Arc};
 use eyeball_im::VectorDiff;
 use futures_util::StreamExt;
 use matrix_sdk::room_directory_search::RoomDirectorySearch as SdkRoomDirectorySearch;
+use ruma::ServerName;
 use tokio::sync::RwLock;
 
 use super::RUNTIME;
@@ -68,6 +69,12 @@ impl From<matrix_sdk::room_directory_search::RoomDescription> for RoomDescriptio
     }
 }
 
+/// A helper for performing room searches in the room directory.
+/// The way this is intended to be used is:
+///
+/// 1. Register a callback using [`RoomDirectorySearch::results`].
+/// 2. Start the room search with [`RoomDirectorySearch::search`].
+/// 3. To get more results, use [`RoomDirectorySearch::next_page`].
 #[derive(uniffi::Object)]
 pub struct RoomDirectorySearch {
     pub(crate) inner: RwLock<SdkRoomDirectorySearch>,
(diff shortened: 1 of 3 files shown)


jmartinesp:
I wanted to add some tests, but I'm not sure how to test this, to be honest 🫤 . Should I just add some mock response where I check the request contains the server parameter?

bindings/matrix-sdk-ffi/src/room_directory_search.rs
```diff
@@ -87,9 +88,15 @@ impl RoomDirectorySearch {
         Ok(())
     }
 
-    pub async fn search(&self, filter: Option<String>, batch_size: u32) -> Result<(), ClientError> {
+    pub async fn search(
+        &self,
+        filter: Option<String>,
+        batch_size: u32,
+        server: Option<String>,
```
pixlwave, line 95: Random thought during a quick skim: As this parameter isn't strongly typed in the FFI, it might be worthwhile calling this `server_name` (or perhaps `remote_server_name`) to indicate that you're not meant to pass a homeserver URL here.

jmartinesp: Good call, done in https://github.com/matrix-org/matrix-rust-sdk/pull/4188/commits/7cb15aa2423956c8bd7f2ed21e563bb24f164535.

bindings/matrix-sdk-ffi/src/room_directory_search.rs
```diff
@@ -87,9 +88,15 @@ impl RoomDirectorySearch {
         Ok(())
     }
 
-    pub async fn search(&self, filter: Option<String>, batch_size: u32) -> Result<(), ClientError> {
+    pub async fn search(
```
bnjbvr, line 91: Since you're here, can you add a doc comment here too please?

bindings/matrix-sdk-ffi/src/room_directory_search.rs
```diff
@@ -87,9 +88,15 @@ impl RoomDirectorySearch {
...
+        &self,
+        filter: Option<String>,
+        batch_size: u32,
+        server_name: Option<String>,
```
bnjbvr, line 95: can we call this "via_server_name"? a cursory look made me think it was a server name we were looking for (before i saw `filter`).

crates/matrix-sdk/src/room_directory_search.rs
```diff
@@ -118,6 +118,7 @@ impl SearchState {
 pub struct RoomDirectorySearch {
     batch_size: u32,
     filter: Option<String>,
+    server: Option<OwnedServerName>,
```
bnjbvr, line 121: ditto, how about `via_server` or just `via` that i think we've used in other places?

crates/matrix-sdk/src/room_directory_search.rs
```diff
@@ -141,14 +143,23 @@ impl RoomDirectorySearch {
...
+    ///
     /// This method will clear the current search results and start a new one.
     // Should never be used concurrently with another `next_page` or a
     // `search`.
-    pub async fn search(&mut self, filter: Option<String>, batch_size: u32) -> Result<()> {
+    pub async fn search(
+        &mut self,
+        filter: Option<String>,
+        batch_size: u32,
+        server: Option<OwnedServerName>,
```
bnjbvr, line 156: ditto via_server or via

bnjbvr (approved):
lgtm, thanks

bnjbvr:
> I wanted to add some tests, but I'm not sure how to test this, to be honest 🫤 . Should I just add some mock response where I check the request contains the server parameter?

You could, but it's strictly testing the behavior of Ruma, so it's not worth IMO. Feel free to call the `server()` function in one test, so we have basic coverage for it.

PR #4191, opened 2024-10-30
chore(ffi): in `Client::account_url` return early when we're not an OIDC session

bnjbvr:
This avoids one spammy error log for sessions not using oidc.

### DIFF
--- a/bindings/matrix-sdk-ffi/src/client.rs
+++ b/bindings/matrix-sdk-ffi/src/client.rs
@@ -606,6 +606,10 @@ impl Client {
         &self,
         action: Option<AccountManagementAction>,
     ) -> Result<Option<String>, ClientError> {
+        if !matches!(self.inner.auth_api(), Some(AuthApi::Oidc(..))) {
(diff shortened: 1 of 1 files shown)


bindings/matrix-sdk-ffi/src/client.rs
```diff
@@ -606,6 +606,9 @@ impl Client {
         &self,
         action: Option<AccountManagementAction>,
     ) -> Result<Option<String>, ClientError> {
+        if !matches!(self.inner.auth_api(), Some(AuthApi::Oidc(..))) {
+            return Ok(None);
+        }
```
poljar, line 611:
```suggestion
        }
        
```

poljar (approved):

bindings/matrix-sdk-ffi/src/client.rs
bnjbvr: You kept some trailing whitespaces, that's pure evil, I admire that.

PR #4192, opened 2024-10-30
feat: Implement cross-process lock for the `EventCache`

Hywan:
This PR must be reviewed commit-by-commit.

It introduces the `LockableEventCacheStore` type, which wraps `DynEventCacheStore`, and that implements `BackingStore` so that it can be used inside the `CrossProcessStoreLock`.

It then introduces the `EventCacheStoreLock` type, which glue all pieces together: the `CrossProcessStoreLock` and the `LockableEventCacheStore`. It exposes a single method: `lock(&'a self) -> Result<EventCacheStoreLockGuard<'a>, _>`, which allows to take the lock. The guard derefs to `DynEventCacheStore`. The resulting API is straightforward to use according to me: `client.event_cache_store().lock().await?`, pretty Rust idiomatic.

Finally, the PR replaces all uses of `DynEventCacheStore` by `EventCacheStoreLock`. There is no more public `DynEventCacheStore`, expect via `EventCacheStoreLock`.

A follow-up patch must update `matrix_sdk::event_cache` to refresh the in-memory data of the event cache when the lock holder has changed. I believe it's nice to introduce that in another PR instead of increasing the number of patches in this one.

---

### DIFF
--- a/crates/matrix-sdk-base/src/client.rs
+++ b/crates/matrix-sdk-base/src/client.rs
@@ -13,11 +13,12 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#[cfg(feature = "e2e-encryption")]
+use std::sync::Arc;
 use std::{
     collections::{BTreeMap, BTreeSet},
     fmt, iter,
     ops::Deref,
-    sync::Arc,
 };
 
 use eyeball::{SharedObservable, Subscriber};
@@ -71,7 +72,7 @@ use crate::RoomMemberships;
 use crate::{
     deserialized_responses::{RawAnySyncOrStrippedTimelineEvent, SyncTimelineEvent},
     error::{Error, Result},
-    event_cache_store::DynEventCacheStore,
+    event_cache_store::EventCacheStoreLock,
     response_processors::AccountDataProcessor,
     rooms::{
         normal::{RoomInfoNotableUpdate, RoomInfoNotableUpdateReasons},
@@ -95,7 +96,7 @@ pub struct BaseClient {
     pub(crate) store: Store,
 
     /// The store used by the event cache.
-    event_cache_store: Arc<DynEventCacheStore>,
+    event_cache_store: EventCacheStoreLock,
 
     /// The store used for encryption.
     ///
@@ -114,8 +115,7 @@ pub struct BaseClient {
     pub(crate) ignore_user_list_changes: SharedObservable<Vec<String>>,
 
     /// A sender that is used to communicate changes to room information. Each
-    /// event contains the room and a boolean whether this event should
-    /// trigger a room list update.
+    /// tick contains the room ID and the reasons that have generated this tick.
     pub(crate) room_info_notable_update_sender: broadcast::Sender<RoomInfoNotableUpdate>,
 
     /// The strategy to use for picking recipient devices, when sending an
@@ -249,14 +249,13 @@ impl BaseClient {
     }
 
     /// Get a reference to the store.
-    #[allow(unknown_lints, clippy::explicit_auto_deref)]
     pub fn store(&self) -> &DynStateStore {
         self.store.deref()
     }
 
     /// Get a reference to the event cache store.
-    pub fn event_cache_store(&self) -> &DynEventCacheStore {
-        self.event_cache_store.as_ref()
+    pub fn event_cache_store(&self) -> &EventCacheStoreLock {
+        &self.event_cache_store
     }
 
     /// Is the client logged in.
--- a/crates/matrix-sdk-base/src/client.rs
+++ b/crates/matrix-sdk-base/src/client.rs
@@ -13,11 +13,10 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#[cfg(feature = "e2e-encryption")]
-use std::ops::Deref;
 use std::{
     collections::{BTreeMap, BTreeSet},
     fmt, iter,
+    ops::Deref,
     sync::Arc,
 };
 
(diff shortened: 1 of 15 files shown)


crates/matrix-sdk/src/client/builder/mod.rs
```diff
@@ -604,7 +620,11 @@ async fn build_indexeddb_store_config(
 
     let store_config = {
         tracing::warn!("The IndexedDB backend does not implement an event cache store, falling back to the in-memory event cache store…");
-        store_config.event_cache_store(matrix_sdk_base::event_cache_store::MemoryStore::new())
+        store_config.event_cache_store(
+            matrix_sdk_base::event_cache_store::MemoryStore::new(),
+            "default-key".to_owned(),
```
bnjbvr, line 625: As said in another comment, we shouldn't use this value here.

crates/matrix-sdk-base/src/store/mod.rs
```diff
@@ -507,8 +507,11 @@ impl StoreConfig {
             #[cfg(feature = "e2e-encryption")]
             crypto_store: matrix_sdk_crypto::store::MemoryStore::new().into_crypto_store(),
             state_store: Arc::new(MemoryStore::new()),
-            event_cache_store: crate::event_cache_store::MemoryStore::new()
-                .into_event_cache_store(),
+            event_cache_store: event_cache_store::EventCacheStoreLock::new(
+                event_cache_store::MemoryStore::new(),
+                "default-key".to_owned(),
+                "matrix-sdk-base".to_owned(),
```
bnjbvr, line 513: The values provided here are scary, and should be controllable by the callers, otherwise it's not possible to distinguish a process from another. The key could be a constant used everywhere in the codebase (maybe buried into a single place, and then passed around — or buried a single time when creating an `EventCacheStoreLock` from an event cache store impl), but the holder's value has to be provided by the embedder.

crates/matrix-sdk/src/client/builder/mod.rs
```diff
@@ -208,6 +208,7 @@ impl ClientBuilder {
             path: path.as_ref().to_owned(),
             cache_path: None,
             passphrase: passphrase.map(ToOwned::to_owned),
+            event_cache_store_lock_holder: "matrix-sdk".to_owned(),
```
bnjbvr, line 211: (See note about the `holder` value in another comment.)

bnjbvr (changes requested):
Great! It's nice that we can reuse so much code for the cross-process locking :partying_face:

crates/matrix-sdk-base/src/store/mod.rs
Hywan: I would like to address that as a follow-up PR: I want to introduce a "process holder name" inside the client that can be used by all cross-process lock, otherwise things start to be really clumsy. What do you think?

crates/matrix-sdk-base/src/client.rs
```diff
@@ -252,12 +251,12 @@ impl BaseClient {
     /// Get a reference to the store.
     #[allow(unknown_lints, clippy::explicit_auto_deref)]
```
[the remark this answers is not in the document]
Hywan: Indeed!

crates/matrix-sdk-base/src/event_cache_store/mod.rs
```diff
@@ -37,6 +39,75 @@ pub use self::{
...
+    /// The inner cross process lock that is used to lock the `EventCacheStore`.
+    cross_process_lock: CrossProcessStoreLock<LockableEventCacheStore>,
+
+    /// The store itself.
+    ///
+    /// That's the only place where the store exists.
+    store: Arc<DynEventCacheStore>,
+}
+
+impl fmt::Debug for EventCacheStoreLock {
```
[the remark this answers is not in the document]
Hywan: Alrighty!

crates/matrix-sdk-base/src/event_cache_store/mod.rs
```diff
@@ -37,6 +39,75 @@ pub use self::{
...
+pub struct EventCacheStoreLockGuard<'a> {
+    /// The cross process lock guard.
+    #[allow(unused)]
+    cross_process_lock_guard: CrossProcessStoreLockGuard,
+
+    /// A reference to the store.
+    store: &'a DynEventCacheStore,
+}
+
+impl<'a> fmt::Debug for EventCacheStoreLockGuard<'a> {
```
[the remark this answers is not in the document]
Hywan: Sir, yes Sir!

crates/matrix-sdk-base/src/event_cache_store/memory_store.rs
```diff
@@ -57,7 +66,44 @@ impl EventCacheStore for MemoryStore {
         key: &str,
         holder: &str,
     ) -> Result<bool, Self::Error> {
-        todo!()
+        let now = Instant::now();
+        let expiration = now + Duration::from_millis(lease_duration_ms.into());
+
+        match self.leases.write().unwrap().entry(key.to_owned()) {
```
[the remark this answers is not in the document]
Hywan: Done!

crates/matrix-sdk-base/src/store/mod.rs
```diff
@@ -528,8 +531,12 @@ impl StoreConfig {
     }
 
     /// Set a custom implementation of an `EventCacheStore`.
-    pub fn event_cache_store(mut self, event_cache_store: impl IntoEventCacheStore) -> Self {
-        self.event_cache_store = event_cache_store.into_event_cache_store();
+    pub fn event_cache_store<S>(mut self, event_cache_store: S, key: String, holder: String) -> Self
```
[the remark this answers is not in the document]
Hywan: And done!

crates/matrix-sdk-base/src/store/mod.rs
Hywan: @bnjbvr and I have agreed to remove these values with a configuration in the `Client`. I'm already working on this as a follow-up PR.

Hywan: Here it is, https://github.com/matrix-org/matrix-rust-sdk/pull/4224

PR #4193, opened 2024-10-30
feat(ffi): add `Client::is_room_alias_available` function

jmartinesp:
This uses the underlying `sdk::Client::resolve_room_alias` fn to check if a room alias is available for room creation.

Signed-off-by:

### DIFF
--- a/bindings/matrix-sdk-ffi/src/client.rs
+++ b/bindings/matrix-sdk-ffi/src/client.rs
@@ -1027,10 +1027,18 @@ impl Client {
     pub async fn resolve_room_alias(
         &self,
         room_alias: String,
-    ) -> Result<ResolvedRoomAlias, ClientError> {
+    ) -> Result<Option<ResolvedRoomAlias>, ClientError> {
         let room_alias = RoomAliasId::parse(&room_alias)?;
         let response = self.inner.resolve_room_alias(&room_alias).await?;
         Ok(response.into())
+        match self.inner.resolve_room_alias(&room_alias).await {
+            Ok(response) => Ok(Some(response.into())),
+            Err(HttpError::Reqwest(http_error)) => match http_error.status() {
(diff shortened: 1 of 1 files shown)


bindings/matrix-sdk-ffi/src/client.rs
```diff
@@ -1126,6 +1127,23 @@ impl Client {
 
         Ok(())
     }
+
+    /// Checks if a room alias is available in the current homeserver.
+    pub async fn is_room_alias_available(&self, alias: String) -> Result<bool, ClientError> {
+        let alias = RoomAliasId::parse(alias)?;
```
poljar, line 1133:
```suggestion
        let alias = RoomAliasId::parse(alias)?;
        
```

bindings/matrix-sdk-ffi/src/client.rs
```diff
@@ -1126,6 +1127,23 @@ impl Client {
 
         Ok(())
     }
+
+    /// Checks if a room alias is available in the current homeserver.
```
poljar, line 1131: I think we could improve the naming, to something like `alias_exists()` or `alias_is_reserved()`.

When you say available it could mean two things:
1. The alias exists and the room can be resolved using this alias.
2. The alias does not exists and it can be created.

poljar (approved):
Left some nits, feel free to merge once they are addressed.

bindings/matrix-sdk-ffi/src/client.rs
poljar: That sounds like a good idea as well, though I imagine that this method could then be a common helper method that uses `Client::resolve_room_alias` under the hood and converts the result to a boolean.

jmartinesp: I added a couple of commits with the changes above in case you want to do another review.

PR #4195, opened 2024-10-30
feat(send queue): send attachments with the send queue

### LINKED ISSUE #1732: [Media+Timeline] Local echo support for media attachments

bnjbvr:
Fixes #1732. This adds support for sending medias via the send queue, which gives us local echoes for free.

The process looks like this:

- put the medias (thumbnail and file) into the media cache (with temporary MXC IDs)
- send a local echo for the media event that references the local media cache
- if thumbnail
  - create a request to upload the thumbnail
  - create a dependent request to upload the file, depending on the thumbnail being uploaded
  - create a dependent request to send the media event, depending on the file upload
- otherwise
  - create a request to upload the file
  - create a dependent request to send the media event
- once the medias have been uploaded, send a local echo edit for the media event, that fixes the MXC IDs to the final ones

The send queue dependency system is made such that, a dependent request depends on *at most* one other request. The design chosen in this PR is that dependencies are kept linear, each request depends on the previous one to be finished. As such, it means that when uploading a thumbnail, we need to remember the thumbnail MXC ID when uploading the file, because the file request will pass that information (when it's completed) to the media event.

---

This can *not* be reviewed commit by commit. Instead, I'd recommend looking at the commit which improves the module doc comment first: https://github.com/matrix-org/matrix-rust-sdk/pull/4195/commits/1d0ba021d52e72f4ef80c2220357c00193535abe

And then review the rest of the changes; we can also have a call to do the review together.

Sits on top of https://github.com/matrix-org/matrix-rust-sdk/pull/4199 and https://github.com/matrix-org/matrix-rust-sdk/pull/4200.

### DIFF
--- a/bindings/matrix-sdk-ffi/src/client.rs
+++ b/bindings/matrix-sdk-ffi/src/client.rs
@@ -717,6 +717,7 @@ impl Client {
     ) -> Result<Vec<u8>, ClientError> {
         let source = (*media_source).clone();
 
+        debug!(?source, "requesting media file");
         Ok(self
             .inner
             .media()
@@ -732,6 +733,7 @@ impl Client {
     ) -> Result<Vec<u8>, ClientError> {
         let source = (*media_source).clone();
 
+        debug!(source = ?media_source, width, height, "requesting media thumbnail");
         Ok(self
             .inner
             .media()
--- a/bindings/matrix-sdk-ffi/src/client.rs
+++ b/bindings/matrix-sdk-ffi/src/client.rs
@@ -8,7 +8,8 @@ use std::{
 use anyhow::{anyhow, Context as _};
 use matrix_sdk::{
     media::{
-        MediaFileHandle as SdkMediaFileHandle, MediaFormat, MediaRequest, MediaThumbnailSettings,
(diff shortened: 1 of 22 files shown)


Hywan (approved):
Did a pair-review via [Element Call](https://call.element.io/). I'm approving this PR, it's an excellent code. @bnjbvr has taken notes for things that are left to do, but nothing blocking the merge.

Impressive PR. Well done!

PR #4197, opened 2024-10-31
task(tests): introduce prebuilt mocks and mocking helpers

bnjbvr:
This introduces new test utilities, starting with `MatrixMockServer`. This is a pair of a `MockServer` and a `Client` (one can retrieve them respectively with `.server()` and `.client()`).

It implements a few mock endpoints, expanding and limited the shared code as much as possible, so the mocks are still flexible to use as scoped/unscoped mounts, named, and so on.

It works like this:

- start with telling which endpoint you'd like to mock, e.g. `mock_room_send()`. This returns a specialized `MockSomething` data structure, with its own impl. For this example, it's `MockRoomSend`.
- configure the response on the endpoint-specific mock data structure. For instance, if you want the sending to result in a transient failure, call `MockRoomSend::error500()`; if you want it to succeed and return the event $42, call `MockRoomSend::ok(event_id!("$42"))`. It's still possible to call `respond_with()`, as we do with wiremock MockBuilder, for maximum flexibility when the helpers aren't sufficient.
- once the endpoint's response is configured, for *any* mock builder, you get a `MatrixMock`; this is a plain `wiremock::Mock` with the server curried, so one doesn't have to pass it around when calling `.mount()` or `.mount_as_scoped()`. As such, it mostly defers its implementations to `wiremock::Mock` under the hood.

## Example

With this, we can go from that code:

```rust
#[async_test]
async fn test_make_reply_event_doesnt_require_event_cache() {
    let (client, server) = logged_in_client_with_server().await;

    let event_id = event_id!("$1");
    let resp_event_id = event_id!("$resp");
    let room_id = room_id!("!galette:saucisse.bzh");

    let f = EventFactory::new();

    let raw_original_event = f
        .text_msg("hi")
        .event_id(event_id)
        .sender(client.user_id().unwrap())
        .room(room_id)
        .into_raw_timeline();

    mock_sync_with_new_room(
        |builder| {
            builder.add_joined_room(JoinedRoomBuilder::new(room_id));
        },
        &client,
        &server,
        room_id,
    )
    .await;

    Mock::given(method("GET"))
        .and(path_regex(r"^/_matrix/client/r0/rooms/.*/event/"))
        .respond_with(ResponseTemplate::new(200).set_body_json(raw_original_event.json()))
        .expect(1)
        .named("/event")
        .mount(&server)
        .await;

    let new_content = RoomMessageEventContentWithoutRelation::text_plain("uh i mean bonjour");

    let room = client.get_room(room_id).unwrap();

    // make_edit_event works, even if the event cache hasn't been enabled.
    room.make_edit_event(resp_event_id, EditedContent::RoomMessage(new_content)).await.unwrap();
}
```

To that code:

```rust
#[async_test]
async fn test_make_reply_event_doesnt_require_event_cache() {
    let mock = MatrixMockServer::new().await;
    let user_id = mock.client().user_id().unwrap().to_owned();

    let room_id = room_id!("!galette:saucisse.bzh");
    let room = mock.sync_joined_room(room_id).await;

    let event_id = event_id!("$1");
    let f = EventFactory::new();
    mock.mock_room_event()
        .ok(f.text_msg("hi").event_id(event_id).sender(&user_id).room(room_id).into_timeline())
        .expect(1)
        .named("/event")
        .mount()
        .await;

    let new_content = RoomMessageEventContentWithoutRelation::text_plain("uh i mean bonjour");

    // make_edit_event works, even if the event cache hasn't been enabled.
    room.make_edit_event(event_id, EditedContent::RoomMessage(new_content)).await.unwrap();
}
```

### DIFF
--- a/crates/matrix-sdk/src/test_utils/mocks.rs
+++ b/crates/matrix-sdk/src/test_utils/mocks.rs
@@ -34,11 +34,8 @@ use wiremock::{
 use super::logged_in_client;
 use crate::{Client, Room};
 
-/// A `wiremock` server along with a client connected to it, with useful methods
-/// to help mocking Matrix client-server API endpoints easily.
-///
-/// This is a pair of a [`MockServer`] and a [`Client]` (one can retrieve them
-/// respectively with [`Self::server()`] and [`Self::client()`]).
+/// A `wiremock` [`MockServer`] along with useful methods to help mocking Matrix
+/// client-server API endpoints easily.
 ///
 /// It implements mock endpoints, limiting the shared code as much as possible,
 /// so the mocks are still flexible to use as scoped/unscoped mounts, named, and
@@ -63,7 +60,6 @@ use crate::{Client, Room};
 ///   mostly defers its implementations to [`wiremock::Mock`] under the hood.
 pub struct MatrixMockServer {
     server: MockServer,
-    client: Client,
 
     /// Make the sync response builder stateful, to keep in memory the batch
     /// token and avoid the client ignoring subsequent responses after the first
@@ -75,19 +71,19 @@ impl MatrixMockServer {
     /// Create a new `wiremock` server specialized for Matrix usage.
     pub async fn new() -> Self {
         let server = MockServer::start().await;
-        let client = logged_in_client(Some(server.uri().to_string())).await;
-        Self { client, server, sync_response_builder: Default::default() }
+        Self { server, sync_response_builder: Default::default() }
     }
 
     /// Creates a new [`MatrixMockServer`] when both parts have been already
     /// created.
-    pub fn from_parts(server: MockServer, client: Client) -> Self {
-        Self { client, server, sync_response_builder: Default::default() }
+    pub fn from_server(server: MockServer) -> Self {
+        Self { server, sync_response_builder: Default::default() }
     }
 
-    /// Return the underlying client.
-    pub fn client(&self) -> Client {
-        self.client.clone()
+    /// Creates a new [`Client`] configured to use this server, preconfigured
+    /// with a session expected by the server endpoints.
+    pub async fn make_client(&self) -> Client {
+        logged_in_client(Some(self.server.uri().to_string())).await
     }
 
     /// Return the underlying server.
@@ -98,11 +94,16 @@ impl MatrixMockServer {
     /// Overrides the sync/ endpoint with knowledge that the given
     /// invited/joined/knocked/left room exists, runs a sync and returns the
     /// given room.
-    pub async fn sync_room(&self, room_id: &RoomId, room_data: impl Into<AnyRoomBuilder>) -> Room {
+    pub async fn sync_room(
+        &self,
+        client: &Client,
+        room_id: &RoomId,
+        room_data: impl Into<AnyRoomBuilder>,
+    ) -> Room {
         let any_room = room_data.into();
 
         self.mock_sync()
-            .ok_and_run(move |builder| match any_room {
+            .ok_and_run(client, move |builder| match any_room {
                 AnyRoomBuilder::Invited(invited) => {
                     builder.add_invited_room(invited);
                 }
@@ -118,13 +119,13 @@ impl MatrixMockServer {
             })
             .await;
 
-        self.client.get_room(room_id).expect("look at me, the room is known now")
+        client.get_room(room_id).expect("look at me, the room is known now")
     }
 
     /// Overrides the sync/ endpoint with knowledge that the given room exists
     /// in the joined state, runs a sync and returns the given room.
-    pub async fn sync_joined_room(&self, room_id: &RoomId) -> Room {
-        self.sync_room(room_id, JoinedRoomBuilder::new(room_id)).await
+    pub async fn sync_joined_room(&self, client: &Client, room_id: &RoomId) -> Room {
+        self.sync_room(client, room_id, JoinedRoomBuilder::new(room_id)).await
     }
 
     /// Verify that the previous mocks expected number of requests match
@@ -144,7 +145,6 @@ impl MatrixMockServer {
             .and(header("authorization", "Bearer 1234"));
         MockSync {
             mock,
-            client: self.client.clone(),
             server: &self.server,
             sync_response_builder: self.sync_response_builder.clone(),
         }
@@ -360,7 +360,6 @@ pub struct MockSync<'a> {
     mock: MockBuilder,
     server: &'a MockServer,
     sync_response_builder: Arc<Mutex<SyncResponseBuilder>>,
-    client: Client,
 }
 
 impl<'a> MockSync<'a> {
@@ -368,7 +367,7 @@ impl<'a> MockSync<'a> {
     /// sync with it.
     ///
     /// After calling this function, the sync endpoint isn't mocked anymore.
-    pub async fn ok_and_run<F: FnOnce(&mut SyncResponseBuilder)>(self, func: F) {
+    pub async fn ok_and_run<F: FnOnce(&mut SyncResponseBuilder)>(self, client: &Client, func: F) {
         let json_response = {
             let mut builder = self.sync_response_builder.lock().unwrap();
             func(&mut builder);
@@ -381,7 +380,7 @@ impl<'a> MockSync<'a> {
             .mount_as_scoped(self.server)
             .await;
 
-        let _response = self.client.sync_once(Default::default()).await.unwrap();
+        let _response = client.sync_once(Default::default()).await.unwrap();
     }
 }
 
--- a/crates/matrix-sdk/src/test_utils/mocks.rs
+++ b/crates/matrix-sdk/src/test_utils/mocks.rs
@@ -24,10 +24,10 @@ use matrix_sdk_test::{
     test_json, InvitedRoomBuilder, JoinedRoomBuilder, KnockedRoomBuilder, LeftRoomBuilder,
     SyncResponseBuilder,
 };
-use ruma::{OwnedEventId, OwnedRoomId, RoomId};
+use ruma::{MxcUri, OwnedEventId, OwnedRoomId, RoomId};
 use serde_json::json;
 use wiremock::{
-    matchers::{header, method, path, path_regex},
+    matchers::{body_partial_json, header, method, path, path_regex},
     Mock, MockBuilder, MockGuard, MockServer, Respond, ResponseTemplate, Times,
 };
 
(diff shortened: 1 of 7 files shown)


crates/matrix-sdk/src/test_utils/mocks.rs
```diff
@@ -0,0 +1,389 @@
...
+    client: Client,
+
+    /// Make the sync response builder stateful, to keep in memory the batch
+    /// token and avoid the client ignoring subsequent responses after the first
+    /// one.
+    sync_response_builder: Arc<Mutex<SyncResponseBuilder>>,
+}
+
+impl MatrixMockServer {
+    /// Create a new mocked server specialized for Matrix usage.
```
andybalaam, line 45: "mocked server" -> "wiremock server"

crates/matrix-sdk/src/test_utils/mocks.rs
```diff
@@ -0,0 +1,389 @@
...
+    pub async fn new() -> Self {
+        let server = MockServer::start().await;
+        let client = logged_in_client(Some(server.uri().to_string())).await;
+        Self { client, server, sync_response_builder: Default::default() }
+    }
+
+    /// Creates a new [`MatrixMockServer`] when both parts have been already
+    /// created.
+    pub fn from_pair(server: MockServer, client: Client) -> Self {
```
andybalaam, line 54: `from_pair` -> `from_server_and_client` or `from_parts` ?

crates/matrix-sdk/src/test_utils/mocks.rs
```diff
@@ -0,0 +1,389 @@
...
+
+    /// Return the underlying server.
+    pub fn server(&self) -> &MockServer {
+        &self.server
+    }
+
+    /// Overrides the sync/ endpoint with knowledge that the given
+    /// invited/joined/knocked/left room exists, runs a sync and returns the
+    /// given room.
+    pub async fn sync_room(&self, room_id: &RoomId, room_data: impl Into<AnyRoomBuilder>) -> Room {
```
andybalaam, line 71: Trying to think of a more informative name for this...

Maybe `create_or_change_room` ? yuck.

crates/matrix-sdk/src/test_utils/mocks.rs
```diff
@@ -0,0 +1,389 @@
...
+        self.client.get_room(room_id).expect("look at me, the room is known now")
+    }
+
+    /// Overrides the sync/ endpoint with knowledge that the given room exists
+    /// in the joined state, runs a sync and returns the given room.
+    pub async fn sync_joined_room(&self, room_id: &RoomId) -> Room {
+        self.sync_room(room_id, JoinedRoomBuilder::new(room_id)).await
+    }
+
+    pub async fn verify_and_reset(&self) {
```
andybalaam, line 100: I personally have no idea what these methods do, so would be nice to have a comment.

crates/matrix-sdk/src/test_utils/mocks.rs
```diff
@@ -0,0 +1,389 @@
...
+}
+
+impl From<KnockedRoomBuilder> for AnyRoomBuilder {
+    fn from(val: KnockedRoomBuilder) -> AnyRoomBuilder {
+        AnyRoomBuilder::Knocked(val)
+    }
+}
+
+/// A wiremock Mock wrapping also the server, to not have to pass it around when
+/// mounting.
```
andybalaam, line 196: I don't understand this. Can you find a different way to write it? I think maybe it's because I don't know much about wiremock.

andybalaam (commented):
This is great. Passing on to @richvdh because a) I think he will have helpful feedback and b) I'm not available next week.

richvdh (commented):
some thoughts. Looks like a great improvement.

crates/matrix-sdk/src/test_utils/mocks.rs
bnjbvr: I like `from_parts`, thanks

crates/matrix-sdk/src/test_utils/mocks.rs
bnjbvr: `mock_and_run_sync_for_room`? :woozy_face:

crates/matrix-sdk/src/test_utils/mocks.rs
```diff
@@ -0,0 +1,389 @@
...
+    }
+
+    pub async fn verify_and_reset(&self) {
+        self.server.verify().await;
+        self.server.reset().await;
+    }
+}
+
+// Specific mount endpoints.
+impl MatrixMockServer {
```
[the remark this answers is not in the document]
bnjbvr: I've seen this in previous code bases in the past: some use different impl blocks to distinguish public vs non-public methods, for instance — or group methods by role, as done here.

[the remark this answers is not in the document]
bnjbvr: Note it's possible to see all the impl blocks for a given type using rust-analyzer, I type "jump to implementations" (same shortcut used when jumping from a trait to its implementations, for instance).

crates/matrix-sdk/src/test_utils/mocks.rs
```diff
@@ -0,0 +1,389 @@
...
+    matchers::{header, method, path, path_regex},
+    Mock, MockBuilder, MockGuard, MockServer, Respond, ResponseTemplate, Times,
+};
+
+use super::logged_in_client;
+use crate::{Client, Room};
+
+pub struct MatrixMockServer {
+    server: MockServer,
+    client: Client,
```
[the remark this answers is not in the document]
bnjbvr: Passing the client upon construction of the mock server is a prerequisite, so the server's URI is injected into the client, at construction; reverting this would make construction a bit unwieldy, as one would have to pass a `ClientBuilder`, or get the URL from the `MatrixMockServer` and use it to create a `Client`, leading to more duplicated code for all callers.

Happy to rename this struct overall to something else, though? Not sure what a good name would be…

bnjbvr:
Should be ready for another round of review! I'm planning to use this for writing more tests for media uploads.

crates/matrix-sdk/src/test_utils/mocks.rs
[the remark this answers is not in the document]
richvdh: > Note it's possible to see all the impl blocks for a given type using rust-analyzer, I type "jump to implementations" (same shortcut used when jumping from a trait to its implementations, for instance).

Yeahhh, but I rather like my code to be legible without an IDE (particularly, when I'm reviewing code in Github).

I'm not convinced I like this, but I'll defer to your preference on it.

crates/matrix-sdk/src/test_utils/mocks.rs
```diff
@@ -0,0 +1,514 @@
...
+use serde_json::json;
+use wiremock::{
+    matchers::{body_partial_json, header, method, path, path_regex},
+    Mock, MockBuilder, MockGuard, MockServer, Respond, ResponseTemplate, Times,
+};
+
+use super::logged_in_client;
+use crate::{Client, Room};
+
+/// A `wiremock` server along a client connected it, with useful methods to help
```
richvdh, line 37:
```suggestion
/// A `wiremock` server along with a client connected it, with useful methods to help
```

crates/matrix-sdk/src/test_utils/mocks.rs
```diff
@@ -0,0 +1,514 @@
...
+/// mocking Matrix client-server API endpoints easily.
+///
+/// This is a pair of a MockServer and a Client (one can retrieve them
```
richvdh, line 40:
```suggestion
/// This is a pair of a [`MockServer`] and a [`Client`] (one can retrieve them
```
possibly?

crates/matrix-sdk/src/test_utils/mocks.rs
```diff
@@ -0,0 +1,514 @@
...
+/// respectively with [`Self::server()`] and [`Self::client()`]).
+///
+/// It implements mock endpoints, expanding and limited the shared code as much
```
richvdh, line 43: I don't know what you mean by "expanding and limit[ing]". Aren't the two kinda opposite?

crates/matrix-sdk/src/test_utils/mocks.rs
```diff
@@ -0,0 +1,514 @@
...
+/// as possible, so the mocks are still flexible to use as scoped/unscoped
+/// mounts, named, and so on.
+///
+///It works like this:
```
richvdh, line 47:
```suggestion
/// It works like this:
```

crates/matrix-sdk/src/test_utils/mocks.rs
```diff
@@ -0,0 +1,514 @@
...
+///
+/// - start with telling which endpoint you'd like to mock, e.g.
```
richvdh, line 49:
```suggestion
/// - start by saying which endpoint you'd like to mock, e.g.
```

["telling" is transitive: you have to say *who* you are telling about a thing.]

crates/matrix-sdk/src/test_utils/mocks.rs
```diff
@@ -0,0 +1,514 @@
...
+        self.server.reset().await;
+    }
+}
+
+// Specific mount endpoints.
+impl MatrixMockServer {
+    /// Mocks a sync endpoint.
+    pub fn mock_sync(&self) -> MockSync<'_> {
+        let mock = Mock::given(method("GET"))
+            .and(path("/_matrix/client/r0/sync"))
```
richvdh, line 143: aside: shouldn't we be using `/v3/` endpoints? `/r0/` is supported by synapse for compatibility with old applications, but new codebases shouldn't be using it.

crates/matrix-sdk/src/test_utils/mocks.rs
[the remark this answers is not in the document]
bnjbvr: Yeah, I found it quite useful, and since there are chances I'll be hacking more on this file…  :-) Resolving this comment.

crates/matrix-sdk/src/test_utils/mocks.rs
bnjbvr: ogod, missing so many words in this sentence…

crates/matrix-sdk/src/test_utils/mocks.rs
bnjbvr: Yeah the default Client is configured with no information about matrix versions, so it supposes the first version by default. It usually doesn't matter, except in the few cases where it does, of course. I'll ask about this in the Rust room, but i'm tempted to keep it like this right now.

poljar: In reality we do use the `/v3` endpoints, but in tests we don't bother filling out the supported version in the `Client` and as @bnjbvr says, it falls back to a default version.

We could modify the constructor of the test `Client` object to have the versions prefilled, but that will automatically mean that you need to update all tests that are using this test client.

I would say that we should do that, but not as part of this PR.

PR #4198, opened 2024-10-31
doc: start an architecture document with a high-level description of the crates

bnjbvr:
Based on reading the https://matklad.github.io/2021/02/06/ARCHITECTURE.md.html blog post, @andybalaam and I started an architecture document explaining at a high-level what the different crates are. We hope to also include the most important high-level components in the mix, as a vague map of where things are implemented. (e.g. Timeline, SyncService, RoomListService, etc.). Very WIP-y, etc.

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -23,17 +23,40 @@ crypto    /
       common (matrix-sdk-common)
 ```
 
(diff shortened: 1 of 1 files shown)


stefanceriu (approved):
Very nice, thank you for setting this up!

PR #4199, opened 2024-10-31
chore(base): refactor internal helpers related to media

bnjbvr:
Notably, make it super clear what parameters are required to create the attachment type, since the function doesn't consume the whole `AttachmentConfig` for realz.

Part of #1732, blocks #4195.


Hywan (approved):
LGTM! Thanks.

PR #4200, opened 2024-10-31
refactor(send queue): generalize stored send queue data

bnjbvr:
See the first commit message, which explains the whole point.

Because of this PR, the send queue events and dependent events tables are both cleared, because there's a change in the format of serde-serialized data, and I didn't bother to write a migration for those. Events stuck in there would only be wedged events, or events that were still in the queue before closing the app; worst case scenario is some users lose some messages they tried to send before an update of an app. Is that OK? otherwise, can look into a sane migration, if needs be.

Part of #1732, blocks #4195.


poljar (approved):
This looks good, but since you removed the migration test, do we have more migration tests that will run and check if everything is fine?

Also, do we want to double check on the app side that this indeed won't break in some unexpected way?

bnjbvr:
Thanks for the review. As discussed in chat, I'll add back the migration test, and will make sure it checks that the content has now been cleared in the send queue, so we have some proof this is going to happen.

PR #4205, opened 2024-11-04
fix(pinned_events): get pinned event ids from the HS as fallback

jmartinesp:
When the required state of a room doesn't contain pinned events info for some reason, try using the `/state` endpoint from the HS to get this data instead.

This should take care of a bug that caused pinned events to be incorrectly removed when the new pinned event ids list was based on an empty one if the required state of the room didn't contain any pinned events info

Signed-off-by:


Hywan (approved):
LGTM if the tests are passing, thanks!

jmartinesp:
I had to fix the tests separately 😅 . Will merge once the tests are green.

PR #4207, opened 2024-11-04
chore(ui): Display the real error of `Error::EventCache`

Hywan:
This patch displays the wrapped error.


bnjbvr (changes requested):
I would even go one step further and mark these as `#[error(transparent)`, `InitializingTimeline` and the other `SlidingSync(SlidingSyncError)` too.

Hywan:
Indeed, done.

bnjbvr (approved):

PR #4208, opened 2024-11-04
feat(ui): Remove `RoomListService::new_with_encryption`

Hywan:
This patch removes `RoomListService::new_with_encryption`. This feature is not used, not useful since it's best to use `EncryptionSyncService`, and it can be racy depending on how it's used. To avoid potential errors and bugs, it's preferable to remove this code.


Hywan:
I want to merge it before because there is some necessary rebasing anyway, it's just an opportunity to simplify the work.

PR #4211, opened 2024-11-04
feat(ui): Add `m.room.tombstone` to the room list `required_state`

Hywan:
This patch adds the `m.room.tombstone` state event to the list of events in `required_state` used by the `RoomListService`. The goal is to offer the possibility for the consumers to know whether a room has been tombstoned or not.

### DIFF
--- a/crates/matrix-sdk-ui/src/room_list_service/mod.rs
+++ b/crates/matrix-sdk-ui/src/room_list_service/mod.rs
@@ -88,6 +88,7 @@ const DEFAULT_REQUIRED_STATE: &[(StateEventType, &str)] = &[
     (StateEventType::RoomCanonicalAlias, ""),
     (StateEventType::RoomPowerLevels, ""),
     (StateEventType::CallMember, "*"),
+    (StateEventType::RoomTombstone, ""),
 ];
 
 /// The default `required_state` constant value for sliding sync room
--- a/crates/matrix-sdk-ui/tests/integration/room_list_service.rs
+++ b/crates/matrix-sdk-ui/tests/integration/room_list_service.rs
@@ -331,6 +331,7 @@ async fn test_sync_all_states() -> Result<(), Error> {
                         ["m.room.canonical_alias", ""],
                         ["m.room.power_levels", ""],
                         ["org.matrix.msc3401.call.member", "*"],
+                        ["m.room.tombstone", ""],
                     ],
                     "include_heroes": true,
                     "filters": {
@@ -2100,6 +2101,7 @@ async fn test_room_subscription() -> Result<(), Error> {
                         ["m.room.canonical_alias", ""],
                         ["m.room.power_levels", ""],
                         ["org.matrix.msc3401.call.member", "*"],
+                        ["m.room.tombstone", ""],
                         ["m.room.create", ""],
                         ["m.room.pinned_events", ""],
                     ],
@@ -2138,6 +2140,7 @@ async fn test_room_subscription() -> Result<(), Error> {
                         ["m.room.canonical_alias", ""],
                         ["m.room.power_levels", ""],
                         ["org.matrix.msc3401.call.member", "*"],
+                        ["m.room.tombstone", ""],
                         ["m.room.create", ""],
                         ["m.room.pinned_events", ""],
                     ],


kevinaboos:
Echoing my [comments from the Matrix Rust SDK room](https://matrix.to/#/!iYnZafYUoXkeVPOSQh:matrix.org/$Iydu40UMRjCqqt0Yyl8sjrSG-bydpnh8QdfYTCbw59U?via=matrix.org&via=element.io&via=envs.net) here, as requested:

I tested this PR out in Robrix, but unfortunately was not able to get any tombstoned info at all. Here's what I'm doing, just in case I missed a key step.

When I receive a diff with a new room from the room list service, i then attempt to call several different tombstone-related functions (which I think are all doing mostly the same thing under the hood):
- `room.tombstone()` always returns `None`.
- `room.get_state_event_static::<RoomTombstoneEventContext>().await` also always returns `None`.
- `room.get_state_events(StateEventType::RoomTombstone).await()` always returns an empty vec.

i know for a fact that I do have tombstoned rooms on the account that I'm testing with, so I'm expecting them to have tombstoned room state. Here's a side-by-side example showing how both rooms exist as separate rooms in the same session: the new version of Element Web/Desktop on the left, and the old tombstoned one on the right.
image

poljar:
What's the state of this PR? Do we need this or do we expect users to enable tombstone events themselves?

> When I receive a diff with a new room from the room list service, i then attempt to call several different tombstone-related functions

Are you sure that the server sent the event to you? As far as I can see, we don't treat those events any differently from other important room state events, if we receive it we put it in the cache.

kevinaboos:
> What's the state of this PR? Do we need this or do we expect users to enable tombstone events themselves?

Hi @poljar, thanks for the response. I would argue that tombstone state should be a part of the sync service's required state, since without it, the client may accidentally display multiple old tombstoned rooms as separate unrelated rooms without being able to properly determine that they're replacements of one another.

> > When I receive a diff with a new room from the room list service, i then attempt to call several different tombstone-related functions
> 
> Are you sure that the server sent the event to you? As far as I can see, we don't treat those events any differently from other important room state events, if we receive it we put it in the cache.

I'm not sure, no. I think that's part of the problem -- even with this change, I don't get any tombstone events from any rooms.

How might I accurately determine whether I've received the tombstone events from the server? (besides using the existing SDK APIs for this, like  https://matrix-org.github.io/matrix-rust-sdk/matrix_sdk/struct.BaseRoom.html#method.tombstone)

poljar:
> How might I accurately determine whether I've received the tombstone events from the server? (besides using the existing SDK APIs for this, like https://matrix-org.github.io/matrix-rust-sdk/matrix_sdk/struct.BaseRoom.html#method.tombstone)

I like to use [`mitmproxy`](https://mitmproxy.org/) for this then you can observe the full `/sync ` response.

Hywan:
@erikjohnston Do you know if sliding sync on synapse filters out the `m.room.tombstone` events by any luck, or mess with them somehow?

Hywan:
I wonder if this can be solved by https://github.com/element-hq/synapse/issues/17540.

Hywan:
Thanks for the test. Things have been resolved by themselves apparently, great news!

jmartinesp (approved):
The changes LGTM, but it seems like something broke the CI tests?

Hywan:
Blocked by https://github.com/matrix-org/matrix-rust-sdk/pull/5048.

kevinaboos:
Excellent, thanks for merging this! I will test it when I get back from vacation.

Hywan:
I'm about to update the room list filters to filter out tombstoned rooms. The idea is to provide a better “native” support for these rooms.

kevinaboos:
I know I'm very late on the response, but I just wanted to confirm that this did fix the problem for us, so thanks very much! We now properly observe tombstoned rooms in Robrix:
```txt
[log] - Room !QQpfJfZvqxbCfeDgCj:matrix.org has been tombstoned:
SuccessorRoom {
    room_id: "!GNPBRmjZKKEGszhMtB:matrix.org",
    reason: Some(
        "This room has been replaced",
    ),
}
```

Hywan:
Great! You may be interested by [`new_filter_deduplicate_versions`](https://matrix-org.github.io/matrix-rust-sdk/matrix_sdk_ui/room_list_service/filters/fn.new_filter_deduplicate_versions.html) for the `RoomList`.

PR #4212, opened 2024-11-04
doc(ui): Clarify how to obtain `room_info_notable_update_receiver` for `RoomList::entries_with_dynamic_adapters()`

kevinaboos:
When migrating to the new `RoomList::entries_with_dynamic_adapters()`, I spent an embarrassingly long time trying to figure out what I should pass in for the `room_info_notable_update_receiver` argument and where to obtain it. This PR clarifies that with a direct link to the `Client` function that provides the receiver.

Signed-off-by: Kevin Boos <kevinaboos@gmail.com>

### DIFF
--- a/crates/matrix-sdk-ui/src/room_list_service/room_list.rs
+++ b/crates/matrix-sdk-ui/src/room_list_service/room_list.rs
@@ -136,6 +136,9 @@ impl RoomList {
     /// entries, and that it's also possible to “paginate” over the entries by
     /// `page_size`. The rooms are also sorted.
     ///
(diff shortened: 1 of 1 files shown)


bnjbvr:
Indeed, thanks for the PR; I think we ought to get rid of this parameter even, since we always have access to the underlying `Client` (and it's never used with a different client instance). Let me try to cook something up.

bnjbvr:
That's it, the argument is now removed :)

kevinaboos:
That works too! Thanks.

PR #4214, opened 2024-11-05
refactor!(room list): remove unneeded argument from `RoomList::entries_with_dynamic_adapters`

bnjbvr:
The parameter `room_info_notable_update_receiver` was removed
 from `RoomList::entries_with_dynamic_adapters`, since it could be
 inferred internally instead.

Alternative to #4212, which I think avoids the problem altogether by providing a better API.


Hywan (approved):
Good to me! Thanks for the improvements.

PR #4215, opened 2024-11-05
chore(timeline): rename `TimelineItemPosition::Update` to `UpdateDecrypted`

bnjbvr:
What it says on the tin. There were ideas to also rename `TimelineItemPosition`, but we're fine with the status quo, until we get more progress on the event cache.


bnjbvr:
This has been updated to use a simpler name without renaming the whole enum type, and as such this is ready for review.

poljar (approved):

PR #4216, opened 2024-11-05
fix(sdk): Do not always remove empty chunks from `LinkedChunk`

Hywan:
This patch introduces `EmptyChunk`, a new enum used to represent whether empty chunks must be removed/unlink or kept from the `LinkedChunk`. It is used by `LinkedChunk::remove_item_at`.

Why is it important? For example, imagine the following situation:

- one inserts a single event in a new chunk (possible if a (sliding) sync is done with `timeline_limit=1`),
- one inserts many events at the position of the previous event, with one of the new events being a duplicate of the first event (possible if a (sliding) sync is done with `timeline_limit=10` this time),
- prior to this patch, the older event was removed, resulting in an empty chunk, which was removed from the `LinkedChunk`, invalidating the insertion position!

So, with this patch:

- `RoomEvents::remove_events` does remove empty chunks, but
- `RoomEvents::remove_events_and_update_insert_position` does NOT remove empty chunks, they are kept in case the position wants to insert in this same chunk.

### DIFF
--- a/crates/matrix-sdk/src/event_cache/linked_chunk/mod.rs
+++ b/crates/matrix-sdk/src/event_cache/linked_chunk/mod.rs
@@ -408,9 +408,18 @@ impl<const CAP: usize, Item, Gap> LinkedChunk<CAP, Item, Gap> {
 
     /// Remove item at a specified position in the [`LinkedChunk`].
     ///
-    /// Because the `position` can be invalid, this method returns a
-    /// `Result`.
-    pub fn remove_item_at(&mut self, position: Position) -> Result<Item, Error> {
+    /// `position` must point to a valid item, otherwise the method returns
+    /// `Err`.
+    ///
+    /// The chunk containing the item represented by `position` may be empty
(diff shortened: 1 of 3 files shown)


crates/matrix-sdk/src/event_cache/linked_chunk/mod.rs
```diff
@@ -410,7 +410,11 @@ impl<const CAP: usize, Item, Gap> LinkedChunk<CAP, Item, Gap> {
     ///
     /// Because the `position` can be invalid, this method returns a
     /// `Result`.
-    pub fn remove_item_at(&mut self, position: Position) -> Result<Item, Error> {
+    pub fn remove_item_at(
+        &mut self,
+        position: Position,
+        empty_chunk: EmptyChunk,
```
poljar, line 416: We should probably update the docs to let people know what the new argument is about. I know `EmptyChunk` contains docs for this as well, but I would expect a small sentence here as well.

poljar (approved):
Makes sense, left a small nit.

PR #4219, opened 2024-11-05
feat(ffi): add room alias format validation & room name to alias

jmartinesp:
Add two FFI functions:

`is_room_alias_format_valid`: to check if a passed string has a valid room alias format in Ruma.
`room_alias_from_room_display_name`: to get a suggestion of a room alias to use given a room name.

Signed-off-by:

### DIFF
--- a/bindings/matrix-sdk-ffi/src/room_alias.rs
+++ b/bindings/matrix-sdk-ffi/src/room_alias.rs
@@ -1,0 +1,7 @@
+use ruma::RoomAliasId;
+
+/// Verifies the passed `String` matches the expected room alias format.
+#[matrix_sdk_ffi_macros::export]
+fn is_room_alias_format_valid(alias: String) -> bool {
+    RoomAliasId::parse(alias).is_ok()
+}
--- a/bindings/matrix-sdk-ffi/src/room_alias.rs
+++ b/bindings/matrix-sdk-ffi/src/room_alias.rs
@@ -1,7 +1,14 @@
+use matrix_sdk::DisplayName;
 use ruma::RoomAliasId;
 
 /// Verifies the passed `String` matches the expected room alias format.
 #[matrix_sdk_ffi_macros::export]
 fn is_room_alias_format_valid(alias: String) -> bool {
     RoomAliasId::parse(alias).is_ok()
+}
+
+/// Transforms a Room's display name into a valid room alias name.
+#[matrix_sdk_ffi_macros::export]
+fn room_alias_name_from_room_display_name(room_name: String) -> String {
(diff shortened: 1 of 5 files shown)


bindings/matrix-sdk-ffi/src/room_alias.rs
```diff
@@ -0,0 +1,22 @@
...
+/// Verifies the passed `String` matches the expected room alias format.
+#[matrix_sdk_ffi_macros::export]
+fn is_room_alias_format_valid(alias: String) -> bool {
+    RoomAliasId::parse(alias).is_ok()
+}
+
+/// Transforms a Room's display name into a valid room alias.
+#[matrix_sdk_ffi_macros::export]
+fn room_alias_from_room_display_name(room_name: String) -> String {
+    let whitespace_regex = Regex::new(r"\s+").unwrap();
```
poljar, line 13: Let's put those strings into a const at the top of the function and let's then document what each of those do.

Additionally, let's replace the `unwrap()` with an `expect()`.

bindings/matrix-sdk-ffi/src/room_alias.rs
```diff
@@ -0,0 +1,22 @@
...
+
+/// Verifies the passed `String` matches the expected room alias format.
+#[matrix_sdk_ffi_macros::export]
+fn is_room_alias_format_valid(alias: String) -> bool {
+    RoomAliasId::parse(alias).is_ok()
+}
+
+/// Transforms a Room's display name into a valid room alias.
+#[matrix_sdk_ffi_macros::export]
+fn room_alias_from_room_display_name(room_name: String) -> String {
```
poljar, line 12: Some simple smoke tests for this function would be nice.

bindings/matrix-sdk-ffi/src/room_alias.rs
```diff
@@ -0,0 +1,22 @@
+use regex::Regex;
+use ruma::RoomAliasId;
+
+/// Verifies the passed `String` matches the expected room alias format.
+#[matrix_sdk_ffi_macros::export]
+fn is_room_alias_format_valid(alias: String) -> bool {
+    RoomAliasId::parse(alias).is_ok()
+}
+
+/// Transforms a Room's display name into a valid room alias.
```
poljar, line 10: I wonder if we should implement this on the `DisplayName` type, seem like it might be generally useful.

poljar (approved):
Left some nits, looks sane.

jmartinesp:
I implemented all those suggestions and some changes to how the room name -> room alias transform works in the last commit.

PR #4220, opened 2024-11-05
feat(ffi): allow `VerificationStateListener` to emit the current state

jmartinesp:
With this, we get notified of the current verification state almost immediately.

Without it, you may either call it too soon and receive an `Unknown` state or you might have to call `Encryption::wait_for_e2ee_initialization_tasks()` and wait until it's finished to request a valid state value.

Also, make sure we get an updated verification value before any network request runs so it can't get deadlocked/time out and we end up not receiving the updated value.

Signed-off-by:

### DIFF
--- a/bindings/matrix-sdk-ffi/src/encryption.rs
+++ b/bindings/matrix-sdk-ffi/src/encryption.rs
@@ -397,14 +397,8 @@ impl Encryption {
     pub fn verification_state_listener(
         self: Arc<Self>,
         listener: Box<dyn VerificationStateListener>,
-        emit_current_value: bool,
     ) -> Arc<TaskHandle> {
         let mut subscriber = self.inner.verification_state();
-
-        if emit_current_value {
-            // Emit current value first
-            listener.on_update(subscriber.get().into());
-        }
 
         Arc::new(TaskHandle::new(RUNTIME.spawn(async move {
             while let Some(verification_state) = subscriber.next().await {
--- a/bindings/matrix-sdk-ffi/src/encryption.rs
+++ b/bindings/matrix-sdk-ffi/src/encryption.rs
@@ -397,8 +397,15 @@ impl Encryption {
     pub fn verification_state_listener(
         self: Arc<Self>,
         listener: Box<dyn VerificationStateListener>,
+        emit_current_value: bool,
     ) -> Arc<TaskHandle> {
         let mut subscriber = self.inner.verification_state();
+
+        if emit_current_value {
+            // Emit current value first
+            listener.on_update(subscriber.get().into());
+        }
+
         Arc::new(TaskHandle::new(RUNTIME.spawn(async move {
             while let Some(verification_state) = subscriber.next().await {
                 listener.on_update(verification_state.into());
--- a/crates/matrix-sdk/src/encryption/mod.rs
+++ b/crates/matrix-sdk/src/encryption/mod.rs
@@ -774,7 +774,7 @@ impl Encryption {
     /// # anyhow::Ok(()) };
     /// ```
     pub fn verification_state(&self) -> Subscriber<VerificationState> {
-        self.client.inner.verification_state.subscribe()
+        self.client.inner.verification_state.subscribe_reset()
     }
 
     /// Get a verification object with the given flow id.
@@ -1698,7 +1698,14 @@ impl Encryption {
 
 #[cfg(all(test, not(target_arch = "wasm32")))]
 mod tests {
-    use std::time::Duration;
+    use std::{
+        ops::Not,
+        sync::{
+            atomic::{AtomicBool, Ordering},
+            Arc,
+        },
+        time::Duration,
+    };
 
     use matrix_sdk_base::SessionMeta;
     use matrix_sdk_test::{
@@ -1713,13 +1720,15 @@ mod tests {
     use serde_json::json;
     use wiremock::{
         matchers::{header, method, path_regex},
-        Mock, MockServer, ResponseTemplate,
+        Mock, MockServer, Request, ResponseTemplate,
     };
 
     use crate::{
+        assert_next_matches_with_timeout,
         config::RequestConfig,
+        encryption::VerificationState,
         matrix_auth::{MatrixSession, MatrixSessionTokens},
-        test_utils::logged_in_client,
+        test_utils::{logged_in_client, no_retry_test_client, set_client_session},
         Client,
     };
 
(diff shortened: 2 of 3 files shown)


jmartinesp:
FWIW, I think all listeners should emit the initial values they have, but since that would modify existing behaviours I followed @pixlwave 's suggestion and added the `emit_current_value` parameter.

poljar (approved):
A regression test for the first commit would be nice. 

I know I know, kinda annoying but if we add more initialization tasks it would ensure that we don't add tasks that require network connectivity before the update of the `VerificationState` and regress again.

jmartinesp:
> A regression test for the first commit would be nice.

Sorry, since you mention the initialisation tasks I'm not sure if you mean the actual first commit (the FFI layer one) or the one in the SDK crate, can you confirm which one is it?

jmartinesp:
Also, if it's the FFI one, do we run tests on the FFI layer in the CI? I thought we didn't.

jmartinesp:
@poljar: since we need this bugfix for Android and you're away today I'm going to merge this as is, but I pinky promise to write a test once I understand what I should be testing 😅 .

bindings/matrix-sdk-ffi/src/encryption.rs
```diff
@@ -397,8 +397,15 @@ impl Encryption {
     pub fn verification_state_listener(
         self: Arc<Self>,
         listener: Box<dyn VerificationStateListener>,
+        emit_current_value: bool,
     ) -> Arc<TaskHandle> {
         let mut subscriber = self.inner.verification_state();
```
bnjbvr, line 402: (You could also have `verification_state` call `subscribe_reset()` in place of `subscribe()`, and not have any change here — that would make it easier to test this at the SDK layer too)

jmartinesp: I'll upload some changes doing this once I've tested them in the Android client.

bindings/matrix-sdk-ffi/src/encryption.rs
```diff
@@ -399,6 +399,10 @@ impl Encryption {
         listener: Box<dyn VerificationStateListener>,
     ) -> Arc<TaskHandle> {
         let mut subscriber = self.inner.verification_state();
+
+        // Reset the subscriber so the current observer value is emitted to the listener
+        subscriber.reset();
```
bnjbvr, line 404: No, that's not what I meant: I meant calling `subscribe_reset()` in place of `subscribe()` here: https://github.com/matrix-org/matrix-rust-sdk/blob/8865e2ff7463ff0b520f53f1a3ffc5500fbb3d86/crates/matrix-sdk/src/encryption/mod.rs#L777

But @poljar might have a different opinion when back, and I'm not clear about the implications of doing that. Let's wait for him to come back (i.e. revert to the previous state).

jmartinesp: Ok, reverted the fixup commit and left it as it was before. Let's wait for him.

bindings/matrix-sdk-ffi/src/encryption.rs
poljar: > Removed that commit, went back to what we had before.

Why, what went wrong.

> You could also have verification_state call subscribe_reset()

Ah, that's the function I was looking for, not `next_now()`, thanks for reminding us.

bindings/matrix-sdk-ffi/src/encryption.rs
poljar: Yes, I think using `subscribe_reset()` would be the correct approach here, I think that re-implementing `subscribe_reset()` using `get()` and `subscribe()` might introduce a small race where we might miss an update.

poljar:
> @poljar: since we need this bugfix for Android and you're away today I'm going to merge this as is, but I pinky promise to write a test once I understand what I should be testing 😅 .

The test should check that we can get the correct `VerificationState` despite the requests that are part of the start up tasks have not finished.

crates/matrix-sdk/src/test_utils.rs
```diff
@@ -105,6 +105,7 @@ pub async fn logged_in_client_with_server() -> (Client, wiremock::MockServer) {
 #[macro_export]
 macro_rules! assert_next_with_timeout {
     ($stream:expr, $timeout_ms:expr) => {{
+        #[allow(unused_imports)]
```
poljar, line 108: Why's that suddenly needed?

crates/matrix-sdk/src/encryption/mod.rs
```diff
@@ -2017,4 +2027,39 @@ mod tests {
...
+
+        // We can get its initial value, and it's Unknown
+        assert_next_matches_with_timeout!(verification_state, VerificationState::Unknown);
+
+        // We set up a mocked request to check this endpoint is not called before
+        // reading the new state
+        let keys_requested = Arc::new(AtomicU8::new(0));
+        let inner_counter = keys_requested.clone();
+        Mock::given(method("GET"))
+            .and(path_regex(
```
poljar, line 2048: You can add `set_delay()` for this mock to make it a bit more certain that we'll not respond in time: https://docs.rs/wiremock/latest/wiremock/struct.ResponseTemplate.html#method.set_delay

poljar (approved):
Yup, looks great, left two nits still but I think that's good to go after that.

crates/matrix-sdk/src/encryption/mod.rs
jmartinesp: But that just makes the actual response be slower, right? The code changing the counter would still be run before the delay.

crates/matrix-sdk/src/test_utils.rs
jmartinesp: Without it I get:

```
warning: unused import: `futures_util::StreamExt`
    --> crates/matrix-sdk/src/test_utils.rs:108:13
     |
108  |         use futures_util::StreamExt as _;
     |             ^^^^^^^^^^^^^^^^^^^^^^^
     |
    ::: crates/matrix-sdk/src/encryption/mod.rs:1735:9
     |
1735 |         assert_next_matches_with_timeout!(verification_state, VerificationState::Unknown);
     |         --------------------------------------------------------------------------------- in this macro invocation
```

When I try to use the macro in the new test 🫤 .

poljar: Hmm, Ok.

crates/matrix-sdk/src/encryption/mod.rs
poljar: Ah duh, I just understood what you said, nevermind with that setting then.

PR #4224, opened 2024-11-06
feat(sdk): Add `Client:cross_process_store_locks_holder_name()`

Hywan:
We use cross-process store locks in 2 places now: in crypto, and in the event cache. The idea of this PR is to create a common place to define the cross-process store locks holder name.

This PR is twofold. First it creates `Client::cross_process_store_locks_holder_name()`, and second it uses this value for the `EventCacheStoreLock`.
This PR **does not** modify the crypto store lock for the moment. I don't know if it has to use the default/common holder name for the moment or not (ping @poljar, @pixlwave).


jmartinesp (commented):
LGTM, but we should wait until someone can verify if we need the 'default' name.

crates/matrix-sdk/src/client/builder/mod.rs
```diff
@@ -770,6 +770,7 @@ pub(crate) mod tests {
     use assert_matches::assert_matches;
     use matrix_sdk_test::{async_test, test_json};
     use serde_json::{json_internal, Value as JsonValue};
+    #[cfg(feature = "experimental-sliding-sync")]
```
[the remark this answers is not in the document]
Hywan: Yes, otherwise we get a warning if this flag is turned off.

bnjbvr:
>  This PR does not modify the crypto store lock for the moment. I don't know if it has to use the default/common holder name for the moment or not.

On iOS, the holder has to be different across the two different processes. This is the value that allows the app deciding who's holding onto the lock; so the main app would have to use one value, and any extension (notification service extension aka NSE, share extension) would have to use a different value uniquely identifying the service.

(The cross-process lock key has to be the same across all processes, though — otherwise processes will coordinate using different entries in the database.)

Hywan:
@bnjbvr Yes, I'm aware of that, but I'm wondering if we should use the holder name from the `Client` or let the actual as it is. Looking at https://github.com/element-hq/element-x-ios/blob/10966ab620929517394ae1a4ceb9a458ff876c18/NSE/Sources/Other/NSEUserSession.swift#L31-L48, there are using `NotificationClient` directly, which uses `Client::notification_client`. I don't see where a new holder name is set for the cross-process lock of the crypto store.
Anyway, now on, the `Client` can setup a cross-process holder name, which can be passed to the `NotificationClient` and will be used everywhere, except by the crypto store for the moment.

In ` crates/matrix-sdk-crypto/src/store/mod.rs`, we have `Store::create_store_lock`, which is used by OIDC and `Encryption::enable_cross_process_store_lock`. The latter is used only by `EncryptionSyncService::new`. And this one is used only by `NotificationClient` which pass its `LOCK_ID` constant. Sooooo. I think we need a little bit more refactoring on this one. Let's go.

jmartinesp (approved):
The new changes LGTM, thanks.

bindings/matrix-sdk-ffi/src/client_builder.rs
```diff
@@ -260,7 +260,8 @@ pub struct ClientBuilder {
     proxy: Option<String>,
     disable_ssl_verification: bool,
     disable_automatic_token_refresh: bool,
-    cross_process_refresh_lock_id: Option<String>,
+    cross_process_store_locks_holder_name: Option<String>,
+    enable_oidc_refresh_crypto_lock: bool,
```
pixlwave, line 264: I don't think the refresh lock is anything to do with crypto. It's purely a lock on making a token refresh request.

pixlwave (approved):
Otherwise, it all appears reasonable to me and a nice tidy-up 👏

bindings/matrix-sdk-ffi/src/client_builder.rs
Hywan: Alrighty. Renamed to `enable_iodc_refresh_lock`.

PR #4225, opened 2024-11-06
feat(media): replace `MediaThumbnailSettings` default ctor with reasonable defaults

bnjbvr:
This new ctor is used when caching a thumbnail everywhere, and when
attempting to retrieve it; it gives us a single place where to
coordinate the default `MediaThumbnailSettings` parameters.

### DIFF
--- a/crates/matrix-sdk/src/send_queue/upload.rs
+++ b/crates/matrix-sdk/src/send_queue/upload.rs
@@ -29,7 +29,6 @@ use ruma::{
         message::{MessageType, RoomMessageEventContent},
         MediaSource, ThumbnailInfo,
     },
-    media::Method,
     uint, OwnedMxcUri, OwnedTransactionId, TransactionId, UInt,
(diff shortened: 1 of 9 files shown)


crates/matrix-sdk/src/send_queue/upload.rs
```diff
@@ -73,15 +72,13 @@ fn make_local_thumbnail_media_request(
...
-        method: Method::Scale,
-        width,
-        height,
-        animated: false,
-    });
-    MediaRequestParameters { source, format }
+    MediaRequestParameters {
+        source: MediaSource::Plain(OwnedMxcUri::from(format!(
+            "mxc://send-queue.localhost/{}",
+            txn_id
```
jplatte, line 78:
```suggestion
            "mxc://send-queue.localhost/{txn_id}",
```

jplatte (approved):

PR #4228, opened 2024-11-06
Accept any string as a key for `m.direct` account data

MatMaul:
This is the follow up of this [Ruma PR](https://github.com/ruma/ruma/pull/1946) for the SDK.

Ruma PR needs to be merged first.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -56,7 +56,7 @@ proptest = { version = "1.5.0", default-features = false, features = ["std"] }
 rand = "0.8.5"
 reqwest = { version = "0.12.4", default-features = false }
 rmp-serde = "1.3.0"
-ruma = { version = "0.11.1", features = [
+ruma = { git = "https://github.com/ruma/ruma.git", rev = "35fda7f2f7811156df2e60b223dbf136fc143bc8", features = [
     "client-api-c",
     "compat-upload-signatures",
     "compat-user-id",
@@ -70,7 +70,7 @@ ruma = { version = "0.11.1", features = [
     "unstable-msc4075",
     "unstable-msc4140",
 ] }
-ruma-common = "0.14.1"
+ruma-common = { git = "https://github.com/ruma/ruma.git", rev = "35fda7f2f7811156df2e60b223dbf136fc143bc8" }
 serde = "1.0.151"
 serde_html_form = "0.2.0"
 serde_json = "1.0.91"
--- a/crates/matrix-sdk-base/src/store/migration_helpers.rs
+++ b/crates/matrix-sdk-base/src/store/migration_helpers.rs
@@ -23,6 +23,7 @@ use std::{
 use matrix_sdk_common::deserialized_responses::SyncTimelineEvent;
 use ruma::{
     events::{
+        direct::OwnedDirectUserIdentifier,
         room::{
             avatar::RoomAvatarEventContent,
             canonical_alias::RoomCanonicalAliasEventContent,
@@ -37,7 +38,7 @@ use ruma::{
         },
         EmptyStateKey, EventContent, RedactContent, StateEventContent, StateEventType,
     },
-    OwnedRoomId, OwnedUserId, RoomId,
+    OwnedRoomId, RoomId,
 };
 use serde::{Deserialize, Serialize};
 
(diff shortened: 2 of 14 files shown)


bnjbvr (changes requested):
I don't think losing static typing is an option, instead we should introduce an enum that can be one of all the types we expect (owned user id OR email address, if I understand correctly).

Also, `BaseRoomInfo` is serialized in the database, so we'd need a test making sure that we don't need a migration there, when the previous content was serialized as a `OwneduserId`; if the test fails, we need a migration.

MatMaul:
> I don't think losing static typing is an option, instead we should introduce an enum that can be one of all the types we expect (owned user id OR email address, if I understand correctly).

Fine for me, but it also needs to support phone numbers I think, basically any [identifier types](https://spec.matrix.org/v1.12/client-server-api/#identifier-types).

Fine for you ?

MatMaul:
Beware that it will not solve some the trouble for some legacy [buggy behavior](https://github.com/element-hq/element-x-android/issues/3074) of Element Web.

bnjbvr:
> Fine for you ?

Fine if we do have some type checking / upfront validation for all these other id types indeed :)

MatMaul:
Is there a way to not fully failed `m.direct` parsing when only some entries are invalid ?

The entry should just be committed in this case I believe.

Should I keep implementing `FromIterator<(String, Vec<OwnedRoomId>)>`, try to `.from` the string with the `UserIdentifier` enum we are talking about, and then put that in a `BTreeMap<UserIdentifier, Vec<OwnedRoomId>>` ?

MatMaul:
So I think the question is, will serde uses `FromIterator<(String, Vec<OwnedRoomId>)>` or not to deserialize the JSON in this case ?

spaetz:
If introducing ENUMS can we have an "UNKNOWN" too please which could catch either nonsense, such as "[object Object]": or other id's that this SDK perhaps does not know about (yet)?

MatMaul:
Yes I think I have a way with something like that @spaetz , thanks.

Cargo.toml
```diff
@@ -45,7 +45,7 @@ once_cell = "1.16.0"
 pin-project-lite = "0.2.9"
 rand = "0.8.5"
 reqwest = { version = "0.12.4", default-features = false }
-ruma = { version = "0.11.1", features = [
+ruma = { git = "https://github.com/ruma/ruma.git", rev = "35fda7f2f7811156df2e60b223dbf136fc143bc8", features = [
```
jplatte, line 48: Potentially stating the obvious here, but if you wanna do a crates.io release soon (I think this was planned), make sure to only merge this PR afterwards 😉

jplatte: Oh. Well I don't think we'll have the Ruma change blocking this PR out on crates.io anytime soon.

MatMaul: No rush in my side, it can wait.

Feel free to review in the meantime, so we can merge when we got a ruma release out.

poljar: Alright, the policy changed back, could you please rebase to pull in #4371, this will get rid of the CI error.

bnjbvr (commented):
Does OwnedDirectUserIdentifier deserialize cleanly from an OwnedUserId? Otherwise this needs a migration, since we're storing the BaseRoomInfo in db

bnjbvr:
> Does OwnedDirectUserIdentifier deserialize cleanly from an OwnedUserId? Otherwise this needs a migration, since we're storing the BaseRoomInfo in db

GOOD NEWS, EVERYONE! Since both are encoded as strings, and use the `IdZst` derive macro, they serialize and deserialize as a string, so this shouldn't require a migration. I would be grateful if a tiny test were added for this, to confirm my understanding: it would serialize an `OwnedUserId` and deserialize it as a `OwnedDirectUserIdentifier` cleanly. With that and the other review comment I've made above, LGTM :+1:

MatMaul:
> I would be grateful if a tiny test were added for this

This feels like something that should be in Ruma, [here is the PR.](https://github.com/ruma/ruma/pull/1968)

crates/matrix-sdk-base/src/store/migration_helpers.rs
```diff
@@ -200,12 +201,17 @@ impl BaseRoomInfoV1 {
             MinimalStateEvent::Redacted(ev) => MinimalStateEvent::Redacted(ev),
         });
 
+        let mut dm_targets_converted = HashSet::new();
+        for dm_target in dm_targets {
+            dm_targets_converted.insert(OwnedDirectUserIdentifier::from(dm_target));
```
bnjbvr, line 206:
```suggestion
        let mut converted_dm_targets = HashSet::new();
        for dm_target in dm_targets {
            converted_dm_targets.insert(OwnedDirectUserIdentifier::from(dm_target));
```

bnjbvr (approved):
lgtm, thanks.

PR #4229, opened 2024-11-06
feat(ffi): Auto approve the required widget capabilities for element call raise hand and reaction feature.

toger5:
This is the same work as done in:
https://github.com/element-hq/element-web/pull/28401

It also refactors the ffi `get_element_call_required_permissions` function to use a `read_send` array that gets copied both (read and send) capability lists.

Signed-off-by:

### DIFF
--- a/bindings/matrix-sdk-ffi/src/widget.rs
+++ b/bindings/matrix-sdk-ffi/src/widget.rs
@@ -5,6 +5,7 @@ use matrix_sdk::{
     async_trait,
     widget::{MessageLikeEventFilter, StateEventFilter},
 };
+use ruma::events::MessageLikeEventType;
 use tracing::error;
 
 use crate::{room::Room, RUNTIME};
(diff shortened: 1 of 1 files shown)


bindings/matrix-sdk-ffi/src/widget.rs
```diff
@@ -278,19 +304,15 @@ pub fn get_element_call_required_permissions(
...
-            // To read encryption keys
-            // TODO change this to the appropriate to-device version once ready
-            WidgetEventFilter::MessageLikeWithType {
-                event_type: "io.element.call.encryption_keys".to_owned(),
-            },
             // This allows the widget to check the room version, so it can know about
             // version-specific auth rules (namely MSC3779).
             WidgetEventFilter::StateWithType { event_type: StateEventType::RoomCreate.to_string() },
-        ],
+            // Custom element call reaction event. (which can be send multiple times for one
```
jmartinesp, line 310: Is this comment correctly placed?

toger5: That is an accident thx.
removed it since the same/similar comment is at the right location.

PR #4231, opened 2024-11-07
fix(event cache): give looser parameters for the deduplicator's bloom filters

bnjbvr:
The previous values would lead to super large memory allocations, as observed with `valgrind --tool=massive` on the tiny test added in this commit:

- for 400 rooms each having 100 events, this led to 540MB of allocations.
- for 1000 rooms each having 100 events, this led to 1.5GB of allocations.

This is not acceptable for any kind of devices, especially for mobile devices which may be more constrained on memory. The bloom filter is an optimisation to avoid going through events in the room's event list, so it shouldn't cause a big toll like that; instead, we can reduce the parameters values given when creating the filters.

With the given parameters, 1000 rooms each having 100 events leads to 1.2MB of allocations.

### DIFF
--- a/crates/matrix-sdk/src/event_cache/deduplicator.rs
+++ b/crates/matrix-sdk/src/event_cache/deduplicator.rs
@@ -41,8 +41,10 @@ impl fmt::Debug for Deduplicator {
 }
 
 impl Deduplicator {
-    const APPROXIMATED_MAXIMUM_NUMBER_OF_EVENTS: usize = 800_000;
-    const DESIRED_FALSE_POSITIVE_RATE: f64 = 0.001;
+    // Note: don't use too high numbers here, or the amount of allocated memory will
+    // explode. See https://github.com/matrix-org/matrix-rust-sdk/pull/4231 for details.
+    const APPROXIMATED_MAXIMUM_NUMBER_OF_EVENTS: usize = 1_000;
+    const DESIRED_FALSE_POSITIVE_RATE: f64 = 0.01;
 
     /// Create a new `Deduplicator`.
     pub fn new() -> Self {
@@ -138,7 +140,7 @@ pub enum Decoration<I> {
 
 #[cfg(test)]
 mod tests {
-    use assert_matches2::assert_let;
+    use assert_matches2::{assert_let, assert_matches};
     use matrix_sdk_base::deserialized_responses::SyncTimelineEvent;
     use ruma::{owned_event_id, user_id, EventId};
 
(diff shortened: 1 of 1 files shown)


crates/matrix-sdk/src/event_cache/deduplicator.rs
```diff
@@ -267,4 +267,33 @@ mod tests {
...
+        // This test was used as a testbed to observe, using `valgrind --tool=massive`,
+        // the total memory allocated by the deduplicator. We keep it checked in
+        // to revive this experiment in the future, if needs be.
+
+        let num_rooms = 10;
+        let num_events = 100;
+        //let num_rooms = std::env::var("ROOMS").unwrap().parse().unwrap();
+        //let num_events = std::env::var("EVENTS").unwrap().parse().unwrap();
+
+        let mut dedups = Vec::with_capacity(num_rooms);
```
jmartinesp, line 282: This isn't really needed for the benchmark, right? We don't do anything with the result that gets pushed.

crates/matrix-sdk/src/event_cache/deduplicator.rs
```diff
@@ -267,4 +267,33 @@ mod tests {
...
+
+    #[test]
+    fn test_bloom_filter_growth() {
+        // This test was used as a testbed to observe, using `valgrind --tool=massive`,
+        // the total memory allocated by the deduplicator. We keep it checked in
+        // to revive this experiment in the future, if needs be.
+
+        let num_rooms = 10;
+        let num_events = 100;
+        //let num_rooms = std::env::var("ROOMS").unwrap().parse().unwrap();
```
poljar, line 279: Can't we use `unwrap_or(100)` here so we don't need to modify the test to rerun the valgrind test?

jmartinesp (approved):
LGTM, just a comment about the test.

crates/matrix-sdk/src/event_cache/deduplicator.rs
```diff
@@ -41,8 +41,8 @@ impl fmt::Debug for Deduplicator {
 }
 
 impl Deduplicator {
-    const APPROXIMATED_MAXIMUM_NUMBER_OF_EVENTS: usize = 800_000;
-    const DESIRED_FALSE_POSITIVE_RATE: f64 = 0.001;
+    const APPROXIMATED_MAXIMUM_NUMBER_OF_EVENTS: usize = 1_000;
```
poljar, line 44: How does the memory usage look like if we just increase the false positive rate?

crates/matrix-sdk/src/event_cache/deduplicator.rs
jmartinesp: Oh wait, it's probably just to know how much memory is used, never mind.

poljar (approved):
Left a question and a nit. Looks good, thanks for investigating.

crates/matrix-sdk/src/event_cache/deduplicator.rs
bnjbvr: Exactly!

crates/matrix-sdk/src/event_cache/deduplicator.rs
bnjbvr: 1GB for 1000 rooms with 100 events each. Oof.

bnjbvr: 12 MB if i lower `APPROXIMATED_MAXIMUM_NUMBER_OF_EVENTS` to 10K. I suppose we can keep 1,000 until proven otherwise.

crates/matrix-sdk/src/event_cache/deduplicator.rs
bnjbvr: Good idea, done something equivalent, thanks.

Hywan:
Oh, nice, thanks!

PR #4234, opened 2024-11-08
refactor(ffi): Make `is_room_alias_format_valid` stricter

jmartinesp:
Previously this only used the Ruma checks, which only handled the initial `#` char and the domain part. With these changes, the name part is also validated, checking it's lowercase, with no whitespaces and containing only allowed chars, similar to what `DisplayName::to_room_alias_name` does.

Signed-off-by:

### DIFF
--- a/bindings/matrix-sdk-ffi/src/room_alias.rs
+++ b/bindings/matrix-sdk-ffi/src/room_alias.rs
@@ -1,10 +1,13 @@
 use matrix_sdk::RoomDisplayName;
-use ruma::RoomAliasId;
 
-/// Verifies the passed `String` matches the expected room alias format.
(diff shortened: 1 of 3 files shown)


stefanceriu (approved):
Looks good to me but I can't help wonder if we should contribute this back up to Ruma

jmartinesp:
I think the id parsing checks were kept very 'light' for compatibility reasons, but I might be wrong. The only validation that seems to do anything besides checking the prefix seems to be [user_id](https://github.com/ruma/ruma/blob/a77a6ad2138363deaeac544759af1ef0b31d1e0a/crates/ruma-identifiers-validation/src/user_id.rs#L20).

PR #4235, opened 2024-11-08
fix(ffi): match the right status code in `Client::is_room_alias_available`

jmartinesp:
Previously I was matching a different error state, so it turns out the code was failing 🫠 .

Signed-off-by:

### DIFF
--- a/bindings/matrix-sdk-ffi/src/client.rs
+++ b/bindings/matrix-sdk-ffi/src/client.rs
@@ -1130,21 +1130,16 @@ impl Client {
         Ok(())
     }
 
-    /// Checks if a room alias is available in the current homeserver.
+    /// Checks if a room alias is not in use yet.
+    ///
+    /// Returns:
+    /// - `Ok(true)` if the room alias is available.
+    /// - `Ok(false)` if it's not (the resolve alias request returned a `404`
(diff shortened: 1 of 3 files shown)


poljar (approved):
How come we don't need to change a test? Could it be that we don't have one? :fearful: 

Anyways, would be nice to add it now.

jmartinesp:
Will FFI layer tests run in the CI? If so, I'll add one, I thought they didn't 🫤 .

poljar:
> Will FFI layer tests run in the CI? If so, I'll add one, I thought they didn't 🫤 .

Doesn't seem like it.

I think that we should either enable them or move the bulk of the logic to the main crate and add a test there. Fixing a bug without a test sounds worse to me.

jmartinesp:
Moved the fn to the SDK crate and added some tests, now the FFI layer only maps some values and calls the inner function.

poljar (approved):
Nice, thanks for adding the test.

Small reminder, please use the `MatrixMockServer` for new tests in the future.

jmartinesp:
I implemented a new endpoint in `MatrixMockServer` for this, I'll merge the PR as soon as the CI is finished.

PR #4236, opened 2024-11-08
tests: Refactor the widget tests to use `MockMatrixServer`

toger5:
This is a follow up PR on: https://github.com/matrix-org/matrix-rust-sdk/pull/3987
And tries to use the `MockMatrixServer` wherever reasonable in the widget integration tests.

Signed-off-by:

### DIFF
--- a/crates/matrix-sdk/src/test_utils/mocks.rs
+++ b/crates/matrix-sdk/src/test_utils/mocks.rs
@@ -333,20 +333,49 @@ impl MatrixMockServer {
     /// see also [mock_room_send] for more context.
     ///
     /// ```
+    /// # tokio_test::block_on(async {
+    /// use matrix_sdk::{ruma::{room_id, event_id}, test_utils::mocks::MatrixMockServer};
+    /// use serde_json::json;
+    ///
+    /// let mock_server = MatrixMockServer::new().await;
+    /// let client = mock_server.client_builder().build().await;
+    ///
+    /// mock_server.mock_room_state_encryption().plain().mount().await;
+    ///
+    /// let room = mock_server
+    ///     .sync_joined_room(&client, room_id!("!room_id:localhost"))
+    ///     .await;
+    ///
     /// let event_id = event_id!("$some_id");
     /// mock_server
-    ///     .mock_room_send_for_type("m.room.message")
+    ///     .mock_room_send_for_type("m.room.message".into())
     ///     .ok(event_id)
     ///     .expect(1)
     ///     .mount()
     ///     .await;
+    ///
+    /// let responseNotMocked = room.send_raw("m.room.reaction", json!({ "body": "Hello world" })).await?;
+    ///
+    /// assert!(
+    ///     responseNotMocked,
+    ///     "The event ID we mocked should match the one we received when we sent the event"
+    /// );
+    ///
+    /// let response = room.send_raw("m.room.message", json!({ "body": "Hello world" })).await?;
+    ///
+    /// assert_eq!(
+    ///     event_id,
+    ///     response.event_id,
+    ///     "The event ID we mocked should match the one we received when we sent the event"
+    /// );
+    /// # anyhow::Ok(()) });
     /// ```
     pub fn mock_room_send_for_type(
         &self,
         event_type: MessageLikeEventType,
     ) -> MockEndpoint<'_, RoomSendEndpoint> {
         let mock = Mock::given(method("PUT"))
-            .and(path_regex(format!(r"^/_matrix/client/r0/rooms/.*/send/{}", event_type)))
+            .and(path_regex(format!(r"^/_matrix/client/v3/rooms/.*/send/{}", event_type)))
             .and(header("authorization", "Bearer 1234"));
         MockEndpoint { mock, server: &self.server, endpoint: RoomSendEndpoint }
     }
@@ -375,7 +404,7 @@ impl MatrixMockServer {
         state_type: StateEventType,
     ) -> MockEndpoint<'_, RoomSendEndpoint> {
         let mock = Mock::given(method("PUT"))
-            .and(path_regex(format!(r"^/_matrix/client/r0/rooms/.*/state/{}", state_type)))
+            .and(path_regex(format!(r"^/_matrix/client/v3/rooms/.*/state/{}", state_type)))
             .and(header("authorization", "Bearer 1234"));
         MockEndpoint { mock, server: &self.server, endpoint: RoomSendEndpoint }
     }
@@ -495,10 +524,11 @@ impl MatrixMockServer {
         }
     }
 
-    /// Create a prebuild mock for reading room message with the `/messages` endpoint.
+    /// Create a prebuild mock for reading room message with the `/messages`
+    /// endpoint.
     pub fn mock_room_messages(&self, limit: Option<u32>) -> MockEndpoint<'_, RoomMessagesEndpoint> {
         let mut mock = Mock::given(method("GET"))
-            .and(path_regex(r"^/_matrix/client/r0/rooms/.*/messages$"))
+            .and(path_regex(r"^/_matrix/client/v3/rooms/.*/messages$"))
             .and(header("authorization", "Bearer 1234"));
         if let Some(l) = limit {
             mock = mock.and(query_param("limit", l.to_string()));
--- a/crates/matrix-sdk/src/test_utils/mocks.rs
+++ b/crates/matrix-sdk/src/test_utils/mocks.rs
@@ -29,7 +29,8 @@ use matrix_sdk_test::{
 };
 use ruma::{
     directory::PublicRoomsChunk,
-    events::{MessageLikeEventType, StateEventType},
+    events::{AnyStateEvent, AnyTimelineEvent, MessageLikeEventType, StateEventType},
+    serde::Raw,
     MxcUri, OwnedEventId, OwnedRoomId, RoomId, ServerName,
 };
 use serde::Deserialize;
@@ -344,7 +345,7 @@ impl MatrixMockServer {
     ///
     /// let event_id = event_id!("$some_id");
     /// mock_server
-    ///     .mock_send_room_state()
+    ///     .mock_room_send_state()
     ///     .ok(event_id)
     ///     .expect(1)
     ///     .mount()
@@ -368,7 +369,7 @@ impl MatrixMockServer {
         let mock = Mock::given(method("PUT"))
             .and(path_regex(r"^/_matrix/client/v3/rooms/.*/state/.*"))
             .and(header("authorization", "Bearer 1234"));
-        MockEndpoint { mock, server: &self.server, endpoint: RoomSendStateEndpoint::new() }
+        MockEndpoint { mock, server: &self.server, endpoint: RoomSendStateEndpoint::default() }
     }
 
     /// Creates a prebuilt mock for asking whether *a* room is encrypted or not.
@@ -495,7 +496,7 @@ impl MatrixMockServer {
         if let Some(l) = limit {
             mock = mock.and(query_param("limit", l.to_string()));
         }
-        MockEndpoint { mock, server: &self.server, endpoint: RoomMessagesEndpoint {} }
+        MockEndpoint { mock, server: &self.server, endpoint: RoomMessagesEndpoint }
     }
 
     /// Create a prebuilt mock for uploading media.
@@ -902,7 +903,7 @@ impl<'a> MockEndpoint<'a, RoomSendEndpoint> {
     ///     .mount()
     ///     .await;
     /// mock_server
-    ///     .mock_send_room_state()
+    ///     .mock_room_send_state()
     ///     .for_type("m.room.message".into())
     ///     .ok(event_id)
     ///     .expect(1)
@@ -973,24 +974,19 @@ impl<'a> MockEndpoint<'a, RoomSendEndpoint> {
 }
 
 /// A prebuilt mock for sending a state event in a room.
+#[derive(Default)]
 pub struct RoomSendStateEndpoint {
     state_key: Option<String>,
     event_type: Option<StateEventType>,
 }
-impl RoomSendStateEndpoint {
-    /// Create new RoomSendStateEndpoint with unset state_key and event_type.
-    pub fn new() -> Self {
-        Self { state_key: None, event_type: None }
-    }
-}
 
 impl<'a> MockEndpoint<'a, RoomSendStateEndpoint> {
     fn generate_path_regex(endpoint: &RoomSendStateEndpoint) -> String {
-        return format!(
+        format!(
             r"^/_matrix/client/v3/rooms/.*/state/{}/{}",
             endpoint.event_type.as_ref().map(|t| t.to_string()).unwrap_or(".*".to_owned()),
             endpoint.state_key.as_ref().map(|k| k.to_string()).unwrap_or(".*".to_owned())
-        );
+        )
     }
     /// Ensures that the body of the request is a superset of the provided
     /// `body` parameter.
@@ -1381,9 +1377,22 @@ pub struct RoomMessagesEndpoint;
 impl<'a> MockEndpoint<'a, RoomMessagesEndpoint> {
     /// Returns a messages endpoint that emulates success, i.e. the messages
     /// provided as `response` could be retrieved.
-    pub fn ok(self, response: impl Into<Value>) -> MatrixMock<'a> {
-        let body: Value = response.into();
-        let mock = self.mock.respond_with(ResponseTemplate::new(200).set_body_json(body));
+    ///
+    /// Note: pass `chunk` in the correct order: topological for forward
+    /// pagination, reverse topological for backwards pagination.
+    pub fn ok(
+        self,
+        start: String,
+        end: Option<String>,
+        chunk: Vec<impl Into<Raw<AnyTimelineEvent>>>,
+        state: Vec<impl Into<Raw<AnyStateEvent>>>,
+    ) -> MatrixMock<'a> {
+        let mock = self.mock.respond_with(ResponseTemplate::new(200).set_body_json(json!({
+            "start": start,
+            "end": end,
+            "chunk": chunk.into_iter().map(|ev| ev.into()).collect::<Vec<_>>(),
+            "state": state.into_iter().map(|ev| ev.into()).collect::<Vec<_>>(),
+        })));
         MatrixMock { server: self.server, mock }
     }
 }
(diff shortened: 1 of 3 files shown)


toger5:
How do i test `/test_utils/mocks.rs`?

crates/matrix-sdk/src/test_utils/mocks.rs
```diff
@@ -316,6 +321,166 @@ impl MatrixMockServer {
...
+    /// ```
+    pub fn mock_room_send_for_type(
+        &self,
+        event_type: MessageLikeEventType,
+    ) -> MockEndpoint<'_, RoomSendEndpoint> {
+        let mock = Mock::given(method("PUT"))
+            .and(path_regex(format!(r"^/_matrix/client/v3/rooms/.*/send/{}", event_type)))
+            .and(header("authorization", "Bearer 1234"));
+        MockEndpoint { mock, server: &self.server, endpoint: RoomSendEndpoint }
+    }
```
bnjbvr, line 379: This should not be implemented this way: instead there should be a single `mock_room_send`, and the `MockEndpoint<RoomSendEndpoint>` impl block should have a method to limit `for_type(self, event_type: MessageLikeEventType) -> Self`. Otherwise we're losing compsability at the endpoint level, and risk duplicating endpoints lots of time, which we'd like to avoid.

crates/matrix-sdk/src/test_utils/mocks.rs
```diff
@@ -316,6 +321,166 @@ impl MatrixMockServer {
...
+    /// let response = room.send_state_event_raw("m.room.message", "my_key", json!({ "body": "Hello world" })).await?;
+    /// // The `/state` endpoint should be mocked by the server.
+    /// assert_eq!(
+    ///     event_id,
+    ///     response.event_id,
+    ///     "The event ID we mocked should match the one we received when we sent the event"
+    /// );
+    /// # anyhow::Ok(()) });
+    /// ```
+    pub fn mock_room_state(&self) -> MockEndpoint<'_, RoomSendEndpoint> {
```
bnjbvr, line 424: Can you call this `mock_room_send_state`?

crates/matrix-sdk/src/test_utils/mocks.rs
```diff
@@ -316,6 +321,166 @@ impl MatrixMockServer {
...
+    /// ```
+    pub fn mock_room_state_for_type(
+        &self,
+        state_type: StateEventType,
+    ) -> MockEndpoint<'_, RoomSendEndpoint> {
+        let mock = Mock::given(method("PUT"))
+            .and(path_regex(format!(r"^/_matrix/client/v3/rooms/.*/state/{}", state_type)))
+            .and(header("authorization", "Bearer 1234"));
+        MockEndpoint { mock, server: &self.server, endpoint: RoomSendEndpoint }
+    }
```
bnjbvr, line 482: Same comment, this should be a specialized method on the return type of `mock_room_send_state`.

crates/matrix-sdk/src/test_utils/mocks.rs
```diff
@@ -431,6 +596,18 @@ impl MatrixMockServer {
...
+    /// Create a prebuild mock for paginating room message with the `/messages`
+    /// endpoint.
+    pub fn mock_room_messages(&self, limit: Option<u32>) -> MockEndpoint<'_, RoomMessagesEndpoint> {
+        let mut mock = Mock::given(method("GET"))
+            .and(path_regex(r"^/_matrix/client/v3/rooms/.*/messages$"))
+            .and(header("authorization", "Bearer 1234"));
+        if let Some(l) = limit {
+            mock = mock.and(query_param("limit", l.to_string()));
+        }
+        MockEndpoint { mock, server: &self.server, endpoint: RoomMessagesEndpoint {} }
```
bnjbvr, line 608:
```suggestion
        MockEndpoint { mock, server: &self.server, endpoint: RoomMessagesEndpoint }
```

crates/matrix-sdk/src/test_utils/mocks.rs
```diff
@@ -980,6 +1157,20 @@ impl<'a> MockEndpoint<'a, RoomEventEndpoint> {
...
 }
 
+/// A prebuilt mock for the `/messages` endpoint.
+pub struct RoomMessagesEndpoint;
+
+/// A prebuilt mock for getting a room messages in a room.
+impl<'a> MockEndpoint<'a, RoomMessagesEndpoint> {
+    /// Returns a messages endpoint that emulates success, i.e. the messages
+    /// provided as `response` could be retrieved.
+    pub fn ok(self, response: impl Into<Value>) -> MatrixMock<'a> {
```
bnjbvr, line 1167: No, we should not use `impl Into<Value>`, otherwise we lose all the usefulness of these mocking APIs by reverting to "we don't know what the return value looks like, it can be an arbitrary JSON". Instead we should use high-level types. We have another place where we do mock messages, grep for `fn mock_messages` (…it's even better, we do have two implementations :D).

bnjbvr (changes requested):

crates/matrix-sdk/src/test_utils/mocks.rs
```diff
@@ -316,6 +321,92 @@ impl MatrixMockServer {
...
+            .and(header("authorization", "Bearer 1234"));
+        MockEndpoint { mock, server: &self.server, endpoint: RoomSendEndpoint }
+    }
+
+    /// Creates a prebuilt mock for sending a state event in a room.
+    ///
+    /// Similar to: [mock_room_send]
+    ///
+    /// Note: works with *any* room.
+    /// Note: works with *any* event type.
```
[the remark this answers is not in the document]
toger5: Done as part of the refactor to use `for_key` `for_type`

poljar (approved):

crates/matrix-sdk/src/test_utils/mocks.rs
```diff
@@ -317,6 +322,56 @@ impl MatrixMockServer {
...
+    ///     .ok(event_id)
+    ///     .expect(1)
+    ///     .mount()
+    ///     .await;
+    ///
+    /// let responseNotMocked = room.send_raw("m.room.create", json!({ "body": "Hello world" })).await;
+    /// // The `/send` endpoint should not be mocked by the server.
+    /// assert!(responseNotMocked.is_err());
+    ///
+    ///
```
bnjbvr, line 358:
```suggestion
```

crates/matrix-sdk/src/test_utils/mocks.rs
```diff
@@ -755,7 +822,7 @@ impl<'a, T> MockEndpoint<'a, T> {
     }
 }
 
-/// A prebuilt mock for sending an event in a room.
+/// A prebuilt mock for sending a message like event in a room.
```
bnjbvr, line 825:
```suggestion
/// A prebuilt mock for sending a message-like event in a room.
```

crates/matrix-sdk/src/test_utils/mocks.rs
```diff
@@ -801,10 +868,286 @@ impl<'a> MockEndpoint<'a, RoomSendEndpoint> {
...
     /// ```
-    pub fn body_matches_partial_json(self, body: serde_json::Value) -> Self {
+    pub fn body_matches_partial_json(self, body: Value) -> Self {
         Self { mock: self.mock.and(body_partial_json(body)), ..self }
     }
 
+    /// Ensures that the send endpoint request uses a specific event type.
+    ///
+    /// Note: works with *any* room.
+    ///
```
bnjbvr, line 878: I would remove this, since there might a future method that makes this obsolete.

```suggestion
```

crates/matrix-sdk/src/test_utils/mocks.rs
```diff
@@ -801,10 +868,286 @@ impl<'a> MockEndpoint<'a, RoomSendEndpoint> {
...
+    ///     .for_type("m.room.message".into())
+    ///     .ok(event_id)
+    ///     .expect(1)
+    ///     .mount()
+    ///     .await;
+    ///
+    /// let responseNotMocked = room.send_raw("m.room.reaction", json!({ "body": "Hello world" })).await;
+    /// // The `m.room.reaction` event type should not be mocked by the server.
+    /// assert!(responseNotMocked.is_err());
+    ///
```
bnjbvr, line 909:
```suggestion
```

crates/matrix-sdk/src/test_utils/mocks.rs
```diff
@@ -801,10 +868,286 @@ impl<'a> MockEndpoint<'a, RoomSendEndpoint> {
...
+    ///     response.event_id,
+    ///     "The event ID we mocked should match the one we received when we sent the event"
+    /// );
+    /// # anyhow::Ok(()) });
+    /// ```
+    pub fn for_type(self, event_type: MessageLikeEventType) -> Self {
+        Self {
+            mock: self
+                .mock
+                .and(path_regex(format!(r"^/_matrix/client/v3/rooms/.*/send/{}", event_type))),
```
bnjbvr, line 924: Ok, I think this works because we're adding a new path regexp for the more precise path. An alternative would have been to lazily provide the path when eventually building the mock into a `MatrixMock`, we can revisit this later.

crates/matrix-sdk/src/test_utils/mocks.rs
```diff
@@ -801,10 +868,286 @@ impl<'a> MockEndpoint<'a, RoomSendEndpoint> {
...
+
+impl<'a> MockEndpoint<'a, RoomSendStateEndpoint> {
+    fn generate_path_regex(endpoint: &RoomSendStateEndpoint) -> String {
+        format!(
+            r"^/_matrix/client/v3/rooms/.*/state/{}/{}",
+            endpoint.event_type.as_ref().map(|t| t.to_string()).unwrap_or(".*".to_owned()),
+            endpoint.state_key.as_ref().map(|k| k.to_string()).unwrap_or(".*".to_owned())
+        )
+    }
+    /// Ensures that the body of the request is a superset of the provided
```
bnjbvr, line 984:
```suggestion
    }
    
    /// Ensures that the body of the request is a superset of the provided
```

crates/matrix-sdk/src/test_utils/mocks.rs
```diff
@@ -801,10 +868,286 @@ impl<'a> MockEndpoint<'a, RoomSendEndpoint> {
...
+pub struct RoomSendStateEndpoint {
+    state_key: Option<String>,
+    event_type: Option<StateEventType>,
+}
+
+impl<'a> MockEndpoint<'a, RoomSendStateEndpoint> {
+    fn generate_path_regex(endpoint: &RoomSendStateEndpoint) -> String {
+        format!(
+            r"^/_matrix/client/v3/rooms/.*/state/{}/{}",
+            endpoint.event_type.as_ref().map(|t| t.to_string()).unwrap_or(".*".to_owned()),
```
bnjbvr, line 980: Not that we care too much about optimizing this code, but

```suggestion
            endpoint.event_type.as_ref().map(|t| t.to_string()).unwrap_or_else(|| ".*".to_owned()),
```

crates/matrix-sdk/src/test_utils/mocks.rs
```diff
@@ -801,10 +868,286 @@ impl<'a> MockEndpoint<'a, RoomSendEndpoint> {
...
+            endpoint.state_key.as_ref().map(|k| k.to_string()).unwrap_or(".*".to_owned())
```
bnjbvr, line 981:
```suggestion
            endpoint.state_key.as_ref().map(|k| k.to_string()).unwrap_or_else(|| ".*".to_owned())
```

crates/matrix-sdk/src/test_utils/mocks.rs
```diff
@@ -801,10 +868,286 @@ impl<'a> MockEndpoint<'a, RoomSendEndpoint> {
...
+    ///     event_id, response.event_id,
+    ///     "The event ID we mocked should match the one we received when we sent the event"
+    /// );
+    ///
+    /// # anyhow::Ok(()) });
+    /// ```
+    pub fn for_type(self, event_type: StateEventType) -> Self {
+        let endpoint = RoomSendStateEndpoint { event_type: Some(event_type), ..self.endpoint };
+        let matcher = path_regex(Self::generate_path_regex(&endpoint));
+        Self { endpoint, mock: self.mock.and(matcher), ..self }
```
bnjbvr, line 1089: I think here it would be better to generate the path regex at the end, just before building the final `MatrixMock`.

crates/matrix-sdk/src/test_utils/mocks.rs
```diff
@@ -1024,6 +1367,33 @@ impl<'a> MockEndpoint<'a, RoomEventEndpoint> {
...
+pub struct RoomMessagesEndpoint;
+
+/// A prebuilt mock for getting a room messages in a room.
+impl<'a> MockEndpoint<'a, RoomMessagesEndpoint> {
+    /// Returns a messages endpoint that emulates success, i.e. the messages
+    /// provided as `response` could be retrieved.
+    ///
+    /// Note: pass `chunk` in the correct order: topological for forward
+    /// pagination, reverse topological for backwards pagination.
+    pub fn ok(
```
bnjbvr, line 1380: noice!

bnjbvr (approved):
Super nice, thanks. Feel free to address the last comments, or let me know if you're bored with this PR, and then I can do it myself as a followup :)

crates/matrix-sdk/src/test_utils/mocks.rs
toger5: Where would that be?
There are numerous places we build the final MatrixMock. Do you want to implement all the methods from `MockEndpoint<'a, T>` in 
`MockEndpoint<'a, RoomSendStateEndpoint>`. What if someone adds a new method to the generic version that can build a matrix mock but forgets to update this version to build the regex?

crates/matrix-sdk/src/test_utils/mocks.rs
```diff
@@ -801,10 +868,286 @@ impl<'a> MockEndpoint<'a, RoomSendEndpoint> {
...
+    ///     "The event ID we mocked should match the one we received when we sent the event"
+    /// );
+    /// # anyhow::Ok(()) });
+    /// ```
+    pub fn for_key(self, state_key: String) -> Self {
+        let endpoint = RoomSendStateEndpoint { state_key: Some(state_key), ..self.endpoint };
+        let matcher = path_regex(Self::generate_path_regex(&endpoint));
+        Self { endpoint, mock: self.mock.and(matcher), ..self }
+    }
+
```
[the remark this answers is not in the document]
toger5: You mean the doc comment for `ok`?
I suppose this is still helpful to have for `ok`. What other test du you have in mind?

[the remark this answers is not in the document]
bnjbvr: Yes. The doc comment *below* is incorrect because it mentions `mock_room_send` while this is `mock_room_send_state`.

crates/matrix-sdk/src/test_utils/mocks.rs
```diff
@@ -432,6 +487,18 @@ impl MatrixMockServer {
         }
     }
 
+    /// Create a prebuild mock for paginating room message with the `/messages`
+    /// endpoint.
+    pub fn mock_room_messages(&self, limit: Option<u32>) -> MockEndpoint<'_, RoomMessagesEndpoint> {
```
bnjbvr, line 492: Can `limit` be set optionally on `MockEndpoint<'_, RoomMessagesEndpoint>`, since this is basically a builder pattern?

crates/matrix-sdk/src/test_utils/mocks.rs
bnjbvr: You can take a look at `RoomEventEndpoint` as an example: it stores an optional `room`, set to `None` first: 

https://github.com/matrix-org/matrix-rust-sdk/blob/80a48f53ad19a2aae050737a6e8dbb9fe502acb3/crates/matrix-sdk/src/test_utils/mocks.rs#L431

Then, we have a builder method to optionally set it:

https://github.com/matrix-org/matrix-rust-sdk/blob/80a48f53ad19a2aae050737a6e8dbb9fe502acb3/crates/matrix-sdk/src/test_utils/mocks.rs#L995-L998

Then, we create the final path when we want to get the final `MatrixMock`:

https://github.com/matrix-org/matrix-rust-sdk/blob/80a48f53ad19a2aae050737a6e8dbb9fe502acb3/crates/matrix-sdk/src/test_utils/mocks.rs#L1009-L1023

If this code were to be duplicated in multiple methods, then it could be factored out as an helper method, like `generate_path_regexp` does, in fact.

Does it make sense now?

crates/matrix-sdk/tests/integration/event_cache.rs
```diff
@@ -257,6 +257,7 @@ async fn test_ignored_unignored() {
 /// Puts a mounting point for /messages for a pagination request, matching
 /// against a precise `from` token given as `expected_from`, and returning the
 /// chunk of events and the next token as `end` (if available).
+// TODO: replace this with the `mock_room_messages` form mocks.rs
```
bnjbvr, line 260: Sorry :nerd_face: 

```suggestion
// TODO: replace this with the `mock_room_messages` from mocks.rs
```

crates/matrix-sdk/src/test_utils/mocks.rs
```diff
@@ -317,6 +322,56 @@ impl MatrixMockServer {
...
+    ///
+    /// let event_id = event_id!("$some_id");
+    /// mock_server
+    ///     .mock_room_send_state()
+    ///     .ok(event_id)
+    ///     .expect(1)
+    ///     .mount()
+    ///     .await;
+    ///
+    /// let responseNotMocked = room.send_raw("m.room.create", json!({ "body": "Hello world" })).await;
```
bnjbvr, line 354: usually we use camel_case for variable names in Rust :eyes: Can you do that here, please?

crates/matrix-sdk/src/test_utils/mocks.rs
```diff
@@ -801,10 +868,284 @@ impl<'a> MockEndpoint<'a, RoomSendEndpoint> {
...
+    /// let event_id = event_id!("$some_id");
+    /// mock_server
+    ///     .mock_room_send()
+    ///     .for_type("m.room.message".into())
+    ///     .ok(event_id)
+    ///     .expect(1)
+    ///     .mount()
+    ///     .await;
+    ///
+    /// let responseNotMocked = room.send_raw("m.room.reaction", json!({ "body": "Hello world" })).await;
```
bnjbvr, line 904: ditto

bnjbvr (changes requested):
We're getting there!

bnjbvr:
To respond to my own suggestion in https://github.com/matrix-org/matrix-rust-sdk/pull/4236#discussion_r1866119824: this is not possible, because it hinders usage of `error500()` and `error_too_large()`, which never define the more restricted path. I've reverted to the previous code, with a code comment explaining that it's fine to redefine a path regex, since it ought to be more specialized than the previous ones.

bnjbvr (approved):
Thanks, let's revisit the approach of finishing the path, because it's a bit calling for misuse, but that one is on me :innocent:

PR #4237, opened 2024-11-08
Introduce a higher level `DisplayName` struct which helps to protect against impersonation via display names

poljar:


crates/matrix-sdk-base/src/deserialized_responses.rs
```diff
@@ -67,6 +70,155 @@ pub struct AmbiguityChanges {
...
+///
+/// // The normalized and sanitized string will be returned by DisplayName.as_str().
+/// assert_eq!(display_name.as_str(), "sahasrahla");
+/// ```
+///
+/// ```
+/// # use matrix_sdk_base::deserialized_responses::DisplayName;
+/// let display_name = DisplayName::new("@alice:localhost");
+///
+/// // The display name looks like na MXID, it's ambiguous because of that.
```
jmartinesp, line 98: Typo: `na` -> `an`.

crates/matrix-sdk-base/src/deserialized_responses.rs
```diff
@@ -67,6 +70,155 @@ pub struct AmbiguityChanges {
...
+    ///     * Arabic Letter RTL mark U+061C
+    ///     * Zero width no-break space (BOM) U+FEFF
+    const HIDDEN_CHARACTERS_PATTERN: &str =
+        "[\u{2000}-\u{200D}\u{300}-\u{036f}\u{2062}-\u{2063}\u{2800}\u{061c}\u{feff}]";
+
+    const MXID_REGEX: Lazy<Regex> = Lazy::new(|| {
+        Regex::new(DisplayName::MXID_PATTERN)
+            .expect("We should be able to create a regex from our static MXID pattern")
+    });
+    const LEFT_RO_RIGHT_REGEX: Lazy<Regex> = Lazy::new(|| {
```
jmartinesp, line 152: Typo: `LEFT_RO_RIGHT` -> `LEFT_TO_RIGHT`.

jmartinesp (approved):
LGTM, thanks!

jmartinesp:
It seems like Rust doesn't like the way the lazy regexes are implemented though.

poljar:
> It seems like Rust doesn't like the way the lazy regexes are implemented though.

Yeah, though we don't mutate things but I'll see how to implement this without any compiler complaints.

crates/matrix-sdk-base/src/deserialized_responses.rs
```diff
@@ -310,3 +464,104 @@ impl SyncOrStrippedState<RoomPowerLevelsEventContent> {
...
+            );
+        }};
+    }
+
+    macro_rules! assert_display_name_ne {
+        ($left:expr, $right:expr $(, $desc:expr)?) => {{
+            let left = crate::deserialized_responses::DisplayName::new($left);
+            let right = crate::deserialized_responses::DisplayName::new($right);
+
+            assert_ne!(
```
Hywan, line 488: Time to contribute to `similar_asserts` to include `assert_ne!`?

crates/matrix-sdk-base/src/store/ambiguity_map.rs
```diff
@@ -275,11 +282,219 @@ impl AmbiguityCache {
...
+        assert_ambiguity!(
+            [("@alice:localhost", "alice"), ("@bob:localhost", "ALICE")],
+            [("alice", true)],
+            "Alice and Bob share a display name that is differently capitalized"
+        );
+
+        assert_ambiguity!(
+            [("@alice:localhost", "alice"), ("@bob:localhost", "аlice")],
+            [("alice", true)],
+            "Bob tries to impresonate Alice using a cyrilic а"
```
Hywan, line 431: s/impresonate/impersonate/

in all tests.

crates/matrix-sdk-base/src/deserialized_responses.rs
poljar: What would you then show? The `ne` variant fails if the two given objects are equal but they shouldn't be. Two equal objects create an empty diff, so what should `similar_asserts` show?

Hywan: 🤦 

/me goes back in its bed

crates/matrix-sdk-base/src/deserialized_responses.rs
poljar: 8069cd36dd2ba8606098d912c4568a42cf2c5c76

crates/matrix-sdk-base/src/deserialized_responses.rs
poljar: 8069cd36dd2ba8606098d912c4568a42cf2c5c76

crates/matrix-sdk-sqlite/src/state_store.rs
```diff
@@ -1509,33 +1523,49 @@ impl StateStore for SqliteStateStore {
...
-            .map(|n| (self.encode_key(keys::DISPLAY_NAME, n), n.as_ref()))
+            .flat_map(|display_name| {
+                // We encode the display name as the `raw_str()` and the normalized string.
+                //
+                // This is for compatibility reasons since:
+                //  1. Previously "Alice" and "alice" were considered to be distinct display
+                //     names, while we now consider them to be the same so we need to merge the
+                //     previously distinct buckets of user IDs.
+                //  2. We can't do a migration to merge the previously distinct buckets of user
+                //     IDs since the dipslay names itself are hashed before they are persisted
```
bnjbvr, line 1545:
```suggestion
                //     IDs since the display names itself are hashed before they are persisted
```

crates/matrix-sdk-base/src/deserialized_responses.rs
```diff
@@ -67,6 +70,170 @@ pub struct AmbiguityChanges {
...
+        self.has_cancer() || self.might_look_like_an_mxid() || self.has_left_to_right_characters()
+    }
+
+    /// Returns the underlying normalized and sanitized string of this
+    /// [`DisplayName`].
+    ///
+    /// If normalization and removal of confusable characters failed, this will
+    /// return the original string provided during the construction of the
+    /// [`DisplayName`].
+    pub fn as_str(&self) -> &str {
```
dkasak, line 198: The doc comment is lying to me at first ("returns the underlying normalised..." when there is actually an exception in which it doesn't return the normalised string) and the name of the method is far too benign for its magical behaviour. Can we do away with `as_str` and use `as_normalised_str` instead?

poljar: I removed the method: https://github.com/matrix-org/matrix-rust-sdk/pull/4237/commits/6f08ae43020acf25c3d6f0da612f38a588be3314

crates/matrix-sdk-base/src/deserialized_responses.rs
```diff
@@ -67,6 +70,170 @@ pub struct AmbiguityChanges {
...
+    /// Returns the underlying normalized and and sanitized string of this
+    /// [`DisplayName`].
+    ///
+    /// Returns `None` if normalization failed during construction of this
+    /// [`DisplayName`].
+    pub fn as_normalized_str(&self) -> Option<&str> {
+        self.decancered.as_deref()
+    }
+
+    fn has_cancer(&self) -> bool {
```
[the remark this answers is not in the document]
poljar: This method has been removed as well.

crates/matrix-sdk-base/src/deserialized_responses.rs
```diff
@@ -67,6 +70,170 @@ pub struct AmbiguityChanges {
...
+            .is_some_and(|decancered| decancered == self.raw.to_lowercase());
+
+        !cancer_free
+    }
+
+    fn has_left_to_right_characters(&self) -> bool {
+        LEFT_TO_RIGHT_REGEX.is_match(&self.raw)
+    }
+
+    fn might_look_like_an_mxid(&self) -> bool {
```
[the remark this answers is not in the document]
poljar: https://github.com/matrix-org/matrix-rust-sdk/pull/4237/commits/450687f5882f4570e0d7bbd5219767ca2c13ac47

crates/matrix-sdk-base/src/deserialized_responses.rs
```diff
@@ -67,6 +70,166 @@ pub struct AmbiguityChanges {
...
+    /// This includes:
+    ///     * various width spaces U+2000 - U+200D
+    ///     * Combining characters U+0300 - U+036F
+    ///     * Blank/invisible characters (U2800, U2062-U2063)
+    ///     * Arabic Letter RTL mark U+061C
+    ///     * Zero width no-break space (BOM) U+FEFF
+    const HIDDEN_CHARACTERS_PATTERN: &str =
+        "[\u{2000}-\u{200D}\u{300}-\u{036f}\u{2062}-\u{2063}\u{2800}\u{061c}\u{feff}]";
+
+    /// Regex pattern to match an uppercase I.
```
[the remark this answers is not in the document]
poljar: This was applied.

poljar:
@dkasak are you happy now with the PR? Can this be approved?

crates/matrix-sdk-base/src/deserialized_responses.rs
```diff
@@ -67,6 +70,170 @@ pub struct AmbiguityChanges {
...
+    ///     * Arabic Letter RTL mark U+061C
+    ///     * Zero width no-break space (BOM) U+FEFF
+    const HIDDEN_CHARACTERS_PATTERN: &str =
+        "[\u{2000}-\u{200D}\u{300}-\u{036f}\u{2062}-\u{2063}\u{2800}\u{061c}\u{feff}]";
+
+    /// Regex pattern to match `i` characters (upper and lower case).
+    ///
+    /// This is used to replace an uppercase I with a lowercase l, i.e. to mark
+    /// "Hello" and "HeIlo" as ambiguous.
+    const I_PATTERN: &str = "[Ii]";
```
davidegirardi, line 172: If we approach it like this we have to add:
- `O` vs `0`
- `rn` vs `m` (up for discussion)

I think these should cover the characters easier to confuse when using a sans-serif font.

Test vectors:
- `Octavius` `0ctavius`
- `matrix` `rmatrix`

davidegirardi (changes requested):

crates/matrix-sdk-base/src/deserialized_responses.rs
poljar: The zero case is already handled: https://github.com/matrix-org/matrix-rust-sdk/pull/4237/files#diff-f089f4ce85deadaee6a01c387ae405895ab1f7aff275d500b763ba88dfdcf638R193

Not really convinced about the `rn` case.

davidegirardi: Is that line replacing the 0 case insensitive? I think it's easier to confuse 0 with O than 0 with o.

Yes, the rn case is up for discussion because it's kind of borderline.

poljar: https://github.com/matrix-org/matrix-rust-sdk/pull/4237/commits/1391e09234529a9f3cd9b51fa48bc3ebac31b02b

dkasak (approved):

PR #4241, opened 2024-11-10
feat(WidgetDriver): Pass Matrix API errors to the widget

toger5:
Currently the WidgetDriver just returns unspecified error strings to the widget that can be used to display an issue description to the user. It is not helpful to run code like a retry or other error mitigation logic.

Here it is proposed to add standardized errors for issues that every widget driver implementation can run into (all matrix cs api errors): https://github.com/matrix-org/matrix-spec-proposals/pull/2762#discussion_r1838804895

This PR forwards the errors that occur during the widget processing to the widget in the correct format.

NOTE:
It does not include request Url and http Headers. See also: https://github.com/matrix-org/matrix-spec-proposals/pull/2762#discussion_r1839802292

Signed-off-by:


crates/matrix-sdk/src/widget/machine/driver_req.rs
```diff
@@ -76,7 +76,7 @@ where
 
     pub(crate) fn then(
         self,
-        response_handler: impl FnOnce(Result<T, String>, &mut WidgetMachine) -> Vec<Action>
+        response_handler: impl FnOnce(Result<T, crate::Error>, &mut WidgetMachine) -> Vec<Action>
```
bnjbvr, line 79: Since you're around, can you add a doc comment to this function please?

crates/matrix-sdk/src/widget/machine/from_widget.rs
```diff
@@ -46,17 +48,63 @@ pub(super) struct FromWidgetErrorResponse {
...
+impl From<&str> for FromWidgetErrorResponse {
+    fn from(error: &str) -> Self {
+        error.to_owned().into()
     }
 }
 
 #[derive(Serialize)]
 struct FromWidgetError {
     message: String,
+    matrix_api_error: Option<FromWidgetMatrixErrorBody>,
```
bnjbvr, line 97: This crate is barely documented, which makes it super hard to understand how it works. Let's change that; of course I wouldn't have you change all of it at once, so let's try to do it incrementally. The doc comments don't have to be long; a single short descriptive line is sufficient to give us a rough idea of why/how it's used.

Please add a doc comment to:

- every single struct/enum you're adding fields to (here, `FromWidgetError`)
- every field/variant from those structs/enums (here, `message` above)
- every new struct/enum/field/variant

crates/matrix-sdk/src/widget/machine/from_widget.rs
```diff
@@ -46,17 +48,63 @@ pub(super) struct FromWidgetErrorResponse {
...
 }
 
+#[derive(Serialize)]
+struct FromWidgetMatrixErrorBody {
+    http_status: u32,
+    // TODO: figure out the which type to use here.
+    http_headers: HashMap<String, String>,
+    url: String,
+    response: Value,
+}
```
bnjbvr, line 107: nit: always add a newline after a block closing
```suggestion
}

```

crates/matrix-sdk/src/widget/machine/from_widget.rs
```diff
@@ -46,17 +48,63 @@ pub(super) struct FromWidgetErrorResponse {
...
+    matrix_api_error: Option<FromWidgetMatrixErrorBody>,
 }
 
+#[derive(Serialize)]
+struct FromWidgetMatrixErrorBody {
+    http_status: u32,
+    // TODO: figure out the which type to use here.
+    http_headers: HashMap<String, String>,
+    url: String,
+    response: Value,
```
bnjbvr, line 106: I want doc comments everywhere here too please :)

crates/matrix-sdk/src/widget/machine/incoming.rs
```diff
@@ -38,7 +38,7 @@ pub(crate) enum IncomingMessage {
         request_id: Uuid,
 
         /// The result of the request: response data or error message.
-        response: Result<MatrixDriverResponse, String>,
+        response: Result<MatrixDriverResponse, crate::Error>,
```
bnjbvr, line 41: The doc comment above needs an update^^

crates/matrix-sdk/src/widget/machine/from_widget.rs
```diff
@@ -46,17 +48,63 @@ pub(super) struct FromWidgetErrorResponse {
...
+
+impl From<String> for FromWidgetErrorResponse {
+    fn from(error: String) -> Self {
+        Self { error: FromWidgetError { message: error, matrix_api_error: None } }
+    }
+}
+
+impl From<&str> for FromWidgetErrorResponse {
+    fn from(error: &str) -> Self {
+        error.to_owned().into()
```
bnjbvr, line 90: Instead of using `From` implementations, it'd be simpler to use plain functions: `from_string(String) -> Self`, `from_error(matrix_sdk::Error) -> Self`, `from_http_error(...)`. Using `into()` at the call sites makes it harder to find which code is used, and thus harder to navigate the code.

crates/matrix-sdk/src/widget/machine/from_widget.rs
```diff
@@ -46,17 +48,63 @@ pub(super) struct FromWidgetErrorResponse {
...
     message: String,
+    matrix_api_error: Option<FromWidgetMatrixErrorBody>,
 }
 
+#[derive(Serialize)]
+struct FromWidgetMatrixErrorBody {
+    http_status: u32,
+    // TODO: figure out the which type to use here.
+    http_headers: HashMap<String, String>,
+    url: String,
```
bnjbvr, line 105: If these two fields are never used, let's not introduce them now?

crates/matrix-sdk/src/widget/machine/mod.rs
```diff
@@ -306,16 +310,21 @@ impl WidgetMachine {
...
+                        (Ok(_), CapabilitiesState::Unset) => {
+                            Err("Received read event request before capabilities negotiation"
+                                .into())
+                        }
+                        (Ok(_), CapabilitiesState::Negotiating) => {
+                            Err("Received read event request while capabilities were negotiating"
+                                .into())
+                        }
+                        (Err(e), _) => Err((&e).into()),
+                    };
```
bnjbvr, line 327: This is messier like this. Can we keep the `and_then`, and after it call `map_err` to convert the error type if needs be, instead?

crates/matrix-sdk/src/widget/matrix.rs
```diff
@@ -54,17 +52,17 @@ impl MatrixDriver {
...
+        self.room.client.send(OpenIdRequest::new(user_id), None).await.map_err(Error::Http)
     }
 
     /// Reads the latest `limit` events of a given `event_type` from the room.
     pub(crate) async fn read_message_like_events(
         &self,
         event_type: MessageLikeEventType,
         limit: u32,
-    ) -> Result<Vec<Raw<AnyTimelineEvent>>> {
+    ) -> Result<Vec<Raw<AnyTimelineEvent>>, Error> {
```
bnjbvr, line 65: nit: can be reverted, because `Result<T>` is a synonym for `Result<T, crate::Error>` if the error type isn't specified.

crates/matrix-sdk/src/widget/matrix.rs
```diff
@@ -80,7 +78,7 @@ impl MatrixDriver {
         &self,
         event_type: StateEventType,
         state_key: &StateKeySelector,
-    ) -> Result<Vec<Raw<AnyTimelineEvent>>> {
+    ) -> Result<Vec<Raw<AnyTimelineEvent>>, Error> {
```
bnjbvr, line 81: ditto here

crates/matrix-sdk/src/widget/matrix.rs
```diff
@@ -118,7 +116,7 @@ impl MatrixDriver {
         state_key: Option<String>,
         content: Box<RawJsonValue>,
         delayed_event_parameters: Option<delayed_events::DelayParameters>,
-    ) -> Result<SendEventResponse> {
+    ) -> Result<SendEventResponse, Error> {
```
bnjbvr, line 119: ditto here

crates/matrix-sdk/src/widget/matrix.rs
```diff
@@ -54,17 +52,17 @@ impl MatrixDriver {
     }
 
     /// Requests an OpenID token for the current user.
-    pub(crate) async fn get_open_id(&self) -> HttpResult<OpenIdResponse> {
+    pub(crate) async fn get_open_id(&self) -> Result<OpenIdResponse, Error> {
```
bnjbvr, line 55: no need to specify the error type here neither

crates/matrix-sdk/src/widget/matrix.rs
```diff
@@ -167,9 +165,9 @@ impl MatrixDriver {
         &self,
         delay_id: String,
         action: UpdateAction,
-    ) -> HttpResult<delayed_events::update_delayed_event::unstable::Response> {
+    ) -> Result<delayed_events::update_delayed_event::unstable::Response, Error> {
```
bnjbvr, line 168: nit: no need to specify the `Error` here too

crates/matrix-sdk/src/widget/mod.rs
```diff
@@ -195,7 +195,7 @@ impl<T: CapabilitiesProvider> ProcessingContext<T> {
                 self.to_widget_tx.send(msg).await.map_err(|_| ())?;
             }
             Action::MatrixDriverRequest { request_id, data } => {
-                let response = match data {
+                let response: Result<MatrixDriverResponse, Error> = match data {
```
bnjbvr, line 198: Is this explicit type declaration needed?

crates/matrix-sdk/tests/integration/widget.rs
```diff
@@ -679,8 +673,67 @@ async fn test_send_delayed_state_event() {
...
+        &driver_handle,
+        json!([
+            "org.matrix.msc4157.send.delayed_event",
+            "org.matrix.msc2762.send.event:m.room.message"
+        ]),
+    )
+    .await;
+
+    Mock::given(method("PUT"))
+        .and(path_regex(r"^/_matrix/client/v3/rooms/.*/send/m.room.message/.*$"))
```
bnjbvr, line 692: We have a mock in the `MatrixMockServer` called `.mock_room_send()`, can you use that instead?

bnjbvr (changes requested):
Makes sense! But we need many more comments for this, please; let's try to add them as we touch these files.

crates/matrix-sdk/src/widget/machine/mod.rs
toger5: Are you sure with the custom error message dependent on the capability state this would become sth like this:
```
let CapabilitiesState::Negotiated(capabilities) = &machine.capabilities
else {
    match &machine.capabilities{
        CapabilitiesState::Unset => return  Err(FromWidgetErrorResponse::from_string(
            "Received read event request before capabilities negotiation",
        )),
        CapabilitiesState::Negotiating => return  Err(FromWidgetErrorResponse::from_string(
            "Received read event request while capabilities were negotiating",
        )),
        CapabilitiesState::Negotiated(capabilities) => todo!(),
    }
};
```

bnjbvr:
(I might not be able to review this this week; if it's urgent, please ask review to somebody else.)

crates/matrix-sdk/src/widget/machine/from_widget.rs
```diff
@@ -41,28 +41,80 @@ pub(super) enum FromWidgetRequest {
...
+            },
+        }
+    }
+
+    /// Create a error response to send to the widget from a matrix sdk error.
+    pub(crate) fn from_error(error: &Error) -> Self {
+        match &error {
+            Error::Http(e) => FromWidgetErrorResponse::from_http_error(e),
+            Error::UnknownError(e) => FromWidgetErrorResponse::from_string(e.to_string()),
+            _ => FromWidgetErrorResponse::from_string(error.to_string()),
```
bnjbvr, line 75: Are the last two arms not equivalent?

crates/matrix-sdk/src/widget/machine/from_widget.rs
```diff
@@ -41,28 +41,80 @@ pub(super) enum FromWidgetRequest {
...
+    }
+
+    /// Create a error response to send to the widget from a string.
+    pub(crate) fn from_string<S: Into<String>>(error: S) -> Self {
+        Self { error: FromWidgetError { message: error.into(), matrix_api_error: None } }
     }
 }
 
+/// The serializable section of an error response send by the client as a
+/// response to a from widget request.
```
bnjbvr, line 86: Unless this is referring to a specific type:

```suggestion
/// response to a request from the widget.
```

If this is referring to a specific type, could it be the type name with a link, e.g. ```[`TheTypeName`]```?

crates/matrix-sdk/src/widget/machine/mod.rs
```diff
@@ -324,17 +339,23 @@ impl WidgetMachine {
...
+                                "Received read event request before capabilities negotiation",
+                            ))
+                        }
+                        (Ok(_), CapabilitiesState::Negotiating) => {
+                            Err(FromWidgetErrorResponse::from_string(
+                                "Received read event request while capabilities were negotiating",
+                            ))
+                        }
+                        (Err(e), _) => Err(FromWidgetErrorResponse::from_error(&e)),
+                    };
```
bnjbvr, line 358: Still think it would be slightly clearer to handle the capabilities error before the other.

crates/matrix-sdk/src/widget/machine/from_widget.rs
```diff
@@ -41,28 +41,80 @@ pub(super) enum FromWidgetRequest {
...
+                        }),
+                        _ => None,
+                    },
+                ),
+            },
+        }
+    }
+
+    /// Create a error response to send to the widget from a matrix sdk error.
+    pub(crate) fn from_error(error: &Error) -> Self {
```
bnjbvr, line 71: Can this take ownership of the error?

crates/matrix-sdk/src/widget/machine/mod.rs
```diff
@@ -500,14 +525,14 @@ impl WidgetMachine {
 
     fn send_from_widget_error_response(
         raw_request: Raw<FromWidgetRequest>,
-        error: impl fmt::Display,
+        error: FromWidgetErrorResponse,
     ) -> Action {
-        Self::send_from_widget_response(raw_request, FromWidgetErrorResponse::new(error))
+        Self::send_from_widget_response(raw_request, error)
```
bnjbvr, line 530: Maybe `send_from_widget_error_response` isn't required anymore, if it only forwards to another function?

crates/matrix-sdk/tests/integration/widget.rs
bnjbvr: This has been resolved without explanation and without a change to the code, can you comment at least please? :)

crates/matrix-sdk/src/widget/machine/from_widget.rs
```diff
@@ -41,28 +41,80 @@ pub(super) enum FromWidgetRequest {
...
+
+/// The serializable section of a widget response that represents a matrix
+/// error.
+#[derive(Serialize)]
+struct FromWidgetMatrixErrorBody {
+    /// The status code of the http response
+    http_status: u32,
+    /// The matrix standard error response including the `errorcode` and the
+    /// `error` message as defined in the spec: https://spec.matrix.org/v1.12/client-server-api/#standard-error-response
+    response: Value,
```
bnjbvr, line 106: Is it really worth saving the `response` field as a json::Value, instead of having the fields in there, automatically serialized by serde?

bnjbvr (changes requested):

crates/matrix-sdk/src/widget/machine/from_widget.rs
toger5: It could but would there be any benefit to that? we will only call methods that use the ref?

crates/matrix-sdk/src/widget/machine/from_widget.rs
toger5: `error.to_string()` generates somehting like: `UnknownError: more information ...`
`e.to_string()` results in: `more information ...`

Since most of the unknown errors are errors originating from the widget api, we dont need the `UnknownError` label in front of the message.

crates/matrix-sdk/src/widget/machine/from_widget.rs
toger5: I think using `StandardErrorBody` makes sense?
This struct was a bit suspicious to me see: https://matrix.to/#/!veagCdDBjKrMsOCzrq:privacytools.io/$2ieYTtk1E4HU36rQZ7r40_bjGaWUN-zfOGz01Yfv3Ng?via=matrix.org&via=envs.net&via=kabou.ro
(I was afraid it is only supposed to be used in a very special occasion since its not even used in the ErrorBody enum)
Also the `ErrorBody` enum seemed to be the perfect fit here but it is not serializable.
So I did not use it before.

crates/matrix-sdk/src/widget/machine/mod.rs
toger5: all the helper methods actually can be removed. `send_from_widget_response` is the only thing we need.

crates/matrix-sdk/src/widget/machine/from_widget.rs
bnjbvr: Avoid cloning in the function's body, and it's a good practice whenever you can do that (as it avoids future changes to this code from cloning too).

crates/matrix-sdk/src/widget/machine/from_widget.rs
bnjbvr: Makes sense, can you add a comment about that, then, please?

crates/matrix-sdk/src/widget/machine/from_widget.rs
bnjbvr: Would it make sense to store the result as an integer and the error message as a String?

crates/matrix-sdk/tests/integration/widget.rs
toger5: I fixed this locally clicked resolve and then noticed I also want to create `with_delay` mocks to replace a couple of more tests with the new mocking pattern.

crates/matrix-sdk/src/widget/machine/from_widget.rs
toger5: We will only call `to_string` and `as_client_api_error` on the error object with both use `&self`. `to_string` will of course allocate new data but I am not sure there is a way to convert an error to a string without that overhead?

crates/matrix-sdk/src/widget/machine/from_widget.rs
toger5: The errorcode is a a string as well. I used the StandardErrorBody which is exactly what we expect here.

crates/matrix-sdk/src/widget/machine/from_widget.rs
bnjbvr: Indeed there's no way, but overall passing by ownership makes for cleaner APIs, whenever we can do it.

bnjbvr (changes requested):
This doesn't pass tests; I'd rather take a look after tests are passing, please! (or, if you need help with tests, please let me know and pinpoint where you need help :-))

bnjbvr (approved):
I've addressed the comments that hadn't been addressed. Please let me know when your PR is not done yet (by removing the review request, or posting a comment), so I'm not reviewing stuff that's just about to change :upside_down_face:

PR #4249, opened 2024-11-12
test(ci): Re-enable Complement Crypto

Hywan:
- Complement Crypto has been disabled in https://github.com/matrix-org/matrix-rust-sdk/pull/4224 because the public FFI API has changed.
- https://github.com/matrix-org/complement-crypto/pull/153 fixes Complement Crypto.
- Now we can re-enable Complement Crypto.

### DIFF
--- a/.github/workflows/bindings_ci.yml
+++ b/.github/workflows/bindings_ci.yml
@@ -177,12 +177,12 @@ jobs:
       - name: Build Framework
         run: target/debug/xtask swift build-framework --target=aarch64-apple-ios --profile=dev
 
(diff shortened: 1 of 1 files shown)


poljar:
Please don't disable complement crypto in the future anymore, if there is an API break open a PR on complement-crypto with the same branch name as your SDK PR.

That way the github action in the SDK will use complement crypto from the PR with the updated API, then CI on the SDK will pass and you can merge both PRs at the same time.

Hywan:
Alrighty.

PR #4252, opened 2024-11-12
fix(utd_hook): Fix regression causing retry to report false late decrypt

BillCarsonFr:
There has been a recent change on `Decryptor#decrypt_event_impl` causing the function to return an TimelineEvent of kind unable to decrypt instead of failing with an error.

The `late_decrypt` detection code was not change, causing any retry to mark UTDs as late decrypt.

Signed-off-by:

### DIFF
--- a/crates/matrix-sdk-ui/src/timeline/controller/mod.rs
+++ b/crates/matrix-sdk-ui/src/timeline/controller/mod.rs
@@ -1064,16 +1064,22 @@ impl<P: RoomDataProvider> TimelineController<P> {
 
                     match decryptor.decrypt_event_impl(original_json).await {
                         Ok(event) => {
-                            trace!(
-                                "Successfully decrypted event that previously failed to decrypt"
-                            );
+                            if let matrix_sdk::deserialized_responses::TimelineEventKind::UnableToDecrypt { utd_info, ..} = event.kind {
+                                info!("Failed to decrypt event after receiving room key: {:?}", utd_info.reason);
+                                None
+                            } else {
+                                trace!(
+                                    kind = ?event.kind,
+                                    "Successfully decrypted event that previously failed to decrypt"
+                                );
 
-                            // Notify observers that we managed to eventually decrypt an event.
-                            if let Some(hook) = unable_to_decrypt_hook {
-                                hook.on_late_decrypt(&remote_event.event_id, *utd_cause).await;
+                                // Notify observers that we managed to eventually decrypt an event.
+                                if let Some(hook) = unable_to_decrypt_hook {
+                                    hook.on_late_decrypt(&remote_event.event_id, *utd_cause).await;
+                                }
+
+                                Some(event)
                             }
-
-                            Some(event)
                         }
                         Err(e) => {
                             info!("Failed to decrypt event after receiving room key: {e}");


crates/matrix-sdk-ui/src/timeline/controller/mod.rs
```diff
@@ -1064,16 +1064,22 @@ impl<P: RoomDataProvider> TimelineController<P> {
...
-
-                            // Notify observers that we managed to eventually decrypt an event.
-                            if let Some(hook) = unable_to_decrypt_hook {
-                                hook.on_late_decrypt(&remote_event.event_id, *utd_cause).await;
+                            if let matrix_sdk::deserialized_responses::TimelineEventKind::UnableToDecrypt { utd_info, ..} = event.kind {
+                                info!("Failed to decrypt event after receiving room key: {:?}", utd_info.reason);
+                                None
+                            } else {
+                                trace!(
+                                    kind = ?event.kind,
```
richvdh, line 1072: this is probably a bit verbose?

richvdh (commented):
This change looks correct to me; sorry for missing this before.

It would be good if we had a regression test.

BillCarsonFr:
> According to @BillCarsonFr: this bug should have caused the UTD rate reported to Posthog to (incorrectly) drop significantly. We haven't actually observed that -- possibly because the relevant change [`f4e8222` (#4070)](https://github.com/matrix-org/matrix-rust-sdk/pull/4070/commits/f4e8222dd0d90483dc63894127cbb5d9a3400de2) hasn't rolled out very far yet?

Actually there is no grace period on rust sdk UTD reporting like on other platform; On other platforms, an event decrypted in less than 4s is graced out and not reported. It is not the case in rust sdk, so the UTD would be reported immediatly but as a temporary UTD (with `timeToDecryptMillis` set to a value).
So we shouldn't see a drop in posthog. What could be expected is a raise of temporary UTD, but this is not observed either.

BillCarsonFr:
> This change looks correct to me; sorry for missing this before.
> 
> It would be good if we had a regression test.

Added a regression test

crates/matrix-sdk-ui/src/timeline/controller/mod.rs
BillCarsonFr: Not changed by this PR. Any other reason to remive it? is it spamming the RS?

crates/matrix-sdk-ui/src/timeline/controller/mod.rs
```diff
@@ -1064,16 +1064,22 @@ impl<P: RoomDataProvider> TimelineController<P> {
...
                     match decryptor.decrypt_event_impl(original_json).await {
                         Ok(event) => {
-                            trace!(
-                                "Successfully decrypted event that previously failed to decrypt"
-                            );
-
-                            // Notify observers that we managed to eventually decrypt an event.
-                            if let Some(hook) = unable_to_decrypt_hook {
-                                hook.on_late_decrypt(&remote_event.event_id, *utd_cause).await;
+                            if let matrix_sdk::deserialized_responses::TimelineEventKind::UnableToDecrypt { utd_info, ..} = event.kind {
```
bnjbvr, line 1067: nit: use `use`, not the fully qualified name here

crates/matrix-sdk-ui/src/timeline/controller/mod.rs
bnjbvr: Contains most of the event's data that's not privacy-sensitive, so it will be spammy. This line can be reached multiple times for each undecrypted event that failed to re-decrypt again, so it will be super spammy, please remove it.

crates/matrix-sdk-ui/src/timeline/tests/encryption.rs
```diff
@@ -170,6 +172,72 @@ async fn test_retry_message_decryption() {
...
+        .await;
+    assert_eq!(timeline.controller.items().await.len(), 2);
+
+    // Wait past the max delay for utd late decryption detection
+    sleep(Duration::from_secs(2)).await;
+
+    {
+        let utds = hook.utds.lock().unwrap();
+        assert_eq!(utds.len(), 1);
+        // Without the regression fix this would be Some(time to decrypt)
```
bnjbvr, line 236:
```suggestion
        // This is the main thing we're testing: if this wasn't identified as a definite UTD, this would be `Some(..)`.
```

crates/matrix-sdk-ui/src/timeline/traits.rs
```diff
@@ -302,8 +302,14 @@ impl Decryptor for (matrix_sdk_base::crypto::OlmMachine, ruma::OwnedRoomId) {
...
-        Ok(event.into())
+        match olm_machine
+            .try_decrypt_room_event(raw.cast_ref(), room_id, &decryption_settings)
+            .await?
+        {
+            RoomEventDecryptionResult::Decrypted(decrypted) => Ok(decrypted.into()),
+            RoomEventDecryptionResult::UnableToDecrypt(utd_info) => {
+                Ok(TimelineEvent::new_utd_event(raw.clone(), utd_info))
+            }
+        }
```
bnjbvr, line 313: Why is this change needed, out of curiosity? This is for the test timeline; is it operating that differently from the non-test timeline?

bnjbvr (changes requested):

bnjbvr:
(Great catch and thanks for fixing, btw)

crates/matrix-sdk-ui/src/timeline/traits.rs
BillCarsonFr: It's needed to make the regression test ~"fails correctly". 
If not the test never detects the behavior change because it is using the mock `Decryptor` that continues to fail with an error when there is an UTD instead of return a OK UTDEvent type.

The underlying problem is that we use a big catch everything Error for all possible errors of the SDK. And here the change that created the regression should have resulted in a change of signature of the trait. It is not anymore Result<, ALLERRORS> but a Result <,SubsetOfErrors>?

crates/matrix-sdk-ui/src/timeline/controller/mod.rs
BillCarsonFr: > nit: use `use`, not the fully qualified name here

There is already an import with the same name `super::event_handler::TimelineEventKind` hence the need for fully qualified name.

richvdh: you can `use matrix_sdk::deserialized_responses::TimelineEventKind as SdkTimelineEventKind` or something, to give it a different name

bnjbvr: I like Rich's suggestion :+1:

crates/matrix-sdk-ui/src/timeline/traits.rs
bnjbvr: Alright, I'm slightly scared that we might be reproducing an error only in testing conditions, but this code is likely going away at some point…

bnjbvr (approved):
Thanks!

crates/matrix-sdk-ui/src/timeline/controller/mod.rs
BillCarsonFr: done

PR #4253, opened 2024-11-12
feat(sdk): Implement Client::observe_events and `Client::observe_room_events`

Hywan:
This patch introduces a mechanism similar to
`Client::add_event_handler` and `Client::add_room_event_handler`
but with a reactive programming pattern. This patch adds
`Client::observe_events` and `Client::observe_room_events`.

```rust
// Get an observer.
let observer =
    client.observe_events::<SyncRoomMessageEvent, (Room, Vec<Action>)>();

// Subscribe to the observer.
let mut subscriber = observer.subscribe();

// Use the subscriber as a `Stream`.
let (message_event, (room, push_actions)) = subscriber.next().await.unwrap();
```

When calling `observe_events`, one has to specify the type of event
(in the example, `SyncRoomMessageEvent`) and a context (in the example,
`(Room, Vec<Action>)`, respectively for the room and the push actions).

Read the documentation of the `ObservableEventHandler` and
`EventHandlerSubscriber` to learn more about how it works.

In order to achieve that, it was necessary to implement `EventHandlerContext`
on tuples where each part implements `EventHandlerContext`. Why? Because,
while `EventHandler` is implemented for `FnOnce()` with 0 to 8 arguments
(e.g. with 2 arguments `client.add_event_handler(|ev: E, ctx1: C1, ctx2: C2| { … })`),
it is _not_ possible to have “variadic” generic parameters. Example, it's not possible
to write `client.observe_events::<E, C1, C2>()`. The trick is to use tuples (!), like:
`client.observe_events::<E, (C1, C2)>()`: the type of `observe_events` remains
“constant”:

```rust
pub fn observe_events<Ev, Ctx>(&self) -> ObservableEventHandler<(Ev, Ctx)>
where
    Ev: SyncEvent + DeserializeOwned + Send + Sync + 'static,
    Ctx: EventHandlerContext + Send + Sync + 'static,
{
   …
}
```


jmartinesp (approved):
Thanks, the changes LGTM in general since I assume it's correct, but maybe a second pair of more experienced eyes wouldn't hurt since I'm not familiar with some parts of it.

crates/matrix-sdk/src/client/mod.rs
```diff
@@ -798,12 +799,96 @@ impl Client {
...
+    ///
+    /// # async fn example(client: Client) {
+    /// let observer =
+    ///     client.observe_events::<SyncRoomMessageEvent, (Room, Vec<Action>)>();
+    ///
+    /// let mut subscriber = observer.subscribe();
+    ///
+    /// let (message_event, (room, push_actions)) =
+    ///     subscriber.next().await.unwrap();
+    /// # }
```
poljar, line 836: If we do this we can get rid of the `unwrap()` and use `?`.

```suggestion
    /// # anyhow::Ok(()) };
```

crates/matrix-sdk/src/client/mod.rs
```diff
@@ -798,12 +799,96 @@ impl Client {
...
+    /// # Example
+    ///
+    /// ```
+    /// use futures_util::StreamExt as _;
+    /// use matrix_sdk::{
+    ///     ruma::{events::room::message::SyncRoomMessageEvent, push::Action},
+    ///     Client, Room,
+    /// };
+    ///
+    /// # async fn example(client: Client) {
```
poljar, line 828: I would like to see some more examples here, time and time again people come and ask how the event handler can do foo. So we added many cases to the event handler examples.

Would be nice to replicate the event handler examples here.

poljar (approved):
Very nice. Quite the clever trick on how to make this pattern generic, great work.

I left some small nits about the documentation but otherwise this is good to go.

crates/matrix-sdk/src/client/mod.rs
Hywan: This is an `Option` but sure, I can remove the `unwrap` :-).

crates/matrix-sdk/src/client/mod.rs
Hywan: Done.

PR #4254, opened 2024-11-12
feat(WidgetDriver): Send state from state sync and not from timeline to widget

toger5:
This PR changes the source of messages send to the widget. Instead of using: 
`let handle = self.room.add_event_handler(move |raw: Raw<AnySyncTimelineEvent>| {})`
which listens to all events form the timeline section of the sync, we use two listeners for state and message like:
`let handle = self.room.add_event_handler(move |raw: Raw<AnySyncStateEvent>| {})`
And
`let handle = self.room.add_event_handler(move |raw: Raw<AnySyncMessageLikeEvent>| {})`

`AnySyncStateEvent` should only get events from the state block of the SSS response.
and `AnySyncMessageLikeEvent` all the message like events from the timeline block.

This makes the behaviour equivalent to: https://github.com/element-hq/element-web/pull/28422

Signed-off-by:

### DIFF
--- a/crates/matrix-sdk/src/widget/matrix.rs
+++ b/crates/matrix-sdk/src/widget/matrix.rs
@@ -184,7 +184,7 @@ impl MatrixDriver {
         let drop_guard_state = self.room.client().event_handler_drop_guard(handle_state);
 
         // The receiver will get a combination of state and messgage like events.
-        // The state events will come from the state section of the sync.
+        // The state events will come from the state section of the sync. (to always represent current resolved state)
         // All state events in the timeline section of the sync will not be forwarded to
         // the widget.
         // TODO annotate the events and send both timeline and state section state events.
--- a/crates/matrix-sdk/src/widget/matrix.rs
+++ b/crates/matrix-sdk/src/widget/matrix.rs
@@ -26,8 +26,9 @@ use ruma::{
     },
     assign,
     events::{
-        AnyMessageLikeEventContent, AnyStateEventContent, AnySyncTimelineEvent, AnyTimelineEvent,
-        MessageLikeEventType, StateEventType, TimelineEventType,
+        AnyMessageLikeEventContent, AnyStateEventContent, AnySyncMessageLikeEvent,
+        AnySyncStateEvent, AnySyncTimelineEvent, AnyTimelineEvent, MessageLikeEventType,
+        StateEventType, TimelineEventType,
     },
     serde::Raw,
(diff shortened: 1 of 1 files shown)


crates/matrix-sdk/src/widget/matrix.rs
```diff
@@ -164,21 +165,39 @@ impl MatrixDriver {
...
+        let handle_state = self.room.add_event_handler(move |raw: Raw<AnySyncStateEvent>| {
+            let _ = tx.send(attach_room_id(raw.cast_ref(), &room_id));
             async {}
         });
-
-        let drop_guard = self.room.client().event_handler_drop_guard(handle);
-        EventReceiver { rx, _drop_guard: drop_guard }
+        let drop_guard_state = self.room.client().event_handler_drop_guard(handle_state);
+
+        // The receiver will get a combination of state and messgage like events.
```
stefanceriu, line 186:
```suggestion
        // The receiver will get a combination of state and message like events.
```

crates/matrix-sdk/src/widget/matrix.rs
```diff
@@ -164,21 +165,39 @@ impl MatrixDriver {
...
+        // The state events will come from the state section of the sync. (to always
+        // represent current resolved state) All state events in the timeline
```
stefanceriu, line 188:
```suggestion
        // represent current resolved state) All state events in the timeline
```
```suggestion
        // The state events will come from the state section of the sync (to always
        // represent current resolved state). All state events in the timeline
```

stefanceriu (changes requested):

stefanceriu (approved):
lgtm 👍

toger5:
@stefanceriu @bnjbvr, I asked this in one of the room but it does not hurt to ask for double confirmation that this comment is actually what is happening:
```
// Get only all state events from the state section of the sync.
let handle_state = self.room.add_event_handler(move |raw: Raw<AnySyncStateEvent>| {
    let _ = tx.send(attach_room_id(raw.cast_ref(), &room_id));
    async {}
});
```

stefanceriu:
Looks like that to me, yes: https://github.com/matrix-org/matrix-rust-sdk/blob/ee7c80c234c10fd370c1e6c20656e57260d94f3d/crates/matrix-sdk/src/event_handler/mod.rs#L351

stefanceriu:
Confirmed with @bnjbvr that this is indeed correct, will go ahead and merge it 👍

PR #4255, opened 2024-11-13
feat(base): Make `ObservableMap::stream` works on `wasm32-unknown-unknown`

Hywan:
This patch updates `eyeball-im` and `eyeball-im-util` to integrate https://github.com/jplatte/eyeball/pull/63/. With this new feature, we can have a single implementation of `ObservableMap` (instead of 2: one for all targets, one for `wasm32-u-u`). It makes it possible to get `Client::rooms_stream` available on all targets now.

This was required to get `matrix-sdk-ui` usable on `wasm32-u-u`, esp. the `RoomListService`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1642,14 +1642,13 @@ dependencies = [
 
 [[package]]
 name = "eyeball-im"
-version = "0.5.0"
+version = "0.5.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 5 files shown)


jplatte:
You didn't upgrade the util crate (latest release is technically breaking, so I bumped the minor version). Seems like it's not necessary for this, but wouldn't hurt either.

poljar (approved):
Nice.

Issue #4201, opened 2024-10-31
send queue: allow editing/aborting a media upload


bnjbvr:
Followup of https://github.com/matrix-org/matrix-rust-sdk/issues/1732:

- [ ] cancel an upload.
  - this needs to interrupt an upload, if it's still ongoing.
- [ ] edit caption / formatted_caption / filename

PR #4262, opened 2024-11-13
feat(send queue): allow aborting a media upload

bnjbvr:
This is a bit tricky, because we need to care about all possible states of the sending state machine, and do something different in each case:

- has the thumbnail upload started yet? y/n
- has the file upload staretd yet? y/n
- has the media event been sent yet? y/n

Fortunately, it's pretty easy to test all the cases, so I've done just that. Many LOCs are just tests, so don't be too afraid by it.

Part of #4201.

### DIFF
--- a/crates/matrix-sdk/src/send_queue.rs
+++ b/crates/matrix-sdk/src/send_queue.rs
@@ -161,7 +161,7 @@ use ruma::{
     serde::Raw,
     OwnedEventId, OwnedRoomId, OwnedTransactionId, TransactionId,
 };
-use tokio::sync::{broadcast, Notify, RwLock};
+use tokio::sync::{broadcast, oneshot, Notify, RwLock};
 use tracing::{debug, error, info, instrument, trace, warn};
 
 #[cfg(feature = "e2e-encryption")]
@@ -542,7 +542,7 @@ impl RoomSendQueue {
                 continue;
             }
 
-            let queued_request = match queue.peek_next_to_send().await {
+            let (queued_request, cancel_upload_rx) = match queue.peek_next_to_send().await {
                 Ok(Some(request)) => request,
 
                 Ok(None) => {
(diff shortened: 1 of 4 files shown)


crates/matrix-sdk/src/send_queue.rs
```diff
@@ -833,7 +833,7 @@ struct QueueStorage {
     /// All the queued requests that are being sent at the moment.
     ///
     /// It also serves as an internal lock on the storage backend.
-    being_sent: Arc<RwLock<BTreeSet<OwnedTransactionId>>>,
+    being_sent: Arc<RwLock<Option<OwnedTransactionId>>>,
```
jmartinesp: Maybe the doc comment doesn't match anymore?

crates/matrix-sdk/src/send_queue.rs
```diff
@@ -822,6 +852,16 @@ struct RoomSendQueueInner {
     _task: JoinHandle<()>,
 }
 
+/// Information about a request being sent right this moment.
+struct BeingSentInfo {
+    /// Transaction id of the thing being sent.
+    transaction_id: OwnedTransactionId,
+
+    /// For an upload request, a trigger to cancel the upload before it
+    /// completed.
```
jmartinesp, line 861: Minor typo maybe? Either 'it's completed' or 'it completes'.

crates/matrix-sdk/src/send_queue/upload.rs
```diff
@@ -398,4 +402,132 @@ impl QueueStorage {
...
+            if store.remove_send_queue_request(&self.room_id, &thumbnail_txn).await? {
+                // The thumbnail upload existed as a request: either it was pending (something
+                // else was being sent), or it was actively being sent.
+                trace!("could remove thumbnail request, removing 2 dependent requests now");
+
+                // 1. Try to abort sending using the being_sent info, in case it was active.
+                if let Some(info) = being_sent.as_ref() {
+                    if info.transaction_id == *thumbnail_txn {
+                        // SAFETY: we knew it was Some(), two lines above.
+                        let info = being_sent.take().unwrap();
```
jmartinesp, line 442: Rookie Rust dev question: why do we need another `info` here?

jmartinesp (approved):
LGTM, just a couple of minor comments and a question!

crates/matrix-sdk/src/send_queue/upload.rs
bnjbvr: Good question! Above it was taken `as_ref()`, so a `&BeingSentInfo`; the cancellation requires triggering the `oneshot::channel`, which consumes it by ownership, so we need to take `BeingSentInfo` by ownership here; at this point it's fine because we know it's the last use. I'll add a comment and maybe a tiny function helper to make this clearer.

jmartinesp: Ah, I see, I didn't realise the channel is consumed 👌 .

crates/matrix-sdk/src/send_queue.rs
bnjbvr: Good point!

PR #4269, opened 2024-11-14
task(send queue): demote some assertions back to logged errors

bnjbvr:
Better safe than panicky. I don't want a bad surprise when coming back from my weekend.

### DIFF
--- a/crates/matrix-sdk/src/send_queue.rs
+++ b/crates/matrix-sdk/src/send_queue.rs
@@ -958,7 +958,13 @@ impl QueueStorage {
                 transaction_id: request.transaction_id.clone(),
                 cancel_upload: cancel_upload_tx,
             });
-            assert!(prev.is_none());
+
+            if let Some(prev) = prev {
+                error!(
+                    prev_txn = ?prev.transaction_id,
+                    "a previous request was still active while picking a new one"
+                );
+            }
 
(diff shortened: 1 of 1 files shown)


crates/matrix-sdk/src/send_queue.rs
```diff
@@ -1023,10 +1031,11 @@ impl QueueStorage {
         // Keep the lock until we're done touching the storage.
         let mut being_sent = self.being_sent.write().await;
         let was_being_sent = being_sent.take();
-        assert_eq!(
-            was_being_sent.as_ref().map(|info| info.transaction_id.as_ref()),
-            Some(transaction_id)
-        );
+
+        let prev_txn = was_being_sent.as_ref().map(|info| info.transaction_id.as_ref());
+        if prev_txn != Some(transaction_id) {
```
poljar, line 1036: The assert would have aborted execution here, while we now continue on, is that fine? Don't we want to perhaps return early?

poljar (approved):

crates/matrix-sdk/src/send_queue.rs
bnjbvr: I think it would be a sign that the system is in a weird state, but the transaction_id passed as a parameter is the important value that is correct above all, so the current code is fine.

PR #4272, opened 2024-11-15
feat(room_preview): Use room directory as another data source

jmartinesp:
For getting a `RoomPreview` we have the following paths, unfolding from `Client::get_room_preview`.

- If the room is known, get the preview from its `RoomInfo`.
- If it's not known:
  - Try using [MSC3266](https://github.com/deepbluev7/matrix-doc/blob/room-summaries/proposals/3266-room-summary.md) to get a preview.
  - If MSC3266 is not enabled, try asking your homeserver for some state and room member data.

The 'not known' part may fail oftentimes if the MSC is not enabled, so we're trying to complement it with the room directory search for public rooms, which will be added as another possible fallback.

The way this works is:

- If we have a room alias, use it for the search, take the first 20 results and look for a room with the expected room id.
- If we don't have it, request 100 rooms and try to find a matching room as a best effort (this may work for small HS, but definitely not for large ones).

In case we have several `via` parameters, we'll start a search in each of them and stop as soon as one of them returns valid results.

Signed-off-by:

### DIFF
--- a/crates/matrix-sdk/src/room_preview.rs
+++ b/crates/matrix-sdk/src/room_preview.rs
@@ -18,9 +18,11 @@
 //! This offers a few capabilities for previewing the content of the room as
 //! well.
 
+use futures_util::future::join_all;
 use matrix_sdk_base::{RoomInfo, RoomState};
 use ruma::{
     api::client::{membership::joined_members, state::get_state_events},
+    directory::PublicRoomJoinRule,
     events::room::{history_visibility::HistoryVisibility, join_rules::JoinRule},
     room::RoomType,
     space::SpaceRoomJoinRule,
@@ -29,7 +31,7 @@ use ruma::{
 use tokio::try_join;
 use tracing::{instrument, warn};
 
-use crate::{Client, Room};
+use crate::{room_directory_search::RoomDirectorySearch, Client, Room};
 
 /// The preview of a room, be it invited/joined/left, or not.
 #[derive(Debug, Clone)]
@@ -140,19 +142,74 @@ impl RoomPreview {
     ) -> crate::Result<Self> {
         // Use the room summary endpoint, if available, as described in
         // https://github.com/deepbluev7/matrix-doc/blob/room-summaries/proposals/3266-room-summary.md
-        match Self::from_room_summary(client, room_id.clone(), room_or_alias_id, via).await {
+        match Self::from_room_summary(client, room_id.clone(), room_or_alias_id, via.clone()).await
+        {
             Ok(res) => return Ok(res),
             Err(err) => {
                 warn!("error when previewing room from the room summary endpoint: {err}");
             }
         }
 
-        // TODO: (optimization) Use the room search directory, if available:
-        // - if the room directory visibility is public,
-        // - then use a public room filter set to this room id
+        // Try room directory search next.
+        match Self::from_room_directory_search(client, &room_id, room_or_alias_id, via).await {
+            Ok(Some(res)) => return Ok(res),
+            Ok(None) => warn!("Room '{room_or_alias_id}' not found in room directory search."),
+            Err(err) => {
+                warn!("Searching for '{room_or_alias_id}' in room directory search failed: {err}");
+            }
+        }
 
         // Resort to using the room state endpoint, as well as the joined members one.
         Self::from_state_events(client, &room_id).await
+    }
+
+    /// Get a [`RoomPreview`] by searching in the room directory for the
+    /// provided room alias or room id and transforming the [`RoomDescription`]
+    /// into a preview.
+    pub(crate) async fn from_room_directory_search(
+        client: &Client,
+        room_id: &RoomId,
+        room_or_alias_id: &RoomOrAliasId,
+        via: Vec<OwnedServerName>,
+    ) -> crate::Result<Option<Self>> {
+        // Get either the room alias or the room id without the leading identifier char
+        let search_term = if room_or_alias_id.is_room_alias_id() {
+            Some(room_or_alias_id.as_str()[1..].to_owned())
+        } else {
+            None
+        };
+
+        // If we have no alias, filtering using a room id is impossible, so just take
+        // the first 100 results and try to find the current room #YOLO
+        let batch_size = if search_term.is_some() { 20 } else { 100 };
+
+        if via.is_empty() {
+            // Just search in the current homeserver
+            search_for_room_preview_in_room_directory(
+                client.clone(),
+                search_term,
+                batch_size,
+                None,
+                room_id,
+            )
+            .await
+        } else {
+            let mut futures = Vec::new();
+            // Search for all servers and retrieve the results
+            for server in via {
+                futures.push(search_for_room_preview_in_room_directory(
+                    client.clone(),
+                    search_term.clone(),
+                    batch_size,
+                    Some(server),
+                    room_id,
+                ));
+            }
+
+            let joined_results = join_all(futures).await;
+
+            Ok(joined_results.into_iter().flatten().next().flatten())
+        }
     }
 
     /// Get a [`RoomPreview`] using MSC3266, if available on the remote server.
@@ -258,4 +315,48 @@ impl RoomPreview {
             state,
         ))
     }
+}
+
+async fn search_for_room_preview_in_room_directory(
+    client: Client,
+    filter: Option<String>,
+    batch_size: u32,
+    server: Option<OwnedServerName>,
+    expected_room_id: &RoomId,
+) -> crate::Result<Option<RoomPreview>> {
+    let mut directory_search = RoomDirectorySearch::new(client);
+    directory_search.search(filter, batch_size, server).await?;
+
+    let (results, _) = directory_search.results();
+
+    for room_description in results {
+        // Iterate until we find a room description with a matching room id
+        if room_description.room_id != expected_room_id {
+            continue;
+        }
+        return Ok(Some(RoomPreview {
+            room_id: room_description.room_id,
+            canonical_alias: room_description.alias,
+            name: room_description.name,
+            topic: room_description.topic,
+            avatar_url: room_description.avatar_url,
+            num_joined_members: room_description.joined_members,
+            num_active_members: None,
+            // Assume it's a room
+            room_type: None,
+            join_rule: match room_description.join_rule {
+                PublicRoomJoinRule::Public => SpaceRoomJoinRule::Public,
+                PublicRoomJoinRule::Knock => SpaceRoomJoinRule::Knock,
+                PublicRoomJoinRule::_Custom(rule) => SpaceRoomJoinRule::_Custom(rule),
+                _ => {
+                    panic!("Unexpected PublicRoomJoinRule {:?}", room_description.join_rule)
+                }
+            },
+            is_world_readable: room_description.is_world_readable,
+            state: None,
+            is_direct: None,
+        }));
+    }
+
+    Ok(None)
 }
--- a/crates/matrix-sdk/src/test_utils/mocks.rs
+++ b/crates/matrix-sdk/src/test_utils/mocks.rs
@@ -17,24 +17,31 @@
 
 #![allow(missing_debug_implementations)]
 
-use std::sync::{Arc, Mutex};
+use std::{
+    collections::BTreeMap,
+    sync::{Arc, Mutex},
+};
 
 use matrix_sdk_base::{deserialized_responses::TimelineEvent, store::StoreConfig, SessionMeta};
 use matrix_sdk_test::{
     test_json, InvitedRoomBuilder, JoinedRoomBuilder, KnockedRoomBuilder, LeftRoomBuilder,
     SyncResponseBuilder,
 };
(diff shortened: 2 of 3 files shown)


crates/matrix-sdk/src/room_preview.rs
```diff
@@ -119,7 +123,7 @@ impl RoomPreview {
 
     /// Create a room preview from a known room (i.e. one we've been invited to,
     /// we've joined or we've left).
-    pub(crate) async fn from_known(room: &Room) -> Self {
+    pub async fn from_known(room: &Room) -> Self {
```
bnjbvr, line 126: Why is this public for everyone now? This is an internal method called from `get_room_preview`, and I think it should remain as such.

crates/matrix-sdk/src/room_preview.rs
```diff
@@ -140,21 +144,76 @@ impl RoomPreview {
...
 
         // Resort to using the room state endpoint, as well as the joined members one.
         Self::from_state_events(client, &room_id).await
     }
 
+    /// Get a [`RoomPreview`] by searching in the room directory for the
+    /// provided room alias or room id and transforming the [`RoomDescription`]
+    /// into a preview.
+    pub(crate) async fn from_room_directory_search(
+        client: &Client,
```
bnjbvr, line 171: nit: since the only thing done with this client is cloning it, let it be passed by ownership here (`Client`, not `&Client`), so the caller decides how to clone it instead. Doing this in general helps avoiding deep clones, in case the caller could just pass it by ownership.

crates/matrix-sdk/src/room_preview.rs
```diff
@@ -140,21 +144,76 @@ impl RoomPreview {
...
+        room_id: &RoomId,
+        room_or_alias_id: &RoomOrAliasId,
+        via: Vec<OwnedServerName>,
+    ) -> crate::Result<Self> {
+        let mut directory_search = RoomDirectorySearch::new(client.clone());
+
+        // Get either the room alias or the room id without the leading identifier char
+        let search_term = if room_or_alias_id.is_room_alias_id() {
+            Some(room_or_alias_id.to_string().chars().skip(1).collect::<String>())
```
bnjbvr, line 180: I appreciate the use of `chars()`, but we know the first char is `!` or `#`, which both are in the ASCII space, so we can likely use a substring method and avoid a few allocations here.

crates/matrix-sdk/src/room_preview.rs
```diff
@@ -140,21 +144,76 @@ impl RoomPreview {
...
+
+        // If we have no alias, filtering using a room id is impossible, so just take
+        // the first 100 results and try to find the current room #YOLO
+        let batch_size = if search_term.is_some() { 20 } else { 100 };
+
+        if via.is_empty() {
+            // Just search in the current homeserver
+            let _ = directory_search.search(search_term, batch_size, None).await;
+        } else {
+            // Iterate all servers until one returns some results
```
bnjbvr, line 193: We can do better: for each server, create a future that will both search and try to find the room; then either `join_all` (to wait for all results), or find a better strategy that waits for any positive result (bonus points if you do this, as there's likely no prebuilt method to do just that) :)

crates/matrix-sdk/src/room_preview.rs
```diff
@@ -258,4 +317,31 @@ impl RoomPreview {
             state,
         ))
     }
+
+    pub(crate) fn from_room_description(room_description: RoomDescription) -> Self {
```
bnjbvr, line 321: Can you inline this? It's called only once.

crates/matrix-sdk/src/room_preview.rs
```diff
@@ -258,4 +317,31 @@ impl RoomPreview {
...
+            room_type: None,
+            join_rule: space_rule_from_public_rule(room_description.join_rule),
+            is_world_readable: room_description.is_world_readable,
+            state: None,
+            is_direct: None,
+        }
+    }
+}
+
+fn space_rule_from_public_rule(join_rule: PublicRoomJoinRule) -> SpaceRoomJoinRule {
```
bnjbvr, line 340: Ditto, this is called only once, can you inline this please?

crates/matrix-sdk/src/test_utils/mocks.rs
```diff
@@ -452,6 +459,12 @@ impl MatrixMockServer {
             Mock::given(method("PUT")).and(path_regex(r"/_matrix/client/v3/directory/room/.*"));
         MockEndpoint { mock, server: &self.server, endpoint: CreateRoomAliasEndpoint }
     }
+
+    /// Create a prebuilt mock for creating room aliases.
```
bnjbvr, line 463: I think Damir wants doc tests for these methods :eyes:

testing/matrix-sdk-integration-testing/src/tests/sliding_sync/room.rs
```diff
@@ -1092,6 +1096,90 @@ async fn test_room_preview() -> Result<()> {
...
+        (
+            via_1.to_owned(),
+            vec![PublicRoomsChunkInit {
+                num_joined_members: uint!(0),
+                room_id: expected_room_id.to_owned(),
+                world_readable: true,
+                guest_can_join: true,
+            }
+            .into()],
+        ),
```
bnjbvr, line 1163: Can you put this one as the second vector element, so that the directory search has to iterate at least once?

testing/matrix-sdk-integration-testing/src/tests/sliding_sync/room.rs
```diff
@@ -1092,6 +1096,90 @@ async fn test_room_preview() -> Result<()> {
...
+        (via_2.to_owned(), Vec::new()),
```
bnjbvr, line 1164: Can you provide a dummy value for another room, to make sure it's not selected over the correct room?

testing/matrix-sdk-integration-testing/src/tests/sliding_sync/room.rs
```diff
@@ -1092,6 +1096,90 @@ async fn test_room_preview() -> Result<()> {
...
+    ]);
+    server
+        .mock_public_rooms()
+        .with_via_params(via_map)
+        // Called only once as the first response returns the room, so no need to check other
+        // homeservers
+        .expect(1)
```
bnjbvr, line 1171: In this case you can even call `mock_once`, which is both "expect to be called once, and mount at most once", aka it must be called exactly once, not less, not more.

But I think that with my other comment, this should change, and have expect(2)?

bnjbvr (changes requested):
Super nice, thanks. I have proposed a slightly more optimized way to get the rooms in case we have multiple via servers, perhaps it's premature and we don't need that, let me know what you think :)

jmartinesp:
I think https://github.com/matrix-org/matrix-rust-sdk/pull/4272/commits/0c19280ff3f437ed49078a8b0ddbff9a7ab53d60 should fix the issues you found 🤞 .

crates/matrix-sdk/src/room_preview.rs
jmartinesp: Reverted, sorry. I think I changed it for some tests outside the crate and forgot to put it back the way it was before.

crates/matrix-sdk/src/room_preview.rs
jmartinesp: In the end I went with the `join_all` approach, nice call!

testing/matrix-sdk-integration-testing/src/tests/sliding_sync/room.rs
jmartinesp: Is this needed after the `join_all` change?

bnjbvr: probably not!

crates/matrix-sdk/src/room_preview.rs
bnjbvr: (still unresolved)

crates/matrix-sdk/src/room_preview.rs
```diff
@@ -140,21 +142,117 @@ impl RoomPreview {
...
+            None
+        };
+
+        // If we have no alias, filtering using a room id is impossible, so just take
+        // the first 100 results and try to find the current room #YOLO
+        let batch_size = if search_term.is_some() { 20 } else { 100 };
+
+        let futures = if via.is_empty() {
+            // Just search in the current homeserver
+            vec![search_homeserver(client.clone(), search_term, batch_size, None, room_id)]
```
bnjbvr, line 231: nit: let's not create the vector if there's only one homeserver to search, we can directly do: `if let Some(preview) = … { return Ok(preview) }` here

crates/matrix-sdk/src/room_preview.rs
```diff
@@ -140,21 +142,117 @@ impl RoomPreview {
...
+            futures
+        };
+
+        let joined_results = join_all(futures).await;
+
+        if let Some(preview) = joined_results.into_iter().flatten().next() {
+            return Ok(preview);
+        }
+
+        Err(crate::Error::InsufficientData)
```
bnjbvr, line 253: Let's have this function return `Result<Option<...>>`, so we can return `Ok(None)` here. It's not really an error if we couldn't find a room preview with this method.

crates/matrix-sdk/src/test_utils/mocks.rs
```diff
@@ -1033,6 +1083,59 @@ impl<'a> MockEndpoint<'a, CreateRoomAliasEndpoint> {
...
+        })));
+        MatrixMock { server: self.server, mock }
+    }
+
+    /// Returns a data endpoint for paginating the public room list with several
+    /// `via` params.
+    ///
+    /// Each `via` param must be in the `server_map` parameter, otherwise it'll
+    /// fail.
+    pub fn with_via_params(
```
bnjbvr, line 1112:
```suggestion
    pub fn ok_with_via_params(
```

testing/matrix-sdk-integration-testing/src/tests/sliding_sync/room.rs
```diff
@@ -1092,6 +1099,100 @@ async fn test_room_preview() -> Result<()> {
...
 
+#[async_test]
+async fn test_room_preview_with_room_directory_search_and_room_alias_only() {
+    let server = MatrixMockServer::new().await;
+    let client = server.client_builder().build().await;
+
+    let room_alias = room_alias_id!("#a-room:matrix.org");
+    let expected_room_id = room_id!("!a-room:matrix.org");
+
+    // Allow retrieving the room id
```
bnjbvr, line 1110:
```suggestion
    // Allow resolving the room via the room directory
```

testing/matrix-sdk-integration-testing/src/tests/sliding_sync/room.rs
```diff
@@ -1092,6 +1099,100 @@ async fn test_room_preview() -> Result<()> {
...
+
+#[async_test]
+async fn test_room_preview_with_room_directory_search_and_room_alias_only_in_several_homeservers() {
+    let server = MatrixMockServer::new().await;
+    let client = server.client_builder().build().await;
+
+    let room_alias = room_alias_id!("#a-room:matrix.org");
+    let expected_room_id = room_id!("!a-room:matrix.org");
+
+    // Allow retrieving the room id
```
bnjbvr, line 1144: ditto

crates/matrix-sdk/src/room_preview.rs
```diff
@@ -140,21 +142,117 @@ impl RoomPreview {
...
+    /// Get a [`RoomPreview`] by searching in the room directory for the
+    /// provided room alias or room id and transforming the [`RoomDescription`]
+    /// into a preview.
+    pub(crate) async fn from_room_directory_search(
+        client: &Client,
+        room_id: &RoomId,
+        room_or_alias_id: &RoomOrAliasId,
+        via: Vec<OwnedServerName>,
+    ) -> crate::Result<Self> {
+        async fn search_homeserver(
```
bnjbvr, line 174: Can this be a helper method instead of an embedded function (so as to reduce indent), and return `Result<Option<...>>` too?

bnjbvr (approved):
Sweet, thanks!

crates/matrix-sdk/src/room_preview.rs
jmartinesp: Maybe I'm wrong, but:

- If we don't pass any via servers, we just clone it once (same as if we just passed the owned value).
- If we pass servers though, we need to clone it once per `RoomDirectorySearch::new(client)` call. If we already had an owned value, we'd be calling clone an extra time unless the do something weird like `let client = if last_iteration { client } else { client.clone() };`.

crates/matrix-sdk/src/room_preview.rs
jmartinesp: Yeah, that's better, thanks!

crates/matrix-sdk/src/room_preview.rs
bnjbvr: Ah hm, we'd be cloning one extra if we have a `via` servers. Nevermind me, thanks for pushing back :)

PR #4275, opened 2024-11-18
feat(utd_hook): Report historical expected UTD with new reason

BillCarsonFr:
Fixes https://github.com/element-hq/element-meta/issues/2581 

Detect and report expected historical UTDs as such. Must be integrated by clients and reported with [this code](https://github.com/matrix-org/matrix-analytics-events/blob/564a8b1386f2f46eee65f429b2a07940916d8257/schemas/Error.json#L23) in Analytics.

To determine if an event is device historical (i.e sent before the current login existed) we use an heuristic based on the event `origin_server_ts` and the device creation local timestamp.

Signed-off-by:

### DIFF
--- a/crates/matrix-sdk-crypto/src/types/events/utd_cause.rs
+++ b/crates/matrix-sdk-crypto/src/types/events/utd_cause.rs
@@ -15,7 +15,7 @@
 use matrix_sdk_common::deserialized_responses::{
     UnableToDecryptInfo, UnableToDecryptReason, VerificationLevel,
 };
-use ruma::{events::AnySyncTimelineEvent, serde::Raw};
+use ruma::{events::AnySyncTimelineEvent, serde::Raw, MilliSecondsSinceUnixEpoch};
 use serde::Deserialize;
 
 /// Our best guess at the reason why an event can't be decrypted.
@@ -47,6 +47,16 @@ pub enum UtdCause {
     /// data was obtained from an insecure source (imported from a file,
     /// obtained from a legacy (asymmetric) backup, unsafe key forward, etc.)
     UnknownDevice = 4,
+
+    /// We are missing the keys for this event, but it is a "device-historical"
+    /// message and no backup is accessible or usable.
+    ///
+    /// Device-historical means that the message was sent before the current
+    /// device existed (but the current user was probably a member of the room
+    /// at the time the message was sent). Not to
+    /// be confused with pre-join or pre-invite messages (see
+    /// [`UtdCause::SentBeforeWeJoined`] for that).
+    HistoricalMessage = 5,
 }
 
 /// MSC4115 membership info in the unsigned area.
@@ -65,10 +75,25 @@ enum Membership {
     Join,
 }
 
+/// Contextual crypto information used by [`UtdCause::determine`] to properly
+/// identify an Unable-To-Decrypt cause in addition to the
+/// [`UnableToDecryptInfo`] and raw event info.
+#[derive(Debug, Clone, Copy)]
+pub struct CryptoContextInfo {
+    /// The current device creation timestamp, used as a heuristic to determine
+    /// if an event is device-historical or not (sent before the current device
+    /// existed).
+    pub device_creation_ts: MilliSecondsSinceUnixEpoch,
+    /// True if key storage is correctly set up and can be used by the current
+    /// client to download and decrypt message keys.
+    pub is_backup_configured: bool,
+}
+
 impl UtdCause {
     /// Decide the cause of this UTD, based on the evidence we have.
     pub fn determine(
-        raw_event: Option<&Raw<AnySyncTimelineEvent>>,
+        raw_event: &Raw<AnySyncTimelineEvent>,
+        crypto_context_info: CryptoContextInfo,
         unable_to_decrypt_info: &UnableToDecryptInfo,
     ) -> Self {
         // TODO: in future, use more information to give a richer answer. E.g.
@@ -76,16 +101,27 @@ impl UtdCause {
             UnableToDecryptReason::MissingMegolmSession
             | UnableToDecryptReason::UnknownMegolmMessageIndex => {
                 // Look in the unsigned area for a `membership` field.
-                if let Some(raw_event) = raw_event {
-                    if let Ok(Some(unsigned)) =
-                        raw_event.get_field::<UnsignedWithMembership>("unsigned")
-                    {
-                        if let Membership::Leave = unsigned.membership {
-                            // We were not a member - this is the cause of the UTD
-                            return UtdCause::SentBeforeWeJoined;
-                        }
+                if let Some(unsigned) =
+                    raw_event.get_field::<UnsignedWithMembership>("unsigned").ok().flatten()
+                {
+                    if let Membership::Leave = unsigned.membership {
+                        // We were not a member - this is the cause of the UTD
+                        return UtdCause::SentBeforeWeJoined;
                     }
                 }
+
+                if let Ok(timeline_event) = raw_event.deserialize() {
+                    if crypto_context_info.is_backup_configured
+                        && timeline_event.origin_server_ts()
+                            < crypto_context_info.device_creation_ts
+                    {
+                        // It's a device-historical message and there is no accessible
+                        // backup. The key is missing and it
+                        // is expected.
+                        return UtdCause::HistoricalMessage;
+                    }
+                }
+
                 UtdCause::Unknown
             }
 
@@ -111,10 +147,10 @@ mod tests {
     use matrix_sdk_common::deserialized_responses::{
         DeviceLinkProblem, UnableToDecryptInfo, UnableToDecryptReason, VerificationLevel,
     };
-    use ruma::{events::AnySyncTimelineEvent, serde::Raw};
+    use ruma::{events::AnySyncTimelineEvent, serde::Raw, uint, MilliSecondsSinceUnixEpoch};
     use serde_json::{json, value::to_raw_value};
 
-    use crate::types::events::UtdCause;
+    use crate::types::events::{utd_cause::CryptoContextInfo, UtdCause};
 
     #[test]
     fn test_a_missing_raw_event_means_we_guess_unknown() {
@@ -122,7 +158,8 @@ mod tests {
         // is unknown.
         assert_eq!(
             UtdCause::determine(
-                None,
+                &raw_event(json!({})),
+                some_crypto_context_info(),
                 &UnableToDecryptInfo {
                     session_id: None,
                     reason: UnableToDecryptReason::MissingMegolmSession,
@@ -137,7 +174,8 @@ mod tests {
         // If our JSON contains no membership info, then we guess the UTD is unknown.
         assert_eq!(
             UtdCause::determine(
-                Some(&raw_event(json!({}))),
+                &raw_event(json!({})),
+                some_crypto_context_info(),
                 &UnableToDecryptInfo {
                     session_id: None,
                     reason: UnableToDecryptReason::MissingMegolmSession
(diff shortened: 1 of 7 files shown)


crates/matrix-sdk-ui/src/timeline/event_handler.rs
```diff
@@ -420,7 +421,11 @@ impl<'a, 'o> TimelineEventHandler<'a, 'o> {
             TimelineEventKind::UnableToDecrypt { content, unable_to_decrypt_info } => {
                 // TODO: Handle replacements if the replaced event is also UTD
                 let raw_event = self.ctx.flow.raw_event();
-                let cause = UtdCause::determine(raw_event, &unable_to_decrypt_info);
+                let client_info = match self.meta.unable_to_decrypt_hook.as_ref() {
+                    Some(hook) => Some(hook.client_info().await),
+                    None => None,
+                };
```
richvdh, line 427: This doesn't belong in the UTD hook. (For example: the `cause` passed to `TimelineItemContent::unable_to_decrypt` should not depend on whether a UTD hook has been registered or not). We need to find another way to access the `ClientInfo` here.

How about:
 * Add a `client_info` method to `RoomDataProvider` (it is implemented by `Room`, which has a reference to the Client, so that should be easy)
 * Add a `UtdCause` filed to `matrix_sdk_ui::timeline::event_handler::TimelineEventKind::UnableToDecrypt`
 * Change `TimelineEventKind::from_event` to take a `RoomDataProvider` instead of a `RoomVersionId`
 * Call `UtdCause::determine()` in `TimelineEventKind::from_event`, to populate the UtdCause

richvdh (changes requested):

crates/matrix-sdk-ui/src/timeline/event_handler.rs
richvdh: (incidentally, we may want to think of a more descriptive name than "ClientInfo" here)

BillCarsonFr: I have implemented some changes, better to look at commit per commit
- https://github.com/matrix-org/matrix-rust-sdk/pull/4275/commits/db1fadeb51f72fe7aeededdce389e4e0cc96b364 Moved utd_cause to the sdk-ui UnableToDecrypt kind
- https://github.com/matrix-org/matrix-rust-sdk/pull/4275/commits/a4092232a49375577b9a11cf5febbc4946399957 As discussed the unable_to_decrypt_info is not needed in the ui crate, the utd_cause is enough
- https://github.com/matrix-org/matrix-rust-sdk/pull/4275/commits/0538bcac347a9013e543b646de7eb5e502771b10 Rename client info
- https://github.com/matrix-org/matrix-rust-sdk/pull/4275/commits/749c0d1a018fcf05efcc9bc484f227ce48a15d11 the e2e feature flag is not visible in the UI crate, so made the API returns an optional `CryptoContextInfo` and return None in sdk if feature not enabled

crates/matrix-sdk-common/src/types.rs
richvdh, line 1: I'm not convinced it makes sense to make a new "types" module for this. Isn't pretty much anything a `type` in Rust?

crates/matrix-sdk-ui/src/timeline/traits.rs
```diff
@@ -76,7 +76,7 @@ pub(super) trait RoomDataProvider:
     fn own_user_id(&self) -> &UserId;
     fn room_version(&self) -> RoomVersionId;
 
-    fn crypto_context_info(&self) -> BoxFuture<'_, CryptoContextInfo>;
+    fn crypto_context_info(&self) -> BoxFuture<'_, Option<CryptoContextInfo>>;
```
richvdh, line 79: If you're going to keep this, please can you document what `None` and `Some` mean?

crates/matrix-sdk/src/room/mod.rs
```diff
@@ -614,15 +616,22 @@ impl Room {
...
-        CryptoContextInfo {
-            device_creation_ts: match self.client.encryption().get_own_device().await {
-                Ok(Some(device)) => device.first_time_seen_ts(),
-                // Should not happen, there will always be an own device
-                _ => MilliSecondsSinceUnixEpoch::now(),
-            },
-            is_backup_configured: self.client.encryption().backups().state()
-                == BackupState::Enabled,
+    pub async fn crypto_context_info(&self) -> Option<CryptoContextInfo> {
+        #[cfg(feature = "e2e-encryption")]
```
richvdh, line 620: Given that we're only going to call this if the `e2e-encryption` feature is enabled (I think?) you could instead do:

```suggestion
        #[cfg(feature = "e2e-encryption")]
        pub async fn crypto_context_info(&self) -> CryptoContextInfo {
           ...
        }
```

(and likewise in the `RoomDataProvider` trait)

... which would mean that the `CryptoContextInfo` type could stay put in `matrix-sdk-crypto`.

crates/matrix-sdk-crypto/src/types/events/utd_cause.rs
```diff
@@ -47,6 +47,10 @@ pub enum UtdCause {
     /// data was obtained from an insecure source (imported from a file,
     /// obtained from a legacy (asymmetric) backup, unsafe key forward, etc.)
     UnknownDevice = 4,
+
+    /// We are missing the keys for this event, but it is an historical message
+    /// and no backup is accessible or usable.
```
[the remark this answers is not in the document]
richvdh: bump

crates/matrix-sdk/src/room/mod.rs
```diff
@@ -610,6 +613,19 @@ impl Room {
         Ok(self.inner.is_encrypted())
     }
 
+    /// todo add the doc
```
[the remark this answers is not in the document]
richvdh: oh, you did the todo, but in a different commit :(

crates/matrix-sdk-ui/src/timeline/event_handler.rs
```diff
@@ -199,7 +202,12 @@ impl TimelineEventKind {
                 Some(AnyMessageLikeEventContent::RoomEncrypted(content)) => {
                     // An event which is still encrypted.
                     if let Some(unable_to_decrypt_info) = unable_to_decrypt_info {
-                        Self::UnableToDecrypt { content, unable_to_decrypt_info }
+                        let utd_cause = UtdCause::determine(
+                            Some(raw_event),
+                            room_data_provider.client_info().await.into(),
```
richvdh: what is the `into()` doing here?

crates/matrix-sdk-ui/src/timeline/event_handler.rs
```diff
@@ -199,7 +202,12 @@ impl TimelineEventKind {
                 Some(AnyMessageLikeEventContent::RoomEncrypted(content)) => {
                     // An event which is still encrypted.
                     if let Some(unable_to_decrypt_info) = unable_to_decrypt_info {
-                        Self::UnableToDecrypt { content, unable_to_decrypt_info }
+                        let utd_cause = UtdCause::determine(
+                            Some(raw_event),
```
richvdh, line 202: given that `UtdCause::determine` is only called here (I think?) we could make `raw_event` be not-`Option`

crates/matrix-sdk-ui/src/timeline/event_handler.rs
```diff
@@ -182,14 +183,16 @@ pub(super) enum TimelineEventKind {
 
 impl TimelineEventKind {
     /// Creates a new `TimelineEventKind` with the given event and room version.
-    pub fn from_event(
+    pub async fn from_event<P: RoomDataProvider>(
         event: AnySyncTimelineEvent,
-        room_version: &RoomVersionId,
+        raw_event: &Raw<AnySyncTimelineEvent>,
```
richvdh, line 184: ugh, having to pass both `event` and `raw_event` is a bit ugly, but it might have to do for now. Let's at least document them though (ie, that `event` is the deserialized `raw_event`)

crates/matrix-sdk-common/src/types.rs
richvdh: see [below](https://github.com/matrix-org/matrix-rust-sdk/pull/4275#discussion_r1858613525): I think we can leave it in the crypto crate

crates/matrix-sdk-ui/src/timeline/event_handler.rs
BillCarsonFr: outdated no more an Option<>

crates/matrix-sdk-ui/src/timeline/traits.rs
```diff
@@ -121,6 +123,10 @@ impl RoomDataProvider for Room {
         (**self).clone_info().room_version_or_default()
     }
 
+    fn client_info(&self) -> BoxFuture<'_, ClientInfo> {
+        async move { self.client_info().await }.boxed()
+    }
```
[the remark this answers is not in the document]
BillCarsonFr: No the struct method has priority other the Trait method. If I wanted to call the trait method I'd have to explicitely do `RoomDataProvider::crypto_context_info(&room)` or maybe if the type was something like `&dyn RoomDataProvider` and not Room.

To avoid confusion maybe we could rename one of the method (adding a `get_` prefix in the trait)?
But it is not the same occurence of that in this file, see `load_user_receipt`

crates/matrix-sdk-ui/src/timeline/traits.rs
BillCarsonFr: Not an Option anymore https://github.com/matrix-org/matrix-rust-sdk/pull/4275/commits/85b907e6a415a5206c466dc012ed93dde72b2464

crates/matrix-sdk/src/room/mod.rs
BillCarsonFr: [revert CryptoContetInfo optional as the ui crate requires e2ee](https://github.com/matrix-org/matrix-rust-sdk/pull/4275/commits/85b907e6a415a5206c466dc012ed93dde72b2464) 
 then https://github.com/matrix-org/matrix-rust-sdk/pull/4275/commits/0f87ea9d3eb58836ccd080092947dd971d992190

crates/matrix-sdk-ui/src/timeline/event_handler.rs
BillCarsonFr: Added some documentation:
```
        // We need also the [`Raw`] version of the `event` because unknown json fields are lost
        // when de-serialising to the actual `AnySyncTimelineEvent` type. And we need some
        // of these field to determine the `UtdCause`.
 ```
 
 I wonder if we couldn't pass both of them to `UtdCause::Determine`? Would avoid a deserialisation there, or maybe we could raw access the `origin_server_ts`

File: crates/matrix-sdk-ui/src/timeline/event_handler.rs (hunk 7 above)
Comment (BillCarsonFr) [reply crates/matrix-sdk-ui/src/timeline/event_handler.rs]: Done https://github.com/matrix-org/matrix-rust-sdk/pull/4275/commits/38f1b607d3e07ae1842594e08a415b6976ffb0b5

File: crates/matrix-sdk-ui/src/timeline/event_handler.rs [hunk 10]
```diff
@@ -181,14 +179,19 @@ pub(super) enum TimelineEventKind {
 
 impl TimelineEventKind {
     /// Creates a new `TimelineEventKind` with the given event and room version.
```
richvdh, line 181: We no longer provide a room version, and this is the place to document the arguments.

```suggestion
    /// Creates a new `TimelineEventKind`.
    ///
    /// # Arguments
    ///
    /// * `event` - The event for which we should create a `TimelineEventKind`.
    /// * `raw_event` - The [`Raw`] JSON for `event`. (Required so that we can access `unsigned` data.)
    /// * `room_data_provider` - An object which will provide information about the room containing the event.
    /// * `unable_to_decrypt_info` - If `event` represents a failure to decrypt, information about that failure. Otherwise, `None`.
```

crates/matrix-sdk-ui/src/timeline/event_handler.rs
```diff
@@ -181,14 +179,19 @@ pub(super) enum TimelineEventKind {
...
-    pub fn from_event(
+    pub async fn from_event<P: RoomDataProvider>(
         event: AnySyncTimelineEvent,
-        room_version: &RoomVersionId,
+        // We need also the [`Raw`] version of the `event` because unknown json fields are lost
+        // when de-serialising to the actual `AnySyncTimelineEvent` type. And we need some
+        // of these field to determine the `UtdCause`.
```
richvdh, line 186: Remove in favour of the above

```suggestion
```

PR #4276, opened 2024-11-18
fix(Room): Use `subscribe_reset` for `subscribe_info` (instead of just `subscribe`)

toger5:
When calling `subscribe_info` users will most likely expect to get the info and not to wait for the next info change.

Has been discussed with @bnjbvr.

Signed-off-by:

### DIFF
--- a/crates/matrix-sdk-base/src/rooms/normal.rs
+++ b/crates/matrix-sdk-base/src/rooms/normal.rs
@@ -856,8 +856,11 @@ impl Room {
     }
 
     /// Subscribe to the inner `RoomInfo`.
+    ///
+    /// Behaves like a `subscribe_reset`:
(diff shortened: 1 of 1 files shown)


bnjbvr (commented):
OK on principle, but this is a draft and not passing tests, so this will require a bit more work.

toger5:
Because this is an intentionally breaking change, we just adapt the tests right?

bnjbvr:
Yes! Would need to make it super explicit that it's a breaking change, please have a look at our contribution guidelines :)

poljar:
What's the status on this? Can we close it if you don't have the time to work on this?

toger5:
Yes I do not have time for this. Sadly. So closing makes sense. Or someone else can reopen and finish it.

PR #4277, opened 2024-11-18
feat(send queue): allow editing media captions

bnjbvr:
This allows editing media captions:

- for remote echo'd events, adding support in `make_edit_event` is sufficient,
- for local events (that are in the send queue), a bit more work is necessary:
  - either the media event is a dependent request, and we need to update it
  - or the media event is a request, and we need to update it
  - or the media event is being sent, and we'll record a dependent request for an edit

Fixes https://github.com/matrix-org/matrix-rust-sdk/issues/4201

### DIFF
--- a/crates/matrix-sdk-base/src/store/traits.rs
+++ b/crates/matrix-sdk-base/src/store/traits.rs
@@ -437,6 +437,16 @@ pub trait StateStore: AsyncTraitDeps {
         sent_parent_key: SentRequestKey,
     ) -> Result<usize, Self::Error>;
 
+    /// Update a dependent send queue request with the new content.
+    ///
+    /// Returns true if the request was found and could be updated.
+    async fn update_dependent_queued_request(
+        &self,
+        room_id: &RoomId,
+        own_transaction_id: &ChildTransactionId,
+        new_content: DependentQueuedRequestKind,
+    ) -> Result<bool, Self::Error>;
+
     /// Remove a specific dependent send queue request by id.
     ///
     /// Returns true if the dependent send queue request has been indeed
@@ -732,6 +742,18 @@ impl<T: StateStore> StateStore for EraseStateStoreError<T> {
         room_id: &RoomId,
     ) -> Result<Vec<DependentQueuedRequest>, Self::Error> {
         self.0.load_dependent_queued_requests(room_id).await.map_err(Into::into)
+    }
+
+    async fn update_dependent_queued_request(
+        &self,
+        room_id: &RoomId,
+        own_transaction_id: &ChildTransactionId,
+        new_content: DependentQueuedRequestKind,
+    ) -> Result<bool, Self::Error> {
+        self.0
+            .update_dependent_queued_request(room_id, own_transaction_id, new_content)
+            .await
+            .map_err(Into::into)
     }
 }
 
--- a/crates/matrix-sdk-base/src/store/traits.rs
+++ b/crates/matrix-sdk-base/src/store/traits.rs
@@ -422,15 +422,15 @@ pub trait StateStore: AsyncTraitDeps {
         content: DependentQueuedRequestKind,
     ) -> Result<(), Self::Error>;
 
-    /// Update a set of dependent send queue requests with a key identifying the
-    /// homeserver's response, effectively marking them as ready.
+    /// Mark a set of dependent send queue requests as ready, using a key
+    /// identifying the homeserver's response.
     ///
     /// ⚠ Beware! There's no verification applied that the parent key type is
     /// compatible with the dependent event type. The invalid state may be
     /// lazily filtered out in `load_dependent_queued_requests`.
     ///
     /// Returns the number of updated requests.
-    async fn update_dependent_queued_request(
+    async fn mark_dependent_queued_requests_as_ready(
         &self,
         room_id: &RoomId,
         parent_txn_id: &TransactionId,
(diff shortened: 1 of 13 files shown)


crates/matrix-sdk/src/room/edit.rs
```diff
@@ -167,47 +185,47 @@ async fn make_edit_event<S: EventSource>(
...
-                        MessageLikeEvent::Original(original),
-                    )) = event.into_full_event(room_id.to_owned())
-                    {
-                        Some(original)
-                    } else {
-                        None
-                    }
+        EditedContent::MediaCaption { caption, formatted_caption } => {
+            // Handle edits of m.room.message.
+            let AnySyncMessageLikeEvent::RoomMessage(SyncMessageLikeEvent::Original(original)) =
```
stefanceriu, line 201: Shouldn't we restrict this to actual media?

crates/matrix-sdk/src/room/edit.rs
```diff
@@ -234,6 +252,101 @@ async fn make_edit_event<S: EventSource>(
...
+        } else {
+            $event.filename = None;
+            $event.body = filename;
+        }
+    };
+}
+
+/// Sets the caption of a [`RoomMessageEventContent`].
+///
+/// Returns true if the event represented a media event (and thus the captions
```
stefanceriu, line 279: Again, why a boolean and not a proper error?

crates/matrix-sdk-base/src/store/traits.rs
```diff
@@ -422,21 +422,31 @@ pub trait StateStore: AsyncTraitDeps {
...
 
+    /// Update a dependent send queue request with the new content.
+    ///
+    /// Returns true if the request was found and could be updated.
+    async fn update_dependent_queued_request(
+        &self,
+        room_id: &RoomId,
+        own_transaction_id: &ChildTransactionId,
+        new_content: DependentQueuedRequestKind,
+    ) -> Result<bool, Self::Error>;
```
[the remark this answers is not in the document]
bnjbvr: It's more flexible this way: the caller can decide whether this is an error or not, in the end.

crates/matrix-sdk-indexeddb/src/state_store/mod.rs
```diff
@@ -1570,6 +1570,48 @@ impl_state_store!({
...
+                break;
+            }
+        }
+
+        if found {
+            obj.put_key_val(&encoded_key, &self.serialize_value(&prev)?)?;
+            tx.await.into_result()?;
+        }
+
+        Ok(found)
```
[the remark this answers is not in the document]
bnjbvr: Again, I find it more flexible to let the consumer of this API decide what to do if there wasn't any change.

crates/matrix-sdk-sqlite/src/state_store.rs
```diff
@@ -1874,6 +1874,35 @@ impl StateStore for SqliteStateStore {
...
+                    r#"UPDATE dependent_send_queue_events
+                       SET content = ?
+                       WHERE own_transaction_id = ?
+                       AND room_id = ?"#,
+                )?
+                .execute((content, own_txn_id, room_id))
+            })
+            .await?;
+
+        Ok(num_updated == 1)
```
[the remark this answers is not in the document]
bnjbvr: Are you suggesting returning an error in this case, or panicking?

crates/matrix-sdk/src/room/edit.rs
bnjbvr: That happens a few lines later under the function that edits the caption.

crates/matrix-sdk-sqlite/src/state_store.rs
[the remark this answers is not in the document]
stefanceriu: I reckon an error is enough but you know best

crates/matrix-sdk/src/room/edit.rs
stefanceriu: Oh I see and then it would go on a `IncompatibleEditType`, fair enough 👍

crates/matrix-sdk/src/room/edit.rs
stefanceriu: I'm not sure I follow, how can the user of a `update_media_caption` know from the signature that the only possible error is an `not media event` one? From the outside a number of things might've gone wrong. Why go throug mental gymanstics based on *only the current implementation* when you can make it explicit through an error (that can also grow in time as need be)?

bnjbvr: I wouldn't call reading the doc comment "mental gymnastics", to be fair :-) I'm not using an error type because [YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it) — and if we need to distinguish among several error kinds later, we can add the error when we need to. I find that using a `Result` here would be optimizing for future possibilities that don't exist quite yet, or, if you prefer, a tiny act of over-engineering.

stefanceriu: As you wish then 👍

crates/matrix-sdk/src/send_queue/upload.rs
```diff
@@ -542,4 +545,121 @@ impl QueueStorage {
...
+        let Some(found) = requests.into_iter().find(|req| req.transaction_id == *txn) else {
+            // Couldn't be found anymore, it's not possible to update captions.
+            return Ok(None);
+        };
+
+        trace!("found the caption to edit as a request");
+
+        let QueuedRequestKind::Event { content: serialized_content } = found.kind else {
+            return Err(InvalidMediaCaptionEdit);
+        };
```
stefanceriu, line 621: Just for my understanding: finding a non-event kind here is a developer error and the wrong txn was passed in, right?

stefanceriu (approved):
Apart from those small comments this is all really nice and makes perfect sense to me ... and the tests are just *chef's kiss* 👌

crates/matrix-sdk/src/send_queue/upload.rs
bnjbvr: Indeed, all these errors would either be caused by a wrong txn_id, or a super inconsistent internal state :+1:

crates/matrix-sdk-sqlite/src/state_store.rs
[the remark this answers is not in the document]
bnjbvr: Made it an error, thanks :)

PR #4293, opened 2024-11-19
media: include the caption and filename for file and audio events too

bnjbvr:
An oversight in `make_attachment_type`, where the creation of file and audio events would lead to losing both the filename (if set, that is, if there was a caption — resulting in the caption being considered as the file name) and the formatted caption, if provided.

Thanks @bmarty for catching this on EX.

Part of #4201.


andybalaam (approved):
Looks good, thanks!

PR #4294, opened 2024-11-19
fix(send queue): when adding a local reaction, look for media events in dependent requests too

bnjbvr:
This allows reacting to a local media upload.

Adding a reaction to a local echo would check that the target of the reaction (identified by its transaction id) is part of the *requests*, but not part of the *dependent requests*. For a media upload, it's more likely the media event will be alive as a dependent request of kind `FinishUpload`, so we need to check there too. With that fix, it's then possible to react to a local media upload.

Part of #4201.


stefanceriu (approved):
Oh super nice, makes perfect sense to me! 👍

PR #4295, opened 2024-11-19
fix(base): clear a room's send queue and dependent event queue after removing it from the state store

bnjbvr:
Since this wasn't properly done for the dependent send queue for rooms, technically we should run a migration to look for the entries in the dependent send queue table, and remove all those that don't correspond to known room ids. I didn't bother here because the value seems quite low (this shouldn't affect many lines), but if the reviewer strongly feels about it, we could clean it up.


bnjbvr:
Thanks for the review!

It was correctly implemented for the indexeddb state store :upside_down_face: (as proven by the tests not failing without any additional changes there). So yeah, our implementations were a bit inconsistent before…

PR #4296, opened 2024-11-19
test: make the `EventFactory` available to all crates

bnjbvr:
This had been requested by crypto folks in the past, notably @BillCarsonFr if I remember correctly: it makes the `EventFactory` available from all crates, by moving it into `matrix-sdk-test`. Because the common crate depended on the test crate, and `EventFactory` depends on the `common` crate, this required cutting the circular dependency by having the common crate use the test-macro crate directly. I find it an acceptable tradeoff.

The two first commits tidy up the test utils implemented in the SDK crate. I don't think they can be moved to another external crate:

- they need concepts defined in the SDK crate, like a `Client`,
- but the SDK crate uses these mock helpers in a few places

So we'd have a circular dependency there too, if we moved this code to another crate (unless I'm missing something).


Hywan (approved):
It's an improvement, thanks. Can you rebase your history before merging please?

PR #4298, opened 2024-11-20
feat(base) Implement `RelationalLinkedChunk`

Hywan:
<figure role="presentation">
<p align="center">A flask of poison</p>
<figcaption><p align="center">A book floating in space in front of a black hole.</p></figcaption>
</figure>

---

A `RelationalLinkedChunk` is like a `LinkedChunk` but with a relational
layout, similar to what we would have in a database.

This is used by memory stores. The idea is to have a data layout that
is similar for memory stores and for relational database stores, to
represent a `LinkedChunk`.

This type is also designed to receive `Update`. Applying `Update`s
directly on a `LinkedChunk` is not ideal and particularly not trivial
as the `Update`s do _not_ match the internal data layout of the
`LinkedChunk`, they have been designed for storages, like a relational
database for example.

This type is not as performant as `LinkedChunk` (in terms of memory
layout, CPU caches etc.). It is only designed to be used in memory
stores, which are mostly used for test purposes or light usages of the
SDK.

---

* ~~Must be merged after https://github.com/matrix-org/matrix-rust-sdk/pull/4299 as it contains its commits~~

### DIFF
--- a/crates/matrix-sdk-base/src/event_cache/store/memory_store.rs
+++ b/crates/matrix-sdk-base/src/event_cache/store/memory_store.rs
@@ -20,7 +20,7 @@ use matrix_sdk_common::{
     ring_buffer::RingBuffer,
     store_locks::memory_store_helper::try_take_leased_lock,
 };
-use ruma::{MxcUri, OwnedMxcUri};
+use ruma::{MxcUri, OwnedMxcUri, RoomId};
 
 use super::{EventCacheStore, EventCacheStoreError, Result};
 use crate::{
@@ -75,9 +75,10 @@ impl EventCacheStore for MemoryStore {
 
     async fn handle_linked_chunk_updates(
         &self,
+        room_id: &RoomId,
         updates: &[Update<Event, Gap>],
     ) -> Result<(), Self::Error> {
-        self.events.write().unwrap().apply_updates(updates);
+        self.events.write().unwrap().apply_updates(room_id, updates);
 
         Ok(())
     }
--- a/crates/matrix-sdk-base/src/event_cache/store/memory_store.rs
+++ b/crates/matrix-sdk-base/src/event_cache/store/memory_store.rs
@@ -16,7 +16,8 @@ use std::{collections::HashMap, num::NonZeroUsize, sync::RwLock as StdRwLock, ti
 
 use async_trait::async_trait;
 use matrix_sdk_common::{
-    linked_chunk::Update, ring_buffer::RingBuffer,
+    linked_chunk::{relational::RelationalLinkedChunk, Update},
+    ring_buffer::RingBuffer,
     store_locks::memory_store_helper::try_take_leased_lock,
 };
 use ruma::{MxcUri, OwnedMxcUri};
@@ -35,6 +36,7 @@ use crate::{
 pub struct MemoryStore {
     media: StdRwLock<RingBuffer<(OwnedMxcUri, String /* unique key */, Vec<u8>)>>,
     leases: StdRwLock<HashMap<String, (String, Instant)>>,
+    events: StdRwLock<RelationalLinkedChunk<Event, Gap>>,
 }
 
 // SAFETY: `new_unchecked` is safe because 20 is not zero.
@@ -45,6 +47,7 @@ impl Default for MemoryStore {
         Self {
             media: StdRwLock::new(RingBuffer::new(NUMBER_OF_MEDIAS)),
             leases: Default::default(),
+            events: StdRwLock::new(RelationalLinkedChunk::new()),
         }
     }
 }
@@ -72,9 +75,11 @@ impl EventCacheStore for MemoryStore {
 
     async fn handle_linked_chunk_updates(
         &self,
-        _updates: &[Update<Event, Gap>],
+        updates: &[Update<Event, Gap>],
     ) -> Result<(), Self::Error> {
-        todo!()
+        self.events.write().unwrap().apply_updates(updates);
+
+        Ok(())
     }
 
     async fn add_media_content(
--- a/crates/matrix-sdk-base/src/event_cache/store/memory_store.rs
+++ b/crates/matrix-sdk-base/src/event_cache/store/memory_store.rs
@@ -85,11 +85,12 @@ impl EventCacheStore for MemoryStore {
     async fn handle_linked_chunk_updates(
         &self,
         room_id: &RoomId,
-        updates: &[Update<Event, Gap>],
+        updates: Vec<Update<Event, Gap>>,
     ) -> Result<(), Self::Error> {
         let mut inner = self.inner.write().unwrap();
+        inner.events.apply_updates(room_id, updates);
 
-        Ok(inner.events.apply_updates(updates))
+        Ok(())
     }
 
     async fn add_media_content(
@@ -128,7 +129,7 @@ impl EventCacheStore for MemoryStore {
     async fn get_media_content(&self, request: &MediaRequestParameters) -> Result<Option<Vec<u8>>> {
         let expected_key = request.unique_key();
 
-        let inner = self.inner.write().unwrap();
+        let inner = self.inner.read().unwrap();
 
         Ok(inner.media.iter().find_map(|(_media_uri, media_key, media_content)| {
             (media_key == &expected_key).then(|| media_content.to_owned())
--- a/crates/matrix-sdk-base/src/event_cache/store/memory_store.rs
+++ b/crates/matrix-sdk-base/src/event_cache/store/memory_store.rs
@@ -34,9 +34,14 @@ use crate::{
 #[allow(clippy::type_complexity)]
 #[derive(Debug)]
 pub struct MemoryStore {
-    media: StdRwLock<RingBuffer<(OwnedMxcUri, String /* unique key */, Vec<u8>)>>,
-    leases: StdRwLock<HashMap<String, (String, Instant)>>,
-    events: StdRwLock<RelationalLinkedChunk<Event, Gap>>,
+    inner: StdRwLock<MemoryStoreInner>,
+}
+
+#[derive(Debug)]
+struct MemoryStoreInner {
+    media: RingBuffer<(OwnedMxcUri, String /* unique key */, Vec<u8>)>,
+    leases: HashMap<String, (String, Instant)>,
+    events: RelationalLinkedChunk<Event, Gap>,
 }
 
 // SAFETY: `new_unchecked` is safe because 20 is not zero.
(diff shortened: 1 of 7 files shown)


bnjbvr (changes requested):
This code is likely wrong, given that the store method should include a reference to a `RoomId` (sorry, should've caught that during review of the patch adding the trait method, but noticed the problem only while doing the implementation for the SQLite backend…).

crates/matrix-sdk-base/src/event_cache/store/memory_store.rs
```diff
@@ -35,6 +36,7 @@ use crate::{
 pub struct MemoryStore {
     media: StdRwLock<RingBuffer<(OwnedMxcUri, String /* unique key */, Vec<u8>)>>,
     leases: StdRwLock<HashMap<String, (String, Instant)>>,
+    events: StdRwLock<RelationalLinkedChunk<Event, Gap>>,
```
bnjbvr, line 39: preexisting, but maybe an opportunity to rewrite it while we're around: could we use a single lock for all the data in that store? this would avoid lock ordering issues like we've had in the past with the other store. (So we could introduce a new `MemoryStoreInner`, and `MemoryStore` would hold an `StdRwLock<MemoryStoreInner>`.)

crates/matrix-sdk-common/src/linked_chunk/mod.rs
```diff
@@ -945,7 +946,7 @@ impl PartialEq<u64> for ChunkIdentifier {
 ///
 /// It's a pair of a chunk position and an item index.
 #[derive(Copy, Clone, Debug, PartialEq)]
-pub struct Position(ChunkIdentifier, usize);
+pub struct Position(pub(super) ChunkIdentifier, pub(super) usize);
```
bnjbvr, line 949: nit: no need to make thse public, they're already public via `Position::chunk_identifier()` and `Position::index()`.

crates/matrix-sdk-common/src/linked_chunk/mod.rs
```diff
@@ -933,7 +934,7 @@ impl ChunkIdentifierGenerator {
 /// Learn more with [`ChunkIdentifierGenerator`].
 #[derive(Copy, Clone, Debug, PartialEq)]
 #[repr(transparent)]
-pub struct ChunkIdentifier(u64);
+pub struct ChunkIdentifier(pub(super) u64);
```
bnjbvr, line 937: Interesting. Fwiw in the sqlite PR, I've introduced an `index()` method to retrieve this value, since I need it in many places; feel free to introduce that here as a new public method, and keep the inner `u64` private.

crates/matrix-sdk-common/src/linked_chunk/relational.rs
```diff
@@ -0,0 +1,658 @@
...
+#[derive(Debug, PartialEq)]
+struct ItemRow<Item, Gap> {
+    room_id: OwnedRoomId,
+    position: Position,
+    item: Either<Item, Gap>,
+}
+
+/// Kind of item.
+#[derive(Debug, PartialEq)]
+enum Either<Item, Gap> {
```
bnjbvr, line 41: Could we use a better name than `Either`? I thought it was a generic `Either<T, V>` enum, taken from the stdlib or an helper library.

crates/matrix-sdk-common/src/linked_chunk/relational.rs
```diff
@@ -0,0 +1,658 @@
...
+/// would have in a database.
+///
+/// This is used by memory stores. The idea is to have a data layout that is
+/// similar for memory stores and for relational database stores, to represent a
+/// [`LinkedChunk`].
+///
+/// This type is also designed to receive [`Update`]. Applying `Update`s
+/// directly on a [`LinkedChunk`] is not ideal and particularly not trivial as
+/// the `Update`s do _not_ match the internal data layout of the `LinkedChunk`,
+/// they are been designed for storages, like a relational database for example.
```
bnjbvr, line 59:
```suggestion
/// they have been designed for storages, like a relational database for example.
```

crates/matrix-sdk-common/src/linked_chunk/relational.rs
```diff
@@ -0,0 +1,658 @@
...
+///
+/// This type is not as performant as [`LinkedChunk`] (in terms of memory
+/// layout, CPU caches etc.). It is only designed to be used in memory stores,
+/// which are mostly used for test purposes or light usage of the SDK.
+///
+/// [`LinkedChunk`]: super::LinkedChunk
+#[derive(Debug)]
+pub struct RelationalLinkedChunk<Item, Gap> {
+    /// Chunks.
```
bnjbvr, line 68: Here and below, I'm fine if you remove the doc comments, since they don't add any new information :nerd_face:

crates/matrix-sdk-common/src/linked_chunk/relational.rs
```diff
@@ -0,0 +1,658 @@
...
+
+impl<Item, Gap> RelationalLinkedChunk<Item, Gap> {
+    /// Create a new relational linked chunk.
+    pub fn new() -> Self {
+        Self { chunks: Vec::new(), items: Vec::new() }
+    }
+
+    /// Apply [`Update`]s. That's the only way to write data inside this
+    /// relational linked chunk.
+    pub fn apply_updates(&mut self, room_id: &RoomId, updates: &[Update<Item, Gap>])
```
bnjbvr, line 83: Pondering the overall trait's method signature: could it receive a straight `updates: Vec<Update<Item, Gap>>`? This would avoid a few useless clones here and there.

crates/matrix-sdk-common/src/linked_chunk/relational.rs
```diff
@@ -0,0 +1,658 @@
...
+            // Find the previous chunk, and update its next chunk.
+            if let Some(previous) = previous {
+                let entry_for_previous_chunk = chunks
+                    .iter_mut()
+                    .find(|ChunkRow { room_id: room_id_candidate, chunk, .. }| {
+                        &room_id == room_id_candidate && *chunk == previous
+                    })
+                    .expect("Previous chunk should be present");
+
+                // Insert the chunk.
```
bnjbvr, line 257: `Insert` isn't the right action, maybe `Link` instead?

```suggestion
                // Link the chunk.
```

crates/matrix-sdk-common/src/linked_chunk/relational.rs
```diff
@@ -0,0 +1,658 @@
...
+            // Find the next chunk, and update its previous chunk.
+            if let Some(next) = next {
+                let entry_for_next_chunk = chunks
+                    .iter_mut()
+                    .find(|ChunkRow { room_id: room_id_candidate, chunk, .. }| {
+                        &room_id == room_id_candidate && *chunk == next
+                    })
+                    .expect("Next chunk should be present");
+
+                // Insert the chunk.
```
bnjbvr, line 270:
```suggestion
                // Link the chunk.
```

crates/matrix-sdk-common/src/linked_chunk/relational.rs
```diff
@@ -0,0 +1,658 @@
...
+                // 0
+                Update::NewItemsChunk { previous: None, new: CId(0), next: None },
+                // 1 after 0
+                Update::NewGapChunk { previous: Some(CId(0)), new: CId(1), next: None, gap: () },
+                // 2 after 1
+                Update::NewItemsChunk { previous: Some(CId(1)), new: CId(2), next: None },
+            ],
+        );
+
+        // Chunks are correctly links.
```
bnjbvr, line 359:
```suggestion
        // Chunks are correctly linked.
```

crates/matrix-sdk-common/src/linked_chunk/relational.rs
```diff
@@ -0,0 +1,658 @@
...
+                // 1 after 0
+                Update::NewGapChunk { previous: Some(CId(0)), new: CId(1), next: None, gap: () },
+                // 2 after 1
+                Update::NewItemsChunk { previous: Some(CId(1)), new: CId(2), next: None },
+                // remove 1
+                Update::RemoveChunk(CId(1)),
+            ],
+        );
+
+        // Chunks are correctly links.
```
bnjbvr, line 413:
```suggestion
        // Chunks are correctly linked.
```

crates/matrix-sdk-common/src/linked_chunk/relational.rs
```diff
@@ -0,0 +1,658 @@
...
+                // new chunk (to test new items are pushed in the correct chunk)
+                Update::NewItemsChunk { previous: Some(CId(0)), new: CId(1), next: None },
+                // new items on 1
+                Update::PushItems { at: Position(CId(1), 0), items: vec!['x', 'y', 'z'] },
+                // new items on 0 again
+                Update::PushItems { at: Position(CId(0), 3), items: vec!['d', 'e'] },
+            ],
+        );
+
+        // Chunks are correctly links.
```
bnjbvr, line 456:
```suggestion
        // Chunks are correctly linked.
```

crates/matrix-sdk-common/src/linked_chunk/relational.rs
```diff
@@ -0,0 +1,658 @@
...
+                // new items on 0
+                Update::PushItems { at: Position(CId(0), 0), items: vec!['a', 'b', 'c', 'd', 'e'] },
+                // remove an item: 'a'
+                Update::RemoveItem { at: Position(CId(0), 0) },
+                // remove an item: 'd'
+                Update::RemoveItem { at: Position(CId(0), 2) },
+            ],
+        );
+
+        // Chunks are correctly links.
```
bnjbvr, line 541:
```suggestion
        // Chunks are correctly linked.
```

:grin:

crates/matrix-sdk-common/src/linked_chunk/relational.rs
```diff
@@ -0,0 +1,658 @@
...
+                // new items on 0
+                Update::PushItems { at: Position(CId(0), 0), items: vec!['a', 'b', 'c', 'd', 'e'] },
+                // new items on 1
+                Update::PushItems { at: Position(CId(1), 0), items: vec!['x', 'y', 'z'] },
+                // detach last items on 0
+                Update::DetachLastItems { at: Position(CId(0), 2) },
+            ],
+        );
+
+        // Chunks are correctly links.
```
bnjbvr, line 595:
```suggestion
        // Chunks are correctly linked.
```

:zany_face:

bnjbvr (approved):
lgtm! minor comments so approving to avoid back-and-forth, and I've included some ideas for bigger improvements (a single inner lock for the memory store; tweak the signature of `handle_linked_chunk_updates`) that can be done here or as follow-up PRs. Thanks!

crates/matrix-sdk-common/src/linked_chunk/mod.rs
Hywan: It's useful to be public to build them. Gonna check if it's still really required. I wouldn't want to introduce a constructor just for such simple types.

crates/matrix-sdk-common/src/linked_chunk/relational.rs
Hywan: `ItemOrGap` :-D?

crates/matrix-sdk-common/src/linked_chunk/mod.rs
Hywan: Okay, let's use constructor.

crates/matrix-sdk-base/src/event_cache/store/memory_store.rs
Hywan: Done.

crates/matrix-sdk-common/src/linked_chunk/relational.rs
Hywan: I've improved the doc insted.

PR #4299, opened 2024-11-20
feat(base): Add `EventCacheStore::handle_linked_chunk_updates`

Hywan:
This patch adds the `handle_linked_chunk_updates` method on the `EventCacheStore` trait.

---


crates/matrix-sdk-base/src/event_cache/store/traits.rs
```diff
@@ -37,6 +40,11 @@ pub trait EventCacheStore: AsyncTraitDeps {
         holder: &str,
     ) -> Result<bool, Self::Error>;
 
+    async fn handle_linked_chunk_updates(
```
bnjbvr, line 43: Can we add a small doc comment here?

bnjbvr (approved):
yay for parallelization \o/

crates/matrix-sdk-base/src/event_cache/store/traits.rs
Hywan: On it!

PR #4303, opened 2024-11-20
feat(timeline): mark media events as editable

pixlwave:
This PR makes audio, file, image and video messages be editable so that the timeline signals when it is possible to use #4277/#4300 for editing captions.

### DIFF
--- a/crates/matrix-sdk-ui/src/timeline/event_item/mod.rs
+++ b/crates/matrix-sdk-ui/src/timeline/event_item/mod.rs
@@ -368,7 +368,15 @@ impl EventTimelineItem {
 
         match self.content() {
             TimelineItemContent::Message(message) => {
-                matches!(message.msgtype(), MessageType::Text(_) | MessageType::Emote(_))
+                matches!(
+                    message.msgtype(),
+                    MessageType::Text(_)
+                        | MessageType::Emote(_)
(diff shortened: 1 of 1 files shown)


crates/matrix-sdk-ui/src/timeline/event_item/mod.rs
```diff
@@ -368,7 +368,15 @@ impl EventTimelineItem {
...
-                matches!(message.msgtype(), MessageType::Text(_) | MessageType::Emote(_))
+                matches!(
+                    message.msgtype(),
+                    MessageType::Text(_)
+                        | MessageType::Emote(_)
+                        | MessageType::Audio(_)
+                        | MessageType::File(_)
+                        | MessageType::Image(_)
+                        | MessageType::Video(_)
+                )
```
bmarty, line 379: I see a bit below a function `is_highlighted`. I am curious, are you using it on iOS?

bnjbvr (approved):

crates/matrix-sdk-ui/src/timeline/event_item/mod.rs
pixlwave: I can't find anything - I don't think it has been exposed over the FFI even if we wanted to (but presumably we're happy with the pill colour to highlight a message?).

Issue #4270, opened 2024-11-14
Message key "withheld" reasons are not shown in UI nor reported to Posthog


richvdh:
If a sender withholds a room key (eg due to the recipient device not being verified), we should show that in the UI, and report it to Posthog.

See [parent story](https://github.com/element-hq/element-meta/issues/2450) for suggested wording

PR #4305, opened 2024-11-21
feat(crypto): Add optional withheld reason to `UnableToDecryptReason`

BillCarsonFr:
Fixes https://github.com/matrix-org/matrix-rust-sdk/issues/4270

Signed-off-by:

### DIFF
--- a/crates/matrix-sdk-common/src/deserialized_responses.rs
+++ b/crates/matrix-sdk-common/src/deserialized_responses.rs
@@ -17,7 +17,10 @@ use std::{collections::BTreeMap, fmt};
 use ruma::{
     events::{AnyMessageLikeEvent, AnySyncTimelineEvent, AnyTimelineEvent},
     push::Action,
-    serde::{JsonObject, Raw},
+    serde::{
+        AsRefStr, AsStrAsRefStr, DebugAsRefStr, DeserializeFromCowStr, FromString, JsonObject, Raw,
+        SerializeAsRefStr,
+    },
     DeviceKeyAlgorithm, OwnedDeviceId, OwnedEventId, OwnedUserId,
 };
 use serde::{Deserialize, Serialize};
@@ -666,12 +669,30 @@ pub struct UnableToDecryptInfo {
     pub session_id: Option<String>,
 
     /// Reason code for the decryption failure
-    #[serde(default = "unknown_utd_reason")]
+    #[serde(default = "unknown_utd_reason", deserialize_with = "deserialize_utd_reason")]
     pub reason: UnableToDecryptReason,
 }
 
 fn unknown_utd_reason() -> UnableToDecryptReason {
     UnableToDecryptReason::Unknown
+}
+
+/// Provides basic backward compatibility for deserializing older serialized
+/// `UnableToDecryptReason` values.
+pub fn deserialize_utd_reason<'de, D>(d: D) -> Result<UnableToDecryptReason, D::Error>
+where
+    D: serde::Deserializer<'de>,
+{
+    // Start by deserializing as to an untyped JSON value.
+    let v: serde_json::Value = Deserialize::deserialize(d)?;
+    // Backwards compatibility: `MissingMegolmSession` used to be stored without the
+    // withheld code.
+    if v.as_str().is_some_and(|s| s == "MissingMegolmSession") {
+        return Ok(UnableToDecryptReason::MissingMegolmSession { withheld_code: None });
+    }
+    // Otherwise, use the derived deserialize impl to turn the JSON into a
+    // UnableToDecryptReason
+    serde_json::from_value::<UnableToDecryptReason>(v).map_err(serde::de::Error::custom)
 }
 
 /// Reason code for a decryption failure
@@ -689,9 +710,11 @@ pub enum UnableToDecryptReason {
 
     /// Decryption failed because we're missing the megolm session that was used
     /// to encrypt the event.
-    ///
-    /// TODO: support withheld codes?
-    MissingMegolmSession,
+    MissingMegolmSession {
+        /// If the key was withheld on purpose, the associated code. `None`
+        /// means no withheld code was received.
+        withheld_code: Option<WithheldCode>,
+    },
 
     /// Decryption failed because, while we have the megolm session that was
     /// used to encrypt the message, it is ratcheted too far forward.
@@ -723,7 +746,86 @@ impl UnableToDecryptReason {
     /// Returns true if this UTD is due to a missing room key (and hence might
     /// resolve itself if we wait a bit.)
     pub fn is_missing_room_key(&self) -> bool {
-        matches!(self, Self::MissingMegolmSession | Self::UnknownMegolmMessageIndex)
+        // In case of MissingMegolmSession with a withheld code we return false here
+        // given that this API is used to decide if waiting a bit will help.
+        matches!(
+            self,
+            Self::MissingMegolmSession { withheld_code: None } | Self::UnknownMegolmMessageIndex
+        )
+    }
+}
+
+/// A machine-readable code for why a Megolm key was not sent.
+///
+/// Normally sent as the payload of an [`m.room_key.withheld`](https://spec.matrix.org/v1.12/client-server-api/#mroom_keywithheld) to-device message.
+#[derive(
+    Clone,
+    PartialEq,
+    Eq,
+    Hash,
+    AsStrAsRefStr,
+    AsRefStr,
+    FromString,
+    DebugAsRefStr,
+    SerializeAsRefStr,
+    DeserializeFromCowStr,
+)]
+pub enum WithheldCode {
+    /// the user/device was blacklisted.
+    #[ruma_enum(rename = "m.blacklisted")]
+    Blacklisted,
+
+    /// the user/devices is unverified.
+    #[ruma_enum(rename = "m.unverified")]
+    Unverified,
+
+    /// The user/device is not allowed have the key. For example, this would
+    /// usually be sent in response to a key request if the user was not in
+    /// the room when the message was sent.
+    #[ruma_enum(rename = "m.unauthorised")]
+    Unauthorised,
+
+    /// Sent in reply to a key request if the device that the key is requested
+    /// from does not have the requested key.
+    #[ruma_enum(rename = "m.unavailable")]
+    Unavailable,
+
+    /// An olm session could not be established.
+    /// This may happen, for example, if the sender was unable to obtain a
+    /// one-time key from the recipient.
+    #[ruma_enum(rename = "m.no_olm")]
+    NoOlm,
+
+    #[doc(hidden)]
+    _Custom(PrivOwnedStr),
+}
+
+impl fmt::Display for WithheldCode {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> Result<(), fmt::Error> {
+        let string = match self {
+            WithheldCode::Blacklisted => "The sender has blocked you.",
+            WithheldCode::Unverified => "The sender has disabled encrypting to unverified devices.",
+            WithheldCode::Unauthorised => "You are not authorised to read the message.",
+            WithheldCode::Unavailable => "The requested key was not found.",
+            WithheldCode::NoOlm => "Unable to establish a secure channel.",
+            _ => self.as_str(),
+        };
+
+        f.write_str(string)
+    }
+}
+
+// The Ruma macro expects the type to have this name.
+// The payload is counter intuitively made public in order to avoid having
+// multiple copies of this struct.
+#[doc(hidden)]
+#[derive(Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub struct PrivOwnedStr(pub Box<str>);
+
+#[cfg(not(tarpaulin_include))]
+impl fmt::Debug for PrivOwnedStr {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        self.0.fmt(f)
     }
 }
 
@@ -817,7 +919,7 @@ mod tests {
     use super::{
         AlgorithmInfo, DecryptedRoomEvent, EncryptionInfo, SyncTimelineEvent, TimelineEvent,
         TimelineEventKind, UnableToDecryptInfo, UnableToDecryptReason, UnsignedDecryptionResult,
-        UnsignedEventLocation, VerificationState,
+        UnsignedEventLocation, VerificationState, WithheldCode,
     };
     use crate::deserialized_responses::{DeviceLinkProblem, VerificationLevel};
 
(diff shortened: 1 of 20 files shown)


crates/matrix-sdk-common/src/deserialized_responses.rs
```diff
@@ -718,7 +733,12 @@ impl UnableToDecryptReason {
     /// Returns true if this UTD is due to a missing room key (and hence might
     /// resolve itself if we wait a bit.)
     pub fn is_missing_room_key(&self) -> bool {
-        matches!(self, Self::MissingMegolmSession | Self::UnknownMegolmMessageIndex)
+        match self {
```
BillCarsonFr, line 736: The doc says here that `true` means that  `( [it] might resolve itself if we wait a bit.)`
If we have a withheld code waiting won't help. It is used by the notification client to decide if it should work for an additional encryption sync.

crates/matrix-sdk-common/src/deserialized_responses.rs
```diff
@@ -669,6 +669,23 @@ fn unknown_utd_reason() -> UnableToDecryptReason {
...
+    /// When the sender refuses to send the key because his security settings in
+    /// that room are not met by your device. Could be that your device is
+    /// not cross-signed (insecure device) or that you are not verified by
+    /// the sender.
+    TrustRequirementMismatch,
+    /// Other reasons. Like fail to establish a secure channel to share the key,
+    /// or if for some reason our device is blocked.
+    Other,
+}
 /// Reason code for a decryption failure
```
richvdh, line 689: missing blank line here

crates/matrix-sdk-common/src/deserialized_responses.rs
```diff
@@ -669,6 +669,23 @@ fn unknown_utd_reason() -> UnableToDecryptReason {
     UnableToDecryptReason::Unknown
 }
 
+/// A room key might be missing because the sender refused to share it or was
+/// not technically able to share it.
+/// In the protocol this is reflected by clients sending `m.room_key.withheld`
+/// to the participant that won't receive the key.
+/// We only want here to consider the subset of codes that clients should
+/// use to display specific UX.
```
richvdh, line 677: as [before](https://github.com/matrix-org/matrix-rust-sdk/pull/4275#discussion_r1860871953), the formatting here is... odd.

If you want a paragraph break, you need a blank line. If you *don't* want a paragraph break, then don't pretend that there is one.

crates/matrix-sdk-common/src/deserialized_responses.rs
```diff
@@ -684,9 +701,7 @@ pub enum UnableToDecryptReason {
 
     /// Decryption failed because we're missing the megolm session that was used
     /// to encrypt the event.
-    ///
-    /// TODO: support withheld codes?
-    MissingMegolmSession,
+    MissingMegolmSession(Option<WithheldReason>),
```
richvdh, line 704: Please document what Some/None mean

crates/matrix-sdk-common/src/deserialized_responses.rs
```diff
@@ -718,7 +733,12 @@ impl UnableToDecryptReason {
     /// Returns true if this UTD is due to a missing room key (and hence might
     /// resolve itself if we wait a bit.)
     pub fn is_missing_room_key(&self) -> bool {
-        matches!(self, Self::MissingMegolmSession | Self::UnknownMegolmMessageIndex)
+        match self {
+            // Return true only if the key is not withheld on purpose.
+            Self::MissingMegolmSession(None) => true,
+            Self::UnknownMegolmMessageIndex => true,
+            _ => false,
+        }
```
richvdh, line 741: Surprised that the linter doesn't complain about this, but this should still be a `matches!` macro I think

crates/matrix-sdk-common/src/deserialized_responses.rs
richvdh, line 704: Could we instead move the existing `WithheldCode` into matrix-sdk-common, rather than inventing a new type that does basically the same thing?

richvdh, line 704: I think this change is going to break deserialization of existing  TimelineEvents? Needs some serde magic. Or a test that proves it's ok

crates/matrix-sdk-common/src/deserialized_responses.rs
```diff
@@ -669,6 +669,23 @@ fn unknown_utd_reason() -> UnableToDecryptReason {
...
+pub enum WithheldReason {
+    /// When the sender refuses to send the key because his security settings in
+    /// that room are not met by your device. Could be that your device is
+    /// not cross-signed (insecure device) or that you are not verified by
+    /// the sender.
+    TrustRequirementMismatch,
+    /// Other reasons. Like fail to establish a secure channel to share the key,
+    /// or if for some reason our device is blocked.
+    Other,
+}
```
richvdh, line 688: Could we instead move the existing `WithheldCode` into matrix-sdk-common, rather than inventing a new type that does basically the same thing?

crates/matrix-sdk-crypto/src/types/events/utd_cause.rs
```diff
@@ -57,6 +57,18 @@ pub enum UtdCause {
     /// be confused with pre-join or pre-invite messages (see
     /// [`UtdCause::SentBeforeWeJoined`] for that).
     HistoricalMessage = 5,
+
+    /// The keys for this event are intentionally withheld.
+    /// The sender has refused to share the key because our device does not meet
+    /// the sender's security requirements.
```
richvdh, line 63: Again, weird paragraph break formatting.

crates/matrix-sdk-crypto/src/types/events/utd_cause.rs
```diff
@@ -98,7 +110,13 @@ impl UtdCause {
     ) -> Self {
         // TODO: in future, use more information to give a richer answer. E.g.
```
richvdh, line 111: we should maybe get rid of this line at some point

crates/matrix-sdk-crypto/src/types/events/utd_cause.rs
```diff
@@ -57,6 +57,18 @@ pub enum UtdCause {
...
+    /// The keys for this event are intentionally withheld.
+    /// The sender has refused to share the key because our device does not meet
+    /// the sender's security requirements.
+    WithheldForUnverifiedOrInsecureDevice = 6,
+
+    /// The keys for this event are missing, likely because the sender was
+    /// unable to share them (e.g., failure to establish an Olm 1:1
+    /// channel). Alternatively, the sender may have deliberately excluded
+    /// this device by cherry-picking and blocking it, no action can be taken on
+    /// our side.
```
richvdh, line 70: > no action can be taken on our side.

Is this actually correct? Eg, if they failed to establish an olm session, we could set one up from our side.

richvdh (changes requested):
Looks like a sensible direction to me; a few comments though

Mostly I'd like to know if we can avoid introducing yet another type

crates/matrix-sdk-common/src/deserialized_responses.rs
BillCarsonFr: Can we have a formatting rule that forbid new lines in comments then? if it makes no sense

crates/matrix-sdk-common/src/deserialized_responses.rs
BillCarsonFr: I though we said it was fine for now, looking at that

crates/matrix-sdk-common/src/deserialized_responses.rs
BillCarsonFr: https://github.com/matrix-org/matrix-rust-sdk/pull/4305/commits/e3b6c5007a8480be414c58717cff6aadd2c042e3
and
https://github.com/matrix-org/matrix-rust-sdk/pull/4305/commits/12bb41daf819be85360610e529a4ec5c5385adc4

crates/matrix-sdk-crypto/src/types/events/utd_cause.rs
BillCarsonFr: I am just talking about the blocking scenario

crates/matrix-sdk-common/src/deserialized_responses.rs
richvdh: new lines are perfectly valid, they just don't mean what you seem to think they mean.

poljar: A related point here, if you put the whole comment on the same line `rustfmt` will happily format this for you correctly, unless of course you do wish paragraph breaks, but then you can put individual paragraphs on a single line and `rustfmt` will help you with each individual paragraph.

crates/matrix-sdk-common/src/deserialized_responses.rs
BillCarsonFr: Not sure to get it, what do you suggest?

crates/matrix-sdk-common/src/deserialized_responses.rs
BillCarsonFr: Here https://github.com/matrix-org/matrix-rust-sdk/pull/4305/commits/15297fb3b7646681fa3b3c6033a5ef5a6ad390be

crates/matrix-sdk-common/src/deserialized_responses.rs
richvdh:
```suggestion
        // Return true only if the key is not withheld on purpose.
        matches!(self, Self::MissingMegolmSession(None) | Self::UnknownMegolmMessageIndex)
```

crates/matrix-sdk-common/src/deserialized_responses.rs
BillCarsonFr: > new lines are perfectly valid, they just don't mean what you seem to think they mean.

I put them for clarity when reading the source code because of the maximum length, there is no such constraint in the doc

poljar: Not exactly sure which constraint you mean, but in case you mean a max comment length, we have this defined: https://github.com/matrix-org/matrix-rust-sdk/blob/e1f0037fd5eb3b263a471a0c515daf6bf95cd784/.rustfmt.toml#L2

So no need to think about the max comment length yourself, `rustfmt` will format comments for you.

crates/matrix-sdk-common/src/deserialized_responses.rs
```diff
@@ -666,15 +669,26 @@ pub struct UnableToDecryptInfo {
...
-    #[serde(default = "unknown_utd_reason")]
+    #[serde(default = "unknown_utd_reason", deserialize_with = "ok_reason_or_default")]
     pub reason: UnableToDecryptReason,
 }
 
 fn unknown_utd_reason() -> UnableToDecryptReason {
     UnableToDecryptReason::Unknown
 }
 
-/// Reason code for a decryption failure
```
richvdh, line 677: I think removing the documentation on `UnableToDecryptReason` is probably an accident?

crates/matrix-sdk-common/src/deserialized_responses.rs
```diff
@@ -666,15 +669,26 @@ pub struct UnableToDecryptInfo {
...
+/// `UnableToDecryptReason` values. If the serialized format has changed in a
+/// way that causes deserialization to fail, the reason will default to
+/// `Unknown`.
+fn ok_reason_or_default<'de, D>(d: D) -> Result<UnableToDecryptReason, D::Error>
+where
+    D: serde::Deserializer<'de>,
+{
+    let v: serde_json::Value = Deserialize::deserialize(d)?;
+    Ok(UnableToDecryptReason::deserialize(v).unwrap_or(UnableToDecryptReason::Unknown))
+}
```
richvdh, line 690: This feels like the wrong approach for three reasons:

 1. It only applies to `UnableToDecryptReason`s that are embedded in `UnableToDecryptInfo`, so this still breaks backwards compatibility for `UnableToDecryptReason`s that have been serialized in any other environment.
 2. It turns existing `MissingMegolmSession`s into `UnableToDecryptReason::Unknown`, which seems like a disappointing loss of data.
 3. It's a bit of a sledgehammer: turning every single decryption error into a `UnableToDecryptReason::Unknown` seems like another loss of data.

I think points 2 and 3 can be fixed by manipulating the json `Value` before calling `UnableToDecryptReason::deserialize(v)`, but I don't know how to fix point 1 without either (a) creating and maintaining another copy of `UnableToDecryptReason` (so that one version can have a custom deserializer that fiddles with the serialized format and then uses the default deserializer of the second version) or (b) hand-writing an entire deserializer. Will ask in the rust room.

richvdh: Having discussed this internally: it seems unlikely that anyone will be serializing `UnableToDecryptReason` outside `UnableToDecryptInfo`, and in any case we aren't yet doing stable releases, so we can disregard point 1.

I'd still quite like us to address points 2 & 3. Something like:


```suggestion
pub fn deserialize_utd_reason<'de, D>(d: D) -> Result<UnableToDecryptReason, D::Error>
where
    D: serde::Deserializer<'de>,
{
    // Start by deserializing as to an untyped JSON value.
    let v: serde_json::Value = Deserialize::deserialize(d)?;

    // Backwards compatibility: `MissingMegolmSession` used to be stored without the withheld code.
    if v.as_str().is_some_and(|s| s == "MissingMegolmSession") {
        return Ok(UnableToDecryptReason::MissingMegolmSession(None));
    }

    // Otherwise, use the derived deserialize impl to turn the JSON into a UnableToDecryptReason
    serde_json::from_value::<UnableToDecryptReason>(v).map_err(serde::de::Error::custom)
}
```

(note also, improved name and the fact that it should probably be `pub`.)

crates/matrix-sdk-common/src/deserialized_responses.rs
```diff
@@ -688,10 +702,9 @@ pub enum UnableToDecryptReason {
     MalformedEncryptedEvent,
 
     /// Decryption failed because we're missing the megolm session that was used
-    /// to encrypt the event.
-    ///
-    /// TODO: support withheld codes?
-    MissingMegolmSession,
+    /// to encrypt the event. If the key was withheld on purpose the
+    /// associated code is attached, None means no withheld code was received.
+    MissingMegolmSession(Option<WithheldCode>),
```
richvdh, line 707: I'd be somewhat in favour of making this a struct variant rather than a tuple variant: I think it would be clearer, and allow for easier extension in the future:

```suggestion
    MissingMegolmSession { withheld_code: Option<WithheldCode> },
```

However, it's not a strong opinion.

richvdh (changes requested):

crates/matrix-sdk-common/src/deserialized_responses.rs
```diff
@@ -723,7 +736,83 @@ impl UnableToDecryptReason {
...
+    }
+}
+
+// Wrapper around `Box<str>` that cannot be used in a meaningful way outside of
+// this crate. Used for string enums because their `_Custom` variant can't be
+// truly private (only `#[doc(hidden)]`).
+#[doc(hidden)]
+#[derive(Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+// XXX had to make the field public to make ruma `StringEnum` macro happy.
+pub struct PrivOwnedStr(pub Box<str>);
```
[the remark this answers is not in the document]
BillCarsonFr: Yes nothing makes sense with the struct anymore (the name, the doc, ..)
Not sure how to properly document it though.

richvdh (commented):
LGTM now, with the exception of PrivOwnedStr, for which I defer to @poljar

crates/matrix-sdk-common/src/deserialized_responses.rs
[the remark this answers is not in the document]
poljar: The Ruma macro expects the type to have this name, it's not something that is in under our control...

The alternatives are to not use the Ruma macro or to have multiple copies of this struct.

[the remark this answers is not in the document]
richvdh: ok, fair enough. It would be helpful if this could be explained in a comment.

Why *does* this need to be `pub`, when it didn't need to be `pub` when it lived in the crypto crate?

crates/matrix-sdk-crypto/src/types/events/utd_cause.rs
```diff
@@ -97,8 +110,14 @@ impl UtdCause {
         unable_to_decrypt_info: &UnableToDecryptInfo,
     ) -> Self {
         // TODO: in future, use more information to give a richer answer. E.g.
-        match unable_to_decrypt_info.reason {
-            UnableToDecryptReason::MissingMegolmSession
+        match &unable_to_decrypt_info.reason {
+            UnableToDecryptReason::MissingMegolmSession { withheld_code: Some(reason) } => {
+                match reason {
+                    WithheldCode::Unverified => UtdCause::WithheldForUnverifiedOrInsecureDevice,
+                    _ => UtdCause::WithheldBySender,
```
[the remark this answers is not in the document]
BillCarsonFr: Done https://github.com/matrix-org/matrix-rust-sdk/pull/4305/commits/0eee8457f38eb3bbba6e81b155d7f8d5d69810dd

crates/matrix-sdk-common/src/deserialized_responses.rs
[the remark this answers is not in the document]
BillCarsonFr: https://github.com/matrix-org/matrix-rust-sdk/pull/4305/commits/1f50e4b1a04b9b3f2f408a9a5b843459e97d711a

> when it didn't need to be pub when it lived in the crypto crate

Because it is also used by other enums (still in crypto) that now uses the one that is in common now.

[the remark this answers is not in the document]
BillCarsonFr: Actually there was another copy left https://github.com/matrix-org/matrix-rust-sdk/pull/4305/commits/2151fcdbe84c09c051a9c42ae9761620c63e1822

I would be ok to revert that and have multiple copies if it makes it simpler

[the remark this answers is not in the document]
poljar: > Why does this need to be pub, when it didn't need to be pub when it lived in the crypto crate?

Because otherwise you can't construct the variant outside of the crate where the type lives unless it's pub. Now I don't exactly know where that is needed but if we remove the pub the compiler might tell us.

poljar (approved):

PR #4306, opened 2024-11-21
refactor: Clean up the Room::compute_display_name() method

poljar:
This method has grown organically into a bit of a beast. Let's try to make it a bit easier to understand in preparation for [MSC4171](https://github.com/matrix-org/matrix-spec-proposals/pull/4171) :saluting_face:.

### DIFF
--- a/crates/matrix-sdk-base/src/rooms/normal.rs
+++ b/crates/matrix-sdk-base/src/rooms/normal.rs
@@ -573,49 +573,113 @@ impl Room {
     ///
     /// [spec]: <https://matrix.org/docs/spec/client_server/latest#calculating-the-display-name-for-a-room>
     pub async fn compute_display_name(&self) -> StoreResult<RoomDisplayName> {
-        let update_cache = |new_val: RoomDisplayName| {
-            self.inner.update_if(|info| {
-                if info.cached_display_name.as_ref() != Some(&new_val) {
-                    info.cached_display_name = Some(new_val.clone());
-                    true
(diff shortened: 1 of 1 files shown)


crates/matrix-sdk-base/src/rooms/normal.rs
```diff
@@ -573,63 +573,61 @@ impl Room {
...
+                    DisplayNameOrSummary::DisplayName(name)
+                }
+                (None, Some(alias)) => {
+                    let name = RoomDisplayName::Aliased(alias.alias().trim().to_owned());
+                    DisplayNameOrSummary::DisplayName(name)
+                }
+                // We can't directly compute the display name from the summary here because Rust
+                // thinks that the `inner` lock is still held even if we explicitly call `drop()`
+                // on it. So we introduced the DisplayNameOrSummary type and do the computation in
+                // two steps.
```
jmartinesp, line 596: Weird, but OK. Is this some kind of bug in the compiler?

jmartinesp (approved):
LGTM, thanks. Just a question about the weird `drop` issue you mentioned.

crates/matrix-sdk-base/src/rooms/normal.rs
poljar: :shrug: I have not investigated further, we had the same situation previously as well, but with some early returns.

I have somewhere seen that it's about thread local storage the lock uses, but no idea realistically.

PR #4309, opened 2024-11-22
feat(ffi): Add `RoomInfo::join_rule` field to bindings

jmartinesp:
This was requested by the clients, to be notified when the join rule of a room changes.

Signed-off-by:


bindings/matrix-sdk-ffi/src/client.rs
```diff
@@ -1969,3 +1972,45 @@ impl TryFrom<AllowRule> for ruma::events::room::join_rules::AllowRule {
...
+                let rules = restricted.allow.into_iter().map(TryInto::try_into).collect::<Result<
+                    Vec<_>,
+                    Self::Error,
+                >>(
+                )?;
+                Ok(JoinRule::Restricted { rules })
+            }
+            RumaJoinRule::Invite => Ok(JoinRule::Invite),
+            RumaJoinRule::_Custom(_) => Ok(JoinRule::Custom { repr: value.as_str().to_owned() }),
+            _ => Err(format!("Invalid JoinRule: {:?}", value)),
```
poljar, line 2001: This won't be an invalid one, it'll be a new variant that was added to `Ruma` and we don't handle it yet.

bindings/matrix-sdk-ffi/src/client.rs
```diff
@@ -1969,3 +1972,45 @@ impl TryFrom<AllowRule> for ruma::events::room::join_rules::AllowRule {
...
+}
+
+impl TryFrom<RumaAllowRule> for AllowRule {
+    type Error = String;
+    fn try_from(value: RumaAllowRule) -> Result<Self, Self::Error> {
+        match value {
+            RumaAllowRule::RoomMembership(membership) => {
+                Ok(AllowRule::RoomMembership { room_id: membership.room_id.to_string() })
+            }
+            _ => Err(format!("Invalid AllowRule: {:?}", value)),
```
poljar, line 2013: Same.

poljar (approved):

bindings/matrix-sdk-ffi/src/client.rs
jmartinesp: Ok, I'll change it to 'unknown' instead.

bindings/matrix-sdk-ffi/src/client.rs
jmartinesp: I'm wondering: should I serialize the custom rule as a JSON and allow the clients to deserialize it instead? It's a possible solution, but maybe we'd rather just go with the 'unkown join rule' error.

bindings/matrix-sdk-ffi/src/room_info.rs
```diff
@@ -54,8 +56,10 @@ pub struct RoomInfo {
     /// Events causing mentions/highlights for the user, according to their
     /// notification settings.
     num_unread_mentions: u64,
-    /// The currently pinned event ids
+    /// The currently pinned event ids.
     pinned_event_ids: Vec<String>,
+    /// The join rule for this room, if known.
+    join_rule: Option<JoinRule>,
```
ganfra, line 62: should we remove the `is_public` as it's checking the `JoinRule`?

jmartinesp: I don't see a `is_public` field in `RoomPreviewInfo`.

jmartinesp: Yep, good call.

bindings/matrix-sdk-ffi/src/client.rs
poljar: Well the official way of handling this in Rust would be to use `as_str()` which can cover the `Custom` as well as the `_` cases.

jmartinesp: I'm not sure we're talking about the same thing: `_Custom` has a boxed `CustomAllowRule`, which is a struct that has a `rule_type` and `extras: BTreeMap<String, JsonValue>`. Its json representation is actually something like:

```json
{
    "type": "some-custom-rule", // becomes `rule_type` 
    "some-random-extra-value: { ... }, // everything else is under extras
    "another-value": 1
}
```

AFAICT we can't just pass a `Map<String, Anything>` to Swift/Kotlin (not now or probably ever), so I was proposing to pass the stringified JSON representation instead. Whether it's through a `serde_json::to_string` or hiding that behind as an `as_str()` function. I'm not sure it's a good idea, but I see no other way of being able to handle the custom case in the clients.

poljar: Oh, those enums are even funkier then usually, usually they only have a `PrivOwnedString`.

Yeah, you're thinking is right, you need to re-serialize the rule again. Seems that this type is one of the few types in Ruma that support reserialization.

I wonder why the thing is again hidden behind a `_Custom` and what the correct™ way to use custom variants even on Rust would be.

jmartinesp: Ok, I'll implement this instead of the error instead then.

jmartinesp:
Asked for another review in case I messed up the 'reserialization' logic, although it seems to work fine AFAICT.

poljar (approved):
Still seems good, thanks.

PR #4310, opened 2024-11-22
refactor(room_preview): make `RoomPreview` use the local known data only for joined rooms

jmartinesp:
When instantiating a room preview, previously the SDK would try to just check if the room exists locally either as joined, invited, knocked, left, etc., and then retrieve the info we cached about it.

While this seems fine for most cases, it turns out for non-joined rooms, the info we have locally will **always** be the one we received when the invite/knock/leave action took place and it'll never be updated, so we may have the case where we knock into a room, never receive a response, someone changes the join rule of the room to something else and we'll think about this room as a 'request to join' room until we clear the local cache. There are other cases such as the room name, topic or other 'public' values changing in the meantime and the user thinking he's trying to join a completely different room too, but these seem like minor issues compared to never being able to join at all.

To prevent all that, I think we can only use the local data for joined rooms, which are constantly updated, and try to use the room summary API and other fallbacks for the rest, even if they're rooms known to us.

Signed-off-by:

### DIFF
--- a/crates/matrix-sdk/src/client/mod.rs
+++ b/crates/matrix-sdk/src/client/mod.rs
@@ -1142,8 +1142,8 @@ impl Client {
         self.base_client().get_room(room_id).map(|room| Room::new(self.clone(), room))
     }
 
-    /// Gets the preview of a room, whether the current user knows it (because
-    /// they've joined/left/been invited to it) or not.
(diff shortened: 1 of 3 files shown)


poljar (approved):
Seems sensible.

PR #4311, opened 2024-11-22
build (apple): add dynamic type to debug package

Velin92:
On the latest Xcode versions this is required to make the app run when building the SDK locally


stefanceriu (approved):
Follow up to https://github.com/element-hq/element-x-ios/pull/3539

PR #4312, opened 2024-11-22
chore: Add missing changelog entries for the ui crate

poljar:


Hywan (approved):

poljar:
@bnjbvr mentions that there are a bunch of entries that are not related to the UI crate.

I think we might need to rethink how we do changelogs a bit more, i.e. perhaps switch to a single changelog file with separate categories per crate instead.

PR #4308, opened 2024-11-21
feat(event cache): enable persistent storage by default :sunglasses: 

bnjbvr:
After weeks and weeks of using and testing the persistent storage for the event cache by default in EX apps, and a few rounds of feedback from the community, we are coming to the conclusion that it is ready to be enabled by default. Doing so will enable a few simplifications in the code base we've wanted to do for a long time, and will reduce the number of possible configurations for embedders, simplifying support and maintenance.

A final touch for #3280.

https://github.com/user-attachments/assets/50983dc7-0d53-4363-b8ee-9b68c363995a

PR #4316, opened 2024-11-25
refactor(event cache): some preparatory refactorings

bnjbvr:
This contains a few commits that are useful in isolation from the large event cache storage PR (#4308).


andybalaam (approved):
Looks good, thanks!

PR #4317, opened 2024-11-25
feat(common): Implement `LinkedChunk::clear`

Hywan:
This patch implements `LinkedChunk::clear`. The code from `impl Drop for LinkedChunk` has been moved inside `Ends::clear`, and replaced by a simple `self.links.clear()`. In addition, `LinkedChunk::clear` indeed calls `self.links.clear()` but also resets all fields.

This patch adds the `Clear` variant to `Update`.

This patch updates `AsVector` to emit a `VectorDiff::Clear` on `Update::Clear`.

Finally, this patch adds the necessary tests.

---

### DIFF
--- a/crates/matrix-sdk-common/src/linked_chunk/mod.rs
+++ b/crates/matrix-sdk-common/src/linked_chunk/mod.rs
@@ -190,6 +190,36 @@ impl<const CAP: usize, Item, Gap> Ends<CAP, Item, Gap> {
             chunk = chunk.previous_mut()?;
         }
     }
+
+    /// Drop all chunks, and re-create the first one.
+    fn clear(&mut self) {
+        // Loop over all chunks, from the last to the first chunk, and drop them.
+        {
+            // Take the latest chunk.
+            let mut current_chunk_ptr = self.last.or(Some(self.first));
+
+            // As long as we have another chunk…
+            while let Some(chunk_ptr) = current_chunk_ptr {
+                // Fetch the previous chunk pointer.
+                let previous_ptr = unsafe { chunk_ptr.as_ref() }.previous;
+
+                // Re-box the chunk, and let Rust does its job.
+                let _chunk_boxed = unsafe { Box::from_raw(chunk_ptr.as_ptr()) };
+
+                // Update the `current_chunk_ptr`.
+                current_chunk_ptr = previous_ptr;
+            }
+
+            // At this step, all chunks have been dropped, including
+            // `self.first`.
+        }
+
+        // Recreate the first chunk.
+        self.first = Chunk::new_items_leaked(ChunkIdentifierGenerator::FIRST_IDENTIFIER);
+
+        // Reset the last chunk.
+        self.last = None;
+    }
 }
 
 /// The [`LinkedChunk`] structure.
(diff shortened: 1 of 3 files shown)


crates/matrix-sdk-common/src/linked_chunk/mod.rs
```diff
@@ -190,6 +190,40 @@ impl<const CAP: usize, Item, Gap> Ends<CAP, Item, Gap> {
...
+            let mut current_chunk_ptr = self.last.or(Some(self.first));
+
+            // As long as we have another chunk…
+            while let Some(chunk_ptr) = current_chunk_ptr {
+                // Disconnect the chunk by updating `previous_chunk.next` pointer.
+                let previous_ptr = unsafe { chunk_ptr.as_ref() }.previous;
+
+                if let Some(mut previous_ptr) = previous_ptr {
+                    unsafe { previous_ptr.as_mut() }.next = None;
+                }
```
bnjbvr, line 208: Do we need to do that unlinking? I fail to see if it achieves anything, considering we're about to drop the chunk on the next line…

crates/matrix-sdk-common/src/linked_chunk/mod.rs
```diff
@@ -846,27 +897,11 @@ impl<const CAP: usize, Item, Gap> LinkedChunk<CAP, Item, Gap> {
...
-
-            // Update the `current_chunk_ptr`.
-            current_chunk_ptr = previous_ptr;
-        }
-
-        // At this step, all chunks have been dropped, including
-        // `self.first`.
+        // Only clear the links. Calling `Self::clear` would be an error as we don't
+        // want to emit an `Update::Clear` when `self` is dropped. Instead, we only care
+        // about freeing memory correctly. Rust can take care of everything except the
```
bnjbvr, line 902:
```suggestion
        // about freeing memory correctly. Rust can take care of everything but the
```

(or `except for the`, I think)

bnjbvr: Dismiss this comment: apparently all three forms are fine in this context! (thanks @andybalaam for the spelling check :))

crates/matrix-sdk-common/src/linked_chunk/mod.rs
Hywan: Indeed 🤔, it seems useless. After further investigation, I don't see any proof it is useful today. Let's remove that.

crates/matrix-sdk-common/src/linked_chunk/mod.rs
```diff
@@ -258,6 +292,23 @@ impl<const CAP: usize, Item, Gap> LinkedChunk<CAP, Item, Gap> {
...
+
+        // Clear `self.length`.
+        self.length = 0;
+
+        // Clear `self.chunk_identifier_generator`.
+        self.chunk_identifier_generator = ChunkIdentifierGenerator::new_from_scratch();
+
+        // “Clear” `self.updates`.
+        if let Some(updates) = self.updates.as_mut() {
+            updates.push(Update::Clear);
```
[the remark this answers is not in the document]
Hywan: `updates` is an `ObservableUpdates` here, which doesn't have a `clear()` method. I understand the optimisation but it has some impacts. If an update reader wants to check _all_ updates (even for debug purposes), we may not want this optimisation. Further more, I'm not sure the situation when we have a lot of pending updates will happen that often. Let's keep that as a todo maybe.

Issue #4314, opened 2024-11-25
linked chunk: the initial chunk isn't forwarded as part of the updates


bnjbvr:
When creating a new linked chunk, there's an initial, empty items chunks that's created. Creation of this item isn't observable, as subscribing to the updates will lead to an empty chunk.

cc @Hywan

PR #4321, opened 2024-11-25
refactor: `RoomEvents::reset` really clears the linked chunk

Hywan:
This PR is twofold:

1. It fixes a bug in `AsVector`: when an `Update::Clear` was received, the internal state of `AsVector` wasn't cleared,
2. It updates `RoomEvents::reset` to use `LinkedChunk::clear` so that events are correctly cleared and not dropped silently, i.e. the `Update`s and `VectorDiff`s are correctly reflecting that. That's important to clear the stores (implementations of `EventCacheStore`, like `MemoryStore` or `SqliteStore`). A new `RoomEvents::updates_as_vector_diffs()` method is added, it's going to be used later and it's useful to test the new behaviour now.

---

* Partly address https://github.com/matrix-org/matrix-rust-sdk/issues/4314

### DIFF
--- a/crates/matrix-sdk-common/src/linked_chunk/as_vector.rs
+++ b/crates/matrix-sdk-common/src/linked_chunk/as_vector.rs
@@ -729,9 +729,9 @@ mod tests {
             assert_eq!(chunks.len(), 1);
             assert_eq!(chunks[0].0, ChunkIdentifierGenerator::FIRST_IDENTIFIER);
             assert_eq!(chunks[0].1, 0);
-        }
 
-        assert!(as_vector.take().is_empty());
+            assert!(as_vector.take().is_empty());
+        }
 
         linked_chunk.push_items_back(['a', 'b', 'c', 'd']);
 
--- a/crates/matrix-sdk-common/src/linked_chunk/as_vector.rs
+++ b/crates/matrix-sdk-common/src/linked_chunk/as_vector.rs
@@ -302,6 +302,12 @@ impl UpdateToVectorDiff {
                             self.chunks.insert(next_chunk_index, (*new, 0));
                         }
 
+                        // First chunk!
+                        (None, None) if self.chunks.is_empty() => {
+                            self.chunks.push_back((*new, 0));
+                        }
+
+                        // Impossible state.
                         (None, None) => {
                             unreachable!(
                                 "Inserting new chunk with no previous nor next chunk identifiers \
@@ -407,6 +413,9 @@ impl UpdateToVectorDiff {
                 }
 
                 Update::Clear => {
+                    // Clean `self.chunks`.
+                    self.chunks.clear();
+
                     // Let's straightforwardly emit a `VectorDiff::Clear`.
                     diffs.push(VectorDiff::Clear);
                 }
@@ -455,10 +464,11 @@ impl UpdateToVectorDiff {
 mod tests {
     use std::fmt::Debug;
 
+    use assert_matches::assert_matches;
     use imbl::{vector, Vector};
 
     use super::{
-        super::{EmptyChunk, LinkedChunk},
+        super::{ChunkIdentifierGenerator, EmptyChunk, LinkedChunk},
         VectorDiff,
     };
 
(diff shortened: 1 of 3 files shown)


crates/matrix-sdk-common/src/linked_chunk/as_vector.rs
```diff
@@ -407,6 +407,14 @@ impl UpdateToVectorDiff {
                 }
 
                 Update::Clear => {
+                    // Clean `self.chunks`.
+                    self.chunks.clear();
+
+                    // An `Update::Clear` happens when `LinkedChunk::clear` is called, which removes
+                    // all chunks, and re-adds one `Chunk` with `FIRST_IDENTIFIER` as its chunk
+                    // identifier.
+                    self.chunks.push_back((ChunkIdentifierGenerator::FIRST_IDENTIFIER, 0));
```
bnjbvr, line 416: I think it's the wrong layer to do that, the linked chunk should save that update and propagate it instead; see also https://github.com/matrix-org/matrix-rust-sdk/issues/4314 which describes the same issue at construction.

crates/matrix-sdk/src/event_cache/room/events.rs
```diff
@@ -158,6 +175,18 @@ impl RoomEvents {
...
+    /// Get all updates from the room events as [`VectorDiff`].
+    ///
+    /// Be careful that each `VectorDiff` is returned only once!
+    ///
+    /// See [`AsVector`] to learn more.
+    ///
+    /// [`Update`]: matrix_sdk_base::linked_chunk::Update
+    #[allow(unused)] // gonna be useful very soon! but we need it now for test purposes
+    pub fn updates_as_vector_diffs(&mut self) -> Vec<VectorDiff<Event>> {
+        self.chunks_updates_as_vectordiffs.take()
```
bnjbvr, line 187: Why do we have to store a `chunks_updates_as_vectordiffs` field, and why is not sufficient to expose the underlying method? Now that the underlying linked chunk is properly reset and not re-created, the stream won't close after a reset, right?

crates/matrix-sdk-common/src/linked_chunk/as_vector.rs
Hywan: Oh, sure! Good idea. I'll update this PR to address that.

crates/matrix-sdk/src/event_cache/room/events.rs
Hywan: This is not a `Stream` :-). If we recreate the `AsVector` everytime, it will lose all its states, and the `VectorDiff` would be pretty messy.

crates/matrix-sdk-common/src/linked_chunk/as_vector.rs
Hywan: Okay, this PR has the correct approach now, but another PR is necessary to fully fix https://github.com/matrix-org/matrix-rust-sdk/issues/4314. I'm already on it, but I think it's nice to keep them separate, otherwise the diff will be a bit confusing.

crates/matrix-sdk/src/event_cache/room/events.rs
```diff
@@ -44,12 +51,22 @@ impl Default for RoomEvents {
...
+            // `as_vector` must return `Some(…)`.
+            .expect("`LinkedChunk` must have been constructor with `new_with_update_history`");
+
+        Self { chunks, chunks_updates_as_vectordiffs, deduplicator: Deduplicator::new() }
     }
 
     /// Clear all events.
+    ///
+    /// All events, all gaps, everything is dropped, move into the void, into
+    /// the ether, forever.
```
bnjbvr, line 67: Having a blast, are you? :D

crates/matrix-sdk/src/event_cache/room/events.rs
```diff
@@ -44,12 +51,22 @@ impl Default for RoomEvents {
 impl RoomEvents {
     /// Build a new [`RoomEvents`] struct with zero events.
     pub fn new() -> Self {
-        Self { chunks: LinkedChunk::new(), deduplicator: Deduplicator::new() }
+        let mut chunks = LinkedChunk::new_with_update_history();
+        let chunks_updates_as_vectordiffs = chunks
+            .as_vector()
+            // SAFETY: The `LinkedChunk` has been built with `new_with_update_history`, so
+            // `as_vector` must return `Some(…)`.
+            .expect("`LinkedChunk` must have been constructor with `new_with_update_history`");
```
bnjbvr, line 59: To avoid this expect, would it make sense that `new_with_update_history()` return the AsVector data structure too?

crates/matrix-sdk/src/event_cache/room/events.rs
bnjbvr: Would we have to re-create an `AsVector`? I'm not sure about that.

Imagine we had:

```rust
pub fn as_vector(&self) -> ... {
  self.chunks.as_vector()
}
```

The resulting `AsVector` could live as long as the caller wants, and it would always be sync'd with the underlying linked chunk, as long as it lived. In that case, we wouldn't have to store anything new in the linked chunk. Is there any reason we cannot do this?

bnjbvr (commented):
Still an open question before I approve, but this looks like this is on the right track!

crates/matrix-sdk/src/event_cache/room/events.rs
Hywan: You have no idea!

crates/matrix-sdk/src/event_cache/room/events.rs
Hywan: Hmm, `new_with_update_history()` could return `(LinkedChunk, Update)`, but I don't see why it should return an `AsVector` directly. Problem is: to compute an `AsVector`, it needs an `&mut Update`, so the constructor would have to return `(LinkedChunk, &mut Update)`, where the `&mut` is a mutable reference to `LinkedChunk` itself. Er, that's not possible.

Hywan:
> Would we have to re-create an `AsVector`?

First off, we create `AsVector` right after `LinkedChunk` because `AsVector` drains all its updates when constructed. What it means is: updates that are past the creation of `AsVector` are ignored. And that's mandatory because `AsVector` computes its initial states by looking at all chunks, that's much more efficient. `AsVector` will map `Update`s to `VectorDiff`s starting from the state of the `LinkedChunk` at the creation time of `AsVector`.

If we re-create an `AsVector` every time, I mean, it's possible and it would be correct, but that's clearly a waste of time and resources. `AsVector` would have to re-calculate its initial states by looping over all chunks. That's not huge, but still.

Also, this `AsVector` will be used by `RoomEvents` to broadcast the `VectorDiff` to the `Timeline` via `RoomEventCacheInner`. So the caller is technically `RoomEventCacherInner`, which holds a `RoomEvents`, but I think it's much simpler for `RoomEvents` to hold `AsVector` rather than `RoomEventCacheInner` to fetch a new `AsVector` every time a `RoomEventCacheUpdate` must be sent.
Either way, `AsVector` has to be stored somewhere: either by `RoomEventCacheInner` or by `RoomEvents`. I thought it was easier on `RoomEvents` because it's more confined and easier to test (one less layer).

I've always seen `RoomEvents` as a bridge between `EventCache` and `LinkedChunk`. Within this frame, it makes sense to have `AsVector` inside `RoomEvents`. `RoomEventCacheInner` and `RoomEvents` have the same lifetime since the former holds the latter.

crates/matrix-sdk/src/event_cache/room/events.rs
bnjbvr: Ok, too bad!

bnjbvr (approved):
> Also, this AsVector will be used by RoomEvents to broadcast the VectorDiff to the Timeline via RoomEventCacheInner. So the caller is technically RoomEventCacherInner, which holds a RoomEvents, but I think it's much simpler for RoomEvents to hold AsVector rather than RoomEventCacheInner to fetch a new AsVector every time a RoomEventCacheUpdate must be sent.
Either way, AsVector has to be stored somewhere: either by RoomEventCacheInner or by RoomEvents. I thought it was easier on RoomEvents because it's more confined and easier to test (one less layer).

Thanks, I'm convinced by this :+1:

PR #4323, opened 2024-11-26
feat(ffi): wrap Ruma MediaSources and run validations before passing them over FFI

stefanceriu:
Ruma doesn't currently validate mxuri's and as such `MediaSource`s passed over FFI can contain invalid/empty URLs. 

This change introduces a wrapper type around Ruma's and failable transformations so that appropiate actions can be taken beforehand e.g. returning a `TimelineItemContent::FailedToParseMessageLike` or nil-ing out the thumbnail info.


poljar (approved):
Indeed, that's the nicest way of handling this, despite the extra wrapper type.

PR #4324, opened 2024-11-26
feat(media)!: Make all fields of `Thumbnail` required

zecakeh:
`Thumbnail` is used to provide the data and information when uploading a thumbnail along an attachment. It seems sensible to assume that if a client is able to generate a thumbnail, it should be able to get all this information for it too.
A thumbnail with no information is not really useful, as we don't know when it could be used instead of the original image.

It allows to simplify the send queue to not have to handle the case where the dimensions of the thumbnail are missing.

Removes `BaseThumbnailInfo` in the process by merging its fields in `Thumbnail` directly.

There are also 2 pre-commits to refactor part of the code to simplify the changes.

cc @bnjbvr.

### DIFF
--- a/bindings/matrix-sdk-ffi/src/timeline/mod.rs
+++ b/bindings/matrix-sdk-ffi/src/timeline/mod.rs
@@ -24,7 +24,7 @@ use matrix_sdk::crypto::CollectStrategy;
 use matrix_sdk::{
     attachment::{
         AttachmentConfig, AttachmentInfo, BaseAudioInfo, BaseFileInfo, BaseImageInfo,
-        BaseThumbnailInfo, BaseVideoInfo, Thumbnail,
+        BaseVideoInfo, Thumbnail,
     },
     deserialized_responses::{ShieldState as SdkShieldState, ShieldStateCode},
     room::edit::EditedContent as SdkEditedContent,
@@ -53,7 +53,7 @@ use ruma::{
         },
         AnyMessageLikeEventContent,
     },
-    EventId,
+    EventId, UInt,
 };
 use tokio::{
     sync::Mutex,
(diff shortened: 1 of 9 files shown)


bindings/matrix-sdk-ffi/src/timeline/mod.rs
```diff
@@ -144,19 +144,26 @@ fn build_thumbnail_info(
...
+                .and_then(|u| UInt::try_from(u).ok())
+                .ok_or(RoomError::InvalidAttachmentData)?;
+            let width = thumbnail_info
+                .width
+                .and_then(|u| UInt::try_from(u).ok())
+                .ok_or(RoomError::InvalidAttachmentData)?;
+            let size = thumbnail_info
+                .width
+                .and_then(|u| UInt::try_from(u).ok())
+                .ok_or(RoomError::InvalidAttachmentData)?;
```
bnjbvr, line 158: Since we're failing if any's missing, maybe the FFI's `ThumbnailInfo` could also have non-optional fields for height/width/size? (also the last one is wrong, `let size = thumbnail_info.width`)

bnjbvr (approved):
Super nice refactoring, thanks. Just one small comment about the FFI layer, and we should be good to go.

stefanceriu:
> @jmartinesp @stefanceriu can you confirm it's OK to always require a height/width/size for a thumbnail?

I think it's fine, yes, we were already enforcing them on the FFI level on the same `if a client is able to generate a thumbnail, it should be able to get all this information for it too` logic

zecakeh:
So I should create a new type for sending? If I'm not mistaken that also means creating new `ImageInfo` and `VideoInfo` for sending, or changing the API to look more like the Rust API (with `Thumbnail`, `BaseImageInfo` and `BaseVideoInfo`).

bnjbvr:
> So I should create a new type for sending? If I'm not mistaken that also means creating new ImageInfo and VideoInfo for sending, or changing the API to look more like the Rust API (with Thumbnail, BaseImageInfo and BaseVideoInfo).

Unless I'm mistaken, I think your change is fine ad we don't need a new type for sending, because it only touches places where we're sending a `Thumbnail`, right?

zecakeh:
I was talking about the FFI crate, if that's not clear.

bnjbvr:
Ah yeah, I missed that the `ThumbnailInfo` type was used for both sending and reading, at the FFI layer. Then we can merge this PR as is; no need to reintroduce moar types there :)

Thanks @zecakeh as usual for the nice PR!

PR #4325, opened 2024-11-26
fix(room): make the room previews more reliable

### LINKED ISSUE #3483: ci: `test_room_preview` intermittently fails

bnjbvr:
This should help make room previews more reliable:

- the behavior of `Room::history_visibility` was to *always* return an `HistoryVisibility` value, even when the event is missing. In case it was missing, it would return `WorldReadable`, which is against the spec, which states that it should be `Shared`, in that case.
- as noted in https://github.com/matrix-org/matrix-rust-sdk/issues/3483#issuecomment-2149249612, the `m.room.create` and `m.room.history_visibility` MUST be present to correctly calculate the `room_type` and `is_world_readable` fields from a `RoomPreview`.

Solution for 1 consists in splitting the method into two: one that returns an `Option<HistoryVisibility>`, and another that returns the better sensible default as specified by the spec.

Solution for 2 consists in adding those two events to the required_state list for the room list service. For sync v2, we're getting all the events anyways, so we should handle them correctly.

Fixes #3483.


poljar (approved):

bnjbvr:
Latest `fixup!` commit includes a fix for the integration test; the behavior has slightly changed when we try to sync a room's members while being invited. If we don't know about the history visibility, we now assume it's "shared" instead of "world visible". As a result, we don't even run the request for `sync_members()`, which means what was a failure before (forward from the synapse server) becomes a soft pass (no request is spawn, `sync_members()` early returns `Ok`).

PR #4327, opened 2024-11-26
fix(common): `LinkedChunk` emits an `Update::NewItemsChunk` when constructed

Hywan:
This patch updates `LinkedChunk::new_with_update_history` to emit an `Update::NewItemsChunk` because the first chunk is created and it must emit an update accordingly.

The fix is 9 lines changes, the rest is only test updates.

---

* Fix https://github.com/matrix-org/matrix-rust-sdk/issues/4314

### DIFF
--- a/crates/matrix-sdk-common/src/linked_chunk/mod.rs
+++ b/crates/matrix-sdk-common/src/linked_chunk/mod.rs
@@ -276,15 +276,24 @@ impl<const CAP: usize, Item, Gap> LinkedChunk<CAP, Item, Gap> {
     /// [`ObservableUpdates::take`] method must be called to consume and
     /// clean the updates. See [`Self::updates`].
     pub fn new_with_update_history() -> Self {
+        let first_chunk_identifier = ChunkIdentifierGenerator::FIRST_IDENTIFIER;
+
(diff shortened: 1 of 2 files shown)


crates/matrix-sdk-common/src/linked_chunk/mod.rs
```diff
@@ -1474,11 +1483,27 @@ mod tests {
...
+    }
+
     #[test]
     fn test_push_items() {
         use super::Update::*;
 
         let mut linked_chunk = LinkedChunk::<3, char, ()>::new_with_update_history();
+
+        // Ignore initial update.
+        let _ = linked_chunk.updates().unwrap().take();
```
bnjbvr, line 1505: Maybe we could also check (here and everywhere below) that the initial update is what we expected?

bnjbvr (approved):
Sweet, thanks!

crates/matrix-sdk-common/src/linked_chunk/mod.rs
Hywan: That's the `test_new_with_initial_update` test, just above this one.

PR #4329, opened 2024-11-26
fix(media): Make sure that local URIs only try to get media from the cache and ignore requested dimensions and change format used in media cache

zecakeh:
Avoids a tricky API where the users need to make exactly the right media request to get the local file in cache, and avoids to make a request over the network when we know that the file should only be found locally. 

cc @bnjbvr, I don't think this is exactly what you had in mind but it does the job.

### DIFF
--- a/crates/matrix-sdk-base/src/store/migration_helpers.rs
+++ b/crates/matrix-sdk-base/src/store/migration_helpers.rs
@@ -31,20 +31,27 @@ use ruma::{
             guest_access::RoomGuestAccessEventContent,
             history_visibility::RoomHistoryVisibilityEventContent,
             join_rules::RoomJoinRulesEventContent,
+            message::{ImageMessageEventContent, MessageType, RoomMessageEventContent},
             name::{RedactedRoomNameEventContent, RoomNameEventContent},
             tombstone::RoomTombstoneEventContent,
             topic::RoomTopicEventContent,
+            MediaSource,
         },
         EmptyStateKey, EventContent, RedactContent, StateEventContent, StateEventType,
     },
-    OwnedRoomId, OwnedUserId, RoomId,
+    owned_mxc_uri, uint, OwnedRoomId, OwnedTransactionId, OwnedUserId, RoomId, TransactionId, UInt,
 };
 use serde::{Deserialize, Serialize};
 
+use super::{
+    ChildTransactionId, DependentQueuedRequest, DependentQueuedRequestKind, SentRequestKey,
+    SerializableEventContent,
+};
 #[cfg(feature = "experimental-sliding-sync")]
 use crate::latest_event::LatestEvent;
 use crate::{
     deserialized_responses::SyncOrStrippedState,
+    media::MediaRequestParameters,
     rooms::{
         normal::{RoomSummary, SyncInfo},
         BaseRoomInfo, RoomNotableTags,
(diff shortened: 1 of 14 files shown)


bnjbvr (changes requested):
Thanks; I haven't looked into it too much, but since this changes the format of the DependentRequestKins, it needs a migration (that can be *at least* a plain removal of pending dependent events). I'll take a deeper look later.

zecakeh:
> Thanks; I haven't looked into it too much, but since this changes the format of the DependentRequestKins, it needs a migration (that can be _at least_ a plain removal of pending dependent events). I'll take a deeper look later.

Oh you are right, I forgot about that. Looking more into it, it seems complicated to migrate the cached media though. We would need the list of migrated URIs from the `StateStore` and then remove or migrate the media in the `EventCacheStore`. We would not need to migrate the cached media if we could access it via its MXC URI alone with a new API, like you suggested. So it seems to be the simplest solution and I am going to work on that.

zecakeh:
Migrations added.

bnjbvr:
Thanks! I'll get back to reviewing this next week.

crates/matrix-sdk-base/src/store/migration_helpers.rs
```diff
@@ -253,3 +260,179 @@ impl From<RoomNameEventContentV1> for RoomNameEventContent {
         RoomNameEventContent::new(value.name.unwrap_or_default())
     }
 }
+
+/// [`DependentQueuedRequest`] v1.
+///
+/// See the docs of [`DependentQueuedRequestKindV1`] for details on the
+/// migration.
```
bnjbvr, line 267: Instead of duplicating `DependentQueuedRequestV1`, any chance we could deserialize `kind` from a new `SerializedDependentQueueRequestKind` struct, which would have a `serde(untagged)` annotation on top, and would try the latest format, and the previous v1 format if the latest didn't work? The v1 format only needs the previous format for the `FinishUpload` variant, not the others, if what I'm thinking about is correct.

crates/matrix-sdk/src/media.rs
```diff
@@ -679,4 +705,44 @@ impl Media {
...
+    ///
+    /// This uses a MXC ID that is only locally valid.
+    pub(crate) fn make_local_uri(txn_id: &TransactionId) -> OwnedMxcUri {
+        // This mustn't represent a potentially valid media server, otherwise it'd be
+        // possible for an attacker to return malicious content under some
+        // preconditions (e.g. the cache store has been cleared before the upload
+        // took place). To mitigate against this, we use the .localhost TLD,
+        // which is guaranteed to be on the local machine. As a result, the only attack
+        // possible would be coming from the user themselves, which we consider a
+        // non-threat.
```
bnjbvr, line 720: Can you put this comment next to `LOCAL_MXC_SERVER_NAME` instead?

crates/matrix-sdk/src/send_queue/upload.rs
```diff
@@ -308,11 +254,10 @@ impl QueueStorage {
...
 
             // Rename the thumbnail too, if needs be.
-            if let Some((info, new_source)) =
-                thumbnail_info.as_ref().zip(sent_media.thumbnail.clone())
+            if let Some((txn, new_source)) =
+                thumbnail_upload_txn.as_ref().zip(sent_media.thumbnail.clone())
             {
-                let from_req =
-                    make_local_thumbnail_media_request(&info.txn, info.height, info.width);
+                let from_req = Media::make_local_file_media_request(txn);
```
bnjbvr, line 260: Are we sure we want to do this? This cached the thumbnail with the exact same dimensions the user uploaded, which is actually nice. With this change, the media stored in the cache is renamed with an incorrect format (File, for a thumbnail), and it's unclear for external users that they can request the thumbnail as a file.

bnjbvr (changes requested):
This PR became a bit big, and I think it could be split into smaller parts to help understanding what's going on, for the sake of reviews, and also for correctness:

- one PR could introduce the new method to get a media entry by URI only (without a format), with the refactoring to move the local URL scheme into the Media object as you did
- one other PR could, as an optimization, migrate the format of the `FinishUpload`. But see my comment below; I think we're now storing the final thumbnail as a `File`, which is not obvious/documented.


So one alternative would be to get rid of the migration altogether, and keep the dimensions for the thumbnail in the DB, *while* keeping the new method for retrieving the local cached thumbnail.

crates/matrix-sdk/src/send_queue/upload.rs
zecakeh: 1. It is not an "incorrect" format. What is important to remember here is that the format reflects only the request that would be made to the homeserver. There is no link between the format of the files stored in the media cache and their purpose in an event.

    If we were to use the `Thumbnail` format, it should reflect that it is a thumbnail of the original file of the event. The way that it is stored actually means that it is a thumbnail of the thumbnail file of the event.
    
    To clarify, with the Rust types:

    ```rust
    // What we have currently.
    // This is requesting a server-generated thumbnail of the thumbnail file
    // at the same dimensions as the thumbnail file.
    let request_thumbnail_of_thumbnail_file = MediaRequestParameters {
      source: thumbnail_file_source,
      format: MediaFormat::Thumbnail(settings_from_thumbnail_metadata),
    };

    // If we wanted to use a `Thumbnail` format, this would make more sense, imo.
    // This is requesting a server-generated thumbnail of the original file
    // at the same dimensions as the thumbnail file.
    let request_thumbnail_of_original_file = MediaRequestParameters {
      source: original_file_source,
      format: MediaFormat::Thumbnail(settings_from_thumbnail_metadata),
    };
    
    // This is correct, we are requesting the thumbnail file to the server.
    let request_thumbnail_file = MediaRequestParameters {
      source: thumbnail_file_source,
      format: MediaFormat::File,
    };
    ```
2. Clients might not request the exact dimensions of the thumbnail. Given that the API allows to request a thumbnail the fits inside given dimensions with `scale`, the client might not bother at all with the original dimensions of the thumbnail and always request, for example, a thumbnail that fits inside a `600x400` rectangle.

    So the need to request _exactly_ the dimensions of the thumbnail to hit the cache for locally-sent thumbnails is already something that needs to be documented for clients.
3. Clients **cannot** request a server-generated thumbnail of a file that is encrypted, it would be a wasteful request that always returns a 404. Therefore a request with the `Thumbnail` format should **never** be made in an encrypted room. Clients can only request the original file or the thumbnail file, so the format is always `File`, which means that the file that was cached will never be accessed.

    With the current code that uses the `Thumbnail` format, if a client wants to hit the media cache for locally-sent events in an encrypted room, they need special code to know if the media was sent from the client, and then only in this case try to request a server-generated thumbnail of the thumbnail file with the thumbnail file's dimensions. Or they don't treat encrypted and non-encrypted media differently and always make a wasteful request.
4. Since we enforce that thumbnails that were sent by the SDK provide all their metadata, a client should be able to request the thumbnail file instead of a server-generated thumbnail in all rooms if it matches the dimensions that they want.

Given all that, storing the thumbnail with a `File` format is less problematic imo.

Since points 3 and 4 mention logic that needs to be implemented in every client, maybe we can do it in the SDK, by improving the code of [`Media::get_thumbnail()`](https://matrix-org.github.io/matrix-rust-sdk/matrix_sdk/media/struct.Media.html#method.get_thumbnail) (which is subpar currently anyway).

zecakeh:
> one PR could introduce the new method to get a media entry by URI only (without a format), with the refactoring to move the local URL scheme into the Media object as you did

Done in #4387.

crates/matrix-sdk/src/send_queue/upload.rs
bnjbvr: Ah yeah, that makes sense, I hadn't realized about the first point :facepalm: Bonus points if we document all of this super clearly/cleanly as part of the `Media::get_thumbnail()` refactor.

PR #4331, opened 2024-11-26
chore(ui): Have a single code for all kind of timeline item insertions

Hywan:
(Extracted from https://github.com/matrix-org/matrix-rust-sdk/pull/3512/)

This patch tries to unify the code to insert timeline item at position `Start` and `End`. We can clearly have the same code for these `TimelineItemPosition`s. It doesn't make sense to have separate implementations, especially knowing that `End` had duplicated events check while `Start` had not.

This code only and strictly code moving, modulo the `position` handling, but nothing has changed apart from that.

---


crates/matrix-sdk-ui/src/timeline/event_handler.rs
```diff
@@ -1150,13 +1155,27 @@ impl<'a, 'o> TimelineEventHandler<'a, 'o> {
...
+                //
+                // * _push back_:
+                //   * if the index represents the end of items
+                //
+                // * _insert at_:
+                //   * for any other index value
+                if must_push_front || (insert_index == 0 && self.items.is_empty().not()) {
+                    trace!("Adding new remote timeline item at the front");
                     self.items.push_front(new_item);
+                } else if insert_index == self.items.len() {
```
bnjbvr, line 1173: The order was very deliberate before, with the assumption that new events coming from sync will happen more frequently that prepending (because loading from cache). Can you revert and put the `if insert_index == self.items.len()` check first, please?

crates/matrix-sdk-ui/src/timeline/event_handler.rs
```diff
@@ -1060,86 +1060,91 @@ impl<'a, 'o> TimelineEventHandler<'a, 'o> {
...
+                        Some(self.items.remove(idx).internal_id.clone())
+                    } else {
+                        None
+                    }
+                };
 
                 // Insert the next item after the latest event item that's not a
                 // pending local echo, or at the start if there is no such item.
-                let insert_idx = latest_event_idx.map_or(0, |idx| idx + 1);
+                let (insert_index, must_push_front) = match position {
```
bnjbvr, line 1130: can we keep the shorter `insert_idx` name, please?

crates/matrix-sdk-ui/src/timeline/event_handler.rs
```diff
@@ -1150,13 +1155,27 @@ impl<'a, 'o> TimelineEventHandler<'a, 'o> {
...
+                // * _push front_:
+                //   * if `must_push_front`, i.e. if the position is a `TimelinePosition::Start`
+                //   * if the index is 0 _but_ the items aren't empty
+                //
+                // * _push back_:
+                //   * if the index represents the end of items
+                //
+                // * _insert at_:
+                //   * for any other index value
+                if must_push_front || (insert_index == 0 && self.items.is_empty().not()) {
```
bnjbvr, line 1170: nit: let's use a `!` operator, please :P `not` is overkill

bnjbvr, line 1170: I found the comment confusing, but that's because the code is a bit confusing too: we don't need `must_push_front`, since the insert_index will be correct in that case. I think you can actually keep the previous code (that preferred push back over push front, see comment below) and it would just work fine. 

Once that's confirmed working, can you keep the previous comment, maybe tweak it to explain why it's not a coincidence that we start with `push_back`?

crates/matrix-sdk-ui/src/timeline/event_handler.rs
bnjbvr, line 1130: (spoiler alert: we don't need `must_push_front`, since inserting at `insert_idx = 0` means pushing front…)

bnjbvr (dismissed):

crates/matrix-sdk-ui/src/timeline/event_handler.rs
Hywan: It becomes:

```rust
if !must_push_front && (insert_index == self.items.len()) {
    self.items.push_back(new_item);
} else if must_push_front || insert_index == 0 {
    self.items.push_front(new_item);
} else {
    self.items.insert(insert_index, new_item);
}
```

which can be rewritten to:

```rust
if !must_push_front {
    if insert_index == self.items.len() {
        self.items.push_back(new_item);
    } else {
        self.items.insert(insert_index, new_item);
    }
} else {
    self.items.push_front(new_item);
}
```

for a bit more clarity.

crates/matrix-sdk-ui/src/timeline/event_handler.rs
Hywan: Rephrased.

crates/matrix-sdk-ui/src/timeline/event_handler.rs
Hywan: It becomes:

```rust
if !must_push_front && (insert_index == self.items.len()) {
    self.items.push_back(new_item);
} else if must_push_front || insert_index == 0 {
    self.items.push_front(new_item);
} else {
    self.items.insert(insert_index, new_item);
}
```

crates/matrix-sdk-ui/src/timeline/event_handler.rs
Hywan: Done.

Hywan: Alternatively, we can insert inside the `match position`. A bit more repetitive code, but it's much much clear. Let me try that in the last commit.

poljar (approved):
The commit message looks a bit rough grammar-wise as well, perhaps something like:

> chore(ui): Unify the logic for timeline item insertions

> This patch unifies the logic for inserting timeline items at Start and End positions. Both TimelineItemPositions can share the same implementation, making separate logic unnecessary. Previously, End included a duplicated events check as well, while Start did not, leading to inconsistency.

> The changes strictly involve moving and refactoring, with no functional modifications.

crates/matrix-sdk-ui/src/timeline/event_handler.rs
```diff
@@ -1060,88 +1060,71 @@ impl<'a, 'o> TimelineEventHandler<'a, 'o> {
...
-                        if old_item.content.is_redacted() && !item.content.is_redacted() {
-                            warn!("Got original form of an event that was previously redacted");
-                            item.content = item.content.redact(&self.meta.room_version);
-                            item.reactions.clear();
+                // This block tries to find duplicated events.
+
+                let removed_event_item_id = {
+                    // Look if we already have a corresponding item somewhere, based on the
+                    // transaction id (if a local echo) or the event id (if a
+                    // duplicate remote event).
```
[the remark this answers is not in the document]
Hywan: :+1:

crates/matrix-sdk-ui/src/timeline/event_handler.rs
```diff
@@ -1060,88 +1060,71 @@ impl<'a, 'o> TimelineEventHandler<'a, 'o> {
...
+                        txn_id.is_some() && it.transaction_id() == txn_id.as_deref()
+                            || it.event_id() == Some(event_id)
+                    });
+
+                    if let Some((idx, old_item)) = result {
+                        if old_item.as_remote().is_some() {
+                            // Item was previously received from the server. This should be very
+                            // rare normally, but with the sliding- sync
+                            // proxy, it is actually very common.
+                            // NOTE: SS proxy workaround.
```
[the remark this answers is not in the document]
Hywan: :+1:

Fortunately, this is going to be removed in fire really soon.

crates/matrix-sdk-ui/src/timeline/event_handler.rs
```diff
@@ -1150,14 +1133,47 @@ impl<'a, 'o> TimelineEventHandler<'a, 'o> {
...
+                            self.items.push_front(new_item);
+                        } else {
+                            trace!(insert_idx, "Adding new remote timeline item at specific index");
+                            self.items.insert(insert_idx, new_item);
+                        }
+                    }
+
+                    p => unreachable!(
+                        "An unexpected `TimelineItemPosition` has been received: {p:?}"
+                    ),
```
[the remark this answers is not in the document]
Hywan: Yeah… it's unfortunate this is needed, but the upper match-arm makes this really unreachable.

crates/matrix-sdk-ui/src/timeline/event_handler.rs
bnjbvr: I still think what I suggested would work, if you both kept only the `insert_idx` and reverted the order of the final push statements as it were back in the original form:

- in the case of inserting at the front, insert_idx would be 0, and when the list is empty that would result in a push_back; when it's not, it would be a push front
- in the case of inserting at the back, we compute the proper insert_idx anyways.

And I mean, it's OK to change the test expectations when they lead to the same outcome ^^ `push_back` or `push_front` on an empty array yields to the same outcome. But OK, the new code works too.

PR #4334, opened 2024-11-27
refactor(linked chunk): replace `LinkedChunk::len()` with a simpler impl

### LINKED ISSUE #4332: linked chunk: get rid of `LinkedChunk::len()`
It's unused, and it's trivial to recompute.

bnjbvr:
It's unused so it's mostly cosmetic, and it's trivial to reimplement using `linked_chunk.items().count()`, as shown in tests.

Fixes #4332.

### DIFF
--- a/crates/matrix-sdk-common/src/linked_chunk/mod.rs
+++ b/crates/matrix-sdk-common/src/linked_chunk/mod.rs
@@ -233,9 +233,6 @@ pub struct LinkedChunk<const CHUNK_CAPACITY: usize, Item, Gap> {
     /// The links to the chunks, i.e. the first and the last chunk.
     links: Ends<CHUNK_CAPACITY, Item, Gap>,
 
-    /// The number of items hold by this linked chunk.
-    length: usize,
-
     /// The generator of chunk identifiers.
     chunk_identifier_generator: ChunkIdentifierGenerator,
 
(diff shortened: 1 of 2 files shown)


crates/matrix-sdk-common/src/linked_chunk/mod.rs
```diff
@@ -1440,6 +1409,17 @@ mod tests {
...
+        /// Returns the number of items of the linked chunk.
+        fn len(&self) -> usize;
+    }
+
+    impl<const CAP: usize, Item, Gap> LinkedChunkLenExt for LinkedChunk<CAP, Item, Gap> {
+        fn len(&self) -> usize {
+            self.items().count()
+        }
+    }
+
```
poljar, line 1422: Why a trait and not a normal function in the `impl` block of `LinkedChunk` that is guarded by a `#[cfg(test)]` attribute?

poljar (approved):

crates/matrix-sdk-common/src/linked_chunk/mod.rs
bnjbvr: Could even be a public function if we wanted to, to maintain backwards compat.

poljar: Wouldn't mind that either, it's just that the trait feels heavy handed for what we're achieving here.

bnjbvr: The good part is that it makes it a non-breaking change :sunglasses:

Hywan (dismissed):
Thanks for the cleanup! I would remove the trait entirely.

Hywan (approved):

PR #4338, opened 2024-11-27
feat(room): add `JoinRequest` subscriptions

jmartinesp:
## What

This PR adds support for listening to changes in requests to join a room (aka: knock memberships): to do this, we listen to new room member events to trigger the emission of a new list of join requests, as well as to the `RoomInfo::members_synced` flag to understand if we need to reload the room members because a gappy sync happened

Since having too many/old requests to join to display can be an UX problem, I added the concept of a 'seen' join request, and an internal set of join request ids (these referencing the knock membership event), so we can mark some of these join requests as seen and filter them out in the clients at will. Also, when a join request is marked as seen, a new list of join requests is emitted with the updated value.

## Why

When a user of a knockable room has the power level to invite to or kick members from the room, there should be a way to display the incoming join requests in a live manner so this user can know about them and act on them ASAP, instead of having to manually reload the room member list when performing some UI interaction: that's why we need a way to subscribe to these kinds of changes in a live manner.

However, having some component that persistently displays the latest requests to join and having no way to dismiss them can make the room UI difficult to use in the clients, so there should be some way to ignore the displayed requests, or mark them as 'seen' so they can still be reviewed but they can be hidden from the user in that case.

This PR should be reviewed commit-by-commit, since it's fairly large.

Signed-off-by:

### DIFF
--- a/bindings/matrix-sdk-ffi/src/room.rs
+++ b/bindings/matrix-sdk-ffi/src/room.rs
@@ -912,17 +912,17 @@ impl Room {
         Ok(())
     }
 
-    /// Subscribes to requests to join this room, using a `listener` to be
-    /// notified of the changes.
+    /// Subscribes to requests to join this room (knock member events), using a
+    /// `listener` to be notified of the changes.
     ///
     /// The current requests to join the room will be emitted immediately
     /// when subscribing, along with a [`TaskHandle`] to cancel the
     /// subscription.
-    pub async fn subscribe_to_join_requests(
+    pub async fn subscribe_to_knock_requests(
         self: Arc<Self>,
-        listener: Box<dyn JoinRequestsListener>,
+        listener: Box<dyn KnockRequestsListener>,
     ) -> Result<Arc<TaskHandle>, ClientError> {
-        let stream = self.inner.subscribe_to_join_requests().await?;
+        let stream = self.inner.subscribe_to_knock_requests().await?;
 
         let handle = Arc::new(TaskHandle::new(RUNTIME.spawn(async move {
             pin_mut!(stream);
@@ -935,8 +935,8 @@ impl Room {
     }
 }
 
-impl From<matrix_sdk::room::request_to_join::JoinRequest> for JoinRequest {
-    fn from(request: matrix_sdk::room::request_to_join::JoinRequest) -> Self {
+impl From<matrix_sdk::room::knock_requests::KnockRequest> for KnockRequest {
+    fn from(request: matrix_sdk::room::knock_requests::KnockRequest) -> Self {
         Self {
             event_id: request.event_id.to_string(),
             user_id: request.member_info.user_id.to_string(),
@@ -946,20 +946,20 @@ impl From<matrix_sdk::room::request_to_join::JoinRequest> for JoinRequest {
             reason: request.member_info.reason.clone(),
             timestamp: request.timestamp.map(|ts| ts.into()),
             is_seen: request.is_seen,
-            actions: Arc::new(JoinRequestActions { inner: request }),
+            actions: Arc::new(KnockRequestActions { inner: request }),
         }
     }
 }
 
 /// A listener for receiving new requests to a join a room.
 #[matrix_sdk_ffi_macros::export(callback_interface)]
-pub trait JoinRequestsListener: Send + Sync {
-    fn call(&self, join_requests: Vec<JoinRequest>);
+pub trait KnockRequestsListener: Send + Sync {
+    fn call(&self, join_requests: Vec<KnockRequest>);
 }
 
 /// An FFI representation of a request to join a room.
 #[derive(Debug, Clone, uniffi::Record)]
-pub struct JoinRequest {
+pub struct KnockRequest {
     /// The event id of the event that contains the `knock` membership change.
     pub event_id: String,
     /// The user id of the user who's requesting to join the room.
@@ -974,44 +974,44 @@ pub struct JoinRequest {
     pub reason: Option<String>,
     /// The timestamp when this request was created.
     pub timestamp: Option<u64>,
-    /// Whether the request to join has been marked as `seen` so it can be
+    /// Whether the knock request has been marked as `seen` so it can be
     /// filtered by the client.
     pub is_seen: bool,
-    /// A set of actions to perform for this request to join.
-    pub actions: Arc<JoinRequestActions>,
+    /// A set of actions to perform for this knock request.
+    pub actions: Arc<KnockRequestActions>,
 }
 
-/// A set of actions to perform for a request to join.
+/// A set of actions to perform for a knock request.
 #[derive(Debug, Clone, uniffi::Object)]
-pub struct JoinRequestActions {
-    inner: matrix_sdk::room::request_to_join::JoinRequest,
+pub struct KnockRequestActions {
+    inner: matrix_sdk::room::knock_requests::KnockRequest,
 }
 
 #[matrix_sdk_ffi_macros::export]
-impl JoinRequestActions {
-    /// Accepts the request to join by inviting the user to the room.
+impl KnockRequestActions {
+    /// Accepts the knock request by inviting the user to the room.
     pub async fn accept(&self) -> Result<(), ClientError> {
         self.inner.accept().await.map_err(Into::into)
     }
 
-    /// Declines the request to join by kicking the user from the room with an
+    /// Declines the knock request by kicking the user from the room with an
     /// optional reason.
     pub async fn decline(&self, reason: Option<String>) -> Result<(), ClientError> {
         self.inner.decline(reason.as_deref()).await.map_err(Into::into)
     }
 
-    /// Declines the request to join by banning the user from the room with an
+    /// Declines the knock request by banning the user from the room with an
     /// optional reason.
     pub async fn decline_and_ban(&self, reason: Option<String>) -> Result<(), ClientError> {
         self.inner.decline_and_ban(reason.as_deref()).await.map_err(Into::into)
     }
 
-    /// Marks the request as 'seen'.
+    /// Marks the knock request as 'seen'.
     ///
     /// **IMPORTANT**: this won't update the current reference to this request,
     /// a new one with the updated value should be emitted instead.
     pub async fn mark_as_seen(&self) -> Result<(), ClientError> {
-        self.inner.clone().mark_as_seen().await.map_err(Into::into)
+        self.inner.mark_as_seen().await.map_err(Into::into)
     }
 }
 
--- a/bindings/matrix-sdk-ffi/src/room.rs
+++ b/bindings/matrix-sdk-ffi/src/room.rs
@@ -1,7 +1,7 @@
 use std::{collections::HashMap, pin::pin, sync::Arc};
 
 use anyhow::{Context, Result};
-use futures_util::StreamExt;
+use futures_util::{pin_mut, StreamExt};
 use matrix_sdk::{
     crypto::LocalTrust,
     event_cache::paginator::PaginatorError,
@@ -910,6 +910,108 @@ impl Room {
         let (room_event_cache, _drop_handles) = self.inner.event_cache().await?;
         room_event_cache.clear().await?;
         Ok(())
+    }
+
+    /// Subscribes to requests to join this room, using a `listener` to be
+    /// notified of the changes.
+    ///
+    /// The current requests to join the room will be emitted immediately
+    /// when subscribing, along with a [`TaskHandle`] to cancel the
+    /// subscription.
+    pub async fn subscribe_to_join_requests(
+        self: Arc<Self>,
+        listener: Box<dyn JoinRequestsListener>,
+    ) -> Result<Arc<TaskHandle>, ClientError> {
+        let stream = self.inner.subscribe_to_join_requests().await?;
+
+        let handle = Arc::new(TaskHandle::new(RUNTIME.spawn(async move {
+            pin_mut!(stream);
+            while let Some(requests) = stream.next().await {
+                listener.call(requests.into_iter().map(Into::into).collect());
+            }
+        })));
+
+        Ok(handle)
+    }
+}
+
+impl From<matrix_sdk::room::request_to_join::JoinRequest> for JoinRequest {
+    fn from(request: matrix_sdk::room::request_to_join::JoinRequest) -> Self {
+        Self {
+            event_id: request.event_id.to_string(),
+            user_id: request.member_info.user_id.to_string(),
+            room_id: request.room_id().to_string(),
+            display_name: request.member_info.display_name.clone(),
+            avatar_url: request.member_info.avatar_url.as_ref().map(|url| url.to_string()),
+            reason: request.member_info.reason.clone(),
+            timestamp: request.timestamp.map(|ts| ts.into()),
+            is_seen: request.is_seen,
+            actions: Arc::new(JoinRequestActions { inner: request }),
+        }
+    }
+}
+
+/// A listener for receiving new requests to a join a room.
+#[matrix_sdk_ffi_macros::export(callback_interface)]
+pub trait JoinRequestsListener: Send + Sync {
+    fn call(&self, join_requests: Vec<JoinRequest>);
+}
+
+/// An FFI representation of a request to join a room.
+#[derive(Debug, Clone, uniffi::Record)]
+pub struct JoinRequest {
+    /// The event id of the event that contains the `knock` membership change.
+    pub event_id: String,
+    /// The user id of the user who's requesting to join the room.
+    pub user_id: String,
+    /// The room id of the room whose access was requested.
+    pub room_id: String,
+    /// The optional display name of the user who's requesting to join the room.
+    pub display_name: Option<String>,
+    /// The optional avatar url of the user who's requesting to join the room.
+    pub avatar_url: Option<String>,
+    /// An optional reason why the user wants join the room.
+    pub reason: Option<String>,
+    /// The timestamp when this request was created.
+    pub timestamp: Option<u64>,
+    /// Whether the request to join has been marked as `seen` so it can be
+    /// filtered by the client.
+    pub is_seen: bool,
+    /// A set of actions to perform for this request to join.
+    pub actions: Arc<JoinRequestActions>,
+}
+
+/// A set of actions to perform for a request to join.
+#[derive(Debug, Clone, uniffi::Object)]
+pub struct JoinRequestActions {
+    inner: matrix_sdk::room::request_to_join::JoinRequest,
+}
+
+#[matrix_sdk_ffi_macros::export]
+impl JoinRequestActions {
+    /// Accepts the request to join by inviting the user to the room.
+    pub async fn accept(&self) -> Result<(), ClientError> {
+        self.inner.accept().await.map_err(Into::into)
+    }
+
+    /// Declines the request to join by kicking the user from the room with an
+    /// optional reason.
+    pub async fn decline(&self, reason: Option<String>) -> Result<(), ClientError> {
+        self.inner.decline(reason.as_deref()).await.map_err(Into::into)
+    }
+
+    /// Declines the request to join by banning the user from the room with an
+    /// optional reason.
+    pub async fn decline_and_ban(&self, reason: Option<String>) -> Result<(), ClientError> {
+        self.inner.decline_and_ban(reason.as_deref()).await.map_err(Into::into)
+    }
+
+    /// Marks the request as 'seen'.
+    ///
+    /// **IMPORTANT**: this won't update the current reference to this request,
+    /// a new one with the updated value should be emitted instead.
+    pub async fn mark_as_seen(&self) -> Result<(), ClientError> {
+        self.inner.clone().mark_as_seen().await.map_err(Into::into)
     }
 }
 
--- a/crates/matrix-sdk-base/src/rooms/normal.rs
+++ b/crates/matrix-sdk-base/src/rooms/normal.rs
@@ -1356,6 +1356,11 @@ impl RoomInfo {
         self.members_synced = false;
     }
 
+    /// Returns whether the room members are synced.
+    pub fn are_members_synced(&self) -> bool {
+        self.members_synced
+    }
+
     /// Mark this Room as still missing some state information.
     pub fn mark_state_partially_synced(&mut self) {
         self.sync_info = SyncInfo::PartiallySynced;
--- a/crates/matrix-sdk-base/src/rooms/normal.rs
+++ b/crates/matrix-sdk-base/src/rooms/normal.rs
@@ -22,7 +22,7 @@ use std::{
 
 use as_variant::as_variant;
 use bitflags::bitflags;
-use eyeball::{SharedObservable, Subscriber};
+use eyeball::{AsyncLock, ObservableWriteGuard, SharedObservable, Subscriber};
 use futures_util::{Stream, StreamExt};
 #[cfg(feature = "experimental-sliding-sync")]
 use matrix_sdk_common::deserialized_responses::TimelineEventKind;
@@ -52,7 +52,7 @@ use ruma::{
         },
         tag::{TagEventContent, Tags},
         AnyRoomAccountDataEvent, AnyStrippedStateEvent, AnySyncStateEvent,
-        RoomAccountDataEventType, SyncStateEvent,
+        RoomAccountDataEventType, StateEventType, SyncStateEvent,
     },
     room::RoomType,
     serde::Raw,
@@ -77,7 +77,8 @@ use crate::{
     read_receipts::RoomReadReceipts,
     store::{DynStateStore, Result as StoreResult, StateStoreExt},
     sync::UnreadNotificationsCount,
-    Error, MinimalStateEvent, OriginalMinimalStateEvent, RoomMemberships,
+    Error, MinimalStateEvent, OriginalMinimalStateEvent, RoomMemberships, StateStoreDataKey,
+    StateStoreDataValue, StoreError,
 };
 
 /// Indicates that a notable update of `RoomInfo` has been applied, and why.
@@ -167,6 +168,12 @@ pub struct Room {
     /// to disk but held in memory.
     #[cfg(all(feature = "e2e-encryption", feature = "experimental-sliding-sync"))]
     pub latest_encrypted_events: Arc<SyncRwLock<RingBuffer<Raw<AnySyncTimelineEvent>>>>,
+
+    /// A map for ids of room membership events in the knocking state linked to
+    /// the user id of the user affected by the member event, that the current
+    /// user has marked as seen so they can be ignored.
+    pub seen_knock_request_ids_map:
+        SharedObservable<Option<BTreeMap<OwnedEventId, OwnedUserId>>, AsyncLock>,
 }
 
 /// The room summary containing member counts and members that should be used to
@@ -289,6 +296,7 @@ impl Room {
                 Self::MAX_ENCRYPTED_EVENTS,
             ))),
             room_info_notable_update_sender,
+            seen_knock_request_ids_map: SharedObservable::new_async(None),
         }
     }
 
--- a/crates/matrix-sdk-base/src/rooms/normal.rs
+++ b/crates/matrix-sdk-base/src/rooms/normal.rs
@@ -1177,6 +1177,88 @@ impl Room {
     pub fn pinned_event_ids(&self) -> Option<Vec<OwnedEventId>> {
         self.inner.read().pinned_event_ids()
     }
+
+    /// Mark a list of requests to join the room as seen, given their state
+    /// event ids.
+    pub async fn mark_knock_requests_as_seen(&self, user_ids: &[OwnedUserId]) -> StoreResult<()> {
+        let raw_user_ids: Vec<&str> = user_ids.iter().map(|id| id.as_str()).collect();
+        let member_raw_events = self
+            .store
+            .get_state_events_for_keys(self.room_id(), StateEventType::RoomMember, &raw_user_ids)
+            .await?;
+        let mut event_to_user_ids = Vec::with_capacity(member_raw_events.len());
+
+        // Map the list of events ids to their user ids, if they are event ids for knock
+        // membership events. Log an error and continue otherwise.
+        for raw_event in member_raw_events {
+            let event = raw_event.cast::<RoomMemberEventContent>().deserialize()?;
+            match event {
+                SyncOrStrippedState::Sync(SyncStateEvent::Original(event)) => {
+                    if event.content.membership == MembershipState::Knock {
+                        event_to_user_ids.push((event.event_id, event.state_key))
+                    } else {
+                        warn!("Could not mark knock event as seen: event {} for user {} is not in Knock membership state.", event.event_id, event.state_key);
+                    }
+                }
+                _ => warn!(
+                    "Could not mark knock event as seen: event for user {} is not valid.",
+                    event.state_key()
+                ),
+            }
+        }
+
+        let mut current_seen_events_guard = self.seen_knock_request_ids_map.write().await;
+        // We're not calling `get_seen_join_request_ids` here because we need to keep
+        // the Mutex's guard until we've updated the data
+        let mut current_seen_events = if current_seen_events_guard.is_none() {
+            self.load_cached_knock_request_ids().await?
+        } else {
+            current_seen_events_guard.clone().unwrap()
+        };
+
+        current_seen_events.extend(event_to_user_ids);
+
+        ObservableWriteGuard::set(
+            &mut current_seen_events_guard,
+            Some(current_seen_events.clone()),
+        );
+
+        self.store
+            .set_kv_data(
+                StateStoreDataKey::SeenKnockRequests(self.room_id()),
+                StateStoreDataValue::SeenKnockRequests(current_seen_events),
+            )
+            .await?;
+
+        Ok(())
+    }
+
+    /// Get the list of seen knock request event ids in this room.
+    pub async fn get_seen_knock_request_ids(
+        &self,
+    ) -> Result<BTreeMap<OwnedEventId, OwnedUserId>, StoreError> {
+        let mut guard = self.seen_knock_request_ids_map.write().await;
+        if guard.is_none() {
+            ObservableWriteGuard::set(
+                &mut guard,
+                Some(self.load_cached_knock_request_ids().await?),
+            );
+        }
+        Ok(guard.clone().unwrap_or_default())
+    }
+
+    /// This loads the current list of seen knock request ids from the state
+    /// store.
+    async fn load_cached_knock_request_ids(
+        &self,
+    ) -> StoreResult<BTreeMap<OwnedEventId, OwnedUserId>> {
+        Ok(self
+            .store
+            .get_kv_data(StateStoreDataKey::SeenKnockRequests(self.room_id()))
+            .await?
+            .and_then(|v| v.into_seen_knock_requests())
+            .unwrap_or_default())
+    }
 }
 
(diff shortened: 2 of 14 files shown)


bindings/matrix-sdk-ffi/src/room.rs
```diff
@@ -840,6 +840,97 @@ impl Room {
...
+
+    /// Declines the request to join by kicking the user from the room with an
+    /// optional reason.
+    pub async fn decline(&self, reason: Option<String>) -> Result<(), ClientError> {
+        self.inner.decline(reason.as_deref()).await.map_err(Into::into)
+    }
+
+    /// Declines the request to join by banning the user from the room with an
+    /// optional reason.
+    pub async fn decline_and_ban(&self, reason: Option<String>) -> Result<(), ClientError> {
```
ganfra, line 931: maybe we want a ban action, and let the user chain? in the case the user doesn't have the right for both

bindings/matrix-sdk-ffi/src/room.rs
```diff
@@ -840,6 +840,97 @@ impl Room {
...
+}
+
+/// A set of actions to perform for a request to join.
+#[derive(Debug, Clone, uniffi::Object)]
+pub struct RequestToJoinActions {
+    inner: matrix_sdk::room::request_to_join::RequestToJoinRoom,
+}
+
+#[matrix_sdk_ffi_macros::export]
+impl RequestToJoinActions {
```
ganfra, line 917: Missing `mark_as_seen`?

bindings/matrix-sdk-ffi/src/room.rs
jmartinesp: You mean kick, then ban? Can we ban a user who already left the room? 

Also, I'd expect the UI to not display the component that triggers this action rather than controlling trying to workaround the issue here.

jmartinesp: But then that's what we'll do. ` decline` just kicks the user from the room, `decline_and_ban` also [just bans them](https://github.com/matrix-org/matrix-rust-sdk/pull/4338/files#diff-fa0c59c707536a04b3c50458f03c0e9da8fe5bd67536b2655d0e5bb3adc0a0cfR59), we don't do anything else.

jmartinesp:
I converted the PR back to draft, I'm trying to fix a test failure I can't reproduce locally 🫤 .

bindings/matrix-sdk-ffi/src/room.rs
```diff
@@ -908,6 +908,108 @@ impl Room {
...
+            timestamp: request.timestamp.map(|ts| ts.into()),
+            is_seen: request.is_seen,
+            actions: Arc::new(JoinRequestActions { inner: request }),
+        }
+    }
+}
+
+/// A listener for receiving new requests to a join a room.
+#[matrix_sdk_ffi_macros::export(callback_interface)]
+pub trait RequestsToJoinListener: Send + Sync {
```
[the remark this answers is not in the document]
jmartinesp: Done in https://github.com/matrix-org/matrix-rust-sdk/pull/4338/commits/2b9e3302a1a86ce923b7d4387f44bc46e47fb37b

crates/matrix-sdk-base/src/store/memory_store.rs
```diff
@@ -239,6 +247,14 @@ impl StateStore for MemoryStore {
                         .expect("Session data not containing server capabilities"),
                 );
             }
+            StateStoreDataKey::SeenRequestsToJoin(room_id) => {
+                self.seen_requests_to_join.write().unwrap().insert(
+                    room_id.to_owned(),
+                    value
+                        .into_seen_join_requests()
+                        .expect("Session data not a set of ignored join requests"),
```
bnjbvr, line 255: s/join/seen/

crates/matrix-sdk-indexeddb/src/state_store/mod.rs
```diff
@@ -574,6 +581,11 @@ impl_state_store!({
             StateStoreDataKey::ComposerDraft(_) => self.serialize_value(
                 &value.into_composer_draft().expect("Session data not a composer draft"),
             ),
+            StateStoreDataKey::SeenRequestsToJoin(_) => self.serialize_value(
+                &value
+                    .into_seen_join_requests()
+                    .expect("Session data not a set of ignored requests to join"),
```
bnjbvr, line 587: ditto here

crates/matrix-sdk-sqlite/src/state_store.rs
```diff
@@ -1029,6 +1035,11 @@ impl StateStore for SqliteStateStore {
             StateStoreDataKey::ComposerDraft(_) => self.serialize_value(
                 &value.into_composer_draft().expect("Session data not a composer draft"),
             )?,
+            StateStoreDataKey::SeenRequestsToJoin(_) => self.serialize_value(
+                &value
+                    .into_seen_join_requests()
+                    .expect("Session data not a set of ignored requests to join"),
```
bnjbvr, line 1041: ditto

crates/matrix-sdk/src/room/mod.rs
```diff
@@ -3241,6 +3246,17 @@ impl Room {
...
+    /// Subscribes to the set of requests to join that have been marked as
+    /// 'seen'.
+    pub async fn subscribe_to_seen_requests_to_join_ids(
+        &self,
+    ) -> Result<(HashSet<OwnedEventId>, impl Stream<Item = HashSet<OwnedEventId>>)> {
+        let current = self.get_seen_requests_to_join_ids().await?;
+        let subscriber =
+            self.seen_requests_to_join_ids.subscribe().map(|values| values.unwrap_or_default());
+        Ok((current, subscriber))
+    }
```
bnjbvr: Unfortunately, this is false observability. `seen_requests_to_join_ids` exists in `Room`, which is created on-demand, e.g. in `Client::get_room()`, `Client::rooms()`. Given a `RoomId`, there might be multiple, different `Room` objects alive at the same time, all representing the room with that id.

As a result, if you do the following:
- create `Room` object for room id XYZ
- create a second `Room` object for the same room id
- call this method ^ on the second object
- set a seen_requests_to_join on the first object

Then the second Room object will NOT emit an update about the seen_requests_to_join emitted from the first object.

Solving this requires getting back to the drawing board, and figuring a better place where to put the `seen_requests_to_join_ids` set:

- the `BaseRoom` aka `matrix_sdk_base::Room` would be a good candidate, as all the data it contains always links to the same identity. (It'd be worth having some kind of `BaseRoomInner` to make this clearer, but oh well, so many wrappers already.) So it'd be possible to include an `Arc` in there for the set, and modify it from all the `BaseRoom`, then `Room` instances.
- an alternative is to put this into `RoomInfo`, which is already unique per room + observable :eyes: If you've included this set in the `RoomInfo` in a subsequent patch, it might be worth doing it sooner rather than later. Otherwise, putting it in `BaseRoom` might be sufficient.

crates/matrix-sdk/src/room/request_to_join.rs
bnjbvr, line 1: Please add a new license header :) (yay for the free bonus in LOC :D)

crates/matrix-sdk/src/room/request_to_join.rs
```diff
@@ -0,0 +1,95 @@
...
+    events::room::member::OriginalSyncRoomMemberEvent, EventId, OwnedEventId, OwnedMxcUri,
+    OwnedUserId, RoomId,
+};
+
+use crate::{room::RoomMember, Error, Room};
+
+/// A request to join a room with `knock` join rule.
+#[derive(Debug, Clone)]
+pub struct RequestToJoinRoom {
+    room: Arc<Room>,
```
bnjbvr, line 11: (This pattern of using an `Arc<Room>` lets me think that you've ran into the `Room` identity issue I've spotted above. We usually don't store `Room` in `Arc`, because we shouldn't have to.)

bnjbvr (changes requested):
Thanks. I might need a bit of time and to take another look after the fundamental issue has been addressed, to understand a bit better the problem's requirements and the proposed solution.

I appreciate trying to make small commits to make the review easier, but I find that commits changing code that was introduced in previous commits doesn't quite help me, as a reviewer: I'm learning about new concept ABC in commit N, but then in commit N+P it's renamed to XYZ. Would it be possible to fold the commits in such a way that the new names are used from the start, instead?

Also, I haven't seen a test until the 5th commit :fearful: Unit tests are not only useful to ensure high code coverage, but also to make sure that the functionality we're introducing is always tested (even if callers stop using it, after some time). So you get bonus points if whenever you introduce a new functionality/bugfix (public API), it gets tested in the same commit :) (I see one of the last commits does this perfectly, so it's a matter of aligning the rest of the commits from this PR.)

Keep up the good work :muscle:

crates/matrix-sdk/src/room/mod.rs
jmartinesp: > the BaseRoom aka matrix_sdk_base::Room would be a good candidate

I might be wrong, but isn't [there](https://github.com/matrix-org/matrix-rust-sdk/pull/4338/files#diff-07c79eee28a9bd6d7e2c163f4a1b1a9654da0065698126793adc636f6fa846e3R155) where it is at the moment? I think it just gives you the impression that it's in `sdk::Room` because it's using the `Defer` impl to access `BaseRoom`. 

I have another question though: since this data lives in the store, the field gets populated the first time `Room::get_seen_join_requests` is called. Maybe there's a better way to handle this?

bnjbvr: > I might be wrong, but isn't [there](https://github.com/matrix-org/matrix-rust-sdk/pull/4338/files#diff-07c79eee28a9bd6d7e2c163f4a1b1a9654da0065698126793adc636f6fa846e3R155) where it is at the moment? I think it just gives you the impression that it's in sdk::Room because it's using the Defer impl to access BaseRoom.

oh my, absolute facepalm of a review, please ignore me then :'D

> I have another question though: since this data lives in the store, the field gets populated the first time Room::get_seen_join_requests is called. Maybe there's a better way to handle this?

It's fine, I get it that it's an in-memory cache for data that lives in the store, so there's no way to get it sooner than that anyways (and it's good that it's filled lazily on demand too).

crates/matrix-sdk/src/room/request_to_join.rs
jmartinesp: Oops, nice catch. I think my IDE broke somehow and wasn't including the headers anymore in new files 😬 .

crates/matrix-sdk/src/room/request_to_join.rs
jmartinesp: Nice, I think I got over-paranoid with not cloning stuff, so I'm reverting this to using a `Room` directly and cloning it as needed.

jmartinesp:
I hope this is now in a better shape to be reviewed. I also added a few extra tests.

crates/matrix-sdk/src/room/mod.rs
```diff
@@ -3195,6 +3201,135 @@ impl Room {
...
+
+            loop {
+                // This is equivalent to a combine stream operation, triggering a new emission
+                // when any of the branches changes
+                tokio::select! {
+                    Some((next, _)) = requests_stream.next() => { new_event = Some(next); }
+                    Some(next) = seen_request_ids_stream.next() => { seen_ids = next; }
+                    Some(next) = room_info_stream.next() => {
+                        missing_room_members = !next.are_members_synced()
+                    }
```
bnjbvr, line 3244: So, if I'm reading the code correctly: it seems the `select!` tries to do the minimal amount of work, and then the actual work is done a bit after. Another way would have been to essentially take the code that reacts to a change here, and put it immediately in the branches. Any reason why it was not done this way? It would preclude the need for a few local variables, and make reading the code a bit more straightforward (instead of `if this or this or this happened, consider that, and maybe do this or that`, we'd have `if this happened, do that; if thot happened, do thut` etc.)

bindings/matrix-sdk-ffi/src/room.rs
```diff
@@ -911,6 +911,108 @@ impl Room {
...
+
+/// An FFI representation of a request to join a room.
+#[derive(Debug, Clone, uniffi::Record)]
+pub struct JoinRequest {
+    /// The event id of the event that contains the `knock` membership change.
+    pub event_id: String,
+    /// The user id of the user who's requesting to join the room.
+    pub user_id: String,
+    /// The room id of the room whose access was requested.
+    pub room_id: String,
```
bnjbvr, line 968: Is it useful to have that here, since it's accessible via a function on `Room` you can get the room id in another way?

bindings/matrix-sdk-ffi/src/room.rs
```diff
@@ -911,6 +911,108 @@ impl Room {
...
+    pub async fn decline_and_ban(&self, reason: Option<String>) -> Result<(), ClientError> {
+        self.inner.decline_and_ban(reason.as_deref()).await.map_err(Into::into)
+    }
+
+    /// Marks the request as 'seen'.
+    ///
+    /// **IMPORTANT**: this won't update the current reference to this request,
+    /// a new one with the updated value should be emitted instead.
+    pub async fn mark_as_seen(&self) -> Result<(), ClientError> {
+        self.inner.clone().mark_as_seen().await.map_err(Into::into)
```
bnjbvr, line 1014: Is the clone useful here?

crates/matrix-sdk/src/room/mod.rs
```diff
@@ -3195,6 +3201,135 @@ impl Room {
...
+    async fn get_current_join_requests(
+        &self,
+        seen_request_ids: &HashSet<OwnedEventId>,
+    ) -> Result<Vec<JoinRequest>> {
+        Ok(self
+            .members(RoomMemberships::KNOCK)
+            .await?
+            .into_iter()
+            .filter_map(|member| {
+                if let Some(event_id) = member.event().event_id() {
```
bnjbvr, line 3317: style suggestion: you can use `let event_id = member.event().event_id()?;` here, which will get rid of all indent \o/

crates/matrix-sdk/tests/integration/room/joined.rs
```diff
@@ -833,3 +837,111 @@ async fn test_enable_encryption_doesnt_stay_unencrypted() {
...
+    server.mock_get_members().ok(vec![alice_knock_event]).mock_once().mount().await;
+
+    let room = server.sync_joined_room(&client, room_id).await;
+    let stream = room.subscribe_to_join_requests().await.unwrap();
+
+    pin_mut!(stream);
+
+    // We receive an initial request to join from Alice
+    let initial = assert_next_with_timeout!(stream, 100);
+    assert!(!initial.is_empty());
```
bnjbvr, line 870: instead of checking `!is_empty()`, if you check the expected `.len()`, you'll be sure there's exactly 1 request, and no more.

crates/matrix-sdk/tests/integration/room/joined.rs
```diff
@@ -833,3 +837,111 @@ async fn test_enable_encryption_doesnt_stay_unencrypted() {
...
+    let alices_request_to_join = &initial[0];
+    assert_eq!(alices_request_to_join.event_id, alice_knock_event_id);
+    assert!(!alices_request_to_join.is_seen);
+
+    // We then mark the request to join as seen
+    room.mark_join_requests_as_seen(&[alice_knock_event_id.to_owned()]).await.unwrap();
+
+    // Now it's received again as seen
+    let seen = assert_next_with_timeout!(stream, 100);
+    assert!(!seen.is_empty());
```
bnjbvr, line 881: ditto

crates/matrix-sdk/tests/integration/room/joined.rs
```diff
@@ -833,3 +837,111 @@ async fn test_enable_encryption_doesnt_stay_unencrypted() {
...
+    let alices_seen_request_to_join = &seen[0];
+    assert_eq!(alices_seen_request_to_join.event_id, alice_knock_event_id);
+    assert!(alices_seen_request_to_join.is_seen);
+
+    // If we then receive a new member event for Alice that's not 'knock'
+    let alice_join_event_id = event_id!("$alice-join:b.c");
+    let joined_room_builder = JoinedRoomBuilder::new(room_id).add_state_bulk(vec![f
```
bnjbvr, line 888: if you use `add_state_event`, you shouldn't need the `vec![]`

crates/matrix-sdk/tests/integration/room/joined.rs
```diff
@@ -833,3 +837,111 @@ async fn test_enable_encryption_doesnt_stay_unencrypted() {
...
+        .event_id(alice_join_event_id)
+        .sender(alice_user_id)
+        .state_key(alice_user_id)
+        .into_raw_timeline()
+        .cast()]);
+    server.sync_room(&client, joined_room_builder).await;
+
+    // The requests to join are now empty
+    let updated_requests = assert_next_with_timeout!(stream, 100);
+    assert!(updated_requests.is_empty());
```
bnjbvr, line 899: Can you check at the end of this test that the stream is empty?

crates/matrix-sdk/src/room/mod.rs
```diff
@@ -3236,6 +3371,17 @@ impl Room {
             };
         Ok(current_join_request_ids)
     }
+
+    /// Subscribes to the set of requests to join that have been marked as
+    /// 'seen'.
+    pub async fn subscribe_to_seen_join_request_ids(
```
bnjbvr, line 3377: I think this isn't tested in isolation. Maybe we can inline this, and not make it public?

bnjbvr (changes requested):
Okay, it makes a lot more sense with the nicely cut commits, thanks a bunch for splitting them that way and adding tests per commit :heart: 

I've got a few blocking comments below, and one overall comment about the approach:

It seems we could spare the existence of `JoinRequest` overall in the SDK, by emitting stream updates about the set of pending knock member events for that room. Then, at the FFI layer, we'd listen to that stream, and apply the extra `is_seen` bit based on a call to the most recent `get_seen_john_requests()`. It's also fine to have some kind of higher-level object at the FFI layer, that calls the `Room` methods to accept/kick/ban/decline a user's knock, _if_ justified that it would avoid crossing the FFI boundary by doing more work at once.

Doing that would avoid introducing new high-level APIs in the SDK, that can be avoided by calling plain functions. It would mean there would be fewer ways to do the same thing, which has some appeal to me. We would have less public API surface, and thus fewer things to test (although I do really appreciate the new mocking endpoints, that I think we should keep in any case). It means a bit more simplicity for consumers of the library, because there are fewer things to look for, in the documentation.

How does that sound?

crates/matrix-sdk-base/src/store/memory_store.rs
```diff
@@ -91,6 +91,7 @@ pub struct MemoryStore {
     custom: StdRwLock<HashMap<Vec<u8>, Vec<u8>>>,
     send_queue_events: StdRwLock<BTreeMap<OwnedRoomId, Vec<QueuedRequest>>>,
     dependent_send_queue_events: StdRwLock<BTreeMap<OwnedRoomId, Vec<DependentQueuedRequest>>>,
+    seen_requests_to_join: StdRwLock<BTreeMap<OwnedRoomId, HashSet<OwnedEventId>>>,
```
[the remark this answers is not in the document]
jmartinesp: Sure, I can rebase.

jmartinesp:
> It seems we could spare the existence of JoinRequest overall in the SDK, by emitting stream updates about the set of pending knock member events for that room. Then, at the FFI layer, we'd listen to that stream, and apply the extra is_seen bit based on a call to the most recent get_seen_john_requests().

I don't think that would work when you mark a request as seen, since you wouldn't have anything to tell you there are new 'seen' knock requests. You'd have to build that stream of seen ids in the FFI layer if possible, or expose it from the SDK, then combine it with the knock request stream in the FFI layer, which seems like maybe too much logic in the FFI layer and we'd have no way to test it.

bnjbvr:
Fair enough! So maybe the original stream can keep on listening to changes to seen, but yield the knock events / event IDs themselves, rather than `JoinRequest`s? or does that cause too much work at the FFI layer? Let's discuss this a bit sync.

bnjbvr:
Sorry to turn my head on the necessity of having `KnockRequest` (née `JoinRequest`), I was a bit torn about expanding the public API surface. But a fair point has been made by @poljar (thanks!) that it's not clear how to act upon a knock request, e.g. knowing that one should "kick" the sending user to reject the knock request. And it's not something we can improve easily with documentation as far as I know, so the current approach is good, and we can keep the stream emitting `KnockRequest`s then :+1:

crates/matrix-sdk/src/room/mod.rs
```diff
@@ -3192,6 +3192,48 @@ impl Room {
     pub fn observe_live_location_shares(&self) -> ObservableLiveLocation {
         ObservableLiveLocation::new(&self.client, self.room_id())
     }
+
+    /// Mark a list of requests to join the room as seen, given their state
+    /// event ids.
+    pub async fn mark_join_requests_as_seen(&self, event_ids: &[OwnedEventId]) -> Result<()> {
```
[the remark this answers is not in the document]
jmartinesp: What if I still take the list of event ids and resolve the knock member events inside the method, failing if any of the provided event ids is not valid (doesn't exist or is not a knock member event) ?

crates/matrix-sdk/src/room/mod.rs
```diff
@@ -3192,6 +3192,48 @@ impl Room {
...
+        let mut current_seen_events = self.get_seen_join_request_ids().await?;
```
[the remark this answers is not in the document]
jmartinesp: You're right, I forgot I'm not dealing with the underlying observable with rw locks directly... I guess I'll have to add a Mutex somewhere in `BaseRoom` for this, maybe? Is there a better way to handle it?

crates/matrix-sdk/src/room/request_to_join.rs
```diff
@@ -0,0 +1,205 @@
...
+        Self {
+            user_id: member.user_id().to_owned(),
+            display_name: member.display_name().map(ToOwned::to_owned),
+            avatar_url: member.avatar_url().map(ToOwned::to_owned),
+            reason: member.event().reason().map(ToOwned::to_owned),
+        }
+    }
+}
+
+#[cfg(all(test, not(target_arch = "wasm32")))]
```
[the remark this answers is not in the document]
jmartinesp: `MatrixMockServer` doesn't seem to work in that arch: https://github.com/matrix-org/matrix-rust-sdk/blob/9a9730d59ece8dd9ed9ef4ea19bd0e7ad7a4101b/crates/matrix-sdk/src/test_utils/mod.rs#L16. I can add a comment about it.

crates/matrix-sdk/src/room/mod.rs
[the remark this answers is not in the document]
bnjbvr: Yep, that could work too. It might be a bit suboptimal if the events haven't been cached yet, and we need to re-resolve them one way or another. Is there no way to pass the actual events here?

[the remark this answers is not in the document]
jmartinesp: There is, but it would mean having to find and deserialize the raw event to the right types in a couple of places instead of a single one. Another option could be to pass the `KnockJoinRequest` directly, maybe?

crates/matrix-sdk/src/room/mod.rs
```diff
@@ -3195,6 +3201,135 @@ impl Room {
...
+    pub async fn subscribe_to_join_requests(&self) -> Result<impl Stream<Item = Vec<JoinRequest>>> {
+        let this = Arc::new(self.clone());
+
+        let requests_observable =
+            this.client.observe_room_events::<SyncRoomMemberEvent, (Client, Room)>(this.room_id());
+
+        let (current_seen_ids, mut seen_request_ids_stream) =
+            this.subscribe_to_seen_join_request_ids().await?;
+
+        let mut room_info_stream = this.subscribe_info();
```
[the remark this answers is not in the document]
jmartinesp: I can try, but I seem to remember something funky happening to lifetimes when I tried before.

crates/matrix-sdk/src/room/mod.rs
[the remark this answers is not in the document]
bnjbvr: Passing the `KnockRequest` could also work nicely, good idea!

bindings/matrix-sdk-ffi/src/room.rs
jmartinesp: It's probably not needed.

crates/matrix-sdk/tests/integration/room/joined.rs
jmartinesp: `add_state_event` only allows you to pass a `StateTestEvent` instance. I could use `StateTestEvent::Custom`, but then I need to use a hardcoded json.

crates/matrix-sdk/src/room/mod.rs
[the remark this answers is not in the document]
bnjbvr: Ah good point that they're state events, so don't require the event cache \o/

Fine to pass the event ids, and check they match some known knock requests. If they don't, I'm not sure if we'd prefer to filter them out (i.e. ignore them), or return an error indicating so to the caller.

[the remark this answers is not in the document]
jmartinesp: I was going to use the event ids but then I realised `store.get_state_events_for_keys` uses the state_key which is the user id, not the event id. So maybe passing the user id is the best way to handle it: you got a knock request, have the user id as the state key, pass that to `mark_knock_requests_as_seen`, if the state event is in knock state, we mark it as seen.

crates/matrix-sdk/src/room/mod.rs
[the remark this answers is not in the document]
jmartinesp: Once moved, should I add a Mutex to it? Or should I add some helper function to initialise the shared observable with the contents in the Store and return it wrapped in a WriteGuard, then reuse that for both getting the ids and modifying them?

jmartinesp:
I think I added fixes for all the comments above in the 3 new commits uploaded. Sorry that they're not atomic, after the renamings it was quite difficult to upload only partial changes in commits that can be built independently. In any case, I plan to rebase again and re-organize everything before merging.

crates/matrix-sdk-base/src/store/traits.rs
```diff
@@ -1093,8 +1093,8 @@ impl StateStoreDataValue {
     }
 
     /// Get this value if it is the data for the ignored join requests.
-    pub fn into_seen_join_requests(self) -> Option<HashSet<OwnedEventId>> {
-        as_variant!(self, Self::SeenJoinRequests)
+    pub fn into_seen_join_requests(self) -> Option<BTreeMap<OwnedEventId, OwnedUserId>> {
```
bnjbvr, line 1096: Let's rename this to `into_seen_knock_requests`

crates/matrix-sdk-base/src/store/traits.rs
```diff
@@ -1127,7 +1127,7 @@ pub enum StateStoreDataKey<'a> {
     ComposerDraft(&'a RoomId),
 
     /// A list of requests to join in a room marked as seen.
-    SeenJoinRequests(&'a RoomId),
+    SeenKnockRequests(&'a RoomId),
```
bnjbvr: Let's update the doc comment here.

crates/matrix-sdk-indexeddb/src/state_store/mod.rs
```diff
@@ -581,7 +581,7 @@ impl_state_store!({
             StateStoreDataKey::ComposerDraft(_) => self.serialize_value(
                 &value.into_composer_draft().expect("Session data not a composer draft"),
             ),
-            StateStoreDataKey::SeenJoinRequests(_) => self.serialize_value(
+            StateStoreDataKey::SeenKnockRequests(_) => self.serialize_value(
                 &value
                     .into_seen_join_requests()
                     .expect("Session data is not a set of seen join request ids"),
```
bnjbvr, line 587: Let's update `seen join request` to `seen knock request` here.

crates/matrix-sdk-sqlite/src/state_store.rs
```diff
@@ -1035,7 +1035,7 @@ impl StateStore for SqliteStateStore {
             StateStoreDataKey::ComposerDraft(_) => self.serialize_value(
                 &value.into_composer_draft().expect("Session data not a composer draft"),
             )?,
-            StateStoreDataKey::SeenJoinRequests(_) => self.serialize_value(
+            StateStoreDataKey::SeenKnockRequests(_) => self.serialize_value(
                 &value
                     .into_seen_join_requests()
                     .expect("Session data is not a set of seen join request ids"),
```
bnjbvr, line 1041: ditto here

crates/matrix-sdk-base/src/rooms/normal.rs
```diff
@@ -1127,6 +1127,88 @@ impl Room {
...
+        if guard.is_none() {
+            ObservableWriteGuard::set(
+                &mut guard,
+                Some(self.load_latest_knock_request_ids().await?),
+            );
+        }
+        Ok(guard.clone().unwrap_or_default())
+    }
+
+    async fn load_latest_knock_request_ids(
```
bnjbvr, line 1202: Let's add a doc-comment here. Does the `latest` carry some meaningful information? Do we mean the "active" knock requests, i.e. those that have not been acted upon?

crates/matrix-sdk-base/src/rooms/normal.rs
```diff
@@ -1127,6 +1127,88 @@ impl Room {
...
+            )
+            .await?;
+
+        drop(current_seen_events_guard);
+
+        Ok(())
+    }
+
+    /// Get the list of seen knock request event ids in this room.
+    pub async fn get_seen_join_request_ids(
```
bnjbvr, line 1189: Let's rename `join` to `knock` here too.

crates/matrix-sdk-base/src/rooms/normal.rs
```diff
@@ -1127,6 +1127,88 @@ impl Room {
...
+        let mut current_seen_events_guard = self.seen_knock_request_ids_map.write().await;
+        let mut current_seen_events = if current_seen_events_guard.is_none() {
+            self.load_latest_knock_request_ids().await?
+        } else {
+            current_seen_events_guard.clone().unwrap()
+        };
+
+        for (event_id, user_id) in event_to_user_ids {
+            current_seen_events.insert(event_id, user_id);
+        }
```
bnjbvr, line 1169: Think you can do `current_seen_events.extend(event_to_user_ids)` or something like that.

crates/matrix-sdk-base/src/rooms/normal.rs
```diff
@@ -1127,6 +1127,88 @@ impl Room {
...
+                    }
+                }
+                _ => warn!(
+                    "Could not mark knock event as seen: event for user {} is not valid.",
+                    event.state_key()
+                ),
+            }
+        }
+
+        let mut current_seen_events_guard = self.seen_knock_request_ids_map.write().await;
```
bnjbvr, line 1160: Can you add a small comment that we're not calling `get_seen_join_request_ids` here because we want to keep the mutex's guard after reloading?

crates/matrix-sdk-base/src/rooms/normal.rs
```diff
@@ -1127,6 +1127,88 @@ impl Room {
...
+        );
+
+        self.store
+            .set_kv_data(
+                StateStoreDataKey::SeenKnockRequests(self.room_id()),
+                StateStoreDataValue::SeenKnockRequests(current_seen_events),
+            )
+            .await?;
+
+        drop(current_seen_events_guard);
```
bnjbvr, line 1183: No need to explicitly drop, since we're hitting the end of the function and it's going to be dropped for us soon.

crates/matrix-sdk/src/room/knock_requests.rs
```diff
@@ -176,24 +193,24 @@ mod tests {
...
+        // When we decline the knock request and ban the user from the room
+        knock_request
             .decline_and_ban(None)
             .await
             .expect("Failed to decline the request and ban the user");
     }
 
-    fn mock_join_request(room: &Room, event_id: Option<&EventId>) -> JoinRequest {
-        JoinRequest::new(
+    fn mock_knock_request(room: &Room, event_id: Option<&EventId>) -> KnockRequest {
```
bnjbvr, line 208: Can we call this `make_knock_request`, so one is not confused that this is not mocking an endpoint? I've come to read the `mock_` prefix as something for endpoints only…

crates/matrix-sdk/src/room/mod.rs
```diff
@@ -3201,103 +3201,104 @@ impl Room {
         ObservableLiveLocation::new(&self.client, self.room_id())
     }
 
-    /// Helper to requests to join this `Room`.
+    /// Helper to subscribe to knock requests in this `Room`.
```
bnjbvr, line 3204: nit: s/Helper to//

crates/matrix-sdk/src/room/mod.rs
```diff
@@ -3201,103 +3201,104 @@ impl Room {
...
+                        if let Some(event) = event.as_original() {
+                            // If we can calculate the membership change, try to emit only when needed
+                            if event.prev_content().is_some() {
+                                match event.membership_change() {
+                                    MembershipChange::Banned |
+                                    MembershipChange::Knocked |
+                                    MembershipChange::KnockAccepted |
+                                    MembershipChange::KnockDenied |
+                                    MembershipChange::KnockRetracted => {
+                                        match this.get_current_join_requests(&seen_ids).await {
```
bnjbvr, line 3255: Instead of duplicating the `match this.get_current_join_requests(,..`) (which ought to be renamed to s/join/knock/), could we have a bool `emit`, and then, `if emit { match this.get_current_join_request() { ... }}` to avoid the code duplication?

crates/matrix-sdk/src/room/mod.rs
```diff
@@ -3201,103 +3201,104 @@ impl Room {
...
                                 }
                             }
-                            _ => (),
                         }
-                    } else {
-                        // If we can't calculate the membership change, assume we need to
-                        // emit updated values
-                        match this.clone().get_current_join_requests(&seen_ids).await {
+                    }
+                    Some(new_seen_ids) = seen_request_ids_stream.next() => {
```
bnjbvr, line 3276: uber nit: please add new lines between match arm variants

crates/matrix-sdk/src/room/mod.rs
```diff
@@ -3201,103 +3201,104 @@ impl Room {
...
                         }
                     }
-                } else if has_new_seen_ids || has_missing_room_members {
-                    // If seen requests have changed or we have missing room members,
-                    // we need to recalculate all the requests to join
-                    match this.clone().get_current_join_requests(&seen_ids).await {
-                        Ok(requests) => yield requests,
-                        Err(e) => {
-                            warn!("Failed to get updated requests to join on seen ids changed: {e:?}")
+                    Some(room_info) = room_info_stream.next() => {
```
bnjbvr: here too

crates/matrix-sdk/src/room/mod.rs
```diff
@@ -3201,103 +3201,104 @@ impl Room {
...
+                        if !room_info.are_members_synced() {
+                            match this.get_current_join_requests(&seen_ids).await {
+                                Ok(requests) => yield requests,
+                                Err(err) => {
+                                    warn!("Failed to get updated knock requests on gappy (limited) sync: {err}")
+                                }
+                            }
                         }
                     }
+                    else => break,
```
bnjbvr, line 3301: What does this do? Can you add a code comment?

bnjbvr (approved):

PR #4341, opened 2024-11-27
test(ui): Fix the `test_echo` test

Hywan:
This patch fixes the `test_echo` test. It was doing the following:

* client sends an event to the server,
* servers acknowledges with the ID `$wWgymRfo7ri1uQx0NXO40vLJ`,
* client syncs and the server returns one event with ID `$7at8sd:localhost`,
* the test expects those 2 events to be the same!, which is incorrect.

The test was working because the transaction IDs are the same, but that's an abuse of the existing code (the code will change soon, another patch is coming). Whatever the code does: the connection must be based on the event ID, not the transaction ID.

---

### DIFF
--- a/crates/matrix-sdk-ui/tests/integration/timeline/echo.rs
+++ b/crates/matrix-sdk-ui/tests/integration/timeline/echo.rs
@@ -67,13 +67,12 @@ async fn test_echo() {
     );
     let (_, mut timeline_stream) = timeline.subscribe().await;
 
+    let event_id = event_id!("$ev");
+
     Mock::given(method("PUT"))
         .and(path_regex(r"^/_matrix/client/r0/rooms/.*/send/.*"))
         .and(header("authorization", "Bearer 1234"))
-        .respond_with(
-            ResponseTemplate::new(200)
-                .set_body_json(json!({ "event_id": "$wWgymRfo7ri1uQx0NXO40vLJ" })),
-        )
+        .respond_with(ResponseTemplate::new(200).set_body_json(json!({ "event_id": event_id })))
         .mount(&server)
         .await;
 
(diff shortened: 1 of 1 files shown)


crates/matrix-sdk-ui/tests/integration/timeline/echo.rs
```diff
@@ -87,15 +86,15 @@ async fn test_echo() {
...
+    assert!(item.event_id().is_none());
     let txn_id = item.transaction_id().unwrap();
 
     assert_let!(Some(VectorDiff::PushFront { value: day_divider }) = timeline_stream.next().await);
     assert!(day_divider.is_day_divider());
 
-    assert_let!(TimelineItemContent::Message(msg) = item.content());
-    assert_let!(MessageType::Text(text) = msg.msgtype());
-    assert_eq!(text.body, "Hello, World!");
-
```
Hywan, line 98: Just a code move.

crates/matrix-sdk-ui/tests/integration/timeline/echo.rs
```diff
@@ -104,13 +103,14 @@ async fn test_echo() {
...
     assert_matches!(item.send_state(), Some(EventSendState::Sent { .. }));
+    assert_eq!(item.event_id(), Some(event_id));
 
     let f = EventFactory::new();
     sync_builder.add_joined_room(
         JoinedRoomBuilder::new(room_id).add_timeline_event(
             f.text_msg("Hello, World!")
                 .sender(user_id!("@example:localhost"))
-                .event_id(event_id!("$7at8sd:localhost"))
+                .event_id(event_id)
```
Hywan, line 113: The real fix.

crates/matrix-sdk-ui/tests/integration/timeline/echo.rs
```diff
@@ -67,13 +67,12 @@ async fn test_echo() {
...
+    let event_id = event_id!("$ev");
+
     Mock::given(method("PUT"))
         .and(path_regex(r"^/_matrix/client/r0/rooms/.*/send/.*"))
         .and(header("authorization", "Bearer 1234"))
-        .respond_with(
-            ResponseTemplate::new(200)
-                .set_body_json(json!({ "event_id": "$wWgymRfo7ri1uQx0NXO40vLJ" })),
-        )
+        .respond_with(ResponseTemplate::new(200).set_body_json(json!({ "event_id": event_id })))
```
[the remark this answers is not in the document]
Hywan: There is an opportunity to rewrite the `Timeline` test suite entirely :-p. Gonna do that as part of the clean up phase of https://github.com/matrix-org/matrix-rust-sdk/issues/3280

PR #4345, opened 2024-11-28
chore: Define the lifetime of some const strings explicitly

poljar:


bnjbvr (approved):
ohh, i think that Clippy complained about this in a previous PR of mine… But if it's this happy here, that's fine!

PR #4346, opened 2024-11-28
Fix indexing issue in `ObservableMap`

gferon:
The client we're building is panicking on startups in some random cases with the following error message `Value should be present or has just been inserted, but it's missing`. There seems to be a pretty serious bug in `observable_map.rs` which I initially demonstrated with an extension of the existing unit tests in 6d6f8e4dada2ec43f89d1aa07118efb608a0a60a.

Any `.remove` operation called on a `ObservableMap` will not re-index `values` _after_ the removed position, which means any later operation on elements inserted after the previously removed key will either be fetched wrongly, or panic when using `.get_or_create`.

This PR fixes these two related bugs and adds extra test cases.

### DIFF
--- a/crates/matrix-sdk-base/src/store/observable_map.rs
+++ b/crates/matrix-sdk-base/src/store/observable_map.rs
@@ -138,10 +138,13 @@ where
         L: Hash + Eq + ?Sized,
     {
         let position = self.mapping.remove(key)?;
-        // reindex every mapped entry that is after the position we're looking to remove
-        for mposition in self.mapping.values_mut().filter(|mposition| **mposition > position) {
-            *mposition = mposition.saturating_sub(1);
+
+        // Reindex every mapped entry that is after the position we're looking to
+        // remove.
+        for mapped_pos in self.mapping.values_mut().filter(|pos| **pos > position) {
+            *mapped_pos = mapped_pos.saturating_sub(1);
         }
+
         Some(self.values.remove(position))
     }
 }
--- a/crates/matrix-sdk-base/src/store/observable_map.rs
+++ b/crates/matrix-sdk-base/src/store/observable_map.rs
@@ -195,6 +195,12 @@ mod tests {
         assert_eq!(map.get(&'a'), Some(&'E'));
         assert_eq!(map.get(&'b'), Some(&'f'));
         assert_eq!(map.get(&'c'), Some(&'G'));
+
+        // remove non-last item
+        assert_eq!(map.remove(&'b'), Some('f'));
+
+        // get_or_create item after the removed one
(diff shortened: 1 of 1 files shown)


bnjbvr:
Duh, great catch, not sure how we've missed that.

gferon:
Do you want me to come up with a fix? I'd gladly spend some time doing that.

bnjbvr:
Absolutely, thanks!

bnjbvr (commented):
Can we do a swap-remove and only update the position of the swapped value?

gferon:
@bnjbvr just gave it a shot, the extra test cases now pass.

> This is a trade-off. This implementation is simple enough for the moment, and basically does the job

It's obviously not the best possible solution in terms of complexity, but I think it fulfills the criteria defined in the code comment quoted above 😄

gferon:
> Can we do a swap-remove and only update the position of the swapped value?

Unfortunately, there's no `ObservableVec::swap_remove` - I could do it in two separate function calls though.

gferon:
It looks like I cannot even call `self.values.swap` on the `&mut self` ref because of some missing trait bound in `eyeball-im` (which seem to use [imbl](https://docs.rs/imbl/latest/imbl/vector/struct.Vector.html#method.swap) behind the scenes). I sent you a DM on Matrix, so we can communicate directly.

crates/matrix-sdk-base/src/store/observable_map.rs
```diff
@@ -138,6 +138,10 @@ where
         L: Hash + Eq + ?Sized,
     {
         let position = self.mapping.remove(key)?;
+        // reindex every mapped entry that is after the position we're looking to remove
+        for mposition in self.mapping.values_mut().filter(|mposition| **mposition > position) {
+            *mposition = mposition.saturating_sub(1);
```
bnjbvr:
```suggestion
        // Reindex every mapped entry that is after the position we're looking to remove.
        for mapped_pos in self.mapping.values_mut().filter(|pos| **pos > position) {
            *mapped_pos = mapped_pos.saturating_sub(1);
```

bnjbvr (approved):
Thanks, this is a good bandaid solution, and we can implement something using a manual swap-remove later, if needs be.

Hywan:
Thanks for the fix!

gferon:
Thanks for reviewing and merging this quickly. I feel like this fix warrants a 0.8.1 release on crates.io since the indexing issue leads to having `get_room` return to the wrong room after any call of `forget_room` (or any operation that deleted a room from the store as a side effect).

bnjbvr:
Our plan is to do a release somewhere around the end of the month (Christmas maybe), which will include this fix.

PR #4347, opened 2024-11-28
feat(event cache): connect the room event cache and persistent storage

bnjbvr:
This implements both directions for interacting between the event cache and the storage backend. I think this is fine design… as long as we don't have any other process writing into the cache, in which case we'll need to reset at different places.

## writing

The main idea in this PR is to use `RoomEventCacheState::with_events_mut(func: impl FnMut(&mut RoomEvents))` to perform any write to the underlying `RoomEvent` data structure. The function takes care of maintaining the storage, after updates have been performed.

Additionally: this behavior isn't enabled by default; callers have to call `EventCache::enable_storage()` when they start using the event cache.

I've put the `RoomEventCacheState` data structure in a private module, so that even the same module can't access its private fields, and we don't make mistakes by accidentally misusing `events`.

## reading

Reading happens by reloading the chunks initially, and prefilling the deduplicator with the reloaded events information. It's mostly boring changes at this point.

---

Part of https://github.com/matrix-org/matrix-rust-sdk/issues/3280. On top of #4362. Split and reworked from https://github.com/matrix-org/matrix-rust-sdk/pull/4308.

### DIFF
--- a/crates/matrix-sdk/src/event_cache/deduplicator.rs
+++ b/crates/matrix-sdk/src/event_cache/deduplicator.rs
@@ -18,6 +18,7 @@
 use std::{collections::BTreeSet, fmt, sync::Mutex};
 
 use growable_bloom_filter::{GrowableBloom, GrowableBloomBuilder};
+use tracing::warn;
 
 use super::room::events::{Event, RoomEvents};
 
@@ -46,16 +47,29 @@ impl Deduplicator {
     const APPROXIMATED_MAXIMUM_NUMBER_OF_EVENTS: usize = 1_000;
     const DESIRED_FALSE_POSITIVE_RATE: f64 = 0.01;
 
-    /// Create a new `Deduplicator`.
+    /// Create a new `Deduplicator` with no prior knowledge of known events.
+    #[cfg(test)]
     pub fn new() -> Self {
-        Self {
-            bloom_filter: Mutex::new(
-                GrowableBloomBuilder::new()
-                    .estimated_insertions(Self::APPROXIMATED_MAXIMUM_NUMBER_OF_EVENTS)
-                    .desired_error_ratio(Self::DESIRED_FALSE_POSITIVE_RATE)
-                    .build(),
-            ),
+        Self::with_initial_events(std::iter::empty())
+    }
+
+    /// Create a new `Deduplicator` filled with initial events.
+    ///
+    /// This won't detect duplicates in the initial events, only learn about
+    /// those events.
+    pub fn with_initial_events<'a>(events: impl Iterator<Item = &'a Event>) -> Self {
+        let mut bloom_filter = GrowableBloomBuilder::new()
+            .estimated_insertions(Self::APPROXIMATED_MAXIMUM_NUMBER_OF_EVENTS)
+            .desired_error_ratio(Self::DESIRED_FALSE_POSITIVE_RATE)
+            .build();
+        for e in events {
+            let Some(event_id) = e.event_id() else {
+                warn!("initial event in deduplicator had no event id");
+                continue;
+            };
+            bloom_filter.insert(event_id);
         }
+        Self { bloom_filter: Mutex::new(bloom_filter) }
     }
 
     /// Scan a collection of events and detect duplications.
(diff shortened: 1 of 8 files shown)


bnjbvr:
Getting back to a draft for this; I suspect a bit more work is required because the back-pagination tokens ought to belong in this state too…

crates/matrix-sdk/src/event_cache/room/events.rs
```diff
@@ -185,6 +188,12 @@ impl RoomEvents {
         self.chunks_updates_as_vectordiffs.take()
     }
 
+    /// Get a mutable reference to the [`LinkedChunk`] updates, aka
+    /// [`ObservableUpdates`].
+    pub fn updates(&mut self) -> &mut ObservableUpdates<Event, Gap> {
```
Hywan, line 207: Please reduce the visiblity of this:

```diff
- pub
+ pub(super)
```

?

crates/matrix-sdk/src/event_cache/room/mod.rs
```diff
@@ -533,29 +540,94 @@ impl RoomEventCacheInner {
...
-    /// Resets this data structure as if it were brand new.
-    pub(super) fn reset(&mut self) {
-        self.events.reset();
-        self.waited_for_initial_prev_token = false;
+    use super::events::RoomEvents;
+    use crate::event_cache::EventCacheError;
+
+    /// State for a single room's event cache.
+    ///
+    /// This contains all inner mutable state that ought to be updated at the
```
Hywan, line 551:
```diff
- mutable state
+ mutable states
```

crates/matrix-sdk/src/event_cache/deduplicator.rs
```diff
@@ -46,16 +47,29 @@ impl Deduplicator {
...
+    #[cfg(test)]
     pub fn new() -> Self {
-        Self {
-            bloom_filter: Mutex::new(
-                GrowableBloomBuilder::new()
-                    .estimated_insertions(Self::APPROXIMATED_MAXIMUM_NUMBER_OF_EVENTS)
-                    .desired_error_ratio(Self::DESIRED_FALSE_POSITIVE_RATE)
-                    .build(),
-            ),
+        Self::with_initial_events(std::iter::empty())
```
Hywan, line 53: :+1:

crates/matrix-sdk/src/event_cache/room/events.rs
```diff
@@ -52,14 +52,28 @@ impl Default for RoomEvents {
...
         let chunks_updates_as_vectordiffs = chunks
             .as_vector()
             // SAFETY: The `LinkedChunk` has been built with `new_with_update_history`, so
             // `as_vector` must return `Some(…)`.
-            .expect("`LinkedChunk` must have been constructor with `new_with_update_history`");
+            .expect("`LinkedChunk` must have been built with `new_with_update_history`");
+
+        // Let the deduplicator know about initial events.
+        let deduplicator =
+            Deduplicator::with_initial_events(chunks.items().map(|(_pos, event)| event));
```
Hywan, line 74: So clean.

crates/matrix-sdk/src/event_cache/room/mod.rs
```diff
@@ -955,6 +958,106 @@ mod tests {
...
+        let client = MockClientBuilder::new("http://localhost".to_owned())
+            .store_config(
+                StoreConfig::new("hodlor".to_owned()).event_cache_store(event_cache_store.clone()),
+            )
+            .build()
+            .await;
+
+        let event_cache = client.event_cache();
+
+        // Don't forget to subscribe and like^W enable storage!
```
Hywan, line 1027: 🙉

crates/matrix-sdk/src/event_cache/room/mod.rs
```diff
@@ -955,6 +958,106 @@ mod tests {
...
+
+        // A new update with one of these events leads to deduplication.
+        let timeline = Timeline { limited: false, prev_batch: None, events: vec![ev1] };
+        room_event_cache
+            .inner
+            .handle_joined_room_update(JoinedRoomUpdate { timeline, ..Default::default() })
+            .await
+            .unwrap();
+
+        // The stream doesn't report these changes *yet*. Use the items vector given
```
Hywan, line 1051: It's part of the PR I'm preparing I suppose, with `VectorDiff` being emitted by `EventCache`?

Hywan (approved):
It looks good. I like the new API to mutate the state. Thanks for putting everything together.

crates/matrix-sdk/src/event_cache/room/events.rs
bnjbvr: Indeed, thanks!

Issue #4189, opened 2024-10-30
UTD event was not decrypted after keys were received


richvdh:
The keys to a given event arrived a few seconds after the event, but there doesn't appear to have been any attempt to retry the decryption.

richvdh:
This is tracked on the EXA and EXI repos as https://github.com/element-hq/element-x-android/issues/3747 and https://github.com/element-hq/element-x-ios/issues/3449 respectively

richvdh:
Possibly related: https://github.com/matrix-org/matrix-rust-sdk/issues/3768

poljar:
I think #3768 is not related.

What I think is happening here is the following:

1. The user has the timeline open for a given room and an event is received in this room.
2. The notification client receives a push and receives the room key for this event.
3. Because the notification client received the room key, the timeline which uses the main client will not be notified about a new room key being received. Thus no attempt at a re-decryption will happen.
4. UTD in the timeline until you reload the timeline it or restart the app.

At least two (or now three) rageshakes were cases where the notification client received the room key.

bnjbvr:
Would it make sense to have a generalization of `Backup::room_keys_for_room_stream` that applied to *all* keys, not only those received from the backup source?

bnjbvr:
We could put it temporarily next to the crypto store cross-process lock, since it has to live in the `Client` for the same reason; maybe we could bundle this into a struct with all the fields that ought to be migrated back to the `OlmMachine` once the migration is over?

poljar:
Well the stream works this way:

1. Room keys get from various sources are prepared to be persisted into the crypto store
2. We collect the info about the room keys
3. We persist the room keys
4. We fire out the previously collected info into a stream

We could move the stream out of the `OlmMachine` but that doesn't change the fact that some parts of the room keys are received in a different process which uses a different `Client`.

Of course the easy fix for EXA is to use the stream from the `OlmMachine` since the machine doesn't get destroyed there.

stefanceriu:
The iOS workaround has been implemented in https://github.com/element-hq/element-x-ios/pull/3628

richvdh:
seems like we should close this, and maybe replace it with one that documents more clearly the situations in which the implemented workaround is insufficient?

poljar:
Sure, leftover stuff is described in https://github.com/matrix-org/matrix-rust-sdk/issues/4445.

PR #4348, opened 2024-11-28
Retry decryptions based on the room key stream of the OlmMachine

poljar:
This is a partial fix for https://github.com/matrix-org/matrix-rust-sdk/issues/4189.

The fix works if you don't use the cross-process lock, otherwise the stream gets destroyed and updates don't come in anymore.

Opening as a draft, still needs a test.

### DIFF
--- a/crates/matrix-sdk-crypto/src/machine/tests/megolm_sender_data.rs
+++ b/crates/matrix-sdk-crypto/src/machine/tests/megolm_sender_data.rs
@@ -1,18 +1,16 @@
-/*
-Copyright 2024 The Matrix.org Foundation C.I.C.
-
-Licensed under the Apache License, Version 2.0 (the "License");
-you may not use this file except in compliance with the License.
-You may obtain a copy of the License at
-
-    http://www.apache.org/licenses/LICENSE-2.0
-
-Unless required by applicable law or agreed to in writing, software
-distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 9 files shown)


richvdh:
So this fixes EXA but not EXI?

poljar:
Gah, trying to write a test for this and I just realized that I need to delay the delivery of the room key so I can have the following conditions:

1. Have the timeline open
2. Let the main client/timeline receive the message, an UTD
3. Have the notification client receive the room key just now
4. Watch the UTD resolve itself because the notification client tells the timeline that the room key has been received

I guess I'll have to write the test in complement-crypto.

bnjbvr:
> delay the delivery of the room key

Our testing suite has something like this: https://github.com/matrix-org/matrix-rust-sdk/blob/a6e1f05957524b0fc28d4d6db89ff06b9bccd334/testing/matrix-sdk-integration-testing/src/tests/sliding_sync/room.rs#L682-L729

ToDevice events are evicted from a server response on purpose, and under the control of the testing code.

poljar:
> ToDevice events are evicted from a server response on purpose, and under the control of the testing code.

Well yes, wiremock can also [delay responses](https://docs.rs/wiremock/latest/wiremock/struct.ResponseTemplate.html#method.set_delay), but then I'd have to mock a lot of things to get the E2EE and notification client working.

poljar:
Alright, backpaginating the event into the timeline worked as well.

PR #4349, opened 2024-11-29
task(crypto_tests): Re-shape tests for UtdCause

andybalaam:
Part of https://github.com/element-hq/element-meta/issues/2638

If you review commit by commit you will see explanations for what I did, but you will read more code in total.


andybalaam:
Rebased on top of CI fixes. No changes to my commits except updated their descriptions.

BillCarsonFr (approved):
Just a minor comment, if we could have const for 1111 / 5555 / 9999 to show the relation between them as it is important (5555 should be in between the 2 others)

PR #653, opened 2022-05-09
Update Dependencies ahead of release

gnunicorn:
Updates a few outdated dependencies - most notably for xhsell-0.2 (for xtask); adds Badge to the Readme for reporting,

PR #4352, opened 2024-11-29
chore(xtask): Disable `unexpected_cfgs` lint

zecakeh:
It is triggered by the `xshell::cmd!` macro.

An alternative to this is to upgrade xshell to the latest version (which was not done in https://github.com/matrix-org/matrix-rust-sdk/pull/653 because there is an API break that makes it more complicated to use), or use a different crate.

### DIFF
--- a/xtask/src/main.rs
+++ b/xtask/src/main.rs
@@ -1,3 +1,5 @@
+#![allow(unexpected_cfgs)] // Triggered by the xshell::cmd!() invocation.
+
 mod ci;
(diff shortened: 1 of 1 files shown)


poljar:
Yeah we should probably look into migrating or upgrading, eventually the old version will break in one way or the other as the ecosystem moves on.

andybalaam (approved):

zecakeh:
Closing in favor of #4358.

PR #4353, opened 2024-11-29
Allow the building of member events in the member factory and utilize them in a coupl of tests

poljar:


stefanceriu (approved):
Super nice, thank you! 👏

PR #4355, opened 2024-11-29
Simplify exists_on_server tests

andybalaam:
This is part of https://github.com/element-hq/element-meta/issues/2638

I am planning to add a `fast_exists_on_server` method which uses a locally-cached value. This change makes the tests ready so I can re-use parts of them in my tests for the new method.

### DIFF
--- a/crates/matrix-sdk/src/encryption/backups/mod.rs
+++ b/crates/matrix-sdk/src/encryption/backups/mod.rs
@@ -1010,7 +1010,7 @@ mod test {
     use serde_json::json;
     use wiremock::{
         matchers::{header, method, path},
-        Mock, MockServer, ResponseTemplate,
+        Mock, MockGuard, MockServer, ResponseTemplate,
     };
 
     use super::*;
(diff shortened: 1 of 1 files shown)


crates/matrix-sdk/src/encryption/backups/mod.rs
```diff
@@ -1282,4 +1254,60 @@ mod test {
 
         server.verify().await;
     }
+
+    async fn mock_backup_exists(server: &MockServer) -> MockGuard {
```
poljar, line 1258: These should become part of the `MatrixMockServer`.

poljar (changes requested):

crates/matrix-sdk/src/encryption/backups/mod.rs
andybalaam: Done in 0d7bbb4472

poljar (approved):
Thanks, the tests look much nicer now.

PR #4357, opened 2024-11-29
fix(room_preview): add fallback server names when possible

jmartinesp:
If no server names are provided for the room summary request and the room's server name doesn't match the current user's server name, add the room alias/id server name as a fallback value. This seems to fix room preview through federation, when there is a server name in either the room id or alias.

Also, when getting a summary for a room list item, if it's an invite one, add the server name of the inviter's user id as another possible fallback.

Signed-off-by:

### DIFF
--- a/crates/matrix-sdk/src/room_preview.rs
+++ b/crates/matrix-sdk/src/room_preview.rs
@@ -26,7 +26,7 @@ use ruma::{
     events::room::{history_visibility::HistoryVisibility, join_rules::JoinRule},
     room::RoomType,
     space::SpaceRoomJoinRule,
-    OwnedMxcUri, OwnedRoomAliasId, OwnedRoomId, OwnedServerName, RoomId, RoomOrAliasId,
+    OwnedMxcUri, OwnedRoomAliasId, OwnedRoomId, OwnedServerName, RoomId, RoomOrAliasId, ServerName,
 };
 use tokio::try_join;
 use tracing::{instrument, warn};
@@ -233,6 +233,9 @@ impl RoomPreview {
         room_or_alias_id: &RoomOrAliasId,
         via: Vec<OwnedServerName>,
     ) -> crate::Result<Self> {
+        let own_server_name = client.session_meta().map(|s| s.user_id.server_name());
+        let via = ensure_server_names_is_not_empty(own_server_name, via, room_or_alias_id);
+
         let request = ruma::api::client::room::get_summary::msc3266::Request::new(
             room_or_alias_id.to_owned(),
             via,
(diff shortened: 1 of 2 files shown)


crates/matrix-sdk/src/room_preview.rs
```diff
@@ -372,3 +375,92 @@ async fn search_for_room_preview_in_room_directory(
...
 }
+
+// Make sure the server name of the room id/alias is
+// included in the list of server names to send if no server names are provided
+fn ensure_server_names_is_not_empty(
+    own_server_name: Option<&ServerName>,
+    server_names: Vec<OwnedServerName>,
+    room_or_alias_id: &RoomOrAliasId,
+) -> Vec<OwnedServerName> {
+    let Some(own_server_name) = own_server_name else { return server_names };
```
poljar, line 386:
```suggestion
    let Some(own_server_name) = own_server_name else { return server_names };
    
```

crates/matrix-sdk/src/room_preview.rs
```diff
@@ -372,3 +375,92 @@ async fn search_for_room_preview_in_room_directory(
...
+    if server_names.is_empty() {
+        if let Some(server_name) = room_or_alias_id.server_name() {
+            if server_name != own_server_name {
+                return server_names.into_iter().chain(vec![server_name.to_owned()]).collect();
+            }
+        }
+    }
```
poljar, line 393:
```suggestion
    }
    
```

crates/matrix-sdk/src/room_preview.rs
```diff
@@ -372,3 +375,92 @@ async fn search_for_room_preview_in_room_directory(
...
+fn ensure_server_names_is_not_empty(
+    own_server_name: Option<&ServerName>,
+    server_names: Vec<OwnedServerName>,
+    room_or_alias_id: &RoomOrAliasId,
+) -> Vec<OwnedServerName> {
+    let Some(own_server_name) = own_server_name else { return server_names };
+    if server_names.is_empty() {
+        if let Some(server_name) = room_or_alias_id.server_name() {
+            if server_name != own_server_name {
+                return server_names.into_iter().chain(vec![server_name.to_owned()]).collect();
```
poljar, line 390: You could have just pushed the additional name to the vec and avoid the early return.

```suggestion
                server_names.push(server_name.to_owned());
```

poljar (approved):
Left some nits, looks good.

crates/matrix-sdk/src/room_preview.rs
jmartinesp: `server_names` is not mutable. I can make it so, but then I'd be creating a new `Vec` even when no changes are needed.

poljar: But you wouldn't, you would return the same `Vec` that was passed to the method, some times you would mutate it and sometimes you won't.

The `collect()` creates a brand new `Vec`.

jmartinesp: Oh, I completely misunderstood how this works then (you'd actually be creating a different type on Kotlin, i.e.). I'll change that, thanks!

PR #4358, opened 2024-11-29
chore(xtask): Upgrade xshell

zecakeh:
Gets rid of an `unexpected_cfgs` warning. Alternative to #4352.

Because most subcommands are functions, this uses a thread-local static shared shell API. If they were methods, we could just use a shared shell API for a given command.

### DIFF
--- a/xtask/src/release.rs
+++ b/xtask/src/release.rs
@@ -1,9 +1,9 @@
 use std::env;
 
 use clap::{Args, Subcommand, ValueEnum};
-use xshell::{cmd, pushd};
+use xshell::cmd;
 
-use crate::{workspace, Result};
+use crate::{sh, workspace, Result};
 
 #[derive(Args)]
 pub struct ReleaseArgs {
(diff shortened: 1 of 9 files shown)


xtask/src/release.rs
```diff
@@ -74,7 +74,8 @@ impl ReleaseArgs {
         //
         // More info: https://git-cliff.org/docs/usage/monorepos
         if self.cmd != ReleaseCommand::Changelog {
-            let _p = pushd(workspace::root_path()?)?;
+            let sh = sh();
+            let _p = sh.push_dir(workspace::root_path()?);
```
zecakeh, line 78: I am not sure if I understand the initial intent here. Because `pushd` / `Shell::push_dir` returns a RAII guard, it stop having an effect as soon as it is out of scope. And it is out of scope right after being created.

poljar: The intention was to not change the directory to the workspace if the `Changelog` command is used.

A `let _p;` before the if branch give the guard the correct lifetime. Could you add that please?

zecakeh: Done.

poljar (approved):
Seems to work as it worked previously, the code looks sane as well.

Thanks for taking care of this.

PR #4360, opened 2024-12-01
feat(ui): Add new properties to `UnableToDecryptInfo`

richvdh:
Adds a number of new properties to the `UnableToDecryptInfo` that is passed to the application-level `UnableToDecryptHook`.

Part of https://github.com/element-hq/element-meta/issues/2582

### DIFF
--- a/crates/matrix-sdk-ui/src/unable_to_decrypt_hook.rs
+++ b/crates/matrix-sdk-ui/src/unable_to_decrypt_hook.rs
@@ -73,6 +73,9 @@ struct PendingUtdReport {
 
     /// The task that will report this UTD to the parent hook.
     report_task: JoinHandle<()>,
+
+    /// The UnableToDecryptInfo structure for this UTD event.
+    utd_info: UnableToDecryptInfo,
 }
 
 /// A manager over an existing [`UnableToDecryptHook`] that deduplicates UTDs
@@ -227,6 +230,7 @@ impl UtdHookManager {
         let reported_utds = self.reported_utds.clone();
         let parent = self.parent.clone();
         let client = self.client.clone();
+        let owned_event_id = event_id.to_owned();
 
         // Spawn a task that will wait for the given delay, and maybe call the parent
         // hook then.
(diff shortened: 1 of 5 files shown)


crates/matrix-sdk-ui/src/unable_to_decrypt_hook.rs
```diff
@@ -63,6 +63,21 @@ pub struct UnableToDecryptInfo {
...
+
+    /// Whether the user had verified their own identity at the point they
+    /// received the UTD event.
+    pub user_trusts_own_identity: bool,
+
+    /// True if the sender of the event was on a different server to our user.
+    pub is_federated: bool,
+
+    /// True if our user is using the `matrix.org` homeserver.
+    pub is_matrix_dot_org: bool,
```
BillCarsonFr, line 80: It feels strange to have a `is_matrix_dot_org` at the `sdk` level. 
Can we have instead a sender_homerserver and a own_homeserver , then the app could do a  `equals("matrix.org")`  and the is_federated?

richvdh: I was torn on this. On the one hand, I agree: this does seem a bit arbitrary. On the other, I was keen to factor out a as much as possible of the EXA/EXI common code.

Hywan: I think @BillCarsonFr's idea is quite nice. I understand your position, but I think it's important to not take this kind of shortcuts/biases in the SDK.

Hywan (changes requested):
Seems good! One note about `is_matrix_dot_org` that I would prefer to not see here. Otherwise, great patches!

crates/matrix-sdk-ui/src/unable_to_decrypt_hook.rs
richvdh: Fair enough. PTAL.

crates/matrix-sdk-ui/src/unable_to_decrypt_hook.rs
```diff
@@ -73,11 +73,12 @@ pub struct UnableToDecryptInfo {
     /// received the UTD event.
     pub user_trusts_own_identity: bool,
 
-    /// True if the sender of the event was on a different server to our user.
-    pub is_federated: bool,
+    /// The homserver of the user that sent the undecryptable event.
```
Hywan, line 76:
```diff
- homserver
+ homeserver
```

Hywan (approved):
Thanks for the fixup! A last typo in the doc, but I'm approving the PR to avoid a new review cycle.

crates/matrix-sdk-ui/src/unable_to_decrypt_hook.rs
richvdh:
```suggestion
    /// The homeserver of the user that sent the undecryptable event.
```

PR #4361, opened 2024-12-02
refactor(ui): Use an iterator instead of `Vec` to represent events

Hywan:
(Extracted from https://github.com/matrix-org/matrix-rust-sdk/pull/3512/)

This patch changes `TimelineStateTransaction::add_remote_events_at` to take an `IntoIterator<Item = Into<SyncTimelineEvent>>` for `events`. In the current code, it saves one
`iter().map(Into::into).collect::<Vec<_>>()`, but it will save another one when we will support `VectorDiff`s coming from the `EventCache`.

It also avoids to allocate a vector to pass new events (this mostly happens in the tests, but it can happen in real life).

---


bnjbvr (approved):
Makes sense!

PR #4362, opened 2024-12-02
feat(event cache): reload a linked chunk from a store

bnjbvr:
This implements the following:

- add a new trait method to reload a linked chunk from the store. In particular, we want to distinguish reloading when there's nothing in the store (could happen if we never saved a chunk before) from reloading a non-empty chunk. The reason is, that there's an invariant in the linked chunk code that when it's empty, a linked chunk has a single empty Events chunk; we let the caller maintain that invariant by creating the linked chunk in this case.
- implement that trait for the sqlite store; it's trivial, since the sqlite store had to implement similar functionality just so as to be tested.
- implement that trait for the in-memory store; while not complicated, it's a bit unwieldy because the in-memory store keeps one line per item in a chunk, instead of storing the entire items vector. But still, relatively easy to manage :-)
- adds integration tests at the trait level for the event cache store, for handling *and* reloading a chunk at the same time. Moves a bit of code from the sqlite testing framework, to help with creating and checking synthetic events.

### DIFF
--- a/crates/matrix-sdk-base/src/event_cache/store/memory_store.rs
+++ b/crates/matrix-sdk-base/src/event_cache/store/memory_store.rs
@@ -16,13 +16,13 @@ use std::{collections::HashMap, num::NonZeroUsize, sync::RwLock as StdRwLock, ti
 
 use async_trait::async_trait;
 use matrix_sdk_common::{
-    linked_chunk::{relational::RelationalLinkedChunk, Update},
+    linked_chunk::{relational::RelationalLinkedChunk, LinkedChunk, Update},
     ring_buffer::RingBuffer,
     store_locks::memory_store_helper::try_take_leased_lock,
 };
 use ruma::{MxcUri, OwnedMxcUri, RoomId};
 
-use super::{EventCacheStore, EventCacheStoreError, Result};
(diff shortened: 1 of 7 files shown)


crates/matrix-sdk-base/src/event_cache/store/memory_store.rs
```diff
@@ -95,10 +95,24 @@ impl EventCacheStore for MemoryStore {
 
     async fn reload_linked_chunk(
         &self,
-        _room_id: &RoomId,
+        room_id: &RoomId,
     ) -> Result<Option<LinkedChunk<DEFAULT_CHUNK_CAPACITY, Event, Gap>>, Self::Error> {
-        // TODO(hywan)
```
Hywan: Good job hywan.

Hywan (approved):
Looks great to me!

bnjbvr:
@Hywan FWIW, I've added a small fix in the last commit (with a regression test). If we added an empty items chunk (which is totally fine, by contract), it would incorrectly lead to an invalid data error. Instead, when we find a chunk with no rows, we have to assume it's an empty events chunk. That's the change happening in the last commit.

Hywan:
All good to me, thanks for the bug fix!

PR #4364, opened 2024-12-02
ci: Build the Mac framework using the reldbg profile

poljar:
This brings our CI times for the `Build framework` step down to 13 minutes after they have risen to 40+ minutes because of https://github.com/matrix-org/matrix-rust-sdk/pull/4354/commits/191d00871f3b50ef90fde2fbef1c92d7036aec22.

### DIFF
--- a/.github/workflows/bindings_ci.yml
+++ b/.github/workflows/bindings_ci.yml
@@ -175,7 +175,7 @@ jobs:
         run: swift test
 
       - name: Build Framework
-        run: target/debug/xtask swift build-framework --target=aarch64-apple-ios --profile=release
+        run: target/debug/xtask swift build-framework --target=aarch64-apple-ios --profile=reldbg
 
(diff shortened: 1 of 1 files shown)


bnjbvr (approved):
That's better! We're not back to the initial situation though: the xtask for apple took 15 minutes, not 13 as initially claimed the PR's description, that I've updated to reflect reality.

Still curious if we could get back to the initial situation, but this is less dramatic at least now, thanks.

poljar:
> That's better! We're not back to the initial situation though: the xtask for apple took 15 minutes, not 13 as initially claimed the PR's description, that I've updated to reflect reality.

I was talking about the step that was modified here, the `Build framework` step:

image

vs

image

I have clarified this in the OP now.

bnjbvr:
Thanks for addressing this :heart_hands:

PR #4365, opened 2024-12-02
feat(ffi): expose a custom timeline builder that can be configured to show only media, images and videos or files and audios

stefanceriu:
Instances of these timelines will be used to power the 2 different tabs shown on the new media browser. The client will be responsible for interacting with it similar to a normal timeline and transforming its data into something renderable e.g. section by date separators (which will be made configurable in a follow up PR)

### DIFF
--- a/bindings/matrix-sdk-ffi/src/room.rs
+++ b/bindings/matrix-sdk-ffi/src/room.rs
@@ -11,7 +11,7 @@ use matrix_sdk::{
     ComposerDraft as SdkComposerDraft, ComposerDraftType as SdkComposerDraftType,
     RoomHero as SdkRoomHero, RoomMemberships, RoomState,
 };
-use matrix_sdk_ui::timeline::{PaginationError, RoomExt, TimelineFocus};
+use matrix_sdk_ui::timeline::{default_event_filter, PaginationError, RoomExt, TimelineFocus};
 use mime::Mime;
 use ruma::{
     api::client::room::report_content,
@@ -23,7 +23,7 @@ use ruma::{
             message::RoomMessageEventContentWithoutRelation,
             power_levels::RoomPowerLevels as RumaPowerLevels, MediaSource,
         },
-        TimelineEventType,
+        AnyMessageLikeEventContent, AnySyncTimelineEvent, TimelineEventType,
     },
     EventId, Int, OwnedDeviceId, OwnedUserId, RoomAliasId, UserId,
 };
@@ -34,7 +34,7 @@ use super::RUNTIME;
 use crate::{
     chunk_iterator::ChunkIterator,
     error::{ClientError, MediaInfoError, RoomError},
-    event::{MessageLikeEventType, StateEventType},
+    event::{MessageLikeEventType, RoomMessageEventMessageType, StateEventType},
     identity_status_change::IdentityStatusChange,
     room_info::RoomInfo,
     room_member::RoomMember,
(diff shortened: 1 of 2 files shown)


bindings/matrix-sdk-ffi/src/room.rs
```diff
@@ -260,6 +267,33 @@ impl Room {
...
 
+    pub async fn media_events_timeline(
+        &self,
+        filter: MediaEventsTimelineFilter,
+    ) -> Result<Arc<Timeline>, ClientError> {
+        let room = &self.inner;
+
+        let mut builder = matrix_sdk_ui::timeline::Timeline::builder(room);
+
+        builder = builder.event_filter(move |event, _| match event {
```
stefanceriu, line 278: Heavily inspired by the [default_event_filter](https://github.com/matrix-org/matrix-rust-sdk/blob/main/crates/matrix-sdk-ui/src/timeline/controller/mod.rs#L182)

bnjbvr: Any chance you could reuse `TimelineEventTypeFilter` that already exists at the FFI layer? Instead of passing it to `RoomListItem::init_timeline`, we could have a new way to create another (non-default) timeline from a filter and a prefix, maybe.

stefanceriu: Ah, good shout, didn't realise that's a thing. Let me have a play 👍

bindings/matrix-sdk-ffi/src/room.rs
```diff
@@ -52,6 +53,12 @@ pub enum Membership {
     Knocked,
 }
 
+#[derive(Debug, Clone, uniffi::Enum)]
+pub enum MediaEventsTimelineFilter {
+    ImageAndVideo,
+    FileAndAudio,
```
bnjbvr, line 59: A few remarks:

- can you add doc comments for these variants, please?
- it seems a bit arbitrary to join image with video, and file with audio. I suppose these come from ElementX requirements, but we're trying to make the SDK a bit more general (although arguably, the FFI layer is quite adhoc to EX apps). Would it make sense to split those into four variants: `Image`, `Video`, `File`, `Audio`, and accept an array of them in `media_events_timeline` instead? That way, embedders can compose these as they want to, and not have to add new variants in there.

bindings/matrix-sdk-ffi/src/room.rs
bnjbvr, line 278: Not sure it will change a thing, but it would be slightly safer to first filter with the [`default_event_filter`](https://github.com/matrix-org/matrix-rust-sdk/blob/46bdcd933a58c575f0fb9e0c8a65a90d3c14c29a/crates/matrix-sdk-ui/src/timeline/controller/mod.rs#L168).

bindings/matrix-sdk-ffi/src/room.rs
```diff
@@ -260,6 +267,33 @@ impl Room {
         Ok(Timeline::new(timeline))
     }
 
+    pub async fn media_events_timeline(
+        &self,
+        filter: MediaEventsTimelineFilter,
+    ) -> Result<Arc<Timeline>, ClientError> {
+        let room = &self.inner;
+
+        let mut builder = matrix_sdk_ui::timeline::Timeline::builder(room);
```
bnjbvr, line 276: Could we also ask for a timeline internal id prefix in this function's signature, and pass it here? This would help tracking issues in rageshakes, since the internal id prefix is used to distinguish across different timeline instances.

bnjbvr: We may want to add a note that we're not passing a UTD hook here, but that's fine because we should be flagging those in the main timeline anyways.

bindings/matrix-sdk-ffi/src/room.rs
```diff
@@ -260,6 +267,33 @@ impl Room {
         Ok(Timeline::new(timeline))
     }
 
+    pub async fn media_events_timeline(
```
bnjbvr, line 270: Can you add a rough doc comment, please?

bnjbvr (changes requested):
Sweet, a few proposals to make it slightly more generic for other consumers; I'd like to read your opinion before approving :-)

bindings/matrix-sdk-ffi/src/room.rs
stefanceriu: > can you add doc comments

sure thing 👍 

> Would it make sense to split those into four variants

Yep yep, makes sense to me, I'll make the change

bindings/matrix-sdk-ffi/src/room.rs
stefanceriu: Oh good point, I'll add it just in case 👍

stefanceriu:
Alrighty, I went with exposing all of the possible RoomMessage message types and allowing any combination of them to be passed in as a filter. I'll squash and update the description before merging if everything else is in order 👍

PR #4366, opened 2024-12-02
refactor: Rename `TimelineMetadata::all_events` to `all_remote_events`

Hywan:
This patch renames the `all_events` field to `all_remote_events` in `TimelineMetada` for the sake of clarity.


bnjbvr (approved):
Makes sense, feel free to rename `add_or_update_event` to `add_or_update_remote_event` too if you'd like (and don't think the name's too long :upside_down_face:)

PR #4368, opened 2024-12-03
refactor: replace static with const for global constants

Integral-Tech:
- For global variables which don't need to be mutable, prefer const to static.

Signed-off-by: Integral <integral@member.fsf.org>


bnjbvr (approved):
Thanks! (For observers: in general consts are better because they can be "inlined" by the optimizer at their use points, instead of occupying space in static memory and needing dereferences. Whenever a static reference is useful, static ought to be preferred, but const is nicer in many cases.)

PR #4369, opened 2024-12-03
refactor: `TimelineStateTransaction::add_or_update_remote_event` always return `true`

Hywan:
The `add_or_update_remote_event` method always returns `true`. This
patch updates the method to return nothing, and cleans up the call sites
accordingly. This patch also adds comments to clarify the code flow.

The bool value returned by `add_or_update_remote_event` was supposed
to be `false` if the event was duplicated. First off, as soon as the
`Timeline` receives its events from the `EventCache` via `VectorDiff`,
the `event_cache::Deduplicator` will take care of deduplication,
so the `Timeline` won't have to handle that itself. Second,
`add_or_update_remote_event` was sometimes removing an event, but it
was re-inserting a new one immediately without returning `false`: it was
never returning `false` because a new event was always added.

### DIFF
--- a/crates/matrix-sdk-ui/src/timeline/controller/state.rs
+++ b/crates/matrix-sdk-ui/src/timeline/controller/state.rs
@@ -454,6 +454,7 @@ impl TimelineStateTransaction<'_> {
 
         let (event_id, sender, timestamp, txn_id, event_kind, should_add) = match raw.deserialize()
         {
+            // Classical path: the event is valid, can be deserialized, everything is alright.
             Ok(event) => {
                 let event_id = event.event_id().to_owned();
                 let room_version = room_data_provider.room_version();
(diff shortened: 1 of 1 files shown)


crates/matrix-sdk-ui/src/timeline/controller/state.rs
```diff
@@ -536,24 +542,30 @@ impl TimelineStateTransaction<'_> {
...
-                    let _event_added_or_updated = self
-                        .add_or_update_remote_event(
-                            event_meta,
-                            position,
-                            room_data_provider,
-                            settings,
-                        )
-                        .await;
+
+                    // Remember the event before returning prematurely.
```
poljar, line 546: Perhaps telling us why we're remembering things would be useful as well, not particularly needed since you can follow the call tree of `add_or_update_remote_event()` yourself, but would avoid a couple of raised eyebrows.

poljar (approved):
Looks good, left a small suggestion.

crates/matrix-sdk-ui/src/timeline/controller/state.rs
Hywan: Good idea. I'll improve the doc.

PR #4372, opened 2024-12-03
test(event cache): rely less on `EventCache::add_initial_events()`

bnjbvr:
`EventCache::add_initial_events()` was supposed to be temporary, until the event cache has its own storage. It's been used in multiple tests, as a way to prefill the event cache. It's used in other place, when constructing a timeline, to prefill the room event cache with the (soon-to-be-deprecated) cached timeline we stored in the sliding sync storage (aka `timeline_queue()`).

This patch gets rids of all uses in tests, so that there remains only a single use (when creating the timeline). We can later remove this function, once the storage has been enabled unconditionally.

Part of https://github.com/matrix-org/matrix-rust-sdk/issues/3280
Split from #4308.

### DIFF
--- a/crates/matrix-sdk/tests/integration/event_cache.rs
+++ b/crates/matrix-sdk/tests/integration/event_cache.rs
@@ -60,7 +60,7 @@ async fn test_must_explicitly_subscribe() {
 }
 
 #[async_test]
-async fn test_add_initial_events() {
+async fn test_event_cache_receives_events() {
     let (client, server) = logged_in_client_with_server().await;
 
     // Immediately subscribe the event cache to sync updates.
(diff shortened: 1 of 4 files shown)


crates/matrix-sdk/tests/integration/event_cache.rs
```diff
@@ -146,46 +149,63 @@ async fn test_add_initial_events() {
...
-                f.text_msg("hey there").sender(dexter).into_sync(),
-                f.text_msg("hoy!").sender(ivan).into_sync(),
+                Update::NewItemsChunk { previous: None, new: ChunkIdentifier::new(0), next: None },
+                Update::PushItems {
+                    at: Position::new(ChunkIdentifier::new(0), 0),
+                    items: vec![
+                        f.text_msg("hey there").sender(dexter).into_sync(),
+                        f.text_msg("hoy!").sender(ivan).into_sync(),
+                    ],
+                },
```
Hywan, line 159: I wonder if we shouldn't use a builder for that 🤔, thoughts?

Hywan (approved):

crates/matrix-sdk/tests/integration/event_cache.rs
bnjbvr: I think we could even avoid that by using a sync, and let the event cache store the events in memory.

PR #4373, opened 2024-12-03
feat(event cache): strip bundled relations before persisting events

bnjbvr:
A bit of serde foo to remove the `unsigned`.`m_relations` of an event, if it contained any, before storing the persisted event to disk. Otherwise, we might end up duplicating events with different contents in the database, and it creates complications for consumers. For instance, the consumer of a reloaded event wouldn't know how to order a bundled edit with respect to other edits (i.e. figure which edit is the most recent one), since there's no generalized notion of ordering across events.

Part of https://github.com/matrix-org/matrix-rust-sdk/issues/3280.

### DIFF
--- a/crates/matrix-sdk/src/event_cache/room/mod.rs
+++ b/crates/matrix-sdk/src/event_cache/room/mod.rs
@@ -539,9 +539,13 @@ impl RoomEventCacheInner {
 mod private {
     use std::sync::Arc;
 
-    use matrix_sdk_base::event_cache::store::EventCacheStoreLock;
+    use matrix_sdk_base::{
+        deserialized_responses::{SyncTimelineEvent, TimelineEventKind},
+        event_cache::store::EventCacheStoreLock,
+        linked_chunk::Update,
+    };
     use once_cell::sync::OnceCell;
-    use ruma::OwnedRoomId;
+    use ruma::{serde::Raw, OwnedRoomId};
 
     use super::events::RoomEvents;
     use crate::event_cache::EventCacheError;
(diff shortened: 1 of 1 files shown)


crates/matrix-sdk/src/event_cache/room/mod.rs
```diff
@@ -587,13 +591,75 @@ mod private {
             Ok(Self { room, store, events, waited_for_initial_prev_token: false })
         }
 
+        /// Removes the bundled relations from an event, if they were present.
+        ///
+        /// Only replaces the present if it contained bundled relations.
+        ///
+        /// Returns an `Option` so as to be able to short-circuit with `?`, but
+        /// the result is meaningless and mustn't be used.
```
andybalaam, line 599: I think I'd prefer to avoid including this `Option` in the type signature, perhaps by using an inner function here, but I'm not going to insist on it.

If you leave it as-is, please can you explicitly say that it always returns `None`? That would help me understand its (lack of) meaning.

crates/matrix-sdk/src/event_cache/room/mod.rs
```diff
@@ -587,13 +591,75 @@ mod private {
...
+        fn maybe_strip_relations<T>(event: &mut Raw<T>) -> Option<()> {
```
andybalaam, line 600: Mild preference for naming this `strip_relations_if_present`.

crates/matrix-sdk/src/event_cache/room/mod.rs
```diff
@@ -587,13 +591,75 @@ mod private {
...
         async fn propagate_changes(&mut self) -> Result<(), EventCacheError> {
             let updates = self.events.updates().take();
 
             if !updates.is_empty() {
                 if let Some(store) = self.store.get() {
                     let locked = store.lock().await?;
+
+                    // Strip relations from the `PushItems` updates.
+                    let updates = updates
+                        .into_iter()
```
andybalaam, line 643: Curious: did you try this with `iter_mut` to modify this in-place? Or do you prefer this pattern?

andybalaam (approved):
Looks good; just a couple of suggestions and questions.

crates/matrix-sdk/src/event_cache/room/mod.rs
bnjbvr: Good catch, thanks!

PR #4374, opened 2024-12-03
feat(event cache): don't clear a timeline if storage's enabled

bnjbvr:
This is one of the last steps for integrating the persistent storage: if a timeline is gappy, then we *don't* need to reset it, either internally or for observers. That's the whole point of persisting the event cache, after all: we should be able to paginate again from any holes we encounter later.

Part of https://github.com/matrix-org/matrix-rust-sdk/issues/3280


stefanceriu (approved):
Nice, makes sense to me and all those test tweaks are the cherry on top 👌

PR #4370, opened 2024-12-03
refactor(ui): Introduce the `AllRemoteEvents` type

Hywan:
<figure role="presentation">
<p align="center"></p>
<figcaption><p align="center">A shepherdess gathering a herd of sheep.</p></figcaption>
</figure>

---

Build on top of https://github.com/matrix-org/matrix-rust-sdk/pull/4369.

The first patch replaces `VecDeque<EventMeta>` by `AllRemoteEvents` which
is a wrapper type around `VecDeque<EventMeta>`, but this new type aims
at adding semantics API rather than a generic API. It also helps to
isolate the use of these values and to know precisely when and how they
are used.

As a first step, `AllRemoteEvents` implements a generic API to not break
the existing code. Next patches are revisiting that a little bit step
by step:

* `AllRemoteEvents::back` becomes `last` to add semantics.
* `AllRemoteEvents::iter_mut` are removed and `get_by_event_id_mut`
  is created.

This PR should be reviewed patch-by-patch.

---

* Address https://github.com/matrix-org/matrix-rust-sdk/issues/3280

PR #4377, opened 2024-12-04
refactor(ui): Create a mapping between remote events to timeline items

Hywan:
<figure role="presentation">
<p align="center"></p>
<figcaption><p align="center">A messenger trying to find its path in an unknown land</p></figcaption>
</figure>

---

Don't be fooled by the number of new lines in this PR, patches are small and relatively easy to read. Most of the new lines are documentation or tests (around 1300).

The idea of this PR is to create a mapping between events to timeline items. More precisely, a mapping between event indexes and timeline item indexes. Why is this needed? We want `Timeline` to receive its updates via `VectorDiff`. Those `VectorDiff` are generated by the `EventCache` (via `linked_chunk::AsVector`). The indexes from these `VectorDiff` —like `VectorDiff::Insert { index, .. }` or `VectorDiff::Remove { index }`— are _event index_. The `Timeline` manipulates a collection of timeline items, where each item has an event ID, but no event index.

Fortunately for us, a recent patch has clarified an internal data type, now known as `AllRemoteEvents` (see https://github.com/matrix-org/matrix-rust-sdk/pull/4370). We can re-use this type to maintain an index between the events and the timeline items.

This PR must be reviewed patch-by-patch:

- The first 3 patches are adding the `EventMeta::timeline_item_index` field, and maintains it. That's it, everything works at this step!
- The other patches are a refactoring to make this mechanism more robust:
  - It introduces `ObservableItems` which is a new type that replaces/wraps `ObservableVector<Arc<TimelineItem>>`: the idea is to provide a small API we control!
  - Then, `AllRemoteEvents` is moved inside `ObservableItems`, so that operations on the items and the events happen “atomically”, which removes many possible errors of misuses.
- Finally, the last patches add documentations and tests.

---

* Follow up https://github.com/matrix-org/matrix-rust-sdk/pull/3512 but these patches have taken a very different approach with new types like `AllRemoteEvents` and `ObservableItems` types.

### DIFF
--- a/crates/matrix-sdk-ui/src/timeline/controller/observable_items.rs
+++ b/crates/matrix-sdk-ui/src/timeline/controller/observable_items.rs
@@ -95,7 +95,7 @@ impl ObservableItems {
     /// # Panics
     ///
     /// Panics if `timeline_item_index > total_number_of_timeline_items`.
-    pub fn set(
+    pub fn replace(
         &mut self,
         timeline_item_index: usize,
         timeline_item: Arc<TimelineItem>,
@@ -161,7 +161,7 @@ impl<'observable_items> ObservableItemsTransaction<'observable_items> {
         self.all_remote_events.get_by_event_id_mut(event_id)
     }
 
-    pub fn set(
+    pub fn replace(
         &mut self,
         timeline_item_index: usize,
         timeline_item: Arc<TimelineItem>,
--- a/crates/matrix-sdk-ui/src/timeline/controller/observable_items.rs
+++ b/crates/matrix-sdk-ui/src/timeline/controller/observable_items.rs
@@ -104,17 +104,17 @@ impl ObservableItems {
     }
 
     /// Get an iterator over all the entries in this `ObservableItems`.
-    pub fn entries(&mut self) -> ObservableVectorEntries<'_, Arc<TimelineItem>> {
-        self.items.entries()
+    pub fn entries(&mut self) -> ObservableItemsEntries<'_> {
+        ObservableItemsEntries(self.items.entries())
     }
 
     /// Call the given closure for every element in this `ObservableItems`,
     /// with an entry struct that allows updating or removing that element.
-    pub fn for_each<F>(&mut self, f: F)
+    pub fn for_each<F>(&mut self, mut f: F)
     where
-        F: FnMut(ObservableVectorEntry<'_, Arc<TimelineItem>>),
+        F: FnMut(ObservableItemsEntry<'_>),
     {
-        self.items.for_each(f)
+        self.items.for_each(|entry| f(ObservableItemsEntry(entry)))
     }
 }
 
@@ -124,6 +124,35 @@ impl Deref for ObservableItems {
 
     fn deref(&self) -> &Self::Target {
         &self.items
+    }
+}
+
+/// An “iterator“ that yields entries into an `ObservableItems`.
+pub struct ObservableItemsEntries<'a>(ObservableVectorEntries<'a, Arc<TimelineItem>>);
+
+impl<'a> ObservableItemsEntries<'a> {
+    /// Advance this iterator, yielding an `ObservableItemsEntry` for the next
+    /// item in the timeline, or `None` if all items have been visited.
+    pub fn next(&mut self) -> Option<ObservableItemsEntry<'_>> {
+        self.0.next().map(ObservableItemsEntry)
+    }
+}
+
+/// A handle to a single timeline item in an `ObservableItems`.
+pub struct ObservableItemsEntry<'a>(ObservableVectorEntry<'a, Arc<TimelineItem>>);
+
+impl<'a> ObservableItemsEntry<'a> {
+    /// Replace the timeline item by `timeline_item`.
+    pub fn replace(this: &mut Self, timeline_item: Arc<TimelineItem>) -> Arc<TimelineItem> {
+        ObservableVectorEntry::set(&mut this.0, timeline_item)
+    }
+}
+
+impl Deref for ObservableItemsEntry<'_> {
+    type Target = Arc<TimelineItem>;
+
+    fn deref(&self) -> &Self::Target {
+        &self.0
     }
 }
 
--- a/crates/matrix-sdk-ui/src/timeline/controller/observable_items.rs
+++ b/crates/matrix-sdk-ui/src/timeline/controller/observable_items.rs
@@ -35,7 +35,11 @@ use super::{state::EventMeta, TimelineItem};
 pub struct ObservableItems {
     /// All timeline items.
     ///
-    /// Yeah, there are here!
+    /// Yeah, there are here! This [`ObservableVector`] contains all the
+    /// timeline items that are rendered in your magnificent Matrix client.
+    ///
+    /// These items are the _core_ of the timeline, see [`TimelineItem`] to
+    /// learn more.
     items: ObservableVector<Arc<TimelineItem>>,
 
     /// List of all the remote events as received in the timeline, even the ones
--- a/crates/matrix-sdk-ui/src/timeline/controller/observable_items.rs
+++ b/crates/matrix-sdk-ui/src/timeline/controller/observable_items.rs
@@ -1,0 +1,142 @@
+// Copyright 2024 The Matrix.org Foundation C.I.C.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use std::{ops::Deref, sync::Arc};
+
+use eyeball_im::{
+    ObservableVector, ObservableVectorEntries, ObservableVectorEntry, ObservableVectorTransaction,
+    ObservableVectorTransactionEntry, VectorSubscriber,
+};
+use imbl::Vector;
+
+use super::TimelineItem;
+
+#[derive(Debug)]
+pub struct ObservableItems {
+    items: ObservableVector<Arc<TimelineItem>>,
+}
+
+impl ObservableItems {
+    pub fn new() -> Self {
+        Self {
+            // Upstream default capacity is currently 16, which is making
+            // sliding-sync tests with 20 events lag. This should still be
+            // small enough.
+            items: ObservableVector::with_capacity(32),
+        }
+    }
+
+    pub fn is_empty(&self) -> bool {
+        self.items.is_empty()
+    }
+
+    pub fn subscribe(&self) -> VectorSubscriber<Arc<TimelineItem>> {
+        self.items.subscribe()
+    }
+
+    pub fn clone(&self) -> Vector<Arc<TimelineItem>> {
+        self.items.clone()
+    }
+
+    pub fn transaction(&mut self) -> ObservableItemsTransaction<'_> {
+        ObservableItemsTransaction { items: self.items.transaction() }
+    }
+
+    pub fn set(
+        &mut self,
+        timeline_item_index: usize,
+        timeline_item: Arc<TimelineItem>,
+    ) -> Arc<TimelineItem> {
+        self.items.set(timeline_item_index, timeline_item)
+    }
+
+    pub fn entries(&mut self) -> ObservableVectorEntries<'_, Arc<TimelineItem>> {
+        self.items.entries()
+    }
+
+    pub fn for_each<F>(&mut self, f: F)
+    where
+        F: FnMut(ObservableVectorEntry<'_, Arc<TimelineItem>>),
+    {
+        self.items.for_each(f)
+    }
+}
+
+// It's fine to deref to an immutable reference to `Vector`.
+impl Deref for ObservableItems {
+    type Target = Vector<Arc<TimelineItem>>;
+
+    fn deref(&self) -> &Self::Target {
+        &self.items
+    }
+}
+
+#[derive(Debug)]
+pub struct ObservableItemsTransaction<'observable_items> {
+    items: ObservableVectorTransaction<'observable_items, Arc<TimelineItem>>,
+}
+
+impl<'observable_items> ObservableItemsTransaction<'observable_items> {
+    pub fn get(&self, timeline_item_index: usize) -> Option<&Arc<TimelineItem>> {
+        self.items.get(timeline_item_index)
+    }
+
+    pub fn set(
+        &mut self,
+        timeline_item_index: usize,
+        timeline_item: Arc<TimelineItem>,
+    ) -> Arc<TimelineItem> {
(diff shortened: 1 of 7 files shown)


Hywan:
I think there is a problem when replacing a timeline item. I need to investigate that. Moving back the PR as a draft.

crates/matrix-sdk-ui/src/timeline/controller/state.rs
```diff
@@ -1208,6 +1212,70 @@ impl FullEventMeta<'_> {
...
+    /// | 4     | content of `$ev5` |                      |
+    ///
+    /// Note the day divider that is a virtual item. Also note ``$ev4`` which is
+    /// a reaction to `$ev2`. Finally note that `$ev1` is not rendered in
+    /// the timeline.
+    ///
+    /// The mapping between remove event index to timeline item index will look
+    /// like this:
+    ///
+    /// | remove event index | timeline item index | comment                                    |
```
Hywan, line 1222:
```diff
- remove event index
+ remote event index
```

crates/matrix-sdk-ui/src/timeline/controller/state.rs
```diff
@@ -1208,6 +1212,70 @@ impl FullEventMeta<'_> {
...
+    /// | 1     | content of `$ev2` | reaction with `$ev4` |
+    /// | 2     | day divider       |                      |
+    /// | 3     | content of `$ev3` |                      |
+    /// | 4     | content of `$ev5` |                      |
+    ///
+    /// Note the day divider that is a virtual item. Also note ``$ev4`` which is
+    /// a reaction to `$ev2`. Finally note that `$ev1` is not rendered in
+    /// the timeline.
+    ///
+    /// The mapping between remove event index to timeline item index will look
```
Hywan, line 1219:
```diff
- remove event
+ remote event
```

crates/matrix-sdk-ui/src/timeline/controller/state.rs
```diff
@@ -1208,6 +1212,70 @@ impl FullEventMeta<'_> {
...
+    /// | 0     | `$ev0`        |
+    /// | 1     | `$ev1`        |
+    /// | 2     | `$ev2`        |
+    /// | 3     | `$ev3`        |
+    /// | 4     | `$ev4`        |
+    /// | 5     | `$ev5`        |
+    ///
+    /// Once rendered in a timeline, it for example produces:
+    ///
+    /// | index | item              | aside items          |
```
poljar, line 1207: That's an odd use of the word `aside`, perhaps you meant `related` items, that would also align with the spec lingo: https://spec.matrix.org/unstable/client-server-api/#forming-relationships-between-events.

crates/matrix-sdk-ui/src/timeline/controller/observable_items.rs
```diff
@@ -0,0 +1,1500 @@
...
+// It's fine to deref to an immutable reference to `Vector`.
+impl Deref for ObservableItems {
+    type Target = Vector<Arc<TimelineItem>>;
+
+    fn deref(&self) -> &Self::Target {
+        &self.items
+    }
+}
+
+/// An “iterator“ that yields entries into an `ObservableItems`.
```
poljar, line 130: Could link to the `Iterator` docs.

crates/matrix-sdk-ui/src/timeline/controller/observable_items.rs
```diff
@@ -0,0 +1,1500 @@
...
+pub struct ObservableItemsEntries<'a>(ObservableVectorEntries<'a, Arc<TimelineItem>>);
+
+impl ObservableItemsEntries<'_> {
+    /// Advance this iterator, yielding an `ObservableItemsEntry` for the next
+    /// item in the timeline, or `None` if all items have been visited.
+    pub fn next(&mut self) -> Option<ObservableItemsEntry<'_>> {
```
poljar, line 136: Wait, why isn't this an `Iterator` implementation?

crates/matrix-sdk-ui/src/timeline/controller/observable_items.rs
```diff
@@ -0,0 +1,1500 @@
...
+    fn deref(&self) -> &Self::Target {
+        &self.0
+    }
+}
+
+/// A transaction that allows making multiple updates to an `ObservableItems` as
+/// an atomic unit.
+///
+/// For updates from the transaction to have affect, it has to be finalized with
+/// [`Self::commit`]. If the transaction is dropped without that method being
```
poljar, line 164: Will you ever get rid of this habit?

```suggestion
/// [`ObservableItemsTransaction::commit`]. If the transaction is dropped without that method being
```

crates/matrix-sdk-ui/src/timeline/controller/observable_items.rs
```diff
@@ -156,40 +156,60 @@ impl Deref for ObservableItemsEntry<'_> {
...
+/// called, the updates will be discarded.
 #[derive(Debug)]
 pub struct ObservableItemsTransaction<'observable_items> {
     items: ObservableVectorTransaction<'observable_items, Arc<TimelineItem>>,
     all_remote_events: &'observable_items mut AllRemoteEvents,
 }
 
 impl<'observable_items> ObservableItemsTransaction<'observable_items> {
+    /// Get a referene to the timeline index at position `timeline_item_index`.
```
poljar: I think you meant [`TimelineItem`] instead of timeline index?

crates/matrix-sdk-ui/src/timeline/controller/observable_items.rs
```diff
@@ -0,0 +1,1500 @@
...
+    pub fn get(&self, timeline_item_index: usize) -> Option<&Arc<TimelineItem>> {
+        self.items.get(timeline_item_index)
+    }
+
+    /// Get a reference to all remote events.
+    pub fn all_remote_events(&self) -> &AllRemoteEvents {
+        self.all_remote_events
+    }
+
+    /// Remove a remote event at position `event_index`.
```
poljar, line 183:
```suggestion
    /// Remove a remote event at the `event_index` position.
```

crates/matrix-sdk-ui/src/timeline/controller/observable_items.rs
```diff
@@ -0,0 +1,1500 @@
...
+    ///
+    /// Not to be confused with removing a timeline item!
+    pub fn remove_remote_event(&mut self, event_index: usize) -> Option<EventMeta> {
+        self.all_remote_events.remove(event_index)
+    }
+
+    /// Push a new remote event at the front of all remote events.
+    ///
+    /// Not to be confused with pushing front a timeline item!
```
poljar, line 192:
```suggestion
    /// Not to be confused with pushing a timeline item to the front!
```

crates/matrix-sdk-ui/src/timeline/controller/observable_items.rs
```diff
@@ -169,7 +169,7 @@ pub struct ObservableItemsTransaction<'observable_items> {
 }
 
 impl<'observable_items> ObservableItemsTransaction<'observable_items> {
-    /// Get a referene to the timeline index at position `timeline_item_index`.
+    /// Get a reference to the timeline index at position `timeline_item_index`.
```
poljar, line 173: Did `typos` notice that? If not, would be good to add it here: https://github.com/crate-ci/typos/issues/1156.

crates/matrix-sdk-ui/src/timeline/controller/observable_items.rs
```diff
@@ -478,3 +478,286 @@ impl AllRemoteEvents {
...
+
+        assert_eq!(events.iter().count(), 3);
+
+        events.clear();
+
+        assert_eq!(events.iter().count(), 0);
+    }
+
+    #[test]
+    fn test_push_front() {
```
poljar, line 1261: Not going to require it, but something to keep in mind, these container types and insertions/removals sound like they would be good candidates for property based tests.

poljar (changes requested):
I think all of this makes sense, a bit challenging to review since some aspects of my review were already addressed in subsequent comments.

I mostly left a bunch of nits here and there. The `unsafe` function makes me uneasy so would be good to rethink if we could express this a bit differently.

crates/matrix-sdk-ui/src/timeline/controller/state.rs
Hywan: Indeed. Renamed. Thanks for the link to the spec!

crates/matrix-sdk-ui/src/timeline/controller/observable_items.rs
```diff
@@ -12,19 +12,36 @@
...
+use ruma::EventId;
 
-use super::TimelineItem;
+use super::{state::EventMeta, TimelineItem};
 
 #[derive(Debug)]
 pub struct ObservableItems {
+    /// All timeline items.
+    ///
+    /// Yeah, there are here!
```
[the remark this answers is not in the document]
Hywan: Done.

crates/matrix-sdk-ui/src/timeline/controller/observable_items.rs
```diff
@@ -12,19 +12,36 @@
...
     items: ObservableVector<Arc<TimelineItem>>,
+
+    /// List of all the remote events as received in the timeline, even the ones
+    /// that are discarded in the timeline items.
+    ///
+    /// This is useful to get this for the moment as it helps the `Timeline` to
```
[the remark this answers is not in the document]
Hywan: Rephrased a liiiitle bit, but used your version. Thanks.

crates/matrix-sdk-ui/src/timeline/controller/observable_items.rs
```diff
@@ -55,29 +59,42 @@ impl ObservableItems {
...
 
+    /// Subscribe to timeline item updates.
     pub fn subscribe(&self) -> VectorSubscriber<Arc<TimelineItem>> {
         self.items.subscribe()
     }
 
+    /// Get a clone of all timeline items.
+    ///
+    /// Note that it doesn't clone `Self`, only the inner timeline items.
     pub fn clone(&self) -> Vector<Arc<TimelineItem>> {
```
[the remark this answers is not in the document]
Hywan: Looks fine to me. I initially wanted to mimic the previous API. I however agree with the renaming, it clarifies things.

crates/matrix-sdk-ui/src/timeline/controller/observable_items.rs
Hywan: I've explained why it doesn't implement `Iterator`, and I've added a link to `Iterator`.

crates/matrix-sdk-ui/src/timeline/controller/observable_items.rs
Hywan: As the doc now says:

```rust
/// An iterator that yields entries into an `ObservableItems`.
///
/// It doesn't implement [`Iterator`] though because of a lifetime conflict: the
/// returned `Iterator::Item` could live longer than the `Iterator` itself.
/// Ideally, `Iterator::next` should take a `&'a mut self`, but this is not
/// possible.
```

I think that's the same reason why [`eyeball_im::ObservableVectorEntries`](https://docs.rs/eyeball-im/latest/eyeball_im/struct.ObservableVectorEntries.html).

crates/matrix-sdk-ui/src/timeline/controller/observable_items.rs
Hywan: > Will you ever get rid of this habit?

Nah!

crates/matrix-sdk-ui/src/timeline/controller/observable_items.rs
Hywan: Indeed. Fixed, thanks!

crates/matrix-sdk-ui/src/timeline/controller/observable_items.rs
Hywan: Fixed, and it's been detected by `typos` already.

crates/matrix-sdk-ui/src/timeline/controller/observable_items.rs
Hywan: Yes, good idea. Let's address that in a future PR. This one is already large enough.

crates/matrix-sdk-ui/src/timeline/controller/state.rs
```diff
@@ -1208,6 +1212,70 @@ impl FullEventMeta<'_> {
...
+    /// The mapping between remove event index to timeline item index will look
+    /// like this:
+    ///
+    /// | remove event index | timeline item index | comment                                    |
+    /// +--------------------+---------------------+--------------------------------------------+
+    /// | 0                  | `Some(0)`           | `$ev0` is rendered as the #0 timeline item |
+    /// | 1                  | `None`              | `$ev1` isn't rendered in the timeline      |
+    /// | 2                  | `Some(1)`           | `$ev2` is rendered as the #1 timeline item |
+    /// | 3                  | `Some(3)`           | `$ev3` is rendered as the #3 timeline item |
+    /// | 4                  | `None`              | `$ev4` is a reaction to item #1            |
```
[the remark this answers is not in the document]
Hywan: We only care if a mapping exists. The nature of the mapping isn't interesting here, and it can be quite tricky to maintain the nature of this mapping too… Also, we won't do anything relevant with _other kinds_ apart from `Index` in your proposal. Since only one variant is useful, I _think_ keeping `Option` is fine here. Let's see in the future if I'm right or not. It's still possible to refactor this. Thoughts?

crates/matrix-sdk-ui/src/timeline/event_handler.rs
```diff
@@ -1189,6 +1187,13 @@ impl<'a, 'o> TimelineEventHandler<'a, 'o> {
                     })
                     .unwrap_or(0);
 
+                let event_index =
+                    Some(self.items.all_remote_events().last_index()
+                        // The last remote event is necessarily associated to this
+                        // timeline item, see the contract of this method.
+                        .expect("A timeline item is being added but its associated remote event is missing")
```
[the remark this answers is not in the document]
poljar: Yeah, that sounds preferable to me.

crates/matrix-sdk-ui/src/timeline/controller/observable_items.rs
```diff
@@ -127,6 +127,35 @@ impl Deref for ObservableItems {
...
+pub struct ObservableItemsEntry<'a>(ObservableVectorEntry<'a, Arc<TimelineItem>>);
+
+impl<'a> ObservableItemsEntry<'a> {
+    /// Replace the timeline item by `timeline_item`.
+    pub fn replace(this: &mut Self, timeline_item: Arc<TimelineItem>) -> Arc<TimelineItem> {
+        ObservableVectorEntry::set(&mut this.0, timeline_item)
+    }
+}
+
+impl Deref for ObservableItemsEntry<'_> {
```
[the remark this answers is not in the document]
poljar: Ah right, `DerefMut` isn't implemented.

crates/matrix-sdk-ui/src/timeline/event_handler.rs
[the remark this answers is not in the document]
Hywan: Okay, then it's already done.

PR #4378, opened 2024-12-04
Revert "chore(ui): Unify the logic for timeline item insertions"

poljar:
This seems to cause incorrect ordering under some more complex insertion scenarios which are not fully understood yet.

This is the likely culprit for: https://github.com/element-hq/element-x-ios/issues/3583.

A second commit, containing doc updates, was reverted as well as the culprit wouldn't revert cleanly on its own.

### DIFF
--- a/crates/matrix-sdk-ui/src/timeline/event_handler.rs
+++ b/crates/matrix-sdk-ui/src/timeline/event_handler.rs
@@ -1158,6 +1158,17 @@ impl<'a, 'o> TimelineEventHandler<'a, 'o> {
                     None => self.meta.new_timeline_item(item),
                 };
 
+                trace!("Adding new remote timeline item after all non-local events");
+
+                // We are about to insert the `new_item`, great! Though, we try to keep
+                // precise insertion semantics here, in this exact order:
+                //
+                // * _push back_ when the new item is inserted after all items,
+                // * _push front_ when the new item is inserted at index 0,
+                // * _insert_ otherwise.
+                //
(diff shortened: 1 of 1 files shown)


stefanceriu (approved):
Tested it on Element X iOS where we could originally easily reproduce the issue and I can't reproduce it anymore.

crates/matrix-sdk-ui/src/timeline/event_handler.rs
```diff
@@ -1085,71 +1085,88 @@ impl<'a, 'o> TimelineEventHandler<'a, 'o> {
...
-                txn_id,
-                event_id,
-                ..
+
+            Flow::Remote {
+                position: TimelineItemPosition::End { .. }, txn_id, event_id, ..
             } => {
-                // This block tries to find duplicated events.
-
-                let removed_event_item_id = {
```
Hywan, line 1102: Can you run this block only if `position` is `TimelineItemPosition::End` please?

poljar: What do you mean? Are you trying to find a fix for https://github.com/matrix-org/matrix-rust-sdk/pull/4378/commits/3b28809c354ca27cff95875cb0d420d2914616e7?

poljar: Status update, yes he was trying to find a fix, but people want to do a release so we'll merge the revert and do the proper fix afterwards with a regression test.

Hywan (approved):

PR #4381, opened 2024-12-06
fix(event cache): don't fill initial items if the room already had events

bnjbvr:
The test requires subtle conditions to trigger:

- initialize a timeline from a room-list-service's room
- start a backpagination with that timeline (so the room event cache's paginator is busy)
- try to initialize another timeline with the same room-list-service's room (e.g. because the first room has been closed, and the app using it doesn't have a room cache)

This would fail, because initializing a timeline calls `EventCache::add_initial_events()` all the time, which tries to reset the paginator's state, which assumes the paginator's not paginating at this point. In a soon future, we'll get rid of the `add_initial_events()` function because the event cache will handle its own persistent storage; in the meantime, a correct fix is to skip `add_initial_events()` if there was already something in the linked chunk. After all, we're likely to fill the initial events with the same events all the time, or a subset of more recent events. By doing that, we're likely keeping *more* events in the linked chunk, instead.

Thanks to @stefanceriu for reporting the issue and confirming the fix works!

### DIFF
--- a/crates/matrix-sdk/src/event_cache/mod.rs
+++ b/crates/matrix-sdk/src/event_cache/mod.rs
@@ -319,6 +319,12 @@ impl EventCache {
 
         let room_cache = self.inner.for_room(room_id).await?;
 
+        // If the linked chunked already has at least one chunk (gap or events), ignore
+        // this request, as it should happen at most once per room.
+        if room_cache.inner.state.read().await.events().chunks().next().is_some() {
+            return Ok(());
+        }
+
         // We could have received events during a previous sync; remove them all, since
         // we can't know where to insert the "initial events" with respect to
         // them.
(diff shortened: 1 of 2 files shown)


stefanceriu (approved):
Wow super nice, thank you very much for tracking it down! 🙏

crates/matrix-sdk/src/event_cache/mod.rs
```diff
@@ -319,6 +319,12 @@ impl EventCache {
 
         let room_cache = self.inner.for_room(room_id).await?;
 
+        // If the linked chunked already has at least one chunk (gap or events), ignore
+        // this request, as it should happen at most once per room.
+        if room_cache.inner.state.read().await.events().chunks().next().is_some() {
```
Hywan, line 324: because the linkedchunk has, by default, one chunk, wouldn't trigger false-positive here?

bnjbvr: Duh, I think you're right, and this effectively disables the hacky temporary cache. And since I've removed `add_initial_events` from our testing recently, we didn't notice. Oh well, will fix today :)

Issue #4186, opened 2024-10-30
Device Dehydration: Cache the dehydration key


BillCarsonFr:
Part of https://github.com/element-hq/element-meta/issues/2603

Currently, the dehydration key is only kept in memory, which means that when the client gets restarted, it will no longer have the dehydration key, and will not be able to create new dehydrated devices. We need to cache it in the crypto store so that the SDK can fetch it on statup.

- [ ] Add functions to store/get dehydration key: store the key, fetch the key (or None if no key is stored), delete the key
- [ ] sqlite migration
- [ ] indexeddb migration

PR #4383, opened 2024-12-06
feat(crypto): Support storing the dehydrated device pickle key

BillCarsonFr:
Fixes https://github.com/matrix-org/matrix-rust-sdk/issues/4186

No need for migration, the pickle key is stored in the generic Key/Value store.

Signed-off-by:

### DIFF
--- a/bindings/matrix-sdk-crypto-ffi/src/dehydrated_devices.rs
+++ b/bindings/matrix-sdk-crypto-ffi/src/dehydrated_devices.rs
@@ -1,8 +1,11 @@
 use std::{mem::ManuallyDrop, sync::Arc};
 
-use matrix_sdk_crypto::dehydrated_devices::{
-    DehydratedDevice as InnerDehydratedDevice, DehydratedDevices as InnerDehydratedDevices,
-    RehydratedDevice as InnerRehydratedDevice,
+use matrix_sdk_crypto::{
+    dehydrated_devices::{
+        DehydratedDevice as InnerDehydratedDevice, DehydratedDevices as InnerDehydratedDevices,
+        RehydratedDevice as InnerRehydratedDevice,
+    },
+    store::DehydratedDeviceKey,
 };
 use ruma::{api::client::dehydrated_device, events::AnyToDeviceEvent, serde::Raw, OwnedDeviceId};
 use serde_json::json;
@@ -177,13 +180,13 @@ impl From<dehydrated_device::put_dehydrated_device::unstable::Request>
     }
 }
 
-fn get_pickle_key(pickle_key: &[u8]) -> Result<Box<[u8; 32]>, DehydrationError> {
+fn get_pickle_key(pickle_key: &[u8]) -> Result<DehydratedDeviceKey, DehydrationError> {
     let pickle_key_length = pickle_key.len();
 
     if pickle_key_length == 32 {
-        let mut key = Box::new([0u8; 32]);
-        key.copy_from_slice(pickle_key);
-
+        let mut raw_bytes = [0u8; 32];
+        raw_bytes.copy_from_slice(pickle_key);
+        let key = DehydratedDeviceKey::from_bytes(&raw_bytes);
         Ok(key)
     } else {
         Err(DehydrationError::PickleKeyLength(pickle_key_length))
--- a/crates/matrix-sdk-crypto/src/store/memorystore.rs
+++ b/crates/matrix-sdk-crypto/src/store/memorystore.rs
@@ -31,8 +31,8 @@ use vodozemac::Curve25519PublicKey;
 
 use super::{
     caches::{DeviceStore, GroupSessionStore},
-    Account, BackupKeys, Changes, CryptoStore, InboundGroupSession, PendingChanges, RoomKeyCounts,
-    RoomSettings, Session,
+    Account, BackupKeys, Changes, CryptoStore, DehydratedDeviceKey, InboundGroupSession,
+    PendingChanges, RoomKeyCounts, RoomSettings, Session,
 };
 use crate::{
     gossiping::{GossipRequest, GossippedSecret, SecretInfo},
@@ -93,6 +93,7 @@ pub struct MemoryStore {
     leases: StdRwLock<HashMap<String, (String, Instant)>>,
     secret_inbox: StdRwLock<HashMap<String, Vec<GossippedSecret>>>,
     backup_keys: RwLock<BackupKeys>,
+    dehydrated_device_pickle_key: RwLock<Option<DehydratedDeviceKey>>,
     next_batch_token: RwLock<Option<String>>,
     room_settings: StdRwLock<HashMap<OwnedRoomId, RoomSettings>>,
 }
@@ -116,6 +117,7 @@ impl Default for MemoryStore {
             custom_values: Default::default(),
             leases: Default::default(),
             backup_keys: Default::default(),
+            dehydrated_device_pickle_key: Default::default(),
             secret_inbox: Default::default(),
             next_batch_token: Default::default(),
             room_settings: Default::default(),
@@ -266,6 +268,11 @@ impl CryptoStore for MemoryStore {
 
         if let Some(version) = changes.backup_version {
             self.backup_keys.write().await.backup_version = Some(version);
+        }
+
+        if let Some(pickle_key) = changes.dehydrated_device_pickle_key {
+            let mut lock = self.dehydrated_device_pickle_key.write().await;
+            *lock = Some(pickle_key);
         }
 
         {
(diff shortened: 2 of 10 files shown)


crates/matrix-sdk-crypto/src/machine/mod.rs
```diff
@@ -2378,6 +2378,33 @@ impl OlmMachine {
         DehydratedDevices { inner: self.to_owned() }
     }
 
+    /// Get the cached dehydrated device pickle key if any.
+    ///
+    /// None if the key was not previously cached (via
+    /// [`Self::save_dehydrated_device_pickle_key`]).
+    ///
+    /// Should be used to periodically rotate the dehydrated device to avoid
+    /// OTK exhaustion and accumulation of to_device messages.
```
poljar, line 2387:
```suggestion
    /// one-time key exhaustion and accumulation of to_device messages.
```

crates/matrix-sdk-crypto/src/machine/mod.rs
```diff
@@ -2378,6 +2378,33 @@ impl OlmMachine {
...
+    pub async fn get_dehydrated_device_pickle_key(
+        &self,
+    ) -> StoreResult<Option<DehydratedDeviceKey>> {
+        self.inner.store.load_dehydrated_device_pickle_key().await
+    }
+
+    /// Store the dehydrated device pickle key in the crypto store.
+    ///
+    /// Use `None` to delete any previously saved pickle key.
```
poljar, line 2396: Doesn't that meant that we're going to accidentally delete the pickle key after every `receive_sync_changes()` call?

crates/matrix-sdk-indexeddb/src/crypto_store/mod.rs
```diff
@@ -487,6 +491,13 @@ impl IndexeddbCryptoStore {
             );
         }
 
+        if let Some(i) = &dehydration_pickle_key {
```
poljar, line 494: This doesn't seem to match to the documentation? There doesn't seem to be a way to delete the pickle key.

crates/matrix-sdk-crypto/src/machine/mod.rs
```diff
@@ -2378,6 +2378,33 @@ impl OlmMachine {
...
+    ///
+    /// This is useful if the client wants to periodically rotate dehydrated
+    /// devices to avoid OTK exhaustion and accumulated to_device problems.
```
poljar, line 2399:
```suggestion
    /// devices to avoid one-time key exhaustion and accumulated to_device problems.
```

crates/matrix-sdk-sqlite/src/crypto_store.rs
```diff
@@ -846,6 +852,11 @@ impl CryptoStore for SqliteCryptoStore {
                     txn.set_kv("backup_version_v1", &serialized_backup_version)?;
                 }
 
+                if let Some(pickle_key) = &changes.dehydrated_device_pickle_key {
```
poljar, line 855: Same as for the indexedDB implementation, docs and implementation don't match.

poljar (changes requested):

crates/matrix-sdk-indexeddb/src/crypto_store/mod.rs
BillCarsonFr: yes I spotted that problem when doing the wasm test. Trying to see how to do that properly

poljar: An enum with three variants could do the trick, something like `JsOption`.

crates/matrix-sdk-crypto/src/machine/mod.rs
BillCarsonFr: Updated

crates/matrix-sdk-sqlite/src/crypto_store.rs
```diff
@@ -1111,6 +1111,12 @@ impl CryptoStore for SqliteCryptoStore {
             .transpose()
     }
 
+    async fn delete_dehydrated_device_pickle_key(&self) -> Result<(), Self::Error> {
+        let conn = self.acquire().await?;
```
BillCarsonFr, line 1115: Should I take a lock on `let _guard = self.save_changes_lock.lock().await;` here ?
In case of a race between that and save_changes??

poljar: I don't think you need to do that since the transaction/connection will already act as a lock. The `save_changes_lock()` is necessary because we have a bunch of await points before we acquire the SQLite transaction.

crates/matrix-sdk-crypto/src/machine/mod.rs
BillCarsonFr: Fixed, there is a specific API to delete the pickle key now

crates/matrix-sdk-indexeddb/src/crypto_store/mod.rs
BillCarsonFr: https://github.com/matrix-org/matrix-rust-sdk/pull/4383/commits/c66a8fc608dd342c5a62c31cdd8de0608cd801d8

crates/matrix-sdk-crypto/src/machine/mod.rs
```diff
@@ -2378,6 +2378,35 @@ impl OlmMachine {
...
+    pub async fn save_dehydrated_device_pickle_key(
+        &self,
+        dehydrated_device_pickle_key: &DehydratedDeviceKey,
+    ) -> Result<(), CryptoStoreError> {
+        let changes = Changes {
+            dehydrated_device_pickle_key: Some(dehydrated_device_pickle_key.clone()),
+            ..Default::default()
+        };
+        self.inner.store.save_changes(changes).await
+    }
```
poljar, line 2408: All the methods related to dehydrated devices are under the "fake", "API" struct called `DehydratedDevices` let's move these under there as well. We should try really hard to avoid adding new methods directly under the `OlmMachine`.

crates/matrix-sdk-crypto/src/store/memorystore.rs
```diff
@@ -486,6 +493,16 @@ impl CryptoStore for MemoryStore {
...
 
+    async fn load_dehydrated_device_pickle_key(&self) -> Result<Option<DehydratedDeviceKey>> {
+        Ok(self.dehydrated_device_pickle_key.read().await.to_owned())
+    }
+
+    async fn delete_dehydrated_device_pickle_key(&self) -> Result<()> {
+        let mut lock = self.dehydrated_device_pickle_key.write().await;
+        *lock = None;
+        Ok(())
+    }
```
poljar, line 504: Maybe I'm missing something, but is this method used anywhere? It's part of the trait, but It's not exposed to users nor do we internally delete the key anywhere.

Would it make sense to delete it automagically when we create a new dehydrated device?

poljar (approved):
Looks good, left a question and some nits about moving the public methods around.

crates/matrix-sdk-crypto/src/machine/mod.rs
```diff
@@ -2378,6 +2378,35 @@ impl OlmMachine {
         DehydratedDevices { inner: self.to_owned() }
     }
 
+    /// Get the cached dehydrated device pickle key if any.
+    ///
+    /// None if the key was not previously cached (via
+    /// [`Self::save_dehydrated_device_pickle_key`]).
```
poljar, line 2384: Oh, another thing. Let's not use `Self::` here, let's expand to the proper struct.

bindings/matrix-sdk-crypto-ffi/src/dehydrated_devices.rs
```diff
@@ -177,13 +180,13 @@ impl From<dehydrated_device::put_dehydrated_device::unstable::Request>
...
-fn get_pickle_key(pickle_key: &[u8]) -> Result<Box<[u8; 32]>, DehydrationError> {
+fn get_pickle_key(pickle_key: &[u8]) -> Result<DehydratedDeviceKey, DehydrationError> {
     let pickle_key_length = pickle_key.len();
 
     if pickle_key_length == 32 {
-        let mut key = Box::new([0u8; 32]);
-        key.copy_from_slice(pickle_key);
-
+        // safe to unwrap because we checked the size
+        let raw_bytes: [u8; 32] = pickle_key.try_into().unwrap();
```
poljar, line 188: I would leave the copy from slice, `try_into()` moves `self` and moves are copies in Rust. Now here it's fine because you move a reference so we only copy the reference instead of the whole byte array.

But we always handle secret key material using `copy_from_slice()` and there it's obvious where the copy is made. The panic/unwrap is then also implicit and it's understood that you need a size check before.

crates/matrix-sdk-crypto/src/machine/mod.rs
BillCarsonFr: Moved https://github.com/matrix-org/matrix-rust-sdk/pull/4383/commits/3d32679d1b10d4bed8eecd87d6de00e6fbadb673

crates/matrix-sdk-crypto/src/store/memorystore.rs
BillCarsonFr: Exposed now. Should be used by apps.
It could make sense if we changed a bit the API, currently create does not need the pickle key, it's upload_keys that use it. Not sure how it would work. Need to talk a bit with @uhoreg to see if can have a better integrated usage of pickle key handling

bindings/matrix-sdk-crypto-ffi/src/dehydrated_devices.rs
BillCarsonFr: https://github.com/matrix-org/matrix-rust-sdk/pull/4383/commits/c6a85e2322b8eb470f58cb3e485dc7955a2219a3

PR #4384, opened 2024-12-06
Provide the reason why an event was an expected UTD

andybalaam:
Part of https://github.com/element-hq/element-meta/issues/2638

Distinguish between 3 cases where we can't find keys for an old message:

- There is no backup -> `HistoricalMessageAndBackupIsDisabled`
- Backup is not working and we have not verified our device -> `HistoricalMessageAndDeviceIsUnverified`
- Something else -> `Unknown`

### DIFF
--- a/crates/matrix-sdk-crypto/src/types/events/utd_cause.rs
+++ b/crates/matrix-sdk-crypto/src/types/events/utd_cause.rs
@@ -24,6 +24,15 @@ use serde::Deserialize;
 pub enum UtdCause {
     /// We don't have an explanation for why this UTD happened - it is probably
     /// a bug, or a network split between the two homeservers.
+    ///
+    /// For example:
+    ///
+    /// - the keys for this event are missing, but a key storage backup exists
+    ///   and is working, so we should be able to find the keys in the backup.
+    ///
+    /// - the keys for this event are missing, and a key storage backup exists
+    ///   on the server, but that backup is not working on this client even
+    ///   though this device is verified.
     #[default]
     Unknown = 0,
 
@@ -49,14 +58,17 @@ pub enum UtdCause {
     UnknownDevice = 4,
 
     /// We are missing the keys for this event, but it is a "device-historical"
-    /// message and no backup is accessible or usable.
+    /// message and there is no key storage backup on the server, presumably
+    /// because the user has turned it off.
     ///
     /// Device-historical means that the message was sent before the current
     /// device existed (but the current user was probably a member of the room
     /// at the time the message was sent). Not to
     /// be confused with pre-join or pre-invite messages (see
     /// [`UtdCause::SentBeforeWeJoined`] for that).
-    HistoricalMessage = 5,
+    ///
+    /// Expected message to user: "History is not available on this device".
+    HistoricalMessageAndBackupIsDisabled = 5,
 
     /// The keys for this event are intentionally withheld.
     ///
@@ -70,6 +82,19 @@ pub enum UtdCause {
     /// this device by cherry-picking and blocking it, in which case, no action
     /// can be taken on our side.
     WithheldBySender = 7,
+
+    /// We are missing the keys for this event, but it is a "device-historical"
+    /// message, and even though a key storage backup does exist, we can't use
+    /// it because our device is unverified.
+    ///
+    /// Device-historical means that the message was sent before the current
+    /// device existed (but the current user was probably a member of the room
+    /// at the time the message was sent). Not to
+    /// be confused with pre-join or pre-invite messages (see
+    /// [`UtdCause::SentBeforeWeJoined`] for that).
+    ///
+    /// Expected message to user: "You need to verify this device".
+    HistoricalMessageAndDeviceIsUnverified = 8,
 }
 
 /// MSC4115 membership info in the unsigned area.
@@ -97,6 +122,14 @@ pub struct CryptoContextInfo {
     /// if an event is device-historical or not (sent before the current device
     /// existed).
     pub device_creation_ts: MilliSecondsSinceUnixEpoch,
+
+    /// True if this device is secure because it has been verified by us
+    pub this_device_is_verified: bool,
+
+    /// True if key storage exists on the server, even if we are unable to use
+    /// it
+    pub backup_exists_on_server: bool,
+
     /// True if key storage is correctly set up and can be used by the current
     /// client to download and decrypt message keys.
     pub is_backup_configured: bool,
@@ -134,14 +167,9 @@ impl UtdCause {
                 }
 
                 if let Ok(timeline_event) = raw_event.deserialize() {
-                    if crypto_context_info.is_backup_configured
-                        && timeline_event.origin_server_ts()
-                            < crypto_context_info.device_creation_ts
-                    {
-                        // It's a device-historical message and there is no accessible
-                        // backup. The key is missing and it
-                        // is expected.
-                        return UtdCause::HistoricalMessage;
+                    if timeline_event.origin_server_ts() < crypto_context_info.device_creation_ts {
+                        // This event was sent before this device existed, so it is "historical"
+                        return UtdCause::determine_historical(crypto_context_info);
                     }
                 }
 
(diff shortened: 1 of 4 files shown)


crates/matrix-sdk-crypto/src/types/events/utd_cause.rs
```diff
@@ -163,6 +190,44 @@ impl UtdCause {
...
+     *
+     * X([Normal UTD error])
+     * Y([History is not available on this device])
+     * Z([You need to verify this device])
+     * ```
+     */
+    fn determine_historical(crypto_context_info: CryptoContextInfo) -> UtdCause {
+        if crypto_context_info.backup_exists_on_server {
+            if crypto_context_info.is_backup_configured
+                || crypto_context_info.this_device_is_verified
```
stefanceriu, line 221: Can non-verified devices even have access to the backups? 🤔

stefanceriu (approved):
Super nice and clear, lgtm! 👏

crates/matrix-sdk-crypto/src/types/events/utd_cause.rs
stefanceriu: > really means "backup is working".

Oh okay, yes, that clears it up, thank you.

> but it's my understanding

That's my understanding as well so let's just leave it at that 😅

crates/matrix-sdk/src/room/mod.rs
```diff
@@ -620,13 +620,26 @@ impl Room {
...
+                (device.first_time_seen_ts(), device.is_verified_with_cross_signing())
+            }
+
+            // Should not happen, there will always be an own device
+            _ => (MilliSecondsSinceUnixEpoch::now(), true),
+        };
+
+        let is_backup_configured = encryption.backups().state() == BackupState::Enabled;
+        let backup_exists_on_server =
+            encryption.backups().exists_on_server().await.unwrap_or(false);
```
richvdh, line 636: it feels a bit odd that we would check if there is a backup on the server, even if the user has explicitly disabled backups? Maybe it's fine though.

crates/matrix-sdk-crypto/src/types/events/utd_cause.rs
```diff
@@ -49,14 +58,16 @@ pub enum UtdCause {
     UnknownDevice = 4,
 
     /// We are missing the keys for this event, but it is a "device-historical"
-    /// message and no backup is accessible or usable.
+    /// message and key storage backup is disabled.
```
richvdh, line 61: The term "disabled" seems a bit handwavey. Could we be more explicit about what the potential causes are?

(explicitly disabled by user, no backup on the server (?) missing signature on the backup, we don't have the keys despite being verified (?), etc)

crates/matrix-sdk-crypto/src/types/events/utd_cause.rs
```diff
@@ -163,6 +190,44 @@ impl UtdCause {
...
+     * Y([History is not available on this device])
+     * Z([You need to verify this device])
+     * ```
+     */
+    fn determine_historical(crypto_context_info: CryptoContextInfo) -> UtdCause {
+        if crypto_context_info.backup_exists_on_server {
+            if crypto_context_info.is_backup_configured
+                || crypto_context_info.this_device_is_verified
+            {
+                UtdCause::Unknown
```
richvdh, line 223: A comment might help clear up the confusion here.

```suggestion
                // Either backup download is working but we still don't have the key, or backup download
                // *isn't* working despite being verified. In either case, we shrug and give an `Unknown` cause.
                UtdCause::Unknown
```

crates/matrix-sdk-crypto/src/types/events/utd_cause.rs
andybalaam: I'm fairly sure we just mean "turned off by the user" here. I'll write that.

andybalaam: Hmm. My logic is "does not exist on the server".

crates/matrix-sdk-crypto/src/types/events/utd_cause.rs
```diff
@@ -163,6 +190,46 @@ impl UtdCause {
...
+     * Z([You need to verify this device])
+     * ```
+     */
+    fn determine_historical(crypto_context_info: CryptoContextInfo) -> UtdCause {
+        if crypto_context_info.backup_exists_on_server {
+            if crypto_context_info.is_backup_configured
+                || crypto_context_info.this_device_is_verified
+            {
+                // Either backup download is working but we still don't have the key, or backup download
+                // *isn't* working despite being verified. In either case, we shrug and give an `Unknown` cause.
```
BillCarsonFr, line 224: As discussed, I think this is inconsistent with web were any device historical UTD is reported using `HistoricalMessage`. Here EX will report those as `OlmKeysNotSentError` https://github.com/element-hq/element-x-ios/blob/9856e3e5b45d3896eba7b8cd84f4cedfffc92d3f/ElementX/Sources/FlowCoordinators/UserSessionFlowCoordinator.swift#L387

BillCarsonFr (changes requested):
I think we need a 3rd variant to stay consistent with web, or we need to update web to be consistent with EX

crates/matrix-sdk-crypto/src/types/events/utd_cause.rs
andybalaam: btw I was working from https://github.com/element-hq/element-meta/issues/2313#issuecomment-2507713793 which was my modification of the flowchart above that Rich made. We should figure this stuff out and write it down somewhere sensible.

richvdh: > As discussed, I think this is inconsistent with web were any device historical UTD is reported using HistoricalMessage.

I must say, that doesn't match my understanding. I thought HistoricalMessage was reserved for the *expected* UTD case of a historical message with no backup on the server.

andybalaam: Yes @BillCarsonFr this is why I think what I implemented matches Web:

From [DecryptionFailureBody.getErrorMessage](https://github.com/element-hq/element-web/blob/d68c5a26af1aa8e17300eb3e0d3f76a804f89c3d/src/components/views/messages/DecryptionFailureBody.tsx#L19)

image

It handles 3 historical cases: no backup exists, backup fails and we are unverified, and everything else - "unable to decrypt".

crates/matrix-sdk-crypto/src/types/events/utd_cause.rs
```diff
@@ -163,6 +191,47 @@ impl UtdCause {
...
+     * C -- Yes --> X
+     * D{Is this device verified?}
+     * D -- No --> Z
+     * D -- Yes --> X
+     *
+     * X([Normal UTD error])
+     * Y([History is not available on this device])
+     * Z([You need to verify this device])
+     * ```
+     */
```
poljar, line 218: Sorry to barge in as an uninvited reviewer, but mermaid is nice if it's going to be rendered, this is a private function inside of the crate and thus will never end up in the rendered docs.

I don't mind the source being here, but can we render this using something like: https://mermaid-ascii.art/ as well?

poljar (dismissed):

crates/matrix-sdk-crypto/src/types/events/utd_cause.rs
andybalaam: We agreed in real life that this change is an improvement even though it's not the final solution, so we are OK to land it.

crates/matrix-sdk-crypto/src/types/events/utd_cause.rs
```diff
@@ -163,6 +191,47 @@ impl UtdCause {
...
+                // Either backup download is working but we still don't have the key, or backup
+                // download *isn't* working despite being verified. In either
+                // case, we shrug and give an `Unknown` cause.
+                UtdCause::Unknown
+            } else {
+                UtdCause::HistoricalMessageAndDeviceIsUnverified
+            }
+        } else {
+            UtdCause::HistoricalMessageAndBackupIsDisabled
+        }
```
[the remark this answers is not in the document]
andybalaam: I made another attempt in 0542c05738

crates/matrix-sdk-crypto/src/types/events/utd_cause.rs
andybalaam: mermaid-ascii produced something terrible, so I just reformatted what we had to be aimed at humans instead of mermaid.

andybalaam: 68cc73e0d9

BillCarsonFr (approved):
I think it's an improvement but I'd like to have some follow up tasks.

For example we fallback to `Unknwown` in `determine_historical`. It's a bit annoying because we report `Unknwown` in posthog as `OlmKeysNotSent` which sounds bad. This key was never sent. So maybe we need a `BackupUnknown` that would be mapped to something else in posthog (or we check the event age before reporting in posthog)


And just another comment, I though that we were caching the `exists_on_server()` but I don't see that in this PR? 
Are we going to do a server query at ~every UTD report?

andybalaam:
> And just another comment, I though that we were caching the `exists_on_server()` but I don't see that in this PR? Are we going to do a server query at ~every UTD report?

Thank you for checking. Yes, that was done in https://github.com/matrix-org/matrix-rust-sdk/pull/4356 so our call to `exists_on_server` in this PR uses the cached value where available.

PR #4385, opened 2024-12-06
feat(send queue): Add an enqueued time to to-be-sent events

zzorba:
Add a new created_at to the send_queue_events and dependent_send_queue_events stored records. This will allow clients to understand how stale a pending message might be in the event that the queue encounters and error and becomes wedged.

This change is exposed through the FFI on the `EventTimelineItem` struct as a new optional field named `local_created_at`.  It will be `None` for any Remote event, and `Some` for Local events (except for those that were enqueued before the migrations were run).

Signed-off-by: Daniel Salinas


crates/matrix-sdk-sqlite/src/state_store.rs
```diff
@@ -1824,26 +1834,28 @@ impl StateStore for SqliteStateStore {
...
-                        .mapped(|row| Ok((row.get(0)?, row.get(1)?, row.get(2)?, row.get(3)?)))
+                        .mapped(|row| Ok((row.get(0)?, row.get(1)?, row.get(2)?, row.get(3)?, row.get(4)?)))
                         .collect()
                 },
             )
             .await?;
 
         let mut requests = Vec::with_capacity(res.len());
         for entry in res {
+            let enqueue_time = MilliSecondsSinceUnixEpoch(UInt::new(entry.4).unwrap());
```
bnjbvr, line 1852: We can't use `unwrap()` here, we'd need to return an error if the `UInt` call failed.

crates/matrix-sdk-sqlite/src/state_store.rs
```diff
@@ -1909,7 +1921,6 @@ impl StateStore for SqliteStateStore {
         // See comment in `save_send_queue_event`.
         let parent_txn_id = parent_txn_id.to_string();
         let own_txn_id = own_txn_id.to_string();
-
```
bnjbvr, line 1912: Can we keep this line, please?

crates/matrix-sdk-sqlite/src/state_store.rs
```diff
@@ -2011,26 +2022,28 @@ impl StateStore for SqliteStateStore {
...
-                        .mapped(|row| Ok((row.get(0)?, row.get(1)?, row.get(2)?, row.get(3)?)))
+                        .mapped(|row| Ok((row.get(0)?, row.get(1)?, row.get(2)?, row.get(3)?, row.get(4)?)))
                         .collect()
                 },
             )
             .await?;
 
         let mut dependent_events = Vec::with_capacity(res.len());
         for entry in res {
+            let enqueue_time = MilliSecondsSinceUnixEpoch(UInt::new(entry.4).unwrap());
```
bnjbvr, line 2040: ditto here

crates/matrix-sdk-sqlite/src/state_store.rs
```diff
@@ -2427,7 +2440,6 @@ mod migration_tests {
         let room_id_value = this.serialize_value(&room_id.to_owned())?;
 
         let content = this.serialize_json(&content)?;
-
```
bnjbvr, line 2430: Can we keep this line here too?

crates/matrix-sdk-base/src/store/send_queue.rs
```diff
@@ -365,6 +369,9 @@ pub struct DependentQueuedRequest {
     /// If the parent request has been sent, the parent's request identifier
     /// returned by the server once the local echo has been sent out.
     pub parent_key: Option<SentRequestKey>,
+
+    /// The time that the request was original attempted.
+    pub enqueue_time: Option<MilliSecondsSinceUnixEpoch>,
```
bnjbvr, line 374: I'm not sure about this, but since this is (de)serialized: should this have a `#[serde(default)]` annotation to indicate that if the field is missing, it should be `None` — or is that the default (heh) behavior for `Option`s?

crates/matrix-sdk-indexeddb/src/state_store/mod.rs
```diff
@@ -444,6 +444,8 @@ struct PersistedQueuedRequest {
     is_wedged: Option<bool>,
 
     event: Option<SerializableEventContent>,
+
+    enqueue_time: Option<MilliSecondsSinceUnixEpoch>,
```
bnjbvr, line 448: Same question about `serde(default)` here.

crates/matrix-sdk-base/src/store/send_queue.rs
```diff
@@ -131,6 +132,9 @@ pub struct QueuedRequest {
     /// The bigger the value, the higher the priority at which this request
     /// should be handled.
     pub priority: usize,
+
+    /// The time that the request was original attempted.
+    pub enqueue_time: Option<MilliSecondsSinceUnixEpoch>,
```
bnjbvr, line 137: What do you think of calling this (everywhere, so in the SQL too) `creation_time`, or even `created_at`?

bnjbvr (commented):
Looks like this is on the right path! It would be nice to expose the value in the local echoes themselves then, and test this all works fine.

crates/matrix-sdk-sqlite/src/state_store.rs
zzorba: Do you think dropping the malformed data is reasonable fallback if it fails to unpack the data?
```
  let enqueue_time = UInt::new(entry.4).map(|e| MilliSecondsSinceUnixEpoch(e));
```

crates/matrix-sdk-base/src/store/send_queue.rs
zzorba: Yeah, it seems like generally we have annotations for serializing (to skip if it is None).  I'll add that here for consistency.

crates/matrix-sdk-base/src/store/send_queue.rs
zzorba: Love that name, will add

zzorba:
@bnjbvr thank you for the review.  Made your suggested corrections (one pending question), and then in a second commit I have attempted to expose this timestamp through the api layer.  

It was more involved than I thought, but I think there was only one spot where I wasn't quite sure which timestamp to use (related to retry of a media message I believe, I left a comment).  Right now I have the storage layer assigning the timestamp and returning it to be used by the SendHandle system.  If you think it would be cleaner to pass in a timestamp I am open to changing this.

If you think this API shape is reasonable, I will work to add some tests (or extend the existing tests) to exercise it.

crates/matrix-sdk-sqlite/src/state_store.rs
bnjbvr: Yeah, it's a good idea to keep it `Option`al :+1:

crates/matrix-sdk/src/send_queue.rs
```diff
@@ -2138,6 +2159,8 @@ impl SendReactionHandle {
             room: self.room.clone(),
             transaction_id: self.transaction_id.clone().into(),
             media_handles: None,
+            // This isn't quite right, should this be the newly enqueued cancel event?
+            created_at: self.created_at.clone(),
```
bnjbvr, line 2163: You're right, it should be `Some(MilliSecondsSinceUnixEpoch::now())`. I think it's a fine value to set `now()` here, since this is the time of the request to cancel.

crates/matrix-sdk-base/src/store/traits.rs
```diff
@@ -361,7 +361,7 @@ pub trait StateStore: AsyncTraitDeps {
         transaction_id: OwnedTransactionId,
         request: QueuedRequestKind,
         priority: usize,
-    ) -> Result<(), Self::Error>;
+    ) -> Result<MilliSecondsSinceUnixEpoch, Self::Error>;
```
bnjbvr, line 364: Instead of having the store return the time of creation: could it be fed by the caller, as a new function parameter? Then we can keep on returning `()` here, and we should have a bit less repetitive code in the stores.

bindings/matrix-sdk-ffi/src/timeline/mod.rs
```diff
@@ -1005,7 +1005,10 @@ impl TimelineItem {
 #[derive(Clone, uniffi::Enum)]
 pub enum EventSendState {
     /// The local event has not been sent yet.
-    NotSentYet,
+    NotSentYet {
+        /// When the send was first attempted.
```
bnjbvr, line 1009: Call me uber-nitpicky :sweat_smile: but in most of the code comments relating to `created_at`: we're not talking about the time where it was first *sent*, but really when the request was created / the event was queued for sending. The request may be pushed while the device is offline, so the first send would happen only when the device gets back online, which could be hours/days later.

crates/matrix-sdk-sqlite/migrations/state_store/010_send_queue_enqueue_time.sql
```diff
@@ -0,0 +1,6 @@
+-- Migration script to add the created_at column to the send_queue_events table
+ALTER TABLE "send_queue_events"
+ADD COLUMN "created_at" INTEGER NOT NULL DEFAULT (strftime('%s', 'now'));
```
bnjbvr, line 3: I think it can be null now that the `created_at` field is optional, right?

crates/matrix-sdk-sqlite/migrations/state_store/010_send_queue_enqueue_time.sql
```diff
@@ -0,0 +1,6 @@
...
+
+ALTER TABLE "dependent_send_queue_events"
+ADD COLUMN "created_at" INTEGER NOT NULL DEFAULT (strftime('%s', 'now'));
```
bnjbvr, line 6: ditto here

bnjbvr (commented):
Looks like this is going in the right direction :+1: I haven't reviewed in depth, but left a few high-level comments in places. Let's go ahead and add some tests :muscle:

bindings/matrix-sdk-ffi/src/timeline/mod.rs
zzorba: Yes you are correct, I will update this language throughout:
```
/// When the send was first enqueued by the user.
```

bindings/matrix-sdk-ffi/src/timeline/mod.rs
```diff
@@ -1005,7 +1005,10 @@ impl TimelineItem {
 #[derive(Clone, uniffi::Enum)]
 pub enum EventSendState {
     /// The local event has not been sent yet.
-    NotSentYet,
+    NotSentYet {
+        /// When the send was first enqueued by the user.
+        created_at: Option<u64>,
```
bnjbvr, line 1010: Random thought: it seems to me that the `created_at` property is not so much depending on the send state, but it's useful for all local echoes in general. How about putting it in `LocalEventTimelineItem`, instead? That might get rid of many of the test changes (and cause many changes to be reverted, sorry!).

zzorba: Note a problem at all, you are correct, this is much cleaner, and makes for a much smaller change.

I was too focused on trying to get it into the NotSentYet state, as opposed to exposing it as a parallel field on the TimelineItem in the FFI.

zzorba:
@bnjbvr I had to mark the `push_media` method as a clippy exception because the additional argument pushed it over the line.  Thank you again for all your feedback on this feature, I think the feature has been greatly improved.

zzorba:
Rebased to resolve merge conflicts and flattened the commits.

zzorba:
Unclear to me what the current build break is related to in this PR.

bnjbvr:
> Unclear to me what the current build break is related to in this PR.

Yep, it's an unrelated intermittent failure, nothing to do with your PR. I'll take another look today :+1:

crates/matrix-sdk-ui/src/timeline/event_handler.rs
```diff
@@ -1033,6 +1033,7 @@ impl<'a, 'o> TimelineEventHandler<'a, 'o> {
                 send_state: EventSendState::NotSentYet,
                 transaction_id: txn_id.to_owned(),
                 send_handle: send_handle.clone(),
+                created_at: send_handle.clone().and_then(|h| h.created_at),
```
bnjbvr, line 1036: Probably the cloning is not required here:

```suggestion
                created_at: send_handle.created_at,
```

(or `created_at.clone()`, if borrowck complains)

crates/matrix-sdk-base/src/store/send_queue.rs
```diff
@@ -131,6 +132,9 @@ pub struct QueuedRequest {
     /// The bigger the value, the higher the priority at which this request
     /// should be handled.
     pub priority: usize,
+
+    /// The time that the request was original attempted.
```
bnjbvr, line 136:
```suggestion
    /// The time that the request was originally attempted.
```

crates/matrix-sdk-base/src/store/send_queue.rs
```diff
@@ -371,6 +375,10 @@ pub struct DependentQueuedRequest {
     /// If the parent request has been sent, the parent's request identifier
     /// returned by the server once the local echo has been sent out.
     pub parent_key: Option<SentRequestKey>,
+
+    /// The time that the request was original attempted.
```
bnjbvr, line 379:
```suggestion
    /// The time that the request was originally attempted.
```

crates/matrix-sdk-indexeddb/src/state_store/mod.rs
```diff
@@ -1391,7 +1402,6 @@ impl_state_store!({
             || Ok(Vec::new()),
             |val| self.deserialize_value::<Vec<PersistedQueuedRequest>>(&val),
         )?;
-
```
bnjbvr, line 1394: nooo i liked this blank line :)

crates/matrix-sdk-base/src/store/send_queue.rs
```diff
@@ -131,6 +132,9 @@ pub struct QueuedRequest {
     /// The bigger the value, the higher the priority at which this request
     /// should be handled.
     pub priority: usize,
+
+    /// The time that the request was original attempted.
+    pub created_at: Option<MilliSecondsSinceUnixEpoch>,
```
bnjbvr, line 137: Could this be non-optional, since this is not serialized we could always provide a value here? There might be places where we upgrade a `DependentQueuedRequest` with no `created_at` (because they were created before this change is introduced) to a `QueuedRequest`, but it seems fine to use `Milliseconds::now()` in those cases, maybe?

crates/matrix-sdk-sqlite/src/state_store.rs
```diff
@@ -2442,6 +2465,29 @@ mod migration_tests {
         txn.prepare_cached("INSERT INTO send_queue_events (room_id, room_id_val, transaction_id, content, wedged) VALUES (?, ?, ?, ?, ?)")?
             .execute((room_id_key, room_id_value, transaction_id.to_string(), content, is_wedged))?;
 
+        Ok(())
+    }
+    fn add_dependent_send_queue_event_v7(
```
bnjbvr, line 2470: I wish we had automatic rustfmt rules for this, but: we try to keep blank lines between functions, to visually separate them (and make moving from paragraph to paragraph fast, in editors like vim :nerd_face:)

```suggestion
    }
    
    fn add_dependent_send_queue_event_v7(
```

crates/matrix-sdk-ui/src/timeline/event_item/mod.rs
```diff
@@ -268,6 +268,11 @@ impl EventTimelineItem {
         as_variant!(&self.kind, EventTimelineItemKind::Local(local) => &local.send_state)
     }
 
+    /// Get the local time that the event was enqueued at.
```
bnjbvr, line 271: "Local time" makes me think we're talking about time zones, and we're not:

```suggestion
    /// Get the time that the local event was pushed in the send queue at.
```

crates/matrix-sdk-ui/src/timeline/tests/echo.rs
```diff
@@ -334,7 +334,7 @@ async fn test_no_reuse_of_counters() {
     let local_id = assert_next_matches_with_timeout!(stream, VectorDiff::PushBack { value: item } => {
         let event_item = item.as_event().unwrap();
         assert!(event_item.is_local_echo());
-        assert_matches!(event_item.send_state(), Some(EventSendState::NotSentYet));
+        assert_matches!(event_item.send_state(), Some(EventSendState::NotSentYet{ .. }));
```
bnjbvr, line 337: I think this change (and every other change related to `NotSentYet`) is not required anymore; could you revert them please?

crates/matrix-sdk-ui/tests/integration/timeline/queue.rs
```diff
@@ -273,7 +273,7 @@ async fn test_reloaded_failed_local_echoes_are_marked_as_failed() {
     // Local echoes are updated with the failed send state as soon as the error
     // response has been received.
     assert_let!(Some(VectorDiff::Set { index: 0, value: first }) = timeline_stream.next().await);
-    let (error, is_recoverable) = assert_matches!(first.send_state().unwrap(), EventSendState::SendingFailed { error, is_recoverable } => (error, is_recoverable));
+    let (error, is_recoverable) = assert_matches!(first.send_state().unwrap(), EventSendState::SendingFailed { error, is_recoverable, .. } => (error, is_recoverable));
```
bnjbvr, line 276: And for `SendingFailed` too.

crates/matrix-sdk/src/send_queue.rs
```diff
@@ -444,7 +444,8 @@ impl RoomSendQueue {
 
         let content = SerializableEventContent::from_raw(content, event_type);
 
-        let transaction_id = self.inner.queue.push(content.clone().into()).await?;
+        let created_at = MilliSecondsSinceUnixEpoch::now();
+        let transaction_id = self.inner.queue.push(content.clone().into(), created_at).await?;
```
bnjbvr, line 448: The PR mixes two ways of feeding the `created_at`:

- sometimes it's created at the `RoomSendQueue` level, and passed as a parameter to a `QueueStorage` function,
- sometimes it's the `QueueStorage` that will call `now()` and pass it directly to the underlying store functions.

We should make this a bit more uniform; I think I prefer the latter where the `QueueStorage` internally sets it. It's nicer, because it implies having fewer parameters to the `QueueStorage` external functions too.

crates/matrix-sdk/src/send_queue.rs
```diff
@@ -1888,6 +1914,9 @@ pub struct SendHandle {
 
     /// Additional handles for a media upload.
     media_handles: Option<MediaHandles>,
+
+    /// The time that this send handle was first created
+    pub created_at: Option<MilliSecondsSinceUnixEpoch>,
```
bnjbvr, line 1919: I see that the only place where this is read is in the timeline code, correct? As a last high-level comment: I think this could be moved to the `LocalEchoContent::Event` variant, even — we might not care about a reaction's `created_at` (or do we? if so, we could move it to the `LocalEcho` struct more generally).

bnjbvr (changes requested):
Looks like we're getting super close to land this. Sorry to ask a bit for more changes, I spotted a few ways to make this even cleaner.

(I will be on holidays starting today for a few weeks, so there might be a bit of delay before the next round of review. Happy end of the year, and thanks for bearing with us :))

bnjbvr:
> > Unclear to me what the current build break is related to in this PR.
> 
> Yep, it's an unrelated intermittent failure, nothing to do with your PR. I'll take another look today 👍

(If you rebase on top of main it should be fixed now.)

crates/matrix-sdk-ui/src/timeline/event_handler.rs
zzorba: So send_handle is an Option so I can't access created_at directly.  And if I try to map and clone, it seems to dislike the Option being mapped (since self.ctx_flow is a reference).  I think the clone is needed?

crates/matrix-sdk-base/src/store/send_queue.rs
zzorba: Yeah, so if we are okay with injecting Milliseconds::now when it fails to parse coming out of the DB (either due to a previous migration or due to ... cosmic rays, then I think we can safely make this non optional.

IMO that's totally fine to do as you explained, just wasn't sure about injecting false data.

crates/matrix-sdk/src/send_queue.rs
zzorba: So the original version had it being injected directly in the send_queue to minimize the API changes.  

But this meant though that the timestamp would have to be returned to be used in constructing a `SendHandle`, which we did not like.  Sometimes the SendHandles come from the persisted store, but sometimes they are constructed directly by people calling the send_queue method. 

Let me know which way you prefer.  Personally I think the current method is cleaner, and also makes it easier to write tests against.

crates/matrix-sdk/src/send_queue.rs
zzorba: So the issue is that it needs to be exposed in the `LocalEventTimelineItem` in `crates/matrix-sdk-ui/src/timeline/event_handler.rs` which doesn't appear to have access to LocalEchos.  Thats why I put it in the SendHandle.

crates/matrix-sdk-base/src/store/send_queue.rs
zzorba: I've updated it to use a ::now in most places and make it required everywhere except when it is exposed in the FFI (because SendHandle might not be present for the Local items)

crates/matrix-sdk-ui/src/timeline/event_handler.rs
jplatte: I think the field is `Copy` so it should work with `send_handle.as_ref().and_then(|h| h.created_at)`.

bnjbvr:
Hi! Sorry, I've run out of time this week and I forgot about this review. I'll try to get back to it early next week.

bnjbvr:
(Note to observers: I addressed my own review comments in the `stylistic changes` commit.)

PR #4386, opened 2024-12-07
chore(xtask): Fix clippy lints

zecakeh:
See the commit messages for details.

Since it was not noticed before, it seems that CI does not lint xtask with clippy. Maybe it should?

### DIFF
--- a/xtask/src/main.rs
+++ b/xtask/src/main.rs
@@ -1,3 +1,5 @@
+#![allow(unexpected_cfgs)]
+
 mod ci;
 mod fixup;
 mod kotlin;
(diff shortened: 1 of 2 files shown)


jplatte (approved):

poljar (approved):

poljar:
> Since it was not noticed before, it seems that CI does not lint xtask with clippy. Maybe it should?

Hmm, CI used to trigger for this but seems that it was only when buidling the docs which I disabled for xtask: 79c8d2c3455bb8dbdbd6b70420fdebe7f1529a15.

PR #4387, opened 2024-12-07
fix(media): Make sure that local MXC URIs only try to get media from the cache and ignore requested dimensions

zecakeh:
Extracted from #4329. This does not change the `MediaFormat` of the request used in the media cache by the send queue.

### DIFF
--- a/crates/matrix-sdk/src/media.rs
+++ b/crates/matrix-sdk/src/media.rs
@@ -40,8 +40,8 @@ use tempfile::{Builder as TempFileBuilder, NamedTempFile, TempDir};
 use tokio::{fs::File as TokioFile, io::AsyncWriteExt};
 
 use crate::{
-    attachment::Thumbnail, config::RequestConfig, futures::SendRequest, utils::not_found_error,
-    Client, Error, Result, TransmissionProgress,
+    attachment::Thumbnail, config::RequestConfig, futures::SendRequest, Client, Error, Result,
+    TransmissionProgress,
 };
 
 /// A conservative upload speed of 1Mbps
@@ -137,6 +137,10 @@ pub enum MediaError {
     /// Preallocated media already had content, cannot overwrite.
     #[error("preallocated media already had content, cannot overwrite")]
     CannotOverwriteMedia,
+
+    /// Local-only media content was not found.
+    #[error("local-only media content was not found")]
+    LocalMediaNotFound,
 }
 
 /// `IntoFuture` returned by [`Media::upload`].
@@ -400,6 +404,8 @@ impl Media {
         request: &MediaRequestParameters,
         use_cache: bool,
     ) -> Result<Vec<u8>> {
+        // Ignore request parameters for local medias, notably those pending in the send
+        // queue.
         if let Some(uri) = Self::as_local_uri(&request.source) {
             return self.get_local_media_content(uri).await;
         }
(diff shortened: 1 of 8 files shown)


crates/matrix-sdk/src/media.rs
```diff
@@ -400,6 +400,10 @@ impl Media {
         request: &MediaRequestParameters,
         use_cache: bool,
     ) -> Result<Vec<u8>> {
+        if let Some(uri) = Self::as_local_uri(&request.source) {
```
bnjbvr, line 403: Can you add a small comment on top: `// Ignore request parameters for local medias, notably those pending in the send queue.`

crates/matrix-sdk/src/media.rs
```diff
@@ -515,6 +519,26 @@ impl Media {
...
+        self.client
+            .event_cache_store()
+            .lock()
+            .await?
+            .get_media_content_for_uri(uri)
+            .await?
+            .ok_or_else(|| {
+                // A request to the server would return a 404 NOT_FOUND error, so let's simulate
+                // that.
+                not_found_error("Unknown media ID".to_owned()).into()
```
bnjbvr, line 538: I found this a bit contrived: instead, let's have `get_local_media_content` return `Result<Option<Vec<u8>>>`, and the caller can do something like

```
if let Some(local_uri) = request.source.as_local() {
  match get_local_media_content(local_uri) {
   Err(err) => { /* log the error, and let the function continue */ }
   Ok(Some(val)) => { return Ok(val) }
   Ok(None) => { /* nothing: let the code continue */ }
  }
}
```

crates/matrix-sdk/src/utils.rs
```diff
@@ -237,6 +246,16 @@ pub fn formatted_body_from(
     }
 }
 
+/// Construct a `404 NOT_FOUND` error response from the server, with the given
+/// message.
+pub(crate) fn not_found_error(message: String) -> HttpError {
```
bnjbvr, line 251: With the other suggestion, we could remove this.

bnjbvr (approved):
Thanks for splitting this out! Looks good to me, small request for simplifying the last commit, and we should be good to go :pray:

crates/matrix-sdk/src/media.rs
zecakeh: We don't want the function to continue though. If `get_media_content` continues it will try to load the media request from the cache _again_ and then make a request to the server with a bogus MXC URI.

bnjbvr: You're right, in that case can it return a more specialized error, and not an HTTP error? No HTTP requests have been done.

crates/matrix-sdk/src/utils.rs
zecakeh: Removed.

crates/matrix-sdk/src/media.rs
zecakeh: Done.

crates/matrix-sdk/src/media.rs
zecakeh: Added.

bnjbvr (approved):
lgtm, thanks!

crates/matrix-sdk/src/media.rs
```diff
@@ -391,6 +404,12 @@ impl Media {
         request: &MediaRequestParameters,
         use_cache: bool,
     ) -> Result<Vec<u8>> {
+        // Ignore request parameters for local medias, notably those pending in the send
```
jplatte, line 407: "media" is already plural.

Issue #4337, opened 2024-11-27
Slow call to `olmMachine.getRoomSettings()`


florianduros:
A bit of context first: we are migrating from the legacy crypto stack to rust crypto stack. In EW, we need to know if the room is encrypted to display it. This assertion is a part of the critical path of EW.

To know if the room is encrypted, we look at the room settings thanks to [`olmMachine.getRoomSettings()`](https://github.com/matrix-org/matrix-js-sdk/blob/develop/src/rust-crypto/rust-crypto.ts#L387).

We have spot a performance issue when trying to get this value from less than 1sec to multiple secs to respond.

poljar:
The relevant method for EW should live in the indexedDB crate:

https://github.com/matrix-org/matrix-rust-sdk/blob/3e7d7e8a3178d6fedf5e97cf4110020111a36251/crates/matrix-sdk-indexeddb/src/crypto_store/mod.rs#L1317-L1327

Nothing really jumps out, looks like a simple key/value get operation.

ANIR1604:
I would like work on this, if that's okay.

poljar:
No objections here, though I don't have any particular guidance on how to improve the situation.

ANIR1604:
Plz Assign it to me

ANIR1604:
> No objections here, though I don't have any particular guidance on how to improve the situation.

Can we use Cache System to speed up the process.

ANIR1604:
> No objections here, though I don't have any particular guidance on how to improve the situation.

Sir Plz Review my code if its ok to go with that ...

PR #4388, opened 2024-12-07
Optimising the DB Search

ANIR1604:
# Advantages : 

- Fixes : #4337 
- Instead of calling object_store and get separately in different lines, the code groups them together in a single chain. This makes the code more concise and improves readability.
- The ? operator is used after each method to propagate errors immediately if any of the calls fail, avoiding unnecessary logic or nesting.
- We avoid multiple method calls and unnecessary intermediate variables. The result of get() is directly assigned, which will later be handled by the map() and transpose() methods.

### DIFF
--- a/crates/matrix-sdk-indexeddb/src/crypto_store/mod.rs
+++ b/crates/matrix-sdk-indexeddb/src/crypto_store/mod.rs
@@ -46,11 +46,6 @@ use tokio::sync::Mutex;
 use tracing::{debug, warn};
 use wasm_bindgen::JsValue;
 use web_sys::IdbKeyRange;
-use std::collections::HashMap;
-use std::sync::Arc;
-use tokio::sync::RwLock;
-
-pub type RoomSettingsCache = Arc<RwLock<HashMap<String, Option<RoomSettings>>>>;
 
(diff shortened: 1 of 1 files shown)


ANIR1604:
> Thanks for the patch, but I don't see how it changes anything to the actual code.

Actually this PR is improving the readability and the error handling of the process which somewhat improving its optimization. 

We can use Cache Memory by Hashmaps to improve the DB calls should I go with that method ?

Hywan:
> Actually this PR is improving the readability and the error handling of the process which somewhat improving its optimization.

I need proof of that. Readibility, yes, a bit, but I'm not sure it improves optimization from the compiler. From the compiler point of view, once reduced to a simple form, there is no difference between the two versions.

ANIR1604:
> > Actually this PR is improving the readability and the error handling of the process which somewhat improving its optimization.
> 
> I need proof of that. Readibility, yes, a bit, but I'm not sure it improves optimization from the compiler. From the compiler point of view, once reduced to a simple form, there is no difference between the two versions.

Sir Can I use Cache Hashmaps to improve optimization ?

Hywan:
You can try for sure. I cannot merge this PR since it doesn't do anything useful and clearly not what the description claims.

PR #4390, opened 2024-12-09
feat(ffi): Add method to send custom events with JSON content

jonasrichardrichter:
This PR adds the send_raw methods to the bindings, making it usable from e.g. Swift.

Signed-off-by: Jonas Richard Richter <jonas-richard.richter@telekom.de>

### DIFF
--- a/bindings/matrix-sdk-ffi/src/room.rs
+++ b/bindings/matrix-sdk-ffi/src/room.rs
@@ -384,19 +384,17 @@ impl Room {
     ///
     /// * `event_type` - The type of the event to send.
     ///
-    /// * `content` - The content of the event to send as JSON string.
-    pub async fn send_custom_event(
+    /// * `content` - The content of the event to send encoded as JSON string.
+    pub async fn send_raw(
         &self,
         event_type: String,
         content: String,
     ) -> Result<(), ClientError> {
-        let content_json: serde_json::Value = match serde_json::from_str(&content) {
-            Ok(json) => json,
-            Err(e) => {
-                return Err(ClientError::Generic { msg: format!("Failed to parse JSON: {}", e) })
-            }
-        };
+        let content_json: serde_json::Value = serde_json::from_str(&content)
+            .map_err(|e| ClientError::Generic { msg: format!("Failed to parse JSON: {e}") })?;
+
         self.inner.send_raw(&event_type, content_json).await?;
+        
         Ok(())
     }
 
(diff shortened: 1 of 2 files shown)


bindings/matrix-sdk-ffi/src/room.rs
```diff
@@ -378,6 +378,28 @@ impl Room {
...
     }
 
+    /// Send a raw event to the room.
+    ///
+    /// # Arguments
+    ///
+    /// * `event_type` - The type of the event to send.
+    ///
+    /// * `content` - The content of the event to send as JSON string.
+    pub async fn send_custom_event(
```
poljar, line 388: We should keep the names the same, please rename this to `send_raw()`.

bindings/matrix-sdk-ffi/src/room.rs
```diff
@@ -378,6 +378,28 @@ impl Room {
...
+        content: String,
+    ) -> Result<(), ClientError> {
+        let content_json: serde_json::Value = match serde_json::from_str(&content) {
+            Ok(json) => json,
+            Err(e) => {
+                return Err(ClientError::Generic { msg: format!("Failed to parse JSON: {}", e) })
+            }
+        };
+        self.inner.send_raw(&event_type, content_json).await?;
+        Ok(())
```
poljar, line 400: This can be more concisely expressed using `map_err()` to convert the error type, a bit of whitespace also doesn't hurt:

```suggestion
       let content_json: serde_json::Value = serde_json::from_str(&content)
            .map_err(|e| ClientError::Generic { msg: format!("Failed to parse JSON: {e}") })?;

        self.inner.send_raw(&event_type, content_json).await?;

        Ok(())
```

bindings/matrix-sdk-ffi/src/room.rs
```diff
@@ -378,6 +378,28 @@ impl Room {
         Ok(())
     }
 
+    /// Send a raw event to the room.
+    ///
+    /// # Arguments
+    ///
+    /// * `event_type` - The type of the event to send.
+    ///
+    /// * `content` - The content of the event to send as JSON string.
```
poljar, line 387:
```suggestion
    /// * `content` - The content of the event to send encoded as JSON string.
```

poljar (changes requested):
Thanks, needs a bit more work but generally looks good.

jonasrichardrichter:
Thanks for your feedback :) These were my fist steps in the Rust space ^^ 
I've added all the requests you made.

poljar:
Seems that `rustfmt` isn't happy with you just yet, you can fix this if you run `cargo xtask fixup style`.

poljar (approved):
Thanks, looks good.

PR #4391, opened 2024-12-09
fix(event cache): use a correcter heuristic to decide whether to add initial events or not

bnjbvr:
Thanks @Hywan for spotting the issue.


Hywan (approved):
LGTM!

PR #4392, opened 2024-12-09
chore: Bump MSRV in READMEs

zecakeh:
To match the version in `Cargo.toml`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,10 +24,6 @@ The rust-sdk consists of multiple crates that can be picked at your convenience:
 - **matrix-sdk-crypto** - No (network) IO encryption state machine that can be
   used to add Matrix E2EE support to your client or client library.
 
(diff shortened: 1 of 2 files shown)


zecakeh:
Alternatively they could just be removed from the README. It used to be nice to have it because there was no other place to set it, but now that it appears explicitly in `Cargo.toml` it is not necessary anymore.

poljar:
> Alternatively they could just be removed from the README. It used to be nice to have it because there was no other place to document it, but now that it appears explicitly in Cargo.toml it is not necessary anymore.

Yeah, I don't think there's much point in having them in two places.

stefanceriu:
I was just about to ask. I disabled automerge and will wait for that change.

zecakeh:
Alright, I pushed a commit that removed the MSRV from the READMEs instead.

PR #4393, opened 2024-12-09
feat(event cache): start paginations from the most recent token, not the oldest token

bnjbvr:
Part of #3280. See the first commit message's description to get the reasoning.

Split from #4308.


andybalaam (approved):
Great write-up in the parent issue, thank you.

Looks good.

PR #4394, opened 2024-12-09
Improve typos configuration

szepeviktor:
Make typos config more strict: miss less future misspellings.

### DIFF
--- a/.typos.toml
+++ b/.typos.toml
@@ -13,13 +13,16 @@ extend-ignore-re = [
     "[1-9A-Za-z]{4}( [1-9A-Za-z]{4}){8,}",
     # some heuristics for base64 strings with no false matches found at the
     # time of writing.
-    "[0-9A-Za-z+=]{38,}",
-    '"ed25519:[^"]+": "[^"]+"',
-    '"sender_key": "[^"]+"',
-    '"x": "[^"]{43}"',
-    "\\bmat ba\\b",
-    "\\bSomOlz-t3ABvRUweGuLCpHFKsI6yhGCoqPyS7o5gpfenizdfHLqq-l7kgyr7lSbW_mTSyYutby",
+    '[0-9A-Za-z_.-]{62,}',
+    '"[0-9a-z_]+": "[0-9A-Za-z/+-]{41,}"',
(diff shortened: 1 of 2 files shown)


.typos.toml
```diff
@@ -1,39 +1,34 @@
...
-    "[1-9A-Za-z]{4}( [1-9A-Za-z]{4}){7,}",
+    "[1-9A-Za-z]{4}( [1-9A-Za-z]{4}){8,}",
     # some heuristics for base64 strings with no false matches found at the
     # time of writing.
-    "[A-Za-z0-9+=]{72,}",
-    "([A-Za-z0-9+=]|\\\\\\s\\*){72,}",
-    "[0-9+][A-Za-z0-9+]{30,}[a-z0-9+]",
-    "\\$[A-Z0-9+][A-Za-z0-9+]{6,}[a-z0-9+]",
-    "\\b[a-z0-9+/=][A-Za-z0-9+/=]{7,}[a-z0-9+/=][A-Z]\\b",
+    "[0-9A-Za-z+=]{38,}",
```
szepeviktor, line 16: This makes all other base64-related regexps unnecessary.

.typos.toml
```diff
@@ -1,39 +1,34 @@
...
+    '"ed25519:[^"]+": "[^"]+"',
+    '"sender_key": "[^"]+"',
+    '"x": "[^"]{43}"',
+    "\\bmat ba\\b",
+    "\\bSomOlz-t3ABvRUweGuLCpHFKsI6yhGCoqPyS7o5gpfenizdfHLqq-l7kgyr7lSbW_mTSyYutby",
+    '\bfHAblQV"',
```
szepeviktor, line 22: These cannot be pin-pointed.

bnjbvr (commented):
Thanks for the patch, but I fail to understand what it improves. It will go into the hidden directories, which might be nice or annoying according to the developer who runs into it. It replaces 6 false positives with 5 false positives. Does it do something else that I'm not seeing?

szepeviktor:
You are right.
It deserves a "close".

I thought those cryptographic values could be - somehow - ignored in a simple way.

PR #4395, opened 2024-12-09
refactor(event cache store): use a single transaction to handle all linked chunk updates at once

bnjbvr:
Instead of one transaction per update. This ensures that if a single update fails, then none is taken into account.

The test triggers a unique constraint violation. Before the patch, it would still contain a single chunk (the first one), while after the patch, it contains none.

Part of #3280.


Hywan (approved):
Looks sane to me! Good catch.

PR #4396, opened 2024-12-09
feat(event cache): add a way to clear all chunks, clear a room's chunks + FFI feature flag

bnjbvr:
This adds a way to:

- clear all the rooms' linked chunk, all at once: this will be useful when disabling the feature flag
- clear a single room's event cache persistent storage: this should be useful for testing purposes, and in case storage is completely borked
- enable a feature flag on an FFI `ClientBuilder` to enable or disable persistent storage. Note: after persistent storage has been enabled, if the embedder decides to disable it, they must explicitly set it to `false` the next time; otherwise the on-disk storage isn't cleared.

The last item isn't very satisfying, but I don't see how to do it otherwise without more involved refactorings. The way persistent storage works right now, is by checking the presence of an event cache store reference in the `RoomEventCacheState` data structure. This is filled at most once, and only once, and never cleared (it's not even clearable). As such, a restart is required for disabling the persistent storage; after restarting is the right time to clear a previous on-disk persisted storage. If we do it sooner than that, then the live `RoomEventCacheState` may keep on trying to store new updates…

Part of #3280.

### DIFF
--- a/bindings/matrix-sdk-ffi/src/client_builder.rs
+++ b/bindings/matrix-sdk-ffi/src/client_builder.rs
@@ -273,7 +273,9 @@ pub struct ClientBuilder {
     decryption_trust_requirement: TrustRequirement,
     request_config: Option<RequestConfig>,
 
-    use_event_cache_persistent_storage: Option<bool>,
+    /// Whether to enable use of the event cache store, for reloading events
+    /// when building timelines et al.
+    use_event_cache_persistent_storage: bool,
 }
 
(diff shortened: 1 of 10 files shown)


Hywan (approved):
Winter clean up.

bnjbvr:
As discussed in private, the last commit changes the semantics so that the feature flag is not optional, and always set to false. It makes it less awkward when used by consumers. (See commit message for details.)

PR #4397, opened 2024-12-09
feat(event cache): improve event cache storage a bit (debug, resilience,…)

bnjbvr:
- commit 1: refactor in preparation for commit 3; instead of rebuilding the linked chunk in the event cache store impl, do it one layer above. It's similar code all the cache store impls had to copy/paste themselves + it will allow us to inspect the chunks whenever something goes wrong.
- commit 2: display a debug string showing the raw state of a linked chunk, which allows to display invalid states, in particular. Use that display whenever rebuilding the linked chunk fails, in the event cache.
- commit 3: when rebuilding a linked chunk fails, clear the storage, and start afresh.
- commit 4: spawn a task when handling updates to storage, so that it can't be interrupted.
  - question for reviewer: this relies on the fact that getting the store lock requires holding onto a real mutex. I'm afraid that if trying to acquire the cross-process lock doesn't mean holding this mutex anymore, in the future, then we might forget about this place, and run into issues. Should I introduce a new lock for that purpose, here?
- commit 5: tweak a log
- commit 6: something spotted while trying the patch in `multiverse`: we get a prev-batch token on every incremental SSS; we only need to store the gap whenever the timeline is limited (aka, the server tells us we are definitely missing events), and not otherwise.

Because of commit 3, we might not even need to have migrations if the format of events data changes over time, which is kinda sloppy but nice? cc @Hywan @poljar

### DIFF
--- a/crates/matrix-sdk-common/src/linked_chunk/mod.rs
+++ b/crates/matrix-sdk-common/src/linked_chunk/mod.rs
@@ -1074,6 +1074,15 @@ pub enum ChunkContent<Item, Gap> {
     Items(Vec<Item>),
 }
 
+impl<Item: Clone, Gap: Clone> Clone for ChunkContent<Item, Gap> {
+    fn clone(&self) -> Self {
+        match self {
+            Self::Gap(gap) => Self::Gap(gap.clone()),
+            Self::Items(items) => Self::Items(items.clone()),
+        }
+    }
+}
+
 /// A chunk is a node in the [`LinkedChunk`].
 pub struct Chunk<const CAPACITY: usize, Item, Gap> {
     /// The previous chunk.
@@ -1433,6 +1442,17 @@ pub struct RawLinkedChunk<Item, Gap> {
 
     /// Link to the next chunk, via its identifier.
     pub next: Option<ChunkIdentifier>,
+}
+
+impl<Item: Clone, Gap: Clone> Clone for RawLinkedChunk<Item, Gap> {
+    fn clone(&self) -> Self {
+        Self {
+            content: self.content.clone(),
+            previous: self.previous,
+            id: self.id,
+            next: self.next,
+        }
+    }
 }
 
 #[cfg(test)]
--- a/crates/matrix-sdk-common/src/linked_chunk/mod.rs
+++ b/crates/matrix-sdk-common/src/linked_chunk/mod.rs
@@ -1064,7 +1064,7 @@ impl<'a, const CAP: usize, Item, Gap> Iterator for Iter<'a, CAP, Item, Gap> {
 }
 
 /// This enum represents the content of a [`Chunk`].
-#[derive(Debug)]
+#[derive(Clone, Debug)]
 pub enum ChunkContent<Item, Gap> {
     /// The chunk represents a gap in the linked chunk, i.e. a hole. It
     /// means that some items are missing in this location.
@@ -1072,15 +1072,6 @@ pub enum ChunkContent<Item, Gap> {
 
     /// The chunk contains items.
     Items(Vec<Item>),
-}
-
-impl<Item: Clone, Gap: Clone> Clone for ChunkContent<Item, Gap> {
-    fn clone(&self) -> Self {
-        match self {
-            Self::Gap(gap) => Self::Gap(gap.clone()),
-            Self::Items(items) => Self::Items(items.clone()),
-        }
-    }
 }
 
 /// A chunk is a node in the [`LinkedChunk`].
@@ -1429,8 +1420,12 @@ impl EmptyChunk {
 }
 
 /// The raw representation of a linked chunk, as persisted in storage.
-#[derive(Debug)]
-pub struct RawLinkedChunk<Item, Gap> {
+///
+/// It may rebuilt into [`Chunk`] and shares the same internal representation,
+/// except that links are materialized using [`ChunkIdentifier`] instead of raw
+/// pointers to the previous and next chunks.
+#[derive(Clone, Debug)]
+pub struct RawChunk<Item, Gap> {
     /// Content section of the linked chunk.
     pub content: ChunkContent<Item, Gap>,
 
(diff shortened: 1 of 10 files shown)


crates/matrix-sdk-common/src/linked_chunk/mod.rs
```diff
@@ -1419,6 +1419,22 @@ impl EmptyChunk {
     }
 }
 
+/// The raw representation of a linked chunk, as persisted in storage.
+#[derive(Debug)]
+pub struct RawLinkedChunk<Item, Gap> {
```
Hywan, line 1433: Nice to get it here.

Question: `Chunk` contains `previous`, `next`, `identiifer` and `content`. So… technically, this type should be `RawChunk`, not `RawLinkedChunk`. Sorry for the rebase it will incur but I think `RawChunk` is the correct naming.

crates/matrix-sdk-common/src/linked_chunk/mod.rs
```diff
@@ -1419,6 +1419,22 @@ impl EmptyChunk {
...
+    /// Content section of the linked chunk.
+    pub content: ChunkContent<Item, Gap>,
+
+    /// Link to the previous chunk, via its identifier.
+    pub previous: Option<ChunkIdentifier>,
+
+    /// Current chunk's identifier.
+    pub id: ChunkIdentifier,
```
Hywan, line 1441: To match the naming of `Chunk`, can we rename `id` to `identifier` please?

crates/matrix-sdk-common/src/linked_chunk/mod.rs
```diff
@@ -1419,6 +1419,22 @@ impl EmptyChunk {
     }
 }
 
+/// The raw representation of a linked chunk, as persisted in storage.
```
Hywan, line 1431: Maybe you can specifiy that this type is like `Chunk` except that ends/links to previous and next chunks are based on `ChunkIdentifier` instead of pointers. That's clearly an important detail.

crates/matrix-sdk-sqlite/src/event_cache_store.rs
```diff
@@ -586,27 +537,42 @@ impl EventCacheStore for SqliteEventCacheStore {
...
-                }
-                ChunkContent::Items(items) => {
-                    builder.push_items(c.previous, c.id, c.next, items);
+        let result = self
+            .acquire()
+            .await?
+            .with_transaction(move |txn| -> Result<_> {
+                let mut items = Vec::new();
+
+                // Use `ORDER BY id` to get a deterministic ordering for testing purposes.
```
Hywan, line 552: Thanks for the comment.

crates/matrix-sdk-common/src/linked_chunk/mod.rs
```diff
@@ -1074,6 +1074,15 @@ pub enum ChunkContent<Item, Gap> {
     Items(Vec<Item>),
 }
 
+impl<Item: Clone, Gap: Clone> Clone for ChunkContent<Item, Gap> {
```
Hywan, line 1077: Doesn't it work with `#[derive(Clone)]` automatically? I thought `derive(Clone)` was adding the bounds for generic parameters automatically.

crates/matrix-sdk-common/src/linked_chunk/mod.rs
```diff
@@ -1435,6 +1444,17 @@ pub struct RawLinkedChunk<Item, Gap> {
     pub next: Option<ChunkIdentifier>,
 }
 
+impl<Item: Clone, Gap: Clone> Clone for RawLinkedChunk<Item, Gap> {
```
Hywan, line 1447: Same here.

crates/matrix-sdk-common/src/linked_chunk/mod.rs
Hywan: Tested here: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=477cdd4969b9e85bb263009447a6acd3

See the macro expand tool result:

```rust
#[derive(Clone)]
struct Foo<A, B> {
    a: A,
    b: B,
}

struct NotClone;

fn test<T: Clone>(value: T) {}

fn main() {
    test(Foo { a: 1, b: 2});
    test(Foo { a: 1, b: NotClone });
}
```

becomes:

```rust
struct Foo<A, B> {
    a: A,
    b: B,
}
#[automatically_derived]
impl<A: ::core::clone::Clone, B: ::core::clone::Clone> ::core::clone::Clone
    for Foo<A, B> {
    #[inline]
    fn clone(&self) -> Foo<A, B> {
        Foo {
            a: ::core::clone::Clone::clone(&self.a),
            b: ::core::clone::Clone::clone(&self.b),
        }
    }
}

// …
```

crates/matrix-sdk/src/event_cache/room/mod.rs
```diff
@@ -611,30 +614,52 @@ mod private {
...
+            Ok(builder.build().map_err(|err| {
+                // Show a debug string representing the known chunks.
+                if tracing::enabled!(tracing::Level::TRACE) {
+                    trace!("couldn't build a linked chunk from the raw parts. Raw chunks:");
+                    for line in raw_chunks_debug_string(raw_chunks) {
+                        trace!("{line}");
+                    }
+                }
+
+                EventCacheStoreError::InvalidData { details: err.to_string() }
```
Hywan, line 642: Shouldn't we erase the storage if we can't deserialize it? What happens if we create a new linked chunk and continue to save it in the storage over a previous one?

Hywan (approved):
I think there is an important renaming, but apart from that, nothing is blocking. Approving this PR to avoid a new review cycle.

crates/matrix-sdk-common/src/linked_chunk/mod.rs
bnjbvr: nice

crates/matrix-sdk-common/src/linked_chunk/mod.rs
bnjbvr: I'd rather go the opposite way, for the sake of conciseness, and rename `identifier` to `id`, to be honest :P but consistency is king, so I'll use `identifier` here :+1:

crates/matrix-sdk-common/src/linked_chunk/mod.rs
bnjbvr: No worries, it's a good point, and I can rename in a separate commit. Thanks.

crates/matrix-sdk/src/event_cache/room/mod.rs
bnjbvr: What you're describing is exactly what happens:

- `try_reload_linked_chunk` fails with an `Err` result
- the caller in `RoomEventCacheState::new()` will log the error, clear storage for the room, and restart with an empty linked chunk

PR #4398, opened 2024-12-10
feat(timeline): Configurable timeline date dividers

stefanceriu:
Expose a timeline configuration option to allow date separators to go in between different days or different months. 
The only real logic change is in the first commit in `DateDividerAdjuster::is_same_date_as` where the configuration option is now taken into account.

This will work together with the newly added `message_filtered_timeline` to power the Element X media gallery and its sections.

This PR also renames all *day* dividers to *date* dividers in the last 2 commits.


crates/matrix-sdk-ui/src/timeline/day_dividers.rs
```diff
@@ -480,6 +481,20 @@ impl DayDividerAdjuster {
             Some(report)
         }
     }
+
+    /// Returns whether the two dates for the given timestamps are the same or not.
+    fn is_same_date_as(
```
andybalaam, line 490: Can we think of a more appropriate name for this function?

The best I can come up with right now is `should_be_separated_by_date_divider` but I would welcome better suggestions.

crates/matrix-sdk-ui/src/timeline/day_dividers.rs
```diff
@@ -855,4 +864,61 @@ mod tests {
         assert!(iter.next().unwrap().is_remote_event());
         assert!(iter.next().is_none());
     }
+
+    #[test]
+    fn test_dayly_divider_mode() {
```
andybalaam, line 873: dayly -> daily

crates/matrix-sdk-ui/src/timeline/day_dividers.rs
```diff
@@ -855,4 +864,61 @@ mod tests {
...
+        let mut items = ObservableVector::new();
+        let mut txn = items.transaction();
+
+        let mut meta = test_metadata();
+
+        txn.push_back(meta.new_timeline_item(event_with_ts(MilliSecondsSinceUnixEpoch(uint!(0)))));
+        txn.push_back(
+            meta.new_timeline_item(event_with_ts(MilliSecondsSinceUnixEpoch(uint!(100000000)))),
+        );
+        txn.push_back(meta.new_timeline_item(event_with_ts(MilliSecondsSinceUnixEpoch::now())));
```
andybalaam, line 883: I am uncomfortable with the use of `now` in a test. I was trying to think of a reason why, but the best I can come up with is what if we want to run the tests on a non-Internet-connected device that does not have a battery-backed clock, so it thinks it's 1970.

Seems unlikely, but is there a good reason why you _want_ to use `now`?

crates/matrix-sdk-ui/src/timeline/day_dividers.rs
```diff
@@ -855,4 +864,61 @@ mod tests {
...
```
andybalaam, line 912: Same comment about `now`

andybalaam (approved):
Looks good; a couple of light suggestions.

Make sure this doesn't get squashed when you merge since the separation you have made between renames and functional changes is really helpful.

crates/matrix-sdk-ui/src/timeline/day_dividers.rs
stefanceriu: 🤦

crates/matrix-sdk-ui/src/timeline/day_dividers.rs
stefanceriu: Haha, but no, there's no particular reason apart from writing conveninence. I'll change it into something more reasonable 👍

crates/matrix-sdk-ui/src/timeline/day_dividers.rs
stefanceriu: I don't think this is always used to `separate` things so how about something very similar like `is_same_date_divider_group_as`?

andybalaam: Yep, sounds good.

PR #4400, opened 2024-12-10
fix(send_queue): Use `MediaFormat::File` when caching attachment thumbnail

zecakeh:
Fix extracted from #4329. 

The `MediaFormat` reflects only the request that would be made to the homeserver. There is no link between the format of the files stored in the media cache and their purpose in an event. `MediaFormat::Thumbnail` means that we request a server-generated thumbnail of a file in the media repository. Since the thumbnail is its own file in the media repository, it makes more sense to use `MediaFormat::File`. See https://github.com/matrix-org/matrix-rust-sdk/pull/4329#discussion_r1873122697 for even more details why this change makes sense.

The fix was simplified a lot here. Instead of requiring a migration, we just minimize the changes in the format of the send queue request in the state store and keep handling the previous case.

This also documents the caching of the media when calling `send_attachment`.

### DIFF
--- a/crates/matrix-sdk/src/send_queue/upload.rs
+++ b/crates/matrix-sdk/src/send_queue/upload.rs
@@ -97,6 +97,11 @@ impl RoomSendQueue {
     /// client's sending queue will be disabled, and it will need to be
     /// manually re-enabled by the caller (e.g. after network is back, or when
     /// something has been done about the faulty requests).
+    ///
+    /// The attachment and its optional thumbnail are stored in the media cache
+    /// and can be retrieved at any time, by calling
+    /// [`Media::get_media_content()`] with the `MediaSource` that can be found
+    /// in the local or remote echo, and using a `MediaFormat::File`.
     #[instrument(skip_all, fields(event_txn))]
     pub async fn send_attachment(
         &self,
--- a/crates/matrix-sdk/src/send_queue/upload.rs
+++ b/crates/matrix-sdk/src/send_queue/upload.rs
@@ -137,20 +137,15 @@ impl RoomSendQueue {
 
             // Process the thumbnail, if it's been provided.
             if let Some(thumbnail) = config.thumbnail.take() {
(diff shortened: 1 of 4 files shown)


crates/matrix-sdk/src/send_queue/upload.rs
```diff
@@ -266,18 +261,20 @@ impl QueueStorage {
             if let Some((info, new_source)) =
                 thumbnail_info.as_ref().zip(sent_media.thumbnail.clone())
             {
-                let from_req =
-                    Media::make_local_thumbnail_media_request(&info.txn, info.height, info.width);
+                // Previously the media request used `MediaFormat::Thumbnail`. Handle this case
+                // for send queue requests that were in the state store before the change.
```
bnjbvr, line 270: state store? or cache store?

crates/matrix-sdk/src/send_queue/upload.rs
```diff
@@ -97,6 +97,11 @@ impl RoomSendQueue {
     /// client's sending queue will be disabled, and it will need to be
     /// manually re-enabled by the caller (e.g. after network is back, or when
     /// something has been done about the faulty requests).
+    ///
+    /// The attachment and its optional thumbnail are stored in the media cache
+    /// and can be retrieved at any time, by calling
+    /// [`Media::get_media_content()`] with the `MediaSource` that can be found
+    /// in the local or remote echo, and using a `MediaFormat::File`.
```
bnjbvr, line 104: <3

bnjbvr (approved):
Thank you, nice serde trickery!

crates/matrix-sdk/src/send_queue/upload.rs
zecakeh: This is about `DependentQueuedRequestKind`, so the state store, right?

bnjbvr: Ah, you're right :facepalm:

PR #4402, opened 2024-12-11
fix(event cache): fix back-pagination ordering in one specific case

bnjbvr:
This started as a refactoring of the event cache's logic for back-paginations, for other purposes, but turns out there was a bug if you were in the following conditions:

- a room has some events
- the room doesn't have a prev-batch token / a gap

In this case, if we ran a back-pagination, the next prev-batch token would be in the wrong position, compared to the events that we just back-paginated. Which means that the next back-pagination would insert events in the wrong place. The integration test I added acts as a regression test, which failed on main.


Hywan (approved):
Looks good to me :-)

PR #4403, opened 2024-12-11
refactor(state store): use a single lock for all memory store accesses

### LINKED ISSUE #3720: in-memory state store: use a single lock for all the data instead of one lock per field

bnjbvr:
The `MemoryStore` implementation of the `StateStore` has grown into a monster, with one lock per field. It's probably overkill, as individual fields don't need fine-grained locks like this; after all, accesses to the store shouldn't be reentrant in general.

Also allows us to get rid of some annoying trace logs, which is the reason why I was reminded of this problem.

Fixes #3720.

Let's see how Complement Crypto likes it…

### DIFF
--- a/crates/matrix-sdk-base/src/store/memory_store.rs
+++ b/crates/matrix-sdk-base/src/store/memory_store.rs
@@ -14,7 +14,7 @@
 
 use std::{
     collections::{BTreeMap, BTreeSet, HashMap},
-    sync::RwLock as StdRwLock,
+    sync::RwLock,
 };
 
 use async_trait::async_trait;
@@ -33,7 +33,7 @@ use ruma::{
     CanonicalJsonObject, EventId, OwnedEventId, OwnedMxcUri, OwnedRoomId, OwnedTransactionId,
     OwnedUserId, RoomId, RoomVersionId, TransactionId, UserId,
 };
-use tracing::{debug, instrument, trace, warn};
+use tracing::{debug, instrument, warn};
 
 use super::{
     send_queue::{ChildTransactionId, QueuedRequest, SentRequestKey},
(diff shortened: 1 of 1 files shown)


crates/matrix-sdk-base/src/store/memory_store.rs
```diff
@@ -47,50 +47,49 @@ use crate::{
...
+    send_queue_events: BTreeMap<OwnedRoomId, Vec<QueuedRequest>>,
+    dependent_send_queue_events: BTreeMap<OwnedRoomId, Vec<DependentQueuedRequest>>,
+}
+
+/// In-memory, non-persistent implementation of the `StateStore`.
+///
+/// Default if no other is configured at startup.
+#[derive(Debug, Default)]
+pub struct MemoryStore {
+    inner: RwLock<MemoryStoreInner>,
```
stefanceriu, line 92: Huh, that's all there is to it, very nice 👏

stefanceriu (approved):
Way better! 🙌

PR #4405, opened 2024-12-11
chore(ffi): avoid hardcoding clang version

richvdh:
Update the workaround for https://github.com/rust-lang/rust/issues/109717 to avoid hardcoding the clang version; instead, run `clang -dumpversion` to figure it out.

While we're there, use the `CC_x86_64-linux-android` env var, which should point to clang, rather than relying on `ANDROID_NDK_HOME` to be set.

### DIFF
--- a/bindings/matrix-sdk-crypto-ffi/build.rs
+++ b/bindings/matrix-sdk-crypto-ffi/build.rs
@@ -1,10 +1,10 @@
-use std::{env, error::Error};
+use std::{env, error::Error, path::PathBuf, process::Command};
 
 use vergen::EmitBuilder;
 
 /// Adds a temporary workaround for an issue with the Rust compiler and Android
 /// in x86_64 devices: https://github.com/rust-lang/rust/issues/109717.
-/// The workaround comes from: https://github.com/mozilla/application-services/pull/5442
+/// The workaround is based on: https://github.com/mozilla/application-services/pull/5442
 ///
 /// IMPORTANT: if you modify this, make sure to modify
 /// [../matrix-sdk-ffi/build.rs] too!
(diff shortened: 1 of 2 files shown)


bindings/matrix-sdk-crypto-ffi/build.rs
```diff
@@ -1,37 +1,53 @@
...
-        );
-        println!("cargo:rustc-link-search={android_ndk_home}/{linux_x86_64_lib_dir}");
+        // Configure rust to statically link against the `libclang_rt.builtins` supplied with clang.
+
+        // cargo-ndk sets CC_x86_64-linux-android to the path to `clang`, within the Android NDK.
+        let cc = PathBuf::from(env::var("CC_x86_64-linux-android").expect("CC_x86_64-linux-android not set"));
+
+        let toolchain_path = cc
+            .ancestors()
+            .nth(2)
```
bnjbvr, line 24: Can you include an example of what `cc` would look like, so we understand a bit better why `nth(2)` is the right thing to do?

bindings/matrix-sdk-crypto-ffi/build.rs
```diff
@@ -1,37 +1,53 @@
...
-        let clang_version =
-            env::var("NDK_CLANG_VERSION").unwrap_or_else(|_| DEFAULT_CLANG_VERSION.to_owned());
-        let linux_x86_64_lib_dir = format!(
-            "toolchains/llvm/prebuilt/{build_os}-x86_64/lib/clang/{clang_version}/lib/linux/"
-        );
-        println!("cargo:rustc-link-search={android_ndk_home}/{linux_x86_64_lib_dir}");
+        // Configure rust to statically link against the `libclang_rt.builtins` supplied with clang.
+
+        // cargo-ndk sets CC_x86_64-linux-android to the path to `clang`, within the Android NDK.
+        let cc = PathBuf::from(env::var("CC_x86_64-linux-android").expect("CC_x86_64-linux-android not set"));
```
bnjbvr, line 20: Can this variable be called `clang_path`?

bindings/matrix-sdk-crypto-ffi/build.rs
```diff
@@ -1,37 +1,53 @@
...
+
+        let clang_version = get_clang_major_version(&cc);
+
+        println!("cargo:rustc-link-search={toolchain_path}/lib/clang/{clang_version}/lib/linux/");
         println!("cargo:rustc-link-lib=static=clang_rt.builtins-x86_64-android");
     }
 }
 
+/// Run the clang binary at `clang_path`, and return its major version number
+fn get_clang_major_version(clang_path: impl AsRef<OsStr>) -> String {
```
bnjbvr, line 37: Since this is called only once, I don't think we need a generic `impl AsRef<...>` parameter here :eyes:

bnjbvr (commented):
Sure :sweat_smile: I'll ask someone who has more Android knowledge to review this, as I'm not sure why the change is needed (supposedly NDK_CLANG_VERSION` is not always set anymore?).

If you wanted to go the extra mile, we could introduce a new tiny crate for building this, that would contain the code for the workaround setup. It seemed fine to duplicate when it was one function-long, but now that it seems to become a bit more involved, sharing code would make sense IMO.

richvdh:
> supposedly `NDK_CLANG_VERSION` is not always set anymore?

I don't think it was ever set, unless you knew about it and set it manually.

richvdh:
> If you wanted to go the extra mile, we could introduce a new tiny crate for building this, that would contain the code for the workaround setup. It seemed fine to duplicate when it was one function-long, but now that it seems to become a bit more involved, sharing code would make sense IMO.

It would, but ugh my life is running short right now. I could do with just shipping this :/

bindings/matrix-sdk-crypto-ffi/build.rs
richvdh: yup. 923fe510d

bindings/matrix-sdk-crypto-ffi/build.rs
richvdh: 923fe510d

bindings/matrix-sdk-crypto-ffi/build.rs
richvdh: 923fe510d

jmartinesp:
> No, the whole point is that CC_x86_64-linux-android is automatically provided by cargo ndk. The intention of this change is very much something that just works most of the time.

Oh, TIL! Then I retract everything I said. I already started a build in the bindings repo to test this out: https://github.com/matrix-org/matrix-rust-components-kotlin/actions/runs/12295378552/job/34312142906

bindings/matrix-sdk-crypto-ffi/build.rs
```diff
@@ -1,37 +1,56 @@
...
-        const DEFAULT_CLANG_VERSION: &str = "18";
-        let clang_version =
-            env::var("NDK_CLANG_VERSION").unwrap_or_else(|_| DEFAULT_CLANG_VERSION.to_owned());
-        let linux_x86_64_lib_dir = format!(
-            "toolchains/llvm/prebuilt/{build_os}-x86_64/lib/clang/{clang_version}/lib/linux/"
+        // Configure rust to statically link against the `libclang_rt.builtins` supplied
+        // with clang.
+
+        // cargo-ndk sets CC_x86_64-linux-android to the path to `clang`, within the
+        // Android NDK.
```
jmartinesp, line 19: I completely missed this comment 🤦 .

jmartinesp (approved):
Approving, since this works as expected. Thanks!

PR #4407, opened 2024-12-12
Ensure that we request the correct state event for the service members to work correctly

poljar:
We forgot that under sliding sync we won't have access to all state events, which meant that #4335 never got working.

The first commit adds the relevant event to the required state for the UI crate.

The second commit fixes another edge case in the case where we homeserver does provide a room summary containing service members in the heroes and thus in the joined members count.

### DIFF
--- a/crates/matrix-sdk-base/src/rooms/normal.rs
+++ b/crates/matrix-sdk-base/src/rooms/normal.rs
@@ -119,6 +119,23 @@ bitflags! {
     }
 }
 
+/// The result of a room summary computation.
+///
+/// If the homeserver does not provide a room summary, we perform a best-effort
+/// computation to generate one ourselves. If the homeserver does provide the
+/// summary, we augment it with additional information about the service members
+/// in the room.
+struct ComputedSummary {
+    /// The list of display names that will be used to calculate the room
+    /// display name.
+    heroes: Vec<String>,
+    /// The number of joined service members in the room.
+    num_service_members: u64,
+    /// The number of joined and invited members, not including any service
+    /// members.
+    num_joined_invited_guess: Option<u64>,
+}
+
 impl Default for RoomInfoNotableUpdateReasons {
     fn default() -> Self {
         Self::empty()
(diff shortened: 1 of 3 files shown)


crates/matrix-sdk-base/src/rooms/normal.rs
```diff
@@ -693,11 +697,20 @@ impl Room {
     /// [`RoomSummary`].
     ///
     /// Returns the display names as a list of strings.
-    async fn extract_heroes(&self, heroes: &[RoomHero]) -> StoreResult<Vec<String>> {
+    async fn extract_heroes(&self, heroes: &[RoomHero]) -> StoreResult<(Vec<String>, u64)> {
```
bnjbvr, line 700: Can you tweak the doc comment and indicate what the second return value is?

crates/matrix-sdk-base/src/rooms/normal.rs
```diff
@@ -721,15 +734,15 @@ impl Room {
...
+        Ok((names, num_service_members))
     }
 
     /// Compute the room summary with the data present in the store.
     ///
     /// The summary might be incorrect if the database info is outdated.
     ///
     /// Returns a `(heroes_names, num_joined_invited)` tuple.
-    async fn compute_summary(&self) -> StoreResult<(Vec<String>, u64)> {
+    async fn compute_summary(&self) -> StoreResult<(Vec<String>, u64, u64)> {
```
bnjbvr, line 745: Same here, can you update the above doc comment? (or return a tiny struct here, since this is returning 3 values that would make sense to me)

bnjbvr (approved):
Thanks!

crates/matrix-sdk-base/src/rooms/normal.rs
poljar: Something like this: https://github.com/matrix-org/matrix-rust-sdk/pull/4407/commits/bb0baf0b0d2f803d72c312de882ec319b1449169?

I made both paths return the struct, now I wonder if it would be simpler if the joined/invted count manipulation should just go into each subfunction as well.

bnjbvr: Yes (to both points) :+1:

PR #4410, opened 2024-12-12
fix(event cache store): use immediate transactions in `handle_linked_chunk_updates`

bnjbvr:
This is unfortunate, but we're hitting a bad case of a read transaction upgrading to a write transaction, and that can plain fail if there's any other write transaction already happening in the background.

To work around this, we need to start the transaction in write mode, so start with the update statements themselves. This means some work is duplicated (fetching a linked chunk), but this should be fine since there's an index on the pair (chunk id, room id).

### DIFF
--- a/crates/matrix-sdk-sqlite/src/event_cache_store.rs
+++ b/crates/matrix-sdk-sqlite/src/event_cache_store.rs
@@ -27,7 +27,7 @@ use matrix_sdk_base::{
 };
 use matrix_sdk_store_encryption::StoreCipher;
 use ruma::{MilliSecondsSinceUnixEpoch, RoomId};
-use rusqlite::{OptionalExtension, Transaction};
+use rusqlite::{OptionalExtension, Transaction, TransactionBehavior};
 use tokio::fs;
 use tracing::{debug, trace};
 
@@ -378,9 +378,7 @@ impl EventCacheStore for SqliteEventCacheStore {
         let room_id = room_id.to_owned();
         let this = self.clone();
 
-        self.acquire()
-            .await?
-            .with_transaction(move |txn| -> Result<_, Self::Error> {
+        with_immediate_transaction(self.acquire().await?, move |txn| {
                 for up in updates {
                     match up {
                         Update::NewItemsChunk { previous, new, next } => {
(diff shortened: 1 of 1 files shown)


crates/matrix-sdk-sqlite/src/event_cache_store.rs
```diff
@@ -532,9 +539,11 @@ impl EventCacheStore for SqliteEventCacheStore {
                     }
                 }
 
+                txn.commit()?;
+
                 Ok(())
             })
-        .await?;
+        .await.unwrap()?;
```
poljar, line 546: What's that unwrap about?

crates/matrix-sdk-sqlite/src/event_cache_store.rs
```diff
@@ -380,7 +380,14 @@ impl EventCacheStore for SqliteEventCacheStore {
 
         self.acquire()
             .await?
-            .with_transaction(move |txn| -> Result<_, Self::Error> {
+            .interact(move |conn| -> Result<_, Self::Error> {
+                // Start the transaction in IMMEDIATE mode since all updates may cause writes, to
+                // avoid read transactions upgrading to write mode and causing SQLITE_BUSY errors.
+                // See also: https://www.sqlite.org/lang_transaction.html#deferred_immediate_and_exclusive_transactions
+                conn.set_transaction_behavior(TransactionBehavior::Immediate);
```
poljar, line 387: Does the thing go back to the default after we call `drop()` on the connection?

bnjbvr: I haven't seen any sign of such thing, but since most accesses are write anyways (even reading from the media cache causes a write access to update the last_touch time), I figured it was OK. That being said, it's not very future-proof, and there are also a few read transactions. So I think I can manually set it back to `Deferred` at the end.

crates/matrix-sdk-sqlite/src/event_cache_store.rs
bnjbvr: Copied from the implementation of `with_transaction()`; the only valid error this can return is because an underlying spawned task has panick'd, which is a valid reason to panic ourselves.

crates/matrix-sdk-sqlite/src/event_cache_store.rs
poljar: Yeah, let's set it back, otherwise me might slow down things or run into more database locked errors.

poljar (approved):

poljar (approved):
Still looks good.

PR #4411, opened 2024-12-12
feat(event cache): have a debug representation of a linked chunk and use it in multiverse

bnjbvr:
I think that this debug representation may show useful to understand issues. Just from using it a few minutes, I've identified a few issues related to storage:

- we shouldn't have empty items chunks; they should be removed, or even better, reused.
- we may have multiple prev-batch token that represent the same events, e.g. if we sync'd a room with timeline_limit=1, then with timeline_limit=20. Not sure if there's anything obvious to do about that.
  - edit: yes there is! do not store a prev-batch token when all the events (at least 1) returned from the backpagination are known. Should be a sufficiently good heuristic 🧠 

image

### DIFF
--- a/bindings/matrix-sdk-ffi/src/room.rs
+++ b/bindings/matrix-sdk-ffi/src/room.rs
@@ -933,6 +933,13 @@ impl Room {
 
         Ok(handle)
     }
+
+    /// Return a debug representation for the internal room events data
+    /// structure, one line per entry in the resulting vector.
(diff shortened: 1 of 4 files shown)


stefanceriu (approved):
Super nice! Any chance we can expose this through FFI? We could dump the contents into a separate screen like we do for message sources.

bnjbvr:
@stefanceriu Done in the last commit; this is a simple change, so not re-requesting a review just for this :) Thanks for the review!

PR #4412, opened 2024-12-12
Use a type alias to allow bindings to take advantage of custom types

zzorba:
This PR adds a new type alias to the matrix-sdk-ffi crate called Date64.  This is aliased to u64 for backwards compatibility (so no change to Swift/Kotlin as is), but as a distinct type it will allow uniffi integrations to customize the unpacking of the type automatically: https://mozilla.github.io/uniffi-rs/latest/udl/custom_types.html

Using this, it is possible to make the uniffi layer automatically convert these from u64 -> an appropriate system Date class in the generated code layer.  

Today a `u64` type can mean either a count or a date, so without an alias implementing a customType would not be possible.  At the very least, the type helps improve the documentation, even if individual bindings choose not to adopt this in their UDL configs.

Signed-off-by: Daniel Salinas

### DIFF
--- a/bindings/matrix-sdk-ffi/src/utils.rs
+++ b/bindings/matrix-sdk-ffi/src/utils.rs
@@ -15,8 +15,19 @@
 use std::{mem::ManuallyDrop, ops::Deref};
 
 use async_compat::TOKIO1 as RUNTIME;
-use ruma::UInt;
+use ruma::{MilliSecondsSinceUnixEpoch, UInt};
 use tracing::warn;
+
+#[derive(Debug, Clone)]
+pub struct Timestamp(u64);
+
+impl From<MilliSecondsSinceUnixEpoch> for Timestamp {
+    fn from(date: MilliSecondsSinceUnixEpoch) -> Self {
+        Self(date.0.into())
+    }
+}
+
+uniffi::custom_newtype!(Timestamp, u64);
 
 pub(crate) fn u64_to_uint(u: u64) -> UInt {
     UInt::new(u).unwrap_or_else(|| {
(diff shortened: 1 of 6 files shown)


zzorba:
I primarily want this for the react-native-matrix-sdk (Typescript), but I believe it is possible to add custom unpackers to Kotlin/Swift as well.  That would of course be a breaking API change for any users of this library, the current iteration of this should fallback to u64 automatically.

jplatte:
Small nitpick about the PR description, it talks about a _marker trait_ but there is no new trait involved here, I think you meant "newtype"?

bindings/matrix-sdk-ffi/src/utils.rs
```diff
@@ -15,9 +15,20 @@
 use std::{mem::ManuallyDrop, ops::Deref};
 
 use async_compat::TOKIO1 as RUNTIME;
-use ruma::UInt;
+use ruma::{MilliSecondsSinceUnixEpoch, UInt};
 use tracing::warn;
 
+#[derive(Debug, Clone)]
+pub struct Date64(u64);
```
Hywan, line 22: I would rename `Date64` to `Timestamp` since it represents a timestamp everywhere this new type is used.

Hywan (dismissed):
Thanks, that's a nice improvement! I'm suggesting a renaming but everything else is nice!

bindings/matrix-sdk-ffi/src/utils.rs
zzorba: Amended the PR with this new name, agree its much clearer.

poljar (approved):
Review comments were addressed and the PR seems fine to me as well.

PR #4414, opened 2024-12-13
feat(room): Separate `RoomState::Ban` from `RoomState::Left`.

jmartinesp:
This is needed to tell apart rooms in left and banned state in places like `RoomInfo` or `RoomPreview`.

The banned rooms will still count as left rooms in the sync processes.

Signed-off-by:

### DIFF
--- a/crates/matrix-sdk-base/src/rooms/normal.rs
+++ b/crates/matrix-sdk-base/src/rooms/normal.rs
@@ -225,14 +225,14 @@ pub enum RoomState {
     Invited,
     /// The room is in a knocked state.
     Knocked,
+    /// The room is in a banned state.
+    Banned,
 }
 
 impl From<&MembershipState> for RoomState {
     fn from(membership_state: &MembershipState) -> Self {
-        // We consider Ban, Knock and Leave to be Left, because they all mean we are not
-        // in the room.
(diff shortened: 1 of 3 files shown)


crates/matrix-sdk-base/src/rooms/normal.rs
```diff
@@ -1919,6 +1928,12 @@ impl RoomStateFilter {
         if self.contains(Self::INVITED) {
             states.push(RoomState::Invited);
         }
+        if self.contains(Self::KNOCKED) {
+            states.push(RoomState::Knocked);
+        }
```
jmartinesp, line 1933: Oops, this was never added before. I added a test including these 2 new cases.

crates/matrix-sdk-base/src/rooms/normal.rs
```diff
@@ -218,14 +218,14 @@ pub enum RoomState {
     Invited,
     /// The room is in a knocked state.
     Knocked,
+    /// The room is in a banned state.
+    Ban,
```
bnjbvr, line 222: Can you name this `Banned`, for consistency with the other states?

bnjbvr (approved):

Issue #4185, opened 2024-10-30
Support Device Dehydration new format


BillCarsonFr:
As per new format defined [here](https://github.com/element-hq/element-meta/issues/2602)
Vodozemac will add [support for serialization/deserialization](https://github.com/matrix-org/vodozemac/issues/196) 

Rust crypto crate already has an enum for the dehydrated device, so this needs to be expanded to support the new format.

Part of https://github.com/element-hq/element-meta/issues/2601

PR #4421, opened 2024-12-16
Use the dehydrated device format implemented by vodozemac

uhoreg:
Depends on https://github.com/matrix-org/vodozemac/pull/199 and https://github.com/ruma/ruma/pull/1987

Fixes https://github.com/matrix-org/matrix-rust-sdk/issues/4185

Signed-off-by:


crates/matrix-sdk-crypto/src/machine/tests/mod.rs
```diff
@@ -1322,8 +1322,8 @@ async fn test_unsigned_decryption() {
 
     // Encrypt a second message, an edit.
     let second_message_text = "This is the ~~original~~ edited message";
-    let second_message_content = RoomMessageEventContent::text_plain(second_message_text)
-        .make_replacement(first_message, None);
+    let second_message_content =
+        RoomMessageEventContent::text_plain(second_message_text).make_replacement(first_message);
```
uhoreg, line 1326: This change is due to https://github.com/ruma/ruma/commit/cae74897e6992833ac148ee41b08400a062f0820

uhoreg:
Looks like the CI failures are due to https://github.com/ruma/ruma/commit/cae74897e6992833ac148ee41b08400a062f0820 since I'm pulling in a newer version of ruma. I had fixed one instance, but other instances need to be fixed too.

zecakeh:
Since we want to avoid making breaking releases too often in Ruma, I would recommend using a branch based on the [ruma-0.12](https://github.com/ruma/ruma/tree/ruma-0.12) branch for now to avoid breaking changes. We will backport all non-breaking changes to it, allowing us to make patch releases whenever the SDK needs to make a release.

zecakeh:
I created #4477 to upgrade Ruma with the necessary commit from https://github.com/ruma/ruma/pull/1987.

Cargo.toml
```diff
@@ -149,6 +149,7 @@ tracing-core = { git = "https://github.com/element-hq/tracing.git", rev = "ca943
 tracing-subscriber = { git = "https://github.com/element-hq/tracing.git", rev = "ca9431f74d37c9d3b5e6a9f35b2c706711dab7dd" }
 tracing-appender = { git = "https://github.com/element-hq/tracing.git", rev = "ca9431f74d37c9d3b5e6a9f35b2c706711dab7dd" }
 paranoid-android = { git = "https://github.com/element-hq/paranoid-android.git", rev = "69388ac5b4afeed7be4401c70ce17f6d9a2cf19b" }
+vodozemac = { git = "https://github.com/uhoreg/vodozemac.git", branch = "dehydration_format" }
```
uhoreg, line 152: @poljar do we need to make a new release of vodozemac, or should I just set the vodozemac dependency to a specific git rev?

poljar: You can pin to the git rev, I'll make a release when we need it for the SDK release.

uhoreg: I've pinned to the git rev.  What do I need to do to make https://github.com/matrix-org/matrix-rust-sdk/actions/runs/12876768825/job/35900251251?pr=4421 happy?

andybalaam (approved):
Looks plausible :-) Thanks!

poljar (approved):

poljar:
Do we want to add a changelog entry for this? Seems to be significant enough. I'll merge this now but open a PR to add a changelog entry if you agree.

uhoreg:
> Do we want to add a changelog entry for this? Seems to be significant enough. I'll merge this now but open a PR to add a changelog entry if you agree.

Yes, probably.  I always forget to add changelog entries.

PR #4422, opened 2024-12-17
feat(room): add `Room::remove_outdated_seen_knock_requests_ids` method

jmartinesp:
In a previous PR (https://github.com/matrix-org/matrix-rust-sdk/pull/4338) we introduced the concept of a 'seen' knock request id, that is, a member event with knock membership that the current user has seen and decided to explicitly ignore so it can be filtered out in the future in the clients and won't clutter the UI. However, if left alone this list will only grow indefinitely, so we need a way to observe the changes in knock room members in the room and remove outdated seen request ids.

Take this PR as a follow-up of the one mentioned above.

A new `matrix_sdk_base::room_member_updates_sender` field was added too, which will broadcast when a room members update is received, both when loading the room members from the HS and when new member state events are received in a sync.

On `fn Room::subscribe_to_knock_requests` a new cleanup task is also spawned to listen to these broadcast and call the `fn Room::remove_outdated_seen_knock_requests_ids`, which will compare the existing seen knock request ids with the room members having a knock membership state in the room and discard the seen knock request ids that are no longer valid.

Signed-off-by:

### DIFF
--- a/crates/matrix-sdk-base/src/client.rs
+++ b/crates/matrix-sdk-base/src/client.rs
@@ -73,7 +73,7 @@ use crate::{
     event_cache::store::EventCacheStoreLock,
     response_processors::AccountDataProcessor,
     rooms::{
-        normal::{RoomInfoNotableUpdate, RoomInfoNotableUpdateReasons},
+        normal::{RoomInfoNotableUpdate, RoomInfoNotableUpdateReasons, RoomMembersUpdate},
         Room, RoomInfo, RoomState,
     },
     store::{
@@ -983,6 +983,9 @@ impl BaseClient {
         let mut new_rooms = RoomUpdates::default();
         let mut notifications = Default::default();
 
+        let mut updated_members_in_room: BTreeMap<OwnedRoomId, BTreeSet<OwnedUserId>> =
+            BTreeMap::new();
+
         for (room_id, new_info) in response.rooms.join {
             let room = self.store.get_or_create_room(
                 &room_id,
@@ -1010,6 +1013,8 @@ impl BaseClient {
                     &mut ambiguity_cache,
                 )
                 .await?;
+
+            updated_members_in_room.insert(room_id.to_owned(), user_ids.clone());
 
             for raw in &new_info.ephemeral.events {
                 match raw.deserialize() {
@@ -1252,6 +1257,13 @@ impl BaseClient {
         // above. Oh well.
         new_rooms.update_in_memory_caches(&self.store).await;
 
+        for (room_id, member_ids) in updated_members_in_room {
+            if let Some(room) = self.get_room(&room_id) {
+                let _ =
+                    room.room_member_updates_sender.send(RoomMembersUpdate::Partial(member_ids));
+            }
+        }
+
         info!("Processed a sync response in {:?}", now.elapsed());
 
         let response = SyncResponse {
@@ -1400,6 +1412,8 @@ impl BaseClient {
 
         self.store.save_changes(&changes).await?;
         self.apply_changes(&changes, Default::default());
+
+        let _ = room.room_member_updates_sender.send(RoomMembersUpdate::FullReload);
 
         Ok(())
     }
--- a/crates/matrix-sdk-base/src/rooms/normal.rs
+++ b/crates/matrix-sdk-base/src/rooms/normal.rs
@@ -15,7 +15,7 @@
 #[cfg(all(feature = "e2e-encryption", feature = "experimental-sliding-sync"))]
 use std::sync::RwLock as SyncRwLock;
 use std::{
-    collections::{BTreeMap, HashSet},
+    collections::{BTreeMap, BTreeSet, HashSet},
     mem,
     sync::{atomic::AtomicBool, Arc},
 };
@@ -1222,20 +1222,60 @@ impl Room {
         }
 
         let current_seen_events_guard = self.get_write_guarded_current_knock_request_ids().await?;
-        // We're not calling `get_seen_join_request_ids` here because we need to keep
-        // the Mutex's guard until we've updated the data
         let mut current_seen_events = current_seen_events_guard.clone().unwrap_or_default();
 
         current_seen_events.extend(event_to_user_ids);
 
         self.update_seen_knock_request_ids(current_seen_events_guard, current_seen_events).await?;
 
-        self.store
-            .set_kv_data(
-                StateStoreDataKey::SeenKnockRequests(self.room_id()),
-                StateStoreDataValue::SeenKnockRequests(current_seen_events),
-            )
-            .await?;
+        Ok(())
+    }
+
+    /// Removes the seen knock request ids that are no longer valid given the
+    /// current room members.
+    pub async fn remove_outdated_seen_knock_requests_ids(&self) -> StoreResult<()> {
+        let current_seen_events_guard = self.get_write_guarded_current_knock_request_ids().await?;
+        let mut current_seen_events = current_seen_events_guard.clone().unwrap_or_default();
+
+        // Get and deserialize the member events for the seen knock requests
+        let keys: Vec<OwnedUserId> = current_seen_events.values().map(|id| id.to_owned()).collect();
+        let raw_member_events: Vec<RawMemberEvent> =
+            self.store.get_state_events_for_keys_static(self.room_id(), &keys).await?;
+        let member_events = raw_member_events
+            .into_iter()
+            .map(|raw| raw.deserialize())
+            .collect::<Result<Vec<MemberEvent>, _>>()?;
+
+        let mut ids_to_remove = Vec::new();
+
+        for (event_id, user_id) in current_seen_events.iter() {
+            // Check the seen knock request ids against the current room member events for
+            // the room members associated to them
+            let matching_member = member_events.iter().find(|event| event.user_id() == user_id);
+
+            if let Some(member) = matching_member {
+                let member_event_id = member.event_id();
+                // If the member event is not a knock or it's different knock, it's outdated
+                if *member.membership() != MembershipState::Knock
+                    || member_event_id.is_some_and(|id| id != event_id)
+                {
+                    ids_to_remove.push(event_id.to_owned());
+                }
+            } else {
+                ids_to_remove.push(event_id.to_owned());
+            }
+        }
+
+        // If there are no ids to remove, do nothing
+        if ids_to_remove.is_empty() {
+            return Ok(());
+        }
+
+        for event_id in ids_to_remove {
+            current_seen_events.remove(&event_id);
+        }
+
+        self.update_seen_knock_request_ids(current_seen_events_guard, current_seen_events).await?;
 
         Ok(())
     }
@@ -2101,7 +2141,6 @@ mod tests {
     use std::{
         collections::BTreeSet,
         ops::{Not, Sub},
-        pin::pin,
         str::FromStr,
         sync::Arc,
         time::Duration,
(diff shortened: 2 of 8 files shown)


crates/matrix-sdk-base/src/rooms/normal.rs
```diff
@@ -262,6 +265,15 @@ fn heroes_filter<'a>(
     move |user_id| user_id != own_user_id && !member_hints.service_members.contains(user_id)
 }
 
+/// The kind of room member updates that just happened.
+#[derive(Debug, Clone)]
+pub enum RoomMembersUpdateKind {
+    /// The whole list room members was reloaded.
+    FullReload,
+    /// A few members were updated, their user ids are included.
+    Partial(BTreeSet<OwnedUserId>),
```
jmartinesp, line 274: While this distinction is not used at the moment, I hope it'll make building a live room members observable in the future a bit easier.

crates/matrix-sdk/src/event_handler/mod.rs
```diff
@@ -1167,4 +1167,79 @@ mod tests {
 
         Ok(())
     }
+
+    #[async_test]
+    #[allow(dependency_on_unit_never_type_fallback)]
```
bnjbvr, line 1172: Can we get rid of it, or comment why it's needed?

crates/matrix-sdk-base/src/client.rs
```diff
@@ -1252,6 +1257,16 @@ impl BaseClient {
         // above. Oh well.
         new_rooms.update_in_memory_caches(&self.store).await;
 
+        for (room_id, member_ids) in updated_members_in_room {
+            if let Some(room) = self.get_room(&room_id) {
+                if let Err(err) =
+                    room.room_member_updates_sender.send(RoomMembersUpdateKind::Partial(member_ids))
+                {
+                    warn!("Failed to broadcast partial room members update: {}", err);
+                }
```
bnjbvr, line 1266: Let's not warn anything, since the only error that can happen is that there's no one listening at this point:

```suggestion
                let _ = room.room_member_updates_sender.send(RoomMembersUpdateKind::Partial(member_ids));
```

crates/matrix-sdk-base/src/client.rs
```diff
@@ -1401,6 +1416,10 @@ impl BaseClient {
         self.store.save_changes(&changes).await?;
         self.apply_changes(&changes, Default::default());
 
+        if let Err(err) = room.room_member_updates_sender.send(RoomMembersUpdateKind::FullReload) {
+            warn!("Failed to broadcast full room members reload: {err}");
+        }
```
bnjbvr, line 1421: Ditto here.

crates/matrix-sdk-base/src/rooms/normal.rs
```diff
@@ -262,6 +265,15 @@ fn heroes_filter<'a>(
     move |user_id| user_id != own_user_id && !member_hints.service_members.contains(user_id)
 }
 
+/// The kind of room member updates that just happened.
+#[derive(Debug, Clone)]
+pub enum RoomMembersUpdateKind {
```
bnjbvr, line 270: Could this be called `RoomMembersUpdate` instead?

crates/matrix-sdk-base/src/rooms/normal.rs
```diff
@@ -297,6 +309,7 @@ impl Room {
             ))),
             room_info_notable_update_sender,
             seen_knock_request_ids_map: SharedObservable::new_async(None),
+            room_member_updates_sender: broadcast::channel(10).0,
```
bnjbvr, line 312: Can you use local variables instead? It's really not clear what the 0 is, so before assigning the sender here, you'd have a line above like `let (sender, _receiver) = broadcast::channel(10);`, and then use `sender` here.

crates/matrix-sdk/src/event_handler/mod.rs
bnjbvr, line 1: Thanks for adding this test! Can it go into a separate commit/PR please?

crates/matrix-sdk/src/room/mod.rs
```diff
@@ -3241,6 +3248,21 @@ impl Room {
 
         let mut room_info_stream = self.subscribe_info();
 
+        // Spawn a task that will clean up the seen knock request ids when updated room
+        // members are received
+        let clear_seen_ids_handle = spawn({
+            let this = this.clone();
```
bnjbvr, line 3254:
```suggestion
            let this = self.clone();
```

crates/matrix-sdk/src/room/mod.rs
```diff
@@ -3241,6 +3248,21 @@ impl Room {
...
+            async move {
+                let mut member_updates_stream = this.room_member_updates_sender.subscribe();
+                while let Ok(_members_update) = member_updates_stream.recv().await {
```
bnjbvr, line 3257:
```suggestion
                while member_updates_stream.recv().await.is_ok() {
```

crates/matrix-sdk-base/src/rooms/normal.rs
```diff
@@ -174,6 +174,9 @@ pub struct Room {
     /// user has marked as seen so they can be ignored.
     pub seen_knock_request_ids_map:
         SharedObservable<Option<BTreeMap<OwnedEventId, OwnedUserId>>, AsyncLock>,
+
+    /// A sender that will notify receivers when room member updates happen.
+    pub room_member_updates_sender: broadcast::Sender<RoomMembersUpdateKind>,
```
bnjbvr, line 179: Since this is `pub`, it's part of the public API, and it must be tested that it works correctly when subscribing :smiling_imp:

crates/matrix-sdk-base/src/rooms/normal.rs
```diff
@@ -1233,6 +1246,59 @@ impl Room {
...
+        );
+
+        let mut current_seen_events_guard = self.seen_knock_request_ids_map.write().await;
+        // We're not calling `get_seen_join_request_ids` here because we need to keep
+        // the Mutex's guard until we've updated the data
+        let mut current_seen_events = if current_seen_events_guard.is_none() {
+            self.load_cached_knock_request_ids().await?
+        } else {
+            current_seen_events_guard.clone().unwrap()
+        };
```
bnjbvr, line 1264: Think this is the third duplication of this piece of code, can we refactor it into a small helper?

crates/matrix-sdk-base/src/rooms/normal.rs
```diff
@@ -1233,6 +1246,59 @@ impl Room {
...
+
+        // If there is nothing to compare the current seen knock ids to, return early
+        if not_knocked_user_ids.is_empty() {
+            return Ok(());
+        }
```
bnjbvr, line 1269: Can you put this code *before* we take the `seen_knock_request_ids_map` mutex?

crates/matrix-sdk-base/src/rooms/normal.rs
```diff
@@ -1233,6 +1246,59 @@ impl Room {
...
+
+        let mut to_remove = Vec::new();
+        for (event_id, user_id) in current_seen_events.iter() {
+            if not_knocked_user_ids.contains(user_id) {
+                to_remove.push(event_id.to_owned());
+            }
+        }
```
bnjbvr, line 1276: I think the approach you've implemented works, but it does a lot of work when the room is large and when there's few members who knocked. Instead, could we iterate over all the user ids from the current_seen_events knock events, get their current membership from the store, and remove them if their own membership has changed?

Now that will mean that instead of loading N memberships (when N is the number of members in the room), we'll only load as many as the number of members who knocked.

bnjbvr (changes requested):

crates/matrix-sdk-base/src/rooms/normal.rs
```diff
@@ -3718,4 +3733,27 @@ mod tests {
...
+        let client = logged_in_base_client(None).await;
+        let room = client.get_or_create_room(room_id!("!a:b.c"), RoomState::Joined);
+
+        let mut receiver = room.room_member_updates_sender.subscribe();
+
+        assert!(receiver.is_empty());
+
+        room.room_member_updates_sender
+            .send(RoomMembersUpdate::FullReload)
+            .expect("broadcasting a room members update failed");
```
bnjbvr, line 3806: I don't think we should send something manually; instead, we should test that the described behavior (will be pinged upon /members or incremental room member update) is correct. So we'd need two tests:

- when I'm calling /members, then I get a `FullReload`
- when I'm receiving new member info from sync, then I get `Partial` with the correct user id set

Does it make sense?

crates/matrix-sdk-base/src/rooms/normal.rs
```diff
@@ -1230,12 +1230,56 @@ impl Room {
...
-            .await?;
+        Ok(())
+    }
+
+    /// Removes the seen knock request ids that are no longer valid given the
+    /// current room members.
+    pub async fn remove_outdated_seen_knock_requests_ids(&self) -> StoreResult<()> {
+        let current_seen_events_guard = self.get_write_guarded_current_knock_request_ids().await?;
+        // We're not calling `get_seen_join_request_ids` here because we need to keep
+        // the Mutex's guard until we've updated the data
```
bnjbvr, line 1241: Think this comment can go now?

crates/matrix-sdk-base/src/rooms/normal.rs
```diff
@@ -1230,12 +1230,56 @@ impl Room {
...
+        let mut current_seen_events = current_seen_events_guard.clone().unwrap_or_default();
+
+        // Get and deserialize the member events for the seen knock requests
+        let keys: Vec<OwnedUserId> = current_seen_events.values().map(|id| id.to_owned()).collect();
+        let raw_member_events: Vec<RawMemberEvent> =
+            self.store.get_state_events_for_keys_static(self.room_id(), &keys).await?;
+        let member_events = raw_member_events
+            .into_iter()
+            .map(|raw| raw.deserialize())
+            .collect::<Result<Vec<MemberEvent>, _>>()?;
```
bnjbvr, line 1251: It's a bit unfortunate that we have to load all the member events from the database systematically, only to filter them out later. The base crate has `BaseRoom::get_member(user_id)` and `StateStore::get_member_event(user_id)`, so we can do better. In pseudo-code:

```
for each (event_id, user_id) in current_seen_events {
  let member = store.get_member(room_id, user_id);
  if member.state is not knocked {
    remove it from current_seen_events
  }
}
```

And that way, we don't have to preload all the member events from the database. Does it make sense?

bnjbvr (changes requested):
Thanks for the splitting of commits, it helps a lot with the reviewing :pray: 

I added a suggestion for not having to reload all member events from the database, as I think it can work.

crates/matrix-sdk-base/src/rooms/normal.rs
jmartinesp: But wouldn't that mean we now need to do N queries instead of a single one? I had that code before (with `store.get_member_event`) and I thought it wasn't as performant as it could be:
- With `store.get_member_event` you make several queries.
- With `room.get_member` you make several queries and also load the presence and other data that's not necessary at this point.

bnjbvr: > The keys passed are the user ids of the seen knock requests, so I don't think we're doing any extra work vs loading the members one by one.

Alright, I missed that, then we should be good to go :+1:

jmartinesp:
https://github.com/matrix-org/matrix-rust-sdk/pull/4422/commits/511328e65daf8e05f548f0f42b0598aa5cd05bc8 should take care of those issues. Sorry I force pushed, I realised I had messed up a doc comment just after uploading the new commit.

bnjbvr (approved):
Thanks, looks great!

PR #4423, opened 2024-12-17
feat(event cache): don't replace a gap chunk by an empty items chunks

bnjbvr:
It was simpler to add a new function rather than trying to fiddle with `replace_gap_at`, especially in terms of the return type of this function. After a few back-and-forth, they're close enough that we could probably merge them together, if you have strong opinions about it.

This was the only place where we'd create an empty chunk, so with this, the storage doesn't keep empty chunks anymore \o/

Part of #3280.

### DIFF
--- a/crates/matrix-sdk-common/src/linked_chunk/mod.rs
+++ b/crates/matrix-sdk-common/src/linked_chunk/mod.rs
@@ -634,6 +634,47 @@ impl<const CAP: usize, Item, Gap> LinkedChunk<CAP, Item, Gap> {
         Ok(())
     }
 
+    /// Remove a gap with the given identifier.
+    ///
+    /// This returns the next insert position, viz. the start of the next
+    /// chunk, if any, or none if there was no next chunk.
+    pub fn remove_gap_at(
+        &mut self,
+        chunk_identifier: ChunkIdentifier,
+    ) -> Result<Option<Position>, Error> {
+        let chunk = self
+            .links
+            .chunk_mut(chunk_identifier)
+            .ok_or(Error::InvalidChunkIdentifier { identifier: chunk_identifier })?;
+
+        if chunk.is_items() {
+            return Err(Error::ChunkIsItems { identifier: chunk_identifier });
+        };
+
+        let next = chunk.next;
+
+        chunk.unlink(&mut self.updates);
+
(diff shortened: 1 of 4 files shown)


crates/matrix-sdk-common/src/linked_chunk/mod.rs
```diff
@@ -634,6 +634,45 @@ impl<const CAP: usize, Item, Gap> LinkedChunk<CAP, Item, Gap> {
...
+            return Err(Error::ChunkIsItems { identifier: chunk_identifier });
+        };
+
+        let next = chunk.next;
+
+        chunk.unlink(&mut self.updates);
+
+        let chunk_ptr = chunk.as_ptr();
+
+        debug_assert!(chunk.is_first_chunk().not(), "A gap cannot be the first chunk");
```
Hywan, line 660: Why this `debug_assert!` is necessary here?

crates/matrix-sdk-common/src/linked_chunk/mod.rs
```diff
@@ -634,6 +634,45 @@ impl<const CAP: usize, Item, Gap> LinkedChunk<CAP, Item, Gap> {
         Ok(())
     }
 
+    /// Remove a gap with the given identifier.
+    ///
+    /// This returns the next insert position, viz. the start of the next
```
Hywan, line 639: TIL https://fr.wiktionary.org/wiki/viz#Breton

crates/matrix-sdk/src/event_cache/room/events.rs
```diff
@@ -166,7 +166,11 @@ impl RoomEvents {
         self.remove_events(duplicated_event_ids);
 
         // Replace the gap by new events.
-        self.chunks.replace_gap_at(unique_events, gap_identifier)
+        if unique_events.is_empty() {
```
Hywan, line 169: Maybe explain why we do 2 different things here?

Hywan (approved):
Looks good, thanks! I think it's fine to keep `remove_gap_at` and `replace_gap_at` separated.

crates/matrix-sdk-common/src/linked_chunk/mod.rs
Hywan: Thanks

Issue #4424, opened 2024-12-17
Failed storage upgrade for "PreviouslyVerifiedButNoLonger"


andybalaam:
In https://github.com/matrix-org/matrix-rust-sdk/pull/4067 , we changed the storage format for identity state, but forgot to provide a migration, so if an identity was in state `PreviouslyVerifiedButNoLonger`, it causes a crash.

See rageshakes here: https://github.com/element-hq/element-android-rageshakes/issues?q=is%3Aissue+is%3Aopen+VerifiedStateOrBool+

```
2024-12-09T07:54:05*227GMT+00:00Z 604 D/ /Tag: TRACE matrix_sdk_crypto::backups: Verifying backup auth data backup_info=MegolmBackupV1Curve25519AesSha2(MegolmV1AuthData { public_key: "curve25519:JDef8cKMe1HDVC53LkLTr2omw572lhdCU/RtjMupCE4", signatures: Signatures({"@docent053:element.utnet.tech": {"ed25519:QIQLPUFJBE": Ok(Ed25519("ed25519:pxQMD6NMH1+kJ45uMMCbybQHn914Jo9xP4HePv7TF8zE54HlXqvtMdyDgVvPahzIPkyI2W87Bw3NzyYWdgl3AQ")), "ed25519:hPHmYphgXT0AcsNSntCTtfPcmpNtiN/q6x+izRHn4K0": Ok(Ed25519("ed25519:tfts0BYUOVaHt2ToL77PQwLlt5XmokI5Z5AwGak/Q0raXQfp5Z6szmlDId3dwOTi7n+ydNRCqCEWm/6Lkx6gCA"))}}), extra: {} })

2024-12-09T07:54:05*252GMT+00:00Z org.matrix.rustcomponents.sdk.crypto.CryptoStoreException$CryptoStore: data did not match any variant of untagged enum VerifiedStateOrBool
	at org.matrix.rustcomponents.sdk.crypto.FfiConverterTypeCryptoStoreError.read(SourceFile:8)
	at org.matrix.rustcomponents.sdk.crypto.FfiConverterTypeCryptoStoreError.read(SourceFile:1)
	at org.matrix.rustcomponents.sdk.crypto.FfiConverter$DefaultImpls.liftFromRustBuffer(SourceFile:13)
	at org.matrix.rustcomponents.sdk.crypto.FfiConverterRustBuffer$DefaultImpls.liftFromRustBuffer(SourceFile:6)
[...]

PR #4425, opened 2024-12-17
Fix error loading older sender data from storage

andybalaam:
Fixes https://github.com/matrix-org/matrix-rust-sdk/issues/4424

### DIFF
--- a/crates/matrix-sdk-common/src/deserialized_responses.rs
+++ b/crates/matrix-sdk-common/src/deserialized_responses.rs
@@ -263,6 +263,7 @@ pub enum ShieldStateCode {
     /// An unencrypted event in an encrypted room.
     SentInClear,
     /// The sender was previously verified but changed their identity.
+    #[serde(alias = "PreviouslyVerified")]
     VerificationViolation,
 }
 
@@ -922,7 +923,7 @@ mod tests {
         TimelineEventKind, UnableToDecryptInfo, UnableToDecryptReason, UnsignedDecryptionResult,
         UnsignedEventLocation, VerificationState, WithheldCode,
     };
-    use crate::deserialized_responses::{DeviceLinkProblem, VerificationLevel};
+    use crate::deserialized_responses::{DeviceLinkProblem, ShieldStateCode, VerificationLevel};
 
     fn example_event() -> serde_json::Value {
         json!({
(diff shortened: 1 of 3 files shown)


crates/matrix-sdk-crypto/src/olm/group_sessions/sender_data.rs
```diff
@@ -215,6 +215,7 @@ enum SenderDataReader {
         legacy_session: bool,
     },
 
+    #[serde(alias = "SenderUnverifiedButPreviouslyVerified")]
     VerificationViolation(KnownSenderData),
```
richvdh, line 219: There are a whole bunch of other renames in #4067 that we need to check for; notably `OwnUserIdentityVerifiedState` (https://github.com/matrix-org/matrix-rust-sdk/pull/4067/files#diff-d28a9285dd9fd0d1ced29b1986dc92d7a754bb846a24cab48ec78073a1581166R874) which was the one that caused problems for Valere, but also `VerificationLevel`, and there may be others.

poljar:
This talks about a crash, but we just changed the serialization format, deserialization failures should not result in crashes. So we need to also find where the crash happens and convert that to an error.

andybalaam:
> This talks about a crash, but we just changed the serialization format, deserialization failures should not result in crashes. So we need to also find where the crash happens and convert that to an error.

It was a crash, right @BillCarsonFr ? Can you paste the stack trace?

crates/matrix-sdk-crypto/src/olm/group_sessions/sender_data.rs
andybalaam: Sorry, was rushing to find a fix and missed the other places. Will expand this.

andybalaam: Added commits dealing with the other cases

richvdh (approved):
LGTM

andybalaam:
> Alright, but that's EA not catching an exception, right? The SDK itself doesn't crash.

Correct. I fixed the PR description and will fix the commits in my rebase.

crates/matrix-sdk-common/src/deserialized_responses.rs
```diff
@@ -179,6 +179,7 @@ pub enum VerificationLevel {
 
     /// The message was sent by a user identity we have not verified, but the
     /// user was previously verified.
+    #[serde(alias = "PreviouslyVerified")]
```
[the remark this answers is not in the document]
andybalaam: I feel like it's quite obvious and the tests serve to document it.

poljar (approved):

andybalaam:
Pausing this PR so we can make a release branch to release this for crypto-wasm based on the version used in matrix-rust-sdk-crypto-wasm 11.0 (70bcddfba5e19f2d0fd11577edb55b9868f89678)

poljar:
Err, please don't pause it? The WASM bindings are not the only ones affected by this.

andybalaam:
I merged the new PR into `release-for-crypto-wasm-11`, then in merged that branch into main in commit 03947618ffd0eb3e603536e6c9575077f6a91a37

PR #4426, opened 2024-12-17
fix(linked chunk): in LinkedChunk::ritems_from, skip as long as we're on the right chunk

bnjbvr:
The previous code would skip based on the position's index, but not the position's chunk. It could be that the position's chunk is different from the first items chunk, as shown in the example, where the linked chunk ends with a gap; in this case, the position's index would be 0, while the first chunk found while iterating backwards had 3 items. As a result, items 'd' and 'e' would be skipped incorrectly.

The fix is to take into account the chunk id when skipping over items.


Hywan (approved):
Oh sh*t, good catch!

PR #4427, opened 2024-12-17
feat(event cache): avoid storing useless prev-batch tokens

bnjbvr:
We have a problem right now, with the event cache storage: as soon as you receive a previous-batch token, either from the sync or from a previous back-pagination, we consider that we may have more events to retrieve in the past. Later, after running back-paginations, we may realize those events are duplicated. But since a back-pagination will return another previous-batch token, until we hit the start of the timeline, then we're getting stuck in back-pagination mode, until we've reached the start of the timeline again.

That's bad, because it makes the event cache store basically useless: every time you restore a session, you may receive a previous-batch token from sync (even more so with sliding sync, which will give one previous-batch token when timeline_limit=1, then another one when the timeline limit expands to 20). And so you back-paginate back until the start of the history.

This series of commits fixes that, by introducing two new rules:

- in back-pagination, don't assume the absence of a gap always means we're waiting for a prev-batch token from sync. This is only true if there was no events stored in the linked chunk before; if there's any, then we don't need to restart back-pagination from the end of the timeline to the beginning.
- in back-pagination or sync, don't store a previous-batch token, if all the events (at least one) we've received were already known (and are duplicated).

With these two rules, we're not storing useless previous-batch tokens anymore, and we'll back-paginate at most one, for a given room. Interestingly, this might uncover some bugs related to back-pagination orderings, so we'll desperately want https://github.com/matrix-org/matrix-rust-sdk/pull/4408 <3

Part of #3280.

### DIFF
--- a/crates/matrix-sdk/src/event_cache/pagination.rs
+++ b/crates/matrix-sdk/src/event_cache/pagination.rs
@@ -185,7 +185,7 @@ impl RoomPagination {
                 let first_event_pos = room_events.events().next().map(|(item_pos, _)| item_pos);
 
                 // First, insert events.
-                let insert_new_gap_pos = if let Some(gap_id) = prev_gap_id {
+                let (add_event_report, insert_new_gap_pos) = if let Some(gap_id) = prev_gap_id {
                     // There is a prior gap, let's replace it by new events!
                     trace!("replaced gap with new events from backpagination");
                     room_events
@@ -195,16 +195,17 @@ impl RoomPagination {
                     // No prior gap, but we had some events: assume we need to prepend events
                     // before those.
                     trace!("inserted events before the first known event");
-                    room_events
+                    let report = room_events
                         .insert_events_at(sync_events, pos)
                         .expect("pos is a valid position we just read above");
-                    Some(pos)
+                    (report, Some(pos))
                 } else {
                     // No prior gap, and no prior events: push the events.
                     trace!("pushing events received from back-pagination");
-                    room_events.push_events(sync_events);
+                    let report = room_events.push_events(sync_events);
                     // A new gap may be inserted before the new events, if there are any.
-                    room_events.events().next().map(|(item_pos, _)| item_pos)
+                    let next_pos = room_events.events().next().map(|(item_pos, _)| item_pos);
+                    (report, next_pos)
                 };
 
                 // And insert the new gap if needs be.
--- a/crates/matrix-sdk/src/event_cache/pagination.rs
+++ b/crates/matrix-sdk/src/event_cache/pagination.rs
@@ -280,6 +280,29 @@ impl RoomPagination {
     }
 }
 
+/// Pagination token data, indicating in which state is the current pagination.
+#[derive(Clone, Debug)]
+pub(super) enum PaginationToken {
+    /// We never had a pagination token, so we'll start back-paginating from the
+    /// end, or forward-paginating from the start.
+    None,
+    /// We paginated once before, and we received a prev/next batch token that
+    /// we may reuse for the next query.
+    HasMore(String),
+    /// We've hit one end of the timeline (either the start or the actual end),
+    /// so there's no need to continue paginating.
+    HitEnd,
+}
+
+impl From<Option<String>> for PaginationToken {
+    fn from(token: Option<String>) -> Self {
+        match token {
+            Some(val) => Self::HasMore(val),
+            None => Self::None,
+        }
+    }
+}
+
 /// A type representing whether the timeline has been reset.
 #[derive(Debug)]
 pub enum TimelineHasBeenResetWhilePaginating {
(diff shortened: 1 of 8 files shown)


Hywan:
Stealing the review to @stefanceriu because <3.

bnjbvr:
Fwiw, there's actually a small issue with this: if we're not storing a previous-batch token, we should *also* not replace the gap with the events returned by back-pagination.

Consider the following timeline, that exists as such on the server, where the numbers represent event ids: [0, 1, 2, 3, 4, 5].

Let's take the notation that `G(n)` is a token for a back-pagination (gap, hence G), starting at (including) the n-th event, and going backwards. For instance, using the pagination token `G(3)` by requesting 2 events will return [2, 3] (in reverse order, but that's an implementation detail we don't care about here).

We start sliding sync:

- first, with a timeline_limit set to 0.
  - We receive from the server `[G(4), 5]`
  - our local state becomes `[G(4), 5]`: we have the pagination token for events 4 and before; we received event 5
- then, we subscribe to the room, and expand the timeline_limit to 3.
  - We receive from the server `[G(2), 3, 4, 5]`. 
  - After sync, we "push" (and deduplicate) the sync response, so our local state becomes `[G(4), G(2), 3, 4, 5]`,  after deduplicating 5.
- we use /messages and consume `G(2)`, requesting 3 events.
  - We receive from the server `[1, 2]`
  - We're replacing `G(2)` with `[1, 2]`, so our local state becomes `[G(4), 1, 2, 3, 4, 5]`.
  - There's no *new* previous-batch token because we've hit the start of the timeline.
- we use /messages and consume `G(4)`, requesting 3 events.
  - We receive from the server `[G(1), 2, 3, 4]`.
  - **With this PR, we're not storing `G(1)`, because all the returned events are known.**
  - We're replacing `G(4)` with the server response, and deduplicate as we go along: our local state becomes `[2, 3, 4, 1, 5]` :boom: 
  - Because we haven't stored `G(1)`, we can't start another back-pagination that would fix the order by deduplicating (1) into its final correct position anymore :exploding_head: 

I think the solution is to *not* do anything when all the events are known. After all, they are at their right location in the linked chunk, and as such, they shouldn't move anymore. I'll need to think a bit more about this, to make sure this is the right solution.

crates/matrix-sdk/src/event_cache/pagination.rs
```diff
@@ -185,7 +185,7 @@ impl RoomPagination {
                 let first_event_pos = room_events.events().next().map(|(item_pos, _)| item_pos);
 
                 // First, insert events.
-                let insert_new_gap_pos = if let Some(gap_id) = prev_gap_id {
```
Hywan, line 178: Are you doing something with `add_event_report`?

crates/matrix-sdk/src/event_cache/room/events.rs
```diff
@@ -398,6 +422,20 @@ impl RoomEvents {
     }
 }
 
+pub(in crate::event_cache) struct AddEventReport {
```
Hywan, line 438: Can you explain/document why this type is needed please?

crates/matrix-sdk/src/event_cache/room/events.rs
```diff
@@ -398,6 +422,20 @@ impl RoomEvents {
...
+    num_new_unique: usize,
+    /// Number of events which have been deduplicated.
+    num_duplicated: usize,
+}
+
+impl AddEventReport {
+    /// Were all the events (at least one) we added already known?
+    pub fn deduplicated_all_new_events(&self) -> bool {
+        self.num_new_unique > 0 && self.num_new_unique == self.num_duplicated
+    }
```
Hywan, line 449: Can't you simply test `self.num_duplicated > 0`?

Actually, I don't understand the code here. The code, the doc, and the method name don't seem to do the same thing. Can you revisit this a bit please?

Hywan (approved):
I'm not sure to understand if more work is required, but so far, it looks good. I'm approving the PR :-).

Thank you for taking care of the tests.

crates/matrix-sdk/src/event_cache/pagination.rs
bnjbvr: (Yep, it's used in a subsequent commit :+1:)

crates/matrix-sdk/src/event_cache/room/events.rs
bnjbvr: I've rewritten the last commit and put a super large comment, hopefully it helps. Let me know if it doesn't, post-merge, and we can rephrase it :pray:

crates/matrix-sdk/src/event_cache/room/events.rs
bnjbvr: I got rid of it \o/

crates/matrix-sdk/src/event_cache/room/events.rs
Hywan: I like you've removed `AddEventReport`, it seems cleaner this way. I may have prefer a `Option<T>` instead of `(bool, T)`, but it's a personal taste (I believe it matches `std` a bit better, but that's a detail, I'll try to submit a PR to not forget that and being able to discuss it).

PR #4428, opened 2024-12-18
test(snapshot): Use snapshot testing in sdk-common

BillCarsonFr:
Tentative to introduce snapshot testing using [insta.rs](https://insta.rs/).
Added some initial doc on the Contributing.md file

As a draft now, because I have some open questions on how to do things.

I need also to see what to do for CI

> Insta behaves differently in CI environments. For instance it will not write new snapshot files. This behavior is enabled by the CI environment variable. Please ensure that your CI system sets it.

`export CI=true`

Signed-off-by:

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -46,6 +46,7 @@ hmac = "0.12.1"
 http = "1.1.0"
 imbl = "3.0.0"
 indexmap = "2.6.0"
+insta = { version = "1.41.1", features = ["json"] }
 itertools = "0.13.0"
 js-sys = "0.3.69"
 mime = "0.3.17"
@@ -124,6 +125,9 @@ debug = 0
 # for the extra time of optimizing it for a clean build of matrix-sdk-ffi.
 quote = { opt-level = 2 }
 sha2 = { opt-level = 2 }
+# faster runs for insta.rs snapshot testing
+insta.opt-level = 3
+similar.opt-level = 3
 
 # Custom profile with full debugging info, use `--profile dbg` to select
 [profile.dbg]
--- a/crates/matrix-sdk-common/Cargo.toml
+++ b/crates/matrix-sdk-common/Cargo.toml
@@ -46,6 +46,7 @@ assert_matches = { workspace = true }
 proptest = { workspace = true }
 matrix-sdk-test-macros = { path = "../../testing/matrix-sdk-test-macros" }
 wasm-bindgen-test = { workspace = true }
+insta = { workspace = true }
 
 [target.'cfg(not(target_arch = "wasm32"))'.dependencies]
 # Enable the test macro.
--- a/crates/matrix-sdk-common/src/deserialized_responses.rs
+++ b/crates/matrix-sdk-common/src/deserialized_responses.rs
@@ -909,21 +909,22 @@ mod tests {
     use std::collections::BTreeMap;
 
     use assert_matches::assert_matches;
+    use insta::{assert_json_snapshot, with_settings};
     use ruma::{
-        event_id,
+        device_id, event_id,
         events::{room::message::RoomMessageEventContent, AnySyncTimelineEvent},
         serde::Raw,
-        user_id,
+        user_id, DeviceKeyAlgorithm,
     };
     use serde::Deserialize;
     use serde_json::json;
 
     use super::{
-        AlgorithmInfo, DecryptedRoomEvent, EncryptionInfo, SyncTimelineEvent, TimelineEvent,
-        TimelineEventKind, UnableToDecryptInfo, UnableToDecryptReason, UnsignedDecryptionResult,
-        UnsignedEventLocation, VerificationState, WithheldCode,
+        AlgorithmInfo, DecryptedRoomEvent, DeviceLinkProblem, EncryptionInfo, ShieldState,
+        ShieldStateCode, SyncTimelineEvent, TimelineEvent, TimelineEventKind, UnableToDecryptInfo,
+        UnableToDecryptReason, UnsignedDecryptionResult, UnsignedEventLocation, VerificationLevel,
+        VerificationState, WithheldCode,
     };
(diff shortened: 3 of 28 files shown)


Cargo.toml
```diff
@@ -124,6 +124,9 @@ debug = 0
 # for the extra time of optimizing it for a clean build of matrix-sdk-ffi.
 quote = { opt-level = 2 }
 sha2 = { opt-level = 2 }
+# faster runs for insta.rs snapshot testing
```
BillCarsonFr, line 127: I think it is the right place to put it? but maybe we need also in other conf? 
It is recommened to do that in https://insta.rs/docs/quickstart/

crates/matrix-sdk-common/src/deserialized_responses.rs
```diff
@@ -1247,4 +1248,126 @@ mod tests {
...
+        let level = VerificationLevel::UnverifiedIdentity;
+        insta::assert_json_snapshot! {
+            serde_json::to_value(&level).unwrap(),
+            @r###""UnverifiedIdentity""###
+        }
+    }
+
+    #[test]
+    fn snapshot_test_verification_states() {
+        let states = vec![
```
BillCarsonFr, line 1283: Is there a way to iterate other all enums? I can see some macro that do that. It might be interesting to do it here?

crates/matrix-sdk-common/src/deserialized_responses.rs
```diff
@@ -1247,4 +1248,126 @@ mod tests {
...
+    #[test]
+    fn snapshot_test_verification_level() {
+        let level = VerificationLevel::VerificationViolation;
+        insta::assert_json_snapshot! {
+            serde_json::to_value(&level).unwrap(),
+        }
+
+        let level = VerificationLevel::UnsignedDevice;
+        insta::assert_json_snapshot! {
+            serde_json::to_value(&level).unwrap(),
```
BillCarsonFr, line 1261: We have the choice here to inline the snapshot by using the `@""` token that insta.rs will replace with the actual string instead of generating a new file in `snapshots/`. WDYT?

Cargo.toml
poljar: Yes.

crates/matrix-sdk-common/Cargo.toml
```diff
@@ -46,6 +46,7 @@ assert_matches = { workspace = true }
 proptest = { workspace = true }
 matrix-sdk-test-macros = { path = "../../testing/matrix-sdk-test-macros" }
 wasm-bindgen-test = { workspace = true }
+insta = { version = "1.41.1", features = ["json"] }
```
poljar, line 49: Let's put this right away as a workspace dependency, it's very likely that we're going to use that in multiple crates.

crates/matrix-sdk-common/src/deserialized_responses.rs
```diff
@@ -1247,4 +1248,126 @@ mod tests {
         let reason = UnableToDecryptReason::UnknownMegolmMessageIndex;
         assert!(reason.is_missing_room_key());
     }
+
+    #[test]
+    fn snapshot_test_verification_level() {
+        let level = VerificationLevel::VerificationViolation;
+        insta::assert_json_snapshot! {
```
poljar, line 1255: Please import `assert_json_snapshot` and use it directly.

crates/matrix-sdk-common/src/deserialized_responses.rs
poljar: I find it nicer if it's like, with the snapshot out of sight.

crates/matrix-sdk-common/src/deserialized_responses.rs
poljar: What you are looking for is Strum: https://docs.rs/strum/latest/strum/trait.VariantIterator.html. 

I don't think it's worth adding this trait to the enum, unless we need an iterator over the variants in the public API as well.

crates/matrix-sdk-common/src/snapshots/matrix_sdk_common__deserialized_responses__tests__snapshot_test_algorithm_info.snap
```diff
@@ -0,0 +1,13 @@
+---
+source: crates/matrix-sdk-common/src/deserialized_responses.rs
+expression: "serde_json::to_value(&info).unwrap()"
+---
+{
+  "MegolmV1AesSha2": {
```
poljar, line 6: Nice, now we don't need to imagine how things will get serialized either, we can see it right here.

poljar (changes requested):
Nice.

crates/matrix-sdk-common/src/deserialized_responses.rs
```diff
@@ -1317,4 +1318,172 @@ mod tests {
         let reason = UnableToDecryptReason::UnknownMegolmMessageIndex;
         assert!(reason.is_missing_room_key());
     }
+
+    #[test]
+    fn snapshot_test_verification_level() {
+        let level = VerificationLevel::VerificationViolation;
+        assert_json_snapshot! {
+            serde_json::to_value(&level).unwrap(),
+        }
```
poljar, line 1327: Wait a minute, what's going on here?

Why are you calling `serde_json::to_value()` manually? You can just do this:

```suggestion
        assert_json_snapshot!(VerificationLevel::VerificationViolation);
        }
```

This applies to all the places wehre you use `assert_json_snapshot()`.

poljar (changes requested):

poljar (approved):
Looks good. Thanks for adding this.

crates/matrix-sdk-common/src/deserialized_responses.rs
BillCarsonFr: 🤯
so much better like that

PR #4437, opened 2024-12-18
fix(linked chunk): fix order handling of initial chunks in `UpdateToVectorDiff::new`

bnjbvr:
The code would use a chunk iterator that moves forward, but call `push_front()` repetitively on each chunk, semantically storing the lengths in *reverse* order.

This could result in subsequent panics, when a new chunk was added, because the links would not match what's expected (e.g. the last chunk must have no successor, etc.).


Hywan (approved):
Of course!! Thanks for spotting this bug.

PR #4438, opened 2024-12-18
test(room): add test to verify `Room::observe_events` with several events

jmartinesp:
The code in `Client::handle_sync_events` and `Client::handle_sync_state_events` call `Client::call_event_handlers`, which takes a single event and it gives the impression of the underlying event handler notifying the subscribers once per each event fed to it, which is the opposite behaviour to what `Room::observe_events` docs state:

> Be careful that only the most recent value can be observed. Subscribers are notified when a new value is sent, but there is no guarantee that they will see all values.

It turns out it works like that because the sender will be fed items way faster than the receiver can handle them if I'm not mistaken.

This behaviour wasn't tested before AFAICT, so I added a test to make sure the described behaviour doesn't regress at some point.

Signed-off-by:

### DIFF
--- a/crates/matrix-sdk/src/event_handler/mod.rs
+++ b/crates/matrix-sdk/src/event_handler/mod.rs
@@ -1167,4 +1167,78 @@ mod tests {
 
         Ok(())
     }
+
+    #[async_test]
+    async fn test_observe_several_room_events() -> crate::Result<()> {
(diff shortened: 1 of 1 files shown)


Hywan (commented):
Sounds good to me.

stefanceriu (approved):
Nice! 👌

PR #4443, opened 2024-12-19
refactor!(client): replace the optional `RequestConfig` arg to `Client::send()` with a `with_request_config()` builder method

bnjbvr:
Instead of `Client::send(request, request_config)`, consumers can now do `Client::send(request).with_request_config(request_config)`. The parameter is still `Option`al to avoid messy patterns, when the `request_config` is an `Option` passed from the caller.

The messy pattern would look like this, if `with_request_config` took a `RequestConfig` and not an `Option<RequestConfig>`:

```
let mut send_req = client.send(req);
if let Some(request_config) = request_config {
  send_req = send_req.with_request_config(request_config);
}
```

### DIFF
--- a/crates/matrix-sdk/CHANGELOG.md
+++ b/crates/matrix-sdk/CHANGELOG.md
@@ -6,6 +6,13 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased] - ReleaseDate
 
+### Refactor
+
+- [**breaking**] Move the optional `RequestConfig` argument of the
+  `Client::send()` method to the `with_request_config()` builder method. You
+  should call `Client::send(request).with_request_config(request_config).awat`
+  now instead.
+
 ## [0.9.0] - 2024-12-18
 
 ### Bug Fixes
--- a/crates/matrix-sdk/src/client/futures.rs
+++ b/crates/matrix-sdk/src/client/futures.rs
@@ -77,6 +77,13 @@ impl<R> SendRequest<R> {
         self
     }
 
+    /// Use the given [`RequestConfig`] for this send request, instead of the
+    /// one provided by default.
+    pub fn with_request_config(mut self, request_config: impl Into<Option<RequestConfig>>) -> Self {
+        self.config = request_config.into();
+        self
+    }
+
     /// Get a subscriber to observe the progress of sending the request
     /// body.
     #[cfg(not(target_arch = "wasm32"))]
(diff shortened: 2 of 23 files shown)


crates/matrix-sdk/src/client/futures.rs
```diff
@@ -77,6 +77,13 @@ impl<R> SendRequest<R> {
         self
     }
 
+    /// Use the given [`RequestConfig`] for this send request, instead of the
+    /// one provided by default.
+    pub fn with_request_config(mut self, request_config: Option<RequestConfig>) -> Self {
```
poljar, line 82: Why would we take an option here? Not calling this is the same as passing `None`, no?

poljar (changes requested):

crates/matrix-sdk/src/client/futures.rs
bnjbvr: It's to avoid the messy pattern described in OP and commit message. I was a bit torn about this.

jplatte: Could do `impl Into<Option<RequestConfig>>` so people can pass a `RequestConfig` directly. There's no reason to ever call it with `None` so type inference should not be an issue.

poljar: That's a good idea.

poljar:
To prevent bitrot on this during the holidays, I took the freedom to implement the change myself, I also added a changelog entry and reworded the now obsolete commit message.

poljar (approved):

crates/matrix-sdk/CHANGELOG.md
```diff
@@ -6,6 +6,13 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased] - ReleaseDate
 
+### Refactor
+
+- [**breaking**] Move the optional `RequestConfig` argument of the
+  `Client::send()` method to the `with_request_config()` builder method. You
+  should call `Client::send(request).with_request_config(request_config).awat`
```
jplatte, line 13:
```suggestion
  should call `Client::send(request).with_request_config(request_config).await`
```

poljar: Oh, that's a nice typo.

poljar: https://github.com/matrix-org/matrix-rust-sdk/pull/4446

Issue #4447, opened 2024-12-20
RoomList NonLeft filter doesn't work with Banned


ganfra:
This [PR ](https://github.com/matrix-org/matrix-rust-sdk/pull/4414) introduced a regression on the RoomList NonLeft filter.
So the banned rooms are not filtered anymore by this.

poljar:
I think #4448 should fix this.

PR #4448, opened 2024-12-20
fix(ui): Consider banned rooms as rooms we left in the non-left rooms matcher

poljar:
Recently we started to differentiate between rooms we've been banned from from rooms we have left on our own.

Sadly the non-left rooms matcher only checked if the room state is not equal to the Left state. This then accidentally moved all the banned rooms to be considered as non-left.

We replace the single if expression with a match and list all the states, this way we're going to be notified by the compiler that we need to consider any new states we add in the future.

This closes: #4447.


stefanceriu (approved):
Nice one, lgtm! 👍

Hywan (approved):
Good!

PR #4449, opened 2024-12-20
test(crypto): Add some basic snapshot testing in crypto crate

BillCarsonFr:
Adding some snapshot testing for the crypto test.

Another benefit of insta is we see right away how things are serialised, and 2 findings here:

- `KnownSenderData` serializes the master_key as an array of numbers in json :/
- `BackupDecryptionKey` does it too, but it is never serialized by itself (but using `serialize_with`). It probably doesn't need to be serializable by itself?

Signed-off-by:


crates/matrix-sdk-crypto/src/types/mod.rs
```diff
@@ -547,4 +549,79 @@ mod test {
 
         assert_eq!(json, serialized, "A serialization cycle should yield the same result");
     }
+
+    #[test]
+    fn snapshot_backup_decryption_key() {
+        let decryption_key = BackupDecryptionKey { inner: Box::new([1u8; 32]) };
+        assert_json_snapshot!(decryption_key);
+
+        // should not log the key !
```
poljar, line 558: Neat.

poljar (approved):
Very nice, thanks.

PR #4450, opened 2024-12-20
fix(crypto): Serialize sender data msk in base64 instead of numbers

BillCarsonFr:
Follow up of https://github.com/matrix-org/matrix-rust-sdk/pull/4449

While adding snapshot test, I detected that sender's data `KnownSenderData` was serializing the cross-signing master key as an array of number. This PR changes that and now serialize the key as a base64 string, adding migration for the old format.

Signed-off-by:

### DIFF
--- a/crates/matrix-sdk-crypto/src/olm/group_sessions/sender_data.rs
+++ b/crates/matrix-sdk-crypto/src/olm/group_sessions/sender_data.rs
@@ -12,13 +12,13 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use std::cmp::Ordering;
+use std::{cmp::Ordering, fmt};
 
 use ruma::{DeviceId, OwnedDeviceId, OwnedUserId, UserId};
-use serde::{Deserialize, Serialize};
+use serde::{de, de::Visitor, Deserialize, Deserializer, Serialize};
 use vodozemac::Ed25519PublicKey;
 
-use crate::types::DeviceKeys;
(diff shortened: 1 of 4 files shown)


crates/matrix-sdk-crypto/src/olm/group_sessions/sender_data.rs
```diff
@@ -33,9 +33,55 @@ pub struct KnownSenderData {
...
+    #[serde(
+        serialize_with = "serialize_ed25519_key",
+        deserialize_with = "deserialize_sender_msk_base64_or_array"
+    )]
     pub master_key: Box<Ed25519PublicKey>,
 }
 
+/// In an initial version the master key was serialized as an array of number,
+/// it is now exported in base64. This code adds backward compatibility.
+pub(crate) fn deserialize_sender_msk_base64_or_array<'de, D>(
```
BillCarsonFr, line 45: I am not familiar with this serde visitor pattern. Let me know if it is sane or if there is a prefferred method? I am not sure about the error mappings

richvdh:
Filed https://github.com/matrix-org/matrix-rust-sdk/issues/4459 about the test failure

andybalaam (approved):
This looks good to me, and the error mappings look about right.

@poljar should Edd25519PublicKey serialize as base64 by default?

poljar:
> This looks good to me, and the error mappings look about right.
> 
> @poljar should Edd25519PublicKey serialize as base64 by default?

No, only if we serialize into JSON.

PR #4451, opened 2024-12-21
feat!(timeline): Allow to send attachments from bytes

zecakeh:
Sometimes we can get the bytes directly, e.g. in Fractal we can get an image from the clipboard. It avoids to have to write the data to a temporary file only to have the data loaded back in memory by the SDK right after.

The first commit to accept any type that implements `Into<String>` for the filename is grouped here because it simplifies slightly the second commit.

Note that we could also use `AttachmentSource` in the other `send_attachment` APIs, on `Room` and `RoomSendQueue`, for consistency.

### DIFF
--- a/crates/matrix-sdk-ui/CHANGELOG.md
+++ b/crates/matrix-sdk-ui/CHANGELOG.md
@@ -13,6 +13,14 @@ All notable changes to this project will be documented in this file.
   non-left room filter did not take the new room stat into account.
   ([#4448](https://github.com/matrix-org/matrix-rust-sdk/pull/4448))
 
+### Features
+
+- [**breaking**] `Timeline::send_attachment()` now takes a type that implements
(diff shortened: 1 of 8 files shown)


zecakeh:
Added changelog entries in the commits.

zecakeh:
I merged main into this branch to solve the conflict. Maybe you would have preferred a rebase to keep the history clean?

bnjbvr:
> I merged main into this branch to solve the conflict. Maybe you would have preferred a rebase to keep the history clean?

Thanks. It's rather up to you:

- I don't tend to merge PRs which contain merge commits as is, instead I'll squash-merge them.
- but if there's only clean atomic commits, I usually rebase-and-merge

So unless you have a strong preference for rebase-and-merge here (in which case, yes, please keep a clean history), I'll squash-merge later :+1:

zecakeh:
It's fine, you can merge as is :slightly_smiling_face: .

PR #4452, opened 2024-12-21
refactor!(sdk): Set thumbnail in `AttachmentConfig` with builder method instead of constructor

zecakeh:
`AttachmentConfig::with_thumbnail()` is replaced by `AttachmentConfig::new().thumbnail()`.

Simplifies the use of `AttachmentConfig`, by avoiding code like:

```rust
let config = if let Some(thumbnail) = thumbnail {
  AttachmentConfig::with_thumbnail(thumbnail)
} else {
  AttachmentConfig::new()
};
```

### DIFF
--- a/bindings/matrix-sdk-ffi/src/timeline/mod.rs
+++ b/bindings/matrix-sdk-ffi/src/timeline/mod.rs
@@ -137,9 +137,9 @@ impl Timeline {
 fn build_thumbnail_info(
     thumbnail_url: Option<String>,
     thumbnail_info: Option<ThumbnailInfo>,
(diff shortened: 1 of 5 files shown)


poljar (changes requested):
Neat, could you add a changelog entry?

zecakeh:
Oh the project is not using git-cliff anymore? Sure, I'll do that.

poljar:
> Oh the project is not using git-cliff anymore? Sure, I'll do that.

Yeah, after a brief adventure we're back to manual changelogs.

poljar (approved):
Looks good, thanks.

PR #4461, opened 2025-01-06
refactor(event cache): make it clearer that vecdiff updates must be handled in `with_events_mut`

bnjbvr:
Every caller to `with_events_mut` must propagate the vector diff updates, otherwise updates would be missing to the room event cache's observers. This slightly tweaks the signature to make this a bit clearer, and adjusts the code comment as well.


Hywan (approved):
Good way to make the API more robust. Thanks.

PR #4463, opened 2025-01-06
feat!(send queue): make unrecoverable errors stop the sending queue

### LINKED ISSUE #4389: Not disabling the send queue when an event is wedged breaks the expectations of users
If the error is temporary, the send queue is disabled, so the events are sent in the right order, which makes sense. However it is not disabled in case of a "permanent" error, the event is marked as wedged and the following events are sent.

A user opened [an issue](https://gitlab.gnome.org/World/fractal/-/issues/1563) on Fractal, surprised by this behavior (among other things).

It seems logical that a user would expect their events to be sent in the right order or not at all, whether the error is permanent or not. It should be their choice whether they want to remove an event or several from the send queue, or all of them, due to the first one failing to send, to avoid losing context in the conversation.

According to the issue, this is already the behavior of EW.

bnjbvr:
Instead of keeping on handling unwedged events from the sending queue, it's now required to re-enable the send queue manually for the room that encountered the sending error, all the time. This is more consistent, and avoids weird behavior when a user would 1. send an event for which sending fails, in an unrecoverable manner, 2. send an event that's actually sendable.

Fixes #4389.

### DIFF
--- a/crates/matrix-sdk/src/send_queue.rs
+++ b/crates/matrix-sdk/src/send_queue.rs
@@ -622,6 +622,9 @@ impl RoomSendQueue {
                         _ => false,
                     };
 
+                    // Disable the queue for this room after any kind of error happened.
+                    locally_enabled.store(false, Ordering::SeqCst);
+
                     if is_recoverable {
                         warn!(txn_id = %txn_id, error = ?err, "Recoverable error when sending request: {err}, disabling send queue");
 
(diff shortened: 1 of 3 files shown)


crates/matrix-sdk/src/send_queue.rs
```diff
@@ -622,22 +622,23 @@ impl RoomSendQueue {
...
-                        // should be the sign that this error is temporary (maybe network
-                        // disconnected, maybe the server had a hiccup).
-                        locally_enabled.store(false, Ordering::SeqCst);
+                        // Let observers know about a failure *after* we've
+                        // marked the item as not
+                        // being sent anymore. Otherwise, there's a possible
+                        // race where a caller might try
+                        // to remove an item, while it's still marked as being
+                        // sent, resulting in a
+                        // cancellation failure.
```
andybalaam, line 641: The formatting of this comment looks like it's messed up somehow.

andybalaam (approved):
Looks good; one tiny comment.

bnjbvr:
Thanks for the review!

PR #4465, opened 2025-01-06
doc(base): update lying doc comment of `compute_display_name`

bnjbvr:
It claimed that it would immediately return when the cached display name value was computed, but that's absolutely not the case.

Spotted while reviewing a PR updating `iamb` to the latest version of the SDK.

### DIFF
--- a/crates/matrix-sdk-base/src/rooms/normal.rs
+++ b/crates/matrix-sdk-base/src/rooms/normal.rs
@@ -623,15 +623,15 @@ impl Room {
         self.inner.read().active_room_call_participants()
     }
 
-    /// Return the cached display name of the room if it was provided via sync,
-    /// or otherwise calculate it, taking into account its name, aliases and
-    /// members.
+    /// Calculate a room's display name, taking into account its name, aliases
+    /// and members.
     ///
     /// The display name is calculated according to [this algorithm][spec].
     ///
-    /// This is automatically recomputed on every successful sync, and the
-    /// cached result can be retrieved in
-    /// [`Self::cached_display_name`].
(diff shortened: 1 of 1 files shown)


crates/matrix-sdk-base/src/rooms/normal.rs
```diff
@@ -623,15 +623,15 @@ impl Room {
...
+    /// Calculate a room's display name, taking into account its name, aliases
+    /// and members.
     ///
     /// The display name is calculated according to [this algorithm][spec].
     ///
-    /// This is automatically recomputed on every successful sync, and the
-    /// cached result can be retrieved in
-    /// [`Self::cached_display_name`].
+    /// ⚠ This may be slowish to compute. As such, the result is cached and can
+    /// be retrieved via [`Self::cached_display_name`], which should be
```
poljar, line 632:
```suggestion
    /// be retrieved via [`Room::cached_display_name`], which should be
```

poljar (approved):
A bit tangentially related, but why don't we have a method that does what the docs claim to do?

A method like this:

```rust
pub async fn foo() -> StoreResult<RoomDisplayName> {
    if let Some(result) = self.cached_display_name() {
        result
    } else {
        self.compute_dislpay_name().await
    }
}
```

We may even call this method just `display_name()` and people wouldn't need to think about the other two methods, unless they really care about never hitting the slow path.

bnjbvr:
> A bit tangentially related, but why don't we have a method that does what the docs claim to do?

Ah nice, I thought that the reason was that `compute_display_name` was _defined_ in the base crate, and used in the _main_ crate, but that's not true! So we could have `compute_display_name` be private in the base crate, and expose a public method like that you propose. Will do as a followup!

PR #4467, opened 2025-01-06
Use Instant from web-time in more places (via ruma re-export)

jplatte:
web-time's Instant type is already used elsewhere in the project. It is an alias for std's Instant type on most targets, but tries to call into JavaScript on wasm32-unknown-unknown (assuming that the wasm blob is used in from a browser context). Its Duration type is a plain re-export of std's Duration, even on wasm32-unknown-unknown. (it is only used to avoid splitting the `time::` import in two)


poljar (approved):
Nice, thanks.

PR #4470, opened 2025-01-07
feat!(base): reintroduce `Room::display_name`

### LINKED ISSUE #4469: Rationalize `compute_display_name()` and `cached_display_name()`
A bit tangentially related, but why don't we have a method that does what the docs claim to do?

A method like this:

```rust
pub async fn foo() -> StoreResult<RoomDisplayName> {
    if let Some(result) = self.cached_display_name() {
        result
    } else {
        self.compute_dislpay_name().await
    }
}
```

We may even call this method just `display_name()` and people wouldn't need to think about the other two methods, unless they really care about never hitting the slow path.

_Originally posted by @poljar in https://github.com/matrix-org/matrix-rust-sdk/pull/4465#pullrequestreview-2533636092_

bnjbvr:
`compute_display_name` is made private again, and used only within the base crate. A new public counterpart `Room::display_name` is introduced, which returns a cached value for, or computes (and fills in cache) the display name. This is simpler to use, and likely what most users expect anyways.

Fixes #4469.

### DIFF
--- a/crates/matrix-sdk-base/CHANGELOG.md
+++ b/crates/matrix-sdk-base/CHANGELOG.md
@@ -6,6 +6,16 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased] - ReleaseDate
 
+### Breaking changes
+
+- Replaced `Room::compute_display_name` with the reintroduced `Room::display_name()`. The new
+  method computes a display name, or return a cached value from the previous successful computation.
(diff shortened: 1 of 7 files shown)


poljar (approved):
Nice, you even opened an issue for maxing out the github points :muscle:

bnjbvr:
Yeah, I was in the mood for postponing, to be honest, but then realized it was a pretty simple change…

PR #4474, opened 2025-01-07
refactor: Make experimental-sliding-sync compile on WASM

jplatte:
Unfortunately the matrix_sdk::Client is not Send on WASM, which is probably not going to be easy to fix. Since PreviousEventsProvider is never used as a trait object, there should be no harm in relaxing its supertraits (on WASM).

I intentionally left out `xtask::ci` changes, because I think this should be tested as part of a `matrix-sdk-ui` WASM check down the line.


Hywan:
Closing in favor of #4483 as it contains a commit to remove `SlidingSyncRoomInner::client` (it's useless, booo dead code, booo).

PR #4477, opened 2025-01-08
chore: Upgrade Ruma

zecakeh:
This is using the `ruma-0.12` branch where non-breaking changes are backported.

Includes the commit necessary for https://github.com/matrix-org/matrix-rust-sdk/pull/4421.


jmartinesp (approved):
Thanks!

PR #4475, opened 2025-01-08
feat(timeline): allow sending mentions along with media

frebib:
Since 8205da898e11c34b77b13fe4794d0b2e1bcdd2a8 it has been possible to attach (intentional) mentions to _edited_ media captions, but the `send_${mediatype}()` timeline APIs provided no way to send them with the initial event. This fixes that.

- [ ] Public API changes documented in changelogs (optional)

image

If I missed something, or completely got the wrong end of the stick, feel free to decline this or point me in the right direction. I'm not familiar with Rust, but this at least compiles and doesn't crash EXA on my phone.

PR #4484, opened 2025-01-08
chore(ffi): fix inconsistent `mentions` parameter layout in `send_voice_message`

stefanceriu:
Introduced in #4475


stefanceriu:
Fair 👍

PR #4485, opened 2025-01-08
fix(ffi): ensure there's always a global directive for logs

bnjbvr:
If it's present, we just let it untouched. Otherwise, we set it to `error` if it's missing. See code comment explaining why we need this.

This makes sure we log panics at the FFI layer, since the log_panics crate will use an error level, with a target that's likely unknown to the user providing the log filter configuration.

### DIFF
--- a/bindings/matrix-sdk-ffi/src/platform.rs
+++ b/bindings/matrix-sdk-ffi/src/platform.rs
@@ -243,8 +243,22 @@ pub struct TracingConfiguration {
 }
 
 #[matrix_sdk_ffi_macros::export]
-pub fn setup_tracing(config: TracingConfiguration) {
+pub fn setup_tracing(mut config: TracingConfiguration) {
     log_panics();
+
+    // If there's no global directive, and a non-empty set of module/crate
+    // directives, then logs from modules and crates *not* appearing in the list
+    // of directives will be silently swallowed. This is bad, especially at the
+    // `error` level, for which we always want to see logs (notably because the
+    // panic logger uses the error level and a target that may not be specified in
+    // the list of user-provided directives).
+    //
+    // As a result, always make sure there's a global directive. If it's not there,
+    // we insert it at the beginning of the filter line.
+
+    if !config.filter.split(',').any(|pair| !pair.contains('=')) {
+        config.filter = "error,".to_owned() + config.filter.as_str();
+    }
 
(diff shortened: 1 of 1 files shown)


poljar (approved):

bindings/matrix-sdk-ffi/src/platform.rs
```diff
@@ -243,9 +243,23 @@ pub struct TracingConfiguration {
...
 
 #[matrix_sdk_ffi_macros::export]
-pub fn setup_tracing(config: TracingConfiguration) {
+pub fn setup_tracing(mut config: TracingConfiguration) {
     log_panics();
 
+    // If there's no global directive, and a non-empty set of module/crate
+    // directives, then logs from modules and crates *not* appearing in the list
+    // of directives will be silently swallowed. This is bad, especially at the
+    // `error` level, for which we always want to see logs (notably because the
```
stefanceriu, line 252: The original reason for removing the global directive is to take **zero** chances that a 3rd party crate would log sensitive data, which would be hard to find and even harder to clean up.

Hywan (approved):

andybalaam (approved):

bindings/matrix-sdk-ffi/src/platform.rs
bnjbvr: Yeah, that's a good point. I've implemented your suggestion to restrict this to only adding `panic=error` to the log filter line, since it's a more cautious choice.

andybalaam (approved):

bindings/matrix-sdk-ffi/src/platform.rs
```diff
@@ -243,11 +260,37 @@ pub struct TracingConfiguration {
...
+    #[test]
+    fn test_add_panic_when_not_provided_non_empty() {
+        let mut filter = String::from("a=b,c=d");
+        add_panic_to_filter(&mut filter);
+        assert_eq!(filter, "a=b,c=d,panic=error");
+    }
+
+    #[test]
+    fn test_do_nothing_when_provided() {
+        let mut filter = String::from("a=b,panic=error,c=d");
```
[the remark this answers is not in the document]
bnjbvr: Indeed, will adjust, thanks!

stefanceriu (approved):
Just tested out the new version and it works great!

PR #4486, opened 2025-01-08
Introduce simplified sync locks that panic when poisoned

poljar:
This should help us to enable the `unwrap_used` clippy lint in the future.

Should be reviewed commit by commit, we first introduce the lock wrappers and then we start using them.

No functional changes.


poljar:
> Thanks, LGTM! I guess we'll be migrating the existing `lock().unwrap()` and RwLock usages little by little before enabling the clippy check?

Yes. There are also a bunch of questionable unwraps that are not caused by the locks.

PR #4487, opened 2025-01-08
fix(base): Correctly name the LeftRoomUpdate in its debug implementation

poljar:


crates/matrix-sdk-base/src/sync.rs
```diff
@@ -215,7 +215,7 @@ impl LeftRoomUpdate {
 #[cfg(not(tarpaulin_include))]
 impl fmt::Debug for LeftRoomUpdate {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        f.debug_struct("JoinedRoom")
+        f.debug_struct("LeftRoom")
```
bnjbvr, line 218:
```suggestion
        f.debug_struct("LeftRoomUpdate")
```

bnjbvr (approved):
:sweat_smile:

crates/matrix-sdk-base/src/sync.rs
Hywan: lulz

poljar: Err, I kept it to be consistent with the `JoinedRoomUpdate` variant.

bnjbvr: I'd rather have us use the real full name everywhere, but not a strong opinion here.

poljar: I don't really care, I just assumed that the `JoinedRoomUpdate` was using the short-hand for a reason.

PR #4488, opened 2025-01-08
fix: Fix incorrect debug_struct calls in several places

andybalaam:


Hywan (approved):
Thanks!

BillCarsonFr (approved):
I think there are some missing ones?
`impl fmt::Debug for LeftRoomUpdate {`
I wonder if it is doing the right thing for PendingEdit or BuilderStoreConfig?

poljar:
> I think there are some missing ones? `impl fmt::Debug for LeftRoomUpdate {` I wonder if it is doing the right thing for PendingEdit or BuilderStoreConfig?

That one was done in https://github.com/matrix-org/matrix-rust-sdk/pull/4487.

PR #4490, opened 2025-01-08
fix(sdk): Ensure a gap has been inserted before removing it

Hywan:
This patch fixes a bug where the code assumes a gap has been inserted, and thus, is always present. But this isn't the case. If `prev_batch` is `None`, a gap is not inserted, and so we cannot remove it. This patch checks that `prev_batch` is `Some(_)`, which means the invariant is correct, and the code can remove the gap.

### DIFF
--- a/crates/matrix-sdk/src/event_cache/room/mod.rs
+++ b/crates/matrix-sdk/src/event_cache/room/mod.rs
@@ -431,7 +431,7 @@ impl RoomEventCacheInner {
 
                     let added_unique_events = room_events.push_events(sync_timeline_events.clone());
 
-                    if !added_unique_events {
+                    if !added_unique_events && prev_batch.is_some() {
                         debug!(
                             "not storing previous batch token, because we deduplicated all new sync events"
                         );


BillCarsonFr:
Should fix this then https://github.com/element-hq/element-x-ios/issues/3648?

bnjbvr:
> Should fix this then https://github.com/element-hq/element-x-ios/issues/3648?

To wit: usually those issues are marked as fixed when the app upstreams the SDK, so we don't have the SDK close EX apps issues.

crates/matrix-sdk/src/event_cache/room/mod.rs
```diff
@@ -436,14 +436,24 @@ impl RoomEventCacheInner {
...
+                            // Note: there can't be any race with another task touching the linked
+                            // chunk at this point, because we're using `with_events_mut` which
+                            // guards access to the data.
+                            trace!("removing gap we just inserted");
+
+                            let prev_gap_id = room_events
+                                .rchunks()
+                                .find_map(|c| {
+                                    let gap = as_variant::as_variant!(c.content(), ChunkContent::Gap)?;
+                                    (gap.prev_token == *prev_token).then_some(c.identifier())
```
bnjbvr, line 449: Since we're touching this code, I also made the check more precise: instead of "give me the previous gap, whatever it is", this is now "give me the previous gap with the matching prev-batch token".

andybalaam: I would prefer a named function for the code inside `find_map` because it took me quite a while to figure out what this was doing, but I won't insist on it, and I'm happy for the change to come later if you like the idea.

andybalaam (approved):
Looks good, one optional comment.

crates/matrix-sdk/src/event_cache/room/mod.rs
bnjbvr: Thanks! added a small code comment above this line, expliciting what it does :+1:

PR #4491, opened 2025-01-08
Some WASM-related refactors

jplatte:
Not sufficient for compiling the UI crate on WASM, but smaller PRs = better.


poljar (approved):
Makes sense.

PR #4492, opened 2025-01-09
feat(ffi): move tracing setup from the final client to the ffi layer

stefanceriu:
Having the final clients define the tracing filters / log levels proved to be tricky to keep in check resulting missing logs (e.g. recently introduced modules like the event cache) or unexpected behaviors (e.g. missing panics because we don't set a global filter). As such we decided to move the tracing setup and default definitions over to the rust side and let rust developers have full control over them.

We will now take a general log level and the final client's target name and apply them on top of the rust side defined default. Targets that log more than the requested by default will remain unchanged while the others will increase their log levels to match. Certain targets like `hyper` will be ignored in this step as they're too verbose.


bindings/matrix-sdk-ffi/src/platform.rs
```diff
@@ -245,12 +231,65 @@ pub struct TracingFileConfiguration {
     max_files: Option<u64>,
 }
 
+#[derive(PartialEq, PartialOrd)]
+enum LogTarget {
```
BillCarsonFr, line 235: I think EXA has more target than that https://github.com/element-hq/element-x-android/blob/65ce91a8fbe287564bbdaa969ce9b925edacdb6c/libraries/matrix/api/src/main/kotlin/io/element/android/libraries/matrix/api/tracing/TracingFilterConfiguration.kt#L48-L66

Would be nice to have a `CustomTarget(rawString)`? that would allow to add a new target wihtout having to compile a custom sdk when developping?

stefanceriu: > I think EXA has more target than that

Seems so. The missing ones are `matrix_sdk_ffi::uniffi_api` , `matrix_sdk` and `matrix_sdk::send_queue`. Not sure if we should include them or not to be honest ¯\_(ツ)_/¯ Perhaps keep `matrix_sdk` on info and not add the rest?


> Would be nice to have a CustomTarget(rawString)? 

Thought about exposing an optional `filter` parameter that will just override everything but I figured it's best we keep the api simple as a first step. I'm not sure how often you'll want to debug something with custom tracing on the app side without having a custom sdk version as well, moment in which tweaking the logs becomes easy.

bindings/matrix-sdk-ffi/src/platform.rs
```diff
@@ -259,38 +299,102 @@ pub struct TracingConfiguration {
...
 }
 
+fn build_tracing_filter(config: &TracingConfiguration) -> String {
+    // We are intentionally not setting a global log level because we don't want to
+    // risk 3rd party crates logging sensitive information.
+    // As such we need to make sure that panics will be properly logged.
+    // On 2025-01-08, `log_panics` uses the `panic` target, at the error log level.
+    let mut filters = vec!["panic=error".to_owned()];
+
+    // We're also need to include the target name that the FII client uses
```
bnjbvr, line 309:
```suggestion
    // We also need to include the target name that the FFI client uses
```

bindings/matrix-sdk-ffi/src/platform.rs
bnjbvr: > Thought about exposing an optional filter parameter that will just override everything but I figured it's best we keep the api simple as a first step. I'm not sure how often you'll want to debug something with custom tracing on the app side without having a custom sdk version as well, moment in which tweaking the logs becomes easy.

Could the `TracingConfiguration` include an `extra_logs: Vec<(String, LogLevel)>`, that would be applied as such without any checks? Having some dynamic way to set some logs is definitely a recurring need, and will pay off super quick.

Then we wouldn't need the mandatory `app_target` field in there, as it could be part of the extra logs. (And that gives you the possibility to use different spans if you'd like!)

bindings/matrix-sdk-ffi/src/platform.rs
```diff
@@ -245,12 +231,66 @@ pub struct TracingFileConfiguration {
...
+    MatrixSDKClient,
+    MatrixSDKCrypto,
+    MatrixSDKCryptoAccount,
+    MatrixSDKOIDC,
+    MatrixSDKHTTPClient,
+    MatrixSDKSlidingSync,
+    MatrixSDKBaseSlidingSync,
+    MatrixSDKUITimeline,
+    MatrixSDKEventCache,
+    MatrixSDKEventCacheStore,
```
bnjbvr, line 247: Let's capitalize only the first letter of acronyms: `MatrixSdkFfi`, `MatrixSdkOidc`, and so on and so forth.

Otherwise it seems that `MatrixSDKOIDC` is yelling at me.

bindings/matrix-sdk-ffi/src/platform.rs
```diff
@@ -245,12 +231,66 @@ pub struct TracingFileConfiguration {
...
+            LogTarget::MatrixSDKClient => "matrix_sdk::client",
+            LogTarget::MatrixSDKCrypto => "matrix_sdk_crypto",
+            LogTarget::MatrixSDKCryptoAccount => "matrix_sdk_crypto::olm::account",
+            LogTarget::MatrixSDKOIDC => "matrix_sdk::oidc",
+            LogTarget::MatrixSDKHTTPClient => "matrix_sdk::http_client",
+            LogTarget::MatrixSDKSlidingSync => "matrix_sdk::sliding_sync",
+            LogTarget::MatrixSDKBaseSlidingSync => "matrix_sdk_base::sliding_sync",
+            LogTarget::MatrixSDKUITimeline => "matrix_sdk_ui::timeline",
+            LogTarget::MatrixSDKEventCache => "matrix_sdk::event_cache",
+            LogTarget::MatrixSDKEventCacheStore => "matrix_sdk_sqlite::event_cache_store",
```
bnjbvr, line 264: For the event cache pack, we'd need also `LogTarget::MatrixSdkBaseEventCache`. But I can do that as a follow-up, if you'd like :)

bindings/matrix-sdk-ffi/src/platform.rs
```diff
@@ -245,12 +231,66 @@ pub struct TracingFileConfiguration {
...
+
+const IMMUTABLE_TARGET_LOG_LEVELS: &[LogTarget] = &[LogTarget::Hyper];
+
 #[derive(uniffi::Record)]
 pub struct TracingConfiguration {
-    /// A filter line following the [RUST_LOG format].
-    ///
-    /// [RUST_LOG format]: https://rust-lang-nursery.github.io/rust-cookbook/development_tools/debugging/config_log.html
-    filter: String,
+    /// The desired LogLevel, corresponding to https://docs.rs/log/latest/log/enum.Level.html
```
bnjbvr, line 288: No need to indicate the type, since docs would include a link to the field's type.

```suggestion
    /// The desired log level.
```

bindings/matrix-sdk-ffi/src/platform.rs
```diff
@@ -245,12 +231,66 @@ pub struct TracingFileConfiguration {
...
+    log_level: LogLevel,
+
+    /// The target name under which the FFI client will log messages using
+    /// `tracing::log_event`
```
bnjbvr, line 292:
```suggestion
    /// [`tracing::log_event`].
```

bindings/matrix-sdk-ffi/src/platform.rs
```diff
@@ -259,38 +299,102 @@ pub struct TracingConfiguration {
     write_to_files: Option<TracingFileConfiguration>,
 }
 
+fn build_tracing_filter(config: &TracingConfiguration) -> String {
+    // We are intentionally not setting a global log level because we don't want to
+    // risk 3rd party crates logging sensitive information.
```
bnjbvr, line 304:
```suggestion
    // risk third party crates logging sensitive information.
```

bnjbvr (changes requested):
Thanks a bunch for doing this, you rock! My only blocking comment is about `extra_logs`, the rest are mostly nits.

bindings/matrix-sdk-ffi/src/platform.rs
stefanceriu: Oh yeah, nice idea, let's do it! 💪

bindings/matrix-sdk-ffi/src/platform.rs
stefanceriu: Fair enough 👍

bindings/matrix-sdk-ffi/src/platform.rs
stefanceriu: I've added`MatrixSdkBaseEventCache` that defaults to `Info` like the others and I've also added an `matrix_sdk` wide one that also defaults to info but won't drop lower than that.

bindings/matrix-sdk-ffi/src/platform.rs
stefanceriu: Okay, so I had a play and ended up with passing in *only* an array of `target`s as opposed to `(target, level)` as doing the latter would force the final client to duplicate the level adjustment logic done here. 

If we pass in only the targets we can keep using the same logic while offering just enough flexibility imho.

bnjbvr (approved):
:partying_face:

poljar:
I think that this should have been merged at the same time https://github.com/matrix-org/complement-crypto/pull/159 got merged, now our CI is busted :cry:.

stefanceriu:
Bah, didn't even cross my mind 😬 
Perhaps we should be more leninent on complement crypto as it's an external project.

poljar:
Well complement-crypto was designed with this in mind and it did work out nicely in the past, I don't think we need to change this.

stefanceriu:
Okay then perhaps we should not try to match the branches and instead always run complement crypto on `main` so it blocks PRs?

poljar:
The matching branches are here so you can develop the complement changes and double check that your changes in the SDK and in complement crypto pass the CI.

It's a chicken and egg problem, if we run always against main of complement crypto then we're never able to get the CI green in the PR. Do note, if we merge the complement crypto PR and forget to merge the SDK PR, then we also get a CI failure.

Sadly, I don't see any better way of doing this.

stefanceriu:
> Do note, if we merge the complement crypto PR and forget to merge the SDK PR, then we also get a CI failure.

Ah right, didn't think about that. Well I guess we'll just have to do to the best with what we have.

Issue #4494, opened 2025-01-09
Sometimes `Timeline::paginate_backward()` returns `true` when it is not at the start of the timeline


zecakeh:
This seems to occur reliably with the live timeline when we received only a handful of new messages in a room via `Client::sync()` since the last time that the client was used.

This does not happen if there are no new messages in the room, or if there are more than a handful new messages in the room, so closing the app and relaunching it makes the issue go away. So it seems like it happens when we receive a `timeline` for the room that is not `limited` in the first sync since the client was created.

These are the logs with an account with a single room, and a single new message received via sync, with `RUST_LOG=matrix_sdk::event_cache=trace,matrix_sdk::sync=trace,matrix_sdk_base::event_cache=trace,matrix_sdk_sqlite::event_cache_store=trace,matrix_sdk_ui=trace,warn`:

<details><summary>Logs</summary>
<pre><code>2025-01-09T16:15:27.969965Z TRACE listen_task: matrix_sdk::event_cache: Spawning the listen task
2025-01-09T16:15:27.974414Z TRACE build{room_id="!BbvbHeskfeWkHQsbJJ:beta.matrix.org" track_read_receipts=true}:set_fully_read_event: matrix_sdk_ui::timeline::controller::state: Updating read marker fully_read_event="$FkaI5ODAPrgJvTOGFhoXToS980hyQqkaPBvPS4PRyIk"
2025-01-09T16:15:27.975741Z TRACE live_update_handler{room_id="!BbvbHeskfeWkHQsbJJ:beta.matrix.org" focus="live"}: matrix_sdk_ui::timeline::builder: Spawned the event subscriber task.
2025-01-09T16:15:27.975892Z TRACE live_update_handler{room_id="!BbvbHeskfeWkHQsbJJ:beta.matrix.org" focus="live"}: matrix_sdk_ui::timeline::builder: Waiting for an event.
2025-01-09T16:15:27.976871Z  INFO local_echo_handler{room_id="!BbvbHeskfeWkHQsbJJ:beta.matrix.org" focus="live"}: matrix_sdk_ui::timeline::builder: spawned the local echo handler!
2025-01-09T16:15:27.977783Z TRACE matrix_sdk_ui::timeline::controller: Creating timeline items signal
2025-01-09T16:15:28.136411Z DEBUG call_sync_response_handlers: matrix_sdk::sync: Ran event handlers in 3.389072ms
[...]

PR #4495, opened 2025-01-09
fix(event cache): keep the previous-batch token when we haven't enabled storage

bnjbvr:
Tentative fix for https://github.com/matrix-org/matrix-rust-sdk/issues/4494, tests to be written.

### DIFF
--- a/crates/matrix-sdk/src/event_cache/room/mod.rs
+++ b/crates/matrix-sdk/src/event_cache/room/mod.rs
@@ -323,7 +323,8 @@ impl RoomEventCacheInner {
             // Only keep the previous-batch token if we have a limited timeline; otherwise,
             // we know about all the events, and we don't need to back-paginate,
             // so we wouldn't make use of the given previous-batch token.
-            let prev_batch = if timeline.limited { timeline.prev_batch } else { None };
+            let prev_batch =
+                if !has_storage || timeline.limited { timeline.prev_batch } else { None };
 
             self.append_new_events(
                 timeline.events,


crates/matrix-sdk/src/test_utils/mocks.rs
```diff
@@ -1697,23 +1697,70 @@ impl<'a> MockEndpoint<'a, RoomMessagesEndpoint> {
...
+            template = template.set_delay(delay);
+        }
+
+        let mock = self.mock.respond_with(template);
         MatrixMock { server: self.server, mock }
     }
 }
 
+/// A response to a [`RoomMessagesEndpoint`] query.
+pub struct RoomMessagesResponse {
```
poljar, line 1718: Can we call this a bit differently? It's not the real Ruma response we use everywhere else, it's a test helper. Perhaps `RoomMessageResponseTemplate`?

Hywan: `MockRoomMessages`?

Hywan (approved):
Approving the real commit :-).

crates/matrix-sdk/src/test_utils/mocks.rs
poljar: Well we're not mocking, and it's not a message. It's about the response that you would send out if you were to send a room message. Then we pass that response to the mock server which will mock things for us.

I picked the `Template` naming because wiremock, which we use and wrap here, is using that term as well: https://docs.rs/wiremock/latest/wiremock/struct.ResponseTemplate.html#.

PR #4497, opened 2025-01-09
More small fixes for compiling the UI crate to WASM

jplatte:
Previously: https://github.com/matrix-org/matrix-rust-sdk/pull/4491

I've got one more PR with a slightly larger diff coming, then it should compile!

Signed-off-by: Jonas Platte <jplatte+matrix@posteo.de>

PR #4498, opened 2025-01-09
refactor(ui): Use RPITIT / AFIT for RoomDataProvider

jplatte:
Based on #4497 (that one should be merged first).
I think this should be the last change required to make the UI crate compile on WASM. I'll make a separate PR to add that to CI afterwards.

### DIFF
--- a/crates/matrix-sdk-ui/src/timeline/traits.rs
+++ b/crates/matrix-sdk-ui/src/timeline/traits.rs
@@ -15,14 +15,12 @@
 use std::future::Future;
 
 use eyeball::Subscriber;
-use futures_util::FutureExt as _;
 use indexmap::IndexMap;
 #[cfg(test)]
 use matrix_sdk::crypto::{DecryptionSettings, RoomEventDecryptionResult, TrustRequirement};
 use matrix_sdk::{
     crypto::types::events::CryptoContextInfo, deserialized_responses::TimelineEvent,
-    event_cache::paginator::PaginableRoom, AsyncTraitDeps, BoxFuture, Result, Room,
-    SendOutsideWasm,
+    event_cache::paginator::PaginableRoom, AsyncTraitDeps, Result, Room, SendOutsideWasm,
 };
 use matrix_sdk_base::{latest_event::LatestEvent, RoomInfo};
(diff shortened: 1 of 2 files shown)


crates/matrix-sdk-ui/src/timeline/traits.rs
```diff
@@ -160,128 +164,110 @@ impl RoomDataProvider for Room {
...
-                    debug!("Could not aggregate push context");
-                    None
-                }
+    async fn push_rules_and_context(&self) -> Option<(Ruleset, PushConditionRoomCtx)> {
+        match self.push_context().await {
+            Ok(Some(push_context)) => match self.client().account().push_rules().await {
+                Ok(push_rules) => Some((push_rules, push_context)),
                 Err(e) => {
-                    error!("Could not get push context: {e}");
+                    error!("Could not get push rules: {e}");
```
Hywan, line 222: :+1:

Hywan (approved):
I'm good with these changes. Thanks!

crates/matrix-sdk-ui/src/timeline/traits.rs
jplatte: Err, this is not something I changed..

jplatte: Checked the diff, it's all good. This is just git displaying the change in a very confusing way because indentation changed. This was already correct before.

jplatte:
Ah, I thought closing / re-opening would trigger CI but apparently there's now merge conflicts.

crates/matrix-sdk-ui/src/timeline/traits.rs
Hywan: Oh yeah right, sorry for the noise.

Hywan:
RPITIT is [_Return Position Implementation Trait in Trait_](https://rustc-dev-guide.rust-lang.org/return-position-impl-trait-in-trait.html).

AFIT is _Async Function In Trait_.

+1 for not using acronyms in titles, or at least explain them in the description. Not everyone is aware of their meaning.

PR #4499, opened 2025-01-10
fix: Implement visit_bytes for the Ed25519PublicKey deserialization

poljar:
This fixes the deserialization of the SenderData since it switched to the base64 encoding for serialization of the master key in one of its variants.

The issue was introduced in 5ff556f6c3026ab53b21ba0b0a24749ffa715e17 (#4450).

### DIFF
--- a/crates/matrix-sdk-crypto/src/olm/group_sessions/sender_data.rs
+++ b/crates/matrix-sdk-crypto/src/olm/group_sessions/sender_data.rs
@@ -70,12 +70,30 @@ where
         where
             A: de::SeqAccess<'de>,
         {
-            let mut buf = [0u8; 32];
+            let mut buf = [0u8; Ed25519PublicKey::LENGTH];
+
             for (i, item) in buf.iter_mut().enumerate() {
                 *item = seq.next_element()?.ok_or_else(|| de::Error::invalid_length(i, &self))?;
             }
+
             let key = Ed25519PublicKey::from_slice(&buf).map_err(|e| de::Error::custom(&e))?;
+
             Ok(Box::new(key))
+        }
+
+        fn visit_bytes<E>(self, v: &[u8]) -> Result<Self::Value, E>
+        where
+            E: de::Error,
+        {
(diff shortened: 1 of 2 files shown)


poljar:
> Approving but I would enjoy having a test 🙂

It's in #4500, though my commit has been squashed in there as well.

crates/matrix-sdk-crypto/src/olm/group_sessions/sender_data.rs
```diff
@@ -71,12 +71,30 @@ where
...
             for (i, item) in buf.iter_mut().enumerate() {
                 *item = seq.next_element()?.ok_or_else(|| de::Error::invalid_length(i, &self))?;
             }
+
             let key = Ed25519PublicKey::from_slice(&buf).map_err(|e| de::Error::custom(&e))?;
+
             Ok(Box::new(key))
         }
+
+        fn visit_bytes<E>(self, v: &[u8]) -> Result<Self::Value, E>
```
andybalaam, line 84: Any tips on how to confirm that `visit_bytes` was the only thing we missed? (I'm not very familiar with these bits of serde)

andybalaam (approved):
Looks good, but definitely needs a test.

crates/matrix-sdk-crypto/src/olm/group_sessions/sender_data.rs
poljar: 1. Tell rust-analyzer to implement all the default members.
2. See that all the relevant other types (`Vec<u8>`, `String`, `&'de [u8]`) call `visit_bytes()` or `visit_str()` by default.

image

You could also read the documentation: https://rust-lang.github.io/hashbrown/serde/de/trait.Visitor.html#method.visit_byte_buf. :nerd_face:

andybalaam: Perfect, thank you.

poljar:
Thanks for the test @BillCarsonFr.

PR #4500, opened 2025-01-10
fix(crypto): migration failure due to optimized handling of &[u8]

BillCarsonFr:
Fix migration bug caused by #4450 when &[u8] serialization is optimized.

=> Migration bug was causing messages to fail to decrypt and the sync to fail with the following errors

```
matrix_sdk_crypto::machine: Failed to decrypt a room event: invalid type: byte array, expected a base64 string or an array of 32 bytes
```
```
ERROR matrix_sdk::sync: Received an invalid response: failed to read or write to the crypto store invalid type: byte array, expected a base64 string or an array of 32 bytes
```

Signed-off-by:

### DIFF
--- a/crates/matrix-sdk-crypto/src/olm/group_sessions/sender_data.rs
+++ b/crates/matrix-sdk-crypto/src/olm/group_sessions/sender_data.rs
@@ -77,6 +77,21 @@ where
             let key = Ed25519PublicKey::from_slice(&buf).map_err(|e| de::Error::custom(&e))?;
             Ok(Box::new(key))
         }
+
+        fn visit_bytes<E>(self, v: &[u8]) -> Result<Self::Value, E>
+        where
+            E: de::Error,
+        {
+            if v.len() == Ed25519PublicKey::LENGTH {
+                let mut buf = [0u8; 32];
+                buf.copy_from_slice(v);
+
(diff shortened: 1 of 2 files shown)


crates/matrix-sdk-crypto/src/olm/group_sessions/sender_data.rs
```diff
@@ -77,6 +77,21 @@ where
             let key = Ed25519PublicKey::from_slice(&buf).map_err(|e| de::Error::custom(&e))?;
             Ok(Box::new(key))
         }
+
+        fn visit_bytes<E>(self, v: &[u8]) -> Result<Self::Value, E>
+        where
+            E: de::Error,
+        {
+            if v.len() == Ed25519PublicKey::LENGTH {
+                let mut buf = [0u8; 32];
```
jmartinesp, line 86: We could use `[0u8; Ed25519PublicKey::LENGTH]` here too I think, to it's more explicit.

jmartinesp (approved):
LGTM if I understood the issue correctly 😅 . Left a small comment, and I think we should wait for confirmation from @poljar .

BillCarsonFr:
Closed. The test was added to the existing PR https://github.com/matrix-org/matrix-rust-sdk/pull/4499

PR #4503, opened 2025-01-10
feat(sdk): Allow to set and check whether an image is animated

zecakeh:
Using MSC4230.

While I was in the area I added a second commit to implement `Default` for the `Base*Info` types since they only have optional fields and it is easier to construct them that way when we don't want to set all fields.


poljar (approved):
Looks good.

poljar (changes requested):

zecakeh:
Fixed a typo too.

PR #4506, opened 2025-01-10
fix(store-encryption): Remove an unwrap that snuck in

poljar:
Found during my crusade against unwraps in the codebase.


andybalaam (approved):
Nice.

PR #4519, opened 2025-01-12
feat(sdk): Expose Client::server_versions publicly

zecakeh:
We have one specific use case at the moment for this in Fractal, to build the [UIAA fallback URL](https://spec.matrix.org/v1.13/client-server-api/#fallback) depending on the Matrix version.  Currently we always use the r0 version that is hardcoded in our code.

But I also don't see why it should not be public.


poljar (approved):
Thanks, indeed should be public.

PR #4521, opened 2025-01-13
feat(event cache): remove events when forgetting a room

bnjbvr:
Does what it says on the tin, and fixes a bug where clearing a room in the in-memory linked chunk would clear *all* rooms at once.

Part of #3280.


jmartinesp (approved):
LGTM, thanks!

PR #4525, opened 2025-01-13
fix(timeline): Add UTDs to the timeline conditionally

jmartinesp:
Should fix https://github.com/element-hq/element-x-ios/issues/3669.

Take into account the `should_add` parameter for including UTDs in the timeline, as is done for other types of events.

Signed-off-by:


jmartinesp:
The test matrix_sdk_integration_testing::tests::sliding_sync::room::test_delayed_invite_response_and_sent_message_decryption seems to be consistently failing [here](https://github.com/matrix-org/matrix-rust-sdk/blob/e19bdbfd59141de101f7d867e6fd398cf4a47667/testing/matrix-sdk-integration-testing/src/tests/sliding_sync/room.rs#L901) because the timeline doesn't contain any event, although the only actual code change in this PR is the `if should_add` check and I've confirmed with a debugger that it's `true`, so it should behave just as before 🫤 .

Hywan (approved):
Thanks for the fix!

PR #4526, opened 2025-01-13
tests: Add an encrypted snapshot of a SQLite db for regression tests

BillCarsonFr:
Adds a crypto SQLite database snapshot for regression tests. 
There was a similar test but with a clear database, this one is encrypted. This allows to cover the specific serialisation used
with encryption (rmp_serde). This would allow to detect regressions like this one #4499 that went un-noticed

Signed-off-by:


jmartinesp (approved):
Thanks!

PR #4529, opened 2025-01-14
feat(room): Add fn Room::own_membership_details

jmartinesp:
This will retrieve the room member info of both the current user and the info for the sender of the current user's room member event and make it accessible in the FFI layer.

This is needed for `RoomPreview` to display messages such as 'you were banned from this room by Foo'.

Signed-off-by:

### DIFF
--- a/crates/matrix-sdk/src/room/mod.rs
+++ b/crates/matrix-sdk/src/room/mod.rs
@@ -2852,6 +2852,35 @@ impl Room {
         Ok(Invite { invitee, inviter })
     }
 
+    /// Get the membership details for the current user.
+    ///
+    /// Returns:
+    ///     - If the current user was present in the room, a tuple of the
+    ///       current user's [`RoomMember`] info and the member info of the
+    ///       sender of that member event.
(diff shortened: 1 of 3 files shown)


crates/matrix-sdk/src/room/mod.rs
```diff
@@ -2852,6 +2852,35 @@ impl Room {
...
     }
 
+    /// Get the membership details for the current user.
+    ///
+    /// Returns:
+    ///     - If the current user was present in the room, a tuple of the
+    ///       current user's [`RoomMember`](RoomMember) info and the member info
+    ///       of the sender of that member event.
+    ///     - If the current user is not present, an error.
+    pub async fn own_membership_details(&self) -> Result<(RoomMember, Option<RoomMember>)> {
```
poljar, line 2862: Can we get a simple smoke test for this method, I know it's a pretty simple method but it would still be nice to test its basic functionality.

poljar (approved):
Looks sensible, a test for the method would be nice.

poljar (approved):

PR #4530, opened 2025-01-14
test: keep a single sync in `test_delayed_invite_response_and_sent_message_decryption()`

bnjbvr:
This removes one sync that happens in the background, because it's likely spurious and may be confusing the server about what's been seen by the current client.


testing/matrix-sdk-integration-testing/src/tests/sliding_sync/room.rs
```diff
@@ -853,74 +853,88 @@ async fn test_delayed_invite_response_and_sent_message_decryption() -> Result<()
...
+            if let Ok(Some(diffs)) = timeout(Duration::from_secs(3), room_stream.next()).await {
+                for diff in diffs {
+                    diff.apply(&mut rooms);
+                }
+            } else {
+                panic!("bob never founds out about the room");
+            }
         }
-    }
+    };
```
andybalaam, line 890: I would prefer if the contents of this {} were a separate function.

andybalaam (approved):
Looks good; one suggestion.

PR #4531, opened 2025-01-14
fix: Remove support for MSC3575

Hywan:
MSC3575 is the “first” and “old” version of sliding sync, implemented by the sliding sync proxy and probably other homeservers. MSC4186 is the “next” and “new” version of sliding sync, being implemented inside Synapse and probably other homeservers.

The Matrix Rust SDK was supporting both versions, aiming at giving time to homeservers to transition from the experimental MSC3575 to the stable, production ready MSC4186.

It is time to say good bye to MSC3575. ~This is what the first patch does.~

~This patch also needs to update ruma (to include https://github.com/ruma/ruma/pull/1995). Updating ruma involves a bit of breaking changes, it's addressed in a standalone second patch.~


crates/matrix-sdk-base/src/read_receipts.rs
```diff
@@ -241,10 +241,9 @@ impl RoomReadReceipts {
         let mut counting_receipts = false;
 
         for event in events {
-            // The sliding sync proxy sometimes sends the same event multiple times, so it
-            // can be at the beginning and end of a batch, for instance. In that
-            // case, just reset every time we see the event matching the
-            // receipt. NOTE: SS proxy workaround.
+            // Sliding sync sometimes sends the same event multiple times, so it can be at
+            // the beginning and end of a batch, for instance. In that case, just reset
+            // every time we see the event matching the
```
bnjbvr, line 246: nit: you lost a word here :D

crates/matrix-sdk-base/src/sliding_sync/http.rs
```diff
@@ -12,38 +12,8 @@
...
-    }
-}
-
-/// HTTP types from MSC4186.
-pub mod msc4186 {
-    pub use ruma::api::client::sync::sync_events::v5::*;
-}
-
-pub use msc4186::*;
+pub use ruma::api::client::sync::sync_events::v5::*;
```
bnjbvr, line 19: Can we get rid of this file now, and just use `use` with the ruma full path?

crates/matrix-sdk/src/client/futures.rs
```diff
@@ -67,15 +67,6 @@ impl<R> SendRequest<R> {
         self
     }
 
-    /// Replace this request's target (homeserver) with a custom one.
-    ///
-    /// This is useful at the moment because the current sliding sync
-    /// implementation uses a proxy server.
-    pub fn with_homeserver_override(mut self, homeserver_override: Option<String>) -> Self {
-        self.homeserver_override = homeserver_override;
```
bnjbvr, line 75: The field could also be removed, right?

crates/matrix-sdk/src/event_cache/room/mod.rs
```diff
@@ -248,11 +248,9 @@ impl RoomEventCacheInner {
         for raw_event in account_data {
             match raw_event.deserialize() {
                 Ok(AnyRoomAccountDataEvent::FullyRead(ev)) => {
-                    // Sometimes the sliding sync proxy sends many duplicates of the read marker
+                    // Sometimes sliding sync sends many duplicates of the read marker
                     // event. Don't forward it multiple times to avoid clutter
                     // the update channel.
```
bnjbvr, line 253: We're not sure the native implementation does that, but my feeling is that we can remove this comment and keep the code as is, maybe with a much shorter comment that sounds less accusatory.

bnjbvr (changes requested):
Exciting! I'd really like the Ruma update to happen separately, because it involves non-trivial changes related to reply fallbacks that I'd like to review without other changes, pretty please :pleading_face: :pray:

Hywan:
Yup, sure, ruma update has been extracted in https://github.com/matrix-org/matrix-rust-sdk/pull/4532.

crates/matrix-sdk-base/src/sliding_sync/http.rs
Hywan: I've removed this file, but `ruma::api::client::sync::sync_events::v5` is re-exported as `http` inside `matrix_sdk_base::sliding_sync`. I think it is still comfortable to use `matrix_sdk_base::sliding_sync::http` to get all the HTTP types for sliding sync without having to look for inside Ruma. It's just a module alias. What do you think?

bnjbvr: Hmm, I do get it's to avoid more changes that you're making this alias public, but if it's now part of our public API, this has semantic versioning implications: any changes to the underlying data structures becomes a breaking change in our public API, that we should mention, be aware of, etc.

Plus, it's another layer of indirection when browsing the code: if I look at a type `http::Something`, I may now think it's defined somewhere along here, find where the `http` declaration is, and find out it's a renaming from another module.

I'd rather not do that, if we can avoid it, and use the full paths in all places we use those types, making it super explicit and direct and non-ambiguous and not in our public API.

Hywan: As you prefer Sir.

Hywan: Done.

bnjbvr (approved):
yay :fire: :fire: :fire:

PR #4532, opened 2025-01-14
chore: Update to latest Ruma version

Hywan:
Extracted from https://github.com/matrix-org/matrix-rust-sdk/pull/4531.

This patch updates ruma to the latest commit at the time of writing.

I'm really not sure about what I'm doing, so please be careful during the review 😛.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -57,7 +57,9 @@ proptest = { version = "1.5.0", default-features = false, features = ["std"] }
 rand = "0.8.5"
 reqwest = { version = "0.12.4", default-features = false }
 rmp-serde = "1.3.0"
-ruma = { git = "https://github.com/ruma/ruma", rev = "b266343136e8470a7d040efc207e16af0c20d374", features = [
+# Be careful to use commits from the https://github.com/ruma/ruma/tree/ruma-0.12
+# branch until a proper release with breaking changes happens.
+ruma = { git = "https://github.com/ruma/ruma", rev = "b868438f5d91918e97d2c3f64d7c82a0d86d29d4", features = [
     "client-api-c",
     "compat-upload-signatures",
     "compat-user-id",
@@ -72,7 +74,7 @@ ruma = { git = "https://github.com/ruma/ruma", rev = "b266343136e8470a7d040efc20
     "unstable-msc4140",
     "unstable-msc4171",
 ] }
-ruma-common = { git = "https://github.com/ruma/ruma", rev = "b266343136e8470a7d040efc207e16af0c20d374" }
+ruma-common = { git = "https://github.com/ruma/ruma", rev = "b868438f5d91918e97d2c3f64d7c82a0d86d29d4" }
 serde = "1.0.151"
 serde_html_form = "0.2.0"
 serde_json = "1.0.91"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4789,7 +4789,7 @@ dependencies = [
 [[package]]
 name = "ruma"
 version = "0.12.0"
-source = "git+https://github.com/ruma/ruma?rev=b266343136e8470a7d040efc207e16af0c20d374#b266343136e8470a7d040efc207e16af0c20d374"
+source = "git+https://github.com/ruma/ruma?rev=b868438f5d91918e97d2c3f64d7c82a0d86d29d4#b868438f5d91918e97d2c3f64d7c82a0d86d29d4"
 dependencies = [
  "assign",
  "js_int",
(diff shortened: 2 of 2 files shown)


poljar:
> This patch updates ruma to the latest commit at the time of writing.

Latest commit on the stable branch of Ruma or on the main branch? The main branch won't see a release in the next 6ish months.

The Ruma folks very nicely provide a stable branch on which they backport non-breaking changes: https://github.com/ruma/ruma/tree/ruma-0.12.

What exactly do you need from the Ruma bump?

richvdh:
I feel like this probably wants a changelog entry or two, assuming that the ruma changes result in some externally-visible change?

Hywan:
@poljar We need this, https://github.com/ruma/ruma/pull/1995.

poljar:
> @poljar Last commit from `main`. We need this, [ruma/ruma#1995](https://github.com/ruma/ruma/pull/1995).

That has been backported to the release branch: https://github.com/ruma/ruma/commit/b868438f5d91918e97d2c3f64d7c82a0d86d29d4.

Please use that rev instead.

crates/matrix-sdk/src/room/edit.rs
```diff
@@ -188,14 +186,10 @@ async fn make_edit_event<S: EventSource>(
                 });
             };
 
-            let mentions = original.content.mentions.clone();
-            let replied_to_original_room_msg =
-                extract_replied_to(source, room_id, original.content.relates_to).await;
+            let mentions = original.content.mentions;
```
bnjbvr, line 189: Can we inline this variable now?

crates/matrix-sdk/src/room/edit.rs
```diff
@@ -662,15 +583,12 @@ mod tests {
 
         cache.events.insert(event_id.to_owned(), event);
 
-        let room_id = room_id!("!galette:saucisse.bzh");
```
bnjbvr, line 665: you don't seem to like my _galette saucisse_ :imp:

crates/matrix-sdk-ui/src/timeline/mod.rs
```diff
@@ -104,8 +104,6 @@ pub struct RepliedToInfo {
     event_id: OwnedEventId,
     /// The sender of the event to reply to.
     sender: OwnedUserId,
-    /// The timestamp of the event to reply to.
-    timestamp: MilliSecondsSinceUnixEpoch,
     /// The content of the event to reply to.
     content: ReplyContent,
```
bnjbvr, line 108: Instead of the full `content`, I think this could store directly the thread information as an `Option`; then we could likely get rid of `ReplyContent` too, which would be nice!

bnjbvr (approved):
LGTM, modulo tweaking the Ruma repo source that @poljar asked for :+1:

Cargo.toml
```diff
@@ -57,7 +57,7 @@ proptest = { version = "1.5.0", default-features = false, features = ["std"] }
 rand = "0.8.5"
 reqwest = { version = "0.12.4", default-features = false }
 rmp-serde = "1.3.0"
-ruma = { git = "https://github.com/ruma/ruma", rev = "b266343136e8470a7d040efc207e16af0c20d374", features = [
+ruma = { git = "https://github.com/ruma/ruma", rev = "5c9ef9f425d8f8b330e46d215c683bf799a28d50", features = [
```
zecakeh, line 60: Maybe a comment should be added to only update Ruma using the [`ruma-0.12`](https://github.com/ruma/ruma/tree/ruma-0.12) branch, to make sure no one tries to update using the `main` branch before we plan to make a breaking release?

poljar: That's a good idea.

richvdh (dismissed):
I'm sorry you found it a chore.

Please can we have a changelog?

poljar:
> I'm sorry you found it a chore.

Can we tone down the snark please?

> Please can we have a changelog?

All of the reviews are moot since this PR points to the main branch which has a bunch of breaking changes, those will be gone once we depend on the `0.12` branch of Ruma. So please a bit of patience, it's very likely that no significant changes happened due to this.

richvdh:
> So please a bit of patience, it's very likely that no significant changes happened due to this.

Sorry, I didn't mean to rush things. I saw it had an approving review from Benji and thought it was ready to go other than the pending review from https://github.com/orgs/matrix-org/teams/rust-crypto-reviewers, so my intention was to turn my earlier request (which I thought might have been missed) into a request for change.

If there are no significant changes, then of course it's fine to omit a changelog. (It feels a bit surprising that we'd bother updating ruma if it didn't result in *some* externally visible change, but I'm happy to take your word for it. I'd just like explicit confirmation that the lack of changelog is deliberate rather than accidental.)

bnjbvr:
For the record, we'd need some kind of changelog entry here to indicate that we stop supporting reply fallbacks in general; this is a breaking change in terms of behavior, for external users.

richvdh:
> > I'm sorry you found it a chore.
> 
> Can we tone down the snark please?

To address this briefly: my comment was meant to be a light-hearted quip. I can see now the humour didn't come across very well and rather felt snarky. Poor judgement on my part; my apologies.

Hywan:
Okay okay, so, I've update this PR to include a single commit that updates Ruma to https://github.com/ruma/ruma/commit/b868438f5d91918e97d2c3f64d7c82a0d86d29d4. No extra modification on our codebase since there is no breaking changes.

bnjbvr (approved):
yay

zecakeh (approved):

PR #4533, opened 2025-01-15
feat(event cache): handle redacted redactions in the `AllEventsCache`

bnjbvr:
This is unlikely that it will affect us, so not worth adding a test IMO, but for the sake of completeness: this handles redacted redactions in the `AllEventsCache` too.


jmartinesp (approved):
LGTM, thanks!

PR #4534, opened 2025-01-15
fix(tests): Fix a flaky test by marking a room's members as synced.

andybalaam:
This is intended to prevent the test
`test_when_user_in_verification_violation_becomes_verified_we_report_it` flaking. I found that sometimes when it called `Room::members` the result was empty due to it trying to fetch the answer from the server. This change prevents that behaviour. I don't know why the behaviour was inconsistent before.


jmartinesp (approved):
LGTM, thanks!

PR #4535, opened 2025-01-15
refactor(live_location_share): exclude live location events of own user

torrybr:
This change ensures that the user's own live location events are excluded from the location sharing stream. Since the user's location is already represented on the map by the blue dot, processing their own events is redundant and unnecessary.

### DIFF
--- a/crates/matrix-sdk/src/room/mod.rs
+++ b/crates/matrix-sdk/src/room/mod.rs
@@ -3244,6 +3244,9 @@ impl Room {
     ///
     /// The returned observable will receive the newest event for each sync
     /// response that contains an `m.beacon` event.
+    ///
+    /// Returns a stream of `ObservableLiveLocation` events from other users
+    /// in the room, excluding the live location events of the room's own user.
     pub fn observe_live_location_shares(&self) -> ObservableLiveLocation {
         ObservableLiveLocation::new(&self.client, self.room_id())
     }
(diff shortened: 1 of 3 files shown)


crates/matrix-sdk/src/room/mod.rs
```diff
@@ -3244,6 +3244,9 @@ impl Room {
     ///
     /// The returned observable will receive the newest event for each sync
     /// response that contains an `m.beacon` event.
+    ///
+    /// Returns a stream of `ObservableLiveLocation` events from other users
```
poljar, line 3248: Let's linkify this, makes things much more pleasant to read and you can always quickly go to the docs of the referenced type.

```suggestion
    /// Returns a stream of [`ObservableLiveLocation`] events from other users
```

crates/matrix-sdk/tests/integration/room/beacon/mod.rs
```diff
@@ -374,3 +374,95 @@ async fn test_observe_single_live_location_share() {
...
+
+    let current_time = MilliSecondsSinceUnixEpoch::now();
+    let millis_time = current_time
+        .to_system_time()
+        .unwrap()
+        .duration_since(UNIX_EPOCH)
+        .expect("Time went backwards")
+        .as_millis() as u64;
+
+    mock_sync(
```
poljar, line 392: Could this please be rewritten using the `MatrixMockServer`, I now added plenty of docs to it so it shouldn't be hard to figure out how that can be translated to it.

Though if you feel lost don't fret to shout.

poljar (changes requested):
Makes sense, though the test could be improved with our new mocking API.

Issue #3872, opened 2024-08-21
Event cache: Move decryption retry in there


bnjbvr:
Right now, it's the Timeline API that takes care of retrying decryption of an event. This has a few bad consequences:

- retrying decryption, while it could be done in other contexts than the Timeline API, can only happen there
- some observers (e.g. the UTD hook) want to know when a previously-UTD event could be decrypted; this happens in the Timeline code too, which is an abstraction leak
- some other code basically retries decryption on its own: the latest event code in the base crate does that

I'm proposing that we move all of this over to the Event Cache. An event in a linked chunk / on disk would be replaced by its decrypted equivalent, if needs be, and we can implement a few enhancements mentioned in a few other issues too.

## High-level rough plan

Names to be refined.

- We have the UTD hook / manager, which helps notifying whenever there's a UTD.
  - It should be moved in a central location, could be in the `Client` for instance.
- The timeline needs a way to call into the UTD hook, to report that a UTD event has been *displayed*.
- I propose to have a new `Redecryptor` component which role is to react to
  1. new UTD events coming from sync
  2. new interesting events / notifications that some UTDs may be decryptable now
  - generic over a new trait `RedecryptorCtx`, which should help testing the component in isolation:

```rust
trait RedecryptorCtx {
  /// Returns the current list of UTD events known by the context.
  /// For instance, the event cache can load those from the database, deserializing only the event type.
  async fn current_utd_events(&self) -> Result<Vec<Event>, SomeErrorType>;

  /// Called back whenever UTD events have been decrypted by the `Redecryptor`.
  async fn on_resolved_utds(&self, events: Vec<Event>) -> Result<(), SomeErrorType>;
}
```
- The event cache will implement the above trait.
  - `on_resolved_utds` will replace now decrypted events in the linked chunk and/or in memory.
    - also trigger an update for observers like the timeline (using `RoomEvents::replace_event_at` + `Update::ReplaceItem` for the store, as it's done [here](https://github.com/matrix-org/matrix-rust-sdk/blob/5acaaf5865d2940903f8511ec99ab9602815f967/crates/matrix-sdk/src/event_cache/room/mod.rs#L1313))
  - `current_utd_events` can either load each time from the DB and deserialize only the event type… or do something better
    - we could index the event type in the database, as a separate column in the `events` table
    - we could maintain a list of UTDs: load it from DB first, then keep it in sync (after we received new UTDs from sync, or we successfully decrypted some previous UTD)
[...]

PR #4536, opened 2025-01-15
feat(event cache): redact previously-seen events in the linked chunk too

bnjbvr:
This makes it possible to:

- replace one item in a linked chunk, and have all the consumers handle the update. This will be useful for the second part of the PR (applying redactions to the linked chunk), and in the future to replace decrypted events ([when decryption happens in the event cache](https://github.com/matrix-org/matrix-rust-sdk/issues/3872)).
- redact, in memory and in storage, an event by its redacted form, so that it's not possible to read it on disk.

We _could_ add a migration to apply redactions for existing users of the nightly cache. I'd rather do this as a followup, since this PR is growing a bit large already (and another possibility is also to clear all events stored in the event cache, as a dirty yet safe migration — we might need this because I spotted an opportunity to not persist unneeded data from storage).

### DIFF
--- a/crates/matrix-sdk-common/src/linked_chunk/as_vector.rs
+++ b/crates/matrix-sdk-common/src/linked_chunk/as_vector.rs
@@ -364,6 +364,14 @@ impl UpdateToVectorDiff {
                     }
                 }
 
+                Update::ReplaceItem { at: position, item } => {
+                    let (offset, (_chunk_index, _chunk_length)) = self.map_to_offset(position);
+
+                    // The chunk length doesn't change.
+
+                    diffs.push(VectorDiff::Set { index: offset, value: item.clone() });
+                }
+
                 Update::RemoveItem { at: position } => {
                     let (offset, (_chunk_index, chunk_length)) = self.map_to_offset(position);
 
@@ -484,15 +492,7 @@ mod tests {
         assert_eq!(diffs, expected_diffs);
 
         for diff in diffs {
-            match diff {
-                VectorDiff::Insert { index, value } => accumulator.insert(index, value),
-                VectorDiff::Append { values } => accumulator.append(values),
-                VectorDiff::Remove { index } => {
-                    accumulator.remove(index);
-                }
-                VectorDiff::Clear => accumulator.clear(),
-                diff => unimplemented!("{diff:?}"),
-            }
+            diff.apply(accumulator);
         }
     }
 
@@ -708,6 +708,31 @@ mod tests {
         assert_eq!(
             accumulator,
             vector!['m', 'a', 'w', 'x', 'y', 'b', 'd', 'i', 'j', 'k', 'l', 'e', 'f', 'g', 'z', 'h']
+        );
+
+        // Replace element 8 by an uppercase J.
+        linked_chunk
+            .replace_item_at(linked_chunk.item_position(|item| *item == 'j').unwrap(), 'J')
+            .unwrap();
+
+        assert_items_eq!(
+            linked_chunk,
+            ['m', 'a', 'w'] ['x'] ['y', 'b'] ['d'] ['i', 'J', 'k'] ['l'] ['e', 'f', 'g'] ['z', 'h']
+        );
+
+        // From an `ObservableVector` point of view, it would look like:
+        //
+        // 0   1   2   3   4   5   6   7   8   9   10  11  12  13  14  15  16
+        // +---+---+---+---+---+---+---+---+---+---+---+---+---+---+---+---+
+        // | m | a | w | x | y | b | d | i | J | k | l | e | f | g | z | h |
+        // +---+---+---+---+---+---+---+---+---+---+---+---+---+---+---+---+
+        //                                 ^^^^
+        //                                 |
+        //                                 new!
+        apply_and_assert_eq(
+            &mut accumulator,
+            as_vector.take(),
+            &[VectorDiff::Set { index: 8, value: 'J' }],
         );
 
         // Let's try to clear the linked chunk now.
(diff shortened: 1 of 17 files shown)


crates/matrix-sdk-common/src/linked_chunk/as_vector.rs
```diff
@@ -484,15 +492,7 @@ mod tests {
...
-            match diff {
-                VectorDiff::Insert { index, value } => accumulator.insert(index, value),
-                VectorDiff::Append { values } => accumulator.append(values),
-                VectorDiff::Remove { index } => {
-                    accumulator.remove(index);
-                }
-                VectorDiff::Clear => accumulator.clear(),
-                diff => unimplemented!("{diff:?}"),
-            }
+            diff.apply(accumulator);
```
Hywan, line 495: 👍

crates/matrix-sdk-common/src/linked_chunk/mod.rs
```diff
@@ -528,6 +528,46 @@ impl<const CAP: usize, Item, Gap> LinkedChunk<CAP, Item, Gap> {
...
+                if item_index > current_items.len() {
+                    return Err(Error::InvalidItemIndex { index: item_index });
+                }
+
+                if let Some(updates) = self.updates.as_mut() {
+                    updates.push(Update::ReplaceItem {
+                        at: Position(chunk_identifier, item_index),
+                        item: item.clone(),
+                    });
+                }
```
Hywan, line 562: Push the update after `current_items[item_index] = item;`: do the action before triggering the update.

bnjbvr: This means we have to clone *before*, and the original thing is unused if `self.updates.as_mut().is_none()`. Plus, the updates aren't reflected in real-time, they're buffered until they're consumed, as far as I can tell? So this should be fine to keep as is. I can add a comment explaining that this avoids a clone if you'd like.

crates/matrix-sdk-common/src/linked_chunk/mod.rs
```diff
@@ -528,6 +528,46 @@ impl<const CAP: usize, Item, Gap> LinkedChunk<CAP, Item, Gap> {
...
+            .chunk_mut(chunk_identifier)
+            .ok_or(Error::InvalidChunkIdentifier { identifier: chunk_identifier })?;
+
+        match &mut chunk.content {
+            ChunkContent::Gap(..) => {
+                return Err(Error::ChunkIsAGap { identifier: chunk_identifier })
+            }
+
+            ChunkContent::Items(current_items) => {
+                if item_index > current_items.len() {
```
andybalaam, line 553: Should this be `>=`?

crates/matrix-sdk-common/src/linked_chunk/as_vector.rs
```diff
@@ -710,6 +710,31 @@ mod tests {
             vector!['m', 'a', 'w', 'x', 'y', 'b', 'd', 'i', 'j', 'k', 'l', 'e', 'f', 'g', 'z', 'h']
         );
 
+        // Replace element 8 by an uppsercase J.
```
andybalaam, line 713: uppsercase -> uppercase

crates/matrix-sdk-common/src/linked_chunk/updates.rs
```diff
@@ -79,6 +79,18 @@ pub enum Update<Item, Gap> {
         items: Vec<Item>,
     },
 
+    /// An item has been replaced in the linked chunk.
+    ///
+    /// The `at` position MUST resolve to the actual position an existing *item*
+    /// (not a gap).
+    ReplaceItem {
```
andybalaam, line 86: Presumably the code doesn't compile at this commit because `match` statements elsewhere don't include `ReplaceItem`? If so, I would mildly prefer if this commit included `todo` impls to make it compile. This will help with any future bisect we need to do.

crates/matrix-sdk/tests/integration/event_cache.rs
```diff
@@ -1453,3 +1457,129 @@ async fn test_dont_delete_gap_that_wasnt_inserted() {
...
+        assert_let!(VectorDiff::Set { index: 0, value: redacted_event } = &diffs[1]);
+        let ev = redacted_event.raw().deserialize().unwrap();
+        assert_let!(
+            AnySyncTimelineEvent::MessageLike(AnySyncMessageLikeEvent::RoomMessage(ev)) = ev
+        );
+        // The event has been redacted!
+        assert_matches!(ev.as_original(), None);
+    }
+
+    // And done for now.
```
andybalaam, line 1533: I'd prefer (again, mildly) separate test for the different cases, with names that describe the different things they check for.

andybalaam (approved):
Looks good. I'm not hugely confident I understand all the implications, but the code looks good. Some small suggestions and questions, nothing blocking.

crates/matrix-sdk-common/src/linked_chunk/updates.rs
bnjbvr: Yep, will fold a few commits together :+1:

crates/matrix-sdk-common/src/linked_chunk/mod.rs
bnjbvr: Oof, good catch. Seems it's preexisting in a few other places too…

andybalaam: Yay, I did something useful today! :tada: A test for this case would be awesome.

bnjbvr: Of course! I added one regression test in the force-pushed commits, with also some other smoke tests (viz. calling this function on a gap chunk).

PR #4537, opened 2025-01-15
crypto: add some more documentation to `DeviceKeys`

richvdh:
This confused me for a while, so I thought more documentation might help.

### DIFF
--- a/crates/matrix-sdk-crypto/src/types/device_keys.rs
+++ b/crates/matrix-sdk-crypto/src/types/device_keys.rs
@@ -38,6 +38,12 @@ use super::{EventEncryptionAlgorithm, Signatures};
 /// Specification, encapsulating essential elements such as the public device
 /// identity keys.
 ///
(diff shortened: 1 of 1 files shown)


crates/matrix-sdk-crypto/src/types/device_keys.rs
```diff
@@ -190,6 +196,8 @@ impl From<Ed25519PublicKey> for DeviceKey {
     }
 }
 
+/// A de/serialization helper for [`DeviceKeys`] which maps the `keys` to/from
+/// [`DeviceKey`]s.
```
bnjbvr, line 200: very light suggestion: One of the two mentions to `DeviceKey` seems redundant; you get to choose which one :D

bnjbvr (approved):

crates/matrix-sdk-crypto/src/types/device_keys.rs
richvdh: one is a mention of `DeviceKey`, the other `DeviceKeys`. They are quire different!

PR #4540, opened 2025-01-16
test: add support for dehydrated devices to `KeyQueryResponseTemplate`

richvdh:
#4476 added some test helpers to generate `/keys/query` responses. We're going to need to test dehydrated devices, so this PR adds support for that.


andybalaam (approved):
Great, thanks!

PR #4541, opened 2025-01-16
test: simplify examples for `KeyQueryResponseTemplate`

richvdh:
Generating keys from slices rather than base64 is easier.

Also, s/builder/template/.

### DIFF
--- a/testing/matrix-sdk-test/src/test_json/keys_query_sets.rs
+++ b/testing/matrix-sdk-test/src/test_json/keys_query_sets.rs
@@ -31,20 +31,20 @@ use crate::{
 /// # use vodozemac::{Curve25519PublicKey, Ed25519SecretKey};
 /// # use matrix_sdk_test::test_json::keys_query_sets::{KeyQueryResponseTemplate, KeyQueryResponseTemplateDeviceOptions};
 ///
-/// let pub_curve_key = "PBo2nKbink/HxgzMrBftGPogsD0d47LlIMsViTpCRn4";
-/// let ed25519_key = "yzj53Kccfqx2yx9lcTwaRfPZX+7jU19harsDWWu5YnM";
+/// // Note that (almost) any 32-byte sequence can be used as a private Ed25519 or Curve25519 key.
+/// // You can also use an arbitrary 32-byte sequence as a *public* key though of course you will
+/// // not know the private key it corresponds to (if indeed there is one).
 ///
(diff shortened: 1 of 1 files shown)


testing/matrix-sdk-test/src/test_json/keys_query_sets.rs
```diff
@@ -31,60 +31,49 @@ use crate::{
...
+/// // You can also use an arbitrary 32-byte sequence as a *public* key though of course you will
+/// // not know the private key it corresponds to (if indeed there is one).
 ///
-/// let builder = KeyQueryResponseTemplate::new(owned_user_id!("@alice:localhost"))
+/// let template = KeyQueryResponseTemplate::new(owned_user_id!("@alice:localhost"))
 ///     .with_device(
 ///         device_id!("TESTDEVICE"),
-///         &Curve25519PublicKey::from_base64(pub_curve_key).unwrap(),
-///         &Ed25519SecretKey::from_base64(ed25519_key).unwrap(),
+///         &Curve25519PublicKey::from(b"curvepubcurvepubcurvepubcurvepub".to_owned()),
```
jmartinesp, line 41: Should this be `from_slice` too?

richvdh: no. For reasons I don't entirely understand, `Curve25519PublicKey` and `Ed25519PublicKey` have slightly different interfaces to `Curve25519SecretKey` and `Ed25519SecretKey`.

[`Curve25519PublicKey::from_slice`](https://docs.rs/vodozemac/latest/vodozemac/struct.Curve25519PublicKey.html#method.from_slice) takes a slice of unspecified length, and returns an Error if the slice is anything other than 32 bytes long. I'm trying to avoid an `unwrap()` here, so that's no good.

We could use [`Curve25519PublicKey::from_bytes`](https://docs.rs/vodozemac/latest/vodozemac/struct.Curve25519PublicKey.html#method.from_bytes), but that's basically the same as `::from`, and I'm not sure it's any clearer. (Annoyingly, both take an *owned* `[u8; 32]`, which means we need the `to_owned()` call here.)

jmartinesp: Oh, I see, the `Curve25519PublicKey::from_slice` takes a slice of any length and that's why it returns a `Result`, while `Curve25519PublicKey::from` takes a slice of length 32.

jmartinesp (approved):

PR #4542, opened 2025-01-16
chore(ffi): silence useless logs coming out of the ffi crate

stefanceriu:
Setting the default log level to `debug` results in logs like:

```
log: log_event
log: latest_event
log: log_event
log: log_event
log: room_info
log: latest_event
log: log_event
log: room_info
```

Presumably they're coming out of the custom tracing configuration and we definitely don't need them


poljar (approved):
Ah, it's telling you which function is called or something like that.

Makes sense.

PR #4543, opened 2025-01-16
Refactor and clean up the SyncService

poljar:
These are mostly internal changes and don't contain any behavioral change, except for some edge cases that are now a bit better handled.

A review commit by commit should ease the task a bit, but reading the whole file shouldn't be too terrible either, it's not that big.

### DIFF
--- a/crates/matrix-sdk-ui/Cargo.toml
+++ b/crates/matrix-sdk-ui/Cargo.toml
@@ -61,6 +61,7 @@ matrix-sdk = { workspace = true, features = ["testing", "sqlite"] }
 matrix-sdk-test = { workspace = true }
 stream_assert = { workspace = true }
 tempfile = { workspace = true }
+url = { workspace = true }
 wiremock = { workspace = true }
 
 [lints]
--- a/crates/matrix-sdk-ui/src/sync_service.rs
+++ b/crates/matrix-sdk-ui/src/sync_service.rs
@@ -360,15 +360,16 @@ impl SyncService {
         let mut inner = self.inner.lock().await;
 
         // Only (re)start the tasks if any was stopped.
-        if matches!(self.state.get(), State::Running) {
-            // It was already true, so we can skip the restart.
-            return;
+        match inner.state.get() {
+            // If we're already running, there's nothing to do.
+            State::Running => (),
+            State::Idle | State::Terminated | State::Error => {
+                trace!("starting sync service");
+
+                inner.supervisor = Some(SyncTaskSupervisor::new(&inner).await);
+                inner.state.set(State::Running);
+            }
         }
-
-        trace!("starting sync service");
-        inner.supervisor = Some(SyncTaskSupervisor::new(&inner).await);
-
-        self.state.set(State::Running);
     }
 
     /// Stop the underlying sliding syncs.
@@ -380,28 +381,27 @@ impl SyncService {
     pub async fn stop(&self) -> Result<(), Error> {
         let mut inner = self.inner.lock().await;
 
-        match self.state.get() {
+        match inner.state.get() {
             State::Idle | State::Terminated | State::Error => {
                 // No need to stop if we were not running.
-                return Ok(());
+                Ok(())
             }
-            State::Running => {}
-        };
+            State::Running => {
+                trace!("pausing sync service");
 
-        trace!("pausing sync service");
+                // First, request to stop the two underlying syncs; we'll look at the results
+                // later, so that we're in a clean state independently of the request to stop.
 
-        // First, request to stop the two underlying syncs; we'll look at the results
-        // later, so that we're in a clean state independently of the request to
-        // stop.
+                // Remove the supervisor from our inner state and request the tasks to be
+                // shutdown.
+                let supervisor = inner.supervisor.take().ok_or_else(|| {
+                    error!("The supervisor was not properly started up");
+                    Error::InternalSupervisorError
+                })?;
 
-        // Remove the supervisor from our inner state and request the tasks to be
-        // shutdown.
-        let supervisor = inner.supervisor.take().ok_or_else(|| {
-            error!("The supervisor was not properly started up");
-            Error::InternalSupervisorError
-        })?;
-
-        supervisor.shutdown().await
+                supervisor.shutdown().await
+            }
+        }
     }
 
     /// Attempt to get a permit to use an `EncryptionSyncService` at a given
--- a/crates/matrix-sdk-ui/src/sync_service.rs
+++ b/crates/matrix-sdk-ui/src/sync_service.rs
@@ -23,7 +23,7 @@
 //! MUST observe. Whenever an error/termination is observed, the user MUST call
 //! [`SyncService::start()`] again to restart the room list sync.
 
-use std::sync::{Arc, Mutex};
+use std::sync::Arc;
 
 use eyeball::{SharedObservable, Subscriber};
 use futures_core::Future;
@@ -67,35 +67,11 @@ pub enum State {
     Error,
 }
 
-pub struct SyncService {
-    /// Room list service used to synchronize the rooms state.
+struct SyncServiceInner {
     room_list_service: Arc<RoomListService>,
-
-    /// Encryption sync taking care of e2ee events.
     encryption_sync_service: Arc<EncryptionSyncService>,
-
-    /// What's the state of this sync service?
     state: SharedObservable<State>,
-
-    /// Use a mutex everytime to modify the `state` value, otherwise it would be
-    /// possible to have race conditions when starting or pausing the
-    /// service multiple times really quickly.
-    modifying_state: AsyncMutex<()>,
-
-    /// Task running the room list service.
-    room_list_task: Arc<Mutex<Option<JoinHandle<()>>>>,
-
-    /// Task running the encryption sync.
-    encryption_sync_task: Arc<Mutex<Option<JoinHandle<()>>>>,
-
-    /// Global lock to allow using at most one `EncryptionSyncService` at all
-    /// times.
-    ///
-    /// This ensures that there's only one ever existing in the application's
-    /// lifetime (under the assumption that there is at most one
-    /// `SyncService` per application).
     encryption_sync_permit: Arc<AsyncMutex<EncryptionSyncPermit>>,
-
     /// Scheduler task ensuring proper termination.
     ///
     /// This task is waiting for a `TerminationReport` from any of the other two
@@ -103,12 +79,32 @@ pub struct SyncService {
     /// that the two services are properly shut up and just interrupted.
     ///
     /// This is set at the same time as the other two tasks.
-    scheduler_task: Arc<Mutex<Option<JoinHandle<()>>>>,
-
+    scheduler_task: Option<JoinHandle<()>>,
     /// `TerminationReport` sender for the [`Self::stop()`] function.
     ///
     /// This is set at the same time as all the tasks in [`Self::start()`].
-    scheduler_sender: Mutex<Option<Sender<TerminationReport>>>,
+    scheduler_sender: Option<Sender<TerminationReport>>,
+}
+
+pub struct SyncService {
+    inner: Arc<AsyncMutex<SyncServiceInner>>,
+
+    /// Room list service used to synchronize the rooms state.
+    room_list_service: Arc<RoomListService>,
+
+    /// What's the state of this sync service? This field is replicated from the
+    /// [`SyncServiceInner`] struct, but it should not be modified in this
+    /// struct. It's re-exposed here so we can subscribe to the state
+    /// without taking the lock on the `inner` field.
+    state: SharedObservable<State>,
+
+    /// Global lock to allow using at most one [`EncryptionSyncService`] at all
+    /// times.
+    ///
+    /// This ensures that there's only one ever existing in the application's
+    /// lifetime (under the assumption that there is at most one [`SyncService`]
+    /// per application).
+    encryption_sync_permit: Arc<AsyncMutex<EncryptionSyncPermit>>,
 }
 
 impl SyncService {
@@ -134,13 +130,14 @@ impl SyncService {
     /// the other one too).
     fn spawn_scheduler_task(
         &self,
+        inner: &SyncServiceInner,
+        room_list_task: JoinHandle<()>,
+        encryption_sync_task: JoinHandle<()>,
         mut receiver: Receiver<TerminationReport>,
     ) -> impl Future<Output = ()> {
-        let encryption_sync_task = self.encryption_sync_task.clone();
-        let encryption_sync = self.encryption_sync_service.clone();
-        let room_list_service = self.room_list_service.clone();
-        let room_list_task = self.room_list_task.clone();
-        let state = self.state.clone();
+        let encryption_sync = inner.encryption_sync_service.clone();
+        let room_list_service = inner.room_list_service.clone();
+        let state = inner.state.clone();
 
         async move {
             let Some(report) = receiver.recv().await else {
@@ -165,13 +162,8 @@ impl SyncService {
                 }
             }
 
-            {
-                let task = room_list_task.lock().unwrap().take();
-                if let Some(task) = task {
-                    if let Err(err) = task.await {
-                        error!("when awaiting room list service: {err:#}");
-                    }
-                }
+            if let Err(err) = room_list_task.await {
+                error!("when awaiting room list service: {err:#}");
             }
 
             if stop_encryption {
@@ -180,13 +172,8 @@ impl SyncService {
                 }
             }
 
-            {
-                let task = encryption_sync_task.lock().unwrap().take();
-                if let Some(task) = task {
-                    if let Err(err) = task.await {
-                        error!("when awaiting encryption sync: {err:#}");
-                    }
-                }
+            if let Err(err) = encryption_sync_task.await {
+                error!("when awaiting encryption sync: {err:#}");
             }
 
             if report.is_error {
@@ -306,7 +293,7 @@ impl SyncService {
     /// - if the stream has been aborted before, it will be properly cleaned up
     ///   and restarted.
     pub async fn start(&self) {
-        let _guard = self.modifying_state.lock().await;
+        let mut inner = self.inner.lock().await;
 
         // Only (re)start the tasks if any was stopped.
         if matches!(self.state.get(), State::Running) {
@@ -319,20 +306,25 @@ impl SyncService {
         let (sender, receiver) = tokio::sync::mpsc::channel(16);
 
         // First, take care of the room list.
-        *self.room_list_task.lock().unwrap() =
-            Some(spawn(Self::room_list_sync_task(self.room_list_service.clone(), sender.clone())));
+        let room_list_task =
+            spawn(Self::room_list_sync_task(self.room_list_service.clone(), sender.clone()));
 
         // Then, take care of the encryption sync.
-        let sync_permit_guard = self.encryption_sync_permit.clone().lock_owned().await;
-        *self.encryption_sync_task.lock().unwrap() = Some(spawn(Self::encryption_sync_task(
-            self.encryption_sync_service.clone(),
+        let sync_permit_guard = inner.encryption_sync_permit.clone().lock_owned().await;
+        let encryption_sync_task = spawn(Self::encryption_sync_task(
+            inner.encryption_sync_service.clone(),
             sender.clone(),
             sync_permit_guard,
-        )));
+        ));
 
         // Spawn the scheduler task.
-        *self.scheduler_sender.lock().unwrap() = Some(sender);
-        *self.scheduler_task.lock().unwrap() = Some(spawn(self.spawn_scheduler_task(receiver)));
+        inner.scheduler_sender = Some(sender);
+        inner.scheduler_task = Some(spawn(self.spawn_scheduler_task(
+            &inner,
+            room_list_task,
+            encryption_sync_task,
+            receiver,
+        )));
 
         self.state.set(State::Running);
     }
@@ -344,7 +336,7 @@ impl SyncService {
     /// necessary.
     #[instrument(skip_all)]
     pub async fn stop(&self) -> Result<(), Error> {
-        let _guard = self.modifying_state.lock().await;
+        let mut inner = self.inner.lock().await;
 
         match self.state.get() {
             State::Idle | State::Terminated | State::Error => {
@@ -360,7 +352,7 @@ impl SyncService {
         // later, so that we're in a clean state independently of the request to
         // stop.
 
-        let sender = self.scheduler_sender.lock().unwrap().clone();
+        let sender = inner.scheduler_sender.clone();
         sender
             .ok_or_else(|| {
                 error!("missing sender");
@@ -377,7 +369,7 @@ impl SyncService {
                 Error::InternalSchedulerError
             })?;
 
-        let scheduler_task = self.scheduler_task.lock().unwrap().take();
+        let scheduler_task = inner.scheduler_task.take();
         scheduler_task
             .ok_or_else(|| {
                 error!("missing scheduler task");
@@ -420,11 +412,8 @@ struct TerminationReport {
 #[doc(hidden)]
 impl SyncService {
     /// Return the existential states of internal tasks.
-    pub fn task_states(&self) -> (bool, bool) {
-        (
-            self.encryption_sync_task.lock().unwrap().is_some(),
-            self.room_list_task.lock().unwrap().is_some(),
-        )
+    pub async fn task_state(&self) -> bool {
+        self.inner.lock().await.scheduler_task.is_some()
     }
 }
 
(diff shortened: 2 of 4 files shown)


crates/matrix-sdk-ui/src/sync_service.rs
```diff
@@ -67,48 +67,40 @@ pub enum State {
     Error,
 }
 
-pub struct SyncService {
-    /// Room list service used to synchronize the rooms state.
```
bnjbvr, line 71: Why did you remove all the fields comments? :/

crates/matrix-sdk-ui/src/sync_service.rs
```diff
@@ -67,48 +67,40 @@ pub enum State {
...
+    /// struct. It's re-exposed here so we can subscribe to the state
+    /// without taking the lock on the `inner` field.
+    state: SharedObservable<State>,
+    /// Global lock to allow using at most one `EncryptionSyncService` at all   
+    /// times.
+    ///                                                                     
+    /// This ensures that there's only one ever existing in the
+    /// application's lifetime (under the assumption that there is at most
+    /// one `SyncService` per application).
+    encryption_sync_permit: Arc<AsyncMutex<EncryptionSyncPermit>>,
```
bnjbvr: Can you keep the newlines between the different fields? Keeping this super close to each other makes it really hard to read.

crates/matrix-sdk-ui/src/sync_service.rs
```diff
@@ -368,28 +369,27 @@ impl SyncService {
...
+                // shutdown.
+                let supervisor = inner.supervisor.take().ok_or_else(|| {
+                    error!("The supervisor was not properly started up");
+                    Error::InternalSupervisorError
+                })?;
 
-        supervisor.shutdown().await
+                supervisor.shutdown().await
+            }
+        }
```
bnjbvr: This change feels like a regression: it adds one extra level of indent, for absolutely no extra value. I see the point of doing it in the other function because the body's size is so small there (in particular, there's no control flow), but it's not the case here, so this just adds visual clutter :/

crates/matrix-sdk-ui/src/sync_service.rs
```diff
@@ -219,14 +224,9 @@ impl SyncTaskSupervisor {
...
-        if let Err(err) = sender
-            .send(TerminationReport {
-                is_error,
-                has_expired,
-                origin: TerminationOrigin::EncryptionSync,
-            })
-            .await
-        {
+        let report =
+            TerminationReport { is_error, has_expired, origin: TerminationOrigin::EncryptionSync };
```
bnjbvr, line 230: I don't see the point of the `report` variables: they only add one level of "indirection" (some extra thing to keep in mind for a reader), they create one extra binding for the compiler frontend to handle, and they're used only once, so might as well be kept inlined at use.

bnjbvr (changes requested):
Haven't looked at all of it, but some of the cosmetic changes seem plain wrong, as they break symmetry with most of the current code style in the UI crate.

crates/matrix-sdk-ui/src/sync_service.rs
poljar: Oh, one of them was lost in the move. Or do you mean all those should be replicated in the `Inner` struct?

Anyways, I added the missing one back: 9ae02355a8041558b5ea19156e1536f6d0e35599.

crates/matrix-sdk-ui/src/sync_service.rs
poljar: Sure: 9ae02355a8041558b5ea19156e1536f6d0e35599.

crates/matrix-sdk-ui/src/sync_service.rs
poljar: I prefer the formatting of the two statements compared to the formatting of the single statement, but ok: b38e8b94b55800bfa290b270107096f3a15b95f0.

crates/matrix-sdk-ui/src/sync_service.rs
poljar: The body size of the SyncService::start() and SyncService::stop() methods was reduced to just a couple of lines in a previous commit, allowing us to do exactly this without introducing visual clutter from indentation.

The value becomes clear when you consider what needs to happen if someone adds another enum variant—something I plan to do.

Consider the examples:

First, the pattern where we flatten out one branch to remove the indentation:
```rust
enum Foo {
    A,
    B,
}

fn bar(foo: Foo) {
    match foo {
        // Do nothing if Foo::A.
        A => return,
        B => (),
    }
    // Do what we would have done in the B branch.
    call_something_for_b();
}
```

Now the way it is in the pull-request:
```rust
enum Foo {
    A,
    B,
}

fn bar(foo: Foo) {
    match foo {
        // Do nothing if Foo::A.
        A => (),
        // Do what we need to do for B.
        B => {
             call_something_for_b();
        },
    }
}
```

When we add a new variant, I’ll either need to convert the first variant into the second, or introduce even more early returns, which would make the flow even less clear:

```rust
enum Foo {
    A,
    B,
    C,
}

fn bar(foo: Foo) {
    match foo {
        // Do nothing if Foo::A.
        A => return,
        B => (),
        C => {
            // Do what we need for C, can't be flattened out like we did for B.
            do_something_for_c();
            return;  
        },
    }
    // Do what we would have done in the B branch.
    call_something_for_b();
}
```

I hope we can agree that adding more variants would make the flow of execution even more convoluted.

In contrast, the second variant naturally provides a clear place to add a new branch, that's precisely what pattern matching is here for.

```rust
enum Foo {
    A,
    B,
    C,
}

fn bar(foo: Foo) {
    match foo {
        A => (),
        B => call_something_for_b(),
        C => do_something_for_c(),
    }
}
```

I would urge you to reconsider your stance on this pattern and code style, and to evaluate how flattening the branches of a pattern match might affect the extensibility of the codebase.

Early returns have their, in my opinion, limited use—but this ain't it, babe. :notes:

bnjbvr: That you planned to add another variant is something I wasn't aware of, and couldn't guess, so that makes sense in this case :+1: 

> I would urge you to reconsider your stance on this pattern and code style, and to evaluate how flattening the branches of a pattern match might affect the extensibility of the codebase.

No, I'd rather have you reconsider your stance: more early returns make for fewer indent levels, and more "linear" / "simple" / "beautiful" code, where all the assumptions are checked upfront, and then the *actual* work is done later. Agree to disagree here :-)

andybalaam (approved):
Looks great, thank you for doing this.

crates/matrix-sdk-ui/src/sync_service.rs
poljar: > That you planned to add another variant is something I wasn't aware of, and couldn't guess, so that makes sense in this case

You seem to be missing the point. Every enum might be expanded to include more variants, the point is to write code that is prepared to easily and exhaustively handle this scenario. The commit did mention this:

```
Now that the various match branches in the start and stop method of the
SyncService are minimized we can remove the early returns.

This should allow us to more easily add new branches.
```

> No, I'd rather have you reconsider your stance

Well I did, otherwise I would not have rewritten this. I explained the problems such code cause multiple times in this very thread, so could you do your due diligence please as well?

bnjbvr: We can talk about this privately.

crates/matrix-sdk-ui/src/sync_service.rs
```diff
@@ -67,48 +67,44 @@ pub enum State {
...
     /// This is set at the same time as all the tasks in [`Self::start()`].
-    scheduler_sender: Mutex<Option<Sender<TerminationReport>>>,
+    scheduler_sender: Option<Sender<TerminationReport>>,
+}
+
+pub struct SyncService {
+    inner: Arc<AsyncMutex<SyncServiceInner>>,
+
+    /// Room list service used to synchronize the rooms state.
+    room_list_service: Arc<RoomListService>,
```
bnjbvr: There's likely no point in storing this into `SyncServiceInner` as well, right?

crates/matrix-sdk-ui/src/sync_service.rs
```diff
@@ -67,48 +67,44 @@ pub enum State {
...
+    /// without taking the lock on the `inner` field.
+    state: SharedObservable<State>,
+
+    /// Global lock to allow using at most one [`EncryptionSyncService`] at all
+    /// times.
+    ///
+    /// This ensures that there's only one ever existing in the application's
+    /// lifetime (under the assumption that there is at most one [`SyncService`]
+    /// per application).
+    encryption_sync_permit: Arc<AsyncMutex<EncryptionSyncPermit>>,
```
bnjbvr: Ditto, I don't see the point of having this cloned in both the outer and inner structs.

crates/matrix-sdk-ui/src/sync_service.rs
```diff
@@ -412,8 +412,8 @@ struct TerminationReport {
 #[doc(hidden)]
 impl SyncService {
     /// Return the existential states of internal tasks.
-    pub async fn task_state(&self) -> bool {
-        self.inner.lock().await.scheduler_task.is_some()
+    pub async fn is_supervisor_running(&self) -> bool {
```
bnjbvr: The doc comment needs an update here.

crates/matrix-sdk-ui/src/sync_service.rs
```diff
@@ -67,69 +67,40 @@ pub enum State {
...
-    /// tasks, or from a user request via [`Self::stop()`]. It makes sure
-    /// that the two services are properly shut up and just interrupted.
-    ///
-    /// This is set at the same time as the other two tasks.
-    supervisor_task: Option<JoinHandle<()>>,
+/// A supervisor that starts two sync tasks, one for the room list and one for
+/// the end-to-end encryption support.
+struct SyncTaskSupervisor {
+    /// The task that supervises the two sync tasks.
+    task: JoinHandle<()>,
```
bnjbvr, line 73: The comment is confusing at first, because it mentions two tasks but there's only one set in the struct.

crates/matrix-sdk-ui/src/sync_service.rs
```diff
@@ -67,69 +67,40 @@ pub enum State {
...
     /// `TerminationReport` sender for the [`Self::stop()`] function.
     ///
     /// This is set at the same time as all the tasks in [`Self::start()`].
```
bnjbvr: Self::start and stop don't seem to exist anymore, do they?

crates/matrix-sdk-ui/src/sync_service.rs
```diff
@@ -67,69 +67,40 @@ pub enum State {
...
-    state: SharedObservable<State>,
-
-    /// Global lock to allow using at most one [`EncryptionSyncService`] at all
-    /// times.
-    ///
-    /// This ensures that there's only one ever existing in the application's
-    /// lifetime (under the assumption that there is at most one [`SyncService`]
-    /// per application).
-    encryption_sync_permit: Arc<AsyncMutex<EncryptionSyncPermit>>,
+    abortion_sender: Sender<TerminationReport>,
```
bnjbvr, line 76: Can this be called `termination_sender`, with respect to `TerminatinoReport`? (that could be a `TerminationRequest` too)

crates/matrix-sdk-ui/src/sync_service.rs
```diff
@@ -196,6 +167,75 @@ impl SyncService {
...
+        }
+    }
+}
+
+struct SyncServiceInner {
+    room_list_service: Arc<RoomListService>,
+    encryption_sync_service: Arc<EncryptionSyncService>,
+    state: SharedObservable<State>,
+    encryption_sync_permit: Arc<AsyncMutex<EncryptionSyncPermit>>,
+    supervisor: Option<SyncTaskSupervisor>,
```
bnjbvr: Did we lose again all the doc comments here? :(

crates/matrix-sdk-ui/src/sync_service.rs
```diff
@@ -139,22 +143,17 @@ impl SyncTaskSupervisor {
...
             if report.is_error {
-                if report.has_expired {
-                    if stop_room_list {
-                        room_list_service.expire_sync_session().await;
-                    }
-                    if stop_encryption {
-                        encryption_sync.expire_sync_session().await;
-                    }
-                }
-
```
bnjbvr: :warning: This commit is changing the semantics: before we'd only expire if we ran into an error, now it'll expire every time it's stopped. This should behave as it did beforehand, otherwise any small network cut will cause spurious sync expirations.

crates/matrix-sdk-ui/src/sync_service.rs
```diff
@@ -300,6 +298,8 @@ impl SyncTaskSupervisor {
             }),
             Err(err) => {
                 error!("when sending termination report: {err}");
+                // Let's abort the task if it won't shut down properly, otherwise we would have
+                // left it as a detached tasks.
```
bnjbvr, line 302: nit: task

crates/matrix-sdk-ui/Cargo.toml
```diff
@@ -61,6 +61,7 @@ matrix-sdk = { workspace = true, features = ["testing", "sqlite"] }
 matrix-sdk-test = { workspace = true }
 stream_assert = { workspace = true }
 tempfile = { workspace = true }
+url = { workspace = true }
```
bnjbvr, line 64: uh, can we make this dependency optional, only active if compiling docs? It seems a bit wasteful to require it overall just for a doc example.

crates/matrix-sdk-ui/src/sync_service.rs
```diff
@@ -321,6 +321,11 @@ struct SyncServiceInner {
 /// with a Matrix server. It can initiate and maintain the necessary
 /// synchronization tasks for you.
 ///
+/// **Note**: The [`SyncService`] requires as server with support for [MSC4186],
```
bnjbvr, line 324: nit: a server, not as server

bnjbvr (changes requested):
I've looked at all the commits now. The commit message for the `Scheduler` -> `Supervisor` renaming was not really convincing, but the offline discussion where you referred to the computing-specific definition of `supervisor` had me agree it's a good name.

My two blocking comments are the ones starting with a :warning: emoji, otherwise mostly nits here and there. I'd like to take another look before approving. Thanks!

crates/matrix-sdk-ui/src/sync_service.rs
poljar: When you spawn the child tasks you need to pass the `RoomListService` to them, it's more convenient to just pass `SyncServiceInner` to the spawning function than to pass this in separately.

bnjbvr: Can we pass the parameter separately, instead? Preferring developer convenience for one or two call sites, over cloning a atomic ref-counted pointer feels like the wrong tradeoff.

crates/matrix-sdk-ui/src/sync_service.rs
poljar: As far as I understand, and please correct me if I'm wrong, `expired` can only be `true` if `error` is also true:

https://github.com/matrix-org/matrix-rust-sdk/blob/e151ea9664b7177cc73ffa6b3df0353750a5f185/crates/matrix-sdk-ui/src/sync_service.rs#L220

But that makes me think that these two booleans which aren't really independent should be replaced with an enum:

```rust
enum TerminationReason {
     Error,
     ExpiredAndError,
     Shrug,
}
```

This way we encode in the type that the fourth state (no error but expired) can't happen.

crates/matrix-sdk-ui/src/sync_service.rs
poljar: Sure.

crates/matrix-sdk-ui/Cargo.toml
poljar: I mean sure, but not really easily as you would need to enable it for `cfg(doc)` but not otherwise. Or you'd need to create a doc feature.

Doesn't seem worth it since that dependency will be compiled as part of the workspace or more directly as part of the `matrix-sdk` crate which the `matrix-sdk-ui` crate directly depends on.

crates/matrix-sdk-ui/src/sync_service.rs
poljar: They do not, this was addressed in a later commit and we point to the `SyncTaskSupervisor::shutdown()` method now.

crates/matrix-sdk-ui/src/sync_service.rs
bnjbvr: Good point! Apart from the naming of `Shrug`, that sounds like a great thing to do (here or in a followup, as is preferred).

crates/matrix-sdk-ui/src/sync_service.rs
poljar: d05784beb09098de4c5fd65d15c0f3dc91c89fe6

crates/matrix-sdk-ui/src/sync_service.rs
poljar: d05784beb09098de4c5fd65d15c0f3dc91c89fe6

crates/matrix-sdk-ui/src/sync_service.rs
poljar: I attempted to reword this in 7707997dc3a92c2d1d3c390da90837bc833797cb.

crates/matrix-sdk-ui/src/sync_service.rs
poljar: Only one of those fields had docs, but indeed it got lost in a rebase for some reason: 20628615cdaeb6fd4764c2d81674b21293496919.

crates/matrix-sdk-ui/src/sync_service.rs
poljar: I don't think I want to make this part of this PR.

crates/matrix-sdk-ui/src/sync_service.rs
poljar: ea3c044bde58f397de44a1eb600a25008ef7049d

crates/matrix-sdk-ui/src/sync_service.rs
poljar: 07f86de2919aaf6158aff4b2f9d32e73f3d76cea

crates/matrix-sdk-ui/Cargo.toml
poljar: Since this doesn't have a blocking emoji sign, I'll resolve this thread without doing anything.

crates/matrix-sdk-ui/src/sync_service.rs
poljar: 5bd85fc9dac8f749630fff878349589c0c9a79cd

crates/matrix-sdk-ui/src/sync_service.rs
poljar: That's still not the point I was making.

https://github.com/matrix-org/matrix-rust-sdk/pull/4543/commits/f522a815b25393fdfb9938e967bc8f2451cfb541

crates/matrix-sdk-ui/src/sync_service.rs
```diff
@@ -476,16 +465,22 @@ impl SyncServiceBuilder {
...
+            encryption_sync_permit: encryption_sync_permit.clone(),
+
+            inner: Arc::new(AsyncMutex::new(SyncServiceInner {
+                scheduler_task: None,
+                scheduler_sender: None,
+                room_list_service,
+                encryption_sync_service: encryption_sync,
+                state,
+                encryption_sync_permit,
+            })),
```
Hywan: Replacing 5 locks by 1 lock, that's what I call an improvement. Well done!

crates/matrix-sdk-ui/src/sync_service.rs
```diff
@@ -72,18 +72,18 @@ struct SyncServiceInner {
     encryption_sync_service: Arc<EncryptionSyncService>,
     state: SharedObservable<State>,
     encryption_sync_permit: Arc<AsyncMutex<EncryptionSyncPermit>>,
-    /// Scheduler task ensuring proper termination.
+    /// Supervisor task ensuring proper termination.
```
Hywan: I agree it's a better name according to your definitions. _Scheduler_ wasn't “wrong” neither, but I believe _Supervisor_ is more adapted.

crates/matrix-sdk-ui/src/sync_service.rs
```diff
@@ -67,69 +67,40 @@ pub enum State {
...
-    supervisor_task: Option<JoinHandle<()>>,
+/// A supervisor that starts two sync tasks, one for the room list and one for
+/// the end-to-end encryption support.
+struct SyncTaskSupervisor {
+    /// The task that supervises the two sync tasks.
+    task: JoinHandle<()>,
     /// `TerminationReport` sender for the [`Self::stop()`] function.
     ///
     /// This is set at the same time as all the tasks in [`Self::start()`].
-    supervisor_sender: Option<Sender<TerminationReport>>,
```
Hywan: Replacing 2 `Option`s by 1 reduces the states in this code, well done. Since it is for the struct itself, and not for its fields, the complexity is also reduced.

crates/matrix-sdk-ui/src/sync_service.rs
```diff
@@ -79,13 +79,11 @@ struct SyncTaskSupervisor {
...
 impl SyncTaskSupervisor {
-    async fn new(
-        inner: &SyncServiceInner,
-        room_list_task: JoinHandle<()>,
-        encryption_sync_task: JoinHandle<()>,
-        sender: Sender<TerminationReport>,
-        receiver: Receiver<TerminationReport>,
-    ) -> Self {
+    async fn new(inner: &SyncServiceInner) -> Self {
+        let (sender, receiver) = tokio::sync::mpsc::channel(16);
```
Hywan, line 91: Are we sure 16 is enough? Don't we want an unbounded channel, or a channel with a larger buffer?

crates/matrix-sdk-ui/src/sync_service.rs
Hywan: Your idea @poljar is nice and that's something I would enjoy reading.

crates/matrix-sdk-ui/src/sync_service.rs
```diff
@@ -183,33 +183,38 @@ impl SyncTaskSupervisor {
         (room_list_task, encryption_sync_task)
     }
 
+    fn check_if_expired(err: &matrix_sdk::Error) -> bool {
```
Hywan, line 205: That's an improvement. It can even be unit tested 😇.

crates/matrix-sdk-ui/src/sync_service.rs
```diff
@@ -315,6 +315,47 @@ struct SyncServiceInner {
     supervisor: Option<SyncTaskSupervisor>,
 }
 
+/// A high level manager for your Matrix syncing needs.
+///
+/// The [`SyncService`] is responsible for managing real-time synchronization
+/// with a Matrix server. It can initiate and maintain the necessary
+/// synchronization tasks for you.
```
Hywan, line 345: Thanks! That's exactly what we often miss to do in the SDK, this kind of doc is really useful.

crates/matrix-sdk-ui/src/sync_service.rs
```diff
@@ -549,15 +565,12 @@ impl SyncServiceBuilder {
 
         Ok(SyncService {
             state: state.clone(),
-            room_list_service: room_list_service.clone(),
-            encryption_sync_permit: encryption_sync_permit.clone(),
-
+            room_list_service,
+            encryption_sync_permit,
```
Hywan: Getting better and better!

crates/matrix-sdk-ui/src/sync_service.rs
```diff
@@ -73,7 +73,7 @@ struct SyncTaskSupervisor {
     task: JoinHandle<()>,
     /// [`TerminationReport`] sender for the [`SyncTaskSupervisor::shutdown()`]
     /// function.
-    abortion_sender: Sender<TerminationReport>,
+    termination_sender: Sender<TerminationReport>,
```
Hywan, line 82: Agree on this.

crates/matrix-sdk-ui/src/sync_service.rs
```diff
@@ -66,10 +66,16 @@ pub enum State {
     Error,
 }
 
-/// A supervisor that starts two sync tasks, one for the room list and one for
-/// the end-to-end encryption support.
+/// A supervisor responsible for managing two sync tasks: one for handling the
+/// room list and another for supporting end-to-end encryption.
+///
+/// The two sync tasks are spawned as child tasks and are contained within t
```
Hywan, line 72:
```diff
- within t
+ within the
```

Hywan (approved):
I see many improvements in this pull request, thank you for taking time working on this!

I've left a couple of comments that you may or may not to address; up to you.

crates/matrix-sdk-ui/src/sync_service.rs
poljar: Thanks for spotting this: https://github.com/matrix-org/matrix-rust-sdk/pull/4543/commits/f4c94abe471d89432bf61df12a9fe92ea2309682.

crates/matrix-sdk-ui/src/sync_service.rs
poljar: I think so, yeah. Basically you get only a message if one of the syncs stops or if the user stops, so it's really low volume.

In any case, this hasn't been modified from what we had previously and we didn't see any problems with this.

Issue #4313, opened 2024-11-22
KeySharing, DevicebasedStrategy | Exclude insecure dehydrated devices when sending messages


uhoreg:
Don't encrypt to devices marked as dehydrated, if they are not cross-signed by the ~~pinned/verified~~ identity.  (Per discussion below: we continue to send to the dehydrated device even if the identity changes) 

~~Also, drop all incoming to-device messages from devices marked as dehydrated.~~ Factored out to #4466

andybalaam:
(Rust) We already have code to avoid encrypting for unsigned devices - we should do this unconditionally for any device with the dehydrated flag set.

richvdh:
@uhoreg I don't see any mention of this behaviour in MSC3814 -- presumably it is something that should be called out there?

richvdh:
Also: why are dehydrated devices being singled out for this special treatment?

uhoreg:
> @uhoreg I don't see any mention of this behaviour in [MSC3814](https://github.com/matrix-org/matrix-spec-proposals/pull/3814) -- presumably it is something that should be called out there?

It's buried in the middle of a paragraph: https://github.com/matrix-org/matrix-spec-proposals/pull/3814/files#diff-1e380a19a30044a5d4a387df67369612e4cdf4dc8b0bcba1853fd221f7dfeeb9R51-R53 I guess it should be called out more.

> Also: why are dehydrated devices being singled out for this special treatment?

This was flagged by Denis at https://github.com/matrix-org/matrix-spec-proposals/pull/3814#discussion_r1822841722 .  The short version is: dehydrated devices are being singled out because clients may hide dehydrated devices or make them less visible, so it may be less noticeable if a dehydrated device is unsigned, compared with normal devices.  Also because dehydrated devices are a new feature and our eventual goal is to drop insecure devices.

richvdh:
ah great, thanks.

richvdh:
> (Rust) We already have code to avoid encrypting for unsigned devices - we should do this unconditionally for any device with the dehydrated flag set.

kinda, though that's implemented as a different `CollectStrategy`, which means we can't directly make use of the same logic. Also: the `CollectStrategy` does not check whether the identity that the device is signed by is the pinned one, so if we want that behaviour, we need to do a bit more work.

richvdh:
> Don't encrypt to devices marked as dehydrated, if they are not cross-signed by the pinned/verified identity.

[...]

PR #4551, opened 2025-01-17
crypto: withhold outgoing messages to unsigned dehydrated devices

richvdh:
Per https://github.com/matrix-org/matrix-rust-sdk/issues/4313, we should not send outgoing messages to dehydrated devices that are not signed by the user's identity

Fixes https://github.com/matrix-org/matrix-rust-sdk/issues/4313

### DIFF
--- a/crates/matrix-sdk-crypto/src/identities/device.rs
+++ b/crates/matrix-sdk-crypto/src/identities/device.rs
@@ -479,9 +479,9 @@ impl Device {
         Ok(raw_encrypted)
     }
 
-    /// Whether or not the device is a dehydrated device.
+    /// True if this device is an [MSC3814](https://github.com/matrix-org/matrix-spec-proposals/pull/3814) dehydrated device.
     pub fn is_dehydrated(&self) -> bool {
-        self.inner.device_keys.dehydrated.unwrap_or(false)
+        self.inner.is_dehydrated()
     }
 }
 
@@ -965,6 +965,11 @@ impl DeviceData {
     /// milliseconds since epoch (client local time).
     pub fn first_time_seen_ts(&self) -> MilliSecondsSinceUnixEpoch {
         self.first_time_seen_ts
+    }
+
+    /// True if this device is an [MSC3814](https://github.com/matrix-org/matrix-spec-proposals/pull/3814) dehydrated device.
+    pub fn is_dehydrated(&self) -> bool {
+        self.device_keys.dehydrated.unwrap_or(false)
     }
 }
 
--- a/crates/matrix-sdk-crypto/src/session_manager/group_sessions/share_strategy.rs
+++ b/crates/matrix-sdk-crypto/src/session_manager/group_sessions/share_strategy.rs
@@ -656,30 +656,33 @@ mod tests {
         CrossSigningKeyExport, EncryptionSettings, LocalTrust, OlmError, OlmMachine,
     };
 
-    async fn set_up_test_machine() -> OlmMachine {
-        let machine = OlmMachine::new(
-            KeyDistributionTestData::me_id(),
-            KeyDistributionTestData::me_device_id(),
-        )
-        .await;
+    /// Returns an `OlmMachine` set up for the test user in
+    /// [`KeyDistributionTestData`], with cross-signing set up and the
+    /// private cross-signing keys imported.
+    async fn test_machine() -> OlmMachine {
+        use KeyDistributionTestData as DataSet;
 
-        let keys_query = KeyDistributionTestData::me_keys_query_response();
-        let txn_id = TransactionId::new();
-        machine.mark_request_as_sent(&txn_id, &keys_query).await.unwrap();
+        // Create the local user (`@me`), and import the public identity keys
+        let machine = OlmMachine::new(DataSet::me_id(), DataSet::me_device_id()).await;
+        let keys_query = DataSet::me_keys_query_response();
+        machine.mark_request_as_sent(&TransactionId::new(), &keys_query).await.unwrap();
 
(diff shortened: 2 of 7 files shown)


crates/matrix-sdk-crypto/src/session_manager/group_sessions/share_strategy.rs
```diff
@@ -160,15 +161,7 @@ pub(crate) async fn collect_session_recipients(
...
-
-                // We only need the user identity if `only_allow_trusted_devices` or
-                // `error_on_verified_user_problem` is set.
-                let device_owner_identity =
-                    if only_allow_trusted_devices || error_on_verified_user_problem {
-                        store.get_user_identity(user_id).await?
-                    } else {
-                        None
-                    };
+                let device_owner_identity = store.get_user_identity(user_id).await?;
```
richvdh, line 164: We now also need the identity if the user has a dehydrated device. Trying to optimise away this call in the cases where we weren't going to need the identity was getting too complicated, so let's just do it every time.

richvdh:
I need to update this in view of https://github.com/matrix-org/matrix-rust-sdk/issues/4313#issuecomment-2602570503

richvdh:
> I need to update this in view of https://github.com/matrix-org/matrix-rust-sdk/issues/4313#issuecomment-2602570503

(Now done)

crates/matrix-sdk-crypto/src/session_manager/group_sessions/share_strategy.rs
```diff
@@ -406,23 +398,50 @@ fn split_devices_for_user(
     for d in user_devices.into_values() {
         if d.is_blacklisted() {
             recipient_devices.denied_devices_with_code.push((d, WithheldCode::Blacklisted));
-        } else if d.local_trust_state() == LocalTrust::Ignored {
+            continue;
```
poljar, line 401: The logic in the loop has, due to the additions of various new settings, grown a bit into a maze, it's quite hard to enumerate that all the different settings and combinations thereof don't have some logic flaw in them.

I think this needs to be put into a separate method which would then allow us to easier test the logic as well.

Something like:

```rust
enum Decision {
    Ok,
    Withhold(WithheldCode),
    Err,
}

fn decide_what_to_do_with_device(device: &Device) -> Decision {
    ...
}
```

Then write a match statement per setting, i.e.

```rust
if only_allow_trusted {
    match (d.is_verified(), d.is_dehydrated)
       ...
    }
}
```

Or something along those lines which would allow us to easily see if our logic is exhaustive and we didn't forget a case.

poljar (changes requested):
Can we refactor the meat of the logic a bit, I find the function to be quite a challenging read.

crates/matrix-sdk-crypto/src/session_manager/group_sessions/share_strategy.rs
richvdh: I agree that pulling out a separate method makes sense; I'll do that.

I had a go at rewriting the logic for the case where neither `only_allow_trusted` nor `error_on_verified_user_problem` is set. It looks like this:

```rust
        match (
            device.local_trust_state(),
            owner_has_verification_violation,
            device_signed_by_owner,
            device.is_dehydrated(),
        ) {
            (LocalTrust::BlackListed, _, _, _) => {
                DeviceBasedStrategySettingsDeviceDecision::Withhold(WithheldCode::Blacklisted)
            }
            (LocalTrust::Ignored | LocalTrust::Verified, _, _, _) => {
                // Ignore the trust state of that device and share
                DeviceBasedStrategySettingsDeviceDecision::Ok
            }

            (LocalTrust::Unset, _, false, true) => {
                // Dehydrated device not signed by owner
                DeviceBasedStrategySettingsDeviceDecision::Withhold(WithheldCode::Unverified)
            }

            (LocalTrust::Unset, true, _, true) => {
                // Dehydrated device from user with verification violation
                DeviceBasedStrategySettingsDeviceDecision::Withhold(WithheldCode::Unverified)
            }

            (LocalTrust::Unset, _, _, _) => DeviceBasedStrategySettingsDeviceDecision::Ok,
        }
```

I'm not convinced that's a material improvement in clarity over a long series of `if` statements. WDYT?

poljar: Yeah, probably not in that case, since we ignore a lot of branches using `_`. Feel free to pick the if/else way if that looks any prettier.

richvdh:
I've now rebased this PR on top of https://github.com/matrix-org/matrix-rust-sdk/pull/4581, which I think/hope will make the logic introduced here much clearer.

Unfortunately there is a failing test, due to incorrect data initialisation in `test_legacy_dehydrated_device_rehydration`. Hopefully @uhoreg is working on a fix for that....

richvdh:
Now rebased on https://github.com/matrix-org/matrix-rust-sdk/pull/4586, which fixes the failing test.

poljar (approved):
Thanks for taking the time to refactor this so this becomes more manageable.

I took the time to re-read all of our strategies and I think the code matches to what we described in the docs.

Great work.

crates/matrix-sdk-crypto/src/identities/device.rs
```diff
@@ -966,6 +966,11 @@ impl DeviceData {
     pub fn first_time_seen_ts(&self) -> MilliSecondsSinceUnixEpoch {
         self.first_time_seen_ts
     }
+
+    /// True if this device is an MSC3814 dehydrated device.
```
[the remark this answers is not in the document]
richvdh: yup, done.

PR #4552, opened 2025-01-18
fix(ui): Fix `latest_edit_json` for live edits

zecakeh:
This was a regression introduced in f0d98602a9e65cf26a607cf10a0715b92da12d4a.

`latest_edit_json` was first set by the call to `EventTimelineItem::with_content()`. It was overwritten in the next section because of the `if let EventTimelineItemKind::Remote(remote_event)` that uses `item` instead of `new_item`.
It means that the updated `RemoteEventTimelineItem` inside`new_item` was replaced by the outdated one inside `item`, so `latest_edit_json` went back to its previous value.

I believe that part of why that went unnoticed is that the code looks more complicated due to the need to set an inner field dependant on an enum variant, so I decided to change the API and move `with_encryption_info` to `EventTimelineItem`, which makes the code look cleaner.

The changes in the tests fail without this commit.


Hywan (approved):
The new API feels less error-prone/more robust. Thanks for tracking this bug and for the test!

PR #4553, opened 2025-01-19
fix(oidc): remove unnecessary e2e-encryption imports

kilimnik:
I noticed that the crate won't build without the  `e2e-encryption` feature enabled. This does not make it build rather it is a step towards it. 
There is still https://github.com/neodyme-labs/matrix-rust-sdk/blob/oidc_e2e/crates/matrix-sdk/src/oidc/mod.rs#L319 missing.

Signed-off-by: Daniel Kilimnik daniel@neoydme.io

### DIFF
--- a/crates/matrix-sdk/src/oidc/cross_process.rs
+++ b/crates/matrix-sdk/src/oidc/cross_process.rs
@@ -1,5 +1,6 @@
 use std::sync::Arc;
 
+#[cfg(feature = "e2e-encryption")]
 use matrix_sdk_base::crypto::{
     store::{LockableCryptoStore, Store},
     CryptoStoreError,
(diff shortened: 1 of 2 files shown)


kilimnik:
I have no Idea why the test failed, it works locally on my machine :)

kilimnik:
It seems like the whole `cross_process` is dependent on the e2e encryption, so it's currently not possible to use oidc without e2e

bnjbvr:
Yeah, OIDC requires the cross-process lock because it requires one, in case a refresh happened in another process (can happen on iOS because the process receiving notifications (NSE) runs concurrently to the app's process). The "plan" to fix this properly would be to have a single cross-process lock throughout the entire app, which has been somewhat sketched in [this comment](https://github.com/matrix-org/matrix-rust-sdk/pull/4285#issuecomment-2488655073). In the meanwhile, I'm not sure there's much we can do; maybe document the requirement would be a nice win, to not confuse other future readers?

PR #4555, opened 2025-01-20
fix(timeline): avoid adding non-pinned events when back paginating

jmartinesp:
When the event cache receives a `RoomEventCacheUpdate::UpdateTimelineEvents` this will be broadcasted to every timeline, we needed a special case in the pinned event timeline to filter out any items loaded in these updates that aren't pinned.

Signed-off-by:

### DIFF
--- a/crates/matrix-sdk-ui/src/timeline/controller/state.rs
+++ b/crates/matrix-sdk-ui/src/timeline/controller/state.rs
@@ -600,31 +600,37 @@ impl TimelineStateTransaction<'_> {
                             .map_or(RemoteEventOrigin::Unknown, |item| item.origin),
                     };
 
-                    match origin {
-                        RemoteEventOrigin::Sync | RemoteEventOrigin::Unknown => {
-                            should_add = match self.timeline_focus {
-                                TimelineFocusKind::PinnedEvents => {
-                                    // Only insert timeline items for pinned events, if the event
-                                    // came from the sync.
-                                    room_data_provider.is_pinned_event(&event_id)
-                                }
-
-                                TimelineFocusKind::Live => {
+                    // If the event should be added according to the general event filter, use a
+                    // second filter to decide whether it should be added depending on the timeline
+                    // focus and events origin, if needed
+                    match self.timeline_focus {
+                        TimelineFocusKind::PinnedEvents => {
+                            // Only add pinned events for the pinned events timeline
+                            should_add = room_data_provider.is_pinned_event(&event_id);
+                        }
+                        TimelineFocusKind::Live => {
+                            match origin {
+                                RemoteEventOrigin::Sync | RemoteEventOrigin::Unknown => {
                                     // Always add new items to a live timeline receiving items from
                                     // sync.
-                                    true
+                                    should_add = true;
                                 }
-
-                                TimelineFocusKind::Event => {
+                                RemoteEventOrigin::Cache | RemoteEventOrigin::Pagination => {
+                                    // Forward the previous decision to add it.
+                                }
+                            }
+                        }
+                        TimelineFocusKind::Event => {
+                            match origin {
+                                RemoteEventOrigin::Sync | RemoteEventOrigin::Unknown => {
                                     // Never add any item to a focused timeline when the item comes
                                     // down from the sync.
-                                    false
+                                    should_add = false;
                                 }
-                            };
-                        }
-
-                        RemoteEventOrigin::Pagination | RemoteEventOrigin::Cache => {
-                            // Forward the previous decision to add it.
+                                RemoteEventOrigin::Cache | RemoteEventOrigin::Pagination => {
+                                    // Forward the previous decision to add it.
+                                }
+                            }
                         }
                     }
                 }
--- a/crates/matrix-sdk-ui/tests/integration/timeline/pinned_event.rs
+++ b/crates/matrix-sdk-ui/tests/integration/timeline/pinned_event.rs
@@ -1,10 +1,14 @@
-use std::time::Duration;
+use std::{ops::ControlFlow, time::Duration};
 
 use assert_matches::assert_matches;
 use eyeball_im::VectorDiff;
 use matrix_sdk::{
-    assert_next_matches_with_timeout, config::SyncSettings, sync::SyncResponse,
-    test_utils::logged_in_client_with_server, Client,
+    assert_next_matches_with_timeout,
+    config::SyncSettings,
+    event_cache::{BackPaginationOutcome, TimelineHasBeenResetWhilePaginating},
+    sync::SyncResponse,
+    test_utils::logged_in_client_with_server,
+    Client,
 };
 use matrix_sdk_base::deserialized_responses::TimelineEvent;
 use matrix_sdk_test::{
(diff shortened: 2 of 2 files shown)


crates/matrix-sdk-ui/tests/integration/main.rs
```diff
@@ -82,22 +82,6 @@ async fn mock_context(
...
-    event: TimelineEvent,
-) {
-    Mock::given(method("GET"))
-        .and(path(format!("/_matrix/client/r0/rooms/{room_id}/event/{event_id}")))
-        .and(header("authorization", "Bearer 1234"))
-        .respond_with(ResponseTemplate::new(200).set_body_json(event.into_raw().json()))
-        .mount(server)
-        .await;
-}
-
```
jmartinesp, line 100: Removed this, as it's no longer used anywhere.

Hywan (approved):
I wonder if the “regression” would be caused by https://github.com/matrix-org/matrix-rust-sdk/pull/4479 that reveals the real bug?

I would have liked better 2 commits, one for the real fix + its test, and another one for the other test updates, but we can live with that.

Thanks for the fix, it looks good to me! Waiting on @bnjbvr for another pair of eyes, but I believe we're all good! Well done.

crates/matrix-sdk-ui/src/timeline/controller/state.rs
```diff
@@ -624,7 +624,15 @@ impl TimelineStateTransaction<'_> {
...
                         RemoteEventOrigin::Pagination | RemoteEventOrigin::Cache => {
-                            // Forward the previous decision to add it.
+                            match self.timeline_focus {
+                                TimelineFocusKind::PinnedEvents => {
+                                    // Only insert timeline items for pinned events, if the event
+                                    // came from the sync.
+                                    should_add = room_data_provider.is_pinned_event(&event_id);
+                                }
+                                // Forward the previous decision to add it.
+                                TimelineFocusKind::Live | TimelineFocusKind::Event => (),
```
bnjbvr, line 634:
```suggestion
                                TimelineFocusKind::Live | TimelineFocusKind::Event => {},
```

crates/matrix-sdk-ui/src/timeline/controller/state.rs
```diff
@@ -624,7 +624,15 @@ impl TimelineStateTransaction<'_> {
                         }
 
                         RemoteEventOrigin::Pagination | RemoteEventOrigin::Cache => {
-                            // Forward the previous decision to add it.
+                            match self.timeline_focus {
+                                TimelineFocusKind::PinnedEvents => {
+                                    // Only insert timeline items for pinned events, if the event
+                                    // came from the sync.
```
bnjbvr, line 630: This comment is a copypasto, the event isn't coming from the sync.

crates/matrix-sdk-ui/src/timeline/controller/state.rs
```diff
@@ -624,7 +624,15 @@ impl TimelineStateTransaction<'_> {
...
+                                    should_add = room_data_provider.is_pinned_event(&event_id);
```
bnjbvr, line 631: This is a `match self.timeline_focus` in a `match origin`. Looking at the other branch of the `match origin`, it might be simpler to invert the two matches, and have the `match self.timeline_focus` be the outermost, while the `match origin` is the innermost? It would at least avoid the repetition for the pinned events focus.

crates/matrix-sdk-ui/tests/integration/main.rs
bnjbvr: Nice!

crates/matrix-sdk-ui/tests/integration/timeline/pinned_event.rs
```diff
@@ -391,12 +376,86 @@ async fn test_pinned_timeline_with_no_pinned_events_and_an_utd_is_just_empty() {
...
+        .sender(&sender_id)
+        .text_msg("Hey")
+        .event_id(event_id)
+        .into_raw_timeline();
+
+    test_helper
+        .server
+        .mock_room_messages()
+        .ok(assign!(RoomMessagesResponseTemplate::default(), {
+            chunk: vec![not_pinned_event.clone()]
```
bnjbvr, line 426: curious that clippy ain't found it…

```suggestion
            chunk: vec![not_pinned_event]
```

crates/matrix-sdk-ui/tests/integration/timeline/pinned_event.rs
```diff
@@ -391,12 +376,86 @@ async fn test_pinned_timeline_with_no_pinned_events_and_an_utd_is_just_empty() {
...
+
+    // Create a live timeline and paginate backwards. This will cause an event cache
+    // update that will affect the previous pinned events timeline too
+    let live_timeline = Timeline::builder(&room)
+        .with_focus(TimelineFocus::Live)
+        .build()
+        .await
+        .expect("Live timeline couldn't be built");
+    let paginated =
+        live_timeline.paginate_backwards(1).await.expect("Live back pagination should work");
```
bnjbvr, line 440: Alternatively, you can get the room event cache and trigger back-pagination there with `room_event_cache.pagination().run_backwards().await.unwrap()`

bnjbvr (changes requested):
We've asked multiple times that commits be a bit cleaner, so as to distinguish changes that are unrelated to each other. In this commit there's a large refactoring of the test helper, going along one new test and the bugfix for the actual issue fixed here. It's unfair to put the complexity load onto the reviewers' shoulders, so today I'll ask for clean commits.

crates/matrix-sdk-ui/tests/integration/timeline/pinned_event.rs
jmartinesp: It's probably because it was inside the macro. It tends to fail on those cases.

jmartinesp:
I just force pushed https://github.com/matrix-org/matrix-rust-sdk/pull/4555/commits/fe5cdc227511993e0cdc2f14802fa73dce9bb9e9 with the changes asked above. I hope just rewriting the test using the legacy mock APIs will be simple enough.

I may upload the upgrade to the new APIs in a new PR if I have the time, since I don't want to add anything else here that would need to be reviewed.

crates/matrix-sdk-ui/src/timeline/controller/state.rs
```diff
@@ -600,31 +600,33 @@ impl TimelineStateTransaction<'_> {
...
+                    match self.timeline_focus {
+                        TimelineFocusKind::PinnedEvents => {
+                            // Only add pinned events for the pinned events timeline
+                            should_add = room_data_provider.is_pinned_event(&event_id);
                         }
-
-                        RemoteEventOrigin::Pagination | RemoteEventOrigin::Cache => {
-                            // Forward the previous decision to add it.
+                        TimelineFocusKind::Live => {
+                            if matches!(
```
bnjbvr, line 612: Here and below, we could use an explicit `match` statement; this way, if we added `RemoteEventOrigin` variants in the future, it would have us pause and think what we need to do here.

bnjbvr (changes requested):

bnjbvr (approved):

PR #4558, opened 2025-01-20
refactor(crypto): Make memory store behave more like other stores

BillCarsonFr:
Attempt to make the memory store behave more like other stores. 
This is following reports of [Playwright flaky tests](https://github.com/element-hq/crypto-internal/issues/414) that are showing some panics of the memory store (used by the bots not the tested app).

The idea is to make the memory store works more like other stores by removing `unwrap()`s, serializing/deserialising structures instead of keep them around to be mutated by everyone. Add a global save_changes lock

### PRO
Will behave more like other stores (might help and make the tests less flaky)
### CONS
Might be slower

### Breakdown
Best reviewed commit per commit (will squash at the end):
- Replace unwrap with expect https://github.com/matrix-org/matrix-rust-sdk/commit/17fa2c791dbdea0212f37117c4aa477ae8dac2a7
- Serialize account and cache static account data https://github.com/matrix-org/matrix-rust-sdk/commit/b7ac36f4be7d02c2267bc5244d627590e0516b86
- Serialize/Deserialise olm sessions https://github.com/matrix-org/matrix-rust-sdk/commit/7c87d1349a4afc196b0d721c1e5cd9bd1355adda
- Ser/Deser inbound group sessions https://github.com/matrix-org/matrix-rust-sdk/commit/9262c303859969bdd0a10719786a22bbb536a684
- Add a global save change lock similar to other stores https://github.com/matrix-org/matrix-rust-sdk/commit/70c2bafc762140f710b2a6e32362205fe8014371

Signed-off-by:

### DIFF
--- a/crates/matrix-sdk-crypto/src/gossiping/machine.rs
+++ b/crates/matrix-sdk-crypto/src/gossiping/machine.rs
@@ -1115,6 +1115,7 @@ mod tests {
     use crate::{
         gossiping::KeyForwardDecision,
         olm::OutboundGroupSession,
+        store::{CryptoStore, DeviceChanges},
         types::requests::AnyOutgoingRequest,
         types::{
             events::{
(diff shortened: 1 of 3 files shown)


andybalaam:
Why squash at the end?

andybalaam (approved):
This looks sensible to me.

PR #4560, opened 2025-01-20
test: get rid of `EventBuilder`

bnjbvr:
This gets rid of `EventBuilder`, and makes more usage of the `EventFactory`, which is more ergonomic to create test events.

A large part of https://github.com/matrix-org/matrix-rust-sdk/issues/3716.


Hywan (approved):
🔥 what a ride! Thanks! I know how irritating it is…

PR #4561, opened 2025-01-21
chore(ffi): add a message focused timeline variant that allows focusing on a particular event

### LINKED ISSUE #4554: API request: load a media timeline from a specific event
The app needs to be able to load a media timeline from a specific event, to be able to navigate through media using the swipe gesture.

Currently we have an API `Room.messageFilteredTimeline` that returns a timeline.

A possible solution could be to add an optional `eventId` parameter to this method.

stefanceriu:
What it says on the tin. Fixes #4554

### DIFF
--- a/bindings/matrix-sdk-ffi/src/room.rs
+++ b/bindings/matrix-sdk-ffi/src/room.rs
@@ -207,15 +207,13 @@ impl Room {
 
         builder = builder.with_focus(configuration.focus.try_into()?);
 
-        if let AllowedMessageTypes::Only(allowed_message_types) =
-            configuration.allowed_message_types
-        {
+        if let AllowedMessageTypes::Only { types } = configuration.allowed_message_types {
             builder = builder.event_filter(move |event, room_version_id| {
                 default_event_filter(event, room_version_id)
                     && match event {
                         AnySyncTimelineEvent::MessageLike(msg) => match msg.original_content() {
                             Some(AnyMessageLikeEventContent::RoomMessage(content)) => {
-                                allowed_message_types.contains(&content.msgtype.into())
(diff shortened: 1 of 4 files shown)


bindings/matrix-sdk-ffi/src/room.rs
```diff
@@ -83,6 +85,36 @@ impl Room {
     pub(crate) fn with_timeline(inner: SdkRoom, timeline: TimelineLock) -> Self {
         Room { inner, timeline }
     }
+
+    fn message_filtered_timeline_builder(
+        &self,
+        internal_id_prefix: Option<String>,
+        allowed_message_types: Vec<RoomMessageEventMessageType>,
+        date_divider_mode: DateDividerMode,
+    ) -> TimelineBuilder {
```
bnjbvr, line 94: It's a bit weird to have a function returning a builder, but with optional arguments. Those optional arguments could become setters on the builder itself, instead, and could be exposed in the public API. This way, we could allow for more control over how the timelines are created, with a likely simpler API. What you made here works, but will need to change the next time anyone requires a new configuration. What do you think?

stefanceriu: I'm a tad confused, don't think I understand what you mean.

Why is it weird? It's just a private helper method to reuse configuration code between the 2 different message filtered timelines🤔 

The only optional argument here is `with_internal_id_prefix` which is already part of both the builder and the existing timeline creation methods. All `message_filtered_timeline_builder ` is make is so it doesn't need to be unwrapped in 2 different place.

Either way, this PR is not about refactoring how timelines are created. If you wish we can tackle that separately.

bnjbvr (approved):
Super nice, thanks!

PR #4562, opened 2025-01-21
test: minimize usage of `EventFactory::state_key`

bnjbvr:
It was used in places where we could make use of other helpers, in some cases. Also introduces the `room_avatar` helper to create the room avatar state event.

Part of https://github.com/matrix-org/matrix-rust-sdk/issues/3716

### DIFF
--- a/crates/matrix-sdk/tests/integration/room/common.rs
+++ b/crates/matrix-sdk/tests/integration/room/common.rs
@@ -12,11 +12,7 @@ use ruma::{
     event_id,
     events::{
         direct::DirectUserIdentifier,
-        room::{
-            avatar::{self, RoomAvatarEventContent},
-            member::MembershipState,
-            message::RoomMessageEventContent,
-        },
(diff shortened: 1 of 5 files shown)


crates/matrix-sdk/tests/integration/room/common.rs
```diff
@@ -881,9 +877,7 @@ async fn test_room_avatar() {
     // Set the avatar, but not the info.
     let avatar_url_1 = mxc_uri!("mxc://server.local/abcdef");
 
-    let mut content = RoomAvatarEventContent::new();
-    content.url = Some(avatar_url_1.to_owned());
-    let event = factory.event(content).state_key("").into_raw_sync();
+    let event = factory.room_avatar().url(avatar_url_1).into_raw_sync();
```
Hywan, line 880: So smooth.

Hywan (approved):
w00t

PR #4565, opened 2025-01-21
test: remove some uses of `sync_timeline_event!` in the base and sdk crates

bnjbvr:
Part of #3716.

### DIFF
--- a/crates/matrix-sdk-base/src/client.rs
+++ b/crates/matrix-sdk-base/src/client.rs
@@ -1749,8 +1749,6 @@ mod tests {
         async_test, ruma_response_from_json, sync_timeline_event, InvitedRoomBuilder,
         LeftRoomBuilder, StateTestEvent, StrippedStateTestEvent, SyncResponseBuilder,
     };
-    #[cfg(feature = "e2e-encryption")]
-    use ruma::UserId;
     use ruma::{api::client as api, room_id, serde::Raw, user_id};
     use serde_json::{json, value::to_raw_value};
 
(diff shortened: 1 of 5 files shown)


crates/matrix-sdk-base/src/client.rs
```diff
@@ -1898,13 +1896,32 @@ mod tests {
...
         let client = logged_in_base_client(Some(user_id)).await;
-        let room = process_room_join_test_helper(&client, room_id, "$1", user_id).await;
+
+        let mut sync_builder = SyncResponseBuilder::new();
+
+        let f = EventFactory::new();
+        let response = sync_builder
+            .add_joined_room(
+                matrix_sdk_test::JoinedRoomBuilder::new(room_id).add_timeline_event(
+                    f.member(user_id)
```
jmartinesp, line 1916: Maybe inline `f` here?

crates/matrix-sdk-base/src/client.rs
```diff
@@ -1768,18 +1771,16 @@ mod tests {
...
-                },
-                "event_id": "$994173582443PhrSn:example.org",
-                "origin_server_ts": 1432135524678u64,
-                "sender": user_id,
-                "state_key": user_id,
-                "type": "m.room.member",
-            })))
+            .add_left_room(
+                LeftRoomBuilder::new(room_id).add_timeline_event(
+                    f.member(user_id)
```
jmartinesp, line 1778: Same, as it's only used once it could be inlined.

jmartinesp (approved):
LGTM, I'd inline some usages of `EventFactory`, but it's not a big deal.

PR #4566, opened 2025-01-21
feat(sdk): (re)Enable HTTP/2 support

zecakeh:
It became an optional default feature in reqwest 0.12, and we disable the default features, so I don't think it was meant to be disabled when the crate was upgraded in https://github.com/matrix-org/matrix-rust-sdk/pull/3362.


andybalaam (approved):
Seems very sensible to me, but deferring to @poljar for the final word.

poljar (approved):

PR #4567, opened 2025-01-21
chore(ui): Remove `Timeline::subscribe`

Hywan:
This patch removes `Timeline::subscribe` and keeps `::subscribe_batched`, which is the only method that is really used.

Edit: I agree that's a non-negligible amount of work, but we want to clean up this API. The more API, the more maintenance and documentation. Also, `subscribe_batched` is more performant from the caller's perspective. Finally, we want the test to use the same API as more users are using, which is the case now with this PR.

### DIFF
--- a/benchmarks/benches/room_bench.rs
+++ b/benchmarks/benches/room_bench.rs
@@ -198,7 +198,7 @@ pub fn load_pinned_events_benchmark(c: &mut Criterion) {
                 .await
                 .expect("Could not create timeline");
 
-            let (items, _) = timeline.subscribe().await;
+            let (items, _) = timeline.subscribe_batched().await;
             assert_eq!(items.len(), PINNED_EVENTS_COUNT + 1);
             timeline.clear().await;
         });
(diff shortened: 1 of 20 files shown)


jplatte (commented):
This is a lot of effort to remove a method that's a couple of lines long. And this is a public method, are you sure it's not useful to _some_ users? (maybe Fractal?)

zecakeh:
Speaking for Fractal, we use `subscribe_batched`.

zecakeh:
I believe that code that uses `subscribe()` should be able to switch to `subscribe_batched()` easily, it just adds a `Vec` around the diffs in the stream.

Hywan:
I agree that's a non-negligible amount of work, but we want to clean up this API. The more API, the more maintenance and documentation. Also, `subscribe_batched` is more performant from the caller's perspective. Finally, we want the test to use the same API as more users are using, which is the case now with this PR.

stefanceriu (approved):
lgtm 👍

Hywan:
Rebased to get the new Lint action running properly.

PR #4568, opened 2025-01-21
refactor!(all the things): fuse `SyncTimelineEvent` into `TimelineEvent`

bnjbvr:
The breakiest of breaking changes:

- The only difference was in their constructors; `SyncTimelineEvent` took a `Raw<AnySyncTimelineEvent>`, and `TimelineEvent` took a `Raw<AnyTimelineEvent>` that it immediately cast into `Raw<AnySyncTimelineEvent>`, so we can use both.
- `TimelineEvent` had its `push_actions` optional, which is the right call; another commit makes that change, partially reverting some changes of the first commit, sorry!
- `SyncTimelineEvent` is eventually renamed to `TimelineEvent`, but as the last commit suggests… maybe this could be plain called `Event`? (Would likely break some serialization formats that rely on type names)


crates/matrix-sdk-base/CHANGELOG.md
```diff
@@ -11,6 +11,9 @@ All notable changes to this project will be documented in this file.
 - Replaced `Room::compute_display_name` with the reintroduced `Room::display_name()`. The new
   method computes a display name, or return a cached value from the previous successful computation.
   If you need a sync variant, consider using `Room::cached_display_name()`.
+- [**breaking**]: The reexported types `SyncTimelineEvent` and `TimelineEvent` have been fused into a single type
+  `TimelineEvent`, and its field `push_actions` has been made `Option`al (it is set to `None` when
+  we couldn't compute the push actions, because we lacked some information).
```
poljar, line 16: Please add a link to the PR introducing this change.

crates/matrix-sdk-base/CHANGELOG.md
```diff
@@ -11,6 +11,9 @@ All notable changes to this project will be documented in this file.
 - Replaced `Room::compute_display_name` with the reintroduced `Room::display_name()`. The new
   method computes a display name, or return a cached value from the previous successful computation.
   If you need a sync variant, consider using `Room::cached_display_name()`.
+- [**breaking**]: The reexported types `SyncTimelineEvent` and `TimelineEvent` have been fused into a single type
```
poljar, line 14:
```suggestion

- [**breaking**]: The reexported types `SyncTimelineEvent` and `TimelineEvent` have been fused into a single type
```

crates/matrix-sdk-common/CHANGELOG.md
```diff
@@ -6,6 +6,10 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased] - ReleaseDate
 
+- [**breaking**]: `SyncTimelineEvent` and `TimelineEvent` have been fused into a single type
+  `TimelineEvent`, and its field `push_actions` has been made `Option`al (it is set to `None` when
+  we couldn't compute the push actions, because we lacked some information).
```
poljar, line 11: Same, PR link please.

crates/matrix-sdk/CHANGELOG.md
```diff
@@ -25,6 +25,10 @@ All notable changes to this project will be documented in this file.
 
 ### Refactor
 
+- [**breaking**]: The reexported types `SyncTimelineEvent` and `TimelineEvent` have been fused into a single type
+  `TimelineEvent`, and its field `push_actions` has been made `Option`al (it is set to `None` when
+  we couldn't compute the push actions, because we lacked some information).
```
poljar, line 30: Another one.

poljar (changes requested):
I think this makes sense and looks good as well. Please just add the missing PR links to the changelog entries.

poljar:
> maybe this could be plain called Event?

Eh, not so sure, we already have another `Event<C>` type in the crypto crate, people usually complain when we have multiple types with the same name.

crates/matrix-sdk-base/CHANGELOG.md
bnjbvr: I'm not against this change, but can I ask why we should do it like this? This will break the list into two lists, now, which seems counter to the point of having *a* list of changes.

bnjbvr:
Had to rebase for the new `Lint` CI job.

crates/matrix-sdk-base/CHANGELOG.md
poljar: I don't particularly care, but if you prefer the rendering without the extra newlines, then we shouldn't have them in the other changelog either.

poljar (approved):

Issue #4466, opened 2025-01-06
Dehydrated devices: drop incoming to-device messages from dehydrated devices


richvdh:
Dehydrated devices should never generate to-device messages, so incoming to-device messages from a dehydrated device indicate a faulty implementation.

We have decided to drop such messages, per [the MSC](https://github.com/uhoreg/matrix-doc/blob/6be907845c1dae03410a4c17f08d601e4dab5c56/proposals/3814-dehydrated-devices-with-ssss.md?plain=1#L53-L55)

PR #4569, opened 2025-01-21
Ignore to-device messages from dehydrated devices

andybalaam:
Fixes https://github.com/matrix-org/matrix-rust-sdk/issues/4466

### DIFF
--- a/crates/matrix-sdk-crypto/src/machine/mod.rs
+++ b/crates/matrix-sdk-crypto/src/machine/mod.rs
@@ -852,9 +852,14 @@ impl OlmMachine {
         let mut decrypted =
             transaction.account().await?.decrypt_to_device_event(&self.inner.store, event).await?;
 
-        // Handle the decrypted event, e.g. fetch out Megolm sessions out of
-        // the event.
-        self.handle_decrypted_to_device_event(transaction.cache(), &mut decrypted, changes).await?;
+        // We ignore all to-device events from dehydrated devices - we should not
+        // receive any
+        if !self.to_device_event_is_from_dehydrated_device(&decrypted, &event.sender).await? {
+            // Handle the decrypted event, e.g. fetch out Megolm sessions out of
+            // the event.
+            self.handle_decrypted_to_device_event(transaction.cache(), &mut decrypted, changes)
+                .await?;
+        }
 
         Ok(decrypted)
     }
@@ -1259,13 +1264,20 @@ impl OlmMachine {
         }
     }
 
+    /// Decrypt the supplied to-device event (if needed, and if we can) and
+    /// handle it.
+    ///
+    /// Return the same event, decrypted if possible and needed.
+    ///
+    /// If we can identify that this to-device event came from a dehydrated
+    /// device, this method does not process it, and returns `None`.
     #[instrument(skip_all, fields(sender, event_type, message_id))]
     async fn receive_to_device_event(
         &self,
         transaction: &mut StoreTransaction,
         changes: &mut Changes,
         mut raw_event: Raw<AnyToDeviceEvent>,
-    ) -> Raw<AnyToDeviceEvent> {
+    ) -> Option<Raw<AnyToDeviceEvent>> {
         Self::record_message_id(&raw_event);
 
         let event: ToDeviceEvents = match raw_event.deserialize_as() {
@@ -1274,7 +1286,7 @@ impl OlmMachine {
                 // Skip invalid events.
                 warn!("Received an invalid to-device event: {e}");
 
-                return raw_event;
+                return Some(raw_event);
             }
         };
 
@@ -1299,9 +1311,29 @@ impl OlmMachine {
                             }
                         }
 
-                        return raw_event;
+                        return Some(raw_event);
                     }
                 };
+
+                // We ignore all to-device events from dehydrated devices - we should not
+                // receive any
+                match self.to_device_event_is_from_dehydrated_device(&decrypted, &e.sender).await {
+                    Ok(true) => {
+                        warn!(
+                            sender = ?e.sender,
+                            session = ?decrypted.session,
+                            "Received a to-device event from a dehydrated device. This is unexpected: ignoring event"
+                        );
+                        return None;
+                    }
+                    Ok(false) => {}
+                    Err(err) => {
+                        error!(
+                            error = ?err,
+                            "Couldn't check whether event is from dehydrated device",
+                        );
+                    }
+                }
 
                 // New sessions modify the account so we need to save that
                 // one as well.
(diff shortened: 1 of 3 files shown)


crates/matrix-sdk-crypto/src/machine/mod.rs
```diff
@@ -852,9 +852,14 @@ impl OlmMachine {
...
-        // Handle the decrypted event, e.g. fetch out Megolm sessions out of
-        // the event.
-        self.handle_decrypted_to_device_event(transaction.cache(), &mut decrypted, changes).await?;
+        // We ignore all to-device events from dehydrated devices - we should not
+        // receive any
+        if !self.event_is_from_dehydrated_device(&decrypted, &event.sender).await {
+            // Handle the decrypted event, e.g. fetch out Megolm sessions out of
+            // the event.
+            self.handle_decrypted_to_device_event(transaction.cache(), &mut decrypted, changes)
+                .await?;
```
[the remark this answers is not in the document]
andybalaam: (To try to clarify what I mean: my tests would still pass if this `if` did not exist.)

crates/matrix-sdk-crypto/src/machine/mod.rs
```diff
@@ -1336,7 +1359,54 @@ impl OlmMachine {
...
+            if device_keys.dehydrated.unwrap_or(false) {
+                return true;
+            }
+            // If not, fall through and check our existing list of devices
+            // below, just in case the sender is sending us
+            // incorrect information embedded in the to-device message, but we
+            // know better.
+        }
+
+        // Do we already know about this device?
```
andybalaam, line 1386: My tests would still pass if the rest of this function did not exist.

I looked into how to construct a to-device message that was from a known dehydrated device, but did not itself contain `device_keys` and it just looked too complex. I even got part-way in to an implementation pairing with Rich, but it just seemed so much!

andybalaam:
Any help with how to construct test cases that exercise the other parts of this code would be much appreciated! I have essentially given up because it seemed too hard.

crates/matrix-sdk-crypto/src/machine/tests/mod.rs
```diff
@@ -434,6 +414,32 @@ async fn test_room_key_sharing() {
     assert_eq!(room_key_updates[0].session_id, alice_session.session_id());
 }
 
+async fn send_room_key_to_device(
+    alice: &OlmMachine,
+    bob: &OlmMachine,
+    room_id: &RoomId,
+) -> OlmResult<(Vec<Raw<AnyToDeviceEvent>>, Vec<RoomKeyInfo>)> {
```
richvdh, line 449: might be worth a bit of documentation saying this sends *from* alice *to* bob. (Alternatively, I guess you could use better names for the users.)

Could you also document what it returns?

crates/matrix-sdk-crypto/src/machine/mod.rs
```diff
@@ -1299,10 +1311,21 @@ impl OlmMachine {
...
                     }
                 };
 
+                // We ignore all to-device events from dehydrated devices - we should not
+                // receive any
+                if self.event_is_from_dehydrated_device(&decrypted, &e.sender).await {
+                    warn!(
+                        sender = ?e.sender,
+                        session = ?decrypted.session,
+                        "Received a to-device event from a dehydrated device. This is unexpected"
```
richvdh, line 1324:
```suggestion
                        "Received a to-device event from a dehydrated device. This is unexpected: ignoring event"
```

crates/matrix-sdk-crypto/src/machine/mod.rs
```diff
@@ -1336,7 +1359,54 @@ impl OlmMachine {
             e => self.handle_to_device_event(changes, &e).await,
         }
 
-        raw_event
+        Some(raw_event)
+    }
+
+    async fn event_is_from_dehydrated_device(
```
richvdh, line 1365:
```suggestion
    async fn to_device_event_is_from_dehydrated_device(
```

crates/matrix-sdk-crypto/src/machine/mod.rs
```diff
@@ -1336,7 +1359,54 @@ impl OlmMachine {
...
+        &self,
+        decrypted: &OlmDecryptionInfo,
+        sender_user_id: &UserId,
+    ) -> bool {
```
richvdh, line 1369: it might come as a bit of a surprise that this accepts an `OlmDecryptionInfo` rather than, say, an `AnyToDeviceEvent`. Might be worth adding a doc-comment that explains that it works on *decrypted* to-device events.

crates/matrix-sdk-crypto/src/machine/mod.rs
```diff
@@ -1336,7 +1359,54 @@ impl OlmMachine {
...
+            .get_device_from_curve_key(sender_user_id, decrypted.result.sender_key)
+            .await
+        {
+            Ok(device) => {
+                // Yes - is it dehydrated?
+                if let Some(device) = device {
+                    device.is_dehydrated()
+                } else {
+                    false
+                }
```
richvdh, line 1398: light suggestion, no strong feelings here

```suggestion
                device.is_some_and(Device::is_dehydrated)
```

crates/matrix-sdk-crypto/src/machine/mod.rs
```diff
@@ -1336,7 +1359,54 @@ impl OlmMachine {
...
+            }
+            Err(e) => {
+                // No - we received a to-device message from a device we don't know about. We
+                // continue to process it, but this is unexpected so we log an error.
+                warn!(
+                    error = ?e,
+                    "Received a to-device message from a device we don't recognise",
+                );
```
richvdh, line 1406: I think if we just don't recognise the device, `get_device_from_curve_key` will return `Ok(None)`. This suggests a problem with the store.

It's probably still ok to return `false` here; but the comment and warning could do with a tweak.

crates/matrix-sdk-crypto/src/machine/mod.rs
```diff
@@ -1336,7 +1359,54 @@ impl OlmMachine {
...
+                // Yes - is it dehydrated?
+                if let Some(device) = device {
+                    device.is_dehydrated()
+                } else {
+                    false
+                }
+            }
+            Err(e) => {
+                // No - we received a to-device message from a device we don't know about. We
+                // continue to process it, but this is unexpected so we log an error.
```
richvdh, line 1402: >  we log an error

We are logging a warning, not an error.

richvdh (changes requested):
A few nits, generally looks about right to me

crates/matrix-sdk-crypto/src/machine/mod.rs
andybalaam: Done in caa557e53f

crates/matrix-sdk-crypto/src/machine/mod.rs
andybalaam: Done in 38ab50ecc2

crates/matrix-sdk-crypto/src/machine/mod.rs
andybalaam: Changed in e7eaa3e539 (exact suggestion didn't because it would have need an `as_ref` (I think))

crates/matrix-sdk-crypto/src/machine/mod.rs
andybalaam: Thanks, fixed in 59caf0c119

crates/matrix-sdk-crypto/src/machine/tests/mod.rs
andybalaam: Fixed in 5ba5ada0e8 (renamed params)

crates/matrix-sdk-crypto/src/machine/mod.rs
andybalaam: Good catch, thanks. I surfaced the error in dd7bfeaf73 , following the pattern of nearby code about what to do with a storage error.

richvdh (approved):

richvdh:
lgtm

PR #4572, opened 2025-01-22
fix(wasm): don't kill task on drop

maan2003:
makes behavior similar to tokio::spawn otherwise spawned tasks immediately die if join handles are not saved somewhere

### DIFF
--- a/crates/matrix-sdk-common/src/executor.rs
+++ b/crates/matrix-sdk-common/src/executor.rs
@@ -47,13 +47,13 @@ where
         let _ = future.await;
     });
 
-    JoinHandle { remote_handle, abort_handle }
+    JoinHandle { remote_handle: Some(remote_handle), abort_handle }
 }
 
 #[cfg(target_arch = "wasm32")]
 #[derive(Debug)]
 pub struct JoinHandle<T> {
-    remote_handle: RemoteHandle<T>,
+    remote_handle: Option<RemoteHandle<T>>,
     abort_handle: AbortHandle,
 }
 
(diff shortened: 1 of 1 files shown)


maan2003:
cc @jplatte

crates/matrix-sdk-common/src/executor.rs
```diff
@@ -64,6 +64,16 @@ impl<T> JoinHandle<T> {
     }
 }
 
+#[cfg(target_arch = "wasm32")]
+impl<T> Drop for JoinHandle<T> {
+    fn drop(&mut self) {
+        // don't abort the spawned future
```
Hywan, line 70: This comment must be elaborated to explain why we don't abort the future when the join handle is dropped:

- Why?
- Does it match the Tokio's behaviour?

Thanks.

Hywan (changes requested):

Hywan:
Thanks for your contribution :-).

It would be nice if you could address my comment, and if tests could be added. This is an important change.

maan2003:
what do you think about using https://docs.rs/n0-future/latest/n0_future/ instead of custom stuff

crates/matrix-sdk-common/src/executor.rs
jplatte: Yes, it matches tokio's behavior and that's exactly the reason. I think we discussed it on Matrix previously that a cancel-on-drop `JoinHandle` might make more sense for the SDK, but then you would have to
- wrap the tokio `JoinHandle` for non-wasm
- update all the existing cases where `JoinHandle`s are dropped prematurely to store the handle somewhere, or call some `detach` method on it

That's a lot harder. I think you should accept a PR that fixes the existing inconsistency (which leads to buggy behavior), and evaluate the different `JoinHandle` semantics afterwards.

Hywan (approved):

PR #4573, opened 2025-01-22
fix(wasm): don't use tokio::time::{timeout,sleep}

maan2003:
tokio timeout,sleep doesn't work on wasm


maan2003:
not sure why crypto tests are failing, this should be noop on non wasm

andybalaam:
> not sure why crypto tests are failing, this should be noop on non wasm

Agreed, re-running in case it's a flake.

clippy.toml
maan2003, line 1: added this to not have this problem ever again

andybalaam:
Hey, please don't force push after you've asked for review, since makes it hard for me to know what I've reviewed or not.

maan2003:
sorry, will not force push again

crates/matrix-sdk-common/src/sleep.rs
```diff
@@ -0,0 +1,44 @@
...
+
+use std::time::Duration;
+
+/// Sleep for the specified duration.
+///
+/// This is a cross-platform sleep implementation that works on both wasm32 and
+/// non-wasm32 targets.
+pub async fn sleep(duration: Duration) {
+    #[cfg(not(target_arch = "wasm32"))]
+    #[allow(clippy::disallowed_methods)]
```
andybalaam, line 23: Do we need this, or is it left over from your clippy experiments?

crates/matrix-sdk-common/src/sleep.rs
```diff
@@ -0,0 +1,44 @@
...
+    tokio::time::sleep(duration).await;
+
+    #[cfg(target_arch = "wasm32")]
+    gloo_timers::future::TimeoutFuture::new(duration.as_millis() as u32).await;
```
andybalaam, line 27: I would prefer `u32::try_from(duration.as_millis()).unwrap_or_else(|_| { // log an error and return u32::MAX})`. I don't think we will have any performance concerns here since this is sleeping anyway, and it means if we ever use this method wrongly in future we will find out about it, non-fatally.

crates/matrix-sdk-common/src/timeout.rs
```diff
@@ -41,17 +41,18 @@ impl Error for ElapsedError {}
 /// an error.
 pub async fn timeout<F, T>(future: F, duration: Duration) -> Result<T, ElapsedError>
 where
-    F: Future<Output = T> + Unpin,
+    F: Future<Output = T>,
 {
     #[cfg(not(target_arch = "wasm32"))]
+    #[allow(clippy::disallowed_methods)]
```
andybalaam, line 47: Again, maybe not needed?

crates/matrix-sdk-common/src/timeout.rs
```diff
@@ -41,17 +41,18 @@ impl Error for ElapsedError {}
...
     return tokio_timeout(duration, future).await.map_err(|_| ElapsedError(()));
 
     #[cfg(target_arch = "wasm32")]
     {
         let timeout_future =
             TimeoutFuture::new(u32::try_from(duration.as_millis()).expect("Overlong duration"));
 
-        match select(future, timeout_future).await {
+        match select(std::pin::pin!(future), timeout_future).await {
```
andybalaam, line 55: This is a nice change! I would have preferred it in a separate PR, but I'm not going to insist on it.

andybalaam (changes requested):
This is great, thank you! I like small PRs, so ideally I would have liked 3 separate ones here: one for removing `Unpin` on `timeout`, one for `sleep` and one for `timeout`.

But I won't insist on that here, and if you address the small changes I asked for I am happy to merge it.

Thank you for working on this!

andybalaam:
> > not sure why crypto tests are failing, this should be noop on non wasm
> Agreed, re-running in case it's a flake.

It passed the second time. This test failed the first time:

❌ TestOnRejoinBobCanSeeButNotDecryptHistoryInPublicRoom (2.14s)
❌ TestOnRejoinBobCanSeeButNotDecryptHistoryInPublicRoom/{rust_hs1}|{rust_hs1} (2.14s)

We should keep an eye on this and see whether we have a flaky test.

andybalaam:
> sorry, will not force push again

No worries! Thank you. There are more tips here: https://github.com/matrix-org/matrix-rust-sdk/blob/main/CONTRIBUTING.md#review-process

crates/matrix-sdk-common/src/timeout.rs
maan2003: I wanted everything in separate PRs, but squashed all of them into one after clippy experiments :/

maan2003:
addressed review

andybalaam (approved):
Great, thank you!

PR #4574, opened 2025-01-22
fix(wasm): remove tokio::time::sleep usage on wasm 

maan2003:
followed the pattern used by timeout.rs


maan2003:
just merging this PR into https://github.com/matrix-org/matrix-rust-sdk/pull/4573

PR #4575, opened 2025-01-22
refactor(sdk): Move `matrix_auth` and `oidc` into the `authentication` module

zecakeh:
`matrix_auth` is renamed to `matrix` in the process since the `auth` is for "authentication".


poljar (changes requested):
Nice, could you please just remove the TODO I linked previously?

https://github.com/matrix-org/matrix-rust-sdk/blob/1d83d42e9fd77f1aed4acb7e3c7d8879ba54ba91/crates/matrix-sdk/src/authentication/mod.rs#L18

zecakeh:
I removed it already.

poljar (approved):
Ah indeed, how did I miss that.

PR #4576, opened 2025-01-22
fix(timeline): maintain aggregations when an event is deduplicated

bnjbvr:
## Some context

An aggregation is an event that relates to another event: for instance, a
reaction, a poll response, and so on and so forth.
                               
## Some requirements
                                              
Because of the sync mechanisms and federation, it can happen that a related
event is received *before* receiving the event it relates to. Those events
must be accounted for, stashed somewhere, and reapplied later, if/when the
related-to event shows up.
                                                                             
In addition to that, a room's event cache can also decide to move events
around, in its own internal representation (likely because it ran into some
duplicate events, or it managed to decrypt a previously UTD event).
When that happens, a timeline opened on the given room
will see a removal then re-insertion of the given event. If that event was
the target of aggregations, then those aggregations must be re-applied when
the given event is reinserted.
                                                                       
## Some solution
      
To satisfy both requirements, the [`Aggregations`] "manager" object provided
by this PR will take care of memoizing aggregations, **for the entire
lifetime of the timeline** (or until it's clear'd by some
caller). Aggregations are saved in memory, and have the same lifetime as
that of a timeline. This makes it possible to apply pending aggregations
to cater for the first use case, and to never lose any aggregations in the
second use case.

## Some points for the reviewer

- I think the most controversial point is that all aggregations are memoized for the entire lifetime of the timeline. Would that become an issue, we can get back to some incremental scheme, in the future: instead of memoizing aggregations for the entire lifetime of the timeline, we'd attach them to a single timeline item. When that item is removed, we'd put the aggregations back into a "pending" stash of aggregations. If the item is reinserted later, we could peek at the pending stash of aggregations, remove any that's in there, and reapply them to the reinserted event. This is what the [first version of this patch](https://github.com/matrix-org/matrix-rust-sdk/pull/4576/commits/ec64b9e0bcc9a2d05dd8c910a8710b72466ef51c) did, in a much more adhoc way, for reactions only; based on the current PR, we could do the same in a simpler manner
- while the PR has small commits, they don't quite make sense to review individually, I'm afraid, as I was trying to find a way to make a general system that would work not only for reactions, poll responses and ends. As a matter of fact, the first commits may have introduced code that is changed in subsequent commits, making the review a bit hazardous. Happy to have a live reviewing party over Element Call, if that helps, considering the size of the patch.
- future work may include using the aggregations manager for edits too, leading to more code removal.

Part of https://github.com/matrix-org/matrix-rust-sdk/issues/3280.


andybalaam:
A live review party sounds good to me, and I've asked some other members of the Rust team whether they'd like to join because I'm not sure I should look at this on my own.

crates/matrix-sdk-ui/src/timeline/algorithms.rs
```diff
@@ -39,7 +39,13 @@ impl EventTimelineItemWithId<'_> {
     /// Create a clone of the underlying [`TimelineItem`] with the given
     /// reactions.
     pub fn with_reactions(&self, reactions: ReactionsByKeyBySender) -> Arc<TimelineItem> {
-        let event_item = self.inner.with_reactions(reactions);
+        let content = self.inner.content().with_reactions(reactions);
+
+        // Do not use `Self::with_content` which may override the latest_edit_json.
+        // TODO: it's likely incorrect?
```
andybalaam, line 45: I don't understand this comment -  can you expand on it? If it's incorrect, do we need to fix it before we can commit this?

crates/matrix-sdk-ui/src/timeline/controller/aggregations.rs
```diff
@@ -0,0 +1,429 @@
...
+    /// Apply an aggregation in-place to a given [`TimelineItemContent`].
+    ///
+    /// In case of success, returns a boolean indicating whether applying the
+    /// aggregation caused a change in the content. If the aggregation could be
+    /// applied, but it didn't cause any change in the passed
+    /// [`TimelineItemContent`], `Ok(false)` will be returned.
+    ///
+    /// In case of error, returns an error detailing why the aggregation
+    /// couldn't be applied.
+    pub fn apply(&self, content: &mut TimelineItemContent) -> Result<bool, AggregationError> {
```
andybalaam, line 131: As discussed irl, returning an enum instead of the bool here would make it clearer, I think.

crates/matrix-sdk-ui/src/timeline/controller/aggregations.rs
```diff
@@ -0,0 +1,429 @@
...
+            }
+        }
+    }
+
+    /// Undo an aggregation in-place to a given [`TimelineItemContent`].
+    ///
+    /// In case of success, returns a boolean indicating whether unapplying the
+    /// aggregation caused a change in the content. If the aggregation could be
+    /// unapplied, but it didn't cause any change in the passed
+    /// [`TimelineItemContent`], `Ok(false)` will be returned.
```
andybalaam, line 187: Also here, an enum return value would help to understand.

crates/matrix-sdk-ui/src/timeline/controller/aggregations.rs
```diff
@@ -0,0 +1,429 @@
...
+        for a in aggregations {
+            a.apply(content)?;
+        }
+        Ok(())
+    }
+
+    /// Mark a target event as being sent (i.e. it transitions from an local
+    /// transaction id to its remote event id counterpart), by updating the
+    /// internal mappings.
+    pub fn mark_target_as_sent(&mut self, txn_id: OwnedTransactionId, event_id: OwnedEventId) {
```
andybalaam, line 307: It would be good to have some unit tests for this.

crates/matrix-sdk-ui/src/timeline/event_handler.rs
```diff
@@ -974,43 +915,42 @@ impl<'a, 'o> TimelineEventHandler<'a, 'o> {
...
+                    self.items.replace(item_pos, TimelineItem::new(new_item, internal_id));
+                    self.result.items_updated += 1;
+                }
+                Ok(false) => {}
+                Err(err) => {
+                    warn!("error when unapplying aggregation: {err}");
+                }
             }
+        } else {
+            warn!("missing related-to item ({target:?}) for aggregation {aggregation_id:?}");
```
andybalaam, line 949: This is probably not a warning since it is fine to happen.

andybalaam (approved):
I'd love to see some tests for the Aggregations manager, but looks good!

crates/matrix-sdk-ui/src/timeline/algorithms.rs
bnjbvr: Ah yeah, something I noticed, and fixed in [another PR](https://github.com/matrix-org/matrix-rust-sdk/pull/4611); I can remove the comment now, thanks for spotting it!

crates/matrix-sdk-ui/src/timeline/controller/aggregations.rs
bnjbvr: Yep, will write a bunch in a subsequent PR!

PR #4577, opened 2025-01-23
feat(ui): expose a method for checking whether a message contains only emojis and should be boosted (use a bigger font size)

stefanceriu:
- supports only text room message types
- enumerates through their body's grapheme clusters and check that every single one of them is an emoji
- part of the `LazyTimelineItemProvider` so that it can be opt in

### DIFF
--- a/crates/matrix-sdk-ui/src/timeline/event_item/mod.rs
+++ b/crates/matrix-sdk-ui/src/timeline/event_item/mod.rs
@@ -620,7 +620,14 @@ impl EventTimelineItem {
     ///   [`MessageType::Video`] if present
     /// - all other message types will not match
     ///
-    /// See `test_emoji_detection` for examples.
+    /// # Examples
+    /// # fn render_timeline_item(timeline_item: TimelineItem) {
+    /// if timeline_item.contains_only_emojis() {
+    ///     // e.g. increase the font size
+    /// }
+    /// # }
+    ///
+    /// See `test_emoji_detection` for more examples.
     pub fn contains_only_emojis(&self) -> bool {
         let body = match self.content() {
             TimelineItemContent::Message(msg) => match msg.msgtype() {
--- a/crates/matrix-sdk-ui/src/timeline/event_item/mod.rs
+++ b/crates/matrix-sdk-ui/src/timeline/event_item/mod.rs
@@ -656,7 +656,9 @@ impl EventTimelineItem {
             let graphemes = body.trim().graphemes(true).collect::<Vec<&str>>();
 
             // Limit the check to 5 graphemes for performance and security
-            // reasons
+            // reasons. This will probably be used for every new message so we
+            // want it to be fast and we don't want to allow a DoS attack by
+            // sending a huge message.
             if graphemes.len() > 5 {
                 return false;
             }
--- a/crates/matrix-sdk-ui/src/timeline/event_item/mod.rs
+++ b/crates/matrix-sdk-ui/src/timeline/event_item/mod.rs
@@ -36,6 +36,7 @@ use ruma::{
     OwnedUserId, RoomId, RoomVersionId, TransactionId, UserId,
 };
 use tracing::warn;
+use unicode_segmentation::UnicodeSegmentation;
 
 mod content;
 mod local;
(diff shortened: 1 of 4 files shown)


crates/matrix-sdk-ui/src/timeline/event_item/mod.rs
```diff
@@ -442,6 +443,19 @@ impl EventTimelineItem {
         }
     }
 
+    pub fn should_boost(&self) -> bool {
+        match self.content() {
+            TimelineItemContent::Message(msg) => match msg.msgtype() {
+                MessageType::Text(text) => {
+                    let graphmes = text.body.graphemes(true).collect::<Vec<&str>>();
+                    graphmes.iter().all(|g| emojis::get(g).is_some())
```
pixlwave, line 451: Would it be easy to strip any leading/trailing whitespace from the body before doing this so that we can boost like EW does?

crates/matrix-sdk-ui/src/timeline/event_item/mod.rs
```diff
@@ -442,6 +443,19 @@ impl EventTimelineItem {
         }
     }
 
+    pub fn should_boost(&self) -> bool {
```
Hywan, line 446: Please add a documentation with an example :-).

Hywan (commented):
Neat!

crates/matrix-sdk-ui/src/timeline/event_item/mod.rs
```diff
@@ -442,6 +443,19 @@ impl EventTimelineItem {
...
+        match self.content() {
+            TimelineItemContent::Message(msg) => match msg.msgtype() {
+                MessageType::Text(text) => {
```
frebib, line 449: What about other event types with captions?

crates/matrix-sdk-ui/src/timeline/event_item/mod.rs
stefanceriu: Documentation sorted but I'm not so sure about the example, seems to involved.

crates/matrix-sdk-ui/src/timeline/event_item/mod.rs
stefanceriu: Yup, good idea, added it in.

crates/matrix-sdk-ui/src/timeline/event_item/mod.rs
stefanceriu: 👍  added caption support for the Audio, File, Image and Video MessageTypes

crates/matrix-sdk-ui/src/timeline/event_item/mod.rs
Hywan: Pro tips: You can create declare a function that takes a `TimelineItem` as a parameter, however the function is “hidden” from the doc:

```rust
# fn test(timeline_item: TimelineItem) {
timeline_item.do_something();
# }
```

`rustdoc` will be able to run the test by checking the types (not the execution though, because `test()` won't run), and `rustdoc` will also be able to render the doc, as:

```rust
timeline_item.do_something();
```

Hywan: See https://doc.rust-lang.org/rustdoc/write-documentation/documentation-tests.html#hiding-portions-of-the-example to learn more.

crates/matrix-sdk-ui/src/timeline/event_item/mod.rs
```diff
@@ -618,6 +605,60 @@ impl EventTimelineItem {
...
+            | TimelineItemContent::CallInvite
+            | TimelineItemContent::CallNotify => None,
+        };
+
+        if let Some(body) = body {
+            // Collect the graphemes after trimming white spaces.
+            let graphemes = body.trim().graphemes(true).collect::<Vec<&str>>();
+
+            // Limit the check to 5 graphemes for performance and security
+            // reasons
```
Hywan, line 659: Performance aspect: why not, I can see.
Security aspect: why? Can you explain a bit please? A link to something maybe?

crates/matrix-sdk-ui/src/timeline/event_item/mod.rs
```diff
@@ -442,6 +443,19 @@ impl EventTimelineItem {
         }
     }
 
+    pub fn should_boost(&self) -> bool {
+        match self.content() {
+            TimelineItemContent::Message(msg) => match msg.msgtype() {
+                MessageType::Text(text) => {
+                    let graphmes = text.body.graphemes(true).collect::<Vec<&str>>();
```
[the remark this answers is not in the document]
stefanceriu: It's already on there, this is from an older commit.

PR #4578, opened 2025-01-23
feat(beacons): expose ffi functions to start, stop and subscribe to live location shares

torrybr:
FFI functions for starting ,stopping and subscribing to Live Location Shares. 

Signed-off-by: 

Torry Brelsford

### DIFF
--- a/bindings/matrix-sdk-ffi/src/room.rs
+++ b/bindings/matrix-sdk-ffi/src/room.rs
@@ -1001,13 +1001,16 @@ impl Room {
         self: Arc<Self>,
         listener: Box<dyn LiveLocationShareListener>,
     ) -> Arc<TaskHandle> {
-        let subscription = self.inner.observe_live_location_shares();
-        let stream = subscription.subscribe();
+        let room = self.inner.clone();
+
         Arc::new(TaskHandle::new(RUNTIME.spawn(async move {
-            pin_mut!(stream);
-            while let Some(event) = stream.next().await {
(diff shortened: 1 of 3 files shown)


bindings/matrix-sdk-ffi/src/room.rs
```diff
@@ -969,6 +970,73 @@ impl Room {
...
+                let last_location = LocationContent {
+                    body: "".to_owned(),
+                    geo_uri: event.last_location.location.uri.clone().to_string(),
+                    description: None,
+                    zoom_level: None,
+                    asset: None,
+                };
+                let beacon_info = event
+                    .beacon_info
+                    .expect("Live location share is missing the associated beacon_info state");
```
jmartinesp, line 1021: I think this should be something like:

```rust
let Some(beacon_info) = event.beacon_info else {
    warn!("Live location share is missing the associated beacon_info state, skipping event.");
    continue;
};
```

Since otherwise this would probably mean a crash in the client, which seems like a bit of an overkill for just finding an error while broadcasting some events.

jmartinesp (changes requested):
Thanks for the changes, I think there might be an issue with unwrapping the `beacon_info` field, and we shouldn't cause a panic when we can't do it.

bindings/matrix-sdk-ffi/src/room.rs
torrybr: Updated! Good catch thanks

jmartinesp (approved):

PR #4584, opened 2025-01-27
refactor(ui): Unify the Timeline pagination API

Hywan:
This patch simplifies the `Timeline` pagination API as follows:

- a unique `paginate_backwards` method (no more
  `focused_paginate_backwards` and `live_paginate_backwards`),
- a unique `paginate_forwards` method (no more
  `focused_paginate_forwards`, the `live` variant was absent).

The idea is to unify pagination by hiding the `live` and `focused` mode.
It was already partially the case with `paginate_backards`, but the
`live` and `focused` variants were also present. I believe it creates
an unnecessary confusion.

### DIFF
--- a/bindings/matrix-sdk-ffi/src/timeline/mod.rs
+++ b/bindings/matrix-sdk-ffi/src/timeline/mod.rs
@@ -19,8 +19,6 @@ use as_variant::as_variant;
 use content::{InReplyToDetails, RepliedToEventDetails};
 use eyeball_im::VectorDiff;
 use futures_util::{pin_mut, StreamExt as _};
-#[cfg(doc)]
-use matrix_sdk::crypto::CollectStrategy;
 use matrix_sdk::{
     attachment::{
         AttachmentConfig, AttachmentInfo, BaseAudioInfo, BaseFileInfo, BaseImageInfo,
@@ -63,8 +61,6 @@ use tracing::{error, warn};
 use uuid::Uuid;
 
 use self::content::{Reaction, ReactionSenderData, TimelineItemContent};
-#[cfg(doc)]
-use crate::client_builder::ClientBuilder;
 use crate::{
     client::ProgressWatcher,
     error::{ClientError, RoomError},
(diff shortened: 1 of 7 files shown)


pixlwave (commented):
Nice, I'm in favour of this from an API perspective - we're already doing this in EX in the [TimelineProxy](https://github.com/element-hq/element-x-ios/blob/a9d206ff8a2ce3bd58d6f12113a90cc577046346/ElementX/Sources/Services/Timeline/TimelineProxy.swift#L92-L96)).

However we currently handle the pagination state differently for [live](https://github.com/element-hq/element-x-ios/blob/a9d206ff8a2ce3bd58d6f12113a90cc577046346/ElementX/Sources/Services/Timeline/TimelineProxy.swift#L114-L117) and [detached](https://github.com/element-hq/element-x-ios/blob/a9d206ff8a2ce3bd58d6f12113a90cc577046346/ElementX/Sources/Services/Timeline/TimelineProxy.swift#L139-L150) timelines as detached timelines [don't support observation](https://github.com/element-hq/element-x-ios/blob/a9d206ff8a2ce3bd58d6f12113a90cc577046346/ElementX/Sources/Services/Timeline/TimelineProxy.swift#L590-L590). So for us to adopt this, would it be possible to add pagination state observation to detached timelines 😇?

pixlwave:
For context on the above, we have observation on live timelines as we push multiple rooms onto the stack. For example, if we push the rooms `[A, B, A]` then as both room `A`s are backed by the same live timeline it is helpful for them both to know the current state.

pixlwave:
After discussion with @Hywan I tested this out and my comments above are entirely irrelevant. We can follow the same 2 branches and nothing changes in terms of listening for the pagination state when live and computing it when detached.

pixlwave (approved):
From an iOS perspective, it works on our side with a single line change 👍

stefanceriu (approved):
Pretty api be pretty 🚀

PR #4585, opened 2025-01-27
chore(ui): Rename `Timeline::subscribe_batched` to `Timeline::subscribe`

Hywan:
This patch renames `Timeline::subscribe_batched` to
`Timeline::subscribe`. Since the `Timeline::subscribe` method has been
removed because unused, it no longer makes sense to have a “batched”
variant here. Let's simplify things!

---

* Follow up of https://github.com/matrix-org/matrix-rust-sdk/pull/4567/.

### DIFF
--- a/crates/matrix-sdk-ui/src/timeline/mod.rs
+++ b/crates/matrix-sdk-ui/src/timeline/mod.rs
@@ -266,11 +266,13 @@ impl Timeline {
         }
     }
 
-    /// Get the current timeline items, and a batched stream of changes.
+    /// Get the current timeline items, along with a stream of updates of
+    /// timeline items.
     ///
-    /// This stream can yield multiple diffs at once. The batching is done such
-    /// that no arbitrary delays are added.
-    pub async fn subscribe_batched(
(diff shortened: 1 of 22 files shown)


crates/matrix-sdk-ui/src/timeline/mod.rs
```diff
@@ -266,11 +266,13 @@ impl Timeline {
...
-    /// Get the current timeline items, and a batched stream of changes.
+    /// Get the current timeline items, along with a stream of updates of
+    /// timeline items.
     ///
-    /// This stream can yield multiple diffs at once. The batching is done such
-    /// that no arbitrary delays are added.
-    pub async fn subscribe_batched(
+    /// The stream produces `Vec<VectorDiff<_>>`, which means multiple updates
+    /// at once. There is no delays, it consumes as much updates as possible and
+    /// batch them.
```
stefanceriu, line 274:
```suggestion
    /// The stream produces `Vec<VectorDiff<_>>`, which means multiple updates
    /// at once. There are no delays, it consumes as many updates as possible and
    /// batches them.
```

stefanceriu (approved):
Noice! 👌

PR #4586, opened 2025-01-27
test: fix cross-signing in legacy dehydrated device test

richvdh:
We missed a call to `sign_device_keys`. Pull out a test helper to make it more obvious where this code is coming from.


richvdh:
Part of me wonders if this should be a test-only method on `DehydratedDevice`, but even then, we'd either have two copies of the code, or we'd be significantly complicating the non-legacy flow.

uhoreg (approved):
lgtm

PR #4589, opened 2025-01-28
feat: Add a method to remove the room topic.

### LINKED ISSUE #4588: [Task] Add a method to remove the room's topic
As part of https://github.com/element-hq/element-meta/issues/2705, problem 1 requires a way to redact the topic instead of setting the string to `""`. Whilst apps could use the timeline to find the topic state event, this is cumbersome and the context in which you edit the room details likely doesn't have a timeline.

Instead it would be great to have a `Room::remove_topic` that finds the event from the store and redacts it directly in the SDK as a complement to `Room:set_topic`.

pixlwave:
This PR adds `Room::remove_room_topic` to allow clients to redact the topic instead of setting it to an empty string.

Closes #4588.

### DIFF
--- a/crates/matrix-sdk/src/room/mod.rs
+++ b/crates/matrix-sdk/src/room/mod.rs
@@ -2270,6 +2270,19 @@ impl Room {
         self.send_state_event(RoomTopicEventContent::new(topic.into())).await
     }
 
+    /// Removes the topic from this room if one is set.
+    pub async fn remove_room_topic(&self) -> Result<()> {
+        let event = self
+            .get_state_event_static::<RoomTopicEventContent>()
+            .await?
+            .ok_or(Error::InsufficientData)?
+            .deserialize()?;
+        let event_id = event.event_id().ok_or(Error::InsufficientData)?;
+
+        self.redact(event_id, None, None).await.map_err(Error::Http)?;
+        Ok(())
+    }
+
     /// Sets the new avatar url for this room.
     ///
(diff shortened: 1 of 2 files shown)


crates/matrix-sdk/src/room/mod.rs
```diff
@@ -2270,6 +2270,19 @@ impl Room {
         self.send_state_event(RoomTopicEventContent::new(topic.into())).await
     }
 
+    /// Removes the topic from this room if one is set.
+    pub async fn remove_room_topic(&self) -> Result<()> {
+        let event = self
```
poljar, line 2275: Could you add a smoke test to check that this calls the correct requests? The `MatrixMockServer` should make this relatively easy.

Let me know if you need some guidance on how the `MatrixMockServer` is supposed to be used.

poljar (changes requested):
Looks good, but a smoke test would make this even better.

crates/matrix-sdk/src/room/mod.rs
pixlwave: Ah nice. Just waiting to hear back if we actually want this method now given https://github.com/element-hq/element-meta/issues/2705#issuecomment-2618827011. Will come back and add that test if we do 👍

pixlwave:
It was decided not to use this (at least for now).

https://github.com/element-hq/element-meta/issues/2705#issuecomment-2619082732

PR #4592, opened 2025-01-28
Introduce an offline mode for the SyncService

poljar:
There are a couple of preparation commits before we come to the meat of this PR, due to this a review commit by commit is advised.

The offline mode attempts to allow users of the UI crate to delegate the restarting of the `SyncService` to the SDK itself. While in offline mode, we're attempting to see if the server is back online using the `/_matrix/client/versions` endpoint. Once we can find the homeserver using the `/versions` endpoint we attempt to sync again.

### DIFF
--- a/crates/matrix-sdk-ui/CHANGELOG.md
+++ b/crates/matrix-sdk-ui/CHANGELOG.md
@@ -27,7 +27,7 @@ All notable changes to this project will be documented in this file.
 - [**breaking**] Add an "offline" mode to the `SyncService`. This allows the
   `SyncService` to attempt to restart the sync automatically. It can be enabled
   with the `SyncServiceBuilder::with_offline_mode` method. Due to this addition,
-  the `SyncService::start` method now returns a `Result` type.
+  the `SyncService::stop` method has been made infallible.
   ([#4592](https://github.com/matrix-org/matrix-rust-sdk/pull/4592))
 
 ### Refactor
--- a/crates/matrix-sdk-ui/CHANGELOG.md
+++ b/crates/matrix-sdk-ui/CHANGELOG.md
@@ -24,6 +24,12 @@ All notable changes to this project will be documented in this file.
   implement `Into<PathBuf>` also implement `Into<AttachmentSource>`.
   ([#4451](https://github.com/matrix-org/matrix-rust-sdk/pull/4451))
 
+- [**breaking**] Add an "offline" mode to the `SyncService`. This allows the
+  `SyncService` to attempt to restart the sync automatically. It can be enabled
+  with the `SyncServiceBuilder::with_offline_mode` method. Due to this addition,
+  the `SyncService::start` method now returns a `Result` type.
+  ([#4592](https://github.com/matrix-org/matrix-rust-sdk/pull/4592))
+
 ### Refactor
 
 - [**breaking**] `Timeline::paginate_forwards` and `Timeline::paginate_backwards`
--- a/crates/matrix-sdk-ui/src/sync_service.rs
+++ b/crates/matrix-sdk-ui/src/sync_service.rs
@@ -448,7 +448,7 @@ impl SyncTaskSupervisor {
         }
     }
 
-    async fn shutdown(self) -> Result<(), Error> {
+    async fn shutdown(self) {
         match self
             .termination_sender
             .send(TerminationReport {
@@ -458,16 +458,21 @@ impl SyncTaskSupervisor {
             })
             .await
         {
-            Ok(_) => self.task.await.map_err(|err| {
-                error!("couldn't finish supervisor task: {err}");
-                Error::InternalSupervisorError
-            }),
+            Ok(_) => {
+                let _ = self.task.await.inspect_err(|err| {
+                    // A `JoinError` indicates that the task was already dead, either because it got
+                    // cancelled or because it panicked. We only cancel the task in the Err branch
+                    // below and the task shouldn't be able to panic.
+                    //
+                    // So let's log an error and return.
+                    error!("The supervisor task has stopped unexpectedly: {err:?}");
+                });
+            }
             Err(err) => {
-                error!("when sending termination report: {err}");
+                error!("Couldn't send the termination report to the supervisor task: {err}");
                 // Let's abort the task if it won't shut down properly, otherwise we would have
                 // left it as a detached task.
                 self.task.abort();
-                Err(Error::InternalSupervisorError)
             }
         }
     }
@@ -504,29 +509,24 @@ impl SyncServiceInner {
         self.state.set(State::Running);
     }
 
-    async fn stop(&mut self) -> Result<(), Error> {
+    async fn stop(&mut self) {
         trace!("pausing sync service");
 
         // Remove the supervisor from our state and request the tasks to be shutdown.
-        let supervisor = self.supervisor.take().ok_or_else(|| {
-            error!("The supervisor was not properly started up");
-            Error::InternalSupervisorError
-        })?;
-
-        supervisor.shutdown().await?;
-
-        Ok(())
+        if let Some(supervisor) = self.supervisor.take() {
+            supervisor.shutdown().await;
+        } else {
+            error!("The sync service was not properly started, the supervisor task doesn't exist");
+        }
     }
 
     async fn restart(
         &mut self,
         room_list_service: Arc<RoomListService>,
         encryption_sync_permit: Arc<AsyncMutex<EncryptionSyncPermit>>,
-    ) -> Result<(), Error> {
-        self.stop().await?;
+    ) {
+        self.stop().await;
         self.start(room_list_service, encryption_sync_permit).await;
-
-        Ok(())
     }
 }
 
@@ -626,7 +626,7 @@ impl SyncService {
     ///   mode and immediately attempt to sync again.
     /// - if the stream has been aborted before, it will be properly cleaned up
     ///   and restarted.
-    pub async fn start(&self) -> Result<(), Error> {
+    pub async fn start(&self) {
         let mut inner = self.inner.lock().await;
 
         // Only (re)start the tasks if it's stopped or if we're in the offline mode.
@@ -637,7 +637,7 @@ impl SyncService {
             State::Offline => {
                 inner
                     .restart(self.room_list_service.clone(), self.encryption_sync_permit.clone())
-                    .await?
+                    .await
             }
             // Otherwise just start.
             State::Idle | State::Terminated | State::Error => {
@@ -646,8 +646,6 @@ impl SyncService {
                     .await
             }
         }
-
-        Ok(())
     }
 
     /// Stop the underlying sliding syncs.
@@ -656,13 +654,13 @@ impl SyncService {
     /// to call this API when the application exits, although not strictly
     /// necessary.
     #[instrument(skip_all)]
-    pub async fn stop(&self) -> Result<(), Error> {
+    pub async fn stop(&self) {
         let mut inner = self.inner.lock().await;
 
         match inner.state.get() {
             State::Idle | State::Terminated | State::Error => {
                 // No need to stop if we were not running.
-                return Ok(());
+                return;
             }
             State::Running | State::Offline => (),
         }
--- a/crates/matrix-sdk-ui/src/sync_service.rs
+++ b/crates/matrix-sdk-ui/src/sync_service.rs
@@ -25,17 +25,13 @@
 use std::sync::Arc;
 
 use eyeball::{SharedObservable, Subscriber};
-use futures_core::Future;
 use futures_util::{pin_mut, StreamExt as _};
 use matrix_sdk::{
     executor::{spawn, JoinHandle},
     Client,
 };
 use thiserror::Error;
-use tokio::sync::{
-    mpsc::{Receiver, Sender},
-    Mutex as AsyncMutex, OwnedMutexGuard,
-};
+use tokio::sync::{mpsc::Sender, Mutex as AsyncMutex, OwnedMutexGuard};
 use tracing::{error, info, instrument, trace, warn, Instrument, Level};
 
 use crate::{
@@ -88,41 +84,38 @@ impl SyncTaskSupervisor {
         room_list_service: Arc<RoomListService>,
         encryption_sync_permit: Arc<AsyncMutex<EncryptionSyncPermit>>,
     ) -> Self {
-        let (sender, receiver) = tokio::sync::mpsc::channel(16);
-        let (room_list_task, encryption_sync_task) = Self::spawn_child_tasks(
-            inner,
-            room_list_service.clone(),
-            encryption_sync_permit,
-            sender.clone(),
-        )
-        .await;
+        let (task, termination_sender) =
+            Self::spawn_supervisor_task(inner, room_list_service, encryption_sync_permit).await;
 
-        let task = spawn(Self::spawn_supervisor_task(
-            inner,
-            room_list_service,
-            room_list_task,
-            encryption_sync_task,
-            receiver,
-        ));
-
-        Self { task, termination_sender: sender }
+        Self { task, termination_sender }
     }
 
     /// The role of the supervisor task is to wait for a termination message
     /// ([`TerminationReport`]), sent either because we wanted to stop both
     /// syncs, or because one of the syncs failed (in which case we'll stop the
     /// other one too).
-    fn spawn_supervisor_task(
+    async fn spawn_supervisor_task(
         inner: &SyncServiceInner,
         room_list_service: Arc<RoomListService>,
-        room_list_task: JoinHandle<()>,
-        encryption_sync_task: JoinHandle<()>,
-        mut receiver: Receiver<TerminationReport>,
-    ) -> impl Future<Output = ()> {
+        encryption_sync_permit: Arc<AsyncMutex<EncryptionSyncPermit>>,
+    ) -> (JoinHandle<()>, Sender<TerminationReport>) {
+        let (sender, mut receiver) = tokio::sync::mpsc::channel(16);
+
         let encryption_sync = inner.encryption_sync_service.clone();
         let state = inner.state.clone();
+        let termination_sender = sender.clone();
 
-        async move {
+        let sync_permit_guard = encryption_sync_permit.clone().lock_owned().await;
+
+        let future = async move {
+            let (room_list_task, encryption_sync_task) = Self::spawn_child_tasks(
+                room_list_service.clone(),
+                encryption_sync.clone(),
+                sync_permit_guard,
+                sender.clone(),
+            )
+            .await;
+
             let report = if let Some(report) = receiver.recv().await {
                 report
             } else {
@@ -179,22 +172,25 @@ impl SyncTaskSupervisor {
                 state.set(State::Terminated);
             }
         }
-        .instrument(tracing::span!(Level::WARN, "supervisor task"))
+        .instrument(tracing::span!(Level::WARN, "supervisor task"));
+
+        let task = spawn(future);
+
+        (task, termination_sender)
     }
 
     async fn spawn_child_tasks(
-        inner: &SyncServiceInner,
         room_list_service: Arc<RoomListService>,
-        encryption_sync_permit: Arc<AsyncMutex<EncryptionSyncPermit>>,
+        encryption_sync_service: Arc<EncryptionSyncService>,
+        sync_permit_guard: OwnedMutexGuard<EncryptionSyncPermit>,
         sender: Sender<TerminationReport>,
     ) -> (JoinHandle<()>, JoinHandle<()>) {
         // First, take care of the room list.
         let room_list_task = spawn(Self::room_list_sync_task(room_list_service, sender.clone()));
 
         // Then, take care of the encryption sync.
-        let sync_permit_guard = encryption_sync_permit.clone().lock_owned().await;
         let encryption_sync_task = spawn(Self::encryption_sync_task(
-            inner.encryption_sync_service.clone(),
+            encryption_sync_service,
             sender.clone(),
             sync_permit_guard,
         ));
--- a/crates/matrix-sdk-ui/src/sync_service.rs
+++ b/crates/matrix-sdk-ui/src/sync_service.rs
@@ -334,6 +334,34 @@ struct SyncServiceInner {
     supervisor: Option<SyncTaskSupervisor>,
 }
 
+impl SyncServiceInner {
+    async fn start(
+        &mut self,
+        room_list_service: Arc<RoomListService>,
+        encryption_sync_permit: Arc<AsyncMutex<EncryptionSyncPermit>>,
+    ) {
+        trace!("starting sync service");
+
+        self.supervisor =
+            Some(SyncTaskSupervisor::new(self, room_list_service, encryption_sync_permit).await);
+        self.state.set(State::Running);
+    }
+
+    async fn stop(&mut self) -> Result<(), Error> {
+        trace!("pausing sync service");
+
+        // Remove the supervisor from our state and request the tasks to be shutdown.
+        let supervisor = self.supervisor.take().ok_or_else(|| {
+            error!("The supervisor was not properly started up");
+            Error::InternalSupervisorError
+        })?;
+
+        supervisor.shutdown().await?;
+
+        Ok(())
+    }
+}
+
 /// A high level manager for your Matrix syncing needs.
 ///
 /// The [`SyncService`] is responsible for managing real-time synchronization
@@ -432,17 +460,9 @@ impl SyncService {
             // If we're already running, there's nothing to do.
             State::Running => (),
             State::Idle | State::Terminated | State::Error => {
-                trace!("starting sync service");
-
-                inner.supervisor = Some(
-                    SyncTaskSupervisor::new(
-                        &inner,
-                        self.room_list_service.clone(),
-                        self.encryption_sync_permit.clone(),
-                    )
-                    .await,
-                );
-                inner.state.set(State::Running);
+                inner
+                    .start(self.room_list_service.clone(), self.encryption_sync_permit.clone())
+                    .await
             }
         }
     }
@@ -464,19 +484,7 @@ impl SyncService {
             State::Running => (),
         }
 
-        trace!("pausing sync service");
-
-        // First, request to stop the two underlying syncs; we'll look at the results
-        // later, so that we're in a clean state independently of the request to stop.
-
-        // Remove the supervisor from our inner state and request the tasks to be
-        // shutdown.
-        let supervisor = inner.supervisor.take().ok_or_else(|| {
-            error!("The supervisor was not properly started up");
-            Error::InternalSupervisorError
-        })?;
-
-        supervisor.shutdown().await
+        inner.stop().await
     }
 
     /// Attempt to get a permit to use an `EncryptionSyncService` at a given
--- a/crates/matrix-sdk-ui/src/sync_service.rs
+++ b/crates/matrix-sdk-ui/src/sync_service.rs
@@ -20,7 +20,9 @@
 //!
 //! The sync service will signal errors via its [`state`](SyncService::state)
 //! that the user MUST observe. Whenever an error/termination is observed, the
-//! user MUST call [`SyncService::start()`] again to restart the room list sync.
+//! user should call [`SyncService::start()`] again to restart the room list
+//! sync, if that is not desirable, the offline support for the [`SyncService`]
+//! may be enabled using the [`SyncServiceBuilder::with_offline_mode`] setting.
 
 use std::{sync::Arc, time::Duration};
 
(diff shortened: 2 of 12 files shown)


crates/matrix-sdk-ui/CHANGELOG.md
```diff
@@ -24,6 +24,12 @@ All notable changes to this project will be documented in this file.
   implement `Into<PathBuf>` also implement `Into<AttachmentSource>`.
   ([#4451](https://github.com/matrix-org/matrix-rust-sdk/pull/4451))
 
+- [**breaking**] Add an "offline" mode to the `SyncService`. This allows the
+  `SyncService` to attempt to restart the sync automatically. It can be enabled
+  with the `SyncServiceBuilder::with_offline_mode` method. Due to this addition,
+  the `SyncService::start` method now returns a `Result` type.
```
stefanceriu, line 30: We touched offline on whether this is actually useful/actionable to the end client and we agreed that eventual (rare) errors can be handled internally here and the start/stop methods simplified.

poljar: The `SyncService::stop()` method has been made infallible in https://github.com/matrix-org/matrix-rust-sdk/pull/4592/commits/bd77e273ecd8e6745db12828d083cd26f8181705.

Due to this, the `SyncService::start()` has become infallible as well.

crates/matrix-sdk-ui/src/sync_service.rs
```diff
@@ -90,6 +111,58 @@ impl SyncTaskSupervisor {
...
+                // services.
+                let report =
+                    receiver.recv().await.unwrap_or_else(TerminationReport::supervisor_error);
+
+                match report.origin {
+                    TerminationOrigin::EncryptionSync | TerminationOrigin::RoomList => (),
+                    // Since the sync service aren't running anymore, we can only receive a report
+                    // from the supervisor. It would have probably made sense to have separate
+                    // channels for reports the sync services send and the user can send using the
+                    // `SyncService::stop()` method.
```
andybalaam, line 152: Should we panic in the unexpected case then? Or log an error?

crates/matrix-sdk-ui/src/sync_service.rs
```diff
@@ -90,6 +111,58 @@ impl SyncTaskSupervisor {
...
+                    TerminationOrigin::Supervisor => break report,
+                }
+            }
+        };
+
+        let wait_to_be_online = async move {
+            loop {
```
andybalaam, line 159: Is it OK to do this repeatedly with no pauses or backoff?

crates/matrix-sdk-ui/tests/integration/sync_service.rs
```diff
@@ -202,3 +205,123 @@ async fn test_sync_service_state() -> anyhow::Result<()> {
...
+
+    let next_state = timeout(states.next(), Duration::from_millis(500))
+        .await
+        .expect("We should have changed states by now")
+        .unwrap();
+
+    assert_eq!(next_state, State::Offline, "We should have entered the offline mode");
+
+    sync_service
+        .start()
```
andybalaam: `.restart()`?

andybalaam: `.restart()`?

andybalaam (approved):
Looks great; a couple of comments.

crates/matrix-sdk-ui/src/sync_service.rs
poljar: Yes, the backoff is handled by the network requesting layer.

crates/matrix-sdk-ui/tests/integration/sync_service.rs
poljar: There is no `SyncService::restart()` method, the `SyncServiceInner` struct has such a method which is called by the `SyncService::start()` method if we're in the offline state.

poljar: Same as https://github.com/matrix-org/matrix-rust-sdk/pull/4592#discussion_r1933674330.

crates/matrix-sdk-ui/src/sync_service.rs
poljar: No, I think it's completely normal that this will happen.

I.e. if we can't reach the server both sync services will get an error and send a termination report.

The supervisor will only handle one of those and stop everything. The second termination report will be left in the channel.

This was fine since we would drop the channel once the supervisor was stopped, but now we don't stop the supervisor and the channel contains the stale termination report.

I think it's fine to just ignore it.

crates/matrix-sdk-ui/src/sync_service.rs
Hywan: Can we explain that in a comment please?

Hywan (approved):
Okay so it looks good to me. I wish you could expand a bit about the “periodic"/back-off requests to `/versions`: is the interval between 2 requests going to be increase _every time_, or is it reaching a plateau at some point? I think this is an important information: if the `SyncService` enters its offline mode for a looong time, and the back-off algorithm takes 5mn between each request, the user experience might feel pretty broken.

crates/matrix-sdk-ui/src/sync_service.rs
poljar: I guess we can add some timeout in case the user configures the `RequestConfig` very aggressively.

I can add a comment, yes. It is our usual `RequestConfig` which uses exponential backoff: https://matrix-org.github.io/matrix-rust-sdk/matrix_sdk/config/struct.RequestConfig.html.

bmarty:
Testing this code using Element X Android, and when I go offline (airplane mode), the sync state moves from `Running` to `Error`. I would expect to get the state `Offline` in this case. When restoring the network condition, the sync does not restart by itself.

 Did I miss anything @poljar ?

poljar:
> Testing this code using Element X Android, and when I go offline (airplane mode), the sync state moves from `Running` to `Error`. I would expect to get the state `Offline` in this case. When restoring the network condition, the sync does not restart by itself.
> 
> Did I miss anything @poljar ?

Yes, you didn't enable "offline" mode. You need to enable it with the `SyncServiceBuilder::with_offline_mode()` function. (You couldn't have enabled it because I forgot to expose this via the FFI, exposed in https://github.com/matrix-org/matrix-rust-sdk/pull/4592/commits/449f2febcb52abf7036f48ef034a164990a218c6).

bmarty:
Thanks. /me build the SDK with the latest commit.

Will the Sending queues be restarted automatically by the SDK when the "network" is back?

We can keep our existing code (https://github.com/element-hq/element-x-android/blob/develop/appnav/src/main/kotlin/io/element/android/appnav/loggedin/SendQueues.kt#L45) if it's not the case.

crates/matrix-sdk-ui/src/sync_service.rs
poljar: Added a timeout, just in case, and a comment in https://github.com/matrix-org/matrix-rust-sdk/pull/4592/commits/7ebdf2ebfb7044f44a9f3291f50ee4e3ceba9240.

poljar:
> Thanks. /me builds the SDK with the latest commit.
> 
> Will the Sending queues be restarted automatically by the SDK when the "network" is back?
> 
> We can keep our existing code (https://github.com/element-hq/element-x-android/blob/develop/appnav/src/main/kotlin/io/element/android/appnav/loggedin/SendQueues.kt#L45) if it's not the case.

This does not interact with the send queue, so you need to continue on doing what you did.

bmarty:
OK, thanks. It's working quite well, the SDK automatically start the sync when the network is back. But I observe some weird behavior, when the network is back, the state goes to `Running` but sometimes, without touching anything, it quickly moves to `Offline` then `Running`. I can share logs if it helps.

Also when the network is back, I can observe that `RoomListServiceSyncIndicatorListener.onUpdate()` is invoked with `SHOW` for a long time (maybe the time for a sync request to timeout?). This is maybe not related to this PR though.

bmarty:
Here is a gif of the problem, you can see the state change while nothing is done on the app / the network condition (once the log is highlighted you need to wait about 30s to see the pb, sorry).

SyncState

poljar:
> Here is a gif of the problem, you can see the state change while nothing is done on the app / the network condition (once the log is highlighted you need to wait about 30s to see the pb, sorry).
> 
> SyncState [ SyncState ](https://private-user-images.githubusercontent.com/3940906/407797229-7f3f103b-9e46-4e30-9483-12b978e870c1.gif?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MzgxNjY3NzAsIm5iZiI6MTczODE2NjQ3MCwicGF0aCI6Ii8zOTQwOTA2LzQwNzc5NzIyOS03ZjNmMTAzYi05ZTQ2LTRlMzAtOTQ4My0xMmI5NzhlODcwYzEuZ2lmP1gtQW16LUFsZ29yaXRobT1BV1M0LUhNQUMtU0hBMjU2JlgtQW16LUNyZWRlbnRpYWw9QUtJQVZDT0RZTFNBNTNQUUs0WkElMkYyMDI1MDEyOSUyRnVzLWVhc3QtMSUyRnMzJTJGYXdzNF9yZXF1ZXN0JlgtQW16LURhdGU9MjAyNTAxMjlUMTYwMTEwWiZYLUFtei1FeHBpcmVzPTMwMCZYLUFtei1TaWduYXR1cmU9MjY0YzM1NmEyOGE2MWEwNDk2OGUyYzZiZGY0MjExY2ZjNmRlMzc5ZjI0MzEwZTUzM2RkMTQ5ZDcxMjJkYzEzNCZYLUFtei1TaWduZWRIZWFkZXJzPWhvc3QifQ.W9qYc6KTDqQPXlCR82vRuLEq3mqRm39wQ2G1_RzXBnw) [ ](https://private-user-images.githubusercontent.com/3940906/407797229-7f3f103b-9e46-4e30-9483-12b978e870c1.gif?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MzgxNjY3NzAsIm5iZiI6MTczODE2NjQ3MCwicGF0aCI6Ii8zOTQwOTA2LzQwNzc5NzIyOS03ZjNmMTAzYi05ZTQ2LTRlMzAtOTQ4My0xMmI5NzhlODcwYzEuZ2lmP1gtQW16LUFsZ29yaXRobT1BV1M0LUhNQUMtU0hBMjU2JlgtQW16LUNyZWRlbnRpYWw9QUtJQVZDT0RZTFNBNTNQUUs0WkElMkYyMDI1MDEyOSUyRnVzLWVhc3QtMSUyRnMzJTJGYXdzNF9yZXF1ZXN0JlgtQW16LURhdGU9MjAyNTAxMjlUMTYwMTEwWiZYLUFtei1FeHBpcmVzPTMwMCZYLUFtei1TaWduYXR1cmU9MjY0YzM1NmEyOGE2MWEwNDk2OGUyYzZiZGY0MjExY2ZjNmRlMzc5ZjI0MzEwZTUzM2RkMTQ5ZDcxMjJkYzEzNCZYLUFtei1TaWduZWRIZWFkZXJzPWhvc3QifQ.W9qYc6KTDqQPXlCR82vRuLEq3mqRm39wQ2G1_RzXBnw)

This looks like the `/versions` succeeds but for some reason the sync right after that fails, the 30s timeout sounds like it could be the normal `/sync` timeout. Do you have logs for this where we can see what exactly is going on?

bmarty:
Here are the logs, let's me know if you need anything else.

[logcat.tar.gz](https://github.com/user-attachments/files/18591098/logcat.tar.gz)

poljar:
> Here are the logs, let me know if you need anything else.
> 
> [logcat.tar.gz](https://github.com/user-attachments/files/18591098/logcat.tar.gz)

So what is happening is the following:

1. You enter offline mode
2. Once a `/versions` requests succeeds you exit offline mode and attempt to sync.
3. The syncs succeed and you are running
4. Then something fails with subsequent syncs
5. You enter offline mode again

Step 4 has this logs:

> 01-29 17:41:09.085 20795 20838 D org.matrix.rust.sdk: matrix_sdk::http_client: Error while sending request: Reqwest(reqwest::Error { kind: Request, url: Url { scheme: "https", cannot_be_a_base: false, username: "", password: None, host: Some(Domain("matrix-client.matrix.org")), port: None, path: "/_matrix/client/unstable/org.matrix.simplified_msc3575/sync", query: Some("pos=0%2Fm5606054066%7E2.5606054070%7E36.5606054070%7E3.5606054070%7E37.5606054070_757284974_35967510_m3389410864%7E38.3389410890_3731314317_263869577_1444040328_11077857250_0_422464&timeout=30000"), fragment: None }, source: hyper_util::client::legacy::Error(SendRequest, hyper::Error(Io, Kind(ConnectionAborted))) }) | crates/matrix-sdk/src/http_client/mod.rs:225 | spans: next_sync_with_lock > sync_once{conn_id="encryption" pos="0/m5606054066~2.5606054070~36.5606054070~3.5606054070~37.5606054070_757284974_35967510_m3389410864~38.3389410890_3731314317_263869577_1444040328_11077857250_0_422464"} > send{server_versions=[V1_0, V1_1, V1_2, V1_3, V1_4, V1_5, V1_6, V1_7, V1_8, V1_9, V1_10, V1_11] config=RequestConfig { timeout: 60s } request_id="REQ-2788" method=POST uri="https://matrix-client.matrix.org/_matrix/client/unstable/org.matrix.simplified_msc3575/sync" request_size="157 B"}

The important bit is this:

> hyper_util::client::legacy::Error(SendRequest, hyper::Error(Io, Kind(ConnectionAborted)))

The [docs](https://doc.rust-lang.org/std/io/enum.ErrorKind.html#variant.ConnectionAborted) for this error tell us the following:

> The connection was aborted (terminated) by the remote server.

Unless the docs are lying and we're actually calling `SyncService::start()` or `SyncService::stop()` right before that error, I don't  think there's anything we can do here.

poljar:
The logs have shown that some failures weren't using our exponential backoff for the `/versions` request, I have enabled this now in https://github.com/matrix-org/matrix-rust-sdk/pull/4592/commits/359e25d2b18fdf2d4d173d5f71bd6d53ad4b56f0

poljar:
https://github.com/matrix-org/matrix-rust-sdk/pull/4592/commits/91aa37be2eceb5956f24570612db84abfea1c81a enables exponential backoff and retrying for the `/sync` requests.

This should avoid the glitch we saw where we went into `Offline` mode right after we exited it due to a `/sync` request failing a short time after we exited `Offline` mode.

crates/matrix-sdk-ui/src/sync_service.rs
```diff
@@ -88,41 +84,38 @@ impl SyncTaskSupervisor {
...
+        let termination_sender = sender.clone();
+
+        let sync_permit_guard = encryption_sync_permit.clone().lock_owned().await;
+
+        let future = async move {
+            let (room_list_task, encryption_sync_task) = Self::spawn_child_tasks(
+                room_list_service.clone(),
+                encryption_sync.clone(),
+                sync_permit_guard,
+                sender.clone(),
```
bnjbvr: This `clone()` might be spurious, since it looks like it's the last use of `sender`.

crates/matrix-sdk-ui/src/sync_service.rs
```diff
@@ -322,6 +430,7 @@ impl SyncTaskSupervisor {
 
 struct SyncServiceInner {
     encryption_sync_service: Arc<EncryptionSyncService>,
+    with_offline_mode: bool,
```
bnjbvr, line 470: Can you add a tiny doc comment explaining what it is? (notably that it's static for the sync service's lifetime, it won't change value over time)

crates/matrix-sdk-ui/src/sync_service.rs
```diff
@@ -481,7 +614,7 @@ impl SyncService {
                 // No need to stop if we were not running.
                 return Ok(());
             }
-            State::Running => (),
+            State::Running | State::Offline => (),
```
bnjbvr, line 649: preexisting/nit: let's avoid using `()` and instead use `{}` to materialize an empty statement

crates/matrix-sdk-ui/src/sync_service.rs
```diff
@@ -360,6 +469,17 @@ impl SyncServiceInner {
 
         Ok(())
     }
+
+    async fn restart(
```
bnjbvr, line 507: Since this is called only once and the body is trivial, might be worth inlining? (unless it's used in more places in subsequent commits)

crates/matrix-sdk-ui/src/sync_service.rs
```diff
@@ -90,6 +111,58 @@ impl SyncTaskSupervisor {
...
+                // we might miss a report with the
+                // `TerminationOrigin::Supervisor` origin and the channel might contain stale
+                // reports from one of the sync services, in case both of them have sent a
+                // report, let's ignore all reports we receive from the sync
+                // services.
+                let report =
+                    receiver.recv().await.unwrap_or_else(TerminationReport::supervisor_error);
+
+                match report.origin {
+                    TerminationOrigin::EncryptionSync | TerminationOrigin::RoomList => (),
```
bnjbvr, line 150: nit: `{}` instead of `()`

crates/matrix-sdk-ui/src/sync_service.rs
```diff
@@ -105,71 +178,106 @@ impl SyncTaskSupervisor {
...
+                        let client = room_list_service.client();
+
+                        if let Some(report) = Self::offline_check(client, &mut receiver).await {
+                            if report.is_error {
+                                state.set(State::Error);
+                                break;
+                            } else {
+                                state.set(State::Idle);
+                                break;
+                            }
```
bnjbvr, line 298: I think it reads slightly better if you put the `break` outside the if and else's bodies…

crates/matrix-sdk-ui/src/sync_service.rs
```diff
@@ -105,71 +178,106 @@ impl SyncTaskSupervisor {
...
+                        } else {
+                            state.set(State::Running);
+                        }
```
bnjbvr, line 301: … then notably, this `else` can be removed because it would be an else-after-break. (Only the `else` disappears, not its content `state.set(State::Running);`

crates/matrix-sdk-ui/src/sync_service.rs
```diff
@@ -105,71 +178,106 @@ impl SyncTaskSupervisor {
...
+                    } else {
+                        state.set(State::Error);
+                        break;
+                    }
+                } else if matches!(report.origin, TerminationOrigin::Supervisor) {
+                    state.set(State::Idle);
+                    break;
+                } else {
```
bnjbvr, line 309: nit: else after break. We can get rid of the `else` here.

crates/matrix-sdk-ui/src/sync_service.rs
```diff
@@ -90,6 +111,58 @@ impl SyncTaskSupervisor {
         Self { task, termination_sender }
     }
 
+    async fn offline_check(
```
bnjbvr, line 132: Since this is not trivial, can you add a small doc comment explaining what it is?

crates/matrix-sdk-ui/tests/integration/sync_service.rs
```diff
@@ -202,3 +205,123 @@ async fn test_sync_service_state() -> anyhow::Result<()> {
...
+            .unwrap();
+
+        assert_eq!(next_state, State::Offline, "We should have entered the offline mode");
+    }
+
+    mock_server.mock_versions().ok().expect(1..).mount().await;
+
+    let next_state = timeout(states.next(), Duration::from_millis(500))
+        .await
+        .expect("We should changed the state")
```
bnjbvr, line 241: typo here

crates/matrix-sdk-ui/tests/integration/sync_service.rs
```diff
@@ -202,3 +205,123 @@ async fn test_sync_service_state() -> anyhow::Result<()> {
...
+
+        sync_service.start().await.expect("We should be able to start the sync service");
+        assert_next_eq!(states, State::Running);
+
+        let next_state = timeout(states.next(), Duration::from_millis(500))
+            .await
+            .expect("We should have changed states by now")
+            .unwrap();
+
+        assert_eq!(next_state, State::Offline, "We should have entered the offline mode");
```
bnjbvr, line 234: we do have the rather verbosely named `assert_next_with_timeout` helper that combines both checks, fwiw.

crates/matrix-sdk-ui/src/sync_service.rs
```diff
@@ -429,16 +429,21 @@ impl SyncTaskSupervisor {
...
         {
-            Ok(_) => self.task.await.map_err(|err| {
-                error!("couldn't finish supervisor task: {err}");
-                Error::InternalSupervisorError
-            }),
+            Ok(_) => {
+                let _ = self.task.await.inspect_err(|err| {
+                    // A `JoinError` indicates that the task was already dead, either because it got
+                    // cancelled or because it panicked. We only cancel the task in the Err branch
+                    // bellow and the task shouldn't be able to panic.
```
bnjbvr, line 452: nit: below

crates/matrix-sdk-ui/src/sync_service.rs
```diff
@@ -472,29 +477,24 @@ impl SyncServiceInner {
...
-            Error::InternalSupervisorError
-        })?;
-
-        supervisor.shutdown().await?;
-
-        Ok(())
+        if let Some(supervisor) = self.supervisor.take() {
+            supervisor.shutdown().await;
+        } else {
+            warn!("The sync service was already stopped");
```
bnjbvr, line 503: It's rather than the state is invalid, because the `self.inner.state` was set to some value that made us think it's running, but the absence of `self.supervisor` shows it's not. It would be more helpful if the warn message indicated this is invalid state, in case we saw it in rageshakes — the current message here suggests this is benign (trying to stop while it's already stopped), despite the `WARN` level.

crates/matrix-sdk-ui/src/sync_service.rs
```diff
@@ -157,9 +158,22 @@ impl SyncTaskSupervisor {
...
+                // or any server error in the 5xx range of HTTP errors are
+                // considered to be transient.
+                //
+                // This means that we're not going to tightloop here, and in the case the
+                // `RequestConfig` has been set up to not have a limit, ever hit the second
+                // iteration of this loop.
+                //
+                // No matter, as a precaution, we're going to sleep for a while here in the
+                // Error case, the user might have configured the RequestConfig
+                // to not contain any backoff or retries.
```
bnjbvr: Doesn't it defeat the purpose of exponential backoff, though? The first retries would happen after less than 500ms, with exponential backoff, so we're artificially delaying the restart here.

crates/matrix-sdk-ui/src/sync_service.rs
```diff
@@ -158,22 +159,23 @@ impl SyncTaskSupervisor {
...
-                // No matter, as a precaution, we're going to sleep for a while here in the
-                // Error case, the user might have configured the RequestConfig
-                // to not contain any backoff or retries.
-                match client.fetch_server_capabilities().await {
+                // Still, as a precaution, we're going to sleep here for a while in the Error
+                // case.
+                match client.fetch_server_capabilities(Some(request_config)).await {
                     Ok(_) => break,
-                    Err(_) => sleep(Duration::from_millis(500)).await,
+                    Err(_) => sleep(Duration::from_millis(100)).await,
```
bnjbvr, line 178: Ok, now with the explicit `RequestConfig`, we don't need the `sleep` for sure, right?

bnjbvr (approved):
Thanks, lgtm. It's nice that we can get rid of the errors when starting/stopping the sync service. The `select` in the `offline_check` function is not quite tested but it's also quite hard to test. Mostly cosmetic requests, so ship it!

poljar:
> Thanks, lgtm. It's nice that we can get rid of the errors when starting/stopping the sync service. The `select` in the `offline_check` function is not quite tested but it's also quite hard to test. Mostly cosmetic requests, so ship it!

What do you mean by not quite tested? What else is missing? Happy to add more tests, but Codecov tells us that both branches are hit in tests:

image

crates/matrix-sdk-ui/src/sync_service.rs
poljar: It's used inside of a loop so it's necessary, though you might be looking at a commit before the loop was introduced.

crates/matrix-sdk-ui/src/sync_service.rs
Hywan: Thanks <3!

crates/matrix-sdk-ui/src/sync_service.rs
poljar: Sure: 0d99447daa2653cdcea16f9ef3fb548ecd5905c7.

crates/matrix-sdk-ui/src/sync_service.rs
poljar: 5df0f2a3f3f5d3b78b65de6f738a03582ca11a31

crates/matrix-sdk-ui/src/sync_service.rs
poljar: I prefer the semantics of a `Offline -> restart()` call instead of having to read that a `stop()` in combination of a `start()` is a restart.

I'll let the compiler inline this.

crates/matrix-sdk-ui/src/sync_service.rs
poljar: 5df0f2a3f3f5d3b78b65de6f738a03582ca11a31

crates/matrix-sdk-ui/src/sync_service.rs
poljar: 0bdd53b45ab34c9b000475450899e0e989be1edf

crates/matrix-sdk-ui/src/sync_service.rs
poljar: 0bdd53b45ab34c9b000475450899e0e989be1edf

crates/matrix-sdk-ui/src/sync_service.rs
poljar: This one doesn't work, because not every branch has a break or continue, which means that this will be executed every time if we remove the else making the loop always execute only once.

crates/matrix-sdk-ui/tests/integration/sync_service.rs
poljar: 997d8212eb8169f3c09b857621556d3c44f1355c

crates/matrix-sdk-ui/tests/integration/sync_service.rs
poljar: Ah I was searching for something like this, though that's still not quite ideal so I added a new helper: 0010d37a55f686014926f997f48bf6167e530f20.

crates/matrix-sdk-ui/src/sync_service.rs
poljar: 997d8212eb8169f3c09b857621556d3c44f1355c

crates/matrix-sdk-ui/src/sync_service.rs
poljar: I attempted to clarify this a bit further and converted this to an `error` in 1e99bf9ca770d8554747afe8d2d823b3f42fc906.

crates/matrix-sdk-ui/src/sync_service.rs
poljar: There are still cases where the backoff won't kick in, so as a precaution I'll leave this in.

crates/matrix-sdk-ui/src/sync_service.rs
poljar: Same argument as https://github.com/matrix-org/matrix-rust-sdk/pull/4592#discussion_r1936844342.

crates/matrix-sdk-ui/src/sync_service.rs
poljar: Sure: a685c463a67271b3654e24eb5648f3214926e598.

poljar:
Rebased and force pushed to get rid of the fixup commits.

maan2003:
with offline mode, should I still handle Terminate and Error states by restarting? or those irrecoverable errors now?

poljar:
> with offline mode, should I still handle Terminate and Error states by restarting? or those irrecoverable errors now?

Realistically I don't think `Terminated` could have happened previously either. The `Error` one can't most likely happen if `offline` mode is enabled, but you can continue on to handle them as you did previously.

PR #4594, opened 2025-01-29
feat(ui): Implement `Timeline` lazy pagination

Hywan:
<figure role="presentation">
<p align="center"></p>
<figcaption><p align="center">A red fox, relaxing under the snow (credits <a href="https://www.roeselienraimond.com/red_foxes/">roeselienraimond.com</a>)</p></figcaption>
</figure>

---

These patches implement lazy pagination for a live `Timeline`. The idea is that `Timeline::subscribe()` keeps returning a `Stream` but [the `Skip` higher-order stream](https://github.com/jplatte/eyeball/pull/72) is applied to skip the first items, exactly the `count` first items:

- When new items are added in the `Timeline`, they will appear in the subscriber's stream (if they are not part of the skipped items, see `Skip`'s documentation to learn more).
- When the `Timeline` paginates backwards, the `Skip`'s `count` value decreases, revealing more items if there is such. When the `count` reaches 0, it fallbacks to the event cache pagination.

For the moment, the event cache reads **all** events from its storage and sends them to the `Timeline`. A follow up of this pull request is to prevent the event cache to read all events, but to read events chunk by chunk. Thus, the `Timeline` do a lazy pagination: when `count` reaches zero, it fallbacks to the event cache, which will load another chunk if possible, otherwise it will fallback to the network. See the documentation in this pull request to learn more. This is a general overview of the mechanism.

---

First off, the first patches do a bit of clean up (removing clones, move and rename `TimelineStream` into a new `subscriber` module).

After that, patches create a `SkipCount` type, and use it inside the `Timeline::subscribe()` flow.

Finally, the last patches add integration tests. Unit tests are already written.

---

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,9 +33,9 @@ as_variant = "1.2.0"
 base64 = "0.22.1"
 byteorder = "1.5.0"
 chrono = "0.4.39"
-eyeball = { version = "0.8.8", features = ["tracing"] }
-eyeball-im = { version = "0.6.0", features = ["tracing"] }
-eyeball-im-util = "0.8.0"
+eyeball = { git = "https://github.com/Hywan/eyeball", branch = "feat-im-util-skip", features = ["tracing"] }
+eyeball-im = { git = "https://github.com/Hywan/eyeball", branch = "feat-im-util-skip", features = ["tracing"] }
+eyeball-im-util = { git = "https://github.com/Hywan/eyeball", branch = "feat-im-util-skip" }
 futures-core = "0.3.31"
 futures-executor = "0.3.31"
 futures-util = "0.3.31"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,9 +33,9 @@ as_variant = "1.2.0"
 base64 = "0.22.1"
 byteorder = "1.5.0"
 chrono = "0.4.39"
-eyeball = { git = "https://github.com/Hywan/eyeball", branch = "feat-im-util-skip", features = ["tracing"] }
-eyeball-im = { git = "https://github.com/Hywan/eyeball", branch = "feat-im-util-skip", features = ["tracing"] }
-eyeball-im-util = { git = "https://github.com/Hywan/eyeball", branch = "feat-im-util-skip" }
+eyeball = { git = "https://github.com/jplatte/eyeball", branch = "main", features = ["tracing"] }
+eyeball-im = { git = "https://github.com/jplatte/eyeball", branch = "main", features = ["tracing"] }
+eyeball-im-util = { git = "https://github.com/jplatte/eyeball", branch = "main" }
 futures-core = "0.3.31"
 futures-executor = "0.3.31"
 futures-util = "0.3.31"
--- a/crates/matrix-sdk-ui/src/timeline/controller/metadata.rs
+++ b/crates/matrix-sdk-ui/src/timeline/controller/metadata.rs
@@ -339,11 +339,11 @@ pub(in crate::timeline) struct EventMeta {
     /// +-------+-------------------+----------------------+
     /// | 0     | content of `$ev0` |                      |
     /// | 1     | content of `$ev2` | reaction with `$ev4` |
-    /// | 2     | day divider       |                      |
+    /// | 2     | date divider      |                      |
     /// | 3     | content of `$ev3` |                      |
     /// | 4     | content of `$ev5` |                      |
     ///
-    /// Note the day divider that is a virtual item. Also note `$ev4` which is
+    /// Note the date divider that is a virtual item. Also note `$ev4` which is
     /// a reaction to `$ev2`. Finally note that `$ev1` is not rendered in
     /// the timeline.
     ///
--- a/crates/matrix-sdk-ui/src/timeline/controller/metadata.rs
+++ b/crates/matrix-sdk-ui/src/timeline/controller/metadata.rs
@@ -20,7 +20,8 @@ use tracing::trace;
 
 use super::{
     super::{
-        reactions::Reactions, rfind_event_by_id, TimelineItem, TimelineItemKind, TimelineUniqueId,
+        reactions::Reactions, rfind_event_by_id, subscriber::skip::SkipCount, TimelineItem,
+        TimelineItemKind, TimelineUniqueId,
     },
     read_receipts::ReadReceipts,
     state::PendingPollEvents,
@@ -36,6 +37,8 @@ pub(in crate::timeline) struct TimelineMetadata {
     ///
     /// This value is constant over the lifetime of the metadata.
     internal_id_prefix: Option<String>,
+
+    pub(super) subscriber_skip_count: SkipCount,
 
     /// The hook to call whenever we run into a unable-to-decrypt event.
     ///
@@ -107,6 +110,7 @@ impl TimelineMetadata {
         is_room_encrypted: Option<bool>,
     ) -> Self {
         Self {
+            subscriber_skip_count: SkipCount::new(),
             own_user_id,
             next_internal_id: Default::default(),
             reactions: Default::default(),
--- a/crates/matrix-sdk-ui/src/timeline/controller/metadata.rs
+++ b/crates/matrix-sdk-ui/src/timeline/controller/metadata.rs
@@ -38,6 +38,8 @@ pub(in crate::timeline) struct TimelineMetadata {
     /// This value is constant over the lifetime of the metadata.
     internal_id_prefix: Option<String>,
 
+    /// The `count` value for the `Skip higher-order stream used by the
+    /// `TimelineSubscriber`. See its documentation to learn more.
     pub(super) subscriber_skip_count: SkipCount,
 
     /// The hook to call whenever we run into a unable-to-decrypt event.
--- a/crates/matrix-sdk-ui/src/timeline/pagination.rs
+++ b/crates/matrix-sdk-ui/src/timeline/pagination.rs
@@ -32,8 +32,21 @@ impl super::Timeline {
     ///
     /// Returns whether we hit the start of the timeline.
     #[instrument(skip_all, fields(room_id = ?self.room().room_id()))]
-    pub async fn paginate_backwards(&self, num_events: u16) -> Result<bool, Error> {
+    pub async fn paginate_backwards(&self, mut num_events: u16) -> Result<bool, Error> {
         if self.controller.is_live().await {
+            match self.controller.live_lazy_paginate_backwards(num_events).await {
+                Some(needed_num_events) => {
+                    num_events = needed_num_events.try_into().expect(
+                        "failed to cast `needed_num_events` (`usize`) into `num_events` (`usize`)",
+                    );
+                }
+                None => {
+                    // TODO: returning `false` is not true everytime, we need a way to know if
+                    // lazy-loading has reached the end of the timeline.
+                    return Ok(false);
+                }
+            }
+
             Ok(self.live_paginate_backwards(num_events).await?)
         } else {
             Ok(self.controller.focused_paginate_backwards(num_events).await?)
--- a/crates/matrix-sdk-ui/src/timeline/subscriber.rs
+++ b/crates/matrix-sdk-ui/src/timeline/subscriber.rs
@@ -1,0 +1,49 @@
+// Copyright 2025 The Matrix.org Foundation C.I.C.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use std::{
+    pin::Pin,
+    sync::Arc,
+    task::{Context, Poll},
+};
+
+use futures_core::Stream;
+use pin_project_lite::pin_project;
+
+use super::TimelineDropHandle;
+
+pin_project! {
+    pub(super) struct TimelineStream<S> {
+        #[pin]
+        inner: S,
+        drop_handle: Arc<TimelineDropHandle>,
+    }
+}
+
+impl<S> TimelineStream<S> {
+    pub fn new(inner: S, drop_handle: Arc<TimelineDropHandle>) -> Self {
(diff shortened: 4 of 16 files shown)


crates/matrix-sdk-ui/src/timeline/pagination.rs
```diff
@@ -32,8 +32,21 @@ impl super::Timeline {
     ///
     /// Returns whether we hit the start of the timeline.
     #[instrument(skip_all, fields(room_id = ?self.room().room_id()))]
-    pub async fn paginate_backwards(&self, num_events: u16) -> Result<bool, Error> {
+    pub async fn paginate_backwards(&self, mut num_events: u16) -> Result<bool, Error> {
         if self.controller.is_live().await {
+            match self.controller.live_lazy_paginate_backwards(num_events).await {
+                Some(needed_num_events) => {
```
stefanceriu, line 38: Would it maybe be better to have `live_lazy_paginate_backwards` return the right type?

Cargo.toml
```diff
@@ -33,9 +33,9 @@ as_variant = "1.2.0"
 base64 = "0.22.1"
 byteorder = "1.5.0"
 chrono = "0.4.39"
-eyeball = { version = "0.8.8", features = ["tracing"] }
-eyeball-im = { version = "0.6.0", features = ["tracing"] }
-eyeball-im-util = "0.8.0"
+eyeball = { git = "https://github.com/Hywan/eyeball", branch = "feat-im-util-skip", features = ["tracing"] }
+eyeball-im = { git = "https://github.com/Hywan/eyeball", branch = "feat-im-util-skip", features = ["tracing"] }
+eyeball-im-util = { git = "https://github.com/Hywan/eyeball", branch = "feat-im-util-skip" }
```
stefanceriu, line 38: We should probably work on these patches from a matrix-org organisation fork as opposed to your personal one.

crates/matrix-sdk-ui/src/timeline/subscriber.rs
```diff
@@ -50,3 +50,344 @@ where
...
+    use super::super::controller::TimelineFocusKind;
+
+    const MAXIMUM_NUMBER_OF_INITIAL_ITEMS: usize = 20;
+
+    #[derive(Clone)]
+    pub struct SkipCount {
+        count: SharedObservable<usize>,
+    }
+
+    impl SkipCount {
```
stefanceriu, line 127: The comments on this file are just 👌 !!

crates/matrix-sdk-ui/src/timeline/subscriber.rs
```diff
@@ -0,0 +1,459 @@
...
+                // |------------------------|--------|--------|--------------|
+                // | 60                     | 20     | 40     | 20 items     |
+                // | 10                     | 20     | 0      | 10 items     |
+                // | 0                      | 20     | 0      | 0 item       |
+                //
+                next_number_of_items.saturating_sub(MAXIMUM_NUMBER_OF_INITIAL_ITEMS)
+            }
+            // Not the initial state: there are items.
+            else {
+                // There is less items than before. Shift to the left `count` by the difference
```
stefanceriu, line 161:
```suggestion
                // There are less items than before. Shift to the left `count` by the difference
```

stefanceriu (approved):
lgtm and I confirm that it's working properly on element x ios.

crates/matrix-sdk-ui/src/timeline/pagination.rs
Hywan: What do you mean by the _right type_?

Cargo.toml
Hywan: Yes! [My patch on eyeball has been merged](https://github.com/jplatte/eyeball/pull/72/), and I'm about to make a release of `eyeball-im-util`, which would remove the need for this change in our `Cargo.toml` file.

crates/matrix-sdk-ui/src/timeline/subscriber.rs
Hywan: ❤️

crates/matrix-sdk-ui/src/timeline/subscriber.rs
Hywan: I thought we should use _there is_ when the quantity is unknown, and _there are_ when the quantity is known. Anyway. Happy to fix that! Thanks.

stefanceriu: Don't know about that, just know that `items` is plural and doesn't work with `is` ¯\_(ツ)_/¯

crates/matrix-sdk-ui/tests/integration/timeline/pagination.rs
```diff
@@ -842,3 +849,246 @@ async fn test_back_pagination_aborted() {
...
+    assert_timeline_stream! {
+        [timeline_stream]
+        // Huh?
+        // TODO: understand why
+        clear;
+        clear;
+
+        // `$ev11`, the 1st item, and its update.
+        append "$ev11";
+        update[0] "$ev11";
```
bnjbvr, line 898: Why do we have an update? (I suspect this might be the read receipt being updated, but worth double-checking?)

crates/matrix-sdk-ui/tests/integration/timeline/pagination.rs
```diff
@@ -842,3 +849,246 @@ async fn test_back_pagination_aborted() {
...
+        //
+        // This is the 20th item, hurray!
+        prepend "$ev10";
+    };
+
+    // Nothing more!
+    assert_pending!(timeline_stream);
+
+    let no_network_pagination =
+        mock_server.mock_room_messages().error500().never().mount_as_scoped().await;
```
bnjbvr, line 967: nit: you do not need to define it: if the SDK tried to reach it, it would then run into a 404, equivalently. (And then you wouldn't need the `drop(no_network_pagination)` below either!)

crates/matrix-sdk-ui/tests/integration/timeline/pagination.rs
```diff
@@ -842,3 +849,246 @@ async fn test_back_pagination_aborted() {
...
+
+        // Receive 3 new items.
+        //
+        // They are inserted after the date divider, hence the indices 1, 2 and 3..
+        assert_timeline_stream! {
+            [timeline_stream]
+            insert[1] "$ev201";
+            insert[2] "$ev200";
+        };
+        // So cool.
```
bnjbvr, line 1089: It is a nice test!

crates/matrix-sdk-ui/src/timeline/controller/metadata.rs
```diff
@@ -38,6 +38,8 @@ pub(in crate::timeline) struct TimelineMetadata {
     /// This value is constant over the lifetime of the metadata.
     internal_id_prefix: Option<String>,
 
+    /// The `count` value for the `Skip higher-order stream used by the
+    /// `TimelineSubscriber`. See its documentation to learn more.
```
bnjbvr, line 42: Can you add brackets around `TimelineSubscriber` so this links to it?

crates/matrix-sdk-ui/tests/integration/timeline/mod.rs
```diff
@@ -831,8 +832,71 @@ async fn test_timeline_without_encryption_can_update() {
...
+    let client = mock_server.client_builder().build().await;
+
+    mock_server.sync_joined_room(&client, room_id).await;
+
+    let event_cache = client.event_cache();
+    event_cache.subscribe().unwrap();
+
+    // The event cache contains 30 events.
+    event_cache
+        .add_initial_events(
```
bnjbvr, line 850: We shouldn't be using `add_initial_events` in testing, since this is going to be removed at some point. Can you use a sync instead, or if it's necessary that the event cache already contains the events, consider using an in-memory event cache store and feed it linked chunk updates that would create the events?

crates/matrix-sdk-ui/tests/integration/timeline/mod.rs
```diff
@@ -831,8 +832,71 @@ async fn test_timeline_without_encryption_can_update() {
...
+    let (timeline_initial_items, mut timeline_stream) = timeline.subscribe().await;
+
+    // The timeline receives 20 initial values, not 30!
+    assert_eq!(timeline_initial_items.len(), 20);
+    assert_pending!(timeline_stream);
+
+    // To get the other, the timeline needs to paginate.
+
+    let _no_network_pagination =
+        mock_server.mock_room_messages().error500().never().mount_as_scoped().await;
```
bnjbvr, line 878: Again, you can remove it, because if the SDK attempts to do pagination, it'll hit a 404 and fail.

bnjbvr (approved):
Cool, thanks for the extensive tests around the skip count. Thanks for implementing this, it will be a sweet improvement over what we have right now and will make the storage more usable in the real world :heart_hands: 

I think it might be interesting to write a test showing the known limitations, esp. with respect to gaps introduced *after* the skip count, as we discussed privately. (For instance, a situation where the room cache contains [e1, e2, gap, e3], the timeline shows [e2, e3] because of a skip count=1, but it will have to get back to e1 before filling the gap after e2).

crates/matrix-sdk-ui/src/timeline/subscriber.rs
```diff
@@ -24,20 +24,23 @@ use pin_project_lite::pin_project;
 use super::TimelineDropHandle;
 
 pin_project! {
-    pub(super) struct TimelineStream<S> {
+    /// A stream that wraps a [`TimelineDropHandle`] so that the `Timeline`
+    /// isn't dropped until the `Stream` is dropped.
+    pub(super) struct WithTimelineDropHandle<S> {
```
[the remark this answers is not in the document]
Hywan: I meant: it's a stream with a `TimelineDropHandle`. But I'm fine with the renaming, it's different but alright too.

crates/matrix-sdk-ui/src/timeline/subscriber.rs
```diff
@@ -50,3 +50,344 @@ where
...
+        ///
+        /// [`Skip`]: eyeball_im_util::vector::Skip
+        pub fn compute_next(
+            &self,
+            previous_number_of_items: usize,
+            next_number_of_items: usize,
+        ) -> usize {
+            let current_count = self.count.get();
+
+            // Initial states: no items is present.
```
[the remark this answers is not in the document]
Hywan: Ah yeah, I miss this every time…

crates/matrix-sdk-ui/src/timeline/subscriber.rs
```diff
@@ -0,0 +1,49 @@
+// Copyright 2022 The Matrix.org Foundation C.I.C.
```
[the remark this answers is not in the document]
Hywan: 🪄

crates/matrix-sdk-ui/src/timeline/subscriber.rs
```diff
@@ -50,3 +50,344 @@ where
...
+            //
+            if current_count >= page_size {
+                (current_count - page_size, None)
+            } else {
+                (0, Some(page_size - current_count))
+            }
+        }
+
+        /// Compute the `count` value for [the `Skip` higher-order
+        /// stream][`Skip`] when a backwards pagination is happening.
```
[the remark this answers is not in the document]
Hywan: Eagle eye 🍒!

crates/matrix-sdk-ui/src/timeline/subscriber.rs
```diff
@@ -50,3 +50,344 @@ where
...
+                         // other kind of timelines will use it, we would need it, it's better to have
+                         // this in case of; everything is tested, the logic is made more robust.
+        pub fn compute_next_when_paginating_forwards(&self, _page_size: usize) -> usize {
+            // Nothing to do, the count remains unchanged as we skip the first values, not
+            // the last values; paginating forwards will add items at the end, not at the
+            // start of the timeline.
+            self.count.get()
+        }
+
+        /// Subscribe to update of the count value.
```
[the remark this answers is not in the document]
Hywan: :+1:

crates/matrix-sdk-ui/tests/integration/timeline/pagination.rs
```diff
@@ -842,3 +849,246 @@ async fn test_back_pagination_aborted() {
...
+        })
+        .await;
+
+    // Only 20 items are broadcasted to the subscriber.
+    assert_timeline_stream! {
+        [timeline_stream]
+        // Huh?
+        // TODO: understand why
+        clear;
+        clear;
```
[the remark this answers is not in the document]
Hywan: Ho ho.

I actually understand the problem. I need to remove `RoomEventCacheUpdate::Clear`.

crates/matrix-sdk-ui/tests/integration/timeline/pagination.rs
Hywan: It's the read receipt, yup.

crates/matrix-sdk-ui/tests/integration/timeline/pagination.rs
Hywan: Oh yeah, sure…

crates/matrix-sdk-ui/tests/integration/timeline/pagination.rs
Hywan: It is!

crates/matrix-sdk-ui/src/timeline/controller/metadata.rs
Hywan: It's a private type and intra-links are disabled for private types, so it won't work (I've tried and got crazy about this 😋).

crates/matrix-sdk-ui/tests/integration/timeline/mod.rs
Hywan: This is necessary the event cache already contains the events. I'll try by feeding with linked chunk updates.

Hywan:
> I think it might be interesting to write a test showing the known limitations, esp. with respect to gaps introduced after the skip count, as we discussed privately. (For instance, a situation where the room cache contains [e1, e2, gap, e3], the timeline shows [e2, e3] because of a skip count=1, but it will have to get back to e1 before filling the gap after e2).

This problem exists with or without the `Skip`'s higher-order stream. I propose to write a test to show the bug with a proper fix, because it _is_ a bug, in a follow up patch.

crates/matrix-sdk-ui/tests/integration/timeline/pagination.rs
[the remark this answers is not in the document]
Hywan: Follow up https://github.com/matrix-org/matrix-rust-sdk/pull/4627/

PR #4601, opened 2025-02-01
fix(ui): Fix performance of `TimelineEventHandler::deduplicate_local_timeline_item`

Hywan:
This patch drastically improves the performance of `TimelineEventHandler::deduplicate_local_timeline_item`.

Before this patch, `rfind_event_item` was used to iterate over all timeline items: for each item in reverse order, if it was an event timeline item, and if it was a local event timeline item, and if it was matching the event ID or transaction ID, then a duplicate was found.

The problem is the following: all items are traversed.

However, local event timeline items are always at the back of the items. Even virtual timeline items are before local event timeline items. Thus, it is not necessary to traverse all items. It is possible to stop the iteration as soon as (i) a non event timeline item is met, or (ii) a non local event timeline item is met.

This patch updates
`TimelineEventHandler::deduplicate_local_timeline_item` to replace to use of `rfind_event_item` by a custom iterator that stops as soon as a non event timeline item, or a non local event timeline item, is met, or —of course— when a local event timeline item is a duplicate.

To do so, [`Iterator::try_fold`] is probably the best companion. [`Iterator::try_find`] would have been nice, but it is available on nightlies, not on stable versions of Rust. However, many methods in `Iterator` are using `try_fold`, like `find` or any other methods that need to do a “short-circuit”. Anyway, `try_fold` works pretty nice here, and does exactly what we need.

Our use of `try_fold` is to return a `ControlFlow<Option<(usize, TimelineItem)>, ()>`. After `try_fold`, we call
`ControlFlow::break_value`, which returns an `Option`. Hence the need to call `Option::flatten` at the end to get a single `Option` instead of having an `Option<Option<(usize, TimelineItem)>>`.

I'm testing this patch with the `test_lazy_back_pagination` test in https://github.com/matrix-org/matrix-rust-sdk/pull/4594. With 10_000 events in the sync, the test was taking 13s to run on my machine. With this patch, it takes 10s to run. It's a 23% improvement. This `deduplicate_local_timeline_item` method was taking a large part of the computation according to the profiler. With this patch, this method is barely visible in the profiler it is so small.

Flamegraph of the `TimelineEventHandler::add_item` method (that calls `deduplicate_local_timeline_item`), before this patch:

Screenshot 2025-02-01 at 11 59 37

And after:

Screenshot 2025-02-01 at 11 26 08

---

[`Iterator::try_fold`]: https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.try_fold
[`Iterator::try_find`]: https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.try_find

### DIFF
--- a/crates/matrix-sdk-ui/src/timeline/event_handler.rs
+++ b/crates/matrix-sdk-ui/src/timeline/event_handler.rs
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use std::sync::Arc;
+use std::{ops::ControlFlow, sync::Arc};
 
 use as_variant::as_variant;
 use indexmap::IndexMap;
@@ -51,7 +51,7 @@ use ruma::{
 use tracing::{debug, error, field::debug, info, instrument, trace, warn};
 
 use super::{
-    algorithms::{rfind_event_by_id, rfind_event_item},
+    algorithms::rfind_event_by_id,
     controller::{
         ObservableItemsTransaction, ObservableItemsTransactionEntry, PendingEdit, PendingEditKind,
         TimelineMetadata, TimelineStateTransaction,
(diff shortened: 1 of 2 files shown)


crates/matrix-sdk-ui/src/timeline/event_handler.rs
```diff
@@ -1261,18 +1261,48 @@ impl<'a, 'o> TimelineEventHandler<'a, 'o> {
...
+            // when a non local event timeline item is met. We want to stop iterating when:
+            //
+            // - a duplicate local event timeline item has been found,
+            // - a non local event timeline item is met,
+            // - a non event timeline is met.
+            //
+            // Indeed, it is a waste of time to iterate over all items in `items`. Local event
+            // timeline items are necessarily at the end of `items`: as soon as they have been
+            // iterated, we can stop the entire iteration.
+            .try_fold((), |(), (nth, timeline_item)| {
```
bnjbvr, line 1281: I think you can use a `find_map()` and not have to bump the minimum rust version, and it would be a bit more noob-friendly :)

bnjbvr (approved):
You don't need the rust-version bump if you use `find_map()`, otherwise lgtm.

crates/matrix-sdk-ui/src/timeline/event_handler.rs
Hywan: [`Iterator::find_map`](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.find_map) is not appropriate here. It returns the first non-`None` value. It means that we cannot differentiate the _break the loop because we've found a value_, vs. _break the loop because no value can be found_.

We need to stop the iteration when all local event timeline items have been traversed. This is not possible to represent that with `Iterator::find_map`. That's exactly the problem this patch is trying to fix :-).

bnjbvr: Why not a plain loop then… :innocent:

Hywan: That, that's a good question ;-).

PR #4602, opened 2025-02-01
sqlite: Optimize the DB and the files sizes

zecakeh:
- First commit: run `VACUUM` on the `SqliteStateStore`. We should have done this when we moved the media cache to `SqlitEventCacheStore`, to defragment the file and reduce its size on the filesystem. Because of that there are databases from that time that still take much more space than they need.
- Second commit: run `PRAGMA optimize` after running the eventual migrations. The [SQLite docs](https://www.sqlite.org/pragma.html#pragma_optimize) recommend to do this regularly, especially after a schema change. Doing this consistently during construction seems to be the easiest solution.
- Third commit: limit the size of the WAL file. While the DB connections of the databases are open, [the size of the WAL file can keep increasing](https://www.sqlite.org/wal.html#avoiding_excessively_large_wal_files) depending on the size needed for the transactions. A critical case is `VACUUM` which basically writes the content of the DB file to the WAL file before writing it back to the DB file, so we end up taking twice the size of the database. The docs say that the WAL file is written to the database when it reaches 4MB, so 10MB seems like a good limit that shouldn't truncate the file too often.

### DIFF
--- a/crates/matrix-sdk-sqlite/CHANGELOG.md
+++ b/crates/matrix-sdk-sqlite/CHANGELOG.md
@@ -12,6 +12,12 @@ All notable changes to this project will be documented in this file.
   `EventCacheStore` for `MediaRetentionPolicy`. See the changelog of
   `matrix-sdk-base` for more details.
   ([#4571](https://github.com/matrix-org/matrix-rust-sdk/pull/4571))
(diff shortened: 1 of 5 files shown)


poljar (approved):
Makes sense.

Thanks for the comprehensive links to the SQLite documentation.

PR #4605, opened 2025-02-03
feat(ffi): Add FFI bindings to `Room::forget`

jmartinesp:
This PR adds bindings for `Room::forget` and makes sure you can also forget rooms with `Banned` state.

Signed-off-by:

### DIFF
--- a/crates/matrix-sdk/src/room/mod.rs
+++ b/crates/matrix-sdk/src/room/mod.rs
@@ -2896,11 +2896,14 @@ impl Room {
     ///
     /// This communicates to the homeserver that it should forget the room.
     ///
-    /// Only left rooms can be forgotten.
+    /// Only left or banned-from rooms can be forgotten.
     pub async fn forget(&self) -> Result<()> {
         let state = self.state();
-        if state != RoomState::Left {
-            return Err(Error::WrongRoomState(WrongRoomState::new("Left", state)));
+        match state {
+            RoomState::Joined | RoomState::Invited | RoomState::Knocked => {
+                return Err(Error::WrongRoomState(WrongRoomState::new("Left / Banned", state)));
+            }
+            RoomState::Left | RoomState::Banned => {}
         }
 
         let request = forget_room::v3::Request::new(self.inner.room_id().to_owned());
--- a/crates/matrix-sdk/tests/integration/room/left.rs
+++ b/crates/matrix-sdk/tests/integration/room/left.rs
@@ -2,7 +2,7 @@ use std::time::Duration;
 
 use assert_matches2::assert_matches;
 use matrix_sdk::config::SyncSettings;
-use matrix_sdk_base::RoomState;
+use matrix_sdk_base::{RoomInfoNotableUpdateReasons, RoomState};
 use matrix_sdk_test::{
     async_test, test_json, GlobalAccountDataTestEvent, LeftRoomBuilder, SyncResponseBuilder,
     DEFAULT_TEST_ROOM_ID,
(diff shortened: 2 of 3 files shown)


crates/matrix-sdk/src/client/mod.rs
```diff
@@ -2471,6 +2471,12 @@ impl Client {
         let base_room = self.inner.base_client.room_knocked(&response.room_id).await?;
         Ok(Room::new(self.clone(), base_room))
     }
+
+    /// Remove the room from the store, removing all data related to it.
+    pub async fn forget_room(&self, room_id: &RoomId) -> Result<()> {
+        self.inner.base_client.forget_room(room_id).await?;
+        Ok(())
+    }
```
bnjbvr, line 2488: There's `Room::forget()`, and it would be easy to mistake one for the other. If you plan to keep this one, now you know there's `Room::forget()`: can you precise in the doc comment this does NOT forget the room at the server level, but only locally? (and as such, the room can still be in the DM list, etc.)

crates/matrix-sdk/src/client/mod.rs
```diff
@@ -3182,4 +3188,19 @@ pub(crate) mod tests {
             .await;
         assert_matches!(ret, Ok(()));
     }
+
+    #[async_test]
+    async fn test_forget_room() {
+        let server = MatrixMockServer::new().await;
+        let client = server.client_builder().build().await;
```
bnjbvr, line 3215: nit: you don't need the server here, you can create the client immediately

bindings/matrix-sdk-ffi/src/client.rs
```diff
@@ -1018,6 +1018,17 @@ impl Client {
         Ok(Arc::new(Room::new(room)))
     }
 
+    /// Leave a room by its room ID.
+    pub async fn leave_room_by_id(
```
bnjbvr, line 1022: What's the purpose of this function, when you have `Room::leave()`? In other words, what are the conditions that make it necessary: a room that's out of sync shouldn't be present at all, and having "bad state locally" seems like a poor justification for API proliferation.

crates/matrix-sdk/src/client/mod.rs
jmartinesp: Oh, I didn't realise this was already there. I'll remove it and add bindings to `Room::forget` instead.

bindings/matrix-sdk-ffi/src/client.rs
jmartinesp: I think you're right and we should focus on fixing the underlying issue in `RoomPreview` instead.

jmartinesp:
I removed the `leave_room_by_id` function and moved the forget room bindings so they match the existing API, including the API call to the `/forget` endpoint.

crates/matrix-sdk/src/room/mod.rs
```diff
@@ -2896,11 +2896,11 @@ impl Room {
     ///
     /// This communicates to the homeserver that it should forget the room.
     ///
-    /// Only left rooms can be forgotten.
+    /// Only left or banned-from rooms can be forgotten.
     pub async fn forget(&self) -> Result<()> {
         let state = self.state();
-        if state != RoomState::Left {
-            return Err(Error::WrongRoomState(WrongRoomState::new("Left", state)));
+        if matches!(state, RoomState::Left | RoomState::Banned).not() {
```
bnjbvr, line 2902: Can you use an explicit `match` here? Next time we add a `RoomState` variant, the compiler will complain that we need to handle it, which is nicer than having to remember to handle it ourselves.

crates/matrix-sdk/tests/integration/room/left.rs
```diff
@@ -77,6 +77,67 @@ async fn test_forget_non_direct_room() {
...
+async fn test_forget_banned_room() {
+    let (client, server) = logged_in_client_with_server().await;
+    let user_id = client.user_id().unwrap();
+
+    let event_cache = client.event_cache();
+    event_cache.subscribe().unwrap();
+    event_cache.enable_storage().unwrap();
+
+    Mock::given(method("POST"))
+        .and(path_regex(r"^/_matrix/client/r0/rooms/.*/forget$"))
```
bnjbvr, line 90: Bonus points if you add new endpoints to the `MatrixMockServer` :eyes:

bnjbvr (approved):
Thanks!

PR #4609, opened 2025-02-03
fix(room_preview): Return room preview info based on local data for banned rooms

jmartinesp:
What the tin says.

We weren't trusting the local room info for rooms that weren't in joined state because for invites and knocks that might mean  displaying outdated info about the room, and left rooms would also return the same info than we had locally. 

However, this missed the `Banned` room state, and the different endpoints would just return a `M_FORBIDDEN` 403 error when trying to get info about a room the user was banned from. So the fix is to trust the local info we have about the room instead of asking the HS about it.

Signed-off-by:

### DIFF
--- a/crates/matrix-sdk/src/client/mod.rs
+++ b/crates/matrix-sdk/src/client/mod.rs
@@ -3186,4 +3186,114 @@ pub(crate) mod tests {
             .await;
         assert_matches!(ret, Ok(()));
     }
+
+    #[async_test]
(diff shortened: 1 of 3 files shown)


Hywan (commented):
Thanks! Looks good to me. Can we have a test to check it works as expected 😇?

jmartinesp:
@Hywan there are now tests for every room state.

Hywan (approved):
Looks good, thanks!

PR #4610, opened 2025-02-03
feat: Upgrade Ruma to 0.12.1

zecakeh:


bnjbvr (approved):
Fine by me, is there a changelog with the new Ruma changes? The release on Github didn't include any meaningful information…

zecakeh:
Oh we should probably consider that a bug in our release task. There is nothing meaningful for the ruma crate in itself but you can see the changelogs of other crates: https://github.com/ruma/ruma/releases.

Note that this is basically the same commit as was used previously. @poljar asked us to make a release because one is planned for the SDK this week.

Hywan (approved):

PR #4611, opened 2025-02-03
fix(timeline): don't lose the `latest_edit_json` under some circumstances

bnjbvr:
Based on our using of `EventTimelineItem::with_content`, we'd sometimes override the `latest_edit_json` field to `None`, because we didn't have a value to provide. The intent was rather to *not* touch the `latest_edit_json` in such cases, since we were only using `None` because we didn't have a JSON value to provide in the first place. We lost the `latest_edit_json` that way when adding a poll response, or ending a poll, or when updating a message because its replied-to message has been updated (because of an edit, because it could be decrypted, etc.).

This adds a new method `with_content_and_latest_edit` that's used in the few places where we actually may want to store a JSON edit, fixing the issue observed in the other cases.


jmartinesp (approved):
LGTM, thanks!

PR #4614, opened 2025-02-03
fix[oidc]: fix docstring in oidc module

JoFrost:
While running `cargo test --doc --features docsrs` with the OIDC feature enabled, I noticed that the tests in the OIDC module were failing due to outdated module resolution in the documentation examples.

This MR updates the module resolution to reflect the current structure.


bnjbvr (approved):
Good catch, thank you! There's a bit more, but our CI didn't run it; I'll address that as a followup. Thanks!

PR #4616, opened 2025-02-04
fix(sdk): Improve performance of `RoomEvents::maybe_apply_new_redaction`

Hywan:
This patch improves the performance of
`RoomEvents::maybe_apply_new_redaction`. This method deserialises all the events it receives, entirely. If the event is not an `m.room.redaction`, then the method returns early. Most of the time, the event is deserialised for nothing because most events aren't of kind `m.room.redaction`!

This patch first uses `Raw::get_field("type")` to detect the type of the event. If it's a `m.room.redaction`, then the event is entirely deserialised, otherwise the method returns.

When running the `test_lazy_back_pagination` from
https://github.com/matrix-org/matrix-rust-sdk/pull/4594 with 10'000 events, prior to this patch, this method takes 11% of the execution time. With this patch, this method takes 2.5%.

Call tree before this patch:

Screenshot 2025-02-04 at 09 11 52

After this patch:

Screenshot 2025-02-04 at 09 15 20

---

### DIFF
--- a/crates/matrix-sdk/src/event_cache/room/events.rs
+++ b/crates/matrix-sdk/src/event_cache/room/events.rs
@@ -22,7 +22,7 @@ use matrix_sdk_common::linked_chunk::{
     ObservableUpdates, Position,
 };
 use ruma::{
-    events::{room::redaction::SyncRoomRedactionEvent, AnySyncTimelineEvent},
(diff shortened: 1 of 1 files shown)


jmartinesp (approved):
LGTM!

bnjbvr:
Nice. This calls for a deserialization cache, because we keep deserializing events all over the place…

Hywan:
Indeed. I was working on this but stopped due to lack of time, but it's definitely something that may improve the performance.

PR #4618, opened 2025-02-04
feat(ffi): add support for starting and responding to user verification requests

stefanceriu:
This patch expands on the existing device verification flows to add support for verifying users as well. 

It does so by:
- moving incoming request detection to the Client level so that it works for both to-device events as well as timeline ones. The client then forwards the data through the generic `process_incoming_verification_request` method
- exposing a new `request_user_verification` method next to the (renamed) `request_device_verification` one on the `SessionVerificationController`
- consolidating subscription logic within `set_ongoing_verification_request`
- overall the mecanisms are very similar which should make it easy to adopt in the final clients

### DIFF
--- a/bindings/matrix-sdk-ffi/src/client.rs
+++ b/bindings/matrix-sdk-ffi/src/client.rs
@@ -32,8 +32,11 @@ use matrix_sdk::{
             user_directory::search_users,
         },
         events::{
-            room::{avatar::RoomAvatarEventContent, encryption::RoomEncryptionEventContent},
-            AnyInitialStateEvent, AnyToDeviceEvent, InitialStateEvent,
+            room::{
+                avatar::RoomAvatarEventContent, encryption::RoomEncryptionEventContent,
+                message::MessageType,
+            },
+            AnyInitialStateEvent, InitialStateEvent,
         },
         serde::Raw,
         EventEncryptionAlgorithm, RoomId, TransactionId, UInt, UserId,
@@ -53,10 +56,12 @@ use ruma::{
     },
     events::{
         ignored_user_list::IgnoredUserListEventContent,
+        key::verification::request::ToDeviceKeyVerificationRequestEvent,
         room::{
             join_rules::{
                 AllowRule as RumaAllowRule, JoinRule as RumaJoinRule, RoomJoinRulesEventContent,
             },
+            message::OriginalSyncRoomMessageEvent,
             power_levels::RoomPowerLevelsEventContent,
         },
         GlobalAccountDataEventType,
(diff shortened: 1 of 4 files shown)


bindings/matrix-sdk-ffi/src/client.rs
```diff
@@ -204,12 +208,27 @@ impl Client {
             tokio::sync::RwLock<Option<SessionVerificationController>>,
         > = Default::default();
         let controller = session_verification_controller.clone();
+        sdk_client.add_event_handler(move |event: AnyToDeviceEvent| async move {
```
poljar, line 211: Since we only care about `m.key.verification.request` events in this handler, perhaps it makes sense to listen to only those types of events.

Like we do here: https://github.com/matrix-org/matrix-rust-sdk/blob/4c46e42201da5d7b1871b05764a35930d578dba6/examples/emoji_verification/src/main.rs#L126-L136.

bindings/matrix-sdk-ffi/src/client.rs
```diff
@@ -204,12 +208,27 @@ impl Client {
...
+            }
+        });
 
-        sdk_client.add_event_handler(move |ev: AnyToDeviceEvent| async move {
-            if let Some(session_verification_controller) = &*controller.clone().read().await {
-                session_verification_controller.process_to_device_message(ev).await;
-            } else {
-                debug!("received to-device message, but verification controller isn't ready");
+        let controller = session_verification_controller.clone();
+        sdk_client.add_event_handler(move |event: OriginalSyncRoomMessageEvent| async move {
```
poljar, line 225: Before you ask, sadly I don't think we can make the handler here smarter like we did for the to-device variant.

bindings/matrix-sdk-ffi/src/room.rs
```diff
@@ -252,13 +252,13 @@ impl Room {
     }
 
     pub async fn member(&self, user_id: String) -> Result<RoomMember, ClientError> {
-        let user_id = UserId::parse(&*user_id).context("Invalid user id.")?;
+        let user_id = UserId::parse(&*user_id)?;
```
poljar, line 255: Is there a point in removing these? Perhaps it would make sense to have a helper that always applies this context to the `UserId` parsing step instead of just removing this?

poljar (approved):
Looks good. I left a couple of nits and it's quite sad that we can't, instead of adding event handlers here, have a higher level stream in the SDK which gives you the `Verification` object once one gets created.

Maybe one of these days when the `OlmMachine` doesn't get permanently killed and resurrected. :smiling_face_with_tear:

bindings/matrix-sdk-ffi/src/client.rs
stefanceriu: Good point, will do 👍

bindings/matrix-sdk-ffi/src/client.rs
stefanceriu: Ha, thank you for not letting me fumble around for nothing 😅

PR #4620, opened 2025-02-04
ci: add a new feature set to test experimental-oidc too

bnjbvr:
This would help find test failures specific to experimental-oidc, as well as doctests failing (which would have prevented the failures fixed in https://github.com/matrix-org/matrix-rust-sdk/pull/4614 to happen in the first place).

---

Let me know if it is preferred to reuse an existing feature set, rather than introduce one; I don't have any strong opinion there.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -35,6 +35,7 @@ jobs:
       fail-fast: true
       matrix:
         name:
+          - experimental-oidc
           - no-encryption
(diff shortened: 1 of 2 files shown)


bnjbvr:
> On the other hand, shouldn't we actually be pushing towards making OIDC non-experimental? We've been running it for a while now and I think we've worked out all the kinks.

Good point, we should discuss this soonish.

PR #4629, opened 2025-02-05
fix(recovery): Delete the known secrets from 4s when disabling recovery

poljar:


crates/matrix-sdk/src/encryption/recovery/mod.rs
```diff
@@ -527,6 +541,20 @@ impl Recovery {
         Ok(())
     }
 
+    async fn delete_all_known_secrets(&self) -> Result<()> {
+        for secret_name in Self::KNOWN_SECRETS {
+            let event_type = GlobalAccountDataEventType::from(secret_name.to_owned());
+            let content = SecretEventContent::new(Default::default());
+            let secret_content = Raw::from_json(
+                to_raw_value(&content)
+                    .expect("We should be able to serialize a raw empty secret event content"),
```
bnjbvr, line 550: Any chance we could use a plain `?` here, since this function returns a `Result` type anyways?

bnjbvr (approved):
TIL there's no endpoint to remove global account data, and one can only reset them to default values :woozy_face: 

Anyhow, thanks, the patch looks good.

crates/matrix-sdk/src/encryption/recovery/mod.rs
poljar: As the recovery module has a specialized error type we can't just use `?` to convert the error. We either need to add a error variant for this or muck it somewhere into the `matrix_sdk::Error` type.

But I don't think it makes much sense, serializing the default event content should never fail.

bnjbvr: Makes sense to keep it like this, then :+1:

poljar:
Oh my god, you can't upload the empty content? Good thing I made an integration test.

davidegirardi (commented):
Sounds reasonable. Do we need to delete (or overwrite with empty JSONs) the room keys too?

crates/matrix-sdk/src/encryption/recovery/mod.rs
```diff
@@ -288,8 +300,10 @@ impl Recovery {
         // Why oh why, can't we delete account data events?
         self.client.account().set_account_data(SecretStorageDisabledContent {}).await?;
         self.client.account().set_account_data(BackupDisabledContent { disabled: true }).await?;
+        // Delete all the known secrets we have in the account data.
+        self.delete_all_known_secrets().await?;
```
BillCarsonFr, line 304: Just before that we deleted `m.secret_storage.default_key`. Here we delete the secrets.
We should probably also delete the key account data itself? the one with type `m.secret_storage.key.ID-OF-THE-DEFAULT-KEY`

BillCarsonFr (approved):
We might want also to delete the actual key

crates/matrix-sdk/src/encryption/recovery/mod.rs
poljar: Is there a point of deleting this? We can't remove the event as there isn't a `DELETE` API for this, so we replace sensitive things with an empty JSON content `{}`.

But the `m.secret_storage.key.ID-OF-THE-DEFAULT-KEY` events, do not contain sensitive data. They only contain information about the key which can be used, once you have the key, to check that it indeed is correct.

crates/matrix-sdk/src/encryption/recovery/mod.rs
```diff
@@ -288,8 +300,10 @@ impl Recovery {
         // Why oh why, can't we delete account data events?
```
richvdh, line 300: ... because nobody was enthusiastic enough about https://github.com/matrix-org/matrix-spec-proposals/pull/3391, ftr

crates/matrix-sdk/src/encryption/recovery/mod.rs
richvdh: > But the `m.secret_storage.key.ID-OF-THE-DEFAULT-KEY` events, do not contain sensitive data. They only contain information about the key which can be used, once you have the key, to check that it indeed is correct.

this is fair. I guess it might be nice just to clean it up a bit as best we can?

richvdh: ok, but you've done it anyway?

crates/matrix-sdk/src/encryption/recovery/mod.rs
```diff
@@ -285,11 +297,36 @@ impl Recovery {
...
         self.client.encryption().backups().disable().await?;
+
         // Why oh why, can't we delete account data events?
+        //
+        // Alright, let's attempt to "delete" the content of our current default key,
+        // for this we first need to check if there is a default key, then
+        // deserialize the content and find out the key ID.
+        //
+        // Then we finally set the event to an empty JSON content.
+        if let Ok(Some(default_event)) =
```
richvdh, line 308: As an aside, I think that referring to account data as an "event" which can be "deleted" or changed in any way, is a confusing anti-pattern. One of the defining properties of an "event" in most environments is that it's something that happened in the past, so changing it is nonsense.

(I know that the spec uses that term, but I'd like that to change (https://github.com/matrix-org/matrix-spec/issues/2030), and we can avoid spreading the mess any further in the meantime.)

crates/matrix-sdk/src/encryption/recovery/mod.rs
poljar: Yes, sorry. I attempted to make a joke.

crates/matrix-sdk/src/encryption/recovery/mod.rs
poljar: So, what should we call them?

richvdh: "Account data". Or possibly, "Account data entries".

PR #4632, opened 2025-02-05
feat(sdk): Implement `EventCache` lazy-loading

Hywan:
WIP


bnjbvr:
Thanks for the detailed summary. Could it be incorporated as part of internal documentation, somehow?

> That's why users were seeing missing messages in their Timeline! And it was impossible to recover, except by clearing the cache.

I don't think it's true: the timeline would recover once it hits the start of the timeline, by triggering back-paginations. Those back-paginations would then fill the gaps at the event cache layer, and those filled gaps would be propagated to the timeline via new vectordiffs. The result would still be confusing to users, though: you got to the top of the apparent timeline, and now the gaps that you *hadn't* seen below would be filled with new messages.

> The previous chunk is of kind Gap —i.e. it contains a… gap—, then a pagination is triggered over the network to fill this gap. The code will stay unchanged here, we already have this mechanism: the Gap is replaced by Items, updates will be broadcasted to the Timeline, happy slow path.

I think that in this case, instead of loading the gap and being satisfied with it (and letting a *future* call to back-pagination resolve it into events), we could block, at this point, and resolve such a gap *right now*. Then we don't have the problem with missing messages, after an app has been backgrounded for a while.

If the app is offline at this point, we would still need a way to display potential gaps.

There's also another caveat to add, if we wanted to represent gaps: we would have many of them that are spurious. In particular, when restarting the app, we don't reuse the SSS's previous `pos`, which will result in an initial sync, that may include events we already know about; in this case, we'll get a prev-batch token => a gap, but when resolving it, it may be spurious (aka, we in fact knew about all the events that have been back-paginated). In terms of user experience, this means we'd display a temporary gap in the timeline, that would later disappear and be replaced by… nothing. Slightly confusing, but that's likely the best we can do without heuristics.

(With heuristics, we could decide to show gaps only after the "initial gap", i.e. a gap that would be observed after an initial SSS response.)

crates/matrix-sdk-common/src/linked_chunk/builder.rs
```diff
@@ -274,10 +274,179 @@ impl<const CAP: usize, Item, Gap> LinkedChunkBuilder<CAP, Item, Gap> {
...
+            // New chunk has a next chunk, and it is the first chunk of the `LinkedChunk`.
+            if next_chunk != expected_next_chunk {
+                return Err(LinkedChunkBuilderError::CannotConnectTwoChunks {
+                    new_chunk: next_chunk,
+                    with_chunk: expected_next_chunk,
+                });
+            }
+
+            // Alright. It's not possible to have a cycle within the chunks or
+            // multiple connected components here. All checks are made.
```
bnjbvr: It's possible to have a cycle in multiple steps, if you start with a linked chunk with ChunkIdentifier (cid) 0, then `insert_new_first_chunk(, { prev: cid(1) })`, then `insert_new_first_chunk(, { prev: cid(0) })`?

crates/matrix-sdk-common/src/linked_chunk/builder.rs
```diff
@@ -274,10 +274,179 @@ impl<const CAP: usize, Item, Gap> LinkedChunkBuilder<CAP, Item, Gap> {
...
+        {
+            // Transform the `RawChunk` into a `Chunk`.
+            let new_first_chunk =
+                Chunk::new_leaked(new_first_chunk.identifier, new_first_chunk.content);
+
+            let links = &mut linked_chunk.links;
+
+            debug_assert!(
+                links.first_chunk().previous.is_none(),
+                "The first chunk is supposed to not have a previous chunk"
```
bnjbvr: nit: here and below, `is not supposed to have a previous chunk` reads more correcter

crates/matrix-sdk/src/event_cache/room/mod.rs
```diff
@@ -580,59 +575,49 @@ mod private {
...
-                EventCacheStoreError::InvalidData { details: err.to_string() }
-            })?)
-        }
-
         /// Create a new state, or reload it from storage if it's been enabled.
+        ///
+        /// Not all events are going to be loaded. Only a portion of them. The
+        /// [`RoomEvents`] relies on a [`LinkedChunk`] to store all events. Only
+        /// the last chunk will be loaded. It means the events are loaded from
+        /// the most recents to the oldest. To load more events, see
```
bnjbvr, line 583: nit: `recents` -> `recent`

crates/matrix-sdk/src/event_cache/room/mod.rs
```diff
@@ -769,72 +754,6 @@ mod private {
...
-                previous: None,
-                next: Some(CId::new(1)),
-            });
-
-            let output = raw_chunks_debug_string(raws);
-            assert_eq!(output.len(), 2);
-            assert_eq!(&output[0], "chunk #0 (prev=<none>, next=1): gap['prev-token']");
-            assert_eq!(&output[1], "chunk #1 (prev=0, next=<none>): events[$12345678, $2]");
-        }
-    }
```
bnjbvr, line 837: Can you move the test over to `RoomEvents::debug_string()`? Or transform it into a minimal test case for `chunk_debug_string` that was called internally and indirectly tested here?

crates/matrix-sdk/src/event_cache/room/mod.rs
```diff
@@ -620,6 +620,79 @@ mod private {
...
+        /// store is disabled.
+        #[must_use = "Updates as `VectorDiff` must probably be propagated via `RoomEventCacheUpdate`"]
+        pub(in super::super) async fn load_more_events_backwards(
+            &mut self,
+        ) -> Result<
+            Option<((Vec<TimelineEvent>, bool), Vec<VectorDiff<TimelineEvent>>)>,
+            EventCacheError,
+        > {
+            // ⚠️ Propagate all pending updates before we do anything.
+            self.propagate_changes().await?;
```
bnjbvr, line 645: :warning: Why do we need to do this? It should be performed just after applying any changes to the in-memory representation of the linked chunk, and having to do it here would suggest there's a bug *elsewhere* (in the place that should've called `propagate_changes()` but didn't.

Strongly think we should not have to do this, and thus remove it.

crates/matrix-sdk/src/event_cache/room/mod.rs
```diff
@@ -620,6 +620,79 @@ mod private {
...
+                // No store: no events to insert.
+                return Ok(None);
+            };
+
+            let first_chunk = self
+                .events
+                .chunks()
+                .next()
+                // SAFETY: A `LinkedChunk` is never empty, it always contains at least one chunk.
+                .expect("The `LinkedChunk` in `RoomEvents` cannot be empty");
```
bnjbvr, line 657: Isn't there a `first_chunk()` method on the links or something?

crates/matrix-sdk/src/event_cache/room/mod.rs
```diff
@@ -620,6 +620,79 @@ mod private {
...
+
+            // Because `first_chunk` is `not `Send`, get these information before the
```
bnjbvr, line 659: nit: information is always singular in English, so `this information`

(and thanks for the comment! this has answered the question i had in my mind, before i could even ask it!)

crates/matrix-sdk/src/event_cache/room/mod.rs
```diff
@@ -620,6 +620,79 @@ mod private {
...
+            // `.await` point, so that this `Future` can implement `Send`.
+            let first_chunk_is_gap = first_chunk.is_gap();
```
bnjbvr, line 661: Turns out this can be inlined, as this is used only once, before any await point.

crates/matrix-sdk/src/event_cache/room/mod.rs
```diff
@@ -620,6 +620,79 @@ mod private {
             Ok(Self { room: room_id, store, events, waited_for_initial_prev_token: false })
         }
 
+        /// Load more events backwards if the last chunk is **not** a gap.
+        ///
+        /// Return `Ok(Some((events, reached_start), _))` if events have been
+        /// inserted, `Ok(None)` if a gap has been inserted or if the
+        /// store is disabled.
```
bnjbvr, line 636: `Ok(None)` could also mean that there's no previous events. Hopefully the callers can handle that properly too :sweat_smile:

crates/matrix-sdk/src/event_cache/pagination.rs
```diff
@@ -121,126 +121,162 @@ impl RoomPagination {
...
+                    // in “reverse order”.
+                    events: events.into_iter().rev().collect(),
+                }));
             }
-        };
+        }
 
-        let paginator = &self.inner.paginator;
+        // Now we can fallback onto the network.
+        {
```
bnjbvr, line 154: Can we avoid opening a new scope here? I don't find it useful, since it's closed at the same time as the function's body. (Only makes the diff harder to read :woozy_face:)

crates/matrix-sdk/src/event_cache/room/mod.rs
```diff
@@ -620,6 +620,79 @@ mod private {
...
+
+            let room_id = &self.room;
+            let store = store.lock().await?;
+
+            // The first chunk is not a gap, we can load its previous chunk.
+            let Some(new_first_chunk) =
+                store.load_previous_chunk(room_id, first_chunk_identifier).await?
+            else {
+                // No previous chunk: no events to insert.
+                return Ok(None);
```
bnjbvr, line 677: I think this return value is incorrect, because it should signal to the caller we've reached the end of the timeline for this room. Otherwise, in the next commit, we'll try to run a back-pagination *after* trying to load a chunk before a first non-gap one.

crates/matrix-sdk/src/event_cache/room/mod.rs
```diff
@@ -1484,10 +1549,9 @@ mod tests {
                         new: ChunkIdentifier::new(0),
                         next: None,
                     },
-                    Update::NewItemsChunk {
-                        previous: Some(ChunkIdentifier::new(0)),
-                        new: ChunkIdentifier::new(1),
-                        next: Some(ChunkIdentifier::new(0)),
+                    Update::PushItems {
+                        at: Position::new(ChunkIdentifier::new(0), 0),
+                        items: iter::repeat_n(event, DEFAULT_CHUNK_CAPACITY + 1).collect(),
```
bnjbvr, line 1570: Hmm, the test shouldn't require any change, since this was testing for cycles. If it failed somehow before this change, it means the incremental rebuilding doesn't correctly handle cycles. Can you revert this please? :nerd_face:

crates/matrix-sdk-common/src/linked_chunk/builder.rs
```diff
@@ -45,33 +234,30 @@ struct TemporaryChunk<Item, Gap> {
 /// then the final call to [`LinkedChunkBuilder::build()`] will result in an
 /// error).
 #[allow(missing_debug_implementations)]
-pub struct LinkedChunkBuilder<const CAP: usize, Item, Gap> {
+#[doc(hidden)]
+pub struct LinkedChunkBuilderTest<const CAP: usize, Item, Gap> {
```
bnjbvr, line 537: If it's only used in tests, consider removing it? :fire:

crates/matrix-sdk-base/src/event_cache/store/integration_tests.rs
```diff
@@ -392,6 +396,220 @@ impl EventCacheStoreIntegrationTests for DynEventCacheStore {
...
+        {
+            let first_chunk = linked_chunk.chunks().next().unwrap().identifier();
+            let mut previous_chunk =
+                self.load_previous_chunk(room_id, first_chunk).await.unwrap().unwrap();
+
+            // Pretend it's the first chunk.
+            previous_chunk.previous = None;
+
+            let _ = LinkedChunkBuilder::insert_new_first_chunk(&mut linked_chunk, previous_chunk)
+                .unwrap();
```
bnjbvr, line 523: Since the caller always has to erase the `.previous` field of the chunk before calling `LinkedChunkBuilder::insert_new_first_chunk`, maybe this could go inside that latter function, instead?

crates/matrix-sdk-sqlite/src/event_cache_store.rs
```diff
@@ -633,9 +633,15 @@ impl EventCacheStore for SqliteEventCacheStore {
                     .query_row(
                         (&hashed_room_id,),
                         |row| {
-                            row.get::<_, u64>(0)
+                            // Read the `MAX(id)` as an `Option<u64>` instead
+                            // of `u64` in case the `SELECT` returns nothing.
+                            // Indeed, if it returns no line, the `MAX(id)` is
+                            // set to `Null`.
```
bnjbvr, line 654: nice, good job on the testing!!

crates/matrix-sdk-common/src/linked_chunk/builder.rs
```diff
@@ -480,7 +583,7 @@ pub enum LinkedChunkBuilderTestError {
 }
 
 #[cfg(test)]
-mod tests {
+mod linked_builder_test_tests {
```
bnjbvr, line 782: uh? we do use `tests` by convention everywhere in the code base, can we maintain this?

crates/matrix-sdk/tests/integration/event_cache.rs
```diff
@@ -1608,3 +1610,256 @@ async fn test_apply_redaction_when_redacted_and_redaction_are_in_same_sync() {
...
+        assert_event_id!(initial_updates[5], "$ev3_5");
+        assert_event_id!(initial_updates[6], "$ev3_6");
+
+        // The stream of updates is waiting, patiently.
+        assert!(updates_stream.is_empty());
+    }
+
+    // Let's paginate to load more!
+    //
+    // One more chunkw will be loaded from the store. This new chunk contains 5
```
bnjbvr, line 1739: nit: `chunkw`

bnjbvr (commented):
Looks super nice at a first glance, well done! I like the thorough testing at each layer, and the nice commenting in the tests.

I have a few fundamental proposals below, and a few nits here and there. Let's discuss them :)

crates/matrix-sdk-base/src/event_cache/store/traits.rs
```diff
@@ -69,11 +69,32 @@ pub trait EventCacheStore: AsyncTraitDeps {
...
+    /// Load the chunk before the chunk identified by `before_chunk_identifier`
+    /// of the `LinkedChunk` holding all events of the room identified by
+    /// `room_id`
+    ///
+    /// This is used to iteratively load events for the `EventCache`.
+    async fn load_previous_chunk(
+        &self,
+        room_id: &RoomId,
+        before_chunk_identifier: ChunkIdentifier,
+    ) -> Result<Option<RawChunk<Event, Gap>>, Self::Error>;
```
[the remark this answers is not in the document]
Hywan: Agree. As discussed, it's gonna be a follow up PR. This is not blocking the overall design, and a rebase would be quite annoying at this step. Let's do that as an immediate follow up!

crates/matrix-sdk/src/event_cache/room/mod.rs
Hywan: As discussed: I've not added this due to a bug, but by precaution. I will remove it and see what happens. It should not change anything, it was pure precaution.

crates/matrix-sdk/src/event_cache/room/mod.rs
Hywan: No, `Ends` (`LinkedChunk::links`) is private. I can add a public method to get the first chunk if you want?

crates/matrix-sdk/src/event_cache/room/mod.rs
Hywan: Correct. I need to update the comment!

crates/matrix-sdk-common/src/linked_chunk/builder.rs
Hywan: As discussed, will be done as an immediate follow up.

crates/matrix-sdk-base/src/event_cache/store/integration_tests.rs
Hywan: Debatable, but I'm fine with that.

crates/matrix-sdk-common/src/linked_chunk/builder.rs
Hywan: Another `mod tests` exists in this file. I can merge both of them if you prefer.

crates/matrix-sdk-common/src/linked_chunk/builder.rs
```diff
@@ -314,6 +315,201 @@ mod tests {
...
+        let new_first_chunk = RawChunk {
+            previous: None,
+            identifier: ChunkIdentifier::new(0),
+            next: None,
+            content: ChunkContent::Gap(()),
+        };
+
+        let mut linked_chunk = LinkedChunk::<2, char, ()>::new();
+
+        let result = LinkedChunkBuilder::insert_new_first_chunk(&mut linked_chunk, new_first_chunk);
```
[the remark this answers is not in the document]
Hywan: Good idea.

crates/matrix-sdk-common/src/linked_chunk/builder.rs
Hywan: :+1:

crates/matrix-sdk/src/event_cache/room/mod.rs
Hywan: :+1:

crates/matrix-sdk/src/event_cache/pagination.rs
```diff
@@ -121,6 +121,37 @@ impl RoomPagination {
     async fn run_backwards_impl(&self, batch_size: u16) -> Result<Option<BackPaginationOutcome>> {
         const DEFAULT_WAIT_FOR_TOKEN_DURATION: Duration = Duration::from_secs(3);
 
+        // First off, remember that's the `RoomEvents` might be partially loaded
+        // (because not all events are fully loaded).
+        //
+        // Knowing that, let's try to load more events from the storage. If there is
+        // nothing from the storage, let's fallback on the network.
```
bnjbvr, line 128: :warning: This is only partially correct: sync may have pushed a new gap, *after* events we partially loaded from the linked chunk. In that case, **we should prefer resolving the new gap, to loading from storage** (i.e. calling network /messages), otherwise we risk apparently losing track of messages.

Suppose we have the following, before client start:

- the room cache contains [A, B, C]
- the server-side room state is [A, B, C, D, E, F]

Then, we start the app:
- we load a linked chunk from storage: let's say chunks are of size 1 (it's arbitrary for the sake of example, but any size would cause the same problems, assuming more events), we only get C in the room.
- sync with timeline limit=1: we receive [Gap1, F]. The chunk is now [C, Gap1, F].
- we request to back-paginate. Because of this code, we'll load [A, B] into the timeline, and will get: [A, B, C, Gap1, F].

In this case, the events from Gap1 seem to be missing. They would be missing *until we get to the start of the known timeline* (i.e. until we've reloaded all the chunks we knew about), so that's a bit problematic. In my example, it's only reloading a single chunk before back-paginating Gap1; in real life, there could be lots and lots of events in the persisted cache storage, and thus the user would see the missing events *way later* after all the previous events have been reloaded. (Probably they wouldn't notice the missing events even have been resolved, because they're looking at the top / beginning of the linked chunk's view, at this point.)

crates/matrix-sdk/src/event_cache/room/mod.rs
Hywan: Done.

crates/matrix-sdk/src/event_cache/room/mod.rs
Hywan: :+1:

crates/matrix-sdk/src/event_cache/room/mod.rs
Hywan: True!

crates/matrix-sdk/src/event_cache/room/mod.rs
Hywan: I think you're correct.

Either the first chunk is a `Gap`, and it can be resolved to more events.
Either the first chunk is an `Items`, and if there is nothing before that, it means we've reached the start of the timeline.

crates/matrix-sdk/src/event_cache/pagination.rs
Hywan: Exactly. This is related to https://github.com/matrix-org/matrix-rust-sdk/pull/4632#discussion_r1952755702 and is now solved!

Hywan:
Patches have been re-organized to be extracted in 3 or 4 parts. First part is here, https://github.com/matrix-org/matrix-rust-sdk/pull/4678.

bnjbvr (approved):
Awesome!!! Thanks for your work there, only minor comments here so I'm approving this :-)

Great job, I'm really delighted with our design, and that the room event cache simulates back-paginations mostly like `/messages`. Very nice!

crates/matrix-sdk/src/event_cache/pagination.rs
```diff
@@ -212,24 +212,38 @@ impl RoomPagination {
...
-
-        let (new_events, duplicated_event_ids, all_deduplicated) =
-            state.collect_valid_and_duplicated_events(sync_events.clone()).await?;
+            // (backwards). The `RoomEvents` API expects the first event to be the oldest.
+            // Let's re-order them for this block.
+            let events = events
+                .iter()
+                .rev()
+                .cloned()
+                .collect::<Vec<_>>();
```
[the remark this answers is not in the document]
bnjbvr: Ok, this is not the same `events` as the `events` outside, which is used for constructing the `BackPaginationOutcome` below :P Let's rename this one `reversed_events` or something like that, to avoid the confusion please :D

crates/matrix-sdk/tests/integration/event_cache.rs
```diff
@@ -1606,3 +1608,427 @@ async fn test_apply_redaction_when_redacted_and_redaction_are_in_same_sync() {
...
+        let pagination_outcome = room_event_cache
+            .pagination()
+            .run_backwards(10, until_at_least_one_event)
+            .await
+            .unwrap();
+
+        // No events, hmmm…
+        assert!(pagination_outcome.events.is_empty());
+
+        // … that's because the start of the timeline is finally reached!
```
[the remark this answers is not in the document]
Hywan: What a ride!

Hywan:
Immediate follow up https://github.com/matrix-org/matrix-rust-sdk/issues/4684

Issue #4593, opened 2025-01-29
Migrate from mas-oidc (adding wasm support to sdk-ffi crate)


ospfranco:
Part of the ongoing effort to add support for WASM to the matrix SDK is adding WASM support to each of the crates/bindings. Several of the crates already have been modified and add support for WASM compilation. One of the final crates that needs modification is the `bindings/matrix-sdk-ffi` crate.

I've done some preliminary work ([in my fork](https://github.com/ospfranco/matrix-rust-sdk/tree/oscar/wasm-ffi)), however, at some point got stuck because the crate depends on `mas-oidc-client` (sub dependency of using [experimental-oidc](https://github.com/matrix-org/matrix-rust-sdk/blob/main/bindings/matrix-sdk-ffi/Cargo.toml#L72)) for the open ID authentication. Not only that it is stuck on 0.11.0 version of the crate, the crate itself using multi-threaded code and crates which make it impossible to add support for a WASM target. I asked the maintaners of the crate and they said that it was too much to keep the crate both a generic openID crate and a next-gen auth for building matrix clients was too much. Moreover, the license has been changed to AGPL-3.0 which means further versions than 0.11.0 can no longer be used on the matrix-sdk.

As such, the intention is to replace the usage of `mas-oidc-client` with a license compatible and wasm compatible implementation, that being a hand-written one or another compatible crate.

poljar:
Due to `mas-oidc-client` not supporting everything we need, we're already using the [openidconnect](https://github.com/ramosbugs/openidconnect-rs/) crate in some places.

We would probably just migrate fully to it. (Unless I misunderstood what those crates are about and they aren't interchangeable).

ospfranco:
They probably do the same :)

zecakeh:
The MSCs have migrated away from OpenID Connect to more basic OAuth 2.0. Since OpenID Connect adds things on top , I am afraid that their API will fail on servers that are not compliant with OIDC but that are compliant with the MSCs.

My thoughts were to use [oauth2](https://github.com/ramosbugs/oauth2-rs) instead of openidconnect, that is from the same developer so that has a very similar API. The only thing is that it doesn't support server metadata discovery and client registration, but it shouldn't be too hard to add. See also the discussion in #4550.

I don't know if you are planning to do this and when, but I was planning to work on this next month if that helps 🙂 .

ospfranco:
[...]

Issue #4550, opened 2025-01-17
Update support of MSC2965 OAuth 2.0 Authorization Server Metadata discovery


sandhose:
MSC2965 was updated to use a C-S API endpoint to retrieve the authorisation server metadata, instead of relying on the OpenID Connect Discovery spec.

One thing that is a little annoying, is that the SDK currently relies on crates originally published by MAS, that aren't updated anymore, and that were pretty pedantic about the server metadata discovery.

This means that a lot of the logic might have to be backported to within the SDK instead, so even if the MSC2965 change is supposed to be relatively small, in practice it might be bigger to do for the SDK.

Note that this isn't too urgent, as MAS intends to keep the old endpoint working for some time, to avoid breaking other apps.

Synapse implementation: https://github.com/element-hq/synapse/pull/18093

zecakeh:
I opened https://github.com/ruma/ruma/pull/1997 to add the new endpoint to Ruma.

I also opened https://github.com/ramosbugs/oauth2-rs/issues/298 on the oauth2 crate to add support for the provider metadata and client registration. These seem to be the only missing RFCs that would be needed currently. In the worst case scenario they could be implemented in the SDK.

zecakeh:
If someone from the SDK team could review the ruma PR, it would help to move things along.

PR #4635, opened 2025-02-06
feat(auth): get rid of the cargo feature flag for OIDC :fire: 

bnjbvr:
OIDC authentication has been used in production in multiple embeddings of the Matrix Rust SDK, some of them for months already, and they're considered stable for everyday use. As such, the feature is not considered experimental anymore, especially since the future of authentication will rely on OIDC and related mechanisms.

---

The cross-process lock used for OIDC relies on the encryption crate, which implements a cross-process locking mechanism. So lots of code have to be guarded against the `e2e-encryption` feature flag, as we need to have access to the crypto features to enable the cross-process lock, unfortunately. This dependency should go away as soon as we generalize the concept of the cross-process lock, so it doesn't make use of the crypto store for that purpose.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -35,7 +35,6 @@ jobs:
       fail-fast: true
       matrix:
         name:
-          - experimental-oidc
           - no-encryption
           - no-sqlite
           - no-encryption-and-sqlite
--- a/bindings/matrix-sdk-ffi/Cargo.toml
+++ b/bindings/matrix-sdk-ffi/Cargo.toml
@@ -55,7 +55,6 @@ workspace = true
 features = [
     "anyhow",
     "e2e-encryption",
-    "experimental-oidc",
     "experimental-widgets",
     "markdown",
     "rustls-tls", # note: differ from block below
@@ -69,7 +68,6 @@ workspace = true
 features = [
     "anyhow",
     "e2e-encryption",
-    "experimental-oidc",
     "experimental-widgets",
     "markdown",
     "native-tls", # note: differ from block above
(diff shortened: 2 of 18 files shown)


bnjbvr:
It's `mas-oidc-client` which is the culprit here. Unfortunately, quick experiments to make it compatible by removing features is unfruitful, because it's using a different version of tracing that's not our special patched version :/ Will investigate further next week.

zecakeh:
There is https://github.com/matrix-org/matrix-rust-sdk/issues/4593, fwiw.

zecakeh:
It might also be a little premature to remove the feature, given that it doesn't implement the latest versions of the MSCs (#4550).

bnjbvr:
> It might also be a little premature to remove the feature, given that it doesn't implement the latest versions of the MSCs (https://github.com/matrix-org/matrix-rust-sdk/issues/4550 and downgrading to OAuth 2.0 instead of OIDC).

We discuss it internally, and we still think it's fine to get rid of the feature. Our implementation is quite stable and "production-ready", in that it's used in the ElementX apps. While it's also been constantly evolving for years, this is the fate of such a feature, and changes are expected to happen slowly and in a non-disruptive manner, which is a sufficient sign of maturity.

---

On the other hand, getting rid of the `mas-oidc-client` crate might be a good justification to hold off, since 1. it prevents compilation on wasm targets at the moment, and it's unclear it's worth investing into making it compile there, 2. there's been discussions about not using it, because it might not be well maintained and support all the features we need/use in the medium term.

bnjbvr:
Closing, this is not going anywhere, not highly urgent, and i have no time investigating getting rid of `mas-oidc-client` :disappointed:

PR #4647, opened 2025-02-08
feat(sdk): Only allow TLS 1.2 or newer

zecakeh:
As recommended by [BCP 195](https://datatracker.ietf.org/doc/bcp195/).

It think it's better if the SDK enforces best current security practices by default.

It shouldn't be a problem with rustls that only supports TLS 1.2 and 1.3, but with native-tls it depends on the implementation.

### DIFF
--- a/crates/matrix-sdk/CHANGELOG.md
+++ b/crates/matrix-sdk/CHANGELOG.md
@@ -11,6 +11,7 @@ All notable changes to this project will be documented in this file.
 - The `MediaRetentionPolicy` can now trigger regular cleanups with its new
   `cleanup_frequency` setting.
   ([#4603](https://github.com/matrix-org/matrix-rust-sdk/pull/4603))
+- The HTTP client only allows TLS 1.2 or newer, as recommended by BCP 195.
 
 ## [0.10.0] - 2025-02-04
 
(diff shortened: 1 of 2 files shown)


Hywan:
I think I'm fine with this change. Having a second review from @poljar or @BillCarsonFr would be valuable.

Hywan:
[BCP stands for _Best Current Practices_](https://datatracker.ietf.org/doc/bcp/) in case someone was wondering (like I was)

crates/matrix-sdk/CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ All notable changes to this project will be documented in this file.
 - The `MediaRetentionPolicy` can now trigger regular cleanups with its new
   `cleanup_frequency` setting.
   ([#4603](https://github.com/matrix-org/matrix-rust-sdk/pull/4603))
+- The HTTP client only allows TLS 1.2 or newer, as recommended by BCP 195.
```
Hywan, line 14: I think it should be marked as a breaking change. Plus, can you add a link to the PR + a link to the BCP please?

Hywan (changes requested):

crates/matrix-sdk/CHANGELOG.md
zecakeh: Done

poljar (approved):
Yup, makes sense.

PR #4649, opened 2025-02-10
feat(timeline): handle more timeline item content kinds in replied-to details

bnjbvr:
Right now, the replied-to details can only contain information about a room message, while we can have other timeline item contents for an item. In particular, the replied-to event could be still encrypted (UTD), and that shouldn't be considered an "unsupported kind", as in https://github.com/element-hq/element-x-ios/issues/3774. Instead, consumers can display the replied-to event was a UTD, and perform manual retries in the background later on (after a backup key has been downloaded, for instance).

Remaining tasks:

- [x] Add tests that a UTD can appear in the replied-to details.
- [x] (and other kinds of contents: poll…
- [x] (…sticker, and so on)
- ~~[ ] Since the code getting a `TimelineItemContent` from a Ruma event is so similar to the code used when handling an event, I'm planning to attempt to merge the two.~~
  - will be a followup

### DIFF
--- a/crates/matrix-sdk-ui/src/timeline/event_item/content/message.rs
+++ b/crates/matrix-sdk-ui/src/timeline/event_item/content/message.rs
@@ -37,7 +37,7 @@ use ruma::{
     serde::Raw,
     OwnedEventId, OwnedUserId, UserId,
 };
-use tracing::{error, trace};
+use tracing::{debug, error, instrument, trace, warn};
 
 use super::TimelineItemContent;
 use crate::{
@@ -354,18 +354,27 @@ impl RepliedToEvent {
         Self::try_from_timeline_event(timeline_event, room_data_provider).await
     }
 
+    #[instrument(skip_all)]
     pub(in crate::timeline) async fn try_from_timeline_event<P: RoomDataProvider>(
         timeline_event: TimelineEvent,
         room_data_provider: &P,
     ) -> Result<Self, TimelineError> {
         let event = match timeline_event.raw().deserialize() {
             Ok(AnySyncTimelineEvent::MessageLike(event)) => event,
-            _ => {
+            Ok(_) => {
+                warn!("can't get details, event isn't a message-like event");
+                return Err(TimelineError::UnsupportedEvent);
+            }
+            Err(err) => {
+                warn!("can't get details, event couldn't be deserialized: {err}");
                 return Err(TimelineError::UnsupportedEvent);
             }
         };
 
+        debug!(event_type = %event.event_type(), "got deserialized event");
+
         let Some(AnyMessageLikeEventContent::RoomMessage(c)) = event.original_content() else {
+            warn!("can't get details, event is redacted or not a RoomMessage");
             return Err(TimelineError::UnsupportedEvent);
         };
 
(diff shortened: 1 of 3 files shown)


bnjbvr:
Note for reviewers: don't be afraid by the number of LOC, there are many repetitive tests in there.

crates/matrix-sdk-ui/src/timeline/event_item/content/message.rs
```diff
@@ -373,20 +379,69 @@ impl RepliedToEvent {
...
+
+                    let utd_cause = if let Some(utd_info) = utd_info {
+                        UtdCause::determine(
+                            timeline_event.raw(),
+                            room_data_provider.crypto_context_info().await,
+                            &utd_info,
+                        )
+                    } else {
+                        UtdCause::Unknown
+                    };
```
jplatte: Seems simpler to use a single `match` for this, no?

jplatte (approved):

crates/matrix-sdk-ui/src/timeline/event_item/content/message.rs
```diff
@@ -373,20 +379,69 @@ impl RepliedToEvent {
...
         debug!(event_type = %event.event_type(), "got deserialized event");
 
-        let Some(AnyMessageLikeEventContent::RoomMessage(c)) = event.original_content() else {
-            warn!("can't get details, event is redacted or not a RoomMessage");
-            return Err(TimelineError::UnsupportedEvent);
-        };
+        let content = match event.original_content() {
+            Some(content) => match content {
+                AnyMessageLikeEventContent::RoomMessage(c) => {
+                    // Assume we're not interested in reactions in this context.
```
poljar, line 385: It might make sense to explain why we're not interested.

crates/matrix-sdk-ui/src/timeline/event_item/content/message.rs
```diff
@@ -354,30 +360,88 @@ impl RepliedToEvent {
...
+                        content, utd_cause,
+                    ))
+                }
+
+                AnyMessageLikeEventContent::UnstablePollStart(
+                    UnstablePollStartEventContent::New(content),
+                ) => {
+                    // Assume we're not interested in reactions in this context.
+                    let reactions = ReactionsByKeyBySender::default();
+                    // TODO could we provide the bundled edit here?
```
poljar, line 428:
```suggestion
                    // TODO: could we provide the bundled edit here?
```

poljar (approved):

crates/matrix-sdk-ui/src/timeline/event_item/content/message.rs
bnjbvr: Good idea!

crates/matrix-sdk-ui/src/timeline/event_item/content/message.rs
bnjbvr: Good point, edited with this comment:

```rust
// Assume we're not interested in reactions in this context: this is
// information for an embedded (replied-to) event, that will usually not
// include detailed information like reactions.
```

PR #4651, opened 2025-02-11
feat(sqlite): Run VACUUM operation after removing a room 

zecakeh:
A room can be associated to a lot of data, depending on the number of members in the room.
So freeing space on the filesystem should be worth it in some cases.

An (extreme) example: I have a test account that is in ~60 rooms, a few of those big public rooms, including Matrix HQ. The size of the `matrix-sdk-state.sqlite3` file is 542 MB. Using this PR and leaving, then forgetting Matrix HQ brings the DB down to 255 MB.

### DIFF
--- a/crates/matrix-sdk-sqlite/CHANGELOG.md
+++ b/crates/matrix-sdk-sqlite/CHANGELOG.md
@@ -10,6 +10,8 @@ All notable changes to this project will be documented in this file.
 
 - Implement the new method of `EventCacheStoreMedia` for `SqliteEventCacheStore`.
   ([#4603](https://github.com/matrix-org/matrix-rust-sdk/pull/4603))
+- Defragment the state store after removing a room.
+  ([#4651](https://github.com/matrix-org/matrix-rust-sdk/pull/4651))
 
 ## [0.10.0] - 2025-02-04
 
--- a/crates/matrix-sdk-sqlite/CHANGELOG.md
+++ b/crates/matrix-sdk-sqlite/CHANGELOG.md
@@ -10,7 +10,7 @@ All notable changes to this project will be documented in this file.
 
 - Implement the new method of `EventCacheStoreMedia` for `SqliteEventCacheStore`.
   ([#4603](https://github.com/matrix-org/matrix-rust-sdk/pull/4603))
-- Defragment the state store after removing a room.
+- Defragment an sqlite state store after removing a room.
   ([#4651](https://github.com/matrix-org/matrix-rust-sdk/pull/4651))
 
 ## [0.10.0] - 2025-02-04
--- a/crates/matrix-sdk-sqlite/src/event_cache_store.rs
+++ b/crates/matrix-sdk-sqlite/src/event_cache_store.rs
@@ -36,8 +36,6 @@ use matrix_sdk_store_encryption::StoreCipher;
 use ruma::{time::SystemTime, MilliSecondsSinceUnixEpoch, MxcUri, RoomId};
 use rusqlite::{params_from_iter, OptionalExtension, Transaction, TransactionBehavior};
 use tokio::fs;
-#[cfg(not(test))]
-use tracing::warn;
 use tracing::{debug, trace};
 
 use crate::{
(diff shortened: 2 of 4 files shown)


crates/matrix-sdk-sqlite/src/state_store.rs
```diff
@@ -1739,40 +1739,42 @@ impl StateStore for SqliteStateStore {
...
-            .await
+            let dependent_send_queue_room_id =
+                this.encode_key(keys::DEPENDENTS_SEND_QUEUE, &room_id);
+            txn.remove_room_dependent_send_queue(&dependent_send_queue_room_id)?;
+
+            Ok(())
+        })
+        .await?;
+
+        conn.vacuum().await
```
bnjbvr, line 1777: I'm wondering if we can vacuum as part of the transaction, or if it would lead to more trouble? It's not clear from the sqlite doc…

crates/matrix-sdk-sqlite/src/event_cache_store.rs
```diff
@@ -959,19 +957,10 @@ impl EventCacheStoreMedia for SqliteEventCacheStore {
...
-                // Since this is an optimisation step, do not propagate the error
-                // but log it.
-                #[cfg(not(test))]
-                warn!("Failed to vacuum database: {error}");
-
-                // We want to know if there is an error with this step during tests.
-                #[cfg(test)]
-                return Err(error.into());
-            }
+            conn.vacuum().await?;
```
bnjbvr, line 963: Preexisting, but does it make sense to vacuum all the time? If only a single media has been removed, it's probably not worth it — especially as vacuuming requires doubling the size of the database on disk, since it creates a copy of the database while it does it. Unfortunately that means heuristics to decide when to trigger, then, and maybe over a certain number of files / total size on disk, that would be profitable and simple enough to trigger often?

crates/matrix-sdk-sqlite/src/utils.rs
```diff
@@ -22,6 +22,8 @@ use matrix_sdk_store_encryption::StoreCipher;
 use ruma::time::SystemTime;
 use rusqlite::{limits::Limit, OptionalExtension, Params, Row, Statement, Transaction};
 use serde::{de::DeserializeOwned, Serialize};
+#[cfg(not(test))]
+use tracing::warn;
```
bnjbvr, line 26: I'd use `tracing::warn!` at the single place where it's used instead, to not avoid a conditional `use` statement (which I personally find ugly).

crates/matrix-sdk-sqlite/src/utils.rs
```diff
@@ -136,6 +138,24 @@ pub(crate) trait SqliteAsyncConnExt {
...
+    /// Only returns an error in tests, otherwise the error is only logged.
+    async fn vacuum(&self) -> Result<()> {
+        if let Err(error) = self.execute_batch("VACUUM").await {
+            // Since this is an optimisation step, do not propagate the error
+            // but log it.
+            #[cfg(not(test))]
+            warn!("Failed to vacuum database: {error}");
+
+            // We want to know if there is an error with this step during tests.
+            #[cfg(test)]
```
bnjbvr, line 153: Can we enable it if we've compiled in debug mode too? (that is, with `#[cfg(debug_assertions)]` too)

crates/matrix-sdk-sqlite/CHANGELOG.md
```diff
@@ -10,6 +10,8 @@ All notable changes to this project will be documented in this file.
 
 - Implement the new method of `EventCacheStoreMedia` for `SqliteEventCacheStore`.
   ([#4603](https://github.com/matrix-org/matrix-rust-sdk/pull/4603))
+- Defragment the state store after removing a room.
```
bnjbvr, line 13:
```suggestion
- Defragment an sqlite state store after removing a room.
```

bnjbvr (approved):
Thanks, makes sense

crates/matrix-sdk-sqlite/src/state_store.rs
zecakeh: No, it's not possible it returns an error.

crates/matrix-sdk-sqlite/src/utils.rs
zecakeh: Done

crates/matrix-sdk-sqlite/src/utils.rs
zecakeh: Done

crates/matrix-sdk-sqlite/CHANGELOG.md
zecakeh: Done

PR #4654, opened 2025-02-11
fix(ui): Demote aggregation target not found log to TRACE level

zecakeh:
We encountered this warning a lot in the logs after upgrading the SDK today. This comes from https://github.com/matrix-org/matrix-rust-sdk/pull/4576.

My understanding is that this path is expected if the target event is not yet in the timeline, which can happen often when paginating backwards, so it's nothing to warn about.

cc @bnjbvr


bnjbvr (approved):
Yeah, this would happen every time we have an aggregation but not the item it relates to; we can demote this for the time being, and see if it's a problem in our future rageshakes :+1:

PR #4655, opened 2025-02-11
refactor(timeline): replace `TimelineEventKind` with `TimelineAction`

bnjbvr:
This centralizes (and simplifies, IMO) the creation of new timeline items vs handling of aggregations. Before, this was done partly when creating the (type previously known as) `TimelineEventKind` and handling it later in the `TimelineEventHandler`. Now, the event is analyzed once, and analyzing the event results in three outcomes:

- either the event type is not supported: we do nothing
- or it would result in a new timeline item: the item is created and wrapped in `TimelineAction::AddItem`
- or it would result in a new aggregation: the aggregation is created and wrapped in `TimelineAction::HandleAggregation`

That way, it should be much simpler to handle aggregation-only events in non-live timelines, which will be handy for thread timelines.

Part of #4869.

### DIFF
--- a/crates/matrix-sdk-ui/src/timeline/event_handler.rs
+++ b/crates/matrix-sdk-ui/src/timeline/event_handler.rs
@@ -154,6 +154,11 @@ pub(super) enum TimelineEventKind {
 }
 
 impl TimelineEventKind {
+    /// Create a new [`TimelineEventKind::AddItem`] with no edit json.
+    fn add_item(content: TimelineItemContent) -> Self {
+        Self::AddItem { content, edit_json: None }
+    }
+
     /// Creates a new `TimelineEventKind`.
     ///
     /// # Arguments
@@ -184,10 +189,7 @@ impl TimelineEventKind {
                 if let Some(redacts) = ev.redacts(&room_version).map(ToOwned::to_owned) {
                     Self::Redaction { redacts }
                 } else {
-                    Self::AddItem {
-                        content: redacted_message_or_none(ev.event_type())?,
-                        edit_json: None,
-                    }
+                    Self::add_item(redacted_message_or_none(ev.event_type())?)
                 }
             }
 
@@ -213,14 +215,11 @@ impl TimelineEventKind {
                             .await;
                         }
 
-                        Self::AddItem {
-                            content: TimelineItemContent::MsgLike(
-                                MsgLikeContent::unable_to_decrypt(EncryptedMessage::from_content(
-                                    content, utd_cause,
-                                )),
-                            ),
-                            edit_json: None,
-                        }
+                        Self::add_item(TimelineItemContent::MsgLike(
+                            MsgLikeContent::unable_to_decrypt(EncryptedMessage::from_content(
+                                content, utd_cause,
+                            )),
+                        ))
                     } else {
                         // If we get here, it means that some part of the code has created a
                         // `TimelineEvent` containing an `m.room.encrypted` event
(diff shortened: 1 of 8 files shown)


crates/matrix-sdk-ui/src/timeline/event_handler.rs
```diff
@@ -135,6 +135,8 @@ pub(super) struct TimelineEventContext {
 #[derive(Clone, Debug)]
 pub(super) enum HandleAggregationKind {
     Reaction { key: String },
+
+    Redaction,
```
Hywan, line 130: Is it possible to add a bit of documentation here please?

Also, it's really cool how everything fits together. Nice refactoring!

crates/matrix-sdk-ui/src/timeline/event_handler.rs
```diff
@@ -137,6 +137,8 @@ pub(super) enum HandleAggregationKind {
     Reaction { key: String },
 
     Redaction,
+
+    Edit { replacement: Replacement<RoomMessageEventContentWithoutRelation> },
```
Hywan, line 132: Is it possible to add a bit of documentation here please?

crates/matrix-sdk-ui/src/timeline/event_handler.rs
```diff
@@ -139,6 +135,12 @@ pub(super) enum HandleAggregationKind {
     Redaction,
 
     Edit { replacement: Replacement<RoomMessageEventContentWithoutRelation> },
+
+    PollResponse { answers: Vec<String> },
+
+    PollEdit { replacement: Replacement<NewUnstablePollStartEventContentWithoutRelation> },
+
+    PollEnd,
```
Hywan: Is it possible to add a bit of documentation here please?

crates/matrix-sdk-ui/src/timeline/event_handler.rs
```diff
@@ -525,18 +528,6 @@ impl<'a, 'o> TimelineEventHandler<'a, 'o> {
...
-                    if should_add {
-                        self.add_item(TimelineItemContent::CallInvite, None);
-                    }
-                }
-
-                AnyMessageLikeEventContent::CallNotify(_) => {
-                    if should_add {
-                        self.add_item(TimelineItemContent::CallNotify, None)
-                    }
-                }
```
Hywan: Hurray to this kind of removals!

crates/matrix-sdk-ui/src/timeline/event_handler.rs
```diff
@@ -477,6 +274,199 @@ impl TimelineEventKind {
         })
     }
 
+    pub(super) fn from_content(
```
Hywan, line 299: Niiice! I have removed a previous comment where I was suggesting to do that \o/.

crates/matrix-sdk-ui/src/timeline/event_handler.rs
```diff
@@ -620,11 +610,9 @@ impl<'a, 'o> TimelineEventHandler<'a, 'o> {
             }
         };
 
-        let should_add = self.ctx.should_add_new_items;
```
Hywan, line 392: 👏

crates/matrix-sdk-ui/src/timeline/event_handler.rs
```diff
@@ -138,30 +138,44 @@ pub(super) enum HandleAggregationKind {
     PollEnd,
 }
 
+/// An action that we want to cause on the timeline.
 #[derive(Clone, Debug)]
-pub(super) enum TimelineEventKind {
-    AddItem { content: TimelineItemContent, edit_json: Option<Raw<AnySyncTimelineEvent>> },
+pub(super) enum TimelineAction {
```
Hywan, line 143: This is really great. I like the name!

PR #4657, opened 2025-02-12
fix(tests): Prevent test flake by using different names for the test stores

andybalaam:
`test_delete_dehydration_pickle_key` was flaking locally for me. This change appears to fix it.


crates/matrix-sdk-crypto/src/store/integration_tests.rs
```diff
@@ -1218,8 +1218,9 @@ macro_rules! cryptostore_integration_tests {
             }
 
             #[async_test]
+            #[ignore]
```
bnjbvr, line 1221: this will not run the test, so admittedly yeah, it won't make it flaky anymore :sweat_smile:

andybalaam: Sorry, fixed in 9dab41e118

PR #4659, opened 2025-02-12
feat(base): Add `EventCacheStore::filter_duplicated_events`

Hywan:
This patch adds and implements the
`EventCacheStore::filter_duplicated_events` method. It is implemented on the `MemoryStore` and the `SqliteEventCacheStore`.

This method remove the unique events and return the duplicated events.

On the `MemoryStore`, it iterates over events only once.
On the `SqliteEventCacheStore`, it runs a single query (modulo the size of events to check, see
`chunk_large_query_over`).

---

* Address #3280
* Unblock #4632

### DIFF
--- a/crates/matrix-sdk-base/src/event_cache/store/memory_store.rs
+++ b/crates/matrix-sdk-base/src/event_cache/store/memory_store.rs
@@ -26,7 +26,7 @@ use matrix_sdk_common::{
 };
 use ruma::{
     time::{Instant, SystemTime},
-    MxcUri, OwnedMxcUri, RoomId,
+    MxcUri, OwnedEventId, OwnedMxcUri, RoomId,
 };
 
 use super::{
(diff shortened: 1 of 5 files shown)


crates/matrix-sdk-base/src/event_cache/store/memory_store.rs
```diff
@@ -148,6 +148,35 @@ impl EventCacheStore for MemoryStore {
...
+        // Collect all duplicated events.
+        let inner = self.inner.read().unwrap();
+
+        let mut duplicated_events = Vec::new();
+
+        for event in inner.events.unordered_events(room_id) {
+            // If `events` is empty, we can short-circuit.
+            if events.is_empty() {
+                break;
+            }
```
bnjbvr, line 165: Can you `if events.is_empty() { return; }` at the top of the events, instead?

crates/matrix-sdk-base/src/event_cache/store/memory_store.rs
```diff
@@ -148,6 +148,35 @@ impl EventCacheStore for MemoryStore {
...
+
+            if let Some(event_id_a) = event.event_id() {
+                // This event exists in the store event!
+                if let Some(position) =
+                    events.iter().position(|event_id_b| &event_id_a == event_id_b)
+                {
+                    duplicated_events.push(events.remove(position));
+                }
```
bnjbvr, line 173: Maybe we could name `event_id_a` a `known_event_id` and `event_id_b` a `new_event_id`?

crates/matrix-sdk-common/src/linked_chunk/relational.rs
```diff
@@ -292,6 +292,21 @@ impl<Item, Gap> RelationalLinkedChunk<Item, Gap> {
...
+
+    /// Return an iterator that yields events of a particular room with no
+    /// particular order.
+    pub fn unordered_events<'a>(&'a self, room_id: &'a RoomId) -> impl Iterator<Item = &'a Item> {
+        self.items.iter().filter_map(move |item_row| {
+            if item_row.room_id == room_id {
+                match &item_row.item {
+                    Either::Item(item) => Some(item),
+                    Either::Gap(..) => None,
+                }
```
bnjbvr, line 304: super light suggestion: `as_variant!`? :-)

bnjbvr (approved):
Thanks! Super nice that the approach we devised for the sqlite store with a single `SELECT` query works flawlessly :ok_hand:

crates/matrix-sdk-base/src/event_cache/store/memory_store.rs
Hywan: No because we remove events one by one from `events`. When it's empty, we can stop iterating all other events.

crates/matrix-sdk-base/src/event_cache/store/memory_store.rs
Hywan: Done!

crates/matrix-sdk-common/src/linked_chunk/relational.rs
Hywan: Thoughts about it, but I decided it wasn't worth it :-(.

PR #4660, opened 2025-02-12
fix: `RoomEventCache::subscribe` is now infallible

Hywan:
This patch updates `RoomEventCache::subscribe` to be infallible. This method wasn't able to return something else than an `Ok`. The return type has been updated from `Result<T>` to `T`.

---

* Extracted from #4632


bnjbvr (approved):
Super nice, thanks!

PR #4662, opened 2025-02-12
feat(event cache): deduplicate using a store when available

bnjbvr:
Based on the glorious work of #4659 and #4656, this introduces a new way to deduplicate events, whether we have enabled storage or not, using a plain old sum type :nerd_face: When storage is enabled by default for everyone, we can get rid of the `BloomFilterDeduplicator` and replace `Deduplicator` with the store-based one.

Part of #3280.

### DIFF
--- a/crates/matrix-sdk/src/event_cache/deduplicator.rs
+++ b/crates/matrix-sdk/src/event_cache/deduplicator.rs
@@ -31,17 +31,17 @@ use super::room::events::{Event, RoomEvents};
 /// positive or not
 ///
 /// [bloom filter]: https://en.wikipedia.org/wiki/Bloom_filter
-pub struct Deduplicator {
+pub struct BloomFilterDeduplicator {
     bloom_filter: Mutex<GrowableBloom>,
 }
 
-impl fmt::Debug for Deduplicator {
+impl fmt::Debug for BloomFilterDeduplicator {
     fn fmt(&self, formatter: &mut fmt::Formatter<'_>) -> fmt::Result {
         formatter.debug_struct("Deduplicator").finish_non_exhaustive()
     }
 }
 
-impl Deduplicator {
+impl BloomFilterDeduplicator {
     // Note: don't use too high numbers here, or the amount of allocated memory will
     // explode. See https://github.com/matrix-org/matrix-rust-sdk/pull/4231 for details.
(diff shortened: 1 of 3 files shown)


crates/matrix-sdk/src/event_cache/deduplicator.rs
```diff
@@ -18,11 +18,102 @@
...
+
+impl Deduplicator {
+    /// Create an empty deduplicator instance that uses an internal Bloom
+    /// filter.
+    ///
+    /// Such a deduplicator is stateful, with no initial known events, and it
+    /// will learn over time by using a Bloom filter which events are
+    /// duplicates or not.
+    ///
+    /// When the persistent storage is enabled by default, this constructor
```
Hywan, line 43: When the persistent storage of the event cache*

crates/matrix-sdk/src/event_cache/deduplicator.rs
```diff
@@ -18,11 +18,102 @@
...
+    pub fn new_memory_based() -> Self {
+        Self::InMemory(BloomFilterDeduplicator::new())
+    }
+
+    /// Create new store-based deduplicator that will run queries against the
+    /// store to find if any event is deduplicated or not.
+    ///
+    /// This deduplicator is stateless.
+    ///
+    /// When the persistent storage is enabled by default, this will become the
```
Hywan, line 54: … of the event cache*

crates/matrix-sdk/src/event_cache/deduplicator.rs
```diff
@@ -18,11 +18,102 @@
...
+            Deduplicator::InMemory(dedup) => Ok(dedup.filter_duplicate_events(events, room_events)),
+            Deduplicator::PersistentStore(dedup) => dedup.filter_duplicate_events(events).await,
+        }
+    }
+}
+
+/// A deduplication mechanism based on the persistent storage associated to the
+/// event cache.
+///
+/// It will use queries to the persistent storage to figure where events are
```
Hywan, line 79: where -> when

Hywan (approved):
Greeeeaaat! Can we have a couple of tests for the store based deduplicator? We already test `filter_duplicated_events` but one test to ensure the behaviour of `retain` works as expected would be nice :-). Thanks!

PR #4663, opened 2025-02-12
Memoize the `Oidc::account_management_url()` method

### LINKED ISSUE #4460: Implement pre-loading and cache for account management URLs
An API is exposed to get the external urls used to managed the account:
`fun accountUrl(action: AccountManagementAction?)`.

The applications are using this API to display the 2 settings entries "Manage account" and "Manage devices", if the respective returned URL are not null, and on click action open the webpage, see below:

image

But this API is performing network request and so can be slow to return, or can even fail when there is no network.
This will lead the application to:
- not show the entries in the root setting screen (no network)
- delay the display of the entries (slow network)

To solve this, the SDK should prefetch the URL for the actions `AccountManagementAction.Profile` and `AccountManagementAction.SessionsList`, in order to speed up the return of the method `fun accountUrl(action: AccountManagementAction?)`.
Not sure when the cache should expire, it's maybe safer to always fetch the URL at each client creation.

Note: mechanism used to pre-fetch data about the user, such as `avatarUrl` and `displayName` could be reused to solve this issue.

poljar:
This will hopefully fix #4460.

### DIFF
--- a/crates/matrix-sdk-common/src/ttl_cache.rs
+++ b/crates/matrix-sdk-common/src/ttl_cache.rs
@@ -1,0 +1,149 @@
+// Copyright 2025 The Matrix.org Foundation C.I.C.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! A TTL cache which can be used to time out repeated operations that might
+//! experience intermittent failures.
+
+use std::{borrow::Borrow, collections::HashMap, hash::Hash, time::Duration};
+
+use ruma::time::Instant;
+
+// One day is the default lifetime.
+const DEFAULT_LIFETIME: Duration = Duration::from_secs(24 * 60 * 60);
+
+#[derive(Debug)]
+struct TtlItem<V: Clone> {
+    value: V,
+    insertion_time: Instant,
+    lifetime: Duration,
+}
+
+impl<V: Clone> TtlItem<V> {
+    fn expired(&self) -> bool {
+        self.insertion_time.elapsed() >= self.lifetime
+    }
+}
+
+/// A TTL cache where items get removed deterministically in the `get()` call.
+#[derive(Debug)]
+pub struct TtlCache<K: Eq + Hash, V: Clone> {
+    lifetime: Duration,
+    items: HashMap<K, TtlItem<V>>,
+}
+
+impl<K, V> TtlCache<K, V>
+where
+    K: Eq + Hash,
+    V: Clone,
+{
+    /// Create a new, empty, [`TtlCache`].
+    pub fn new() -> Self {
+        Self { items: Default::default(), lifetime: DEFAULT_LIFETIME }
+    }
+
+    /// Does the cache contain an non-expired item with the matching key.
+    pub fn contains<Q>(&self, key: &Q) -> bool
+    where
+        K: Borrow<Q>,
+        Q: Hash + Eq + ?Sized,
+    {
+        let cache = &self.items;
+        let contains = if let Some(item) = cache.get(key) { !item.expired() } else { false };
+
+        contains
+    }
+
+    /// Add a single item to the cache.
+    pub fn insert(&mut self, key: K, value: V) {
+        self.extend([(key, value)]);
+    }
+
+    /// Extend the cache with the given iterator of items.
+    pub fn extend(&mut self, iterator: impl IntoIterator<Item = (K, V)>) {
+        let cache = &mut self.items;
+
+        let now = Instant::now();
+
+        for (key, value) in iterator {
+            let item = TtlItem { value, insertion_time: now, lifetime: self.lifetime };
+
+            cache.insert(key, item);
+        }
+    }
+
+    /// Remove the item that matches the given key.
+    pub fn remove<Q>(&mut self, key: &Q) -> Option<V>
+    where
+        K: Borrow<Q>,
+        Q: Hash + Eq + ?Sized,
+    {
+        self.items.remove(key.borrow()).map(|item| item.value)
+    }
+
+    /// Get the item that matches the given key, if the item has expired `None`
+    /// will be returned and the item will be evicted from the cache.
+    pub fn get<Q>(&mut self, key: &Q) -> Option<V>
+    where
+        K: Borrow<Q>,
+        Q: Hash + Eq + ?Sized,
+    {
+        // Remove all expired items.
+        self.items.retain(|_, value| !value.expired());
+        // Now get the wanted item.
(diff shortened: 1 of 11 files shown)


crates/matrix-sdk-common/src/ttl_cache.rs
```diff
@@ -0,0 +1,148 @@
...
+    {
+        // Remove all expired items.
+        self.items.retain(|_, value| !value.expired());
+        // Now get the wanted item.
+        self.items.get(key.borrow()).map(|item| item.value.clone())
+    }
+
+    /// Force the expiry of the given item, if it is present in the cache.
+    ///
+    /// This doesn't remove the item, it just marks it as expired.
```
bnjbvr, line 108: Why not remove it immediately? If it's expired, is there any point in keeping it in the mapping?

crates/matrix-sdk/src/authentication/oidc/mod.rs
```diff
@@ -644,6 +644,46 @@ impl Oidc {
...
+    /// [`Oidc::fetch_account_management_url()`] and cache the resulting URI
+    /// before returning it.
+    pub async fn account_management_url(
+        &self,
+        action: Option<AccountManagementActionFull>,
+    ) -> Result<Option<Url>, OidcError> {
+        let mut cache = self.client.inner.caches.account_management_url.lock().await;
+
+        let serialized_action = serde_json::to_string(&action)
+            .expect("We should be able to serialize the account management acction");
```
bnjbvr, line 672: nit: acction => action

crates/matrix-sdk/src/authentication/oidc/mod.rs
```diff
@@ -644,6 +644,46 @@ impl Oidc {
...
+    ///   management URL can be found in the [`VerifiedProviderMetadata`], or
+    ///   directly with [`Oidc::account_management_actions_supported()`].
+    ///
+    /// Returns `Ok(None)` if the URL was not found. Returns an error if the
+    /// request to get the provider metadata fails or the URL could not be
+    /// parsed.
+    ///
+    /// This method will cache the URL for a while, if the cache is not
+    /// populated it will internally call
+    /// [`Oidc::fetch_account_management_url()`] and cache the resulting URI
```
bnjbvr, line 663: how nitpicky of me, but there's an inconsistency between `URI` and `URL` in this paragraph :nerd_face:

crates/matrix-sdk/src/authentication/oidc/tests.rs
```diff
@@ -452,3 +456,48 @@ async fn test_insecure_clients() -> anyhow::Result<()> {
 
     Ok(())
 }
+
+#[async_test]
+async fn test_management_url_cache() {
+    let server = MatrixMockServer::new().await;
```
bnjbvr, line 462: Think you don't need the mock server at all, since mocking OIDC endpoints is so adhoc (my fault!), and can use `MockClientBuilder` to create a "serverless" client.

crates/matrix-sdk/src/authentication/oidc/tests.rs
```diff
@@ -452,3 +456,48 @@ async fn test_insecure_clients() -> anyhow::Result<()> {
...
+        .expect("We should be able to restore an OIDC session");
+
+    // The cache should not contain the entry.
+    assert!(!client
+        .inner
+        .caches
+        .account_management_url
+        .lock()
+        .await
+        .contains(&serde_json::to_string::<Option<AccountManagementAction>>(&None).unwrap()));
```
bnjbvr, line 486: light suggestion: consider a local variable for the atrocity in parenthesis? (Or could we use `Some()` variant of `AccountManagementAction`, so that we don't have to precise the generic type used in `to_string()`?)

bnjbvr (approved):
Makes sense, thanks!

crates/matrix-sdk-common/src/ttl_cache.rs
poljar: That method is for testing purposes mostly, perhaps I should make it`cfg(test)`.

It expires the item and then we check that a `get()` correctly removes it.

bnjbvr: Fair!

crates/matrix-sdk/src/authentication/oidc/mod.rs
```diff
@@ -644,6 +644,46 @@ impl Oidc {
...
+            Ok(Some(url))
+        } else {
+            let maybe_url = self.fetch_account_management_url(action).await?;
+
+            if let Some(url) = &maybe_url {
+                cache.insert(serialized_action, url.clone());
+            }
+
+            Ok(maybe_url)
+        }
```
zecakeh, line 684: If I am understanding this correctly, we are caching one URL per action? It seems wasteful to make one request per action since the base URL is shared by all actions.

We could just cache the base URL and call `build_account_management_url()` to get the full URL after that.

crates/matrix-sdk-common/src/ttl_cache.rs
```diff
@@ -0,0 +1,148 @@
...
+// limitations under the License.
+
+//! A TTL cache which can be used to time out repeated operations that might
+//! experience intermittent failures.
+
+use std::{borrow::Borrow, collections::HashMap, hash::Hash, time::Duration};
+
+use ruma::time::Instant;
+
+const DEFAULT_LIFETIME: Duration = Duration::from_secs(15 * 60);
```
zecakeh, line 22: Why only 15 minutes? It seems like with the given issue, the goal is to be able to have the URL with no connection, or a slow connection. The URL is unlikely to change so maybe a week would be fine, like for the server capabilities?

crates/matrix-sdk/src/authentication/oidc/mod.rs
poljar: https://github.com/matrix-org/matrix-rust-sdk/pull/4663/commits/768ec36376896384407fcfa42d968d5344402075

crates/matrix-sdk/src/authentication/oidc/tests.rs
poljar: https://github.com/matrix-org/matrix-rust-sdk/pull/4663/commits/768ec36376896384407fcfa42d968d5344402075

crates/matrix-sdk/src/authentication/oidc/tests.rs
poljar: https://github.com/matrix-org/matrix-rust-sdk/pull/4663/commits/768ec36376896384407fcfa42d968d5344402075

crates/matrix-sdk/src/authentication/oidc/mod.rs
poljar: I guess you're right, caching the whole metadata seems more reasonable: https://github.com/matrix-org/matrix-rust-sdk/pull/4663/commits/9575c6058154dc4d41173c1000b295eb490e876d.

zecakeh: Good, I was wondering if that would make sense after #4550 as well.

crates/matrix-sdk-common/src/ttl_cache.rs
poljar: No particular reason, just  seemed to be the minimum viable value to remove the flicker from the linked issue.

Let's put it to a day: d114feea80b74196bb72677b5f83fa846083b2f4.

PR #4664, opened 2025-02-13
refactor(oidc): Remove method to authorize arbitrary scope

zecakeh:
Only the scopes necessary during login are specified in MSC2967 now.


zecakeh:
The MSC was changed in https://github.com/matrix-org/matrix-spec-proposals/pull/2967/commits/f65aef3952ce0f9656abde251065239e5192ffae.

There used to be a section called "Insufficient privilege response" that used to say that we could have an error response about an "insufficient scope", and that we needed to request another authorization with the given scope, which is why that method was necessary.

Given that this section was removed, the only use case for the authorization endpoint right now is for logging in, so this method has no use case anymore.

poljar (approved):
Thanks for the clarification, makes sense.

Could you just get rid of the merge conflict?

zecakeh:
Done

PR #4665, opened 2025-02-13
fix(event cache store): shortcut when there's no duplicate events to check at all

bnjbvr:
Otherwise this causes a panic when repeating the events variable, when generating the SQL query below.

### DIFF
--- a/crates/matrix-sdk-sqlite/src/event_cache_store.rs
+++ b/crates/matrix-sdk-sqlite/src/event_cache_store.rs
@@ -627,6 +627,13 @@ impl EventCacheStore for SqliteEventCacheStore {
         room_id: &RoomId,
         events: Vec<OwnedEventId>,
     ) -> Result<Vec<OwnedEventId>, Self::Error> {
+        // If there's no events for which we want to check duplicates, we can return
+        // early. It's not only an optimization to do so: it's required, otherwise the
+        // `repeat_vars` call below will panic.
+        if events.is_empty() {
+            return Ok(Vec::new());
+        }
+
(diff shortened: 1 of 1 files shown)


crates/matrix-sdk-sqlite/src/event_cache_store.rs
```diff
@@ -627,6 +627,13 @@ impl EventCacheStore for SqliteEventCacheStore {
         room_id: &RoomId,
         events: Vec<OwnedEventId>,
     ) -> Result<Vec<OwnedEventId>, Self::Error> {
+        // If there's no events for which we want to check duplicates, we can return
+        // early. It's not only an optimization to do so: it's required, otherwise the
+        // `repeat_vars` call below will panic.
+        if events.is_empty() {
+            return Ok(Vec::new());
+        }
```
poljar, line 635: Don't we have a similar problem in the other direction? When we have too many events and the number of parameters doesn't fit into a single query.

poljar (approved):

crates/matrix-sdk-sqlite/src/event_cache_store.rs
Hywan: No because I'm using `chunk_large_query_over` already.

Hywan (approved):
Thanks for the fix. I think we should also update `repeat_vars` to not panic when there is nothing to repeat :-].

crates/matrix-sdk-sqlite/src/event_cache_store.rs
poljar: Ah Indeed.

PR #4668, opened 2025-02-13
bench: add a new benchmark to measure the time to fill a timeline with many events

bnjbvr:
Also introduces a new `profiling` profile to be used in benchmarking or profiling contexts.

cc @Hywan


Hywan (approved):
Niiiiice!

PR #4669, opened 2025-02-13
perf(timeline): make replacing replies much faster by indexing replies

bnjbvr:
`TimelineEventHandler::maybe_update_responses` was used after a message has been edited or redacted, so as to update replies to it. To find replies to a message, it would iterate over every single timeline item, find if it's an event item that's a message that's a reply, and look if the replied-to id matched the event that was redacted or edited.

This is super inefficient, because it may be doing work for nothing, if there's no such replies.

The alternative implemented in this patch is a classical memory / runtime speed tradeoff: every time we receive a new message that's a reply, we do insert it in a reverse mapping from `reply-target` to `set of replies`. Then, finding the replies to a message after it's edited/redacted becomes instant, because we can use this reverse mapping.

The reverse mapping will only be applied to events that are remote echo. This means that a local echo that's a reply may not get updated until it's remote echoed, if the original message is redacted/edited. We could implement that as well, if we wanted to, by reverting to a linear search in this case, as it should be less common; or storing a set of `TimelineEventItemId` in the reverse mapping (and maintaining it over time).

This reverse mapping will only grow over time. In theory, a redacted reply should cause the entry in the mapping to be removed to, but I haven't bothered with that: the code to update the reply can handle stale entries just fine, and will at worst log a warning in this case. We could try to do better and maintain it over time, though.

In the benchmark introduced in #4668, this lowers the total run time from 101ms to 65ms, a ~35% speedup.

### DIFF
--- a/crates/matrix-sdk-ui/src/timeline/event_handler.rs
+++ b/crates/matrix-sdk-ui/src/timeline/event_handler.rs
@@ -38,7 +38,7 @@ use ruma::{
         room::{
             encrypted::RoomEncryptedEventContent,
             member::RoomMemberEventContent,
-            message::{self, RoomMessageEventContent, RoomMessageEventContentWithoutRelation},
+            message::{Relation, RoomMessageEventContent, RoomMessageEventContentWithoutRelation},
         },
         AnyMessageLikeEventContent, AnySyncMessageLikeEvent, AnySyncStateEvent,
         AnySyncTimelineEvent, BundledMessageLikeRelations, EventContent, FullStateEventContent,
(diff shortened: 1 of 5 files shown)


Hywan (commented):
Good job on finding this bottleneck but I wonder… we already have a mapping from remote event to timeline items: Shouldn't we extend this by having a mapping from remote event **ID** to timeline items? That's basically what you do, more more generalised. It can be used in various other places.

crates/matrix-sdk-ui/src/timeline/event_handler.rs
```diff
@@ -1266,29 +1292,47 @@ impl<'a, 'o> TimelineEventHandler<'a, 'o> {
...
+            trace!("item has no replies");
+            return;
+        };
+
+        for reply_id in replies {
+            let Some(timeline_item_index) = items
+                .get_remote_event_by_event_id(reply_id)
+                .and_then(|meta| meta.timeline_item_index)
+            else {
+                warn!(%reply_id, "event not known as a timeline item in the timeline");
```
jmartinesp, line 1310: nit: 'as an item in the timeline' or 'as a timeline item in this timeline' maybe?

jmartinesp (approved):
LGTM, thanks for the optimisation!

PR #4670, opened 2025-02-14
Have the `RoomIdentityProvider` return processing changes when identities transition to `IdentityState::Verified` too

stefanceriu:
While implementing user verification on Element X I realized that the `IdentityStatusChanges` stream does not notify us when an user becomes `Verified`, which is a shame as it's perfect for powering app updates after verifying users (i.e. all the decorations we show throughout the app: room header, room details, room member list, room member profile etc.)

Had a look at the existing code, and while that seems completely intentional, I see no reason why we can't expand it's purview.


crates/matrix-sdk-crypto/CHANGELOG.md
```diff
@@ -21,6 +21,10 @@ All notable changes to this project will be documented in this file.
 
 - Room keys are not shared with unsigned dehydrated devices.
   ([#4551](https://github.com/matrix-org/matrix-rust-sdk/pull/4551))
+  
+- Have the `RoomIdentityProvider` return processing changes when identities transition
+  to `IdentityState::Verified` too.
+  ([#1234](https://github.com/matrix-org/matrix-rust-sdk/pull/????))
```
andybalaam, line 27: You have the PR number now, so can add it here

crates/matrix-sdk/src/room/identity_status_changes.rs
```diff
@@ -287,10 +287,16 @@ mod tests {
         // When Bob becomes verified
         t.verify_bob().await;
 
+        // Then we are notified about the unpinning part
```
andybalaam, line 290:
```suggestion
        // Then we are notified about Bob being verified
```

crates/matrix-sdk-crypto/src/identities/room_identity_state.rs
```diff
@@ -219,16 +227,11 @@ impl<R: RoomIdentityProvider> RoomIdentityState<R> {
...
-            // good -> good - don't report - no message needed in either case
+            // Changed the type of good - report so can change the message
             (Pinned, Verified) |
-            (Verified, Pinned) => {
-                // The state has changed, so we update it
-                self.set_state(user_id, new_state);
-                // but there is no need to report a change to the UI
-                None
-            }
+            (Verified, Pinned) => Some(self.set_state(user_id, new_state)),
```
andybalaam, line 234: This whole match has now become `if old_state == new_state` (right?) so let's just change it to that. All those comments were to explain something irritatingly complicated that you have now removed :-)

crates/matrix-sdk-crypto/src/identities/room_identity_state.rs
```diff
@@ -409,8 +412,14 @@ mod tests {
             identity_change(&mut room, user_id, IdentityState::Verified, false, false).await;
         let update = state.process_change(updates).await;
 
-        // Then we emit no update
-        assert_eq!(update, vec![]);
```
andybalaam, line 413: This test needs renaming.

andybalaam (approved):
Yep, this makes a lot of sense. I've added some comments, but happy for it to go in once they are addressed. Thanks!

crates/matrix-sdk-crypto/src/identities/room_identity_state.rs
stefanceriu: Ha, I was so busy tweaking it I didn't realise I can simplify the whole thing, love it!

PR #4673, opened 2025-02-15
refactor(oidc): Use the `GET /auth_metadata` Matrix endpoint

### LINKED ISSUE #4550: Update support of MSC2965 OAuth 2.0 Authorization Server Metadata discovery

zecakeh:
This is the method to get the server metadata in the latest draft of [MSC2965](https://github.com/matrix-org/matrix-spec-proposals/pull/2965).

We still keep the old behavior with `GET /auth_issuer` as fallback for now because it has wider server support.

There are some pre-main commit cleanups to simplify the main commit. This can be reviewed commit by commit.

The changes were tested with the oidc_cli example on beta.matrix.org.

Closes #4550.

### DIFF
--- a/bindings/matrix-sdk-ffi/src/client.rs
+++ b/bindings/matrix-sdk-ffi/src/client.rs
@@ -283,27 +283,18 @@ impl Client {
     /// Information about login options for the client's homeserver.
     pub async fn homeserver_login_details(&self) -> Arc<HomeserverLoginDetails> {
         let oidc = self.inner.oidc();
-        let (supports_oidc_login, supported_oidc_prompts) = match oidc
-            .fetch_authentication_issuer()
-            .await
-        {
-            Ok(issuer) => match &oidc.given_provider_metadata(&issuer).await {
-                Ok(metadata) => {
-                    let prompts = metadata
-                        .prompt_values_supported
-                        .as_ref()
-                        .map_or_else(Vec::new, |prompts| prompts.iter().map(Into::into).collect());
+        let (supports_oidc_login, supported_oidc_prompts) = match &oidc.provider_metadata().await {
+            Ok(metadata) => {
+                let prompts = metadata
+                    .prompt_values_supported
+                    .as_ref()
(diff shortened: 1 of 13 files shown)


poljar (approved):
Nice, thanks for keeping the old way around. Hopefully we'll be able to remove it relatively quickly.

Could you just get rid of the merge conflict?

bnjbvr:
When a PR is marked as rebase-and-merge when CI passes, it's usually conflicting with merge commits, because a merge commits makes rebase somewhat harder, if not impossible. In this case, switching to squash-and-merge would resolve it; would you like me to press the button, or would you like to rewrite the history so it doesn't include a merge commit?

zecakeh:
Sure, you can squash and merge, thanks!

bindings/matrix-sdk-ffi/src/client.rs
```diff
@@ -285,27 +285,18 @@ impl Client {
...
+                    .as_ref()
+                    .map_or_else(Vec::new, |prompts| prompts.iter().map(Into::into).collect());
+
+                (true, prompts)
+            }
             Err(error) => {
-                error!("Failed to fetch authentication issuer: {error}");
-                (false, Default::default())
+                error!("Failed to fetch OIDC provider metadata: {error}");
+                (true, Default::default())
```
bmarty, line 299: For the record: regression has been spotted and fixed here: https://github.com/matrix-org/matrix-rust-sdk/pull/4717

PR #4674, opened 2025-02-17
refactor(oidc): Align logout behavior with MSC4254

zecakeh:
Token revocation was split out from [MSC2964](https://github.com/matrix-org/matrix-spec-proposals/pull/2964) (commit https://github.com/matrix-org/matrix-spec-proposals/pull/2964/commits/c859c0bf2fb3fd5f9b7695b21eeec28623489c22, section called "Logout initiated from a Matrix client") to [MSC4254](https://github.com/matrix-org/matrix-spec-proposals/pull/4254).

According to MSC4254:

- OIDC RP-Initiated logout, which allowed to generate a URL to log out the user from the web UI, is now only listed as an alternative.
- Making a request to revoke a single token is enough to revoke both the access token and the refresh token.


poljar (changes requested):
Alright, sorry for the delay. I wanted to test this PR since we don't have integration tests with a real OIDC provider. Can confirm, it works with beta.matrix.org.

I left a small nit and could you please get rid of the merge conflict, then we can merge.

zecakeh:
I rebased on main to solve the conflicts

poljar (approved):

PR #4675, opened 2025-02-17
feat(common): Introduce `Chunk::lazy_previous` in the `LinkedChunk`

Hywan:
This patch introduces `Chunk::lazy_previous` which is a key feature to support lazy-loading of a `LinkedChunk`. When a chunk is loaded, if it is the first, it keeps in memory whether it has a previous chunk or not. Thus, it is possible to insert new chunk in front of the `LinkedChunk`, and `Update`s will correctly continue to link chunks between them (with `NewItemsChunk` and `NewGapChunk`).

Example, imagine the following chunks: [0] <-> [1] <-> [2]. If [2] is the only one being loaded. Then its previous chunk, [1], is loaded from the store (because [2]'s previous is [1] in the store). Then [1] is replaced by [3] and [4]. We get this: [4] <-> [3] <-> [1] <-> [2]. If the `Update::New*Chunk` for [4] doesn't contain a `previous`, the store is out of sync: in the store, [4] has no previous, but [0] still has [1] for its `next`.

With this `lazy_previous`, the links are correctly computed.

---

* Extracted from https://github.com/matrix-org/matrix-rust-sdk/pull/4632

### DIFF
--- a/crates/matrix-sdk-common/src/linked_chunk/mod.rs
+++ b/crates/matrix-sdk-common/src/linked_chunk/mod.rs
@@ -1175,6 +1175,12 @@ pub struct Chunk<const CAPACITY: usize, Item, Gap> {
     /// The previous chunk.
     previous: Option<NonNull<Chunk<CAPACITY, Item, Gap>>>,
 
+    /// If this chunk is the first one, and if the `LinkedChunk` is loaded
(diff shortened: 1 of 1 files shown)


bnjbvr (approved):
Seeing how it's used in #4632, makes sense!

PR #4677, opened 2025-02-17
perf(timeline): avoid linear searches for an event, and use a reverse mapping from event-id to index instead

bnjbvr:
Before this PR, in the timeline, to know if an event id had an attached event timeline item, we'd do a linear search over the timeline items, from the end to the back.

This PR changes it so that we:

- maintain a mapping of event-id -> index in the `all_remote_events` array
- then use that mapping instead of doing the linear search

Supposedly, this should improve performance; there's now a benchmark to confirm this.

### DIFF
--- a/crates/matrix-sdk-ui/src/timeline/algorithms.rs
+++ b/crates/matrix-sdk-ui/src/timeline/algorithms.rs
@@ -52,10 +52,13 @@ impl Deref for EventTimelineItemWithId<'_> {
     }
 }
 
-#[inline(always)]
-fn rfind_event_item_internal(
+/// Finds an item in the vector of `items` given a predicate `f`.
+///
+/// WARNING/FIXME: this does a linear scan of the items, so this can be slow if
+/// there are many items in the timeline.
+pub(super) fn rfind_event_item(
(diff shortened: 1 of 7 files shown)


bnjbvr:
Turns out, adding a mapping between event id and index (and maintaining it over time) is *slower* than doing linear searches in the timeline: from ~66.7ms to ~67.8ms to run the timeline benchmark I wrote last week.

I think this makes sense, because:

- maintaining the index may end up doing work that's linear in the # of events, for every added/removed item => becomes quadratic when adding a batch of N items.
- it involves hashing, which can be slow (I tried replacing the stdlib's hashmap with FxHashMap, and it was similar).
- the linear search relies on the locality property of those lookups: we usually look up an event that's close to the _end_ of the timeline. It's more common that some event relates to something else that's recent (so around the end of the timeline), rather than old (so around the start of the timeline). If those lookups were more random, this might be more of a win.

Even if the linear search is slower than the lookup in the hashmap (which would be nearly instant), maintaining the hashmap so it's correct entails more work, so overall it's not a win because of that.

As a result, I will likely close this PR :smiling_face_with_tear:

PR #4678, opened 2025-02-17
task(base): Update `EventCacheStore` to add `load_last_chunk` and `load_previous_chunk`

Hywan:
Extracted from https://github.com/matrix-org/matrix-rust-sdk/pull/4632 for the sake of easy review.

### DIFF
--- a/crates/matrix-sdk-common/src/linked_chunk/relational.rs
+++ b/crates/matrix-sdk-common/src/linked_chunk/relational.rs
@@ -468,6 +468,7 @@ where
 
 #[cfg(test)]
 mod tests {
+    use assert_matches::assert_matches;
     use ruma::room_id;
 
     use super::{ChunkIdentifier as CId, *};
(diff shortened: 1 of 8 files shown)


crates/matrix-sdk-sqlite/src/event_cache_store.rs
```diff
@@ -1948,6 +1948,172 @@ mod tests {
...
+                .await
+                .unwrap();
+
+            let previous_chunk =
+                store.load_previous_chunk(room_id, ChunkIdentifier::new(42)).await.unwrap();
+
+            assert!(previous_chunk.is_none());
+        }
+
+        // Case #3: there is two chunks.
```
bnjbvr, line 2126: nit: there *are*

crates/matrix-sdk-sqlite/src/event_cache_store.rs
```diff
@@ -1948,6 +1948,172 @@ mod tests {
...
+        }
+    }
+
+    #[async_test]
+    async fn test_load_previous_chunk() {
+        let room_id = room_id!("!r0:matrix.org");
+        let event = |msg: &str| make_test_event(room_id, msg);
+        let store = get_event_cache_store().await.expect("creating cache store failed");
+
+        // Case #1: no chunk at all, equivalent to having an inexistent
```
bnjbvr, line 2096: I think I've seen more often `nonexistent` than `inexistent` in English, or "lacking"/"missing".

crates/matrix-sdk-common/src/linked_chunk/relational.rs
```diff
@@ -345,13 +347,36 @@ where
...
 
         // Find the last chunk.
-        let Some(chunk_row) = self
-            .chunks
-            .iter()
-            .find(|chunk_row| chunk_row.room_id == room_id && chunk_row.next_chunk.is_none())
-        else {
-            // Chunk is not found.
-            return Ok((None, chunk_identifier_generator));
+        let chunk_row = match self.chunks.iter().try_fold(0, |mut number_of_chunks, chunk_row| {
```
bnjbvr, line 350: Can you use a plain loop instead of `try_fold`? I don't understand what this code does (what does it mean to continue with a value?), so I'd rather have a plain loop, so as to not confuse readers.

crates/matrix-sdk-common/src/linked_chunk/relational.rs
```diff
@@ -345,13 +347,36 @@ where
...
+                    return ControlFlow::Break((number_of_chunks, chunk_row));
+                }
+            }
+
+            ControlFlow::Continue(number_of_chunks)
+        }) {
+            // Chunk has been found, all good.
+            ControlFlow::Break((_, chunk_row)) => chunk_row,
+
+            // Chunk is not found and there is zero chunk for this room, this is consistent, all
```
bnjbvr, line 364: nit: zero is plural, so `there are zero chunks`

bnjbvr (changes requested):
Thanks! I'm a bit uneasy about the `try_fold` loop, and strongly recommends a plain old loop there instead.

Also, don't we need cycle detection later too (when loading another non-last chunk), because the cycle can involve chunks that are not the last one? `[ { prev: 1, id: 2, next : 1 }, { prev: 2, id: 1, next: 0 }, { prev: 1, id: 0, next: null } ]`

This makes me quite unconvinced that it's the DB's responsibility to run this consistency check:

- the error is stringly typed, so there's no way to cleanly report it to callers
- as demonstrated above, cycles can happen in other places,
- the DB isn't checked for consistency in general: there could be chunks in DB that aren't properly linked to the rest of the "main" line, aka disconnected components, and we're not checking for this anymore (before, it was the case because we'd load *all* chunks for a given room id).

Maybe we should think about this cycle detection a bit more, and separately (as a separate stack of patches…)

crates/matrix-sdk-sqlite/src/event_cache_store.rs
```diff
@@ -668,8 +670,23 @@ impl EventCacheStore for SqliteEventCacheStore {
                     )
                     .optional()?
                 else {
-                    // Chunk is not found.
-                    return Ok((None, chunk_identifier_generator));
+                    // Chunk is not found and there is zero chunk for this room, this is consistent, all
+                    // good.
+                    if number_of_chunks == 0 {
```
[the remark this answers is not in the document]
Hywan: This is a `let … else`, sadly `let else if` doesn't exist yet I think.

Hywan:
Replaced `try_fold` with a loop, fixed typos. Cycle detection for the last chunk is very particular as it can only be detected from the database. For other chunks, it happens in the `LinkedChunkBuilder` (see https://github.com/matrix-org/matrix-rust-sdk/pull/4632 which contains the patches). Changing the stringy errors to a beautiful `enum` should be addressed in another patch to avoid stacking PR for the eternity.

bnjbvr (approved):
Thanks :heart_hands:

PR #4679, opened 2025-02-17
refactor(oidc): Remove support for deserializing previous `UserSession` format

zecakeh:
The format changed 10 months ago and since it contains the tokens, it should have been re-serialized already in that time.

Afaict EX clients do not serialize that type, the bindings have their own `Session` type for that.


Hywan (approved):
🤞

PR #4685, opened 2025-02-19
feat(sqlite) Add an index on `events.event_id` and `.room_id`

Hywan:
This patch adds an index on `events.event_id` and on `events.room_id` so that queries on this column are faster. It mostly happens in the `Deduplicator`, which runs for every backwards pagination or sync.

This patch also updates the query in `filter_duplicated_events` to sort event by their `chunk_id` and `position` so that the results are constant, it helps when testing.

---

* Follow up of #4632

### DIFF
--- a/crates/matrix-sdk-sqlite/src/event_cache_store.rs
+++ b/crates/matrix-sdk-sqlite/src/event_cache_store.rs
@@ -62,7 +62,7 @@ mod keys {
 /// This is used to figure whether the SQLite database requires a migration.
 /// Every new SQL migration should imply a bump of this number, and changes in
 /// the [`run_migrations`] function.
-const DATABASE_VERSION: u8 = 4;
+const DATABASE_VERSION: u8 = 5;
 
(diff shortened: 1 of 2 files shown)


crates/matrix-sdk-sqlite/src/event_cache_store.rs
```diff
@@ -339,6 +339,16 @@ async fn run_migrations(conn: &SqliteAsyncConn, version: u8) -> Result<()> {
         .await?;
     }
 
+    if version < 5 {
+        conn.with_transaction(|txn| {
+            txn.execute_batch(include_str!(
+                "../migrations/event_cache_store/005_events_index_on_event_id.sql"
+            ))?;
+            txn.set_db_version(4)
```
bnjbvr, line 347: This should help with the tests :D

```suggestion
            txn.set_db_version(5)
```

Hywan: Glooups, thanks!

bnjbvr (approved):

PR #4687, opened 2025-02-19
refactor(oidc): Move `qrcode` module inside `oidc`

zecakeh:
Because it uses the OIDC API and it is accessed via `Oidc::login_with_qr_code()`.

The second commit makes `LoginWithQrCode` use `OidcBackend` instead of its own `OauthClient`. It will allow to share code when the server backend is switched to the oauth2 crate too. It will also allow to expose the device authorization grant directly in `Oidc`, if necessary.

### DIFF
--- a/crates/matrix-sdk/CHANGELOG.md
+++ b/crates/matrix-sdk/CHANGELOG.md
@@ -63,6 +63,8 @@ All notable changes to this project will be documented in this file.
     `NotSupported` variant and an `is_not_supported()` method to check if the
     error is due to the server not supporting OAuth 2.0.
   - `OidcError::MissingAuthenticationIssuer` was removed.
+- [**breaking**]: The `authentication::qrcode` module was moved inside
+  `authentication::oidc`, because it is only available through the `Oidc` API.
 
 ## [0.10.0] - 2025-02-04
 


crates/matrix-sdk/src/authentication/oidc/backend/mock.rs
```diff
@@ -240,4 +240,33 @@ impl OidcBackend for MockImpl {
...
+        _device_authorization_response: &oauth2::StandardDeviceAuthorizationResponse,
+    ) -> Result<
+        OidcSessionTokens,
+        oauth2::RequestTokenError<
+            oauth2::HttpClientError<reqwest::Error>,
+            oauth2::DeviceCodeErrorResponse,
+        >,
+    > {
+        unimplemented!()
+    }
```
poljar, line 271: To be honest, I don't really like this backend trait and that we're using it to test the OIDC submodule.

I wish we properly mocked things instead and added a `oidc()` subtype to the `MatrixMockServer`.

I know it's annoying to mock the OIDC stuff, but if the server changes under our feet we don't have any proof that it did so. Furthermore we don't have any integration tests that use MAS either.

I guess it's still a good step to centralize the logic instead of having this additional OIDC client thing, and then we can mock things using `wiremock` and `MatrixMockServer` and eventually get rid of this mock backend and trait.

poljar (approved):
Sorry I left some musings on our current test situation, not really anything that's blocking this PR.

crates/matrix-sdk/src/authentication/oidc/backend/mock.rs
zecakeh: I had planned to slightly refactor the tests before changing to the oauth2 crate, to reduce duplication and simplify the migration. Maybe I can do this instead then.

PR #4688, opened 2025-02-19
Add a new constructor to the InboundGroupSession struct

poljar:
As part of https://github.com/element-hq/element-meta/issues/2720 we're going to add a new fields to the `InboundGroupSession` struct.

A new constructor will be helpful to initialize the new field when we receive a room key as an `m.room_key` event.


andybalaam (approved):
Great, thanks!

PR #4689, opened 2025-02-19
refactor(event cache): simplify back-paginations

bnjbvr:
This is a stepping stone so as to be able to run back-paginations automatically from the event cache

If we kept the previous method `run_backwards()`, then we should've added extra `Send` / `Sync` bounds so that it can run from a spawned task owned by the `RoomEventCache`; this would have made it super unwieldy to pass data inside the closure passed as a parameter. As a matter of fact, most use case of `run_backwards()` should be equivalent to one of the two newly introduced methods that replace it:

- `run_backwards_until()`, that will rerun back-pagination until a certain number of events have been received,
- `run_backwards_once()`, that will run a single back-pagination (but keep the previous behavior of restarting if the timeline has been reset in the meanwhile).


jmartinesp (approved):
LGTM, thanks!

crates/matrix-sdk-ui/src/timeline/pagination.rs
```diff
@@ -75,38 +73,30 @@ impl super::Timeline {
...
-                        // As an exceptional contract: if there were no events in the response,
-                        // and we've not hit the start of the timeline, retry until we get
-                        // some events or reach the start of the timeline.
-                        return ControlFlow::Continue(());
+        loop {
+            match self.event_cache.pagination().run_backwards_once(batch_size).await {
+                Ok(outcome) => {
+                    // As an exceptional contract, restart the back-pagination if we received an
+                    // empty chunk.
+                    if outcome.reached_start || outcome.events.len() > 0 {
```
Hywan: I think clippy will tell you that `!outcome.events.is_empty()` is better:

```diff
- outcome.events.len() > 0
+ !outcome.events.is_empty()
```

crates/matrix-sdk/CHANGELOG.md
```diff
@@ -8,6 +8,13 @@ All notable changes to this project will be documented in this file.
 
 ### Features
 
+- [**breaking**]: The `RoomPagination::run_backwards` method has been removed, and replaced by two
+simpler methods:
+  - `RoomPagination::run_backwards_until()`, which will retrigger back-paginations until a certain
+  number of events have been received (and retry if the timeline has been reset in the background).
+  - `RoomPagination::run_backwards_once()`, which will run a single back-pagination (and retry if
+  the timeline has been reset in the background).
+  ([#4688](https://github.com/matrix-org/matrix-rust-sdk/pull/4688))
```
Hywan: s/4688/4689/

crates/matrix-sdk/src/event_cache/pagination.rs
```diff
@@ -131,7 +95,7 @@ impl RoomPagination {
         //
         // Knowing that, if all gaps have been resolved in the linked chunk, let's try
         // to load more events from the storage!
-        if self.get_or_wait_for_token(None).await.has_more().not() {
+        if !self.get_or_wait_for_token(None).await.has_more() {
```
Hywan, line 98: 😏

crates/matrix-sdk/tests/integration/event_cache.rs
```diff
@@ -518,33 +494,16 @@ async fn test_backpaginate_many_times_with_one_iteration() {
...
+        if !global_reached_start {
+            global_reached_start = outcome.reached_start;
+        }
+        global_events.extend(outcome.events);
         num_iterations += 1;
     }
 
     // I'll get all the previous events,
-    assert_eq!(num_iterations, 1); // in one iteration…
-    assert_eq!(num_paginations, 2); // … we get two paginations!
```
Hywan, line 547: I wonder if this test is not a duplicate of the previous one now.

crates/matrix-sdk/tests/integration/event_cache.rs
```diff
@@ -1748,25 +1685,21 @@ async fn test_lazy_loading() {
     // One more chunk will be loaded from the store. This new chunk contains 5
     // items. No need to reach the network.
     {
-        let pagination_outcome = room_event_cache
-            .pagination()
-            .run_backwards(10, until_at_least_one_event)
-            .await
-            .unwrap();
+        let outcome = room_event_cache.pagination().run_backwards_until(1).await.unwrap();
```
Hywan, line 1688: I was happy with `pagination_outcome` because it tells it's the outcome of the pagination 😛. Now we don't know what outcome it is about except by looking at the expression.

Hywan (approved):
Thanks for the patch. It simplifies the API. The previous API was already a simplification of something older that was even more complex. With time it appears we don't need this level of complexity. Nice!

crates/matrix-sdk/tests/integration/event_cache.rs
bnjbvr: The name is a bit too long, and makes statements overflow in multiple lines, when they could hold into a single one. "outcome" only refers to "pagination outcome" in the entire file. Also, each definition is at most 4 or 5 lines away from each use, so I found that having a short name is fine in this context.

PR #4690, opened 2025-02-19
Use the `DecryptedOlmV1Event` when encrypting to-device events

poljar:
This requires a bit of a hack, but I think the stronger typing and having the fields defined in a single place make this worth it.

### DIFF
--- a/crates/matrix-sdk-crypto/src/gossiping/machine.rs
+++ b/crates/matrix-sdk-crypto/src/gossiping/machine.rs
@@ -539,13 +539,15 @@ impl GossipMachine {
         device: &Device,
         content: SecretSendContent,
     ) -> OlmResult<Session> {
-        let event_type = content.event_type();
-        let (used_session, content) = device.encrypt(event_type, content).await?;
+        let event_type = content.event_type().to_owned();
+        let (used_session, content) = device.encrypt(&event_type, content).await?;
+
+        let encrypted_event_type = content.event_type().to_owned();
 
         let request = ToDeviceRequest::new(
             device.user_id(),
             device.device_id().to_owned(),
-            content.event_type(),
+            &encrypted_event_type,
             content.cast(),
         );
 
@@ -568,10 +570,12 @@ impl GossipMachine {
         let (used_session, content) =
             device.encrypt_room_key_for_forwarding(session.clone(), message_index).await?;
 
+        let event_type = content.event_type().to_owned();
+
         let request = ToDeviceRequest::new(
             device.user_id(),
             device.device_id().to_owned(),
-            content.event_type(),
+            &event_type,
             content.cast(),
         );
 
--- a/crates/matrix-sdk-crypto/src/identities/device.rs
+++ b/crates/matrix-sdk-crypto/src/identities/device.rs
@@ -425,18 +425,19 @@ impl Device {
         session: InboundGroupSession,
         message_index: Option<u32>,
     ) -> OlmResult<(Session, Raw<ToDeviceEncryptedEventContent>)> {
-        let (event_type, content) = {
+        let content: ForwardedRoomKeyContent = {
             let export = if let Some(index) = message_index {
                 session.export_at_index(index).await
             } else {
                 session.export().await
             };
-            let content: ForwardedRoomKeyContent = export.try_into()?;
 
-            (content.event_type(), content)
+            export.try_into()?
         };
 
-        self.encrypt(event_type, content).await
+        let event_type = content.event_type().to_owned();
+
+        self.encrypt(&event_type, content).await
     }
 
     /// Encrypt an event for this device.
@@ -832,9 +833,9 @@ impl DeviceData {
     ) -> OlmResult<MaybeEncryptedRoomKey> {
         let content = session.as_content().await;
         let message_index = session.message_index().await;
-        let event_type = content.event_type();
+        let event_type = content.event_type().to_owned();
 
-        match self.encrypt(store, event_type, content).await {
+        match self.encrypt(store, &event_type, content).await {
             Ok((session, encrypted)) => Ok(MaybeEncryptedRoomKey::Encrypted {
                 share_info: ShareInfo::new_shared(
                     session.sender_key().to_owned(),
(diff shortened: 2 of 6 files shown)


Hywan (approved):
Looks good to me, thanks!

crates/matrix-sdk-crypto/src/olm/session.rs
```diff
@@ -146,26 +150,59 @@ impl Session {
...
+        }
+
+        impl EventType for Content<'_> {
+            // This is a bit of a hack, usually we just define the `EVENT_TYPE` and use the
+            // default implementation of `event_type()`, we can't do this here
+            // because the event type isn't static.
+            //
+            // So we just leave EVENT_TYPE out. This works because the serialization is
+            // using `event_type()` and this type is contained to this function.
+            const EVENT_TYPE: &'static str = "";
```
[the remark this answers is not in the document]
poljar: It's used in all the other types that implement `EventType`. Only this one implementation is a special hack.

crates/matrix-sdk-crypto/src/types/events/mod.rs
```diff
@@ -42,7 +42,7 @@ pub trait EventType {
     ///
     /// **Note**: This should never be implemented manually, this takes the
     /// event type from the constant.
```
richvdh, line 44: This is no longer true? Or at least, the word "never" should be softened to "not usually" or something.

You could fix the comma splice while you're there.

crates/matrix-sdk-crypto/src/identities/device.rs
```diff
@@ -425,18 +425,20 @@ impl Device {
...
             let export = if let Some(index) = message_index {
                 session.export_at_index(index).await
             } else {
                 session.export().await
             };
             let content: ForwardedRoomKeyContent = export.try_into()?;
 
-            (content.event_type(), content)
+            content
         };
```
richvdh, line 437:
```suggestion

        let content: ForwardedRoomKeyContent = {
            let export = if let Some(index) = message_index {
                session.export_at_index(index).await
            } else {
                session.export().await
            };
            export.try_into()?
        };

File: crates/matrix-sdk-crypto/src/gossiping/machine.rs [hunk 4]
```diff
@@ -539,13 +539,15 @@ impl GossipMachine {
         device: &Device,
         content: SecretSendContent,
     ) -> OlmResult<Session> {
-        let event_type = content.event_type();
-        let (used_session, content) = device.encrypt(event_type, content).await?;
+        let event_type = content.event_type().to_owned();
+        let (used_session, content) = device.encrypt(&event_type, content).await?;
+
+        let event_type = content.event_type().to_owned();
```
richvdh, line 545: this line looks redundant?

crates/matrix-sdk-crypto/src/types/events/olm_v1.rs
```diff
@@ -162,7 +162,10 @@ impl AnyDecryptedOlmEvent {
 }
 
 /// An `m.olm.v1.curve25519-aes-sha2` decrypted to-device event.
-#[derive(Clone, Debug, Deserialize, Serialize)]
+///
+/// **Note**: This event will reserialize events lossy, unknown fields will be
```
richvdh, line 166:
```suggestion
/// **Note**: This event will reserialize events lossily; unknown fields will be
```

crates/matrix-sdk-crypto/src/types/events/olm_v1.rs
```diff
@@ -182,6 +185,52 @@ where
     pub content: C,
 }
 
+impl<C: EventType + Debug + Sized + Serialize> Serialize for DecryptedOlmV1Event<C> {
```
richvdh, line 188: A comment as to why this is needed/what it does differently from `derive(Serialize)` might be helpful.

crates/matrix-sdk-crypto/src/olm/session.rs
```diff
@@ -146,26 +150,59 @@ impl Session {
...
+        #[derive(Debug)]
+        struct Content<'a> {
+            event_type: &'a str,
+            content: Raw<Value>,
+        }
+
+        impl EventType for Content<'_> {
+            // This is a bit of a hack, usually we just define the `EVENT_TYPE` and use the
+            // default implementation of `event_type()`, we can't do this here
+            // because the event type isn't static.
```
richvdh, line 162: multi-way comma splice!

```suggestion
            // This is a bit of a hack: usually we just define the `EVENT_TYPE` and use the
            // default implementation of `event_type()`. We can't do this here
            // because the event type isn't static.
```

crates/matrix-sdk-crypto/src/olm/session.rs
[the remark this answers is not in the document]
richvdh: I assume Ivan means we should add this to the documentation of this implementation, to which, +1.

```suggestion
            // We have to provide `EVENT_TYPE` to conform to the `EventType` trait, but
            // don't actually use it, so we just leave it empty.
            //
            // This works because the serialization uses `event_type()` and this type is contained
            // to this function.
            const EVENT_TYPE: &'static str = "";
```

richvdh (approved):
LGTM in general, a few nits

crates/matrix-sdk-crypto/src/gossiping/machine.rs
poljar: At first glance one would think so. But the first `event_type` is for `content` while the second `event_type` is for `content` after the `device.encrypt()` step.

Perhaps renaming the second `content` to `encrypted_content` might make this clearer.

richvdh: oh right, yes. Some different names might help.

crates/matrix-sdk-crypto/src/olm/session.rs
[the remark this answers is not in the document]
Hywan: Yes, this. Thanks for clarifying my words.

crates/matrix-sdk-crypto/src/identities/device.rs
poljar: Wouldn't let me commit the suggestion, but applied manually in 042925eb435e49ed10ff7a090e84f73e00dfb271.

crates/matrix-sdk-crypto/src/types/events/mod.rs
poljar: Did so in 49d119a397b53e45b81197904c09a83687f503b4.

PR #4691, opened 2025-02-19
refactor(event cache): simplify flow when deciding to resolve a gap

bnjbvr:
The code before this patch was doing this:

- look if there's any prev-batch token available right now, aka look if there's a gap in the in-memory linked chunk
- look at the first chunk; if it's a gap, return to the caller so it resolves it

The check is done twice at two different levels, which is confusing. Instead, this patch rewrites it so that the chunk is done only in `load_more_event_backwards()`.

Note this is also correct for the case storage is disabled; in this case, we early return and always try to resolve the gap anyways.


Hywan (approved):
Fair enough, thanks for the refactoring!

PR #4695, opened 2025-02-19
chore(common): Remove `LinkedChunkBuilderTest`

Hywan:
This patch adds the new `from_all_chunks` function in the
`linked_chunk::lazy_loader` module. It is only used for testing
purposes. It aims at replacing `LinkedChunkBuilderTest`.
Why? Because `from_all_chunks` uses `from_last_chunk` and
`insert_new_first_chunk`: if `from_all_chunks` is able to find all
errors that `LinkedChunkBuilderTest` finds, it's a bingo. Transitively,
it proves that `from_last_chunk` and `insert_new_first_chunk` are
correct!

The other patches replaces `LinkedChunkBuilderTest` and remove it.

---

* Follow up of #4632

### DIFF
--- a/crates/matrix-sdk-common/src/linked_chunk/lazy_loader.rs
+++ b/crates/matrix-sdk-common/src/linked_chunk/lazy_loader.rs
@@ -196,6 +196,65 @@ where
     Ok(())
 }
 
+/// A pretty inefficient, test-only, function to rebuild a full `LinkedChunk`.
+#[doc(hidden)]
+pub fn from_all_chunks<const CAP: usize, Item, Gap>(
+    mut chunks: Vec<RawChunk<Item, Gap>>,
+) -> Result<Option<LinkedChunk<CAP, Item, Gap>>, LazyLoaderError>
+where
+    Item: Clone,
+    Gap: Clone,
+{
+    if chunks.is_empty() {
+        return Ok(None);
+    }
+
+    // Sort by `next` so that the search for the next chunk is faster (it should
+    // come first). The chunk with the biggest next chunk identifier comes first.
+    // Chunk with no next chunk comes last.
+    chunks.sort_by(|a, b| b.next.cmp(&a.next));
+
+    let last_chunk = chunks
+        .pop()
+        // SAFETY: `chunks` is guaranteed to not be empty, `pop` cannot fail.
+        .expect("`chunks` is supposed to not be empty, we must be able to `pop` an item");
+    let last_chunk_identifier = last_chunk.identifier;
+    let chunk_identifier_generator =
+        ChunkIdentifierGenerator::new_from_previous_chunk_identifier(last_chunk_identifier);
+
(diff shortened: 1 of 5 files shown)


crates/matrix-sdk-common/src/linked_chunk/lazy_loader.rs
```diff
@@ -196,6 +196,59 @@ where
...
+    if chunks.is_empty() {
+        return Ok(None);
+    }
+
+    // Sort by `next` so that the search for the next chunk is faster (it should
+    // come first). The chunk with the biggest next chunk identifier comes first.
+    // Chunk with no next chunk comes last.
+    chunks.sort_by(|a, b| b.next.cmp(&a.next));
+
+    let last_chunk = chunks.pop().unwrap();
```
poljar, line 214: I know that this is meant to be a test function, but it isn't marked as `cfg(test)` and those unneeded unwraps bother me.

Why not replace the `is_empty()` check with a:

```rust
let Some(last_chunk) = vec.last() else {
     return Ok(None)
};
```

crates/matrix-sdk-common/src/linked_chunk/lazy_loader.rs
```diff
@@ -196,6 +196,59 @@ where
...
+    let last_chunk_identifier = last_chunk.identifier;
+    let chunk_identifier_generator =
+        ChunkIdentifierGenerator::new_from_previous_chunk_identifier(last_chunk_identifier);
+
+    let mut linked_chunk = from_last_chunk(Some(last_chunk), chunk_identifier_generator)?.unwrap();
```
poljar, line 219: Not sure if you can get rid of this one, but at least put an `expect()` there instead.

poljar (approved):
Makes sense, I left a couple of nits.

crates/matrix-sdk-common/src/linked_chunk/lazy_loader.rs
Hywan: It's not behind a `cfg(test)` because it's used by other crates :-/.

I can do a `pop() else` yup.

crates/matrix-sdk-common/src/linked_chunk/lazy_loader.rs
Hywan: Sure, let's add an `expect`. For the record, it's purely used for testing purposes, but I share your concerns.

Hywan: I've changed for a `else { return Ok(None); }` here.

crates/matrix-sdk-common/src/linked_chunk/lazy_loader.rs
Hywan: Actually the `pop` cannot fail because `chunks` is guaranteed to not be empty. I am adding a `// SAFETY` comment.

PR #4699, opened 2025-02-20
refactor(oidc): Align `OidcAuthCodeUrlBuilder` behavior with next-gen auth MSCs

zecakeh:
The first commit removes support for Pushed Authorization Requests. PAR is a small optimization that sends the parameters via a request first, and returns an ID that we used in the authorization URL instead of the parameters, so it only makes the authorization URL smaller by reducing the size of the query part. It is no mentioned in the next-gen auth MSCs and it is not supported by the oauth2 crate so let's remove it already.

The second commit removes most of the optional parameters of the authorization URL. All those parameters are specified in OpenID Connect, but the next-gen auth MSCs were downgraded to OAuth 2.0 which doesn't have those parameters.

Only the following parameters are defined in the MSCs:
- `prompt=create` is defined in [MSC2964](https://github.com/sandhose/matrix-spec-proposals/blob/msc/sandhose/oauth2-profile/proposals/2964-oauth2-profile.md#user-registration)
- `login_hint=mxid:@user:server.name` is defined in MSC4198

### DIFF
--- a/crates/matrix-sdk/CHANGELOG.md
+++ b/crates/matrix-sdk/CHANGELOG.md
@@ -80,6 +80,10 @@ simpler methods:
   - Only one request is made to revoke the access token, since the server is
     supposed to revoke both the access token and the associated refresh token
     when the request is made.
+- [**breaking**]: Remove most of the parameter methods of
+  `OidcAuthCodeUrlBuilder`, since they were parameters defined in OpenID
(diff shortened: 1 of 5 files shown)


poljar (approved):
Nice, a lot of code gone. Tested that our login still works. Could you please just get rid of the merge conflict.

zecakeh:
Rebased on main to resolve the conflict.

poljar (approved):

PR #4701, opened 2025-02-20
test: rename some MatrixMockServer matcher functions

bnjbvr:
The `.from()`, `.with_delay()` and `.limit()` functions are not very explicit about what they did, and the `with_delay` one in particular led me to think that it would introduce a delay *in the response*, while it indicated a delay was expected as part of the matched URL.

Instead, this patch proposes to prefix all matchers with `match_`, to make it clearer that… they will match the incoming query: match_from, match_delay, match_limit.

Thanks to this change, the `RoomMessagesResponseTemplate` can be renamed from `delayed` to `with_delay`, which was my original intent, before I noticed another `with_delay` with totally different semantics.

### DIFF
--- a/crates/matrix-sdk/src/test_utils/mocks.rs
+++ b/crates/matrix-sdk/src/test_utils/mocks.rs
@@ -1332,6 +1332,8 @@ impl<'a> MockEndpoint<'a, RoomSendEndpoint> {
 
     /// Ensures the event was sent as a delayed event.
     ///
+    /// See also [the MSC](https://github.com/matrix-org/matrix-spec-proposals/pull/4140).
+    ///
     /// Note: works with *any* room.
     ///
     /// # Examples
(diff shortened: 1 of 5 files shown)


crates/matrix-sdk/src/test_utils/mocks.rs
```diff
@@ -1387,7 +1387,7 @@ impl<'a> MockEndpoint<'a, RoomSendEndpoint> {
     /// assert_eq!("$some_id", response.delay_id);
     /// # anyhow::Ok(()) });
     /// ```
-    pub fn with_delay(self, delay: Duration) -> Self {
+    pub fn match_delay(self, delay: Duration) -> Self {
```
poljar, line 1390: Perhaps `match_delayted_event()` would be an even better name, also the docs should really link to the MSC.

crates/matrix-sdk/src/test_utils/mocks.rs
```diff
@@ -1615,7 +1615,7 @@ impl<'a> MockEndpoint<'a, RoomSendStateEndpoint> {
     ///
     /// # anyhow::Ok(()) });
     /// ```
-    pub fn with_delay(self, delay: Duration) -> Self {
+    pub fn match_delay(self, delay: Duration) -> Self {
```
poljar, line 1618: Same argument here.

poljar (approved):

PR #4703, opened 2025-02-20
feat(event cache): automatically shrink a room's linked chunk when all subscribers are gone

bnjbvr:
On top of #4694.

Every subscriber to a `RoomEventCache` now gets a shared reference to a RAII `AutoShrinkLinkedChunk` guard; when the last one is dropped, it's spawning a task (from the `Drop` impl — _sigh to no async drop_) that will shrink the linked chunk to the last chunk.

This is a poor man's implementation that doesn't cause too much operational overhead, in particular, because it doesn't require a new task listening to commands to shrink the linked chunk. (This new task should then live in the main `EventCache` instance, so as not to have one such task per `RoomEventCache`, aka per room.)

In the worst case, we can tweak the implementation and (mostly) keep the test, because it's testing the behavior we'd want here.


crates/matrix-sdk/src/event_cache/room/mod.rs
```diff
@@ -55,6 +63,84 @@ impl fmt::Debug for RoomEventCache {
...
+
+impl Drop for AutoShrinkLinkedChunk {
+    fn drop(&mut self) {
+        let room_event_cache = self.room_event_cache.take().expect("drop called multiple times");
+
+        // I hear you from the future: "but, spawning a detached task in a drop
+        // implementation is real bad! Maybe there will be multiple shrinks
+        // happening at the same time, and that's bad!". However, this can't
+        // happen, because the whole `state` variable is guarded by a fair lock,
+        // which will run queries in the order they happen. Should be fine™.
```
Hywan, line 82: You forgot to explain why you need to `spawn` here. It's because `drop` cannot be async, and you're doing async operations.

bnjbvr: Yep, introduced a mechanism with one (1) task listening forever in the event cache, that should be more sound.

crates/matrix-sdk/src/event_cache/mod.rs
```diff
@@ -129,6 +129,9 @@ pub struct EventCacheDropHandles {
 
     /// Task that listens to updates to the user's ignored list.
     ignore_user_list_update_task: JoinHandle<()>,
+
+    /// The task used to automatically shrink the linked chunks.
+    auto_shrink_linked_chunk_tasks: JoinHandle<()>,
```
Hywan, line 134:
```diff
- auto_shrink_linked_chunk_tasks
+ auto_shrink_linked_chunk_task
```

crates/matrix-sdk/src/event_cache/mod.rs
```diff
@@ -213,7 +218,19 @@ impl EventCache {
                 client.subscribe_to_ignore_user_list_changes(),
             ));
 
-            Arc::new(EventCacheDropHandles { listen_updates_task, ignore_user_list_update_task })
+            let (tx, rx) = mpsc::channel(32);
```
Hywan, line 221: Does it mean we are bounded to 32 subscribers? It seems a bit low. Can we increase to 256?

crates/matrix-sdk/src/event_cache/room/mod.rs
```diff
@@ -55,6 +64,62 @@ impl fmt::Debug for RoomEventCache {
...
+            // We were the last instance of the listener; let the auto-shrinker know by
+            // notifying it of our room id.
+
+            let mut room_id = self.room_id.clone();
+
+            // Try to send without waiting for channel capacity, and restart in a spin-loop
+            // if it failed; the channel shouldn't be super busy in general, so
+            // this should resolve quickly enough.
+
+            while let Err(err) = self.auto_shrink_sender.try_send(room_id) {
```
Hywan, line 97: Hmm, this loop worries me a lot. Can we panic if it executes too many times? The idea is to avoid having an infinite loop (in practise it should not happen, but better being safe than sorry).

crates/matrix-sdk/src/event_cache/room/mod.rs
```diff
@@ -584,6 +669,9 @@ mod private {
         /// the first time we try to run backward pagination. We reset
         /// that upon clearing the timeline events.
         pub waited_for_initial_prev_token: bool,
+
+        /// An atomic count of the current number of listeners.
```
Hywan, line 673: Listeners… of the `RoomEventCache`.

Hywan (approved):
Excellent :-]. Thank you!

crates/matrix-sdk/src/event_cache/mod.rs
bnjbvr: No, it's not exactly the number of subscribers, it's the number of entries in the channel. In the worst case, it's the number of subscribers who try to send a notification at the same time, but with he `try_send` mechanism used in the drop impl, that should be sufficient to handle them all in order.

crates/matrix-sdk/src/event_cache/room/mod.rs
bnjbvr: ok, will add a large guard, like 1024 attempts

PR #4705, opened 2025-02-21
fix(client): Add `handle_verification_events` to `BaseClient`

jmartinesp:
<del>This can be used to ignore verification requests in this sliding sync instance, preventing issues found where several sliding sync instances with the same client process events simultaneously and re-process the same verification request events during their initial syncs.</del> We went with a solution based on `BaseClient` instead.

This is a blocker for user verification on Android clients, since both the normal sync and the one initiated to get the event for a notification will process the same verification request events and end up with conflicts, making the verification flow fail most of the time.

Signed-off-by:

### DIFF
--- a/crates/matrix-sdk-base/src/client.rs
+++ b/crates/matrix-sdk-base/src/client.rs
@@ -124,6 +124,10 @@ pub struct BaseClient {
     /// The trust requirement to use for decrypting events.
     #[cfg(feature = "e2e-encryption")]
     pub decryption_trust_requirement: TrustRequirement,
+
+    /// If the client should handle verification events received when syncing.
+    #[cfg(feature = "e2e-encryption")]
+    pub handle_verification_events: bool,
 }
 
 #[cfg(not(tarpaulin_include))]
@@ -160,6 +164,8 @@ impl BaseClient {
             room_key_recipient_strategy: Default::default(),
             #[cfg(feature = "e2e-encryption")]
             decryption_trust_requirement: TrustRequirement::Untrusted,
+            #[cfg(feature = "e2e-encryption")]
+            handle_verification_events: true,
         }
     }
 
@@ -169,6 +175,7 @@ impl BaseClient {
     pub async fn clone_with_in_memory_state_store(
         &self,
         cross_process_store_locks_holder_name: &str,
+        handle_verification_events: bool,
     ) -> Result<Self> {
         let config = StoreConfig::new(cross_process_store_locks_holder_name.to_owned())
             .state_store(MemoryStore::new());
@@ -189,6 +196,7 @@ impl BaseClient {
             room_info_notable_update_sender: self.room_info_notable_update_sender.clone(),
             room_key_recipient_strategy: self.room_key_recipient_strategy.clone(),
             decryption_trust_requirement: self.decryption_trust_requirement,
+            handle_verification_events,
         };
 
         if let Some(session_meta) = self.session_meta().cloned() {
@@ -207,6 +215,7 @@ impl BaseClient {
     pub async fn clone_with_in_memory_state_store(
         &self,
         cross_process_store_locks_holder: &str,
+        _handle_verification_events: bool,
     ) -> Result<Self> {
         let config = StoreConfig::new(cross_process_store_locks_holder.to_owned())
             .state_store(MemoryStore::new());
@@ -339,6 +348,10 @@ impl BaseClient {
         event: &AnySyncMessageLikeEvent,
         room_id: &RoomId,
     ) -> Result<()> {
+        if !self.handle_verification_events {
+            return Ok(());
+        }
+
         if let Some(olm) = self.olm_machine().await.as_ref() {
             olm.receive_verification_event(&event.clone().into_full_event(room_id.to_owned()))
                 .await?;
--- a/crates/matrix-sdk-base/CHANGELOG.md
+++ b/crates/matrix-sdk-base/CHANGELOG.md
@@ -15,6 +15,9 @@ All notable changes to this project will be documented in this file.
   - `EventCacheStoreMedia` has a new method `last_media_cleanup_time_inner`
   - There are new `'static` bounds in `MediaService` for the media cache stores
 - `event_cache::store::MemoryStore` implements `Clone`.
+- `BaseClient` now has a `handle_verification_events` field which is `true` by 
+  default and can be negated so the `NotificationClient` won't handle received 
+  verification events too, causing errors in the `VerificationMachine`.
 
 ## [0.10.0] - 2025-02-04
 
--- a/crates/matrix-sdk/CHANGELOG.md
+++ b/crates/matrix-sdk/CHANGELOG.md
@@ -29,6 +29,9 @@ simpler methods:
   [BCP 195](https://datatracker.ietf.org/doc/bcp195/).
   ([#4647](https://github.com/matrix-org/matrix-rust-sdk/pull/4647))
 - Add `Room::report_room` api. ([#4713](https://github.com/matrix-org/matrix-rust-sdk/pull/4713))
+- `Client::notification_client` will create a copy of the existing `Client`, but now it'll make sure 
+  it doesn't handle any verification events to avoid an issue with these events being received and 
+  processed twice if `NotificationProcessSetup` was `SingleSetup`.
 
 ### Bug Fixes
 
--- a/crates/matrix-sdk/src/client/mod.rs
+++ b/crates/matrix-sdk/src/client/mod.rs
@@ -2422,7 +2422,7 @@ impl Client {
                 self.inner.http_client.clone(),
                 self.inner
                     .base_client
-                    .clone_with_in_memory_state_store(&cross_process_store_locks_holder_name)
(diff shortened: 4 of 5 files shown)


crates/matrix-sdk-base/src/client.rs
```diff
@@ -496,13 +498,18 @@ impl BaseClient {
                                 SyncMessageLikeEvent::Original(original_event),
                             ) => match &original_event.content.msgtype {
                                 MessageType::VerificationRequest(_) => {
-                                    Box::pin(self.handle_verification_event(e, room.room_id()))
-                                        .await?;
+                                    if !ignore_verification_requests {
+                                        Box::pin(self.handle_verification_event(e, room.room_id()))
+                                            .await?;
+                                    }
```
poljar, line 504: Shouldn't we ignore all verification events? I guess the crypto crate already does so if it doesn't receive an `m.key.verification.request` event, but I think it'll log a bunch of warnings in that case.

poljar (changes requested):
This feels a bit like a hack, though sadly I don't have any better ideas how to deal with this problem.

I'm wondering if we should ignore all verification events. Additionally could you write a test that the notification client won't create `Verification` objects?

An integration test would probably be the easiest to write.

crates/matrix-sdk-base/src/client.rs
jmartinesp: Isn't that what I'm doing below? I'm not sure I fully understand the underlying logic, but I think this branch takes the verification requests and the one below any other verification event.

poljar: Ah, how did I miss that one. You are indeed right.

Can we then rename the flag and setting? Since we indeed don't just ignore the requests `ignore_verification_events` might make more sense.

crates/matrix-sdk-base/src/client.rs
```diff
@@ -496,13 +498,18 @@ impl BaseClient {
                                 SyncMessageLikeEvent::Original(original_event),
                             ) => match &original_event.content.msgtype {
                                 MessageType::VerificationRequest(_) => {
-                                    Box::pin(self.handle_verification_event(e, room.room_id()))
-                                        .await?;
+                                    if !ignore_verification_requests {
+                                        Box::pin(self.handle_verification_event(e, room.room_id()))
```
Hywan, line 502: Shouldn't we use this `ignore_verification_requests` inside the `handle_verification_event` method instead of here?

crates/matrix-sdk-base/src/sliding_sync.rs
```diff
@@ -124,11 +124,14 @@ impl BaseClient {
...
     /// * `previous_events_provider` - Timeline events prior to the current
     ///   sync.
+    /// * `ignore_verification_requests` - Whether verification requests found
+    ///   in the response should be ignored.
     #[instrument(skip_all, level = "trace")]
     pub async fn process_sliding_sync<PEP: PreviousEventsProvider>(
         &self,
         response: &http::Response,
         previous_events_provider: &PEP,
+        ignore_verification_requests: bool,
```
Hywan, line 134: Can't we get this information from the `BaseClient` instead of passing an argument everywhere?

Hywan (dismissed):

crates/matrix-sdk-base/src/client.rs
jmartinesp: To me it feels a bit weird that a method called `handle_verification_event` gets a parameter telling it if it should ignore the event, but I can move it there if you'd rather centralise it. However, this still needs to get passed as a parameter since `BaseClient` shouldn't hold this value.

poljar: What's the reasoning for not putting the parameter into the `BaseClient`?

jmartinesp: https://github.com/matrix-org/matrix-rust-sdk/pull/4705#discussion_r1967996296

crates/matrix-sdk-base/src/sliding_sync.rs
poljar: Ah, here is the reasoning about the parameter, I'm not sure this is related to sliding sync, it has more to do with the notification client.

While our notification client exclusively uses sliding sync, one could imagine one that uses classic sync.

poljar: Which one? The `BaseClient`? We already do so:

https://github.com/matrix-org/matrix-rust-sdk/blob/19df945155cefe80dada6c18e341cd5d1fe5727d/crates/matrix-sdk/src/client/mod.rs#L2423-L2426

We could set the flag there, then the public API of the `SlidingSyncBuilder` wouldn't need to be changed.

jmartinesp: Oh ok, that makes sense! Let me try that instead then.

jmartinesp: Ok, so I think the latest changes should handle this the way you and @Hywan wanted it to. Could you check?

crates/matrix-sdk/tests/integration/client.rs
```diff
@@ -1249,7 +1247,7 @@ async fn test_rooms_stream() {
     assert!(client.get_room(room_id_3).is_some());
 
     // We receive 3 diffs…
-    assert_let!(Some(diffs) = rooms_stream.next().await);
+    assert_let!(Some(diffs) = futures_util::StreamExt::next(&mut rooms_stream).await);
```
poljar, line 1250: What's this about? Seems to be unrelated?

crates/matrix-sdk-crypto/src/verification/machine.rs
```diff
@@ -389,18 +389,22 @@ impl VerificationMachine {
...
-                            // This won't produce an outgoing content
-                            let _ = sas.receive_any_event(event.sender(), &content);
+                    if !verification.is_cancelled() {
+                        match verification {
+                            Verification::SasV1(sas) => {
+                                // This won't produce an outgoing content
+                                let _ = sas.receive_any_event(event.sender(), &content);
+                            }
+                            #[cfg(feature = "qrcode")]
+                            Verification::QrV1(qr) => qr.receive_cancel(event.sender(), c),
```
poljar, line 406: What is this about? This doesn't seem to be directly related to the PR?

crates/matrix-sdk-base/src/client.rs
```diff
@@ -169,6 +175,7 @@ impl BaseClient {
     pub async fn clone_with_in_memory_state_store(
         &self,
         cross_process_store_locks_holder_name: &str,
+        handle_verification_events: bool,
```
poljar, line 178: I don't think that this necessarily needed to be an argument for this function, after all, we're only using it with the `false` parameter.

I guess some other unlikely user of this method might want to set it to `true`.

poljar (changes requested):
Yeah, this aligns what I had in mind, I think it's also what @Hywan suggested.

There are some unrelated changes that have slipped through, please get rid of them or at least justify them and if possible put them in a separate commit.

crates/matrix-sdk/tests/integration/client.rs
jmartinesp: I made this change because the compiler was complaining about not being able to decide on which import to use for some reason and I should make it explicit. The `use` above didn't seem to help. I'll try reverting these.

crates/matrix-sdk-crypto/src/verification/machine.rs
jmartinesp: Oh damn, these were part of some experiments I did to check whether the issues could be fixed this way and I was *sure* I had reverted them. Apparently not, sorry.

crates/matrix-sdk-base/src/client.rs
jmartinesp: So should I get rid of the parameter? I left it both because handling verifications doesn't have much to do with cloning the client so it's a bit weird that we assume it must be false, and also in case some user of the SDK had any use case in mind that needed the parameter to be true.

poljar: Let's leave it in.

It's very likely that this function is only used to create a notification client object, then we could rename the method to `notification_client()`.

I'm not completely certain that this is true, so let's leave it as it is.

crates/matrix-sdk/tests/integration/client.rs
jmartinesp: Yep, not needed anymore. I don't know what happened when I tried this before.

jmartinesp: Removed in https://github.com/matrix-org/matrix-rust-sdk/pull/4705/commits/deaaca1aef4448add7ed1cce54db35bb97c6b610

crates/matrix-sdk-crypto/src/verification/machine.rs
jmartinesp: Removed in https://github.com/matrix-org/matrix-rust-sdk/pull/4705/commits/deaaca1aef4448add7ed1cce54db35bb97c6b610

poljar (approved):
Alright, please just add a changelog entry and feel free to rebase to get a nicer git history.

Looks good.

PR #4706, opened 2025-02-21
fix: Duplicated events are always removed

Hywan:
When a duplicated event is found, sometimes we keep the older one and the new one is dropped, but sometimes the older one is removed and the new is kept. So we need to be able to remove an event.

However, since https://github.com/matrix-org/matrix-rust-sdk/pull/4662 and https://github.com/matrix-org/matrix-rust-sdk/pull/4632, this can fail. Why? Because we can find an duplicated event in the store, and that is not loaded in the in-memory `LinkedChunk`. What happens in this case? We simply give up! See the `error!` case?

https://github.com/matrix-org/matrix-rust-sdk/blob/bdf5fad992dc5c0d412a0a039a816503a0d8077f/crates/matrix-sdk/src/event_cache/room/events.rs#L246-L254

Well, this is pretty bad, because it means the duplicated event will not be removed. And it means a new event with the same ID will be inserted. Kaboom 💥.

These patches fix that in simple steps:

1. `EventCacheStore::filter_duplicated_events` returns the event positions,
2. `Deduplicator` is redesigned to produce a `DeduplicatorOutcome` which makes the difference between in-memory vs. in-store duplicated events,
3. Use this information accordingly.

These patches also fixes inconsistency between the `BloomFilterDeduplicator` (used when there is no store) and `StoreDeduplicator` (used when there is a store), like: the new events can contain duplicates (`new_events = [$ev0, $ev1, $ev0]`) in theory (not in practise so far). This check was only done in the bloom filter deduplicator. Now it's done for all of them, directly in `Deduplicator`. Same for detecting events with no ID, it's done beforehand. See the patches one by one.


crates/matrix-sdk/src/event_cache/deduplicator.rs
```diff
@@ -60,16 +61,47 @@ impl Deduplicator {
...
+    ) -> Result<DeduplicationOutcome, EventCacheError> {
+        // Remove all events with no ID, or that is duplicated inside `events`, i.e.
+        // `events` contains duplicated events in itself, e.g. `[$e0, $e1, $e0]`, here
+        // `$e0` is duplicated in within `events`.
+        {
+            let mut event_ids = BTreeSet::new();
+
+            events.retain(|event| {
+                let Some(event_id) = event.event_id() else {
+                    // No event ID? Bye bye.
```
stefanceriu, line 78: Under what circumstances would an event passed into here not have an event it? 🤔

crates/matrix-sdk/src/event_cache/room/events.rs
```diff
@@ -241,27 +241,23 @@ impl RoomEvents {
...
-                    event_position,
-                    // If removing an event results in an empty chunk, the empty chunk is removed
-                    // because nothing is going to be inserted in it apparently.
-                    EmptyChunk::Remove,
-                )
-                .expect("Failed to remove an event we have just found");
+    /// Empty chunks are going to be removed.
+    ///
+    /// Events **must** be sorted by their position (descending, i.e. from
+    /// newest to oldest).
```
stefanceriu, line 247: This feels a bit footgun-y, can't we sort them here too, similar to what the deduplicator is doing in its `sort_events_by_position_descending`?

crates/matrix-sdk/src/event_cache/deduplicator.rs
```diff
@@ -260,22 +299,63 @@ enum Decoration<I> {
...
+    Duplicated((I, Position)),
 
     /// This event is invalid (i.e. not well formed).
     Invalid(I),
 }
 
+pub(super) struct DeduplicationOutcome {
+    /// All events passed to the deduplicator.
+    ///
+    /// All events in this collection has a valid event ID.
```
stefanceriu, line 311:
```suggestion
    /// All events in this collection have a valid event ID.
```

stefanceriu (approved):
Left some comments inline but otherwise makes sense and lgtm 👍

crates/matrix-sdk-base/src/event_cache/store/integration_tests.rs
```diff
@@ -787,9 +787,18 @@ impl EventCacheStoreIntegrationTests for DynEventCacheStore {
             .unwrap();
 
         assert_eq!(duplicated_events.len(), 3);
-        assert_eq!(duplicated_events[0], event_comte.event_id().unwrap());
-        assert_eq!(duplicated_events[1], event_morbier.event_id().unwrap());
-        assert_eq!(duplicated_events[2], event_mont_dor.event_id().unwrap());
+        assert_eq!(
+            duplicated_events[0],
+            (event_comte.event_id().unwrap(), Position::new(CId::new(0), 0))
```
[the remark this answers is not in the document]
Hywan: Yes…

crates/matrix-sdk/src/event_cache/room/events.rs
```diff
@@ -241,27 +241,23 @@ impl RoomEvents {
...
+    pub fn remove_events_by_position<P>(&mut self, positions: P) -> Result<(), Error>
+    where
+        P: Iterator<Item = Position>,
+    {
+        for position in positions {
+            self.chunks.remove_item_at(
+                position,
+                // If removing an event results in an empty chunk, the empty chunk is removed.
+                EmptyChunk::Remove,
```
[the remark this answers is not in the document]
Hywan: It makes a lot more sense, thanks!

bnjbvr (commented):
Nice work! just skimmed the high-level changes so I could understand it, and it makes sense to me :+1:

Issue #3886, opened 2024-08-23
Reconcile or unify `RoomEvents` event chunks with `AllEventsCache` events


jmartinesp:
In `EventCache::save_event` there is [this comment](https://github.com/matrix-org/matrix-rust-sdk/blob/be133886b7fef136264537633da039559a473afc/crates/matrix-sdk/src/event_cache/mod.rs#L573) that says:

> This doesn't insert the event into the linked chunk. In the future there'll be no distinction between the linked chunk and the separate cache.

And we have 2 separate caches now: one for chunks of events for timelines of rooms (`RoomEvents`), which come from room timeline updates during syncs and back pagination, and another one for any event in any room and their related events (`AllEventsCache`), which may come from any source. 

I believe the end goal here is using a persistent storage to hold a cache that unifies both of these, and it's probably related to the *'implement a basic reconciliation of events; don't handle separate connected components of events yet'* point in https://github.com/matrix-org/matrix-rust-sdk/issues/3058. I'm opening this issue to both discuss if this is the way we want to proceed and, if it is, keep some recording of it.

bnjbvr:
After we've implemented proper persistent storage for events, I'd expect that the second `AllEventsCache` would read from the database, instead of storing any information in memory, so we can likely fold this as one item of #3280?

bnjbvr:
For what it's worth, we might even not need migrations for the sqlite backend, if we consider using sqlite's [`json_extract`](https://www.sqlite.org/json1.html#jex) function to inspect the event's content. This might be kind of equivalent to the parsing done by Ruma, though, so we might as well add new columns to the events table (one for the relation type and another one for the related event id).

Hywan:
I know it's not something easy, but I think we may want to bench having a new column for the event type vs. using `json_extract`.  We've talk about this in the past already, but I didn't get time to do this yet on my side.

bnjbvr:
`json_extract` would work for most cases, because we could get the relation type and the related event using two calls:

- the relation type is *mostly* readable from ```$.content.`m.relates_to`.rel_type```
- the related event id is *mostly* readable from ```$.content.`m.relates_to`.event_id```

Unfortunately, this is not sufficient: the `AllEventsCache` also wants to let us know about redacted events. Redacted events may store the redacted event id in either `$.content.redacts` or `$.redacts`, according to the room version (redacted redactions seem to only store it in `$.content.redacts`).

[...]

PR #4708, opened 2025-02-24
fix(sdk): `RoomEventCache::event` looks inside the store

Hywan:
This patch fixes a bug in `RoomEventCache::event` since https://github.com/matrix-org/matrix-rust-sdk/pull/4632 has landed. `RoomEvents` now likely contains a partial view of the events, i.e. they are not all loaded from the store to memory.

A new `EventCacheStore::find_event` is added to look for an event in the store. `RoomEventCache::event` uses it to use the store, with a fallback to `AllEventsCache` because we can't get rid of it right now (see https://github.com/matrix-org/matrix-rust-sdk/issues/3886).

### DIFF
--- a/crates/matrix-sdk-common/src/linked_chunk/relational.rs
+++ b/crates/matrix-sdk-common/src/linked_chunk/relational.rs
@@ -293,9 +293,9 @@ impl<Item, Gap> RelationalLinkedChunk<Item, Gap> {
         }
     }
 
-    /// Return an iterator that yields events of a particular room with no
+    /// Return an iterator that yields items of a particular room with no
     /// particular order.
-    pub fn unordered_events<'a>(
+    pub fn unordered_items<'a>(
         &'a self,
         room_id: &'a RoomId,
     ) -> impl Iterator<Item = (&'a Item, Position)> {
@@ -305,6 +305,17 @@ impl<Item, Gap> RelationalLinkedChunk<Item, Gap> {
                     Either::Item(item) => Some((item, item_row.position)),
                     Either::Gap(..) => None,
                 }
+            } else {
+                None
+            }
+        })
+    }
+
+    /// Return an iterator over all items.
+    pub fn items(&self) -> impl Iterator<Item = (Position, &Item, &RoomId)> {
+        self.items.iter().filter_map(|item_row| {
+            if let Either::Item(item) = &item_row.item {
+                Some((item_row.position, item, item_row.room_id.as_ref()))
             } else {
                 None
             }
@@ -1131,7 +1142,7 @@ mod tests {
             ],
         );
 
-        let mut events = relational_linked_chunk.unordered_events(room_id);
+        let mut events = relational_linked_chunk.unordered_items(room_id);
 
         assert_eq!(events.next().unwrap(), (&'a', Position::new(CId::new(0), 0)));
         assert_eq!(events.next().unwrap(), (&'b', Position::new(CId::new(0), 1)));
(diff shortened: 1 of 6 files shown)


crates/matrix-sdk/src/event_cache/room/mod.rs
```diff
@@ -938,6 +942,31 @@ mod private {
             &self.events
         }
 
+        /// Find a single event in this room.
+        pub async fn event(
```
bnjbvr, line 946: Can we call this `find_event`? Also, can we add in the comment that it will try to load from memory first, and then attempt to load it from persisted storage, if it couldn't find it in memory?

crates/matrix-sdk/src/event_cache/room/mod.rs
```diff
@@ -938,6 +942,31 @@ mod private {
...
+            &self,
+            event_id: &EventId,
+        ) -> Result<Option<(Position, TimelineEvent)>, EventCacheError> {
+            let room_id = self.room.as_ref();
+
+            // There is supposedly less events loaded in memory than in the store. Let's
```
bnjbvr, line 952:
```suggestion
            // There are supposedly fewer events loaded in memory than in the store. Let's
```

crates/matrix-sdk/src/event_cache/room/mod.rs
```diff
@@ -938,6 +942,31 @@ mod private {
             &self.events
         }
 
+        /// Find a single event in this room.
+        pub async fn event(
+            &self,
+            event_id: &EventId,
+        ) -> Result<Option<(Position, TimelineEvent)>, EventCacheError> {
```
bnjbvr, line 949: What's the value of returning the position? No callers make uses of it. (If the answer is "but maybe later they do!" my counter-answer is "then let's add it later :nerd_face:")

bnjbvr (approved):
Nice, thanks!

crates/matrix-sdk/src/event_cache/room/mod.rs
Hywan: We need it for `maybe_apply_new_redaction` 😛.

bnjbvr: I hear the song of "we need it later" 😏

Hywan: hehe, but this PR is a prequel of the work for `maybe_apply_new_redaction`, I should have shared that.

crates/matrix-sdk-common/src/linked_chunk/relational.rs
```diff
@@ -310,6 +310,17 @@ impl<Item, Gap> RelationalLinkedChunk<Item, Gap> {
...
+    /// Return an iterator over all items.
+    pub fn items(&self) -> impl Iterator<Item = (Position, &Item, &RoomId)> {
+        self.items.iter().filter_map(|item_row| {
+            if let Either::Item(item) = &item_row.item {
+                Some((item_row.position, item, item_row.room_id.as_ref()))
+            } else {
+                None
+            }
+        })
+    }
```
bnjbvr, line 323: Seems this is `unordered_items` with an extra return value; can we fuse those two functions?

bnjbvr (commented):
Just one more question about fusing two methods in the in-memory event cache store implementation, and we should be good to go, thanks!

bnjbvr:
lol i forgot i approved yesterday

Hywan:
Oh damn, sorry!!

crates/matrix-sdk-common/src/linked_chunk/relational.rs
bnjbvr: oh i found the difference now, `unordered_items` is keyed by room. Maybe the names could be clearer…

bnjbvr:
opened https://github.com/matrix-org/matrix-rust-sdk/pull/4711 with my review suggestions!

PR #4709, opened 2025-02-24
feat(event cache): leaner updates when clearing/shrinking to the last chunk

bnjbvr:
I wanted to look at what it'd take to implement the `TODO`s about clearing, and I think I got them right, in the first commit; the only thing to pay attention to was the `last_index_per_reader` that needed an update when clearing updates.

But this wasn't sufficient to actually get rid of the spurious updates I've introduced in a previous test, because we're _extending the previous updates with the ones caused by the shrink/reset_. Instead, I've tried passing the updates diff vector as a `&mut` parameter to reset/shrink etc., but it makes the APIs a bit uglier, but eventually it behaves as I expected. What do you think @Hywan?

### DIFF
--- a/crates/matrix-sdk-common/src/linked_chunk/updates.rs
+++ b/crates/matrix-sdk-common/src/linked_chunk/updates.rs
@@ -139,8 +139,8 @@ impl<Item, Gap> ObservableUpdates<Item, Gap> {
     }
 
     /// Clear all pending updates.
-    pub(super) fn clear(&mut self) {
-        self.inner.write().unwrap().clear();
+    pub(super) fn clear_pending(&mut self) {
+        self.inner.write().unwrap().clear_pending();
     }
 
     /// Take new updates.
@@ -248,7 +248,7 @@ impl<Item, Gap> UpdatesInner<Item, Gap> {
     }
 
     /// Clear all pending updates.
-    fn clear(&mut self) {
+    fn clear_pending(&mut self) {
         self.updates.clear();
 
         // Reset all the per-reader indices.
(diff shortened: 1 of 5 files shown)


crates/matrix-sdk/src/event_cache/room/mod.rs
```diff
@@ -925,12 +930,21 @@ mod private {
 
         /// Resets this data structure as if it were brand new.
         #[must_use = "Updates as `VectorDiff` must probably be propagated via `RoomEventCacheUpdate`"]
-        pub async fn reset(&mut self) -> Result<Vec<VectorDiff<TimelineEvent>>, EventCacheError> {
+        pub async fn reset(
+            &mut self,
+            sync_timeline_events_diffs: &mut Vec<VectorDiff<TimelineEvent>>,
```
Hywan, line 935: Why do you pass a `&mut` instead of keeping to return the diffs?

Hywan (commented):
I'm convinced by the first patch, thanks! I've a question about the second patch.

crates/matrix-sdk/src/event_cache/room/mod.rs
bnjbvr: This was to follow the pattern introduced in `replace_with`, which makes use of this function. I can isolate using the pattern to `replace_with` / `shrink_to_last_chunk`, if you'd prefer, and have the override happen there?

Hywan: I'm struggling to see how it will look like but if the idea is to remove this `&mut`, I think I would prefer that yes!

crates/matrix-sdk-common/src/linked_chunk/updates.rs
```diff
@@ -242,6 +247,19 @@ impl<Item, Gap> UpdatesInner<Item, Gap> {
         }
     }
 
+    /// Clear all pending updates.
+    fn clear(&mut self) {
```
stefanceriu, line 251: Would it maybe be better if we call this `clear_pending`?

bnjbvr: Why not!

PR #4712, opened 2025-02-25
test(oidc): Use MatrixMockServer and MockClientBuilder where possible

zecakeh:
As discussed in https://github.com/matrix-org/matrix-rust-sdk/pull/4687#discussion_r1967556139, we want to get rid of the `OidcBackend` trait to mock a real authorization server instead.

This implements an `OauthMockServer` to handle the OAuth 2.0-related endpoints and uses it as much as possible.

We keep the mock backend for now because mocking the server would require generating ID tokens on the fly (which involves encryption with private keys). Since ID tokens are only part of OpenID Connect, and not OAuth 2.0, we will remove support for them, so it doesn't seem worth it to add temporary code for this. We can remove support for them easily and get rid of the backend trait when this and #4699 are merged.

### DIFF
--- a/crates/matrix-sdk/src/authentication/oidc/backend/mock.rs
+++ b/crates/matrix-sdk/src/authentication/oidc/backend/mock.rs
@@ -27,7 +27,7 @@ use mas_oidc_client::{
         client_credentials::ClientCredentials,
         errors::ClientErrorCode,
         iana::oauth::OAuthTokenTypeHint,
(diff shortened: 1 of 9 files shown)


poljar (approved):
Some of the new mocks could take some more constructors, i.e. the `/keys/upload` one is hard-coded to return 20 one-time keys. This would have come in handy in the last commit where the previous mock returned 50 one-time keys in the response.

But ok, we can add this later and this particular test doesn't really care about the exact count.

poljar:
Oh also, thanks a lot for tackling this. This makes the tests much more pleasant.

PR #4715, opened 2025-02-25
Retry decryption from a long-lived async task, instead of lots of detached ones

andybalaam:
Part of https://github.com/element-hq/element-meta/issues/2697

In response to a [comment on #4644](https://github.com/matrix-org/matrix-rust-sdk/pull/4644/files#r1963599148) - avoid spawning a detached async task every time we have some things to decrypt because some Megolm sessions updated.

Instead, we create one async task and hold it inside a new struct called `RetryDecryptionTask`. This struct holds one end of channel, and sends retry requests to the async task, which stops when the channel is closed, which will happen when the `RetryDecryptionTask` struct is dropped.

So we keep the feature that retrying decryption does not block the main processing, but we lose the spawning of lots of async tasks that are not kept track of.

As written, the task is not cancellable, but this could be added if we think we need it.

I strongly suggest reviewing commit by commit, since this is structured as a set of refactorings that hopefully make sense individually.

### DIFF
--- a/crates/matrix-sdk-ui/src/timeline/controller/decryption_retry_task.rs
+++ b/crates/matrix-sdk-ui/src/timeline/controller/decryption_retry_task.rs
@@ -107,8 +107,17 @@ async fn decryption_task<D: Decryptor>(
     mut receiver: Receiver<DecryptionRetryRequest<D>>,
 ) {
     debug!("Decryption task starting.");
+
     while let Some(request) = receiver.recv().await {
-        let retry_indices = retry_indices(state.clone(), &request.session_ids).await;
+        let should_retry = |session_id: &str| {
+            if let Some(session_ids) = &request.session_ids {
+                session_ids.contains(session_id)
+            } else {
+                true
+            }
+        };
+
+        let retry_indices = item_indices_to_retry(state.clone(), &should_retry).await;
         if !retry_indices.is_empty() {
             debug!("Retrying decryption");
             decrypt_by_index(
@@ -116,28 +125,24 @@ async fn decryption_task<D: Decryptor>(
                 settings.clone(),
                 room_data_provider.clone(),
                 request.decryptor,
-                &request.session_ids,
+                should_retry,
                 retry_indices,
             )
             .await
         }
     }
+
     debug!("Decryption task stopping.");
 }
 
-async fn retry_indices(
+/// Return a list of the items within the timeline that we should retry
+/// decrypting because their session updated. Items are identified by their
+/// index in the supplied `state`'s list of items.
+async fn item_indices_to_retry(
     state: Arc<RwLock<TimelineState>>,
-    session_ids: &Option<BTreeSet<String>>,
+    should_retry: impl Fn(&str) -> bool,
 ) -> Vec<usize> {
     let state = state.read_owned().await;
-
-    let should_retry = |session_id: &str| {
-        if let Some(session_ids) = &session_ids {
-            session_ids.contains(session_id)
-        } else {
-            true
-        }
-    };
 
     state
         .items
@@ -161,17 +166,9 @@ async fn decrypt_by_index<D: Decryptor>(
     settings: TimelineSettings,
     room_data_provider: impl RoomDataProvider,
     decryptor: D,
-    session_ids: &Option<BTreeSet<String>>,
+    should_retry: impl Fn(&str) -> bool,
     retry_indices: Vec<usize>,
 ) {
-    let should_retry = move |session_id: &str| {
-        if let Some(session_ids) = &session_ids {
-            session_ids.contains(session_id)
-        } else {
-            true
-        }
-    };
-
     let mut state = state.clone().write_owned().await;
 
     let push_rules_context = room_data_provider.push_rules_and_context().await;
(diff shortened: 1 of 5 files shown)


andybalaam:
This lines that codecov is complaining about missing coverage are pretty much all lifted-and-shifted from the their old location into `decryption_retry_task.rs`. This code is tested by existing tests in `crates/matrix-sdk-ui/src/timeline/tests/encryption.rs` and `crates/matrix-sdk-ui/src/timeline/tests/read_receipts.rs`. Whilst it would be nice to add unit tests closer to the code, these cover the more real-world cases, and this PR is a detour-on-a-detour for me :-)

andybalaam:
@poljar is this something like what you intended in https://github.com/matrix-org/matrix-rust-sdk/pull/4644/files#r1963599148 ?

crates/matrix-sdk-ui/src/timeline/controller/decryption_retry_task.rs
```diff
@@ -0,0 +1,241 @@
...
+    pub(crate) fn new<P: RoomDataProvider>(
+        state: Arc<RwLock<TimelineState>>,
+        room_data_provider: P,
+    ) -> Self {
+        // We will send decryption requests down this channel to the long-running task
+        let (sender, receiver) = mpsc::channel(CHANNEL_BUFFER_SIZE);
+
+        // Spawn the long-running task, providing the receiver so we can listen for
+        // decryption requests
+        matrix_sdk::executor::spawn(decryption_task(state, room_data_provider, receiver));
```
poljar, line 63: Hmm, we're still dropping the `JoinHandle` though the task will get killed because the `sender` gets dropped, which in turn means the receiver will get a `None`.

That's a bit implicit compared to holding on to the `JoinHandle` and cancelling in the `drop()` call of `DecryptionRetryTask, but I guess there's nothing wrong with it.

poljar:
> @poljar is this something like what you intended in https://github.com/matrix-org/matrix-rust-sdk/pull/4644/files#r1963599148 ?

Yeah, I think that's fine, I explained things a bit more here: https://github.com/matrix-org/matrix-rust-sdk/pull/4715#pullrequestreview-2644286256

crates/matrix-sdk-ui/src/timeline/tests/read_receipts.rs
```diff
@@ -485,18 +486,17 @@ async fn test_read_receipts_updates_on_message_decryption() {
         )
         .await;
 
-    assert_eq!(timeline.controller.items().await.len(), 2);
```
poljar, line 488: Why was this removed?

poljar (approved):
Nice work, thanks for splitting this into many small commits, made the review easier. I left one nit, but I'll approve this in advance.

There's also a typo in one commit:

> Adjust tsts that retry decryption to wait with

In any case, this should make it much easier to move the redecryption logic into the correct place, wherever we decide this to be.

crates/matrix-sdk-ui/src/timeline/controller/decryption_retry_task.rs
andybalaam: Yeah, I tried to join in `drop()` but it's async, so I can't do it.

crates/matrix-sdk-ui/src/timeline/tests/read_receipts.rs
andybalaam: Because it is not immediately true: we need to wait for the async events to happen. Given that in the rest of this test we check that the timeline stream is as we expect, I didn't think there was much benefit in this check. If we want to keep it, we can move it later so we know the async work has been processed.

crates/matrix-sdk-ui/src/timeline/controller/decryption_retry_task.rs
poljar: Oh, you wouldn't await the join in the `drop()` call, you would just `abort()` the task. But perhaps it's cleaner to let the task naturally die due to the `Sender` being gone.

crates/matrix-sdk-ui/src/timeline/tests/read_receipts.rs
poljar: Ah ok, fine by me to remove it then.

crates/matrix-sdk-ui/src/timeline/controller/decryption_retry_task.rs
bnjbvr: I think keeping the join handle and manually aborting shows a clear ownership of the task, and would be a tiny bit clearer, even in terms of discoverability: we'd see the `JoinHandle` in the struct, thus know that this is a task's owner.

bnjbvr: Fair; my argument in terms of discoverability still holds IMO, can we store the task in the struct at least?

poljar: Sure, that's probably a wise idea.

poljar (changes requested):
Sorry to change my mind, could you please put the `JoinHandle` into the `DecryptionRetryTask`.

I shouldn't be used for anything, just to let people know that the task is owned by this struct.

crates/matrix-sdk-ui/src/timeline/controller/decryption_retry_task.rs
andybalaam: Done in f39a969da3

poljar (approved):
Thanks for the last minute change, we can merge this after you make the compiler happy with the right types.

PR #4716, opened 2025-02-25
ffi: make room and timeline creation orthogonal

bnjbvr:
This gets rid of the `Timeline` stored in an FFI `Room`. There's no good reason why we should create and hold on to a live timeline instance when we have a `Room` instance, since we might want a `Room` to only perform `Room`-related operations.

In general, holding onto live timelines in scattered places may lead to having the [auto-linked-chunk-shrink optimization](https://github.com/matrix-org/matrix-rust-sdk/pull/4703) **not** trigger, so we need to be careful and minimize the number of live timelines at all costs.

As a result, this gets rid of `Room::timeline`, which would return the cached timeline instance. `Room::timeline_with_configuration()` can be used to create a new timeline instance for a given room, or `RoomListItem::init_timeline()` and then `RoomListItem::timeline()`.

Also renamed the FFI `Room::inner` (which is a… SDK Room) to `Room::sdk`. Is this too cute? :pleading_face: 

cc @jmartinesp @stefanceriu, if you could test it out please :-)

### DIFF
--- a/bindings/matrix-sdk-ffi/src/room.rs
+++ b/bindings/matrix-sdk-ffi/src/room.rs
@@ -10,7 +10,7 @@ use matrix_sdk::{
     ComposerDraft as SdkComposerDraft, ComposerDraftType as SdkComposerDraftType,
     RoomHero as SdkRoomHero, RoomMemberships, RoomState,
 };
-use matrix_sdk_ui::timeline::{default_event_filter, RoomExt};
+use matrix_sdk_ui::timeline::{default_event_filter, Timeline as SdkTimeline};
 use mime::Mime;
 use ruma::{
     api::client::room::report_content,
@@ -27,7 +27,6 @@ use ruma::{
     },
     EventId, Int, OwnedDeviceId, OwnedUserId, RoomAliasId, UserId,
 };
-use tokio::sync::RwLock;
 use tracing::{error, warn};
 
 use super::RUNTIME;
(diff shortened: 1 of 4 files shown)


bindings/matrix-sdk-ffi/src/room.rs
```diff
@@ -611,14 +592,21 @@ impl Room {
...
     ///
     /// Note: this does NOT unset the unread flag; it's the caller's
     /// responsibility to do so, if needs be.
     pub async fn mark_as_read(&self, receipt_type: ReceiptType) -> Result<(), ClientError> {
-        let timeline = self.timeline().await?;
+        // Create a throwaway timeline for this.
+        let timeline = SdkTimeline::builder(&self.inner)
+            .track_read_marker_and_receipts()
+            .build()
+            .await
```
Hywan: Is there a risk that the creation of the `Timeline` and the installation of all its tasks, take a lot of time? Maybe we should add a big warning in the doc?

Hywan (approved):
Thanks. FEWER! TIMELINE! FEWER! TIMELINE!

bindings/matrix-sdk-ffi/src/room.rs
bnjbvr: We know it can be slow, if there are many events, especially because of read receipts, so yeah I'll add a comment :+1:

bnjbvr:
Dropping this PR: `timeline_with_configuration()` is actually not a good replacement, as it:

- doesn't fill the event cache with the old naive event cache from sliding sync, when the new shiny event cache persistent storage isn't enabled
- doesn't make use of the UTD hook

See dependencies in https://github.com/matrix-org/matrix-rust-sdk/issues/4718.

PR #4717, opened 2025-02-25
FFI: Correctly indicate OIDC support when fetching metadata fails.

pixlwave:
There was a small regression introduced into the FFI layer in #4673 where an error that returned `false` was updated to return `true`. This PR fixes it.


stefanceriu (approved):
Nice catch 👍

PR #4719, opened 2025-02-25
feat(event cache): don't auto-shrink if the sync returned 0 events

bnjbvr:
Tests to be written.


bnjbvr:
I am now convinced this is incorrect: if the sync included a gap but no events, we may be rendering events from before the gap in the timeline, but not the gap, leading to the impression that some events (those that would be returned by resolving the gap) have been lost by the app. Closing for now.

PR #4724, opened 2025-02-26
fix(event cache): wait for the initial previous-batch token, when storage's enabled

bnjbvr:
If during back-pagination, we lazy-load a chunk from storage and realize there's no previous chunk, we might think we've reached the start of the timeline and that we're done. This is not true, if this was the first default chunk, and we never waited long enough to get the initial gap from sync. This patch fixes that, and includes a regression test showing the error.

Part of #3280.

### DIFF
--- a/crates/matrix-sdk/src/event_cache/room/mod.rs
+++ b/crates/matrix-sdk/src/event_cache/room/mod.rs
@@ -845,9 +845,16 @@ mod private {
                     }
 
                     Ok(None) => {
-                        // No previous chunk: no events to insert. Better, it means we've reached
-                        // the start of the timeline!
-                        return Ok(LoadMoreEventsBackwardsOutcome::StartOfTimeline);
+                        // No previous chunk: no events to insert. This means one of two things:
+                        // - either the linked chunk is at the start of the timeline,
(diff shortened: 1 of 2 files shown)


crates/matrix-sdk/src/event_cache/room/mod.rs
```diff
@@ -1182,8 +1188,19 @@ mod private {
             func: F,
         ) -> Result<(O, Vec<VectorDiff<TimelineEvent>>), EventCacheError> {
             let output = func(&mut self.events);
+
             self.propagate_changes().await?;
+
+            // If we've never waited for an initial previous-batch token, and we now have at
```
Hywan, line 1194: Why is it here? I feel like this is not the correct place, but it's a gut feeling.

Hywan (approved):
Thanks for the test. I don't understand one thing but it's documented and tested, I approve this PR.

crates/matrix-sdk/src/event_cache/room/mod.rs
Hywan: Thanks for the explanation. Makes sense to me.

PR #4725, opened 2025-02-26
chore(event cache): remove unused errors

bnjbvr:
Wish we had a way to detect those.


poljar:
> Wish we had a way to detect those.

Make the type private so the variants are considered dead code. But that's annoying unless the type is used in only a couple of places.

poljar (approved):

PR #4726, opened 2025-02-26
refactor(oidc): Remove support for ID tokens and get rid of OidcBackend

zecakeh:
ID tokens are a feature of OpenID Connect, we don't need them to support OAuth 2.0, so we remove the corresponding fields.

Since ID tokens were the complicated part to mock, we can now use `OauthMockServer` everywhere we couldn't before, and we can get rid of `OidcBackend`.

Tested against beta.matrix.org with the oidc-cli example.

Next step: actually porting to the oauth2 crate.


poljar:
Why was the `ClientMetadata` stuff removed as well?

zecakeh:
Because it is not used outside of registration so we don't need to keep it around.

We used it for requests in two places:
- To decide if Pushed Authorization Requests were mandatory. Support was removed in #4699.
- To know the signing algorithm that must be used for ID tokens. Support is removed in this PR.

poljar (approved):
Sorry for the delay. This looks good to me and I tested it with the OIDC example as well.

PR #4731, opened 2025-02-27
refactor(event cache): call /messages directly in the room pagination

bnjbvr:
And don'y rely on the `Paginator`. This simplifies the code a bit, avoids a few methods on the `Paginator`, and makes it more straightforward the pagination happens.

### DIFF
--- a/crates/matrix-sdk/src/event_cache/pagination.rs
+++ b/crates/matrix-sdk/src/event_cache/pagination.rs
@@ -17,21 +17,23 @@
 use std::{sync::Arc, time::Duration};
 
 use eyeball::{SharedObservable, Subscriber};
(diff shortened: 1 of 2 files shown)


crates/matrix-sdk/src/event_cache/pagination.rs
```diff
@@ -233,20 +235,28 @@ impl RoomPagination {
             }
         };
 
-        let (pagination_result, new_gap) = {
+        let (events, new_gap) = {
             // Use a throw-away paginator instance.
```
Hywan, line 239: This comment is now outdated.

Hywan (approved):

crates/matrix-sdk/src/event_cache/pagination.rs
bnjbvr: Good catch, thanks!

PR #4735, opened 2025-02-27
refactor(ffi): unify a bit more `Room::timeline_with_configuration` and `RoomListItem::init_timeline`

bnjbvr:
The two last missing pieces will be the UTD hook and loading events from the persistent storage.

Note timelines created with `Room::timeline_with_configuration` don't get a UTD hook yet, nor do they force-load the naive event cache we had a long time ago, when the shiny persistent event cache is disabled. As such, they shouldn't be used to create live timelines *yet*.

cc @stefanceriu @jmartinesp; feel free to steal the review :)


bindings/matrix-sdk-ffi/src/timeline/configuration.rs
```diff
@@ -82,4 +139,26 @@ pub struct TimelineConfiguration {
 
     /// How often to insert date dividers
     pub date_divider_mode: DateDividerMode,
+
+    /// Should the read receipts and read markers be tracked for the timeline
+    /// items in this instance?
+    ///
+    /// As this has a non negligible performance impact, make sure to enable it
+    /// only when you need it.
+    pub track_read_receipts: bool,
```
stefanceriu, line 148: Just to confirm, looking at the previous code it looks like this should be always false for now and it should only come in play when we get rid of the `init_timeline` method, right?

bindings/matrix-sdk-ffi/src/timeline/configuration.rs
```diff
@@ -82,4 +139,26 @@ pub struct TimelineConfiguration {
...
+}
+
+/// Create a configuration for a live timeline, that will react to sync events
+/// in live.
+///
+/// Note: as of 2025-02-27, doesn't include the UTD hook.
+#[matrix_sdk_ffi_macros::export]
+pub fn live_timeline_configuration(internal_id_prefix: Option<String>) -> TimelineConfiguration {
```
stefanceriu, line 156: I'm not sure this brings much value tbh, I thought we were moving away fom multiple ways of creating timelines which this sort of diverges from.

bindings/matrix-sdk-ffi/src/timeline/configuration.rs
```diff
@@ -52,28 +110,27 @@ impl From<DateDividerMode> for matrix_sdk_ui::timeline::DateDividerMode {
 }
 
 #[derive(uniffi::Enum)]
-pub enum AllowedMessageTypes {
+pub enum TimelineFilter {
+    /// Show all the events in the timeline, independent of their type.
     All,
-    Only { types: Vec<RoomMessageEventMessageType> },
+    /// Show only `m.room.messages` of the given room message types.
+    OnlyMessages {
```
stefanceriu, line 117: Nit: I would call this one `OnlyMessage` as `.onlyMessage(types:)` rolls better of the tongue.

stefanceriu (approved):
Looks and more importantly works great on my side 👍

bindings/matrix-sdk-ffi/src/room.rs
```diff
@@ -206,30 +206,50 @@ impl Room {
...
-                default_event_filter(event, room_version_id)
-                    && match event {
-                        AnySyncTimelineEvent::MessageLike(msg) => match msg.original_content() {
-                            Some(AnyMessageLikeEventContent::RoomMessage(content)) => {
-                                types.contains(&content.msgtype.into())
+        if configuration.track_read_receipts {
+            builder = builder.track_read_marker_and_receipts();
+        }
+
+        match configuration.filter {
```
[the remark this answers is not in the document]
bnjbvr: Yep, both filters are tested in the SDK. Not the particular instances of those filters, but heh this is the FFI crate, and the logic is relatively plain here…

bindings/matrix-sdk-ffi/src/timeline/configuration.rs
bnjbvr: Exactly!

bindings/matrix-sdk-ffi/src/timeline/configuration.rs
bnjbvr: It's a way to create a timeline *configuration*, not a timeline. But maybe too early indeed; I'll keep that for when we get rid of the `init_timeline` method.

PR #4738, opened 2025-02-28
chore: Tweak the weekly-report command to include PR numbers

poljar:


stefanceriu (approved):
lgtm but are you sure you don't want to keep the author?

poljar:
> lgtm but are you sure you don't want to keep the author?

In the past 3-4 TWIM posts we didn't include authors, if there is one we do want to call out it's one click away to get it from the PR link.

PR #4740, opened 2025-03-03
fix(sdk): `maybe_apply_new_redaction` updates in-store events

Hywan:
This is the last known task required after the event cache lazy-loading feature.

This is unblocked by https://github.com/matrix-org/matrix-rust-sdk/pull/4708.

`maybe_apply_new_redaction` was looking for events only from `RoomEvents`, so only in-memory. And the redaction was only happening in-memory too. With this patch, `maybe_apply_new_redaction` handles in memory and in store.

This pull request should be reviewed patch by patch for the sake of clarity.

A clean up in the `Deduplicator` is possible by using the new `EventLocation` enum, but it's not part of this set of patches.

---

### DIFF
--- a/crates/matrix-sdk/src/event_cache/room/events.rs
+++ b/crates/matrix-sdk/src/event_cache/room/events.rs
@@ -102,7 +102,11 @@ impl RoomEvents {
     /// If the given event is a redaction, try to retrieve the to-be-redacted
     /// event in the chunk, and replace it by the redacted form.
     #[instrument(skip_all)]
-    fn maybe_apply_new_redaction(&mut self, room_version: &RoomVersionId, event: &Event) {
+    pub(super) fn maybe_apply_new_redaction(
+        &mut self,
+        room_version: &RoomVersionId,
+        event: &Event,
+    ) {
         let raw_event = event.raw();
 
         // Do not deserialise the entire event if we aren't certain it's a
@@ -176,17 +180,6 @@ impl RoomEvents {
         }
 
         // TODO: remove all related events too!
-    }
-
-    /// Callback to call whenever we touch events in the database.
-    pub fn on_new_events<'a>(
-        &mut self,
-        room_version: &RoomVersionId,
-        events: impl Iterator<Item = &'a Event>,
-    ) {
-        for ev in events {
-            self.maybe_apply_new_redaction(room_version, ev);
-        }
     }
 
     /// Push events after all events or gaps.
--- a/crates/matrix-sdk/src/event_cache/room/events.rs
+++ b/crates/matrix-sdk/src/event_cache/room/events.rs
@@ -16,7 +16,6 @@ use as_variant::as_variant;
 use eyeball_im::VectorDiff;
 pub use matrix_sdk_base::event_cache::{Event, Gap};
 use matrix_sdk_base::{
-    apply_redaction,
     event_cache::store::DEFAULT_CHUNK_CAPACITY,
     linked_chunk::{
         lazy_loader::{self, LazyLoaderError},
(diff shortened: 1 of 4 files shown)


crates/matrix-sdk/src/event_cache/room/mod.rs
```diff
@@ -1347,6 +1351,11 @@ pub(super) enum EventsPostProcessing {
     None,
 }
 
+pub(super) enum EventLocation {
+    InMemory,
+    InStore,
```
bnjbvr: Can these be called `Memory` and `Store`, only? Would be more concise, and the enum's name already indicates it's a location. (Otherwise it's temping to read it as "where is the event location itself?".)

crates/matrix-sdk/src/event_cache/room/mod.rs
```diff
@@ -1337,7 +1337,11 @@ mod private {
                                 .expect("should have been a valid position of an item");
                         }
                         EventLocation::InStore => {
-                            todo!()
+                            self.send_updates_to_store(vec![Update::ReplaceItem {
```
bnjbvr, line 1340: Out of curiosity, why to include the fix to your own previous commit as a separate commit?

bnjbvr (changes requested):
Thanks, looks good! can you add an integration test, where the event is only in the store, and not in memory, please?

crates/matrix-sdk/src/event_cache/room/mod.rs
Hywan: I wanted to go step-by-step: first the refactoring, then the fix (that's only possible because of the refactoring). Glad to merge these 2 patches in a single one when review will be finished.

crates/matrix-sdk/src/event_cache/room/mod.rs
```diff
@@ -1230,8 +1237,10 @@ mod private {
 
             match post_processing {
                 EventsPostProcessing::HaveBeenInserted(events) => {
+                    let room_version = self.room_version.clone();
```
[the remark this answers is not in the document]
Hywan: Good eyes!

crates/matrix-sdk/src/event_cache/room/mod.rs
Hywan: Sure. Done.

crates/matrix-sdk/src/event_cache/room/events.rs
```diff
@@ -262,6 +173,14 @@ impl RoomEvents {
         Ok(())
     }
 
+    /// Replace event at a specified position.
+    ///
+    /// `position` must point to a valid item, otherwise the method returns
+    /// `Err`.
```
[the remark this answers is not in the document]
Hywan: Yup.

crates/matrix-sdk/src/event_cache/room/mod.rs
```diff
@@ -1095,7 +1093,9 @@ mod private {
                                 .map(|(_event_id, position)| position)
                                 .collect(),
                         )
-                        .expect("failed to remove an event")
+                        .expect("failed to remove an event");
+
+                    EventsPostProcessing::None
```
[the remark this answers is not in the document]
Hywan: Okay. I've added a new commit (happy to squash it) that removes `EventsPostProcessing` and updates the documentation of `with_events_mut`.

bnjbvr (approved):
Nice, thank you!

Hywan:
I forgot to add the test… here it is, https://github.com/matrix-org/matrix-rust-sdk/pull/4760/!

PR #4747, opened 2025-03-04
chore(sdk): Reduce log verbosity from info -> trace in a few functions

kevinaboos:
At the default `INFO` log level, the log gets inundated with thousands of emitted statements, primarily related to `is_display_name_ambiguous()`. The execution of this tiny function certainly doesn't need to be traced every time, at least not at the info log level.

Note: some of these could be debatably reduced to debug level rather than trace level, but I went with "trace" because they all seem to be trace statements rather than actual debug dump outputs (there is no actual program state dumped out).

Signed-off-by: Kevin Boos [kevinaboos@gmail.com](mailto:kevinaboos@gmail.com)


crates/matrix-sdk-base/src/store/ambiguity_map.rs
```diff
@@ -86,7 +86,7 @@ pub(crate) struct AmbiguityCache {
     pub changes: BTreeMap<OwnedRoomId, BTreeMap<OwnedEventId, AmbiguityChange>>,
 }
 
-#[instrument(ret)]
+#[instrument(ret, level = "trace")]
```
bnjbvr, line 89: This will only create a span (i.e. instrument) if the level is trace, does it actually change anything, since the only log i can see below is already at the trace log level?

crates/matrix-sdk-ui/src/timeline/builder.rs
```diff
@@ -212,7 +212,7 @@ impl TimelineBuilder {
             let room_event_cache = room_event_cache.clone();
             let inner = controller.clone();
 
-            let span = info_span!(
+            let span = trace_span!(
```
bnjbvr, line 215: I think we should keep these spans at the regular log level, as they're important to realize where a log message is coming from.

crates/matrix-sdk-ui/src/timeline/builder.rs
```diff
@@ -315,7 +315,7 @@ impl TimelineBuilder {
                     timeline.handle_local_echo(echo).await;
                 }
 
-                let span = info_span!(
+                let span = trace_span!(
```
bnjbvr, line 318: ditto here

bnjbvr (changes requested):
Thanks! Makes sense, these logs can be verbose. I'm less certain about the span changes, though: I think those span make sense to keep, to isolate where things are happening; could we keep them, please?

crates/matrix-sdk-base/src/store/ambiguity_map.rs
kevinaboos: oh, perhaps i misunderstood what you were asking. if you were asking about how this change affects the `trace!("Checking if a display name is ambiguous");` line within this function, then the answer is that they're separate. 

This attribute adds an implicit trace/log output every time that this function is entered, which is what caused the ~4000 INFO-level log outputs in Robrix. The latter `trace!()` statement in line 94 is an independent, explicit trace statement that is independent from the instrumentation attribute.

bnjbvr: so, I learned about the log level parameter for a span only as of today, so please bear with me :)

My understanding was that if a span is at log level ABC, then this log level ABC must be enabled at runtime for the span to be instantiated; otherwise, it is not.

So when if I have:

```
#[instrument(level = "trace")
fn my_func(val: u32) {
  tracing::trace!("le trace");
  tracing::warn!("le warning");
}
```

and I'm running this with `RUST_LOG=warn`, then I'll see, with the default tracing subscriber formatter:

```
le warning
```

but if I run with `RUST_LOG=trace`, then I'll get:

```
my_func(val = 42): le trace
my_func(val = 42): le warning
```

Is that understanding correct? If so, maybe I misunderstood your intent, and you only wanted to reduce the size of an individual line, by having fewer recorded fields in a span, etc.?

(In that case: i'll need to check on a case-by-case basis, but I think the context brought by some of these spans is *always* useful, so we might want to enable them *all the time*.)

bnjbvr: @kevinaboos any news about this^?

kevinaboos: Sorry for the delay! totally missed this.

> Is that understanding correct? If so, maybe I misunderstood your intent, and you only wanted to reduce the size of an individual line, by having fewer recorded fields in a span, etc.?

That wasn't my intent. I was hoping to reduce the log level of the `#[instrument]`-emitted log statement itself. See below.

As far as I understand (and have experienced locally),  the `#[instrument]` attribute does not just affect the span of the function. When I test it within Robrix, the existence of the `#[instrument]` attribute _itself_ results in a bunch of log statements about the `is_display_name_ambiguous()` function being _entered_. It's not just that it affects the span level of the subsequent line
```rust
    trace!("Checking if a display name is ambiguous");
```
... although it does also do that, probably.

Question: with the trace logging level enabled, do you not observe thousands of logs about the `is_display_name_ambiguous()` function being entered? That's the main thing that's absolutely flooding my logs, not the `trace!()` statement within that function.

kevinaboos: Anyway, what I had hoped to achieve here was to demote the `#[instrument]` log statement that logs the function entry from the `info` level (which is default) to the `trace` level.

If you're saying that the `#[instrument]` attribute _also_ affects the span generated for the following `trace!()` log line, then wouldn't that still be okay since it's just a `trace!()` statement anyway? If not, then maybe we can keep the attribute as `#[instrument(ret, level = "trace")]` but then manually create an info-level span in the first line of `is_display_name_ambiguous()`. Would that work?

kevinaboos: ahhhhh ok ok, I got it now. Ignore what i wrote above (which is only partially correct).

Your reproduction did not include the `ret` parameter, it only included the `#[instrument]` bare attribute. That explains why you did not observe the log statement that dumps out the function's return value upon exit.

And then, my change resulted in the level being set to TRACE for the entire _span_, not _just_ for the `ret` parameter.

Turns out we can get what we both want, by doing:
```rust
#[instrument(ret(level = Level::TRACE)]
```

kevinaboos: Fixed in my latest commit. Please resolve this if you agree this fix works.

kevinaboos:
Should be ready to go now. Sorry for the delay!

crates/matrix-sdk-ui/src/timeline/builder.rs
kevinaboos: restored.

crates/matrix-sdk-ui/src/timeline/builder.rs
kevinaboos: restored.

bnjbvr (approved):
Cool, thanks!

bnjbvr:
Approved, but github tells me there are conflicts that need to be resolved; can you rebase please?

kevinaboos:
> Approved, but github tells me there are conflicts that need to be resolved; can you rebase please?

done via a merge commit. I presume you'll just do a squash merge.

kevinaboos:
looks like there's an unrelated CI failure:
```
thread 'tests::timeline::test_new_users_first_messages_dont_warn_about_insecure_device_if_it_is_secure' panicked at testing/matrix-sdk-integration-testing/src/tests/timeline.rs:876:19:
Next event timed out: Elapsed(())
```

might just be a transient time out, dunno. I don't have the permissions to re-run the failing actions though.

bnjbvr:
Indeed, thanks for the ping; I've opened an [intermittent failure issue](https://github.com/matrix-org/matrix-rust-sdk/issues/4871), and kicked CI to restart it :crossed_fingers:

PR #4749, opened 2025-03-04
fix(timeline): don't duplicate read receipts of hidden events after a pagination

bnjbvr:
WIP


stefanceriu:
> One remark: [8001d71](https://github.com/matrix-org/matrix-rust-sdk/commit/8001d71720e712f36cc4230ca3fd60e4a8739ab3) we should probably delete, otherwise the logs and rageshakes will grow a lot. This has been quite useful for debugging though, so maybe should be behind a feature toggle to enable those issues.

`matrix_sdk_ui::timeline` currently [defaults to info](https://github.com/matrix-org/matrix-rust-sdk/blob/9c8762591034ee53bcbf4702666a7ef140bfc316/bindings/matrix-sdk-ffi/src/platform.rs#L290) on the FFI layer and I see all the new logs are Trace so let's perhaps just keep them?

stefanceriu (approved):
What a great PR! The explanations, documentation and test are super super useful, thank you for taking the time!

Hywan (approved):
The changes look sain to me too. I'm merging it.

PR #4750, opened 2025-03-05
test(sqlite): Improve a test to check the uniqueness constraint

Hywan:
Title says it all.

### DIFF
--- a/crates/matrix-sdk-sqlite/src/event_cache_store.rs
+++ b/crates/matrix-sdk-sqlite/src/event_cache_store.rs
@@ -1867,6 +1867,9 @@ mod tests {
         let store = get_event_cache_store().await.expect("creating cache store failed");
 
         let room_id = *DEFAULT_TEST_ROOM_ID;
+        let event_0 = make_test_event(room_id, "hello");
+        let event_1 = make_test_event(room_id, "world");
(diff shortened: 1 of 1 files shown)


crates/matrix-sdk-sqlite/src/event_cache_store.rs
```diff
@@ -1899,6 +1898,25 @@ mod tests {
 
         let chunks = store.load_all_chunks(room_id).await.unwrap();
         assert!(chunks.is_empty());
+
+        // It's okay to re-insert a past event.
```
bnjbvr, line 1902: Do we want to add a `TODO: it's not okay actually!!` or something like this?

bnjbvr (approved):

crates/matrix-sdk-sqlite/src/event_cache_store.rs
Hywan: It's actually okay to do that: if we remove an event because it is duplicated, then it is re-inserted, it's perfectly normal, right?

bnjbvr: Ah, I didn't notice it was removed in the middle; so yes that's fine then!

PR #4756, opened 2025-03-05
fix(ffi): `Room::is_encrypted` is now async

Hywan:
This patch updates `Room::is_encrypted` to be async, otherwise it could block forever in case of bad network or offline network.

### DIFF
--- a/bindings/matrix-sdk-ffi/src/room.rs
+++ b/bindings/matrix-sdk-ffi/src/room.rs
@@ -257,8 +257,8 @@ impl Room {
         self.inner.room_id().to_string()
     }
 
-    pub fn is_encrypted(&self) -> Result<bool, ClientError> {
-        Ok(RUNTIME.block_on(self.inner.is_encrypted())?)
+    pub async fn is_encrypted(&self) -> Result<bool, ClientError> {
+        Ok(self.inner.is_encrypted().await?)
     }
 
     pub async fn members(&self) -> Result<Arc<RoomMembersIterator>, ClientError> {
--- a/bindings/matrix-sdk-ffi/CHANGELOG.md
+++ b/bindings/matrix-sdk-ffi/CHANGELOG.md
@@ -30,7 +30,10 @@ Breaking changes:
           or fails to present).
         - The rest of `AuthenticationError` is now found in the OidcError type.
 - `OidcAuthenticationData` is now called `OidcAuthorizationData`.
+
 - The `get_element_call_required_permissions` function now requires the device_id.
+
+- `Room::is_encrypted` is now async.
 
 Additions:
 


bindings/matrix-sdk-ffi/src/room.rs
```diff
@@ -257,8 +257,8 @@ impl Room {
         self.inner.room_id().to_string()
     }
 
-    pub fn is_encrypted(&self) -> Result<bool, ClientError> {
-        Ok(RUNTIME.block_on(self.inner.is_encrypted())?)
+    pub async fn is_encrypted(&self) -> Result<bool, ClientError> {
```
poljar, line 260: AFAIK Android will be quite unhappy about that. So please ask them before merging.

ganfra: Yeah, we do rely on this value to be synced, it's really unfortunate that this value is not part of `RoomInfo`

Hywan: Well, it's blocking the iOS team. It's a bit annoying. Can't it be async on your side (block the call or something)?

ganfra: The question is, could we get this value in `RoomInfo` as an optional? So this value can get refreshed automatically

Hywan: `RoomInfo::new` in `matrix-sdk-ffi` is async, so it should be possible, but I remember we have removed `is_encrypted` from it because it could have sent a request, which would create a lag in some cases. That's why we have extracted it in `Room::is_encrypted` directly.

See this patch from @jmartinesp: https://github.com/matrix-org/matrix-rust-sdk/commit/ad4d24f7d0bbc91dffe2c21efc2ce3492cbd6719

Hywan: Also, @jmartinesp did introduce this feature one patch earlier with https://github.com/matrix-org/matrix-rust-sdk/commit/004941b6b41ec4d8e4206035ec8c236e56d28ceb.

ganfra: Sure, thats why I'm asking if we can this value as optional, so if we don't have the info, just trigger a background request which updates the room info asynchronously

Hywan: @ganfra You are asking a field that is `Option<bool>`:

- `Some(true)` if it's synced and the room is encrypted
- `Some(false)` if it's synced and the room is NOT encrypted
- `None` if it's not synced (and you sync yourself)

Is that what you're asking?

Is the SDK user (so Android in this case) to run the request and update `RoomInfo` asynchronously? How do you imagine handling that?

ganfra: Yes it's what I'm asking, or it could also be an enum if it's clearer.

And yes, I'd expect the client to run the request whenever he needs to (ie. probably when retrieving `None`). I'd just expect the sdk to ensure only one request is made at a time.

ganfra: Then nice, one thing less to do. Just let the data goes into `RoomInfo` and we are good!

PR #4757, opened 2025-03-05
chore(ffi): Remove useless `block_on`

Hywan:
This patch removes the last `block_on` calls in `matrix-sdk-ffi`.

By reducing the number of blocking contexts, we are reducing the number of blocking threads, which is good performance wide. See https://github.com/matrix-org/matrix-rust-sdk/pull/4766 and https://github.com/matrix-org/matrix-rust-sdk/pull/4786.

---

* Address https://github.com/element-hq/element-x-ios/issues/3785


ganfra (approved):
LGTM

PR #4760, opened 2025-03-05
test(sdk): Add test for `maybe_apply_new_redaction`

Hywan:
Follow up of https://github.com/matrix-org/matrix-rust-sdk/pull/4740. I forgot to add a test!

This patch adds a test for `maybe_apply_new_redaction` when the redacted event is not loaded in-memory, i.e. when it lives in the store only.

---

### DIFF
--- a/crates/matrix-sdk/tests/integration/event_cache.rs
+++ b/crates/matrix-sdk/tests/integration/event_cache.rs
@@ -28,6 +28,12 @@ use ruma::{
 use serde_json::json;
 use tokio::{spawn, sync::broadcast, task::yield_now, time::sleep};
 
+macro_rules! assert_event_id {
+    ($timeline_event:expr, $event_id:literal) => {
+        assert_eq!($timeline_event.event_id().unwrap().as_str(), $event_id);
+    };
+}
+
 #[async_test]
 async fn test_must_explicitly_subscribe() {
     let server = MatrixMockServer::new().await;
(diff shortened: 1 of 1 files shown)


crates/matrix-sdk/tests/integration/event_cache.rs
```diff
@@ -1543,6 +1549,145 @@ async fn test_apply_redaction_when_redaction_comes_later() {
...
+    assert_matches!(update, RoomEventCacheUpdate::UpdateTimelineEvents { diffs, .. } => {
+        assert_eq!(diffs.len(), 1);
+
+        assert_matches!(&diffs[0], VectorDiff::Insert { index: 0, value: event } => {
+            assert_event_id!(event, "$ev0");
+            assert_matches!(
+                event.raw().deserialize().unwrap(),
+                AnySyncTimelineEvent::MessageLike(AnySyncMessageLikeEvent::RoomMessage(event)) => {
+                    // The event has been redacted!
+                    assert!(event.as_original().is_none());
```
poljar, line 1682:
```suggestion
                AnySyncTimelineEvent::MessageLike(event) => {
                    // The event has been redacted!
                    assert!(event.is_redacted());
```

Not quite the same, but if we only care about the fact that the event is redacted might be nicer to read.

poljar (approved):
Thanks for adding the forgotten test. I left a small suggestion but feel free to ignore.

crates/matrix-sdk/tests/integration/event_cache.rs
Hywan: Cool, didn't know about this one! Fixed up and rebased.

PR #4765, opened 2025-03-06
fix(multiverse): Don't wait for sync service state changes when shutting down

poljar:
The `SyncService::stop()` method guarantees that the sync service will be stopped after it has completed so there's no need to wait for state changes.

The state change might not even come, if you pressed `S` to stop the sync service manually.

### DIFF
--- a/labs/multiverse/src/main.rs
+++ b/labs/multiverse/src/main.rs
@@ -27,7 +27,7 @@ use matrix_sdk::{
 };
 use matrix_sdk_ui::{
     room_list_service::{self, filters::new_filter_non_left},
-    sync_service::{self, SyncService},
+    sync_service::SyncService,
     timeline::{TimelineItem, TimelineItemContent, TimelineItemKind, VirtualTimelineItem},
     Timeline as SdkTimeline,
 };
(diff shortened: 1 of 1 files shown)


bnjbvr (approved):
Hmm, I'm fine with this, since the service's going to be dropped anyways, but I kinda liked that we could prove it was shutting down. We *could* in theory maintain an atomic bool meaning "do we know the sync service is running", and toggle it when pressing `S`, so as to know whether or not it's worth waiting when exiting, but… maybe not worth doing.

LGTM!

poljar:
> Hmm, I'm fine with this, since the service's going to be dropped anyways, but I kinda liked that we could prove it was shutting down. We _could_ in theory maintain an atomic bool meaning "do we know the sync service is running", and toggle it when pressing `S`, so as to know whether or not it's worth waiting when exiting, but… maybe not worth doing.
> 
> LGTM!

Or you could use `subsribe_reset()` to get the current value as the first item in the stream. 

But once `stop()` stop finishes it's guaranteed that the service is stopped so I don't see the need for any proof.

PR #4766, opened 2025-03-06
feat(ffi): new way to provide a lightweight tokio runtime for memory-constrained apps

bnjbvr:
Creating many threads may use a bit of memory: on a machine with N
devices, exactly N*2 MB of memory may be consumed.

That might be a lot for a NSE process on iOS, which can only have up to
16 MB of RAM allocated for it. For this case, we introduce a new FFI
method `setup_lightweight_tokio_runtime` which will spawn at most 4
worker threads and 2 blocking threads. This should be sufficient for most
use cases.

---

* Address https://github.com/element-hq/element-x-ios/issues/3785

### DIFF
--- a/bindings/matrix-sdk-ffi/src/platform.rs
+++ b/bindings/matrix-sdk-ffi/src/platform.rs
@@ -358,6 +358,38 @@ pub fn setup_tracing(config: TracingConfiguration) {
         .init();
 }
 
+/// Set up a lightweight tokio runtime, for processes that have memory
+/// limitations (like the NSE process on iOS).
+#[matrix_sdk_ffi_macros::export]
+pub fn setup_lightweight_tokio_runtime() {
+    async_compat::set_runtime_builder(Box::new(|| {
+        eprintln!("spawning a lightweight tokio runtime");
+
+        // Get the number of available cores through the system, if possible.
+        let num_available_cores =
+            std::thread::available_parallelism().map(|n| n.get()).unwrap_or(1);
+
+        // The number of worker threads will be either that or 4, whichever is smaller.
+        let num_worker_threads = num_available_cores.min(4);
+
+        // Chosen by a fair dice roll.
+        let num_blocking_threads = 2;
+
+        // 1 MiB of memory per worker thread. Should be enough for everyone™.
+        let max_memory_bytes = 1024 * 1024;
+
+        let mut builder = tokio::runtime::Builder::new_multi_thread();
+
+        builder
+            .enable_all()
+            .worker_threads(num_worker_threads)
+            .thread_stack_size(max_memory_bytes)
+            .max_blocking_threads(num_blocking_threads);
+
+        builder
+    }));
+}
+
 #[cfg(test)]
 mod tests {
     use super::build_tracing_filter;
--- a/.deny.toml
+++ b/.deny.toml
@@ -10,7 +10,9 @@ exclude = [
 version = 2
 ignore = [
     { id = "RUSTSEC-2023-0071", reason = "We are not using RSA directly, nor do we depend on the RSA crate directly" },
+    { id = "RUSTSEC-2024-0436", reason = "Unmaintained paste crate, not critical." },
     { id = "RUSTSEC-2024-0384", reason = "Unmaintained backoff crate, not critical. We'll migrate soon." },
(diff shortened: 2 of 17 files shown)


bnjbvr:
That might not be sufficient, actually:

- the caller of the async-compat code is uniffi itself
- when uniffi calls, we haven't entered any tokio runtime yet: it's likely the first to request a runtime, for exported async functions
- so if async-compat tries to look if there's an live runtime, it won't see the one we're trying to create here, because it's not been entered *yet* (it might be later, if the callee makes use of `RUNTIME.something()`).

So we don't have a choice but to use our own fork of async-compat, but tweaked for the right number of cores, etc. We should likely make a PR to async-compat so we can provide a runtime builder function to be used as a default runtime, if none is available otherwise.

bindings/matrix-sdk-ffi/src/platform.rs
```diff
@@ -358,6 +358,41 @@ pub fn setup_tracing(config: TracingConfiguration) {
...
+
+        // Chosen by a fair dice roll.
+        let num_blocking_threads = 1;
+
+        // 1 MiB of memory per worker thread. Should be enough for everyone™.
+        let max_memory_bytes = 1024 * 1024;
+
+        // Create a multithreaded runtime with this number of worker threads, each
+        // having access to 2 MiB of stack space (this stack space size is the
+        // default, but we keep it explicit here).
```
stefanceriu, line 383: This part of the comment can go I guess

bindings/matrix-sdk-ffi/src/platform.rs
```diff
@@ -358,6 +358,41 @@ pub fn setup_tracing(config: TracingConfiguration) {
...
+pub fn setup_lightweight_tokio_runtime() {
+    async_compat::set_runtime_builder(Box::new(|| {
+        eprintln!("spawning a lightweight tokio runtime");
+
+        // Get the number of available cores through the system, if possible.
+        let num_available_cores =
+            std::thread::available_parallelism().map(|n| n.get()).unwrap_or(1);
+
+        // The number of worker threads will be either that or 4, whichever is smaller.
+        let num_worker_threads = usize::min(num_available_cores, 4);
```
poljar, line 373: Why not `cmp:min()`? I think this one is expected to be used as `a.min(b)`, at least that's what the examples show.

poljar (approved):
Looks good.

bindings/matrix-sdk-ffi/src/platform.rs
bnjbvr: my mathematical brain wants `function(param1, param2)` instead of `param1.function(param2)`, in this case, but both work.

bindings/matrix-sdk-ffi/src/platform.rs
```diff
@@ -358,6 +358,38 @@ pub fn setup_tracing(config: TracingConfiguration) {
...
+        eprintln!("spawning a lightweight tokio runtime");
+
+        // Get the number of available cores through the system, if possible.
+        let num_available_cores =
+            std::thread::available_parallelism().map(|n| n.get()).unwrap_or(1);
+
+        // The number of worker threads will be either that or 4, whichever is smaller.
+        let num_worker_threads = num_available_cores.min(4);
+
+        // Chosen by a fair dice roll.
```
Hywan, line 375: 🎲

jplatte:
AFAIK the number of blocking threads in the default configuration is a lot larger than the number of worker threqds - are you sure having just a single blocking thread is a good idea? Note that a lot of async functionality internally spawns blocking tasks (for example for filesystem access). I would imagine 1 worker, 4 blocking to perform a lot better (though I am not certain how tokio spawn these threads - whether blocking threads have a smaller or larger stack size).

bnjbvr:
> AFAIK the number of blocking threads in the default configuration is a lot larger than the number of worker threads - are you sure having just a single blocking thread is a good idea? Note that a lot of async functionality internally spawns blocking tasks (for example for filesystem access). I would imagine 1 worker, 4 blocking to perform a lot better (though I am not certain how tokio spawn these threads - whether blocking threads have a smaller or larger stack size).

Good question; for the lightweight app we're interested in (EX's NSE on iOS — say that thrice), i/o performance shouldn't be a bottleneck, but we can tweak those values easily if needs be.

PR #4769, opened 2025-03-07
feat(timeline): Provide is_utd on EventTimelineItem

andybalaam:
Adds a little utility method. I didn't add tests because the nearby similar methods don't have them, but happy to add them if we want.

See https://github.com/matrix-org/matrix-rust-sdk/pull/4644#discussion_r1984968706 for a context where this is useful.

### DIFF
--- a/crates/matrix-sdk-ui/src/timeline/event_item/mod.rs
+++ b/crates/matrix-sdk-ui/src/timeline/event_item/mod.rs
@@ -233,6 +233,11 @@ impl EventTimelineItem {
         matches!(self.kind, EventTimelineItemKind::Remote(_))
     }
 
+    /// Check whether this item's content is UTD (unable to decrypt).
+    pub fn is_utd(&self) -> bool {
+        matches!(self.content(), TimelineItemContent::UnableToDecrypt(_))
+    }
+
     /// Get the `LocalEventTimelineItem` if `self` is `Local`.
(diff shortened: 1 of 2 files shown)


poljar (approved):
Looks good, but we should start to use this method:

https://github.com/matrix-org/matrix-rust-sdk/blob/53853c2d9a2cc385e2a9307b24bf5fb5a89814f3/crates/matrix-sdk-ui/src/timeline/tests/encryption.rs#L751

and 

https://github.com/matrix-org/matrix-rust-sdk/blob/53853c2d9a2cc385e2a9307b24bf5fb5a89814f3/testing/matrix-sdk-integration-testing/src/tests/sliding_sync/room.rs#L925

andybalaam:
> https://github.com/matrix-org/matrix-rust-sdk/blob/53853c2d9a2cc385e2a9307b24bf5fb5a89814f3/crates/matrix-sdk-ui/src/timeline/tests/encryption.rs#L751

Done

> https://github.com/matrix-org/matrix-rust-sdk/blob/53853c2d9a2cc385e2a9307b24bf5fb5a89814f3/testing/matrix-sdk-integration-testing/src/tests/sliding_sync/room.rs#L925

This is not an `EventTimelineItem` unfortunately.

poljar:
> This is not an `EventTimelineItem` unfortunately.

🥲

PR #4772, opened 2025-03-08
refactor(sdk): Merge session tokens

zecakeh:
Since both authentication APIs now have identical session tokens type, we can merge them an put the data in `AuthCtx`.

I decided to not reuse a `SharedObservable` because the `session_tokens_stream` behavior can be done by subscribing to `Client::subscribe_to_session_changes()` and then calling `Client::session_tokens()`.

### DIFF
--- a/crates/matrix-sdk/src/authentication/oidc/tests.rs
+++ b/crates/matrix-sdk/src/authentication/oidc/tests.rs
@@ -33,9 +33,8 @@ use crate::{
     },
     test_utils::{
         client::{
-            oauth::{
-                mock_client_metadata, mock_prev_session_tokens, mock_session, mock_session_tokens,
-            },
+            mock_prev_session_tokens_with_refresh, mock_session_tokens_with_refresh,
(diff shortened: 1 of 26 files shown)


crates/matrix-sdk/src/authentication/oidc/tests.rs
```diff
@@ -450,18 +449,24 @@ async fn test_insecure_clients() -> anyhow::Result<()> {
...
+        assert_eq!(
+            session_changes.try_recv(),
+            Ok(SessionChange::TokensRefreshed),
+            "The session changes should be notified of the tokens refresh"
+        );
+        assert_eq!(
+            session_changes.try_recv(),
+            Err(TryRecvError::Empty),
+            "There should be no more session changes"
+        );
```
Hywan, line 469: I think you can use `session_changes.is_empty()` too in this case, but that's fine, just a side little note.

Hywan (approved):
I applaude this very welcomed clean up. Excellent work as usual. Thank you for your time and skills!

PR #4773, opened 2025-03-09
test(sdk): Allow any MockEndpoint to override the expected access token

zecakeh:
Generalizes the behavior introduced in #4767 for the `/whoami` mock endpoint.

It requires creators of `MockEndpoint`s to use the built-in methods instead of custom methods on `MockBuilder` and sadly there is no way to enforce this programmatically.


bnjbvr (approved):
Nice, thanks for the cleanups!

PR #4777, opened 2025-03-10
feat: Introduce `EncryptionState`

Hywan:
This patch introduces the new `EncryptionState` to represent the 3
possible states: `Encrypted`, `NotEncrypted` or `Unknown`. All the
`is_encrypted` methods have been replaced by `encryption_state`.
The most noticable change is in `matrix_sdk::Room` where `async fn
is_encrypted(&self) -> Result<bool>` has been replaced by `fn fn
encryption_state(&self) -> EncryptionState`. However, a new `async
fn latest_encryption_state(&self) -> Result<EncryptionState>` method
“restores” the previous behaviour by calling `request_encryption_state`
if necessary.

The idea is that the caller is now responsible to call
`request_encryption_state` if desired, or use `latest_encryption_state`
to automate the call if necessary. `encryption_state` is now non-async
and infallible everywhere.

`matrix-sdk-ffi` has been updated but no methods have been added for
the moment.

---

* Replaces https://github.com/matrix-org/matrix-rust-sdk/pull/4756


crates/matrix-sdk/CHANGELOG.md
```diff
@@ -21,17 +21,36 @@ simpler methods:
...
+  room.is_encrypted().await?
+  ```
+
+  by
+
+  ```rust
+  room.latest_encryption_state().await?.is_encrypted()
+  ```
+
+>>>>>>> b6c5e758bd (doc: Update the `CHANGELOG.md`s.)
```
poljar, line 53: What's that? :sweat_smile:

crates/matrix-sdk-base/src/rooms/normal.rs
```diff
@@ -3589,8 +3612,32 @@ mod tests {
...
+        });
+
+        assert_matches!(room.encryption_state(), EncryptionState::NotEncrypted);
+    }
+
+    #[test]
+    fn test_encryption_state() {
+        assert!(EncryptionState::Unknown.is_unknown());
+        assert!(EncryptionState::Encrypted.is_unknown().not());
+        assert!(EncryptionState::NotEncrypted.is_unknown().not());
```
poljar, line 3636: I will `not` tell anyone about this :zipper_mouth_face:

crates/matrix-sdk-base/src/rooms/normal.rs
```diff
@@ -1576,9 +1566,15 @@ impl RoomInfo {
...
-        self.base_info.encryption.is_some()
+    /// Returns the encryption state of this room.
+    pub fn encryption_state(&self) -> EncryptionState {
+        if !self.encryption_state_synced {
+            EncryptionState::Unknown
+        } else if self.base_info.encryption.is_some() {
+            EncryptionState::Encrypted
+        } else {
+            EncryptionState::NotEncrypted
+        }
```
poljar, line 1577: It would probably be nice to replace `encryption_state_synced` and `base_info.encryption`with `EncryptionState` where `Encrypted` keeps the `m.room.encryption` content as associated data.

But that's probably something for a different PR and might not even be worth the trouble.

poljar (approved):
Looks good, you left in a merge conflict marker which we should get rid of, but that's about it.

crates/matrix-sdk-base/src/rooms/normal.rs
Hywan: Talk about that in a meeting and agreed to address that in another PR.

crates/matrix-sdk/CHANGELOG.md
Hywan: oopsy

PR #4778, opened 2025-03-10
feat: Always know when a room is encrypted or not (with sliding sync)

Hywan:
<figure role="presentation">
<p align="center"></p>
<figcaption><p align="center"><a href="https://commons.wikimedia.org/wiki/File:John_Collier_-_Priestess_of_Delphi_-_Google_Art_Project.jpg">Priestess of Delphi (John Collier)</a></p></figcaption>
</figure>

---

So, we have a problem with encryption state.

First off, a patch has clarified the different states for encryption, see https://github.com/matrix-org/matrix-rust-sdk/pull/4777.

Second, the problem is pretty basic to understand:

- the _presence_ of a `m.room.encryption` state event for a room in a sync response means the encryption state for this room is `Encrypted`,
- the _absence_ of a `m.room.encryption` state event means two things:

  1. if the `m.room.encryption` state event was requested (via `required_state` in MSC4186, aka (simplified) sliding sync), it means the encryption state is `NotEncrypted`,
  2. if the`m.room.encryption` state event was **not** requested, it means the encryption state is `Unknown`; indeed, there is no way to interpret the meaning of the absence of this event.

So far, the presence of `m.room.encryption` was always interpreted as `Encrypted`, which is fine. However, the absence was always interpreted as `Unknown`, which is _not incorrect_, but it means a request to [`/…/state/m.room.encryption/`](https://spec.matrix.org/latest/client-server-api/#get_matrixclientv3roomsroomidstateeventtypestatekey) (with `matrix_sdk::Room::request_encryption_state`) was necessary to resolve that. This implies a network call, and it creates slowness, or worst, in offline mode, it can block some clients if not used appropriately (note: this method is async, but it can be misused)!

With this patch, when sliding sync is used, the `Unknown` is impossible to get if `m.room.encryption` is present in `required_state`. Thus, the encryption state for a room is always `Encrypted` or `NotEncrypted` with `matrix_sdk_ui::RoomListService` or `matrix_sdk_ui::NotificationClient` for example.

---

* Fix https://github.com/element-hq/element-x-ios/issues/3857

### DIFF
--- a/crates/matrix-sdk-base/src/rooms/normal.rs
+++ b/crates/matrix-sdk-base/src/rooms/normal.rs
@@ -1582,20 +1582,41 @@ impl RoomInfo {
         self.base_info.encryption = event;
     }
 
+    /// Handle the encryption state.
+    pub fn handle_encryption_state(
+        &mut self,
+        requested_required_states: &[(StateEventType, String)],
+    ) {
+        if requested_required_states
+            .iter()
+            .any(|(state_event, _)| state_event == &StateEventType::RoomEncryption)
+        {
+            // The `m.room.encryption` event was requested during the sync. Whether we have
+            // received a `m.room.encryption` event in return doesn't matter: we must mark
+            // the encryption state as synced; if the event is present, it means the room
+            // _is_ encrypted, otherwise it means the room _is not_ encrypted.
+
+            self.mark_encryption_state_synced();
+        }
+    }
+
     /// Handle the given state event.
     ///
     /// Returns true if the event modified the info, false otherwise.
     pub fn handle_state_event(&mut self, event: &AnySyncStateEvent) -> bool {
-        let ret = self.base_info.handle_state_event(event);
+        // Store the state event in the `BaseRoomInfo` first.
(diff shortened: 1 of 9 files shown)


crates/matrix-sdk/src/sliding_sync/mod.rs
```diff
@@ -543,6 +547,7 @@ impl SlidingSync {
         debug!("Sending request");
 
         // Prepare the request.
+        let requested_required_states = RequestedRequiredStates::from(&request);
```
Hywan, line 550: The source of the magic is here.

crates/matrix-sdk-base/src/rooms/normal.rs
```diff
@@ -1582,20 +1582,41 @@ impl RoomInfo {
         self.base_info.encryption = event;
     }
 
+    /// Handle the encryption state.
+    pub fn handle_encryption_state(
```
Hywan, line 1586: The rest of the magic is here.

stefanceriu (approved):
Very nice! Lots of little changes that I just skimmed but the core logic lgtm! 👍

PR #4779, opened 2025-03-10
fix(event cache): don't remove a gap when it's the only chunk in memory

### LINKED ISSUE #4784: event cache: having any ignored user will aggressively clear all rooms' caches
If the sync returns a state event of the kind `IgnoredUserList`, then this triggers the observable `ignore_user_list_changes` in the base client:

https://github.com/matrix-org/matrix-rust-sdk/blob/c8536e9e46cc99b946db50f20838a0de928ee143/crates/matrix-sdk-base/src/client.rs#L1294-L1301

Unfortunately, this will cause a clear of *all* the rooms in the event cache:

https://github.com/matrix-org/matrix-rust-sdk/blob/c8536e9e46cc99b946db50f20838a0de928ee143/crates/matrix-sdk/src/event_cache/mod.rs#L269-L274

This might not be an issue, if only the initial sync contained the ignored-user-list event, and the event cache has been subscribed to *after* the first initial sync has been received. But in general, this is a footgun that would be nice to avoid. Things I can imagine to solve this:

- only trigger the observable when the value changed, compared to a previous value we knew about. Meaning we'd compare the new value with the one stored in the database.
- equivalently, maintain a list of ignored users in the event cache, and only clear the room if this list has changed

Likely a blocker for #3280.

cc @Hywan

### LINKED ISSUE #4785: event cache: running into unique constraint failures
Running `multiverse` locally with a big test account that has some ignored users, I can easily trigger unique constraint failures, for room gaps:

- start multiverse
- let it run in the background for a couple minutes
- shut it down
- stash the rust logs somewhere
- restart it
- let it run in the background for a couple minutes

as it's triggering #4784, this will clear the storage. But! in the logs of the second run, I can see issues related to unique constraints failures:

> 2025-03-11T08:26:30.763163Z ERROR listen_task:handle_room_updates: matrix_sdk::event_cache: handling joined room update: UNIQUE constraint failed: gaps.room_id, gaps.chunk_id room_id=!abcdef:matrix.org

Blocker for #3280.

cc @Hywan

bnjbvr:
See the commit messages for more explanations. This makes the event cache much more stable for `multiverse`, by addressing three issues:

- if a gap was the only chunk loaded in a linked chunk, don't try to remove it after resolving it to an empty set of events
- #4785: enable the foreign keys pragma for each sqlite connection, to make sure cascading works
- #4784 (perf): now we'll only wake the subscribers if the ignored user list has changed, when compared to the previous version we knew about

Bugs found with `multiverse`.

Fixes #4785.
Fixes #4784.

Part of #3280.

### DIFF
--- a/crates/matrix-sdk-base/src/client.rs
+++ b/crates/matrix-sdk-base/src/client.rs
@@ -1254,9 +1254,10 @@ impl BaseClient {
 
         {
             let _sync_lock = self.sync_lock().lock().await;
+            let prev_ignored_user_list = self.load_previous_ignored_user_list().await;
             self.store.save_changes(&changes).await?;
             *self.store.sync_token.write().await = Some(response.next_batch.clone());
-            self.apply_changes(&changes, room_info_notable_updates);
+            self.apply_changes(&changes, room_info_notable_updates, prev_ignored_user_list);
         }
 
         // Now that all the rooms information have been saved, update the display name
@@ -1286,10 +1287,17 @@ impl BaseClient {
         Ok(response)
     }
 
+    pub(crate) async fn load_previous_ignored_user_list(
+        &self,
+    ) -> Option<Raw<IgnoredUserListEvent>> {
+        self.store().get_account_data_event_static().await.ok().flatten()
+    }
+
     pub(crate) fn apply_changes(
         &self,
         changes: &StateChanges,
         room_info_notable_updates: BTreeMap<OwnedRoomId, RoomInfoNotableUpdateReasons>,
+        prev_ignored_user_list: Option<Raw<IgnoredUserListEvent>>,
     ) {
         if let Some(event) = changes.account_data.get(&GlobalAccountDataEventType::IgnoredUserList)
         {
@@ -1298,8 +1306,27 @@ impl BaseClient {
                     let user_ids: Vec<String> =
                         event.content.ignored_users.keys().map(|id| id.to_string()).collect();
 
-                    self.ignore_user_list_changes.set(user_ids);
+                    // Try to only trigger the observable if the ignored user list has changed,
+                    // from the previous time we've seen it. If we couldn't load the previous event
+                    // for any reason, always trigger.
+                    if let Some(prev_user_ids) =
+                        prev_ignored_user_list.and_then(|raw| raw.deserialize().ok()).map(|event| {
+                            event
+                                .content
+                                .ignored_users
+                                .keys()
+                                .map(|id| id.to_string())
+                                .collect::<Vec<_>>()
+                        })
+                    {
+                        if user_ids != prev_user_ids {
+                            self.ignore_user_list_changes.set(user_ids);
+                        }
+                    } else {
+                        self.ignore_user_list_changes.set(user_ids);
+                    }
                 }
+
                 Err(error) => {
                     error!("Failed to deserialize ignored user list event: {error}")
                 }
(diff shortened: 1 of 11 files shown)


crates/matrix-sdk/src/event_cache/room/events.rs
```diff
@@ -149,7 +150,24 @@ impl RoomEvents {
...
+
+            // If there's no chunks at all, then we won't be able to find the gap chunk.
+            let _ =
+                it.next().ok_or(Error::InvalidChunkIdentifier { identifier: gap_identifier })?;
+
+            // If there's no next chunk, we can conclude there's only one.
+            it.next().is_none()
+        };
+
+        let next_pos = if events.is_empty() && !has_only_one_chunk {
```
Hywan, line 170: :+1:

crates/matrix-sdk-sqlite/src/event_cache_store.rs
```diff
@@ -1915,6 +1915,27 @@ mod tests {
         let chunks = store.load_all_chunks(room_id).await.unwrap();
         assert!(chunks.is_empty());
 
+        // Check that cascading worked. Yes, sqlite, I doubt you.
```
Hywan, line 1926: How dare you?

crates/matrix-sdk-sqlite/src/event_cache_store.rs
```diff
@@ -156,7 +156,15 @@ impl SqliteEventCacheStore {
...
 
     async fn acquire(&self) -> Result<SqliteAsyncConn> {
-        Ok(self.pool.get().await?)
+        let connection = self.pool.get().await?;
+
+        // Per https://www.sqlite.org/foreignkeys.html#fk_enable, foreign key support must be
+        // enabled on a per-connection basis. Execute it every time we try to get a
+        // connection, since we can't guarantee a previous connection did
+        // enabled it before.
+        connection.execute_batch("PRAGMA foreign_keys = ON;").await?;
```
Hywan, line 165: :+1:

crates/matrix-sdk-base/src/client.rs
```diff
@@ -1298,8 +1306,27 @@ impl BaseClient {
...
+                            event
+                                .content
+                                .ignored_users
+                                .keys()
+                                .map(|id| id.to_string())
+                                .collect::<Vec<_>>()
+                        })
+                    {
+                        if user_ids != prev_user_ids {
+                            self.ignore_user_list_changes.set(user_ids);
```
Hywan, line 1323: It may be a good use case for [`update_if`](https://docs.rs/eyeball/latest/eyeball/struct.Observable.html#method.update_if) here, though it's not blocking the approval of this patch.

Hywan (approved):
Excellent, thanks for the bug fixes!

crates/matrix-sdk/src/event_cache/pagination.rs
```diff
@@ -367,17 +367,15 @@ impl RoomPagination {
...
-                    // prior gap that we just filled.
-                    trace!("removing previous gap, as all events have been deduplicated");
-                    room_events.remove_empty_chunk_at(gap_id).expect("gap identifier is a valid gap chunk id we read previously")
-                } else {
-                    trace!("replacing previous gap with the back-paginated events");
-
-                    // Replace the gap with the events we just deduplicated.
-                    room_events.replace_gap_at(reversed_events.clone(), gap_id)
-                        .expect("gap_identifier is a valid chunk id we read previously")
+                    assert!(events.is_empty());
```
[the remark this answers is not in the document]
bnjbvr: both are the same, right?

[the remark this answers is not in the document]
bnjbvr: you're right it's better :+1:

PR #4780, opened 2025-03-10
Invites room summary fallback

stefanceriu:
This PR introduces a fallback mechanism for dealing with invites when the room summary API [MSC3266](https://github.com/matrix-org/matrix-spec-proposals/pull/3266) isn't enabled on the homeserver. 
It does so by trying to use whatever data the client has cached about the room, and even though it won't contain a lot of information, it will still provide an usable instance which can be used to accept/decline the invite and partially populate the join room screen (in EX).

It also deals away with the now unnecessary `invited_room` and `inviter` and expose a new client one for fetching arbitrary rooms based on their identifier.

### DIFF
--- a/crates/matrix-sdk/tests/integration/room_preview.rs
+++ b/crates/matrix-sdk/tests/integration/room_preview.rs
@@ -1,5 +1,9 @@
+use assert_matches::assert_matches;
 use js_int::uint;
-use matrix_sdk::{config::SyncSettings, test_utils::logged_in_client_with_server};
+use matrix_sdk::{
+    config::SyncSettings,
+    test_utils::{logged_in_client_with_server, mocks::MatrixMockServer},
+};
 use matrix_sdk_base::RoomState;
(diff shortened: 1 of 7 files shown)


crates/matrix-sdk/tests/integration/room_preview.rs
```diff
@@ -50,6 +50,28 @@ async fn test_room_preview_leave_invited() {
...
+    server.reset().await;
+
+    mock_leave(room_id, &server).await;
+
+    let room_preview = client.get_room_preview(room_id.into(), Vec::new()).await.unwrap();
+    assert_eq!(room_preview.state.unwrap(), RoomState::Invited);
+
+    client.get_room(room_id).unwrap().leave().await.unwrap();
+
+    assert_eq!(client.get_room(room_id).unwrap().state(), RoomState::Left);
```
bnjbvr, line 72: What is this last line checking? It doesn't use the preview state, but the room state…

PR #4782, opened 2025-03-11
refactor(timeline): move `is_utd()` to `TimelineItemContent`

bnjbvr:
It's unusual to have the method on the parent type when the field type could also hold the method. In fact, this was the only bool getter inspecting the timeline's content, so let's move the method next to as its siblings, for consistency, and let's spell it out fully for clarity.


stefanceriu (approved):
👌

PR #4787, opened 2025-03-11
feat(base): Reduce memory usage of `BaseClient::room_info_notable_update_sender` by 65'536 times for a regular user

Hywan:
This patch reduces the memory usage of the broadcast channel used by
`BaseClient::room_info_notable_update_sender`. So far, its size was
`u16::MAX`. Considering `RoomInfoNotableUpdate` is 24 bytes, the channel
was allocating 1.5Mb of memory, which is way too much. It is creating
problems on systems where the process has limited resources, like the
[Notification Service Extension][NSE] on iOS.

For a regular users with 200 rooms, the memory usage becomes 24Kb, which
is 65'536 times less.

[NSE]: https://developer.apple.com/documentation/usernotifications/unnotificationserviceextension

---

* Follow up of https://github.com/matrix-org/matrix-rust-sdk/pull/4013

### DIFF
--- a/crates/matrix-sdk-base/src/client.rs
+++ b/crates/matrix-sdk-base/src/client.rs
@@ -148,11 +148,24 @@ impl BaseClient {
     /// * `config` - An optional session if the user already has one from a
     ///   previous login call.
     pub fn with_store_config(config: StoreConfig) -> Self {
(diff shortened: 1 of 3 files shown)


stefanceriu (approved):
🚀

PR #4789, opened 2025-03-11
refactor(oidc): Remove dependency on mas-oidc-client 

zecakeh:
To get rid of mas-oidc-client, we need to import a few methods and types, but nothing very complex.

I decided to create a more opinionated `ClientMetadata` type so that it is easy to construct and we don't need to validate it.

### DIFF
--- a/crates/matrix-sdk/Cargo.toml
+++ b/crates/matrix-sdk/Cargo.toml
@@ -48,9 +48,7 @@ experimental-oidc = [
     "ruma/unstable-msc2967",
     "ruma/unstable-msc4108",
     "dep:language-tags",
-    "dep:mas-oidc-client",
     "dep:sha2",
-    "dep:tower",
     "dep:oauth2",
 ]
 experimental-widgets = ["dep:language-tags", "dep:uuid"]
@@ -81,7 +79,6 @@ imbl = { workspace = true, features = ["serde"] }
 indexmap = { workspace = true }
 js_int = "0.2.2"
 language-tags = { version = "0.3.2", optional = true }
-mas-oidc-client = { version = "0.11.0", default-features = false, optional = true }
 matrix-sdk-base = { workspace = true }
 matrix-sdk-common = { workspace = true }
 matrix-sdk-ffi-macros = { workspace = true, optional = true }
--- a/crates/matrix-sdk/src/authentication/oidc/error.rs
+++ b/crates/matrix-sdk/src/authentication/oidc/error.rs
@@ -61,11 +61,10 @@ pub enum OidcError {
     #[error("authorization server discovery failed: {0}")]
     Discovery(#[from] OauthDiscoveryError),
 
-    /// The OpenID Connect Provider doesn't support dynamic client registration.
-    ///
-    /// The provider probably offers another way to register clients.
-    #[error("no dynamic registration support")]
-    NoRegistrationSupport,
+    /// An error occurred when registering the client with the authorization
+    /// server.
+    #[error("client registration failed: {0}")]
+    ClientRegistration(#[from] OauthClientRegistrationError),
 
     /// The client has not registered while the operation requires it.
     #[error("client not registered")]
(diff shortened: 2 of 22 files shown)


crates/matrix-sdk/src/authentication/oidc/http_client.rs
```diff
@@ -0,0 +1,191 @@
...
+
+                let mut uri_parts = request.uri().clone().into_parts();
+                uri_parts.scheme = Some(http::uri::Scheme::HTTP);
+                *request.uri_mut() = http::uri::Uri::from_parts(uri_parts)
+                    .expect("reconstructing URI from parts should work");
+
+                request
+            } else {
+                request
+            };
```
Hywan, line 57: Why do you need this for testing?

crates/matrix-sdk/src/authentication/oidc/http_client.rs
```diff
@@ -0,0 +1,191 @@
...
+) -> Result<(), RequestTokenError<HttpClientError<reqwest::Error>, T>> {
+    let Some(content_type) = http_response.headers().get(http::header::CONTENT_TYPE) else {
+        return Ok(());
+    };
+
+    if content_type
+        .to_str()
+        // Check only the beginning of the content type, because there might be extra
+        // parameters, like a charset.
+        .is_ok_and(|ct| ct.to_lowercase().starts_with(mime::APPLICATION_JSON.essence_str()))
```
Hywan, line 99: A `Content-Type: application/jsonfoo` would pass this test. Thinking about the `;charset` is smart, but I don't think it's safe that way.

crates/matrix-sdk/src/authentication/oidc/error.rs
```diff
@@ -249,3 +248,56 @@ pub enum OauthTokenRevocationError {
...
+    /// The value of one of the client metadata fields is invalid and the server
+    /// has rejected this request.
+    InvalidClientMetadata,
+
+    /// The software statement presented is invalid.
+    InvalidSoftwareStatement,
+
+    /// The software statement presented is not approved for use by this
+    /// authorization server.
+    UnapprovedSoftwareStatement,
```
Hywan, line 283: Did you remove this type of action in the previous commit? I wonder if this error becomes useless then.

crates/matrix-sdk/src/authentication/oidc/registration.rs
```diff
@@ -80,3 +89,375 @@ pub struct ClientRegistrationResponse {
...
+/// different values are supported by this API. To set other fields, use your
+/// own type or construct directly the JSON representation.
+///
+/// The original format is defined in [RFC 7591].
+///
+/// [MSC2966]: https://github.com/matrix-org/matrix-spec-proposals/pull/2966
+/// [RFC 7591]: https://datatracker.ietf.org/doc/html/rfc7591
+#[derive(Debug, Clone, Serialize)]
+#[serde(into = "ClientMetadataSerializeHelper")]
+pub struct ClientMetadata {
```
Hywan, line 107: Nice!

crates/matrix-sdk/src/authentication/oidc/registration.rs
```diff
@@ -80,3 +89,375 @@ pub struct ClientRegistrationResponse {
...
+                client_name,
+                logo_uri,
+                policy_uri,
+                tos_uri,
+            },
+        }
+    }
+}
+
+struct ClientMetadataLocalizedFields {
```
Hywan, line 302: Why do you need this type?

crates/matrix-sdk/Cargo.toml
```diff
@@ -81,7 +79,6 @@ imbl = { workspace = true, features = ["serde"] }
 indexmap = { workspace = true }
 js_int = "0.2.2"
 language-tags = { version = "0.3.2", optional = true }
-mas-oidc-client = { version = "0.11.0", default-features = false, optional = true }
```
Hywan, line 84: w00t!

Hywan (changes requested):
I truly appreciate this kind of patches. Clear. Tested. Well documented. Thank you!

While I liked the effort put in `mas-oidc-client`, it is correct that we save many dependencies (and probably many lines of code) by removing it from our tree.

I let a couple of feedbacks, but nothing blocking.

crates/matrix-sdk/src/authentication/oidc/http_client.rs
zecakeh: The oauth2 crate checks that the revocation has an HTTPS scheme, so we have to work around that for the mock server.

Note that I opened an issue to know if we could have a method to avoid the check but they said to handle it with a custom HTTP client.

crates/matrix-sdk/src/authentication/oidc/error.rs
zecakeh: I was wondering about it too, but I decided to just copy the possible values from the RFC.

Technically there is also nothing is preventing clients from adding a software statement in the `Raw<ClientMetadata>`.

But in the end I don't have strong opinion about this.

crates/matrix-sdk/src/authentication/oidc/registration.rs
zecakeh: Because (de)serialization of `Localized` values is complicated so we need to implement it manually. But we don't need to things manually for other fields so we just group the localizable fields in a struct and handle their serialization separately.

crates/matrix-sdk/src/authentication/oidc/http_client.rs
zecakeh: I don't think that this is actually a safety issue. It's just a code path to error quickly if we didn't get the format that we expected.

Nothing prevents the server from announcing that they are sending JSON but send something else entirely. There is no point in making up a weird content-type to circumvent that check.

crates/matrix-sdk/src/authentication/oidc/http_client.rs
Hywan: Okay, thanks.

crates/matrix-sdk/src/authentication/oidc/error.rs
Hywan: Let's keep it then.

crates/matrix-sdk/src/authentication/oidc/registration.rs
Hywan: Great, thanks. It may deserves a small documentation line to explain that maybe?

Hywan:
Please resolve the conflicts and we are good!

crates/matrix-sdk/src/authentication/oidc/registration.rs
zecakeh: I added docs to the type

zecakeh:
I rebased on main to solve the conflicts

crates/matrix-sdk/src/authentication/oidc/registration.rs
Hywan: Thanks!

Hywan (approved):

PR #4790, opened 2025-03-12
feat!(ffi): rename `setup_tracing` into `init_platform`

bnjbvr:
And make it take a boolean indicating whether we want to set up a lightweight tokio runtime or not, instead of having `setup_lightweight_tokio_runtime` as a public function + another function, both of which would have to be called anyways.

cc @stefanceriu @jmartinesp


poljar (approved):

Hywan (approved):
LGTM!

Issue #4762, opened 2025-03-06
Only the initial chunk of events gets decrypted if events are loaded from the event cache store.


poljar:
When using the event cache with storage enabled, events are persisted in the database. Since room keys may not always arrive in the same sync as the event—sometimes arriving later—it is expected that m.room.encrypted events will be stored in the event cache.

Upon reloading and inserting these events into the timeline, the initial chunk of events will get decrypted during timeline construction:

https://github.com/matrix-org/matrix-rust-sdk/blob/6c9b1ef3c1c8c493c1620f566207d6786c43b133/crates/matrix-sdk-ui/src/timeline/builder.rs#L490-L495

However, when additional events are loaded from the cache, no decryption logic is exists. As a result, undecryptable events (UTDs) will not fix themselves unless the application explicitly requests a redecryption.

PR #4794, opened 2025-03-12
fix(ui): Decrypt `TimelineEventKind::UnableToDecrypt` coming from the event cache

Hywan:
The fix is 3 LOC, the test is 345 LOC, oh yeah.

So, currently, it is possible to store an `TimelineEventKind::UnableToDecrypt` (UTD) event inside the event cache. When received by the `Timeline`, prior this patch, such event wasn't decrypted. The fix is actually quite small: everything that comes from the cache (`EventsOrigin::Cache`) and that contains a UTD, triggers a redecryption. Hoping that the keys are in the `OlmMachine` of course.

This patch tests two situations where it can happen: when the UTD is part of the initial items, and when the UTD is part of a pagination. The former was already working, but not tested via the event cache as far as I know, and the latter wasn't tested at all (because it wasn't implemented).

---

* Fix https://github.com/matrix-org/matrix-rust-sdk/issues/4762
* Fix https://github.com/element-hq/element-x-ios/issues/3858

### DIFF
--- a/crates/matrix-sdk-ui/src/timeline/builder.rs
+++ b/crates/matrix-sdk-ui/src/timeline/builder.rs
@@ -393,6 +393,10 @@ async fn room_event_cache_updates_task(
                         },
                     )
                     .await;
+
+                if matches!(origin, EventsOrigin::Cache) {
+                    timeline_controller.retry_event_decryption(None).await;
+                }
             }
 
             RoomEventCacheUpdate::AddEphemeralEvents { events } => {
--- a/crates/matrix-sdk-ui/src/timeline/builder.rs
+++ b/crates/matrix-sdk-ui/src/timeline/builder.rs
@@ -448,8 +448,6 @@ where
     pin_mut!(stream);
 
     while let Some(update) = stream.next().await {
-        let room = timeline_controller.room();
-
         match update {
             Ok(info) => {
                 let mut session_ids = BTreeSet::new();
@@ -458,10 +456,10 @@ where
                     session_ids.extend(set);
                 }
 
-                timeline_controller.retry_event_decryption(room, Some(session_ids)).await;
+                timeline_controller.retry_event_decryption(Some(session_ids)).await;
             }
             // We lagged, so retry every event.
-            Err(_) => timeline_controller.retry_event_decryption(room, None).await,
+            Err(_) => timeline_controller.retry_event_decryption(None).await,
         }
     }
 }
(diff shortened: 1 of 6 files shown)


crates/matrix-sdk-ui/src/timeline/builder.rs
```diff
@@ -393,6 +393,10 @@ async fn room_event_cache_updates_task(
                         },
                     )
                     .await;
+
+                if matches!(origin, EventsOrigin::Cache) {
+                    timeline_controller.retry_event_decryption(None).await;
+                }
```
poljar, line 399: Hmm, as far as I understand, this gets called once per event, while `retry_event_decryption(None)` tries to decrypt all the events in the timeline.

I wouldn't be surprised if that turns out to be a problem.

We should really move this decryption logic ASAP.

crates/matrix-sdk-ui/tests/integration/timeline/decryption.rs
```diff
@@ -0,0 +1,345 @@
+// Copyright 2023 The Matrix.org Foundation C.I.C.
```
poljar, line 1: :older_man: 

```suggestion
// Copyright 2025 The Matrix.org Foundation C.I.C.
```

crates/matrix-sdk-ui/tests/integration/timeline/decryption.rs
```diff
@@ -0,0 +1,345 @@
...
+                                None,
+                            ))
+                            .event_id(event_id!("$ev0"))
+                            .into_utd_sync_timeline_event(),
+                          ],
+                    },
+                ],
+            )
+            .await
+            .unwrap();
```
poljar, line 116: Instead of unwrapping everywhere, it would be really nice to move some of those comments into `expect()` calls. That way if the test fails, we have the description right in the shell that run the test, instead of searching for the source file.

poljar (approved):
Looks good, I left some philosophical musings and nits.

crates/matrix-sdk-ui/src/timeline/builder.rs
Hywan: It is called per batch of updates, so it can cover multiple events.

crates/matrix-sdk-ui/tests/integration/timeline/decryption.rs
Hywan: 👶

crates/matrix-sdk-ui/src/timeline/builder.rs
poljar: Ah ok, all good then.

crates/matrix-sdk-ui/tests/integration/timeline/decryption.rs
Hywan: Yup, done!

PR #4796, opened 2025-03-12
chore(ui): Extract `TimelineBuilder` tasks into their own functions

Hywan:
Extracted from https://github.com/matrix-org/matrix-rust-sdk/pull/4794.

This patch rewrites the `TimelineBuilder::build()` method by extracting all the spawned task code block into their own function. The resulting code is simpler to read, and we can reference each task by its function name. Moreover, each task is isolated, it's a bit easier to reason about.

### DIFF
--- a/crates/matrix-sdk-ui/src/timeline/builder.rs
+++ b/crates/matrix-sdk-ui/src/timeline/builder.rs
@@ -14,6 +14,7 @@
 
 use std::{collections::BTreeSet, sync::Arc};
 
+use futures_core::Stream;
 use futures_util::{pin_mut, StreamExt};
 use matrix_sdk::{
     encryption::backups::BackupState,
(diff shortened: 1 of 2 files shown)


crates/matrix-sdk-ui/src/timeline/builder.rs
```diff
@@ -502,3 +313,229 @@ impl TimelineBuilder {
...
+
+                // The updates might have lagged, but the room event cache might have
+                // events, so retrieve them and add them back again to the timeline,
+                // after clearing it.
+                let (initial_events, _stream) = room_event_cache.subscribe().await;
+
+                timeline_controller
+                    .replace_with_initial_remote_events(
+                        initial_events.into_iter(),
+                        RemoteEventOrigin::Sync,
```
Hywan, line 362: It becomes `::Cache` in https://github.com/matrix-org/matrix-rust-sdk/pull/4795. It should be updated before merge.

stefanceriu (approved):
Lovely! 👨‍🍳 👌

PR #4799, opened 2025-03-14
fix(base): Do not define the `RoomInfoNotableUpdate`'s channel based on number of rooms

Hywan:
When `BaseClient` is created, the rooms aren't loaded yet. Then, calculating the size of the channel for `RoomInfoNotableUpdate` is useless, as it will always be zero, and we will fallback to 500 every time. By the way, 500 is a nice default. This patch uses this value as the only channel's size.


poljar (approved):
Err, alright.

PR #4800, opened 2025-03-14
fix(encryption): Return the uiaa error if we have one in the identity reset loop

poljar:
Seems that we don't handle an error here correctly which ends up creating an infinite loop.

Opening as a draft since we need a test.

### DIFF
--- a/crates/matrix-sdk/src/encryption/mod.rs
+++ b/crates/matrix-sdk/src/encryption/mod.rs
@@ -286,8 +286,13 @@ impl CrossSigningResetHandle {
                 return Ok(());
             }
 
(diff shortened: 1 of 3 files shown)


jmartinesp:
Tested on Android and this fixes the issue AFAICT.

jmartinesp:
> Tested on Android and this fixes the issue AFAICT.

@poljar just a ping so this draft PR is not forgotten (I had, in fact, forgotten about it 😅 ).

poljar:
Ah, thanks for reminding me.

I need to write some tests for this before we can undraft it.

Hywan (approved):
Lovely, thanks!

PR #4804, opened 2025-03-14
feat(sdk): Expose a local server builder

zecakeh:
This exposes the server builder code that was used behind the `sso-login` feature so that it can be reused in other places, like during OAuth 2.0 login. It's called `LocalServerBuilder` and is behind the `local-server` feature.

I added types around the channels to make the API easier to use and harder to misuse (if the shutdown signal is not sent, the server will keep running). They are inspired by tokio's `JoinHandle` and `AbortHandle`.


Hywan (approved):
I think this is a great improvement. Thanks for your work!

PR #4810, opened 2025-03-18
test(common): Ensure inserting a new first chunk is as expected

Hywan:
This patch adds a `debug_assert_eq` to ensure the new first chunk is the one that matches the `lazy_previous` of the current first chunk.

---

### DIFF
--- a/crates/matrix-sdk-common/src/linked_chunk/lazy_loader.rs
+++ b/crates/matrix-sdk-common/src/linked_chunk/lazy_loader.rs
@@ -96,7 +96,8 @@ where
             }
         }
 
-        let expected_next_chunk = linked_chunk.links.first_chunk().identifier();
+        let first_chunk = linked_chunk.links.first_chunk();
+        let expected_next_chunk = first_chunk.identifier();
 
         // New chunk has a next chunk.
         let Some(next_chunk) = new_first_chunk.next else {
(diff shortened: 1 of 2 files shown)


crates/matrix-sdk-common/src/linked_chunk/lazy_loader.rs
```diff
@@ -111,15 +112,25 @@ where
...
     }
 
     // Insert the new first chunk.
     {
         // Transform the `RawChunk` into a `Chunk`.
         let lazy_previous = new_first_chunk.previous.take();
+        let new_first_chunk_identifier = new_first_chunk.identifier;
         let mut new_first_chunk =
-            Chunk::new_leaked(new_first_chunk.identifier, new_first_chunk.content);
+            Chunk::new_leaked(new_first_chunk_identifier, new_first_chunk.content);
```
bnjbvr, line 133: Could we revert these changes, and only keep what we had before? avoids one local variable, which is cognitive overhead

bnjbvr (approved):
I like checking invariants, nice!

crates/matrix-sdk-common/src/linked_chunk/lazy_loader.rs
Hywan: oh, rebasing failure, thanks!

PR #4811, opened 2025-03-18
doc(sdk): Precise in which context a variable can or cannot be used

Hywan:
Yup.

---


crates/matrix-sdk/src/event_cache/room/mod.rs
```diff
@@ -922,6 +922,9 @@ mod private {
 
             // We've reached the start on disk, if and only if, there was no chunk prior to
             // the one we just loaded.
+            //
+            // This value is correct, if and only if, it is used for a chunk content of kind
+            // `Items`.
```
bnjbvr, line 927: We could even say events :)

bnjbvr (approved):

PR #4814, opened 2025-03-19
refactor(oauth): Rename `OidcRegistrations` to `OAuthRegistrationStore` and other improvements

zecakeh:
We rename the types to use the same prefix as the other types in the OAuth 2.0 API, and we use the
same suffix as other data-persisting APIs for consistency. It also avoids to have two modules with very similar names, the only difference being a trailing `s`.

We also refactor the error types to have more precise variants, we don't ignore errors when reading from the file, and we make the methods async because they perform blocking I/O.

### DIFF
--- a/crates/matrix-sdk/src/authentication/oauth/registration_store.rs
+++ b/crates/matrix-sdk/src/authentication/oauth/registration_store.rs
@@ -110,9 +110,12 @@ impl OAuthRegistrationStore {
     /// Returns the client ID registered for a particular issuer or `None` if a
     /// registration hasn't been made.
     pub fn client_id(&self, issuer: &Url) -> Option<ClientId> {
-        let mut data = self.read_or_generate_registration_data();
-        data.dynamic_registrations.extend(self.static_registrations.clone());
-        data.dynamic_registrations.get(issuer).cloned()
+        if let Some(client_id) = self.static_registrations.get(issuer) {
+            return Some(client_id.clone());
+        }
+
+        let mut data = self.read_registration_data()?;
+        data.dynamic_registrations.remove(issuer)
     }
 
     /// Stores a new client ID registration for a particular issuer.
@@ -124,7 +127,13 @@ impl OAuthRegistrationStore {
         client_id: ClientId,
         issuer: Url,
     ) -> Result<(), OAuthRegistrationStoreError> {
-        let mut data = self.read_or_generate_registration_data();
+        let mut data = self.read_registration_data().unwrap_or_else(|| {
+            tracing::info!("Generating new OAuth 2.0 client registration data");
+            FrozenRegistrationData {
+                metadata: self.metadata.clone(),
+                dynamic_registrations: Default::default(),
+            }
+        });
         data.dynamic_registrations.insert(issuer, client_id);
 
         let writer = BufWriter::new(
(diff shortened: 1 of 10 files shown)


zecakeh:
Rebased on main to solve conflicts

crates/matrix-sdk/src/authentication/oauth/registration_store.rs
```diff
@@ -104,15 +105,19 @@ impl OAuthRegistrationStore {
...
+    pub async fn new(
+        file: PathBuf,
         metadata: Raw<ClientMetadata>,
         static_registrations: HashMap<Url, ClientId>,
     ) -> Result<Self, OAuthRegistrationStoreError> {
-        let parent = file.parent().ok_or(OAuthRegistrationStoreError::NotAFilePath)?;
-        fs::create_dir_all(parent)?;
-
-        Ok(OAuthRegistrationStore { file_path: file.to_owned(), metadata, static_registrations })
+        spawn_blocking(move || {
```
Hywan, line 110: I think `spawn_blocking` is _acceptable_ here as it's probably not a long task. However, using [`tokio::fs::create_dir_all`](https://docs.rs/tokio/latest/tokio/fs/fn.create_dir_all.html) would be preferable here.

Looking at the code of [`create_dir_all`](https://github.com/tokio-rs/tokio/blob/0ec4d0db4d24af74cea40cf0ceed6b7a0a5ff183/tokio/src/fs/create_dir_all.rs#L48-L51), it calls [`asyncify`](https://github.com/tokio-rs/tokio/blob/0ec4d0db4d24af74cea40cf0ceed6b7a0a5ff183/tokio/src/fs/mod.rs#L305-L317), which basically call `spawn_blocking`. So it's technically the same, but can you change to `tokio::fs::create_dir_all` so that the cognitive “complexity” of `spawn_blocking` is hidden. Also, the day Tokio changes for another API, we will get it for free. Also, it removes the need for one more blocking thread, which aren't cheap.

crates/matrix-sdk/src/authentication/oauth/registration_store.rs
```diff
@@ -164,8 +172,14 @@ impl OAuthRegistrationStore {
...
         data.dynamic_registrations.insert(issuer, client_id);
 
-        let writer = BufWriter::new(File::create(&self.file_path)?);
-        serde_json::to_writer(writer, &data).map_err(OAuthRegistrationStoreError::IntoJson)
+        let file_path = self.file_path.clone();
+
+        spawn_blocking(move || {
+            let writer = BufWriter::new(File::create(&file_path)?);
+            serde_json::to_writer(writer, &data).map_err(OAuthRegistrationStoreError::IntoJson)
+        })
```
Hywan, line 194: Too bad `serde_json` doesn't support `AsyncWrite` here.

crates/matrix-sdk/src/authentication/oauth/registration_store.rs
```diff
@@ -174,44 +188,56 @@ impl OAuthRegistrationStore {
...
+                    return Err(error.into());
                 }
+            };
 
-                // Forward the error.
-                return Err(error.into());
-            }
-        };
+            let registration_data: FrozenRegistrationData =
+                serde_json::from_reader(BufReader::new(file))
```
Hywan, line 225: Same, too bad it doesn't support `AsyncRead` neither.

crates/matrix-sdk/src/authentication/oauth/registration_store.rs
```diff
@@ -102,24 +103,35 @@ impl OAuthRegistrationStore {
     /// * `metadata` - The metadata used to register the client. If this changes
     ///   compared to the value stored in the file, any stored registrations
     ///   will be invalidated so the client can re-register with the new data.
-    ///
-    /// * `static_registrations` - Pre-configured registrations for use with
-    ///   servers that don't support dynamic client registration.
     pub async fn new(
         file: PathBuf,
         metadata: Raw<ClientMetadata>,
-        static_registrations: HashMap<Url, ClientId>,
```
Hywan: Good move.

Hywan (changes requested):
Nice changes, thanks for taking the time to simplify the API and to make some of them async! I enjoy the fact we have more errors and no more `UnknownError`.
I've a tiny feedback regarding a `spawn_blocking`, but overall it looks really great!

crates/matrix-sdk/src/authentication/oauth/registration_store.rs
zecakeh: We could also just serialize the whole file and then write it, if you'd prefer to use tokio APIs.

Hywan: Hmm that could be a good idea, yes. If it doesn't annoy you to change this part, it would be nice!

zecakeh: I also prefer that, I just didn't want to change the code too much if it was some kind of necessary optimization to not have the whole file in memory.

Hywan: Let's change that as another commit then, no need to rebase or change your existing patches.

zecakeh: Done

crates/matrix-sdk/src/authentication/oauth/registration_store.rs
zecakeh: Done

Hywan (approved):
Thanks for removing all the `spawn_blocking`s!

PR #4816, opened 2025-03-19
feat(timeline): add a virtual start of timeline item

bnjbvr:
Might be better than letting the apps figure out whether they need to put one or not.

The next step would be to add a virtuel item for gaps as well, `Gap { loading: bool }`:

1. this would pave the way for inserting gaps at random positions in the timeline,
2. when the timeline is back-paginating, we'd have a gap with `loading: true`
3. when the timeline is idle and we suspect there *might* be an history we haven't loaded, we'd have a gap with `loading: false`. The next call to `Timeline::paginate_backwards()` would resolve it into a `TimelineStart` or another `Gap { loading: false }`.

Part of #4821.

### DIFF
--- a/crates/matrix-sdk-ui/src/timeline/controller/mod.rs
+++ b/crates/matrix-sdk-ui/src/timeline/controller/mod.rs
@@ -71,7 +71,7 @@ use super::{
     traits::{Decryptor, RoomDataProvider},
     DateDividerMode, Error, EventSendState, EventTimelineItem, InReplyToDetails, Message,
     PaginationError, Profile, RepliedToEvent, TimelineDetails, TimelineEventItemId, TimelineFocus,
-    TimelineItem, TimelineItemContent, TimelineItemKind,
+    TimelineItem, TimelineItemContent, TimelineItemKind, VirtualTimelineItem,
 };
 use crate::{
     timeline::{
@@ -1308,6 +1308,18 @@ impl<P: RoomDataProvider, D: Decryptor> TimelineController<P, D> {
                 info!(txn_id = %related_to, "some media for a media event has been uploaded");
             }
         }
+    }
+
+    /// Insert a timeline start item at the beginning of the room, if it's
+    /// missing.
+    pub async fn insert_timeline_start_if_missing(&self) {
+        let mut state = self.state.write().await;
+        let mut txn = state.transaction();
+        if txn.items.get(0).is_some_and(|item| item.is_timeline_start()) {
+            return;
+        }
+        txn.items.push_front(txn.meta.new_timeline_item(VirtualTimelineItem::TimelineStart), None);
+        txn.commit();
     }
 }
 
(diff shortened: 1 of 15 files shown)


crates/matrix-sdk-ui/src/timeline/controller/mod.rs
```diff
@@ -1309,6 +1309,20 @@ impl<P: RoomDataProvider, D: Decryptor> TimelineController<P, D> {
...
         }
     }
+
+    /// Insert a timeline start item at the beginning of the room, if it's
+    /// missing.
+    pub async fn insert_timeline_start_if_missing(&self) {
+        let mut state = self.state.write().await;
+        let mut txn = state.transaction();
+        for item in txn.items.iter() {
+            if item.is_timeline_start() {
```
Hywan, line 1319: Can we have a `VirtualTimelineItem::TimelineStart` at another position than the first position? I don' think so.
Consequently, this loop is unecessary (and even dangerous as we will loop over all items if there is no `TimelineStart`):

```rust
if self.items.get(0).is_some_and(|item| item.is_timeline_start()) {
    return;
}

txn.items.push_front(…);
txn.commit();
```

Hywan (changes requested):
Nice! I like having this marker. One note about a loop that can be simplified. Otherwise everything else is good.

One note while reading your code (it's **NOT** something I'm asking you to do, I can do it, just wanted to share my idea). I wonder if `ObservableItems` shouldn't get a method to push or insert specific timeline items, so that we move the logic of where to insert or shift index in there instead of having it inside the `EventHandler`:

* `insert_timeline_start`: will emit a `PushFront` with a `TimelineStart`
* `insert_remote_timeline_item(index, item)`: depending of `index` and the presence of some items, it will emit a `PushFront`, `Insert` or `PushBack`
* `insert_local_timeline_item`: always push at the end etc.

The idea is to get a cleaner API and a central place where to have all the invariants regarding the position of the items. Thoughts?

crates/matrix-sdk-ui/src/timeline/controller/mod.rs
bnjbvr: I was worried that the date separator or the read marker could go before, but maybe you're right and I could add a test that a date separator can't be first.

Hywan: You're checking the first item in the EventHandler. Be sure to be consistent.

bnjbvr:
> One note while reading your code (it's **NOT** something I'm asking you to do, I can do it, just wanted to share my idea). I wonder if `ObservableItems` shouldn't get a method to push or insert specific timeline items, so that we move the logic of where to insert or shift index in there instead of having it inside the `EventHandler`:
> 
>     * `insert_timeline_start`: will emit a `PushFront` with a `TimelineStart`
> 
>     * `insert_remote_timeline_item(index, item)`: depending of `index` and the presence of some items, it will emit a `PushFront`, `Insert` or `PushBack`
> 
>     * `insert_local_timeline_item`: always push at the end etc.
> 
> 
> The idea is to get a cleaner API and a central place where to have all the invariants regarding the position of the items. Thoughts?

I think that makes sense overall, although I'm not quite sure what it buys us. More concise names would be sweet, like `prepend_item` / `insert_item` / `push_item`.

Overall, I think the most valuable part would be to make it possible in the tests to assert against somewhat different expectations, using one of the macros that asserts what's the next timeline update is: e.g. asserting `PushBack` against a `Insert(index: vector.length(), value)` would be accepted, and conversely (ditto for Insert(0) vs PushFront). This would help testing the timeline in a much more robust manner.

Hywan (approved):
Good to me. Thanks for the changes!

PR #4817, opened 2025-03-19
refactor(oauth): Simplify behavior to finish login

zecakeh:
This can be reviewed commit by commit:

1. Merge `finish_authorization()` and `finish_login()`: now the user only needs to call `finish_login()`, since currently the only way to use the authorization code flow is for logging in it makes more sense.
2. Allow to log into the same session several times: useful for soft logouts, we just check that we are indeed logged into the same session.
3. Get rid of `OAuthError::MissingDeviceId`: since we are the ones providing the device ID to the server, there is clearly a way to work around the ID missing in the response.

### DIFF
--- a/crates/matrix-sdk/src/authentication/oauth/error.rs
+++ b/crates/matrix-sdk/src/authentication/oauth/error.rs
@@ -65,10 +65,6 @@ pub enum OAuthError {
     #[error("client not registered")]
     NotRegistered,
 
-    /// The device ID was not returned by the homeserver after login.
-    #[error("missing device ID in response")]
-    MissingDeviceId,
-
     /// The client is not authenticated while the request requires it.
     #[error("client not authenticated")]
     NotAuthenticated,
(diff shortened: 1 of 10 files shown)


crates/matrix-sdk/src/authentication/oauth/error.rs
```diff
@@ -94,6 +94,14 @@ pub enum OAuthError {
     #[error(transparent)]
     LockError(#[from] CrossProcessRefreshLockError),
 
+    /// The user logged into a session that is different than the one the client
+    /// is already using.
+    ///
+    /// This only happens if the session was already restored, and the user logs
+    /// into a new session that its different than the old one.
```
andybalaam, line 97: "its" -> "is"

andybalaam (approved):
Looks good! One typo in a code comment, and one in the last commit: "changelof" -> "changelog". Feel free to fix and rebase, and I'll merge.

crates/matrix-sdk/src/authentication/oauth/error.rs
zecakeh: Fixed

zecakeh:
Typo in the commit fixed two, thanks!

PR #4818, opened 2025-03-19
fix(ffi): propagate the back-pagination status immediately after a subscriber asked for it

bnjbvr:
And add more logs to the event cache computation of the `reached_start` value.

Might be a reason why we ran into https://github.com/element-hq/element-x-ios/issues/3768: spawning the task can be deferred a bit, so EXI doesn't receive the initial value until much later, causing it to be confused as its initial value was `timelineEndReached` (it's been changed [here](https://github.com/element-hq/element-x-ios/pull/3924), which should also be sufficient to fix the issue).

Part of #3280.


andybalaam (approved):
Looks good, thanks!

PR #4819, opened 2025-03-19
feat(timeline): add functions for sending messages in threads

Johennes:
`send_reply` is geared towards clients that don't render threads themselves. It sends a reply and optionally forwards it onto any existing thread.

This PR adds `send_thread_reply` for clients that do render threads themselves. This sends a message onto a thread, regardless of whether a thread existed before, and includes the fallback for non-threaded clients only if the message is not itself a reply on the thread.

### DIFF
--- a/bindings/matrix-sdk-ffi/src/timeline/mod.rs
+++ b/bindings/matrix-sdk-ffi/src/timeline/mod.rs
@@ -48,7 +48,7 @@ use ruma::{
         },
         receipt::ReceiptThread,
         room::message::{
-            ForwardThread, LocationMessageEventContent, MessageType,
+            ForwardThread, LocationMessageEventContent, MessageType, ReplyWithinThread,
             RoomMessageEventContentWithoutRelation,
         },
         AnyMessageLikeEventContent,
@@ -461,6 +461,10 @@ impl Timeline {
         Ok(())
     }
 
+    /// Send a reply.
+    ///
+    /// If the replied to event has a thread relation, it is forwarded on the reply
+    /// so that clients that support threads can render the reply inside the thread.
     pub async fn send_reply(
         &self,
         msg: Arc<RoomMessageEventContentWithoutRelation>,
@@ -475,6 +479,29 @@ impl Timeline {
 
         self.inner
             .send_reply((*msg).clone(), replied_to_info, ForwardThread::Yes)
+            .await
+            .map_err(|err| anyhow::anyhow!(err))?;
+        Ok(())
+    }
+
+    /// Send a message on a thread.
+    ///
+    /// If the replied to event does not have a thread relation, it becomes the root
+    /// of a new thread.
+    pub async fn send_thread_reply(
+        &self,
+        msg: Arc<RoomMessageEventContentWithoutRelation>,
+        event_id: String,
+    ) -> Result<(), ClientError> {
+        let event_id = EventId::parse(event_id)?;
+        let replied_to_info = self
+            .inner
+            .replied_to_info_from_event_id(&event_id)
+            .await
+            .map_err(|err| anyhow::anyhow!(err))?;
+
+        self.inner
+            .send_thread_reply((*msg).clone(), replied_to_info, ReplyWithinThread::No)
             .await
             .map_err(|err| anyhow::anyhow!(err))?;
         Ok(())
--- a/bindings/matrix-sdk-ffi/src/timeline/mod.rs
+++ b/bindings/matrix-sdk-ffi/src/timeline/mod.rs
@@ -479,7 +479,11 @@ impl Timeline {
             .map_err(|err| anyhow::anyhow!(err))?;
 
         self.inner
-            .send_reply((*msg).clone(), replied_to_info, ForwardThread::Yes)
+            .send_reply(
+                (*msg).clone(),
+                replied_to_info,
+                timeline::EnforceThread::No(ForwardThread::Yes),
+            )
             .await
             .map_err(|err| anyhow::anyhow!(err))?;
         Ok(())
@@ -502,7 +506,11 @@ impl Timeline {
             .map_err(|err| anyhow::anyhow!(err))?;
 
         self.inner
-            .send_thread_reply((*msg).clone(), replied_to_info, ReplyWithinThread::No)
+            .send_reply(
+                (*msg).clone(),
+                replied_to_info,
+                timeline::EnforceThread::Yes(ReplyWithinThread::No),
+            )
             .await
             .map_err(|err| anyhow::anyhow!(err))?;
         Ok(())
--- a/bindings/matrix-sdk-ffi/src/timeline/mod.rs
+++ b/bindings/matrix-sdk-ffi/src/timeline/mod.rs
@@ -493,10 +493,19 @@ impl Timeline {
     ///
     /// If the replied to event does not have a thread relation, it becomes the
     /// root of a new thread.
+    ///
+    /// # Arguments
+    ///
+    /// * `msg` - Message content to send
+    ///
+    /// * `event_id` - ID of the event to reply to
+    ///
+    /// * `is_reply` - Whether the message is a reply on a thread
     pub async fn send_thread_reply(
         &self,
         msg: Arc<RoomMessageEventContentWithoutRelation>,
         event_id: String,
+        is_reply: bool,
     ) -> Result<(), ClientError> {
         let event_id = EventId::parse(event_id)?;
         let replied_to_info = self
@@ -509,7 +518,11 @@ impl Timeline {
             .send_reply(
                 (*msg).clone(),
                 replied_to_info,
-                timeline::EnforceThread::Yes(ReplyWithinThread::No),
+                timeline::EnforceThread::Yes(if is_reply {
+                    ReplyWithinThread::Yes
+                } else {
+                    ReplyWithinThread::No
+                }),
             )
             .await
             .map_err(|err| anyhow::anyhow!(err))?;
--- a/bindings/matrix-sdk-ffi/src/timeline/mod.rs
+++ b/bindings/matrix-sdk-ffi/src/timeline/mod.rs
@@ -463,8 +463,9 @@ impl Timeline {
 
     /// Send a reply.
     ///
-    /// If the replied to event has a thread relation, it is forwarded on the reply
-    /// so that clients that support threads can render the reply inside the thread.
+    /// If the replied to event has a thread relation, it is forwarded on the
+    /// reply so that clients that support threads can render the reply
+    /// inside the thread.
     pub async fn send_reply(
         &self,
         msg: Arc<RoomMessageEventContentWithoutRelation>,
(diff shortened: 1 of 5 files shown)


crates/matrix-sdk-ui/src/timeline/mod.rs
```diff
@@ -369,6 +371,107 @@ impl Timeline {
...
+                let thread_root =
+                    if let Some(ruma::events::room::encrypted::Relation::Thread(Thread {
+                        event_id,
+                        ..
+                    })) = &relates_to
+                    {
+                        event_id.to_owned()
+                    } else {
+                        event_id.to_owned()
+                    };
```
bnjbvr, line 458: ogod this is the most evil use of shadowing i've seen in a while :D let's get rid of the top-level `event_id` local variable, and use `replied_to_info.event_id` in the else arm :)

crates/matrix-sdk-ui/src/timeline/mod.rs
```diff
@@ -369,6 +371,107 @@ impl Timeline {
...
+    /// * `replied_to_info` - A wrapper that contains the event ID, sender,
+    ///   content and timestamp of the event to reply to
+    ///
+    /// * `is_reply` - Whether or not the message is a reply inside a thread
+    #[instrument(skip(self, content, replied_to_info))]
+    pub async fn send_thread_reply(
+        &self,
+        content: RoomMessageEventContentWithoutRelation,
+        replied_to_info: RepliedToInfo,
+        is_reply: ReplyWithinThread,
```
bnjbvr, line 393: Instead of duplicating the method, could `send_reply` include a variant that determines whether this is to be used in a thread or not, and for the thread variant have the `ReplyWithinThread` field? It seems it could avoid lots of code duplication here, even though it might make the `Raw` case a bit gnarlier.

If the code becomes too complicated, we can keep separate methods, but just curious if it does make things too complicated or not.

crates/matrix-sdk-ui/src/timeline/mod.rs
```diff
@@ -369,6 +371,108 @@ impl Timeline {
...
+                // the next ruma release.
+                #[derive(Deserialize)]
+                struct ContentDeHelper {
+                    #[serde(rename = "m.relates_to")]
+                    relates_to: Option<ruma::events::room::encrypted::Relation>,
+                }
+
+                let previous_content =
+                    raw_event.get_field::<ContentDeHelper>("content").ok().flatten();
+                let relates_to = previous_content.as_ref().and_then(|c| c.relates_to.as_ref());
```
bnjbvr, line 435: Can this be inlined in the one place where it's used?

crates/matrix-sdk-ui/tests/integration/timeline/replies.rs
```diff
@@ -906,6 +908,106 @@ async fn test_send_reply_with_event_id() {
...
+        assert_next_matches!(timeline_stream, VectorDiff::PushBack { value } => value);
+
+    // Clear the timeline to make sure the old item does not need to be
+    // available in it for the reply to work.
+    timeline.clear().await;
+    assert_next_matches!(timeline_stream, VectorDiff::Clear);
+
+    // Now, let's reply to a message sent by `BOB`.
+    Mock::given(method("PUT"))
+        .and(path_regex(r"^/_matrix/client/v3/rooms/.*/send/.*"))
```
bnjbvr, line 946: you can use `server.mock_room_send()` here

crates/matrix-sdk-ui/tests/integration/timeline/replies.rs
```diff
@@ -906,6 +908,106 @@ async fn test_send_reply_with_event_id() {
...
+    assert_eq!(in_reply_to.event_id, event_id_from_bob);
+
+    // Right now, we don't pass along the replied-to event to the event handler,
+    // so it's not available if the timeline got cleared. Not critical, but
+    // there's notable room for improvement here.
+    //
+    // let replied_to_event =
+    // assert_matches!(&in_reply_to.event, TimelineDetails::Ready(ev) => ev);
+    // assert_eq!(replied_to_event.sender(), *BOB);
+
```
bnjbvr, line 994: Let's remove this, as it's copy pasta from the test that was the inspiration here :)

crates/matrix-sdk-ui/tests/integration/timeline/replies.rs
```diff
@@ -906,6 +908,106 @@ async fn test_send_reply_with_event_id() {
...
+    let reply_message = reply_item_remote_echo.content().as_message().unwrap();
+    assert_eq!(reply_message.body(), "Replying to Bob");
+    let in_reply_to = reply_message.in_reply_to().unwrap();
+    assert_eq!(in_reply_to.event_id, event_id_from_bob);
+
+    // Same as above.
+    //
+    // let replied_to_event =
+    // assert_matches!(&in_reply_to.event, TimelineDetails::Ready(ev) =>
+    // ev); assert_eq!(replied_to_event.sender(), *BOB);
```
bnjbvr, line 1008: …same as above :D

crates/matrix-sdk-ui/tests/integration/timeline/replies.rs
```diff
@@ -906,6 +908,106 @@ async fn test_send_reply_with_event_id() {
...
+    // Let the send queue handle the event.
+    yield_now().await;
+
+    let reply_item = assert_next_matches!(timeline_stream, VectorDiff::PushBack { value } => value);
+
+    assert_matches!(reply_item.send_state(), Some(EventSendState::NotSentYet));
+    let reply_message = reply_item.content().as_message().unwrap();
+    assert_eq!(reply_message.body(), "Replying to Bob");
+    let in_reply_to = reply_message.in_reply_to().unwrap();
+    assert_eq!(in_reply_to.event_id, event_id_from_bob);
```
bnjbvr, line 985: Can you also assert the thread information is correctly set?

crates/matrix-sdk-ui/tests/integration/timeline/replies.rs
```diff
@@ -906,6 +908,106 @@ async fn test_send_reply_with_event_id() {
...
+    // assert_eq!(replied_to_event.sender(), *BOB);
+
+    let diff = timeout(timeline_stream.next(), Duration::from_secs(1)).await.unwrap().unwrap();
+    assert_let!(VectorDiff::Set { index: 0, value: reply_item_remote_echo } = diff);
+
+    assert_matches!(reply_item_remote_echo.send_state(), Some(EventSendState::Sent { .. }));
+    let reply_message = reply_item_remote_echo.content().as_message().unwrap();
+    assert_eq!(reply_message.body(), "Replying to Bob");
+    let in_reply_to = reply_message.in_reply_to().unwrap();
+    assert_eq!(in_reply_to.event_id, event_id_from_bob);
```
bnjbvr, line 1002: And here too, let's check the thread relation is correctly set too.

bnjbvr (changes requested):
Nice, this should provide the ability to start a new thread.

crates/matrix-sdk-ui/src/timeline/mod.rs
Johennes: Guilty. 😅

jplatte: Also destructuring the `Thread` inside the `if let` seems harder to read than not doing it.

jplatte: No, I mean instead of

```rust
if let Some(encrypted::Relation::Thread(Thread { event_id, .. })) = &relates_to {
    event_id.to_owned()
}
```

you could write

```rust
if let Some(encrypted::Relation::Thread(thread)) = &relates_to {
    thread.event_id.to_owned()
}
```

which would be easier to read.

Johennes: Oh. Yes, that's better indeed. 👍

crates/matrix-sdk-ui/src/timeline/mod.rs
Johennes: Great idea! I have tried to implement this with a new `EnforceThread` enum. I think complexity-wise it's ok and most of the lengthy raw case will go away with the next ruma release.

bnjbvr (changes requested):
Super nice, thanks for reworking it, the changes are nicely minimal now :sparkles:

crates/matrix-sdk-ui/src/timeline/mod.rs
```diff
@@ -346,22 +357,78 @@ impl Timeline {
...
+                match enforce_thread {
+                    EnforceThread::Yes(is_reply) => {
+                        // Some of the code below technically belongs into ruma. However,
+                        // reply fallbacks have been removed in Matrix 1.13 which means
+                        // both match arms can use the successor of make_for_thread in
+                        // the next ruma release.
+                        #[derive(Deserialize)]
+                        struct ContentDeHelper {
+                            #[serde(rename = "m.relates_to")]
+                            relates_to: Option<ruma::events::room::encrypted::Relation>,
```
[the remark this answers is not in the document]
Johennes: Hm, at least it is what `make_reply_to_raw` also uses. Comparing the two structs, the one from `encrypted` seems broader as it also supports annotations which the one from `message` doesn't. Given that we only check for thread relations, I suppose it doesn't make a huge difference though.

bindings/matrix-sdk-ffi/src/timeline/mod.rs
```diff
@@ -474,7 +479,38 @@ impl Timeline {
...
+            .inner
+            .replied_to_info_from_event_id(&event_id)
+            .await
+            .map_err(|err| anyhow::anyhow!(err))?;
+
+        self.inner
+            .send_reply(
+                (*msg).clone(),
+                replied_to_info,
+                timeline::EnforceThread::Yes(ReplyWithinThread::No),
```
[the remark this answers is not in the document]
Johennes: Good idea 👍

crates/matrix-sdk-ui/src/timeline/mod.rs
```diff
@@ -135,6 +137,19 @@ pub enum ReplyContent {
...
+#[allow(clippy::exhaustive_enums)]
+pub enum EnforceThread {
+    /// A thread relation is enforced. If the original message does not have a
+    /// thread relation itself, a new thread is started.
+    Yes(ReplyWithinThread),
+
+    /// A thread relation is not enforced. If the original message has a thread
+    /// relation, it is optionally forwarded.
+    No(ForwardThread),
+}
```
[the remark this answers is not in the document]
Johennes: Yes, this feels better than the nested enums. Have changed it around now.

crates/matrix-sdk-ui/src/timeline/mod.rs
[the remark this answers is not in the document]
Johennes: I have shortened the statements with `use` now.

bnjbvr (approved):
Awesome, thank you!

PR #4820, opened 2025-03-20
fix(ffi): propagate initial values before the future is picked by a runtime

bnjbvr:
Same spirit as #4818; doing work after the `async move` "moves" the code inside the future, and it will be executed only when a tokio runtime is available, instead of doing it right away. We can improve and do it right away in a few cases:

- when getting the initial timeline items
- when looking at the room directory search results

I've audited all the `spawn`s at the FFI layer, so we should be in a good shape now.

### DIFF
--- a/bindings/matrix-sdk-ffi/src/room.rs
+++ b/bindings/matrix-sdk-ffi/src/room.rs
@@ -607,29 +607,28 @@ impl Room {
         })))
     }
 
-    pub fn subscribe_to_identity_status_changes(
+    pub async fn subscribe_to_identity_status_changes(
         &self,
         listener: Box<dyn IdentityStatusChangeListener>,
-    ) -> Arc<TaskHandle> {
+    ) -> Result<Arc<TaskHandle>, ClientError> {
         let room = self.inner.clone();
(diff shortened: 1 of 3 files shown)


Hywan (approved):
Yup, makes sense! Thanks for the fixes.

Hywan:
I honestly don't understand the error on Complement Crypto.

https://github.com/matrix-org/complement-crypto/blob/23821cf6320beb15b20233dc5728a019fcc02f15/internal/api/rust/rust.go#L720-L721

The listener should be called the same way. If we are hitting a timeout, it means the listener isn't called, which is surprising as it should be called… faster.

bnjbvr:
Kegan reported that the complement-crypto test expects to get the first events *after* AddTimelineListener has returned, not while it happens, so it's likely deadlocking somewhere. @kegsay would you kindly help us and rework the Go code so it works with this breaking change, please? Otherwise we can also disable the complement-crypto testing in the meanwhile.

Hywan:
Re-run the complement crypto job now there is https://github.com/matrix-org/complement-crypto/pull/169.

Issue #4793, opened 2025-03-12
Element Call widget generation to support latest including embedded package


hughns:
This is to support https://github.com/element-hq/element-call/issues/2994

Requirements:
 
- MUST Don't add `/room` path suffix
- MUST Add support for optional `posthogApiHost` and `posthogApiKey` and `posthogUserId` URL parameters
- MUST Add support for optional `rageshakeSubmitUrl` URL parameter
- MUST Add support for optional `sentryDsn` and `sentryEnvironment` URL parameters
- SHOULD Deprecate `analyticsID`
- SHOULD Add support for optional `hideScreensharing`
- SHOULD
  - Add support for optional `intent`: `start_call` or `join_existing`
  - `skipLobby` is now deprecated and `intent` should be used instead

The documentation for the new URL parameters is currently drafted at https://github.com/element-hq/element-call/pull/3089.

hughns:
The merged PR breaks the UX against Element Call 0.7.x where the `skipLobby` flag was being relied upon.

BillCarsonFr:
> The merged [PR](https://github.com/matrix-org/matrix-rust-sdk/pull/4802) breaks the UX against Element Call 0.7.x where the `skipLobby` flag was being relied upon.

Any details on what breaks? is there an issue? some logs? how to reproduce

PR #4802, opened 2025-03-14
feat(widgets, element-call) Update the widget url generation

toger5:
Fixes: https://github.com/matrix-org/matrix-rust-sdk/issues/4793
- [x] Public API changes documented in changelogs (optional)

Signed-off-by:

PR #4822, opened 2025-03-20
feat(widgets, element-call) Update the widget url generation

hughns:
Fixes: #4793

The was a previous PR https://github.com/matrix-org/matrix-rust-sdk/pull/4802 which attempted to implement this, but missed some backwards compatibility needs.

This updated PR has the original commit and then additional commits to add the compatibility (along with tests for the new intent param generally).

Signed-off-by:

### DIFF
--- a/crates/matrix-sdk/src/widget/settings/element_call.rs
+++ b/crates/matrix-sdk/src/widget/settings/element_call.rs
@@ -41,15 +41,15 @@ struct ElementCallParams {
     base_url: String,
     // Non template parameters
     parent_url: String,
-    /// Deprecated since Element Call v0.8.0. Included for backwards compatibility.
-    /// Set to `true` if intent is `Intent::StartCall`.
+    /// Deprecated since Element Call v0.8.0. Included for backwards
+    /// compatibility. Set to `true` if intent is `Intent::StartCall`.
     skip_lobby: Option<bool>,
     confine_to_room: bool,
     app_prompt: bool,
     hide_header: bool,
     preload: bool,
-    /// Deprecated since Element Call v0.9.0. Included for backwards compatibility.
-    /// Set to the same as `posthog_user_id`.
+    /// Deprecated since Element Call v0.9.0. Included for backwards
+    /// compatibility. Set to the same as `posthog_user_id`.
     analytics_id: Option<String>,
     /// Supported since Element Call v0.9.0.
     posthog_user_id: Option<String>,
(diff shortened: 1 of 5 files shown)


crates/matrix-sdk/src/widget/settings/element_call.rs
```diff
@@ -220,6 +245,7 @@ impl WidgetSettings {
                 _ => None,
             },
             intent: props.intent,
+            skip_lobby: skip_lobby,
```
bnjbvr: nit, use shorthand destructuring: `skip_lobby,`

crates/matrix-sdk/src/widget/settings/element_call.rs
```diff
@@ -245,7 +244,7 @@ impl WidgetSettings {
                 _ => None,
             },
             intent: props.intent,
-            skip_lobby: skip_lobby,
+            skip_lobby,
```
bnjbvr, line 247: hah, i'm such a clippy

bnjbvr (approved):
lgtm, thanks for the new doc comments!

crates/matrix-sdk/src/widget/settings/element_call.rs
```diff
@@ -157,7 +212,13 @@ impl WidgetSettings {
...
         props: VirtualElementCallWidgetOptions,
     ) -> Result<Self, url::ParseError> {
-        let mut raw_url: Url = Url::parse(&format!("{}/room", props.element_call_url))?;
+        let mut raw_url: Url = Url::parse(&props.element_call_url)?;
+
+        let skip_lobby = if props.intent.as_ref().is_some_and(|x| x == &Intent::StartCall) {
+            Some(true)
+        } else {
+            None
+        };
```
bnjbvr, line 221:
```suggestion
        let skip_lobby = props.intent.as_ref().map(|intent| *intent == Intent::StartCall);
```

bnjbvr (approved):

hughns:
> Reverted my commit: it would slightly change semantics by setting `skipLobby` to false when the intent is set, which is a waste of bandwidth, since I assume EC considers `skipLobby=false` as an equivalent to the absence of `skipLobby`.

Key thing is that we can't just set `skipLobby=false` as this will override the `intent` in Element Call. Instead `skipLobby` should not be sent. So, the original behaviour (which is what was merged) is intended (hence the tests being written that way). 👍

PR #4824, opened 2025-03-20
feat(ffi): trace log bundles + log the log levels

bnjbvr:
This is a new convenient way to enable the TRACE log levels for multiple targets at the same time, when they relate to the same feature. For instance, there's a log pack for the event cache, that will enable the trace level for `matrix_sdk::event_cache`, `matrix_sdk_base::event_cache` and `matrix_sdk_sqlite::event_cache_store` at the same time.

I haven't made it possible to configure the log *level* (instead of TRACE) for those targets, because I think we won't need it; the only interesting case is debugging purposes, which usually requires trace logs.

Also, log the env filter string, so we know when looking at rageshakes what logs are enabled.

### DIFF
--- a/bindings/matrix-sdk-ffi/src/platform.rs
+++ b/bindings/matrix-sdk-ffi/src/platform.rs
@@ -410,10 +410,15 @@ fn build_tracing_filter(config: &TracingConfiguration) -> String {
 pub fn init_platform(config: TracingConfiguration, use_lightweight_tokio_runtime: bool) {
     log_panics();
 
+    let env_filter = build_tracing_filter(&config);
+
     tracing_subscriber::registry()
-        .with(EnvFilter::new(build_tracing_filter(&config)))
+        .with(EnvFilter::new(&env_filter))
         .with(text_layers(config))
         .init();
+
+    // Log the log levels 🧠.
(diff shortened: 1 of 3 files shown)


bindings/matrix-sdk-ffi/src/platform.rs
```diff
@@ -358,11 +412,16 @@ fn build_tracing_filter(config: &TracingConfiguration) -> String {
...
 
+    let env_filter = build_tracing_filter(&config);
+
     tracing_subscriber::registry()
-        .with(EnvFilter::new(build_tracing_filter(&config)))
+        .with(EnvFilter::new(&env_filter))
         .with(text_layers(config))
         .init();
 
+    // Log the log levels 🧠.
```
stefanceriu, line 422: 😂

stefanceriu (approved):
Looks awesome, nicely done! 👏

Hywan:
`timeline::tests::encryption::test_false_positive_late_decryption_regression` seems flaky. It has been introduced in https://github.com/matrix-org/matrix-rust-sdk/pull/4252, 4 months ago. It's the first time I see it being flaky though. cc @BillCarsonFr

Issue #4801, opened 2025-03-14
Specialize the NotificationClient more


poljar:
The [NotificationClient](https://github.com/matrix-org/matrix-rust-sdk/blob/c8536e9e46cc99b946db50f20838a0de928ee143/crates/matrix-sdk-ui/src/notification_client.rs#L75-L80) is a lightweight, specialized version of our Client object.

It relies on an in-memory store and requires room data for incoming notifications to be provided via sliding sync while it fetches the event which triggered the notification.

However, creating a `NotificationClient`currently requires the creation of a full `Client` instance, which loads all `RoomInfo` objects from storage. For users with thousands of rooms, this can quickly consume excessive memory, especially within Apple's heavily restricted notification processes.

To reduce memory usage, we should further optimize the `NotificationClient` by:

1. Avoiding unnecessary RoomInfo loading – Ideally, no room data should be loaded from storage, or at most, only the room receiving the notification.
2. Limiting outgoing E2EE-related requests, we don't need to upload one-time keys in the `NotificationClient`.
3. Optimizing batch sizes for E2EE requests – Some E2EE operations exchange large amounts of data. If batch sizes are configurable, they should be reduced where necessary. For instance, the outgoing /keys/query request currently uses a [batch size of 250](https://github.com/matrix-org/matrix-rust-sdk/blob/c8536e9e46cc99b946db50f20838a0de928ee143/crates/matrix-sdk-crypto/src/identities/manager.rs#L98).
4. Tweak the cache sizes of the SQLite database – The Notification process still uses a SQLite database for the crypto store, which has some caches that might be lowered. For example [`PRAGMA cache_size`](https://www.sqlite.org/pragma.html#pragma_cache_size) uses 2MB by default
5. Tweak the pool size of the SQLite connection manager – The database for the crypto store uses [deadpool](https://docs.rs/deadpool/0.12.2/deadpool/) to handle a pool of SQLite connections. By default a pool size of `num_cpu * 4` will be used. The NSE could lower the number of connections in the pool with the [`max_size()`](https://docs.rs/deadpool/latest/deadpool/managed/struct.PoolBuilder.html#method.max_size) method when creating the [pool](https://github.com/matrix-org/matrix-rust-sdk/blob/7f03093e84da16244b8063b9b42f5137ad23a9d8/crates/matrix-sdk-sqlite/src/crypto_store.rs#L85).

ara4n:
[...]

PR #4826, opened 2025-03-21
feat(sqlite): Add `StoreOpenConfig` and `open_with_config` for all stores

Hywan:
This patch adds a new `StoreOpenConfig` type to configure the store when opening it and when creating the pool of connections to SQLite via `deadpool_sqlite`.

This patch also adds a new `open_with_config` constructor on all stores, namely `SqliteCryptoStore`, `SqliteEventCacheStore` and `SqliteStateStore`.

---

* Address https://github.com/matrix-org/matrix-rust-sdk/issues/4801

### DIFF
--- a/crates/matrix-sdk-sqlite/src/lib.rs
+++ b/crates/matrix-sdk-sqlite/src/lib.rs
@@ -24,6 +24,9 @@ mod event_cache_store;
 #[cfg(feature = "state-store")]
 mod state_store;
 mod utils;
+use std::path::{Path, PathBuf};
+
+use deadpool_sqlite::PoolConfig;
 
 #[cfg(feature = "crypto-store")]
 pub use self::crypto_store::SqliteCryptoStore;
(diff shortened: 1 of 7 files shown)


crates/matrix-sdk-sqlite/src/lib.rs
```diff
@@ -35,3 +38,53 @@ pub use self::state_store::SqliteStateStore;
 
 #[cfg(test)]
 matrix_sdk_test::init_tracing_for_tests!();
+
+/// A configuration structure used for opening a store.
+pub struct StoreOpenConfig {
```
poljar, line 43: Perhaps a better name would be `StoreConfig` or `StoreConfiguration`. Maybe even mention SQLite somewhere since this is specific for the SQLite store implementation.

crates/matrix-sdk-sqlite/src/lib.rs
```diff
@@ -35,3 +38,53 @@ pub use self::state_store::SqliteStateStore;
...
+    /// Passphrase to open the store, if any.
+    passphrase: Option<String>,
+    /// The pool configuration for [`deadpool_sqlite`].
+    pool_config: PoolConfig,
+}
+
+impl StoreOpenConfig {
+    /// Create a new [`StoreOpenConfig`] with a path representing the directory
+    /// containing the store database, and its passphrase if encoded.
+    pub fn new<P>(path: P, passphrase: Option<&str>) -> Self
```
poljar, line 55: If we're creating a builder, perhaps it's better to move the optional `passphrase` parameter into its own setter.

poljar (approved):
Left some minor nits and suggestions.

Looks good.

crates/matrix-sdk-sqlite/src/lib.rs
Hywan: Okay to me to rename to `StoreConfig` (I don't like `Config`, I would prefer `Configuration`, but `Config` is used everywhere else in the code, consistency matters most). However, this is the `matrix-sdk-sqlite`, I wonder why we need to prefix everything by `Sqlite`, but again… consistency matters most. I'm changing to `SqliteStoreConfig`.

poljar: >  I wonder why we need to prefix everything by Sqlite

People complain about too many types with the same name. I think the complaint is warranted.

PR #4827, opened 2025-03-21
refactor(oauth): Allow to use any registration method with any login function

zecakeh:
This creates a `ClientRegistrationMethod` enum and use it for `OAuth::login` and `OAuth::login_with_qr_code`, allowing notable to use the `OAuthRegistrationStore` every time.

We can get rid of `url_for_oidc` since everything can be done with `OAuth::login` now, and we also merge `OAuth::finish_login` and `OAuth::login_with_oidc_callback()` to use the simplest API.

Finally there are some optimizations to reduce the number of requests by reusing the `AuthorizationServerMetadata`.


Hywan (approved):
More clean ups 👏! More clean ups 👏!

Thanks 🙇.

PR #4829, opened 2025-03-21
refactor(examples): Update OAuth examples

zecakeh:
To mention OAuth 2.0 instead of OpenID Connect.

Also use `OAuthRegistrationStore` in the oauth_cli example, since it's probably the recommended way to manage registrations.


andybalaam (approved):
Great, thank you!

PR #4831, opened 2025-03-22
refactor(oauth): Introduce `AccountManagementUrlBuilder`

zecakeh:
It allows to reuse the URL for different actions more easily than having to call `OAuth::account_management_url` every time for a different action.

### DIFF
--- a/bindings/matrix-sdk-ffi/src/client.rs
+++ b/bindings/matrix-sdk-ffi/src/client.rs
@@ -8,7 +8,7 @@ use anyhow::{anyhow, Context as _};
 use async_compat::get_runtime_handle;
 use matrix_sdk::{
     authentication::oauth::{
-        AccountManagementActionFull, ClientId, OAuthAuthorizationData, OAuthSession,
+        AccountManagementActionFull, ClientId, OAuthAuthorizationData, OAuthError, OAuthSession,
     },
     event_cache::EventCacheError,
     media::{
@@ -606,11 +606,24 @@ impl Client {
             return Ok(None);
         }
 
-        match self.inner.oauth().account_management_url(action.map(Into::into)).await {
-            Ok(url) => Ok(url.map(|u| u.to_string())),
+        let mut url_builder = match self.inner.oauth().account_management_url().await {
+            Ok(Some(url_builder)) => url_builder,
+            Ok(None) => return Ok(None),
             Err(e) => {
                 tracing::error!("Failed retrieving account management URL: {e}");
-                Err(e.into())
+                return Err(e.into());
+            }
+        };
+
+        if let Some(action) = action {
+            url_builder = url_builder.action(action.into());
+        }
+
+        match url_builder.build() {
+            Ok(url) => Ok(Some(url.to_string())),
+            Err(e) => {
+                tracing::error!("Failed to build account management URL: {e}");
+                Err(OAuthError::AccountManagementUrl(e).into())
             }
         }
     }
--- a/crates/matrix-sdk/src/authentication/oauth/account_management_url.rs
+++ b/crates/matrix-sdk/src/authentication/oauth/account_management_url.rs
@@ -73,43 +73,63 @@ pub enum AccountManagementActionFull {
     CrossSigningReset,
 }
 
-/// Build the URL for accessing the account management capabilities, as defined
-/// in [MSC].
+/// Builder for the URL for accessing the account management capabilities, as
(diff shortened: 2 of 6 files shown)


zecakeh:

bindings/matrix-sdk-ffi/src/client.rs
```diff
@@ -606,11 +606,24 @@ impl Client {
...
+        };
+
+        if let Some(action) = action {
+            url_builder = url_builder.action(action.into());
+        }
+
+        match url_builder.build() {
+            Ok(url) => Ok(Some(url.to_string())),
+            Err(e) => {
+                tracing::error!("Failed to build account management URL: {e}");
```
poljar, line 625: Let's import `error!()` if it's used in multiple places.

crates/matrix-sdk/src/authentication/oauth/account_management_url.rs
```diff
@@ -73,119 +73,174 @@ pub enum AccountManagementActionFull {
...
-/// A URL to be opened in a web browser where the end-user will be able to
-/// access the account management capabilities of the issuer.
-///
-/// # Errors
-///
-/// Returns an error if serializing the action fails.
+/// [`AccountManagementUrlBuilder::build()`] and
+/// [`AccountManagementUrlBuilder::build_or_ignore_action()`] return a URL to be
+/// opened in a web browser where the end-user will be able to access the
+/// account management capabilities of the issuer.
```
poljar, line 85: Can we get an example here or on the method that returns this struct? There's now enough indirection for this to warrant some examples.

poljar (changes requested):
Thanks, I left some small nits.

Could you clarify what you mean by:

> AccountManagementUrlBuilder::build_or_ignore_action() also allows to ignore errors when serializing actions to always have a URL to use (although I'm not sure how to get errors there).

Do you mean that the serialization should be infallible?

zecakeh:
> Could you clarify what you mean by:
> 
> > AccountManagementUrlBuilder::build_or_ignore_action() also allows to ignore errors when serializing actions to always have a URL to use (although I'm not sure how to get errors there).
> 
> Do you mean that the serialization should be infallible?

It means that I am not sure when serialization in `serde_html_form` would fail, I couldn't find anything in the docs about it.

From what I understand, it should not be possible for it to fail though. The only variable input is the `DeviceId` which can be any string, and according to the [`x-www-form-urlencoded`](https://url.spec.whatwg.org/#application/x-www-form-urlencoded) serializing rules, it should be possible to serialize any scalar value string. According to the Rust docs, [`char`](https://doc.rust-lang.org/std/primitive.char.html)s are Unicode scalar values, so there shouldn't be a problem there.

poljar:
> > Could you clarify what you mean by:
> > > AccountManagementUrlBuilder::build_or_ignore_action() also allows to ignore errors when serializing actions to always have a URL to use (although I'm not sure how to get errors there).
> > 
> > 
> > Do you mean that the serialization should be infallible?
> 
> It means that I am not sure when serialization in `serde_html_form` would fail, I couldn't find anything in the docs about it.
> 
> From what I understand, it should not be possible for it to fail though. The only variable input is the `DeviceId` which can be any string, and according to the [`x-www-form-urlencoded`](https://url.spec.whatwg.org/#application/x-www-form-urlencoded) serializing rules, it should be possible to serialize any scalar value string. According to the Rust docs, [`char`](https://doc.rust-lang.org/std/primitive.char.html)s are Unicode scalar values, so there shouldn't be a problem there.

Ah ok, yeah it seems like it should™ be infallible to me as well. But let us be cautious here.

zecakeh:
Actually, there is a way to avoid the error if we don't use `serde_html_form` at all, but instead use `Url::query_mut` and append the key-value pairs manually.

bindings/matrix-sdk-ffi/src/client.rs
zecakeh: Changed. It was already imported :stuck_out_tongue:.

crates/matrix-sdk/src/authentication/oauth/account_management_url.rs
zecakeh: I added an example.

zecakeh:
I got rid of potential errors by doing as I said.

poljar:
> I got rid of potential errors by doing as I said.

Ah neat.

Issue #4835, opened 2025-03-24
timeline: media attachments as (thread) replies


Johennes:
Following https://github.com/matrix-org/matrix-rust-sdk/pull/4819 I would like to enable sending media as (thread) replies.

I considered mimicking [`matrix_sdk_ui::Timeline::send_reply`](https://github.com/matrix-org/matrix-rust-sdk/blob/f8236a8b96ff685b23a80c6eb8c479c56f445174/crates/matrix-sdk-ui/src/timeline/mod.rs#L343) by routing `RepliedToInfo` and `EnforceThread` through from [`matrix_sdk_ffi::Timeline::send_attachment`](https://github.com/matrix-org/matrix-rust-sdk/blob/f8236a8b96ff685b23a80c6eb8c479c56f445174/bindings/matrix-sdk-ffi/src/timeline/mod.rs#L103). However, these types live in `matrix_sdk_ui` and for attachments the event content is only constructed down in `matrix_sdk` (for instance [here](https://github.com/matrix-org/matrix-rust-sdk/blob/f8236a8b96ff685b23a80c6eb8c479c56f445174/crates/matrix-sdk/src/send_queue/upload.rs#L186) for the send queue path).

Pulling the construction of the content up into `matrix_sdk_ui` seems tricky because it depends on the media sources. That leaves pushing the types down into `matrix_sdk` as the only option?

@bnjbvr curious if you have thoughts on this possibly.

Johennes:
Well this is much more detailed a plan than I had hoped for! 🤩

Sounds great, I'll give this a shot.

PR #4842, opened 2025-03-25
refactor(timeline): push the reply logic down into matrix_sdk

Johennes:
This achieves step 2 of #4835.


crates/matrix-sdk/src/room/reply.rs
```diff
@@ -0,0 +1,404 @@
...
+    Ok(RepliedToInfo {
+        event_id: event_id.to_owned(),
+        sender: event.sender().to_owned(),
+        timestamp: event.origin_server_ts(),
+        content: reply_content,
+    })
+}
+
+#[cfg(test)]
+mod tests {
```
Johennes, line 297: I wasn't sure how many tests to add here since we already have the tests in `replies.rs`.

crates/matrix-sdk-ui/src/timeline/error.rs
```diff
@@ -73,6 +74,10 @@ pub enum Error {
     #[error(transparent)]
     EditError(#[from] EditError),
 
+    /// An error happened while attempting to reply to an event.
+    #[error(transparent)]
+    ReplyError(#[from] ReplyError),
```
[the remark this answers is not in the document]
bnjbvr: Yep, it's fine :+1:

crates/matrix-sdk-ui/CHANGELOG.md
```diff
@@ -12,6 +12,8 @@ All notable changes to this project will be documented in this file.
 
 - [**breaking**] Optionally allow starting threads with `Timeline::send_reply`.
   ([4819](https://github.com/matrix-org/matrix-rust-sdk/pull/4819))
+- [**breaking**] Push the reply logic down into matrix_sdk.
```
bnjbvr, line 15: Agree it's a breaking change, but can you comment to which extent? Ideally we'd include all the names of types that have been changed here, but at least having some steps explaining how to migrate would be sweet :)

crates/matrix-sdk/src/room/reply.rs
```diff
@@ -0,0 +1,392 @@
...
+    MissingJson,
+    /// The event to reply to could not be found.
+    #[error("event to reply to not found")]
+    MissingEvent,
+    /// The event to reply to could not be deserialized.
+    #[error("failed to deserialize event to reply to")]
+    FailedToDeserializeEvent,
+    /// State events cannot be replied to.
+    #[error("tried to reply to a state event")]
+    StateEvent,
```
bnjbvr, line 81: Let's uniformize the names, since we're touching this error type anyways :)

- FailedToDeserializeEvent -> Deserialization

crates/matrix-sdk/src/room/reply.rs
```diff
@@ -0,0 +1,392 @@
...
+}
+
+/// Whether or not to enforce a [`Relation::Thread`] when sending a reply.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+#[allow(clippy::exhaustive_enums)]
```
bnjbvr, line 86: Out of curiosity, why is this line useful?

crates/matrix-sdk/src/room/reply.rs
```diff
@@ -0,0 +1,392 @@
...
+    ) -> impl Future<Output = Result<TimelineEvent, ReplyError>> + SendOutsideWasm;
+}
+
+impl EventSource for &Room {
+    async fn get_event(&self, event_id: &EventId) -> Result<TimelineEvent, ReplyError> {
+        self.load_or_fetch_event(event_id, None)
+            .await
+            .map_err(|err| ReplyError::Fetch(Box::new(err)))
+    }
+}
```
bnjbvr, line 139: Consider sharing the trait with `room/edit.rs` :)

crates/matrix-sdk/src/room/reply.rs
```diff
@@ -0,0 +1,392 @@
...
+    #[async_test]
+    async fn test_reply_to_state_event() {
+        let event_id = event_id!("$1");
+        let own_user_id = user_id!("@me:saucisse.bzh");
+
+        let mut cache = TestEventCache::default();
+
+        cache.events.insert(
+            event_id.to_owned(),
+            // TODO: use the EventFactory for state events too.
```
bnjbvr, line 325: We can, now! :D

crates/matrix-sdk/src/room/reply.rs
bnjbvr: It would be sweet to have a smoke test + have tests that reach each `ReplyError` variant. (We can also *remove* some tests from the timeline's code, if they were only testing the code moved here; otherwise the timeline is now testing this layer, and that's a layer mismatch.)

bnjbvr (changes requested):
Thanks, looks like this is on the right track! A few small comments below, but looks close to merging.

crates/matrix-sdk/src/room/reply.rs
Johennes: Sorry, this crept in with https://github.com/matrix-org/matrix-rust-sdk/pull/4819/files#diff-7f49982c21dd9af0fcce15d1d1349716f42c41268483df13c1dc8086c6448d57R143, I think because I was initially copying the `ForwardThread` enum from ruma. I'll remove it.

crates/matrix-sdk/src/room/reply.rs
Johennes: How shall we deal with the different error types in the two traits? I could use an associated type but then we'd still be duplicating the implementations. 🤔

```
trait EventSource {
    type Error: std::error::Error + Send + Sync + 'static;

    fn get_event(
        &self,
        event_id: &EventId,
    ) -> impl Future<Output = Result<TimelineEvent, Self::Error>> + SendOutsideWasm;
}

impl EventSource for &Room {
    type Error = ReplyError;

    async fn get_event(&self, event_id: &EventId) -> Result<TimelineEvent, ReplyError> {
        self.load_or_fetch_event(event_id, None)
            .await
            .map_err(|err| ReplyError::Fetch(Box::new(err)))
    }
}

impl EventSource for &Room {
    type Error = EditError;

    async fn get_event(&self, event_id: &EventId) -> Result<TimelineEvent, ReplyError> {
        self.load_or_fetch_event(event_id, None)
            .await
            .map_err(|err| EditError::Fetch(Box::new(err)))
    }
}
```

crates/matrix-sdk/src/room/reply.rs
Johennes: Why do I always step into these? 🤦‍♂️😅

crates/matrix-sdk-ui/CHANGELOG.md
Johennes: Yes, true. Have expanded it.

crates/matrix-sdk/src/room/reply.rs
Johennes: Done. The two `Missing*` variants also turned out to be unused. So I removed them.

crates/matrix-sdk/src/room/reply.rs
Johennes: Fixed here and in `edit.rs`.

bnjbvr (approved):
Thanks! Sorry I forgot to send the review comment explaining how to share more code yesterday evening.

Johennes:
No worries at all. Your turnaround time on this is amazing. Thanks a ton! 👌

bnjbvr (approved):
Nice, thanks!

PR #4850, opened 2025-03-26
refactor(base): Introduce `BaseClient::activate` and split `BaseStateStore::set_session_meta` to 3 methods

Hywan:
This set of patches is twofold:

1. it renames `set_session_meta` to `set_or_reload_session` to clarify that the state _can be reloaded_, which is not obvious for a setter,
2. a new `Store::set_or_reload_session_from_other` method is added, to derive a state store from another one.

Finally, this patch adds missing unit tests and adds one for the `…from_other` method.

See https://github.com/matrix-org/matrix-rust-sdk/pull/4850#issuecomment-2761547063 to see the next step we went for: `BaseClient::activate` and `BaseStateStore::set_session_meta` + `::load_rooms` + `::load_sync_token` + `::derive_from_other`.

This refactoring is necessary regarding https://github.com/matrix-org/matrix-rust-sdk/issues/4801 because I'm about to add another data in `Store`. Having this `set_or_reload_session_from_other` is a nice abstraction for now, but also in a short future.

---

* Address https://github.com/matrix-org/matrix-rust-sdk/issues/4801

### DIFF
--- a/benchmarks/benches/room_bench.rs
+++ b/benchmarks/benches/room_bench.rs
@@ -61,7 +61,7 @@ pub fn receive_all_members_benchmark(c: &mut Criterion) {
     );
 
     runtime
-        .block_on(base_client.set_or_reload_session(
+        .block_on(base_client.activate(
             SessionMeta {
                 user_id: user_id!("@somebody:example.com").to_owned(),
                 device_id: device_id!("DEVICE_ID").to_owned(),
--- a/benchmarks/benches/room_bench.rs
+++ b/benchmarks/benches/room_bench.rs
@@ -61,7 +61,7 @@ pub fn receive_all_members_benchmark(c: &mut Criterion) {
     );
 
     runtime
-        .block_on(base_client.set_session_meta(
+        .block_on(base_client.set_or_reload_session(
             SessionMeta {
                 user_id: user_id!("@somebody:example.com").to_owned(),
                 device_id: device_id!("DEVICE_ID").to_owned(),
--- a/crates/matrix-sdk-base/src/client.rs
+++ b/crates/matrix-sdk-base/src/client.rs
@@ -115,7 +115,7 @@ pub struct BaseClient {
 
     /// The olm-machine that is created once the
     /// [`SessionMeta`][crate::session::SessionMeta] is set via
-    /// [`BaseClient::set_or_reload_session`]
+    /// [`BaseClient::activate`]
     #[cfg(feature = "e2e-encryption")]
     olm_machine: Arc<RwLock<Option<OlmMachine>>>,
 
@@ -289,16 +289,21 @@ impl BaseClient {
         &self.event_cache_store
     }
 
-    /// Is the client logged in.
-    pub fn logged_in(&self) -> bool {
+    /// Check whether the client has been activated.
+    ///
+    /// See [`BaseClient::activate`] to know what it means.
+    pub fn is_activated(&self) -> bool {
         self.state_store.session_meta().is_some()
     }
 
-    /// Set the [`SessionMeta`] for this client, and if a state exists for this
-    /// session, reload it.
+    /// Activate the client.
     ///
-    /// If encryption is enabled, this also initializes or restores the
-    /// `OlmMachine`.
+    /// A client is considered active when:
+    ///
+    /// 1. It has a `SessionMeta` (user ID, device ID and access token),
+    /// 2. Has loaded cached data from storage,
+    /// 3. If encryption is enabled, it also initialized or restored its
+    ///    `OlmMachine`.
     ///
     /// # Arguments
     ///
@@ -317,14 +322,14 @@ impl BaseClient {
     /// # Panics
     ///
     /// This method panics if it is called twice.
-    pub async fn set_or_reload_session(
+    pub async fn activate(
         &self,
         session_meta: SessionMeta,
         #[cfg(feature = "e2e-encryption")] custom_account: Option<
             crate::crypto::vodozemac::olm::Account,
         >,
     ) -> Result<()> {
-        debug!(user_id = ?session_meta.user_id, device_id = ?session_meta.device_id, "Restoring login");
+        debug!(user_id = ?session_meta.user_id, device_id = ?session_meta.device_id, "Activating the client");
         self.state_store
             .set_or_reload_session(session_meta.clone(), &self.room_info_notable_update_sender)
             .await?;
@@ -814,7 +819,7 @@ impl BaseClient {
         } else {
             // If we have no `OlmMachine`, just return the events that were passed in.
             // This should not happen unless we forget to set things up by calling
-            // `Self::set_or_reload_session()`.
+            // `Self::activate()`.
             Ok(encryption_sync_changes.to_device_events)
         }
     }
@@ -2260,7 +2265,7 @@ mod tests {
         let client =
             BaseClient::new(StoreConfig::new("cross-process-store-locks-holder-name".to_owned()));
         client
-            .set_or_reload_session(
+            .activate(
                 SessionMeta { user_id: user_id.to_owned(), device_id: "FOOBAR".into() },
                 #[cfg(feature = "e2e-encryption")]
                 None,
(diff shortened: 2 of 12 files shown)


crates/matrix-sdk-base/src/store/mod.rs
```diff
@@ -243,6 +241,25 @@ impl Store {
...
     }
 
+    /// Similar to [`Store::set_session_meta_and_reload_state`] but takes values
+    /// from another existing [`Store`].
+    #[cfg(any(feature = "e2e-encryption", test))]
+    pub(crate) async fn set_session_meta_and_reload_state_from_other(
+        &self,
+        other: &Self,
+        room_info_notable_update_sender: &broadcast::Sender<RoomInfoNotableUpdate>,
+    ) -> Result<()> {
```
bnjbvr, line 251: Instead of introducing a new function, could the caller call the existing `set_session_meta` with the session_meta extracted from the other store, instead? I don't see the point of duplicating the function when we could compose function calls instead :thinking:

crates/matrix-sdk-base/src/store/mod.rs
```diff
@@ -205,13 +205,11 @@ impl Store {
...
-    /// Set the meta of the session.
-    ///
-    /// Restores the state of this `Store` from the given `SessionMeta` and the
-    /// inner `StateStore`.
+    /// Set a [`SessionMeta`] and reload its state and its inner
+    /// [`StateStore`].
     ///
     /// This method panics if it is called twice.
-    pub async fn set_session_meta(
+    pub async fn set_session_meta_and_reload_state(
```
bnjbvr, line 212: Agreed that a renaming was long overdue here, because it's easy to miss that it's also reloading all the rooms. However, this name is super long, and phrase-looking names are painful to read. I would suggest one of two things:

1. either split the function so that setting the session meta happens independently of restoring the room infos,
2. or consider another shorter name; `restore_session`? `reload_session`? etc.

bnjbvr (changes requested):

crates/matrix-sdk-base/src/store/mod.rs
Hywan: Because I'm going to add a new parameter, as the commit description mentions. Instead of exposing the internal data and let the caller deriving/composing with them, I believe it's a better design to have a dedicated method for that. It's also unit testable and documentable.

crates/matrix-sdk-base/src/store/mod.rs
Hywan: I spent more time I would like to admit on finding a new name here. 

- `restore_` is not always correct because this method is used when the state store is created from scratch or when it is restored, so it's a half lie.
- `reload_` has the same problem, if the state store is created for the first time, it reloads nothing.

That's why I went for an `and`.

Also splitting in two methods is not something I want to do here. All call sites will have to call two methods.

What we could do is:

- having `set_session_meta` alone, and private
- having `load_room_infos` alone, and private
- having `load` (bad name), a public method, that calls `set_session_meta` internally
- having `reload`, a public method, that calls `set_session_meta` and `load_room_infos` internally

I don't like the `load` name. It's convenient because it plays nicely with `reload` but it loads nothing. Maybe `set_session` and `reset_session`? I think I like those. Thoughts?

The problem is that this state store isn't really set up until the session meta isn't set. The store isn't fully constructed semantically speaking.

bnjbvr: > I spent more time I would like to admit on finding a new name here.
> 
>     * `restore_` is not always correct because this method is used when the state store is created from scratch or when it is restored, so it's a half lie.
> 
>     * `reload_` has the same problem, if the state store is created for the first time, it reloads nothing.
> 
> 
> That's why I went for an `and`.

Well, `reload` in the name you've chosen is also a lie, then? :innocent: I appreciate your points, but I would suggest keeping it simple and short, rather than making it perfectly self-explanatory (which would imply a renaming like `set_session_meta_and_maybe_reload_state`). Having a doc comment explaining what it does is fine. `restore_session` IMO conveys the meaning: it's taking a session meta, so that makes sense; it's restoring Some™ state. In the initial setup it might not restore any state, but that's fine; it's not supposed to be used externally anyways, except when you're manipulating a `ClientBase` by yourself, which you ought not to :innocent:

Hywan: I've renamed `set_session_meta` to `set_or_reload_session`, which I believe convey the idea better, and is still a short name.

Hywan:
Rebased on top of `main` to solve conflicts.

crates/matrix-sdk-base/src/store/mod.rs
```diff
@@ -562,3 +562,84 @@ impl StoreConfig {
...
+        store
+            .set_or_reload_session(session_meta.clone(), &room_info_notable_update_sender)
+            .await
+            .unwrap();
+
+        // Kaboom.
+        store
+            .set_or_reload_session(session_meta.clone(), &room_info_notable_update_sender)
+            .await
+            .unwrap();
```
poljar, line 617: I think an `expect_err()` explaining why this should panic would have been much nicer.

benchmarks/benches/room_bench.rs
```diff
@@ -61,7 +61,7 @@ pub fn receive_all_members_benchmark(c: &mut Criterion) {
     );
 
     runtime
-        .block_on(base_client.set_session_meta(
+        .block_on(base_client.set_or_reload_session(
```
poljar, line 64: I'm probably late to the party, but let me add my suggestion anyways since `set_or_reload()` seems to imply we're doing one or the other.

Since the issue here is that there's more to the function than just setting the session metadata, perhaps a more higher level concept could be introduced.

Something that describes in fully that after you call this method the `Client` object will become usable to send/receive messages.

I was thinking that the `Client` might become `active`, so perhaps `activate_client()` and we describe what an active client is:

1. One that has the session meta set (user id, device id, access token)
2. Has loaded cached data from storage

But feel free to ignore this suggestion since this has been going on for quite a bit now.

poljar (approved):

crates/matrix-sdk-base/src/store/mod.rs
poljar: Oh indeed, ignore me then.

Hywan:
As discussed in private, we have found a satisfying refactoring and renaming for these methods.

---

A patch renames `BaseClient::set_or_reload_session`
to `BaseClient::activate`, and `BaseClient::logged_in` to
`BaseClient::is_activated`.

The idea behind these renamings is to introduce a “state” for the
`BaseClient`: it is activated when is has a `SessionMeta`, has loaded
its data from the storages, and has an `OlmMachine`. Consequently, the
`logged_in` method is renamed `is_activated` for the symmetry. If one
wants to know if the client is logged in, it can use `is_activated`, or
also `MatrixAuth::logged_in`.

---

Another patch splits `BaseStateStore::set_or_reload_session` into 3
distinct methods:

1. `set_session_meta` (hello again!),
2. `load_rooms`,
3. `load_sync_token`.

This patch also renames
`BaseStateStore::set_or_reload_session_from_other` into
`derive_from_other` to clarify its semantics. It calls these 3 methods
above as a combo.

Finally, this patch also updates `BaseClient::activate` to call these 3
methods above individually.

---

Here we are.

crates/matrix-sdk-base/src/client.rs
```diff
@@ -288,15 +286,21 @@ impl BaseClient {
         &self.event_cache_store
     }
 
-    /// Is the client logged in.
-    pub fn logged_in(&self) -> bool {
+    /// Check whether the client has been activated.
+    ///
+    /// See [`BaseClient::activate`] to know what it means.
+    pub fn is_activated(&self) -> bool {
```
poljar, line 292: `has_been_activated` or `is_active` sounds like the more correct grammar here, like you used in the description of this method :sweat_smile:.

poljar (approved):
Left another small nit about the `is_activated()` naming, looks good otherwise.

crates/matrix-sdk-base/src/client.rs
Hywan: 🤦  done!

PR #4851, opened 2025-03-26
refactor(base): `Store` is renamed `BaseStateStore`, and `store()` methods are renamed `state_store()`.

Hywan:
This set of patches rename the `Store` struture to `BaseStateStore` (`StateStore` was already taken by `traits::StateStore`, let's avoid confusions). It also renames `BaseClient::store()` and `Client::store()` to `state_store()` for the sake of consistency.

---

* Part of the https://github.com/matrix-org/matrix-rust-sdk/issues/4801 journey


stefanceriu (approved):
Don't know much about these but it makes sense to me ¯\_(ツ)_/¯

PR #4852, opened 2025-03-26
feat!(timeline): allow sending media as (thread) replies

Johennes:
This implements step 3 of #4835.


crates/matrix-sdk/src/room/reply.rs
```diff
@@ -111,6 +116,45 @@ impl Room {
...
+    ///
+    /// # Arguments
+    ///
+    /// * `event_id` - ID of the event to reply to
+    pub async fn replied_to_info_from_event_id(
+        &self,
+        event_id: &EventId,
+    ) -> Result<RepliedToInfo, ReplyError> {
+        replied_to_info_from_event_id(self, event_id).await
+    }
```
Johennes, line 157: I extracted these methods so that, when the requested event cannot be replied to, we can fail early and before starting to process any media. It does, unfortunately, make the interface a little more complicated.

bindings/matrix-sdk-ffi/src/timeline/mod.rs
```diff
@@ -201,14 +228,27 @@ pub struct UploadParameters {
     caption: Option<String>,
     /// Optional HTML-formatted caption, for clients that support it.
     formatted_caption: Option<FormattedBody>,
-    // Optional intentional mentions to be sent with the media.
+    /// Optional intentional mentions to be sent with the media.
     mentions: Option<Mentions>,
+    /// Optional parameters for sending the media as (threaded) reply.
+    reply_params: Option<ReplyParameters>,
```
Johennes, line 234: I've added this here to not have to duplicate the `send_*` methods for the various media types. Maybe we could also use this to merge `send_reply` and `send_thread_reply`?

crates/matrix-sdk/src/room/reply.rs
bnjbvr: Yeah, this gets us back to a place where responding is a two-process step:

1. get the replied-to info
2. call the `make_reply_event_with_info` method

We could avoid this new method and the one above, by reusing `make_reply_event`. I think this would make `Room::make_attachment_event` async+fallible (while it's not any of these two right now), but that seems fine; both callers are already async+fallible too.

Can you try this out please? If it turns out to be too complicated, we can consider a two-steps process like now, but having a super streamlined API would be sweet.

crates/matrix-sdk/src/attachment.rs
```diff
@@ -188,6 +190,8 @@ pub struct AttachmentConfig {
     pub(crate) caption: Option<String>,
     pub(crate) formatted_caption: Option<FormattedBody>,
     pub(crate) mentions: Option<Mentions>,
+    pub(crate) replied_to_info: Option<RepliedToInfo>,
+    pub(crate) enforce_thread: Option<EnforceThread>,
```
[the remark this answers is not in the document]
bnjbvr: Exactly; with respect to my other comment, I think this could be `Option<Reply>`, with `struct Reply { event_id: OwnedEventId, enforce_thread: EnforceThread }` or something like that. The builder methods could be fused too.

crates/matrix-sdk-ui/CHANGELOG.md
```diff
@@ -17,6 +17,9 @@ All notable changes to this project will be documented in this file.
   [`Timeline::send_reply()`] now takes an event ID rather than a `RepliedToInfo`.
   `Timeline::replied_to_info_from_event_id` has been made private in `matrix_sdk`.
   ([4842](https://github.com/matrix-org/matrix-rust-sdk/pull/4842))
+- [**breaking**] Allow sending media as (thread) replies. The reply behaviour
```
bnjbvr, line 20: Not sure it's a breaking change this time, is it? :D

crates/matrix-sdk-ui/tests/integration/timeline/media.rs
```diff
@@ -128,6 +137,10 @@ async fn test_send_attachment_from_file() {
         assert_let!(MessageType::File(file) = msg.msgtype());
         assert_let!(MediaSource::Plain(uri) = &file.source);
         assert!(uri.to_string().contains("localhost"));
+
+        // The message should be considered part of the thread.
+        let aggregated = item.content().as_aggregated().unwrap();
+        assert!(aggregated.is_threaded());
```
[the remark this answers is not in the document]
bnjbvr: (Confirming we'd like more tests, please :))

bnjbvr (changes requested):
Thanks! A first round of review comments here, but it looks like this is on the right track!

crates/matrix-sdk-ui/CHANGELOG.md
Johennes: Err .. no, I think it's not. 🤦‍♂️

crates/matrix-sdk/src/room/reply.rs
Johennes: Yeah, this seems to work well actually. 👍

crates/matrix-sdk-ui/tests/integration/timeline/media.rs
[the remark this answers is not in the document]
Johennes: I've added more tests but kept this test. Interestingly this fails to compile on the CI now. I'm not sure what the issue is. I cannot reproduce this issue locally. Do you have any clues possibly?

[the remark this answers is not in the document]
bnjbvr: Do you reproduce the issues when running `cargo xtask ci` (or one of its subcommands)?

bindings/matrix-sdk-ffi/src/timeline/mod.rs
```diff
@@ -99,7 +99,7 @@ impl Timeline {
         unsafe { Arc::from_raw(Arc::into_raw(inner) as _) }
     }
 
-    fn send_attachment(
+    async fn send_attachment(
```
bnjbvr, line 102: Looks like this doesn't need to be async anymore? (and ditto for callers)

bindings/matrix-sdk-ffi/src/timeline/mod.rs
bnjbvr: Yep, that would make sense (maybe as a followup PR, to not have this one drag for too long?)

bindings/matrix-sdk-ffi/src/error.rs
```diff
@@ -288,6 +288,8 @@ pub enum RoomError {
     TimelineUnavailable,
     #[error("Invalid thumbnail data")]
     InvalidThumbnailData,
+    #[error("Invalid reply parameters")]
+    InvalidReplyParameters,
```
bnjbvr, line 292: Seems it's only used in one case when the event id is incorrect; can we specialize the error message to reflect that, please?

crates/matrix-sdk/src/send_queue/mod.rs
```diff
@@ -1841,6 +1841,10 @@ pub enum RoomSendQueueError {
     /// Error coming from storage.
     #[error(transparent)]
     StorageError(#[from] RoomSendQueueStorageError),
+
+    /// The attachment event failed to be created.
+    #[error("the attachment event could not be created")]
+    AttachmentError,
```
bnjbvr, line 1847: nit: since it's not a transparent error type, let's not use the `Error` suffix, and rename it to `FailedToCreateAttachment`

crates/matrix-sdk-ui/tests/integration/timeline/media.rs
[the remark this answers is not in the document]
bnjbvr: If you've rebased to latest main, this function might have been renamed to `as_msglike()`.

crates/matrix-sdk/tests/integration/room/attachment/mod.rs
```diff
@@ -297,6 +303,328 @@ async fn test_room_attachment_send_mentions() {
...
+                })),
+        )
+        .await
+        .unwrap();
+
+    assert_eq!(expected_event_id, response.event_id);
+}
+
+#[async_test]
+async fn test_room_attachment_reply_on_thread() {
```
bnjbvr, line 432: I think this test is semantically equivalent to the `start_thread` one; maybe we can get rid of one of the twos, unless I've missed some important distinction between those?

bnjbvr (approved):
Nice, we're getting close to landing this! LGTM on principle that we can address the last comments without large changes.

bindings/matrix-sdk-ffi/src/timeline/mod.rs
Johennes: 🤦‍♂️ Sorry, yes.

bindings/matrix-sdk-ffi/src/timeline/mod.rs
Johennes: Separate PR makes sense. 👍

crates/matrix-sdk-ui/tests/integration/timeline/media.rs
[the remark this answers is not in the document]
Johennes: Merging `main` helped me to reproduce it locally (and, yes, a switch to `as_msglike` was needed apparently). Thanks a lot for the pointers.

crates/matrix-sdk/tests/integration/room/attachment/mod.rs
Johennes: Yeah, you're right. It doesn't add much more. Have deleted it.

Johennes:
Thanks a lot for your patience with me figuring things out through multiple rounds of review here. 🙏

bnjbvr:
No worries, thanks a lot for this contribution!

PR #4853, opened 2025-03-27
timeline: Aggregated timeline item content cleanup

stefanceriu:
This is a follow up to #4839 that:
* simplifies the tests by using the `as_message`/`is_message` and sibling helper methods
* addresses the `fetch_replied_to_event` simplification from 4839
* renames `AggregatedTimelineItemContent` to `MsgLikeContent` and `AggregatedTimelineItemContentKind` to `MsgLikeKind`.


crates/matrix-sdk-ui/src/timeline/event_item/content/mod.rs
```diff
@@ -348,6 +348,13 @@ impl TimelineItemContent {
         aggregated.unwrap_or(None)
     }
 
+    pub fn as_sticker(&self) -> Option<&Sticker> {
+        let aggregated = as_variant!(self, Self::Aggregated)
+            .map(|f| as_variant!(&f.kind, AggregatedTimelineItemContentKind::Sticker));
+
+        aggregated.unwrap_or(None)
```
bnjbvr, line 355: Let's wait for #4854 :)

crates/matrix-sdk-ui/src/timeline/controller/mod.rs
```diff
@@ -1523,15 +1514,12 @@ impl<P: RoomDataProvider> TimelineController<P, (OlmMachine, OwnedRoomId)> {
...
-#[allow(clippy::too_many_arguments)]
 async fn fetch_replied_to_event(
     mut state: RwLockWriteGuard<'_, TimelineState>,
     index: usize,
     item: &EventTimelineItem,
     internal_id: TimelineUniqueId,
-    message: &Message,
-    reactions: &ReactionsByKeyBySender,
-    thread_root: &Option<OwnedEventId>,
+    aggregated: &AggregatedTimelineItemContent,
```
bnjbvr, line 1522: I don't think you need this parameter either; `item` holds this aggregated content under the hood, right?

bnjbvr (changes requested):
Thanks, looks good so far! I think in terms of naming, there's no obvious super nice choice here. For consistency with Ruma, we could call these `MsgLike`, as this is the nomenclature for such Ruma events. Would make for lines like `content().as_msglike()?.as_message()?`, which isn't *too* weird once you're acquainted with it. We'd define in the doc comment what this is more precisely. So, let's roll with `MsgLikeContent` / `MsgLikeKind`?

crates/matrix-sdk-ui/src/timeline/controller/mod.rs
bnjbvr: Ah good point. We can either let it like this, or use an `.expect()` statement that the item is an aggregated item (and specify it clearly in this function's signature). But overall, since this function is only called once, I'm also tempted to suggest inlining it into its caller…

stefanceriu: I tried both but not quite happy with how it looked, I think I would like to just leave it alone for now if you don't mind 🤔

crates/matrix-sdk-ui/src/timeline/event_item/content/msg_like.rs
```diff
@@ -60,4 +61,10 @@ impl MsgLikeContent {
     pub fn with_in_reply_to(&self, in_reply_to: InReplyToDetails) -> Self {
         Self { in_reply_to: Some(in_reply_to), ..self.clone() }
     }
+
+    /// If `kind` is of the [`MsgLikeKind`][MsgLikeKind::Message] variant,
+    /// return the inner [`Message`].
+    pub fn as_message(&self) -> Option<Message> {
+        as_variant!(self.kind.clone(), MsgLikeKind::Message(message) => message)
```
bnjbvr, line 68: nit: move the `.clone()` to the right-hand side, so that we don't clone when `self` isn't a message

bnjbvr (approved):
Yay for shorter names! Thanks for tackling this quickly :partying_face:

PR #4854, opened 2025-03-27
refactor(timeline): bump as_variant and make more use of its pattern matching form

bnjbvr:
Thanks to @jplatte for a quick merge and release of `as_variant!`.


crates/matrix-sdk-ui/src/timeline/event_item/content/mod.rs
```diff
@@ -333,19 +333,29 @@ impl TimelineItemContent {
...
-        aggregated.unwrap_or(None)
+        as_variant!(
+            self,
+            Self::Aggregated(
+                AggregatedTimelineItemContent {
+                    kind: AggregatedTimelineItemContentKind::Poll(poll),
+                    ..
+                }
+            ) => poll
+        )
```
jplatte, line 358: Potential alternative:

```rust
let aggregated = as_variant!(self, Self::Aggregated)?;
as_variant!(aggregated.kind, AggregatedTimelineItemContentKind::Poll)
```

poljar: This variant improves the readability indeed.

poljar (approved):

crates/matrix-sdk-ui/src/timeline/event_item/content/mod.rs
bnjbvr: Yeah, that would've been equivalent to the simplified previous implementation. I don't have strong opinions here; OTOH in like the dense pattern matching that shows precisely what we extract where; OTOH, rustfmt didn't format it automatically, and a single-line construct was super long to read.

bnjbvr: Well I guess we didn't need the new pattern matching rule then :D Will update to the shorter form.

Issue #4147, opened 2024-10-17
Option to exclude insecure devices for Olm traffic, to be consistent with megolm sharing?


BillCarsonFr:
Following the work on invisible crypto it is now possible to exclude devices not signed by their owner to participate in encrypted room:
- When encrypting via [CollectStrategy/IdentityBased](https://github.com/matrix-org/matrix-rust-sdk/blob/c7708d6154f2c56165494280e4c14cfdf7226a53/crates/matrix-sdk-crypto/src/session_manager/group_sessions/share_strategy.rs#L40C10-L40C25)
- When decrypting via [TrustRequirement](https://github.com/matrix-org/matrix-rust-sdk/blob/bdfe64179bd0b5ef5598774ac510d1539efc2184/crates/matrix-sdk-crypto/src/lib.rs#L127)

We also have an API to encrypt a message directly to a device (using olm):
https://github.com/matrix-org/matrix-rust-sdk/blob/bdfe64179bd0b5ef5598774ac510d1539efc2184/crates/matrix-sdk-crypto/src/identities/device.rs#L407-L412

This can lead to a strange situation when the `exclude insecure devices` option is enabled where:
- Sharing a room key won't work for a given device
- Encrypting directly to this device would work

Maybe we would want to use some `EncryptionSetting`/`DecryptionSetting` for olm traffic?
Notice that we also don't look at the device trust status `BlackListed`/`Ignored` when encrypting in olm

fkwp:
@toger5

dkasak:
I had a short discussion about this with @poljar now.

Ideally we would reject all encrypted to-device message types if the sender device is insecure. However, there are some caveats with some standardised message types:

- Message keys, i.e. `m.room_key`.
  - If there's a delay in securing the device and you receive a message key in the meantime, we would reject it. But suppose we learn that the device is secured a short time afterwards—now we'd accept the message key we've rejected and the consequence of the rejection is that we'll now get ~100 UTDs from this user.
  - So we accept room keys from insecure devices but keep them in a staged area.
- There are to-device message types we also accept in cleartext. It might be silly to accept them in cleartext but reject them when they're encrypted just because they're coming from an insecure device.
  - These include `m.key.verification`, `m.secret.request`, `m.room_key.withheld`.
  - We accept these from insecure devices.

[...]

PR #4856, opened 2025-03-27
WidgetDriver: add to_device support

toger5:
Updated version of: https://github.com/matrix-org/matrix-rust-sdk/pull/4155

This introduces widget driver support for sending to devices messages: "A widget (element call) can request send an action via postmessage to the client (element X). This PRs job is to do everything from parsing the json to sending the to_device cs api request to sending back the response."

Signed-off-by:

### DIFF
--- a/crates/matrix-sdk/src/widget/capabilities.rs
+++ b/crates/matrix-sdk/src/widget/capabilities.rs
@@ -18,13 +18,14 @@
 use std::fmt;
 
 use async_trait::async_trait;
-use ruma::{events::AnyTimelineEvent, serde::Raw};
 use serde::{ser::SerializeSeq, Deserialize, Deserializer, Serialize, Serializer};
-use tracing::{debug, error};
+use tracing::debug;
 
 use super::{
-    filter::MatrixEventFilterInput, EventFilter, MessageLikeEventFilter, StateEventFilter,
+    filter::{Filter, FilterInput},
+    MessageLikeEventFilter, StateEventFilter,
 };
+use crate::widget::filter::ToDeviceEventFilter;
 
 /// Must be implemented by a component that provides functionality of deciding
 /// whether a widget is allowed to use certain capabilities (typically by
@@ -42,9 +43,9 @@ pub trait CapabilitiesProvider: Send + Sync + 'static {
 #[cfg_attr(test, derive(PartialEq))]
 pub struct Capabilities {
     /// Types of the messages that a widget wants to be able to fetch.
-    pub read: Vec<EventFilter>,
+    pub read: Vec<Filter>,
     /// Types of the messages that a widget wants to be able to send.
-    pub send: Vec<EventFilter>,
+    pub send: Vec<Filter>,
     /// If this capability is requested by the widget, it can not operate
     /// separately from the matrix client.
     ///
@@ -58,17 +59,17 @@ pub struct Capabilities {
 }
 
 impl Capabilities {
-    /// Tells if a given raw event matches the read filter.
-    pub fn raw_event_matches_read_filter(&self, raw: &Raw<AnyTimelineEvent>) -> bool {
-        let filter_in = match raw.deserialize_as::<MatrixEventFilterInput>() {
-            Ok(filter) => filter,
-            Err(err) => {
-                error!("Failed to deserialize raw event as MatrixEventFilterInput: {err}");
-                return false;
-            }
-        };
-
-        self.read.iter().any(|f| f.matches(&filter_in))
+    pub(super) fn allow_reading(&self, event_filter_input: &FilterInput) -> bool {
+        self.read.iter().any(|f| f.matches(event_filter_input))
+    }
+    pub(super) fn allow_sending(&self, event_filter_input: &FilterInput) -> bool {
+        self.send.iter().any(|f| f.matches(event_filter_input))
+    }
+    /// Checks if a filter exists for the given event type, useful for
+    /// optimization. Avoids unnecessary requests when no matching filter is
+    /// present.
+    pub(super) fn has_read_filter_for_type(&self, event_type: &str) -> bool {
+        self.read.iter().any(|f| f.filter_event_type() == event_type)
     }
 }
 
@@ -85,12 +86,13 @@ impl Serialize for Capabilities {
     where
         S: Serializer,
     {
-        struct PrintEventFilter<'a>(&'a EventFilter);
+        struct PrintEventFilter<'a>(&'a Filter);
         impl fmt::Display for PrintEventFilter<'_> {
             fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
                 match self.0 {
-                    EventFilter::MessageLike(filter) => PrintMessageLikeEventFilter(filter).fmt(f),
-                    EventFilter::State(filter) => PrintStateEventFilter(filter).fmt(f),
+                    Filter::MessageLike(filter) => PrintMessageLikeEventFilter(filter).fmt(f),
+                    Filter::State(filter) => PrintStateEventFilter(filter).fmt(f),
+                    Filter::ToDevice(filter) => filter.fmt(f),
                 }
             }
         }
@@ -136,15 +138,15 @@ impl Serialize for Capabilities {
         }
         for filter in &self.read {
             let name = match filter {
-                EventFilter::MessageLike(_) => READ_EVENT,
-                EventFilter::State(_) => READ_STATE,
+                Filter::MessageLike(_) => READ_EVENT,
+                Filter::State(_) => READ_STATE,
             };
             seq.serialize_element(&format!("{name}:{}", PrintEventFilter(filter)))?;
         }
         for filter in &self.send {
             let name = match filter {
-                EventFilter::MessageLike(_) => SEND_EVENT,
-                EventFilter::State(_) => SEND_STATE,
+                Filter::MessageLike(_) => SEND_EVENT,
+                Filter::State(_) => SEND_STATE,
             };
             seq.serialize_element(&format!("{name}:{}", PrintEventFilter(filter)))?;
         }
@@ -162,8 +164,8 @@ impl<'de> Deserialize<'de> for Capabilities {
             RequiresClient,
             UpdateDelayedEvent,
             SendDelayedEvent,
-            Read(EventFilter),
-            Send(EventFilter),
+            Read(Filter),
+            Send(Filter),
             Unknown,
         }
 
@@ -184,17 +186,17 @@ impl<'de> Deserialize<'de> for Capabilities {
                 }
 
                 match s.split_once(':') {
-                    Some((READ_EVENT, filter_s)) => Ok(Permission::Read(EventFilter::MessageLike(
+                    Some((READ_EVENT, filter_s)) => Ok(Permission::Read(Filter::MessageLike(
                         parse_message_event_filter(filter_s),
                     ))),
-                    Some((SEND_EVENT, filter_s)) => Ok(Permission::Send(EventFilter::MessageLike(
+                    Some((SEND_EVENT, filter_s)) => Ok(Permission::Send(Filter::MessageLike(
                         parse_message_event_filter(filter_s),
                     ))),
                     Some((READ_STATE, filter_s)) => {
-                        Ok(Permission::Read(EventFilter::State(parse_state_event_filter(filter_s))))
+                        Ok(Permission::Read(Filter::State(parse_state_event_filter(filter_s))))
                     }
                     Some((SEND_STATE, filter_s)) => {
-                        Ok(Permission::Send(EventFilter::State(parse_state_event_filter(filter_s))))
+                        Ok(Permission::Send(Filter::State(parse_state_event_filter(filter_s))))
                     }
                     _ => {
                         debug!("Unknown capability `{s}`");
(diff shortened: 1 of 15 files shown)


Hywan:
Hello,

The PR title contains WIP but a review is requested. The PR description contains almost nothing. Please, move the PR as a draft. Ping me when it's ready, I'll be delighted to review it then.

crates/matrix-sdk/src/widget/matrix.rs
```diff
@@ -204,19 +222,170 @@ impl MatrixDriver {
...
+                        device_or_all_id,
+                        &user_id,
+                        &event_type,
+                    )
+                    .await
+                    .map_err(|e| info!("could not encrypt content for to device widget event content: {}. because, {}", content.json(), e))
+                    .ok()
+                }}));
+            // The first flatten takes the iterator of Option<Device>'s iterators and converts it to just a iterator over Option<(Device, data)>'s.
+            // The second flatten takes the the iterator over Option<(Device,data)>'s and converts it to just a iterator over Device
```
Hywan: So the “data” are computed and dropped?

crates/matrix-sdk/src/widget/matrix.rs
```diff
@@ -204,19 +222,170 @@ impl MatrixDriver {
...
+            join_all(device_map_futures).await.into_iter().flatten().flatten().flatten().collect()
+        }
+
+        // We first want to get all missing session before we start any to device sending!
+        client.claim_one_time_keys(messages.iter().map(|(u, _)| u.as_ref())).await?;
+
+        let request = match encrypted {
+            true => {
```
Hywan, line 353:
```diff
let request = if encrypted {
    …
} else {
    …
};
```

crates/matrix-sdk/src/widget/matrix.rs
```diff
@@ -204,19 +222,170 @@ impl MatrixDriver {
...
+                    encrypted_content,
+                )
+            }
+            false => RumaToDeviceRequest::new_raw(event_type, TransactionId::new(), messages),
+        };
+
+        let response = client.send(request).await;
+        // if let Ok(res){
+        //     client.mark_request_as_sent(request.request_id(), &res).await;
+        // };
```
Hywan, line 391: Clean up.

crates/matrix-sdk/src/widget/matrix.rs
```diff
@@ -204,19 +222,170 @@ impl MatrixDriver {
...
+        response.map_err(Into::into)
     }
 }
 
 /// A simple entity that wraps an `UnboundedReceiver`
 /// along with the drop guard for the room event handler.
-pub(crate) struct EventReceiver {
-    rx: UnboundedReceiver<Raw<AnyTimelineEvent>>,
-    _drop_guards: [EventHandlerDropGuard; 2],
+
```
Hywan: Remove the empty line.

crates/matrix-sdk/src/widget/mod.rs
```diff
@@ -269,16 +291,42 @@ impl WidgetDriver {
...
+                if self.to_device_event_forwarding_guard.is_none() {
+                    let (stop_forwarding, guard) = {
+                        let token = CancellationToken::new();
+                        (token.child_token(), token.drop_guard())
+                    };
+
+                    self.to_device_event_forwarding_guard = Some(guard);
+                    let (mut matrix, events_tx) =
+                        (matrix_driver.to_device_events(), incoming_msg_tx.clone());
+                    tokio::spawn(async move {
```
Hywan: Please use `matrix_sdk_common::executor::spawn` instead.

crates/matrix-sdk/src/widget/capabilities.rs
```diff
@@ -62,39 +65,37 @@ pub struct Capabilities {
...
-                    Err(err) => {
-                        error!("Failed to deserialize raw event as MatrixEventFilterInput: {err}");
-                        return false;
-                    }
-                }
+    pub fn raw_timeline_event_matches_read_filter(&self, raw: &Raw<AnyTimelineEvent>) -> bool {
+        let filter_in = match raw.deserialize_as::<MatrixEventFilterInputData>() {
+            Ok(filter) => MatrixEventFilterInput::Timeline(filter),
+            Err(err) => {
+                error!("Failed to deserialize raw event as MatrixEventFilterInput: {err}");
```
Hywan: Should be rebased onto the first commit.

crates/matrix-sdk/src/widget/capabilities.rs
```diff
@@ -62,39 +65,37 @@ pub struct Capabilities {
...
-const RECEIVE_STATE: &str = "org.matrix.msc2762.receive.state_event";
-const SEND_TODEVICE: &str = "org.matrix.msc3819.send.to_device";
-const RECEIVE_TODEVICE: &str = "org.matrix.msc3819.receive.to_device";
-const REQUIRES_CLIENT: &str = "io.element.requires_client";
+pub(super) const SEND_EVENT: &str = "org.matrix.msc2762.send.event";
+pub(super) const RECEIVE_EVENT: &str = "org.matrix.msc2762.receive.event";
+pub(super) const SEND_STATE: &str = "org.matrix.msc2762.send.state_event";
+pub(super) const RECEIVE_STATE: &str = "org.matrix.msc2762.receive.state_event";
+pub(super) const SEND_TODEVICE: &str = "org.matrix.msc3819.send.to_device";
+pub(super) const RECEIVE_TODEVICE: &str = "org.matrix.msc3819.receive.to_device";
```
Hywan, line 75: Same here and everywhere else.

Hywan (changes requested):
Thanks for your work, but I don't think we are there yet. This is hard to review.

I did review up to the 5th patches.

Please, rewrite the history of this pull request. The first commit is super large and, because it has no description, it's quite hard to know what it does. The code is not completely documented. The code is not formatted. Some parts of the code are marked as dead. Some parts of the code are commented.

Maybe the next commits fix that up. In this case, these commits should be merged together.

Also please, add commit descriptions! A good clue is: if a commit does too many things, it means the commit must be split into smaller ones. Also, if providing a description for a commit is complicated, then imagine how hard it is for the reviewer to check the diff.

crates/matrix-sdk/src/widget/mod.rs
toger5: We need to be careful here for wasm since it will drop the handle and consequently stop the async process.
So this would need a `forget` call on the handle or keeping a reference to the handle while the process should be running (tokio and the wasm spawn behave different in that regard)

crates/matrix-sdk/src/widget/machine/from_widget.rs
```diff
@@ -237,3 +240,16 @@ impl From<update_delayed_event::unstable::Response> for UpdateDelayedEventRespon
         Self {}
     }
 }
+
+/// A wrapper type for the empty okay response from
+/// [`update_delayed_event`](update_delayed_event::unstable::Response)
+/// which derives Serialize. (The response struct from Ruma does not derive
+/// serialize)
+#[derive(Serialize, Debug)]
+pub(crate) struct SendToDeviceEventResponse {}
```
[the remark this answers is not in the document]
Hywan: Can't we update the tests? What's the impact of changing that?
If you can't change, okay, but please indicate this information in the documentation.

crates/matrix-sdk/src/widget/mod.rs
Hywan: So please update `matrix_sdk_common::executor::spawn` to do what you need, but don't use `tokio::spawn` directly here.

Hywan (changes requested):
I'm still struggling to review the essence of the PR because there is too much noise to fix beforehand.

Your patches are still a bit big but I can live with it _if_ you add a description to each commit explaining what it does and why. Thanks :-).

crates/matrix-sdk/src/widget/filter.rs
```diff
@@ -103,271 +109,382 @@ pub enum StateEventFilter {
...
 }
 
+impl fmt::Display for ToDeviceEventFilter {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "{0}", self.event_type)
+    }
+}
+
+// Filter input:
+
```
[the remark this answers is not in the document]
toger5: Seperates the following section which does not implement filters anymore but Filter inputs.
Is there a common style how such seperator is done (or is that considered unhelpful?)

crates/matrix-sdk/src/widget/filter.rs
```diff
@@ -103,271 +109,382 @@ pub enum StateEventFilter {
...
+    }
+}
+
+// Filter input state:
 #[derive(Debug, Deserialize)]
-pub(super) struct MatrixEventFilterInput {
+pub struct FilterInputState {
     #[serde(rename = "type")]
-    pub(super) event_type: TimelineEventType,
+    pub(super) event_type: String,
```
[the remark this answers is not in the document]
toger5: This allow to have a seperate `FilterInputState` vs `FilterInputMessageLike` so that we do not have the single data type that contained state key and msgtype.
(the distinction between state and messageLike was done via event_type before which is easy to misconfigure when using `into()` especially with unknown even types from widgets. Now the event_type is a string and the distinction is done explicitly by Filter::...)

crates/matrix-sdk/src/widget/filter.rs
```diff
@@ -103,271 +109,382 @@ pub enum StateEventFilter {
...
-            event_type,
-            state_key: None,
-            content: MatrixEventContent { msgtype: Some(msgtype) },
+/// Create a filter input of type FilterInput::ToDevice.
+impl TryFrom<Raw<AnyToDeviceEvent>> for FilterInput {
+    type Error = serde_json::Error;
+    fn try_from(raw_event: Raw<AnyToDeviceEvent>) -> Result<Self, Self::Error> {
+        // make sure Raw actually was a message like event
+        raw_event.deserialize()?;
+        raw_event.deserialize_as::<FilterInputToDevice>().map(FilterInput::ToDevice)
```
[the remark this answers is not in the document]
toger5: Here this first one is actually unnecassary. A to device message is already valid with only a `type` field so the second implies that the first is working.

crates/matrix-sdk/src/widget/matrix.rs
```diff
@@ -204,19 +212,194 @@ impl MatrixDriver {
...
+                if let Some(content) = event_with_encryption_flag.get_mut("content") {
+                    content["encrypted"] = json!(encryption_info.is_some());
+                }
+                let ev_for_widget = match Raw::<AnyToDeviceEvent>::from_json_string(
+                    event_with_encryption_flag.to_string(),
+                ) {
+                    Ok(ev) => ev,
+                    Err(_) => raw,
+                };
```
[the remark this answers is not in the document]
toger5: I ended up refactoring this to be equivalent to what we do for adding the `room_id`
This has the advantage that we deserialize as: `BTreeMap<String, Box<RawJsonValue>` which should be quicker considering `RawJsonValue` is a trivial deserialization.

crates/matrix-sdk/src/widget/filter.rs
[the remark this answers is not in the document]
toger5: I updated it now to be: The correct event type even though its already definde by the Filter enum.
`StateEventType`, `ToDeviceEventType`, `MessageLikeEventType`

toger5:
@Hywan I addressed the more obvious changes and added more details to the larger commits: https://github.com/matrix-org/matrix-rust-sdk/pull/4856/commits/f091aac2961e6c470e26e3e2ccc55f40359a62d2

toger5:
Just `rebased` on main

crates/matrix-sdk/src/widget/matrix.rs
```diff
@@ -208,25 +235,214 @@ impl MatrixDriver {
...
+            let content: Value =
+                unencrypted_content.deserialize_as().map_err(Into::<Error>::into)?;
+            let event_type = event_type.clone();
+            let device_content_tasks = devices.into_iter().map(|device| spawn({
+                let event_type = event_type.clone();
+                let content = content.clone();
+
+                async move {
+                    if !device.is_cross_signed_by_owner() {
+                        info!("Device {} is not verified, skipping encryption", device.device_id());
```
BillCarsonFr, line 294: `verified` is not the correct term, the device is insecure.
There are more checks to do here, I wonder if we should manage that in `encrypt_event_raw` with some specific setting.

crates/matrix-sdk/src/widget/matrix.rs
```diff
@@ -208,25 +235,214 @@ impl MatrixDriver {
...
+            event_type: &ToDeviceEventType,
+            device_content_map: BTreeMap<DeviceIdOrAllDevices, Raw<AnyToDeviceEventContent>>,
+        ) -> BTreeMap<DeviceIdOrAllDevices, Raw<AnyToDeviceEventContent>> {
+            let device_map_futures =
+                device_content_map.into_iter().map(|(device_or_all_id, content)| spawn({
+                    let client = client.clone();
+                    let user_id = user_id.to_owned();
+                    let event_type = event_type.clone();
+                    async move {
+                        let Ok(user_devices) = client.encryption().get_user_devices(&user_id).await else {
```
BillCarsonFr, line 327: We shuold probably handle that in the crypto crate as per https://github.com/matrix-org/matrix-rust-sdk/issues/4147

crates/matrix-sdk/src/widget/matrix.rs
```diff
@@ -208,25 +235,214 @@ impl MatrixDriver {
...
+            // and flattens the Option (drops None items)
+            // The third takes the iterator over iterators: impl Iterator<Item =
+            // (DeviceIdOrAllDevices, Raw<AnyToDeviceEventContent>)>
+            // and flattens it to just an iterator over (DeviceIdOrAllDevices,
+            // Raw<AnyToDeviceEventContent>)
+            content_map_iterator.flatten().flatten().flatten().collect()
+        }
+
+        let request = if encrypted {
+            // We first want to get all missing session before we start any to device
```
BillCarsonFr, line 370: actually you should do that before encrypting.

crates/matrix-sdk/src/widget/matrix.rs
```diff
@@ -208,25 +235,214 @@ impl MatrixDriver {
...
+    match raw.deserialize_as::<BTreeMap<String, Box<RawJsonValue>>>() {
+        Ok(mut ev_mut) => {
+            ev_mut.insert("room_id".to_owned(), serde_json::value::to_raw_value(room_id)?);
+            Ok(Raw::new(&ev_mut)?.cast())
+        }
+        Err(e) => Err(Error::from(e)),
+    }
+}
+
+fn with_attached_encryption_flag(
```
BillCarsonFr, line 436: why are we doing that?

crates/matrix-sdk/tests/integration/widget.rs
```diff
@@ -417,32 +419,63 @@ async fn test_receive_live_events() {
...
+                    "content": {
+                      "a": "test",
+                    }
+                  }
+            ));
         })
         .await;
 
-    let msg = recv_message(&driver_handle).await;
-    assert_eq!(msg["api"], "toWidget");
```
BillCarsonFr, line 424: Sounds like something for insta:assert_json_snapshot ?

Hywan (changes requested):

toger5:
Lets close this since we have a plan how to restructure this into new PR's.

PR #4858, opened 2025-03-28
integration-test: Remove postgres container

richvdh:
Followup on https://github.com/matrix-org/matrix-rust-sdk/pull/3983: now that we don't have a sliding sync proxy, we don't need a postgres container.


stefanceriu (approved):
Noice 👌

PR #4860, opened 2025-03-28
chore: move `RedactedMessage` and `UnableToDecrypt` to `MsgLike`

stefanceriu:
This patch moves `RedactedMessage` and `UnableToDecrypt` to the newly introduced `TimelineItemContent::MsgLike` variant to bring consistency with Ruma and have them provide aggregations if necessary (e.g. UTD on a thread).


andybalaam (approved):
Thanks for the nice small commits!

PR #4862, opened 2025-03-28
refactor(encryption): Simplify the parsing of OAuthCrossSigningResetInfo

poljar:
Now that OAuth isn't behind a feature flag a bunch of things can be simplified.

This was also tested with a  real server.


bnjbvr (approved):
Noice!

PR #4866, opened 2025-03-31
feat(ffi): expose `report_room` in Client

Velin92:
This API will make things a lot easier in cases in which we want to report the room, but we don't have access to the `Room` itself, but only to its identifier.


bnjbvr (commented):
Thanks! So as to avoid duplicating functionality in the SDK, and now that we've exposed `Client::get_room` in the FFI layer: can you use, in the embedding, `client.getRoom(room_id).reportRoom()`?

bnjbvr:
Sorry, there's been a misunderstanding: my point was rather that there's nothing to add to the SDK; if you have a `Client` instance in the swift code, then you could use the swift methods to retrieve the room by id, then report it.

Velin92:
> Thanks! So as to avoid duplicating functionality in the SDK, and now that we've exposed `Client::get_room` in the FFI layer: can you use, in the embedding, `client.getRoom(room_id).reportRoom()`?

Oh okay now I understand. So I'll explain the problem, the issue is that for room in which you have not joined, like room invites, the `full_room` and the `invited_room` should not be used in favour of the `room_preview` APIs. However this API lacks a report room action, and reporting an offensive invite is quite a big use case.

bnjbvr:
We now have a way to get a room via ID, independently of the `RoomPreview` API (and the other methods you're mentioned here). For instance, this method is mapped by uniffi and exposed [here](https://github.com/vector-im/element-x-ios/blob/36903bf3fbb270775822ea790b26df14278c14ce/ElementX/Sources/Mocks/Generated/SDKGeneratedMocks.swift#L1621) in EXI code. With this method, you ought to be able to get a room via its room id, and then call `reportRoom()` on it, maybe?

jmartinesp:
> Oh okay now I understand. So I'll explain the problem, the issue is that for room in which you have not joined, like room invites, the `full_room` and the `invited_room` should not be used in favour of the `room_preview` APIs

To expand a bit on @bnjbvr 's response, we actually talked about this with @stefanceriu and him and IIRC we agreed on removing the `RoomPreview` object from the FFI layer at some point and instead use only the `RoomPreviewInfo` contents + `Client::get_room` to access any needed methods now that we don't need an actual timeline to call that one. Sorry, we should have made sure this was known by all team members 😓 .

Velin92:
> > Oh okay now I understand. So I'll explain the problem, the issue is that for room in which you have not joined, like room invites, the `full_room` and the `invited_room` should not be used in favour of the `room_preview` APIs
> 
> To expand a bit on @bnjbvr 's response, we actually talked about this with @stefanceriu and him and IIRC we agreed on removing the `RoomPreview` object from the FFI layer at some point and instead use only the `RoomPreviewInfo` contents + `Client::get_room` to access any needed methods now that we don't need an actual timeline to call that method. Sorry, we should have made sure this was known by all team members 😓 .

Okay I think that's okay then, but the `Room` from the `get_room` comes from the `state_store` right?
For now this would be sufficient but if in the future we want to allow reporting a public room you have just seen in the public room directory that contains offensive content (like an avatar or just the name) and have not joined or been invited to, I guess that would not be possible.

jmartinesp:
> if in the future we want to allow reporting a public room you have just seen in the public room directory that contains offensive content (like an avatar or just the name) and have not joined or been invited to, I guess that would not be possible right?

True, for that case maybe we'd need some `Client::report_room_by_id` method.

Velin92:
> > if in the future we want to allow reporting a public room you have just seen in the public room directory that contains offensive content (like an avatar or just the name) and have not joined or been invited to, I guess that would not be possible right?
> 
> True, for that case maybe we'd need some `Client::report_room_by_id` method.

Well as I said this out of scope for now, so I guess we can use the proposed solution, and then maybe in the future deal with a `Client::report_room_by_id` method.

PR #4879, opened 2025-04-02
refactor(oauth): Remove `OAuthRegistrationStore` and `issuer`

zecakeh:
MSC2966 was updated pre-FCP (https://github.com/matrix-org/matrix-spec-proposals/pull/2966#discussion_r2013883503), clients should re-register for every login, so we don't need to store the client IDs between logins.


zecakeh:
I added another commit that also removes the `issuer` that we were holding in `OAuthAuthData`. It was mostly unused for a while, and now that users can just use the homeserver URL for static registrations, it's not necessary to have the `OAuth::issuer()` helper method.

stefanceriu (approved):
Great work, thank you very much! 🙏

PR #4880, opened 2025-04-02
feat!(ffi): merge send_reply and send_thread_reply

Johennes:
This is a small follow-up on https://github.com/matrix-org/matrix-rust-sdk/pull/4852/files#r2016594024.

### DIFF
--- a/bindings/matrix-sdk-ffi/src/timeline/mod.rs
+++ b/bindings/matrix-sdk-ffi/src/timeline/mod.rs
@@ -116,31 +116,13 @@ impl Timeline {
             params.formatted_caption.map(Into::into),
         );
 
-        let reply = if let Some(reply_params) = params.reply_params {
-            let event_id = EventId::parse(reply_params.event_id)
-                .map_err(|_| RoomError::InvalidRepliedToEventId)?;
-            let enforce_thread = if reply_params.enforce_thread {
-                EnforceThread::Threaded(if reply_params.reply_within_thread {
-                    ReplyWithinThread::Yes
-                } else {
-                    ReplyWithinThread::No
-                })
-            } else {
-                EnforceThread::MaybeThreaded
-            };
-
-            Some(Reply { event_id, enforce_thread })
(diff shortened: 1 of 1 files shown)


bindings/matrix-sdk-ffi/src/timeline/mod.rs
```diff
@@ -502,45 +502,11 @@ impl Timeline {
...
-            .send_reply(
-                (*msg).clone(),
-                event_id,
-                EnforceThread::Threaded(if is_reply {
-                    ReplyWithinThread::Yes
-                } else {
-                    ReplyWithinThread::No
-                }),
-            )
+            .send_reply((*msg).clone(), reply.event_id, reply.enforce_thread)
```
Johennes, line 509: This could maybe use a single parameter of type `Reply` rather than separate parameters for the event ID and threading behaviour. The same applies to a few other methods down the call tree and would probably require moving `Reply` from `attachment.rs` into `repply.rs`.

Happy to make those changes if that seems reasonable.

bindings/matrix-sdk-ffi/src/timeline/mod.rs
```diff
@@ -241,6 +223,24 @@ pub struct ReplyParameters {
     reply_within_thread: bool,
 }
 
+impl ReplyParameters {
+    fn to_reply(&self) -> Result<Reply, RoomError> {
```
bnjbvr, line 227: Any chance we can make use of a `TryInto` impl here?

bnjbvr (commented):
yay, thanks! I think pushing down `Reply` into the SDK makes sense :+1:

bindings/matrix-sdk-ffi/src/timeline/mod.rs
Johennes: Ohhh, yes of course! 🤯🤦‍♂️

bnjbvr (approved):
Thanks, lgtm!

PR #4881, opened 2025-04-03
fix(ffi): correctly populate all audio content fields when converting from FFI types to Ruma

stefanceriu:
This patch fixes forwarding audio and voice messages that would previously show up as files because of missing fields.

Props go out to @jmartinesp for finding what the actual problem was 🍻

### DIFF
--- a/bindings/matrix-sdk-ffi/src/ruma.rs
+++ b/bindings/matrix-sdk-ffi/src/ruma.rs
@@ -34,7 +34,7 @@ use ruma::{
                 MessageType as RumaMessageType,
                 NoticeMessageEventContent as RumaNoticeMessageEventContent,
                 RoomMessageEventContentWithoutRelation,
-                TextMessageEventContent as RumaTextMessageEventContent,
+                TextMessageEventContent as RumaTextMessageEventContent, UnstableAmplitude,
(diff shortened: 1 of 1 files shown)


bmarty (commented):
Not tested, by I believe that this will work according to the diff of Event source that I observe between a voice message and a forwarded voice message.

Nice!

bindings/matrix-sdk-ffi/src/ruma.rs
```diff
@@ -362,6 +362,15 @@ impl TryFrom<MessageType> for RumaMessageType {
                         .info(content.info.map(Into::into).map(Box::new));
                 event_content.formatted = content.formatted_caption.map(Into::into);
                 event_content.filename = filename;
+
+                if let Some(audio) = content.audio {
+                    event_content.audio = Some(audio.into());
+                }
```
poljar, line 368: Can't you just assign that always, something like:

```rust
event_content.audio = content.audio.map(Into::into)
```

poljar (approved):
Left a suggestion, makes sense.

bindings/matrix-sdk-ffi/src/ruma.rs
stefanceriu: Oh right, Jorge told me something about mapping but I just blanked out 😅

PR #4882, opened 2025-04-03
Clean up OAuth docs

zecakeh:
There are two commits here:

1. Make sure that the docs of the `oauth` module are up-to-date and mention all the login APIs.
2. Try to clean up the changelog: there were a lot of changes around the Oidc/OAuth API this cycle, and it was hard to follow, especially with methods and types that changed and were then removed.

    This groups everything under a single entry, with a short summary of the changes, followed by a list of per-PR changes.

    The detailed list is reorganized to put the biggest changes first, like the renaming, and a few entries were cleaned up or removed, because they mention a method or type that is removed in another entry.

### DIFF
--- a/crates/matrix-sdk/src/authentication/oauth/mod.rs
+++ b/crates/matrix-sdk/src/authentication/oauth/mod.rs
@@ -27,53 +27,76 @@
 //!
 //! # Registration
 //!
-//! Registration is only required the first time a client encounters a
-//! homeserver.
+//! Clients must register with the homeserver before being able to interact with
+//! an OAuth 2.0 server.
+//!
+//! The registration consists in providing client metadata to the authorization
+//! server, to declare the interactions that the client supports with the
+//! homeserver. This step is important because the client cannot use a feature
+//! that is not declared during registration. In return, the server assigns an
+//! ID and eventually credentials to the client, which will allow to identify
+//! the client when authorization requests are made.
 //!
 //! Note that only public clients are supported by this API, i.e. clients
 //! without credentials.
 //!
-//! If the server supports dynamic registration, it can be done by using
-//! [`OAuth::register_client()`]. After registration, the client ID should be
-//! persisted and reused for every session that interacts with that same server.
+//! The registration step can be done automatically by providing a
+//! [`ClientRegistrationData`] to the login method.
 //!
-//! If dynamic registration is not available, the homeserver should document how
-//! to obtain a client ID.
-//!
-//! To provide the client ID and metadata if dynamic registration is not
-//! available, or if the client is already registered with the issuer, call
-//! [`OAuth::restore_registered_client()`].
+//! If the server supports dynamic registration, registration can be performed
+//! manually by using [`OAuth::register_client()`]. If dynamic registration is
+//! not available, the homeserver should document how to obtain a client ID. The
+//! client ID can then be provided with [`OAuth::restore_registered_client()`].
 //!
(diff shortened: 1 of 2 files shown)


crates/matrix-sdk/src/authentication/oauth/mod.rs
```diff
@@ -27,53 +27,63 @@
 //!
 //! # Registration
 //!
-//! Registration is only required the first time a client encounters a
-//! homeserver.
+//! Clients must register with the homeserver before
```
poljar, line 30: I think we should explain what the hell a registration achieves, especially since this is very much not the same as registering an user.

The dynamic registration [RFC](https://datatracker.ietf.org/doc/html/rfc7591), which we might link to as well, has a nice paragraph about this:

>  This specification defines mechanisms for dynamically registering
   OAuth 2.0 clients with authorization servers.  Registration requests
   send a set of desired client metadata values to the authorization
   server.  The resulting registration responses return a client
   identifier to use at the authorization server and the client metadata
   values registered for the client.  The client can then use this
   registration information to communicate with the authorization server
   using the OAuth 2.0 protocol.  This specification also defines a set
   of common client metadata fields and values for clients to use during
   registration.

Perhaps we can quote this verbatim, or maybe shorten and adapt it a bit.

Either way works for me.

crates/matrix-sdk/src/authentication/oauth/mod.rs
```diff
@@ -27,53 +27,63 @@
...
+//! provided redirect URI, with a code in the query that will allow to finish
+//! the login process by calling [`OAuth::finish_login()`].
+//!
+//! If the login needs to be cancelled before its completion,
+//! [`OAuth::abort_login()`] should be called to clean up the local data.
+//!
+//! ## Login by scanning a QR Code
+//!
+//! This method requires to have another logged-in Matrix device that can
+//! display a QR Code. This is a method that uses the Device Code flow
```
poljar, line 67: Would be nice to link to what the `Device Code flow` is as well.

crates/matrix-sdk/src/authentication/oauth/mod.rs
```diff
@@ -873,7 +892,11 @@ impl OAuth {
         )
     }
 
-    /// Login via OAuth 2.0 with the Authorization Code flow.
+    /// Log in via OAuth 2.0 with the Authorization Code flow.
```
poljar, line 895: Can we perhaps link to the section in the module level docs where we explain what the Authorization Code flow is?

poljar (changes requested):
Thanks for updating the docs, very nice improvements. I left some nits that might improve things even further, especially if somebody that isn't familiar with the OAuth lingo comes around.

crates/matrix-sdk/src/authentication/oauth/mod.rs
zecakeh: I added links to the MSC and RFC.

crates/matrix-sdk/src/authentication/oauth/mod.rs
zecakeh: I am not sure how to do that except by cheating with a relative URL. I copied the paragraph that explains what the flow does.

poljar:
Could you just get rid of the self-inflicted merge conflict :sweat_smile:

zecakeh:
Rebased on main to solve the conflict.

PR #4885, opened 2025-04-04
chore(cargo): Update `rand` to 0.9

Hywan:
This patch updates `rand` to 0.9. It brings several nice advantages:

- it is compatible with the new Rust edition 2024,
- it simplifies several signatures since `try_fill` is replaced by
  `fill`, there is no failure.

---

* Based on https://github.com/matrix-org/matrix-rust-sdk/pull/4884


poljar:
I'm afraid that this will have to wait for https://github.com/matrix-org/vodozemac/pull/206 and its deps.

Hywan:
Why?

`rand` exists in 2 versions in the deps tree with this PR, but I think waiting on vodozemac isn't blocking.

Hywan:
We have 2 versions of `getrandom`, the 0.3.2 from our crates, and 0.2.15 from:

- `chacha20poly1305` which still requires `rand` 0.6.4,
- `backoff`, which still requires `rand` 0.8.5,
- `vodozemac` which still requires `rand` 0.6.4,
- `ctr`, same,
- `hkdf`, same,
- `indexed_db_futures`,

We have to move many crates…

poljar:
Let's close this for now. We're going to collect merge conflicts for quite a bit here otherwise.

PR #4886, opened 2025-04-04
feat(sdk): Add `Client::logout()` to log out regardless of the auth API

zecakeh:
It simplifies code for users, and avoids to have to match on `AuthApi`, which is a non-exhaustive enum.


poljar (approved):
That's indeed a nice simplification.

PR #4887, opened 2025-04-04
Reduce the sizes of various types and mostly `matrix_sdk::Error`

Hywan:
This PR is a collection of small patches that reduce the sizes of various types. A particular focus has been made on `matrix_sdk::Error` where its size has been reduced by 6.5x (from 160 bytes to 24 bytes). This is important as it impacts all the methods and functions that return a `Result`.


poljar (approved):
Neat.

PR #4897, opened 2025-04-07
refactor(send_queue): generalize SentRequestKey::Media and DependentQueuedRequestKind::UploadFileWithThumbnail to prepare for MSC4274 gallery uploads

Johennes:
This was broken out of https://github.com/matrix-org/matrix-rust-sdk/pull/4838 and is a preliminary step towards implementing [MSC4274](https://github.com/matrix-org/matrix-spec-proposals/pull/4274). `SentRequestKey::Media` and `DependentQueuedRequestKind::UploadFileWithThumbnail` are generalized to allow chaining dependent media uploads and accumulating sent media sources.


crates/matrix-sdk/src/send_queue/upload.rs
```diff
@@ -348,36 +350,58 @@ impl QueueStorage {
...
-        // The media we just uploaded was a thumbnail, so the thumbnail shouldn't have
+        // If the previous upload was a thumbnail, it shouldn't have
         // a thumbnail itself.
-        debug_assert!(sent_media.thumbnail.is_none());
-        if sent_media.thumbnail.is_some() {
-            warn!("unexpected thumbnail for a thumbnail!");
+        if depends_on_thumbnail {
+            debug_assert!(sent_media.thumbnail.is_none());
+            if sent_media.thumbnail.is_some() {
+                warn!("unexpected thumbnail for a thumbnail!");
```
Hywan, line 375: Oh good, there is a check for that, nice.

crates/matrix-sdk/src/send_queue/upload.rs
```diff
@@ -348,36 +350,58 @@ impl QueueStorage {
...
+            }
         }
 
-        trace!(related_to = %event_txn, "done uploading thumbnail, now queuing a request to send the media file itself");
+        trace!(related_to = %event_txn, "done uploading file or thumbnail, now queuing the dependent file or thumbnail upload request");
+
+        // If the previous upload was a thumbnail, forward the accumulated sent media
+        // unchangend. Otherwise, append the sent media to the accumulated
```
Hywan, line 382: unchangend?

crates/matrix-sdk/src/send_queue/upload.rs
```diff
@@ -348,36 +350,58 @@ impl QueueStorage {
...
+            });
+            accumulated
+        };
 
         let request = QueuedRequestKind::MediaUpload {
             content_type,
             cache_key,
-            // The thumbnail for the next upload is the file we just uploaded here.
-            thumbnail_source: Some(sent_media.file),
+            // If the previous upload was a thumbnail, it becomse the thumbnail source for the next
```
Hywan, line 399: becomes*

Hywan (changes requested):
Thanks for the patch.
I believe it would be lovely to add a couple of tests just to ensure the `accumulated` field correctly accumulates the media info. Is it doable?

crates/matrix-sdk-base/Cargo.toml
```diff
@@ -43,6 +43,9 @@ testing = [
     "matrix-sdk-crypto?/testing",
 ]
 
+# Add support for inline media galleries via msgtypes
+unstable-msc4274 = []
```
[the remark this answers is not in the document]
Johennes: Technically yes but this PR doesn't actually use any of the MSC4274 code in Ruma yet. Since that also requires updating the SHA that ruma is pinned to, I thought it would be better to only do it in a later PR when I'm actually starting to use things from ruma. Does that makes sense?

crates/matrix-sdk/src/send_queue/upload.rs
Johennes: Yeah, I couldn't think of a better way to do this but it is essentially the same issue in the current code already.

crates/matrix-sdk/src/send_queue/upload.rs
Johennes: In the current code for media the order is always thumbnail first and then the file as a dependent request on it. For galleries, I'm planning to it the same way so you'll have:

thumbnail_1 ← file_1 ← thumbnail_2 ← file_2 ← ...

Or if there are no thumbnails, it'd be:

file_1 ← file_2 ← ...

The `accumulated` vector is meant to only contain completed uploads of both thumbnail (if it exists) and file for a single media.

So if we get here and the depended upon request is a thumbnail, we know that the file upload is still missing and, therefore, we only pass `accumulated` through unchanged. If the depended upon request is a file, however, we know that this file (and possibly its thumbnail) have uploaded and we push it onto `accumulated`.

We could maybe also store uploaded thumbnails into `accumulated` directly but then `AccumulatedSentMediaInfo::file` would have to become optional which might then complicate things elsewhere.

Johennes:
> I believe it would be lovely to add a couple of tests just to ensure the accumulated field correctly accumulates the media info. Is it doable?

It is slightly tricky because I think currently it can only be tested indirectly through the final event that is being sent. For galleries the code to send the event is still missing, however. I was planning to add it with a separate PR, including tests. I realize this isn't ideal but I'm not sure how to best test it in this PR here. 😕

crates/matrix-sdk-base/Cargo.toml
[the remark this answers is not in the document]
Hywan: It does, thanks.

crates/matrix-sdk/src/send_queue/upload.rs
Johennes: Ohhh! Looks like I can neither write nor read. 🤦‍♂️ Sorry, I totally misunderstood. Typo is fixed now.

bnjbvr:
Yep, thanks for the ping.

crates/matrix-sdk-base/src/store/send_queue.rs
```diff
@@ -219,17 +225,22 @@ pub enum DependentQueuedRequestKind {
         key: String,
     },
 
-    /// Upload a file that had a thumbnail.
-    UploadFileWithThumbnail {
-        /// Content type for the file itself (not the thumbnail).
+    /// Upload a file or thumbnail depending on another file or thumbnail
+    /// upload.
+    UploadFileOrThumbnail {
```
bnjbvr, line 230: I think the renaming of this field is a data semantic breaking change: at least the SQLite backend serializes the `DependentQueuedRequestKind` as JSON, and depending on the strategy used for (de)serializing an enum, this might be a change in the data format.

We need to figure out whether that's the case: one way would be to write a test case with content serialized with the previous format (so an enum with only that variant, called `UploadFileWithThumbnail` is used and saved into the database), then deserialized with the new format. If it deserializes without trouble (which I doubt here), then we need to take further steps.

Further steps may include:

- using `serde` magic to allow for the field to have a different previous name, or a different name in the serialized format; that's fine, since we mostly care about the Rust name here.
- running a data migration for each store.
- or bailing out from running a data migration, and clear all the pending dependent requests in all send queues. I would like to avoid this.

crates/matrix-sdk-base/src/store/send_queue.rs
```diff
@@ -324,6 +335,29 @@ pub struct SentMediaInfo {
     ///
     /// When uploading a thumbnail, this is set to `None`.
     pub thumbnail: Option<MediaSource>,
+
+    /// Accumulated list of infos for previously uploaded files and thumbnails
+    /// if used during a gallery transaction. Otherwise empty.
+    #[cfg(feature = "unstable-msc4274")]
+    #[serde(default)]
+    pub accumulated: Vec<AccumulatedSentMediaInfo>,
```
bnjbvr, line 343: There might be a different possible data scheme here, if I understand correctly:

- `SentMediaInfo` contains *only* the vec of `AccumulatedSentMediaInfo` (since a vec can hold the pair for the media, when there's only a single one and no gallery)
- `AccumulatedSentMediaInfo` (maybe rename `SingleSentMediaInfo`?) can keep on holding its current fields.

Again, I see this struct `SentMediaInfo` is also marked as `Serialized/Deserialize`, so maybe what I'm suggesting implies having to a data migration :woozy_face:.

crates/matrix-sdk-base/src/store/send_queue.rs
```diff
@@ -322,6 +333,29 @@ pub struct SentMediaInfo {
...
+    /// File that was uploaded by this request.
+    ///
+    /// If the request related to a thumbnail upload, this contains the
+    /// thumbnail media source.
+    pub file: MediaSource,
+
+    /// Optional thumbnail previously uploaded, when uploading a file.
+    ///
+    /// When uploading a thumbnail, this is set to `None`.
+    pub thumbnail: Option<MediaSource>,
```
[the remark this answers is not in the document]
bnjbvr: I don't think so :)

crates/matrix-sdk/src/send_queue/upload.rs
```diff
@@ -348,36 +350,58 @@ impl QueueStorage {
...
+        trace!(related_to = %event_txn, "done uploading file or thumbnail, now queuing the dependent file or thumbnail upload request");
+
+        // If the previous upload was a thumbnail, forward the accumulated sent media
+        // unchanged. Otherwise, append the sent media to the accumulated
+        // vector.
+        #[cfg(feature = "unstable-msc4274")]
+        let accumulated = if depends_on_thumbnail {
+            sent_media.accumulated
+        } else {
+            let mut accumulated = sent_media.accumulated.clone();
```
bnjbvr, line 388: Is the clone required here?

crates/matrix-sdk/src/send_queue/mod.rs
```diff
@@ -1589,22 +1599,28 @@ impl QueueStorage {
...
                     client,
                     dependent_request.own_transaction_id.into(),
                     parent_key,
                     content_type,
                     cache_key,
                     related_to,
+                    #[cfg(feature = "unstable-msc4274")]
+                    depends_on_thumbnail,
+                    #[cfg(not(feature = "unstable-msc4274"))]
+                    true,
```
bnjbvr, line 1623: To make it easier to read, can you define a local variable `depends_on_thumbnail` and define it accordingly? (maybe with `cfg_if`)

crates/matrix-sdk/src/send_queue/upload.rs
```diff
@@ -348,36 +350,58 @@ impl QueueStorage {
...
+                warn!("unexpected thumbnail for a thumbnail!");
+            }
         }
 
-        trace!(related_to = %event_txn, "done uploading thumbnail, now queuing a request to send the media file itself");
+        trace!(related_to = %event_txn, "done uploading file or thumbnail, now queuing the dependent file or thumbnail upload request");
+
+        // If the previous upload was a thumbnail, forward the accumulated sent media
+        // unchanged. Otherwise, append the sent media to the accumulated
+        // vector.
```
bnjbvr, line 383: Can we reword this like `If the parent previous request was a thumbnail upload, don't add it to the list of accumulated medias, since it is included elsewhere.`?

(I wonder if renaming `depends_on_thumbnail` to `parent_is_thumbnail_upload` would clarify things, or just be too wordy?)

crates/matrix-sdk/src/send_queue/upload.rs
```diff
@@ -348,36 +350,58 @@ impl QueueStorage {
     }
 
     /// Consumes a finished upload of a thumbnail and queues the file upload.
-    pub(super) async fn handle_dependent_file_upload_with_thumbnail(
+    #[allow(clippy::too_many_arguments)]
+    pub(super) async fn handle_dependent_file_or_thumbnail_upload(
```
bnjbvr, line 354: Doc comment needs an update :)

bnjbvr (changes requested):
Thanks! There might be a subtle breaking data format change here, so requesting changes to double-check this with a test, at least.

crates/matrix-sdk/src/send_queue/upload.rs
Johennes: Ah, no I think it isn't actually. Good spot.

crates/matrix-sdk/src/send_queue/upload.rs
Johennes: I have tried to phrase the comment differently and renamed the field. I think the `parent_...` renaming helps make things clearer and feels ok-ish length-wise.

crates/matrix-sdk-base/src/store/send_queue.rs
Johennes: Argh, good catch. You're right, the variant name is encoded during serialization so deserialization fails after the renaming. I fixed it with `#[serde(rename = "UploadFileWithThumbnail")]` and added a unit test. I also noticed, that I was missing a function for the default value of `parent_is_thumbnail_upload`. It needs to be `true` rather than `false`.

crates/matrix-sdk-base/src/store/send_queue.rs
Johennes: I think technically this would work, yes. I didn't go for it originally because I wanted to minimize breaking changes and because it felt slightly cleaner to have `accumulated` only contain the requests that are actually finished. This way, we only push to `accumulated`. If we do it the other way around, we'll need to push or modify the last item depending on what the request is for. Did you have specific advantages of using this variant in mind?

crates/matrix-sdk-base/src/store/send_queue.rs
```diff
@@ -219,17 +225,23 @@ pub enum DependentQueuedRequestKind {
         key: String,
     },
 
-    /// Upload a file that had a thumbnail.
-    UploadFileWithThumbnail {
-        /// Content type for the file itself (not the thumbnail).
+    /// Upload a file or thumbnail depending on another file or thumbnail
+    /// upload.
+    #[serde(rename = "UploadFileWithThumbnail")]
```
bnjbvr, line 230: Thanks; i think we could use `serde(alias)` here, which is slightly better (will use the Rust name or the one specified in the alias, if the Rust name was missing).

```suggestion
    #[serde(alias = "UploadFileWithThumbnail")]
```

crates/matrix-sdk-base/src/store/send_queue.rs
```diff
@@ -248,6 +260,13 @@ pub enum DependentQueuedRequestKind {
     },
 }
 
+/// If parent_is_thumbnail_upload is missing, we assume the request is for a
+/// file upload following a thumbnail upload.
```
bnjbvr, line 264: Hmm, is that because that was the case before the data changes here? Maybe worth mentioning in the comment if so.

crates/matrix-sdk-base/src/store/send_queue.rs
bnjbvr: The only advantage is technical, in that it avoids containing what's effectively another flattened `AccumulatedSentMediaInfo` next to the `accumulated` vector, and also reduces the number of concepts.

I'm a bit torn, because this is likely a non-trivial change, but on the other hand that *might* be a nice simplification. Maybe that could be attempted as a follow-up, and we could open an issue to not forget about it?

crates/matrix-sdk/CHANGELOG.md
```diff
@@ -292,6 +292,11 @@ simpler methods:
     - The `issuer` field of `UserSession` was removed.
 - `SendHandle::media_handles` was generalized into a vector
   ([#4898](https://github.com/matrix-org/matrix-rust-sdk/pull/4898))
+- [**breaking] `QueueStorage::handle_dependent_file_upload_with_thumbnail`
+  was renamed to `handle_dependent_file_or_thumbnail_upload`. Under the
+  `unstable-msc4274` feature, it was generalized to allow chaining multiple
+  dependent file / thumbnail uploads.
+  ([#4897](https://github.com/matrix-org/matrix-rust-sdk/pull/4897))
```
bnjbvr, line 299: This is a private API, so no need to include in the `CHANGELOG` \o/

crates/matrix-sdk/src/send_queue/mod.rs
```diff
@@ -1589,22 +1599,34 @@ impl QueueStorage {
...
                     // Not finished yet, we should retry later => false.
                     return Ok(false);
                 };
-                self.handle_dependent_file_upload_with_thumbnail(
+                let parent_is_thumbnail_upload = {
+                    cfg_if::cfg_if! {
+                        if #[cfg(feature = "unstable-msc4274")] {
+                            parent_is_thumbnail_upload
+                        } else {
+                            true
```
bnjbvr, line 1618: Here too, can you add a tiny comment explaining why it's `true` in this case?

crates/matrix-sdk/src/send_queue/upload.rs
```diff
@@ -347,37 +349,62 @@ impl QueueStorage {
...
         }
 
-        trace!(related_to = %event_txn, "done uploading thumbnail, now queuing a request to send the media file itself");
+        trace!(related_to = %event_txn, "done uploading file or thumbnail, now queuing the dependent file or thumbnail upload request");
+
+        // If the parent request was a thumbnail upload, don't add it to the list of
+        // accumulated medias yet because its dependent file upload is still
+        // pending. If the parent request was a file upload, we know that both
+        // the file and its thumbnail (if any) have finished uploading and we
+        // can add them to the accumulated sent media.
```
bnjbvr, line 386: Super nice comment, thanks!

crates/matrix-sdk/src/send_queue/upload.rs
```diff
@@ -347,37 +349,62 @@ impl QueueStorage {
...
+        };
 
         let request = QueuedRequestKind::MediaUpload {
             content_type,
             cache_key,
-            // The thumbnail for the next upload is the file we just uploaded here.
-            thumbnail_source: Some(sent_media.file),
+            // If the previous upload was a thumbnail, it becomes the thumbnail source for the next
+            // upload.
+            thumbnail_source: if parent_is_thumbnail_upload { Some(sent_media.file) } else { None },
```
bnjbvr, line 404:
```suggestion
            thumbnail_source: parent_is_thumbnail_upload.then_some(sent_media.file),
```

bnjbvr (approved):
Looks good to me, thanks. Approving so that we can merge this as soon as my review comments have been addressed. Any chance we can also include some unit tests relevant to the changes here, or is that not worth it?

crates/matrix-sdk-base/src/store/send_queue.rs
Johennes: Yes, that's exactly it. Have amended the comment, good idea. 👍

crates/matrix-sdk/CHANGELOG.md
Johennes: Oh, right! Sorry, I keep getting these changelogs wrong. 🤦‍♂️

bnjbvr: No worries, better to include in changelogs by default, thanks for thinking about embedders :)

Johennes:
Thanks for the merge! Re the unit tests, I just wanted to say that it's a bit tricky on this PR because most of the changes are just routing new fields through existing APIs but nothing really uses these new fields yet. Therefore, I was planning to add tests in future PRs when the accumulation is actually being used by galleries.

PR #4895, opened 2025-04-07
misc(ffi): add `thread_id` to `NotificationItem`

jmartinesp:
This is needed to identify the event as being in a thread, and to reply to it inside the thread instead of in the room's timeline.

- [ ] Public API changes documented in changelogs (optional)

Signed-off-by:

PR #4899, opened 2025-04-08
fix(ui): extract NotificationEvent::thread_id to its own function

jmartinesp:
Follow up PR of https://github.com/matrix-org/matrix-rust-sdk/pull/4895 with @poljar 's suggestion.

I also added a test, I didn't do it in the previous PR because I somehow thought this was on the FFI layer instead of the UI crate.

Signed-off-by:


crates/matrix-sdk-ui/src/notification_client.rs
```diff
@@ -632,6 +632,20 @@ impl NotificationEvent {
             NotificationEvent::Invite(ev) => &ev.sender,
         }
     }
+
+    fn thread_id(&self) -> Option<OwnedEventId> {
```
poljar, line 636: Let's just add a small docstring explaining this function, i.e when it's useful.

poljar (approved):
Thanks for revisiting this and thanks a bunch for the test as well.

Left a small request but approving ahead of time anyways.

PR #4901, opened 2025-04-08
refactor(base): Continue to extract response processors

Hywan:
Sequel of https://github.com/matrix-org/matrix-rust-sdk/pull/4889.


poljar (approved):
Very nice improvements, I really like the new names you found for some of the methods.

PR #4903, opened 2025-04-08
crypto: prep work for sharing room key history bundles

richvdh:
A set of non-functional changes in `matrix-sdk-crypto` which may not make a whole lot of sense on their own, but are enablers for the actual sharing of room key history bundles (#4864). I decided to separate the refactoring work from the functional changes for easeier review.


poljar (approved):
You're right that some of this makes little sense, though I'm going to trust you that this is just a clever setup in a mysterious book you're writing.

PR #4912, opened 2025-04-09
multiverse: allow recovery keys with `h` or `l` in them

richvdh:
Currently, `h` and `l` are intercepted by the parent view to change tab,
meaning it's impossible to enter a recovery key which contains those
characters.

The fix here is very blunt: it just disables `h` and `l` for tab-changing. I
considered making it dependent on which tab is open, or what's going on in the
'Encryption' tab, but given you need to know about the alternatives (tab/cursor
keys) to switch away from the Encryption tab, I don't think that makes sense.


poljar (approved):
Oh god, right, this wasn't part of the tabbed settings view initially.

PR #4913, opened 2025-04-09
multiverse: reduce redraw rate

richvdh:
This reduces the framerate from ~62fps to 10fps.

This is a workaround for a problem in my terminal, so apologies for inflicting
it on everyone else, but here we are, and 10fps seems like it should be enough
for anyone.

The problem in question is specifically when I try to select some text by
dragging the mouse (eg, to copy a generated recovery key). If I start a drag,
but a redraw happens before the mouse has moved [a certain distance?], then the
drag doesn't work, and nothing gets selected. By reducing the framerate, I have
a much better chance of successfully starting a drag.

### DIFF
--- a/labs/multiverse/src/main.rs
+++ b/labs/multiverse/src/main.rs
@@ -368,7 +368,7 @@ impl App {
         loop {
             terminal.draw(|f| f.render_widget(&mut *self, f.area()))?;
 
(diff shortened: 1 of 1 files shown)


Hywan (approved):
We should move to an event-based drawing instead of time-based, but I'm okay with this change!

PR #4914, opened 2025-04-09
refactor(base): Continue to extract response processors, take 3

Hywan:
Sequel of https://github.com/matrix-org/matrix-rust-sdk/pull/4901.

This time, it's a bit more interesting as we are clearly reusing existing processors. Even some processors are re-using other processors.

I've cleaned up the `BaseClient::handle_timeline` quite a lot, to finally be able to move it as the new `timeline::build` processor. It is still a bit too broad from my point of view as the commit description says, but at least `BaseClient` is cleaner.

The detection and processing of the verification events is also now unified: previously it was more or less copy-pasted in three different locations.

### DIFF
--- a/crates/matrix-sdk-base/src/client.rs
+++ b/crates/matrix-sdk-base/src/client.rs
@@ -858,7 +858,7 @@ impl BaseClient {
             let processors::e2ee::to_device::Output {
                 decrypted_to_device_events: to_device,
                 room_key_updates,
-            } = processors::e2ee::to_device::from_sync_v3(
+            } = processors::e2ee::to_device::from_sync_v2(
                 &mut context,
                 &response,
                 olm_machine.as_ref(),
--- a/crates/matrix-sdk-base/src/client.rs
+++ b/crates/matrix-sdk-base/src/client.rs
@@ -381,20 +381,17 @@ impl BaseClient {
         context: &mut Context,
         room: &Room,
         limited: bool,
-        events: Vec<Raw<AnySyncTimelineEvent>>,
-        ignore_state_events: bool,
+        raw_events: Vec<Raw<AnySyncTimelineEvent>>,
         prev_batch: Option<String>,
         push_rules: &Ruleset,
-        user_ids: &mut BTreeSet<OwnedUserId>,
         room_info: &mut RoomInfo,
         notifications: &mut BTreeMap<OwnedRoomId, Vec<Notification>>,
-        ambiguity_cache: &mut AmbiguityCache,
     ) -> Result<Timeline> {
         let mut timeline = Timeline::new(limited, prev_batch);
         let mut push_context =
             self.get_push_room_context(room, room_info, &context.state_changes).await?;
 
-        for raw_event in events {
+        for raw_event in raw_events {
             // Start by assuming we have a plaintext event. We'll replace it with a
             // decrypted or UTD event below if necessary.
             let mut event = TimelineEvent::new(raw_event);
@@ -403,45 +400,9 @@ impl BaseClient {
                 Ok(e) => {
                     #[allow(clippy::single_match)]
                     match &e {
-                        AnySyncTimelineEvent::State(s) if !ignore_state_events => {
-                            match s {
-                                AnySyncStateEvent::RoomMember(member) => {
-                                    Box::pin(ambiguity_cache.handle_event(
-                                        &context.state_changes,
-                                        room.room_id(),
-                                        member,
-                                    ))
-                                    .await?;
-
-                                    match member.membership() {
-                                        MembershipState::Join | MembershipState::Invite => {
-                                            user_ids.insert(member.state_key().to_owned());
-                                        }
-                                        _ => {
-                                            user_ids.remove(member.state_key());
-                                        }
-                                    }
-
-                                    processors::profiles::upsert_or_delete(
-                                        context,
-                                        room.room_id(),
-                                        member,
-                                    );
-                                }
-                                _ => {
-                                    room_info.handle_state_event(s);
-                                }
-                            }
-
-                            let raw_event: Raw<AnySyncStateEvent> = event.raw().clone().cast();
-                            context.state_changes.add_state_event(
-                                room.room_id(),
-                                s.clone(),
-                                raw_event,
-                            );
+                        AnySyncTimelineEvent::State(_) => {
+                            // do nothing
                         }
-
-                        AnySyncTimelineEvent::State(_) => { /* do nothing */ }
 
                         AnySyncTimelineEvent::MessageLike(
                             AnySyncMessageLikeEvent::RoomRedaction(r),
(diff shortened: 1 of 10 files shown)


crates/matrix-sdk-base/src/client.rs
```diff
@@ -831,19 +790,32 @@ impl BaseClient {
...
+                    &new_info.timeline.events,
+                );
+
+            let mut other_new_user_ids = processors::state_events::dispatch_and_get_new_users(
+                &mut context,
+                (&raw_state_events_from_timeline, &state_events_from_timeline),
+                &mut room_info,
+                &mut ambiguity_cache,
+            )
+            .await?;
```
Hywan: Because `handle_timeline` is no longer handling the state events, we have to do this prior the call to `handle_timeline`, so here.

crates/matrix-sdk-base/src/client.rs
```diff
@@ -912,27 +884,39 @@ impl BaseClient {
...
+            new_user_ids.append(&mut other_new_user_ids);
```
Hywan: Same. Because `handle_timeline` doesn't handle state events, let's do it here, prior the call to `handle_timeline`.

crates/matrix-sdk-base/src/response_processors/state_events.rs
```diff
@@ -50,3 +60,52 @@ where
...
+
+/// Dispatch the state events and return the new users for this room.
+///
+/// `raw_events` and `events` must be generated from [`collect_sync`]. Events
+/// must be exactly the same list of events that are in raw_events, but
+/// deserialised. We demand them here to avoid deserialising multiple times.
+#[instrument(skip_all, fields(room_id = ?room_info.room_id))]
+pub async fn dispatch_and_get_new_users(
+    context: &mut Context,
+    (raw_events, events): (&[Raw<AnySyncStateEvent>], &[AnySyncStateEvent]),
```
andybalaam, line 93: I like this: indicates that they are a pair and should not be out of sync.

crates/matrix-sdk-base/src/client.rs
```diff
@@ -610,54 +610,6 @@ impl BaseClient {
...
-        context: &mut Context,
-        raw_events: &[Raw<AnySyncStateEvent>],
-        events: &[AnySyncStateEvent],
-        room_info: &mut RoomInfo,
-        ambiguity_cache: &mut AmbiguityCache,
-    ) -> StoreResult<BTreeSet<OwnedUserId>> {
-        let mut state_events = BTreeMap::new();
-        let mut user_ids = BTreeSet::new();
-
-        assert_eq!(raw_events.len(), events.len());
```
andybalaam, line 696: Did you deliberately remove this assertion? Isn't it a little helpful?

crates/matrix-sdk-base/src/response_processors/state_events.rs
```diff
@@ -31,26 +34,44 @@ use crate::{
...
     _context: &mut Context,
     raw_events: &[Raw<AnyStrippedStateEvent>],
 ) -> (Vec<Raw<AnyStrippedStateEvent>>, Vec<AnyStrippedStateEvent>) {
     collect(raw_events)
 }
 
-fn collect<'a, T>(raw_events: &'a [Raw<T>]) -> (Vec<Raw<T>>, Vec<T>)
+/// Collect sync timeline event, filter the sync state events, and map them with
+/// their deserialized form.
+pub fn collect_sync_from_timeline(
```
andybalaam, line 55: Maybe `collect_state_events` ? I think "state" is more important than "sync" here.

crates/matrix-sdk-base/src/response_processors/verification.rs
```diff
@@ -13,12 +13,59 @@
...
+    RoomId,
+};
 
 use super::Context;
 use crate::Result;
 
-pub async fn verification(
+/// Process the given event as a verification event if it is a candidate. The
+/// event must be decrypted.
+pub async fn process_if_candidate(
```
andybalaam, line 29: `process_if_relevant`? `process` ? Not sure these are better. In other places we just have names like `handle`, even though the method filters and only handles if relevant.

crates/matrix-sdk-base/src/sliding_sync.rs
```diff
@@ -393,7 +393,7 @@ impl BaseClient {
         room_info.mark_state_partially_synced();
         room_info.handle_encryption_state(requested_required_states);
 
-        let new_user_ids = processors::state_events::dispatch_and_get_new_users(
+        let _new_user_ids = processors::state_events::dispatch_and_get_new_users(
```
andybalaam, line 396: A comment here about why the name starts with `_` would be good.

andybalaam (approved):
This is my kind of change, thank you!

crates/matrix-sdk-base/src/client.rs
Hywan: I thought the assertion wasn't particularly helpful. Either it should become a `debug_assert_eq`, or we should introduce a type indicating those two values must be "synced", or to use a tuple as you've noticed.

crates/matrix-sdk-base/src/response_processors/state_events.rs
Hywan: The module name is already `state_events`. Having _state_ in the process name would feel a bit repetitive according to me. The idea was to indicate we want to collect `AnySyncStateEvent`. The general naming I had in mind was:  `collect_<out>[_from<in>](Any<in>StateEvent) -> Any<out>StateEvent`:

- `collect_sync(AnySyncStateEvent) -> AnySyncStateEvent`
- `collect_stripped(AnyStrippedStateEvent) -> AnyStrippedStateEvent`
- `collect_sync_from_timeline(AnySyncTimelineEvent) -> AnySyncStateEvent`

I understand that `collect_sync_from_timeline` is a bit an exceptional here.

I propose to rename to:

- `collect_from_sync(AnySyncStateEvent) -> AnySyncStateEvent`
- `collect_from_stripped(AnyStrippedStateEvent) -> AnyStrippedStateEvent`
- `collect_from_sync_timeline(AnySyncTimelineEvent) -> AnySyncStateEvent`

The name is saying nothing about the return type, but at least names are short and still clear. Also, since this is the `state_events` module, we expec to collect… state events, so either `AnySyncStateEvent` or `AnyStrippedStateEvents`. Thoughts?

I can also create submodule:

- `state_events::sync::collect_from_sync`
- `state_events::sync::collect_from_sync_timeline`
- `state_events::stripped::collect_from_stripped`

If the input type is the same as the output type, we can even from the `from_…` suffix:

- `state_events::sync::collect`
- `state_events::sync::collect_from_sync_timeline`
- `state_events::stripped::collect`


Thoughts?

crates/matrix-sdk-base/src/response_processors/verification.rs
Hywan: I like `process_if_relevant`. Have changed to that.

crates/matrix-sdk-base/src/response_processors/state_events.rs
andybalaam: I think I like what you have now. I still have a lingering feeling that the word `collect` isn't expressing what we mean, but I can't think of anything better.

Hywan: Same feeling.

PR #4915, opened 2025-04-09
feat(bindings): expose `is_public` and `join_rule` in notifications

Velin92:
Would be useful to tell from a notification if the room that generated it, is public or not, or if has some specific join rule, since a client may want to hide or show the media based on the join rule of the room for safety reasons


Hywan (approved):
Looks good, thanks!

poljar (approved):
Looks good.

PR #4917, opened 2025-04-09
Cleanups and extensions to tracing span logic

richvdh:
A grab-bag of improvements to how we handle tracing spans, in an attempt to make debugging easier.

### DIFF
--- a/crates/matrix-sdk-ui/src/sync_service.rs
+++ b/crates/matrix-sdk-ui/src/sync_service.rs
@@ -42,7 +42,7 @@ use tokio::sync::{
     mpsc::{Receiver, Sender},
     Mutex as AsyncMutex, OwnedMutexGuard,
 };
-use tracing::{error, info, instrument, trace, warn, Instrument, Level};
+use tracing::{error, info, instrument, trace, warn, Instrument, Level, Span};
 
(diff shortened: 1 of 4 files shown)


crates/matrix-sdk-ui/src/sync_service.rs
```diff
@@ -487,6 +496,14 @@ struct SyncServiceInner {
     /// The offline mode is described in the [`State::Offline`] enum variant.
     with_offline_mode: bool,
     state: SharedObservable<State>,
+
+    /// The parent tracing span to use for the tasks within this service.
+    ///
+    /// Normally this will be [`Span::none`], but it may be useful to assign a
+    /// defined span, for example if there is more than one active sync
+    /// service.
```
poljar, line 504: Did you somewhere encounter the "more than one active sync service" case?

That should only happen in a multi-account scenario.

poljar (approved):
Looks good, just wondering about the use-case.

crates/matrix-sdk-ui/src/sync_service.rs
richvdh: Yeah, I was struggling with this in an integration test.

PR #4925, opened 2025-04-11
refactor(base): Continue to extract response processors, take 4

Hywan:
Sequel of https://github.com/matrix-org/matrix-rust-sdk/pull/4914.

This time, we remove the last `handle_` method on `BaseClient`! And we create another set of response processors, `ephemeral_events`, which remove duplicated code.


stefanceriu (approved):
Very nice 👌

PR #4928, opened 2025-04-14
refactor(base): Continue to extract response processors, take 5

Hywan:
Sequel of https://github.com/matrix-org/matrix-rust-sdk/pull/4925.

This time, the sync v2 (`BaseClient::receive_sync_response`) is now fully using response processors. This PR introduces the new `rooms::sync_v2::update_(joined|left|invited|knocked)_room` processors! It also removes duplicated code around push notifications by improving the `notification::Notification` type.

### DIFF
--- a/crates/matrix-sdk-base/src/response_processors/room/sync_v2.rs
+++ b/crates/matrix-sdk-base/src/response_processors/room/sync_v2.rs
@@ -85,7 +85,7 @@ pub async fn update_joined_room(
     updated_members_in_room.insert(room_id.to_owned(), new_user_ids.clone());
 
     #[cfg(feature = "e2e-encryption")]
-    let olm_machine = e2ee.olm_machine.clone();
+    let olm_machine = e2ee.olm_machine;
 
     let timeline = timeline::build(
         context,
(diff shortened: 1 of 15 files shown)


crates/matrix-sdk-base/src/response_processors/room/sync_v2.rs
```diff
@@ -0,0 +1,140 @@
...
+use super::super::{account_data, ephemeral_events, state_events, timeline, Context};
+use crate::{
+    store::{ambiguity_map::AmbiguityCache, BaseStateStore},
+    sync::JoinedRoomUpdate,
+    RequestedRequiredStates, Result, RoomInfoNotableUpdate, RoomState,
+};
+
+/// Process updates of a joined room.
+#[allow(clippy::too_many_arguments)]
+pub async fn update_joined_room(
```
stefanceriu, line 34: Probably missing something here but can't we name all the processor main methods `process_fo` e.g. `process_joined_room` in this case? 🤔 
`update_foo` makes me think it updates something in place where it actually returns an update.

stefanceriu (approved):
Just lovely, thank you! 👏

crates/matrix-sdk-base/src/response_processors/room/sync_v2.rs
stefanceriu: Ah, got it, fair enough!

PR #4929, opened 2025-04-14
fix(sdk): Remove `include_heroes` in sliding sync requests

Hywan:
This patch removes all mentions of `include_heroes`. This field isn't part of the MSC4186. We have the `test_compute_heroes_from_sliding_sync` which continues to pass, everything fine on that side. There is no mention of `include_heroes` in Synapse at the time of writing. It seems to be an artifact from the sliding sync proxy experiment.

---

* Related to https://github.com/ruma/ruma/pull/2058


bnjbvr (approved):
Nice!

PR #4933, opened 2025-04-15
feat(ffi): introduce a `ThreadSummary` type within `MsgLikeContent`

stefanceriu:
…that holds information on the thread the given item is the root of

- it holds the latest event content and sender at the moment but will hold more information in the future e.g. number of replies, if it's unread etc.
- the field is not currently being populate but is delivered earlier so it can power shipping the UI side on the embedders


bindings/matrix-sdk-ffi/src/timeline/msg_like.rs
```diff
@@ -56,10 +60,12 @@ pub enum MsgLikeKind {
 pub struct MsgLikeContent {
     pub kind: MsgLikeKind,
     pub reactions: Vec<Reaction>,
-    /// Event ID of the thread root, if this is a threaded message.
-    pub thread_root: Option<String>,
     /// The event this message is replying to, if any.
     pub in_reply_to: Option<Arc<InReplyToDetails>>,
+    /// Event ID of the thread root, if this is a threaded message.
```
bnjbvr, line 65:
```suggestion
    /// Event ID of the thread root, if this is a message in a thread.
```

crates/matrix-sdk-ui/src/timeline/event_item/content/msg_like.rs
```diff
@@ -36,17 +36,31 @@ pub enum MsgLikeKind {
     UnableToDecrypt(EncryptedMessage),
 }
 
+#[derive(Clone, Debug)]
+pub struct ThreadSummary {
+    pub latest_event_details: TimelineDetails<Box<ThreadSummaryLatestEvent>>,
```
bnjbvr, line 41: Let's name this `latest_event` (details refers to `TimelineDetails` and is rather an implementation detail (heh))

crates/matrix-sdk-ui/src/timeline/event_item/content/msg_like.rs
```diff
@@ -36,17 +36,31 @@ pub enum MsgLikeKind {
...
 /// information.
 #[derive(Clone, Debug)]
 pub struct MsgLikeContent {
     pub kind: MsgLikeKind,
     pub reactions: ReactionsByKeyBySender,
-    /// Event ID of the thread root, if this is a threaded message.
-    pub thread_root: Option<OwnedEventId>,
     /// The event this message is replying to, if any.
     pub in_reply_to: Option<InReplyToDetails>,
+    /// Event ID of the thread root, if this is a threaded message.
```
bnjbvr, line 60: Let's try to make it super clear the difference between thread root and messages-in-thread. "Threaded message" could be the root, in my head.

```suggestion
    /// Event ID of the thread root, if this is a message in a thread.
```

crates/matrix-sdk-ui/src/timeline/event_item/content/reply.rs
```diff
@@ -135,11 +135,13 @@ impl RepliedToEvent {
         let content = match event.original_content() {
             Some(content) => match content {
                 AnyMessageLikeEventContent::RoomMessage(c) => {
-                    // Assume we're not interested in reactions and thread info in this context:
+                    // Assume we're not interested in aggregations in this context:
                     // this is information for an embedded (replied-to) event, that will usually not
                     // include detailed information like reactions.
```
bnjbvr, line 140: Now that the function has four `None` parameters, maybe it's a sign that the `content` bound above should be a `MsgLikeKind`, not a `MsgLike`? (can be a future refactor)

crates/matrix-sdk-ui/src/timeline/event_handler.rs
```diff
@@ -418,6 +418,7 @@ impl<'a, 'o> TimelineEventHandler<'a, 'o> {
                                 reactions: Default::default(),
                                 thread_root: None,
                                 in_reply_to: None,
+                                thread_summary: None,
```
bnjbvr, line 421: hmm, can't stickers be part of threads? Can we add a TODO comment about this, and link it to a github issue (new or existing one), please?

crates/matrix-sdk-ui/src/timeline/event_handler.rs
```diff
@@ -762,6 +765,7 @@ impl<'a, 'o> TimelineEventHandler<'a, 'o> {
                 reactions: reactions.clone(),
                 thread_root: thread_root.clone(),
                 in_reply_to: in_reply_to.clone(),
+                thread_summary: thread_summary.clone(),
```
bnjbvr, line 768: It might be time to introduce `MsgLikeContent::with_kind()` :)

crates/matrix-sdk-ui/src/timeline/event_handler.rs
```diff
@@ -869,6 +874,7 @@ impl<'a, 'o> TimelineEventHandler<'a, 'o> {
...
```
bnjbvr, line 877: (and use it here too)

crates/matrix-sdk-ui/src/timeline/event_handler.rs
```diff
@@ -921,6 +927,7 @@ impl<'a, 'o> TimelineEventHandler<'a, 'o> {
                 reactions: Default::default(),
                 thread_root: None,
                 in_reply_to: None,
+                thread_summary: None,
```
bnjbvr, line 930: Can't poll be part of a thread or thread roots? (if they can: let's also add a TODO here)

crates/matrix-sdk-ui/src/timeline/event_handler.rs
```diff
@@ -1403,6 +1410,7 @@ impl<'a, 'o> TimelineEventHandler<'a, 'o> {
                 reactions: msglike.reactions.clone(),
                 thread_root: msglike.thread_root.clone(),
                 in_reply_to,
+                thread_summary: msglike.thread_summary.clone(),
```
bnjbvr, line 1413: (and we could use `MsgLikeContent::with_kind()` here too)

crates/matrix-sdk-ui/src/timeline/event_item/content/mod.rs
```diff
@@ -198,6 +198,7 @@ impl TimelineItemContent {
                 let reactions = Default::default();
                 let thread_root = None;
                 let in_reply_to = None;
+                let thread_summary = None;
```
bnjbvr, line 201: Probably the latest preview item should also return a MsgLikeKind, not a MsgLike, then. (Future refactor :innocent:)

bnjbvr (approved):
lgtm, thanks!

crates/matrix-sdk-ui/src/timeline/event_handler.rs
stefanceriu: Stickers can be both part of a thread and the thread root, at least on web. The `thread_summary` is None here just because this PR doesn't populated it. I didn't consider a todo necessary as we're going to have to handle that properly soon anyway.

The `thread_root` on the other hand is a remnant of the previous (pre MsgLike) implementation where it was only available on Message types. Seeing as we're building threads properly I don't think we should bother with fixing that, I would just remove the whole thing instead.

crates/matrix-sdk-ui/src/timeline/event_handler.rs
stefanceriu: To make things fun polls can indeed be thread roots but they can't be sent inside threads 🤪

PR #4938, opened 2025-04-15
feat(multiverse): Add support to join rooms you've been invited to

poljar:
This is a bit bigger than intended because I've added mouse support to press the "Accept" button.

Rejecting the invite magnifies an issue we have. If a room we are viewing gets removed from the room list, it will still be shown in the room view despite another room being selected.

But I think that's a fix for another time.

### DIFF
--- a/labs/multiverse/src/main.rs
+++ b/labs/multiverse/src/main.rs
@@ -8,7 +8,12 @@ use std::{
 
 use clap::Parser;
 use color_eyre::Result;
-use crossterm::event::{self, Event, KeyCode, KeyEvent, KeyEventKind, KeyModifiers};
+use crossterm::{
+    event::{
+        self, DisableMouseCapture, EnableMouseCapture, Event, KeyCode, KeyEvent, KeyModifiers,
+    },
+    execute,
+};
 use futures_util::{pin_mut, StreamExt as _};
 use imbl::Vector;
 use layout::Flex;
@@ -110,6 +115,7 @@ async fn main() -> Result<()> {
     event_cache.enable_storage()?;
 
     let terminal = ratatui::init();
+    execute!(stdout(), EnableMouseCapture)?;
     let mut app = App::new(client).await?;
 
     app.run(terminal).await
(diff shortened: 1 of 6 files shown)


labs/multiverse/src/main.rs
```diff
@@ -315,37 +321,46 @@ impl App {
         self.state.global_mode = mode;
     }
 
-    async fn handle_global_key_press(&mut self, key: KeyEvent) -> Result<bool> {
+    async fn handle_global_key_press(&mut self, event: Event) -> Result<bool> {
```
bnjbvr, line 324: The function name is now a lie, since it can call into an event handler for a non-key event; let's maybe rename it `handle_global_event`?

bnjbvr (approved):
Nice, thanks.

The room list focus bug I was talking about in private Matrix channel happens again here. If you're focused on the first room, then you get an invite to a new room, then the focus moves to the invited room, but you still see the room view of the previous room. Might be a multiverse render bug to fix later, as a followup.

Code looks good. Approving, assuming you're fine with the suggested renaming; if not, let's discuss again.

labs/multiverse/src/main.rs
poljar: Sure: https://github.com/matrix-org/matrix-rust-sdk/pull/4938/commits/a3795e7038520cee4fae614e74f71911a385ddb8.

labs/multiverse/src/widgets/room_view/invited_room.rs
```diff
@@ -0,0 +1,276 @@
...
+            Mode::Loading { task } => task.abort(),
+            Mode::Joining { task } => task.abort(),
+            Mode::Leaving { task } => task.abort(),
+            Mode::Loaded { .. } => {}
+            Mode::Done => {}
+        }
+    }
+}
+
+enum SelectedButton {
```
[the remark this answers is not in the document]
poljar: The lingo has been adopted from the `tui-framework-experiments` crate: https://github.com/joshka/tui-framework-experiment/blob/main/examples/widgets/tabs/buttons.rs.

Meaning, we can't really get rid of the `button.select()` calls to focus a button.

Renamed in https://github.com/matrix-org/matrix-rust-sdk/pull/4938/commits/6ea77db0c4ccadd5445c440fed2e7f6eebe4202a.

Issue #4825, opened 2025-03-21
Heroes are empty from room to which you have been invited


Velin92:
This applies both to the `RoomSummary` from the `RoomList` and the `RoomPreviewInfo` of their preview.

However the inviter is always available, which is a member of the room and therefore should be a hero, so this is a bit weird.

This creates some issue when you want to render the avatar of a DM invite, where it should be determined by the heroes of the roon.

bnjbvr:
Thanks for opening an issue! Likely same underlying issue as https://github.com/matrix-org/matrix-rust-sdk/issues/4051 but manifesting in a different way; let's keep both and conclude when 4051 is fixed.

Hywan:
thanks for the triaging, I think you're right @bnjbvr, let's see.

PR #4934, opened 2025-04-15
refactor(base): Continue to extract response processors, take 6

Hywan:
Sequel of https://github.com/matrix-org/matrix-rust-sdk/pull/4928.

This PR continues to refactor and clean up the response workflow as so-called response processors.

Each patch should be reviewed individually. There is nothing fancy, it's mostly code moves.

This time, the sliding sync (`BaseClient::process_sliding_sync`) is now fully using response processors. This PR introduces the new `rooms::msc4186` processors! It does several things, but the most notable is: a nice clean up around 4 `Option`s being returned by a method, and the removal of a large `clone` of room account data.

PR #4940, opened 2025-04-15
fix(base): The room computed display name is persisted

Hywan:
Sequel of https://github.com/matrix-org/matrix-rust-sdk/pull/4934.

This PR should be reviewed patch-by-patch.

The room display name is computed during the sync (for both sync v2 and sliding sync), and is kept in memory. To do that, the `StateChanges` needs to be saved in the state store, because computing the display name requires data from the state store. Problem is: the display name isn't saved after that. One needs to wait on another sync to happen for this particular room to persist/save the computed display name.

This PR fixes that by using the recently added response processors. A test confirms the problem is fixed.

---

Combo.

* Fixes https://github.com/matrix-org/matrix-rust-sdk/issues/4051
* Fixes https://github.com/element-hq/element-x-ios/issues/3365
* Fixes https://github.com/element-hq/element-x-android/issues/3920
* Fixes https://github.com/matrix-org/matrix-rust-sdk/issues/4825
* Closes https://github.com/element-hq/customer-success/issues/405
* Closes https://github.com/element-hq/customer-success/issues/469

### DIFF
--- a/crates/matrix-sdk-base/src/response_processors/changes.rs
+++ b/crates/matrix-sdk-base/src/response_processors/changes.rs
@@ -25,6 +25,13 @@ use crate::{
     Result, StateChanges,
 };
 
+/// Save the [`StateChanges`] from the [`Context`] inside the [`BaseStateStore`]
+/// only! The changes aren't applied on the in-memory rooms.
+#[instrument(skip_all)]
+pub async fn save_only(context: Context, state_store: &BaseStateStore) -> Result<()> {
+    save_changes(&context.state_changes, state_store, None).await
+}
+
 /// Save the [`StateChanges`] from the [`Context`] inside the
 /// [`BaseStateStore`], and apply them on the in-memory rooms.
 #[instrument(skip_all)]
--- a/crates/matrix-sdk-base/src/response_processors/changes.rs
+++ b/crates/matrix-sdk-base/src/response_processors/changes.rs
@@ -22,14 +22,17 @@ use tracing::{error, instrument, trace};
 use super::Context;
 use crate::{
     store::{BaseStateStore, StateStoreExt as _},
-    Result, StateChanges,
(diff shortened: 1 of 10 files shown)


crates/matrix-sdk-base/src/response_processors/changes.rs
```diff
@@ -25,6 +25,13 @@ use crate::{
     Result, StateChanges,
 };
 
+/// Save the [`StateChanges`] from the [`Context`] inside the [`BaseStateStore`]
+/// only! The changes aren't applied on the in-memory rooms.
+#[instrument(skip_all)]
+pub async fn save_only(context: Context, state_store: &BaseStateStore) -> Result<()> {
```
stefanceriu, line 31: Perhaps `save_to_store` is a better name here?

Hywan:
The last two commits should solve the problem. Now, a `RoomInfoNotableUpdateReasons::DISPLAY_NAME` is emitted.

stefanceriu:
> The last two commits should solve the problem. Now, a `RoomInfoNotableUpdateReasons::DISPLAY_NAME` is emitted.

And they do, complete and utter success! 👏👏👏

stefanceriu (approved):

Hywan:
@stefanceriu Would you like to review again with the [_regression fix_](https://github.com/matrix-org/matrix-rust-sdk/pull/4940/commits/46eacf81319e611fc1038bf89150845fc1198e83) due to the [_bug fix_](https://github.com/matrix-org/matrix-rust-sdk/pull/4940/commits/3ccb3603f1e1690b6059d869a9610f47dc2d43cc)?

stefanceriu (approved):
Makes perfect sense to me. Let's remove that hack when it's not holiday season and try to track down all the required update reasons.

Issue #4869, opened 2025-03-31
[meta] Threads


bnjbvr:
The SDK is ripe for supporting threads, which allows parallel topical discussions and reduce confusion. And, it will make it possible to add support for it in embeddings, notably in ElementX, closing the gap with the Element app's feature set.

We'll update this issue as soon as we'll identify the next steps.
Here is the high-level overview of what we're planning to implement:

## Interacting with threads

- https://github.com/matrix-org/matrix-rust-sdk/pull/4819
- https://github.com/matrix-org/matrix-rust-sdk/issues/4835
- #4973 
  - [x] List all the threads of a room with `/threads`
  - [x] Retrieve all the thread events with `/relations`
- https://github.com/matrix-org/matrix-rust-sdk/issues/5535

## Thread-focused timeline

A new timeline focus will be introduced that will represent a thread. Threads will include message-like events, and will be updated in real-time.

- https://github.com/matrix-org/matrix-rust-sdk/issues/4833
- https://github.com/matrix-org/matrix-rust-sdk/pull/4839
- https://github.com/matrix-org/matrix-rust-sdk/pull/4853
- https://github.com/matrix-org/matrix-rust-sdk/pull/4854
- https://github.com/matrix-org/matrix-rust-sdk/pull/4860
- https://github.com/matrix-org/matrix-rust-sdk/pull/5032
- How to filter out in-thread events from a timeline? Using the `EventFilter` might be sufficient, but if it's not we'll need a different type.
- [x] Handle events related to an in-thread events, from sync, in real-time. Probably means finishing https://github.com/matrix-org/matrix-rust-sdk/pull/4655 so we can only observe aggregations from sync, and apply them in the thread if needs be.

## Thread data model and persistent storage

spin off in their own issues:

- #5123 
- #5122

## Read receipts

We want to support thread read receipts. This will *not* interact with unread status on rooms, yet.

- [x] show thread receipts
- https://github.com/matrix-org/matrix-rust-sdk/issues/4834

## Thread subscriptions

See also https://github.com/matrix-org/matrix-rust-sdk/issues/5038, which includes more details.

## Bonus / Moonshots / Nice-to-have

- https://github.com/matrix-org/matrix-rust-sdk/issues/4113
- https://github.com/matrix-org/matrix-rust-sdk/issues/5216 
- [ ] #5121
- [x] #5223

PR #4941, opened 2025-04-15
feat(event cache): don't strip a bundled thread relation when storing it in the event cache

bnjbvr:
We may be reusing the bundled thread info as a location where to save the thread summary in a near future, so it's useful to keep it from sync, for the sake of our experiments.

Part of #4869.


bnjbvr:
Drafting; need to think about it a bit more.

PR #4947, opened 2025-04-16
Widget driver: Distinguish room state and timeline events

robintown:
This is an implementation of [an update to MSC2762](https://github.com/matrix-org/matrix-spec-proposals/pull/4237). It changes the widget driver to split state updates out into an `update_state` action and use the `send_event` action exclusively for forwarding timeline events. Accordingly, `read_events` now always reads from /messages, never the state store.

For https://github.com/element-hq/voip-internal/issues/289

### DIFF
--- a/crates/matrix-sdk/src/widget/machine/driver_req.rs
+++ b/crates/matrix-sdk/src/widget/machine/driver_req.rs
@@ -21,7 +21,7 @@ use ruma::{
         account::request_openid_token, delayed_events::update_delayed_event,
         to_device::send_event_to_device,
     },
-    events::{AnyTimelineEvent, AnyToDeviceEventContent},
+    events::{AnyStateEvent, AnyTimelineEvent, AnyToDeviceEventContent},
     serde::Raw,
     to_device::DeviceIdOrAllDevices,
     OwnedUserId,
@@ -48,11 +48,11 @@ pub(crate) enum MatrixDriverRequestData {
     /// Get OpenId token for a given request ID.
     GetOpenId,
 
-    /// Read message event(s).
-    ReadMessageLikeEvent(ReadMessageLikeEventRequest),
+    /// Read events from the timeline.
+    ReadEvents(ReadEventsRequest),
 
-    /// Read state event(s).
-    ReadStateEvent(ReadStateEventRequest),
+    /// Read room state entries.
+    ReadState(ReadStateRequest),
 
     /// Send Matrix event that corresponds to the given description.
     SendEvent(SendEventRequest),
@@ -170,26 +170,32 @@ impl FromMatrixDriverResponse for request_openid_token::v3::Response {
     }
 }
 
-/// Ask the client to read Matrix event(s) that corresponds to the given
+/// Ask the client to read Matrix events that correspond to the given
 /// description and return a list of events as a response.
 #[derive(Clone, Debug)]
-pub(crate) struct ReadMessageLikeEventRequest {
+pub(crate) struct ReadEventsRequest {
     /// The event type to read.
     // TODO: This wants to be `MessageLikeEventType`` but we need a type which supports `as_str()`
     // as soon as ruma supports `as_str()` on `MessageLikeEventType` we can use it here.
     pub(crate) event_type: String,
 
+    /// The `state_key` to read. If None, this will read events regardless of
+    /// whether they are state events. If `Some(Any)`, this will only read state
+    /// events of the given type. If set to a specific state key, this will only
+    /// read state events of the given type matching that state key.
+    pub(crate) state_key: Option<StateKeySelector>,
+
     /// The maximum number of events to return.
     pub(crate) limit: u32,
 }
 
-impl From<ReadMessageLikeEventRequest> for MatrixDriverRequestData {
-    fn from(value: ReadMessageLikeEventRequest) -> Self {
-        MatrixDriverRequestData::ReadMessageLikeEvent(value)
+impl From<ReadEventsRequest> for MatrixDriverRequestData {
+    fn from(value: ReadEventsRequest) -> Self {
+        MatrixDriverRequestData::ReadEvents(value)
     }
 }
 
-impl MatrixDriverRequest for ReadMessageLikeEventRequest {
+impl MatrixDriverRequest for ReadEventsRequest {
     type Response = Vec<Raw<AnyTimelineEvent>>;
 }
 
@@ -205,28 +211,40 @@ impl FromMatrixDriverResponse for Vec<Raw<AnyTimelineEvent>> {
     }
 }
 
-/// Ask the client to read Matrix event(s) that corresponds to the given
-/// description and return a list of events as a response.
+/// Ask the client to read Matrix room state entries corresponding to the given
+/// description and return a list of state events as a response.
 #[derive(Clone, Debug)]
-pub(crate) struct ReadStateEventRequest {
+pub(crate) struct ReadStateRequest {
     /// The event type to read.
     // TODO: This wants to be `TimelineEventType` but we need a type which supports `as_str()`
     // as soon as ruma supports `as_str()` on `TimelineEventType` we can use it here.
     pub(crate) event_type: String,
 
-    /// The `state_key` to read, or `Any` to receive any/all events of the given
-    /// type, regardless of their `state_key`.
+    /// The `state_key` to read, or `Any` to receive any/all room state entries
+    /// of the given type, regardless of their `state_key`.
     pub(crate) state_key: StateKeySelector,
 }
 
-impl From<ReadStateEventRequest> for MatrixDriverRequestData {
-    fn from(value: ReadStateEventRequest) -> Self {
-        MatrixDriverRequestData::ReadStateEvent(value)
+impl From<ReadStateRequest> for MatrixDriverRequestData {
+    fn from(value: ReadStateRequest) -> Self {
+        MatrixDriverRequestData::ReadState(value)
     }
 }
 
-impl MatrixDriverRequest for ReadStateEventRequest {
-    type Response = Vec<Raw<AnyTimelineEvent>>;
+impl MatrixDriverRequest for ReadStateRequest {
+    type Response = Vec<Raw<AnyStateEvent>>;
+}
+
+impl FromMatrixDriverResponse for Vec<Raw<AnyStateEvent>> {
+    fn from_response(ev: MatrixDriverResponse) -> Option<Self> {
+        match ev {
+            MatrixDriverResponse::StateRead(response) => Some(response),
+            _ => {
+                error!("bug in MatrixDriver, received wrong event response");
+                None
+            }
+        }
+    }
 }
 
 /// Ask the client to send Matrix event that corresponds to the given
--- a/crates/matrix-sdk/src/widget/machine/driver_req.rs
+++ b/crates/matrix-sdk/src/widget/machine/driver_req.rs
@@ -55,7 +55,7 @@ pub(crate) enum MatrixDriverRequestData {
     ReadStateEvent(ReadStateEventRequest),
 
     /// Send Matrix event that corresponds to the given description.
-    SendMatrixEvent(SendEventRequest),
+    SendEvent(SendEventRequest),
 
     /// Send a to-device message over the Matrix homeserver.
     SendToDeviceEvent(SendToDeviceRequest),
@@ -196,7 +196,7 @@ impl MatrixDriverRequest for ReadMessageLikeEventRequest {
 impl FromMatrixDriverResponse for Vec<Raw<AnyTimelineEvent>> {
     fn from_response(ev: MatrixDriverResponse) -> Option<Self> {
         match ev {
-            MatrixDriverResponse::MatrixEventRead(response) => Some(response),
+            MatrixDriverResponse::EventsRead(response) => Some(response),
             _ => {
                 error!("bug in MatrixDriver, received wrong event response");
                 None
@@ -251,7 +251,7 @@ pub(crate) struct SendEventRequest {
 
 impl From<SendEventRequest> for MatrixDriverRequestData {
     fn from(value: SendEventRequest) -> Self {
-        MatrixDriverRequestData::SendMatrixEvent(value)
+        MatrixDriverRequestData::SendEvent(value)
     }
 }
 
@@ -262,7 +262,7 @@ impl MatrixDriverRequest for SendEventRequest {
 impl FromMatrixDriverResponse for SendEventResponse {
     fn from_response(ev: MatrixDriverResponse) -> Option<Self> {
         match ev {
-            MatrixDriverResponse::MatrixEventSent(response) => Some(response),
+            MatrixDriverResponse::EventSent(response) => Some(response),
             _ => {
                 error!("bug in MatrixDriver, received wrong event response");
                 None
@@ -303,7 +303,7 @@ impl TryInto<send_event_to_device::v3::Response> for MatrixDriverResponse {
 
     fn try_into(self) -> Result<send_event_to_device::v3::Response, Self::Error> {
         match self {
-            MatrixDriverResponse::MatrixToDeviceSent(response) => Ok(response),
+            MatrixDriverResponse::ToDeviceSent(response) => Ok(response),
             _ => Err(de::Error::custom("bug in MatrixDriver, received wrong event response")),
         }
     }
@@ -330,7 +330,7 @@ impl MatrixDriverRequest for UpdateDelayedEventRequest {
 impl FromMatrixDriverResponse for update_delayed_event::unstable::Response {
     fn from_response(ev: MatrixDriverResponse) -> Option<Self> {
         match ev {
-            MatrixDriverResponse::MatrixDelayedEventUpdate(response) => Some(response),
+            MatrixDriverResponse::DelayedEventUpdated(response) => Some(response),
             _ => {
                 error!("bug in MatrixDriver, received wrong event response");
                 None
--- a/crates/matrix-sdk/src/widget/machine/from_widget.rs
+++ b/crates/matrix-sdk/src/widget/machine/from_widget.rs
@@ -35,7 +35,7 @@ pub(super) enum FromWidgetRequest {
     #[serde(rename = "get_openid")]
     GetOpenId {},
     #[serde(rename = "org.matrix.msc2876.read_events")]
-    ReadEvent(ReadEventRequest),
+    ReadEvent(ReadEventsRequest),
     SendEvent(SendEventRequest),
     SendToDevice(SendToDeviceRequest),
     #[serde(rename = "org.matrix.msc4157.update_delayed_event")]
@@ -133,6 +133,7 @@ impl SupportedApiVersionsResponse {
                 ApiVersion::V0_0_1,
                 ApiVersion::V0_0_2,
                 ApiVersion::MSC2762,
+                ApiVersion::MSC2762UpdateState,
                 ApiVersion::MSC2871,
                 ApiVersion::MSC3819,
             ],
@@ -154,6 +155,10 @@ pub(super) enum ApiVersion {
     /// Supports sending and receiving of events.
     #[serde(rename = "org.matrix.msc2762")]
     MSC2762,
+
+    /// Supports receiving of room state with the `update_state` action.
+    #[serde(rename = "org.matrix.msc2762_update_state")]
+    MSC2762UpdateState,
 
     /// Supports sending of approved capabilities back to the widget.
     #[serde(rename = "org.matrix.msc2871")]
@@ -181,22 +186,15 @@ pub(super) enum ApiVersion {
 }
 
 #[derive(Deserialize, Debug)]
-#[serde(untagged)]
-pub(super) enum ReadEventRequest {
-    ReadStateEvent {
-        #[serde(rename = "type")]
-        event_type: String,
-        state_key: StateKeySelector,
-    },
-    ReadMessageLikeEvent {
-        #[serde(rename = "type")]
-        event_type: String,
-        limit: Option<u32>,
-    },
+pub(super) struct ReadEventsRequest {
+    #[serde(rename = "type")]
+    pub(super) event_type: String,
+    pub(super) state_key: Option<StateKeySelector>,
+    pub(super) limit: Option<u32>,
 }
 
 #[derive(Debug, Serialize)]
-pub(super) struct ReadEventResponse {
+pub(super) struct ReadEventsResponse {
     pub(super) events: Vec<Raw<AnyTimelineEvent>>,
 }
 
--- a/crates/matrix-sdk/src/widget/machine/from_widget.rs
+++ b/crates/matrix-sdk/src/widget/machine/from_widget.rs
@@ -152,15 +152,15 @@ pub(super) enum ApiVersion {
     #[serde(rename = "0.0.2")]
     V0_0_2,
 
-    /// Supports sending and receiving of events.
+    /// Supports sending and receiving events.
     #[serde(rename = "org.matrix.msc2762")]
     MSC2762,
 
-    /// Supports receiving of room state with the `update_state` action.
+    /// Supports receiving room state with the `update_state` action.
     #[serde(rename = "org.matrix.msc2762_update_state")]
     MSC2762UpdateState,
 
-    /// Supports sending of approved capabilities back to the widget.
+    /// Supports sending approved capabilities back to the widget.
     #[serde(rename = "org.matrix.msc2871")]
     MSC2871,
 
@@ -176,7 +176,7 @@ pub(super) enum ApiVersion {
     #[serde(rename = "org.matrix.msc2876")]
     MSC2876,
 
-    /// Supports sending and receiving of to-device events.
+    /// Supports sending and receiving to-device events.
     #[serde(rename = "org.matrix.msc3819")]
     MSC3819,
 
--- a/crates/matrix-sdk/src/widget/machine/incoming.rs
+++ b/crates/matrix-sdk/src/widget/machine/incoming.rs
@@ -64,13 +64,16 @@ pub(crate) enum MatrixDriverResponse {
     /// A response to an `Action::GetOpenId` command.
     OpenIdReceived(request_openid_token::v3::Response),
     /// Client read some Matrix event(s).
-    /// A response to an `Action::ReadMatrixEvent` commands.
-    MatrixEventRead(Vec<Raw<AnyTimelineEvent>>),
+    /// A response to a `Action::ReadEvent` command.
+    EventsRead(Vec<Raw<AnyTimelineEvent>>),
     /// Client sent some Matrix event. The response contains the event ID.
-    /// A response to an `Action::SendMatrixEvent` command.
-    MatrixEventSent(SendEventResponse),
-    MatrixToDeviceSent(send_event_to_device::v3::Response),
-    MatrixDelayedEventUpdate(delayed_events::update_delayed_event::unstable::Response),
+    /// A response to a `Action::SendEvent` command.
+    EventSent(SendEventResponse),
+    /// A response to a `Action::SendToDevice` command.
+    ToDeviceSent(send_event_to_device::v3::Response),
+    /// Client updated a delayed event.
+    /// A response to a `Action::UpdateDelayedEvent` command.
+    DelayedEventUpdated(delayed_events::update_delayed_event::unstable::Response),
 }
 
 pub(super) struct IncomingWidgetMessage {
--- a/crates/matrix-sdk/src/widget/machine/incoming.rs
+++ b/crates/matrix-sdk/src/widget/machine/incoming.rs
@@ -14,13 +14,15 @@
 
 use ruma::{
     api::client::{account::request_openid_token, delayed_events, to_device::send_event_to_device},
-    events::{AnyTimelineEvent, AnyToDeviceEvent},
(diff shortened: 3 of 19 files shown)


bnjbvr (changes requested):
I've only looked at the first three commits, a first batch of comments; the change in the second commit is likely wrong.

testing/matrix-sdk-test/src/event_factory.rs
```diff
@@ -116,16 +116,14 @@ impl<C: EventContent> Default for Unsigned<C> {
...
+#[derive(Clone, Debug)]
 pub struct EventBuilder<C: EventContent> {
     sender: Option<OwnedUserId>,
     /// Whether the event is an ephemeral one. As such, it doesn't require a
     /// room id or a sender.
     is_ephemeral: bool,
     room: Option<OwnedRoomId>,
     event_id: Option<OwnedEventId>,
-    /// Whether the event should *not* have an event id. False by default.
-    no_event_id: bool,
```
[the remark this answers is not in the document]
robintown: The idea is you can still remove the event ID by calling `EventBuilder::no_event_id`, as before. The generation of an event ID has been changed to happen immediately upon creation, rather than at the point of converting to JSON, so that we don't need this third "event ID needed but not yet generated" state.

testing/matrix-sdk-test/src/event_factory.rs
```diff
@@ -263,6 +264,14 @@ where
         Raw::new(&self.construct_json(false)).unwrap().cast()
     }
 
+    pub fn into_raw_state(self) -> Raw<AnyStateEvent> {
+        Raw::new(&self.construct_json(true)).unwrap().cast()
+    }
+
+    pub fn into_raw_sync_state(self) -> Raw<AnySyncStateEvent> {
+        Raw::new(&self.construct_json(false)).unwrap().cast()
+    }
```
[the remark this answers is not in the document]
bnjbvr: (Getting the Raw event in general is fine, since then the caller decides what to do with it, and the caller may have more context on the inner type.)

bnjbvr (changes requested):
A few more remarks. It took me lot of time to read the widget-related code, as it's poorly documented, so let's try to improve the situation here: for each new function/type/trait you're introduced, or that you've touched in this PR, can you add a doc comment, please?

Would you also consider breaking the PR into smaller chunks, or smaller commits? The last one has quite the "draw the rest of the freaking howl" vibe :) and without a commit message or extra guidance where to start reading from, it's a bit overwhelming to review…

crates/matrix-sdk/src/widget/machine/from_widget.rs
```diff
@@ -174,22 +179,15 @@ pub(super) enum ApiVersion {
...
-    ReadMessageLikeEvent {
-        #[serde(rename = "type")]
-        event_type: MessageLikeEventType,
-        limit: Option<u32>,
-    },
+pub(super) struct ReadEventsRequest {
+    #[serde(rename = "type")]
+    pub(super) event_type: TimelineEventType,
+    pub(super) state_key: Option<StateKeySelector>,
+    pub(super) limit: Option<u32>,
```
bnjbvr, line 186: This will badly break the data format, and I suppose that's the reason why you're asking for a coordinated release; is there a chance we could support both the previous and new formats, instead?

crates/matrix-sdk/src/widget/machine/incoming.rs
```diff
@@ -49,23 +49,34 @@ pub(crate) enum IncomingMessage {
...
+    /// This means that the machine previously subscribed to some events
+    /// ([`crate::widget::Action::Subscribe`] request).
+    StateUpdateReceived(Vec<Raw<AnyStateEvent>>),
 }
 
 pub(crate) enum MatrixDriverResponse {
     /// Client acquired capabilities from the user.
-    ///
-    /// A response to an `Action::AcquireCapabilities` command.
+    /// A response to a `MatrixDriverRequestData::AcquireCapabilities` command.
```
bnjbvr, line 63: nit: here and below, use the cargo docs link notation, so we get links for free when rendered:

```rust
/// A response to a [`MatrixDriverRequestData::AcquireCapabilities`] command.
```

crates/matrix-sdk/src/widget/machine/mod.rs
```diff
@@ -314,83 +353,79 @@ impl WidgetMachine {
...
                             // content doesn't matter for state events
                             content: MatrixEventContent::default(),
                         };
 
                         capabilities.read.iter().any(|filter| filter.matches(&filter_in))
                     }
-                };
+                } || capabilities.read.iter().any(|f| {
+                    f.matches_message_like_event_type(&request.event_type.to_string().into())
+                });
```
bnjbvr, line 388: Can you use a local variable, or break the if into two ifs? using a match statement as a bool expression in a compound condition makes the code harder to read.

crates/matrix-sdk/src/widget/machine/mod.rs
```diff
@@ -616,10 +651,94 @@ impl WidgetMachine {
...
 
+                let state_filters: Vec<_> = approved
+                    .read
+                    .clone()
+                    .into_iter()
+                    .filter_map(|f| match f {
+                        EventFilter::State(f) => Some(f),
+                        _ => None,
+                    })
+                    .collect();
```
bnjbvr, line 662: You can avoid the preemptive cloning here:

```rust
                    .iter()
                    .filter_map(|f| match f {
                        EventFilter::State(f) => Some(f),
                        _ => None,
                    })
                    .cloned()
                    .collect();
```

crates/matrix-sdk/src/widget/machine/mod.rs
```diff
@@ -616,10 +651,94 @@ impl WidgetMachine {
...
-                let (_request, action) = machine.send_to_widget_request(update);
+                let (_request, notify_action) = machine.send_to_widget_request(update);
+
+                let (initial_state_pending, guard) = {
+                    let token = CancellationToken::new();
+                    (token.child_token(), token.drop_guard())
+                };
+
+                if !state_filters.is_empty() {
+                    machine.pending_state_updates = Some(InitialStateUpdate {
```
bnjbvr, line 673: Are we sure there's no previous `machine.pending_state_updates` here? If so, can you add an `assert!()` checking for this? Or a code comment explaining why it's safe to override, otherwise?

crates/matrix-sdk/src/widget/machine/mod.rs
```diff
@@ -616,10 +651,94 @@ impl WidgetMachine {
...
+                        initial_state: Vec::with_capacity(state_filters.len()),
+                        postponed_updates: Vec::new(),
+                        _drop_guard: guard,
+                    })
+                }
+
+                let request_count = state_filters.len();
+                let initial_state_actions = state_filters.iter().flat_map(|filter| {
```
bnjbvr, line 681: Please explain in a code comment what this is doing.

crates/matrix-sdk/src/widget/machine/mod.rs
```diff
@@ -616,10 +651,94 @@ impl WidgetMachine {
...
+                    action
+                });
 
-                subscribe_required.then_some(Action::Subscribe).into_iter().chain(action).collect()
+                subscribe_required
+                    .then_some(Action::Subscribe)
+                    .into_iter()
+                    .chain(initial_state_actions)
+                    .chain(notify_action)
+                    .collect()
```
bnjbvr, line 741: I know this is pre-existing, but this chaining of iterators is unnecessary complicated: instead, let's use a mutable vector of `Action` and push into it as we need to (like I've done in commits in #4966).

crates/matrix-sdk/src/widget/machine/mod.rs
```diff
@@ -121,6 +135,9 @@ pub(crate) struct WidgetMachine {
     /// Outstanding requests sent to the matrix driver (mapped by uuid).
     pending_matrix_driver_requests: PendingRequests<MatrixDriverRequestMeta>,
 
+    /// Outstanding state updates waiting to be sent to the widget.
+    pending_state_updates: Option<InitialStateUpdate>,
```
bnjbvr, line 139: Please explain in the doc comment a few more things: why is this needed? when is it set? when it's unset, will it stay unset forever, or can it be reset later?

crates/matrix-sdk/src/widget/matrix.rs
```diff
@@ -76,17 +83,30 @@ impl MatrixDriver {
         });
 
         let messages = self.room.messages(options).await?;
-        Ok(messages.chunk.into_iter().map(|ev| ev.into_raw().cast()).collect())
+        Ok(messages
+            .chunk
+            .into_iter()
+            .map(|ev| ev.into_raw().cast())
+            .filter(|ev| match &state_key {
```
bnjbvr, line 90: Any chance this can be fused into a `filter_map`?

crates/matrix-sdk/src/widget/matrix.rs
```diff
@@ -76,17 +83,30 @@ impl MatrixDriver {
...
+                Some(state_key) => {
+                    ev.get_field::<String>("state_key").is_ok_and(|k| match state_key {
```
bnjbvr, line 92: nit: please give a meaningful name to `k`

crates/matrix-sdk/tests/integration/widget.rs
```diff
@@ -181,6 +181,35 @@ async fn test_negotiate_capabilities_immediately() {
...
+            "body": "hello",
+            "msgtype": "m.text",
+        },
+        "event_id": "$msda7m0df9E9op3",
+        "origin_server_ts": 152037280,
+        "sender": "@example:localhost",
+        "type": "m.room.message",
+        "room_id": &*ROOM_ID,
+    })
+});
```
bnjbvr, line 196: This looks like this is code that's been just moved, so a good candidate for a separate commit with only code motion.

bnjbvr (dismissed):
(Moar review comments, now that I'm mostly understanding the flow between Machine/Widget/Driver)

crates/matrix-sdk/src/widget/matrix.rs
robintown: I could, though what would be the benefit? My understanding is that operations on Rust iterators are fused automatically.

testing/matrix-sdk-test/src/event_factory.rs
```diff
@@ -420,8 +410,10 @@ impl EventFactory {
             is_ephemeral: false,
             room: self.room.clone(),
             server_ts: self.next_server_ts(),
-            event_id: None,
-            no_event_id: false,
+            event_id: Some(match &self.room {
+                Some(room_id) => EventId::new(room_id.server_name().unwrap()),
+                None => EventId::new(server_name!("dummy.org")),
+            }),
```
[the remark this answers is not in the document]
robintown: I see, I missed that the `room` field of an EventBuilder is mutable. I wanted to be able to clone EventBuilders so that I could convert the same event to a `Raw<AnySyncStateEvent>` and also a `Raw<AnyStateEvent>`, which must have the same event ID but lack the `room_id` attribute, so I could [compare them in tests](https://github.com/matrix-org/matrix-rust-sdk/blob/089c552122eb705435cb895ebb3f7b45c0f6e451/crates/matrix-sdk/tests/integration/widget.rs#L543).

But if they can't be made cloneable, I can work around this with a bit more type-casting and verbose assertions, I guess.

crates/matrix-sdk/src/widget/matrix.rs
```diff
@@ -76,17 +84,34 @@ impl MatrixDriver {
         });
 
         let messages = self.room.messages(options).await?;
-        Ok(messages.chunk.into_iter().map(|ev| ev.into_raw().cast()).collect())
+        Ok(messages
```
poljar, line 87:
```suggestion
        
        Ok(messages
```

poljar (changes requested):
Alright, I think most of this is good. The one thing that I'm objecting to mostly is the complexity of the capability negotiation.

The biggest issue is the naked unwrap in the mentioned function.

crates/matrix-sdk/src/widget/machine/mod.rs
```diff
@@ -632,14 +672,109 @@ impl WidgetMachine {
...
+                            }
+                        })
+                        .map(|(request, action)| {
+                            request.then(move |result, machine| {
+                                // No-op if the initial state update was canceled
+                                if initial_state_pending.is_cancelled() {
+                                    Vec::new()
+                                } else {
+                                    let mut updates =
+                                        mem::take(&mut machine.pending_state_updates).unwrap();
```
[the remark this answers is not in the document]
robintown: I think I have a genuine need to use a `CancellationToken` in addition to the `Option` and also to `take` the state updates and put them back. I've written a couple more comments, which for quick reference are:

```rust
                        .map(|(request, action)| {
                            request.then(move |result, machine| {
                                // No-op if the initial state update was canceled.
                                // N.b. it would not suffice to just match on
                                // `machine.pending_state_updates` being `Some`/`None` here, because
                                // it's possible for a capability renegotiation to happen while an
                                // earlier initial state update was in progress, and we don't want
                                // to assign this state request's results to the wrong update (as it
                                // would race/duplicate one of the requests from the renegotiation).
                                if initial_state_pending.is_cancelled() {
                                    Vec::new()
                                } else {
                                    // Pull the updates struct out of the machine temporarily so
                                    // that we can unwrap it in one place, mutate it, and still be
                                    // able to call `send_to_widget_request` later in this block
                                    // (which borrows the machine mutably)
                                    let mut updates =
                                        mem::take(&mut machine.pending_state_updates).expect("initial state updates must be canceled when set to `None`");
```

crates/matrix-sdk/src/widget/machine/driver_req.rs
```diff
@@ -50,7 +50,7 @@ pub(crate) enum MatrixDriverRequestData {
     ReadStateEvent(ReadStateEventRequest),
 
     /// Send matrix event that corresponds to the given description.
-    SendMatrixEvent(SendEventRequest),
+    SendEvent(SendEventRequest),
```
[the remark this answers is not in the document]
robintown: There's nothing else that we're supposed to be calling an "event", to my knowledge (I've commented on the linked PR)

crates/matrix-sdk/src/widget/machine/driver_req.rs
```diff
@@ -152,26 +152,32 @@ impl FromMatrixDriverResponse for request_openid_token::v3::Response {
     }
 }
 
-/// Ask the client to read matrix event(s) that corresponds to the given
+/// Ask the client to read matrix events that correspond to the given
 /// description and return a list of events as a response.
 #[derive(Clone, Debug)]
-pub(crate) struct ReadMessageLikeEventRequest {
+pub(crate) struct ReadEventsRequest {
```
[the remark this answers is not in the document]
robintown: OK, noted for next time.

crates/matrix-sdk/src/widget/machine/mod.rs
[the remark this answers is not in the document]
poljar: > I think I have a genuine need to use a CancellationToken in addition to the Option 

Hmm, I can't figure out why.

> // it's possible for a capability renegotiation to happen while an
  // earlier initial state update was in progress

But that would need to borrow the `WidgetMachine` and the `pending_state_updates` as well? Which can't happen until our first borrow is done. The `response_handler` callback isn't an async function so there's nothing funky going on with the borrow either.

Isn't the mutable borrow enough to protect us from other pending state updates?

For what is worth, this diff doesn't seem to trigger any test failures:


```diff
diff --git a/crates/matrix-sdk/src/widget/machine/mod.rs b/crates/matrix-sdk/src/widget/machine/mod.rs
index 2833db1b1..d34c3647d 100644
--- a/crates/matrix-sdk/src/widget/machine/mod.rs
+++ b/crates/matrix-sdk/src/widget/machine/mod.rs
@@ -14,7 +14,7 @@
 
 //! No I/O logic of the [`WidgetDriver`].
 
-use std::{iter, mem, time::Duration};
+use std::{iter, time::Duration};
 
 use driver_req::{ReadStateRequest, UpdateDelayedEventRequest};
 use from_widget::UpdateDelayedEventResponse;
@@ -26,7 +26,6 @@ use ruma::{
 };
 use serde::Serialize;
 use serde_json::value::RawValue as RawJsonValue;
-use tokio_util::sync::{CancellationToken, DropGuard};
 use tracing::{error, info, instrument, warn};
 use uuid::Uuid;
 
@@ -112,8 +111,6 @@ struct InitialStateUpdate {
     /// read the initial state. These should be pushed to the widget immediately
     /// after pushing the initial state to ensure no data is lost.
     postponed_updates: Vec<Vec<Raw<AnyStateEvent>>>,
-    /// Guard which cancels all requests to populate this initial state on drop.
-    _drop_guard: DropGuard,
 }
 
 /// No I/O state machine.
@@ -682,11 +679,6 @@ impl WidgetMachine {
                     .cloned()
                     .collect();
 
-                let (initial_state_pending, guard) = {
-                    let token = CancellationToken::new();
-                    (token.child_token(), token.drop_guard())
-                };
-
                 if !state_filters.is_empty() {
                     // Begin accumulating the initial state to be pushed to the widget. Note that we
                     // don't care what the previous value of `pending_state_updates` was; we
@@ -698,7 +690,6 @@ impl WidgetMachine {
                     machine.pending_state_updates = Some(InitialStateUpdate {
                         initial_state: Vec::with_capacity(state_filters.len()),
                         postponed_updates: Vec::new(),
-                        _drop_guard: guard,
                     })
                 }
 
@@ -707,7 +698,6 @@ impl WidgetMachine {
                 // room state entries
                 let request_count = state_filters.len();
                 actions.extend(state_filters.iter().flat_map(|filter| {
-                    let initial_state_pending = initial_state_pending.child_token();
                     machine
                         .send_matrix_driver_request(match filter {
                             StateEventFilter::WithType(event_type) => ReadStateRequest {
@@ -723,16 +713,12 @@ impl WidgetMachine {
                         })
                         .map(|(request, action)| {
                             request.then(move |result, machine| {
-                                // No-op if the initial state update was canceled
-                                if initial_state_pending.is_cancelled() {
-                                    Vec::new()
-                                } else {
-                                    let mut updates =
-                                        mem::take(&mut machine.pending_state_updates).unwrap();
+                                if let Some(mut updates) = machine.pending_state_updates.take() {
                                     updates.initial_state.push(result.unwrap_or_else(|e| {
                                         error!("Reading initial room state failed: {e}");
                                         Vec::new()
                                     }));
+
                                     // Once the number of initial state responses reaches
                                     // `request_count`, the initial state is complete; combine the
                                     // data and push it to the widget in a single action.
@@ -762,6 +748,8 @@ impl WidgetMachine {
                                         machine.pending_state_updates = Some(updates);
                                         Vec::new()
                                     }
+                                } else {
+                                    Vec::new()
                                 }
                             });
 
```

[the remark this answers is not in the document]
robintown: Well, in the meantime I noticed that the widget driver *doesn't* actually implement capability renegotiation ([MSC2974](https://github.com/matrix-org/matrix-spec-proposals/pull/2974)) :facepalm: so it's indeed pointless to have that extra stuff with the `DropGuard` as it can't yet be tested.

I took your suggested diff with the exception that I'm still calling `.expect()` on the updates struct, since it's a logical error for it to be `None` while there are still state requests in flight, and I'd rather have it fail loudly if we break that invariant.

crates/matrix-sdk/src/widget/machine/driver_req.rs
```diff
@@ -152,26 +152,32 @@ impl FromMatrixDriverResponse for request_openid_token::v3::Response {
...
     /// The event type to read.
     // TODO: This wants to be `MessageLikeEventType`` but we need a type which supports `as_str()`
     // as soon as ruma supports `as_str()` on `MessageLikeEventType` we can use it here.
     pub(crate) event_type: String,
 
+    /// The `state_key` to read. If None, this will read events regardless of
+    /// whether they are state events. If `Some(Any)`, this will only read state
+    /// events of the given type. If set to a specific state key, this will only
+    /// read state events of the given type matching that state key.
+    pub(crate) state_key: Option<StateKeySelector>,
```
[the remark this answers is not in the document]
robintown: I'll bring this up on the MSC. I think the use case for reading *all* events of a given type, state event or not, is actually pretty hard to imagine, and so maybe it's best to repurpose the `None` state here to represent a message-like event filter as we had before.

This does need an update to the MSC (which can probably be phrased as a "clarification") so for now I'd like to align this with what Element Web and matrix-widget-api are doing.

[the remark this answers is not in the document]
robintown: https://github.com/matrix-org/matrix-spec-proposals/pull/4288

crates/matrix-sdk/src/widget/machine/mod.rs
```diff
@@ -595,62 +647,169 @@ impl WidgetMachine {
...
+    /// [`NotifyCapabilitiesChanged`] `toWidget` action to tell the widget
+    /// which capabilities were approved.
+    fn negotiate_capabilities(&mut self) -> Vec<Action> {
+        let mut actions = Vec::new();
 
-                let notify_caps_changed = NotifyCapabilitiesChanged {
-                    approved: approved_capabilities,
-                    requested: requested_capabilities,
-                };
+        // XXX This branch appears to be accounting for capability **re**negotiation
```
poljar, line 798:
```suggestion
        // XXX: This branch appears to be accounting for capability **re**negotiation
```

poljar (changes requested):
Ok, I think there's still a bit that can be improved as part of this PR.

I think after this we're good. Please don't use force pushes once a review has started. It makes it quite hard to see the changes between the review rounds.

If you'd like to keep the history clean you can use `fixup` commits and rebase/autosquash once the PR has been approved.

More info about our review process can be found in the contributing guide: https://github.com/matrix-org/matrix-rust-sdk/blob/main/CONTRIBUTING.md#review-process.

crates/matrix-sdk/src/widget/machine/mod.rs
```diff
@@ -595,62 +647,169 @@ impl WidgetMachine {
...
+        &mut self,
+        events: Result<Vec<Raw<AnyStateEvent>>, Error>,
+    ) -> Vec<Action> {
+        // Pull the updates struct out of the machine temporarily so that we can unwrap
+        // it in one place, mutate it, and still be able to call
+        // `send_to_widget_request` later in this block (which borrows the
+        // machine mutably)
+        let mut updates = mem::take(&mut self.pending_state_updates).expect(
+            "initial state updates must only be set to `None` once all requests are complete",
+        );
```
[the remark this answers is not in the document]
robintown: OK, I can turn both panics into `error!` logs and then I suppose we're on the safe side.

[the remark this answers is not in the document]
poljar: To be clear, you don't need to do this to get this merged, it's mostly me just grumbling here.

Though if you agree that it sounds like it could potentially become dangerous, here's a diff I created when trying to figure out what was going on:

```diff
diff --git a/crates/matrix-sdk/src/widget/machine/mod.rs b/crates/matrix-sdk/src/widget/machine/mod.rs
index 29130a623..ef6c3df04 100644
--- a/crates/matrix-sdk/src/widget/machine/mod.rs
+++ b/crates/matrix-sdk/src/widget/machine/mod.rs
@@ -14,7 +14,7 @@
 
 //! No I/O logic of the [`WidgetDriver`].
 
-use std::{iter, mem, time::Duration};
+use std::time::Duration;
 
 use driver_req::{ReadStateRequest, UpdateDelayedEventRequest};
 use from_widget::UpdateDelayedEventResponse;
@@ -653,40 +653,60 @@ impl WidgetMachine {
     /// [`NotifyStateUpdate`] action.
     fn process_read_initial_state_response(
         &mut self,
-        events: Result<Vec<Raw<AnyStateEvent>>, Error>,
+        events: Vec<Raw<AnyStateEvent>>,
     ) -> Vec<Action> {
         // Pull the updates struct out of the machine temporarily so that we can unwrap
         // it in one place, mutate it, and still be able to call
-        // `send_to_widget_request` later in this block (which borrows the
-        // machine mutably)
-        let mut updates = mem::take(&mut self.pending_state_updates).expect(
-            "initial state updates must only be set to `None` once all requests are complete",
-        );
-        updates.initial_state.push(events.unwrap_or_else(|e| {
-            error!("Reading initial room state failed: {e}");
-            Vec::new()
-        }));
-        // Once the number of initial state responses reaches `updates.request_count`,
-        // the initial state is complete; combine the data and push it to the
-        // widget in a single action.
-        if updates.initial_state.len() == updates.request_count {
-            iter::once(
-                self.send_to_widget_request(NotifyStateUpdate {
-                    state: updates.initial_state.into_iter().flatten().collect(),
-                })
-                .map(|(_request, action)| action),
-            )
-            // Also flush any state updates that had been postponed until after the initial state
-            // push
-            .chain(updates.postponed_updates.into_iter().map(|state| {
-                self.send_to_widget_request(NotifyStateUpdate { state })
-                    .map(|(_request, action)| action)
-            }))
-            .flatten()
-            .collect()
+        // `send_to_widget_request` later in this block (which borrows the machine
+        // mutably)
+
+        fn state_events_to_action(
+            machine: &mut WidgetMachine,
+            state_events: Vec<Raw<AnyStateEvent>>,
+        ) -> Option<Action> {
+            machine
+                .send_to_widget_request(NotifyStateUpdate { state: state_events })
+                .map(|(_request, action)| action)
+        }
+
+        let result = if let Some(updates) = &mut self.pending_state_updates {
+            updates.initial_state.push(events);
+
+            // Once the number of initial state responses reaches `updates.request_count`,
+            // the initial state is complete; combine the data and push it to the
+            // widget in a single action.
+            if updates.initial_state.len() == updates.request_count {
+                // Gather the initial events.
+                // TODO: Why the flatten? Why are the initial events flattened out before
+                // the conversion into an action, but the postponed events afterwards?
+                let initial: Vec<_> = updates.initial_state.drain(0..).flatten().collect();
+
+                // Also flush any state updates that had been postponed until after the
+                // initial state push
+                let postponed: Vec<_> = updates.postponed_updates.drain(0..).collect();
+
+                // Now convert the two sets of state events into actions.
+                let initial = state_events_to_action(self, initial);
+                let postponed =
+                    postponed.into_iter().map(|events| state_events_to_action(self, events));
+
+                // Combine them before we return them.
+                let combined: Vec<_> = initial.into_iter().chain(postponed.flatten()).collect();
+
+                Some(combined)
+            } else {
+                None
+            }
+        } else {
+            None
+        };
+
+        // If we're returning the actions, then we're done with the
+        // `pending_state_updates` and we can set them to `None`.
+        if let Some(result) = result {
+            self.pending_state_updates.take();
+            result
         } else {
-            // Continue accumulating the initial state
-            self.pending_state_updates = Some(updates);
             Vec::new()
         }
     }
@@ -752,7 +772,7 @@ impl WidgetMachine {
             })
             .map(|(request, action)| {
                 request.then(move |result, machine| {
-                    machine.process_read_initial_state_response(result)
+                    machine.process_read_initial_state_response(result.unwrap_or_default())
                 });
                 action
             })
```

I would probably go even a bit further, return `None` when we're still collecting things, return `Some(actions)` when we're done collecting things.

That would clear it up further, compared to an empty `Vec` as the "We're still not done" return value.

robintown:
> Please don't use force pushes once a review has started. It makes it quite hard to see the changes between the review rounds. If you'd like to keep the history clean you can use `fixup` commits and rebase/autosquash once the PR has been approved.

Noted. I had to resolve some merge conflicts so I haven't used `fixup` commits for this most recent round of changes, but if you would prefer I absorb them into the previous commits once this is approved, I could do that. (Just means I have to resolve the conflicts a second time.)

crates/matrix-sdk/src/widget/machine/mod.rs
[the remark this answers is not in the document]
robintown: 25e2789ef280d93ed9760737f7aa566db05aac86
6f507ed8cbc0393568ea951a66fdcabd587c124c

poljar:
> > Please don't use force pushes once a review has started. It makes it quite hard to see the changes between the review rounds. If you'd like to keep the history clean you can use `fixup` commits and rebase/autosquash once the PR has been approved.
> 
> Noted. I had to resolve some merge conflicts so I haven't used `fixup` commits for this most recent round of changes, but if you would prefer I absorb them into the previous commits once this is approved, I could do that. (Just means I have to resolve the conflicts a second time, unless you know of another way.)

It's fine, I'll take a look tomorrow.

poljar (approved):
Alright, I think that this is fine.

You can now get rid of the fixup commits and rebase, or if you don't want to spend time on this we can squash merge as well.

robintown:
Thanks a bunch for your thorough reviews! As a reminder, this still needs to wait on some Element Call release candidate stuff before it's ready for merging.

robintown:
We landed a [fallback](https://github.com/element-hq/element-call/pull/3277) for widget drivers that don't support to-device messages, so this is unblocked

robintown:
(rebased to resolve conflicts)

robintown:
@poljar Now that the necessary changes to Element Call [have been released](https://github.com/element-hq/element-call/releases/tag/v0.12.0), this is ready to be merged. (Oh and the Complement crypto test failure looks like an unrelated flake, needs a re-run I guess.) Thank you :pray:

poljar:
Hmm, doesn't look like a flake. You might need to rebase again to get a matching complement-crypto version that works with the current state of the SDK.

Issue #4945, opened 2025-04-16
test_history_share_on_invite fails intermitently


poljar:
Instance of this can be found here: https://github.com/matrix-org/matrix-rust-sdk/actions/runs/14493485258/job/40655224967?pr=4944#step:10:2603

bnjbvr:
https://github.com/matrix-org/matrix-rust-sdk/actions/runs/14589790091/job/40922205596#step:10:2783

PR #4948, opened 2025-04-16
FFI: Allow file data to be passed through bindings when sending attachment

mgoldenberg:
### Problem

`matrix_sdk_ffi` exposes functions for sending attachments through the [`Timeline`](https://github.com/matrix-org/matrix-rust-sdk/blob/bc50cae35fd32b902eac4d86d61ebe2ee7e11df2/bindings/matrix-sdk-ffi/src/timeline/mod.rs#L93). These functions, however, only allow the foreign language to provide the name of a file (see [UploadParameters](https://github.com/matrix-org/matrix-rust-sdk/blob/bc50cae35fd32b902eac4d86d61ebe2ee7e11df2/bindings/matrix-sdk-ffi/src/timeline/mod.rs#L203)) and require that the linked Rust library actually handle reading the files.

### Proposal

Add functionality to make it possible for a foreign language to read in a file's contents natively and then upload those contents through the `Timeline`.  

### Implementation

Currently, `UploadParameters.filename` is eventually transformed into an [`AttachmentSource::File`](https://github.com/matrix-org/matrix-rust-sdk/blob/bc50cae35fd32b902eac4d86d61ebe2ee7e11df2/crates/matrix-sdk-ui/src/timeline/mod.rs#L739). The addition of an extra field, `file_data: Option<Vec<u8>>`, allows mapping `UploadParameters` into an [`AttachmentSource::Data`](https://github.com/matrix-org/matrix-rust-sdk/blob/bc50cae35fd32b902eac4d86d61ebe2ee7e11df2/crates/matrix-sdk-ui/src/timeline/mod.rs#L728) when `file_data` is present.

#### Alternative

Another option would be to add a wrapped `AttachmentSource` to `matrix_sdk_ffi`, expose it through UniFFI, and replace `UploadParameters.filename` with `UploadParameters.source`. 

#### Breaking Changes

Both implementations mentioned result in breaking changes. I opted for the current implementation, as opposed to the alternative mentioned above because, I believe, there are some cases where the current implementation does not result in a breaking change, as it is simply adding an optional field. On the other hand, the alternative implementation would result in a breaking change in all cases as it both removes a field and adds a field.

#### Tests

I tested these changes by generating Python bindings and running some of the `Timeline` functions, as I wasn't able to find any unit tests or integration tests against the bindings - I only found tests ensuring the bindings would be generated.

Are there tests to ensure the bindings work as they should? Or is this not of concern within this repository?

---

- [x] Public API changes documented in changelogs

### DIFF
--- a/bindings/matrix-sdk-ffi/src/timeline/mod.rs
+++ b/bindings/matrix-sdk-ffi/src/timeline/mod.rs
@@ -130,10 +130,11 @@ impl Timeline {
             .reply(params.reply_params.map(|p| p.try_into()).transpose()?);
 
         let handle = SendAttachmentJoinHandle::new(get_runtime_handle().spawn(async move {
-            let source = if let Some(bytes) = params.file_data {
-                AttachmentSource::Data { bytes, filename: params.filename }
-            } else {
-                AttachmentSource::File(params.filename.into())
+            let source = match params.source {
+                UploadSource::Data { data, filename } => {
+                    AttachmentSource::Data { bytes: data, filename }
+                }
+                UploadSource::File { filename } => AttachmentSource::File(filename.into()),
             };
             let mut request = self.inner.send_attachment(source, mime_type, attachment_config);
 
@@ -205,10 +206,8 @@ fn build_thumbnail_info(
 
 #[derive(uniffi::Record)]
 pub struct UploadParameters {
-    /// Filename (previously called "url") for the media to be sent.
-    filename: String,
-    /// Optional file data - if not present, then data is read from `filename`
-    file_data: Option<Vec<u8>>,
+    /// Source from which to upload data
+    source: UploadSource,
     /// Optional non-formatted caption, for clients that support it.
     caption: Option<String>,
     /// Optional HTML-formatted caption, for clients that support it.
--- a/bindings/matrix-sdk-ffi/src/timeline/mod.rs
+++ b/bindings/matrix-sdk-ffi/src/timeline/mod.rs
@@ -227,9 +227,9 @@ pub enum UploadSource {
     },
     /// Upload source is data in memory
     Data {
-        /// Data being uploaded
-        data: Vec<u8>,
-        /// Filename to associate with data
+        /// Bytes being uploaded
+        bytes: Vec<u8>,
+        /// Filename to associate with bytes
         filename: String,
     },
 }
@@ -238,7 +238,7 @@ impl From<UploadSource> for AttachmentSource {
     fn from(value: UploadSource) -> Self {
         match value {
             UploadSource::File { filename } => Self::File(filename.into()),
-            UploadSource::Data { data, filename } => Self::Data { bytes: data, filename },
+            UploadSource::Data { bytes, filename } => Self::Data { bytes, filename },
         }
     }
 }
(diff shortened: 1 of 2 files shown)


mgoldenberg:
Hmm, looks like the following test is failing: [`tests::e2ee::test_history_share_on_invite`](https://github.com/matrix-org/matrix-rust-sdk/blob/bc50cae35fd32b902eac4d86d61ebe2ee7e11df2/testing/matrix-sdk-integration-testing/src/tests/e2ee.rs#L1187) in `matrix-sdk-integration-testing`.

I'm not sure this has to do with my changes though, as `matrix-sdk-integration-testing` does not depend on `matrix-sdk-ffi`... In any case, I already tried running this test locally and it seems to work.

Any thoughts on why this might not be passing? Perhaps a time-related issue?

bindings/matrix-sdk-ffi/src/timeline/mod.rs
```diff
@@ -203,6 +207,8 @@ fn build_thumbnail_info(
 pub struct UploadParameters {
     /// Filename (previously called "url") for the media to be sent.
     filename: String,
+    /// Optional file data - if not present, then data is read from `filename`
```
zzorba, line 210: So this might be slightly more idiomatic as a new enum with a `Filename` and `Data` option.  It would be a 'breaking' change for the API, but is more in keeping with the way the FFI is generally structured.

mgoldenberg: I'm definitely open to it! I partly avoided this because wrapping `AttachmentSource` seemed a bit esoteric and duplicating it seemed a bit crude. Of course, what I have here is also rather crude 😅

poljar:
> I'm not sure this has to do with my changes though, as `matrix-sdk-integration-testing` does not depend on `matrix-sdk-ffi`...

It doesn't it was intermittently failing on some other PRs as well. 

> In any case, I also tried running this test locally and it seems to work.
> 
> Any thoughts on why this might not be passing? Perhaps a time-related issue?

It's a timing-based issue, it works locally because your computer is much faster than what is running the CI.

I have opened https://github.com/matrix-org/matrix-rust-sdk/issues/4945 to track and eventually address this.

poljar:
The bindings are mostly tested with [complement-crypto](https://github.com/matrix-org/complement-crypto/), while it has crypto in its name, it tests that the bindings can send/receive messages using a real homeserver.

bindings/matrix-sdk-ffi/src/timeline/mod.rs
poljar: I agree that the enum is the better approach, can we have that please?

poljar (changes requested):

bindings/matrix-sdk-ffi/src/timeline/mod.rs
mgoldenberg: No problem, I will update it!

poljar: Thanks, and sorry for the delay, somehow this PR fell through the cracks.

zzorba:
Looks good from my pov, we might eventually need a cfg guard on the filename enum variant for wasm32 target, but that's definitely something for a future PR.

mgoldenberg:
> Looks good from my pov, we might eventually need a cfg guard on the filename enum variant for wasm32 target, but that's definitely something for a future PR.

Agree!

poljar (approved):
Alright, this looks good. Thanks.

PR #4953, opened 2025-04-17
fix(sdk): Don't overwrite previously added state events for this room in state_event sync processing

### LINKED ISSUE #4952: regular sync: power levels and other state events are missing
Was trying to upgrade to latest `main` when I noticed that asking any room for `power_levels` comes back with `InsufficientData`. We are using regular old sync (are in the process of migrating to sliding sync) and it appears that on latest `main` the state events aren't saved to the state_store anymore.

Some git bisecting points me to db6d1b4cd6893912387867831f069cdc712034a8 being the commit it all breaks with. 

After a brief introspection it seems that there is no new equivalent of the removed state-events adding for non-user-events: https://github.com/matrix-org/matrix-rust-sdk/commit/db6d1b4cd6893912387867831f069cdc712034a8#diff-c559762e72a5a86b43382e9ebb7f1bf5d8144f409c1789ad5bb4294eabf44d07L437-L441 

I'll investigate further and see if I can come up with a patch.

gnunicorn:
Fixes #4952 .

Signed-off-by: Benjamin Kampmann <ben@acter.global>

### DIFF
--- a/crates/matrix-sdk-base/src/response_processors/state_events.rs
+++ b/crates/matrix-sdk-base/src/response_processors/state_events.rs
@@ -20,10 +20,7 @@ use super::Context;
 
 /// Collect [`AnySyncStateEvent`].
 pub mod sync {
-    use std::{
-        collections::{BTreeMap, BTreeSet},
-        iter,
-    };
+    use std::{collections::BTreeSet, iter};
 
     use ruma::{
         events::{room::member::MembershipState, AnySyncTimelineEvent},
@@ -81,8 +78,6 @@ pub mod sync {
             return Ok(user_ids);
         }
 
-        let mut state_events = BTreeMap::new();
-
         for (raw_event, event) in iter::zip(raw_events, events) {
             room_info.handle_state_event(event);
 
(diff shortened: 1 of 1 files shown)


gnunicorn:
looks like I have another failure. redoing the change.

gnunicorn:
Okay, this does seem to do the trick now, actually not overwriting the wrong items.

bnjbvr (changes requested):
Thanks, looks sensible; can you write a regression test please?

crates/matrix-sdk-base/src/response_processors/state_events.rs
```diff
@@ -216,3 +213,30 @@ where
...
+    use crate::test_utils::logged_in_base_client;
+
+    #[async_test]
+    async fn test_state_events_after_sync() {
+        // Given a room
+        let user_id = user_id!("@u:u.to");
+
+        let client = logged_in_base_client(Some(user_id)).await;
+
+        let response = sync_response(SyncResponseFile::DefaultWithStateInTimeline);
```
bnjbvr, line 231: Ouch, we try to not use the `SyncResponseFile` anymore, and instead use the glorious `EventFactory`, which makes it more ergonomic to create smaller sync responses. See for instance some usage in tests from this module: https://github.com/matrix-org/matrix-rust-sdk/blob/a81ac262d681fb77444ec94e6c4e9fe03bd794bf/crates/matrix-sdk-base/src/read_receipts.rs#L643-L656

And then let's get rid of the new `DefaultWithStateInTimeline`.

bnjbvr (changes requested):
Thanks; let's try to not use the `SyncResponseFile` anymore (see #3716 for context).

crates/matrix-sdk-base/src/response_processors/state_events.rs
gnunicorn: ack. I've replaced the last commit with a new one in this format.

bnjbvr (approved):
Nice, thanks!

Issue #2729, opened 2023-10-18
Make `collect_session_recipients` more resistent


BillCarsonFr:
Currently when `collect_session_recipients` is figuring out to which devices a key was distributed it only looks at `shared_with_set` and not `to_share_with_set` (inflight).

That means that if the caller calls share_room_key, but fails to distribute the to_device, at the next call the device will not be seen has having the key but will eventually get it after ~~next call that processes outgoing requests~~ next message is sent in room.

As an example there was a bug in webR, where to-device was [not marked as sent](https://github.com/matrix-org/matrix-rust-sdk-crypto-wasm/pull/41).
This leads to several strange side effect. 
For example is you change the `only_allow_trusted_devices` setting to true (with an unverified device in the room). Then collect session recipient won't rotate the megolm session, as the session as not been delivered (not in `shared_with_set`, only in `to_share_with_set`).

As a result the sdk will give you 2 to-devices, a room key for the device and a withheld for the same device and same session.

We could make the sdk more resistent by considering both`shared_with_set` and `to_share_with_set`

poljar:
Closed by #4975.

PR #4954, opened 2025-04-18
Crypto: Fixed a bug where room keys would be rotated unecessarily in the presence of blacklisted/withheld devices

procr1337:
Previously, `is_session_overshared_for_user` did not take into account that `shared_with_set` also contains withheld device IDs who explicitly have never received the session keys. This would lead to it mistakenly determining oversharing for those devices for every event being sent in the presence of blacklisted/withheld devices in the room, and rotating the group session accordingly.

The fix is to correctly exclude devices with `ShareInfo::Withheld` from the enumeration.

Signed-off-by: Niklas Baumstark niklas.baumstark@gmail.com

### DIFF
--- a/crates/matrix-sdk-crypto/CHANGELOG.md
+++ b/crates/matrix-sdk-crypto/CHANGELOG.md
@@ -16,6 +16,10 @@ All notable changes to this project will be documented in this file.
   (as per [MSC4147](https://github.com/matrix-org/matrix-spec-proposals/pull/4147)).
   ([#4922](https://github.com/matrix-org/matrix-rust-sdk/pull/4922))
 
+- Fix bug which caused room keys to be unnecessarily rotated on every send in the
+  presence of blacklisted/withheld devices in the room.
+  ([#4954](https://github.com/matrix-org/matrix-rust-sdk/pull/4954))
+
 ## [0.11.0] - 2025-04-11
 
 ### Features
--- a/crates/matrix-sdk-crypto/src/olm/group_sessions/outbound.rs
+++ b/crates/matrix-sdk-crypto/src/olm/group_sessions/outbound.rs
@@ -160,7 +160,7 @@ pub struct OutboundGroupSession {
     pub(crate) shared_with_set:
         Arc<StdRwLock<BTreeMap<OwnedUserId, BTreeMap<OwnedDeviceId, ShareInfo>>>>,
     #[allow(clippy::type_complexity)]
-    pub(crate) to_share_with_set:
+    to_share_with_set:
         Arc<StdRwLock<BTreeMap<OwnedTransactionId, (Arc<ToDeviceRequest>, ShareInfoSet)>>>,
 }
 
--- a/crates/matrix-sdk-crypto/src/olm/group_sessions/outbound.rs
+++ b/crates/matrix-sdk-crypto/src/olm/group_sessions/outbound.rs
@@ -160,7 +160,7 @@ pub struct OutboundGroupSession {
     pub(crate) shared_with_set:
         Arc<StdRwLock<BTreeMap<OwnedUserId, BTreeMap<OwnedDeviceId, ShareInfo>>>>,
     #[allow(clippy::type_complexity)]
-    to_share_with_set:
+    pub(crate) to_share_with_set:
         Arc<StdRwLock<BTreeMap<OwnedTransactionId, (Arc<ToDeviceRequest>, ShareInfoSet)>>>,
 }
 
--- a/crates/matrix-sdk-crypto/src/session_manager/group_sessions/share_strategy.rs
+++ b/crates/matrix-sdk-crypto/src/session_manager/group_sessions/share_strategy.rs
@@ -427,22 +427,49 @@ fn is_session_overshared_for_user(
     let recipient_device_ids: BTreeSet<&DeviceId> =
         recipient_devices.iter().map(|d| d.device_id()).collect();
 
+    let mut shared: Vec<&DeviceId> = Vec::new();
+
+    // This duplicates a conservative subset of the logic in OutbundGroupSession::is_shared_with,
+    // because we don't have corresponding DeviceData at hand
+    fn is_actually_shared(info: &ShareInfo) -> bool {
+        match info {
+            ShareInfo::Shared(_) => true,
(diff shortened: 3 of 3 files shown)


procr1337:
I don't really understand the failing test: It seems to withhold room keys for Dan's devices because of `m.no_olm`, so it doesn't actually share the keys with Dan, and hence correctly determines that no rotation is necessary after Dan's second devices logs out.

The correct behavior should be to actually do share the keys with both devices first, so that a rotation is in fact necessary.

procr1337:
https://github.com/matrix-org/matrix-rust-sdk/issues/2729 (which is kind of the opposite of the behavior I was trying to fix here) would indicate that there are probably other considerations about which devices may actually eventually get the key, e.g. `to_share_with_set`

poljar (changes requested):
> I don't really understand the failing test: It seems to withhold room keys for Dan's devices because of `m.no_olm`, so it doesn't actually share the keys with Dan, and hence correctly determines that no rotation is necessary after Dan's second device logs out.

I think the test might have been written before the support for the `m.room_key.withheld` messages was added. It therefore didn't set up the required Olm sessions for the room key to be shared even though it obviously wanted Dan's devices to be in the shared set.

To fix the test we'll probably want to rewrite the test so it doesn't call `share_room_key()` and `mark_request_as_sent()` to modify the state of the outbound group session. Instead we can just call `OutboundGroupSession::mark_shared_with()`. This won't require any setup of the Olm sessions.

The bug you discovered was likely introduced with the `.withheld` messages as well.

I think that this PR makes sense, could you just fix the failing test and introduce a regression test which checks the case where we don't want to rotate as well.

procr1337:
@poljar Thank you for your feedback. Added the two tests. 

I believe there was an additional bug in the previous implementation for not taking into account pending to-device requests. I can't prove to myself why it should not be possible for pending requests to exist before share_room_key is called, especially because it specifically handles this case after calling `collect_session_recipients`. I could be wrong and be missing another invariant at the API boundaries here.

There is some potential for refactoring, especially to properly encapsulate the `to_share_with_set` field of OutboundGroupSession, and also to make `OutboundGroupSession::is_shared_with` accessible without full `DeviceData`

procr1337:
On sceond read, I believe the second bug is pretty much exactly #2729, so this should potentially be a fix for that IIUC

poljar:
> @poljar Thank you for your feedback. Added the two tests.
> 
> I believe there was an additional bug in the previous implementation for not taking into account pending to-device requests. I can't prove to myself why it should not be possible for pending requests to exist before share_room_key is called, especially because it specifically handles this case after calling `collect_session_recipients`.

They can exists if you call `share_room_key()` multiple times, this is what https://github.com/matrix-org/matrix-rust-sdk/issues/2729 is about.

They can't exists if you  call `share_room_key()` the first time. This can bee seen if you inspect how the pending requests are added.

The way to add the pending requests is to call `OutboundGroupSession::add_request()`. There are exactly two places where this is happening, once for the withheld to-device requests and once for the requests containing the room key.

Both of those cases happen after the call to `collect_session_recipients()`.

> I could be wrong and be missing another invariant at the API boundaries here.

You are right. The reason why this works is that we discard the room key if we don't manage to mark all the pending requests as sent:

https://github.com/matrix-org/matrix-rust-sdk/blob/a60e336f8572252e26f8dacb42e096f0f6d1fce3/crates/matrix-sdk/src/room/mod.rs#L1774-L1785

Now, there is still an edge case where https://github.com/matrix-org/matrix-rust-sdk/issues/2729 is important, if the future gets cancelled and we fail to discard the room key. This can happen if the application closes unexpectedly, then we are left with an `OutboundGroupSession` in the store that might contain pending requests.

> I guess another way to fix this would be to cancel pending requests to share keys with devices who are no longer in the set of recipients. This seems more complex to me and I'm not sure if races will be possible in this case (between some task picking up a request to send and it being canceled).

I was imagining that we should have a stateful API for this, i.e. `share_room_key()` returns an object that contains some state.

For example it would contain a per-room lock, users can't attempt to share a room key for the same room concurrently, it would also contain logic that would only persist and "activate" the room key if we managed to send out all the pending requests.

So something like:

```rust

async fn share_room_key() -> ShareRoomKeyState {
   ...
} 

impl Drop for ShareRoomKeyState {
    fn drop(&mut self) {
        if !self.is_done() {
            self.session_manager.discard_room_key(&self.room_id);
        }
    }
}

let state = olm_machine.share_room_key().await;

for request in state.pending_requests {
    // The `?` is fine because `drop()` will discard the room key if we fail.    
    let response = client.send_request(request).await?;
    // This will set the `done` flag for the state if this is our final request.
    state.mark_request_as_sent(response).await;
}
```

This still has a problem with cancel safety, unless we also change the way the `OutboundGroupSession` is stored, it should be only stored once the all the request have been shared. Furthermore expressing this over the FFI is pretty tricky.

All that being said, your approach might be better since it removes the cancel-safety problem.

poljar (changes requested):
Could we please split out the fix for https://github.com/matrix-org/matrix-rust-sdk/issues/2729 into a separate PR?

Your original fix and the modified tests look good to me, so we can merge that part right away.

The fix for https://github.com/matrix-org/matrix-rust-sdk/issues/2729 might need some additional tests.

poljar (approved):
Thanks, looks good.

procr1337:
@poljar Done.

> You are right. The reason why this works is that we discard the room key if we don't manage to mark all the pending requests as sent

FYI there is at least one test which fails an `assert!(outbound.pending_requests().is_empty())` placed before the call to `collect_session_recipients`:

```
thread 'session_manager::group_sessions::tests::test_no_olm_sent_once' panicked at crates/matrix-sdk-crypto/src/session_manager/group_sessions/mod.rs:677:9:
assertion failed: outbound.pending_requests().is_empty()


failures:
    session_manager::group_sessions::tests::test_no_olm_sent_once
```

poljar:
> FYI there is at least one test which fails an assert!(outbound.pending_requests().is_empty()) placed before the call to collect_session_recipients:

That doesn't surprise me, as mentioned, the API of the crypto crate doesn't prevent this state.

The `matrix-sdk` crate, which uses the crypto crate, does guard against this via room key discarding.

PR #4957, opened 2025-04-22
WidgetDriver: to-device support, refactor `Filter`

toger5:
## WidgetDriver: to-device support step 1

This is a split out of: https://github.com/matrix-org/matrix-rust-sdk/pull/4856
It is the first step to introduce to-device events to the widget api.
With this PR we only focus on the current filter implementation and change improve it so that adding to-device
filters can be done in a cleaner way. The refactor is a result of what worked well when doing the full implementation in https://github.com/matrix-org/matrix-rust-sdk/pull/4856
(This PR does not yet include any to-device changes. It will make the introduction of the filtering for to-device events much cleaner and simpler to read)

 
The API for the filters was a bit confusing since it had a lot of special cases.
This commit also simplifies the filter public api.

Rethinking the public api we only need:
 - to know if events can be sent based on the capabilities
 - to know if events can be sent to the widget (read) based on the capabilities
 - if it even makes sense to sent a cs api read request or if all possibly returned events
   would not match the type.

To simplify the code in the machine it also made sense to add `From` implementation
to the FilterInputs instead of gathering the relevant data from all kinds of Raw events.

The new api is simpler:
All possible events we need to check can be converted into filter inputs (using `into()`).
`capabilites` has `allow_read`/`allow_send` that consume filter inputs.
`capabilites` can be asked if there is any filter for specific event types
to allow not send unnecassary requests.

Signed-off-by:

### DIFF
--- a/crates/matrix-sdk/src/widget/capabilities.rs
+++ b/crates/matrix-sdk/src/widget/capabilities.rs
@@ -18,11 +18,13 @@
 use std::fmt;
 
 use async_trait::async_trait;
-use ruma::{events::AnyTimelineEvent, serde::Raw};
 use serde::{ser::SerializeSeq, Deserialize, Deserializer, Serialize, Serializer};
-use tracing::{debug, error};
+use tracing::{debug, warn};
 
-use super::{filter::FilterInput, Filter, MessageLikeEventFilter, StateEventFilter};
+use super::{
+    filter::{Filter, FilterInput},
+    MessageLikeEventFilter, StateEventFilter,
+};
 
 /// Must be implemented by a component that provides functionality of deciding
 /// whether a widget is allowed to use certain capabilities (typically by
@@ -56,17 +58,47 @@ pub struct Capabilities {
 }
 
 impl Capabilities {
-    /// Tells if a given raw event matches the read filter.
-    pub fn raw_event_matches_read_filter(&self, raw: &Raw<AnyTimelineEvent>) -> bool {
-        let filter_in = match raw.deserialize_as::<FilterInput>() {
-            Ok(filter) => filter,
-            Err(err) => {
-                error!("Failed to deserialize raw event as MatrixEventFilterInput: {err}");
-                return false;
+    /// Checks if a given event is allowed to be forwarded to the widget.
+    ///
+    /// - `event_filter_input` is a minimized event respresntation that contains
+    ///   only the information needed to check if the widget is allowed to
+    ///   receive the event. (See [`FilterInput`])
+    pub(super) fn allow_reading<'a>(
+        &self,
+        event_filter_input: impl TryInto<FilterInput<'a>>,
+    ) -> bool {
+        match &event_filter_input.try_into() {
+            Err(_) => {
+                warn!("Failed to convert event into filter input for `allow_reading`.");
+                false
             }
-        };
+            Ok(filter_input) => self.read.iter().any(|f| f.matches(filter_input)),
+        }
+    }
 
-        self.read.iter().any(|f| f.matches(&filter_in))
+    /// Checks if a given event is allowed to be sent by the widget.
+    ///
+    /// - `event_filter_input` is a minimized event respresntation that contains
+    ///   only the information needed to check if the widget is allowed to send
+    ///   the event to a matrix room. (See [`FilterInput`])
+    pub(super) fn allow_sending<'a>(
+        &self,
+        event_filter_input: impl TryInto<FilterInput<'a>>,
+    ) -> bool {
+        match &event_filter_input.try_into() {
+            Err(_) => {
+                warn!("Failed to convert event into filter input for `allow_sending`.");
+                false
+            }
+            Ok(filter_input) => self.send.iter().any(|f| f.matches(filter_input)),
+        }
+    }
+
+    /// Checks if a filter exists for the given event type, useful for
+    /// optimization. Avoids unnecessary read event requests when no matching
+    /// filter is present.
+    pub(super) fn has_read_filter_for_type(&self, event_type: &str) -> bool {
+        self.read.iter().any(|f| f.filter_event_type() == event_type)
     }
 }
 
--- a/crates/matrix-sdk/src/widget/capabilities.rs
+++ b/crates/matrix-sdk/src/widget/capabilities.rs
@@ -22,9 +22,7 @@ use ruma::{events::AnyTimelineEvent, serde::Raw};
 use serde::{ser::SerializeSeq, Deserialize, Deserializer, Serialize, Serializer};
 use tracing::{debug, error};
 
-use super::{
-    filter::MatrixEventFilterInput, EventFilter, MessageLikeEventFilter, StateEventFilter,
-};
+use super::{filter::FilterInput, Filter, MessageLikeEventFilter, StateEventFilter};
 
 /// Must be implemented by a component that provides functionality of deciding
 /// whether a widget is allowed to use certain capabilities (typically by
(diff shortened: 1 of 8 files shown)


crates/matrix-sdk/src/widget/filter.rs
```diff
@@ -103,271 +103,325 @@ pub enum StateEventFilter {
...
+
+// Filter input:
+
+/// The input data for the filter. This can either be constructed from a
+/// [`Raw<AnyTimelineEvent>`] or a [`SendEventRequest`].
+#[derive(Debug)]
+pub enum FilterInput {
+    State(FilterInputState),
+    MessageLike(FilterInputMessageLike),
+}
```
poljar, line 137: Considering how these get created it feels like you'd want these to be types that hold a reference to event type instead of taking ownership of the type and content.

crates/matrix-sdk/src/widget/filter.rs
```diff
@@ -103,271 +103,325 @@ pub enum StateEventFilter {
...
+    fn from(request: &SendEventRequest) -> Self {
+        match &request.state_key {
+            None => match request.event_type {
+                TimelineEventType::RoomMessage => {
+                    if let Some(msgtype) =
+                        serde_json::from_str::<MessageLikeFilterEventContent>(request.content.get())
+                            .unwrap_or_else(|e| {
+                                debug!("Failed to deserialize event content for filter: {e}");
+                                // Fallback to empty content is safe because there is no filter
+                                // that matches with it when it otherwise wouldn't.
```
poljar, line 220: That matches with what?

Is it really safe to default to an empty content here?

crates/matrix-sdk/src/widget/capabilities.rs
```diff
@@ -58,17 +58,29 @@ pub struct Capabilities {
...
-    /// Tells if a given raw event matches the read filter.
-    pub fn raw_event_matches_read_filter(&self, raw: &Raw<AnyTimelineEvent>) -> bool {
-        let filter_in = match raw.deserialize_as::<MatrixEventFilterInput>() {
-            Ok(filter) => filter,
-            Err(err) => {
-                error!("Failed to deserialize raw event as MatrixEventFilterInput: {err}");
-                return false;
-            }
-        };
+    /// Checks if a givent event is allowed to be forwarded to the widget.
```
poljar, line 61:
```suggestion
    /// Checks if a given event is allowed to be forwarded to the widget.
```

crates/matrix-sdk/src/widget/capabilities.rs
```diff
@@ -58,17 +58,29 @@ pub struct Capabilities {
...
+    ///
+    /// - `event_filter_input` is a minimized event respresntation that contains
+    ///   only the information needed to check if the widget is allowed to
+    ///   receive the event. (See [`FilterInput`])
+    pub(super) fn allow_reading(&self, event_filter_input: &FilterInput) -> bool {
```
poljar, line 66: What do you think about changing this to take `impl TryInto<FilterInput>` instead of `FilterInput`.

This would work doubly nicely if `FilterInput` just borrows values.

I think that this might simplify a lot of call sites, i.e. this complex expression:

```rust
events.retain(|e|
    matches!(e.clone().try_into(), Ok(f_in) if capabilities.allow_reading(&f_in))
);
```

Would become:

```rust
events.retain(|e| capabilities.allow_reading(e))
```

This code snippet:

```rust
let Ok(event_filter_in) = event_raw.clone().try_into() else {
    error!("Failed to convert event to filter input");
    return Vec::new();
};
if capabilities.allow_reading(&event_filter_in) {
    let action = self.send_to_widget_request(NotifyNewMatrixEvent(event_raw)).1;
    return action.map(|a| vec![a]).unwrap_or_default();
}
Vec::new()
```

Would become:

```rust
if capabilities.allow_reading(&event_filter_in) {
    let (_, action) = self.send_to_widget_request(NotifyNewMatrixEvent(event_raw));
    action.map(|a| vec![a]).unwrap_or_default()
} else {
    Vec::new()
}
```

Of course this would mean that the `try_into()` call would be moved into `allow_reading()`, and thus the `Err` case for `try_into()` would always need to return something, i.e. `false`. But I think that this is exactly what you want?

crates/matrix-sdk/src/widget/filter.rs
```diff
@@ -103,271 +103,325 @@ pub enum StateEventFilter {
...
             }
             StateEventFilter::WithTypeAndStateKey(event_type, filter_state_key) => {
-                matrix_event.event_type == TimelineEventType::from(event_type.clone())
-                    && state_key == filter_state_key
+                &state_filter_input.event_type == event_type
+                    && state_filter_input.state_key == *filter_state_key
             }
         }
     }
+    fn filter_event_type(&self) -> String {
```
poljar, line 121: Why not return `&str`? I can see at least one case where you can avoid the copying.

poljar (changes requested):
I think I found a way to improve this PR quite a bit, I didn't go through all of the code because my suggestion would change things quite a bit.

Please let me know if I'm completely wrong about my suggestion.

crates/matrix-sdk/src/widget/filter.rs
toger5: Where would the owned value of the event type live when we take the reference to it?
If we use the `TryFrom` construction, we dont store the deserialized value anywhere.

crates/matrix-sdk/src/widget/filter.rs
toger5: How can I construct `&str` from a `&StateEventType`. I think the only public method i have is `to_string()`

poljar: Yeah, a bit annoying.

The overhead doesn't particularly matter I'd imagine, so you can pick whichever you prefer here.

crates/matrix-sdk/src/widget/filter.rs
poljar: Very nice.

crates/matrix-sdk/src/widget/machine/driver_req.rs
```diff
@@ -190,7 +190,7 @@ impl FromMatrixDriverResponse for Vec<Raw<AnyTimelineEvent>> {
 #[derive(Clone, Debug)]
 pub(crate) struct ReadStateEventRequest {
     /// The event type to read.
-    pub(crate) event_type: StateEventType,
+    pub(crate) event_type: String,
```
poljar, line 193: Always a bit sad to see weaker typing, but I guess it doesn't matter much here.

crates/matrix-sdk/src/widget/filter.rs
```diff
@@ -102,272 +99,343 @@ pub enum StateEventFilter {
...
+            None => match request.event_type.as_str() {
+                "m.room.message" => {
+                    if let Some(msgtype) =
+                        serde_json::from_str::<MessageLikeFilterEventContent<'a>>(
+                            request.content.get(),
+                        )
+                        .unwrap_or_else(|e| {
+                            debug!("Failed to deserialize event content for filter: {e}");
+                            // Fallback to empty content is safe.
+                            // If we do have a filter matching any content type, it will march
```
poljar, line 202:
```suggestion
                            // If we do have a filter matching any content type, it will match
```

crates/matrix-sdk/src/widget/filter.rs
```diff
@@ -102,272 +99,343 @@ pub enum StateEventFilter {
...
+    pub(super) content: MessageLikeFilterEventContent<'a>,
+}
 
-    fn message_event(event_type: TimelineEventType) -> MatrixEventFilterInput {
-        MatrixEventFilterInput { event_type, state_key: None, content: Default::default() }
+/// Create a filter input based on [`AnyTimelineEvent`].
+/// This will create a [`FilterInput::State`] or [`FilterInput::MessageLike`]
+/// depending on the event type.
+impl<'a> TryFrom<&'a Raw<AnyTimelineEvent>> for FilterInput<'a> {
+    type Error = serde_json::Error;
```
poljar, line 184:
```suggestion
    type Error = serde_json::Error;
    
```

poljar (changes requested):
I found a typo, but other than that I think this looks good.

PR #4959, opened 2025-04-22
WidgetDriver: Use `matrix_sdk_common::executor::spawn` instead of `tokio::spawn` to make it wasm compatible.

toger5:
This PR is a revived version of: https://github.com/matrix-org/matrix-rust-sdk/pull/4707 since it is now possible to easily add wasm support thanks to: https://github.com/matrix-org/matrix-rust-sdk/pull/4572

Using the `executer::spawn` will select `tokio::spawn` or `wasm_bindgen_futures::spawn_local` based on what platform we are on.

~~They behave differently in that `tokio` actually starts the async block inside the spawn but the wasm `spawn` wrapper will only start the part that is not inside the `async` block and the join handle needs to be awaited for it to work.~~

This has now changed with: https://github.com/matrix-org/matrix-rust-sdk/pull/4572.
Now they behave the same and this PR becomes a very simple change.

Signed-off-by:

### DIFF
--- a/crates/matrix-sdk/Cargo.toml
+++ b/crates/matrix-sdk/Cargo.toml
@@ -85,7 +85,7 @@ oauth2 = { version = "5.0.0", default-features = false, features = ["reqwest", "
 once_cell = { workspace = true }
 percent-encoding = "2.3.1"
 pin-project-lite = { workspace = true }
-rand = { workspace = true , optional = true }
+rand = { workspace = true, optional = true }
 ruma = { workspace = true, features = [
     "rand",
     "unstable-msc2448",
@@ -104,6 +104,7 @@ sha2 = { workspace = true }
 tempfile = { workspace = true }
 thiserror = { workspace = true }
 tokio-stream = { workspace = true, features = ["sync"] }
+tokio-util = "0.7.13"
 tower = { version = "0.5.2", features = ["util"], optional = true }
 tracing = { workspace = true, features = ["attributes"] }
 uniffi = { workspace = true, optional = true }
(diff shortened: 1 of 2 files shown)


crates/matrix-sdk/Cargo.toml
jplatte, line 1: The changes to this file seem unrelated and other than the formatting change, likely wrong.

toger5: This change moves `tokio-util` from `not(target_arch = "wasm32")` to an unconditional dependency since it is needed for the `wasm` and the `not wasm` target.
See the commit description for further details why its needed.

I is related to the PR in the sense that both changes are needed to make the widget driver build for wasm: 
 - use `tokio-util` with wasm target
 - use `executor::spawn`
 (but this change is not needed for the `executor::spawn` change.)

jplatte: Okay, so you're saying it just didn't build before on wasm due to `tokio_util` being referenced but not a dependency on that target? Then this definitely makes sense.

I first thought this was a `dev-dependency` becoming a regular dependency. Didn't catch that the lower block was still regular runtime dependencies, but cfg'd.

Just one ask: Can you move `tokio-util` up a little so the dependencies are sorted alphabetically?

jplatte (approved):

crates/matrix-sdk/src/widget/mod.rs
```diff
@@ -139,7 +139,7 @@ impl WidgetDriver {
         let (incoming_msg_tx, mut incoming_msg_rx) = unbounded_channel();
 
         // Forward all of the incoming messages from the widget.
-        tokio::spawn({
+        matrix_sdk_common::executor::spawn({
```
bnjbvr, line 142: Could you use a `use` statement at the top, to avoid the fully-qualified path here and below, please?

bnjbvr (approved):
Thanks! just one small nit :)

crates/matrix-sdk/src/widget/mod.rs
toger5: We do this in a lot of places. I was wondering if this is desired as it allows for easier ctrl+F searching.
But moving it to the import definitely sounds more idiomatic.

bnjbvr (approved):

PR #4961, opened 2025-04-22
test: attempt to deflake history-sharing integ test

richvdh:
Rather than attempting to trigger Alice's encryption sync loop with an incoming message from Bob, instead have Alice send a message once Bob has joined the room.

Hopefully, fixes https://github.com/matrix-org/matrix-rust-sdk/issues/4945.

### DIFF
--- a/crates/matrix-sdk-common/src/timeout.rs
+++ b/crates/matrix-sdk-common/src/timeout.rs
@@ -24,7 +24,7 @@ use tokio::time::timeout as tokio_timeout;
 
 /// Error type notifying that a timeout has elapsed.
 #[derive(Clone, Copy, Debug, PartialEq, Eq)]
-pub struct ElapsedError(());
+pub struct ElapsedError();
 
 impl fmt::Display for ElapsedError {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
@@ -44,7 +44,7 @@ where
     F: Future<Output = T>,
 {
     #[cfg(not(target_arch = "wasm32"))]
-    return tokio_timeout(duration, future).await.map_err(|_| ElapsedError(()));
+    return tokio_timeout(duration, future).await.map_err(|_| ElapsedError());
 
     #[cfg(target_arch = "wasm32")]
     {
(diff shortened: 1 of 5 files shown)


crates/matrix-sdk-common/src/timeout.rs
```diff
@@ -24,7 +24,7 @@ use tokio::time::timeout as tokio_timeout;
 
 /// Error type notifying that a timeout has elapsed.
 #[derive(Clone, Copy, Debug, PartialEq, Eq)]
-pub struct ElapsedError(());
+pub struct ElapsedError(pub ());
```
poljar, line 27: If construction of this has become important, let's rather remove the field completely.

```suggestion
pub struct ElapsedError;
```

testing/matrix-sdk-integration-testing/src/helpers.rs
```diff
@@ -212,19 +215,50 @@ impl Deref for SyncTokenAwareClient {
...
+/// of type [`ElapsedError`].
+pub async fn wait_until_some<F, Fut, R>(
+    mut cb: F,
+    timeout_duration: Duration,
+) -> Result<R, ElapsedError>
+where
+    F: FnMut(u32) -> Fut,
+    Fut: Future<Output = Option<R>>,
+{
+    let start = Instant::now();
```
poljar, line 235: I know that this is trivial but perhaps we could have used `backon` for this.

The body could also handle a bit of whitespace.

That being said, I'm not requesting anything to be changed.

poljar (approved):
Left a small nit, let's see if that behaves better.

crates/matrix-sdk-common/src/timeout.rs
richvdh: sure, I was aiming to avoid breaking changes to existing code, but this makes sense

testing/matrix-sdk-integration-testing/src/helpers.rs
richvdh: eh, yes, probably, on both fronts.

PR #4964, opened 2025-04-23
feat(crypto): Start using the stable identifier for the sender device keys

poljar:
This patch updates the sending side of the `sender_device_keys` field introduced in MSC4147.

Since the MSC got merged, we're switching from the unstable identifier to the stable one.

A couple of snapshot tests were added modified to make this happen.

### DIFF
--- a/crates/matrix-sdk-crypto/CHANGELOG.md
+++ b/crates/matrix-sdk-crypto/CHANGELOG.md
@@ -8,6 +8,10 @@ All notable changes to this project will be documented in this file.
 
 ### Features
 
+- Send stable identifier `sender_device_keys` for MSC4147 (Including device
+  keys with Olm-encrypted events).
+  ([#4964](https://github.com/matrix-org/matrix-rust-sdk/pull/4964))
+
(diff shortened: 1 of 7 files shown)


poljar:
Should close: https://github.com/element-hq/element-meta/issues/2683.

poljar:
> I have absolutely no clues about what this does, but I assume you do, and the code looks good to me 👍

It just renames an identifier, nothing fancy.

> (In case this question is useful: since the `OlmV1Event` seem to be serialized/deserialized, this might mean that a previous `sender_device_keys` inserted _before_ this patch will be missing _after_ this patch: is this OK, or does this warrant a data migration?)

Hmm, I don't think it will be missing _after_ this patch. The deserialization is still using a `serde` alias:

https://github.com/matrix-org/matrix-rust-sdk/blob/8131301110089c09e0e5b0c3fc6c849b91b28b3b/crates/matrix-sdk-crypto/src/types/events/olm_v1.rs#L194-L195

richvdh (approved):
LGTM

PR #4967, opened 2025-04-23
refactor(widget): rename some concepts

bnjbvr:
After reading the widget code for *a long time* this afternoon, I think I finally understand some things about it. It seems there are two channels between the widget and the "machine", and the direction is indicated in the name of some structs. For instance, `ToWidgetRequest` = "some request has been sent from the machine to the widget", aka it's always from the point of view of the *sender*.  For the second channel, there's `FromWidgetRequest`, aka it's a request sent from the widget to the machine.

Now, this is where it gets confusing for me: `ToWidgetResponse` is the response from the widget to the machine. In other words, what I think motivated the name is that, this is the response to a request of kind "ToWidget". Bizarrely, there's no equivalent for responses to a `FromWidgetRequest`; the responses are plain serialized from their original types.

If I'm right, then I'm super super confused by this, because I'd imagine a response to a `ToWidgetRequest` (from machine to widget) would be named after the fact **it's sent back by the machine to the widget**, so I'd intuit it should be the `FromWidgetResponse`. But that's the opposite: it is the `ToWidgetResponse`, which I imagined was the response from the machine to the widget.

I think some renaming is due here, but I'm not sure how yet.

### DIFF
--- a/crates/matrix-sdk/src/widget/machine/from_widget.rs
+++ b/crates/matrix-sdk/src/widget/machine/from_widget.rs
@@ -29,7 +29,7 @@ use crate::{widget::StateKeySelector, Error, HttpError, RumaApiError};
 
 #[derive(Deserialize, Debug)]
 #[serde(tag = "action", rename_all = "snake_case", content = "data")]
-pub(super) enum FromWidgetRequest {
+pub(super) enum WidgetToDriverRequest {
     SupportedApiVersions {},
     ContentLoaded {},
     #[serde(rename = "get_openid")]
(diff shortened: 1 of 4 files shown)


bnjbvr:
I realize there's a third concept called the `Driver` that's the SDK client being actually used; so my initial post is probably incorrect and should `s/driver/machine`. Oh well :smiling_face_with_tear:

toger5:
Your original post seems about right.
 - there are `from_widget` and `to_widget` actions
 - they can either be sent as a request or as a response (containing a response field vs not containing a response field) -> both actions can be sent in both directions... It uses directional terminology for a property that behaves more like a type.

This is how the widget MSC's are written as of now however. So I do think the naming should be as it is in the: [MSC](https://github.com/matrix-org/matrix-spec-proposals/blob/travis/msc/widgets-send-receive-events/proposals/2762-widget-event-receiving.md)'s (["not up to date general widget msc"](https://github.com/matrix-org/matrix-spec-proposals/issues/3803)), js-sdk, EW, [matrix-widget-api](https://github.com/matrix-org/matrix-widget-api/pulls)

bnjbvr:
Thanks! Closing this PR, the new documentation you've written (thanks!) does help reduce my confusion in this area now :)

bnjbvr:
Wait, some of these changes might be good in isolation, without the renamings…

bnjbvr:
Emotional rollercoaster: I forgot the other commits landed as part of another PR. Closing for realz this time.

PR #4971, opened 2025-04-24
docs(common): add a comment explaining how to use the `timer!` macro

bnjbvr:
"How long would it take to figure out how to use a macro? Less than 10 minutes, right? RIGHT?"

### DIFF
--- a/crates/matrix-sdk-common/src/tracing_timer.rs
+++ b/crates/matrix-sdk-common/src/tracing_timer.rs
@@ -15,7 +15,7 @@
 use ruma::time::Instant;
 use tracing::{callsite::DefaultCallsite, Callsite as _};
 
-/// A named RAII that will show on Drop how long its covered section took to
+/// A named RAII that will show on `Drop` how long its covered section took to
 /// execute.
 pub struct TracingTimer {
     id: String,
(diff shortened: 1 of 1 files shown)


crates/matrix-sdk-common/src/tracing_timer.rs
```diff
@@ -77,6 +77,12 @@ impl TracingTimer {
 ///
 /// The tracing level can be specified as a first argument, but it's optional.
 /// If it's missing, this will use the debug level.
+///
```
poljar, line 80: Hmm, I would expect people to use the LSP to import things, the LSP knows how to import things for you.

Is this a useful comment? If so, why would we put it only on this macro?

Instead of this comment, perhaps it would make more sense to fix the first line in the docs and add an example:

> /// Macro to create a RAII timer that will log a `tracing` event once it's
/// dropped.

This doesn't explain what we're timing like the documentation for the struct does:

> /// A named RAII that will show on Drop how long its covered section took to
/// execute.

Adding an example to this will showcase how to use the macro as well as how to import it properly.

poljar (approved):
Approving since there's not much to it, but I think an example would convey how to import the macro much better.

crates/matrix-sdk-common/src/tracing_timer.rs
bnjbvr: > Hmm, I would expect people to use the LSP to import things, the LSP knows how to import things for you.

In this case, either the LSP didn't help me, or I was dumb and didn't trigger the LSP action to auto-import things :smiling_face_with_tear: 

Will add an example :+1:

PR #4973, opened 2025-04-24
feat(threads): add methods to list a room's threads + list all related events

bnjbvr:
The tests show how to use them for listing threads (`Room::list_threads()`), and the events within a thread (`Room::relations()`).

This should be sufficient to implement an in-memory (viz. not persisted on disk) `RoomThreadEventCache` :eyes: according to the plan described in #4869.

Part of #4869.


poljar (approved):
It's a bit strange that the last commit introduces test helpers that were used in the previous commit, there's also a typo in the last commit message.

Overall looks good, I would liked some examples for some of those new methods but alas.

I'm going to fix the typo and commit ordering myself.

PR #4977, opened 2025-04-25
feat(send_queue): Implement sending of MSC4274 galleries

Johennes:
This was broken out of https://github.com/matrix-org/matrix-rust-sdk/pull/4838 and is a step towards implementing [MSC4274](https://github.com/matrix-org/matrix-spec-proposals/pull/4274).

* The entry point for sending galleries via the send queue is a new method [`RoomSendQueue::send_gallery`](https://github.com/matrix-org/matrix-rust-sdk/pull/4977/files#diff-8752e86459c22cff470d7ca617240dcbdf222c3c6c98be2af2e43ddec071154cR362) which is a generalization of `RoomSendQueue::send_attachment`.
  * `send_gallery` takes as input parameters a [`GalleryConfig`](https://github.com/matrix-org/matrix-rust-sdk/pull/4977/files#diff-d38d74cec6159cf769c32bed496199146175f05428fc23ab13bc2c629900da3eR283) (containing info about the gallery itself, such as its caption)  and a vector of [`GalleryItemInfo`](https://github.com/matrix-org/matrix-rust-sdk/pull/4977/files#diff-d38d74cec6159cf769c32bed496199146175f05428fc23ab13bc2c629900da3eR355)s (containing info about each image / file / etc. in the gallery).
  * `send_gallery` creates the gallery event content via [`Room:make_message_event`](https://github.com/matrix-org/matrix-rust-sdk/pull/4977/files#diff-474f20e47fdcf60feac3f839a81f82fbb2c1fd0bb406388b0262adb60216ce3bR2281) which was renamed from `make_attachment_event` to reflect the fact that it creates general `msgtype` events now.
  * `send_gallery` maps the passed item infos into [`GalleryItemQueueInfo`](https://github.com/matrix-org/matrix-rust-sdk/pull/4977/files#diff-d569f54c7901b5cd660aae77045c80dabbd3c251f0fa5891530469f35941327aR2008)s. This additional struct allows grouping all the metadata for a single gallery item together.
* Finally `send_gallery` invokes [`QueueStorage::push_gallery`](https://github.com/matrix-org/matrix-rust-sdk/pull/4977/files#diff-d569f54c7901b5cd660aae77045c80dabbd3c251f0fa5891530469f35941327aR1294) which is a generalization of `QueueStorage::push_media`.
  * `send_gallery` pushes upload requests for the media and thumbnails to the queue in a "daisy chain" manner. The first thumbnail (or media if no thumbnail exists) is pushed as a `QueuedRequestKind::MediaUpload`. The remaining thumbnails and media are pushed as `DependentQueuedRequestKind::UploadFileOrThumbnail`s while chaining each request to the previous one.
  * Finally a [`DependentQueuedRequestKind::FinishGallery`](https://github.com/matrix-org/matrix-rust-sdk/pull/4977/files#diff-42a1a7ebe2446d916c9e013293bafc5eb16fd89bfe89248e3877ef031cc83ef2R265) is pushed to finalize the gallery upload (analogous to the existing `FinishUpload` for single media uploads).
* The `FinishGallery` request is handled in [`QueueStorage::handle_dependent_finish_gallery_upload`](https://github.com/matrix-org/matrix-rust-sdk/pull/4977/files#diff-8752e86459c22cff470d7ca617240dcbdf222c3c6c98be2af2e43ddec071154cR628) which was modeled after `handle_dependent_finish_upload`.
  * To be able to map the temporary event source for each gallery item to the final `AccumulatedSentMediaInfo`, a hash map is used.
  * Using the hash map, the gallery event is then updated to use the actual media sources via [`update_gallery_event_after_upload`](https://github.com/matrix-org/matrix-rust-sdk/pull/4977/files#diff-8752e86459c22cff470d7ca617240dcbdf222c3c6c98be2af2e43ddec071154cR104) and a new method [`Room::make_gallery_item_type`](https://github.com/matrix-org/matrix-rust-sdk/pull/4977/files#diff-474f20e47fdcf60feac3f839a81f82fbb2c1fd0bb406388b0262adb60216ce3bR2303)
* An [integration test](https://github.com/matrix-org/matrix-rust-sdk/pull/4977/files#diff-21532ad5467a69489d7913b8da7afd4d618b7e357ce94f769e6e60e395b58055R2048) has been added to demonstrate the functionality.

This is relatively large, unfortunately, but including everything needed to actually send the event made it possible to also add a test for it. It would be nice if the amount of new code could be reduced but I'm struggling a bit to find ways to integrate galleries with the existing media uploads further.


crates/matrix-sdk/src/room/mod.rs
```diff
@@ -2294,6 +2296,99 @@ impl Room {
...
+                let info = assign!(info.map(ImageInfo::from).unwrap_or_default(), {
+                    mimetype: Some(content_type.as_ref().to_owned()),
+                    thumbnail_source,
+                    thumbnail_info
+                });
+                let content = assign!(ImageMessageEventContent::new(body, source), {
+                    info: Some(Box::new(info)),
+                    formatted: formatted_caption,
+                    filename
+                });
```
Johennes, line 2334: These blocks could maybe share the logic with `make_attachment_type` by extracting a function to create the content for each mime-type? I wasn't sure if it's really worth it because it might not be shorter due to the number of arguments on the call.

Johennes:
I'm mindful that it's probably just a matter of lack of time but if there's anything I can do to simplify reviewing this PR, I'd be happy to follow any suggestions.

crates/matrix-sdk/src/send_queue/mod.rs
```diff
@@ -1284,6 +1286,199 @@ impl QueueStorage {
...
+                            upload_file_txn.clone().into(),
+                            created_at,
+                            DependentQueuedRequestKind::UploadFileOrThumbnail {
+                                content_type: content_type.to_string(),
+                                cache_key: file_media_request.clone(),
+                                related_to: send_event_txn.clone(),
+                                parent_is_thumbnail_upload: false,
+                            },
+                        )
+                        .await?;
```
bnjbvr, line 1450: This is always done in both branches, so it can be extracted out of the if/else and be put just after it (by passing `parent_is_thumbnail_upload`, and mutating `last_upload_file_txn` in place in the thumbnail case). Less code FTW!

crates/matrix-sdk/src/send_queue/mod.rs
```diff
@@ -1284,6 +1286,199 @@ impl QueueStorage {
...
+                last_upload_file_txn = Some(upload_file_txn.clone());
+            }
+        }
+
+        // Push the request for the event itself as a dependent request of the last file
+        // upload.
+        store
+            .save_dependent_queued_request(
+                &self.room_id,
+                &last_upload_file_txn.unwrap(),
```
bnjbvr, line 1469: Hey, don't hide an `unwrap()` like this :D I think this is incorrect, if there's no gallery items in the gallery. This should be impossible based on another previous comment I made, but let's make sure of it by putting this code in the `if ... split_first()` above.

Then, we can reduce the entire function's body indent by one level, by having some kind of `let Some((first, rest)) = ...split_first() else { return Ok(()); }`.

Which means that `last_upload_file_txn` will now always be set, something that the compile ought to be able to infer, so we can make it non `Option`al, and start by not setting it initially.

crates/matrix-sdk/src/send_queue/upload.rs
```diff
@@ -349,6 +621,134 @@ impl QueueStorage {
...
+            .into_media()
+            .ok_or(RoomSendQueueError::StorageError(RoomSendQueueStorageError::InvalidParentKey))?;
+
+        let mut sent_media_vec = sent_gallery.accumulated;
+        sent_media_vec.push(AccumulatedSentMediaInfo {
+            file: sent_gallery.file,
+            thumbnail: sent_gallery.thumbnail,
+        });
+
+        let mut sent_infos = HashMap::<String, AccumulatedSentMediaInfo>::new();
```
bnjbvr, line 648: 1. Do we have to define the type of the hashmap here, or can Rust infer it?
2. Is it worth using a BTreeMap, for making the order of iteration deterministic?

crates/matrix-sdk/src/send_queue/upload.rs
```diff
@@ -349,6 +621,134 @@ impl QueueStorage {
...
+                cache_store
+                    .replace_media_key(
+                        &from_req,
+                        &MediaRequestParameters {
+                            source: sent_media.file.clone(),
+                            format: MediaFormat::File,
+                        },
+                    )
+                    .await
+                    .map_err(RoomSendQueueStorageError::EventCacheStoreError)?;
```
bnjbvr, line 684: Any chance some of this can also be factored out, to avoid duplicating the code of `handle_dependent_finish_upload`?

crates/matrix-sdk/src/send_queue/upload.rs
```diff
@@ -85,6 +98,89 @@ fn update_media_event_after_upload(echo: &mut RoomMessageEventContent, sent: Sen
...
 
+/// Replace the sources by the final ones in all the media types handled by
+/// [`Room::make_gallery_item_type()`].
+#[cfg(feature = "unstable-msc4274")]
+fn update_gallery_event_after_upload(
+    echo: &mut RoomMessageEventContent,
+    sent: HashMap<String, AccumulatedSentMediaInfo>,
+) {
+    match &mut echo.msgtype {
+        MessageType::Gallery(gallery) => {
```
bnjbvr, line 109: Let's reduce indent by one level by doing something like this:

```rust
let MessageType::Gallery(gallery) = &mut echo.msgtype else {
  // All `GalleryItemType` created by `Room::make_gallery_item_type` should be
  // handled here. The only way to end up here is that a item type has
  // been tampered with in the database.
  error!("Invalid gallery item types in database");
  // Only crash debug builds.
  debug_assert!(false, "invalid item type in database {:?}", echo.msgtype());
  return;
};
```

crates/matrix-sdk-base/src/store/send_queue.rs
```diff
@@ -285,6 +298,17 @@ pub struct FinishUploadThumbnailInfo {
     pub height: Option<UInt>,
 }
 
+#[cfg(feature = "unstable-msc4274")]
+#[derive(Clone, Debug, Serialize, Deserialize)]
+/// Detailed record about a file and thumbnail used when finishing a gallery
+/// upload.
```
bnjbvr, line 304: 1. Let's put the doc comment above the `cfg` and derive.
2. The name suggests this is for a single item (`finish gallery item`), but the code comment suggests it's when finishing a gallery upload (so I'd expect after all the items have been uploaded). Can you precise in the doc comment / name what's the truth?

bnjbvr (changes requested):
Looks great! The test is quite comprehensive, that's nice.

A first batch of comments; I think we might be able to reduce the code duplication with the existing send_attachment functionality in various places, so let's try it out. If it turns out to be more trouble (e.g. require doing too much data/parameters shuffling, etc.), in some places, please don't hesitate to *not* do it.

I'd like to take another look after that, but it looks like this is on the right track. Thanks!

crates/matrix-sdk/src/send_queue/upload.rs
```diff
@@ -241,6 +337,182 @@ impl RoomSendQueue {
...
+    ) -> Result<SendHandle, RoomSendQueueError> {
+        let Some(room) = self.inner.room.get() else {
+            return Err(RoomSendQueueError::RoomDisappeared);
+        };
+
+        if room.state() != RoomState::Joined {
+            return Err(RoomSendQueueError::RoomNotJoined);
+        }
+
+        let send_event_txn = config.txn_id.map_or_else(ChildTransactionId::new, Into::into);
```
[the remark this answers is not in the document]
Johennes: Good call. I cannot think of any good use cases for sending an empty gallery so I made it return an error.

crates/matrix-sdk/src/room/mod.rs
```diff
@@ -2294,6 +2296,99 @@ impl Room {
         Ok(content)
     }
 
+    /// Creates the inner [`GalleryItemType`] for an already-uploaded media file
+    /// provided by its source.
+    #[cfg(feature = "unstable-msc4274")]
+    #[allow(clippy::too_many_arguments)]
+    pub(crate) fn make_gallery_item_type(
+        &self,
```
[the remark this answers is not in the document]
Johennes: Ah, right. Have fixed it in both.

crates/matrix-sdk/src/send_queue/upload.rs
```diff
@@ -241,6 +337,182 @@ impl RoomSendQueue {
...
+    /// The attachments and their optional thumbnails are stored in the media
+    /// cache and can be retrieved at any time, by calling
+    /// [`Media::get_media_content()`] with the `MediaSource` that can be found
+    /// in the local or remote echo, and using a `MediaFormat::File`.
+    #[cfg(feature = "unstable-msc4274")]
+    #[instrument(skip_all, fields(event_txn))]
+    pub async fn send_gallery(
+        &self,
+        item_infos: Vec<GalleryItemInfo>,
+        config: GalleryConfig,
```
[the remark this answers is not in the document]
Johennes: Yes, agreed. Have bundled the items with the config, added `is_empty` and `len` methods and used them elsewhere.

Johennes:
Ok, I think I've addressed everything now.

PR #4980, opened 2025-04-25
refactor(crypto): Properly encapsulate internal `OutboundGroupSession` state

procr1337:
As discussed in #4975. Previously, the `share_strategy` was breaking the abstraction provided by `OutboundGroupSession` by accessing its internal fields in an inconsistent and adhoc way. Now all fields are private and a proper abstraction was added to access the required state in a consistent API.

NOTE: Mostly used this to practice my Rust, I'm not heavily invested in it

### DIFF
--- a/crates/matrix-sdk-crypto/src/olm/group_sessions/outbound.rs
+++ b/crates/matrix-sdk-crypto/src/olm/group_sessions/outbound.rs
@@ -16,9 +16,10 @@ use std::{
     cmp::max,
     collections::{BTreeMap, BTreeSet},
     fmt,
+    ops::Bound,
     sync::{
         atomic::{AtomicBool, AtomicU64, Ordering},
-        Arc,
+        Arc, RwLockReadGuard,
     },
     time::Duration,
 };
@@ -70,7 +71,7 @@ const ONE_WEEK: Duration = Duration::from_secs(60 * 60 * 24 * 7);
 const ROTATION_PERIOD: Duration = ONE_WEEK;
 const ROTATION_MESSAGES: u64 = 100;
 
-#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
 /// Information about whether a session was shared with a device.
 pub(crate) enum ShareState {
     /// The session was not shared with the device.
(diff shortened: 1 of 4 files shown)


crates/matrix-sdk-crypto/src/olm/group_sessions/outbound.rs
```diff
@@ -206,6 +206,88 @@ pub struct SharedWith {
...
+                            message_index: info.message_index,
+                            olm_wedging_index: info.olm_wedging_index,
+                        }
+                    } else {
+                        ShareState::SharedButChangedSenderKey
+                    }
+                }
+                ShareInfo::Withheld(_) => ShareState::NotShared,
+            })
+            .max()
```
poljar, line 234: Heh, this is a nice trick, though a bit obscure. I guess the enum is unlikely to be expanded for the `PartialEq` implementation to become invalid.

But perhaps we should add a unit test that ensures the `PartialEq` implementation does the right thing. I.e. `max()` and `min()` always return the correct variants.

poljar (changes requested):
This looks good, very neat tricks were used to replace a bunch of duplicate code.

Could we just please add a unit test to prevent unsuspecting modifications of the `ShareState` enum making the `PartialEq` implementation behave completely differently.

crates/matrix-sdk-crypto/src/olm/group_sessions/outbound.rs
procr1337: done

crates/matrix-sdk-crypto/src/olm/group_sessions/outbound.rs
```diff
@@ -206,6 +206,88 @@ pub struct SharedWith {
...
+            set: &'a ShareInfoSet,
+            user_ids: (Bound<&'b UserId>, Bound<&'b UserId>),
+            device_ids: (Bound<&'c DeviceId>, Bound<&'c DeviceId>),
+        ) -> impl Iterator<Item = (&'a UserId, &'a DeviceId, &'a ShareInfo)> + use<'a, 'b, 'c>
+        {
+            set.range::<UserId, _>(user_ids).flat_map(move |(uid, d)| {
+                d.range::<DeviceId, _>(device_ids)
+                    .map(|(id, info)| (uid.as_ref(), id.as_ref(), info))
+            })
+        }
```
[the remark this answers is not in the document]
procr1337: ty

PR #4981, opened 2025-04-28
refactor(crypto): Move `session_id` from `EncryptionInfo` to `AlgorithmInfo` as it is megolm specific

BillCarsonFr:
Moves the `session_id` field from EncryptionInfo to AlgorithmInfo::MegolmV1AesSha2 as it is specific to megolm. Provides transparent migration of serialized data.

In the future we plan to reuse `EncryptionInfo` for to_device decryption (using olm not megolm). So megolm session_id should move to algorithm specific data.

This is a follow-up of https://github.com/matrix-org/matrix-rust-sdk/pull/4935 and also ground work for future changes.
As discussed here https://github.com/matrix-org/matrix-rust-sdk/pull/4935#discussion_r2049153547

This PR also provides migration for old serialized `EncryptionInfo` (serialized in the event cache).

I only provide specific migration for `EncryptionInfo`.
I don't think more migration it is needed for `AlgorithmInfo::MegolmV1AesSha2` as the added `session_id` is an Option and a `serde/default`is added. Migration for AlgorithmInfo is just done via `#[serde(default, skip_serializing_if = "Option::is_none")]`. `skip_serializing_if` is used in order to not break existing snapshots to avoid noise. But not sure if the right choice.

Signed-off-by:

### DIFF
--- a/crates/matrix-sdk-common/src/deserialized_responses.rs
+++ b/crates/matrix-sdk-common/src/deserialized_responses.rs
@@ -599,14 +599,11 @@ impl TimelineEventKind {
     pub fn session_id(&self) -> Option<&str> {
         match self {
             TimelineEventKind::Decrypted(decrypted_room_event) => {
-                let AlgorithmInfo::MegolmV1AesSha2 { session_id, .. } =
-                    &decrypted_room_event.encryption_info.algorithm_info;
-                session_id.as_ref()
+                decrypted_room_event.encryption_info.session_id()
             }
-            TimelineEventKind::UnableToDecrypt { utd_info, .. } => utd_info.session_id.as_ref(),
+            TimelineEventKind::UnableToDecrypt { utd_info, .. } => utd_info.session_id.as_deref(),
             TimelineEventKind::PlainText { .. } => None,
         }
-        .map(String::as_str)
     }
 }
 
--- a/crates/matrix-sdk-common/src/deserialized_responses.rs
+++ b/crates/matrix-sdk-common/src/deserialized_responses.rs
@@ -322,42 +322,37 @@ impl<'de> Deserialize<'de> for EncryptionInfo {
     where
         D: serde::Deserializer<'de>,
     {
-        // Deserialize as a generic JSON value
-        let value = serde_json::Value::deserialize(deserializer)?;
-
-        // Backwards compatibility:  Capture session_id at root if exists, in legacy
+        // Backwards compatibility: Capture session_id at root if exists. In legacy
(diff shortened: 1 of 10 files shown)


poljar (changes requested):
I think I found a way to simplify this.

poljar:
Wait, where is my comment?!?

crates/matrix-sdk-common/src/deserialized_responses.rs
```diff
@@ -302,9 +307,59 @@ pub struct EncryptionInfo {
...
+            binding.as_object().and_then(|obj| obj.get("session_id")).and_then(|s| s.as_str());
+
+        // Now deserialize normally without legacy fields
+        #[derive(Deserialize)]
+        struct Helper {
+            pub sender: OwnedUserId,
+            pub sender_device: Option<OwnedDeviceId>,
+            pub algorithm_info: AlgorithmInfo,
+            pub verification_state: VerificationState,
+        }
```
poljar, line 341: Can't this just be:

```rust
struct Helper {
    sender: OwnedUserId,
    sender_device: Option<OwnedDeviceId>,
    algorithm_info: AlgorithmInfo,
    verification_state: VerificationState,
    #[serde(rename = "session_id"]
    old_session_id,
}
```

The helper being the same as the real struct might confuse people. This also avoids two deserialization steps and the dependence on `serde_json::Value`.

Then you can merge the session IDs using:

```rust
let session_id = session_id.or(old_session_id);
```

It might make sense to destruct the helper, so any new additions to the struct don't get forgotten:

```rust
let Helper { sender, sender_device, algorithm_info, verification_state, session_id} = Helper::deserialize(deserializer)?;
```

poljar (changes requested):
Ok, now for real? With a comment.

crates/matrix-sdk-common/src/deserialized_responses.rs
BillCarsonFr: Thanks it is a lot better. see here https://github.com/matrix-org/matrix-rust-sdk/pull/4981/commits/69fe1db925a0ad77fed7467783fe98251bad37b0

We might write some sort of cheatsheet with all the strategies we use migrate to with serde (via a Helper, custom deserialize, field attributes, serde from/try_from, untagged...). I'll try to come with a first version

poljar (changes requested):
Looks much better. Let's just add a changelog entry, since this is a breaking API change.

poljar:
Huh, did you forget to push after my last review? The changelog entry is still missing.

poljar (approved):

PR #4984, opened 2025-04-28
feat(bindings): check if report room api is supported

Velin92:
Since only homeservers that support matrix v1.13 are able to use this API, clients might need to know if the report room API is available or not, to display an action to the user


stefanceriu (approved):
lgtm 👍

PR #4987, opened 2025-04-29
fix(sdk): Upload encrypted media with `application/octet-stream` mime type

jmartinesp:
This is apparently the right way to do it, both because some HS expect only this mime type and also so we don't leak the mime type of the encrypted media.

This was initially reported in https://github.com/element-hq/element-x-android/issues/3848.

Signed-off-by:


crates/matrix-sdk/src/encryption/mod.rs
```diff
@@ -435,24 +435,22 @@ impl Client {
...
     /// room.send(CustomEventContent { encrypted_file }).await?;
     /// # anyhow::Ok(()) };
     /// ```
     pub fn upload_encrypted_file<'a, R: Read + ?Sized + 'a>(
         &'a self,
-        content_type: &'a mime::Mime,
         reader: &'a mut R,
     ) -> UploadEncryptedFile<'a, R> {
-        UploadEncryptedFile::new(self, content_type, reader)
+        UploadEncryptedFile::new(self, &mime::APPLICATION_OCTET_STREAM, reader)
```
poljar, line 447: I think we can remove this even further, no need for `UploadEncryptedFile` to take this as an argument.

PR #4988, opened 2025-04-29
crypto: improve SenderData stored with room key bundle data

richvdh:
If we already have cross-signing details for the owner of the device at the point we receive the to-device message, we should store that rather than just the device info.

To enable this, we have a couple of preparatory commits which factor out some code from `SenderDataFinder` so that we can reuse the same logic when we receive the key bundle data.

Suggest review commit-by-commit

Part of https://github.com/element-hq/element-meta/issues/2685

### DIFF
--- a/crates/matrix-sdk-crypto/src/machine/mod.rs
+++ b/crates/matrix-sdk-crypto/src/machine/mod.rs
@@ -952,9 +952,17 @@ impl OlmMachine {
             return Ok(());
         };
 
+        // We already checked that `sender_device_keys` matches the actual sender of the
+        // message when we decrypted the message, which included doing
+        // `DeviceData::try_from` on it, so it can't fail.
+
+        let sender_device_data =
+            DeviceData::try_from(sender_device_keys).expect("failed to verify sender device keys");
+        let sender_device = self.store().wrap_device_data(sender_device_data).await?;
+
         changes.received_room_key_bundles.push(StoredRoomKeyBundleData {
             sender_user: event.sender.clone(),
-            sender_data: SenderData::device_info(sender_device_keys.clone()),
+            sender_data: SenderData::from_device(&sender_device),
             bundle_data: event.content.clone(),
         });
         Ok(())
--- a/crates/matrix-sdk-crypto/src/olm/group_sessions/sender_data.rs
+++ b/crates/matrix-sdk-crypto/src/olm/group_sessions/sender_data.rs
@@ -16,9 +16,13 @@ use std::{cmp::Ordering, fmt};
 
 use ruma::{DeviceId, OwnedDeviceId, OwnedUserId, UserId};
 use serde::{de, de::Visitor, Deserialize, Deserializer, Serialize};
+use tracing::error;
 use vodozemac::Ed25519PublicKey;
 
-use crate::types::{serialize_ed25519_key, DeviceKeys};
+use crate::{
+    types::{serialize_ed25519_key, DeviceKeys},
+    Device,
+};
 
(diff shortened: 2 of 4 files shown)


crates/matrix-sdk-crypto/src/machine/mod.rs
```diff
@@ -952,9 +952,17 @@ impl OlmMachine {
             return Ok(());
         };
 
+        // We already checked that `sender_device_keys` matches the actual sender of the
+        // message when we decrypted the message, which included doing
+        // `DeviceData::try_from` on it, so it can't fail.
```
poljar, line 957: Can't we instead treat that the same way as we treat the `None` case up above?

And let a test ensure that we always handle this properly?

crates/matrix-sdk-crypto/src/olm/group_sessions/sender_data.rs
```diff
@@ -216,6 +220,61 @@ impl SenderData {
...
+        if let Some(master_key) = master_key {
+            // We have user_id and master_key for the user sending the to-device message.
+            let master_key = Box::new(master_key);
+            let known_sender_data = KnownSenderData { user_id, device_id, master_key };
+            if sender_device.is_cross_signing_trusted() {
+                Self::SenderVerified(known_sender_data)
+            } else if sender_device
+                .device_owner_identity
+                .as_ref()
+                .expect("User with master key must have identity")
```
poljar, line 262: I know that this is existing code but this could have been expressed a bit differently to avoid the panic.

Something like:

```rust
match (device_owner, master_key) {
    (Some(..), Some(..) => {
        // Happy path
    },
    (_, _) => {
        // Surprisingly, there was no key in the MasterPubkey. We did not expect this:
        // treat it as if the device was not signed by this master key.
        //
        error!("MasterPubkey for user {user_id} does not contain any keys!");
        Self::device_info(sender_device.as_device_keys().clone())
    } 
}

poljar (approved):
Left a couple nits about the panics here.

crates/matrix-sdk-crypto/src/machine/mod.rs
richvdh: You mean, ignore the to-device message, rather than panic?

I suppose so, though generally I take the view that "unreachable" code should cause a crash so that you get an early warning that your previous assumptions no longer hold.

poljar: Well ok, I guess it's marginally easier to see where the assumption changed if we panic here instead in a test.

richvdh: As for a test: what did you have in mind? `receive_room_key_bundle_data` is private so can't easily be called directly from a test, as are its callers. And, for obvious reasons, we can't test this by injecting a to-device message with a mis-signed sender_device_keys.

crates/matrix-sdk-crypto/src/olm/group_sessions/sender_data.rs
richvdh: Yup. done in 8e762662b.

crates/matrix-sdk-crypto/src/machine/mod.rs
poljar: I just assumed that testing the happy path here would be enough to ensure that the assumption didn't change.

richvdh: If you don't have strong opinions on this one, I'd like to leave it as-is. I think otherwise we end up with more unreachable (and untestable) code.

PR #4990, opened 2025-04-30
WidgetDriver: Add `io.element.call.encryption_keys` to-device capability.

toger5:
This needs to be part of the send/read capabilities so that to-device keys can be used.

Blocked by (depends on): https://github.com/matrix-org/matrix-rust-sdk/pull/4963

Signed-off-by:


toger5:
now part of: https://github.com/matrix-org/matrix-rust-sdk/pull/4963

PR #4991, opened 2025-04-30
Notification client call members

stefanceriu:
This PR tweaks the NotificationClient so that it also requests the `org.matrix.msc3401.call.member` state and exposes a method for fetching rooms from its internal in-memory store backed client.

These will be used to check whether a room still has an ongoing call before showing the ringing screen.


jmartinesp (approved):
LGTM! The tests seem broken after the required state changes though.

toger5 (approved):

PR #4994, opened 2025-04-30
feat(ffi): Add methods for observing account data changes

Johennes:

### DIFF
--- a/bindings/matrix-sdk-ffi/src/client.rs
+++ b/bindings/matrix-sdk-ffi/src/client.rs
@@ -647,7 +647,7 @@ impl Client {
                     }
                 })))
             }
-            AccountDataEventType::SecretStorageKey(key_id) => {
+            AccountDataEventType::SecretStorageKey { key_id } => {
                 // Using an Arc here is mandatory or else the subscriber will never trigger
                 let observer = Arc::new(self.inner.observe_events::<RumaGlobalAccountDataEvent<SecretStorageKeyEventContent>, ()>());
 
--- a/bindings/matrix-sdk-ffi/src/client.rs
+++ b/bindings/matrix-sdk-ffi/src/client.rs
@@ -573,97 +573,77 @@ impl Client {
         event_type: AccountDataEventType,
         listener: Box<dyn AccountDataListener>,
     ) -> Arc<TaskHandle> {
+        macro_rules! observe {
+            ($t:ty, $cb: expr) => {{
+                // Using an Arc here is mandatory or else the subscriber will never trigger
+                let observer = Arc::new(self.inner.observe_events::<$t, ()>());
+
+                Arc::new(TaskHandle::new(get_runtime_handle().spawn(async move {
+                    let mut subscriber = observer.subscribe();
+                    loop {
+                        if let Some(next) = subscriber.next().await {
+                            $cb(next.0);
+                        }
+                    }
+                })))
+            }};
+        }
+
         match event_type {
             AccountDataEventType::Direct => {
-                // Using an Arc here is mandatory or else the subscriber will never trigger
-                let observer = Arc::new(
-                    self.inner
-                        .observe_events::<RumaGlobalAccountDataEvent<DirectEventContent>, ()>(),
-                );
-
-                Arc::new(TaskHandle::new(get_runtime_handle().spawn(async move {
-                    let mut subscriber = observer.subscribe();
-                    loop {
-                        if let Some(next) = subscriber.next().await {
-                            listener.on_change(next.0.into());
-                        }
+                observe!(
+                    RumaGlobalAccountDataEvent<DirectEventContent>,
+                    |event: RumaGlobalAccountDataEvent<DirectEventContent>| {
+                        listener.on_change(event.into());
                     }
-                })))
+                )
             }
             AccountDataEventType::IdentityServer => {
-                // Using an Arc here is mandatory or else the subscriber will never trigger
-                let observer = Arc::new(self.inner.observe_events::<RumaGlobalAccountDataEvent<IdentityServerEventContent>, ()>());
-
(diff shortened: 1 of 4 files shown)


bindings/matrix-sdk-ffi/src/ruma.rs
```diff
@@ -43,16 +53,39 @@ use ruma::{
...
-    MatrixToUri, MatrixUri as RumaMatrixUri, OwnedUserId, UInt, UserId,
+    KeyDerivationAlgorithm as RumaKeyDerivationAlgorithm, MatrixToUri, MatrixUri as RumaMatrixUri,
+    OwnedRoomId, OwnedUserId, UInt, UserId,
 };
 use tracing::info;
 
 use crate::{
     error::{ClientError, MediaInfoError},
     helpers::unwrap_or_clone_arc,
+    notification_settings::{Action, PushCondition},
```
Johennes, line 88: The mapping between matrix-sdk-ffi types and ruma types for pushes is now split between `notification_settings.rs` and `ruma.rs`. I wasn't sure if maybe we'd want all of the mappers in `ruma.rs`?

bindings/matrix-sdk-ffi/src/client.rs
```diff
@@ -534,6 +563,198 @@ impl Client {
...
+
+                Arc::new(TaskHandle::new(get_runtime_handle().spawn(async move {
+                    let mut subscriber = observer.subscribe();
+                    loop {
+                        if let Some(next) = subscriber.next().await {
+                            listener.on_change(next.0.into());
+                        }
+                    }
+                })))
+            }
```
Johennes, line 592: There's quite a bit of duplicated boilerplate between the various match arms. I couldn't find a good way to avoid this but maybe somebody with stronger rust-fu has a tip?

bindings/matrix-sdk-ffi/src/ruma.rs
```diff
@@ -985,3 +1018,591 @@ pub fn content_without_relation_from_message(
...
+    IdentityServer,
+    /// m.ignored_user_list
+    IgnoredUserList,
+    /// m.push_rules
+    PushRules,
+    /// m.secret_storage.default_key
+    SecretStorageDefaultKey,
+    /// m.secret_storage.key.*
+    SecretStorageKey { key_id: String },
+}
```
poljar, line 1037: There's a lot of duplication of Ruma types here. I remember the Ruma folks being open towards adding `uniffi` attributes to Ruma types directly.

It's prbobably a big ask, but would you be open on attempting this?

The account data types might be sufficiently small that this might not be too big of a problem.

bindings/matrix-sdk-ffi/src/notification_settings.rs
```diff
@@ -161,7 +161,7 @@ pub enum PushCondition {
 }
 
 impl TryFrom<SdkPushCondition> for PushCondition {
-    type Error = ();
+    type Error = String;
```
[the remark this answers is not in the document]
poljar: Yeah, `String` is always better than `()`.

bindings/matrix-sdk-ffi/src/client.rs
poljar: How about a macro?

Something like:

```rust
($t:ty) => {{
    // Using an Arc here is mandatory or else the subscriber will never trigger
    let observer = Arc::new(self.inner.observe_events::<$t, ()>());

    Arc::new(TaskHandle::new(get_runtime_handle().spawn(async move {
        let mut subscriber = observer.subscribe();
        loop {
            if let Some(next) = subscriber.next().await {
                listener.on_change(next.0.into());
            }
        }
    })))
}};
```

poljar (changes requested):
I left some suggestions, I think at least the macro would improve this quite a bit.

Velin92:
I have a question about this PR since it may be very useful for: https://github.com/matrix-org/matrix-rust-sdk/pull/5014

I see that you can subscribe to either stable or unstable content, couldn't the SDK unify the stable and unstable and internally use the appropriate the stable one when available, or how would that be achieved without having to do the check client side?

Johennes:
> I see that you can subscribe to either stable or unstable content, couldn't the SDK unify the stable and unstable and internally use the appropriate the stable one when available, or how would that be achieved without having to do the check client side?

Good point. I think we can map both the stable and the unstable type from Ruma to a single FFI type. I can make that change for `MarkedUnread` and `UnstableMarkedUnread`.

> Also what if the value is missing and we want to return a Default?

Hm, I'm not sure. Maybe we could return `None` and leave the default to the caller? Or do you have a specific event type in mind where the SDK could define a default value?

bindings/matrix-sdk-ffi/src/ruma.rs
Johennes: > I remember the Ruma folks being open towards adding uniffi attributes to Ruma types directly.

Oh, I wasn't aware. That would actually be amazing. The mapping of Ruma to FFI types adds a lot of boilerplate in many places.

I can try to push things done. It might just take me a little. Just saying in case there's any urgency on https://github.com/matrix-org/matrix-rust-sdk/pull/5014 (or its successor).

poljar: I think we can also merge this as is, and if/when `uniffi` attributes are added to the Ruma type we can just switch over.

Johennes:
> Good point. I think we can map both the stable and the unstable type from Ruma to a single FFI type. I can make that change for `MarkedUnread` and `UnstableMarkedUnread`.

Err, actually that doesn't work because then you might get two competing updates on the listener when both events exist. I think the logic needs to go down into the SDK. It should only trigger the observer for the unstable event if the stable one doesn't exist. Not sure how hard this would be given the generic way that event listening is set up though.

PR #4617, opened 2025-02-04
feat(event cache store): add an IndexedDB implementation

ospfranco:
This PR is WIP.

This PR is meant to add a EventCache store for the the `matrix-sdk-indexeddb` crate of the SDK meant to be used in WASM contexts. Similar to the sqlite implementation is supposed to provide a disk persisted implementation for storing Events in LinkedChunks. 

It follows the general structure of `matrix-sdk-indexeddb/src/crypto_store` and `matrix-sdk-indexeddb/src/state_store`, a new module `matrix-sdk-indexeddb/src/event_cache_store` has been added.

There are still many TODOs and questions in the codebase to which I would appreciate some guidance from the SDK maintainers.

- [ ] Public API changes documented in changelogs (optional)

Signed-off-by: Oscar Franco

PR #4996, opened 2025-05-02
IndexedDB: Make `IndexeddbSerializer` shareable between modules in `matrix-sdk-indexeddb`

mgoldenberg:
## Background

This pull request has a high-level focus on integrating work on an IndexedDB implementation of the `EventCacheStore` (see #4617). Those changes were rather large and sprawling and included code duplication, so this is part of a series of pull requests which will break that work into smaller, more manageable changes.

## Changes

This change is focused on de-duplicating code. In #4617, `crypto_store::indexeddb_serializer` was copy-and-pasted to `event_cache_store::indexeddb_serializer` with minor modifications (see [here](https://github.com/matrix-org/matrix-rust-sdk/pull/4617#discussion_r1940789943)). In preparation for integrating this work, I have made the following changes.

1. Moved the `IndexeddbSerializer` to the top-level so it may be shared between modules in the crate.
2. Created a new error type, `IndexeddbSerializerError` to be used with `IndexeddbSerializer`.
    1. Previously, it was using `IndexeddbCryptoStoreError`, which was too specific for a generalized use-case.

Additionally, I have prioritized making changes in such a way that does not change the public API of the crate. This means some of the changes weren't as sensible as they could have been, but it helped to keep the pull request small.

## Future Work

The next pull request in this series will make additions to the `IndexeddbSerializer` - there are two functions which #4617 added to its copy of the serializer and which are used by `IndexeddbEventCacheStore`.

---


mgoldenberg:
Apologies about the failures, I am fixing them now!

Hywan (approved):
Thanks for your patches. They are making sense to me. Thanks for doing small patches, I appreciate.

PR #4998, opened 2025-05-02
feat(crypto): Add new `encrypt_and_send_custom_to_device` to the client

BillCarsonFr:
Created a follow-up PR to extract the mock helpers with the other mock helpers, see https://github.com/matrix-org/matrix-rust-sdk/pull/5005

Signed-off-by:


crates/matrix-sdk-crypto/src/machine/mod.rs
```diff
@@ -1111,6 +1111,49 @@ impl OlmMachine {
...
+    ///
+    /// # Returns
+    /// A list of `ToDeviceRequest` to send out the event, and the list of
+    /// devices where encryption did not succeed (device excluded or no olm)
+    pub async fn encrypt_content_for_devices(
+        &self,
+        devices: Vec<DeviceData>,
+        event_type: &str,
+        content: &Value,
+    ) -> OlmResult<(Vec<ToDeviceRequest>, Vec<(DeviceData, WithheldCode)>)> {
```
poljar, line 1132: Something I mentioned on the [to-device PR for widgets](https://github.com/matrix-org/matrix-rust-sdk/pull/4963/files#r2077618445), and in a conversation with you as well.

Is this the right API shape we'd want here? Let people collect devices and then share things with them?

It sounds more like we would want an API where you pass a set of user IDs and then the crypto crate collects the devices using the usual mechanisms.

The API would essentially be a customized version of `share_room_key()` method. We can call it `share_secret_with_room()`. That method will then respect the users wishes with regards to the `CollectStrategy`.

This would also solve your `TODO` item at the start of this method. 

As this is implemented, each and every widget would need to reimplement collect strategies. Which I think is just too big of a task.

poljar (commented):
Not reviewed in fully yet, but there's a bit to discuss.

crates/matrix-sdk-crypto/src/machine/mod.rs
BillCarsonFr: > It sounds more like we would want an API where you pass a set of user IDs and then the crypto crate collects the devices using the usual mechanisms.

For the VOIP use case, users joins the call per device (call join events are per device). So the user knows exaclty to what (`user_id,device_id`) he want to encrypt to/

poljar: Ok, but how do you know which `user_id` and `device_id` should participate? Do you inspect the device verification status? 

Furthermore if you base this on the `device_id`, do you protect against the fact that the underlying cryptographic device that is identified by a `user_id` and `device_id` can be switched out?

BillCarsonFr: For now the "trust/secure" concerns are out of scope.
Yes, I think we would have to expose/inspect the verification status.

> user_id and device_id can be switched out

Yes I know, I have several comments that I want address in the MSC regarding usage of claimed/insecure info and usage of device_id instead of device key in call membership events. 
This is for the next work package

poljar: Alright, in that case let's put the public method in the `matrix-sdk` crate behind a `experimental-something` feature flag.

That will prevent people for shooting themselves in the foot with this and we can see if the API needs some more reconsideration without much fear.

BillCarsonFr: > Alright, in that case let's put the public method in the `matrix-sdk` crate behind a `experimental-something` feature flag.
> 
> That will prevent people for shooting themselves in the foot with this and we can see if the API needs some more reconsideration without much fear.

I give it a try https://github.com/matrix-org/matrix-rust-sdk/pull/4998/commits/2254e7211d41278952f50d1853c61f55fb2524c7

First time I do a feature flag (I looked at automatic-room-key-forwarding), let me know if I missed something

crates/matrix-sdk-crypto/src/machine/mod.rs
```diff
@@ -1111,6 +1115,50 @@ impl OlmMachine {
...
+    ///
+    /// # Returns
+    /// A list of `ToDeviceRequest` to send out the event, and the list of
+    /// devices where encryption did not succeed (device excluded or no olm)
+    #[cfg(feature = "experimental-send-custom-to-device")]
+    pub async fn encrypt_content_for_devices(
+        &self,
+        devices: Vec<DeviceData>,
+        event_type: &str,
+        content: &Value,
```
poljar, line 1136: I think we'll want to take a `&Raw` here, just like `OutboundGroupSession::encrypt()` does.

But ok, we can tweak the types later on.

poljar (changes requested):
Alright, did a fuller review, there are some docs that are wrong and I left some nits, but I think this can soon go in as an experimental feature.

crates/matrix-sdk-crypto/src/machine/mod.rs
```diff
@@ -1111,6 +1115,50 @@ impl OlmMachine {
         self.inner.group_session_manager.share_room_key(room_id, users, encryption_settings).await
     }
 
+    /// Encrypts the given content using Olm for each of the given devices.
```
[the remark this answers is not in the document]
BillCarsonFr: Done here https://github.com/matrix-org/matrix-rust-sdk/pull/4998/commits/d875aa7d2ce80871e42571b96da347e0d6a3bf6c

BillCarsonFr:
> Alright, did a fuller review, there are some docs that are wrong and I left some nits, but I think this can soon go in as an experimental feature.

Thx for the reviews and comments. I have updated the PR to adress the comments. I also added a changelog entry

crates/matrix-sdk/CHANGELOG.md
```diff
@@ -25,6 +25,10 @@ All notable changes to this project will be documented in this file.
 - `Room::set_unread_flag()` now sets the stable `m.marked_unread` room account data, which was
   stabilized in Matrix 1.12. `Room::is_marked_unread()` also ignores the unstable
   `com.famedly.marked_unread` room account data if the stable variant is present.
+- `Encryption::encrypt_and_send_raw_to_device`: Introduced as an experimental method for
+  sending custom encrypted to-device events. This feature is gated behind the
+  `experimental-send-custom-to-device` flag, as it remains under active development and may undergo changes.
+
```
poljar, line 31: Please add a link to the PR.

poljar (changes requested):
One last thing and I think it's good to merge.

crates/matrix-sdk/CHANGELOG.md
BillCarsonFr: https://github.com/matrix-org/matrix-rust-sdk/pull/4998/commits/230c9581589f71317d9c3c6462359818ff33a96d

Issue #4976, opened 2025-04-24
When `Timeline` only contains `TimelineStart`, new events are pushed in front of `TimelineStart`


mgoldenberg:
## Problem

When creating a `Timeline` that is empty, due for instance to a certain kind of filtering, we can `Timeline.paginate_backwards` until we reach the start of the `Timeline`, at which point, the `Timeline` only contains a `TimelineStart`.

This can be confirmed by checking that `Timeline.latest_event` returns `None` and by `Timeline.subscribe`ing and inspecting the returned `TimelineItem`s.

Once new events start arriving, however, the `VectorDiff`s report that the new events are being pushed in front of the `TimelineStart` and this can be confirmed by inspecting that `Timeline.latest_event` continues returning `None`.

**Note:** that when the `Timeline` has even one event aside from `TimelineStart` before calling `Timeline.subscribe`, then new events are properly push behind `TimelineStart` and `Timeline.latest_event` returns the expected event.

### Code

I have adapted the `example/timeline` crate to illustrate the problem. For the full crate, see [`mgoldenberg:example-sticky-timeline-start`](https://github.com/mgoldenberg/matrix-rust-sdk/tree/example-sticky-timeline-start/examples/sticky-timeline-start).

```rust
use anyhow::Result;
use clap::Parser;
use futures_util::StreamExt;
use matrix_sdk::{
    config::SyncSettings,
    ruma::{
        api::client::room::create_room::v3::Request,
        events::{room::message::RoomMessageEventContent, AnySyncTimelineEvent},
    },
    Client,
};
use matrix_sdk_ui::timeline::{default_event_filter, RoomExt};
use url::Url;

#[derive(Parser, Debug)]
struct Cli {
    /// The homeserver to connect to.
    #[clap(value_parser)]
    homeserver: Url,

    /// The user name that should be used for the login.
    #[clap(value_parser)]
    user_name: String,

    /// The password that should be used for the login.
    #[clap(value_parser)]
    password: String,

    /// Set the proxy that should be used for the connection.
    #[clap(short, long)]
    proxy: Option<Url>,

    /// Enable verbose logging output.
    #[clap(short, long, action)]
    verbose: bool,

    /// An initial message to send in the room before subscribing to timeline events
    #[clap(short, long)]
    initial_message: Option<String>,

    /// Number of messages to send after subscribing to timeline events
    #[clap(short, long, default_value = "3")]
    num_messages_after_subscribing: usize,
}

[...]

PR #5000, opened 2025-05-05
fix(ui): Introduce `Timeline` regions

Hywan:
This patch is twofold.

### First part

This patch fixes the insertion of a new `TimelineItem` in the presence of a `TimelineStart` that shifts/offsets the timeline index of 1. It fixes https://github.com/matrix-org/matrix-rust-sdk/issues/4976. It also contains a regression test.

### Second part

Based on https://github.com/matrix-org/matrix-rust-sdk/pull/4816#pullrequestreview-2705148141, I'm proposing a new concept in `timeline::controller::ObservableItems`: **regions**.

The `ObservableItems` holds all the invariants about the _position_ of the items. It defines three regions where items can live:

1. the _start_ region, which can only contain a single `TimelineStart`,
2. the _remotes_ region, which can only contain many `Remote` timeline items with their decorations (only `DateDivider`s and `ReadMarker`s),
3. the _locals_ region, which can only contain many `Local` timeline items with their decorations (only `DateDivider`s).

The `iter_all_regions` method allows to iterate over all regions. `iter_remotes_region` will restrict the iterator over the _remotes_ region, and so on. These iterators provide the absolute indices of the items, so that it's harder to make mistakes when manipulating the indices of items with operations like `insert`, `remove`, `replace` etc.

Other methods like `push_local` or `push_date_divider` insert the items in the correct region, and check a couple of invariants. I've first introduced the invariants, then 12 tests were failing, which indicates that the `TimelineStart` could have been a problem in multiple situations. Moving the code to use `iter_remotes_regions` & siblings have fixed the tests, which indicates that this class of bugs has been fixed.

### Review

It's better to review patch-by-patch. There are quite small and I hope easy to understand. The concept of _regions_ removes an entire class of bugs in our `Timeline`, and I'm pretty happy with that 🙂.

---

* Fix https://github.com/matrix-org/matrix-rust-sdk/issues/4976
* Address https://github.com/element-hq/customer-success/issues/482

### DIFF
--- a/crates/matrix-sdk-ui/src/timeline/controller/observable_items.rs
+++ b/crates/matrix-sdk-ui/src/timeline/controller/observable_items.rs
@@ -303,6 +303,31 @@ impl<'observable_items> ObservableItemsTransaction<'observable_items> {
             .timeline_item_has_been_inserted_at(self.items.len().saturating_sub(1), event_index);
     }
 
+    /// Push a new [`TimelineStart`] virtual timeline item.
+    ///
+    /// # Invariant
+    ///
+    /// A [`TimelineStart`] is always the first item if present..
+    ///
+    /// # Panics
+    ///
+    /// It panics if the provided `timeline_item` is not a [`TimelineStart`].
+    ///
+    /// [`TimelineStart`]: super::VirtualTimelineItem::TimelineStart
+    pub fn push_timeline_start_if_missing(&mut self, timeline_item: Arc<TimelineItem>) {
+        assert!(
+            timeline_item.is_timeline_start(),
+            "The provided `timeline_item` is not a `TimelineStart`"
+        );
+
+        // The timeline start virtual item is necessarily the first item.
+        if self.get(0).is_some_and(|item| item.is_timeline_start()) {
+            return;
+        }
+
+        self.push_front(timeline_item, None);
+    }
+
     /// Clear all timeline items and all remote events.
     pub fn clear(&mut self) {
         self.items.clear();
@@ -390,7 +415,7 @@ mod observable_items_tests {
         controller::{EventTimelineItemKind, RemoteEventOrigin},
         event_item::RemoteEventTimelineItem,
         EventTimelineItem, Message, MsgLikeContent, MsgLikeKind, TimelineDetails,
-        TimelineItemContent, TimelineUniqueId,
+        TimelineItemContent, TimelineUniqueId, VirtualTimelineItem,
     };
 
     fn item(event_id: &str) -> Arc<TimelineItem> {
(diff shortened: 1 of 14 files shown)


crates/matrix-sdk-ui/src/timeline/date_dividers.rs
```diff
@@ -192,8 +192,16 @@ impl DateDividerAdjuster {
 
         // Only record the initial state if we've enabled the trace log level, and not
         // otherwise.
-        let initial_state =
-            if event_enabled!(Level::TRACE) { Some(items.iter().cloned().collect()) } else { None };
+        let initial_state = if event_enabled!(Level::TRACE) {
+            Some(
+                items
+                    .iter_remotes_and_locals_regions()
```
andybalaam, line 198: For recording the initial state, don't we want all regions?

crates/matrix-sdk-ui/src/timeline/date_dividers.rs
```diff
@@ -599,7 +604,14 @@ impl Display for DateDividerInvariantsReport<'_, '_> {
             }
 
             writeln!(f, "\nFinal state:")?;
-            write_items(f, self.final_state.iter().cloned().collect::<Vec<_>>().as_slice())?;
+            write_items(
+                f,
+                self.final_state
+                    .iter_remotes_and_locals_regions()
```
andybalaam, line 610: Again, maybe we want to write out the full state including the start event. Or maybe not - not sure.

crates/matrix-sdk-ui/src/timeline/controller/observable_items.rs
```diff
@@ -553,6 +553,7 @@ impl<'e> ObservableItemsTransactionIterBuilder<'e> {
     }
 
     /// Build the iterator.
+    #[allow(clippy::iter_skip_zero)]
```
andybalaam, line 556: Would not be needed if you like my suggestion, but maybe you like the regularity of always skipping something.

andybalaam: Oh wait, you did it so you can return a concrete type! I guess I am ok with that then...

A `Box<dyn Iterator>` would be simpler, but there is a cost...

andybalaam (approved):
Really good change, thanks!

crates/matrix-sdk-ui/src/timeline/controller/observable_items.rs
```diff
@@ -369,13 +435,230 @@ impl<'observable_items> ObservableItemsTransaction<'observable_items> {
...
+            //
+            // This combination isn't implemented yet (because it contains a hole), but it's also
+            // not necessary in our current code base; it's fine to ignore it.
+            (true, false, true) => unimplemented!(
+                "Iterating over the start and the locals regions is not implemented yet"
+            ),
+
+            // The remotes and the locals regions.
+            (false, true, true) => {
+                iter.skip(size_of_start_region).take(size_of_remotes_region + size_of_locals_region)
```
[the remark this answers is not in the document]
Hywan: Same as https://github.com/matrix-org/matrix-rust-sdk/pull/5000#discussion_r2084698995

crates/matrix-sdk-ui/src/timeline/date_dividers.rs
Hywan: It's the initial state for the date divider: it doesn't need all the regions; only the remotes and locals regions!

crates/matrix-sdk-ui/src/timeline/controller/observable_items.rs
Hywan: There is a cost indeed. Also, not all `Iterable` implements `ExactSizeIterator` nor `DoubleEndedIterator`, which is a requirement. We could use `Box<dyn Iterator + ExactSizeIterator + DoubleEndedIterator>` but I'm not sure I want this for the moment. We don't need _this power_ for the moment, a concrete type is fine.

crates/matrix-sdk-ui/src/timeline/controller/observable_items.rs
```diff
@@ -315,16 +350,47 @@ impl<'observable_items> ObservableItemsTransaction<'observable_items> {
...
+    ) {
+        assert!(
+            timeline_item.is_date_divider(),
+            "The provided `timeline_item` is not a `DateDivider`"
+        );
+
+        // We are not inserting in the start region.
+        if timeline_item_index == 0 {
+            assert!(
+                matches!(self.items.get(timeline_item_index), Some(timeline_item) if !timeline_item.is_timeline_start())
```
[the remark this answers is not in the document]
Hywan: You're right! `self.items` must not be empty in that case. Updated!

crates/matrix-sdk-ui/src/timeline/controller/observable_items.rs
```diff
@@ -369,13 +435,230 @@ impl<'observable_items> ObservableItemsTransaction<'observable_items> {
...
+    ///
+    /// [`Local`]: super::EventTimelineItemKind::Local
+    pub fn has_local(&self) -> bool {
+        matches!(self.items.last(), Some(timeline_item) if timeline_item.is_local_echo())
+    }
+
+    /// Return the index where to insert the first remote timeline
+    /// item.
+    pub fn first_remotes_region_index(&self) -> usize {
+        if self.items.get(0).is_some_and(|item| item.is_timeline_start()) {
```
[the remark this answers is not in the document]
Hywan: What check would you like that add?

crates/matrix-sdk-ui/src/timeline/controller/observable_items.rs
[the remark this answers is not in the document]
Hywan: I've added 4-5 more tests to ensure everything works correctly, including the invariants. Thanks!

PR #5004, opened 2025-05-05
Test: Replace `sync_timeline_event!` with `EventFactory` in notification integration test

yousefmoazzam:
Part of #3716

I think I covered almost all the information in the events that were replaced by using `EventFactory` instead, the only thing missing is the "mention" of a user in the first message. I had a look at `EventFactory` and didn't see anything that immediately jumped out as being able to handle this, and from the assertions in the test I thought that maybe the "mention" in the first message wasn't fundamental to what the test was checking, so I left out the "mention" for now.

Please feel free to let me know if that was an incorrect assumption, I'm happy to make changes accordingly, and the same for any other feedback :slightly_smiling_face: 

### DIFF
--- a/crates/matrix-sdk/tests/integration/notification.rs
+++ b/crates/matrix-sdk/tests/integration/notification.rs
@@ -2,15 +2,10 @@ use assert_matches2::assert_matches;
 use matrix_sdk::{config::SyncSettings, sync::Notification};
 use matrix_sdk_base::deserialized_responses::RawAnySyncOrStrippedTimelineEvent;
 use matrix_sdk_test::{
-    async_test, stripped_state_event, sync_state_event, sync_timeline_event, test_json,
+    async_test, event_factory::EventFactory, stripped_state_event, sync_state_event, test_json,
     InvitedRoomBuilder, JoinedRoomBuilder, SyncResponseBuilder,
 };
-use ruma::{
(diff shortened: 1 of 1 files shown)


yousefmoazzam:
Apologies for the failing CI job. Taking a quick look at the step that's failing in the job, it looks like a lot of go tests are failing with a similar error about a conflict in container naming:
```
❌ TestAliceBobEncryptionWorks/{rust_hs1}|{rust_hs1} (0s)
  2025/05/05 17:05:09 Sharing [SERVER_NAME=hs1 SYNAPSE_COMPLEMENT_DATABASE=sqlite] host environment variables with container
      test_package.go:183: CreateDirtyDeployment failed: CreateDirtyServer: Failed to deploy image homeserver : Error response from daemon: Conflict. The container name "/complement_crypto_dirty_hs1" is already in use by container "093dd0783bb2a5820aeec275fd9437f5939c01d9640fb6a4598595d120816c5e". You have to remove (or rename) that container to be able to reuse that name.
```

This looks like it could be some issue with the CI job (as far as container creation), unrelated to the changes made in the PR, but without having much knowledge of the repo setup it's quite possible the changes I made could have caused that.

Do let me know if my changes need to be amended to fix that, any pointers would be appreciated!

poljar (approved):
Thanks. The test failure was indeed random.

PR #5005, opened 2025-05-06
refactor(test): Extract common crypto mock server helper

BillCarsonFr:
Extracted a helper to create a wiremock server that allows to perform full end to end encryption tests.

Note: In the existing verification test there was a `known_devices` list, but this was not used (the test work fine if you remove it), so I removed it when extracting the mock helpers

Signed-off-by:

### DIFF
--- a/crates/matrix-sdk/src/encryption/mod.rs
+++ b/crates/matrix-sdk/src/encryption/mod.rs
@@ -99,6 +99,7 @@ pub use matrix_sdk_base::crypto::{
     SessionCreationError, SignatureError, VERSION,
 };
 
+use crate::config::RequestConfig;
 pub use crate::error::RoomKeyImportError;
 
 /// All the data related to the encryption state.
@@ -570,6 +571,20 @@ impl Client {
         );
 
         self.send(request).await
+    }
+
+    pub(crate) async fn send_to_device_with_config(
+        &self,
+        request: &ToDeviceRequest,
+        config: RequestConfig,
+    ) -> HttpResult<ToDeviceResponse> {
+        let request = RumaToDeviceRequest::new_raw(
+            request.event_type.clone(),
+            request.txn_id.clone(),
+            request.messages.clone(),
+        );
+
+        self.send_inner(request, Some(config), Default::default()).await
     }
 
     pub(crate) async fn send_verification_request(
--- a/crates/matrix-sdk/src/test_utils/mocks/encryption.rs
+++ b/crates/matrix-sdk/src/test_utils/mocks/encryption.rs
@@ -1,0 +1,578 @@
+// Copyright 2024 The Matrix.org Foundation C.I.C.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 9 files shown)


crates/matrix-sdk/src/test_utils/mocks/encryption.rs
```diff
@@ -0,0 +1,578 @@
...
+///     .unwrap()
+///     .expect("alice sees bob's device");
+/// ```
+pub struct MatrixKeysMockServer {
+    /// The underlying [`wiremock`] [`MockServer`]
+    pub server: MockServer,
+    keys: Arc<Mutex<Keys>>,
+    token_to_user_id_map: Arc<Mutex<BTreeMap<String, OwnedUserId>>>,
+    token_counter: AtomicU32,
+}
```
poljar, line 100: This doesn't really incorporate the `Keys` struct into `MatrixMockServer` instead it creates another mock server concept with much less functionality.

I don't think we'd want to have two of those.

poljar (changes requested):

crates/matrix-sdk/src/test_utils/mocks/encryption.rs
BillCarsonFr: Got it, I did refactor it to extend the existing MockServer https://github.com/matrix-org/matrix-rust-sdk/pull/5005/commits/9766ca15e99db1f9dc2cdda7e03d2d3966cbd669
I kept the crypto API mocking in the new file for code organisation, Is it ok?

I did a second commit to take benefit of the refactoring and re-use a functionnality already in the base mock server instead of using another custom method https://github.com/matrix-org/matrix-rust-sdk/pull/5005/commits/fcb9070d497eaaaf2b5c1065fb209f124267632b

crates/matrix-sdk/src/encryption/mod.rs
```diff
@@ -1786,15 +1786,15 @@ impl Encryption {
...
+            let ruma_request = RumaToDeviceRequest::new_raw(
                 request.event_type.clone(),
                 request.txn_id.clone(),
                 request.messages.clone(),
             );
 
             let send_result = self
                 .client
-                .send_inner(request, Some(RequestConfig::short_retry()), Default::default())
+                .send_inner(ruma_request, Some(RequestConfig::short_retry()), Default::default())
```
poljar, line 1797: Was this added by accident? It's not really topical to the PR.

crates/matrix-sdk/src/test_utils/mocks/encryption.rs
```diff
@@ -0,0 +1,488 @@
...
+/// * Start by creating the mock server like this [`MatrixMockServer::new`].
+/// * Then mock the crypto API endpoints
+///   [`MatrixMockServer::mock_crypto_endpoints_preset`].
+/// * Create your test client using
+///   [`MatrixMockServer::client_builder_for_crypto_end_to_end`], this is
+///   important as it will set up an access token that will allow to know what
+///   client is doing what request.
+///
+/// The [`MatrixMockServer::set_up_alice_and_bob_for_encryption`] will set up
+/// two olm machines aware of each other and ready to communicate.
```
poljar, line 65: That won't get rendered anywhere, right?

poljar (approved):
I would have preferred some doc examples on the individual methods instead on the `impl` block which won't be shown anywhere besides in the source as far as I know.

Some renaming wouldn't have hurt either. I think we can address that in separate PRs so we don't block your stack of PRs on this for too long.

PR #5009, opened 2025-05-07
refactor(widget): Use streams to streamline the action processing logic

poljar:


andybalaam (approved):

crates/matrix-sdk/src/widget/mod.rs
```diff
@@ -159,23 +169,17 @@ impl WidgetDriver {
...
-                    &incoming_msg_tx,
-                    &capabilities_provider,
-                    action,
-                )
+        // Let's combine our set of initial actions with the stream of received actions.
+        let mut combined = tokio_stream::iter(initial_actions).chain(stream);
+
+        // Let's now process all actions we receive forever.
+        while let Some(action) = combined.next().await {
+            self.process_action(&matrix_driver, &incoming_msg_tx, &capabilities_provider, action)
```
toger5, line 181: Is it possible to change it to:
```rust
let stream = tokio_stream::iter(initial_actions).chain(
    UnboundedReceiverStream::new(incoming_msg_rx)
    .flat_map(|message| tokio_stream::iter(widget_machine.process(message))
).flat_map(|action| self.process_action(&matrix_driver, &incoming_msg_tx, &capabilities_provider, action);
// awaiting on the stream...
```
or similar, to make it more obvious that the main thing that is happeing is: `widget_machine.process` -> `self.process_action`.

toger5 (approved):
This is nice.
The while loops makes total sens but I was searching for a `flat_map(process_action)` similar to `flat_map(process)`.

Is there a way to do this?
In the comment i did some pseudo-code for what I was searching.

But probably there are good reasons to the while loop as well?

crates/matrix-sdk/src/widget/mod.rs
poljar: Well once you get to `self.process_action()` you are not mapping pure data types between each other. As the name of the function suggests, you are executing an action.

If you take a look at the function signature of `map()`, it takes a function that converts the item of the stream `T1` into a different item `T2`. It might help to remind ourselves that most functional tools we have in our programming toolset have a direct counterpart in mathematics. And look, it's even using the same name: https://en.wikipedia.org/wiki/Map_(mathematics).

`flatten()` on the other hand takes a list of lists and produces a single list, or in this case a stream of streams.

`flat_map()` combines the two concepts.

So `flat_map()` is certainly not what you are looking for. What you seem to be looking for is the `then()` function.

But I would argue that you don't want to use `then()`. If we take a look at the Javascript world, futures are equivalent to promises, and look we again have a `then()` function:

```javascript
fetchData()
  .then((data) => {
    return data.toUpperCase();
  })
  .then((upperData) => {
    console.log(upperData);
  })
  .catch((error) => {
    console.error("Error:", error);
  });
```

But any sufficiently complex usage the `then()` function results in deeply nested code, each promise adds more nesting. Thus the `async/await` syntax was introduced as syntactic sugar:

```javascript
async function handleData() {
  try {
    const data = await fetchData();
    const upperData = data.toUpperCase();
    console.log(upperData);
  } catch (error) {
    console.error("Error:", error);
  }
}

handleData();
```

Mind you, all of this probably can trace it's roots to functional programming and perhaps even the `do` notation for Monads in Haskell[[1]]. Look, this time we have a `then` operator `>>`:

```haskell
putStr "Hello" >> 
putStr " " >> 
putStr "world!" >> 
putStr "\n"
```

Put the "Hello" string, then put a space character, then put "world", and finally put a newline. Now let's take a look at their syntactic sugar they introduced to make life simpler:

```haskell
do { putStr "Hello"
   ; putStr " "
   ; putStr "world!"
   ; putStr "\n" }
```

Wait a minute, but that's almost the same, well yes and no. Each put statement, or action is separated by a ;. And the do notation has flattened out the single nested statement from above into 4 separate statements.

It's not really noticeable here, until your actions start to return values or you need to handle errors. Let's take a quick look at a more complex case:

```haskell
action1 >>= (\ x1 -> action2 >>= (\ x2 -> mk_action3 x1 x2 ))
```

This time the `bind` (`>>=`) operator is used, we call `action1` and we bind the result of `action1` to `x1`, we call `action2` and bind the result to `x2`, finally we call `mk_action3` and provide `x1` and `x2` as arguments to it.

This is quite tough to read, but in `do` notation:

```haskell
do { x1 <- action1
   ; x2 <- action2
   ; mk_action3 x1 x2 }
```

Hold on... This looks just like an imperative language now. Well yes, the nesting has been flattened out just like in our previous examples.

What does this all have to do with Rust? Well it's the same. Futures not only behave like Monads, they are and instance of a Monad. We have introduced `async/await` as our special purpose `do` notation to make our life simpler.

Do note, that the same thing happened with `Option<T>` and `Result<T>` which also can produce deeply nested logic, that's why we introduced yet another instance of the `do` notation, the `?` operator.

To finally conclude this overly long comment, a general rule of thumb is to use functional patterns when you operate on pure data, but as soon as you have actions you should use the `do`, err `async/await` and imperative notation.
 
[1]: https://en.wikibooks.org/wiki/Haskell/do_notation#Translating_the_then_operator

PR #5011, opened 2025-05-07
FFI: Expose the QrCodeData server name.

pixlwave:
Small PR to expose the scanned `server_name` in QR code data so that clients with e.g. MDM can choose not to continue with the authorisation for particular servers.


poljar (approved):
Thanks.

PR #5012, opened 2025-05-07
FFI: Add support for OIDC login hints

pixlwave:
This PR makes 2 changes:
- Add a generic `login_hint` method to the OAuth URL builder (see https://github.com/element-hq/matrix-authentication-service/pull/4512 for more context).
- Expose the generic `login_hint` on the FFI's `url_for_oidc` method.


crates/matrix-sdk/src/authentication/oauth/auth_code_builder.rs
```diff
@@ -73,6 +73,17 @@ impl OAuthAuthCodeUrlBuilder {
         self
     }
 
+    /// Set a generic login hint to help an identity provider pre-fill the login
+    /// form.
+    ///
+    /// Note: This is not the same as the [`Self::user_id_hint()`] method, which
+    /// is specifically designed to a) take a `UserId` and no other type of
+    /// hint and b) be used directly by MAS and not the identity provider.
```
poljar, line 81: Hmm, this overwrites whatever we set with `user_id_hint()`, we should mention this.

Out of interest, what will you set as the hint?

poljar (changes requested):

crates/matrix-sdk/src/authentication/oauth/auth_code_builder.rs
pixlwave: Good point, I've added the same `The following methods are mutually exclusive` line that we use in the client builder.

It will likely be an email address but it depends on the upstream provider, we would pass through whatever is given to us without modifying it. The use case is an organisation creating provisioning links like `https://app.example.com/?server_name=example.org&login_hint=alice` where their upstream provider knows how to handle `alice` as a hint.

There are docs here: https://openid.net/specs/openid-connect-core-1_0.html#AuthRequest

poljar: Ah makes sense. Can we perhaps more explicitly mention this example use-case and link to these docs?

poljar:
Ah, the CI isn't happy yet.

pixlwave:
I don't understand what the failure is:

```
❌ TestAliceBobEncryptionWorks/{rust_hs1}|{rust_hs1} (0s)
  2025/05/07 15:07:48 Sharing [SERVER_NAME=hs1 SYNAPSE_COMPLEMENT_DATABASE=sqlite] host environment variables with container
      test_package.go:183: CreateDirtyDeployment failed: CreateDirtyServer: Failed to deploy image homeserver : Error response from daemon: Conflict. The container name "/complement_crypto_dirty_hs1" is already in use by container "fe7d8290369fd8f61878229944b863e7b67ab074c1f6bfb67228ed36e9945b1c". You have to remove (or rename) that container to be able to reuse that name.
```

pixlwave:
Ah seems like this is hitting [other](https://github.com/matrix-org/matrix-rust-sdk/pull/5013) PRs and the fix is here: https://github.com/matrix-org/complement-crypto/pull/174

MatMaul:
Thanks a lot for that and the corresponding bits in MAS, you beat us to it by a week or 2 I believe :)

PR #5014, opened 2025-05-07
feat: MSC 4278 implementation

Velin92:
This PR updates Ruma and implements new APIs to use the unstable [MSC 4278](https://github.com/matrix-org/matrix-spec-proposals/pull/4278).

This new global account data setting, will be used by clients to decide what policy to follow in display the media previews in the timeline, and the avatars in invite requests

### DIFF
--- a/bindings/matrix-sdk-ffi/src/client.rs
+++ b/bindings/matrix-sdk-ffi/src/client.rs
@@ -47,6 +47,9 @@ use ruma::{
     events::{
         ignored_user_list::IgnoredUserListEventContent,
         key::verification::request::ToDeviceKeyVerificationRequestEvent,
+        media_preview_config::{
+            InviteAvatars as RumaInviteAvatars, MediaPreviews as RumaMediaPreviews,
+        },
         room::{
             history_visibility::RoomHistoryVisibilityEventContent,
             join_rules::{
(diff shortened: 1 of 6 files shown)


crates/matrix-sdk/src/account.rs
```diff
@@ -916,6 +920,88 @@ impl Account {
...
+        {
+            unstable_media_preview_config
+        } else {
+            UnstableMediaPreviewConfigEventContent::new(Default::default(), Default::default())
+        };
+        media_preview_config_to_update.invite_avatars = new_setting;
+        // Updating the account data
+        self.set_account_data(media_preview_config_to_update).await?;
+        // TODO: I think I should reset all the storage and perform a new local sync
+        // here but I don't know how
```
Velin92, line 957: Not really sure what to do here

crates/matrix-sdk/src/account.rs
```diff
@@ -916,6 +920,88 @@ impl Account {
...
+        {
+            unstable_media_preview_config
+        } else {
+            UnstableMediaPreviewConfigEventContent::new(Default::default(), Default::default())
+        };
+        media_preview_config_to_update.media_previews = new_setting;
+        // Updating the account data
+        self.set_account_data(media_preview_config_to_update).await?;
+        // TODO: I think I should reset all the storage and perform a new local sync
+        // here but I don't know how
```
Velin92, line 938: Not really sure what to do here

crates/matrix-sdk/src/account.rs
```diff
@@ -916,6 +920,91 @@ impl Account {
...
+    /// This is an unstable event, so we always default to setting the unstable
+    /// version of the event for now.
+    pub async fn set_media_previews_setting(&self, new_setting: MediaPreviews) -> Result<()> {
+        let mut media_preview_config_to_update = if let Some(unstable_media_preview_config) =
+            self.get_unstable_media_preview_config().await?
+        {
+            unstable_media_preview_config
+        } else {
+            UnstableMediaPreviewConfigEventContent::new(Default::default(), Default::default())
+        };
```
andybalaam, line 934: Would this work?

```
let mut media_preview_config_to_update = self.get_unstable_media_preview_config().await?.unwrap_or_else(|| UnstableMediaPreviewConfigEventContent::new(Default::default(), Default::default()));
```

If so, this would probably match the house style a bit more.

crates/matrix-sdk/src/account.rs
andybalaam: I was actually looking at this recently and was surprised to see the store not getting updated when we set some account data. @poljar is there a way to tell the store that a piece of account data has been changed?

crates/matrix-sdk/src/account.rs
```diff
@@ -916,6 +920,91 @@ impl Account {
...
+    /// Get the current avatars in invite requests display policy for the
+    /// account.
+    pub async fn get_invite_avatars_setting(&self) -> Result<InviteAvatars> {
+        if let Some(media_preview_config) = self.get_media_preview_config().await? {
+            return Ok(media_preview_config.invite_avatars);
+        } else if let Some(unstable_media_preview_config) =
+            self.get_unstable_media_preview_config().await?
+        {
+            return Ok(unstable_media_preview_config.invite_avatars);
+        }
```
andybalaam, line 984: This is tricky to change to `unwrap_or_else` style because the first else part requires `.await` and `?`, both of which don't fit inside a closure.

I would definitely prefer to remove the `return`s by making the whole function a single `if` block i.e. making the last part:

```
} else {
	Default::default()
}
```

andybalaam (changes requested):
This looks great - marking as "request changes" so it pings me when we figure out the remaining questions and you re-request review.

bindings/matrix-sdk-ffi/src/client.rs
```diff
@@ -1207,6 +1210,91 @@ impl Client {
...
+    Off,
+}
+
+impl From<RumaMediaPreviews> for MediaPreviews {
+    fn from(value: RumaMediaPreviews) -> Self {
+        match value {
+            RumaMediaPreviews::On => MediaPreviews::On,
+            RumaMediaPreviews::Private => MediaPreviews::Private,
+            RumaMediaPreviews::Off => MediaPreviews::Off,
+            // Use the default value of On for unhandled cases.
```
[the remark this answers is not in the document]
Velin92: this would mean that we never display any media preview anymore, and the MSC spec defines ON as the default value, since is the original intended behaviour.
So I would say we should be consistent with the spec, and in case an unhandled value will be added in the future we should probably use the default MSC value.

bindings/matrix-sdk-ffi/src/client.rs
```diff
@@ -1207,6 +1210,91 @@ impl Client {
...
+    On,
+    Off,
+}
+
+impl From<RumaInviteAvatars> for InviteAvatars {
+    fn from(value: RumaInviteAvatars) -> Self {
+        match value {
+            RumaInviteAvatars::On => InviteAvatars::On,
+            RumaInviteAvatars::Off => InviteAvatars::Off,
+            // Use the default value of On for unhandled cases.
```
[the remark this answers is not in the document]
Velin92: Same as before, I think that for unhandled values we should default to the MSC value which is `On`, also I imagine that unhandled values would be probably be introduced as a new MSC, and would likely be added later.

crates/matrix-sdk/src/account.rs
Velin92: the problem is that the unstable and stable ruma types are different, so using a single `unwrap_or_else` can't be used.

Maybe I could implement a trait on the SDK side for both of them? But I am not sure if is worth to just remove an elif

crates/matrix-sdk/src/account.rs
Velin92: I just looked at the the `ignore_user` and `unignore_user` function which has the exact same comment. So I assume this is a known issue maybe? I imagine that since the sync might take some time this means that if we change the value and immediately check the account data again, it won't be updated right?
And if so how long may it take to be update, are we talking about seconds or milliseconds? 
Since this may affect how we design the UI.

bindings/matrix-sdk-ffi/src/client.rs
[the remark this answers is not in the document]
andybalaam: Yep, I guess if it says it in the spec, we should follow that :-(

crates/matrix-sdk/src/account.rs
poljar: As long as a sync takes, you can subscribe to listen to account data events. There's actually a PR in flight to add exactly this to the bindings: https://github.com/matrix-org/matrix-rust-sdk/pull/4994.

I suppose you'll have to subscribe to this anyways, in case the value gets updated on another device.

Velin92: nice this would be very useful!

bindings/matrix-sdk-ffi/src/client.rs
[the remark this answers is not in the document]
Half-Shot: (Ftr, the spec was influenced by hard requirements on the product side so we'd need to do this anyway)

Velin92:
I'll close this PR then in favour of this one: https://github.com/matrix-org/matrix-rust-sdk/pull/5017 that only updates Ruma, then I'll open another PR based on https://github.com/matrix-org/matrix-rust-sdk/pull/4994

PR #5015, opened 2025-05-08
Test: Replace `sync_timeline_event!` with `EventFactory` for room create and room member events in room integration test

yousefmoazzam:
Part of #3716

I was working on replacing all uses of `sync_timeline_event!` in this test file, but after seeing that it probably requires adding more than one new method on `EventFactory`, I thought I'd make a PR for a subset of the changes I've done so far which I think are still useful in isolation and self-contained, to keep the diff smaller.

On the other hand, if these code changes alone are too small to warrant a PR then I'm happy to keep adding the rest of the changes to get rid of `sync_timeline_event!` from that file :slightly_smiling_face: 


poljar (approved):
The only PR that is small enough is one that doesn't compile.

Thanks, this looks good.

PR #5019, opened 2025-05-09
fix(tests): Avoid depending on the local time zone in tests

robintown:
This test fails when run in the Americas, because over here the Unix epoch took place in 1969 =D


poljar (approved):
Wat :duck:

PR #5020, opened 2025-05-10
Test: Replace `sync_timeline_event!` with `EventFactory` for room power level events in room integration test

yousefmoazzam:
Part of #3716

I went with a `BTreeMap` for the parameter type on the new helper method on `EventFactory` purely because that's the type of the `RoomPowerLevelsEventContent.users` field, there was no other reason for that choice of map type.


poljar (approved):
Thanks, I think `BTreeMap` is a sensible choice.

PR #5024, opened 2025-05-12
Test: Replace `sync_timeline_event!` with `EventFactory` for room canonical alias events in room integration test

yousefmoazzam:
Part of #3716

I inlined the creation of the `EventFactory` in the second test since it's only used once and I thought it was a tad neater than having a separate binding. Hopefully the extra level of indentation is tolerable, but I'm happy to define the `EventFactory` instance separately if it's too unsightly :slightly_smiling_face: 


Hywan (approved):
Well done, thank you!

PR #5025, opened 2025-05-13
feat(state store): enable the busy timeout to automatically retry operations on busy DBs

bnjbvr:
Under some very particular circumstances, the "database is locked" error can still happen, even in WAL mode, even if the database connection is not being upgraded from a read transaction to a write transaction.

We *think* this might be the reason behind errors like https://github.com/element-hq/element-x-ios/issues/3582, so we're enabling the sqlite busy_timeout, which will retry the operation after a short sleep, until the busy timeout is being hit.

See also https://github.com/element-hq/element-x-ios/issues/3582#issuecomment-2875836411

### DIFF
--- a/crates/matrix-sdk-sqlite/src/state_store.rs
+++ b/crates/matrix-sdk-sqlite/src/state_store.rs
@@ -446,7 +446,18 @@ impl SqliteStateStore {
     }
 
     async fn acquire(&self) -> Result<SqliteAsyncConn> {
-        Ok(self.pool.get().await?)
+        let conn = self.pool.get().await?;
+
+        // Specify a busy timeout so that operations are automatically retried, in case
+        // the database was marked as locked, which can happen under very
+        // peculiar circumstances in WAL mode.
+        //
+        // The timeout value is in milliseconds.
+        //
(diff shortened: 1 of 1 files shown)


crates/matrix-sdk-sqlite/src/state_store.rs
```diff
@@ -446,7 +446,16 @@ impl SqliteStateStore {
...
     async fn acquire(&self) -> Result<SqliteAsyncConn> {
-        Ok(self.pool.get().await?)
+        let conn = self.pool.get().await?;
+
+        // Specify a busy timeout so that operations are automatically retried, in case
+        // the database was marked as locked, which can happen under very
+        // peculiar circumstances in WAL mode.
+        //
+        // See also https://www.sqlite.org/wal.html#sometimes_queries_return_sqlite_busy_in_wal_mode.
+        conn.execute_batch("PRAGMA busy_timeout = 2000;").await?;
```
poljar, line 456: Would be nice to specify that this is 2000 milliseconds.

It's not clear without searching through the docs.

I would probably define a `const` `Duration` for this, so the type tells us which unit we're using.

Hywan: Do we want to add this `PRAGMA` to other databases?

PR #5030, opened 2025-05-13
refactor(timeline): get rid of `HandleEventResult`

bnjbvr:
Another set of simplifications after #4655 landed. I'm a bit less sure about the final commits, but this complexity is hard to tame…


crates/matrix-sdk-ui/src/timeline/controller/state_transaction.rs
```diff
@@ -229,6 +232,55 @@ impl<'a> TimelineStateTransaction<'a> {
         }
     }
 
+    /// Whether the event should be added to the timeline as a new item.
+    fn should_add_event_item<P: RoomDataProvider>(
```
Hywan, line 234: This is nice!

Issue #4833, opened 2025-03-24
timeline: add thread-focused timeline


bnjbvr:
The parameter for the focus would be the thread's root (event id). Then, to start with, we can use the /relations endpoint to query the current events in the thread.

We'd need to make sure this includes bundled events for each thread event (e.g. reaction to a thread message).

stefanceriu:
All the subtask have now been implemented. Threaded read receipts will be handled in https://github.com/matrix-org/matrix-rust-sdk/issues/3123

PR #5032, opened 2025-05-14
feat(ui): add new `Thread` timeline focus mode and associated events loader

stefanceriu:
… that allows building a timeline instance specific to a particular thread root.

Creating a timeline in this mode will start by backpaginating root event relations with `num_events` and automatically insert the thread root event when reaching the end. It will include `RelationsOfType(RelationType::Thread)` but also recurse over the retrieved events to fetch reactions. 
It will not however react to new events received over sync or that the user sends (for now).

This patch will also help incrementally deliver the upstream client support for creating such a timeline.

Part of #4833 (meta #4869).

### DIFF
--- a/crates/matrix-sdk-ui/src/timeline/controller/mod.rs
+++ b/crates/matrix-sdk-ui/src/timeline/controller/mod.rs
@@ -181,7 +181,7 @@ impl Default for TimelineSettings {
     }
 }
 
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, Copy)]
 pub(super) enum TimelineFocusKind {
     Live,
     Event,
@@ -488,7 +488,7 @@ impl<P: RoomDataProvider, D: Decryptor> TimelineController<P, D> {
     ) -> Result<bool, PaginationError> {
         let PaginationResult { events, hit_end_of_timeline } = match &*self.focus.read().await {
             TimelineFocusData::Live | TimelineFocusData::PinnedEvents { .. } => {
-                return Err(PaginationError::NotEventFocusMode)
+                return Err(PaginationError::NotSupported)
             }
             TimelineFocusData::Event { paginator, .. } => paginator
                 .paginate_backward(num_events.into())
@@ -520,14 +520,14 @@ impl<P: RoomDataProvider, D: Decryptor> TimelineController<P, D> {
         num_events: u16,
     ) -> Result<bool, PaginationError> {
         let PaginationResult { events, hit_end_of_timeline } = match &*self.focus.read().await {
-            TimelineFocusData::Live | TimelineFocusData::PinnedEvents { .. } => {
-                return Err(PaginationError::NotEventFocusMode)
-            }
+            TimelineFocusData::Live
+            | TimelineFocusData::PinnedEvents { .. }
+            | TimelineFocusData::Thread { .. } => return Err(PaginationError::NotSupported),
+
             TimelineFocusData::Event { paginator, .. } => paginator
                 .paginate_forward(num_events.into())
                 .await
                 .map_err(PaginationError::Paginator)?,
-            TimelineFocusData::Thread { .. } => return Err(PaginationError::NotSupported),
         };
 
         // Events are in topological order.
--- a/crates/matrix-sdk-ui/src/timeline/controller/mod.rs
+++ b/crates/matrix-sdk-ui/src/timeline/controller/mod.rs
@@ -67,6 +67,7 @@ use super::{
     event_item::{ReactionStatus, RemoteEventOrigin},
     item::TimelineUniqueId,
     subscriber::TimelineSubscriber,
+    threaded_events_loader::ThreadedEventsLoader,
     traits::{Decryptor, RoomDataProvider},
     DateDividerMode, Error, EventSendState, EventTimelineItem, InReplyToDetails, PaginationError,
     Profile, RepliedToEvent, TimelineDetails, TimelineEventItemId, TimelineFocus, TimelineItem,
@@ -108,6 +109,13 @@ enum TimelineFocusData<P: RoomDataProvider> {
         paginator: Paginator<P>,
         /// Number of context events to request for the first request.
         num_context_events: u16,
+    },
+
+    Thread {
+        loader: ThreadedEventsLoader<P>,
+
+        /// Number of relations events to requests for the first request
+        num_events: u16,
     },
 
     PinnedEvents {
@@ -173,10 +181,11 @@ impl Default for TimelineSettings {
     }
 }
 
-#[derive(Debug, Clone, Copy)]
+#[derive(Debug, Clone)]
 pub(super) enum TimelineFocusKind {
     Live,
     Event,
+    Thread,
     PinnedEvents,
 }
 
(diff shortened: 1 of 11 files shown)


crates/matrix-sdk-ui/src/timeline/controller/state_transaction.rs
```diff
@@ -295,6 +295,9 @@ impl<'a> TimelineStateTransaction<'a> {
                                 }
                             }
                         }
+                        TimelineFocusKind::Thread => {
+                            should_add = true;
```
bnjbvr, line 299: (This is going to change in the PR I've made, which I'll land after yours, but: no need to do anything in this match arm, since `should_add` is true (line 264))

crates/matrix-sdk-ui/src/timeline/controller/mod.rs
```diff
@@ -173,10 +181,11 @@ impl Default for TimelineSettings {
     }
 }
 
-#[derive(Debug, Clone, Copy)]
+#[derive(Debug, Clone)]
```
bnjbvr, line 184: I think this can still be `Copy`, and then you can avoid adding `.clone()` everytime this is copied :)

crates/matrix-sdk-ui/src/timeline/controller/mod.rs
```diff
@@ -488,6 +527,7 @@ impl<P: RoomDataProvider, D: Decryptor> TimelineController<P, D> {
                 .paginate_forward(num_events.into())
                 .await
                 .map_err(PaginationError::Paginator)?,
+            TimelineFocusData::Thread { .. } => return Err(PaginationError::NotSupported),
```
bnjbvr, line 530: I think you can reuse the same match arm than the one above for `Live` and `PinnedEvents`; i like the error `NotSupported` more than the `NotEventFocusMode`, so feel free to use one or the other (and ditch the one you're not using anymore).

crates/matrix-sdk-ui/src/timeline/threaded_events_loader.rs
```diff
@@ -0,0 +1,127 @@
+// Copyright 2024 The Matrix.org Foundation C.I.C.
```
bnjbvr, line 1: correct dates? where we're going, we don't need dates!

```suggestion
// Copyright 2025 The Matrix.org Foundation C.I.C.
```

crates/matrix-sdk-ui/src/timeline/threaded_events_loader.rs
```diff
@@ -0,0 +1,127 @@
...
+            root_event_id,
+            tokens: Mutex::new(PaginationTokens { previous: None.into(), next: None.into() }),
+        }
+    }
+
+    pub async fn paginate_backwards(
+        &self,
+        num_events: UInt,
+    ) -> Result<PaginationResult, PaginatorError> {
+        self.paginate(Direction::Backward, num_events).await
```
bnjbvr, line 54: Since we're not likely to implement forward pagination, let's inline `paginate` in there, so then we remove the `Direction` parameter, and likely simplify the code in `paginate` too!

crates/matrix-sdk-ui/tests/integration/timeline/thread.rs
```diff
@@ -0,0 +1,308 @@
+// Copyright 2023 The Matrix.org Foundation C.I.C.
```
bnjbvr, line 1:
```suggestion
// Copyright 2025 The Matrix.org Foundation C.I.C.
```

crates/matrix-sdk-ui/tests/integration/timeline/thread.rs
```diff
@@ -0,0 +1,308 @@
...
+    assert_pending!(timeline_stream);
+
+    let hit_start = timeline.paginate_backwards(1).await.unwrap();
+    assert!(hit_start);
+
+    assert_let!(Some(timeline_updates) = timeline_stream.next().await);
+
+    // Remove date separator and insert a new one plus the remaining threaded
+    // even and the thread root
+    assert_eq!(timeline_updates.len(), 4);
```
bnjbvr, line 148: Sorry, but it would be nice to check that the updates are those we expect, here, please :) (We don't have to do it for every single test, though.)

crates/matrix-sdk-ui/tests/integration/timeline/thread.rs
```diff
@@ -0,0 +1,308 @@
...
+        .with_focus(TimelineFocus::Thread { root_event_id: thread_root_event_id, num_events: 1 })
+        .build()
+        .await
+        .unwrap();
+
+    timeline.send(RoomMessageEventContent::text_plain("Random message").into()).await.unwrap();
+
+    let (items, mut timeline_stream) = timeline.subscribe().await;
+
+    assert_eq!(items.len(), 2); // a date separator and the root
```
bnjbvr, line 306: Can you check on the items, please?

crates/matrix-sdk-ui/tests/integration/timeline/thread.rs
```diff
@@ -0,0 +1,308 @@
...
+        "Threaded event 3"
+    );
+    assert_eq!(
+        items[5].as_event().unwrap().content().as_message().unwrap().body(),
+        "Threaded event 4"
+    );
+}
+
+#[async_test]
+async fn test_thread_live_updates() {
```
bnjbvr, line 267: This test doesn't do much yet, right? Maybe we can get rid of it until it does :)

crates/matrix-sdk-ui/tests/integration/timeline/thread.rs
```diff
@@ -0,0 +1,308 @@
...
+    );
+
+    assert_eq!(
+        items[3].as_event().unwrap().content().as_message().unwrap().body(),
+        "Threaded event 2"
+    );
+}
+
+#[async_test]
+async fn test_thread_ordering() {
```
bnjbvr, line 169: Sorry, but what does this test do than the other isn't doing? Seems like it could be fused with the previous one. Or if it does something more, can you add a small comment at the top of the test's body, indicating what we're planning to test here, in a sentence or two?

bnjbvr (approved):
So neat <3

crates/matrix-sdk-ui/src/timeline/controller/mod.rs
stefanceriu: Ah sorry, this was left over from the other branch where I introduced an OwnedEventId as an associated value on the Thread here.

crates/matrix-sdk-ui/src/timeline/controller/state_transaction.rs
stefanceriu: Uh, good point. I rebased this on top of your PR and got rid of it.

crates/matrix-sdk-ui/src/timeline/threaded_events_loader.rs
stefanceriu: We will most likely have to implement forward pagination at some point for thread permalink support but I agree that it won't probably be here.

crates/matrix-sdk-ui/tests/integration/timeline/thread.rs
stefanceriu: Fair, I didn't know whether we want one that will fail later or the opposite tbh.

crates/matrix-sdk-ui/tests/integration/timeline/thread.rs
stefanceriu: Coming back to it yeah, they can definitely be merged together. The idea was for the order one to focus or just how messages are ordered in the thread, skipping a lot of the other checks.

PR #5034, opened 2025-05-14
feat(sdk): Send stable `m.marked_unread` room account data

zecakeh:


crates/matrix-sdk-base/src/rooms/mod.rs
```diff
@@ -587,6 +591,17 @@ impl RoomMemberships {
     }
 }
 
+/// The possible sources of an account data type.
+#[derive(Debug, Clone, Copy, Default, PartialEq, Eq, Serialize, Deserialize)]
+pub(crate) enum AccountDataSource {
+    /// The source is account data with the stable prefix.
+    Stable,
+
+    /// The source is account data with the unstable prefix.
```
Hywan, line 600: Would be nice to improve the doc by saying that the stable prefix is `m.` and the unstable prefix is `com.famedly`. I think it would clarify things greatly. Thoughts?

Hywan (changes requested):
Looks good to me. I'm requesting a tiny change, just to clarify the documentation of `AccountDataSource`. Other than that, I think we are good. Thanks for working on this!

crates/matrix-sdk-base/src/rooms/mod.rs
zecakeh: I made this type vague on purpose to be able to reuse it eventually in the future since I doubt this is going to be the only case of room account data that we will need to migrate. Although the stable prefix should always be `m.`, the unstable prefix will vary.

Hywan: Okay then.

Hywan (approved):

PR #5040, opened 2025-05-14
feat: implement and observe MSC4278 config value

Velin92:
This PR
- Updates Ruma to use the improved MediaPreviewConfig event type that also supports a `Default` for the content type
- Implemented a way to observe the stable and unstable values of the event and return the used one accordingly, if no one is present the default will be used
- Set the value (will only use unstable type for now)

### DIFF
--- a/bindings/matrix-sdk-ffi/src/ruma.rs
+++ b/bindings/matrix-sdk-ffi/src/ruma.rs
@@ -1111,9 +1111,12 @@ pub enum AccountDataEvent {
 /// The policy that decides if media previews should be shown in the timeline.
 #[derive(Clone, uniffi::Enum, Default)]
 pub enum MediaPreviews {
+    /// Always show media previews in the timeline.
     #[default]
     On,
+    /// Show media previews in the timeline only if the room is private.
     Private,
+    /// Never show media previews in the timeline.
     Off,
 }
 
@@ -1138,11 +1141,13 @@ impl From<MediaPreviews> for RumaMediaPreviews {
     }
 }
 
-/// The policy that decides if avarars should be shown in invite requests.
+/// The policy that decides if avatars should be shown in invite requests.
 #[derive(Clone, uniffi::Enum, Default)]
 pub enum InviteAvatars {
+    /// Always show avatars in invite requests.
     #[default]
     On,
+    /// Never show avatars in invite requests.
     Off,
 }
 
--- a/bindings/matrix-sdk-ffi/src/ruma.rs
+++ b/bindings/matrix-sdk-ffi/src/ruma.rs
@@ -30,6 +30,10 @@ use ruma::{
         ignored_user_list::{IgnoredUser as RumaIgnoredUser, IgnoredUserListEventContent},
         location::AssetType as RumaAssetType,
         marked_unread::{MarkedUnreadEventContent, UnstableMarkedUnreadEventContent},
+        media_preview_config::{
+            InviteAvatars as RumaInviteAvatars, MediaPreviewConfigEventContent,
+            MediaPreviews as RumaMediaPreviews,
+        },
         poll::start::PollKind as RumaPollKind,
         push_rules::PushRulesEventContent,
         room::{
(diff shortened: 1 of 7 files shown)


crates/matrix-sdk/src/client/mod.rs
```diff
@@ -3319,24 +3323,130 @@ pub(crate) mod tests {
...
+                "content": {
+                    "media_previews": "off",
+                    "invite_avatars": "on"
+                },
+                "type": "m.media_preview_config"
+                  })));
+            })
+            .await;
+
+        let next_value = stream.next().await.unwrap();
```
poljar, line 3364: There's `assert_next_eq!()` in the `stream_asserts` crate which combines the next and await call, making this less verbose.

crates/matrix-sdk/src/test_utils/mocks/mod.rs
```diff
@@ -1016,6 +1016,13 @@ impl MatrixMockServer {
         let mock = Mock::given(method("GET"));
         self.mock_endpoint(mock, RoomRelationsEndpoint::default()).expect_default_access_token()
     }
+
+    /// Create a prebuilt mock for the endpoint used to get the global account
+    /// data.
+    pub fn mock_global_account_data(&self) -> MockEndpoint<'_, GlobalAccountDataEndpoint> {
```
poljar, line 1022: I like to add examples to these, because it's not infrequent that people ask how these can be used.

Let's add one here as well.

crates/matrix-sdk/src/account.rs
```diff
@@ -980,6 +986,107 @@ impl Account {
...
+    ///
+    /// This will return the initial value of the configuration and a stream
+    /// that will yield new values as they are received.
+    ///
+    /// The initial value is the one that was stored in the account data
+    /// when the client was started. If no value was found in either the stable
+    /// and unstable type, the default configuration will be returned.
+    /// This is a temporary solution until we know which matrix version will
+    /// support the stable type.
+    pub async fn observe_media_preview_config(
```
poljar, line 1000: An example here would be nice. Streams tend to be a tricky part of Rust.

crates/matrix-sdk/src/account.rs
```diff
@@ -980,6 +986,107 @@ impl Account {
...
+    pub async fn set_invite_avatars_display_policy(&self, policy: InviteAvatars) -> Result<()> {
+        let mut media_preview_config = self.get_media_preview_config_event_content().await?;
+        media_preview_config.0.invite_avatars = policy;
+
+        // Updating the account data
+        self.set_account_data(media_preview_config).await?;
+        Ok(())
+    }
+
+    async fn get_media_preview_config_event_content(
```
poljar, line 1079: Let's document what and why this is returning a default value.

poljar (changes requested):
Looks mostly good, there are some missing docs here an there.

It would have been nice if this was spit out into separate commits. There are clear separation points. For example the extending of the mocking server could have been one commit, the functionality in the SDK a second one, and finally the bindings the third one.

But ok, it's not that big of a PR so it's still reviewable without this separation. Just something to keep in mind.

crates/matrix-sdk/src/client/mod.rs
Velin92: the problem is that the content type cannot be equated

bindings/matrix-sdk-ffi/src/ruma.rs
```diff
@@ -1104,6 +1108,63 @@ pub enum AccountDataEvent {
     },
 }
 
+/// The policy that decides if media previews should be shown in the timeline.
+#[derive(Clone, uniffi::Enum, Default)]
+pub enum MediaPreviews {
+    #[default]
```
poljar, line 1114: The enum variants aren't documented.

bindings/matrix-sdk-ffi/src/ruma.rs
```diff
@@ -1104,6 +1108,63 @@ pub enum AccountDataEvent {
...
+    fn from(value: MediaPreviews) -> Self {
+        match value {
+            MediaPreviews::On => Self::On,
+            MediaPreviews::Private => Self::Private,
+            MediaPreviews::Off => Self::Off,
+        }
+    }
+}
+
+/// The policy that decides if avarars should be shown in invite requests.
```
poljar, line 1141: Might be nice to elaborate which avatars, the room avatar, the inviter avatar, both?

```suggestion
/// The policy that decides if avatars should be shown in invite requests.
```

bindings/matrix-sdk-ffi/src/ruma.rs
```diff
@@ -1104,6 +1108,63 @@ pub enum AccountDataEvent {
...
+#[derive(Clone, uniffi::Enum, Default)]
+pub enum InviteAvatars {
+    #[default]
+    On,
```
poljar, line 1145: Docs on the variants missing.

crates/matrix-sdk/src/account.rs
```diff
@@ -980,6 +987,131 @@ impl Account {
...
+            .observe_events::<GlobalAccountDataEvent<MediaPreviewConfigEventContent>, ()>();
+
+        let stream = first_observer.subscribe().map(|event| event.0.content);
+
+        let second_observer = self
+            .client
+            .observe_events::<GlobalAccountDataEvent<UnstableMediaPreviewConfigEventContent>, ()>();
+
+        let second_stream = second_observer.subscribe().map(|event| event.0.content.0);
+
```
poljar, line 1046: I just realized, we should probably first create the streams, then fetch the initial value, as we'd otherwise will have a race where we'll miss an update that happens while we fetch the initial value.

So let's fix the order here.

poljar (changes requested):
There's still some docs missing and I just realized that there's going to be a small race.

Looks almost ready to be merged.

I know that I complained about force pushes, but now the CI is broken unless you rebase onto main.

So feel free to do that.

poljar (approved):
Thanks, looks good now.

PR #4989, opened 2025-04-29
WIP: receiving room key bundles

richvdh:
Work-in-progress branch for implementation of https://github.com/matrix-org/matrix-rust-sdk/issues/4513 and https://github.com/matrix-org/matrix-rust-sdk/issues/4514

Issue #4513, opened 2025-01-10
Encrypted history sharing: download and import keys from encrypted key blob


richvdh:
Our proposed solution to [sharing encrypted room history on invite](https://github.com/element-hq/element-meta/issues/39) involves sharing the keys to an encrypted blob of room keys via to-device messages.

https://github.com/matrix-org/matrix-rust-sdk/issues/4505 tracks uploading that encrypted blob. This issue tracks downloading, decrypting, and importing the keys.

PR #5041, opened 2025-05-14
refactor(crypto): Make the room key importing logic more generic

poljar:
Part of https://github.com/matrix-org/matrix-rust-sdk/issues/4513 and necessary for https://github.com/matrix-org/matrix-rust-sdk/pull/4989.

Making this generic right now is a bit of a no-op, but eventually we'll also import the `HistoricRoomKey` type which requires this to be generic.

### DIFF
--- a/crates/matrix-sdk-crypto/src/store/mod.rs
+++ b/crates/matrix-sdk-crypto/src/store/mod.rs
@@ -74,7 +74,8 @@ use crate::{
     },
     types::{
         events::room_key_withheld::RoomKeyWithheldEvent, BackupSecrets, CrossSigningSecrets,
-        EventEncryptionAlgorithm, MegolmBackupV1Curve25519AesSha2Secrets, SecretsBundle,
+        EventEncryptionAlgorithm, MegolmBackupV1Curve25519AesSha2Secrets, RoomKeyExport,
+        SecretsBundle,
     },
     verification::VerificationMachine,
     CrossSigningStatus, OwnUserIdentityData, RoomKeyImportResult,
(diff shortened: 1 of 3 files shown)


crates/matrix-sdk-crypto/src/store/mod.rs
```diff
@@ -1832,6 +1833,54 @@ impl Store {
...
+        &self,
+        exported_keys: Vec<ExportedRoomKey>,
+        progress_listener: impl Fn(usize, usize),
+    ) -> Result<RoomKeyImportResult> {
+        self.import_room_keys(exported_keys, None, progress_listener).await
+    }
+
+    async fn import_sessions_impl<T>(
+        &self,
+        room_keys: Vec<T>,
```
andybalaam, line 1876: Can this be `IntoIter<T>` or something, to avoid the `collect()` above when we call it?

andybalaam (approved):
Looks good, one question.

crates/matrix-sdk-crypto/src/store/mod.rs
poljar: Kinda, though it can't be just that, because we have a `room_keys.len()` call.

Which then makes all of the trait bounds even more complex.

The other option is to remove the `len()` call, but then our progress listener doesn't work anymore.

Don't think it's worth it to avoid this `collect()` call, after all, allocating a vector of references isn't that big of a deal.

andybalaam: Yep, agree it's fine as-is.

andybalaam: (Or pass in the length?)

poljar: > (Or pass in the length?)

That sounds like it would be a bit of a clunky API, not terrible but clunky enough for me to rather take the allocation.

The only reason why we take a `Vec<&Foo>` vs a `Vec<Foo>` is so we can gather info about the specific key in the error branch if creating a `InboundGroupSession` fails. 

Perhaps gathering the info before the `try_into()` call consumes the key would be the wiser choice if we really want to avoid the allocation.

poljar: Alright, let's keep the allocation, if we change our mind I think gathering the info before the consumption of the key would be the way to go.

PR #5043, opened 2025-05-15
fix(sdk): mark invited DM rooms joined with `Client::join_room_by_id` or `Client::join_room_by_id_or_alias` as DMs

### LINKED ISSUE #4730: Starting a DM with the same user twice results in 2 entries in `m.direct`.
As [reported](https://matrix.to/#/!amwlFMejXQcoCxNYHH:matrix.org/$QRFjKl9WrXMH6w54XI4l_dBocNp-hjKRAWGhjbDVuko?via=matrix.org&via=element.io&via=beeper.com) in the Element X public room, it seems that if you create a DM with someone after having left a DM with them (we don't allow creating 2 DMs with the same user), the `m.direct` ends up with 2 entires for that user ID. There is then a second issue at play, where if there are 2 entries for the same user ID, the SDK doesn't recognise the new DM as a `direct` (I haven't dug into what gets decoded).

Reported steps to reproduce:

	1.	Both sides cleaned `m.direct`
	2.	A starts DM with B
	3.	Chat is shown as DM by EX of A and as Room in EX of B
	4.	B opens EW in Browser where it is shown as DM and clicks on DM with A
	5.	Chat is shown as DM in EX of B
	6.	A and B leave room
	7.	A starts DM with B
	8.	Chat is shown as DM by EX of A and as Room in EX of B
	9.	B opens EW in Browser where it is shown as DM and clicks on DM with A
	10.   Chat is NOT shown as DM in EX of B

I'm not sure what is going on with `4`/`5` but I think the important thing here is that by `10` you have two entires in the `m.direct`.

We're creating DMs with the following:

```
let parameters = CreateRoomParameters(name: nil,
                                      topic: nil,
                                      isEncrypted: true,
                                      isDirect: true,
                                      visibility: .private,
                                      preset: .trustedPrivateChat,
                                      invite: [userID],
                                      avatar: nil,
                                      powerLevelContentOverride: .init(usersDefault: nil,
                                                                       eventsDefault: nil,
                                                                       stateDefault: nil,
                                                                       ban: nil,
                                                                       kick: nil,
                                                                       redact: nil,
                                                                       invite: nil,
                                                                       notifications: nil,
                                                                       users: [:],
                                                                       events: [
                                                                        "m.call.member": Int32(0),
                                                                        "org.matrix.msc3401.call.member": Int32(0)
                                                                       ]))
let roomID = try await client.createRoom(request: parameters)
```

bnjbvr:
This moves the logic from `Room::join()` to these two methods. This is isofunctional, because `Room::join()` does call into `Client::join_room_by_id()` internally.

Fix #4730.

### DIFF
--- a/crates/matrix-sdk/src/client/mod.rs
+++ b/crates/matrix-sdk/src/client/mod.rs
@@ -1411,10 +1411,34 @@ impl Client {
         }
     }
 
+    /// Finish joining a room.
+    ///
+    /// If the room was an invite that should be marked as a DM, will include it
+    /// in the DM event after creating the joined room.
+    async fn finish_join_room(&self, room_id: &RoomId) -> Result<Room> {
+        let mark_as_dm = if let Some(room) = self.get_room(room_id) {
+            room.state() == RoomState::Invited
+                && room.is_direct().await.unwrap_or_else(|e| {
+                    warn!(%room_id, "is_direct() failed: {e}");
+                    false
+                })
+        } else {
+            false
+        };
+
+        let base_room = self.base_client().room_joined(room_id).await?;
+        let room = Room::new(self.clone(), base_room);
+
+        if mark_as_dm {
+            room.set_is_direct(true).await?;
+        }
+
+        Ok(room)
+    }
+
     /// Join a room by `RoomId`.
     ///
-    /// Returns a `join_room_by_id::Response` consisting of the
-    /// joined rooms `RoomId`.
+    /// Returns the `Room` in the joined state.
     ///
     /// # Arguments
     ///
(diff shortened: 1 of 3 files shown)


pixlwave (approved):
Thank you, testing this on EXI it works well 👏

crates/matrix-sdk/src/client/mod.rs
```diff
@@ -1411,30 +1411,51 @@ impl Client {
...
+
+        let mark_as_dm = self.must_mark_joined_room_as_dm(&response.room_id).await;
         let base_room = self.base_client().room_joined(&response.room_id).await?;
-        Ok(Room::new(self.clone(), base_room))
+        let room = Room::new(self.clone(), base_room);
+        if mark_as_dm {
+            room.set_is_direct(true).await?;
+        }
+
+        Ok(room)
```
poljar, line 1446: Can't we put this code snippet into a function? Seems like we have the same thing in two places now, unless I'm missing something.

bnjbvr: You're right, I've put all of this inside `must_mark_joined_room_as_dm` and renamed it `finish_join_room` :+1:

crates/matrix-sdk/src/client/mod.rs
```diff
@@ -1411,30 +1411,54 @@ impl Client {
         }
     }
 
+    /// Finish joining a room.
+    ///
+    /// If the room was an invite that should be marked as a DM, will include it
+    /// in the DM event after creating the joined room.
+    async fn finish_join_room(&self, room_id: &RoomId) -> Result<Room> {
+        let mark_as_dm = if let Some(room) = self.get_room(room_id) {
+            room.state() == RoomState::Invited
```
richvdh, line 1420: won't this race with the `/sync` response, causing it to do the wrong thing if the `/sync` response arrives before the `/join` completes?

bnjbvr: Good catch :+1: I suppose passing the previous room state, in the callers, captured before sending the `join_room` requests, would be sufficient to avoid the race (but would also need some kind of regression test to prove it can happen in the first place).

PR #5044, opened 2025-05-15
crypto: rewrite `check_sender_trust_requirement` to use `VerificationState`

richvdh:
There are two reasons for this.

Firstly. we've already done a bunch of work to map `SenderData` into a `VerificationState`, and the decision tree from `VerificationState` to allow/reject is simpler than going from `SenderData`, even if we have to fudge it a bit to get the "legacy" flag. (Note that it allows us to get rid of an `unreachable!` panic.)

Secondly, `VerficationState` represents the state of an *event*, whereas `SenderData` is about the session as a whole. A session can be fine, whilst events (claiming to be) encrypted with it can be suspect. What we want here is to check a specific message. Currently, this doesn't make any functional difference, but conceptually it's cleaner to check the `VerificationState`.

Note that there are a bunch of tests for this method in `matrix-sdk-crypto/src/machine/tests/decryption_verification_state.rs`, called `test_decryption_trust_requirement`.

### DIFF
--- a/crates/matrix-sdk-crypto/src/machine/mod.rs
+++ b/crates/matrix-sdk-crypto/src/machine/mod.rs
@@ -53,7 +53,7 @@ use tokio::sync::Mutex;
 use tracing::{
     debug, error,
     field::{debug, display},
-    info, instrument, warn, Span,
+    info, instrument, trace, warn, Span,
 };
 use vodozemac::{
     megolm::{DecryptionError, SessionOrdering},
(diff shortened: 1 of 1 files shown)


crates/matrix-sdk-crypto/src/machine/mod.rs
```diff
@@ -1786,51 +1786,60 @@ impl OlmMachine {
...
+                let legacy_session = match &session.sender_data {
+                    SenderData::DeviceInfo { legacy_session, .. } => legacy_session,
+                    SenderData::UnknownDevice { legacy_session, .. } => legacy_session,
+                    _ => &false,
+                };
+
+                match (verification_level, legacy_session) {
+                    (VerificationLevel::UnverifiedIdentity, _) => Ok(()),
+                    (VerificationLevel::UnsignedDevice, true) => Ok(()),
+                    (VerificationLevel::None(_), true) => Ok(()),
```
richvdh, line 1835: correspondances between `VerificationLevel` and `SenderData`:

`SenderData::VerificationViolation` -> `VerificationLevel::VerificationViolation`
`SenderData::SenderUnverified` -> `VerificationLevel::UnverifiedIdentity`
`SenderData::DeviceInfo` ->  `VerificationLevel::UnsignedDevice`
`SenderData::UnknownDevice` -> `VerificationLevel::None`

There is no equivalent to `SenderData::SenderVerified`; that is handled as `VerificationState::Verified` above.

crates/matrix-sdk-crypto/src/machine/mod.rs
```diff
@@ -1786,52 +1786,62 @@ impl OlmMachine {
...
+                    SenderData::DeviceInfo { legacy_session, .. } => legacy_session,
+                    SenderData::UnknownDevice { legacy_session, .. } => legacy_session,
+                    _ => &false,
+                };
+
+                match (verification_level, legacy_session) {
+                    (VerificationLevel::UnverifiedIdentity, _) => true,
+                    (VerificationLevel::UnsignedDevice, true) => true,
+                    (VerificationLevel::None(_), true) => true,
+                    _ => false,
```
richvdh, line 1831: correspondances between `VerificationLevel` and `SenderData`:

`SenderData::VerificationViolation` -> `VerificationLevel::VerificationViolation`
`SenderData::SenderUnverified` -> `VerificationLevel::UnverifiedIdentity`
`SenderData::DeviceInfo` ->  `VerificationLevel::UnsignedDevice`
`SenderData::UnknownDevice` -> `VerificationLevel::None`

There is no equivalent to `SenderData::SenderVerified`; that is handled as `VerificationState::Verified` above.

crates/matrix-sdk-crypto/src/machine/mod.rs
```diff
@@ -1786,52 +1786,61 @@ impl OlmMachine {
...
+                    SenderData::UnknownDevice { legacy_session, .. } => legacy_session,
+                    _ => &false,
+                };
+
+                matches!(
+                    (verification_level, legacy_session),
+                    (VerificationLevel::UnverifiedIdentity, _)
+                        | (VerificationLevel::UnsignedDevice, true)
+                        | (VerificationLevel::None(_), true)
+                )
```
richvdh, line 1832: correspondances between `VerificationLevel` and `SenderData`:

`SenderData::VerificationViolation` -> `VerificationLevel::VerificationViolation`
`SenderData::SenderUnverified` -> `VerificationLevel::UnverifiedIdentity`
`SenderData::DeviceInfo` ->  `VerificationLevel::UnsignedDevice`
`SenderData::UnknownDevice` -> `VerificationLevel::None`

There is no equivalent to `SenderData::SenderVerified`; that is handled as `VerificationState::Verified` above.

poljar, line 1832: I don't like using `matches!()` in crypto related code, it's very easy to forget to modify the statement once new variants are added to the enum. I know it's concise and looks neat, it still makes me worry for the future.

What do you think about listing it all out like so:

```rust
// In the CrossSignedOrLegacy case the following rules apply:
//
// 1. Identities we have not yet verified can be decrypted regardless of the
//    legacy state of the session.
// 2. Devices that aren't signed by the owning identity of the device can only
//    be decrypted if it's a legacy session.
// 3. If we have no information about the device, we should only decrypt if it's
//    a legacy session.
// 4. Anything else, should throw an error.
match (verification_level, legacy_session) {
    // Case 1
    (VerificationLevel::UnverifiedIdentity, _) => true,

    // Case 2
    (VerificationLevel::UnsignedDevice, true) => true,

    // Case 3
    (VerificationLevel::None(_), true) => true,

    // Case 4
    (VerificationLevel::VerificationViolation, _)
    | (VerificationLevel::UnsignedDevice, _)
    | (VerificationLevel::None(_), false) => false,
}
```

poljar (approved):
I left some small suggestions, I think the logic is fine.

crates/matrix-sdk-crypto/src/machine/mod.rs
richvdh: I agree it's clearer. I started with that, but I got a lint failure. I'll disable the lint, I guess.

poljar:
The CI failures are unrelated, they are being fixed by @Hywan.

PR #5045, opened 2025-05-15
fix(ui): populate the `thread_root` and `in_reply_to` fields for stickers and polls

stefanceriu:
They have never been set and there was no way of telling if stickers and polls belong on a thread or come in reply of any other message.

This patch also exposes methods for setting these relations on the event factory level for polls.

### DIFF
--- a/testing/matrix-sdk-test/src/event_factory.rs
+++ b/testing/matrix-sdk-test/src/event_factory.rs
@@ -45,7 +45,8 @@ use ruma::{
             member::{MembershipState, RoomMemberEventContent},
             message::{
                 FormattedBody, ImageMessageEventContent, MessageType, Relation,
-                RoomMessageEventContent, RoomMessageEventContentWithoutRelation,
+                RelationWithoutReplacement, RoomMessageEventContent,
+                RoomMessageEventContentWithoutRelation,
             },
             name::RoomNameEventContent,
             power_levels::RoomPowerLevelsEventContent,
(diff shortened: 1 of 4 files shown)


testing/matrix-sdk-test/src/event_factory.rs
```diff
@@ -360,6 +361,32 @@ impl EventBuilder<RoomMessageEventContent> {
...
 
+impl EventBuilder<UnstablePollStartEventContent> {
+    /// Adds a reply relation to the current event.
+    pub fn reply_to(mut self, event_id: &EventId) -> Self {
+        if let UnstablePollStartEventContent::New(content) = &mut self.content {
+            content.relates_to = Some(RelationWithoutReplacement::Reply {
+                in_reply_to: InReplyTo::new(event_id.to_owned()),
+            });
+
+            self.content = UnstablePollStartEventContent::New(content.to_owned());
```
bnjbvr, line 372: Is this required? We should've modified the `content` in place above, because of the `&mut self.content`…

testing/matrix-sdk-test/src/event_factory.rs
```diff
@@ -360,6 +361,32 @@ impl EventBuilder<RoomMessageEventContent> {
...
+    }
+
+    /// Adds a thread relation to the root event, setting the reply to
+    /// event id as well.
+    pub fn in_thread(mut self, root: &EventId, reply_to_event_id: &EventId) -> Self {
+        let thread = Thread::reply(root.to_owned(), reply_to_event_id.to_owned());
+
+        if let UnstablePollStartEventContent::New(content) = &mut self.content {
+            content.relates_to = Some(RelationWithoutReplacement::Thread(thread));
+            self.content = UnstablePollStartEventContent::New(content.to_owned());
```
bnjbvr, line 384: ditto for this line

bnjbvr (approved):
Thanks, lgtm!

testing/matrix-sdk-test/src/event_factory.rs
stefanceriu: Nop, that's just me botching the language 🙈

Issue #4823, opened 2025-03-20
timeline: finish the aggregations refactoring


bnjbvr:
https://github.com/matrix-org/matrix-rust-sdk/pull/4576 introduced a manager for aggregations (edits, reactions, poll responses etc.), that will record them over time, and applying them when the aggregation is observed before the event it applies to (which could happen with SSS and backpaginations, notably).

The work there hasn't been finished, and it's promising in that it should allow for a few refactorings reducing the complexity of adding new items to the timeline:

- [ ] add unit tests for the aggregation manager
- [x] take care of redactions in the aggregation manager (with test)
- [x] take care of edits in the aggregation manager (with test).
  - This one might be fun, because of bundled edits, and we should only recall the *latest* edit, whatever that means — likely comparing the previous latest edit with a new edit's event position in the `all_events` array.
  - [ ] I think an event-focused/pinned-events timeline should always prioritize the bundled edit over other seen edits, as it's stateless. A live timeline should always prioritize the latest edit, and resort to the bundled edit otherwise.
- [x] Finish https://github.com/matrix-org/matrix-rust-sdk/pull/4655
- [x] Make use of the above when handling a reply preview, instead of manually re-doing the work.
  - https://github.com/matrix-org/matrix-rust-sdk/pull/5057
- [x] With all of the above, we can selectively choose to only handle live aggregations, and not all live events in general. Resolve https://github.com/matrix-org/matrix-rust-sdk/issues/4643
  - https://github.com/matrix-org/matrix-rust-sdk/pull/5060

PR #5046, opened 2025-05-15
refactor(timeline): finish the aggregations refactoring

bnjbvr:
This uses the `Aggregations` manager to take care of redactions and edits, allowing us to get rid of the `pending_edits` in the `TimelineMetadata` \o/ Aggregations are now handled in a single manner, which is nice for consistency.

For redactions, this means we now stash a redaction even if we've seen it before the event it redacts, making sure we won't display something that should've been redacted.

For edits, we now do "the right thing", and will look at the position of each edit event we know about, and use the most "recent" one, if we've found any. This will help handling edits within threads as well (as such, helps with #4869), instead of having us implement some more complicated logic for those.

Part of #4823.

### DIFF
--- a/crates/matrix-sdk-ui/src/timeline/controller/aggregations.rs
+++ b/crates/matrix-sdk-ui/src/timeline/controller/aggregations.rs
@@ -353,7 +353,7 @@ impl Aggregations {
     /// Will return an error at the first aggregation that couldn't be applied;
     /// see [`Aggregation::apply`] which explains under which conditions it can
     /// happen.
-    pub fn apply(
+    pub fn apply_all(
         &self,
         item_id: &TimelineEventItemId,
         event: &mut EventTimelineItem,
--- a/crates/matrix-sdk-ui/src/timeline/controller/aggregations.rs
+++ b/crates/matrix-sdk-ui/src/timeline/controller/aggregations.rs
@@ -41,6 +41,12 @@ use std::{borrow::Cow, collections::HashMap};
 
 use as_variant::as_variant;
 use ruma::{
+    events::{
+        poll::unstable_start::NewUnstablePollStartEventContentWithoutRelation,
+        relation::Replacement, room::message::RoomMessageEventContentWithoutRelation,
+        AnySyncTimelineEvent,
+    },
+    serde::Raw,
     MilliSecondsSinceUnixEpoch, OwnedEventId, OwnedTransactionId, OwnedUserId, RoomVersionId,
 };
 use tracing::{info, trace, warn};
@@ -50,6 +56,30 @@ use crate::timeline::{
     EventTimelineItem, MsgLikeContent, MsgLikeKind, PollState, ReactionInfo, ReactionStatus,
     TimelineEventItemId, TimelineItem, TimelineItemContent,
 };
+
+#[derive(Clone)]
+pub(in crate::timeline) enum PendingEditKind {
+    RoomMessage(Replacement<RoomMessageEventContentWithoutRelation>),
+    Poll(Replacement<NewUnstablePollStartEventContentWithoutRelation>),
+}
+
+impl std::fmt::Debug for PendingEditKind {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            Self::RoomMessage(_) => f.debug_struct("RoomMessage").finish_non_exhaustive(),
+            Self::Poll(_) => f.debug_struct("Poll").finish_non_exhaustive(),
+        }
+    }
+}
+
+#[derive(Clone, Debug)]
+pub(in crate::timeline) struct PendingEdit {
+    /// The kind of edit this is.
+    pub kind: PendingEditKind,
+
+    /// The raw JSON for the edit.
+    pub edit_json: Option<Raw<AnySyncTimelineEvent>>,
+}
 
 /// Which kind of aggregation (related event) is this?
 #[derive(Clone, Debug)]
@@ -88,6 +118,8 @@ pub(crate) enum AggregationKind {
 
     /// An event has been redacted.
     Redaction,
+
+    Edit(PendingEdit),
 }
 
 /// An aggregation is an event related to another event (for instance a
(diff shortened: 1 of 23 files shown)


bnjbvr:
Now rebased and ready for review!

crates/matrix-sdk-ui/tests/integration/timeline/edit.rs
```diff
@@ -492,7 +492,14 @@ async fn test_edit_to_replied_updates_reply() {
...
 
-    // The reply events are updated with the edited replied-to content.
+    // The edit happens.
+    assert_next_matches!(timeline_stream, VectorDiff::Set { index: 0, value } => {
+        let msg = value.content().as_message().unwrap();
+        assert_eq!(msg.body(), "hello world");
+        assert!(msg.is_edited());
+    });
+
+    // And then the reply events are updated with the edited replied-to content.
```
Hywan, line 502: It's cool to have these effects in that order. It seems a bit more as expected.

crates/matrix-sdk-ui/src/timeline/event_handler.rs
```diff
@@ -491,6 +493,8 @@ impl TimelineAction {
                                         new_content,
                                     )),
                                     edit_json,
+                                    // TODO maybe we could provide it?
```
Hywan, line 437: Has it been addressed?

crates/matrix-sdk-ui/src/timeline/event_handler.rs
```diff
@@ -416,6 +416,8 @@ impl TimelineAction {
...
+                                    encryption_info: None,
```
Hywan, line 391: Has it been addressed?

crates/matrix-sdk-ui/src/timeline/controller/metadata.rs
```diff
@@ -79,9 +77,6 @@ pub(in crate::timeline) struct TimelineMetadata {
     /// Given an event, what are all the events that are replies to it?
     pub replies: HashMap<OwnedEventId, BTreeSet<OwnedEventId>>,
 
-    /// Edit events received before the related event they're editing.
-    pub pending_edits: RingBuffer<PendingEdit>,
```
Hywan, line 83: 🎉

crates/matrix-sdk-ui/src/timeline/controller/observable_items.rs
```diff
@@ -1947,7 +1947,16 @@ impl AllRemoteEvents {
...
+        // Reverse the iterator to start looking at the end. Since this will give us the
+        // "reverse" position, reverse the index after finding the event.
+        //
+        // If the reversed index was 0 in a array of size 1, then it should stay 0, so
+        // that justifies the `- 1` in the formula below.
+        self.0
+            .iter()
+            .rev()
+            .position(|event_meta| event_meta.event_id == event_id)
+            .map(|rev_index| self.0.len() - 1 - rev_index)
```
Hywan, line 1959: Could be simplified as:

```rust
self.0
    .iter()
    .enumerate()
    .rev()
    .find_map(|(index, event_meta)| { (event_meta.event_id == event_id).then_some(index))
```

So that you don't need to do the math yourself.

It works because [`Enumerate` implements `DoubleEndedIterator`](https://doc.rust-lang.org/std/iter/struct.Enumerate.html#impl-DoubleEndedIterator-for-Enumerate%3CI%3E).

crates/matrix-sdk-ui/src/timeline/controller/metadata.rs
```diff
@@ -75,6 +75,8 @@ pub(in crate::timeline) struct TimelineMetadata {
     pub aggregations: Aggregations,
 
     /// Given an event, what are all the events that are replies to it?
+    ///
+    /// Only works for remote events *and* replies which are remote-echoed.
```
Hywan, line 79: Should we rename `replies` to `remote_replies`? Not convinced, just asking.

Hywan (approved):
It's a big PR with many changes. I've spotted a couple of optimisations. Approving ahead of time to avoid another round a review. The main purpose of this patch is reached I think. Good job!

crates/matrix-sdk-ui/src/timeline/controller/metadata.rs
bnjbvr: It's unclear in `remote_replies` which side is `remote`; in this case, it's both the original and the reply which must have been remote-echoed, and I think this information is more clearly described from the code comments + types used (only `OwnedEventId`, not `TimelineEventItemId` anywhere). If you don't mind too much, I'll keep this name.

crates/matrix-sdk-ui/src/timeline/controller/observable_items.rs
bnjbvr: Good idea, thanks!

crates/matrix-sdk-ui/src/timeline/controller/aggregations.rs
```diff
@@ -156,31 +161,29 @@ impl Aggregation {
...
+            },
 
             AggregationKind::Reaction { key, sender, timestamp, reaction_status } => {
-                let Some(reactions) = event.content_mut().reactions_mut() else {
+                let Some(reactions) = event.content().reactions() else {
                     // These items don't hold reactions.
                     return ApplyAggregationResult::LeftItemIntact;
                 };
 
+                let mut reactions = reactions.clone();
```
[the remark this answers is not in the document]
bnjbvr: It's possible to do this, but there's one change in the code excerpt you posted that would be needed:

```
let Some(mut reactions) = event.to_mut().content_mut().reactions_mut() else { … }
```

This eagerly marks the `Cow` as written to, so eagerly clones the underlying content if it was borrowed. In that case, we might be cloning even though the final item isn't actually mutated. So we'd need to do another dance:

- on the read-only version of reaction, check first if there's a similar reaction
- if not, do the mutable transition and add it.

I've done that locally; it's a bit more code, but it's OK and should avoid a few useless clones in some edge cases.

crates/matrix-sdk-ui/src/timeline/event_handler.rs
bnjbvr: It can be done, I'll open a followup PR to keep changes tidy.

PR #5048, opened 2025-05-16
refactor: Reduce the size of different types

Hywan:
Since the migration to Rust 1.87, Clippy is emitting warnings: [`enum_large_variant`](https://rust-lang.github.io/rust-clippy/master/index.html#large_enum_variant). The tl;dr is:

> Enum size is bounded by the largest variant. Having one large variant can penalize the memory layout of that enum.

In most cases, this is relevant. In a few places, we consider this is a false-positive (esp. in `matrix-sdk-ffi` where it could hurt performances). Please, review this pull request patch-by-patch for more clarity.

### DIFF
--- a/crates/matrix-sdk-crypto/src/types/requests/room_message.rs
+++ b/crates/matrix-sdk-crypto/src/types/requests/room_message.rs
@@ -28,5 +28,5 @@ pub struct RoomMessageRequest {
     pub txn_id: OwnedTransactionId,
 
     /// The event content to send.
-    pub content: AnyMessageLikeEventContent,
+    pub content: Box<AnyMessageLikeEventContent>,
 }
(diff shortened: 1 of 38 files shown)


crates/matrix-sdk-crypto/src/types/requests/room_message.rs
```diff
@@ -28,5 +28,5 @@ pub struct RoomMessageRequest {
     pub txn_id: OwnedTransactionId,
 
     /// The event content to send.
-    pub content: AnyMessageLikeEventContent,
+    pub content: Box<AnyMessageLikeEventContent>,
```
richvdh, line 31: @Hywan it would be nice to document breaking changes like this in the changelog, where possible!

richvdh: Bleh, it was actually the change to `AnyOutgoingRequest::RoomMessage` which broke my build. Still, the point stands.

Hywan: Sure, will do next time.

PR #5051, opened 2025-05-16
Test: Replace `sync_timeline_event!` with `EventFactory` for beacon events in room integration test

yousefmoazzam:
Part of #3716

I _think_ I've done something sensible for adding the ability to put age into the unsigned data of an event, but it's possible I've misunderstood how the fields on `Unsigned` are being used.

It seemed like fields from different `*Unsigned` types from `ruma` were all put onto the single `Unsigned` struct (ie, `transaction_id` and `relations` from `RoomMemberUnsigned`, and `redacted_because` from `RedactedUnsigned`), and that some but not all fields on an `Unsigned` instance within an `EventBuilder` would be made a `Some` value depending on what event type was being created.

So, I just stuck another field onto `Unsigned` to allow `age` to be in the unsigned data, even though it's not a field that's available on every `*Unsigned` type, imagining it'd be used only for events which actually could have `age` in the unsigned data. Hope that makes sense!

And apologies for the slightly gnarly `<&EventId>::try_from` bit. I would have just used the `event_id!` macro, but rust doesn't allow doing `event_id!(format!(...))` (ie, trying to run one macro and then pass its output as input to another macro).

### DIFF
--- a/crates/matrix-sdk/tests/integration/room/beacon/mod.rs
+++ b/crates/matrix-sdk/tests/integration/room/beacon/mod.rs
@@ -234,13 +234,14 @@ async fn test_most_recent_event_in_stream() {
 
     let f = EventFactory::new();
     for nth in 0..25 {
+        let event_id = format!("$event_for_stream_{nth}");
         timeline_events.push(
             f.beacon(
                 owned_event_id!("$15139375514XsgmR:localhost"),
                 format!("geo:{nth}.9575274619722,12.494122581370175;u={nth}"),
                 Some(MilliSecondsSinceUnixEpoch(1_636_829_458u32.into())),
             )
-            .event_id(<&EventId>::try_from(format!("$event_for_stream_{nth}").as_str()).unwrap())
+            .event_id(<&EventId>::try_from(event_id.as_str()).unwrap())
             .server_ts(1_636_829_458)
             .sender(user_id!("@example2:localhost"))
             .age(598971)
--- a/crates/matrix-sdk/tests/integration/room/beacon/mod.rs
+++ b/crates/matrix-sdk/tests/integration/room/beacon/mod.rs
@@ -7,8 +7,8 @@ use matrix_sdk::{
     test_utils::mocks::MatrixMockServer,
 };
 use matrix_sdk_test::{
-    async_test, event_factory::EventFactory, mocks::mock_encryption_state, sync_timeline_event,
-    test_json, JoinedRoomBuilder, SyncResponseBuilder, DEFAULT_TEST_ROOM_ID,
+    async_test, event_factory::EventFactory, mocks::mock_encryption_state, test_json,
+    JoinedRoomBuilder, SyncResponseBuilder, DEFAULT_TEST_ROOM_ID,
 };
 use ruma::{
     event_id,
--- a/crates/matrix-sdk/tests/integration/room/beacon/mod.rs
+++ b/crates/matrix-sdk/tests/integration/room/beacon/mod.rs
@@ -17,7 +17,7 @@ use ruma::{
     },
     owned_event_id, room_id,
     time::SystemTime,
-    user_id, MilliSecondsSinceUnixEpoch,
+    user_id, EventId, MilliSecondsSinceUnixEpoch,
 };
 use serde_json::json;
 use wiremock::{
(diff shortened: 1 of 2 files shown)


crates/matrix-sdk/tests/integration/room/beacon/mod.rs
```diff
@@ -232,26 +232,20 @@ async fn test_most_recent_event_in_stream() {
...
-            "sender": "@example2:localhost",
-            "type": "org.matrix.msc3672.beacon",
-            "unsigned": {
-                "age": 598971
-            }
-        }));
+        timeline_events.push(
+            f.beacon(
+                owned_event_id!("$15139375514XsgmR:localhost"),
+                format!("geo:{nth}.9575274619722,12.494122581370175;u={nth}"),
```
poljar, line 240: This is a bit too stringly typed for my taste.

What do you think about passing in the longitude, latitude and uncertainty as arguments and the string gets built in the `beacon()` function?

crates/matrix-sdk/tests/integration/room/beacon/mod.rs
```diff
@@ -232,26 +232,20 @@ async fn test_most_recent_event_in_stream() {
...
+                Some(MilliSecondsSinceUnixEpoch(1_636_829_458u32.into())),
+            )
+            .event_id(<&EventId>::try_from(format!("$event_for_stream_{nth}").as_str()).unwrap())
```
poljar, line 243: Putting the string or the event ID into a variable at the top might be a bit less gnarly. 

```rust
let event_id: &EventId = <&EventId>::try_from(format!(...));
```

testing/matrix-sdk-test/src/event_factory.rs
```diff
@@ -790,6 +806,16 @@ impl EventFactory {
         self.event(event)
     }
 
+    /// Create a new `org.matrix.msc3672.beacon` event.
```
poljar, line 809: An example for this method would be nice, though not required.

poljar (changes requested):
Looks mostly good, I left some minor suggestions if you're up for it.

crates/matrix-sdk/tests/integration/room/beacon/mod.rs
yousefmoazzam: Yeah it's a good point, I agree that the geo URI could be handled more robustly rather than just passing a string, I'll do that!

crates/matrix-sdk/tests/integration/room/beacon/mod.rs
yousefmoazzam: This is a fair suggestion, I think inlining the entire thing was a bit much, I'll split it up

testing/matrix-sdk-test/src/event_factory.rs
yousefmoazzam: Sure thing, I'm happy to make the docstring more useful by adding an example, thanks for the suggestion!

crates/matrix-sdk/tests/integration/room/beacon/mod.rs
yousefmoazzam: Done in fixup commit eb9ae84

crates/matrix-sdk/tests/integration/room/beacon/mod.rs
yousefmoazzam: Done in fixup commit dbf6516 + corresponding changes to tests done in fixup commit 4c3957c.

I modified the asserts because the printing of the float latitude and longitude values was "slightly off", most likely due to floating-point representation oddities. An example I saw from one of the tests would probably explain it more succinctly; this code:
```rust
println!("value is {}", 12.494122581370175);
```

prints this:
```
value is 12.494122581370174
```

Similarly for `10.000000` getting printed as `10` for the other test.

So, I opted to use the `format!` macro to produced the expected value/LHS of the assertions.

The nicest way to do this could be pretty subjective, so no worries at all about further comments on making it better. I decided to interpolate the float values as they're the potentially problematic ones, and I put the integer value inside the string literal since integers aren't problematic like decimals represented with floating-point.

testing/matrix-sdk-test/src/event_factory.rs
yousefmoazzam: Done in fixup commit 819ee49

poljar (approved):
Looks good. Thanks for addressing the comments.

One suggestion for the future, if you're doing fixup commits, the individual commits should still compile.

I.e. if you change the function signature in a fixup commit, all the call sites need to be updated in the same commit.

It's not terribly important here, because the commits are simple and I'll just squash merge this anyways.

yousefmoazzam:
> One suggestion for the future, if you're doing fixup commits, the individual commits should still compile.
> 
> I.e. if you change the function signature in a fixup commit, all the call sites need to be updated in the same commit.

Ah my bad, sorry! I think I misunderstood the [fixup commit](https://github.com/matrix-org/matrix-rust-sdk/blob/main/CONTRIBUTING.md#addressing-review-comments-using-fixup-commits) section of the contributing guide.

I was thinking that, once the PR was approved, I'd then run `git rebase main --interactive --autosquash` locally to squash all the fixup commits (so then the individual fixup commits wouldn't need to compile in isolation, as they'd be squashed appropriately), before then force pushing that autosquashed final version of the branch to my fork.

Apologies again for my misunderstanding, I'll keep this in mind for future PRs :slightly_smiling_face:

PR #3119, opened 2024-02-12
timeline: mark as read based on the latest folded/invisible event

bnjbvr:
See commit by commit for an easier review.

TL;DR:

- FFI `mark_as_{un}read` has been changed to `set_unread_flag(bool)`
- a new method `Timeline::mark_as_read` is introduced in both the UI and FFI layers, attaching a read receipt to the latest event (not the event attached to the latest item)
- instead of `mark_as_unread_and_send_receipt`, it's expected of the callers to first call `set_unread_flag(false)` then `mark_as_read`
- instead of using `send_read_receipt` on the latest item's event id, when scrolled to the bottom in a room view, one would call `mark_as_read`

Pinging some app developers so they're aware of the new methods, cc @stefanceriu @ganfra @bmarty

Should fix with stuck unreads like https://github.com/element-hq/element-x-ios-rageshakes/issues/1409.

PR #5055, opened 2025-05-19
feat(sdk,ui): Unset the unread flag when sending unthreaded receipts

zecakeh:
This allows to follow [these recommendations from the spec](https://spec.matrix.org/v1.14/client-server-api/#client-behaviour-6) automatically:

> Clients SHOULD reset the unread marker by setting unread to false when opening a room to display its timeline.
>
> Clients that offer functionality to mark a room as read by sending a read receipt for the last event, SHOULD reset the unread marker simultaneously.

Note that there was a function in the bindings doing a similar thing that was removed in https://github.com/matrix-org/matrix-rust-sdk/pull/3119, but it doesn't seem that it was necessary to fix the bug that it was trying to fix.

### DIFF
--- a/crates/matrix-sdk-ui/src/timeline/mod.rs
+++ b/crates/matrix-sdk-ui/src/timeline/mod.rs
@@ -545,7 +545,7 @@ impl Timeline {
                 "not sending receipt, because we already cover the event with a previous receipt"
             );
 
-            if thread == ReceiptThread::Unthreaded && self.room().is_marked_unread() {
+            if thread == ReceiptThread::Unthreaded {
                 // Unset the read marker.
                 self.room().set_unread_flag(false).await?;
             }
@@ -608,7 +608,7 @@ impl Timeline {
 
         if !receipts.is_empty() {
             self.room().send_multiple_receipts(receipts).await?;
-        } else if self.room().is_marked_unread() {
+        } else {
             self.room().set_unread_flag(false).await?;
         }
 
@@ -632,10 +632,8 @@ impl Timeline {
         } else {
             trace!("can't mark room as read because there's no latest event id");
 
-            if self.room().is_marked_unread() {
-                // Unset the read marker.
-                self.room().set_unread_flag(false).await?;
-            }
+            // Unset the read marker.
+            self.room().set_unread_flag(false).await?;
 
             Ok(false)
         }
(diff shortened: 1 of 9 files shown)


crates/matrix-sdk-ui/src/timeline/mod.rs
```diff
@@ -595,7 +606,13 @@ impl Timeline {
             }
         }
 
-        self.room().send_multiple_receipts(receipts).await
+        if !receipts.is_empty() {
+            self.room().send_multiple_receipts(receipts).await?;
+        } else if self.room().is_marked_unread() {
+            self.room().set_unread_flag(false).await?;
+        }
```
Hywan, line 613: Can we add `let room = self.room();` to avoid calling `self.room()` multiple time please?

crates/matrix-sdk-ui/src/timeline/mod.rs
```diff
@@ -605,13 +622,21 @@ impl Timeline {
...
     #[instrument(skip(self), fields(room_id = ?self.room().room_id()))]
     pub async fn mark_as_read(&self, receipt_type: ReceiptType) -> Result<bool> {
         if let Some(event_id) = self.controller.latest_event_id().await {
             self.send_single_receipt(receipt_type, ReceiptThread::Unthreaded, event_id).await
         } else {
             trace!("can't mark room as read because there's no latest event id");
+
+            if self.room().is_marked_unread() {
+                // Unset the read marker.
+                self.room().set_unread_flag(false).await?;
```
Hywan, line 637: Can we add `let room = self.room();` to avoid calling `self.room()` multiple time please?

Hywan (changes requested):
Thanks for the PR. Apart from the minor feedback, we are good!

crates/matrix-sdk-ui/src/timeline/mod.rs
```diff
@@ -595,7 +606,13 @@ impl Timeline {
             }
         }
 
-        self.room().send_multiple_receipts(receipts).await
+        if !receipts.is_empty() {
+            self.room().send_multiple_receipts(receipts).await?;
+        } else if self.room().is_marked_unread() {
+            self.room().set_unread_flag(false).await?;
```
bnjbvr, line 612: Instead of specializing each call site of `Room::set_unread_flag()`, consider making `Room::set_unread_flag` a no-op if `!self.is_marked_unread()`?

crates/matrix-sdk-ui/src/timeline/mod.rs
zecakeh: With the removal of the condition, there is only one call here now.

crates/matrix-sdk-ui/src/timeline/mod.rs
zecakeh: Done.

zecakeh:

Hywan (approved):
Great, thanks!

PR #5056, opened 2025-05-19
chore(base): Move `BaseRoomInfo` and `RoomInfo` into their own module

Hywan:
These patches are a bit of clean up: the `BaseRoomInfo` type and `RoomInfo` type, in addition to their sibling types and implementations, are moved inside their own new `matrix_sdk_base::rooms::room_info` module.

More clean up are possible, but let's start with these ones.

There is no code change, it's purely code moves.


stefanceriu (approved):
🚀

PR #5057, opened 2025-05-19
refactor(timeline): use `TimelineAction::from_event` to create a `RepliedToEvent` struct

bnjbvr:
One more bullet item of #4823. The reply preview will now support all the types that are supported by the timeline, by default, and the UTD hook will also be called when creating a `RepliedToEvent` when the replied-to event is a UTD.

On top of #5046, so draft. Only the latest commit has been added on top of the aforementioned PR.

### DIFF
--- a/crates/matrix-sdk-ui/src/timeline/event_item/content/reply.rs
+++ b/crates/matrix-sdk-ui/src/timeline/event_item/content/reply.rs
@@ -23,7 +23,7 @@ use super::TimelineItemContent;
 use crate::timeline::{
     controller::TimelineMetadata,
     event_handler::TimelineAction,
-    event_item::{content::MsgLikeContent, EventTimelineItem, Profile, TimelineDetails},
+    event_item::{EventTimelineItem, Profile, TimelineDetails},
     traits::RoomDataProvider,
     Error as TimelineError, TimelineItem,
 };
@@ -97,7 +97,7 @@ impl RepliedToEvent {
         room_data_provider: &P,
         timeline_items: &Vector<Arc<TimelineItem>>,
         meta: &mut TimelineMetadata,
-    ) -> Result<Self, TimelineError> {
+    ) -> Result<Option<Self>, TimelineError> {
         let (raw_event, unable_to_decrypt_info) = match timeline_event.kind {
             matrix_sdk::deserialized_responses::TimelineEventKind::UnableToDecrypt {
                 utd_info,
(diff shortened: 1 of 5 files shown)


crates/matrix-sdk-ui/src/timeline/event_item/content/reply.rs
```diff
@@ -132,98 +116,26 @@ impl RepliedToEvent {
...
-
-                _ => {
-                    warn!("unsupported event type");
-                    return Err(TimelineError::UnsupportedEvent);
-                }
-            },
-
-            None => TimelineItemContent::MsgLike(MsgLikeContent::redacted()),
+        let sender = event.sender().to_owned();
+        let action = TimelineAction::from_event(
```
stefanceriu, line 120: Super nice 👌 .. although `TimelineAction` might no longer be the right name here.

stefanceriu: On another note: I wonder if populating those previously empty relations will create any issues 🤔

crates/matrix-sdk-ui/tests/integration/timeline/replies.rs
```diff
@@ -350,7 +350,13 @@ async fn test_fetch_details_utd() {
         let in_reply_to = in_reply_to.clone().unwrap();
         assert_let!(TimelineDetails::Ready(replied_to) = &in_reply_to.event);
         assert_eq!(replied_to.sender(), *ALICE);
-        assert!(replied_to.content().is_unable_to_decrypt());
```
stefanceriu, line 353: Why change this? 🤔

stefanceriu (approved):
Just some general questions, otherwise lgtm, thank you!

crates/matrix-sdk-ui/tests/integration/timeline/replies.rs
bnjbvr: Better error, when the things don't match (spoiler: it didn't match, at some point, during dev). It's easy to revert, though, if you feel strongly about it.

stefanceriu: No no, just curious

crates/matrix-sdk-ui/src/timeline/event_item/content/reply.rs
bnjbvr: Do you have in mind any particular issue it could cause?

bnjbvr: There's nothing that comes to mind, but we can watch out on this part during the next few releases :+1:

Issue #4718, opened 2025-02-25
ffi: refactor timeline creation and don't maintain timeline instances alive


bnjbvr:
There are multiple ways to create a timeline as of now, in the SDK-UI and the FFI layer, and some of these ways even involve holding onto a live timeline instance.

This is problematic, because we now have a mechanism to automatically make opening a timeline instance super fast, but for this, we need to make sure that there's no alive timeline in the background, otherwise it won't trigger.

So as to avoid this, and also simplify the FFI APIs to create a timeline, we should get rid of those cached timeline instances. I've spotted a few places:

- the FFI `Room` may hold onto a live timeline instance, when it's created from `RoomListItem::full_room`. This is used only to retrieve that timeline, or mark a room as read; in this case, we can plainly create a throwaway timeline and do it immediately
  - https://github.com/matrix-org/matrix-rust-sdk/pull/4716
- the FFI `RoomListItem` may hold onto a live time instance. In fact, if you want to get the `Room` from the `RoomListItem::full_room`, you'll need the latter's timeline to be initialized. This is a bit strange, because the room list item is likely transient, and holding a live timeline item into it doesn't quite make sense, as the timeline's refcount will decrease as soon as the room list item is gone, likely leading to the timeline being dropped. Also, you might want to get a handle to the underlying `Room` without manipulating its timeline, and the timeline MUST be initialized so as to get the room. The pattern of initializing the timeline of a `RoomListItem` is a bit contrived, involving several steps (checking if it's been initialized before, and initializing it only in that case).

Instead of these two, it might be possible to only use `Room::timeline_with_configuration`, provided that it implements as much as `RoomListItem::init_timeline`. For that to happen, we need the following:

- [ ] make sure that we have access to the UTD hook in `Room::timeline_with_configuration`.
  - either by providing it as an argument (bleh), or by stashing it in one FFI structure and reusing it there, or by having a UTD hook in the SDK's client. 
- [ ] make sure that we can filter on event type, in addition to message types, in `Room::timeline_with_configuration`.
- [ ] Then, the `room_list_service::room::RoomInner` shouldn't make use of its inner `timeline`; it's only used in a direct getter, or to compute the latest room event, but it's not working as intended, since local echoes aren't properly displayed in the room list. This non-working feature can be removed, in favor of https://github.com/matrix-org/matrix-rust-sdk/issues/4112 instead.
[...]

PR #5058, opened 2025-05-19
ffi: Refactor timeline creation and lifetimes

stefanceriu:
This PR implements the plan laid out in #4718: it removes long living timeline instances from the room list service and its items in favor of directly creating a timeline through its parent room and the `timeline_with_configuration` method.

In order to do so it also moves the previously internally configured UTD hook to the FFI client and exposes an option to use it on the aforementioned timeline configuration method.

NB: the event filter is already exposed in the configuration and the event cache is now enabled by default.

### DIFF
--- a/bindings/matrix-sdk-ffi/src/client.rs
+++ b/bindings/matrix-sdk-ffi/src/client.rs
@@ -220,7 +220,7 @@ impl From<matrix_sdk::TransmissionProgress> for TransmissionProgress {
 #[derive(uniffi::Object)]
 pub struct Client {
     pub(crate) inner: AsyncRuntimeDropped<MatrixClient>,
-    delegate: RwLock<Option<Arc<dyn ClientDelegate>>>,
+    delegate: OnceLock<Arc<dyn ClientDelegate>>,
     utd_hook: OnceLock<Arc<UtdHookManager>>,
     session_verification_controller:
         Arc<tokio::sync::RwLock<Option<SessionVerificationController>>>,
@@ -265,7 +265,7 @@ impl Client {
 
         let client = Client {
             inner: AsyncRuntimeDropped::new(sdk_client.clone()),
-            delegate: RwLock::new(None),
+            delegate: OnceLock::new(),
             utd_hook: OnceLock::new(),
             session_verification_controller,
         };
@@ -774,7 +774,7 @@ impl Client {
                 }
             });
 
-            *self.delegate.write().unwrap() = Some(Arc::from(delegate));
+            self.delegate.get_or_init(|| Arc::from(delegate));
             Arc::new(TaskHandle::new(session_change_task))
         })
     }
@@ -1486,7 +1486,7 @@ impl From<&search_users::v3::User> for UserProfile {
 
 impl Client {
     fn process_session_change(&self, session_change: SessionChange) {
-        if let Some(delegate) = self.delegate.read().unwrap().clone() {
+        if let Some(delegate) = self.delegate.get().cloned() {
             debug!("Applying session change: {session_change:?}");
             get_runtime_handle().spawn_blocking(move || match session_change {
                 SessionChange::UnknownToken { soft_logout } => {
--- a/bindings/matrix-sdk-ffi/src/client.rs
+++ b/bindings/matrix-sdk-ffi/src/client.rs
@@ -2,6 +2,7 @@ use std::{
     collections::HashMap,
     fmt::Debug,
     sync::{Arc, RwLock},
+    time::Duration,
 };
 
 use anyhow::{anyhow, Context as _};
@@ -38,9 +39,12 @@ use matrix_sdk::{
     store::RoomLoadSettings as SdkRoomLoadSettings,
     AuthApi, AuthSession, Client as MatrixClient, SessionChange, SessionTokens,
 };
-use matrix_sdk_ui::notification_client::{
-    NotificationClient as MatrixNotificationClient,
-    NotificationProcessSetup as MatrixNotificationProcessSetup,
+use matrix_sdk_ui::{
+    notification_client::{
+        NotificationClient as MatrixNotificationClient,
+        NotificationProcessSetup as MatrixNotificationProcessSetup,
+    },
+    unable_to_decrypt_hook::UtdHookManager,
 };
 use mime::Mime;
 use ruma::{
@@ -94,6 +98,7 @@ use crate::{
     },
     sync_service::{SyncService, SyncServiceBuilder},
     task_handle::TaskHandle,
+    utd::{UnableToDecryptDelegate, UtdHook},
     utils::AsyncRuntimeDropped,
     ClientError,
 };
@@ -216,6 +221,7 @@ impl From<matrix_sdk::TransmissionProgress> for TransmissionProgress {
 pub struct Client {
     pub(crate) inner: AsyncRuntimeDropped<MatrixClient>,
     delegate: RwLock<Option<Arc<dyn ClientDelegate>>>,
+    utd_hook_manager: Option<Arc<UtdHookManager>>,
     session_verification_controller:
         Arc<tokio::sync::RwLock<Option<SessionVerificationController>>>,
 }
@@ -225,6 +231,7 @@ impl Client {
         sdk_client: MatrixClient,
         enable_oidc_refresh_lock: bool,
         session_delegate: Option<Arc<dyn ClientSessionDelegate>>,
+        utd_delegate: Option<Arc<dyn UnableToDecryptDelegate>>,
     ) -> Result<Self, ClientError> {
         let session_verification_controller: Arc<
             tokio::sync::RwLock<Option<SessionVerificationController>>,
@@ -257,9 +264,10 @@ impl Client {
         let cross_process_store_locks_holder_name =
             sdk_client.cross_process_store_locks_holder_name().to_owned();
 
-        let client = Client {
-            inner: AsyncRuntimeDropped::new(sdk_client),
+        let mut client = Client {
+            inner: AsyncRuntimeDropped::new(sdk_client.clone()),
             delegate: RwLock::new(None),
+            utd_hook_manager: None,
             session_verification_controller,
         };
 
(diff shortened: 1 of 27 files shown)


bindings/matrix-sdk-ffi/src/client.rs
```diff
@@ -216,6 +221,7 @@ impl From<matrix_sdk::TransmissionProgress> for TransmissionProgress {
 pub struct Client {
     pub(crate) inner: AsyncRuntimeDropped<MatrixClient>,
     delegate: RwLock<Option<Arc<dyn ClientDelegate>>>,
+    utd_hook_manager: Option<Arc<UtdHookManager>>,
```
bnjbvr: Ah sorry, there's been a misunderstanding here: I was suggesting to move the `UtdHookManager` into the *SDK* `Client` (likely in the `ClientInner`), not the *FFI* `Client`. This way, you could avoid having to pass the `UtdHookManager` to every room you're creating in the next commit.

stefanceriu: Well I can do that as well but I would handle it as a separate PR, in the spirit of keeping it rather contained. The goal here was to simplify how we deal with timelines for which this should do the trick, no?

crates/matrix-sdk-ui/src/room_list_service/mod.rs
```diff
@@ -431,12 +429,6 @@ pub enum Error {
     #[error("Room `{0}` not found")]
     RoomNotFound(OwnedRoomId),
 
-    #[error("A timeline instance already exists for room {0}")]
-    TimelineAlreadyExists(OwnedRoomId),
-
-    #[error(transparent)]
-    InitializingTimeline(#[from] timeline::Error),
-
```
bnjbvr, line 439: Yep!

bnjbvr:
(removing Ivan from the reviewer list, as I've reviewed it.)

crates/matrix-sdk-ui/src/room_list_service/room.rs
```diff
@@ -77,56 +70,13 @@ impl Room {
...
-    /// doesn't have any latest event, it comes from the cache. This method
-    /// does not fetch any events or calculate anything — if it's not already
-    /// available, we return `None`.
+    /// Get the room's latest event from the cache. This method does not fetch
+    /// any events or calculate anything — if it's not already available, we
+    /// return `None`.
     ///
     /// Reminder: this method also returns `None` is the latest event is not
     /// suitable for use in a message preview.
     pub async fn latest_event(&self) -> Option<EventTimelineItem> {
```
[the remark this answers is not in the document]
stefanceriu: I believe this might also be an optimization reducing the amount of back and forth through FFI for the Android side 🤔

bindings/matrix-sdk-ffi/src/client.rs
```diff
@@ -797,6 +779,26 @@ impl Client {
         })
     }
 
+    pub async fn set_utd_delegate(self: Arc<Self>, utd_delegate: Box<dyn UnableToDecryptDelegate>) {
```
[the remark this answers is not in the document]
stefanceriu: > What was the hardship with having it in the builder

These events need to eventually get passed down into the analytics service which isn't normally availalbe before out client (the client proxy) is setup. The actions flow from this client proxy to the user session flow coordinator and the analytics service through it's own actions publisher. It's funky to setup the utd delegate before the proxy and its action publisher is setup. There's ways around it of course but it just makes for weirder code.

I guess the base question is: do you expect this thing to be act like a hot observable, to potentially fire as soon as the client is built because e.g. the event cache detected a UTD?

Another perhaps simpler option would be to expoxe the full UTD manager, have it store UTD infos and clients subscribe directly to it.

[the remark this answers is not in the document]
stefanceriu: Talked about this offline and decided that a `OnceCell` (OnceLock in this case) would be a good compromise solution.

bindings/matrix-sdk-ffi/src/room.rs
```diff
@@ -656,11 +656,11 @@ impl Room {
     /// Mark a room as read, by attaching a read receipt on the latest event.
     ///
     /// Note: this does NOT unset the unread flag; it's the caller's
-    /// responsibility to do so, if needs be.
+    /// responsibility to do so, if need be.
     pub async fn mark_as_read(&self, receipt_type: ReceiptType) -> Result<(), ClientError> {
-        let timeline = self.timeline().await?;
+        let timeline = TimelineBuilder::new(&self.inner).build().await?;
```
bnjbvr, line 667: Could this be: if there's a timeline created for this room, then use it; otherwise create a temporary one (like you're doing here)?

bindings/matrix-sdk-ffi/src/client.rs
```diff
@@ -221,7 +221,7 @@ impl From<matrix_sdk::TransmissionProgress> for TransmissionProgress {
 pub struct Client {
     pub(crate) inner: AsyncRuntimeDropped<MatrixClient>,
     delegate: RwLock<Option<Arc<dyn ClientDelegate>>>,
-    utd_hook_manager: Option<Arc<UtdHookManager>>,
+    utd_hook: Option<Arc<UtdHookManager>>,
```
bnjbvr: `utd_hook_manager` was the right name (it's a thin wrapper over the actual UTD hook provided by the user), can you keep it?

bindings/matrix-sdk-ffi/src/room.rs
```diff
@@ -75,16 +78,17 @@ pub(crate) type TimelineLock = Arc<RwLock<Option<Arc<Timeline>>>>;
 #[derive(uniffi::Object)]
 pub struct Room {
     pub(super) inner: SdkRoom,
+    utd_hook: Option<Arc<UtdHookManager>>,
```
bnjbvr, line 81: Let's call it `utd_hook_manager` here too

bindings/matrix-sdk-ffi/src/timeline/configuration.rs
```diff
@@ -157,4 +157,8 @@ pub struct TimelineConfiguration {
     /// As this has a non negligible performance impact, make sure to enable it
     /// only when you need it.
     pub track_read_receipts: bool,
+
+    /// Whether this timeline instance should reportd UTDs through the client's
+    /// delegate
```
bnjbvr, line 162: nit: dot at the end of the comment

bindings/matrix-sdk-ffi/src/room.rs
```diff
@@ -237,8 +237,10 @@ impl Room {
             builder = builder.with_internal_id_prefix(internal_id_prefix);
         }
 
-        if let Some(utd_hook) = self.utd_hook.clone() {
-            builder = builder.with_unable_to_decrypt_hook(utd_hook);
+        if configuration.report_utds {
+            if let Some(utd_hook) = self.utd_hook.clone() {
+                builder = builder.with_unable_to_decrypt_hook(utd_hook);
+            }
```
bnjbvr, line 239: Can we report an error when `report_utds` is true but there's no utd hook manager around?

labs/multiverse/src/main.rs
```diff
@@ -280,14 +280,13 @@ impl App {
                     }
                 };
 
-                if let Err(err) = ui_room.init_timeline_with_builder(builder).await {
-                    error!("error when creating default timeline: {err}");
+                let Ok(timeline) = builder.build().await else {
+                    error!("error when creating default timeline");
                     continue;
-                }
+                };
```
bnjbvr, line 286: Can you use a `match` statement so as to log the error message?

labs/multiverse/src/widgets/room_view/mod.rs
```diff
@@ -320,21 +319,14 @@ impl RoomView {
...
-        let Some(room) = selected.and_then(|room_id| self.ui_rooms.lock().get(room_id).cloned())
-        else {
-            self.status_handle.set_message("missing room or nothing to show".to_owned());
-            return;
-        };
-
-        // Mark as read!
-        let Ok(timeline) = TimelineBuilder::new(&room).build().await else {
-            self.status_handle.set_message("failed creating timeline".to_owned());
+        let Some(sdk_timeline) = self.selected_room.as_deref().and_then(|room_id| {
```
bnjbvr, line 322: nit: let's call this `timeline`, not `sdk_timeline`; we're not in the FFI layer here.

bindings/matrix-sdk-ffi/src/client.rs
```diff
@@ -796,7 +796,7 @@ impl Client {
             // because the UTD hook failed to load its data.
         }
 
-        *self.utd_hook.write().unwrap() = Some(Arc::new(utd_hook));
+        self.utd_hook.get_or_init(|| Arc::new(utd_hook));
```
bnjbvr, line 798: We'd probably want to throw an error if the utd hook manager was already set, because it could be super confusing for callers if they assumed that the UTD hook was overwritten, as before.

Let's add some note about that in this method's doc comment too, please.

bindings/matrix-sdk-ffi/src/client.rs
```diff
@@ -774,7 +774,7 @@ impl Client {
                 }
             });
 
-            *self.delegate.write().unwrap() = Some(Arc::from(delegate));
+            self.delegate.get_or_init(|| Arc::from(delegate));
```
bnjbvr, line 776: Same: let's throw if the delegate was already set, and let's add a doc comment explaining this invariant that it's a no op if the delegate was already set.

Probably you need to do the check at the start of the function, to avoid spawning a task that will be detached forever if the delegate was already set.

(Overall, I would have preferred this to be in a separate PR, because it's unrelated to the current one.)

bindings/matrix-sdk-ffi/src/client.rs
```diff
@@ -1488,13 +1487,10 @@ impl Client {
     fn process_session_change(&self, session_change: SessionChange) {
         if let Some(delegate) = self.delegate.get().cloned() {
             debug!("Applying session change: {session_change:?}");
-            get_runtime_handle().spawn_blocking(move || match session_change {
-                SessionChange::UnknownToken { soft_logout } => {
+            get_runtime_handle().spawn_blocking(move || {
+                if let SessionChange::UnknownToken { soft_logout } = session_change {
```
bnjbvr, line 1491: nit: let's keep a match statement, even if the `TokensRefreshed` arm is empty, that means that if we ever added another arm, we'd have to pause and think about what to do here (because the compiler would complain the match arm isn't handled)

bnjbvr (approved):
lgtm, thanks! we can likely get rid of more room wrappers in the SDK, after this…

Ideally, in terms of clean history management, we'd do some tidying up of the commits here:

- the "make ci happy" only contains changes that ought to be included in other commits. [`git absorb`](https://github.com/tummychow/git-absorb) is your friend in this kind of situations!
- the commit introducing the `OnceLock` for the UTD hook manager ought to be squashed with the other one that introduced the `Arc<Mutex<...>>` because this commit is undoing things that another commit made (fixup commit FTW!).

bindings/matrix-sdk-ffi/src/room.rs
stefanceriu: This was actually a miss on my part, the ffi room should no longer hold on to a timeline instance, which I've taken care of in the last commit.

PR #5059, opened 2025-05-19
feat(timeline): don't reset the pinned events timeline when underlying events haven't changed

bnjbvr:
This avoids a few spurious pinned event timeline resets, as shown by the test changes.


Hywan (commented):
The code looks okay, but I don't understand why we get an update with the same pinned events? Does it come from the server?

bnjbvr:
> The code looks okay, but I don't understand why we get an update with the same pinned events? Does it come from the server?

Every time the `RoomInfo` gets updated for *any* reason, the pinned events notifier would emit an update.

Hywan (approved):

PR #5060, opened 2025-05-19
feat(timeline): handle live aggregations on non-live timelines

bnjbvr:
This makes it possible to handle reactions/redactions/edits/etc. on non-live timelines. As a result, the pinned and focused timelines will now get live reactions/redactions and so on. This makes it possible to also have the thread timelines handle those live events, although it's unclear how it will pane out in the end, when the event cache is also involved.

Sits on top of #5046, hence drafting at the moment.


andybalaam (approved):
Looks good, thanks!

PR #5062, opened 2025-05-19
feat(bindings): added APIs to get the media preview config from the store

Velin92:
The reason behind this PR, is to extend https://github.com/matrix-org/matrix-rust-sdk/pull/5040 
While the observable value is very useful in the context of the main app, it may be problematic in the context of parallel processes that aim to aid the main app process, like for example the Notification Service Extension on iOS, which is treated as a different process, and has some very harsh memory contraints, and a limited lifetime, which would make the observable values not quite useful.

Also while we do have a [`get_media_preview_config_event_content`](https://github.com/matrix-org/matrix-rust-sdk/blob/0cfb4b1c55dd2ff8e90a6bb804944592ea819222/crates/matrix-sdk/src/account.rs#L1067) that is used by the observable API, this sends a request to the server to check for the current account data value of such event, and in the context of the NSE which already waits for a sync, we probably do not want to slow it down with another request to the server (not mentioning that sending a request for each notification might be also problematic for the servers).
For the NSE scenario I think getting whatever is currently in the store would be fine.

### DIFF
--- a/bindings/matrix-sdk-ffi/src/client.rs
+++ b/bindings/matrix-sdk-ffi/src/client.rs
@@ -1401,12 +1401,32 @@ impl Client {
     }
 
     /// Get the media previews timeline display policy
+    /// currently stored in the cache.
+    pub async fn get_stored_media_preview_display_policy(
+        &self,
+    ) -> Result<MediaPreviews, ClientError> {
+        let configuration =
+            self.inner.account().get_stored_media_preview_config_event_content().await?;
+        Ok(configuration.media_previews.into())
+    }
+
+    /// Set the invite request avatars display policy
     pub async fn set_invite_avatars_display_policy(
         &self,
         policy: InviteAvatars,
     ) -> Result<(), ClientError> {
         self.inner.account().set_invite_avatars_display_policy(policy.into()).await?;
         Ok(())
+    }
+
+    /// Get the invite request avatars display policy
+    /// currently stored in the cache.
+    pub async fn get_stored_invite_avatars_display_policy(
+        &self,
+    ) -> Result<InviteAvatars, ClientError> {
+        let configuration =
+            self.inner.account().get_stored_media_preview_config_event_content().await?;
+        Ok(configuration.invite_avatars.into())
     }
 }
 
--- a/bindings/matrix-sdk-ffi/src/client.rs
+++ b/bindings/matrix-sdk-ffi/src/client.rs
@@ -1402,11 +1402,8 @@ impl Client {
 
     /// Get the media previews timeline display policy
     /// currently stored in the cache.
-    pub async fn get_stored_media_preview_display_policy(
-        &self,
-    ) -> Result<MediaPreviews, ClientError> {
-        let configuration =
-            self.inner.account().get_stored_media_preview_config_event_content().await?;
+    pub async fn get_media_preview_display_policy(&self) -> Result<MediaPreviews, ClientError> {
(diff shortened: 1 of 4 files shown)


Velin92:
@poljar I wonder if we shouldn't actually use also in the [observable API](https://github.com/matrix-org/matrix-rust-sdk/blob/0cfb4b1c55dd2ff8e90a6bb804944592ea819222/crates/matrix-sdk/src/account.rs#L1026-L1088) the currently stored value, instead of doing a request to the server... WDYT?

crates/matrix-sdk/src/account.rs
```diff
@@ -1087,6 +1087,30 @@ impl Account {
         Ok(media_preview_config)
     }
 
+    /// Get the media preview configuration event content stored in the cache.
+    ///
+    /// This will return the default configuration if no value was found.
+    /// Will check first for the stable event and then for the unstable one.
+    pub async fn get_stored_media_preview_config_event_content(
```
zecakeh, line 1094: Could `observe_media_preview_config` use this method to get the initial value, or is there a reason that it must make a request to the homeserver?

poljar: Well, historically that has been a bit of a footgun because the sliding sync proxy tended to be a bit slow to sync the account data.

So you would get a stale value. I suppose it's not that bad to get a stale value here since you're going to receive the update eventually.

In conclusion, yeah I think you're right.

Velin92: okay perfect then I'll rework the existing code to use this in place of the API call

Velin92: Another question should I also do the same on the setters?
https://github.com/matrix-org/matrix-rust-sdk/blob/0cfb4b1c55dd2ff8e90a6bb804944592ea819222/crates/matrix-sdk/src/account.rs#L1114-L1142

Or the setters should keep fetching the API directly?

poljar: Answered in the chat room, but for completeness sake. The setters should still send out the request, in case the sync is late or not running, so we don't edit stale data.

bindings/matrix-sdk-ffi/src/client.rs
```diff
@@ -1401,13 +1401,33 @@ impl Client {
     }
 
     /// Get the media previews timeline display policy
+    /// currently stored in the cache.
+    pub async fn get_stored_media_preview_display_policy(
```
poljar, line 1405: The `stored` is a bit rendundant here. We have a kinda soft policy that we call functions that get things from the store `get_foo()` while we call functions that get things from the server `fetch_foo()`.

```suggestion
    pub async fn get_media_preview_display_policy(
```

crates/matrix-sdk/src/account.rs
```diff
@@ -1059,12 +1059,16 @@ impl Account {
...
 
         Ok((initial_value, result_stream))
     }
 
-    async fn get_media_preview_config_event_content(
+    /// Get the media preview configuration event content from the server.
+    ///
+    /// If no value was found in either the stable and unstable type, the
+    /// default configuration will be returned.
+    pub async fn get_media_preview_config_event_content(
```
poljar, line 1071: Now that we have two methods, let's call this one `fetch_...()` and the one that access the store `get()`, to avoid the overly long name.

poljar (changes requested):
Looks good, let's just rename the methods a bit.

poljar (approved):
Thanks, looking good.

PR #5064, opened 2025-05-20
feat(timeline): pass the encryption info for a bundled edit

bnjbvr:
This is a direct follow up to #4823, because it was deemed a bit too big for that PR. Also nicely groups all the parameters of `TimelineAction::from_content` when the event is a remote echo.

(I get more and more ideas to refactor this part of the timeline, please send help, in the form of contributor time or financial support to Element :D)

### DIFF
--- a/crates/matrix-sdk-ui/src/timeline/event_handler.rs
+++ b/crates/matrix-sdk-ui/src/timeline/event_handler.rs
@@ -186,6 +186,7 @@ impl TimelineAction {
         raw_event: &Raw<AnySyncTimelineEvent>,
         room_data_provider: &P,
         unable_to_decrypt_info: Option<UnableToDecryptInfo>,
+        bundled_edit_encryption_info: Option<EncryptionInfo>,
         timeline_items: &Vector<Arc<TimelineItem>>,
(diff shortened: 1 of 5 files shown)


crates/matrix-sdk-ui/src/timeline/event_handler.rs
```diff
@@ -243,10 +252,12 @@ impl TimelineAction {
...
-                            Some(raw_event),
-                            Some(ev.relations()),
-                            Some(ev.event_id()),
-                            bundled_edit_encryption_info,
+                            Some(RemoteEventContext {
+                                event_id: ev.event_id(),
+                                raw_event,
+                                relations: ev.relations(),
+                                bundled_edit_encryption_info,
+                            }),
```
Hywan, line 260: That's much nicer. Now we clearly identify they are all related.

Hywan (approved):
Here is my help: 🧀 and 🍫.

bnjbvr:
This builds properly on my machine. Caches busted?

bnjbvr:
Ah well, a rebase showed there were actual compilation errors. Alrighty then!

PR #5065, opened 2025-05-20
feat(sdk): add support for reporting errors in Sentry

### LINKED ISSUE #4843: structured logging: set up Sentry reporting
The FFI layer sets up the `tracing` infrastructure for recording traces and spans. This logic could be moved over to the SDK, to make it easier to setup tracing with sensible defaults, or with more details, if needs be. In addition to that, we could set up the tracing Sentry integration, so as to be able to specify a custom Sentry DSN and forward errors/panics (maybe that include a custom filtering tag) to Sentry. This would be super useful for:

- keeping track of the most frequent errors happening in the SDK
- also getting some rough performance information about some steps in the SDK (e.g. track performance of a sync, from starting the request to handling the response, etc.)

The next steps for this would be the following:

- [ ] Move the logic for setting up the logs from the FFI crate to the main crate, with sensible defaults + access to the tracing registry, so that users can customize it as they'd like to.
- [ ] Add support for a default sentry layer, and keep the sentry context alive in the `Client` struct

See also https://github.com/matrix-org/matrix-rust-sdk/issues/1463 which talks about related ideas.

bnjbvr:
This adds a new way to configure an optional Sentry layer, at the FFI layer. The thing is that Sentry requires keeping context around, for the whole time it's enabled, so we need to have some new global context for holding onto it. Initially this global context was part of the `Client` data, since the `Client` is kind of our global context in the SDK layer, but as pointed out by Stefan, these have different lifetimes: multiple `Client` instances can be created and destroyed while the logging must outlive all of them. As a matter of fact, I've decided to put this logging context in its own global variable, that's process-wide.

It also makes sense, with the perspective that our SDK shouldn't mandate any logging / tracing library over another, that it stays at the FFI layer.

Another comment then adds an error statement for the `database is locked` error we've observed a lot in the while, just so as to get a rough idea of how it evolves over time.

Fixes #4843.


stefanceriu (approved):
This looks great and it seems to be working but I don't know how to manually tigger a database busy error to test out the whole flow. It _should_ work so fingers crossed 🤞

PR #5067, opened 2025-05-21
chore(base): Reorganize the `matrix_sdk_base::rooms` module

Hywan:
This set of patches aims at reorganizing the `matrix_sdk_base::rooms` module. That's something I wanted to do since a long time, and that time has finally come.

This is a big diff, I recognized; but everything is gonna be alright, it's puuurely code move, nothing else. The patches contain descriptions to explain what moves every time;

The hierarchy was:

```
rooms/
├── members.rs
├── mod.rs
├── normal.rs
└── room_info.rs
```

Now it becomes:

```
room/
├── call.rs
├── create.rs
├── display_name.rs
├── encryption.rs
├── knock.rs
├── latest_event.rs
├── members.rs
├── mod.rs
├── room_info.rs
├── state.rs
└── tags.rs
```

The `rooms::normal` module was kind of huge, and was containing a LOT of types, methods, tests, logic etc. The idea of this set of patches is to split the `Room` implementation into many smaller implementations. For example, `rooms::display_name` has its own `impl Room` to add specific methods. There are pros and cons of that:

- pros: smaller modules, with isolated methods, isolated tests, clearer API
- cons: looking for a method directly in the code without LSP might be a bit unusual for some users

Nevertheless, I consider this is a nice improvement over the large module we had before.


poljar:
> This set of patches aims at reorganizing the `matrix_sdk_base::rooms` module. That's something I wanted to do since a long time, and that time has finally come.
> 
> This is a big diff, I recognized; but everything is gonna be alright, it's puuurely code move, nothing else. The patch contains descriptions to explain what moves every time;
> 
> The hierarchy was:
> 
> ```
> rooms/
> ├── members.rs
> ├── mod.rs
> ├── normal.rs
> └── room_info.rs
> ```
> 
> Now it becomes:
> 
> ```
> rooms/
> ├── call.rs
> ├── create.rs
> ├── display_name.rs
> ├── encryption.rs
> ├── knock.rs
> ├── latest_event.rs
> ├── members.rs
> ├── mod.rs
> ├── room_info.rs
> ├── state.rs
> └── tags.rs
> ```

Might make to get rid of the plural as well? `rooms/` -> `room/`.

Hywan:
Good idea. I initially wanted to do this in a separate PR, but doing this here also makes sense.

crates/matrix-sdk-base/src/rooms/room_info.rs
```diff
@@ -356,9 +356,10 @@ where
 /// Holds all the info needed to persist a room into the state store.
 #[derive(Clone, Debug, Serialize, Deserialize)]
 pub struct RoomInfo {
-    /// The version of the room info.
-    #[serde(default)]
-    pub(crate) version: u8,
+    /// The version of the room info type. It is used to migrate the `RoomInfo`
+    /// serialization from one version to another.
+    #[serde(default, rename = "version")]
```
poljar, line 400: Why not an alias? So we can deserialize using `version` but new entries will use the new name?

crates/matrix-sdk-base/src/rooms/members.rs
```diff
@@ -13,28 +13,188 @@
...
 
+use super::Room;
 use crate::{
     deserialized_responses::{DisplayName, MemberEvent, SyncOrStrippedState},
-    store::ambiguity_map::is_display_name_ambiguous,
+    store::{ambiguity_map::is_display_name_ambiguous, Result as StoreResult, StateStoreExt},
     MinimalRoomMemberEvent,
 };
 
+impl Room {
```
poljar, line 43: It would probably make sense to create an empty API struct like we have for `Client::encryption()` for this and some of the others, that way we wouldn't have `impl Room` blocks scattered around and the docs would then be more consumable as well.

I.e. if you want to know about `members` methods on the `Room` you would look at whichever struct we add here.

Though this might be something for a separate PR and would be a pretty big breaking change.

poljar (approved):
Very nice, this was something that was needed for a long time. Thanks for tackling this.

crates/matrix-sdk-base/src/rooms/room_info.rs
Hywan: Matter of taste. `version` is shorter, it takes less time in the storage. But I'm fine with an alias.

PR #5069, opened 2025-05-21
refactor(ui/ffi): get rid of `RoomListService::Room` and `RoomListItem`

bnjbvr:
Following #5058, these are now just thin wrappers over a SDK `Room`, so we might as well just use that instead.

After this PR, we now have """only""" 3 `Room`s: base / SDK / FFI. Not sure we'll achieve better than this :)


Hywan (approved):
We are removing most code we've written this past year. Strange but positive feeling! Thanks!

Hywan:
Now you need to update Complement Crypto 😋.

Issue #5050, opened 2025-05-16
Room list: Deduplication logic for tombstoned rooms


manuroe:
In order to fix upgraded rooms appearing twice or more in the room list like on https://github.com/element-hq/element-x-ios/issues/3105, we need to adopt a filtering logic so that only the most interesting room versions in the list.

The most interesting version is the latest version of the room the user is in. Ideally, the user should be in the version of the room that is not tombstoned but it is not guaranteed. They may not have joined that room yet. In this case, I see 2 options so far: 
- Use the room with the higher version number. We need to check if room upgrading is a monotonically increasing function
- Use `m.room.tombstone` events that form a linked list and pick the tail

We can implement such logic only if the SSS API returns `m.room.tombstone`. It seems to be the case [nowadays](https://github.com/matrix-org/matrix-rust-sdk/pull/4211#issuecomment-2877173857) and it is going to be enforced by the merge of https://github.com/matrix-org/matrix-rust-sdk/pull/4211.

Hywan:
Tasks:

- [x] Implement a predecessor/successor API
- [x] Check for loops in tombstoned rooms
- [x] Implement the filter for the room list
- [x] Implementing FFI bindings

Hywan:
@manuroe When a user leaves a room, should we also cascade to leave all the predecessor rooms? It can be implemented inside the SDK.

Hywan:
All good on the Rust side. Closing the issue :-).

manuroe:
> [@manuroe](https://github.com/manuroe) When a user leaves a room, should we also cascade to leave all the predecessor rooms? It can be implemented inside the SDK.

yes to align with Element Web that leaves every room.

PR #5070, opened 2025-05-21
feat(base): Create `Room::predecessor_room()` and `Room::successor_room()`

Hywan:
This patch adds 2 methods on `Room`:

1. `predecessor_room()`,
2. `successor_room()`.

If a room A is tombstoned, it is replaced by a room B. Room A is the predecessor, and room B is the successor. Having this clear semantics is helpful, and will help to solve https://github.com/matrix-org/matrix-rust-sdk/issues/5050.

Documentation has been improved, along with adding missing tests, and writing new tests.

---

* Address https://github.com/matrix-org/matrix-rust-sdk/issues/5050

### DIFF
--- a/crates/matrix-sdk-base/src/room/tombstone.rs
+++ b/crates/matrix-sdk-base/src/room/tombstone.rs
@@ -116,11 +116,9 @@ mod tests {
 
     use assert_matches::assert_matches;
     use matrix_sdk_test::{
-        async_test, event_factory::EventFactory, JoinedRoomBuilder, StateTestEvent,
-        SyncResponseBuilder,
+        async_test, event_factory::EventFactory, JoinedRoomBuilder, SyncResponseBuilder,
     };
-    use ruma::{event_id, room_id, user_id};
-    use serde_json::json;
+    use ruma::{event_id, room_id, user_id, RoomVersionId};
 
     use crate::{test_utils::logged_in_base_client, RoomState};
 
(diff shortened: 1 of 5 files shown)


crates/matrix-sdk-base/src/room/tombstone.rs
```diff
@@ -12,18 +12,291 @@
...
+///
+/// [`m.room.tombstone`]: https://spec.matrix.org/v1.14/client-server-api/#mroomtombstone
+/// [`m.room.create`]: https://spec.matrix.org/v1.14/client-server-api/#mroomcreate
+#[derive(Debug)]
+pub struct PredecessorRoom {
+    /// The ID of the old room.
+    pub room_id: OwnedRoomId,
+
+    /// The event ID of the last known event in the predecesssor room.
+    pub event_id: OwnedEventId,
```
bnjbvr, line 110: Could we name this `last_event_id`? Otherwise it's very unclear what event id this is…

crates/matrix-sdk-base/src/room/tombstone.rs
```diff
@@ -12,18 +12,291 @@
...
+                    },
+                    "event_id": "$ev0",
+                    "origin_server_ts": 42,
+                    "sender": sender,
+                    "state_key": "",
+                    "type": "m.room.create",
+                    "unsigned": {
+                        "age": 43
+                    }
+                })),
```
bnjbvr, line 290: ditto here

crates/matrix-sdk-base/src/room/tombstone.rs
```diff
@@ -12,18 +12,291 @@
...
+                },
+                RoomLoadSettings::default(),
+                #[cfg(feature = "e2e-encryption")]
+                None,
+            )
+            .await
+            .unwrap();
+
+        base_client
+    }
```
bnjbvr, line 148: No chance this helper could be reused? Looks like if it's needed here, it would be needed in every other file that makes use of a `BaseClient` for testing purposes…

bnjbvr (approved):
Ok, it's hard to see the value of these structs renaming the `create` and `tombstone` event contents. And might be confusing too; at first, I thought we were talking about predecessor and successor in the room list, since those two terms are so generic. Why not, I guess.

Can you fold 12608b6fa7b2baebe05c8ac8c3c46127df4b9e79 into the commit it pertains to? `git absorb` is your friend :)

crates/matrix-sdk-base/src/room/tombstone.rs
```diff
@@ -12,18 +12,291 @@
...
+                    },
+                    "event_id": "$ev0",
+                    "origin_server_ts": 42,
+                    "sender": sender,
+                    "state_key": "",
+                    "type": "m.room.create",
+                    "unsigned": {
+                        "age": 43
+                    }
+                })),
```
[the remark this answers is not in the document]
Hywan: Done.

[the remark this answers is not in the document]
Hywan: We should really get rid off `StateTestEvent`…

crates/matrix-sdk-base/src/room/tombstone.rs
Hywan: Of course, we have `logged_in_base_client` already. Updated.

crates/matrix-sdk-base/src/room/tombstone.rs
Hywan: Sure! Good idea.

PR #5072, opened 2025-05-22
refactor: when leaving an `Invited` room also forget it

jmartinesp:
This behaviour was added only at the `RoomPreview::leave` method, but since we're slowly moving away from it we should move the forget action to the `Room::leave` method instead.

Signed-off-by:

### DIFF
--- a/crates/matrix-sdk/tests/integration/room/joined.rs
+++ b/crates/matrix-sdk/tests/integration/room/joined.rs
@@ -19,8 +19,8 @@ use matrix_sdk_test::{
     event_factory::EventFactory,
     mocks::mock_encryption_state,
     test_json::{self, sync::CUSTOM_ROOM_POWER_LEVELS},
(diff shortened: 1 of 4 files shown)


jmartinesp:
And now that GH actions work codecov got stuck again 🥲

crates/matrix-sdk/tests/integration/room/joined.rs
```diff
@@ -126,6 +126,28 @@ async fn test_leave_room() -> Result<(), anyhow::Error> {
     Ok(())
 }
 
+#[async_test]
+async fn test_leave_invited_room_also_forgets_it() -> Result<(), anyhow::Error> {
```
andybalaam, line 130: Can we test that we _don't_ forget a room that we were joined?

andybalaam (approved):
Looks good, one light suggestion.

crates/matrix-sdk/tests/integration/room/joined.rs
jmartinesp: Done and force pushed the new test.

PR #5074, opened 2025-05-22
feat(crypto): Add `EncryptionInfo` to `Decrypted` to-device variant

BillCarsonFr:
The `ProcessedToDeviceEvent::Decrypted` variant now also have an `EncryptionInfo` field.
Format changed from `Decrypted(Raw<AnyToDeviceEvent>)` to `Decrypted { raw: Raw<AnyToDeviceEvent>, encryption_info: EncryptionInfo) }`

It is needed for future work on widget in order to support encrypted to device  properly via`add_event_handler` 

This is a follow-up on 
- https://github.com/matrix-org/matrix-rust-sdk/pull/4935
- https://github.com/matrix-org/matrix-rust-sdk/pull/4981

Signed-off-by:


crates/matrix-sdk-crypto/src/types/mod.rs
```diff
@@ -633,8 +636,14 @@ mod test {
 #[derive(Clone, Debug, Serialize, Deserialize)]
 pub enum ProcessedToDeviceEvent {
     /// A successfully-decrypted encrypted event.
-    /// Contains the raw decrypted event .
-    Decrypted(Raw<AnyToDeviceEvent>),
+    /// Contains the raw decrypted event and encryption info
+    #[serde(deserialize_with = "deserialize_decrypted_variant")]
+    Decrypted {
+        /// The raw decrypted event
+        raw: Raw<AnyToDeviceEvent>,
```
poljar, line 643: This doesn't need to be accessible?

```suggestion
        pub raw: Raw<AnyToDeviceEvent>,
```

crates/matrix-sdk-crypto/src/types/mod.rs
```diff
@@ -633,8 +636,14 @@ mod test {
...
+        /// The olm encryption info
+        encryption_info: EncryptionInfo,
```
poljar, line 645: Likewise:

```suggestion
        pub encryption_info: EncryptionInfo,
```

crates/matrix-sdk-crypto/src/types/mod.rs
```diff
@@ -648,12 +657,66 @@ pub enum ProcessedToDeviceEvent {
...
+    // In the old format the Raw<AnyToDeviceEvent> was just flatten out here.
+    // With the new format the event is encapsulated under `raw` and there is a
+    // `encryption_info` field
+    if json.get("raw").is_some() && json.get("encryption_info").is_some() {
+        // If the "raw" field is present, it's a NewFormat
+        #[derive(Deserialize)]
+        struct NewFormat {
+            raw: Raw<AnyToDeviceEvent>,
+            encryption_info: EncryptionInfo,
+        }
```
poljar, line 682: What happens if someone puts a `raw` and `encryption_info` field into the JSON of the event?

It seems to me that the sender of the event, perhaps in collusion with the homeserver, might insert an `encryption_info` on their own.

Come to think of, why does this need to support `Serialize` and `Deserialize`? Because the widgets API will transfer this verbatim to the widget using JSON?

In any case, I think it's safer to not have a migration here since we don't seem to need it.

poljar (changes requested):

crates/matrix-sdk-common/src/deserialized_responses.rs
```diff
@@ -287,6 +287,12 @@ pub enum AlgorithmInfo {
         #[serde(default, skip_serializing_if = "Option::is_none")]
         session_id: Option<String>,
     },
+
+    /// The info if the event was encrypted using m.olm.v1.curve25519-aes-sha2
+    OlmV1Curve25519AesSha2 {
+        // The sender device key, base64 encoded
+        curve25519_public_key_base64: String,
```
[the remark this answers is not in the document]
BillCarsonFr: It tried to keep it consistent with the other variant. 
I suppose I have then to use `serde(deserialize_with = "deserialize_curve_key", .. "serialize_curve_key"`?
I'll do it too for `MegolmV1AesSha2` to stay consistant

crates/matrix-sdk-crypto/src/types/mod.rs
BillCarsonFr: > Come to think of, why does this need to support Serialize and Deserialize?

I'll check but I think it mostly to export then to bindings?
But I agree that those should not be serialized, but we previously said that if we put something as serializable we should provide migration. I wish we had another trait for things that can be persisted

crates/matrix-sdk-crypto/src/types/mod.rs
BillCarsonFr: It is already public because the "parent" enum is public.
> note: enum variants and their fields always share the visibility of the enum they are in

crates/matrix-sdk-crypto/src/types/mod.rs
BillCarsonFr: So you are right.. Serialize and Deserialize are not used at all (I also checked in wasm bindings and it is not). Removing all that

crates/matrix-sdk-common/src/deserialized_responses.rs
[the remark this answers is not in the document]
BillCarsonFr: Actually I am not sure about this change.
To do it, I need to add vodozemac as a dependency of matrix-sdk-common (it is not yet), would I need also to add some `feature = "e2e-encryption"` cfg?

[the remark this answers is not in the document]
BillCarsonFr: I had another look, and if I just refactor both Olm and Megolm variant to use `Curve25519Publickey` instead of strings, I'll need to 
- make the `common` crate to depend on `vodozemac`
- Then for ser/deser compatibility (particularly for Megolm variant as part of `TimelineEventKind`) we need to move  `deserialize_curve_key` and `serialize_curve_key` from crypto crate down to common crate (that will also bring back a dependency to zeroize).

This is touching several tests files. Looks like a candidate for a separate PR

crates/matrix-sdk-crypto/src/types/mod.rs
poljar: Ah duh. It's not a struct.

crates/matrix-sdk-crypto/src/olm/account.rs
```diff
@@ -127,6 +130,7 @@ pub(crate) struct DecryptionResult {
     pub event: Box<AnyDecryptedOlmEvent>,
     pub raw_event: Raw<AnyToDeviceEvent>,
     pub sender_key: Curve25519PublicKey,
```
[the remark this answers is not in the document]
poljar: Hmm, but wouldn't that method then return an `Option`? If so, I think that this would be pretty annoying to deal with.

crates/matrix-sdk-common/Cargo.toml
```diff
@@ -50,10 +50,11 @@ wasm-bindgen-futures = { version = "0.4.33", optional = true }
 web-sys = { workspace = true, features = ["console"] }
 
 [dev-dependencies]
-assert_matches.workspace = true
+assert_matches = { workspace = true }
+assert_matches2. workspace = true
+proptest.workspace = true
```
poljar: Why the move of the proptest import here? This should ideally be sorted alphabetically.

crates/matrix-sdk-crypto/CHANGELOG.md
```diff
@@ -10,7 +10,11 @@ All notable changes to this project will be documented in this file.
 
 - [**breaking**] Move `session_id` from `EncryptionInfo` to `AlgorithmInfo` as it is megolm specific. 
   Use `EncryptionInfo::session_id()` helper for quick access.
-  ([#4981](https://github.com/matrix-org/matrix-rust-sdk/pull/4775), [#4864](https://github.com/matrix-org/matrix-rust-sdk/pull/4981))
+  ([4981](https://github.com/matrix-org/matrix-rust-sdk/pull/4981))
+
+- [**breaking**] The `ProcessedToDeviceEvent::Decrypted` variant now also have an `EncryptionInfo` field.
+  Format changed from `Decrypted(Raw<AnyToDeviceEvent>)` to `Decrypted { raw: Raw<AnyToDeviceEvent>, encryption_info: EncryptionInfo) }`
+  ([5074](https://github.com/matrix-org/matrix-rust-sdk/pull/5074))
```
poljar, line 13: Let's move this to the top, we should add new entries at the top. Also did you intentionally remove the second PR link from the previous entry?

crates/matrix-sdk-crypto/src/olm/account.rs
```diff
@@ -1502,12 +1506,52 @@ impl Account {
...
+                )));
+
+            let encryption_info = EncryptionInfo {
+                sender: event.sender().to_owned(),
+                sender_device: sender_device.map(|d| d.device_id().to_owned()),
+                algorithm_info: AlgorithmInfo::OlmV1Curve25519AesSha2 {
+                    curve25519_public_key_base64: sender_key.to_base64(),
+                },
+                verification_state,
+            };
```
poljar: Shame that this can't be a method on the `Device` but let's at least put this into a separate method in the `Account`.

poljar (changes requested):
Alright, I still have a couple of nits I would like to see fixed.

But I think we're good after that.

crates/matrix-sdk-crypto/CHANGELOG.md
BillCarsonFr: Done https://github.com/matrix-org/matrix-rust-sdk/pull/5074/commits/ef03d242c2bc450cd0b241e28c65b2ee0f0e1be3

> Also did you intentionally remove the second PR link from the previous entry?

Not sure what happened but it was added when merging main. This was pointing to an unrelated PR (that is a couple of line below for history key sharing)

crates/matrix-sdk-crypto/src/olm/account.rs
BillCarsonFr: done https://github.com/matrix-org/matrix-rust-sdk/pull/5074/commits/df2863a4d98bff40e1e867789f19f074e1e6011b

crates/matrix-sdk-common/Cargo.toml
BillCarsonFr: I did some cleanup. The merge maked it a bit difficult to make sense out of that. 
I changed with proper order and usage of short syntax `.workspace` when applicable https://github.com/matrix-org/matrix-rust-sdk/pull/5074/commits/3bfc862e88d5fee72d885e09dfd9cf994f176964

crates/matrix-sdk-crypto/CHANGELOG.md
poljar: Ah, ok.

poljar (approved):
Looks good.

PR #5076, opened 2025-05-22
chore: Clean up Cargo.toml formatting

jplatte:
First commit mostly automated using `cargo sort`. A few manual changes were made because the edits invalidated some comments, and because there were extra newlines that didn't make sense.

CI integration semi-blocked on https://github.com/taiki-e/install-action/pull/980. Should be a matter of installing + running `cargo sort --workspace --grouped`.


jplatte:
Huh, looks like after the second commit, `cargo sort` starts to put target-specific dependencies before non-target-specific dependencies in many manifests. There's definitely a bug there 😅

Hywan (approved):
Thanks for this! I didn't know about [`cargo-sort`](https://crates.io/crates/cargo-sort). Of course, you're one of the owners 😉.

Do you think it could be added in the CI as a check to ensure we don't regress on the `Cargo.toml` style?

jplatte:
Yes, later. I already noted why I didn't add it to CI in this PR 😅

PR #5078, opened 2025-05-22
Test: Replace `sync_timeline_event!` with `EventFactory` for plain text room message event in notification client test

yousefmoazzam:
Part of #3716

I added some extra variables to avoid duplicating hardcoded values which could go out of sync if one is changed but the other isn't, hope that's a reasonable thing to do alongside the change I was making!


andybalaam:
Thanks for the PR! I think we have a known problem with the complement-crypto tests, so probably unrelated to your changes.

andybalaam (approved):
Thanks, this looks great. I'll try to merge it today or ask someone else to look after it next week.

PR #5079, opened 2025-05-22
Add WASM32 core platform abstractions

zzorba:
This commit adds foundational WASM32 support to matrix-sdk-common by introducing platform-agnostic abstractions for async operations:

- async_lock.rs: WASM32-compatible async RwLock implementation using std::sync primitives with async interface
- stream.rs: WASM32 stream utilities using LocalBoxStream instead of BoxStream
- executor.rs: Enhanced with comprehensive WASM runtime abstraction including WasmRuntimeHandle, unified Handle/Runtime types, and improved JoinError handling
- Dependencies: Added futures-executor for WASM32 and async-compat for non-WASM32 targets

These abstractions enable matrix-rust-sdk to run on both WASM32 and native targets without conditional compilation in consuming code.

This is part of a larger effort by Filament to add wasm32 support to the matrix-rust-sdk and matrix-sdk-ffi bindings.  We have these bindings working on our fork, and are now taking steps to cherry pick them incrementally back to main to make them more reviewable.

Signed-off-by: Daniel Salinas


crates/matrix-sdk-common/src/async_lock.rs
jplatte, line 1: That's a lot of `#[cfg(target_arch = "wasm32")]`s. How about doing it like this:

```rust
#[cfg(not(target_arch = "wasm32"))]
pub use tokio::sync::{ ... };

pub use self::async_lock_wasm::*;

mod async_lock_wasm {
    // all the current cfg'd items
}
```

crates/matrix-sdk-common/src/executor.rs
```diff
@@ -47,24 +73,28 @@ where
...
-    JoinHandle { remote_handle: Some(remote_handle), abort_handle }
+    JoinHandle { remote_handle: Some(remote_handle), the_abort_handle: abort_handle }
 }
 
 #[cfg(target_arch = "wasm32")]
 #[derive(Debug)]
 pub struct JoinHandle<T> {
     remote_handle: Option<RemoteHandle<T>>,
-    abort_handle: AbortHandle,
+    the_abort_handle: AbortHandle,
```
jplatte, line 83: Why rename this?

crates/matrix-sdk-common/src/executor.rs
```diff
@@ -83,17 +113,120 @@ impl<T: 'static> Future for JoinHandle<T> {
...
             Pin::new(handle).poll(cx).map(Ok)
         } else {
-            Poll::Ready(Err(JoinError))
+            Poll::Ready(Err(JoinError::Panic))
         }
     }
 }
 
+#[cfg(target_arch = "wasm32")]
+use futures_executor;
```
jplatte, line 128: This is a no-op, you can remove it.

crates/matrix-sdk-common/src/async_lock.rs
maan2003, line 1: why? tokio::sync::RwLock works on wasm

crates/matrix-sdk-common/src/stream.rs
```diff
@@ -0,0 +1,78 @@
...
+#[cfg(not(target_arch = "wasm32"))]
+pub use futures_util::{stream::BoxStream, StreamExt};
+
+/// Future for the [`next`](super::StreamExt::next) method.
+#[cfg(target_arch = "wasm32")]
+#[derive(Debug)]
+#[must_use = "futures do nothing unless you `.await` or poll them"]
+pub struct Next<'a, St: ?Sized> {
+    stream: &'a mut St,
+}
```
maan2003, line 23: why is this needed? I think futures_util::stream::StreamExt's next should just work

crates/matrix-sdk-common/src/executor.rs
maan2003, line 1: would strongly recommend https://docs.rs/n0-future/latest/n0_future/ instead of making our own abstractions

Hywan (changes requested):
Thanks for your contribution!

I've a couple of remarks though.

- I think it's better to split your single patch into one patch per _features_: one patch for `async_lock`, one patch for `stream` etc. It would help the reviewer,
- I don't understand why `async_lock` is necessary at all since `tokio::sync` is Wasm-compatible (see [`tokio` Wasm support](https://docs.rs/tokio/latest/tokio/#wasm-support)),
- I suggest to use the `#[cfg(target…)] mod sys` + `pub use sys::*` pattern more broadly to reduce most of the `#[cfg(target…)]` you have; it will also greatly clarify the code!
- I suggest to use `target_family = "wasm"` to target Wasm instead of `target_arch = "wasm32"`,
- I don't understand the need for `WasmRuntime` because `tokio::runtime` is Wasm-compatible (see the same link for `tokio::sync`),
- I don't understand why you re-implement `Next`, it is by-design Wasm-compatible. I understand the need to “alias” `LocalBoxStream` to `BoxStream` though.

In general, please justify the reason of being of those types in their respective documentation please.

maan2003:
> I don't understand the need for WasmRuntime because tokio::runtime is Wasm-compatible (see the same link for tokio::sync),

actually no, tokio rt on wasm doesn't integrate with the browser (for fetch() etc), many wasm compatible crates require the wasm-bindgen-futures runtime

crates/matrix-sdk-common/src/async_lock.rs
jplatte: I guess this thread is irrelevant given the other one on this file ^^

zzorba: Okay, I will make these corrections in these shim files.  It seemed like the other was the norm but happy to adjust it.

zzorba: You are correct, the async_compat is not necessary and I have removed it.

crates/matrix-sdk-common/src/stream.rs
zzorba: The issue is the call to .boxed() on the stream it returns, the StreamExt has has a `Send` trait requirement that prevents .boxed() from being called when used in situations like this:

```
        let (initial, mut subscriber): (RoomPaginationStatus, BoxStream<'_, RoomPaginationStatus>) =
            match self.inner.live_back_pagination_status().await {
                Some((status, stream)) => (status, stream.boxed()),
                None => {
                    let (status, stream) = self.inner.focused_pagination_status().await.context(
                        "can't subscribe to the back-pagination status on a focused timeline",
                    )?;
                    (status, stream.boxed())
                }
            };
```

Here is the definition of .boxed() from the futures_util::stream::StreamExt

```
    /// Wrap the stream in a Box, pinning it.
    ///
    /// This method is only available when the `std` or `alloc` feature of this
    /// library is activated, and it is activated by default.
    #[cfg(feature = "alloc")]
    fn boxed<'a>(self) -> BoxStream<'a, Self::Item>
    where
        Self: Sized + Send + 'a,
    {
        assert_stream::<Self::Item, _>(Box::pin(self))
    }
```

zzorba:
1) I will split up patch into two commits (one for Stream, one for Executor, as async_lock is indeed not necessary)
2) Have adjusted code to use mod sys exports.
3) Moved to target_family = "wasm"
4) Tokio rt does not work with wasm, I believe an alternative runtime is necessary as a result.  
5) The issue is the .boxed() call on StreamExt, which has a Send trait.  I will try to find a way to add this as an extension for Wasm in a way that doesn't require redefining Stream.

crates/matrix-sdk-common/src/executor.rs
jplatte: That's not true, you can have a method with the same name as a field just fine. It's somewhat common, even.

crates/matrix-sdk-common/src/executor.rs
```diff
@@ -83,17 +113,120 @@ impl<T: 'static> Future for JoinHandle<T> {
...
+
+#[cfg(target_arch = "wasm32")]
+pub type Handle = WasmRuntimeHandle;
+#[cfg(target_arch = "wasm32")]
+pub type Runtime = WasmRuntimeHandle;
+
+#[cfg(target_arch = "wasm32")]
+#[derive(Debug)]
+/// A dummy guard that does nothing when dropped.
+/// This is used for the WASM implementation to match
```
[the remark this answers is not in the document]
zzorba: More spots in the codebase are referring to it as WASM not Wasm, including in Readme files.  Are we sure we want to change it here to `Wasm`

crates/matrix-sdk-common/src/executor.rs
zzorba: Thank you for this suggestion, I will update this in the split out PR to use this library for the shim instead.

zzorba:
Okay, I have split up this PR:

Stream extension: https://github.com/matrix-org/matrix-rust-sdk/pull/5085
Executor/JoinHandle extensions: https://github.com/matrix-org/matrix-rust-sdk/pull/5088
Runtime: https://github.com/matrix-org/matrix-rust-sdk/pull/5089 (dependent on 5088, will mark as ready once 5088 is merged)

PR #5081, opened 2025-05-23
fix: Do not retrieve notification events sent by ignored users

jmartinesp:
When retrieving notifications using either sliding sync or the `/context` endpoint, check if the sender is ignored by the user before returning it. If it is, just filter out the notification.

Signed-off-by:

### DIFF
--- a/crates/matrix-sdk-ui/tests/integration/notification_client.rs
+++ b/crates/matrix-sdk-ui/tests/integration/notification_client.rs
@@ -4,10 +4,13 @@ use std::{
 };
 
 use assert_matches::assert_matches;
-use matrix_sdk::{config::SyncSettings, test_utils::logged_in_client_with_server};
+use matrix_sdk::{
+    config::SyncSettings,
+    test_utils::{logged_in_client_with_server, mocks::MatrixMockServer},
+};
 use matrix_sdk_test::{
-    async_test, mocks::mock_encryption_state, sync_timeline_event, JoinedRoomBuilder,
-    StateTestEvent, SyncResponseBuilder,
+    async_test, event_factory::EventFactory, mocks::mock_encryption_state, sync_timeline_event,
+    JoinedRoomBuilder, StateTestEvent, SyncResponseBuilder,
 };
 use matrix_sdk_ui::{
     notification_client::{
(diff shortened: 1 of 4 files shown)


crates/matrix-sdk/src/client/mod.rs
```diff
@@ -2541,6 +2541,13 @@ impl Client {
         let base_room = self.inner.base_client.room_knocked(&response.room_id).await?;
         Ok(Room::new(self.clone(), base_room))
     }
+
+    /// Checks whether the provided `user_id` belongs to an ignored user.
+    pub fn is_user_ignored(&self, user_id: &UserId) -> bool {
+        let raw_user_id = user_id.to_string();
+        let current_ignored_user_list = self.subscribe_to_ignore_user_list_changes().get();
+        current_ignored_user_list.contains(&raw_user_id)
+    }
```
Hywan, line 2550: `subscribe_to_ignore_user_list_changes` is only “runtime-aware”: it knows the changes that happen during the current run of the SDK. If a user has been ignored in a previous instance of the SDK, you will miss it.

I think it's better to use something like this (not tested):

```rust
let state_store = get_it_from_somewhere();
let list  = state_store.get_account_data_event_static::<IgnoredUserListEvent>()…;
```

The new `is_user_ignored` is nice! That's a good idea. It's just the body of the method that should change.

Also, the method should be on `matrix_sdk_base::Client`, not `matrix_sdk::Client`.

Hywan (changes requested):

crates/matrix-sdk-base/src/client.rs
```diff
@@ -960,6 +961,27 @@ impl BaseClient {
...
+    /// Checks whether the provided `user_id` belongs to an ignored user.
+    pub async fn is_user_ignored(&self, user_id: &UserId) -> bool {
+        match self.state_store.get_account_data_event_static::<IgnoredUserListEventContent>().await
+        {
+            Ok(Some(raw_ignored_user_list)) => match raw_ignored_user_list.deserialize() {
+                Ok(current_ignored_user_list) => {
+                    current_ignored_user_list.content.ignored_users.contains_key(user_id)
+                }
+                Err(error) => {
+                    warn!("Failed to deserialize ignored user list event: {error:?}.");
```
Hywan, line 974:
```suggestion
                    warn!(?error, "Failed to deserialize the ignored user list event");
```

crates/matrix-sdk-base/src/client.rs
```diff
@@ -960,6 +961,27 @@ impl BaseClient {
...
+                    false
+                }
+            },
+            Ok(None) => false,
+            Err(error) => {
+                warn!("Could not get ignored user list. {error:?}.");
```
Hywan, line 980:
```suggestion
                warn!(?error, "Could not get the ignored user list from the state store");
```

crates/matrix-sdk-base/src/client.rs
```diff
@@ -1597,4 +1619,38 @@ mod tests {
...
+
+        client
+            .activate(
+                SessionMeta { user_id: user_id.to_owned(), device_id: "FOOBAR".into() },
+                RoomLoadSettings::default(),
+                #[cfg(feature = "e2e-encryption")]
+                None,
+            )
+            .await
+            .unwrap();
```
Hywan, line 1637:
```suggestion
        let client = logged_in_base_client(None).await;
```

Hywan (changes requested):
Almost there! Thanks for the new patch.

crates/matrix-sdk-ui/tests/integration/notification_client.rs
```diff
@@ -1059,3 +1062,227 @@ async fn test_notification_client_mixed() {
...
+                                "event_id": "$151800140517rfvjc:example.org",
+                                "membership": "join",
+                                "origin_server_ts": 151800140,
+                                "sender": sender,
+                                "state_key": sender,
+                                "type": "m.room.member",
+                                "unsigned": {
+                                    "age": 2970366,
+                                }
+                            },
```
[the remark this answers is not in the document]
jmartinesp: These are mostly copy-pasted from other tests, but sure, I can do that.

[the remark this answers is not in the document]
jmartinesp: I can probably do that on a separate PR.

PR #5082, opened 2025-05-23
Comprehensive Send+Sync bound improvements for WASM32 compatibility

zzorba:
This commit systematically replaces Send+Sync trait bounds throughout the matrix-rust-sdk codebase to enable WASM32 compatibility while maintaining thread safety on native targets.

Key changes:
- Use SendOutsideWasm/SyncOutsideWasm traits instead of Send/Sync for trait bounds
- Apply conditional compilation for error types and trait objects
- Update FFI trait definitions to use WASM32-compatible bounds
- Fix event handler and type alias definitions for cross-platform compatibility
- Maintain existing functionality while enabling WebAssembly target support

The SendOutsideWasm/SyncOutsideWasm traits are empty on WASM32 (allowing all types) and alias to Send/Sync on native targets, ensuring zero-cost abstraction.

Files updated:
- All FFI bindings (30+ trait definitions)
- Core SDK error types and type aliases
- Event handler infrastructure
- Store and crypto abstractions
- UI service filters and sorters
- Timeline and authentication modules

Signed-off-by:


Hywan (changes requested):
Thank you for the patches.

I think you might have conflict with https://github.com/matrix-org/matrix-rust-sdk/pull/5083.

I also wonder why sometimes you use `(Sync|Send)OutsideWasm`, and sometimes you add a `#[cfg]` to get 2 different implementations.

PS: It's WebAssembly, or Wasm, or `wasm32`, but it's never Web Assembly, or WASM, or WASM32 <3.

crates/matrix-sdk/src/sliding_sync/list/builder.rs
```diff
@@ -93,6 +97,15 @@ impl SlidingSyncListBuilder {
         self
     }
 
+    #[cfg(target_family = "wasm")]
+    pub fn once_built<C>(mut self, callback: C) -> Self
+    where
+        C: Fn(SlidingSyncList) -> SlidingSyncList + 'static,
```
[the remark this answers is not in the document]
jplatte: Here the `OutsideWasm` trait should actually work, I think.

crates/matrix-sdk-common/src/store_locks.rs
```diff
@@ -214,7 +227,7 @@ impl<S: BackingStore + Clone + SendOutsideWasm + 'static> CrossProcessStoreLock<
         //   operation running in a transaction.
 
         if let Some(_prev) = renew_task.take() {
-            #[cfg(not(target_arch = "wasm32"))]
+            #[cfg(not(target_family = "wasm"))]
```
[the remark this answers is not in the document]
zzorba: I'm hoping it should auto-merge them on rebase, but if it doesn't I can address those once one of the PRs has landed at main.

zzorba:
Fixed the flagged issue and rebased to address the upstream conflicts @Hywan

crates/matrix-sdk-common/src/store_locks.rs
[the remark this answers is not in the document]
Hywan: #5083 has landed in `main`

zzorba:
That's correct, matrix-sdk-ffi is my goal in the compilation target.  There are also a handful of items that current have been carved out of wasm32-unknown-unknown targets that are now usable.

Hywan:
Please resolve the CI and I'll do another review :-).

bindings/matrix-sdk-ffi/src/client.rs
```diff
@@ -167,39 +167,39 @@ impl From<PushFormat> for RumaPushFormat {
...
 
 #[matrix_sdk_ffi_macros::export(callback_interface)]
-pub trait ClientDelegate: Sync + Send {
+pub trait ClientDelegate: SyncOutsideWasm + SendOutsideWasm {
     fn did_receive_auth_error(&self, is_soft_logout: bool);
 }
 
 #[matrix_sdk_ffi_macros::export(callback_interface)]
-pub trait ClientSessionDelegate: Sync + Send {
+pub trait ClientSessionDelegate: AsyncTraitDeps {
```
jplatte: This trait doesn't use async so the supertrait looks a bit weird. Also you only used `AsyncTraitDeps` here, not for the other traits. Why is that?

zzorba: Good catch, in my fork I have made these functions async (because the keychain on react-native is async), so this was a stray change.  I've changed it back to the normal SendOutsideWasm + SyncOutsideWasm calls.

Hywan (approved):

crates/matrix-sdk-crypto/src/store/error.rs
```diff
@@ -77,8 +77,14 @@ pub enum CryptoStoreError {
 
     /// A problem with the underlying database backend
     #[error(transparent)]
+    #[cfg(not(target_family = "wasm"))]
     Backend(Box<dyn std::error::Error + Send + Sync>),
 
+    /// A problem with the underlying database backend
+    #[error(transparent)]
+    #[cfg(target_family = "wasm")]
+    Backend(Box<dyn std::error::Error>),
```
richvdh, line 86: @zzorba I'm just wondering what the driver for this particular change (removing the `Send + Sync` bounds on `Backend`) was?

(For context: it turns out we have some code in matrix-sdk-crypto-wasm that relies on `CryptoStoreError` being `Send + Sync`. Not the end of the world to work around it, but I'm interested to know why it was necessary.)

zzorba: I'll reconfirm now, but I believe the uniffi boundary can't handle the Send/Sync requirement on wasm platforms (I suspect because the base future system is not sync/sync for wasm).

Our goal is to use the uniffi annotation to generate JS/Wasm bindings. This is working in our fork, but the changes are being added slowly to the main repo.

zzorba: Hmm, I might be mistaken.  Looks like removing a second trait was able to work.  I will open a PR for this one.

zzorba: Okay, so this particular one is not needed.  

One of them is, the `UnknownError` in matrix-sdk/src/error.rs runs into anyhow, which struggles with the sync bounds.  

I will open a PR to relax the ones I can though.

PR #5085, opened 2025-05-23
StreamExt that supports wasm platforms

zzorba:
The .boxed() method requires a Send trait that makes it unusable on Wasm platforms. This re-exports the existing StreamExt, while providing a new extension for the wasm family of targets.

Break out of https://github.com/matrix-org/matrix-rust-sdk/pull/5079

Signed-off-by: Daniel Salinas


zzorba:
Updated the comments, and hooked up the extension in matrix-sdk-ffi/client.rs (at the moment this is the only call site I have found necessary to swap extensions for, but I suspect there might be one or two that come up later).

Hywan (approved):
Looks good to me, thanks!

PR #5086, opened 2025-05-23
change: have media preview config return an optional

Velin92:
The reason is that, while the `default` was quite handy, it prevents clients to perform migrations. This will allow a client to check if the server or the store do have a definition for the account data, and if not decide to perform a migration or present a different default if they so wish.

### DIFF
--- a/crates/matrix-sdk/src/account.rs
+++ b/crates/matrix-sdk/src/account.rs
@@ -995,8 +995,7 @@ impl Account {
     /// that will yield new values as they are received.
     ///
     /// The initial value is the one that was stored in the account data
-    /// when the client was started. If no value was found in either the stable
-    /// and unstable type, the default configuration will be returned.
+    /// when the client was started.
     /// and the following code is using a temporary solution until we know which
     /// Matrix version will support the stable type.
     ///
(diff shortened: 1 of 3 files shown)


crates/matrix-sdk/src/account.rs
```diff
@@ -1085,42 +1086,40 @@ impl Account {
...
     }
 
     /// Set the media previews display policy in the timeline.
     ///
     /// This will always use the unstable event until we know which Matrix
     /// version will support it.
     pub async fn set_media_previews_display_policy(&self, policy: MediaPreviews) -> Result<()> {
-        let mut media_preview_config = self.fetch_media_preview_config_event_content().await?;
+        let mut media_preview_config =
+            self.fetch_media_preview_config_event_content().await?.unwrap_or_default();
```
zecakeh, line 1122: While the description of the PR says that this would allow for a client to use its own default when there is no value, changing one of the two settings if the account data was not present will also set the default value of the other setting because of how Ruma serializes and deserializes this type. So the other setting might not match the default value used by the client.

For example:

1. The account data is not set, the clients uses as default `private` for `media_previews`, and `on` for `invite_avatars`.
2. The client calls `account.set_invite_avatars_display_policy(InviteAvatars::Off)`.
3. The method also sets the value for `media_previews` to `on` (the MSC default), which the client then receives via the observer stream.

One way to work around this would be to change Ruma's type to have both fields as `Option`s, in which case we wouldn't need to wrap the type in an `Option`.

poljar (approved):
Your commit descriptions seem a bit funny to me. `change` isn't telling us much, every patch is a change in some sort, otherwise there's no patch.

I think this looks good otherwise.

PR #5087, opened 2025-05-23
feat(ffi): expose the inviter directly on the room

stefanceriu:
… to avoid having to fetch the room preview in order to access it.

The room preview doesn't actually contain the inviter but instead retrieves it from the room invite details.

### DIFF
--- a/bindings/matrix-sdk-ffi/src/room.rs
+++ b/bindings/matrix-sdk-ffi/src/room.rs
@@ -134,6 +134,16 @@ impl Room {
         self.inner.alt_aliases().iter().map(|a| a.to_string()).collect()
     }
 
+    /// Get the user who created the invite, if any.
+    pub async fn inviter(&self) -> Result<Option<RoomMember>, ClientError> {
+        let invite_details = self.inner.invite_details().await?;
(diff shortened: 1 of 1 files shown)


bindings/matrix-sdk-ffi/src/room.rs
```diff
@@ -134,6 +134,12 @@ impl Room {
         self.inner.alt_aliases().iter().map(|a| a.to_string()).collect()
     }
 
+    /// Get the user who created the invite, if any.
+    pub async fn inviter(&self) -> Option<RoomMember> {
+        let invite_details = self.inner.invite_details().await.ok()?;
+        invite_details.inviter.and_then(|m| m.try_into().ok())
```
poljar, line 140: Don't we want to log anything in the failure cases?

poljar (approved):

bindings/matrix-sdk-ffi/src/room.rs
stefanceriu: Fair point, I went one step further and exposed the underlying errors to the final client to handle accordingly.

PR #5090, opened 2025-05-23
IndexedDB: Add initial implementation of `EventCacheStore`

mgoldenberg:
## Background

This pull request provides an initial implementation of the `EventCacheStore` and is mostly taken from the work completed in #4617, with a few additions and modifications to accommodate more recent updates to the trait.

Given that #4617 was a rather large pull request, this is part of a series of pull requests which will break that work into smaller, more manageable changes (see #4996). **_Notably, to minimize changes, this pull request omits the implementation of `EventCacheStoreMedia` and simply uses a nested `MemoryStore` for media queries for the time being._**

## Changes

This change is focused on integrating the implementation of `EventCacheStore` from #4617, while using a nested `MemoryStore` for media queries. Additionally, since the original implementation was written before the latest version of the `EventCacheStore`, I have added implementations of new functions.

There are virtually no changes to existing code, except for the addition of a new feature - `event-cache-store` - to the `matrix-sdk-indexeddb` crate. Most other changes occur in entirely new files.

## Caveats

Given that the original implementation from #4617 was written before some of the more recent additions to the `EventCacheStore` trait, I had to retrofit implementations for a number of functions which are rather inefficient. I have left them inefficient, however, in order to keep this pull request small, as the changes are a bit involved. 

I plan to improve them in a follow-up pull request.

### Finding the Last Chunk

Loading the last chunk in a list - i.e., `EventCacheStore::load_last_chunk` - currently reads in and deserializes all chunks in a room and looks for any that have no `next` value. This is a rather inefficient process and could be greatly improved by adding another index on the `LINKED_CHUNKS` object store.

### Querying Events by ID

The following functions read and deserialize all events in a room in order to perform their operations.

- `EventCacheStore::filter_duplicated_events`
- `EventCacheStore::find_event`
- `EventCacheStore::find_event_relations`
- `EventCacheStore::save_event`

These operations could also be greatly improved by the addition of a few new indices on the `EVENTS` object store.

## Future Work

- Add an index to the `LINKED_CHUNKS` object store in order to improve chunk-related queries.
- Add indices to the `EVENTS` object store in order to improve event-related queries.
- Add an `EventCacheStoreMedia` implementation in order to have persistent media storage.

---

### DIFF
--- a/crates/matrix-sdk-indexeddb/CHANGELOG.md
+++ b/crates/matrix-sdk-indexeddb/CHANGELOG.md
@@ -6,6 +6,12 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased] - ReleaseDate
 
+### Features
+
+- Add `IndexeddbEventCacheStore` for providing an IndexedDB implementation
+  of the `EventCacheStore`. Expose the type through `IndexeddbEventCacheStoreBuilder`
+  which ensures object stores in IndexedDB are properly initialized.
+
 ## [0.11.0] - 2025-04-11
 
 No notable changes in this release.
--- a/crates/matrix-sdk-indexeddb/Cargo.toml
+++ b/crates/matrix-sdk-indexeddb/Cargo.toml
@@ -14,9 +14,10 @@ default-target = "wasm32-unknown-unknown"
 rustdoc-args = ["--cfg", "docsrs", "--generate-link-to-definition"]
 
 [features]
-default = ["e2e-encryption", "state-store"]
+default = ["e2e-encryption", "state-store", "event-cache-store"]
 state-store = ["dep:matrix-sdk-base", "growable-bloom-filter"]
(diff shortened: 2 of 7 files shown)


Hywan (commented):
Oh wow, it looks great! Thanks for the hardwork.

First off, it is possible to split this PR in smaller ones? It would radically simplify the review process, both for you and for me?

mgoldenberg:
@Hywan, I think that would mostly work okay, except that the integration tests cannot be added incrementally as far as I can tell. I suppose I could copy relevant tests over one-by-one, and then when we have all the functions implemented, I could remove the copies and make a call to the macro.

Does that sound reasonable? Or is there a better way of going about that?

Hywan:
> Does that sound reasonable? Or is there a better way of going about that?

It sounds great to me!

mgoldenberg:
Closing this, as work is continuing piecemeal in #5138.

Issue #5091, opened 2025-05-24
SlidingSync: deadlock when calling room.members() from EventHandler


aeoncl:
## Context

When using native Sliding Sync, in the handle_response method of the SlidingSync impl, we take the `sync_lock` and then we call `process_and_take_response()`.

```rust
//crates/matrix-sdk/sliding_sync/mod.rs - line 259
//...
        let sync_response = {
            // Take the lock to avoid concurrent sliding syncs overwriting each other's room
            // infos.
            let _sync_lock = self.inner.client.base_client().sync_lock().lock().await;

//...

            response_processor.process_and_take_response().await?
        };
```

`process_and_take_response()` calls all the `sync_response_handlers` while the `sync_lock` is still in scope.

```rust
pub async fn process_and_take_response(mut self) -> Result<SyncResponse> {
        let mut response = self.response.take().unwrap_or_default();

        response.to_device.extend(self.to_device_events);

        self.client.call_sync_response_handlers(&response).await?;

        Ok(response)
    }
```

I've registered an AccountDataEvent Handler, which calls room.members(), 

if members are not synced, we end up in `receive_all_members()` which also takes the `sync_lock`, ending up in a deadlock

```rust
//crates/matrix-sdk-base/client.rs - line 787
    pub async fn receive_all_members(
        &self,
        room_id: &RoomId,
        request: &api::membership::get_member_events::v3::Request,
        response: &api::membership::get_member_events::v3::Response,
    ) -> Result<()> {
   // ...

        let _sync_lock = self.sync_lock().lock().await;
        let mut room_info = room.clone_info();
        room_info.mark_members_synced();
        context.state_changes.add_room(room_info);
    // ...



}
```

## Proposed fix

We could release the lock right before calling `process_and_take_response()`

PR #5092, opened 2025-05-24
5091/sliding sync deadlock

aeoncl:
Fixed Sliding Sync deadlock due to sync_lock not being released before calling the Event Handlers

(see issue #5091)

Signed-off-by: aeoncl <c.lopez.dev@protonmail.com>


aeoncl:
Thanks :D 
I added the regression test, can confirm that it breaks before the fix.

Hywan (approved):

PR #5093, opened 2025-05-24
Test: Replace `sync_timeline_event!` with `EventFactory` for events in event item tests

yousefmoazzam:
Part of #3716

I did notice that the `sender` and `member` methods have some overlap in what values get set for the "sender" and "state key", and I tried to make sure that in my changes to use `EventFactory` the original event configuration is being replicated, so please do double-check me on that note in particular.

### DIFF
--- a/crates/matrix-sdk-ui/src/timeline/event_item/mod.rs
+++ b/crates/matrix-sdk-ui/src/timeline/event_item/mod.rs
@@ -1120,9 +1120,9 @@ mod tests {
         let room_id = room_id!("!q:x.uk");
         let user_id = user_id!("@t:o.uk");
         let client = logged_in_client(None).await;
-        let f = EventFactory::new().room(room_id);
+        let f = EventFactory::new().room(room_id).sender(user_id);
 
-        let mut event = f.text_html("🤷‍♂️ No boost 🤷‍♂️", "").sender(user_id).into_event();
+        let mut event = f.text_html("🤷‍♂️ No boost 🤷‍♂️", "").into_event();
         let mut timeline_item =
             EventTimelineItem::from_latest_event(client.clone(), room_id, LatestEvent::new(event))
                 .await
@@ -1131,7 +1131,7 @@ mod tests {
         assert!(!timeline_item.contains_only_emojis());
 
         // Ignores leading and trailing white spaces
-        event = f.text_html(" 🚀 ", "").sender(user_id).into_event();
+        event = f.text_html(" 🚀 ", "").into_event();
         timeline_item =
             EventTimelineItem::from_latest_event(client.clone(), room_id, LatestEvent::new(event))
                 .await
@@ -1140,7 +1140,7 @@ mod tests {
         assert!(timeline_item.contains_only_emojis());
 
         // Too many
-        event = f.text_html("👨‍👩‍👦1️⃣🚀👳🏾‍♂️🪩👍👍🏻🫱🏼‍🫲🏾🙂👋", "").sender(user_id).into_event();
+        event = f.text_html("👨‍👩‍👦1️⃣🚀👳🏾‍♂️🪩👍👍🏻🫱🏼‍🫲🏾🙂👋", "").into_event();
         timeline_item =
             EventTimelineItem::from_latest_event(client.clone(), room_id, LatestEvent::new(event))
                 .await
@@ -1149,7 +1149,7 @@ mod tests {
         assert!(!timeline_item.contains_only_emojis());
 
         // Works with combined emojis
-        event = f.text_html("👨‍👩‍👦1️⃣👳🏾‍♂️👍🏻🫱🏼‍🫲🏾", "").sender(user_id).into_event();
+        event = f.text_html("👨‍👩‍👦1️⃣👳🏾‍♂️👍🏻🫱🏼‍🫲🏾", "").into_event();
         timeline_item =
             EventTimelineItem::from_latest_event(client.clone(), room_id, LatestEvent::new(event))
                 .await


crates/matrix-sdk-ui/src/timeline/event_item/mod.rs
```diff
@@ -1209,18 +1200,12 @@ mod tests {
...
-                "body": body,
-                "format": "org.matrix.custom.html",
-                "formatted_body": formatted_body,
-                "msgtype": "m.text"
-            },
-        }))
+        EventFactory::new()
+            .room(room_id)
+            .text_html(body, formatted_body)
+            .event_id(event_id!("$eventid6"))
```
bnjbvr, line 1206: Since we're here: can you try to drop this line, and see if the tests still pass? Maybe it was provided because it's always required to have an event id attached to an event, but now the event factory will automatically generate event ids when they were not provided (unless instructed not to do that, which is an opt-out behavior).

crates/matrix-sdk-ui/src/timeline/event_item/mod.rs
```diff
@@ -1209,18 +1200,12 @@ mod tests {
...
-            "type": "m.room.message",
-            "room_id": room_id.to_string(),
-            "content": {
-                "body": body,
-                "format": "org.matrix.custom.html",
-                "formatted_body": formatted_body,
-                "msgtype": "m.text"
-            },
-        }))
+        EventFactory::new()
```
bnjbvr, line 1203: If you're interested in bonus points :D This method `message_event` is used only in two callers, so we could actually inline it there, and reuse the same `EventFactory` in that case. I think I'd like it slightly better, because then we see immediately what are the events, as code readers, and we don't have to do an "context switch" in the reading. Of course, that's totally optional and we could do it later if you don't feel like it :)

crates/matrix-sdk-ui/src/timeline/event_item/mod.rs
```diff
@@ -1150,24 +1148,17 @@ mod tests {
...
+            .room(room_id)
+            .sender(user_id!("@example:example.org"))
+            .member(user_id)
+            .avatar_url(avatar_url.into())
+            .display_name(display_name)
+            .reason("")
+            .event_id(event_id!("$143273582443PhrSn:example.org"))
+            .server_ts(143273583)
+            .age(1234)
```
bnjbvr, line 1160: I wonder if these 3 fields are meaningful; maybe if they're not checked in the test, we could remove these 3 lines?

crates/matrix-sdk-ui/src/timeline/event_item/mod.rs
```diff
@@ -1150,24 +1148,17 @@ mod tests {
...
-            "origin_server_ts": 143273583,
-            "room_id": room_id,
-            "sender": "@example:example.org",
-            "state_key": user_id,
-            "type": "m.room.member",
-            "unsigned": {
-              "age": 1234
-            }
-        })
+        EventFactory::new()
```
bnjbvr, line 1151: Bonus points here too: this method is only used once in a single caller, could we inline it please? :)

crates/matrix-sdk-ui/src/timeline/event_item/mod.rs
```diff
@@ -1150,24 +1148,17 @@ mod tests {
...
+            .room(room_id)
+            .sender(user_id!("@example:example.org"))
+            .member(user_id)
```
bnjbvr, line 1154: That's the correct order, fwiw: the `user_id` passed to `member()` ends up being the state_key (and the sender, when it's not set, but here it's set so we're all good!). Good job!

bnjbvr (approved):
Cheers! A few suggestions for improvements, if you feel like it, otherwise we can merge and consider doing those next steps as a followup.

crates/matrix-sdk-ui/src/timeline/event_item/mod.rs
yousefmoazzam: Thanks for the feedback, this was also how I had initially understood the interplay between the two methods, so I'm glad I was following things correctly :slightly_smiling_face:

crates/matrix-sdk-ui/src/timeline/event_item/mod.rs
yousefmoazzam: Done in 55c5b97

crates/matrix-sdk-ui/src/timeline/event_item/mod.rs
yousefmoazzam: Done in 14f603a

yousefmoazzam:
> **_If you're bored with this PR_**, I'm happy to take over, as I'm a bit torn that I'm asking so much :)

Not at all, I'm enjoying getting the feedback and learning a bit more about how things could be done more optimally, I appreciate the time you're taking to guide me and give me tips! From that perspective, I'm more than happy to address the latest feedback and to keep going with the PR :slightly_smiling_face: And I'm glad that you're "asking so much", there's more opportunity to get to know things better if there's more information to work with.

However, I also know it may be a bit time-consuming and slow on your end to be walking me through things, so I'd completely understand if you'd prefer to take over and get this done more quickly.

So, I'll go ahead and address the latest feedback; if there's more feedback after that, then I'm cool to take it on, but I'll let you decide if you'd prefer to take it from there, no worries :slightly_smiling_face: 

> Thanks for bearing with us!

Haha, it's me who should be saying that to you...

crates/matrix-sdk-ui/src/timeline/event_item/mod.rs
```diff
@@ -1061,11 +1069,23 @@ mod tests {
...
+            .text_html("**My M**", "<b>My M</b>")
+            .sender(user_id)
+            .server_ts(122344)
+            .into_event();
         let client = logged_in_client(None).await;
 
         let member_event = MinimalStateEvent::Original(
-            member_event(room_id, user_id, "Alice Margatroid", "mxc://e.org/SEs")
+            EventFactory::new()
+                .room(room_id)
```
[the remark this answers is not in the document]
yousefmoazzam: Done in 0e78238

crates/matrix-sdk-ui/src/timeline/event_item/mod.rs
```diff
@@ -1016,7 +1019,12 @@ mod tests {
         use ruma::owned_mxc_uri;
         let room_id = room_id!("!q:x.uk");
         let user_id = user_id!("@t:o.uk");
-        let event = message_event(room_id, user_id, "**My M**", "<b>My M</b>", 122344);
+        let event = EventFactory::new()
+            .room(room_id)
+            .text_html("**My M**", "<b>My M</b>")
+            .sender(user_id)
+            .server_ts(122344)
```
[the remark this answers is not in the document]
yousefmoazzam: Done in cc191bc

crates/matrix-sdk-ui/src/timeline/event_item/mod.rs
```diff
@@ -1061,11 +1069,23 @@ mod tests {
...
```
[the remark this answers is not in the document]
yousefmoazzam: Done in cc191bc

crates/matrix-sdk-ui/src/timeline/event_item/mod.rs
```diff
@@ -1107,8 +1127,9 @@ mod tests {
         let room_id = room_id!("!q:x.uk");
         let user_id = user_id!("@t:o.uk");
         let client = logged_in_client(None).await;
+        let f = EventFactory::new().room(room_id);
 
-        let mut event = message_event(room_id, user_id, "🤷‍♂️ No boost 🤷‍♂️", "", 0);
+        let mut event = f.text_html("🤷‍♂️ No boost 🤷‍♂️", "").sender(user_id).server_ts(0).into_event();
```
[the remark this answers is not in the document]
yousefmoazzam: Done in 6383704

PR #5096, opened 2025-05-26
feat(ffi): remove the state store when clearing caches

### LINKED ISSUE #4908: dev option: Clearing the cache should also clear the room list
It seems that people are using the dev option to clear the cache now, in the regular builds, and that they expect us to get rid of the state store. It should be implemented in a hacky manner, as long as the callers are fine rebuilding the Client instance / restart the app.

---

Original issue: element-hq/element-x-android#4520
---
### Steps to reproduce

1. Have a room that no longer exists cached on your room list
2. Go to developer settings
3. Clear cache

#### What did you expect?
Cache to be fully cleared

#### What happened instead?
Room list is still in the same state as before with the rooms that no longer exist and cache still takes up about 400MB

### Your phone model

Google Pixel 8

### Operating system version

Android 16 (Beta 3.1)

### Application version and app store

25.03.4-nightly (3bff64b9)

### Homeserver

Synapse 1.127.1

### Will you send logs?

Yes

### Are you willing to provide a PR?

No

bnjbvr:
This is a dangerous operation, and requires that the sync service must be stopped while we're touching this. Since this is an internal method, that shouldn't be used in most clients anyways (as it usually papers over actual issues happening elsewhere, on the server for instance), I kept it simple with a scary doc comment explaining the preconditions, but we could assert them at runtime, with a little bit more effort.

Fixes #4908.
(edit by @Hywan) Closes https://github.com/element-hq/element-x-android-rageshakes/issues/5977

### DIFF
--- a/bindings/matrix-sdk-ffi/src/client.rs
+++ b/bindings/matrix-sdk-ffi/src/client.rs
@@ -1,6 +1,7 @@
 use std::{
     collections::HashMap,
     fmt::Debug,
+    path::PathBuf,
     sync::{Arc, OnceLock, RwLock},
     time::Duration,
 };
@@ -38,6 +39,7 @@ use matrix_sdk::{
     sliding_sync::Version as SdkSlidingSyncVersion,
     store::RoomLoadSettings as SdkRoomLoadSettings,
     AuthApi, AuthSession, Client as MatrixClient, SessionChange, SessionTokens,
+    STATE_STORE_DATABASE_NAME,
 };
 use matrix_sdk_ui::{
     notification_client::{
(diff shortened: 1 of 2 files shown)


bmarty:
> but we could assert them at runtime

Probably a safer approach? You should never trust an Android developer :)

bmarty:
It seems that the clear cache is removing too many data. I end up with an un-verifed session, and was not able to verify it again

bmarty (dismissed):
Please read my comment above.

Hywan (approved):
Good to me :-).

PR #5098, opened 2025-05-26
feat(widgetDriver): Add to-device encryption support

toger5:
This is based on the main branch and the commits from matrix-org:toger5/widget-to-device-without-encryption.

It adds two commits:
 - add capabilites to the element call widget for sending call encryption to device messages
 - add temp to-device encryption methods that should be part of the crypto crate. (they are in the submodule of `widget::matrix::to_device_crypto` which will be removed once the same methods are available through the matrix sdk crypto crate.

For now this pr is mostly a stop gap to work/use the sdk with the full to-device feature but is subject to change.

Signed-off-by:


poljar:
I believe this PR was superseded by https://github.com/matrix-org/matrix-rust-sdk/pull/5252.

PR #5099, opened 2025-05-26
feat(crypto): Support EncryptionInfo for olm encrypted messages when using `add_event_handler`

BillCarsonFr:
Make it possible to make a difference between a plaintext and a successfully decrypted to-device message. 
There is now a new variant of `EncryptionInfo` that will be passed to the to-device event handlers.

Based on https://github.com/matrix-org/matrix-rust-sdk/pull/5074

Signed-off-by:

### DIFF
--- a/crates/matrix-sdk-base/src/client.rs
+++ b/crates/matrix-sdk-base/src/client.rs
@@ -495,7 +495,7 @@ impl BaseClient {
         #[cfg(feature = "e2e-encryption")]
         let to_device = {
             let processors::e2ee::to_device::Output {
-                decrypted_to_device_events: to_device,
+                processed_to_device_events: to_device,
                 room_key_updates,
             } = processors::e2ee::to_device::from_sync_v2(&response, olm_machine.as_ref()).await?;
 
--- a/crates/matrix-sdk-base/src/response_processors/e2ee/to_device.rs
+++ b/crates/matrix-sdk-base/src/response_processors/e2ee/to_device.rs
@@ -14,7 +14,9 @@
 
 use std::collections::BTreeMap;
 
-use matrix_sdk_crypto::{store::RoomKeyInfo, EncryptionSyncChanges, OlmMachine};
+use matrix_sdk_crypto::{
+    store::RoomKeyInfo, types::ProcessedToDeviceEvent, EncryptionSyncChanges, OlmMachine,
+};
 use ruma::{
     api::client::sync::sync_events::{v3, v5, DeviceLists},
     events::AnyToDeviceEvent,
@@ -93,28 +95,31 @@ async fn process(
         let (events, room_key_updates) =
             olm_machine.receive_sync_changes(encryption_sync_changes).await?;
 
-        let events = events
-            .iter()
-            // TODO: There is loss of information here, after calling `to_raw` it is not
-            // possible to make the difference between a successfully decrypted event and a plain
-            // text event. This information needs to be propagated to top layer at some point if
-            // clients relies on custom encrypted to device events.
-            .map(|p| p.to_raw())
-            .collect();
-
-        Output { decrypted_to_device_events: events, room_key_updates: Some(room_key_updates) }
+        Output { processed_to_device_events: events, room_key_updates: Some(room_key_updates) }
     } else {
-        // If we have no `OlmMachine`, just return the events that were passed in.
+        // If we have no `OlmMachine`, just return the clear events that were passed in.
+        // The encrypted ones are dropped as they are un-usable.
         // This should not happen unless we forget to set things up by calling
         // `Self::activate()`.
         Output {
-            decrypted_to_device_events: encryption_sync_changes.to_device_events,
+            processed_to_device_events: encryption_sync_changes
+                .to_device_events
+                .into_iter()
+                .filter(|raw| {
+                    if let Ok(Some(event_type)) = raw.get_field::<String>("type") {
+                        event_type != "m.room.encrypted"
+                    } else {
+                        false // Exclude events with no type or encrypted
+                    }
+                })
+                .map(ProcessedToDeviceEvent::PlainText)
+                .collect(),
             room_key_updates: None,
         }
     })
 }
 
 pub struct Output {
-    pub decrypted_to_device_events: Vec<Raw<AnyToDeviceEvent>>,
+    pub processed_to_device_events: Vec<ProcessedToDeviceEvent>,
     pub room_key_updates: Option<Vec<RoomKeyInfo>>,
 }
--- a/crates/matrix-sdk-base/src/sync.rs
+++ b/crates/matrix-sdk-base/src/sync.rs
@@ -17,6 +17,7 @@
 use std::{collections::BTreeMap, fmt};
 
 use matrix_sdk_common::{debug::DebugRawEvent, deserialized_responses::TimelineEvent};
+use matrix_sdk_crypto::types::ProcessedToDeviceEvent;
 pub use ruma::api::client::sync::sync_events::v3::{
     InvitedRoom as InvitedRoomUpdate, KnockedRoom as KnockedRoomUpdate,
 };
@@ -24,7 +25,7 @@ use ruma::{
     api::client::sync::sync_events::UnreadNotificationsCount as RumaUnreadNotificationsCount,
     events::{
         presence::PresenceEvent, AnyGlobalAccountDataEvent, AnyRoomAccountDataEvent,
-        AnySyncEphemeralRoomEvent, AnySyncStateEvent, AnyToDeviceEvent,
+        AnySyncEphemeralRoomEvent, AnySyncStateEvent,
     },
     push::Action,
     serde::Raw,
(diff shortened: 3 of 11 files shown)


crates/matrix-sdk-base/src/client.rs
```diff
@@ -519,7 +519,19 @@ impl BaseClient {
...
+            .events
+            .into_iter()
+            .filter(|raw| {
+                if let Ok(Some(event_type)) = raw.get_field::<String>("type") {
+                    event_type != "m.room.encrypted"
+                } else {
+                    false // Exclude events with no type or encrypted
+                }
+            })
+            .map(matrix_sdk_common::deserialized_responses::ProcessedToDeviceEvent::PlainText)
```
BillCarsonFr: When there is no encryption, to-device are not processed by the olm machine. There is no `UnProcessed` variant of `ProcessedToDeviceEvent` (not sure it would make sense :/).
So anyhow, we can just ignore the encrypted to-device as they are useless. So we can just map to PlainText variant?
I wonder if we should at some point remove the optional `e2e-encryption` feature flag

crates/matrix-sdk/src/event_handler/mod.rs
```diff
@@ -821,6 +856,45 @@ mod tests {
...
+    #[async_test]
+    #[allow(dependency_on_unit_never_type_fallback)]
+    async fn test_add_to_device_event_handler() -> crate::Result<()> {
+        let client = logged_in_client(None).await;
+
+        let captured_event: Arc<Mutex<Option<AnyToDeviceEvent>>> = Arc::new(Mutex::new(None));
+        let captured_info: Arc<Mutex<Option<EncryptionInfo>>> = Arc::new(Mutex::new(None));
+
+        client.add_event_handler({
+            let captured = captured_event.clone();
```
andybalaam, line 868: I'd prefer this to be called `captured_event` too.

crates/matrix-sdk/src/event_handler/mod.rs
```diff
@@ -821,6 +856,45 @@ mod tests {
...
+            let captured_info = captured_info.clone();
+            move |ev: AnyToDeviceEvent, encryption_info: Option<EncryptionInfo>| {
+                let mut captured_lock = captured.lock();
+                *captured_lock = Some(ev);
```
andybalaam, line 872: Can we lock and update in the same line of code? Might be nicer?

crates/matrix-sdk/src/event_handler/mod.rs
BillCarsonFr: Thx @andybalam I updated the test https://github.com/matrix-org/matrix-rust-sdk/pull/5099/commits/041e50d7e46c4ca21de2fa73014f6ca4b9fda130

crates/matrix-sdk-base/src/client.rs
poljar: Huh, but how are events we fail to decrypt treated? Why not treat them just like that?

crates/matrix-sdk-base/src/response_processors/e2ee/to_device.rs
```diff
@@ -93,28 +94,31 @@ async fn process(
...
-            // clients relies on custom encrypted to device events.
-            .map(|p| p.to_raw())
-            .collect();
-
-        Output { decrypted_to_device_events: events, room_key_updates: Some(room_key_updates) }
+        Output { processed_to_device_events: events, room_key_updates: Some(room_key_updates) }
     } else {
-        // If we have no `OlmMachine`, just return the events that were passed in.
+        // If we have no `OlmMachine`, just return the clear events that were passed in.
+        // The encrypted ones are dropped as they are un-usable.
```
poljar, line 100: I mean sure, they are unusable in the sense that they are not decrypted. But it might be useful to know that you received an encrypted event, if for nothing else for logs.

crates/matrix-sdk-base/src/sync.rs
```diff
@@ -61,7 +64,12 @@ impl fmt::Debug for SyncResponse {
         f.debug_struct("SyncResponse")
             .field("rooms", &self.rooms)
             .field("account_data", &DebugListOfRawEventsNoId(&self.account_data))
-            .field("to_device", &DebugListOfRawEventsNoId(&self.to_device))
+            .field(
+                "to_device",
+                &DebugListOfRawEventsNoId(
+                    self.to_device.iter().map(|p| p.to_raw()).collect::<Vec<_>>().as_slice(),
```
poljar: Feels a bit expensive to clone the whole thing for a debug output, can't we have an `as_raw()` that will borrow the thing instead?

crates/matrix-sdk-common/src/deserialized_responses.rs
```diff
@@ -1147,6 +1147,43 @@ impl From<SyncTimelineEventDeserializationHelperV0> for TimelineEvent {
     }
 }
 
+/// Represents a to-device event after it has been processed by the olm machine.
```
poljar:
```suggestion
/// Represents a to-device event after it has been processed by the Olm machine.
```

crates/matrix-sdk-common/src/deserialized_responses.rs
```diff
@@ -1147,6 +1147,43 @@ impl From<SyncTimelineEventDeserializationHelperV0> for TimelineEvent {
...
+#[derive(Clone, Debug)]
+pub enum ProcessedToDeviceEvent {
+    /// A successfully-decrypted encrypted event.
+    /// Contains the raw decrypted event and encryption info
+    Decrypted {
+        /// The raw decrypted event
+        raw: Raw<AnyToDeviceEvent>,
+        /// The olm encryption info
```
poljar:
```suggestion
        /// The Olm encryption info
```

crates/matrix-sdk/src/sync.rs
```diff
@@ -74,7 +75,12 @@ impl fmt::Debug for SyncResponse {
             .field("next_batch", &self.next_batch)
             .field("rooms", &self.rooms)
             .field("account_data", &DebugListOfRawEventsNoId(&self.account_data))
-            .field("to_device", &DebugListOfRawEventsNoId(&self.to_device))
+            .field(
+                "to_device",
+                &DebugListOfRawEventsNoId(
+                    self.to_device.iter().map(|p| p.to_raw()).collect::<Vec<_>>().as_slice(),
```
poljar: Same, as for the previous place where we had this, an `as_raw()` method might be able to avoid all the cloning.

crates/matrix-sdk/src/test_utils/mocks/encryption.rs
```diff
@@ -120,6 +120,15 @@ impl MatrixMockServer {
             alice_olm.update_tracked_users([bob_user_id.as_ref()]).await.unwrap();
         }
 
+        // let bob be aware of Alice keys in order to be able to decrypt custom
+        // to-device (the device keys check are not delayed for non-crypto to-device
```
poljar: I think they are in fact only delayed or rather deferred for `m.room.key` events since dropping those would lead to a lot of UTDs.

crates/matrix-sdk-base/src/response_processors/e2ee/to_device.rs
BillCarsonFr: Right. it is better => https://github.com/matrix-org/matrix-rust-sdk/pull/5099/commits/7639807c806fce6c37c3c74b32d11109d89772bc

crates/matrix-sdk-base/src/sync.rs
BillCarsonFr: I did a similar thing to DebugListOfRawEventsNoId for list of processed to device https://github.com/matrix-org/matrix-rust-sdk/pull/5099/commits/3eaaa30e59f3d3c351de00dbdfd62470bf446766

crates/matrix-sdk/src/test_utils/mocks/encryption.rs
BillCarsonFr: I clarified (I am not sure what is exactly done for key bundles, but no need to go in details here)

crates/matrix-sdk/CHANGELOG.md
```diff
@@ -10,6 +10,9 @@ All notable changes to this project will be documented in this file.
 
 ### Features
 
+- `Client::add_event_handler`: Set `Option<EncryptionInfo>` in `EventHandlerData` for to-device messages.
+  If the to-device message was encrypted, the `EncryptionInfo` will be set. If it is `None` the message was sent in clear.
+  ([#5099](https://github.com/matrix-org/matrix-rust-sdk/pull/5099))
```
poljar: This has been put in the wrong place, it's not part of the 0.12.0 release.

poljar (changes requested):
I think this is now good to go, we just need to fix the changelog entry.

To double check, this isn't a breaking change, right? Existing event handlers will continue on working, they'll just ignore the new `EncryptionInfo` field.

crates/matrix-sdk/CHANGELOG.md
BillCarsonFr: oh right, got hit by the merge. https://github.com/matrix-org/matrix-rust-sdk/pull/5099/commits/d92eceb073ca8a3db6719b92d5acb4bf605399db

> To double check, this isn't a breaking change, right? Existing event handlers will continue on working

Yes no change for existing event handlers. 
Existing event handlers for to_device that requested the `EncryptionInfo` were always getting None, now they will get something if the event was encrypted.

poljar (approved):
Thanks, looks good now.

Should I just squash merge?

PR #5102, opened 2025-05-27
refactor(base): Improve the `dispatch_and_get_new_users` response processor

Hywan:
Okay, that's a small clean up, but a nice one. The `dispatch_and_get_new_users` response processor is renamed `dispatch`. Less allocations, one less clone, and clear code, please review each patch individually, it's rather simple and straightforward.

---

- Address https://github.com/matrix-org/matrix-rust-sdk/issues/5050


bnjbvr (approved):
lgtm!

PR #5103, opened 2025-05-27
test: Improve `NotificationClient` tests using the modern test utils

jmartinesp:

Signed-off-by:

### DIFF
--- a/crates/matrix-sdk-ui/src/notification_client.rs
+++ b/crates/matrix-sdk-ui/src/notification_client.rs
@@ -968,8 +968,12 @@ impl NotificationItem {
         if sender_display_name.is_none() || sender_avatar_url.is_none() {
             let sender_id = event.sender();
             for ev in state_events {
-                let Ok(ev) = ev.deserialize() else {
-                    continue;
+                let ev = match ev.deserialize() {
+                    Ok(ev) => ev,
+                    Err(error) => {
+                        warn!(?error, "Failed to deserialize a state event");
+                        continue;
+                    }
                 };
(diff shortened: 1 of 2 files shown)


crates/matrix-sdk-ui/src/notification_client.rs
```diff
@@ -968,8 +968,12 @@ impl NotificationItem {
...
             let sender_id = event.sender();
             for ev in state_events {
-                let Ok(ev) = ev.deserialize() else {
-                    continue;
+                let ev = match ev.deserialize() {
+                    Ok(ev) => ev,
+                    Err(error) => {
+                        warn!("Failed to deserialize state event: {:?}", error);
+                        continue;
+                    }
```
jmartinesp, line 976: I added this because figuring out what was wrong when generating the fake events was quite difficult. I can revert it, but I think it's useful.

crates/matrix-sdk-ui/tests/integration/notification_client.rs
```diff
@@ -127,7 +108,6 @@ async fn test_notification_client_with_context() {
         assert_eq!(event.event_type(), TimelineEventType::RoomMessage);
     });
     assert_eq!(item.sender_display_name.as_deref(), Some("John Mastodon"));
-    assert_eq!(item.sender_avatar_url.as_deref(), Some("https://example.org/avatar.jpeg"));
```
jmartinesp, line 130: I had to remove the avatar comparison because the `EventBuilder` expects an `MxcUri` value, but returns a `String`, and I couldn't find a good way to compare them.

PR #5105, opened 2025-05-27
fix(sdk): Handle `520` HTTP status code as a permanent error

jmartinesp:
The status code is usually returned by Cloudflare to indicate an unknown server error, so we should cancel the upload and let the user retry if they want to.

Signed-off-by:


poljar (approved):
This might be changelog worthy. Looks good otherwise.

PR #5116, opened 2025-05-28
feat(widget): Receive custom to-device messages in widgets in e2ee rooms

BillCarsonFr:
Proper support for receiving to-device messages for widgets

If the widget is in an e2ee room, clear to-device traffic will be excluded.
Also filter out internal to-device messages that widgets should not be aware off

Draft for now because depends on 
 - https://github.com/matrix-org/matrix-rust-sdk/pull/5099
 - and https://github.com/matrix-org/matrix-rust-sdk/pull/5074

I want to add more test, but for that I'd like to have this [PR for new common mock helpers](https://github.com/matrix-org/matrix-rust-sdk/pull/5005) to be merged.

Signed-off-by:

### DIFF
--- a/crates/matrix-sdk-crypto/src/machine/mod.rs
+++ b/crates/matrix-sdk-crypto/src/machine/mod.rs
@@ -1440,6 +1440,11 @@ impl OlmMachine {
                     Ok(event) => {
                         self.handle_to_device_event(changes, &event).await;
 
+                        // TODO: we should have access to some decryption settings here
+                        // (TrustRequirement) and use it (at least for
+                        // custom to-devices) to manually reject the decryption.
+                        // Similar to check_sender_trust_requirement for room events
+
                         raw_event = event
                             .serialize_zeroized()
                             .expect("Zeroizing and reserializing our events should always work")
--- a/crates/matrix-sdk/src/widget/matrix.rs
+++ b/crates/matrix-sdk/src/widget/matrix.rs
@@ -240,7 +240,6 @@ impl MatrixDriver {
                     return;
                 };
 
-                // Or should I use room.encryption_settings()?
                 let room_encrypted = room.latest_encryption_state().await
                     .map(|s| s.is_encrypted())
                     // Default consider encrypted
--- a/crates/matrix-sdk/src/widget/matrix.rs
+++ b/crates/matrix-sdk/src/widget/matrix.rs
@@ -38,10 +38,12 @@ use ruma::{
 };
 use serde_json::{value::RawValue as RawJsonValue, Value};
 use tokio::sync::mpsc::{unbounded_channel, UnboundedReceiver};
-use tracing::error;
+use tracing::{error, trace, warn};
 
 use super::{machine::SendEventResponse, StateKeySelector};
-use crate::{event_handler::EventHandlerDropGuard, room::MessagesOptions, Error, Result, Room};
+use crate::{
+    event_handler::EventHandlerDropGuard, room::MessagesOptions, Client, Error, Result, Room,
+};
 
 /// Thin wrapper around a [`Room`] that provides functionality relevant for
 /// widgets.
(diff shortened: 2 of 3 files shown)


crates/matrix-sdk-crypto/src/machine/mod.rs
```diff
@@ -1444,6 +1444,11 @@ impl OlmMachine {
                     Ok(event) => {
                         self.handle_to_device_event(changes, &event).await;
 
+                        // TODO: we should have access to some decryption settings here
+                        // (TrustRequirement) and use it (at least for
+                        // custom to-devices) to manually reject the decryption.
+                        // Similar to check_sender_trust_requirement for room events
```
poljar: That's probably not the place where we would reject the decryption as we're already dealing with decrypted events here.

A better place for this would be the `parse_decrypted_to_device_event()` method.

crates/matrix-sdk/src/widget/matrix.rs
```diff
@@ -235,21 +235,86 @@ impl MatrixDriver {
...
 
+    fn should_filter_message_to_widget(raw_message: &Raw<AnyToDeviceEvent>) -> bool {
+        let Ok(Some(event_type)) = raw_message.get_field::<String>("type") else {
+            return true;
+        };
+
+        if event_type == "m.room.encrypted" {
+            // Unable to decrypt,
+            return true;
+        }
```
poljar: Why treat this as a special case? This would fit equally well into the `matches!()` down bellow.

crates/matrix-sdk/src/widget/matrix.rs
```diff
@@ -235,21 +235,86 @@ impl MatrixDriver {
...
-            move |raw: Raw<AnyToDeviceEvent>, _: Option<EncryptionInfo>| {
-                let _ = tx.send(raw);
-                async {}
+
+            async move |raw: Raw<AnyToDeviceEvent>, encryption_info: Option<EncryptionInfo>, client: Client| {
+
+                // Some to-device traffic is used by the sdk for internal machinery.
+                // They should not be exposed to widgets.
+                if Self::should_filter_message_to_widget(&raw) {
+                    trace!("Internal or UTD to-device message filtered out by widget driver.");
```
poljar: You are not attaching any data to this log line, it will be hard to know what got filtered out. Consider moving this into the `should_filter_message_...()` method so we can at least log the event type.

crates/matrix-sdk/src/widget/matrix.rs
```diff
@@ -235,21 +235,86 @@ impl MatrixDriver {
...
-            // EncryptionInfo. The widgetAPI expects a boolean `encrypted` to be added
-            // (!) to the raw content to know if the to-device message was encrypted or
-            // not (as per MSC3819).
-            move |raw: Raw<AnyToDeviceEvent>, _: Option<EncryptionInfo>| {
-                let _ = tx.send(raw);
-                async {}
+
+            async move |raw: Raw<AnyToDeviceEvent>, encryption_info: Option<EncryptionInfo>, client: Client| {
+
+                // Some to-device traffic is used by the sdk for internal machinery.
```
poljar:
```suggestion
                // Some to-device traffic is used by the SDK for internal machinery.
```

crates/matrix-sdk/src/widget/matrix.rs
```diff
@@ -235,21 +235,86 @@ impl MatrixDriver {
...
+                // They should not be exposed to widgets.
+                if Self::should_filter_message_to_widget(&raw) {
+                    trace!("Internal or UTD to-device message filtered out by widget driver.");
+                    return;
+                }
+
+                // Encryption can be enabled after the widget has been instantiated,
+                // we want to keep track of the latest status
+                let Some(room) = client.get_room(&room_id) else {
+                    warn!("Room {} not found in client.", room_id);
```
poljar:
```suggestion
                    warn!("Room {room_id} not found in client.");
```

crates/matrix-sdk/src/widget/matrix.rs
```diff
@@ -235,21 +235,86 @@ impl MatrixDriver {
...
+                    return;
+                };
+
+                let room_encrypted = room.latest_encryption_state().await
+                    .map(|s| s.is_encrypted())
+                    // Default consider encrypted
+                    .unwrap_or(true);
```
poljar, line 259: Nice, we're defaulting to encryption on, this is probably one of the first places where we do so.

crates/matrix-sdk/src/widget/matrix.rs
```diff
@@ -235,21 +235,86 @@ impl MatrixDriver {
...
+                    if encryption_info.is_none() {
+                        warn!(
+                            ?room_id,
+                            "Received to-device event in clear for a widget in an e2e room, dropping."
+                        );
+                        return;
+                    };
+
+                    // There no per-room specific decryption setting, so we can just send to the
+                    // widget
```
poljar:
```suggestion
                    // There are no per-room specific decryption setting, so we can just send to the
                    // widget
```

crates/matrix-sdk/src/test_utils/mocks/encryption.rs
```diff
@@ -178,6 +182,70 @@ impl MatrixMockServer {
...
+    ///
+    /// * `sender` - The user ID of the message sender
+    ///
+    /// # Returns
+    ///
+    /// Returns a tuple containing:
+    /// - A `MockGuard` the end-point mock is scoped to this guard
+    /// - A `Future` that resolves to a `Value` containing the captured
+    ///   encrypted to-device message.
+    pub async fn mock_capture_put_to_device(
```
poljar, line 251: That's a neat function, could we add an example to the doc so:

1. The functions i kind off tested as a doctest
2. Other people will more easily discover how to use it

crates/matrix-sdk/src/test_utils/mocks/encryption.rs
```diff
@@ -178,6 +182,70 @@ impl MatrixMockServer {
...
+    ///
+    /// # Arguments
+    ///
+    /// * `sender` - The user ID of the message sender
+    ///
+    /// # Returns
+    ///
+    /// Returns a tuple containing:
+    /// - A `MockGuard` the end-point mock is scoped to this guard
+    /// - A `Future` that resolves to a `Value` containing the captured
```
poljar: Does a `Value` make the most sense? Couldn't we return `EncryptedToDeviceEvent` or `Raw<EncryptedToDeviceEvent>`?

This would ensure that nobody misses that this is only for `m.room.encrypted` to-device messages.

edit: Looking at the usage of this, you pass it directly to a mocked sync, so a `Value` fits, though I think `Raw<T>` would work as well. Additionally it might make sense to add a function that does this capture and mocks the sync in once call.

crates/matrix-sdk/tests/integration/widget.rs
```diff
@@ -524,9 +567,176 @@ async fn test_receive_live_events() {
...
+    assert_eq!(data["sender"], "@bob:example.org");
+    // TODO: Currently the rust-sdk is exposing more than type/content/sender in
+    // the event this should be exposed to the widget but it should be fixed
+    // at the crypto crate level see https://github.com/matrix-org/matrix-rust-sdk/pull/5201
+    // assert_eq!(data.len(), 3);
+}
+
+/// Test that "internal" to-device messages are never forwarded to the widgets.
+/// Here we test that a `m.room_key` message is never forwarded to the widget.
+/// These events are already zeroized by the sdk, but let's not leak them
```
poljar:
```suggestion
/// These events are already zeroized by the SDK, but let's not leak them
```

crates/matrix-sdk-crypto/src/machine/mod.rs
BillCarsonFr: right, moved it to the correct place https://github.com/matrix-org/matrix-rust-sdk/pull/5116/commits/618bf31e1ca964c36212fe8ed010f3eeae921b91

crates/matrix-sdk/src/widget/matrix.rs
BillCarsonFr: moved it, and added better logs

crates/matrix-sdk/src/test_utils/mocks/encryption.rs
BillCarsonFr: I did that and I also added a utility to combine capturing and sending back to sync https://github.com/matrix-org/matrix-rust-sdk/pull/5116/commits/51523e13f346981125efdd60fde3d8aceb33cf2e let me know if it is the right way

crates/matrix-sdk/src/widget/matrix.rs
```diff
@@ -235,21 +235,91 @@ impl MatrixDriver {
...
+                | "m.key.verification.done"
+                | "m.secret.request"
+                | "m.secret.send"
+                // drop utd traffic
+                | "m.room.encrypted"
+        );
+
+        if filtered {
+            trace!(
+                "To-device message of type <{}> filtered out by widget driver.",
```
poljar: This works as well, right?

```suggestion
                "To-device message of type <{event_type}> filtered out by widget driver.",
```

poljar (changes requested):
This looks good, left still one nit.

If the nit doesn't make sense let me know and we can merge after that.

crates/matrix-sdk/src/widget/matrix.rs
BillCarsonFr: yes that works, missed it

PR #5118, opened 2025-05-28
Make the sharing and accepting of historic room key bundles more robust

poljar:
These two patches ensure that we:

1. Have all the devices of the recipient of the room key bundle before we send it out.
2. Accept a room key bundle from a device even if the device which has sent us the bundle isn't found in our local store.

The second point depends on [MSC4147](https://github.com/matrix-org/matrix-spec-proposals/pull/4147).

### DIFF
--- a/crates/matrix-sdk-crypto/src/olm/account.rs
+++ b/crates/matrix-sdk-crypto/src/olm/account.rs
@@ -1465,34 +1465,35 @@ impl Account {
             .into())
         } else {
             // If the event contained sender_device_keys, check them now.
+            // WARN: If you move or modify this check, ensure that the code below is still
+            // valid. The processing of the historic room key bundle depends on this being
+            // here.
             Self::check_sender_device_keys(event.as_ref(), sender_key)?;
 
-            // If this event is an `m.room_key` event, defer the check for the
-            // Ed25519 key of the sender until we decrypt room events. This
-            // ensures that we receive the room key even if we don't have access
-            // to the device.
-            if !matches!(event.as_ref(), AnyDecryptedOlmEvent::RoomKey(_)) {
-                let device = if let Some(device) =
-                    store.get_device_from_curve_key(event.sender(), sender_key).await?
-                {
-                    device
-                } else if let AnyDecryptedOlmEvent::RoomKeyBundle(_) = event.as_ref() {
-                    // If this is a room key bundle, and we don't have the device in our store,
-                    // we're requiring the device keys to be part of the `AnyDecryptedOlmEvent`.
-                    //
-                    // Othrwise we'll throw an error refusing to decrypt the room key bundle.
-                    let device_keys =
-                        event.sender_device_keys().ok_or(EventError::MissingSigningKey)?;
-                    let device_data = DeviceData::try_from(device_keys).unwrap();
+            if let AnyDecryptedOlmEvent::RoomKey(_) = event.as_ref() {
+                // If this event is an `m.room_key` event, defer the check for
+                // the Ed25519 key of the sender until we decrypt room events.
+                // This ensures that we receive the room key even if we don't
+                // have access to the device.
+            } else if let AnyDecryptedOlmEvent::RoomKeyBundle(_) = event.as_ref() {
+                // If this is a room key bundle we're requiring the device keys to be part of
+                // the `AnyDecryptedOlmEvent`. This ensures that we can skip the check for the
+                // Ed25519 key below since `Self::check_sender_device_keys` already did so.
+                //
+                // If the event didn't contain any sender device keys we'll throw an error
+                // refusing to decrypt the room key bundle.
+                event.sender_device_keys().ok_or(EventError::MissingSigningKey).inspect_err(
+                    |_| {
+                        warn!("The room key bundle was missing the sender device keys in the event")
+                    },
+                )?;
+            } else {
+                let device = store
+                    .get_device_from_curve_key(event.sender(), sender_key)
+                    .await?
+                    .ok_or(EventError::MissingSigningKey)?;
 
-                    store.wrap_device_data(device_data).await?
-                } else {
-                    return Err(EventError::MissingSigningKey.into());
-                };
-
-                let Some(key) = device.ed25519_key() else {
-                    return Err(EventError::MissingSigningKey.into());
-                };
+                let key = device.ed25519_key().ok_or(EventError::MissingSigningKey)?;
 
                 if key != event.keys().ed25519 {
                     return Err(EventError::MismatchedKeys(
--- a/crates/matrix-sdk-crypto/src/olm/account.rs
+++ b/crates/matrix-sdk-crypto/src/olm/account.rs
@@ -1471,10 +1471,22 @@ impl Account {
             // Ed25519 key of the sender until we decrypt room events. This
             // ensures that we receive the room key even if we don't have access
             // to the device.
-            if !matches!(*event, AnyDecryptedOlmEvent::RoomKey(_)) {
-                let Some(device) =
+            if !matches!(event.as_ref(), AnyDecryptedOlmEvent::RoomKey(_)) {
+                let device = if let Some(device) =
                     store.get_device_from_curve_key(event.sender(), sender_key).await?
-                else {
+                {
+                    device
+                } else if let AnyDecryptedOlmEvent::RoomKeyBundle(_) = event.as_ref() {
+                    // If this is a room key bundle, and we don't have the device in our store,
+                    // we're requiring the device keys to be part of the `AnyDecryptedOlmEvent`.
+                    //
+                    // Othrwise we'll throw an error refusing to decrypt the room key bundle.
+                    let device_keys =
+                        event.sender_device_keys().ok_or(EventError::MissingSigningKey)?;
+                    let device_data = DeviceData::try_from(device_keys).unwrap();
+
+                    store.wrap_device_data(device_data).await?
+                } else {
                     return Err(EventError::MissingSigningKey.into());
                 };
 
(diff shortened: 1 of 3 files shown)


crates/matrix-sdk-crypto/src/olm/account.rs
```diff
@@ -1471,10 +1471,22 @@ impl Account {
...
+                } else if let AnyDecryptedOlmEvent::RoomKeyBundle(_) = event.as_ref() {
+                    // If this is a room key bundle, and we don't have the device in our store,
+                    // we're requiring the device keys to be part of the `AnyDecryptedOlmEvent`.
+                    //
+                    // Othrwise we'll throw an error refusing to decrypt the room key bundle.
+                    let device_keys =
+                        event.sender_device_keys().ok_or(EventError::MissingSigningKey)?;
+                    let device_data = DeviceData::try_from(device_keys).unwrap();
+
+                    store.wrap_device_data(device_data).await?
```
richvdh, line 1488: I think/hope/believe that `Self::check_sender_device_keys` has already done the check on the Ed25519 key, so there's no point in wrapping the `sender_device_keys` as a Device and doing it again.

Also, the level of indentation is getting painful here.

Could we write this as:

```rust
if let AnyDecryptedOlmEvent::RoomKeyBundle(_) = event.as_ref() {
    // We require sender device keys for room key bundle events.
    event.sender_device_keys().ok_or(EventError::MissingSigningKey)?;
} else if matches!(event.as_ref(), AnyDecryptedOlmEvent::RoomKey(_)) {
   // no-op
} else {
   // ... existing code
}
```

... but with more comments?

(why are we using different syntaxes for matching `RoomKey` and `RoomKeyBundle`? Let's clean that up too, if possible)

crates/matrix-sdk/src/room/shared_room_history.rs
```diff
@@ -66,10 +66,17 @@ pub async fn share_room_history(room: &Room, user_id: OwnedUserId) -> Result<()>
         "Uploaded encrypted key blob"
     );
 
-    // 3. Establish Olm sessions with all of the recipient's devices.
+    // 3. Ensure that we get a fresh list of devices for the invited user.
+    let (req_id, request) = olm_machine.query_keys_for_users(iter::once(user_id.as_ref()));
+
+    if !request.device_keys.is_empty() {
+        room.client.keys_query(&req_id, request.device_keys).await?;
+    }
```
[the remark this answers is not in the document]
poljar: I guess we can check if the user is tracked and decide based on that.

crates/matrix-sdk-crypto/src/olm/account.rs
poljar: Hmm, but this is not quite the same. In your snippet we unconditionally require `sender_device_keys` for room key bundles, while in my patch we only require them if no device is found in the store. That is, we fall back to the `sender_device_keys`.

I guess that's fine?

Though if we do the fallback to `sender_device_keys` method, we'll be able to more easily migrate to this logic for all other to-device message types.

richvdh: I think if we're going to require `sender_device_keys` sometimes, we might as well require it all the time. We don't really want the sender trying to second-guess whether to include them. 

Further, if there's a future client impl that *doesn't* put `sender_device_keys` in the key bundle event, and we only require them if we don't know about the device, then that's going to work 99% of the time, which is the worst possible situation for debugging.

poljar: Alright, that's a fair point.

poljar: Ok, refactored all of that in https://github.com/matrix-org/matrix-rust-sdk/pull/5118/commits/22b7377626fc8980ca37ee77f23286e85e95e3c1.

crates/matrix-sdk-crypto/CHANGELOG.md
```diff
@@ -8,6 +8,10 @@ All notable changes to this project will be documented in this file.
 
 ### Features
 
+- Add `OlmMachine::is_user_tracked_and_up_to_date`. This method checks whether
+  a given user is being tracked for device changes and whether a `/keys/query`
+  request is needed to refresh the user's device list.
```
richvdh, line 13:
```suggestion
- Add `OlmMachine::is_user_tracked_and_up_to_date`. This method checks whether
  a given user is being tracked for device changes and whether a `/keys/query`
  request is needed to refresh the user's device list.
  ([#5118](https://github.com/matrix-org/matrix-rust-sdk/pull/5118))
```

crates/matrix-sdk-crypto/src/identities/manager.rs
```diff
@@ -1564,6 +1565,10 @@ pub(crate) mod tests {
         let key_query_manager = manager.key_query_manager.synced(&cache).await.unwrap();
 
         assert!(key_query_manager.tracked_users().is_empty(), "No users are initially tracked");
+        assert!(
+            !key_query_manager.is_user_tracked_and_up_to_date(alice).await,
+            "Alice sholdn't be tracked yet"
```
richvdh, line 1570:
```suggestion
            "Alice shouldn't be tracked yet"
```

crates/matrix-sdk-crypto/src/identities/manager.rs
```diff
@@ -1578,6 +1583,43 @@ pub(crate) mod tests {
...
+
+        key_query_manager
+            .mark_tracked_users_as_up_to_date(users, SequenceNumber::default())
+            .await
+            .unwrap();
+
+        assert!(key_query_manager.tracked_users().contains(bob), "Bob should still be tracked");
+        assert!(
+            !key_query_manager.users_for_key_query().await.0.contains(bob),
+            "Bob should'nt be queued up for a /keys/query anymore"
```
richvdh, line 1615:
```suggestion
            "Bob shouldn't be queued up for a /keys/query any more"
```

(though apparently the anymore/any more thing might be a British vs American English thing. Whatever. "should'nt" is definitely wrong though :) )

crates/matrix-sdk-crypto/src/store/mod.rs
```diff
@@ -342,6 +342,21 @@ impl SyncedKeyQueryManager<'_> {
...
+
+        if is_tracked {
+            // If we're tracked, we want to know if we're queued up for a `/keys/query`.
+            //
+            // Being in this map means that we're not up-to-date.
+            !self.manager.users_for_key_query.lock().await.contains_user(user_id)
+        } else {
+            false
+        }
+    }
```
richvdh, line 358: I'm sorry about this, because I suspect it's going to make this scope-creep even more, and maybe just doing a `/keys/query` every time was the lesser evil, but...

I don't think this is sufficient. If I *previously* shared a room with Bob, but have now left, then Bob will still be present in `tracked_users` (I think?), but the device list may well be outdated.

crates/matrix-sdk-crypto/src/olm/account.rs
```diff
@@ -1465,34 +1465,35 @@ impl Account {
             .into())
         } else {
             // If the event contained sender_device_keys, check them now.
+            // WARN: If you move or modify this check, ensure that the code bellow is still
```
richvdh, line 1468:
```suggestion
            // WARN: If you move or modify this check, ensure that the code below is still
```

crates/matrix-sdk-crypto/src/olm/account.rs
```diff
@@ -1465,22 +1465,35 @@ impl Account {
...
+                // This ensures that we receive the room key even if we don't
+                // have access to the device.
+            } else if let AnyDecryptedOlmEvent::RoomKeyBundle(_) = event.as_ref() {
+                // If this is a room key bundle we're requiring the device keys to be part of
+                // the `AnyDecryptedOlmEvent`. This ensures that we can skip the check for the
+                // Ed25519 key bellow since `Self::check_sender_device_keys` already did so.
+                //
+                // If the event didn't contain any sender device keys we'll throw an error
+                // refusing to decrypt the room key bundle.
+                event.sender_device_keys().ok_or(EventError::MissingSigningKey).inspect_err(
```
richvdh, line 1485: TIL `inspect_err` (and `inspect`)

crates/matrix-sdk-crypto/src/olm/account.rs
```diff
@@ -1465,22 +1465,35 @@ impl Account {
...
-                };
+            if let AnyDecryptedOlmEvent::RoomKey(_) = event.as_ref() {
+                // If this event is an `m.room_key` event, defer the check for
+                // the Ed25519 key of the sender until we decrypt room events.
+                // This ensures that we receive the room key even if we don't
+                // have access to the device.
+            } else if let AnyDecryptedOlmEvent::RoomKeyBundle(_) = event.as_ref() {
+                // If this is a room key bundle we're requiring the device keys to be part of
+                // the `AnyDecryptedOlmEvent`. This ensures that we can skip the check for the
+                // Ed25519 key bellow since `Self::check_sender_device_keys` already did so.
```
richvdh, line 1481:
```suggestion
                // Ed25519 key below since `Self::check_sender_device_keys` already did so.
```

crates/matrix-sdk-crypto/src/store/mod.rs
richvdh: yeah, sorry, trying to optimise away the `/keys/query` was a terrible idea.

Turns out cache invalidation is hard. Who knew.

crates/matrix-sdk-crypto/CHANGELOG.md
poljar: This got reverted.

crates/matrix-sdk-crypto/src/identities/manager.rs
poljar: Reverted, so doesn't apply anymore.

crates/matrix-sdk-crypto/src/identities/manager.rs
poljar: Reverted.

crates/matrix-sdk-crypto/src/olm/account.rs
poljar: Yeah, they come in handy sometimes. Avoids some unnecessary branches in some simpler cases.

PR #5119, opened 2025-05-28
fix(sdk): check max_upload_size before attempting to upload any media

jmartinesp:
## Changes:

- Add `ClientInner::server_max_upload_size` field to hold the max upload size coming from the media config endpoint response.
- Add `Client::load_or_fetch_max_upload_size()` fn that will either load the value from the cached `OnceCell` or fetch it from the homeserver using the media config endpoint, then set it in the cell. This means the value is only valid for the current Client and will have to be re-fetched when the client is restarted.
- Create `SendMediaUploadRequest` wrapper for media upload requests, which will check the size of the file to upload and cancel the upload with an error if the file is larger than the allowed size.
- Expose `ffi::Client::get_max_upload_size()` fn to get this value so we can use it when compressing media in the platform side to check if further compression is needed before trying to upload anything.

Also fixed a few tests related to media upload since they now need to mock the media config endpoint to work.

Signed-off-by:

### DIFF
--- a/bindings/matrix-sdk-ffi/src/client.rs
+++ b/bindings/matrix-sdk-ffi/src/client.rs
@@ -1553,6 +1553,13 @@ impl Client {
     ) -> Result<Option<MediaPreviewConfig>, ClientError> {
         Ok(self.inner.account().fetch_media_preview_config_event_content().await?.map(Into::into))
     }
+
+    /// Gets the `max_upload_size` value from the homeserver, which controls the
+    /// max size a media upload request can have.
+    pub async fn get_max_media_upload_size(&self) -> Result<u64, ClientError> {
+        let max_upload_size = self.inner.load_or_fetch_max_upload_size().await?;
+        Ok(max_upload_size.into())
+    }
 }
 
 #[matrix_sdk_ffi_macros::export(callback_interface)]
--- a/crates/matrix-sdk/src/client/futures.rs
+++ b/crates/matrix-sdk/src/client/futures.rs
@@ -48,6 +48,20 @@ pub struct SendRequest<R> {
 }
 
 impl<R> SendRequest<R> {
+    /// Replace the default `SharedObservable` used for tracking upload
+    /// progress.
+    ///
+    /// Note that any subscribers obtained from
+    /// [`subscribe_to_send_progress`][Self::subscribe_to_send_progress]
+    /// will be invalidated by this.
+    pub fn with_send_progress_observable(
+        mut self,
+        send_progress: SharedObservable<TransmissionProgress>,
+    ) -> Self {
+        self.send_progress = send_progress;
+        self
+    }
+
     /// Use the given [`RequestConfig`] for this send request, instead of the
     /// one provided by default.
     pub fn with_request_config(mut self, request_config: impl Into<Option<RequestConfig>>) -> Self {
@@ -165,7 +179,7 @@ impl SendMediaUploadRequest {
         mut self,
         send_progress: SharedObservable<TransmissionProgress>,
     ) -> Self {
-        self.send_request.send_progress = send_progress;
+        self.send_request = self.send_request.with_send_progress_observable(send_progress);
         self
     }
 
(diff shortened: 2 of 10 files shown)


crates/matrix-sdk/src/client/futures.rs
```diff
@@ -42,33 +48,12 @@ pub struct SendRequest<R> {
 }
 
 impl<R> SendRequest<R> {
-    /// Replace the default `SharedObservable` used for tracking upload
-    /// progress.
-    ///
-    /// Note that any subscribers obtained from
-    /// [`subscribe_to_send_progress`][Self::subscribe_to_send_progress]
-    /// will be invalidated by this.
-    pub fn with_send_progress_observable(
```
Hywan: I would keep this method here, as it can be useful for something else than a media.

crates/matrix-sdk/src/client/futures.rs
```diff
@@ -42,33 +48,12 @@ pub struct SendRequest<R> {
...
     /// one provided by default.
     pub fn with_request_config(mut self, request_config: impl Into<Option<RequestConfig>>) -> Self {
         self.config = request_config.into();
         self
     }
-
-    /// Get a subscriber to observe the progress of sending the request
-    /// body.
-    #[cfg(not(target_arch = "wasm32"))]
-    pub fn subscribe_to_send_progress(&self) -> Subscriber<TransmissionProgress> {
```
Hywan: Same here, please keep it.

crates/matrix-sdk/src/client/futures.rs
```diff
@@ -149,3 +134,62 @@ where
...
+    /// progress.
+    ///
+    /// Note that any subscribers obtained from
+    /// [`subscribe_to_send_progress`][Self::subscribe_to_send_progress]
+    /// will be invalidated by this.
+    pub fn with_send_progress_observable(
+        mut self,
+        send_progress: SharedObservable<TransmissionProgress>,
+    ) -> Self {
+        self.send_request.send_progress = send_progress;
```
Hywan: Then, this code would become `self.send_request = self.send_request.with_send_progress_observable(send_progress)`.

crates/matrix-sdk/src/client/mod.rs
```diff
@@ -2546,6 +2552,27 @@ impl Client {
     pub async fn is_user_ignored(&self, user_id: &UserId) -> bool {
         self.base_client().is_user_ignored(user_id).await
     }
+
+    /// Gets the `max_upload_size` value from the homeserver, getting either a
+    /// cached value or with a `/_matrix/client/v1/media/config` request if it's
+    /// missing.
```
Hywan, line 2570: Can you add a link to the spec please? Would be adorable.

bindings/matrix-sdk-ffi/src/client.rs
```diff
@@ -1552,6 +1552,13 @@ impl Client {
     ) -> Result<Option<MediaPreviewConfig>, ClientError> {
         Ok(self.inner.account().fetch_media_preview_config_event_content().await?.map(Into::into))
     }
+
+    /// Gets the `max_upload_size` value from the homeserver, which controls the
+    /// max size a media upload request can have.
+    pub async fn get_max_media_upload_size(&self) -> Result<u64, ClientError> {
```
Hywan, line 1559: No need for the `get_` prefix in Rust. You can rename this method to `max_media_upload_size`.

crates/matrix-sdk/src/test_utils/mocks/mod.rs
```diff
@@ -3027,4 +3027,11 @@ impl<'a> MockEndpoint<'a, AuthenticatedMediaConfigEndpoint> {
             "m.upload.size": max_upload_size,
         })))
     }
+
+    /// Returns a successful response with a maxed out max upload size.
+    pub fn ok_default(self) -> MatrixMock<'a> {
+        self.respond_with(ResponseTemplate::new(200).set_body_json(json!({
+            "m.upload.size": UInt::MAX,
```
Hywan, line 3031: Poor server :'-).

Hywan (approved):
Approving ahead of the feedback fix, which are tiny. Thanks for the PR!

bindings/matrix-sdk-ffi/src/client.rs
jmartinesp: It is fairly used in the FFI clients that will call this though.

PR #5124, opened 2025-05-28
refactor: When mapping FFI event contents, log unsupported event types

jmartinesp:
At the moment, the error just says `Unsupported Event Type`, which is not that helpful.

Signed-off-by:


stefanceriu (approved):
TIL about `bail!`, lgtm 👍

PR #5125, opened 2025-05-28
feat(timeline): Expose method to send galleries in matrix-sdk-ui

Johennes:
This was broken out of https://github.com/matrix-org/matrix-rust-sdk/pull/4838 and is a step towards implementing https://github.com/matrix-org/matrix-spec-proposals/pull/4274. Building upon https://github.com/matrix-org/matrix-rust-sdk/pull/4977, a new method `Timeline::send_gallery` in matrix-sdk-ui for sending galleries.

Signed-off-by:

### DIFF
--- a/crates/matrix-sdk-ui/src/timeline/controller/mod.rs
+++ b/crates/matrix-sdk-ui/src/timeline/controller/mod.rs
@@ -237,6 +237,7 @@ pub fn default_event_filter(event: &AnySyncTimelineEvent, room_version: &RoomVer
                                 MessageType::Audio(_)
                                     | MessageType::Emote(_)
                                     | MessageType::File(_)
+                                    | MessageType::Gallery(_)
                                     | MessageType::Image(_)
                                     | MessageType::Location(_)
                                     | MessageType::Notice(_)
--- a/crates/matrix-sdk-ui/src/timeline/futures.rs
+++ b/crates/matrix-sdk-ui/src/timeline/futures.rs
@@ -1,6 +1,8 @@
 use std::future::IntoFuture;
 
 use eyeball::SharedObservable;
+#[cfg(feature = "unstable-msc4274")]
+use matrix_sdk::attachment::GalleryConfig;
 use matrix_sdk::{attachment::AttachmentConfig, TransmissionProgress};
 use matrix_sdk_base::boxed_into_future;
 use mime::Mime;
@@ -85,6 +87,40 @@ impl<'a> IntoFuture for SendAttachment<'a> {
                     .store_in_cache();
                 fut.await.map_err(|_| Error::FailedSendingAttachment)?;
             }
+
+            Ok(())
+        };
+
+        Box::pin(fut.instrument(tracing_span))
+    }
+}
+
+#[cfg(feature = "unstable-msc4274")]
+pub struct SendGallery<'a> {
+    timeline: &'a Timeline,
+    gallery: GalleryConfig,
+    tracing_span: Span,
+}
+
+#[cfg(feature = "unstable-msc4274")]
+impl<'a> SendGallery<'a> {
+    pub(crate) fn new(timeline: &'a Timeline, gallery: GalleryConfig) -> Self {
+        Self { timeline, gallery, tracing_span: Span::current() }
+    }
+}
+
+#[cfg(feature = "unstable-msc4274")]
+impl<'a> IntoFuture for SendGallery<'a> {
+    type Output = Result<(), Error>;
+    boxed_into_future!(extra_bounds: 'a);
+
+    fn into_future(self) -> Self::IntoFuture {
+        let Self { timeline, gallery, tracing_span } = self;
+
+        let fut = async move {
+            let send_queue = timeline.room().send_queue();
+            let fut = send_queue.send_gallery(gallery);
+            fut.await.map_err(|_| Error::FailedSendingAttachment)?;
 
             Ok(())
         };
--- a/crates/matrix-sdk-ui/tests/integration/timeline/media.rs
+++ b/crates/matrix-sdk-ui/tests/integration/timeline/media.rs
@@ -18,6 +18,8 @@ use assert_matches::assert_matches;
 use assert_matches2::assert_let;
 use eyeball_im::VectorDiff;
 use futures_util::{FutureExt, StreamExt};
+#[cfg(feature = "unstable-msc4274")]
+use matrix_sdk::attachment::{AttachmentInfo, BaseFileInfo, GalleryConfig, GalleryItemInfo};
 use matrix_sdk::{
     assert_let_timeout,
     attachment::AttachmentConfig,
@@ -26,6 +28,8 @@ use matrix_sdk::{
 };
 use matrix_sdk_test::{async_test, event_factory::EventFactory, JoinedRoomBuilder, ALICE};
 use matrix_sdk_ui::timeline::{AttachmentSource, EventSendState, RoomExt};
+#[cfg(feature = "unstable-msc4274")]
+use ruma::events::room::message::GalleryItemType;
 use ruma::{
     event_id,
     events::room::{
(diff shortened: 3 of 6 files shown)


crates/matrix-sdk-ui/src/timeline/futures.rs
```diff
@@ -92,3 +94,37 @@ impl<'a> IntoFuture for SendAttachment<'a> {
         Box::pin(fut.instrument(tracing_span))
     }
 }
+
+#[cfg(feature = "unstable-msc4274")]
```
bnjbvr, line 96: We could have a tiny `mod galleries { ... }` accompanied with a `pub use galleries::*`; that would require only two `cfg` guards (even the `GalleryConfig` import could go in that mod). What do you think?

crates/matrix-sdk-ui/tests/integration/timeline/media.rs
```diff
@@ -262,6 +266,128 @@ async fn test_send_attachment_from_bytes() {
...
+    )
+    .await;
+
+    // Sanity check.
+    assert_let_timeout!(Some(VectorDiff::PushBack { value: item }) = timeline_stream.next());
+    assert_let!(Some(msg) = item.content().as_message());
+    assert_eq!(msg.body(), "hello");
+
+    // No other updates.
+    assert!(timeline_stream.next().now_or_never().is_none());
```
bnjbvr: Can we use `assert_pending!` here?

crates/matrix-sdk-ui/tests/integration/timeline/media.rs
```diff
@@ -262,6 +266,128 @@ async fn test_send_attachment_from_bytes() {
...
+        assert_eq!(filename, file.filename());
+        assert_eq!(Some("item caption"), file.caption());
+
+        // The URI refers to the local cache.
+        assert_let!(MediaSource::Plain(uri) = &file.source);
+        assert!(uri.to_string().contains("localhost"));
+    }
+
+    // Eventually, the media is updated with the final MXC IDs…
+    sleep(Duration::from_secs(2)).await;
```
bnjbvr: Do we need the sleep, considering that below we're using `assert_let_timeout!` (which can be configured with a higher timeout than the default)?

crates/matrix-sdk-ui/src/timeline/controller/mod.rs
```diff
@@ -232,20 +232,21 @@ pub fn default_event_filter(event: &AnySyncTimelineEvent, room_version: &RoomVer
...
+                                | MessageType::File(_)
+                                | MessageType::Image(_)
+                                | MessageType::Location(_)
+                                | MessageType::Notice(_)
+                                | MessageType::ServerNotice(_)
+                                | MessageType::Text(_)
+                                | MessageType::Video(_)
+                                | MessageType::VerificationRequest(_) => true,
+                                #[cfg(feature = "unstable-msc4274")]
+                                MessageType::Gallery(_) => false,
```
bnjbvr: I think this should be true, so as to display galleries by default. Wouldn't this make the test you've added fail, since it'd filter out the gallery event from rendering into an item? Maybe that's why CI's failing now.

bnjbvr (approved):
Nice, thank you! Approving because my comments are mostly nits. Soo, is this the last feature, or maybe we need a bit more to display and send galleries at the FFI layer?

Johennes:
> Soo, is this the last patch, or maybe we need a bit more to display and send galleries at the FFI layer?

Yes indeed, there's one more PR needed to expose galleries on the FFI layer. I just tried to keep this one here as small as possible because the other one will probably have some lengthy mapping to FFI types again. 🙈

crates/matrix-sdk-ui/src/timeline/futures.rs
Johennes: Good idea. It's slightly annoying to have to repeat the imports inside the submodule but overall this feels easier to read than the excessive `cfg` guards.

crates/matrix-sdk-ui/tests/integration/timeline/media.rs
Johennes: Interestingly I have to use 3 seconds inside `assert_let_timeout` to replace the 2 seconds from `sleep` but this still feels better because the code is shorter.

crates/matrix-sdk-ui/src/timeline/controller/mod.rs
Johennes: Oh, right! 🤦‍♂️

Have switched it around. It didn't seem to affect the test, however, which is mildly concerning. 🤔

bnjbvr: The test is likely not running; investigating.

bnjbvr: Hah:

- CI didn't run the test because the feature flag isn't ever enabled for the UI crate. I think we could decide that we support galleries by default in the timeline, so we could get rid of it now. Alternatively, we could enable the feature just a bit ahead of time in the FFI layer `Cargo.toml`, so that it's enabled workspace-wide, and included in testing. I think the latter would be great, since the FFI layer likely will want it anyways.
- @stefanceriu and I noticed a bug that local echoes never get filtered out based on content, and this is what's happening here, since the current is sending a gallery. Can you add, in this test or another one, a gallery message received via sync from another user, so we can make sure that the filtering actually works? (and temporarily revert the patch that switched this from false to true, to make sure it does filter? no need to add a specific test for the filtering)

bnjbvr (changes requested):
Requesting changes, because of the latest in-thread comment i've posted, to be super explicit. Let's test this in CI! :)

crates/matrix-sdk-ui/src/timeline/controller/mod.rs
Johennes: Oh! 🤯

Ok, I hope I got all of this right. The test now _does_ fail for me also locally (with the erroneous `false`). When I switch back to `true`, it succeeds.

testing/matrix-sdk-test/Cargo.toml
```diff
@@ -27,7 +27,7 @@ matrix-sdk-test-macros = { version = "0.11.0", path = "../matrix-sdk-test-macros
 once_cell.workspace = true
 # Enable the unstable feature for polls support.
 # "client-api-s" enables need the "server" feature of ruma-client-api, which is needed to serialize Response objects to JSON.
-ruma = { workspace = true, features = ["canonical-json", "client-api-s", "rand", "unstable-msc3381"] }
+ruma = { workspace = true, features = ["canonical-json", "client-api-s", "rand", "unstable-msc3381", "unstable-msc4274"] }
```
Johennes, line 30: I'm not sure if this should be guarded behind a feature of matrix-sdk-test?

bnjbvr: It's fine like this, thanks!

crates/matrix-sdk-ui/tests/integration/timeline/media.rs
Johennes: Oh, looks like the `sleep`delay was just increased from 2 to 4 seconds in https://github.com/matrix-org/matrix-rust-sdk/commit/8a5d4f0d82d8ac371d4f48e9b0110aeb8215dde7.

crates/matrix-sdk-ui/src/timeline/controller/mod.rs
Johennes: Test now fails as expected on the CI: https://github.com/matrix-org/matrix-rust-sdk/actions/runs/15415843410/job/43378308042?pr=5125

bnjbvr: [Does it?](https://github.com/matrix-org/matrix-rust-sdk/actions/runs/15415843410/job/43378308042?pr=5125#step:8:1195) Looks like it's the sending of the media gallery that fails, not the filtering out…

Johennes: Oh, you're right. 🤦‍♂️ Ok, will have to look into what's going on there now. 🤔

Johennes: Looks like the media config requests need mocking after merging in `main`. Added those and switched back to `false` so the CI should now fail timing out waiting for the sent gallery.

bnjbvr: Thanks, it seems to be failing as intended now: no item is pushed, while we expected one :partying_face:  Good job making the test meaningful and winning over CI!

Johennes: Lets hope it actually turns green now. 😅🤞

bnjbvr:
Thanks @Johennes, and great job as usual!

Issue #5036, opened 2025-05-14
[Task] Threads: Viewing Root Message Details


manuroe:
SDK companion issue of [[Story] Threads: Viewing Root Message Details](https://github.com/element-hq/element-meta/issues/2771) for EX.
We need to check the SDK exposes everything expected.

PR #5126, opened 2025-05-28
feat(threads): add a `ThreadSummary` to `TimelineEvent` and maintain it over time

bnjbvr:
This adds the following functionality:

- we add a new `ThreadSummary` data structure (empty at the moment; it will be filled later), and add it as a field of `TimelineEvent`
- when a thread root event contains a bundled thread summary, we translate it into a `ThreadSummary` on the carrying event.
- event cache: when an event is a threaded reply, we identify the thread root and attach a `ThreadSummary` to it
- we forward this new `ThreadSummary` into the timeline's `ThreadSummary`

This allows a main timeline to identify which events are thread roots, by querying whether they have a (timeline) `ThreadSummary` with `event_item.content().thread_summary().is_some()`. If that's the case, then it means the item can lead to the opening of a thread.

Subsequent PRs will introduce Interesting Fields™ in the common ThreadSummary. Note related to the persistent storage: as I haven't migrated any event stored on disk (to fill their thread summary if needs be), the thread summaries will be incorrect until we either perform such a migration, or clear the event cache.

Part of #5036.

### DIFF
--- a/crates/matrix-sdk-common/src/deserialized_responses.rs
+++ b/crates/matrix-sdk-common/src/deserialized_responses.rs
@@ -392,6 +392,15 @@ impl ThreadSummaryStatus {
     fn is_unknown(&self) -> bool {
         matches!(self, ThreadSummaryStatus::Unknown)
     }
+
+    /// Transforms the [`ThreadSummaryStatus`] into an optional thread summary,
+    /// for cases where we don't care about distinguishing unknown and none.
+    pub fn summary(&self) -> Option<&ThreadSummary> {
+        match self {
+            ThreadSummaryStatus::Unknown | ThreadSummaryStatus::None => None,
+            ThreadSummaryStatus::Some(thread_summary) => Some(thread_summary),
+        }
+    }
 }
 
 /// Represents a Matrix room event that has been returned from `/sync`,
--- a/crates/matrix-sdk-common/src/deserialized_responses.rs
+++ b/crates/matrix-sdk-common/src/deserialized_responses.rs
@@ -1052,6 +1052,7 @@ mod tests {
         UnableToDecryptReason, UnsignedDecryptionResult, UnsignedEventLocation, VerificationLevel,
         VerificationState, WithheldCode,
     };
+    use crate::deserialized_responses::{ThreadSummary, ThreadSummaryStatus};
 
     fn example_event() -> serde_json::Value {
         json!({
--- a/crates/matrix-sdk-common/src/deserialized_responses.rs
+++ b/crates/matrix-sdk-common/src/deserialized_responses.rs
@@ -466,7 +466,7 @@ impl TimelineEvent {
         self.kind.encryption_info()
     }
 
-    /// Takes ownership of this `TimelineEvent`, returning the (potentially
+    /// Takes ownership of this [`TimelineEvent`], returning the (potentially
     /// decrypted) Matrix event within.
     pub fn into_raw(self) -> Raw<AnySyncTimelineEvent> {
         self.kind.into_raw()
--- a/crates/matrix-sdk-common/src/deserialized_responses.rs
+++ b/crates/matrix-sdk-common/src/deserialized_responses.rs
@@ -29,7 +29,10 @@ use serde::{Deserialize, Serialize};
 #[cfg(target_arch = "wasm32")]
 use wasm_bindgen::prelude::*;
 
-use crate::debug::{DebugRawEvent, DebugStructExt};
+use crate::{
+    debug::{DebugRawEvent, DebugStructExt},
+    serde_helpers::extract_bundled_thread_summary,
+};
 
 const AUTHENTICITY_NOT_GUARANTEED: &str =
     "The authenticity of this encrypted message can't be guaranteed on this device.";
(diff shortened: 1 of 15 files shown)


crates/matrix-sdk-common/src/serde_helpers.rs
```diff
@@ -0,0 +1,140 @@
...
+#[allow(missing_debug_implementations)]
+#[derive(Deserialize)]
+struct SimplifiedContent {
+    #[serde(rename = "m.relates_to")]
+    relates_to: Option<RelatesTo>,
+}
+
+/// Try to extract the thread root from a timeline event, if provided.
+///
+/// The thread root is the field located at `content`.`m.relates_to`.`event_id`,
```
Hywan, line 50: I don't understand why the dot is outside the `<code>` here. It expresses a path, which is fine and easily understandable. Thanks for the markup effort, but I'm not sure it was worth it 😛.

crates/matrix-sdk-common/src/deserialized_responses.rs
```diff
@@ -386,11 +386,17 @@ pub struct TimelineEvent {
...
-    /// If it's set to `None`, then it means we couldn't compute those actions.
-    #[serde(skip_serializing_if = "Option::is_none")]
+    /// If it's set to `None`, then it means we couldn't compute those actions,
+    /// or that they could be computed but there were none.
+    #[serde(skip_serializing_if = "skip_serialize_push_actions")]
     pub push_actions: Option<Vec<Action>>,
 }
 
+// Don't serialize push actions if they're `None` or an empty vec.
+fn skip_serialize_push_actions(push_actions: &Option<Vec<Action>>) -> bool {
```
Hywan, line 446: 👍

crates/matrix-sdk-common/src/deserialized_responses.rs
```diff
@@ -357,6 +357,31 @@ impl<'de> Deserialize<'de> for EncryptionInfo {
     }
 }
 
+/// A simplified thread summary.
```
Hywan, line 363: Explaining what a thread summary is would probably be appreciable 🙂.

crates/matrix-sdk-common/src/serde_helpers.rs
```diff
@@ -137,4 +174,103 @@ mod tests {
...
+        assert_matches!(extract_bundled_thread_summary(&event), ThreadSummaryStatus::None);
+
+        // When there's a bundled replace, we can assert there's no thread summary.
+        let event = Raw::new(&json!({
+            "event_id": "$eid:example.com",
+            "type": "m.room.message",
+            "sender": "@alice:example.com",
+            "origin_server_ts": 42,
+            "content": {
+                "body": "Bonjour, monde!",
```
Hywan, line 234: Salut, toi.

crates/matrix-sdk-common/src/deserialized_responses.rs
```diff
@@ -383,6 +383,15 @@ impl ThreadSummaryStatus {
     fn is_unknown(&self) -> bool {
         matches!(self, ThreadSummaryStatus::Unknown)
     }
+
+    /// Transforms the [`ThreadSummaryStatus`] into an optional thread summary,
+    /// for cases where we don't care about distinguishing unknown and none.
+    pub fn maybe_summary(&self) -> Option<&ThreadSummary> {
```
Hywan: I would name this method `summary` instead of `maybe_summary`. It feels a bit more Rust idiomatic, and consistent with the rest of this code base. Not a blocker though.

testing/matrix-sdk-test/src/event_factory.rs
```diff
@@ -193,18 +193,35 @@ where
...
-    ///
-    /// Ideally, we'd type-check that an event passed as a relation is the same
-    /// type as this one, but it's not trivial to do so because this builder
-    /// is only generic on the event's *content*, not the event type itself;
-    /// doing so would require many changes, and this is testing code after
-    /// all.
-    pub fn bundled_relations(
+    /// Create a bundled thread summary in the unsigned bundled relations of
+    /// this event.
+    pub fn with_bundled_thread_summary(
```
Hywan, line 198: Pretty nice helpers!

crates/matrix-sdk/src/event_cache/room/mod.rs
```diff
@@ -1190,6 +1194,44 @@ mod private {
             Ok(updates_as_vector_diffs)
         }
 
+        /// If the event has a thread root, ensure the thread root event has a thread summary.
```
Hywan: I think this method would deserve a bit more explanation. What does it do exactly? When is it useful to extract a thread root? What is a thread root? Stuff that could bring some context.

crates/matrix-sdk/src/event_cache/room/mod.rs
```diff
@@ -1190,6 +1194,44 @@ mod private {
...
+            target_event.thread_summary = ThreadSummaryStatus::Some(ThreadSummary {});
+
+            match location {
+                EventLocation::Memory(position) => {
+                    self.events
+                        .replace_event_at(position, target_event)
+                        .expect("should have been a valid position of an item");
+                }
+                EventLocation::Store => self.save_event([target_event]).await?,
+            }
```
Hywan, line 1232: I don't understand what it does. The method is extracting the thread root and is saving it?

Also, maybe we could put this `EventLocation` logic inside `save_event`, what do you think?

crates/matrix-sdk-common/src/serde_helpers.rs
bnjbvr: I think it's worth it, because of the dot in `m.relates_to`, which is confusing if read as a single jq query. If I had put `content.m.relates_to.event_id`, you might have thought that the object resembles this shape:

```json
{
  "content": {
    "m": {
      "relates_to": {
        "event_id": "$blah"
      }
    }
  }
}
```

while the correct one is that one:

```json
{
  "content": {
    "m.relates_to": {
        "event_id": "$blah"
    }
  }
}
```

crates/matrix-sdk-common/src/deserialized_responses.rs
bnjbvr: I like it!

crates/matrix-sdk/src/event_cache/room/mod.rs
bnjbvr: > I don't understand what it does. The method is extracting the thread root and is saving it?

Not quite: if the event is a threaded reply, it makes sure the thread root event (first event in thread) has thread summary attached to it, so that clients can notice it's an entrypoint into a thread.

I think we could make something out of your suggestion, but I'd rather do it in a followup, because I think there's a bug hidden in there (and a preexisting bug in the similar code handling redactions).

crates/matrix-sdk-common/src/serde_helpers.rs
bnjbvr: another way to write it down might be `content["m.relates_to"].event_id`, but at this point not sure which one is more readable xD

Hywan: Oh you're right!

PR #5131, opened 2025-05-29
refactor(ci): Don't rerun most CI jobs when un-drafting a PR

jplatte:
This only makes sense to do for workflows that branch off of github.event.pull_request.draft, which only bindings_ci.yml does at this point in time.


jplatte:
Nope.

poljar (approved):
Alright, then let's do this.

PR #5137, opened 2025-05-29
refactor: Use native async fn in traits for BackingStore

jplatte:


bnjbvr:
Thanks for helping getting rid of `async-trait`; it's been a major culprit in increasing compile times in the past :partying_face:

PR #5140, opened 2025-05-30
Thread filtering

stefanceriu:
Yet another way to filter threaded events, based on our recent discussion.

This time it's done on top of the `TimelineAction` by moving  relation processing to the metadata so they're available before processing content then passing them into the `from_content` method together with the timeline focus and whether that particular content is a thread root or not.

The thread root processing happens on the `from_event` method where we have access to the event id so it can be compared to the one set in the timeline focus.

The next step would be to also pass in the timeline settings so that the threads feature flag can be checked as well.

Open questions: 
* should the timeline focus and settings be put in the TimelineMeta?
* should we still pass in a mutable meta to `from_event` or should we move that part of the processing to meta as well? We need to build the content before extracting relations.
* how should this work for redacted and encrypted events?
* `TimelineAction` is also used for building replies which doesn't know about the timeline focus 🤔

### DIFF
--- a/crates/matrix-sdk-ui/src/timeline/controller/metadata.rs
+++ b/crates/matrix-sdk-ui/src/timeline/controller/metadata.rs
@@ -17,15 +17,44 @@ use std::{
     sync::Arc,
 };
 
-use ruma::{EventId, OwnedEventId, OwnedUserId, RoomVersionId};
+use imbl::Vector;
+use matrix_sdk::deserialized_responses::EncryptionInfo;
+use ruma::{
+    events::{
+        poll::unstable_start::UnstablePollStartEventContent, relation::Replacement,
+        room::message::RelationWithoutReplacement, AnyMessageLikeEventContent,
+        AnySyncMessageLikeEvent, AnySyncTimelineEvent, BundledMessageLikeRelations,
+    },
+    serde::Raw,
+    EventId, OwnedEventId, OwnedUserId, RoomVersionId,
+};
 use tracing::trace;
 
 use super::{
     super::{subscriber::skip::SkipCount, TimelineItem, TimelineItemKind, TimelineUniqueId},
     read_receipts::ReadReceipts,
-    Aggregations, AllRemoteEvents, ObservableItemsTransaction,
+    Aggregation, AggregationKind, Aggregations, AllRemoteEvents, ObservableItemsTransaction,
+    PendingEdit, PendingEditKind,
 };
-use crate::unable_to_decrypt_hook::UtdHookManager;
+use crate::{
+    timeline::{
+        event_item::{
+            extract_bundled_edit_event_json, extract_poll_edit_content,
+            extract_room_msg_edit_content,
+        },
+        InReplyToDetails, TimelineEventItemId,
+    },
+    unable_to_decrypt_hook::UtdHookManager,
+};
+
+/// All parameters to [`TimelineAction::from_content`] that only apply if an
+/// event is a remote echo.
+pub struct RemoteEventContext<'a> {
+    pub event_id: &'a EventId,
+    pub raw_event: &'a Raw<AnySyncTimelineEvent>,
+    pub relations: BundledMessageLikeRelations<AnySyncMessageLikeEvent>,
+    pub bundled_edit_encryption_info: Option<Arc<EncryptionInfo>>,
+}
 
 #[derive(Clone, Debug)]
 pub(in crate::timeline) struct TimelineMetadata {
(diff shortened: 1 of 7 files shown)


crates/matrix-sdk-ui/src/timeline/controller/metadata.rs
```diff
@@ -278,6 +307,144 @@ impl TimelineMetadata {
...
+                    .map(|in_reply_to| InReplyToDetails::new(in_reply_to.event_id, timeline_items))
+            }
+            _ => None,
+        });
+
+        (in_reply_to, thread_root)
+    }
+
+    fn mark_response(
+        meta: &mut TimelineMetadata,
```
bnjbvr, line 434: nit: &mut self

crates/matrix-sdk-ui/src/timeline/event_handler.rs
```diff
@@ -319,10 +334,26 @@ impl TimelineAction {
...
+                }
+            }
+            TimelineFocusKind::Thread { root_event_id } => {
+                if is_thread_root || thread_root.as_ref().is_some_and(|r| r == root_event_id) {
+                } else {
+                    return None;
+                }
+            }
+            _ => {}
+        }
```
bnjbvr, line 355: Oops, this is not the idea I had in mind. Instead, it's `TimelineStateTransaction::should_add_event_item` which should make this decision, and it already has the `TimelineSettings` as a parameter (so it can follow the user decision to exclude in-thread events or not). You don't have to worry about the threaded timeline having to filter out events, since it'll be listening to a stream of `VectorDiff TimelineItem` that only includes the target thread's events, in the future.

So eventually the flow would look like this, in `handle_remote_event`:

- deserialize (as we do now; nothing to change)
- if deserialization was successful: extract the reply_to/threads info, and process `meta.replies` (using the methods you've moved over to `TimelineMetadata` — great stuff btw!)
- then call `should_add_event_item` with the new optional thread we just extracted

And adapt other callers too. A sign that you're on the right track is that `TimelineAction` should only take a `&TimelineMetadata` parameter, and not `&mut TimelineMetadata` anymore.

crates/matrix-sdk-ui/tests/integration/timeline/thread.rs
```diff
@@ -193,3 +191,47 @@ async fn test_thread_backpagination() {
...
+                        .event_id(event_id!("$original_event"))
+                        .in_thread(&thread_root_event_id, event_id!("$first_reply_in_thread")),
+                ),
+        )
+        .await;
+
+    timeline_stream.next().await;
+
+    // The main timeline only contains the date separator and the thread root
+    assert_eq!(timeline.items().await.len(), 2);
```
bnjbvr, line 236: We'll make sure to check what the timeline stream update is, and what are the items.

crates/matrix-sdk-ui/src/timeline/event_handler.rs
```diff
@@ -198,16 +188,25 @@ impl TimelineAction {
...
+            };
+
         let redacted_message_or_none = |event_type: MessageLikeEventType| {
             (event_type != MessageLikeEventType::Reaction)
                 .then_some(TimelineItemContent::MsgLike(MsgLikeContent::redacted()))
         };
 
         Some(match event {
             AnySyncTimelineEvent::MessageLike(AnySyncMessageLikeEvent::RoomRedaction(ev)) => {
+                // How does this work for threads? I see Element Web doesn't show redactions 🤔
```
bnjbvr, line 209: (make sure to include `XXX` in such comments that are open questions, so we don't forget to remove then from the code later :))

crates/matrix-sdk-ui/src/timeline/event_handler.rs
```diff
@@ -240,41 +239,57 @@ impl TimelineAction {
                             .await;
                         }
 
+                        // Same question here, how does this work for threads?
```
bnjbvr, line 242: ditto here

crates/matrix-sdk-ui/src/timeline/controller/mod.rs
```diff
@@ -181,11 +181,11 @@ impl Default for TimelineSettings {
     }
 }
 
-#[derive(Debug, Clone, Copy)]
+#[derive(Debug, Clone)]
 pub(super) enum TimelineFocusKind {
     Live,
     Event,
-    Thread,
+    Thread { root_event_id: OwnedEventId },
```
bnjbvr, line 188: (Don't think we'll need this, see other comments!)

crates/matrix-sdk-ui/src/timeline/event_handler.rs
stefanceriu: > Instead, it's TimelineStateTransaction::should_add_event_item which should make this decision

I thought the idea was to let `TimelineAction` do all the filtering and not "scatter the responsibility of filtering into two places" 🤔 
More importantly though, how would this work for `State.handle_local_event`  which has its own `should_add_new_items`? 

> if deserialization was successful: extract the reply_to/threads info

If I am to extract those out of the raw event in `StateTransaction` I will have to move the `TimelineAction::from_event` part to meta too. Is that the way you want to go?

stefanceriu: Also, doesn't

> A sign that you're on the right track is that TimelineAction should only take a &TimelineMetadata parameter

conflict with 

> should we still pass in a mutable meta to from_event or should we move that part of the processing to meta as well? We need to build the content before extracting relations.
> see big main comment, but TLDR: yes i think so!

stefanceriu:
> how should this work for redacted and encrypted events?
> I don't think these would require any special treatment, but maybe I'm missing something?

The problem there was that `from_event` directly builds the content without checking relations so `from_content` no longer has access to filter out or in those events.

crates/matrix-sdk-ui/src/timeline/event_handler.rs
bnjbvr: (To latest comment: i guess the "A or B" question shouldn't have been answered with "yes", my bad :))

> I thought the idea was to let TimelineAction do all the filtering and not "scatter the responsibility of filtering into two places" 🤔

The filtering would happen only in `should_add` this way, as it's being done right now, so that's the right place where to do it. The `TimelineAction` code shouldn't have to filter in or out some events. in general.

> More importantly though, how would this work for State.handle_local_event which has its own should_add_new_items?

We'd also call the meta extractors here; if this leads to much code duplication, we can factor it out in a small function helper.

> If I am to extract those out of the raw event in StateTransaction I will have to move the TimelineAction::from_event part to meta too. Is that the way you want to go?

I think so, yes.

bnjbvr:
> > how should this work for redacted and encrypted events?
> > I don't think these would require any special treatment, but maybe I'm missing something?
> 
> The problem there was that `from_event` directly builds the content without checking relations so `from_content` no longer has access to filter out or in those events.

Ok, so I think with the new way I suggested (where it's `should_add` that filters out or not), this issue is moot, as `from_content` doesn't have to decide to filter or not. Does it make more sense now?

stefanceriu:
Closing this in favor of https://github.com/matrix-org/matrix-rust-sdk/pull/5185

PR #5141, opened 2025-05-30
SDK: Add `ClientBuilder::with_enable_share_history_on_invite`

richvdh:
Replace `experimental-share-history-on-invite` feature flag with a runtime flag on the `Client`.

Part of https://github.com/element-hq/element-meta/issues/2871

Signed-off-by:

### DIFF
--- a/crates/matrix-sdk/src/room/mod.rs
+++ b/crates/matrix-sdk/src/room/mod.rs
@@ -385,7 +385,6 @@ impl Room {
             ))));
         }
 
-        #[cfg(all(feature = "experimental-share-history-on-invite", feature = "e2e-encryption"))]
         let inviter = if prev_room_state == RoomState::Invited {
             match self.invite_details().await {
                 Ok(details) => details.inviter,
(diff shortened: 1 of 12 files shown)


crates/matrix-sdk/src/room/mod.rs
```diff
@@ -400,11 +399,17 @@ impl Room {
...
+        #[cfg(feature = "e2e-encryption")]
+        if self.client.inner.enable_share_history_on_invite {
+            if let Some(inviter) = inviter {
+                shared_room_history::maybe_accept_key_bundle(self, inviter.user_id()).await?;
+            }
         }
 
+        #[cfg(not(feature = "e2e-encryption"))]
+        // Suppress "unused variable" lint
+        let _inviter = inviter;
```
richvdh, line 411: I don't know if there is a less awful way to do this. I wanted to keep the code flow above consistent between `feature = "e2e-encryption"` and not, so didn't really want another `#[cfg(feature = "e2e-encryption")]` where `inviter` is calculated.

stefanceriu: Perhaps inline the whole thing?

```
if prev_room_state == RoomState::Invited {
    match self.invite_details().await {
        Ok(details) =>
        {
            if self.client.inner.enable_share_history_on_invite {
                if let Some(inviter) = details.inviter {
                    shared_room_history::maybe_accept_key_bundle(self, inviter.user_id())
                        .await?;
                }
            }
        }
        Err(e) => {
            warn!("No invite details were found, can't attempt to find a room key bundle to accept: {e:?}");
        }
    }
};
```

Either way it looks fine to me, doesn't seem like a big issue.

richvdh: unfortunately we need to call `invite_details` before doing the join.

richvdh:
@poljar and I discussed whether this would be better represented as a field in `EncryptionSettings` rather than `Client`. The answer is: probably, but adding it would be a breaking API change (since `EncryptionSettings`'s fields are currently public, so it can be instantiated in application code), as would removing it when the feature becomes stable (unless we took the opportunity of the *first* breaking change to require that `EncryptionSettings` be created via a constructor).

Although removing `with_enable_share_history_on_invite` is *technically* a breaking change, we can argue that you're setting yourself up for failure by using an experimental feature.

I'm going to leave this as-is for now, but we should consider using `EncryptionSettings` for future changes of this sort.

PR #5143, opened 2025-05-30
Test: Replace `sync_timeline_event!` with `EventFactory` for events in timeline subscribe tests

yousefmoazzam:
Part of #3716

For the tests which were modified, this additionally:
- removes server timestamp information on events that weren't necessary for the associated test to pass
- for places where multiple events were added at the same time, the code was very slightly tidied by adding them in bulk rather than adding them individually

If it's preferred to do those changes separate to the removal of the `sync_timeline_event!` macro, I'm happy to split them off into a separate PR :slightly_smiling_face:


poljar:
> If it's preferred to do those changes separate to the removal of the sync_timeline_event! macro, I'm happy to split them off into a separate PR 🙂

Nah, it's fine, the PR is small enough already.

poljar (approved):

PR #5144, opened 2025-06-02
Upgrade Ruma

zecakeh:
Use the newly released version, since a new release of the SDK is planned soon.


andybalaam (approved):
Great, thank you!

PR #5148, opened 2025-06-02
feat(multiverse): Add a /leave command

poljar:


andybalaam (approved):
Great

labs/multiverse/src/widgets/room_view/mod.rs
```diff
@@ -288,9 +288,26 @@ impl RoomView {
         }
     }
 
+    async fn leave_room(&mut self) {
+        let Some(room) = self
+            .selected_room
+            .as_deref()
+            .and_then(|room_id| self.ui_rooms.lock().get(room_id).cloned())
```
richvdh: When I looked at this, it struck me that we could usefully factor this out as a helper method. YMMV

richvdh:
Sorry this will have conflicted with https://github.com/matrix-org/matrix-rust-sdk/pull/5146

labs/multiverse/src/widgets/room_view/mod.rs
```diff
@@ -288,9 +288,26 @@ impl RoomView {
...
+        else {
+            self.status_handle
+                .set_message(format!("Coulnd't find the room object to leave the room"));
+            return;
+        };
+
+        let _ = room.leave().await.inspect_err(|e| {
+            self.status_handle.set_message(format!("Couldn't leave the room {e:?}"))
+        });
```
richvdh: Do we not need to `self.input.clear()` here?

poljar: Hmm, I tried this out and I think the input was gone. Will check again.

labs/multiverse/src/widgets/room_view/mod.rs
poljar: Yup that's probably a good idea, more commands might reuse the same snippet. Will add a helper.

poljar: This gave me the opportunity to use async closures: https://github.com/matrix-org/matrix-rust-sdk/pull/5148/commits/fa43f7375d6d3f921c48a3cbd5f263296bd696f7.

labs/multiverse/src/widgets/room_view/mod.rs
poljar: Ah, you were right. I just didn't pay attention: https://github.com/matrix-org/matrix-rust-sdk/pull/5148/commits/b4a49e03f848faebc9b900ce0da7b9a6a4831ada.

poljar:
> Sorry this will have conflicted with #5146

Nah, that was an easy resolve.

PR #5149, opened 2025-06-02
feat(wasm): Add Wasm aware async_trait proc macro

zzorba:
Create a new proc-macro to eliminate conditional cfgs wherever async_trait is defined. Convert existing use of conditional wasm cfg to use this macro, as well as other call site thats were using async_trait::async_trait directly

Signed-off-by: Daniel Salinas

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2875,7 +2875,6 @@ dependencies = [
  "assert_matches2",
  "async-channel",
  "async-stream",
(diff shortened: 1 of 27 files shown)


zzorba:
Given the new change is much simpler, going to close this PR in favor of https://github.com/matrix-org/matrix-rust-sdk/pull/5156

PR #5150, opened 2025-06-02
feat(base): Detecting invalid states in room upgrades

Hywan:
This set of patches adds a new `check_tombstone` response processor, and a new `InvalidTombstonedRooms` error in `matrix_sdk_base::Error` when some invalid states with room upgrades are detected. So far three problematic states are known:

- loop: when room upgrades are creating a loop, e.g. room A is upgrading to itself, or another example, when roo A is upgrading to room B that is also upgrading to room A.
- merger: when room A upgrades to room C, and when room B upgrades to room C
- splitter: it's the opposite of a _merger_, e.g. room A upgrades to both room B and room C

The first two are detecting: loop and merger. The splitter can be easily detected but it seems okay in some case, and the spec is unclear about what to do in this case. This is under discussion. We can decide to be stricter than the spec for "safety" reason though. I let the reviewer gives me feedback here.


crates/matrix-sdk-base/src/response_processors/state_events.rs
```diff
@@ -248,15 +251,492 @@ where
...
+/// a loop or a merger.
+///
+/// Please see [`Error::InconsistentTombstonedRooms`] to learn more about the
+/// kind of invalid states it can detect.
+pub fn check_tombstone(
+    context: &mut Context,
+    room_updates: &RoomUpdates,
+    state_store: &BaseStateStore,
+) -> Result<()> {
+    let mut room_updates = room_updates
```
stefanceriu, line 264: Perhaps `room_upgrades` works better here

crates/matrix-sdk-base/src/response_processors/state_events.rs
```diff
@@ -248,15 +251,492 @@ where
...
+
+            // First time we see this room: let's remember it.
+            already_seen.insert(room_id);
+
+            let Some(replacement_room_id) = maybe_replacement_room_id else {
+                // No replacement room. Stop looking for one.
+                break;
+            };
+
+            // Look up for the replacement room in the `room_updates`.
```
stefanceriu:
```suggestion
            // Look up a subsequent replacement room in `room_updates`.
```

crates/matrix-sdk-base/src/response_processors/state_events.rs
Hywan: I prefer to stick to the `RoomUpdates` naming for the moment 🙂.

crates/matrix-sdk-base/src/response_processors/state_events.rs
Hywan: yup, better, thanks!

PR #5152, opened 2025-06-02
fix(event cache): when replacing an item in memory, also save it in the store backend

bnjbvr:
A followup inspired by https://github.com/matrix-org/matrix-rust-sdk/pull/5126#discussion_r2120662405. There was a subtle bug (proven by the changes to one existing redaction test), where a redaction would affect the in-memory linked chunk, but not propagate the change to the store backend; the next time we'd open the room event cache, the redacted event would be available in the clear (even though any observers would also re-read the redaction event and likely redact the to-be-redacted event, in that case, so it wasn't dramatic).

Same issue applied when adding a thread summary to an event. I haven't added a test for it, as it's exactly the same bug as the redaction, that's correctly tested separately.

Part of #5036.

### DIFF
--- a/crates/matrix-sdk/src/event_cache/room/mod.rs
+++ b/crates/matrix-sdk/src/event_cache/room/mod.rs
@@ -1222,13 +1222,34 @@ mod private {
 
             target_event.thread_summary = ThreadSummaryStatus::Some(ThreadSummary {});
 
+            self.replace_event_at(location, target_event).await?;
+
+            Ok(())
+        }
+
+        /// Replaces a single event, be it saved in memory or in the store.
+        ///
+        /// If it was saved in memory, this will emit a notification to
+        /// observers that a single item has been replaced. Otherwise,
+        /// such a notification is not emitted, because observers are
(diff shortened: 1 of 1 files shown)


crates/matrix-sdk/src/event_cache/room/mod.rs
```diff
@@ -1302,6 +1302,9 @@ mod private {
                             self.events
                                 .replace_event_at(position, copy)
                                 .expect("should have been a valid position of an item");
+                            // We just changed the in-memory representation; synchronize this with
+                            // the store.
+                            self.propagate_changes().await?;
```
Hywan: We have `with_events_mut` to automate that. Should we use it here?

Hywan: Ok, looking at the next patch, it makes sense once put in its own method.

bnjbvr: Yeah, and it's called from within `with_events_mut`. Infinite recursion, anyone? :woozy_face:

PR #5153, opened 2025-06-02
feat(event cache): include useful information in the `ThreadSummary`

### LINKED ISSUE #5036: [Task] Threads: Viewing Root Message Details

bnjbvr:
Based on top of #5152.

Include some useful information in the `ThreadSummary`:

- the number of replies in thread (excluding the thread root). That's equivalent to what a bundled thread summary would include. It's also easy to compute a meaningful value for this: it's all the events that `m.relate_to` point to the thread root event; since only room messages (and not aggregations) can have it, it's meaningful by design.
- the latest thread event. Note that I've implemented it mostly for demonstration purposes, and calling this first version done; when each thread has a baking linked chunk, it will be computed properly and more efficiently. Now, what it does is override the latest event with the event if it came from sync; or use the first paginated in-thread event it sees, if there wasn't any prior latest event, when the event came from a (room) network backpagination. It's a bit hacky, but ought to be correct most of the time.

Fixes #5036.

### DIFF
--- a/crates/matrix-sdk-ui/src/timeline/controller/state_transaction.rs
+++ b/crates/matrix-sdk-ui/src/timeline/controller/state_transaction.rs
@@ -16,9 +16,11 @@ use std::collections::{HashMap, HashSet};
 
 use eyeball_im::VectorDiff;
 use itertools::Itertools as _;
-use matrix_sdk::deserialized_responses::{TimelineEvent, TimelineEventKind, UnsignedEventLocation};
+use matrix_sdk::deserialized_responses::{
+    ThreadSummaryStatus, TimelineEvent, TimelineEventKind, UnsignedEventLocation,
+};
 use ruma::{
-    events::AnySyncTimelineEvent, push::Action, serde::Raw, MilliSecondsSinceUnixEpoch,
+    events::AnySyncTimelineEvent, push::Action, serde::Raw, EventId, MilliSecondsSinceUnixEpoch,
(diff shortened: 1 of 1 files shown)


crates/matrix-sdk/src/event_cache/room/mod.rs
```diff
@@ -1224,8 +1224,17 @@ mod private {
...
 
+            if let ThreadSummaryStatus::Some(existing) = &target_event.thread_summary {
+                if existing == &new_summary {
+                    trace!("thread summary is already up-to-date");
+                    return Ok(());
+                }
+            }
+
+            // Cause an update to observers.
+            target_event.thread_summary = ThreadSummaryStatus::Some(new_summary);
```
Hywan, line 1258: Do you think it should be tested?

Hywan (approved):
Looks good to me. Thanks for the effort of doing patches as small as possible.

crates/matrix-sdk/src/event_cache/room/mod.rs
bnjbvr: It is tested!… in a timeline integration test, in the final test that has adds more in-thread events, and observes that the thread summary is updated over time :+1:

PR #5156, opened 2025-06-02
feat(wasm): Replace async_trait with wasm aware cfg conditional

zzorba:
Change the use of async_trait to remove required Send for wasm family of platforms.

Signed-off-by: Daniel Salinas


zzorba:
@Hywan I don't have permissions to request reviewers, but wanted to put this on your plate.

PR #5157, opened 2025-06-02
Test: Replace `sync_timeline_event!` with `EventFactory` for sticker event in timeline test

yousefmoazzam:
Part of #3716

I'm fairly sure that I've used the correct relation in the sticker event content (a reply to a specific message within a thread?), but please do double-check me on that in particular.

I also removed some information attached to the event that wasn't being used in the test:
- the server timestamp
- the various fields on the `ImageInfo` (it seemed like, in its current state, the test is primarily checking that the event is a sticker event and that the relation to another event is correct, and so the actual image info isn't important) 

The first one likely doesn't matter if it's done or not, but maybe the second one shouldn't be done if it'd make sense to be asserting on the image information - I'm happy to remove that change if desired.


Hywan (approved):
Fantastic, thank you very much!

PR #5158, opened 2025-06-03
feat(wasm): Add Wasm platform aware test to export proc-macro

zzorba:
Change the uniffi export to not use tokio on Wasm platforms.

Signed-off-by: Daniel Salinas


poljar (approved):
Makes sense.

PR #5161, opened 2025-06-03
feat(event cache): save a bundled thread's latest event into the event cache

bnjbvr:
Builds on top of #5153, only the last commit is relevant; drafting for now.

The hardship is that we're building a `TimelineEvent` out of the `Raw<AnyMessageLikeEvent>` included in the bundled thread summary, and as such we need to include the encryption information, which is… tricky and error-prone. I wonder if there's a way to do something simpler there.


bnjbvr:
Superseded by #5165

PR #5008, opened 2025-05-06
feat(send_queue): report progress for media uploads

Johennes:
This makes it possible to listen to the media upload progress when using the send queue. The progress is communicated via `EventSendState::NotSentYet`.

I haven't yet fixed / enhanced the tests or added a changelog because I'm not 100% sure about this approach and would like to get some preliminary feedback first.

- [ ] Public API changes documented in changelogs (optional)

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1210187425391164

---

Edit by @Hywan:

- Address https://github.com/element-hq/element-x-ios/issues/4055

PR #5163, opened 2025-06-03
feat(ffi): Expose method for sending galleries

Johennes:
Addendum to https://github.com/matrix-org/matrix-rust-sdk/pull/5125 to allow sending galleries from the FFI crate. This is the final PR for galleries (apart from possible enhancements to report the upload status in https://github.com/matrix-org/matrix-rust-sdk/pull/5008).

This is somewhat lengthy again, apologies. Most of the changes are just wrappers and type mapping, however. So I was hoping that it's relatively straightforward to review in bulk. Happy to try and elaborate on the changes or break them up into smaller commits if that helps, however.

### DIFF
--- a/bindings/matrix-sdk-ffi/src/ruma.rs
+++ b/bindings/matrix-sdk-ffi/src/ruma.rs
@@ -345,15 +345,38 @@ impl From<Mentions> for ruma::events::Mentions {
 
 #[derive(Clone, uniffi::Enum)]
 pub enum MessageType {
-    Emote { content: EmoteMessageContent },
-    Image { content: ImageMessageContent },
-    Audio { content: AudioMessageContent },
-    Video { content: VideoMessageContent },
-    File { content: FileMessageContent },
-    Notice { content: NoticeMessageContent },
-    Text { content: TextMessageContent },
-    Location { content: LocationContent },
-    Other { msgtype: String, body: String },
+    Emote {
+        content: EmoteMessageContent,
+    },
+    Image {
+        content: ImageMessageContent,
+    },
+    Audio {
+        content: AudioMessageContent,
+    },
+    Video {
+        content: VideoMessageContent,
+    },
+    File {
+        content: FileMessageContent,
+    },
+    #[cfg(feature = "unstable-msc4274")]
+    Gallery {
+        content: GalleryMessageContent,
+    },
+    Notice {
+        content: NoticeMessageContent,
+    },
+    Text {
+        content: TextMessageContent,
+    },
+    Location {
+        content: LocationContent,
+    },
+    Other {
+        msgtype: String,
+        body: String,
+    },
 }
 
 /// From MSC2530: https://github.com/matrix-org/matrix-spec-proposals/blob/main/proposals/2530-body-as-caption.md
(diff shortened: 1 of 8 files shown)


crates/matrix-sdk-ui/src/timeline/mod.rs
```diff
@@ -830,3 +835,157 @@ where
...
+    }
+
+    /// Checks whether the gallery contains any media items or not.
+    pub fn is_empty(&self) -> bool {
+        self.items.is_empty()
+    }
+}
+
+#[cfg(feature = "unstable-msc4274")]
+impl TryFrom<GalleryConfig> for matrix_sdk::attachment::GalleryConfig {
```
Johennes, line 939: Having to duplicate `GalleryConfig` into another type here is really annoying but this way we can delay loading the data from the `AttachmentSource` until we're inside the `SendGallery` future. This means that `send_gallery` returns earlier which, I think, is desirable?

Johennes:
I think the CI failures are caused by complement-crypto using uniffi 0.25.0. Is there a reason why it's not using 0.28.0 like the SDK? Apart from that, I think this is ready for review.

bnjbvr:
> I think the CI failures are caused by complement-crypto using uniffi 0.25.0. Is there a reason why it's not using 0.28.0 like the SDK? Apart from that, I think this is ready for review.

Yeah, complement-crypto uses the Go bindings generator, which is slightly outdated and requires using uniffi 0.25.0, so it is our least common denominator here. In the meanwhile, the code has to be compatible with uniffi 0.25.0, until this is upgraded in complement-crypto.

bindings/matrix-sdk-ffi/src/ruma.rs
```diff
@@ -568,6 +526,31 @@ pub struct ImageMessageContent {
     pub info: Option<ImageInfo>,
 }
 
+impl From<ImageMessageContent> for RumaImageMessageEventContent {
+    fn from(value: ImageMessageContent) -> Self {
+        let (body, filename) = get_body_and_filename(value.filename, value.caption);
+        let mut event_content = Self::new(body, (*value.source).clone().into())
```
bnjbvr, line 532: no ideas if that idea applies cleanly, but: is there no way to avoid the clone here? It seems spurious since the value is passed by ownership and consumer by this method.

(and it happens a few times below as well)

bindings/matrix-sdk-ffi/src/ruma.rs
```diff
@@ -1694,3 +1756,107 @@ impl From<RumaRoomAccountDataEvent<UnstableMarkedUnreadEventContent>> for RoomAc
...
+                    body: value.body().to_owned(),
+                },
+            })
+        }
+    }
+
+    #[uniffi::export]
+    pub fn formatted_body_from_html(body: String) -> FormattedBody {
+        FormattedBody::from(&RumaFormattedBody::html(body))
+    }
```
bnjbvr: 1. Can you add a doc comment?
2. Why is this needed? (I'm not saying it shouldn't exist, but I'm curious to know if we have some similar feature in the FFI layer already, or if it's allowing doing new things)

bindings/matrix-sdk-ffi/src/timeline/mod.rs
```diff
@@ -1367,3 +1367,270 @@ impl LazyTimelineItemProvider {
...
+
+    #[derive(uniffi::Record)]
+    pub struct AudioGalleryItemInfo {
+        /// Audio metadata.
+        pub audio_info: AudioInfo,
+        /// Filename.
+        pub filename: String,
+        /// Caption.
+        pub caption: Option<String>,
+        /// Rich caption.
```
bnjbvr: Here and below: I don't think these doc comments are useful, so let's remove them.

bindings/matrix-sdk-ffi/src/timeline/mod.rs
```diff
@@ -1367,3 +1367,270 @@ impl LazyTimelineItemProvider {
...
+        pub formatted_caption: Option<FormattedBody>,
+        /// Path to the thumbnail.
+        pub thumbnail_path: Option<String>,
+    }
+
+    #[derive(uniffi::Object)]
+    pub struct SendGalleryJoinHandle {
+        join_hdl: Arc<Mutex<JoinHandle<Result<(), RoomError>>>>,
+        abort_hdl: AbortHandle,
```
bnjbvr: Might be a copy pasta from some other code, but I find `hdl` to be hard to understand and expand: let's fully expand `handle` in the names here.

crates/matrix-sdk-ui/src/timeline/mod.rs
bnjbvr: Yep, it's fine like this. Can you add a doc comment to that effect, explaining why we're duplicating the struct, on top of `GalleryConfig` please?

crates/matrix-sdk-ui/src/timeline/mod.rs
```diff
@@ -830,3 +835,157 @@ where
...
+    pub formatted_caption: Option<FormattedBody>,
+    /// The thumbnail.
+    pub thumbnail: Option<Thumbnail>,
+}
+
+#[cfg(feature = "unstable-msc4274")]
+impl TryFrom<GalleryItemInfo> for matrix_sdk::attachment::GalleryItemInfo {
+    type Error = Error;
+
+    fn try_from(value: GalleryItemInfo) -> std::result::Result<Self, Self::Error> {
```
bnjbvr: nit: do we have to precise `std::result::Result` here? If so, it might mean that the `Result` type we're defining crate/module wide could use two generic parameters, with the second one defaulting to the error type which might be hardcoded right now.

bnjbvr (changes requested):
Looks great! only requesting changes because we need to make it work for complement-crypto, as alluded to in the previous comment (unless you're interested in a side quest to bump the go bindings in complement-crypto :D).

crates/matrix-sdk-ui/src/timeline/mod.rs
Johennes: No, actually not. Sorry, I think auto-completion in my IDE might have inserted this and I hadn't noticed.

bindings/matrix-sdk-ffi/src/timeline/mod.rs
Johennes: Good idea. Have fixed it in the types I copied from, too.

bindings/matrix-sdk-ffi/src/ruma.rs
Johennes: Actually, sorry, I don't remember how I ended up with this in the project I've been prototyping in. I don't think this is needed because `FormattedBody` could just be constructed directly across the FFI instead. Have removed it.

bindings/matrix-sdk-ffi/src/ruma.rs
Johennes: Hm, I haven't found a way to avoid the clone. `Self::new` requires a plain `MediaSource` but `value.source` wraps it in an `Arc`. So I have to `clone` the inner value because I cannot move it out of the `Arc`. I'm not sure I'm fully following why because `value` is also consumed by the method. My Rust-fu might just be too weak. 😕

bnjbvr (approved):
Thanks a bunch, and congrats on shipping the gallery feature on the Rust SDK, end-to-end :partying_face:

Johennes:
Thanks a ton for reviewing all these bulky PRs! ❤️

PR #5164, opened 2025-06-04
fix(base): Remove false positives in `check_tombstone`.

Hywan:
This patch removes false positives in `check_tombstone`. See the new `test_check_tombstone_no_loop_within_misordered_rooms` test which explains the problem. In a nutshell, the `Response` type holds room updates in a `HashMap`, thus the room updates are _re-ordered_ by the `HashMap`. It creates problems with the algorithm in `check_tombstone`. This patch re-orders room updates by room versions. Now we can have false negatives if the room version is unknown. A discussion is happening to make the `m.room.create` not optional in all the sync mechanisms.

Kudos to @stefanceriu for being the wise person that has found the problem (see https://github.com/matrix-org/matrix-rust-sdk/compare/stefan/failingInvalidRoomUpgradesAlgorithm?expand=1).

### DIFF
--- a/crates/matrix-sdk-base/src/response_processors/state_events.rs
+++ b/crates/matrix-sdk-base/src/response_processors/state_events.rs
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use std::collections::{BTreeMap, BTreeSet};
+use std::collections::BTreeSet;
 
 use ruma::{events::AnySyncStateEvent, serde::Raw};
 use serde::Deserialize;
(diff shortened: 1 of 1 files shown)


Hywan:
@andybalaam told me the room version is an opaque value and we should not sort by room version. I don't find this in https://spec.matrix.org/v1.14/client-server-api/#mroomcreate, but apparently https://spec.matrix.org/v1.14/rooms/ uses a regular versioning, even if the grammar is quite liberal https://spec.matrix.org/v1.14/rooms/#room-version-grammar…

Hywan:
It is possible for a room `12` to ugprade to `2` for example. Nothing prevents that. So we can't compare room versions. We need another algorithm.

PR #5165, opened 2025-06-04
refactor(sdk): extract a bundled thread's latest event as a `TimelineEvent`

bnjbvr:
A few refactorings first around the `TimelineEvent` creation, which was a bit scattered and messy. Then, this patch makes it possible to extract a bundled thread's latest event as a `TimelineEvent`, and store it temporarily in the `TimelineEvent`; it's not persisted, since it's information that could in theory be recovered, and our only use case right now is saving it (the first time we observe it via sync/pagination) in the event cache, so it's sufficient and efficient to do it like this.

This is a followup to #5153, so part of #4869 and #5036.

### DIFF
--- a/crates/matrix-sdk/src/event_cache/room/mod.rs
+++ b/crates/matrix-sdk/src/event_cache/room/mod.rs
@@ -1172,9 +1172,15 @@ mod private {
             // Update the store before doing the post-processing.
             self.propagate_changes().await?;
 
-            for event in &events_to_post_process {
-                self.maybe_apply_new_redaction(event).await?;
-                self.analyze_thread_root(event, is_live_sync).await?;
+            for event in events_to_post_process {
+                self.maybe_apply_new_redaction(&event).await?;
+
+                self.analyze_thread_root(&event, is_live_sync).await?;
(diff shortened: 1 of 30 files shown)


crates/matrix-sdk/src/event_cache/room/mod.rs
```diff
@@ -1172,9 +1172,16 @@ mod private {
...
-                self.maybe_apply_new_redaction(event).await?;
-                self.analyze_thread_root(event, is_live_sync).await?;
+            for event in events_to_post_process {
+                self.maybe_apply_new_redaction(&event).await?;
+
+                self.analyze_thread_root(&event, is_live_sync).await?;
+
+                // Save a bundled thread event, if there was one.
+                if let Some(bundled_thread) = event.bundled_latest_thread_event {
+                    warn!("FOUND IT");
```
stefanceriu: Maybe remove this warning?

bnjbvr: ohnoes, i got caught doing printf debugging xD

PR #5166, opened 2025-06-04
fix(timeline): add the timeline start of the current pagination state too

bnjbvr:
Might fix a weird race condition if a room has few events/is still fully loaded in the room event cache.


ganfra:
Can confirm it seems to fix the issue on my device 👍

bnjbvr:
@Hywan re-requesting review because I've changed the strategy: instead of waiting for the user to call `live_pagination_status()`, the timeline start item is now inserted eagerly upon creation of the timeline. This is even better. You can see the difference in behavior by looking at the test changes in the last commit.

Hywan (approved):
Looks good to me, thanks!

PR #5171, opened 2025-06-04
refactor: Add extra logs to Client::send_call_notifications_if_needed

jmartinesp:
Also make it return `Result<bool>` instead of `Result<()>` so we can check if the event was sent.

Add new test for the 'active call' case and adapt the rest to the new return type.

Signed-off-by:


jmartinesp:
This is mostly so we can debug some issues on the EX client where this event is not sent intermittently.

PR #5172, opened 2025-06-04
refactor(event cache): prepare the event cache for storing a linked chunk per thread

bnjbvr:

Signed-off-by:

### DIFF
--- a/crates/matrix-sdk/src/event_cache/room/mod.rs
+++ b/crates/matrix-sdk/src/event_cache/room/mod.rs
@@ -390,16 +390,6 @@ impl RoomEventCacheInner {
 
         let mut state = self.state.write().await;
 
-        // Ditch the previous-batch token if the sync isn't limited and we've seen at
-        // least one event in the past.
-        //
-        // In this case (and only this one), we should definitely know what the head of
-        // the timeline is (either we know about all the events, or we have a
-        // gap somewhere), since storage is enabled by default.
-        if !timeline.limited && state.events().events().next().is_some() {
-            prev_batch = None;
-        }
-
         let (
             DeduplicationOutcome {
                 all_events: events,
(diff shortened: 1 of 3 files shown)


crates/matrix-sdk/src/event_cache/room/mod.rs
```diff
@@ -409,6 +399,22 @@ impl RoomEventCacheInner {
             all_duplicates,
         ) = state.collect_valid_and_duplicated_events(timeline.events).await?;
 
+        // If the timeline isn't limited, and we already knew about some past events,
+        // then this definitely know what the timeline head is (either we know
```
stefanceriu: *knows (?)

crates/matrix-sdk/src/event_cache/room/mod.rs
```diff
@@ -409,6 +399,22 @@ impl RoomEventCacheInner {
...
+        // about all the events persisted in storage, or we have a gap
+        // somewhere). In this case, we can ditch the previous-batch
+        // token, which is an optimization to avoid unnecessary future back-pagination
+        // requests.
+        //
+        // We can also ditch it, if we knew about all the events that came from sync,
```
stefanceriu: Don't need a comma after `it` as the first part cannot stand on its own as a sentence.

crates/matrix-sdk/src/event_cache/room/mod.rs
```diff
@@ -409,6 +399,22 @@ impl RoomEventCacheInner {
...
+        // viz. they were all deduplicated. In this case, using the
```
stefanceriu: What does `viz.` mean here? 🤔

crates/matrix-sdk/src/event_cache/room/mod.rs
```diff
@@ -409,6 +399,22 @@ impl RoomEventCacheInner {
...
+        // previous-batch token would only result in fetching other events we
+        // knew about. This is slightly incorrect in the presence of
+        // network splits, but this has shown to be Good Enough™.
+        if !timeline.limited && state.events().events().next().is_some() || all_duplicates {
```
stefanceriu: We should do something about these `events.events()`. I see it returns an iterator so perhaps we should call it just that `events.iterator()`.

stefanceriu (approved):
Makes sense to me, looks good 👍

crates/matrix-sdk/src/event_cache/room/mod.rs
bnjbvr: > The abbreviation `viz.` is short for the Latin videlicet, which itself is a contraction of the Latin phrase videre licet, meaning "it is permitted to see". It is used as a synonym for "namely", "that is to say", "to wit", "which is", or "as follows". 

I'll replace with `namely` :)

crates/matrix-sdk/src/event_cache/room/mod.rs
bnjbvr: Yep, not in this PR but that's something that's bugged me in the past too!

crates/matrix-sdk/src/event_cache/room/mod.rs
stefanceriu: TIL 👍

PR #5176, opened 2025-06-05
Multiverse: Initial threadding support

poljar:
This PR adds two things:

1. The ability to select timeline items
2. Rendering of threads

The first option isn't particularly useful yet, but it will be used in the future to enter a thread.

The thread rendering attempts to mimick what the Element applications do:

image

### DIFF
--- a/labs/multiverse/src/widgets/room_view/timeline.rs
+++ b/labs/multiverse/src/widgets/room_view/timeline.rs
@@ -54,6 +54,8 @@ impl StatefulWidget for &mut TimelineView<'_> {
 fn format_timeline_item(item: &Arc<TimelineItem>) -> Option<ListItem<'_>> {
     let item = match item.kind() {
         TimelineItemKind::Event(ev) => {
+            // TODO: Once the SDK allows you to filter out messages that are part of a
+            // thread, switch to that mechanism instead of manually returning `None` here.
             if ev.content().thread_root().is_some() {
                 return None;
             }
@@ -119,8 +121,10 @@ fn format_text_message(
 
         if let Some(thread_summary) = thread_summary {
             match thread_summary.latest_event {
-                TimelineDetails::Unavailable => {}
-                TimelineDetails::Pending => {}
+                TimelineDetails::Unavailable | TimelineDetails::Pending => {
+                    let thread_line = Line::from("  💬 ...");
+                    lines.push(thread_line);
+                }
                 TimelineDetails::Ready(e) => {
                     let sender = e.sender;
                     let content = e.content.as_message().map(|m| m.msgtype());


labs/multiverse/src/widgets/room_view/timeline.rs
```diff
@@ -39,20 +39,58 @@ impl StatefulWidget for &mut TimelineView<'_> {
                             kind: MsgLikeKind::Message(message),
                             ..
                         }) => {
+                            if ev.content().thread_root().is_some() {
```
bnjbvr: Can you add a `TODO: will be done in the live timeline configuration later` please?

labs/multiverse/src/widgets/room_view/timeline.rs
```diff
@@ -187,3 +50,130 @@ impl StatefulWidget for &mut TimelineView<'_> {
...
+    thread_summary: Option<ThreadSummary>,
+) -> Option<ListItem<'static>> {
+    if let MessageType::Text(text) = message.msgtype() {
+        let mut lines = Vec::new();
+        let first_line = Line::from(format!("{}: {}", sender, text.body));
+
+        lines.push(first_line);
+
+        if let Some(thread_summary) = thread_summary {
+            match thread_summary.latest_event {
```
bnjbvr, line 123: I think it'd be nice to have an indication that this message is a thread, even when the details of the latest event aren't available. Could we have a :thread: emoji next to the message? Also, could we use `format_text_message` recursively, so we get a regular text line to append after the number of replies?

poljar: > Could we have a 🧵 emoji next to the message? 

Next to the message or bellow it and show an empty space where the threaded message would be? We could even have a throbber there to indicate the pending state.

> Also, could we use format_text_message recursively, so we get a regular text line to append after the number of replies?

I'm not sure I understand what you're asking here for. Using `format_text_message` recursively for the latest thread message is part of the plan. That way we would show more than just the text messages like we do now.

But from the question it sounds like no messages are shown for you?

labs/multiverse/src/widgets/room_view/timeline.rs
poljar: Could you expand on that? What will be done in the timeline configuration? In which timeline? The ratatui widget? The SDK one?

bnjbvr: The configuration of the UI timeline builder will include a knob to indicate whether you want to filter out in-thread items by default from the (main) live timeline, or not.

labs/multiverse/src/widgets/room_view/timeline.rs
bnjbvr: > Next to the message or bellow it and show an empty space where the threaded message would be? We could even have a throbber there to indicate the pending state.

No strong opinions. We need a way to know that a message has a thread attached to it, so we can enter the thread (even when the latest event is unavailable/unknown/etc).

> Using format_text_message recursively for the latest thread message is part of the plan.

Alright, sounds like we agree on the plan :+1: (but this is not done here, right? unless i've missed it in the code)

poljar: > No strong opinions. We need a way to know that a message has a thread attached to it, so we can enter the thread (even when the latest event is unavailable/unknown/etc).

Ok, I'll keep the design the same and put on a throbber or `...`.

> Alright, sounds like we agree on the plan 👍 (but this is not done here, right? unless i've missed it in the code)

It's not, I started to refactor the giant match block but felt that the PR would grow too big if I continued on.

poljar: Added in https://github.com/matrix-org/matrix-rust-sdk/pull/5176/commits/c0b68fbb161ad248f56a36b688fbfabb590e2edc.

labs/multiverse/src/widgets/room_view/timeline.rs
poljar: Added in https://github.com/matrix-org/matrix-rust-sdk/pull/5176/commits/c0b68fbb161ad248f56a36b688fbfabb590e2edc.

Issue #4926, opened 2025-04-11
History sharing: handle the key bundle arriving *after* the invite


richvdh:
The initial implementation of [sharing encrypted history](https://github.com/element-hq/element-meta/issues/39) assumes that the key bundle will arrive before the invite. It's entirely possible that the order will be reversed, and the bundle will arrive after the invite.

PR #5177, opened 2025-06-05
Refactor the crypto store module a bit

poljar:
The cryptostore module has collected over the years a lot of storage related types.

Let's create a submodule for the pure data types and move the caches into the caches submodule.

There's also a slightly unrelated commit, I can move that into a separate PR if needed, but it's pretty trivial.

Oh, this is in preparation for #4926, since the store needs to notify us that a bundle has been received. Meaning we'll need to add more types and I didn't want to contribute to the mess further.


andybalaam (approved):
Looks good, thanks.

PR #5179, opened 2025-06-05
feat(ffi): Subscribe to a room's `RoomInfo` through `Client`

jmartinesp:
This helps in the case we want to observe the membership state changes - or some other info - for a room that's still not known so we can't just use `Client::get_room` to fetch it.

Signed-off-by:

### DIFF
--- a/bindings/matrix-sdk-ffi/src/client.rs
+++ b/bindings/matrix-sdk-ffi/src/client.rs
@@ -37,6 +37,7 @@ use matrix_sdk::{
     },
     sliding_sync::Version as SdkSlidingSyncVersion,
     store::RoomLoadSettings as SdkRoomLoadSettings,
+    sync::RoomUpdate,
     AuthApi, AuthSession, Client as MatrixClient, SessionChange, SessionTokens,
     STATE_STORE_DATABASE_NAME,
 };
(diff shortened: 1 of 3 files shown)


bindings/matrix-sdk-ffi/src/client.rs
```diff
@@ -1561,6 +1563,42 @@ impl Client {
         let max_upload_size = self.inner.load_or_fetch_max_upload_size().await?;
         Ok(max_upload_size.into())
     }
+
+    /// Subscribe to [`RoomInfo`] updates given a provided [`RoomId`].
+    ///
+    /// This works even for rooms we haven't received yet, so we can subscribe
+    /// to this and wait until we receive updates from them when sync responses
+    /// are processed.
```
bnjbvr, line 1571: Can you add about being careful to not maintain too many of these alive / that the tasks should be short-lived, and should be explicitly dropped as soon as they're not useful for the caller anymore?

jmartinesp: What's the main rationale for this, so I can explain it? Is it because keeping several of these is expensive?

bnjbvr: Yeah, performance doesn't scale if we maintained one alive task per room across the entire application's lifetime. Other mechanisms exist to receive information about *all* rooms, and should be preferred when we want to get a global view of rooms

jmartinesp: Done, thanks for the explanation!

PR #5181, opened 2025-06-05
feat(ffi): Add ability to configure indexedb instead of sqlite for wasm platforms

zzorba:
Introduces `SqliteSessionBuilder` and `IndexedDbSessionBuilder` to allow for a choice of session store.  Non-wasm platforms will only support sqlite, while Wasm platforms will support indexedb.  In order to maintain a compatible-ffi, these restrictions are handled at runtime rather than compile time.  

We also conditionally remove things related to network features (proxy, ssl validation, etc), that are not available on wasm platforms.

Signed-off-by: Daniel Salinas

### DIFF
--- a/bindings/matrix-sdk-ffi/Cargo.toml
+++ b/bindings/matrix-sdk-ffi/Cargo.toml
@@ -14,17 +14,17 @@ publish = false
 release = true
 
 [lib]
-crate-type = ["cdylib", "staticlib"]
+crate-type = ["cdylib", "staticlib", "lib"]
 
 [features]
 default = ["bundled-sqlite", "unstable-msc4274"]
 bundled-sqlite = ["matrix-sdk/bundled-sqlite"]
 unstable-msc4274 = ["matrix-sdk-ui/unstable-msc4274"]
+js = []
 
 [dependencies]
 anyhow.workspace = true
 as_variant.workspace = true
-async-compat = "0.2.4"
 extension-trait = "1.0.1"
 eyeball-im.workspace = true
 futures-util.workspace = true
@@ -36,21 +36,18 @@ matrix-sdk-ui = { workspace = true, features = ["uniffi"] }
 mime = "0.3.16"
 once_cell.workspace = true
 ruma = { workspace = true, features = ["html", "unstable-unspecified", "unstable-msc3488", "compat-unset-avatar", "unstable-msc3245-v1-compat", "unstable-msc4278"] }
-sentry-tracing = "0.36.0"
 serde.workspace = true
 serde_json.workspace = true
 thiserror.workspace = true
-tokio = { workspace = true, features = ["rt-multi-thread", "macros"] }
 tracing.workspace = true
 tracing-appender = { version = "0.2.2" }
 tracing-core.workspace = true
 tracing-subscriber = { workspace = true, features = ["env-filter"] }
-uniffi = { workspace = true, features = ["tokio"] }
 url.workspace = true
 uuid = { version = "1.4.1", features = ["v4"] }
 zeroize.workspace = true
 
-[target.'cfg(not(target_os = "android"))'.dependencies.matrix-sdk]
+[target.'cfg(all(not(target_os = "android"), not(target_family = "wasm")))'.dependencies.matrix-sdk]
 workspace = true
 features = [
     "anyhow",
@@ -64,7 +61,7 @@ features = [
     "uniffi",
 ]
 
-[target.'cfg(not(target_os = "android"))'.dependencies.sentry]
+[target.'cfg(all(not(target_os = "android"), not(target_family = "wasm")))'.dependencies.sentry]
 version = "0.36.0"
 default-features = false
 features = [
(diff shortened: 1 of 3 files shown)


jplatte:
PR title should probably make it clear that this is about the FFI crate :)

bindings/matrix-sdk-ffi/Cargo.toml
```diff
@@ -14,17 +14,17 @@ publish = false
 release = true
 
 [lib]
-crate-type = ["cdylib", "staticlib"]
+crate-type = ["cdylib", "staticlib", "lib"]
```
Hywan, line 17: I'm not sure it's related to what the patch says it does. Can you split this in another patch please?

bindings/matrix-sdk-ffi/Cargo.toml
```diff
@@ -77,6 +74,31 @@ features = [
...
+features = [
+    "anyhow",
+    "e2e-encryption",
+    "experimental-widgets",
+    "markdown",
+    "rustls-tls",
+    "socks",
+    "indexeddb",
+    "uniffi",
+]
```
Hywan: All the changes in `Cargo.toml` just be in a dedicated patch.

bindings/matrix-sdk-ffi/src/client_builder.rs
```diff
@@ -357,17 +299,17 @@ impl ClientBuilder {
...
         let mut builder = unwrap_or_clone_arc(self);
-        builder.session_paths = Some(SessionPaths { data_path, cache_path });
+        builder.paths = Some(SessionPaths { data_path, cache_path });
         Arc::new(builder)
     }
 
     /// Set the passphrase for the stores given to
-    /// [`ClientBuilder::session_paths`].
-    pub fn session_passphrase(self: Arc<Self>, passphrase: Option<String>) -> Arc<Self> {
+    /// [`ClientBuilder::paths`].
```
Hywan:
```suggestion
    /// [`SqliteSessionBuilder::paths`].
```

bindings/matrix-sdk-ffi/src/client_builder.rs
```diff
@@ -430,80 +372,171 @@ impl ClientBuilder {
         builder.system_is_memory_constrained = true;
         Arc::new(builder)
     }
+}
 
-    pub fn username(self: Arc<Self>, username: String) -> Arc<Self> {
+#[derive(Clone, uniffi::Object)]
+pub struct IndexedDbSessionBuilder {
```
Hywan, line 381: Should be put behind a `#[cfg(target_family = "wasm")]`.

bindings/matrix-sdk-ffi/src/client_builder.rs
```diff
@@ -430,80 +372,171 @@ impl ClientBuilder {
...
+    #[cfg_attr(not(target_family = "wasm"), allow(unused))]
```
Hywan, line 382: We don't use `IndexedDb` outside Wasm, all fields can be present.

bindings/matrix-sdk-ffi/src/client_builder.rs
```diff
@@ -430,80 +372,171 @@ impl ClientBuilder {
...
+#[derive(Clone)]
+pub enum ClientSessionConfig {
+    /// Setup the client to use the SQLite store.
+    #[cfg_attr(target_family = "wasm", allow(unused))]
+    Sqlite(SqliteSessionBuilder),
+
+    /// Setup the client to use the IndexedDB store.
+    #[cfg_attr(not(target_family = "wasm"), allow(unused))]
+    IndexedDb(IndexedDbSessionBuilder),
+}
```
Hywan: I think you can remove all this, and add:

```suggestion
#[cfg(target_family = "wasm")]
pub type SessionConfig = IndexedDbSessionBuilder;

#[cfg(not(target_family = "wasm"))]
pub type SessionConfig = SqliteSessionBuilder;
```

bindings/matrix-sdk-ffi/src/client_builder.rs
```diff
@@ -430,80 +372,171 @@ impl ClientBuilder {
...
+
+#[derive(Clone, uniffi::Object)]
+pub struct ClientBuilder {
+    session: Option<ClientSessionConfig>,
```
Hywan, line 418: Then…

```suggestion
    session: Option<SessionConfig>
```

bindings/matrix-sdk-ffi/src/client_builder.rs
```diff
@@ -430,80 +372,171 @@ impl ClientBuilder {
...
         Arc::new(builder)
     }
 
-    pub fn homeserver_url(self: Arc<Self>, url: String) -> Arc<Self> {
+    pub fn session_indexeddb(self: Arc<Self>, config: Arc<IndexedDbSessionBuilder>) -> Arc<Self> {
         let mut builder = unwrap_or_clone_arc(self);
-        builder.homeserver_cfg = Some(HomeserverConfig::Url(url));
+        builder.session = Some(ClientSessionConfig::IndexedDb(config.as_ref().clone()));
         Arc::new(builder)
     }
```
Hywan: And finally…

```suggestion
    pub fn session(self: Arc<Self>, config: Arc<SessionConfig>) -> Arc<Self> {
        let mut builder = unwrap_or_clone_arc(self);
        builder.session = Some(config);
        Arc::new(builder)
    }
```

Hywan (changes requested):
Thanks for the patch.

First off, I believe you should split all the modifications from the `Cargo.toml` file into its own patch.

Second, I think you can simplify things by introducing a type alias `SessionConfig`. We don't want to expose SQLite or IndexedDb directly, we want to hide that from the consumer/user. See my suggestions.

What do you think?

bindings/matrix-sdk-ffi/Cargo.toml
jplatte: Yeah this seems weird, what's the idea with this?

zzorba: The uniffi bindings generator requires the library to be built as a "lib" so it can be included in a second generated-rust library that is fed through.  But I will break that out into a second PR.

bindings/matrix-sdk-ffi/src/client_builder.rs
zzorba: So, if you are sharing bindings between two JS platforms (react-native ios/android + web), it is very convenient if they have the same interface.  The typescript system cannot differentiate between bindings in a *.web.ts file and a *.ts file, it assumes they will share an API).

That's the reason I've gone with runtime guards on behavior, to make sure we can generate unified uniffi bindings between the targets.  I could put this behind a feature flag if you'd like, but I think it will be necessary for the maintenance of the react-native bindings (which will include ios/android/web).

bindings/matrix-sdk-ffi/Cargo.toml
zzorba: Opened up https://github.com/matrix-org/matrix-rust-sdk/pull/5211 which is nearly merged.  I will rebase this branch with those changes (and other feedback) once it lands.

zzorba:
Going to close this in favor of https://github.com/matrix-org/matrix-rust-sdk/pull/5211 and https://github.com/matrix-org/matrix-rust-sdk/pull/5245

Issue #5122, opened 2025-05-28
threads: persistent storage


bnjbvr:
Make it possible to save threads events in the event cache (even if they haven't been received from sync). We also want to persist and restore a thread's summary to/from persisted storage.

- [x] Generalize the SQLite backend to hold thread linked chunk. It might be sufficient that a linked chunk's key isn't a room anymore, but a generalized `enum LinkedChunkKey { Room(OwnedRoomId), Thread(OwnedRoomid, OwnedEventId) }`. Migration = clear all the chunks?
  - #5182. 
- [x] Add a table in storage to store thread summaries separately from the root event; or instead of storing `TimelineEvent` into storage, store an augmented data structure that also includes a `ThreadSummary` (⚠ this second alternative would require a migration).
  - [x] Add a method in the event cache store to `upsert_thread_summary`
  - Keep `participated` in clear, so that we can query for all the threads we've participated to.
  - Status (2025-06-09): Done differently: we're storing the `ThreadSummary` in the `TimelineEvent`, which is stored in the persistent storage.
- [ ] For each event that had more than one edit (>1 or >=1 TBD), introduce an edit linked list, to order them relative to each other, and figure out which is the actual latest.
- What to do after a gappy sync?
  - Thread linked chunk
    - In a non-persisted world, we can clear the linked chunk of a gappy thread immediately. 
    - ~~Upon gappy sync, do a `/threads` query to retrieve all the threads and their latest event; if it matches what we knew from the thread linked chunk, mark it as NOT `maybe_outdated`. If not, start back-pagination until all the thread events are known.~~
      - This has been deemed impractical, because `/threads` can be super slow for large rooms like MatrixHQ. Instead, we could, for each *known* thread, run a single `/relations` query from the back, and keep on back-paginating until we know all the events; or let the embedder manually request to back-paginate, like we do for live main timelines.
      - Likely recompute the thread summary afterwards 👀 (Think about conflicts with sync.)

Part of #4869

Hywan:
Closed by #6317 (that was merged in the `features/event-cache-refactoring` branch, part of the https://github.com/matrix-org/matrix-rust-sdk/issues/6152 issue, which has been merged in `main` in https://github.com/matrix-org/matrix-rust-sdk/pull/6671).

PR #5182, opened 2025-06-05
refactor!(event cache): introduce `LinkedChunkId` in the backends

bnjbvr:
In a "soon" future, threads have their own linked chunk. All our code has been written with the fact that a linked chunk belong to *a room* in mind, so it needs some biggish update. Fortunately, most of the changes are mechanical, so they should be rather easy to review.

Part of #4869, namely #5122.

### DIFF
--- a/crates/matrix-sdk-base/src/event_cache/store/memory_store.rs
+++ b/crates/matrix-sdk-base/src/event_cache/store/memory_store.rs
@@ -21,8 +21,8 @@ use std::{
 use async_trait::async_trait;
 use matrix_sdk_common::{
     linked_chunk::{
-        relational::RelationalLinkedChunk, ChunkIdentifier, ChunkIdentifierGenerator, Position,
-        RawChunk, Update,
+        relational::RelationalLinkedChunk, ChunkIdentifier, ChunkIdentifierGenerator,
+        LinkedChunkId, Position, RawChunk, Update,
     },
     ring_buffer::RingBuffer,
     store_locks::memory_store_helper::try_take_leased_lock,
@@ -128,46 +128,46 @@ impl EventCacheStore for MemoryStore {
 
     async fn handle_linked_chunk_updates(
         &self,
-        room_id: &RoomId,
+        lcid: LinkedChunkId<'_>,
         updates: Vec<Update<Event, Gap>>,
     ) -> Result<(), Self::Error> {
         let mut inner = self.inner.write().unwrap();
-        inner.events.apply_updates(room_id, updates);
+        inner.events.apply_updates(lcid, updates);
 
         Ok(())
     }
 
     async fn load_all_chunks(
         &self,
-        room_id: &RoomId,
+        lcid: LinkedChunkId<'_>,
     ) -> Result<Vec<RawChunk<Event, Gap>>, Self::Error> {
         let inner = self.inner.read().unwrap();
         inner
             .events
-            .load_all_chunks(room_id)
+            .load_all_chunks(lcid)
             .map_err(|err| EventCacheStoreError::InvalidData { details: err })
     }
 
     async fn load_last_chunk(
         &self,
-        room_id: &RoomId,
+        lcid: LinkedChunkId<'_>,
     ) -> Result<(Option<RawChunk<Event, Gap>>, ChunkIdentifierGenerator), Self::Error> {
         let inner = self.inner.read().unwrap();
         inner
             .events
-            .load_last_chunk(room_id)
+            .load_last_chunk(lcid)
             .map_err(|err| EventCacheStoreError::InvalidData { details: err })
     }
 
     async fn load_previous_chunk(
         &self,
-        room_id: &RoomId,
+        lcid: LinkedChunkId<'_>,
         before_chunk_identifier: ChunkIdentifier,
     ) -> Result<Option<RawChunk<Event, Gap>>, Self::Error> {
         let inner = self.inner.read().unwrap();
         inner
             .events
-            .load_previous_chunk(room_id, before_chunk_identifier)
+            .load_previous_chunk(lcid, before_chunk_identifier)
             .map_err(|err| EventCacheStoreError::InvalidData { details: err })
     }
 
@@ -178,7 +178,7 @@ impl EventCacheStore for MemoryStore {
 
     async fn filter_duplicated_events(
         &self,
-        room_id: &RoomId,
+        lcid: LinkedChunkId<'_>,
         mut events: Vec<OwnedEventId>,
     ) -> Result<Vec<(OwnedEventId, Position)>, Self::Error> {
         // Collect all duplicated events.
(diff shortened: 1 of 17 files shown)


crates/matrix-sdk-base/src/event_cache/store/traits.rs
```diff
@@ -105,20 +107,22 @@ pub trait EventCacheStore: AsyncTraitDeps {
     /// ⚠ This is meant only for super specific use cases, where there shouldn't
     /// be any live in-memory linked chunks. In general, prefer using
     /// `EventCache::clear_all_rooms()` from the common SDK crate.
+    // TODO(bnjbvr): rename to `clear_all_linked_chunks`, since it's going to also
```
bnjbvr: Remaining todo here

crates/matrix-sdk-base/src/event_cache/store/traits.rs
```diff
@@ -105,20 +107,22 @@ pub trait EventCacheStore: AsyncTraitDeps {
...
+        lcid: LinkedChunkId<'_>,
         events: Vec<OwnedEventId>,
     ) -> Result<Vec<(OwnedEventId, Position)>, Self::Error>;
 
-    /// Find an event by its ID.
+    /// Find an event by its ID in a linked chunk.
     async fn find_event(
         &self,
-        room_id: &RoomId,
+        lcid: LinkedChunkId<'_>,
```
bnjbvr: Thinking about it, it might be better keyed on the `&RoomId`, to reflect that events are saved per room (not per linked chunk).

crates/matrix-sdk-base/src/event_cache/store/traits.rs
```diff
@@ -133,7 +137,7 @@ pub trait EventCacheStore: AsyncTraitDeps {
     /// a certain relationship.
     async fn find_event_relations(
         &self,
-        room_id: &RoomId,
+        lcid: LinkedChunkId<'_>,
```
bnjbvr: ditto

crates/matrix-sdk-common/src/linked_chunk/mod.rs
```diff
@@ -97,15 +97,114 @@ pub mod relational;
...
+    // TODO(bnjbvr): Soon™.
+    // Thread(&'a RoomId, &'a EventId),
+}
+
+impl AsRef<[u8]> for LinkedChunkId<'_> {
+    fn as_ref(&self) -> &[u8] {
+        match self {
+            LinkedChunkId::Room(room_id) => room_id.as_ref(),
+            // ouch, good luck future me for handling thread keys here
+        }
```
bnjbvr: Probably better to rename `persistent_key()` or something like this and have it return `impl AsRef<[u8]>`, so that the thread variant can return a `String` formed by concatenating the room and thread ids.

(Alternative: for the room variant, return something like `room:{room_id}`, but it's not really necessary)

crates/matrix-sdk-common/src/linked_chunk/mod.rs
```diff
@@ -97,15 +97,114 @@ pub mod relational;
...
+        match (self, other) {
+            (LinkedChunkId::Room(a), OwnedLinkedChunkId::Room(b)) => (*a).partial_cmp(b),
+        }
+    }
+}
+
+impl PartialEq<LinkedChunkId<'_>> for OwnedLinkedChunkId {
+    fn eq(&self, other: &LinkedChunkId<'_>) -> bool {
+        match (self, other) {
+            (OwnedLinkedChunkId::Room(a), LinkedChunkId::Room(b)) => a == *b,
```
bnjbvr: Can be implemented in terms of the other `PartialEq` impl, likely

crates/matrix-sdk-common/src/linked_chunk/mod.rs
```diff
@@ -97,15 +97,114 @@ pub mod relational;
...
+        }
+    }
+}
+
+impl PartialOrd<LinkedChunkId<'_>> for OwnedLinkedChunkId {
+    fn partial_cmp(&self, other: &LinkedChunkId<'_>) -> Option<cmp::Ordering> {
```
bnjbvr: Maybe ditto? (Would need to revert the `Ordering`)

crates/matrix-sdk-common/src/linked_chunk/mod.rs
```diff
@@ -97,15 +97,114 @@ pub mod relational;
...
+    Room(OwnedRoomId),
+    // TODO(bnjbvr): Soon™.
+    // Thread(RoomId, EventId),
+}
+
+impl Display for OwnedLinkedChunkId {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            OwnedLinkedChunkId::Room(room_id) => write!(f, "{room_id}"),
+            // ouch, good luck future me for handling thread keys here
```
bnjbvr: lolwat

bnjbvr (commented):
(Self-review, since I thought about the code this morning…)

crates/matrix-sdk-common/src/linked_chunk/mod.rs
bnjbvr: even better: it's actually unused, so let's get rid of it!

crates/matrix-sdk-common/src/linked_chunk/relational.rs
```diff
@@ -335,15 +369,15 @@ where
 
     /// Return an iterator that yields items of a particular room, in no
```
poljar: Might make sense to update the doc as well, it's still talking about a room.

crates/matrix-sdk-common/src/linked_chunk/relational.rs
```diff
@@ -183,18 +194,23 @@ where
...
                     let mut entry_to_remove = None;
 
-                    for (nth, ItemRow { room_id: room_id_candidate, position, .. }) in
-                        self.items_chunks.iter_mut().enumerate()
+                    for (
+                        nth,
+                        ItemRow { linked_chunk_id: linked_chunk_id_candidate, position, .. },
+                    ) in self.items_chunks.iter_mut().enumerate()
                     {
                         // Filter by room ID.
```
poljar: We're not filtering by room ID anymore.

crates/matrix-sdk-base/src/event_cache/store/memory_store.rs
```diff
@@ -128,46 +128,46 @@ impl EventCacheStore for MemoryStore {
 
     async fn handle_linked_chunk_updates(
         &self,
-        room_id: &RoomId,
+        lcid: LinkedChunkId<'_>,
```
poljar: Can we rename this to `chunk_id` or `linked_chunk_id` like the rest of the patch does?

`lcid` is a bit opaque on first sight.

poljar (approved):
I left a couple of nits. This looks mostly good.

I guess you'll squash merge this PR since individual commits don't build on their own?

My main gripe here is that we now have two things that are called a chunk identifier.

crates/matrix-sdk-base/src/event_cache/store/memory_store.rs
bnjbvr: Sure!

bnjbvr:
Thanks for the review! Good point about the new dual `ChunkId` vs `LinkedChunkId`; probably we should rename one of these to something closer to linked list lingo. Not doing it in this PR here, we can discuss this as a group later.

PR #5089, opened 2025-05-23
feat(wasm): Wasm equivalent of get_runtime_handle and corresponding tokio types

zzorba:
Adds a Wasm equivalent of the get_runtime_handle method provided by tokio, as well as Handle/Runtime types that can be used on either Wasm or non-Wasm platforms interchangeably.

Dependent on https://github.com/matrix-org/matrix-rust-sdk/pull/5088

- [ ] Public API changes documented in changelogs (optional)

Signed-off-by: Daniel Salinas

PR #5184, opened 2025-06-05
fix: Move runtime module from matrix-sdk-common to matrix-sdk-ffi

jplatte:
This module only builds on non-wasm with the patched async-compat from the workspace Cargo.toml's patch section, and it is only used by the ffi crate. It is currently breaking the use of the SDK as a git dependency, and would prevent the publishing of matrix-sdk-common (unless using --no-verify, but then that would just break all users of the newly published crates.io version).

This bug was introduced in https://github.com/matrix-org/matrix-rust-sdk/pull/5089.


poljar (approved):
Hah, thanks for the fix. That would have been annoying to deal with.

PR #5185, opened 2025-06-06
timeline: Thread filtering

stefanceriu:
This set of patches introduces options for filtering out threaded messages from the main timelines. In order to do so it:
* moves relation processing from `TimelineAction` to `TimelineMetadata` so that they're available beforehand and can be used in `TimelineStateTransaction::handle_remote_event` and `TimelineState::handle_local_event`
* uses the thread root in the `TimelineStateTransaction::should_add_event_item` and the newly introduced `TimelineState::should_add_event_item`
* adds new `hide_threaded_events` associated values to the `Live` and `Event` focuses so that the old behavior can be kept until clients migrate 

Relates to https://github.com/element-hq/element-meta/issues/2771 and https://github.com/matrix-org/matrix-rust-sdk/issues/4833

### DIFF
--- a/bindings/matrix-sdk-ffi/src/timeline/configuration.rs
+++ b/bindings/matrix-sdk-ffi/src/timeline/configuration.rs
@@ -63,10 +63,28 @@ impl From<FilterTimelineEventType> for TimelineEventType {
 
 #[derive(uniffi::Enum)]
 pub enum TimelineFocus {
-    Live { hide_threaded_events: bool },
-    Event { event_id: String, num_context_events: u16, hide_threaded_events: bool },
-    Thread { root_event_id: String, num_events: u16 },
-    PinnedEvents { max_events_to_load: u16, max_concurrent_requests: u16 },
+    Live {
+        /// Whether to hide in-thread replies from the live timeline.
+        hide_threaded_events: bool,
+    },
+    Event {
+        /// The initial event to focus on. This is usually the target of a
+        /// permalink.
+        event_id: String,
+        /// The number of context events to load around the focused event.
+        num_context_events: u16,
+        /// Whether to hide in-thread replies from the live timeline.
+        hide_threaded_events: bool,
+    },
+    Thread {
+        /// The thread root event ID to focus on.
+        root_event_id: String,
+        num_events: u16,
+    },
+    PinnedEvents {
+        max_events_to_load: u16,
+        max_concurrent_requests: u16,
+    },
 }
 
 impl TryFrom<TimelineFocus> for matrix_sdk_ui::timeline::TimelineFocus {
--- a/bindings/matrix-sdk-ffi/src/timeline/configuration.rs
+++ b/bindings/matrix-sdk-ffi/src/timeline/configuration.rs
@@ -63,8 +63,8 @@ impl From<FilterTimelineEventType> for TimelineEventType {
 
 #[derive(uniffi::Enum)]
 pub enum TimelineFocus {
-    Live,
-    Event { event_id: String, num_context_events: u16 },
+    Live { hide_threaded_events: bool },
+    Event { event_id: String, num_context_events: u16, hide_threaded_events: bool },
     Thread { root_event_id: String, num_events: u16 },
     PinnedEvents { max_events_to_load: u16, max_concurrent_requests: u16 },
 }
@@ -76,15 +76,19 @@ impl TryFrom<TimelineFocus> for matrix_sdk_ui::timeline::TimelineFocus {
         value: TimelineFocus,
     ) -> Result<matrix_sdk_ui::timeline::TimelineFocus, Self::Error> {
         match value {
-            TimelineFocus::Live => Ok(Self::Live),
-            TimelineFocus::Event { event_id, num_context_events } => {
+            TimelineFocus::Live { hide_threaded_events } => Ok(Self::Live { hide_threaded_events }),
+            TimelineFocus::Event { event_id, num_context_events, hide_threaded_events } => {
                 let parsed_event_id =
                     EventId::parse(&event_id).map_err(|err| FocusEventError::InvalidEventId {
                         event_id: event_id.clone(),
                         err: err.to_string(),
                     })?;
 
-                Ok(Self::Event { target: parsed_event_id, num_context_events })
+                Ok(Self::Event {
+                    target: parsed_event_id,
+                    num_context_events,
+                    hide_threaded_events,
+                })
             }
             TimelineFocus::Thread { root_event_id, num_events } => {
                 let parsed_root_event_id = EventId::parse(&root_event_id).map_err(|err| {
--- a/crates/matrix-sdk-ui/src/timeline/controller/metadata.rs
+++ b/crates/matrix-sdk-ui/src/timeline/controller/metadata.rs
@@ -17,15 +17,44 @@ use std::{
     sync::Arc,
 };
 
-use ruma::{EventId, OwnedEventId, OwnedUserId, RoomVersionId};
+use imbl::Vector;
+use matrix_sdk::deserialized_responses::EncryptionInfo;
+use ruma::{
+    events::{
+        poll::unstable_start::UnstablePollStartEventContent, relation::Replacement,
+        room::message::RelationWithoutReplacement, AnyMessageLikeEventContent,
+        AnySyncMessageLikeEvent, AnySyncTimelineEvent, BundledMessageLikeRelations,
+    },
+    serde::Raw,
+    EventId, OwnedEventId, OwnedUserId, RoomVersionId,
+};
 use tracing::trace;
 
 use super::{
     super::{subscriber::skip::SkipCount, TimelineItem, TimelineItemKind, TimelineUniqueId},
     read_receipts::ReadReceipts,
-    Aggregations, AllRemoteEvents, ObservableItemsTransaction,
+    Aggregation, AggregationKind, Aggregations, AllRemoteEvents, ObservableItemsTransaction,
+    PendingEdit, PendingEditKind,
 };
-use crate::unable_to_decrypt_hook::UtdHookManager;
+use crate::{
+    timeline::{
+        event_item::{
+            extract_bundled_edit_event_json, extract_poll_edit_content,
+            extract_room_msg_edit_content,
+        },
+        InReplyToDetails, TimelineEventItemId,
+    },
+    unable_to_decrypt_hook::UtdHookManager,
+};
+
+/// All parameters to [`TimelineAction::from_content`] that only apply if an
+/// event is a remote echo.
+pub struct RemoteEventContext<'a> {
+    pub event_id: &'a EventId,
+    pub raw_event: &'a Raw<AnySyncTimelineEvent>,
+    pub relations: BundledMessageLikeRelations<AnySyncMessageLikeEvent>,
+    pub bundled_edit_encryption_info: Option<Arc<EncryptionInfo>>,
+}
 
 #[derive(Clone, Debug)]
 pub(in crate::timeline) struct TimelineMetadata {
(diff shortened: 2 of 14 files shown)


crates/matrix-sdk-ui/src/timeline/controller/metadata.rs
```diff
@@ -278,6 +307,140 @@ impl TimelineMetadata {
             }
         }
     }
+
+    pub(crate) fn process_content_relations(
```
bnjbvr, line 339: We should add some doc comment explaining what is the returned `Option<OwnedEventId>` here.

crates/matrix-sdk-ui/src/timeline/controller/metadata.rs
```diff
@@ -278,6 +307,140 @@ impl TimelineMetadata {
...
+                }
+
+                (in_reply_to, thread_root)
+            }
+
+            _ => (None, None),
+        }
+    }
+
+    fn extract_reply_and_thread_root(
```
bnjbvr, line 441: Also doc comment here and in `mark_response`

crates/matrix-sdk-ui/src/timeline/controller/metadata.rs
```diff
@@ -278,6 +307,140 @@ impl TimelineMetadata {
...
+            )) => {
+                let (in_reply_to, thread_root) =
+                    Self::extract_reply_and_thread_root(c.relates_to.clone(), timeline_items);
+
+                // Record the bundled edit in the aggregations set, if any.
+                if let Some(ctx) = remote_ctx {
+                    if let Some(new_content) = extract_poll_edit_content(ctx.relations) {
+                        // It is replacing the current event.
+                        if let Some(edit_event_id) =
+                            ctx.raw_event.get_field::<OwnedEventId>("event_id").ok().flatten()
```
bnjbvr, line 370: Hmm preexisting, but I do realize this is incorrect: this should be the event id of the edit event, so the one in `ctx.relations`. It's likely harmless, unless someone starts redacting edits… I will open a followup issue, because I don't think this happens in real-world clients.

crates/matrix-sdk-ui/src/timeline/controller/metadata.rs
```diff
@@ -17,15 +17,44 @@ use std::{
...
+        },
+        InReplyToDetails, TimelineEventItemId,
+    },
+    unable_to_decrypt_hook::UtdHookManager,
 };
-use crate::unable_to_decrypt_hook::UtdHookManager;
+
+/// All parameters to [`TimelineAction::from_content`] that only apply if an
+/// event is a remote echo.
+pub struct RemoteEventContext<'a> {
```
bnjbvr: Let's keep it `pub(crate)` at best.

crates/matrix-sdk-ui/src/timeline/controller/state_transaction.rs
```diff
@@ -619,6 +619,13 @@ impl<'a> TimelineStateTransaction<'a> {
         {
             // Classical path: the event is valid, can be deserialized, everything is alright.
             Ok(event) => {
+                let (in_reply_to, thread_root) = self.meta.process_event_relations(
+                    event.clone(),
```
bnjbvr: Let's avoid the cloning here by having `process_event_relations` take a `&Any…`

crates/matrix-sdk-ui/src/timeline/event_item/content/reply.rs
```diff
@@ -113,16 +113,23 @@ impl EmbeddedEvent {
         // We don't need to fill the thread information of an embedded reply.
         let thread_summary = None;
 
+        let (in_reply_to, thread_root) = meta.process_event_relations(
+            event.clone(),
```
bnjbvr: ditto here

crates/matrix-sdk-ui/src/timeline/controller/mod.rs
```diff
@@ -1349,15 +1349,9 @@ impl<P: RoomDataProvider, D: Decryptor> TimelineController<P, D> {
     ) -> Result<Option<EmbeddedEvent>, Error> {
         // Reborrow, to avoid that the automatic deref borrows the entire guard (and we
         // can't borrow both items and meta).
-        let state = &mut *self.state.write().await;
+        let state = &*self.state.write().await;
```
bnjbvr: nit: `read()` \o/ and can likely remove the comment and reborrow

crates/matrix-sdk-ui/src/timeline/controller/mod.rs
```diff
@@ -1587,15 +1581,10 @@ async fn fetch_replied_to_event(
     trace!("Fetching replied-to event");
     let res = match room.load_or_fetch_event(in_reply_to, None).await {
         Ok(timeline_event) => {
-            let state = &mut *state_lock.write().await;
+            let state = &*state_lock.write().await;
```
bnjbvr: ditto

crates/matrix-sdk-ui/src/timeline/controller/state_transaction.rs
```diff
@@ -393,6 +393,7 @@ impl<'a> TimelineStateTransaction<'a> {
         room_data_provider: &P,
         settings: &TimelineSettings,
         event: &AnySyncTimelineEvent,
+        thread_root: &Option<OwnedEventId>,
```
bnjbvr: nit: it's usually slightly more idiomatic to use `Option<&EventId>` (i.e. move the reference inside the option), since niche optimizations could kick in in this case (and not with `&Option<...>`).

crates/matrix-sdk-ui/src/timeline/controller/state_transaction.rs
```diff
@@ -401,13 +402,17 @@ impl<'a> TimelineStateTransaction<'a> {
...
 
-        match self.timeline_focus {
+        match &self.timeline_focus {
             TimelineFocusKind::PinnedEvents => {
                 // Only add pinned events for the pinned events timeline.
                 room_data_provider.is_pinned_event(event.event_id())
             }
 
             TimelineFocusKind::Event => {
+                if thread_root.is_some() {
```
bnjbvr: Let's add a tiny comment explaining what's going on here.

crates/matrix-sdk-ui/src/timeline/controller/state_transaction.rs
```diff
@@ -429,16 +434,12 @@ impl<'a> TimelineStateTransaction<'a> {
...
             }
 
-            TimelineFocusKind::Thread => {
-                // The thread timeline doesn't apply any additional
-                // for now. It will however do so in the future, as
-                // will the live one
-                true
+            TimelineFocusKind::Thread { root_event_id } => {
+                event.event_id() == root_event_id
+                    || thread_root.as_ref().is_some_and(|r| r == root_event_id)
```
bnjbvr, line 447: Unfortunately, I think this is slightly incorrect in the presence of aggregations applying to in-thread events, but we can deal with this later.

crates/matrix-sdk-ui/src/timeline/controller/state.rs
```diff
@@ -160,11 +160,16 @@ impl TimelineState {
         send_handle: Option<SendHandle>,
         content: AnyMessageLikeEventContent,
     ) {
-        // Only add new items if the timeline is live.
-        let should_add_new_items = match self.timeline_focus {
-            TimelineFocusKind::Live | TimelineFocusKind::Thread { .. } => true,
-            TimelineFocusKind::Event | TimelineFocusKind::PinnedEvents => false,
-        };
+        let timeline_focus = self.timeline_focus.clone();
```
bnjbvr: Probably a spurious clone?

crates/matrix-sdk-ui/src/timeline/controller/state.rs
```diff
@@ -196,6 +194,21 @@ impl TimelineState {
...
+        match &timeline_focus {
+            TimelineFocusKind::Live => thread_root.is_none(),
+
+            TimelineFocusKind::Thread { root_event_id } => {
+                thread_root.as_ref().is_some_and(|r| r == root_event_id)
+            }
+
+            TimelineFocusKind::Event | TimelineFocusKind::PinnedEvents => false,
+        }
+    }
```
bnjbvr: I'm a bit worried that we have an undocumented function, that partially replicates some behavior of `Transaction::should_add_event_item`; let's see if we can share more code here.

crates/matrix-sdk-ui/src/timeline/controller/state.rs
```diff
@@ -196,6 +194,21 @@ impl TimelineState {
...
+        thread_root: &Option<OwnedEventId>,
+    ) -> bool {
+        match &timeline_focus {
+            TimelineFocusKind::Live => thread_root.is_none(),
+
+            TimelineFocusKind::Thread { root_event_id } => {
+                thread_root.as_ref().is_some_and(|r| r == root_event_id)
+            }
+
+            TimelineFocusKind::Event | TimelineFocusKind::PinnedEvents => false,
```
bnjbvr: This means that local echoes for aggregations won't apply to those focus kinds, which is likely wrong.

crates/matrix-sdk-ui/src/timeline/controller/state.rs
```diff
@@ -196,6 +194,21 @@ impl TimelineState {
         txn.commit();
     }
 
+    fn should_add_event_item(
+        timeline_focus: &TimelineFocusKind,
+        thread_root: &Option<OwnedEventId>,
```
bnjbvr: nit: use `Option<&` here too

bindings/matrix-sdk-ffi/src/timeline/configuration.rs
```diff
@@ -63,8 +63,8 @@ impl From<FilterTimelineEventType> for TimelineEventType {
 
 #[derive(uniffi::Enum)]
 pub enum TimelineFocus {
-    Live,
-    Event { event_id: String, num_context_events: u16 },
+    Live { hide_threaded_events: bool },
```
bnjbvr: Let's add a small nice comment here :)

bindings/matrix-sdk-ffi/src/timeline/configuration.rs
```diff
@@ -63,8 +63,8 @@ impl From<FilterTimelineEventType> for TimelineEventType {
...
+    Event { event_id: String, num_context_events: u16, hide_threaded_events: bool },
```
bnjbvr: and here too

crates/matrix-sdk-ui/src/timeline/controller/state.rs
```diff
@@ -199,13 +199,15 @@ impl TimelineState {
         thread_root: &Option<OwnedEventId>,
     ) -> bool {
         match &timeline_focus {
-            TimelineFocusKind::Live => thread_root.is_none(),
+            TimelineFocusKind::Live { hide_threaded_events } => {
+                thread_root.is_none() || thread_root.is_some() && !hide_threaded_events
```
bnjbvr: In the right hand side, we can remove `thread_root.is_some()`, since it's implied from the negation of the LHS. Boole's logic ftw!

crates/matrix-sdk-ui/src/timeline/mod.rs
```diff
@@ -121,10 +121,10 @@ pub struct Timeline {
 pub enum TimelineFocus {
     /// Focus on live events, i.e. receive events from sync and append them in
     /// real-time.
-    Live,
+    Live { hide_threaded_events: bool },
```
bnjbvr: Let's add a doc comment here too.

crates/matrix-sdk-ui/src/timeline/mod.rs
```diff
@@ -121,10 +121,10 @@ pub struct Timeline {
...
 
     /// Focus on a specific event, e.g. after clicking a permalink.
-    Event { target: OwnedEventId, num_context_events: u16 },
+    Event { target: OwnedEventId, num_context_events: u16, hide_threaded_events: bool },
```
bnjbvr: And here too

crates/matrix-sdk-ui/tests/integration/timeline/thread.rs
```diff
@@ -97,12 +97,32 @@ async fn test_thread_backpagination() {
         .await;
 
     let batch1 = vec![
-        factory.text_msg("Threaded event 4").event_id(event_id!("$3")).into_raw_sync().cast(),
-        factory.text_msg("Threaded event 3").event_id(event_id!("$4")).into_raw_sync().cast(),
+        factory
+            .text_msg("Threaded event 4")
+            .event_id(event_id!("$3"))
+            .in_thread(&thread_root_event_id, event_id!("$latest"))
```
bnjbvr: Fwiw, the second event ids are all incorrect, as this is the one used for the reply fallback, it should be set to the previous in-thread event, each time.

crates/matrix-sdk-ui/tests/integration/timeline/thread.rs
```diff
@@ -399,3 +419,75 @@ async fn test_new_thread_reply_causes_thread_summary() {
...
+                        .text_msg("Thread root")
+                        .sender(sender_id)
+                        .event_id(&thread_root_event_id),
+                )
+                .add_timeline_event(
+                    factory
+                        .text_msg("Within thread")
+                        .sender(sender_id)
+                        .event_id(event_id!("$threaded_event"))
+                        .in_thread(&thread_root_event_id, event_id!("$first_reply_in_thread")),
```
bnjbvr: second event id should be `thread_root_event_id` too (reply fallback)

crates/matrix-sdk-ui/tests/integration/timeline/thread.rs
```diff
@@ -399,3 +419,75 @@ async fn test_new_thread_reply_causes_thread_summary() {
...
+        .await;
+
+    filtered_timeline_stream.next().await;
+
+    let items = timeline.items().await;
+
+    // A thread filtered timeline should only contain the date separator and the
+    // thread root.
+    assert!(items[0].is_date_divider());
+    assert_eq!(items[1].as_event().unwrap().content().as_message().unwrap().body(), "Thread root");
```
bnjbvr: Instead of using this pattern, usually we'd do the following:

- assert let with a timeout on the stream
- run assertions directly on the stream update

crates/matrix-sdk-ui/tests/integration/timeline/thread.rs
```diff
@@ -399,3 +419,75 @@ async fn test_new_thread_reply_causes_thread_summary() {
...
+
+    let items = timeline.items().await;
+
+    // A non-thread filtered timeline should contain all the items.
+    assert!(items[0].is_date_divider());
+    assert_eq!(items[1].as_event().unwrap().content().as_message().unwrap().body(), "Thread root");
+    assert_eq!(
+        items[2].as_event().unwrap().content().as_message().unwrap().body(),
+        "Within thread"
+    );
```
bnjbvr: And here as well

crates/matrix-sdk-ui/src/timeline/controller/state_transaction.rs
bnjbvr: Or maybe not! aggregations don't care about the result of `should_add_event_item` \o/

crates/matrix-sdk-ui/src/timeline/controller/state.rs
bnjbvr: (there's `txn.timeline_focus` that can be used instead)

crates/matrix-sdk-ui/src/timeline/controller/state.rs
bnjbvr: errr, *I'm* wrong! aggregations are handled separately.

PR #5186, opened 2025-06-06
fix(base): Revisit `check_tombstone` entirely

Hywan:
This patch renames `check_tombstone` to `check_room_upgrades`. Then it rewrites it **entirely** to remove all false-positives and false-negatives. More importantly, the room versions are no longer involved: they can't be compared or ordered, they must be treated as opaque values.

This new version of `check_room_upgrades` does a first path to check predecessor-successor consistency. Then it does a second version to detect loops. This new algorithm is robust to absent `m.room.create` events. Making them mandatory is left to another patch.

More tests are added, especially to ensure that `m.room.create` cannot be overwritten, and to ensure loops or inconsistent predecessors and successors are correctly detected.

---

- Sequel of https://github.com/matrix-org/matrix-rust-sdk/pull/5150


pixlwave (commented):
4aa048a2e8477b38e2dc442c2f5117d0d9c305a30f2d64e1ebc4c52f9792ecf7-1985628428

PR #5187, opened 2025-06-06
refactor(ffi): Improve `Client::subscribe_to_room_info`

jmartinesp:
Use `Client::room_info_notable_update_receiver` instead of `RoomUpdates`

This allows us to also react to local changes that would take several extra seconds to be received in a new sync response.

@ganfra found out this problem when testing this with unknown rooms he requested to join, the response using the previous implementation took way longer than it was needed. This new one behaves way better.

Signed-off-by:

### DIFF
--- a/bindings/matrix-sdk-ffi/src/client.rs
+++ b/bindings/matrix-sdk-ffi/src/client.rs
@@ -37,7 +37,6 @@ use matrix_sdk::{
     },
     sliding_sync::Version as SdkSlidingSyncVersion,
     store::RoomLoadSettings as SdkRoomLoadSettings,
-    sync::RoomUpdate,
     AuthApi, AuthSession, Client as MatrixClient, SessionChange, SessionTokens,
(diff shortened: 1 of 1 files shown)


bindings/matrix-sdk-ffi/src/client.rs
```diff
@@ -1588,17 +1586,20 @@ impl Client {
...
+                while let Ok(room_update) = receiver.recv().await {
+                    if room_update.room_id != room_id {
+                        continue;
+                    }
+
+                    if let Some(room) = client.get_room(&room_id) {
+                        if let Ok(room_info) = RoomInfo::new(&room).await {
+                            listener.call(room_info);
+                        }
+                    }
```
poljar, line 1602: Might make sense to ponder if we can remove `subscribe_to_room_updates()` if we can achieve the same with the notable receiver.

I guess the notable one is sliding sync specific.

PR #5188, opened 2025-06-06
feat(multiverse): Show read receipts for individual messages

poljar:
I'm not really sure how we want to display the read receipts. I put them in the details view of the read receipts tab.

We can move them around or create a separate tab for them if that's preferable.

image

### DIFF
--- a/labs/multiverse/src/widgets/room_view/details/read_receipts.rs
+++ b/labs/multiverse/src/widgets/room_view/details/read_receipts.rs
@@ -3,10 +3,10 @@ use matrix_sdk_base::read_receipts::RoomReadReceipts;
 use matrix_sdk_ui::timeline::TimelineItem;
 use ratatui::{
     prelude::*,
-    widgets::{Block, Borders, HighlightSpacing, List, ListItem, ListState, Paragraph, Wrap},
+    widgets::{Block, Borders, Paragraph, Wrap},
 };
 
-use crate::{widgets::room_view::DetailsState, SELECTED_STYLE_FG, TEXT_COLOR};
+use crate::{widgets::room_view::DetailsState, TEXT_COLOR};
 
 pub struct ReadReceipts<'a> {
     state: &'a DetailsState<'a>,
@@ -47,14 +47,11 @@ impl Widget for &mut ReadReceipts<'_> {
             match self.state.selected_item.as_deref() {
                 Some(selected_event) => {
                     if let Some(item) = format_timeline_item(selected_event) {
-                        let list = List::new(vec![item])
-                            .highlight_spacing(HighlightSpacing::Always)
-                            .highlight_symbol(">")
-                            .highlight_style(SELECTED_STYLE_FG);
-
-                        let mut state = ListState::default();
-
-                        StatefulWidget::render(list, area, buf, &mut state);
+                        Paragraph::new(item)
+                            .fg(TEXT_COLOR)
+                            .wrap(Wrap { trim: false })
+                            .block(Block::new().borders(Borders::BOTTOM))
+                            .render(area, buf);
                     } else {
                         render_room_stats(room, area, buf);
                     }
@@ -68,7 +65,7 @@ impl Widget for &mut ReadReceipts<'_> {
     }
 }
 
-fn format_timeline_item(item: &TimelineItem) -> Option<ListItem<'_>> {
+fn format_timeline_item(item: &TimelineItem) -> Option<Vec<Line<'_>>> {
     let event = item.as_event()?;
     let receipts = event.read_receipts();
     let sender = event.sender();
@@ -77,5 +74,5 @@ fn format_timeline_item(item: &TimelineItem) -> Option<ListItem<'_>> {
     let first_line = Line::from(format!("{sender} - {event_id:?}"));
     let second_line = Line::from(format!("{receipts:?}"));
 
-    Some(ListItem::from(vec![first_line, second_line]))
+    Some(vec![first_line, second_line])
 }


bnjbvr (commented):
At a first glance, I don't understand the output, and if one has to make a connection between the events IDs and timeline items (without seeing the timeline items IDs in the message view), I don't think this is practical in its current shape.

Could we display the read receipts in a tiny popover, on top of the messages, when we've selected one message in the room view? This would visually "link" the read receipts closer to the item they're referring to, and we could reuse this popover for other purposes (add edit/redact/open thread buttons, for instance). What do you think?

poljar:
> Could we display the read receipts in a tiny popover, on top of the messages, when we've selected one message in the room view?

This sounds pretty complicated, though perhaps we should spend the time to have a per-event debug view which shows the source as well.

Though then the details views become a bit redundant. Are you sure you are fine seeing this only for a single event at a time?

What do you think about this TODO comment: https://github.com/matrix-org/matrix-rust-sdk/pull/5188/files#diff-e95372f3a9ac5cdd7829584d1591a511e0b204b4d8c1241e628c77950fbdf1eeR63-R66?

That way you could see all events and see which one is selected in the timeline as well as see the selection in the read receipts view.

bnjbvr:
If the detailed view can be synchronized with the current event we're looking at, that's perfect; we could have a view for showing read receipts, reactions, and so on. Sounds like a great idea, thanks!

poljar:
> If the detailed view can be synchronized with the current event we're looking at, that's perfect; we could have a view for showing read receipts, reactions, and so on. Sounds like a great idea, thanks!

Well it requires a bit of shared state between the two views but I think it's doable.

poljar:
> If the detailed view can be synchronized with the current event we're looking at, that's perfect; we could have a view for showing read receipts, reactions, and so on. Sounds like a great idea, thanks!

Alright, seems that I got it working:

image

poljar:
Alright, after some chat discussion, @bnjbvr requested that we render read receipts only for the selected event.

There's a slight annoyance with this, we can briefly select events that are part of the `Vector` of timeline items but aren't rendered.

This is due to how the `ListState` and the `ListState::select_next()` method works in Ratatui:

> /// Note: until the list is rendered, the number of items is not known, so the index is set to
>    /// `0` and will be corrected when the list is rendered

So we briefly select an item that isn't rendered, but it gets corrected.

Not sure if we should attempt to fix this, it will probably fix itself once the SDK filters out threaded events. At least in the case where a threaded event is the last event in the vector.

For other event types, we would need to start to render everything the vector contains, or manually remember the number of items so `ListState::select_next()` doesn't go to an unrendered item.

bnjbvr:
@poljar Ah yeah, I had to deal with this problem in #5210, by [mapping the indices of rendered items to the timeline item indices](https://github.com/matrix-org/matrix-rust-sdk/pull/5210/files#diff-1c406501f13c4d0011ba1279a310f9d24ddc95641f828f37f2170ec0bd519dc1R27-R29): maybe there's a way to reuse that logic here?

poljar:
> @poljar Ah yeah, I had to deal with this problem in #5210, by [mapping the indices of rendered items to the timeline item indices](https://github.com/matrix-org/matrix-rust-sdk/pull/5210/files#diff-1c406501f13c4d0011ba1279a310f9d24ddc95641f828f37f2170ec0bd519dc1R27-R29): maybe there's a way to reuse that logic here?

Well then that's already fixed, since we use the `ListState` of the timeline here, or rather we use the index of the `ListState` to figure out the selected event ID.

bnjbvr:
I run into a strange bug when trying this out, where I can't see anything in the room's view, after selecting a room with CtrlJ/CtrlK. I'm trying to figure out whether that's something wrong on my side, network turning bad or something like this.

> Well then that's already fixed, since we use the ListState of the timeline here, or rather we use the index of the ListState to figure out the selected event ID.

I'm pretty sure it's not, unfortunately. If a timeline item isn't rendered on screen, then the rendered list indices won't match the underlying timeline item indices (any item for which `format_timeline_item()` would return `None`: a non-`Message` msg-like, a `Poll`, a state membership without a change, etc.), so there'll be an offset between those. The consequence is that the event details view wouldn't show you the details for the event you selected before, but for another one in the list. I've actually run into this bug in the other PR adding threading support, hence the logic to map a rendered list item index to the underlying timeline item index. Maybe we could get to review and merge my PR, and you could reuse that logic here then?

Example of an off-by-one here: (I've taken your branch, and added printing the message in the event read receipt view, if it was available)

image

poljar:
> I run into a strange bug when trying this out, where I can't see anything in the room's view, after selecting a room with CtrlJ/CtrlK. I'm trying to figure out whether that's something wrong on my side, network turning bad or something like this.

I saw that as well, it resolved itself eventually so I think it was some network based delay.

poljar:
> I'm pretty sure it's not, unfortunately. If a timeline item isn't rendered on screen, then the rendered list indices won't match the underlying timeline item indices (any item for which format_timeline_item() would return None: a non-Message msg-like, a Poll, a state membership without a change, etc.), so there'll be an offset between those.

Yes, but your `TimelineListState` ensures that we get the index of the rendered item.

Then from this index I fetch the item. If the item contains read receipts we render them, otherwise we render the global room read receipt stats.

I think that this works now correctly. For good measure I'll disabled the selecting past the list behavior as well in https://github.com/matrix-org/matrix-rust-sdk/pull/5188/commits/ec51a4cf07c0d60d2ab7c720c12e60551c79a524.

labs/multiverse/src/widgets/room_view/details/read_receipts.rs
```diff
@@ -1,48 +1,81 @@
...
+                Some(selected_event) => {
+                    if let Some(item) = format_timeline_item(selected_event) {
+                        let list = List::new(vec![item])
+                            .highlight_spacing(HighlightSpacing::Always)
+                            .highlight_symbol(">")
+                            .highlight_style(SELECTED_STYLE_FG);
+
+                        let mut state = ListState::default();
+
+                        StatefulWidget::render(list, area, buf, &mut state);
```
bnjbvr: Any particular interesting in having this be a `List`, and not a plain `Paragraph`? As far as I can tell, there's no way to select the individual items in this list…

labs/multiverse/src/widgets/room_view/mod.rs
```diff
@@ -565,6 +579,17 @@ impl RoomView {
...
+    fn get_selected_event(&self) -> Option<Arc<TimelineItem>> {
+        let selected = self.timeline_list.selected()?;
+        let selected_room = self.room_id()?;
+
+        let timelines = self.timelines.lock();
+        let current_timeline = timelines.get(selected_room)?;
+        let items = current_timeline.items.lock();
+
+        items.get(selected).cloned()
+    }
```
bnjbvr: I think that, to work nicely for threads, this should be:

```suggestion
    fn get_selected_event(&self) -> Option<Arc<TimelineItem>> {
        let selected = self.timeline_list.selected()?;
        let items = self.get_selected_timeline_items()?;
        items.get(selected).cloned()
    }
```

But the one caller of this function also calls `get_selected_timeline_items()`, so there's likely a way to avoid this function (that would be now 3 lines), and have the caller get the selected item from `items` (at line 648) immediately.

labs/multiverse/src/widgets/room_view/mod.rs
```diff
@@ -620,6 +645,9 @@ impl Widget for &mut RoomView {
             let maybe_room = self.client.get_room(room_id);
             let mut maybe_room = maybe_room.as_ref();
 
+            let items = self.get_selected_timeline_items();
+            let selected_event = self.get_selected_event();
```
bnjbvr: Consider pushing these two lines down just before (current) line 671, so they're closer to their use (and not unused if the mode is normal)?

bnjbvr (changes requested):
Code looks good, but it seems that the change to `TimelineListState::select_next()` makes it impossible to select the next items, sometimes…


https://github.com/user-attachments/assets/87b91a96-5ecf-4611-a3c9-940ed1c367a6

labs/multiverse/src/widgets/room_view/mod.rs
poljar: You can't, because `self` will get borrowed mutably further below and then you can't borrow immutably anymore.

labs/multiverse/src/widgets/room_view/details/read_receipts.rs
poljar: No, it's just what it used to be when receipts for all items were shown.

Though I would imagine that we'd want to have prettier formatting for this, which might be nice to format using a list/table?

labs/multiverse/src/widgets/room_view/mod.rs
poljar: > But the one caller of this function also calls get_selected_timeline_items()

No, it's not doing that anymore.

poljar:
> Code looks good, but it seems that the change to `TimelineListState::select_next()` makes it impossible to select the next items, sometimes…
> Screencast.From.2025-06-12.14-56-22.mp4

Hmm, ok. I'll revert that then.

PR #5189, opened 2025-06-06
Upgrade to uniffi 29.x to support wasm targets

zzorba:
Upgrade to what will soon be uniffi 0.29.3 (release is planned for next week).  This mostly concerns an upgrade of the xtask orchestration system, but there were some small changes to the .udl spec that were needed as well.

This will likely produce issues with the complement-crypto package, which seems to still be using uniffi 0.25 to build.
  
Signed-off-by: Daniel Salinas

### DIFF
--- a/xtask/src/config_supplier.rs
+++ b/xtask/src/config_supplier.rs
@@ -1,0 +1,74 @@
+use std::{collections::HashMap, fs};
+
+use anyhow::{bail, Context};
+use camino::Utf8PathBuf;
+use cargo_metadata::Metadata;
+use toml::value::Table;
+use uniffi_bindgen::BindgenCrateConfigSupplier;
+
+#[derive(Debug, Clone, Default)]
+pub struct CrateConfigSupplier {
+    paths: HashMap<String, Utf8PathBuf>,
+}
+
+fn load_toml_file(path: Option<&Utf8PathBuf>) -> anyhow::Result<Option<Table>> {
+    if let Some(path) = path {
+        if path.exists() {
+            let contents = fs::read_to_string(path)?;
+            let table: Table = toml::from_str(&contents)?;
+            Ok(Some(table))
+        } else {
+            Ok(None)
+        }
+    } else {
+        Ok(None)
+    }
+}
+
+impl BindgenCrateConfigSupplier for CrateConfigSupplier {
+    fn get_toml(&self, crate_name: &str) -> anyhow::Result<Option<Table>> {
+        load_toml_file(self.get_toml_path(crate_name).as_ref())
+    }
+
+    fn get_toml_path(&self, crate_name: &str) -> Option<Utf8PathBuf> {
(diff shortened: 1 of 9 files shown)


xtask/src/config_supplier.rs
```diff
@@ -0,0 +1,74 @@
+use std::{collections::HashMap, fs};
+
+use anyhow::{bail, Context};
+use camino::Utf8PathBuf;
+use cargo_metadata::Metadata;
+use toml::value::Table;
+use uniffi_bindgen::BindgenCrateConfigSupplier;
+
+#[derive(Debug, Clone, Default)]
+pub struct CrateConfigSupplier {
```
jplatte, line 12: Would be nice to have some docs. What does this config supplier do differently from the default upstream one?

xtask/src/kotlin.rs
```diff
@@ -121,7 +122,24 @@ fn build_android_library(
...
 
-    generate_bindings(library_path, None, &KotlinBindingGenerator, None, ffi_generated_dir, false)?;
+    let manifest_path = std::env::current_dir()?.join("Cargo.toml");
+    println!("manifest path {:?}", manifest_path);
+
+    // Get metadata using cargo_metadata
+    let metadata = MetadataCommand::new().manifest_path(&manifest_path).exec()?;
+
+    // Convert the Metadata into a CrateConfigSupplier using From
+    let config_supplier = CrateConfigSupplier::from(metadata);
```
jplatte: These comments do not add anything.

xtask/src/config_supplier.rs
zzorba: Which upstream class one are you seeing?  I only see the trait and an 'Empty' supplier in the 2.9.x versions of Uniffi.

Hywan:
What's the status of this PR?

xtask/src/config_supplier.rs
jplatte: I thought I had a seen a "default config supplier". There seem to be two upstream though: https://docs.rs/uniffi_bindgen/0.29.3/uniffi_bindgen/cargo_metadata/struct.CrateConfigSupplier.html is the other one.

zzorba:
@Hywan this PR is still necessary, it upgrades us to the most recent uniffi (0.29.3) which contains support for Wasm targets.  Without it the matrix-sdk-ffi crate will not compile. 

It is unfortunately blocked on the complement-crypto library, which I believe is blocked on NordSecurity considering an upgrade I submitted https://github.com/NordSecurity/uniffi-bindgen-go/pull/77. 

I believe Matthew has asked someone to look into the complement-crypto issue, which I believe is currently monkey-patching to work with a very old version of uniffi.

poljar:
Superseded by https://github.com/matrix-org/matrix-rust-sdk/pull/5808.

I'm not particularly sure why this PR is so much more complex compared to #5808, so if some of this complexity is needed please open a new PR on top of #5808.

PR #5192, opened 2025-06-07
refactor: Clean up tracing and formatting macro uses

jplatte:
I noticed one thing and couldn't stop myself from looking a bit further, recursively 😅 

### DIFF
--- a/bindings/matrix-sdk-ffi/src/authentication.rs
+++ b/bindings/matrix-sdk-ffi/src/authentication.rs
@@ -171,7 +171,7 @@ impl OidcConfiguration {
                 .iter()
                 .filter_map(|(issuer, client_id)| {
                     let Ok(issuer) = Url::parse(issuer) else {
-                        tracing::error!("Failed to parse {:?}", issuer);
+                        tracing::error!("Failed to parse {issuer:?}");
                         return None;
                     };
                     Some((issuer, ClientId::new(client_id.clone())))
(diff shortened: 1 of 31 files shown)


andybalaam (approved):
Looks good, but a compile error slipped in somehow. (Thanks!)

jplatte:
Yeah I had a git fail and reverted parts of https://github.com/matrix-org/matrix-rust-sdk/pull/5191 by accident. Thanks for reminding me to fix it ^^

bindings/matrix-sdk-ffi/src/authentication.rs
```diff
@@ -171,7 +171,7 @@ impl OidcConfiguration {
                 .iter()
                 .filter_map(|(issuer, client_id)| {
                     let Ok(issuer) = Url::parse(issuer) else {
-                        tracing::error!("Failed to parse {:?}", issuer);
+                        tracing::error!("Failed to parse {issuer:?}");
```
richvdh, line 174: Is there a particular reason to prefer one format over the other here, out of interest?

jplatte: Where possible, the inline syntax is preferable IMO. However, Rust's formatting macros only support the inline syntax for bare identifiers, i.e. thingw like `format!("{structure.field:?}")` are not supported.

Issue #5190, opened 2025-06-07
matrix-sdk-crypto from crates.io fails to compile


VerdeQuar:
# Maintainer update

The issue has been identified as an accidental breaking change in a dependency. While we are looking into the best way foward, you can work around the issue by downgrading `tracing-attributes`:

```
cargo update -p tracing-attributes --precise 0.1.28
```

# Original issue description

I was doing some work on a different project and everything was working fine, until i did `cargo update`, despite `matrix_sdk` being still in version `0.11.0` like it was before, it for some reason stopped compiling and gave me this error:
```
error[E0573]: expected type, found variant `CryptoStoreError`
    --> /home/verdek/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/matrix-sdk-crypto-0.11.0/src/identities/manager.rs:1182:21
     |
44   |     CryptoStoreError, LocalTrust, OwnUserIdentity, SignatureError, UserIdentity,
     |     ---------------- you might have meant to refer to this enum
...
1182 |     ) -> Result<(), CryptoStoreError> {
     |                     ^^^^^^^^^^^^^^^^ not a type

For more information about this error, try `rustc --explain E0573`.
error: could not compile `matrix-sdk-crypto` (lib) due to 1 previous error
```
i tested it on an empty project in which i just did `cargo add matrix_sdk` and it also doesn't compile
i'm on stable rust `1.87.0` but when i switched to `1.86.0` i got the same error
same thing with downgrading to `matrix_sdk` `0.10.0`

seems very weird that that kind of error would go unnoticed for so long, so i have no idea if this is even reproducible, but i found that removing [this](https://github.com/matrix-org/matrix-rust-sdk/blob/47e81818bcbc8cf6859c23ac0faddd54d4a70cee/crates/matrix-sdk-crypto/src/identities/manager.rs#L1127) wildcard import solves the issue
which seems even weirder cause i would assume that import would only work inside the scope of the function, but it somehow triggers an error in its signature
in any case, wildcard imports of enum variants are very error prone so it shouldn't even be there in the first place, should i make a pr?

jplatte:
What an interesting bug! I suspect you will be able to reproduce it here in the repo but updating the lockfile with `cargo update`. And I agree with removing the wildcard import as a fix.

VerdeQuar:
Actually this repo as well as `crates/matrix-sdk-crypto` on its own, both do compile for me after cargo update
[...]

PR #5194, opened 2025-06-09
feat: Support releasing a specific package

poljar:
Since #5190 causes issues for an existing release, it would be nice to do a patch release for this specific crate.

Our tooling, which wraps `cargo release`, doesn't support this as of yet. Let's change that.

### DIFF
--- a/xtask/src/release.rs
+++ b/xtask/src/release.rs
@@ -1,5 +1,5 @@
 use clap::{Args, Subcommand, ValueEnum};
-use xshell::cmd;
+use xshell::{cmd, Cmd};
 
 use crate::{sh, Result};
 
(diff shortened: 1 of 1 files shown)


xtask/src/release.rs
```diff
@@ -88,9 +99,15 @@ fn check_prerequisites() {
...
+fn prepare(version: ReleaseVersion, execute: bool, package: Option<String>) -> Result<()> {
     let sh = sh();
-    let cmd = cmd!(sh, "cargo release --workspace --no-publish --no-tag --no-push");
+    let cmd = cmd!(sh, "cargo release --no-publish --no-tag --no-push");
+
+    let cmd = if let Some(package) = package {
+        cmd.args(["--package", &package])
+    } else {
+        cmd.arg("--workspace")
+    };
```
bnjbvr: Since this bit is duplicated 4 times, it's possible that a small function helper would remove redundancies, but since this is the xtask code, I don't have a strong opinion at all :)

poljar: Sure: https://github.com/matrix-org/matrix-rust-sdk/pull/5194/commits/42d373cb219818486cb837e40b18b1f5372cc8f9

PR #5198, opened 2025-06-09
feat(ui): don't mark each thread reply as an actual reply to the previous message, in threaded timelines

bnjbvr:
This correctly handles the reply fallback behavior:

- the behavior isn't changed for a live timeline,
- when a timeline is thread-focused, we will extract the `replied_to` field if and only if the thread relation is *not* marked as behaving in a fallback manner.

This makes it possible to distinguish actual in-thread replies.

Part of #4833.

### DIFF
--- a/crates/matrix-sdk-ui/src/timeline/controller/metadata.rs
+++ b/crates/matrix-sdk-ui/src/timeline/controller/metadata.rs
@@ -38,6 +38,7 @@ use super::{
 };
 use crate::{
     timeline::{
+        controller::TimelineFocusKind,
         event_item::{
             extract_bundled_edit_event_json, extract_poll_edit_content,
             extract_room_msg_edit_content,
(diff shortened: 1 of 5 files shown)


crates/matrix-sdk-ui/src/timeline/controller/metadata.rs
```diff
@@ -450,9 +464,25 @@ impl TimelineMetadata {
...
+
+                if matches!(timeline_focus, TimelineFocusKind::Thread { .. })
+                    && thread.is_falling_back
+                {
+                    // In general, a threaded event is marked as a response to the previous message
+                    // in the thread, to maintain backwards compatibility with clients not
+                    // supporting threads.
+                    //
+                    // But we can have actual replies to other in-thread events. The
+                    // `is_falling_back` bool helps distinguishing both use cases.
```
stefanceriu, line 476: TIL and makes perfect sense to me 👍

crates/matrix-sdk-ui/tests/integration/timeline/thread.rs
```diff
@@ -159,14 +159,17 @@ async fn test_thread_backpagination() {
...
+    assert_eq!(event_item.content().as_message().unwrap().body(), "Threaded event 3");
+    // In a threaded timeline, threads aren't using the reply fallback, unless
+    // they're an actual reply to another thread event.
+    assert_matches!(event_item.content().in_reply_to(), None);
+
+    let event_item = items[2].as_event().unwrap();
+    assert_eq!(event_item.content().as_message().unwrap().body(), "Threaded event 4");
+    // But this one is an actual reply to another thread event, so it has the
+    // replied-to event correctly set.
+    assert_eq!(event_item.content().in_reply_to().unwrap().event_id, event_id!("$2"));
```
stefanceriu, line 172: Super nice, thank you!

stefanceriu (approved):
🙌

PR #5199, opened 2025-06-09
Revert the improved tombstone event handling

poljar:
Seems that we have a couple of bugs in the tombstone event handling, a relevant bug report can be found here https://github.com/element-hq/element-x-android-rageshakes/issues/6206.

This reverts 0478037b57ce21a5a1a4fe6acf5b7cf75e6fc890, 8e3ad22d92f2bd949be307f44dd62d3dfd7248e2, and c7f6190cffe289f97ec92fa7b7b195169c5fbc9a from https://github.com/matrix-org/matrix-rust-sdk/pull/5186 and https://github.com/matrix-org/matrix-rust-sdk/pull/5150.


bnjbvr (approved):

bnjbvr:
For what it's worth, this is due to the fact that we've observed at least one public room, in the wild, that caused an error in the successor/predecessor relationships, because the successor room didn't have the predecessor room marked as a predecessor. It's https://matrix.to/#/#fosdem2020:matrix.org, which is an upgraded room, but the create event doesn't mention it:

```json
{
  "content": {
    "creator": "@Half-Shot:half-shot.uk",
    "room_version": "4"
  },
  "origin_server_ts": 1572270366956,
  "sender": "@Half-Shot:half-shot.uk",
  "state_key": "",
  "type": "m.room.create",
  "unsigned": {
    "age": 196156
  },
  "event_id": "$Y5IUaHVFK8ie-Xre4zjkCoxQHdhWAluoXLMO0rgZp34",
  "room_id": "!LNrXvgubOkADypzhHy:half-shot.uk"
}
```

cc @Hywan

Hywan:
So yeah, `check_room_upgrades` did its job to detect an error. Since this is specific error isn't critical, we may prefer to log an error instead of returning an error. I'm on it. Thanks for the investigation.

PR #5201, opened 2025-06-10
crypto: Only expose type/sender/content for raw decrypted to-device msg

BillCarsonFr:
The `DecryptionResult#raw_event` is exposing a lot of "internal" crypto related fields as part of the raw json encoded object. 
All these properties are then exposed outside of the crypto crate, in a sort of hidden way because you have to serialize as a Value to get them. For example in the wasm-bindings they will be exposed to the js-sdk and easilly accessible and that could cause the sdk to wrongly use these fields (source of footgun?)

Just like we zeroized content of to-device messages, I think we should also clean-up these crypto fields before exposing them outside?

Before:
```
{
    "type": "my.custom.to.device",
    "content": {
        "call_id": ""
    },
    "keys": {
        "ed25519": "hk5X9uSb0OwhdAFrlcdtzNkE+AmL/85XdPO5v1Ocgg"
    },
    "recipient": "@alice:example.org",
    "recipient_keys": {
        "ed25519": "WgEu0w6T511+R03R4dUQP53d9OQTVJr6F/cYmF3Civs"
    },
    "sender": "@bob:example.org",
    "sender_device_keys": {
        "algorithms": [
            "m.olm.v1.curve25519-aes-sha2",
            "m.megolm.v1.aes-sha2"
        ],
        "device_id": "B0B0B0B0B",
        "keys": {
            "curve25519:B0B0B0B0B": "Bwl3kZjTnxxKNufqMiG1XoJg1VxGQko5asIy6aq4pUk",
            "ed25519:B0B0B0B0B": "hk5X9uSb0OwhdAFrlcdtzNkE+AmLT/85XdPO5v1Ocgg"
        },
        "signatures": {
            "@bob:example.org": {
                "ed25519:B0B0B0B0B": "XvyTR/3piBBCl4Xq4wHjrjhnJdgMMtXvpMh9itG1Nr9GUI80n8bQ32CtHTZCENk77esgUZ4Tkzy0bUGJRMzZAg"
            }
        },
    	"user_id": "@bob:example.org",
    },
}
```

After
```
{
    "type": "my.custom.to.device",
    "content": {
        "call_id": ""
    },
    "sender": "@bob:example.org",
}
```

Signed-off-by:

### DIFF
--- a/crates/matrix-sdk-crypto/src/olm/account.rs
+++ b/crates/matrix-sdk-crypto/src/olm/account.rs
@@ -42,10 +42,7 @@ use ruma::{
     SecondsSinceUnixEpoch, UInt, UserId,
 };
 use serde::{de::Error, Deserialize, Serialize};
-use serde_json::{
-    value::{to_raw_value, RawValue as RawJsonValue},
-    Value,
-};
+use serde_json::{json, value::to_raw_value, Value};
 use sha2::{Digest, Sha256};
 use tokio::sync::Mutex;
 use tracing::{debug, field::debug, info, instrument, trace, warn, Span};
@@ -128,6 +125,7 @@ pub(crate) struct OlmDecryptionInfo {
 pub(crate) struct DecryptionResult {
     // AnyDecryptedOlmEvent is pretty big at 512 bytes, box it to reduce stack size
     pub event: Box<AnyDecryptedOlmEvent>,
+    // The raw decrypted event, with only `type`, `sender` and `content` fields.
     pub raw_event: Raw<AnyToDeviceEvent>,
     pub sender_key: Curve25519PublicKey,
     pub encryption_info: EncryptionInfo,
@@ -1511,12 +1509,17 @@ impl Account {
 
             let encryption_info = Self::get_olm_encryption_info(sender_key, sender, &sender_device);
 
-            Ok(DecryptionResult {
-                event,
-                raw_event: Raw::from_json(RawJsonValue::from_string(plaintext)?),
-                sender_key,
-                encryption_info,
-            })
+            let parsed: Value = serde_json::from_str(&plaintext)?;
+
+            let raw_event = json!({
+                "type": event.event_type().to_owned(),
+                "content": parsed["content"].clone(),
+                "sender": event.sender().to_owned(),
+            });
+
+            let raw_event: Raw<AnyToDeviceEvent> = Raw::from_json_string(raw_event.to_string())?;
+
+            Ok(DecryptionResult { event, raw_event, sender_key, encryption_info })
         }
     }
 
--- a/crates/matrix-sdk-crypto/src/machine/tests/olm_encryption.rs
+++ b/crates/matrix-sdk-crypto/src/machine/tests/olm_encryption.rs
@@ -24,7 +24,7 @@ use ruma::{
     events::{dummy::ToDeviceDummyEventContent, AnyToDeviceEvent},
     user_id, DeviceKeyAlgorithm, DeviceKeyId, SecondsSinceUnixEpoch,
 };
-use serde_json::json;
+use serde_json::{json, Value};
 use vodozemac::Ed25519SecretKey;
 
 use crate::{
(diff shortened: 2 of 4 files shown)


crates/matrix-sdk-crypto/src/olm/account.rs
```diff
@@ -128,6 +125,7 @@ pub(crate) struct OlmDecryptionInfo {
 pub(crate) struct DecryptionResult {
     // AnyDecryptedOlmEvent is pretty big at 512 bytes, box it to reduce stack size
     pub event: Box<AnyDecryptedOlmEvent>,
+    // The raw decrypted event, with only `type`, `sender` and `content` fields.
```
BillCarsonFr, line 128: Or maybe we should remove `raw_event` and add `clear_type` and `clear_content`?
Not sure what was the original reason to keep the raw_event. Notice that all the fields in there are still properly accessible via `AnyDecryptedOlmEvent`

crates/matrix-sdk-crypto/src/olm/account.rs
```diff
@@ -1511,12 +1509,17 @@ impl Account {
...
+
+            let raw_event = json!({
+                "type": event.event_type().to_owned(),
+                "content": parsed["content"].clone(),
+                "sender": event.sender().to_owned(),
+            });
+
+            let raw_event: Raw<AnyToDeviceEvent> = Raw::from_json_string(raw_event.to_string())?;
+
+            Ok(DecryptionResult { event, raw_event, sender_key, encryption_info })
```
Hywan, line 1522: It doesn't look super efficient. You are deserializing the entire `plaintext` as `Value`, then you are creating another `Value` based on the previous one, then you're serializing it with `to._string`, to finally re-deserializing it here.

I wonder if you could not “simply” replace `Raw::from_json(RawJsonValue::from_string(…))` by `Raw::<AnyToDeviceEvent>::from_json_string(plaintext);`, something like this:

```suggestion
            Ok(DecryptionResult { event, raw_event: Raw::<AnyToDeviceEvent>::from_json_string(plaintext)?, sender_key, encryption_info })
```

Not tested.

Thoughts?

Hywan (changes requested):
Thanks for the patch.

The idea is excellent, but the proposed code is suboptimal. I think we can make it simpler.

poljar:
> Just like we zeroized content of to-device messages, I think we should also clean-up these crypto fields before exposing them outside?

Hmm, but those are part of the event/message, defined here: https://spec.matrix.org/unstable/client-server-api/#molmv1curve25519-aes-sha2.

And those fields aren't exactly secrets like the things we do zeroize.

BillCarsonFr:
> > Just like we zeroized content of to-device messages, I think we should also clean-up these crypto fields before exposing them outside?
> 
> Hmm, but those are part of the event/message, defined here: https://spec.matrix.org/unstable/client-server-api/#molmv1curve25519-aes-sha2.
> 
> And those fields aren't exactly secrets like the things we do zeroize.

Right, but they are already available in the `AnyDecryptedOlmEvent` struct.
To give more context, I am working on the to-device message support for widgets. And widgets are not supposed to be aware of any encryption related things. So I need to clean-up that before forwarding.

But maybe the solution is to return a `clearType` and a `clearContent` instead of the `raw` event. I don't think the raw event is used anyhow (used to access all the  plain text `OlmPayload` fields?

poljar:
> To give more context, I am working on the to-device message support for widgets. And widgets are not supposed to be aware of any encryption related things. So I need to clean-up that before forwarding.

Alright, but is it crucial from preventing them from seeing these fields? If so, this might make sense to filter out in the widget stuff.

I don't think it's necessary to do this in the crypto crate.

> But maybe the solution is to return a clearType and a clearContent instead of the raw event. I don't think the raw event is used anyhow (used to access all the plain text OlmPayload fields?

Well the raw event is mostly for consumers of the SDK here, no idea if they do something with these fields. We just want to pass the JSON as it was received, except of course, for fields containing secrets.

crates/matrix-sdk-crypto/src/olm/account.rs
poljar: We discussed this in chat, the SDK needs to let people get the complete JSON as it was received and decrypted. Higher level layers can decide to filter out things.

In this case, the widget driver is such a higher layer.

BillCarsonFr:
As per discussion will close this PR. for the specific case of widget to-device messages the clean-up will be done at the widget driver level

PR #5209, opened 2025-06-10
fix(base): Ignore invalid state events instead of throwing an error

Hywan:
This patch changes the strategy of `check_room_upgrades`. Instead of checking all state events in `StateChanges` and throwing an error in case of an invalid state, this patch updates `state_events::sync::dispatch` to filter out invalid state events (specifically `m.room.create` and `m.room.tombstone`), and log the error. That way, the sync doesn't stop and the app can continue working smoothly.

So `check_room_upgrades` is splited into two functions: `is_create_event_valid` and `is_tombstone_event_valid`. It's no longer necessary to detect mergers or splitters because those are _soft errors_, however loops are still critical errors (hence the fact the state events aren't stored nor saved nor applied, and that a log is emitted).

Note: `check_room_upgrades` has been reverted in a previous commit, that's why it doesn't appear in this diff.

---

- Sequel of https://github.com/matrix-org/matrix-rust-sdk/pull/5199 (thanks @poljar for the revert)
- Address https://github.com/matrix-org/matrix-rust-sdk/issues/5050

### DIFF
--- a/crates/matrix-sdk-base/src/response_processors/state_events.rs
+++ b/crates/matrix-sdk-base/src/response_processors/state_events.rs
@@ -546,7 +546,7 @@ mod tests {
             assert!(room_0.successor_room().is_none());
         }
 
-        // Room 0 <-> room 1.
+        // Room 0 and room 1.
         {
             let tombstone_event_id = event_id!("$ev0");
             let response = response_builder
@@ -585,7 +585,7 @@ mod tests {
             assert!(room_1.successor_room().is_none(), "room 1 must not have a successor");
         }
 
-        // Room 1 <-> room 2.
+        // Room 1 and room 2.
         {
             let tombstone_event_id = event_id!("$ev1");
             let response = response_builder
@@ -743,7 +743,7 @@ mod tests {
 
         let client = logged_in_base_client(None).await;
 
-        // Room 0 -> room 0.
+        // Room 0.
         {
             let tombstone_event_id = event_id!("$ev0");
             let response = response_builder
@@ -780,7 +780,7 @@ mod tests {
 
         let client = logged_in_base_client(None).await;
 
-        // Room 0 <-> room 1.
+        // Room 0 and room 1.
         {
             let tombstone_event_id = event_id!("$ev0");
             let response = response_builder
@@ -819,7 +819,7 @@ mod tests {
             assert!(room_1.successor_room().is_none(), "room 1 must not have a successor");
         }
 
-        // Room 1 <-> room 2 -> room 0.
+        // Room 1, room 2 and room 0.
         {
             let tombstone_event_id = event_id!("$ev1");
             let response = response_builder
@@ -891,7 +891,7 @@ mod tests {
 
         let client = logged_in_base_client(None).await;
 
-        // Room 0 <- room 0.
+        // Room 0.
         {
             let tombstone_event_id = event_id!("$ev0");
             let response = response_builder
@@ -927,7 +927,7 @@ mod tests {
 
         let client = logged_in_base_client(None).await;
 
-        // Room 0 <-> room 0.
+        // Room 0.
         {
             let tombstone_event_id = event_id!("$ev0");
             let response = response_builder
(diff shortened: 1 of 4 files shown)


crates/matrix-sdk-base/src/response_processors/state_events.rs
```diff
@@ -248,16 +307,762 @@ where
...
+
+            assert!(client.receive_sync_response(response).await.is_ok());
+
+            let room_0 = client.get_room(room_id_0).unwrap();
+
+            assert!(room_0.predecessor_room().is_none());
+            assert!(room_0.successor_room().is_none());
+        }
+
+        // Room 0 <-> room 1.
```
andybalaam: The `<->` confused me here. I understand why it is double-headed, but it reads like it's a loop.

Could do: `Room 0 --tombstone-->    <--create-- Room 1` ?

crates/matrix-sdk-base/src/response_processors/state_events.rs
```diff
@@ -76,22 +86,71 @@ pub mod sync {
...
+                    } else {
+                        error!(
+                            room_id = ?room_info.room_id(),
+                            ?create,
+                            "`m.create.tombstone` event is invalid, it creates a loop"
+                        );
+
+                        // Do not add the event to `room_info`.
+                        // Do not add the event to `context.state_changes.state`.
+                        break;
```
poljar: Hmm, if we break here, won't this potentially ignore a bunch of valid events as well?

Was that supposed to be a continue?

crates/matrix-sdk-base/src/response_processors/state_events.rs
```diff
@@ -76,22 +86,71 @@ pub mod sync {
...
+                    } else {
+                        error!(
+                            room_id = ?room_info.room_id(),
+                            ?tombstone,
+                            "`m.room.tombstone` event is invalid, it creates a loop"
+                        );
+
+                        // Do not add the event to `room_info`.
+                        // Do not add the event to `context.state_changes.state`.
+                        break;
```
poljar: Same thing, why do we break here, why don't we continue, or since nothing comes after this, do nothing.

crates/matrix-sdk-base/src/response_processors/state_events.rs
```diff
@@ -248,16 +307,762 @@ where
...
+                    .room(&predecessor_room_id)
+                    .and_then(|room| Some(room.predecessor_room()?.room_id))
+            })
+        else {
+            // No more predecessor found. Everything seems alright. No loop.
+            break;
+        };
+
+        predecessor_room_id = next_predecessor_room_id;
+    }
```
poljar, line 357: This seems like a fine check, though do we want to limit the loop iteration count somehow?

crates/matrix-sdk-base/src/response_processors/state_events.rs
```diff
@@ -248,16 +307,762 @@ where
...
+        else {
+            // No more successor found. Everything seems alright. No loop.
+            break;
+        };
+
+        successor_room_id = next_successor_room_id;
+    }
+
+    true
+}
```
poljar, line 409: Same comment, do you think we should limit the number of iterations here?

poljar (changes requested):
This is looking mostly good.

It worries me that we're intentionally skipping more state events than just the invalid tombstone/create events, or did I misunderstand that?

crates/matrix-sdk-base/src/response_processors/state_events.rs
Hywan: This function checks for a single new state event, so at worst, it skips a single state event.

Hywan: It's supposed to be a `continue`, you're right! Damn…

crates/matrix-sdk-base/src/response_processors/state_events.rs
Hywan: Bugger, what was I thinking about…?

Note: we must use `continue`, otherwise the state event will be added to `context.state_changes`, which isn't desired.

crates/matrix-sdk-base/src/response_processors/state_events.rs
Hywan: Not sure it's necessary. It will always exit the loop, so I don't believe this is required.

poljar: Alright.

poljar (approved):

crates/matrix-sdk-base/src/response_processors/state_events.rs
andybalaam: Can we add a test to cover this?

PR #5210, opened 2025-06-10
feat(multiverse): add support for threaded timelines

bnjbvr:
This adds supports for threaded timelines to multiverse, by pressing `Ctrl+T` on a given thread root event. This will be useful for debugging, and understanding how one would use the APIs, for sure. Already noticed #5216 because of this.

Part of #4869.

### DIFF
--- a/labs/multiverse/src/widgets/room_view/mod.rs
+++ b/labs/multiverse/src/widgets/room_view/mod.rs
@@ -231,8 +231,7 @@ impl RoomView {
 
         let status_handle = self.status_handle.clone();
 
-        // Start a new one, request batches of 20 events, stop after 10 timeline items
-        // have been added.
+        // Request to back-paginate 20 events.
         *pagination = Some(spawn(async move {
             if let Err(err) = sdk_timeline.paginate_backwards(20).await {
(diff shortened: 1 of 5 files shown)


labs/multiverse/src/widgets/room_view/mod.rs
```diff
@@ -47,26 +74,124 @@ pub struct RoomView {
...
+
+        let Some(root_event_id) = root_event.event_id().map(ToOwned::to_owned) else {
+            self.status_handle.set_message("can't open thread on a local echo".to_owned());
+            return;
+        };
+
+        self.status_handle.set_message(format!(
+            "Opening thread for event {root_event_id} in room {}",
+            room.room_id()
+        ));
```
poljar: This status message is a bit visually displeasing, especially since it doesn't use the room name.

Unless you really need it for debugging purposes I would remove it.

poljar (approved):
LGTM

labs/multiverse/src/widgets/room_view/mod.rs
bnjbvr: Yeah, I've kept it in the logs at least, because it helped show how long it takes to open a threaded view :+1:

PR #5215, opened 2025-06-11
feat(ffi): pass live events to threaded timelines and allow sending locations in them

stefanceriu:
This PR makes 2 changes:

1) exposes `ReplyParameters` on `send_location` on the FFI layer so that locations can be sent as replies, or, more importantly, within threads.
2) Updates the definition of what a "live" timeline is on the builder level so that events coming down from /sync are forwarded to threaded timelines.

### DIFF
--- a/bindings/matrix-sdk-ffi/src/timeline/mod.rs
+++ b/bindings/matrix-sdk-ffi/src/timeline/mod.rs
@@ -587,7 +587,8 @@ impl Timeline {
         description: Option<String>,
         zoom_level: Option<u8>,
         asset_type: Option<AssetType>,
-    ) {
+        reply_params: Option<ReplyParameters>,
+    ) -> Result<(), ClientError> {
         let mut location_event_message_content =
             LocationMessageEventContent::new(body, geo_uri.clone());
 
@@ -604,8 +605,13 @@ impl Timeline {
         let room_message_event_content = RoomMessageEventContentWithoutRelation::new(
             MessageType::Location(location_event_message_content),
         );
-        // Errors are logged in `Self::send` already.
-        let _ = self.send(Arc::new(room_message_event_content)).await;
+
+        if let Some(reply_params) = reply_params {
+            self.send_reply(Arc::new(room_message_event_content), reply_params).await
+        } else {
+            self.send(Arc::new(room_message_event_content)).await?;
+            Ok(())
+        }
     }
 
     /// Toggle a reaction on an event.
--- a/crates/matrix-sdk-ui/src/timeline/builder.rs
+++ b/crates/matrix-sdk-ui/src/timeline/builder.rs
@@ -172,8 +172,6 @@ impl TimelineBuilder {
         let (room_event_cache, event_cache_drop) = room.event_cache().await?;
         let (_, event_subscriber) = room_event_cache.subscribe().await;
 
-        let is_live = matches!(focus, TimelineFocus::Live { .. });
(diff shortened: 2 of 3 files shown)


crates/matrix-sdk-ui/src/timeline/builder.rs
```diff
@@ -172,7 +172,7 @@ impl TimelineBuilder {
         let (room_event_cache, event_cache_drop) = room.event_cache().await?;
         let (_, event_subscriber) = room_event_cache.subscribe().await;
 
-        let is_live = matches!(focus, TimelineFocus::Live { .. });
+        let is_live = matches!(focus, TimelineFocus::Live { .. } | TimelineFocus::Thread { .. });
```
Hywan: Can we view thread on a pinned event? Can we view thread on a permalink? In those case, `is_live` would be `true` and I think it's an error, thought?

stefanceriu: You can start a threaded timeline around *any* event, pinned or not, and that timeline should receive live updates.

Visually pinned timelines don't show threads (not sure they even build relations) but that's EX specific.

so I'm not sure I'm following where the error is 🤔

bindings/matrix-sdk-ffi/src/timeline/mod.rs
```diff
@@ -604,8 +605,13 @@ impl Timeline {
         let room_message_event_content = RoomMessageEventContentWithoutRelation::new(
             MessageType::Location(location_event_message_content),
         );
-        // Errors are logged in `Self::send` already.
-        let _ = self.send(Arc::new(room_message_event_content)).await;
+
+        if let Some(reply_params) = reply_params {
+            self.send_reply(Arc::new(room_message_event_content), reply_params).await
+        } else {
+            _ = self.send(Arc::new(room_message_event_content)).await?;
```
bnjbvr: is the `_ =` useful here?

crates/matrix-sdk-ui/src/timeline/builder.rs
bnjbvr: Since `is_live` is used downstream, how about making the downstream users take some variant of the focus? (either the simplified focus kind, or the full focus as a reference) so as to make the consequence of being live closer to the decision? Even `is_pinned_events` below would likely benefit from such a treatment, IMO

stefanceriu: Good idea, yeah, I'll make that happen 👍

bindings/matrix-sdk-ffi/src/timeline/mod.rs
stefanceriu: Huh, it was there before which I figured was to supress some sort of unused result warning. Guess not.

bnjbvr (approved):
looks great, thanks! (Please make sure to follow the PR name format, as described in the contributing guidelines; it's basically the same as commits, but as a short overview of the entire PR :))

Also, a note: I wonder if, with the recent improvements to `should_add…`, we could even have the logic in there, instead. If so, I suspect we might get rid of the other branch handling aggregations, because `should_add` doesn't control adding aggregations, so live aggregations would be handled live all the time. Something that we can experiment with in future PRs, likely, unless you want to play with that now (and re-request a review :))

stefanceriu:
In the interest of delivering this and fixing things on iOS for the next nightly I'll merge this PR as is and will look into the other changes separately.

PR #5220, opened 2025-06-11
feat(wasm): Fix cargo runtime on Wasm platforms

zzorba:
When file was [moved](https://github.com/matrix-org/matrix-rust-sdk/commit/2a140770a02cbeeaedc8c4dec90de9135aebc679) it looks like an update was missed, since wasm is not in the CI yet.

Signed-off-by: Daniel Salinas

### DIFF
--- a/bindings/matrix-sdk-ffi/src/runtime.rs
+++ b/bindings/matrix-sdk-ffi/src/runtime.rs
@@ -39,7 +39,7 @@ mod sys {
 mod sys {
     use std::future::Future;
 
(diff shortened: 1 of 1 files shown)


zzorba:
Also with this in the new location, the `runtime` won't be usable in the matrix-sdk-crypto-ffi crates. I don't need that for my purposes, but that was the original motivation for placing it in a more central location.  Perhaps a crate in the bindings folder would be more appropriate though.

jplatte (approved):

PR #5221, opened 2025-06-11
Add Wasm version of AsyncRuntimeDropped

zzorba:
The stock implementation of AsyncRuntimeDropped makes use of ManuallyDropped which is not usable on Wasm.  Introduce a Wasm specific version of this struct.

Signed-off-by: Daniel Salinas

### DIFF
--- a/bindings/matrix-sdk-ffi/src/utils.rs
+++ b/bindings/matrix-sdk-ffi/src/utils.rs
@@ -12,12 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use std::{mem::ManuallyDrop, ops::Deref};
-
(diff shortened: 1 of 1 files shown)


jplatte:
> ManuallyDropped which is not usable on Wasm.

This makes no sense. `ManuallyDrop` does not have anything target-specific about it.

What problems did you see with the current `AsyncRuntimeDropped` on Wasm?

jplatte (approved):
That said, the code changes look okay. It just doesn't make sense that `ManuallyDrop` would cause issues in Wasm.

zzorba:
You are absolutely correct, I just tried building it without this.  I'm working through the last stragglers on my long leveled branch, looking for things, but will try to test them in isolation.

PR #5222, opened 2025-06-12
feat(ffi): Expose legacy SSO support infomation

ShadowRZ:
Currently Element X can't distinguish the cases where a homeserver only
supports legacy SSO without OIDC (and password login isn't avaliable),
and other server unreachable scenarios.

This patch exposes legacy SSO support infomation so that Element X side
can give a dedicated error message when it encounters a homeserver that
can only support legacy SSO.

Signed-off-by: Yorusaka Miyabi <23130178+ShadowRZ@users.noreply.github.com>

### DIFF
--- a/bindings/matrix-sdk-ffi/src/client.rs
+++ b/bindings/matrix-sdk-ffi/src/client.rs
@@ -341,7 +341,24 @@ impl Client {
             }
         };
 
-        let supports_password_login = self.supports_password_login().await.ok().unwrap_or(false);
+        let login_types = self.inner.matrix_auth().get_login_types().await.ok();
+        let supports_password_login = login_types
+            .as_ref()
+            .map(|login_types| {
+                login_types.flows.iter().any(|login_type| {
+                    matches!(login_type, get_login_types::v3::LoginType::Password(_))
+                })
+            })
+            .unwrap_or(false);
+        let supports_sso_login = login_types
+            .as_ref()
+            .map(|login_types| {
+                login_types
+                    .flows
+                    .iter()
+                    .any(|login_type| matches!(login_type, get_login_types::v3::LoginType::Sso(_)))
+            })
+            .unwrap_or(false);
         let sliding_sync_version = self.sliding_sync_version();
 
         Arc::new(HomeserverLoginDetails {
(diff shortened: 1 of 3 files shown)


bindings/matrix-sdk-ffi/src/client.rs
```diff
@@ -336,12 +336,14 @@ impl Client {
...
         let supports_password_login = self.supports_password_login().await.ok().unwrap_or(false);
+        let supports_sso_login = self.supports_sso_login().await.ok().unwrap_or(false);
         let sliding_sync_version = self.sliding_sync_version();
 
         Arc::new(HomeserverLoginDetails {
             url: self.homeserver(),
             sliding_sync_version,
             supports_oidc_login,
+            supports_sso_login,
             supported_oidc_prompts,
```
stefanceriu: I think it's nicer to leave the oidc fields together and move `supports_sso_login` one line lower.

bindings/matrix-sdk-ffi/src/client.rs
```diff
@@ -743,6 +745,16 @@ impl Client {
             .any(|login_type| matches!(login_type, get_login_types::v3::LoginType::Password(_)));
         Ok(supports_password)
     }
+
+    /// Whether or not the client's homeserver supports the legacy SSO login flow.
+    pub(crate) async fn supports_sso_login(&self) -> anyhow::Result<bool> {
```
stefanceriu: This method is 99% the same as the `supports_password_login` one and neither has any other users so I'm thinking we should inline them at the place where we create the `HomeserverLoginDetails`.

stefanceriu (dismissed):

bindings/matrix-sdk-ffi/src/client.rs
ShadowRZ: I forget to change some identifiers 😅
But you're right, I'll change it.

poljar (changes requested):
Could you rebase this to get the CI going? Also a changelog entry for the new method would be nice.

Sorry this fell through, people went on holidays.

poljar (approved):
Thanks.

PR #5225, opened 2025-06-12
feat(event cache): introduce an absolute local event ordering

bnjbvr:
This PR introduces a local absolute ordering for items of a linked chunk, or equivalently, for events within a room's timeline. The idea is to reuse the same underlying mechanism we had for `AsVector`, but restricting it to only counting the number of items in a chunk; given an item's `Position`, we can then compute its absolute order as the total number of items *before* its containing chunk + its *index* within the chunk.

This will help us order edits that would apply to a thread event, for instance; this is deferred to a future PR, to not make this one too heavyweight.

Attention to reviewers: sorry, this is a bulky PR (mostly because of tests), but I think it's important to see how the `OrderTracker` methods are used in 280bd3288f8d5c867f0556bd4caf3c8ab57edece, to make sense of their *raison d'être*.

Part of #4869 / #5122.

### DIFF
--- a/crates/matrix-sdk-common/src/linked_chunk/as_vector.rs
+++ b/crates/matrix-sdk-common/src/linked_chunk/as_vector.rs
@@ -86,13 +86,10 @@ impl<Item, Gap> AsVector<Item, Gap> {
 
 /// Interface for a type accumulating updates from [`UpdateToVectorDiff::map`],
 /// and being returned as a result of this.
-pub(super) trait UpdatesAccumulator<Item> {
+pub(super) trait UpdatesAccumulator<Item>: Extend<VectorDiff<Item>> {
     /// Create a new accumulator with a rough estimation of the number of
     /// updates this accumulator is going to receive.
     fn new(num_updates_hint: usize) -> Self;
-
-    /// Fold the accumulator with the given new updates.
-    fn fold(&mut self, updates: impl IntoIterator<Item = VectorDiff<Item>>);
 }
 
 // Simple implementation for a `Vec<VectorDiff<Item>>` collection for
@@ -100,10 +97,6 @@ pub(super) trait UpdatesAccumulator<Item> {
 impl<Item> UpdatesAccumulator<Item> for Vec<VectorDiff<Item>> {
     fn new(num_updates_hint: usize) -> Vec<VectorDiff<Item>> {
         Vec::with_capacity(num_updates_hint)
-    }
-
-    fn fold(&mut self, updates: impl IntoIterator<Item = VectorDiff<Item>>) {
-        self.extend(updates)
     }
 }
 
@@ -375,7 +368,7 @@ impl<Item, Acc: UpdatesAccumulator<Item>> UpdateToVectorDiff<Item, Acc> {
                         .expect("Removing an index out of the bounds");
 
                     // Removing at the same index because each `Remove` shifts items to the left.
-                    acc.fold(repeat_n(VectorDiff::Remove { index: offset }, number_of_items));
+                    acc.extend(repeat_n(VectorDiff::Remove { index: offset }, number_of_items));
                 }
 
                 Update::PushItems { at: position, items } => {
@@ -395,13 +388,12 @@ impl<Item, Acc: UpdatesAccumulator<Item>> UpdateToVectorDiff<Item, Acc> {
 
                     // Optimisation: we can emit a `VectorDiff::Append` in this particular case.
                     if is_pushing_back && !detaching {
-                        acc.fold([VectorDiff::Append { values: items.into() }]);
+                        acc.extend([VectorDiff::Append { values: items.into() }]);
                     }
                     // No optimisation: let's emit `VectorDiff::Insert`.
                     else {
-                        acc.fold(items.iter().enumerate().map(|(nth, item)| VectorDiff::Insert {
-                            index: offset + nth,
-                            value: item.clone(),
+                        acc.extend(items.iter().enumerate().map(|(nth, item)| {
+                            VectorDiff::Insert { index: offset + nth, value: item.clone() }
                         }));
                     }
                 }
@@ -411,7 +403,7 @@ impl<Item, Acc: UpdatesAccumulator<Item>> UpdateToVectorDiff<Item, Acc> {
 
                     // The chunk length doesn't change.
 
-                    acc.fold([VectorDiff::Set { index: offset, value: item.clone() }]);
+                    acc.extend([VectorDiff::Set { index: offset, value: item.clone() }]);
                 }
 
                 Update::RemoveItem { at: position } => {
@@ -426,7 +418,7 @@ impl<Item, Acc: UpdatesAccumulator<Item>> UpdateToVectorDiff<Item, Acc> {
                     }
 
                     // Let's emit a `VectorDiff::Remove`.
-                    acc.fold([VectorDiff::Remove { index: offset }]);
+                    acc.extend([VectorDiff::Remove { index: offset }]);
                 }
 
                 Update::DetachLastItems { at: position } => {
@@ -469,7 +461,7 @@ impl<Item, Acc: UpdatesAccumulator<Item>> UpdateToVectorDiff<Item, Acc> {
                     self.chunks.clear();
 
                     // Let's straightforwardly emit a `VectorDiff::Clear`.
-                    acc.fold([VectorDiff::Clear]);
+                    acc.extend([VectorDiff::Clear]);
                 }
             }
         }
--- a/crates/matrix-sdk-common/src/linked_chunk/order_tracker.rs
+++ b/crates/matrix-sdk-common/src/linked_chunk/order_tracker.rs
@@ -14,12 +14,28 @@
 
 use std::sync::{Arc, RwLock};
 
+use eyeball_im::VectorDiff;
+
 use super::{
     updates::{ReaderToken, Update, UpdatesInner},
     Position,
 };
 use crate::linked_chunk::{ChunkMetadata, UpdateToVectorDiff};
 
+/// A tracker for the order of items in a linked chunk.
+///
+/// This can be used to determine the absolute ordering of an item, and thus the
+/// relative ordering of two items in a linked chunk, in an
+/// efficient manner, thanks to [`OrderTracker::ordering`]. Internally, it
+/// keeps track of the relative ordering of the chunks themselves; given a
+/// [`Position`] in a linked chunk, the item ordering is the lexicographic
+/// ordering of the chunk in the linked chunk, and the internal position within
+/// the chunk. For the sake of ease, we return the absolute vector index of the
+/// item in the linked chunk.
+///
+/// It requires the full links' metadata to be provided at creation time, so
+/// that it can also give an order for an item that's not loaded yet, in the
+/// context of lazy-loading.
 #[derive(Debug)]
 pub struct OrderTracker<Item, Gap> {
     /// Strong reference to [`UpdatesInner`].
@@ -47,9 +63,11 @@ impl<Item> super::UpdatesAccumulator<Item> for NullAccumulator<Item> {
     fn new(_num_updates_hint: usize) -> Self {
         Self { _phantom: std::marker::PhantomData }
     }
+}
 
-    fn fold(&mut self, _updates: impl IntoIterator<Item = eyeball_im::VectorDiff<Item>>) {
-        // Nothing to do, this is a no-op accumulator.
+impl<Item> Extend<VectorDiff<Item>> for NullAccumulator<Item> {
+    fn extend<T: IntoIterator<Item = VectorDiff<Item>>>(&mut self, _iter: T) {
+        // This is a no-op, as we don't want to accumulate anything.
     }
 }
 
@@ -57,12 +75,16 @@ impl<Item, Gap> OrderTracker<Item, Gap>
 where
     Item: Clone,
 {
-    /// Create a new [`AsOrdering`].
+    /// Create a new [`OrderTracker`].
     ///
-    /// `all_chunks_iterator` is the iterator of all [`Chunk`](super::Chunk)s,
-    /// used to set up its internal state. Note this must include *all* the
-    /// chunks that have ever been known, even in the presence of
-    /// lazy-loading.
+    /// The `all_chunks_metadata` parameter must include the metadata for *all*
+    /// chunks (the full collection, even if the linked chunk is
+    /// lazy-loaded).
+    ///
+    /// They must be ordered by their links in the linked chunk, i.e. the first
+    /// chunk in the vector is the first chunk in the linked chunk, the
+    /// second in the vector is the first's next chunk, and so on. If that
+    /// precondition doesn't hold, then the ordering of items will be undefined.
     pub(super) fn new(
         updates: Arc<RwLock<UpdatesInner<Item, Gap>>>,
         token: ReaderToken,
@@ -79,8 +101,10 @@ where
 
     /// Force flushing of the updates manually.
     ///
-    /// If `inhibit` is `true`, the updates are ignored; otherwise, they are
-    /// consumed normally.
+    /// If `inhibit` is `true` (which is useful in the case of lazy-loading
+    /// related updates, which shouldn't affect the canonical, persisted
+    /// linked chunk), the updates are ignored; otherwise, they are consumed
+    /// normally.
     pub fn flush_updates(&mut self, inhibit: bool) {
         if inhibit {
             // Ignore the updates.
--- a/crates/matrix-sdk/src/event_cache/room/events.rs
+++ b/crates/matrix-sdk/src/event_cache/room/events.rs
@@ -225,7 +225,7 @@ impl RoomEvents {
         self.chunks.items()
     }
 
-    /// Return the order of an event in the room (main) linked chunk.
+    /// Return the order of an event in the room linked chunk.
     ///
     /// Can return `None` if the event can't be found in the linked chunk.
     pub fn event_order(&self, event_pos: Position) -> Option<usize> {


crates/matrix-sdk-common/src/linked_chunk/as_vector.rs
```diff
@@ -83,20 +83,45 @@ impl<Item, Gap> AsVector<Item, Gap> {
...
+/// Interface for a type accumulating updates from [`UpdateToVectorDiff::map`],
+/// and being returned as a result of this.
+pub(super) trait UpdatesAccumulator<Item> {
+    /// Create a new accumulator with a rough estimation of the number of
+    /// updates this accumulator is going to receive.
+    fn new(num_updates_hint: usize) -> Self;
+
+    /// Fold the accumulator with the given new updates.
+    fn fold(&mut self, updates: impl IntoIterator<Item = VectorDiff<Item>>);
+}
```
Hywan: Couldn't we use [`std::default::Default`](https://doc.rust-lang.org/std/default/trait.Default.html) and [`std::iter::Extend`](https://doc.rust-lang.org/std/iter/trait.Extend.html) here? It would avoid re-inventing the wheel probably. Just a personal taste.

I should roughly look like:

```rust
pub(super) trait UpdatesAccumulator<Item>: Default + Extend<Item> {}
```

(a trait might even not be necessary actually)

We loose the `num_updates_hint` though, don't know how impactful it is.

crates/matrix-sdk-common/src/linked_chunk/as_vector.rs
```diff
@@ -83,20 +83,45 @@ impl<Item, Gap> AsVector<Item, Gap> {
...
+
+// Simple implementation for a `Vec<VectorDiff<Item>>` collection for
+// `AsVector<Item, Gap>`.
+impl<Item> UpdatesAccumulator<Item> for Vec<VectorDiff<Item>> {
+    fn new(num_updates_hint: usize) -> Vec<VectorDiff<Item>> {
+        Vec::with_capacity(num_updates_hint)
+    }
+
+    fn fold(&mut self, updates: impl IntoIterator<Item = VectorDiff<Item>>) {
```
Hywan: If you prefer to not use `Default` and `Extend`, please rename `fold` to `extend`:

- `fold` produces a single value by combining a previous value and a new one,
- `extend` takes a mutable reference to a collection and pushes items at its end.

The meanings are too different.

crates/matrix-sdk-common/src/linked_chunk/mod.rs
```diff
@@ -1085,6 +1087,35 @@ impl<const CAP: usize, Item, Gap> LinkedChunk<CAP, Item, Gap> {
...
+    /// compare the relative position of two events in this linked chunk.
+    ///
+    /// A pre-requisite is that the linked chunk has been constructed with
+    /// [`Self::new_with_update_history`], and that if the linked chunk is
+    /// lazily-loaded, an iterator over the fully-loaded linked chunk is
+    /// passed at construction time here.
+    pub fn order_tracker(
+        &mut self,
+        all_chunks_iterator: Option<Iter<'_, CAP, Item, Gap>>,
+    ) -> Option<OrderTracker<Item, Gap>>
```
Hywan: Idea 💡: If the order is defined as `i64` (which can represent negative indexes), you may not need to get an overview of all the chunks or items, thus supporting lazy-loaded as fully-loaded seamlessly.

crates/matrix-sdk-common/src/linked_chunk/order_tracker.rs
```diff
@@ -0,0 +1,476 @@
...
+use std::sync::{Arc, RwLock};
+
+use super::{
+    updates::{ReaderToken, Update, UpdatesInner},
+    Iter, Position,
+};
+use crate::linked_chunk::UpdateToVectorDiff;
+
+#[derive(Debug)]
+pub struct OrderTracker<Item, Gap> {
```
Hywan, line 40: Can we have a documentation for this type please 🥹?

crates/matrix-sdk-common/src/linked_chunk/order_tracker.rs
```diff
@@ -0,0 +1,476 @@
...
+
+impl<Item> super::UpdatesAccumulator<Item> for NullAccumulator<Item> {
+    fn new(_num_updates_hint: usize) -> Self {
+        Self { _phantom: std::marker::PhantomData }
+    }
+
+    fn fold(&mut self, _updates: impl IntoIterator<Item = eyeball_im::VectorDiff<Item>>) {
+        // Nothing to do, this is a no-op accumulator.
+    }
+}
```
Hywan: If we use `Default` + `Extend` as suggested earlier, `NullAccumulator` can be `()` (because, yeah, [`Extend` is implemented for `()`](https://doc.rust-lang.org/std/iter/trait.Extend.html#impl-Extend%3C()%3E-for-())).

Also, removing the need for the special trait `UpdatesAccumulator`, mechanically remove the need for the `PhantomData` (for the `Item` generic).

crates/matrix-sdk-common/src/linked_chunk/as_vector.rs
```diff
@@ -373,7 +378,7 @@ impl<Item, Acc: UpdatesAccumulator<Item>> UpdateToVectorDiff<Item, Acc> {
                     }
 
                     // Optimisation: we can emit a `VectorDiff::Append` in this particular case.
-                    if is_pushing_back && detaching.not() {
+                    if is_pushing_back && !detaching {
```
Hywan, line 390: 🙈

crates/matrix-sdk-common/src/linked_chunk/order_tracker.rs
```diff
@@ -0,0 +1,476 @@
...
+
+impl<Item, Gap> OrderTracker<Item, Gap>
+where
+    Item: Clone,
+{
+    /// Create a new [`AsOrdering`].
```
Hywan:
```suggestion
    /// Create a new [`OrderTracker`].
```

crates/matrix-sdk-common/src/linked_chunk/order_tracker.rs
```diff
@@ -0,0 +1,476 @@
...
+        let _ = self.mapper.map(updates);
+    }
+
+    /// Given an event's position, returns its final ordering in the current
+    /// state of the linked chunk as a vector.
+    ///
+    /// Useful to compare the ordering of multiple events.
+    ///
+    /// Precondition: the reader must be up to date, i.e.
+    /// [`Self::flush_updates`] must have been called before this method.
```
Hywan, line 134: Idea 💡: How annoying it is to call `flush_updates` directly inside this method?

crates/matrix-sdk-common/src/linked_chunk/updates.rs
```diff
@@ -294,6 +294,12 @@ impl<Item, Gap> UpdatesInner<Item, Gap> {
         slice
     }
 
+    /// Has the given reader, identified by its [`ReaderToken`], some pending
+    /// updates, or has it consumed all the pending updates?
+    pub(super) fn is_reader_up_to_date(&self, token: ReaderToken) -> bool {
```
Hywan, line 296: Can we use [`debug_assertions`](https://doc.rust-lang.org/reference/conditional-compilation.html#r-cfg.debug_assertions) here please?

```rust
#[cfg(debug_assertions)]
```

So that it exists only for non-release build. [That's what `debug_assert!` uses internally](https://github.com/rust-lang/rust/blob/a17780db7b8eebbf42a1cbe6bc9cc83013820ba5/library/core/src/macros/mod.rs#L308).

crates/matrix-sdk-common/src/linked_chunk/order_tracker.rs
```diff
@@ -0,0 +1,476 @@
...
+            let _ = updates.take_with_token(token);
+        }
+
+        Self { updates, token, mapper: UpdateToVectorDiff::new(all_chunks_iterator) }
+    }
+
+    /// Force flushing of the updates manually.
+    ///
+    /// If `inhibit` is `true`, the updates are ignored; otherwise, they are
+    /// consumed normally.
```
Hywan: When do we need to use `inhibit = true`?

crates/matrix-sdk/src/event_cache/room/events.rs
```diff
@@ -48,24 +51,27 @@ impl Default for RoomEvents {
...
     /// The provided [`LinkedChunk`] must have been built with update history.
     pub fn with_initial_linked_chunk(
         linked_chunk: Option<LinkedChunk<DEFAULT_CHUNK_CAPACITY, Event, Gap>>,
+        fully_loaded_linked_chunk: Option<LinkedChunk<DEFAULT_CHUNK_CAPACITY, Event, Gap>>,
     ) -> Self {
         let mut linked_chunk = linked_chunk.unwrap_or_else(LinkedChunk::new_with_update_history);
 
         let chunks_updates_as_vectordiffs = linked_chunk
             .as_vector()
-            // SAFETY: The `LinkedChunk` has been built with `new_with_update_history`, so
```
Hywan, line 64: Why removing the `SAFETY` comment?

crates/matrix-sdk/src/event_cache/room/events.rs
```diff
@@ -231,6 +255,35 @@ impl RoomEvents {
         self.chunks.items()
     }
 
+    /// Return the order of an event in the room (main) linked chunk.
```
Hywan: You can remove “(main)” I think:

- it can only create confusion,
- to me, it's clear that a room has a linked chunk, and that each thread will have its own linked chunk,
- you are saying “in the room”, so we know it's not for the thread.

The comment is already crystal clear, you don't need “(main)”.

crates/matrix-sdk/src/event_cache/room/mod.rs
```diff
@@ -530,30 +533,39 @@ mod private {
...
 
             let linked_chunk_id = LinkedChunkId::Room(&room_id);
-            let linked_chunk = match store_lock
+
+            // Load the full linked chunk, so as to feed the order tracker.
+            //
+            // If loading the full linked chunk failed, we'll clear the event cache, as it
+            // indicates that at some point, there's some malformed data.
+            let fully_loaded_linked_chunk =
+                match Self::load_full_linked_chunk(&*store_lock, linked_chunk_id).await {
```
Hywan: It seems we are loading all chunks with all their items. It seems to be quite annoying and quite a blocker to me. It will impact performance too much. At best we must only load the chunks **without** their items.

crates/matrix-sdk/src/event_cache/room/events.rs
bnjbvr: Because the `expect()` says the same thing one line below, so there's no real value in repeating it in a code comment IMO.

bnjbvr:
Thanks for the review! These are very valid questions.

> I'm wondering how the event order will be used.

The requirement is that we want to be able to compare the relative position of two events, *be they loaded in the in-memory linked chunk or not*. So, this must work, independently of the actual position of the event.

Some alternative ways to build this:

- as you suggest, the most obvious optimization is to *not* load all chunks with their events, but only some "metadata" (prev/next/number of items in it).
- we could also move the entire computation inside the store trait, and make it stateless; in this case, comparing the relative positions of two events would be equivalent to finding their owning chunks, i.e. find the relative ordering of the chunks (they belong to). But we have a linked list of chunks, here, so we can't immediately answer that query without going through the next/prev link of one of the two chunks. I suspect the optimal algorithm would be a "ping-pong" strategy: start with the next; if it's not this one, continue to the previous; if it's not this one, continue with the next's next; if if it's not this one, continue with the previous' previous, and so on. But worst case scenario, we now have to iterate over *all* the chunks every time we want to order two events.
  - Could be mitigated by keeping a local cache for the ordering of chunks, that's blown up as soon as we add / remove a chunk, maybe… but at this point, we might as well keep the current solution, that constructs the full ordering, then maintains it over time.
- maybe there's a way to have a mixed, incremental solution where we only get the initial chunk in the total ordering; every time we don't find the chunk owning an event in the local state, we'd lazily load the previous chunk in the order tracker (and maybe we could use negative indices as you suggest). Worst case, we'd get to the initial chunk (but then we'd have a fully-loaded order tracker, which will be faster every time it's called subsequently). Now, this might be tricky because we don't want to lazy-load *the room's linked chunk* just so as to maintain the order tracking up to date; but if the room's linked chunk *does* lazy-load, we want the order tracker to benefit from this. Grmpf.

It sounds like the first and last solutions might be preferable in the short term, and would both require loading only the metadata of a chunk. So I could start with this.

Then, I find the concern around performance totally legit. The best way to resolve it would be benchmarking or using some real-world measures, so I may look into this and get back to you here.

> Are you loading all the chunks to always return the position of an event? It means you want the position of an event from the store, not from the in-memory room event cache? If this is the case, I think the strategy should entirely move onto the EventCacheStore store trait. If that's not the case, I don't understand why we need to load all the chunks. Can you explain to me please?

As said above, a solution only based on the `EventCacheStore` trait would likely be slow, as in the worst case you go through the entire prev/next chain of nodes in the linked list (in sqlite, that's probably one query per "deref").

Loading all the chunks at start seemed like a reasonable solution to build the initial state, and then maintain it cleanly over time. But yeah, at the very least we'd need to only load the minimal amount of data for the order tracker to work correctly.

crates/matrix-sdk-common/src/linked_chunk/order_tracker.rs
bnjbvr: It's annoying, because it requires the method to take a `&mut self` instead of `&self`, and it propagates through the caller's chain. The first version of my patch did that, but it was painful and caused bad smells (like, the debug functionality now required a `&mut RoomEventCacheState` :weary:)

bnjbvr:
We've discussed this, and lazy loading the order tracker brings many other complications, so we're going to roll with the first suggestion only (load a limited set of metadata about each chunk of a linked chunk, and use that instead of the fully loaded linked chunk), and see what comes out of that in terms of performance. We can load the entire metadata in a single SQL query, so that ought to be rather efficient.

crates/matrix-sdk/src/event_cache/room/events.rs
bnjbvr: You're right! I suppose I've confused myself with respect to the "main conversation thread", but the room is so much more than that…

crates/matrix-sdk-common/src/linked_chunk/as_vector.rs
```diff
@@ -137,6 +138,17 @@ impl<Item, Acc: UpdatesAccumulator<Item>> UpdateToVectorDiff<Item, Acc> {
         Self { chunks: initial_chunk_lengths, _phantom: std::marker::PhantomData }
     }
 
+    /// Construct [`UpdateToVectorDiff`], based on a linked chunk's full
+    /// metadata, used to set up its own internal state.
```
bnjbvr, line 135: nit: add that the metadata must be correctly ordered

crates/matrix-sdk-common/src/linked_chunk/as_vector.rs
bnjbvr: The `num_updates_hint` is nice for getting a good approximation for the number of updates in the case of the `AsVector` mapper, so I'll keep this one; good eye on the `std::iter::Extend`, though!

crates/matrix-sdk-common/src/linked_chunk/order_tracker.rs
bnjbvr: hmm, `Extend` is implemented for `()` only when the extended type is `()` too; in this case, we want to extend with `VectorDiff<Item>`, so it doesn't plain work. I've kept the `NullAccumulator` struct, since this is an internal impl detail and we shouldn't leak the `UpdatesAccumulator` impl for `()` in general.

bnjbvr:
Rebased on top of main, but only the last three commits are really new. Time for another round of review \o/

crates/matrix-sdk-common/src/linked_chunk/updates.rs
bnjbvr: Actually we can't: https://github.com/matrix-org/matrix-rust-sdk/actions/runs/15900921446/job/44843483636?pr=5225

`debug_assert!` uses a `if cfg!` internally, so this means the code below must exist, even if it's not called.

crates/matrix-sdk-common/src/linked_chunk/as_vector.rs
```diff
@@ -137,6 +138,17 @@ impl<Item, Acc: UpdatesAccumulator<Item>> UpdateToVectorDiff<Item, Acc> {
...
+    ///
+    /// See [`Self::map`] to learn more about the algorithm.
+    pub fn from_metadata(metas: Vec<ChunkMetadata>) -> Self {
+        let initial_chunk_lengths =
+            metas.into_iter().map(|meta| (meta.identifier, meta.num_items)).collect();
```
Hywan, line 144: Not a blocker!

We are doing a double allocation here. One `collect` in `from_metadata`, and one `collect` in `LinkedChunk::order_tracker`. I propose the following:

```suggestion
    pub fn from_metadata<M>(metas: M) -> Self
        where M: Iterator<Item = ChunkMetadata>
    {
        let initial_chunk_lengths =
            metas.map(|meta| (meta.identifier, meta.num_items)).collect();
```

and in `order_tracker`, we pass the iterator directly, without calling `collect`. I don't know how it will work with the `unwrap_or_else` though.

crates/matrix-sdk-common/src/linked_chunk/updates.rs
Hywan: Ohhhh, damn…

crates/matrix-sdk-common/src/linked_chunk/as_vector.rs
bnjbvr: Hmm, it's a good idea and I've tried it, but the one caller looks like this:

```rust
        Some(OrderTracker::new(
            updates,
            token,
            all_chunks.unwrap_or_else(|| {
                // Consider the linked chunk as fully loaded.
                self.chunks()
                    .map(|chunk| ChunkMetadata {
                        identifier: chunk.identifier(),
                        num_items: chunk.num_items(),
                        previous: chunk.previous().map(|prev| prev.identifier()),
                        next: chunk.next().map(|next| next.identifier()),
                    })
                    .collect()
            }),
        ))
```

where `all_chunks` is a `Option<Vec<ChunkMetadata>>`. If I wanted `OrderTracker::new()` to accept an `Iterator<Item = ChunkMetadata>` as the last parameter, then I'd have some code like this:

```rust
all_chunks.map(Vec::into_iter).unwrap_or_else(|| {
                // Consider the linked chunk as fully loaded.
                self.chunks()
                    .map(|chunk| ChunkMetadata {
                        identifier: chunk.identifier(),
                        num_items: chunk.num_items(),
                        previous: chunk.previous().map(|prev| prev.identifier()),
                        next: chunk.next().map(|next| next.identifier()),
                    })
            }
```

But now, you can see what the problem is:

- when `all_chunks` is `Some`, then the iterator has the concrete type `vec::IntoIter`
- and when it's `None`, then the iterator has the concrete type `Map<...>`

so the compiler complains (correctly) they're not the same type. I *think* I recall some Rust proposals to basically allow this, but it's not a reality right now, unfortunately. If we found a better solution later, happy to rewrite this code in another way that would make this possible, of course.

PR #5226, opened 2025-06-12
IndexedDB: Add (de)serialization functionality for `EventCacheStore` backend

mgoldenberg:
## Background

This pull request is part of a series of pull requests to add an IndexedDB implementation of the `EventCacheStore` (see #4617, #4996, #5090, #5138). This particular pull request provides (de)serialization functionality for the types defined in #5138.

## Changes

This change add a few traits to aid in (de)serialization and implements those traits on a relevant types. Furthermore, there is a `struct` for conveniently calling into those (de)serialization traits - i.e., `IndexeddbEventCacheStoreSerializer`.

### `trait`s and `impl`s

- `Indexed` - converts high-level types to/from indexed types
    - `impl`s:
        - `Chunk` ↔ `IndexedChunk`
        - `Event` ↔ `IndexedEvent`
        - `Gap` ↔ `IndexedGap`
- `IndexedKey` - encodes a key for an indexed type
    - `impl`s:
        - `Chunk`
            - `IndexedChunkIdKey`
            - `IndexedNextChunkIdKey`
        - `Event`
            - `IndexedEventIdKey`
            - `IndexedEventPositionKey`
            - `IndexedEventRelationKey`
        - `Gap`
            - `IndexedGapIdKey`   
- `IndexedKeyBounds` - encodes the lower and upper keys for an indexed type
    - `impl`s:
        - `Chunk`
            - `IndexedChunkIdKey`
            - `IndexedNextChunkIdKey`
        - `Event`
            - `IndexedEventIdKey`
            - `IndexedEventPositionKey`
            - `IndexedEventRelationKey`
        - `Gap`
            - `IndexedGapIdKey` 

## Future Work

- Add convenient query functions that wrap (de)serialization logic
- Add implementation of `EventCacheStore` which uses query functions mentioned above

---

### DIFF
--- a/crates/matrix-sdk-indexeddb/src/event_cache_store/migrations.rs
+++ b/crates/matrix-sdk-indexeddb/src/event_cache_store/migrations.rs
@@ -120,6 +120,8 @@ pub mod v1 {
         pub const EVENTS_POSITION_KEY_PATH: &str = "position";
         pub const EVENTS_RELATION: &str = "events_relation";
         pub const EVENTS_RELATION_KEY_PATH: &str = "relation";
+        pub const EVENTS_RELATION_RELATED_EVENTS: &str = "events_relation_related_event";
+        pub const EVENTS_RELATION_RELATION_TYPES: &str = "events_relation_relation_type";
         pub const GAPS: &str = "gaps";
         pub const GAPS_KEY_PATH: &str = "id";
     }
--- a/crates/matrix-sdk-indexeddb/src/event_cache_store/migrations.rs
+++ b/crates/matrix-sdk-indexeddb/src/event_cache_store/migrations.rs
@@ -109,6 +109,7 @@ pub mod v1 {
 
     pub mod keys {
         pub const CORE: &str = "core";
+        pub const ROOMS: &str = "rooms";
         pub const LINKED_CHUNKS: &str = "linked_chunks";
         pub const LINKED_CHUNKS_KEY_PATH: &str = "id";
         pub const LINKED_CHUNKS_NEXT: &str = "linked_chunks_next";
--- a/crates/matrix-sdk-indexeddb/src/event_cache_store/migrations.rs
+++ b/crates/matrix-sdk-indexeddb/src/event_cache_store/migrations.rs
@@ -20,20 +20,26 @@ use thiserror::Error;
 use wasm_bindgen::JsValue;
 use web_sys::{DomException, IdbIndexParameters};
 
-const CURRENT_DB_VERSION: Version = Version::V1;
+/// The current version and keys used in the database.
+pub mod current {
+    use super::{v1, Version};
+
+    pub const VERSION: Version = Version::V1;
+    pub use v1::keys;
+}
 
 /// Opens a connection to the IndexedDB database and takes care of upgrading it
 /// if necessary.
 #[allow(unused)]
 pub async fn open_and_upgrade_db(name: &str) -> Result<IdbDatabase, DomException> {
-    let mut request = IdbDatabase::open_u32(name, CURRENT_DB_VERSION as u32)?;
+    let mut request = IdbDatabase::open_u32(name, current::VERSION as u32)?;
     request.set_on_upgrade_needed(Some(|event: &IdbVersionChangeEvent| -> Result<(), JsValue> {
         let mut version =
             Version::try_from(event.old_version() as u32).map_err(DomException::from)?;
-        while version < CURRENT_DB_VERSION {
+        while version < current::VERSION {
             version = match version.upgrade(event.db())? {
                 Some(next) => next,
-                None => CURRENT_DB_VERSION, /* No more upgrades to apply, jump forward! */
+                None => current::VERSION, /* No more upgrades to apply, jump forward! */
             };
         }
         Ok(())
--- a/crates/matrix-sdk-indexeddb/src/event_cache_store/serializer/traits.rs
+++ b/crates/matrix-sdk-indexeddb/src/event_cache_store/serializer/traits.rs
@@ -63,3 +63,14 @@ pub trait IndexedKey<T: Indexed> {
         serializer: &IndexeddbSerializer,
     ) -> Self;
 }
+
+/// A trait for constructing the bounds of an [`IndexedKey`].
+///
+/// This is useful when constructing range queries in IndexedDB.
+pub trait IndexedKeyBounds<T: Indexed>: IndexedKey<T> {
+    /// Encodes the lower bound of the key.
+    fn encode_lower(room_id: &RoomId, serializer: &IndexeddbSerializer) -> Self;
+
+    /// Encodes the upper bound of the key.
+    fn encode_upper(room_id: &RoomId, serializer: &IndexeddbSerializer) -> Self;
+}
--- a/crates/matrix-sdk-indexeddb/src/event_cache_store/serializer/traits.rs
+++ b/crates/matrix-sdk-indexeddb/src/event_cache_store/serializer/traits.rs
@@ -1,0 +1,44 @@
+// Copyright 2025 The Matrix.org Foundation C.I.C.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License
+
+use ruma::RoomId;
+
+use crate::serializer::IndexeddbSerializer;
+
+/// A conversion trait for preparing high-level types into indexed types
+/// which are better suited for storage in IndexedDB.
+///
+/// Note that the functions below take an [`IndexeddbSerializer`] as an
+/// argument, which provides the necessary context for encryption and
+/// decryption, in the case the high-level type must be encrypted before
+/// storage.
+pub trait Indexed: Sized {
+    /// The indexed type that is used for storage in IndexedDB.
+    type IndexedType;
+    /// The error type that is returned when conversion fails.
+    type Error;
+
+    /// Converts the high-level type into an indexed type.
+    fn to_indexed(
+        &self,
+        room_id: &RoomId,
+        serializer: &IndexeddbSerializer,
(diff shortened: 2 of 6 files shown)


crates/matrix-sdk-indexeddb/src/event_cache_store/serializer/traits.rs
```diff
@@ -42,3 +42,24 @@ pub trait Indexed: Sized {
         serializer: &IndexeddbSerializer,
     ) -> Result<Self, Self::Error>;
 }
+
+/// A trait for encoding types which will be used as keys in IndexedDB.
+///
+/// Each implementation represents a key on an [`Indexed`] type.
+pub trait IndexedKey<T: Indexed> {
+    /// Any extra data used to construct the key.
+    type KeyComponents: ?Sized;
```
Hywan: Why `?Sized` is necessary here?

crates/matrix-sdk-indexeddb/src/event_cache_store/serializer/traits.rs
```diff
@@ -63,3 +63,14 @@ pub trait IndexedKey<T: Indexed> {
...
+
+/// A trait for constructing the bounds of an [`IndexedKey`].
+///
+/// This is useful when constructing range queries in IndexedDB.
+pub trait IndexedKeyBounds<T: Indexed>: IndexedKey<T> {
+    /// Encodes the lower bound of the key.
+    fn encode_lower(room_id: &RoomId, serializer: &IndexeddbSerializer) -> Self;
+
+    /// Encodes the upper bound of the key.
+    fn encode_upper(room_id: &RoomId, serializer: &IndexeddbSerializer) -> Self;
```
Hywan, line 75: I wonder if we could have a single method here. I'll see in this patch if each method is used alone, or if they are always used together.

crates/matrix-sdk-indexeddb/src/event_cache_store/migrations.rs
```diff
@@ -20,20 +20,26 @@ use thiserror::Error;
 use wasm_bindgen::JsValue;
 use web_sys::{DomException, IdbIndexParameters};
 
-const CURRENT_DB_VERSION: Version = Version::V1;
+/// The current version and keys used in the database.
+pub mod current {
+    use super::*;
```
Hywan: I think you only need `Version` and `v1` right?

If possible, we prefer to avoid `*` as it can import too much things.

```suggestion
    use super::{v1, Version};
```

crates/matrix-sdk-indexeddb/src/event_cache_store/serializer/types.rs
```diff
@@ -84,6 +154,41 @@ pub enum IndexedNextChunkIdKey {
...
+    ) -> Self {
+        next_chunk_id
+            .map(|id| {
+                Self::Some(<IndexedChunkIdKey as IndexedKey<Chunk>>::encode(
+                    room_id, &id, serializer,
+                ))
+            })
+            .unwrap_or_else(|| {
+                let room_id = serializer.encode_key_as_string(keys::ROOMS, room_id);
+                Self::None((room_id,))
```
Hywan: Why do you need a unary tuple for `None`? Can't it be simplified as `Self::None(room_id)`, just like the `Some` variant?

crates/matrix-sdk-indexeddb/src/event_cache_store/serializer/types.rs
```diff
@@ -29,18 +29,22 @@
...
         migrations::current::keys,
         serializer::traits::{Indexed, IndexedKey, IndexedKeyBounds},
-        types::Chunk,
+        types::{Chunk, Event, Position},
     },
     serializer::{IndexeddbSerializer, MaybeEncrypted},
 };
 
+const INDEXED_KEY_LOWER_CHARACTER: char = '\u{0000}';
+const INDEXED_KEY_UPPER_CHARACTER: char = '\u{FFFF}';
```
Hywan: Why these values? The explanation should ideally be a code comment :-).

crates/matrix-sdk-indexeddb/src/event_cache_store/serializer/traits.rs
Hywan: I think they are always used together. Up to you to decide if it's worth merging them together.

crates/matrix-sdk-indexeddb/src/event_cache_store/serializer/traits.rs
mgoldenberg: It is because the `KeyComponent` for an `IndexedEventIdKey` is an `EventId` (see [here](https://github.com/matrix-org/matrix-rust-sdk/blob/2ac6751ede6be78df7952e039dc8a316610ad52c/crates/matrix-sdk-indexeddb/src/event_cache_store/serializer/types.rs#L263-L271)). I believe the loosened restriction is due to the fact that `EventId` is a wrapper around `str`, rather than `&str`.

I could also change the `KeyComponent` for an `IndexedEventIdKey` to be `OwnedEventId`. Does that sound good?

crates/matrix-sdk-indexeddb/src/event_cache_store/serializer/traits.rs
mgoldenberg: There are some places that they will be used separately in future pull requests. That being said, we could do either of the following to make this a little more concise.

1. Return a tuple of the bounds and simply discard one of them if we don't use both - a bit wasteful, but not too bad.
2. Add an `enum`, say `Bound`, that we pass into the function to let it know if we want `Bound::Upper` or `Bound::Lower`

Thoughts or preferences on this?

crates/matrix-sdk-indexeddb/src/event_cache_store/serializer/types.rs
mgoldenberg: So, there's a [comment](https://github.com/matrix-org/matrix-rust-sdk/blob/106b9c004af23529d05c420f2782e007242a5045/crates/matrix-sdk-indexeddb/src/event_cache_store/serializer/types.rs#L139-L145) on the type about why this is necessary.

Basically, we need the two variants to serialize into JavaScript arrays because IndexedDB requires that keys be the same type. If we were to simply pass the `RoomId`, we would get a key that serializes into a JavaScript string. 

The alternative would be to write a custom (de)serializer, but that is a bit involved and, ultimately, these types will likely not even be exposed outside the `serializer` module.

That being said, I could add a function `IndexedNextChunkIdKey::none(RoomId)` that would abstract this logic and make it more readable.

What do you think?

crates/matrix-sdk-indexeddb/src/event_cache_store/migrations.rs
mgoldenberg: fixup: [`25975b48d817a01a9f633381511603ff0936931e`](https://github.com/matrix-org/matrix-rust-sdk/pull/5226/commits/25975b48d817a01a9f633381511603ff0936931e)

crates/matrix-sdk-indexeddb/src/event_cache_store/serializer/types.rs
mgoldenberg: fixup: [`1efe4874fcc62658a35a34584072932781894c9a`](https://github.com/matrix-org/matrix-rust-sdk/pull/5226/commits/1efe4874fcc62658a35a34584072932781894c9a)

crates/matrix-sdk-indexeddb/src/event_cache_store/serializer/traits.rs
Hywan: I think it's better to keep the `Sized` property here, so please, change for `OwnedEventId`. We don't particularly need dyn compatibility here, but I rather suspect the compiler can run more optimisations if the size of the types are known.

crates/matrix-sdk-indexeddb/src/event_cache_store/serializer/traits.rs
Hywan: Let's keep 2 methods, it's fine!

crates/matrix-sdk-indexeddb/src/event_cache_store/serializer/types.rs
Hywan: Ah, okay it makes sense. I'm in favor for your `IndexedNextChunkIdKey::none()` method 👍.

mgoldenberg: `EventId` → `OwnedEventId`, fixup: [`2151ab7676244d71028df24f5ef73efaf17bd458`](https://github.com/matrix-org/matrix-rust-sdk/pull/5226/commits/2151ab7676244d71028df24f5ef73efaf17bd458)
`KeyComponents: ?Sized` → `KeyComponents`, fixup: [`49cb8e1d6d10f004fb42e19b9256094b4f61952e`](https://github.com/matrix-org/matrix-rust-sdk/pull/5226/commits/49cb8e1d6d10f004fb42e19b9256094b4f61952e)

crates/matrix-sdk-indexeddb/src/event_cache_store/serializer/traits.rs
mgoldenberg: `EventId` → `OwnedEventId`, fixup: [`2151ab7676244d71028df24f5ef73efaf17bd458`](https://github.com/matrix-org/matrix-rust-sdk/pull/5226/commits/2151ab7676244d71028df24f5ef73efaf17bd458)
`KeyComponents: ?Sized` → `KeyComponents`, fixup: [`49cb8e1d6d10f004fb42e19b9256094b4f61952e`](https://github.com/matrix-org/matrix-rust-sdk/pull/5226/commits/49cb8e1d6d10f004fb42e19b9256094b4f61952e)

crates/matrix-sdk-indexeddb/src/event_cache_store/serializer/types.rs
mgoldenberg: Add `IndexedNextChunkIdKey::none`, fixup: [`438c709f7e9672f2380703d6240be28ca1c12a84`](https://github.com/matrix-org/matrix-rust-sdk/pull/5226/commits/438c709f7e9672f2380703d6240be28ca1c12a84)
Use `IndexedNextChunkidKey::none` in `IndexedNextChunkIdKey::encode`, fixup: [`3bd8b6ec9c1a3b4c06018564399b2c1449b08165`](https://github.com/matrix-org/matrix-rust-sdk/pull/5226/commits/3bd8b6ec9c1a3b4c06018564399b2c1449b08165)

Hywan:
Can you rebase on top of `main` to get the new code coverage infra please?

PR #5227, opened 2025-06-13
feat: add thread support to the room message draft facilities

### LINKED ISSUE #5223: Handle drafts for threaded timelines
The draft storage facilities that the rooms currently offer are limited to one per room and now, with the introduction of threaded timelines, that's no longer enough.

A simple solution would be to expand the `ComposerDraft` StateStore key to include an optional thread root and export that all the way up to the FFI layer interfaces.

stefanceriu:
This patch adds optional thread root event id parameters to the drafting functions exposed on the room level allowing unfinished messages to be managed for the main room as well as any inner thread.

Fixes #5223

### DIFF
--- a/bindings/matrix-sdk-ffi/src/room.rs
+++ b/bindings/matrix-sdk-ffi/src/room.rs
@@ -828,18 +828,31 @@ impl Room {
 
     /// Store the given `ComposerDraft` in the state store using the current
     /// room id, as identifier.
-    pub async fn save_composer_draft(&self, draft: ComposerDraft) -> Result<(), ClientError> {
-        Ok(self.inner.save_composer_draft(draft.try_into()?).await?)
+    pub async fn save_composer_draft(
+        &self,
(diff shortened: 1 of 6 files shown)


andybalaam (approved):
This doesn't break clients with existing stores, because any existing drafts will load up fine and just be unthreaded (right?).

I would very mildly prefer a struct instead of the `(room_id, thread_id)` tuple, that could be used in `StateStoreDataKey::ComposerDraft` and as the key of the `composer_drafts` hashmap, but I won't insist on it.

Looks great.

stefanceriu:
> This doesn't break clients with existing stores

Yes, that should be the case as thread root is unwrapped before creating the key variants (except for the in memory store where well.. it's all in memory anyway)

> I would very mildly prefer a struct

I honestly tought about it but felt like one step too many, at least for now. I think I'm going to leave it like this and see how/if it grows over time.

PR #5229, opened 2025-06-16
Valere/crypto/widget send encrypted to device

BillCarsonFr:
Based on https://github.com/matrix-org/matrix-rust-sdk/pull/5116

Adds the capability for widgets to send encrytped to-device messages properly.

It is getting rid of the old `encrypted` boolean that was used by widget to request encryption. As said by the base MSC 
for widget send/receive messages: Widgets should not be aware of encryption.
If the widget is in an encrytped room then to-device traffic will be encrypted.

Signed-off-by:


BillCarsonFr:
Something went wrong with all the merge and rebase, will create a new branch and cherry pick

BillCarsonFr:
re-opened here https://github.com/matrix-org/matrix-rust-sdk/pull/5252

PR #5231, opened 2025-06-16
feat(wasm): Address use of errors and panic::resume_unwind for wasm targets

zzorba:
Address use of anyhow's error context + use of panic::resume_unwind on wasm targets.

Signed-off-by: Daniel Salinas

### DIFF
--- a/bindings/matrix-sdk-ffi/src/timeline/mod.rs
+++ b/bindings/matrix-sdk-ffi/src/timeline/mod.rs
@@ -634,7 +634,7 @@ impl Timeline {
         self.inner
             .fetch_details_for_event(event_id)
             .await
-            .map_err(|_| ClientError::from_str("Fetching event details".to_owned(), None))?;
+            .map_err(|e| ClientError::from_str(e, Some("Fetching event details".to_owned())))?;
         Ok(())
     }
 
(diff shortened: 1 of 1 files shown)


bindings/matrix-sdk-ffi/src/timeline/mod.rs
```diff
@@ -634,7 +631,10 @@ impl Timeline {
 
     pub async fn fetch_details_for_event(&self, event_id: String) -> Result<(), ClientError> {
         let event_id = <&EventId>::try_from(event_id.as_str())?;
-        self.inner.fetch_details_for_event(event_id).await.context("Fetching event details")?;
+        self.inner
+            .fetch_details_for_event(event_id)
+            .await
+            .map_err(|_| ClientError::from_str("Fetching event details".to_owned(), None))?;
```
poljar: This isn't quite the same, right? Previously we were returning the error `fetch_details_for_event()` was returning but adding some context to the error.

Now we dropped the original error completely. Why not pass in the error to `from_str()` and put the context into the second argument instead of `None`.

Why not

```rust
.map_err(|e| ClientError::from_str(e, Some("Fetching event details".to_owned()))?;
```

poljar (dismissed):
This could have been handled a bit better, as it is this is dropping potentially useful information.

Could you also explain what the problem with `anyhow` on Wasm is? Sorry if you explained this in another PR already, I haven't seen it.

bindings/matrix-sdk-ffi/src/timeline/mod.rs
zzorba: You are right, this is much better.  I think when I first wrote these changes the ClientError convenience constructors were not available yet, and I haven't internalized the full set of options.

zzorba:
Oh sorry, the issue with anyhow's `context` is that it is requiring a `Send`, which is not present on the underlying Errors on Wasm platforms.  The .context on `Option` works fine, but the .context for `Result` does not

PR #5236, opened 2025-06-17
refactor(crypto): Extract a method for handling encrypted to-device events

andybalaam:
This is a cut-and-paste of the logic from within a match branch into a separate method: the diff doesn't necessarily make that easy to see but I promise I didn't change anything.

This is purely because I found the existing method too long for me to understand easily, and because I may possibly want to unit test the new method as part of my work.

### DIFF
--- a/crates/matrix-sdk-crypto/src/machine/mod.rs
+++ b/crates/matrix-sdk-crypto/src/machine/mod.rs
@@ -89,13 +89,14 @@ use crate::{
             room::encrypted::{
                 EncryptedEvent, EncryptedToDeviceEvent, RoomEncryptedEventContent,
                 RoomEventEncryptionScheme, SupportedEventEncryptionSchemes,
+                ToDeviceEncryptedEventContent,
             },
             room_key::{MegolmV1AesSha2Content, RoomKeyContent},
             room_key_bundle::RoomKeyBundleContent,
             room_key_withheld::{
                 MegolmV1AesSha2WithheldContent, RoomKeyWithheldContent, RoomKeyWithheldEvent,
             },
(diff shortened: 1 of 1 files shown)


crates/matrix-sdk-crypto/src/machine/mod.rs
```diff
@@ -1385,87 +1385,100 @@ impl OlmMachine {
...
+            Err(err) => {
+                if let OlmError::SessionWedged(sender, curve_key) = err {
+                    if let Err(e) =
+                        self.inner.session_manager.mark_device_as_wedged(&sender, curve_key).await
+                    {
                         error!(
-                            error = ?err,
-                            "Couldn't check whether event is from dehydrated device",
+                            error = ?e,
+                            "Couldn't mark device from to be unwedged",
```
poljar:
```suggestion
                            "Couldn't mark device to be unwedged",
```

crates/matrix-sdk-crypto/src/machine/mod.rs
```diff
@@ -1385,87 +1385,100 @@ impl OlmMachine {
...
+                    session = ?decrypted.session,
+                    "Received a to-device event from a dehydrated device. This is unexpected: ignoring event"
+                );
+                return None;
+            }
+            Ok(false) => {}
+            Err(err) => {
+                error!(
+                    error = ?err,
+                    "Couldn't check whether event is from dehydrated device",
```
poljar:
```suggestion
                    "Couldn't check whether the event is from a dehydrated device",
```

poljar:
If you rebase this PR onto main, you'll get your coverage running again.

PR #5240, opened 2025-06-17
feat(wasm): Fix unwrap error on Wasm platforms caused by UInt::MAX conversions

zzorba:
UInt::MAX.try_into().unwrap() was causing errors on Wasm platforms, due to the result being unrepresentable. 

This `unwrap_or` was also always being calculated regardless, so I think using `usize::MAX` is preferable on all platforms.

Signed-off-by: Daniel Salinas


crates/matrix-sdk-common/src/serde_helpers.rs
```diff
@@ -80,9 +82,9 @@ pub fn extract_bundled_thread_summary(
     match event.get_field::<Unsigned>("unsigned") {
         Ok(Some(Unsigned { relations: Some(Relations { thread: Some(bundled_thread) }) })) => {
             // Take the count from the bundled thread summary, if available. If it can't be
-            // converted to a `u64`, we use `UInt::MAX` as a fallback, as this is unlikely
+            // converted to a `u64`, we use `usize::MAX` as a fallback, as this is unlikely
```
jplatte: This doesn't make sense. `UInt::MAX` is always representable as `u64`. Looks like this comment has been wrong about what happens here for a while.


```suggestion
            // converted to a `usize`, we use `usize::MAX` as a fallback, as this is unlikely
```

Hywan: `count` is a `usize` I guess. Your suggestion _is_ the correct version of the comment.

jplatte: Well, this is an _addition_ to the edit made by @zzorba. This is fixing a pre-existing issue with the comment that @zzorba missed, does not undo any of his changes.

bnjbvr: I think an even better fix would be to make the `ThreadSummary::num_replies` a fixed-width type, like `u32`; depending on platform-specific sized types isn't a good practice, especially as it means we have different semantics per platform that we need to think about.

(Good catch on the bug, though!)

Hywan: > Well, this is an _addition_ to the edit made by @zzorba. This is fixing a pre-existing issue with the comment that @zzorba missed, does not undo any of his changes.

Yes absolutely. My phrasing was probably too vague. I think your suggestion should be applied.

Hywan (dismissed):
How comfortable do you feel to address https://github.com/matrix-org/matrix-rust-sdk/pull/5240#discussion_r2154141592?

zzorba:
Suggestion is to change to u32? 

I can do that but feels like it should be a separate PR versus the regression here.

bnjbvr:
Yeah, I think `u32` would be nice. I don't think it would cause many changes, so I don't think it warrants its own PR, but it's up to you at the end of the day :) (I'd be happy to give it a quick review, in any case.)

zzorba:
@bnjbvr updated with the shift from usize to u32.

It did require one additional `try_into().unwrap` since it is normally calculated based on the `len()` of a vec.

zzorba:
Looks like the CI flaked out during the github outage

jplatte:
Restarted the failed coverage thing, but not sure if it's really gonna succeed before https://github.com/matrix-org/matrix-rust-sdk/pull/5244 or something similar is merged.

bnjbvr:
@zzorba I think this needs a rebase, and this should be good to go!

jplatte:
I think close + reopen should also do it?

jplatte:
No clue what happened in CI there. It used the new workflow but didn't find the right Cargo profile it seems. Hopefully rebase will fix it.

poljar:
> No clue what happened in CI there. It used the new workflow but didn't find the right Cargo profile it seems. Hopefully rebase will fix it.

I think it's using the action and the xtask after the migration to `llvm-cov` but the `.config/nextest.toml` file from this PR, which doesn't have a `ci` profile.

jplatte:
But.. how? It should be checking out a merge commit of this PR merged into `main` (at the time of last push or opening PR or something). It should not be possible for it to use _some_ of the current files from `main` and some old files that weren't even changed in this PR 😅

poljar:
> But.. how? It should be checking out a merge commit of this PR merged into `main` (at the time of last push or opening PR or something). It should not be possible for it to use _some_ of the current files from `main` and some old files that weren't even changed in this PR 😅

I don't know, I'm just trying to parse what it looks like :sweat_smile:

zzorba:
Should I rebase this to pick up the CI fixes at main?

PR #5242, opened 2025-06-18
feat(wasm): Add lib to matrix-sdk-ffi target

zzorba:
The uniffi tool for generating JS/Wasm bindings utilizes rust as its intermediate language.  

As a result, the 'target' uniffi annotated library needs to be marked as a 'lib' so that the generated rust code can utilize it to generate the Wasm create + typescript bindings.

Signed-off-by: Daniel Salinas


jplatte (approved):
Okay, that PR description helps. Maybe I just missed where you explained this before. I would love if you could make this a multi-line array with explanations what each individual crate-type is for, but I think this can merge without that too. (I'll let the actual team do the merging though)

zzorba:
So`staticlib` is used by iOS and Android, `lib` is used for Wasm.  But I don't know what actually uses `cdylib`, do you?

jplatte:
I seem to remember it was staticlib -> iOS, cdylib -> android.

zzorba:
Yep, just read the uniffi docs and that's what it says.  I think my android react-native tool might be using the static lib, which was the source of my confusion.  Updated the Cargo.toml file with comments.

poljar:
Could you rebase this to get the CI going?

zzorba:
Rebased

PR #5244, opened 2025-06-18
 Switch to llvm-cov for coverage reports

poljar:
A continuation of #5241, somehow caches got messed up with that PR and CI was completely borked.

Let's see if the suffering continues or we get somewhere with this.


Hywan (approved):
Looks good to me, thanks!

PR #5211, opened 2025-06-10
feat(wasm): Change matrix-sdk-ffi to rely on features over platform targets

zzorba:
The system of platform targets was already quite messy, and becoming even worse as we start preparing for Wasm support. Switch to features instead to make this easier to work with.

The idea is that whether than a large number of platform specific checks in the toml file, these should be cfg features that can be enabled/disabled on the various platforms as needed.  The xtask build system for the matrix-sdk-ffi bindings has been updated to use this to maintain existing behavior.

The new features are:
- sentry: whether sentry is included in setup, the rust sentry library does not work on wasm platforms.
- rustls-tls: whether rustls-tls should be used for network, this was previously implemented with target based conditionals.
- native-tls: whether native-tls should be used for network, this was previously implemented with target based conditionals.
- js: whether existing configs meant to run in a wasm environment targeting javascript are needed.

In the future, we will add the following additional features:
- react-native: used to ensure the uniffi exposed interfaces are the same across platforms.  This will be used more in future PRs, and is meant to address a feature requirement of react-native (where native ios+android and web bindings must share a public interface, a limitation of react-native + typescript).
- sqlite: Will be used to control if sqlite session store is included and exposed.
- indexeddb: Will be used to control if indexeddb session store is included and exposed.

- [x] Public API changes documented in changelogs (optional)

Signed-off-by: Daniel Salinas

PR #5245, opened 2025-06-18
feat(wasm): add feature support for indexedb and sqlite to matrix-sdk-ffi

zzorba:
Expose a choice of session storage in the matrix-sdk-ffi crate.  The initial choices are Sqlite (supported on non-Wasm platforms) and IndexedDb (Wasm only).  It was agreed though that feature flags make more sense here, and make the code more maintainable.

The existing bundled-sqlite is setup to opt-in sqlite, however several methods for configuring a Sqlite store have been moved onto a sub-builder for organizational reasons. 

This PR is stacked on top of https://github.com/matrix-org/matrix-rust-sdk/pull/5211/files, I will rebase it against main when that has landed.

Corresponding PR for complement-crypto: https://github.com/matrix-org/complement-crypto/pull/195

Signed-off-by: Daniel Salinas

### DIFF
--- a/bindings/matrix-sdk-ffi/src/session_store.rs
+++ b/bindings/matrix-sdk-ffi/src/session_store.rs
@@ -241,10 +241,10 @@ impl SessionStoreBuilder {
     pub(crate) fn build(&self) -> Result<SessionStoreResult, ClientBuildError> {
         match self {
             #[cfg(feature = "sqlite")]
-            SessionStoreBuilder::Sqlite(config) => config.build(),
+            Self::Sqlite(config) => config.build(),
 
             #[cfg(feature = "indexeddb")]
-            SessionStoreBuilder::IndexedDb(config) => config.build(),
+            Self::IndexedDb(config) => config.build(),
         }
     }
 }


Hywan:
Waiting on https://github.com/matrix-org/matrix-rust-sdk/pull/5211 to be merged before reviewing this one.

zzorba:
Would be nice to get opinions on the new session_store and client_builder changes. Those are wholly new, and structural feedback would be very welcome

poljar:
Can this one be rebased as well?

zzorba:
Yep, rebased it to adopt the Cargo.toml changes.  This PR now concerns only the configuration of the session store.

Let me know if a different approach would be preferred here, adding sub-builders seemed more self-documenting than having a large number of 'optional' methods on the main builder, but happy to do it a different way if you prefer.

zzorba:
Here is a corresponding complement-crypto PR: https://github.com/matrix-org/complement-crypto/pull/195

bindings/matrix-sdk-ffi/CHANGELOG.md
```diff
@@ -6,6 +6,29 @@ All notable changes to this project will be documented in this file.
...
+```
+builder
+   .session_store_sqlite(
+     SqliteSessionStoreBuilder.new({ dataPath: "data_path", cachePath: "cache_path" })
+       .passphrase("foobar")
+   )
+```
+
+The following methods from `ClientBuilder` have been moved onto `SqliteSessionStoreBuilder`:
+`session_paths`, `session_passphrase`, `session_pool_max_size`, `session_cache_size`, and  `session_journal_size_limit`.
```
Hywan: Please fix the markup here. If you are unsure about the Markdown rendering, you can use the Github interface to view the file:

Screenshot 2025-06-23 at 10 35 14

```suggestion
### Features

- Configuration of the session store has been changed to allow for use of either IndexedDb
or SQLite on relevant platforms.  These can be enabled via features, `indexeddb` or `sqlite`.

  Previously the code to configure a sqlite session would look something like this in a host language:

  \```rust
  builder
    .sessionPaths("data_path", "cache_path")
    .passphrase("foobar")
  \```

  With the new system, a helper object is exposed for either SQLite or IndexedDB to group those settings.

// etc
```

bindings/matrix-sdk-ffi/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "matrix-sdk-ffi"
-version = "0.12.0"
+version = "0.11.0"
```
Hywan: Nop.

bindings/matrix-sdk-ffi/Cargo.toml
```diff
@@ -25,17 +25,24 @@ crate-type = [
 
 [features]
 default = ["bundled-sqlite", "unstable-msc4274"]
-bundled-sqlite = ["matrix-sdk/bundled-sqlite"]
+bundled-sqlite = ["sqlite", "matrix-sdk/bundled-sqlite"]
 unstable-msc4274 = ["matrix-sdk-ui/unstable-msc4274"]
+# Use indexeddb for session storage, supported only on Wasm platforms.
```
Hywan:
```suggestion
# Use IndexedDB for session storage, only supported on Wasm targets.
```

bindings/matrix-sdk-ffi/Cargo.toml
```diff
@@ -25,17 +25,24 @@ crate-type = [
...
+indexeddb = ["matrix-sdk/indexeddb"]
+# Use sqlite for session storage, not supported on Wasm platforms.
```
Hywan:
```suggestion
# Use SQLite for session storage, not supported on Wasm targets.
```

bindings/matrix-sdk-ffi/Cargo.toml
```diff
@@ -25,17 +25,24 @@ crate-type = [
...
+sqlite = ["matrix-sdk/sqlite"]
+# provide a unified API regardless of target platforms, for use with multi-target platforms like react-native
```
Hywan:
```suggestion
# This feature indicates the crate must be compiled for the React-native host.
```

or to mimic the `js` feature:

```suggestion
# Required when targeting the React-native environment.
```

A link to React-native would be a nice bonus, but it's not required.

bindings/matrix-sdk-ffi/Cargo.toml
```diff
@@ -25,17 +25,24 @@ crate-type = [
...
+react-native = []
 # Required when targeting a Javascript environment, like Wasm in a browser.
-js = ["matrix-sdk-ui/js"]
+js = ["matrix-sdk-ui/js", "uuid/js"]
 # Use the TLS implementation provided by the host system, necessary on iOS and Wasm platforms.
-native-tls = ["matrix-sdk/native-tls", "sentry?/native-tls"]
+native-tls = ["matrix-sdk/native-tls", "sentry?/native-tls", "graphql_client/reqwest"]
```
Hywan: Why?

bindings/matrix-sdk-ffi/README.md
```diff
@@ -12,6 +12,8 @@ Given the number of platforms targeted, we have broken out a number of features
 ### Functionality
 - `sentry`: Enable error monitoring using Sentry, not supports on Wasm platforms.
 - `bundled-sqlite`: Use an embedded version of sqlite instead of the system provided one.
+- `sqlite`: Use sqlite for session storage, not available on Wasm platforms.
+- `indexeddb`: Use IndexedDb for session storage, only available on Wasm platforms.
```
Hywan: Same suggestions as in the `Cargo.toml` file apply.

bindings/matrix-sdk-ffi/build.rs
```diff
@@ -56,7 +56,31 @@ fn get_clang_major_version(clang_path: &Path) -> String {
...
+    }
+}
+
 fn main() -> Result<(), Box<dyn Error>> {
+    let sqlite_set = env_is_set("CARGO_FEATURE_SQLITE");
+    let indexeddb_set = env_is_set("CARGO_FEATURE_INDEXEDDB");
+    ensure(
+        sqlite_set || indexeddb_set,
+        "one of the features 'sqlite' or 'indexeddb' must be enabled",
+    );
```
Hywan: You must also ensure that they are not both activated at the same time, it's a [XOR](https://en.wikipedia.org/wiki/Exclusive_or):

```suggestion
    ensure(
        sqlite_set ^ indexeddb_set,
        "one of the features 'sqlite' or (exclusive) 'indexeddb' must be enabled",
    );
```

bindings/matrix-sdk-ffi/src/client_builder.rs
```diff
@@ -425,50 +350,23 @@ impl ClientBuilder {
...
-                sqlite_store_config = sqlite_store_config.pool_max_size(size);
-            }
-
-            if let Some(size) = builder.session_cache_size {
-                sqlite_store_config = sqlite_store_config.cache_size(size);
-            }
-
-            if let Some(limit) = builder.session_journal_size_limit {
-                sqlite_store_config = sqlite_store_config.journal_size_limit(limit);
+        let mut store_path = None;
```
Hywan: I prefer the old flow with `let store_path = if let Some(…)`. Why changing this? It was avoiding the need for a mutable variable, which is always preferable.

bindings/matrix-sdk-ffi/src/session_store.rs
```diff
@@ -0,0 +1,222 @@
...
+    pool_max_size: Option<usize>,
+    cache_size: Option<u32>,
+    journal_size_limit: Option<u32>,
+    system_is_memory_constrained: bool,
+}
+
+#[cfg(feature = "sqlite")]
+#[matrix_sdk_ffi_macros::export]
+impl SqliteSessionStoreBuilder {
+    /// Construct a SqliteSessionStoreBuilder and set the paths that the client
```
Hywan:
```suggestion
    /// Construct a `SqliteSessionStoreBuilder` and set the paths that the client
```

bindings/matrix-sdk-ffi/src/session_store.rs
```diff
@@ -0,0 +1,222 @@
...
+            passphrase: Zeroizing::new(None),
+            pool_max_size: None,
+            cache_size: None,
+            journal_size_limit: None,
+            system_is_memory_constrained: false,
+        }
+    }
+
+    /// Set the passphrase for the stores given to
+    /// [`ClientBuilder::paths`].
```
Hywan: The `ClientBuilder::paths` method doesn't exist. The previous comment was:

```rust
    /// Set the passphrase for the stores given to
    /// [`ClientBuilder::session_paths`].
```

Alright, you've renamed the method, but the method is no longer on `ClientBuilder`, it's on `SqliteSessionStoreBuilder`.

```suggestion
    /// [`SqliteSessionStoreBuidler::paths`].
```

bindings/matrix-sdk-ffi/src/session_store.rs
```diff
@@ -0,0 +1,222 @@
...
+    pub fn passphrase(self: Arc<Self>, passphrase: Option<String>) -> Arc<Self> {
+        let mut builder = unwrap_or_clone_arc(self);
+        builder.passphrase = Zeroizing::new(passphrase);
+        Arc::new(builder)
+    }
+
+    /// Set the pool max size for the SQLite stores given to
+    /// [`ClientBuilder::session_paths`].
```
Hywan: Same here. You've also forgot to rename the method.

```suggestion
    /// [`SqliteSessionStoreBuilder::paths`].
```

bindings/matrix-sdk-ffi/src/session_store.rs
```diff
@@ -0,0 +1,222 @@
...
+    /// See [`SqliteStoreConfig::pool_max_size`] to learn more.
+    pub fn pool_max_size(self: Arc<Self>, pool_max_size: Option<u32>) -> Arc<Self> {
+        let mut builder = unwrap_or_clone_arc(self);
+        builder.pool_max_size = pool_max_size
+            .map(|size| size.try_into().expect("`pool_max_size` is too large to fit in `usize`"));
+        Arc::new(builder)
+    }
+
+    /// Set the cache size for the SQLite stores given to
+    /// [`ClientBuilder::session_paths`].
```
Hywan: Same here.

bindings/matrix-sdk-ffi/src/session_store.rs
```diff
@@ -0,0 +1,222 @@
...
+    ///
+    /// See [`SqliteStoreConfig::cache_size`] to learn more.
+    pub fn cache_size(self: Arc<Self>, cache_size: Option<u32>) -> Arc<Self> {
+        let mut builder = unwrap_or_clone_arc(self);
+        builder.cache_size = cache_size;
+        Arc::new(builder)
+    }
+
+    /// Set the size limit for the SQLite WAL files of stores given to
+    /// [`ClientBuilder::session_paths`].
```
Hywan: Same here.

bindings/matrix-sdk-ffi/src/session_store.rs
```diff
@@ -0,0 +1,222 @@
...
+        builder.journal_size_limit = limit;
+        Arc::new(builder)
+    }
+
+    /// Tell the client that the system is memory constrained, like in a push
+    /// notification process for example.
+    ///
+    /// So far, at the time of writing (2025-04-07), it changes the defaults of
+    /// [`SqliteStoreConfig`], so one might not need to call
+    /// [`ClientBuilder::session_cache_size`] and siblings for example. Please
```
Hywan: Same here.

bindings/matrix-sdk-ffi/src/session_store.rs
```diff
@@ -0,0 +1,222 @@
...
+
+#[cfg(feature = "indexeddb")]
+#[matrix_sdk_ffi_macros::export]
+impl IndexedDbSessionStoreBuilder {
+    #[uniffi::constructor]
+    pub fn new(name: String) -> Arc<Self> {
+        Arc::new(Self { name, passphrase: None })
+    }
+
+    /// Set the passphrase for the IndexedDB store.
```
Hywan: You may want to explain that a passphrase is optional.

bindings/matrix-sdk-ffi/src/session_store.rs
```diff
@@ -0,0 +1,222 @@
...
+        Arc::new(builder)
+    }
+}
+
+/// Internal struct for tracking the choice of session store config.
+#[derive(Clone)]
+pub enum SessionStoreConfig {
+    #[cfg(feature = "sqlite")]
+    /// Setup the client to use the SQLite store.
+    Sqlite(SqliteSessionStoreBuilder),
```
Hywan: Documentation before `#[cfg(…)]`.

```suggestion
    /// Setup the client to use the SQLite store.
    #[cfg(feature = "sqlite")]
    Sqlite(SqliteSessionStoreBuilder),
```

bindings/matrix-sdk-ffi/src/session_store.rs
```diff
@@ -0,0 +1,222 @@
...
+
+    /// Set the passphrase for the IndexedDB store.
+    pub fn passphrase(self: Arc<Self>, passphrase: Option<String>) -> Arc<Self> {
+        let mut builder = unwrap_or_clone_arc(self);
+        builder.passphrase = passphrase;
+        Arc::new(builder)
+    }
+}
+
+/// Internal struct for tracking the choice of session store config.
```
Hywan:
```suggestion
/// Represent the kind of store the client will configure.
```

bindings/matrix-sdk-ffi/src/session_store.rs
```diff
@@ -0,0 +1,222 @@
...
+#[derive(Clone)]
+pub enum SessionStoreConfig {
+    #[cfg(feature = "sqlite")]
+    /// Setup the client to use the SQLite store.
```
Hywan: The comment is wrong.

```suggestion
    /// Represents the builder for the SQLite store.
```

bindings/matrix-sdk-ffi/src/session_store.rs
```diff
@@ -0,0 +1,222 @@
...
+    pub fn passphrase(self: Arc<Self>, passphrase: Option<String>) -> Arc<Self> {
+        let mut builder = unwrap_or_clone_arc(self);
+        builder.passphrase = passphrase;
+        Arc::new(builder)
+    }
+}
+
+/// Internal struct for tracking the choice of session store config.
+#[derive(Clone)]
+pub enum SessionStoreConfig {
```
Hywan: I think we can rename this type to:

```suggestion
pub enum SessionStoreBuilder {
```

because it only holds builders for various stores.

bindings/matrix-sdk-ffi/src/session_store.rs
```diff
@@ -0,0 +1,222 @@
...
+    #[cfg(feature = "sqlite")]
+    /// Setup the client to use the SQLite store.
+    Sqlite(SqliteSessionStoreBuilder),
+
+    #[cfg(feature = "indexeddb")]
+    /// Setup the client to use the IndexedDB store.
+    IndexedDb(IndexedDbSessionStoreBuilder),
```
Hywan: Same.

bindings/matrix-sdk-ffi/src/session_store.rs
```diff
@@ -0,0 +1,222 @@
...
+}
+
+/// The result of building a [`SessionStoreConfig`], with data that
+/// can be passed directly to a ClientBuilder.
```
Hywan:
```suggestion
/// can be passed directly to a [`ClientBuilder`].
```

bindings/matrix-sdk-ffi/src/session_store.rs
```diff
@@ -0,0 +1,222 @@
...
+pub enum SessionStoreResult {
```
Hywan: In Rust, the term _result_ has a particular meaning, it's attached to [the `std::result::Result` type](https://doc.rust-lang.org/std/result/enum.Result.html). There is no `Ok` or `Err` here.
If you want to express the _output_ of a computation, you can use the term _outcome_.

However, in this case, it's just a `SessionStore` that is built!

```suggestion
pub enum SessionStore {
```

bindings/matrix-sdk-ffi/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "matrix-sdk-ffi"
-version = "0.11.0"
+version = "0.12.0"
```
Hywan: Should be a fixup commit of your first commit.

bindings/matrix-sdk-ffi/src/session_store.rs
```diff
@@ -1,152 +1,227 @@
...
+    use tracing::debug;
+    use zeroize::Zeroizing;
+
+    use crate::helpers::unwrap_or_clone_arc;
+    use super::SessionStoreResult;
+    use crate::client_builder::ClientBuildError;
+
+    /// The store paths the client will use when built.
+    #[derive(Clone)]
+    struct SessionPaths {
```
Hywan, line 29: Don't you need a `pub(crate)` here?

xtask/src/ci.rs
```diff
@@ -231,7 +231,7 @@ fn check_clippy() -> Result<()> {
         "rustup run {NIGHTLY} cargo clippy --workspace --all-targets
             --exclude matrix-sdk-crypto --exclude xtask
             --no-default-features
-            --features native-tls,sso-login,testing
+            --features native-tls,sso-login,sqlite,testing
```
Hywan, line 234: The same patch is mixing two things. Proof is your commit message contains “and”.

Please split this patch into two.

Hywan: By the way, this change should be a fixup of the patch introducing the `sqlite` feature.

bindings/matrix-sdk-ffi/src/session_store.rs
```diff
@@ -55,15 +55,15 @@ mod sqlite_session_store {
         /// capable of handling multiple users, however it is valid to use the
         /// same path for both stores on a single session.
         #[uniffi::constructor]
-        pub fn new(data_path: String, cache_path: String) -> Self {
-            Self {
+        pub fn new(data_path: String, cache_path: String) -> Arc<Self> {
```
Hywan, line 58: Why?

It should be a fixup patch.

bindings/matrix-sdk-ffi/src/session_store.rs
```diff
@@ -182,6 +182,11 @@ mod sqlite_session_store {
 
 #[cfg(feature = "indexeddb")]
 mod indexeddb_session_store {
+    use std::sync::Arc;
+
+    use super::SessionStoreResult;
+    use crate::{client_builder::ClientBuildError, helpers::unwrap_or_clone_arc};
+
```
Hywan, line 189: A fixup patch.

Hywan (changes requested):
Thanks for the patch. The general idea is good, but the patches deserve more work.

xtask/src/ci.rs
zzorba: It is not mixing two different things, it is adding support for indexedb which requires parameterizing the sqlite configuration.  I do not believe these are separable changes.

bindings/matrix-sdk-ffi/src/session_store.rs
zzorba: I don't know what that means.  What is a fixup patch?

bindings/matrix-sdk-ffi/build.rs
zzorba: I don't think they are necessarily exclusive.  

At the moment they are because of platform limitations, but the code to enable IndexeDB + SQLite at the same time would work perfectly fine someday when there is a bundled sqlite solution on web (not an impossibility).

bindings/matrix-sdk-ffi/src/session_store.rs
zzorba: No, because it is only used in this local `sqlite_session_store` package, it is not exposed anywhere else, the code is encapsulated. Purely an implementation detail.

bindings/matrix-sdk-ffi/README.md
zzorba: I don't know what this means

Hywan: These sentences are copied-pasted from the `Cargo.toml` comment. I made suggestions on `Cargo.toml`. So please re-apply them here too :-).

bindings/matrix-sdk-ffi/build.rs
Hywan: I doubt the `Client` can have 2 stores, so, yeah, these features are mutually exclusive 🙂.
Edit: except if you use for example IndexedDB for one store, like the state store, and SQLite for the crypto store, but it seems pretty weird. Please, make it mutually exclusive for the moment, as having them two at the same time isn't supported right now, nor tested.

zzorba: At the compilation layer, you can definitely have two stores enabled.

zzorba: I have updated this as you requested, although I do still disagree with your reasoning.  

The builder system written here is explicitly designed to choose between the two, nothing is mutually exclusive about the feature as written.  The fact that indexedb and sqlite have platform restrictions is an orthogonal issue entirely, and doesn't seem worth guarding against at the feature level.

bindings/matrix-sdk-ffi/src/session_store.rs
Hywan: A fixup patch is a patch that is “melt”/“merged” with another one so that we end up with a single patch, see https://git-scm.com/docs/git-rebase.

manuroe: The contributing guide also explains it: https://github.com/matrix-org/matrix-rust-sdk/blob/main/CONTRIBUTING.md#addressing-review-comments-using-fixup-commits.

bindings/matrix-sdk-ffi/build.rs
Hywan: At the compilation layer, you can have IDB and SQLite if SQLite can be compiled to Wasm. Right now:

```sh
$ cargo build --target wasm32-unknown-unknown --package matrix-sdk-sqlite
```

does not compile.

I prefer to be safe and to prevent someone doing the wrong thing.

If one day a user is asking for SQLite+Wasm support, I'll be happy to revisit this. For the moment, I prefer to provide a satisfying developer experience.

bindings/matrix-sdk-ffi/CHANGELOG.md
```diff
@@ -7,26 +7,21 @@ All notable changes to this project will be documented in this file.
...
+
+  \```rust
+  builder
+    .sessionPaths("data_path", "cache_path")
+    .passphrase("foobar")
+  \```
+
+  With the new system, a helper object is exposed for either SQLite or IndexedDB to group those settings.
+
+// etc
```
Hywan: You did apply my suggestion straightforwardly, but I didn't suggest all the fixes needed to be done. Please revisit this.

bindings/matrix-sdk-ffi/CHANGELOG.md
```diff
@@ -7,26 +7,21 @@ All notable changes to this project will be documented in this file.
...
-     SqliteSessionStoreBuilder.new({ dataPath: "data_path", cachePath: "cache_path" })
-       .passphrase("foobar")
-   )
-```
-
-The following methods from `ClientBuilder` have been moved onto `SqliteSessionStoreBuilder`:
-`session_paths`, `session_passphrase`, `session_pool_max_size`, `session_cache_size`, and  `session_journal_size_limit`. 
+  Previously the code to configure a sqlite session would look something like this in a host language:
+
+  \```rust
```
Hywan: Please remove the `\`.

bindings/matrix-sdk-ffi/CHANGELOG.md
Hywan: I see your fixup patch in https://github.com/matrix-org/matrix-rust-sdk/pull/5245/commits/7987cc2fceaab0d7256cc6c5576fcfcefb010643.

bindings/matrix-sdk-ffi/src/session_store.rs
```diff
@@ -227,24 +227,24 @@ use crate::client_builder::ClientBuildError;
...
     IndexedDb(IndexedDbSessionStoreBuilder),
 }
 
-impl SessionStoreConfig {
+impl SessionStoreBuilder {
     pub(crate) fn build(&self) -> Result<SessionStoreResult, ClientBuildError> {
         match self {
             #[cfg(feature = "sqlite")]
-            SessionStoreConfig::Sqlite(config) => config.build(),
+            SessionStoreBuilder::Sqlite(config) => config.build(),
```
Hywan: Here, you can even use `Self::Sqlite`, it's probably easier.

bindings/matrix-sdk-ffi/src/session_store.rs
```diff
@@ -227,24 +227,24 @@ use crate::client_builder::ClientBuildError;
...
 
             #[cfg(feature = "indexeddb")]
-            SessionStoreConfig::IndexedDb(config) => config.build(),
+            SessionStoreBuilder::IndexedDb(config) => config.build(),
```
Hywan: Same with `Self::IndexedDb`.

Hywan:
I will address the feedback myself and will merge the PR.

Hywan:
Closing in favor of https://github.com/matrix-org/matrix-rust-sdk/pull/5811.

PR #5246, opened 2025-06-18
refactor(crypto): Extract a method for handling encrypted to-device events

andybalaam:
We ended up calling `to_device_event_is_from_dehydrated_device` but if we shuffle things around a bit we only need to call it once.

This _should_ have no effect on behaviour in any way.  I have used my thinking brain to try and prove to myself that the new code is equivalent to the old, but I'd really appreciate someone else double-checking.


crates/matrix-sdk-crypto/src/machine/mod.rs
```diff
@@ -843,10 +843,15 @@ impl OlmMachine {
...
     }
 
-    /// Decrypt a to-device event.
+    /// Decrypt and handle a to-device event.
     ///
-    /// Returns a decrypted `ToDeviceEvent` if the decryption was successful,
-    /// an error indicating why decryption failed otherwise.
+    /// If decryption (or checking the sender device) fails, returns an error.
+    ///
+    /// If the sender device is dehdrated, does no handling and immediately
```
Hywan:
```suggestion
    /// If the sender device is dehydrated, does no handling and immediately
```

crates/matrix-sdk-crypto/src/machine/mod.rs
```diff
@@ -856,20 +861,40 @@ impl OlmMachine {
...
+        let from_dehydrated_device = self
+            .to_device_event_is_from_dehydrated_device(&decrypted, &event.sender)
+            .await
+            .map_err(|e| {
+                error!(
+                    error = ?e,
+                    "Couldn't check whether the event is from a dehydrated device",
+                );
+                e
+            })?;
```
Hywan: Why logging the error if the error is returned? Is the error not accurate enough?

crates/matrix-sdk-crypto/src/machine/mod.rs
```diff
@@ -1409,7 +1434,7 @@ impl OlmMachine {
         e: ToDeviceEvent<ToDeviceEncryptedEventContent>,
     ) -> Option<ProcessedToDeviceEvent> {
         let decrypted = match self.decrypt_to_device_event(transaction, &e, changes).await {
-            Ok(e) => e,
+            Ok(decrypted) => decrypted,
```
Hywan: To remove the `?` after the `match`, you could do this instead:

```suggestion
            Ok(maybe_decrypted) => maybe_decrypted?,
```

I think it's a bit clearer. I know it's a personal taste.

crates/matrix-sdk-crypto/src/machine/mod.rs
```diff
@@ -1424,27 +1449,7 @@ impl OlmMachine {
...
-            }
-            Ok(false) => {}
-            Err(err) => {
-                error!(
-                    error = ?err,
-                    "Couldn't check whether the event is from a dehydrated device",
-                );
-            }
-        }
+        }?;
```
Hywan: If you decide to apply my suggestion above:

```suggestion
        };
```

Hywan (approved):
Looks good to me, thanks! I left tiny suggestions, nothing mandatory.

crates/matrix-sdk-crypto/src/machine/mod.rs
andybalaam: Hmm, good point. I was just making sure I recreate the behaviour of the original code, which logs inside decrypt_to_device_event but I don't see a good reason to do this, so I took it out in 9ab164221a

crates/matrix-sdk-crypto/src/machine/mod.rs
andybalaam: Yeah, I much prefer that too. Thanks! eb287c1750

PR #5247, opened 2025-06-18
feat(sdk): Introduce `GenericRoomEventCacheUpdate`, one channel to get update of all rooms

Hywan:
`RoomEventCache::subscribe` is nice to subscribe to every update
happening inside a room in the event cache. However, the returned
`RoomEventCacheSubscriber` has side-effects when dropped (see
auto-shrink to save memory space). In some situation, this is pretty
annoying. for example, if one wants to listen to multiple room updates,
all the room event cache subscribers must be kept in memory, thus
breaking the side-effects. This isn't always the desired output. In
addition, listening to multiple channels/subscribers at the same is
quite complex, as it implies non-trivial async runtime efforts or
complex future types.

To solve this problem, this patch introduces a new
`EventCache::subscribe_to_generic_room_updates` method, which returns a
single `Receiver<GenericRoomEventCacheUpdate>`.

First off, it hides the details of `RoomEventCacheUpdate` (returned by
`RoomEventCacheSubscriber`), which might be desired, but particularly
lighter because events aren't part of the payload.

Second, one no longer needs to subscribe to all rooms. Only one channel
can be listened to get updates for all rooms. It reduces the complexity
on the caller side, plus `Receiver<GenericRoomEventCacheUpdate>` doesn't
have any side-effect.

This patch tests this feature in 4 situations:

1. when a room is created/loaded empty,
2. when a room is loaded and is not empty because data exists in the storage,
3. when a room receives data from the sync,
4. when a room receives data from the pagination.

### DIFF
--- a/crates/matrix-sdk/src/event_cache/mod.rs
+++ b/crates/matrix-sdk/src/event_cache/mod.rs
@@ -193,13 +193,13 @@ impl EventCache {
             // Force-initialize the sender in the [`RoomEventCacheInner`].
             self.inner.auto_shrink_sender.get_or_init(|| tx);
 
-            let auto_shrink_linked_chunk_tasks =
+            let auto_shrink_linked_chunk_task =
                 spawn(Self::auto_shrink_linked_chunk_task(self.inner.clone(), rx));
 
             Arc::new(EventCacheDropHandles {
                 listen_updates_task,
                 ignore_user_list_update_task,
-                auto_shrink_linked_chunk_task: auto_shrink_linked_chunk_tasks,
+                auto_shrink_linked_chunk_task,
             })
         });
 
--- a/crates/matrix-sdk/src/event_cache/mod.rs
+++ b/crates/matrix-sdk/src/event_cache/mod.rs
@@ -188,13 +188,15 @@ impl EventCache {
                 client.subscribe_to_ignore_user_list_changes(),
             ));
 
-            let (tx, rx) = mpsc::channel(32);
+            let (auto_shrink_sender, auto_shrink_receiver) = mpsc::channel(32);
 
             // Force-initialize the sender in the [`RoomEventCacheInner`].
-            self.inner.auto_shrink_sender.get_or_init(|| tx);
+            self.inner.auto_shrink_sender.get_or_init(|| auto_shrink_sender);
 
-            let auto_shrink_linked_chunk_task =
-                spawn(Self::auto_shrink_linked_chunk_task(self.inner.clone(), rx));
+            let auto_shrink_linked_chunk_task = spawn(Self::auto_shrink_linked_chunk_task(
+                self.inner.clone(),
+                auto_shrink_receiver,
+            ));
 
             Arc::new(EventCacheDropHandles {
                 listen_updates_task,
--- a/crates/matrix-sdk/src/event_cache/mod.rs
+++ b/crates/matrix-sdk/src/event_cache/mod.rs
@@ -63,7 +63,7 @@ mod room;
 
 pub mod paginator;
 pub use pagination::{PaginationToken, RoomPagination, RoomPaginationStatus};
-pub use room::{RoomEventCache, RoomEventCacheListener};
+pub use room::{RoomEventCache, RoomEventCacheSubscriber};
 
 /// An error observed in the [`EventCache`].
 #[derive(thiserror::Error, Debug)]
@@ -276,14 +276,14 @@ impl EventCache {
     /// The auto-shrink mechanism works this way:
     ///
     /// - Each time there's a new subscriber to a [`RoomEventCache`], it will
-    ///   increment the active number of listeners to that room, aka
-    ///   [`RoomEventCacheState::listener_count`].
+    ///   increment the active number of subscribers to that room, aka
+    ///   [`RoomEventCacheState::subscriber_count`].
     /// - When that subscriber is dropped, it will decrement that count; and
     ///   notify the task below if it reached 0.
     /// - The task spawned here, owned by the [`EventCacheInner`], will listen
     ///   to such notifications that a room may be shrunk. It will attempt an
     ///   auto-shrink, by letting the inner state decide whether this is a good
-    ///   time to do so (new listeners might have spawned in the meanwhile).
+    ///   time to do so (new subscribers might have spawned in the meanwhile).
     #[instrument(skip_all)]
     async fn auto_shrink_linked_chunk_task(
         inner: Arc<EventCacheInner>,
@@ -303,11 +303,11 @@ impl EventCache {
             trace!("waiting for state lock…");
             let mut state = room.inner.state.write().await;
 
-            match state.auto_shrink_if_no_listeners().await {
+            match state.auto_shrink_if_no_subscribers().await {
                 Ok(diffs) => {
                     if let Some(diffs) = diffs {
                         // Hey, fun stuff: we shrunk the linked chunk, so there shouldn't be any
-                        // listeners, right? RIGHT? Especially because the state is guarded behind
+                        // subscribers, right? RIGHT? Especially because the state is guarded behind
                         // a lock.
                         //
                         // However, better safe than sorry, and it's cheap to send an update here,
--- a/crates/matrix-sdk/src/event_cache/mod.rs
+++ b/crates/matrix-sdk/src/event_cache/mod.rs
@@ -49,7 +49,7 @@ use ruma::{
     events::AnySyncEphemeralRoomEvent, serde::Raw, OwnedEventId, OwnedRoomId, RoomId, RoomVersionId,
 };
 use tokio::sync::{
-    broadcast::{error::RecvError, Receiver},
+    broadcast::{channel, error::RecvError, Receiver, Sender},
     mpsc, Mutex, RwLock,
 };
 use tracing::{debug, error, info, info_span, instrument, trace, warn, Instrument as _, Span};
@@ -155,6 +155,8 @@ impl Debug for EventCache {
 impl EventCache {
     /// Create a new [`EventCache`] for the given client.
     pub(crate) fn new(client: WeakClient, event_cache_store: EventCacheStoreLock) -> Self {
+        let (generic_room_event_cache_update_sender, _) = channel(32);
+
         Self {
             inner: Arc::new(EventCacheInner {
                 client,
@@ -163,6 +165,7 @@ impl EventCache {
                 by_room: Default::default(),
                 drop_handles: Default::default(),
                 auto_shrink_sender: Default::default(),
+                generic_room_event_cache_update_sender,
             }),
         }
     }
@@ -353,6 +356,21 @@ impl EventCache {
     pub async fn clear_all_rooms(&self) -> Result<()> {
         self.inner.clear_all_rooms().await
     }
+
+    /// Subscribe to _generic_ room updates.
+    ///
+    /// If one wants to listen what has changed in a specific room, the
+    /// [`RoomEventCache::subscribe`] is recommended. However, the
+    /// [`RoomEventCacheSubscriber`] type triggers side-effects.
+    ///
+    /// If one wants to get a high-overview, generic, updates for rooms, and
+    /// without side-effects, this method is recommended. For example, it
+    /// doesn't provide a list of new events, but rather a
+    /// [`GenericRoomEventCacheUpdate::TimelineUpdated`] message. Also, dropping
+    /// the receiver of this channel will not trigger any side-effect.
+    pub fn subscribe_to_generic_room_updates(&self) -> Receiver<GenericRoomEventCacheUpdate> {
+        self.inner.generic_room_event_cache_update_sender.subscribe()
+    }
 }
 
 struct EventCacheInner {
@@ -384,6 +402,12 @@ struct EventCacheInner {
     ///
     /// See doc comment of [`EventCache::auto_shrink_linked_chunk_task`].
     auto_shrink_sender: OnceLock<mpsc::Sender<AutoShrinkChannelPayload>>,
+
+    /// A sender for generic room update.
+    ///
+    /// See doc comment of [`GenericRoomEventCacheUpdate`] and
+    /// [`EventCache::subscribe_to_generic_room_updates`].
+    generic_room_event_cache_update_sender: Sender<GenericRoomEventCacheUpdate>,
 }
 
 type AutoShrinkChannelPayload = OwnedRoomId;
@@ -473,7 +497,14 @@ impl EventCacheInner {
         for (room_id, left_room_update) in updates.left {
             let room = self.for_room(&room_id).await?;
 
-            if let Err(err) = room.inner.handle_left_room_update(left_room_update).await {
+            if let Err(err) = room
+                .inner
+                .handle_left_room_update(
+                    left_room_update,
+                    &self.generic_room_event_cache_update_sender,
+                )
+                .await
+            {
                 // Non-fatal error, try to continue to the next room.
                 error!("handling left room update: {err}");
             }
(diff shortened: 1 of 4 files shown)


crates/matrix-sdk/src/event_cache/mod.rs
```diff
@@ -188,13 +188,15 @@ impl EventCache {
                 client.subscribe_to_ignore_user_list_changes(),
             ));
 
-            let (tx, rx) = mpsc::channel(32);
+            let (auto_shrink_sender, auto_shrink_receiver) = mpsc::channel(32);
```
bnjbvr: I find that brevity has its merits, and in this case, the long names don't really buy us anything more; the useful scope area of these two variables is 4 lines…

crates/matrix-sdk/src/event_cache/room/mod.rs
```diff
@@ -66,15 +66,15 @@ impl fmt::Debug for RoomEventCache {
...
 ///
 /// The current side-effect is: auto-shrinking the [`RoomEventCache`] when no
-/// more listeners are active. This is an optimisation to reduce the number of
-/// data held in memory by a [`RoomEventCache`]: when no more listeners are
+/// more subscribers are active. This is an optimisation to reduce the number of
+/// data held in memory by a [`RoomEventCache`]: when no more subscribers are
 /// active, all data are reduced to the minimum.
 #[allow(missing_debug_implementations)]
-pub struct RoomEventCacheListener {
+pub struct RoomEventCacheSubscriber {
```
bnjbvr, line 79: I would rather do the opposite and keep `RoomEventCacheListener` everywhere, so as to catch the attention of a reader that it's not a plain `Subscriber`, and has an extra cost. To me, "listener" seems way more familiar, but YMMV there. Also, it's not using a `Subscriber` but a `Receiver` under the hood, so the renaming doesn't make sense.

crates/matrix-sdk/src/event_cache/mod.rs
```diff
@@ -578,6 +629,21 @@ pub struct BackPaginationOutcome {
     pub events: Vec<TimelineEvent>,
 }
 
+/// Represents an update of a room. It hides the details of
+/// [`RoomEventCacheUpdate`] by being more generic.
+///
+/// This is used by [`EventCache::subscribe_to_generic_room_updates`]. Please
+/// read it to learn more about the motivation behind this type.
+#[derive(Clone, Debug)]
+pub enum GenericRoomEventCacheUpdate {
```
bnjbvr: nit: I think the `Generic` is misplaced in this name. It makes it sound like the `room` is generic, while it's rather the `update`; so how about `RoomEventCacheGenericUpdate`, instead? (would also mean renaming all the associated fields/methods/getters etc.)

It's still a mouthful though, so if anyone has ideas for a shorter and explicit name. What do you think of `AnyRoomUpdate`? It encodes in a shorter way the fact that it's an update to any room in the cache.

crates/matrix-sdk/src/event_cache/room/mod.rs
```diff
@@ -359,37 +370,57 @@ impl RoomEventCacheInner {
...
-        self.handle_timeline(
-            updates.timeline,
-            updates.ephemeral.clone(),
-            updates.ambiguity_changes,
-        )
-        .await?;
+    pub(super) async fn handle_joined_room_update(
+        &self,
+        updates: JoinedRoomUpdate,
+        generic_room_event_cache_update_sender: &Sender<GenericRoomEventCacheUpdate>,
```
bnjbvr: It's likely not useful to pass it as a parameter, since the `self` (`RoomEventCacheInner`) contains a clone of it.

crates/matrix-sdk/src/event_cache/room/mod.rs
```diff
@@ -359,37 +370,57 @@ impl RoomEventCacheInner {
...
+    ) -> Result<()> {
+        if self
+            .handle_timeline(updates.timeline, updates.ephemeral.clone(), updates.ambiguity_changes)
+            .await?
+        {
+            let _ = generic_room_event_cache_update_sender.send(
```
bnjbvr: `self.generic_update_sender.send(`

crates/matrix-sdk/src/event_cache/room/mod.rs
```diff
@@ -359,37 +370,57 @@ impl RoomEventCacheInner {
...
         Ok(())
     }
 
     #[instrument(skip_all, fields(room_id = %self.room_id))]
-    pub(super) async fn handle_left_room_update(&self, updates: LeftRoomUpdate) -> Result<()> {
-        self.handle_timeline(updates.timeline, Vec::new(), updates.ambiguity_changes).await?;
+    pub(super) async fn handle_left_room_update(
+        &self,
+        updates: LeftRoomUpdate,
+        generic_room_event_cache_update_sender: &Sender<GenericRoomEventCacheUpdate>,
```
bnjbvr: ditto

crates/matrix-sdk/src/event_cache/mod.rs
```diff
@@ -551,10 +593,21 @@ impl EventCacheInner {
...
                     room_id.to_owned(),
                     auto_shrink_sender,
+                    self.generic_room_event_cache_update_sender.clone(),
                 );
 
                 by_room_guard.insert(room_id.to_owned(), room_event_cache.clone());
 
+                // If at least one event has been loaded, it means there is a timeline. Let's
+                // emit a generic update.
+                if has_at_least_one_event {
```
andybalaam: So the creation of an empty timeline is not something we want to hear about?

Can timelines _become_ empty through redactions?

crates/matrix-sdk/src/event_cache/room/mod.rs
```diff
@@ -359,37 +370,57 @@ impl RoomEventCacheInner {
...
+        }
+
         Ok(())
     }
 
+    /// Handle a [`Timeline`], i.e. new events receivedy by a sync for this
+    /// room.
+    ///
+    /// It returns `Ok(true)` if events have been added, redacted or removed,
+    /// `Ok(false)` otherwise.
```
bnjbvr: Since `RoomEventCacheInner` has the generic sender in it, I think it could immediately send the updates itself, instead of propagating the information upwards that it has new changes?

crates/matrix-sdk/src/event_cache/room/mod.rs
```diff
@@ -66,10 +66,15 @@ impl fmt::Debug for RoomEventCache {
     }
 }
 
-/// Thin wrapper for a room event cache listener, so as to trigger side-effects
-/// when all listeners are gone.
+/// Thin wrapper for a room event cache subscriber, so as to trigger
+/// side-effects when all subscribers are gone.
+///
+/// The current side-effect is: auto-shrinking the [`RoomEventCache`] when no
```
andybalaam: This side effect happens on `Drop`, right? Might be worth saying that explicitly.

crates/matrix-sdk/src/event_cache/room/mod.rs
```diff
@@ -352,37 +370,57 @@ impl RoomEventCacheInner {
...
+    pub(super) async fn handle_joined_room_update(
+        &self,
+        updates: JoinedRoomUpdate,
+        generic_room_event_cache_update_sender: &Sender<GenericRoomEventCacheUpdate>,
+    ) -> Result<()> {
+        if self
+            .handle_timeline(updates.timeline, updates.ephemeral.clone(), updates.ambiguity_changes)
+            .await?
+        {
+            let _ = generic_room_event_cache_update_sender.send(
```
andybalaam: So we ignore errors from `send`? Why? Should we log them at least?

crates/matrix-sdk/src/event_cache/room/mod.rs
```diff
@@ -352,37 +370,57 @@ impl RoomEventCacheInner {
...
+        if self.handle_timeline(updates.timeline, Vec::new(), updates.ambiguity_changes).await? {
+            let _ = generic_room_event_cache_update_sender.send(
+                GenericRoomEventCacheUpdate::TimelineUpdated { room_id: self.room_id.clone() },
+            );
+        }
+
         Ok(())
     }
 
+    /// Handle a [`Timeline`], i.e. new events receivedy by a sync for this
```
andybalaam:
```suggestion
    /// Handle a [`Timeline`], i.e. new events received by a sync for this
```

crates/matrix-sdk/src/event_cache/room/mod.rs
```diff
@@ -352,37 +370,57 @@ impl RoomEventCacheInner {
...
+    /// room.
+    ///
+    /// It returns `Ok(true)` if events have been added, redacted or removed,
+    /// `Ok(false)` otherwise.
     async fn handle_timeline(
```
andybalaam, line 396: Is there a test for this function? If it can be added easily it seems worth it now we have the boolean return value that we need to get right.

andybalaam (approved):
Looks good, with a couple of questions. Thanks!

crates/matrix-sdk/src/event_cache/mod.rs
Hywan: I've extracted these patches into their own PR https://github.com/matrix-org/matrix-rust-sdk/pull/5269 because I need the `GenericRoomEventCacheUpdate` feature to be merged and I don't want to be blocked by renaming discussions.

crates/matrix-sdk/src/event_cache/room/mod.rs
Hywan: See https://github.com/matrix-org/matrix-rust-sdk/pull/5247#discussion_r2161626528.

crates/matrix-sdk/src/event_cache/mod.rs
Hywan: You're right, I think `RoomEventCacheGenericUpdate` is better.

crates/matrix-sdk/src/event_cache/room/mod.rs
Hywan: Indeed! 🤦

crates/matrix-sdk/src/event_cache/room/mod.rs
Hywan: Thanks!

crates/matrix-sdk/src/event_cache/mod.rs
Hywan: > So the creation of an empty timeline is not something we want to hear about?

We may, at some point, but so far only `RoomEventCacheGenericUpdate::TimelineUpdated` exists. If we are interested by another event, we can change the logic here.

> Can timelines become empty through redactions?

No, because it will contain at least 2 events: one for the original event, and one for the redaction event. In all cases, this is a _timeline update_.

crates/matrix-sdk/src/event_cache/room/mod.rs
Hywan: Resolved since this method no longer returns a boolean 😛.

crates/matrix-sdk/src/event_cache/room/mod.rs
Hywan: Indeed, I'm okay with that. Done.

crates/matrix-sdk/src/event_cache/room/mod.rs
Hywan: Indeed. Addressed!

crates/matrix-sdk/src/event_cache/room/mod.rs
Hywan: [Quoting the doc](https://docs.rs/tokio/latest/tokio/sync/broadcast/error/struct.SendError.html):

> A send operation can only fail if there are no active receivers, implying that the message could never be received. The error contains the message being sent as a payload so it can be recovered.

We don't care if there is no more receivers, it's _fine_ to ignore the error in this case.

crates/matrix-sdk/src/event_cache/mod.rs
```diff
@@ -353,6 +356,21 @@ impl EventCache {
...
+    /// Subscribe to room _generic_ updates.
+    ///
+    /// If one wants to listen what has changed in a specific room, the
+    /// [`RoomEventCache::subscribe`] is recommended. However, the
+    /// [`RoomEventCacheSubscriber`] type triggers side-effects.
+    ///
+    /// If one wants to get a high-overview, generic, updates for rooms, and
+    /// without side-effects, this method is recommended. For example, it
+    /// doesn't provide a list of new events, but rather a
+    /// [`RoomEventCacheGenericUpdate::TimelineUpdated`] message. Also, dropping
```
bnjbvr: light suggestion: let's not say "message", as the meaning is a bit conflated in the context of Matrix. Maybe let's say "notification" or "update"?

crates/matrix-sdk/src/event_cache/mod.rs
```diff
@@ -540,6 +564,8 @@ impl EventCacheInner {
                 )
                 .await?;
 
+                let has_at_least_one_event = room_state.events().revents().next().is_some();
+
                 // SAFETY: we must have subscribed before reaching this coed, otherwise
```
bnjbvr: preexisting: code, not coed

crates/matrix-sdk/src/event_cache/mod.rs
```diff
@@ -540,6 +564,8 @@ impl EventCacheInner {
                 )
                 .await?;
 
+                let has_at_least_one_event = room_state.events().revents().next().is_some();
```
bnjbvr: What about a slightly shorter name `nonempty_timeline`?

crates/matrix-sdk/src/event_cache/mod.rs
```diff
@@ -578,6 +615,21 @@ pub struct BackPaginationOutcome {
...
 
+/// Represents an update of a room. It hides the details of
+/// [`RoomEventCacheUpdate`] by being more generic.
+///
+/// This is used by [`EventCache::subscribe_to_room_generic_updates`]. Please
+/// read it to learn more about the motivation behind this type.
+#[derive(Clone, Debug)]
+pub enum RoomEventCacheGenericUpdate {
+    /// The timeline has been updated, i.e. an event has been added, redacted or
+    /// removed.
```
bnjbvr: `added, redacted, removed, or reloaded`?

crates/matrix-sdk/src/event_cache/mod.rs
```diff
@@ -778,4 +839,176 @@ mod tests {
...
+            assert!(generic_stream.recv().now_or_never().is_none());
+        }
+    }
+
+    #[async_test]
+    async fn test_generic_update_when_paginating_room() {
+        // Create 1 room, with 4 chunks in the event cache storage.
+        let user = user_id!("@mnt_io:matrix.org");
+        let client = logged_in_client(None).await;
+        let room_id = room_id!("!raclette:patate.ch");
```
bnjbvr, line 913: yummy

crates/matrix-sdk/src/event_cache/room/mod.rs
```diff
@@ -404,23 +417,34 @@ impl RoomEventCacheInner {
             self.pagination_batch_token_notifier.notify_one();
         }
 
+        let mut events_update_the_timeline = false;
```
bnjbvr: Always feels weird to me to have a boolean that's a sentence; usually I'd advocate for `verb_nouns`. How about `trigger_timeline_update`, or `forward_generic_update`?

bnjbvr (approved):
thanks!

crates/matrix-sdk/src/event_cache/mod.rs
Hywan: You're correct!

crates/matrix-sdk/src/event_cache/mod.rs
Hywan: Sure.

crates/matrix-sdk/src/event_cache/mod.rs
Hywan: Sure.

crates/matrix-sdk/src/event_cache/mod.rs
Hywan: 🫕

crates/matrix-sdk/src/event_cache/room/mod.rs
Hywan: I've renamed to `update_has_been_sent`.

PR #5248, opened 2025-06-18
feat(wasm): Remove network config features from ffi ClientBuilder for Wasm

zzorba:
Features to configure UserAgent, Proxy, Disabling SSL, and additional certificates are not available on Wasm platforms. We remove these configuration options from the FFI layer, while preserving them on non-Wasm platforms.

Signed-off-by: Daniel Salinas

### DIFF
--- a/bindings/matrix-sdk-ffi/src/client_builder.rs
+++ b/bindings/matrix-sdk-ffi/src/client_builder.rs
@@ -286,7 +286,7 @@ pub struct ClientBuilder {
     enable_share_history_on_invite: bool,
     request_config: Option<RequestConfig>,
 
-    // The settings below here will be ignored on Wasm platforms,
+    // The options below here will be ignored on Wasm platforms,
     // because the network stack does not support them.
     user_agent: Option<String>,
     proxy: Option<String>,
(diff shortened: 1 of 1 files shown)


poljar:
Can we rebase this as well, to kickstart the CI.

bindings/matrix-sdk-ffi/src/client_builder.rs
```diff
@@ -274,21 +275,23 @@ pub struct ClientBuilder {
...
     session_delegate: Option<Arc<dyn ClientSessionDelegate>>,
-    additional_root_certificates: Vec<Vec<u8>>,
-    disable_built_in_root_certificates: bool,
     encryption_settings: EncryptionSettings,
     room_key_recipient_strategy: CollectStrategy,
     decryption_trust_requirement: TrustRequirement,
     enable_share_history_on_invite: bool,
     request_config: Option<RequestConfig>,
+    // The settings below here will be ignored on Wasm platforms,
+    // because the network stack does not support them.
```
Hywan: Then why not put them behind `#[cfg(not(target_familiy = "wasm"))]`?

Hywan (changes requested):
Thanks for the changes. I'm okay with them, except that the fields should be behind `#[cfg(not(target_family = "wasm32"))]` instead of being commented as not available. Apart from that, all good.

zzorba:
Our goal is to target wasm32-unknown-unknown for browser support, not wasi.  These features do not exist in the network browser system.

Hywan:
But are they blocking you? Does it forbid `matrix-sdk-ffi` to work on `wasm32-u-u`?

zzorba:
The Certificates and Proxies types do not exist on Wasm platforms, so it does not typecheck/compile.

PR #5250, opened 2025-06-19
change(ffi): expose the full RoomPowerLevels object and move corresponding methods to it

stefanceriu:
This patch expands on the already existent `RoomPowerLevels` record (which it renames to `RoomPowerLevelsValues` to work around uniffi not exposing public fields) and nests them inside a new exported object that also provides methods for retrieving the actions that an user can take, methods moved from the room object. This reduces the amount of async calls the clients need to make, simplifies the API and groups the code better together.

On the sdk/base level those methods can be removed as the underlying power levels are already exposed.


poljar:
> Looks sensible to me, but the CI is moaning about some formatting.

It just needs a bit of a rebase onto main.

stefanceriu:
All green, thank you gents!

PR #5251, opened 2025-06-19
ci: Use the correct token to upload JUnit reports

poljar:


poljar:
> If it fixes things, sure :)

I can't guarantee that, only that it won't make things worse.

Issue #4859, opened 2025-03-28
Support sending to_device messages in the rust sdk widget driver. (MSC3819)


toger5:
The rust sdk widget driver already supports all widget actions required for ElementCall.

It is missing to_device messages however.

## MSC:
https://github.com/matrix-org/matrix-spec-proposals/pull/3819

## WIP PR:
https://github.com/matrix-org/matrix-rust-sdk/pull/4856

fkwp:
https://github.com/matrix-org/matrix-rust-sdk/pull/4963

PR #5252, opened 2025-06-19
feat(widget): Send custom to-device messages in widgets in e2ee rooms

BillCarsonFr:
Adds the capability for widgets to send encrytped to-device messages properly.

It is getting rid of the old encrypted boolean that was used by widget to request encryption. As said by the base MSC
for widget send/receive messages: Widgets should not be aware of encryption.
If the widget is in an encrytped room then to-device traffic will be encrypted.

Fixes https://github.com/matrix-org/matrix-rust-sdk/issues/4859

Signed-off-by:

### DIFF
--- a/crates/matrix-sdk/src/test_utils/mocks/encryption.rs
+++ b/crates/matrix-sdk/src/test_utils/mocks/encryption.rs
@@ -180,8 +180,6 @@ impl MatrixMockServer {
 
     /// Creates a new device and returns a new client for it.
     /// The new and old clients will be aware of each other.
-    /// If possible, the new client will import the secret bundle from the
-    /// initial client
     ///
     /// # Arguments
     ///
@@ -206,19 +204,6 @@ impl MatrixMockServer {
 
         let new_client =
             self.client_builder_for_crypto_end_to_end(&user_id, &new_device_id).build().await;
-
-        // sync the keys
-        self.mock_sync().ok_and_run(&new_client, |_| {}).await;
-
-        {
-            // If cross-signing is enabled, import the secret bundle from the old device.
-            let olm = existing_client.olm_machine_for_testing().await;
-            let olm = olm.as_ref().unwrap();
-            let bundle = olm.store().export_secrets_bundle().await;
-            if let Ok(bundle) = bundle {
-                new_client.encryption().import_secrets_bundle(&bundle).await.unwrap();
-            }
-        }
 
         // sync the keys
         self.mock_sync().ok_and_run(&new_client, |_| {}).await;
--- a/crates/matrix-sdk/src/test_utils/mocks/encryption.rs
+++ b/crates/matrix-sdk/src/test_utils/mocks/encryption.rs
@@ -127,12 +127,12 @@ impl MatrixMockServer {
             self.client_builder_for_crypto_end_to_end(&bob_user_id, &bob_device_id).build().await;
 
         // Have Alice track Bob, so she queries his keys later.
-        MatrixMockServer::update_tracked_users(&alice, [bob_user_id.as_ref()]).await;
+        alice.update_tracked_users_for_testing([bob_user_id.as_ref()]).await;
 
         // let bob be aware of Alice keys in order to be able to decrypt custom
         // to-device (the device keys check are deferred for `m.room.key` so this is not
         // needed for sending room messages for example).
-        MatrixMockServer::update_tracked_users(&bob, [alice_user_id.as_ref()]).await;
+        bob.update_tracked_users_for_testing([alice_user_id.as_ref()]).await;
 
         // Have Alice and Bob upload their signed device keys.
         self.mock_sync().ok_and_run(&alice, |_x| {}).await;
@@ -157,10 +157,10 @@ impl MatrixMockServer {
         self.mock_sync().ok_and_run(&carl, |_| {}).await;
 
         // Have Alice track Carl, so she queries his keys later.
-        MatrixMockServer::update_tracked_users(alice, [carl.user_id().unwrap()]).await;
+        alice.update_tracked_users_for_testing([carl.user_id().unwrap()]).await;
 
         // Have Bob track Carl, so she queries his keys later.
-        MatrixMockServer::update_tracked_users(bob, [carl.user_id().unwrap()]).await;
+        bob.update_tracked_users_for_testing([carl.user_id().unwrap()]).await;
 
         // Have Alice and Bob upload their signed device keys, and download Carl's keys.
         {
@@ -169,11 +169,8 @@ impl MatrixMockServer {
         }
 
         // Let carl be aware of Alice and Bob keys.
-        MatrixMockServer::update_tracked_users(
-            &carl,
-            [alice.user_id().unwrap(), bob.user_id().unwrap()],
-        )
-        .await;
+        carl.update_tracked_users_for_testing([alice.user_id().unwrap(), bob.user_id().unwrap()])
+            .await;
 
         // A last sync for carl to get the keys.
         self.mock_sync().ok_and_run(alice, |_| {}).await;
@@ -242,12 +239,6 @@ impl MatrixMockServer {
         }
 
         new_client
-    }
-
-    async fn update_tracked_users(client: &Client, user_ids: impl IntoIterator<Item = &UserId>) {
-        let olm = client.olm_machine_for_testing().await;
-        let olm = olm.as_ref().unwrap();
-        olm.update_tracked_users(user_ids).await.unwrap();
     }
 
     /// Mock up the various crypto API so that it can serve back keys when
--- a/crates/matrix-sdk/src/test_utils/mocks/encryption.rs
+++ b/crates/matrix-sdk/src/test_utils/mocks/encryption.rs
@@ -173,6 +173,69 @@ impl MatrixMockServer {
         carl
     }
 
+    /// Creates a new device and returns a new client for it.
+    /// The new and old clients will be aware of each other.
+    /// If possible, the new client will import the secret bundle from the
+    /// initial client
+    ///
+    /// # Arguments
+    ///
+    /// * `existing_client` - The original client for which a new device will be
+    ///   created
+    /// * `device_id` - The device ID to use for the new client
+    /// * `clients_to_update` - A vector of client references that should be
+    ///   notified about the new device. These clients will receive a device
+    ///   list change notification during their next sync.
+    ///
+    /// # Returns
+    ///
+    /// Returns the newly created client instance configured for the new device.
+    pub async fn set_up_new_device_for_encryption(
+        &self,
+        existing_client: &Client,
+        device_id: &DeviceId,
+        clients_to_update: Vec<&Client>,
+    ) -> Client {
+        let user_id = existing_client.user_id().unwrap().to_owned();
+        let new_device_id = device_id.to_owned();
+
+        let new_client =
+            self.client_builder_for_crypto_end_to_end(&user_id, &new_device_id).build().await;
+
+        // sync the keys
+        self.mock_sync().ok_and_run(&new_client, |_| {}).await;
+
+        {
+            // If cross-signing is enabled, import the secret bundle from the old device.
+            let olm = existing_client.olm_machine_for_testing().await;
+            let olm = olm.as_ref().unwrap();
+            let bundle = olm.store().export_secrets_bundle().await;
+            if let Ok(bundle) = bundle {
+                new_client.encryption().import_secrets_bundle(&bundle).await.unwrap();
+            }
+        }
+
+        // sync the keys
+        self.mock_sync().ok_and_run(&new_client, |_| {}).await;
+
+        // Notify existing device of a change
+        self.mock_sync()
+            .ok_and_run(existing_client, |builder| {
+                builder.add_change_device(&user_id);
+            })
+            .await;
+
+        for client_to_update in clients_to_update {
+            self.mock_sync()
+                .ok_and_run(client_to_update, |builder| {
+                    builder.add_change_device(&user_id);
+                })
+                .await;
+        }
+
+        new_client
+    }
+
     async fn update_tracked_users(client: &Client, user_ids: impl IntoIterator<Item = &UserId>) {
         let olm = client.olm_machine_for_testing().await;
         let olm = olm.as_ref().unwrap();
--- a/crates/matrix-sdk/src/test_utils/mocks/encryption.rs
+++ b/crates/matrix-sdk/src/test_utils/mocks/encryption.rs
@@ -21,14 +21,17 @@ use std::{
     sync::{atomic::Ordering, Arc, Mutex},
 };
 
+use assert_matches2::assert_let;
 use matrix_sdk_test::test_json;
 use ruma::{
     api::client::{
         keys::upload_signatures::v3::SignedKeys, to_device::send_event_to_device::v3::Messages,
     },
     encryption::{CrossSigningKey, DeviceKeys, OneTimeKey},
+    events::AnyToDeviceEvent,
     owned_device_id, owned_user_id,
     serde::Raw,
+    to_device::DeviceIdOrAllDevices,
     CrossSigningKeyId, DeviceId, MilliSecondsSinceUnixEpoch, OneTimeKeyAlgorithm, OwnedDeviceId,
     OwnedOneTimeKeyId, OwnedUserId, UserId,
 };
@@ -46,6 +49,11 @@ use crate::{
     },
     Client,
 };
+
+/// Stores pending to-device messages for each user and device.
+/// To be used with [`MatrixMockServer::capture_put_to_device_traffic`].
+pub type PendingToDeviceMessages =
+    BTreeMap<OwnedUserId, BTreeMap<OwnedDeviceId, Vec<Raw<AnyToDeviceEvent>>>>;
 
 /// Extends the `MatrixMockServer` with useful methods to help mocking
 /// matrix crypto API and perform integration test with encryption.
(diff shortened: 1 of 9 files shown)


crates/matrix-sdk/src/widget/matrix.rs
```diff
@@ -309,38 +318,137 @@ impl MatrixDriver {
...
+            .latest_encryption_state()
+            .await
+            .map(|s| s.is_encrypted())
+            // Default consider encrypted
+            .unwrap_or(true);
+
+        if room_encrypted {
+            trace!("Sending to-device message in encrypted room <{}>", self.room.room_id());
+
+            //  Attempt to re-batch these by content to batch them up into a single request
```
BillCarsonFr: This part is a bit annoying, and the root cause is I believe a bad design choice on the widget API that wants the same structure than `/sendToDevice` HTTP API.
It is annoying because the crypto API allows to encrypt a same content for a list of recipients (not a bag of different contents). I added a comment on the MSC https://github.com/matrix-org/matrix-spec-proposals/pull/3819#discussion_r2097482594

but anyhow, for now we have to invert this map and re-group per similar content (have to serialise for checking that)

crates/matrix-sdk/src/widget/machine/from_widget.rs
```diff
@@ -245,10 +254,23 @@ impl From<update_delayed_event::unstable::Response> for UpdateDelayedEventRespon
...
-/// serializes to `{}` instead of `Null` when returned to the widget as json.
-#[derive(Serialize, Debug)]
-pub(crate) struct SendToDeviceEventResponse {}
+/// Send to device request response.
+#[derive(Serialize, Debug, Default)]
+pub(crate) struct SendToDeviceEventResponse {
+    /// It is possible that sending to some device failed (failed to encrypt or
+    /// failed to send)
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub failures: Option<BTreeMap<String, Vec<String>>>,
```
BillCarsonFr: For now there is no reason/code for the failures. Will be handled in a follow-up PR

crates/matrix-sdk/src/test_utils/mocks/encryption.rs
```diff
@@ -147,6 +137,48 @@ impl MatrixMockServer {
...
+        self.mock_sync().ok_and_run(alice, |_| {}).await;
+
+        carl
+    }
+
+    async fn update_tracked_users(client: &Client, user_ids: impl IntoIterator<Item = &UserId>) {
+        let olm = client.olm_machine_for_testing().await;
+        let olm = olm.as_ref().unwrap();
+        olm.update_tracked_users(user_ids).await.unwrap();
+    }
```
poljar: This could be a function on the `Client` that is guarded with `#[cfg(test)]`, no?

crates/matrix-sdk/src/widget/machine/from_widget.rs
```diff
@@ -245,10 +254,23 @@ impl From<update_delayed_event::unstable::Response> for UpdateDelayedEventRespon
...
 }
 
-/// The response to the widget that it received the to-device event.
-/// Only used as the response for the successful send case.
-/// FromWidgetErrorResponse will be used otherwise.
-/// This is intentionally an empty tuple struct (not a unit struct), so that it
-/// serializes to `{}` instead of `Null` when returned to the widget as json.
-#[derive(Serialize, Debug)]
-pub(crate) struct SendToDeviceEventResponse {}
+/// Send to device request response.
```
poljar: This comment is much terser, could we please bring back some information the comment contained?

crates/matrix-sdk/src/widget/machine/from_widget.rs
```diff
@@ -245,10 +254,23 @@ impl From<update_delayed_event::unstable::Response> for UpdateDelayedEventRespon
...
+#[derive(Serialize, Debug, Default)]
+pub(crate) struct SendToDeviceEventResponse {
+    /// It is possible that sending to some device failed (failed to encrypt or
+    /// failed to send)
```
poljar: Ok, it's possible that sending may fail, but could we explain how this field relates to this fact? What's contained in the map? What does a `None` mean?

Can't we fold `None` into the empty `BTreeMap` case and skip the serialization if the map is empty?

crates/matrix-sdk/src/widget/matrix.rs
```diff
@@ -309,38 +318,137 @@ impl MatrixDriver {
...
             BTreeMap<DeviceIdOrAllDevices, Raw<AnyToDeviceEventContent>>,
         >,
-    ) -> Result<send_event_to_device::v3::Response> {
+    ) -> Result<SendToDeviceEventResponse> {
+        // TODO: block this at the negotiation stage, no reason to let widget believe
+        // they can do that
+        if Self::is_internal_type(&event_type.to_string()) {
+            warn!("Widget tried to send internal to-device message <{}>, ignoring", event_type);
+            // Silently return a success response, the widget will not receive the message
+            // TODO! this will later be blocked at negotiation level
```
poljar:
```suggestion
            // TODO: this will later be blocked at negotiation level
```

crates/matrix-sdk/src/widget/matrix.rs
```diff
@@ -309,38 +318,137 @@ impl MatrixDriver {
...
-        encrypted: bool,
         messages: BTreeMap<
             OwnedUserId,
             BTreeMap<DeviceIdOrAllDevices, Raw<AnyToDeviceEventContent>>,
         >,
-    ) -> Result<send_event_to_device::v3::Response> {
+    ) -> Result<SendToDeviceEventResponse> {
+        // TODO: block this at the negotiation stage, no reason to let widget believe
+        // they can do that
+        if Self::is_internal_type(&event_type.to_string()) {
```
poljar, line 359: Does this not work?

```suggestion
        if Self::is_internal_type(event_type.as_str()) {
```

crates/matrix-sdk/src/widget/matrix.rs
```diff
@@ -309,38 +318,137 @@ impl MatrixDriver {
...
+            for (user_id, device_map) in messages.iter() {
+                for (device_id, content) in device_map.iter() {
+                    content_grouped
+                        .entry(content.json().get())
+                        .or_default()
+                        .entry(user_id.clone())
+                        .or_default()
+                        .push(device_id.to_owned());
+                }
+            }
```
poljar: This one is a bit trippy, you are reversing the mapping from user -> device -> content to content -> user -> device.

As far as I understand this is an optimization so we can use the function that encrypts the same content for multiple devices.

Could you document this fact, it isn't immediately obvious why this is needed.

Renaming the variable to `content_to_recipients_map` or some similar more obvious name might help as well.

crates/matrix-sdk/src/widget/matrix.rs
```diff
@@ -309,38 +318,137 @@ impl MatrixDriver {
...
 
-        response.map_err(Into::into)
+            // Encrypt and send
```
poljar: This sounds like a perfect place where we could introduce a helper function.

This is getting a bit long I would say.

crates/matrix-sdk/src/widget/matrix.rs
```diff
@@ -309,38 +318,137 @@ impl MatrixDriver {
...
+                            }
+                            DeviceIdOrAllDevices::AllDevices => {
+                                // get all the devices and add them
+                                let user_devices =
+                                    client.encryption().get_user_devices(&user_id).await?;
+                                let devices = user_devices.devices();
+                                devices.for_each(|device| {
+                                    device_set.insert((
+                                        device.user_id().to_owned(),
+                                        device.device_id().to_owned(),
```
poljar: This is a bit weird, you are doing the following:

1. Get all the devices of the user
2. Record the user/device id
3. Throw the devices away
4. Iterate through each recorded user/device id and load each device individually from the store.

It seems to me that it would be much better to:

1. Get all the devices of a user
2. If an alias was used, you have your set.
3. If a device IDs were used, filter out any device that isn't part of the recipient set.

crates/matrix-sdk/src/widget/matrix.rs
```diff
@@ -309,38 +318,137 @@ impl MatrixDriver {
...
+                for (u, d) in device_set {
+                    let found_device = client.encryption().get_device(&u, &d).await?;
+                    let Some(device) = found_device else {
+                        warn!("Device {} not found for user {}", d, u);
+                        // device not found, skip it and log
+                        failures.entry(u).or_default().push(d);
+                        continue;
+                    };
+                    recipients.push(device);
+                }
```
poljar: Once you have all the recipient devices collected, getting the failures would be a simple set operation.

```
let failures = recipient_list - recipient_devices
```

poljar (changes requested):
I think this looks mostly good, there is some surprising logic that we could make a bit more obvious and perhaps even more performant.

crates/matrix-sdk/src/widget/matrix.rs
BillCarsonFr: No there is no such thing for `ToDeviceEventType`, there used to be? but it is now deprecated? like the to_cow_str? (all of that is mind-gated by macro)

crates/matrix-sdk/src/widget/machine/from_widget.rs
BillCarsonFr: It's just me not knowing I could just skip serialization if map empty `[serde(skip_serializing_if = "BTreeMap::is_empty")]`

crates/matrix-sdk/src/widget/matrix.rs
BillCarsonFr: https://github.com/matrix-org/matrix-rust-sdk/pull/5252/commits/8fe836ba607386f625eae7cb9cbb46c52ae31c88

crates/matrix-sdk/src/widget/matrix.rs
BillCarsonFr: Update here https://github.com/matrix-org/matrix-rust-sdk/pull/5252/commits/159a7a1b4a164b7baca87041e0f56b4045a96270

I haven't found the simple way of finding failure like you suggested `let failures = recipient_list - recipient_devices`, I tried something else

crates/matrix-sdk/src/widget/machine/from_widget.rs
BillCarsonFr: https://github.com/matrix-org/matrix-rust-sdk/pull/5252/commits/6647da20617a16b0c08ae67a7ac2c7b050be0ae0

crates/matrix-sdk/src/widget/machine/from_widget.rs
BillCarsonFr: udpated https://github.com/matrix-org/matrix-rust-sdk/pull/5252/commits/c65193771e2c3d48258d4b6d8f1f34c0004560c0

crates/matrix-sdk/src/widget/matrix.rs
BillCarsonFr: done

crates/matrix-sdk/src/test_utils/mocks/encryption.rs
BillCarsonFr: right, I knew it was not the best place but I didn't thought about the cfg(test) thing https://github.com/matrix-org/matrix-rust-sdk/pull/5252/commits/912c7ef16947ad8fc30afed66e621574aeef917a

crates/matrix-sdk/src/widget/machine/from_widget.rs
poljar: You can implement arbitrary functions that tell you when serialization should be skipped. Most `is_empty` functions for any struct should work as well.

crates/matrix-sdk/src/widget/matrix.rs
poljar: Ah :(

crates/matrix-sdk/src/widget/matrix.rs
poljar: This still feels pretty suboptimal:

1. You fetch all devices for each `DeviceId` entry in the list.
2. You collect a bunch of duplicates, which wastes memory until...
3. You deduplicate the list of recipient devices.

What I had in mind would be something like the following diff:

```diff
Commit ID: d205fe05c3577be6ca56d4a994dccbc00be52b7b
Change ID: unuvrlxqtqmnuptxnkxukwuoxmknkktl
Author   : Damir Jelić <poljar@termina.org.uk> (2025-06-26 12:22:21)
Committer: Damir Jelić <poljar@termina.org.uk> (2025-06-26 13:05:20)

    (no description set)

diff --git a/crates/matrix-sdk/src/widget/matrix.rs b/crates/matrix-sdk/src/widget/matrix.rs
index 8fd210baa7..8b1df37c11 100644
--- a/crates/matrix-sdk/src/widget/matrix.rs
+++ b/crates/matrix-sdk/src/widget/matrix.rs
@@ -15,7 +15,7 @@
 //! Matrix driver implementation that exposes Matrix functionality
 //! that is relevant for the widget API.
 
-use std::collections::BTreeMap;
+use std::collections::{BTreeMap, BTreeSet};
 
 use as_variant::as_variant;
 use matrix_sdk_base::deserialized_responses::{EncryptionInfo, RawAnySyncOrStrippedState};
@@ -436,37 +436,48 @@
         let client = self.room.client();
         let mut recipient_devices = Vec::<Device>::new();
 
-        for (user_id, device_id_or_alias) in user_to_list_of_device_id_or_all {
+        for (user_id, recipient_device_ids) in user_to_list_of_device_id_or_all {
             let user_devices = client.encryption().get_user_devices(&user_id).await?;
 
-            if device_id_or_alias.contains(&DeviceIdOrAllDevices::AllDevices) {
-                recipient_devices.extend(user_devices.devices());
+            let user_devices: Vec<Device> = if recipient_device_ids
+                .contains(&DeviceIdOrAllDevices::AllDevices)
+            {
+                // If the user wants to send to all devices there's nothing to filter and no
+                // need to inspect other entries in the users device list.
+                user_devices.devices().collect()
             } else {
-                recipient_devices.extend(user_devices.devices().filter(|d| {
-                    device_id_or_alias
-                        .contains(&DeviceIdOrAllDevices::DeviceId(d.device_id().to_owned()))
-                }));
-
-                let missing_devices = device_id_or_alias
-                    .iter()
-                    .filter_map(|d| {
-                        as_variant!(d, DeviceIdOrAllDevices::DeviceId)
-                            .map(|device_id| device_id.to_owned())
-                    })
-                    .filter(|device_id| user_devices.get(device_id).is_none())
-                    .collect::<Vec<_>>();
-
-                if !missing_devices.is_empty() {
-                    failures.entry(user_id).or_default().extend(missing_devices);
-                }
-            }
+                // If the user wants to send to only some devices, filter out any devices that
+                // aren't part of the recipient_device_ids list.
+                let filtered_devices = user_devices
+                    .devices()
+                    .map(|device| (device.device_id().to_owned(), device))
+                    .filter(|(device_id, _)| {
+                        recipient_device_ids
+                            .contains(&DeviceIdOrAllDevices::DeviceId(device_id.clone()))
+                    });
+
+                let (found_device_ids, devices): (BTreeSet<_>, Vec<_>) = filtered_devices.unzip();
+
+                let list_of_devices: BTreeSet<_> = recipient_device_ids
+                    .into_iter()
+                    .filter_map(|d| as_variant!(d, DeviceIdOrAllDevices::DeviceId))
+                    .collect();
+
+                // Let's now find any devices that are part of the recipient_device_ids list but
+                // were not found in our store.
+                // TODO: Do we want to do this for the case where all devices are part of the
+                // recipient list, in case someone specified all devices as well as some devices
+                // explicitly?
+                let missing_devices =
+                    list_of_devices.difference(&found_device_ids).map(|d| d.to_owned()).collect();
+                failures.insert(user_id, missing_devices);
+
+                devices
+            };
+
+            recipient_devices.extend(user_devices);
         }
 
-        // The user-provided list of devices may contain duplicates, ensure no
-        // duplicates are sent.
-        recipient_devices
-            .dedup_by(|da, db| da.user_id().eq(db.user_id()) && da.device_id().eq(db.device_id()));
-
         if !recipient_devices.is_empty() {
             // need to group by content
             let encrypt_and_send_failures = client

```

poljar (changes requested):
Still not quite to my liking. I left a diff that I think makes things clearer and doesn't load all devices for each user multiple times.

Hopefully it helps out a bit.

crates/matrix-sdk/src/widget/matrix.rs
poljar: > I haven't found the simple way of finding failure like you suggested `let failures = recipient_list - recipient_devices`, I tried something else

Sorry, I meant [BTreeSet::difference()](https://doc.rust-lang.org/std/collections/struct.BTreeSet.html#method.difference) but couldn't remember the name of the method.

crates/matrix-sdk/src/widget/matrix.rs
```diff
@@ -330,38 +340,171 @@ impl MatrixDriver {
...
+            return Ok(Default::default());
+        }
+
         let client = self.room.client();
 
-        let request = if encrypted {
-            return Err(Error::UnknownError(
-                "Sending encrypted to-device events is not supported by the widget driver.".into(),
-            ));
+        let mut failures: BTreeMap<OwnedUserId, Vec<OwnedDeviceId>> = BTreeMap::new();
```
poljar, line 367: Is there a point to initializing this here, mutably borrowing it in the `encrypt_and_send_content_to_devices_helper` call?

Can't `encrypt_and_send_content_to_devices_helper` just return this?

crates/matrix-sdk/src/widget/matrix.rs
```diff
@@ -330,38 +340,171 @@ impl MatrixDriver {
...
+    // Helper to encrypt a single content to several devices
+    async fn encrypt_and_send_content_to_devices_helper(
+        &self,
+        event_type: &ToDeviceEventType,
+        content: &str,
+        user_to_list_of_device_id_or_all: BTreeMap<OwnedUserId, Vec<DeviceIdOrAllDevices>>,
+        failures: &mut BTreeMap<OwnedUserId, Vec<OwnedDeviceId>>,
+    ) -> Result<()> {
+        let client = self.room.client();
+        let mut recipient_devices = Vec::<Device>::new();
```
poljar: Can't it infer the type that goes into the `Vec`?

crates/matrix-sdk/src/widget/matrix.rs
```diff
@@ -330,38 +340,171 @@ impl MatrixDriver {
...
+
+        for (user_id, recipient_device_ids) in user_to_list_of_device_id_or_all {
+            let user_devices = client.encryption().get_user_devices(&user_id).await?;
+
+            let user_devices: Vec<Device> = if recipient_device_ids
```
poljar: The `Vec` might be needed, though I'm doubting that as well, but the `Device` shouldn't. I isn't inside the if/else.   

```suggestion
            let user_devices: Vec<_> = if recipient_device_ids
```

poljar (changes requested):
Ok, I left some two small nits, but I think this is now mostly good.

crates/matrix-sdk/src/widget/matrix.rs
BillCarsonFr: Even if I return a failure map from `encrypt_and_send_content_to_devices_helper` I will still have to intialize a map outside of the loop (`for .. in content_to_recipients_map` )in order to accumalate all the failures. Not sure if it really makes a differenc

crates/matrix-sdk/src/widget/matrix.rs
BillCarsonFr: Right it is not needed, updated.
Sometimes I don't get why type inference need more info. It is often when using `collect`, suddenly it needs type annotation, but a partial one like `Vec<_>`. I need to look out the reason for that, and the `_` thing

poljar: Collect needs to know if you'd like to collect things into a `Vec` or `BTreeSet` or any other collection type.

It knows the type that is inside of the collection, that's why you can use `Vec<_>` instead of the full type.

crates/matrix-sdk/src/widget/matrix.rs
poljar: Alright, let's leave it as is then.

poljar:
Ok if I squash merge this?

BillCarsonFr:
> Ok if I squash merge this?

yes 👍

poljar:
Oh one thing. We should add a changelog entry, though let's do so in a separate PR.

BillCarsonFr:
> Oh one thing. We should add a changelog entry, though let's do so in a separate PR.

https://github.com/matrix-org/matrix-rust-sdk/pull/5323/files

Issue #5123, opened 2025-05-28
threads: data model MVP


bnjbvr:
This is about making it possible to play with thread events, but not necessarily storing them in the persisted event cache yet.

### Core types

- a timeline event may have a `ThreadSummary`, if it's a thread root
  - see also https://github.com/matrix-org/matrix-rust-sdk/issues/5036
  - count (likely not the raw event count, but renderable event count)
  - latest event id (observers can use the `load_or_fetch_event()` method to retrieve it from the cache when needed)
  - participated: TBD
  - maybe_outdated: bool indicating whether the summary might be outdated, after a gappy sync
- for each thread (identified by room id + root event id), a `ThreadEventCache { linked_chunk: LinkedChunk<TimelineEvent, ThreadGap> }` (ThreadGap could be the same as `Gap`)
- the `RoomEventCache` contains a mapping of all the threads roots to their `ThreadEventCache`

### Tasks

- https://github.com/matrix-org/matrix-rust-sdk/issues/3886
- https://github.com/matrix-org/matrix-rust-sdk/issues/4841
- [x] Extract the bundled aggregation from a thread root, and feed it into the `RoomEventCache`, which forwards it to the `ThreadEventCache` ONLY if the thread was not known beforehand
  - [x] Save the bundled thread's latest event into the event cache (don't insert into the thread linked chunk, only save it into the events table with `save_event()`)
- [ ] Have a linked chunk per thread.
- [ ] Allow subscribing to a thread by receiving the initial events in that thread, if known, and a stream of `VectorDiff<Event>` (this will power the thread-focused timeline).
  - [ ] the timeline loads related events itself (to retrieve reactions and so on)
  - [x] make sure relations are correctly ordered, namely edits
- [ ] implement back-pagination at the `ThreadEventCache` level; this will be used by the thread-focused timeline `paginate_backwards()` method.

### MVP

With this version, we might have many empty threads (every time there's a gappy sync), and thread counts might be incorrect.

- When a new sync event is received, and it belongs to a thread:
  - if it's a known thread: append the event to the thread linked chunk
  - if it belongs to an unknown thread: don't push the event, only record the thread summary for the thread root (latest event + increase count?)
- When a sync is gappy (limited):
  - clear _all_ the thread linked chunks 😔 we don't know which threads have missing events, so we have to assume each could've been modified
- When a back-pagination result is returned:
[...]

PR #5258, opened 2025-06-19
feat(event cache): retrieve relations in their local 

bnjbvr:
This makes it so that the relations returned by `RoomEventCache::event_with_relations` are now sorted by their position in the linked chunk, if available (events that can't be located in a linked chunk will be at the front of the array, as they might be "bundled" relations they should be handled in priority).

This will make it possible, for a threaded timeline, to retrieve all related edits in a thread, and to correctly order them (if we've seen them in the main linked chunk first).

(In the meantime, we could even use it to include the edit to a latest thread event as the latest event :eyes: Will look into that as a separate PR.)

On top of #5225. Part of #4869 / #5123.


crates/matrix-sdk-common/src/linked_chunk/relational.rs
```diff
@@ -365,24 +365,29 @@ where
     /// particular order.
     pub fn unordered_linked_chunk_items<'a>(
         &'a self,
-        target: LinkedChunkId<'a>,
+        target: &OwnedLinkedChunkId,
```
poljar, line 371: It's quite strange to have `OwnedLinkedChunkId` and `LinkedChunkId` but also use `&OwnedLinkedChunkId`.

Why don't we have `borrow()` implemented to convert from one to the other?

crates/matrix-sdk-sqlite/src/event_cache_store.rs
```diff
@@ -1190,19 +1193,34 @@ impl EventCacheStore for SqliteEventCacheStore {
...
                     })?
                 {
-                    let ev = ev?;
-                    let ev = serde_json::from_slice(&this.decode_value(&ev)?)?;
-                    related.push(ev);
+                    let (event_blob, chunk_id, index) = result?;
+
+                    let event: Event = serde_json::from_slice(&this.decode_value(&event_blob)?)?;
+
+                    // Only build the position of both the chunk_id and position were present; in
```
poljar:
```suggestion
                    // Only build the position if both the chunk_id and position were present; in
```

crates/matrix-sdk/src/event_cache/room/mod.rs
```diff
@@ -1204,6 +1218,32 @@ mod private {
...
+                        // still consider that not having a position means you'll end at the start
+                        // of the array.
+                        match (lhs, rhs) {
+                            (None, None) => Ordering::Equal,
+                            (None, Some(_)) => Ordering::Less,
+                            (Some(_), None) => Ordering::Greater,
+                            (Some(lhs), Some(rhs)) => lhs.cmp(&rhs),
+                        }
+                    }
+                }
```
poljar, line 1244: This looks like it could be its own function to make it more easily testable. Though it might still be hard to test because it depends on state in `self`. So perhaps this point is moot.

poljar (approved):
Looks good.

crates/matrix-sdk/src/event_cache/room/mod.rs
bnjbvr: Indeed, it's a bit convoluted to make this its own function; would require a trait to get the order. I can change it this way, though, if you think it reads better. It's a bit more concise, but maybe less explicit with respect to events that have a position but not orderings (which are internal errors — maybe I should warn about them), 

```suggestion
                use std::cmp::Ordering;
                match (
                    lhs.and_then(|lhs| self.room_event_order(lhs)),
                    rhs.and_then(|rhs| self.room_event_order(rhs)),
                ) {
                    (None, None) => Ordering::Equal,
                    (None, Some(_)) => Ordering::Less,
                    (Some(_), None) => Ordering::Greater,
                    (Some(lhs), Some(rhs)) => lhs.cmp(&rhs),
                }
            });
```

poljar: I think the `and_then` makes this `then()` a bit harder to reason about.

So the first variant is better if you ask me.

crates/matrix-sdk-common/src/linked_chunk/relational.rs
bnjbvr: I don't think I can do this, because the `Borrow` trait resembles this:

```rust
pub trait Borrow<Borrowed: ?Sized> {
    fn borrow(&self) -> &Borrowed;
}
``` 

So if I implemented it for `OwnedLinkedChunk`, then I'd need to return `&LinkedChunk<'_>`. Unfortunately, it's impossible to construct a temporary `LinkedChunk` and return a reference to it (since it's dropped at the end of the function's scope). I've added a fake `OwnedLinkedChunkId::as_ref()` method, but it can't be used for a `HashMap<OwnedLinkedChunkId, T>` lookup (as the lookups really want `&impl Borrow<HashMapKey>`). Hence this sub-optimal API.

It's something I would appreciate help with, because I've tried many things when implementing the pair `LinkedChunkId` / `OwnedLinkedChunkId`, and couldn't find something that's nice to implement for hashmap lookups.

crates/matrix-sdk/src/event_cache/room/mod.rs
bnjbvr: Alright, let's keep it like this then

crates/matrix-sdk-common/src/linked_chunk/relational.rs
poljar: Yes, for maps you'll need a `Borrow`. I think the docs on `Borrow` explain why:

```rust
/// In particular `Eq`, `Ord` and `Hash` must be equivalent for
/// borrowed and owned values: `x.borrow() == y.borrow()` should give the
/// same result as `x == y`.
///
/// If generic code merely needs to work for all types that can
/// provide a reference to related type `T`, it is often better to use
/// [`AsRef<T>`] as more types can safely implement it.
```

> So if I implemented it for OwnedLinkedChunk, then I'd need to return &LinkedChunk<'_>. Unfortunately, it's impossible to construct a temporary LinkedChunk and return a reference to it (since it's dropped at the end of the function's scope).

Sadly for borrow you usually need a bit of unsafe, you can transmute from one type to the other, this is probably a minimal example:

```rust
use std::collections::HashMap;
use std::borrow::Borrow;

#[derive(PartialEq, Eq, Hash)]
struct Foo(str);

impl From<&str> for &Foo {
    fn from(s: &str) -> Self {
        unsafe { std::mem::transmute(s) }
    }
}

#[derive(PartialEq, Eq, Hash)]
struct OwnedFoo(String);

impl Borrow<Foo> for OwnedFoo {
    fn borrow(&self) -> &Foo {
        self.0.as_str().into()
    }
}

fn main() {
    let owned_foo = OwnedFoo("foo".to_owned());
    let mut map = HashMap::new();
    
    map.insert(owned_foo, 1);
    
    let foo = <&Foo>::from("foo");
    let get = map.get(foo).expect("We should get foo");
    
    println!("Hello {get}")
}
```

https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=870437c03819f5771cf1904d0e68b7f7

You can also take a look at what the Ruma types do using `cargo expand -p ruma-common`:

For example `OwnedRoomId` is defined as:

```rust
pub struct OwnedRoomId {
    inner: Box<RoomId>,
}
```

Which makes a Borrow and AsRef implementation are pretty easy. Though they do require unsafe somewhere:

```rust
impl RoomId {
    pub(super) const fn from_borrowed(s: &str) -> &Self {
        unsafe { std::mem::transmute(s) }
    }
}
```

PR #5259, opened 2025-06-20
refactor(crypto): Rename raw_event to processed_event to reflect its changed state

andybalaam:
At this point in the code it is no longer raw.


Hywan (approved):
Thanks!

PR #5260, opened 2025-06-20
sdk: add logging to `Room::join()`

richvdh:
The receipient side of e2e history sharing doesn't seem to be working. Add some
logging to help diagnose it.


crates/matrix-sdk/src/room/mod.rs
```diff
@@ -397,19 +398,28 @@ impl Room {
...
+        debug!(
+            ?prev_room_state,
+            inviter=?inviter.as_ref().map(|room_member| room_member.user_id()),
+            enable_share_history_on_invite,
+            "Joining room",
+        );
+
         self.client.join_room_by_id(self.room_id()).await?;
 
         #[cfg(feature = "e2e-encryption")]
```
andybalaam, line 416: We could remove this if we were OK with the extra runtime check?

andybalaam (approved):
One very light suggestion.

crates/matrix-sdk/src/room/mod.rs
richvdh: unfortunately that leads to a build error due to the absence of `shared_room_history::maybe_accept_key_bundle`.

PR #5261, opened 2025-06-20
refactor(crypto): Hold a DecryptionSettings in base Client

andybalaam:
Previously, we held a `TrustRequirement` inside `Client`, and had to wrap it in a `DecryptionSettings` manually a couple of times.

It feels more right to hold on to a `DecryptionSettings` directly. If there were ever some additional settings, most likely we would need them all here.

If I haven't screwed it up, this should not affect behaviour in any way.

Prep for https://github.com/matrix-org/matrix-rust-sdk/issues/4147


poljar (approved):
Hmm, I wonder if we should eventually migrate towards a unified setting for the sending and receiving.

Feels like we could have too many options if we don't do that.

andybalaam:
> Hmm, I wonder if we should eventually migrate towards a unified setting for the sending and receiving.

Definitely sounds sensible to me.

PR #5264, opened 2025-06-20
feat(wasm): Address timeline panic to avoid resume_unwind on Wasm

zzorba:
It doesn't seem to be possible to do a resume_unwind on Wasm. Given how severe a panic is, this seems like an acceptable workaround for now.

Signed-off-by: Daniel Salinas


Hywan (approved):
Looks acceptable to me.

PR #5268, opened 2025-06-23
feat(ffi): expose the `RoomPowerLevels` on the `RoomInfo` directly

stefanceriu:
Currently, clients have to make async requests to `Room::get_power_levels` from multiple places throughout the app in order to correctly configure the various UI components and again, on randomly decided events, to keep them up to date.

This patch starts publishing the power levels directly on the `RoomInfo` and allows them to be handled (and updated!) through the normal `subscribe_to_room_info_updates` mechanism.


bnjbvr (commented):
some comment I made elsewhere: I know it's tempting to put [more stuff in RoomInfo](https://github.com/matrix-org/matrix-rust-sdk/pull/5268/files), but (1) that makes it slower to compute every time, and we'll compute it for every single room in the room list, (2) that makes it bigger, and likely increases the cost of getting it over the FFI. Is it really the best place to put it?

It seems there's also some overlap with the `user_levels` computation done a few lines above, maybe that could be reunited?

(Nothing blocking, I just wanted to highlight those possible changes :))

stefanceriu:
As it turns out the existing `user_power_levels` field was already computed by async fetching the `RoomPowerLevels`. I moved that mapping into the newly exposed FFI `RoomPowerLevels` object so that they fetched only once and the mapping computed only on request (iOS is not using it for example). 

So tl;dr; I should actually be a tad faster now 😅

bindings/matrix-sdk-ffi/src/room/power_levels.rs
```diff
@@ -61,6 +85,11 @@ impl RoomPowerLevels {
         Ok(self.inner.user_can_redact_event_of_other(&user_id))
     }
 
+    /// Returns true if the current user is able to kick in the room.
```
bnjbvr: nit: invite

bnjbvr (approved):
Looks great, thanks!

PR #5270, opened 2025-06-23
fix(wasm): Avoid using of tokio::time:sleep which does not work on Wasm

zzorba:
Logging in with a qr-code was failing on Wasm platforms because of the use of tokio::time:sleep.
We already have a wrapped sleep in matrix-sdk-base, so use that instead.

Signed-off-by: Daniel Salinas


stefanceriu (approved):
lgtm 👍

PR #5271, opened 2025-06-23
fix(timeline): take into account the skip count when computing the pagination status

bnjbvr:
The pagination status was only mapped onto the "global" state, that is, the state of the event cache pagination.

Now, consider the following case, where two timeline instances are live:

- the first one could have backpaginated all the items back to the start of the room
- the second one is created later. Because of the initial value for the skip count, it will only return a subset of event items (~20).

However, listening to the pagination status for the second timeline would incorrectly state that the timeline was entirely paginated (because it returned the "global" pagination status). As such, an observer might think that there are no more items in the timeline, while a subsequent pagination would adjust the skip count and return more items.

This fixes it by combining the global pagination state with the local timeline state (aka the skip count value). If the skip count is positive (meaning, we could set it to 0 later and thus returning more events), we pretend we haven't reached the start of the timeline. This way, an observer can call pagination later, which may adjust the skip count and "return" more items.

### DIFF
--- a/crates/matrix-sdk-ui/src/timeline/controller/mod.rs
+++ b/crates/matrix-sdk-ui/src/timeline/controller/mod.rs
@@ -1553,7 +1553,7 @@ impl TimelineController {
     /// case: if the timeline has a skip count greater than 0, it will
     /// ensure that the pagination status says that we haven't reached the
     /// timeline start yet.
-    pub(super) async fn combine_pagination_status(
+    pub(super) async fn map_pagination_status(
         &self,
         status: RoomPaginationStatus,
     ) -> RoomPaginationStatus {
--- a/crates/matrix-sdk-ui/src/timeline/pagination.rs
+++ b/crates/matrix-sdk-ui/src/timeline/pagination.rs
@@ -112,7 +112,7 @@ impl super::Timeline {
 
         let mut status = pagination.status();
 
-        let current_value = self.controller.combine_pagination_status(status.next_now()).await;
+        let current_value = self.controller.map_pagination_status(status.next_now()).await;
 
         let controller = self.controller.clone();
         let stream = Box::pin(stream! {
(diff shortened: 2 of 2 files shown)


crates/matrix-sdk-ui/tests/integration/timeline/pagination.rs
```diff
@@ -169,6 +170,149 @@ async fn test_back_pagination() {
...
+    }
+
+    let (initial_pagination_status2, mut back_pagination_status2) =
+        timeline2.live_back_pagination_status().await.unwrap();
+
+    // …so a caller must have the information that we haven't hit the timeline start
+    // yet.
+    assert_eq!(
+        initial_pagination_status2,
+        RoomPaginationStatus::Idle { hit_timeline_start: false }
```
bnjbvr, line 275: note for the reviewer: the regression test would previously fail here, with `hit_timeline_start` set to `true` (global pagination status for the room).

crates/matrix-sdk-ui/src/timeline/controller/mod.rs
```diff
@@ -1545,6 +1545,36 @@ impl TimelineController {
...
     }
+
+    /// Combine the global (event cache) pagination status with the local state
+    /// of the timeline.
+    ///
+    /// This only changes the global pagination status of this room, in one
+    /// case: if the timeline has a skip count greater than 0, it will
+    /// ensure that the pagination status says that we haven't reached the
+    /// timeline start yet.
+    pub(super) async fn combine_pagination_status(
```
Hywan: I don't understand `combine` here. I'm okay to use this name, but I'm proposing another verb in case it didn't pop in your head already: `map_pagination_status`, because it takes a `RoomPaginationStatus` and returns a `RoomPaginationStatus`. The name isn't perfect though.

PR #5274, opened 2025-06-24
IndexedDB: Add transaction functionality for `EventCacheStore` backend

mgoldenberg:
## Background

This pull request is part of a series of pull requests to add an IndexedDB implementation of the `EventCacheStore` (see #4617, #4996, #5090, #5138, #5226). This particular pull request primarily focuses on providing functionality for easily issuing IndexedDB transaction.

## Changes

The primary change is the addition of `IndexeddbEventCacheStoreTransaction`, which wraps `IdbTransaction` and `IndexedbEventCacheStoreSerializer`. It provides a convenient interface for composing atomic IndexedDB operations. For the sake of brevity, I have only added the generic versions of the functions for these operations, but as we add implementations of the functions in `EventCacheStore`, I will bring in their typed versions where relevant.

Additionally, in order to support some of the generic functions in `IndexeddbEventCacheStoreTransaction`, I have introduced another trait and some minor changes to the existing ones.

- **Added** `IndexedKeyComponentBounds` (trait)
    - Similar to `IndexedKeyBounds`, but generates the lower and upper bounds of the key components, rather than the key itself. This is useful when constructing key ranges, except in the case where a particular key component is encrypted, in which case one must construct the ranges from the keys themselves.
- **Added** `IndexedKeyRange` (type)
    - A type that can be used to express individual keys, as well as key ranges.
- **Changed** `Indexed` (trait)
    - Tracks the name of the object store in which the implementing type stored using an associated constant, `OBJECT_STORE: &'static str`
- **Changed** `IndexedKey` (trait)
    - Tracks the name of the index with which the key is associated (if any) using a function, `index() -> Option<&'static str>`
- **Changed** `IndexedKeyBounds` (trait)
    - `encode_lower_key` -> `lower_key` to more closely mirror `IndexedKeyComponentBounds`
    - `encode_upper_key` -> `upper_key` to more closely mirror `IndexedKeyComponentBounds` 
- **Changed** `IndexeddbEventCacheStoreSerializerError`
    - `Indexing` variant uses generic type, rather than dynamic type so that `Send` and `Sync` don't have to be used as constraints where they aren't necessary.
 
## Future Work

- Add implementation of `EventCacheStore` which uses the transaction type added in this pull request.

---

### DIFF
--- a/crates/matrix-sdk-indexeddb/src/event_cache_store/error.rs
+++ b/crates/matrix-sdk-indexeddb/src/event_cache_store/error.rs
@@ -12,10 +12,12 @@
 // See the License for the specific language governing permissions and
 // limitations under the License
 
-#![allow(unused)]
+use matrix_sdk_base::{SendOutsideWasm, SyncOutsideWasm};
 
-mod error;
-mod migrations;
-mod serializer;
-mod transaction;
-mod types;
+/// A trait that combines the necessary traits needed for asynchronous runtimes,
+/// but excludes them when running in a web environment - i.e., when
+/// `#[cfg(target_family = "wasm")]`.
+pub trait AsyncErrorDeps: std::error::Error + SendOutsideWasm + SyncOutsideWasm + 'static {}
+
+impl<T> AsyncErrorDeps for T where T: std::error::Error + SendOutsideWasm + SyncOutsideWasm + 'static
+{}
--- a/crates/matrix-sdk-indexeddb/src/event_cache_store/serializer/traits.rs
+++ b/crates/matrix-sdk-indexeddb/src/event_cache_store/serializer/traits.rs
@@ -73,6 +73,22 @@ pub trait IndexedKey<T: Indexed> {
 /// A trait for constructing the bounds of an [`IndexedKey`].
 ///
 /// This is useful when constructing range queries in IndexedDB.
+///
+/// The [`IndexedKeyComponentBounds`] helps to specify the upper and lower
+/// bounds of the components that are used to create the final key, while the
+/// `IndexedKeyBounds` are the upper and lower bounds of the final key itself.
+///
+/// While these concepts are similar and often produce the same results, there
+/// are cases where these two concepts produce very different results. Namely,
+/// when any of the components are encrypted in the process of constructing the
+/// final key, then the component bounds and the key bounds produce very
+/// different results.
+///
+/// So, for instance, consider the `EventId`, which may be encrypted before
+/// being used in a final key. One cannot construct the upper and lower bounds
+/// of the final key using the upper and lower bounds of the `EventId`, because
+/// once the `EventId` is encrypted, the resultant value will no longer express
+/// the proper bound.
 pub trait IndexedKeyBounds<T: Indexed>: IndexedKey<T> {
     /// Constructs the lower bound of the key.
     fn lower_key(room_id: &RoomId, serializer: &IndexeddbSerializer) -> Self;
@@ -102,6 +118,9 @@ where
 /// This is useful when constructing range queries in IndexedDB. Note that this
 /// trait should not be implemented for key components that are going to be
 /// encrypted as ordering properties will not be preserved.
+///
+/// One may be interested to read the documentation of [`IndexedKeyBounds`] to
+/// get a better overview of how these two interact.
 pub trait IndexedKeyComponentBounds<T: Indexed>: IndexedKeyBounds<T> {
     /// Constructs the lower bound of the key components.
     fn lower_key_components() -> Self::KeyComponents;
--- a/crates/matrix-sdk-indexeddb/src/event_cache_store/serializer/traits.rs
+++ b/crates/matrix-sdk-indexeddb/src/event_cache_store/serializer/traits.rs
@@ -24,6 +24,9 @@ use crate::serializer::IndexeddbSerializer;
 /// decryption, in the case the high-level type must be encrypted before
 /// storage.
 pub trait Indexed: Sized {
+    /// The name of the object store in IndexedDB.
+    const OBJECT_STORE: &'static str;
+
     /// The indexed type that is used for storage in IndexedDB.
     type IndexedType;
     /// The error type that is returned when conversion fails.
--- a/crates/matrix-sdk-indexeddb/src/event_cache_store/serializer/traits.rs
+++ b/crates/matrix-sdk-indexeddb/src/event_cache_store/serializer/traits.rs
@@ -50,6 +50,9 @@ pub trait Indexed: Sized {
 ///
 /// Each implementation represents a key on an [`Indexed`] type.
 pub trait IndexedKey<T: Indexed> {
+    /// The index name for the key, if it represents an index.
+    const INDEX: Option<&'static str> = None;
+
     /// Any extra data used to construct the key.
     type KeyComponents;
 
--- a/crates/matrix-sdk-indexeddb/src/event_cache_store/serializer/traits.rs
+++ b/crates/matrix-sdk-indexeddb/src/event_cache_store/serializer/traits.rs
@@ -68,8 +68,14 @@ pub trait IndexedKey<T: Indexed> {
 ///
 /// This is useful when constructing range queries in IndexedDB.
 pub trait IndexedKeyBounds<T: Indexed>: IndexedKey<T> {
+    /// Constructs the lower bound of the key components.
+    fn lower_key_components() -> Self::KeyComponents;
+
     /// Encodes the lower bound of the key.
     fn encode_lower(room_id: &RoomId, serializer: &IndexeddbSerializer) -> Self;
+
+    /// Constructs the upper bound of the key components.
+    fn upper_key_components() -> Self::KeyComponents;
 
     /// Encodes the upper bound of the key.
     fn encode_upper(room_id: &RoomId, serializer: &IndexeddbSerializer) -> Self;
--- a/crates/matrix-sdk-indexeddb/src/event_cache_store/serializer/traits.rs
+++ b/crates/matrix-sdk-indexeddb/src/event_cache_store/serializer/traits.rs
@@ -68,25 +68,38 @@ pub trait IndexedKey<T: Indexed> {
 ///
 /// This is useful when constructing range queries in IndexedDB.
 pub trait IndexedKeyBounds<T: Indexed>: IndexedKey<T> {
-    /// Constructs the lower bound of the key components.
-    fn lower_key_components() -> Self::KeyComponents;
-
     /// Constructs the lower bound of the key.
-    fn lower_key(room_id: &RoomId, serializer: &IndexeddbSerializer) -> Self
-    where
-        Self: Sized,
-    {
+    fn lower_key(room_id: &RoomId, serializer: &IndexeddbSerializer) -> Self;
+
+    /// Constructs the upper bound of the key.
+    fn upper_key(room_id: &RoomId, serializer: &IndexeddbSerializer) -> Self;
+}
+
+impl<T, K> IndexedKeyBounds<T> for K
+where
+    T: Indexed,
+    K: IndexedKeyComponentBounds<T> + Sized,
+{
+    /// Constructs the lower bound of the key.
+    fn lower_key(room_id: &RoomId, serializer: &IndexeddbSerializer) -> Self {
         <Self as IndexedKey<T>>::encode(room_id, &Self::lower_key_components(), serializer)
     }
 
-    /// Constructs the upper bound of the key components.
-    fn upper_key_components() -> Self::KeyComponents;
-
     /// Constructs the upper bound of the key.
-    fn upper_key(room_id: &RoomId, serializer: &IndexeddbSerializer) -> Self
-    where
-        Self: Sized,
-    {
+    fn upper_key(room_id: &RoomId, serializer: &IndexeddbSerializer) -> Self {
         <Self as IndexedKey<T>>::encode(room_id, &Self::upper_key_components(), serializer)
     }
+}
+
+/// A trait for constructing the bounds of the components of an [`IndexedKey`].
+///
+/// This is useful when constructing range queries in IndexedDB. Note that this
+/// trait should not be implemented for key components that are going to be
+/// encrypted as ordering properties will not be preserved.
+pub trait IndexedKeyComponentBounds<T: Indexed>: IndexedKeyBounds<T> {
+    /// Constructs the lower bound of the key components.
+    fn lower_key_components() -> Self::KeyComponents;
+
+    /// Constructs the upper bound of the key components.
+    fn upper_key_components() -> Self::KeyComponents;
 }
--- a/crates/matrix-sdk-indexeddb/src/event_cache_store/serializer/traits.rs
+++ b/crates/matrix-sdk-indexeddb/src/event_cache_store/serializer/traits.rs
@@ -71,8 +71,8 @@ pub trait IndexedKeyBounds<T: Indexed>: IndexedKey<T> {
     /// Constructs the lower bound of the key components.
     fn lower_key_components() -> Self::KeyComponents;
 
-    /// Encodes the lower bound of the key.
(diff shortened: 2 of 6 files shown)


crates/matrix-sdk-indexeddb/src/event_cache_store/serializer/types.rs
```diff
@@ -292,6 +308,14 @@ impl IndexedKey<Event> for IndexedEventIdKey {
 }
 
 impl IndexedKeyBounds<Event> for IndexedEventIdKey {
+    fn lower_key_components() -> Self::KeyComponents {
+        OwnedEventId::try_from(format!("${INDEXED_KEY_LOWER_CHARACTER}")).expect("valid event id")
+    }
+
+    fn upper_key_components() -> Self::KeyComponents {
+        OwnedEventId::try_from(format!("${INDEXED_KEY_UPPER_CHARACTER}")).expect("valid event id")
```
Hywan, line 366: Same here, for `INDEXED_KEY_UPPER_CHARACTER` please?

crates/matrix-sdk-indexeddb/src/event_cache_store/serializer/types.rs
```diff
@@ -292,6 +308,14 @@ impl IndexedKey<Event> for IndexedEventIdKey {
 }
 
 impl IndexedKeyBounds<Event> for IndexedEventIdKey {
+    fn lower_key_components() -> Self::KeyComponents {
+        OwnedEventId::try_from(format!("${INDEXED_KEY_LOWER_CHARACTER}")).expect("valid event id")
```
Hywan, line 362: Can we add a documentation line on `INDEXED_KEY_LOWER_CHARACTER` that it must follow [the event ID format](https://spec.matrix.org/v1.14/appendices/#event-ids)?

```suggestion
        OwnedEventId::try_from(format!("${INDEXED_KEY_LOWER_CHARACTER}")).expect("invalid event ID")
```

crates/matrix-sdk-indexeddb/src/event_cache_store/serializer/types.rs
```diff
@@ -292,6 +308,14 @@ impl IndexedKey<Event> for IndexedEventIdKey {
...
+    }
+
+    fn upper_key_components() -> Self::KeyComponents {
+        OwnedEventId::try_from(format!("${INDEXED_KEY_UPPER_CHARACTER}")).expect("valid event id")
+    }
+
     fn encode_lower(room_id: &RoomId, serializer: &IndexeddbSerializer) -> Self {
         let room_id = serializer.encode_key_as_string(keys::ROOMS, room_id);
         let event_id = String::from(INDEXED_KEY_LOWER_CHARACTER);
```
Hywan: We don't need a `$` for the `event_id` here (and in `encode_upper`?

crates/matrix-sdk-indexeddb/src/event_cache_store/serializer/traits.rs
```diff
@@ -72,11 +72,21 @@ pub trait IndexedKeyBounds<T: Indexed>: IndexedKey<T> {
     fn lower_key_components() -> Self::KeyComponents;
 
     /// Encodes the lower bound of the key.
-    fn encode_lower(room_id: &RoomId, serializer: &IndexeddbSerializer) -> Self;
+    fn encode_lower(room_id: &RoomId, serializer: &IndexeddbSerializer) -> Self
+    where
+        Self: Sized,
+    {
+        <Self as IndexedKey<T>>::encode(room_id, &Self::lower_key_components(), serializer)
```
Hywan, line 107: Neat!

crates/matrix-sdk-indexeddb/src/event_cache_store/serializer/types.rs
```diff
@@ -60,6 +60,57 @@ const INDEXED_KEY_LOWER_CHARACTER: char = '\u{0000}';
 /// [1]: https://en.wikipedia.org/wiki/Plane_(Unicode)#Basic_Multilingual_Plane
 const INDEXED_KEY_UPPER_CHARACTER: char = '\u{FFFF}';
 
+/// Representation of a range of keys of type `K`. This is loosely
+/// correlated with IndexedDB key ranges, with some negligible differences.
```
Hywan: What are the differences 😛? (you knew I would ask, aren't you?)

crates/matrix-sdk-indexeddb/src/event_cache_store/serializer/traits.rs
```diff
@@ -53,6 +53,11 @@ pub trait IndexedKey<T: Indexed> {
     /// Any extra data used to construct the key.
     type KeyComponents;
 
+    /// The index name for the key, if it represents an index.
+    fn index() -> Option<&'static str> {
```
Hywan: It can be a const:

```rust
const INDEX: Option<&'static str>;
```

Just like you did with `Indexed::OBJECT_STORE` in https://github.com/matrix-org/matrix-rust-sdk/pull/5274/commits/fea765edb61d05139c3fed7e4e53ac9fb7fbf69b.

crates/matrix-sdk-indexeddb/src/event_cache_store/error.rs
```diff
@@ -0,0 +1,23 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License
+
+use matrix_sdk_base::{SendOutsideWasm, SyncOutsideWasm};
+
+/// A trait that combines the necessary traits needed for asynchronous runtimes,
+/// but excludes them when running in a web environment - i.e., when
+/// `#[cfg(target_arch = "wasm32")]`.
```
Hywan:
```suggestion
/// `#[cfg(target_family = "wasm")]`.
```

crates/matrix-sdk-indexeddb/src/event_cache_store/transaction.rs
```diff
@@ -66,4 +79,192 @@ impl<'a> IndexeddbEventCacheStoreTransaction<'a> {
...
+    ) -> Result<Vec<T>, IndexeddbEventCacheStoreTransactionError>
+    where
+        T: Indexed,
+        T::IndexedType: DeserializeOwned,
+        T::Error: AsyncErrorDeps,
+        K: IndexedKeyBounds<T> + Serialize,
+    {
+        let range = self.serializer.encode_key_range::<T, K>(room_id, range)?;
+        let object_store = self.transaction.object_store(T::OBJECT_STORE)?;
+        let array = if let Some(index) = K::index() {
```
Hywan: With the `const`, it would be become:

```suggestion
        let array = if let Some(index) = K::INDEX {
```

crates/matrix-sdk-indexeddb/src/event_cache_store/transaction.rs
```diff
@@ -66,4 +79,192 @@ impl<'a> IndexeddbEventCacheStoreTransaction<'a> {
...
+            object_store.index(index)?.get_all_with_key(&range)?.await?
+        } else {
+            object_store.get_all_with_key(&range)?.await?
+        };
+        let mut items = Vec::new();
```
Hywan: Would be nice if `range` could tell how many items it will fetch, so that we can `Vec::with_capacity(range.len());`. Is it possible?

Edit: even simpler:

```suggestion
        let mut items = Vec::with_capacity(array.length() as usize);
```

crates/matrix-sdk-indexeddb/src/event_cache_store/serializer/types.rs
mgoldenberg: I think this comment is only relevant to a particular set of commits, as this code is not present in the latest version of this branch...

crates/matrix-sdk-indexeddb/src/event_cache_store/serializer/types.rs
mgoldenberg: I think this comment is only relevant to a particular set of commits, as this code is not present in the latest version of this branch...

crates/matrix-sdk-indexeddb/src/event_cache_store/serializer/types.rs
mgoldenberg: I think this comment is only relevant to a particular set of commits, as this code is not present in the latest version of this branch...

mgoldenberg:
A quick comment about the [`IndexedKeyComponentBounds`](https://github.com/matrix-org/matrix-rust-sdk/blob/2f0d2e466d706c9b3f69c15f73caa7e35e798165/crates/matrix-sdk-indexeddb/src/event_cache_store/serializer/traits.rs#L107) and [`IndexedKeyBounds`](https://github.com/matrix-org/matrix-rust-sdk/blob/2f0d2e466d706c9b3f69c15f73caa7e35e798165/crates/matrix-sdk-indexeddb/src/event_cache_store/serializer/traits.rs#L78) traits to help clarify some things.

The `IndexedKeyComponentBounds` helps to specify the upper and lower bounds of the components that are used to create the final key, while the `IndexedKeyBounds` are the upper and lower bounds of the final key itself. 

While these concepts are similar and often produce the same results, there are cases where these two concepts produce very different results. Namely, when any of the components are encrypted in the process of constructing the final key, then the component bounds and the key bounds produce very different results. 

So, for instance, consider the `EventId`, which may be encrypted before being used in a final key. One cannot construct the upper and lower bounds of the final key using the upper and lower bounds of the `EventId`, because once the `EventId` is encrypted, the resultant value will no longer express the proper bound.

Note that this is mentioned in the final version of the traits, see [here](https://github.com/matrix-org/matrix-rust-sdk/blob/2f0d2e466d706c9b3f69c15f73caa7e35e798165/crates/matrix-sdk-indexeddb/src/event_cache_store/serializer/traits.rs#L104-L107).

crates/matrix-sdk-indexeddb/src/event_cache_store/serializer/traits.rs
mgoldenberg: Nice, good thinking!

mgoldenberg: I had to do this in a few fixup commits, not sure if there was a better way...

- Add `const INDEX`, fixup: [`d94846a9b297bcf680cb382c5969e56a5ab9b683`](https://github.com/matrix-org/matrix-rust-sdk/pull/5274/commits/d94846a9b297bcf680cb382c5969e56a5ab9b683)
- Use `const INDEX` in get functions, fixup: [`3cd7a14178974054223abb91b6bd68ca5acdf2f3`](https://github.com/matrix-org/matrix-rust-sdk/pull/5274/commits/3cd7a14178974054223abb91b6bd68ca5acdf2f3)
- Use `const INDEX` in delete functions, fixup: [`a1c3eda844c403ae8ad7718743b3f4ec90f38a49`](https://github.com/matrix-org/matrix-rust-sdk/pull/5274/commits/a1c3eda844c403ae8ad7718743b3f4ec90f38a49)
- Remove `fn index()`, fixup [`b86d13a9754f1f67859eafae15de57d4f41a988d`](https://github.com/matrix-org/matrix-rust-sdk/pull/5274/commits/b86d13a9754f1f67859eafae15de57d4f41a988d)

crates/matrix-sdk-indexeddb/src/event_cache_store/transaction.rs
mgoldenberg: I had to do this in a few fixup commits, not sure if there was a better way...

- Add `const INDEX`, fixup: [`d94846a9b297bcf680cb382c5969e56a5ab9b683`](https://github.com/matrix-org/matrix-rust-sdk/pull/5274/commits/d94846a9b297bcf680cb382c5969e56a5ab9b683)
- Use `const INDEX` in get functions, fixup: [`3cd7a14178974054223abb91b6bd68ca5acdf2f3`](https://github.com/matrix-org/matrix-rust-sdk/pull/5274/commits/3cd7a14178974054223abb91b6bd68ca5acdf2f3)
- Use `const INDEX` in delete functions, fixup: [`a1c3eda844c403ae8ad7718743b3f4ec90f38a49`](https://github.com/matrix-org/matrix-rust-sdk/pull/5274/commits/a1c3eda844c403ae8ad7718743b3f4ec90f38a49)
- Remove `fn index()`, fixup [`b86d13a9754f1f67859eafae15de57d4f41a988d`](https://github.com/matrix-org/matrix-rust-sdk/pull/5274/commits/b86d13a9754f1f67859eafae15de57d4f41a988d)

crates/matrix-sdk-indexeddb/src/event_cache_store/error.rs
mgoldenberg: fixup: [`3a4150571ff3a0a36db4442ed1868051bc0983e3`](https://github.com/matrix-org/matrix-rust-sdk/pull/5274/commits/3a4150571ff3a0a36db4442ed1868051bc0983e3)

crates/matrix-sdk-indexeddb/src/event_cache_store/transaction.rs
mgoldenberg: fixup: [`1f997d546e1736737b99f352d62d788b0fbcc68f`](https://github.com/matrix-org/matrix-rust-sdk/pull/5274/commits/1f997d546e1736737b99f352d62d788b0fbcc68f)

crates/matrix-sdk-indexeddb/src/event_cache_store/serializer/types.rs
mgoldenberg: fixup: [`bb8c01b28f40d6ec72af34c637af49e12ed5032a`](https://github.com/matrix-org/matrix-rust-sdk/pull/5274/commits/bb8c01b28f40d6ec72af34c637af49e12ed5032a)

mgoldenberg:
> I was wondering: can we add tests for this? Or is it for a next PR?

So, in my branch with all the changes, I only have tests of the `EventCacheStore` functions, as I believe that will ultimately be the only public interface. 

Do you think we should have more than that? If so, I'd be happy to add more now or in a future pull request. 

For context, I'll also be adding typed versions of all the functions in `IndexeddbEventCacheStoreTransaction` - e.g., `get_chunks_by_id`, `add_event`, `delete_gap_by_id`, etc. - as they are needed in the implementation of the `EventCacheStore` functions. Would you want to see tests of those typed functions and the generic ones?

Let me know what you're thinking!

crates/matrix-sdk-indexeddb/src/event_cache_store/serializer/types.rs
Hywan: Hmm yes. I was reviewing commit by commit. All good.

crates/matrix-sdk-indexeddb/src/event_cache_store/serializer/types.rs
Hywan: :+1:

crates/matrix-sdk-indexeddb/src/event_cache_store/serializer/traits.rs
Hywan: It's fine, thanks!

Hywan:
About https://github.com/matrix-org/matrix-rust-sdk/pull/5274#issuecomment-3001669854 — It's great the last part is mentioned in the doc but the first 2/3 of the comment is also truly valuable! Can you add that on the documentation of `IndexedKeyBounds` for example.

About https://github.com/matrix-org/matrix-rust-sdk/pull/5274#issuecomment-3001846036 — It depends of how much efforts it represents for you. We already have unit tests and integration tests for the `EventCacheStore` implementations (via macros). But for particular implementation, we also have custom unit tests. If you believe this is necessary, I'm all for it of course, but it's not a blocker for me yet.

Hywan:
You can rebase your fixup commits, then we can merge the PR.

Hywan (approved):

Issue #5107, opened 2025-05-27
History sharing: support for indexeddb


richvdh:
We've added a bit more functionality to the sqlite store; need to do the same for indexeddb.

Part of https://github.com/element-hq/element-meta/issues/2829

PR #5276, opened 2025-06-24
Indexeddb: support for received room key bundles

richvdh:
Add a new store to keep track of the information we have received about room key bundles for MSC4268.

Fixes: https://github.com/matrix-org/matrix-rust-sdk/issues/5107

### DIFF
--- a/crates/matrix-sdk-indexeddb/src/crypto_store/migrations/v11_to_v12.rs
+++ b/crates/matrix-sdk-indexeddb/src/crypto_store/migrations/v11_to_v12.rs
@@ -12,9 +12,6 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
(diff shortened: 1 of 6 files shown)


crates/matrix-sdk-indexeddb/src/crypto_store/migrations/v11_to_v12.rs
```diff
@@ -12,9 +12,6 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-//! Migration code that moves from inbound_group_sessions2 to
-//! inbound_group_sessions3, shrinking the values stored in each record.
-
```
richvdh, line 17: a driveby fix: this comment had been incorrectly c&ped from somewhere else

PR #5279, opened 2025-06-24
feat(event cache): add one linked chunk per thread

### LINKED ISSUE #5123: threads: data model MVP

bnjbvr:
This adds one linked chunk per thread, in the event cache. The main benefit for the embedders will be that a threaded timeline won't always start fresh network requests, if it had some threads in the in-memory cache. Persistent storage of threads will come in a second time.

At this point, a bit of functionality is still duplicated, because we can't rely on the exact same mechanism that exist for the room linked chunk (since the latter lives both in memory and on disk). Fortunately, the feature set is quite small.

There are few tests in this PR, because most of it was indirectly tested in some timeline integration tests, which still pass, a sign that the feature is working as intended :)

Fixes #5123. Part of #4869.

### DIFF
--- a/labs/multiverse/src/widgets/room_view/mod.rs
+++ b/labs/multiverse/src/widgets/room_view/mod.rs
@@ -396,9 +396,9 @@ impl RoomView {
 
         let status_handle = self.status_handle.clone();
 
-        // Request to back-paginate 20 events.
+        // Request to back-paginate 5 events.
         *pagination = Some(spawn(async move {
-            if let Err(err) = sdk_timeline.paginate_backwards(20).await {
+            if let Err(err) = sdk_timeline.paginate_backwards(5).await {
                 status_handle.set_message(format!("Error during backpagination: {err}"));
             }
         }));


crates/matrix-sdk/src/event_cache/room/mod.rs
```diff
@@ -1434,6 +1661,17 @@ mod private {
                 prev_batch = None;
             }
 
+            if prev_batch.is_some() {
+                // Sad time: there's a gap, somewhere, in the timeline, and there's at least one
+                // non-duplicated event. We don't know which threads might have gappy, so we
+                // must invalidate them all :(
+                // TODO(bnjbvr): absorb moar caffeine and create a good solution to this
```
Hywan, line 1702: ☕

Hywan (approved):
Looks good to me :-). Kudos.

PR #5280, opened 2025-06-25
fix(ffi): make `RoomInfo` power levels optional as they can be missing depending on the required state

stefanceriu:
The `m.room.power_levels` state event is not part of the room list required state and computing the RoomInfo would fail in that case.


Hywan (approved):
Can we add this state event to the room list too? Maybe in this PR?

Hywan:
Edit: the state event is already fetched by the room list, see:

https://github.com/matrix-org/matrix-rust-sdk/blob/22c99f30f3fee6c0c32ad2eb4baa45e56f87a9bc/crates/matrix-sdk-ui/src/room_list_service/mod.rs#L86

I think there is something else.

stefanceriu:
> Can we add this state event to the room list too? Maybe in this PR?

Turns out it's already in there ¯\_(ツ)_/¯

PR #5282, opened 2025-06-25
refactor(crypto): Provide a specific error type for to-device events …

andybalaam:
…from dehydrated devices

This will get more usage soon, when we add a variant for events with unverified senders.

Prep for https://github.com/matrix-org/matrix-rust-sdk/issues/4147


poljar (approved):
Sensible, the fact that we're not unwedging sessions from dehydrated devices anymore is a win as well.

PR #5284, opened 2025-06-25
fix(sdk): correctly import e2ee history in `join_room_by_id`

richvdh:
It turns out that downstream clients can and do call `Client::join_room_by_id()` rather than `Room::join`, so we need to do the room key history import in the lower-level method.

### DIFF
--- a/crates/matrix-sdk/src/client/mod.rs
+++ b/crates/matrix-sdk/src/client/mod.rs
@@ -93,6 +93,7 @@ use crate::{
     http_client::HttpClient,
     media::MediaError,
     notification_settings::NotificationSettings,
+    room::RoomMember,
     room_preview::RoomPreview,
     send_queue::SendQueueData,
     sliding_sync::Version as SlidingSyncVersion,
@@ -1429,11 +1430,42 @@ impl Client {
         }
     }
 
+    /// Prepare to join a room by ID, by getting the current details about it
+    async fn prepare_join_room_by_id(&self, room_id: &RoomId) -> Option<PreJoinRoomInfo> {
+        let Some(room) = self.get_room(room_id) else {
+            // We know nothing about this room.
+            return None;
+        };
+
+        let inviter = match room.invite_details().await {
+            Ok(details) => details.inviter,
+            Err(Error::WrongRoomState(_)) => None,
+            Err(e) => {
+                warn!("Error fetching invite details for room: {e:?}");
+                None
+            }
+        };
+
+        Some(PreJoinRoomInfo { inviter })
+    }
+
     /// Finish joining a room.
     ///
     /// If the room was an invite that should be marked as a DM, will include it
     /// in the DM event after creating the joined room.
-    async fn finish_join_room(&self, room_id: &RoomId) -> Result<Room> {
+    ///
+    /// If encrypted history sharing is enabled, will check to see if we have a
+    /// key bundle, and import it if so.
+    ///
+    /// # Arguments
+    ///
+    /// * `room_id` - The `RoomId` of the room that was joined.
+    /// * `pre_join_room_info` - Information about the room before we joined.
+    async fn finish_join_room(
+        &self,
+        room_id: &RoomId,
+        pre_join_room_info: Option<PreJoinRoomInfo>,
+    ) -> Result<Room> {
         let mark_as_dm = if let Some(room) = self.get_room(room_id) {
             room.state() == RoomState::Invited
                 && room.is_direct().await.unwrap_or_else(|e| {
@@ -1451,6 +1483,16 @@ impl Client {
             room.set_is_direct(true).await?;
         }
 
+        #[cfg(feature = "e2e-encryption")]
+        if self.inner.enable_share_history_on_invite {
+            if let Some(inviter) =
+                pre_join_room_info.as_ref().and_then(|info| info.inviter.as_ref())
+            {
+                crate::room::shared_room_history::maybe_accept_key_bundle(&room, inviter.user_id())
+                    .await?;
+            }
+        }
+
         Ok(room)
     }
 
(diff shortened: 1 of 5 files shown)


crates/matrix-sdk/src/client/mod.rs
```diff
@@ -1429,11 +1430,42 @@ impl Client {
         }
     }
 
+    /// Prepare to join a room by ID, by getting the current details about it
+    async fn prepare_join_room_by_id(&self, room_id: &RoomId) -> Option<PreJoinRoomInfo> {
+        let Some(room) = self.get_room(room_id) else {
+            // We know nothing about this room.
+            return None;
+        };
```
bnjbvr:
```suggestion
        let room = self.get_room(room_id)?;
```

crates/matrix-sdk/src/client/mod.rs
```diff
@@ -2736,6 +2783,13 @@ impl<Value> CachedValue<Value> {
     }
 }
 
+/// Information about the state of a room before we joined it.
+#[derive(Debug, Clone, Default)]
+struct PreJoinRoomInfo {
+    /// The user that invited us to the room, if any
+    pub inviter: Option<RoomMember>,
+}
```
bnjbvr: Can you include the room state prior to accepting the join, or even better, a `mark_as_dm` bool that follows the logic that's currently in `finish_join_room`? Then, we can fix the race you've spotted with respect to sync vs join (and include a test for it :upside_down_face:).

crates/matrix-sdk/src/client/mod.rs
```diff
@@ -1486,7 +1533,7 @@ impl Client {
             via: server_names.to_owned(),
         });
         let response = self.send(request).await?;
-        self.finish_join_room(&response.room_id).await
+        self.finish_join_room(&response.room_id, None).await
```
bnjbvr: We can likely do better, if the `alias` is a room id?

crates/matrix-sdk/src/client/mod.rs
```diff
@@ -2736,6 +2783,13 @@ impl<Value> CachedValue<Value> {
     }
 }
 
+/// Information about the state of a room before we joined it.
+#[derive(Debug, Clone, Default)]
+struct PreJoinRoomInfo {
+    /// The user that invited us to the room, if any
```
bnjbvr: I've been taught it's not polite to use "that" when talking about a person, what do you think?

```suggestion
    /// The user who invited us to the room, if any.
```

crates/matrix-sdk/src/client/mod.rs
```diff
@@ -1429,11 +1430,42 @@ impl Client {
...
     /// in the DM event after creating the joined room.
-    async fn finish_join_room(&self, room_id: &RoomId) -> Result<Room> {
+    ///
+    /// If encrypted history sharing is enabled, will check to see if we have a
+    /// key bundle, and import it if so.
+    ///
+    /// # Arguments
+    ///
+    /// * `room_id` - The `RoomId` of the room that was joined.
+    /// * `pre_join_room_info` - Information about the room before we joined.
```
bnjbvr, line 1460: Do you find that it brings extra usefulness, here? I don't, but would be open to keeping this if you find it useful.

richvdh: well to be honest no, it's not useful in its own right. You can make the same argument for the docs on `join_room_by_id`, though.

Generally I favour adding documentation even where it's "obvious", because (a) it saves having to think about the decision about where the line is between "obvious" and not, and (b) it means we have a better starting point for documentation as the api evolves: for example in this case, if we add a third argument which *does* need documentation, that's easier because the documentation is already there for the other two.

Personally then, I think having the arguments listed is neutral-to-marginally-positive. But not a strong opinion.

crates/matrix-sdk/src/client/mod.rs
richvdh: I'd quite like to make that a separate problem, rather than scope-creeping the current fix into fixing other things.

crates/matrix-sdk/src/client/mod.rs
richvdh: 77eec1b73

crates/matrix-sdk/src/client/mod.rs
bnjbvr: Yeah, I'm generally against this list of arguments, unless they bring extra value (and agreed it's not any useful in `join_room_by_id` either). I've seen *so* many instances where those argument lists were desynchronized from the actual code, or just repeating the type with more words, that I find that we should avoid them, unless they provide some real value. But it's fine to keep as well, no strong opinions here, once again.

richvdh: > I've seen so many instances where those argument lists were desynchronized from the actual code,

Counterpoint: having the argument list in the documentation as a matter of course means that you start to remember that you can't change arguments without changing documentation. If they're only listed occasionally when they bring extra value, then the chances are much higher that people forget to maintain them.

richvdh: Since nobody has strong opinions here, I'm just going to leave it.

richvdh:
> is it worth adding a new test, or is the test change you've made sufficient?

It's difficult to think of what a new test would look like, other than an exact repeat of the test I've changed.

We talked about this on the crypto team, and concluded that the fundamental problem here was that there is no functional test at the application level to check that the functionality actually works (i.e. that the app is holding matrix-sdk in the way that matrix-sdk expects to be held). But, testing at that level brings its own set of problems.

crates/matrix-sdk/src/client/mod.rs
poljar: I like the argument list, if they don't contain useful information to me I can ignore them. If they are missing I can't manifest them back.

I'd like to see them everywhere in the public API as well.

richvdh:
Ignoring the coverage metric because it seems to be bogus

poljar:
One thing that comes to mind about the API here is that you can join a room using the `join_room_by_id` method without ever having synced or having a sync running.

This means that we might not have the bundle available if people just don't sync. I guess that this isn't much different from the bundle arriving late.

Still, I wish that the clients using this over the FFI were more normal and would just use the correct API. :disappointed:

PR #5288, opened 2025-06-26
Add a stream that notifies us about received room key bundles

poljar:
Part of https://github.com/matrix-org/matrix-rust-sdk/issues/4926.

This PR doesn't yet handle room key bundles received out of order. It is rather the building block which will notify us that we need to consider that a bundle might have arrived out of order.

### DIFF
--- a/crates/matrix-sdk-crypto/src/types/events/room_key_bundle.rs
+++ b/crates/matrix-sdk-crypto/src/types/events/room_key_bundle.rs
@@ -17,10 +17,12 @@
 //!
 //! [MSC4268]: https://github.com/matrix-org/matrix-spec-proposals/pull/4268
 
-use ruma::OwnedRoomId;
+use ruma::{events::room::EncryptedFile, OwnedRoomId};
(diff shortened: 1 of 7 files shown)


crates/matrix-sdk-crypto/src/types/events/room_key_bundle.rs
```diff
@@ -17,11 +17,13 @@
 //!
 //! [MSC4268]: https://github.com/matrix-org/matrix-spec-proposals/pull/4268
 
-use ruma::OwnedRoomId;
+use ruma::{events::room::EncryptedFile, OwnedRoomId};
 use serde::{Deserialize, Serialize};
 
 use super::EventType;
 
+// TODO: We need implement zeroize for this type.
```
Hywan, line 25: Is it possible to open an issue to track this? It seems rather important isn't it?

poljar: https://github.com/ruma/ruma/issues/2117

PR #5291, opened 2025-06-26
refactor(common): don't parse the full bundled latest thread event, but only its type

bnjbvr:
The previous code would parse the entire bundled event, only to look at its type indirectly. We can do better than this, and only look at the type instead. This brings a few benefits:

- it is faster, as we don't have to deserialize the entire event
- while the spec seems to indicate that the latest thread event has a `room_id`, it seems that, under some circumstances, it does not, as indicated by some rageshakes. As such, not parsing as `AnyMessageLike` (which mandates a room id) makes it more robust to the absence of a room id in there, marking more events as latest events.

Part of #4869.


andybalaam (approved):
Looks good, thanks!

bnjbvr:
(Pushed some tiny changes, so as to include a test; the previous test wasn't checking the bundled latest thread event was indeed included, and the bundled event was malformed so it would've failed in the first place!)

Issue #5285, opened 2025-06-25
Removing an avatar does not update the cached value.


Velin92:
Original issue: https://github.com/element-hq/element-x-ios/issues/4117

To explain the issue I'll add two sync responses, one for when the avatar is updated with a new avatar URL, and another one when the avatar is updated with a null value.

## Updated the room avatar with a new avatar
```
{
  "pos": "92278017/s6020200519_757284974_6174708_3755728849_4242429717_265999676_1501755517_11151313124_0_501813",
  "lists": {
    "all_rooms": {
      "count": 1,
      "ops": [
        {
          "op": "SYNC",
          "range": [
            0,
            0
          ],
          "room_ids": [
            "!yaCGbiOMYrDStgWeJr:matrix.org"
          ]
        }
      ]
    }
  },
  "rooms": {
    "!yaCGbiOMYrDStgWeJr:matrix.org": {
      "notification_count": 0,
      "highlight_count": 0,
      "avatar": "mxc://matrix.org/vIjDaVKBEbwZnqWgGCUICiFZ",
      "timeline": [
        {
          "content": {
            "info": {
              "mimetype": "image/jpeg"
            },
            "url": "mxc://matrix.org/vIjDaVKBEbwZnqWgGCUICiFZ"
          },
          "origin_server_ts": 1750870056555,
          "sender": "@mauro-fresh-test:matrix.org",
          "state_key": "",
          "type": "m.room.avatar",
          "unsigned": {
            "replaces_state": "$ZhOcRvDWYId3ezMib5reiI3UhIDcdElU74ke3snQF0o",
            "prev_content": {
              "info": {
                "mimetype": "image/jpeg"
              },
              "url": "mxc://matrix.org/LyxCQrNKUopMetbFJnAMbuwE"
            },
            "prev_sender": "@mauro-fresh-test:matrix.org",
            "membership": "join",
            "age": 253
          },
          "event_id": "$h9cbzBomcAA48mip7UhKfgU7f6EXIrL0nMuYfFAfOG0"
        }
      ],
      "limited": false,
      "prev_batch": "s6020200518_757284974_6174708_3755728849_4242429717_265999676_1501755517_11151313124_0_501813",
      "num_live": 1
    }
  },
  "extensions": {
    "account_data": {
      "global": [],
      "rooms": {}
    },
    "receipts": {
      "rooms": {}
    },
    "typing": {
      "rooms": {}
    }
  }
} 
```
[...]

PR #5293, opened 2025-06-26
fix: Add `m.room.avatar` to the required state

jmartinesp:
This works around the issue with removed avatars not being explicitly flagged by sliding sync until https://github.com/element-hq/synapse/issues/18598 is fixed.

This is a workaround for https://github.com/matrix-org/matrix-rust-sdk/issues/5285.

Signed-off-by:


jmartinesp:
> I smell incoming test failures, but LGTM once they're addressed!

Agh, I forgot to run `nextest`. Let's see how many need to be fixed 🤞 .

PR #5296, opened 2025-06-26
refactor(sdk): introduce the new `paginators` module

bnjbvr:
There was a room paginator implementation in the event cache, but unused by the event cache. There also was a thread paginator implementation in the timeline. I've moved both to a new mod in the SDK, called `paginators`. The rest is mostly uninteresting details. Can be reviewed commit by commit.

Part of #4869.

### DIFF
--- a/crates/matrix-sdk-ui/src/timeline/tests/mod.rs
+++ b/crates/matrix-sdk-ui/src/timeline/tests/mod.rs
@@ -30,7 +30,7 @@ use matrix_sdk::{
     config::RequestConfig,
     crypto::OlmMachine,
     deserialized_responses::{EncryptionInfo, TimelineEvent},
-    paginators::{PaginableRoom, PaginatorError},
+    paginators::{thread::PaginableThread, PaginableRoom, PaginatorError},
     room::{EventWithContextResponse, Messages, MessagesOptions, PushContext, Relations},
     send_queue::RoomSendQueueUpdate,
     BoxFuture,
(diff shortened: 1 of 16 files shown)


crates/matrix-sdk-ui/src/timeline/tests/mod.rs
```diff
@@ -303,6 +303,23 @@ impl PaginableRoom for TestRoomDataProvider {
     }
 }
 
+impl PaginableThread for TestRoomDataProvider {
```
stefanceriu, line 306: I wasn't sure whether this needed it's own trait when first building this as relations aren't thread specific.

crates/matrix-sdk/src/paginators/thread.rs
```diff
@@ -12,28 +12,75 @@
...
     /// Create a new [`Paginator`], given a room implementation.
     pub fn new(room: P, root_event_id: OwnedEventId) -> Self {
         Self { room, root_event_id, token: Mutex::new(None.into()) }
     }
 
+    /// Run a single pagination backwards, returning the next set of events and
+    /// information whether we've reached the start of the thread.
+    ///
+    /// Note: when the thread start is reached, the root event *will* be
+    /// included in the result.
```
stefanceriu, line 83: Noice 👌

stefanceriu (approved):
Nice cleanup, lgtm 👍

crates/matrix-sdk-ui/src/timeline/tests/mod.rs
bnjbvr: Yep, the trait ought to make it easier to test in isolation, which we might have to do, since I won't be using this loader anymore in a soon future :see_no_evil:

Issue #4112, opened 2024-10-10
[meta] Latest Event API, the missing one


bnjbvr:
The Latest Event represents the event that is displayed in the Room List under each room item, or displayed as a Thread Summary. It is important, and must be reactive. It follows several constraints, such as:

- Not all events are a candidate to be a Latest Event, e.g. state events aren't candidate; thus the Latest Event has to paginate through events in reverse order (from newest to oldest) to find a suitable event if any, otherwise the information that no candidate has been found must be present,
- It must represent remote _and_ local events for a better user experience.

The Latest Event API in its current form was added a bit in an ad-hoc way, since we needed the feature but lacked the proper storage layer for events, and the proper send queue. What we can re-use though is all the logic to find a suitable event, which was not trivial to write in an efficient way knowing we have to deal with encrypted event and so on.

Today is different: We have the Event Cache API, and we have the Send Queue API. Those are the two components we need to write a proper Latest Event API! I love it when a plan comes together.

The Latest Event API in its current form is scattered in multiple crates. It's going to be hard to “migrate”, or move things together. The plan we came up with the team is to write a new API from scratch, and then migrate the components using a Latest Event one by one.

> [!IMPORTANT]
>
> ### Storage
>
> A note about the storage. The Event Cache API has its own storage. The Send Queue API has its own storage. The Latest Event API will be _computation only_, no storage is required, and that will solve multiple problems. Right now, the Latest Event is stored in `RoomInfo`, which is not a good place.

> [!IMPORTANT]
>
> ### Thread
>
> Another note about threads. The Latest Event API in its current form is only used by the Rooms. But we need it to work with Threads too, as it is required for Thread Summary.
>
> See https://github.com/matrix-org/matrix-rust-sdk/issues/6013

## Product decisions/questions

- If the latest event is an edit, show it if it edited the previous latest rendered event.
- If the latest event has been redacted, do we show "a message has been redacted", or the previous preview event?
- What counts as an interesting latest preview event?
  - We already have made some decisions there, but what about reactions?
- If there's a local echo that hasn't been sent yet (and it should've been sent, say, 8 hours ago), but there's been new messages in the room since then, what do we show: the local echo or the most up to date message?
- If there's a new in-thread message in a room, can it become the latest preview event for that room, or are only main-thread messages candidates for being the latest preview event?
  - Product proposes that only the main-timeline messages are candidates for the latest preview event for a room.

### Create the Latest Event API

[...]

PR #5298, opened 2025-06-27
feat(sdk): Introduce the new `LatestEvents` API, part 1

Hywan:
This patch is the first part of the new `LatestEvents` API. It contains
the “framework”, i.e. the structure, tasks, logic to make it work, but
no latest events are computed for the moment.

The Latest Events API provides a lazy, reactive and efficient way to
compute the latest event for a room or a thread.

The latest event represents the last displayable and relevant event
a room or a thread has been received. It is usually displayed in a
_summary_, e.g. below the room title in a room list.

The entry point is `LatestEvents`. It is preferable to get a reference
to it from `Client::latest_events`, which already plugs everything to
build it. `LatestEvents` is using the `EventCache` and the `SendQueue`
to respectively get known remote events (i.e. synced from the server),
or local events (i.e. ones being sent).

## Laziness

`LatestEvents` is lazy, it means that, despite `LatestEvents` listening 
to all `EventCache` or `SendQueue` updates, it will
only do something if one is expected to get the latest event for a
particular room or a particular thread. Concretely, it means that
until `LatestEvents::listen_to_room` is called for a particular room, no
latest event will ever be computed for that room (and similarly with
`LatestEvents::listen_to_thread`).

If one is no longer interested to get the latest event for a
particular room or thread, the `LatestEvents::forget_room` and
`LatestEvents::forget_thread` methods must be used.

## Reactive

`LatestEvents` is designed to be reactive.
`LatestEvents::listen_and_subscribe_to_room` must be used, it
returns a `Subscriber`.

---


crates/matrix-sdk/src/latest_events/subscriber.rs
stefanceriu, line 1: I imagine this isn't empty by mistake, right?

crates/matrix-sdk/src/latest_events/mod.rs
```diff
@@ -0,0 +1,861 @@
...
+//! it from [`Client::latest_events`][crate::Client::latest_events], which
+//! already plugs everything to build it. [`LatestEvents`] is using the
+//! [`EventCache`] and the [`SendQueue`] to respectively get known remote events
+//! (i.e. synced from the server), or local events (i.e. ones being sent).
+//!
+//! ## Laziness
+//!
+//! [`LatestEvents`] is lazy, it means that, despites [`LatestEvents`] is
+//! listening to all [`EventCache`] or [`SendQueue`] updates, it will only do
+//! something if one is expected to get the latest event for a particular room
```
stefanceriu:
```suggestion
//! [`LatestEvents`] is lazy, it means that, despite [`LatestEvents`]
//! listening to all [`EventCache`] or [`SendQueue`] updates, it will only do
//! something if one is expected to get the latest event for a particular room
```

crates/matrix-sdk/src/latest_events/mod.rs
```diff
@@ -0,0 +1,861 @@
...
+//! or a particular thread. Concretely, it means that until
+//! [`LatestEvents::for_room`] is called for a particular room, no latest event
+//! will ever be computed for that room (and similarly with
```
stefanceriu: Does that mean I won't have a latest event if I subscribe to it later, after messages come in?

crates/matrix-sdk/src/latest_events/latest_event.rs
```diff
@@ -0,0 +1,55 @@
...
+    pub(super) fn new(room_id: &RoomId, thread_id: Option<&EventId>) -> Option<Self> {
+        Some(Self {
+            _room_id: room_id.to_owned(),
+            _thread_id: thread_id.map(ToOwned::to_owned),
+            value: SharedObservable::new_async(LatestEventValue::None),
+        })
+    }
+
+    /// Return the current value for the latest event.
+    pub async fn get(&self) -> LatestEventValue {
```
stefanceriu: Should this be called plainly `value` maybe? Or perhaps even clearer `current_value` as a counterpart to `subscribe`.

`LatestEvent.value` feels more comfortable than `LatestEvent.get` to me ¯\_(ツ)_/¯

crates/matrix-sdk/src/latest_events/mod.rs
```diff
@@ -0,0 +1,861 @@
...
+    /// Start listening to updates (if not already) for a particular room, and
+    /// fetch the [`LatestEvent`] for this room.
+    ///
+    /// It returns `None` if the room doesn't exist.
+    ///
+    /// Note that a read lock guard is returned if the [`LatestEvent`] exists.
+    /// One doesn't want to keep this lock alive for too long, otherwise the
+    /// whole [`LatestEvents`] API is blocked. Use [`LatestEvent::get`] or
+    /// [`LatestEvent::subscribe`] as quick as possible, and then drop this
+    /// lock guard.
```
stefanceriu: This sounds a bit footgun-y, do we really need to expose the internal lock guard or can we maybe instead expose the underlying `get/value` and `subscribe` methods directly (similar to how `forget_thread` just makes magic happen)? e.g. subscribe & get room + subscribe & get thread

crates/matrix-sdk/src/latest_events/mod.rs
```diff
@@ -0,0 +1,861 @@
...
+    ///
+    /// The [`RoomLatestEvents`], and the associated [`LatestEvent`], will be
+    /// created if missing. It means that write lock is taken if necessary, but
+    /// it's always downgraded to a read lock at the end.
+    async fn room_latest_event(
+        &self,
+        room_id: &RoomId,
+        thread_id: Option<&EventId>,
+    ) -> Option<RwLockReadGuard<'_, RoomLatestEvents>> {
+        match thread_id {
```
stefanceriu, line 224: It looks like a lot of this code is duplicated between the 2 match branches. Can we if let the thread in the second step and reduce the indentation here?

crates/matrix-sdk/src/latest_events/mod.rs
```diff
@@ -0,0 +1,861 @@
...
+        self.state.registered_rooms.for_thread(room_id, thread_id).await
+    }
+
+    /// Forget a room.
+    ///
+    /// It means that [`LatestEvents`] will stop listening to updates for the
+    /// `LatestEvent`s of the room and all its threads.
+    ///
+    /// If [`LatestEvents`] is not listening for `room_id`, nothing happens.
+    pub async fn forget_room(&self, room_id: &RoomId) {
```
stefanceriu, line 175: Can't help but wonder, is there any version of this in which the user doesn't need to care about deregestering from subscribers manually? If I get a `LatestEvent` and subscribe to its values, can dropping that take care of the forgetting  maybe?

stefanceriu (approved):
Nice start! 👏 Left some comments inline

crates/matrix-sdk/src/latest_events/mod.rs
Hywan: If you want the `LatestEvent`, you will need to `LatestEvents::for_room(…).get()`. We will try our best to compute it, including the initial value for it. If you call `LatestEvent::for_room(…).subscribe()`, you will get a `Stream` of updates. If just want to register this room, simply call `LatestEvent::for_room();` and that's it. The system will listen to this room. Later, each time a new `LatestEvent` will be computed, a `RoomInfoNotableUpdate` will be triggered, which will update the `RoomListService` for example. Is it clear?

crates/matrix-sdk/src/latest_events/latest_event.rs
Hywan: We are mimicing the API of `eyeball::Subscriber` and I think it's great for the consistency (we are also using this in other parts of the SDK). Happy to revisit that if needed.

crates/matrix-sdk/src/latest_events/mod.rs
Hywan: Sadly no, because in one case we take a write-lock immediately, another time, we take a read-lock then a write-lock, it's a bit different.

crates/matrix-sdk/src/latest_events/mod.rs
Hywan: I've replaced `for_room` by `listen_to_room(…) -> bool` and `listen_and_subscribe_to_room(…) -> Option<Subscriber<LatestEventValue>>`. Same for thread.

crates/matrix-sdk/src/latest_events/latest_event.rs
Hywan: See https://github.com/matrix-org/matrix-rust-sdk/pull/5298#discussion_r2176754481

crates/matrix-sdk/src/latest_events/mod.rs
Hywan: See https://github.com/matrix-org/matrix-rust-sdk/pull/5298#discussion_r2176754481

crates/matrix-sdk/src/latest_events/mod.rs
Hywan: A `LatestEvent` is no longer returned, so all good :-p.

PR #5301, opened 2025-06-27
Fix new lints in nightly

zecakeh:
See each commit message.


zecakeh:
Should I also update the version of nightly in CI?

poljar:
> Should I also update the version of nightly in CI?

I wouldn't mind that. Though we can do that in a separate PR.

poljar (approved):

PR #5305, opened 2025-06-28
ci: Bump the version of Rust nightly

zecakeh:


poljar (approved):
Seems like nothing broke.

PR #5306, opened 2025-06-29
refactor(common): Use a constant for the room version to use as a fallback

zecakeh:
It avoids using different versions in several places for consistency. It also allows to be able to change it in a single place when needed.

This also bumps the fallback to v11 everywhere, since it is the default version for new rooms since Matrix 1.14 and it has the sanest redaction rules.


jplatte (approved):

poljar (approved):
Very nice, thank you.

PR #5309, opened 2025-06-29
Upgrade matrix-sdk-ui to Rust edition 2024

jplatte:
I made sure that the two remaining PRs that blocked the large rustfmt PR (https://github.com/matrix-org/matrix-rust-sdk/pull/5225, https://github.com/matrix-org/matrix-rust-sdk/pull/5258) do not touch the UI crate, so this is not going to result in conflicts there.

There were a few more `+ use<_>` clauses added in private interfaces, but they weren't necessary for compilation to succeed, so I omitted them. I doubt they will become required for anything later, but if they do, it's gonna be the same sort of situation as newly-added `-> impl Trait` return types, where surely nobody thinks to add `+ use<>`, and the compiler is hopefully going to suggest doing that where helpful.

Signed-off-by: Jonas Platte

### DIFF
--- a/crates/matrix-sdk-ui/Cargo.toml
+++ b/crates/matrix-sdk-ui/Cargo.toml
@@ -2,7 +2,7 @@
 name = "matrix-sdk-ui"
 description = "GUI-centric utilities on top of matrix-rust-sdk (experimental)."
 version = "0.13.0"
-edition = "2021"
+edition = "2024"
 repository = "https://github.com/matrix-org/matrix-rust-sdk"
 license = "Apache-2.0"
 rust-version.workspace = true
--- a/crates/matrix-sdk-ui/src/room_list_service/filters/fuzzy_match_room_name.rs
+++ b/crates/matrix-sdk-ui/src/room_list_service/filters/fuzzy_match_room_name.rs
@@ -12,9 +12,9 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-pub use fuzzy_matcher::{skim::SkimMatcherV2, FuzzyMatcher as _};
+pub use fuzzy_matcher::{FuzzyMatcher as _, skim::SkimMatcherV2};
 
-use super::{normalize_string, Filter};
+use super::{Filter, normalize_string};
 
 struct FuzzyMatcher {
     matcher: SkimMatcherV2,
--- a/crates/matrix-sdk-ui/src/room_list_service/filters/fuzzy_match_room_name.rs
+++ b/crates/matrix-sdk-ui/src/room_list_service/filters/fuzzy_match_room_name.rs
@@ -44,7 +44,7 @@ impl FuzzyMatcher {
 ///
 /// Rooms are fetched from the `Client`. The pattern and the room names are
 /// normalized with `normalize_string`.
-pub fn new_filter(pattern: &str) -> impl Filter {
+pub fn new_filter(pattern: &str) -> impl Filter + use<> {
     let searcher = FuzzyMatcher::new().with_pattern(pattern);
 
     move |room| -> bool {
--- a/crates/matrix-sdk-ui/src/timeline/controller/aggregations.rs
+++ b/crates/matrix-sdk-ui/src/timeline/controller/aggregations.rs
@@ -42,17 +42,17 @@ use std::{borrow::Cow, collections::HashMap, sync::Arc};
 use as_variant::as_variant;
 use matrix_sdk::deserialized_responses::EncryptionInfo;
 use ruma::{
+    MilliSecondsSinceUnixEpoch, OwnedEventId, OwnedTransactionId, OwnedUserId, RoomVersionId,
     events::{
+        AnySyncTimelineEvent,
         poll::unstable_start::NewUnstablePollStartEventContentWithoutRelation,
         relation::Replacement, room::message::RoomMessageEventContentWithoutRelation,
-        AnySyncTimelineEvent,
     },
     serde::Raw,
-    MilliSecondsSinceUnixEpoch, OwnedEventId, OwnedTransactionId, OwnedUserId, RoomVersionId,
 };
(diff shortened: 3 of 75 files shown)


crates/matrix-sdk-ui/Cargo.toml
```diff
@@ -2,7 +2,7 @@
 name = "matrix-sdk-ui"
 description = "GUI-centric utilities on top of matrix-rust-sdk (experimental)."
 version = "0.12.0"
-edition = "2021"
+edition = "2024"
```
stefanceriu, line 5: 🥳

crates/matrix-sdk-ui/src/room_list_service/filters/fuzzy_match_room_name.rs
```diff
@@ -44,7 +44,7 @@ impl FuzzyMatcher {
 ///
 /// Rooms are fetched from the `Client`. The pattern and the room names are
 /// normalized with `normalize_string`.
-pub fn new_filter(pattern: &str) -> impl Filter {
+pub fn new_filter(pattern: &str) -> impl Filter + use<> {
```
stefanceriu, line 47: TIL https://blog.rust-lang.org/2024/09/05/impl-trait-capture-rules/

crates/matrix-sdk-ui/src/timeline/controller/aggregations.rs
```diff
@@ -642,11 +642,7 @@ fn resolve_edits(
         }
     }
 
-    if let Some(edit) = best_edit {
-        edit_item(event, edit)
-    } else {
-        false
-    }
+    if let Some(edit) = best_edit { edit_item(event, edit) } else { false }
```
stefanceriu, line 645: Ending a function in `if let` feels super strange to me, I think I liked the previous version better.

crates/matrix-sdk-ui/src/timeline/tests/reactions.rs
```diff
@@ -144,13 +144,15 @@ async fn test_redact_reaction_success() {
...
     let event = assert_item_update!(stream, &event_id, item_pos);
-    assert!(event
-        .content()
-        .reactions()
-        .cloned()
-        .unwrap_or_default()
-        .get(&REACTION_KEY.to_owned())
-        .is_none());
+    assert!(
+        event
```
stefanceriu, line 148: Got mixed feelings about this one too.

stefanceriu (commented):
Apart from what I commented this looks good to me, thank you Jonas!

I don't think I should be the only one reviewing it though so I'll let the review gods reassign.

crates/matrix-sdk-ui/src/room_list_service/filters/fuzzy_match_room_name.rs
Hywan: I find it particularly frustrating to add empty `use`… I don't understand why it's not the default.

crates/matrix-sdk-ui/src/timeline/controller/mod.rs
```diff
@@ -586,7 +586,7 @@ impl<P: RoomDataProvider, D: Decryptor> TimelineController<P, D> {
         &self,
     ) -> (
         Vector<Arc<TimelineItem>>,
-        impl Stream<Item = VectorDiff<Arc<TimelineItem>>> + SendOutsideWasm,
+        impl Stream<Item = VectorDiff<Arc<TimelineItem>>> + SendOutsideWasm + use<P, D>,
```
Hywan: Can't we change the `impl Stream` to `VectorSubscriberStream`? Having a named type would help removing the `use<P, D>` I think, thought?

crates/matrix-sdk-ui/src/timeline/controller/mod.rs
```diff
@@ -602,7 +602,7 @@ impl<P: RoomDataProvider, D: Decryptor> TimelineController<P, D> {
     pub(super) async fn subscribe_filter_map<U, F>(
         &self,
         f: F,
-    ) -> (Vector<U>, impl Stream<Item = VectorDiff<U>>)
+    ) -> (Vector<U>, impl Stream<Item = VectorDiff<U>> + use<U, F, P, D>)
```
Hywan: Same here. I believe `impl + use` creates more confusion that using a named type. In case there isn't a named type, I think it's great, but since we've one, can't we use it?

crates/matrix-sdk-ui/src/timeline/pagination.rs
```diff
@@ -103,7 +103,7 @@ impl super::Timeline {
     /// call to [`Self::paginate_backwards()`].
     pub async fn live_back_pagination_status(
         &self,
-    ) -> Option<(RoomPaginationStatus, impl Stream<Item = RoomPaginationStatus>)> {
+    ) -> Option<(RoomPaginationStatus, impl Stream<Item = RoomPaginationStatus> + use<>)> {
```
Hywan, line 110: Here it's fine because there is no named type we can use: `impl Stream` is mandatory.

crates/matrix-sdk-ui/src/timeline/controller/aggregations.rs
bnjbvr: It's the same, but formatted differently, right?

stefanceriu: Yep, just one lined.

crates/matrix-sdk-ui/src/room_list_service/filters/fuzzy_match_room_name.rs
jplatte: Probably because it's backwards-compatible to reduce captured generics (including down to nothing, i.e. `use<>`), but capturing _more_ generics is a breaking change.

PR #5312, opened 2025-06-30
refactor: Don't use `AnyMessageLikeEventContent` with the event factory

zecakeh:
When we upgrade Ruma, the `EventContent` bound on `EventBuilder` will be changed to `StaticEventContent`, which is not implemented by the `Any*EventContent` enums.

### DIFF
--- a/crates/matrix-sdk-ui/src/timeline/tests/polls.rs
+++ b/crates/matrix-sdk-ui/src/timeline/tests/polls.rs
@@ -3,16 +3,9 @@ use fakes::poll_a2;
 use matrix_sdk_test::{async_test, ALICE, BOB};
 use ruma::{
     event_id,
-    events::{
-        poll::{
-            unstable_end::UnstablePollEndEventContent,
-            unstable_response::UnstablePollResponseEventContent,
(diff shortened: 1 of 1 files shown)


crates/matrix-sdk-ui/src/timeline/tests/polls.rs
```diff
@@ -297,9 +280,9 @@ impl TestTimeline {
         original_id: &EventId,
         content: UnstablePollStartContentBlock,
     ) {
-        let content =
-            ReplacementUnstablePollStartEventContent::new(content, original_id.to_owned());
-        let event_content = AnyMessageLikeEventContent::UnstablePollStart(content.into());
+        let event_content = UnstablePollStartEventContent::from(
+            ReplacementUnstablePollStartEventContent::new(content, original_id.to_owned()),
+        );
```
bnjbvr, line 285: Note to self: There's also a `EventFactory::poll_edit` function, but it can't easily be used without also changing the signature (thus the callers) here.

bnjbvr (approved):
Cheers!

PR #5315, opened 2025-06-30
refactor(timeline): simplify the concept of "focus" and refactor building

bnjbvr:
There were three different APIs for a `TimelineFocus`, before this PR:

- `TimelineFocus`, which is the public API to be used when building a `Timeline`
- `TimelineFocusKind`, which is a simplified view of the focus of a timeline, used in misc places to make some lightweight decisions
- `TimelineFocusData`, which includes some (inner) mutable data to make useful changes to a timeline: pagination, etc.

This PR unifies the latter two into a single type, that's eventually renamed `TimelineFocusKind`. I find it makes it simpler to understand the different choices, and I even think this might lead to more simplifications later.

### DIFF
--- a/crates/matrix-sdk-ui/src/timeline/controller/mod.rs
+++ b/crates/matrix-sdk-ui/src/timeline/controller/mod.rs
@@ -285,6 +285,7 @@ impl<P: RoomDataProvider, D: Decryptor> TimelineController<P, D> {
             TimelineFocus::Live { hide_threaded_events } => {
                 TimelineFocusKind::Live { hide_threaded_events }
             }
+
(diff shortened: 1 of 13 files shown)


crates/matrix-sdk-ui/src/timeline/controller/mod.rs
```diff
@@ -186,14 +186,6 @@ impl Default for TimelineSettings {
     }
 }
 
-#[derive(Debug, Clone)]
-pub(super) enum TimelineFocusKind {
```
stefanceriu, line 182: 🔥

PR #5319, opened 2025-07-01
Refuse to decrypt to-device messages from unverified devices (when in exclude insecure mode)

andybalaam:
First half of https://github.com/matrix-org/matrix-rust-sdk/issues/4147 - this is the receiving part.


crates/matrix-sdk-crypto/README.md
```diff
@@ -33,19 +32,27 @@ async fn main() -> Result<(), OlmError> {
...
+            },
+            &decryption_settings,
+        )
+        .await?;
 
     // Pull requests that we need to send out.
     let outgoing_requests = machine.outgoing_requests().await?;
 
-    // Send the requests here out and call machine.mark_request_as_sent().
+    // Send the requests out here and call machine.mark_request_as_sent().
```
poljar, line 55: Hmm, this sounds off to me. Perhaps removing `here` sounds the best.

crates/matrix-sdk-crypto/src/machine/mod.rs
```diff
@@ -868,27 +868,13 @@ impl OlmMachine {
...
-        } else {
-            // Device is not dehydrated: handle it as normal e.g. create a Megolm session
-            self.handle_decrypted_to_device_event(transaction.cache(), &mut decrypted, changes)
-                .await?;
+        // Return early if the sending device is decrypted
+        self.check_to_device_event_is_not_from_dehydrated_device(&decrypted, &event.sender).await?;
 
-            Ok(decrypted)
-        }
+        // Device is not dehydrated: handle it as normal e.g. create a Megolm session
```
poljar, line 883: This is an existing comment but it's a bit misleading, we don't just create a `Megolm` session in `handle_decrypted_to_device_event()`. It depends on the received event type what will happen, if anything.

crates/matrix-sdk-crypto/src/machine/mod.rs
```diff
@@ -868,27 +868,13 @@ impl OlmMachine {
...
-                session = ?decrypted.session,
-                "Received a to-device event from a dehydrated device. This is unexpected: ignoring event"
-            );
-            Err(DecryptToDeviceError::FromDehydratedDevice)
-        } else {
-            // Device is not dehydrated: handle it as normal e.g. create a Megolm session
-            self.handle_decrypted_to_device_event(transaction.cache(), &mut decrypted, changes)
-                .await?;
+        // Return early if the sending device is decrypted
+        self.check_to_device_event_is_not_from_dehydrated_device(&decrypted, &event.sender).await?;
```
poljar, line 881: It's kinda tripping me up that we flattened out the branches here. A bit worried that we'll somehow miss this subtle `?` that prevents us from accepting secrets from dehydrated devices.

That being said, I don't have an idea how to improve it.

crates/matrix-sdk-common/src/deserialized_responses.rs
```diff
@@ -1196,6 +1196,15 @@ pub enum ProcessedToDeviceEvent {
     /// required information to be processed (like no event `type` for
     /// example)
     Invalid(Raw<AnyToDeviceEvent>),
+
+    /// A to device event that was ignored because the sender device was not
```
poljar:
```suggestion
    /// A to-device event that was ignored because the sender device was not
```

crates/matrix-sdk-crypto/src/machine/mod.rs
```diff
@@ -862,32 +870,24 @@ impl OlmMachine {
...
-                sender = ?event.sender,
-                session = ?decrypted.session,
-                "Received a to-device event from a dehydrated device. This is unexpected: ignoring event"
-            );
-            Err(DecryptToDeviceError::FromDehydratedDevice)
-        } else {
-            // Device is not dehydrated: handle it as normal e.g. create a Megolm session
-            self.handle_decrypted_to_device_event(transaction.cache(), &mut decrypted, changes)
-                .await?;
+        // Return early if the sending device is decrypted
```
poljar:
```suggestion
        // Return early if the sending device is a dehydrated device
```

?

crates/matrix-sdk-crypto/CHANGELOG.md
```diff
@@ -11,6 +11,21 @@ All notable changes to this project will be documented in this file.
 - [**breaking**] Add a new `VerificationLevel::MismatchedSender` to indicate that the sender of an event appears to have been tampered with.
   ([#5219](https://github.com/matrix-org/matrix-rust-sdk/pull/5219))
 
+- [**breaking**]: When in "exclude insecure devices" mode, refuse to decrypt
+  incoming to-device messages from unverified devices, except for some
+  exceptions for certain event types (see the documentation on
+  `OlmMachine::decrypt_to_device_event` for details). To support this, a new
```
poljar: `OlmMachine::decrypt_to_device_event` is a private method, so it's gonna be a bit hard for users to look at the docs of that method.

crates/matrix-sdk-common/src/deserialized_responses.rs
```diff
@@ -1196,6 +1196,15 @@ pub enum ProcessedToDeviceEvent {
...
+    /// verified.
+    UnverifiedSender {
+        /// The raw decrypted event
+        raw: Raw<AnyToDeviceEvent>,
+        /// The Olm encryption info
+        encryption_info: EncryptionInfo,
+    },
```
poljar: Hmm, but is this something we want? In the case of Megolm we just give you the UTD.

I would have expected us to do the same here, though I'm happy to be convinced that we need to do something different for Olm/to-device messages.

If we want to explain to people why this is an UTD we could expand the `UnableToDecrypt` variant.

If you agree with this, perhaps we should move all this logic into the `Account` somewhere in the `parse_decrypted_to_device_event()`. Not sure if we should also move the device dehydration check there.

andybalaam: It just felt rude to decrypt a message and then throw away all we know about it, so it would be hard to debug why it failed to decrypt, so I wanted to include the decrypted event (mainly so you could see the real event type) and encryption info.

If you like we could add to UTD like:

```
struct DecryptedCandidate {
    event: Raw<AnyToDeviceEvent>,
    encryption_info: EncryptionInfo,
}

...
pub enum ProcessedToDeviceEvent {
    UnableToDecrypt{
        encrypted_event: Raw<AnyToDeviceEvent>,
        decrypted_candidate: Option<SomeNewType>
    }
...
```

But I'd also be happy to just return the existing UTD if you think that's sensible.

> If you agree with this, perhaps we should move all this logic into the Account somewhere in the parse_decrypted_to_device_event(). Not sure if we should also move the device dehydration check there.

Happy to do that if you prefer.

crates/matrix-sdk-crypto/CHANGELOG.md
andybalaam: Yeah. Do you think it's worth repeating ourselves by listing the exceptions in a public method doc? I looked and it felt a bit over-detailed. Maybe I should just remove this part from the release notes.

crates/matrix-sdk-crypto/README.md
andybalaam: It's saying "insert blah blah code here", so I thought using "here" made sense, but happy to change if you feel strongly.

crates/matrix-sdk-crypto/src/machine/mod.rs
andybalaam: Yeah, I don't love it, but I think splitting into a separate function is good.

I could make it an explicit `if` and `return` but as the method name says "check" and we have a comment saying "Return early" maybe it's good enough?

crates/matrix-sdk-crypto/src/machine/mod.rs
andybalaam: But "e.g." means "for example" so that seems OK to me, but happy to delete it.

poljar:
I see that you re-requested review, but what about my other comments? Most importantly about https://github.com/matrix-org/matrix-rust-sdk/pull/5319#discussion_r2179649604?

andybalaam:
> I see that you re-requested review, but what about my other comments? Most importantly about [#5319 (comment)](https://github.com/matrix-org/matrix-rust-sdk/pull/5319#discussion_r2179649604)?

I'm waiting for replies from you on several comments including that one.

poljar:
> > I see that you re-requested review, but what about my other comments? Most importantly about [#5319 (comment)](https://github.com/matrix-org/matrix-rust-sdk/pull/5319#discussion_r2179649604)?
> 
> I'm waiting for replies from you on several comments including that one.

Err, did you reply to any of my comments? Or do you expect me to reply to my own comments?

I can't see any comments after my initial comments:

image

Is there supposed to be something else there? I'm confused.

andybalaam:
> I can't see any comments after my initial comments:

Odd. I can:

image

andybalaam:
D'oh, it was a pending review, sorry. I think I've clicked the relevant button now.

crates/matrix-sdk-crypto/README.md
poljar: Not feeling strongly, feel free to leave as it is.

crates/matrix-sdk-crypto/src/machine/mod.rs
poljar: Yeah, I suppose you're right.

crates/matrix-sdk-crypto/CHANGELOG.md
poljar: I think removing it from the changelog is fine.

crates/matrix-sdk-common/src/deserialized_responses.rs
poljar: > It just felt rude to decrypt a message

Yeah but that's what we do for room messages as well

> and then throw away all we know about it, so it would be hard to debug why it failed to decrypt

We should probably return the error then. We can include more metadata i.e. the decrypted event type if you think that this is sensible.

> If you like we could add to UTD like:

Yes, this seems sensible. It's also mimicking what we do for room events more or less. That is we can include the UTD reason/info like we did for room events:

https://github.com/matrix-org/matrix-rust-sdk/blob/6de403276afeb9c316a762e0dfcad13c72fe6eea/crates/matrix-sdk-common/src/deserialized_responses.rs#L720-L729

> Happy to do that if you prefer.

Yeah, sounds like it would centralize this "refusing to decrypt" logic.

poljar (changes requested):
Hopefully I replied to all the outstanding comments. If I missed some, please ping me.

crates/matrix-sdk-crypto/src/machine/mod.rs
```diff
@@ -2816,6 +2939,28 @@ impl OlmMachine {
     }
 }
 
+fn sending_device_is_unverified(encryption_info: &EncryptionInfo) -> bool {
+    if let VerificationState::Unverified(level) = &encryption_info.verification_state {
```
BillCarsonFr: nit: I think that a full `match` with `Verified`/`Unverified` will be easier to read than the `if let` and `else`

Also the name `is_unverified` can be better? because now when `UnverifiedIdentity` is true then `is_unverified` is false.

Can we do something similar than for megolm? something around `check_sender_trust_requirement`?

crates/matrix-sdk-common/src/deserialized_responses.rs
BillCarsonFr: I was thinking about that in the context of element-call. 
EC uses to-device messages to send the key to decrypt the video stream. 
I initialy thought that it would be good for me to know that the incoming message was a call key even if we decided to not decrypt it because of trust settings. This could allow me to put some feedback on screen.

But actually I think I don't really need it. The call membership event has the sender device info in it, so I already know the verification state of the device, so I will know that it will be impossible to receive a key from it.

So the `refusing to decrypt` logic is ok for me.

If we really need some info maybe the `UnableToDecryptInfo` could hold the clear type? but not sure how useful it would be (appart for some metrics/analytics)

poljar: Yeah, I'm fine with adding the clear type to the info if necessary, just not the whole content.

andybalaam: Updated to make this a UTD in 56a51a49c3 . Working on moving the logic to Account.

andybalaam: Updated to move the logic to Account in e6748a0d15

andybalaam: > Yeah, I'm fine with adding the clear type to the info if necessary, just not the whole content.

I have not done this since we don't see a clear need but it should be simple to do in future.

crates/matrix-sdk-crypto/src/machine/mod.rs
andybalaam: Attempted to make it clearer in e5fac474d2

crates/matrix-sdk-crypto/CHANGELOG.md
andybalaam: Removed in 39240427b5

crates/matrix-sdk-crypto/src/machine/mod.rs
andybalaam: Thanks, improved with a `match` in 02ba63c832 and re-arranged to try and look more like `check_sender_trust_requirement` in 550fa11bb7 . What do you think?

andybalaam:
> Hopefully I replied to all the outstanding comments. If I missed some, please ping me.

I think I addressed everything - let me know if not. Thanks!

poljar (approved):
Alright, I think that this is good now.

There's an incomplete commit message in 550fa11bb7b1722851983ea77d2e9802654043f4.

Not sure if you want to clean up the history and the commit message or if you'd like to just squash merge or something else.

PR #5322, opened 2025-07-01
feat(sdk): Add a tasks that listens for historic room keys if they arrive out of order

poljar:
Historic room key bundles are uploaded as an encrypted file to the media
repo and the key to decrypt the file is sent as a to-device message to
the recipient device.

In the nominal case, the invite and this to-device message should arrive
at the same time and accepting the invite would download and import the
bundle.

If the to-device message arrives after the invite has already been
accepted we would never download and import the bundle.

To mitigate this problem, this patch introduces a task that listens for
bundles that arrive. If the bundle is for a room that we have joined we
will consider importing the bundle.

Please note that the acceptance rules aren't yet decided and this PR needs to be updated with some better acceptance rules before it gets merged.


crates/matrix-sdk/src/encryption/tasks.rs
```diff
@@ -385,15 +390,77 @@ impl BackupDownloadTaskListenerState {
...
+        // we'll have to implement the bundle event in Ruma.
+        while let Some(bundle_info) = stream.next().await {
+            let Some(client) = client.get() else {
+                // The client was dropped while we were waiting on the stream. Let's end the
+                // loop, since this means that the application has shut down.
+                break;
+            };
+
+            let Some(room) = client.get_room(&bundle_info.room_id) else {
+                warn!(room_id = %bundle_info.room_id, "Received a historic room key bundle for an unknown room");
```
andybalaam, line 423: Couldn't this happen if the bundle arrived before we saw the invitation?

crates/matrix-sdk/src/encryption/tasks.rs
```diff
@@ -385,15 +390,77 @@ impl BackupDownloadTaskListenerState {
...
+            };
+
+            Self::handle_bundle(&room, &bundle_info).await;
+        }
+    }
+
+    #[instrument(skip(room), fields(room_id = %room.room_id()))]
+    async fn handle_bundle(room: &Room, bundle_info: &RoomKeyBundleInfo) {
+        if Self::should_accept_bundle(room, bundle_info) {
+            info!("Accepting an out of order key bundle.");
```
andybalaam: Not really "out of order", right? More "late-arriving".

crates/matrix-sdk/src/encryption/tasks.rs
```diff
@@ -385,15 +390,77 @@ impl BackupDownloadTaskListenerState {
...
+
+            if let Err(e) =
+                shared_room_history::maybe_accept_key_bundle(room, &bundle_info.sender).await
+            {
+                warn!("Couldn't accept a late room key bundle {e:?}");
```
andybalaam, line 439: Would probably be helpful to include info about the bundle info event e.g. event ID, or the bundle URL?

crates/matrix-sdk/src/encryption/tasks.rs
```diff
@@ -385,15 +390,77 @@ impl BackupDownloadTaskListenerState {
...
+            }
+        } else {
+            info!("Refusing to accept a historic room key bundle.");
```
andybalaam, line 442: Also here would be useful to have more info about the bundle.

crates/matrix-sdk/src/encryption/tasks.rs
andybalaam: Or will there be another PR later to hold on to these for a while, in case the invite comes later?

poljar: We already hold on to these for a while. That's the normal case. The bundle arrives and we hold on to it, the invitation can arrive either at the same time as the bundle or after it.

We consider bundles for acceptance only when the user accepts the invitation.

crates/matrix-sdk/src/encryption/tasks.rs
poljar: Those are included as fields in the span using the `instrument` call above the method.

crates/matrix-sdk/src/encryption/tasks.rs
poljar: Same as https://github.com/matrix-org/matrix-rust-sdk/pull/5322#discussion_r2177655981.

poljar:
I think we'll want to rebase this on top of #5333. We should at least restrict this behavior to accept such bundles only if the user explicitly accepted the invite, so the homeserver can't join you into a room.

poljar:
Ok, I think this is finally ready for the real review. I had to rework the way we test this since we can't really guarantee that a room and a to-device event don't arrive in the same sync.

poljar:
Rebased to get rid of the revert, the revert was for a commit introduced in this PR.

crates/matrix-sdk/src/encryption/tasks.rs
poljar: Sorry, forgot to mention. I reworded this.

crates/matrix-sdk/src/test_utils/mocks/mod.rs
```diff
@@ -382,7 +414,7 @@ impl MatrixMockServer {
     pub fn mock_room_send(&self) -> MockEndpoint<'_, RoomSendEndpoint> {
         let mock = Mock::given(method("PUT"))
             .and(path_regex(r"^/_matrix/client/v3/rooms/.*/send/.*".to_owned()));
-        self.mock_endpoint(mock, RoomSendEndpoint).expect_default_access_token()
+        self.mock_endpoint(mock, RoomSendEndpoint)
```
zecakeh, line 417: Those are not copy-paste errors. Those endpoints require an access token according to the spec so they require it by default here.

poljar: Hmm, ok. Doesn't seem that we need to expect an acccess token for all of those.

But I guess I can revert that commit and add an override to my mock that doesn't use the default access token.

Thanks.

zecakeh: To be clear, I am not asking to revert the commit, I was just explaining why it was like that because of what I read in the commit message. I think that when mocking an endpoint it makes sense that by default it has the same behavior as a homeserver and if a test doesn't want those defaults it should overwrite them.

Issue #3123, opened 2024-02-13
timeline: Support threaded receipts


bnjbvr:
Right now, the SDK assumes in multiple places that we only support unthreaded receipts. We also try our best to be compatible with receipts threaded from the `main` thread.

When we'll want to support threaded receipts, we'll need to modify a bunch of code, so I'm opening this issue so as to be able to link to it from the code.

To make expectations clear: it's not a priority for us at the moment to support threads or threaded receipts, as far as I understand (but as usual, contributions would be welcome :)).

PR #5325, opened 2025-07-02
feat(sdk): send threaded read receipts

stefanceriu:
This series of patches introduce proper threading support for sending read receipts and computing room unread counts.

They add a new, client wide, `threads_enabled` feature flag which overtake the old `TimelineFocus:: hide_threaded_events`.

If this feature flag is turned on then the timelines will automatically send [the correct read receipt](https://spec.matrix.org/latest/client-server-api/#threaded-read-receipts) when invoked through the existing `send_read_receipt`/`send_single_receipt`/ method.
The base SDK's read receipts handler will also start ignoring threaded messages when computing room unread counts. A room will be unread only if there are new messages in its main timeline.

Part of https://github.com/matrix-org/matrix-rust-sdk/issues/3123

### DIFF
--- a/crates/matrix-sdk-base/src/client.rs
+++ b/crates/matrix-sdk-base/src/client.rs
@@ -76,11 +76,12 @@ use crate::{
 /// rather through `matrix_sdk::Client`.
 ///
 /// ```rust
-/// use matrix_sdk_base::{store::StoreConfig, BaseClient};
+/// use matrix_sdk_base::{store::StoreConfig, BaseClient, ThreadingSupport};
 ///
-/// let client = BaseClient::new(StoreConfig::new(
-///     "cross-process-holder-name".to_owned(),
-/// ));
+/// let client = BaseClient::new(
+///     StoreConfig::new("cross-process-holder-name".to_owned()),
+///     ThreadingSupport::Disabled,
+/// );
 /// ```
 #[derive(Clone)]
 pub struct BaseClient {
@@ -122,6 +123,9 @@ pub struct BaseClient {
     /// If the client should handle verification events received when syncing.
     #[cfg(feature = "e2e-encryption")]
     pub handle_verification_events: bool,
+
+    /// Whether the client supports threads or not.
+    pub threading_support: ThreadingSupport,
 }
 
 #[cfg(not(tarpaulin_include))]
@@ -134,6 +138,25 @@ impl fmt::Debug for BaseClient {
     }
 }
 
+/// Whether this client instance supports threading or not. Currently used to
+/// determine how the client handles read receipts and unread count computations
+/// on the base SDK level.
+///
+/// Timelines on the other hand have a separate `TimelineFocus`
+/// `hide_threaded_events` associated value that can be used to hide threaded
+/// events but also to enable threaded read receipt sending. This is because
+/// certain timeline instances should ignore threading no matter what's defined
+/// at the client level. One such example are media filtered timelines which
+/// should contain all the room's media no matter what thread its in (unless
+/// explicitly opted into).
+#[derive(Clone, Copy, Debug)]
+pub enum ThreadingSupport {
+    /// Threading enabled
+    Enabled,
+    /// Threading disabled
+    Disabled,
+}
+
 impl BaseClient {
     /// Create a new client.
     ///
@@ -141,7 +164,7 @@ impl BaseClient {
     ///
     /// * `config` - the configuration for the stores (state store, event cache
     ///   store and crypto store).
-    pub fn new(config: StoreConfig) -> Self {
+    pub fn new(config: StoreConfig, threading_support: ThreadingSupport) -> Self {
         let store = BaseStateStore::new(config.state_store);
 
         // Create the channel to receive `RoomInfoNotableUpdate`.
@@ -173,6 +196,7 @@ impl BaseClient {
             },
             #[cfg(feature = "e2e-encryption")]
             handle_verification_events: true,
+            threading_support,
         }
     }
 
@@ -204,6 +228,7 @@ impl BaseClient {
             room_key_recipient_strategy: self.room_key_recipient_strategy.clone(),
             decryption_settings: self.decryption_settings.clone(),
             handle_verification_events,
+            threading_support: self.threading_support,
         };
 
         copy.state_store
@@ -224,7 +249,7 @@ impl BaseClient {
     ) -> Result<Self> {
         let config = StoreConfig::new(cross_process_store_locks_holder.to_owned())
             .state_store(MemoryStore::new());
-        Ok(Self::new(config))
+        Ok(Self::new(config, ThreadingSupport::Disabled))
     }
 
     /// Get the session meta information.
@@ -1080,6 +1105,7 @@ mod tests {
 
     use super::{BaseClient, RequestedRequiredStates};
     use crate::{
+        client::ThreadingSupport,
         store::{RoomLoadSettings, StateStoreExt, StoreConfig},
         test_utils::logged_in_base_client,
         RoomDisplayName, RoomState, SessionMeta,
@@ -1374,8 +1400,10 @@ mod tests {
         let user_id = user_id!("@alice:example.org");
         let room_id = room_id!("!ithpyNKDtmhneaTQja:example.org");
 
-        let client =
-            BaseClient::new(StoreConfig::new("cross-process-store-locks-holder-name".to_owned()));
+        let client = BaseClient::new(
+            StoreConfig::new("cross-process-store-locks-holder-name".to_owned()),
+            ThreadingSupport::Disabled,
+        );
         client
             .activate(
                 SessionMeta { user_id: user_id.to_owned(), device_id: "FOOBAR".into() },
@@ -1434,8 +1462,10 @@ mod tests {
         let inviter_user_id = user_id!("@bob:example.org");
         let room_id = room_id!("!ithpyNKDtmhneaTQja:example.org");
 
-        let client =
-            BaseClient::new(StoreConfig::new("cross-process-store-locks-holder-name".to_owned()));
+        let client = BaseClient::new(
+            StoreConfig::new("cross-process-store-locks-holder-name".to_owned()),
+            ThreadingSupport::Disabled,
+        );
         client
             .activate(
                 SessionMeta { user_id: user_id.to_owned(), device_id: "FOOBAR".into() },
@@ -1496,8 +1526,10 @@ mod tests {
         let inviter_user_id = user_id!("@bob:example.org");
         let room_id = room_id!("!ithpyNKDtmhneaTQja:example.org");
 
-        let client =
-            BaseClient::new(StoreConfig::new("cross-process-store-locks-holder-name".to_owned()));
+        let client = BaseClient::new(
+            StoreConfig::new("cross-process-store-locks-holder-name".to_owned()),
+            ThreadingSupport::Disabled,
+        );
         client
             .activate(
                 SessionMeta { user_id: user_id.to_owned(), device_id: "FOOBAR".into() },
@@ -1568,8 +1600,10 @@ mod tests {
     #[async_test]
     async fn test_ignored_user_list_changes() {
         let user_id = user_id!("@alice:example.org");
-        let client =
-            BaseClient::new(StoreConfig::new("cross-process-store-locks-holder-name".to_owned()));
+        let client = BaseClient::new(
+            StoreConfig::new("cross-process-store-locks-holder-name".to_owned()),
+            ThreadingSupport::Disabled,
+        );
 
         client
             .activate(
--- a/crates/matrix-sdk-base/src/read_receipts.rs
+++ b/crates/matrix-sdk-base/src/read_receipts.rs
@@ -122,7 +122,10 @@ use std::{
     num::NonZeroUsize,
 };
 
-use matrix_sdk_common::{deserialized_responses::TimelineEvent, ring_buffer::RingBuffer};
+use matrix_sdk_common::{
+    deserialized_responses::TimelineEvent, ring_buffer::RingBuffer,
+    serde_helpers::extract_thread_root,
+};
 use ruma::{
     events::{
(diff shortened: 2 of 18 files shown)


crates/matrix-sdk-ui/src/timeline/controller/mod.rs
```diff
@@ -1449,6 +1449,27 @@ impl TimelineController {
         Ok(())
     }
 
+    pub(super) fn thread_for_read_receipt(&self, receipt_type: &SendReceiptType) -> ReceiptThread {
```
bnjbvr: Let's add a small doc comment explaining what it does. At first I thought it would return the thread for a given read receipt, based on the name, so I think we should also rename it: maybe `infer_thread_for_receipt`? (+/- `read_`)

crates/matrix-sdk-ui/src/timeline/mod.rs
```diff
@@ -582,9 +582,10 @@ impl Timeline {
     pub async fn send_single_receipt(
         &self,
         receipt_type: ReceiptType,
-        thread: ReceiptThread,
```
bnjbvr, line 585: Since this is a public method, this warrants a CHANGELOG entry (sorry!).

Can you also add in the doc comment that the receipt thread is inferred from the context?

crates/matrix-sdk-ui/tests/integration/timeline/read_receipts.rs
```diff
@@ -414,7 +414,12 @@ async fn test_send_single_receipt() {
 
     server.mock_room_state_encryption().plain().mount().await;
 
-    let timeline = room.timeline().await.unwrap();
+    let timeline = room
+        .timeline_builder()
+        .with_focus(TimelineFocus::Live { hide_threaded_events: false })
+        .build()
+        .await
+        .unwrap();
```
bnjbvr, line 424: Is that not equivalent to the default? If so, did you make this change to make the test more explicit, maybe?

crates/matrix-sdk-ui/tests/integration/timeline/read_receipts.rs
```diff
@@ -885,14 +811,13 @@ async fn test_send_single_receipt_with_unread_flag() {
...
 
-        timeline
-            .send_single_receipt(
-                CreateReceiptType::Read,
-                ReceiptThread::Main,
-                second_receipts_event_id,
-            )
-            .await
-            .unwrap();
+        room.send_single_receipt(
```
bnjbvr, line 956: It's a bit weird to have the `room` send the read receipt in this part of the test. While I understand the change, could we, instead, move this part of the test (the entire scope, maybe) into another thread-specific test, later, that would make use of a threaded timeline?

crates/matrix-sdk-base/src/read_receipts.rs
```diff
@@ -457,7 +471,6 @@ pub(crate) fn compute_unread_counts(
         previous_events.extend(new_events.iter().cloned());
         previous_events
     };
-
```
bnjbvr: nit: please keep the newline :P it's nice to have some breathing air between large statements.

crates/matrix-sdk-base/src/read_receipts.rs
```diff
@@ -202,6 +206,10 @@ impl RoomReadReceipts {
     /// Returns whether a new event triggered a new unread/notification/mention.
     #[inline(always)]
     fn process_event(&mut self, event: &TimelineEvent, user_id: &UserId) {
+        if Self::should_ignore(event) {
+            return;
+        }
```
bnjbvr: This new behavior would need a test :) Feel free to write one (or more) unit tests in this file, since this functionality can be tested standalone, without an integration test.

crates/matrix-sdk-base/src/client.rs
```diff
@@ -141,7 +145,7 @@ impl BaseClient {
     ///
     /// * `config` - the configuration for the stores (state store, event cache
     ///   store and crypto store).
-    pub fn new(config: StoreConfig) -> Self {
+    pub fn new(config: StoreConfig, threads_enabled: bool) -> Self {
```
bnjbvr: Having a bool in public APIs may be quite confusing; let's use a simple enum (that behaves as a bool) so as to make a nice public API.

```rust
pub enum ThreadSupport {
  Enabled,
  Disabled
}
```

and here we'd have `thread_support: ThreadSupport`.

crates/matrix-sdk-base/src/read_receipts.rs
```diff
@@ -179,8 +178,14 @@ pub struct RoomReadReceipts {
     /// not the event ids of the receipt events themselves.
     #[serde(default = "new_nonempty_ring_buffer")]
     pending: RingBuffer<OwnedEventId>,
+
+    /// Whether the threads feature flag has been enabled and threaded
+    /// messages should be ignored from the unread computations.
+    #[serde(default)]
+    threads_enabled: bool,
```
bnjbvr: I don't think it should be deserialized; instead, it should be a parameter passed to `compute_unread_counts` (and propagated throughout the call tree, down to `process_event`). Otherwise, there's a risk of state mismatch (e.g. a client has `!threads_enabled` but a previously serialized `RoomReadReceipts` had it set). It should also require fewer code changes, namely a `BaseRoom` shouldn't have to contain the flag either.

crates/matrix-sdk-ui/tests/integration/timeline/read_receipts.rs
stefanceriu: Yes that but I think we should also move away from that `room.timeline()` method and eventually remove it entirely.

crates/matrix-sdk-ui/src/timeline/mod.rs
stefanceriu: Don't be sorry, my bad 👍

crates/matrix-sdk-ui/tests/integration/timeline/read_receipts.rs
stefanceriu: I'm not sure tbh, this is more about being able to configure a read receipts thread separate from the timeline instance you're using, it wouldn't belogn in the threading tests either necessarily 🤔

crates/matrix-sdk-base/src/client.rs
stefanceriu: Good point 👍

crates/matrix-sdk-base/src/read_receipts.rs
```diff
@@ -181,6 +180,7 @@ pub struct RoomReadReceipts {
     pending: RingBuffer<OwnedEventId>,
 }
 
+/// `Default` conformance is useful for store migrations
```
bnjbvr: Not sure what this comment means, consider removing it?

crates/matrix-sdk-base/src/read_receipts.rs
```diff
@@ -276,8 +277,8 @@ impl RoomReadReceipts {
 
     /// Checks whether the given `TimelineEvent` is part of a thread and should
     /// be ignored.
-    fn should_ignore(event: &TimelineEvent) -> bool {
-        return extract_thread_root(&event.raw()).is_some();
+    fn should_ignore(&self, event: &TimelineEvent, threads_enabled: bool) -> bool {
+        threads_enabled && extract_thread_root(event.raw()).is_some()
     }
```
bnjbvr: We should probably inline this function (and the comment) into its one caller; the name doesn't say anything about thread, and makes sense only after reading the comment.

crates/matrix-sdk-base/src/response_processors/room/msc4186/mod.rs
```diff
@@ -60,6 +60,7 @@ pub enum RoomUpdateKind {
     Knocked(KnockedRoomUpdate),
 }
 
+#[allow(clippy::too_many_arguments)]
```
bnjbvr: This can be removed now

crates/matrix-sdk-base/src/response_processors/room/msc4186/mod.rs
```diff
@@ -224,6 +225,7 @@ pub async fn update_any_room(
 ///
 /// If there is any invite state events, the room can be considered an invited
 /// or knocked room, depending of the membership event (if any).
+#[allow(clippy::too_many_arguments)]
```
bnjbvr: Here too this can be removed now

crates/matrix-sdk-base/src/room/room_info.rs
```diff
@@ -480,7 +480,7 @@ impl RoomInfo {
             sync_info: SyncInfo::NoState,
             encryption_state_synced: false,
             latest_event: None,
-            read_receipts: Default::default(),
+            read_receipts: RoomReadReceipts::default(),
```
bnjbvr: Looks like all the changes in this file can now be reverted.

crates/matrix-sdk-base/src/sliding_sync.rs
```diff
@@ -284,6 +285,7 @@ impl BaseClient {
                 room_previous_events,
                 &joined_room_update.timeline.events,
                 &mut room_info.read_receipts,
+                matches!(self.threading_support, ThreadingSupport::Enabled),
```
bnjbvr: (light suggestion) Maybe `compute_unread_counts` could also take a `ThreadingSupport` parameter?

crates/matrix-sdk-base/src/store/migration_helpers.rs
```diff
@@ -115,7 +116,7 @@ impl RoomInfoV1 {
             sync_info,
             encryption_state_synced,
             latest_event: latest_event.map(|ev| Box::new(LatestEvent::new(ev))),
-            read_receipts: Default::default(),
+            read_receipts: RoomReadReceipts::default(),
```
bnjbvr: Changes in this file can be reverted too

crates/matrix-sdk/src/client/builder/mod.rs
```diff
@@ -113,6 +113,7 @@ pub struct ClientBuilder {
     #[cfg(feature = "e2e-encryption")]
     enable_share_history_on_invite: bool,
     cross_process_store_locks_holder_name: String,
+    threads_enabled: bool,
```
bnjbvr: Think it should store a `ThreadingSupport` that's defined in the setter.

crates/matrix-sdk/src/client/builder/mod.rs
```diff
@@ -479,6 +481,14 @@ impl ClientBuilder {
         self
     }
 
+    /// Whether the threads feature is enabled throuoghout the SDK.
+    /// This will affect how timelines are setup, how read receipts are sent
+    /// and how room unreads are computed.
+    pub fn threads_enabled(mut self, enabled: bool) -> Self {
```
bnjbvr: Consider calling this `enable_thread_support`, maybe?

crates/matrix-sdk-base/src/client.rs
```diff
@@ -134,14 +138,23 @@ impl fmt::Debug for BaseClient {
     }
 }
 
+/// Whether this client instance supports threading or not
```
bnjbvr: We should explain a bit more what this entails: in-thread events might be excluded from timeline by default, unread counts will ignore threads, and so on.

crates/matrix-sdk-base/src/read_receipts.rs
```diff
@@ -1547,4 +1560,56 @@ mod tests {
...
+        let own_alice = user_id!("@alice:example.org");
+        let bob = user_id!("@bob:example.org");
+
+        // Threaded messages from myself or other users shouldn't change the
+        // unread counts.
+        receipts.process_event(
+            &make_event(own_alice, event_id!("$some_thread_root")),
+            own_alice,
+            true,
+        );
```
bnjbvr: Note that in this commit (2nd), `process_event` doesn't take a boolean parameter yet, so the test wouldn't compile if checked out on this specific commit. Consider removing the bool param here, or squashing commits 2 and 3 :)

bnjbvr (approved):
Cheers! A few cosmetic requests, and then this should be good to go :ok_hand:

crates/matrix-sdk-base/src/read_receipts.rs
stefanceriu: Artefact from the previous feature flag implementation when I introduced a "new" constructor but had to keep this around for the migration.

crates/matrix-sdk-base/src/room/room_info.rs
stefanceriu: To be honest I find this version a lot easier to read and find, `Default::default` is way too generic

bnjbvr: I don't wanna brag, but my editor shows me inline what the actual type is, so I don't need this. Agreed it makes it slightly more useful for code reviews, though, but I'm not sure I find enough worth that it justifies the import. It's fine to keep as is, though.

stefanceriu: Yeah, this is more about discovery and ease of reading outside an IDE. We have 500+ `Default::default`s throughout the SDK though so I guess I might as well get with the program.

crates/matrix-sdk-base/src/client.rs
```diff
@@ -134,14 +138,33 @@ impl fmt::Debug for BaseClient {
...
+
 impl BaseClient {
     /// Create a new client.
     ///
     /// # Arguments
     ///
     /// * `config` - the configuration for the stores (state store, event cache
     ///   store and crypto store).
-    pub fn new(config: StoreConfig) -> Self {
+    pub fn new(config: StoreConfig, threading_support: ThreadingSupport) -> Self {
```
poljar, line 167: This was a breaking change to the public API of the base crate. Could we get a changelog entry for this?

bnjbvr: Thanks, added in https://github.com/matrix-org/matrix-rust-sdk/pull/5359

PR #5326, opened 2025-07-02
refactor(ffi): Drop support for observing `AccountDataEventType::SecretStorageKey`

zecakeh:
This type cannot be observed because it is a dynamic event type and this branch will fail to compile with the next version of Ruma.

### DIFF
--- a/bindings/matrix-sdk-ffi/src/client.rs
+++ b/bindings/matrix-sdk-ffi/src/client.rs
@@ -69,9 +69,7 @@ use ruma::{
             message::OriginalSyncRoomMessageEvent,
             power_levels::RoomPowerLevelsEventContent,
         },
-        secret_storage::{
-            default_key::SecretStorageDefaultKeyEventContent, key::SecretStorageKeyEventContent,
-        },
+        secret_storage::default_key::SecretStorageDefaultKeyEventContent,
         tag::TagEventContent,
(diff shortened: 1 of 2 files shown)


bindings/matrix-sdk-ffi/src/client.rs
```diff
@@ -573,11 +571,14 @@ impl Client {
     /// Be careful that only the most recent value can be observed. Subscribers
     /// are notified when a new value is sent, but there is no guarantee that
     /// they will see all values.
+    ///
+    /// Doesn't work for `m.secret_storage.key.<key_id>`, because it is
+    /// currently not observable.
```
zecakeh, line 576: Note that we are adding stuff to Ruma that should make this possible in the future (https://github.com/ruma/ruma/pull/2136), although the event handler code will need to be adapted in the SDK.

So maybe it's better not to touch the function signature and panic for now?

zecakeh:
Actually what we have in https://github.com/ruma/ruma/pull/2136 is probably not the right answer. We might need to do another breaking change that will require changes to the event handler code and fix this.

zecakeh:
There is a fix for this in #5337 so we might not need this. We could make #5337 smaller by merging a version of this (since the current code doesn't work anyway), and then applying the fix in another PR after #5337.

zecakeh:
Actually we found a way where we don't need to fix anything until after Ruma is bumped.

PR #5328, opened 2025-07-02
chore(ci): use the dev profile when building the swift bindings

### LINKED ISSUE #4416: Build for iOS (aarch64-apple-ios) is broken
Hello.

It seems I cannot build the SDK for aarch64-apple-ios.

I've added a pipeline to demonstrate this here:

https://github.com/olliiiver/matrix-rust-sdk-build/actions/runs/12320983696/job/34391426770

The error message is:

````
  = note: ld: warning: object file (/Users/runner/work/matrix-rust-sdk-build/matrix-rust-sdk-build/matrix-rust-sdk/target/aarch64-apple-ios/debug/deps/libblake3-e40b55db2257280c.rlib[6](db3b6bfb95261072-blake3_neon.o)) was built for newer 'iOS' version (17.5) than being linked (10.0)
          Undefined symbols for architecture arm64:
            "___chkstk_darwin", referenced from:
                _blake3_hash4_neon in libblake3-e40b55db2257280c.rlib[6](db3b6bfb95261072-blake3_neon.o)
          ld: symbol(s) not found for architecture arm64
          clang: error: linker command failed with exit code 1 (use -v to see invocation)
````

How can this be fixed?

### PR DESCRIPTION
**Pull request by bnjbvr**
The swift bindings aren't getting tested (they don't run) in CI anymore, so building with the reldbg profile (that's a workaround to make it run and not crash in production) doesn't provide more value than building in debug mode, while taking much longer to build. Let's use the default dev profile for this; we have to specify it manually, because the default for the xtask command is to use the `reldbg` profile otherwise.

Fixes #4416.

### DIFF
--- a/.github/workflows/bindings_ci.yml
+++ b/.github/workflows/bindings_ci.yml
@@ -175,7 +175,7 @@ jobs:
         run: swift test
 
       - name: Build Framework
-        run: target/debug/xtask swift build-framework --target=aarch64-apple-ios --profile=dev --iphoneos-deployment-target=18.0
+        run: target/debug/xtask swift build-framework --target=aarch64-apple-ios --profile=dev --ios-deployment-target=18.0
 
   complement-crypto:
     name: "Run Complement Crypto tests"

### CODE REVIEW

**Comment by stefanceriu** [comment]
This makes sense to me but unfortunately I can reproduce the CI failure locally 
```
Undefined symbols for architecture arm64:
            "___chkstk_darwin", referenced from:
                _blake3_hash4_neon in libblake3-74136ca20ab62b27.rlib[6](a1edd97dd51cd48d-blake3_neon.o)
          ld: symbol(s) not found for architecture arm64
```

**Comment by stefanceriu** [comment]
Same on `dev` and `dbg` so I reckon changing it to anything other than `reldbg` defeats the purpose of having the test to beging with.

**Comment by bnjbvr** [comment]
Yep. I'll try one potential fix (inspired from https://github.com/olliiiver/Blake_test/pull/2/files), that we may include in the swift xtask directly, but if it doesn't work I'll close this PR instead :+1:

File: .github/workflows/bindings_ci.yml [hunk 1]
```diff
@@ -175,7 +175,7 @@ jobs:
         run: swift test
 
       - name: Build Framework
-        run: target/debug/xtask swift build-framework --target=aarch64-apple-ios --profile=reldbg
+        run: target/debug/xtask swift build-framework --target=aarch64-apple-ios --profile=dev
```
bnjbvr:
```suggestion
        run: target/debug/xtask swift build-framework --target=aarch64-apple-ios
```

xtask/src/swift.rs
```diff
@@ -64,12 +72,21 @@ impl SwiftArgs {
...
                 target: targets,
                 sequentially,
+                iphoneos_deployment_target,
             } => {
-                // The dev profile seems to cause crashes on some platforms so we default to
-                // reldbg (https://github.com/matrix-org/matrix-rust-sdk/issues/4009)
-                let profile =
-                    profile.as_deref().unwrap_or(if release { "release" } else { "reldbg" });
-                build_xcframework(profile, targets, components_path, sequentially)
+                let profile = profile.as_deref().unwrap_or(if release { "release" } else { "dev" });
```
stefanceriu: We should really keep this comment and `reldbg` as the default

bnjbvr:
This reduces build times for the swift binding task from around 13 or 14 minutes (in other PRs) down to 8 minutes :partying_face:

PR #5329, opened 2025-07-02
refactor(event cache): generalize `RoomEvents` in preparation for threads

bnjbvr:
This includes a series of refactorings / renamings I've been meaning to do for a long time, and that should help with reusing code maximally for a threaded timeline. Next steps would involve including the removal of duplicate events into the mix, but the PR is already large enough.

PR #5330, opened 2025-07-02
feat(sdk): `Add RoomEventCache::rfind_event_in_memory_by`

Hywan:
This patch introduces a new method named `RoomEventCache::rfind_event_in_memory_by` to look for an in-memory event, starting from the most recent event, by applying a predicate closure on each event.

Because this method's name has the pattern `find_event*_by`, this patch is also renaming `event` to `find_event_by_id` and `event_with_relations` to `find_event_by_id_with_relations`. I think it is mandatory and it's more Rust idiomatic as it follows the standard library namings. Those methods are no more getters, but methods going over an iterator. Proof is: `RoomEventCache::event` was entirely using `RoomEventCacheState::find_event` (this last naming was correct!).

### DIFF
--- a/crates/matrix-sdk/src/event_cache/room/mod.rs
+++ b/crates/matrix-sdk/src/event_cache/room/mod.rs
@@ -214,8 +214,11 @@ impl RoomEventCache {
         RoomPagination { inner: self.inner.clone() }
     }
 
-    /// Try to find an event by id in this room.
-    pub async fn event(&self, event_id: &EventId) -> Option<Event> {
+    /// Try to find an event by ID in this room.
+    ///
+    /// It starts by looking into loaded events before looking inside the
+    /// storage.
+    pub async fn find_event(&self, event_id: &EventId) -> Option<Event> {
         self.inner
             .state
             .read()
@@ -2210,8 +2213,8 @@ mod timed_tests {
 
         // The rooms knows about all cached events.
         {
-            assert!(room_event_cache.event(event_id1).await.is_some());
-            assert!(room_event_cache.event(event_id2).await.is_some());
+            assert!(room_event_cache.find_event(event_id1).await.is_some());
+            assert!(room_event_cache.find_event(event_id2).await.is_some());
         }
 
         // But only part of events are loaded from the store
@@ -2251,7 +2254,7 @@ mod timed_tests {
 
         // Events individually are not forgotten by the event cache, after clearing a
         // room.
-        assert!(room_event_cache.event(event_id1).await.is_some());
+        assert!(room_event_cache.find_event(event_id1).await.is_some());
 
         // But their presence in a linked chunk is forgotten.
         let items = room_event_cache.events().await;
@@ -2365,8 +2368,8 @@ mod timed_tests {
         assert!(stream.is_empty());
 
         // The event cache knows only all events though, even if they aren't loaded.
-        assert!(room_event_cache.event(event_id1).await.is_some());
-        assert!(room_event_cache.event(event_id2).await.is_some());
+        assert!(room_event_cache.find_event(event_id1).await.is_some());
+        assert!(room_event_cache.find_event(event_id2).await.is_some());
 
         // Let's paginate to load more events.
         room_event_cache.pagination().run_backwards_once(20).await.unwrap();
--- a/crates/matrix-sdk/src/event_cache/room/mod.rs
+++ b/crates/matrix-sdk/src/event_cache/room/mod.rs
@@ -214,6 +214,18 @@ impl RoomEventCache {
         RoomPagination { inner: self.inner.clone() }
     }
 
+    /// Try to find a single event in this room, starting from the most recent
+    /// event.
+    ///
+    /// **Warning**! It looks into the loaded events from the in-memory linked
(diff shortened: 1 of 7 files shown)


bnjbvr:
I'll steal review on this one. Let's wait for #5329 which makes changes in the surrounding area, and it has been opened slightly earlier.

Hywan:
Yeah sure, I'm fine with that.

crates/matrix-sdk/src/event_cache/room/mod.rs
```diff
@@ -219,7 +219,7 @@ impl RoomEventCache {
             .state
             .read()
             .await
-            .find_event(event_id)
+            .find_event_by_id(event_id)
```
bnjbvr: As discussed, maybe we can keep the same name, as the `_by_id` suffix doesn't give us extra information?

crates/matrix-sdk/src/event_cache/room/mod.rs
```diff
@@ -213,8 +213,11 @@ impl RoomEventCache {
         RoomPagination { inner: self.inner.clone() }
     }
 
-    /// Try to find an event by id in this room.
-    pub async fn event(&self, event_id: &EventId) -> Option<Event> {
+    /// Try to find an event by ID in this room.
+    ///
+    /// It starts by looking into loaded events before looking inside the
+    /// storage.
+    pub async fn find_event(&self, event_id: &EventId) -> Option<Event> {
```
bnjbvr: I like this one, though, thanks!

crates/matrix-sdk/src/event_cache/room/mod.rs
```diff
@@ -213,6 +213,18 @@ impl RoomEventCache {
         RoomPagination { inner: self.inner.clone() }
     }
 
+    /// Try to find a single event in this room, starting from the most recent
+    /// event.
+    ///
+    /// **Warning**! It looks into the loaded events in `RoomEvents`
```
bnjbvr: Consider using ```[`RoomEvents`]``` here, so the renaming in my PR would have become a doc build error here :)

(or better, don't use the type name, and rephrase as: "it looks into the in-memory linked chunk only"

crates/matrix-sdk/src/event_cache/room/mod.rs
```diff
@@ -1140,6 +1152,20 @@ mod private {
             &self.events
         }
 
+        //// Find a single event in this room, starting from the most recent event.
+        ///
+        /// **Warning**! It looks into the loaded events in `RoomEvents`
```
bnjbvr: And here too.

crates/matrix-sdk/src/event_cache/room/mod.rs
```diff
@@ -1140,6 +1152,20 @@ mod private {
...
+        /// **only**. It doesn't look inside the storage, contrary to
+        /// [`Self::find_event_by_id`].
+        pub fn rfind_event_in_memory_by<P>(&self, mut predicate: P) -> Option<Event>
+        where
+            P: FnMut(&Event) -> bool,
+        {
+            self.events
+                .revents()
+                .find_map(|(_position, event)| predicate(event).then_some(event.clone()))
```
bnjbvr: I think it's better to use `.then(|| Some(event.clone())` here, otherwise if the predicate is false, the cloned value will be unused (and thus the clone will be wasted).

crates/matrix-sdk/src/event_cache/room/mod.rs
```diff
@@ -3035,4 +3035,80 @@ mod timed_tests {
         assert_eq!(events3.len(), 1);
         assert_eq!(events3[0].event_id().as_deref(), Some(evid2));
     }
+
+    #[async_test]
+    async fn test_rfind_event_in_memory_by() {
```
bnjbvr, line 2988: What I would like to see in this test is proof that it doesn't look inside the store; can you have a partially filled in-memory linked chunk, and look for an event that's in the store but not in memory, please?

bnjbvr (approved):
Some tiny change request about the test, and then we're good to go. Cheers!

crates/matrix-sdk/src/event_cache/room/mod.rs
Hywan: Done.

crates/matrix-sdk/src/event_cache/room/mod.rs
Hywan: Ah true! How did I miss that… thanks.

For the record, it's:

```rust
.then(|| event.clone())
```

The `Some(…)` isn't necessary.

bnjbvr (changes requested):
Can you drop the `find_event_by_id` renaming commit, as discussed, please?

crates/matrix-sdk/src/event_cache/room/mod.rs
Hywan: Yup, good idea, especially to detect regression. On it.

Hywan:
Yup yup it's done Sir.

bnjbvr (approved):
:heart_hands:

PR #5331, opened 2025-07-02
feat(timeline): expose timestamp & identifier on embedded events

Johennes:
These are available in normal timeline items. Having them on embedded items, too, allows clients to render them (timestamp) or use them for further operations (event ID).

### DIFF
--- a/crates/matrix-sdk-ui/src/timeline/event_item/content/reply.rs
+++ b/crates/matrix-sdk-ui/src/timeline/event_item/content/reply.rs
@@ -16,7 +16,7 @@ use std::sync::Arc;
 
 use imbl::Vector;
 use matrix_sdk::deserialized_responses::TimelineEvent;
-use ruma::{OwnedEventId, OwnedUserId};
+use ruma::{MilliSecondsSinceUnixEpoch, OwnedEventId, OwnedUserId};
 use tracing::{debug, instrument, warn};
 
 use super::TimelineItemContent;
@@ -25,7 +25,7 @@ use crate::timeline::{
     event_handler::TimelineAction,
     event_item::{EventTimelineItem, Profile, TimelineDetails},
     traits::RoomDataProvider,
-    Error as TimelineError, TimelineItem,
+    Error as TimelineError, TimelineEventItemId, TimelineItem,
 };
 
 /// Details about an event being replied to.
@@ -68,6 +68,13 @@ pub struct EmbeddedEvent {
     pub sender: OwnedUserId,
     /// The profile of the sender of the related embedded event.
     pub sender_profile: TimelineDetails<Profile>,
+    /// The timestamp of the event.
+    pub timestamp: MilliSecondsSinceUnixEpoch,
+    /// The unique identifier of this event.
+    ///
+    /// This is the transaction ID for a local echo that has not been sent and
+    /// the event ID for a local echo that has been sent or a remote event.
+    pub identifier: TimelineEventItemId,
 }
 
 impl EmbeddedEvent {
@@ -77,6 +84,8 @@ impl EmbeddedEvent {
             content: timeline_item.content.clone(),
             sender: timeline_item.sender.clone(),
             sender_profile: timeline_item.sender_profile.clone(),
+            timestamp: timeline_item.timestamp,
+            identifier: timeline_item.identifier(),
         }
     }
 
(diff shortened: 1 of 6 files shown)


crates/matrix-sdk-ui/src/timeline/event_item/content/reply.rs
```diff
@@ -68,6 +68,12 @@ pub struct EmbeddedEvent {
     pub sender: OwnedUserId,
     /// The profile of the sender of the related embedded event.
     pub sender_profile: TimelineDetails<Profile>,
+    /// The timestamp of the event.
+    pub timestamp: MilliSecondsSinceUnixEpoch,
+    /// The unique identifier of this event. This is the transaction ID
+    /// for a local echo that has not been sent and the event ID for a
+    /// local echo that has been sent or a remote event.
```
bnjbvr: nit: let's use the shorthand format for getting a short description with LSP:

```suggestion
    /// The unique identifier of this event.
    ///
    /// This is the transaction ID for a local echo that has not been sent and the event ID for a
    /// local echo that has been sent or a remote event.
```

crates/matrix-sdk-ui/tests/integration/timeline/replies.rs
```diff
@@ -89,6 +89,7 @@ async fn test_in_reply_to_details() {
         // We get the original message.
         assert_let!(VectorDiff::PushBack { value: first } = &timeline_updates[0]);
         assert!(first.as_event().unwrap().content().is_message());
+        let timestamp = first.as_event().unwrap().timestamp();
```
bnjbvr: Can you specify a custom timestamp when building the event, instead? (that way, we don't rely on the internals of the `EventFactory`, which may or may not set a 0 timestamp for all events :P — I know it doesn't at the moment, but it's usually better to get full control over the data, in testing situations)

bnjbvr (approved):
lgtm, thanks!

crates/matrix-sdk-ui/tests/integration/timeline/replies.rs
Johennes: Good call. Have made it so.

crates/matrix-sdk-ui/src/timeline/event_item/content/reply.rs
bnjbvr: (sowwy i broke formatting with my suggestion)

crates/matrix-sdk-ui/src/timeline/event_item/content/reply.rs
```diff
@@ -68,6 +68,13 @@ pub struct EmbeddedEvent {
     pub sender: OwnedUserId,
     /// The profile of the sender of the related embedded event.
     pub sender_profile: TimelineDetails<Profile>,
+    /// The timestamp of the event.
+    pub timestamp: MilliSecondsSinceUnixEpoch,
+    /// The unique identifier of this event.
+    ///
+    /// This is the transaction ID for a local echo that has not been sent and the event ID for a
+    /// local echo that has been sent or a remote event.
```
bnjbvr:
```suggestion
    /// This is the transaction ID for a local echo that has not been sent and
    /// the event ID for a local echo that has been sent or a remote event.
```

bnjbvr:
@Johennes Can you provide a signoff in a commit's description (and if you do so, feel free to also reorganize commits to make a neat history, so we can rebase-merge), or in the PR's description, please?

Johennes:
Sorry. I swear I had a sign-off but I probably lost it when rebasing. 🤦 

To double check: I tried using fixup commits earlier but the CI blocked them. So is rebasing the intended way to address review comments now?

bnjbvr:
> To double check: I tried using fixup commits earlier but the CI blocked them. So is rebasing the intended way to address review comments now?

The CI complains that we have fixup commits, just so as to make sure we notice them, and not merge a PR with fixup commits (because Github unfortunately doesn't auto-squash them, when rebase-merging). It really is the preferred way, though, as mentioned in our [contribution guidelines](https://github.com/matrix-org/matrix-rust-sdk/blob/c4c108d525d7cd333bcb7dba746327c5cdf1d30c/CONTRIBUTING.md?plain=1#L229); the idea would be that, after getting a review, you'd rebase and autosquash the fixup commits yourself.

It's also totally fine to not do this at all; depending on the reviewer, the full PR would be squash-merged (to keep a tidy commit history), or just rebase-merged as is. I tend to squash-merge, in this case, as I find it's not worth bothering a contributor to ask for a perfectly tidy history all the time :+1:

bnjbvr:
Thanks for the signoff, btw!

PR #5333, opened 2025-07-02
feat(base): Remember when a user accepts an invite

poljar:
I think that this is necessary to make #5322 somewhat secure.

Part of https://github.com/matrix-org/matrix-rust-sdk/issues/4926. 

### DIFF
--- a/crates/matrix-sdk-base/src/room/room_info.rs
+++ b/crates/matrix-sdk-base/src/room/room_info.rs
@@ -46,8 +46,8 @@ use ruma::{
     },
     room::RoomType,
     serde::Raw,
-    EventId, MxcUri, OwnedEventId, OwnedMxcUri, OwnedRoomAliasId, OwnedRoomId, OwnedUserId,
-    RoomAliasId, RoomId, RoomVersionId, UserId,
+    EventId, MilliSecondsSinceUnixEpoch, MxcUri, OwnedEventId, OwnedMxcUri, OwnedRoomAliasId,
+    OwnedRoomId, OwnedUserId, RoomAliasId, RoomId, RoomVersionId, UserId,
 };
 use serde::{Deserialize, Serialize};
 use tracing::{debug, field::debug, info, instrument, warn};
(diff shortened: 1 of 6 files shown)


crates/matrix-sdk-base/src/room/room_info.rs
```diff
@@ -464,6 +464,13 @@ pub struct RoomInfo {
     /// more accurate than relying on the latest event.
     #[serde(default)]
     pub(crate) recency_stamp: Option<u64>,
+
+    /// A timestamp remembering when we observed the user accepting an invite on
+    /// this current device.
+    ///
+    /// This is useful to remember if the user accepted this a join on using
+    /// this specific client.
```
stefanceriu: Something's off here, needs rewording I reckon.

crates/matrix-sdk-base/src/room/room_info.rs
```diff
@@ -464,6 +464,13 @@ pub struct RoomInfo {
...
+    pub(crate) invite_accepted_at: Option<MilliSecondsSinceUnixEpoch>,
```
stefanceriu, line 474: I never do know how these work: does it need a serde default or some special migration to not break old instances?

stefanceriu (approved):
Makes sense to me 👍

crates/matrix-sdk-base/src/room/room_info.rs
poljar: No: https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=dfb0cddf843b3f7eb2bda51fe094b5e1

A missing field or a `null` are treated the same if `Option` is used. I guess we could add a `skip_serializing` here since this field will in most cases be `None`.

stefanceriu: > No: https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=dfb0cddf843b3f7eb2bda51fe094b5e1

Heh, nice, thank you! 👍

crates/matrix-sdk-base/src/room/room_info.rs
poljar: Just a leftover from the previous version of this sentence.

PR #5337, opened 2025-07-02
Bump Ruma (breaking)

zecakeh:
This is ~~more or less~~ ready, with all the changes that are required due to breaking changes in Ruma. ~~I still need to fix a few tests. I will probably open other PRs soon to try to make this slightly smaller (but there is not much to do).~~

I tried to split the commits by source of the change, so I recommend to review this commit by commit. However after the first commit the code doesn't compile unless all the other commits are applied, so I recommend to squash everything when merging.

~~Note that currently Ruma is not ready for a release with those breaking changes, however it should be ready in a few weeks.~~ Ruma should be ready for release when the support for [room version 12](https://matrix.org/blog/2025/07/security-predisclosure/) is added in the next few days.

This is supposed to be part of 3 PRs:

1. This one, that bumps Ruma to a commit on main that has the same features that we already use in the SDK, while having the least amount of changes possible. (probably the biggest one).
2. Bump to the latest commit before support for room version 12. Will have ~300 lines changed.
3. Bump to the latest commit with support for room version 12.

### DIFF
--- a/bindings/matrix-sdk-ffi/CHANGELOG.md
+++ b/bindings/matrix-sdk-ffi/CHANGELOG.md
@@ -15,6 +15,13 @@ All notable changes to this project will be documented in this file.
 
 ### Breaking changes:
 
+- The `reason` argument of `Room::report_room` is now required, do to a clarification in the spec.
+  ([#5337](https://github.com/matrix-org/matrix-rust-sdk/pull/5337))
+- `PublicRoomJoinRule` has more variants, supporting all the known values from the spec.
+  ([#5337](https://github.com/matrix-org/matrix-rust-sdk/pull/5337))
+- The fields of `MediaPreviewConfig` are both optional, allowing to use the type for room account
+  data as well as global account data.
+  ([#5337](https://github.com/matrix-org/matrix-rust-sdk/pull/5337))
 - The `event_id` field of `PredecessorRoom` was removed, due to its removal in the Matrix
   specification with MSC4291.
   ([#5419](https://github.com/matrix-org/matrix-rust-sdk/pull/5419))
--- a/bindings/matrix-sdk-ffi/CHANGELOG.md
+++ b/bindings/matrix-sdk-ffi/CHANGELOG.md
@@ -15,6 +15,9 @@ All notable changes to this project will be documented in this file.
 
 ### Breaking changes:
 
+- `RoomPreview::info()` doesn't return a result anymore. All unknown join rules are handled in the
+  `JoinRule::Custom` variant.
+  ([#5337](https://github.com/matrix-org/matrix-rust-sdk/pull/5337))
 - The `reason` argument of `Room::report_room` is now required, do to a clarification in the spec.
   ([#5337](https://github.com/matrix-org/matrix-rust-sdk/pull/5337))
 - `PublicRoomJoinRule` has more variants, supporting all the known values from the spec.
--- a/bindings/matrix-sdk-ffi/src/room_preview.rs
+++ b/bindings/matrix-sdk-ffi/src/room_preview.rs
@@ -1,10 +1,10 @@
 use anyhow::Context as _;
 use matrix_sdk::{room_preview::RoomPreview as SdkRoomPreview, Client};
-use ruma::{room::RoomType as RumaRoomType, space::SpaceRoomJoinRule};
+use ruma::room::{JoinRuleSummary, RoomType as RumaRoomType};
 use tracing::warn;
 
 use crate::{
-    client::JoinRule,
+    client::{AllowRule, JoinRule},
     error::ClientError,
     room::{Membership, RoomHero},
     room_member::{RoomMember, RoomMemberWithSenderInfo},
@@ -40,7 +40,7 @@ impl RoomPreview {
                 .as_ref()
                 .map(TryInto::try_into)
                 .transpose()
-                .map_err(|_| anyhow::anyhow!("unhandled SpaceRoomJoinRule kind"))?,
+                .map_err(|_| anyhow::anyhow!("unhandled JoinRuleSummary kind"))?,
             is_direct: info.is_direct,
             heroes: info
                 .heroes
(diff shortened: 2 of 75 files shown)


testing/matrix-sdk-test/src/event_factory.rs
```diff
@@ -232,7 +232,12 @@ impl<E: StaticEventContent> EventBuilder<E> {
         self.state_key = Some(state_key.into());
         self
     }
+}
 
+impl<E: StaticEventContent> EventBuilder<E>
+where
+    E: Serialize,
+{
```
jplatte: This looks a bit weird. I'd move both bounds on `E` to the same place (probably the where bound).

zecakeh: Changed.

crates/matrix-sdk-common/src/deserialized_responses.rs
```diff
@@ -559,7 +562,9 @@ impl TimelineEvent {
                             // information around.
                             return Some(Box::new(TimelineEvent::from_decrypted(
                                 DecryptedRoomEvent {
-                                    event: latest_event,
+                                    // Safety: A decrypted event always includes a room_id in its
+                                    // payload.
```
jplatte, line 566: I wouldn't put `Safety:` if it's not about `unsafe` code.

crates/matrix-sdk/src/authentication/oauth/qrcode/rendezvous_channel.rs
```diff
@@ -114,11 +114,18 @@ impl RendezvousChannel {
...
 
         let request = create_rendezvous_session::unstable::Request::default();
         let response = client
-            .send(request, None, rendezvous_server.to_string(), None, &[], Default::default())
+            .send(
+                request,
+                None,
+                rendezvous_server.to_string(),
+                None,
+                &SupportedVersions { versions: Default::default(), features: Default::default() },
```
jplatte, line 126: Should we have an associated `const` for empty `SupportedVersions` upstream?

zecakeh: In this case, it looks to me like we should call `/versions` first, otherwise when this becomes stable we won't know which version of the path to call.

crates/matrix-sdk-common/src/deserialized_responses.rs
zecakeh: I copied this comment from other places in the code that do the same thing.

jplatte:
I think you can rebase with `-Xours` (or `-Xtheirs`? whichever one uses your changes) and `cargo fmt` after to have an automated clean rebase.

crates/matrix-sdk-common/src/deserialized_responses.rs
bnjbvr: I think that since https://github.com/ruma/ruma/issues/2061 (thanks, btw!), the `.cast()` can only happen when it's safe to do so, hence we could probably remove all those pseudo-safety comments, right?

zecakeh: This bumps Ruma to a commit before that PR (because that PR adds a lot more changes). Those `.cast()` will become `.cast_unchecked()` because it is not safe to `.cast()` from a sync to a non-sync type, so the comment will still apply.

crates/matrix-sdk/src/authentication/oauth/qrcode/rendezvous_channel.rs
poljar: Can't we use the version when support for this endpoint initially was added as the fallback? I guess there's nothing wrong with calling `/versions` either since this side will know the homeserver URL anyways.

crates/matrix-sdk/src/client/builder/mod.rs
```diff
@@ -101,7 +101,7 @@ pub struct ClientBuilder {
     store_config: BuilderStoreConfig,
     request_config: RequestConfig,
     respect_login_well_known: bool,
-    server_versions: Option<Box<[MatrixVersion]>>,
+    server_versions: Option<BTreeSet<MatrixVersion>>,
```
poljar, line 104: Nice, makes much more sense than an array.

bindings/matrix-sdk-ffi/src/room_preview.rs
```diff
@@ -122,20 +122,32 @@ pub struct RoomPreviewInfo {
...
+            },
+            JoinRuleSummary::KnockRestricted(summary) => JoinRule::KnockRestricted {
+                rules: summary
+                    .allowed_room_ids
+                    .iter()
+                    .map(|room_id| AllowRule::RoomMembership { room_id: room_id.to_string() })
+                    .collect(),
+            },
+            JoinRuleSummary::Public => JoinRule::Public,
+            JoinRuleSummary::_Custom(_) => JoinRule::Custom { repr: join_rule.as_str().to_owned() },
```
poljar: We're not supposed to use `_Custom()` this is supposed to be the wildcard one, or did something change in that regard?

poljar (approved):
Nice, thank you very much for doing the bump. Bonus kudos for the pedantic changelog entries.

I left a question about the usage of the `_Custom` variant.

A couple of merge conflicts have snuck in. Could you also please get rid of them.

bindings/matrix-sdk-ffi/src/room_preview.rs
zecakeh: We are not supposed to indeed. What do you prefer, keeping the `TryFrom` implementation without the `Custom` variant, or keeping the `Custom` variant and replacing this with a `From` implementation?

crates/matrix-sdk/src/authentication/oauth/qrcode/rendezvous_channel.rs
zecakeh: Ruma doesn't support that currently, only the unstable path is used if no supported versions are provided. See https://github.com/ruma/ruma/issues/2063.

bindings/matrix-sdk-ffi/src/room_preview.rs
poljar: Wouldn't a `From` need a panic for the `_` case? If so, let's keep the `TryFrom` and throw an error if something unsupported appears.

zecakeh: If we keep the `Custom` variant from the FFI type, we can just do `repr: join_rule.as_str().to_owned()`, like the current version.

poljar: Ah right, there's a `Custom` on the FFI side. Yeah, let's convert this to a `From` then.

zecakeh: Done.

bindings/matrix-sdk-ffi/CHANGELOG.md
```diff
@@ -15,6 +15,8 @@ All notable changes to this project will be documented in this file.
 
 ### Breaking changes:
 
+- `RoomPreview::info()` doesn't return a result anymore. All unknown join rules are handled in the
+  `JoinRule::Custom` variant.
```
poljar, line 19: PR number please. Looks good otherwise.

poljar (changes requested):

PR #5338, opened 2025-07-02
feat: return the notification status in the `NotificationClient` APIs

bnjbvr:
This helps distinguishing two cases:

- an event couldn't be found
- an event has been filtered out (because of push rules, or because the sending user is ignored)

Also a collection of tiny refactorings around the notification client, that in my opinion make the code slightly simpler to understand :)

### DIFF
--- a/bindings/matrix-sdk-ffi/src/notification.rs
+++ b/bindings/matrix-sdk-ffi/src/notification.rs
@@ -1,8 +1,8 @@
 use std::{collections::HashMap, sync::Arc};
 
 use matrix_sdk_ui::notification_client::{
-    NotificationClient as SdkNotificationClient, NotificationItem as SdkNotificationItem,
-    NotificationStatus as SdkNotificationStatus,
+    NotificationClient as SdkNotificationClient, NotificationEvent as SdkNotificationEvent,
+    NotificationItem as SdkNotificationItem, NotificationStatus as SdkNotificationStatus,
 };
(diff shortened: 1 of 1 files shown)


stefanceriu (approved):
Nice, lgtm 👍

bnjbvr:
Looks like this will need a change over [complement-crypto](https://github.com/matrix-org/matrix-rust-sdk/actions/runs/16053977564/job/45303593861#step:14:29) too :smiling_face_with_tear:
May have to wait for next week, if not later.

bnjbvr:
Complement-crypto fix: https://github.com/matrix-org/complement-crypto/pull/198

bindings/matrix-sdk-ffi/src/notification.rs
```diff
@@ -142,26 +181,22 @@ impl NotificationClient {
     pub async fn get_notifications(
         &self,
         requests: Vec<NotificationItemsRequest>,
-    ) -> Result<HashMap<String, NotificationItem>, ClientError> {
+    ) -> Result<HashMap<String, BatchNotificationResult>, ClientError> {
```
bmarty, line 184: Integration note: documentation is outdated.

bnjbvr: Thanks, addressed in https://github.com/matrix-org/matrix-rust-sdk/pull/5359

PR #5340, opened 2025-07-02
feat(sdk): Introduce the new `LatestEvents API`, part 2: Computing from the Event Cache

Hywan:
In this patch, `LatestEventValue` gains more variants to represent all kind of possible latest event values, like `RoomMessage`, `CallInvite`, `KnockedStateEvent`.

The flow is the following:

- `compute_latest_events_task` is run when rooms have received an update and the latest event must be recalculated, it calls `compute_latest_events`,
- `compute_latest_events` calls `RoomLatestEvents::update` for each room that must be updated,
- `RoomLatestEvents::update` calls `LatestEventValue::update` for the latest event of the room and of each thread (that is listened),
- Ultimately, `LatestEventValue::update` is called, which will call the `LatestEventValue::new_with_power_levels` constructor, and **will set** the new value :tada:,
- This constructor will use the `RoomEventCache` to iterate on the events backwards (from the most recent to the oldest), and will `find_and_map` the best latest event candidate value.

That's the flow and that's almost anything we need to know to understand this patch. However, one can ask themself:

- Why having two constructors for `LatestEventValue`? There is `new` and `new_with_power_levels`. The `RoomPowerLevels` are necessary to calculate the `LatestEventValue::KnockedStateEvent` variant. `new` will fetch the `RoomPowerLevels` itself, whilst `new_with_power_levels` receives a `&RoomPowerLevels`. It's an optimisation for `RoomLatestEvents::update` which calculates the `PowerRoomLevels` once for the room and the thread, it saves the need to recalculate them for each call to `LatestEventValue::update`.
- Why `WeakClient` and `WeakRoom`? Because if the `Client` is dropped, we are sure these tasks will error and will terminate.
- Why `LatestEvent` doesn't hold an instance of `RoomEventCache` but receives it by reference in `update`? Because `RoomLatestEvents` is responsible to hold `RoomEventCache`; plus it creates a smaller structure, so storing `LatestEvent` takes less space in memory.

There are a lot of new tests, but `test_latest_event_value_is_updated_via_event_cache` is the most fun as it uses `listen_and_subscribe_to_room` and runs the full workflow via the sync.

---

- Sequel of https://github.com/matrix-org/matrix-rust-sdk/pull/5298

### DIFF
--- a/crates/matrix-sdk/src/latest_events/latest_event.rs
+++ b/crates/matrix-sdk/src/latest_events/latest_event.rs
@@ -81,12 +81,7 @@ impl LatestEvent {
         )
         .await;
 
-        if let LatestEventValue::None = new_value {
-            // The new value is `None`. It means no new value has been
-            // computed. Let's keep the old value.
-        } else {
-            self.value.set(new_value).await;
-        }
+        self.value.set(new_value).await;
     }
 }
 


Hywan:
Thanks for the fast review.

PR #5341, opened 2025-07-02
Update README to make Element sponsorship explicit

neilisfragile:


poljar (changes requested):
Now I'm not a designer, but something feels a bit off here :sweat_smile:.

image

It would be nice if the new image respects the background color/theme github is using, like the main logo does. It does feel a bit too big as well.

README.md
```diff
@@ -18,11 +18,27 @@
...
+<picture>
+  <source srcset="contrib/element-logo-light.png" media="(prefers-color-scheme: dark)">
+  <source srcset="contrib/element-logo-dark.png" media="(prefers-color-scheme: light)">
+  <img src="contrib/element-logo-fallback.png" alt="Element logo">
+</picture>
+</div>
+<br><br>
+Development of the SDK is proudly sponsored and maintained by [Element](https://element.io). Element uses the SDK in their next-generation mobile apps Element X on [iOS](https://github.com/element-hq/element-x-ios) and [Android](https://github.com/element-hq/element-x-android) and has plans to introduce it to the web and desktop clients as well.
+
+The SDK is also the basis for multiple Matrix projects and we welcome contributions from all.
```
poljar: I think that this might look still better if we center the whole section, or at least the first paragraph. As it is now, we have a left/center/left scenario making the logo still look a bit out of place.

neilisfragile: Yeah I like that, done

PR #5345, opened 2025-07-03
feat(state store): diagnose and send to sentry deserialization errors

bnjbvr:
There's a bad regression hiding, where some data can't be deserialized from the state store, and we have no clues what caused it. This kind of diagnosis should help figuring out what it is, by (1) logging an error with extra information in the rageshake, (2) sending that error to sentry, when it's been configured at the FFI / tracing layer.

### DIFF
--- a/crates/matrix-sdk-sqlite/src/state_store.rs
+++ b/crates/matrix-sdk-sqlite/src/state_store.rs
@@ -390,7 +390,26 @@ impl SqliteStateStore {
 
     fn deserialize_json<T: DeserializeOwned>(&self, data: &[u8]) -> Result<T> {
         let decoded = self.decode_value(data)?;
-        Ok(serde_json::from_slice(&decoded)?)
+
+        let json_deserializer = &mut serde_json::Deserializer::from_slice(&decoded);
+
+        serde_path_to_error::deserialize(json_deserializer).map_err(|err| {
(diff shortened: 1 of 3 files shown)


crates/matrix-sdk-sqlite/src/state_store.rs
```diff
@@ -395,25 +395,25 @@ impl SqliteStateStore {
...
                 sentry = true,
                 %err,
-                maybe_room_id = ?room_id,
-                maybe_event_id = ?event_id,
                 "Failed to deserialize {target_type} in the state state: {serde_path}",
             );
 
+            if let Some(raw) = raw_json {
+                if let Some(room_id) = raw.get_field::<OwnedRoomId>("room_id").ok().flatten() {
+                    warn!("Found a room id in the source data to deserialize: {room_id}");
```
stefanceriu, line 410: Maybe we want these on the same log level as the error?

bnjbvr: There are not "errors" per se, and they would always be logged, since we do usually log everything at the `WARN` level. No strong opinions here, I'm tempted to keep it like this.

bnjbvr: (Said in private he's fine with it :))

PR #5346, opened 2025-07-03
refactor(sqlite): share the encode/decode/serialize/deserialize sqlite store helpers

bnjbvr:
These were duplicated for each SQLite store implementation, and with #5345, we'll get better error reporting upon deserialization; I figured we could have all the stores benefit from it, instead.


poljar (approved):
Neat.

PR #5349, opened 2025-07-04
refactor(test): Change the bound on `EventBuilder` to `StaticEventContent`

zecakeh:
The `EventContent` trait is gone in Ruma so this will ease the upgrade to the next breaking release.


stefanceriu (approved):
lgtm, thank you Kevin! 👍

PR #5358, opened 2025-07-07
fix(event cache): add more tests for threads in the timeline and event caches

bnjbvr:
These tests found some small incongruities. Testing is good, actually, you know.

Followup to #5123, part of #4869.


Hywan (approved):
Looks good to me. It fixes real problems, let's not waste time, kudos for the tests!

PR #5359, opened 2025-07-07
chore: update outdated FFI doc and add missing changelog entry

bnjbvr:
What it says on the tin. 

Also drive-by fixes a doctest build failure that made its way in `main`.


Hywan (approved):
Thanks!

PR #5361, opened 2025-07-07
fix(timeline): reload the related events for the initial thread events

bnjbvr:
Another test that I was pretty sure would fail. I was *not* wrong.

### DIFF
--- a/crates/matrix-sdk-ui/src/timeline/controller/mod.rs
+++ b/crates/matrix-sdk-ui/src/timeline/controller/mod.rs
@@ -380,11 +380,32 @@ impl<P: RoomDataProvider, D: Decryptor> TimelineController<P, D> {
                 let (events, _) = room_event_cache.subscribe_to_thread(root_event_id.clone()).await;
                 let has_events = !events.is_empty();
 
+                // For each event, we also need to find the related events, as they don't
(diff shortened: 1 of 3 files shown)


crates/matrix-sdk-ui/src/timeline/controller/mod.rs
```diff
@@ -380,12 +380,33 @@ impl<P: RoomDataProvider, D: Decryptor> TimelineController<P, D> {
                 let (events, _) = room_event_cache.subscribe_to_thread(root_event_id.clone()).await;
                 let has_events = !events.is_empty();
 
+                // For each event, we also need to find the related events, as they don't
+                // include the thread relationship, they won't be included in
+                // the initial list of events.
+                let mut related_events = Vector::new();
+                for event_id in events.iter().filter_map(|event| event.event_id()) {
+                    if let Some((_original, related)) =
+                        room_event_cache.find_event_with_relations(&event_id, None).await
```
Hywan, line 389: Does it look over the network? If yes, it can be pretty quickly annoying.

bnjbvr: No, the `await` is only caused because it will look up in the storage, but not network.

Hywan (approved):

PR #5363, opened 2025-07-08
feat(sdk): Add `RoomEventCacheGenericUpdate::Clear`

Hywan:
This patch adds the `RoomEventCacheGenericUpdate::Cleared` variant. It's helpful to know when a room has been cleared.

This is particularly useful for https://github.com/matrix-org/matrix-rust-sdk/pull/5340 to recalculate the `LatestEvent` when the `EventCache` clears a room.

---


poljar (approved):
Makes sense, but both of those commits are breaking changes and need to be added to the changelog.

PR #5367, opened 2025-07-08
Upgrade Ruma

zecakeh:
This is a bugfix for the `compat-lax-room-create-deser` and `compat-lax-room-topic-deser` features, which didn't work with events in the wild.


jplatte (approved):

poljar (approved):
Thanks for swiftly fixing this and re-relasing Ruma.

PR #5369, opened 2025-07-08
feat(ui): `RoomListService::subscribe_to_rooms` calls `LatestEvents::listen_to_room`

Hywan:
This patch updates `RoomListService::subscribe_to_rooms` to call `LatestEvents::listen_to_room` automatically. This method becomes async, which propagates to a couple of callers.

The idea is that when one is interested by a specific room, a subscription will be applied. This is an opportunity to also “activate” the computation of the `LatestEvent` for this specific room, so that the user doesn't have to do that manually (except if room subscription is never used).

---

### DIFF
--- a/bindings/matrix-sdk-ffi/src/room_list.rs
+++ b/bindings/matrix-sdk-ffi/src/room_list.rs
@@ -121,7 +121,7 @@ impl RoomListService {
         })))
     }
 
-    fn subscribe_to_rooms(&self, room_ids: Vec<String>) -> Result<(), RoomListError> {
+    async fn subscribe_to_rooms(&self, room_ids: Vec<String>) -> Result<(), RoomListError> {
         let room_ids = room_ids
             .into_iter()
             .map(|room_id| {
@@ -129,7 +129,9 @@ impl RoomListService {
             })
             .collect::<Result<Vec<_>, _>>()?;
 
-        self.inner.subscribe_to_rooms(&room_ids.iter().map(AsRef::as_ref).collect::<Vec<_>>());
+        self.inner
+            .subscribe_to_rooms(&room_ids.iter().map(AsRef::as_ref).collect::<Vec<_>>())
+            .await;
 
         Ok(())
     }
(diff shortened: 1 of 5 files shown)


bindings/matrix-sdk-ffi/src/room_list.rs
```diff
@@ -121,15 +121,17 @@ impl RoomListService {
         })))
     }
 
-    fn subscribe_to_rooms(&self, room_ids: Vec<String>) -> Result<(), RoomListError> {
+    async fn subscribe_to_rooms(&self, room_ids: Vec<String>) -> Result<(), RoomListError> {
```
poljar, line 124: We have a changelog for the bindings nowadays as well, so please document that this changed semantics and has become async.

crates/matrix-sdk-ui/src/room_list_service/mod.rs
```diff
@@ -393,7 +393,15 @@ impl RoomListService {
...
-    pub fn subscribe_to_rooms(&self, room_ids: &[&RoomId]) {
+    ///
+    /// [`LatestEvents::listen_to_room`][listen_to_room] will be called for each
+    /// room in `room_ids`, so that the [`LatestEventValue`] will automatically
+    /// be calculated and updated for these rooms, for free.
+    ///
+    /// [listen_to_room]: matrix_sdk::latest_events::LatestEvents::listen_to_room
+    /// [`LatestEventValue`]: matrix_sdk::latest_events::LatestEventValue
+    pub async fn subscribe_to_rooms(&self, room_ids: &[&RoomId]) {
+        // Calculate the settings for the room subscriptions.
```
poljar, line 404: This warrants a changelog entry as well.

bindings/matrix-sdk-ffi/src/room_list.rs
Hywan: Done.

crates/matrix-sdk-ui/src/room_list_service/mod.rs
Hywan: Done.

PR #5370, opened 2025-07-08
feat(sdk): Introduce `SendQueue::subscribe`

Hywan:
To get updates from the `SendQueue`, one needs to use `RoomSendQueue::subscribe`, it emits `RoomSendQueueUpdate`s. However, if one wants to receive updates for all rooms managed by the `SendQueue`, instead of subscribing to all `RoomSendQueue` individually, this patch introduces a new `SendQueue::subscribe` method, which emits `SendQueueUpdate`s. It's basically a wrapper around `RoomSendQueueUpdate` by adding an `OwnedRoomId`.

---


poljar (approved):
Nice hack to avoid writing new tests :sweat_smile:

PR #5376, opened 2025-07-09
feat(sdk): Add the `request_duration` log field in `HttpClient::send_request`

Hywan:
This patch adds a new log field, named `request_duration` in `HttpClient::send_request`. It helps to know how much time the request took to be executed.


crates/matrix-sdk/src/http_client/native.rs
```diff
@@ -76,15 +76,18 @@ impl HttpClient {
             async {
                 let num_attempt = retry_count.fetch_add(1, Ordering::SeqCst);
                 debug!(num_attempt, "Sending request");
+                let before = ruma::time::Instant::now();
```
bnjbvr, line 79: Why not std::time::Instant?

crates/matrix-sdk/src/http_client/native.rs
```diff
@@ -76,15 +76,18 @@ impl HttpClient {
...
                     send_request(&self.inner, &request, config.timeout, send_progress).await?;
 
+                let request_duration = ruma::time::Instant::now().saturating_duration_since(before);
                 let status_code = response.status();
                 let response_size = ByteSize(response.body().len().try_into().unwrap_or(u64::MAX));
                 tracing::Span::current()
                     .record("status", status_code.as_u16())
-                    .record("response_size", response_size.display().si_short().to_string());
+                    .record("response_size", response_size.display().si_short().to_string())
+                    .record("request_duration", tracing::field::debug(request_duration));
```
bnjbvr, line 91: Do we need the `debug` wrapper here? I've seen contexts where numbers were accepted as plain tracing values, and didn't need any display/debug modifier.

bnjbvr (approved):
To clarify the intent: this PR makes it easier to spot the duration of a given request, since the information is already available (there are usually two log entries for a request start and a request end, and both should include the time).

Thanks!

crates/matrix-sdk/src/http_client/native.rs
Hywan: Because of wasm32 support. `ruma` automatically exports `web-time` which manages this fallback.

crates/matrix-sdk/src/http_client/native.rs
Hywan: This isn't a number but a `Duration` :-).

PR #5378, opened 2025-07-09
feat(ui): Add `RoomListService::new_with_share_pos`

Hywan:
This patch adds the new `RoomListService::new_with_share_pos` constructor. It decides whether the `share_pos` feature of sliding sync should be enabled or not.

`SyncServiceBuilder` gains a new `with_share_pos` method to configure the way the `RoomListService` is built.

The FFI bindings are updated accordingly.

### DIFF
--- a/bindings/matrix-sdk-ffi/src/sync_service.rs
+++ b/bindings/matrix-sdk-ffi/src/sync_service.rs
@@ -119,6 +119,12 @@ impl SyncServiceBuilder {
         Arc::new(Self { builder, ..this })
     }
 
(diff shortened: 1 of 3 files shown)


poljar (approved):
I was for a bit confused where the logic to turn this on/off is, the sneaky `if` branch wasn't immediately clear from the diff.

Looks sensible. @jmartinesp care to try it out? Though there's nothing controversial, we can also just merge if you don't have the time.

jmartinesp:
Sure, I'll take a look now.

Hywan:
Yes, the previous value. Thanks for the test!

PR #5380, opened 2025-07-09
FFI: Allow tracing's RollingFileAppender to be updated at runtime.

### LINKED ISSUE #5375: Allow tracing to be reconfigured with a different log path a runtime.
There are circumstances when an app (that doesn't support multi-account) would find it useful to segregate its log files based on its `Client`'s user and/or homeserver, such as making it easier to manage which logs are selected to send in a particular rageshake.

Currently there is an [init_platform](https://github.com/matrix-org/matrix-rust-sdk/blob/54e6a7d8d1bb7fde3dae12e5a18825c7a846fd8d/bindings/matrix-sdk-ffi/src/platform.rs#L531) function to configure tracing in platform.rs which sets this [OnceLock](https://github.com/matrix-org/matrix-rust-sdk/blob/54e6a7d8d1bb7fde3dae12e5a18825c7a846fd8d/bindings/matrix-sdk-ffi/src/platform.rs#L366) as tracing can only be configured once. However from what I can tell, a potential route to implement this is by wrapping the [`file_layer`](https://github.com/matrix-org/matrix-rust-sdk/blob/54e6a7d8d1bb7fde3dae12e5a18825c7a846fd8d/bindings/matrix-sdk-ffi/src/platform.rs#L133) inside a [reload layer](https://docs.rs/tracing-subscriber/latest/tracing_subscriber/reload/index.html) in order to dynamically switch out the `RollingFileAppender` for a new one with the desired path.

pixlwave:
As described in #5375, this PR allows apps to update the directory used to write log files at runtime, so that logs could be segregated by e.g. user/homeserver etc.

Closes #5375

### DIFF
--- a/bindings/matrix-sdk-ffi/src/platform.rs
+++ b/bindings/matrix-sdk-ffi/src/platform.rs
@@ -14,164 +14,181 @@ use tracing_subscriber::{
         time::FormatTime,
         FormatEvent, FormatFields, FormattedFields,
     },
-    layer::SubscriberExt as _,
+    layer::{Layered, SubscriberExt as _},
     registry::LookupSpan,
+    reload::{self, Handle},
     util::SubscriberInitExt as _,
-    Layer,
+    EnvFilter, Layer, Registry,
 };
 
 use crate::{error::ClientError, tracing::LogLevel};
 
-fn text_layers<S>(config: TracingConfiguration) -> impl Layer<S>
+// Adjusted version of tracing_subscriber::fmt::Format
+struct EventFormatter {
+    display_timestamp: bool,
+    display_level: bool,
+}
+
+impl EventFormatter {
+    fn new() -> Self {
+        Self { display_timestamp: true, display_level: true }
+    }
+
+    #[cfg(target_os = "android")]
+    fn for_logcat() -> Self {
+        // Level and time are already captured by logcat separately
+        Self { display_timestamp: false, display_level: false }
+    }
+
+    fn format_timestamp(&self, writer: &mut fmt::format::Writer<'_>) -> std::fmt::Result {
+        if fmt::time::SystemTime.format_time(writer).is_err() {
+            writer.write_str("<unknown time>")?;
+        }
+        Ok(())
+    }
+
+    fn write_filename(
+        &self,
+        writer: &mut fmt::format::Writer<'_>,
+        filename: &str,
+    ) -> std::fmt::Result {
+        const CRATES_IO_PATH_MATCHER: &str = ".cargo/registry/src/index.crates.io";
+        let crates_io_filename = filename
+            .split_once(CRATES_IO_PATH_MATCHER)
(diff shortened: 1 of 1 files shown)


bindings/matrix-sdk-ffi/src/platform.rs
```diff
@@ -611,6 +643,33 @@ fn setup_lightweight_tokio_runtime() {
     }));
 }
 
+#[matrix_sdk_ffi_macros::export]
+pub fn reload_tracing_file_writer(
```
pixlwave: I was split whether this should have a more generic name such as `update_platform(file_configuration:)` that would align closer with `init_platform`, but went with this for now. Happy to update if anyone has a preference :)

Also realising this needs a doc to say you must call `init_platform` first (especially if not renamed).

pixlwave:
<del>It's maybe worth noting I've gone for the simpler solution here of another global function rather than directly returning an optional reload handle from `init_platform`, which might be a nicer API given you couldn't even try to change the file writer before calling `init_platform`, but seems to add more complexity for relatively little benefit.</del>

pixlwave:
Answering both of my previous questions here by seeing that `enable_sentry_logging` is a thing which is both a global function and isn't aligned in naming with `init_platform` 😄

bindings/matrix-sdk-ffi/src/platform.rs
```diff
@@ -14,164 +14,158 @@ use tracing_subscriber::{
...
+        &self,
+        writer: Writer<'writer>,
+        fields: R,
+    ) -> std::fmt::Result {
+        self.0.format_fields(writer, fields)
+    }
+}
 
-    Layer::and_then(
+type HandleType = Handle<
```
jplatte:
```suggestion
type ReloadHandle = Handle<
```

pixlwave: Makes sense, updated 👍

bindings/matrix-sdk-ffi/src/platform.rs
```diff
@@ -570,6 +602,38 @@ pub fn enable_sentry_logging(enabled: bool) {
...
+        });
+    };
+
+    let layer = make_file_layer(configuration);
+    reload_handle.reload(layer).map_err(|error| ClientError::Generic {
+        msg: format!("Failed to reload file config: {error}"),
+        details: None,
+    })?;
+
+    Ok(())
```
poljar: I think this is unnecessary.

```suggestion
    reload_handle.reload(layer).map_err(|error| ClientError::Generic {
        msg: format!("Failed to reload file config: {error}"),
        details: None,
    })
```

poljar (changes requested):
Looks good. I would have separated the refactoring into individual commits before adding the reload handle to ease up the review. The actual changes are quite minimal. But ok, the PR isn't that big itself.

Left a small nit, after that we can merge.

PR #5382, opened 2025-07-09
feat(sqlite): `SqliteEventCacheStore` has 1 write connection

Hywan:
Until now, `SqliteEventCacheStore` manages a pool of connections. A connection is fetched from this pool and operations are executed on it, regardless whether these are read operations or write operations.

We are seeing more and more _database is busy_ errors. We believe this is because too many write operations are executed concurrently.

The solution to solve this is to use multiple connections for read operations, and a single connection for write operations. That way, concurrent writings are no longer a thing, and we hope it will reduce the number of _database is busy_ errors to zero. That's our guess.

This patch does that. When the pool of connections is created, a connection is elected as the `write_connection`. To get a connection for read operations, one has to use the new `SqliteEventCacheStore::read` method (it replaces the `acquire` method). To get a connection for write operations, one has to use the new `SQliteEventCacheStore::write` method. It returns a `OwnedMutexGuard` from an async `Mutex`. All callers that want to do write operations on this store have to wait their turn, this `Mutex` is fair, and the first to wait on the lock is the first that will take the lock (FIFO). It guarantees the execution ordering the code expects.

The rest of the patch updates all spots where `acquire` was used and replaces them by `read()` or `write()`. A particular care was made to see if other places are using `SqliteEventCacheStore::pool` directly. No place remains except in `read()` and `write()`.

Finally, other commits are adding more logs and timer logs.

### DIFF
--- a/crates/matrix-sdk-sqlite/src/event_cache_store.rs
+++ b/crates/matrix-sdk-sqlite/src/event_cache_store.rs
@@ -47,7 +47,7 @@ use tokio::{
     fs,
     sync::{Mutex, OwnedMutexGuard},
 };
-use tracing::{debug, error, trace};
+use tracing::{debug, error, instrument, trace};
 
 use crate::{
     error::{Error, Result},
@@ -173,8 +173,13 @@ impl SqliteEventCacheStore {
     }
 
     // Acquire a connection for executing read operations.
+    #[instrument(skip_all)]
     async fn read(&self) -> Result<SqliteAsyncConn> {
+        trace!("Taking a `read` connection");
+
         let connection = self.pool.get().await?;
+
+        trace!("`read` connection taken");
 
         // Per https://www.sqlite.org/foreignkeys.html#fk_enable, foreign key
         // support must be enabled on a per-connection basis. Execute it every
@@ -186,8 +191,13 @@ impl SqliteEventCacheStore {
     }
 
     // Acquire a connection for executing write operations.
+    #[instrument(skip_all)]
     async fn write(&self) -> Result<OwnedMutexGuard<SqliteAsyncConn>> {
+        trace!("Taking a `write` connection");
+
         let connection = self.write_connection.clone().lock_owned().await;
+
+        trace!("`write` connection taken");
 
         // Per https://www.sqlite.org/foreignkeys.html#fk_enable, foreign key
         // support must be enabled on a per-connection basis. Execute it every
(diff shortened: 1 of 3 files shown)


Hywan:
Bonus: I'm thinking about it right now, but we may want to add logs around `write()` to know how much it took to acquire the write connection. It could be a very useful information to debug slowness if any. I'll do that later. Please don't merge this PR immediately.

crates/matrix-sdk-sqlite/src/event_cache_store.rs
```diff
@@ -1583,33 +1615,35 @@ async fn with_immediate_transaction<
...
-        // been run, whether it was successful or not.
-        conn.set_transaction_behavior(TransactionBehavior::Deferred);
-
-        res
-    })
-    .await
-    // SAFETY: same logic as in [`deadpool::managed::Object::with_transaction`].`
-    .unwrap()
+    this.write()
+        .await?
```
Hywan, line 1700: This change ensures `with_immediate_transaction` takes a write connection.

Hywan:
Quoting myself (because why not):

> Bonus: I'm thinking about it right now, but we may want to add logs around write() to know how much it took to acquire the write connection. It could be a very useful information to debug slowness if any. I'll do that later. Please don't merge this PR immediately.

Done. I've added `#[instrument]` + `timer!` logs, almost everywhere. It will help spot slowness more easily.

crates/matrix-sdk-sqlite/src/event_cache_store.rs
```diff
@@ -465,6 +465,8 @@ impl EventCacheStore for SqliteEventCacheStore {
         key: &str,
         holder: &str,
     ) -> Result<bool> {
+        let _timer = timer!("method");
```
poljar, line 468: Are we sure we want to put `method` everywhere?

poljar (approved):
Looks good, left a question.

crates/matrix-sdk-sqlite/src/event_cache_store.rs
Hywan: With `#[instrument]`, we already have all the information required. I really didn't know exactly what to put here. We are measuring the time for the whole method, I thought it was alright. Thought?

poljar: Sure.

Issue #5366, opened 2025-07-08
The `test_login_username_refresh_token()` method should use the `MatrixMockServer`


poljar:
Related to https://github.com/matrix-org/matrix-rust-sdk/issues/3716.

Many of the integration tests for the refresh token support still don't use the `MatrixMockServer`. A good starting point to kickstart the conversion would be the `test_login_username_refresh_token()` method:

https://github.com/matrix-org/matrix-rust-sdk/blob/a4e4b5534befc481b7c018aed9c8d60b5d138f56/crates/matrix-sdk/tests/integration/refresh_token.rs#L44-L69

The test can't be converted right of the bat, we first need to extend the `MatrixMockServer` a bit:

- [ ] The `LoginEndpoint` mock needs to be extended so we can ensure that the request body contains certain JSON fields.
- [ ] We need to be able to configure what the `LoginEndpoint` mock will respond with.

For the first part we can take a look at the `ReceiptEndpoint`:

https://github.com/matrix-org/matrix-rust-sdk/blob/a4e4b5534befc481b7c018aed9c8d60b5d138f56/crates/matrix-sdk/src/test_utils/mocks/mod.rs#L3172-L3176

The second point should add a parametrized variant of the `ok()` method we have:

https://github.com/matrix-org/matrix-rust-sdk/blob/a4e4b5534befc481b7c018aed9c8d60b5d138f56/crates/matrix-sdk/src/test_utils/mocks/mod.rs#L3229-L3232

After this, the test itself can be converted to utilize the `MatrixMockServer` with the newly introduced methods.

poljar:
This has been closed by #5383.

PR #5383, opened 2025-07-09
Fix issue #5366

dragonfly1033:
Change test_login_username_refresh_token to use MatrixMockServer.
Add builders for configurable reesponses for mock login endpoint.

Signed-off-by: 
Shrey Patel <shreyp@element.io>

### DIFF
--- a/crates/matrix-sdk/src/test_utils/mocks/mod.rs
+++ b/crates/matrix-sdk/src/test_utils/mocks/mod.rs
@@ -3231,9 +3231,85 @@ impl<'a> MockEndpoint<'a, LoginEndpoint> {
         self.respond_with(ResponseTemplate::new(200).set_body_json(&*test_json::LOGIN))
     }
 
-    /// Returns a successful response with a given message.
-    pub fn ok_with(self, response: ResponseTemplate) -> MatrixMock<'a> {
-        self.respond_with(response)
+    /// Returns a given response on POST /login requests
+    ///
+    /// This function
+    /// to-device messages sent via the `/sendToDevice` endpoint.
+    ///
+    /// # Arguments
+    ///
+    /// * `response` - The response that the mock server sends on POST /login
+    ///   requests.
+    ///
+    /// # Returns
+    ///
+    /// Returns a [`MatrixMock`] which can be mounted.
+    ///
+    /// # Examples
+    ///
+    /// ```rust
+    /// use matrix_sdk::test_utils::mocks::{
+    ///     LoginResponseTemplate200, MatrixMockServer,
+    /// };
+    /// use matrix_sdk_test::async_test;
+    /// use ruma::{device_id, time::Duration, user_id};
+    ///
+    /// #[async_test]
+    /// async fn test_ok_with() {
+    ///     let server = MatrixMockServer::new().await;
+    ///     server
+    ///         .mock_login()
+    ///         .ok_with(LoginResponseTemplate200::new(
+    ///             "qwerty",
+    ///             device_id!("DEADBEEF"),
+    ///             user_id!("@cheeky_monkey:matrix.org"),
+    ///         ))
+    ///         .mount()
+    ///         .await;
+    ///
+    ///     let client = server.client_builder().unlogged().build().await;
+    ///
+    ///     let result = client
+    ///         .matrix_auth()
+    ///         .login_username("example", "wordpass")
+    ///         .send()
+    ///         .await
+    ///         .unwrap();
+    ///
+    ///     assert!(
+    ///         result.access_tokesn.unwrap() == "qwerty",
+    ///         "wrong access token in response"
+    ///     );
+    ///     assert!(
+    ///         result.device_id.unwrap() == "DEADBEEF",
+    ///         "wrong device id in response"
+    ///     );
+    ///     assert!(
+    ///         result.user_id.unwrap() == "@cheeky_monkey:matrix.org",
+    ///         "wrong user id in response"
+    ///     );
+    /// }
+    /// ```
+    pub fn ok_with(self, response: LoginResponseTemplate200) -> MatrixMock<'a> {
+        self.respond_with(ResponseTemplate::new(200).set_body_json(json!({
+            "access_token": response.access_token,
+            "device_id": response.device_id,
+            "user_id": response.user_id,
+            "expires_in": response.expires_in.map(|duration| { duration.as_millis() }),
+            "refresh_token": response.refresh_token,
+            "well_known": response.well_known.map(|vals| {
+                json!({
+                    "m.homeserver": {
+                        "base_url": vals.homeserver_url
+                    },
+                    "m.identity_server": vals.identity_url.map(|url| {
+                        json!({
+                            "base_url": url
+                        })
+                    })
+                })
+            }),
+        })))
     }
 
     /// Ensures that the body of the request is a superset of the provided
@@ -3243,6 +3319,80 @@ impl<'a> MockEndpoint<'a, LoginEndpoint> {
     }
 }
 
+#[derive(Default)]
+struct LoginResponseWellKnown {
+    /// Required if well_known is used: The base URL for the homeserver for
+    /// client-server connections.
+    homeserver_url: String,
+
+    /// Required if well_known and m.identity_server are used: The base URL for
+    /// the identity server for client-server connections.
+    identity_url: Option<String>,
+}
+
+/// A response to a [`LoginEndpoint`] query with status code 200.
+#[derive(Default)]
+pub struct LoginResponseTemplate200 {
+    /// Required: An access token for the account. This access token can then be
+    /// used to authorize other requests.
+    access_token: Option<String>,
+
+    /// Required: ID of the logged-in device. Will be the same as the
+    /// corresponding parameter in the request, if one was specified.
+    device_id: Option<OwnedDeviceId>,
+
+    /// The lifetime of the access token, in milliseconds. Once the access token
+    /// has expired a new access token can be obtained by using the provided
+    /// refresh token. If no refresh token is provided, the client will need
+    /// to re-log in to obtain a new access token. If not given, the client
+    /// can assume that the access token will not expire.
+    expires_in: Option<Duration>,
+
+    /// A refresh token for the account. This token can be used to obtain a new
+    /// access token when it expires by calling the /refresh endpoint.
+    refresh_token: Option<String>,
+
+    /// Required: The fully-qualified Matrix ID for the account.
+    user_id: Option<OwnedUserId>,
+
+    /// Optional client configuration provided by the server.
+    well_known: Option<LoginResponseWellKnown>,
+}
+
+impl LoginResponseTemplate200 {
+    /// Constructor for empty response
+    pub fn new<T1: Into<OwnedDeviceId>, T2: Into<OwnedUserId>>(
+        access_token: &str,
+        device_id: T1,
+        user_id: T2,
+    ) -> Self {
+        Self {
+            access_token: Some(access_token.to_owned()),
+            device_id: Some(device_id.into()),
+            user_id: Some(user_id.into()),
+            ..Default::default()
+        }
+    }
+
+    /// sets expires_in
+    pub fn expires_in(mut self, value: Duration) -> Self {
+        self.expires_in = Some(value);
+        self
+    }
+
+    /// sets refresh_token
+    pub fn refresh_token(mut self, value: &str) -> Self {
+        self.refresh_token = Some(value.to_owned());
+        self
+    }
+
+    /// sets well_known which takes a homeserver_url and an optional
+    /// identity_url
+    pub fn well_known(mut self, homeserver_url: String, identity_url: Option<String>) -> Self {
+        self.well_known = Some(LoginResponseWellKnown { homeserver_url, identity_url });
+        self
+    }
+}
 
 /// A prebuilt mock for `GET /devices` requests.
 pub struct DevicesEndpoint;
--- a/crates/matrix-sdk/src/test_utils/mocks/mod.rs
+++ b/crates/matrix-sdk/src/test_utils/mocks/mod.rs
@@ -3230,7 +3230,19 @@ impl<'a> MockEndpoint<'a, LoginEndpoint> {
     pub fn ok(self) -> MatrixMock<'a> {
         self.respond_with(ResponseTemplate::new(200).set_body_json(&*test_json::LOGIN))
     }
+
+    /// Returns a successful response with a given message.
+    pub fn ok_with(self, response: ResponseTemplate) -> MatrixMock<'a> {
+        self.respond_with(response)
+    }
+
+    /// Ensures that the body of the request is a superset of the provided
+    /// `body` parameter.
+    pub fn body_matches_partial_json(self, body: Value) -> Self {
+        Self { mock: self.mock.and(body_partial_json(body)), ..self }
+    }
 }
+
 
 /// A prebuilt mock for `GET /devices` requests.
 pub struct DevicesEndpoint;
--- a/crates/matrix-sdk/tests/integration/refresh_token.rs
+++ b/crates/matrix-sdk/tests/integration/refresh_token.rs
@@ -11,8 +11,8 @@ use matrix_sdk::{
     executor::spawn,
     store::RoomLoadSettings,
     test_utils::{
-        client::mock_session_meta, logged_in_client_with_server, no_retry_test_client_with_server,
-        test_client_builder_with_server,
+        client::mock_session_meta, logged_in_client_with_server, mocks::MatrixMockServer,
+        no_retry_test_client_with_server, test_client_builder_with_server,
     },
     HttpError, RefreshTokenError, SessionChange, SessionTokens,
 };
@@ -43,18 +43,15 @@ fn session() -> MatrixSession {
 
 #[async_test]
 async fn test_login_username_refresh_token() {
-    let (client, server) = no_retry_test_client_with_server().await;
-
-    Mock::given(method("POST"))
-        .and(path("/_matrix/client/r0/login"))
-        .and(body_partial_json(json!({
-            "refresh_token": true,
-        })))
-        .respond_with(
-            ResponseTemplate::new(200).set_body_json(&*test_json::LOGIN_WITH_REFRESH_TOKEN),
-        )
-        .mount(&server)
+    let server = MatrixMockServer::new().await;
+    server
+        .mock_login()
+        .body_matches_partial_json(json!({"refresh_token": true}))
+        .ok_with(ResponseTemplate::new(200).set_body_json(&*test_json::LOGIN_WITH_REFRESH_TOKEN))
+        .mount()
         .await;
+
+    let client = server.client_builder().unlogged().build().await;
 
     let res = client
         .matrix_auth()
@@ -63,7 +60,7 @@ async fn test_login_username_refresh_token() {
         .send()
         .await
         .unwrap();
-
+    
     assert!(client.is_active(), "Client should be active");
     res.refresh_token.unwrap();
 }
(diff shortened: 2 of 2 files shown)


crates/matrix-sdk/tests/integration/refresh_token.rs
```diff
@@ -43,27 +43,24 @@ fn session() -> MatrixSession {
...
-        .respond_with(
-            ResponseTemplate::new(200).set_body_json(&*test_json::LOGIN_WITH_REFRESH_TOKEN),
-        )
-        .mount(&server)
+    let server = MatrixMockServer::new().await;
+    server
+        .mock_login()
+        .body_matches_partial_json(json!({"refresh_token": true}))
+        .ok_with(ResponseTemplate::new(200).set_body_json(&*test_json::LOGIN_WITH_REFRESH_TOKEN))
+        .mount()
```
poljar: I think a `ok()` method that uses the default resposne from `test_json` would be nicer here.

We can leave the `ok_with()` method as well, in case somebody needs a custom response.

crates/matrix-sdk/src/test_utils/mocks/mod.rs
```diff
@@ -3243,6 +3243,244 @@ impl<'a> MockEndpoint<'a, LoginEndpoint> {
...
+/// A struct for building a login response with status code 200
+#[derive(Default)]
+pub struct LoginResponseBuilder200 {
+    access_token: Option<String>, // required
+    device_id: Option<String>,   // required
+    expires_in_ms: Option<u32>,
+    refresh_token: Option<String>,
+    user_id: Option<String>,    // required
+    homeserver_url: Option<String>, // required if well_known used
+    identity_url: Option<String>, // required if well_known and m.identity_server used
```
poljar: It would be nice to document these.

crates/matrix-sdk/src/test_utils/mocks/mod.rs
```diff
@@ -3243,6 +3243,244 @@ impl<'a> MockEndpoint<'a, LoginEndpoint> {
     }
 }
 
+/// A struct for building a login response with status code 200
```
poljar: You can link to the method where this can be used with. Take a look at the relations response template:

https://github.com/matrix-org/matrix-rust-sdk/blob/72d133260c29e132a75597ae6ca77963813e74fa/crates/matrix-sdk/src/test_utils/mocks/mod.rs#L3115-L3117.

crates/matrix-sdk/src/test_utils/mocks/mod.rs
```diff
@@ -3243,6 +3243,244 @@ impl<'a> MockEndpoint<'a, LoginEndpoint> {
...
+#[derive(Default)]
+pub struct LoginResponseBuilder200 {
```
poljar: We call these concepts templates, to be consistent with wiremock `ResponseTemplate` lingo.

crates/matrix-sdk/src/test_utils/mocks/mod.rs
```diff
@@ -3243,6 +3243,244 @@ impl<'a> MockEndpoint<'a, LoginEndpoint> {
...
+    device_id: Option<String>,   // required
+    expires_in_ms: Option<u32>,
+    refresh_token: Option<String>,
+    user_id: Option<String>,    // required
+    homeserver_url: Option<String>, // required if well_known used
+    identity_url: Option<String>, // required if well_known and m.identity_server used
+}
+
+impl LoginResponseBuilder200 {
+    /// constructor for empty reponse
```
poljar: Docstrings are normal sentences, they should be capitalized as such.

crates/matrix-sdk/src/test_utils/mocks/mod.rs
```diff
@@ -3243,6 +3243,244 @@ impl<'a> MockEndpoint<'a, LoginEndpoint> {
...
+    expires_in_ms: Option<u32>,
+    refresh_token: Option<String>,
+    user_id: Option<String>,    // required
+    homeserver_url: Option<String>, // required if well_known used
+    identity_url: Option<String>, // required if well_known and m.identity_server used
+}
+
+impl LoginResponseBuilder200 {
+    /// constructor for empty reponse 
+    pub fn new() -> Self { Self::default() }
```
poljar: It would be nice if the constructor told us which things *must* be set. It might be sensible to pick some default values for some fields to leave them optional in the constructor.

crates/matrix-sdk/src/test_utils/mocks/mod.rs
```diff
@@ -3243,6 +3243,244 @@ impl<'a> MockEndpoint<'a, LoginEndpoint> {
...
+
+    /// sets access_token
+    pub fn set_access_token(&mut self, value: &str) -> &mut Self { 
+        self.access_token = Some(value.to_string()); 
+        self 
+    }
+    
+    /// sets device_id
+    pub fn set_device_id(&mut self, value: &str) -> &mut Self {
```
poljar: We have the `DeviceId` type for device IDs.

crates/matrix-sdk/src/test_utils/mocks/mod.rs
```diff
@@ -3243,6 +3243,244 @@ impl<'a> MockEndpoint<'a, LoginEndpoint> {
...
+    }
+    
+    /// sets device_id
+    pub fn set_device_id(&mut self, value: &str) -> &mut Self { 
+        self.device_id = Some(value.to_string()); 
+        self 
+    }
+    
+    /// sets expires_in_ms
+    pub fn set_expires_in_ms(&mut self, value: u32) -> &mut Self {
```
poljar: For values that contain time we use the `Duration` type, this removes the need to mention the unit in the function name.

crates/matrix-sdk/src/test_utils/mocks/mod.rs
```diff
@@ -3243,6 +3243,244 @@ impl<'a> MockEndpoint<'a, LoginEndpoint> {
...
+    }
+    
+    /// sets refresh_token
+    pub fn set_refresh_token(&mut self, value: &str) -> &mut Self { 
+        self.refresh_token = Some(value.to_string()); 
+        self 
+    }
+    
+    /// sets user_id
+    pub fn set_user_id(&mut self, value: &str) -> &mut Self {
```
poljar: The `UserId` type is for user IDs.

crates/matrix-sdk/src/test_utils/mocks/mod.rs
```diff
@@ -3243,6 +3243,244 @@ impl<'a> MockEndpoint<'a, LoginEndpoint> {
...
+    homeserver_url: Option<String>, // required if well_known used
+    identity_url: Option<String>, // required if well_known and m.identity_server used
+}
+
+impl LoginResponseBuilder200 {
+    /// constructor for empty reponse 
+    pub fn new() -> Self { Self::default() }
+
+    /// sets access_token
+    pub fn set_access_token(&mut self, value: &str) -> &mut Self {
```
poljar: Our naming convention avoids the usage of `set`, it's obvious that those will be setters.

poljar: There's two ways to create a builder pattern, one mutates `self` the other takes ownership of `self` and returns `self`.

We already have a similar template here, so let's be consistent and use the later:

https://github.com/matrix-org/matrix-rust-sdk/blob/72d133260c29e132a75597ae6ca77963813e74fa/crates/matrix-sdk/src/test_utils/mocks/mod.rs#L3136-L3160

crates/matrix-sdk/src/test_utils/mocks/mod.rs
```diff
@@ -3243,6 +3243,244 @@ impl<'a> MockEndpoint<'a, LoginEndpoint> {
...
+
+            if self.identity_url.is_some() {
+                let mut identity_server = Map::new();
+                identity_server.insert("base_url".to_string(), Value::String(self.identity_url.clone().unwrap()));
+                well_known.insert("m.identityserver".to_string(), Value::Object(identity_server));
+            }
+
+            body.insert("well_known".to_string(), Value::Object(well_known));
+        }
+
```
poljar: This is a very manual way of building the `json::Value` from the `ResponseTemplate`.

You could have just used `json!({})` like the `RoomRelationsResponseTemplate` does:

https://github.com/matrix-org/matrix-rust-sdk/blob/72d133260c29e132a75597ae6ca77963813e74fa/crates/matrix-sdk/src/test_utils/mocks/mod.rs#L3067-L3072

There is no difference between an unset field and a field being set to `None`, so you don't need to make this conditional at all.

crates/matrix-sdk/src/test_utils/mocks/mod.rs
```diff
@@ -3243,6 +3243,244 @@ impl<'a> MockEndpoint<'a, LoginEndpoint> {
...
+        if self.retry_after_ms.is_some() {
+            body.insert("retry_after_ms".to_string(), Value::Number(Number::from(self.retry_after_ms.clone().unwrap())));
+        }
+
+        ResponseTemplate::new(429).set_body_json(Value::Object(body))
+    }
+}
+
```
poljar: These aren't really used yet, so let's leave them out of the PR.

crates/matrix-sdk/tests/integration/refresh_token.rs
```diff
@@ -47,7 +47,16 @@ async fn test_login_username_refresh_token() {
     server
         .mock_login()
         .body_matches_partial_json(json!({"refresh_token": true}))
-        .ok_with(ResponseTemplate::new(200).set_body_json(&*test_json::LOGIN_WITH_REFRESH_TOKEN))
+        // .ok_with(ResponseTemplate::new(200).set_body_json(&*test_json::LOGIN_WITH_REFRESH_TOKEN))
```
poljar: We should remove this old code snippet instead of commenting it out.

crates/matrix-sdk/src/test_utils/mocks/mod.rs
```diff
@@ -3243,6 +3243,244 @@ impl<'a> MockEndpoint<'a, LoginEndpoint> {
...
+    }
+    
+    /// sets identity_url
+    pub fn set_identity_url(&mut self, value: &str) -> &mut Self { 
+        self.identity_url = Some(value.to_string()); 
+        self 
+    }
+    
+    /// builds a ResponseTemplate from builder
+    pub fn build(&mut self) -> ResponseTemplate {
```
poljar: I think the `build()` method won't be necessary once the constructor takes all the required fields.

This would make the template consistent with the other custom templates we have.

crates/matrix-sdk/src/test_utils/mocks/mod.rs
```diff
@@ -3230,8 +3230,20 @@ impl<'a> MockEndpoint<'a, LoginEndpoint> {
     pub fn ok(self) -> MatrixMock<'a> {
         self.respond_with(ResponseTemplate::new(200).set_body_json(&*test_json::LOGIN))
     }
+
+    /// Returns a successful response with a given message.
+    pub fn ok_with(self, response: ResponseTemplate) -> MatrixMock<'a> {
+        self.respond_with(response)
+    }
```
poljar: It would be nice to add an example to this docstring, so newcomers can more easily discover how this should be used. Plenty of methods in the mocks module have examples, feel free to copy most of it from another example and modify it for this method.

crates/matrix-sdk/tests/integration/refresh_token.rs
dragonfly1033: The existing ok method uses the LOGIN response not the LOGIN_WITH_REFRESH_TOKEN response. Should I change the default to just use it as is with LOGIN?

poljar: Ah sorry, I was supposed to remove this comment when I realized that an `ok()` already exists. Feel free to ignore this comment.

crates/matrix-sdk/src/test_utils/mocks/mod.rs
dragonfly1033: If I'm understanding correctly then you mean that the constructor takes required field values as arguments and sets default values for the rest. Then the setters can be used to set the optional args. Is this correct?

poljar: Yes, correct.

crates/matrix-sdk/src/test_utils/mocks/mod.rs
```diff
@@ -3230,6 +3230,175 @@ impl<'a> MockEndpoint<'a, LoginEndpoint> {
...
+    ///     assert!(
+    ///         result.user_id.unwrap() == "@cheeky_monkey:matrix.org",
+    ///         "wrong user id in response"
+    ///     );
+    /// }
+    /// ```
+    pub fn ok_with(self, response: LoginResponseTemplate200) -> MatrixMock<'a> {
+        if response.homeserver_url.is_none() && response.identity_url.is_some() {
+            panic!("identity uri given but not homeserver uri. homeserver uri required for well known field.")
+        }
```
poljar: If those two have a strong relationship and you shouldn't be able to set one without the other, then we should express this in the setter.

Meaning, instead of having two setters that set each of those fields separately, let's have one setter that sets both.

You would need to convert the field in the struct to be a single option that holds both the homeserver URL as well as the identity server URL.

crates/matrix-sdk/src/test_utils/mocks/mod.rs
```diff
@@ -3230,6 +3230,175 @@ impl<'a> MockEndpoint<'a, LoginEndpoint> {
...
+
+        let identity = if response.identity_url.is_none() {
+            None
+        } else {
+            Some(json!({
+                "base_url": response.identity_url
+            }))
+        };
```
poljar: This snippet will change, since the structure of the fields will change.

But to mention it anyways, for such patterns where we need to transform the type into something else in case it is a `Some` we have the `map()` method:

```suggestion
        let identity = response.identity_url.map(|url| {
            json!({ "base_url": url })
        });
```

This way the `None` stays a `None` but if the `Option` contains a `Some` it is mapped into a different type.

crates/matrix-sdk/src/test_utils/mocks/mod.rs
```diff
@@ -3230,6 +3230,175 @@ impl<'a> MockEndpoint<'a, LoginEndpoint> {
...
+    ///     );
+    ///     assert!(
+    ///         result.user_id.unwrap() == "@cheeky_monkey:matrix.org",
+    ///         "wrong user id in response"
+    ///     );
+    /// }
+    /// ```
+    pub fn ok_with(self, response: LoginResponseTemplate200) -> MatrixMock<'a> {
+        if response.homeserver_url.is_none() && response.identity_url.is_some() {
+            panic!("identity uri given but not homeserver uri. homeserver uri required for well known field.")
```
poljar: In Rust you generally won't use these `.is_none()` or `.is_some()` methods that often.

You would either destructure the `Option` using `if let` statements or use an `expect()` call.

So either:

```rust
if let Some(foo) = bar {
   // Do something if we have `Some`
} else {
   // Do something if we have `None`
}
```

or

```rust
let foo = bar.expect("identity uri given but not");
```

But see my next comment on how to tie these fields a bit better together so the whole logic here isn't needed.

crates/matrix-sdk/src/test_utils/mocks/mod.rs
```diff
@@ -3230,6 +3230,175 @@ impl<'a> MockEndpoint<'a, LoginEndpoint> {
...
+        };
+
+        self.respond_with(ResponseTemplate::new(200).set_body_json(json!({
+            "access_token": response.access_token,
+            "device_id": response.device_id,
+            "user_id": response.user_id,
+            "expires_in": response.expires_in,
+            "refresh_token": response.refresh_token,
+            "well_known": well_known,
+        })))
```
poljar: This looks much better than the manual `Value` crafting we had before, sorry for not mentioning this before you went on a whole side-quest of discovering how `Value` works.

crates/matrix-sdk/src/test_utils/mocks/mod.rs
```diff
@@ -3230,6 +3230,175 @@ impl<'a> MockEndpoint<'a, LoginEndpoint> {
...
+    /// Required: ID of the logged-in device. Will be the same as the
+    /// corresponding parameter in the request, if one was specified.
+    device_id: Option<String>,
+
+    /// The lifetime of the access token, in milliseconds. Once the access token
+    /// has expired a new access token can be obtained by using the provided
+    /// refresh token. If no refresh token is provided, the client will need
+    /// to re-log in to obtain a new access token. If not given, the client
+    /// can assume that the access token will not expire.
+    expires_in: Option<u128>,
```
poljar: Didn't we want to use a `Duration` here? Sorry if I forgot to mention this, but time should always use a stronger type in Rust. 

For this case a `Duration` is fine.

crates/matrix-sdk/src/test_utils/mocks/mod.rs
```diff
@@ -3230,6 +3230,175 @@ impl<'a> MockEndpoint<'a, LoginEndpoint> {
...
+/// A response to a [`LoginEndpoint`] query with status code 200.
+#[derive(Default)]
+pub struct LoginResponseTemplate200 {
+    /// Required: An access token for the account. This access token can then be
+    /// used to authorize other requests.
+    access_token: Option<String>,
+
+    /// Required: ID of the logged-in device. Will be the same as the
+    /// corresponding parameter in the request, if one was specified.
+    device_id: Option<String>,
```
poljar: I see that you used the stronger types in the setters. But I think it's even better to use the stronger types in these fields as well.

Then when we serialize things into a `Value` we can do the necessary conversion. For the `DeviceId` and `UserId` you don't need to convert them manually into a string, the serialization rules of those types will do the right thing.

If you are wondering which type needs to be put into the field, since the compiler will complain if you try to put a `DeviceId` here, it's `OwnedDeviceId` and you can get that type with the `to_owned()` method. The same applies to `UserId` and `OwnedUserId`.

These types behave the same way `str` and `String` behave, i.e the former is a DST (dynamically sized type) and does not own the data it points to, the later owns the underlying data. Not sure if you got to this point in the Rust book, and it isn't strictly important to figure out all of this right now.

The gist is, that you can go to the `Owned` type to make the compiler happy.

crates/matrix-sdk/src/test_utils/mocks/mod.rs
dragonfly1033: No worries! I understand how they work now so that's a benefit in my book.

crates/matrix-sdk/src/test_utils/mocks/mod.rs
dragonfly1033: That's useful thanks!

crates/matrix-sdk/src/test_utils/mocks/mod.rs
dragonfly1033: The relationship is one-way, identity depends on homeserver but not vice versa. How can I capture this in one option?

poljar: The setter accepts a homeserver and an optional identity server.

Expressed as a small sub-struct:

```rust
struct HomeserverAndMaybeIdentity {
    homeserver: String,
    identity_server: Option<String>
}
```

Or you can use a tuple as well:

```rust
struct LoginBuilder {
    ....
    homeserver_and_identity: (String, Option<String>),
    ....,
?
```

With some better names if possible :sweat_smile:.

poljar (approved):
Alright, looks good.

Nice touch on the `Into<OwnedDeviceId>` usage. I think that this could have been `impl Into<OwnedDeviceId>` which is a bit of a newer and more pleasant syntax for the same thing, but not that important.

Congrats on your first PR in the SDK, pop open your favorite drink to celebrate :champagne:

PR #5389, opened 2025-07-11
indexeddb: Remove incorrect line from changelog

richvdh:
I would say that support for encrypted history is notable.


poljar (approved):

poljar:
Whoops. Thanks.

PR #5390, opened 2025-07-11
feat(base): Remember the inviter if we accept an invite

poljar:
Another PR that is a prerequisite of #5322.

Par of https://github.com/matrix-org/matrix-rust-sdk/issues/4926.

### DIFF
--- a/crates/matrix-sdk-base/src/room/room_info.rs
+++ b/crates/matrix-sdk-base/src/room/room_info.rs
@@ -50,7 +50,7 @@ use ruma::{
     OwnedRoomId, OwnedUserId, RoomAliasId, RoomId, RoomVersionId, UserId,
 };
 use serde::{Deserialize, Serialize};
-use tracing::{debug, field::debug, info, instrument, warn};
+use tracing::{debug, error, field::debug, info, instrument, warn};
 
 use super::{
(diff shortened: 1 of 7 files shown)


crates/matrix-sdk-base/src/room/room_info.rs
```diff
@@ -525,6 +537,9 @@ impl RoomInfo {
 
     /// Set the membership RoomState of this Room
     pub fn set_state(&mut self, room_state: RoomState) {
+        // Changing our state removes the invite details since we can't know that they
```
Hywan, line 543: Do you think we should set `invite_acceptance_details` to `None` if and only if the previous state was `RoomState::Invited`? At least it would be make things clearer, and it wouldn't hide a possible bug if any. Like, we may want to throw an (log) error if the previous state is not `Invited` and if `invite_acceptance_details` is `Some`?

Hywan (approved):
Approving, thanks for the work. I'm leaving one question, not a blocker.

crates/matrix-sdk-base/src/room/room_info.rs
poljar: Hmm, I guess we can add a log for that case.

We also have some tests that confirm that it isn't set when it shouldn't be.

poljar: Added something.

PR #5394, opened 2025-07-13
FFI: Pass existing device ID to url_for_oidc

### LINKED ISSUE #5392: FFI: Pass existing device ID to url_for_oidc
Currently, each time `Client.url_for_oidc` is called, a new device ID is generated as `OAuth.login` is passed `None` as the `device_id` parameter.
I'd like to allow setting the `device_id` also for the FFI binding.

Let me know if this feature is wanted :)

fl0lli:
Currently, each time `Client.url_for_oidc` is called, a new device ID is generated as `OAuth.login` is passed `None` as the `device_id` parameter.

This closes #5392.

Signed-off-by: fl0lli <github@fl0lli.de>

### DIFF
--- a/bindings/matrix-sdk-ffi/CHANGELOG.md
+++ b/bindings/matrix-sdk-ffi/CHANGELOG.md
@@ -11,6 +11,8 @@ All notable changes to this project will be documented in this file.
 - `ClientBuilder::build_with_qr_code` has been removed. Instead, the Client should be built by passing
   `QrCodeData::server_name` to `ClientBuilder::server_name_or_homeserver_url`, after which QR login can be performed by
   calling `Client::login_with_qr_code`. ([#5388](https://github.com/matrix-org/matrix-rust-sdk/pull/5388))
+- `Client::url_for_oidc` now allows passing an optional existing device id from a previous login call.
+  ([#5394](https://github.com/matrix-org/matrix-rust-sdk/pull/5394))
 
 ## [0.13.0] - 2025-07-10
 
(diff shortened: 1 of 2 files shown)


poljar (changes requested):
I guess adding this is fine since it more closely matches the SDK API.

Still could you expand on your use-case here? Also a changelog entry is required for this since this is a breaking change.

fl0lli:
Hi, thanks for your fast reply! :)

> Still could you expand on your use-case here? 

For a custom integration I need to authenticate a user twice. Relating to #5395, after the first "plain" login I check which scopes a user can request (as I do not know that beforehand), and call `url_for_oidc` again using a new scope set (e.g. the Synapse Admin Scope or a custom Scope).
Without this change, after the second call, the SDK will fail as the device ids from first and second call mismatch and the CryptoStore complains.

> Also a changelog entry is required for this since this is a breaking change.

Forgot. Added

bindings/matrix-sdk-ffi/CHANGELOG.md
```diff
@@ -11,6 +11,8 @@ All notable changes to this project will be documented in this file.
 - `ClientBuilder::build_with_qr_code` has been removed. Instead, the Client should be built by passing
   `QrCodeData::server_name` to `ClientBuilder::server_name_or_homeserver_url`, after which QR login can be performed by
   calling `Client::login_with_qr_code`. ([#5388](https://github.com/matrix-org/matrix-rust-sdk/pull/5388))
+- `Client::url_for_oidc` now allows passing an optional existing device id from a previous login call.
+  ([#5394](https://github.com/matrix-org/matrix-rust-sdk/pull/5394))
```
poljar: We add new entries to the top.

poljar (changes requested):
Alright makes sense. Let's just move the entry to the top.

poljar (approved):

PR #5401, opened 2025-07-14
base: Refactoring around room power levels

zecakeh:
This tries to reduce duplication for code that accesses the power levels of a room.


Hywan (approved):
Reaaaallly nice! I've really enjoyed reading the commits. Thanks for the good work!

PR #5384, opened 2025-07-09
IndexedDB: Add IndexedDB implementation of `EventCacheStore::load_last_chunk`

mgoldenberg:
## Background

This pull request is part of a series of pull requests to add a full IndexedDB implementation of the `EventCacheStore` (see #4617, #4996, #5090, #5138, #5226, #5274, #5343). This particular pull request focuses on providing an implementation of `EventCacheStore::load_last_chunk`.

## Changes

The primary change is the addition of `EventCacheStore::load_last_chunk`, as well as tests for ensuring this is functioning correctly.

There are also a number of additions to `IndexeddbEventCacheStoreTransaction` which support the implementation of `EventCacheStore::load_last_chunk`.

- `get_chunk_by_next_chunk_id`
- `get_chunks_count_in_room`
- `get_max_chunk_id`

These are simply typed wrappers around the generic functions in `IndexeddbEventCacheStoreTransaction`.

## Future Work

- Add remaining implementations of `EventCacheStore` functions without relying on `MemoryStore`.

---

- [ ] Public API changes documented in changelogs (optional)

Signed-off-by: Michael Goldenberg <m@mgoldenberg.net>

PR #5343, opened 2025-07-02
IndexedDB: Add initial implementation of `EventCacheStore`

mgoldenberg:
## Background

This pull request is part of a series of pull requests to add a full IndexedDB implementation of the `EventCacheStore` (see #4617, #4996, #5090, #5138, #5226, #5274). This particular pull request focuses on providing implementations of a few `EventCacheStore` functions.

## Changes

The primary change is the addition of a top-level type, `IndexeddbEventCacheStore`, as well as a way of constructing this type through `IndexeddbEventCacheStoreBuilder`. Additionally, `IndexeddbEventCacheStore` also provides an implementation of `EventCacheStore`.

In order to limit the size of this pull request, the implementation of `EventCacheStore` is almost entirely delegated to a nested `MemoryStore` except for the following two functions, which have a genuine IndexedDB implementation.

- `EventCacheStore::handle_linked_chunk_updates`
- `EventCacheStore::load_all_chunks`

Additionally, there are a number of additions to `IndexeddbEventCacheStoreTransaction` which support the implementation of the `EventCacheStore` functions above.

- `get_chunk_by_id` 
- `get_chunks_in_room` 
- `load_chunk_by_id` 
- `add_chunk` 
- `delete_chunk_by_id` 
- `delete_chunks_in_room` 
- `get_events_by_position` 
- `get_events_by_chunk` 
- `put_event` 
- `delete_events_by_position` 
- `delete_event_by_position` 
- `delete_events_by_chunk` 
- `delete_events_by_chunk_from_index` 
- `delete_events_in_room` 
- `get_gap_by_id` 
- `delete_gap_by_id` 
- `delete_gaps_in_room` 

These are mostly typed wrappers around the generic functions in `IndexeddbEventCacheStoreTransaction`, though some of them add extra functionality as well.

## Tests

A quick note about the tests. Because most of the functions are delegated to a nested `MemoryStore` and only two are actually implemented, the `event_cache_store_integration_tests` macro will generate many failures. Consequently, I have made a local copy of the macro which only includes tests which are relevant to `EventCacheStore::handle_linked_chunk_updates` and `EventCacheStore::load_all_chunks`. Once all functionality is added, this local macro will disappear and I will use the original one from `matrix_sdk_base`.

Additionally, there are a number of tests which have been copied and adapted from SQLite implementation of `EventCacheStore`. Many of these tests will likely be generalized and included as part of the `event_cache_store_integration_tests` macro and will also, hopefully, disappear (see #5342).

## Future Work

[...]

PR #5406, opened 2025-07-15
IndexedDB: Add IndexedDB implementation of `LinkedChunk`-related functions in `EventCacheStore`

mgoldenberg:
## Background

This pull request is part of a series of pull requests to add a full IndexedDB implementation of the `EventCacheStore` (see #4617, #4996, #5090, #5138, #5226, #5274, #5343, #5384). This particular pull request focuses on providing implementations for the remaining `LinkedChunk`-related functions in `EventCacheStore`.

## Changes

The changes include implementations of the following functions, as well as tests for each of them.

- `EventCacheStore::load_all_chunks_metadata`
- `EventCacheStore::load_previous_chunk`
- `EventCacheStore::clear_all_linked_chunks`

There are also two additions to `IndexeddbEventCacheStoreTransaction` which support the implementation of the `EventCacheStore::load_all_chunks_metadata`.

- `get_events_count_by_position`
- `get_events_count_by_chunk`

## Future Work

- Add remaining implementations of `EventCacheStore` functions without relying on `MemoryStore`

---

### DIFF
--- a/crates/matrix-sdk-indexeddb/src/event_cache_store/mod.rs
+++ b/crates/matrix-sdk-indexeddb/src/event_cache_store/mod.rs
@@ -28,9 +28,10 @@ use matrix_sdk_base::{
         RawChunk, Update,
     },
     media::MediaRequestParameters,
+    timer,
 };
 use ruma::{events::relation::RelationType, EventId, MxcUri, OwnedEventId, RoomId};
-use tracing::trace;
+use tracing::{instrument, trace};
 use web_sys::IdbTransactionMode;
 
 use crate::event_cache_store::{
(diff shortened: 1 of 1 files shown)


crates/matrix-sdk-indexeddb/src/event_cache_store/mod.rs
```diff
@@ -287,10 +287,27 @@ impl_event_cache_store! {
...
+        let chunks = transaction.get_chunks_in_room(room_id).await?;
+        for chunk in chunks {
+            let chunk_id = ChunkIdentifier::new(chunk.identifier);
+            let num_items = transaction.get_events_count_by_chunk(room_id, &chunk_id).await?;
+            raw_chunks.push(ChunkMetadata {
+                num_items,
+                previous: chunk.previous.map(ChunkIdentifier::new),
+                identifier: ChunkIdentifier::new(chunk.identifier),
+                next: chunk.next.map(ChunkIdentifier::new),
+            });
```
Hywan, line 330: If we have thousands chunks, this isn't going to scale well. I mean, it's probably fine for the moment, but a big `TODO` + a `timer!` macro is necessary here.

I suggest to add logs as I did in https://github.com/matrix-org/matrix-rust-sdk/pull/5382 (see commits by commits) with the `timer!` log macro. It's going to be pretty helpful to find which method is slow.

Hywan (changes requested):
Excellent! I left a feedback regarding a query made in a loop. Otherwise, it's good! Thanks.

crates/matrix-sdk-indexeddb/src/event_cache_store/mod.rs
mgoldenberg: Ah, interesting - I didn't think about that - thanks! And will do!

mgoldenberg: Okay, added timers and put a note in `EventCacheStore::load_all_chunks_metadata` about the performance issue. I will study #5382 and think about how to apply this to IndexedDB.

Hywan (approved):

PR #5407, opened 2025-07-15
fix(sdk): Disable `OrderTracker` for the moment

Hywan:
This patch removes the use of `OrderTracker` because the implementation of `EventCacheStore::load_all_chunks_metadata` for `SqliteEventCacheStore` is the cause of severe slowness (up to 100s for some accounts).

We are going to undo this patch once the problem has been solved.

### Metrics

Using [this rageshake](https://github.com/element-hq/element-x-ios-rageshakes/issues/4248), the logs introduced in [this PR](https://github.com/matrix-org/matrix-rust-sdk/pull/5382), and this command-line to only list executions of `load_all_chunks_metadata` taking **more than one second**:

```shell
$ rg '_method_ finished in.*> load_all_chunks_metadata' foo.log | rg '\d+(\.\d+)?s' --only-matching | sort --numeric-sort --reverse
107.121747125s
79.909931458s
10.348993583s
8.827636417s
8.614481625s
8.009787875s
5.99637875s
4.118492334s
3.910040333s
3.718858334s
3.689340667s
3.661383208s
2.310609666s
2.221779s
2.087249333s
1.898318333s
1.320944334s
1.005408875s
```


Hywan:
For the record, perf is no longer a problem. It confirms this is the culprit. Now let's fix that.

PR #5408, opened 2025-07-15
chore(bench): Update Criterion to 0.6

Hywan:
Does exactly what it says on the tin.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -482,7 +482,7 @@ dependencies = [
 name = "benchmarks"
 version = "1.0.0"
 dependencies = [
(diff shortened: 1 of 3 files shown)


Hywan:
Cannot update to Criterion 0.16 because pprof is still using 0.15, https://crates.io/crates/pprof/0.15.0/dependencies. Need to wait on https://github.com/tikv/pprof-rs/issues/274 to be addressed.

poljar:
This has been superseded by https://github.com/matrix-org/matrix-rust-sdk/pull/5443.

Issue #5397, opened 2025-07-14
Meta: Encrypted state


andybalaam:
@kaylendog will be working on encrypted state - we will track our plans and progress here.

## Goal

Investigate the feasibility of [MSC3414](https://github.com/matrix-org/matrix-spec-proposals/pull/3414) and prototype it in rust-sdk.

## Context

Currently, all [state events](https://spec.matrix.org/v1.15/client-server-api/#types-of-room-events) in a room are unencrypted: only non-state events can be encrypted.

To enhance privacy, we should encrypt state events where possible.

## Problems

- The server holds the current state of a room, and it uses `type`+`state_key` as the key of the map. This causes 2 problems:
    1. Encrypted events have a fixed `type` of `m.room.encrypted` so events of different types will clash
    2. We potentially want the true `state_key` to be inaccessible to people without decryption keys

We will start by not considering this problem, then progressively enhance our solution.

## Plan

- [x] Set-up, logins, passwords etc.
- [x] Build matrix-rust-sdk including [multiverse](https://github.com/matrix-org/matrix-rust-sdk/tree/main/labs/multiverse)
- [x] Run a local [synapse](https://github.com/element-hq/synapse/) and connect to it from multiverse
- [x] Encrypt a test event. Write the minimum possible code to get a state event to send and receive whose contents are encrypted. Look at [OlmMachine::encrypt_room_event](https://github.com/matrix-org/matrix-rust-sdk/blob/main/crates/matrix-sdk-crypto/src/machine/mod.rs#L1064). You will probably need to write a similar method called `encrypt_state_event` and figure out how to call it from multiverse. You will probably need to add code to decrypt state events if they arrive encrypted.
- [x] Consider stuffing the event type into `state_key` so the server can hold two events of different `type` with the same `state_key` even if they are encrypted. E.g. change the state_key to be "{type}:${state_key}" (note: careful choice of separator will eventually be needed, but no need to worry too much now). Confirm that the server correctly returns the right state in this circumstance. Figure out how to get the client to "unstuff" it and understand the right type and state_key.
- [x] Investigate how well this works. E.g. can I do [live location sharing](https://github.com/matrix-org/matrix-spec-proposals/pull/3489) using encrypted state events?
- [x] Implement runtime feature flag for encrypted state
- [x] Merge into upstream SDK
- [x] Investigate implementing encrypted state into existing clients - feature/labs flag.
- [ ] Implement encrypted state in Element Web as a labs flag.

[...]

PR #5409, opened 2025-07-15
Crypto: Split `types::events::room::encrypted` to `scheme`, `room`, and `to_device` modules.

kaylendog:
This PR serves as a primer for my work on #5397, which would be simplified by isolating the type definitions in this module.

- Split `types::events::room::encrypted` into three modules:
    - `scheme` - defines the various schemes used to encrypt room events;
    - `room` - defines how `m.room.encrypted` message events are encrypted;
    - `to_device` - defines how `m.room.encrypted` to-device events are encrypted;
- Modified `types::events::room::encrypted` to re-export from these new locations to avoid rewriting imports in the entire crate.

Having `types::events::room::room` is a little silly, so perhaps further refactoring could be done.


crates/matrix-sdk-crypto/src/types/events/room/encrypted.rs
```diff
@@ -12,511 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-//! Types for the `m.room.encrypted` room events.
+//! Types for the `m.room.encrypted` room events. Re-exports for compatibility with existing code.
```
poljar: Not sure that this belongs in a doc comment. The fact that these are re-exports is an implementation detail.

crates/matrix-sdk-crypto/src/types/events/room/encrypted.rs
```diff
@@ -12,511 +12,8 @@
...
-
-        let serialized = serde_json::to_value(event)?;
-        assert_eq!(json, serialized);
-
-        Ok(())
-    }
-}
+pub use super::room::*;
+pub use super::scheme::*;
+pub use super::to_device::*;
```
poljar: Why would you put these modules into `room` instead of creating submodules in the `encrypted` module?

kaylendog:
@poljar I've made another adjustment to the structure. It feels more intuitive (at least for me) to store the encrypted types for each event type alongside the rest of their definitions, even if they are implemented under the hood using `m.room.encrypted`. Since I will be adding types for the state events in #5397, this structure should reduce the scattering of encrypted vs non-encrypted types between `types/events/room/encrypted/bla.rs` and `types/events/bla.rs`. 

I can *probably* introduce a better structure conforming to [your comment](https://github.com/matrix-org/matrix-rust-sdk/pull/5409#discussion_r2207398662) that works well with my next work, but this feels better to me? I think it's a matter of whether the protocol event type (namely `m.room.encrypted`) dominates the decision of where to place files.

My work in #5397 will need to store state event types in the state key, while preserving the `m.room.encrypted` event type, which is something to also consider.

kaylendog:
Converting this to draft for now since it's becoming more unclear whether I'll actually need to add types to the crypto crate.

poljar:
I assume we can close this since you went in another direction?

poljar:
Closing now due to inactivity.

PR #5414, opened 2025-07-15
feat(indexeddb): implement `Event`-related functions in `EventCacheStore`

mgoldenberg:
## Background

This pull request is part of a series of pull requests to add a full IndexedDB implementation of the `EventCacheStore` (see #4617, #4996, #5090, #5138, #5226, #5274, #5343, #5384, #5406). This particular pull request focuses on providing implementations for the `Event`-related functions in `EventCacheStore`.

## Changes

The changes include implementations of the following functions, as well as tests for each of them.

- `EventCacheStore::filter_duplicated_events`
- `EventCacheStore::find_event`
- `EventCacheStore::find_event_relations`
- `EventCacheStore::save_event`

I have also added a few other functions which support the implementation of the functions above.

- `IndexedEventRelationKey::set_related_event`
- `IndexeddbEventCacheStoreTransaction::get_event_by_id`
- `IndexeddbEventCacheStoreTransaction::get_events_by_relation`
- `IndexeddbEventCacheStoreTransaction::get_events_by_related_event`

## Bug Fixes

There is also a small change to how `Update::PushItems` is handled. Namely, if a given item already exists in the database, it will update that item, rather than returning an error. The new behavior is consistent with the SQLite implementation of `EventCacheStore` (see [here](https://github.com/matrix-org/matrix-rust-sdk/blob/471e3c340c0c98418fc3c0384c07ecc763d9d596/crates/matrix-sdk-sqlite/src/event_cache_store.rs#L595)).

This bug was surfaced by [`matrix_sdk_base::event_cache::store::integration_tests::test_find_event_relations`](https://github.com/matrix-org/matrix-rust-sdk/blob/471e3c340c0c98418fc3c0384c07ecc763d9d596/crates/matrix-sdk-base/src/event_cache/store/integration_tests.rs#L988).

## Future Work

- Add implementation of `EventCacheStore::try_take_leased_lock` functions without relying on `MemoryStore`
- Add implementation of `EventCacheStoreMedia`

---

### DIFF
--- a/crates/matrix-sdk-indexeddb/src/event_cache_store/mod.rs
+++ b/crates/matrix-sdk-indexeddb/src/event_cache_store/mod.rs
@@ -462,10 +462,14 @@ impl_event_cache_store! {
         event_id: &EventId,
     ) -> Result<Option<Event>, IndexeddbEventCacheStoreError> {
         let _timer = timer!("method");
-        self.memory_store
-            .find_event(room_id, event_id)
+
+        let transaction =
+            self.transaction(&[keys::EVENTS], IdbTransactionMode::Readonly)?;
+        transaction
+            .get_event_by_id(room_id, &event_id.to_owned())
             .await
-            .map_err(IndexeddbEventCacheStoreError::MemoryStore)
+            .map(|ok| ok.map(Into::into))
+            .map_err(Into::into)
     }
 
     #[instrument(skip(self, event_id, filters))]
--- a/crates/matrix-sdk-indexeddb/src/event_cache_store/mod.rs
+++ b/crates/matrix-sdk-indexeddb/src/event_cache_store/mod.rs
@@ -480,10 +480,32 @@ impl_event_cache_store! {
         filters: Option<&[RelationType]>,
     ) -> Result<Vec<(Event, Option<Position>)>, IndexeddbEventCacheStoreError> {
         let _timer = timer!("method");
-        self.memory_store
-            .find_event_relations(room_id, event_id, filters)
-            .await
-            .map_err(IndexeddbEventCacheStoreError::MemoryStore)
+
+        let transaction =
+            self.transaction(&[keys::EVENTS], IdbTransactionMode::Readonly)?;
+
+        let mut related_events = Vec::new();
+        match filters {
+            Some(relation_types) if !relation_types.is_empty() => {
+                for relation_type in relation_types {
+                    let relation = (event_id.to_owned(), relation_type.clone());
+                    let events = transaction.get_events_by_relation(room_id, &relation).await?;
+                    for event in events {
+                        let position = event.position().map(Into::into);
+                        related_events.push((event.into(), position));
+                    }
+                }
+            }
+            _ => {
+                for event in
+                    transaction.get_events_by_related_event(room_id, &event_id.to_owned()).await?
+                {
+                    let position = event.position().map(Into::into);
+                    related_events.push((event.into(), position));
+                }
+            }
+        }
+        Ok(related_events)
     }
 
     #[instrument(skip(self, event))]
--- a/crates/matrix-sdk-indexeddb/src/event_cache_store/mod.rs
+++ b/crates/matrix-sdk-indexeddb/src/event_cache_store/mod.rs
@@ -435,10 +435,24 @@ impl_event_cache_store! {
         events: Vec<OwnedEventId>,
     ) -> Result<Vec<(OwnedEventId, Position)>, IndexeddbEventCacheStoreError> {
         let _timer = timer!("method");
-        self.memory_store
-            .filter_duplicated_events(linked_chunk_id, events)
-            .await
-            .map_err(IndexeddbEventCacheStoreError::MemoryStore)
+
+        if events.is_empty() {
+            return Ok(Vec::new());
+        }
+
+        let linked_chunk_id = linked_chunk_id.to_owned();
+        let room_id = linked_chunk_id.room_id();
+        let transaction =
+            self.transaction(&[keys::EVENTS], IdbTransactionMode::Readonly)?;
+        let mut duplicated = Vec::new();
+        for event_id in events {
+            if let Some(types::Event::InBand(event)) =
+                transaction.get_event_by_id(room_id, &event_id).await?
+            {
+                duplicated.push((event_id, event.position.into()));
+            }
+        }
+        Ok(duplicated)
     }
 
     #[instrument(skip(self, event_id))]
--- a/crates/matrix-sdk-indexeddb/src/event_cache_store/mod.rs
+++ b/crates/matrix-sdk-indexeddb/src/event_cache_store/mod.rs
@@ -31,14 +31,14 @@ use matrix_sdk_base::{
     timer,
 };
 use ruma::{events::relation::RelationType, EventId, MxcUri, OwnedEventId, RoomId};
-use tracing::{instrument, trace};
+use tracing::{error, instrument, trace};
 use web_sys::IdbTransactionMode;
 
 use crate::event_cache_store::{
     migrations::current::keys,
     serializer::IndexeddbEventCacheStoreSerializer,
     transaction::{IndexeddbEventCacheStoreTransaction, IndexeddbEventCacheStoreTransactionError},
-    types::{ChunkType, InBandEvent},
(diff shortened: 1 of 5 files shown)


crates/matrix-sdk-indexeddb/src/event_cache_store/mod.rs
```diff
@@ -435,10 +435,24 @@ impl_event_cache_store! {
...
-            .filter_duplicated_events(linked_chunk_id, events)
-            .await
-            .map_err(IndexeddbEventCacheStoreError::MemoryStore)
+
+        if events.is_empty() {
+            return Ok(Vec::new());
+        }
+
+        let linked_chunk_id = linked_chunk_id.to_owned();
+        let room_id = linked_chunk_id.room_id();
```
bnjbvr, line 444: This is incorrect: deduplication should happen on a linked chunk basis, not a room basis. In some future, we'll store threads on disk, so we'll have a linked chunk for each thread, and one for the room. Duplicated in a thread isn't the same thing as duplicated in the room (some in-thread events might be present in the room linked chunk, which is fine in the current design), so we need to focus the query to "in this given linked chunk, are the given events duplicated or not".

For what it's worth, the sqlite backend stores the event ids and their positions, in each linked chunk. The full event JSONs themselves are stored in a separate table; this indirection makes it so that the duplication of events in the room linked chunk and the threads linked chunks isn't too costly. I don't know if the IDB backend does something similar, but it might be handy to do that!

In any case: it would be nice to have a test for this, but I'm afraid it'll require a bit of setup (i.e. add the `Thread` variants to `LinkedChunkId` / `OwnedLinkedChunkId`, and complete the implementations), so we can defer this to another PR (and I can do it, if you're not interested, of course :)).

crates/matrix-sdk-indexeddb/src/event_cache_store/types.rs
```diff
@@ -94,6 +94,15 @@ impl Event {
             Event::OutOfBand(e) => e.relation(),
         }
     }
+
+    /// Sets the content of the underlying [`GenericEvent`]
```
bnjbvr: Is this the right name? it seems there's no such `GenericEvent` in this file.

bnjbvr:
Thanks for the PR! Stealing review, as I'm more acquainted with the event cache.

crates/matrix-sdk-indexeddb/src/event_cache_store/types.rs
```diff
@@ -94,6 +94,15 @@ impl Event {
...
+    pub fn set_content(mut self, content: TimelineEvent) -> Self {
```
bnjbvr: nitpick: I'd name this `with_content`, instead of `set_content`, because I'd expect `set_content` to take a `&mut self` and not return a new instance.

crates/matrix-sdk-indexeddb/src/event_cache_store/serializer/types.rs
```diff
@@ -417,6 +417,23 @@ pub type IndexedEventPositionIndex = usize;
 #[derive(Debug, Serialize, Deserialize)]
 pub struct IndexedEventRelationKey(IndexedRoomId, IndexedEventId, IndexedRelationType);
 
+impl IndexedEventRelationKey {
+    /// Sets the related event field of this key. This is helpful
+    /// when searching for all events which are related to the given
+    /// event.
+    pub fn set_related_event_id(
```
bnjbvr: ditto, I'd name this `with_related_event_id`

bnjbvr (changes requested):
Sorry about the long review time! Holidays make it so that there are fewer of us working around this time.

Looks good in general, there's a tiny logic issue in the first commit, but the rest looks great, thanks!

crates/matrix-sdk-indexeddb/src/event_cache_store/mod.rs
mgoldenberg: Ah, okay, this is illuminating! I was wondering what `LinkedChunkId::Thread` was for, but I saw that it was commented, so I figured I'd just ignore it until it became a problem - you'll see that all the other functions in this implementation do the same.

So, I think I mostly understand what _should_ be done - i.e., focus the query - but it seems like I am not able to accomplish this without the `LinkedChunkId::Thread` variant - am I understanding that correctly? Or is there something that I could be improving here even with the `LinkedChunkId::Thread` still commented?

In any case, a test case would definitely be helpful in clarifying some things and I will also take a closer look at the SQLite implementation!

Appreciate your time!

crates/matrix-sdk-indexeddb/src/event_cache_store/types.rs
mgoldenberg: It's just a bit below, line [115](https://github.com/matrix-org/matrix-rust-sdk/blob/07b5d025d435ba03d87e8172ac9cbeb359783d0f/crates/matrix-sdk-indexeddb/src/event_cache_store/types.rs#L115)!

crates/matrix-sdk-indexeddb/src/event_cache_store/types.rs
mgoldenberg: Yes, agree - will do!

crates/matrix-sdk-indexeddb/src/event_cache_store/serializer/types.rs
mgoldenberg: Agree - will do!

crates/matrix-sdk-indexeddb/src/event_cache_store/types.rs
mgoldenberg: fixup: [d534857b69cd3d6dba364e4db512cdd35ef3d509](https://github.com/matrix-org/matrix-rust-sdk/pull/5414/commits/d534857b69cd3d6dba364e4db512cdd35ef3d509)

crates/matrix-sdk-indexeddb/src/event_cache_store/serializer/types.rs
mgoldenberg: fixup: [4799ce1a34bd2725cf75e52ff3ac6f66258d52d4](https://github.com/matrix-org/matrix-rust-sdk/pull/5414/commits/4799ce1a34bd2725cf75e52ff3ac6f66258d52d4)

crates/matrix-sdk-indexeddb/src/event_cache_store/mod.rs
bnjbvr: > So, I think I mostly understand what should be done - i.e., focus the query - but it seems like I am not able to accomplish this without the LinkedChunkId::Thread variant - am I understanding that correctly? Or is there something that I could be improving here even with the LinkedChunkId::Thread still commented?

I don't think you need to implement the `Thread` variant so as to implement the correct behavior here. The new variant should only be required for the test.

The idea is that a linked chunk is "keyed" by the `LinkedChunkId`. Think of the storage as a map from `LinkedChunkId` to the actual storage data backing the linked chunk. The bytes representing the linked chunk id can be retrieved from `LinkedChunkId::storage_key()`; so in our metaphorical map here, the return type of `storage_key()` is effectively the map's key type.

In SQL, this mapping means we're selecting the linked chunk by its linked chunk id, on every query that involves a `LinkedChunkId` (in the `EventCacheStore` trait's methods). For instance, `handle_linked_chunk_updates` must identify the linked chunk it's acting upon, based on the linked chunk id; every `INSERT`/`UPDATE`/`DELETE` will have a `WHERE linked_chunk_id = ?` clause in it. On the other hand, `save_event` only uses a `RoomId`, so there's no way to save it into a linked chunk, because we couldn't know which one (and at which position). As such, the store must only save the event "out of band" (I think I've seen the IDB impl had that already).

Does it make sense?

mgoldenberg: Yes, I think I understand now, thank you!

mgoldenberg: As per our conversation, I am going to resolve this comment and put these fixes in a future pull request as a proper fix requires some refactoring that is outside the scope of this pull request.

bnjbvr (approved):
Thanks, with the linked-chunk-id keying postponed to later, LGTM :+1:

Issue #5381, opened 2025-07-09
Leaving a room must leave all the predecessor rooms


Hywan:
Currently, when a user is leaving a room, it leaves the room and that's it. However, we must also leave all the predecessor rooms, i.e. all its the tombstoned rooms.

razvp:
To solve this we have to recursively find all the predecessor rooms and make a `leave` request for each of them. I'm not sure there is a way to do this atomically, so if the client disconnects while leaving the rooms, he will remain joined in some rooms. Is there a better way to do this?

There are also other ways for a user to `leave` a room (like a kick from an Admin) and this implementation wouldn't cover that. Is this ok? Maybe this feature would be better implemented on the homeserver?

Hywan:
It could be implemented server-side, as long as the state events aren't not encrypted.

> I'm not sure there is a way to do this atomically, so if the client disconnects while leaving the rooms, he will remain joined in some rooms. Is there a better way to do this?

That's a real concern yes. I don't believe it's a critical operation, so we may consider it's up to the user to handle that for the moment.

> There are also other ways for a user to leave a room (like a kick from an Admin) and this implementation wouldn't cover that. Is this ok?

Let's start with a normal leave before working on kick or ban. Those have different semantics I reckon.

poljar:
This was closed by #5418.

PR #5418, opened 2025-07-16
Fix for issue #5381: Leaving room leaves predecessors 

dragonfly1033:
Made the ```Room.leave``` function recursive to leave predecessor rooms. Also added an integration test to check this.  

### DIFF
--- a/crates/matrix-sdk/src/error.rs
+++ b/crates/matrix-sdk/src/error.rs
@@ -276,18 +276,6 @@ impl RetryKind {
 #[derive(Error, Debug)]
 #[non_exhaustive]
 pub enum Error {
-    /// Error during room.leave().
-    #[error("an error occurred during room leave")]
-    ErrorDuringRoomLeave,
-
-    /// Error doing a get_room(predecessor).
-    #[error("the predecessor is wrong or forgotten")]
-    PredecessorWrongOrForgotten,
-
-    /// Error doing room.leave()
-    #[error("the predecessor is not Joined or Left")]
-    PredecessorWrongState,
-
     /// Error doing an HTTP request.
     #[error(transparent)]
     Http(Box<HttpError>),
--- a/crates/matrix-sdk/src/error.rs
+++ b/crates/matrix-sdk/src/error.rs
@@ -277,7 +277,7 @@ impl RetryKind {
 #[non_exhaustive]
 pub enum Error {
     /// Error during room.leave().
-    #[error("an error occured during room leave")]
+    #[error("an error occurred during room leave")]
     ErrorDuringRoomLeave,
 
     /// Error doing a get_room(predecessor).
--- a/crates/matrix-sdk/src/error.rs
+++ b/crates/matrix-sdk/src/error.rs
@@ -276,6 +276,18 @@ impl RetryKind {
 #[derive(Error, Debug)]
 #[non_exhaustive]
 pub enum Error {
+    /// Error during room.leave().
+    #[error("an error occured during room leave")]
+    ErrorDuringRoomLeave,
+
+    /// Error doing a get_room(predecessor).
+    #[error("the predecessor is wrong or forgotten")]
+    PredecessorWrongOrForgotten,
+
+    /// Error doing room.leave()
+    #[error("the predecessor is not Joined or Left")]
+    PredecessorWrongState,
+
     /// Error doing an HTTP request.
     #[error(transparent)]
     Http(Box<HttpError>),
--- a/crates/matrix-sdk/src/room/mod.rs
+++ b/crates/matrix-sdk/src/room/mod.rs
@@ -365,6 +365,11 @@ impl Room {
             }
         }
 
+        if let Some(pred_room) = self.predecessor_room() {
+            let pred = self.client.get_room(&pred_room.room_id).expect("Couldn't find predecessor");
+            Box::pin(pred.leave()).await?;
+        }
+
         Ok(())
     }
 
--- a/crates/matrix-sdk/src/room/mod.rs
+++ b/crates/matrix-sdk/src/room/mod.rs
@@ -415,17 +415,13 @@ impl Room {
             if let (Err(e), room) = result {
                 if room.room_id() == self.room_id() {
                     maybe_this_room_failed_with = Some(e);
-                    continue;
+                } else {
+                    warn!("Failure while attempting to leave predecessor room: {e:?}");
                 }
-                warn!("{e:?}");
             }
         }
 
-        if let Some(error) = maybe_this_room_failed_with {
-            return Err(error);
-        }
-
-        Ok(())
+        maybe_this_room_failed_with.map_or(Ok(()), Err)
     }
 
     /// Join this room.
--- a/crates/matrix-sdk/src/room/mod.rs
+++ b/crates/matrix-sdk/src/room/mod.rs
@@ -379,6 +379,8 @@ impl Room {
     /// automatically.
     ///
     /// Only invited and joined rooms can be left.
+    /// Will return an error if the current room fails to leave but
+    /// will only warn if a predecessor fails to leave.
     pub async fn leave(&self) -> Result<()> {
         let mut rooms: Vec<Room> = vec![self.clone()];
         let mut current_room = self;
--- a/crates/matrix-sdk/src/room/mod.rs
+++ b/crates/matrix-sdk/src/room/mod.rs
@@ -414,6 +414,8 @@ impl Room {
                 .iter()
                 .map(|room| match room.state() {
                     RoomState::Joined => Ok(Some(room.leave_this())),
+                    RoomState::Invited => Ok(Some(room.leave_this())),
+                    RoomState::Knocked => Ok(Some(room.leave_this())),
                     RoomState::Left => Ok(None),
                     _ => Err(Error::PredecessorWrongState),
                 })
@@ -423,7 +425,7 @@ impl Room {
                 .collect::<Vec<_>>();
 
             if join_all(batch).await.iter().any(Result::is_err) {
-                error!("Error occured while leaving room.");
+                error!("Error occurred while leaving room.");
                 return Err(Error::ErrorDuringRoomLeave);
             }
         }
--- a/crates/matrix-sdk/src/room/mod.rs
+++ b/crates/matrix-sdk/src/room/mod.rs
@@ -26,7 +26,9 @@ use std::{
 use async_stream::stream;
 use eyeball::SharedObservable;
 use futures_core::Stream;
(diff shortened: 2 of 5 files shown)


crates/matrix-sdk/src/room/mod.rs
```diff
@@ -365,6 +365,11 @@ impl Room {
             }
         }
 
+        if let Some(pred_room) = self.predecessor_room() {
+            let pred = self.client.get_room(&pred_room.room_id).expect("Couldn't find predecessor");
```
poljar: This will panic as soon as someone manually leaves and forgets a predecessor room before leaving this room.

Furthermore, a homeserver can give us a room with a fake predecessor.

crates/matrix-sdk/src/room/mod.rs
```diff
@@ -365,6 +365,11 @@ impl Room {
...
+            Box::pin(pred.leave()).await?;
```
poljar: Is the pin needed here?

It would probably make sense to make the server request in parallel.

If we need to leave 5 rooms, and each of them takes `500ms` we'll be waiting for 2.5 seconds. If we make each of those requests at the same time, we might be done in `500ms`.

We should limit the amount of parallel requests we make, 5-10 at a time sounds like a good ballpark value.

dragonfly1033: I *think* the pin is required because it is a recursive call in an async function and the typing would be unbounded otherwise. An alternative strategy could be to make a wrapper that accumulates the whole linked list and then leave them in concurrent batches with threads?

poljar: > An alternative strategy could be to make a wrapper that accumulates the whole linked list and then leave them in concurrent batches with threads?

Yes, though you wouldn't spawn any threads you would do something like the [`buffered`](https://docs.rs/futures/latest/futures/stream/trait.StreamExt.html#method.buffered) extension or [`join_all()`](https://docs.rs/futures-util/latest/futures_util/future/fn.join_all.html) to let futures run concurrently.

crates/matrix-sdk/src/room/mod.rs
```diff
@@ -368,6 +368,71 @@ impl Room {
...
+            }
+
+            let batch = batch
+                .iter()
+                .map(|room| match room.state() {
+                    RoomState::Joined => Ok(Some(room.leave_this())),
+                    RoomState::Invited => Ok(Some(room.leave_this())),
+                    RoomState::Knocked => Ok(Some(room.leave_this())),
+                    RoomState::Left => Ok(None),
+                    _ => Err(Error::PredecessorWrongState),
```
poljar: Is that an error? I suppose for the loop it can be considered but being banned from a room before leaving the successor room sounds like a normal operation, I don't think that this warrants a `error!()` log.

crates/matrix-sdk/src/room/mod.rs
```diff
@@ -368,6 +368,71 @@ impl Room {
...
+                })
+                .collect::<Result<Vec<Option<_>>, _>>()?
+                .into_iter()
+                .flatten()
+                .collect::<Vec<_>>();
+
+            if join_all(batch).await.iter().any(Result::is_err) {
+                error!("Error occurred while leaving room.");
```
poljar: We probably want to log why the leave failed.

crates/matrix-sdk/src/room/mod.rs
```diff
@@ -317,7 +317,7 @@ impl Room {
     /// Only invited and joined rooms can be left.
     #[doc(alias = "reject_invitation")]
     #[instrument(skip_all, fields(room_id = ?self.inner.room_id()))]
-    pub async fn leave(&self) -> Result<()> {
+    async fn leave_this(&self) -> Result<()> {
```
poljar:
```suggestion
    async fn leave_impl(&self) -> Result<()> {
```

crates/matrix-sdk/src/room/mod.rs
```diff
@@ -368,6 +368,71 @@ impl Room {
...
+                        rooms.push(room);
+                        cur_room = rooms.last().expect("rooms just pushed so can't be empty");
+                    }
+                    Err(e) => {
+                        debug!("Tombstone following terminated early: {e:?}");
+                        break 'leave_aggregate_loop;
+                    }
+                },
+            }
+        }
```
poljar: I think this whole section could be written like this:

```rust
let mut rooms: Vec<Room> = vec![self.clone()];
let mut current_room = self;

while let Some(predecessor) = current_room.predecessor_room() {
    let maybe_previous_room = current_room.client.get_room(&predecessor.room_id);

    if let Some(previous_room) = maybe_previous_room {
        rooms.push(previous_room.clone());
        current_room = rooms.last().expect("The room was just pushed to the vec");
    }
}
```

This removes the need for the breaks which might be a bit hard to follow since they are effectively gotos.

crates/matrix-sdk/src/room/mod.rs
```diff
@@ -368,6 +368,71 @@ impl Room {
...
     }
 
+    /// Leave this room and all predecessors.
+    /// If any room was in [`RoomState::Invited`] state, it'll also be forgotten
+    /// automatically.
+    ///
+    /// Only invited and joined rooms can be left.
+    pub async fn leave(&self) -> Result<()> {
+        let mut rooms: Vec<Room> = vec![self.clone()];
+        let mut cur_room = self;
```
poljar: `cur` isn't that shorter than `current`, let's avoid unnecessary abbreviations.

```suggestion
        let mut current_room = self;
```

crates/matrix-sdk/src/room/mod.rs
```diff
@@ -368,6 +368,71 @@ impl Room {
...
+            if batch.is_empty() {
+                break 'leave_action_loop;
+            }
+
+            let batch = batch
+                .iter()
+                .map(|room| match room.state() {
+                    RoomState::Joined => Ok(Some(room.leave_this())),
+                    RoomState::Invited => Ok(Some(room.leave_this())),
+                    RoomState::Knocked => Ok(Some(room.leave_this())),
```
poljar: Seems like `join_all!()` turned out to be the wrong pattern here.

Using the `buffered()` stream extension seems to produce more readable results:

```rust
let futures: Vec<_> = rooms
    .into_iter()
    .filter_map(|room| match room.state() {
        RoomState::Joined | RoomState::Invited | RoomState::Knocked => {
            Some(async move { room.leave_this().await })
        }
        RoomState::Left | RoomState::Banned => None,
    })
    .collect();

let mut stream = futures_util::stream::iter(futures).buffered(BATCH_SIZE);

for result in stream.next().await {
    match result {
        Ok(_) => todo!(),
        Err(_) => todo!(),
    }
}
```

crates/matrix-sdk/src/error.rs
```diff
@@ -276,6 +276,18 @@ impl RetryKind {
...
+    #[error("an error occurred during room leave")]
+    ErrorDuringRoomLeave,
+
+    /// Error doing a get_room(predecessor).
+    #[error("the predecessor is wrong or forgotten")]
+    PredecessorWrongOrForgotten,
+
+    /// Error doing room.leave()
+    #[error("the predecessor is not Joined or Left")]
+    PredecessorWrongState,
```
poljar: I don't think those errors should ever be exposed to the caller.

If you'd like to use errors somewhere you might want to introduce a separate error type for this, but you can also just log things directly where the error happens.

poljar (changes requested):
Hah, doing things concurrently isn't easy, and there are many ways to do this.

I think we can do a bit better.

crates/matrix-sdk/src/room/mod.rs
dragonfly1033: Thanks for the restructure. Would we not want to log if following the linked list finished early? Or do we just add it in the future if/when we need it for debugging?

crates/matrix-sdk/src/error.rs
dragonfly1033: The leave() returns a Result<(), Error> so which Error should it return to indicate failure to leave? Since it may have multiple Error's from leave_impl(), should I make a new error type which holds a Vec<Error> for leave failures?

dragonfly1033: Perhaps, I can return an error if and only if the room self fails and only log all other errors?

poljar: Yup, only returning the error if we can't leave the room we actually want to leave sounds good to me.

crates/matrix-sdk/src/room/mod.rs
poljar: You can add an else branch there that logs if we can't find the previous room.

crates/matrix-sdk/src/room/mod.rs
```diff
@@ -375,61 +381,50 @@ impl Room {
...
+
+        let mut maybe_this_room_failed_with: Option<Error> = None;
+
+        while let Some(result) = futures_stream.next().await {
+            if let (Err(e), room) = result {
+                if room.room_id() == self.room_id() {
+                    maybe_this_room_failed_with = Some(e);
+                    continue;
+                }
+                warn!("{e:?}");
```
poljar: We should explain what happened here, instead of just showing the error. The error will just show that a request failed with a specific reason, but we won't know why we made that request. 

This could also be put in an else branch which makes the `continue` in the if branch unnecessary.

crates/matrix-sdk/src/room/mod.rs
```diff
@@ -375,61 +381,50 @@ impl Room {
...
             }
         }
 
+        if let Some(error) = maybe_this_room_failed_with {
+            return Err(error);
+        }
+
```
poljar: This can be written as `maybe_this_room_failed_with.unwrap_or(Ok(()))`.

poljar (changes requested):
Almost there, a couple of nits left but I think that we're good after that.

crates/matrix-sdk/src/room/mod.rs
dragonfly1033: It would be ```map_or(Ok(()), |error| Err(error))``` I believe

poljar: Right, you need to wrap it into the `Result` enum since `maybe_this_room_failed` contains an `Error` type not a `Result` type.

poljar (changes requested):
This looks good now. Unfortunately there have been some breaking changes on main in the meantime.

Luckily they seem to be easy to address.

Could you also add a changelog entry to describe that the behavior of `leave()` changed a bit.

PR #5424, opened 2025-07-18
refactor: Clean up some uses of `Raw`.

zecakeh:
This will allow to have a slightly smaller diff when bumping Ruma.

It also adds `Account::fetch_account_data_static()` to allow to fetch account data using a statically-known type.

This can be reviewed commit by commit. The commit messages have more details about what is happening.

### DIFF
--- a/crates/matrix-sdk/src/encryption/recovery/types.rs
+++ b/crates/matrix-sdk/src/encryption/recovery/types.rs
@@ -13,10 +13,7 @@
 // limitations under the License.
 
 use matrix_sdk_base::crypto::store::types::RoomKeyCounts;
-use ruma::{
-    events::{EventContent, GlobalAccountDataEventType},
-    exports::ruma_macros::EventContent,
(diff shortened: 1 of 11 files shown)


crates/matrix-sdk/src/encryption/recovery/types.rs
```diff
@@ -115,12 +112,3 @@ pub(super) struct SecretStorageDisabledContent {}
 pub(super) struct BackupDisabledContent {
     pub disabled: bool,
 }
-
-impl BackupDisabledContent {
-    /// Get the event type of the [`BackupDisabledContent`] global account data
-    /// event.
-    pub(super) fn event_type() -> GlobalAccountDataEventType {
-        // This is dumb, there's got to be a better way to get to the event type?
-        Self { disabled: false }.event_type()
```
poljar, line 124: Hah, thanks for getting rid of this.

PR #5425, opened 2025-07-21
feat(sqlite): make sqlite's implementation of `load_all_chunks_metadata` even faster

bnjbvr:
https://github.com/matrix-org/matrix-rust-sdk/pull/5411 was already an immense improvement for the loading of the chunks metadata, but running the benchmark on my machine it can take up to 25 ms for a single room with 10K events, which is unfortunately still a lot (considering we can have hundreds/thousands of rooms).

The strategy used there is good:

1. select all chunks
2. for each event chunk, run a query to count the number of events

This implies that we'll have one additional `select` query per event chunk. This gave me the idea for this strategy:

- in a single query, get all the chunk ids and associated number of events, using a plain `COUNT` + `GROUP BY`, and put that into a hashmap (from chunk id to number of events in that chunk)
- in another query, get all the chunk ids / next / prev.

Then, when reforming the chunk metadata, use the hashmap to get the associated number of events. This is doing step 2 of the initial strategy in the app, to avoid running an extra `SELECT` query per event chunk.

This leads to the following improvement, as reported by `cargo bench` on the new benchmark:

```
reading/metadata/sqlite store/10000
                        time:   [1.9737 ms 2.0377 ms 2.0966 ms]
                        thrpt:  [4.7697 Melem/s 4.9075 Melem/s 5.0666 Melem/s]
                 change:
                        time:   [-92.816% -92.595% -92.388%] (p = 0.00 < 0.05)
                        thrpt:  [+1213.7% +1250.5% +1292.0%]
                        Performance has improved.
reading/metadata/sqlite store/100000
                        time:   [23.049 ms 23.256 ms 23.424 ms]
                        thrpt:  [4.2692 Melem/s 4.3000 Melem/s 4.3385 Melem/s]
                 change:
                        time:   [-99.277% -99.269% -99.260%] (p = 0.00 < 0.05)
                        thrpt:  [+13410% +13583% +13729%]
                        Performance has improved.
```

That is, it takes around 2ms for loading the metadata of a single room, instead of the initial 25ms that it took on my machine.

### DIFF
--- a/crates/matrix-sdk-sqlite/src/event_cache_store.rs
+++ b/crates/matrix-sdk-sqlite/src/event_cache_store.rs
@@ -14,7 +14,7 @@
 
 //! An SQLite-based backend for the [`EventCacheStore`].
 
-use std::{fmt, iter::once, path::Path, sync::Arc};
+use std::{collections::HashMap, fmt, iter::once, path::Path, sync::Arc};
 
 use async_trait::async_trait;
 use deadpool_sqlite::{Object as SqliteAsyncConn, Pool as SqlitePool, Runtime};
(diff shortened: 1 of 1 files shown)


crates/matrix-sdk-sqlite/src/event_cache_store.rs
```diff
@@ -920,11 +938,17 @@ impl EventCacheStore for SqliteEventCacheStore {
...
                 })?
                 .map(|data| -> Result<_> {
-                    let (id, previous, next, num_items) = data?;
+                    let (id, previous, next, chunk_type) = data?;
+
+                    let num_items = if chunk_type == CHUNK_TYPE_GAP_TYPE_STRING {
+                        0
+                    } else {
+                        num_events_by_chunk_ids.get(&id).copied().unwrap_or(0)
+                    };
```
poljar, line 956: Doesn't a gap have 0 events by definition? Making this if/else redundant?

bnjbvr: It is, but my local testing showed it's 1% slower to not have this branch. The theory is that, at least on my machine, it's faster to allocate the chunk type string and compare against it, than it is to do a hashmap lookup that results in a cache miss.

But 1% doesn't matter much, so happy to get rid of it if you prefer?

poljar: Nah it's fine. A comment explaining this might be nice though.

Hywan:
Fantastic!

Issue #5422, opened 2025-07-18
threaded timeline: responding to an in-thread message doesn't use the correct fallback


bnjbvr:
It seems that if one uses a threaded timeline, and answers an in-thread message, the reply fallback's event is set to the thread's root event, not the thread's latest event. It's unclear whether it's a bug in EXI or in the SDK, I'll investigate when I'm back.

bnjbvr:
When sending an in-thread message, that's not an explicit reply to another in-thread message, EXI will say it's a response to the thread root:

https://github.com/element-hq/element-x-ios/blob/57469fb8eb1476480c8c82b28de0ae5356abb005/ElementX/Sources/Services/Timeline/TimelineProxy.swift#L468-L472

It's incorrect: the replied-to event id should always be set to the latest in-thread message, for the fallback to work as intended.

Arguably, the API is bad here, so closing as a dup of #5216.

I think another action to take would be for EXI to fix the API usage as well, until #5216 is fixed, cc @stefanceriu 🥺

PR #5427, opened 2025-07-21
refactor(timeline): enhance the `Timeline::send()` and `Timeline::send_reply()` APIs when the timeline is threaded

### LINKED ISSUE #5216: timeline: improve API of sending thread replies
While implementing it in `multiverse`, I realized that sending a reply in a thread-focused timeline is kinda weird: one must use `Timeline::send_reply()` and enforce the thread relationship in there.

I would expect that, as a user of the threaded-timeline, all I need is use `Timeline::send` when I want to answer in that thread, and `Timeline::send_reply` when I want to send an explicit reply to another in-thread event. Now it means that `Timeline::send` would fill in the `m.relates_to` relationship field to the thread, *if and only it's missing*. This way, it's still possible to send other aggregations, notably edits, to an in-thread event (an edit sets `m.relates_to` to `replace`).

If we made the APIs like this, it would likely simplify `Timeline::send_reply` so it doesn't need the `Reply` parameters, and only take a simpler enum as it did before (enforce thread relationship or not — which would be ignored in the case of a thread timeline).

This means that to start a thread, one would need to open a thread-focused timeline, and then use that timeline's `send()` (vs what we have now: using the main timeline, use `send_reply` with a thread relationship in the `Reply` parameter).

I think it might be worth it, but I'm curious to get feedback about this idea; cc @stefanceriu @Johennes, what do you think of this?

bnjbvr:
This should make it simpler to send in-thread messages and in-thread replies in the future, but not mandating that the caller passes so much information.

- To send a message in a threaded timeline, one can now use `Timeline::send()`
- To send a reply in a threaded timeline, one can now use `Timeline::send_reply()` which only takes an `OwnedEventId` param instead of a `Reply`
- (To start a new thread, you'd create a threaded timeline and use `Timeline::send()`.)

The improvement is notable in the code of `multiverse`, for instance, where sending an event is now the same code between a regular or a threaded timeline.

If the threaded timeline isn't empty, it'll correctly use the latest in-thread event as the replied-to fallback event id (a check has been added in one test for that, and this avoids the mess observed in #5422). We could go one step further and trigger a pagination if the threaded timeline was empty; we'll see if we'll ever need it.

cc @Johennes, curious to get your thoughts on the new API!

Fixes #5216

### DIFF
--- a/crates/matrix-sdk-ui/src/timeline/mod.rs
+++ b/crates/matrix-sdk-ui/src/timeline/mod.rs
@@ -47,8 +47,12 @@ use ruma::{
         AnyMessageLikeEventContent, AnySyncTimelineEvent, Mentions,
         poll::unstable_start::{NewUnstablePollStartEventContent, UnstablePollStartEventContent},
         receipt::{Receipt, ReceiptThread},
+        relation::Thread,
         room::{
-            message::{FormattedBody, ReplyWithinThread, RoomMessageEventContentWithoutRelation},
+            message::{
+                FormattedBody, Relation, RelationWithoutReplacement, ReplyWithinThread,
+                RoomMessageEventContentWithoutRelation,
+            },
             pinned_events::RoomPinnedEventsEventContent,
         },
     },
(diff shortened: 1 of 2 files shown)


Johennes:
This sounds sensible and much easier than fiddling with `enforce_thread` or `reply_within_thread`. 👍

bnjbvr:
Stefan identified a few issues, so might as well add some tests later:

- [ ] add tests for stickers
- [ ] add tests for polls

crates/matrix-sdk-ui/src/timeline/mod.rs
```diff
@@ -270,18 +274,65 @@ impl Timeline {
...
     #[instrument(skip(self, content), fields(room_id = ?self.room().room_id()))]
-    pub async fn send(
-        &self,
-        content: AnyMessageLikeEventContent,
-    ) -> Result<SendHandle, RoomSendQueueError> {
-        self.room().send_queue().send(content).await
+    pub async fn send(&self, content: AnyMessageLikeEventContent) -> Result<SendHandle, Error> {
+        // If this is a room event we're sending in a threaded timeline, we add the
+        // thread relation ourselves.
+        if let AnyMessageLikeEventContent::RoomMessage(ref room_msg_content) = content
```
stefanceriu: For completeness: this is not going to work as expected for non-room messages i.e. polls which would just be sent in the main timeline no matter the orginating instance.

bnjbvr: Thanks! Fixed in the latest commit, and added new tests that failed before the changes in the timeline code :+1:

crates/matrix-sdk-ui/src/timeline/mod.rs
```diff
@@ -294,31 +298,50 @@ impl Timeline {
...
+                        )));
+                    }
+                }
+
+                AnyMessageLikeEventContent::Sticker(sticker) => {
+                    if let Some(thread_root) = self.controller.thread_root() {
+                        sticker.relates_to =
+                            Some(Relation::Thread(Thread::plain(thread_root, reply.event_id)));
+                    }
+                }
```
stefanceriu, line 338: My god this so much simpler than what I imagined from seeing what the `RoomMessage` `make_reply_event` does. Love it!

crates/matrix-sdk-ui/src/timeline/mod.rs
stefanceriu: Can confirm that it works properly now and that sending and interacting with polls works in threaded timelines! 👏

PR #5430, opened 2025-07-22
misc: Include file name when uploading unencrypted media

jmartinesp:
Changes:

- Allow uploading unencrypted media files with their original file name.
- Implement this for the serialized requests for the send queue too.
- Also add support for this in the media gallery APIs.
- Add a special case for thumbnails too so they can reference the original file name too.

Signed-off-by:


crates/matrix-sdk-base/src/media.rs
```diff
@@ -163,6 +170,10 @@ impl MediaEventContent for AudioMessageEventContent {
         Some(self.source.clone())
     }
 
+    fn filename(&self) -> Option<String> {
+        self.filename.clone()
```
bnjbvr: Unfortunately, I think this might need to be more complicated, see also: https://github.com/matrix-org/matrix-rust-sdk/blob/7c8f870d16581f21361db40896391f177a1fe36d/crates/matrix-sdk/src/room/mod.rs#L222-L228

crates/matrix-sdk/src/media.rs
```diff
@@ -748,8 +757,9 @@ impl Media {
 
         let (data, content_type, thumbnail_info) = thumbnail.into_parts();
 
+        let filename = filename.map(|name| format!("thumbnail-{name}"));
```
jmartinesp, line 744: I'm not sure if there would be a better way to reflect this is a thumbnail and at the same time reference the original file 🫤 .

crates/matrix-sdk/src/media.rs
```diff
@@ -200,13 +202,15 @@ impl Media {
         &self,
         content_type: &Mime,
         data: Vec<u8>,
+        filename: Option<String>,
         request_config: Option<RequestConfig>,
```
bnjbvr: Having two optional parameters in this method isn't pretty, because it's unclear what `None` is about. I think we can refactor this method, in the same way as we do for `Client.send()`, since this already returns a `SendMediaUploadRequest` object that can be configured:

- move the creation of the `media::create_content::v3::Request` to `SendMediaUploadRequest::into_future`. This means storing the request parameters + `&client` instance in the `SendMediaUploadRequest`.
- add a builder method on `SendMediaUploadRequest` to allow providing a custom `RequestConfig`; it'd be set to `None` in the ctor of `SendMediaUploadRequest`, and extracted and used in the actual request created in `into_future`
- add a builder method for providing a custom filename, the same way. 

How does that sound?

crates/matrix-sdk/src/media.rs
bnjbvr: What's the use case for providing a file for the thumbnail, out of curiosity? What values does it provide?

crates/matrix-sdk-base/src/media.rs
jmartinesp: I took that into account [here](https://github.com/matrix-org/matrix-rust-sdk/pull/5430/commits/e33c40eb1cb7cdc80f16576a5c2337e18509d31b#diff-d569f54c7901b5cd660aae77045c80dabbd3c251f0fa5891530469f35941327aR1296) for the send queue, although maybe I missed some case and I think the non-send queue part will just use the provided `filename` parameter.

---

EDIT: I ended up doing https://github.com/matrix-org/matrix-rust-sdk/pull/5430/commits/2e32c252e805639fdedf43ba784be736d4e1f414

crates/matrix-sdk/src/media.rs
jmartinesp: If you want to search for a certain filename to delete some media I guess you'd want to remove the associated thumbnail too. I don't know if the media server has a better way to handle deleting both, this might not be needed.

jmartinesp:
> In commit 3: why do we need to pass the filename for each media upload? Isn't the filename supposed to live in the media event content itself? I'm confused.

You mean in the bindings one? I added the new optional filename parameter because on Android i.e. when we pre-process some media for upload we create a temp file for it which will use an UUID for its filename and we don't want to upload the media with that filename, but with its original one `my-cat.jpg` instead.

crates/matrix-sdk/src/media.rs
jmartinesp: Sounds good, the API would be way better this way!

jmartinesp:
I'm about to rebase to fix the conflict, but nothing changed in the existing commits. A couple of new commits were added.

crates/matrix-sdk-base/src/media.rs
```diff
@@ -142,6 +142,9 @@ pub trait MediaEventContent {
     /// Returns `None` if `Self` has no file.
     fn source(&self) -> Option<MediaSource>;
 
+    /// Get the name of the uploaded file for `Self`.
+    fn filename_or_body(&self) -> Option<String>;
```
bnjbvr, line 146: I don't think this is a good API; a final user wants the filename, or the caption, but they shouldn't have to know about the Matrix Client-Server API to know how to compute one or the other.

Instead, I'd recommend having both `fn filename()` and `fn caption()` that do the right thing, and figure the expected value, according to the specification.

crates/matrix-sdk-ui/src/timeline/mod.rs
```diff
@@ -821,21 +821,24 @@ pub enum AttachmentSource {
     /// An attachment loaded from a file.
     ///
     /// The bytes and the filename will be read from the file at the given path.
-    File(PathBuf),
+    File { filename: Option<String>, path: PathBuf },
```
bnjbvr, line 824: Why to require that the filename be provided in addition to the `PathBuf`, which also includes the filename?

crates/matrix-sdk/src/client/futures.rs
```diff
@@ -179,14 +212,24 @@ impl SendMediaUploadRequest {
...
 
     /// Get a subscriber to observe the progress of sending the request
     /// body.
     pub fn subscribe_to_send_progress(&self) -> Subscriber<TransmissionProgress> {
-        self.send_request.send_progress.subscribe()
+        self.progress_observable.subscribe()
+    }
+
+    /// Creates the [`SendRequest`] using the provided parameters.
+    pub fn build_send_request(self) -> SendRequest<media::create_content::v3::Request> {
```
bnjbvr, line 226: I don't think this should be `pub`, in fact it could be inlined in the `into_future` impl instead

crates/matrix-sdk/src/client/futures.rs
```diff
@@ -156,17 +156,50 @@ where
...
+            request: media::create_content::v3::Request::new(file),
+            request_config: Default::default(),
+            progress_observable: SharedObservable::new(TransmissionProgress::default()),
+        }
+    }
+
+    /// Returns the data to upload.
+    pub fn data(&self) -> &Vec<u8> {
+        &self.request.file
+    }
```
bnjbvr, line 185: Don't think it should be needed? (If it is, then it should return `&[u8]` by convention)

crates/matrix-sdk/src/media.rs
```diff
@@ -165,53 +165,40 @@ impl Media {
...
-        request_config: Option<RequestConfig>,
-    ) -> SendMediaUploadRequest {
-        let request_config = request_config.unwrap_or_else(|| {
-            self.client.request_config().timeout(Self::reasonable_upload_timeout(&data))
-        });
-
-        let request = assign!(media::create_content::v3::Request::new(data), {
-            content_type: Some(content_type.essence_str().to_owned()),
+        send_media_upload_request: SendMediaUploadRequest,
+    ) -> Result<media::create_content::v3::Response, Error> {
```
bnjbvr, line 195: This is not right. Instead, it should take the basic parameters that are always needed for a `SendMediaUploadRequest`, and return it. As a result, it should not be async, and it should resolve the response. The way to use it would be:

```
client.media().upload(file_bytes_vec).with_request_config(RequestConfig::default()).await?;
```

and so on.

It may be better to split this change into a separate PR? The history of this one is already a bit messy, and this is a problem orthogonal to the one at hand. (I'd be happy reviewing the split up PR too, if that helps speeding up the review!)
```

File: crates/matrix-sdk/src/media.rs (hunk 2 above)
Comment (bnjbvr) [reply crates/matrix-sdk/src/media.rs:744]: I think we'd need to take an optional thumbnail filename, to make it entirely user controlled, and not something based on our internals. What do you think? (can come as a subsequent PR/commit too)

File: bindings/matrix-sdk-ffi/src/client.rs [hunk 9]
```diff
@@ -1026,21 +1026,24 @@ impl Client {
         &self,
         mime_type: String,
         data: Vec<u8>,
+        filename: Option<String>,
         progress_watcher: Option<Box<dyn ProgressWatcher>>,
     ) -> Result<String, ClientError> {
         let mime_type: mime::Mime = mime_type.parse().context("Parsing mime type")?;
-        let request = self.inner.media().upload(&mime_type, data, None);
+        let send_media_request = SendMediaUploadRequest::new((*self.inner).clone(), data)
+            .with_content_type(mime_type.essence_str().to_owned())
```
bnjbvr, line 1034: It's surprising that `with_content_type()` doesn't take the same parameter type as the previous `&mime_type` in the previous call site. Can we make it so that it does, please?

bindings/matrix-sdk-ffi/src/timeline/mod.rs
```diff
@@ -219,7 +219,11 @@ pub enum UploadSource {
     /// Upload source is a file on disk
     File {
         /// Path to file
-        filename: String,
+        path: String,
+
+        /// An optional filename, if the one in the path is not the one we want
+        /// to use for the file.
+        filename: Option<String>,
```
bnjbvr, line 226: Again, not sure why the path's filename isn't sufficient here…

crates/matrix-sdk-base/src/store/send_queue.rs
```diff
@@ -109,6 +109,9 @@ pub enum QueuedRequestKind {
         #[cfg(feature = "unstable-msc4274")]
         #[serde(default)]
         accumulated: Vec<AccumulatedSentMediaInfo>,
+
+        /// The name of the file to upload, if known or public.
```
bnjbvr, line 113:
```suggestion
        /// The name of the file to upload.
```

crates/matrix-sdk-base/src/store/send_queue.rs
```diff
@@ -241,6 +244,9 @@ pub enum DependentQueuedRequestKind {
         #[cfg(feature = "unstable-msc4274")]
         #[serde(default = "default_parent_is_thumbnail_upload")]
         parent_is_thumbnail_upload: bool,
+
+        /// The name of the file to upload, if known.
```
bnjbvr, line 248:
```suggestion
        /// The name of the file to upload.
```

crates/matrix-sdk/src/client/futures.rs
```diff
@@ -156,17 +156,50 @@ where
...
+    /// The [`RequestConfig`] to use for uploading the media file.
+    pub request_config: Option<RequestConfig>,
+    progress_observable: SharedObservable<TransmissionProgress>,
 }
 
 impl SendMediaUploadRequest {
-    pub fn new(request: SendRequest<media::create_content::v3::Request>) -> Self {
-        Self { send_request: request }
+    /// Creates a new instance.
+    pub fn new(client: Client, file: Vec<u8>) -> Self {
```
bnjbvr, line 173: (This ctor shouldn't be public, if taking into account the change I suggested with respect to `Media::upload()` returning the `SendMediaUploadRequest`)

crates/matrix-sdk/src/send_queue/mod.rs
```diff
@@ -1873,7 +1906,9 @@ impl QueueStorage {
                 related_to,
                 #[cfg(feature = "unstable-msc4274")]
                 parent_is_thumbnail_upload,
+                filename,
             } => {
+                warn!("Saving send queue request for filename {filename:?}");
```
bnjbvr, line 1911: nit: debug `warn!` remaining here

crates/matrix-sdk/src/test_utils/mocks/mod.rs
```diff
@@ -2624,6 +2628,13 @@ impl<'a> MockEndpoint<'a, UploadEndpoint> {
     /// Returns a upload endpoint that emulates success, i.e. the media has been
     /// uploaded to the media server and can be accessed using the given
     /// event has been sent with the given [`MxcUri`].
+    /// Expect the filename query param to match the provided value.
+    pub fn expect_filename(self, filename: &str) -> Self {
```
bnjbvr, line 2632: I've started to use the `match_` prefix to mean "a parameter must match this value", so lightly suggesting to do it here too. (Strictly speaking it's not "expecting" as it's not asserting; if the filename isn't provided, the mock will silently not catch the request, and it may run into a 404 instead)

…but i realize the equivalent for mime type is called `expect_mime_type`, so it's fun to use `expect_` here, or better to rename the other one to `match_mime_type` as well. Your choice :)

bnjbvr (changes requested):
Sorry for the long review time! I really wanted the media upload progress changes to land first, as the progress PR was slightly bigger and older.

This PR is touching many files, and I think some of my suggestions have been lost in translation unfortunately, so if you're in the mood, it might be sweet to split it up.

jmartinesp:
This PR has been open for a long time and I sadly don't have the time to finish it, so I'll close it for now.

PR #5431, opened 2025-07-22
Bump the base crate to 2024 edition and use let-chains in there

bnjbvr:
…which requires the base crate to use the 2024 edition.

### DIFF
--- a/crates/matrix-sdk-base/src/read_receipts.rs
+++ b/crates/matrix-sdk-base/src/read_receipts.rs
@@ -264,15 +264,15 @@ impl RoomReadReceipts {
             // Sliding sync sometimes sends the same event multiple times, so it can be at
             // the beginning and end of a batch, for instance. In that case, just reset
             // every time we see the event matching the receipt.
-            if let Some(event_id) = event.event_id() {
-                if event_id == receipt_event_id {
-                    // Bingo! Switch over to the counting state, after resetting the
-                    // previous counts.
-                    trace!("Found the event the receipt was referring to! Starting to count.");
-                    self.reset();
-                    counting_receipts = true;
-                    continue;
-                }
+            if let Some(event_id) = event.event_id()
+                && event_id == receipt_event_id
+            {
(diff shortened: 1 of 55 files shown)


bnjbvr:
Reverting to draft at the moment; we should first land the Ruma bump, otherwise this will cause some unnecessary rebase pain for the author.

crates/matrix-sdk-base/src/room/room_info.rs
```diff
@@ -324,25 +324,42 @@ impl BaseRoomInfo {
...
-            self.history_visibility.as_mut().unwrap().redact(&room_version);
-        } else if self.join_rules.has_event_id(redacts) {
-            self.join_rules.as_mut().unwrap().redact(&room_version);
-        } else if self.name.has_event_id(redacts) {
-            self.name.as_mut().unwrap().redact(&room_version);
-        } else if self.tombstone.has_event_id(redacts) {
-            self.tombstone.as_mut().unwrap().redact(&room_version);
-        } else if self.topic.has_event_id(redacts) {
-            self.topic.as_mut().unwrap().redact(&room_version);
+        if let Some(ev) = self.avatar.as_mut()
```
jplatte:
```suggestion
        if let Some(ev) = &mut self.avatar
```

crates/matrix-sdk-base/src/read_receipts.rs
```diff
@@ -388,16 +388,15 @@ impl ReceiptSelector {
         for (event_id, receipts) in &receipt_event.0 {
             for ty in [ReceiptType::Read, ReceiptType::ReadPrivate] {
                 if let Some(receipt) = receipts.get(&ty).and_then(|receipts| receipts.get(user_id))
```
jplatte:
```suggestion
                if let Some(receipts) = receipts.get(&ty)
                	&& Some(receipt) = receipts.get(user_id)
```

jplatte (approved):
Other than CI failing due to the no-longer-used `OptionExt` trait, and my (very much not important) nit above, LGTM.

PR #5434, opened 2025-07-22
contrib: Set our check command to check instead of clippy

poljar:
Clippy is a bit too heavy to be used as the check on save command in this repo.

Let's set a better default.

### DIFF
--- a/contrib/ide/vscode/settings.json
+++ b/contrib/ide/vscode/settings.json
@@ -1,4 +1,4 @@
 {
-  "rust-analyzer.checkOnSave.command": "clippy",
+  "rust-analyzer.checkOnSave.command": "check",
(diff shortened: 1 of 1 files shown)


jplatte:
Hm, clippy, especially recent versions, shouldn't be much heavier than `check`.
See also https://blog.goose.love/posts/clippy-performance-status-update/

poljar:
> Hm, clippy, especially recent versions, shouldn't be much heavier than `check`. See also https://blog.goose.love/posts/clippy-performance-status-update/

One of our interns had to change the check on save command to get their IDE usable. This PR is a reaction to this.

While it isn't problematic for me either I think many people might still have trouble with it.

Also, that background on the blogpost is impressively distracting.

PR #5436, opened 2025-07-23
Upgrade Ruma (phase 3)

zecakeh:
Follow-up to #5433.

This upgrade introduces support for [room version 12](https://matrix.org/blog/2025/07/security-predisclosure/). If clients want to test their support on servers that use the unstable `org.matrix.hydra.11` version (like beta.matrix.org), the `unstable-hydra` feature of Ruma must be enabled.

Split into commits by theme for easier review, however they should be squashed together when merging.


crates/matrix-sdk/src/room/member.rs
```diff
@@ -95,6 +98,17 @@ impl RoomMember {
 #[derive(Clone, Debug, PartialEq)]
 #[cfg_attr(feature = "uniffi", derive(uniffi::Enum))]
 pub enum RoomMemberRole {
+    /// The member is a creator.
+    ///
+    /// A creator has infinite power levels and cannot be demoted, so this role
```
jplatte:
```suggestion
    /// A creator has an infinite power level and cannot be demoted, so this role
```


Every user has only one power level, right? x)

zecakeh: Right, fixed.

crates/matrix-sdk/src/room/member.rs
```diff
@@ -105,22 +119,29 @@ pub enum RoomMemberRole {
...
+            UserPowerLevel::Int(value) => {
+                if value >= int!(100) {
+                    Self::Administrator
+                } else if value >= int!(50) {
+                    Self::Moderator
+                } else {
+                    Self::User
+                }
+            }
+            _ => unimplemented!(),
```
poljar, line 138: Hmm, is there a way to remind us that we have new variants? I wonder if we'll forget to adapt this if Ruma ever gets new variants.

bindings/matrix-sdk-ffi/src/room_member.rs
```diff
@@ -130,3 +136,40 @@ impl TryFrom<matrix_sdk::room::RoomMemberWithSenderInfo> for RoomMemberWithSende
...
+    /// The user has the given power level.
+    Value { value: i64 },
+}
+
+impl From<UserPowerLevel> for PowerLevel {
+    fn from(value: UserPowerLevel) -> Self {
+        match value {
+            UserPowerLevel::Infinite => Self::Infinite,
+            UserPowerLevel::Int(value) => Self::Value { value: value.into() },
+            _ => unimplemented!(),
```
poljar: Same for this place, can we do something better here?

zecakeh: We could change the `From` to a `TryFrom` and return an error in this case?

poljar: Yeah, let's do that. I don't really want app crashes because we forgot about this.

crates/matrix-sdk/src/room/member.rs
jplatte: There's the `non_exhaustive_omitted_patterns` lint in rustc, but it's still unstable. I'm trying to fix the one known issue with it in https://github.com/rust-lang/rust/pull/144331, afterwards I might send a stabilization PR :)

jmartinesp:
I'm still testing some of the changes, but so far these changes seem to work fine on Element X Android. Thanks!

crates/matrix-sdk/src/room/member.rs
poljar: Hey I remember this lint. Yeah, that one would be great.

jplatte: Originally implemented by Devin in one of his GSoC projects at Ruma :)

zecakeh: What should we do in the meantime? The compiler will complain with Rust stable if we use unstable features, right?

poljar: > Originally implemented by Devin in one of his GSoC projects at Ruma :)

Hah right.

jplatte: Could do something like `#![cfg_attr(feature = "nightly", feature(non_exhaustive_omitted_patterns_lint))]` (or a `cfg` that's enabled via `build.rs` instead of a feature; can't use `rustversion` though because you can't use a proc-macro as an inner attribute).

poljar: > What should we do in the meantime? The compiler will complain with Rust stable if we use unstable features, right?

Let's add a TODO and ignore the issue till the lint is ready.

Hopefully it's unlikely that we're going to add more variants before the lint is stabilized.

bindings/matrix-sdk-ffi/src/room_member.rs
zecakeh: Done.

crates/matrix-sdk/src/room/member.rs
zecakeh: Comment added.

poljar (approved):
Looks good. Once we get rid of the merge conflicts we can get this in.

poljar (approved):

PR #5437, opened 2025-07-23
refactor: Add `read_timeout` for HTTP requests, make `timeout` optional

jmartinesp:
Having the `read_timeout` value set allows us to have long network requests that will get cancelled automatically when no data is received for a certain amount of time.

If we combine it with being able to remove the existing default `timeout` value for some requests, it would fix the issue with downloading large files in poor network connectivity scenarios where the download would just cancel automatically after the timeout (30s) was reached.

Signed-off-by:

### DIFF
--- a/crates/matrix-sdk/src/client/mod.rs
+++ b/crates/matrix-sdk/src/client/mod.rs
@@ -20,6 +20,7 @@ use std::{
     future::{ready, Future},
     pin::Pin,
     sync::{Arc, Mutex as StdMutex, RwLock as StdRwLock, Weak},
+    time::Duration,
 };
 
 use caches::ClientCaches;
@@ -2312,7 +2313,8 @@ impl Client {
         });
         let mut request_config = self.request_config();
         if let Some(timeout) = sync_settings.timeout {
-            request_config.timeout = Some(timeout);
+            let base_timeout = request_config.timeout.unwrap_or(Duration::from_secs(30));
+            request_config.timeout = Some(base_timeout + timeout);
         }
 
         let response = self.send(request).with_request_config(request_config).await?;
--- a/crates/matrix-sdk/src/client/mod.rs
+++ b/crates/matrix-sdk/src/client/mod.rs
@@ -2312,7 +2312,7 @@ impl Client {
         });
         let mut request_config = self.request_config();
         if let Some(timeout) = sync_settings.timeout {
-            request_config.timeout += timeout;
+            request_config.timeout = Some(timeout);
         }
 
         let response = self.send(request).with_request_config(request_config).await?;
--- a/bindings/matrix-sdk-ffi/src/client_builder.rs
+++ b/bindings/matrix-sdk-ffi/src/client_builder.rs
@@ -133,6 +133,11 @@ pub struct ClientBuilder {
     threads_enabled: bool,
 }
 
+/// The timeout applies to each read operation, and resets after a successful
+/// read. This is more appropriate for detecting stalled connections when the
+/// size isn’t known beforehand.
+const DEFAULT_READ_TIMEOUT: Duration = Duration::from_secs(60);
+
 #[matrix_sdk_ffi_macros::export]
 impl ClientBuilder {
     #[uniffi::constructor]
@@ -539,6 +544,7 @@ impl ClientBuilder {
             if let Some(timeout) = config.timeout {
                 updated_config = updated_config.timeout(Duration::from_millis(timeout));
             }
+            updated_config = updated_config.read_timeout(DEFAULT_READ_TIMEOUT);
             if let Some(max_concurrent_requests) = config.max_concurrent_requests {
                 if max_concurrent_requests > 0 {
                     updated_config = updated_config.max_concurrent_requests(NonZeroUsize::new(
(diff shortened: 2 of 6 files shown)


crates/matrix-sdk/src/client/mod.rs
```diff
@@ -2312,7 +2312,7 @@ impl Client {
         });
         let mut request_config = self.request_config();
         if let Some(timeout) = sync_settings.timeout {
-            request_config.timeout += timeout;
+            request_config.timeout = Some(timeout);
```
poljar: This is not quite the same. If we set the poll timeout of the sync to 30 seconds and the timeout itself was 30, then we would have had 60 here.

Instead now the connection aborts right when the poll timeout expires. I think this potentially breaks syncs for everybody.

You'll have to `unwrap_or_default()` the timeout here.

poljar (changes requested):
I think this has one small problem.

The rest looks fine.

crates/matrix-sdk/src/client/mod.rs
jmartinesp: Oh, true. For some reason I didn't realise what this was doing was just adding the 2 durations.

crates/matrix-sdk/src/client/mod.rs
```diff
@@ -2312,7 +2312,7 @@ impl Client {
         });
         let mut request_config = self.request_config();
         if let Some(timeout) = sync_settings.timeout {
-            request_config.timeout = Some(timeout);
+            request_config.timeout = Some(request_config.timeout.unwrap_or_default() + timeout);
```
poljar: Hmm, the default here would be 0, let's `unwrap()` into a `Duration` of 30s instead.

Sorry I didn't realize that the default value wouldn't work either in my first comment.

poljar (changes requested):
Sorry, I think we need a better default yet :sweat_smile:

crates/matrix-sdk/src/client/mod.rs
jmartinesp: No worries, it should be fixed in https://github.com/matrix-org/matrix-rust-sdk/pull/5437/commits/d8ecab0cfa55307c4b0622f61acf828634e2a247.

poljar:
Would you like to get rid of the fixup commit or should I just squash everything?

poljar:
Also, would you like to mention this as a bugfix in the changelog? Might make sense to mention this.

jmartinesp:
I've already added a changelog entry and squashed the fixup 👍 .

frebib:
Would this apply to sync too? I ask because if you've ever tried to use EX on an airplane, you'd understand what "unusable network" means 😂 Sync is also affected by this too in extreme cases

jmartinesp:
> Would this apply to sync too? I ask because if you've ever tried to use EX on an airplane, you'd understand what "unusable network" means 😂 Sync is also affected by this too in extreme cases

For `/sync` we keep the existing timeout strategy, with the extra read timeout.

poljar:
> Would this apply to sync too? I ask because if you've ever tried to use EX on an airplane, you'd understand what "unusable network" means 😂 Sync is also affected by this too in extreme cases

You mean that 30 seconds isn't enough to get a sync response?

This `read_timeout` applies to every request now, if no data is received in 60 seconds we abort and possibly retry the request.

Though the `/sync` request has an additional shorter timeout. We might want to get rid of this shorter timeout but I think it would be better if we could configure `read_timeout` on a per-request basis before this happens.

In other words we'd like: https://github.com/seanmonstar/reqwest/pull/2764.

frebib:
> > Would this apply to sync too? I ask because if you've ever tried to use EX on an airplane, you'd understand what "unusable network" means 😂 Sync is also affected by this too in extreme cases
> 
> You mean that 30 seconds isn't enough to get a sync response?

In my experience, no. Provided the `read_timeout` is never hit, I don't see a good reason to timeout the sync request to allow for slow networks. Several times I've been on extremely unusably slow networks (think bytes per second with big pauses of no data at all) where EX has never been able to sync for the duration of the flight, but both Signal and WhatsApp are able to sync eventually after several minutes.

See https://github.com/element-hq/element-x-android/issues/1841

poljar:
> > > Would this apply to sync too? I ask because if you've ever tried to use EX on an airplane, you'd understand what "unusable network" means 😂 Sync is also affected by this too in extreme cases
> > 
> > 
> > You mean that 30 seconds isn't enough to get a sync response?
> 
> In my experience, no. Provided the `read_timeout` is never hit, I don't see a good reason to timeout the sync request to allow for slow networks. Several times I've been on extremely unusably slow networks (think bytes per second with big pauses of no data at all) where EX has never been able to sync for the duration of the flight, but both Signal and WhatsApp are able to sync eventually after several minutes.

Well that would explain why it doesn't work for you, if the request doesn't finish we abort after 30 seconds no matter if there's still data coming in.

The `read_timeout` is a relatively recent addition to the `reqwest` crate.  It only kills the connection if not data is being received for a certain amount of time.

We should probably migrate away from `timeout` for most requests if not for all of them.

frebib:
> We should probably migrate away from `timeout` for most requests if not for all of them.

Yeah that sounds like a good improvement! It would certainly fix/prevent https://github.com/element-hq/element-x-android/issues/1841 in the future
Despite that issue being closed, I'm pretty sure this is still a problem because of the 30s timeout

PR #5438, opened 2025-07-23
FFI: Update `Client::get_url` to return raw data and to throw on HTTP errors.

pixlwave:
`Client::get_url` is there for SDK consumers to be able to use the existing HTTP stack (configuration and all) however in practice it was a bit weird:
- Responses came back as strings limiting the types of resource that could be fetched (as well as requiring the string to be converted back to data before handed to a JSON decoder).
- HTTP errors weren't being raised and instead you would find the (e.g. 404) error response in the `Ok` case.

This PR fixes both of these issues.


stefanceriu:
Having talked about this alredy with Doug offline I believe somebody else should take the review here.

PR #5439, opened 2025-07-23
feat(sdk): add support for thread subscription endpoints (MSC4306)

bnjbvr:
This implements the three new endpoints introduced by [MSC4306](https://github.com/matrix-org/matrix-spec-proposals/pull/4306), with tests and support in multiverse.

Depends on https://github.com/ruma/ruma/pull/2169 and https://github.com/element-hq/synapse/pull/18722.
Part of #4869.

### DIFF
--- a/bindings/matrix-sdk-ffi/src/room/mod.rs
+++ b/bindings/matrix-sdk-ffi/src/room/mod.rs
@@ -1139,6 +1139,59 @@ impl Room {
 
         Ok(Arc::new(RoomPreview::new(AsyncRuntimeDropped::new(client), room_preview)))
     }
+
+    /// Toggle a MSC4306 subscription to a thread in this room, based on the
+    /// thread root event id.
+    ///
+    /// If `subscribed` is `true`, it will subscribe to the thread, with a
+    /// precision that the subscription was manually requested by the user
+    /// (i.e. not automatic).
+    ///
+    /// If the thread was already subscribed to (resp. unsubscribed from), while
+    /// trying to subscribe to it (resp. unsubscribe from it), it will do
+    /// nothing, i.e. subscribing (resp. unsubscribing) to a thread is an
+    /// idempotent operation.
+    pub async fn set_thread_subscription(
+        &self,
+        thread_root_event_id: String,
+        subscribed: bool,
+    ) -> Result<(), ClientError> {
+        let thread_root = EventId::parse(thread_root_event_id)?;
+        if subscribed {
+            // This is a manual subscription.
+            let automatic = false;
+            self.inner.subscribe_thread(thread_root, automatic).await?;
+        } else {
+            self.inner.unsubscribe_thread(thread_root).await?;
+        }
+        Ok(())
+    }
+
+    /// Return the current MSC4306 thread subscription for the given thread root
+    /// in this room.
+    ///
+    /// Returns `None` if the thread doesn't exist, or isn't subscribed to, or
+    /// the server can't handle MSC4306; otherwise, returns the thread
+    /// subscription status.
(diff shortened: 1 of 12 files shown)


crates/matrix-sdk/tests/integration/room/thread.rs
```diff
@@ -0,0 +1,57 @@
...
+
+    // I can subscribe to a thread.
+    room.subscribe_thread(root_id.clone(), true).await.unwrap();
+
+    server
+        .mock_get_thread_subscription()
+        .match_room_id(room_id.to_owned())
+        .match_thread_id(root_id.clone())
+        .ok(true)
+        .mock_once()
```
andybalaam, line 32: I see a `mock_once` here and then two calls to `fetch_thread_subscription` - is this OK?

bindings/matrix-sdk-ffi/src/room/mod.rs
```diff
@@ -1139,6 +1139,59 @@ impl Room {
...
+    ///
+    /// If `subscribed` is `true`, it will subscribe to the thread, with a
+    /// precision that the subscription was manually requested by the user
+    /// (i.e. not automatic).
+    ///
+    /// If the thread was already subscribed to (resp. unsubscribed from), while
+    /// trying to subscribe to it (resp. unsubscribe from it), it will do
+    /// nothing, i.e. subscribing (resp. unsubscribing) to a thread is an
+    /// idempotent operation.
+    pub async fn toggle_thread_subscription(
```
andybalaam: "toggle" to me means flip to true if it's false or to false if it's true.

So I'd suggest `set_thread_subscription`.

bindings/matrix-sdk-ffi/src/room/mod.rs
```diff
@@ -1139,6 +1139,59 @@ impl Room {
...
+        &self,
+        thread_root_event_id: String,
+        subscribed: bool,
```
andybalaam, line 1157: Why are we not providing a way to say whether this is automatic or not? Won't clients need to do that e.g. if the user is mentioned?

andybalaam (approved):
Looks good - some small questions. Thanks!

crates/matrix-sdk/tests/integration/room/thread.rs
bnjbvr: Second call is supposed to fail with a 404, so we don't need to set up an endpoint for it.

bindings/matrix-sdk-ffi/src/room/mod.rs
bnjbvr: By definition, the automatic subscriptions are caused by the SDK, not external clients; at least that's my current understanding of the MSC.

andybalaam: Ah right, makes sense.

crates/matrix-sdk/CHANGELOG.md
```diff
@@ -8,6 +8,11 @@ All notable changes to this project will be documented in this file.
 
 ### Features
 
+- Add experimental support for
+  [MSC4306](https://github.com/matrix-org/matrix-spec-proposals/pull/4306), with the
+  `Room::fetch_thread_subscription()`, `Room::subscribe_thread()` and `Room::unsubscribe_thread()`
+  methods.
+  ([#5442](https://github.com/matrix-org/matrix-rust-sdk/pull/5442))
```
Hywan, line 15: It's 5439, not 5442 :-D.

bnjbvr:
Feel free to PR :P

PR #5441, opened 2025-07-24
feat(push): add experimental support for MSC3768 (in-app-only notifications)

Johennes:
This adds experimental support for [MSC3768](https://github.com/matrix-org/matrix-spec-proposals/pull/3768) by introducing the push rule action `NotifyInApp` and a room notification mode that is similar to mentions & keywords but notifies in-app rather than muting other notifications.

Signed-off-by:

### DIFF
--- a/bindings/matrix-sdk-ffi/src/notification_settings.rs
+++ b/bindings/matrix-sdk-ffi/src/notification_settings.rs
@@ -326,7 +326,7 @@ pub enum Action {
     /// remote / push).
     Notify,
     /// Causes matching events to generate an in-app notification but no remote
-    /// / push notification.
+    /// (push) notification.
     #[cfg(feature = "unstable-msc3768")]
     NotifyInApp,
     /// Sets an entry in the 'tweaks' dictionary sent to the push gateway.
--- a/bindings/matrix-sdk-ffi/src/notification_settings.rs
+++ b/bindings/matrix-sdk-ffi/src/notification_settings.rs
@@ -370,11 +370,12 @@ pub enum RoomNotificationMode {
     /// Receive remote and in-app notifications for all messages.
     AllMessages,
     /// Receive remote and in-app notifications for mentions and keywords only.
-    MentionsAndKeywordsOnly,
-    /// Receive remote and in-app notifications for mentions and keywords and
-    /// in-app notifications only for other room messages.
-    #[cfg(feature = "unstable-msc3768")]
-    MentionsAndKeywordsOnlyTheRestInApp,
+    MentionsAndKeywordsOnly {
+        /// If true, receive in-app-only notifications (no pushes) for other
+        /// room messages. Otherwise, mute all other room messages.
+        #[cfg(feature = "unstable-msc3768")]
+        notify_in_app: bool,
+    },
     /// Do not receive any notifications.
     Mute,
 }
@@ -383,11 +384,13 @@ impl From<SdkRoomNotificationMode> for RoomNotificationMode {
     fn from(value: SdkRoomNotificationMode) -> Self {
         match value {
             SdkRoomNotificationMode::AllMessages => Self::AllMessages,
-            SdkRoomNotificationMode::MentionsAndKeywordsOnly => Self::MentionsAndKeywordsOnly,
-            #[cfg(feature = "unstable-msc3768")]
-            SdkRoomNotificationMode::MentionsAndKeywordsOnlyTheRestInApp => {
-                Self::MentionsAndKeywordsOnlyTheRestInApp
-            }
+            SdkRoomNotificationMode::MentionsAndKeywordsOnly {
+                #[cfg(feature = "unstable-msc3768")]
+                notify_in_app,
+            } => Self::MentionsAndKeywordsOnly {
+                #[cfg(feature = "unstable-msc3768")]
+                notify_in_app,
+            },
             SdkRoomNotificationMode::Mute => Self::Mute,
         }
     }
@@ -397,11 +400,13 @@ impl From<RoomNotificationMode> for SdkRoomNotificationMode {
     fn from(value: RoomNotificationMode) -> Self {
         match value {
             RoomNotificationMode::AllMessages => Self::AllMessages,
-            RoomNotificationMode::MentionsAndKeywordsOnly => Self::MentionsAndKeywordsOnly,
-            #[cfg(feature = "unstable-msc3768")]
-            RoomNotificationMode::MentionsAndKeywordsOnlyTheRestInApp => {
-                Self::MentionsAndKeywordsOnlyTheRestInApp
-            }
+            RoomNotificationMode::MentionsAndKeywordsOnly {
+                #[cfg(feature = "unstable-msc3768")]
+                notify_in_app,
+            } => Self::MentionsAndKeywordsOnly {
+                #[cfg(feature = "unstable-msc3768")]
+                notify_in_app,
+            },
             RoomNotificationMode::Mute => Self::Mute,
         }
     }
--- a/bindings/matrix-sdk-ffi/src/notification_settings.rs
+++ b/bindings/matrix-sdk-ffi/src/notification_settings.rs
@@ -370,12 +370,11 @@ pub enum RoomNotificationMode {
     /// Receive remote and in-app notifications for all messages.
     AllMessages,
     /// Receive remote and in-app notifications for mentions and keywords only.
-    MentionsAndKeywordsOnly {
-        /// If true, receive in-app-only notifications (no pushes) for other
-        /// room messages. Otherwise, mute all other room messages.
-        #[cfg(feature = "unstable-msc3768")]
-        notify_in_app: bool,
-    },
+    MentionsAndKeywordsOnly,
+    /// Receive remote and in-app notifications for mentions and keywords and
+    /// in-app notifications only for other room messages.
+    #[cfg(feature = "unstable-msc3768")]
+    MentionsAndKeywordsOnlyTheRestInApp,
     /// Do not receive any notifications.
     Mute,
 }
@@ -384,13 +383,11 @@ impl From<SdkRoomNotificationMode> for RoomNotificationMode {
     fn from(value: SdkRoomNotificationMode) -> Self {
         match value {
             SdkRoomNotificationMode::AllMessages => Self::AllMessages,
-            SdkRoomNotificationMode::MentionsAndKeywordsOnly {
-                #[cfg(feature = "unstable-msc3768")]
-                notify_in_app,
-            } => Self::MentionsAndKeywordsOnly {
-                #[cfg(feature = "unstable-msc3768")]
-                notify_in_app,
-            },
+            SdkRoomNotificationMode::MentionsAndKeywordsOnly => Self::MentionsAndKeywordsOnly,
+            #[cfg(feature = "unstable-msc3768")]
+            SdkRoomNotificationMode::MentionsAndKeywordsOnlyTheRestInApp => {
+                Self::MentionsAndKeywordsOnlyTheRestInApp
+            }
             SdkRoomNotificationMode::Mute => Self::Mute,
         }
     }
@@ -400,13 +397,11 @@ impl From<RoomNotificationMode> for SdkRoomNotificationMode {
     fn from(value: RoomNotificationMode) -> Self {
         match value {
             RoomNotificationMode::AllMessages => Self::AllMessages,
-            RoomNotificationMode::MentionsAndKeywordsOnly {
-                #[cfg(feature = "unstable-msc3768")]
-                notify_in_app,
-            } => Self::MentionsAndKeywordsOnly {
-                #[cfg(feature = "unstable-msc3768")]
-                notify_in_app,
-            },
+            RoomNotificationMode::MentionsAndKeywordsOnly => Self::MentionsAndKeywordsOnly,
+            #[cfg(feature = "unstable-msc3768")]
+            RoomNotificationMode::MentionsAndKeywordsOnlyTheRestInApp => {
+                Self::MentionsAndKeywordsOnlyTheRestInApp
+            }
             RoomNotificationMode::Mute => Self::Mute,
         }
     }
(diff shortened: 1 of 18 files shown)


crates/matrix-sdk-base/src/notification_settings.rs
```diff
@@ -19,10 +19,14 @@ use serde::{Deserialize, Serialize};
...
-    /// Receive notifications for all messages.
+    /// Receive remote and in-app notifications for all messages.
     AllMessages,
-    /// Receive notifications for mentions and keywords only.
+    /// Receive remote and in-app notifications for mentions and keywords only.
     MentionsAndKeywordsOnly,
+    /// Receive remote and in-app notifications for mentions and keywords and
+    /// in-app notifications only for other room messages.
+    #[cfg(feature = "unstable-msc3768")]
+    MentionsAndKeywordsOnlyTheRestInApp,
```
bnjbvr: The name and my understanding of the feature suggest we could:

- make `MentionsAndKeywordsOnly` a (embedded) struct variant now
- with a boolean field in it `notify_all_in_apps` or something like this

I'm not quite sure what the name entails, to be honest, even after reading the MSC. Does it mean, in this mode, that I would receive a notification for every single *message* in the app? (is that based on the event type? or does it mean every push rule now applies within the app?) It might be worth elaborating in the code comment here.

crates/matrix-sdk/src/notification_settings/command.rs
```diff
@@ -80,3 +72,28 @@ impl Command {
...
+    None,
+}
+
+impl Notify {
+    fn get_actions(&self) -> Vec<Action> {
+        match self {
+            Self::All => vec![Action::Notify, Action::SetTweak(Tweak::Sound("default".into()))],
+            #[cfg(feature = "unstable-msc3768")]
+            Self::InAppOnly => vec![Action::NotifyInApp],
+            Self::None => vec![],
```
bnjbvr: nit: let's avoid the macro expansion:

```suggestion
            Self::None => Vec::new()
```

crates/matrix-sdk/src/notification_settings/rule_commands.rs
```diff
@@ -235,7 +238,7 @@ mod tests {
         assert_matches!(&rule_commands.commands[0],
             Command::SetRoomPushRule { room_id: command_room_id, notify } => {
                 assert_eq!(command_room_id, &room_id);
-                assert!(notify);
+                assert_eq!(&Notify::All, notify);
```
bnjbvr: nit: here and below, we can make this slightly more readable (usually RHS is the expectation, otherwise this reads as a ["yoda condition"](https://en.wikipedia.org/wiki/Yoda_conditions)):

```suggestion
                assert_eq!(*notify, Notify::All);
```

crates/matrix-sdk/src/notification_settings/rules.rs
```diff
@@ -85,8 +86,16 @@ impl Rules {
...
-            // if this rule contains a `Notify` action
             if rule.triggers_notification() {
+                cfg_if! {
+                    if #[cfg(feature = "unstable-msc3768")] {
+                        if !rule.triggers_remote_notification() {
+                            // This rule contains a `NotifyInApp` action.
+                            return Some(RoomNotificationMode::MentionsAndKeywordsOnlyTheRestInApp);
+                        }
+                    }
+                }
```
bnjbvr: nit: maybe we don't need a `cfg_if` here? suggestion modulo rustfmt. I think the outer block is needed because one can't put cfg guards on statements, so we need a fake expression statement to make it work. But this might be worth checking without the outer block, in case it works too.

```suggestion
                #[cfg(feature = "unstable-msc3768")] 
                {
                        if !rule.triggers_remote_notification() {
                            // This rule contains a `NotifyInApp` action.
                            return Some(RoomNotificationMode::MentionsAndKeywordsOnlyTheRestInApp);
                        }
                }
```

Also, maybe can we use a custom helper? Here we're doing `.triggers_notifications()` which iterates over all the rules, then `.triggers_remote_notification()` which iterates again. We could iterate once, and figure out on the go if it would trigger a remote or non-remote notification, instead?

crates/matrix-sdk/src/notification_settings/rules.rs
```diff
@@ -112,18 +121,24 @@ impl Rules {
...
+        if let Some(rule) = self.ruleset.get(RuleKind::Underride, rule_id).filter(|r| r.enabled()) {
+            if rule.triggers_notification() {
+                cfg_if! {
+                    if #[cfg(feature = "unstable-msc3768")] {
+                        if !rule.triggers_remote_notification() {
+                            // This rule contains a `NotifyInApp` action.
+                            return RoomNotificationMode::MentionsAndKeywordsOnlyTheRestInApp;
+                        }
+                    }
+                }
```
bnjbvr: I think the two comments above apply here too.

crates/matrix-sdk/src/notification_settings/command.rs
```diff
@@ -80,3 +72,28 @@ impl Command {
         }
     }
 }
+
+/// Enum describing if and how to deliver a notification.
+#[derive(Clone, Debug, Eq, PartialEq)]
```
bnjbvr: Can you make it `Copy` too?

crates/matrix-sdk-ui/src/notification_client.rs
```diff
@@ -771,6 +772,17 @@ impl NotificationClient {
...
 
+fn should_action_notify_remote(action: &Action) -> bool {
+    cfg_if! {
+        if #[cfg(feature = "unstable-msc3768")] {
+            action.should_notify_remote()
+        } else {
+            // Before MSC3768 only combined remote/local notifications existed
+            action.should_notify()
+        }
+    }
```
bnjbvr, line 783: it might have been slightly nicer for all Ruma's consumers, that the Ruma impl of `should_notify_remote()` be available all the time, and the current `should_notify_remote` body if the feature's enabled, or only redirecting to `should_notify` otherwise :)
Nothing to act upon here, though.

bindings/matrix-sdk-ffi/src/notification_settings.rs
```diff
@@ -322,8 +322,13 @@ impl TryFrom<Tweak> for SdkTweak {
...
 /// Enum representing the push notification actions for a rule.
 pub enum Action {
-    /// Causes matching events to generate a notification.
+    /// Causes matching events to generate a notification (both in-app and
+    /// remote / push).
     Notify,
+    /// Causes matching events to generate an in-app notification but no remote
+    /// / push notification.
+    #[cfg(feature = "unstable-msc3768")]
+    NotifyInApp,
```
bnjbvr: Does an embedder need to know that the notification was generated in app? Or is it required to allow creating custom rules that could use this variant?

bindings/matrix-sdk-ffi/src/notification_settings.rs
```diff
@@ -358,10 +367,14 @@ impl TryFrom<Action> for SdkAction {
...
-    /// Receive notifications for all messages.
+    /// Receive remote and in-app notifications for all messages.
     AllMessages,
-    /// Receive notifications for mentions and keywords only.
+    /// Receive remote and in-app notifications for mentions and keywords only.
     MentionsAndKeywordsOnly,
+    /// Receive remote and in-app notifications for mentions and keywords and
+    /// in-app notifications only for other room messages.
+    #[cfg(feature = "unstable-msc3768")]
+    MentionsAndKeywordsOnlyTheRestInApp,
```
bnjbvr: Same remark as the one below for the shape of this enum

bnjbvr (changes requested):
Interesting MSC. How can it be used, right now, by an embedder? Is everything already wired up, so that timeline items would be marked as notifying the user if this rule is set?

I would also like to see some tests for the new expected behaviors, please, and at both the base, SDK, and UI layers, to at least smoke-test the behavior is correctly implemented.

crates/matrix-sdk-base/src/notification_settings.rs
Johennes: So `MentionsAndKeywordsOnly`, adds a room-specific rule to mute the room. The default override rules for mentions and keywords make these still cause notifications though.

`MentionsAndKeywordsOnlyTheRestInApp` does the same only that instead of muting the room, it sets it to notify in-app only. So you'll still get all the unread counts, badges and notifications inside the app. You'll only get _push_ notifications for mentions and keywords, however.

crates/matrix-sdk/src/notification_settings/rule_commands.rs
Johennes: Fixed both with ef96531cfc8f3cc4d70f94bee49c9acba114d703.

crates/matrix-sdk/src/notification_settings/rules.rs
Johennes: Oh, you're right about `cfg_if`. I have eliminated it with cd21783c92e646f2dd7c9e37880fbec580b8f7c7.

As for the second comment, it's actually not iterating over the rules but only over the actions on each rule. That array should normally have at most two items. So I think doing that twice is ok-ish?

crates/matrix-sdk/src/notification_settings/rules.rs
Johennes: See https://github.com/matrix-org/matrix-rust-sdk/pull/5441#discussion_r2257103423.

crates/matrix-sdk/src/notification_settings/command.rs
Johennes: Fixed in 6e3b7f6da03ad811f10e35212b2abcb10feafa14.

crates/matrix-sdk-ui/src/notification_client.rs
Johennes: Yeeeeeah, I can see that now. 🤦‍♂️

bindings/matrix-sdk-ffi/src/notification_settings.rs
Johennes: It would be required for creating custom rules, yes.

crates/matrix-sdk/src/notification_settings/command.rs
Johennes: Sorry, I realized too late that accepting this suggestion won't produce a fixup commit and then had to patch this up again with 3b158ab3f39258d4bcd3f465a3fd861304401852 because a comma was missing. Since the diff is quite small, I hope this is ok. I'll make sure to squash everything into one commit in case this gets approved.

crates/matrix-sdk-base/src/notification_settings.rs
Johennes: Have addressed this with deef9610d75d0d1b88423a800a1cabb1f74f29e2. It's possibly getting a little unwieldy with the struct variant and the `cfg` annotations now though.

bindings/matrix-sdk-ffi/src/notification_settings.rs
Johennes: See https://github.com/matrix-org/matrix-rust-sdk/pull/5441#discussion_r2257291842.

Johennes:
For the tests, I'd like to integrate something into `matrix-sdk-ui/tests/integration/notification_client.rs` but I'm struggling a bit to see how to inject custom push rules in the client. It looks like all the current tests just run with the default push rules?

bindings/matrix-sdk-ffi/src/notification_settings.rs
```diff
@@ -322,8 +322,13 @@ impl TryFrom<Tweak> for SdkTweak {
 #[derive(Clone, uniffi::Enum)]
 /// Enum representing the push notification actions for a rule.
 pub enum Action {
-    /// Causes matching events to generate a notification.
+    /// Causes matching events to generate a notification (both in-app and
+    /// remote / push).
     Notify,
+    /// Causes matching events to generate an in-app notification but no remote
+    /// / push notification.
```
bnjbvr: uh, rustfmt did wrap the line at a confusing place xD

```suggestion
    /// (push) notification.
```

crates/matrix-sdk-base/src/notification_settings.rs
```diff
@@ -19,10 +19,15 @@ use serde::{Deserialize, Serialize};
...
     AllMessages,
-    /// Receive notifications for mentions and keywords only.
-    MentionsAndKeywordsOnly,
+    /// Receive remote and in-app notifications for mentions and keywords only.
+    MentionsAndKeywordsOnly {
+        /// If true, receive in-app-only notifications (no pushes) for other
+        /// room messages. Otherwise, mute all other room messages.
+        #[cfg(feature = "unstable-msc3768")]
+        notify_in_app: bool,
+    },
```
bnjbvr: Sorry to go back and forth with this, but I've realized that since this `enum` is `Deserialize`, this would need a data migration, at the risk of making `RoomInfo` deserialization fail (thus loading of initial rooms fail at start).

But thanks to the nicely split commits (thanks!), I think you can revert deef9610d75d0d1b88423a800a1cabb1f74f29e2 and we should be back to the previous place. Then we'd rename `MentionsAndKeywordsOnlyTheRestInApp` to something like `PushMentionsAndKeywordsOnly`, and we'll be good, for the time being, in terms of client-side implementation.

bnjbvr (changes requested):
Sorry, noticed a thing that makes it not ideal to mix the bool within the field. I should've noticed the first time :pensive: So let's get back to the new variant instead (with a suggested new name). And yeah let's add some notification client tests for the new notify-in-app mode. Likely we'll need to sync/fetch account data to tweak the default push rules for the user.

crates/matrix-sdk-base/src/notification_settings.rs
Johennes: No worries. Was an easy revert. Fixed with b23bc278a701c752e9a71ea3dbee7885f9705cc0 and e6bd99f296e7ad03b0b1b307223c2c2bed0f917f.

Johennes:
Tests are still missing. I will need to figure out how to extend the existing ones.

Johennes:
Moving back to draft to re-evaluate alternatives to [MSC3768](https://github.com/matrix-org/matrix-spec-proposals/pull/3768).

Issue #5440, opened 2025-07-24
The timeline doesn't consider unthreaded read receipts on thread aware instances


stefanceriu:
As reported in https://github.com/element-hq/element-x-ios/issues/4347

On thread aware timeline instances (i.e. timeline foci of `Live { hide_threaded_events: true }` or `Thread { root_event_id`) read receipts sent from unthreaded clients (i.e. missing a `thread_id`) are not considered when computing the user's latest receipts, resulting in incorrect information being reported. 

Even though the remote party has read all the messages (on a flat timeline), the current user will remain stuck on their previously sent one, as they were the latest having a `thread_id`.

Issue #4113, opened 2024-10-10
Event cache: handle read receipts and unread counts


bnjbvr:
Now that we have some storage backend for the event cache, we can move the handling of read receipts in there.

Rough list of implementation steps:

- [ ] Move support for handling read receipts
  - [ ] Identify what the timeline needs
- [x] Move support for unread count too
  - [ ] Fuse the "room info" subscriber with the unread count subscriber into a single one (same as in #4112) 
  - [x] Avoid recomputing read receipts on every linked chunk update
    - Edit by @Hywan: It has moved in `matrix_sdk::event_cache`. It's updated via sync, which solves the problem.
  - [ ] Avoid losing information after shrinking a linked chunk
  - [x] Use `compute_unread_counts` for threads?
- [ ] Add function to "mark room as read" using read receipts

Part of #3058.

bnjbvr:
Copying some notes from #5723:

The event cache would have the responsibility to track read receipts on events:

- tracking means that the read receipt would move, from one event to another, after we receive new read receipt events via sync
- the event cache would propagate dedicated pings for this kind of tracked receipt updates, so the timeline can use that to move read receipts on its items
- it would offer new APIs to load the tracked receipt on a given event (loading all the receipts of an event is already possible via the state store), so the timeline can still load the initial tracked receipts on each event

manuroe:
Discussed today with the team. We have most of what is described here for threads. 2 years later we can close this meta issue.

PR #5442, opened 2025-07-24
feat(ui): consider unthreaded read receipts when computing the user's latest on thread aware timeline instances

stefanceriu:
This change modifies the existing `latest_user_read_receipt` method to retrieve and use unthreaded read receipts and chosing the most recent one.

We do this in order improve compatibility with non-threaded clients and correctly show what the other party has read on flattened out timelines.

Fixes https://github.com/matrix-org/matrix-rust-sdk/issues/5440

### DIFF
--- a/crates/matrix-sdk-ui/src/timeline/controller/read_receipts.rs
+++ b/crates/matrix-sdk-ui/src/timeline/controller/read_receipts.rs
@@ -721,20 +721,23 @@ impl<P: RoomDataProvider> TimelineState<P> {
             )
             .await;
 
-        // If the current client is thread aware, we maintain maximal
-        // compatibility with other clients using either unthreaded or
-        // main-thread read receipts by allowing both here and taking the most
-        // recent one.
-        let unthreaded_public_read_receipt = self
-            .meta
-            .user_receipt(
-                user_id,
-                ReceiptType::Read,
-                ReceiptThread::Unthreaded,
-                room_data_provider,
-                all_remote_events,
-            )
-            .await;
+        // If the current client is thread aware and the own receipt thread is main,
+        // we maintain maximal compatibility with other clients using either unthreaded
+        // or main-thread read receipts by allowing both here and taking the
+        // most recent one.
+        let unthreaded_public_read_receipt = if receipt_thread == ReceiptThread::Main {
+            self.meta
+                .user_receipt(
+                    user_id,
+                    ReceiptType::Read,
+                    ReceiptThread::Unthreaded,
+                    room_data_provider,
+                    all_remote_events,
+                )
+                .await
+        } else {
+            None
+        };
 
         let public_read_receipt = match self.meta.compare_optional_receipts(
             threaded_public_read_receipt.as_ref(),
@@ -757,20 +760,23 @@ impl<P: RoomDataProvider> TimelineState<P> {
             )
             .await;
 
-        // If the current client is thread aware, we maintain maximal
-        // compatibility with other clients using either unthreaded or
-        // main-thread read receipts by allowing both here and taking the most
-        // recent one.
-        let unthreaded_private_read_receipt = self
-            .meta
-            .user_receipt(
-                user_id,
-                ReceiptType::ReadPrivate,
-                ReceiptThread::Unthreaded,
-                room_data_provider,
-                all_remote_events,
-            )
-            .await;
+        // If the current client is thread aware and the own receipt thread is main,
+        // we maintain maximal compatibility with other clients using either unthreaded
+        //  or main-thread read receipts by allowing both here and taking the
+        // most recent one.
+        let unthreaded_private_read_receipt = if receipt_thread == ReceiptThread::Main {
+            self.meta
+                .user_receipt(
+                    user_id,
+                    ReceiptType::ReadPrivate,
+                    ReceiptThread::Unthreaded,
+                    room_data_provider,
+                    all_remote_events,
+                )
+                .await
+        } else {
+            None
+        };
 
         let private_read_receipt = match self.meta.compare_optional_receipts(
             threaded_private_read_receipt.as_ref(),
--- a/crates/matrix-sdk-ui/src/timeline/controller/read_receipts.rs
+++ b/crates/matrix-sdk-ui/src/timeline/controller/read_receipts.rs
@@ -515,8 +515,9 @@ impl<P: RoomDataProvider> TimelineStateTransaction<'_, P> {
                 }
 
                 for (user_id, receipt) in receipts {
-                    if own_receipt_thread == ReceiptThread::Unthreaded {
-                        // If the own receipt thread is unthreaded, we maintain maximal
+                    if matches!(own_receipt_thread, ReceiptThread::Unthreaded | ReceiptThread::Main)
+                    {
+                        // If the own receipt thread is unthreaded or main, we maintain maximal
                         // compatibility with clients using either unthreaded or main-thread read
                         // receipts by allowing both here.
                         if !matches!(
--- a/crates/matrix-sdk-ui/src/timeline/controller/read_receipts.rs
+++ b/crates/matrix-sdk-ui/src/timeline/controller/read_receipts.rs
@@ -710,7 +710,7 @@ impl<P: RoomDataProvider> TimelineState<P> {
     ) -> Option<(OwnedEventId, Receipt)> {
         let all_remote_events = self.items.all_remote_events();
 
-        let public_read_receipt = self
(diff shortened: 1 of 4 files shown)


crates/matrix-sdk-ui/src/timeline/tests/read_receipts.rs
```diff
@@ -560,6 +560,58 @@ async fn test_initial_public_main_thread_receipt() {
     assert_eq!(receipt_event_id, event_id);
 }
 
+#[async_test]
+async fn test_initial_public_unthreaded_receipt_main_threaded_timeline() {
+    let event_id = owned_event_id!("$event_with_receipt");
+
+    // Add initial private receipt on the main thread.
```
bnjbvr: Wrong copy/pasto? this is an unthreaded receipt, which is different from a receipt on the main thread. Also it's not the "private" variant that's used here.

Can you add at the top of this test a small comment explaining what we're expecting here?

crates/matrix-sdk-ui/src/timeline/tests/read_receipts.rs
```diff
@@ -560,6 +560,58 @@ async fn test_initial_public_main_thread_receipt() {
...
+
+    let (receipt_event_id, _) = timeline.controller.latest_user_read_receipt(*ALICE).await.unwrap();
+    assert_eq!(receipt_event_id, event_id);
+}
+
+#[async_test]
+async fn test_initial_public_unthreaded_receipt_other_threaded_timeline() {
+    let event_id = owned_event_id!("$event_with_receipt");
+
+    // Add initial private receipt on the main thread.
```
bnjbvr: Ditto, comment is borked.

+ add also a comment explaining what the test does. At first sight, I thought it was a duplicate of the first one…

crates/matrix-sdk-ui/src/timeline/tests/read_receipts.rs
```diff
@@ -610,6 +662,58 @@ async fn test_initial_private_main_thread_receipt() {
     assert_eq!(receipt_event_id, event_id);
 }
 
+#[async_test]
+async fn test_initial_private_unthreaded_receipt_main_threaded_timeline() {
+    let event_id = owned_event_id!("$event_with_receipt");
+
+    // Add initial private receipt on the main thread.
```
bnjbvr: nit: unthreaded, not main thread

and test comment please :)

crates/matrix-sdk-ui/src/timeline/tests/read_receipts.rs
```diff
@@ -610,6 +662,58 @@ async fn test_initial_private_main_thread_receipt() {
...
+
+    let (receipt_event_id, _) = timeline.controller.latest_user_read_receipt(*ALICE).await.unwrap();
+    assert_eq!(receipt_event_id, event_id);
+}
+
+#[async_test]
+async fn test_initial_private_unthreaded_receipt_other_threaded_timeline() {
+    let event_id = owned_event_id!("$event_with_receipt");
+
+    // Add initial private receipt on the main thread.
```
bnjbvr: ditto for both

bnjbvr:
Stealing review.

crates/matrix-sdk-ui/src/timeline/controller/read_receipts.rs
```diff
@@ -721,17 +721,67 @@ impl<P: RoomDataProvider> TimelineState<P> {
...
+        // If the current client is thread aware, we maintain maximal
+        // compatibility with other clients using either unthreaded or
+        // main-thread read receipts by allowing both here and taking the most
+        // recent one.
+        let unthreaded_public_read_receipt = self
+            .meta
+            .user_receipt(
+                user_id,
+                ReceiptType::Read,
+                ReceiptThread::Unthreaded,
```
bnjbvr: I think this is wrong, because the bound `receipt_thread` variable can be `Unthreaded` as well, in which case we're computing the same thing twice. Also, a thread-focused timeline should not take into account the unthreaded receipt at all, as it's not even certain that the event pointed to by the unthreaded receipt belongs to the thread or not. (So there are more chances it can't be ordered relative to the correct thread receipt at all, since the unthreaded receipt more likely points to another event of another thread.)

crates/matrix-sdk-ui/src/timeline/controller/read_receipts.rs
```diff
@@ -721,17 +721,67 @@ impl<P: RoomDataProvider> TimelineState<P> {
...
+        // If the current client is thread aware, we maintain maximal
+        // compatibility with other clients using either unthreaded or
+        // main-thread read receipts by allowing both here and taking the most
+        // recent one.
+        let unthreaded_private_read_receipt = self
+            .meta
+            .user_receipt(
+                user_id,
+                ReceiptType::ReadPrivate,
+                ReceiptThread::Unthreaded,
```
bnjbvr: Same comment here.

bnjbvr (changes requested):
Thanks for trying it out, but I don't think this is the right fix, for multiple reasons:

- on the one hand, read receipts attached to items aren't computed with this method, so this likely wouldn't fix the problem observed in the linked issue. The state of read receipts in the timeline is the messiest part of the timeline's code, so that's not on you.
- on the other hand, a threaded timeline should likely not take into account the unthreaded receipt at all. I agree that a main thread timeline should use unthreaded|main, though, like a non-threaded timeline uses both.

crates/matrix-sdk-ui/src/timeline/controller/read_receipts.rs
andybalaam: If the meaning of an unthreaded receipt is "I have read everything before this" then I argue an unthreaded receipt should be considered when deciding the unread state of a thread. Otherwise, if I switch from an unthreaded client to a threaded one, all threads from all time will be unread.

On the other hand, in practice, if you flip back to an unthreaded client, read the latest messages and then go back to a threaded one, all your threads will be read, so this might need us to be more pragmatic and follow what you are suggesting Benji - I don't know.

bnjbvr: Unfortunately, the read receipts in the timeline are computed locally, using only the current timeline's events. To know if an event is read, we compare the position of the event timeline item pointed to by the receipt, and the positions of event timeline items. It's more likely that the unthreaded receipt points to an event that's outside a threaded timeline, so it won't be found by the timeline code. As a result, it's not possible, in the current form of the code, to figure that a thread is fully read or not read, based on the threaded timeline. All the information is around, though, so we *could* compute this information (on a user basis), with some extra effort.

stefanceriu:
Thanks for comments, let's try this again:
* 3d9d5c6c60306bcd90e3e4b13297ded3ae1576d7 uses untreaded read receipts within `latest_user_read_receipt` **only** for `Main` timelines
* 227f68c7f764edebddfbf2908aa25261518e8f8e then goes and fixes the real issue reported initially (tested against EX)
* and finally, on @bnjbvr's suggestion, 2ca4d6c0bdbd00287a5030d35784fb4f551d338b rewrites `latest_user_read_receipt` to make use of the timeline, going through timeline items in reverse order until finding a read receipt, to have one single source of truth. This means changing the tests to have them inject timeline items and work around not having access to one's own receipts.

bnjbvr (changes requested):
Sorry about the long review time, but I needed a bit of time to think about this.

I think the actual fix is the 4th commit, and is untested; any chances you could add a test for it?

Other than that: I think that, on the one hand, all the changes in the other commits could be isolated in a separate PR. But, on the other hand, it doesn't look like the current implementation really needs a timeline; it could even go inside the event cache (and that would be a good move towards moving [the read receipt code into the event cache](https://github.com/matrix-org/matrix-rust-sdk/issues/4113)). But this is completely outside the scope of this bugfix, sooooo…

1. We could either get rid of all the commits that aren't fix to the actual problem at hand (provided we'd have at least one regression test for the initial problem).
2. Or we could keep those commits, and put them in a different PR. The reason I'm suggesting that, is that the `latest_user_read_receipt()` API is not used in the SDK, so we'd need to clarify what the expectations are for this public API (i.e. how it's supposed to behave).
3. Or, with enough effort :) we could move the `latest_user_read_receipt()` code inside the `EventCache`, likely in a `RoomReadReceipts` API object (that includes a `Arc<RoomEventCacheInner>`).

I'm tempted to suggest (1) or (2) for the time being, and I could look into (3) myself later. How does that sound?

stefanceriu:
This PR grew a bit too much. Going with solution 1) and closing in favor #5527

Issue #5417, opened 2025-07-16
Monitor benchmarks in CI


manuroe:
We have benchmarks in the project but there are not part of our CI process because we are lacking of time reliable runners. We can only run them locally, which is not something we do every time. 
Services like https://codspeed.io/ may be a good solution for us.

Another alternative would be to count the number of required CPU instructions instead of measuring time.

Having such checks in the CI would avoid us performance regressions that happened in https://github.com/matrix-org/matrix-rust-sdk/pull/5407 that was discovered lately in the apps.

PR #5443, opened 2025-07-24
Enable benchmarks on the CI

poljar:
This sadly removes our ability to create flamegraphs locally due to the usage of a fork of criterion.

Codspeed by default generates those for us, so it might not be that bad. But if needs be we might investigate how to get them back locally as well.

Some of our benchmarks seem to panic when run, so I'm only enabling a couple of benchmarks for now.

Closes: https://github.com/matrix-org/matrix-rust-sdk/issues/5417.

### DIFF
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -1,54 +1,37 @@
 name: Benchmarks
 
 on:
+  push:
+    branches:
+      - "main"
+  pull_request:
   workflow_dispatch:
 
 jobs:
   benchmarks:
     name: Run Benchmarks
     runs-on: ubuntu-latest
-    environment: matrix-rust-bot
-    if: github.event_name == 'push'
+    strategy:
+      matrix:
+        benchmark:
+          - crypto_bench
+          - event_cache
 
     steps:
-    - name: Checkout the repo
-      uses: actions/checkout@v4
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683
 
-    - name: Install Rust
-      uses: dtolnay/rust-toolchain@master
-      with:
-        toolchain: nightly-2025-06-27
-        components: rustfmt
+      - name: Setup rust toolchain, cache and cargo-codspeed binary
+        uses: moonrepo/setup-rust@ede6de059f8046a5e236c94046823e2af11ca670
+        with:
+          channel: stable
+          cache-target: release
+          bins: cargo-codspeed
 
-    - name: Run Benchmarks
-      run: cargo bench | tee benchmark-output.txt
+      - name: Build the benchmark target(s)
+        run: cargo codspeed build ${{ matrix.benchmarks }}
 
-    - name: Check benchmark result for PR
-      if: github.event_name == 'pull_request'
-      uses: benchmark-action/github-action-benchmark@v1
-      with:
-        name: Rust Benchmark
-        tool: 'cargo'
-        output-file-path: benchmark-output.txt
-        auto-push: false
-        # comment to alert the user this has gone bad
-        github-token: ${{ secrets.MRB_ACCESS_TOKEN }}
-        alert-threshold: '120%'
-        comment-on-alert: true
-        fail-threshold: '150%'
-        fail-on-alert: true
-
-    - name: Store benchmark result
-      if: github.event_name != 'pull_request'
-      uses: benchmark-action/github-action-benchmark@v1
-      with:
-        name: Rust Benchmark
-        tool: 'cargo'
-        output-file-path: benchmark-output.txt
-        github-token: ${{ secrets.GITHUB_TOKEN }}
-        auto-push: true
-        # Show alert with commit comment on detecting possible performance regression
-        alert-threshold: '150%'
-        comment-on-alert: true
-        fail-on-alert: true
-        alert-comment-cc-users: '@gnunicornBen,@jplatte,@poljar'
+      - name: Run the benchmarks
+        uses: CodSpeedHQ/action@0b6e7a3d96c9d2a6057e7bcea6b45aaf2f7ce60b
+        with:
+          run: cargo codspeed run
+          token: ${{ secrets.CODSPEED_TOKEN }}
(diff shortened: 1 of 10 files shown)


benchmarks/benches/timeline.rs
```diff
@@ -118,18 +118,7 @@ pub fn create_timeline_with_initial_events(c: &mut Criterion) {
...
-            100,
-            pprof::criterion::Output::Flamegraph(None),
-        ))
-    }
-
-    #[cfg(not(target_os = "linux"))]
-    {
-        Criterion::default()
-    }
+    Criterion::default()
```
bnjbvr: Nice. Any chance we can remove the `criterion()` function altogether, and remove the `config = criterion()` in the `criterion_group!` macro invoke? (If so, we could do this for every file)

.github/workflows/benchmarks.yml
```diff
@@ -1,54 +1,37 @@
...
+        uses: moonrepo/setup-rust@ede6de059f8046a5e236c94046823e2af11ca670
+        with:
+          channel: stable
+          cache-target: release
+          bins: cargo-codspeed
 
-    - name: Run Benchmarks
-      run: cargo bench | tee benchmark-output.txt
+      - name: Build the benchmark target(s)
+        run: cargo codspeed build ${{ matrix.benchmarks }}
```
bnjbvr: Do we need to build with the special `profiling` profile, so as to have precise function names and so on? Not sure if this is required…

bnjbvr (approved):
Sweet, thanks! We need to investigate the other benchmarks, but yeah let's do that later. Any chance we could have a benchmark report posted on this issue, to get an example? Otherwise it's fine, we'll wait for the next run.

bnjbvr:
Ah, I'm only noticing now that the benchmarks are running, and they've been running for 18 minutes already, and it's still compiling :scream: This might be the optimized profile that has heavyweight LTO kicking in; I did put `lto = false` in the `profiling` profile exactly for this reason. Let's do this as well for the benchmarks here, to not slow down a CI pipeline too much?

(I'd argue it's fine to do so: overall we're interested in the evolution of performance over time, not what's the absolute best performance we could have if we enabled all the optimization compile flags.)

benchmarks/benches/timeline.rs
poljar: I intended to leave this in as it will be easier to bring back local pprof support, but sure we can remove it if you prefer the functions to be gone.

poljar:
> Ah, I'm only noticing now that the benchmarks are running, and they've been running for 18 minutes already, and it's still compiling 😱 This might be the optimized profile that has heavyweight LTO kicking in; I did put `lto = false` in the `profiling` profile exactly for this reason. Let's do this as well for the benchmarks here, to not slow down a CI pipeline too much?
> 
> (I'd argue it's fine to do so: overall we're interested in the evolution of performance over time, not what's the absolute best performance we could have if we enabled all the optimization compile flags.)

I messed up and didn't enable just a single benchmark, instead we have two runs of all benchmarks.

Let us instead introduce a `bench` profile, since the tooling will use the `bench` profile by default.

.github/workflows/benchmarks.yml
```diff
@@ -28,7 +28,7 @@ jobs:
           bins: cargo-codspeed
 
       - name: Build the benchmark target(s)
-        run: cargo codspeed build ${{ matrix.benchmarks }}
+        run: cargo codspeed build -p benchmarks ${{ matrix.benchmark }}
```
bnjbvr: Shouldn't it be `-p bench`, since the profile is named `bench`?

Cargo.toml
```diff
@@ -180,6 +180,10 @@ lto = false
 # Get symbol names for profiling purposes.
 debug = true
 
+[profile.bench]
+inherits = "dbg"
```
bnjbvr: release? :D

.github/workflows/benchmarks.yml
poljar: It's the name of the package, the `bench` profile is used by default so we don't need to specify `--profile`.

bnjbvr:
[ohnoes](https://github.com/matrix-org/matrix-rust-sdk/actions/runs/16497152654/job/46645580095?pr=5443#step:5:89):

>   Skipping: benchmarks/benches/event_cache.rs::event_cache::handle_room_updates::reading::handle_room_updates/memory store[1] (CodSpeed doesn't support custom iterations)

Cargo.toml
poljar: Don't we want to go for the best compile time for this? What would `opt-level = 3` give us?

poljar:
> [ohnoes](https://github.com/matrix-org/matrix-rust-sdk/actions/runs/16497152654/job/46645580095?pr=5443#step:5:89):
> 
> > Skipping: benchmarks/benches/event_cache.rs::event_cache::handle_room_updates::reading::handle_room_updates/memory store[1] (CodSpeed doesn't support custom iterations)

At least it finished in 5 minutes :sweat_smile:

Cargo.toml
bnjbvr: The dbg profile includes debug assertions, and may be super slow to run in general, as it disables inlining ("the mother of all compiler optimizations"). Benchmarking is usually meaningless in debug mode. Even though we don't want LTO as it makes the compile times explode, we still do want something that should be representative of real-world performance, so setting the release opt-level would be more realistic :+1:

poljar: https://github.com/matrix-org/matrix-rust-sdk/pull/5443/commits/c636ec63f44367d2ecea5f7a229d6a9eefe2260b

benchmarks/benches/timeline.rs
poljar: Removed this in https://github.com/matrix-org/matrix-rust-sdk/pull/5443/commits/d40f04e32c47f3df782807bb923ca74d35eb405d.

.github/workflows/benchmarks.yml
poljar: The `bench` profile was added.

PR #5444, opened 2025-07-24
fix: Properly support event types with variable suffixes

zecakeh:
… or add a bound to make it clear that they are not supported by the API.

Thanks to a recent addition in Ruma, we can now identify event types where only the prefix is static with the `IsPrefix` associated type on `StaticEventContent` (`SecretStorageKeyEventContent` is currently the only type where it is `True`).

In the SDK all the APIs that use that bound only support full event types, not prefixes. So we have two commits here:

1. The first one adds a bound `IsPrefix = False` on the `get_*_static` APIs, because they also have sibling APIs that take a non-static event type. We also add the same bound to `EventBuilder` because it's the only supported way currently, and it's not a public API.
2. The second one adds support for event type prefixes to event handlers, since there is no alternative API where we can provide a non-static event type to watch.

### DIFF
--- a/crates/matrix-sdk-base/src/store/mod.rs
+++ b/crates/matrix-sdk-base/src/store/mod.rs
@@ -578,7 +578,9 @@ impl StateChanges {
         state_key: &K,
     ) -> Option<&Raw<SyncStateEvent<C>>>
     where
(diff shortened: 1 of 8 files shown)


poljar (changes requested):
Nice, we should advertise this in the changelog.

Looks good otherwise.

PR #5445, opened 2025-07-24
feat(common): add the `Thread` variant to `OwnedLinkedChunkId` and `LinkedChunkId`

bnjbvr:
This should make it clearer what's the expected behavior for the IDB implementation of the event cache. Also, testing is good, because this was actually incorrectly implemented in the memory store.

I've added it in the IDB implementation as well, because I think it should fail (and can't run the IDB tests locally for some reason). [edit] Yes, it did fail, so I've removed it from this PR; someone can investigate it later.

Part of #5122, so part of #4869 too.

### DIFF
--- a/crates/matrix-sdk-base/CHANGELOG.md
+++ b/crates/matrix-sdk-base/CHANGELOG.md
@@ -20,6 +20,9 @@ All notable changes to this project will be documented in this file.
   ([#5390](https://github.com/matrix-org/matrix-rust-sdk/pull/5390))
 
 ### Refactor
+- [**breaking**] `RelationalLinkedChunk::items` now takes a `RoomId` instead of an
+  `&OwnedLinkedChunkId` parameter.
+  ([#5445](https://github.com/matrix-org/matrix-rust-sdk/pull/5445))
 - [**breaking**] Add an `IsPrefix = False` bound to the
   `get_state_event_static()`, `get_state_event_static_for_key()` and
   `get_state_events_static()`, `get_account_data_event_static()` and


crates/matrix-sdk-common/src/linked_chunk/relational.rs
```diff
@@ -376,20 +376,22 @@ where
...
     ///
     /// This will include out-of-band items.
-    pub fn items(
-        &self,
-        target: &OwnedLinkedChunkId,
-    ) -> impl Iterator<Item = (&Item, Option<Position>)> {
+    pub fn items<'a>(
+        &'a self,
+        room_id: &'a RoomId,
+    ) -> impl Iterator<Item = (&'a Item, Option<Position>)> {
```
poljar, line 389: That's a breaking change, right?

Changelog entry please?

bnjbvr: I don't expect anyone using this data structure outside our own crates, the same way mostly no one might have good reasons to use `matrix-sdk-base`, so adding a changelog entry for this sounds overkill to me.

poljar: While the base crate itself is indeed less important for end users and its changelogs might be less interesting, it is part of our release. I don't think it makes sense to add special exceptions on which changes to the public API should make it to the changelog or not. 

We're already pretty bad at remembering to add changelog entries, myself included. I'd much rather have us build a healthy habit where changelog updates are expected for every change. Adding too many exceptions risks reinforcing the idea that it's okay to skip them, which I think we should avoid.

One of these days we should set up [cargo-semver-checks](https://github.com/obi1kenobi/cargo-semver-checks) to remind us about breaking changes. Ruma did so recently as well.

PR #5446, opened 2025-07-24
Allow Element Call to send call notifications

robintown:
For clients which integrate Element Call: Currently matrix-rust-sdk is responsible for sending the call notification event when joining MatrixRTC sessions, but this is planned to be changed soon. As of the upcoming Element Call 0.14.0 release, it will request the capability to send call notifications itself, and we should auto-approve this capability.


stefanceriu (approved):
Makes sense to me 👍

Will there be a separate PR removing the previous implementation (`send_call_notification_if_needed`)?

robintown:
> Will there be a separate PR removing the previous implementation (`send_call_notification_if_needed`)?

Yep, working on that now

PR #5449, opened 2025-07-28
refactor(ffi): expose room_version in the FFI RoomInfo

jmartinesp:
## Changes:

- Exposes the room version in the FFI RoomInfo.
- Exposes the `privileged_creators_role` boolean in the FFI RoomInfo, to understand how the retrieved creators ids should be treated.

This is needed on the clients so we can understand whether `RoomInfo.creators` are the ids of the room creators (with no special privileges) or they're users with infinite power level and should be treated differently.

Signed-off-by:


poljar (approved):
Please just add changelog entries for these.

PR #5451, opened 2025-07-28
ci: Enable more benchmarks

poljar:
Let's see if we can get some more benchmarks running.

### DIFF
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -16,6 +16,9 @@ jobs:
         benchmark:
           - crypto_bench
           - event_cache
+          - linked_chunk
+          - store_bench
+          - timeline
 
     steps:
       - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683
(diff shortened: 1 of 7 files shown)


stefanceriu (approved):
🐎

bnjbvr:
Super nice, thanks for enabling those. I wonder if we should take care of renaming all the benchmarks before merging this, maybe? Right now they appear as `metadata/sqlite store[10]` or (worse case) `none[10]`, but they include nothing about the benchmark itself. I suspect that's because, in the past, we'd run it with the `cargo bench` command by including the bench's name (= the file name), so that would give us the main context this way, but now we don't have this main context anymore. So what I'm suggesting here, is that we'd add this context in the bench string description, so that it appears in the codspeed reports. Doing this before merging this PR would avoid having a few benchmarks "disappear". What do you think?

poljar:
> Doing this before merging this PR would avoid having a few benchmarks "disappear". What do you think?

Yeah, that's something I wanted to do in the first PR as well, then I forgot a bit about this and went the "fuck it, we'll do it live" route.

PR #5453, opened 2025-07-28
feat(send_queue): report progress for media uploads

Johennes:
This contains the send queue changes broken out from #5008.


andybalaam:
Assigning to @bnjbvr since it's part of a review you were already doing I think. Feel free to assign something to me in return :-)

PR #5454, opened 2025-07-28
feat(timeline): report progress for media uploads

Johennes:
This is meant to only contain the timeline changes broken out from #5008 (together with #5453 which this depends on).

### DIFF
--- a/bindings/matrix-sdk-ffi/src/timeline/mod.rs
+++ b/bindings/matrix-sdk-ffi/src/timeline/mod.rs
@@ -30,7 +30,8 @@ use matrix_sdk_common::{
     stream::StreamExt,
 };
 use matrix_sdk_ui::timeline::{
-    self, AttachmentConfig, AttachmentSource, EventItemOrigin, Profile, TimelineDetails,
+    self, AttachmentConfig, AttachmentSource, EventItemOrigin,
+    MediaUploadProgress as SdkMediaUploadProgress, Profile, TimelineDetails,
     TimelineUniqueId as SdkTimelineUniqueId,
 };
 use mime::Mime;
@@ -235,6 +236,50 @@ impl From<UploadSource> for AttachmentSource {
         match value {
             UploadSource::File { filename } => Self::File(filename.into()),
             UploadSource::Data { bytes, filename } => Self::Data { bytes, filename },
+        }
+    }
+}
+
+/// This type represents the progress of a media (consisting of a file and
+/// possibly a thumbnail) being uploaded.
+#[derive(Clone, Copy, uniffi::Record)]
+pub struct MediaUploadProgress {
+    /// The index of the media within the transaction. A file and its
+    /// thumbnail share the same index. Will always be 0 for non-gallery
+    /// media uploads.
+    pub index: u64,
+
+    /// The current combined upload progress for both the file and,
+    /// if it exists, its thumbnail.
+    pub progress: AbstractProgress,
+}
+
+impl From<SdkMediaUploadProgress> for MediaUploadProgress {
+    fn from(value: SdkMediaUploadProgress) -> Self {
+        Self { index: value.index, progress: value.progress.into() }
+    }
+}
+
+/// Progress of an operation in abstract units.
+///
+/// Contrary to [`TransmissionProgress`], this allows tracking the progress
+/// of sending or receiving a payload in estimated pseudo units representing a
+/// percentage. This is helpful in cases where the exact progress in bytes isn't
+/// known, for instance, because encryption (which changes the size) happens on
+/// the fly.
+#[derive(Clone, Copy, uniffi::Record)]
+pub struct AbstractProgress {
+    /// How many units were already transferred.
+    pub current: u64,
+    /// How many units there are in total.
+    pub total: u64,
+}
+
+impl From<matrix_sdk::send_queue::AbstractProgress> for AbstractProgress {
+    fn from(value: matrix_sdk::send_queue::AbstractProgress) -> Self {
+        Self {
+            current: value.current.try_into().unwrap_or(u64::MAX),
+            total: value.total.try_into().unwrap_or(u64::MAX),
         }
     }
 }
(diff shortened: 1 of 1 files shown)


Johennes:
I have addressed all comments from https://github.com/matrix-org/matrix-rust-sdk/pull/5008, except for:

> I do realize only now that it's a bit weird that message events also get a progress optional value here, despite this option being never set. This suggests that some variant EventSendState::Uploading that we could use for medias only, maybe?
> [...]
> Based on other comments: we might prefer to have a new variant NotUploadedYet, for medias. The progress field isn't used, for non-media uploads.

This is turning out a little more involved than expected, sadly.

There are places like [here](https://github.com/Johennes/matrix-rust-sdk/blob/johannes/timeline-media-upload-progress/crates/matrix-sdk-ui/src/timeline/event_handler.rs#L733) when a timeline item is added where I would have to peek into the content and figure out if it's a media event or not. There's also the retry handling [here](https://github.com/Johennes/matrix-rust-sdk/blob/johannes/timeline-media-upload-progress/crates/matrix-sdk-ui/src/timeline/controller/mod.rs#L1367) where I only have the transaction ID to figure out if it's a media upload transaction or not. It feels like working in a new enum variant there might introduce quite a bit of complexity.

Alternatively, we could only use `NotUploadedYet` once the upload actually starts and stick to `NotSentYet` otherwise. That might make things even more confusing than right now though.

I currently feel like keeping the progress optional in `NotSentYet` might do the least damage?

bindings/matrix-sdk-ffi/src/timeline/mod.rs
```diff
@@ -241,6 +242,56 @@ impl From<UploadSource> for AttachmentSource {
     }
 }
 
+/// This type represents the "send progress" of a local event timeline item.
+#[derive(Clone, Copy, uniffi::Enum)]
+pub enum EventSendProgress {
+    /// A media is being uploaded.
+    MediaUpload {
```
bnjbvr: Since this isn't serialized: maybe we could make this enum with one variant a plain struct, instead? `MediaUploadProgress` or something like that?

bindings/matrix-sdk-ffi/src/timeline/mod.rs
```diff
@@ -990,7 +1041,10 @@ impl TimelineItem {
 #[derive(Clone, uniffi::Enum)]
 pub enum EventSendState {
     /// The local event has not been sent yet.
-    NotSentYet,
+    NotSentYet {
+        /// The progress of the sending operation, if any is available.
```
bnjbvr: Can we precise it's only available for media uploads, at the moment?

crates/matrix-sdk-ui/src/timeline/controller/mod.rs
```diff
@@ -1072,7 +1072,11 @@ impl<P: RoomDataProvider, D: Decryptor> TimelineController<P, D> {
                 warn!("We looked for a local item, but it transitioned as remote??");
                 return false;
             };
-            prev_local_item.with_send_state(EventSendState::NotSentYet)
+            // If the local echo had an upload progress, retain it.
+            let progress = as_variant!(&prev_local_item.send_state,
+                EventSendState::NotSentYet { progress } => progress.clone())
+            .flatten();
```
bnjbvr, line 1078: Let's let the code breath a bit, and simplify the `as_variant!` to do as little as possible:

```suggestion

            // If the local echo had an upload progress, retain it.
            let progress = as_variant!(&prev_local_item.send_state, EventSendState::NotSentYet)
              .and_then(|progress| progress.cloned());
            
```

crates/matrix-sdk-ui/src/timeline/event_item/local.rs
```diff
@@ -65,7 +68,10 @@ impl LocalEventTimelineItem {
 #[derive(Clone, Debug)]
 pub enum EventSendState {
     /// The local event has not been sent yet.
-    NotSentYet,
+    NotSentYet {
+        /// The progress of the sending operation, if any is available.
```
bnjbvr: Can we precise it's only available for media uploads, please?

crates/matrix-sdk-ui/src/timeline/event_item/local.rs
```diff
@@ -84,3 +90,21 @@ pub enum EventSendState {
         event_id: OwnedEventId,
     },
 }
+
+/// This type represents the "send progress" of a local event timeline item.
+#[derive(Clone, Debug)]
+pub enum EventSendProgress {
+    /// A media (consisting of a file and possibly a thumbnail) is being
+    /// uploaded.
+    MediaUpload {
```
bnjbvr: ditto, since this is an enum with a single variant, and that's not ser/de, let's use a struct instead

bnjbvr (changes requested):
We've discussed this with the Rust SDK core team, and we think this is a nice API. We might experiment with a new `EventSendState::Uploading` state later, but since this is unclear whether it'll be good or not, I won't request it now.

My final comments are mostly cosmetic improvements, so a last round and we should be good to go. Thanks for this PR!

crates/matrix-sdk-ui/src/timeline/controller/mod.rs
Johennes: This seems to only work for tuple structs / tuple variants?

crates/matrix-sdk-ui/src/timeline/event_item/local.rs
Johennes: Fixed with 1bcf040ab904024ed12ac050b2cefcf625cb2c2e

bindings/matrix-sdk-ffi/src/timeline/mod.rs
Johennes: Fixed with 5689fcea1bb2f943f26600f4cfbf04616553290c.

crates/matrix-sdk-ui/src/timeline/event_item/local.rs
Johennes: Fixed with 68f7aea1e63860a11c2157b28e5cc72ac38b587f.

bindings/matrix-sdk-ffi/src/timeline/mod.rs
Johennes: Fixed with ccd4c48eac5adc9f9caf14fd6bad5408a5622bff.

crates/matrix-sdk-ui/src/timeline/controller/mod.rs
bnjbvr: ah crap. Thanks for trying though!

bnjbvr (approved):
Looks good to me, thanks! Mind fixing the merge conflit and auto-squashing the fixup commits please? Then let's merge it \o/

PR #5455, opened 2025-07-28
feat: store the thread subscription status in the database

bnjbvr:
Some implementation notes:

- instead of using kvdata (`StateStoreDataKey` et al.), I've added a new table for the thread subscriptions, so we can later do queries like "where status is subscribed".
- I've made it so that unsubscribing is remembered on the local device, so the next attempt to fetch a thread subscription may return unsubscribed. The MSC should soon evolve in that direction, in such a way that a thread status can be unsubscribed on the server as well, so I'm just being a bit early here.

### DIFF
--- a/crates/matrix-sdk/src/room/mod.rs
+++ b/crates/matrix-sdk/src/room/mod.rs
@@ -34,6 +34,7 @@ use http::StatusCode;
 pub use identity_status_changes::IdentityStatusChanges;
 #[cfg(feature = "e2e-encryption")]
 use matrix_sdk_base::crypto::{IdentityStatusChange, RoomIdentityProvider, UserIdentity};
+pub use matrix_sdk_base::store::ThreadStatus;
 #[cfg(feature = "e2e-encryption")]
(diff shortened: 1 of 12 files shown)


crates/matrix-sdk/src/room/mod.rs
```diff
@@ -3773,42 +3792,57 @@ impl Room {
...
+                    // return something slightly more precise than what the server returned.
+                    let stored_status = self
+                        .client
+                        .state_store()
+                        .load_thread_subscription(self.room_id(), &thread_root)
+                        .await?;
+
+                    if let Some(ThreadStatus::Unsubscribed) = stored_status {
+                        // The thread was unsubscribed from before, so maintain this information.
+                        Ok(Some(ThreadStatus::Unsubscribed))
```
stefanceriu, line 3754: 👌

stefanceriu (approved):
Nice, lgtm 👍

stefanceriu (approved):
still lgtm 👍

PR #5456, opened 2025-07-28
feat: Encrypted state events (MSC3414)

kaylendog:
Implements support for encrypted state events under `m.room.encrypted`, as outlined in [MSC3414](https://github.com/matrix-org/matrix-spec-proposals/blob/travis/msc/encrypted-state/proposals/3414-encrypted-state.md).

- (base) Modifies `state_events::sync::dispatch` to require `E2EE` when end-to-end encryption is enabled, using it to decrypt `m.room.encrypted` state events when received.
- (base) Declares `m.room.encrypted` as required state by default.
- (crypto) Introduces `encrypt_state_event` and `encrypt_state_event_raw` to `OlmMachine` and `OutboundGroupSession`;
- (sdk) Modifies `Room::send_state_event` and `Room::send_state_event_Raw` to return `SendStateEvent` and `SendStateEventRaw` futures respectifely, that each function similarly to their message-like counterparts.
- (integration) Add an integration test for encrypted room renaming.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -68,6 +68,7 @@ ruma = { git = "https://github.com/ruma/ruma", rev = "e73f302e4df7f5f0511fca1aa4
     "compat-lax-room-create-deser",
     "compat-lax-room-topic-deser",
     "unstable-msc3401",
+    "unstable-msc3414",
     "unstable-msc3488",
     "unstable-msc3489",
     "unstable-msc4075",
--- a/crates/matrix-sdk-base/src/response_processors/state_events.rs
+++ b/crates/matrix-sdk-base/src/response_processors/state_events.rs
@@ -26,6 +26,8 @@ use serde::Deserialize;
 use tracing::warn;
 
 use super::Context;
+#[cfg(feature = "e2e-encryption")]
+use super::e2ee;
 use crate::store::BaseStateStore;
 
 /// Collect [`AnySyncStateEvent`].
@@ -89,6 +91,7 @@ pub mod sync {
         ambiguity_cache: &mut AmbiguityCache,
         new_users: &mut U,
         state_store: &BaseStateStore,
+        #[cfg(feature = "e2e-encryption")] e2ee: super::e2ee::E2EE<'_>,
     ) -> StoreResult<()>
     where
         U: NewUsers,
@@ -140,6 +143,80 @@ pub mod sync {
                         // Do not add the event to `context.state_changes.state`.
                         continue;
                     }
+                }
+
+                #[cfg(feature = "e2e-encryption")]
+                AnySyncStateEvent::RoomEncrypted(outer) => {
+                    use matrix_sdk_crypto::RoomEventDecryptionResult;
+                    use tracing::{debug, warn};
+
+                    debug!(event_id = ?outer.event_id(), "Received encrypted state event, attempting decryption...");
+
+                    let Some(olm_machine) = e2ee.olm_machine else {
+                        panic!();
+                    };
+
+                    let decrypted_event = olm_machine
+                        .try_decrypt_room_event(
+                            raw_event.cast_ref_unchecked(),
+                            &room_info.room_id,
+                            e2ee.decryption_settings,
+                        )
+                        .await
+                        .expect("OlmMachine was not started");
+
+                    // Skip state events that failed to decrypt.
+                    let RoomEventDecryptionResult::Decrypted(room_event) = decrypted_event else {
+                        warn!(event_id = ?outer.event_id(), "Failed to decrypt state event");
+                        continue;
+                    };
+
+                    // Unpack event type and state key from outer, or discard if this fails.
+                    let Some((outer_event_type, outer_state_key)) =
+                        outer.state_key().split_once(":")
+                    else {
+                        warn!(
+                            event_id = outer.event_id().as_str(),
+                            state_key = event.state_key(),
+                            "Malformed state key"
+                        );
+                        continue;
+                    };
+
+                    let inner =
+                        match room_event.event.deserialize_as_unchecked::<AnySyncStateEvent>() {
+                            Ok(inner) => inner,
+                            Err(e) => {
+                                warn!("Malformed event body: {e}");
+                                continue;
+                            }
+                        };
+
+                    // Check event types match, discard if not.
+                    let inner_event_type = inner.event_type().to_string();
+                    if outer_event_type != inner_event_type {
+                        warn!(
+                            event_id = outer.event_id().as_str(),
+                            expected = outer_event_type,
+                            found = inner_event_type,
+                            "Mismatched event type"
+                        );
+                        continue;
+                    }
+
+                    // Check state keys match, discard if not.
+                    if outer_state_key != inner.state_key() {
+                        warn!(
+                            event_id = outer.event_id().as_str(),
+                            expected = outer_state_key,
+                            found = inner.state_key(),
+                            "Mismatched state key"
+                        );
+                        continue;
+                    }
+
+                    debug!(event_id = ?outer.event_id(), "Decrypted state event successfully.");
+                    room_info.handle_state_event(&inner);
                 }
 
                 _ => {
--- a/crates/matrix-sdk-base/src/room/encryption.rs
+++ b/crates/matrix-sdk-base/src/room/encryption.rs
@@ -36,6 +36,10 @@ pub enum EncryptionState {
     /// The room is encrypted.
     Encrypted,
 
+    /// The room is encrypted, additionally requiring state events to be
+    /// encrypted.
+    StateEncrypted,
+
     /// The room is not encrypted.
     NotEncrypted,
 
@@ -45,9 +49,16 @@ pub enum EncryptionState {
 }
 
 impl EncryptionState {
-    /// Check whether `EncryptionState` is [`Encrypted`][Self::Encrypted].
+    /// Check whether `EncryptionState` is [`Encrypted`][Self::Encrypted] or
+    /// [`StateEncrypted`][Self::StateEncrypted].
     pub fn is_encrypted(&self) -> bool {
-        matches!(self, Self::Encrypted)
+        matches!(self, Self::Encrypted | Self::StateEncrypted)
+    }
+
+    /// Check whether `EncryptionState` is
+    /// [`StateEncrypted`][Self::StateEncrypted].
+    pub fn is_state_encrypted(&self) -> bool {
+        matches!(self, Self::StateEncrypted)
     }
 
     /// Check whether `EncryptionState` is [`Unknown`][Self::Unknown].
--- a/crates/matrix-sdk-crypto/src/machine/mod.rs
+++ b/crates/matrix-sdk-crypto/src/machine/mod.rs
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 use std::{
+    borrow::Borrow,
     collections::{BTreeMap, HashMap, HashSet},
     sync::Arc,
     time::Duration,
@@ -44,8 +45,9 @@ use ruma::{
     },
     assign,
     events::{
+        room::encrypted::unstable_state::StateRoomEncryptedEventContent,
         secret::request::SecretName, AnyMessageLikeEvent, AnyMessageLikeEventContent,
-        AnyToDeviceEvent, MessageLikeEventContent,
+        AnyStateEventContent, AnyToDeviceEvent, MessageLikeEventContent, StateEventContent,
     },
     serde::{JsonObject, Raw},
     DeviceId, MilliSecondsSinceUnixEpoch, OneTimeKeyAlgorithm, OwnedDeviceId, OwnedDeviceKeyId,
@@ -1098,6 +1100,62 @@ impl OlmMachine {
         content: &Raw<AnyMessageLikeEventContent>,
     ) -> MegolmResult<Raw<RoomEncryptedEventContent>> {
         self.inner.group_session_manager.encrypt(room_id, event_type, content).await
+    }
+
+    /// Encrypt a state event for the given room.
+    ///
+    /// # Arguments
+    ///
+    /// * `room_id` - The id of the room for which the event should be
+    ///   encrypted.
+    ///
+    /// * `content` - The plaintext content of the event that should be
+    ///   encrypted.
+    ///
+    /// * `state_key` - The associated state key of the event.
+    pub async fn encrypt_state_event<C, K>(
+        &self,
+        room_id: &RoomId,
+        content: C,
+        state_key: K,
+    ) -> MegolmResult<Raw<StateRoomEncryptedEventContent>>
+    where
+        C: StateEventContent,
+        C::StateKey: Borrow<K>,
+        K: AsRef<str>,
+    {
+        let event_type = content.event_type().to_string();
+        let content = Raw::new(&content)?.cast_unchecked();
+        self.encrypt_state_event_raw(room_id, &event_type, state_key.as_ref(), &content).await
+    }
+
+    /// Encrypt a state event for the given state event using its raw JSON
+    /// content and state key.
+    ///
+    /// This method is equivalent to [`OlmMachine::encrypt_state_event`]
+    /// method but operates on an arbitrary JSON value instead of strongly-typed
+    /// event content struct.
+    ///
+    /// # Arguments
+    ///
+    /// * `room_id` - The id of the room for which the message should be
+    ///   encrypted.
+    ///
+    /// * `content` - The plaintext content of the event that should be
+    ///   encrypted as a raw JSON value.
+    ///
+    /// * `state_key` - The associated state key of the event.
+    pub async fn encrypt_state_event_raw(
+        &self,
+        room_id: &RoomId,
+        event_type: &str,
+        state_key: &str,
+        content: &Raw<AnyStateEventContent>,
+    ) -> MegolmResult<Raw<StateRoomEncryptedEventContent>> {
+        self.inner
+            .group_session_manager
+            .encrypt_state(room_id, event_type, state_key, content)
+            .await
     }
 
     /// Forces the currently active room key, which is used to encrypt messages,
--- a/crates/matrix-sdk-crypto/src/olm/group_sessions/outbound.rs
+++ b/crates/matrix-sdk-crypto/src/olm/group_sessions/outbound.rs
@@ -491,35 +491,12 @@ impl OutboundGroupSession {
         }
 
         let payload = Payload { event_type, content, room_id: &self.room_id };
-        let payload_json =
-            serde_json::to_string(&payload).expect("payload serialization never fails");
 
         let relates_to = content
             .get_field::<serde_json::Value>("m.relates_to")
             .expect("serde_json::Value deserialization with valid JSON input never fails");
 
-        let ciphertext = self.encrypt_helper(payload_json).await;
-        let scheme: RoomEventEncryptionScheme = match self.settings.algorithm {
-            EventEncryptionAlgorithm::MegolmV1AesSha2 => MegolmV1AesSha2Content {
-                ciphertext,
-                sender_key: Some(self.account_identity_keys.curve25519),
-                session_id: self.session_id().to_owned(),
-                device_id: Some(self.device_id.clone()),
-            }
-            .into(),
-            #[cfg(feature = "experimental-algorithms")]
-            EventEncryptionAlgorithm::MegolmV2AesSha2 => {
-                MegolmV2AesSha2Content { ciphertext, session_id: self.session_id().to_owned() }
-                    .into()
-            }
-            _ => unreachable!(
-                "An outbound group session is always using one of the supported algorithms"
-            ),
-        };
-
-        let content = RoomEncryptedEventContent { scheme, relates_to, other: Default::default() };
-
-        Raw::new(&content).expect("m.room.encrypted event content can always be serialized")
+        self.encrypt_inner(&payload, relates_to).await
     }
 
     /// Encrypt a room state event for the given room.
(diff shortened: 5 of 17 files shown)


richvdh:
This is part of https://github.com/matrix-org/matrix-rust-sdk/issues/5397

crates/matrix-sdk-base/src/room/encryption.rs
```diff
@@ -33,9 +33,12 @@ impl Room {
 #[derive(Debug)]
 #[cfg_attr(feature = "uniffi", derive(uniffi::Enum))]
 pub enum EncryptionState {
-    /// The room is encrypted.
+    /// The room is encrypted, not including state events.
```
jplatte:
```suggestion
    /// The room is encrypted, excluding state events.
```

zecakeh:
For information, #5473 upgrades Ruma and brings in the encrypted state events changes.

Cargo.toml
```diff
@@ -95,8 +96,8 @@ tracing = { version = "0.1.40", default-features = false, features = ["std"] }
 tracing-core = "0.1.32"
 tracing-subscriber = "0.3.18"
 unicode-normalization = "0.1.24"
-uniffi = { version = "0.28.0" }
-uniffi_bindgen = { version = "0.28.0" }
+uniffi = { version = "=0.28.0" }
+uniffi_bindgen = { version = "=0.28.0" }
```
kaylendog: `0.28.0` with no SemVer operators resolves to `^0.28.0` which is satisfied by the latest `0.28.3` which Cargo can automatically upgrade to when dependencies are touched or the lockfile is regenerated, as per the [docs](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html). This version has a breaking change in the method `generate_bindings`, which breaks xtask. Locking this to `=0.28.0` resolves the issue for now.

jplatte: It would be good to report breaking changes in a patch release to UniFFI: https://github.com/mozilla/uniffi-rs/issues

labs/multiverse/src/widgets/room_view/mod.rs
```diff
@@ -552,12 +552,24 @@ impl RoomView {
...
+        self.input.clear();
+    }
+
     async fn handle_command(&mut self, command: input::Command) {
         match command {
             input::Command::Invite { user_id } => self.invite_member(&user_id).await,
             input::Command::Leave => self.leave_room().await,
             input::Command::Subscribe => self.subscribe_thread().await,
             input::Command::Unsubscribe => self.unsubscribe_thread().await,
+            input::Command::Rename { to } => self.rename_room(to).await,
```
richvdh: could you make this a separate PR, since it's not really anything to do with encrypted state?

testing/matrix-sdk-integration-testing/src/tests/e2ee/state_events.rs
```diff
@@ -85,6 +86,7 @@ async fn test_e2e_location_sharing() -> Result<()> {
     let bob_response = bob.sync_once().instrument(bob_span.clone()).await?;
 
     // Bob should have received a to-device event with the payload
+    println!("{:#?}", bob_response.rooms.invited.iter().next().unwrap().1.invite_state);
```
richvdh: this printf-debugging seems to have got left over

richvdh: hrm well, maybe it got removed later, but that suggests the commits need a bit more cleanup

crates/matrix-sdk-crypto/src/olm/group_sessions/outbound.rs
```diff
@@ -519,6 +522,71 @@ impl OutboundGroupSession {
...
+                MegolmV2AesSha2Content { ciphertext, session_id: self.session_id().to_owned() }
+                    .into()
+            }
+            _ => unreachable!(
+                "An outbound group session is always using one of the supported algorithms"
+            ),
+        };
+
+        let content =
+            RoomEncryptedEventContent { scheme, relates_to: None, other: Default::default() };
```
richvdh: We could do with factoring out this code that is duplicated from `encrypt_event` to a separate method.

crates/matrix-sdk-crypto/src/session_manager/group_sessions/mod.rs
```diff
@@ -224,6 +228,49 @@ impl GroupSessionManager {
...
+    ///
+    /// Returns the raw encrypted state event content.
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if saving changes to the store fails.
+    ///
+    /// # Panics
+    ///
+    /// Panics if no such session exists for the given room ID, or the session
```
richvdh, line 251:
```suggestion
    /// Panics if no session exists for the given room ID, or the session
```

crates/matrix-sdk-crypto/src/machine/mod.rs
```diff
@@ -1100,6 +1102,62 @@ impl OlmMachine {
...
+    ///
+    /// # Arguments
+    ///
+    /// * `room_id` - The id of the room for which the message should be
+    ///   encrypted.
+    ///
+    /// * `content` - The plaintext content of the event that should be
+    ///   encrypted as a raw JSON value.
+    ///
+    /// * `state_key` - The associated state key of the event.
```
richvdh, line 1147: `event_type` is missing here, and the order is inconsistent with the actual arguments

crates/matrix-sdk/src/room/futures.rs
```diff
@@ -319,3 +321,83 @@ impl<'a> IntoFuture for SendAttachment<'a> {
         Box::pin(fut.instrument(tracing_span))
     }
 }
+
+/// TODO: Future returned by `Room::send_state_event_raw`.
```
richvdh: what does this TODO mean?

crates/matrix-sdk/src/room/futures.rs
```diff
@@ -319,3 +321,83 @@ impl<'a> IntoFuture for SendAttachment<'a> {
...
+#[allow(missing_debug_implementations)]
+pub struct SendStateEventRaw<'a> {
```
richvdh, line 378: :thinking: I wonder if this should be called `sendRawStateLikeEvent` for consistency with `SendRawMessageLikeEvent`.

crates/matrix-sdk/src/room/futures.rs
```diff
@@ -319,3 +321,83 @@ impl<'a> IntoFuture for SendAttachment<'a> {
...
+    /// # Returns
+    ///
+    /// Returns `true` if the event should be encrypted, otherwise returns
+    /// `false`.
+    #[cfg(feature = "e2e-encryption")]
+    fn should_encrypt(room: &Room, event_type: &str) -> bool {
+        use tracing::debug;
+
+        if !room.encryption_state().is_state_encrypted() {
+            debug!("Sending plaintext event as the room does NOT support encrypted state events.");
```
richvdh, line 430: I'd guess this and the `debug!` below are going to be a bit spammy: maybe turn them down to `trace!` ?

crates/matrix-sdk/src/room/futures.rs
```diff
@@ -401,3 +401,74 @@ impl<'a> SendStateEventRaw<'a> {
...
+            room.ensure_room_joined()?;
+
+            #[cfg(feature = "e2e-encryption")]
+            let mut state_key = state_key.to_owned();
+            #[cfg(not(feature = "e2e-encryption"))]
+            let state_key = state_key.to_owned();
+
+            #[cfg(feature = "e2e-encryption")]
+            if Self::should_encrypt(room, event_type) {
+                use tracing::debug;
```
richvdh, line 479: best hoist this to the top of the file

crates/matrix-sdk/src/room/futures.rs
```diff
@@ -401,3 +401,74 @@ impl<'a> SendStateEventRaw<'a> {
...
+
+                Span::current().record("is_room_encrypted", true);
+                debug!(
+                    room_id = ?room.room_id(),
+                    "Sending encrypted event because the room is encrypted.",
```
richvdh, line 484:
```suggestion
                    "Sending encrypted state event because the room is encrypted.",
```

crates/matrix-sdk/src/room/futures.rs
```diff
@@ -401,3 +401,74 @@ impl<'a> SendStateEventRaw<'a> {
...
+                );
+
+                if !room.are_members_synced() {
+                    room.sync_members().await?;
+                }
+
+                room.query_keys_for_untracked_or_dirty_users().await?;
+                room.preshare_room_key().await?;
```
richvdh: I wonder if we could move all this into `room.preshare_room_key`. Otherwise, maybe we could pull it out to a helper function, so it can be shared with `SendRawMessageLikeEvent` ?

crates/matrix-sdk/src/room/futures.rs
```diff
@@ -401,3 +401,74 @@ impl<'a> SendStateEventRaw<'a> {
...
+
+                let olm = room.client.olm_machine().await;
+                let olm = olm.as_ref().expect("Olm machine wasn't started");
+
+                content = olm
+                    .encrypt_state_event_raw(room.room_id(), event_type, &state_key, &content)
+                    .await?
+                    .cast_unchecked();
```
richvdh, line 495: what's going on with this `cast_unchecked`?

It would be nice to get rid of it, but at the very least we need a comment explaining why it is needed and why we believe it to be safe.

crates/matrix-sdk/src/room/futures.rs
```diff
@@ -322,7 +322,7 @@ impl<'a> IntoFuture for SendAttachment<'a> {
     }
 }
 
-/// TODO: Future returned by `Room::send_state_event_raw`.
+/// Future returned by [`Room::send_state_event_raw`].
```
richvdh, line 376: oh, the TODO got removed in this apparently unrelated commit...

crates/matrix-sdk/src/room/mod.rs
```diff
@@ -2745,22 +2746,13 @@ impl Room {
     /// # anyhow::Ok(()) };
     /// ```
     #[instrument(skip_all)]
-    pub async fn send_state_event_raw(
-        &self,
-        event_type: &str,
-        state_key: &str,
+    pub fn send_state_event_raw<'a>(
```
richvdh, line 2778: we probably need some text in the documentation that says the event will be encrypted if that is enabled (c.f. `send_raw`)

crates/matrix-sdk/src/room/futures.rs
```diff
@@ -322,6 +325,37 @@ impl<'a> IntoFuture for SendAttachment<'a> {
     }
 }
 
+/// TODO: Future returned by `Room::send_state_event`.
```
richvdh, line 331: again, not sure what this TODO is for

crates/matrix-sdk/src/room/futures.rs
```diff
@@ -356,6 +356,20 @@ impl<'a> SendStateEvent<'a> {
     }
 }
 
+impl<'a> IntoFuture for SendStateEvent<'a> {
```
richvdh, line 362: FWIW I think it would have made sense to combine the commits that add the `impl IntoFuture`s with the commits that introduce the `struct` definition in the first place. The commits would still be small enough to be digestible, and obviously `struct SendStateEvent` doesn't make much sense without its associated `impl IntoFuture`.

crates/matrix-sdk/src/room/mod.rs
```diff
@@ -2647,11 +2647,11 @@ impl Room {
     /// # anyhow::Ok(()) };
     /// ```
     #[instrument(skip_all)]
-    pub async fn send_state_event(
-        &self,
+    pub fn send_state_event<'a>(
```
richvdh, line 2683: Again, probably need to update the documentation on these methods to mention encryption.

crates/matrix-sdk/src/room/mod.rs
```diff
@@ -1955,6 +1955,49 @@ impl Room {
...
+
+                self.client.state_store().save_changes(&changes).await?;
+                self.set_room_info(room_info, RoomInfoNotableUpdateReasons::empty());
+            } else {
+                debug!("room successfully marked as encrypted");
+            }
+        }
+
+        Ok(())
+    }
```
richvdh: This is so similar to `enable_encryption` that I'm having a hard time figuring out what's different. Let's factor out a common helper that takes a boolean argument or something.

crates/matrix-sdk-base/src/response_processors/state_events.rs
```diff
@@ -142,6 +145,80 @@ pub mod sync {
...
+                            event_id = outer.event_id().as_str(),
+                            expected = outer_state_key,
+                            found = inner.state_key(),
+                            "Mismatched state key"
+                        );
+                        continue;
+                    }
+
+                    debug!(event_id = ?outer.event_id(), "Decrypted state event successfully.");
+                    room_info.handle_state_event(&inner);
```
richvdh, line 219: this looks a bit complicated to be in the middle of a `match` statement. Pull it out to a separate function, please.

crates/matrix-sdk-base/src/response_processors/state_events.rs
```diff
@@ -142,6 +145,80 @@ pub mod sync {
...
+                    // Check state keys match, discard if not.
+                    if outer_state_key != inner.state_key() {
+                        warn!(
+                            event_id = outer.event_id().as_str(),
+                            expected = outer_state_key,
+                            found = inner.state_key(),
+                            "Mismatched state key"
+                        );
+                        continue;
+                    }
```
richvdh, line 216: This functionality belongs in the crypto crate: it is not specific to /sync, but should happen whereever we decrypt a state event.

It seems like we we need an `olm_machine.decrypt_state_event` which does these checks somehow. Not sure exactly what the types for that look like off the top of my head: see if you can figure something out, if not give me a ping to see if we can figure something out together.

crates/matrix-sdk/src/test_utils/mod.rs
```diff
@@ -266,6 +266,35 @@ macro_rules! assert_decrypted_message_eq {
...
+
+        let content =
+            deserialized_event.original_content().expect("The event should not have been redacted");
+
+        assert_matches2::assert_let!($pat = content);
+    };
+    ($pat:pat = $event:expr) => {
+        assert_let_decrypted_state_event_content!(
+            $pat = $event,
+            "The decrypted event did not match to the expected value"
```
richvdh, line 293:
```suggestion
            "The decrypted event did not match the expected value"
```

crates/matrix-sdk/src/test_utils/mod.rs
```diff
@@ -266,6 +266,35 @@ macro_rules! assert_decrypted_message_eq {
...
+
+        let deserialized_event = decrypted_event
+            .event
+            .deserialize_as_unchecked::<$crate::ruma::events::AnyStateEvent>()
+            .expect("We should be able to deserialize the decrypted event");
+
+        let content =
+            deserialized_event.original_content().expect("The event should not have been redacted");
+
+        assert_matches2::assert_let!($pat = content);
```
richvdh, line 288: Looks like you've forgotten to use `$msg` here.

crates/matrix-sdk/src/test_utils/mod.rs
```diff
@@ -266,6 +266,35 @@ macro_rules! assert_decrypted_message_eq {
     }};
 }
 
+/// Given a [`TimelineEvent`], assert that the event is a decrypted state
+/// event, and that its content matches the given pattern via a let binding.
```
richvdh, line 270: Some examples might help me understand how to use this.

crates/matrix-sdk/src/test_utils/mod.rs
```diff
@@ -266,6 +266,35 @@ macro_rules! assert_decrypted_message_eq {
...
+#[macro_export]
+macro_rules! assert_let_decrypted_state_event_content {
+    ($pat:pat = $event:expr, $($msg:tt)*) => {
```
richvdh, line 273: The documentation could do with mentioning that the macro uses the second and subsequent arguments, if provided, as an error message if the content does not match.

testing/matrix-sdk-integration-testing/src/tests/e2ee/state_events.rs
```diff
@@ -0,0 +1,150 @@
...
+    let alice_room = alice
+        .create_room(assign!(CreateRoomRequest::new(), {
+            name: Some("Cat Photos".to_owned()),
+            preset: Some(RoomPreset::PublicChat),
+        }))
+        .await?;
+    alice_room.enable_encryption_with_state().await?;
+
+    // HACK: wait for sync
+    let _ = tokio::time::sleep(Duration::from_secs(3)).await;
```
richvdh, line 65: We need to find something to wait for, rather than just sleeping for 3 seconds. There's a `wait_until_some` helper which I imagine you could combine with `is_state_encrypted` to do what you need.

richvdh: that said... didn't we agree this was a bug in `enable_encryption[_with_state]` which should be fixed in a separate PR?

crates/matrix-sdk-crypto/src/store/types.rs
```diff
@@ -411,6 +411,9 @@ pub struct RoomSettings {
     /// The encryption algorithm that should be used in the room.
     pub algorithm: EventEncryptionAlgorithm,
 
+    /// Whether state event encryption is enabled.
+    pub encrypt_state_events: bool,
```
richvdh, line 415: RoomSettings is serialized and persisted in clients' local stores. We'll need to make sure it behaves sensibly when the serialized version omits the field. (In other words, needs `#[serde(default)]` or something)

The commit comment is a bit weird too: it says "WASM SDK" but ... this isn't the wasm bindings project.

TBH I'd suggest pulling this RoomSettings stuff out to a separate PR.

richvdh (changes requested):
General comments:

 * Yay! excites! It generally looks very sensible.

 * I wonder if all this functionality should be behind compile-time cfg guards, since it's entirely unspecced. @poljar: any thoughts on this?

 * For future reference: your commit comments could do with a little more detail. Try and explain how the bit of functionality that bit introduces fits in with the whole.

 * Before this is ready to land in the rust SDK, I think we need some more (or, indeed, any) unit tests, which exercise individual parts of the functionality without rolling it into a big integration test.
 
   For sure an integration test is nice to make sure it all works as a whole, but integration tests are slooow and make it painful to figure out which bit of the code is misbehaving. Plus, unit tests make it easier to exercise the non-happy paths since you can mock out the homeserver. Codecov is having one of its "special" days, but when it recovers I'd expect it to complain about lack of code coverage. There's a concept called the [Test Pyramid](https://martinfowler.com/articles/practical-test-pyramid.html#TheTestPyramid) which says you should have lots of low-level tests and a few high-level tests. 
 
   (Generally, I'd expect to see code, and unit tests that exercise that code, landing in the same commit.)

 * We should be able to break this PR into smaller chunks, especially if we have more tests. Smaller PRs are generally a good thing since it means that author and reviewer have less state to keep in their head as the PR goes through multiple rounds of review. (Some guy called richvdh wrote a [blog post](https://richvdh.org/rubber-stamp-reviews-considered-harmful.html#keep-prs-small) mentioning this.)

crates/matrix-sdk/src/room/futures.rs
kaylendog: I was going for consistency with Ruma - it uses `StateEvent` and `MessageLikeEvent`.

crates/matrix-sdk/src/room/futures.rs
kaylendog: I've done this in a couple of places where the only usage of an import is behind a feature gate, which causes the CI to complain. This saves on `#[cfg(feature = "e2e-encryption")]` needing to be applied to imports, but I can do that if that's preferable.

crates/matrix-sdk/src/room/futures.rs
kaylendog: Hmm, this particular cast is not necessary and I will remove it, but casting remains at the [heart](https://github.com/kaylendog/matrix-rust-sdk/blob/61d40bd330e1f8f914ae92e0c2fd463813c2c4b8/crates/matrix-sdk-crypto/src/olm/group_sessions/outbound.rs#L582) of the way state encryption is done at the moment. @poljar initially suggested I use casting, but I think it is worth exploring ways to do this in a type-safe manner.

crates/matrix-sdk/src/room/futures.rs
jplatte: (and Ruma uses "message-like" to disambiguate from `m.room.message` / `m.message`)

crates/matrix-sdk-crypto/src/store/types.rs
kaylendog: I included WASM SDK in the commit message since this change (afaik) will only affect it and not the rest of the Rust SDK, since it uses the `m.room.encryption` event to decide whether to encrypt rather than this - I can change?

richvdh: I think this probably just falls into the more general bracket of "your commit comments could do with a little more detail". You're allowed to use more than one line ;)

kaylendog: I have pulled this out to https://github.com/matrix-org/matrix-rust-sdk/pull/5511 - I'll drop the commit that introduces this after our meeting.

kaylendog:
Closing in favour of split pull requests:

- #5511
- #5512
- #5523 
- #5519
- SDK encrypt
- SDK decrypt/sync

PR #5460, opened 2025-07-29
refactor(test): avoid proliferation of builder submethods in the `MockClientBuilder`

bnjbvr:
Instead of having one static method duplicating an underlying `ClientBuilder` method, we can pass the builder directly to a closure, that will replace it. Call sites are a bit more verbose, but that would avoid having to add duplicate `MockClientBuilder` methods for each `ClientBuilder` method.

Not too sure about this one; wondering if the current way is better, or if the proposed enhanced way is?

### DIFF
--- a/crates/matrix-sdk/src/test_utils/client.rs
+++ b/crates/matrix-sdk/src/test_utils/client.rs
@@ -14,10 +14,7 @@
 
 //! Augmented [`ClientBuilder`] that can set up an already logged-in user.
 
-use matrix_sdk_base::{
-    store::{RoomLoadSettings, StoreConfig},
-    SessionMeta,
-};
+use matrix_sdk_base::{store::RoomLoadSettings, SessionMeta};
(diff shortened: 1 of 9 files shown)


crates/matrix-sdk/src/test_utils/client.rs
```diff
@@ -121,30 +101,9 @@ impl MockClientBuilder {
...
-    pub fn sqlite_store(mut self, path: impl AsRef<std::path::Path>) -> Self {
-        self.builder = self.builder.sqlite_store(path, None);
-        self
-    }
-
-    /// Handle refreshing access tokens automatically.
-    pub fn handle_refresh_tokens(mut self) -> Self {
-        self.builder = self.builder.handle_refresh_tokens();
+    /// Apply changes to the underlying [`ClientBuilder`].
+    pub fn on_builder<F: FnOnce(ClientBuilder) -> ClientBuilder>(mut self, f: F) -> Self {
```
poljar: Can we add an example to this docstring, I can imagine newcomers being confused by the `FnOnce`.

PR #5461, opened 2025-07-29
feat(threads): filter out notifications from threads we haven't subscribed to 

bnjbvr:
Based on #5455, this makes it possible to filter out notifications for a thread we haven't subscribed to.

The rules are the following:

- if the thread is subscribed to: keep the notification
- if the thread has no subscription status:
  - if the new event would cause a subscription: keep the notification
  - if the new event wouldn't cause a subscription: dismiss the notification
- if the thread is unsubscribed from: dismiss the notification


bnjbvr:
Closing: we don't need this, since the evaluation of push rules does it for us (duh! but now I know better about push rules).

The only remaining interesting things have been / will be extracted to other PRs:

- #5494 for the simplifications spotted during this work
- the boolean for filtering out unsubscribed threads will go in #5483

PR #5462, opened 2025-07-29
feat(threads): automatically subscribe a user to a thread under the semantics of MSC4306

bnjbvr:
Based on #5455 + #5461, this implements automatic subscription to threads which explicitly mention the current user (via an intentional room or user mention), and likewise sends a manual subscription to a thread when replying into it.

There's a single background task, owned by the event cache, that will listen to two input streams, and create automatic thread subscriptions based on these streams.

- one stream is the global event cache linked chunk update stream, introduced in this PR (and that could be fused in the future with the `RoomEventCacheGenericUpdate` stream). Based on the new events propagated in a linked chunk update, we can compute the notifications for those, and decide to subscribe to a thread, based on those.
- the other stream is the global send queue stream (thanks for adding this in another PR!). Ideally, we'd have the full event's content when it's sent, and we could only react to `Sent` updates. We could embed the sent event in the `Sent` variant, as a simplification; otherwise, we keep just a bit of state, to link the to-be-sent event (transaction id) to its thread root.

Topics to discuss:

- having the event cache own this task is a bit strange since it also listens to the send queue. We could put it in the client, but at this point, since both the send q and the event cache should be enabled at the same time, and we'd want to have the event cache stream around, I thought it was just as simple to put it in the event cache.
- as hinted above: the send q update `Sent` could embed the sent event, which would avoid the mapping of to-be-sent event to a thread root.

PR is a bit big, but mostly because of the tests; the core code isn't that big.

Part of #4869.

### DIFF
--- a/crates/matrix-sdk/src/event_cache/mod.rs
+++ b/crates/matrix-sdk/src/event_cache/mod.rs
@@ -426,6 +426,12 @@ impl EventCache {
         self.inner.generic_update_sender.subscribe()
     }
 
+    /// React to a given linked chunk update by subscribing the user to a
+    /// thread, if needs be (when the user got mentioned in a thread reply, for
+    /// a thread they were not subscribed to).
+    ///
+    /// Returns a boolean indicating whether the task should keep on running or
+    /// not.
     #[instrument(skip(client, thread_subscriber_sender))]
     async fn handle_thread_subscriber_linked_chunk_update(
         client: &WeakClient,
@@ -438,7 +444,7 @@ impl EventCache {
             return false;
         };
 
-        let OwnedLinkedChunkId::Thread(room_id, thread_root) = &up.linked_chunk else {
+        let OwnedLinkedChunkId::Thread(room_id, thread_root) = &up.linked_chunk_id else {
             trace!("received an update for a non-thread linked chunk, ignoring");
             return true;
         };
@@ -450,8 +456,9 @@ impl EventCache {
 
         let thread_root = thread_root.clone();
 
-        let new_events = up.events();
-        if new_events.is_empty() {
+        let mut new_events = up.events().peekable();
+
+        if new_events.peek().is_none() {
             // No new events, nothing to do.
             return true;
         }
@@ -482,8 +489,9 @@ impl EventCache {
         let mut subscribe_up_to = None;
 
         // Find if there's an event that would trigger a mention for the current
-        // user, iterating from the end of the new events towards the oldest,
-        for ev in new_events.into_iter().rev() {
+        // user, iterating from the end of the new events towards the oldest, so we can
+        // find the most recent event to subscribe to.
+        for ev in new_events.rev() {
             if push_context
                 .for_event(ev.raw())
                 .await
@@ -513,6 +521,12 @@ impl EventCache {
         true
     }
 
+    /// React to a given send queue update by subscribing the user to a
+    /// thread, if needs be (when the user sent an event in a thread they were
+    /// not subscribed to).
+    ///
+    /// Returns a boolean indicating whether the task should keep on running or
+    /// not.
     #[instrument(skip(client, thread_subscriber_sender))]
     async fn handle_thread_subscriber_send_queue_update(
         client: &WeakClient,
@@ -703,7 +717,8 @@ struct EventCacheInner {
     /// A sender for a persisted linked chunk update.
     ///
     /// This is used to notify that some linked chunk has persisted some updates
-    /// to a store, and can be used by observers to look for new events.
+    /// to a store, during sync or a back-pagination of *any* linked chunk.
+    /// This can be used by observers to look for new events.
     ///
     /// See doc comment of [`RoomEventCacheLinkedChunkUpdate`].
     linked_chunk_update_sender: Sender<RoomEventCacheLinkedChunkUpdate>,
@@ -957,34 +972,37 @@ pub struct RoomEventCacheGenericUpdate {
 #[derive(Clone, Debug)]
 struct RoomEventCacheLinkedChunkUpdate {
     /// The linked chunk affected by the update.
-    linked_chunk: OwnedLinkedChunkId,
+    linked_chunk_id: OwnedLinkedChunkId,
 
-    /// A vector of all the updates that happened during this update.
+    /// A vector of all the linked chunk updates that happened during this event
+    /// cache update.
     updates: Vec<linked_chunk::Update<TimelineEvent, Gap>>,
 }
 
 impl RoomEventCacheLinkedChunkUpdate {
     /// Return all the new events propagated by this update, in topological
     /// order.
-    pub fn events(self) -> Vec<TimelineEvent> {
-        self.updates
-            .into_iter()
-            .flat_map(|update| match update {
-                linked_chunk::Update::PushItems { items, .. } => items,
-                linked_chunk::Update::ReplaceItem { item, .. } => vec![item],
-                linked_chunk::Update::RemoveItem { .. }
-                | linked_chunk::Update::DetachLastItems { .. }
-                | linked_chunk::Update::StartReattachItems
-                | linked_chunk::Update::EndReattachItems
-                | linked_chunk::Update::NewItemsChunk { .. }
-                | linked_chunk::Update::NewGapChunk { .. }
-                | linked_chunk::Update::RemoveChunk(..)
-                | linked_chunk::Update::Clear => {
-                    // All these updates don't contain any new event.
-                    vec![]
-                }
-            })
-            .collect()
+    pub fn events(self) -> impl DoubleEndedIterator<Item = TimelineEvent> {
+        use itertools::Either;
+        self.updates.into_iter().flat_map(|update| match update {
+            linked_chunk::Update::PushItems { items, .. } => {
+                Either::Left(Either::Left(items.into_iter()))
+            }
+            linked_chunk::Update::ReplaceItem { item, .. } => {
+                Either::Left(Either::Right(std::iter::once(item)))
+            }
+            linked_chunk::Update::RemoveItem { .. }
+            | linked_chunk::Update::DetachLastItems { .. }
+            | linked_chunk::Update::StartReattachItems
+            | linked_chunk::Update::EndReattachItems
+            | linked_chunk::Update::NewItemsChunk { .. }
+            | linked_chunk::Update::NewGapChunk { .. }
+            | linked_chunk::Update::RemoveChunk(..)
+            | linked_chunk::Update::Clear => {
+                // All these updates don't contain any new event.
+                Either::Right(std::iter::empty())
+            }
+        })
     }
 }
 
--- a/crates/matrix-sdk/src/event_cache/room/threads.rs
+++ b/crates/matrix-sdk/src/event_cache/room/threads.rs
@@ -55,6 +55,10 @@ pub(crate) struct ThreadEventCache {
     /// A sender for live events updates in this thread.
     sender: Sender<ThreadEventCacheUpdate>,
 
+    /// A sender for the globally observable linked chunk updates that happened
+    /// during a sync or a back-pagination.
+    ///
+    /// See also [`super::super::EventCacheInner::linked_chunk_update_sender`].
     linked_chunk_update_sender: Sender<RoomEventCacheLinkedChunkUpdate>,
 }
 
@@ -102,7 +106,7 @@ impl ThreadEventCache {
 
         let _ = self.linked_chunk_update_sender.send(RoomEventCacheLinkedChunkUpdate {
             updates,
-            linked_chunk: OwnedLinkedChunkId::Thread(
+            linked_chunk_id: OwnedLinkedChunkId::Thread(
                 self.room_id.clone(),
                 self.thread_root.clone(),
             ),
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3084,6 +3084,7 @@ dependencies = [
  "imbl",
  "indexmap",
  "insta",
+ "itertools 0.14.0",
  "js_int",
  "language-tags",
  "matrix-sdk-base",
--- a/crates/matrix-sdk/Cargo.toml
+++ b/crates/matrix-sdk/Cargo.toml
@@ -84,6 +84,7 @@ futures-util.workspace = true
 http.workspace = true
 imbl = { workspace = true, features = ["serde"] }
 indexmap.workspace = true
+itertools.workspace = true
 js_int = "0.2.2"
 language-tags = { version = "0.3.2" }
 matrix-sdk-base.workspace = true
--- a/crates/matrix-sdk/src/event_cache/room/mod.rs
+++ b/crates/matrix-sdk/src/event_cache/room/mod.rs
@@ -1260,7 +1260,7 @@ mod private {
 
             // Forward that the store got updated to observers.
             let _ = self.linked_chunk_update_sender.send(RoomEventCacheLinkedChunkUpdate {
-                linked_chunk: OwnedLinkedChunkId::Room(self.room.clone()),
+                linked_chunk_id: OwnedLinkedChunkId::Room(self.room.clone()),
                 updates,
             });
 


crates/matrix-sdk/src/event_cache/pagination.rs
```diff
@@ -298,6 +303,8 @@ impl RoomPagination {
                 });
             }
 
+            subscribe_to_new_threads(&room, new_thread_subs).await;
```
Hywan: Can you add a comment here to explain why we call `subscribe_to_new_threads` at this place please?

crates/matrix-sdk/src/event_cache/room/threads.rs
```diff
@@ -258,11 +267,126 @@ impl ThreadEventCache {
...
+pub struct AutomaticThreadSubscriptions(pub BTreeMap<OwnedEventId, OwnedEventId>);
+
+/// A thin wrapper over a specialized [`PushContext`] that is used for computing
+/// the threads subscriptions automatically.
+///
+/// This can be created only using the
+/// [`push_context_for_threads_subscriptions`] function, available in the same
+/// module.
+#[derive(Debug, Default)]
+pub struct ThreadPushContext(Option<PushContext>);
```
Hywan: Why not `type ThreadPushContext = Option<PushContext>`? What the benefits of having a new type?

crates/matrix-sdk/src/event_cache/room/threads.rs
```diff
@@ -258,11 +267,126 @@ impl ThreadEventCache {
...
     }
 }
+
+/// Mapping from the thread root event id, to the latest event in thread
+/// that should cause a subscription.
+///
+/// Note: it's not necessarily the latest event in the thread, but the latest in
+/// the thread that matches a push condition rule.
+#[derive(Debug, Default)]
+pub struct AutomaticThreadSubscriptions(pub BTreeMap<OwnedEventId, OwnedEventId>);
```
Hywan: Why not `type AutomaticThreadSubscriptions  = BTreeMap<OwnedEventId, OwnedEventId>`? What the benefits of having a new type?

crates/matrix-sdk/src/room/mod.rs
```diff
@@ -3053,7 +3065,19 @@ impl Room {
...
     /// actions for events.
     pub async fn push_context(&self) -> Result<Option<PushContext>> {
-        let Some(push_condition_room_ctx) = self.push_condition_room_ctx().await? else {
+        self.push_context_internal(self.client.enabled_thread_subscriptions()).await
+    }
+
+    /// Retrieves a [`PushContext`] that can be used to compute the push actions
+    /// for events, with a choice to include thread subscriptions or not,
+    /// based on the extra `with_threads_subscriptions` parameter.
+    pub(crate) async fn push_context_internal(
```
Hywan, line 3075: I would add an `#[instrument]` here, so that the `debug!` a couple lines below will provide more context (like `with_threads_subscriptions`).

crates/matrix-sdk/src/event_cache/pagination.rs
Hywan: Re-reading this, I don't understand why we subscribe to threads here :-].

Hywan (commented):
Looks good. I don't understand all the details for the moment. I've left a couple of questions. Thanks for your contributions!

crates/matrix-sdk/src/event_cache/room/threads.rs
bnjbvr: Newtypes allow hiding the inner workings. I don't want users to pass a general `ThreadPushContext` to the method that determines whether we should subscribe to a thread or not, because the push context shouldn't have the thread-subscription check. So it has to be a specific flavor of `PushContext` to be passed in there, and I don't want the user to care about this; they can create a `ThreadPushContext` with the other function, and not think too much about the implementation details.

crates/matrix-sdk/src/event_cache/room/threads.rs
bnjbvr: Yep, for this one a type alias would make sense :+1:

crates/matrix-sdk/src/event_cache/pagination.rs
bnjbvr: This is the "main" events timeline, which has threaded *and* non-threaded events. So when we're back-paginating, we might run into threaded for which no subscriptions have ever been computed (i.e. the threads existed before we created the subscription mechanism). In this case, we still want to be able to "catch up" and automatically subscribe the user.

bnjbvr:
Ready for another round of review :partying_face: This also implements support for manually subscribing when an event is sent via the send queue, as per the semantics of MSC4306.

PTAL :partying_face:

crates/matrix-sdk/src/event_cache/mod.rs
```diff
@@ -657,6 +671,42 @@ pub struct RoomEventCacheGenericUpdate {
...
 }
 
+/// An update being triggered when events change in the persisted event cache
+/// for any room.
+#[derive(Clone, Debug)]
+struct RoomEventCacheLinkedChunkUpdate {
+    /// The linked chunk affected by the update.
+    linked_chunk: OwnedLinkedChunkId,
+
+    /// A vector of all the updates that happened during this update.
```
Hywan:
```suggestion
    /// A vector of all the linked chunk updates that happened during this event cache update.
```

crates/matrix-sdk/src/event_cache/mod.rs
```diff
@@ -657,6 +671,42 @@ pub struct RoomEventCacheGenericUpdate {
     pub room_id: OwnedRoomId,
 }
 
+/// An update being triggered when events change in the persisted event cache
+/// for any room.
+#[derive(Clone, Debug)]
+struct RoomEventCacheLinkedChunkUpdate {
+    /// The linked chunk affected by the update.
+    linked_chunk: OwnedLinkedChunkId,
```
Hywan: Small suggestion to clarify the manipulated data:

```suggestion
    linked_chunk_id: OwnedLinkedChunkId,
```

crates/matrix-sdk/src/event_cache/mod.rs
```diff
@@ -657,6 +671,42 @@ pub struct RoomEventCacheGenericUpdate {
...
+                | linked_chunk::Update::NewGapChunk { .. }
+                | linked_chunk::Update::RemoveChunk(..)
+                | linked_chunk::Update::Clear => {
+                    // All these updates don't contain any new event.
+                    vec![]
+                }
+            })
+            .collect()
+    }
+}
```
Hywan: How great this method takes a `self` by value! We can remove all `Vec` allocations here with:

We can use `Either`, [which implements `Iterator` on specific case](https://docs.rs/either/1.15.0/either/enum.Either.html#impl-Iterator-for-Either%3CL,+R%3E), to represent all possible cases. Since we have 3 cases: (i) many items, (ii) one item, (iii) zero item, the `Either` type is like `Either<C0, Either<C1, C2>>`.

```rust
    /// Return all the new events propagated by this update, in topological
    /// order.
    pub fn events(self) -> impl Iterator<Item = TimelineEvent> {
        use std::iter;

        use itertools::Either;

        self.updates.into_iter().flat_map(|update| match update {
            linked_chunk::Update::PushItems { items, .. } => Either::Left(items.into_iter()),
            linked_chunk::Update::ReplaceItem { item, .. } => {
                Either::Right(Either::Left(iter::once(item)))
            }
            linked_chunk::Update::RemoveItem { .. }
            | linked_chunk::Update::DetachLastItems { .. }
            | linked_chunk::Update::StartReattachItems
            | linked_chunk::Update::EndReattachItems
            | linked_chunk::Update::NewItemsChunk { .. }
            | linked_chunk::Update::NewGapChunk { .. }
            | linked_chunk::Update::RemoveChunk(..)
            | linked_chunk::Update::Clear => {
                // All these updates don't contain any new event.
                Either::Right(Either::Right(iter::empty()))
            }
        })
    }
```

Note that `events` now returns a `impl Iterator<Item = T>` instead of a `Vec<T>`: no more allocations at all.

(you may need to add `itertools.workspace = true` in the `matrix-sdk/Cargo.toml` file).

crates/matrix-sdk/src/event_cache/room/threads.rs
```diff
@@ -51,12 +54,24 @@ pub(crate) struct ThreadEventCache {
 
     /// A sender for live events updates in this thread.
     sender: Sender<ThreadEventCacheUpdate>,
+
+    linked_chunk_update_sender: Sender<RoomEventCacheLinkedChunkUpdate>,
```
Hywan, line 62: Documentation is missing.

crates/matrix-sdk/src/event_cache/room/threads.rs
```diff
@@ -78,6 +93,22 @@ impl ThreadEventCache {
         }
     }
 
+    // TODO(bnjbvr): share more code with `RoomEventCacheState` to avoid the
```
Hywan, line 100: Is this `TODO` addressed later? (just adding a comment to see if that's the case)

crates/matrix-sdk/src/room/mod.rs
```diff
@@ -3740,6 +3740,26 @@ impl Room {
         }
     }
 
+    /// Subscribe to a thread if needed, based on a current subscription to it.
+    ///
+    /// This is like [`Self::subscribe_thread`], but it first checks if the user
+    /// has already subscribed to a thread, so as to minimize sending
+    /// unnecessary subscriptions which would be ignored by the server.
```
Hywan, line 3772: What's the point of keeping `Self::subscribe_thread` in the public scope then? When one would need to call it?

crates/matrix-sdk/src/event_cache/mod.rs
```diff
@@ -657,6 +671,42 @@ pub struct RoomEventCacheGenericUpdate {
     pub room_id: OwnedRoomId,
 }
 
+/// An update being triggered when events change in the persisted event cache
+/// for any room.
+#[derive(Clone, Debug)]
+struct RoomEventCacheLinkedChunkUpdate {
```
Hywan, line 973: This type isn't tested in this commit neither the next commit.

crates/matrix-sdk/src/event_cache/mod.rs
```diff
@@ -183,10 +191,20 @@ impl EventCache {
...
+                thread_subscriber_receiver,
             }),
         }
     }
 
+    /// Subscribes to updates that a thread subscription has been sent.
+    ///
+    /// For testing purposes only.
+    #[doc(hidden)]
+    pub fn subscribe_thread_subscriber_updates(&self) -> Receiver<()> {
```
Hywan, line 215: Can't we have a `#[cfg(test)]` instead?

crates/matrix-sdk/src/event_cache/mod.rs
```diff
@@ -396,6 +414,115 @@ impl EventCache {
...
+
+                    let Some(room) = client.get_room(&room_id) else {
+                        warn!(%room_id, "unknown room");
+                        continue;
+                    };
+
+                    let thread_root = thread_root.clone();
+
+                    let new_events = up.events();
+                    if new_events.is_empty() {
```
Hywan: Since `events()` could return a `impl Iterator`, we may want to change `.is_empty()` for `.peekable().peek().is_none()`.

crates/matrix-sdk/src/event_cache/mod.rs
```diff
@@ -441,6 +568,21 @@ struct EventCacheInner {
...
+    /// One important constraint is that there is only one such task per
+    /// [`EventCache`], so it does listen to *all* rooms at the same time.
+    _thread_subscriber_task: AbortOnDrop<()>,
+
+    /// A test helper receiver that will be emitted every time the thread
+    /// subscriber task subscribed to a new thread.
+    ///
+    /// This is helpful for tests to coordinate that a new thread subscription
+    /// has been sent or not.
+    thread_subscriber_receiver: Receiver<()>,
```
Hywan, line 739: Can we put it behind a `#[cfg(test)]`?

crates/matrix-sdk/src/event_cache/mod.rs
```diff
@@ -415,6 +418,92 @@ impl EventCache {
         self.inner.generic_update_sender.subscribe()
     }
 
+    async fn handle_thread_subscriber_linked_chunk_update(
```
Hywan, line 436: Can you explain what the returned `bool` means please? A `ControlFlow` would have clarified that 😇 😉.

crates/matrix-sdk/src/event_cache/mod.rs
```diff
@@ -501,7 +511,102 @@ impl EventCache {
...
+        true
+    }
+
+    #[instrument(skip(client, thread_subscriber_sender))]
+    async fn handle_thread_subscriber_send_queue_update(
+        client: &WeakClient,
+        thread_subscriber_sender: &Sender<()>,
+        events_being_sent: &mut HashMap<OwnedTransactionId, OwnedEventId>,
+        up: SendQueueUpdate,
+    ) -> bool {
```
Hywan, line 536: Same for `-> bool`. Also, I believe these methods deserve a bit of documentation, thought?

crates/matrix-sdk-common/src/serde_helpers.rs
```diff
@@ -45,6 +48,22 @@ struct SimplifiedContent {
...
 }
 
+/// Try to extract the thread root from an event's content, if provided.
+///
+/// The thread root is the field located at `m.relates_to`.`event_id`,
+/// if the field at `m.relates_to`.`rel_type` is `m.thread`.
+///
+/// Returns `None` if we couldn't find a thread root, or if there was an issue
+/// during deserialization.
+pub fn extract_thread_root_from_content(
```
Hywan, line 58: Nice! I was mentioning such an optimisation in a previous comment, glad to see it here!

crates/matrix-sdk/src/event_cache/room/threads.rs
bnjbvr: To be honest, it wouldn't be quite useful to have a doc comment here, as it would only repeat the name and types :P But sure.

bnjbvr: I'm getting out of my way to make this comment useful lol

crates/matrix-sdk/src/event_cache/room/threads.rs
bnjbvr: Nope it is not. I'd expect all these refactorings to happen around the same time, when we unify backpagination in threads and in the main room -> when persisting threads happens.

crates/matrix-sdk/src/room/mod.rs
bnjbvr: It's a good question. I'd say it's useful only until we've proven the other API isn't buggy, or when some user of this API is super paranoid and would rather make sure to always propagate the update to the server.

If that's fine with you, I'd suggest to keep it for the time being, and remove it later, when the sync of thread subscriptions is in place.

crates/matrix-sdk/src/event_cache/mod.rs
bnjbvr: What do you mean by a type that's not tested? Do you mean testing the methods on the type? As the method is tested in the integration with the thread subscriber task, and the type itself is mostly POD, I'm not sure to see the value in testing it in isolation…

crates/matrix-sdk/src/event_cache/mod.rs
bnjbvr: No, the integration tests are considered to be outside the scope of the crate, so we need a public method (or to resurrect the infamous `testing` feature).

bnjbvr: that being said, the `testing` feature is still alive, so we don't need to resurrect it, i could use that there :+1:

bnjbvr: uh, actually, we'd need to sprinkle `cfg` guards everywhere, which would be annoying, so i'm tempted to let it as is.

crates/matrix-sdk/src/event_cache/mod.rs
bnjbvr: (Same as above)

crates/matrix-sdk/src/event_cache/mod.rs
bnjbvr: Wow, you iterators magician! Thanks, that applies cleanly.

crates/matrix-sdk/src/event_cache/room/threads.rs
Hywan: No problem.

crates/matrix-sdk/src/room/mod.rs
Hywan: Alrighty.

crates/matrix-sdk/src/event_cache/mod.rs
Hywan: If it's tested via an integration test, that's good.

PR #5466, opened 2025-07-30
ffi (bindings): add suggested role and power level value based functions

Velin92:
The reason for doing this is that in the EX clients we also allow the creators to set other members as creators even if they are not among the original ones, this will set them with a power level of 150 (like superadmins).

This introduces two extra functions that have the purpose of be used if you want an equivalent power level value for a creator.


Velin92:
No worries I'll close this PR... the problem is that I think the best solution would be to define a specific role for the 150 PL admin, that is in between the admin itself and the creator. This would still be a bit of a hack.

Issue #5420, opened 2025-07-17
`matrix-sdk-indexeddb`  is only tested against Firefox


mgoldenberg:
Currently, the tests for `matrix-sdk-indexeddb` are [only run against Firefox](https://github.com/matrix-org/matrix-rust-sdk/blob/1a9e5b904b6ea28caf9efaea543aaefc3960a229/xtask/src/ci.rs#L432) and not against Chrome (or any other browsers).

Furthermore, when running tests against Chrome, there are a number of failures. Note that some of these are tests of code that I introduced - i.e., those in `matrix_sdk_indexeddb::event_cache_store`.

```bash
> wasm-pack test --chrome --headless -- crates/matrix-sdk-indexeddb

# -- snip --

failures:

    state_store::encrypted_tests::statestore_integration_tests::test_topic_redaction_outer
    state_store::migrations::tests::test_migrating_to_v4_outer
    event_cache_store::tests::encrypted::indexeddb_event_cache_store_integration_tests::test_linked_chunk_replace_item
    event_cache_store::tests::encrypted::indexeddb_event_cache_store_integration_tests::test_linked_chunk_remove_item
    event_cache_store::tests::encrypted::indexeddb_event_cache_store_integration_tests::test_linked_chunk_detach_last_items
    state_store::tests::statestore_integration_tests::test_topic_redaction_outer
    event_cache_store::tests::unencrypted::indexeddb_event_cache_store_integration_tests::test_linked_chunk_replace_item
    event_cache_store::tests::unencrypted::indexeddb_event_cache_store_integration_tests::test_linked_chunk_remove_item
    event_cache_store::tests::unencrypted::indexeddb_event_cache_store_integration_tests::test_linked_chunk_detach_last_items

test result: FAILED. 168 passed; 9 failed; 2 ignored; 0 filtered out; finished in 19.61s
```

All of these fail due to errors with the use of cursors. The error messages are of the following form.

```rust
DomException { name: "InvalidStateError", message: "Failed to execute '<operation>' on 'IDBCursor': The cursor is being iterated or has iterated past its end.", code: 11 })
```

[...]

PR #5467, opened 2025-07-30
IndexedDB: prepare `matrix-sdk-indexeddb` crate for updating `indexed_db_futures` to latest version

mgoldenberg:
## Background

This pull request makes the preliminary changes necessary to update [`indexed_db_futures`](https://docs.rs/indexed_db_futures/latest/indexed_db_futures/index.html) in the `matrix-sdk-indexeddb` crate. The reason we'd like to update this dependency is because the version currently used does not fully support the Chrome browser (see #5420).

The latest version of `indexed_db_futures` has significant changes. Many of these changes can be integrated without issue. There is, however, a single change which requires substantive modification to the `matrix-sdk-indexeddb` crate. Namely, one cannot access the active transaction in the callback to update the database (for details, see Alorel/rust-indexed-db#66).

This is only an issue in the `crypto_store` module, where migrations are provided access to the active transaction (see [here](https://github.com/matrix-org/matrix-rust-sdk/blob/ca89700dfe9f29dcd823bb10861807f9d75e0634/crates/matrix-sdk-indexeddb/src/crypto_store/migrations/mod.rs#L211)). Furthermore, there is only a single migration (`v11_to_v12`) in the `crypto_store` module which makes use of the active transaction (see [here](https://github.com/matrix-org/matrix-rust-sdk/blob/ca89700dfe9f29dcd823bb10861807f9d75e0634/crates/matrix-sdk-indexeddb/src/crypto_store/migrations/v11_to_v12.rs#L23)).

For clarity, the reason `v11_to_v12` is problematic in the latest versions of `indexed_db_futures` is because it is simply adding an index to an object store which was created in a different migration and this requires access to the active transaction. All the other migrations create object stores and indices in the same migration, which does not suffer from the same issue.

## Changes

Due to the issues above, I have taken the following steps to prepare `matrix-sdk-indexeddb` for an update to `indexed_db_futures`.

1. Modified the schema upgrade logic so that none of the individual migrations expect access to the active transaction 
2. Removed the logic from the `v11_to_v12` migration, which could no longer be accomplished without the active transaction
3. Added a new set of migrations, `v13_to_v15`, which replicate the logic from `v11_to_v12` in a way that is compatible with later versions of `indexed_db_futures`.

## Alternatives

My guess is that this problem is an oversight in the interface developed in the `indexed_db_futures` crate and not a deeper issue - i.e., it seems it would be possible to expose the active transaction in later versions, as is the case in the current [version](https://docs.rs/indexed_db_futures/0.5.0/indexed_db_futures/struct.IdbVersionChangeEvent.html#method.transaction) used in `matrix-sdk-indexeddb`.

Alternatives to the changes in this pull request include the following.

1. Wait for the maintainer of `indexed_db_futures` to make the necessary modifications in a future version
2. Vendor a version with the necessary modifications specifically for `matrix-sdk-indexeddb`

Neither of these seemed particularly expedient or attractive, so I opted for making changes to the migrations.

## Future Work

- Update `indexed_db_futures` to the latest version - i.e., `0.6.4`

---

### DIFF
--- a/crates/matrix-sdk-indexeddb/src/crypto_store/migrations/mod.rs
+++ b/crates/matrix-sdk-indexeddb/src/crypto_store/migrations/mod.rs
@@ -200,13 +200,12 @@ type OldVersion = u32;
 /// * `name` - name of the indexeddb database to be upgraded.
 /// * `version` - version we are upgrading to.
 /// * `f` - closure which will be called if the database is below the version
-///   given. It will be called with three arguments `(db, txn, oldver)`, where:
+///   given. It will be called with two arguments `(db, oldver)`, where:
 ///   * `db` - the [`IdbDatabase`]
-///   * `txn` - the database transaction: a [`IdbTransaction`]
 ///   * `oldver` - the version number before the upgrade.
 async fn do_schema_upgrade<F>(name: &str, version: u32, f: F) -> Result<(), DomException>
 where
-    F: Fn(&IdbDatabase, IdbTransaction<'_>, OldVersion) -> Result<(), JsValue> + 'static,
+    F: Fn(&IdbDatabase, OldVersion) -> Result<(), JsValue> + 'static,
 {
     info!("IndexeddbCryptoStore upgrade schema -> v{version} starting");
     let mut db_req: OpenDbRequest = IdbDatabase::open_u32(name, version)?;
@@ -218,7 +217,7 @@ where
         let old_version = evt.old_version() as u32;
 
         // Run the upgrade code we were supplied
-        f(evt.db(), evt.transaction(), old_version)
+        f(evt.db(), old_version)
     }));
 
     let db = db_req.await?;
--- a/crates/matrix-sdk-indexeddb/src/crypto_store/migrations/mod.rs
+++ b/crates/matrix-sdk-indexeddb/src/crypto_store/migrations/mod.rs
@@ -25,6 +25,7 @@ mod v0_to_v5;
 mod v10_to_v11;
 mod v11_to_v12;
 mod v12_to_v13;
+mod v13_to_v15;
 mod v5_to_v7;
 mod v7;
 mod v7_to_v8;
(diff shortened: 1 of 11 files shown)


mgoldenberg:
@andybalaam: thanks for your response! And yes, I believe you have understood correctly!

Now my turn to make sure I am understanding you: the problem is that the web-based Element client uses the `crypto_store` module of `matrix-sdk-indexeddb` - presumably as a WASM module? And that the user interface is essentially unresponsive during migrations. Finally, the migration I have added will likely take a long time for many users - depending on how much data they have in `inbound_group_sessions3` - and, furthermore, provide no added benefit as the schema hasn't _really_ changed.

That's a rather tricky problem... I hadn't thought about how long the migration might take.

So, the concern with waiting for an upstream patch to `indexed_db_futures` seems like it will be rather open-ended. Unfortunately, the project appears to be maintained by a single person and their current status on GitHub is: ["I'm probs not having a look at that issue you opened anytime soon. Hands full."](https://github.com/Alorel). This, of course, isn't your problem to fix. It just isn't instilling a lot of confidence that we will see a change any time soon 😅

I will confer with the people at @filament-dm, as this work is being done in service of their client. Perhaps we will have to keep this in their fork until we see some movement elsewhere.

Out of curiosity, do you have any concrete sense of when element-hq/element-web#26892 might be resolved? And is this something I can help with?

Also, tagging @zzorba here, in case he wants to weigh in or has other ideas about how to move forward.

andybalaam:
@mgoldenberg yes, I think we are understanding each other very well.

I'd really like to avoid a slow migration for every user when it is not needed. If `rust-indexed-db` is not being updated, it might make sense to fork it, but I guess the first step would be to open a PR against the project and see whether it gets attention.

Help with fixing https://github.com/element-hq/element-web/issues/26892 would be great - it might need some input from Element Design team to decide how it should look, but we could probably get that, or merge a good-enough solution in the meantime.

BUT the more I think about this the less I like adding a delay for all users simply because of a dependency issue.

I guess part of the question is how urgent this is. Maybe we are prepared to wait a while to see whether `rust-indexed-db` accepts a contribution. If the only downsides are that we can't test on Chrome and Filament needs to keep this in their fork for a while, maybe we can accept those for a while?

mgoldenberg:
Closing this in favor of #5722.

PR #5469, opened 2025-07-31
refactor(send queue): tidy up the media upload progress reporting code

bnjbvr:
I wanted to add these as tiny commits to the media upload progress [PR](https://github.com/matrix-org/matrix-rust-sdk/pull/5453), but the changes are nontrivial, so this warrants a review IMO.

The most notable change is that the media upload progress updates won't go through the global send queue listener (that's used to compute the latest event). I don't think every single progress update is significant for an external, global listener, that wants to know about major changes in all rooms. Glad to discuss this, though, if people think differently.

Based on top of https://github.com/matrix-org/matrix-rust-sdk/pull/5453; will rebase as soon as it's merged.


Hywan:
> The most notable change is that the media upload progress updates won't go through the global send queue listener (that's used to compute the latest event). I don't think every single progress update is significant for an external, global listener, that wants to know about major changes in all rooms. Glad to discuss this, though, if people think differently.

Looks sensible to me. The Latest Event can simply ignore the `MediaUpload` state, but I think it's a good think generally speaking. We can revisit that later if required.

PR #5470, opened 2025-07-31
refactor(client): have `upload_encrypted_file` own the `Client` instance

### LINKED ISSUE #5465: api: make it so that `UploadEncryptedFile` clones the `Client` instead of borrowing it
Right now, to create an encrypted upload, it's necessary to have a live client, like this:

```rust
// assuming a `Room` object is around
let mut cursor = std::io::Cursor::new(Vec::new());
let client = room.client();
let mut req = client
    .upload_encrypted_file(&mut cursor)
    .with_request_config(RequestConfig::short_retry());
```

We can't use the following pattern, because the `UploadEncryptedFile` request returned from `Client::upload_encrypted_file` borrows from the `Client` instance:

```
let mut req = room.client()
    .upload_encrypted_file(&mut cursor)
    .with_request_config(RequestConfig::short_retry());
```

However, `SendRequest` (and thus `SendMediaUploadRequest`) do clone the `Client` within them instead of borrowing it. This makes it simpler, and to make the API a bit more consistent, I propose here to do the same for `UploadEncryptedFile`, i.e. change the borrow to an owned `Client` here:

https://github.com/matrix-org/matrix-rust-sdk/blob/92192c549b3f53889c23954386743867a73b31b1/crates/matrix-sdk/src/encryption/futures.rs#L31

This should be a pretty simple change for a newcomer. Note it will need a CHANGELOG entry as it touches the public API (the lifetime of `UploadEncryptedFile` is now tied to that of the reader only, instead of reader+client).

multisme:

Signed-off-by: multi multiestunhappydev@gmail.com

Fixes #5465

### DIFF
--- a/crates/matrix-sdk/CHANGELOG.md
+++ b/crates/matrix-sdk/CHANGELOG.md
@@ -1,10 +1,11 @@
 # Changelog
-
+<t_k€>ku
 All notable changes to this project will be documented in this file.
 
 <!-- next-header -->
 
 ## [Unreleased] - ReleaseDate
+- [**breaking**] Change the upload_encrypted_file and make it clone the client instead of owning it. 
 
 ### Features
 
--- a/crates/matrix-sdk/CHANGELOG.md
+++ b/crates/matrix-sdk/CHANGELOG.md
@@ -1,11 +1,9 @@
 # Changelog
-<t_k€>ku
 All notable changes to this project will be documented in this file.
 
 <!-- next-header -->
 
 ## [Unreleased] - ReleaseDate
-- [**breaking**] Change the upload_encrypted_file and make it clone the client instead of owning it. 
 
 ### Features
 
@@ -31,6 +29,7 @@ All notable changes to this project will be documented in this file.
 
 ### Refactor
 
+- [**breaking**] Change the upload_encrypted_file and make it clone the client instead of owning it. The lifetime of the `UploadEncryptedFile` request returned by `Client::upload_encrypted_file()` only depends on the request lifetime now.
 - [**breaking**] Add an `IsPrefix = False` bound to the `account_data()` and
   `fetch_account_data_static()` methods of `Account`. These methods only worked
   for events where the full event type is statically-known, and this is now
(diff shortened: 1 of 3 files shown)


crates/matrix-sdk/CHANGELOG.md
```diff
@@ -1,10 +1,11 @@
 # Changelog
-
+<t_k�>ku
```
bnjbvr: oopsy :-)

crates/matrix-sdk/CHANGELOG.md
```diff
@@ -1,10 +1,11 @@
...
 All notable changes to this project will be documented in this file.
 
 <!-- next-header -->
 
 ## [Unreleased] - ReleaseDate
+- [**breaking**] Change the upload_encrypted_file and make it clone the client instead of owning it.
```
bnjbvr: Can you put this at the top of the `Refactor` section, and word it like ```The lifetime of the `UploadEncryptedFile` request returned by `Client::upload_encrypted_file()` doesn't depend on the `Client` lifetime anymore.```?

multisme:
Sorry I had a weird local git thing that made me force push

crates/matrix-sdk/CHANGELOG.md
```diff
@@ -1,5 +1,4 @@
 # Changelog
```
bnjbvr, line 1:
```suggestion
# Changelog

```

crates/matrix-sdk/CHANGELOG.md
```diff
@@ -1,5 +1,4 @@
...
-
```
bnjbvr: Let's revert this newline deletion.

bnjbvr (changes requested):
Thanks, but I think you missed some of the feedback from the previous review:

> Can you also change line 864 of send_queue/mod.rs, removing the TODO in there and replace client by room.client() in the one place where it appears, please?

crates/matrix-sdk/CHANGELOG.md
multisme: done

Issue #3697, opened 2024-07-15
Improve performance of encrypted sqlite DB


kegsay:
As seen in https://github.com/matrix-org/complement-crypto/pull/122 which times out due to how slow it gets with encryption enabled via `.Passphrase` being set on the `client_builder`.

Example timings:
 - `TestAliceBobEncryptionWorks/{rust_hs1}|{rust_hs1}` - this creates 2 clients, 1 room and Alice sends 1 message to Bob. It's the simplest E2EE test. 910ms without encrypted DB, 4.86s with.
 - `TestFallbackKeyIsUsedIfOneTimeKeysRunOut` - this consumes all OTKs (blocking the client uploading more) then claims the fallback key, so it will cause the client to re-upload 50 OTKs. 2.52s without, 8.43s with.
 - Overall: 3m29s to run without, 9m28s to run with.

This is 3-5x worse.

This means I can't realistically enable encrypted DBs in tests, which is desirable in order to match EX.

poljar:
I think that this is mostly due to our PBKDF usage in the default case when we encrypt the store. The usage of a random key is possible with our primitive itself:

https://github.com/matrix-org/matrix-rust-sdk/blob/7a85b7abdcfc05881ccbddf54a2094273be90110/crates/matrix-sdk-store-encryption/src/lib.rs#L151-L182

But it's not easily exposed in the `ClientBuilder` like the passphrase based mechanism is:

https://github.com/matrix-org/matrix-rust-sdk/blob/7a85b7abdcfc05881ccbddf54a2094273be90110/crates/matrix-sdk/src/client/builder.rs#L216-L228

We would need a new method to let us create stores with a random key instead.

multisme:
Hi, can you extend more on the need here? Are you talking about adding to BuilderStoreConfig a Option field that would contain the encryption  key?

poljar:
Yes, instead of requiring the usage of a passphrase that is used to derive the encryption key, we should let users provide an encryption key themselves.

This avoids the slow PBKDF step.

PR #5472, opened 2025-07-31
Refactor/improve performance sqlite db

multisme:

Signed-off-by: multi [multiestunhappydev@gmail.com](mailto:multiestunhappydev@gmail.com)

Fixes https://github.com/matrix-org/matrix-rust-sdk/issues/3697

### DIFF
--- a/crates/matrix-sdk-sqlite/CHANGELOG.md
+++ b/crates/matrix-sdk-sqlite/CHANGELOG.md
@@ -13,7 +13,7 @@ No notable changes in this release.
 ## [0.13.0] - 2025-07-10
 
 ### Refactor
-- Change the logic for opening a store so as to use a `key` instead of a `passphrase`
+- [breaking] Change the logic for opening a store so as to use a `Secret` enum in the function `open_with_pool` instead of a `passphrase` (https://github.com/matrix-org/matrix-rust-sdk/pull/5472)
 
 ### Security Fixes
 
--- a/crates/matrix-sdk-sqlite/CHANGELOG.md
+++ b/crates/matrix-sdk-sqlite/CHANGELOG.md
@@ -6,14 +6,19 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased] - ReleaseDate
 
+### Features
+- Implement a new constructrot that allow to open SqliteCryptoStore with a Key
+  ([#5472](https://github.com/matrix-org/matrix-rust-sdk/pull/5472))
+
+### Refactor
+- [breaking] Change the logic for opening a store so as to use a `Secret` enum in the function `open_with_pool` instead of a `passphrase`
+  ([#5472](https://github.com/matrix-org/matrix-rust-sdk/pull/5472))
+
 ## [0.14.0] - 2025-09-04
 
 No notable changes in this release.
 
 ## [0.13.0] - 2025-07-10
-
-### Refactor
-- [breaking] Change the logic for opening a store so as to use a `Secret` enum in the function `open_with_pool` instead of a `passphrase` (https://github.com/matrix-org/matrix-rust-sdk/pull/5472)
 
 ### Security Fixes
 
--- a/crates/matrix-sdk-sqlite/CHANGELOG.md
+++ b/crates/matrix-sdk-sqlite/CHANGELOG.md
@@ -12,6 +12,9 @@ No notable changes in this release.
 
 ## [0.13.0] - 2025-07-10
 
+### Refactor
+- Change the logic for opening a store so as to use a `key` instead of a `passphrase`
+
 ### Security Fixes
 
 - Fix SQL injection vulnerability in `find_event_relations()`.
--- a/crates/matrix-sdk-sqlite/CHANGELOG.md
+++ b/crates/matrix-sdk-sqlite/CHANGELOG.md
@@ -7,7 +7,7 @@ All notable changes to this project will be documented in this file.
 ## [Unreleased] - ReleaseDate
 
 ### Features
-- Implement a new constructrot that allow to open SqliteCryptoStore with a Key
+- Implement a new constructor that allows to open SqliteCryptoStore with a cryptographic key
   ([#5472](https://github.com/matrix-org/matrix-rust-sdk/pull/5472))
 
 ### Refactor
--- a/crates/matrix-sdk-sqlite/src/crypto_store.rs
+++ b/crates/matrix-sdk-sqlite/src/crypto_store.rs
@@ -1543,11 +1543,12 @@ mod tests {
         tmpdir
     }
 
-    async fn get_test_db(data_path: &str, key: Option<&[u8; 32]>) -> TestDb {
+    async fn get_test_db(data_path: &str, passphrase: Option<&str>) -> TestDb {
         let tmpdir = copy_db(data_path);
 
-        let database =
-            SqliteCryptoStore::open(tmpdir.path(), key).await.expect("Can't open the test store");
+        let database = SqliteCryptoStore::open(tmpdir.path(), passphrase)
+            .await
+            .expect("Can't open the test store");
 
         TestDb { _dir: tmpdir, database }
     }
@@ -1943,7 +1944,7 @@ mod encrypted_tests {
             let _ = fs::remove_dir_all(&tmpdir_path).await;
         }
 
-        SqliteCryptoStore::open(tmpdir_path.to_str().unwrap(), Some(passphrase))
+        SqliteCryptoStore::open(tmpdir_path.to_str().unwrap(), Some(pass))
             .await
             .expect("Can't create a key protected store")
     }
--- a/crates/matrix-sdk-sqlite/src/crypto_store.rs
+++ b/crates/matrix-sdk-sqlite/src/crypto_store.rs
@@ -1900,7 +1900,11 @@ mod tests {
         assert!(backup_keys.decryption_key.is_some());
     }
 
-    async fn get_store(name: &str, passphrase: Option<&str>, clear_data: bool) -> SqliteCryptoStore {
+    async fn get_store(
+        name: &str,
+        passphrase: Option<&str>,
+        clear_data: bool,
+    ) -> SqliteCryptoStore {
         let tmpdir_path = TMP_DIR.path().join(name);
 
         if clear_data {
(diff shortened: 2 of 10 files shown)


crates/matrix-sdk-sqlite/src/crypto_store.rs
```diff
@@ -84,17 +84,17 @@ impl EncryptableStore for SqliteCryptoStore {
 
 impl SqliteCryptoStore {
     /// Open the SQLite-based crypto store at the given path using the given
-    /// passphrase to encrypt private data.
+    /// key to encrypt private data.
     pub async fn open(
         path: impl AsRef<Path>,
-        passphrase: Option<&str>,
+        key: Option<&[u8; 32]>,
```
poljar, line 99: You can't just change the constructor like that, that's a pretty big breaking change and nobody will be able to open their existing stores anymore.

This needs to be an alternative constructor. I would imagine that this would then also solve the majority of your CI woes.

crates/matrix-sdk-sqlite/CHANGELOG.md
```diff
@@ -8,6 +8,9 @@ All notable changes to this project will be documented in this file.
 
 ## [0.13.0] - 2025-07-10
 
+### Refactor
+- Change the logic for opening a store so as to use a `key` instead of a `passphrase`
+
```
poljar: This would be a new feature that adds a new constructor. A PR link is also required for the changelog entry.

crates/matrix-sdk-sqlite/src/crypto_store.rs
multisme: So I'm looking in the code and if I understand well,  for example it implies having both a `open_with_pool` function and an  `open_with_pool_with_key`. one?

poljar: Sure, or one function could take a single enum argument that could be the key or passphrase, or a builder style store opener could be created.

The gist of it is that we can't remove the existing functionality.

multisme:
Hey I'm unsure about the `get_or_create_store_cipher` could you point me in the right direction?

crates/matrix-sdk-sqlite/src/utils.rs
```diff
@@ -393,7 +393,7 @@ impl SqliteKeyValueStoreConnExt for rusqlite::Connection {
 ///
 /// ```sql
 /// CREATE TABLE "kv" (
-///     "key" TEXT PRIMARY KEY NOT NULL,
+///     "key" TEXT PRIMARY KEY NOT NULL,d
```
poljar:
```suggestion
///     "key" TEXT PRIMARY KEY NOT NULL,
```

crates/matrix-sdk-sqlite/src/crypto_store.rs
```diff
@@ -83,18 +83,27 @@ impl EncryptableStore for SqliteCryptoStore {
 }
 
 impl SqliteCryptoStore {
-    /// Open the SQLite-based crypto store at the given path using the given
-    /// passphrase to encrypt private data.
+    /// Open the SQLite-based event cache store at the given path using the
```
poljar: This ain't the event cache? Why make this change?

crates/matrix-sdk-sqlite/src/crypto_store.rs
```diff
@@ -1861,7 +1870,7 @@ mod tests {
 
         SqliteCryptoStore::open(tmpdir_path.to_str().unwrap(), passphrase)
             .await
-            .expect("Can't create a passphrase protected store")
+            .expect("Can't create a key protected store")
```
poljar:
```suggestion
            .expect("Can't create an encrypted store")
```

crates/matrix-sdk-sqlite/src/crypto_store.rs
```diff
@@ -1893,7 +1902,7 @@ mod encrypted_tests {
 
         SqliteCryptoStore::open(tmpdir_path.to_str().unwrap(), Some(pass))
             .await
-            .expect("Can't create a passphrase protected store")
+            .expect("Can't create a key protected store")
```
poljar: Same here.

crates/matrix-sdk-sqlite/src/lib.rs
```diff
@@ -43,13 +42,19 @@ pub use self::state_store::{SqliteStateStore, DATABASE_NAME as STATE_STORE_DATAB
 #[cfg(test)]
 matrix_sdk_test::init_tracing_for_tests!();
 
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub enum Secret {
+    Key([u8; 32]),
+    PassPhrase(String),
+}
```
poljar: This needs to be zeroized or not take ownership of the key/passphrase. Not taking ownership would be preferred but might not be easy to achieve.

crates/matrix-sdk-sqlite/src/state_store.rs
```diff
@@ -90,18 +90,27 @@ impl fmt::Debug for SqliteStateStore {
 }
 
 impl SqliteStateStore {
-    /// Open the SQLite-based state store at the given path using the given
-    /// passphrase to encrypt private data.
+    /// Open the SQLite-based event cache store at the given path using the
```
poljar: This ain't the event cache either.

crates/matrix-sdk-sqlite/src/utils.rs
```diff
@@ -457,18 +457,24 @@ pub(crate) trait SqliteKeyValueStoreAsyncConnExt: SqliteAsyncConnExt {
...
-            let export = cipher.export(passphrase);
-            #[cfg(test)]
-            let export = cipher._insecure_export_fast_for_testing(passphrase);
+            //#[cfg(not(test))]
+            let export = match secret {
+                Secret::PassPhrase(passphrase) => cipher.export(&passphrase),
+                Secret::Key(key) => cipher.export_with_key(&key),
+            };
+            //#[cfg(test)]
+            //let export = cipher._insecure_export_fast_for_testing(passphrase);
```
poljar: We want to continue on using this for tests, as we don't want to slow down our CI runtime or local test runs.

multisme: I would need to write a function that would take a key instead of a passphrase in this case.

crates/matrix-sdk-sqlite/src/lib.rs
multisme: If I don't take ownership, is there anyway without using a lifetime, hence having the data live as long as SqliteStoreConfig?

multisme: I went with the zeroize solution

crates/matrix-sdk-sqlite/src/utils.rs
multisme: Rewrote this passage as to use the normal function when a key is given as secret.

multisme:
@poljar Hey I think I managed the implementation, could you provide me with feedback please?

crates/matrix-sdk-sqlite/CHANGELOG.md
```diff
@@ -6,6 +6,14 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased] - ReleaseDate
 
+### Features
+- Implement a new constructrot that allow to open SqliteCryptoStore with a Key
```
poljar:
```suggestion
- Implement a new constructor that allows to open the SqliteCryptoStore with a cryptographic key.
```

crates/matrix-sdk-sqlite/src/lib.rs
```diff
@@ -43,13 +44,21 @@ pub use self::state_store::{SqliteStateStore, DATABASE_NAME as STATE_STORE_DATAB
 #[cfg(test)]
 matrix_sdk_test_utils::init_tracing_for_tests!();
 
+#[derive(Clone, Debug, Eq, PartialEq, Zeroize)]
+pub enum Secret {
+    #[zeroize]
+    Key([u8; 32]),
+    #[zeroize]
+    PassPhrase(String),
```
poljar: The enum variants need to be documented.

crates/matrix-sdk-sqlite/src/lib.rs
```diff
@@ -43,13 +44,21 @@ pub use self::state_store::{SqliteStateStore, DATABASE_NAME as STATE_STORE_DATAB
 #[cfg(test)]
 matrix_sdk_test_utils::init_tracing_for_tests!();
 
+#[derive(Clone, Debug, Eq, PartialEq, Zeroize)]
```
poljar: I think we should derive `ZeroizeOnDrop` here so we don't need the manual `zeroize()` calls.

crates/matrix-sdk-sqlite/src/lib.rs
```diff
@@ -43,13 +44,21 @@ pub use self::state_store::{SqliteStateStore, DATABASE_NAME as STATE_STORE_DATAB
...
+pub enum Secret {
+    #[zeroize]
+    Key([u8; 32]),
```
poljar: This should be inside a `Box` or inside a `Zeroizing` struct so we avoid unintended copies due to moves being a `memcpy` call in Rust land. More info here: https://benma.github.io/2020/10/16/rust-zeroize-move.html.

In short, every time you pass the `Secret` to a function a `memcpy` call will copy the contents of the array.

If a `Box` is used, the pointer is instead copied.

crates/matrix-sdk-sqlite/src/lib.rs
poljar: Same here, this will produce unintended copies, `Zeroizing<String>` would make sense.

poljar (changes requested):
Sorry for the late reply, I was unavailable the past two weeks. This seems almost good to go.

Though we have a tricky problem with hidden copies of the secret key material.

multisme:
Sorry for the force push, I need to manage rebase better.

crates/matrix-sdk-sqlite/src/lib.rs
```diff
@@ -47,13 +48,22 @@ pub use self::state_store::{SqliteStateStore, DATABASE_NAME as STATE_STORE_DATAB
 #[cfg(test)]
 matrix_sdk_test_utils::init_tracing_for_tests!();
 
+/// A enum used to store the secret that gives access to a store
```
Hywan, line 51:
```diff
- /// A enum
+ /// An enum
```

PR #5473, opened 2025-08-01
Upgrade Ruma

zecakeh:
This brings in a new breaking change from Ruma, because not all events are stripped in a room's stripped state. For simplicity and minimal changes, we still consider the events as stripped during processing since we mostly care about the `content`.

This also brings in the support for encrypted state events, fyi.


poljar (approved):
Looks good.

PR #5476, opened 2025-08-01
feature(search): initialize integrate search crate

dragonfly1033:
A new crate with a basic API for a creating, populating and searching a room message index.  
Added it as an optional feature in matrix-sdk with intgration such that:
- joining a room creates an index for it
- messages are added to a room's index on arrival.


crates/matrix-sdk-search/src/index.rs
```diff
@@ -0,0 +1,243 @@
+#![forbid(missing_docs)]
```
poljar: We'll want configure this for the whole crate, isn't it already so? Also we'll want to include copyright headers in each file that is copyrightable.

crates/matrix-sdk-search/src/index.rs
```diff
@@ -0,0 +1,243 @@
...
+
+        Ok(last_commit_opstamp)
+    }
+
+    /// Forces [`RoomIndex`] to commit changes.
+    /// Nominally [`RoomIndex`] stores up changes and batch commits.
+    /// However, for some circumstances (such as testing), it may be
+    /// beneficial to know a commit occurred before moving on.
+    pub fn force_commit(&mut self) -> Result<OpStamp, IndexError> {
+        let opstamp = self.writer.force_commit()?;
```
poljar: Not important now, but I think that this will block. The writer in Tantivy is operating inside a separate thread, documents are added using a channel and once you commit the writer will write the updated index to the file.

I think that reloading the reader will wait for the writer to finish an block until it does so. We'll need to convert this to a future or async function using `tokio::spawn_blocking()`.

Not in this PR, but let's add a note or TODO item for this.

crates/matrix-sdk-search/src/index.rs
```diff
@@ -0,0 +1,243 @@
...
+        Ok(opstamp)
+    }
+
+    /// Search the [`RoomIndex`] for some query. Returns a list of
+    /// results with a maximum given length.
+    pub fn search(
+        &self,
+        query: &str,
+        max_number_of_results: usize,
+    ) -> Result<Vec<String>, IndexError> {
```
poljar: This should return the event IDs.

crates/matrix-sdk-search/src/index.rs
```diff
@@ -0,0 +1,243 @@
...
+        let query = self.query_parser.parse_query(query)?;
+        let searcher = self.reader.searcher();
+
+        let results = searcher.search(&query, &TopDocs::with_limit(max_number_of_results))?;
+        let mut ret: Vec<String> = Vec::new();
+        let pk = self.schema.primary_key();
+
+        for (_score, doc_address) in results {
+            let retrieved_doc: TantivyDocument = searcher.doc(doc_address)?;
+            for f in retrieved_doc.get_all(pk) {
```
poljar: If we're certain that we only add a single value for the field we can use `get_first()` instead.

crates/matrix-sdk-search/Cargo.toml
```diff
@@ -0,0 +1,29 @@
...
+repository = "https://github.com/matrix-org/matrix-rust-sdk"
+rust-version.workspace = true
+version = "0.1.0"
+
+[package.metadata.docs.rs]
+all-features = true
+rustdoc-args = ["--cfg", "docsrs", "--generate-link-to-definition"]
+
+[dependencies]
+tantivy = "0.22"
```
poljar: What year is it :sweat_smile:

```suggestion
tantivy = "0.24.2"
```

crates/matrix-sdk-search/src/index.rs
```diff
@@ -0,0 +1,243 @@
...
+                match f {
+                    OwnedValue::Str(s) => ret.push(s.to_string()),
+                    _ => println!("how"),
```
poljar: If the returned thing isn't a string, i.e. it has an unexpected type we should log a `error!()` or `warn!()` log line.

crates/matrix-sdk-search/src/lib.rs
```diff
@@ -0,0 +1,12 @@
...
+mod schema;
+mod util;
+mod writer;
+
+/// A module for errors relating to the search crate.
+pub mod error;
+/// A module for the search index.
+pub mod index;
+/// A module for testing the search crate.
+pub mod testing;
```
poljar: I think that this module duplicates a lot of the stuff we have in the `matrix-sdk-test` crate, more specifically the `EventFactory`.

crates/matrix-sdk-search/src/util.rs
```diff
@@ -0,0 +1,7 @@
+#![forbid(missing_docs)]
+
+use std::time::Duration;
+
+pub(crate) const TANTIVY_INDEX_MEMORY_BUDGET: usize = 50_000_000;
+pub(crate) const MIN_COMMIT_SIZE: usize = 500;
+pub(crate) const MAX_COMMIT_TIME: Duration = Duration::from_secs(5);
```
poljar: This can live in the `lib.rs` file. they aren't really utilities more like constans that are used across the library.

crates/matrix-sdk-search/src/writer.rs
```diff
@@ -0,0 +1,61 @@
...
+            || (Instant::now() - self.last_commit_time > MAX_COMMIT_TIME)
+        {
+            self.last_commit_opstamp = self.commit_impl()?;
+            self.last_commit_time = Instant::now();
+            self.document_count = 0;
+        } else {
+            self.document_count += 1;
+        }
+
+        Ok(self.last_commit_opstamp)
```
poljar: This is a trick from Seshat, my memory is not completely gone yet. It was used to lower the amount of actual commits because the JS SDK or EW don't let you easily know when a `/sync` response is done, i.e. when you have added all the events that were part of a single response.

In the Rust SDK we can give you batches of events which will server as the natural boundary for a commit call.

crates/matrix-sdk-search/src/writer.rs
```diff
@@ -0,0 +1,61 @@
...
+    }
+
+    pub(crate) fn force_commit(&mut self) -> Result<OpStamp, IndexError> {
+        self.last_commit_opstamp = self.commit_impl()?;
+        self.last_commit_time = Instant::now();
+        self.document_count = 0;
+        Ok(self.last_commit_opstamp)
+    }
```
poljar: If we actually commit in the commit function instead lying to you, then we don't need this method.

Makes for a simpler API.

crates/matrix-sdk/Cargo.toml
```diff
@@ -62,6 +62,8 @@ docsrs = ["e2e-encryption", "sqlite", "indexeddb", "sso-login", "qrcode"]
 # Add support for inline media galleries via msgtypes
 unstable-msc4274 = ["ruma/unstable-msc4274", "matrix-sdk-base/unstable-msc4274"]
 
+search = ["matrix-sdk-search"]
```
poljar:
```suggestion
experimental-search = ["matrix-sdk-search"]
```

crates/matrix-sdk/Cargo.toml
```diff
@@ -16,7 +16,7 @@ features = ["docsrs"]
 rustdoc-args = ["--cfg", "docsrs", "--generate-link-to-definition"]
 
 [features]
-default = ["e2e-encryption", "automatic-room-key-forwarding", "sqlite", "native-tls"]
+default = ["e2e-encryption", "automatic-room-key-forwarding", "sqlite", "native-tls", "search"]
```
poljar: Let's not make it default until we have left the experimentation stage.

crates/matrix-sdk/src/client/builder/mod.rs
```diff
@@ -823,6 +858,11 @@ pub enum ClientBuildError {
     #[cfg(feature = "sqlite")]
     #[error(transparent)]
     SqliteStore(#[from] matrix_sdk_sqlite::OpenStoreError),
+
+    /// No index base directory was configured
+    #[cfg(feature = "search")]
+    #[error("no index base directory was configured")]
+    MissingIndexBaseDirectory,
```
poljar: We're defaulting to a in-RAM index, I don't think this is needed because of that.

crates/matrix-sdk/src/client/mod.rs
```diff
@@ -2800,6 +2826,44 @@ impl Client {
...
         &self.base_client().decryption_settings
     }
+
+    /// Add [`AnyMessageLikeEvent`] to [`RoomIndex`] of given [`RoomId`]
+    #[cfg(feature = "search")]
+    pub async fn index_event(&self, event: AnyMessageLikeEvent, room_id: &RoomId) -> Result<()> {
+        let mut hash_map = self.inner.room_indexes.lock().await;
+        if let Some(index) = hash_map.get_mut(room_id) {
+            let res = index.add_event(event).map(|_| ());
+            drop(hash_map);
```
poljar: This drop isn't needed. The drop is automatically happening right in that place. Before the result is returned to the caller.

crates/matrix-sdk/src/client/mod.rs
```diff
@@ -2800,6 +2826,44 @@ impl Client {
     pub fn decryption_settings(&self) -> &DecryptionSettings {
         &self.base_client().decryption_settings
     }
+
+    /// Add [`AnyMessageLikeEvent`] to [`RoomIndex`] of given [`RoomId`]
+    #[cfg(feature = "search")]
+    pub async fn index_event(&self, event: AnyMessageLikeEvent, room_id: &RoomId) -> Result<()> {
+        let mut hash_map = self.inner.room_indexes.lock().await;
+        if let Some(index) = hash_map.get_mut(room_id) {
+            let res = index.add_event(event).map(|_| ());
```
poljar: Let's make our life a bit easier, since we can't bubble up this right to the `Client` authors we can just log this here.

Let's log a `debug!()` message for event parsing errors, since those are user-provided content and can always contain malformed data. For the Tantivy errors where the writer fails to add/stage a document let's log a `error!()` but I think that with our setup, where the writer is alive as long as the `Client`, those errors can't be triggered.

crates/matrix-sdk/src/client/mod.rs
```diff
@@ -2800,6 +2826,44 @@ impl Client {
...
+    pub async fn add_index(&self, room_id: &RoomId) -> Result<()> {
+        let mut hash_map = self.inner.room_indexes.lock().await;
+        if !hash_map.contains_key(room_id) {
+            let index = match &self.inner.index_base_dir {
+                IndexBaseDir::Directory(path) => RoomIndex::new(path, room_id)?,
+                IndexBaseDir::Ram => RoomIndex::new_in_ram(room_id)?,
+            };
+            hash_map.insert(room_id.to_owned(), index);
+        }
+        drop(hash_map);
```
poljar: Another redundant drop.

crates/matrix-sdk/src/error.rs
```diff
@@ -419,6 +421,11 @@ pub enum Error {
     /// An error happened while attempting to change power levels.
     #[error("power levels error: {0}")]
     PowerLevels(#[from] PowerLevelsError),
+
+    /// A search index error happened.
+    #[cfg(feature = "search")]
+    #[error(transparent)]
+    Index(#[from] IndexError),
```
poljar: I'm not 100% sure, but I think all the new methods can return `std::Result<(), IndexError>`.

crates/matrix-sdk/src/event_cache/mod.rs
```diff
@@ -116,6 +116,10 @@ pub enum EventCacheError {
         /// A string containing details about the error.
         details: String,
     },
+
+    /// An error occurred higher up in the SDK.
+    #[error(transparent)]
+    SdkError(#[from] error::Error),
```
poljar: If we return a `IndexError` for the relevant indexing methods, then this can be a `IndexError`.

crates/matrix-sdk/src/event_cache/room/mod.rs
```diff
@@ -523,6 +529,45 @@ impl RoomEventCacheInner {
...
+    #[cfg(feature = "search")]
+    async fn index_timeline_event(&self, event: &TimelineEvent) -> Result<()> {
+        if let Some(message_event) = self.parse_timeline_event(event) {
+            if let Some(room) = self.weak_room.get() {
+                return room.index_event(message_event).await.map_err(EventCacheError::from);
+            } else {
+                return Err(EventCacheError::RoomNotFound { room_id: self.room_id.to_owned() });
+            }
+        }
+        Ok(())
```
poljar: The Ok isn't needed if you get rid of the `return` calls.

crates/matrix-sdk/src/event_cache/room/mod.rs
```diff
@@ -523,6 +529,45 @@ impl RoomEventCacheInner {
...
+    #[cfg(feature = "search")]
+    fn parse_timeline_event(&self, event: &TimelineEvent) -> Option<AnyMessageLikeEvent> {
+        let maybe_try_event = match &event.kind {
+            TimelineEventKind::Decrypted(d) => Some(d.event.deserialize()),
+            TimelineEventKind::PlainText { event } => match event.deserialize() {
+                Ok(event_obj) => match event_obj {
+                    AnySyncTimelineEvent::MessageLike(sync_event) => {
+                        Some(Ok(sync_event.into_full_event(self.room_id.clone())))
+                    }
+                    AnySyncTimelineEvent::State(_) => None,
```
poljar: Maybe add a `TODO:` item that we need to handle redactions.

crates/matrix-sdk/src/room/mod.rs
```diff
@@ -3636,6 +3641,12 @@ impl Room {
         opts.send(self, event_id).await
     }
 
+    /// Add an [`AnyMessageLikeEvent`] to this room's [`RoomIndex`]
+    #[cfg(feature = "search")]
+    pub async fn index_event(&self, event: AnyMessageLikeEvent) -> Result<()> {
+        self.client.index_event(event, self.room_id()).await
+    }
+
```
poljar: I don't think we want to make this `pub`. Users of the create should not really need to manually index events themselves, we'll take care of that if they correctly set up their sync.

crates/matrix-sdk/src/sliding_sync/client.rs
```diff
@@ -199,6 +201,23 @@ impl SlidingSyncResponseProcessor {
...
+                .collect();
+            let batch_size = 5;
+            let mut buffer = futures_util::stream::iter(futures).buffer_unordered(batch_size);
+            while let Some(result) = buffer.next().await {
+                if let Err(e) = result {
+                    error!("error occurred while adding index for joined room: {e:?}");
+                }
+            }
+        }
+
```
poljar: Let's just rely on the lazy creation of the index when we search/add events to the index.

Otherwise we'll need to add a similar code snippet to the classical sync response handling path and have too many places where we create the index.

crates/matrix-sdk-search/src/index.rs
```diff
@@ -0,0 +1,243 @@
...
+
+        assert!(index.is_ok(), "failed to make index in ram: {:?}", index.unwrap_err());
+    }
+
+    #[test]
+    fn test_add_event() {
+        let room_id = room_id!("!room_id:localhost");
+        let index = RoomIndex::new_in_ram(room_id);
+
+        assert!(index.is_ok(), "failed to make index in ram: {:?}", index.unwrap_err());
```
poljar: This is a bit more idiomatic.

```rust
index.expect("Failed to make an in-RAM index.");
```

crates/matrix-sdk-search/src/index.rs
```diff
@@ -0,0 +1,243 @@
...
+
+        let mut index = index.unwrap();
+
+        let event = NewEventBuilder::new()
+            .content(MessageEventContent::plain("event message"))
+            .event_id(event_id!("$event_id:localhost"))
+            .build_as(AnyMessageLikeEvent::Message);
+
+        let res = index.add_event(event);
```
poljar: Another place for an `expect` call.

crates/matrix-sdk-search/src/index.rs
```diff
@@ -0,0 +1,243 @@
...
+    #[test]
+    fn test_add_event() {
+        let room_id = room_id!("!room_id:localhost");
+        let index = RoomIndex::new_in_ram(room_id);
+
+        assert!(index.is_ok(), "failed to make index in ram: {:?}", index.unwrap_err());
+
+        let mut index = index.unwrap();
+
+        let event = NewEventBuilder::new()
```
poljar: I think that this can be converted to use the `EventFactory` without any additions to it.

poljar (changes requested):
Alright, I left a bunch of comments. As a first step this looks roughly like the correct shape for this.

One thing that we should do is, split the PR up into multiple PRs. The natural place where a split would happen is between the introduction of the `matrix-sdk-search` and the integration into the rest of our crates.

So after addressing the comments please split this up by creating a new branch and using `git checkout -p` or some other mechanism you might prefer.

dragonfly1033:
Changes made. PR split into 2 parts: Adding of the search crate, integration with matrix-sdk. This PR is now only the addition of the search crate.

crates/matrix-sdk-search/src/error.rs
```diff
@@ -0,0 +1,164 @@
+// Copyright 2024 The Matrix.org Foundation C.I.C.
```
Hywan, line 1: 2025 ^^

crates/matrix-sdk-search/src/index.rs
```diff
@@ -0,0 +1,285 @@
+// Copyright 2024 The Matrix.org Foundation C.I.C.
```
Hywan, line 1: 2025 ^^

PR #5477, opened 2025-08-02
refactor(base): Don't drop whole `m.room.create` if `predecessor` is invalid

zecakeh:
The `m.room.create` event contains at lot of important information for a room, like its type (i.e. whether it is a space), its version and its creator(s) (which are important in room version 12). So ignoring the event completely might break a room.

Instead what we do here is simply ignore the `predecessor` field if it is considered invalid, allowing us to access the other fields.

In the process, this also adds a bug fix, where the content of a redacted create event was not validated, although it still contains the same fields.

### DIFF
--- a/crates/matrix-sdk-base/src/response_processors/state_events.rs
+++ b/crates/matrix-sdk-base/src/response_processors/state_events.rs
@@ -107,24 +107,16 @@ pub mod sync {
                 }
 
                 AnySyncStateEvent::RoomCreate(create) => {
-                    if super::is_create_event_valid(
+                    let edited_create = super::validate_create_event_predecessor(
                         context,
                         room_info.room_id(),
                         create,
                         state_store,
-                    ) {
-                        room_info.handle_state_event(event);
-                    } else {
-                        error!(
-                            room_id = ?room_info.room_id(),
-                            ?create,
-                            "`m.create.tombstone` event is invalid, it creates a loop"
-                        );
+                    );
 
-                        // Do not add the event to `room_info`.
-                        // Do not add the event to `context.state_changes.state`.
-                        continue;
-                    }
+                    room_info.handle_state_event(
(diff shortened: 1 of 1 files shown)


crates/matrix-sdk-base/src/response_processors/state_events.rs
```diff
@@ -317,31 +309,47 @@ where
...
     already_seen.insert(room_id.to_owned());
 
-    let Some(mut predecessor_room_id) = event
-        .as_original()
-        .and_then(|event| Some(event.content.predecessor.as_ref()?.room_id.clone()))
+    // Redacted and non-redacted create events use the same content type.
+    let content = match event {
+        SyncStateEvent::Original(event) => &event.content,
+        SyncStateEvent::Redacted(event) => &event.content,
+    };
```
Hywan, line 329: I think `m.room.create` cannot be redacted. That's why `as_original()` was called previously. I'm fine with the change (better be safe), but maybe it should be mentioned.

Hywan (changes requested):
Thanks for the fix!

Can we add a test ensuring the `m.room.create` state event isn't dropped when the predecessor is invalid please?

crates/matrix-sdk-base/src/response_processors/state_events.rs
zecakeh: It can be redacted, but redaction has no effect in recent room versions. Given that Ruma relies on the detection of `redacted_because` in `unsigned`, not on the format of the content, it is entirely possible to get the `Redacted` variant here.

zecakeh:
> Can we add a test ensuring the m.room.create state event isn't dropped when the predecessor is invalid please?

I added extra checks on the current tests for invalid predecessors, is that not enough?

Hywan (approved):

Hywan:
Sorry I misread the patch. All good!

PR #5478, opened 2025-08-03
fix(sdk): Allow legacy push rules to be missing when changing `NotificationSettings`

zecakeh:
They are being removed from the spec with MSC4210, so we can't error if they are not found.


bnjbvr (approved):
Thanks! do you think it's test-worthy, if it's not too much hassle writing such a test?

zecakeh:
Test added!

PR #5479, opened 2025-08-04
change(room_list): request space rooms through sliding sync and expose a room list filter for them

stefanceriu:
This is a breaking change as spaces are now requested through sliding sync and they need to manually be excluded from the room list by using the newly introduced non-space filter.


crates/matrix-sdk-ui/src/room_list_service/filters/non_space.rs
```diff
@@ -0,0 +1,59 @@
+// Copyright 2024 The Matrix.org Foundation C.I.C.
```
Hywan: Time flies. We are in 2025.

crates/matrix-sdk-ui/src/room_list_service/filters/non_space.rs
```diff
@@ -0,0 +1,59 @@
...
+impl<F> NonSpaceRoomMatcher<F>
+where
+    F: Fn(&Room) -> bool,
+{
+    fn matches(&self, room: &Room) -> bool {
+        (self.is_not_space)(room)
+    }
+}
+
+/// Create a new filter that will filter out rooms that are not spaces.
```
Hywan: Should we explain what a space is? Maybe with a link to the spec, like https://spec.matrix.org/latest/client-server-api/#spaces?

crates/matrix-sdk-ui/CHANGELOG.md
```diff
@@ -7,7 +7,9 @@ All notable changes to this project will be documented in this file.
 ## [Unreleased] - ReleaseDate
 
 ### Features
-
+- [**breaking**] Space rooms are now being retrieved through sliding sync and the newly introduced 
+  `NonSpace` filter should be used to exclude them from any room list.
```
Hywan: The filter name is `new_filter_non_space`. Ideally, you can prefix with `room_list_service::filters::new_filter_non_space` to indicate it's about the RoomList service, because so far the changelog entry doesn't mention it.

PR #5480, opened 2025-08-04
fix(xtask): Use `--package` instead of `-p` on `cargo ndk`

Hywan:
We use `cargo ndk -p {package_name}`, where `-p` is short for `--package`. However, `cargo ndk` has introduced the `-p` option (see https://github.com/bbqsrc/cargo-ndk/commit/c6b93a89a2723ff0fac99b5ac86ae6636a6cf54a), short for `--platform`. It creates a confusion and the command line doesn't execute properly. Let's use the long option `--package` to clarify everything.

### DIFF
--- a/xtask/src/kotlin.rs
+++ b/xtask/src/kotlin.rs
@@ -164,7 +164,7 @@ fn build_for_android_target(
     let sh = sh();
     cmd!(
         sh,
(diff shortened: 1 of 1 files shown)


stefanceriu (approved):
Makes sense to me 👍

PR #5481, opened 2025-08-04
Add setting to ignore the `timeout` setting on first sync

zecakeh:
The `timeout` settings on the `/sync` endpoint is the maximum allowed time for the server to send its response, because this is a poll-based API. It means that if there is no new data to show, the server will wait until the end of `timeout` before returning a response.

It can be an undesirable behavior when starting a client and informing the user that we are "catching up" while waiting for the first response.

By not setting a `timeout` on the first request to `/sync`, the homeserver should reply immediately, whether the response is empty or not.

### DIFF
--- a/crates/matrix-sdk/src/client/mod.rs
+++ b/crates/matrix-sdk/src/client/mod.rs
@@ -2607,14 +2607,17 @@ impl Client {
 
         let parent_span = Span::current();
 
-        async_stream::stream! {
+        async_stream::stream!({
             loop {
                 trace!("Syncing");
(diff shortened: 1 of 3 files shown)


crates/matrix-sdk/src/client/mod.rs
```diff
@@ -2607,14 +2607,17 @@ impl Client {
 
         let parent_span = Span::current();
 
-        async_stream::stream! {
+        async_stream::stream!({
```
Hywan, line 2613: Interesting hack. I've learned something, thanks!

PR #5482, opened 2025-08-04
feat(sdk): Introduce the new `LatestEvents` API, part 3: Computing from the Send Queue

Hywan:
WIP

---


crates/matrix-sdk/src/latest_events/mod.rs
```diff
@@ -567,10 +588,13 @@ async fn listen_to_event_cache_and_send_queue_updates_task(
...
     event_cache_generic_updates_subscriber: &mut broadcast::Receiver<RoomEventCacheGenericUpdate>,
+    send_queue_generic_updates_subscriber: &mut broadcast::Receiver<SendQueueUpdate>,
     listened_rooms: &mut HashSet<OwnedRoomId>,
-    latest_event_queue_sender: &mpsc::UnboundedSender<OwnedRoomId>,
+    latest_event_queue_sender: &mpsc::UnboundedSender<LatestEventQueueUpdate>,
 ) -> ControlFlow<()> {
-    select! {
+    // We need a biased select here: `room_registration_receiver` must have the
+    // priority over other futures.
+    select_biased! {
```
bnjbvr: It this different from `tokio::select! { biased; … }` that we've used in a few other places in the code base? Would be sweet to consistently use one or the other.

crates/matrix-sdk/src/latest_events/mod.rs
```diff
@@ -592,14 +616,31 @@ async fn listen_to_event_cache_and_send_queue_updates(
...
+        send_queue_generic_update = send_queue_generic_updates_subscriber.recv().fuse() => {
+            if let Ok(SendQueueUpdate { room_id, update }) = send_queue_generic_update {
+                if listened_rooms.contains(&room_id) {
+                    let _ = latest_event_queue_sender.send(LatestEventQueueUpdate::SendQueue {
+                        room_id,
+                        update
+                    });
+                }
+            } else {
+                error!("`send_queue_generic_updates` channel has been closed");
```
bnjbvr: It's a bit harsh to use the `error!` level here; this might just be the client shutting down, right?

crates/matrix-sdk/src/latest_events/latest_event.rs
```diff
@@ -139,26 +135,50 @@ impl LatestEventValue {
...
             .await
+            .map(Self::Remote)
             .unwrap_or_default()
     }
 }
 
-fn find_and_map(
+/// A latest event value!
+#[derive(Debug, Clone)]
+pub enum LatestEventKind {
```
bnjbvr: I think it would make sense to rename this `LatestEventContent`, since it contains the content of each variant?

crates/matrix-sdk/src/latest_events/latest_event.rs
```diff
@@ -139,26 +135,50 @@ impl LatestEventValue {
...
-            .rfind_map_event_in_memory_by(|event| find_and_map(event, power_levels))
+            .rfind_map_event_in_memory_by(|event| find_and_map_timeline_event(event, power_levels))
             .await
+            .map(Self::Remote)
             .unwrap_or_default()
     }
 }
 
-fn find_and_map(
+/// A latest event value!
```
bnjbvr: This doc comment is a copy/pasto of the one for `LatestEventValue`; let's update it :)

crates/matrix-sdk/src/latest_events/latest_event.rs
```diff
@@ -139,26 +135,50 @@ impl LatestEventValue {
...
+
+    /// A `m.call.invite` event.
+    CallInvite(CallInviteEventContent),
+
+    /// A `m.call.notify` event.
+    CallNotify(CallNotifyEventContent),
+
+    /// A `m.room.member` event, more precisely a knock membership change that
+    /// can be handled by the current user.
+    KnockedStateEvent(Option<RoomMemberEventContent>),
```
bnjbvr: Why is the content optional? Can you add the response to that question in the doc comment please?

crates/matrix-sdk/src/latest_events/latest_event.rs
```diff
@@ -242,18 +229,58 @@ fn find_and_map(
...
+) -> Option<LatestEventKind> {
+    match event {
+        AnyMessageLikeEventContent::RoomMessage(message) => {
+            // Don't show incoming verification requests.
+            if let MessageType::VerificationRequest(_) = message.msgtype {
+                return None;
+            }
+
+            // Check if this is a replacement for another message. If it is, ignore
+            // it.
```
bnjbvr, line 620: Can you add a TODO, instead? I think that if we want to support showing an edit to the latest message in the room, we should support it, at some point.

crates/matrix-sdk/src/latest_events/latest_event.rs
```diff
@@ -242,18 +229,58 @@ fn find_and_map(
     }
 }
 
+fn find_and_map_any_message_like_event_content(
```
bnjbvr: This name is a bit long and not very descriptive IMO. How about `extract_suitable_content` or `extract_content_from_suitable_event`?

crates/matrix-sdk/src/latest_events/mod.rs
```diff
@@ -1259,10 +1260,7 @@ mod tests {
             .sync_room(
                 &client,
                 JoinedRoomBuilder::new(&room_id).add_timeline_event(
-                    event_factory
-                        .text_msg("venez découvrir cette nouvelle raclette !")
```
bnjbvr, line 1076: what happened to the nice raclette ad :(

crates/matrix-sdk/src/latest_events/mod.rs
```diff
@@ -536,10 +536,30 @@ impl RoomLatestEvents {
...
     /// send queue update.
     async fn update_with_send_queue(&mut self, send_queue_update: &RoomSendQueueUpdate) {
-        self.for_the_room.update_with_send_queue(send_queue_update).await;
+        // Get the power levels of the user for the current room if the `WeakRoom` is
+        // still valid.
+        //
+        // Get it once for all the updates of all the latest events for this room (be
+        // the room and its threads).
+        let room = self.weak_room.get();
+        let power_levels = match &room {
```
bnjbvr: (not technically a big deal, but this is a lie, as it's `user_id_and_power_levels`, right?

crates/matrix-sdk/src/latest_events/latest_event.rs
```diff
@@ -150,6 +169,332 @@ impl LatestEventValue {
...
+        match send_queue_update {
+            // A new local event is being sent.
+            //
+            // Let's create the `LatestEventValue` and push it in the buffer of values.
+            RoomSendQueueUpdate::NewLocalEvent(LocalEcho {
+                transaction_id,
+                content: local_echo_content,
+            }) => match local_echo_content {
+                LocalEchoContent::Event { serialized_event: content, .. } => {
+                    if let Ok(content) = content.deserialize() {
```
bnjbvr: Can you `warn!` with the error, when deserialization failed?

crates/matrix-sdk/src/latest_events/latest_event.rs
```diff
@@ -150,6 +169,332 @@ impl LatestEventValue {
...
+                    power_levels,
+                )
+                .await
+            }
+
+            // A local event has successfully been sent!
+            //
+            // Unwedge all wedged values after the one matching `transaction_id`. Indeed, if
+            // an event has been sent, it means the send queue is working, so if any value has been
+            // marked as wedged, it must be marked as unwedged. Then, remove the calculated
```
bnjbvr: :bug: The second sentence starting with `Indeed` shows "spooky action at a distance" tight coupling. On the one hand, it's bad in terms of design, on the other hand, it's likely wrong in the sense that "wedged" means there was a fatal error and there's no way a retry without updating the event could fix it. So values marked as wedged shouldn't be marked unwedged, in this case.

crates/matrix-sdk/src/latest_events/latest_event.rs
```diff
@@ -150,6 +169,332 @@ impl LatestEventValue {
...
+                .await
+            }
+
+            // A local event has been replaced by another one.
+            //
+            // Replace the latest event value matching `transaction_id` in the buffer if it exists
+            // (note: it should!), and return the last `LatestEventValue` or calculate a new one.
+            RoomSendQueueUpdate::ReplacedLocalEvent { transaction_id, new_content: content } => {
+                if let Some(position) = buffer_of_values_for_local_events.position(transaction_id) {
+                    if let Ok(content) = content.deserialize() {
```
bnjbvr: ditto regarding logging the deserialization error here

crates/matrix-sdk/src/latest_events/latest_event.rs
```diff
@@ -150,6 +169,332 @@ impl LatestEventValue {
...
+                        if let Some(kind) = find_and_map_any_message_like_event_content(content) {
+                            buffer_of_values_for_local_events.replace_kind(position, kind);
+                        }
```
bnjbvr: :bug: It's a bit of an edge case, but a suitable event content can become unsuitable after the `ReplacedLocalEvent`. E.g. a `m.room.message` local echo has been replaced with one including a replacement relation; as edits aren't handled (per my previous review comment), there won't be any extracted content here, so we will never replace the local value. In this case, I think we should likely remove it from the buffer altogether  (so `else { buffer….remove(position); }`).

Can you also add a test for this, please?

crates/matrix-sdk/src/latest_events/latest_event.rs
```diff
@@ -150,6 +169,332 @@ impl LatestEventValue {
...
+    /// # Panics
+    ///
+    /// Panics if `position` is strictly greater than buffer's length.
+    fn remove(&mut self, position: usize) -> (OwnedTransactionId, LatestEventValue) {
+        self.buffer.remove(position)
+    }
+
+    /// Mark the `LatestEventValue` matching `transaction_id`, and all the
+    /// following values, as wedged.
+    fn wedged_from(&mut self, transaction_id: &TransactionId) {
```
bnjbvr: It's weird to have an effectful method with a name that's in the passive form. Consider renaming `mark_wedged_from`?

crates/matrix-sdk/src/latest_events/latest_event.rs
```diff
@@ -150,6 +169,332 @@ impl LatestEventValue {
...
+                if let LatestEventValue::LocalIsSending(kind) = value_to_wedge {
+                    *value_to_wedge = LatestEventValue::LocalIsWedged(kind.clone());
+                }
+            }
+        }
+    }
+
+    /// Mark the `LatestEventValue` matching `transaction_id`, and all the
+    /// following values, as unwedged.
+    fn unwedged_from(&mut self, transaction_id: &TransactionId) {
```
bnjbvr: Same here, `mark_unwedged_from`?

crates/matrix-sdk/src/latest_events/latest_event.rs
```diff
@@ -150,6 +169,332 @@ impl LatestEventValue {
...
+/// 2. [`RoomSendQueueUpdate::NewLocalEvent`]: new local event,
+/// 3. [`RoomSendQueueUpdate::SendError`]: the first local event has failed to
+///    be sent.
+///
+/// Because a `SendError` is received (targeting the first `NewLocalEvent`), the
+/// send queue is stopped. However, the `LatestEventValue` targets the second
+/// `NewLocalEvent`. The system must consider that when a local event is wedged,
+/// all the following local events must also be marked as wedged. And vice
+/// versa, when the send queue is able to send an event again, all the following
+/// local events must be marked as unwedged.
```
bnjbvr: :hammer: I think making an online algorithm updating the latest event based on local echoes is a fine intent, but that it would be much simpler to recompute the latest event, from the list of local echoes (the first `Vec<LocalEcho>` member of the pair returned by `SendQueue::subscribe())` based on a send queue update ping. After all, these updates shouldn't happen that often, so it should be cheap to trigger them; and it would avoid the duplication of each room's send queue in memory, as well as the complexity of reacting to send queue room updates, maintaining wedged/unwedged state, etc.

crates/matrix-sdk/src/latest_events/latest_event.rs
```diff
@@ -150,6 +169,332 @@ impl LatestEventValue {
...
+            }
+        }
+    }
+
+    /// Mark all the following values after the `LatestEventValue` matching
+    /// `transaction_id` as unwedged.
+    ///
+    /// Note that contrary to [`Self::unwedged_from`], the `LatestEventValue` is
+    /// untouched. However, its position is returned (if any).
+    fn unwedged_after(&mut self, transaction_id: &TransactionId) -> Option<usize> {
```
bnjbvr: Maybe `mark_unwedged_after`?

crates/matrix-sdk/src/send_queue/mod.rs
```diff
@@ -236,7 +236,20 @@ impl SendQueue {
...
+
+    /// Get or create a new send queue for a given room, and insert it into our
+    /// memoized rooms mapping.
+    #[cfg(test)]
+    pub(crate) fn for_room(&self, room: Room) -> RoomSendQueue {
+        self.for_room_impl(room)
+    }
+
+    #[inline(always)]
+    fn for_room_impl(&self, room: Room) -> RoomSendQueue {
```
bnjbvr: Oof, it's fine to mark the initial `for_room` as `pub(crate)` IMO, especially if it's only for use in testing. It confused me and took me 15 seconds to find the difference between the two impls here…

crates/matrix-sdk/src/latest_events/mod.rs
Hywan: `tokio::select` automatically fuse the streams, while here I prefer to control the fusing manually for clarity. If you prefer to rely on `tokio::select`, I'm fine switching.

crates/matrix-sdk/src/latest_events/mod.rs
Hywan: You're right! How did I forget about this… I've updated the `error!` to `warn!`.

crates/matrix-sdk/src/latest_events/latest_event.rs
Hywan: I like the new name, and it makes sense. Changed!

crates/matrix-sdk/src/latest_events/latest_event.rs
Hywan: Done, thanks!

crates/matrix-sdk/src/latest_events/latest_event.rs
Hywan: You're right. I've renamed to `extract_content_from_any_message_like`.

crates/matrix-sdk/src/latest_events/mod.rs
Hywan: Haha ^^.

crates/matrix-sdk/src/latest_events/mod.rs
Hywan: Yes! I wanted to keep it short this time, because it's the user ID associated to the room power levels… I can rename if you prefer?

crates/matrix-sdk/src/latest_events/latest_event.rs
Hywan: Yup, sure. I've done that for all `deserialize()` calls.

crates/matrix-sdk/src/latest_events/latest_event.rs
Hywan: Haha, I've spotted that while addressing the precedent feedback.

Sure, on it.

crates/matrix-sdk/src/send_queue/mod.rs
Hywan: 👍

crates/matrix-sdk/src/latest_events/latest_event.rs
Hywan: Done.

crates/matrix-sdk/src/latest_events/latest_event.rs
Hywan: I understand your intent, but I've argument against this proposal.

First off, we are not duplicating each room's send queue in memory because not all events in the send queue have a corresponding `LatestEventValue`. Media and reactions are an example of that.

Second, we are using the _generic_ update, which you call _ping_. This update already contains a copy of the update. If, once we get a _generic_ update, we call `RoomSendQueue::subscribe` to get all the updates, we will copy of all updates including one update copied twice.

Third, the _wedged_/_unwedged_ state in `latest_events` is different from `send_queue` sadly. If the event A, B and C are sent, the latest event is C; if A is wedged, then C should be marked as wedged too. The semantics is different. (As we have talked about today in private, it would be better to even use a different terminology).

These reasons, to me, justify that adopting a different approach won't reduce memory usage and won't simplify the code 🙂.

PR #5483, opened 2025-08-04
refactor(threads): add support for MSC4306 latest changes and push rules

bnjbvr:
- Updates to the new subscription request, that includes the ID of the event that triggered an automatic subscription.
- Get rid of the "unsubscribed" thread status. Technically a breaking change, but since this is all very experimental and shouldn't have shipped yet, I will likely not do a DB migration for it.
- Updates to Ruma's async evaluation of push rules, that's going to be needed to evaluate the thread subscription push rule
- Fetch the thread subscription status when evaluating thread-related event push conditions.

Part of #4869. Depends on https://github.com/ruma/ruma/pull/2180

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4422,7 +4422,7 @@ dependencies = [
 [[package]]
 name = "ruma"
 version = "0.12.6"
-source = "git+https://github.com/ruma/ruma?rev=a626e6b8521bcaa01bf8b7c4161b26b361b72aff#a626e6b8521bcaa01bf8b7c4161b26b361b72aff"
+source = "git+https://github.com/ruma/ruma?rev=e73f302e4df7f5f0511fca1aa43853d4cf8416c8#e73f302e4df7f5f0511fca1aa43853d4cf8416c8"
 dependencies = [
  "assign",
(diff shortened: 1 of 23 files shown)


bnjbvr:
For what it's worth, I'm going to add one commit, so that the evaluation of the new push rule doesn't happen by default, and must be opt-in.

bnjbvr:
@andybalaam This should be ready for review again, please take a look :pray: (and forward to somebody else)

andybalaam:
> @andybalaam This should be ready for review again, please take a look 🙏 (and forward to somebody else)

I got github's "Refresh" button and used it :-)

Why forward to somebody else? Shouldn't github assign the right number of reviewers?

bnjbvr:
Thanks a bunch @andybalaam, weird timing in my comment indeed :grin: Github's UI didn't show yours.

> Why forward to somebody else? Shouldn't github assign the right number of reviewers?

My "and" was supposed to be an "or", doh!

I'll be waiting for the merge in Ruma, and then I'll use a definite (non-PR) Ruma commit and merge this. Thanks for the quick review again!

Issue #5485, opened 2025-08-05
No more notifications for @room mentions


manuroe:
With the recent changes for room v12, the SDK does not trigger anymore a notification for such content:
```json
"content": {
  "body": "@room Big boom!",
  "m.mentions": {
    "room": true
  },
  "msgtype": "m.text"
}
```

The discussion is happening in [#matrix-rust-sdk-dev:flipdot.org](https://matrix.to/#/!qSsPTKDfMGYqhgiLPJ:flipdot.org/$E59WRL4BGigD0IV9WDJ5nWz3Q5QgGqbf8eqZbxElQoo?via=flipdot.org&via=matrix.org&via=element.io).

PR #5486, opened 2025-08-05
fix(notification_client): Add `m.room.create` to the required sliding sync state

jmartinesp:
With this missing in the `SlidingSync` instance created for the notification client, power levels couldn't be computed, and some push rules didn't take effect for notifications.

Should fix https://github.com/matrix-org/matrix-rust-sdk/issues/5485 (https://github.com/element-hq/element-x-android/issues/5113 being the original report on the EXA repo) and the equivalent issue on iOS of `@room` and call notifications being filtered out because they didn't match any push rules.

Kudos to @zecakeh for his help on debugging the issue.

Signed-off-by:


poljar (approved):
Looks reasonable.

Issue #3557, opened 2024-06-14
Error: invalid type: newtype struct, expected any valid JSON value at line 1 column 1061


CyberShadow:
I wrote a Matrix bot using this library a while ago. Today, it won't start.

On startup, it just prints:

> `Error: invalid type: newtype struct, expected any valid JSON value at line 1 column 1061`

and exits.

I'm guessing some file got corrupted. But how can I find which?

My usual debugging methods failed me.

- Searching for the error message on the Internet did not provide any insight.
- `RUST_BACKTRACE=1` has no effect.
- `RUST_BACKTRACE=full` has no effect.
- Running under gdb has no effect.
- Breakpointing `write` did allow me to get a backtrace of what is printing that error message, but it's useless:

```
Thread 1 "matrix_bot" hit Breakpoint 1, 0x00007ffff751a170 in write () from /usr/lib/libc.so.6
(gdb) where
#0  0x00007ffff751a170 in write () from /usr/lib/libc.so.6
#1  0x0000555557d29ebd in std::sys::pal::unix::fd::FileDesc::write () at library/std/src/sys/pal/unix/fd.rs:264
#2  std::sys::pal::unix::stdio::{impl#5}::write () at library/std/src/sys/pal/unix/stdio.rs:71
#3  std::io::Write::write_all<std::sys::pal::unix::stdio::Stderr> () at library/std/src/io/mod.rs:1706
#4  std::io::stdio::{impl#2}::write_all () at library/std/src/io/stdio.rs:176
#5  0x0000555557d2bd59 in std::io::stdio::{impl#21}::write_all () at library/std/src/io/stdio.rs:976
#6  std::io::Write::write_fmt::{impl#0}::write_str<std::io::stdio::StderrLock> () at library/std/src/io/mod.rs:1835
#7  0x0000555557d5873c in core::fmt::write () at library/core/src/fmt/mod.rs:1118
#8  0x0000555557d2a67b in std::io::Write::write_fmt<std::io::stdio::StderrLock> () at library/std/src/io/mod.rs:1846
#9  std::io::stdio::{impl#20}::write_fmt () at library/std/src/io/stdio.rs:956
#10 0x0000555557d2aab2 in std::io::stdio::{impl#19}::write_fmt () at library/std/src/io/stdio.rs:930
#11 std::io::stdio::attempt_print_to_stderr () at library/std/src/io/stdio.rs:1057
[...]

PR #5487, opened 2025-08-05
Upgrade Ruma

zecakeh:
This brings 2 important bug fixes:

- Make deprecated fields of `m.room.encrypted` optional: it seems that there are events without these fields in the wild.
- Fix deserialization of `RedactedRoomJoinRulesEventContent`. This was found by a bug report in Fractal that caused the same error as #3557 when restoring the client. So maybe we could consider that this bug is fixed? It is still possible that there is another deserialization error. 

There is also a breaking change in the format of the `state` field in response to `GET /v3/sync`.

### DIFF
--- a/crates/matrix-sdk-base/src/response_processors/room/sync_v2.rs
+++ b/crates/matrix-sdk-base/src/response_processors/room/sync_v2.rs
@@ -16,9 +16,10 @@ use std::collections::{BTreeMap, BTreeSet};
 
 use ruma::{
     OwnedRoomId, OwnedUserId, RoomId,
-    api::client::sync::sync_events::v3::{InvitedRoom, JoinedRoom, KnockedRoom, LeftRoom},
+    api::client::sync::sync_events::v3::{InvitedRoom, JoinedRoom, KnockedRoom, LeftRoom, State},
 };
 use tokio::sync::broadcast::Sender;
+use tracing::error;
 
 #[cfg(feature = "e2e-encryption")]
 use super::super::e2ee;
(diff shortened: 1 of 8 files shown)


crates/matrix-sdk-base/src/response_processors/room/sync_v2.rs
```diff
@@ -61,39 +62,48 @@ pub async fn update_joined_room(
...
+                state_events::sync::collect_from_timeline(&joined_room.timeline.events);
+            state_events::sync::dispatch(
+                context,
+                (&raw_state_events_from_timeline, &state_events_from_timeline),
+                &mut room_info,
+                ambiguity_cache,
+                &mut new_user_ids,
+                state_store,
+            )
+            .await?;
```
bnjbvr, line 90: Out of curiosity: could this part be *after* the `match`, if each match arm returned `raw_state_events` + `state_events`?

crates/matrix-sdk-base/src/response_processors/room/sync_v2.rs
```diff
@@ -179,30 +189,39 @@ pub async fn update_left_room(
...
+                state_events::sync::collect_from_timeline(&left_room.timeline.events);
+            state_events::sync::dispatch(
+                context,
+                (&raw_state_events_from_timeline, &state_events_from_timeline),
+                &mut room_info,
+                ambiguity_cache,
+                &mut (),
+                state_store,
+            )
+            .await?;
```
bnjbvr, line 215: and ditto here

bnjbvr (commented):
Looks good to me; one question to rearrange the code and process timeline items even when we don't know about the state kind.

crates/matrix-sdk-base/src/response_processors/room/sync_v2.rs
zecakeh: No, because the other variant is `State::After` (I will add support for that in another PR), where we **don't** process the state events in the timeline (like for SSS).

zecakeh: The behavior to apply is detailed in the docs of the variant, fwiw.

bnjbvr (approved):
Alright, thanks for the explanation!

PR #5488, opened 2025-08-05
feat(sdk): Use `state_after` in sync v2 

zecakeh:
It is supposed to be an improvement over `state`, since it allows the
server to send updates to the state that might not be reflected in the
timeline.

This is also the same behavior as in Simplified Sliding Sync.

This is MSC4222 that was accepted and is about to get merged in the spec.

### DIFF
--- a/crates/matrix-sdk-base/src/sync.rs
+++ b/crates/matrix-sdk-base/src/sync.rs
@@ -180,10 +180,10 @@ pub struct JoinedRoomUpdate {
     /// The timeline of messages and state changes in the room.
     pub timeline: Timeline,
     /// Updates to the state, between the time indicated by the `since`
-    /// parameter, and the start of the `timeline` (or all state up to the
-    /// start of the `timeline`, if `since` is not given, or `full_state` is
-    /// true).
-    pub state: Vec<Raw<AnySyncStateEvent>>,
+    /// parameter, and the start or the end of the `timeline` (or all state up
+    /// to the start or the end of the `timeline`, if `since` is not given,
+    /// or `full_state` is true).
+    pub state: State,
     /// The private data that this user has attached to this room.
     pub account_data: Vec<Raw<AnyRoomAccountDataEvent>>,
     /// The ephemeral events in the room that aren't recorded in the timeline or
@@ -202,7 +202,7 @@ impl fmt::Debug for JoinedRoomUpdate {
         f.debug_struct("JoinedRoomUpdate")
             .field("unread_notifications", &self.unread_notifications)
             .field("timeline", &self.timeline)
-            .field("state", &DebugListOfRawEvents(&self.state))
+            .field("state", &self.state)
             .field("account_data", &DebugListOfRawEventsNoId(&self.account_data))
             .field("ephemeral", &self.ephemeral)
             .field("ambiguity_changes", &self.ambiguity_changes)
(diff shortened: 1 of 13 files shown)


crates/matrix-sdk-base/src/sync.rs
```diff
@@ -180,10 +180,10 @@ pub struct JoinedRoomUpdate {
     /// The timeline of messages and state changes in the room.
     pub timeline: Timeline,
     /// Updates to the state, between the time indicated by the `since`
-    /// parameter, and the start of the `timeline` (or all state up to the
-    /// start of the `timeline`, if `since` is not given, or `full_state` is
-    /// true).
-    pub state: Vec<Raw<AnySyncStateEvent>>,
+    /// parameter, and the start or the end of the `timeline` (or all state up
+    /// to the start or the end of the `timeline`, if `since` is not given,
+    /// or `full_state` is true).
```
andybalaam: This comment is becoming very hard to understand! I tried to reword it below but it is difficult! I'm not 100% sure it's correct, so please read it carefully, but maybe something like this?

```
/// Updates to the state.
///
/// If `since` is missing or `full_state` is true, the start point of the update is the
/// beginning of the timeline. Otherwise, the start point is the time specified in `since`.
///
/// If `state_after` was used, the end point of the update is the end of the
/// `timeline`. Otherwise, the end point of these updates is the start of
// the `timeline`, and to calculate room state we must scan the `timeline`
/// for state events as well as using this information in this property.
```

crates/matrix-sdk-base/src/sync.rs
```diff
@@ -249,10 +249,10 @@ pub struct LeftRoomUpdate {
     /// when the user left.
     pub timeline: Timeline,
     /// Updates to the state, between the time indicated by the `since`
-    /// parameter, and the start of the `timeline` (or all state up to the
-    /// start of the `timeline`, if `since` is not given, or `full_state` is
-    /// true).
-    pub state: Vec<Raw<AnySyncStateEvent>>,
+    /// parameter, and the start or the end of the `timeline` (or all state up
+    /// to the start or the end of the `timeline`, if `since` is not given, or
+    /// `full_state` is true).
```
andybalaam: I guess re-use whatever comment we choose here too.

crates/matrix-sdk/src/event_handler/mod.rs
```diff
@@ -422,6 +418,11 @@ impl Client {
             unsigned: Option<UnsignedDetails>,
         }
 
+        let state_events = match state {
+            State::Before(events) => events,
+            State::After(events) => events,
+        };
```
andybalaam, line 424: I guess we don't need to use `collect` here because these events are not being used to calculate room state?

crates/matrix-sdk/src/widget/matrix.rs
```diff
@@ -534,9 +535,13 @@ impl StateUpdateReceiver {
         loop {
             match self.room_updates.recv().await? {
                 RoomUpdate::Joined { room, updates } => {
-                    if !updates.state.is_empty() {
-                        return Ok(updates
-                            .state
+                    let state_events = match updates.state {
+                        State::Before(events) => events,
+                        State::After(events) => events,
+                    };
```
andybalaam, line 541: We are losing the information here about whether these state events cover the whole update or not. Should the widget API take the result of `collect`, so it can always just update state based on what we pass it?

crates/matrix-sdk/src/event_handler/mod.rs
zecakeh: The thing about `collect` is that it filters out events that we fail to deserialize. Other users of the data might be able to deserialize them.

crates/matrix-sdk-base/src/sync.rs
zecakeh: Yes, this looks better. Thanks!

crates/matrix-sdk/src/widget/matrix.rs
zecakeh: I really don't know, I guess that it depends on what the widget is expecting. But the information was already lost before between sync v2 and SSS.

andybalaam (approved):

PR #5489, opened 2025-08-05
refactor(crypto): Make deprecated `sender_key` and `device_id` optional in `RoomEncryptedEventContent` and `RoomKeyRequestContent`

zecakeh:
They were deprecated in Matrix 1.3 and are now optional.

According to a user report in the Ruma room, there seems to be events in the wild without those fields.

### DIFF
--- a/crates/matrix-sdk-crypto/src/types/events/room/encrypted.rs
+++ b/crates/matrix-sdk-crypto/src/types/events/room/encrypted.rs
@@ -23,12 +23,12 @@ use vodozemac::{megolm::MegolmMessage, olm::OlmMessage, Curve25519PublicKey};
 
 use super::Event;
 use crate::types::{
-    deserialize_curve_key, deserialize_curve_key_option,
+    deserialize_curve_key,
     events::{
         room_key_request::{self, SupportedKeyInfo},
         EventType, ToDeviceEvent,
     },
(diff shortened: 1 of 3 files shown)


crates/matrix-sdk-crypto/src/types/events/room/encrypted.rs
```diff
@@ -309,11 +309,17 @@ pub struct MegolmV1AesSha2Content {
     pub ciphertext: MegolmMessage,
 
     /// The Curve25519 key of the sender.
-    #[serde(deserialize_with = "deserialize_curve_key", serialize_with = "serialize_curve_key")]
-    pub sender_key: Curve25519PublicKey,
+    #[serde(
+        default,
+        deserialize_with = "deserialize_curve_key_option",
+        serialize_with = "serialize_curve_key_option",
```
jplatte: Could bundle them in a module and use `#[serde(with)]`.

poljar: Would be nice, but not that important if you don't feel like it.

zecakeh: Done.

poljar (approved):
Still looking good, thanks.

PR #5490, opened 2025-08-05
chore: Upgrade matrix-sdk-common to edition 2024

jplatte:
Signed-off-by: Jonas Platte

### DIFF
--- a/crates/matrix-sdk-common/src/failures_cache.rs
+++ b/crates/matrix-sdk-common/src/failures_cache.rs
@@ -102,10 +102,7 @@ where
         Q: Hash + Eq + ?Sized,
     {
         let lock = self.inner.items.read();
-
-        let contains = if let Some(item) = lock.get(key) { !item.expired() } else { false };
-
-        contains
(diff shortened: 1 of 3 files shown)


crates/matrix-sdk-common/src/failures_cache.rs
```diff
@@ -102,10 +102,7 @@ where
         Q: Hash + Eq + ?Sized,
     {
         let lock = self.inner.items.read();
-
-        let contains = if let Some(item) = lock.get(key) { !item.expired() } else { false };
-
-        contains
+        if let Some(item) = lock.get(key) { !item.expired() } else { false }
```
Hywan, line 105: Honestly, a:

```rust
lock.get(key).map(|item| !item.expired()).unwrap_or_default()
```

would be simpler and more Rust idiomatic here. A detail.

jplatte: Even better: `is_some_and` https://github.com/matrix-org/matrix-rust-sdk/pull/5498

PR #5491, opened 2025-08-05
Introduce TestResult and use it in a couple random places

jplatte:
Let's you use `?` in tests (only on `Result`s, not `Option`s unfortunately). Error backtrace will point to the (first line of the) expression being `?`ed. Pretty close to `.unwrap()`!

Signed-off-by: Jonas Platte


Hywan:
Can you please fix the conflict?

PR #5494, opened 2025-08-06
refactor(notification client): misc simplifications

bnjbvr:
Some simplifications extracted from #5461.

### DIFF
--- a/crates/matrix-sdk-ui/src/notification_client.rs
+++ b/crates/matrix-sdk-ui/src/notification_client.rs
@@ -606,6 +606,34 @@ impl NotificationClient {
         get_notifications_result.remove(event_id).unwrap_or(Ok(NotificationStatus::EventNotFound))
     }
 
+    /// Given a (decrypted or not) event, figure out whether it should be
+    /// filtered out for other client-side reasons (such as the sender being
+    /// ignored, for instance), and returns the corresponding
+    /// [`NotificationStatus`].
+    async fn compute_status(
+        &self,
+        room: &Room,
+        push_actions: Option<&[Action]>,
+        raw_event: RawNotificationEvent,
+        state_events: Vec<Raw<AnyStateEvent>>,
+    ) -> Result<NotificationStatus, Error> {
+        if let Some(actions) = push_actions
+            && !actions.iter().any(|a| a.should_notify())
+        {
+            // The event shouldn't notify: return early.
+            return Ok(NotificationStatus::EventFilteredOut);
+        }
+
+        let notification_item =
+            NotificationItem::new(room, raw_event, push_actions, state_events).await?;
+
+        if self.client.is_user_ignored(notification_item.event.sender()).await {
+            Ok(NotificationStatus::EventFilteredOut)
+        } else {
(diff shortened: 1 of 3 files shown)


crates/matrix-sdk-ui/src/notification_client.rs
```diff
@@ -684,30 +684,23 @@ impl NotificationClient {
...
-                    }
-                }
-                _ => {
-                    batch_result.insert(event_id, Ok(status));
-                }
+            if self.client.is_user_ignored(notification_item.event.sender()).await {
+                batch_result.insert(event_id, Ok(NotificationStatus::EventFilteredOut));
+            } else {
+                batch_result
+                    .insert(event_id, Ok(NotificationStatus::Event(Box::new(notification_item))));
```
Hywan: I suspect the modification might be incorrect.

The previous version of the code was:

```rust
match status {
    NotificationStatus::Event(event) => { … }
    _ => {
        batch_result.insert(event_id, Ok(status));
    }
}
```

The new version is:

```rust
if … { … }
else {
    batch_result.insert(event_id, Ok(NotificationStatus::Event(…));
}
```

It means that the semantics of the `else` case (for both the `match` in the previous version, or the `else` in the new version) aren't equivalent. In the former, we are inserting a status that is **not** `NotificationStatus::Event`, whilst in the other, we are always inserting a status of type `NotificationStatus::Event`.

Did I miss something? I'm not familiar with this code base.

crates/matrix-sdk-ui/src/notification_client.rs
```diff
@@ -606,6 +606,34 @@ impl NotificationClient {
         get_notifications_result.remove(event_id).unwrap_or(Ok(NotificationStatus::EventNotFound))
     }
 
+    /// Given a (decrypted or not) event, figure out whether it should be
+    /// filtered out for other client-side reasons (such as the sender being
+    /// ignored, for instance), and returns the corresponding
+    /// [`NotificationStatus`].
+    async fn compute_status(
```
Hywan, line 613: This commit might be impacted by my previous comment.

crates/matrix-sdk-ui/src/notification_client.rs
bnjbvr: If you read the previous code:

```rust
let status =
                match NotificationItem::new(&room, raw_event, push_actions.as_deref(), Vec::new())
                    .await
                    .map(|event| NotificationStatus::Event(Box::new(event))) {
 Ok(status) => status,
 Err(err) => /* record the error and continue */
};

match status {
  NotificationStatus::Event(event) => do_some(),
  _ => {
   do_something_else();
  }
}
```

the `do_something_else()` arm is actually unreachable, because `status`, if we've reached that branch, is always a `NotificationStatus::Event`, because of the `.map()` call above.

The code was confusing, since it included bits that couldn't be reached. The new version avoids that confusion by removing this unreachable code path.

Hywan: Oh okay, I see. Thanks for the explanations!

Issue #5139, opened 2025-05-30
History sharing: check `sender_data` before downloading room key bundles


richvdh:
Currently, we do not check the cross-signing status for the device that sends us room key bundles. This could mean that we accept a room key bundle even if the device is signed by the **incorrect** cross-signing key. We should make sure we check the cross-signing status and only accept key bundles from devices that are cross-signed.

Part of https://github.com/element-hq/element-meta/issues/2829

richvdh:
There is a risk of cache-poisoning here. What happens if:
 * Alice sends information about key bundle to Bob, and sends the invite
 * Mallory sends information about key bundle to Bob, posing as Alice
 * Bob accepts the invite

Since we only store the most recent key bundle data from each user (IIRC?), we will overwrite the genuine info from Alice with the false data from Mallory.

However: provided we actually check the sender device for whatever bundle we use is correctly cross-signed, I think that reduces this to a DoS (in which Bob doesn't get the keys), rather than anything more serious. And, I venture to suggest, that if Mallory can pose as Alice, she probably has other ways of causing a DoS.

PR #5495, opened 2025-08-06
Remember the public Curve25519 key of the sender of the historic room key bundle

poljar:
Part of https://github.com/matrix-org/matrix-rust-sdk/issues/5139.


crates/matrix-sdk-crypto/src/machine/mod.rs
```diff
@@ -1287,7 +1289,7 @@ impl OlmMachine {
             }
             AnyDecryptedOlmEvent::RoomKeyBundle(e) => {
                 debug!("Received a room key bundle event {:?}", e);
```
bnjbvr, line 1291: pre-existing, but is it fine to log a full key bundle event? (no keys will appear in there?)

poljar: Good thinking https://github.com/ruma/ruma/pull/2185.

PR #5496, opened 2025-08-06
Log a special error when try to upload duplicate one-time keys

poljar:


crates/matrix-sdk/src/encryption/mod.rs
```diff
@@ -622,7 +623,26 @@ impl Client {
...
-                self.keys_upload(r.request_id(), request).await?;
+                self.keys_upload(r.request_id(), request).await.inspect_err(|e| {
+                    match e.as_ruma_api_error() {
+                        Some(RumaApiError::ClientApi(e)) if e.status_code == 400 => {
+                            if let ErrorBody::Standard { message, .. } = &e.body {
+                                // This is one of the nastiest errors we can have. The server
+                                // telling us that we already have a one-time key uploaded means
+                                // that we forgot about some of our one-time keys. This will lead to
+                                // UTDs.
+                                if message.starts_with("One time key") {
```
stefanceriu, line 634: Is this by any chance synapse specific? Would it maybe be better to check for the request's `one_time_keys` list instead? 🤔

crates/matrix-sdk/src/encryption/mod.rs
```diff
@@ -622,7 +623,26 @@ impl Client {
                 self.keys_query(r.request_id(), request.device_keys.clone()).await?;
             }
             AnyOutgoingRequest::KeysUpload(request) => {
-                self.keys_upload(r.request_id(), request).await?;
+                self.keys_upload(r.request_id(), request).await.inspect_err(|e| {
+                    match e.as_ruma_api_error() {
+                        Some(RumaApiError::ClientApi(e)) if e.status_code == 400 => {
```
stefanceriu, line 628: Are these response codes defined anywhere in the spec cause I can't seem to find them?

crates/matrix-sdk/src/encryption/mod.rs
poljar: Yes, I think this is synapse specific. What do you mean by checking the request's list?

This error happens if the list of one-time keys we have locally diverges from the list of one-time keys the server knows about.

I.e. we have something called a key ID, which is actually a counter. So we upload keys with a key ID from 1-10 to the server. The next time we try to upload keys with a key ID from 1-10 the server will tell us that we already uploaded those keys.

crates/matrix-sdk/src/encryption/mod.rs
poljar: I don't think they are, they seem to be Synapse specific. But we saw this error in logs:

> 2025-07-07T14:04:22.936320Z DEBUG matrix_sdk::http_client: Error while sending request: Api(Server(ClientApi(Error { status_code: 400, body: Standard { kind: Unknown, message: "One time key signed_curve25519:AAAAAAAAAUM already exists. Old key: {\"key\":\"qPPCHIFnIQC9Sf87EDC3ePHsx9K5igXARQh8SnZESWs\",\"signatures\":{\"xxxxx:xxxxx.org\":{\"ed25519:sPojQN8l8QrkRI4Dnus+aJS7e/k3kUeRY177rdrlaAw\":\"vi/xBcNqlvbxvn8xI8lO5Qj+Nzf5G5xK8JP2QwqeylEiRPk9FMdxwIngN5duuuWdWJyDg6/f/jjJ5jolP/+uDw\"}}}; new key: {'key': 'WDaBVLDrQJVTAtY1H8kHP06q9Mkjqpp9nmWQuNbNrFo', 'signatures': {'xxxxx:xxxxx.org': {'ed25519:sPojQN8l8QrkRI4Dnus+aJS7e/k3kUeRY177rdrlaAw': '1GScTpjsnc0eLMyi818g/ZxokKCbATpWaXfud6JaQirzRmQ8t58932gPXH0q1OGRWl7xK6SrudAlbli4Mq7qDA'}}}" } })))

crates/matrix-sdk/src/encryption/mod.rs
stefanceriu: Just meant to only check requests that contain one time keys in their payload, we don't care about errors on any of the others, but as long as we check the string it doesn't really matter, does it.

stefanceriu (approved):
Looks sensible to me 👍

crates/matrix-sdk/src/encryption/mod.rs
poljar: Sorry, still not sure what you mean here. :sweat_smile:

I'll try to answer anyways. 

We are only logging this for `/keys/upload` requests, the first `/keys/upload` request might not contain any `one-time` keys but every subsequent request will.

So it's not particularly useful to check if the request contained one-time keys, the first request a client ever does by definition can't fail with this error either.

stefanceriu: > but every subsequent request will.

Oh, that was the bit of info I was missing. I was under the impression one time keys don't get used and replenished that often.

poljar: That depends on how many new devices want to talk to you, but this requests purpose is to replenish and upload new one-time keys to the server.

PR #5501, opened 2025-08-07
refactor(base): Don't take room ID in `Notification::push_notification_from_event_if`

zecakeh:
It is already in the `PushConditionRoomCtx`, so we don't need an extra argument for it.


bnjbvr (approved):
Sweet!

zecakeh:
Rebased on main to fix CI.

PR #5502, opened 2025-08-07
Bump rand

poljar:
Ok, the rand bump in vodozemac will take a bit longer due to it having many more dependencies that require a `rand` bump as well.

In the meantime we can bump `rand` in the SDK.

### DIFF
--- a/crates/matrix-sdk-common/Cargo.toml
+++ b/crates/matrix-sdk-common/Cargo.toml
@@ -66,6 +66,7 @@ wasm-bindgen-test.workspace = true
 [target.'cfg(target_family = "wasm")'.dev-dependencies]
 # Enable the JS feature for getrandom.
 getrandom = { workspace = true, default-features = false, features = ["wasm_js"] }
+getrandom3 = { version = "0.3.4", package = "getrandom", default-features = false, features = ["wasm_js"] }
 js-sys.workspace = true
 
 [lints]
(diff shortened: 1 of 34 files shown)


andybalaam (approved):

poljar:
@andybalaam do you want to retake your look here?

crates/matrix-sdk-common/Cargo.toml
```diff
@@ -66,6 +66,7 @@ wasm-bindgen-test.workspace = true
 [target.'cfg(target_family = "wasm")'.dev-dependencies]
 # Enable the JS feature for getrandom.
 getrandom = { workspace = true, default-features = false, features = ["wasm_js"] }
+getrandom2 = { version = "0.3.4", package = "getrandom", default-features = false, features = ["wasm_js"] }
```
andybalaam: Is this needed? (Just wondering because I don't see it when we are not in the wasm family)

poljar: It is as some of our dependencies still use 0.3.x. Not enabling it leads to compilation issues https://github.com/matrix-org/matrix-rust-sdk/actions/runs/23748646788/job/69184209001?pr=5502.

andybalaam: Fine, thanks.

Issue #5350, opened 2025-07-04
Meta: Full text search support


poljar:
# Introduction

As users post messages and the room history grows over time, manually scrolling to find a past conversation becomes inefficient and frustrating.

In today's world, users expect to be able to find past conversations easily and quickly. Adding full text search would greatly improve usability by allowing users to search across messages with filters like sender, date ranges, and on a per-room basis.

Element Desktop has been supporting full text search for quite a while now. It utilizes [Seshat](https://github.com/matrix-org/seshat) which uses a SQLite database in combination with a full-text search index (backed by [Tantivy](https://github.com/quickwit-oss/tantivy/)) to provide a simple API to feed events into the database/index and search and retrieve the events.

# Plan

The Rust SDK already contains a persistent store for events we encounter. This fact and Seshat itself being stuck on a quite old version of the used full-text search library makes direct adoption of Seshat unfeasible.

Nevertheless, some parts of Seshat will be useful, for example Seshat contains a storage backend for `Tantivy` which encrypts the index.

This issue lies down the tasks that would be necessary to bring full-text search support into the Rust SDK.

Because we can't adopt Seshat directly, we will begin with the creation of a new crate in this very repository. After that we can add a simple API to index events and search the index.

Once a functioning API exists, we can experiment with the search functionality using benchmarks, small test clients and finally multiverse.

# MVP tasks for internship

The following list of tasks lays out the rough plan:

- [x] Create a new create called matrix-sdk-search.
- [x] Implement a basic API to add events to and index using. [Tantivy](https://github.com/quickwit-oss/tantivy/)
- [x] Add support to search for events.
- [x] Extend multiverse so we can test out the search API and functionality.
- [x] Utilize the matrix-sdk-search crate in the `EventCache`, every time the `EventCache` inserts an event, index the event.
- [x] Add support for edits and redactions
- [x] Test out and benchmark if a per-room index is better than a global index.
- [x] Pagination
- [x] Implement room spidering so we index the whole room history.
    - [x] Fix search on first viewing
- [x] Support encrypted index. 
- [x] Update [README](https://github.com/matrix-org/matrix-rust-sdk/blob/main/crates/matrix-sdk-search/README.md) with information and code examples.
- [x] Record a demo of search in Multiverse.

### Nice to have

- [ ] Global search.
- [ ] Background spidering
- [x] Try bulk operations to reduce index commits.
- [ ] Make index writes/commits non-blocking.
[...]

PR #5505, opened 2025-08-07
feat(sdk): Add creation of indexes and indexing of messages

dragonfly1033:
Integrate matrix-sdk-search into matrix-sdk.
When a room is joined, a corresponding index is created.
When a message is received via sync or via a back-pagination, it is added to the corresponding room's index.

### DIFF
--- a/crates/matrix-sdk-search/src/index.rs
+++ b/crates/matrix-sdk-search/src/index.rs
@@ -12,22 +12,9 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-//! The event cache is an abstraction layer, sitting between the Rust SDK and a
-//! final client, that acts as a global observer of all the rooms, gathering and
-//! inferring some extra useful information about each room. In particular, this
-//! doesn't require subscribing to a specific room to get access to this
-//! information.
-//!
-//! It's intended to be fast, robust and easy to maintain, having learned from
-//! previous endeavours at implementing middle to high level features elsewhere
-//! in the SDK, notably in the UI's Timeline object.
-//!
-//! See the [github issue](https://github.com/matrix-org/matrix-rust-sdk/issues/3058) for more
-//! details about the historical reasons that led us to start writing this.
-
 use std::{fmt, fs, path::Path, sync::Arc};
 
-use ruma::{OwnedEventId, OwnedRoomId, RoomId, events::AnyMessageLikeEvent};
+use ruma::{OwnedEventId, OwnedRoomId, RoomId, events::AnySyncMessageLikeEvent};
 use tantivy::{
     Index, IndexReader, TantivyDocument,
     collector::TopDocs,
@@ -43,6 +30,11 @@ use crate::{
     schema::{MatrixSearchIndexSchema, RoomMessageSchema},
     writer::SearchIndexWriter,
 };
+
+/// A struct to represent the operations on a [`RoomIndex`]
+pub(crate) enum RoomIndexOperation {
+    Add(TantivyDocument),
+}
 
 /// A struct that holds all data pertaining to a particular room's
 /// message index.
@@ -91,9 +83,9 @@ impl RoomIndex {
         RoomIndex::new_with(index, schema, room_id)
     }
 
-    /// Create new [`RoomIndex`] which stores the index in RAM.
+    /// Create new [`RoomIndex`] which stores the index in memory.
     /// Intended for testing.
-    pub fn new_in_ram(room_id: &RoomId) -> Result<RoomIndex, IndexError> {
+    pub fn new_in_memory(room_id: &RoomId) -> Result<RoomIndex, IndexError> {
         let schema = RoomMessageSchema::new();
         let index = Index::create_in_ram(schema.as_tantivy_schema());
         RoomIndex::new_with(index, schema, room_id)
@@ -130,10 +122,14 @@ impl RoomIndex {
         RoomIndex::new_with(index, schema, room_id)
     }
 
-    /// Add [`AnyMessageLikeEvent`] to [`RoomIndex`]
-    pub fn add_event(&mut self, event: AnyMessageLikeEvent) -> Result<(), IndexError> {
-        let doc = self.schema.make_doc(event)?;
-        self.writer.add_document(doc)?; // TODO: This is blocking. Handle it.
+    /// Handle [`AnySyncMessageLikeEvent`]
+    ///
+    /// This which will add/remove/edit an event in the index based on the
+    /// event type.
+    pub fn handle_event(&mut self, event: AnySyncMessageLikeEvent) -> Result<(), IndexError> {
+        match self.schema.handle_event(event)? {
+            RoomIndexOperation::Add(document) => self.writer.add_document(document)?,
+        };
         Ok(())
     }
 
@@ -198,64 +194,64 @@ mod tests {
     use crate::index::RoomIndex;
 
     #[test]
-    fn test_make_index_in_ram() {
+    fn test_make_index_in_memory() {
         let room_id = room_id!("!room_id:localhost");
-        let index = RoomIndex::new_in_ram(room_id);
+        let index = RoomIndex::new_in_memory(room_id);
 
         index.expect("failed to make index in ram: {index:?}");
     }
 
     #[test]
-    fn test_add_event() {
+    fn test_handle_event() {
         let room_id = room_id!("!room_id:localhost");
         let mut index =
-            RoomIndex::new_in_ram(room_id).expect("failed to make index in ram: {index:?}");
+            RoomIndex::new_in_memory(room_id).expect("failed to make index in ram: {index:?}");
 
         let event = EventFactory::new()
             .text_msg("event message")
             .event_id(event_id!("$event_id:localhost"))
             .room(room_id)
             .sender(user_id!("@user_id:localhost"))
-            .into_any_message_like_event();
+            .into_any_sync_message_like_event();
 
-        index.add_event(event).expect("failed to add event: {res:?}");
+        index.handle_event(event).expect("failed to add event: {res:?}");
     }
 
     #[test]
     fn test_search_populated_index() -> Result<(), Box<dyn Error>> {
         let room_id = room_id!("!room_id:localhost");
         let mut index =
-            RoomIndex::new_in_ram(room_id).expect("failed to make index in ram: {index:?}");
+            RoomIndex::new_in_memory(room_id).expect("failed to make index in ram: {index:?}");
 
         let event_id_1 = event_id!("$event_id_1:localhost");
         let event_id_2 = event_id!("$event_id_2:localhost");
         let event_id_3 = event_id!("$event_id_3:localhost");
 
-        index.add_event(
+        index.handle_event(
             EventFactory::new()
                 .text_msg("This is a sentence")
                 .event_id(event_id_1)
                 .room(room_id)
                 .sender(user_id!("@user_id:localhost"))
-                .into_any_message_like_event(),
+                .into_any_sync_message_like_event(),
         )?;
 
-        index.add_event(
+        index.handle_event(
             EventFactory::new()
                 .text_msg("All new words")
                 .event_id(event_id_2)
                 .room(room_id)
                 .sender(user_id!("@user_id:localhost"))
-                .into_any_message_like_event(),
+                .into_any_sync_message_like_event(),
         )?;
 
-        index.add_event(
+        index.handle_event(
             EventFactory::new()
                 .text_msg("A similar sentence")
                 .event_id(event_id_3)
                 .room(room_id)
                 .sender(user_id!("@user_id:localhost"))
-                .into_any_message_like_event(),
+                .into_any_sync_message_like_event(),
         )?;
 
         index.commit_and_reload()?;
@@ -275,7 +271,7 @@ mod tests {
     fn test_search_empty_index() -> Result<(), Box<dyn Error>> {
         let room_id = room_id!("!room_id:localhost");
         let mut index =
-            RoomIndex::new_in_ram(room_id).expect("failed to make index in ram: {index:?}");
+            RoomIndex::new_in_memory(room_id).expect("failed to make index in ram: {index:?}");
 
         index.commit_and_reload()?;
 
--- a/crates/matrix-sdk-search/src/schema.rs
+++ b/crates/matrix-sdk-search/src/schema.rs
@@ -12,39 +12,29 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-//! The event cache is an abstraction layer, sitting between the Rust SDK and a
-//! final client, that acts as a global observer of all the rooms, gathering and
-//! inferring some extra useful information about each room. In particular, this
-//! doesn't require subscribing to a specific room to get access to this
-//! information.
-//!
-//! It's intended to be fast, robust and easy to maintain, having learned from
-//! previous endeavours at implementing middle to high level features elsewhere
-//! in the SDK, notably in the UI's Timeline object.
-//!
-//! See the [github issue](https://github.com/matrix-org/matrix-rust-sdk/issues/3058) for more
-//! details about the historical reasons that led us to start writing this.
-
-use ruma::{
-    MilliSecondsSinceUnixEpoch, OwnedEventId, OwnedUserId,
-    events::{
-        AnyMessageLikeEvent, MessageLikeEvent, MessageLikeEventContent, RedactContent,
-        RedactedMessageLikeEventContent, room::message::MessageType,
-    },
+use ruma::events::{
+    AnySyncMessageLikeEvent, MessageLikeEventContent, RedactContent,
+    RedactedMessageLikeEventContent, SyncMessageLikeEvent, room::message::MessageType,
 };
 use tantivy::{
     DateTime, TantivyDocument, doc,
     schema::{DateOptions, DateTimePrecision, Field, INDEXED, STORED, STRING, Schema, TEXT},
 };
 
-use crate::error::{IndexError, IndexSchemaError};
+use crate::{
+    error::{IndexError, IndexSchemaError},
+    index::RoomIndexOperation,
+};
 
 pub(crate) trait MatrixSearchIndexSchema {
     fn new() -> Self;
     fn default_search_fields(&self) -> Vec<Field>;
     fn primary_key(&self) -> Field;
     fn as_tantivy_schema(&self) -> Schema;
-    fn make_doc(&self, event: AnyMessageLikeEvent) -> Result<TantivyDocument, IndexError>;
+    fn handle_event(
+        &self,
+        event: AnySyncMessageLikeEvent,
+    ) -> Result<RoomIndexOperation, IndexError>;
 }
 
 #[derive(Debug, Clone)]
@@ -58,47 +48,30 @@ pub(crate) struct RoomMessageSchema {
 }
 
 impl RoomMessageSchema {
-    fn parse_event<C: MessageLikeEventContent + RedactContent, F>(
+    /// Given an [`AnySyncMessageLikeEvent`] and a function to convert the
+    /// content into a String to be indexed, return a [`TantivyDocument`] to
+    /// index.
+    fn make_doc<C: MessageLikeEventContent + RedactContent, F>(
         &self,
-        event: MessageLikeEvent<C>,
-        get_body: F,
-    ) -> Result<(OwnedEventId, String, MilliSecondsSinceUnixEpoch, OwnedUserId), IndexError>
+        event: SyncMessageLikeEvent<C>,
+        get_body_from_content: F,
+    ) -> Result<TantivyDocument, IndexError>
     where
         <C as RedactContent>::Redacted: RedactedMessageLikeEventContent,
         F: FnOnce(&C) -> Result<String, IndexError>,
     {
         let unredacted = event.as_original().ok_or(IndexError::CannotIndexRedactedMessage)?;
 
-        let body = get_body(&unredacted.content)?;
+        let body = get_body_from_content(&unredacted.content)?;
 
-        Ok((
-            unredacted.event_id.clone(),
-            body,
-            unredacted.origin_server_ts,
-            unredacted.sender.clone(),
+        Ok(doc!(
+            self.event_id_field => unredacted.event_id.to_string(),
+            self.body_field => body,
+            self.date_field =>
+                DateTime::from_timestamp_millis(
+                    unredacted.origin_server_ts.get().into()),
+            self.sender_field => unredacted.sender.to_string(),
         ))
-    }
-
-    fn parse_any_event(
-        &self,
-        event: AnyMessageLikeEvent,
-    ) -> Result<(OwnedEventId, String, MilliSecondsSinceUnixEpoch, OwnedUserId), IndexError> {
-        match event {
-            // old m.room.message behaviour
-            AnyMessageLikeEvent::RoomMessage(event) => {
-                self.parse_event(event, |content| match &content.msgtype {
-                    MessageType::Text(content) => Ok(content.body.clone()),
-                    _ => Err(IndexError::MessageTypeNotSupported),
-                })
-            }
-
-            // new m.message behaviour
-            AnyMessageLikeEvent::Message(event) => self.parse_event(event, |content| {
-                content.text.find_plain().ok_or(IndexError::EmptyMessage).map(|v| v.to_owned())
-            }),
-
-            _ => Err(IndexError::MessageTypeNotSupported),
-        }
     }
 }
 
@@ -140,17 +113,28 @@ impl MatrixSearchIndexSchema for RoomMessageSchema {
         self.inner.clone()
     }
 
-    fn make_doc(&self, event: AnyMessageLikeEvent) -> Result<TantivyDocument, IndexError> {
-        let (event_id, body, timestamp, sender) = self.parse_any_event(event)?;
+    fn handle_event(
+        &self,
+        event: AnySyncMessageLikeEvent,
+    ) -> Result<RoomIndexOperation, IndexError> {
+        match event {
+            // m.room.message behaviour
+            AnySyncMessageLikeEvent::RoomMessage(event) => self
+                .make_doc(event, |content| match &content.msgtype {
+                    MessageType::Text(content) => Ok(content.body.clone()),
+                    _ => Err(IndexError::MessageTypeNotSupported),
+                })
+                .map(RoomIndexOperation::Add),
 
-        Ok(doc!(
-            self.event_id_field => event_id.to_string(),
-            self.body_field => body,
-            self.date_field =>
-                DateTime::from_timestamp_millis(
-                    timestamp.get().into()),
-            self.sender_field => sender.to_string(),
-        ))
+            // new MSC-1767 m.message behaviour
+            AnySyncMessageLikeEvent::Message(event) => self
+                .make_doc(event, |content| {
+                    content.text.find_plain().ok_or(IndexError::EmptyMessage).map(|v| v.to_owned())
+                })
+                .map(RoomIndexOperation::Add),
+
+            _ => Err(IndexError::MessageTypeNotSupported),
+        }
     }
 }
 
--- a/crates/matrix-sdk/Cargo.toml
+++ b/crates/matrix-sdk/Cargo.toml
@@ -92,9 +92,9 @@ matrix-sdk-base.workspace = true
 matrix-sdk-common.workspace = true
 matrix-sdk-ffi-macros = { workspace = true, optional = true }
 matrix-sdk-indexeddb = { workspace = true, optional = true }
+matrix-sdk-search = { workspace = true, optional = true }
 matrix-sdk-sqlite = { workspace = true, optional = true }
 matrix-sdk-test = { workspace = true, optional = true }
-matrix-sdk-search = { workspace = true, optional = true }
 mime.workspace = true
 mime2ext = "0.1.53"
 oauth2.workspace = true
--- a/crates/matrix-sdk/src/client/builder/mod.rs
+++ b/crates/matrix-sdk/src/client/builder/mod.rs
@@ -27,12 +27,8 @@ use homeserver_config::*;
 #[cfg(feature = "e2e-encryption")]
 use matrix_sdk_base::crypto::DecryptionSettings;
 use matrix_sdk_base::{store::StoreConfig, BaseClient, ThreadingSupport};
-#[cfg(feature = "experimental-search")]
-use matrix_sdk_search::index::RoomIndex;
 #[cfg(feature = "sqlite")]
 use matrix_sdk_sqlite::SqliteStoreConfig;
-#[cfg(feature = "experimental-search")]
-use ruma::OwnedRoomId;
 use ruma::{
     api::{error::FromHttpResponseError, MatrixVersion, SupportedVersions},
     OwnedServerName, ServerName,
@@ -42,6 +38,10 @@ use tokio::sync::{broadcast, Mutex, OnceCell};
 use tracing::{debug, field::debug, instrument, Span};
 
 use super::{Client, ClientInner};
+#[cfg(feature = "experimental-search")]
+use crate::client::search::SearchIndex;
+#[cfg(feature = "experimental-search")]
+use crate::client::search::SearchIndexStoreKind;
 #[cfg(feature = "e2e-encryption")]
 use crate::crypto::{CollectStrategy, TrustRequirement};
 #[cfg(feature = "e2e-encryption")]
@@ -123,7 +123,7 @@ pub struct ClientBuilder {
     cross_process_store_locks_holder_name: String,
     threading_support: ThreadingSupport,
     #[cfg(feature = "experimental-search")]
-    index_base_dir: IndexBaseDir,
+    search_index_store_kind: SearchIndexStoreKind,
 }
 
 impl ClientBuilder {
@@ -156,7 +156,7 @@ impl ClientBuilder {
                 Self::DEFAULT_CROSS_PROCESS_STORE_LOCKS_HOLDER_NAME.to_owned(),
             threading_support: ThreadingSupport::Disabled,
             #[cfg(feature = "experimental-search")]
-            index_base_dir: IndexBaseDir::Ram,
+            search_index_store_kind: SearchIndexStoreKind::InMemory,
         }
     }
 
@@ -503,8 +503,8 @@ impl ClientBuilder {
 
     /// The base directory in which each room's index directory will be stored.
     #[cfg(feature = "experimental-search")]
-    pub fn index_base_directory(mut self, path: IndexBaseDir) -> Self {
-        self.index_base_dir = path;
+    pub fn search_index_store(mut self, kind: SearchIndexStoreKind) -> Self {
+        self.search_index_store_kind = kind;
         self
     }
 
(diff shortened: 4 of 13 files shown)


andybalaam:
Handing to @poljar as I think you are looking after this. If this is just a standard review, feel free to hand it back to me.

crates/matrix-sdk/Cargo.toml
```diff
@@ -92,6 +94,7 @@ matrix-sdk-ffi-macros = { workspace = true, optional = true }
 matrix-sdk-indexeddb = { workspace = true, optional = true }
 matrix-sdk-sqlite = { workspace = true, optional = true }
 matrix-sdk-test = { workspace = true, optional = true }
+matrix-sdk-search = { workspace = true, optional = true }
```
bnjbvr: Let's try to keep the alphabetical ordering here, and move this before `-sqlite` :)

crates/matrix-sdk/src/client/builder/mod.rs
```diff
@@ -590,6 +609,10 @@ impl ClientBuilder {
         let event_cache = OnceCell::new();
         let latest_events = OnceCell::new();
 
+        #[cfg(feature = "experimental-search")]
+        let room_indexes: Arc<Mutex<HashMap<OwnedRoomId, RoomIndex>>> =
```
bnjbvr: I think you can remove the fully-qualified type on the left-hand side; Rust should infer it from downstream's use :)

crates/matrix-sdk/src/client/builder/mod.rs
```diff
@@ -717,6 +744,14 @@ async fn build_indexeddb_store_config(
     panic!("the IndexedDB is only available on the 'wasm32' arch")
 }
 
+#[cfg(feature = "experimental-search")]
+#[allow(dead_code)]
+#[derive(Clone, Debug)]
+pub enum IndexBaseDir {
+    Directory(PathBuf),
+    Ram,
+}
```
bnjbvr: - Can we add doc comments explaining what's an `IndexBaseDir`, and what does each variant represent?
- Usually we don't refer to `Ram`, but we could call it `InMemory`, if that's what you had in mind?
- What do you think of naming this `SearchIndexStoreKind` or `SearchIndexBackend`, since this isn't in fact a base directory when it's stored in memory?

crates/matrix-sdk/src/client/mod.rs
```diff
@@ -350,6 +362,14 @@ pub(crate) struct ClientInner {
...
     /// [`LatestEvent`]: crate::latest_event::LatestEvent
     latest_events: OnceCell<LatestEvents>,
+
+    /// HashMap that links each joined room to its RoomIndex
+    #[cfg(feature = "experimental-search")]
+    room_indexes: Arc<Mutex<HashMap<OwnedRoomId, RoomIndex>>>,
+
+    /// Base directory that stores the directories for each RoomIndex
+    #[cfg(feature = "experimental-search")]
+    index_base_dir: IndexBaseDir,
```
bnjbvr: I think we can have a slightly different structure, and there's an opportunity to add an "API object" `Search` here.

- Let's gather these two fields into a `SearchContext` data structure, that's defined in a new `matrix-sdk/src/client/search.rs` module (and this module is guarded by a single `#[cfg(feature = "experimental-search")]`, so we don't have to sprinkle it all over the place :))
  - the `ClientInner` would now have a `SearchContext` data structure, which contains these two fields
- let's have an "API object" (like `Account` or `Media`), that's called `Search`:
  - it would be defined as `struct Search { client: Client }`
  - the new methods `index_event` / `add_index` and so on can all go under an `impl Search` block instead
  - we'd have a new method `Client::search(&self) -> Search`, only defined under the experimental-search feature

crates/matrix-sdk/src/client/mod.rs
```diff
@@ -2810,6 +2842,85 @@ impl Client {
...
+    ) -> Result<(), IndexError> {
+        let mut hash_map = self.inner.room_indexes.lock().await;
+
+        let result = if let Some(index) = hash_map.get_mut(room_id) {
+            index.add_event(event)
+        } else {
+            self.add_index_impl(room_id, &mut hash_map)?;
+            let index = hash_map.get_mut(room_id).expect("key just added");
+            index.add_event(event)
+        };
```
bnjbvr: I'd prefer some pattern like:

```rust
let index = hash_map.entry(room_id.to_owned()).or_insert_with(|| {
  // add the new index
});
```

Then below, you could do `match index.add_event(event)`, which reads nicely.

crates/matrix-sdk/src/client/mod.rs
```diff
@@ -2810,6 +2842,85 @@ impl Client {
...
+            };
+            hash_map.insert(room_id.to_owned(), index);
+        }
+        Ok(())
+    }
+
+    /// Search a [`Room`]'s index for the query and return at most
+    /// max_number_of_results results.
+    #[cfg(feature = "experimental-search")]
+    pub async fn search_index(
```
bnjbvr: Could this be internal to the crate, instead? I think it's better to keep the public API on the `Room`, as you've done :)

crates/matrix-sdk/src/event_cache/room/mod.rs
```diff
@@ -551,6 +600,37 @@ impl RoomEventCacheInner {
             self.pagination_batch_token_notifier.notify_one();
         }
 
+        // We can also add the event to the index.
+        #[cfg(feature = "experimental-search")]
+        for diff in timeline_event_diffs.iter() {
```
bnjbvr: So, while this likely achieves what you want, there's a simpler way to do it :)

There's a `post_process_new_events` in the event cache, that will be called both for back-pagination and sync. For it to work great, I think the following is required (but we can implement it in subsequent PRs):

- indexing must be idempotent, i.e. if I index the same event twice, it shouldn't lead to an error
- indexing must handle redactions in any order, i.e. it must redact, even if the redaction event is seen before the redacted event (yay for decentralized systems :))
- indexing must handle edits in any order (same as above)

We can focus on those edge cases in the future, though :) In the meantime, can you make it so that this code is called under `post_process_new_events`, please?

crates/matrix-sdk/src/client/mod.rs
```diff
@@ -2810,6 +2842,107 @@ impl Client {
...
+        if let Some(index) = hash_map.get_mut(room_id) {
+            let _ = index.commit().inspect_err(|err| {
+                error!("error occured while committing: {err:?}");
+            });
+        }
+    }
+
+    /// Commit a [`Room`]'s [`RoomIndex`] and reload searchers
+    #[cfg(feature = "experimental-search")]
+    pub async fn commit_and_reload(&self, room_id: &RoomId) {
```
bnjbvr: Being not familiar with search, I'm wondering if this should be a public API, or if we should keep it private, and automatically commit at the right time, on behalf of the embedder?

crates/matrix-sdk/src/event_cache/room/mod.rs
```diff
@@ -523,6 +529,49 @@ impl RoomEventCacheInner {
...
+                    AnySyncTimelineEvent::MessageLike(sync_event) => {
+                        Some(Ok(sync_event.into_full_event(self.room_id.clone())))
+                    }
+                    AnySyncTimelineEvent::State(_) => None,
+                },
+                Err(e) => Some(Err(e)),
+            },
+
+            TimelineEventKind::UnableToDecrypt { event: _, utd_info: _ } => None,
+        };
```
bnjbvr: Can this be `let maybe_try_event = event.raw().deserialize_as()` instead? Or is it important to keep UTD out of the index? (I suspect they wouldn't be indexed just thereafter)

crates/matrix-sdk/src/client/builder/mod.rs
dragonfly1033: Yep these make sense, will do.

crates/matrix-sdk/src/client/mod.rs
dragonfly1033: Ahh, thats much cleaner. Can we rename the `Search` object because it is more of a `SearchIndexHandler`. Also because in the future I expect to have a `SearchQuery` type so it may be ambiguous without reading docs.

bnjbvr: I'm open to a slightly longer name, maybe `SearchIndex`. The intent behind `Search` was that the end-user (embedders of the SDK) likely aren't interested in the implementation details themselves, but mostly care about the feature (which is… search). And I like short names. But `Search` or `SearchIndex` could work, IMO.

crates/matrix-sdk/src/client/mod.rs
dragonfly1033: `or_insert_with` doesn't allow results and the nicest I have found to handle this is 
```rust
if !self.index_map.contains_key(room_id) {
    let index = self.create_index(room_id)?;
    self.index_map.insert(room_id.to_owned(), index);
}

let index = self.index_map.get_mut(room_id).expect("index should exist");
let result = index.add_event(event);
```
is this ok?
(result needs to exist so I can use it in the dbug logs directly like `{result:?}`)

crates/matrix-sdk/src/event_cache/room/mod.rs
dragonfly1033: How would I get a `Room` object for the given `RoomId` in the `RoomEventCacheState`? On the other hand I can leave `parse/index_timeline_event` in `RoomEventCacheInner` if I can get a reference to it in `RoomEventCacheState` but I don't think I can.

crates/matrix-sdk/src/client/mod.rs
bnjbvr: You're right, indeed there's no good stdlib API for this. Yep, your suggestion looks great to me :+1:

crates/matrix-sdk/src/client/mod.rs
dragonfly1033: The plan is to index events in batches along with the event cache. Upon doing this we can call this internally but until then it has to be called externally. Calling this on every insertion is a bad idea but if it is preferrable then I can have it called internally right before `search_index` until we get the batching done.

crates/matrix-sdk/src/event_cache/room/mod.rs
dragonfly1033: We can only index messages we can parse (so far only `m.room.message`). I'm aware there is some kind of deferred decryption that I need to handle as well but I don't think that happens here.

crates/matrix-sdk/src/event_cache/room/mod.rs
bnjbvr: I see. I've got this problem in another PR as well, where the `Room` isn't available, so likely we'd need to pass it as a new parameter to `post_process_new_event`; both callers have room not too far away:

- `RoomEventCacheState::handle_sync` can have a new `&Room` parameter, and its [one caller](https://github.com/matrix-org/matrix-rust-sdk/blob/6c944a9b39250dce7a01090ab8761ce136d1a378/crates/matrix-sdk/src/event_cache/room/mod.rs#L545-L546) can materialize it with `self.weak_room.get()`.
- `RoomEventCacheState::handle_back_pagination` can have a new `&Room` parameter, and its [one caller](https://github.com/matrix-org/matrix-rust-sdk/blob/6c944a9b39250dce7a01090ab8761ce136d1a378/crates/matrix-sdk/src/event_cache/pagination.rs#L291) can reuse the previous `Room` that was [materialized a few lines above](https://github.com/matrix-org/matrix-rust-sdk/blob/6c944a9b39250dce7a01090ab8761ce136d1a378/crates/matrix-sdk/src/event_cache/pagination.rs#L291)

(I might need something similar for the sake of my thread subscription PR)

crates/matrix-sdk/src/client/mod.rs
bnjbvr: Makes sense. I think, as a first intent, to commit upon `search_index`, is a good approximation to start with. Later, we could use dirty state tracking (i.e. if new events have been indexed, set a dirty bit to true; if attempting to call `search_index` and the dirty bit is set, commit first), and add automatic background committing on top of that.

crates/matrix-sdk/src/event_cache/room/mod.rs
bnjbvr: Indeed, let's not worry about deferred decryption quite yet :+1: 

It's fine to keep as is, at the moment, but a small note: `MessageLike` doesn't mean you'll get a `m.room.message` at the end of the day. It merely means that this is not a state event, so it can be e.g. a redaction event (which doesn't have the type `m.room.message`).

andybalaam: Side note: some other event types would be really good to parse - especially polls [MSC3381](https://github.com/matrix-org/matrix-spec-proposals/pull/3381)

dragonfly1033: > Side note: some other event types would be really good to parse - especially polls [MSC3381](https://github.com/matrix-org/matrix-spec-proposals/pull/3381)

After a brief look, this will require further conversation but I'll add it to the TODO list.

dragonfly1033: > Indeed, let's not worry about deferred decryption quite yet 👍
> 
> It's fine to keep as is, at the moment, but a small note: `MessageLike` doesn't mean you'll get a `m.room.message` at the end of the day. It merely means that this is not a state event, so it can be e.g. a redaction event (which doesn't have the type `m.room.message`).

Yep, the `AnyMessageLikeEvent` is further parsed by the schema in `crates/matrix-sdk-search/src/schema.rs` and returns an error for unsupported event types.

crates/matrix-sdk/src/client/builder/mod.rs
```diff
@@ -489,6 +501,13 @@ impl ClientBuilder {
         self
     }
 
+    /// The base directory in which each room's index directory will be stored.
+    #[cfg(feature = "experimental-search")]
+    pub fn index_base_directory(mut self, path: SearchIndexStoreKind) -> Self {
```
bnjbvr: Can we rename this helper `search_index_store`, and rename `path` to `kind` or `store_kind`?

crates/matrix-sdk/src/client/builder/mod.rs
```diff
@@ -590,6 +609,12 @@ impl ClientBuilder {
         let event_cache = OnceCell::new();
         let latest_events = OnceCell::new();
 
+        #[cfg(feature = "experimental-search")]
+        let room_indexes = Arc::new(Mutex::new(HashMap::new()));
```
bnjbvr: light suggestion: if you inline `room_indexes` below into its one use, then we can get rid of one `cfg` guard \o/

crates/matrix-sdk/src/client/search.rs
```diff
@@ -0,0 +1,146 @@
...
+    ) -> Option<Vec<OwnedEventId>> {
+        if let Some(index) = self.index_map.get(room_id) {
+            index
+                .search(query, max_number_of_results)
+                .inspect_err(|err| {
+                    error!("error occurred while searching index: {err:?}");
+                })
+                .ok()
+        } else {
+            warn!("Tried to search in a room with no index");
```
bnjbvr: It's probably not `warn`-worthy, in that it might happen a lot of times; let's make it `debug!` at best :)

crates/matrix-sdk/src/event_cache/room/mod.rs
```diff
@@ -542,8 +542,16 @@ impl RoomEventCacheInner {
...
-        let (stored_prev_batch_token, timeline_event_diffs) =
-            self.state.write().await.handle_sync(timeline).await?;
+        let (stored_prev_batch_token, timeline_event_diffs) = self
+            .state
+            .write()
+            .await
+            .handle_sync(
+                timeline,
+                #[cfg(feature = "experimental-search")]
+                &self.weak_room.get().expect("Should be able to get room"),
```
bnjbvr: This `.expect()` is wrong, I think; it could be that the client is shutting down. Instead, let's do `let Some(room) = self.weak_room.get() else { warn!("..."); return; }`

crates/matrix-sdk/src/event_cache/room/mod.rs
```diff
@@ -1434,6 +1448,49 @@ mod private {
...
+                    Err(e) => Some(Err(e)),
+                },
+
+                TimelineEventKind::UnableToDecrypt { event: _, utd_info: _ } => None,
+            };
+
+            match maybe_try_event {
+                Some(Ok(event)) => match event {
+                    // TODO: Handle room redaction.
+                    AnyMessageLikeEvent::RoomRedaction(_) | _ => Some(event),
```
bnjbvr: Regarding responsibilities: who's in charge of saying that such or such event is fine to be indexed? I'd expect this to happen at the lowest common denominator, so, within the `Room::index_event` function, or even something called deeper in there.

If we're agreeing here, then I think that this whole function could be simplified to a mere `event.raw().deserialize().ok()` or something like this. We wouldn't inspect the event's kind here, and one of the functions we're calling under the hood would do that for us, dismiss UTDs, and so on and so forth. How does that sound?

crates/matrix-sdk/tests/integration/room/joined.rs
```diff
@@ -740,6 +740,38 @@ async fn test_room_message_send() {
...
+                .sender(user_id)
+                .into_raw_sync()]),
+        )
+        .await;
+
+    let room = client.get_room(room_id).unwrap();
+
+    let response = room.search_index("this", 5).await.expect("search should have 1 result");
+
+    assert!(!response.is_empty(), "no results found {response:?}");
```
bnjbvr: We can do even better, and assert the number of responses:

```suggestion
    assert_eq!(response.len(), 1, "unexpected numbers of responses: {response:?}");
```

crates/matrix-sdk/src/client/search.rs
```diff
@@ -0,0 +1,146 @@
+// Copyright 2024 The Matrix.org Foundation C.I.C.
```
bnjbvr: How times flies :)

bnjbvr (changes requested):
A few more questions, but this is almost good to land :partying_face:

crates/matrix-sdk/src/event_cache/room/mod.rs
dragonfly1033: `Schema::parse_any_event` is responsible for this and is called only when indexing. The idea behind this line was that `AnyMessageLikeEvent::RoomRedaction` (unlike every other variant) should 
1. remove the message from the index 
2. return None so it is skipped in indexing.

I think, however, that the redaction handling will/should happen previous to this in `maybe_apply_new_redaction`. In which case this handling is not needed here and the match can be removed.   

The alternative of passing the whole thing down to the search crate and doing everything there is possible but then you would have to change the semantics from `index_event` to `handle_event` because it may or may not add/remove from the index depending on the event.

crates/matrix-sdk-search/src/schema.rs
```diff
@@ -140,17 +122,25 @@ impl MatrixSearchIndexSchema for RoomMessageSchema {
         self.inner.clone()
     }
 
-    fn make_doc(&self, event: AnyMessageLikeEvent) -> Result<TantivyDocument, IndexError> {
-        let (event_id, body, timestamp, sender) = self.parse_any_event(event)?;
+    fn handle_event(&self, event: AnyMessageLikeEvent) -> Result<RoomIndexOperation, IndexError> {
+        match event {
+            // old m.room.message behaviour
```
bnjbvr: pre-existing, but what do you mean by `old` here? All messages are `m.room.message`, so it seems pretty current to me, while I've never seen a `m.message` in the wild :grin:

crates/matrix-sdk-search/src/schema.rs
```diff
@@ -58,48 +58,30 @@ pub(crate) struct RoomMessageSchema {
...
+        get_body_from_content: F,
+    ) -> Result<TantivyDocument, IndexError>
     where
         <C as RedactContent>::Redacted: RedactedMessageLikeEventContent,
         F: FnOnce(&C) -> Result<String, IndexError>,
     {
         let unredacted = event.as_original().ok_or(IndexError::CannotIndexRedactedMessage)?;
 
-        let body = get_body(&unredacted.content)?;
+        let body = get_body_from_content(&unredacted.content)?;
```
bnjbvr, line 65: kinda preexisting: `AnyMessageLikeEvent` has `sender()`, `origin_server_ts()` and `event_id()`. So I think the code could be reorganized to avoid the generic function `make_doc` with the generic parameter `get_body_from_content`.

In `handle_event`, you'd get the sender/origin_server_ts/event_id from the `AnyMessageLikeEvent` getter, than match on the event type to read the body.

(We can address this as a followup, to avoid exploding the size of this PR.)

crates/matrix-sdk/src/client/search.rs
```diff
@@ -0,0 +1,147 @@
...
+                error!("failed to handle event in index: {err:?}")
+            }
+            Err(_) => error!("unexpected error during indexing: {result:?}"),
+        }
+        Ok(())
+    }
+
+    /// Search a [`Room`]'s index for the query and return at most
+    /// max_number_of_results results.
+    pub(crate) fn search_index(
```
bnjbvr: How do you feel about calling this `search`? the index mostly sounds like an implementation detail to me

crates/matrix-sdk/src/event_cache/room/mod.rs
```diff
@@ -542,8 +542,22 @@ impl RoomEventCacheInner {
         // Add all the events to the backend.
         trace!("adding new events");
 
-        let (stored_prev_batch_token, timeline_event_diffs) =
-            self.state.write().await.handle_sync(timeline).await?;
+        #[cfg(feature = "experimental-search")]
+        let Some(room) = self.weak_room.get() else {
+            warn!("Couldn't get room while handling timeline");
+            return Err(EventCacheError::RoomNotFound { room_id: self.room_id.clone() });
```
bnjbvr: It's likely a minor error, because this usually means the `Client` is shutting down (the main reference to the client has been dropped), so we can return `Ok(())` here

crates/matrix-sdk/src/event_cache/room/mod.rs
```diff
@@ -1434,6 +1454,46 @@ mod private {
...
         }
 
+        #[cfg(feature = "experimental-search")]
+        fn parse_timeline_event(&self, event: &TimelineEvent) -> Option<AnyMessageLikeEvent> {
+            let maybe_try_event = match &event.kind {
+                TimelineEventKind::Decrypted(d) => Some(d.event.deserialize()),
+                TimelineEventKind::PlainText { event } => match event.deserialize() {
+                    Ok(event_obj) => match event_obj {
+                        AnySyncTimelineEvent::MessageLike(sync_event) => {
+                            Some(Ok(sync_event.into_full_event(self.room.clone())))
```
bnjbvr: Since the search indexes are per room, we don't need to store the room id into each event, right? So we could keep an `AnySyncMessageLikeEvent` (i.e. without room id) instead of `AnyMessageLikeEvent` (i.e. with room id), right?

crates/matrix-sdk/src/event_cache/room/mod.rs
```diff
@@ -1434,6 +1454,46 @@ mod private {
...
+                    Err(e) => Some(Err(e)),
+                },
+
+                TimelineEventKind::UnableToDecrypt { event: _, utd_info: _ } => None,
+            };
+
+            match maybe_try_event {
+                Some(Ok(event)) => Some(event),
+                Some(Err(e)) => {
+                    warn!("failed to index event: {e:?}");
```
bnjbvr:
```suggestion
                    warn!("failed to parse event: {e:?}");
```

crates/matrix-sdk/src/event_cache/room/mod.rs
```diff
@@ -1434,6 +1454,46 @@ mod private {
...
+                    None
+                }
+                None => None,
+            }
+        }
+
+        #[cfg(feature = "experimental-search")]
+        async fn index_handle_timeline_event(
```
bnjbvr: What's the "handle" about in this name? `index_timeline_event` sounds fine to me, fwiw :)

crates/matrix-sdk/src/room/mod.rs
```diff
@@ -3671,6 +3678,38 @@ impl Room {
...
+        &self,
+        event: AnyMessageLikeEvent,
+    ) -> Result<(), IndexError> {
+        self.client.search_index().lock().await.handle_event(event, self.room_id())
+    }
+
+    /// Search this room's [`RoomIndex`] for query and return at most
+    /// max_number_of_results results.
+    #[cfg(feature = "experimental-search")]
+    pub async fn search_index(
```
bnjbvr: (Same as above, could we name this `search` maybe?)

crates/matrix-sdk/src/room/mod.rs
```diff
@@ -3671,6 +3678,38 @@ impl Room {
...
+        self.client.search_index().lock().await.search_index(
+            query,
+            max_number_of_results,
+            self.room_id(),
+        )
+    }
+
+    /// Commit a [`Room`]'s [`RoomIndex`] and reload searchers
+    #[cfg(feature = "experimental-search")]
+    pub(crate) async fn commit_and_reload(&self) {
```
bnjbvr: On the other hand, I think this function name should refer to the search index.

crates/matrix-sdk/tests/integration/room/joined.rs
```diff
@@ -740,6 +740,38 @@ async fn test_room_message_send() {
...
+            &client,
+            JoinedRoomBuilder::new(room_id).add_timeline_bulk(vec![event_factory
+                .text_msg("this is a sentence")
+                .event_id(event_id)
+                .sender(user_id)
+                .into_raw_sync()]),
+        )
+        .await;
+
+    let room = client.get_room(room_id).unwrap();
```
bnjbvr:
```suggestion
    let room = mock_server
        .sync_room(
            &client,
            JoinedRoomBuilder::new(room_id).add_timeline_bulk(vec![event_factory
                .text_msg("this is a sentence")
                .event_id(event_id)
                .sender(user_id)
                .into_raw_sync()]),
        )
        .await;
```

crates/matrix-sdk/src/event_cache/room/mod.rs
dragonfly1033: The next steps (laid out in #5350 ) will involve benchmarking per-room indexes vs global index for global searching to see if there's a significant performace impact. Hence, I'm passing it down so it's one less change to make for the experiment but if we decide to stick ewith per-room then you're right.

crates/matrix-sdk/src/event_cache/room/mod.rs
dragonfly1033: `index_timeline_event` to me means to index (verb) the timeline event (noun) which is not what that function does. But `index_handle_timeline_event` was my best approximation for "Handle this event in the index" or "Index handles this event". 

I didn't really like the name either but it's internal and its body is relatively self describing so I thought it's good enought without being too verbose :man_shrugging: .

crates/matrix-sdk-search/src/schema.rs
dragonfly1033: I saw there were 2 different message types the ~~old~~ current `m.room.message` and the new `msc1767` `m.message` . You're technically right, I just assumed we'd migrate eventually. I'll change to current and I'll label the new one with the msc

crates/matrix-sdk/src/event_cache/room/mod.rs
bnjbvr: Sorry, I still don't understand the reasoning here. The indexes are per room. For getting the index, you need to have a `Room` object around. If you have a `Room` object around, you can get the room id via `Room::room_id()`. So, if you were to search in the index, you don't need to get the room id from the message; it's always the one for the room you're looking at. So we don't need to include the room id in the indexed event.

crates/matrix-sdk/src/event_cache/room/mod.rs
bnjbvr: It's quite uncommon to have `subject verb noun` in a function's name; usually we'd do more `verb noun`, at best. Also, the current name reads like `IndexHandle / TimelineEvent`, which is definitely confusing. Can't we use `index` as a verb? Is this not indexing the timeline event?

bnjbvr (changes requested):
See new responses in review comments.

crates/matrix-sdk/src/client/search.rs
```diff
@@ -0,0 +1,150 @@
...
+//! inferring some extra useful information about each room. In particular, this
+//! doesn't require subscribing to a specific room to get access to this
+//! information.
+//!
+//! It's intended to be fast, robust and easy to maintain, having learned from
+//! previous endeavours at implementing middle to high level features elsewhere
+//! in the SDK, notably in the UI's Timeline object.
+//!
+//! See the [github issue](https://github.com/matrix-org/matrix-rust-sdk/issues/3058) for more
+//! details about the historical reasons that led us to start writing this.
```
bnjbvr: Bad copy pasto? :)

```suggestion
```

crates/matrix-sdk/src/event_cache/room/mod.rs
```diff
@@ -542,8 +542,22 @@ impl RoomEventCacheInner {
         // Add all the events to the backend.
         trace!("adding new events");
 
-        let (stored_prev_batch_token, timeline_event_diffs) =
-            self.state.write().await.handle_sync(timeline).await?;
+        #[cfg(feature = "experimental-search")]
+        let Some(room) = self.weak_room.get() else {
+            warn!("Couldn't get room while handling timeline");
```
bnjbvr:
```suggestion
            debug!("Couldn't get room while handling timeline");
```

crates/matrix-sdk-search/src/index.rs
```diff
@@ -130,10 +122,11 @@ impl RoomIndex {
         RoomIndex::new_with(index, schema, room_id)
     }
 
-    /// Add [`AnyMessageLikeEvent`] to [`RoomIndex`]
-    pub fn add_event(&mut self, event: AnyMessageLikeEvent) -> Result<(), IndexError> {
-        let doc = self.schema.make_doc(event)?;
-        self.writer.add_document(doc)?; // TODO: This is blocking. Handle it.
+    /// Handle [`AnySyncMessageLikeEvent`]
+    pub fn handle_event(&mut self, event: AnySyncMessageLikeEvent) -> Result<(), IndexError> {
```
bnjbvr: Would be sweet to write a more descriptive comment here; this is repeating the function name and type, so doesn't quite help figuring out what it does really, or why it does it :)

crates/matrix-sdk-search/src/index.rs
```diff
@@ -130,10 +122,11 @@ impl RoomIndex {
...
+        match self.schema.handle_event(event)? {
+            RoomIndexOperation::Add(document) => self.writer.add_document(document)?,
+        };
```
bnjbvr, line 132: Not sure this semicolon is necessary:

```suggestion
        match self.schema.handle_event(event)? {
            RoomIndexOperation::Add(document) => self.writer.add_document(document)?,
        }
```

crates/matrix-sdk-search/src/index.rs
```diff
@@ -44,6 +31,11 @@ use crate::{
     writer::SearchIndexWriter,
 };
 
+/// A struct to represent the operations on a [`RoomIndex`]
+pub(crate) enum RoomIndexOperation {
+    Add(TantivyDocument),
```
bnjbvr, line 36: Would be sweet to add a doc-comment to this variant too.

crates/matrix-sdk/src/client/search.rs
```diff
@@ -0,0 +1,137 @@
...
+
+    /// Base directory that stores the directories for each RoomIndex
+    search_index_store_kind: &'a SearchIndexStoreKind,
+}
+
+impl SearchIndexGuard<'_> {
+    fn create_index(&self, room_id: &RoomId) -> Result<RoomIndex, IndexError> {
+        let index = match self.search_index_store_kind {
+            SearchIndexStoreKind::Directory(path) => RoomIndex::open_or_create(path, room_id)?,
+            SearchIndexStoreKind::InMemory => RoomIndex::new_in_ram(room_id)?,
```
bnjbvr: (for a follow-up, and it may be preexisting: think `new_in_ram` could be named `new_in_memory`)

crates/matrix-sdk/src/event_cache/room/mod.rs
```diff
@@ -542,8 +544,22 @@ impl RoomEventCacheInner {
         // Add all the events to the backend.
         trace!("adding new events");
 
-        let (stored_prev_batch_token, timeline_event_diffs) =
-            self.state.write().await.handle_sync(timeline).await?;
+        #[cfg(feature = "experimental-search")]
+        let Some(room) = self.weak_room.get() else {
+            debug!("Couldn't get room while handling timeline");
```
bnjbvr: This only happens because the `Client`'s instance is shutting down (i.e. being `Drop`ped elsewhere), so let's put it to a trace level, as it's mostly uninteresting.

```suggestion
            trace!("Couldn't get room while handling timeline");
```

crates/matrix-sdk/src/event_cache/room/mod.rs
```diff
@@ -1434,6 +1456,49 @@ mod private {
...
+            };
+
+            match maybe_try_event {
+                Some(Ok(event)) => Some(event),
+                Some(Err(e)) => {
+                    warn!("failed to parse event: {e:?}");
+                    None
+                }
+                None => None,
+            }
```
bnjbvr, line 1481: Still unclear why we can't do a plain `event.raw().deserialize()` at the one call site of this function, but let's keep it like this for this current version.

crates/matrix-sdk/src/room/mod.rs
```diff
@@ -3671,6 +3677,37 @@ impl Room {
...
+        self.commit_and_reload_search_index().await;
+
+        self.client.search_index().lock().await.search(query, max_number_of_results, self.room_id())
+    }
+
+    /// Commit a [`Room`]'s [`RoomIndex`] and reload searchers
+    #[cfg(feature = "experimental-search")]
+    pub(crate) async fn commit_and_reload_search_index(&self) {
+        self.client.search_index().lock().await.commit_and_reload(self.room_id());
+    }
```
bnjbvr: Apparently this is called only once; if i'm not mistaken, could you inline it at its one call site above?

crates/matrix-sdk/src/room/mod.rs
```diff
@@ -3671,6 +3677,37 @@ impl Room {
...
+    ///
+    /// This which will add/remove/edit an event in the index based on the
+    /// event type.
+    #[cfg(feature = "experimental-search")]
+    pub(crate) async fn index_event(
+        &self,
+        event: AnySyncMessageLikeEvent,
+    ) -> Result<(), IndexError> {
+        self.client.search_index().lock().await.handle_event(event, self.room_id())
+    }
```
bnjbvr, line 3690: I do realize only now; it might be better to have the callers of this method directly call the `Client`'s method, since if they have a `Room` around they have a `Client` around. And having `Room`-methods is mostly nice for public APIs (like `search` below), otherwise there's no need to have an indirect call structure like this.

crates/matrix-sdk/tests/integration/room/joined.rs
```diff
@@ -740,6 +740,36 @@ async fn test_room_message_send() {
     assert_eq!(event_id!("$h29iv0s8:example.com"), response.event_id)
 }
 
+#[cfg(feature = "experimental-search")]
+#[async_test]
+async fn test_sending_message_indexes_message() {
```
bnjbvr: uber-nit: a message received during sync'd is not necessarily a message we've sent (could have been another user, another device, etc.)

```suggestion
async fn test_sync_message_is_indexed() {
```

PR #5507, opened 2025-08-08
feat(multiverse): Add room creation to multiverse

dragonfly1033:
Pressing Ctrl+R in multiverse will open a popup dialogue box where you can enter a room name which will be created when you press Enter.
Relevant information has been added to the status bar and help screen.


labs/multiverse/src/main.rs
```diff
@@ -411,12 +421,41 @@ impl App {
                             self.set_global_mode(GlobalMode::Default);
                         }
                     }
+                    GlobalMode::CreateRoom { view } => {
+                        if let Event::Key(key) = event
+                            && let KeyModifiers::NONE = key.modifiers
+                        {
+                            match key.code {
+                                Enter => {
+                                    if let Some(room_name) = view.get_text() {
```
andybalaam, line 430: We should check whether they are trying to create a room with an empty name and prevent this, I think.

andybalaam: Oh, I see you handle this, never mind :-)

labs/multiverse/src/widgets/room_view/mod.rs
```diff
@@ -525,11 +525,7 @@ impl RoomView {
                     Ok(Some(subscription)) => {
                         status_handle.set_message(format!(
                             "Thread subscription status: {}",
-                            if subscription.automatic {
-                                "subscribed (automatic)"
-                            } else {
-                                "subscribed (manual)"
-                            }
+                            subscription.as_str()
```
andybalaam: This seems irrelevant - should be removed or a separate PR?

dragonfly1033: Yep, it's a fix from a breaking change in another crate.

dragonfly1033: Change removed and no longer breaking (I was working off main from 3 days ago :man_facepalming:)

andybalaam:
By the way, you should not force-push while the PR is being reviewed because it makes it hard for the reviewer to see what happened. You can squash-merge or rebase and force-push when the review process is over. (Which arguably it was here, because I ticked the PR, so this is just FYI :-)

Do you need me to hit merge, or can you?

dragonfly1033:
Ok, noted thx :+1: . I, probably for the best, cannot merge so could you please?

PR #5509, opened 2025-08-08
feat(sdk): Spaces

stefanceriu:
This series of patches introduces a top level UI oriented interface for interacting with [spaces](https://spec.matrix.org/latest/client-server-api/#spaces). 

The `SpaceService` offers:
* a method for retrieving one's top level `joined_spaces` and a `subscribe_to_joined_spaces` stream that automatically updates based on client activity
    * to do so the service builds a graph from all the `m.space.parent` and `m.space.children` state events it knows about, removing cycles and then only keeping parent spaces
* a `SpaceRoomList` that can be used to interact with the [rooms hierarchy](https://spec.matrix.org/latest/client-server-api/#get_matrixclientv1roomsroomidhierarchy) endpoint and retrieve rooms referencing a certain parent
    * the room list is responsible for holding the pagination state as well as listening to client activity and automatically publishing room state updates through its `subscribe_to_room_updates` publisher (e.g. when joining one of the rooms)

I do appreciate this is a pretty chunky PR but it should be straight forward to review commit by commit.


bindings/matrix-sdk-ffi/src/spaces.rs
```diff
@@ -0,0 +1,180 @@
...
+    }
+}
+
+#[matrix_sdk_ffi_macros::export]
+impl SpaceServiceRoomList {
+    pub fn pagination_state(&self) -> SpaceServiceRoomListPaginationState {
+        self.inner.pagination_state().into()
+    }
+
+    pub fn subscribe_to_pagiation_state_updates(
```
pixlwave:
```suggestion
    pub fn subscribe_to_pagination_state_updates(
```

crates/matrix-sdk-ui/src/spaces/mod.rs
```diff
@@ -0,0 +1,148 @@
...
+impl SpaceService {
+    pub fn new(client: Client) -> Self {
+        Self { client }
+    }
+
+    pub fn joined_spaces(&self) -> Vec<OwnedRoomId> {
+        self.client
+            .joined_rooms()
+            .into_iter()
+            .filter_map(|room| if room.is_space() { Some(room.room_id().to_owned()) } else { None })
```
Hywan: Another funny way to write it:

```rust
.filter_map(|room| room.is_space().then(|| Some(room.room_id().to_owned())))
```

crates/matrix-sdk-ui/src/spaces/room.rs
```diff
@@ -0,0 +1,76 @@
...
+// limitations under the License.
+
+use matrix_sdk::{Room, RoomHero, RoomState};
+use ruma::events::room::guest_access::GuestAccess;
+use ruma::events::room::history_visibility::HistoryVisibility;
+use ruma::room::{JoinRuleSummary, RoomSummary, RoomType};
+use ruma::{OwnedMxcUri, OwnedRoomAliasId, OwnedRoomId};
+
+#[derive(Debug)]
+pub struct SpaceServiceRoom {
```
Hywan: I really don't like this name :-D. What about `space_service::DiscoveredRoom`?

crates/matrix-sdk-ui/src/spaces/room_list.rs
```diff
@@ -0,0 +1,223 @@
...
+use std::sync::Mutex;
+
+use eyeball::{SharedObservable, Subscriber};
+use matrix_sdk::{Client, Error, paginators::PaginationToken};
+use ruma::{OwnedRoomId, api::client::space::get_hierarchy};
+
+use crate::spaces::SpaceServiceRoom;
+
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub enum SpaceServiceRoomListPaginationState {
```
Hywan: Remove the `SpaceService` prefix.

crates/matrix-sdk-ui/src/spaces/room_list.rs
```diff
@@ -0,0 +1,223 @@
...
+    Idle { end_reached: bool },
+    Loading,
+}
+
+pub struct SpaceServiceRoomList {
```
Hywan: Remove the `SpaceService` prefix.

crates/matrix-sdk-ui/src/spaces/room_list.rs
```diff
@@ -0,0 +1,223 @@
...
+    client: Client,
+
+    parent_space_id: OwnedRoomId,
+
+    token: Mutex<PaginationToken>,
+
+    pagination_state: SharedObservable<SpaceServiceRoomListPaginationState>,
+
+    rooms: SharedObservable<Vec<SpaceServiceRoom>>,
```
Hywan: What about `ObservableVector<_>` instead?

crates/matrix-sdk-ui/src/spaces/room_list.rs
```diff
@@ -0,0 +1,223 @@
...
+                current_rooms.extend(
+                    result
+                        .rooms
+                        .iter()
+                        .map(|room| (&room.summary, self.client.get_room(&room.summary.room_id)))
+                        .map(|(summary, room)| SpaceServiceRoom::new_from_summary(summary, room))
+                        .collect::<Vec<_>>(),
+                );
+
+                self.rooms.set(current_rooms.clone());
```
Hywan: With `ObservableVector`:

```rust
self.rooms.extend(result.rooms.iter().map(…));
```

or:

```rust
let rooms = self.rooms.read();

for room in result.rooms.iter().map(…)… {
    rooms.push_back(room);
}
```

crates/matrix-sdk-ui/src/spaces/mod.rs
```diff
@@ -26,37 +29,87 @@ pub mod room_list;
 
 pub struct SpaceService {
     client: Client,
+
+    joined_spaces: SharedObservable<Vec<SpaceServiceRoom>>,
```
Hywan: `ObservableVector` again.

crates/matrix-sdk-ui/src/spaces/mod.rs
```diff
@@ -26,37 +29,87 @@ pub mod room_list;
...
+                        if new_spaces != joined_spaces_clone.get() {
+                            joined_spaces_clone.set(new_spaces);
+                        }
+                    }
+                    Err(err) => {
+                        error!("error when listening to room updates: {err}");
+                    }
+                }
+            }
+        });
```
Hywan: Be careful to use `AbortOnDrop` here, we don't want the task to be detached.

Hywan (commented):
First fast review. Keep going!

crates/matrix-sdk-ui/src/spaces/room_list.rs
stefanceriu: Definitely, it's the second thing on the to do list. Decided to start simple with full vector publishing and add diffing as a nicety towards the end.

crates/matrix-sdk-ui/src/spaces/mod.rs
stefanceriu: Oh, isn't manually aborting it in `Drop for SpaceService` enough? 🤔

bindings/matrix-sdk-ffi/src/spaces.rs
pixlwave: Resolved but remains unfixed 😢

bindings/matrix-sdk-ffi/src/client.rs
```diff
@@ -1259,6 +1261,11 @@ impl Client {
         SyncServiceBuilder::new((*self.inner).clone(), self.utd_hook_manager.get().cloned())
     }
 
+    pub fn spaces_service(&self) -> Arc<SpaceService> {
```
pixlwave: Minor nit: This would be more consistent with the types.
```suggestion
    pub fn space_service(&self) -> Arc<SpaceService> {
```

bindings/matrix-sdk-ffi/src/spaces.rs
```diff
@@ -0,0 +1,214 @@
...
+    TaskHandle,
+};
+
+#[derive(uniffi::Object)]
+pub struct SpaceService {
+    inner: UISpaceService,
+}
+
+impl SpaceService {
+    pub fn new(inner: UISpaceService) -> Self {
```
pixlwave: Does this one need to be `pub`?

bindings/matrix-sdk-ffi/src/spaces.rs
```diff
@@ -0,0 +1,214 @@
...
+    }
+}
+
+#[derive(uniffi::Object)]
+pub struct SpaceServiceRoomList {
+    inner: UISpaceServiceRoomList,
+}
+
+impl SpaceServiceRoomList {
+    pub fn new(inner: UISpaceServiceRoomList) -> Self {
```
pixlwave: Same here?

bindings/matrix-sdk-ffi/src/spaces.rs
```diff
@@ -0,0 +1,214 @@
...
+        match state {
+            UISpaceServiceRoomListPaginationState::Idle { end_reached } => {
+                SpaceServiceRoomListPaginationState::Idle { end_reached }
+            }
+            UISpaceServiceRoomListPaginationState::Loading => {
+                SpaceServiceRoomListPaginationState::Loading
+            }
+        }
+    }
+}
```
pixlwave: Is there a reason not to put the `#[derive(uniffi::Enum)]` on the enum in the UI crate and remove the FFI's copy entirely?

bindings/matrix-sdk-ffi/src/spaces.rs
```diff
@@ -0,0 +1,214 @@
...
+
+#[matrix_sdk_ffi_macros::export(callback_interface)]
+pub trait SpaceServiceRoomListEntriesListener: SendOutsideWasm + SyncOutsideWasm + Debug {
+    fn on_update(&self, rooms: Vec<SpaceServiceRoom>);
+}
+
+#[matrix_sdk_ffi_macros::export(callback_interface)]
+pub trait SpaceServiceJoinedSpacesListener: SendOutsideWasm + SyncOutsideWasm + Debug {
+    fn on_update(&self, rooms: Vec<SpaceServiceRoom>);
+}
```
pixlwave: Up to you but I noticed whilst conforming out `SDKListener` to these that they're identical. So maybe a single generic `SpaceServiceSpaceRoomsListener` could suffice here 🌵

pixlwave: Not sure how the cactus got in there, but leaving it because it is cute 😆

pixlwave (commented):
This is only a review on the FFI layer, but looks great!. Can't wait for it to land 🤩

spacex-2-3537887872

crates/matrix-sdk-ui/src/spaces/room.rs
stefanceriu: Let's just keep it simple and just call it a SpaceRoom so it matches the spec `A space is defined by the [m.space room type](https://spec.matrix.org/latest/client-server-api/#types), making it known as a “space-room”.`

bindings/matrix-sdk-ffi/src/spaces.rs
stefanceriu: Yes because it's accessed from client.rs but I can change it to `pub(crate)` instead.

bindings/matrix-sdk-ffi/src/spaces.rs
stefanceriu: Seeing as these 2 components are rather independent I would like to keep them separate, at least it's super clear what's what.

bindings/matrix-sdk-ffi/src/spaces.rs
stefanceriu: Nop, good idea 👍

crates/matrix-sdk-ui/src/spaces/mod.rs
stefanceriu: Heh, nice 👌

stefanceriu: Haha, clippy is informing me that there's an even better way `.filter_map(|room| room.is_space().then_some(room))`

pixlwave:
Something weird is going on with the latest version, all of my `joined_spaces` have a `children_count` of `0`.

There's a branch here if you want to test the updates with diffing:
https://github.com/element-hq/element-x-ios/tree/doug/sdk-space-service

stefanceriu:
> Something weird is going on with the latest version, all of my `joined_spaces` have a `children_count` of `0`.
> 
> There's a branch here if you want to test the updates with diffing: https://github.com/element-hq/element-x-ios/tree/doug/sdk-space-service

Went ahead and fixed the remaining issues with VectorDiffs, seems fine now on your branch. Thanks for testing!

crates/matrix-sdk-ui/src/spaces/mod.rs
Hywan: I wonder if we shouldn't add a method on the `Client` directly to get the top-level, joined spaces. `joined_rooms()` is iterating, filtering and collecting all the `Room`s in a `Vec`. Then here, we're iterating, filtering an collecting again.

Alternatively, we could also improve `RoomStateFilter` to include a `SPACE` flag maybe, and then you can use `Client::rooms_filtered`.

crates/matrix-sdk-ui/src/spaces/mod.rs
```diff
@@ -0,0 +1,148 @@
...
+
+    pub fn joined_spaces(&self) -> Vec<OwnedRoomId> {
+        self.client
+            .joined_rooms()
+            .into_iter()
+            .filter_map(|room| if room.is_space() { Some(room.room_id().to_owned()) } else { None })
+            .collect()
+    }
+
+    pub async fn top_level_children_for(
```
Hywan: Documentation is required :-].

crates/matrix-sdk-ui/src/spaces/mod.rs
```diff
@@ -0,0 +1,148 @@
...
+        &self,
+        space_id: OwnedRoomId,
+    ) -> Result<Vec<OwnedRoomId>, Error> {
+        let request = get_hierarchy::v1::Request::new(space_id.clone());
+
+        let result = self.client.send(request).await?;
+
+        println!("Top level children for space {}: {:?}", space_id, result.rooms);
+
+        Ok(vec![])
```
Hywan: I think this is a debug code?

crates/matrix-sdk-ui/src/spaces/mod.rs
```diff
@@ -29,25 +33,29 @@ impl SpaceService {
...
-    pub fn joined_spaces(&self) -> Vec<OwnedRoomId> {
+    pub fn joined_spaces(&self) -> Vec<SpaceServiceRoom> {
         self.client
             .joined_rooms()
             .into_iter()
-            .filter_map(|room| if room.is_space() { Some(room.room_id().to_owned()) } else { None })
-            .collect()
+            .filter_map(|room| if room.is_space() { Some(room) } else { None })
+            .map(|room| SpaceServiceRoom::new_from_known(room))
+            .collect::<Vec<_>>()
```
Hywan: The `::<Vec<_>>` is unnecessary here, as the type is inferred by the method returned type.

crates/matrix-sdk-ui/src/spaces/mod.rs
```diff
@@ -29,25 +33,29 @@ impl SpaceService {
...
 
-    pub fn joined_spaces(&self) -> Vec<OwnedRoomId> {
+    pub fn joined_spaces(&self) -> Vec<SpaceServiceRoom> {
         self.client
             .joined_rooms()
             .into_iter()
-            .filter_map(|room| if room.is_space() { Some(room.room_id().to_owned()) } else { None })
-            .collect()
+            .filter_map(|room| if room.is_space() { Some(room) } else { None })
+            .map(|room| SpaceServiceRoom::new_from_known(room))
```
Hywan: You may want to simplify to:

```suggestion
            .filter_map(|room| room.is_space().then(SpaceServiceRoom::new_from_known))
```

crates/matrix-sdk-ui/src/spaces/mod.rs
```diff
@@ -133,7 +141,7 @@ mod tests {
 
         // All joined
         assert_eq!(
-            space_service.joined_spaces(),
+            space_service.joined_spaces().iter().map(|s| s.room_id.to_owned()).collect::<Vec<_>>(),
```
Hywan: Again, the `::<Vec<_>>` is unnecessary here.

crates/matrix-sdk-ui/src/spaces/mod.rs
```diff
@@ -29,25 +33,29 @@ impl SpaceService {
...
 
-        println!("Top level children for space {}: {:?}", space_id, result.rooms);
-
-        Ok(vec![])
+        Ok(result
+            .rooms
+            .iter()
+            .map(|room| (&room.summary, self.client.get_room(&room.summary.room_id)))
+            .map(|(summary, room)| SpaceServiceRoom::new_from_summary(summary, room))
+            .collect::<Vec<_>>())
```
Hywan: The `::<Vec<_>>` is not necessary here as it is inferred by the method returned type.

crates/matrix-sdk-ui/src/spaces/room.rs
Hywan: Can we name it `Room` please?

crates/matrix-sdk-ui/src/spaces/mod.rs
Hywan, line 1: What about `crates/matrix-sdk-ui/src/space_service/` instead of `spaces/`, so that it matches `room_list_service` and `sync_service`?

crates/matrix-sdk-ui/src/spaces/room.rs
```diff
@@ -0,0 +1,76 @@
...
+            num_joined_members: summary.num_joined_members.into(),
+            join_rule: Some(summary.join_rule.clone()),
+            world_readable: Some(summary.world_readable),
+            guest_can_join: summary.guest_can_join,
+            state: known_room.as_ref().map(|r| r.state()),
+            heroes: known_room.map(|r| r.heroes()),
+        }
+    }
+
+    pub fn new_from_known(known_room: Room) -> Self {
```
Hywan: What about `new_from_room`?

crates/matrix-sdk-ui/src/spaces/room_list.rs
```diff
@@ -0,0 +1,223 @@
...
+use eyeball::{SharedObservable, Subscriber};
+use matrix_sdk::{Client, Error, paginators::PaginationToken};
+use ruma::{OwnedRoomId, api::client::space::get_hierarchy};
+
+use crate::spaces::SpaceServiceRoom;
+
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub enum SpaceServiceRoomListPaginationState {
+    Idle { end_reached: bool },
+    Loading,
```
Hywan, line 34: Comments?

crates/matrix-sdk-ui/src/spaces/room_list.rs
```diff
@@ -0,0 +1,223 @@
...
+    client: Client,
+
+    parent_space_id: OwnedRoomId,
+
+    token: Mutex<PaginationToken>,
+
+    pagination_state: SharedObservable<SpaceServiceRoomListPaginationState>,
+
+    rooms: SharedObservable<Vec<SpaceServiceRoom>>,
+}
```
Hywan: Comments?

crates/matrix-sdk-ui/src/spaces/mod.rs
Hywan: Ah yeah, I didn't see it. It works yes, well done! Or you can use `AbortOnDrop` to make it super explicit and type-checked, which I think is better :-].

crates/matrix-sdk-ui/src/spaces/mod.rs
```diff
@@ -26,37 +29,87 @@ pub mod room_list;
...
+                        error!("error when listening to room updates: {err}");
+                    }
+                }
+            }
+        });
+
+        Self { client, joined_spaces, room_update_handle: handle }
+    }
+
+    pub fn subscribe_to_joined_spaces(&self) -> Subscriber<Vec<SpaceServiceRoom>> {
```
Hywan: Do we want to return the current list of joined spaces, as we usually do with the `subscribe*` methods, i.e. `-> (Vec<Room>, Subscriber<Room>)`?

crates/matrix-sdk-ui/src/spaces/mod.rs
```diff
@@ -17,9 +17,11 @@
 //! See [`SpaceService`] for details.
 
 use eyeball::{SharedObservable, Subscriber};
+use futures_util::pin_mut;
 use matrix_sdk::Client;
 use ruma::OwnedRoomId;
 use tokio::task::JoinHandle;
+use tokio_stream::StreamExt;
```
Hywan: I think it's preferable to pull `StreamExt` from `futures_util` rather than `tokio_stream`.

crates/matrix-sdk-ui/src/spaces/mod.rs
```diff
@@ -84,13 +86,29 @@ impl SpaceService {
...
+    async fn joined_spaces_for(client: &Client) -> Vec<SpaceServiceRoom> {
+        let joined_spaces = client
             .joined_rooms()
             .into_iter()
             .filter_map(|room| if room.is_space() { Some(room) } else { None })
-            .map(SpaceServiceRoom::new_from_known)
-            .collect::<Vec<_>>()
+            .collect::<Vec<_>>();
+
+        let top_level_spaces: Vec<SpaceServiceRoom> = tokio_stream::iter(joined_spaces)
```
Hywan: The type `: Vec<SpaceServiceRoom>` is not necessary here, as it's inferred by the returned type.

Also, the `top_level_spaces` variable is not necessary neither, you can return the whole expression directly.

crates/matrix-sdk-ui/src/spaces/mod.rs
```diff
@@ -84,13 +86,29 @@ impl SpaceService {
...
+            .then(|room| async move {
+                let ok = if let Ok(parents) = room.parent_spaces().await {
+                    pin_mut!(parents);
+                    parents.any(|p| p.is_ok()).await == false
+                } else {
+                    false
+                };
+                (room, ok)
```
Hywan: What are you trying to do here? The `ok` refers to the “back-link”, like a space has a link to the room (from parent to child), and the room has a link to the space (from child to parent)?

I think it deserves a log when the child has no link to the parent, it's an important information.

Also, to remove the need to use `tokio_stream`, you can do:

```rust
let mut top_level_spaces = Vec::new();

for room in joined_spaces {
    if let Ok(parents) = room.parent_spaces().await {
        // do checks

        if everything_is_alright {
            top_level_spaces.push(room);
        }
    } else {
        warn!("…");
    }
}
```

crates/matrix-sdk-ui/src/spaces/room_list.rs
```diff
@@ -36,18 +39,60 @@ pub struct SpaceServiceRoomList {
...
 }
 
+impl Drop for SpaceServiceRoomList {
+    fn drop(&mut self) {
+        self.room_update_handle.abort();
+    }
+}
 impl SpaceServiceRoomList {
     pub fn new(client: Client, parent_space_id: OwnedRoomId) -> Self {
+        let rooms = SharedObservable::new(Vec::<SpaceServiceRoom>::new());
```
Hywan: Pretty sure you can remove `::<SpaceServiceRoom>` part after `Vec`.

crates/matrix-sdk-ui/src/spaces/room_list.rs
```diff
@@ -36,18 +39,60 @@ pub struct SpaceServiceRoomList {
...
+                        if new_rooms != rooms_clone.get() {
+                            rooms_clone.set(new_rooms);
+                        }
+                    }
+                    Err(err) => {
+                        error!("error when listening to room updates: {err}");
+                    }
+                }
+            }
+        });
```
Hywan: I wonder if you couldn't use `Client::rooms_stream` with filters instead of having all these tasks and code logic 🤔.

Hywan (changes requested):
I've reviewed up to the graph thingy. Is it possible to squash commits that are updating the same part of the code, they are mostly fixup at this point and it would simplifiy the review process.

I'm also wondering if using `Client::rooms_stream` could remove the need for some tasks, what do you think?

crates/matrix-sdk-ui/src/spaces/mod.rs
stefanceriu: Oh nice, thank you! Took care of it in 0f16eea7121a2707c4ca6ac4681edbceca35aafc

crates/matrix-sdk-ui/src/spaces/mod.rs
stefanceriu: This has since been replaced by `joined_spaces` which is also documented

crates/matrix-sdk-ui/src/spaces/room.rs
stefanceriu: Overall naming has been adjusted in 511c975ca2819f340593511f970c504b018bb37f

crates/matrix-sdk-ui/src/spaces/mod.rs
stefanceriu: I would expect `spaces` to grow and potentiall encompass more than just this initial service. Figured it's safer to have something that can group all space related functionality instead of narrowing down its scope too soon.

crates/matrix-sdk-ui/src/spaces/room.rs
stefanceriu: I wanted to make it clear that it's a Room already known by the client as opposed to any of the other Room types we have. I don't really have a problem changing it if you think that's already clear enough.

crates/matrix-sdk-ui/src/spaces/mod.rs
stefanceriu: It is, with this being the initial commit. I'm happy to squash these if you think evolution history isn't relevant.

crates/matrix-sdk-ui/src/spaces/room_list.rs
stefanceriu: With the APIs being very fluid and VectorDiffs being introduced todawards the end, I intentionally left documentation as one of the last steps. Handled in c6a25e99ab399139d716c76f156931bea3bbec05

crates/matrix-sdk-ui/src/spaces/room_list.rs
stefanceriu: `SharedObservable`s have since been replaced with `ObservableVector`s and this is no longer an issue.

crates/matrix-sdk-ui/src/spaces/room_list.rs
stefanceriu: Talked about if offline: `rooms_stream` will trigger only when rooms get left or joined and not when any of their data changes.

crates/matrix-sdk-ui/src/spaces/mod.rs
stefanceriu: Sorted 👍

crates/matrix-sdk-ui/src/spaces/mod.rs
stefanceriu: Oh, I didn't realise that, thanks!

crates/matrix-sdk-ui/src/spaces/mod.rs
stefanceriu: Yes, that's what the version of the `ObservableVector` API introduced in 52ee21cd0a58bf324e84669d587f7893f9a33c20 does too.

crates/matrix-sdk-ui/src/spaces/mod.rs
stefanceriu: This was trying to figure out what the top level parent spaces are, the spaces that don't have any parents of their own. It turned out to be the wrong approach as child to parent or parent to child relations aren't enforced so it can result in wrong results. It was replaced with a proper graph structure that fills in missing edges in b2ea99cba214e2f31d301b2e2e7d72aa05a9826a

crates/matrix-sdk-ui/src/spaces/mod.rs
stefanceriu: Fair point. I think adding a `joined_space_rooms` method is the most straight forward solution, changing `RoomStateFilter` to do more than membership feels a bit too intrusive. One note though, it will return *any* joined space room, not the top level ones. We need the graph and what this service provides to compute those.

crates/matrix-sdk-ui/src/spaces/mod.rs
stefanceriu: Another one for the toolbelt, thanks! 👍

bindings/matrix-sdk-ffi/src/spaces.rs
```diff
@@ -0,0 +1,232 @@
...
+    room::{Membership, RoomHero},
+    room_preview::RoomType,
+    runtime::get_runtime_handle,
+    TaskHandle,
+};
+
+#[derive(uniffi::Object)]
+pub struct SpaceService {
+    inner: UISpaceService,
+}
```
poljar, line 43: Are docs on the struct not useful on the FFI side?

bindings/matrix-sdk-ffi/src/spaces.rs
```diff
@@ -0,0 +1,232 @@
...
+    }
+}
+
+#[matrix_sdk_ffi_macros::export]
+impl SpaceService {
+    pub async fn joined_spaces(&self) -> Vec<SpaceRoom> {
+        self.inner.joined_spaces().await.into_iter().map(Into::into).collect()
+    }
+
+    pub async fn subscribe_to_joined_spaces(
```
poljar, line 63: Docs?

bindings/matrix-sdk-ffi/src/spaces.rs
```diff
@@ -0,0 +1,232 @@
...
+            values: initial_values.into_iter().map(Into::into).collect(),
+        }]);
+
+        Arc::new(TaskHandle::new(get_runtime_handle().spawn(async move {
+            while let Some(diffs) = stream.next().await {
+                listener.on_update(diffs.into_iter().map(Into::into).collect());
+            }
+        })))
+    }
+    #[allow(clippy::unused_async)]
```
poljar, line 81: It looks a bit strange to me when this `#[allow()]` is above some comments, let's put it above the function definition.

bindings/matrix-sdk-ffi/src/spaces.rs
```diff
@@ -0,0 +1,232 @@
...
+    // This method doesn't need to be async but if its not the FFI layer panics
+    // with "there is no no reactor running, must be called from the context
+    // of a Tokio 1.x runtime" error because the underlying constructor spawns
+    // an async task.
+    pub async fn space_room_list(
```
poljar, line 86: Not sure how this will look but let's also document this.

bindings/matrix-sdk-ffi/src/spaces.rs
```diff
@@ -0,0 +1,232 @@
...
+    ) -> Result<Arc<SpaceRoomList>, ClientError> {
+        let space_id = RoomId::parse(space_id)?;
+        Ok(Arc::new(SpaceRoomList::new(self.inner.space_room_list(space_id))))
+    }
+}
+
+#[derive(uniffi::Object)]
+pub struct SpaceRoomList {
+    inner: UISpaceRoomList,
+}
```
poljar, line 107: I wont' repeat much further, but we need to document the FFI layer as well.

bindings/matrix-sdk-ffi/src/sync_service.rs
```diff
@@ -90,6 +90,15 @@ impl SyncService {
...
     }
+
+    /// Force expiring both sliding sync sessions.
+    ///
+    /// This ensures that the sync service is stopped before expiring both
+    /// sessions. It should be used sparingly, as it will cause a restart of
+    /// the sessions on the server as well.
+    pub async fn expire_sessions(&self) {
+        self.inner.expire_sessions().await;
+    }
```
poljar: Let's put this into a separate PR. Functionally-wise it's separate from the rest of the PR.

testing/matrix-sdk-test/src/event_factory.rs
```diff
@@ -871,9 +873,11 @@ impl EventFactory {
         &self,
         creator_user_id: &UserId,
         room_version: RoomVersionId,
+        room_type: Option<RoomType>,
```
poljar: Instead of adding this argument here, we should add a new method here:

https://github.com/matrix-org/matrix-rust-sdk/blob/3aeface647183345973e310930fa037bc603db04/testing/matrix-sdk-test/src/event_factory.rs#L453-L465

This removes the need to modify *every* call site of this method which in turn makes the PR smaller. This change could have been a separate PR as well.

crates/matrix-sdk-ui/Cargo.toml
```diff
@@ -67,6 +67,7 @@ assert-json-diff.workspace = true
 assert_matches.workspace = true
 assert_matches2.workspace = true
 eyeball-im-util.workspace = true
+futures-executor.workspace = true
```
poljar: What's this for?

crates/matrix-sdk-ui/src/spaces/mod.rs
```diff
@@ -0,0 +1,510 @@
...
+///
+/// # Examples
+///
+/// ```no_run
+/// use futures_util::StreamExt;
+/// use matrix_sdk::Client;
+/// use matrix_sdk_ui::spaces::SpaceService;
+/// use ruma::owned_room_id;
+///
+/// # futures_executor::block_on(async {
```
poljar: Oh, it's for this? Since the example is `no_run` you don't really need this, having the `async` block is enough, for example like here: https://github.com/matrix-org/matrix-rust-sdk/blob/3aeface647183345973e310930fa037bc603db04/crates/matrix-sdk-ui/src/sync_service.rs#L572-L603.

testing/matrix-sdk-test/src/event_factory.rs
```diff
@@ -987,6 +991,30 @@ impl EventFactory {
...
+        &self,
+        parent: OwnedRoomId,
+        child: OwnedRoomId,
+    ) -> EventBuilder<SpaceParentEventContent> {
+        let mut event = self.event(SpaceParentEventContent::new(vec![]));
+        event.state_key = Some(parent.to_string());
+        event.room = Some(child);
+        event
+    }
+
```
poljar: Let us pack this into the same PR as the rest of the event factory changes.

crates/matrix-sdk-base/src/sync.rs
```diff
@@ -90,13 +90,22 @@ impl RoomUpdates {
...
     }
+
+    /// Returns whether or not this update contains any changes to the list
+    /// of invited, joined, knocked or left rooms.
+    pub fn is_empty(&self) -> bool {
+        self.invited.is_empty()
+            && self.joined.is_empty()
+            && self.knocked.is_empty()
+            && self.left.is_empty()
+    }
```
poljar: We can put this into a separate PR, those methods are useful on their own.

Don't forget a changelog entry.

crates/matrix-sdk/src/client/mod.rs
```diff
@@ -1235,6 +1235,15 @@ impl Client {
...
 
+    /// Returns the joined space rooms this client knows about.
+    pub fn joined_space_rooms(&self) -> Vec<Room> {
+        self.base_client()
+            .rooms_filtered(RoomStateFilter::JOINED)
+            .into_iter()
+            .flat_map(|room| room.is_space().then_some(Room::new(self.clone(), room)))
+            .collect()
+    }
+
```
poljar: Can be a separate PR, with a changelog.

crates/matrix-sdk/src/room/mod.rs
```diff
@@ -1166,7 +1166,7 @@ impl Room {
         // Implements this algorithm:
         // https://spec.matrix.org/v1.8/client-server-api/#mspaceparent-relationships
 
-        // Get all m.room.parent events for this room
+        // Get all m.space.parent events for this room
```
poljar: These small comment fixes could also go into a separate PR.

crates/matrix-sdk-ui/src/spaces/mod.rs
stefanceriu: Ah sorry, this was supposed to work without `no_run` through that `futures-executor` but it ends up just failing miseraby with something along the lines of:

```
% cargo test spaces -p matrix-sdk-ui --doc

test crates/matrix-sdk-ui/src/spaces/mod.rs - spaces::SpaceService (line 60) ... FAILED

failures:

---- crates/matrix-sdk-ui/src/spaces/mod.rs - spaces::SpaceService (line 60) stdout ----
Test executable failed (exit status: 101).

stderr:

thread 'main' (14505151) panicked at /var/folders/xy/jh61h2x979g3r7c5rglgd_800000gn/T/rustdoctestbJlRWS/doctest_bundle_2024.rs:48:22:
not yet implemented
```

poljar: We discussed this in chat. Using `no_run` is completely fine, we don't want our examples to try to reach real homeservers anyways.

testing/matrix-sdk-test/src/event_factory.rs
stefanceriu: Oh this is neat, didn't realise it was a thing! 👍

stefanceriu:
Status update:
* I've extracted several small PRs from this one in #5592, #5593, #5594, #5595 and #5596
* with those merged onto `main` I went ahead and rebased this PR on top of it and cleaned up the history

crates/matrix-sdk-ui/src/spaces/graph.rs
```diff
@@ -0,0 +1,217 @@
...
+    children: BTreeSet<OwnedRoomId>,
+}
+
+impl SpaceGraphNode {
+    fn new(id: OwnedRoomId) -> Self {
+        Self { id, parents: BTreeSet::new(), children: BTreeSet::new() }
+    }
+}
+
+#[derive(Debug)]
```
poljar: Derive goes right at the top of the function definition, in other words this should go after the docs.

crates/matrix-sdk-ui/src/spaces/graph.rs
```diff
@@ -0,0 +1,217 @@
...
+}
+
+impl Default for SpaceGraph {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+impl SpaceGraph {
+    pub fn new() -> Self {
```
poljar: Docs please. A single sentence is enough but something has to be here since it's pub.

crates/matrix-sdk-ui/src/spaces/graph.rs
```diff
@@ -0,0 +1,217 @@
...
+
+    /// Adds a directed edge from `parent_id` to `child_id`, creating nodes if
+    /// they do not already exist in the graph.
+    pub fn add_edge(&mut self, parent_id: OwnedRoomId, child_id: OwnedRoomId) {
+        self.nodes.entry(parent_id.clone()).or_insert(SpaceGraphNode::new(parent_id.clone()));
+
+        self.nodes.entry(child_id.clone()).or_insert(SpaceGraphNode::new(child_id.clone()));
+
+        self.nodes.get_mut(&parent_id).unwrap().children.insert(child_id.clone());
+        self.nodes.get_mut(&child_id).unwrap().parents.insert(parent_id);
```
poljar: These unwraps, while trivially safe, aren't necessary. We should use the entry to insert things into the node, after all that's what the entry is for.

In Rust this would look like:

```rust
let parent_entry =
self.nodes.entry(parent_id.clone()).or_insert(SpaceGraphNode::new(parent_id.clone()));
parent_entry.children.insert(child_id.clone());

let child_entry =
self.nodes.entry(child_id.clone()).or_insert(SpaceGraphNode::new(child_id));
child_entry.parents.insert(parent_id);
```

crates/matrix-sdk-ui/src/spaces/graph.rs
```diff
@@ -0,0 +1,217 @@
...
+    }
+}
+
+#[derive(Debug)]
+/// A graph structure representing a space hierarchy. Contains functionality
+/// for mapping parent-child relationships between rooms, removing cycles and
+/// retrieving top-level parents/roots.
+pub struct SpaceGraph {
+    nodes: BTreeMap<OwnedRoomId, SpaceGraphNode>,
+}
```
poljar: I wonder if we could have used a library for this, but perhaps adding a dependency is not worth the cost considering how little code we have here.

crates/matrix-sdk-ui/src/spaces/graph.rs
```diff
@@ -0,0 +1,217 @@
...
+        self.nodes.entry(child_id.clone()).or_insert(SpaceGraphNode::new(child_id.clone()));
+
+        self.nodes.get_mut(&parent_id).unwrap().children.insert(child_id.clone());
+        self.nodes.get_mut(&child_id).unwrap().parents.insert(parent_id);
+    }
+
+    /// Removes cycles in the graph by performing a depth-first search (DFS) and
+    /// remembering the visited nodes. If a node is revisited while still in the
+    /// current path (i.e. it's on the stack), it indicates a cycle.
+    pub fn remove_cycles(&mut self) {
```
poljar: This is pub because it's too expensive to call at each `add_edge()` call right?

I wonder if a better API, where we can't forget to call this, would have been a bulk `add_edges()` API where it then becomes feasible to call this after all edges were added.

crates/matrix-sdk-ui/src/spaces/room.rs
```diff
@@ -0,0 +1,86 @@
...
+    pub avatar_url: Option<OwnedMxcUri>,
+    pub room_type: Option<RoomType>,
+    pub num_joined_members: u64,
+    pub join_rule: Option<JoinRuleSummary>,
+    pub world_readable: Option<bool>,
+    pub guest_can_join: bool,
+
+    pub children_count: u64,
+    pub state: Option<RoomState>,
+    pub heroes: Option<Vec<RoomHero>>,
```
poljar: These fields, while mostly a copy paste from the other Room struct we have, need some documentation.

crates/matrix-sdk-ui/src/spaces/room.rs
```diff
@@ -0,0 +1,86 @@
...
+}
+
+impl SpaceRoom {
+    pub fn new_from_summary(
```
poljar: Let's describe what this does as well. Also, does this need to be pub or could it be `pub(crate)`? I don't think we expect users to create a `SpaceRoom` manually, no?

crates/matrix-sdk-ui/src/spaces/room.rs
```diff
@@ -0,0 +1,86 @@
...
+            join_rule: Some(summary.join_rule.clone()),
+            world_readable: Some(summary.world_readable),
+            guest_can_join: summary.guest_can_join,
+            children_count,
+            state: known_room.as_ref().map(|r| r.state()),
+            heroes: known_room.map(|r| r.heroes()),
+        }
+    }
+
+    pub fn new_from_known(known_room: Room, children_count: u64) -> Self {
```
poljar: Same as for the other constructor.

crates/matrix-sdk-ui/src/spaces/mod.rs
poljar: I think the reasoning for the different naming scheme makes sense.

crates/matrix-sdk-ui/src/spaces/mod.rs
```diff
@@ -0,0 +1,491 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for that specific language governing permissions and
+// limitations under the License.
+
+//! High level interfaces for working with Spaces
+//!
+//! The `SpaceService` is an UI oriented, high-level interface for working with
+//! Matrix Spaces. It provides methods to retrieve joined spaces, subscribe
+//! to updates, and navigate space hierarchies.
```
poljar: Could we link to the spec so people can discover what `Matrix Spaces` are?

crates/matrix-sdk-ui/src/spaces/mod.rs
```diff
@@ -0,0 +1,491 @@
...
+//!
+//! It consists of 3 main components:
+//! - `SpaceService`: The main service for managing spaces. It
+//! - `SpaceGraph`: An utility that maps the `m.space.parent` and
+//!   `m.space.child` fields into a graph structure, removing cycles and
+//!   providing access to top level parents.
```
poljar, line 26: Looking at the implementation of the service, it seems that the `SpaceGraph` is just an implementation detail?

Can't we hide the `SpaceGraph` from our users and the docs?

crates/matrix-sdk-ui/src/spaces/room_list.rs
```diff
@@ -0,0 +1,406 @@
...
+    pagination_state: SharedObservable<SpaceRoomListPaginationState>,
+
+    rooms: Arc<Mutex<ObservableVector<SpaceRoom>>>,
+
+    _room_update_handle: AbortOnDrop<()>,
+}
+
+impl SpaceRoomList {
+    /// Creates a new `SpaceRoomList` for the given space identifier.
+    pub fn new(client: Client, parent_space_id: OwnedRoomId) -> Self {
```
poljar, line 113: Can we make this a bit more type safe? What happens with this if the space ID isn't a space?

Can't we have something like `SpaceRoom::room_list() -> SpaceRoomList`? Would that be annoying or slow to use on the FFI?

crates/matrix-sdk-ui/src/spaces/mod.rs
```diff
@@ -0,0 +1,491 @@
...
+/// # Examples
+///
+/// ```no_run
+/// use futures_util::StreamExt;
+/// use matrix_sdk::Client;
+/// use matrix_sdk_ui::spaces::SpaceService;
+/// use ruma::owned_room_id;
+///
+/// # async {
+/// let client: Client = todo!();
```
poljar: We should probably hide this, if people got this far with the SDK they know how to create a `Client`.

```suggestion
/// # let client: Client = todo!();
```

crates/matrix-sdk-ui/src/spaces/mod.rs
```diff
@@ -0,0 +1,491 @@
...
+pub struct SpaceService {
+    client: Client,
+
+    joined_spaces: Arc<Mutex<ObservableVector<SpaceRoom>>>,
+
+    room_update_handle: Mutex<Option<AbortOnDrop<()>>>,
+}
+
+impl SpaceService {
+    pub fn new(client: Client) -> Self {
```
poljar, line 104: Doc please.

crates/matrix-sdk-ui/src/spaces/mod.rs
```diff
@@ -0,0 +1,491 @@
...
+    /// joined or left, the stream will yield diffs that reflect the changes.
+    pub async fn subscribe_to_joined_spaces(
+        &self,
+    ) -> (Vector<SpaceRoom>, VectorSubscriberBatchedStream<SpaceRoom>) {
+        if self.room_update_handle.lock().is_none() {
+            let client = self.client.clone();
+            let joined_spaces = Arc::clone(&self.joined_spaces);
+            let all_room_updates_receiver = self.client.subscribe_to_all_room_updates();
+
+            *self.room_update_handle.lock() = Some(AbortOnDrop::new(spawn(async move {
```
poljar: While unlikely there's a race here between the first time you take the lock for the `is_none()` check and this line.

We should take the lock only once here:

```rust
let guard = self.room_update_handle.lock();

if guard.is_none() {
    ...
} else {
    ...
}
```

crates/matrix-sdk-ui/src/spaces/mod.rs
```diff
@@ -0,0 +1,491 @@
...
+        let mut graph = SpaceGraph::new();
+
+        // Iterate over all joined spaces and populate the graph with edges based
+        // on `m.space.parent` and `m.space.child` state events.
+        for space in joined_spaces.iter() {
+            graph.add_node(space.room_id().to_owned());
+
+            if let Ok(parents) = space.get_state_events_static::<SpaceParentEventContent>().await {
+                parents.into_iter()
+                .flat_map(|parent_event| match parent_event.deserialize() {
```
poljar, line 193: Doesn't make a difference result-wise, but I think you're filtering here. You get all the parents but filter out invalid ones. So `filter_map` sounds like the correct one to use.
```suggestion
                .filter_map(|parent_event| match parent_event.deserialize() {
```

crates/matrix-sdk-ui/src/spaces/mod.rs
```diff
@@ -0,0 +1,491 @@
...
+                        None
+                    }
+                }).for_each(|parent| graph.add_edge(parent, space.room_id().to_owned()));
+            } else {
+                error!(room_id = ?space.room_id(), "Could not get m.space.parent events");
+            }
+
+            if let Ok(children) = space.get_state_events_static::<SpaceChildEventContent>().await {
+                children.into_iter()
+                .flat_map(|child_event| match child_event.deserialize() {
```
poljar:
```suggestion
                .filter_map(|child_event| match child_event.deserialize() {
```

crates/matrix-sdk-ui/src/spaces/mod.rs
```diff
@@ -0,0 +1,491 @@
...
+                    }
+                }).for_each(|child| graph.add_edge(space.room_id().to_owned(), child));
+            } else {
+                error!(room_id = ?space.room_id(), "Could not get m.space.child events");
+            }
+        }
+
+        // Remove cycles from the graph. This is important because they are not
+        // enforced backend side.
+        graph.remove_cycles();
```
poljar, line 228: I guess it isn't too bad that we have to call this manually since we need to do so in a single place, but it seems that a bulk API would have been possible.

crates/matrix-sdk-ui/src/spaces/mod.rs
```diff
@@ -0,0 +1,491 @@
...
+
+        let root_notes = graph.root_nodes();
```
poljar: wat :duck: 

```suggestion
        let root_nodes = graph.root_nodes();
```

crates/matrix-sdk-ui/src/spaces/mod.rs
```diff
@@ -0,0 +1,491 @@
...
+
+        joined_spaces
+            .iter()
+            .flat_map(|room| {
+                let room_id = room.room_id().to_owned();
+
+                if root_notes.contains(&&room_id) {
```
poljar:
```suggestion
                if root_nodes.contains(&&room_id) {
```

crates/matrix-sdk-ui/src/spaces/graph.rs
```diff
@@ -0,0 +1,217 @@
...
+
+impl SpaceGraph {
+    pub fn new() -> Self {
+        Self { nodes: BTreeMap::new() }
+    }
+
+    /// Returns the root nodes of the graph, which are nodes without any
+    /// parents.
+    pub fn root_nodes(&self) -> Vec<&OwnedRoomId> {
+        self.nodes.values().filter(|node| node.parents.is_empty()).map(|node| &node.id).collect()
```
poljar: I see, the borrow type made life a bit painful for you.

```suggestion
    pub fn root_nodes(&self) -> Vec<&RoomId> {
        self.nodes.values().filter(|node| node.parents.is_empty()).map(|node| node.id.as_ref()).collect()
```

crates/matrix-sdk-ui/src/spaces/graph.rs
```diff
@@ -0,0 +1,217 @@
...
+    }
+
+    /// Returns the children of a given node. If the node does not exist, it
+    /// returns an empty vector.
+    pub fn children_of(&self, node_id: &OwnedRoomId) -> Vec<&OwnedRoomId> {
+        self.nodes.get(node_id).map_or(vec![], |node| node.children.iter().collect())
+    }
```
poljar:
```suggestion
    pub fn children_of(&self, node_id: &RoomId) -> Vec<&RoomId> {
        self.nodes.get(node_id).map_or(vec![], |node| node.children.iter().map(|id| id.as_ref()).collect())
    }
```

crates/matrix-sdk-ui/src/spaces/mod.rs
poljar: I thought that this looks strange, because you pass a double ref further bellow. So I went back to look at the graph implementation and added some more suggestions to it.

If you do the changes in the graph implementation to work with `&RoomId` then this can look as follows:

```suggestion
                let room_id = room.room_id();

                if root_notes.contains(&room_id) {
```

crates/matrix-sdk-ui/src/spaces/mod.rs
```diff
@@ -0,0 +1,491 @@
...
+
+        // Remove cycles from the graph. This is important because they are not
+        // enforced backend side.
+        graph.remove_cycles();
+
+        let root_notes = graph.root_nodes();
+
+        joined_spaces
+            .iter()
+            .flat_map(|room| {
```
poljar: That's also a `filter_map()` as far as I can tell.
```suggestion
            .filter_map(|room| {
```

crates/matrix-sdk-ui/src/spaces/room_list.rs
```diff
@@ -0,0 +1,406 @@
...
+/// );
+///
+/// // Subscribe to pagination state updates
+/// let pagination_state_stream =
+///     room_list.subscribe_to_pagination_state_updates();
+///
+/// // And to room list updates
+/// let (_, room_stream) = room_list.subscribe_to_room_updates();
+///
+/// // spawn {
```
poljar: What are those spawns for?

crates/matrix-sdk-ui/src/spaces/room_list.rs
```diff
@@ -0,0 +1,406 @@
...
+/// while let Some(pagination_state) = pagination_state_stream.next().await {
+///     println!("Received pagination state update: {pagination_state:?}");
+/// }
+/// // }
+///
+/// // spawn {
+/// while let Some(diffs) = room_stream.next().await {
+///     println!("Received room list update: {diffs:?}");
+/// }
+/// // }
```
poljar: Oh, these are supposed to be in a background task? Perhaps we should just comment as such if you don't want to showcase how to spawn tasks for real.

The way it is it might just confuse people because this wouldn't compile if uncommented.

crates/matrix-sdk-ui/src/spaces/room_list.rs
```diff
@@ -0,0 +1,406 @@
...
+    _room_update_handle: AbortOnDrop<()>,
+}
+
+impl SpaceRoomList {
+    /// Creates a new `SpaceRoomList` for the given space identifier.
+    pub fn new(client: Client, parent_space_id: OwnedRoomId) -> Self {
+        let rooms = Arc::new(Mutex::new(ObservableVector::<SpaceRoom>::new()));
+
+        let client_clone = client.clone();
+        let rooms_clone = rooms.clone();
```
poljar: Not important, but you can get rid of those annoying names using this trick:

```rust
let handle = spawn({
    let room = room.clone();
    let client = client.clone();
    async move {
        ...
    }
});
```

The clones don't destroy the original variables because they are inside a subscope.

crates/matrix-sdk-ui/src/spaces/room_list.rs
```diff
@@ -0,0 +1,406 @@
...
+                        }
+
+                        let mut mutable_rooms = rooms_clone.lock();
+
+                        updates.iter_all_room_ids().for_each(|updated_room_id| {
+                            if let Some((position, room)) = mutable_rooms
+                                .clone()
+                                .iter()
+                                .find_position(|room| &room.room_id == updated_room_id)
+                                && let Some(update_room) = client_clone.get_room(updated_room_id)
```
poljar: Look at you fancy pants, using let chains.

crates/matrix-sdk-ui/src/spaces/room_list.rs
```diff
@@ -0,0 +1,406 @@
...
+    pub async fn paginate(&self) -> Result<(), Error> {
+        match *self.pagination_state.read() {
+            SpaceRoomListPaginationState::Idle { end_reached } if end_reached => {
+                return Ok(());
+            }
+            SpaceRoomListPaginationState::Loading => {
+                return Ok(());
+            }
+            _ => {}
+        }
```
poljar: Don't we have a race here again?

Between the read and the set the state could have changed under you. I think you'll need to use `pagination_state.set_if_not_eq()`

I think this one might not even be this theoretical, since users might spam a button or scroll gestures.

crates/matrix-sdk-ui/src/spaces/room_list.rs
```diff
@@ -0,0 +1,406 @@
...
+
+        self.pagination_state.set(SpaceRoomListPaginationState::Loading);
+
+        let mut request = get_hierarchy::v1::Request::new(self.parent_space_id.clone());
+        request.max_depth = Some(uint!(1)); // We only want the immediate children of the space
+
+        if let PaginationToken::HasMore(ref token) = *self.token.lock() {
+            request.from = Some(token.clone());
+        }
```
poljar: Can't this be an async lock we hold on to for the duration of the request?

The `pagination_state` already protect us from two concurrent calls to this function, but it might still be a neater design to hold on to the token lock while it is used/modified.

crates/matrix-sdk-ui/src/spaces/room_list.rs
```diff
@@ -0,0 +1,406 @@
...
+                *token = match &result.next_batch {
+                    Some(val) => PaginationToken::HasMore(val.clone()),
+                    None => PaginationToken::HitEnd,
+                };
+
+                let mut rooms = self.rooms.lock();
+                result
+                    .rooms
+                    .iter()
+                    .flat_map(|room| {
```
poljar:
```suggestion
                    .filter_map(|room| {
```

poljar (changes requested):
Alright, there are a couple of improvements that we should make, and a bunch of smaller nits, but overall this looks great.

Thank you again for splitting it up a bit, there's much less noise in the PR now.

crates/matrix-sdk-ui/src/spaces/room_list.rs
stefanceriu: Now that's cool!

crates/matrix-sdk-ui/src/spaces/room_list.rs
stefanceriu: Don't think so, that should be absolutely fine but I do wonder if it should be a result instead of an optional 🤔

poljar: Hmm, but it isn't an optional?

stefanceriu: It would be if I apply your suggestion?

poljar: Oh, you mean return `None` if the space ID isn't a space? Or which suggestion? I didn't post a github code suggestion here, am I growing senile? :older_man:

stefanceriu: Lol, nop, it's not you, it's me. I completely missread your original comment as wanting to return `SpaceRoomList?` from that existing constructor to which figured we might as well return a full result.

Yeah, the `SpaceRoom` is supposed to be a lightweight record that powers items in the rendered list, adding a client to them so we can build the RoomList might be a bit too much.

poljar: Right, makes sense.

crates/matrix-sdk-ui/src/spaces/graph.rs
stefanceriu: I thought about it but as this rather simple at the moment and an implementation detail of the spaces service I considered it not worth indeed.

crates/matrix-sdk-ui/src/spaces/room.rs
stefanceriu: Good point, yes 👍

crates/matrix-sdk-ui/src/spaces/mod.rs
stefanceriu: And by that do you mean making it all `pub(super)`? 🤔

crates/matrix-sdk-ui/src/spaces/room_list.rs
stefanceriu: Ups, sure thing 👍

crates/matrix-sdk-ui/src/spaces/mod.rs
poljar: Yes :sweat_smile:

crates/matrix-sdk-ui/src/spaces/mod.rs
stefanceriu: #nocomment 😅

crates/matrix-sdk-ui/src/spaces/graph.rs
stefanceriu: I can potentially be expensive, yes. That sounds like a good idea yes although from an usability point of view it might be a bit more annoying as they would all need to be collected and mapped beforehand.

crates/matrix-sdk-ui/src/spaces/mod.rs
stefanceriu: Yeah, let's see how this evolves, we might end up just replacing the whole graph component.

crates/matrix-sdk-ui/src/spaces/mod.rs
stefanceriu: 🤦

crates/matrix-sdk-ui/src/spaces/graph.rs
stefanceriu: Definitely, nice one! 👍

crates/matrix-sdk-ui/src/spaces/mod.rs
stefanceriu: Noice, thank you! 🙏

crates/matrix-sdk-ui/src/spaces/room_list.rs
stefanceriu: Nice catch! I would've expected the SharedObservable to not notify in case the value doesn't change 🤔

crates/matrix-sdk-ui/src/spaces/room_list.rs
stefanceriu: TIL and sounds like a good idea to me 👍

crates/matrix-sdk-ui/src/spaces/mod.rs
stefanceriu: Nice catch, thanks!

crates/matrix-sdk-ui/src/spaces/room_list.rs
```diff
@@ -201,7 +201,7 @@ impl SpaceRoomList {
             _ => {}
         }
 
-        self.pagination_state.set(SpaceRoomListPaginationState::Loading);
+        self.pagination_state.set_if_not_eq(SpaceRoomListPaginationState::Loading);
```
poljar: This isn't exactly what I meant :sweat_smile:

You need to replace the `read` + `set` with something that is atomic. It's a classic [Time-of-check to time-of-use](https://en.wikipedia.org/wiki/Time-of-check_to_time-of-use) issue.

Rereading what `set_if_not_eq()` does, I think my suggestion was wrong.

Taking a `write()` lock instead might be the proper way to do this.

poljar (changes requested):
Alright, almost good to go.

There's one thing that I miscommunication a bit and even gave you an incorrect suggestion, sorry about that.

Let me know if you are out of time and want me to fix this myself.

crates/matrix-sdk-ui/src/spaces/room_list.rs
stefanceriu: Talked about if offline, makes sense, made the change 👍

poljar: “Offline” :sweat_smile:

poljar (approved):
Nice, I think this is good to go.

Congrats, now you have your own personal subsystem in the SDK that you have the privilege to maintain *forever* :scream:

stefanceriu:
>Congrats, now you have your own personal subsystem in the SDK that you have the privilege to maintain forever 😱

Wooo, thank you, what an absolute priviledge! 😁

PR #5510, opened 2025-08-08
Check the sender data before we accept historic room key bundles

poljar:
This closes: https://github.com/matrix-org/matrix-rust-sdk/issues/5139.

The solution is a bit different from what #5139 describes, here we do the check before we accept the bundle, not before we download it.

This is done to avoid the duplication the Rust implementation and js-sdk implementation of shared historic bundles would need to do in case the check is done before the download. Furthermore it makes for a simpler API, implementers can't accidentally forgot to do this check.

### DIFF
--- a/crates/matrix-sdk-crypto/src/store/mod.rs
+++ b/crates/matrix-sdk-crypto/src/store/mod.rs
@@ -60,7 +60,7 @@ use thiserror::Error;
 use tokio::sync::{Mutex, Notify, OwnedRwLockWriteGuard, RwLock};
 use tokio_stream::wrappers::errors::BroadcastStreamRecvError;
 use tracing::{error, info, instrument, trace, warn};
-use types::RoomKeyBundleInfo;
+use types::{RoomKeyBundleInfo, StoredRoomKeyBundleData};
 use vodozemac::{megolm::SessionOrdering, Curve25519PublicKey};
 
 use self::types::{
@@ -1341,14 +1341,11 @@ impl Store {
     ///
     /// while let Some(bundle_info) = bundle_stream.next().await {
     ///     // Try to find the bundle content in the store and if it's valid accept it.
-    ///     if let Some(bundle_content) = machine.store().get_received_room_key_bundle_data(&bundle_info.room_id, &bundle_info.sender).await? {
-    ///         let StoredRoomKeyBundleData { sender_user, sender_data, bundle_data, .. } = bundle_content;
+    ///     if let Some(bundle_data) = machine.store().get_received_room_key_bundle_data(&bundle_info.room_id, &bundle_info.sender).await? {
     ///         // Download the bundle now and import it.
     ///         let bundle: RoomKeyBundle = todo!("Download the bundle");
     ///         machine.store().receive_room_key_bundle(
-    ///             &bundle_info.room_id,
-    ///             &sender_user,
-    ///             &sender_data,
+    ///             &bundle_data,
     ///             bundle,
     ///             |_, _| {},
     ///         ).await?;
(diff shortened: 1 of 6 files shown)


testing/matrix-sdk-integration-testing/src/tests/e2ee/shared_history.rs
```diff
@@ -128,3 +133,195 @@ async fn test_history_share_on_invite() -> Result<()> {
...
+        .encryption()
+        .get_user_identity(alice.user_id().unwrap())
+        .await?
+        .expect("Bob should have access to Alice's identity");
+
+    info!("Bob is verifying Alice");
+
+    alice_identity.verify().await?;
+
+    // Now Alice creates a room.
```
andybalaam: I think this comment should not be here - it is repeated below.

testing/matrix-sdk-integration-testing/src/tests/e2ee/shared_history.rs
```diff
@@ -128,3 +133,195 @@ async fn test_history_share_on_invite() -> Result<()> {
...
+        .await
+        .expect("We should be able to get the bundle stream");
+
+    let identity_stream = bob
+        .encryption()
+        .user_identities_stream()
+        .await
+        .expect("We should be able to get the identity stream");
+
+    error!("Alice is resetting the identity");
```
andybalaam: Was this an error for your debugging purposes? Shouldn't be an error, right?

testing/matrix-sdk-integration-testing/src/tests/e2ee/shared_history.rs
```diff
@@ -128,3 +133,195 @@ async fn test_history_share_on_invite() -> Result<()> {
...
+    info!(room_id = ?alice_room.room_id(), "Alice has created and enabled encryption in the room");
+
+    // ... and sends a message
+    let event_id = alice_room
+        .send(RoomMessageEventContent::text_plain("Hello Bob"))
+        .await
+        .expect("We should be able to send a message to the room")
+        .event_id;
+
+    let bundle_stream = bob
```
andybalaam, line 233: The comments go a bit quiet here - would be useful to have some more to help me understand what's happening.

testing/matrix-sdk-integration-testing/src/tests/e2ee/shared_history.rs
```diff
@@ -128,3 +133,195 @@ async fn test_history_share_on_invite() -> Result<()> {
...
+    // Alice is done. Bob has been invited and the room key bundle should have been
+    // sent out. Let's stop syncing so the logs contain less noise.
+    alice_sync_service.stop().await;
+
+    // Let's wait for the bundle to arrive.
+    pin_mut!(bundle_stream);
+    assert_next_with_timeout!(bundle_stream, 3000);
+
+    // Let us now wait till Alice's identity gets updated.
+    error!("Bob is checking if alice's identity has changed");
```
andybalaam: Again, should be `info` probably

crates/matrix-sdk-crypto/src/store/mod.rs
```diff
@@ -1611,66 +1608,86 @@ impl Store {
...
+            bundle_info.sender_data.clone()
+        };
+
+        tracing::Span::current().record("sender_data", tracing::field::debug(&sender_data));
+
+        match sender_data {
+            SenderData::UnknownDevice { .. }
+            | SenderData::VerificationViolation(_)
+            | SenderData::DeviceInfo { .. } => {
+                warn!("Not accepting a historic room key bundle due to inssuficient trust in the sender");
```
andybalaam: inssuficient -> insufficient

andybalaam (approved):
Looks good. Minor comments inline.

testing/matrix-sdk-integration-testing/src/tests/e2ee/shared_history.rs
poljar: Ah duh, yes. `error!()` gets a red color which is easier to spot.

testing/matrix-sdk-integration-testing/src/tests/e2ee/shared_history.rs
poljar: I added some more. Hopefully that makes it a bit clearer.

testing/matrix-sdk-integration-testing/src/tests/e2ee/shared_history.rs
poljar: Removed it.

PR #5512, opened 2025-08-11
feat: Change type of `DecryptedRoomEvent::event` to `Raw<AnyTimelineEvent>`

kaylendog:
- [x] Change `DecryptedRoomEvent::event` to `Raw<AnyTimelineEvent>`
- [x] Update usages to pattern match on `AnyTimelineEvent::MessageLike` where necessary

### DIFF
--- a/crates/matrix-sdk-common/src/deserialized_responses.rs
+++ b/crates/matrix-sdk-common/src/deserialized_responses.rs
@@ -14,12 +14,10 @@
 
 use std::{collections::BTreeMap, fmt, sync::Arc};
 
-#[cfg(doc)]
-use ruma::events::AnyTimelineEvent;
 use ruma::{
     DeviceKeyAlgorithm, OwnedDeviceId, OwnedEventId, OwnedUserId,
     events::{
-        AnyMessageLikeEvent, AnySyncMessageLikeEvent, AnySyncTimelineEvent, AnyToDeviceEvent,
+        AnySyncMessageLikeEvent, AnySyncTimelineEvent, AnyTimelineEvent, AnyToDeviceEvent,
         MessageLikeEventType,
     },
     push::Action,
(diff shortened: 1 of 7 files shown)


andybalaam:
Test failure seems to be an internal problem at codspeed

andybalaam (changes requested):
Looks good, and kudos for breaking into small commits. I like each commit to compile and pass tests, however, because it makes `git bisect` work nicely, so I'd prefer all in one commit if these don't compile separately :-)

No need to change that now though: we can squash merge this if we agree it should all be combined.

For the changelog: please can you list the `pub` functions that are affected and what users should do to if they use these methods?

crates/matrix-sdk-common/src/deserialized_responses.rs
```diff
@@ -843,7 +841,7 @@ pub struct DecryptedRoomEvent {
     /// encrypted payload *always contains* a room id, by the [spec].
     ///
     /// [spec]: https://spec.matrix.org/v1.12/client-server-api/#mmegolmv1aes-sha2
-    pub event: Raw<AnyMessageLikeEvent>,
+    pub event: Raw<AnyTimelineEvent>,
```
richvdh, line 845: The documentation on this still refers to `AnyMessageLikeEvent`: could you update it please?

It should probably say:

```rust
     /// Note: it's not an error that this contains an [`AnyTimelineEvent`]
     /// (as opposed to an [`AnySyncTimelineEvent`]): an
     /// encrypted payload *always contains* a room id, by the [spec].
```

andybalaam (approved):

Hywan:
Thanks for the patch. Can you explain me why you need that? The changes are good and nice, kudos for that, but you forgot to explain the motivations for that ^^.

kaylendog:
> Thanks for the patch. Can you explain me why you need that? The changes are good and nice, kudos for that, but you forgot to explain the motivations for that ^^.

It's fundamentally about being able to avoid casting when implementing encrypted state events. At the moment, using `Raw<AnyMessageLikeEvent>` requires that a decrypted event can never be a state event (unless it is unchecked-casted). There are a couple of places in my older #5456 PR where I [needed to cast](https://github.com/kaylendog/matrix-rust-sdk/blob/a8b7d119d85a24189aedc0a29d3e72942f76a93b/crates/matrix-sdk-base/src/response_processors/state_events.rs#L187) between `Any(Sync)StateEvent` and `AnyMessageLikeEvent` and vice versa. This change would allow me to avoid the casting when decrypting state events.

I would have liked to feature gate this behind `experimental-encrypted-state-events` but that would require conditional compilation flags all over the place.

PR #5516, opened 2025-08-12
sdk: Improvements and fixes around media APIs

zecakeh:
This includes 3 small commits:

1. Use the new support for (un)stable feature flags in Ruma to decide on whether we will use the authenticated media endpoints, instead of having custom code.
2. Since #5437, the request timeout was ignored in media requests, however that didn't apply to unauthenticated media endpoints because they didn't use the request config. This fixes it.
3. #5119 added a mandatory check when uploading files, using the allowed upload size advertised by the server at the media config endpoint. However that check only supports the authenticated media config endpoint, so users on servers that don't support authenticated media cannot send media anymore. We got a bug report because of that in Fractal, from a user using a private homeserver instance that doesn't support authenticated media. This adds support for the unauthenticated media config endpoint too.

### DIFF
--- a/crates/matrix-sdk/src/client/mod.rs
+++ b/crates/matrix-sdk/src/client/mod.rs
@@ -45,6 +45,7 @@ use ruma::{
         client::{
             account::whoami,
             alias::{create_alias, delete_alias, get_alias},
+            authenticated_media,
             device::{delete_devices, get_devices, update_device},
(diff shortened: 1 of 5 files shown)


zecakeh:
All right, I fixed the CI.

stefanceriu:
Beautiful, thank you!

Issue #5474, opened 2025-01-27
UTDs not retried when key arrives at almost the same time


andybalaam:
### Steps to reproduce

See this rageshake: https://github.com/element-hq/element-x-android-rageshakes/issues/4076

The logs in the receiver side show a pattern like this:

```
2025-01-20T15:46:43.519041Z Failed to decrypt a room event
... (9 more Failed to decrypt)
2025-01-20T15:46:46.398276Z Failed to decrypt a room event
2025-01-20T15:46:46.407885Z Received a new megolm room key
2025-01-20T15:46:46.418528Z Failed to decrypt a room event
.. (2 more Failed to decrypt)
```

So the megolm key arrived at almost the same moment as the events.

There is no mention of "Retrying decryption" anywhere in the receiver logs.

I am wondering whether the async task pulling from `room_keys_received_stream()` might have retried decryption too early, before the incoming batch of events was finished processing?

richvdh:
This looks to happen on EXI as well (I don't see any reason it wouldn't). I'm going to move it to the rust-sdk repo.

poljar:
Oh, I think I had an epiphany about this.

Since in our sliding sync world we have two syncs running, one for the events and one for the to-device messages, I think we have a race:

1. The event is received, we try to decrypt it bu there's no key yet.
2. The room key is received, we notify that a key has been received.
3. The event is pushed to the timeline.

andybalaam:
So to fix this, we could re-attempt decryption whenever an undecrypted event is pushed to the timeline. I'm going to experiment with this and see whether it helps.

andybalaam:
@poljar is taking the test code I wrote and using it to test his refactor that moves redecryption into the event cache. That refactor should fix this race, so I'm assigning the task to him for now.

poljar:
PR fixing this can be found here: https://github.com/matrix-org/matrix-rust-sdk/pull/5881

PR #5517, opened 2025-08-12
refactor(timeline): Make RoomDataProvider provide Decryptor to simplify redecryption

andybalaam:
Part of the work on https://github.com/matrix-org/matrix-rust-sdk/issues/5474

This should help us (in future changes) to call into the decryptor thread from code that adds events to the timeline.

### DIFF
--- a/crates/matrix-sdk-ui/src/timeline/controller/mod.rs
+++ b/crates/matrix-sdk-ui/src/timeline/controller/mod.rs
@@ -21,9 +21,8 @@ use eyeball_im_util::vector::{FilterMap, VectorObserverExt};
 use futures_core::Stream;
 use imbl::Vector;
 #[cfg(test)]
-use matrix_sdk::crypto::OlmMachine;
+use matrix_sdk::Result;
 use matrix_sdk::{
-    Result, Room,
     deserialized_responses::TimelineEvent,
     event_cache::{RoomEventCache, RoomPaginationStatus},
     paginators::{PaginationResult, Paginator},
@@ -31,6 +30,8 @@ use matrix_sdk::{
         LocalEcho, LocalEchoContent, RoomSendQueueUpdate, SendHandle, SendReactionHandle,
     },
 };
+#[cfg(test)]
+use ruma::events::receipt::ReceiptEventContent;
 use ruma::{
     EventId, MilliSecondsSinceUnixEpoch, OwnedEventId, OwnedTransactionId, TransactionId, UserId,
     api::client::receipt::create_receipt::v3::ReceiptType as SendReceiptType,
@@ -46,8 +47,6 @@ use ruma::{
     room_version_rules::RoomVersionRules,
     serde::Raw,
 };
-#[cfg(test)]
-use ruma::{OwnedRoomId, RoomId, events::receipt::ReceiptEventContent};
 use tokio::sync::{RwLock, RwLockWriteGuard};
 use tracing::{debug, error, field::debug, info, instrument, trace, warn};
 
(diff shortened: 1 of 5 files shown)


crates/matrix-sdk-ui/src/timeline/traits.rs
```diff
@@ -88,7 +86,7 @@ impl RoomExt for Room {
 }
 
 pub(super) trait RoomDataProvider:
-    Clone + PaginableRoom + PaginableThread + PinnedEventsRoom + 'static
+    Clone + Decryptor + PaginableRoom + PaginableThread + PinnedEventsRoom + 'static
```
andybalaam, line 89: This is the core of this PR.

crates/matrix-sdk-ui/src/timeline/tests/mod.rs
```diff
@@ -461,3 +470,66 @@ impl RoomDataProvider for TestRoomDataProvider {
...
+
+impl Decryptor for TestRoomDataProvider {
+    async fn decrypt_event_impl(
+        &self,
+        raw: &Raw<AnySyncTimelineEvent>,
+        push_ctx: Option<&PushContext>,
+    ) -> matrix_sdk::Result<TimelineEvent> {
+        let Some(decryptor) = &self.decryptor else {
+            panic!(
+                "No TestDecryptor supplied! Use TestTimelineBuilder::decryptor() to provide one."
```
bnjbvr:
```suggestion
                "No TestDecryptor supplied! Use TestTimelineBuilder::with_decryptor() to provide one."
```

crates/matrix-sdk-ui/src/timeline/tests/mod.rs
```diff
@@ -271,6 +272,9 @@ struct TestRoomDataProvider {
     /// The [`EncryptionInfo`] describing the Megolm sessions that were used to
     /// encrypt events.
     pub encryption_info: HashMap<String, Arc<EncryptionInfo>>,
+
+    /// If we are going to do event decryption, the decryptor that provides
```
bnjbvr: Is a word missing at the end? (At least, a dot is :))

crates/matrix-sdk-ui/src/timeline/controller/mod.rs
```diff
@@ -161,15 +160,15 @@ pub(super) struct TimelineController<P: RoomDataProvider = Room, D: Decryptor =
 
     /// A [`RoomDataProvider`] implementation, providing data.
     ///
-    /// Useful for testing only; in the real world, it's just a [`Room`].
+    /// In the real world, this is just a [`Room`].
```
bnjbvr: I think I prefer the new comment much less: this begs me the question, as a reader, "then why not use a plain `Room` here?", which the previous comment responded to clearly. Are you open to reverting this line?

andybalaam:
> This is just a refactor, with no changes in behavior, at this point, right?

Yes
 
> Can you tweak the PR name so that it also follows the commit message convention, please?

Done, thanks.

crates/matrix-sdk-ui/src/timeline/tests/mod.rs
andybalaam: Good spot, fixed in 911602fb06d510ca4

crates/matrix-sdk-ui/src/timeline/tests/mod.rs
andybalaam: Thanks, fixed in 7d67b901f9

crates/matrix-sdk-ui/src/timeline/controller/mod.rs
andybalaam: Fixed in 948012d722 . I was confused by the original comment because it implied (to me) that the field is only useful for testing, whereas the correct meaning is that the generic type is only used for testing.

PR #5511, opened 2025-08-11
feat(crypto): Add RoomSettings::encrypt_state_events

kaylendog:
- [x] Add `RoomSettings::encrypt_state_events`, which will be used when implementing encrypted state support to the WASM SDK.

PR #5519, opened 2025-08-12
feat(crypto): Add support for encrypted state events to `matrix-sdk-crypto`

kaylendog:
Depends on #5511, #5512 and #5523.

- [ ] Add `OutboundGroupSession::encrypt_state`.
- [ ] Add `GroupSessionManager::encrypt_state` and a private helper function `::encrypt_inner`.
- [ ] Add `OlmMachine::encrypt_state_event` and `::encrypt_state_event_raw`.
- [ ] Add naive state key unpacking and verification to `OlmMachine::decrypt_room_event_inner`.

### DIFF
--- a/crates/matrix-sdk-common/src/deserialized_responses.rs
+++ b/crates/matrix-sdk-common/src/deserialized_responses.rs
@@ -998,6 +998,11 @@ pub enum UnableToDecryptReason {
     /// cross-signing identity did not satisfy the requested
     /// `TrustRequirement`.
     SenderIdentityNotTrusted(VerificationLevel),
+
+    /// The outer state key could not be verified against the inner encrypted
+    /// state key and type.
+    #[cfg(feature = "experimental-encrypted-state-events")]
+    StateKeyVerificationFailed,
 }
 
 impl UnableToDecryptReason {
--- a/crates/matrix-sdk-crypto/src/error.rs
+++ b/crates/matrix-sdk-crypto/src/error.rs
@@ -133,6 +133,12 @@ pub enum MegolmError {
     /// The nested value is the sender's current verification level.
     #[error("decryption failed because trust requirement not satisfied: {0}")]
     SenderIdentityNotTrusted(VerificationLevel),
+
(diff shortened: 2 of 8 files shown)


crates/matrix-sdk-crypto/src/error.rs
```diff
@@ -133,6 +133,12 @@ pub enum MegolmError {
     /// The nested value is the sender's current verification level.
     #[error("decryption failed because trust requirement not satisfied: {0}")]
     SenderIdentityNotTrusted(VerificationLevel),
+
+    /// The outer state key could not be verified against the inner encrypted
+    /// state key and type.
+    #[cfg(feature = "experimental-encrypted-state-events")]
+    #[error("decryption failed because the state key failed to validate")]
+    StateKeyVerificationFailed,
```
kaylendog, line 141: This could be split into a larger variety of cases:

- `MalformedStateKey` - when the outer encrypted event state key is not formatted as `str:str`.
- `MismatchBodyType` - when the decrypted ciphertext is not a state event but the outer event is.
- `EventTypeMismatch` - when the outer event type does not match the inner.
- `StateKeyMismatch` - when the outer state key does not match the inner.

These may not be worth implementing?

crates/matrix-sdk-common/src/deserialized_responses.rs
```diff
@@ -998,6 +998,11 @@ pub enum UnableToDecryptReason {
     /// cross-signing identity did not satisfy the requested
     /// `TrustRequirement`.
     SenderIdentityNotTrusted(VerificationLevel),
+
+    /// The outer state key could not be verified against the inner encrypted
+    /// state key and type.
+    #[cfg(feature = "experimental-encrypted-state-events")]
+    StateKeyVerificationFailed,
```
kaylendog, line 1005: See https://github.com/matrix-org/matrix-rust-sdk/pull/5519/files#r2276885909

kaylendog:
Closing in favour of a branch local to this repository for branch-matching.

PR #5520, opened 2025-08-12
feat(send queue): automatically subscribe when replying in a thread

bnjbvr:
This adds a manual thread subscription when a user is replying to a thread, via the send queue, according to the semantics of MSC4306.

Probably needs some kind of centralized mechanism to automatically subscribe when sending a message in the send queue, for both the send queue and the event cache.

Part of #4869.


bnjbvr:
Fold in the other PR

PR #5523, opened 2025-08-12
feat(base): Add `EncryptionState::StateEncrypted`

kaylendog:
- [ ] Add `EncryptionState::StateEncrypted` to represent the encryption state of rooms supporting state event encryption.


crates/matrix-sdk-base/src/room/encryption.rs
```diff
@@ -36,6 +36,11 @@ pub enum EncryptionState {
     /// The room is encrypted.
     Encrypted,
 
+    /// The room is encrypted, additionally requiring state events to be
+    /// encrypted.
+    #[cfg(feature = "experimental-encrypted-state-events")]
+    StateEncrypted,
```
Hywan, line 42: I think there is a design problem here.

`EncryptionState` is an enum. But adding this `StateEncrypted` variant makes it act like a bit flags, like if it should be combined with another variant. And this reflects with the rest of the patch, where you consider `is_encrypted` as `StateEncrypted | Encrypted`.

So, either you move `EncryptionState` to a real bit flags style (with [`bitflags`](https://crates.io/crates/bitflags) — we are already using this crate in the repository), or you change `StateEncrypted`.

1. **Bit flags style** It makes sense somehow, where `0b00` represents `NotEncrypted`, `0b01` represents `Encrypted` and `0b10` represents encrypted states. So `0b11` represents `Encrypted` + `StateEncrypted`.
2. **New enum** You can change `Encrypted` to be `Encrypted { with_encrypted_states: bool }` and the problem is solved.

kaylendog: I'm not sure those would be particularly compatible with feature gating:

1. I could convert `EncryptionState` to use bit-flags and conditionally add a flag for encrypted state, but the bit flag approach implies the existence of a room where state events are encrypted but not message-like events, which I don't believe should ever happen under MSC3414.
2. The boolean field would require conditional expansion of the struct variant whenever the feature is enabled, which I imagine will be quite a lot of changes.

Perhaps it's a naming issue? Maybe `EncryptedIncludingStateEvents` would be a better name, although it's a bit verbose.

crates/matrix-sdk-base/src/room/encryption.rs
```diff
@@ -155,5 +175,16 @@ mod tests {
...
+        #[cfg(feature = "experimental-encrypted-state-events")]
+        {
+            assert!(EncryptionState::StateEncrypted.is_unknown().not());
+            assert!(EncryptionState::StateEncrypted.is_encrypted());
+
+            assert!(EncryptionState::Unknown.is_state_encrypted().not());
+            assert!(EncryptionState::Encrypted.is_state_encrypted().not());
+            assert!(EncryptionState::StateEncrypted.is_state_encrypted());
+            assert!(EncryptionState::NotEncrypted.is_state_encrypted().not());
+        }
```
kaylendog, line 188: I'd like for this to not be feature gated so it gets tested properly, but I can't figure out how to persuade the tests to run with `experimental-encrypted-state-events` on.

crates/matrix-sdk-base/src/room/encryption.rs
Hywan:
```rust
struct EncryptionState {
    Encrypted {
        with_encrypted_states: bool,
    },
    NotEncrypted,
}
```

`with_encrypted_states` is always `false` if `#[cfg(not(feature = "experimental-encrypted-state-events"))]`. Otherwise, it depends of the MSC3414.

I don't see why it creates a lot of changes. Am I missing something?

kaylendog: I interpreted your suggestion as
```rs
enum EncryptionState {
    #[cfg(not(feature = "experimental-encrypted-state-events"))]
    Encrypted,
    
    #[cfg(feature = "experimental-encrypted-state-events")]
    Encrypted {
        with_encrypted_states: bool,
    }
    
    // ...
}
```
You'd need to modify all the pattern matches on `EncryptionState::Encrypted` to `EncryptionState::Encrypted { .. }`. I can do that? Just seems undesirable to me.

kaylendog:
Looks like this [breaks the Go FFI bindings](https://github.com/matrix-org/complement-crypto/blob/273970292fb78eaec009561ab0387e78a05d216f/internal/api/rust/rust.go#L545) in `complement-crypto`

crates/matrix-sdk-base/src/room/encryption.rs
```diff
@@ -33,8 +33,11 @@ impl Room {
 #[derive(Debug)]
 #[cfg_attr(feature = "uniffi", derive(uniffi::Enum))]
 pub enum EncryptionState {
-    /// The room is encrypted.
-    Encrypted,
+    /// The room is encrypted, possibly including state events.
+    Encrypted {
+        /// Determines whether experimental state encryption is enabled.
+        encrypt_state_events: bool,
```
Hywan: Why not `encrypted_state_events`?

Hywan:
> Looks like this [breaks the Go FFI bindings](https://github.com/matrix-org/complement-crypto/blob/273970292fb78eaec009561ab0387e78a05d216f/internal/api/rust/rust.go#L545) in `complement-crypto`

Yes, you may need to update Complement Crypto. Ideally, you can open a branch on Complement Crypto with the same name as your branch here, to trigger a special build that will help tests both PR at the same time.

kaylendog:
Reverted back to the enum variant after the discussion with @richvdh.

PR #5525, opened 2025-08-13
refactor!(ffi): always use the send queue and remove legacy media upload progress

bnjbvr:
Now that the we media upload progress in the timeline, thanks to @Johennes, we can remove the legacy support, so as to make the API less confusing (before, one could use a `progress_watcher` *and* enable the send queue media upload progress at the same time, and it wasn't clear how these two API would interact / which would be enabled).

Also always use the send queue for media uploads, as it's generally better, stable (after months in production in the EX apps).


stefanceriu (approved):
Yey! 🔥

PR #5526, opened 2025-08-13
Upgrade the version of Rust nightly

zecakeh:
The upgrade makes cargo check detect new dead code so this removes it.


andybalaam (approved):
Looks great, thank you!

andybalaam:
@zecakeh any clue why this would have happened?

> Merging https://github.com/matrix-org/matrix-rust-sdk/pull/5526 will degrade performances by 26.23%

It feels like a false negative to me...

Or a Rust compiled performance regression?

zecakeh:
This check seems flaky to me, right now it says that the performance is not altered.

PR #5527, opened 2025-08-13
feat(timeline): consider unthreaded read receipts for `ReceiptThread::Main` timelines when computing timeline items states.

stefanceriu:
This patch updates the Timeline Controller's `handle_explicit_read_receipts` method to also consider unthreaded read receipts on **threaded** **main** timelines when calculating timeline items states and adds a test for it.

This picks up from #5442 and fixes https://github.com/matrix-org/matrix-rust-sdk/issues/5440

### DIFF
--- a/crates/matrix-sdk-ui/src/timeline/tests/read_receipts.rs
+++ b/crates/matrix-sdk-ui/src/timeline/tests/read_receipts.rs
@@ -894,4 +894,6 @@ async fn test_unthreaded_client_updates_threaded_read_receipts() {
     let event_b = timeline.controller.items().await[2].as_event().unwrap().to_owned();
     assert_eq!(event_b.read_receipts().len(), 1);
     assert!(event_b.read_receipts().get(*BOB).is_some());
+
+    assert_pending!(stream);
 }
--- a/crates/matrix-sdk-ui/src/timeline/tests/read_receipts.rs
+++ b/crates/matrix-sdk-ui/src/timeline/tests/read_receipts.rs
@@ -845,6 +845,7 @@ async fn test_unthreaded_client_updates_threaded_read_receipts() {
     assert_next_matches!(stream, VectorDiff::PushBack { .. });
     assert_next_matches!(stream, VectorDiff::PushFront { .. });
     assert_next_matches!(stream, VectorDiff::PushBack { .. });
+    assert_pending!(stream);
 
     // Bob reads the last one from an unthreaded client
     timeline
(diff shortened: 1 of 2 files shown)


crates/matrix-sdk-ui/src/timeline/tests/read_receipts.rs
```diff
@@ -826,3 +826,39 @@ async fn test_threaded_latest_user_read_receipt() {
...
+    let event_b = event_id!("$event_b");
+
+    // Alice sends 2 messages
+    let f = &timeline.factory;
+    timeline.handle_live_event(f.text_msg("A").sender(*ALICE)).await;
+    timeline.handle_live_event(f.text_msg("B").sender(*ALICE).event_id(event_b)).await;
+
+    assert_next_matches!(stream, VectorDiff::PushBack { .. });
+    assert_next_matches!(stream, VectorDiff::PushFront { .. });
+    assert_next_matches!(stream, VectorDiff::PushBack { .. });
```
bnjbvr, line 847: Can you assert the stream is empty after this?

crates/matrix-sdk-ui/src/timeline/tests/read_receipts.rs
```diff
@@ -826,3 +826,39 @@ async fn test_threaded_latest_user_read_receipt() {
...
+            BOB.to_owned(),
+            ReceiptThread::Unthreaded,
+        )])
+        .await;
+
+    // Alice's timeline gets updated
+    let item_b = assert_next_matches!(stream, VectorDiff::Set { index: 2, value } => value);
+    let event_b = item_b.as_event().unwrap();
+    assert_eq!(event_b.read_receipts().len(), 1);
+    assert!(event_b.read_receipts().get(*BOB).is_some());
```
bnjbvr, line 864: can you assert the stream's empty after this too?

crates/matrix-sdk-ui/src/timeline/tests/read_receipts.rs
```diff
@@ -826,3 +826,39 @@ async fn test_threaded_latest_user_read_receipt() {
...
+        .focus(TimelineFocus::Live { hide_threaded_events: true })
+        .build();
+    let mut stream = timeline.subscribe().await;
+
+    let event_b = event_id!("$event_b");
+
+    // Alice sends 2 messages
+    let f = &timeline.factory;
+    timeline.handle_live_event(f.text_msg("A").sender(*ALICE)).await;
+    timeline.handle_live_event(f.text_msg("B").sender(*ALICE).event_id(event_b)).await;
```
bnjbvr, line 843: I'm curious to propose a small change in this test, that I think wouldn't require a change in the rest of the code, in theory :)

Let's add a third event with id `event_c`, that's sent by Alice *in a thread*.

This event should be filtered out, because the timeline is set to `hide_threaded_events`.

Now, make Bob's read receipt point to `event_c`.

I think the outcome should remain the same, and the read receipt on `event_b` should be set to Bob's.

(This enhanced version proves that, even if the unthreaded read receipt is on an in-thread event, we can put it "at the right position" in the no-threaded-events timeline.)

bnjbvr (approved):
Great job, thanks! Small proposal to change the test (or if you prefer, add a new one), and that shouldn't require any changes in the timeline's code, if I understand correctly

crates/matrix-sdk-ui/src/timeline/tests/read_receipts.rs
stefanceriu: Great idea! Added it in 875ec25e435233989974d257c82320bcb7d81c4a, fits right in 👍

crates/matrix-sdk-ui/src/timeline/tests/read_receipts.rs
```diff
@@ -826,3 +826,72 @@ async fn test_threaded_latest_user_read_receipt() {
...
+            ReceiptType::Read,
+            BOB.to_owned(),
+            ReceiptThread::Thread(thread_event_id.to_owned()),
+        )])
+        .await;
+
+    // The main timeline read receipts are still correct
+    let event_b = timeline.controller.items().await[2].as_event().unwrap().to_owned();
+    assert_eq!(event_b.read_receipts().len(), 1);
+    assert!(event_b.read_receipts().get(*BOB).is_some());
```
bnjbvr, line 896:
```suggestion
    assert!(event_b.read_receipts().get(*BOB).is_some());

    assert_pending!(stream);
```

PR #5532, opened 2025-08-14
chore: check in the updated compatible ruma version

bnjbvr:
Likely a fallout of a race merging two PRs (a ruma bump + the introduction of the ruma-signatures usage).

### DIFF
--- a/crates/matrix-sdk/src/client/mod.rs
+++ b/crates/matrix-sdk/src/client/mod.rs
@@ -3033,7 +3033,6 @@ pub(crate) mod tests {
 
     use super::Client;
     use crate::{
-        assert_let_timeout,
         client::{futures::SendMediaUploadRequest, WeakClient},
         config::{RequestConfig, SyncSettings},
         futures::SendRequest,
@@ -3963,7 +3962,12 @@ pub(crate) mod tests {
 
         // Call the endpoint once to check the timeout.
         let mut stream = Box::pin(client.sync_stream(SyncSettings::new()).await);
-        assert_let_timeout!(Some(Ok(_)) = stream.next());
+
+        timeout(Duration::from_secs(1), async {
+            stream.next().await.unwrap().unwrap();
+        })
+        .await
+        .unwrap();
     }
 
     #[async_test]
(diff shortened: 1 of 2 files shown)


crates/matrix-sdk/src/client/mod.rs
```diff
@@ -3963,7 +3962,12 @@ pub(crate) mod tests {
 
         // Call the endpoint once to check the timeout.
         let mut stream = Box::pin(client.sync_stream(SyncSettings::new()).await);
-        assert_let_timeout!(Some(Ok(_)) = stream.next());
+
+        timeout(Duration::from_secs(1), async {
+            stream.next().await.unwrap().unwrap();
+        })
+        .await
+        .unwrap();
```
Hywan, line 3970: Why not `assert_let_timeout!(Duration::from_secs(1), Some(Ok(_)) = stream.next());`?

crates/matrix-sdk/src/client/mod.rs
```diff
@@ -3992,7 +3996,12 @@ pub(crate) mod tests {
         let mut stream = Box::pin(
             client.sync_stream(SyncSettings::new().ignore_timeout_on_first_sync(true)).await,
         );
-        assert_let_timeout!(Some(Ok(_)) = stream.next());
-        assert_let_timeout!(Some(Ok(_)) = stream.next());
+
+        timeout(Duration::from_secs(1), async {
+            stream.next().await.unwrap().unwrap();
+            stream.next().await.unwrap().unwrap();
+        })
```
Hywan, line 4003: Same here.

crates/matrix-sdk/src/client/mod.rs
bnjbvr: we're not inspecting the result, so binding the value is useless

crates/matrix-sdk/src/client/mod.rs
bnjbvr: Same here.

Hywan (approved):

Issue #5513, opened 2025-08-11
History sharing: improve efficiency of building key bundle


richvdh:
Currently, when constructing a key bundle for sharing an invite, we pull *all* keys out of the datastore and then filter by room. This can take many minutes on a large account.

We should only pull out the keys for the right room (and ideally, only those marked as shareable).

See: https://github.com/matrix-org/matrix-rust-sdk@3a98d46/crates/matrix-sdk-crypto/src/store/mod.rs#L1586-L1589

multisme:
Doesn't this require to write a migration to index the keys by room_id?

richvdh:
I believe they are already indexed by room_id, both in sqlite and indexeddb.

multisme:
Can we query a partial key in indexedDB?
https://github.com/matrix-org/matrix-rust-sdk/blob/1eb8f6ac161fbf1c773a7ddb8243d7378489fdc8/crates/matrix-sdk-indexeddb/src/safe_encode.rs#L79-L90

richvdh:
@multisme I believe so. You can use a key range. I think `serializer.encode_to_range` does what we need.

Are you asking these questions because you are considering contributing an implementation? It would be much appreciated 😇

multisme:
Yes, I want to finish something first and I will try to solve this one.

poljar:
Is this now done since https://github.com/matrix-org/matrix-rust-sdk/pull/5534 was merged?

richvdh:
No, #5534 added functionality to the store implementations, but we still have to use the new impl. That should be easy though. @multisme were you planning to make a PR for that, or would you prefer us to pick it up?

multisme:
Yes, but I can only be back on it next week?

poljar:
Alright #5619 got merged. Unless I'm missing something that means we can close this.

andybalaam:
PR for matrix-js-sdk to get this into Element Web: https://github.com/matrix-org/matrix-js-sdk/pull/4991

PR #5534, opened 2025-08-14
 Implement querying inboundgroupsessions by room_id

multisme:
History sharing: improve efficiency of building key bundle

Signed-off-by: multi [multiestunhappydev@gmail.com](mailto:multiestunhappydev@gmail.com)

Partially Implement https://github.com/matrix-org/matrix-rust-sdk/issues/5513


multisme:
@richvdh 
Hi,
Am I on the right path regarding the implementation of that functionality with IndexedDB?

crates/matrix-sdk-indexeddb/src/crypto_store/mod.rs
```diff
@@ -967,6 +967,31 @@ impl_crypto_store! {
         }
     }
 
+    async fn get_inbound_group_session_by_room_id(
```
richvdh:
```suggestion
    async fn get_inbound_group_sessions_by_room_id(
```

richvdh: Also, this belongs *after* `get_inbound_group_sessions`.

richvdh (commented):
Looks plausible to me. It will need defining in the trait, and some tests, obviously.

crates/matrix-sdk-crypto/src/store/traits.rs
```diff
@@ -121,6 +121,15 @@ pub trait CryptoStore: AsyncTraitDeps {
...
     async fn get_inbound_group_sessions(&self) -> Result<Vec<InboundGroupSession>, Self::Error>;
 
+    /// Get all the inbound group sessions linked to a room
+    ///
+    /// # Arguments
+    /// * `room_id` - the room id of the room that the sessions are linked to.
+    async fn get_inbound_group_sessions_by_room_id(
+        &self,
+        room_id: &RoomId,
+    ) -> Result<Vec<InboundGroupSession>, Self::Error>;
```
richvdh: suggest slotting this in between `inbound_group_session_counts` and `get_inbound_group_sessions_for_device_batch`.

Likewise in `MemoryStore` and `SqliteCryptoStore`.

crates/matrix-sdk-crypto/src/store/traits.rs
```diff
@@ -121,6 +121,15 @@ pub trait CryptoStore: AsyncTraitDeps {
     /// Get all the inbound group sessions we have stored.
     async fn get_inbound_group_sessions(&self) -> Result<Vec<InboundGroupSession>, Self::Error>;
 
+    /// Get all the inbound group sessions linked to a room
```
richvdh:
```suggestion
    /// Get all the inbound group sessions for a given room.
```

crates/matrix-sdk-crypto/src/store/traits.rs
```diff
@@ -121,6 +121,15 @@ pub trait CryptoStore: AsyncTraitDeps {
...
+    ///
+    /// # Arguments
+    /// * `room_id` - the room id of the room that the sessions are linked to.
```
richvdh:
```suggestion
    /// * `room_id` - The ID of the room to return sessions for.
```

crates/matrix-sdk-crypto/src/store/memorystore.rs
```diff
@@ -443,6 +443,26 @@ impl CryptoStore for MemoryStore {
...
+    async fn get_inbound_group_sessions_by_room_id(
+        &self,
+        room_id: &RoomId,
+    ) -> Result<Vec<InboundGroupSession>> {
+        let inbounds: Vec<InboundGroupSession> = self
+            .inbound_group_sessions
+            .read()
+            .get(room_id)
+            .into_iter()
+            .flat_map(HashMap::values)
```
richvdh: Converting the `Option` into an iterator seems like a bit of an obscure way of dealing with it. Maybe you could do something with `map` or a `match` statement.

crates/matrix-sdk-crypto/src/store/integration_tests.rs
```diff
@@ -567,6 +567,64 @@ macro_rules! cryptostore_integration_tests {
                 assert_eq!(store.inbound_group_session_counts(None).await.unwrap().total, 1);
             }
 
+            #[async_test]
+            async fn test_load_inbound_group_sessions_by_room_id() {
+                let dir = "load_inbound_group_session_by_room_id";
```
richvdh: I don't know why the other tests are called `load_*`, but there doesn't seem to be any reason to propagate it.

```suggestion
            async fn test_get_inbound_group_sessions_by_room_id() {
                let dir = "get_inbound_group_session_by_room_id";
```

crates/matrix-sdk-crypto/src/store/integration_tests.rs
```diff
@@ -567,6 +567,64 @@ macro_rules! cryptostore_integration_tests {
...
+                assert_eq!(store.get_inbound_group_sessions().await.unwrap().len(), 0);
+
+                let room_id = &room_id!("!testing:localhost");
+                let (_, session_1) = account.create_group_session_pair_with_defaults(room_id).await;
+                let (_, session_2) = account.create_group_session_pair_with_defaults(room_id).await;
+                let (_, session_3) = account.create_group_session_pair_with_defaults(room_id).await;
+                let (_, session_4) = account.create_group_session_pair_with_defaults(room_id).await;
+                let (_, session_5) = account.create_group_session_pair_with_defaults(room_id).await;
+                let (_, session_6) = account.create_group_session_pair_with_defaults(room_id).await;
+                let (_, session_7) = account.create_group_session_pair_with_defaults(room_id).await;
```
richvdh: I'm not convinced we need 7 sessions to test with. How about 2?

crates/matrix-sdk-crypto/src/store/integration_tests.rs
```diff
@@ -567,6 +567,64 @@ macro_rules! cryptostore_integration_tests {
...
+
+
+                let mut sessions = vec![
+                    session_1.clone(),
+                    session_2.clone(),
+                    session_3.clone(),
+                    session_4.clone(),
+                    session_5.clone(),
+                    session_6.clone(),
+                    session_7.clone(),
```
richvdh: I think the `.clone` calls are redundant?

crates/matrix-sdk-crypto/src/store/integration_tests.rs
richvdh: I think we also need a session for a *different* room, to assert that is *not* returned.

crates/matrix-sdk-crypto/src/store/integration_tests.rs
```diff
@@ -567,6 +567,64 @@ macro_rules! cryptostore_integration_tests {
...
+                    .unwrap();
+
+                for session in loaded_sessions.iter() {
+                    session.export().await;
+                }
+                assert_eq!(loaded_sessions.len(), 7);
+                assert_session_lists_eq(sessions, loaded_sessions, "room by id sessions");
+
+                assert_eq!(store.get_inbound_group_sessions().await.unwrap().len(), 7);
+                assert_eq!(store.inbound_group_session_counts(None).await.unwrap().total, 7);
```
richvdh: this seems unnecessary and unrelated to the function under test?

crates/matrix-sdk-crypto/src/store/memorystore.rs
multisme: I changed this and use an expect that panic now.

crates/matrix-sdk-crypto/src/store/integration_tests.rs
multisme: I kept it as to show the session in the other room is inside the store.

crates/matrix-sdk-crypto/src/store/memorystore.rs
```diff
@@ -485,6 +465,26 @@ impl CryptoStore for MemoryStore {
...
 
+    async fn get_inbound_group_sessions_by_room_id(
+        &self,
+        room_id: &RoomId,
+    ) -> Result<Vec<InboundGroupSession>> {
+        let inbounds: Vec<InboundGroupSession> = self
+            .inbound_group_sessions
+            .read()
+            .get(room_id)
+            .expect("No inbound group sessions for that roomId")
```
richvdh: Panicking in rust should happen only in situations that you strongly believe should not happen. Having no sessions for a given room is a perfectly normal situation.

crates/matrix-sdk-crypto/src/store/integration_tests.rs
```diff
@@ -604,8 +597,13 @@ macro_rules! cryptostore_integration_tests {
                 };
                 store.save_changes(changes).await.expect("Can't save group session");
 
+
```
richvdh: why did you add this extra blank line?

crates/matrix-sdk-crypto/src/store/integration_tests.rs
```diff
@@ -568,29 +568,22 @@ macro_rules! cryptostore_integration_tests {
...
-                    session_4.clone(),
-                    session_5.clone(),
-                    session_6.clone(),
-                    session_7.clone(),
+                    session_1,
+                    session_2,
+                    session_3
                 ];
 
                 for session in sessions.iter_mut() {
```
richvdh: can't this be simply:

```suggestion
                for session in sessions {
```

?

crates/matrix-sdk-crypto/src/store/integration_tests.rs
```diff
@@ -604,8 +597,13 @@ macro_rules! cryptostore_integration_tests {
                 };
                 store.save_changes(changes).await.expect("Can't save group session");
 
+
                 drop(store);
 
+                // Remove the last sessions since it is in an other room
```
richvdh: ... so?

```suggestion
                // The last session is in a different room, so should not be returned by 
                // get_inbound_group_sessions_by_room_id. Remove it from the list.
```

crates/matrix-sdk-crypto/src/store/integration_tests.rs
```diff
@@ -618,11 +616,12 @@ macro_rules! cryptostore_integration_tests {
                 for session in loaded_sessions.iter() {
                     session.export().await;
                 }
```
richvdh: what is the purpose of this loop?

crates/matrix-sdk-crypto/src/store/integration_tests.rs
richvdh: fair enough, but:
 * we only need one of these calls to prove that
 * it would be more logical to put it nearer the start of the test (just after you have saved them in the store)
 * please add a comment to say what it is for

crates/matrix-sdk-crypto/src/store/integration_tests.rs
```diff
@@ -604,8 +597,13 @@ macro_rules! cryptostore_integration_tests {
                 };
                 store.save_changes(changes).await.expect("Can't save group session");
 
+
                 drop(store);
 
+                // Remove the last sessions since it is in an other room
+                sessions.pop();
+
+
```
richvdh: again, why two blank lines here?

crates/matrix-sdk-crypto/src/store/integration_tests.rs
```diff
@@ -568,29 +568,22 @@ macro_rules! cryptostore_integration_tests {
             }
 
             #[async_test]
-            async fn test_load_inbound_group_sessions_by_room_id() {
-                let dir = "load_inbound_group_session_by_room_id";
+            async fn test_get_inbound_group_sessions_by_room_id() {
```
richvdh, line 581: please could you add another test that checks we get an empty list back when there are no sessions in the room?

richvdh (changes requested):

crates/matrix-sdk-crypto/src/store/integration_tests.rs
multisme: I need to keep the sessions to compare later with loaded _sessions. while exporting them.

crates/matrix-sdk-crypto/src/store/integration_tests.rs
multisme: To export the loaded_sessions before making a comparaison with the sessions generated at the beginning of the file

crates/matrix-sdk-crypto/src/store/integration_tests.rs
multisme: Done

crates/matrix-sdk-crypto/src/store/integration_tests.rs
```diff
@@ -586,7 +586,7 @@ macro_rules! cryptostore_integration_tests {
                     session_3
                 ];
 
-                for session in sessions.iter_mut() {
+                for session in sessions.iter_mut(){
```
richvdh: :-1:

crates/matrix-sdk-crypto/src/store/integration_tests.rs
```diff
@@ -619,9 +621,7 @@ macro_rules! cryptostore_integration_tests {
 
                 assert_eq!(loaded_sessions.len(), 2);
                 assert_session_lists_eq(sessions, loaded_sessions, "room by id sessions");
-
-                assert_eq!(store.get_inbound_group_sessions().await.unwrap().len(), 3);
-                assert_eq!(store.inbound_group_session_counts(None).await.unwrap().total, 3);
+
```
richvdh: redundant blank line here

crates/matrix-sdk-crypto/src/store/integration_tests.rs
```diff
@@ -574,6 +574,8 @@ macro_rules! cryptostore_integration_tests {
                 assert_eq!(store.get_inbound_group_sessions().await.unwrap().len(), 0);
 
                 let room_id = &room_id!("!testing:localhost");
+                assert_eq!(store.get_inbound_group_sessions_by_room_id(room_id).await.unwrap().len(), 0);
```
richvdh: this should really have been a separate test, but I guess it's fine.

crates/matrix-sdk-crypto/src/store/integration_tests.rs
richvdh: Ok, but why? Export it to where? Isn't `session.export()` a no-op if you throw away the result?

crates/matrix-sdk-crypto/src/store/integration_tests.rs
richvdh: Actually, my question should have been: what is the point of this loop? Why do we need to `export` and re-import the sessions? Can't we just remove this loop?

crates/matrix-sdk-crypto/src/store/integration_tests.rs
multisme: A test was not working without but I can't reproduce it so I remove it.

richvdh:
@multisme if you could click "re-request review" after you make changes, that would help me remember to take another look

PR #5538, opened 2025-08-14
Widget driver: allow state keys to have a _m.call suffix for Element Call

fkwp:

Signed-off-by:

### DIFF
--- a/bindings/matrix-sdk-ffi/src/widget.rs
+++ b/bindings/matrix-sdk-ffi/src/widget.rs
@@ -220,6 +220,8 @@ pub fn get_element_call_required_permissions(
                 event_type: StateEventType::CallMember.to_string(),
                 state_key: format!("{own_user_id}_{own_device_id}"),
             },
+            // Same as above for [MSC3779] and [MSC4143](https://github.com/matrix-org/matrix-spec-proposals/pull/4143),
+            // with application suffix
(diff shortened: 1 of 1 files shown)


toger5 (approved):
Good catch that we need a custom capability for sending since we are very explicit what keys we allow the widget to send.

bindings/matrix-sdk-ffi/src/widget.rs
```diff
@@ -220,13 +220,21 @@ pub fn get_element_call_required_permissions(
                 event_type: StateEventType::CallMember.to_string(),
                 state_key: format!("{own_user_id}_{own_device_id}"),
             },
+            WidgetEventFilter::StateWithTypeAndStateKey {
```
andybalaam, line 225: This needs a comment to explain it like the others (also the second one).

andybalaam (changes requested):
Looks good, just needs a couple of comments

bindings/matrix-sdk-ffi/src/widget.rs
fkwp: addressed with https://github.com/matrix-org/matrix-rust-sdk/pull/5538/commits/b09c82aa355ee85503124fcfacb8b6f1da48a61d

PR #5539, opened 2025-08-15
feat(crypto): Add support for encrypted state events to `matrix-sdk-crypto`

kaylendog:
Depends on #5511, #5512 and #5523.

- [x] Add `OutboundGroupSession::encrypt_state`.
- [x] Add `GroupSessionManager::encrypt_state` and a private helper function `::encrypt_inner`.
- [x] Add `OlmMachine::encrypt_state_event` and `::encrypt_state_event_raw`.
- [x] Add naive state key unpacking and verification to `OlmMachine::decrypt_room_event_inner`.

### DIFF
--- a/crates/matrix-sdk-crypto/src/machine/mod.rs
+++ b/crates/matrix-sdk-crypto/src/machine/mod.rs
@@ -2261,9 +2261,72 @@ impl OlmMachine {
                 .await;
         }
 
-        let event = serde_json::from_value::<Raw<AnyTimelineEvent>>(decrypted_event.into())?;
+        let decrypted_event =
+            serde_json::from_value::<Raw<AnyTimelineEvent>>(decrypted_event.into())?;
 
-        Ok(DecryptedRoomEvent { event, encryption_info, unsigned_encryption_info })
+        #[cfg(feature = "experimental-encrypted-state-events")]
+        self.verify_packed_state_key(&event, &decrypted_event)?;
+
+        Ok(DecryptedRoomEvent { event: decrypted_event, encryption_info, unsigned_encryption_info })
+    }
+
+    /// If the passed event is a state event, verify its outer packed state key
+    /// matches the inner state key once unpacked.
+    ///
+    /// * `original` - The original encrypted event received over the wire.
+    /// * `decrypted` - The decrypted event.
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if any of the following are true:
+    ///
+    /// * The original event's state key failed to unpack;
+    /// * The decrypted event could not be deserialised;
+    /// * The unpacked event type does not match the type of the decrypted
+    ///   event;
+    /// * The unpacked event state key does not match the state key of the
+    ///   decrypted event.
+    #[cfg(feature = "experimental-encrypted-state-events")]
+    fn verify_packed_state_key(
+        &self,
+        original: &EncryptedEvent,
+        decrypted: &Raw<AnyTimelineEvent>,
+    ) -> MegolmResult<()> {
+        use serde::Deserialize;
+
+        // We only need to verify state events.
+        let Some(raw_state_key) = &original.state_key else { return Ok(()) };
+
+        // Unpack event type and state key from the raw state key.
+        let (outer_event_type, outer_state_key) =
+            raw_state_key.split_once(":").ok_or(MegolmError::StateKeyVerificationFailed)?;
+
+        // Helper for deserializing.
+        #[derive(Deserialize)]
+        struct PayloadDeserializationHelper {
+            state_key: String,
+            #[serde(rename = "type")]
+            event_type: String,
+        }
+
+        // Deserialize the decrypted event.
+        let PayloadDeserializationHelper {
+            state_key: inner_state_key,
+            event_type: inner_event_type,
+        } = decrypted
+            .deserialize_as_unchecked()
+            .map_err(|_| MegolmError::StateKeyVerificationFailed)?;
+
+        // Check event types match, discard if not.
+        if outer_event_type != inner_event_type {
+            return Err(MegolmError::StateKeyVerificationFailed);
+        }
+
+        // Check state keys match, discard if not.
+        if outer_state_key != inner_state_key {
+            return Err(MegolmError::StateKeyVerificationFailed);
+        }
+        Ok(())
     }
 
     /// Try to decrypt the events bundled in the `unsigned` object of the given
@@ -3034,6 +3097,8 @@ fn megolm_error_to_utd_info(
         JsonError(_) => UnableToDecryptReason::PayloadDeserializationFailure,
         MismatchedIdentityKeys(_) => UnableToDecryptReason::MismatchedIdentityKeys,
         SenderIdentityNotTrusted(level) => UnableToDecryptReason::SenderIdentityNotTrusted(level),
+        #[cfg(feature = "experimental-encrypted-state-events")]
+        StateKeyVerificationFailed => UnableToDecryptReason::StateKeyVerificationFailed,
 
         // Pass through crypto store errors, which indicate a problem with our
         // application, rather than a UTD.
--- a/crates/matrix-sdk-crypto/src/machine/mod.rs
+++ b/crates/matrix-sdk-crypto/src/machine/mod.rs
@@ -12,6 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#[cfg(feature = "experimental-encrypted-state-events")]
+use std::borrow::Borrow;
 use std::{
     collections::{BTreeMap, HashMap, HashSet},
     sync::Arc,
@@ -31,6 +33,8 @@ use matrix_sdk_common::{
     locks::RwLock as StdRwLock,
     BoxFuture,
 };
+#[cfg(feature = "experimental-encrypted-state-events")]
+use ruma::events::{AnyStateEventContent, StateEventContent};
 use ruma::{
     api::client::{
         dehydrated_device::DehydratedDeviceData,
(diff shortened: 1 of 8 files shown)


crates/matrix-sdk-crypto/src/olm/group_sessions/outbound.rs
```diff
@@ -458,6 +460,50 @@ impl OutboundGroupSession {
...
 
+    /// Encrypt an arbitrary event for the given room.
+    ///
+    /// Beware that a room key needs to be shared before this method
+    /// can be called using the `share_room_key()` method.
+    ///
+    /// # Arguments
+    ///
+    /// * `payload` - The plaintext content of the event that should be
+    ///   encrypted in raw JSON form.
```
richvdh: it's not yet in raw JSON, right? We will serialize it to JSON.

crates/matrix-sdk-crypto/src/olm/group_sessions/outbound.rs
```diff
@@ -458,6 +460,50 @@ impl OutboundGroupSession {
...
+    ///
+    /// # Panics
+    ///
+    /// Panics if the content can't be serialized.
+    async fn encrypt_inner<T: Serialize>(
+        &self,
+        payload: &T,
```
richvdh, line 478: Nice idea to take a generic `T: Serialize` here, though I'm slightly hesitant because IIRC this means we'll end up with a separate copy of `encrypt_inner` for each `Payload` type. It's probably fine though.

crates/matrix-sdk-crypto/src/machine/tests/mod.rs
```diff
@@ -727,6 +732,105 @@ async fn test_megolm_encryption() {
...
+        })
+        .await
+        .unwrap()
+        .inbound_group_session
+        .unwrap();
+    let sessions = std::slice::from_ref(&group_session);
+    bob.store().save_inbound_group_sessions(sessions).await.unwrap();
+
+    // when we decrypt the room key, the
+    // inbound_group_session_streamroom_keys_received_stream should tell us
```
richvdh:
```suggestion
    // room_keys_received_stream should tell us
```

crates/matrix-sdk-crypto/src/machine/tests/mod.rs
```diff
@@ -727,6 +732,105 @@ async fn test_megolm_encryption() {
...
+    // about it.
+    let room_keys = room_keys_received_stream
+        .next()
+        .now_or_never()
+        .flatten()
+        .expect("We should have received an update of room key infos")
+        .unwrap();
+    assert_eq!(room_keys.len(), 1);
+    assert_eq!(room_keys[0].session_id, group_session.session_id());
```
richvdh: not sure we actually need this bit of the test. We're already testing `room_keys_received_stream` elsewhere and there's no reason to imagine it would be any different for state encryption than for other events.

crates/matrix-sdk-crypto/src/machine/tests/mod.rs
```diff
@@ -727,6 +732,105 @@ async fn test_megolm_encryption() {
...
+        assert_eq!(&content.topic, plaintext);
+    } else {
+        panic!("Decrypted room event has the wrong type");
+    }
+
+    // Just decrypting the event should *not* cause an update on the
+    // inbound_group_session_stream.
+    if let Some(igs) = room_keys_received_stream.next().now_or_never() {
+        panic!("Session stream unexpectedly returned update: {igs:?}");
+    }
```
richvdh: `room_keys_received_stream` rather than `inbound_group_session_stream`, but again: we probably don't need to assert this.

crates/matrix-sdk-crypto/src/machine/mod.rs
```diff
@@ -2261,9 +2261,62 @@ impl OlmMachine {
...
+    /// Returns an error if
+    ///
+    /// * The original event's state key failed to unpack;
+    /// * The decrypted event could not be deserialised;
+    /// * The unpacked event type does not match the type of the decrypted
+    ///   event;
+    /// * The unpacked event state key does not match the state key of the
+    ///   decrypted event.
+    #[cfg(feature = "experimental-encrypted-state-events")]
+    pub fn verify_packed_state_key(
```
richvdh: should this be `pub`? I don't think this should be `pub`.

crates/matrix-sdk-crypto/src/machine/mod.rs
```diff
@@ -2197,9 +2261,62 @@ impl OlmMachine {
...
+
+    /// If the passed event is a state event, verify its outer packed state key
+    /// matches the inner state key once unpacked.
+    ///
+    /// * `original` - The original encrypted event received over the wire.
+    /// * `decrypted` - The decrypted event.
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if
```
richvdh:
```suggestion
    /// Returns an error if any of the following are true:
```

crates/matrix-sdk-crypto/src/machine/mod.rs
```diff
@@ -2261,9 +2261,62 @@ impl OlmMachine {
...
+        // Unpack event type and state key from the raw state key.
+        let (outer_event_type, outer_state_key) =
+            raw_state_key.split_once(":").ok_or(MegolmError::StateKeyVerificationFailed)?;
+
+        // Deserialize the decrypted event.
+        let AnyTimelineEvent::State(inner) =
+            decrypted.deserialize().map_err(MegolmError::JsonError)?
+        else {
+            return Err(MegolmError::StateKeyVerificationFailed);
+        };
```
richvdh: There's a couple of weird things here.

The first is that we end up completely deserializing state events (and throwing away most of the result), but not regular events. Given we only need a couple of fields here, maybe we should instead do something like:

```rust
    struct PayloadDeserializationHelper {
        state_key: String,
        type: String,
    }
```

... and deserialize into that?

The second weird thing is that we end up returning a `MegolmError::JsonError` error rather than a `MegolmError::StateKeyVerificationFailed` if deserialization fails (whether or not it has a `state_key`). Maybe that's ok, but it feels a bit odd for a ficntion named `verify_packed_state_key` to return  `MegolmError::JsonError` sometimes. Possibly we want a `let Ok(...) else` rather than the `?` operator.

crates/matrix-sdk-crypto/src/machine/mod.rs
```diff
@@ -2261,9 +2261,62 @@ impl OlmMachine {
...
+        // Check event types match, discard if not.
+        let inner_event_type = inner.event_type().to_string();
+        if outer_event_type != inner_event_type {
+            return Err(MegolmError::StateKeyVerificationFailed);
+        }
+
+        // Check state keys match, discard if not.
+        if outer_state_key != inner.state_key() {
+            return Err(MegolmError::StateKeyVerificationFailed);
+        }
```
richvdh: it would be nice to hit at least one of these code paths in a test.

richvdh (changes requested):
Generally, looks really good. Thanks for breaking it up for easy review.

A few comments, nits, and suggestsions.

crates/matrix-sdk-crypto/src/olm/group_sessions/outbound.rs
andybalaam: Will there be lots of `Payload` types? My naive expectation would be that there would only be the two that are explicitly declared.

crates/matrix-sdk-crypto/src/olm/group_sessions/outbound.rs
```diff
@@ -488,35 +534,51 @@ impl OutboundGroupSession {
...
+            #[serde(rename = "type")]
+            event_type: &'a str,
+            state_key: &'a str,
+            content: &'a Raw<AnyStateEventContent>,
+            room_id: &'a RoomId,
+        }
 
-        Raw::new(&content).expect("m.room.encrypted event content can always be serialized")
+        let payload = Payload { event_type, state_key, content, room_id: &self.room_id };
+        self.encrypt_inner(&payload, None).await.cast_unchecked()
```
andybalaam: Do we need the `cast_unchecked`? I notice we don't have it in `encrypt`.

crates/matrix-sdk-crypto/src/olm/group_sessions/outbound.rs
kaylendog: All the calls will be monomorphized but as Andy says that'll be two functions here. An alternative may be to do the serialization outside of the call and accept a raw or similar?

richvdh: Yeah, I don't think it's important enough to worry about.

crates/matrix-sdk-crypto/src/machine/tests/mod.rs
```diff
@@ -804,14 +792,61 @@ async fn test_megolm_state_encryption() {
...
+            ..
+        })
+    );
+    assert_matches!(
+        bad_state_key_decryption_result,
+        RoomEventDecryptionResult::UnableToDecrypt(UnableToDecryptInfo {
+            reason: UnableToDecryptReason::StateKeyVerificationFailed,
+            ..
+        })
+    );
```
richvdh: eh, sorry, not a fan of this style of test that tests 15 different things. Could you pull out a separate `test_megolm_state_encryption_bad_state_key` or something?

You might want to create a couple of helper functions to share functionality with the existing test.

crates/matrix-sdk-crypto/src/machine/tests/mod.rs
```diff
@@ -727,6 +732,174 @@ async fn test_megolm_encryption() {
     }
 }
 
+#[cfg(feature = "experimental-encrypted-state-events")]
```
richvdh, line 748: It would be nice to add a doc-comment explaining what this does. Something like: "Helper for testing Megolm encryption and decryption between two devices. Creates two devices, Alice and Bob, and has Alice create an outgoing Megolm session in the given room, whose decryption key is shared with Bob via a to-device message. Returns the two devices."

richvdh: Also: I think this is common to `test_megolm_encryption` as well? I don't think there is any problem with removing the `cfg` guard and using it from there: it will help break up that rather big test method into more digestible chunks.

crates/matrix-sdk-crypto/src/machine/tests/mod.rs
```diff
@@ -727,6 +732,174 @@ async fn test_megolm_encryption() {
...
+        assert_eq!(&sender, alice.user_id());
+        assert_eq!(&content.topic, plaintext);
+    } else {
+        panic!("Decrypted room event has the wrong type");
+    }
+}
+
+#[cfg(feature = "experimental-encrypted-state-events")]
+#[async_test]
+async fn test_megolm_state_encryption_bad_type() {
```
richvdh, line 842: A doc-comment giving a couple of words on what this is testing would be helpful.

"When the type encoded in the top-level `state_key` does not match that in the payload, the event should fail to decrypt."

Likewise on `test_megolm_state_encryption_bad_state_key`

crates/matrix-sdk-crypto/src/machine/tests/mod.rs
```diff
@@ -727,6 +732,174 @@ async fn test_megolm_encryption() {
...
+
+    let room_id = room_id!("!test:example.org");
+    let (alice, bob) = megolm_encryption_setup_helper(room_id).await;
+
+    let plaintext = "It is a secret to everybody";
+    let content = RoomTopicEventContent::new(plaintext.to_owned());
+    let encrypted_content =
+        alice.encrypt_state_event(room_id, content, EmptyStateKey).await.unwrap();
+
+    // Malformed events
```
richvdh: This is a bit confusing/redundant now.

```suggestion
```

crates/matrix-sdk-crypto/src/machine/tests/mod.rs
```diff
@@ -727,6 +732,174 @@ async fn test_megolm_encryption() {
...
+
+    let decryption_settings =
+        DecryptionSettings { sender_device_trust_requirement: TrustRequirement::Untrusted };
+
+    let bad_state_key_decryption_result = bob
+        .try_decrypt_room_event(&bad_state_key_event, room_id, &decryption_settings)
+        .await
+        .unwrap();
+
+    // Require malformed events fail verification
```
richvdh: this doesn't seem to make sense, and I don't think it's needed

richvdh (approved):
LGTM other than a few nits about comments in tests

crates/matrix-sdk-crypto/src/machine/tests/mod.rs
kaylendog: > Also: I think this is common to `test_megolm_encryption` as well? I don't think there is any problem with removing the `cfg` guard and using it from there: it will help break up that rather big test method into more digestible chunks.

`test_megolm_encryption` does some assertions on the `group_session` and `room_keys_received_stream` that are squirreled away into the helper, so I'm not sure it makes sense to refactor it as well?

richvdh: ok, fair enough :+1:

PR #5506, opened 2025-08-07
IndexedDB: implement `EventCacheStore::try_take_leased_lock`

mgoldenberg:
## Background

This pull request is part of a series of pull requests to add a full IndexedDB implementation of the `EventCacheStore` (see #4617, #4996, #5090, #5138, #5226, #5274, #5343, #5384, #5406, #5414, #5497). This particular pull request adds an IndexedDB-backed implementation of `EventCacheStore::try_take_leased_lock`.

## Changes

The following types have been added for representing and indexing leased locks.

- `Lease` - the primary type for representing a leased lock.
- `IndexedLease` - an indexed version of `Lease`.
- `IndexedLeaseIdKey` - the primary key of an `IndexedLease`.

To store the types above, an object store (`v1::keys::LEASES`) gets created when upgrading the database to version `1` with primary key `id`.

Additionally, the following transaction functions are added for conveniently working with leases.

- `IndexeddbEventCacheStoreTransaction::get_lease_by_id`
- `IndexeddbEventCacheStoreTransaction::put_lease`

### Tests

`matrix_sdk_base::event_cache_store_integration_tests_time` previously used `tokio::time::sleep` which is not supported on `wasm` targets.

> The time module will only work on WASM platforms that have support for timers (e.g. wasm32-wasi). The timing functions will panic if used on a WASM platform that does not support timers.

(Link to text [here](https://docs.rs/tokio/latest/tokio/index.html#wasm-support)).

I have modified this macro so that when running on `wasm` targets with an unknown operating system, we use a compatible crate, i.e., `gloo_timers`. This way, these tests can be run against the `matrix-sdk-indexeddb` crate.

## Future Work

- Proper handling of `LinkedChunkId` (see [here](https://github.com/matrix-org/matrix-rust-sdk/pull/5414#discussion_r2222985920)).
- Add implementation of `EventCacheStoreMedia`

---

- [ ] Public API changes documented in changelogs (optional)

Signed-off-by: Michael Goldenberg <m@mgoldenberg.net>

PR #5497, opened 2025-08-06
IndexedDB: remove assumption that all `EventCacheStore` keys contain a `RoomId`

mgoldenberg:
## Background

This pull request is part of a series of pull requests to add a full IndexedDB implementation of the `EventCacheStore` (see #4617, #4996, #5090, #5138, #5226, #5274, #5343, #5384, #5406, #5414). This particular pull request focuses on removing the assumption that all `EventCacheStore` keys contain a `RoomId`. This refactoring is necessary in order to support the queries related to leases and media, which make no reference to rooms - e.g., [here](https://docs.rs/matrix-sdk-base/0.13.0/matrix_sdk_base/event_cache/store/trait.EventCacheStore.html#tymethod.try_take_leased_lock), [here](https://docs.rs/matrix-sdk-base/0.13.0/matrix_sdk_base/event_cache/store/trait.EventCacheStore.html#tymethod.get_media_content_for_uri), [here](https://docs.rs/matrix-sdk-base/0.13.0/matrix_sdk_base/event_cache/store/trait.EventCacheStore.html#tymethod.media_retention_policy), etc.

### No References to `RoomId` in Serialization Traits

The overarching change is that each of the traits in [`event_cache_store::serializer::traits`](https://github.com/mgoldenberg/matrix-rust-sdk/blob/indexeddb-event-cache-store-room-id-key-component/crates/matrix-sdk-indexeddb/src/event_cache_store/serializer/traits.rs) no longer makes reference to a `RoomId`, which ultimately lays the foundation for keys which do not contain a `RoomId`. For the existing keys - all of which contain a `RoomId` - their components are now fully represented in [`IndexedKey::KeyComponents`](https://github.com/mgoldenberg/matrix-rust-sdk/blob/3054131840b21c3f3d35fedff4c1a65d6bb15b8d/crates/matrix-sdk-indexeddb/src/event_cache_store/serializer/traits.rs#L56). 

So, for instance, [`IndexedChunkIdKey`](https://github.com/matrix-org/matrix-rust-sdk/blob/872713c4bc024ac9246dfa72f834584ebe92a3d7/crates/matrix-sdk-indexeddb/src/event_cache_store/serializer/types.rs#L188) previously had the following [implementation](https://github.com/matrix-org/matrix-rust-sdk/blob/872713c4bc024ac9246dfa72f834584ebe92a3d7/crates/matrix-sdk-indexeddb/src/event_cache_store/serializer/types.rs#L190-L202).

```rust
[...]

PR #5540, opened 2025-08-17
IndexedDB: handle `LinkedChunkId` properly in `EventCacheStore` implementation

mgoldenberg:
## Background

This pull request is part of a series of pull requests to add a full IndexedDB implementation of the `EventCacheStore` (see #4617, #4996, #5090, #5138, #5226, #5274, #5343, #5384, #5406, #5414, #5497, #5506). This particular pull request adds proper handling of the `LinkedChunkId` argument provided in a number of `EventCacheStore` functions by making modifications to key formats in a number of object stores.

Prior to #5182, many functions of `EventCacheStore` were provided a `RoomId` argument. In order to accommodate the possibility of thread-based operations, `LinkedChunkId` replaced many of the `RoomId` arguments. Then, in #5445, the `Thread` variant was added to `LinkedChunkId` and would require that implementations of `EventCacheStore` properly handle thread-based operations.

This issue was surfaced by @bnjbvr [here](https://github.com/matrix-org/matrix-rust-sdk/pull/5414#discussion_r2222985920).

### `LinkedChunkId` and `OwnedLinkedChunkId`

For ease of use, I have made `OwnedLinkedChunkId::as_ref` accessible outside of `[#cfg(test)]` contexts, as well as adding `From` implementations between `LinkedChunkId` and `OwnedLinkedChunkId`. This also made it possible to consolidate the `Display` implementation for these two types.

Additionally, I have derived `Serialize` and `Deserialize` on `OwnedLinkedChunkId` in order to allow this type to be easily stored in a database, e.g., IndexedDB.

Note that these changes were discussed with @bnjbvr, though he may want to give final confirmation that this is aligned with his intention for the types.

### Keys and Indices

All existing keys replace `RoomId` with `LinkedChunkId`, with one exception: `IndexedEventRelationKey` still uses a `RoomId` in its key because [`EventCacheStore::find_event_relations`](https://docs.rs/matrix-sdk-base/0.13.0/matrix_sdk_base/event_cache/store/trait.EventCacheStore.html#tymethod.find_event_relations) is provided with a `RoomId` and not a `LinkedChunkId`.

Due to other event-related functions also continuing to operate on `RoomId`s rather than `LinkedChunkId`s, a new index (and associated key) was added to the event object store.

- Index 
    - Unique: true
    - Name: `events_room`
    - Key:
        - Path: `room`
        - Format: (`RoomId`, `EventId`) 

This index (and key) are necessary to support the following operations.

- [`EventCacheStore::find_event`](https://docs.rs/matrix-sdk-base/0.13.0/matrix_sdk_base/event_cache/store/trait.EventCacheStore.html#tymethod.find_event) - finds an event in a particular room.
- [`EventCacheStore::save_event`](https://docs.rs/matrix-sdk-base/0.13.0/matrix_sdk_base/event_cache/store/trait.EventCacheStore.html#tymethod.save_event) - saves an event to a particular room.

### Tests

Up until this pull request, `matrix_sdk_base::event_cache_store_integration_tests` was recreated in `matrix_sdk_indexeddb::event_cache_store::integration_tests` and selectively chose tests based on what was properly implemented. This has now been removed and this `EventCacheStore` implementation is now being tested against the integration tests in `matrix_sdk_base`.

## Future Work

- Add implementation of `EventCacheStoreMedia`

---


crates/matrix-sdk-common/src/linked_chunk/mod.rs
```diff
@@ -168,12 +179,7 @@ pub enum OwnedLinkedChunkId {
 
 impl Display for OwnedLinkedChunkId {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        match self {
-            OwnedLinkedChunkId::Room(room_id) => write!(f, "{room_id}"),
-            OwnedLinkedChunkId::Thread(room_id, thread_root) => {
-                write!(f, "{room_id}:thread:{thread_root}")
-            }
-        }
+        write!(f, "{}", self.as_ref())
```
Hywan: Could we simplify to `self.as_ref().fmt(f)`? It avoids the need to unfold the `write!` macro.

crates/matrix-sdk-indexeddb/src/serializer.rs
```diff
@@ -88,6 +88,27 @@ impl IndexeddbSerializer {
...
 
+    /// Hash the given key securely for the given tablename using the store
+    /// cipher.
+    ///
+    /// This works similarly to [`encode_key`](Self::encode_key), but skips
+    /// formatting and base64 encoding. This is useful for dealing with keys
+    /// that are represented as byte arrays, as it prevents having to convert
+    /// the byte array into a string and then back into a byte array.
+    ///
+    /// **Note:** that when dealing with keys which will be encoded as strings,
```
Hywan:
```suggestion
    /// **Note** that when dealing with keys which will be encoded as strings,
```

crates/matrix-sdk-indexeddb/src/event_cache_store/transaction.rs
```diff
@@ -601,6 +602,17 @@ impl<'a> IndexeddbEventCacheStoreTransaction<'a> {
         self.get_item_by_key::<Event, IndexedEventIdKey>(key).await
     }
 
+    /// Query IndexedDB for events that match the given event id in the given
+    /// room. If more than one item is found, an error is returned.
```
Hywan, line 673: Events aren't uniquely indexed by their ID?

bnjbvr:
Thanks, only read the description, and this all looks great!

crates/matrix-sdk-indexeddb/src/event_cache_store/transaction.rs
mgoldenberg: You are right, they are uniquely indexed by their ID!

That being said, the IndexedDB API's function for getting a single value - i.e., [`IDBObjectStore.get`](https://developer.mozilla.org/en-US/docs/Web/API/IDBObjectStore/get) - from an object store will return the [_"first record matching the given key"_](https://developer.mozilla.org/en-US/docs/Web/API/IDBObjectStore/get#return_value). Consequently, I have no way of knowing whether the returned record from calling this function was the one and only record or the first of a set of records.

So, throughout the codebase, I opt for getting a set of records and then verifying that there is only one. That way, if for some reason I forgot to make a particular index unique, requests for a single value that should be unique but actually match multiple records do not go unnoticed.

Does that make sense?

crates/matrix-sdk-common/src/linked_chunk/mod.rs
mgoldenberg: fixup: [`d319f19351aae37f4fcf72ea219f4f8608d2a07e`](https://github.com/matrix-org/matrix-rust-sdk/pull/5540/commits/d319f19351aae37f4fcf72ea219f4f8608d2a07e)

crates/matrix-sdk-indexeddb/src/serializer.rs
mgoldenberg: fixup: [`563f573b8e83fc51686c42f4ea48341eaa1e07c8`](https://github.com/matrix-org/matrix-rust-sdk/pull/5540/commits/563f573b8e83fc51686c42f4ea48341eaa1e07c8)

crates/matrix-sdk-indexeddb/src/event_cache_store/transaction.rs
Hywan: It does make sense, yes. Thanks for the explanation.

PR #5544, opened 2025-08-18
Add search feature to multiverse

dragonfly1033:
Adds search feature to multiverse. 
`Ctrl+S` to open search box.
Shows a list of `EventId`'s that match the query.


labs/multiverse/src/main.rs
```diff
@@ -448,6 +456,29 @@ impl App {
...
+                            && let KeyModifiers::NONE = key.modifiers
+                        {
+                            match key.code {
+                                Enter => {
+                                    if let Some(query) = view.get_text() {
+                                        if let Some(room) = self.room_view.room() {
+                                            if let Some(results) = room.search(&query, 100).await {
+                                                view.results(results);
+                                            } else {
+                                                println!("sadge no results")
```
andybalaam: Is `println!` the right way to log? I see `error` above, so maybe `warn` would be better here?

andybalaam (approved):
Looks good, 1 comment to answer. Ping me when it's ready to merge!

labs/multiverse/src/main.rs
dragonfly1033: Oops,sorry I forgot about that. I think `debug` would be sufficient here and `warn` for no room.

andybalaam (approved):

PR #5545, opened 2025-08-18
feat(sdk): Support room key downloading using `JsonCastable<EncryptedEvent>`

kaylendog:
This PR introduces feature-gated changes to room key downloading that will allow encrypted state events to trigger downloads using polymorphism. Although this will increase binary size due to monomorphisation, it allows for backwards compatibility with existing SDK code.

- [ ] Declares `AnyStateEventContent` as castable to `RoomEncryptedEventContent`.
- [ ] Declares `EncryptedEvent` as castable to `OriginalSyncStateRoomEncryptedEvent`.
- [ ] Modify `RoomKeyDownloadRequest` to accept a `Raw<EncryptedEvent>`.
- [ ] Introduces a modified `Backups::maybe_download_room_key` and `BackupDownloadTask::trigger_download_for_utd_event` that accepts any `T` castable to `EncryptedEvent`

### DIFF
--- a/crates/matrix-sdk-crypto/Cargo.toml
+++ b/crates/matrix-sdk-crypto/Cargo.toml
@@ -22,7 +22,8 @@ experimental-send-custom-to-device = []
 # Enable experimental support for encrypting state events; see
 # https://github.com/matrix-org/matrix-rust-sdk/issues/5397.
 experimental-encrypted-state-events = [
-    "matrix-sdk-common/experimental-encrypted-state-events"
+    "matrix-sdk-common/experimental-encrypted-state-events",
+    "ruma/unstable-msc3414"
 ]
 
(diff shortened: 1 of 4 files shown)


andybalaam (approved):
As far as I understand it, looks good.

I'd give extra points for the two variants of `trigger_download_for_utd_event` both being able to take `&download_request.event` so we can avoid `#[cfg` in the call to `is_room_key_available`

richvdh:
@kaylendog said:

> i think Andys suggestion would need JsonCastable<T> for T (which makes sense to me) to be added to Ruma
> i modify `RoomKeyDownloadRequest::event` to be `EncryptedEvent` when the feature is enabled, which gets passed into the call andy is referring to
> if i make `OlmMachine::is_room_key_available` accept a `T: JsonCastable<EncryptedEvent>` then this would break

PR #5547, opened 2025-08-18
feat(sdk): Add top-level support for encrypting state events

kaylendog:
Adds feature-gated support to `matrix-sdk` for encrypted state events, making modifications to `Room`.

- [ ] Add `SendRawStateEvent` future and provide a modified `Room::send_state_event_raw`, with an accompanying unit test.
- [ ] Add `SendStateEvent` future and provide a modified `Room::send_state_event` and `Room::send_state_event_for_key`.
- [ ] Adds testing utilities for encrypted state events.

### DIFF
--- a/crates/matrix-sdk/src/room/futures.rs
+++ b/crates/matrix-sdk/src/room/futures.rs
@@ -414,7 +414,7 @@ impl<'a> IntoFuture for SendRawStateEvent<'a> {
             if Self::should_encrypt(room, event_type) {
                 use tracing::debug;
 
-                Span::current().record("is_room_encrypted", true);
+                Span::current().record("should_encrypt", true);
                 debug!(
                     room_id = ?room.room_id(),
                     "Sending encrypted event because the room is encrypted.",
@@ -433,7 +433,7 @@ impl<'a> IntoFuture for SendRawStateEvent<'a> {
                 state_key = format!("{event_type}:{state_key}");
                 event_type = "m.room.encrypted";
             } else {
-                Span::current().record("is_room_encrypted", false);
+                Span::current().record("should_encrypt", false);
             }
 
             let request = send_state_event::v3::Request::new_raw(


crates/matrix-sdk/src/room/futures.rs
```diff
@@ -153,6 +162,24 @@ impl<'a> SendRawMessageLikeEvent<'a> {
...
+    }
+
+    // Query keys in case we don't have them for newly synced members.
+    //
+    // Note we do it all the time, because we might have sync'd members before
+    // sending a message (so didn't enter the above branch), but
+    // could have not query their keys ever.
+    room.query_keys_for_untracked_or_dirty_users().await?;
+    room.preshare_room_key().await?;
+
```
richvdh: This isn't part of querying keys, so needs a space from that and its comment.

```suggestion

    room.preshare_room_key().await?;
```

crates/matrix-sdk/src/room/futures.rs
```diff
@@ -153,6 +162,24 @@ impl<'a> SendRawMessageLikeEvent<'a> {
...
+    Ok(())
+}
```
richvdh: This probably needs to go at the bottom of the file, rather than in the middle of the `impl`s for `SendRawMessageLikeEvent`

crates/matrix-sdk/src/room/futures.rs
```diff
@@ -319,3 +335,188 @@ impl<'a> IntoFuture for SendAttachment<'a> {
...
+
+        let fut = async move {
+            room.ensure_room_joined()?;
+
+            let mut state_key = state_key.to_owned();
+
+            if Self::should_encrypt(room, event_type) {
+                use tracing::debug;
+
+                Span::current().record("is_room_encrypted", true);
```
richvdh: the name of this field is slightly misleading... maybe:

```suggestion
                Span::current().record("should_encrypt", true);
```

... and likewise below for the `false` case`.

richvdh (approved):
LGTM otherwise

crates/matrix-sdk/src/room/futures.rs
kaylendog: `SendRawMessageLikeEvent` currently uses `is_room_encrypted`, I can change in this PR and then open one to modify that there?

richvdh: I don't necessarily think `SendRawMessageLikeEvent` needs to match, tbh.

crates/matrix-sdk/src/room/futures.rs
```diff
@@ -319,3 +317,207 @@ impl<'a> IntoFuture for SendAttachment<'a> {
...
+                Span::current().record("should_encrypt", true);
+                debug!(
+                    room_id = ?room.room_id(),
+                    "Sending encrypted event because the room is encrypted.",
+                );
+
+                ensure_room_encryption_ready(room).await?;
+
+                let olm = room.client.olm_machine().await;
+                let olm = olm.as_ref().expect("Olm machine wasn't started");
```
andybalaam, line 426: Is unwrapping what we normally do when the Olm machine is not ready?

crates/matrix-sdk/tests/integration/encryption/state_events.rs
```diff
@@ -0,0 +1,80 @@
...
+        .mount()
+        .await;
+
+    // Get room and send a topic state event.
+    alice
+        .get_room(room_id)
+        .expect("Alice should have access to the room")
+        .send_state_event(RoomTopicEventContent::new("Encrypted topic".to_owned()))
+        .await
+        .unwrap();
```
andybalaam, line 75: Can we assert anything here? E.g. the event arrived, it was encrypted, it can be decrypted?

andybalaam (changes requested):
Looks good - I'd just like to see some assertions in that test, so we can confirm things are working.

crates/matrix-sdk/tests/integration/encryption/state_events.rs
kaylendog: The `mock_room_send_state` call above does exactly this! It's looking for a `m.room.encrypted` event with state key `m.room.topic:` and checks that the body contains the `algorithm` and `device_id` fields. As far as I can tell these assertions get tested when the mock gets dropped (so at the end of the function body).

crates/matrix-sdk/src/room/futures.rs
kaylendog: Yes, the existing `SendRawMessageLike` future asserts this on [line 205](https://github.com/matrix-org/matrix-rust-sdk/blob/f27e733e965c78e8f9614633e22ca4b2d0e7533d/crates/matrix-sdk/src/room/futures.rs#L205).

crates/matrix-sdk/tests/integration/encryption/state_events.rs
andybalaam: Great, thanks!

andybalaam (approved):

PR #5551, opened 2025-08-19
fix(ci): free more space in the CI runner for the `codecov` task

bnjbvr:
This frees 13 GB of disk space in the Codecov image runner, by getting rid of tools, binaries and libs we're not interested in.

These tools have been identified thanks to the new lines added to detect large files, e.g. `du -hsc /usr/local/bin/* | sort -h`:

- `du -hsc /usr/local/bin/*` will output a summary of the size, in human size format, of every top-level entry in `/usr/local/bin`
- `sort -h` will sort from smallest to largest, in the human size format too

[Example](https://github.com/matrix-org/matrix-rust-sdk/actions/runs/17067856858/job/48389222421#step:4:89).

This idea courtesy of apache/arrow, which has a [script](https://github.com/apache/arrow/blob/main/ci/scripts/util_free_space.sh) to remove such unused space (but many binaries seem to not be present in our image at start, so it might be a bit outdated).

Should help unblocking the codecov step for #5544.

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -43,9 +43,58 @@ jobs:
       # space here.
       - name: Create some more space
         run: |
+          echo "Disk space before cleanup"
+          df -h
+
           cd /opt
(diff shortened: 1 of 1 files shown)


bnjbvr:
This is now ready for review!

Hywan:
It frees 20Gb. Kudos.

PR #5552, opened 2025-08-19
feat: Add top-level support for decrypting state events

kaylendog:
Implements support for decryption of state events. Needs #5545.

- [ ] Introduce a case for `AnySyncStateEvent::RoomEncrypted` to the `state_events` sync response processor.
- [ ] Introduce modified `Room::decrypt_event` and `::try_decrypt_room_event`.
- [ ] Introduce testing macro `assert_let_decrypted_state_event_content`.
- [ ] Add casts and explicit type hints where necessary.

### DIFF
--- a/crates/matrix-sdk/src/room/mod.rs
+++ b/crates/matrix-sdk/src/room/mod.rs
@@ -32,6 +32,8 @@ use futures_util::{
 use http::StatusCode;
 #[cfg(feature = "e2e-encryption")]
 pub use identity_status_changes::IdentityStatusChanges;
+#[cfg(feature = "experimental-encrypted-state-events")]
+use matrix_sdk_base::crypto::types::events::room::encrypted::EncryptedEvent;
 #[cfg(feature = "e2e-encryption")]
 use matrix_sdk_base::crypto::{IdentityStatusChange, RoomIdentityProvider, UserIdentity};
 pub use matrix_sdk_base::store::ThreadSubscription;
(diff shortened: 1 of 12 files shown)


crates/matrix-sdk/src/room/mod.rs
```diff
@@ -636,17 +645,21 @@ impl Room {
...
     ///
     /// Only logs from the crypto crate will indicate a failure to decrypt.
+    #[cfg(not(feature = "experimental-encrypted-state-events"))]
     #[allow(clippy::unused_async)] // Used only in e2e-encryption.
     async fn try_decrypt_event(
         &self,
         event: Raw<AnyTimelineEvent>,
         push_ctx: Option<&PushContext>,
     ) -> TimelineEvent {
+        // If we have either an encrypted message-like or state event, try to decrypt.
```
richvdh: is this correct? we don't seem to check for state events here?

crates/matrix-sdk/src/room/mod.rs
```diff
@@ -636,17 +645,21 @@ impl Room {
...
         #[cfg(feature = "e2e-encryption")]
         if let Ok(AnySyncTimelineEvent::MessageLike(AnySyncMessageLikeEvent::RoomEncrypted(
             SyncMessageLikeEvent::Original(_),
         ))) = event.deserialize_as::<AnySyncTimelineEvent>()
         {
+            // Cast safety: The state key is not used during decryption, and the types
+            // overlap sufficiently.
```
richvdh: what state key?

crates/matrix-sdk/src/room/mod.rs
kaylendog: Seems to be an extraneous comment, will remove this and the other you addressed below.

richvdh (approved):
Thanks! I don't love the duplication of methods that we're sprouting here, but I guess the alternatives suck a bit too.

PR #5553, opened 2025-08-19
Dragonfly1033/listen updates

dragonfly1033:
Remove previous code that got updates in `RoomEventCacheState::post_process_new_events`.
Add new task in `EventCache` that subscribes to `linked_chunk_update_sender` and forwards new events to `client::search::SearchIndex` same as before.


crates/matrix-sdk/src/event_cache/mod.rs
```diff
@@ -683,6 +693,79 @@ impl EventCache {
...
+    /// the event type.
+    #[cfg(feature = "experimental-search")]
+    async fn search_indexing_task(
+        client: WeakClient,
+        linked_chunk_update_sender: Sender<RoomEventCacheLinkedChunkUpdate>,
+    ) {
+        let Some(client) = client.get() else {
+            trace!("Client is shutting down, not spawning thread subscriber task");
+            return;
+        };
```
bnjbvr: Doing this will upgrade the `weak` client, making it into a strong reference. As a result, it won't be dropped when the `Client` drops, because the `Client` owns the `EventCache`, which in turns owns this task, which in turn owns the `Client` — creating a cycle.

I think you'd need to put this code below, just before the `for event in timeline_events` loop.

crates/matrix-sdk/src/event_cache/mod.rs
```diff
@@ -683,6 +693,79 @@ impl EventCache {
...
+                    };
+
+                    let timeline_events = room_ec_lc_update.events();
+
+                    for event in timeline_events {
+                        if let Some(message_event) = Self::parse_timeline_event(&event) {
+                            let _ = client
+                                .search_index()
+                                .lock()
+                                .await
```
bnjbvr: To help perf, you can lock the `search_index` only once, above the loop, if `!timeline_events.is_empty()`, to avoid locking/unlocking in a loop. This is called "loop invariant code motion", in compiler engineering; hoist the "constant" parts of a loop on top of it, to avoid repeating the same work.

crates/matrix-sdk/src/event_cache/mod.rs
```diff
@@ -683,6 +693,79 @@ impl EventCache {
...
+                                .handle_event(message_event, &room_id)
+                                .map_err(EventCacheError::from)
+                                .inspect_err(|err| {
+                                    warn!("Failed to handle event for indexing: {err:?}")
+                                });
```
bnjbvr: - we can remove the `map_err` call here, since the error is only used for inspection purposes
- an error can be `Display`ed, so you can use `{err}` instead of `{err:?}` (which would invoke the `Debug` impl instead)
- since we're only inspecting the error, and not doing anything with the result, we could transform this statement into a `if let Err(err) = … .handle_event(…) { warn!("Failed to handle event for indexing: {err}")`

crates/matrix-sdk/src/event_cache/pagination.rs
```diff
@@ -281,13 +281,7 @@ impl RoomPagination {
...
             .write()
             .await
-            .handle_backpagination(
-                response.chunk,
-                response.end,
-                prev_token,
-                #[cfg(feature = "experimental-search")]
-                &room,
-            )
+            .handle_backpagination(response.chunk, response.end, prev_token)
```
bnjbvr: Sweet! Can you also revert to the previous version of the code, which kept the `Room` instance in a smaller sub-scope, around the pagination itself?

crates/matrix-sdk/src/event_cache/room/mod.rs
```diff
@@ -1513,12 +1459,6 @@ mod private {
             for event in events {
                 self.maybe_apply_new_redaction(&event).await?; // TODO: Handle redaction for search index
```
bnjbvr: Let's also remove the `TODO` here; it's going to go to the search indexing task instead.

crates/matrix-sdk/src/room/mod.rs
```diff
@@ -3702,18 +3697,6 @@ impl Room {
...
 
-    /// Handle an [`AnySyncMessageLikeEvent`] in this room's [`RoomIndex`].
-    ///
-    /// This which will add/remove/edit an event in the index based on the
-    /// event type.
-    #[cfg(feature = "experimental-search")]
-    pub(crate) async fn index_event(
-        &self,
-        event: AnySyncMessageLikeEvent,
-    ) -> Result<(), IndexError> {
```
bnjbvr, line 3713: I think you can remove the `EventCacheError::IndexError`, with all the changes I've proposed in this PR?

crates/matrix-sdk/src/event_cache/mod.rs
```diff
@@ -683,6 +693,79 @@ impl EventCache {
...
+                None
+            }
+        }
+    }
+
+    /// Takes a [`TimelineEvent`] and passes it to the [`RoomIndex`] of the
+    /// given room which will add/remove/edit an event in the index based on
+    /// the event type.
+    #[cfg(feature = "experimental-search")]
+    async fn search_indexing_task(
```
bnjbvr, line 695: Can you add an `#[instrument(skip_all)]` line on top of this, so the log lines would be tagged accordingly?

crates/matrix-sdk/src/event_cache/mod.rs
```diff
@@ -683,6 +693,79 @@ impl EventCache {
             }
         }
     }
+
+    #[cfg(feature = "experimental-search")]
+    fn parse_timeline_event(event: &TimelineEvent) -> Option<AnySyncMessageLikeEvent> {
```
bnjbvr: Can you make this a freestanding function? It's weird that this is a method attached to the `EventCache` (and it was already a bit surprising to have it in the previous place where it was).

crates/matrix-sdk/src/event_cache/mod.rs
bnjbvr: Hah nice, we had a test for this!

https://github.com/matrix-org/matrix-rust-sdk/actions/runs/17071224044/job/48400427680?pr=5553#step:8:458

crates/matrix-sdk/src/event_cache/mod.rs
```diff
@@ -683,6 +686,80 @@ impl EventCache {
...
+                Err(RecvError::Lagged(num_skipped)) => {
+                    warn!(num_skipped, "Lagged behind linked chunk updates");
+                }
+            }
+        }
+    }
+}
+
+#[cfg(feature = "experimental-search")]
+fn parse_timeline_event(event: &TimelineEvent) -> Option<AnySyncMessageLikeEvent> {
```
bnjbvr: Let's rename this to make it clear it's related to search indexing.

crates/matrix-sdk/src/event_cache/mod.rs
```diff
@@ -683,6 +686,80 @@ impl EventCache {
...
+                                    {
+                                        warn!("Failed to handle event for indexing: {err}")
+                                    }
+                                }
+                            }
+                        } else {
+                            trace!("Client is shutting down, not spawning thread subscriber task");
+                            return;
+                        }
+                    }
```
bnjbvr: To reduce indent here, we can do the following:

```rust
if timeline_events.peek().is_none() {
  continue;
}

let Some(client) = client.get() else {
  trace!("Client is shutting down etc");
  return;
};

// then the interesting part
```

crates/matrix-sdk/src/event_cache/mod.rs
```diff
@@ -738,6 +815,15 @@ struct EventCacheInner {
     /// [`EventCache`], so it does listen to *all* rooms at the same time.
     _thread_subscriber_task: AbortOnDrop<()>,
 
+    /// A background task listening to room and send queue updates, and
```
bnjbvr: nit: we're only listening to room updates, not send queue updates here

bnjbvr (approved):
Looks great, thanks!

crates/matrix-sdk/src/event_cache/mod.rs
dragonfly1033: I've had mixed messages on using this guard clause pattern in the project. Personally, I prefer them but I have been told that I shouldn't because it's (paraphrase) "harder to follow".

I'll change it here anyway since we both prefer it.

crates/matrix-sdk/Cargo.toml
```diff
@@ -16,7 +16,7 @@ features = ["docsrs"]
 rustdoc-args = ["--cfg", "docsrs", "--generate-link-to-definition"]
 
 [features]
-default = ["e2e-encryption", "automatic-room-key-forwarding", "sqlite", "native-tls"]
+default = ["e2e-encryption", "automatic-room-key-forwarding", "sqlite", "native-tls", "experimental-search"]
```
bnjbvr: oopsy, actually?

Issue #5554, opened 2025-08-19
Bad contention of the cross-process lock / sqlite write locks


bnjbvr:
- The event cache store uses a cross-process lock to make sure that there's at most one process writing to its DB.
- At another layer, the sqlite backend uses a [write lock](https://github.com/matrix-org/matrix-rust-sdk/pull/5382) to prevent against bad `database busy` errors, which has proven super effective.

Unfortunately, it seems that the interaction of both can dramatically slow down the SDK: taking the cross-process lock will read/write into the DB into a "leases" database, making use of the sqlite write lock.

This is even worse, because the event cache store is also used for medias at the moment; ideally it'd be split off, and there's https://github.com/matrix-org/matrix-rust-sdk/issues/5410 for that.

In the meanwhile: we should also explore ways to reduce contention around the cross-process lock, so as to keep it as long as possible, maybe for a full sync update, and pass it down to methods that need it, instead of taking it randomly as we see fit. We might need to be careful around the dining philosopher's problem, though; if our locks are more fine-grained, then there are more chances we're taking multiple ones at the same time.

bnjbvr:
This issue originally named `Bad contention of the cross-process lock / sqlite write locks`, but I'm being less and less sure this is only about that, so broadening a bit here.

Hywan:
I've found https://github.com/matrix-org/matrix-rust-sdk/issues/5572, maybe it's the slowness cause we were looking for.

PR #5555, opened 2025-08-19
feat(event cache): take the cross-process lock once per sync update

bnjbvr:
An attempt at taming https://github.com/matrix-org/matrix-rust-sdk/issues/5554.


kathleenmotley49-maker (approved):

bnjbvr:
This has bitrotten, so I'll close it, and also the numbers were not as great as I expected. We'd need a structured way to look at lock contention.

PR #5556, opened 2025-08-19
feat(sdk): Introduce the new `LatestEvents` API, part 4: The Send Queue has the priority over the Event Cache

Hywan:
In this patch, we are implementing a requirement where a `LatestEventValue` created by a send queue update cannot be overwritten by an event cache update.

We are also fixing another subtlety: when a local event is sent, the `LatestEventValue` represents this just sent local event as long as an event cache update is received.

---

### DIFF
--- a/crates/matrix-sdk/src/latest_events/latest_event.rs
+++ b/crates/matrix-sdk/src/latest_events/latest_event.rs
@@ -79,7 +79,7 @@ impl LatestEvent {
     }
 
     /// Update the inner latest event value, based on the event cache
-    /// (specifically with a [`RoomEventCache`]), if and only if there is no
+    /// (specifically with the [`RoomEventCache`]), if and only if there is no
     /// local latest event value waiting.
     ///
     /// It is only necessary to compute a new [`LatestEventValue`] from the
@@ -94,7 +94,7 @@ impl LatestEvent {
     ) {
         if self.buffer_of_values_for_local_events.is_empty().not() {
             // At least one `LatestEventValue` exists for local events (i.e. coming from the
-            // send queue). In this case, we don't overwrite the current value with a new
+            // send queue). In this case, we don't overwrite the current value with a newly
             // computed one from the event cache.
             return;
         }
@@ -106,7 +106,7 @@ impl LatestEvent {
     }
 
     /// Update the inner latest event value, based on the send queue
-    /// (specifically with a [`RoomSendQueueUpdate`]).
+    /// (specifically with the [`RoomSendQueueUpdate`]).
     pub async fn update_with_send_queue(
         &mut self,
         send_queue_update: &RoomSendQueueUpdate,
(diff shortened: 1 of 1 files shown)


crates/matrix-sdk/src/latest_events/latest_event.rs
```diff
@@ -244,16 +456,30 @@ impl LatestEventValue {
...
-                Self::new_local_or_remote(
+                // First, compute the new value. Then we remove the sent local event from the
+                // buffer.
+                //
+                // Why in this order? Because once the local event is sent, it's not yet
+                // received by the sync and consequently not stored in the event cache. So once
+                // the local event is sent, it won't show up as a `LatestEventValue` as it will
+                // immediately be replaced by an event from the event cache. By computing the
+                // new value before removing it from the buffer, we ensure the
+                // `LatestEventValue` represents the just sent local event.
```
stefanceriu:
```suggestion
                // Why in this order? Because in between sending and remote echoing the event 
                // will only be stored as an local echo and not as a full event in the event cache. 
                // Just after sending it won't show up as a `LatestEventValue` as it will
                // immediately be replaced by an event from the event cache. By computing the
                // new value before removing it from the buffer, we ensure the
                // `LatestEventValue` represents the just sent local event.
```

stefanceriu: Found yours a bit hard to parse initially and I think using the remote/local echo naming scheme makes things easier to follow.

crates/matrix-sdk/src/latest_events/latest_event.rs
```diff
@@ -79,12 +79,26 @@ impl LatestEvent {
     }
 
     /// Update the inner latest event value, based on the event cache
-    /// (specifically with a [`RoomEventCache`]).
+    /// (specifically with a [`RoomEventCache`]), if and only if there is no
```
stefanceriu:
```suggestion
    /// (specifically the [`RoomEventCache`]), if and only if there is no
```

crates/matrix-sdk/src/latest_events/latest_event.rs
```diff
@@ -79,12 +79,26 @@ impl LatestEvent {
...
+    /// event cache if there is no [`LatestEventValue`] to be compute from the
+    /// send queue. Indeed, anything coming from the send queue has the priority
+    /// over the anything coming from the event cache. We believe it provides a
+    /// better user experience.
     pub async fn update_with_event_cache(
         &mut self,
         room_event_cache: &RoomEventCache,
         power_levels: &Option<(&UserId, RoomPowerLevels)>,
     ) {
+        if self.buffer_of_values_for_local_events.is_empty().not() {
```
stefanceriu, line 95: Imho reads better with `if !self.buffer_of_values_for_local_events.is_empty()`

crates/matrix-sdk/src/latest_events/latest_event.rs
```diff
@@ -79,12 +79,26 @@ impl LatestEvent {
...
+            // At least one `LatestEventValue` exists for local events (i.e. coming from the
+            // send queue). In this case, we don't overwrite the current value with a new
```
stefanceriu:
```suggestion
            // send queue). In this case, we don't overwrite the current value with a newly
```

Hywan: Thanks!

crates/matrix-sdk/src/latest_events/latest_event.rs
Hywan: Thanks!

crates/matrix-sdk/src/latest_events/latest_event.rs
Hywan: Thanks for the rephrasing suggestion!

PR #5557, opened 2025-08-19
feat(sdk): Add `Room::enable_encryption_with_state`

kaylendog:
- [ ] Adds `Room::enable_encryption_with_state` to enable state event encryption in a room.
- [ ] Adds testing utilities used in the provided unit test to verify the above


Hywan (changes requested):
The code looks good, thanks. I'm not sure about the new method name though: `Room::enable_encryption_with_state`, I think the name isn't clear. Since we have an `EncryptionState` type, one can expect `enable_encryption_with_state` to take… an `EncryptionState`! It's not clear that state refers to state events here.

Thoughts? Can we think of a different name?

Hywan (approved):

PR #5559, opened 2025-08-20
feat(sdk): Poll encryption state on sync event for up to 3 seconds.

kaylendog:
This fixes a small issue encountered while developing the [integration test](https://github.com/matrix-org/matrix-rust-sdk/pull/5456#discussion_r2260782187) for encrypted state events. If the first sync response received from the server after enabling encryption does not contain the `m.room.encryption` event, the client will report the encryption state as unknown, even if the next sync response does contain the event.

- [ ] Modify `Room::enable_encryption_inner` to spin on the room encryption state, rather than bailing after the first sync response is received.

Signed-off-by: Skye Elliot

### DIFF
--- a/crates/matrix-sdk/src/room/mod.rs
+++ b/crates/matrix-sdk/src/room/mod.rs
@@ -1929,9 +1929,24 @@ impl Room {
             )
             .await;
 
-            // If encryption was enabled, return.
             let _sync_lock = self.client.base_client().sync_lock().lock().await;
+
+            // If encryption was enabled, return.
+            #[cfg(not(feature = "experimental-encrypted-state-events"))]
             if res.is_ok() && self.inner.encryption_state().is_encrypted() {
+                debug!("room successfully marked as encrypted");
+                return Ok(());
+            }
+
+            // If encryption with state event encryption was enabled, return.
+            #[cfg(feature = "experimental-encrypted-state-events")]
+            if res.is_ok() && {
+                if encrypted_state_events {
+                    self.inner.encryption_state().is_state_encrypted()
+                } else {
+                    self.inner.encryption_state().is_encrypted()
+                }
+            } {
                 debug!("room successfully marked as encrypted");
                 return Ok(());
             }


crates/matrix-sdk/src/room/mod.rs
```diff
@@ -1908,29 +1908,47 @@ impl Room {
...
             // could be quite useful if someone wants to enable encryption and
             // send a message right after it's enabled.
-            _ = timeout(self.client.inner.sync_beat.listen(), SYNC_WAIT_TIME).await;
+            let res = timeout(
+                async {
+                    loop {
+                        // Listen for sync events, then check if the encryption state has been set.
+                        self.client.inner.sync_beat.listen().await;
+                        let _sync_lock = self.client.base_client().sync_lock().lock().await;
+                        if self.inner.encryption_state().is_encrypted() {
```
stefanceriu: Are we not interested in **any** encryption state that's not `Unknown` here? 🤔

kaylendog: You make a valid point! I shall change.

stefanceriu:
Okay, now this makes sense to me but I would still like somebody from the crypto team to have a quick look.

crates/matrix-sdk/src/room/mod.rs
stefanceriu: Thanks! 👍

crates/matrix-sdk/src/room/mod.rs
```diff
@@ -1908,29 +1908,62 @@ impl Room {
...
+            if res.is_ok() && {
+                if encrypted_state_events {
+                    self.inner.encryption_state().is_state_encrypted()
+                } else {
+                    self.inner.encryption_state().is_encrypted()
+                }
+            } {
                 debug!("room successfully marked as encrypted");
+                return Ok(());
             }
```
kaylendog, line 1952: This might belong in its own PR?

richvdh: you mean, separate the "poll for 3 seconds" part from the "make it support encrypted state" part? Yes, a separate PR would be optimal, but I think it's fine in practice.

PR #5560, opened 2025-08-20
feat(element-call url params): split url params into configuration and properties

toger5:
This PR is part of an onging effort to move responsiblity to the EC app and out of the EX apps.

4 intends (f.ex `join_existing` `start_new_dm`... ) (as url paramters) are introduced in recent element call versions. Those intends behave like defaults. If an intend is set a set of url parameters are predefined.
Not all params can be covered by the intend (for insteance the `widget_id` or the `host_url`).
This PR splits the url parameters into configuration (things that can be configured by the intent) and properties (things that still need to be passed one by one)

The goal with this change is that EX only needs to configre the intent once and the EC codebase can update the behavior in those 4 specific scenarios in case new features come along (auto hangup when other participants leave, send call ring notification...)

Signed-off-by: Timo K <toger5@hotmail.de>

Signed-off-by:

### DIFF
--- a/crates/matrix-sdk/CHANGELOG.md
+++ b/crates/matrix-sdk/CHANGELOG.md
@@ -84,6 +84,40 @@ All notable changes to this project will be documented in this file.
   ([#5431](https://github.com/matrix-org/matrix-rust-sdk/pull/5431))
 - [**breaking**] `Room::send_call_notification` and `Room::send_call_notification_if_needed` have been removed, since the event type they send is outdated, and `Client` is not actually supposed to be able to join MatrixRTC sessions (yet). In practice, users of these methods probably already rely on another MatrixRTC implementation to participate in sessions, and such an implementation should be capable of sending notifications itself.
   ([#5452](https://github.com/matrix-org/matrix-rust-sdk/pull/5452))
+- [**breaking**] The `new_virtual_element_call_widget` now uses a `props` and a `config` parameter instead of only `props`.
+  This splits the configuration of the widget into required properties ("widget_id", "parent_url"...) so the widget can work
+  and optional config parameters ("skip_lobby", "header", "...").
+  The config option should in most cases only provide the `"intent"` property.
+  All other config options will then be chosen by EC based on platform + `intent`.
+
+  Before:
+
+  ```rust
+  new_virtual_element_call_widget(
+    VirtualElementCallWidgetProperties {
+      widget_id: "my_widget_id", // required property
+      skip_lobby: Some(true), // optional configuration
+      preload: Some(true), // optional configuration
+      // ...
+    }
+  )
+  ```
+
+  Now:
+
+  ```rust
+  new_virtual_element_call_widget(
+    VirtualElementCallWidgetProperties {
+      widget_id: "my_widget_id", // required property
+      // ... only required properties
+    },
+    VirtualElementCallWidgetConfig {
+      intend: Intend.StartCallDM, // defines the default values for all other configuration
+      skip_lobby: Some(false), // overwrite a specific default value
+      ..VirtualElementCallWidgetConfig::default() // set all other config options to `None`. Use defaults from intent
+    }
+  )
+  ```
 
 ### Bugfix
 
--- a/crates/matrix-sdk/CHANGELOG.md
+++ b/crates/matrix-sdk/CHANGELOG.md
@@ -114,7 +114,7 @@ All notable changes to this project will be documented in this file.
     VirtualElementCallWidgetConfig {
       intend: Intend.StartCallDM, // defines the default values for all other configuration
       skip_lobby: Some(false), // overwrite a specific default value
-      ..VirtualElementCallWidgetConfig::default() // set all other config options to `None`. Use defaults from intent
+      ..VirtualElementCallWidgetConfig::default() // set all other config options to `None`. Use defaults from intent.
     }
   )
   ```
--- a/crates/matrix-sdk/src/widget/settings/element_call.rs
+++ b/crates/matrix-sdk/src/widget/settings/element_call.rs
@@ -26,9 +26,9 @@ use super::{url_params, WidgetSettings};
 
 #[derive(Serialize)]
 #[serde(rename_all = "camelCase")]
-/// Parameters for the Element Call widget.
+/// Serialization struct for URL parameters for the Element Call widget.
 /// These are documented at https://github.com/element-hq/element-call/blob/livekit/docs/url-params.md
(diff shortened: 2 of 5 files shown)


crates/matrix-sdk/CHANGELOG.md
```diff
@@ -55,7 +55,11 @@ All notable changes to this project will be documented in this file.
 - [**breaking**] The MSRV has been bumped to Rust 1.88.
   ([#5431](https://github.com/matrix-org/matrix-rust-sdk/pull/5431))
 - [**breaking**] `Room::send_call_notification` and `Room::send_call_notification_if_needed` have been removed, since the event type they send is outdated, and `Client` is not actually supposed to be able to join MatrixRTC sessions (yet). In practice, users of these methods probably already rely on another MatrixRTC implementation to participate in sessions, and such an implementation should be capable of sending notifications itself.
-
+- [**breaking**] The `new_virtual_element_call_widget` now uses a `props` and a `config` parameter instead of only `props`.
+  This splits the configuration of the widget into required properties ("widget_id", "parent_url"...) so the widget can work
+  and optional config parameters ("skip_lobby", "header" ...).
+  The config option should in most cases only provide the `"intent"` property.
+  All other config options will then be chosen by EC based on platform + `intent`.
```
Hywan: Can you add backtick around `"…"` please?

Can you add a before/after example too? That would be helpful for users.

Hywan (changes requested):
Thanks for the patches.

Can you amend your patches to provide a description please? Your first patch's description is “d”, and the second is “fix tests”. I don't think “d” is enough information to help the reviewer to understand it.

crates/matrix-sdk/src/widget/settings/element_call.rs
```diff
@@ -44,13 +44,13 @@ struct ElementCallParams {
     /// compatibility. Set to `true` if intent is `Intent::StartCall`.
     skip_lobby: Option<bool>,
     confine_to_room: bool,
-    app_prompt: bool,
+    app_prompt: Option<bool>,
```
Hywan, line 70: What `Option<bool>` means here? Specifically, how `Some(false)` is different from `None`?

crates/matrix-sdk/src/widget/settings/element_call.rs
```diff
@@ -59,7 +59,7 @@ struct ElementCallParams {
     font_scale: Option<f64>,
     font: Option<String>,
     #[serde(rename = "perParticipantE2EE")]
-    per_participant_e2ee: bool,
+    per_participant_e2ee: Option<bool>,
```
Hywan, line 85: Same question as above.

crates/matrix-sdk/src/widget/settings/element_call.rs
```diff
@@ -44,13 +44,13 @@ struct ElementCallParams {
...
-    app_prompt: bool,
+    app_prompt: Option<bool>,
     /// Supported since Element Call v0.13.0.
-    header: HeaderStyle,
+    header: Option<HeaderStyle>,
     /// Deprecated since Element Call v0.13.0. Included for backwards
     /// compatibility. Use header: "standard"|"none" instead.
     hide_header: Option<bool>,
-    preload: bool,
+    preload: Option<bool>,
```
Hywan, line 76: Same question as above.

crates/matrix-sdk/src/widget/settings/element_call.rs
```diff
@@ -73,8 +73,10 @@ struct ElementCallParams {
     sentry_dsn: Option<String>,
     /// Supported since Element Call v0.9.0. Only used by the embedded package.
     sentry_environment: Option<String>,
-    hide_screensharing: bool,
-    controlled_media_devices: bool,
+    /// Supported since Element Call v0.9.0.
+    hide_screensharing: Option<bool>,
```
Hywan, line 100: Same question as above.

crates/matrix-sdk/src/widget/settings/element_call.rs
```diff
@@ -195,6 +191,53 @@ pub struct VirtualElementCallWidgetOptions {
     /// Default: `true`
     pub confine_to_room: Option<bool>,
 
+    /// Do not show the screenshare button.
+    pub hide_screensharing: Option<bool>,
```
Hywan, line 223: Same question as above :-).

crates/matrix-sdk/src/widget/settings/element_call.rs
```diff
@@ -195,6 +191,53 @@ pub struct VirtualElementCallWidgetOptions {
...
+
+    /// Make the audio devices be controlled by the os instead of the
+    /// element-call webview.
+    pub controlled_audio_devices: Option<bool>,
```
Hywan, line 227: Guess what ^^?

crates/matrix-sdk/src/widget/settings/element_call.rs
```diff
@@ -195,6 +191,53 @@ pub struct VirtualElementCallWidgetOptions {
...
+    /// it might break custom implementations.
+    ///
+    /// Defaults to `element_call_url` for the non-iframe (dedicated webview)
+    /// usecase.
+    pub parent_url: Option<String>,
+
+    /// The font scale which will be used inside element call.
+    ///
+    /// Default: `1`
+    pub font_scale: Option<f64>,
```
Hywan, line 267: What happens if it's 0? Is it allowed?

toger5:
@Hywan sorry for the high latency on this PR. (Holidays)

I will work on a rebase/merge today.

> What Option<bool> means here? Specifically, how Some(false) is different from None?

The `ElementCallParams` are directly translated into url query parameters.
All fields are optionals where the `None` case implies, that it will not be part of the url parameters.

For example:

```
ElementCallParams {
   hide_screensharing: Some(true),
   ..ElementCallParams::default()
}
```
will become:
```
my.url?hide_screensharing=true
```
The reason it might be desirable to not list those configurations in the urls parameters is that the `intent` implies default for all configuration values in widget itself. Setting the url parameter specifically will overwrite those defaults. So it often is desired for them to be `None` values (not part of the url).

I will update the docstring to contain this information as well.

crates/matrix-sdk/src/widget/settings/element_call.rs
toger5: It would set the font size to 0. Since EC is a webview this would imply that text is basically "not rendered".

I think this is the expected behavior. Since css allso does not do any specific allow rules on this I feel like the url params can have the same flexibility.

PR #5561, opened 2025-08-20
feat(sdk): `LatestEventValue` implements `Serialize` and `Deserialize`

Hywan:
The problem is: `LatestEventContent` cannot be serialized. It's annoying because it means we can't store a `LatestEventValue` (that wraps a `LatestEventContent`) in the database.

This patch revisits `LatestEventValue`. Before we got:

```rust
pub enum LatestEventValue {
     None,
     Remote(LatestEventContent),
     LocalIsSending(LatestEventContent),
     LocalCannotBeSent(LatestEventContent),
}

pub enum LatestEventContent {
    RoomMessage(RoomMessageEventContent),
    Sticker(StickerEventContent),
    Poll(UnstablePollStartEventContent),
    CallInvite(CallInviteEventContent),
    CallNotify(CallNotifyEventContent),
    KnockedStateEvent(RoomMemberEventContent),
    Redacted(AnySyncMessageLikeEvent),
}
```

`LatestEventContent::Redacted` contains an `AnySyncMessageLikeEvent`. That's the part that is not serializable.

It appears that `LatestEventContent` isn't necessary! The only thing we need is to _filter_ the events by their type, no need to _find and map_. The `LatestEventValue` can contain the entry event directly (e.g. a `TimelineEvent` for the event cache). Okay, let's do that.

```rust
pub enum LatestEventValue {
    None,
    Remote(RemoteLatestEventValue),
    LocalIsSending(???),
    LocalCannotBeSent(???),
}

type RemoteLatestEventValue = TimelineEvent;
```

What about the `Local*` variants? We can't use a `TimelineEvent`. We need a new type for that:

```rust
pub enum LatestEventValue {
    None,
    Remote(RemoteLatestEventValue),
    LocalIsSending(LocalLatestEventValue),
    LocalCannotBeSent(LocalLatestEventValue),
}

pub struct LocalLatestEventValue {
    pub timestamp: MilliSecondsSinceUnixEpoch,
    pub content: Raw<AnyMessageLikeEventContent>,
    pub event_type: String,
}
```

We don't need the event ID nor the transaction ID in `LocalLatestEventValue`.

That's the only change. All the other changes are about the tests.

---


bnjbvr:
I think that instead of:

```
    pub content: Raw<AnyMessageLikeEventContent>,
    pub event_type: String,
```

Could you use `SerializableEventContent`, which likely encodes the semantics you're interested in?

crates/matrix-sdk/src/latest_events/latest_event.rs
```diff
@@ -341,14 +344,38 @@ pub enum LatestEventValue {
...
+#[derive(Debug, Clone)]
+pub struct LocalLatestEventValue {
+    /// The time where the event has been created (by this module).
+    pub timestamp: MilliSecondsSinceUnixEpoch,
+
+    /// The content of the local event.
+    pub content: Raw<AnyMessageLikeEventContent>,
+
+    /// The event type of the local event.
+    pub event_type: String,
```
bnjbvr: Instead of this, could you use the send queue's `SerializableEventContent`, instead? It seems you're reimplementing the same functionality here to serialize/deserialize it :eyes:

crates/matrix-sdk/src/latest_events/latest_event.rs
```diff
@@ -341,14 +344,38 @@ pub enum LatestEventValue {
...
+    LocalCannotBeSent(LocalLatestEventValue),
+}
+
+/// Represents the value for [`LatestEventValue::Remote`].
+pub type RemoteLatestEventValue = TimelineEvent;
+
+/// Represents the value for [`LatestEventValue::LocalIsSending`] and
+/// [`LatestEventValue::LocalCannotBeSent`].
+#[derive(Debug, Clone)]
+pub struct LocalLatestEventValue {
```
bnjbvr: I would be in favor of dropping the `Value` suffixes in both names, as they don't add any… value :drum:

crates/matrix-sdk/src/latest_events/latest_event.rs
```diff
@@ -860,133 +887,117 @@ fn extract_content_from_any_message_like(
...
+    use ruma::{event_id, events::room::message::RoomMessageEventContent, user_id};
 
-    use super::{find_and_map_timeline_event, LatestEventContent, RoomMessageEventContent};
+    use super::filter_timeline_event;
 
     macro_rules! assert_latest_event_content {
-        ( with | $event_factory:ident | $event_builder:block
-          it produces $match:pat ) => {
+        ( with | $event_factory:ident | $event_builder:block,
+          $bool:literal $(,)* ) => {
```
bnjbvr: Maybe rename the macro? Invoke sites don't quite make sense, with the `true` or `false` at the end:

```
assert_latest_event_content!(
    with | event_factory | { event_factory.text_msg("hello").into_event() },
    true,
);
```

crates/matrix-sdk/src/latest_events/latest_event.rs
Hywan: You're right! Good idea.

crates/matrix-sdk/src/latest_events/latest_event.rs
Hywan: Hmm yeah, you're right. I'm revisiting this.

crates/matrix-sdk/src/latest_events/latest_event.rs
Hywan: I agree it feels a bit heavy. However, going to `RemoteLatestEvent` and `LocalLatestEvent` feels confusing to me. The type hierarchy looks like this:

* `LatestEvent`
  * `LatestEventValue`
    * `LocalLatestEventValue`
    * `RemoteLatestEventValue`

If we drop the `Value` suffix, we would get something that might read like so:

* `LatestEvent`
  * `LatestEventValue`
* `LocalLatestEvent`
* `RemoteLatestEvent`

And I believe it's confusing. One might expect `LocalLatestEvent` to get a _value_ inner type for example.

I don't find another approach with shorter names and still satisfying for the moment.

crates/matrix-sdk/src/latest_events/latest_event.rs
```diff
@@ -860,133 +859,126 @@ fn extract_content_from_any_message_like(
...
-    use super::{find_and_map_timeline_event, LatestEventContent, RoomMessageEventContent};
+    use super::filter_timeline_event;
 
     macro_rules! assert_latest_event_content {
-        ( with | $event_factory:ident | $event_builder:block
-          it produces $match:pat ) => {
+        ( event | $event_factory:ident | $event_builder:block
+          is a candidate ) => {
+            assert_latest_event_content!(@_ | $event_factory | $event_builder, true);
+        };
```
bnjbvr, line 891: Cute!

bnjbvr (approved):
lgtm!

PR #5563, opened 2025-08-21
refactor(bench): rewrite the `event_cache` benchmark to measure what it's supposed to

bnjbvr:
- Fixes a nasty cycle involving `Arc<EventCacheInner>`, which would cause a leak of it, after a client/`EventCache` has been `Drop`ped. (This caused "too many files" / "connection refused" errors in the updated benchmark.)
- Rewrite the event cache benchmark so that it doesn't reuse the same filled database every single time. Before the rewrite, the benchmark would fill the database in the first iteration, then try to add the same events in the same database: they would all be duplicated, and as such, no insertions would happen! Instead, the updated benchmark in https://github.com/matrix-org/matrix-rust-sdk/commit/767997ad82efe794785cc6d407a97f5c300a195a removes the event cache directory, and creates a fresh new `Client` instance to use for measuring.
- The above still requires an async setup, which Criterion doesn't natively support, and which means we can't run it in CI. The third commit changes the approach so that we're using the same `Client` instance, and clear the event cache store across runs of the benchmark. As a result, we can run it in CI. See the commit message for details about the tradeoffs involved there; I think they're worth it, overall.


crates/matrix-sdk/src/event_cache/mod.rs
```diff
@@ -352,12 +352,16 @@ impl EventCache {
...
-        inner: Arc<EventCacheInner>,
+        inner: Weak<EventCacheInner>,
         mut rx: mpsc::Receiver<AutoShrinkChannelPayload>,
     ) {
         while let Some(room_id) = rx.recv().await {
             trace!(for_room = %room_id, "received notification to shrink");
 
+            let Some(inner) = inner.upgrade() else {
+                return;
+            };
```
Hywan, line 363: Do you believe we can test the previous buggy behaviour?

bnjbvr: We could write some tests that counts the number of strong references to an `Arc<EventCacheInner>`, which could act as some kind of regression test.

Hywan: I think it would be quite important to get this honestly. Thoughts?

bnjbvr: I don't think so, but I won't be opposed to you writing a test :)

Issue #5038, opened 2025-05-14
Threads: Implement the new participation model API


manuroe:
SDK companion issue of [[Prep] Threads: Design a participation model API in Matrix](https://github.com/element-hq/element-meta/issues/2777) for EX.

---

- [x] Add MSC4306 endpoints in Ruma
  - https://github.com/ruma/ruma/pull/2169 
- [x] Add MSC4306  endpoint requests in the SDK, and expose them at the FFI layer
  - https://github.com/matrix-org/matrix-rust-sdk/pull/5439
- [x] Store subscription state in persistent storage
  - https://github.com/matrix-org/matrix-rust-sdk/pull/5455
- [x] Add support for the new push rules
  - https://github.com/matrix-org/matrix-rust-sdk/pull/5483 
- [x] Automatically subscribe under the thread participation model
  - if the client's user sent an event in the thread (including the thread's root), send a manual subscription
  - if the client's been mentioned in the thread, send an automatic subscription
- [x] Add the new `postcontent` section to the push rules, and update the predefined thread subs rules from MSC4306
  - https://github.com/ruma/ruma/pull/2193
  - [x] update SDK after the Ruma PR has been merged
- [x] Receive new thread subscriptions via sliding sync (MSC4308)
  - [x] Rejigger stored subscription data format, so as to store the bumpstamp in there
    - https://github.com/matrix-org/matrix-rust-sdk/pull/5589
  - [x] Store the (pos, prev_batch) pairs into the DB
  - [x] Automatically catch up all subscriptions in the background (MSC4308)
  - [x] Make `Room::load_or_fetch_thread_subscription()` well, actually load the subscriptions from the store.
- [x] Add support for the new endpoint for catching up subscriptions (MSC4308)
  - https://github.com/ruma/ruma/pull/2194 
  - https://github.com/ruma/ruma/pull/2200
  - https://github.com/matrix-org/matrix-rust-sdk/pull/5590

bnjbvr:
Some small status update. I'll need to rejigger the data format used to save the subscriptions in the DB; after all, we need to save the bumpstamp and the status (unsubscribed/subscribed), so as to be able to order subscriptions against unsubscriptions (and conversely).

[...]

PR #5564, opened 2025-08-21
tests: add support for global account data in the `EventFactory`

bnjbvr:
Let's make it simpler to test the global account data, notably push rules; this will be handy for testing the behavior of the new thread push rules.

Part of #3716, useful for #5038.

### DIFF
--- a/crates/matrix-sdk/src/notification_settings/mod.rs
+++ b/crates/matrix-sdk/src/notification_settings/mod.rs
@@ -647,7 +647,7 @@ mod tests {
             .ok_and_run(&client, |sync_response_builder| {
                 let f = EventFactory::new();
                 sync_response_builder.add_global_account_data(
-                    f.push_rules(Ruleset::server_default(client.user_id().unwrap())).into_raw(),
+                    f.push_rules(Ruleset::server_default(client.user_id().unwrap())),
                 );
             })
             .await;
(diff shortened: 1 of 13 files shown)


crates/matrix-sdk/src/notification_settings/mod.rs
```diff
@@ -91,36 +91,40 @@ pub struct NotificationSettings {
...
     pub(crate) fn new(client: Client, ruleset: Ruleset) -> Self {
         let changes_sender = broadcast::Sender::new(100);
         let rules = Arc::new(RwLock::new(Rules::new(ruleset)));
 
-        // Listen for PushRulesEvent
+        // Listen for PushRulesEvent.
         let push_rules_event_handler_handle = client.add_event_handler({
             let changes_sender = changes_sender.clone();
-            let rules = Arc::clone(&rules);
+            let rules = rules.clone();
```
andybalaam, line 113: I was under the impression that `Arc::clone` is more idiomatic. I do find that style here: https://doc.rust-lang.org/stable/book/ch16-03-shared-state.html and here: https://doc.rust-lang.org/std/sync/struct.Arc.html#method.clone . Does the house style do it differently?

andybalaam (approved):
Great, thanks!

crates/matrix-sdk/src/notification_settings/mod.rs
bnjbvr: Yep, we definitely do `thing.clone()` immediately in general. I guess the one thing it would allow you to distinguish with, is when you're cloning the wrappee, but that would look like `(*rules).clone()` anyways, and my intuition is that the explicit deref there would catch the reader's attention that this is not cloning the `Arc` wrapper.

PR #5567, opened 2025-08-21
fix(search): Make adding events to index idempotent.

dragonfly1033:
Adding events to the index is now idempotent.

### DIFF
--- a/crates/matrix-sdk-search/src/index.rs
+++ b/crates/matrix-sdk-search/src/index.rs
@@ -137,7 +137,7 @@ impl RoomIndex {
                     self.writer.add_document(document)?;
                 }
             }
-        };
+        }
         Ok(())
     }
 
@@ -192,12 +192,11 @@ impl RoomIndex {
     }
 
     fn contains(&self, event_id: &EventId) -> bool {
-        let search_result = self.search(format!("event_id:\"{}\"", event_id).as_str(), 5);
+        let search_result = self.search(format!("event_id:\"{event_id}\"").as_str(), 1);
         match search_result {
-            // If there are > 1 entry (which there shouldn't be), still return true
-            Ok(results) => results.len() != 0,
+            Ok(results) => !results.is_empty(),
             Err(err) => {
-                warn!("Failed to check if event has been indexed, assuming it has: {err:?}");
+                warn!("Failed to check if event has been indexed, assuming it has: {err}");
                 true
             }
         }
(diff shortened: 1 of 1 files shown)


crates/matrix-sdk-search/src/index.rs
```diff
@@ -184,6 +190,18 @@ impl RoomIndex {
 
         Ok(ret)
     }
+
+    fn contains(&self, event_id: &EventId) -> bool {
+        let search_result = self.search(format!("event_id:\"{event_id}\"").as_str(), 5);
```
bnjbvr: Maybe we can limit the search to 1 result?

crates/matrix-sdk-search/src/index.rs
```diff
@@ -184,6 +190,18 @@ impl RoomIndex {
...
+        match search_result {
+            // If there are > 1 entry (which there shouldn't be), still return true
+            Ok(results) => !results.is_empty(),
+            Err(err) => {
+                warn!("Failed to check if event has been indexed, assuming it has: {err:?}");
```
bnjbvr: nit: don't use `{:?}` (i.e. calling the `Debug` impl) for errors, but `{}` (i.e. calling the `Display` impl). The reason is that `Error` implements `Display`, and this is the way those errors are supposed to be rendered.

crates/matrix-sdk-search/src/index.rs
```diff
@@ -184,6 +190,18 @@ impl RoomIndex {
 
         Ok(ret)
     }
+
+    fn contains(&self, event_id: &EventId) -> bool {
+        let search_result = self.search(format!("event_id:\"{event_id}\"").as_str(), 5);
+        match search_result {
+            // If there are > 1 entry (which there shouldn't be), still return true
```
bnjbvr: I don't understand this comment: why "still" return true? (also, missing dot at the end)

bnjbvr (commented):
Slightly worried that we have to run a specific search for what's conceptually a set lookup. Could you write a tiny benchmark and measure how long it takes to add the same document, say, 1000 / 10000 / 100000 times? (not necessarily one in the `benches` dir and that we'd check in, but you'd get bonus points for that!)

crates/matrix-sdk-search/src/index.rs
dragonfly1033: Yep we can.

crates/matrix-sdk-search/src/index.rs
dragonfly1033: Sorry, keep forgetting this

crates/matrix-sdk-search/src/index.rs
dragonfly1033: The comment was redundant. forgot to remove

bnjbvr (changes requested):
Please look at my previous review comment where I asked for running a tiny benchmark.

dragonfly1033:
Sorry, got a bit eager. Here's the results, looks like nothing to worry about.
image

For reference: this is benching the `RoomIndex::handle_event` function on the same event.

PR #5568, opened 2025-08-22
Split media store from event cache store.

dragonfly1033:
This PR is a start to the process of splitting the media store from the event cache store. #5410 

It contains:
* Split `MediaStore` trait from `EventCacheStore`. 
* Rename `EventCacheStoreMedia` to `MediaStoreInner`. 
* Move relevant tests into `MediaStoreIntegrationTests`.

This will be done over 3 PR's (reviewing 1, 2, 3 then merging 3 into 2 into 1). 

A reminder comment for my own sanity:
This PR will not pass tests until after merging.

Current state of this PR:
- [x] Step 1 reviewed #5568
- [x] Step 2 reviewed #5569 
- [x] Step 3 reviewed #5571 
- [x] Step 3 merged into Step 2
- [x] Step 2 merged into Step 1
- [ ] Add changes to changelog.
- [ ] Ready to merge :tada: 

Note, may also want to: 
* Re-organize file structure
* Split/refactor benchmarks namely `benchmarks/benches/event_cache.rs`

- [ ] Public API changes documented in changelogs (optional)

Signed-off-by: Shrey Patel shreyp@element.io

PR #5570, opened 2025-08-22
Part 3 of Split media store from event cache store 

dragonfly1033:
Read #5568 . This is part 3.

Contains:
* Instantiate `MediaStore`s and use where required.


dragonfly1033:
Closed because it should be PR'd off Part 2 not main

PR #5571, opened 2025-08-22
Part 3 of Split media store from event cache store

dragonfly1033:
Read #5568 . This is part 3.

Contains:
* Instantiate `MediaStore`s and use where required.


crates/matrix-sdk-base/src/event_cache/store/media/mod.rs
Hywan, line 1: Maybe we want to move that into `crates/matrix-sdk-base/src/media/store.rs` or something? The idea is to move out of the `event_cache` directory.

crates/matrix-sdk-base/src/event_cache/store/media/traits.rs
Hywan, line 1: Same, we should move it `crates/matrix-sdk-base/src/media/traits.rs` maybe?

crates/matrix-sdk-base/src/store/mod.rs
```diff
@@ -730,6 +730,7 @@ pub struct StoreConfig {
     pub(crate) crypto_store: Arc<DynCryptoStore>,
     pub(crate) state_store: Arc<DynStateStore>,
     pub(crate) event_cache_store: event_cache_store::EventCacheStoreLock,
+    pub(crate) media_store: event_cache_store::media::MediaStoreLock,
```
Hywan: Ideally, we want to get `media_store::MediaStoreLock`.

crates/matrix-sdk-indexeddb/src/media_store/builder.rs
```diff
@@ -31,6 +24,7 @@ impl IndexeddbMediaStoreBuilder {
     /// Opens the IndexedDB database with the provided name. If successfully
     /// opened, builds the [`IndexeddbMediaStore`] with that database
     /// and the provided store cipher.
+    #[allow(dead_code)]
```
Hywan: Why?

crates/matrix-sdk-indexeddb/src/media_store/mod.rs
```diff
@@ -45,7 +45,8 @@ pub struct IndexeddbMediaStore {
 
 impl IndexeddbMediaStore {
     /// Provides a type with which to conveniently build an
-    /// [`IndexeddbEventCacheStore`]
+    /// [`IndexeddbMediaStore`]
+    #[allow(dead_code)]
```
Hywan: Why?

crates/matrix-sdk-indexeddb/src/media_store/mod.rs
```diff
@@ -222,17 +223,12 @@ mod tests {
...
-            Self::Backend(Box::new(value))
-        }
-    }
-
     mod unencrypted {
         use super::*;
 
         wasm_bindgen_test::wasm_bindgen_test_configure!(run_in_browser);
 
+        #[allow(clippy::unused_async)]
```
Hywan: Why :-)? Do we have another part that fixes that?

dragonfly1033: The test macro requires an async `get_media_store` function. For other, namely sqlite, implementations this is required and satisfied but for indexeddb this function doesn't contain an await but still needs to work with the testing macro. AFAIK there's no way to have a "maybe async" function so I just did this. It is only used during testing anyway and there are other instances of its use with a comment providing a reason. As such I have also added a comment to provide a reason :slightly_smiling_face:

Hywan: Thanks!

PR #5573, opened 2025-08-25
fix(media): Use the right value to get no timeouts when downloading media

jmartinesp:
Using `Option::None` will just make `reqwest` use the default `Client`'s timeout value, instead of setting an infinite timeout value. Using `Some(Duration::MAX)` should achieve the desired behaviour instead.

Fixes https://github.com/element-hq/element-x-ios/issues/4437.

Signed-off-by:


jmartinesp:
I tried adding some regression tests for this, but I couldn't find a way to have wiremock use the `tokio` time for mocking the delay in any way and I don't think adding ~30s to our tests just for this is worth it.

PR #5574, opened 2025-08-25
IndexedDB: Add implementation of `MediaRetentionPolicy`-related functions in `EventCacheStoreMedia`

mgoldenberg:
## Background

This pull request is part of a series of pull requests to add a full IndexedDB implementation of the `EventCacheStore` (see #4617, #4996, #5090, #5138, #5226, #5274, #5343, #5384, #5406, #5414, #5497, #5506, #5540). This particular pull request adds a `MemoryStore`-backed implementation of `EventCacheStoreMedia`, as well as, IndexedDB-backed implementations for `MediaRetentionPolicy`-related functions.

## Changes

The overarching change adds a `MemoryStore`-backed implementation of `EventCacheStoreMedia` for `IndexeddbEventCacheStore`. Furthermore, the implementation of `EventCacheStore` now delegates to the implementation of `EventCacheStoreMedia` via a `MediaService` nested in `IndexeddbEventCacheStore`.

Additionally, there are now implementations and tests for the following `EventCacheStoreMedia` functions.

- `EventCacheStoreMedia::media_retention_policy_inner`
    - Supported by `IndexeddbEventCacheStoreTransaction::get_media_retention_policy`
- `EventCacheStoreMedia::set_media_retention_policy_inner`

And finally, to support these queries, the `CORE` object store now has a primary key with key path `id`.

## Caveats

`EventCacheStoreMedia` requires that the implementing type be `Clone`able. Consequently, some rather minimal changes were required in order to support `Clone`. 

There is, however, one notable change for supporting `Clone`, which is that `IndexeddbEventCacheStore.inner` went from being an `IdbDatabase` to an `Rc<IdbDatabase>`.

While this may seem like a problem in a concurrent environment, it is ultimately not an issue, as IndexedDB ensures that only a single transaction is running at a time. Any interfering transaction will be canceled by IndexedDB and return an error.

The downside of this architecture, however, is that it would require re-trying in a loop if multiple instances of `IndexeddbEventCacheStore` were contending for access. 

An alternative would be to use a `tokio::sync::Mutex` in conjunction with an `Arc`, which would replace the loop with an `await`. The consequences of this are that the implementation of `IndexeddbEventCacheStoreTransaction` becomes rather tricky, as it seems to require a self-referential struct.

I'm not entirely sure why the `EventCacheStoreMedia` requires `Clone` and how it will be used, so I opted for using an `Rc` as it was far more straightforward and still a sound approach. 

## Future Work

- Add implementation of remaining `EventCacheStoreMedia` functions

---

### DIFF
--- a/crates/matrix-sdk-indexeddb/src/event_cache_store/builder.rs
+++ b/crates/matrix-sdk-indexeddb/src/event_cache_store/builder.rs
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License
 
-use std::sync::Arc;
+use std::{rc::Rc, sync::Arc};
 
 use matrix_sdk_base::event_cache::store::MemoryStore;
 use matrix_sdk_store_encryption::StoreCipher;
@@ -66,7 +66,7 @@ impl IndexeddbEventCacheStoreBuilder {
     /// and the provided store cipher.
     pub async fn build(self) -> Result<IndexeddbEventCacheStore, IndexeddbEventCacheStoreError> {
         Ok(IndexeddbEventCacheStore {
-            inner: open_and_upgrade_db(&self.database_name).await?,
+            inner: Rc::new(open_and_upgrade_db(&self.database_name).await?),
             serializer: IndexeddbEventCacheStoreSerializer::new(IndexeddbSerializer::new(
                 self.store_cipher,
             )),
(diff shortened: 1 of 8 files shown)


Hywan:
The `MediaService` trait implements `Clone` because of this:

https://github.com/matrix-org/matrix-rust-sdk/blob/f4ce4356ab674ad33177fa5548813d9638dbe4cb/crates/matrix-sdk-base/src/event_cache/store/media/media_service.rs#L319-L328

The cleanup phase is done in a separate task. The task handle is stored here:

https://github.com/matrix-org/matrix-rust-sdk/blob/f4ce4356ab674ad33177fa5548813d9638dbe4cb/crates/matrix-sdk-base/src/event_cache/store/media/media_service.rs#L53-L57

If it is really annoying for you, we may want to revisit this.

Issue #5572, opened 2025-08-25
`matrix_sdk_sqlite::EventCacheStore::find_event_relations` is too slow


Hywan:
I've investigated [this rageshake](https://github.com/element-hq/element-x-ios-rageshakes/issues/4418). I noticed that `find_event_relations` is too slow, 3-4s in average, up to 14mn (yes!):

```sh
$ rg 'finished in.*> find_event_relations' | rg '\d+(\.\d+)?s' --only-matching | sort --numeric-sort --reverse
843.875936208s
9.140288875s
7.676449416s
6.553326875s
6.112908041s
6.1045985s
6.103136458s
6.100351625s
6.098929833s
6.098878s
6.045843709s
6.028541875s
5.919690042s
5.71187025s
5.61592125s
4.657586375s
4.449823167s
4.364553583s
4.241941958s
4.166927458s
4.155965792s
3.900056458s
3.723121125s
3.343189208s
3.207252542s
3.183746875s
3.160940958s
3.16089425s
3.160786166s
3.160109292s
3.160023625s
3.152290542s
3.13718875s
3.081865292s
2.676727583s
```

The entry code is here:

https://github.com/matrix-org/matrix-rust-sdk/blob/4e2655aa1bd56469930f3ce11a57f08ac6f716aa/crates/matrix-sdk-sqlite/src/event_cache_store.rs#L1250-L1281

The real code is here:

https://github.com/matrix-org/matrix-rust-sdk/blob/4e2655aa1bd56469930f3ce11a57f08ac6f716aa/crates/matrix-sdk-sqlite/src/event_cache_store.rs#L1685-L1768

bnjbvr:
Ouch, good find! Looks like these queries could make use of indexes…

bnjbvr:
I'll try to address this.

bnjbvr:
I've implemented a benchmark and an optimization to add the indexes, but I can't reproduce the initial slowness (unless we're talking about more than 1,000 related events; e.g. a thread with more than 1000 responses). So there has to be another explanation (maybe lock contention? maybe something else).

bnjbvr:
Ok, after playing with the benchmark, and adding more unrelated events (either events which have no relations, or events which are related to another target event), I can reproduce the slowness, and it seems indeed to scale with the number of events.

Good news is, the perf fix in #5579 makes it so that the query's time becomes independent of the number of events in DB (caps it at 30ms on my machine, independently of the number of in-DB events).

[...]

PR #5578, opened 2025-08-26
bench: add a benchmark for finding related events

bnjbvr:
Does what it says on the tin. Split from the performance fix, so we can get some initial numbers on the CI bench runs too. 

Part of investigating https://github.com/matrix-org/matrix-rust-sdk/issues/5572

### DIFF
--- a/benchmarks/benches/event_cache.rs
+++ b/benchmarks/benches/event_cache.rs
@@ -152,6 +152,14 @@ fn handle_room_updates(c: &mut Criterion) {
 }
 
 fn find_event_relations(c: &mut Criterion) {
+    // Number of other events to saturate the DB, but that will not be affected by
+    // the benchmark. A small multiple of this number will be added.
+    // When running locally, run with more events than in Codespeed CI.
+    #[cfg(feature = "codspeed")]
+    const NUM_OTHER_EVENTS: usize = 100;
+    #[cfg(not(feature = "codspeed"))]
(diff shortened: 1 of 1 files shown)


benchmarks/benches/event_cache.rs
```diff
@@ -147,10 +151,192 @@ fn handle_room_updates(c: &mut Criterion) {
...
+                client.event_cache().subscribe().unwrap();
+
+                // Sync the updates before starting the benchmark.
+                let mut update_recv = client.event_cache().subscribe_to_room_generic_updates();
+
+                client.event_cache().handle_room_updates(room_updates.clone()).await.unwrap();
+
+                // Wait for the event cache to notify us of the update.
+                let update = update_recv.recv().await.unwrap();
+                assert_eq!(update.room_id, room_id);
```
poljar: This seems to panic for me:

```
thread 'main' panicked at benchmarks/benches/event_cache.rs:280:17:
assertion `left == right` failed
  left: "!other-room:ben.ch"
 right: "!room:ben.ch"
```

Seems that there's some non-determinism here?

poljar (changes requested):
Sadly panics for me :smiling_face_with_tear:

poljar:
Oh, it panics on the CI as well.

benchmarks/benches/event_cache.rs
bnjbvr: uh, last minute changes to the PR made it inconsistent, will fix.

PR #5581, opened 2025-08-26
feat(search): Add redactions and edits to search index.

dragonfly1033:
This PR adds support for edit and redaction events to affect the search index.
Redactions will remove events from the index.
Edits will remove the old event and add the new event. 

Added some tests for this behaviour as well.

Problems to be solved:
All edits reference the original which is replaced. Currently this means that edits 2+ will simply be added after deleting the original (which is a no-op because it was already deleted by edit 1) rather than deleting the previous edits.

The solution is to simply construct a new faux-event with the original event id and the new content and forward that to the index. 


crates/matrix-sdk-search/src/schema.rs
```diff
@@ -133,6 +135,27 @@ impl MatrixSearchIndexSchema for RoomMessageSchema {
...
+            AnySyncMessageLikeEvent::RoomRedaction(redaction_event) => {
+                if let SyncRoomRedactionEvent::Original(redaction_event) = redaction_event {
+                    if let Some(redacted_event_id) = redaction_event.content.redacts {
+                        // From room version 11, redacted event id is here. Prefer this if it
+                        // exists since it's newer.
+                        Ok(RoomIndexOperation::Remove(redacted_event_id))
+                    } else if let Some(redacted_event_id) = redaction_event.redacts {
+                        // Before room version 11, redacted event id is here.
+                        // Fallback to this.
+                        Ok(RoomIndexOperation::Remove(redacted_event_id))
```
bnjbvr: You shouldn't have to do this; instead, you want to use `redaction_event.redacts()`, which needs the room (redaction) rules. You can take some inspiration by looking at other call sites of `.redacts()`. This might require adding the rules as a parameter to this function, or as state in the owning object.

crates/matrix-sdk-search/src/schema.rs
```diff
@@ -133,6 +135,27 @@ impl MatrixSearchIndexSchema for RoomMessageSchema {
...
+                    } else {
+                        // If not acting on anything, we can just ignore it.
+                        trace!("Room redaction in indexing redacts nothing, ignoring.");
+                        Ok(RoomIndexOperation::NoOp)
+                    }
+                } else {
+                    // If redaction itself is redacted, we can ignore it.
+                    Ok(RoomIndexOperation::NoOp)
```
bnjbvr: Don't think this is valid, because a redacted redaction still contains the event id of the redacted event. By using `SyncRoomRedactionEvent::redacts()` (instead of the `if let ::Original = ` branch), you can handle all the cases for free.

crates/matrix-sdk-search/src/index.rs
```diff
@@ -34,6 +34,8 @@ use crate::{
 /// A struct to represent the operations on a [`RoomIndex`]
 pub(crate) enum RoomIndexOperation {
     Add(TantivyDocument),
+    Remove(OwnedEventId),
+    NoOp,
```
bnjbvr: light suggestion: I've got a small preference for the `Noop` capitalization, because `NoOp` looks a bit weird to me.

crates/matrix-sdk-search/src/index.rs
```diff
@@ -378,4 +378,42 @@ mod tests {
 
         Ok(())
     }
+
+    #[test]
+    fn test_redaction_removes_event() -> Result<(), Box<dyn Error>> {
```
bnjbvr, line 397: nit: either use only `.expect()` / `.unwrap()` in the test (and never use `?`, so this function wouldn't return a result), or use only `?` with the `TestResult` type being returned by the test function here.

crates/matrix-sdk-search/src/index.rs
```diff
@@ -378,4 +378,42 @@ mod tests {
...
+            .into_any_sync_message_like_event();
+
+        index.handle_event(event.clone())?;
+
+        index.commit_and_reload()?;
+
+        assert!(index.contains(event_id), "Index should contain event");
+
+        let redaction_event_id = event_id!("$redaction_event_id:localhost");
+        let redaction = EventFactory::new()
```
bnjbvr: nit: reuse the `EventFactory` by storing it into a local variable :P

You can even use "currying" of the sender/room parameters, for the event factory, by preconfiguring it before creating an event:

```
let f = EventFactory::new().room(room_id).sender(user_id);
```

Then, all events created with `f` will automatically have a sender and room already set.

crates/matrix-sdk-search/src/schema.rs
```diff
@@ -28,6 +34,27 @@ use crate::{
...
+    fn relates_to(&self) -> Option<Relation<Self::T>>;
+}
+
+impl AnySyncMessageContent for MessageEventContent {
+    type T = MessageEventContentWithoutRelation;
+
+    fn relates_to(&self) -> Option<Relation<Self::T>> {
+        self.relates_to.clone()
+    }
+}
```
bnjbvr: Honestly, I would remove all support for `m.message` because as far as I know, it's unused, and we're not planning to add support for it (it would require lots of extra support in many places, for it to work correctly anyways). Can you remove support for it from the search index, at the moment, please? (This would also simplify this commit, likely)

crates/matrix-sdk-search/src/index.rs
```diff
@@ -35,6 +35,7 @@ use crate::{
 pub(crate) enum RoomIndexOperation {
     Add(TantivyDocument),
     Remove(OwnedEventId),
+    Edit(OwnedEventId, TantivyDocument),
```
bnjbvr, line 47: Let's add a doc comment explaining what the event id is, here.

crates/matrix-sdk/src/event_cache/search.rs
```diff
@@ -0,0 +1,96 @@
...
+use matrix_sdk_base::deserialized_responses::TimelineEvent;
+use ruma::{
+    events::{room::message::Relation, AnySyncMessageLikeEvent},
+    EventId,
+};
+use tracing::warn;
+
+use crate::event_cache::RoomEventCache;
+
+async fn get_most_recent_edit(
```
bnjbvr, line 26: For each of these functions, let's add a small doc comment explaining what it does.

crates/matrix-sdk/src/event_cache/search.rs
```diff
@@ -0,0 +1,96 @@
...
+    related
+        .last()
+        .and_then(|latest| match latest.raw().deserialize() {
+            Ok(AnySyncTimelineEvent::MessageLike(latest)) => Some(latest),
+            _ => None,
+        })
+        .or_else(|| match original_ev.raw().deserialize() {
+            Ok(AnySyncTimelineEvent::MessageLike(original_ev)) => Some(original_ev),
+            _ => None,
+        })
```
bnjbvr: Maybe there's a way to use `.map_or_else()` here, and keep the deserialization out of the `.map_or_else()`?

crates/matrix-sdk/src/event_cache/search.rs
```diff
@@ -0,0 +1,96 @@
...
+}
+
+async fn handle_possible_edits(
```
bnjbvr, line 47: doc comment here too

crates/matrix-sdk/src/event_cache/search.rs
```diff
@@ -0,0 +1,96 @@
...
+                _ => event.event_id(),
+            })?;
+
+            get_most_recent_edit(cache, original_ev_id).await
+        }
+        _ => None,
+    }
+}
+
+pub(crate) async fn parse_timeline_event(
```
bnjbvr, line 66: doc comment here too

crates/matrix-sdk-search/src/index.rs
```diff
@@ -150,7 +150,7 @@ impl RoomIndex {
                 }
             }
             RoomIndexOperation::NoOp => {}
-        };
+        }
```
bnjbvr, line 167: Fwiw, we should not have commits like 50ec802948c534d4ddf4de08bab03ee9989e49e5 "fix clippy issues in tests". You should integrate these changes in the commits where they should've happened. You can use `git rebase -i` to modify these commits, or, add the changes and use [`git-absorb`](https://github.com/tummychow/git-absorb) to incorporate the changes in the original commits.

crates/matrix-sdk-search/src/schema.rs
```diff
@@ -70,6 +71,7 @@ pub(crate) trait MatrixSearchIndexSchema {
 pub(crate) struct RoomMessageSchema {
     inner: Schema,
     event_id_field: Field,
+    original_event_id_field: Field,
```
bnjbvr, line 53: Consider adding a doc comment on top of this and `event_id_field` to explain their differences.

crates/matrix-sdk/src/event_cache/mod.rs
```diff
@@ -725,10 +725,19 @@ impl EventCache {
                         return;
                     };
 
+                    let Ok((room_cache, _drop_handles)) =
+                        client.event_cache().for_room(&room_id).await
+                    else {
+                        warn!(for_room = %room_id, "Failed to get RoomEventCache");
+                        continue;
+                    };
```
bnjbvr: In this case, please log the error when the `for_room()` call fails.

crates/matrix-sdk-search/src/index.rs
```diff
@@ -372,4 +387,85 @@ mod tests {
...
+            .into_any_sync_message_like_event();
+
+        index.handle_event(old_event)?;
+
+        index.commit_and_reload()?;
+
+        assert!(index.contains(old_event_id), "Index should contain event");
+
+        let new_event_id = event_id!("$new_event_id:localhost");
+        let edit = EventFactory::new()
```
bnjbvr: ditto, let's reuse the `EventFactory`, and use fixed `room`/`sender`s on it, to make the test less verbose.

crates/matrix-sdk/tests/integration/room/joined.rs
```diff
@@ -770,6 +770,95 @@ async fn test_sync_message_is_indexed() {
     assert_eq!(response[0], event_id, "event id doesn't match: {response:?}");
 }
 
+#[cfg(feature = "experimental-search")]
+#[async_test]
+async fn test_search_index_edit_ordering() {
```
bnjbvr: Can you put this test in the `event_cache` file, instead?

crates/matrix-sdk/tests/integration/room/joined.rs
```diff
@@ -770,6 +770,95 @@ async fn test_sync_message_is_indexed() {
...
+
+    assert_eq!(results.len(), 0, "Search should return 0 results, got {results:?}");
+
+    // Adding the original after some pending edits should add the latest edit
+    // instead of the original.
+    server
+        .mock_sync()
+        .ok_and_run(&client, |builder| {
+            builder.add_joined_room(JoinedRoomBuilder::new(room_id).add_timeline_event(original));
+        })
```
bnjbvr: nit: here and below, you can use `.sync_room(JoinedRoomBuilder::new(room_id).add_timeline_event(original)))` too.

crates/matrix-sdk/tests/integration/room/joined.rs
dragonfly1033: The test above it is also for search. Should I move that too?

crates/matrix-sdk/src/event_cache/search.rs
dragonfly1033: I came up with this?
```rust
match related.last().unwrap_or(&original_ev).raw().deserialize() {
    Ok(AnySyncTimelineEvent::MessageLike(latest)) => Some(latest),
    _ => None,
}
```

bnjbvr: Very nice!

dragonfly1033:
> Great start!
> 
> Sorry to throw a wrench at you, but there's a bit more to take care of. An edit can be redacted too (EW allows that!), in which case, it's necessary to reindex the new latest edit (or the original unedited event, in case the redacted edit was the last edit).
> 
> Do you want to add that here, or in a follow-up PR?

I'll fix this in the next PR :slightly_smiling_face: .

crates/matrix-sdk/tests/integration/room/joined.rs
bnjbvr: Feel free, yeah!

crates/matrix-sdk-search/src/index.rs
```diff
@@ -372,4 +392,75 @@ mod tests {
...
+            )
+            .event_id(new_event_id)
+            .into_any_sync_message_like_event();
+
+        index.handle_event(edit, &RedactionRules::V11)?;
+
+        index.commit_and_reload()?;
+
+        assert!(!index.contains(old_event_id), "Index should not contain old event");
+        assert!(index.contains(new_event_id), "Index should contain edited event");
```
bnjbvr, line 462: Not sure this would be the desired behavior. Ideally, when searching for the updated string, we'd get the edited event id as well as the original event id, so we can create a timeline focused on it. But let's see later, we can tweak that behavior easily.

crates/matrix-sdk/src/client/search.rs
```diff
@@ -135,3 +139,127 @@ impl SearchIndexGuard<'_> {
...
+                JoinedRoomBuilder::new(room_id)
+                    .add_timeline_event(dummy.clone())
+                    .add_timeline_event(edit1.clone())
+                    .add_timeline_event(edit2.clone()),
+            )
+            .await;
+
+        let results = room.search("message", 3).await.unwrap();
+
+        assert_eq!(results.len(), 0, "Search should return 0 results, got {results:?}");
```
bnjbvr, line 243: Is it expected that it doesn't return the edited event's contents, even though they're indexed?

bnjbvr (approved):
Thanks, let's take it like this and move forward :+1:

crates/matrix-sdk/src/client/search.rs
dragonfly1033: Yes, indexing edits has the behaviour of getting the original, finding its most recent edit  and indexing that. If the search for the original fails then this process fails. You could argue that in this case you should just index this event but I would argue that an edit of an unknown event should be considered invalid.

crates/matrix-sdk-search/src/index.rs
dragonfly1033: If you have the edit event id you can get the original through the event cache. I'm not quite seeing the usecase but you're right, it's an easy change and we can discuss this later.

PR #5582, opened 2025-08-26
Don't error in `import_secrets` if `maybe_enable_backups` errors

uhoreg:
Fixes https://github.com/element-hq/element-x-android/issues/5099

Since enabling backups is optional, if it fails, it shouldn't cause the whole function to error.

Signed-off-by:


uhoreg:
Discussion about whether this approach is correct at https://github.com/element-hq/element-x-android/issues/5099#issuecomment-3226423654  (But no matter what, I think we want to include the extra logging.)

PR #5583, opened 2025-08-27
feat(crypto): Add EncryptionSettings::encrypt_state_events

kaylendog:
This will be used inside the WASM SDK to introduce a similar field to its `EncryptionSettings` struct.

- [ ] Adds `EncryptionSettings::encrypt_state_events` with Serde default tag.

Signed-off-by: Skye Elliot


poljar (changes requested):
Looks good, though seems that you'll need to shorten the file name even further.

PR #5584, opened 2025-08-27
test: Add a test to check that we don't create event cache reference cycles

poljar:


bnjbvr (approved):
Thank you!

crates/matrix-sdk/src/event_cache/mod.rs
```diff
@@ -1519,4 +1524,45 @@ mod tests {
...
+            // Have the client know the room.
+            let response = SyncResponseBuilder::default()
+                .add_joined_room(JoinedRoomBuilder::new(room_id))
+                .build_sync_response();
+            client.inner.base_client.receive_sync_response(response).await.unwrap();
+
+            client.event_cache().subscribe().unwrap();
+
+            let (_room_event_cache, _drop_handles) =
+                client.get_room(room_id).unwrap().event_cache().await.unwrap();
```
Hywan, line 1553: Do you think it is worth keeping `room_event_cache` alive after the `Client` drops?

crates/matrix-sdk/src/event_cache/mod.rs
```diff
@@ -1519,4 +1524,45 @@ mod tests {
...
+
+        drop(client);
+
+        // Give a bit of time for background tasks to die.
+        sleep(Duration::from_secs(1)).await;
+
+        // No strong counts should exist now that the Client has been dropped.
+        assert_eq!(
+            event_cache_weak.strong_count(),
+            0,
```
Hywan, line 1564: Before https://github.com/matrix-org/matrix-rust-sdk/pull/5563, the strong count is at 1, did you test that?

poljar: Yes.

crates/matrix-sdk/src/event_cache/mod.rs
poljar: Hmm, but we would expect at least some things to stay alive if we do that.

I wanted to test that, if the user drops everything, everything indeed gets dropped.

So perhaps this is for another test, but not sure what exactly you want to test in that case.

Hywan: Alrighty.

Hywan (approved):

PR #5585, opened 2025-08-27
feat(common): Smaller serialised `TimelineEvent`

Hywan:
This set of patches is a draft in order to collect opinions from code owners with concrete examples.

The `TimelineEvent` type represents the majority of the events in the Matrix Rust SDK. It's the type used in the Event Cache. This type is stored in the Event Cache database. The goal of this set of patches is to reduce the storage space. How? By renaming many struct fields or enum variants to shorten names. It's a simple trick, low-hanging fruit, that can have a nice impact.

For example:

```diff
pub struct TimelineEvent {
    /// The event itself, together with any information on decryption.
+   #[serde(rename = "k", alias = "kind")]
    pub kind: TimelineEventKind,
```

With this simple change, we save 4 bytes.

If we apply this to `TimelineEventKind`, down to all the sub-types, **excluding** the crypto-related types, we get this:

| for … | saving for 1 value | saving for 100'000 values |
|-|-|-|
| a decrypted event | 113 bytes | 10.78 MiB |
| a UTD | 64 bytes | 6.1 MiB |
| an uncrypted event | 40 bytes | 3.81 MiB |

If we include the crypto-related types, like `AlgorithmInfo`, `VerificationState`, `UnableToDecryptInfo` etc.:

| for … | saving for 1 value | saving for 100'000 values |
|-|-|-|
| a decrypted event | 207 bytes | 19.74 MiB |
| a UTD | ~92 bytes | 8.77 MiB |

Before going further, I would like to collect your feedback. What do you think? Is it worth the changes?

At the time of writing, this set of patches handle the first case (a decrypted event) without changing the crypto-related types.

An (dead) alternative we (@poljar and I) have explored is to compress the data before going inside the database. Compression is effective when sequence of chars are repeating. Here, per row, the word `kind` exists once (excluding the event's content). Compression won't work that much.

### DIFF
--- a/crates/matrix-sdk-common/src/deserialized_responses.rs
+++ b/crates/matrix-sdk-common/src/deserialized_responses.rs
@@ -324,18 +324,25 @@ pub enum AlgorithmInfo {
 pub struct EncryptionInfo {
     /// The user ID of the event sender, note this is untrusted data unless the
     /// `verification_state` is `Verified` as well.
+    #[serde(rename = "s")]
     pub sender: OwnedUserId,
+
     /// The device ID of the device that sent us the event, note this is
     /// untrusted data unless `verification_state` is `Verified` as well.
+    #[serde(rename = "d")]
     pub sender_device: Option<OwnedDeviceId>,
+
     /// Information about the algorithm that was used to encrypt the event.
+    #[serde(rename = "a")]
     pub algorithm_info: AlgorithmInfo,
+
     /// The verification state of the device that sent us the event, note this
     /// is the state of the device at the time of decryption. It may change in
     /// the future if a device gets verified or deleted.
     ///
     /// Callers that persist this should mark the state as dirty when a device
     /// change is received down the sync.
+    #[serde(rename = "v")]
     pub verification_state: VerificationState,
 }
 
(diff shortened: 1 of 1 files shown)


bnjbvr (commented):
Thanks for trying this out!

I'd be curious to know how much storage we're sparing out of the total? Would be super sweet to get rough distribution indicators around the events sizes, e.g. average / median / total size, for a few users (including one of our in-house <del>power</del> exponential users). That would give us an estimate of the win provided by this PR.

Overall, I'm not sure if we have some embedders downstream who would be annoyed that the serde keys are all obfuscated. Usually, when you want to have something user-readable, you use JSON and you're happy having the full key names. If you want to optimize for on-disk size, you'd use some other serialization format, likely something binary. We're encoding the events' content in JSON in the database, because it makes a few things simpler ("migrations"), but there's nothing that absolutely mandates it, so we could explore using a binary format there.

Then in terms of approach: maybe, instead of retrofitting serde annotations to all the types, we could even craft a new type especially fit to store events in the database, that optimizes for size? Then we could even flatten some of the encrypted information into it.

(Just throwing random ideas / streams of thoughts here, so nothing really prescriptive here.)

richvdh:
What Benji said, mostly. It's definitely an interesting idea to explore, though as I said yesterday, even if you can save tens of megabytes, it's not really worth it if the store as a whole is measured in gigabytes -- so +1 to trying to measure relative numbers.

poljar:
I presume we're not moving forward with this. So I'll close this now.

PR #5589, opened 2025-08-27
feat(threads): store the "bump_stamps" in the thread subscriptions

bnjbvr:
See also MSC4308 for details about the bumpstamps. They allow to order two subscriptions, and avoid bad race conditions where one would get a thread subscription by both paginating and sync'ing them.

It's a bit of a shame that the code comparing bumpstamps looks a bit duplicated, but note that, among the 3 implementations, only 2 are duplicated, for real: the last one uses the `PersistedThreadSubscription` in lieu of `StoreThreadSubscription` type. 

Part of #5038.

### DIFF
--- a/crates/matrix-sdk-base/src/store/memory_store.rs
+++ b/crates/matrix-sdk-base/src/store/memory_store.rs
@@ -968,15 +968,37 @@ impl StateStore for MemoryStore {
         &self,
         room: &RoomId,
         thread_id: &EventId,
-        subscription: StoredThreadSubscription,
+        mut new: StoredThreadSubscription,
     ) -> Result<(), Self::Error> {
-        self.inner
-            .write()
-            .unwrap()
-            .thread_subscriptions
-            .entry(room.to_owned())
-            .or_default()
-            .insert(thread_id.to_owned(), subscription);
+        let mut inner = self.inner.write().unwrap();
+        let room_subs = inner.thread_subscriptions.entry(room.to_owned()).or_default();
+
+        if let Some(previous) = room_subs.get(thread_id) {
+            // Nothing to do.
+            if *previous == new {
+                return Ok(());
+            }
+
+            match (previous.bump_stamp, new.bump_stamp) {
+                // If the previous subscription had a bump stamp, and the new one
+                // doesn't, keep the previous one.
+                (Some(prev_bump), None) => {
+                    new.bump_stamp = Some(prev_bump);
+                }
+
+                // If the previous bump stamp is newer than the new one, don't store the value at
+                // all.
+                (Some(prev_bump), Some(new_bump)) if new_bump <= prev_bump => {
+                    return Ok(());
+                }
+
+                // In all other cases, keep the new bumpstamp.
+                _ => {}
+            }
+        }
+
+        room_subs.insert(thread_id.to_owned(), new);
+
         Ok(())
     }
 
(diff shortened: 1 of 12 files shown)


crates/matrix-sdk-base/src/store/mod.rs
```diff
@@ -452,31 +452,74 @@ pub enum RoomLoadSettings {
...
+    pub fn as_str(&self) -> &'static str {
+        match self {
+            ThreadSubscriptionStatus::Subscribed { automatic } => {
+                if *automatic {
+                    "automatic"
+                } else {
+                    "manual"
+                }
+            }
+            ThreadSubscriptionStatus::Unsubscribed => "unsubscribed",
```
poljar, line 503: Any reason why we're using `as_str()` for database storage instead of our usual `Serialize` serde stuff?

Additionally if we use any of the more usual traits for this, then we can use `insta` to add more protection to this. `Display`, `Debug`, or `Serialize` would work.

crates/matrix-sdk-base/src/store/memory_store.rs
```diff
@@ -957,15 +957,37 @@ impl StateStore for MemoryStore {
...
+                // doesn't, keep the previous one.
+                (Some(prev_bump), None) => {
+                    new.bump_stamp = Some(prev_bump);
+                }
+
+                // If the previous bump stamp is newer than the new one, don't store the value at
+                // all.
+                (Some(prev_bump), Some(new_bump)) if new_bump <= prev_bump => {
+                    return Ok(());
+                }
```
poljar: Hmm, but this will avoid a transition from unsubscribed to subscribed if the bump stamps are equal, no?

Is such a transition impossible or undesirable?

Furthermore, as you said, this logic is duplicated a couple of times. Can't we make the bump stamp a separate type `BumpStamp(u64)` and the implement either a `is_better()`  method or even add a  `PartialEq` implementation that will contain this logic?

poljar (changes requested):
Not really requesting many changes but I'm not sure if I misunderstood something of if this is incorrect in some edge cases.

crates/matrix-sdk-base/src/store/memory_store.rs
bnjbvr: I think it's not possible, because the bump stamps will increase if the thread subscription goes from subscribed to unsubscribed. My understanding is that same bump stamp = same subscription information; it forms an "index" when combined with the (room id, thread root event id) pair. Will ask to clarify this in the MSC, and maybe I can add a small comment here.

But yeah, the duplication isn't great, will find a way to avoid it.

crates/matrix-sdk-base/src/store/mod.rs
bnjbvr: It makes for a denser format to use `as_str()` (it flattens the subscription into a string, instead of a JSON map), and it allows future use cases like "find the list of all subscribed threads in this room" (which we couldn't do if we serialized the `ThreadSubscription` as a blob inside the DB, b/o encrypted DBs).

poljar: Well I'm not saying that you shouldn't flatten out the serialized format, just that a different trait is usually used for this. A custom `Serialize` implementation can produce the same serialized format as you already have.

bnjbvr: No, it'd be wrong to use a custom `Serialize` impl in the sqlite impl, which has `status` and `bump_stamps` columns.

PR #5591, opened 2025-08-27
common: js_tracing: drop TRACE logs

richvdh:
Now that (since https://github.com/matrix-org/matrix-js-sdk/pull/4918) Element Web uses separate subscribers for each OlmMachine, rather than a single global one with a separate LevelFilter, EW's logs are very verbose because they receive all the TRACE logs.

Dropping the TRACE logs on the floor isn't my favourite solution, but everything else seems to be a bit harder than I have time for right now:

* Sending TRACE logs up to the JS side in case it wants to print them would (a) increase overhead and (b) be an annoying breaking change in JsLogger

* Ideally we'd let the application configure its logging more precisely via an `EnvFilter` or something, but I'm not really sure what the API would look like for that.

So for now, we take the easy path.

### DIFF
--- a/crates/matrix-sdk-common/CHANGELOG.md
+++ b/crates/matrix-sdk-common/CHANGELOG.md
@@ -8,6 +8,8 @@ All notable changes to this project will be documented in this file.
 
 ### Features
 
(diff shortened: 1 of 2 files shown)


richvdh:
> EW is using multiple `OlmMachine`s? Or is that just in case it will do so for multi-account support?

Sorry, that was a bit unclear. EW will not use multiple `OlmMachine`s (at least, not at the same time). 

matrix-js-sdk allows applications to use a separate `Logger` for each `MatrixClient`, and it now uses that `Logger`for the `OlmMachine` that is associated with the client. We therefore have a separate `tracing::Subscriber` instance for each `OlmMachine` in matrix-sdk-crypto-wasm.

EW doesn't really make use of that functionality currently; it's more useful for test setups where we have multiple matrix clients active at a time.

PR #5595, opened 2025-08-28
feat(test): add event factory methods for creating space rooms and populating parent-children relationships

stefanceriu:
What it says on the tin, they're going to be used in the spaces PR (after rebase)


testing/matrix-sdk-test/src/event_factory.rs
```diff
@@ -462,6 +464,11 @@ impl EventBuilder<RoomCreateEventContent> {
         self.content.predecessor = None;
         self
     }
+
+    pub fn with_space_type(mut self) -> Self {
```
poljar, line 469: Docs please.

poljar (approved):
Looks good, let's just document the new method.

PR #5597, opened 2025-08-28
Bump most of our deps

poljar:
This is in preparation for a release, some notable exceptions are made (`rand`, `tracing`, `uniffi`) but I think those are known.


benchmarks/Cargo.toml
```diff
@@ -24,7 +24,7 @@ matrix-sdk-ui.workspace = true
 ruma.workspace = true
 serde.workspace = true
 serde_json.workspace = true
-tempfile = "3.3.0"
+tempfile = "3.21.0"
```
Hywan:
```diff
- tempfile = "3.21.0"
+ tempfile.workspace = true
```

poljar:
> Thanks a lot for taking care of that!

Don't thank me yet, It's all red on the CI and I might just close this :sweat_smile:

poljar:
The perf regression seems to be in the TLS implementation?

image

Which sounds to me that our benchmark is running more code than it should.

PR #5599, opened 2025-08-29
crypto: log message index for megolm sessions received over olm

richvdh:
When we receive a to-device message that contains a megolm decryption key, log the ratchet index of the received key, for debugging.


richvdh:
> We don't log this already 😨

Apparently not :/

PR #5604, opened 2025-09-01
feat: implement msc4308 + automatic background catchup of thread subscriptions

### LINKED ISSUE #5038: Threads: Implement the new participation model API

bnjbvr:
This adds support for:

- msc4308 sliding sync extension: receiving new changes to thread subscriptions over sync, as well as the `prev_batch` token that indicates if there are more subscriptions to catch up with the companion endpoint.
- saving such catchup tokens in the state store
- automatically catch up for extra thread subscriptions in a background task
- in `Room::load_or_fetch_thread_subscriptions()`, use network if we haven't caught up yet, or the local store only, if we have.

I've tested this in isolation, with a few sliding sync tests (using the `MatrixMockServer`; this adds quite a bit of code to it, expanding the PR's size), as well as against the Synapse impl from https://github.com/element-hq/synapse/pull/18695, and it works great in both cases.

Fixes #5038.

### DIFF
--- a/crates/matrix-sdk-base/src/store/traits.rs
+++ b/crates/matrix-sdk-base/src/store/traits.rs
@@ -1151,6 +1151,36 @@ pub enum StateStoreDataValue {
 
     /// A list of knock request ids marked as seen in a room.
     SeenKnockRequests(BTreeMap<OwnedEventId, OwnedUserId>),
+
+    /// A list of tokens to continue thread subscriptions catchup.
+    ///
+    /// See documentation of [`ThreadSubscriptionCatchupToken`] for more
+    /// details.
+    ThreadSubscriptionsCatchupTokens(Vec<ThreadSubscriptionCatchupToken>),
+}
+
+/// Tokens to use when catching up on thread subscriptions.
+///
+/// These tokens are created when the client receives some thread subscriptions
+/// from sync, but the sync indicates that there are more thread subscriptions
+/// available on the server. In this case, it's expected that the client will
+/// call the MSC4308 companion endpoint to catch up (back-paginate) on previous
+/// thread subscriptions.
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
+pub struct ThreadSubscriptionCatchupToken {
+    /// The token to use as the lower bound when fetching new threads
+    /// subscriptions.
+    ///
+    /// In sliding sync, this is the `prev_batch` value of a sliding sync
+    /// response.
+    pub from: String,
+
+    /// The token to use as the upper bound when fetching new threads
+    /// subscriptions.
+    ///
+    /// In sliding sync, it must be set to the `pos` value of the sliding sync
+    /// *request*, which response received a `prev_batch` token.
+    pub to: Option<String>,
 }
 
 /// Current draft of the composer for the room.
@@ -1222,6 +1252,14 @@ impl StateStoreDataValue {
     pub fn into_seen_knock_requests(self) -> Option<BTreeMap<OwnedEventId, OwnedUserId>> {
         as_variant!(self, Self::SeenKnockRequests)
     }
+
+    /// Get this value if it is the data for the thread subscriptions catchup
+    /// tokens.
+    pub fn into_thread_subscriptions_catchup_tokens(
+        self,
+    ) -> Option<Vec<ThreadSubscriptionCatchupToken>> {
+        as_variant!(self, Self::ThreadSubscriptionsCatchupTokens)
+    }
 }
 
 /// A key for key-value data.
@@ -1258,6 +1296,9 @@ pub enum StateStoreDataKey<'a> {
 
     /// A list of knock request ids marked as seen in a room.
     SeenKnockRequests(&'a RoomId),
+
+    /// A list of thread subscriptions catchup tokens.
+    ThreadSubscriptionsCatchupTokens,
 }
 
 impl StateStoreDataKey<'_> {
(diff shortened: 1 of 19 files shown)


crates/matrix-sdk-base/src/store/traits.rs
```diff
@@ -1151,6 +1151,36 @@ pub enum StateStoreDataValue {
...
+    /// details.
+    ThreadSubscriptionsCatchupTokens(Vec<ThreadSubscriptionCatchupToken>),
+}
+
+/// Tokens to use when catching up on thread subscriptions.
+///
+/// These tokens are created when the client receives some thread subscriptions
+/// from sync, but the sync indicates that there are more thread subscriptions
+/// available on the server. In this case, it's expected that the client will
+/// call the MSC4308 companion endpoint to catch up (back-paginate) on previous
```
poljar: Linky please on the MSC.

crates/matrix-sdk/src/client/thread_subscriptions.rs
```diff
@@ -0,0 +1,427 @@
...
+                        StoredThreadSubscription {
+                            status: ThreadSubscriptionStatus::Subscribed {
+                                automatic: thread_sub.automatic,
+                            },
+                            bump_stamp: Some(thread_sub.bump_stamp.into()),
+                        },
+                    )
+                    .await?;
+            }
+        }
```
poljar: Not for now, but I think it'll make sense to turn this into a bulk API so we don't take a transaction per iteration.

Let's just put this into a issue or into a separate PR.

crates/matrix-sdk/src/client/thread_subscriptions.rs
```diff
@@ -0,0 +1,427 @@
...
+    /// tokens list.
+    async fn lock(&self) -> Option<GuardedStoreAccess> {
+        let client = self.client.get()?;
+        let mutex_guard = self.uniq_mutex.clone().lock_owned().await;
+        Some(GuardedStoreAccess {
+            _mutex: mutex_guard,
+            client,
+            is_outdated: self.is_outdated.clone(),
+        })
+    }
```
poljar, line 258: This lock seems useful, but we're not using it in the part of the code that stores the subscriptions.

Probably not that important since there shouldn't thousands of subscriptions flying in and out, but might be prudent to utilize the lock in more places.

crates/matrix-sdk/src/client/thread_subscriptions.rs
```diff
@@ -0,0 +1,427 @@
...
+        if let Some(token) = token {
+            trace!(?token, "Saving catchup token");
+            tokens.push(token);
+        } else {
+            trace!("No catchup token to save");
+        }
+
+        let has_tokens = !tokens.is_empty();
+
+        guard.save_catchup_tokens(tokens).await?;
```
poljar: Could we return the value for `has_tokens` in `save_catchup_tokens()`?

As it is now, we're calculating this flag twice. Once in `save_catchup_tokens()` and once here.

If we ever change this we'd need to remember to modify two places.

crates/matrix-sdk/src/client/thread_subscriptions.rs
```diff
@@ -0,0 +1,427 @@
...
+        // Wake up the catchup task, in case it's waiting.
+        if has_tokens {
+            let _ = self.ping_sender.send(()).await;
+        }
+
+        Ok(())
+    }
+
+    /// The background task listening to new catchup tokens, and using them to
+    /// catch up the thread subscriptions via the msc4308 companion
```
poljar: Linky to the MSC.

crates/matrix-sdk/src/sliding_sync/mod.rs
```diff
@@ -659,15 +659,19 @@ impl SlidingSync {
             || !self.inner.lists.read().await.is_empty()
     }
 
+    /// Sned a single sliding sync request, and returns the response summary.
```
poljar: Send*

poljar (approved):
Well that was fast.

PR #5608, opened 2025-09-02
feat(ui+ffi): Create dedicated `LatestEventValue` types

Hywan:
This set of patches creates a new `LatestEventValue` type in `matrix_sdk_ui::timeline::latest_event::LatestEventValue` mimicking the `matrix_sdk_base::latest_event::LatestEventValue` type but tailored to the `timeline` module. The same thing is done for the `matrix_sdk_ffi` crate. Finally, a new `matrix_sdk_ffi::Room::new_latest_event_value` is created.

Note that `EventTimelineItem` is no longer used to represent a latest event: it has its own dedicated type.


Hywan:
It doesn't support thread for the moment, as for the “previous” API. Thanks for having tested it out! I appreciate!

PR #5609, opened 2025-09-02
refactor(search): Move event processing out of search crate.

dragonfly1033:
Clean up the search crate integration by moving event processing out and having a new 
`RoomIndex::execute(op: RoomIndexOperation)` api.

### DIFF
--- a/crates/matrix-sdk-search/src/index.rs
+++ b/crates/matrix-sdk-search/src/index.rs
@@ -15,8 +15,7 @@
 use std::{fmt, fs, path::Path, sync::Arc};
 
 use ruma::{
-    EventId, OwnedEventId, OwnedRoomId, RoomId, events::AnySyncMessageLikeEvent,
-    room_version_rules::RedactionRules,
+    EventId, OwnedEventId, OwnedRoomId, RoomId, events::room::message::OriginalSyncRoomMessageEvent,
 };
 use tantivy::{
     Index, IndexReader, TantivyDocument,
@@ -35,16 +34,17 @@ use crate::{
 };
 
 /// A struct to represent the operations on a [`RoomIndex`]
-pub(crate) enum RoomIndexOperation {
-    /// Add this document to the index.
-    Add(TantivyDocument),
+#[derive(Debug)]
+pub enum RoomIndexOperation {
+    /// Add this event to the index.
+    Add(OriginalSyncRoomMessageEvent),
     /// Remove all documents in the index where
-    /// [`MatrixSearchIndexSchema::deletion_key()`] matches this event id.
+    /// `MatrixSearchIndexSchema::deletion_key()` matches this event id.
     Remove(OwnedEventId),
     /// Replace all documents in the index where
-    /// [`MatrixSearchIndexSchema::deletion_key()`] matches this event id with
-    /// the new document.
-    Edit(OwnedEventId, TantivyDocument),
+    /// `MatrixSearchIndexSchema::deletion_key()` matches this event id with
+    /// the new event.
+    Edit(OwnedEventId, OriginalSyncRoomMessageEvent),
     /// Do nothing.
     Noop,
 }
@@ -137,37 +137,6 @@ impl RoomIndex {
         RoomIndex::new_with(index, schema, room_id)
     }
 
-    /// Handle [`AnySyncMessageLikeEvent`]
-    ///
-    /// This which will add/remove/edit an event in the index based on the
-    /// event type.
-    pub fn handle_event(
-        &mut self,
-        event: AnySyncMessageLikeEvent,
-        redaction_rules: &RedactionRules,
-    ) -> Result<(), IndexError> {
-        let event_id = event.event_id().to_owned();
-
-        match self.schema.handle_event(event, redaction_rules)? {
-            RoomIndexOperation::Add(document) => {
-                if !self.contains(&event_id) {
-                    self.writer.add(document)?;
-                }
-            }
-            RoomIndexOperation::Remove(event_id) => {
-                self.writer.remove(&event_id);
-            }
-            RoomIndexOperation::Edit(remove_event_id, document) => {
-                self.writer.remove(&remove_event_id);
-                if !self.contains(&event_id) {
-                    self.writer.add(document)?;
-                }
-            }
-            RoomIndexOperation::Noop => {}
-        }
-        Ok(())
-    }
-
     /// Commit added events to [`RoomIndex`]
     pub fn commit(&mut self) -> Result<OpStamp, IndexError> {
         let last_commit_opstamp = self.writer.commit()?; // TODO: This is blocking. Handle it.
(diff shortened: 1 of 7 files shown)


crates/matrix-sdk-search/src/index.rs
```diff
@@ -218,6 +187,32 @@ impl RoomIndex {
         Ok(ret)
     }
 
+    /// Execute [`RoomIndexOperation`]
+    ///
+    /// This which will add/remove/edit an event in the index based on the
+    /// operation.
+    pub fn execute(&mut self, operation: RoomIndexOperation) -> Result<(), IndexError> {
+        warn!("doing {operation:?}");
```
poljar: Why is that a warning?

crates/matrix-sdk-search/src/index.rs
```diff
@@ -35,16 +34,17 @@ use crate::{
...
+    /// `MatrixSearchIndexSchema::deletion_key()` matches this event id.
     Remove(OwnedEventId),
     /// Replace all documents in the index where
-    /// [`MatrixSearchIndexSchema::deletion_key()`] matches this event id with
-    /// the new document.
-    Edit(OwnedEventId, TantivyDocument),
+    /// `MatrixSearchIndexSchema::deletion_key()` matches this event id with
+    /// the new event.
+    Edit(OwnedEventId, OriginalSyncRoomMessageEvent),
     /// Do nothing.
```
poljar, line 48: That's pre-existing, but what is the point of the `Noop`?

crates/matrix-sdk-search/src/index.rs
```diff
@@ -235,12 +230,57 @@ mod tests {
...
+        index: &mut RoomIndex,
+        event: AnySyncMessageLikeEvent,
+    ) -> Result<(), IndexError> {
+        if let AnySyncMessageLikeEvent::RoomMessage(ev) = event
+            && let Some(ev) = ev.as_original()
+            && ev.content.relates_to.is_none()
+        {
+            return index.execute(RoomIndexOperation::Add(ev.clone()));
+        }
+        panic!("Event was not a relationless OriginalSyncRoomMessageEvent.")
```
poljar, line 261: Can't this be a no-op with a warning? Even better, why not accept a `OriginalSyncRoomMessageEvent` if that's the only type we accept?

Generally having an API that panics because you pass in the wrong thing is an anti-pattern. We have strong types that can enforce such a contract much better.

crates/matrix-sdk-search/src/index.rs
```diff
@@ -235,12 +230,57 @@ mod tests {
...
+        index: &mut RoomIndex,
+        event_id: &EventId,
+        new: AnySyncMessageLikeEvent,
+    ) -> Result<(), IndexError> {
+        if let AnySyncMessageLikeEvent::RoomMessage(ev) = new
+            && let Some(ev) = ev.as_original()
+        {
+            return index.execute(RoomIndexOperation::Edit(event_id.to_owned(), ev.clone()));
+        }
+        panic!("Event was not a relationless OriginalSyncRoomMessageEvent.")
```
poljar: Again, this contract could be enforced using types.

crates/matrix-sdk/src/event_cache/search.rs
```diff
@@ -37,36 +45,79 @@ async fn get_most_recent_edit(
     };
 
     match related.last().unwrap_or(&original_ev).raw().deserialize() {
-        Ok(AnySyncTimelineEvent::MessageLike(latest)) => Some(latest),
+        Ok(AnySyncTimelineEvent::MessageLike(AnySyncMessageLikeEvent::RoomMessage(latest))) => {
+            warn!("most recent is {}", latest.event_id());
```
poljar: Another warning.

poljar (changes requested):
There are some problems we have to fix here, but overall it looks sensible.

crates/matrix-sdk-search/src/index.rs
dragonfly1033: When handling a chain of possibilities represented by functions which return `Option<RoomIndexOperation>` `None` can represent "this possibility is not the case try something else" or "do nothing". To disambiguate I introduced `Noop`. Here's the code where it is used:

```rust
/// If the given [`OriginalSyncRoomMessageEvent`] is an edit we make an
/// [`RoomIndexOperation::Edit`] with the new most recent version of the
/// original.
async fn handle_possible_edit(
    event: &OriginalSyncRoomMessageEvent,
    cache: &RoomEventCache,
) -> Option<RoomIndexOperation> {
    if let Some(Relation::Replacement(replacement_data)) = &event.content.relates_to {
        if let Some(recent) = get_most_recent_edit(cache, &replacement_data.event_id).await {
            return Some(RoomIndexOperation::Edit(replacement_data.event_id.clone(), recent));
        } else {
            return Some(RoomIndexOperation::Noop);
        }
    }
    None
}

/// Return a [`RoomIndexOperation::Edit`] or [`RoomIndexOperation::Add`]
/// depending on the message.
async fn handle_room_message(
    event: SyncRoomMessageEvent,
    cache: &RoomEventCache,
) -> Option<RoomIndexOperation> {
    if let Some(event) = event.as_original() {
        return handle_possible_edit(event, cache).await.or(get_most_recent_edit(
            cache,
            &event.event_id,
        )
        .await
        .map(RoomIndexOperation::Add));
    }
    None
}
```

Is there a different pattern you would prefer?

crates/matrix-sdk-search/src/index.rs
dragonfly1033: Sorry, was used during deebugging and I forgot to remove it.

crates/matrix-sdk/src/event_cache/search.rs
dragonfly1033: Same as above

crates/matrix-sdk-search/src/index.rs
dragonfly1033: How would I enforce `ev.content.relates_to = None` with types? The intent here is to ensure test writers don't use `index_message` for edit events. The intent in the one below is to ensure test writers don't try editing an edit because edits should refer to the original.

poljar: Ah indeed, there are additional restrictions I gleamed over.

I guess it's fine to leave it as is, since this isn't a `pub` method.

crates/matrix-sdk-search/src/index.rs
poljar: Am I wrong to think that if you just take `OriginalSyncRoomMessageEvent` instead of `AnySyncMessageLikeEvent` the panic can't happen anymore and the logic didn't change?

But perhaps not important enough to change this for a private method.

crates/matrix-sdk-search/src/index.rs
poljar: Oh I see, the `or()`:

https://github.com/matrix-org/matrix-rust-sdk/blob/4a2d4b6d945b30bc9980bba73583b6343a07e61c/crates/matrix-sdk/src/event_cache/search.rs#L103-L107

wouldn't behave correctly if we don't differentiate between `None` and `Some(Noop)`.

Alright.

crates/matrix-sdk-search/src/index.rs
dragonfly1033: Yes, I think here that would work.

poljar: Wanna change it or leave it as is?

PR #5612, opened 2025-09-02
feat(sdk): add more timers to sync processing

bnjbvr:
Not for the fainted of heart.

This would help pinpoint what is taking so long during some sync processing.


Hywan (approved):

Hywan (approved):
Still good to me!

Issue #5580, opened 2025-08-26
Expose sending and receiving the `m.rtc.decline` events in the `FFI` layer


toger5:
To implement call declines on EX we need to expose this event type to the FFI.

Most context can be found in the PR implementing this event in ruma: https://github.com/ruma/ruma/pull/2198

We need:
 - to send the event from Element X in the notification worker (while showing the full screen notification)
 - we need to be able to listen to a room timeline when we are currently playing a call ring for this event type to stop ringing on all other devices if such an event arrives.

PR #5614, opened 2025-09-03
bindings: MSC4310 call decline and subscribe to decline events

BillCarsonFr:
Fixes https://github.com/matrix-org/matrix-rust-sdk/issues/5580 

Signed-off-by:

### DIFF
--- a/bindings/matrix-sdk-ffi/src/room/mod.rs
+++ b/bindings/matrix-sdk-ffi/src/room/mod.rs
@@ -1006,6 +1006,41 @@ impl Room {
         Ok(())
     }
 
+    /// Declines a call (and stop ringing).
+    ///
+    /// # Arguments
+    ///
+    /// * `rtc_notification_event_id` - the event id of the m.rtc.notification
+    ///   event.
+    pub async fn decline_call(&self, rtc_notification_event_id: String) -> Result<(), ClientError> {
+        let parsed_id = EventId::parse(rtc_notification_event_id.as_str())?;
+        let content = self.inner.make_decline_call_event(&parsed_id).await?;
+
+        self.inner.send_queue().send(content.into()).await?;
+
+        Ok(())
+    }
+
+    /// Subscribes to call decline for a currently ringing call, using a
+    /// `listener` to be notified when someone declines.
+    ///
+    /// Will error if `rtc_notification_event_id` is not a valid event id.
+    /// Use the [`TaskHandle`] to cancel the subscription.
+    pub fn subscribe_to_call_decline_events(
+        self: Arc<Self>,
+        rtc_notification_event_id: String,
+        listener: Box<dyn CallDeclineListener>,
+    ) -> Result<Arc<TaskHandle>, ClientError> {
+        let parsed_id = EventId::parse(rtc_notification_event_id.as_str())?;
+        Ok(Arc::new(TaskHandle::new(get_runtime_handle().spawn(async move {
+            let (_event_handler_drop_guard, mut subscriber) =
+                self.inner.subscribe_to_call_decline_events(&parsed_id);
+            while let Ok(user_id) = subscriber.recv().await {
+                listener.call(user_id.to_string());
+            }
+        }))))
+    }
+
     /// Subscribes to live location shares in this room, using a `listener` to
     /// be notified of the changes.
     ///
@@ -1151,6 +1186,12 @@ pub struct ThreadSubscription {
 #[matrix_sdk_ffi_macros::export(callback_interface)]
 pub trait LiveLocationShareListener: SyncOutsideWasm + SendOutsideWasm {
     fn call(&self, live_location_shares: Vec<LiveLocationShare>);
+}
+
+/// A listener for receiving call decline events in a room.
+#[matrix_sdk_ffi_macros::export(callback_interface)]
+pub trait CallDeclineListener: SyncOutsideWasm + SendOutsideWasm {
+    fn call(&self, decliner_user_id: String);
 }
 
 impl From<matrix_sdk::room::knock_requests::KnockRequest> for KnockRequest {
--- a/bindings/matrix-sdk-ffi/src/room/mod.rs
+++ b/bindings/matrix-sdk-ffi/src/room/mod.rs
@@ -1014,6 +1014,7 @@ impl Room {
     ///   event.
     pub async fn decline_call(&self, rtc_notification_event_id: String) -> Result<(), ClientError> {
         let parsed_id = EventId::parse(rtc_notification_event_id.as_str())?;
+
         let content = self.inner.make_decline_call_event(&parsed_id).await?;
 
         self.inner.send_queue().send(content.into()).await?;
@@ -1032,9 +1033,11 @@ impl Room {
         listener: Box<dyn CallDeclineListener>,
     ) -> Result<Arc<TaskHandle>, ClientError> {
         let parsed_id = EventId::parse(rtc_notification_event_id.as_str())?;
+
         Ok(Arc::new(TaskHandle::new(get_runtime_handle().spawn(async move {
             let (_event_handler_drop_guard, mut subscriber) =
                 self.inner.subscribe_to_call_decline_events(&parsed_id);
+
             while let Ok(user_id) = subscriber.recv().await {
                 listener.call(user_id.to_string());
             }
--- a/crates/matrix-sdk/Cargo.toml
+++ b/crates/matrix-sdk/Cargo.toml
@@ -64,7 +64,8 @@ sso-login = ["local-server"]
 
 uniffi = ["dep:uniffi", "matrix-sdk-base/uniffi", "dep:matrix-sdk-ffi-macros"]
 
-experimental-widgets = ["dep:uuid", "experimental-send-custom-to-device"]
+unstable-msc4310 = ["ruma/unstable-msc4310"]
+experimental-widgets = ["dep:uuid", "experimental-send-custom-to-device", "unstable-msc4310"]
 
 docsrs = ["e2e-encryption", "sqlite", "indexeddb", "sso-login", "qrcode"]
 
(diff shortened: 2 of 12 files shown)


crates/matrix-sdk/Cargo.toml
```diff
@@ -64,7 +64,8 @@ sso-login = ["local-server"]
 
 uniffi = ["dep:uniffi", "matrix-sdk-base/uniffi", "dep:matrix-sdk-ffi-macros"]
 
-experimental-widgets = ["dep:uuid", "experimental-send-custom-to-device"]
+unstable-msc4310 = ["ruma/unstable-msc4310"]
```
poljar: Is there a reason we'd like to have this as a separate feature? In other words, is this features useful on its own, without the `experimental-widgets` feature?

bindings/matrix-sdk-ffi/src/room/mod.rs
```diff
@@ -1006,6 +1006,41 @@ impl Room {
...
+    /// `listener` to be notified when someone declines.
+    ///
+    /// Will error if `rtc_notification_event_id` is not a valid event id.
+    /// Use the [`TaskHandle`] to cancel the subscription.
+    pub fn subscribe_to_call_decline_events(
+        self: Arc<Self>,
+        rtc_notification_event_id: String,
+        listener: Box<dyn CallDeclineListener>,
+    ) -> Result<Arc<TaskHandle>, ClientError> {
+        let parsed_id = EventId::parse(rtc_notification_event_id.as_str())?;
```
poljar, line 1035: A bit of white space helps with the readability here.

```suggestion
        let parsed_id = EventId::parse(rtc_notification_event_id.as_str())?;

```

bindings/matrix-sdk-ffi/src/room/mod.rs
```diff
@@ -1006,6 +1006,41 @@ impl Room {
...
+        Ok(Arc::new(TaskHandle::new(get_runtime_handle().spawn(async move {
+            let (_event_handler_drop_guard, mut subscriber) =
+                self.inner.subscribe_to_call_decline_events(&parsed_id);
```
poljar, line 1039:
```suggestion
                self.inner.subscribe_to_call_decline_events(&parsed_id);

```

crates/matrix-sdk/tests/integration/room/calls.rs
```diff
@@ -0,0 +1,161 @@
+#![cfg(feature = "unstable-msc4310")]
```
poljar: No need for this as you already feature gated it in `room/mod.rs`.

crates/matrix-sdk/src/room/calls.rs
```diff
@@ -0,0 +1,115 @@
...
+    let AnySyncTimelineEvent::MessageLike(AnySyncMessageLikeEvent::CallNotify(notify)) = event
+    else {
+        return Err(CallError::BadEventType);
+    };
+
+    if notify.sender() == own_user_id {
+        return Err(CallError::DeclineOwnCall);
+    }
+
+    Ok(RtcDeclineEventContent::new(notification_event_id))
```
poljar: Those returns are really breaking the flow here. I think this would read much nicer as an `if let/else if/else` each of them without the explicit return.

crates/matrix-sdk/src/room/calls.rs
```diff
@@ -0,0 +1,115 @@
...
+    }
+
+    /// Subscribe to decline call evvent for this room.
+    ///
+    /// The returned receiver will receive the sender UserID for each decline
+    /// for the matching notify event.
+    pub fn subscribe_to_call_decline_events(
+        &self,
+        notification_event_id: &EventId,
+    ) -> (EventHandlerDropGuard, broadcast::Receiver<OwnedUserId>) {
```
poljar: Can we get an example on in the docs how this is supposed to be used?

crates/matrix-sdk/src/room/calls.rs
```diff
@@ -0,0 +1,115 @@
...
+    ///
+    /// The event can then be sent with [`Room::send`] or a
+    /// [`crate::send_queue::RoomSendQueue`].
+    #[instrument(skip(self), fields(room = %self.room_id()))]
+    pub async fn make_decline_call_event(
+        &self,
+        notification_event_id: &EventId,
+    ) -> Result<RtcDeclineEventContent, CallError> {
+        make_call_decline_event(self, self.own_user_id(), notification_event_id).await
+    }
```
poljar, line 61: A bit of a strange API to have on the room object. We don't have such helpers anywhere else, but ok I don't have a better idea if you want to be able to use the send queue as well as the room object to send this.

The `make_call_decline_event` method only checks that the `notification_event_id` is for a valid call that we should be able to decline.

Perhaps it would make sense to rename the `make_call_decline()` function into something like `check_if_foo_valid()` and do the actual content creation in this method.

This would make the intention of the functions more clear and highlight the separate concerns.

crates/matrix-sdk/src/room/calls.rs
```diff
@@ -0,0 +1,115 @@
...
+        let drop_guard = self.client().event_handler_drop_guard(decline_call_event_handler_handle);
+        (drop_guard, receiver)
+    }
+}
+
+async fn make_call_decline_event<S: EventSource>(
+    source: S,
+    own_user_id: &UserId,
+    notification_event_id: &EventId,
+) -> Result<RtcDeclineEventContent, CallError> {
```
poljar: Is it necessary for this to be a generic function? As far as I can see you're always mocking things to get a room and never use any other `EventSource`.

poljar (changes requested):
Looking mostly goood, I left some nits here and there.

crates/matrix-sdk/src/room/calls.rs
BillCarsonFr: > A bit of a strange API to have on the room object. We don't have such helpers anywhere else, but ok I don't have a better idea if you want to be able to use the send queue as well as the room object to send this.

There are helpers like that I think (edits. replies) It was a suggestion for @bnjbvr :
> You shouldn't have a function to send an event, instead you should have functions that allow you to create the final event, and the caller decides whether to send via the Room.send or the send queue
That's why we have methods like Room::make_reply_event, for instance

I mostly tried to be consistent to how things are done for edits and replies

poljar: Oh right, we do have those. :older_man:

crates/matrix-sdk/src/room/calls.rs
BillCarsonFr: I tried to be consistent with the way `make_edit_event` is doing it, as it is also a relation.
I am not testing it to simplify the mocking by using the event cache, but I can add a test without event_cache and then mock the `/message` API?

poljar: Well ok, but if we don't use the generics then it's a bit pointless to introduce them. They don't hurt that much but they do slow down compile times a bit.

> I am not testing it to simplify the mocking by using the event cache, but I can add a test without event_cache and then mock the /message API?

I don't think that's necessary, writing the test the way you did is a more realistic scenario. Those generics are there if it's very hard to inject the necessary test data into various places.

crates/matrix-sdk/tests/integration/room/calls.rs
BillCarsonFr: It is not in that module that I did like that.
But I could do the same to be consistent. Fells bad to guard a hole `calls.rs` file behind a very specific `msc4310` for declining call... Need to think about it

crates/matrix-sdk/Cargo.toml
BillCarsonFr: Actually we are not doing that for the notify event (unstable-msc4075), so I removed it.
If we really want to put it under a feature flag, it would rather be an experimental-call features that depends on widget, with conditional ruma features... but this would require more work

crates/matrix-sdk/tests/integration/room/calls.rs
BillCarsonFr: removed that feature

crates/matrix-sdk/src/room/calls.rs
BillCarsonFr: Updated

crates/matrix-sdk/src/room/calls.rs
BillCarsonFr: added

crates/matrix-sdk/src/room/calls.rs
BillCarsonFr: simplified to be just Room

BillCarsonFr:
> Thanks, looks good. Let's just add a changelog entry for this.

Added https://github.com/matrix-org/matrix-rust-sdk/pull/5614/commits/f70afad456975af3f65f95d04dee68fd9dc49da5

poljar:
The cargo-deny failure is due to a new RUSTSEC advisory, fixed by: https://github.com/matrix-org/matrix-rust-sdk/pull/5633.

No need to rebase but I'll merge this PR after #5633 is merged.

crates/matrix-sdk/src/room/calls.rs
```diff
@@ -0,0 +1,163 @@
...
+    if let AnySyncTimelineEvent::MessageLike(AnySyncMessageLikeEvent::CallNotify(notify)) = event {
+        if notify.sender() == own_user_id {
+            // Cannot decline own call.
+            Err(CallError::DeclineOwnCall)
+        } else {
+            Ok(RtcDeclineEventContent::new(notification_event_id))
+        }
+    } else {
+        Err(CallError::BadEventType)
+    }
```
toger5, line 162: EC will always sent both events (`...rtc.notifcation` + `...call.notify`) but the spec does not define a relation to a call.notify event (they are deprecated).

To make all platforms consistent, we would need to decline on the rtc.notification event. Otherwise EW will not detect the decline and will not stop ringing.
As a stop gap we could patch EW to listen to any decline event independent of its `m.relation.event_id`.

But this is only really correct if it checks for m.rtc notification introduced here: https://github.com/ruma/ruma/pull/2199

PR #5617, opened 2025-09-03
feat(ui): Add the new `latest_event` room list sorter, and update the `recency` sorter

Hywan:
This set of patches updates the room list to sort rooms by their “latest event”. What does it mean?

- We want a room with a **local** latest event to be at the top of the room list,
- We want a room to have an accurate position in the room list, which means we want to rely on the latest event instead of the room `recency_stamp` (also known as the `bump_stamp` in sliding sync terminology):
  - Why was `recency_stamp` inaccurate? Because the server might _bump_ a room based on certain events, so the room will _bump_ in the room list, but the latest event won't change because the event causing the bump is **not** a latest event candidate. Because the latest event is used by som apps to display the room “last updated time” in the room, the user might see unordered rooms!
  - One may argue that we should keep the list of events triggering a bump on the server in-sync with the list of latest event candidates on the client side. I would happily counter-argue that it's server-dependent and client-dependent. Not all servers and not all clients want the same behaviour. The list on the server is “best-effort”. Also, it's not configurable anymore with Simplified Sliding Sync (MSC4186).
  - One may also argue that sorting by latest event's timestamp means we use [the famous `origin_server_rs` value](https://spec.matrix.org/v1.15/client-server-api/), which is a mistake because it _can_ be forged by malicious servers! Yes, that's true, but:
    - There is no security risk involved here. At worst, a room is _sticked_ at the top or the bottom of the room list, that's it.
    - We were using this value in the past and we agreed it was fine (happy to re-discuss that if needed).
    - Some homeservers can use the `origin_server_ts` value of the last event in the room as the value for `bump_stamp`. We can't control this, it's an opaque value.

All-in-all, this set of patches creates a new `latest_event` sorter for the room list. Then it updates the `recency` sorter to be “latest-event aware”. Finally, it installs the new `latest_event` sorter.

### DIFF
--- a/crates/matrix-sdk-base/src/latest_event.rs
+++ b/crates/matrix-sdk-base/src/latest_event.rs
@@ -41,6 +41,35 @@ pub enum LatestEventValue {
     LocalCannotBeSent(LocalLatestEventValue),
 }
 
+impl LatestEventValue {
+    /// Get the timestamp of the [`LatestEventValue`].
+    ///
+    /// If it's [`None`], it returns `None`. If it's [`Remote`], it returns the
+    /// `origin_server_ts`. If it's [`LocalIsSending`] or [`LocalCannotBeSent`],
+    /// it returns the [`timestamp`] value.
+    ///
+    /// [`None`]: LatestEventValue::None
+    /// [`Remote`]: LatestEventValue::Remote
+    /// [`LocalIsSending`]: LatestEventValue::LocalIsSending
+    /// [`LocalCannotBeSent`]: LatestEventValue::LocalCannotBeSent
+    /// [`timestamp`]: LocalLatestEventValue::timestamp
+    pub fn timestamp(&self) -> Option<u64> {
+        match self {
+            Self::None => None,
+            Self::Remote(remote_latest_event_value) => remote_latest_event_value
+                .kind
+                .raw()
+                .get_field::<u64>("origin_server_ts")
+                .ok()
+                .flatten(),
+            Self::LocalIsSending(LocalLatestEventValue { timestamp, .. })
+            | Self::LocalCannotBeSent(LocalLatestEventValue { timestamp, .. }) => {
+                Some(timestamp.get().into())
+            }
+        }
+    }
+}
+
 /// Represents the value for [`LatestEventValue::Remote`].
 pub type RemoteLatestEventValue = TimelineEvent;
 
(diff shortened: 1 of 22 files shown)


crates/matrix-sdk-ui/src/room_list_service/sorters/latest_event.rs
```diff
@@ -0,0 +1,267 @@
...
+use matrix_sdk::latest_events::LatestEventValue;
+
+use super::{Room, Sorter};
+
+struct LatestEventMatcher<F>
+where
+    F: Fn(&Room, &Room) -> (LatestEventValue, LatestEventValue),
+{
+    latest_events: F,
+}
```
poljar: Is there a point of this being a struct? Seems like function would have done just fine?

crates/matrix-sdk-ui/src/room_list_service/sorters/latest_event.rs
```diff
@@ -0,0 +1,267 @@
...
+                LatestEventValue::LocalIsSending(_) | LatestEventValue::LocalCannotBeSent(_),
+                LatestEventValue::None | LatestEventValue::Remote(_),
+            ) => Ordering::Less,
+
+            // `Local*` == `Local*`
+            (
+                LatestEventValue::LocalIsSending(_) | LatestEventValue::LocalCannotBeSent(_),
+                LatestEventValue::LocalIsSending(_) | LatestEventValue::LocalCannotBeSent(_),
+            ) => Ordering::Equal,
+        }
```
poljar: This looks awfully similar to a `PartialOrd`/`Ord` implementation for `LatestEventValue`. Why can't it be one? Matches is also a bit of a confusing name for something that returns a `Ordering`.

crates/matrix-sdk-base/src/latest_event.rs
```diff
@@ -41,6 +41,35 @@ pub enum LatestEventValue {
     LocalCannotBeSent(LocalLatestEventValue),
 }
 
+impl LatestEventValue {
+    /// Get the timestamp of the [`LatestEventValue`].
+    ///
+    /// If it's [`None`], it returns `None`. If it's [`Remote`], it returns the
+    /// `origin_server_ts`. If it's [`LocalIsSending`] or [`LocalCannotBeSent`],
```
poljar, line 48: We discussed this in chat but the discussion died down, so I'll repeat here.

To avoid malicious homeservers and users manipulating our room ordering using the `origin_server_ts` can't we record a local timestamp whenever we create a new `LatestEventValue`?

crates/matrix-sdk-ui/src/room_list_service/sorters/recency.rs
```diff
@@ -62,26 +64,185 @@ where
...
 
     move |left, right| -> Ordering { matcher.matches(left, right) }
 }
 
+/// Extract the recency stamp from either the [`RoomInfo::new_latest_event`] or
+/// from [`RoomInfo::recency_stamp`].
+///
+/// We must be very careful to return data of the same nature: either a _recency
+/// stamp_ from the [`LatestEventValue`], or from the
+/// [`RoomInfo::recency_stamp`], but we **must never** mix both. The
```
poljar, line 81: Sounds like some stronger types that would prevent mixing of those `u64` values would have been useful here.

crates/matrix-sdk-ui/src/room_list_service/sorters/latest_event.rs
Hywan: Hmm, it was simpler from a test perspective, and that's how we've been doing in the other filters and sorters, but sure, we can change for a function! Even in the test, it can work. Do you want me to change for all filters and sorters (no problem to do it, it's pretty straightforward). Do you think it can be more performant?

crates/matrix-sdk-ui/src/room_list_service/sorters/latest_event.rs
Hywan: I agree, at first glance. However, the meaning of `PartialOrd`/`Ord` here is very specific to this sorter, and may not be suitable for a generic implementation over `LatestEventValue`.

Moreover, note that `LatestEventValue::None` and `LatestEventValue::Remote` are considered `Ordering::Equal`, so that we are sure `LatestEventValue::Local*` come first, and it can be combined with the lexicographic sorter nicely. Again, it's super specific to room list sorters, and I'm not sure having this semantics for all `LatestEventValue` is desired.

Finally, if we decide to define the semantics of sorting rooms based on their `LatestEventValue` (if any) directly in `LatestEventValue`, it's fragile and can be broken easily since `RoomList` and `LatestEventValue` don't live in the same crates.

However, I agree we can rename `matches` to `cmp` (or `compare` to avoid confusion with `Ord`)! Do you agree with this name?

crates/matrix-sdk-ui/src/room_list_service/sorters/latest_event.rs
poljar: > Do you want me to change for all filters and sorters (no problem to do it, it's pretty straightforward).

Well it would be simpler code, so if you don't mind, sure.

> Do you think it can be more performant?

Theoretically sure, though not in a meaningful manner. Realistically I expect the compiler to figure this out.

crates/matrix-sdk-base/src/latest_event.rs
Hywan: Yeah. I've been thinking about that.

When we receive an event via the sync, we can record the `now()` as the _local timestamp_. However, it means that all events from all rooms for a particular sync will have the same _local timestamp_, and thus, we can't use that to sort the rooms (because all “recorded timestamps” will be equal).

I think we have to keep `origin_server_ts` for the moment. We can warn it's a known limitations. It's also used to display the event's time in the timeline for example. If we can't trust this value, it's a protocol issue I suppose. I don't see how to solve that for the moment.

crates/matrix-sdk-ui/src/room_list_service/sorters/recency.rs
Hywan: Good idea!

crates/matrix-sdk-ui/src/room_list_service/sorters/latest_event.rs
poljar: Sure, `compare` sounds fine.

crates/matrix-sdk-ui/src/room_list_service/sorters/recency.rs
Hywan: I've created `RoomRecencyStamp` in https://github.com/matrix-org/matrix-rust-sdk/pull/5617/commits/949f4275f8b46f905c26f41506c5f9e6e0c24cb9.

At the end `extract_recency_stamp` (renamed to avoid confusion with `RoomRecencyStamp`) still return a pair of `Option<u64>`, but the code makes it harder to confuse the recency stamp and the timestamp.

poljar (approved):
Alright, looks good.

There's still the question of using `origin_server_ts` exclusively for this, but as you said, this can be a separate PR.

crates/matrix-sdk-ui/src/room_list_service/sorters/latest_event.rs
Hywan: Done! In https://github.com/matrix-org/matrix-rust-sdk/pull/5617/commits/5a2c89f187db798b1c8eaa0bf7f322f238f2ee34?diff=split&w=1 and https://github.com/matrix-org/matrix-rust-sdk/pull/5617/commits/d9db53e237a8457b291652796120b1352e805b86?diff=split&w=1.

PR #5618, opened 2025-09-03
fix(room service): enable the thread subscriptions extension based on server support

bnjbvr:
Otherwise, when running against a server that doesn't support thread subscriptions, the client may falsely think that its thread subscriptions are *not* outdated (when receiving an empty, defaulted `ThreadSubscriptions` section in the SSS response). As a result, it would always use the store, not find most thread subscriptions, and conclude "nope, you haven't subscribed to any thread", which is wrong.

(Looking for support of MSC4306 is _close_ to what we want, which would be support of MSC4308 itself, coming as part of https://github.com/element-hq/synapse/pull/18695, and that will land despite MSC4306 being already advertised, but that's good enough, since it's unlikely many servers have enabled it by default yet.)

### DIFF
--- a/crates/matrix-sdk-ui/tests/integration/room_list_service.rs
+++ b/crates/matrix-sdk-ui/tests/integration/room_list_service.rs
@@ -1,4 +1,4 @@
-use std::{ops::Not, sync::Arc};
+use std::{collections::BTreeMap, ops::Not, sync::Arc};
 
 use assert_matches::assert_matches;
 use eyeball_im::VectorDiff;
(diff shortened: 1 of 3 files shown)


bnjbvr:
Sure, should be fairly trivial albeit verbose to test.

crates/matrix-sdk-ui/tests/integration/room_list_service.rs
```diff
@@ -2893,3 +2893,196 @@ async fn test_multiple_timeline_init() {
...
+                    "all_rooms": {
+                        "filters": {},
+                        "ranges": [ [ 0, 19, ], ],
+                        "required_state": [
+                            [ "m.room.name", "" ],
+                            [ "m.room.encryption", "" ],
+                            [ "m.room.member", "$LAZY" ],
+                            [ "m.room.member", "$ME" ],
+                            [ "m.room.topic", "", ],
+                             [ "m.room.avatar", "", ],
```
Hywan: Indentation seems broken here.

crates/matrix-sdk-ui/tests/integration/room_list_service.rs
```diff
@@ -2893,3 +2893,196 @@ async fn test_multiple_timeline_init() {
...
+                "all_rooms": {
+                    "filters": {},
+                    "ranges": [ [ 0, 19, ], ],
+                    "required_state": [
+                        [ "m.room.name", "" ],
+                        [ "m.room.encryption", "" ],
+                        [ "m.room.member", "$LAZY" ],
+                        [ "m.room.member", "$ME" ],
+                        [ "m.room.topic", "", ],
+                         [ "m.room.avatar", "", ],
```
Hywan: Same.

PR #5621, opened 2025-09-03
perf: make room notification caching up to 4000% faster

bnjbvr:
The room notification modes will only change when the push rules have changed; otherwise, recomputing them is a no-op, and only wastes time.

Here's a distribution of the worst processing times under this method, as measured in `multiverse` with some big test account running (fresh, i.e. after an initial response) for a few minutes:

Before:
0.036192141s
0.037368775s
0.039177686s
0.039883797s
0.041679405s

After:
0.000000249s
0.000000397s
0.000000899s
0.001080307s
0.001340611s

The new distribution makes sense: the push rules event comes only once, in the initial response, so they're processed only once. Then, there's no need to process it ever again (in my session I've never changed the processing times), so it's almost instantaneous the subsequent times.

Found thanks to https://github.com/matrix-org/matrix-rust-sdk/pull/5612

### DIFF
--- a/crates/matrix-sdk/src/sliding_sync/client.rs
+++ b/crates/matrix-sdk/src/sliding_sync/client.rs
@@ -5,9 +5,13 @@ use matrix_sdk_base::{
     RequestedRequiredStates, ThreadSubscriptionCatchupToken, sync::SyncResponse, timer,
 };
 use matrix_sdk_common::deserialized_responses::ProcessedToDeviceEvent;
-use ruma::api::{
-    FeatureFlag, SupportedVersions,
-    client::sync::sync_events::v5::{self as http, response},
+use ruma::{
+    OwnedRoomId,
+    api::{
+        FeatureFlag, SupportedVersions,
+        client::sync::sync_events::v5::{self as http, response},
+    },
+    events::GlobalAccountDataEventType,
 };
 use tracing::error;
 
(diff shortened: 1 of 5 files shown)


crates/matrix-sdk/src/sliding_sync/client.rs
```diff
@@ -242,13 +245,25 @@ impl SlidingSyncResponseProcessor {
...
+        for room_id in response.rooms.joined.keys() {
+            let Some(room) = client.get_room(room_id) else {
+                error!(
+                    ?room_id,
+                    "Cannot post process a room in sliding sync because it is missing"
+                );
+                continue;
+            };
+
+            room.user_defined_notification_mode().await;
```
bnjbvr: Note: an extra perf improvement would be to create the `NotificationSettings` instance once (instead of once per call to `user_defined_notification_mode()`), and use it to update all the room notification modes at once, instead of creating the `NotificationSettings` (which loads the push rules from DB) once per room.

Also, I think it's still necessary to figure what are the _newly_ joined rooms, and compute the user defined notification mode for those.

Not as simple as I imagined first, unfortunately.

bnjbvr:
It is indeed *somewhat* tested, but my self-review comment highlights at least two issues that must exist, and that will require new tests. Back to the drawing board!

bnjbvr:
@andybalaam I've now added a regression test for the thing I've spotted while implementing this (the room notification modes should be updated when the push rules event is updated, even if the room isn't included in the sliding sync response), and tweaked the implementation to avoid reloading the push rules event from the DB every single time.

Also updated the performance numbers in the initial comment. The PR is different enough that I think a new review might be useful.

PR #5587, opened 2025-08-27
feat(sdk,base): Introduce the new `LatestEvents` API, part 5: Make it persistent

Hywan:
This set of patches make the `LatestEventValue` persistent by:

- setting it in `RoomInfo`,
- persisting the `RoomInfo` in the `StateStore`,
- emitting a `RoomInfoNotableUpdateReasons::LATEST_EVENT`.

---

* Address https://github.com/matrix-org/matrix-rust-sdk/issues/4112

PR #5624, opened 2025-09-03
chore: Remove the old Latest Event API

Hywan:
This set of patches remove the old latest event API, since the new one is now available and working.

Finally, this set of patches also remove the `new_` prefix from the new latest event API.

🎉 

---

* Address https://github.com/element-hq/element-x-ios/issues/4814

### DIFF
--- a/bindings/matrix-sdk-ffi/CHANGELOG.md
+++ b/bindings/matrix-sdk-ffi/CHANGELOG.md
@@ -6,6 +6,13 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased] - ReleaseDate
 
+### Features
+
+- [**breaking**]: The new Latest Event API replaces the old API.
+  `Room::new_latest_event` overwrites the `Room::latest_event` method. See the
+  documentation of `matrix_sdk::latest_event` to learn about the new API.
+  [#5624](https://github.com/matrix-org/matrix-rust-sdk/pull/5624/)
+
 ## [0.16.0] - 2025-12-04
 
 ### Breaking changes
--- a/crates/matrix-sdk-ui/CHANGELOG.md
+++ b/crates/matrix-sdk-ui/CHANGELOG.md
@@ -6,6 +6,14 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased] - ReleaseDate
 
+### Features
+
+- [**breaking**]: The new Latest Event API replaces the old API. All the
+  `new_` prefixes have been removed. The following methods are removed:
+  `EventTimelineItem::from_latest_event`, and `Timeline::latest_event`. See the
+  documentation of `matrix_sdk::latest_event` to learn about the new API.
+  [#5624](https://github.com/matrix-org/matrix-rust-sdk/pull/5624/)
+
 ## [0.16.0] - 2025-12-04
 
 ### Features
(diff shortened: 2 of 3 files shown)


poljar:
This depends on #5617, right?

Hywan:
There are orthogonal. #5617 is adding a new feature to the new Latest Event API, while this set of patches removes the old API.

The last patch of this set introduce a renaming by removing the `new_` prefix. It can conflict with #5617, but it's not a big deal.

Hywan:
#5617 can be merged after this PR, it won't add any regression. Right now, with the old Latest Event API was not supporting local events, so it was impossible to sort by them.

However, https://github.com/matrix-org/matrix-rust-sdk/issues/3872 must be addressed so that re-decryption happens in the Event Cache. The old Latest Event API has its own re-decryption mechanism. Knowing that the new Latest Event API relies on the Event Cache, as long as the Event Cache doesn't support redecryption, we will have a regression.

So:

- https://github.com/matrix-org/matrix-rust-sdk/issues/3872 must be addressed before merging this PR,
- https://github.com/matrix-org/matrix-rust-sdk/pull/5617 can be merged at anytime.

poljar:
> #5617 can be merged after this PR, it won't add any regression. Right now, with the old Latest Event API was not supporting local events, so it was impossible to sort by them.

Ah, because https://github.com/matrix-org/matrix-rust-sdk/pull/5587 added the new way to do latest events, and #5617 will only do the sorting based on the latest event.

Right, sorry for being confused.

kathleenmotley49-maker (approved):

Hywan:
Rebased on top of `main`. Ready to be reviewed!

crates/matrix-sdk-base/src/room/room_info.rs
```diff
@@ -464,16 +462,9 @@ pub struct RoomInfo {
...
-    ///
-    /// TODO(@hywan): Remove.
-    pub(crate) latest_event: Option<Box<LatestEvent>>,
-
     /// The latest event value of this room.
-    ///
-    /// TODO(@hywan): Rename to `latest_event`.
     #[serde(default)]
-    pub(crate) new_latest_event: LatestEventValue,
+    pub(crate) latest_event_value: LatestEventValue,
```
poljar, line 467: Won't this cause some values to be lost? Don't you need a serde alias for this rename to be a no-op?

poljar (changes requested):
Nice, though I think we don't want to just drop the old values because a field got renamed.

Also, I think that this warrants a changelog entry.

crates/matrix-sdk-base/src/room/room_info.rs
Hywan: Is it annoying? This feature hasn't been part of the stable version of most clients. At worse, we are just losing a “cached” data, which is easily computed. My question is: do we need to maintain compatibility in this case?

poljar: Didn't the EX clients switch to this API? Won't they lose the latest event in that case?

If not, feel free to leave the alias out.

Hywan: We agreed in our internal chat room that it's fine to drop. Only “nightly” or “beta” versions of Element X are using it, and no other clients seem to have switched to it so far. Moreover, Element X is using the `LatestEventValue` only recently in the UI: prior to that, it was only used by the Room List to sort items, but it was “invisible” to the app.

Hywan:
I've resolved the merge conflict. All good now.

poljar (approved):
Very nice -3k. Thanks.

PR #5625, opened 2025-09-04
widget-driver: Add read/send capabilities for rtc decline event

BillCarsonFr:
Add capabilities for widgets to read/send rtc decline events.

Signed-off-by:

### DIFF
--- a/bindings/matrix-sdk-ffi/src/widget.rs
+++ b/bindings/matrix-sdk-ffi/src/widget.rs
@@ -175,6 +175,13 @@ pub fn get_element_call_required_permissions(
         WidgetEventFilter::MessageLikeWithType {
             event_type: MessageLikeEventType::RoomRedaction.to_string(),
         },
+        // This allows declining an incoming call and detect if someone declines a call.
+        // Accepts both `org.matrix.msc4310.rtc.decline` and `m.rtc.decline` events to ease future
+        // transition.
+        WidgetEventFilter::MessageLikeWithType {
+            event_type: "org.matrix.msc4310.rtc.decline".to_owned(),
+        },
+        WidgetEventFilter::MessageLikeWithType { event_type: "m.rtc.decline".to_owned() },
     ];
 
     WidgetCapabilities {
--- a/bindings/matrix-sdk-ffi/src/widget.rs
+++ b/bindings/matrix-sdk-ffi/src/widget.rs
@@ -535,5 +535,11 @@ mod tests {
         );
         cap_assert("org.matrix.msc2762.send.event:org.matrix.rageshake_request");
         cap_assert("org.matrix.msc2762.send.event:io.element.call.encryption_keys");
+
+        // RTC decline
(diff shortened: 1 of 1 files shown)


bindings/matrix-sdk-ffi/src/widget.rs
```diff
@@ -175,6 +175,13 @@ pub fn get_element_call_required_permissions(
         WidgetEventFilter::MessageLikeWithType {
             event_type: MessageLikeEventType::RoomRedaction.to_string(),
         },
+        // This allows declining an incoming call and detect if someone declines a call.
+        // Accepts both `org.matrix.msc4310.rtc.decline` and `m.rtc.decline` events to ease future
+        // transition.
+        WidgetEventFilter::MessageLikeWithType {
+            event_type: "org.matrix.msc4310.rtc.decline".to_owned(),
+        },
```
Hywan, line 183: One note about the bindings.

The FFI bindings are supposed to a tiny layer between the foreign language and the Rust SDK. I'm not sure it's a good practise to get “business code” in it. And I think this is what we have here. Thoughts?

Do you think we should refactor that to move this logic from the bindings to the SDK instead?

Hywan (approved):
Looks good.

One note about the fact the bindings should not contain real implementation, but should rather be a thin “glue layer” between the foreign language and the SDK.

bindings/matrix-sdk-ffi/src/widget.rs
BillCarsonFr: I agree, it is a bit strange to have this here.
I was expecting the app to request for permissions. Eventually the SDK could have a list of "auto-granted" permissions that the SDK give without prompting the user (like web is doing?).
So if we do a refactoring I would move that up to the apps (not the SDK), or maybe to a separate crate for calls that the apps could use

BillCarsonFr: I will discuss with the team and see what we can do about that

Hywan: I'm okay to validate that here and now if that's urgent for you, but we must open an issue to track the problem and to not forget to solve it. Otherwise, we could include the refactoring in this PR.

toger5: For context: The decision to move this into the ffi was because of a push back agains element call specific code in the rust-sdk's main crate. This way we limit it to just be available in uniffi sdk's but not when using the rust sdk as a dependency directly.
@BillCarsonFr proposal of having the list of permissions in EX ios+android is probably the "cleanest". It was decided agains however, because of obvious code duplication and any change like this PR would require two PR's.

poljar:
Can we merge this if it has been approved?

BillCarsonFr:
> Can we merge this if it has been approved?

yes please, I created the issue to track it https://github.com/matrix-org/matrix-rust-sdk/issues/5646 as suggested by @Hywan

PR #5626, opened 2025-09-04
Release prep for the 0.14.0 release

poljar:


poljar:
The regressed benchmark seems to be a noisy one. Previously a regression was reported in the TLS crate we're using, now inside of tokio. Ignoring this for now.

poljar:
There's something funky going on with the caches for the benchmarks. First it claims that it can't restore a cache, then it claims that the cache exists when it tries to store it.

I cancelled the last benchmark for that reason.

PR #5629, opened 2025-09-04
feat(multiverse): Show messages in search results.

dragonfly1033:
This PR changes search in multiverse to show the sender and message in the results instead of just event id's.

image

### DIFF
--- a/labs/multiverse/src/main.rs
+++ b/labs/multiverse/src/main.rs
@@ -20,13 +20,16 @@ use futures_util::{StreamExt as _, pin_mut};
 use imbl::Vector;
 use layout::Flex;
 use matrix_sdk::{
-    AuthSession, Client, SearchIndexStoreKind, SqliteCryptoStore, SqliteEventCacheStore,
+    AuthSession, Client, Room, SearchIndexStoreKind, SqliteCryptoStore, SqliteEventCacheStore,
     SqliteStateStore, ThreadingSupport,
     authentication::matrix::MatrixSession,
     config::StoreConfig,
+    deserialized_responses::TimelineEvent,
     encryption::{BackupDownloadStrategy, EncryptionSettings},
     reqwest::Url,
-    ruma::{OwnedRoomId, api::client::room::create_room::v3::Request as CreateRoomRequest},
+    ruma::{
+        OwnedEventId, OwnedRoomId, api::client::room::create_room::v3::Request as CreateRoomRequest,
+    },
 };
 use matrix_sdk_common::locks::Mutex;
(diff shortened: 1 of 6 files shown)


labs/multiverse/src/main.rs
```diff
@@ -465,7 +465,29 @@ impl App {
...
+                                                                )
+                                                            }),
+                                                        )
+                                                        .await
+                                                    {
+                                                        view.results(
+                                                            results.into_iter().flatten().collect(),
+                                                        );
+                                                    } else {
+                                                        debug!("Couldnt get search result events.")
```
poljar: Don't we want to attempt to fetch the event from the server in this case? I guess we assume the event isn't in the index if it isn't in the cache?

If we do have this assumption, let's add a comment explaining this.

labs/multiverse/src/widgets/search/mod.rs
```diff
@@ -40,29 +63,96 @@ impl SearchingView {
...
+    }
+    None
+}
+
+#[derive(Debug, Clone)]
+pub struct MessageWidget {
+    title: String,
+    content: String,
+    style: Style,
+}
```
poljar: Let's add the timestamp of the message to this. It's very useful for the case when we have multiple messages with the same content.

poljar (changes requested):
This kinda works, but it's missing some info like the timestamp.

The search popout also has some problems:

image

It looks a bit weird like it's bleeding into the existing UI elements instead of popping in front of them.

Adding a dedicated background color might help as well as a border. The help screen for example doesn't define a background color but it doesn't do so uniformly.

The search popout in some parts has a background color defined while in the unused parts does not.

There's also the problem that I can't input upercase letters into the search field. As a final goodie that would be nice to have, I can't delete the text in the search field using `CTRL+W` which is possible in the message input field.

labs/multiverse/src/main.rs
dragonfly1033: Yep, although whenever we remove something from the cache should we not rmove it from the index? Anyway, for now I have added the server as backup.

labs/multiverse/src/widgets/search/mod.rs
dragonfly1033: Sure, done :+1:

labs/multiverse/src/main.rs
poljar: Well kinda, we have this method that clears the cache and I'm not certain it has been hooked up to remove the index as well.

poljar (approved):
Alright, I think that's good enough.

Issue #2822, opened 2023-11-06
Sync service: provide more precise errors


bnjbvr:
Notably, it would be nice to distinguish a sliding sync reset, which is not really a hard error, from other timeouts, DNS errors, and so on.

pixlwave:
Some more useful errors would be when the sync proxy is unreachable and when the endpoint doesn't exist at the given host - specifically this would be enough to know that we should alert new users that there's a configuration issue with their well-known not pointing to the right host.

morguldir:
Someone in #element-x-ios was having issues only being able to log in with their subdomain and not their base domain
```

**Comment by morguldir** [comment]
Someone in #element-x-ios was having issues with logging into their base domain (no rooms loading), but the subdomain worked fine, this was because their subdomain had a well-known pointing at the proxy `matrix.example.org/sliding-sync`, but their domain was pointing at `matrix.example.org`, which is not a proxy, but synapse:
```json
{"errcode":"M_UNRECOGNIZED","error":"Unrecognized request"}
```

I've also seen quite a few people run into this when they try to set up the proxy via nginx, and they put the synapse location first, so the sync requests don't go to the proxy (/client might still work)

I'm not sure what a good way to avoid this would be, but perhaps a warning if the endpoint repeatedly 404's?

Hywan:
The log should display the request and the response of the HTTP payloads. I'm not sure it's the role of `matrix_sdk_ui::SyncService` to display these errors. At worst, it should be `matrix_sdk::SlidingSync` itself, but again, it uses the `matrix_sdk::Client`, which logs everything. Is there an access to the logs?

pixlwave:
>  I'm not sure it's the role of `matrix_sdk_ui::SyncService` to display these errors.

The service doesn't need to "display" the errors, but without communicating the error to the app, we have no way of knowing what is the most sensible course for recovery, so for EX we blindly restart not knowing if that will fix the issue or not.

pixlwave:
I'm going to bang on this drum louder given the current matrix.org outage. Element X doesn't know why the sync failed (even though Rust is logging that the "Server is unavailable") so has no other choice but to restart the sync loop and hope for the best.

Hywan:
Louder drum heard. On it Sir.

Hywan:
[...]

PR #5631, opened 2025-09-05
feat(ui): `sync_service::TerminationReport` and `State` contain the error

Hywan:
This set of patches makes `TerminationReport` and `State::Error` to contain the error leading to this state.

So far, this is not yet mapped to new error type in FFI land because it's unclear what we want for the moment. However, things are here, ready to be used, for any other Rust consumers.

It's best to review this set patch after patch.

---

* Fixes https://github.com/matrix-org/matrix-rust-sdk/issues/2822#issuecomment-3248280281


andybalaam (approved):
Great, thanks!

PR #5640, opened 2025-09-08
feat(ui): add custom events to timeline when explicitly filtered

### LINKED ISSUE #5598: Use custom events in matrix_sdk_ui's Timeline
Hi, 

The title resumes it well : I'd want to use a Custom event in the `Timeline` struct exposed by matrix_sdk_ui. Right now it seems that the only possible option to handle custom events is to register an event handler with `client.add_event_handler`, which is, imho, not flexible enough.

Right now I did this : 

## Create the custom event
```rust
 #[derive(Clone, Debug, Deserialize, Serialize, EventContent)]
#[ruma_event(type = "rs.matrix-sdk.example.ack", kind = MessageLike)]
pub struct AckEventContent {
    // the event ID of the ping.
    ping_id: OwnedEventId,
}
```

## Build my timeline with custom filters

```rust
room.timeline_builder()
                .event_filter(|event, room_version| {
                    if event.event_type() == TimelineEventType::from("rs.matrix-sdk.example.ack") {
                        println!("Custom sync event found !");
                        return true;
                    } else {
                        return matrix_sdk_ui::timeline::default_event_filter(event, room_version);
                    };
                })
                .track_read_marker_and_receipts()
                .build()
                .await?
```

## Send a custom event to my room

```rust
let message = AckEventContent::new("hey".to_string());
timeline.room().send(message).await?
```

## Try to read custom events

When I'm processing the diff stream that I got from `timeline.subscribe()`, I cannot find any of my custom event. However, the sync event is correctly sent because my terminal prints "Custom sync event found !". The problem lays between those two layers. 
**Thus, I suppose the conversion between the sync event to a timeline event isn't supported**. 

## Possible solutions ? 

Unfortunately I couldn't find any example of how to manage this use case. The [example provided](https://github.com/matrix-org/matrix-rust-sdk/blob/main/examples/custom_events/src/main.rs) in this repo is mostly dedicated to basic use cases with bots, not for using custom events in the timeline. 

Am I obliged to fork the sdk to add my custom events ? Or is there a way to forward them in the timeline as MsgLike events ? 

Thank you in advance,

IT-ess:
closes #5598 

This allows custom message like events (created by the `EventContent` macro from ruma) to be added to the timeline if they are explictly allowed when building the timeline with an `event_filter`. 
The custom event content is not available directly to the consumer, but it can still fetch it from the matrix-sdk client with its event_id, or display a "this type of event is not supported". 

Signed-off-by: Itess <me@aloiseau.com>

### DIFF
--- a/crates/matrix-sdk-ui/src/timeline/event_item/content/other.rs
+++ b/crates/matrix-sdk-ui/src/timeline/event_item/content/other.rs
@@ -15,8 +15,6 @@
 //! Timeline item content for other message-like events created by the
 //! EventContent macro from ruma.
 
-use std::fmt;
-
 use ruma::events::MessageLikeEventType;
 
 /// A custom event created by the EventContent macro from ruma.


bindings/matrix-sdk-ffi/src/event.rs
```diff
@@ -382,6 +382,54 @@ impl From<MessageLikeEventType> for ruma::events::MessageLikeEventType {
     }
 }
 
+impl TryFrom<ruma::events::MessageLikeEventType> for MessageLikeEventType {
+    type Error = ClientError;
+
+    fn try_from(val: ruma::events::MessageLikeEventType) -> Result<Self, Self::Error> {
+        match val {
+            ruma::events::MessageLikeEventType::CallAnswer => Ok(Self::CallAnswer),
```
bnjbvr: Style suggestion: consider adding `use ruma::events::MessageLikeEventType as RumaEventType;` at the top of this function's body, so we don't have the fully-qualified path on every single line here?

bindings/matrix-sdk-ffi/src/timeline/msg_like.rs
```diff
@@ -182,6 +186,19 @@ impl TryFrom<matrix_sdk_ui::timeline::MsgLikeContent> for MsgLikeContent {
...
                 thread_summary,
             },
+            Kind::Other(other) => Self {
+                kind: MsgLikeKind::Other {
+                    // This should always fail since we cannot construct a custom
+                    // MessageLikeEventType, because the enum must be
+                    // exhaustive.
+                    event_type: MessageLikeEventType::try_from(other.event_type().clone())
+                        .map_err(|e| (e, other.event_type().to_string()))?,
+                },
```
bnjbvr: I'm not sure to follow; does it mean that *all* FFI consumers will run into an error here? If so, I think we should change our strategy here. Maybe we could add a `Custom` (or `Other`) variant to our own `MessageLikeEventType`, so that it's handled properly, and FFI embedders can properly consume it, instead?

(If we did this, I think we could even cover all variants from ruma's `MessageLikeEventType`, and as such the `TryFrom` could be changed to a `From` impl)

crates/matrix-sdk-ui/src/timeline/event_handler.rs
```diff
@@ -381,12 +382,18 @@ impl TimelineAction {
...
             },
 
-            _ => {
-                debug!(
-                    "Ignoring message-like event of type `{}`, not supported (yet)",
-                    content.event_type()
-                );
-                return None;
+            _event => {
+                let other = OtherMessageLike { event_type: _event.event_type() };
```
bnjbvr: nit: no need to use the `_` prefix, since the event is being used here.

crates/matrix-sdk-ui/src/timeline/event_item/content/msg_like.rs
```diff
@@ -74,6 +79,7 @@ impl MsgLikeContent {
             MsgLikeKind::Poll(_) => "a poll",
             MsgLikeKind::Redacted => "a redacted message",
             MsgLikeKind::UnableToDecrypt(_) => "an encrypted message we couldn't decrypt",
+            MsgLikeKind::Other(_) => "a custom message",
```
bnjbvr:
```suggestion
            MsgLikeKind::Other(_) => "a custom message-like event",
```

crates/matrix-sdk-ui/src/timeline/tests/event_filter.rs
```diff
@@ -137,6 +137,25 @@ async fn test_custom_filter() {
...
+            event_filter: Arc::new(|ev, _| matches!(ev, AnySyncTimelineEvent::MessageLike(_))),
+            ..Default::default()
+        })
+        .build();
+    let mut stream = timeline.subscribe().await;
+
+    let f = &timeline.factory;
+    timeline.handle_live_event(f.custom_message_like_event().sender(&ALICE)).await;
+    let _item = assert_next_matches!(stream, VectorDiff::PushBack { value } => value);
+    let _date_divider = assert_next_matches!(stream, VectorDiff::PushFront { value } => value);
```
bnjbvr: Can you do additional checks on the item and date_divider? i.e., assert that the item is from the expected event type, and that the date divider is a date divider.

crates/matrix-sdk-ui/tests/integration/timeline/mod.rs
```diff
@@ -784,6 +788,51 @@ async fn test_timeline_receives_a_limited_number_of_events_when_subscribing() {
...
+            JoinedRoomBuilder::new(room_id).add_timeline_event(
+                f.custom_message_like_event().event_id(event_id).sender(user_id!("@a:b.c")),
+            ),
+        )
+        .await;
+
+    assert_let!(Some(timeline_updates) = timeline_stream.next().await);
+    assert_let!(VectorDiff::PushBack { value: first } = &timeline_updates[0]);
+    let event_type = MessageLikeEventType::from("rs.matrix-sdk.custom.test");
+    let _other_msglike = OtherMessageLike::from_event_type(event_type);
```
bnjbvr: nit: the variable is being used, so don't use the `_` prefix

crates/matrix-sdk-ui/tests/integration/timeline/mod.rs
```diff
@@ -784,6 +788,51 @@ async fn test_timeline_receives_a_limited_number_of_events_when_subscribing() {
...
+    server.mock_room_state_encryption().plain().mount().await;
+
+    let timeline = room
+        .timeline_builder()
+        .event_filter(|event, room_version| {
+            if event.event_type() == TimelineEventType::from("rs.matrix-sdk.custom.test") {
+                true
+            } else {
+                default_event_filter(event, room_version)
+            }
```
bnjbvr:
```suggestion
            event.event_type() == TimelineEventType::from("rs.matrix-sdk.custom.test") || default_event_filter(event, room_version)
```

(This may need rustfmt'ing)

testing/matrix-sdk-test/src/event_factory.rs
```diff
@@ -1264,3 +1270,15 @@ impl From<MembershipState> for PreviousMembership {
         Self::new(state)
     }
 }
+
+#[derive(Clone, Debug, Serialize, EventContent)]
+#[ruma_event(type = "rs.matrix-sdk.custom.test", kind = MessageLike)]
+pub struct CustomMessageLikeEventContent {
+    whatever_field: String,
```
bnjbvr: Since the field isn't used, even in testing, can you remove it? It's fine to use an empty event content, for testing purposes.

bindings/matrix-sdk-ffi/src/timeline/msg_like.rs
IT-ess: Ok I will do that, I wasn't sure if this was still the scope of this PR, but let's do this !

crates/matrix-sdk-ui/tests/integration/timeline/mod.rs
IT-ess: Rust compiler doesn't agree :thinking: 
image

bindings/matrix-sdk-ffi/src/timeline/msg_like.rs
IT-ess: In fact it was fairly easy, we just store the event_type in the ffi "Other" variant :sweat_smile:

crates/matrix-sdk-ui/tests/integration/timeline/mod.rs
bnjbvr: Ah, so this is a different issue: the `other_msglike` in the `matches!` statement will in fact bind a new variable named `other_msglike`, that's shadowing the previous binding defined in the `let` statement.

In this case, I think you could use `assert_matches!`, which supports adding bindings, and a block for extra assertions, so something like (I'm writing this from memory, so not sure it's the exact correct syntax):

```rust
let other_msglike = OtherMessageLike::from_event_type(event_type);
assert_matches!(
  &first.as_event().unwrap().content().as_msglike().unwrap().kind,
  MsgLikeKind::Other(observed_other) => {
    assert_eq!(observed_other, other_msglike);
});
```

bindings/matrix-sdk-ffi/src/timeline/msg_like.rs
```diff
@@ -182,6 +186,18 @@ impl TryFrom<matrix_sdk_ui::timeline::MsgLikeContent> for MsgLikeContent {
                 thread_root,
                 thread_summary,
             },
+            Kind::Other(other) => Self {
+                kind: MsgLikeKind::Other {
+                    // This should always fail since we cannot construct a custom
+                    // MessageLikeEventType, because the enum must be
+                    // exhaustive.
```
bnjbvr: Comment can now be removed, i think?

crates/matrix-sdk-ui/src/timeline/tests/event_filter.rs
```diff
@@ -137,6 +137,30 @@ async fn test_custom_filter() {
...
+
+    let f = &timeline.factory;
+    timeline.handle_live_event(f.custom_message_like_event().sender(&ALICE)).await;
+    let item = assert_next_matches!(stream, VectorDiff::PushBack { value } => value);
+    let date_divider = assert_next_matches!(stream, VectorDiff::PushFront { value } => value);
+
+    let _item = item.as_event().unwrap().content.as_msglike().unwrap().kind.clone();
+    let _date_divider = date_divider.as_virtual().unwrap();
+    assert!(matches!(MsgLikeKind::Other, _item));
+    assert!(matches!(VirtualTimelineItem::ReadMarker, _date_divider));
```
bnjbvr: These two items are being used; i think this code suffers from the same problem I've described in the other comment, where the second `_item` binding shadows the first one, so it's not doing what you think it's doing (it's not comparing the two versions, but instead it does nothing - so Rust is correct that the two variables are unused). Can you use `assert_matches` here and add `assert_eq` assertions too, please?

testing/matrix-sdk-test/src/event_factory.rs
```diff
@@ -1297,3 +1303,19 @@ impl From<MembershipState> for PreviousMembership {
...
+    pub fn new() -> Self {
+        Self {}
+    }
+}
+
+impl Default for CustomMessageLikeEventContent {
+    fn default() -> Self {
+        Self::new()
+    }
+}
```
bnjbvr: conciseness ftw

```suggestion
#[derive(Clone, Default, Debug, Serialize, EventContent)]
#[ruma_event(type = "rs.matrix-sdk.custom.test", kind = MessageLike)]
pub struct CustomMessageLikeEventContent;
```

testing/matrix-sdk-test/src/event_factory.rs
```diff
@@ -1076,6 +1077,11 @@ impl EventFactory {
         event
     }
 
+    /// Create a new `rs.matrix-sdk.custom.test` custom event
+    pub fn custom_message_like_event(&self) -> EventBuilder<CustomMessageLikeEventContent> {
+        self.event(CustomMessageLikeEventContent::new())
```
bnjbvr: (with the suggestion below)


```suggestion
        self.event(CustomMessageLikeEventContent::default())
```

crates/matrix-sdk-ui/tests/integration/timeline/mod.rs
IT-ess: Oops, I'm still quite a neophyte with Rust :sweat_smile:

IT-ess: I had to derive the PartialEq trait on the `OtherMessageLike` struct for this to work. Is that ok or should I gate it behind a `#[cfg(test)]` ?

bnjbvr: sounds fine to me :+1:

crates/matrix-sdk-ui/src/timeline/event_item/content/other.rs
```diff
@@ -0,0 +1,45 @@
...
+    }
+}
+
+#[cfg(not(tarpaulin_include))]
+impl fmt::Debug for OtherMessageLike {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        let Self { event_type } = self;
+        f.debug_struct("OtherMessageLike").field("event_type", event_type).finish()
+    }
+}
```
bnjbvr: Actually that's the same as an automatically derived `Debug` impl, since `MessageLikeEventType` implements `Debug`, so we could use that instead of a custom Debug impl.

bnjbvr (approved):
Nice! I might amend the PR myself to take care of the last comment. Thanks for adding this!

IT-ess:
Thank you for the merge and the review ! 
This will make the event_filter a "real" filter which doesn't ignore custom events 😄

PR #5641, opened 2025-09-09
feat: Add Element recent emojis for shared emoji reactions

jmartinesp:
This is part of the results of the HHCIB hackathon, equivalent - kind of - to [this functionality in Element Web](https://github.com/element-hq/element-web/commit/a7f92f35f5a27a53a5a030ea7c471be97751a67a). It's not based on an MSC, as I explicitly asked the Web team what the plan was for this when it was merged and they said we only wanted to make it work on Element clients, so there was no MSC involved or planned 🤷 .

With this we can sync recently used emojis in Web and Element X. The list is sorted by recency of usage and also contains the number of times the emoji has been used in case we want to use that to filter out the items in the future.

This is behind a new `experimental-element-recent-emojis` feature, that's enabled at the FFI layer.

Signed-off-by:

### DIFF
--- a/crates/matrix-sdk/src/account.rs
+++ b/crates/matrix-sdk/src/account.rs
@@ -17,6 +17,8 @@
 use futures_core::Stream;
 use futures_util::{stream, StreamExt};
 #[cfg(feature = "experimental-element-recent-emojis")]
+use itertools::Itertools;
+#[cfg(feature = "experimental-element-recent-emojis")]
 use js_int::uint;
 #[cfg(feature = "experimental-element-recent-emojis")]
 use matrix_sdk_base::recent_emojis::RecentEmojisContent;
@@ -1206,7 +1208,7 @@ impl Account {
     /// homeserver instead of the local storage.
     #[cfg(feature = "experimental-element-recent-emojis")]
     pub async fn get_recent_emojis(&self, refresh: bool) -> Result<Vec<(String, UInt)>> {
-        let event = if refresh {
+        let content = if refresh {
             let Some(user_id) = self.client.user_id() else {
                 return Err(Error::AuthenticationRequired);
             };
@@ -1229,8 +1231,17 @@ impl Account {
                 .transpose()?
         };
 
-        if let Some(content) = event {
-            Ok(content.recent_emoji.into_iter().collect())
+        if let Some(content) = content {
+            // Sort by count, descending. For items with the same count, since they were
+            // previously ordered by recency in the list, more recent emojis will be
+            // returned first.
+            let sorted_emojis = content
+                .recent_emoji
+                .into_iter()
+                // Items with higher counts should be first
+                .sorted_by(|(_, count_a), (_, count_b)| count_b.cmp(count_a))
+                .collect();
+            Ok(sorted_emojis)
         } else {
             Ok(Vec::new())
         }
@@ -1297,9 +1308,19 @@ mod test_recent_emojis {
         let recent_emojis = client.account().get_recent_emojis(false).await.expect("recent emojis");
         assert!(recent_emojis.is_empty());
 
+        let emoji_list = vec![
+            (":/".to_owned(), uint!(1)),
+            (":)".to_owned(), uint!(12)),
+            (":D".to_owned(), uint!(12)),
+        ];
+
         server
             .mock_global_account_data()
-            .ok(user_id, RecentEmojisContent::default().event_type(), json!({ "recent_emoji": [] }))
+            .ok(
+                user_id,
+                RecentEmojisContent::default().event_type(),
+                json!({ "recent_emoji": emoji_list }),
+            )
             .named("Fetch recent emojis")
             .mock_once()
             .mount()
@@ -1310,7 +1331,7 @@ mod test_recent_emojis {
         server
             .mock_sync()
             .ok(|builder| {
-                let content = RecentEmojisContent::new(vec![(":)".to_owned(), uint!(1))]);
+                let content = RecentEmojisContent::new(emoji_list);
                 let event_builder = EventFactory::new().global_account_data(content);
                 builder.add_global_account_data(event_builder);
             })
@@ -1321,6 +1342,13 @@ mod test_recent_emojis {
         client.sync_once(SyncSettings::default()).await.expect("sync failed");
 
         let recent_emojis = client.account().get_recent_emojis(false).await.expect("recent emojis");
-        assert_eq!(recent_emojis.len(), 1);
+
+        // Assert size
+        assert_eq!(recent_emojis.len(), 3);
+
+        // Assert ordering: first by times used, then by recency
+        assert_eq!(recent_emojis[0].0, ":)");
+        assert_eq!(recent_emojis[1].0, ":D");
+        assert_eq!(recent_emojis[2].0, ":/");
     }
 }


Hywan:
Thanks @jmartinesp!

I'm questioning myself: Should we accept features not based on an MSC? cc @poljar

crates/matrix-sdk-base/src/lib.rs
```diff
@@ -45,6 +45,9 @@ pub mod sync;
 mod test_utils;
 mod utils;
 
+#[cfg(feature = "element-recent-emojis")]
```
Hywan: Should be `experimental-element-recent-emojis` I believe.

poljar:
> Thanks @jmartinesp!
> 
> I'm questioning myself: Should we accept features not based on an MSC? cc @poljar

We discussed this before @jmartinesp opened the PR. Yes, this is simple enough that it's fine to accept without a MSC.

crates/matrix-sdk-base/src/recent_emojis.rs
```diff
@@ -0,0 +1,25 @@
+//! Data types used for handling the recently used emojis.
+
+use ruma::{UInt, events::macros::EventContent};
+use serde::{Deserialize, Serialize};
+
+#[cfg(feature = "element-recent-emojis")]
+#[derive(Clone, Debug, Default, Deserialize, Serialize, EventContent)]
+#[ruma_event(type = "io.element.recent_emoji", kind = GlobalAccountData)]
+/// An event type containing a list of recently used emojis for reactions.
```
poljar: Docs go on top of all the derives an other attributes.

crates/matrix-sdk-base/src/recent_emojis.rs
```diff
@@ -0,0 +1,25 @@
+//! Data types used for handling the recently used emojis.
```
poljar, line 1: Is this documented anywhere at all? If it is could we link to such a document?

crates/matrix-sdk/src/client/mod.rs
```diff
@@ -2938,6 +2944,50 @@ impl Client {
...
+    #[cfg(feature = "element-recent-emojis")]
+    /// Gets the list of recently used emojis from the `io.element.recent_emoji`
+    /// global account data.
+    pub async fn get_recent_emojis(&self) -> Result<Vec<(String, u64)>> {
+        let Some(content) = self.get_recent_emoji_content().await? else {
+            return Ok(Vec::new());
+        };
+
+        Ok(content.recent_emoji.into_iter().map(|(emoji, count)| (emoji, count.into())).collect())
+    }
```
poljar: Since all of this uses account data, we should move these methods to the `account.rs` file and under the `Account` struct.

crates/matrix-sdk/src/client/mod.rs
```diff
@@ -2938,6 +2944,50 @@ impl Client {
...
+        let request = UpdateGlobalAccountDataRequest::new(session_id.to_owned(), &content)?;
+        let _ = self.send(request).await?;
+
+        Ok(())
+    }
+
+    #[cfg(feature = "element-recent-emojis")]
+    /// Gets the list of recently used emojis from the `io.element.recent_emoji`
+    /// global account data.
+    pub async fn get_recent_emojis(&self) -> Result<Vec<(String, u64)>> {
```
poljar: Again, the cfg goes below the docs.

crates/matrix-sdk/src/client/mod.rs
```diff
@@ -2938,6 +2944,50 @@ impl Client {
...
+            .get_account_data_event_static::<RecentEmojisContent>()
+            .await?
+            .map(|raw| raw.deserialize().map(|event| event.content))
+            .transpose()
+            .map_err(Into::into)
+    }
+
+    #[cfg(feature = "element-recent-emojis")]
+    /// Adds a recently used emoji to the list and uploads the updated
+    /// `io.element.recent_emoji` content to the global account data.
```
poljar: Docs on top, cfg attribute below.

testing/matrix-sdk-test/src/event_factory.rs
```diff
@@ -1076,6 +1076,15 @@ impl EventFactory {
...
     }
+
+    pub fn global_account_data<C>(&self, content: C) -> EventBuilder<C>
+    where
+        C: GlobalAccountDataEventContent + StaticEventContent<IsPrefix = False>,
+    {
+        let mut event = self.event(content);
+        event.is_global = true;
+        event
+    }
```
poljar, line 1088: Docs please.

crates/matrix-sdk/src/client/mod.rs
```diff
@@ -2938,6 +2944,50 @@ impl Client {
...
+    pub async fn add_recent_emoji(&self, emoji: &str) -> Result<()> {
+        let mut content =
+            self.get_recent_emoji_content().await?.unwrap_or_else(RecentEmojisContent::default);
```
poljar: This is a common pattern in the account data handling which tends to delete your account data because we might be operating on stale data.

Are you sue you don't want to get the account data from the server to avoid this?

crates/matrix-sdk-base/src/recent_emojis.rs
```diff
@@ -0,0 +1,25 @@
...
+#[cfg(feature = "experimental-element-recent-emojis")]
+#[derive(Clone, Debug, Default, Deserialize, Serialize, EventContent)]
+#[ruma_event(type = "io.element.recent_emoji", kind = GlobalAccountData)]
+/// An event type containing a list of recently used emojis for reactions.
+pub struct RecentEmojisContent {
+    /// The list of recently used emojis, ordered by recency. The tuple of
+    /// `String`, `UInt` values represent the actual emoji and the number of
+    /// times it's been used in total, for those clients that might be
+    /// interested.
+    pub recent_emoji: Vec<(String, UInt)>,
```
poljar, line 18: You sure that this is the correct type to model this?

It seems to me that a `BTreeMap<String, UInt>` would be the more natural type for this.

crates/matrix-sdk-base/src/recent_emojis.rs
```diff
@@ -0,0 +1,25 @@
+//! Data types used for handling the recently used emojis.
+
+use ruma::{UInt, events::macros::EventContent};
+use serde::{Deserialize, Serialize};
+
+#[cfg(feature = "experimental-element-recent-emojis")]
+#[derive(Clone, Debug, Default, Deserialize, Serialize, EventContent)]
```
poljar, line 11: Can we have a test for the deserialization/serialization of this type?

Otherwise someomen might tweak the type and break things. An insta snapshot might be nice as well.

crates/matrix-sdk/src/client/mod.rs
```diff
@@ -2938,6 +2944,50 @@ impl Client {
...
+        let index = content.recent_emoji.iter().position(|(unicode, _)| unicode == emoji);
+
+        let count =
+            if let Some(index) = index { content.recent_emoji.remove(index).1 } else { uint!(0) };
+
+        content.recent_emoji.insert(0, (emoji.to_owned(), count + uint!(1)));
+
+        let Some(session_id) = self.user_id() else {
+            return Err(Error::AuthenticationRequired);
+        };
```
poljar: It would probably make sense to do this at the top of the function, also shouldn't we call the variable a `user_id`?

crates/matrix-sdk/src/client/mod.rs
```diff
@@ -2938,6 +2944,50 @@ impl Client {
...
+
+    #[cfg(feature = "experimental-element-recent-emojis")]
+    /// Gets the list of recently used emojis from the `io.element.recent_emoji`
+    /// global account data.
+    pub async fn get_recent_emojis(&self) -> Result<Vec<(String, u64)>> {
+        let Some(content) = self.get_recent_emoji_content().await? else {
+            return Ok(Vec::new());
+        };
+
+        Ok(content.recent_emoji.into_iter().map(|(emoji, count)| (emoji, count.into())).collect())
```
poljar: Not going to insist on this, but this looks nicer to me if you just use `if let Some() {...} else {...}` avoiding the goto in the form of the early return.

poljar (changes requested):
Looks mostly sensible, there are a bunch of small nits I left.

poljar:
Oh, to get the CI green, a rebase onto latest main should help.

crates/matrix-sdk-base/src/recent_emojis.rs
jmartinesp: Not really: https://github.com/element-hq/element-web/issues/12315

crates/matrix-sdk-base/src/recent_emojis.rs
jmartinesp: Can the `BTreeMap` keep the ordering of items? It's used as a LRU cache, where recent items are removed from whatever index they had before and moved to the start of the vec.

crates/matrix-sdk/src/client/mod.rs
jmartinesp: Since this data comes from `/sync` I was hoping it would be up to date, but I can add the fetch from server call here too, if you think that's better.

crates/matrix-sdk-base/src/recent_emojis.rs
jmartinesp: There is this commit with the implementation: https://github.com/element-hq/element-web/commit/a7f92f35f5a27a53a5a030ea7c471be97751a67a

crates/matrix-sdk-base/src/recent_emojis.rs
poljar: Oh, not sure if that's clever or a bit crazy. In that case a `HashMap` would be needed, as the `BTreeMap` orders entries by the key, in this case alphabetically.

jmartinesp: But HashMap doesn't have any ordering, AFAICT. We need to keep the ordering here, since the most recently used emojis will be removed from their current index, if any, then placed at the start, the format it has in its JSON form is:

```json
"recent_emoji": [["👍️", 2406], ["🙏", 250], ["😂", 112], ["💯", 54], ["🚢", 8], ...]
```

jmartinesp: Actually, looking at the actual EW implementation this ordering is kind of important, but the `count` parameter is the really important one, since they sort the list using it, then keep the list ordering for items with the same count, i.e:

```json
"recent_emoji": [["👍️", 100], ["🙏", 200], ["😂", 100], ["💯", 320], ["🚢", 8]]
```

Will return the items in this order:

```json
["💯", "🙏", "👍️", "😂", "🚢"]
```

 I need to give this another try.

poljar: Ah right. I was thinking of `IndexMap` which preserves the insertion order which you might preserve during deserialization.

bindings/matrix-sdk-ffi/src/client.rs
```diff
@@ -1745,6 +1745,37 @@ impl Client {
...
+    #[cfg(feature = "experimental-element-recent-emojis")]
+    /// Adds a recently used emoji to the list and uploads the updated
+    /// `io.element.recent_emoji` content to the global account data.
+    pub async fn add_recent_emoji(&self, emoji: String) -> Result<(), ClientError> {
+        Ok(self.inner.account().add_recent_emoji(&emoji).await?)
+    }
+
+    #[cfg(feature = "experimental-element-recent-emojis")]
+    /// Gets the list of recently used emojis from the `io.element.recent_emoji`
+    /// global account data.
```
poljar, line 1757: Same as above.

bindings/matrix-sdk-ffi/src/client.rs
```diff
@@ -1745,6 +1745,37 @@ impl Client {
...
+            .await?
+            .into_iter()
+            .map(|(emoji, count)| RecentEmoji { emoji, count: count.into() })
+            .collect::<Vec<RecentEmoji>>())
+    }
+}
+
+#[cfg(feature = "experimental-element-recent-emojis")]
+#[derive(Debug, uniffi::Record)]
+/// Represents an emoji recently used for reactions.
```
poljar: Another one.

crates/matrix-sdk-base/src/recent_emojis.rs
poljar: I guess let's link to the EW PR or commit in that case.

crates/matrix-sdk/src/client/mod.rs
poljar: Yeah, the problem is that people can call this method without syncing and there's a race between a sync and a person calling this method.

I think it's reasonable to fetch from the server at least when we don't find anything in the store.

jmartinesp: I made it so the data is always fetched before being modified so we're sure we're modifying the updated values. Would that work?

poljar: Sure, if you think a network request every time you make a reaction is fine to do.

jmartinesp: It's not ideal, since you'll have 3 requests in total (reaction, fetch recent emojis, update recent emojis), but I think it's the only way to ensure there are no races.

crates/matrix-sdk/src/account.rs
```diff
@@ -1165,6 +1173,79 @@ impl Account {
...
+        let Some(user_id) = self.client.user_id() else {
+            return Err(Error::AuthenticationRequired);
+        };
+        let mut recent_emojis = self.get_recent_emojis(true).await?;
+
+        let index = recent_emojis.iter().position(|(unicode, _)| unicode == emoji);
+
+        let count = if let Some(index) = index { recent_emojis.remove(index).1 } else { uint!(0) };
+
+        recent_emojis.insert(0, (emoji.to_owned(), count + uint!(1)));
```
Johennes, line 1193: This doesn't seem to limit the number of emojis that are stored. The JS SDK implementation starts truncating at a count of 100. Should something similar happen here, too?

jmartinesp: Oh, probably. I will add a parameter to the FFI layer so we can set how many emojis we want to retrieve.

jmartinesp: Just to confirm: do you truncate only when retrieving items, or also when add items so the account data event doesn't grow indefinitely?

Johennes: Oh, I'm only talking about the latter. If I'm reading the JS SDK implementation correctly, it truncates the account data event to 100 emoji when adding new ones to prevent it from growing unboundedly.

jmartinesp: I see, now the implementation makes more sense, because if an emoji was used a lot in the past but it hasn't been used in ages it can still be removed by more recently used emojis.

In that case I can either add a `limit/trauncate_at` parameter to both add and get emojis or I can hardcode the value in the SDK. The latter makes more sense in my mind, to be honest. @poljar what do you think?

poljar: Yeah, I think people shouldn't need to think about this and we should just agree to a sensible default.

PR #5642, opened 2025-09-09
Return a MessageType from verification_request_content

andybalaam:

Signed-off-by:

### DIFF
--- a/crates/matrix-sdk-crypto/src/identities/user.rs
+++ b/crates/matrix-sdk-crypto/src/identities/user.rs
@@ -25,9 +25,7 @@ use as_variant::as_variant;
 use matrix_sdk_common::locks::RwLock;
 use ruma::{
     api::client::keys::upload_signatures::v3::Request as SignatureUploadRequest,
-    events::{
-        key::verification::VerificationMethod, room::message::KeyVerificationRequestEventContent,
-    },
+    events::{key::verification::VerificationMethod, room::message::MessageType},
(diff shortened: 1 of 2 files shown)


richvdh (commented):
looks fine to me, though it could do with a test to demonstrate it's fixing the bug.

richvdh:
I have a working test, which I can share once https://github.com/matrix-org/matrix-rust-sdk/pull/5654 is merged

crates/matrix-sdk-crypto/CHANGELOG.md
```diff
@@ -6,6 +6,11 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased] - ReleaseDate
 
+### Bug Fixes
+
+- Fix a bug introduced in 0.14.0 which meant that the serialization of the value returned by `OtherUserIdentity::verification_request_content` did not include a `msgtype` field.
+  ([#5642](https://github.com/matrix-org/matrix-rust-sdk/pull/5642))
```
richvdh, line 12: does this count as a breaking change? It changes the type of `OtherUserIdentity::verification_request_content`, but OTOH given the intended use is just to shove it into `serde_json::to_value`, maybe that doesn't matter in practice.

andybalaam: Given it fixes clearly-wrong behaviour, I'm guessing it will only unbreak someone's code anyway, so I'd be inclined to say it's fine.

PR #5647, opened 2025-09-09
Create a separate error variant to indicate a failure importing a secret.

uhoreg:
Part of the fix to https://github.com/element-hq/element-x-android/issues/5099

Allows applications to distinguish between errors that occur when unlocking Secret Storage, or errors that occur when importing a secret, so that they can display appropriate feedback (or not) to the user.

Signed-off-by:

### DIFF
--- a/bindings/matrix-sdk-ffi/src/encryption.rs
+++ b/bindings/matrix-sdk-ffi/src/encryption.rs
@@ -79,9 +79,14 @@ pub enum RecoveryError {
     #[error(transparent)]
     Client { source: crate::ClientError },
 
-    /// Error in the secret storage subsystem.
+    /// Error in the secret storage subsystem, except for when importing a
+    /// secret.
     #[error("Error in the secret-storage subsystem: {error_message}")]
     SecretStorage { error_message: String },
+
+    /// Error when importing a secret from secret storage.
+    #[error("Error importing a secret: {error_message}")]
+    Import { error_message: String },
 }
 
 impl From<matrix_sdk::encryption::recovery::RecoveryError> for RecoveryError {
@@ -89,6 +94,9 @@ impl From<matrix_sdk::encryption::recovery::RecoveryError> for RecoveryError {
         match value {
             recovery::RecoveryError::BackupExistsOnServer => Self::BackupExistsOnServer,
             recovery::RecoveryError::Sdk(e) => Self::Client { source: ClientError::from(e) },
+            recovery::RecoveryError::SecretStorage(
+                matrix_sdk::encryption::secret_storage::SecretStorageError::ImportError(e),
+            ) => Self::Import { error_message: e.to_string() },
             recovery::RecoveryError::SecretStorage(e) => {
                 Self::SecretStorage { error_message: e.to_string() }
             }
--- a/crates/matrix-sdk/src/encryption/secret_storage/secret_store.rs
+++ b/crates/matrix-sdk/src/encryption/secret_storage/secret_store.rs
@@ -295,7 +295,8 @@ impl SecretStore {
     }
 
     async fn maybe_enable_backups(&self) -> Result<()> {
-        if let Some(mut secret) = self.get_secret(SecretName::RecoveryKey).await? {
+        let get_secret_result = self.get_secret(SecretName::RecoveryKey).await;
+        if let Ok(Some(mut secret)) = get_secret_result {
             let ret = self.client.encryption().backups().maybe_enable_backups(&secret).await;
 
             if let Err(e) = &ret {
@@ -305,6 +306,10 @@ impl SecretStore {
             secret.zeroize();
 
             Ok(ret.map(|_| ())?)
+        } else if let Err(e) = get_secret_result {
+            warn!("Could not enable backups from secret storage: {e:?}");
+
+            Err(e)
         } else {
             info!("No backup recovery key found.");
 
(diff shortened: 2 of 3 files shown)


uhoreg:
Not sure why the one unit test is failing.  It looks completely unrelated to what I changed, and the test succeeds on my machine.

crates/matrix-sdk/src/encryption/secret_storage/secret_store.rs
```diff
@@ -295,7 +295,8 @@ impl SecretStore {
     }
 
     async fn maybe_enable_backups(&self) -> Result<()> {
-        if let Some(mut secret) = self.get_secret(SecretName::RecoveryKey).await? {
+        let get_secret_result = self.get_secret(SecretName::RecoveryKey).await;
+        if let Ok(Some(mut secret)) = get_secret_result {
```
Hywan: I believe a `match` statement would be better here:

```rust
match self.get_secret(SecretName::RecoveryKey).await {
    Ok(Some(mut secret)) => {
        // …
    }
    Err(e) => {
        warn!("Could not enable backups from secret storage: {e:?}");

        Err(e)
    }
    _ => {
        info!("No backup recovery key found.");

        Ok(())
    }
}
```

crates/matrix-sdk/src/encryption/secret_storage/secret_store.rs
```diff
@@ -151,7 +151,9 @@ impl SecretStore {
         let secret_name = secret_name.into();
         let event_type = GlobalAccountDataEventType::from(secret_name.to_owned());
 
-        if let Some(secret_content) = self.client.account().fetch_account_data(event_type).await? {
+        if let Some(secret_content) =
+            self.client.account().fetch_account_data(event_type).await.map_err(ImportError::from)?
```
Hywan: Should be a fixup of the prefix patch.

bindings/matrix-sdk-ffi/src/encryption.rs
```diff
@@ -79,16 +79,24 @@ pub enum RecoveryError {
...
 
-    /// Error in the secret storage subsystem.
+    /// Error in the secret storage subsystem, except for when importing a
+    /// secret.
     #[error("Error in the secret-storage subsystem: {error_message}")]
     SecretStorage { error_message: String },
+
+    /// Error when importing a secret from secret storage.
+    #[error("Error importing a secret: {error_message}")]
+    Import { error_message: String },
```
poljar, line 89: Sorry for the drive-by comment, but don't we want to remember which secret couldn't have been imported?

uhoreg: It wasn't needed for the issue I was working on, but it does sound like it could be useful.  I guess it means that I can't just use `ImportError::from` (or I can't just use the automatic conversion from `ImportError` to `SecretStorageError`.  Is there some sort of convention that I should follow?

poljar: You won't be able to use the automatic conversion, but you can just `map_err()` from one error to the other and include the context that we can't infer automatically this way.

uhoreg: Sorry, I wasn't clear with my question.  I was wondering if there was a convention for what to put in the `map_err`.  Should I just throw an anonymous function in there?  Or should I pre-define some functions in `ImportError`?  Or something else?

poljar: Depends on how many times you need to do the same map, two or more use a pre-defined function, otherwise anonymous functions are fine.

uhoreg: I've changed SecretStorageError::ImportError to have `name` and `error` fields, where `name` is the name of the secret that was being imported.  It all seems fine *except* that the way `SecretStore.import_secrets` imports the cross signing keys is that it loads all the secrets from Secret Storage (which is fine, it calls `SecretStore.get_secret` individually on each secret name), but then it bundles them all up and calls `OlmMachine.import_cross_signing_keys` on the whole bundle.  In there, it will error, for example, if one of the private keys doesn't match the public key.

But that means that I can't tell which secret caused the error, unless I modify the error type returned by `OlmMachine.import_cross_signing_keys` so that it contains the secret name, which means changing stuff in the crypto crate.  I'm a bit hesitant to go deeper into the Rust SDK for this, since it will probably affect more stuff, but I think it's the only way to get the right information.  So, just wanted to double check that it's OK for me to do that.

I think the alternative would be to hard-code something in `SecretStore.import_secrets` saying that it's, say, the Master key that caused the error, and hope for the best, but I'm not too excited about that either.

crates/matrix-sdk/src/encryption/secret_storage/secret_store.rs
uhoreg: Changed the commit to be a fixup.  Hopefully I did it right.

uhoreg: But the CI is complaining about the fixup, so not sure what's going on. :-/

poljar: That's just a reminder for us to squash the fixup commits.

bindings/matrix-sdk-ffi/src/encryption.rs
poljar: I think it's fine to include that in the crypto crate as well. Sure that's a breaking change but a fairly minor one.

uhoreg:
Ping @Hywan: can you give this a re-review?

uhoreg:
@Hywan or @poljar: one of you will need to hit the merge button because I don't have permissions in this repo

PR #5648, opened 2025-09-10
feat(common): Partially protect against malicious `origin_server_ts`

Hywan:
The first patch adds the `timestamp` field to `TimelineEvent`. It's a copy of the `origin_server_ts` value, parsed as an `Option<MilliSecondsSinceUnixEpoch>`. It's `None` if the parsing failed, or if the `TimelineEvent` was deserialised from a version before this new field was added.

A new `extract_timestamp` function is added for this purpose. It protects against malicious `origin_server_ts` where the value can be set to year 2100 for example. The only protection we are adding here is to take the `min(origin_server_ts, now())`, so that the event can never been “in the future”.

It doesn't protect against a malicious value like 0. It's non-trivial to define a minimum timestamp for an event.

When a `TimelineEvent` is mapped from one kind to another kind, the `timestamp` is carried over. To achieve that, new `to_decrypted` and `to_utd` methods are added.

The second patch updates `LatestEventValue::timestamp` to use the new `TimelineEvent::timestamp` method in case of a `LatestEventValue::Remote`. As such, it fixes the comment in https://github.com/matrix-org/matrix-rust-sdk/pull/5617#discussion_r2325247420.

---

* Follow up of https://github.com/matrix-org/matrix-rust-sdk/pull/5617

### DIFF
--- a/crates/matrix-sdk-base/src/latest_event.rs
+++ b/crates/matrix-sdk-base/src/latest_event.rs
@@ -45,27 +45,20 @@ impl LatestEventValue {
     /// Get the timestamp of the [`LatestEventValue`].
     ///
     /// If it's [`None`], it returns `None`. If it's [`Remote`], it returns the
-    /// `origin_server_ts`. If it's [`LocalIsSending`] or [`LocalCannotBeSent`],
-    /// it returns the [`timestamp`] value.
+    /// [`TimelineEvent::timestamp`]. If it's [`LocalIsSending`] or
+    /// [`LocalCannotBeSent`], it returns the
+    /// [`LocalLatestEventValue::timestamp`] value.
     ///
     /// [`None`]: LatestEventValue::None
     /// [`Remote`]: LatestEventValue::Remote
     /// [`LocalIsSending`]: LatestEventValue::LocalIsSending
     /// [`LocalCannotBeSent`]: LatestEventValue::LocalCannotBeSent
-    /// [`timestamp`]: LocalLatestEventValue::timestamp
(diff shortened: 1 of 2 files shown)


crates/matrix-sdk-common/src/serde_helpers.rs
```diff
@@ -116,17 +116,35 @@ pub fn extract_bundled_thread_summary(
...
+/// from the future.
+pub fn extract_timestamp(
+    event: &Raw<AnySyncTimelineEvent>,
+    max_value: MilliSecondsSinceUnixEpoch,
+) -> Option<MilliSecondsSinceUnixEpoch> {
+    let mut origin_server_ts = event.get_field("origin_server_ts").ok().flatten()?;
+
+    if origin_server_ts > max_value {
+        origin_server_ts = max_value;
+    }
```
stefanceriu, line 132: I would argue the right solution here is to set it to zero/something small and not `max_value` otherwise the it would still probably stay on top of everything.

Hywan: When the event is received via the sync, its origin_server_ts will be capped to now(). It won't change after that. So it won't stay at the top of everything.

It can happen though, for a previously received event with a malicious origin_server_ts. Clearing the event cache is the only solution here.

About _something small_, it's non-trivial to define sadly :-(.

stefanceriu: Ain't that the truth 😞

PR #5651, opened 2025-09-10
chore(sdk): Better logs for the duplicate one-time key error

poljar:


crates/matrix-sdk/src/encryption/mod.rs
```diff
@@ -380,6 +384,66 @@ struct OAuthCrossSigningResetUiaaResetParameter {
...
+    fn from_str(s: &str) -> std::result::Result<Self, Self::Err> {
+        use regex::Regex;
+
+        let re = Regex::new(
+            r"^One time key signed_curve25519:[A-Za-z0-9+/=]+ already exists\. Old key:",
+        )
+        .expect("We should be able to build the one-time key error message regex");
+
+        // First we split the string into two parts, the part containing the old key and
+        // the part containing the new key. The parts are conveninetly separated
```
andybalaam: conveninetly -> conveniently

crates/matrix-sdk/src/encryption/mod.rs
```diff
@@ -380,6 +384,66 @@ struct OAuthCrossSigningResetUiaaResetParameter {
...
+        // by a `;` character.
+        let mut split = s.split_terminator(';');
```
andybalaam, line 404: TIL `split_terminator`. Rust has all the nice toys.

crates/matrix-sdk/src/encryption/mod.rs
```diff
@@ -380,6 +384,66 @@ struct OAuthCrossSigningResetUiaaResetParameter {
...
+        let old_key = split
+            .next()
+            .ok_or(serde_json::Error::custom("Old key is missing in the error message"))?;
+        let new_key = split
+            .next()
+            .ok_or(serde_json::Error::custom("New key is missing in the error message"))?;
+
+        // Now we remove the lengthy prefix from the part containing the old key, we
+        // should be left with just the JSON of the signed key.
+        let old_key = re.replace(old_key, "");
```
andybalaam: Is it worth the regex dependency? We could just search for "Old key:"...

crates/matrix-sdk/src/encryption/mod.rs
```diff
@@ -644,18 +708,23 @@ impl Client {
...
+                                            DuplicateOneTimeKeyErrorMessage::from_str(message)
+                                        {
+                                            tracing::error!(
+                                                sentry = true,
+                                                old_key = %message.old_key,
+                                                new_key = %message.new_key,
+                                                "Duplicate one-time keys have been uploaded"
+                                            );
+                                        } else {
+                                            self.state_store()
```
andybalaam: Previously we always did this step, and now we only do it if parsing the error message fails. Can you explain why? (Maybe a comment here about what we are doing would help?)

crates/matrix-sdk/src/encryption/mod.rs
```diff
@@ -2327,4 +2399,19 @@ mod tests {
         OAuthCrossSigningResetInfo::from_auth_info(&auth_info)
             .expect("We should be able to fetch the cross-signing reset info from the auth info");
     }
+
+    #[test]
+    fn test_duplicate_one_time_key_error_parsing() {
```
andybalaam, line 2409: Would be good to test an error case of some kind.

crates/matrix-sdk/src/encryption/mod.rs
```diff
@@ -2327,4 +2399,19 @@ mod tests {
...
+        let message = concat!(
+            r#"One time key signed_curve25519:AAAAAAAAAAA already exists. "#,
+            r#"Old key: {"key":"dBcZBzQaiQYWf6rBPh2QypIOB/dxSoTeyaFaxNNbeHs","#,
+            r#""signatures":{"@example:matrix.org":{"ed25519:AAAAAAAAAA":""#,
+            r#"Fk45zHAbrd+1j9wZXLjL2Y/+DU/Mnz9yuvlfYBOOT7qExN2Jdud+5BAuNs8nZ/caS4wTF39Kg3zQpzaGERoCBg"}}};"#,
+            r#" new key: {'key': 'CY0TWVK1/Kj3ZADuBcGe3UKvpT+IKAPMUsMeJhSDqno', "#,
+            r#"'signatures': {'@example:matrix.org': {'ed25519:AAAAAAAAAA': "#,
+            r#"'BQ9Gp0p+6srF+c8OyruqKKd9R4yaub3THYAyyBB/7X/rG8BwcAqFynzl1aGyFYun4Q+087a5OSiglCXI+/kQAA'}}}"#
+        );
+        DuplicateOneTimeKeyErrorMessage::from_str(message)
```
andybalaam: We can assert on the values we found here, right?

crates/matrix-sdk/src/encryption/mod.rs
```diff
@@ -2327,4 +2399,19 @@ mod tests {
...
+    #[test]
+    fn test_duplicate_one_time_key_error_parsing() {
+        let message = concat!(
+            r#"One time key signed_curve25519:AAAAAAAAAAA already exists. "#,
+            r#"Old key: {"key":"dBcZBzQaiQYWf6rBPh2QypIOB/dxSoTeyaFaxNNbeHs","#,
+            r#""signatures":{"@example:matrix.org":{"ed25519:AAAAAAAAAA":""#,
+            r#"Fk45zHAbrd+1j9wZXLjL2Y/+DU/Mnz9yuvlfYBOOT7qExN2Jdud+5BAuNs8nZ/caS4wTF39Kg3zQpzaGERoCBg"}}};"#,
+            r#" new key: {'key': 'CY0TWVK1/Kj3ZADuBcGe3UKvpT+IKAPMUsMeJhSDqno', "#,
+            r#"'signatures': {'@example:matrix.org': {'ed25519:AAAAAAAAAA': "#,
+            r#"'BQ9Gp0p+6srF+c8OyruqKKd9R4yaub3THYAyyBB/7X/rG8BwcAqFynzl1aGyFYun4Q+087a5OSiglCXI+/kQAA'}}}"#
```
andybalaam, line 2417: Would something like this be clearer?

```
            r#"\
                One time key signed_curve25519:AAAAAAAAAAA already exists. \
                Old key: \
                {
                    "key":"dBcZBzQaiQYWf6rBPh2QypIOB/dxSoTeyaFaxNNbeHs",\
                    "signatures":{\
                        "@example:matrix.org":{\
                            "ed25519:AAAAAAAAAA": "Fk45zHAbrd+1j9wZXLjL2Y/+DU/Mnz9yuvlfYBOOT7qExN2Jdud+5BAuNs8nZ/caS4wTF39Kg3zQpzaGERoCBg"\
...
             "#
```

(Matter of taste, so feel free to ignore)

crates/matrix-sdk/src/encryption/mod.rs
poljar: The dependency is already in the tree, i.e. already used by the base crate. That being said, yeah that's a good idea.

crates/matrix-sdk/src/encryption/mod.rs
poljar: Oh boy, the else was supposed to log the error without the one-time keys. Fixed in https://github.com/matrix-org/matrix-rust-sdk/pull/5651/commits/214dd8ddbda4e1c27092d13805f92c4135f3efa2

crates/matrix-sdk/src/encryption/mod.rs
poljar: That doesn't work with raw string literals.

crates/matrix-sdk/src/encryption/mod.rs
poljar: Fixed in https://github.com/matrix-org/matrix-rust-sdk/pull/5651/commits/9d8429e6334d79f072ada1bba366684bce105d0e.

crates/matrix-sdk/src/encryption/mod.rs
poljar: https://github.com/matrix-org/matrix-rust-sdk/pull/5651/commits/523a996e6c753be9225d9b16d575105e80151469

poljar: Also https://github.com/crate-ci/typos/issues/1370#issuecomment-3274780422.

crates/matrix-sdk/src/encryption/mod.rs
andybalaam: Darnit.

crates/matrix-sdk/src/encryption/mod.rs
```diff
@@ -2327,4 +2404,25 @@ mod tests {
...
+        );
+        let message = DuplicateOneTimeKeyErrorMessage::from_str(message)
+            .expect("We should be able to parse the error message");
+
+        assert_eq!(message.old_key.to_base64(), "dBcZBzQaiQYWf6rBPh2QypIOB/dxSoTeyaFaxNNbeHs");
+        assert_eq!(message.new_key.to_base64(), "CY0TWVK1/Kj3ZADuBcGe3UKvpT+IKAPMUsMeJhSDqno");
+
+        DuplicateOneTimeKeyErrorMessage::from_str("One time key already exists.")
+            .expect_err("We should't be able to parse an incomplete error message");
```
andybalaam: should't -> shouldn't

poljar: https://github.com/matrix-org/matrix-rust-sdk/pull/5651/commits/8a45483895ae98802f339d199abb463df649c7e9

crates/matrix-sdk-base/Cargo.toml
```diff
@@ -82,7 +82,7 @@ matrix-sdk-crypto = { workspace = true, optional = true }
 matrix-sdk-store-encryption.workspace = true
 matrix-sdk-test = { workspace = true, optional = true }
 once_cell.workspace = true
-regex = "1.11.2"
+regex.workspace = true
```
andybalaam, line 85: Don't need this any more

poljar: True, but doesn't hurt to leave it like this.

PR #5652, opened 2025-09-10
Test for keys arriving in the middle of adding UTD to timeline

andybalaam:

Signed-off-by:


kathleenmotley49-maker (approved):

kathleenmotley49-maker (approved):

poljar:
Alright, I have ported the test over to the event cache redecryptor now.

The problem continues to exist as we're again have two tasks, one that adds the event to the event cache and one that attempts to redecrypt the events.

I'm going to close this now and open a new PR, hopefully containing the ported test and fix.

Thanks a lot for the test, it is indeed a great way to test this using a delayed store.

PR #5653, opened 2025-09-10
refactor(common): Rename `store_locks` to `cross_process_lock`

Hywan:
This set of patches refactors the cross-process lock API. Each patch can be reviewed individually, but in short:

- `CrossProcessStoreLock*` are renamed to `CrossProcessLock*` (the `Store` part is removed),
- `BackingStore` is renamed to `TryLock`,
- The `store_locks` module becomes `cross_process_lock`,
- `LockStoreError` becomes `CrossProcessLockError`.

The idea is: cross-process lock isn't tailored to stores, it can be used for something else. Plus, it makes name shorter. The name `TryLock` has been decided with the rest of the team.


poljar (approved):
Makes sense.

PR #5654, opened 2025-09-10
crypto: Simplify `PrivateCrossSigningIdentity::with_account`

richvdh:
`PrivateCrossSigningIdentity::with_account` is called in lots of places where we end up throwing away most of its work. Indeed, there is only one place that actually uses the results (`olm::Account::bootstrap_cross_signing`).

It makes more sense to me to replace it with something similar, and hoist the remaining functionality up to `bootstrap_cross_signing`.

### DIFF
--- a/crates/matrix-sdk-crypto/src/olm/account.rs
+++ b/crates/matrix-sdk-crypto/src/olm/account.rs
@@ -806,11 +806,30 @@ impl Account {
         device_keys
     }
 
-    /// Bootstrap Cross-Signing
+    /// Bootstraps cross-signing, generating new cross-signing keys and creating
+    /// the necessary upload and signature requests.
+    ///
+    /// # Returns
+    /// A tuple containing:
+    /// - [`PrivateCrossSigningIdentity`]: The newly-generated cross-signing
+    ///   identity (including a signature from this device).
+    /// - [`UploadSigningKeysRequest`]: The request to upload the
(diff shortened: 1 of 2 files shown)


crates/matrix-sdk-crypto/CHANGELOG.md
```diff
@@ -6,6 +6,12 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased] - ReleaseDate
 
+### Refactor
+
+- [**breaking**] Remove `matrix_sdk_crypto::olm::PrivateCrossSigningIdentity::with_account`. Applications should instead use `matrix_sdk_crypto::olm::Account::bootstrap_cross_signing` which has identical behaviour.
+  `PrivateCrossSigningIdentity::for_account` is added for test situations where the upload requests are not required.
+  ([#5654](https://github.com/matrix-org/matrix-rust-sdk/pull/5654))
```
poljar: This was not a public method, as such this can't be a breaking change.

Do we now need to expose a public function for this?

richvdh: > This was not a public method, as such this can't be a breaking change.

ohhh it was pub(crate), I missed that. And no, we don't need a public function. Let me have another go at this.

richvdh: @poljar I've made the new method `pub(crate)`, and removed the changelog entry.

Issue #4951, opened 2025-04-17
Add disable_badge_count parameter to the set_pusher method.


bmarty:
The method `client.set_pusher` should add `org.matrix.msc4076.disable_badge_count: true` in the data sent to the server to let the server know that the application do not want to receive push just about updating the unread count.

This should reduce the number of Pushes that the application is receiving (with high priority) and which are at the end ignored by both Element X Android and Element X iOS.

Not sure if this parameter should be exposed in the Client API or if this can be just a hard-coded value in the SDK.

Ref:
- MSC: https://github.com/matrix-org/matrix-spec-proposals/pull/4076
- Synapse impl (there is a feature flag): https://github.com/element-hq/synapse/pull/17975

PR #5658, opened 2025-09-11
feat: Add `disable_badge_count` field to set pusher request

jmartinesp:
This allows the clients to stop receiving push notifications for unread counts in their devices, when it's not useful.

Needs https://github.com/ruma/ruma/pull/2214.

Implements https://github.com/matrix-org/matrix-rust-sdk/issues/4951

Signed-off-by:


jmartinesp:
In the end it seems like we're not interested on implementing the MSC (yet?).

PR #5659, opened 2025-09-11
feat(base): Add `get_room_events` to `EventCacheStore` trait and impls.

dragonfly1033:
Add a `get_room_events` function to the `EventCacheStore` trait and implement it for `MemoryStore`, `SqliteEventCacheStore` and `IndexeddbEventCacheStore`.

Also add a test `test_get_room_events` for this new function in `EventCacheStoreIntegrationTests`

### DIFF
--- a/crates/matrix-sdk-base/src/event_cache/store/traits.rs
+++ b/crates/matrix-sdk-base/src/event_cache/store/traits.rs
@@ -154,6 +154,12 @@ pub trait EventCacheStore: AsyncTraitDeps {
         filter: Option<&[RelationType]>,
     ) -> Result<Vec<(Event, Option<Position>)>, Self::Error>;
 
+    /// Get all events in this room.
+    ///
+    /// This method must return events saved either in any linked chunks, *or*
+    /// events saved "out-of-band" with the [`Self::save_event`] method.
+    async fn get_room_events(&self, room_id: &RoomId) -> Result<Vec<Event>, Self::Error>;
+
     /// Save an event, that might or might not be part of an existing linked
     /// chunk.
     ///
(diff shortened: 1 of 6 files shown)


crates/matrix-sdk-sqlite/src/event_cache_store.rs
```diff
@@ -1282,6 +1282,32 @@ impl EventCacheStore for SqliteEventCacheStore {
             .await
     }
 
+    #[instrument(skip(self))]
+    async fn get_room_events(&self, room_id: &RoomId) -> Result<Vec<Event>, Self::Error> {
+        let _timer = timer!("method");
```
poljar, line 1287: Which method?

crates/matrix-sdk-base/src/event_cache/store/traits.rs
```diff
@@ -154,6 +154,12 @@ pub trait EventCacheStore: AsyncTraitDeps {
         filter: Option<&[RelationType]>,
     ) -> Result<Vec<(Event, Option<Position>)>, Self::Error>;
 
+    /// Get all events in this room.
+    ///
+    /// This method must return events saved either in any linked chunks, *or*
+    /// events saved "out-of-band" with the [`Self::save_event`] method.
+    async fn get_room_events(&self, room_id: &RoomId) -> Result<Vec<Event>, Self::Error>;
```
poljar, line 161: Hmm, this should give us some more options than just load everything all at once.

Letting you get events of a certain type would be nice, or limit things and paginate. For the decryptor it would be nice to get only a certain amount of `m.room.encrypted` events that contain a certain `session ID` in the content.

But I see that the tables for the events doesn't contain even a `type` row, so this isn't completely trivial.

I guess we leave this for another PR. I'll need this soon, so I'll probably will tackle this.

crates/matrix-sdk-sqlite/src/event_cache_store.rs
poljar: Ah I see what's going on, there's no reason to specify the method again since the `instrument` will include the name.

Alright, nevermind me then.

PR #5660, opened 2025-09-11
fix(timeline): avoid replacing timeline items when the encryption info is unchanged

Johennes:
`make_replacement_for` is intended to update the `EncryptionInfo` on `TimelineItem`s. However, it also creates replacements when the info is unchanged from before which can cause massive, unbatched timeline updates. This change avoids creating replacements and triggering timeline updates when the encryption info is unchanged.

### DIFF
--- a/crates/matrix-sdk-ui/src/timeline/controller/decryption_retry_task.rs
+++ b/crates/matrix-sdk-ui/src/timeline/controller/decryption_retry_task.rs
@@ -404,7 +404,8 @@ pub(super) async fn retry_fetch_encryption_info<P: RoomDataProvider>(
 
 /// Create a replacement TimelineItem for the supplied one, with new
 /// [`EncryptionInfo`] from the supplied `room_data_provider`. Returns None if
-/// the supplied item is not a remote event, or if it doesn't have a session ID.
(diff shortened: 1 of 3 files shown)


crates/matrix-sdk-ui/src/timeline/controller/decryption_retry_task.rs
```diff
@@ -416,6 +416,12 @@ async fn make_replacement_for<P: RoomDataProvider>(
 
     let new_encryption_info =
         room_data_provider.get_encryption_info(session_id, &event.sender).await;
+
+    // Only create a replacement if the encryption info actually changed.
+    if remote.encryption_info == new_encryption_info {
+        return None;
```
andybalaam, line 423: The doc comment for this function could do with an update to explain the new behaviour.

Johennes: Oh, good point. Have updated the comment and squashed everything into a single commit. I forgot using fixups, sorry just came back from a 3-week vacation. 🙈

andybalaam: No worries - fine for something this small anyway. I hope it was a good vacation!

PR #5662, opened 2025-09-11
perf(common): Compute the `message` in `TracingTimer` when required

Hywan:
This patch moves the creation of the `message` in `TracingTimer` if and only if the log is enabled. Computing it every time is useless, and can even slow things down (because of the time calculation).

### DIFF
--- a/crates/matrix-sdk-common/src/tracing_timer.rs
+++ b/crates/matrix-sdk-common/src/tracing_timer.rs
@@ -33,6 +33,8 @@ impl std::fmt::Debug for TracingTimer {
 
 impl Drop for TracingTimer {
     fn drop(&mut self) {
+        let elapsed = self.start.elapsed();
+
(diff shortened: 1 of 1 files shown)


crates/matrix-sdk-common/src/tracing_timer.rs
```diff
@@ -45,6 +43,8 @@ impl Drop for TracingTimer {
             return;
         }
 
+        let message = format!("_{}_ finished in {:?}", self.id, self.start.elapsed());
```
bnjbvr: Hmm, this means the elapsed time now includes the time to check the tracing level is enabled. Could we call `.elapsed()` *before*, store it in a temporary local, and use it here, instead? This means we'd always compute `.elapsed()`, which would result in being more precise, at the cost of doing the system call to compute the "now" time, so I'm a bit torn myself with doing that…

bnjbvr (approved):

PR #5663, opened 2025-09-12
feat(spaces): expose `SpaceRoomList` `SpaceRoom` and an `is_direct` flag

stefanceriu:
This series of patches add features requested by the final client teams:
* the "owner" space of a particular `SpaceRoomList` is now exposed as as `SpaceRoom`
    * converged on plaining calling it `space` in that context
* `SpaceRooms` now have an `is_direct` flag

### DIFF
--- a/crates/matrix-sdk-ui/src/spaces/room_list.rs
+++ b/crates/matrix-sdk-ui/src/spaces/room_list.rs
@@ -256,26 +256,35 @@ impl SpaceRoomList {
                 };
 
                 let mut rooms = self.rooms.lock();
-                result
-                    .rooms
+
+                // The space is part of the /hierarchy response. Partition the room array
+                // so we can use its details but also filter it out of the room list
+                let (space, children): (Vec<_>, Vec<_>) =
+                    result.rooms.into_iter().partition(|f| f.summary.room_id == self.space_id);
+
+                if let Some(room) = space.first() {
+                    let mut space = self.space.write();
+
+                    if space.is_none() {
+                        ObservableWriteGuard::set(
+                            &mut space,
+                            Some(SpaceRoom::new_from_summary(
+                                &room.summary,
+                                self.client.get_room(&room.summary.room_id),
+                                room.children_state.len() as u64,
+                            )),
+                        );
+                    }
+                }
+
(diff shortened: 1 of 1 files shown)


crates/matrix-sdk-ui/src/spaces/room_list.rs
```diff
@@ -229,15 +257,22 @@ impl SpaceRoomList {
...
-                                &room.summary,
-                                self.client.get_room(&room.summary.room_id),
-                                room.children_state.len() as u64,
-                            ))
+                            let mut parent_space = self.parent_space.write();
+                            if parent_space.is_none() {
+                                ObservableWriteGuard::set(&mut parent_space, space_room);
+                            }
+
+                            return None;
```
poljar: Are you adding here side-effects that are unrelated to the filter operation?

Why does the populating of `self.parent_space` belong here?

stefanceriu: Well the API allows you to create a `SpaceRoomList` for any room id.  If the space happens to be known by the client at creation time then we can populate it directly. 

Otherwise we can wait for the first `/hierarchy` pagination request to complete and read it from there instead as it will be included (we were intentionally filtering it out previously). 

This is in order to avoid making a separate call to the `/room_summary` endpoint which will result in the same exact data.

If you're referring _only_ to the `.filter_map` then should I rewrite it using a for each or..? 🤔

poljar: I meant the `filter_map` part, I think we should just separate the operation into a separate step so we don't obfuscate the filter logic too much.

Modifying elements that will be filtered out in the `filter_map()` call sounds like a abuse of the API. `filter_map` is a combination of `filter()` and `map()` so you first filter, then you're supposed to map the values that remain into a different type.

If you want to mutate items that will be filtered out, that should happen in a loop before the filtering.

You can also use `partition()` to get two buckets so you don't have to loop twice over all items.

stefanceriu: Oh okay, that's fair, let me see what I can do 👍

PR #5666, opened 2025-09-12
fix(crypto): Report inner-outer state key differences as invalid.

kaylendog:
Fixes a [nit](https://github.com/matrix-org/matrix-js-sdk/pull/4994/files#r2344224610) @richvdh thought of while reviewing my changes to the JS SDK. 

- [ ] Modifies `OlmMachine::verify_packed_state_key` to correctly discard events that do not contain cleartext state keys, but do in their content, and vice versa.
- [ ] Add tests to test correct handling of both cases.


poljar (approved):
Looks good.

PR #5668, opened 2025-09-15
feat(rtc): Remove deprecated CallNotify in favour of RtcNotification

BillCarsonFr:
`CallNotify` event has been deprecated in favour of `RtcNotification` event https://github.com/ruma/ruma/pull/2199

Needs an update on ruma dependency

Signed-off-by:

### DIFF
--- a/bindings/matrix-sdk-ffi/src/ruma.rs
+++ b/bindings/matrix-sdk-ffi/src/ruma.rs
@@ -23,7 +23,6 @@ use matrix_sdk::attachment::{BaseAudioInfo, BaseFileInfo, BaseImageInfo, BaseVid
 use ruma::{
     assign,
     events::{
-        call::notify::NotifyType as RumaNotifyType,
         direct::DirectEventContent,
         fully_read::FullyReadEventContent,
         identity_server::IdentityServerEventContent,
@@ -57,6 +56,7 @@ use ruma::{
             ImageInfo as RumaImageInfo, MediaSource as RumaMediaSource,
             ThumbnailInfo as RumaThumbnailInfo,
         },
+        rtc::notification::NotificationType as RumaNotificationType,
         secret_storage::{
             default_key::SecretStorageDefaultKeyEventContent,
             key::{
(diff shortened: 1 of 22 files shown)


bindings/matrix-sdk-ffi/src/ruma.rs
```diff
@@ -487,25 +487,25 @@ impl TryFrom<RumaMessageType> for MessageType {
 }
 
 #[derive(Clone, uniffi::Enum)]
-pub enum NotifyType {
+pub enum RtcNotificationType {
     Ring,
     Notify,
```
bnjbvr: Should it be renamed `Notification` to align with Ruma?

testing/matrix-sdk-test/src/event_factory.rs
```diff
@@ -1003,15 +1003,21 @@ impl EventFactory {
...
+        mentions: Option<Mentions>,
+        rtc_membership_event_id: Option<OwnedEventId>,
+    ) -> EventBuilder<RtcNotificationEventContent> {
+        let mut content = RtcNotificationEventContent::new(
+            MilliSecondsSinceUnixEpoch::now(),
+            Duration::new(20, 0),
+            notification_type,
+        );
+        content.relates_to = rtc_membership_event_id.map(Reference::new);
+        content.mentions = mentions;
```
bnjbvr: The more idiomatic way with the `EventFactory` builder pattern would be to add `fn mentions()` and `fn rtc_membership_event()` methods on `impl EventBuilder<RtcNotificationEventContent>`, so that we don't have `Option`al parameters in this ctor. Also, I would rather make it so that the timestamp and durations are configurable (aka function parameters), so we don't have tight coupling between the test framework and the tests themselves (or, at least let's rename `rtc_notification` to `predefined_rtc_notification` or something like that)

bnjbvr (changes requested):
Thanks. Two minor comments below, and a few extra questions:

- being unaware of the MSC et al., does this maintain backwards compatibility with the existing system and events?
- linked to the other question: does it require a timed merge with an Element Call release, or can this PR be merged at any time?

testing/matrix-sdk-test/src/event_factory.rs
BillCarsonFr: Thanks for all the tips! I missed the fact that one could do `impl EventBuilder<CONTENT>`. 
Done here https://github.com/matrix-org/matrix-rust-sdk/pull/5668/commits/0505ce28c3fbf4a11e93a61f913b178eb23d9412

BillCarsonFr:
> Thanks. Two minor comments below, and a few extra questions:
> 
>     * being unaware of the MSC et al., does this maintain backwards compatibility with the existing system and events?
> 
>     * linked to the other question: does it require a timed merge with an Element Call release, or can this PR be merged at any time?

Good Q. Given that EC is already sending both the old `mscXXX.call.decline` as well as the new `mscXXX.rtc.notification`,  we decided that we don't need to support both nor to require timed merged. It can be merged anytime

bindings/matrix-sdk-ffi/src/ruma.rs
BillCarsonFr: right, done https://github.com/matrix-org/matrix-rust-sdk/pull/5668/commits/9d53bc1a801a21cbd527ec693c9e36217d98373b

bnjbvr (approved):
Perfect then, thanks!

Issue #5535, opened 2025-08-14
timeline: support permalink to a threaded message


bnjbvr:
Right now, the permalink feature works with an event-focused timeline. Under the hood, we start a `/context` query that focused on the permalink's target, through a `Paginator` instance, and paginates backwards and forwards based on it.

I think the scheme could be tweaked to support permalink to in-thread events:

- add a new bool/enum in `TimelineFocusKind::Event` that would tell if we're on a threaded timeline or not
- after receiving the initial response from /context, consider if the event is part of a thread (using `extract_thread_root` or the deserialized event `relates_to` field)
- if that's the case, switch the paginator to a thread `Paginator` implementation instead (note: it's already implemented and available in the SDK, but mostly unused and lightly tested), focused on the given event, and restart an initial query

I think the timeline would use a "facade" paginator instance, that would select either the room paginator (/context) or the thread paginator (/relations) based on the initial response of /context.

Some timeline methods might need to be tweaked (`infer_reply`, `infer_thread_for_read_receipt`, based on the new bool, etc.), to take into account that a `TimelineFocusKind::Event` can now be threaded too, and then most methods should work out of the box.

zecakeh:
If I may, I believe that this issue highlights a problem with `TimelineFocus`, it translates to not one but 2 settings: the kind of timeline and the initial position in the timeline.

I think that it would be worth it to split it into 2 separate settings, for exemple:

```rust
enum TimelineKind {
  Unthreaded // Or maybe Default?
  MainThread,
  Thread { .. },
  PinnedEvents { .. },
}

enum TimelineInitialPosition {
  End,
  Event { .. },
  // There could even be a Start variant later.
}
```

`TimelineInitialPosition` could be a no-op if it is not supported for a given `TimelineKind` (or log a warning, or trigger an error). Ideally a timeline should become what is called "Live" now whenever we reach the end of the timeline.

bnjbvr:
Thought about this too, but now it's possible to make invalid combinations that would only be detected at runtime (e.g. `Pinned` + `TimelineInitialPosition::Event { ... }`). Still worth considering, though.

bnjbvr:
This has been achieved, with the combination of mostly #5678 (kudos to @jmartinesp for that one!), and a few preparatory refactorings and followups, including but not limited to #5669, #5701, #5708, #5709, #6050.

PR #5669, opened 2025-09-15
refactor(timeline): some preparatory refactorings for in-thread permalinks

bnjbvr:
In support of https://github.com/matrix-org/matrix-rust-sdk/issues/5535, this changes two things:

- centralize the decision whether a timeline is thread-focused or not, by using a `match` with fields explicitly named (so that when adding/changing a field, we have to consider those `match` statements).
- add a new `Timeline::is_threaded()` function that will return whether a timeline is on a thread or not. This will be useful in the `Events` case, to know if the permalinked message belonged to a thread or to the main timeline.

### DIFF
--- a/crates/matrix-sdk-ui/src/timeline/controller/mod.rs
+++ b/crates/matrix-sdk-ui/src/timeline/controller/mod.rs
@@ -148,6 +148,17 @@ impl<P: RoomDataProvider> TimelineFocusKind<P> {
             TimelineFocusKind::PinnedEvents { .. } => ReceiptThread::Unthreaded,
         }
     }
+
+    /// Whether the focus is on a thread (from a live thread or a thread
+    /// permalink).
+    fn is_thread(&self) -> bool {
+        match self {
+            TimelineFocusKind::Live { hide_threaded_events: _ }
+            | TimelineFocusKind::Event { paginator: _, hide_threaded_events: _ }
+            | TimelineFocusKind::PinnedEvents { loader: _ } => false,
+            TimelineFocusKind::Thread { root_event_id: _ } => true,
+        }
+    }
 }
 
 #[derive(Clone, Debug)]
--- a/crates/matrix-sdk-ui/src/timeline/controller/mod.rs
+++ b/crates/matrix-sdk-ui/src/timeline/controller/mod.rs
@@ -133,30 +133,37 @@ impl<P: RoomDataProvider> TimelineFocusKind<P> {
     /// general, unless they hide in-thread events, in which case they will
     /// use the main thread.
     pub(super) fn receipt_thread(&self) -> ReceiptThread {
+        if let Some(thread_root) = self.thread_root() {
+            ReceiptThread::Thread(thread_root.to_owned())
+        } else if self.hide_threaded_events() {
+            ReceiptThread::Main
+        } else {
+            ReceiptThread::Unthreaded
+        }
+    }
+
+    /// Whether to hide in-thread events from the timeline.
+    fn hide_threaded_events(&self) -> bool {
         match self {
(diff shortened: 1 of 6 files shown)


crates/matrix-sdk-ui/src/timeline/controller/mod.rs
```diff
@@ -148,6 +148,17 @@ impl<P: RoomDataProvider> TimelineFocusKind<P> {
...
     }
+
+    /// Whether the focus is on a thread (from a live thread or a thread
+    /// permalink).
+    fn is_thread(&self) -> bool {
+        match self {
+            TimelineFocusKind::Live { hide_threaded_events: _ }
+            | TimelineFocusKind::Event { paginator: _, hide_threaded_events: _ }
+            | TimelineFocusKind::PinnedEvents { loader: _ } => false,
+            TimelineFocusKind::Thread { root_event_id: _ } => true,
```
stefanceriu: Surely all of these except `TimelineFocusKind::Event` can keep their unamed parameters, no? 

Wouldn't expect parameters changing the "threadness" of `Thread`, `Live` or `PinnedEvents` no matter what, no?

crates/matrix-sdk-ui/tests/integration/timeline/mod.rs
```diff
@@ -67,6 +68,69 @@ mod thread;
...
+        let f = EventFactory::new();
+        let event = f
+            .text_msg("hello world")
+            .event_id(event_id!("$target"))
+            .room(room_id)
+            .sender(&ALICE)
+            .into_event();
+        server.mock_room_event_context().ok(event, "", "", Vec::new()).mount().await;
+
+        // An event-focused timeline isn't threaded.
```
stefanceriu, line 107: Yet?

crates/matrix-sdk-ui/src/timeline/controller/mod.rs
```diff
@@ -604,11 +604,18 @@ impl<P: RoomDataProvider> TimelineController<P> {
 
     /// Is this timeline focused on a thread?
     pub(super) fn is_threaded(&self) -> bool {
-        matches!(&*self.focus, TimelineFocusKind::Thread { .. })
+        self.focus.is_thread()
     }
 
+    /// The root of the current thread, for a live thread timeline or a
+    /// permalink to a thread message.
```
stefanceriu, line 618: > permalink to a thread message.

Is the idea here for `TimelineFocusKind::Event` to eventually handle threaded permalinks too?

stefanceriu (approved):
Makes sense to me apart from a couple of small comments 👍

crates/matrix-sdk-ui/src/timeline/controller/mod.rs
bnjbvr: Yeah, I'll keep them in the ellipsis :+1:

crates/matrix-sdk-ui/tests/integration/timeline/mod.rs
bnjbvr: Indeed! I don't think it's worth commenting "yet" in the code, though, since this is going to happen, and it's unlikely that we'll remember to update it in the test :)

crates/matrix-sdk-ui/src/timeline/controller/mod.rs
bnjbvr: Exactly, you can read the full plan in https://github.com/matrix-org/matrix-rust-sdk/issues/5535

bnjbvr:
@stefanceriu I've pushed another commit, but it includes more changes that are slightly different from the initial version, so retagging you for review.

stefanceriu (approved):
lgtm 👍

Issue #4874, opened 2025-04-01
Introduce poison inside the cross-process lock


Hywan:
[During our work on the event cache store](https://github.com/matrix-org/matrix-rust-sdk/issues/3280), we have worked on [improving the event cache store to introduce `lock` and `lock_unchecked` with the notion of a poison](https://github.com/matrix-org/matrix-rust-sdk/pull/4285). If the lock has been taken by another process, the lock is poisoned and returned an error. This is helpful to know when to refresh the state of the data owned by such a lock.

This work is nice but we realise it should be extended to all cross-process locks, not only the event cache store.

Tasks:

- [ ] Introduce a generation mechanism right inside the cross-process lock type,
- [ ] Update `lock()` to return a poison if any,
- [ ] Add `lock_unchecked()` to lock regardless of the poison,
- [ ] Implement this new mechanism in the event cache store
- [ ] Remove the existing similar mechanism in the crypto store
  - [ ] A clean up in the crypto store is necessary to put all the states in a single place, so that cache invalidation is easier

Hywan:
Also see this comment https://github.com/matrix-org/matrix-rust-sdk/pull/4285#issuecomment-2488655073

PR #5672, opened 2025-09-15
feat(common): Add cross-process lock dirtying

Hywan:
This set of patches updates the cross-process lock to introduce “dirtiness”. The lock is considered “dirty” when the lock is acquired from another holder:

- holder A takes the lock for the first time: not dirty,
- holder A takes the lock again, not dirty,
- holder B takes the lock for the first time: not dirty,
- holder A takes the lock again: **dirty**

To achieve so, a _generation_ value is added in the stores. This generation is a `u64`. Every time the lock is acquired from a previous holder, this generation is incremented by one. That's basically it. The first value for the generation is 0, which acts as a guard value.

The `TryLock::try_lock` method now returns an `Option<CrossProcessLockGeneration>` instead of `bool`.

The `CrossProcessLock::try_lock_once` method now returns a `CrossProcessLockResult` instead of an `Option<CrossProcessLockGuard>`. It's a new type that looks like this:

```rust
enum CrossProcessLockResult {
    Clean(CrossProcessLockGuard),
    Dirty(CrossProcessLockGuard),
    Unobtained,
}
```

The `CrossProcessLock::spin_lock` method doesn't change for the moment. Let's do that in another PR where we will use the new API to decide how to invalidate the various states here and there. I wanted to keep this PR as small as possible first.

---

### DIFF
--- a/crates/matrix-sdk-common/CHANGELOG.md
+++ b/crates/matrix-sdk-common/CHANGELOG.md
@@ -19,7 +19,7 @@ All notable changes to this project will be documented in this file.
 
 - [**breaking**] Use `Raw<AnyTimelineEvent>` in place of `Raw<AnyMessageLikeEvent>`
   in `DecryptedRoomEvent::event`.
-  ([#5512](https://github.com/matrix-org/matrix-rust-sdk/pull/5512/files)).
+  ([#5512](https://github.com/matrix-org/matrix-rust-sdk/pull/5512)).
   Affects the following functions:
   - `OlmMachine::decrypt_room_event` - existing matches on the result's event field
      should be updated to `AnyTimelineEvent::MessageLike(AnyMessageLikeEvent::...)`


kathleenmotley49-maker (approved):

Hywan:
Must be merged after https://github.com/matrix-org/matrix-rust-sdk/pull/5829.

crates/matrix-sdk-common/src/cross_process_lock.rs
```diff
@@ -26,7 +26,7 @@
 //! timestamp on the side; see also `CryptoStore::try_take_leased_lock` for more
 //! details.
 //!
-//! The lock is initially acquired for a certain period of time (namely, the
+//! The lock is initially obtainedd for a certain period of time (namely, the
```
stefanceriu: ### 
```suggestion
//! The lock is initially obtained for a certain period of time (namely, the
```

crates/matrix-sdk-sqlite/src/crypto_store.rs
```diff
@@ -1488,37 +1499,52 @@ impl CryptoStore for SqliteCryptoStore {
...
+                            expiration = excluded.expiration,
+                            generation =
+                                CASE holder
+                                    WHEN excluded.holder THEN generation
+                                    ELSE generation + 1
+                                END
+                        WHERE
+                            holder = excluded.holder
+                            OR expiration < ?4
+                    RETURNING generation
```
poljar, line 1538: It feels a bit wrong to have this three times, but moving this to a common DB seems like it would cause other kind of trouble.

I don't have a better idea here :sweat_smile:

crates/matrix-sdk-common/src/cross_process_lock.rs
```diff
@@ -360,6 +462,42 @@ where
...
+            Self::Clean(guard) | Self::Dirty(guard) => Some(guard),
+            Self::Unobtained => None,
+        }
+    }
+
+    /// Return `true` if the lock has been obtained, `false` otherwise.
+    pub fn is_ok(&self) -> bool {
+        matches!(self, Self::Clean(_) | Self::Dirty(_))
+    }
+}
```
poljar, line 499: Something we mentioned in chat but for archeological reasons I'll repeat it here.

I don't think it's worth it to reimplement the `Result` type to avoid having `Result<Result<T>>`.

But I think we left the modification of the return type to another PR.

So nothing to do here.

poljar (approved):
Alright, I think all of this makes sense to me and it looks good.

crates/matrix-sdk-sqlite/src/crypto_store.rs
Hywan: We could refactor that later to have a common implementation.

crates/matrix-sdk-common/src/cross_process_lock.rs
Hywan: Thanks for the future us.

PR #5673, opened 2025-09-15
feat(sdk): `LatestEvents` auto-listens from the sync

Hywan:
This patch updates `SlidingSyncResponseProcessor::handle_room_response` to automatically call `LatestEvents::listen_to_room` based on `http::Response`'s `rooms`.

Why? Because when a sync is received, we want its `LatestEventValue` to be computed, so that it can trigger a `RoomInfoNotableUpdate`, which will update the `RoomList`, which will re-sort the rooms. So far, `LatestEvents::listen_to_room` was called by `RoomListService::subscribe_to_rooms`, but it's possible to receive a room update via the sync for a room that is not subscribed, i.e. out of the viewport of the room list in Matrix clients (it's recommended to subscribe to rooms that “enter” the viewport). Without this patch, rooms are receiving updates but the room list is not entirely refreshed/recalculated.

Other patches also solve a dead lock found by Complement Crypto. Read the patches' description to learn more. 

---

* https://github.com/matrix-org/matrix-rust-sdk/issues/4112

### DIFF
--- a/crates/matrix-sdk/src/sliding_sync/client.rs
+++ b/crates/matrix-sdk/src/sliding_sync/client.rs
@@ -16,7 +16,7 @@ use ruma::{
 use tracing::error;
 
 use super::{SlidingSync, SlidingSyncBuilder};
-use crate::{Client, Result};
+use crate::{Client, Result, sync::subscribe_to_room_latest_events};
 
 /// A sliding sync version.
 #[derive(Clone, Debug)]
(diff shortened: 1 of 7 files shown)


crates/matrix-sdk/src/sliding_sync/client.rs
```diff
@@ -237,6 +239,21 @@ impl SlidingSyncResponseProcessor {
     }
 }
 
+/// Automatically calls `LatestEvents::listen_to_room` for rooms in `response`.
+///
+/// That way, the latest event is computed and updated for all rooms receiving
+/// an update from the sync.
+#[instrument(skip_all)]
+async fn auto_listen_to_latest_events(client: &Client, response: &http::Response) {
```
stefanceriu: This doesn't feel particularly automatic to me. I would've expected a background task or something. 
How about we just call it `subscribe_to_room_latest_events` or something along those longs?

Hywan: I agree. Let's change the name!

stefanceriu (approved):
Woo! 👏

PR #5677, opened 2025-09-16
chore: bump `matrix-sdk` to the 2024 edition

bnjbvr:
It all started with me noticing that `cargo nextest` warned about usage of the never type fallback in the SDK crate… and then it escalated a bit. This bumps the Rust edition of the matrix-sdk crate to 2024.

### Some notes

- Sorry that some commits include some reformatting changes. I should've started with that, and started to address the compilation errors first, then auto-format-on-save kicked in and included some `use` formatting changes in the mix. With some extra effort I could put them back into the `rustfmt` commit, tell me if you think it's worth it.
- I wonder if I can use the Rust edition 2024 in the `matrix-sdk` crate while some of its dependencies aren't using the 2024 edition (but some idioms that seem 2024-y to me, like `use<>` etc.). Let's see what CI thinks of it.

### DIFF
--- a/crates/matrix-sdk-crypto/src/store/crypto_store_wrapper.rs
+++ b/crates/matrix-sdk-crypto/src/store/crypto_store_wrapper.rs
@@ -321,7 +321,7 @@ impl CryptoStoreWrapper {
     /// the reader.
     pub fn room_keys_received_stream(
         &self,
-    ) -> impl Stream<Item = Result<Vec<RoomKeyInfo>, BroadcastStreamRecvError>> {
+    ) -> impl Stream<Item = Result<Vec<RoomKeyInfo>, BroadcastStreamRecvError>> + use<> {
         BroadcastStream::new(self.room_keys_received_sender.subscribe())
     }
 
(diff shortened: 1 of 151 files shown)


bnjbvr:
```
error: unneeded unit return type
   --> crates/matrix-sdk/src/event_handler/mod.rs:954:56
    |
954 |             async move |_ev: OriginalSyncRoomNameEvent| -> () {
    |                                                        ^^^^^^ help: remove the `-> ()`
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#unused_unit
    = note: `-D clippy::unused-unit` implied by `-D warnings`
    = help: to override `-D warnings` add `#[allow(clippy::unused_unit)]`
```

Ouch, Clippy and Rustc disagree:

- Rustc wants to know explicitly that we're coercing the `!` type (returned from `panic!` below) to a `()`, because of the new 2024 edition rule,
- but Clippy thinks the return type is useless, while it's used to have the coercion happen.

I'll add special ignore lines to this Clippy lint for these 3 lines.

andybalaam (approved):
Looks good to me! I can live with the formatting.

PR #5678, opened 2025-09-16
feat: Add event focused pagination for threads

jmartinesp:
Changes:

- `TimelineFocusKind::Event` now has its `paginator` wrapped into an `AnyPaginator` enum, so we can decide if the event pagination should be done on the main timeline or on a thread one.
- `ThreadedEventsLoader` now has a paginate forwards method too so it has a similar API to that of `Paginator`.
- There is a new `async fn Room::load_or_fetch_event(&self, event_id: &EventId) -> Result<TimelineEvent>` that returns an event. Along with `fn TimelineEvent::thread_root_event_id`, this is used by the clients to check if the event is part of a thread or not so we can decide which navigation flow should be used.

Signed-off-by:

### DIFF
--- a/crates/matrix-sdk-ui/CHANGELOG.md
+++ b/crates/matrix-sdk-ui/CHANGELOG.md
@@ -6,6 +6,11 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased] - ReleaseDate
 
+### Refactor
+
+- `TimelineFocusKind::Event` can now handle both the existing event pagination and thread pagination if the focused 
+  event is part of a thread ([#5678](https://github.com/matrix-org/matrix-rust-sdk/pull/5678)).
+
 ### Bug Fixes
 
 - Avoid replacing timeline items when the encryption info is unchanged.
@@ -62,7 +67,7 @@ All notable changes to this project will be documented in this file.
 ### Refactor
 
 - [**breaking**] The MSRV has been bumped to Rust 1.88.
-  ([#5431](https://github.com/matrix-org/matrix-rust-sdk/pull/5431))
+  ([#5431](https://github.com/matrix-org/matrix-rust-sdk/pull/5431)) 
 
 ### Bug Fixes
 
--- a/crates/matrix-sdk/CHANGELOG.md
+++ b/crates/matrix-sdk/CHANGELOG.md
@@ -60,6 +60,10 @@ All notable changes to this project will be documented in this file.
 
 - [**breaking**] `OAuth::login` now allows requesting additional scopes for the authorization code grant.
   ([#5395](https://github.com/matrix-org/matrix-rust-sdk/pull/5395))
+- [**breaking**] `ThreadedEventsLoader::new` now takes optional `tokens` parameter to customise where the pagination 
+  begins ([#5678](https://github.com/matrix-org/matrix-rust-sdk/pull/5678).
+- Make `PaginationTokens` `pub`, as well as its `previous` and `next` tokens so they can be assigned from other files 
+  ([#5678](https://github.com/matrix-org/matrix-rust-sdk/pull/5678).
 
 ### Refactor
 
--- a/bindings/matrix-sdk-ffi/CHANGELOG.md
+++ b/bindings/matrix-sdk-ffi/CHANGELOG.md
@@ -15,6 +15,11 @@ All notable changes to this project will be documented in this file.
 - Remove the deprecated `CallNotify` event (`org.matrix.msc4075.call.notify`) in favor of the new
   `RtcNotification` event (`org.matrix.msc4075.rtc.notification`).
 
+### Features:
+
+- Add `Room::load_or_fetch_event` so we can get a `TimelineEvent` given its event id ([#5678](https://github.com/matrix-org/matrix-rust-sdk/pull/5678)).
+- Add `TimelineEvent::thread_root_event_id` to expose the thread root event id for this type too ([#5678](https://github.com/matrix-org/matrix-rust-sdk/pull/5678)).
+
 ## [0.14.0] - 2025-09-04
 
 ### Features:


crates/matrix-sdk/src/paginators/thread.rs
```diff
@@ -63,17 +63,26 @@ pub struct ThreadedEventsLoader<P: PaginableThread> {
     room: P,
 
     /// The thread root event ID (the event that started the thread).
-    root_event_id: OwnedEventId,
+    pub root_event_id: OwnedEventId,
 
     /// The current pagination token, which is used to keep track of the
     /// pagination state.
-    token: Mutex<PaginationToken>,
+    tokens: Mutex<PaginationTokens>,
```
bnjbvr, line 70: nit: need to update the code comment now! (at least to pluralize)

crates/matrix-sdk/src/paginators/thread.rs
```diff
@@ -63,17 +63,26 @@ pub struct ThreadedEventsLoader<P: PaginableThread> {
     room: P,
 
     /// The thread root event ID (the event that started the thread).
-    root_event_id: OwnedEventId,
+    pub root_event_id: OwnedEventId,
```
bnjbvr: Why is this public? Instead of making it public, if you absolutely need to expose it: can you add an accessor function?

crates/matrix-sdk-ui/src/timeline/controller/mod.rs
```diff
@@ -108,7 +112,7 @@ pub(in crate::timeline) enum TimelineFocusKind<P: RoomDataProvider> {
     /// direction or another.
     Event {
         /// The paginator instance.
-        paginator: Paginator<P>,
+        paginator: OnceCell<AnyPaginator<P>>,
```
bnjbvr, line 113: Hmm, it doesn't make sense to have the `hide_threaded_events` in the context of a thread, because this would be unused then. I think, instead that the `AnyPaginator::Unthreaded` variant should own this field, now:

```rust
enum AnyPaginator {
 Unthreaded {
    paginator: Paginator<P>,
    hide_threaded_events: bool,
 }, // etc
}
```

crates/matrix-sdk-ui/src/timeline/controller/mod.rs
```diff
@@ -160,9 +170,15 @@ impl<P: RoomDataProvider> TimelineFocusKind<P> {
...
-            | TimelineFocusKind::Event { paginator: _, hide_threaded_events: _ }
-            | TimelineFocusKind::PinnedEvents { .. } => None,
+            TimelineFocusKind::Event { paginator, hide_threaded_events: _ } => {
+                match paginator.get() {
+                    Some(AnyPaginator::Threaded(threaded_event_loader)) => {
+                        Some(&threaded_event_loader.root_event_id)
+                    }
+                    _ => None,
+                }
+            }
```
bnjbvr: Instead, can you make a small `AnyPaginator::thread_root(&self)` method in a new `AnyPaginator` impl block? (can likely also use `as_variant!` here)

crates/matrix-sdk-ui/src/timeline/controller/mod.rs
```diff
@@ -402,16 +417,46 @@ impl<P: RoomDataProvider> TimelineController<P> {
...
 
+                let thread_root_event_id = start_from_result
+                    .events
+                    .iter()
+                    .find(
+                        |event| {
+                            if let Some(id) = event.event_id() { id == *event_id } else { false }
+                        },
+                    )
+                    .and_then(|raw_event| match &raw_event.kind {
```
bnjbvr: nit: it's not a `raw` event (i.e. it isn't of the form `Raw<T>`) as I think it's a `TimelineEvent`, right? Let's name it `event` :)

crates/matrix-sdk-ui/src/timeline/controller/mod.rs
```diff
@@ -402,16 +417,46 @@ impl<P: RoomDataProvider> TimelineController<P> {
...
+                        TimelineEventKind::Decrypted(event) => {
+                            let raw_event = event.event.clone().cast_unchecked();
+                            extract_thread_root(&raw_event)
+                        }
+                        TimelineEventKind::PlainText { event } => extract_thread_root(event),
+                        TimelineEventKind::UnableToDecrypt { .. } => None,
+                    });
```
bnjbvr: I think this can be simplified with `extact-thread_root(event.raw())`, so we don't have to worry about the timeline event kind? (my suggestion would likely be better in the presence of UTD too, since they can appear in a thread)

crates/matrix-sdk-ui/src/timeline/controller/mod.rs
```diff
@@ -402,16 +417,46 @@ impl<P: RoomDataProvider> TimelineController<P> {
...
+                });
+
                 let has_events = !start_from_result.events.is_empty();
+                let events = start_from_result.events;
 
                 self.replace_with_initial_remote_events(
-                    start_from_result.events.into_iter(),
+                    events.into_iter(),
                     RemoteEventOrigin::Pagination,
                 )
```
bnjbvr: Hmm, this will make it so that the timeline will include the events received from /context, while there might be messages in there that are *not* from the current thread. Instead, here's what should happen:

- if the any paginator is set to Unthreaded, then we can do what you're doing here: the events returned from /context are likely those we want to show now.
- if the any paginator is set to a Threaded instance, then we should run a first round of /relations, and use the resulting events here, instead

Can you also add a regression test, that would have shown this issue, please?

crates/matrix-sdk-ui/src/timeline/controller/mod.rs
```diff
@@ -557,10 +602,17 @@ impl<P: RoomDataProvider> TimelineController<P> {
...
-                .map_err(PaginationError::Paginator)?,
+            TimelineFocusKind::Event { paginator, .. } => match paginator.get() {
+                Some(AnyPaginator::Unthreaded(event_paginator)) => event_paginator
+                    .paginate_backward(num_events.into())
+                    .await
+                    .map_err(PaginationError::Paginator)?,
+                Some(AnyPaginator::Threaded(threaded_paginator)) => threaded_paginator
+                    .paginate_backwards(num_events.into())
+                    .await
+                    .map_err(PaginationError::Paginator)?,
```
bnjbvr: Can you add `AnyPaginator::paginate_backwards()`, so this logic is encapsulated, instead?

crates/matrix-sdk-ui/src/timeline/controller/mod.rs
```diff
@@ -587,10 +639,17 @@ impl<P: RoomDataProvider> TimelineController<P> {
...
-                .map_err(PaginationError::Paginator)?,
+            TimelineFocusKind::Event { paginator, .. } => match paginator.get() {
+                Some(AnyPaginator::Unthreaded(event_paginator)) => event_paginator
+                    .paginate_forward(num_events.into())
+                    .await
+                    .map_err(PaginationError::Paginator)?,
+                Some(AnyPaginator::Threaded(threaded_paginator)) => threaded_paginator
+                    .paginate_forwards(num_events.into())
+                    .await
+                    .map_err(PaginationError::Paginator)?,
```
bnjbvr: ditto here

crates/matrix-sdk-ui/tests/integration/timeline/focus_event.rs
```diff
@@ -418,3 +421,172 @@ async fn test_focused_timeline_doesnt_show_local_echoes() {
...
+    let threaded_event_id = threaded_event.event_id().unwrap().clone();
+
+    // Mock the initial /context request to check if the event is in a thread
+    server
+        .mock_room_event_context()
+        .room(room_id)
+        .ok(threaded_event, "prev_token_1", "next_token_1", Vec::new())
+        .mock_once()
+        .mount()
+        .await;
```
bnjbvr, line 457: to reproduce the failure mode i've talked about above, i think we'd need this mocked endpoint to allow specifying a list of before/after events, and provide some here. Those events should include events that are not in the `root_thread_id` thread, and as such they shouldn't appear in the final timeline.

crates/matrix-sdk-ui/tests/integration/timeline/focus_event.rs
```diff
@@ -418,3 +421,172 @@ async fn test_focused_timeline_doesnt_show_local_echoes() {
...
+            chunk: vec![
+                f.text_msg("Root").sender(user_id).event_id(root_thread_id).into_raw_timeline(),
+            ],
+            prev_batch: Some("prev_token_1".to_owned()),
+            next_batch: None,
+            recursion_depth: None,
+        })
+        .mock_once()
+        .mount()
+        .await;
```
bnjbvr, line 533: I don't think the root event is part of the /relations response for the first related events (because the root event doesn't have the `m.relates_to` field set), so this should return an empty `chunk` (or a chunk with another previous thread response).

crates/matrix-sdk/src/paginators/room.rs
```diff
@@ -339,6 +339,16 @@ impl<PR: PaginableRoom> Paginator<PR> {
...
+
+    /// Returns the current `prev_token` used for pagination.
+    pub fn prev_token(&self) -> PaginationToken {
+        self.tokens.lock().unwrap().previous.clone()
+    }
+
+    /// Returns the current `next_token` used for pagination.
+    pub fn next_token(&self) -> PaginationToken {
+        self.tokens.lock().unwrap().next.clone()
+    }
```
bnjbvr: I think it would be fine to have a single method returning both at once, since they're used both at once in the calling code; that would avoid getting the lock twice (even if it's not contended for sure!)

bindings/matrix-sdk-ffi/src/event.rs
```diff
@@ -51,6 +52,18 @@ impl TimelineEvent {
         };
         Ok(event_type)
     }
+
+    pub fn thread_root_event_id(&self) -> Option<String> {
+        match self.0.deref() {
```
bnjbvr: nit: it's unusual to call `deref` directly, you can do `*self.0` instead, which is more idiomatic

bnjbvr (changes requested):
Thanks, looks very close to landing. Think I've spotted a bug, so I'd like a confirmation first, and then we can land this :)

crates/matrix-sdk-ui/src/timeline/controller/mod.rs
jmartinesp: > if the any paginator is set to a Threaded instance, then we should run a first round of /relations, and use the resulting events here, instead

I'm not sure I understand how that would work: 
- `/relations` runs using the thread root, not the focused event id, so we need some way to say 'start looking here', and it always paginates forwards or backwards.
- If we don't use the tokens I'm not sure what would happen, I guess it'll start from the root?
- And if we provide tokens, the results of paginating backwards and forwards won't contain the previously loaded events, I think. I guess we could maybe do a forward pagination from the prev token taken from `/context`?

Can't we iterate through the returned events and add only those that are part of the thread, since we already got them, by using `extract_thread_root`?

bnjbvr: > Can't we iterate through the returned events and add only those that are part of the thread, since we already got them, by using extract_thread_root?

Then, the first round of /relations would run into duplicates, which is really not wanted. Plus, it's extra work, since a threaded timeline wouldn't have to filter out events that are outside the thread, by default; it's implied from the data source that all the events belong to the thread.

But I think there's a misunderstanding, I'm not suggesting to start only with the /relations. You start with /context, and after figuring out the target event is in a thread, you continue with /relations and the tokens. Once you've got the result, you use them in the `replace_with_initial_remote_events` call.

Otherwise, if /context returned [event A in thread, event B not in thread, event C in thread], then B would show up, while it should not; this is the bug I alluded to in the previous comment.

jmartinesp: > Then, the first round of /relations would run into duplicates, which is really not wanted. Plus, it's extra work, since a threaded timeline wouldn't have to filter out events that are outside the thread, by default; it's implied from the data source that all the events belong to the thread.

I think I'm misunderstanding how the pagination works then. Let me know if this is wrong:

If I ask for item `A` and 5 events using `/context`, I'll receive:

1. Event A.
2. 2 items before A, 2 items after A if available.
3. 2 tokens, `before_context` and `after_context`.

Event A will definitely go into the timeline. Any of the 2 items before or after *could* go into the timeline, that's why I said we could check if they are part of the thread before adding them. Using the `before_context` with backwards pagination will give you events before the results of this `/context` call and with forward pagination and `after_context` you'd get the ones after it. 

---

I think like that we'd have all the events we need, and we wouldn't skip any by mistake - if my understanding of how this works is accurate. When you said you wanted to use `/relations` instead using the results of `/context` did you mean using i.e. the `before_context` and forward pagination so we'd also get the focused event (or `after_context` and backwards pagination)?

bindings/matrix-sdk-ffi/src/event.rs
jmartinesp: I think it's because if we just use `*` we move the boxed value and we'd have to use `&*` instead. I just copied it from the method above, which returns the `TimelineEventType`.

I can replace the deref with `&*` if that's better.

crates/matrix-sdk/src/paginators/thread.rs
jmartinesp: Sure, added the accessor in the latest changes.

crates/matrix-sdk/CHANGELOG.md
```diff
@@ -52,6 +52,8 @@ All notable changes to this project will be documented in this file.
 
 - [**breaking**] `OAuth::login` now allows requesting additional scopes for the authorization code grant.
   ([#5395](https://github.com/matrix-org/matrix-rust-sdk/pull/5395))
+- [**breaking**] `ThreadedEventsLoader::new` now takes optional `prev_token` and `next_token` parameters ([#5678](https://github.com/matrix-org/matrix-rust-sdk/pull/5678).
+- Make `PaginationTokens` `pub(crate)`, as well as its `previous` and `next` tokens so they can be assigned from other files. Also added public `Paginator::prev_token` and `Paginator::next_token` methods to be able to fetch the currently used tokens from other crates ([#5678](https://github.com/matrix-org/matrix-rust-sdk/pull/5678).
```
bnjbvr: The last sentence isn't true anymore, right?

crates/matrix-sdk-ui/src/timeline/controller/mod.rs
```diff
@@ -557,10 +681,16 @@ impl<P: RoomDataProvider> TimelineController<P> {
             | TimelineFocusKind::Thread { .. } => {
                 return Err(PaginationError::NotSupported);
             }
-            TimelineFocusKind::Event { paginator, .. } => paginator
-                .paginate_backward(num_events.into())
-                .await
-                .map_err(PaginationError::Paginator)?,
+            TimelineFocusKind::Event { paginator, .. } => {
+                if let Some(paginator) = paginator.get() {
```
bnjbvr: I think I'd do `paginator.get().expect("The paginator must be set at this point")`, and remove the `NotInstantiated` error variant altogether. If a timeline has been successfully built, the paginator instance *will* be set, or the timeline would be in a partially-built state.

crates/matrix-sdk-ui/src/timeline/controller/mod.rs
```diff
@@ -587,10 +717,16 @@ impl<P: RoomDataProvider> TimelineController<P> {
             | TimelineFocusKind::PinnedEvents { .. }
             | TimelineFocusKind::Thread { .. } => return Err(PaginationError::NotSupported),
 
-            TimelineFocusKind::Event { paginator, .. } => paginator
-                .paginate_forward(num_events.into())
-                .await
-                .map_err(PaginationError::Paginator)?,
+            TimelineFocusKind::Event { paginator, .. } => {
+                if let Some(paginator) = paginator.get() {
```
bnjbvr: ditto here

crates/matrix-sdk-ui/src/timeline/controller/mod.rs
```diff
@@ -145,8 +216,11 @@ impl<P: RoomDataProvider> TimelineFocusKind<P> {
     /// Whether to hide in-thread events from the timeline.
     fn hide_threaded_events(&self) -> bool {
         match self {
-            TimelineFocusKind::Live { hide_threaded_events }
-            | TimelineFocusKind::Event { hide_threaded_events, .. } => *hide_threaded_events,
+            TimelineFocusKind::Live { hide_threaded_events } => *hide_threaded_events,
+            TimelineFocusKind::Event { paginator } => paginator
+                .get()
+                .map(|paginator| paginator.hide_threaded_events())
+                .unwrap_or_default(),
```
bnjbvr: I'm not a fan to use "default" for booleans, let's explicitly say "false":

```suggestion
                .map_or(false, |paginator| paginator.hide_threaded_events())
```

crates/matrix-sdk-ui/src/timeline/controller/mod.rs
```diff
@@ -125,6 +126,76 @@ pub(in crate::timeline) enum TimelineFocusKind<P: RoomDataProvider> {
     },
 }
 
+#[derive(Debug)]
+pub(in crate::timeline) enum AnyPaginator<P: RoomDataProvider> {
+    Unthreaded { paginator: Paginator<P>, hide_threaded_events: bool },
```
bnjbvr: Can you keep the doc comment for `hide_threaded_events`?

crates/matrix-sdk-ui/src/timeline/controller/mod.rs
```diff
@@ -396,25 +468,77 @@ impl<P: RoomDataProvider> TimelineController<P> {
...
 
+                let thread_root_event_id = start_from_result
+                    .events
+                    .iter()
+                    .find(
+                        |event| {
+                            if let Some(id) = event.event_id() { id == *event_id } else { false }
+                        },
+                    )
+                    .and_then(|event| extract_thread_root(event.raw()));
```
bnjbvr, line 491: Could you use `find_map` here?

crates/matrix-sdk-ui/src/timeline/controller/mod.rs
```diff
@@ -396,25 +468,77 @@ impl<P: RoomDataProvider> TimelineController<P> {
...
+
+                let _ = paginator.set(match thread_root_event_id {
+                    Some(root_id) => {
+                        let tokens = event_paginator.tokens();
+                        AnyPaginator::Threaded(ThreadedEventsLoader::new(
+                            self.room_data_provider.clone(),
+                            root_id,
+                            tokens.previous,
+                            tokens.next,
```
bnjbvr: I think it'd be nicer if `ThreadedEventsLoader`'s ctor would take the `tokens` as a single variable; then there's no risk to be confused as to which must be passed first (is it previous first, or next first?)

crates/matrix-sdk-ui/src/timeline/controller/mod.rs
```diff
@@ -396,25 +468,77 @@ impl<P: RoomDataProvider> TimelineController<P> {
...
+                match paginator.get() {
+                    Some(AnyPaginator::Unthreaded { .. }) => {
+                        self.replace_with_initial_remote_events(
+                            events.into_iter(),
+                            RemoteEventOrigin::Pagination,
+                        )
+                        .await;
+                    }
+                    Some(AnyPaginator::Threaded(threaded_events_loader)) => {
+                        let events_in_thread = events
```
bnjbvr, line 522: Can you add a tiny comment saying that we only keep events from the /context response, which belong to the target thread?

crates/matrix-sdk-ui/src/timeline/controller/mod.rs
```diff
@@ -396,25 +468,77 @@ impl<P: RoomDataProvider> TimelineController<P> {
...
+
                 let has_events = !start_from_result.events.is_empty();
+                let events = start_from_result.events;
 
-                self.replace_with_initial_remote_events(
-                    start_from_result.events.into_iter(),
-                    RemoteEventOrigin::Pagination,
-                )
-                .await;
+                match paginator.get() {
```
bnjbvr: To remove the `None` arm, can you use `paginator.get().expect("...")` as suggested in other places, please?

crates/matrix-sdk-ui/src/timeline/controller/state_transaction.rs
```diff
@@ -410,10 +410,14 @@ impl<'a, P: RoomDataProvider> TimelineStateTransaction<'a, P> {
...
 
-            TimelineFocusKind::Event { hide_threaded_events, paginator: _ } => {
+            TimelineFocusKind::Event { paginator } => {
                 // If the timeline's filtering out in-thread events, don't add items for
                 // threaded events.
-                if thread_root.is_some() && *hide_threaded_events {
+                let hide_threaded_events = paginator
+                    .get()
+                    .map(|paginator| paginator.hide_threaded_events())
+                    .unwrap_or_default();
```
bnjbvr:
```suggestion
                    .map_or(false, |paginator| paginator.hide_threaded_events())
```

crates/matrix-sdk-ui/CHANGELOG.md
```diff
@@ -6,6 +6,12 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased] - ReleaseDate
 
+### Refactor
+
+- [**breaking**] `TimelineFocusKind::Event`'s `paginator` is now an instance of an `AnyPaginator` enum
+  which can be either `Paginator` or `ThreadedEventsLoader`, this way we can paginate events using any 
+  of those ([#5678](https://github.com/matrix-org/matrix-rust-sdk/pull/5678)).
```
bnjbvr: Is it changelog worthy? as in, does it leak through our public APIs? If not, then I'd remove it; internal changes aren't that interesting.

…But since you're around, you could tell that the `Events` focus will now properly use thread-pagination when the focused event is in a thread :partying_face:

bindings/matrix-sdk-ffi/src/event.rs
bnjbvr: Alright, let's keep it like this for consistency.

bindings/matrix-sdk-ffi/src/event.rs
```diff
@@ -51,6 +52,18 @@ impl TimelineEvent {
         };
         Ok(event_type)
     }
+
+    pub fn thread_root_event_id(&self) -> Option<String> {
```
bnjbvr, line 56: Can you add a small doc comment here?

crates/matrix-sdk-ui/tests/integration/timeline/focus_event.rs
```diff
@@ -418,3 +421,176 @@ async fn test_focused_timeline_doesnt_show_local_echoes() {
...
+
+    let (items, mut timeline_stream) = timeline.subscribe().await;
+    assert_eq!(items.len(), 2);
+    assert!(items[0].is_date_divider());
+    assert_eq!(items[1].as_event().unwrap().content().as_message().unwrap().body(), "Hey");
+
+    // We paginate back once
+    server
+        .mock_room_relations()
+        .ok(RoomRelationsResponseTemplate {
```
bnjbvr, line 488: (light suggestion) Here and below, could we match against the expected pagination token here?

bnjbvr (approved):
Thanks! Approving so we can move forward. A bunch of nits, but nothing that should be hard to address or that should require another round of review :)
Good job!

crates/matrix-sdk/CHANGELOG.md
jmartinesp: Yes, I need to double check the changelogs. Thanks for the reminder!

crates/matrix-sdk-ui/src/timeline/controller/mod.rs
jmartinesp: I was using it at the beginning, but it looked way more verbose and didn't seem to help legibility.

PR #5680, opened 2025-09-17
fix(ui): The `recency` room list sorter stops using the latest event's timestamp

Hywan:
This patch updates the `recency` room list sorter to no longer use the `LatestEventValue::timestamp` method. It keeps using the `Room::recency_stamp` for the moment, as it was the case before. This patch is a test to try finding the problem in some Matrix clients where the room list becomes unusable. We suspect it's because of this patch sorter.

---


pixlwave (approved):
LGTM from what I understand, thanks!

PR #5681, opened 2025-09-17
feat(multiverse): Add backfilling indexing to multiverse on startup.

dragonfly1033:
This PR adds bulk processing to the search crate, uses this in a blocking task in multiverse that indexes all messages in all rooms from the event cache and the server.

It also has mild refactoring for search related functions.
 

crates/matrix-sdk/Cargo.toml
```diff
@@ -16,7 +16,7 @@ features = ["docsrs"]
 rustdoc-args = ["--cfg", "docsrs", "--generate-link-to-definition"]
 
 [features]
-default = ["e2e-encryption", "automatic-room-key-forwarding", "sqlite", "native-tls"]
+default = ["e2e-encryption", "automatic-room-key-forwarding", "sqlite", "native-tls", "experimental-search"]
```
stefanceriu: Are you sure we want to enable this by default? 🤔

crates/matrix-sdk/src/search_index/mod.rs
```diff
@@ -138,6 +158,140 @@ impl SearchIndexGuard<'_> {
...
+async fn get_most_recent_edit(
+    cache: &RoomEventCache,
+    original: &EventId,
+) -> Option<OriginalSyncRoomMessageEvent> {
+    use ruma::events::{relation::RelationType, AnySyncTimelineEvent};
+
+    let Some((original_ev, related)) =
+        cache.find_event_with_relations(original, Some(vec![RelationType::Replacement])).await
+    else {
+        warn!("Couldn't find relations for {}", original);
```
stefanceriu: I would assume this is expected most of the times, is the warn worth it?

crates/matrix-sdk/src/search_index/mod.rs
```diff
@@ -138,6 +158,140 @@ impl SearchIndexGuard<'_> {
...
+}
+
+/// Return a [`RoomIndexOperation::Edit`] or [`RoomIndexOperation::Remove`]
+/// depending on the message.
+async fn handle_room_redaction(
+    event: SyncRoomRedactionEvent,
+    cache: &RoomEventCache,
+    rules: &RedactionRules,
+) -> Option<RoomIndexOperation> {
+    if let Some(redacted_event_id) = event.redacts(rules) {
```
stefanceriu: Can you maybe use let chains here or let elses to reduce the indentation levels?

crates/matrix-sdk-search/src/index.rs
```diff
@@ -196,35 +215,161 @@ impl RoomIndex {
...
+        Ok(())
+    }
+
+    /// Bulk execute [`RoomIndexOperation`]s
+    ///
+    /// If an error occurs in the batch it retries 5 times if possible.
+    ///
+    /// This which will add/remove/edit an events in the index based on the
+    /// operations.
+    pub fn bulk_execute(&mut self, operations: Vec<RoomIndexOperation>) -> Result<(), IndexError> {
```
stefanceriu, line 346: I would've expected this to be async 🤔

crates/matrix-sdk-search/src/index.rs
```diff
@@ -56,8 +56,8 @@ pub struct RoomIndex {
     schema: RoomMessageSchema,
     query_parser: QueryParser,
     room_id: OwnedRoomId,
-    uncommited_adds: HashSet<OwnedEventId>,
-    uncommited_removes: HashSet<OwnedEventId>,
+    uncommitted_adds: HashSet<OwnedEventId>,
+    uncommitted_removes: HashSet<OwnedEventId>,
```
stefanceriu, line 60: Would've loved to see small bits like these in separate commits, there's quite a few in here.

labs/multiverse/src/main.rs
```diff
@@ -95,6 +104,8 @@ pub enum GlobalMode {
     CreateRoom { view: CreateRoomView },
     /// Mode where we have opened create room screen
     Searching { view: SearchingView },
+    /// Mode where we have opened create room screen
```
stefanceriu: Copy pasta

Also.. "_the_ create room screen"?

labs/multiverse/src/main.rs
```diff
@@ -206,15 +223,27 @@ impl App {
 
         let room_view = RoomView::new(client.clone(), timelines.clone(), status.handle());
 
+        let (indexing_send, indexing_recv) = channel::<(bool, IndexingMessage)>(1024);
+        let client_2 = client.clone();
```
stefanceriu: Can you not clone it directly in the call to `App::indexing_task`?

labs/multiverse/src/main.rs
```diff
@@ -206,15 +223,27 @@ impl App {
 
         let room_view = RoomView::new(client.clone(), timelines.clone(), status.handle());
 
+        let (indexing_send, indexing_recv) = channel::<(bool, IndexingMessage)>(1024);
```
stefanceriu: Not a fan of these shortened names to be honest

labs/multiverse/src/widgets/search/mod.rs
stefanceriu, line 1: Same here, could've/should've been a different commit

labs/multiverse/src/widgets/room_view/timeline.rs
```diff
@@ -25,14 +25,21 @@ impl<'a> TimelineView<'a> {
...
     list_index_to_item_index: Vec<usize>,
 }
 
+impl Default for TimelineListState {
+    fn default() -> Self {
+        let mut state = ListState::default();
+        state.select_last();
+        Self { state, list_index_to_item_index: Vec::default() }
+    }
+}
```
stefanceriu, line 41: What's this about?

labs/multiverse/src/main.rs
```diff
@@ -314,6 +343,162 @@ impl App {
...
+            count += no_of_events;
+            let _ = update_sender.send((false, IndexingMessage::Progress(count))).await;
+        }
+
+        // Now index from the server
+        debug!("Start indexing from the server.");
+
+        let batch_size = 25;
+
+        let mut rooms = client.rooms_filtered(RoomStateFilter::JOINED.union(RoomStateFilter::LEFT));
```
stefanceriu: Can one still paginate left rooms? 😮

crates/matrix-sdk-search/src/index.rs
dragonfly1033: Yep, this was going to be another PR. Index writes and commits are blocking so the plan was to make them non-blocking with `spawn_blocking` and await them (which would then make the API's async).

crates/matrix-sdk/src/search_index/mod.rs
dragonfly1033: Good point, probably not. I'll make it a `debug!`.

labs/multiverse/src/main.rs
dragonfly1033: How did I not realise that :joy:

labs/multiverse/src/widgets/room_view/timeline.rs
dragonfly1033: This is to send the timeline view to the bottom (most recent message) after the repeated backpaginations in the indexing task. The default would otherwise be to have the view at the top of messagee history.

crates/matrix-sdk/Cargo.toml
dragonfly1033: No! Sorry, forgot to remove it.

labs/multiverse/src/widgets/room_view/timeline.rs
stefanceriu: Oh I see, okay then 👍

crates/matrix-sdk-search/src/index.rs
```diff
@@ -196,35 +215,161 @@ impl RoomIndex {
...
+        writer: &mut SearchIndexWriter,
+        operation: &RoomIndexOperation,
+        retries: usize,
+    ) -> Result<(), IndexError> {
+        let mut num_tries = 0;
+
+        while let Err(err) = self.execute_impl(writer, operation) {
+            if num_tries == retries {
+                return Err(err);
+            }
```
poljar, line 294: We should probably reuse the `backon` crate for this kind of logic. It's already in our dependency tree for the network retry logic.

But ok, there's not much time left for this and it isn't particularly important.

labs/multiverse/src/main.rs
```diff
@@ -313,6 +342,162 @@ impl App {
...
+        client: Client,
+        update_sender: Sender<(bool, IndexingMessage)>,
+        sync_service: Arc<SyncService>,
+    ) {
+        let timeout_secs = 30;
+
+        let start_time = Instant::now();
+
+        // Spin until there are rooms to index
+        while (Instant::now() - start_time) < Duration::from_secs(timeout_secs) {
```
poljar: wat :duck: 

`RoomListService::state()` returns a subscriber which you can listen to. It will wake you up when a new state is triggered.

```rust
let mut subscriber = room_list_service.state();

while let Some(update) = subscriber.next().await {
    ...
}
```

labs/multiverse/src/main.rs
```diff
@@ -313,6 +342,162 @@ impl App {
...
+                }
+            } else {
+                idx = (idx + 1) % rooms.len();
+            }
+        }
+
+        while let Err(err) = update_sender.send((true, IndexingMessage::Progress(count))).await {
+            debug!("couldn't send final update {err}, trying again.");
+        }
+    }
```
poljar: This function is a bit on the long side. Would be nice to split this out into the logical parts you already noticed.

Namely, helper functions for these parts would have been nice:

1. Wait for the sync to calm down
2. Index from the cache
3. Index from the server

This would split out the complexity of the function into smaller and more consumable parts.

labs/multiverse/src/main.rs
```diff
@@ -313,6 +342,208 @@ impl App {
...
+                        while let Err(e) =
+                            update_sender.send((true, IndexingMessage::Progress(0))).await
+                        {
+                            debug!("Failed to send final message, trying again: {e:?}");
+                        }
+                        return Ok(());
+                    }
+                    _ => {
+                        debug!("Sync service not running. Waiting to start indexing. {state:?}");
+                        sleep(Duration::from_secs(5)).await;
```
poljar: No need to sleep here anymore, no?

labs/multiverse/src/main.rs
```diff
@@ -313,6 +342,208 @@ impl App {
...
 
+    async fn wait_for_room_sync(
+        update_sender: &Sender<(bool, IndexingMessage)>,
+        sync_service: Arc<SyncService>,
+    ) -> Result<(), ()> {
+        let mut sync_subscriber = sync_service.room_list_service().state();
+
+        // Spin until there are rooms to index
+        loop {
+            match timeout(Duration::from_secs(30), sync_subscriber.next()).await {
```
poljar: I think you'll want to apply the timeout to the `wait_for_room_sync()` method, this way we're waiting 30 seconds for each update we receive over the subscriber.

PR #5603, opened 2025-09-01
IndexedDB: Add initial types and database migrations for `EventCacheStoreMedia` backend

mgoldenberg:
## Background

This pull request is part of a series of pull requests to add a full IndexedDB implementation of the `EventCacheStore` (see #4617, #4996, #5090, #5138, #5226, #5274, #5343, #5384, #5406, #5414, #5497, #5506, #5540, #5574). This particular pull request adds types and migrations that will support storing media in IndexedDB via `EventCacheStoreMedia`.

### Queries

Based on the SQLite-backed implementation of `EventCacheStoreMedia`, the following queries for interacting with stored media should be supported.

1. Operate on media by [`MediaRequestParameters`](https://docs.rs/matrix-sdk/latest/matrix_sdk/media/struct.MediaRequestParameters.html)
2. Operate on media by [`MediaSource`](https://docs.rs/matrix-sdk/latest/matrix_sdk/ruma/events/room/enum.MediaSource.html#)
3. Operate on media by last time it was accessed and whether it obeys the [`MediaRetentionPolicy`](https://docs.rs/matrix-sdk/latest/matrix_sdk/media/struct.MediaRetentionPolicy.html). 
4. Operate on media by the size of its contents and whether it obeys the [`MediaRetentionPolicy`](https://docs.rs/matrix-sdk/latest/matrix_sdk/media/struct.MediaRetentionPolicy.html).
5. Operate on media by the last time it was accessed, the size of its contents, and whether it obeys the [`MediaRetentionPolicy`](https://docs.rs/matrix-sdk/latest/matrix_sdk/media/struct.MediaRetentionPolicy.html).

### Types

Two types have been added - `Media` and `MediaMetadata` to represent each media object in the database.

```rust
pub struct Media {
    pub metadata: MediaMetadata,
    pub content: Vec<u8>,
}

pub struct MediaMetadata {
    pub request_parameters: MediaRequestParameters,
    pub last_access: Duration,
    pub ignore_policy: IgnoreMediaRetentionPolicy,
}
```

The fields in `MediaMetadata` along with the size of `Media::content` after it is serialized, are used to derive keys for indexing `Media` in service of the queries above.

### Schema

In order to support the queries above, the following object stores and indices have been created.

- `MEDIA` - Object Store
[...]

PR #5682, opened 2025-09-17
IndexedDB: add functionality to get, add, replace, and remove media via `MediaStore`

mgoldenberg:
## Background

This pull request is part of a series of pull requests to add a full IndexedDB implementation of the `EventCacheStore` and  `MediaStore` (see #4617, #4996, #5090, #5138, #5226, #5274, #5343, #5384, #5406, #5414, #5497, #5506, #5540, #5574, #5603, #5676). This particular pull request adds IndexedDB-backed implementations for retrieving, storing, replacing, and removing media in IndexedDB via `MediaStore`.

### IndexedDB implementations of top-level `MediaStore` functions

The overarching change is the addition of IndexedDB-backed implementations for the following functions.

- `MediaStore::add_media_content`
- `MediaStore::replace_media_key`
- `MediaStore::get_media_content`
- `MediaStore::remove_media_content`
- `MediaStore::get_media_content_for_uri`
- `MediaStore::remove_media_content_for_uri`
- `MediaStoreInner::add_media_content_inner`
- `MediaStoreInner::get_media_content_inner`
- `MediaStoreInner::get_media_content_for_uri_inner`

Additionally, there were a number of changes that needed to be made in order to support the implementations above.

### Index media object store by [`MxcUri`](https://docs.rs/matrix-sdk/latest/matrix_sdk/ruma/struct.MxcUri.html) rather than [`MediaSource`](https://docs.rs/matrix-sdk/latest/matrix_sdk/ruma/events/room/enum.MediaSource.html)

Prior to this pull request, the `MEDIA` object store contained an index that allowed querying by `MediaSource`; however, what was actually needed was to query by `MxcUri` given the expectations of the `*_for_uri` functions above.

So, the `MEDIA_SOURCE` index has been replaced with a `MEDIA_URI` index, which allows querying by the `MxcUri` of the associated media.

### Add `UnixTime` type to better represent [`Media::last_access`](https://github.com/matrix-org/matrix-rust-sdk/blob/62d2d0ff94c1b132030d3f5d9e150a8eab24ee1d/crates/matrix-sdk-indexeddb/src/media_store/types.rs#L52)

In order to better represent time relative to the Unix Epoch, the following type was added.

```rust
pub enum UnixTime {
    BeforeEpoch(Duration),
    AfterEpoch(Duration),
}
```

This type adds convenience as it is possible to convert a `SystemTime` to a `UnixTime` infallibly, given that it supports times before and after the Unix Epoch.  

## Future Work

- Add implementation of remaining functions in `MediaStore` and `MediaStoreInner`
- Refactor feature flags

---


Hywan (commented):
It looks good to me. Should we have a migration to handle the `media_source` to `media_uri` change?

mgoldenberg:
> It looks good to me. Should we have a migration to handle the `media_source` to `media_uri` change?

I think it probably isn't necessary, given that none of this is exposed publicly outside the crate. I figure once everything is stabilized, we'll make it all public and then future changes will require migrations.

What do you think?

Hywan (approved):
Good to me! Thank you very much.

PR #5683, opened 2025-09-17
fix(sdk): Use a max recent emoji count of 100

jmartinesp:
Iteration on https://github.com/matrix-org/matrix-rust-sdk/pull/5641.

This prevents the account data from growing indefinitely and makes sure values that aren't recently used can be forgotten.

cc @Johennes 

Signed-off-by:


crates/matrix-sdk/src/test_utils/mocks/mod.rs
```diff
@@ -3553,6 +3553,24 @@ impl<'a> MockEndpoint<'a, UpdateGlobalAccountDataEndpoint> {
...
         MatrixMock { server: self.server, mock }
     }
+
+    /// Returns a mock for a successful global account data event, as long as it
+    /// matches the provided `body`.
+    pub fn ok_with_request_body(
+        self,
+        user_id: &UserId,
+        event_type: GlobalAccountDataEventType,
+        body: Value,
```
bnjbvr: Conceptually, matching on the body is optional. What you could do instead:

- add a new `body: Option<Value>` in the `UpdateGlobalAccountDataEndpoint` struct
- add a new setter function `fn match_body(self, value: Option<Value>) -> Self`,  that sets the `body` to a set value
- when calling the existing `ok` function, if `body` is defined, configure the `mock` to match on it

But overall, I don't think `mock_global_account_data` and `mock_update_global_account_data` are the right fit for the mocking endpoints. In particular, if, as a user of this API, you have to provide the raw JSON `Value` objects, then you have to know what shape it will have. The goal of these mocking APIs was that we don't have to know the shape of the requests/responses, so that we don't have to specify (and duplicate all over the code base) the JSON values themselves. Instead, we'd use dedicated endpoints with high-level APIs, like `fn ok(self, recent_emojis: Vec<(String, UInt)>)`, and this function would return the JSON response in the format the spec requires.

Would you be open to making that change here, please? :pleading_face:

bnjbvr (changes requested):
Thanks, looks good; only requesting changes because of mocking :pleading_face:

crates/matrix-sdk/src/account.rs
```diff
@@ -1188,10 +1192,17 @@ impl Account {
 
         let index = recent_emojis.iter().position(|(unicode, _)| unicode == emoji);
 
+        // Remove the emoji from the list if it was present and get it's `count` value
         let count = if let Some(index) = index { recent_emojis.remove(index).1 } else { uint!(0) };
 
+        // Insert the emoji with the updated count at the start of the list, so it's
+        // considered the most recently used emoji
```
poljar, line 1201: The `insert` is `O(n)`, so if we're worried about the size, perhaps we should truncate before the insert.

poljar (approved):
Left a suggestions, looks good otherwise.

crates/matrix-sdk/src/test_utils/mocks/mod.rs
poljar: Ah, sorry I didn't expect you to review this this late and I'm already familiar with the topic so wanted to get this off your back.

jmartinesp: @bnjbvr I'm not sure I fully follow what you're asking me to do here, just to replace `fn ok_with_request_body` with `fn ok(self, recent_emojis: Vec<(String, UInt)>)`? Or fully replacing `mock_global_account_data` and `mock_update_global_account_data` with specific `mock_...` methods for the ones with recent emojis, and each new account data get/update for an event type should create their own too?

crates/matrix-sdk/src/account.rs
jmartinesp: Sure, I can do that. As for the overhead of using `insert(0, _)`, would it be better if I created a new `Vec` with just the updated item, then appended the previous list to it?  `Vec::append` and `Vec::extend` don't mention their complexity, but I guess they'll also be `O(n)` 🫤 .

crates/matrix-sdk/src/test_utils/mocks/mod.rs
bnjbvr: Well, both :) Having `fn ok(self, recent_emojis: Vec<(String, UInt)>)` on `mock_global_account_data` (resp. update) would be overly specific to the emojis, since there's nothing in the `ok` name that would say it's for emojis. So I'm recommending to have new `mock_get_recent_emojis()` and `mock_update_recent_emojis()` instead :+1:

jmartinesp: Makes sense, let's see what I can do.

crates/matrix-sdk/src/account.rs
```diff
@@ -1192,15 +1193,20 @@ impl Account {
 
         let index = recent_emojis.iter().position(|(unicode, _)| unicode == emoji);
 
+        // Truncate to the max allowed size, which will remove any emojis that
+        // haven't been used in a very long time. This will also ease the pressure on
+        // `remove` and `insert` shifting lots of elements in the list
+        recent_emojis.truncate(MAX_RECENT_EMOJI_COUNT);
```
bnjbvr, line 1195: If the emoji wasn't present in the list before, the recent emojis list ends up having MAX+1 items. I think the number to truncate should be MAX-1 in this case, so this line should be: `if index.is_some() { MAX_RECENT_EMOJI_COUNT } else { MAX_RECENT_EMOJI_COUNT - 1 }`

crates/matrix-sdk/src/account.rs
```diff
@@ -1192,15 +1193,20 @@ impl Account {
...
-        // Insert the emoji with the updated count at the start of the list, so it's
+        // Create a new list with the emoji in the first position, so it's
         // considered the most recently used emoji
         recent_emojis.insert(0, (emoji.to_owned(), count + uint!(1)));
 
-        // Then truncate to the max allowed size, which will remove any emojis that
-        // haven't been used in a very long time
+        // If the item was a new one, the list will now be `MAX_RECENT_EMOJI_COUNT` + 1,
+        // so truncate it again (this is a no-op if it already has the right size)
         recent_emojis.truncate(MAX_RECENT_EMOJI_COUNT);
```
bnjbvr, line 1206: No need to re-truncate after the operation, with my other comment taken into account.

crates/matrix-sdk/src/account.rs
```diff
@@ -1192,15 +1193,20 @@ impl Account {
...
+        // Truncate to the max allowed size, which will remove any emojis that
+        // haven't been used in a very long time. This will also ease the pressure on
+        // `remove` and `insert` shifting lots of elements in the list
+        recent_emojis.truncate(MAX_RECENT_EMOJI_COUNT);
+
         // Remove the emoji from the list if it was present and get it's `count` value
         let count = if let Some(index) = index { recent_emojis.remove(index).1 } else { uint!(0) };
 
-        // Insert the emoji with the updated count at the start of the list, so it's
+        // Create a new list with the emoji in the first position, so it's
```
bnjbvr: Comment's wrong, it doesn't create a new list, does it?

crates/matrix-sdk/src/test_utils/mocks/mod.rs
```diff
@@ -3500,75 +3501,73 @@ impl<'a> MockEndpoint<'a, RoomRelationsEndpoint> {
...
-            )))
-            .and(body_json(body))
-            .respond_with(ResponseTemplate::new(200).set_body_json(()));
+    /// Returns a mock for a successful update of the recent emojis account data
+    /// event. The request body contents should match the provided emoji
+    /// list.
+    #[cfg(feature = "experimental-element-recent-emojis")]
+    pub fn ok(self, user_id: &UserId) -> MatrixMock<'a> {
+        let mock =
+            global_account_data_mock_builder(self.mock, user_id, "io.element.recent_emoji".into())
```
bnjbvr, line 3569: Can we use a `GlobalAccountDataEventType` variant instead of the custom variant here?

crates/matrix-sdk/src/account.rs
jmartinesp: I don't think it's that easy: what if the item was present in the original list (potentially > MAX_RECENT_EMOJI_COUNT) and *after* `MAX_RECENT_EMOJI_COUNT`? We end up with nested ifs and kind of complex logic, which I tried to avoid with the double truncation, which I hope is performant enough.

crates/matrix-sdk/src/account.rs
jmartinesp: Oops, true. I had some code from previous attempt to optimise this that made me change this comment, but I reverted it. Thanks for spotting it!

crates/matrix-sdk/src/test_utils/mocks/mod.rs
jmartinesp: Not unless we instantiate a dummy `RecentEmojisContent` instance so we can call `content.event_type()`, since the type is not part of ruma (yet) so that means it'll be a `GlobalAccountDataEventType::_Custom(here_be_dragons)` case.

crates/matrix-sdk/src/account.rs
jmartinesp: Also, this ensures the shifting done in the `remove` operation is done in `O(n)` with `n <= 100`, instead of whatever the original length of the list was.

PR #5684, opened 2025-09-17
fix(ui): Fix performance problem with the `latest_event` and `recency` room list sorters

Hywan:
WIP


crates/matrix-sdk-ui/src/room_list_service/room_list.rs
```diff
@@ -353,3 +361,84 @@ impl RoomListDynamicEntriesController {
...
+/// Why caching data? [`RoomInfo`] is behind a lock. Every time a filter or a
+/// sorter calls a method on [`matrix_sdk::Room`], it's likely to hit the lock
+/// in front of [`RoomInfo`]. It creates a big contention. By caching the data,
+/// it avoids hitting the lock, improving the performance greatly.
+///
+/// Data are refreshed in `merge_stream_and_receiver` (private function).
+///
+/// [`RoomInfo`]: matrix_sdk::RoomInfo
+#[derive(Clone, Debug)]
+pub struct Room {
```
andybalaam: I really don't like things with the same name in different namespaces. Is there any chance we can find a different name? `RoomListItem` is the best I can come up with right now, but it's not great.

Hywan: I've adopted `RoomListItem`. Thanks for the suggestion!

PR #5685, opened 2025-09-18
chore(ffi): rejigger recent emoji code around so as to work with defaut features disabled

bnjbvr:
For some reason, running `cargo xtask ci clippy` locally would now fail, complaining that the recent emoji functions didn't exist, in the FFI layer. I suspect it's because some of the uniffi derive macro to export functions incorrectly propagates the `cfg` guards; so a solution is to move all this code under a new mod, that's enabled if and only if the feature's enabled.

@jmartinesp can you try it around and confirm the emoji functions are still available for you, when the experimental feature's enabled, please?

### DIFF
--- a/bindings/matrix-sdk-ffi/src/client.rs
+++ b/bindings/matrix-sdk-ffi/src/client.rs
@@ -1745,37 +1745,42 @@ impl Client {
             }
         }))))
     }
-
-    /// Adds a recently used emoji to the list and uploads the updated
-    /// `io.element.recent_emoji` content to the global account data.
-    #[cfg(feature = "experimental-element-recent-emojis")]
-    pub async fn add_recent_emoji(&self, emoji: String) -> Result<(), ClientError> {
-        Ok(self.inner.account().add_recent_emoji(&emoji).await?)
-    }
-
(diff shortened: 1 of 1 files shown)


jmartinesp:
I was using `cargo xtask ci clippy` locally before I submitted the PR, maybe I'm behind in the Rust version we should be using?

> can you try it around and confirm the emoji functions are still available for you, when the experimental feature's enabled, please?

I built some bindings locally and the 2 methods still seem to be available.

---

> I was using `cargo xtask ci clippy` locally before I submitted the PR, maybe I'm behind in the Rust version we should be using?

Ah, I see, new Rust version just dropped (1.90).

bnjbvr:
Hmm, I was running into this failure <del>before it was cool</del> before the new Rust release, and I was surely using 1.89. Wonder what would make it so that you wouldn't run into it. Were you using stable 1.89 before?

jmartinesp:
`1.88` actually 😅 . Maybe it was a `1.89+` issue.

bnjbvr:
So, I still need this to run `cargo xtask fixup clippy`, but I don't need it with `cargo xtask ci clippy`, fwiw.

PR #5687, opened 2025-09-18
feat(push): add experimental support for applying "Do not Disturb" notification settings as per MSC4359

Johennes:
This adds partial support for https://github.com/matrix-org/matrix-spec-proposals/pull/4359 and suppresses push notifications when a room is set to "Do not Disturb" (DnD). The DnD setting is sourced from a dedicated global account data event.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2253,16 +2253,13 @@ dependencies = [
 
 [[package]]
 name = "html5ever"
-version = "0.29.0"
+version = "0.35.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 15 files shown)


bnjbvr (commented):
Hi @Johennes, thanks for the contribution! I'm not against it in principle, but we're a bit concerned that adding support for experimental features like this adds significantly to the maintenance overhead of matrix-rust-sdk, and as such feel like it's probably a bit early to merge support for this into mainline, even when guarded by feature flags. Perhaps you could keep it in a fork for now, and we can revisit once the MSC (or the other) gets a bit more traction?

Johennes:
Oh. 😯 This actually seemed relatively minimal to me, especially when compared to its predecessor #5441. I understand the rationale though and keeping this in a fork is fine. If it's ok, I'll leave this pull request open in draft state because it's currently the only public implementation of the MSC.

PR #5688, opened 2025-09-18
perf: make loading the last chunk of a linked chunk faster using a sql index

bnjbvr:
(perf numbers incoming)


bnjbvr:
Nope, turns out that loading the latest chunk is already quite fast! In a context where we have 1,000 rooms with 10,000 events in each room, before the new INDEX:

|                      | Lower bound    | Estimate      | Upper bound   |
|----------------------|----------------|---------------|---------------|
| **Slope**            | 44.512 µs      | 46.765 µs     | 51.965 µs     |
| **Throughput**       | 192.44 Melem/s | 213.84 Melem/s| 224.66 Melem/s|
| **R²**               | 0.6061338      | 0.6357869     | 0.5042838     |
| **Mean**             | 47.627 µs      | 54.614 µs     | 63.388 µs     |
| **Std. Dev.**        | 4.8466 µs      | 13.565 µs     | 18.699 µs     |
| **Median**           | 44.346 µs      | 51.222 µs     | 61.903 µs     |
| **MAD**              | 1.3159 µs      | 9.5452 µs      | 18.359 µs     |

So it's not slow in the first place, so the index would not be useful. This leaves us with the more likely hypothesis that the event cache store lock was super contended in this user's rageshake. Closing.

PR #5689, opened 2025-09-18
ffi: Expose is_direct on SpaceRoom.

pixlwave:
I think this was missed in #5663


bnjbvr:
(You might need a rebase over the latest main to fix the unrelated CI failure)

PR #5695, opened 2025-09-19
fix(ui): `room_list::sorters::recency` is no longer based on 2 data

Hywan:
This patch fixes an issue where the `recency` sorter is based on either the latest event's timestamp, or the room recency stamp. This cannot work with a sort algorithm as the position of a particular room can be different based on what it is compared to (i.e. if the rooms have a latest event value or not).

This patch updates the `recency` sorter to only use the recency stamp for now, as the latest event is not yet computed for all rooms.


pixlwave (approved):
👍

PR #5698, opened 2025-09-22
feat(ui): The `recency` sorter handles recency stamp _and_ latest event's timestamp

Hywan:
This patch revisits a feature we have disabled a couple of days ago: the `recency` sorter was initially only supporting the recency stamp, then later the recency stamp _and_ the latest event's timestamp. It was however buggy and we had to revert it. Now it's time to re-introduce it but with a different approach.

The previous rules were:

1. if two rooms have a latest event, use their latest event's timestamps as their _scores_,
2. if one of the rooms has a latest event, use the recency stamp as their _scores_ for both rooms.

Rule 2 was buggy because one room was sometimes using its latest event's timestamp, and sometimes its recency stamp, based on what it was compared to. It was an error!

The new rules are the following:

1. unchanged
2. if one room has a latest event, use its latest event's timestamp as its _score_, and use no _score_ for the other room,
3. if two rooms have NO latest event, use the recency stamp as their _scores_.

It means that a room with no latest event will always be sorted _after_ a room with a latest event. It can feel cruel, but it should be an edge case. When a room is synchronised, it should receive events, which should trigger the computation of a latest event.

Note that this patch also renames _rank_ to _score_, as I consider it's a better vocabulary. It could be confusing to use _rank_ as one can expect all rooms to be indexed and get a rank, but it's not the case. _Score_ sounds better.

---

* Follow up of https://github.com/matrix-org/matrix-rust-sdk/pull/5680
* Follow up of https://github.com/matrix-org/matrix-rust-sdk/pull/5695

### DIFF
--- a/crates/matrix-sdk-ui/src/room_list_service/sorters/recency.rs
+++ b/crates/matrix-sdk-ui/src/room_list_service/sorters/recency.rs
@@ -16,11 +16,11 @@ use std::cmp::Ordering;
 
 use super::{RoomListItem, Sorter};
 
-fn cmp<F>(timestamps: F, left: &RoomListItem, right: &RoomListItem) -> Ordering
+fn cmp<F>(scores: F, left: &RoomListItem, right: &RoomListItem) -> Ordering
 where
-    F: Fn(&RoomListItem, &RoomListItem) -> (Option<Rank>, Option<Rank>),
+    F: Fn(&RoomListItem, &RoomListItem) -> (Option<Score>, Option<Score>),
 {
-    match timestamps(left, right) {
+    match scores(left, right) {
         (Some(left), Some(right)) => left.cmp(&right).reverse(),
 
         (Some(_), None) => Ordering::Less,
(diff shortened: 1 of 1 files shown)


jmartinesp:
AFAICT this works well in a real client. The sorting seemed right and I couldn't reproduce the empty room list issue - not that I could reproduce that easily.

crates/matrix-sdk-ui/src/room_list_service/sorters/recency.rs
```diff
@@ -41,33 +41,84 @@ where
...
+/// [`RoomInfo::recency_stamp`], but we **must never** mix both. The
+/// `RoomInfo::recency_stamp` is not a timestamp, while `LatestEventValue` uses
+/// a timestamp.
+fn extract_score(left: &RoomListItem, right: &RoomListItem) -> (Option<Score>, Option<Score>) {
+    // Warning 1.
+    //
+    // Be careful. This method is called **a lot** in the context of a sorter. Using
+    // `Room::new_latest_event` would be dramatic as it returns a clone of the
+    // `LatestEventValue`. It's better to use the more specific method
+    // `Room::new_latest_event_timestamp`, where the value is cached in this
```
bnjbvr: It's better to do so, but the code isn't doing so; the code comment seems outdated.

crates/matrix-sdk-ui/src/room_list_service/sorters/recency.rs
```diff
@@ -41,33 +41,84 @@ where
...
-fn extract_rank(left: &RoomListItem, right: &RoomListItem) -> (Option<Rank>, Option<Rank>) {
-    (left.cached_recency_stamp.map(Into::into), right.cached_recency_stamp.map(Into::into))
+/// Extract the recency _score_ from either the [`RoomInfo::new_latest_event`]
+/// or from [`RoomInfo::recency_stamp`].
+///
+/// We must be very careful to return data of the same nature: either a
+/// _score_ from the [`LatestEventValue`]'s timestamp, or from the
+/// [`RoomInfo::recency_stamp`], but we **must never** mix both. The
+/// `RoomInfo::recency_stamp` is not a timestamp, while `LatestEventValue` uses
+/// a timestamp.
```
bnjbvr, line 57: What about using an `enum` for the `Score`? Then we could use static type checks that we're not comparing timestamps with bumpstamps? And this entire caveat would disappear from existence.

crates/matrix-sdk-ui/src/room_list_service/sorters/recency.rs
```diff
@@ -41,33 +41,84 @@ where
 /// [`RoomInfo::new_latest_event`]: matrix_sdk_base::RoomInfo::new_latest_event
 /// [`LatestEventValue::None`]: matrix_sdk_base::latest_event::LatestEventValue::None
 pub fn new_sorter() -> impl Sorter {
-    let ranks = |left: &RoomListItem, right: &RoomListItem| extract_rank(left, right);
+    let scores = |left: &RoomListItem, right: &RoomListItem| extract_score(left, right);
```
bnjbvr: preexisting: is the `scores` closure needed? Looks like it has the exact same signature as `extract_score`, so I wonder if you could use `extract_score` in the call to `cmp` below :eyes:

crates/matrix-sdk-ui/src/room_list_service/sorters/recency.rs
```diff
@@ -41,33 +41,84 @@ where
...
+    // `Room::new_latest_event_timestamp`, where the value is cached in this
+    // module's `RoomListItem` type.
+
+    // Warning 2.
+    //
+    // A `RoomListItem` must have a unique score when sorting. Its `Score`
+    // must always be the same while sorting the rooms. Thus, the following
+    // rules must apply:
+    //
+    // - Nomical case: Two rooms with a latest event can be compared together based
```
bnjbvr:
```suggestion
    // - Nominal case: Two rooms with a latest event can be compared together based
```

crates/matrix-sdk-ui/src/room_list_service/sorters/recency.rs
```diff
@@ -41,33 +41,84 @@ where
...
+    //   first room has a score but the other doesn't have one,
+    // - Case #2: If none of the room has a latest event, we fallback to the recency
+    //   stamp for both rooms.
+    //
+    // The most important aspect is: if room returns its latest event's
+    // timestamp or its recency stamp, _once_, it must return it every time
+    // it's compared to another room, if possible, whatever the room is,
+    // otherwise it must return `None`.
+    match (left.cached_latest_event_timestamp, right.cached_latest_event_timestamp) {
+        // Nomical case.
```
bnjbvr:
```suggestion
        // Nominal case.
```

crates/matrix-sdk-ui/src/room_list_service/sorters/recency.rs
```diff
@@ -41,33 +41,84 @@ where
...
+        (left @ Some(_), right @ Some(_)) |
+
+        // Case #1.
+        (left @ None, right @ Some(_)) |
+        (left @ Some(_), right @ None)
+        => {
+            (left.map(|ms| ms.get().into()), right.map(|ms| ms.get().into()))
```
bnjbvr: I usually love short names, but what does `ms` stand for, in this case? I'd expect either `ts` for timestamp, or some shorter way to spell `cached_latest_event_timestamp`.

crates/matrix-sdk-ui/src/room_list_service/sorters/recency.rs
Hywan: Good catch.

crates/matrix-sdk-ui/src/room_list_service/sorters/recency.rs
Hywan: Comment updated to make the “where the value is cached in the `RoomListItem` type” part clearer.

crates/matrix-sdk-ui/src/room_list_service/sorters/recency.rs
Hywan: Because it would make the comparison a bit harder in the `match` where it's used. If we add an enum for `Score`, we could get a`into_u64` method that could run some asserts though, like so:

```rust
struct Scores(Score, Score);

impl Scores {
    fn into_u64(self) -> (u64, u64) {
        debug_assert!(…);
    }
}
```

Thoughts?

crates/matrix-sdk-ui/src/room_list_service/sorters/recency.rs
Hywan: Let's go with `ts`!

crates/matrix-sdk-ui/src/room_list_service/sorters/recency.rs
Hywan: Done in https://github.com/matrix-org/matrix-rust-sdk/pull/5698/commits/97e5772b9d65f563967f82f9775ddff1057bd6a0.

crates/matrix-sdk-ui/src/room_list_service/sorters/recency.rs
```diff
@@ -31,43 +31,143 @@ where
...
+            )
+        }
+
+        // Case #2.
+        (None, None) => {
+            Scores(
+                left.cached_recency_stamp.map(Into::into).map(Score::RecencyStamp),
+                right.cached_recency_stamp.map(Into::into).map(Score::RecencyStamp)
+            )
+        }
```
bnjbvr: Not sure a match is the best in terms of readability here? sounds like

```rust
if left.cached_latest_event_timestamp.is_some() || right.cached_latest_event_timestamp.is_some() {
 // use timestamps
} else {
 // use recency stamps
}
```

is shorter and easier on my smol brain. What do you think?

crates/matrix-sdk-ui/src/room_list_service/sorters/recency.rs
```diff
@@ -31,43 +31,143 @@ where
...
+                unreachable!("`Score::TimeStamp` cannot be compared to a `Score::RecencyStamp`")
+            }
+        }
+    }
+}
+
+/// Represent a score from either a latest event's timestamp or a room's recency
+/// stamp.
+#[derive(Debug)]
+enum Score {
```
bnjbvr: Hmm, I didn't imagine this would be that much code to do the same thing, and this didn't quite achieve the goal of concision and type-checking safety I hoped for. Sorry to ask this, but maybe we could either revert to the previous version, or (equivalently, I think) have `extract_scores` return the pair of int64, that can be immediately compared, without having to resort to `Score` and `Scores`?

Hywan: 😢 but 👍.

crates/matrix-sdk-ui/src/room_list_service/sorters/recency.rs
Hywan: Done.

PR #5699, opened 2025-09-22
Add function to check if the user has another device to verify against

uhoreg:
Part of the fix for https://github.com/element-hq/element-x-android/issues/4864 and https://github.com/element-hq/element-x-ios/issues/4190

Allows applications to determine whether the user can verify against another device in order to cross-sign their new device.

Signed-off-by:

### DIFF
--- a/bindings/matrix-sdk-ffi/src/encryption.rs
+++ b/bindings/matrix-sdk-ffi/src/encryption.rs
@@ -287,8 +287,8 @@ impl Encryption {
         Ok(self.inner.recovery().is_last_device().await?)
     }
 
-    /// Does the user has other devices that the current device can verify
-    /// against.
+    /// Does the user have other devices that the current device can verify
+    /// against?
     ///
     /// The device must be signed by the user's cross-signing key, must have an
     /// identity, and must not be a dehydrated device.
--- a/crates/matrix-sdk/src/encryption/mod.rs
+++ b/crates/matrix-sdk/src/encryption/mod.rs
@@ -865,8 +865,8 @@ impl Encryption {
         Some(machine.cross_signing_status().await)
     }
 
-    /// Does the user has other devices that the current device can verify
-    /// against.
+    /// Does the user have other devices that the current device can verify
+    /// against?
     ///
     /// The device must be signed by the user's cross-signing key, must have an
     /// identity, and must not be a dehydrated device.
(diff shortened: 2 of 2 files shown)


bindings/matrix-sdk-ffi/src/encryption.rs
```diff
@@ -287,6 +287,15 @@ impl Encryption {
         Ok(self.inner.recovery().is_last_device().await?)
     }
 
+    /// Does the user has other devices that the current device can verify
```
andybalaam: has->have

bindings/matrix-sdk-ffi/src/encryption.rs
```diff
@@ -287,6 +287,15 @@ impl Encryption {
...
+    /// against.
```
andybalaam: Complete nit: this sentence should end with `?`

bindings/matrix-sdk-ffi/src/encryption.rs
andybalaam: (Same for the version in `crates/matrix-sdk/src/encryption/mod.rs`)

crates/matrix-sdk/src/encryption/mod.rs
```diff
@@ -2428,4 +2451,197 @@ mod tests {
...
+    /// Test that we can detect whether the user has devices that they can
+    /// verify against.
+    async fn test_devices_to_verify_against() {
+        let server = MockServer::start().await;
+        let client = logged_in_client(Some(server.uri())).await;
+        let user_id = client.user_id().unwrap();
+        let olm_machine = client.olm_machine().await;
+        let olm_machine = olm_machine.as_ref().unwrap();
+
+        // We have no devices, so we can't verify against anything
```
andybalaam: I would prefer 3 separate tests instead of one big one.

crates/matrix-sdk/src/encryption/mod.rs
```diff
@@ -2428,4 +2451,197 @@ mod tests {
...
+        }
+
+        // We have a cross-signed device, so we can verify against it
+        {
+            // user ID: "@example:localhost"
+            // device ID: "DEVICEID"
+            let _mock_guard = Mock::given(method("POST"))
+                .and(path_regex(r"^/_matrix/client/r0/keys/query".to_owned()))
+                .respond_with(ResponseTemplate::new(200).set_body_json(json!({
+                    "device_keys": {
```
andybalaam: I would find this test easier to read if the actual `json!` macro were broken out into a helper function that took a few parameters. Could we re-use that helper function with the 3rd test below?

bindings/matrix-sdk-ffi/src/encryption.rs
uhoreg: I'm pretty sure I thought about doing that, but then I saw other function descriptions of the form "Does ... " that ended with a period.  But now I can't see the descriptions that I thought I saw.  :shrug:

uhoreg:
@andybalaam I've made the changes.  Can you hit the merge button?  I don't have permissions in this repo.

PR #5705, opened 2025-09-24
Add some doc to `add_event_handler` for invites and stripped state

MatMaul:
Signed-off-by: Mathieu Velten <mathieu.velten@beta.gouv.fr>

### DIFF
--- a/crates/matrix-sdk/src/client/mod.rs
+++ b/crates/matrix-sdk/src/client/mod.rs
@@ -818,7 +818,11 @@ impl Client {
     ///         events::{
     ///             macros::EventContent,
     ///             push_rules::PushRulesEvent,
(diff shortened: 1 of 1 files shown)


crates/matrix-sdk/src/client/mod.rs
```diff
@@ -871,6 +875,18 @@ impl Client {
     ///     // Use the context
     /// });
     ///
+    /// // This will handle membership events in joined rooms. Invites are special, cf below.
```
poljar: You'd need a dot for the `cf.`, but let's just change it to something simpler.

```suggestion
    /// // This will handle membership events in joined rooms. Invites are special, see below.
```

PR #5707, opened 2025-09-24
refactor(base): Rename `sync_lock` to `state_store_lock`

Hywan:
This patch renames the `sync_lock` to `state_store_lock` because it is what it is. It's not about the sync, it's about the state store.


Hywan:
You mean “I state_store_lock this is great”?

PR #5708, opened 2025-09-24
refactor(timeline): in thread permalinks, avoid back-paginating if the root event is part of the /context response

bnjbvr:
For thread permalinks, we start with a /context query that will load the focused event, and maybe a few other in-thread events. In fact, it can also include the thread root event, which was excluded before. Instead, we would get a previous-token for back-paginations, which would be used in /relations. When the request to /relations returns an empty previous token, that means we've reached the start of the thread, and in this case we would manually load the root with /event.

We can do better, if the root event is part of the initial /context response: skip the back-paginations altogether, and make sure to include the root event in `init_focus()`.

### DIFF
--- a/crates/matrix-sdk-ui/src/timeline/controller/mod.rs
+++ b/crates/matrix-sdk-ui/src/timeline/controller/mod.rs
@@ -25,7 +25,7 @@ use matrix_sdk::Result;
 use matrix_sdk::{
     deserialized_responses::TimelineEvent,
     event_cache::{RoomEventCache, RoomPaginationStatus},
-    paginators::{PaginationResult, Paginator},
+    paginators::{PaginationResult, PaginationToken, Paginator},
     send_queue::{
         LocalEcho, LocalEchoContent, RoomSendQueueUpdate, SendHandle, SendReactionHandle,
     },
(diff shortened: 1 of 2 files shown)


crates/matrix-sdk-ui/src/timeline/controller/mod.rs
```diff
@@ -492,13 +494,28 @@ impl<P: RoomDataProvider> TimelineController<P> {
 
                 let _ = paginator.set(match thread_root_event_id {
                     Some(root_id) => {
-                        let tokens = event_paginator.tokens();
+                        let mut tokens = event_paginator.tokens();
+
+                        // Look if the thread root event is part of the /context response. This
+                        // allows us to spare some backwards pagination with
+                        // /relations.
+                        let includes_root_event = start_from_result.events.iter().any(|event| {
```
Hywan, line 502: I suppose the thread root event —if present— is the first one, right? Do we need to iterate over all the events then? Shouldn't we just look at the first one?

bnjbvr: Unfortunately no: there could be unrelated events before the target (the "focused" event) — also after it, but we don't really care in this context.

Say your thread root is $r, the target event is $t, other events are $ev1/2/3.

All these `start_from_result.events` lists are possible and valid:

- `$t`
- `$r`, `$ev1`, `$t`, `$ev2`
- `$ev1`, `$r`, `$ev2`, `$t`, `$ev3`

PR #5711, opened 2025-09-24
feat(oauth): add flow for logging in using a QR code generated on the logged out device

Johennes:


poljar:
Could we split this PR up a bit?

- Commit 1 is an easy code move.
- Commit 2 is a breaking change lacking a changelog entry
- Commit 3 is a breaking change lacking a changelog entry and a motivation in the commit message.
- Commit 4 is the meat of the PR.

Ideally we would merge 1-3 as separate PRs and then discuss 4 in a more detailed manner.

Johennes:
Ok, sure. I'll put this one back into draft until the others have landed.

kathleenmotley49-maker (approved):

Johennes:
This is now reduced to the formerly last commit with the previous ones having landed on `main`.

crates/matrix-sdk/src/authentication/oauth/mod.rs
```diff
@@ -459,6 +459,31 @@ impl OAuth {
...
+    ///   the server. If this is not provided, an error will occur unless
+    ///   [`OAuth::register_client()`] or [`OAuth::restore_registered_client()`]
+    ///   was called previously.
+    #[cfg(feature = "e2e-encryption")]
+    pub fn login_with_generated_qr_code<'a>(
+        &'a self,
+        registration_data: Option<&'a ClientRegistrationData>,
+    ) -> LoginWithGeneratedQrCode<'a> {
+        LoginWithGeneratedQrCode::new(&self.client, registration_data)
+    }
```
poljar: I think a separate method here is a bit annoying, we'll have 4 methods in total just for the QR code feature if we do it this way.

We could instead extend `login_with_qr_code()` to return a builder which then either accepts a `QrCodeData` and gives you a named future which logs you in by scanning a QR code. Otherwise it gives you a future which logs you in by generating a QR code.

crates/matrix-sdk/src/authentication/oauth/mod.rs
```diff
@@ -459,6 +459,31 @@ impl OAuth {
...
+    /// as verified and transfer all end-to-end encryption related secrets, like
+    /// the private cross-signing keys and the backup key from the existing
+    /// device to the new device.
+    ///
+    /// # Arguments
+    ///
+    /// * `registration_data` - The data to restore or register the client with
+    ///   the server. If this is not provided, an error will occur unless
+    ///   [`OAuth::register_client()`] or [`OAuth::restore_registered_client()`]
+    ///   was called previously.
```
poljar: We should add an example to this docs, it's never easy to figure out what you need to do to use such complex futures.

crates/matrix-sdk/src/authentication/oauth/qrcode/login.rs
```diff
@@ -264,12 +264,68 @@ pub enum LoginProgress<Q> {
...
+/// Metadata to be used with [`LoginProgress::EstablishingSecureChannel`] when
+/// this device is the one generating the QR code.
+///
+/// We have established the secure channel, but we need to let the
+/// other side know about the [`QrCodeData`] so they can send us the
+/// [`CheckCode`] and verify that the secure channel is indeed secure.
+#[derive(Clone, Debug)]
+pub enum GeneratedQrProgress {
+    /// The QR code data that must be sent to the existing device.
+    QrReady(QrCodeData),
```
poljar, line 286: I wonder if it would be nicer if the login method gave you the future and QR code as a tuple.

I guess no, because getting the rendezvous channel set up might take some time and UI clients will want to show a spinner during this.
 
Having it here makes showing a spinner a natural step in the UI.

crates/matrix-sdk/src/authentication/oauth/qrcode/login.rs
```diff
@@ -339,6 +395,111 @@ impl<'a> LoginWithQrCode<'a> {
...
+    }
+
+    async fn establish_secure_channel(
+        &self,
+    ) -> Result<EstablishedSecureChannel, SecureChannelError> {
+        let http_client = self.client.inner.http_client.clone();
+
+        let secure_channel = SecureChannel::login(http_client, &self.client.homeserver()).await?;
+
+        assert_eq!(QrCodeModeData::Login, secure_channel.qr_code_data().mode_data);
```
poljar: We usually don't put asserts in code but if you'd like to leave it give it a more descriptive error message as the third argument to the `assert_eq!()`.

crates/matrix-sdk/src/authentication/oauth/qrcode/login.rs
```diff
@@ -339,6 +395,111 @@ impl<'a> LoginWithQrCode<'a> {
...
+        self.state.set(LoginProgress::EstablishingSecureChannel(GeneratedQrProgress::QrScanned(
+            CheckCodeSender::new(tx),
+        )));
+
+        let check_code = rx.await.map_err(|_| SecureChannelError::CannotReceiveCheckCode)?;
+
+        trace!("Received check code.");
+
+        channel.confirm(check_code)
+    }
```
poljar, line 543: This function needs some more comments to describe what each step is achieving similar to how the `LoginWithQrCode` future does. Bonus points if you note the step numbers from the MSC in each block, something I failed to do as well.

Otherwise this it's pretty hard to double check which code blocks relate to which step in the process.

crates/matrix-sdk/src/authentication/oauth/qrcode/secure_channel.rs
```diff
@@ -32,21 +30,32 @@ use crate::{config::RequestConfig, http_client::HttpClient};
...
+        let channel = RendezvousChannel::create_outbound(http_client, homeserver_url).await?;
+        let rendezvous_url = channel.rendezvous_url().to_owned();
+        let mode_data = QrCodeModeData::Login;
+
+        let ecies = Ecies::new();
+        let public_key = ecies.public_key();
+
+        let qr_code_data = QrCodeData { public_key, rendezvous_url, mode_data };
+
+        Ok(Self { channel, qr_code_data, ecies })
```
poljar, line 54: The `new()` method could use this method and just mutate or create a new `QrCodeModeData`.

This would avoid much of the code duplication here.

crates/matrix-sdk/CHANGELOG.md
```diff
@@ -79,6 +79,9 @@ All notable changes to this project will be documented in this file.
   begins ([#5678](https://github.com/matrix-org/matrix-rust-sdk/pull/5678).
 - Make `PaginationTokens` `pub`, as well as its `previous` and `next` tokens so they can be assigned from other files
   ([#5678](https://github.com/matrix-org/matrix-rust-sdk/pull/5678).
+- Add `OAuth::login_with_generated_qr_code` for generating a QR code on a new device
+  and logging it in with the help of an existing device scanning the code.
+  ([#5711](https://github.com/matrix-org/matrix-rust-sdk/pull/5711))
```
poljar: That's not the only change, some error types got new variants as well. Extending the error types is a breaking change.

crates/matrix-sdk/src/authentication/oauth/qrcode/login.rs
```diff
@@ -264,12 +264,68 @@ pub enum LoginProgress<Q> {
...
+}
+
+impl CheckCodeSender {
+    fn new(tx: tokio::sync::oneshot::Sender<u8>) -> Self {
+        Self { inner: Arc::new(Mutex::new(Some(tx))) }
+    }
+
+    /// Send the [`CheckCode`].
+    ///
+    /// Calling this method more than once will result in an error.
```
poljar, line 306: We should probably mention that you should send the digits representation of the check code, the `CheckCode` struct has two representations.

crates/matrix-sdk/src/authentication/oauth/qrcode/login.rs
```diff
@@ -264,12 +264,68 @@ pub enum LoginProgress<Q> {
...
+/// Possible errors when calling [`CheckCodeSender::send`].
+#[derive(Debug, thiserror::Error)]
+pub enum CheckCodeSenderError {
+    /// The check code has already been sent.
+    #[error("check code already sent.")]
+    AlreadySent,
+    /// The check code cannot be sent.
+    ///
+    /// This probably means that the new device log-in process
+    /// crashed somehow.
```
poljar: No need to crash, you can drop the `Future` to close the channel.

crates/matrix-sdk/src/authentication/oauth/qrcode/login.rs
```diff
@@ -339,6 +395,111 @@ impl<'a> LoginWithQrCode<'a> {
...
+        Box::pin(async move {
+            let mut channel = self.establish_secure_channel().await?;
+
+            trace!("Established the secure channel.");
+
+            // Receive m.login.protocols message.
+            let message = channel.receive_json().await?;
+
+            match message {
+                QrAuthMessage::LoginProtocols { protocols, .. } => {
```
poljar: A step is missing here which becomes obvious once you look at the rest of the fields the `LoginProtocols` message contains.

image

You'll need to modify the homeserver we're going to use and re-resolve well-known.

A test that swapping out the homeserver works would be nice as well.

I think that this is important if the client picks up a default homeserver for the rendezvous but actually wants to log in into the homeserver the existing device is using.

Please add comments to this method as well, describing what each code block does and at which step in the sequence diagram we're at.

crates/matrix-sdk/src/authentication/oauth/qrcode/secure_channel.rs
Johennes: Good point. Fixed in 2de41589a8bf364cc08a6a752aca793acf29051d.

crates/matrix-sdk/src/authentication/oauth/qrcode/login.rs
Johennes: Actually, I think this isn't really needed anymore. Have removed it in 1183efad52fda9e67040c76f08b21741d494af5a.

crates/matrix-sdk/src/authentication/oauth/qrcode/login.rs
Johennes: Ok, true. Have removed the misleading part in 9c4e005683a05cbe6077bbc362de81fd1c911a35.

crates/matrix-sdk/src/authentication/oauth/qrcode/login.rs
Johennes: Have extended the comment in e5e59e49eea6a64e0ee1fa7a0d15e326b02a4adc.

crates/matrix-sdk/src/authentication/oauth/mod.rs
Johennes: Have added one together with the builder switch in 5e385199f8c971af8dfb81e911bcaa50c7c568f9.

crates/matrix-sdk/src/authentication/oauth/mod.rs
Johennes: Good idea. Have switched to a builder in 5e385199f8c971af8dfb81e911bcaa50c7c568f9.

poljar: Much nicer, thanks.

crates/matrix-sdk/src/authentication/oauth/qrcode/login.rs
Johennes: Have tried to annotate the different steps in both flows with 8d2735f02342255a030383e8d787a06160c2d35f. It doesn't match super clearly in all cases but I hope it's acceptable.

crates/matrix-sdk/src/authentication/oauth/qrcode/login.rs
Johennes: Good catch. Have added the code and a test for handling homeserver swaps in 2a61188b7d9918a3d0ab9224ac9aac6b2c4f96e5 and also sprinkled on some comments.

crates/matrix-sdk/CHANGELOG.md
Johennes: Sorry, I missed that. Have amended the changelog entry with 329a4c659b22327dc3d268d3346b61c8b43c097b.

crates/matrix-sdk/src/authentication/oauth/mod.rs
Johennes: Sorry, I forgot to update the doc comments on the two futures accordingly. This is now done with 6cb59d081a8af261bd3fca3bfd1271e4af5cd5ce.

Johennes:
Sorry, took me a bit longer to address everything and unbreak the CI. Should be ready for re-review now.

crates/matrix-sdk/src/authentication/oauth/qrcode/login.rs
```diff
@@ -339,6 +408,139 @@ impl<'a> LoginWithQrCode<'a> {
...
+                    });
+                }
+            };
+
+            // Change the login homeserver if it is different from the server hosting the
+            // secure channel.
+            if self.client.homeserver() != homeserver {
+                self.client.set_homeserver(homeserver);
+                self.client.reset_server_info().await.map_err(QRCodeLoginError::ServerReset)?;
+            }
```
poljar, line 486: I'm afraid that this isn't enough. The MSC explicitly tells us that we still need to resolve the well-known to get the URL.

In the case where we scan the QR code, this happens at the `Client` building stage, but here we need to do that after the server swap.

Image

crates/matrix-sdk/src/authentication/oauth/mod.rs
```diff
@@ -1403,6 +1324,181 @@ impl OAuth {
...
+    /// This method allows you to log in with a scanned QR code.
+    ///
+    /// The existing device needs to display the QR code which this device can
+    /// scan and call this method to log in.
+    ///
+    /// A successful login using this method will automatically mark the device
+    /// as verified and transfer all end-to-end encryption related secrets, like
+    /// the private cross-signing keys and the backup key from the existing
+    /// device to the new device.
+    ///
```
poljar, line 1349: I know that this is fairly easy to discover, but can we still link from one method to the other and vice-versa?

crates/matrix-sdk/src/authentication/oauth/qrcode/login.rs
```diff
@@ -296,19 +354,30 @@ impl<'a> IntoFuture for LoginWithQrCode<'a> {
...
     fn into_future(self) -> Self::IntoFuture {
         Box::pin(async move {
+            // Before we get here, the other device has created a new rendezvous session
+            // and presented a QR code which this device has scanned.
+            // -- MSC4108 Secure channel setup steps 1-3
+
             // First things first, establish the secure channel. Since we're the one that
             // scanned the QR code, we're certain that the secure channel is
             // secure, under the assumption that we didn't scan the wrong QR code.
+            // -- MSC4108 Secure channel setup steps 3-5
```
poljar, line 364: Thanks for adding those, even though I failed to add them myself.

crates/matrix-sdk/src/authentication/oauth/mod.rs
Johennes: Yeah, sounds helpful. Have extended the comments with 68a95112abfa1a0abdf97ea0517a9e195f669466.

crates/matrix-sdk/src/authentication/oauth/qrcode/login.rs
Johennes: Sorry, somehow I was under the misassumption that this would happen automatically after resetting the cache but looks like I was wrong. Fixed with 17c7dc09281e3db57048df819c80de83882fc478.

crates/matrix-sdk/src/client/mod.rs
```diff
@@ -529,10 +529,26 @@ impl Client {
...
+    pub(crate) async fn switch_homeserver_and_re_resolve_well_known(
+        &self,
+        homeserver_url: Url,
+    ) -> Result<()> {
+        self.set_homeserver(homeserver_url);
+        self.reset_server_info().await?;
+        if let ServerInfo { well_known: Some(well_known), .. } =
+            self.load_or_fetch_server_info().await?
+            && let Ok(homeserver) = Url::parse(&well_known.homeserver.base_url)
+        {
```
poljar: Don't we need to throw an error here, otherwise this turns into a no-op if we can't parse the URL.

Am I missing something?

Johennes: Hm, I think you're right. I guess that should throw in `maybe_update_login_well_known`, too, then?

Johennes: Stopped swallowing the error in this method with 1383ff8fdcbec3f1293827f52b13affc25d1050e.

poljar: `maybe_update_login_well_known` sounds like a best effort method, so it might be fine there.

Though looking at the place where it's used it might make more sense to error out there as well.

poljar (approved):
Alright, this looks good now. Thanks for patiently working on this.

I know it's a bit slow and annoying but looking at the resulting code/docs/comments I think it was very much worth it.

Let me know if you'd like to clean up the history or if I should just squash merge.

Johennes:
> I know it's a bit slow and annoying but looking at the resulting code/docs/comments I think it was very much worth it.

Oh, not annoying at all and a lot better than the initial PR. Thanks for the review. 👍 

> Let me know if you'd like to clean up the history or if I should just squash merge.

If squash-merging is an option, I'll take it. Otherwise, I would squash it into a single commit locally .. which sort of leads to the same result.

PR #5712, opened 2025-09-25
feat(spaces): automatically subscribe to `SpaceRoomList` "parent" space room info updates 

stefanceriu:
… when known to the client and forward updates through the existing `subscribe_to_space_updates` mechanisms.

This allows clients to listen to updates without having to resort to a separate room info subscription on their side.

### DIFF
--- a/crates/matrix-sdk-ui/src/spaces/room_list.rs
+++ b/crates/matrix-sdk-ui/src/spaces/room_list.rs
@@ -103,7 +103,7 @@ pub struct SpaceRoomList {
 
     space_id: OwnedRoomId,
 
-    space: SharedObservable<Option<SpaceRoom>>,
+    space: Arc<SharedObservable<Option<SpaceRoom>>>,
 
     via_parameters: Mutex<Option<HashMap<OwnedRoomId, Vec<OwnedServerName>>>>,
 
@@ -112,6 +112,8 @@ pub struct SpaceRoomList {
     pagination_state: SharedObservable<SpaceRoomListPaginationState>,
 
     rooms: Arc<Mutex<ObservableVector<SpaceRoom>>>,
+
+    _space_update_handle: Option<AbortOnDrop<()>>,
 
     _room_update_handle: AbortOnDrop<()>,
 }
@@ -123,7 +125,7 @@ impl SpaceRoomList {
 
         let all_room_updates_receiver = client.subscribe_to_all_room_updates();
 
-        let handle = spawn({
+        let room_update_handle = spawn({
             let client = client.clone();
             let rooms = rooms.clone();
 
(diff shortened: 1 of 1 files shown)


poljar:
A rebase might make your CI happy.

crates/matrix-sdk-ui/src/spaces/room_list.rs
```diff
@@ -164,28 +166,53 @@ impl SpaceRoomList {
...
 
-            Some(SpaceRoom::new_from_known(&parent, children_count))
+            let mut subscriber = parent.subscribe_info();
+            let space_update_handle = spawn({
+                let client = client.clone();
+                let space_id = space_id.clone();
+                let space_observable = space_observable.clone();
+                async move {
+                    loop {
+                        if subscriber.next().await.is_some()
```
poljar: Why the infinite loop and the `is_some()` call vs the usual:

```rust
while let Some(update) = subscriber.next().await {
    ...
}
```

poljar (changes requested):
Makes sense, though the execution seems a bit unusual compared to our other patterns.

crates/matrix-sdk-ui/src/spaces/room_list.rs
stefanceriu: Hmm I think I got that from somewhere on the FFI layer, the client I believe. To be honest I was a bit confused by the API but I now realise it's returning None only when/after being dropped.

stefanceriu:
Yey, thank you! 🙌

crates/matrix-sdk-ui/src/spaces/room_list.rs
```diff
@@ -103,7 +103,7 @@ pub struct SpaceRoomList {
 
     space_id: OwnedRoomId,
 
-    space: SharedObservable<Option<SpaceRoom>>,
+    space: Arc<SharedObservable<Option<SpaceRoom>>>,
```
Hywan, line 106: Note: This is a `SharedObservable`. You can already share it. The `Arc` isn't necessary here.

stefanceriu: Does that apply to clones cause I tried that and it didn't actually work? 🤔

Hywan: https://github.com/matrix-org/matrix-rust-sdk/pull/5720

Hywan: https://docs.rs/eyeball/latest/eyeball/struct.SharedObservable.html

> Unlike [Observable](https://docs.rs/eyeball/latest/eyeball/struct.Observable.html), `SharedObservable` can be `Clone`d but does’t dereference to `T`.

PR #5713, opened 2025-09-25
ci: Install clippy for the test-crypto CI run

poljar:


poljar:
> Sure, but wait, why would this only be needed for this peculiar CI step? Why do the others keep on working? 🤔

Well most of our steps don't require clippy, we have a dedicated clippy step and this `test-crypto` step uses clippy as well.

I didn't go through each step to completely confirm this but I took a brief glance.

PR #5714, opened 2025-09-25
fix(ffi): `ffi::Room::load_or_fetch_event` fails with missing room_id

jmartinesp:
The raw event was being deserialized in a wrong way and it could have a missing room id in some cases, returning an error even when the event was found.

Signed-off-by:


poljar (approved):
Makes sense, events coming from sync won't have the room id so we'll need to add it if we want to deserialize it into a `TimelineEvent`.

PR #5717, opened 2025-09-25
FFI: Expose a room list filter for spaces.

pixlwave:
Currently only a `NonSpace` filter is available. This new one doesn't help re only showing top-level spaces, but it does allow Element X to show space invites in the room list.


pixlwave:
Thanks Andy, any chance you could merge this for me please? 😇

PR #5723, opened 2025-09-29
fix(timeline): align initial loading with sync of read receipts in compatibility mode

bnjbvr:
In code elsewhere, when getting new receipts via sync, and the timeline was to configured to display the `main` thread (that is, a live timeline without threaded events), then we would either use the "unthreaded" or "main" read receipts to compute the read receipts of an event, for maximal compatibility with people not using thread-aware clients.

There was a discrepancy, though, in the initial loading of receipts, which, in the same situation, would only use the "main" read receipt. As a result, one can see read receipts jumping "backwards" in time, after closing and re-opening a timeline view, if an unthreaded read receipt came in real-time through sync, but there's no associated main read receipt.

I think this fixes https://github.com/element-hq/element-x-android/issues/5411.

### DIFF
--- a/crates/matrix-sdk-ui/tests/integration/timeline/thread.rs
+++ b/crates/matrix-sdk-ui/tests/integration/timeline/thread.rs
@@ -1320,6 +1320,84 @@ async fn test_initial_read_receipts_are_correctly_populated() {
 }
 
 #[async_test]
+async fn test_initial_read_receipts_compatibility_mode() {
+    // If there are initial read receipts in the store, that are using the
+    // "unthreaded" kind of receipt, then they're used as "main" receipts as
+    // well.
+
+    let server = MatrixMockServer::new().await;
+    let client = client_with_threading_support(&server).await;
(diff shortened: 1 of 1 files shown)


bnjbvr:
Undrafting, as I've added a regression test now. PTAL!

bnjbvr:
> Those receipts are a bit headache inducing if you ask me.

Agreed! I'm starting to think about a new mental model about it (in the context of #4113), where the event cache would have the responsibility to track read receipts on events:

- tracking means that the read receipt would move, from one event to another, after we receive new read receipt events via sync
- the event cache would propagate dedicated pings for this kind of tracked receipt updates, so the timeline can use that to move read receipts on its items
- it would offer new APIs to load the tracked receipt on a given event (loading all the receipts of an event is already possible via the state store), so the timeline can still load the initial tracked receipts on each event

I'm confident that, by putting all of this under the event cache, we can simplify the logic that's currently living in the timeline.

PR #5727, opened 2025-09-30
refactor(event cache): random refactorings in preparation for caching threads

bnjbvr:
I was looking at some thread summary bugs, and while looking at other code, noticed a few places where slightly refactoring the code would pave the way for caching in-thread events.

### DIFF
--- a/crates/matrix-sdk/src/event_cache/room/mod.rs
+++ b/crates/matrix-sdk/src/event_cache/room/mod.rs
@@ -912,21 +912,15 @@ mod private {
                 return Ok(LoadMoreEventsBackwardsOutcome::Gap { prev_token: Some(prev_token) });
             }
 
-            // Because `first_chunk` is `not `Send`, get this information before the
-            // `.await` point, so that this `Future` can implement `Send`.
-            let first_chunk_identifier = self
-                .room_linked_chunk
-                .chunks()
-                .next()
-                .expect("a linked chunk is never empty")
-                .identifier();
-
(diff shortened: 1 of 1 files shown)


crates/matrix-sdk/src/event_cache/room/mod.rs
```diff
@@ -912,21 +912,15 @@ mod private {
                 return Ok(LoadMoreEventsBackwardsOutcome::Gap { prev_token: Some(prev_token) });
             }
 
-            // Because `first_chunk` is `not `Send`, get this information before the
-            // `.await` point, so that this `Future` can implement `Send`.
```
Hywan, line 946: I hope this isn't necessary anymore.

Hywan (approved):
Thanks! It looks good to me.

crates/matrix-sdk/src/event_cache/room/mod.rs
bnjbvr: Yep, this was referring to a previous implementation, since there wasn't even a `first_chunk` variable here. I managed to reproduce the issue, by having `let first_chunk: &Chunk = ...;` (and then using `first_chunk.identifier()`) below. The `first_chunk` would be borrowed across the `.await` point for getting the store lock, which would indeed cause the borrowck issue.

But by moving the definition right to *after* the lock was taken, the problem disappears: the borrowed chunk is only borrowed for a short time.

(That's why I've removed the comment too.)

PR #5728, opened 2025-09-30
feat(timeline): allow an edit to be the latest event in a thread summary

### LINKED ISSUE #5715: thread summary: last message does not update on edit
Steps to reproduce:
- Alice and Bob have a room with a thread. They both uses EX.
- Alice sends a message "t0" in the thread
- Alice sends a message "t2" in the thread
- Alice edits "t2" into "t1"

The problem:
For Alice and Bob, the main timeline of the room continues to show "t2" as the last message whereas "t1" is expected.

Notes
* Clearing the cache does nothing
* There is actually no need to have 2 users in the room. Alice alone meets the bug with those steps.

bnjbvr:
## event cache changes

When a postprocessed event is an edit, and the edited event is in a thread, mark the related thread as requiring a thread update, which will recompute the thread summary (and propagate an event cache update, if needs be).

## timeline changes

Allow edits to be represented as `EmbeddedEvent`s, so they can appear in a reply context, or a thread's latest event context.

Then a few refactorings to make it also possible to allow poll edits to be `EmbeddedEvent`s too.

---

Fixes #5715.

### DIFF
--- a/crates/matrix-sdk-ui/src/timeline/event_item/content/reply.rs
+++ b/crates/matrix-sdk-ui/src/timeline/event_item/content/reply.rs
@@ -21,7 +21,8 @@ use tracing::{debug, instrument, warn};
 
 use super::TimelineItemContent;
 use crate::timeline::{
-    Error as TimelineError, MsgLikeContent, PollState, TimelineEventItemId, TimelineItem,
+    Error as TimelineError, MsgLikeContent, MsgLikeKind, PollState, TimelineEventItemId,
+    TimelineItem,
     controller::TimelineMetadata,
     event_handler::{HandleAggregationKind, TimelineAction},
     event_item::{EventTimelineItem, Profile, TimelineDetails},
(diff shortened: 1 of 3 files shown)


crates/matrix-sdk-common/src/serde_helpers.rs
```diff
@@ -72,9 +75,23 @@ pub fn extract_thread_root_from_content(
 /// Returns `None` if we couldn't find a thread root, or if there was an issue
 /// during deserialization.
 pub fn extract_thread_root(event: &Raw<AnySyncTimelineEvent>) -> Option<OwnedEventId> {
+    extract_thread_root_from_content(event.get_field("content").ok().flatten()?)
+}
+
+/// Try to extract the target of an edit event, from a ra timeline event, if
```
poljar:
```suggestion
/// Try to extract the target of an edit event, from a raw timeline event, if
```

poljar (approved):
This looks good, left a small nit.

PR #5729, opened 2025-09-30
refactor(stores): get rid of the temporary `compute_filter_strings` now that Ruma has been updated

bnjbvr:
This was a local fix for a bug in Ruma, that has been fixed upstream since then, so we can get rid of the workaround now.

### DIFF
--- a/crates/matrix-sdk-sqlite/src/event_cache_store.rs
+++ b/crates/matrix-sdk-sqlite/src/event_cache_store.rs
@@ -21,7 +21,7 @@ use deadpool_sqlite::{Object as SqliteAsyncConn, Pool as SqlitePool, Runtime};
 use matrix_sdk_base::{
     deserialized_responses::TimelineEvent,
     event_cache::{
-        store::{compute_filters_string, extract_event_relation, EventCacheStore},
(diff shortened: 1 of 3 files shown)


crates/matrix-sdk-sqlite/src/event_cache_store.rs
```diff
@@ -1399,7 +1399,13 @@ fn find_event_relations_transaction(
             WHERE relates_to = ? AND room_id = ? AND rel_type IN ({question_marks})"
         );
 
-        let filters: Vec<_> = filters.iter().map(|f| f.to_sql().unwrap()).collect();
+        // First the filters need to be stringified; because `.to_sql()` will borrow
+        // from them, they also need to be stringified onto the stack, so as to
+        // get a stable address (to avoid returning a temporary reference in the
+        // map closure below).
+        let filter_strings: Vec<_> = filters.iter().map(|f| f.to_string()).collect();
+        let filters_params: Vec<_> = filter_strings.iter().map(|f| f.to_sql().unwrap()).collect();
```
poljar: Existing, but would be nice to expect, if for nothing else so that we can more easily disallow naked unwraps throughout the codebase.

PR #5731, opened 2025-10-01
feat(spaces): leave space mechanism

### LINKED ISSUE #5718: New APIs to leave Space
In the context of https://github.com/element-hq/element-meta/issues/2914, we will need 2 new APIs exposed to the FFI layer:

- one API to get all the joined rooms/spaces of a specific space, and recursively (flattened list of `SpaceRoom`). We will also need to know if the user is the last admin of each `SpaceRoom`. The FFI api could look like:
```
SpaceRoomListInterface.getAllJoinedSpaceRooms(): List<Pair<SpaceRoom, Boolean>>
```
 where the second item of the result is true when the user is the last admin of the SpaceRoom. 

- one API to leave the space (by roomId) and an optional list of other `SpaceRoom` to leave. The FFI api could look like: 
```
SpaceRoomListInterface.leave(subSpaceRoomIdsToLeave: List<String>)
```

stefanceriu:
This PR introduce a mechanism for leaving spaces for which we want the following:
* starting from a given room id representing a space graph subtree root, retrieve all joined rooms
* sort the rooms in reverse BFS order so that the first to be left will be the leafs and the last the initially given space parent
* annotate the results with membership details to inform the user when they are the last admin in a room to be left
* allow clients to only keep a subset of the provided rooms
* use that subset to leave

Fixes #5718

Note: I initially wanted to use a list of `LeaveSpaceRoom`s as the parameter for `LeaveSpaceHandle:: leave` to make it extra safe but that complicated things significantly on the FFI layer and made it not worth it.

### DIFF
--- a/bindings/matrix-sdk-ffi/src/spaces.rs
+++ b/bindings/matrix-sdk-ffi/src/spaces.rs
@@ -341,7 +341,7 @@ impl LeaveSpaceHandle {
     /// include leave specific information.
     pub fn rooms(&self) -> Vec<LeaveSpaceRoom> {
         let rooms = self.inner.rooms();
-        rooms.into_iter().map(|room| room.into()).collect()
+        rooms.iter().map(|room| room.clone().into()).collect()
     }
 
     /// Bulk leave the given rooms. Stops when encountering an error.
--- a/crates/matrix-sdk-ui/src/spaces/leave.rs
+++ b/crates/matrix-sdk-ui/src/spaces/leave.rs
@@ -74,13 +74,13 @@ impl LeaveSpaceHandle {
 
     /// A list of rooms to be left which next to normal [`SpaceRoom`] data also
     /// include leave specific information.
-    pub fn rooms(&self) -> Vec<LeaveSpaceRoom> {
-        self.rooms.clone()
+    pub fn rooms(&self) -> &Vec<LeaveSpaceRoom> {
+        &self.rooms
     }
 
     /// Bulk leave the given rooms. Stops when encountering an error.
     pub async fn leave(&self, filter: impl FnMut(&LeaveSpaceRoom) -> bool) -> Result<(), Error> {
-        for room in self.rooms().into_iter().filter(filter) {
+        for room in self.rooms.clone().into_iter().filter(filter) {
             if let Some(room) = self.client.get_room(&room.space_room.room_id) {
                 room.leave().await.map_err(Error::LeaveSpace)?;
             } else {


crates/matrix-sdk-ui/src/spaces/graph.rs
```diff
@@ -222,4 +252,76 @@ mod tests {
...
+        // │   └── Train
+        // ├── Personal
+        // │   ├── Car
+        // │   ├── Bicycle
+        // │   │   ├── Road
+        // │   │   ├── Gravel
+        // │   │   └── Mountain
+        // │   └── Helicopter
+        // └── Cargo
+        //     └── Plane
```
poljar, line 304: Boo, so generic. Should have put your favorite car/vehicle models up there. Like for example [Rexy](https://upload.wikimedia.org/wikipedia/commons/thumb/4/4f/AO_Porsche_WGI23_10.jpg/2560px-AO_Porsche_WGI23_10.jpg).

crates/matrix-sdk-ui/src/spaces/mod.rs
```diff
@@ -95,6 +95,8 @@ pub mod room_list;
 pub struct SpaceService {
     client: Client,
 
+    space_graph: Arc<Mutex<SpaceGraph>>,
+
```
poljar: Alright, those two pieces of data, the `space_graph` and `joined_spaces` are intimately related. I think it makes sense to put both of them behind a single `Mutex`.

crates/matrix-sdk-ui/src/spaces/leave.rs
```diff
@@ -0,0 +1,195 @@
...
+
+/// Space leaving specific room that groups normal [`SpaceRoom`] details with
+/// information about the leaving user's role.
+pub struct LeaveSpaceRoom {
+    /// The underlying [`SpaceRoom`]
+    pub space_room: SpaceRoom,
+    /// Whether the user is the last admin in the room. This helps clients
+    /// better inform the user about the consequences of leaving the room.
+    pub is_last_admin: bool,
+}
```
poljar: Can we put this at the top? You're referencing this type in the methods above, but to know what the type is I had to scroll down here.

crates/matrix-sdk-ui/src/spaces/leave.rs
```diff
@@ -0,0 +1,195 @@
...
+        for room_id in rooms_ids {
+            if let Some(room) = self.client.get_room(room_id) {
+                room.leave().await.map_err(Error::LeaveSpace)?;
+            } else {
+                return Err(Error::RoomNotFound(room_id.to_owned()));
+            }
+        }
+
+        Ok(())
+    }
```
poljar, line 92: Do we want to ensure somehow that the given `room_ids` are part of the `LeaveSpaceHandle.room_ids`?

Perhaps, instead of passing room IDs in this function we could pass a closure which will filter out `self.room_ids`:

```rust
/// Bulk leave the given rooms. Stops when encountering an error.
pub async fn leave(&self, filter: impl FnMut(&LeaveSpaceRoom) -> bool) -> Result<(), Error> {
    let rooms = self.rooms().await?;

    for room in rooms.into_iter().filter(filter) {
        if let Some(room) = self.client.get_room(&room.space_room.room_id) {
            room.leave().await.map_err(Error::LeaveSpace)?;
        } else {
            return Err(Error::RoomNotFound(room.space_room.room_id));
        }
    }

    Ok(())
}
```

This might remove the need for the `LeaveSpaceHandle` concept and you could just have the `leave_space()` method. Of course, unless removing the handle makes it painful to have this work over the FFI.

crates/matrix-sdk-ui/src/spaces/mod.rs
```diff
@@ -33,26 +33,49 @@ use eyeball_im::{ObservableVector, VectorSubscriberBatchedStream};
...
+pub mod leave;
 pub mod room;
 pub mod room_list;
 
+/// Possible [`SpaceService`] errors.
+#[derive(Debug, Error)]
+pub enum Error {
+    /// Unknown error.
+    #[error("Unknown error")]
+    Unknown,
```
poljar: Is this one used? Can't find it.

bindings/matrix-sdk-ffi/src/spaces.rs
```diff
@@ -85,6 +86,19 @@ impl SpaceService {
...
     }
+
+    /// Start a space leave process returning a [`LeaveSpaceHandle`] from which
+    /// rooms can be retrieved in reversed BFS order starting from the requested
+    /// `space_id` graph node. If the room is unknown then an error will be
+    /// returned.
+    ///
+    /// Once the rooms to be left are chosen the handle can be used to leave
+    /// them.
+    pub fn leave_space(&self, space_id: String) -> Result<LeaveSpaceHandle, ClientError> {
```
poljar: Yeah, I think you could just pass in a callback here. Would make for a simpler and more flexible API.

crates/matrix-sdk-ui/src/spaces/graph.rs
stefanceriu: I honestly wanted to but the decided best I stay serious. I did use `floo_powder` in one of the test though 😉 

Otherwise I'm more of a "baby" fan myself 😅 

Supernatural-Impala-feature copy 2

bindings/matrix-sdk-ffi/src/spaces.rs
stefanceriu: Can I actually do that over FFI?

crates/matrix-sdk-ui/src/spaces/graph.rs
poljar: > I honestly wanted to but the decided best I stay serious.

Right, we need to remain super cereal at all times.

> Otherwise I'm more of a "baby" fan myself 😅

Neat.

crates/matrix-sdk-ui/src/spaces/leave.rs
stefanceriu: My initial idea was to get back a list of `LeaveSpaceRoom`s so it's clear that they're originating from the handle. I'm not sure the filter will fly over ffi but I can certainly check that the passed in room identifiers exists in the stored list and return an error otherwise?

crates/matrix-sdk-ui/src/spaces/mod.rs
stefanceriu: How do I do that, `space_state: Arc<Mutex<(SpaceGraph, ObservableVector<SpaceRoom>)>>`-ish?

bindings/matrix-sdk-ffi/src/spaces.rs
stefanceriu: On the other hand, I believe that's going to make uglier to use in the client where they need to present a list of rooms, let the user chose and then invoke the leaving. Having to store that callback isn't amazing.

poljar: You should be able to, yes. How are you doing the room filters if not using a predicate/callback?

Something like this should™ work:

```rust
#[matrix_sdk_ffi_macros::export(callback_interface)]
pub trait SpaceFilter: Send + Sync {
    fn predicate(&self, space_room: LeaveSpaceRoom) -> bool;
}

...

/// Bulk leave the given rooms. Stops when encountering an error.
pub async fn leave(&self, predicate: Box<dyn SpaceFilter>) -> Result<(), ClientError> {
    self.inner
        .leave(|space_room| predicate.predicate(space_room.to_owned().into()))
        .await
        .map_err(ClientError::from)
}
```

crates/matrix-sdk-ui/src/spaces/mod.rs
poljar: Yeah, just put them into a struct instead of a tuple. Makes for more descriptive names and avoids the `<<(<>)>>` hieroglyphs a bit.

bindings/matrix-sdk-ffi/src/spaces.rs
poljar: > 

On the other hand, I believe that's going to make uglier to use in the client where they need to present a list of rooms, let the user chose and then invoke the leaving. Having to store that callback isn't amazing.

You don't need to store the callback, if you leave the `LeaveSpaceHandle` and the `LeaveSpaceHandle::rooms()` method you can still store the room IDs, and then pass the room IDs into the predicate.

crates/matrix-sdk-ui/src/spaces/mod.rs
stefanceriu: It's gone! Curious that it didn't warn me that it's unused 🤔

poljar: Public enum, so the variant is `pub` as well. Which means someone outside of the crate might construct this error variant.

stefanceriu: Oh of course, makes perfect sense, thank you!

bindings/matrix-sdk-ffi/src/spaces.rs
stefanceriu: Talked about it offline, decided to go with filters on the SDK side and a room id list based solution for the FFI layer to keep thing simple.

crates/matrix-sdk-ui/src/spaces/leave.rs
```diff
@@ -78,12 +78,14 @@ impl LeaveSpaceHandle {
     }
 
     /// Bulk leave the given rooms. Stops when encountering an error.
-    pub async fn leave(&self, rooms_ids: Vec<&RoomId>) -> Result<(), Error> {
-        for room_id in rooms_ids {
-            if let Some(room) = self.client.get_room(room_id) {
+    pub async fn leave(&self, filter: impl FnMut(&LeaveSpaceRoom) -> bool) -> Result<(), Error> {
+        let rooms = self.rooms().await?;
```
poljar: Something I forgot to ask, would it make sense to compute this once and store it in `LeaveSpaceHandle` in the `LeaveSpaceHandle::new()` call? This way the call to `SpaceHandle::rooms()` would be cheaper and you could return a `&[LeaveSpaceRoom]` instead of a `Vec`.

bindings/matrix-sdk-ffi/src/spaces.rs
```diff
@@ -304,3 +318,56 @@ impl From<VectorDiff<UISpaceRoom>> for SpaceListUpdate {
...
+/// The `LeaveSpaceHandle` processes rooms to be left in the order they were
+/// provided by the [`SpaceService`] and annotates them with extra data to
+/// inform the leave process e.g. if the current user is the last room admin.
+///
+/// Once the upstream client decides what rooms should actually be left, the
+/// handle provides a method to execute that too.
+#[derive(uniffi::Object)]
+pub struct LeaveSpaceHandle {
+    inner: UILeaveSpaceHandle,
+}
```
poljar, line 336:
```suggestion
}

```

crates/matrix-sdk-ui/src/spaces/leave.rs
stefanceriu: I thought about it and had the space service level method as async to being with but then I figured this can be used to reload your room list if it ever happens to change in between when you requested the leave and when you actually interact with the `LeaveSpaceRoom` list.

Then again, if the data changes from underneath you the changes of the graph itself being different are high as well.

tl;dr; I think I'll do that, yes, thank you for the suggestion 😁

stefanceriu: Or maybe not `std::sync::MutexGuard<'_, matrix_sdk_ui::spaces::SpaceState> cannot be sent between threads safely
within {async block@bindings/matrix-sdk-ffi/src/spaces.rs:53:1: 53:33}, the trait std::marker::Send is not implemented for std::sync::MutexGuard<'_, matrix_sdk_ui::spaces::SpaceState>`

This is because leave space is aquiring a mutex on the space service state I guess.

poljar: Hmm, the error talks about the newly added `SpaceState` mutex. Can you show me the diff that causes this?

stefanceriu: Sure thing, pushed all of it on `stefan/leavingSpaceAsyncHandle` here https://github.com/matrix-org/matrix-rust-sdk/commit/21b011f7af16954375652212f69836dfe8812fce

poljar: Can't hold sync locks across await points:

```diff
diff --git a/crates/matrix-sdk-ui/src/spaces/mod.rs b/crates/matrix-sdk-ui/src/spaces/mod.rs
index 5f9745cab3..dce6917428 100644
--- a/crates/matrix-sdk-ui/src/spaces/mod.rs
+++ b/crates/matrix-sdk-ui/src/spaces/mod.rs
@@ -209,13 +209,15 @@
     /// Once the rooms to be left are chosen the handle can be used to leave
     /// them.
     pub async fn leave_space(&self, space_id: &RoomId) -> Result<LeaveSpaceHandle, Error> {
-        let space_state = self.space_state.lock();
+        let room_ids = {
+            let space_state = self.space_state.lock();
 
-        if !space_state.graph.has_node(space_id) {
-            return Err(Error::RoomNotFound(space_id.to_owned()));
-        }
+            if !space_state.graph.has_node(space_id) {
+                return Err(Error::RoomNotFound(space_id.to_owned()));
+            }
 
-        let room_ids = space_state.graph.flattened_bottom_up_subtree(space_id);
+            space_state.graph.flattened_bottom_up_subtree(space_id)
+        };
 
         let handle = LeaveSpaceHandle::new(self.client.clone(), room_ids).await;
 
```

stefanceriu: Oh I'm such a dummy, I got thrown by the FFI layer error 🤦‍♂️

crates/matrix-sdk-ui/src/spaces/leave.rs
```diff
@@ -0,0 +1,198 @@
...
+            }
+        }
+
+        Self { client, rooms }
+    }
+
+    /// A list of rooms to be left which next to normal [`SpaceRoom`] data also
+    /// include leave specific information.
+    pub fn rooms(&self) -> Vec<LeaveSpaceRoom> {
+        self.rooms.clone()
```
poljar: The clone isn't required here, though you might need it in the FFI layer.

PR #5732, opened 2025-10-01
refactor(sdk): Allow to send waveform for any audio message

zecakeh:
We plan to use this in Fractal to always be able to present a waveform for an audio message.

The second commit might be more controversial: it changes the format of the waveform from a list of `u16` between 0 and 1024 to a list of `f32` between 0 and 1. This is done because the value between 0 and 1024 used in the event is quite arbitrary (and they have changed in MSC3246 since then), most clients should end up with values between 0 and 1, and need to convert it for sending it. So this centralizes this conversion in the SDK.

### DIFF
--- a/bindings/matrix-sdk-ffi/CHANGELOG.md
+++ b/bindings/matrix-sdk-ffi/CHANGELOG.md
@@ -8,6 +8,10 @@ All notable changes to this project will be documented in this file.
 
 ### Breaking changes:
 
+- The `waveform` parameter in `Timeline::send_voice_message` format changed to a list of `f32`
+  between 0 and 1.
+  ([#5732](https://github.com/matrix-org/matrix-rust-sdk/pull/5732))
+
 - The `normalized_power_level` field has been removed from the `RoomMember`
   struct.
   ([#5635](https://github.com/matrix-org/matrix-rust-sdk/pull/5635))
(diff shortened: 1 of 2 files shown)


crates/matrix-sdk/src/attachment.rs
```diff
@@ -67,7 +67,9 @@ pub struct BaseAudioInfo {
     /// The file size of the audio clip in bytes.
     pub size: Option<UInt>,
     /// The waveform of the audio clip.
-    pub waveform: Option<Vec<u16>>,
+    ///
+    /// Must only includes values between 0 and 1.
```
bnjbvr: nit: `include`

Hywan (approved):
The changes look good to me. I'm a bit annoyed by the regression with the waveform format, but since it's a different type and it's documented in the `CHANGELOG`, I think it's fine.

Hywan:
Re-running the benchmarks. I suspect the result is misleading and could be due to noise.

@zecakeh can you fix the conflicts please?

bnjbvr (commented):
Thanks, we're on board with this change, and as you said it'll make the switch to the new MSC simpler. We might delay the landing of this PR, since we want to limit landing breaking changes at the FFI layer until the Matrix conference.

One comment I would have is: why to use `f32` and not `f64`? We'd have a much better precision when it comes to the [0; 1] range, leading to a better precision in the normalized range; and I don't think we're getting any notable speedup for using 32 bits instead of 64 bits here.

Would it make sense to change to a f64 in this patch?

zecakeh:
I don't mind changing to f64 but I chose f32 because it is precise enough to be converted to an integer value between 0 and 1024 and it should be enough to draw it too so I don't see what we gain by increasing the precision?

bnjbvr:
I wary of using float32s, because they can't represent finely small values, viz. values in the [0; 1] range. That is, many small float32 values would be normalized to the same int in the [0; 1024] range, so if you looked at a flat projection of the f32 range to the int10 range, you *might* see holes in the int range. Since float64 have so much more precision, it's unlikely that this problem would happen with those.

zecakeh:
I am not convinced that this is true, because f32 has a precision of [6 significant digits](https://doc.rust-lang.org/std/primitive.f32.html#associatedconstant.DIGITS), and for integer values between 0 and 1024 we only need 4.

zecakeh:
No problem, I'll rebase this after the Matrix conference then to solve the conflicts.

poljar:
> No problem, I'll rebase this after the Matrix conference then to solve the conflicts.

The EX releases for the conf have been done, you can now rebase so we can merge.

zecakeh:
> The EX releases for the conf have been done, you can now rebase so we can merge.

Done

zecakeh:
Oh I forgot to fix the typo.

zecakeh:
Done.

PR #5733, opened 2025-10-01
refactor(sdk): Use TextMessageEventContent to send a caption

zecakeh:
It doesn't make sense to send a formatted caption without a plain text caption so using TextMessageEventContent forces the latter to be present.

This also allows to use the helpful constructors of TextMessageEventContent.


poljar (approved):
Thanks, makes sense.

PR #5734, opened 2025-10-01
task(crypto): Warn API users to fetch device info before processing verification requests

andybalaam:


bindings/matrix-sdk-crypto-ffi/src/machine.rs
```diff
@@ -873,6 +873,14 @@ impl OlmMachine {
...
     ///   lower severity warnings and unverified identities are not decorated).
     /// * `decryption_settings` - The setting for decrypting messages.
+    ///
+    /// **Note**: If the supplied event is an `m.room.message` event with
+    /// `msgtype: m.key.verification.request`, then the device information for
+    /// the sending user must be up-to-date before calling this method
+    /// (otherwise, the request will be ignored). It is hard to guarantee this
+    /// is the case, but you can maximize your chances by explicitly making a
+    /// request to /keys/query for the user's device info, and processing the
+    /// response with [`OlmMachine::mark_request_as_sent`].
```
poljar: This is a bit heavy to do on every `decrypt_room_event()` call, don't you think?

I think it's better to discourage the usage of this with `handle_verification_events` set to `true`.

andybalaam: How about 21dceb52c3 ?

poljar: Yup.

Issue #5721, opened 2025-09-26
Wrong `cached_is_space` on clear cache / fresh install leads to broken space room filter until restart


SpiritCroc:
### Observed on:
- SchildiChat Next since merging Element 25.09.2 with [my own Rust-SDK fork](https://github.com/SchildiChat/matrix-rust-sdk/commit/1763ed972d150f7bc89284f5b1f1127e869c8091)
- Element X Android that I checked to confirm this happens too on an SDK version without my own changes, more precisely built today from develop with `org.matrix.rustcomponents:sdk-android:25.9.25"`, which I think is fairly up-to-date at 5b52c729a5dd1c373cc17569711310f280bb9181 if I see that correctly.

### What happens:
When clearing cache, spaces show up in the normal room list until app restart, after which the filter starts working as expected. This *does not* affect Element X's experimental spaces feature, which seems to work differently.
On SchildiChat Next it is worse, since we still use the same room list mechanism to also build the space list, which thus remains empty until restart.

### Investigation / workaround "fix"

On my fork I tracked the issue down to `cached_is_space`, i.e. this commit "fixes" it on my end: https://github.com/SchildiChat/matrix-rust-sdk/commit/e2feb756c5740d6231df41c8408bb968b6229738

This makes the impression to me like a regression in https://github.com/matrix-org/matrix-rust-sdk/pull/5684/commits to me. My linked commit is likely not the appropriate solution since it voids some of the intended performance improvements.

I probably won't have much time to figure out the proper fix for this soon, so here you have my report, hope it helps :)

Hywan:
Thanks for the bug report! I believe https://github.com/matrix-org/matrix-rust-sdk/pull/5735 might fix your issue.

SpiritCroc:
Can confirm this fixes it, thank you!

PR #5735, opened 2025-10-01
fix(ui): `RoomListItem` refreshes its `cache_is_space`

Hywan:
The `RoomListItem` caches some data. It's been introduced in #5684. When the cached data are refreshed, the `Room::is_space()` value wasn't refreshed because I thought it wasn't possible to change the type of a room. However, if the state is not fully loaded (because the `m.room.create` event isn't present), the result of `Room::is_space()`… can change. So this patch refreshes the `cached_is_space` value unconditionally.

---

* Fixes https://github.com/matrix-org/matrix-rust-sdk/issues/5721


stefanceriu (approved):
From my testing this indeed fixes the problem, nicely done 👍

Issue #5726, opened 2025-09-30
Invites no longer appear at the expected position in the room list.


pixlwave:
Discovered on Element X iOS which is currently using d6d720c01529b9aadb0518ef745ed5817a216824

**This invite was sent moments before taking the screenshot:**
<img width="390" height="844" alt="Image" src="https://github.com/user-attachments/assets/a7aed7b3-4674-4a76-aad4-e79a38d3cd07" />

Hywan:
I suspect that this invite room has no `LatestEventValue`. I'm trying to adjust

https://github.com/matrix-org/matrix-rust-sdk/blob/681b22142fd265e47caa9190a3f3c53997d47a10/crates/matrix-sdk/src/latest_events/latest_event.rs#L900

to see if it handles invites.

Hywan:
See my comment https://github.com/matrix-org/matrix-rust-sdk/pull/5736#issuecomment-3364764120 to know why https://github.com/matrix-org/matrix-rust-sdk/pull/5736 doesn't fully fix the issue. The `EventCache` isn't handling invite room yet, so the `LatestEventValue` cannot be computed, and the invite room cannot be sorted as expected in the `RoomList`:

https://github.com/matrix-org/matrix-rust-sdk/blob/a4bd36cbe88ba5c5b23752ca7228e14d0566ab82/crates/matrix-sdk/src/event_cache/mod.rs#L965-L966

ganfra:
I think we have the same issue when a room is just created, without any sent message, it won't appear at the top

Hywan:
Yes, it happens to me too. There is no `LatestEventValue` for freshly joined room. We should find a workaround for that case.

An idea would be to consider `m.room.member` (for the current user) as another valid state event candidate, so that we would get a `LatestEventValue::Remote(_)`:

https://github.com/matrix-org/matrix-rust-sdk/blob/11d430563b9479d8af2097cded25281e1e6db126/crates/matrix-sdk/src/latest_events/latest_event.rs#L1267

Thoughts?

pixlwave:
> An idea would be to consider m.room.member (for the current user) as another valid state event candidate, so that we would get a LatestEventValue::Remote(_):

Yep, sounds good to me! I think the old apps used this same trick of only considering `m.room.member` for the current user too.

pixlwave:
[...]

PR #5736, opened 2025-10-01
feat(sdk): `m.room.member` can be a latest event candidate

Hywan:
WIP

---

* Fix https://github.com/matrix-org/matrix-rust-sdk/issues/5726

### DIFF
--- a/crates/matrix-sdk/src/latest_events/latest_event.rs
+++ b/crates/matrix-sdk/src/latest_events/latest_event.rs
@@ -1267,6 +1267,21 @@ mod tests_latest_event_content {
             is not a candidate
         );
     }
+
+    #[test]
+    fn test_invite() {
+        use ruma::events::room::member::MembershipState;
+
+        assert_latest_event_content!(
+            event | event_factory | {
+                event_factory
+                    .member(user_id!("@mnt.io:matrix.org"))
+                    .membership(MembershipState::Invite)
+                    .into_event()
+            }
+            is a candidate
+        );
+    }
 }
 
 #[cfg(test)]


pixlwave:
I'm afraid they're still not shown in the right place for me:

Simulator Screenshot - iPhone 17 - 2025-10-02 at 11 38 51

Just a guess, but I thought we didn't have `m.room.member` as SSS required state in the room list?

crates/matrix-sdk/src/latest_events/latest_event.rs
```diff
@@ -87,7 +90,8 @@ impl LatestEvent {
     pub async fn update_with_event_cache(
         &mut self,
         room_event_cache: &RoomEventCache,
-        power_levels: &Option<(&UserId, RoomPowerLevels)>,
+        own_user_id: Option<&UserId>,
+        power_levels: &Option<RoomPowerLevels>,
```
bnjbvr: It's usually more idiomatic to use `Option<&T>` over `&Option<T>` (it's a clippy lint, that we haven't enabled as far as I know).

crates/matrix-sdk/src/latest_events/latest_event.rs
```diff
@@ -109,12 +116,14 @@ impl LatestEvent {
         &mut self,
         send_queue_update: &RoomSendQueueUpdate,
         room_event_cache: &RoomEventCache,
-        power_levels: &Option<(&UserId, RoomPowerLevels)>,
+        own_user_id: Option<&UserId>,
+        power_levels: &Option<RoomPowerLevels>,
```
bnjbvr: ditto here

crates/matrix-sdk/src/latest_events/latest_event.rs
```diff
@@ -485,28 +494,29 @@ impl LatestEventValueBuilder {
...
+        Self::new_remote_with_power_levels(room_event_cache, own_user_id, &power_levels).await
     }
 
     /// Create a new [`LatestEventValue::Remote`] based on existing power
     /// levels.
     async fn new_remote_with_power_levels(
         room_event_cache: &RoomEventCache,
-        power_levels: &Option<(&UserId, RoomPowerLevels)>,
+        own_user_id: Option<&UserId>,
+        power_levels: &Option<RoomPowerLevels>,
```
bnjbvr: and here

crates/matrix-sdk/src/latest_events/latest_event.rs
```diff
@@ -519,7 +529,8 @@ impl LatestEventValueBuilder {
         send_queue_update: &RoomSendQueueUpdate,
         buffer_of_values_for_local_events: &mut LatestEventValuesForLocalEvents,
         room_event_cache: &RoomEventCache,
-        power_levels: &Option<(&UserId, RoomPowerLevels)>,
+        own_user_id: Option<&UserId>,
+        power_levels: &Option<RoomPowerLevels>,
```
bnjbvr: and here

crates/matrix-sdk/src/latest_events/latest_event.rs
```diff
@@ -705,12 +721,13 @@ impl LatestEventValueBuilder {
     async fn new_local_or_remote(
         buffer_of_values_for_local_events: &mut LatestEventValuesForLocalEvents,
         room_event_cache: &RoomEventCache,
-        power_levels: &Option<(&UserId, RoomPowerLevels)>,
+        own_user_id: Option<&UserId>,
+        power_levels: &Option<RoomPowerLevels>,
```
bnjbvr: and… here!

crates/matrix-sdk/src/latest_events/latest_event.rs
```diff
@@ -899,7 +916,8 @@ impl LatestEventValuesForLocalEvents {
 
 fn filter_timeline_event(
     event: &TimelineEvent,
-    power_levels: &Option<(&UserId, RoomPowerLevels)>,
+    own_user_id: Option<&UserId>,
+    power_levels: &Option<RoomPowerLevels>,
```
bnjbvr: … you see where this is going

crates/matrix-sdk/src/latest_events/latest_event.rs
```diff
@@ -970,14 +990,15 @@ fn filter_any_message_like_event_content(event: AnyMessageLikeEventContent) -> b
 
 fn filter_any_sync_state_event(
     event: AnySyncStateEvent,
-    power_levels: &Option<(&UserId, RoomPowerLevels)>,
+    own_user_id: Option<&UserId>,
+    power_levels: &Option<RoomPowerLevels>,
```
bnjbvr: lalala

crates/matrix-sdk/src/latest_events/latest_event.rs
```diff
@@ -1017,8 +1017,16 @@ fn filter_any_sync_state_event(
...
+                    // The current _is_ invited (not someone else).
+                    match member {
+                        // We can only decide whether the user is invited if the event isn't
+                        // redacted.
+                        SyncStateEvent::Original(state) => {
+                            Some(state.state_key.deref()) == own_user_id
+                        }
+
+                        _ => false,
+                    }
```
bnjbvr, line 1030: I like it, but where was this behavior requested, out of curiosity?

crates/matrix-sdk/src/latest_events/latest_event.rs
```diff
@@ -993,6 +988,55 @@ fn filter_any_message_like_event_content(event: AnyMessageLikeEventContent) -> b
...
+    power_levels: &Option<RoomPowerLevels>,
+) -> bool {
+    match event {
+        AnySyncStateEvent::RoomMember(member) => {
+            match member.membership() {
+                MembershipState::Knock => {
+                    let can_accept_or_decline_knocks = match (own_user_id, power_levels) {
+                        (Some(own_user_id), Some(room_power_levels)) => {
+                            room_power_levels.user_can_invite(own_user_id)
+                                || room_power_levels.user_can_kick(own_user_id)
```
zecakeh, line 1004: This code was already there, but it should be:

```suggestion
                                || room_power_levels.user_can_kick_user(own_user_id, member.state_key())
```

Another option is to always use `RoomPowerLevels::can_do_to_user` whenever performing an action on another user's event, it will always perform the same checks as the authorization rules.

In this case, although it is unlikely, a user wouldn't be able to reject a knock from another user that has the same or a higher power level than them.

For example:

1. A mod leaves the room, but the power level event isn't changed.
2. The mod knocks.
 
Another mod in the room with the same power level cannot reject the knock.

crates/matrix-sdk/src/latest_events/latest_event.rs
```diff
@@ -993,6 +988,55 @@ fn filter_any_message_like_event_content(event: AnyMessageLikeEventContent) -> b
...
+                    false
+                }
+
+                MembershipState::Invite => {
+                    // The current _is_ invited (not someone else).
+                    match member {
+                        // We can only decide whether the user is invited if the event isn't
+                        // redacted.
+                        SyncStateEvent::Original(state) => {
+                            Some(state.state_key.deref()) == own_user_id
```
zecakeh, line 1026: In this case you wouldn't get the event in the timeline, but in the stripped state of the room. You would only be able to see this after joining the room.

Hywan:
> I'm afraid they're still not shown in the right place for me: […]
>
> Just a guess, but I thought we didn't have `m.room.member` as SSS required state in the room list?

I've found why. The computation of a `LatestEventValue` is triggered by either the `SendQueue` or the `EventCache`. But the `EventCache` doesn't handle invite room yet:

https://github.com/matrix-org/matrix-rust-sdk/blob/a4bd36cbe88ba5c5b23752ca7228e14d0566ab82/crates/matrix-sdk/src/event_cache/mod.rs#L965-L966

So we need a bit more work here.

crates/matrix-sdk/src/latest_events/latest_event.rs
Hywan: I've updated the description of the PR:

> The idea behind this patch is to compute a `LatestEventValue` for a invite room, so that the room can be sorted correctly in the room list by [the `recency` sorter](https://matrix-org.github.io/matrix-rust-sdk/matrix_sdk_ui/room_list_service/sorters/fn.new_sorter_recency.html). That's a necessary step to solve https://github.com/matrix-org/matrix-rust-sdk/issues/5726.

crates/matrix-sdk/src/latest_events/latest_event.rs
Hywan: This API receives events from the `EventCache`, it's not implemented yet (see https://github.com/matrix-org/matrix-rust-sdk/pull/5736#issuecomment-3364764120), but I think we should broadcast the stripped state events too.

crates/matrix-sdk/src/latest_events/latest_event.rs
Hywan: Thanks!

Hywan: Addressed in https://github.com/matrix-org/matrix-rust-sdk/pull/5743

PR #5737, opened 2025-10-01
feat: Append withheld info from room key bundle to store.

kaylendog:
Part of https://github.com/element-hq/element-meta/issues/2876

Signed-off-by:  Skye Elliot

### DIFF
--- a/crates/matrix-sdk-crypto/src/machine/mod.rs
+++ b/crates/matrix-sdk-crypto/src/machine/mod.rs
@@ -1003,7 +1003,9 @@ impl OlmMachine {
 
         if let RoomKeyWithheldContent::MegolmV1AesSha2(
             MegolmV1AesSha2WithheldContent::BlackListed(c)
-            | MegolmV1AesSha2WithheldContent::Unverified(c),
+            | MegolmV1AesSha2WithheldContent::Unverified(c)
+            | MegolmV1AesSha2WithheldContent::Unauthorised(c)
+            | MegolmV1AesSha2WithheldContent::Unavailable(c),
         ) = &event.content
         {
             changes
--- a/crates/matrix-sdk-crypto/src/store/mod.rs
+++ b/crates/matrix-sdk-crypto/src/store/mod.rs
@@ -59,7 +59,7 @@ use serde::{de::DeserializeOwned, Serialize};
 use thiserror::Error;
 use tokio::sync::{Mutex, Notify, OwnedRwLockWriteGuard, RwLock};
 use tokio_stream::wrappers::errors::BroadcastStreamRecvError;
-use tracing::{error, info, instrument, trace, warn};
+use tracing::{info, instrument, trace, warn};
 use types::{RoomKeyBundleInfo, StoredRoomKeyBundleData};
 use vodozemac::{megolm::SessionOrdering, Curve25519PublicKey};
 
@@ -76,6 +76,7 @@ use crate::{
         Account, ExportedRoomKey, InboundGroupSession, PrivateCrossSigningIdentity, SenderData,
         Session, StaticAccountData,
     },
+    store::types::RoomKeyWithheldEntry,
     types::{
         BackupSecrets, CrossSigningSecrets, MegolmBackupV1Curve25519AesSha2Secrets, RoomKeyExport,
         SecretsBundle,
@@ -1645,6 +1646,7 @@ impl Store {
         }
 
         self.import_room_key_bundle_sessions(bundle_info, &bundle, progress_listener).await?;
+        self.import_room_key_bundle_withheld_info(bundle_info, &bundle).await?;
 
         Ok(())
     }
@@ -1700,6 +1702,43 @@ impl Store {
         Ok(())
     }
 
+    async fn import_room_key_bundle_withheld_info(
+        &self,
+        bundle_info: &StoredRoomKeyBundleData,
+        bundle: &RoomKeyBundle,
+    ) -> Result<(), CryptoStoreError> {
+        let mut changes = Changes::default();
+        for withheld in &bundle.withheld {
+            let (room_id, session_id) = match withheld {
+                RoomKeyWithheldContent::MegolmV1AesSha2(c) => match (c.room_id(), c.session_id()) {
+                    (Some(room_id), Some(session_id)) => (room_id, session_id),
+                    _ => continue,
+                },
+                #[cfg(feature = "experimental-algorithms")]
+                RoomKeyWithheldContent::MegolmV2AesSha2(c) => match (c.room_id(), c.session_id()) {
+                    (Some(room_id), Some(session_id)) => (room_id, session_id),
+                    _ => continue,
+                },
+                RoomKeyWithheldContent::Unknown(_) => continue,
+            };
+
+            if room_id != bundle_info.bundle_data.room_id {
+                trace!("Ignoring withheld info for incorrect room {} in bundle", room_id);
+                continue;
+            }
+
+            changes.withheld_session_info.entry(room_id.to_owned()).or_default().insert(
+                session_id.to_owned(),
+                RoomKeyWithheldEntry {
+                    sender: bundle_info.sender_user.clone(),
+                    content: withheld.to_owned(),
+                },
+            );
+        }
+        self.save_changes(changes).await?;
+
+        Ok(())
+    }
 }
 
 impl Deref for Store {
@@ -1735,16 +1774,25 @@ mod tests {
     use futures_util::StreamExt;
     use insta::{_macro_support::Content, assert_json_snapshot, internals::ContentPath};
     use matrix_sdk_test::async_test;
-    use ruma::{device_id, owned_device_id, room_id, user_id, RoomId};
+    use ruma::{
+        device_id,
+        events::room::{EncryptedFileInit, JsonWebKeyInit},
+        owned_device_id, owned_mxc_uri, room_id,
+        serde::Base64,
+        user_id, RoomId,
+    };
     use serde_json::json;
     use vodozemac::megolm::SessionKey;
 
     use crate::{
         machine::test_helpers::get_machine_pair,
         olm::{InboundGroupSession, SenderData},
-        store::types::{DehydratedDeviceKey, RoomKeyWithheldEntry},
+        store::types::{DehydratedDeviceKey, RoomKeyWithheldEntry, StoredRoomKeyBundleData},
         types::{
-            events::room_key_withheld::{MegolmV1AesSha2WithheldContent, RoomKeyWithheldContent},
+            events::{
+                room_key_bundle::RoomKeyBundleContent,
+                room_key_withheld::{MegolmV1AesSha2WithheldContent, RoomKeyWithheldContent},
+            },
             EventEncryptionAlgorithm,
         },
         OlmMachine,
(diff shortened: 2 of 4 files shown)


crates/matrix-sdk-crypto/src/types/events/room_key_withheld.rs
```diff
@@ -17,14 +17,57 @@
...
+    /// A to-device event containing withheld room key information.
+    ToDevice(RoomKeyWithheldEvent),
+    /// Content held within a room key bundle.
+    Bundle {
+        /// The ID of the user who sent the bundle.
+        sender: OwnedUserId,
+        /// The contents of a single withheld entry in the bundle.
+        content: RoomKeyWithheldContent,
+    },
+}
```
richvdh: I think this type probably belongs in `matrix_sdk_crypto::store::types` rather than here.

richvdh: Also: using an enum gives a bit of a confusing api for `Changes::withheld_session_info` and `CryptoStore::get_withheld_info`. Why not just use the `Bundle` format, i.e:

```
pub struct RoomKeyWithheldEntry {
    sender: OwnedUserId,
    content: RoomKeyWithheldContent,
}
```

I *think* that, as it happens, data in the db using the *existing* format can also be deserialized into that (we'll need a test), so we don't need any migration logic. But if necessary, you can add a deserialization helper type which *is* an enum, but which is immediately transformed into a public `RoomKeyWithheldEntry`.

kaylendog: I'm seeing some regression tests for serialising/deserialising to the store. Would these suffice, or should I introduce more?

richvdh: > I'm seeing some regression tests for serialising/deserialising to the store. Would these suffice, or should I introduce more?

Serialize-and-deserialize would be insufficient, because the "serialize" step would use the new format. IIRC there are some tests that use `insta` to store pre-serialized JSON blobs containing legacy formats, you probably want something like that.

It doesn't even necessarily need to hit the store, provided you can test that an old-format object can be successfully deserialized into a new-format object.

kaylendog: This would be fine if I serialise the old format, though, right? The test I wrote converts a to-device `RoomKeyWithheldEvent` (the previous format) to a `serde_json::Value`, then tries to deserialise this back to the new `RoomKeyWithheldEntry`.

I'm investigating `insta`, but it seems more geared towards comparing against existing snapshots, rather than loading a snapshot and testing if it satisfies a given condition.

richvdh: yeah, sorry, `insta` isn't the right thing here. It looks like the tests I'm thinking of just hardocode a `json!({ ... })` value and test that it deserializes as expected.

> This would be fine if I serialise the old format, though, right? The test I wrote converts a to-device `RoomKeyWithheldEvent` (the previous format) to a `serde_json::Value`, then tries to deserialise this back to the new `RoomKeyWithheldEntry`.

Yeah I guess that's fine in this case, as long as we can be sure that serializing a `RoomKeyWithheldEvent` continues to give the same format as whatever people have in their database today. I kinda think that hardcoding a `json!({...})` tests this more explicitly and robustly, but it's not a strong opinion.

Your test could do with some comments explaining *why* that's an important thing to test though (i.e. that the format has changed and we need to make sure that the old format is still handled correctly)

andybalaam: +1 for being explicit in tests i.e. using `json!`

https://goodenoughtesting.com/articles/tests-should-be-simple

kaylendog: Updated to use `json`!

crates/matrix-sdk-crypto/src/machine/mod.rs
```diff
@@ -998,7 +998,7 @@ impl OlmMachine {
         Ok(())
     }
 
-    fn add_withheld_info(&self, changes: &mut Changes, event: &RoomKeyWithheldEvent) {
+    pub(crate) fn add_withheld_info(&self, changes: &mut Changes, event: &RoomKeyWithheldEvent) {
```
richvdh: if we're going to increase the visibility of this, we should really add some docs... but do we need to increase the visibility?

crates/matrix-sdk-crypto/src/store/types.rs
```diff
@@ -486,7 +489,23 @@ pub struct RoomKeyWithheldInfo {
 
     /// The `m.room_key.withheld` event that notified us that the key is being
     /// withheld.
```
richvdh: this comment isn't really accurate any more, since it may not have been an  `m.room_key.withheld`

crates/matrix-sdk-crypto/src/store/types.rs
```diff
@@ -486,7 +489,23 @@ pub struct RoomKeyWithheldInfo {
...
-    pub withheld_event: RoomKeyWithheldEvent,
+    pub withheld_event: RoomKeyWithheldEntry,
+}
+
+/// Represents an entry for a withheld room key event, which can be either a
+/// to-device event or a bundle entry.
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct RoomKeyWithheldEntry {
+    /// The ID of the user who sent the bundle.
```
richvdh: may not have been a bundle

crates/matrix-sdk-crypto/src/store/types.rs
```diff
@@ -486,7 +489,23 @@ pub struct RoomKeyWithheldInfo {
...
+    pub sender: OwnedUserId,
+    /// The contents of a single withheld entry in the bundle.
```
richvdh: may not have been a bundle

crates/matrix-sdk-crypto/src/store/mod.rs
```diff
@@ -1681,10 +1681,30 @@ impl Store {
...
+        let mut changes = Changes::default();
+        for withheld in &bundle.withheld {
+            if let RoomKeyWithheldContent::MegolmV1AesSha2(
+                MegolmV1AesSha2WithheldContent::BlackListed(c)
+                | MegolmV1AesSha2WithheldContent::Unverified(c)
+                | MegolmV1AesSha2WithheldContent::Unauthorised(c)
+                | MegolmV1AesSha2WithheldContent::Unavailable(c),
+            ) = withheld
+            {
+                changes.withheld_session_info.entry(c.room_id.to_owned()).or_default().insert(
```
richvdh: I think we should check that c.room_id matches the expected room id, as we do with the received keys.

crates/matrix-sdk-crypto/src/store/mod.rs
```diff
@@ -1681,10 +1681,30 @@ impl Store {
...
+                    RoomKeyWithheldEntry {
+                        sender: bundle_info.sender_user.clone(),
+                        content: withheld.to_owned(),
+                    },
+                );
             }
         }
+        self.save_changes(changes).await?;
+
+        Ok(())
```
richvdh, line 1740: this method is getting unpleasantly long: probably time to find a way to break it up?

crates/matrix-sdk-crypto/src/store/mod.rs
```diff
@@ -1681,10 +1681,30 @@ impl Store {
...
+        }
 
-                Ok(())
+        let mut changes = Changes::default();
+        for withheld in &bundle.withheld {
+            if let RoomKeyWithheldContent::MegolmV1AesSha2(
+                MegolmV1AesSha2WithheldContent::BlackListed(c)
+                | MegolmV1AesSha2WithheldContent::Unverified(c)
+                | MegolmV1AesSha2WithheldContent::Unauthorised(c)
+                | MegolmV1AesSha2WithheldContent::Unavailable(c),
```
richvdh: it feels like `MegolmV1AesSha2WithheldContent` should have `session_id` and `room_id` getters, which can take care of this mess.

crates/matrix-sdk-crypto/src/store/mod.rs
```diff
@@ -1681,10 +1681,30 @@ impl Store {
                         self.import_sessions_impl(good, None, progress_listener).await?;
                     }
                 }
+            }
+        }
 
-                Ok(())
+        let mut changes = Changes::default();
+        for withheld in &bundle.withheld {
+            if let RoomKeyWithheldContent::MegolmV1AesSha2(
```
richvdh: probably want a `match` here rather than an `if let`, so that when we introduce a new type of `RoomKeyWithheldContent` we don't forget about this code.

richvdh: Relatedly, we probably want to do the same for `MegolmV2AesSha2`, if the relevant feature flag is set.

crates/matrix-sdk-crypto/src/store/mod.rs
```diff
@@ -1969,6 +2003,136 @@ mod tests {
...
+        assert_eq!(sender, "@alice:s.co");
+        assert_eq!(withheld_content.room_id, "!roomid:s.co");
+        assert_eq!(withheld_content.session_id, "session123");
+        assert_eq!(
+            withheld_content.sender_key.to_base64(),
+            "7hIcOrEroXYdzjtCBvBjUiqvT0Me7g+ymeXqoc65RS0"
+        );
+        assert_eq!(withheld_content.from_device, Some(owned_device_id!("ALICE")));
+    }
     /// Create an inbound Megolm session for the given room.
```
richvdh, line 2192:
```suggestion

    /// Create an inbound Megolm session for the given room.
```

richvdh (changes requested):

crates/matrix-sdk-crypto/src/machine/mod.rs
kaylendog: Hmm, latest revision doesn't use this method at all, so i'll just revert this.

crates/matrix-sdk-crypto/src/store/snapshots/matrix_sdk_crypto__store__tests__build_room_key_bundle.snap
```diff
@@ -5,7 +5,7 @@ expression: bundle
 {
   "room_keys": [
     {
-      "algorithm": "m.megolm.v1.aes-sha2",
+      "algorithm": "[algorithm]",
```
kaylendog, line 8: Fixes tests when running with `experimental-algorithms` feature enabled.

kathleenmotley49-maker (approved):

crates/matrix-sdk-crypto/src/store/types.rs
```diff
@@ -487,18 +487,24 @@ pub struct RoomKeyWithheldInfo {
     /// The ID of the session that the key is for.
     pub session_id: String,
 
-    /// The `m.room_key.withheld` event that notified us that the key is being
-    /// withheld.
+    /// The withheld entry from a `m.room_key.withheld` event or [MSC4268] room
+    /// key bundle.
+    ///
+    /// [MSC4268]: https://github.com/matrix-org/matrix-spec-proposals/pull/4362
```
richvdh: 4362 != 4268

crates/matrix-sdk-crypto/src/store/types.rs
```diff
@@ -487,18 +487,24 @@ pub struct RoomKeyWithheldInfo {
...
 
 /// Represents an entry for a withheld room key event, which can be either a
 /// to-device event or a bundle entry.
 #[derive(Clone, Debug, Serialize, Deserialize)]
 pub struct RoomKeyWithheldEntry {
-    /// The ID of the user who sent the bundle.
+    /// The user ID responsible for this entry, either from a
+    /// `m.room_key.withheld` to-device event or an [MSC4268] room key bundle.
+    ///
+    /// [MSC4268]: https://github.com/matrix-org/matrix-spec-proposals/pull/4362
```
richvdh: 4362 != 4268

crates/matrix-sdk-crypto/src/store/mod.rs
```diff
@@ -1632,59 +1633,112 @@ impl Store {
...
+                RoomKeyWithheldContent::MegolmV1AesSha2(c) => match (c.room_id(), c.session_id()) {
+                    (Some(room_id), Some(session_id)) => (room_id, session_id),
+                    _ => continue,
+                },
+                #[cfg(feature = "experimental-algorithms")]
+                RoomKeyWithheldContent::MegolmV2AesSha2(c) => match (c.room_id(), c.session_id()) {
+                    (Some(room_id), Some(session_id)) => (room_id, session_id),
+                    _ => continue,
+                },
+                _ => continue,
```
richvdh: can we make this an explicit match, so that we remember to add handling for new variants?

```suggestion
                RoomKeyWithheldContent::Unknown(_) => continue,
```

crates/matrix-sdk-crypto/src/store/mod.rs
```diff
@@ -2150,7 +2207,10 @@ mod tests {
             room_id,
             session_key,
             SenderData::unknown(),
+            #[cfg(not(feature = "experimental-algorithms"))]
             EventEncryptionAlgorithm::MegolmV1AesSha2,
+            #[cfg(feature = "experimental-algorithms")]
+            EventEncryptionAlgorithm::MegolmV2AesSha2,
```
richvdh, line 2218: could do with documenting this behaviour in the doc-comment.

richvdh (changes requested):
a few little nits

PR #5740, opened 2025-10-02
fix(docs): Replace doc_auto_cfg with doc_cfg feature

zecakeh:
The former has been merged in the latter, and it errors when generating the docs in a recent version of nightly, like the one used on docs.rs.

This also requires to bump the version of nightly used in CI, otherwise it would break the docs generation.

### DIFF
--- a/crates/matrix-sdk-base/src/event_cache/store/memory_store.rs
+++ b/crates/matrix-sdk-base/src/event_cache/store/memory_store.rs
@@ -237,6 +237,7 @@ impl EventCacheStore for MemoryStore {
 }
 
 #[cfg(test)]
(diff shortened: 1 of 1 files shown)


zecakeh:
I am not sure how to solve the remaining error in CI. It seems to only occur when running clippy with `--all-targets`.

poljar:
> I am not sure how to solve the remaining error in CI. It seems to only occur when running clippy with `--all-targets`.

I think silencing this using an explicit allow sounds fine, it's a strange occurrence indeed.

zecakeh:
Done!

stefanceriu (approved):
👌

PR #5741, opened 2025-10-03
feat(spaces): sort space room list rooms as defined in the spec

stefanceriu:
The ordering criteria is defined at https://spec.matrix.org/latest/client-server-api/#ordering-of-children-within-a-space. The gist is that `order` comes first, then `timestamp` and finally the `room_id`

This is not available for top level spaces, but there is an MSC that addresses it at matrix-org/matrix-spec-proposals#3230 and Ruma support has been added in ruma/ruma#2231. The SDK side implementation for that will come in a later PR.


crates/matrix-sdk-ui/src/spaces/room_list.rs
```diff
@@ -346,10 +353,41 @@ impl SpaceRoomList {
...
+        let a_state = children_state.get(&a.room_id);
+        let b_state = children_state.get(&b.room_id);
+
+        match (a_state, b_state) {
+            (Some(a_state), Some(b_state)) => {
+                match (&a_state.content.order, &b_state.content.order) {
+                    (Some(a_order), Some(b_order)) => a_order
+                        .cmp(b_order)
+                        .then(a_state.origin_server_ts.cmp(&b_state.origin_server_ts))
+                        .then(a.room_id.cmp(&b.room_id)),
```
poljar, line 373: This turned out to be quite neat.

stefanceriu: Yup yup!

stefanceriu:
Supa nice, thank you for the guidance!

Issue #5362, opened 2025-07-08
SQLite stores sporadically fail with `database is locked` errors


bnjbvr:
We sometimes observe in the apps a `database is locked` error, in the SQLite backends. This seems to happen quite a bit with the state store and the event cache store, as far as Sentry reports tell us (we've had thousands of occurrences of these errors there in the last few weeks only, after [adding Sentry support](https://github.com/matrix-org/matrix-rust-sdk/pull/5065)).

When this error happens, it can cause several issues. For instance, when this is happening for the event cache store, in the best case we can miss loading a media, but in the worst case this can lead to desynchronization of the in-memory representation of a timeline and the on-disk persisted representation of the same timeline, leading to further confusing issues (best case: duplication of or missing events; worst case: panics). When this happens with the state store, this can lead to some sync data being missing (fortunately the sync should be retried after it, so it's partially mitigated *in this case*).

We've already enabled [WAL mode](https://www.sqlite.org/wal.html), which is the de-facto solution to allow concurrent accesses to a database (that was needed at least for iOS, which may run multiple processes, all trying to read into the same databases).

Unfortunately, this error can happen for lots and lots of reasons, including but not limited to:

- a read-only transaction is upgraded to a write-transaction
- another database connection has the database mode open in [exclusive locking mode](https://www.sqlite.org/pragma.html#pragma_locking_mode) then all queries against the database will return [SQLITE_BUSY](https://www.sqlite.org/rescode.html#busy).
- when the last connection to a particular database is closing, that connection will acquire an exclusive lock for a short time while it cleans up the WAL and shared-memory files. If a separate attempt is made to open and query the database while the first connection is still in the middle of its cleanup process, the second connection might get an [SQLITE_BUSY](https://www.sqlite.org/rescode.html#busy) error.
- if the last connection to a database crashed, then the first new connection to open the database will start a recovery process. An exclusive lock is held during recovery. So if a third database connection tries to jump in and query while the second connection is running recovery, the third connection will get an [SQLITE_BUSY](https://www.sqlite.org/rescode.html#busy) error.
- and many searches on the web will show some other sources and reasons where this can happen.

We've tried setting up the busy timeout in the past, but it didn't seem to help, so we've reverted that.

## possible ways forward

- retry the busy timeout with higher values
- implement misc retry mechanisms, either at the database layer, or the applicative layer
[...]

PR #5744, opened 2025-10-03
feat(sqlite): `SqliteStateStore` has 1 write connection

Hywan:
This patch introduces a write-only connection in `SqliteStateStore` _à la_ `SqliteEventCacheStore` (see https://github.com/matrix-org/matrix-rust-sdk/pull/5382 to learn more). The idea is to get many read-only connections, and a single write-only connections behind a lock, so that there is a single writer at a time.

This patch renames the `acquire` method to `read`, and it introduces a new `write` connection.

---

- Address https://github.com/element-hq/customer-success/issues/591
- May fix https://github.com/matrix-org/matrix-rust-sdk/issues/5362

### DIFF
--- a/crates/matrix-sdk-sqlite/src/event_cache_store.rs
+++ b/crates/matrix-sdk-sqlite/src/event_cache_store.rs
@@ -165,7 +165,7 @@ impl SqliteEventCacheStore {
         })
     }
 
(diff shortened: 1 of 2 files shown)


poljar (approved):
Looks good.

PR #5745, opened 2025-10-03
refactor(tests): Use EventFactory for some tests

zecakeh:
There is a breaking change that landed in Ruma and those types are now non-exhaustive so they can't be built with the struct declaration anymore.


zecakeh:
I fixed the linting error found by clippy.

PR #5746, opened 2025-10-03
Implement the redecryption logic in the event cache

poljar:
This is very much a WIP PR. Eventually it will get split into different smaller PRs.

Things that still need to be addressed:

* [ ] Don't load all events for a certain room when attempting to redecrypt events.
* [ ] Decide what to do when the room key stream has lagged.
* [ ] Connect the UTD hook to the new redecryptor
* [ ] Completely remove the timeline redecryption logic

### DIFF
--- a/crates/matrix-sdk/src/event_cache/mod.rs
+++ b/crates/matrix-sdk/src/event_cache/mod.rs
@@ -935,7 +935,7 @@ impl EventCacheInner {
             self.multiple_room_updates_lock.lock().await
         };
 
-        // Note: bnjbvr tried to make this concurrent at some point, but it turned out
+        // NOTE: bnjbvr tried to make this concurrent at some point, but it turned out
         // to be a performance regression, even for large sync updates. Lacking
         // time to investigate, this code remains sequential for now. See also
         // https://github.com/matrix-org/matrix-rust-sdk/pull/5426.
--- a/crates/matrix-sdk/src/event_cache/mod.rs
+++ b/crates/matrix-sdk/src/event_cache/mod.rs
@@ -74,6 +74,7 @@ use crate::{
 
 mod deduplicator;
 mod pagination;
+mod redecryptor;
 mod room;
 
 pub use pagination::{RoomPagination, RoomPaginationStatus};
--- a/crates/matrix-sdk/src/event_cache/mod.rs
+++ b/crates/matrix-sdk/src/event_cache/mod.rs
@@ -69,6 +69,7 @@ use tracing::{Instrument as _, Span, debug, error, info, info_span, instrument, 
 use crate::{
     Client,
     client::WeakClient,
+    event_cache::redecryptor::Redecryptor,
     send_queue::{LocalEchoContent, RoomSendQueueUpdate, SendQueueUpdate},
 };
 
@@ -149,6 +150,9 @@ pub struct EventCacheDropHandles {
 
     /// The task used to automatically shrink the linked chunks.
     auto_shrink_linked_chunk_task: JoinHandle<()>,
+
+    /// The task used to automatically redecrypt UTDs.
+    redecryption_task: JoinHandle<()>,
 }
 
 impl fmt::Debug for EventCacheDropHandles {
@@ -162,6 +166,7 @@ impl Drop for EventCacheDropHandles {
         self.listen_updates_task.abort();
         self.ignore_user_list_update_task.abort();
         self.auto_shrink_linked_chunk_task.abort();
+        self.redecryption_task.abort();
     }
 }
 
@@ -258,10 +263,13 @@ impl EventCache {
                 auto_shrink_receiver,
             ));
 
+            let redecryption_task = Redecryptor::new(client, Arc::downgrade(&self.inner));
+
             Arc::new(EventCacheDropHandles {
                 listen_updates_task,
                 ignore_user_list_update_task,
                 auto_shrink_linked_chunk_task,
+                redecryption_task,
             })
         });
 
--- a/crates/matrix-sdk/src/event_cache/mod.rs
+++ b/crates/matrix-sdk/src/event_cache/mod.rs
@@ -152,7 +152,7 @@ pub struct EventCacheDropHandles {
     auto_shrink_linked_chunk_task: JoinHandle<()>,
 
     /// The task used to automatically redecrypt UTDs.
-    redecryption_task: JoinHandle<()>,
+    redecryptor: Redecryptor,
 }
 
 impl fmt::Debug for EventCacheDropHandles {
@@ -166,7 +166,6 @@ impl Drop for EventCacheDropHandles {
         self.listen_updates_task.abort();
         self.ignore_user_list_update_task.abort();
         self.auto_shrink_linked_chunk_task.abort();
-        self.redecryption_task.abort();
     }
 }
 
@@ -263,13 +262,13 @@ impl EventCache {
                 auto_shrink_receiver,
             ));
 
-            let redecryption_task = Redecryptor::new(client, Arc::downgrade(&self.inner));
+            let redecryptor = Redecryptor::new(client, Arc::downgrade(&self.inner));
 
             Arc::new(EventCacheDropHandles {
                 listen_updates_task,
                 ignore_user_list_update_task,
                 auto_shrink_linked_chunk_task,
-                redecryption_task,
+                redecryptor,
             })
         });
 
--- a/crates/matrix-sdk/src/event_cache/mod.rs
+++ b/crates/matrix-sdk/src/event_cache/mod.rs
@@ -262,7 +262,7 @@ impl EventCache {
                 auto_shrink_receiver,
             ));
 
-            let redecryptor = Redecryptor::new(client, Arc::downgrade(&self.inner));
+            let redecryptor = Redecryptor::new(Arc::downgrade(&self.inner));
 
             Arc::new(EventCacheDropHandles {
                 listen_updates_task,
--- a/crates/matrix-sdk/src/event_cache/redecryptor.rs
+++ b/crates/matrix-sdk/src/event_cache/redecryptor.rs
@@ -32,7 +32,9 @@ use tracing::{info, instrument, trace, warn};
 
 use crate::{
     Client,
-    event_cache::{EventCache, EventCacheError, EventCacheInner},
+    event_cache::{
+        EventCache, EventCacheError, EventCacheInner, EventsOrigin, RoomEventCacheUpdate,
+    },
 };
 
 pub(crate) trait RedecryptorCtx {
@@ -57,16 +59,18 @@ impl RedecryptorCtx for EventCache {
         &self,
         room_key_info: &RoomKeyInfo,
     ) -> Result<Vec<(OwnedEventId, Raw<AnySyncTimelineEvent>)>, Self::Error> {
-        let only_utd_events = |event: TimelineEvent| {
-            // We need the event ID to be able to replace the event.
+        let filter_non_utds = |event: TimelineEvent| {
             let event_id = event.event_id();
             // We only care about events fort his particular room key, identified by the
             // session ID.
             let session_id = event.kind.session_id();
 
             if session_id == Some(&room_key_info.session_id) {
-                // Only pick out events that are UTDs.
+                // Only pick out events that are UTDs, get just the Raw event as this is what
+                // the OlmMachine needs.
                 let event = as_variant!(event.kind, TimelineEventKind::UnableToDecrypt { event, .. } => event);
+                // Zip the event ID and event together so we don't have to pick out the event ID
+                // again. We need the event ID to replace the event in the cache.
                 event_id.zip(event)
             } else {
                 None
@@ -80,9 +84,10 @@ impl RedecryptorCtx for EventCache {
         let store = self.inner.store.lock().await?;
         let events = store.get_room_events(&room_key_info.room_id).await?;
 
-        Ok(events.into_iter().filter_map(only_utd_events).collect())
+        Ok(events.into_iter().filter_map(filter_non_utds).collect())
     }
 
+    #[instrument(skip_all, fields(room_id))]
     async fn on_resolved_utds(
         &self,
         room_id: &RoomId,
@@ -93,14 +98,28 @@ impl RedecryptorCtx for EventCache {
         let (room_cache, _) = self.for_room(room_id).await?;
         let mut state = room_cache.inner.state.write().await;
 
+        let event_ids: Vec<_> = events.iter().map(|(event_id, _)| event_id).collect();
+
+        trace!(?event_ids, "Replacing successfully re-decrypted events");
+
         for (event_id, decrypted) in events {
             // The event isn't in the cache, nothing to replace. Realistically this can't
-            // happen since we retrieved the list of events from the cache itself.
+            // happen since we retrieved the list of events from the cache itself and
+            // `find_event()` will look into the store as well.
             if let Some((location, mut target_event)) = state.find_event(&event_id).await? {
                 target_event.kind = TimelineEventKind::Decrypted(decrypted);
                 state.replace_event_at(location, target_event).await?
             }
         }
+
+        // We replaced a bunch of events, reactive updates for those replacements have
+        // been queued up. We need to send them out to our subscribers now.
+        let diffs = state.room_linked_chunk_mut().updates_as_vector_diffs();
+
+        let _ = room_cache.inner.sender.send(RoomEventCacheUpdate::UpdateTimelineEvents {
+            diffs,
+            origin: EventsOrigin::Cache,
+        });
 
         Ok(())
     }
@@ -168,7 +187,7 @@ impl Redecryptor {
         match machine.decrypt_room_event(event, room_id, client.decryption_settings()).await {
             Ok(decrypted) => Some(decrypted),
             Err(e) => {
-                warn!("Failed to redecrypt an event {e:?}");
+                warn!("Failed to redecrypt an event despite receiving a room key for it {e:?}");
                 None
             }
         }
--- a/crates/matrix-sdk/src/event_cache/redecryptor.rs
+++ b/crates/matrix-sdk/src/event_cache/redecryptor.rs
@@ -106,7 +106,47 @@ impl EventCache {
 
         Ok(())
     }
-}
+
+    async fn decrypt_event(
+        &self,
+        room_id: &RoomId,
+        event: &Raw<EncryptedEvent>,
+    ) -> Option<DecryptedRoomEvent> {
+        let client = self.inner.client().ok()?;
+        let machine = client.olm_machine().await;
+        let machine = machine.as_ref()?;
+
+        match machine.decrypt_room_event(event, room_id, client.decryption_settings()).await {
+            Ok(decrypted) => Some(decrypted),
+            Err(e) => {
+                warn!("Failed to redecrypt an event despite receiving a room key for it {e:?}");
+                None
+            }
+        }
+    }
+    /// Attempt to redecrypt events after a room key with the given session ID
+    /// has been received.
+    #[instrument(skip_all, fields(room_key_info))]
+    async fn retry_decryption(&self, room_key_info: RoomKeyInfo) -> Result<(), EventCacheError> {
+        trace!("Retrying to decrypt");
+
+        let events = self.get_utds(&room_key_info).await?;
+        let mut decrypted_events = Vec::with_capacity(events.len());
+
+        for (event_id, event) in events {
+            // If we managed to decrypt the event, and we should have to since we received
+            // the room key for this specific event, then replace the event.
+            if let Some(decrypted) =
+                self.decrypt_event(&room_key_info.room_id, event.cast_ref_unchecked()).await
+            {
+                decrypted_events.push((event_id, decrypted));
+            }
+        }
+
+        self.on_resolved_utds(&room_key_info.room_id, decrypted_events).await?;
+
+        Ok(())
+    }
 
 pub(crate) struct Redecryptor {
     cache: Weak<EventCacheInner>,
@@ -129,52 +169,6 @@ impl Redecryptor {
 
         task
     }
-    /// Attempt to redecrypt events after a room key with the given session ID
-    /// has been received.
-    #[instrument(skip_all, fields(room_key_info))]
-    async fn retry_decryption(
-        &self,
-        cache: &EventCache,
-        room_key_info: RoomKeyInfo,
-    ) -> Result<(), EventCacheError> {
-        trace!("Retrying to decrypt");
-
-        let events = cache.get_utds(&room_key_info).await?;
-        let mut decrypted_events = Vec::with_capacity(events.len());
-
-        for (event_id, event) in events {
-            // If we managed to decrypt the event, and we should have to since we received
-            // the room key for this specific event, then replace the event.
-            if let Some(decrypted) =
-                self.decrypt_event(&cache, &room_key_info.room_id, event.cast_ref_unchecked()).await
-            {
-                decrypted_events.push((event_id, decrypted));
-            }
-        }
-
-        cache.on_resolved_utds(&room_key_info.room_id, decrypted_events).await?;
-
-        Ok(())
-    }
-
-    async fn decrypt_event(
-        &self,
-        cache: &EventCache,
-        room_id: &RoomId,
-        event: &Raw<EncryptedEvent>,
-    ) -> Option<DecryptedRoomEvent> {
-        let client = cache.inner.client().ok()?;
-        let machine = client.olm_machine().await;
-        let machine = machine.as_ref()?;
-
-        match machine.decrypt_room_event(event, room_id, client.decryption_settings()).await {
-            Ok(decrypted) => Some(decrypted),
-            Err(e) => {
-                warn!("Failed to redecrypt an event despite receiving a room key for it {e:?}");
-                None
-            }
-        }
-    }
 
     async fn listen_for_room_keys_task(
         self,
@@ -193,8 +187,8 @@ impl Redecryptor {
                 let cache = EventCache { inner: event_cache };
 
                 for key in room_keys {
-                    let _ = self
-                        .retry_decryption(&cache, key)
+                    let _ = cache
+                        .retry_decryption(key)
                         .await
                         .inspect_err(|e| warn!("Error redecrypting {e:?}"));
                 }
--- a/crates/matrix-sdk/src/event_cache/redecryptor.rs
+++ b/crates/matrix-sdk/src/event_cache/redecryptor.rs
@@ -206,3 +206,177 @@ impl Redecryptor {
         info!("Shutting down the event cache redecryptor");
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use std::time::Duration;
+
+    use assert_matches2::assert_matches;
+    use eyeball_im::VectorDiff;
+    use matrix_sdk_base::deserialized_responses::TimelineEventKind;
+    use matrix_sdk_test::{
+        JoinedRoomBuilder, StateTestEvent, async_test, event_factory::EventFactory,
+    };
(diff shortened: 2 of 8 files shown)


crates/matrix-sdk/src/event_cache/room/mod.rs
```diff
@@ -1301,6 +1301,12 @@ mod private {
             &self.room_linked_chunk
         }
 
+        /// Returns a mutable reference to the underlying room linked chunk.
+        #[cfg(feature = "e2e-encryption")]
+        pub(in crate::event_cache) fn room_linked_chunk_mut(&mut self) -> &mut EventLinkedChunk {
```
Hywan: I think it breaks the purpose of `RoomEventCacheState`, not a big deal for the moment, we can still refactor later. Can we reduce the visibility scope, maybe like `pub(in super::super)`?

crates/matrix-sdk/src/event_cache/redecryptor.rs
```diff
@@ -0,0 +1,208 @@
...
+        // Load the relevant events from the event cache store and attempt to redecrypt
+        // things.
+        let store = self.inner.store.lock().await?;
+        let events = store
+            .get_room_events(
+                &room_key_info.room_id,
+                Some("m.room.encrypted"),
+                Some(&room_key_info.session_id),
+            )
+            .await?;
```
Hywan: Can we scope the lock at these 2 instructions please? If there is a large number of events, I prefer to iteration + filter-map + collect to happen with the lock being freed.

```suggestion
        let events = {
            let store = self.inner.store.lock().await;
            store
                .get_room_events(
                    &room_key_info.room_id,
                    Some("m.room.encrypted"),
                    Some(&room_key_info.session_id),
                )
                .await?
        };
```

crates/matrix-sdk/src/event_cache/redecryptor.rs
```diff
@@ -0,0 +1,208 @@
...
+use ruma::{OwnedEventId, RoomId, events::AnySyncTimelineEvent, serde::Raw};
+use tokio_stream::wrappers::errors::BroadcastStreamRecvError;
+use tracing::{info, instrument, trace, warn};
+
+use crate::{
+    Client,
+    event_cache::{
+        EventCache, EventCacheError, EventCacheInner, EventsOrigin, RoomEventCacheUpdate,
+    },
+};
```
Hywan: I tend to refer having relative paths for the current submodule I'm working in. Just a personal taste. I find it facilitates code moves.

```suggestion
use crate::Client;
use super::{EventCache, EventCacheError, EventCacheInner, EventsOrigin, RoomEventCacheUpdate};
```

crates/matrix-sdk/src/event_cache/redecryptor.rs
```diff
@@ -0,0 +1,208 @@
...
+        Ok(events.into_iter().filter_map(map_timeline_event).collect())
+    }
+
+    #[instrument(skip_all, fields(room_id))]
+    async fn on_resolved_utds(
+        &self,
+        room_id: &RoomId,
+        events: Vec<(OwnedEventId, DecryptedRoomEvent)>,
+    ) -> Result<(), EventCacheError> {
+        // Get the cache for this particular room and lock the state for the duration of
```
Hywan, line 280: Can we add a `timer!` log, just in case… 😛?

crates/matrix-sdk/src/event_cache/redecryptor.rs
```diff
@@ -0,0 +1,208 @@
...
+        // the decryption.
+        let (room_cache, _drop_handles) = self.for_room(room_id).await?;
+        let mut state = room_cache.inner.state.write().await;
+
+        let event_ids: Vec<_> = events.iter().map(|(event_id, _)| event_id).collect();
```
Hywan: Can we compute `event_ids` if and only if `trace!` is enabled? Maybe with [`tracing::enabled!`](https://docs.rs/tracing/latest/tracing/macro.enabled.html)?

Alternatively, I confess it's a bit more logs but, I would put the `trace!` inside the loop, closer to `replace_event_at`, because it's where we are really replacing the event. Thoughts?

crates/matrix-sdk/src/event_cache/redecryptor.rs
```diff
@@ -0,0 +1,208 @@
...
+
+                // TODO: `replace_event_at()` propagates changes to the store for every event,
+                // we should probably have a bulk version of this?
+                state.replace_event_at(location, target_event).await?
+            }
+        }
+
+        // We replaced a bunch of events, reactive updates for those replacements have
+        // been queued up. We need to send them out to our subscribers now.
+        let diffs = state.room_linked_chunk_mut().updates_as_vector_diffs();
```
Hywan, line 311: Note that this is safe only because we have a write-lock over the `RoomEventCache` at the beginning. Otherwise, the risk would have been to read updates from another async operation. All good then.

crates/matrix-sdk/src/event_cache/redecryptor.rs
```diff
@@ -0,0 +1,208 @@
...
+        trace!("Retrying to decrypt");
+
+        let events = self.get_utds(&room_key_info).await?;
+        let mut decrypted_events = Vec::with_capacity(events.len());
+
+        for (event_id, event) in events {
+            // If we managed to decrypt the event, and we should have to since we received
+            // the room key for this specific event, then replace the event.
+            if let Some(decrypted) =
+                self.decrypt_event(&room_key_info.room_id, event.cast_ref_unchecked()).await
```
Hywan: I wonder if we can optimise this part. Inside `decrypt_event`, we fetch the client, we fetch the `OlmMachine`, we fetch the decryption settings. And we will do that for each event in `events`.

Can we fetch these data outside `decrypt_event` and pass them to `decrypt_event`?

crates/matrix-sdk/src/event_cache/redecryptor.rs
```diff
@@ -0,0 +1,208 @@
...
+            }
+        }
+
+        self.on_resolved_utds(&room_key_info.room_id, decrypted_events).await?;
+
+        Ok(())
+    }
+}
+
+pub(crate) struct Redecryptor {
```
Hywan: Can you add documentation please?

crates/matrix-sdk/src/event_cache/redecryptor.rs
```diff
@@ -0,0 +1,208 @@
...
+    task: JoinHandle<()>,
+}
+
+impl Drop for Redecryptor {
+    fn drop(&mut self) {
+        self.task.abort();
+    }
+}
```
Hywan: Can't we use `matrix_sdk::executor::AbortOnDrop` here? It avoids the necessity to repeat this pattern everywhere, and at least the type system clarifies what we do with this task.

You can even use `matrix_sdk::executor::JoinHandlExt` to call `abort_on_drop()` directly on the `JoinHandle`: it will create the `AbortOnDrop` for you.

crates/matrix-sdk/src/event_cache/redecryptor.rs
```diff
@@ -0,0 +1,208 @@
...
+    }
+
+    async fn listen_for_room_keys_task(
+        cache: Weak<EventCacheInner>,
+        received_stream: impl Stream<Item = Result<Vec<RoomKeyInfo>, BroadcastStreamRecvError>>,
+    ) {
+        pin_mut!(received_stream);
+
+        // TODO: We need to relisten to this stream if it dies due to the cross-process
+        // lock reloading the Olm machine.
```
Hywan: Good point! Is it addressed in this PR? Just a note.

crates/matrix-sdk/src/event_cache/redecryptor.rs
```diff
@@ -206,3 +206,178 @@ impl Redecryptor {
...
+            })
+            .await;
+
+        // Bob should receive a new update from the cache.
+        assert_let_timeout!(
+            Duration::from_secs(1),
+            Ok(RoomEventCacheUpdate::UpdateTimelineEvents { diffs, .. }) = subscriber.recv()
+        );
+
+        // It should replace the UTD with a decrypted event.
```
Hywan, line 984: 🎉

crates/matrix-sdk/src/event_cache/room/mod.rs
```diff
@@ -1597,7 +1597,7 @@ mod private {
         /// observers that a single item has been replaced. Otherwise,
         /// such a notification is not emitted, because observers are
         /// unlikely to observe the store updates directly.
-        async fn replace_event_at(
+        pub(crate) async fn replace_event_at(
```
Hywan, line 1600: It must be visible to `crate::event_cache`, not `crate`.

crates/matrix-sdk/src/event_cache/redecryptor.rs
```diff
@@ -206,3 +206,178 @@ impl Redecryptor {
...
+    use ruma::{device_id, event_id, room_id, user_id};
+    use serde_json::json;
+
+    use crate::{
+        assert_let_timeout, encryption::EncryptionSettings, event_cache::RoomEventCacheUpdate,
+        test_utils::mocks::MatrixMockServer,
+    };
+
+    #[async_test]
+    async fn test_redecryptor() {
```
Hywan: At this step of the review, we've meet our new `Redecryptor` friend, but I didn't see where it was plugged into the `EventCache`. Am I missing something?

crates/matrix-sdk/src/event_cache/redecryptor.rs
```diff
@@ -161,46 +167,98 @@ impl Drop for Redecryptor {
...
+        cache: &Weak<EventCacheInner>,
+    ) -> Option<impl Stream<Item = Result<Vec<RoomKeyInfo>, BroadcastStreamRecvError>>> {
+        let event_cache = cache.upgrade()?;
+        let client = event_cache.client().ok()?;
+        let machine = client.olm_machine().await;
+
+        machine.as_ref().map(|m| m.store().room_keys_received_stream())
+    }
+
+    fn upgrade_event_cache(cache: &Weak<EventCacheInner>) -> Option<EventCache> {
```
Hywan, line 633: In case `rustc` is drunk: can we add `#[inline(always)]` here please?

crates/matrix-sdk/src/event_cache/redecryptor.rs
```diff
@@ -336,6 +394,14 @@ mod tests {
         let event = event_receiver.await.expect("Alice should have sent the event by now");
         let room_key = room_key.await;
 
+        // We regenerate the Olm machine to check if the room key stream is recreated to
+        // correctly.
```
Hywan, line 939: 👍

crates/matrix-sdk/src/event_cache/redecryptor.rs
```diff
@@ -12,14 +12,63 @@
...
+//! Another gotcha is that room keys might be received on another process if the
+//! Client is operating on a iOS device. A separate process is used in this case
+//! to receive push notifications. In this case the room key will be received
+//! and Rd won't get notified about it. To work around this decryption requests
+//! can be explicitly sent to Rd.
+//!
+//!
+//!                              ┌─────────────┐
+//!                              │             │
+//!                  ┌───────────┤  Timeline   │◄────────────┐
```
Hywan: What do you call “Timeline” here? Is it `matrix_sdk_ui::Timeline`? Is it a timeline update, like from `/sync`, or `/messages` and so on?

crates/matrix-sdk/src/event_cache/mod.rs
```diff
@@ -149,6 +152,10 @@ pub struct EventCacheDropHandles {
 
     /// The task used to automatically shrink the linked chunks.
     auto_shrink_linked_chunk_task: JoinHandle<()>,
+
+    /// The task used to automatically redecrypt UTDs.
+    #[cfg(feature = "e2e-encryption")]
+    _redecryptor: redecryptor::Redecryptor,
```
Hywan, line 158: 👍

crates/matrix-sdk/src/event_cache/redecryptor.rs
```diff
@@ -230,10 +288,28 @@ impl EventCache {
...
+        let _ =
+            self.inner.redecryption_channels.decryption_request_sender.send(request).inspect_err(
+                |_| warn!("Requesting a decryption while the redecryption task has been shut down"),
+            );
+    }
+
+    /// Subscribe to reports that the redecryptor generates.
+    ///
+    /// TODO: Explain when the redecryptor might send such reports.
+    pub fn subscrube_to_decryption_reports(
```
Hywan:
```suggestion
    pub fn subscribe_to_decryption_reports(
```

crates/matrix-sdk/src/event_cache/redecryptor.rs
```diff
@@ -248,23 +324,17 @@ impl Redecryptor {
...
 
-        Self { task, request_decryption_sender }
-    }
-
-    #[allow(dead_code)]
-    pub(super) fn request_decryption(&self, request: DecryptionRetryRequest) {
-        let _ = self.request_decryption_sender.send(request).inspect_err(|_| {
-            warn!("Requesting a decryption while the redecryption task has been shut down")
-        });
+        Self { task }
```
Hywan: Nice, much simpler.

crates/matrix-sdk/src/event_cache/redecryptor.rs
```diff
@@ -326,9 +401,21 @@ impl Redecryptor {
...
                             }
+
+                            // TODO: Deal with encryption info updating as well.
                         },
                         Some(Err(_)) => {
-                            todo!("Handle lagging here, how?")
+                            // We missed some room keys, we need to report this in case a listener
+                            // has and idea which UTDs we should attempt to redecrypt.
+                            //
+                            // This would most likely be the timeline. The timeline might attempt
```
Hywan: Are you talking about `matrix_sdk_ui::Timeline`? If yes, can you precise it please?

crates/matrix-sdk/src/event_cache/redecryptor.rs
```diff
@@ -167,13 +176,31 @@ impl EventCache {
...
+            let event_id = event.event_id();
+
+            let event = as_variant!(event.kind, TimelineEventKind::Decrypted(event) => event);
+            // Zip the event ID and event together so we don't have to pick out the event ID
+            // again. We need the event ID to replace the event in the cache.
+            event_id.zip(event)
+        };
+
+        let store = self.inner.store.lock().await?;
+        let events = store.get_room_events(room_id, None, Some(session_id)).await?;
```
Hywan: Same as before: can we take the lock for a smaller region:

```rust
let events = {
    let store = self.inner.store.lock().await?;
    store.get_room_events(…)
};
```

crates/matrix-sdk/src/event_cache/redecryptor.rs
```diff
@@ -218,24 +227,29 @@ impl EventCache {
     async fn on_resolved_utds(
         &self,
         room_id: &RoomId,
-        events: Vec<(OwnedEventId, DecryptedRoomEvent)>,
+        events: Vec<(OwnedEventId, DecryptedRoomEvent, Option<Vec<Action>>)>,
```
Hywan: I think it starts deserving its own type, or at least type alias?

crates/matrix-sdk/src/event_cache/redecryptor.rs
```diff
@@ -364,13 +368,21 @@ impl EventCache {
             }
         }
 
+        let event_ids: BTreeSet<_> =
```
Hywan, line 422: Maybe we can early-return if `get_utds` returns an empty vector?

crates/matrix-sdk/src/event_cache/redecryptor.rs
```diff
@@ -403,6 +415,13 @@ impl EventCache {
...
```
Hywan, line 476: Same.

crates/matrix-sdk/src/event_cache/redecryptor.rs
```diff
@@ -256,10 +256,15 @@ impl EventCache {
...
                 // we should probably have a bulk version of this?
-                state.replace_event_at(location, target_event).await?
+                state.replace_event_at(location, target_event.clone()).await?;
+                new_events.push(target_event);
             }
         }
 
+        // TODO: Is this useful? Do I need to call this once per event as well because
+        // replace event at works on a single event?
+        state.post_process_new_events(new_events, false).await?;
```
Hywan: This is the correct usage. Calling it with multiple events is better performance-wise.

crates/matrix-sdk/src/event_cache/room/mod.rs
```diff
@@ -1439,7 +1439,7 @@ mod private {
         /// linked chunk.
         ///
         /// Flushes updates to disk first.
-        async fn post_process_new_events(
+        pub(crate) async fn post_process_new_events(
```
Hywan: `pub(super)` maybe?

crates/matrix-sdk/src/event_cache/redecryptor.rs
```diff
@@ -31,28 +41,49 @@
...
+//!                              __:='|____`-:__
+//!                             ||[] ||====|| []||
+//!                             ||[] ||====|| []||
+//!                             |:== ||====|| ==:|
+//!                             ||[] ||====|| []||
+//!                             ||[] ||====|| []||
+//!                            _||_  ||====||  _||_
+//!                           (====) |:====:| (====)
+//!                            }--{  | |  | |  }--{
+//!                           (____) |_|  |_| (____)
```
Hywan, line 73: Absolutely necessary.

Hywan (approved):
Wow. That was an impressive set of patches.

How a simple problem gently piles more and more complexity. The resulting code is not complex however, and that's really great. Thank you for the great documentation, including inline documentation!

I've left a couple of feedback, but nothing requiring a huge change.

crates/matrix-sdk/src/event_cache/redecryptor.rs
Hywan: Marked as resolved, as it's done in subsequent patches.

crates/matrix-sdk/src/event_cache/redecryptor.rs
poljar: I would have thought that the module level documentation would be enough, the whole module is about this single struct.

I guess I can say as much in the struct docs? Please take a look at the module level docs. Would that suffice?

crates/matrix-sdk/src/event_cache/redecryptor.rs
poljar: It is, yes. Specifically in https://github.com/matrix-org/matrix-rust-sdk/pull/5746/commits/e93423504569bf280f4dd6b3405647056ff55832.

crates/matrix-sdk/src/event_cache/redecryptor.rs
poljar: Yes, will do.

crates/matrix-sdk/src/event_cache/room/mod.rs
poljar: We can, but isn't `super::super` the `crate::event_cache` module? I.e. isn't the visibility the same in those two cases?

Hywan: Ah, that's possible indeed. I also tend to prefer having relative path, in case we move the code to another place. Just a personal taste.

crates/matrix-sdk/src/event_cache/redecryptor.rs
Hywan: I wrote this comment before seeing the patch that adds the doc.

I think adding a note like the one you've suggested, saying “please take a look at the module level docs” with an intra-link is fine!

crates/matrix-sdk/src/event_cache/redecryptor.rs
Hywan: :+1:

crates/matrix-sdk/src/event_cache/redecryptor.rs
poljar: > Can we compute `event_ids` if and only if `trace!` is enabled? Maybe with [`tracing::enabled!`](https://docs.rs/tracing/latest/tracing/macro.enabled.html)?

Sadly we can't, as the event IDs are sent out as part of the `RedecryptorReport`, that's also the reason why we're cloning it in later parts of this patchset:

https://github.com/matrix-org/matrix-rust-sdk/pull/5746/commits/a2f89e85b967175f8283cefcdb7ea7a3f574c321#diff-5c3b83b7260e87e30c1b6b3bfad8b5be005c31d142b825e75ff0e047680c2923R201


> Alternatively, I confess it's a bit more logs but, I would put the `trace!` inside the loop, closer to `replace_event_at`, because it's where we are really replacing the event. Thoughts?

Hmm, I'm really worried about log spam here. If you're suggesting this because you want to avoid the building of the `BTreeSet` then I would say that this comment resolved itself since the `BTreeSet` is necessary for the `RedecryptorReport`.

crates/matrix-sdk/src/event_cache/room/mod.rs
poljar: Ah ok, I thought `super::super` would look silly and confusing, but I don't have strong opinions here.

I'll change it to `super::super`.

crates/matrix-sdk/src/event_cache/redecryptor.rs
poljar: We can, but notice that the normal case is to use the `room` to decrypt the event.

The room is fetched outside of `decrypt_event` for this every reason. 

It's abnormal to attempt to decrypt an event for a room we don't know about. So I thought that the fallback to the `OlmMachine` is fine to be done inline like that.

Does this change your opinion, or do you think we should still do it outside of `decrypt_event()`?

crates/matrix-sdk/src/event_cache/redecryptor.rs
poljar: This was already fixed in d7d4730b, I guess it's kinda bad that this made it into the original commit introducing this function.

I can move the fixing into the right place afterwards if that's wanted.

crates/matrix-sdk/src/event_cache/redecryptor.rs
poljar: Type alias sounds good.

crates/matrix-sdk/src/event_cache/room/mod.rs
poljar: Would need to be a `in super::super` but can do.

crates/matrix-sdk/src/event_cache/room/mod.rs
poljar: https://github.com/matrix-org/matrix-rust-sdk/pull/5746/commits/d2eab603c1d34a4f0d310183064a2f39071dfd2d

crates/matrix-sdk/src/event_cache/redecryptor.rs
poljar: https://github.com/matrix-org/matrix-rust-sdk/pull/5746/commits/84a21a42d0a54aa73708eb070f97dcb5ce839a9c

crates/matrix-sdk/src/event_cache/redecryptor.rs
poljar: https://github.com/matrix-org/matrix-rust-sdk/pull/5746/commits/3ad70623bbc5e26cdc50abed79af206a1445d335

crates/matrix-sdk/src/event_cache/redecryptor.rs
poljar: https://github.com/matrix-org/matrix-rust-sdk/pull/5746/commits/952c5af07c44f07d8b13a646d31c21b58c69b6c2

crates/matrix-sdk/src/event_cache/redecryptor.rs
poljar: https://github.com/matrix-org/matrix-rust-sdk/pull/5746/commits/717f016f21b867da5465a962c0ef334fe2f21df1

crates/matrix-sdk/src/event_cache/redecryptor.rs
poljar: https://github.com/matrix-org/matrix-rust-sdk/pull/5746/commits/f9c23b361256b84cabc336c65da24cc061d952b8

crates/matrix-sdk/src/event_cache/room/mod.rs
poljar: https://github.com/matrix-org/matrix-rust-sdk/pull/5746/commits/38df621b8a594ba348781f91ccd4453f3442df55

crates/matrix-sdk/src/event_cache/redecryptor.rs
poljar: https://github.com/matrix-org/matrix-rust-sdk/pull/5746/commits/2e9e9aedd77b4b3019277c30ab6161909ae33822

crates/matrix-sdk/src/event_cache/redecryptor.rs
poljar: https://github.com/matrix-org/matrix-rust-sdk/pull/5746/commits/e45994dbf93717a3c20bda78bcaba321b8952667

crates/matrix-sdk/src/event_cache/redecryptor.rs
poljar: https://github.com/matrix-org/matrix-rust-sdk/pull/5746/commits/a011c8ae44a1054e9bb2c6347f720ba30cbb9493

crates/matrix-sdk/src/event_cache/redecryptor.rs
poljar: Yup, fixed as well in https://github.com/matrix-org/matrix-rust-sdk/pull/5746/commits/84a21a42d0a54aa73708eb070f97dcb5ce839a9c.

crates/matrix-sdk/src/event_cache/redecryptor.rs
poljar: Added some aliases in https://github.com/matrix-org/matrix-rust-sdk/pull/5746/commits/f3dde1662431587bc0361d514e38c903b1764b88.

crates/matrix-sdk/src/event_cache/redecryptor.rs
poljar: Good thinking https://github.com/matrix-org/matrix-rust-sdk/pull/5746/commits/c3fb01d35d44d77485bcf0227019ddb4978d0887.

crates/matrix-sdk/src/event_cache/redecryptor.rs
poljar: I think this was about the early return as well, if so, fixed in https://github.com/matrix-org/matrix-rust-sdk/pull/5746/commits/c3fb01d35d44d77485bcf0227019ddb4978d0887 as well.

crates/matrix-sdk/src/event_cache/room/mod.rs
poljar: Did so in https://github.com/matrix-org/matrix-rust-sdk/pull/5746/commits/38df621b8a594ba348781f91ccd4453f3442df55.

poljar:
I'm going to merge this since you approved it. If you feel like there's more to do here please shout and I'll handle it in a separate PR.

richvdh:
> This is very much a WIP PR. Eventually it will get split into different smaller PRs.

@poljar: since this PR got merged, I guess the description became outdated?

poljar:
> > This is very much a WIP PR. Eventually it will get split into different smaller PRs.
> 
> @poljar: since this PR got merged, I guess the description became outdated?

Sure, updated the description.

PR #5748, opened 2025-10-04
Upgrade Ruma

zecakeh:
Brings a breaking change with event structs being non-exhaustive now, so they need to be constructed with methods rather than with a struct declaration.

A few other crates are also upgraded, to avoid having two versions of the same crate in the dependency tree if possible.


poljar (approved):
Looks good, thanks.

PR #5749, opened 2025-10-04
IndexedDB: add functionality for cleaning up `MediaStore`

mgoldenberg:
## Background

This pull request is part of a series of pull requests to add a full IndexedDB implementation of the `EventCacheStore` and  `MediaStore` (see #4617, #4996, #5090, #5138, #5226, #5274, #5343, #5384, #5406, #5414, #5497, #5506, #5540, #5574, #5603, #5676, #5682). This particular pull request adds IndexedDB-backed implementations for cleaning expired and over-sized media in IndexedDB via `MediaStore`.

### IndexedDB implementations of top-level `MediaStore` functions

The overarching change is the removal of the nested `MemoryStore` and the addition of IndexedDB-backed implementations for the following functions.

- `MediaStoreInner::clean_inner`
- `MediaStoreInner::last_media_cleanup_time_inner`
- `MediaStoreInner::set_ignore_media_retention_policy`

Additionally, the following types, trait implementations, and functions were added to support the functions above.

**Types**
- `MediaCleanupTime` - a basic wrapper around `UnixTime` for representing the last time the `MediaStore` was cleaned. This also includes indexed versions of this type optimized for IndexedDB storage.

**Trait Implementations**
- `Add<Duration>` and `Sub<Duration>` for `UnixTime`
    - Helps calculate expiration times, etc. for `Media`

**Functions**
- `Transaction::get_keys`
- `Transaction::fold_keys_while`
- `IndexeddbMediaStoreTransaction::get_media_cleanup_time`
- `IndexeddbMediaStoreTransaction::put_media_cleanup_time`
- `IndexeddbMediaStoreTransaction::get_media_keys_by_content_size`
- `IndexeddbMediaStoreTransaction::fold_media_keys_by_retention_metadata_while`
- `IndexeddbMediaStoreTransaction::get_cache_size`
- `IndexeddbMediaStoreTransaction::put_media`
- `IndexeddbMediaStoreTransaction::delete_media_by_content_size`
- `IndexeddbMediaStoreTransaction::delete_media_by_content_size_greater_than`
- `IndexeddbMediaStoreTransaction::delete_media_by_last_access`
- `IndexeddbMediaStoreTransaction::delete_media_by_last_access_earlier_than`
- `IndexeddbMediaStoreTransaction::delete_media_by_retention_metadata`
- `IndexeddbMediaStoreTransaction::delete_media_by_retention_metadata_to`

## Bug Fixes

1. [`3a379e061a5d2266ae2be3d5e897d089d0e83df7`](https://github.com/matrix-org/matrix-rust-sdk/commit/3a379e061a5d2266ae2be3d5e897d089d0e83df7) - Integration tests for `MediaStore` that assure media size constraints are properly implemented expect that content will maintain its size after encoding. For this reason, these tests are recommended to be run against an unencrypted `MediaStore` only (see [documentation](https://github.com/matrix-org/matrix-rust-sdk/blob/68075b65fbbea5b34ea50ba27f68aea7d7129282/crates/matrix-sdk-base/src/media/store/integration_tests.rs#L969-L972)). Previously, unencrypted media contents were Base64 encoded, which increased the size of the contents, causing the tests to fail. Now, an unencrypted `MediaStore` does not modify the contents for serialization, but rather stores it as it is provided.
2. [`377271814a0284183be2c4cdb49c7ef31f189760`](https://github.com/matrix-org/matrix-rust-sdk/commit/377271814a0284183be2c4cdb49c7ef31f189760) - `MediaStoreInner::set_media_retention_policy_inner` previously failed to commit the transaction used to set the `MediaRetentionPolicy` and so the policy was not properly recorded. The transaction is now being commited.
3. [`870e1e26718c75e9bb71206fd4085a98557afd14`](https://github.com/matrix-org/matrix-rust-sdk/commit/870e1e26718c75e9bb71206fd4085a98557afd14) - `IndexeddbMediaStoreTransaction::put_media_if_policy_compliant` previously failed to put `Media` items in IndexedDB when the given `Media` was set to ignore the `MediaRetentionPolicy`. The condition for adding `Media` through this function was modified to correct the error.

## Caveats

IndexedDB does not seem to allow partial updates to existing objects. So, if one would like to update a single field on an object in IndexedDB, they must do the following.

1. Deserialize the object from the database.
2. Make the desired modifications to the deserialized object. 
3. Re-serialize the modified object.
4. Overwrite the old object with the modified object.

This can, of course, be very inefficient when an object store's schema contains large fields. 
And, unfortunately, this issue affects the existing schema for the `Media` object store.

The issue is that the `Media` object store stores metadata and content alongside each other in a single object. So, for instance, to set the `last_access` field of an object in IndexedDB, one has to deserialize and re-serialized the entire content of the `Media`, which might be very large.

The following functions are affected by this issue.

- [`MediaStore::replace_media_key`](https://github.com/mgoldenberg/matrix-rust-sdk/blob/a82a0db02350438cd78be2e73c18e437d4a47229/crates/matrix-sdk-indexeddb/src/media_store/mod.rs#L145)
- [`MediaStore::get_media_content`](https://github.com/mgoldenberg/matrix-rust-sdk/blob/a82a0db02350438cd78be2e73c18e437d4a47229/crates/matrix-sdk-indexeddb/src/media_store/mod.rs#L164)
    - Due to the use of [`IndexeddbMediaStoreTransaction::access_media_by_id`](https://github.com/mgoldenberg/matrix-rust-sdk/blob/68075b65fbbea5b34ea50ba27f68aea7d7129282/crates/matrix-sdk-indexeddb/src/media_store/transaction.rs#L97) which sets the `last_access` field of a `Media` object
- [`MediaStore::get_media_content_for_uri`](https://github.com/mgoldenberg/matrix-rust-sdk/blob/a82a0db02350438cd78be2e73c18e437d4a47229/crates/matrix-sdk-indexeddb/src/media_store/mod.rs#L185)
    - Due to the use of [`IndexeddbMediaStoreTransaction::access_media_by_id`](https://github.com/mgoldenberg/matrix-rust-sdk/blob/68075b65fbbea5b34ea50ba27f68aea7d7129282/crates/matrix-sdk-indexeddb/src/media_store/transaction.rs#L97) which sets the `last_access` field of a `Media` object
- [`MediaStore::set_ignore_media_retention_policy`](https://github.com/mgoldenberg/matrix-rust-sdk/blob/a82a0db02350438cd78be2e73c18e437d4a47229/crates/matrix-sdk-indexeddb/src/media_store/mod.rs#L221)

### Proposed Solution

I am currently working on a solution to this problem which separates the metadata and content of a `Media` object into two separate object stores. As a result, this would require that retrieving media contents would require two separate queries to IndexedDB, but I think this is preferable to the excessive (de)serialization described above.

## Future Work

- Separate metadata and contents in `Media` object store
- Refactor feature flags
- Expose `EventCacheStore` and `MediaStore` outside of the `matrix-sdk-indexeddb` 

---

### DIFF
--- a/crates/matrix-sdk-indexeddb/src/media_store/mod.rs
+++ b/crates/matrix-sdk-indexeddb/src/media_store/mod.rs
@@ -434,8 +434,8 @@ impl MediaStoreInner for IndexeddbMediaStore {
 #[cfg(all(test, target_family = "wasm"))]
 mod tests {
     use matrix_sdk_base::{
-        media::store::MediaStoreError, media_store_integration_tests,
-        media_store_integration_tests_time,
+        media::store::MediaStoreError, media_store_inner_integration_tests,
+        media_store_integration_tests, media_store_integration_tests_time,
     };
     use uuid::Uuid;
 
@@ -456,6 +456,9 @@ mod tests {
 
         #[cfg(target_family = "wasm")]
         media_store_integration_tests_time!();
+
+        #[cfg(target_family = "wasm")]
+        media_store_inner_integration_tests!(with_media_size_tests);
     }
 
     mod encrypted {
@@ -473,5 +476,8 @@ mod tests {
 
         #[cfg(target_family = "wasm")]
         media_store_integration_tests_time!();
+
+        #[cfg(target_family = "wasm")]
+        media_store_inner_integration_tests!();
     }
 }
--- a/crates/matrix-sdk-indexeddb/src/media_store/mod.rs
+++ b/crates/matrix-sdk-indexeddb/src/media_store/mod.rs
@@ -33,7 +33,7 @@ use matrix_sdk_base::{
     media::{
         store::{
             IgnoreMediaRetentionPolicy, MediaRetentionPolicy, MediaService, MediaStore,
-            MediaStoreInner, MemoryMediaStore,
+            MediaStoreInner,
         },
         MediaRequestParameters,
     },
@@ -66,12 +66,6 @@ pub struct IndexeddbMediaStore {
     // A service for conveniently delegating media-related queries to an `MediaStoreInner`
     // implementation
     media_service: MediaService,
-    // An in-memory store for providing temporary implementations for
-    // functions of `MediaStore`.
-    //
-    // NOTE: This will be removed once we have IndexedDB-backed implementations for all
-    // functions in `MediaStore`.
-    memory_store: MemoryMediaStore,
 }
 
 impl IndexeddbMediaStore {
--- a/crates/matrix-sdk-indexeddb/src/media_store/mod.rs
+++ b/crates/matrix-sdk-indexeddb/src/media_store/mod.rs
@@ -305,10 +305,17 @@ impl MediaStoreInner for IndexeddbMediaStore {
         ignore_policy: IgnoreMediaRetentionPolicy,
     ) -> Result<(), IndexeddbMediaStoreError> {
         let _timer = timer!("method");
-        self.memory_store
-            .set_ignore_media_retention_policy_inner(request, ignore_policy)
-            .await
-            .map_err(IndexeddbMediaStoreError::MemoryStore)
+
+        let transaction =
+            self.transaction(&[Media::OBJECT_STORE], IdbTransactionMode::Readwrite)?;
+        if let Some(mut media) = transaction.get_media_by_id(request).await? {
+            if media.metadata.ignore_policy != ignore_policy {
+                media.metadata.ignore_policy = ignore_policy;
+                transaction.put_media(&media).await?;
+                transaction.commit().await?;
+            }
+        }
+        Ok(())
     }
 
     #[instrument(skip_all)]
--- a/crates/matrix-sdk-indexeddb/src/media_store/mod.rs
+++ b/crates/matrix-sdk-indexeddb/src/media_store/mod.rs
@@ -28,7 +28,9 @@ use std::{rc::Rc, time::Duration};
 
 pub use builder::IndexeddbMediaStoreBuilder;
 pub use error::IndexeddbMediaStoreError;
-use indexed_db_futures::{cursor::CursorDirection, database::Database, Build};
+use indexed_db_futures::{
+    cursor::CursorDirection, database::Database, transaction::TransactionMode, Build,
+};
 use matrix_sdk_base::{
     media::{
         store::{
@@ -41,7 +43,6 @@ use matrix_sdk_base::{
 };
 use ruma::{time::SystemTime, MilliSecondsSinceUnixEpoch, MxcUri};
 use tracing::instrument;
-use web_sys::IdbTransactionMode;
 
 use crate::{
     media_store::{
@@ -81,7 +82,7 @@ impl IndexeddbMediaStore {
     pub fn transaction<'a>(
         &'a self,
         stores: &[&str],
-        mode: IdbTransactionMode,
+        mode: TransactionMode,
     ) -> Result<IndexeddbMediaStoreTransaction<'a>, IndexeddbMediaStoreError> {
         Ok(IndexeddbMediaStoreTransaction::new(
             self.inner
@@ -110,8 +111,7 @@ impl MediaStore for IndexeddbMediaStore {
 
         let now = Duration::from_millis(MilliSecondsSinceUnixEpoch::now().get().into());
 
-        let transaction =
-            self.transaction(&[Lease::OBJECT_STORE], IdbTransactionMode::Readwrite)?;
+        let transaction = self.transaction(&[Lease::OBJECT_STORE], TransactionMode::Readwrite)?;
 
         if let Some(lease) = transaction.get_lease_by_id(key).await? {
             if lease.holder != holder && !lease.has_expired(now) {
@@ -149,8 +149,7 @@ impl MediaStore for IndexeddbMediaStore {
     ) -> Result<(), IndexeddbMediaStoreError> {
         let _timer = timer!("method");
 
-        let transaction =
-            self.transaction(&[Media::OBJECT_STORE], IdbTransactionMode::Readwrite)?;
+        let transaction = self.transaction(&[Media::OBJECT_STORE], TransactionMode::Readwrite)?;
         if let Some(mut media) = transaction.get_media_by_id(from).await? {
             // delete before adding, in case `from` and `to` generate the same key
             transaction.delete_media_by_id(from).await?;
@@ -177,8 +176,7 @@ impl MediaStore for IndexeddbMediaStore {
     ) -> Result<(), IndexeddbMediaStoreError> {
         let _timer = timer!("method");
 
-        let transaction =
-            self.transaction(&[Media::OBJECT_STORE], IdbTransactionMode::Readwrite)?;
+        let transaction = self.transaction(&[Media::OBJECT_STORE], TransactionMode::Readwrite)?;
         transaction.delete_media_by_id(request).await?;
         transaction.commit().await.map_err(Into::into)
     }
@@ -199,8 +197,7 @@ impl MediaStore for IndexeddbMediaStore {
     ) -> Result<(), IndexeddbMediaStoreError> {
         let _timer = timer!("method");
 
-        let transaction =
-            self.transaction(&[Media::OBJECT_STORE], IdbTransactionMode::Readwrite)?;
+        let transaction = self.transaction(&[Media::OBJECT_STORE], TransactionMode::Readwrite)?;
         transaction.delete_media_by_uri(uri).await?;
         transaction.commit().await.map_err(Into::into)
     }
@@ -247,7 +244,7 @@ impl MediaStoreInner for IndexeddbMediaStore {
         &self,
     ) -> Result<Option<MediaRetentionPolicy>, IndexeddbMediaStoreError> {
         let _timer = timer!("method");
-        self.transaction(&[MediaRetentionPolicy::OBJECT_STORE], IdbTransactionMode::Readonly)?
+        self.transaction(&[MediaRetentionPolicy::OBJECT_STORE], TransactionMode::Readonly)?
             .get_media_retention_policy()
             .await
             .map_err(Into::into)
@@ -261,7 +258,7 @@ impl MediaStoreInner for IndexeddbMediaStore {
         let _timer = timer!("method");
 
         let transaction =
-            self.transaction(&[MediaRetentionPolicy::OBJECT_STORE], IdbTransactionMode::Readwrite)?;
+            self.transaction(&[MediaRetentionPolicy::OBJECT_STORE], TransactionMode::Readwrite)?;
         transaction.put_item(&policy).await?;
         transaction.commit().await.map_err(Into::into)
     }
@@ -277,8 +274,7 @@ impl MediaStoreInner for IndexeddbMediaStore {
     ) -> Result<(), IndexeddbMediaStoreError> {
         let _timer = timer!("method");
 
-        let transaction =
-            self.transaction(&[Media::OBJECT_STORE], IdbTransactionMode::Readwrite)?;
+        let transaction = self.transaction(&[Media::OBJECT_STORE], TransactionMode::Readwrite)?;
 
         let media = Media {
             metadata: MediaMetadata {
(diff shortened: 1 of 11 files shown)


mgoldenberg:
(Test failure is due to an intermittent timing issue in how `media_store_integration_tests_time::test_leased_locks` is implemented).

crates/matrix-sdk-indexeddb/src/media_store/serializer/indexed_types.rs
```diff
@@ -67,6 +67,10 @@ pub type IndexedLeaseContent = MaybeEncrypted;
 /// A (possibly) encrypted representation of a [`MediaRetentionPolicy`]
 pub type IndexedMediaRetentionPolicyContent = MaybeEncrypted;
 
+/// A (possibly) encrypted representation of a the last time the store was
```
bnjbvr: nit: `of a the last time`

crates/matrix-sdk-indexeddb/src/transaction/mod.rs
```diff
@@ -264,6 +265,85 @@ impl<'a> Transaction<'a> {
...
+            .serde()?
+            .await?
+        {
+            while let Some(key) = cursor.next_key_ser().await? {
+                match f(&state, &key) {
+                    Some(s) => state = s,
+                    None => return Ok((state, Some(key))),
+                }
+            }
+        }
```
bnjbvr, line 343: Are the two `Some(mut cursor)` of the same type? If so, it might be nice to first bind the `cursor` to a temporary variable (with an early `else { return Ok((init, None));` branch, if both branches returned None). And then, put the common `while let Some(key) = … { }` thereafter.

(Of course that only applies if the cursor types are the same…)

crates/matrix-sdk-indexeddb/src/media_store/transaction.rs
```diff
@@ -153,6 +157,51 @@ impl<'a> IndexeddbMediaStoreTransaction<'a> {
...
+    /// accumulator while the given function `f` returns [`Some`].
+    ///
+    /// This function returns the final value of the accumulator and the key, if
+    /// any, which caused the fold to short circuit.
+    ///
+    /// Note that the use of cursor means that keys are read lazily from
+    /// IndexedDB.
+    ///
+    /// [1]: IndexedMediaRetentionMetadataKey
+    pub async fn fold_media_keys_by_retention_metadata_while<B, F>(
```
bnjbvr: ditto for the `B` generic type name here

crates/matrix-sdk-indexeddb/src/transaction/mod.rs
```diff
@@ -264,6 +265,85 @@ impl<'a> Transaction<'a> {
...
+    /// This function returns the final value of the accumulator and the key, if
+    /// any, which caused the fold to short circuit.
+    ///
+    /// Note that the use of cursor means that keys are read lazily from
+    /// IndexedDB.
+    pub async fn fold_keys_while<T, K, B, F>(
+        &self,
+        direction: IdbCursorDirection,
+        range: impl Into<IndexedKeyRange<K>>,
+        init: B,
```
bnjbvr: I *think* `B` means `base` here, but could we rename the type name variable to `Acc`, instead, since this is the accumulator type, that's mentioned in the code comment?

crates/matrix-sdk-indexeddb/src/media_store/serializer/indexed_types.rs
```diff
@@ -437,6 +437,20 @@ pub struct IndexedMediaLastAccessKey(
     #[serde(with = "unix_time")] UnixTime,
 );
 
+impl IndexedMediaLastAccessKey {
+    /// Returns whether the associated [`IndexedMedia`] record should ignore the
+    /// global [`MediaRetentionPolicy`]
+    pub fn ignore_policy(&self) -> IgnoreMediaRetentionPolicy {
```
bnjbvr: Doc comment here looks wrong? If it was indeed returning a "should ignore", that should be a bool return type, likely.

crates/matrix-sdk-indexeddb/src/media_store/serializer/indexed_types.rs
```diff
@@ -493,6 +507,25 @@ pub struct IndexedMediaRetentionMetadataKey(
     IndexedMediaContentSize,
 );
 
+impl IndexedMediaRetentionMetadataKey {
+    /// Returns whether the associated [`IndexedMedia`] record should ignore the
+    /// global [`MediaRetentionPolicy`]
+    pub fn ignore_policy(&self) -> IgnoreMediaRetentionPolicy {
+        self.0
+    }
```
bnjbvr: and here

crates/matrix-sdk-indexeddb/src/media_store/serializer/indexed_types.rs
```diff
@@ -374,8 +374,8 @@ pub struct IndexedMediaContentSizeKey(
 impl IndexedMediaContentSizeKey {
     /// Returns whether the associated [`IndexedMedia`] record should ignore the
     /// global [`MediaRetentionPolicy`]
-    pub fn ignore_policy(&self) -> bool {
-        self.0.is_yes()
+    pub fn ignore_policy(&self) -> IgnoreMediaRetentionPolicy {
+        self.0
```
bnjbvr, line 386: Doc comment now needs an update!

crates/matrix-sdk-indexeddb/src/media_store/mod.rs
```diff
@@ -348,10 +348,66 @@ impl MediaStoreInner for IndexeddbMediaStore {
...
+                .await?
+                .ok_or(Self::Error::CacheSizeTooBig)?;
+            if cache_size > (max_cache_size as usize) {
+                let (_, upper_key) = transaction
+                    .fold_media_keys_by_retention_metadata_while(
+                        CursorDirection::Prev,
+                        ignore_policy,
+                        0usize,
+                        |total, key| {
+                            web_sys::console::log_1(&format!("total={total}, key={key:?}").into());
```
bnjbvr: Probably a temporary log that was useful for development, but might want to delete?

crates/matrix-sdk-indexeddb/src/media_store/transaction.rs
```diff
@@ -242,8 +242,9 @@ impl<'a> IndexeddbMediaStoreTransaction<'a> {
         policy: MediaRetentionPolicy,
     ) -> Result<(), TransactionError> {
         self.put_item_if(media, |indexed| {
-            !indexed.content_size.ignore_policy().is_yes()
-                && !policy.exceeds_max_file_size(indexed.content_size.content_size() as u64)
+            indexed.content_size.ignore_policy().is_yes()
+                || (!indexed.content_size.ignore_policy().is_yes()
+                    && !policy.exceeds_max_file_size(indexed.content_size.content_size() as u64))
```
bnjbvr: This conceptually does: `A || (!A && B)`. You can simplify it to `A || B`, because the second branch will only be taken if the left hand side is false.

So, comparing to the initial code: only need to remove the leading `!` line 245.

crates/matrix-sdk-indexeddb/src/media_store/mod.rs
```diff
@@ -66,12 +66,6 @@ pub struct IndexeddbMediaStore {
     // A service for conveniently delegating media-related queries to an `MediaStoreInner`
     // implementation
     media_service: MediaService,
-    // An in-memory store for providing temporary implementations for
-    // functions of `MediaStore`.
-    //
-    // NOTE: This will be removed once we have IndexedDB-backed implementations for all
-    // functions in `MediaStore`.
-    memory_store: MemoryMediaStore,
```
bnjbvr, line 74: \o/ \o/ \o/

crates/matrix-sdk-indexeddb/src/transaction/mod.rs
mgoldenberg: Unfortunately, they are not... The first is a `KeyCursor<'_, Index<'_>>` while the second is a `KeyCursor<'_, ObjectStore<'_>>`. 

In theory, it would be possible to rely on their implementations of [`QuerySource`](https://docs.rs/indexed_db_futures/latest/indexed_db_futures/query_source/trait.QuerySource.html#) in order to create a `KeyCursor<'_, Box<dyn QuerySource>>`; however, [`QuerySource` is not implemented for `Box<dyn QuerySource>`](https://docs.rs/indexed_db_futures/latest/indexed_db_futures/query_source/trait.QuerySource.html#implementors) and, furthermore, [`QuerySource` is a sealed trait](https://docs.rs/indexed_db_futures/latest/src/indexed_db_futures/query_source.rs.html#29).

I do agree, though, that this would be a really nice improvement, so perhaps it's something we could add to our fork of `rust-indexed-db` and then try to upstream.

mgoldenberg: Or perhaps I could add a trait, say `CursorExt`, that I implement on both `KeyCursor<'_, Index<'_>>` and `KeyCursor<'_, ObjectStore<'_>>` and then wrap each in a `Box<dyn CursorExt>`...

bnjbvr: Nah don't worry about it, let's keep it simple like this 👍

mgoldenberg: Sounds good!

crates/matrix-sdk-indexeddb/src/transaction/mod.rs
mgoldenberg: Yes, agree - that would be a much better name!

crates/matrix-sdk-indexeddb/src/media_store/serializer/indexed_types.rs
mgoldenberg: fixup: [`19aec68e4a11c023b4af5e6d9595ecdb52ea3fe8`](https://github.com/matrix-org/matrix-rust-sdk/pull/5749/commits/19aec68e4a11c023b4af5e6d9595ecdb52ea3fe8)

crates/matrix-sdk-indexeddb/src/transaction/mod.rs
mgoldenberg: fixup: [`3b967d7c4be9cb57ec6aba2b6e634d84bb621a09`](https://github.com/matrix-org/matrix-rust-sdk/pull/5749/commits/3b967d7c4be9cb57ec6aba2b6e634d84bb621a09)

crates/matrix-sdk-indexeddb/src/media_store/transaction.rs
mgoldenberg: fixup: [`997ff0dc22ab9c252794038b07b0b967ff137406`](https://github.com/matrix-org/matrix-rust-sdk/pull/5749/commits/997ff0dc22ab9c252794038b07b0b967ff137406)

crates/matrix-sdk-indexeddb/src/media_store/serializer/indexed_types.rs
mgoldenberg: fixup: [`7b1a8be908f6bf9e6820886903df8507ae921add`](https://github.com/matrix-org/matrix-rust-sdk/pull/5749/commits/7b1a8be908f6bf9e6820886903df8507ae921add)

crates/matrix-sdk-indexeddb/src/media_store/serializer/indexed_types.rs
mgoldenberg: fixup: [`7b1a8be908f6bf9e6820886903df8507ae921add`](https://github.com/matrix-org/matrix-rust-sdk/pull/5749/commits/7b1a8be908f6bf9e6820886903df8507ae921add)

crates/matrix-sdk-indexeddb/src/media_store/serializer/indexed_types.rs
mgoldenberg: fixup: [`7625fd478ece7b3e36b72d403ac11a8c89be6c29`](https://github.com/matrix-org/matrix-rust-sdk/pull/5749/commits/7625fd478ece7b3e36b72d403ac11a8c89be6c29)

crates/matrix-sdk-indexeddb/src/media_store/transaction.rs
mgoldenberg: fixup: [`c971a2c597bd195fcafb8611163704f5c7fd8791`](https://github.com/matrix-org/matrix-rust-sdk/pull/5749/commits/c971a2c597bd195fcafb8611163704f5c7fd8791)

mgoldenberg:
> Do you want to squash the fixup commits, or should I squash the entire PR into a single commit?

Ivan has opted for squashing the fixup commits for the other pull requests, so let's stick with that! Let me know if you want me to rebase!

PR #5750, opened 2025-10-06
Make LoginProgres::EstablishingSecureChannel generic in order to reuse it for the other QR login flow

Johennes:
This was split off from https://github.com/matrix-org/matrix-rust-sdk/pull/5711 and is a preparatory step for implementing the missing QR login path

CC @poljar


Johennes:
Wait but it's not actually a breaking change in matrix-sdk-ffi or is it? We're still returning the non-generic `QrLoginProgress::EstablishingSecureChannel`. It's just mapped from a different SDK value.

poljar:
> Wait but it's not actually a breaking change in matrix-sdk-ffi or is it? We're still returning the non-generic `QrLoginProgress::EstablishingSecureChannel`. It's just mapped from a different SDK value.

Ah duh, I looked at the enum variant changing from a struct to a tuple variant but that also only happened on the SDK side.

Sorry for the confusion.

Johennes:
No worries at all. Thanks for the review. 👍

PR #5752, opened 2025-10-06
feat(spaces): reuse existing room display name computation logic for spaces

stefanceriu:
This PR add a new public static method on the base-sdk's Room that allows computing room display names with much of the same logic as real rooms but from discrete values.

### DIFF
--- a/crates/matrix-sdk-base/src/room/display_name.rs
+++ b/crates/matrix-sdk-base/src/room/display_name.rs
@@ -17,7 +17,7 @@ use std::fmt;
 use as_variant::as_variant;
 use regex::Regex;
 use ruma::{
-    OwnedMxcUri, OwnedUserId, UserId,
+    OwnedMxcUri, OwnedUserId, RoomAliasId, UserId,
     events::{SyncStateEvent, member_hints::MemberHintsEventContent},
 };
 use serde::{Deserialize, Serialize};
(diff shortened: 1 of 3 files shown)


crates/matrix-sdk-ui/src/spaces/room.rs
```diff
@@ -91,10 +102,20 @@ impl SpaceRoom {
...
+            room_info.heroes().to_vec(),
+            known_room.joined_members_count(),
+        )
+        .to_string();
+
         Self {
             room_id: room_info.room_id().to_owned(),
             canonical_alias: room_info.canonical_alias().map(ToOwned::to_owned),
-            name: room_info.name().map(ToOwned::to_owned),
+            name,
```
ganfra, line 117: Should we keep same names than for RoomSummary, ie. `raw_name` and `name`?

stefanceriu: Well the the underlying BaseRoom and BaseRoomInfo call them `name` and `display_name` so I went with that ¯\_(ツ)_/¯

crates/matrix-sdk-base/src/room/display_name.rs
```diff
@@ -60,6 +60,31 @@ impl Room {
...
+            (None, None) => {
+                let hero_display_names =
+                    heroes.into_iter().filter_map(|hero| hero.display_name).collect::<Vec<_>>();
+
+                compute_display_name_from_heroes(
+                    num_joined_members,
+                    hero_display_names.iter().map(|name| name.as_str()).collect(),
+                )
+            }
+        }
```
poljar, line 85: Shame that this is almost exactly the same as this code snippet: https://github.com/matrix-org/matrix-rust-sdk/blob/b32ee9f80b9049519614506dbcb7329a8ca0f656/crates/matrix-sdk-base/src/room/display_name.rs#L83-L97

The added complexity in the snippet is due to the handling of the service members.

Couldn't you store the service members event in your `SpaceRoom` so we can reuse the exact logic we use for normal rooms?

This doesn't need to happen as part of this PR, but I think it would avoid this code duplication and it would be even more correct™.

crates/matrix-sdk-ui/src/spaces/room.rs
stefanceriu: https://github.com/matrix-org/matrix-rust-sdk/pull/5757

PR #5758, opened 2025-10-07
fix: Make impl Stream return type not use any lifetime

zecakeh:
With Rust 2024, by default `impl Trait` return types use any hidden generic parameter that is in scope, so in these cases the lifetime of `self`.

But since the return type is actually owned, the returned `impl Stream` shouldn't use any lifetime, which is what `use<>` does.

For more details: https://blog.rust-lang.org/2024/09/05/impl-trait-capture-rules/.


bnjbvr (approved):
Thanks! Can we also bump the Rust edition for the crypto crate to 2024 then?

zecakeh:
Oh I didn't notice that it was not on Rust 2024.

zecakeh:
Do you mean in this PR by the way? There should be a lot of changes due to rustfmt so I think it should be in its own PR.

bnjbvr:
You're right, we should likely keep that for a different PR.

PR #5760, opened 2025-10-08
feat(oauth): add LoginProgress::SyncingSecrets

Johennes:
This was split off from https://github.com/matrix-org/matrix-rust-sdk/pull/5711. It is technically not a required preparatory step for implementing the missing QR login path. It helps clients display progress during the login process and is a UX improvement regardless of which flavor of QR login is used.

Signed-off-by:


poljar (approved):
Thanks for adding both changelog entries and a motivation for this.

Issue #5613, opened 2025-09-02
History-sharing-on-invite is broken when "Exclude insecure devices" is set


richvdh:
STR:
 * Enable "Exclude insecure devices when sending/receiving messages" and "Share encrypted history with new members" labs flags
 * Have a friend enable "Share encrypted history with new members", and invite you to an encrypted room
 * Fail to decrypt history

I think this used to work, but has regressed in https://github.com/matrix-org/matrix-rust-sdk/pull/5074 and https://github.com/matrix-org/matrix-rust-sdk/pull/5319. The problem is:

 * `sender_device` is set to `None` [here](https://github.com/matrix-org/matrix-rust-sdk/pull/5074/files#diff-909c5de6771a2992465b595c86bbc6baacc92acd231a79cee788f16c8d4fcbd5R1512) when the event type is `io.element.msc4268.room_key_bundle`
 * so, `encryption_info.verification_state` is always set to `VerificationState::Unverified(VerificationLevel::None(DeviceLinkProblem::MissingDevice)))`,
 * so, `is_from_verified_device_or_allowed_type` returns `false` when  "Exclude insecure devices" is enabled
 * so, the event is ignored.

PR #5763, opened 2025-10-09
crypto: Fix bugs in processing incoming encrypted to-device messages

richvdh:
This PR fixes a pair of bugs around how device verification on incoming encrypted to-device messages. See individual commits for more details.

Fixes: https://github.com/matrix-org/matrix-rust-sdk/issues/5613
Fixes: #5768

### DIFF
--- a/crates/matrix-sdk/src/room/shared_room_history.rs
+++ b/crates/matrix-sdk/src/room/shared_room_history.rs
@@ -132,6 +132,9 @@ pub(crate) async fn maybe_accept_key_bundle(room: &Room, inviter: &UserId) -> Re
 
     // Ensure that we get a fresh list of devices for the inviter, in case we need
     // to recalculate the `SenderData`.
+    // XXX: is this necessary, given (with exclude-insecure-devices), we should have
+    // checked that the inviter device was cross-signed when we received the
+    // to-device message?
     let (req_id, request) =
         olm_machine.query_keys_for_users(iter::once(bundle_info.sender_user.as_ref()));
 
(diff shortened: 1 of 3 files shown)


crates/matrix-sdk-crypto/src/olm/account.rs
```diff
@@ -1569,23 +1573,41 @@ impl Account {
...
+                    .into());
+                }
+                Ok(Some(device))
+            }
 
-        let key = device.ed25519_key().ok_or(EventError::MissingSigningKey)?;
+            (None, Some(sender_device_keys)) => {
+                // We have already validated the signature on `sender_device_keys`, so this
+                // try_into cannot fail.
+                let sender_device_data = sender_device_keys.try_into().unwrap();
```
poljar: No naked unwraps please, use an `expect()` to explain why it can't fail. 

Alternatively, is there a good reason to not do this conversion in `check_sender_device_keys()` so the `expect()` isn't necessary at all?

richvdh: I'll replace this with an `expect` for now, but it's my opinion that there should be a way to convert a `DeviceKeys` whose signature we have already checked into a `DeviceData` (so, a `DeviceData::from_validated_device_keys` or similar): this isn't the only place we have to do it.

If we *really* want to track whether the signature has been checked via the type system, we should do so with a `ValidatedDeviceKeys` which wraps `DeviceKeys`, without all the other clutter that comes with `DeviceData`.

> Alternatively, is there a good reason to not do this conversion in `check_sender_device_keys()` so the `expect()` isn't necessary at all?

It just feels like a bit of a confusing API for `check_sender_device_keys` to wrap the `DeviceKeys` with all this other data which we're *probably* going to throw away, and to provide a bunch of methods which don't really make sense.

PR #5764, opened 2025-10-09
fix(spaces): handle empty string room names when computing the display names

### LINKED ISSUE #5762: SpaceRoom.displayName can be wrong
The following scenario seems to break the logic somehow : 

- Create a room in a space from any client supporting this
- Invite a bunch of people in this room (one might be enough)
- Remove the name of the room 

At this point, the SpaceRoom.displayName gets messy.
Exemple if I have "user 1" and "user 2" in the room (plus me), I can get randomly(?) : 
- "user1, user2"
- "3 people"
- an empty string

When I open the room, `RoomInfo.displayName` has the correct value "user1, user2"

stefanceriu:
Fixes #5762

### DIFF
--- a/crates/matrix-sdk-base/src/room/display_name.rs
+++ b/crates/matrix-sdk-base/src/room/display_name.rs
@@ -70,6 +70,10 @@ impl Room {
         heroes: Vec<RoomHero>,
         num_joined_members: u64,
     ) -> RoomDisplayName {
+        // Handle empty string names. The `Room` level implementation relies
+        // on `RoomInfo` doing the same thing.
+        let name = name.and_then(|name| (!name.is_empty()).then_some(name));
+
         match (name, canonical_alias) {
             (Some(name), _) => RoomDisplayName::Named(name.trim().to_owned()),
             (None, Some(alias)) => RoomDisplayName::Aliased(alias.alias().trim().to_owned()),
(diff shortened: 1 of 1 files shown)


crates/matrix-sdk-base/src/room/display_name.rs
```diff
@@ -70,6 +70,10 @@ impl Room {
         heroes: Vec<RoomHero>,
         num_joined_members: u64,
     ) -> RoomDisplayName {
+        // Handle empty string names. The `Room` level implementation relies
+        // on `RoomInfo` doing the same thing.
```
stefanceriu, line 74: I would argue we should remove the special handling from the RoomInfo level too but I'm not entirely convinced. 
Everybody should be using the display name and not we shoud take the training wheels off and deliver what we actually have. 🤔

poljar: But according to the [spec](https://spec.matrix.org/v1.16/client-server-api/#mroomname), empty string should be treated the same as if there's no such event:

> If a room has an m.room.name event with an absent, null, or empty name field, it should be treated the same as a room with no m.room.name event.

stefanceriu: Well TIL but okay then ¯\_(ツ)_/¯

PR #5772, opened 2025-10-11
feat(timeline): Use read receipt as fallback for read marker

VAWVAW:
According to the [spec](https://spec.matrix.org/v1.16/client-server-api/#events-7) if a `m.fully_read` event does not exist, the normal `m.read` marker should be used as a fallback.

### DIFF
--- a/crates/matrix-sdk-ui/src/timeline/controller/mod.rs
+++ b/crates/matrix-sdk-ui/src/timeline/controller/mod.rs
@@ -1012,11 +1012,18 @@ impl<P: RoomDataProvider> TimelineController<P> {
                 .await;
         }
 
-        if track_read_markers
-            && let Some(fully_read_event_id) =
+        if track_read_markers {
+            if let Some(fully_read_event_id) =
                 self.room_data_provider.load_fully_read_marker().await
-        {
-            state.handle_fully_read_marker(fully_read_event_id);
+            {
+                state.handle_fully_read_marker(fully_read_event_id);
+            } else if let Some(latest_receipt_event_id) = state
+                .latest_user_read_receipt_timeline_event_id(self.room_data_provider.own_user_id())
+            {
+                // Fall back to read receipt if no fully read marker exists.
+                debug!("no `m.fully_read` marker found, falling back to read receipt");
+                state.handle_fully_read_marker(latest_receipt_event_id);
+            }
         }
     }
 
--- a/crates/matrix-sdk-ui/src/timeline/controller/read_receipts.rs
+++ b/crates/matrix-sdk-ui/src/timeline/controller/read_receipts.rs
@@ -65,7 +65,7 @@ impl ReadReceipts {
 
     /// Read the latest read receipt of the given type for the given user, from
     /// the in-memory cache.
-    fn get_latest(
+    pub(crate) fn get_latest(
         &self,
         user_id: &UserId,
         receipt_type: &ReceiptType,
--- a/crates/matrix-sdk-ui/tests/integration/timeline/thread.rs
+++ b/crates/matrix-sdk-ui/tests/integration/timeline/thread.rs
@@ -30,6 +30,7 @@ use matrix_sdk_test::{
 };
 use matrix_sdk_ui::timeline::{
     RoomExt as _, TimelineBuilder, TimelineDetails, TimelineEventItemId, TimelineFocus,
+    VirtualTimelineItem,
 };
 use ruma::{
     MilliSecondsSinceUnixEpoch,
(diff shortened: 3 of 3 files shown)


zecakeh:
This change doesn't really make sense to me. Could you quote the part of the spec that says that? I don't see it.

zecakeh:
Ah nevermind sorry I found it:

> If the event does not exist on the user’s account data, the fully read marker should be considered to be the user’s read receipt location.

I am not convinced that it makes sense because if there is no fully read marker it should mean that the user hasn't "fully" read the room previously, but oh well.

crates/matrix-sdk-ui/src/timeline/controller/mod.rs
```diff
@@ -1012,11 +1012,20 @@ impl<P: RoomDataProvider> TimelineController<P> {
...
+            if let Some(fully_read_event_id) =
                 self.room_data_provider.load_fully_read_marker().await
-        {
-            state.handle_fully_read_marker(fully_read_event_id);
+            {
+                state.handle_fully_read_marker(fully_read_event_id);
+            } else if let Some((last_read_event_id, _)) = state
+                .meta
+                .read_receipts
+                .get_latest(self.room_data_provider.own_user_id(), &ReceiptType::Read)
```
andybalaam: I may be understanding this wrong, but does this point at the read receipt event itself? If so I think that's not right: surely it should point at the event _pointed to_ by the read receipt, not point at the read receipt?

andybalaam:
Also, thank you for the contribution, and for the short and easy to read change!

crates/matrix-sdk-ui/src/timeline/controller/mod.rs
VAWVAW: The read receipt is ephemeral and not an event in the timeline. The function `get_latest` returns the event pointed to by the receipt and some context about when the receipt was sent and for which thread it is.

andybalaam: Great, thank you.

andybalaam: Should we also check `ReceiptType::ReadPrivate` ?

VAWVAW: Probably. When I wrote the code, I wasn't sure whether the fully read marker is public but it seems to be private.

How do you think we should handle the case where both a public and private receipt exist, but we haven't loaded their events. In this case we can't determine which receipt is newer. I think the safest option is to use the public read marker since some clients might not set private read markers.

andybalaam: > Probably. When I wrote the code, I wasn't sure whether the fully read marker is public but it seems to be private.

The fully read marker is part of account data, so private to the user, but I'm not sure that is relevant.

> How do you think we should handle the case where both a public and private receipt exist, but we haven't loaded their events. In this case we can't determine which receipt is newer. I think the safest option is to use the public read marker since some clients might not set private read markers.

The spec says that we should use the latest of the public and private receipt, if both exist, but as you say if we don't have the events we don't know which is latest. If we have only one of the two events, we should use that one, since the event we don't have is likely to be old. If we have neither, it probably doesn't matter too much: I think we can use either. Some clients will only set private read receipts, some will only set public ones, and some will be a mixture, especially if the user recently updated their settings e.g. to make them private.

So I think to be fully correct your code should get both events pointed at by receipts, try to figure out which is latest, pick the one we have if we only have one, and then pick one arbitrarily if we have two and we can't decide their order.

andybalaam: ... which fits the change you have done in f210b3b so if you rebase I will merge. Thanks again!

andybalaam: Oh sorry one more thing: please can we have some tests that check the new code you have added?

crates/matrix-sdk-ui/src/timeline/controller/mod.rs
```diff
@@ -1012,11 +1012,47 @@ impl<P: RoomDataProvider> TimelineController<P> {
...
+            if let Some(fully_read_event_id) =
                 self.room_data_provider.load_fully_read_marker().await
-        {
-            state.handle_fully_read_marker(fully_read_event_id);
+            {
+                state.handle_fully_read_marker(fully_read_event_id);
+            } else {
+                // Fall back to read receipt if no fully read marker exists.
+                debug!("no `m.fully_read` marker found, falling back to read receipt");
+                let public_receipt = state
```
zecakeh: It looks like you could use `TimelineState::latest_user_read_receipt` instead of all this.

VAWVAW:
I squashed the commits and used `TimelineState::latest_user_read_receipt_timeline_event_id` to calculate the latest receipt. I think the edit to `test_send_read_receipts` should cover this PR.

crates/matrix-sdk-ui/src/timeline/controller/mod.rs
VAWVAW: Yes, you are right.

crates/matrix-sdk-ui/src/timeline/controller/mod.rs
VAWVAW: After the latest change there isn't that much new code, so I think the edit to the existing test should be enough.

andybalaam: Yeah agreed.

andybalaam:
Spec reference: https://spec.matrix.org/v1.16/client-server-api/#events-7

PR #5778, opened 2025-10-13
feat(threads): better support redactions of in-thread events

bnjbvr:
This adds better support for redactions of in-thread events:

- when an in-thread event has been redacted, *remove* the redacted event from the thread (see also: https://github.com/matrix-org/matrix-js-sdk/pull/3798 for the underlying reasoning).
- also correctly update the thread summary (latest event and replies count)
- in the timeline, make sure that a redaction will always update replies to the redacted event item, even if the replied-to item isn't in the timeline anymore
- remove a thread summary when the number of replies has reached 0

Part of #4869.

### DIFF
--- a/crates/matrix-sdk/tests/integration/event_cache/threads.rs
+++ b/crates/matrix-sdk/tests/integration/event_cache/threads.rs
@@ -797,9 +797,15 @@ async fn test_redact_touches_threads() {
     let thread_resp1 = s.events[0].get_field::<OwnedEventId>("event_id").unwrap().unwrap();
     let thread_resp2 = s.events[1].get_field::<OwnedEventId>("event_id").unwrap().unwrap();
 
+    let room = s.server.sync_joined_room(&s.client, &s.room_id).await;
+
+    let (room_event_cache, _drop_handles) = room.event_cache().await.unwrap();
+
+    let (thread_events, mut thread_stream) =
+        room_event_cache.subscribe_to_thread(thread_root_id.to_owned()).await;
+
     // Receive a thread root, and a threaded reply.
-    let room = s
-        .server
+    s.server
         .sync_room(
             &s.client,
             JoinedRoomBuilder::new(&s.room_id)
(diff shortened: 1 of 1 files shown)


crates/matrix-sdk/src/event_cache/room/threads.rs
```diff
@@ -147,6 +147,32 @@ impl ThreadEventCache {
         }
     }
 
+    /// Remove an event from an thread event linked chunk, if it exists.
+    ///
+    /// If the event has been found and removed, then an update will be
+    /// propagated to observers.
+    pub(crate) fn try_remove_event(&mut self, event_id: &EventId) {
```
andybalaam: Does "try" in the name imply we return `Result`? Maybe `remove_event_if_present`? Or even just `remove`?

crates/matrix-sdk/tests/integration/event_cache/threads.rs
```diff
@@ -780,3 +780,112 @@ async fn test_auto_subscribe_on_thread_paginate_root_event() {
...
+
+        // The room event is redacted.
+        {
+            assert_let!(VectorDiff::Set { index: 1, value: new_event } = &diffs[1]);
+            let deserialized = new_event.raw().deserialize().unwrap();
+
+            // TODO: replace with https://github.com/ruma/ruma/pull/2254 when it's been merged in Ruma.
+            assert!(match deserialized {
+                AnySyncTimelineEvent::MessageLike(ev) => ev.is_redacted(),
+                AnySyncTimelineEvent::State(_ev) => unreachable!(),
```
andybalaam, line 884: Is it OK to use `unreachable` in a test? Isn't it undefined behaviour to read an unreachable statement, so in theory the generated code might not check whether we get here?

andybalaam: (meaning if we introduce a bug making us get here, the test could still pass)

andybalaam: Nope, ignore me: https://doc.rust-lang.org/std/macro.unreachable.html says this always panics. I was thinking of `unreachable_unchecked`.

bnjbvr: Exactly, this would panic in a regular build 👍

crates/matrix-sdk/src/event_cache/room/threads.rs
bnjbvr: Good point, will rename it to `remove_if_present` 👍

PR #5780, opened 2025-10-13
Verify membership when test if name is ambiguous

### LINKED ISSUE #5774: Room member display names incorrectly marked as ambiguous
The [spec](https://spec.matrix.org/v1.16/client-server-api/#calculating-the-display-name-for-a-user) explicitly says that only users with membership `join` or `invite` should be considered as possibly ambiguous.

But `RoomMember::name_ambiguous` for a member returned from `matrix_sdk::room::Room::members_no_sync` will return `true` if another user with the same name exists with membership `leave`.

multisme:
Verify if the membership is not Leave when checking if a name in a room is ambiguous.

Signed-off-by: 
multi

Fix https://github.com/matrix-org/matrix-rust-sdk/issues/5774

### DIFF
--- a/crates/matrix-sdk-base/src/room/members.rs
+++ b/crates/matrix-sdk-base/src/room/members.rs
@@ -179,10 +179,13 @@ impl Room {
             .transpose()?
             .map(|e| e.content.ignored_users.into_keys().collect());
 
+        let active_users = self.store.get_user_ids(self.room_id(), RoomMemberships::ACTIVE).await?;
+
         Ok(MemberRoomInfo {
             power_levels: power_levels.into(),
             max_power_level,
             users_display_names,
+            active_users,
             ignored_users,
         })
     }
@@ -211,18 +214,23 @@ impl RoomMember {
         presence: Option<PresenceEvent>,
         room_info: &MemberRoomInfo<'_>,
     ) -> Self {
-        let MemberRoomInfo { power_levels, max_power_level, users_display_names, ignored_users } =
-            room_info;
+        let MemberRoomInfo {
+            power_levels,
+            max_power_level,
+            users_display_names,
+            ignored_users,
+            active_users,
+        } = room_info;
 
         let display_name = event.display_name();
-        let membership = event.membership();
 
-        println!("{:?} {:?}", users_display_names, &display_name);
         let display_name_ambiguous = users_display_names.get(&display_name).is_some_and(|s| {
+            // s.filter(|n| )
             if !is_display_name_ambiguous(&display_name, s) {
                 return false;
             }
-            matches!(*membership, MembershipState::Join | MembershipState::Invite)
+            //We check of many active_users with the same surname exist
+            active_users.iter().filter(|u| s.contains(*u)).count() > 1
         });
         let is_ignored = ignored_users.as_ref().is_some_and(|s| s.contains(event.user_id()));
 
@@ -396,6 +404,7 @@ pub(crate) struct MemberRoomInfo<'a> {
     pub(crate) max_power_level: i64,
     pub(crate) users_display_names: HashMap<&'a DisplayName, BTreeSet<OwnedUserId>>,
     pub(crate) ignored_users: Option<BTreeSet<OwnedUserId>>,
+    pub(crate) active_users: Vec<OwnedUserId>,
 }
 
 /// The kind of room member updates that just happened.
@@ -574,12 +583,9 @@ mod tests {
                 .or_default()
                 .entry(StateEventType::RoomMember)
                 .or_default();
-            
-            let ambiguity_maps = changes
-                .ambiguity_maps
-                .entry(room.room_id().to_owned())
-                .or_default();
 
+            let ambiguity_maps =
+                changes.ambiguity_maps.entry(room.room_id().to_owned()).or_default();
 
             let display_name = DisplayName::new("Carol");
             members.insert(carol.into(), f.member(carol).display_name("Carol").into());
@@ -618,7 +624,6 @@ mod tests {
             members.insert(mewto.into(), f.member(mewto).display_name("Me").into());
             ambiguity_maps.entry(display_name.clone()).or_default().insert(mewto.to_owned());
 
-
             store.save_changes(&changes).await.unwrap();
         }
 
@@ -633,7 +638,6 @@ mod tests {
 
         assert!(!room.get_member(julie).await.unwrap().expect("Julie user").name_ambiguous());
         assert!(!room.get_member(bob).await.unwrap().expect("Bob user").name_ambiguous());
-
 
         assert!(room.get_member(me).await.unwrap().expect("Me user").name_ambiguous());
         assert!(room.get_member(mewto).await.unwrap().expect("Mewto user").name_ambiguous());
--- a/crates/matrix-sdk-base/src/room/members.rs
+++ b/crates/matrix-sdk-base/src/room/members.rs
@@ -216,6 +216,8 @@ impl RoomMember {
 
         let display_name = event.display_name();
         let membership = event.membership();
+
+        println!("{:?} {:?}", users_display_names, &display_name);
         let display_name_ambiguous = users_display_names.get(&display_name).is_some_and(|s| {
             if !is_display_name_ambiguous(&display_name, s) {
                 return false;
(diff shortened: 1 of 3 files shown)


multisme:
It's pretty straightforward but should I had some tests?

crates/matrix-sdk-base/src/room/members.rs
```diff
@@ -215,9 +215,10 @@ impl RoomMember {
             room_info;
 
         let display_name = event.display_name();
-        let display_name_ambiguous = users_display_names
-            .get(&display_name)
-            .is_some_and(|s| is_display_name_ambiguous(&display_name, s));
+        let membership = event.membership();
+        let display_name_ambiguous = users_display_names.get(&display_name).is_some_and(|s| {
+            is_display_name_ambiguous(&display_name, s) && *membership != MembershipState::Leave
```
stefanceriu: So we *don't* want to disambiguate the name of a left user, why is that? I see the AmbiguityCache has a concept of active members so what are we trying to achieve here?

multisme: It's because of the [spec](https://spec.matrix.org/v1.16/client-server-api/#calculating-the-display-name-for-a-user). It only sees member with an join or an invite membership as needing disambiguation

multisme: I updated the code to follow more precisely the spec.

crates/matrix-sdk-base/src/room/members.rs
```diff
@@ -215,9 +215,11 @@ impl RoomMember {
             room_info;
 
         let display_name = event.display_name();
-        let display_name_ambiguous = users_display_names
-            .get(&display_name)
-            .is_some_and(|s| is_display_name_ambiguous(&display_name, s));
+        let membership = event.membership();
+        let display_name_ambiguous = users_display_names.get(&display_name).is_some_and(|s| {
+            is_display_name_ambiguous(&display_name, s)
+                && (*membership == MembershipState::Join || *membership == MembershipState::Invite)
```
bnjbvr: Can you write the second part with a match statement, please? This will make it so that we explicitly handle all the membership cases, and we don't forget to think about handling it, would we have a new membership kind later on.

```rust
if !is_display_name_ambiguous(&display_name, s) {
  return false;
}

match *membership {
 // handle all variants explicitly here
}
```

bnjbvr (changes requested):
Thanks for the patch! Can you add a test please? Ideally it would show the case that's failing on main (but supposed to work), and that would be working with your patch now.

crates/matrix-sdk-base/src/room/members.rs
multisme: Ok I will

multisme:
@bnjbvr 
I have a problem here. There is a test that is failing that should be failing 
https://github.com/matrix-org/matrix-rust-sdk/blob/7a93c57419f1414cc37885afe91d48f1392e6fa8/crates/matrix-sdk-base/src/room/members.rs#L615-L622
Unless I misunderstood, the name should be here ambiguous and the test should on line 621 should not be failing. Should I make a new issue ?

bnjbvr:
It sounds like the code you've changed introduced a regression, so, no, we can't merge this as is and open a new issue. Instead, one should investigate the issue and attempt to fix it :)

multisme:
So i checked, and it seems that when using saving_changes to store the member events in MemoryStore, the HashMap `display_names: HashMap<OwnedRoomId, HashMap<DisplayName, BTreeSet<OwnedUserId>>>`  is never updated unless you expressly  do it. But this is the Hashmap used by is_the_display_name_ambiguous.
If this is not the behavior expected, I can make an issue and a PR for making sure that hashmap is populated, wait for it to get merged, and then finish this PR.

multisme:
@bnjbvr 
So i checked, and it seems that when using saving_changes to store the member events in MemoryStore, the HashMap `display_names: HashMap<OwnedRoomId, HashMap<DisplayName, BTreeSet<OwnedUserId>>>`  is never updated unless you expressly  do it. But this is the Hashmap used by is_the_display_name_ambiguous.
If this is not the behavior expected, I can make an issue and a PR for making sure that hashmap is populated, wait for it to get merged, and then finish this PR.

multisme:
I made a mistake anyway, I should check the membership of every RoomMember with the same displayname, I'm getting back into IT.

multisme:
@poljar Should I make an issue regarding this?

> So i checked, and it seems that when using saving_changes to store the member events in MemoryStore, the HashMap `display_names: HashMap<OwnedRoomId, HashMap<DisplayName, BTreeSet<OwnedUserId>>>` is never updated unless you expressly do it. But this is the Hashmap used by is_the_display_name_ambiguous. If this is not the behavior expected, I can make an issue and a PR for making sure that hashmap is populated, wait for it to get merged, and then finish this PR.

zecakeh:
Just my 2 cents, but why don't we keep only the active members in the display names cache, rather than having unnecessary data in it, and then use a list of active members to filter it?

multisme:
tbh that would make sense, but that would be a breaking change.

zecakeh:
Why would this be a breaking change? Isn't this data used only internally when building a `RoomMember`?

crates/matrix-sdk-base/src/room/members.rs
Hywan: Why not updating `is_display_name_ambiguous` to include this membership constraint? It seems to be really part of it.

crates/matrix-sdk-base/src/room/members.rs
Hywan: I think it's clearly worth adding this logic inside `is_display_name_ambiguous`.

crates/matrix-sdk-base/src/room/members.rs
```diff
@@ -216,6 +216,8 @@ impl RoomMember {
 
         let display_name = event.display_name();
         let membership = event.membership();
+
+        println!("{:?} {:?}", users_display_names, &display_name);
```
Hywan: Should be removed.

Hywan (changes requested):
Nice contribution, thank you.

I think we should merge the logic inside `is_display_name_ambiguous` to make less error-prone. What do you think?

multisme:
I messed up the history big will force push because of it.

crates/matrix-sdk-base/src/room/members.rs
multisme: Ok I will update is_display_name_ambiguous to take a MemberEvent instead of a string

crates/matrix-sdk-base/src/room/members.rs
```diff
@@ -226,13 +214,24 @@ impl RoomMember {
...
+            active_users,
+        } = room_info;
 
         let display_name = event.display_name();
-        let display_name_ambiguous = users_display_names
-            .get(&display_name)
-            .is_some_and(|s| is_display_name_ambiguous(&display_name, s));
+
+        let display_name_ambiguous = users_display_names.get(&display_name).is_some_and(|s| {
+            // s.filter(|n| )
```
Hywan, line 228: should be removed

Hywan (commented):
My comments don't seem to be addressed. Is it on purpose :-)?

crates/matrix-sdk-base/src/room/members.rs
multisme: No, taking care of that now.

multisme: Ok I looked back into it, and the reason I didn't remove it was that I would need to make is_display_name_ambiguous accept the list of current active users, which imply a lot of changes in `the matrix-sdk-base/src/store/ambiguity_map.rs`, so I wanted to confirm before starting that work.

Hywan: You can as new commits and see where it goes. If it's too complex, we will take another path. Thanks for asking.

multisme:
@Hywan I have created a new function get_active_users_display_name to replace get_users_display_name in the proper situations but in the case of IndexDB store, I have to either do two queries or write a migration. Writing a migration sounds cleaner but it feels like I should check before starting.

poljar:
@Hywan friendly ping, I think this one fell through the cracks somewhere.

Netfloex:
To add some context, the mautrix WhatsApp bridge migrated phone numbers to @lid identifiers last year ([v0.12.0](https://github.com/mautrix/whatsapp/releases/tag/v0.12.0)). This makes this issue extra annoying as this happens for most bridged rooms.

crates/matrix-sdk-sqlite/src/state_store.rs
```diff
@@ -1672,7 +1676,7 @@ impl StateStore for SqliteStateStore {
             .transpose()?
             .unwrap_or_default())
     }
-
+
```
Hywan, line 1679: Please, remove empty spaces.

crates/matrix-sdk-sqlite/src/state_store.rs
```diff
@@ -1720,6 +1724,76 @@ impl StateStore for SqliteStateStore {
...
+        let room_id = self.encode_key(keys::DISPLAY_NAME, room_id);
+        let mut names_map = display_names
+            .iter()
+            .flat_map(|display_name| {
+                // We encode the display name as the `raw_str()` and the normalized string.
+                //
+                // This is for compatibility reasons since:
+                //  1. Previously "Alice" and "alice" were considered to be distinct display
+                //     names, while we now consider them to be the same so we need to merge the
+                //     previously distinct buckets of user IDs.
```
Hywan, line 1770: Can you quote the spec where such operations are allowed? I can understand where the idea comes from, though.

Hywan (commented):
A couple more comments 🙂.

Hywan:
@multisme I think I would prefer a migration too. It's not a problem to migrate, we do that quite often.

Hywan:
@Netfloex Thanks for the feedback.

multisme:
@Hywan Ok I'm on it.

poljar:
It's been a while since the latest request for changes.

Feel free to reopen if you'd like to continue work on this.

PR #5781, opened 2025-10-14
Upgrade Ruma

zecakeh:
Brings changes to the requests metadata. It was changed from a struct to a trait, and the authentication scheme is now an associated type.

This allows to forbid at compile time requests that use an unsupported authentication scheme.

### DIFF
--- a/crates/matrix-sdk/src/http_client/mod.rs
+++ b/crates/matrix-sdk/src/http_client/mod.rs
@@ -245,6 +245,9 @@ async fn response_to_http_response(
 
 /// Marker trait to identify the authentication schemes that the [`HttpClient`]
 /// supports.
+///
+/// This trait can also be implemented for custom
+/// [`AuthScheme`](auth_scheme::AuthScheme)s if necessary.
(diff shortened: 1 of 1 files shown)


crates/matrix-sdk/src/http_client/mod.rs
```diff
@@ -254,6 +243,20 @@ async fn response_to_http_response(
     Ok(http_builder.body(body).expect("Can't construct a response using the given body"))
 }
 
+/// Marker trait to identify the authentication schemes that the [`HttpClient`]
+/// supports.
+pub trait SupportedAuthScheme: auth_scheme::AuthScheme {}
```
bnjbvr, line 251: It makes sense to me to make it public so that external users can also implement it, but might be worth commenting in the doc comment above that it's intended? (and if it's not intended, consider sealing the trait)

zecakeh: I added a sentence, is it okay like this?

bnjbvr: Sweet, thanks!

PR #5782, opened 2025-10-14
Add a totality test for the recency sorter

poljar:


crates/matrix-sdk-ui/src/room_list_service/sorters/latest_event.rs
```diff
@@ -40,7 +41,7 @@ where
         // - `Local*` < `Remote`.
         (true, false) => Ordering::Less,
 
-        // `true` and `false, i.e.:
+        // `true` and `true, i.e.:
```
bnjbvr:
```suggestion
        // `true` and `true`, i.e.:
```

crates/matrix-sdk-ui/src/room_list_service/sorters/recency.rs
```diff
@@ -302,4 +305,73 @@ mod tests {
...
+            let b_a_comparison = cmp_impl(b, a);
+
+            eprintln!("a.cmp(b) = {a_b_comparison:?}");
+            eprintln!("b.cmp(a) = {b_a_comparison:?}");
+
+            assert!(
+                (a_b_comparison == Less || a_b_comparison == Equal)
+                || (b_a_comparison == Less || b_a_comparison == Equal)
+            );
+        }
```
bnjbvr, line 376: I don't think this test brings any value, unfortunately: in Rust an order is total if it implements `Ord` in addition to `PartialOrd`, so by definition they're total.

That being said, something that would be nice to test, is that the comparison values are "opposite" of each other: if `a_b_comparison` is Less, then `b_a_comparison` is Greater, etc.

PR #5784, opened 2025-10-15
feat(ffi): add bindings for listening to global send queue updates

Johennes:
This is a small addendum to #5761 that does the same thing for the global update reporter.

CC @bnjbvr as the resident send queue expert.


poljar:
Could you please get rid of the merge conflict, looks good otherwise.

Johennes:
Have rebased on latest `main`.

PR #5786, opened 2025-10-16
feat(ffi): add bindings for logging in by generating a QR code on the new device

Johennes:
This adds FFI bindings for the QR login flow added in #5711.

### DIFF
--- a/bindings/matrix-sdk-ffi/src/client.rs
+++ b/bindings/matrix-sdk-ffi/src/client.rs
@@ -102,7 +102,7 @@ use crate::{
     encryption::Encryption,
     notification::NotificationClient,
     notification_settings::NotificationSettings,
-    qr_code::{HumanQrLoginError, QrCodeData, QrLoginProgressListener},
+    qr_code::LoginWithQrCodeHandler,
     room::{RoomHistoryVisibility, RoomInfoListener},
     room_directory_search::RoomDirectorySearch,
     room_preview::RoomPreview,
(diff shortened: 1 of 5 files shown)


bindings/matrix-sdk-ffi/src/qr_code.rs
```diff
@@ -172,3 +304,71 @@ impl From<qrcode::LoginProgress<QrProgress>> for QrLoginProgress {
...
+    /// we can verify that the channel is indeed secure.
+    QrScanned { check_code_sender: Arc<CheckCodeSender> },
+    /// We are waiting for the login and for the OAuth 2.0 authorization server
+    /// to give us an access token.
+    WaitingForToken { user_code: String },
+    /// We are syncing secrets.
+    SyncingSecrets,
+    /// The login has successfully finished.
+    Done,
+}
```
Johennes, line 329: It looks like UniFFI doesn't support generics in enumerations. So we either have to duplicate the enum, as it is done here, or use a union type where some variants are not used in one flow or the other. I'm not sure which of these options is least bad. Thoughts welcome.

bindings/matrix-sdk-ffi/src/client.rs
```diff
@@ -543,43 +543,17 @@ impl Client {
...
-    ///
-    /// The usage of the progress_listener is required to transfer the
-    /// [`CheckCode`] to the existing client.
+    /// # Arguments
     ///
-    /// [MSC4108]: https://github.com/matrix-org/matrix-spec-proposals/pull/4108
-    pub async fn login_with_qr_code(
+    /// * `oidc_configuration` - The data to restore or register the client with
+    ///   the server.
+    pub fn login_with_qr_code(
```
pixlwave, line 552: Retroactive comment having just adopted this API: Should this method be renamed to something like `start_qr_code_login`/`prepare_qr_code_login` or simply `new_qr_code_login_handler` (or similar), seeing as the method no longer performs a login directly and you need to use the handler to do the login?

I was a bit surprised at the shape of the API given the name (but it wasn't exactly hard to figure it out in the end) 🙂

poljar: Hmm, it's a typical builder pattern, perhaps it should have been named as such :thinking:

Johennes: I had tried to mimic the shape of the underlying API in the SDK. It's admittedly a bit odd here because nothing is being built which is why I called it "handler" rather than "builder". Given that the method is really just a factory, I like `new_qr_code_login_handler` best out of the suggested options. I don't feel strongly though.

PR #5788, opened 2025-10-16
fix: `NotificationSettings::unmute_room` didn't clear the cached notification mode

jmartinesp:
As a result, the clients still received the cached value in the `RoomInfo`s even though the push rule was updated.

A way to solve this could be to clear this value in `unmute_room` or some other point of this flow, but `SlidingSyncResponseProcessor::update_in_memory_caches` already updates this value *if* the received room has a custom notification mode, it just never clears it, so maybe it's the best place to also add its counterpart.

Fixed https://github.com/element-hq/element-x-android/issues/5546 when I tested this locally.

Signed-off-by:


jmartinesp:
True, let me see if I can find an easy way to test it.

poljar (approved):
Thanks, looks good.

PR #5795, opened 2025-10-21
IndexedDB: split media content and media metadata into separate object stores in `MediaStore`

mgoldenberg:
## Background

This pull request is part of a series of pull requests to add a full IndexedDB implementation of the `EventCacheStore` and  `MediaStore` (see #4617, #4996, #5090, #5138, #5226, #5274, #5343, #5384, #5406, #5414, #5497, #5506, #5540, #5574, #5603, #5676, #5682, #5749). This particular pull request changes the schema for how media is stored by splitting content and metadata into separate object stores.

The reason for this change is that IndexedDB does not allow partial updates to existing objects. So, when content and metadata are stored in the same object store, updating a metadata field requires deserializing and re-serializing the entire object, which includes content that has the potential to be very large - e.g., an image, a document, etc.

### Adding separate stores for content and metadata

The overarching change is to add two new object stores, one for content and one for metadata.

#### [`MediaContent`](https://github.com/mgoldenberg/matrix-rust-sdk/blob/85b814273f93aa3047cd21619357e8de3e72bcf9/crates/matrix-sdk-indexeddb/src/media_store/types.rs#L76-L83)

The content store is very straightforward and simply maps an identifier (`u64`) to the content, illustrated in the type below.

```rust
pub struct MediaContent {
    pub id: u64,
    pub data: Vec<u8>,
}
```

When storing a `MediaContent` object in IndexedDB, one must find an unused `u64`. Currently, this is accomplished by querying the object store for the largest identifier, incrementing it, and then using it as the identifier for the desired object.

Note that IndexedDB does offer [auto-incrementing keys](https://developer.mozilla.org/en-US/docs/Web/API/IDBObjectStore/autoIncrement); however, it's not clear if it's possible to retrieve the generated key upon insertion into the database via [`indexed_db_futures`](https://docs.rs/indexed_db_futures/latest/indexed_db_futures/index.html). So, one must ultimately query the database to get access to the key in a similar fashion to that described above.

#### [`MediaMetadata`](https://github.com/mgoldenberg/matrix-rust-sdk/blob/85b814273f93aa3047cd21619357e8de3e72bcf9/crates/matrix-sdk-indexeddb/src/media_store/types.rs#L57-L74)

The metadata store is almost identical to the original media store, but contains some additional information for tracking the identifier and the size of the [`MediaContent`](https://github.com/mgoldenberg/matrix-rust-sdk/blob/85b814273f93aa3047cd21619357e8de3e72bcf9/crates/matrix-sdk-indexeddb/src/media_store/types.rs#L76-L83), as illustrated in the type below.

```rust
pub struct MediaMetadata {
    pub request_parameters: MediaRequestParameters,
    pub last_access: UnixTime,
    pub ignore_policy: IgnoreMediaRetentionPolicy,
    pub content_id: u64,
    pub content_size: usize,
}
```

When storing a `MediaMetadata` object in IndexedDB, one must first store the `MediaContent`. Once the `MediaContent` is stored, one can determine its identifier and its encoded size which can be used to populate `MediaMetadata::content_id` and `MediaMetadata::content_size`.

Note that this means that retrieving `MediaContent` via `MediaRequestParameters` requires two steps.

1. Using `MediaRequestParameters` to retrieve `MediaMetadata` 
2. Using `MediaMetadata::content_id` to retrieve `MediaContent`

### Removing original media store

After the two object stores above were created and the implementations of various functions were updated to use those object stores, the original media store and its associated types were removed. There is one exception, which is that the top-level [`Media`](https://github.com/mgoldenberg/matrix-rust-sdk/blob/c4f676a191c11dcac1889b110b78de6867812ad5/crates/matrix-sdk-indexeddb/src/media_store/types.rs#L40-L56) type was kept in place, as it proved to be a useful top-level abstraction.

### Improvements

These changes offer significant improvements on the following operations.

- [`MediaStore::replace_media_key`](https://github.com/mgoldenberg/matrix-rust-sdk/blob/a82a0db02350438cd78be2e73c18e437d4a47229/crates/matrix-sdk-indexeddb/src/media_store/mod.rs#L145) - _changes request parameters - i.e., primary key - of media_
    - **Before:** read and write media metadata and media content
    - **After:** read and write only media metadata
- [`MediaStore::get_media_content`](https://github.com/mgoldenberg/matrix-rust-sdk/blob/a82a0db02350438cd78be2e73c18e437d4a47229/crates/matrix-sdk-indexeddb/src/media_store/mod.rs#L164) - _retrieves media by request parameters and sets last access time_
    - **Before:** read and write media metadata and media content
    - **After:** read media metadata and media content, write media metadata 
- [`MediaStore::get_media_content_for_uri`](https://github.com/mgoldenberg/matrix-rust-sdk/blob/a82a0db02350438cd78be2e73c18e437d4a47229/crates/matrix-sdk-indexeddb/src/media_store/mod.rs#L185) - _retrieves media by URI and sets the last access time_
    - **Before:** read and write media metadata and media content
    - **After:** read media metadata and media content, write media metadata
- [`MediaStore::set_ignore_media_retention_policy`](https://github.com/mgoldenberg/matrix-rust-sdk/blob/a82a0db02350438cd78be2e73c18e437d4a47229/crates/matrix-sdk-indexeddb/src/media_store/mod.rs#L221) - _sets whether to ignore media retention policy_
    - **Before:** read and write media metadata and media content
    - **After:** read and write only media metadata

### Penalties

On the other hand, there are also some penalties due to the updated schema.

- [`MediaStore::add_media_content`](https://github.com/mgoldenberg/matrix-rust-sdk/blob/c4f676a191c11dcac1889b110b78de6867812ad5/crates/matrix-sdk-indexeddb/src/media_store/mod.rs#L134) - _adds media_
    - **Before:** write media in one operation
    - **After:** write media in three operations (read metadata, write content, write metadata)
- [`MediaStore::get_media_content`](https://github.com/mgoldenberg/matrix-rust-sdk/blob/a82a0db02350438cd78be2e73c18e437d4a47229/crates/matrix-sdk-indexeddb/src/media_store/mod.rs#L164) - _retrieves media by request parameters and sets last access time_
    - **Before:** read and write media in two operations (read media, write media)
    - **After:** read and write media in three operations (read metadata, read content, write metadata)
- [`MediaStore::remove_media_content`](https://github.com/mgoldenberg/matrix-rust-sdk/blob/c4f676a191c11dcac1889b110b78de6867812ad5/crates/matrix-sdk-indexeddb/src/media_store/mod.rs#L174) - _removes media by request parameters_
    - **Before:** remove media in a single operation
    - **After:** remove media in three operations (read metadata, remove content, remove metadata)
- [`MediaStore::get_media_content_for_uri`](https://github.com/mgoldenberg/matrix-rust-sdk/blob/a82a0db02350438cd78be2e73c18e437d4a47229/crates/matrix-sdk-indexeddb/src/media_store/mod.rs#L185) - _retrieves media by URI and sets the last access time_
    - **Before:** read and write media in two operations (read media, write media)
    - **After:** read and write media in three operations (read metadata, read content, write metadata)
- [`MediaStore::remove_media_content_for_uri`](https://github.com/mgoldenberg/matrix-rust-sdk/blob/c4f676a191c11dcac1889b110b78de6867812ad5/crates/matrix-sdk-indexeddb/src/media_store/mod.rs#L198) - _removes media by URI_
    - **Before:** remove media in a single operation
    - **After:** remove media in three operations (read metadata, remove content, remove metadata)
- [`MediaStore::clean`](https://github.com/mgoldenberg/matrix-rust-sdk/blob/c4f676a191c11dcac1889b110b78de6867812ad5/crates/matrix-sdk-indexeddb/src/media_store/mod.rs#L238) - _clean store by removing oversized and old media_
    - **Before:** remove media ranges in a single operation
    - **After:** remove media ranges in many operations (read metadata, remove content, remove metadata)

## Conclusions

My feeling is that this implementation is an improvement overall. That being said, some benchmarking would offer a greater degree of confidence that [`MediaStore::clean`](https://github.com/mgoldenberg/matrix-rust-sdk/blob/c4f676a191c11dcac1889b110b78de6867812ad5/crates/matrix-sdk-indexeddb/src/media_store/mod.rs#L238) has not deteriorated significantly. If this is desired, we can pursue this to get a better sense of the penalty.  

In any case, I don't think it would be wise to return to a single object store, but perhaps there is some way to improve upon the split object stores.

## Future Work

- Refactor feature flags
- Expose `EventCacheStore` and `MediaStore` outside of the `matrix-sdk-indexeddb` 

---

### DIFF
--- a/crates/matrix-sdk-indexeddb/CHANGELOG.md
+++ b/crates/matrix-sdk-indexeddb/CHANGELOG.md
@@ -11,6 +11,11 @@ All notable changes to this project will be documented in this file.
 - [**breaking**] `IndexeddbCryptoStore::get_withheld_info` now returns `Result<Option<RoomKeyWithheldEntry>, ...>`
   ([#5737](https://github.com/matrix-org/matrix-rust-sdk/pull/5737))
 
+### Performance
+
+- Improve performance of certain media queries in `MediaStore` implementation by storing media content and media metadata
+  in separate object stores in IndexedDB (see [#5795](https://github.com/matrix-org/matrix-rust-sdk/pull/5795)).
+
 ## [0.14.0] - 2025-09-04
 
 No notable changes in this release.
@@ -48,6 +53,7 @@ No notable changes in this release.
 
 - `save_change` performance improvement, all encryption and serialization
   is done now outside of the db transaction.
+
 ### Bug Fixes
 
 - Use the `DisplayName` struct to protect against homoglyph attacks.


crates/matrix-sdk-indexeddb/src/media_store/types.rs
```diff
@@ -37,28 +37,50 @@ impl Lease {
...
+    pub content_id: u64,
+    /// The size in bytes of the associated [`MediaContent`]
+    pub content_size: usize,
+}
+
+/// A representation of media content which can be stored in IndexedDB.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct MediaContent {
+    /// The identifier associated with the given [`MediaContent::data`].
+    pub id: u64,
```
poljar: As far as I gather `MediaMetadata::content_id` and `MediaContent::id` are the unique key which is used to reference each other from one type to the other.

Can we name the fields the same and perhaps add a type alias or even wrapper type for the `u64` to make this a bit more obvious?

edit: Seems that the type alias was added in a later commit, so this part seems fine.

crates/matrix-sdk-indexeddb/src/media_store/serializer/indexed_types.rs
```diff
@@ -80,6 +82,9 @@ pub type IndexedMediaMetadata = MaybeEncrypted;
 /// is suitable for use in an IndexedDB key
 pub type IndexedMediaContentSize = usize;
 
+/// A (possibly) encrypted representation of [`MediaContent`]
+pub type IndexedMediaContentContent = Vec<u8>;
```
poljar: While I do enjoy the silliness of `ContentContent`, perhaps a better name would be `DataContent` or some similar variant of this.

crates/matrix-sdk-indexeddb/src/media_store/serializer/indexed_types.rs
```diff
@@ -569,3 +574,84 @@ impl<'a> IndexedPrefixKeyComponentBounds<'a, Media, IgnoreMediaRetentionPolicy>
...
+}
+
+/// The primary key of the [`MEDIA_CONTENT`][1] object store, which is
+/// constructed from:
+///
+/// - The identifier associated with the [`IndexedMediaContent`]
+///
+/// [1]: crate::media_store::migrations::v1::create_media_content_object_store
+#[derive(Debug, Serialize, Deserialize)]
+pub struct IndexedMediaContentIdKey(u64);
```
poljar: Why don't we use this type, perhaps with a more generic name, in the media structs?

crates/matrix-sdk-indexeddb/src/transaction/mod.rs
```diff
@@ -289,6 +269,37 @@ impl<'a> Transaction<'a> {
...
+            .with_query(range)
+            .with_direction(direction)
+            .serde()?
+            .await?
+        {
+            return cursor.next_key_ser().await.map_err(Into::into);
+        }
+        Ok(None)
+    }
+
```
poljar, line 304: Couldn't we just used random identifiers? A random `u64` sounds like it would do the job. Alternatively an UUID could be used as well. If the ordering is important, then an `ULID` could be used.

This would avoid the need for this function completely.

poljar (changes requested):
This looks pretty good, thanks for splitting this up into many small commits.

I wonder what you think about utilizing a different type for the IDs.

Besides that I left a couple of nits about consistent naming of the content ID.

poljar:
Oh, another thing. Please add a changelog entry.

crates/matrix-sdk-indexeddb/src/media_store/types.rs
mgoldenberg: Just making sure I understand correctly: you are asking that `MediaMetadata::content_id` and `MediaContent::id` have the same name and use `IndexedMediaContentIdKey` as their type. Is that correct?

This seems fine to me and, in fact, I even had it that way at one point! How would you feel about them both being named `content_id`? While `MediaContent::content_id` feels redundant, I think it is preferable to `MediaMetadata::id`, which seems far too general to indicate that it is actually a type of foreign key.

Or if you have another suggestion that is better than either of those, happy to hear that as well.

crates/matrix-sdk-indexeddb/src/media_store/serializer/indexed_types.rs
mgoldenberg: Yes, I agree - it is not ideal. I was trying to be consistent about the suffixes, but I actually think in this case it would make sense to break the pattern.

crates/matrix-sdk-indexeddb/src/media_store/serializer/indexed_types.rs
mgoldenberg: I am okay with this - as mentioned in the discussion above. One question though: by _"with a more generic name"_ do you mean that we should have a type synonym and use that in the media structs?

crates/matrix-sdk-indexeddb/src/media_store/types.rs
poljar: Both being named `content_id` is perfectly fine. I think the redundancy is worth the clarity in this case.

Yes, having a stronger type compared to `u64` is preferable, not sure if it should be `IndexedMediaContentIdKey` or if we're migrating towards ULIDs, but something better than `u64` sounds to me like a good idea.

crates/matrix-sdk-indexeddb/src/media_store/serializer/indexed_types.rs
poljar: Yup.

crates/matrix-sdk-indexeddb/src/transaction/mod.rs
mgoldenberg: Yes, I think it would be reasonable to use a random `u64`. I had originally been using the auto-incrementing feature of IndexedDB, and then ended up replicating it in the code because of quirks in the `indexed_db_futures` API. 

As for using a UUID, my understanding is that a UUID is in essence a `u128`, is that correct? If so, I don't know if that is a sensible choice. There is a [maximum integer value](https://docs.rs/js-sys/latest/js_sys/struct.Number.html#associatedconstant.MAX_SAFE_INTEGER) which can reliably be used across the Rust-JavaScript boundary. It is somewhere between a `u32` and a `u64`, so adding more bits won't be much help.

mgoldenberg:
> Oh, another thing. Please add a changelog entry.

So, I haven't left changelog entries on any of the pull requests in this series because none of this code is accessible outside of the crate at the moment. Have I been thinking about this correctly? I am happy to add a changelog, just want to make sure we're on the same page.

crates/matrix-sdk-indexeddb/src/transaction/mod.rs
poljar: A UUID as well as a ULID are in essence a `u128` correct. They do have standardized serialization formats.

Because of the Javascript numerical limitations you would have to serialize the UUID for storage.

mgoldenberg: Ah, I think I see what you mean. It would be serialized as a string, yes?

poljar:
> > Oh, another thing. Please add a changelog entry.
> 
> So, I haven't left changelog entries on any of the pull requests in this series because none of this code is accessible outside of the crate at the moment. Have I been thinking about this correctly? I am happy to add a changelog, just want to make sure we're on the same page.

This should improve the performance of the media access. As such it sounds like it would be a nice thing to announce to our users.

crates/matrix-sdk-indexeddb/src/transaction/mod.rs
poljar: Yup. Not sure though if this messes up your logic for the ranges.

mgoldenberg: Sounds good! And it shouldn't make a difference, as queries for media content don't assume any ordering.

crates/matrix-sdk-indexeddb/src/media_store/types.rs
mgoldenberg: - Rename `MediaContent::id`: [`1283f04b5a1137ef8d8e09d437468fea725be3e9`](https://github.com/matrix-org/matrix-rust-sdk/pull/5795/commits/1283f04b5a1137ef8d8e09d437468fea725be3e9)
- Use UUID instead of `u64`: [`c1b5acb59ac4ab61d822fa8af0361850ffac79ce`](https://github.com/matrix-org/matrix-rust-sdk/pull/5795/commits/c1b5acb59ac4ab61d822fa8af0361850ffac79ce)

crates/matrix-sdk-indexeddb/src/media_store/serializer/indexed_types.rs
mgoldenberg: Use UUID instead of `u64`: [`c1b5acb59ac4ab61d822fa8af0361850ffac79ce`](https://github.com/matrix-org/matrix-rust-sdk/pull/5795/commits/c1b5acb59ac4ab61d822fa8af0361850ffac79ce)

crates/matrix-sdk-indexeddb/src/transaction/mod.rs
mgoldenberg: Use UUID instead of `u64`: [`c1b5acb59ac4ab61d822fa8af0361850ffac79ce`](https://github.com/matrix-org/matrix-rust-sdk/pull/5795/commits/c1b5acb59ac4ab61d822fa8af0361850ffac79ce)

crates/matrix-sdk-indexeddb/src/media_store/serializer/indexed_types.rs
mgoldenberg: fixup: [`913332feb5b9ec26c3b89ec8623a8fa1c52e28b5`](https://github.com/matrix-org/matrix-rust-sdk/pull/5795/commits/913332feb5b9ec26c3b89ec8623a8fa1c52e28b5)

mgoldenberg:
@poljar: I added a changelog entry, but this is now in conflict with `main`. Mind if I rebase and force push? I can also squash the fixup commit above.

poljar:
> @poljar: I added a changelog entry, but this is now in conflict with `main`. Mind if I rebase and force push? I can also squash the fixup commit above.

Please wait with rebasing till we're dong with the review phase.

mgoldenberg:
Okay, tests don't seem to run when there is a conflict, so figured it might be worth doing it before.

poljar (commented):
Ok, the followup commits look good. Not sure if CI agrees that everything is fine, so let's rebase now to get it working.

PR #5799, opened 2025-10-23
feat(spaces): add support for MSC3230 and top level space order

stefanceriu:
This is an unstable feature but as per [MSC3230](https://github.com/matrix-org/matrix-spec-proposals/pull/3230) each space room might have an optional `m.space_order`/`org.matrix.msc3230.space_order` string field in its room account data defining the lexicographical order in which the spaces should be displayed, with spaces missing this field shown at the bottom and ordered by their room id.


poljar (approved):
Looks good, please just add a changelog entry.

PR #5801, opened 2025-10-24
feat(oauth): add flow for reciprocating a login using a QR code generated on the existing device

Johennes:
This adds the flow for reciprocating a login by generating a QR code on the logged in device as per [MSC4108](https://github.com/matrix-org/matrix-spec-proposals/pull/4108). Two required refactoring commits are also included which I hope is ok as they felt too small to justify being put into their own PR. I'm planning to add the complementary flow where the existing device scans the QR code in a separate PR.

### DIFF
--- a/crates/matrix-sdk/src/authentication/oauth/mod.rs
+++ b/crates/matrix-sdk/src/authentication/oauth/mod.rs
@@ -214,7 +214,7 @@ mod tests;
 #[cfg(feature = "e2e-encryption")]
 use self::cross_process::{CrossProcessRefreshLockGuard, CrossProcessRefreshManager};
 #[cfg(feature = "e2e-encryption")]
-use self::qrcode::{LoginWithGeneratedQrCode, LoginWithQrCode};
+use self::qrcode::{GrantLoginWithGeneratedQrCode, LoginWithGeneratedQrCode, LoginWithQrCode};
 pub use self::{
     account_management_url::{AccountManagementActionFull, AccountManagementUrlBuilder},
     auth_code_builder::{OAuthAuthCodeUrlBuilder, OAuthAuthorizationData},
@@ -364,7 +364,7 @@ impl OAuth {
         as_variant!(data, AuthData::OAuth)
     }
 
-    /// Log in using a QR code.
+    /// Log in this device using a QR code.
     ///
     /// # Arguments
     ///
@@ -378,6 +378,12 @@ impl OAuth {
         registration_data: Option<&'a ClientRegistrationData>,
     ) -> LoginWithQrCodeBuilder<'a> {
         LoginWithQrCodeBuilder { client: &self.client, registration_data }
+    }
+
+    /// Grant login to a new device using a QR code.
+    #[cfg(feature = "e2e-encryption")]
+    pub fn grant_login_with_qr_code<'a>(&'a self) -> GrantLoginWithQrCodeBuilder<'a> {
+        GrantLoginWithQrCodeBuilder { client: &self.client }
     }
 
     /// Restore or register the OAuth 2.0 client for the server with the given
@@ -1450,7 +1456,7 @@ impl<'a> LoginWithQrCodeBuilder<'a> {
     ///     ruma::serde::Raw,
     ///     Client,
     /// };
-    /// use std::io::stdin;
+    /// use std::{error::Error, io::stdin};
     /// # fn client_metadata() -> Raw<ClientMetadata> { unimplemented!() }
     /// # _ = async {
     /// // Build the client as usual.
@@ -1481,9 +1487,9 @@ impl<'a> LoginWithQrCodeBuilder<'a> {
     ///             LoginProgress::EstablishingSecureChannel(GeneratedQrProgress::QrScanned(cctx)) => {
     ///                 println!("Please enter the code displayed on your other device");
     ///                 let mut s = String::new();
-    ///                 stdin().read_line(&mut s).unwrap();
-    ///                 let check_code = s.trim().parse::<u8>().unwrap();
-    ///                 cctx.send(check_code).await.unwrap()
+    ///                 stdin().read_line(&mut s)?;
+    ///                 let check_code = s.trim().parse::<u8>()?;
+    ///                 cctx.send(check_code).await?
     ///             }
     ///             LoginProgress::WaitingForToken { user_code } => {
     ///                 println!("Please use your other device to confirm the log in {user_code}")
@@ -1491,6 +1497,7 @@ impl<'a> LoginWithQrCodeBuilder<'a> {
     ///             LoginProgress::Done => break,
     ///         }
     ///     }
+    ///     Ok::<(), Box<dyn Error + Send + Sync>>(())
     /// });
     ///
     /// // Now run the future to complete the login.
@@ -1505,6 +1512,82 @@ impl<'a> LoginWithQrCodeBuilder<'a> {
     }
 }
 
+/// Builder for QR login grant handlers.
+#[cfg(feature = "e2e-encryption")]
+#[derive(Debug)]
+pub struct GrantLoginWithQrCodeBuilder<'a> {
+    /// The underlying Matrix API client.
+    client: &'a Client,
+}
+
+#[cfg(feature = "e2e-encryption")]
+impl<'a> GrantLoginWithQrCodeBuilder<'a> {
+    /// Grant login by generating a QR code on this device to be scanned by the
+    /// new device.
+    ///
+    /// # Example
+    ///
+    /// ```no_run
+    /// use anyhow::bail;
+    /// use futures_util::StreamExt;
+    /// use matrix_sdk::{
+    ///     Client, authentication::oauth::{
+    ///         qrcode::{GeneratedQrProgress, GrantLoginProgress}
+    ///     }
+    /// };
+    /// use std::{error::Error, io::stdin};
+    /// # _ = async {
+    /// // Build the client as usual.
+    /// let client = Client::builder()
+    ///     .server_name_or_homeserver_url("matrix.org")
+    ///     .handle_refresh_tokens()
+    ///     .build()
+    ///     .await?;
+    ///
+    /// let oauth = client.oauth();
+    ///
+    /// // Subscribing to the progress is necessary since we need to capture the
+    /// // QR code, feed the checkcode back in and obtain the verification URL to
+    /// // open it in a browser so the user can consent to the new login.
+    /// let mut grant = oauth.grant_login_with_qr_code().generate();
+    /// let mut progress = grant.subscribe_to_progress();
+    ///
+    /// // Create a task which will show us the progress and allows us to receive
+    /// // and feed back data.
+    /// let task = tokio::spawn(async move {
+    ///     while let Some(state) = progress.next().await {
+    ///         match state {
+    ///             GrantLoginProgress::Starting | GrantLoginProgress::SyncingSecrets => (),
+    ///             GrantLoginProgress::EstablishingSecureChannel(GeneratedQrProgress::QrReady(qr_code_data)) => {
+    ///                 println!("Please scan the QR code on your other device: {:?}", qr_code_data);
+    ///             }
+    ///             GrantLoginProgress::EstablishingSecureChannel(GeneratedQrProgress::QrScanned(checkcode_sender)) => {
+    ///                 println!("Please enter the code displayed on your other device");
+    ///                 let mut s = String::new();
+    ///                 stdin().read_line(&mut s)?;
+    ///                 let check_code = s.trim().parse::<u8>()?;
+    ///                 checkcode_sender.send(check_code).await?;
+    ///             }
+    ///             GrantLoginProgress::WaitingForAuth { verification_uri } => {
+    ///                 println!("Please open {verification_uri} to confirm the new login")
+    ///             },
+    ///             GrantLoginProgress::Done => break,
+    ///         }
+    ///     }
+    ///     Ok::<(), Box<dyn Error + Send + Sync>>(())
+    /// });
+    ///
+    /// // Now run the future to grant the login.
+    /// grant.await?;
+    /// task.abort();
+    ///
+    /// println!("Successfully granted login");
+    /// # anyhow::Ok(()) };
+    /// ```
+    pub fn generate(self) -> GrantLoginWithGeneratedQrCode<'a> {
+        GrantLoginWithGeneratedQrCode::new(self.client)
+    }
+}
 /// A full session for the OAuth 2.0 API.
 #[derive(Debug, Clone)]
 pub struct OAuthSession {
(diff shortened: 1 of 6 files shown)


crates/matrix-sdk/src/authentication/oauth/mod.rs
```diff
@@ -1505,6 +1513,82 @@ impl<'a> LoginWithQrCodeBuilder<'a> {
     }
 }
 
+/// Builder for QR login reciprocation handlers.
+#[cfg(feature = "e2e-encryption")]
+#[derive(Debug)]
+pub struct GrantLoginWithQrCodeBuilder<'a> {
```
poljar, line 1518: Hey, I like this grant login terminology. I think it's much easier to understand compared to reciprocate.

crates/matrix-sdk/src/authentication/oauth/qrcode/reciprocate/mod.rs
```diff
@@ -0,0 +1,222 @@
...
+}
+
+impl From<SecretsBundleExportError> for Error {
+    fn from(e: SecretsBundleExportError) -> Self {
+        Self::MissingSecretsBackup(Some(e))
+    }
+}
+
+/// The progress of the reciprocation process.
+enum ReciprocateProgress {
```
poljar: This needs to be pub, no?

I think naming this GrangLoginProgress might be nicer as well.

crates/matrix-sdk/src/authentication/oauth/qrcode/reciprocate/mod.rs
```diff
@@ -0,0 +1,222 @@
...
+    let request = device::get_device::v3::Request::new(device_id.to_base64().into());
+    if client.send(request).await.is_ok() {
+        channel
+            .send_json(QrAuthMessage::LoginFailure {
+                reason: LoginFailureReason::DeviceAlreadyExists,
+                homeserver: None,
+            })
+            .await?;
+        return Err(Error::DeviceIDAlreadyInUse);
+    }
```
poljar: This sounds a bit too lenient, we should specifically check for a 404 error.

Otherwise we'll continue with the login even if we get any other form of error despite the device existing.

crates/matrix-sdk/src/authentication/oauth/qrcode/reciprocate/mod.rs
```diff
@@ -0,0 +1,222 @@
...
+        return Err(Error::Unknown(
+            "Receiving unexpected message when expecting LoginSuccess".to_owned(),
+        ));
+    };
+
+    // We check that the new device was created successfully allowing for up to 10
+    // seconds of delay. -- MSC4108 Secret sharing and device verification step
+    // 1
+    let request = device::get_device::v3::Request::new(device_id.to_base64().into());
+    let deadline = Instant::now() + Duration::from_secs(10);
```
poljar:
```suggestion
    let deadline = Instant::now() + Duration::from_secs(10);

```

crates/matrix-sdk/src/authentication/oauth/qrcode/reciprocate/mod.rs
```diff
@@ -0,0 +1,222 @@
...
+    let QrAuthMessage::LoginSuccess = message else {
+        return Err(Error::Unknown(
+            "Receiving unexpected message when expecting LoginSuccess".to_owned(),
+        ));
+    };
+
+    // We check that the new device was created successfully allowing for up to 10
+    // seconds of delay. -- MSC4108 Secret sharing and device verification step
+    // 1
+    let request = device::get_device::v3::Request::new(device_id.to_base64().into());
```
poljar: Since this is the second instance where we need this, can we add a convenience method to the `Client` for this?

Maybe even something like `fn device_exists(device_id: ...) -> Result<bool>` which handles the `404` for you so the snippet bellow becomes:

```rust
if client.device_exists(device_id).await? {
    ...
} else {
    ...
}
```

crates/matrix-sdk/src/authentication/oauth/qrcode/reciprocate/mod.rs
```diff
@@ -0,0 +1,222 @@
...
+            continue;
+        }
+        // The deadline has passed. Let's fail the login process.
+        channel
+            .send_json(QrAuthMessage::LoginFailure {
+                reason: LoginFailureReason::DeviceNotFound,
+                homeserver: None,
+            })
+            .await?;
+        return Err(Error::DeviceIDAlreadyInUse);
```
poljar: This is an if/else if/else branch with the branches flattened out.

Can we make this a proper if/else if/else to make it clear from a glance that only one of the branches will be executed at each iteration.

crates/matrix-sdk/src/authentication/oauth/qrcode/reciprocate/scanned.rs
```diff
@@ -0,0 +1,728 @@
...
+use matrix_sdk_base::crypto::types::{SecretsBundle, qr_login::QrCodeData};
+use url::Url;
+
+use super::{
+    AlmostEstablishedSecureChannel, Error, SecureChannel, export_secrets_bundle, finish_login_grant,
+};
+use crate::Client;
+
+/// A handler for reciprocating an existing login to a new device.
+pub struct ReciprocateHandler {
```
poljar: Can we reuse the grant login terminology here as well?

crates/matrix-sdk/src/authentication/oauth/qrcode/reciprocate/scanned.rs
```diff
@@ -0,0 +1,728 @@
...
+    qr_code_data: QrCodeData,
+    /// The current state of the reciprocation process.
+    state: SharedObservable<ReciprocateProgress>,
+}
+
+impl std::fmt::Debug for ReciprocateHandler {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("ReciprocateHandler")
+            .field("secrets_bundle", &"SecretsBundle")
+            .field("channel", &"Channel")
```
poljar: Is there a reason to include those two? Why not remove the fields and just use `finish_non_exhaustive()` to indicate that not all fields are printed.

crates/matrix-sdk/src/authentication/oauth/qrcode/reciprocate/scanned.rs
```diff
@@ -0,0 +1,728 @@
...
+        // display it. We emit a progress update to let the caller prompt the
+        // user to enter the checkcode and feed it back to us.
+        // -- MSC4108 Secure channel setup step 6
+        self.state.set(ReciprocateProgress::WaitingForCheckCode);
+
+        Ok(())
+    }
+
+    /// Confirm the check code entered by the user and finish the reciprocation.
+    pub async fn confirm_check_code_and_finish(&mut self, check_code: u8) -> Result<(), Error> {
```
poljar: Is there a reason for this API to be inconsistent with your previous PR?

Why do we have two methods that need to be called instead of using a check-code sender like in the other methods.

Having two steps here means that you need to account for the channel being used up, or `Poisoned`. It also means that you need to clone the QR code data so you have always access to it.

If we reuse the patterns we have for the QR login we would have a named future that allows you to get the QR code data and subscribe and once you await all this data gets moved into the future and you don't need to have the `Poisoned` state nor the clone of the QR code data. I think you could even avoid the clone of the `SecretsBundle`.

crates/matrix-sdk/src/authentication/oauth/qrcode/reciprocate/mod.rs
```diff
@@ -0,0 +1,222 @@
...
+            .as_str(),
+    )
+    .map_err(|_| Error::UnableToCreateDevice)?;
+    state.set(ReciprocateProgress::WaitingForAuth { verification_uri }.into());
+
+    // We send the new device the m.login.protocol_accepted message to let it know
+    // that the consent process is in progress.
+    // -- MSC4108 OAuth 2.0 login step 4 continued
+    let message = QrAuthMessage::LoginProtocolAccepted;
+    channel.send_json(&message).await?;
```
poljar: Don't we want to set the `WaitingForAuth` state only after we notified the new device that it can poll the server for the access token?

I guess it doesn't matter much but it sounds like it's more correct to first let the device know and only then let the user open the verification URI.

crates/matrix-sdk/src/authentication/oauth/qrcode/reciprocate/mod.rs
```diff
@@ -0,0 +1,222 @@
...
+    // 1
+    let request = device::get_device::v3::Request::new(device_id.to_base64().into());
+    let deadline = Instant::now() + Duration::from_secs(10);
+    loop {
+        if client.send(request.clone()).await.is_ok() {
+            break;
+        }
+        // If the deadline hasn't yet passed, give it some time and retry the request.
+        if Instant::now() < deadline {
+            tokio::time::sleep(Duration::from_secs(1)).await;
```
poljar: That sleep duration might be a bit heavy, what do you think about 500ms?

crates/matrix-sdk/src/authentication/oauth/qrcode/reciprocate/mod.rs
Johennes: Hm, the MSC lists it the other way around.

> If no existing device was found then the existing device opens the `verification_uri_complete` [...]
> 
> The existing device then sends an acknowledgement message to let the other device know that the consent process is in progress:
> 
> *Existing device => New device via secure channel*
> 
> ```json
> {
>     "type": "m.login.protocol_accepted"
> }
> ```

I'm not sure if that was strictly intentional though.

crates/matrix-sdk/src/authentication/oauth/qrcode/reciprocate/mod.rs
Johennes: Good point. Fixed in 9cb7ba175ae6b1700be8e3d73f276ad60fcb14ab.

crates/matrix-sdk/src/authentication/oauth/qrcode/reciprocate/mod.rs
Johennes: Yes, good idea. Fixed in 9cb7ba175ae6b1700be8e3d73f276ad60fcb14ab.

crates/matrix-sdk/src/authentication/oauth/qrcode/reciprocate/mod.rs
Johennes: Fixed in 93ec5995219c27d824a64178ac35fdda1d3c1708.

crates/matrix-sdk/src/authentication/oauth/qrcode/reciprocate/scanned.rs
Johennes: No, I guess these don't provide a lot of value in the output. Have removed them with ac28865e5b6538793d6e99cd34c4deacc2f54ffe.

crates/matrix-sdk/src/authentication/oauth/qrcode/reciprocate/mod.rs
Johennes: Sounds good. I wasn't really sure what interval to pick to be honest.

Tweaked in 0503388c29c4c79861fea575a4ecd1bb93db4c8c.

crates/matrix-sdk/src/authentication/oauth/mod.rs
Johennes: Oh, thank God I'm not the only one. 😅 I've only used this because I thought it was an established terminology already. Have renamed things with f9fa7f9a0e7133a9e1e84fcbaf71f01d1f0c4b1f.

crates/matrix-sdk/src/authentication/oauth/qrcode/reciprocate/mod.rs
Johennes: Ah, you're right. I missed this. 🤦‍♂️ Fixed in e40b10fb938fe12ff4b0ad85944f05345b5a0bdc.

Naming fixed in f9fa7f9a0e7133a9e1e84fcbaf71f01d1f0c4b1f.

crates/matrix-sdk/src/authentication/oauth/qrcode/reciprocate/mod.rs
Johennes: Have applied this as part of 93ec5995219c27d824a64178ac35fdda1d3c1708.

crates/matrix-sdk/src/authentication/oauth/qrcode/reciprocate/scanned.rs
Johennes: Yes, fixed in f9fa7f9a0e7133a9e1e84fcbaf71f01d1f0c4b1f.

crates/matrix-sdk/src/authentication/oauth/qrcode/reciprocate/scanned.rs
Johennes: Hm, I think you're making a convincing case for why the other / existing API shape is better. Let me try to swap this around.

crates/matrix-sdk/src/authentication/oauth/qrcode/reciprocate/mod.rs
poljar: Alright, let's leave it as is. Shouldn't matter one-way or the other as things happen at the sameish time.

crates/matrix-sdk/src/authentication/oauth/qrcode/reciprocate/scanned.rs
Johennes: Have swapped things around and this indeed feels much simpler. ca26d0b35d1608b43b0cfd2dc52ad7667e6118b2 and 645905b4324e1dda438fd083c00f50a6ab3b86f2 are two preparatory commits and c2d655474cb111e409373dc83912e6343561420c contains the bulk of the change. Unfortunately, that last one wasn't really feasible to break down further into smaller commits.

crates/matrix-sdk/src/authentication/oauth/mod.rs
```diff
@@ -1548,17 +1546,27 @@ impl<'a> GrantLoginWithQrCodeBuilder<'a> {
...
-    ///             GrantLoginProgress::Created | GrantLoginProgress::WaitingForCheckCode | GrantLoginProgress::SyncingSecrets => (),
+    ///             GrantLoginProgress::Starting | GrantLoginProgress::SyncingSecrets => (),
+    ///             GrantLoginProgress::EstablishingSecureChannel(GeneratedQrProgress::QrReady(qr_code_data)) => {
+    ///                 println!("Please scan the QR code on your other device: {:?}", qr_code_data);
+    ///             }
+    ///             GrantLoginProgress::EstablishingSecureChannel(GeneratedQrProgress::QrScanned(checkcode_sender)) => {
+    ///                 println!("Please enter the code displayed on your other device");
+    ///                 let mut s = String::new();
+    ///                 stdin().read_line(&mut s).unwrap();
+    ///                 let check_code = s.trim().parse::<u8>().unwrap();
```
poljar: Those uwnraps could use a `?` since we're using `anyhow` in the example.

Johennes: Good point. Have fixed it here and in the other example with 5e8798fb00031253fa424824159c7a8d4e47032d.

Johennes:
The history is cleaned up now. Thanks for the review!

PR #5806, opened 2025-10-28
feat(ffi): Add Room::mark_as_fully_read_unchecked

jmartinesp:
## Changes:

- Add `ffi::Room::mark_as_fully_read_unchecked`: this method shouldn't be widely used, but it's useful when we want to mark the room as fully read when leaving it and at the same time we have to destroy the room and timeline instances immediately so their in-memory cache is cleared.
- Add `ffi::Timeline::latest_event_id`: It will allow us to fetch the latest event id coming from the SDK instead of deciding which one to use in the clients, which could be altered by filters, post-processing, etc.
- Make sure `ui::Timeline::latest_event` returns the latest event and not the latest item if it's an event, so it matches the logic of `latest_event_id` and other methods that take the latest event in the SDK.

Signed-off-by:

### DIFF
--- a/crates/matrix-sdk-ui/CHANGELOG.md
+++ b/crates/matrix-sdk-ui/CHANGELOG.md
@@ -12,6 +12,8 @@ All notable changes to this project will be documented in this file.
 
 ### Refactor
 
+- `Timeline::latest_event` will return the latest event in the timeline, not the latest item of the timeline if it's
+  an event.
 - `TimelineFocusKind::Event` can now handle both the existing event pagination and thread pagination if the focused 
   event is part of a thread ([#5678](https://github.com/matrix-org/matrix-rust-sdk/pull/5678)).
 
--- a/bindings/matrix-sdk-ffi/CHANGELOG.md
+++ b/bindings/matrix-sdk-ffi/CHANGELOG.md
@@ -34,6 +34,8 @@ All notable changes to this project will be documented in this file.
 
 ### Features:
 
+- Add `Room::mark_as_fully_read_unchecked` so clients can mark a room as read without needing a `Timeline` instance. Note this method is not recommended as it can potentially cause incorrect read receipts, but it can needed in certain cases.
+- Add `Timeline::latest_event_id` to be able to fetch the event id of the latest event of the timeline.
 - Add `Room::load_or_fetch_event` so we can get a `TimelineEvent` given its event id ([#5678](https://github.com/matrix-org/matrix-rust-sdk/pull/5678)).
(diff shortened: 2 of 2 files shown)


jmartinesp:
Some context for why we may need this: on Element X Android we both want to mark the room as fully read and clear the timeline cache as soon as we leave the room screen. However: 

1. Sending the read receipt is an async operation and we need the timeline to be around to do it.
2. If we wait for the read receipt to be sent before destroying the timeline reference so its cache is cleared, in bad connectivity scenarios this could potentially take several seconds. If the user then tries to open the recently left room before this is finished, the newly instantiated timeline will load the previously cached items and instead of loading the initial (20?) items it will load hundreds or thousands of them if you paginated back long enough, making the timeline loading very slow.
3. There is no reliable way to do 'as soon as you start sending the read receipt you can now destroy the timeline reference'. The best approach I could find was using delays (prone to breaking) or, as done here, split the logic of what would usually be done inside the SDK.

bindings/matrix-sdk-ffi/src/timeline/mod.rs
```diff
@@ -365,6 +365,11 @@ impl Timeline {
         Ok(())
     }
 
+    /// Returns the latest [`EventId`] of the timeline.
```
poljar: The event ID of the timeline? This doesn't sound sensible to me :sweat_smile:

jmartinesp: Yeah, I could have documented it better 😅 . https://github.com/matrix-org/matrix-rust-sdk/pull/5806/commits/7df9e7b4e60d1b3f74084d02302526cc926d01b8

crates/matrix-sdk-ui/CHANGELOG.md
```diff
@@ -14,6 +14,8 @@ All notable changes to this project will be documented in this file.
 
 - `TimelineFocusKind::Event` can now handle both the existing event pagination and thread pagination if the focused 
   event is part of a thread ([#5678](https://github.com/matrix-org/matrix-rust-sdk/pull/5678)).
+- `Timeline::latest_event` will return the latest event in the timeline, not the latest item of the timeline if it's 
+  an event.
```
poljar: New entries should be added to the top.

poljar (approved):
Looks good, except changelog entries should be at the top.

Approving ahead of time.

crates/matrix-sdk-ui/CHANGELOG.md
jmartinesp: Ah, I forgot about that. Will change it.

PR #5807, opened 2025-10-28
crypto: use a new withheld code when history is marked as "not shareable"

richvdh:
MSC4268 now defines a new withheld code for use when the session is not marked as `shared_history`: start using it, and add a test to make sure it is sent correctly.

part of https://github.com/element-hq/element-meta/issues/2876

### DIFF
--- a/testing/matrix-sdk-integration-testing/src/tests/e2ee/shared_history.rs
+++ b/testing/matrix-sdk-integration-testing/src/tests/e2ee/shared_history.rs
@@ -5,19 +5,26 @@ use assert_matches2::{assert_let, assert_matches};
 use assign::assign;
 use futures::{FutureExt, StreamExt, future, pin_mut};
 use matrix_sdk::{
-    assert_decrypted_message_eq, assert_next_with_timeout,
+    Room, assert_decrypted_message_eq, assert_next_with_timeout,
     deserialized_responses::TimelineEventKind,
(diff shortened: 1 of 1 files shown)


crates/matrix-sdk-crypto/src/types/events/room_key_withheld.rs
```diff
@@ -84,7 +85,9 @@ macro_rules! construct_withheld_content {
                         .into(),
                 ))
             }
-            _ => unreachable!("Can't create an unknown withheld code content"),
+            WithheldCode::_Custom(_) => {
```
richvdh, line 88: I've removed the wildcard pattern in a few enums, so that it's harder to miss places when adding new withheld codes.

crates/matrix-sdk-common/src/deserialized_responses.rs
```diff
@@ -1200,6 +1200,13 @@ pub enum WithheldCode {
     #[ruma_enum(rename = "m.no_olm")]
     NoOlm,
 
+    /// Normally used when sharing history, per [MSC4268][MSC4268]: indicates
```
poljar: You don't need `[MSC4268]` twice, no?

richvdh: indeed.

PR #5808, opened 2025-10-28
feat(ffi): Upgrade UniFFI to `v0.30.0`

jmartinesp:
Upgrade UniFFI to `v0.30.0`. There is a matching branch in [Complement Crypto](https://github.com/matrix-org/complement-crypto/pull/214) to workaround the issues caused by the upgrade.

Some parts are based on #1433 .

Signed-off-by:

### DIFF
--- a/bindings/matrix-sdk-ffi/CHANGELOG.md
+++ b/bindings/matrix-sdk-ffi/CHANGELOG.md
@@ -8,10 +8,10 @@ All notable changes to this project will be documented in this file.
 
 ### Breaking changes:
 
+- UniFFI was upgraded to `v0.30.0` ([#5808](https://github.com/matrix-org/matrix-rust-sdk/pull/5808)).
 - The `waveform` parameter in `Timeline::send_voice_message` format changed to a list of `f32`
   between 0 and 1.
   ([#5732](https://github.com/matrix-org/matrix-rust-sdk/pull/5732))
-
 - The `normalized_power_level` field has been removed from the `RoomMember`
   struct.
   ([#5635](https://github.com/matrix-org/matrix-rust-sdk/pull/5635))


bindings/matrix-sdk-ffi/uniffi.toml
```diff
@@ -1,4 +1,14 @@
...
+
+[bindings.typescript]
+[bindings.typescript.customTypes.Date64]
+typeName = "Date"
+intoCustom = """((v: FfiType) => {
+    return new Date(Number(v));
+})({})"""
+fromCustom = """((v: TsType) => {
+    return BigInt(v.getTime());
+})({})"""
```
poljar: Missing newline at the end.

bindings/matrix-sdk-ffi/uniffi.toml
```diff
@@ -1,4 +1,14 @@
 [bindings.kotlin]
 package_name = "org.matrix.rustcomponents.sdk"
 cdylib_name = "matrix_sdk_ffi"
-android_cleaner = true
\ No newline at end of file
+android_cleaner = true
+
+[bindings.typescript]
+[bindings.typescript.customTypes.Date64]
```
poljar: Could we add a comment on why this is needed and what it does?

jmartinesp: To be honest, I'm not fully sure this is needed. It was added in https://github.com/matrix-org/matrix-rust-sdk/pull/5189, so I assumed it was needed for the RN bindings.

poljar: I guess we'll remove it from this PR and let follow up PRs merge and document this.

jmartinesp: I removed it. If those mappings are needed they can be added in a separate PR.

PR #5810, opened 2025-10-29
chore: Fix a comment

poljar:
I don't know, seems like a copy paste issue or a joke I don't understand.

### DIFF
--- a/crates/matrix-sdk/src/event_cache/room/mod.rs
+++ b/crates/matrix-sdk/src/event_cache/room/mod.rs
@@ -2302,7 +2302,7 @@ mod timed_tests {
 
         let event_cache = client.event_cache();
 
-        // Don't forget to subscribe and like^W enable storage!
+        // Don't forget to subscribe and like.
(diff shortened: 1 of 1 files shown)


poljar:
> I believe it's a joke based on YouTube's culture where video makers constantly ask the viewers to “like & subscribe!”. It's a funny joke, but I'm okay with the change.

That part I understand, but what about the part that I removed?

bnjbvr:
> That part I understand, but what about the part that I removed?

At some point, enabling the persistent storage was a manual action, so there would be another call to `client.event_cache().enable_storage();` as well. This has disappeared when persistent storage has been enabled by default, but the comment hadn't been updated. Doh!

poljar:
I imagined that it had to do with the manual storage enabling, but I couldn't figure out if the `^W` was some clever way to say "with". Perhaps the `^` was supposed to be the logical `and` operation, but what would the `W` then be?

Doesn't matter as the storage enabling isn't done manually anymore.

PR #5813, opened 2025-10-29
chore: Convert a Note to a NOTE

poljar:
All uppercase is the correct convention and some editors even highlight things if the correct convention is used.


Hywan (approved):
Looks good to me, thanks!

PR #5817, opened 2025-10-30
event-cache: Allow room events to be loaded of a certain type

poljar:
Part of #3872.

This PR modifies the `get_room_events()` method to allow a subset of events to be fetched from storage.

Once the redecryption logic moves into the event cache we'll need to load only `m.room.encrypted` events.

### DIFF
--- a/crates/matrix-sdk-base/src/event_cache/store/integration_tests.rs
+++ b/crates/matrix-sdk-base/src/event_cache/store/integration_tests.rs
@@ -14,14 +14,17 @@
 
 //! Trait and macro of integration tests for `EventCacheStore` implementations.
 
-use std::{collections::BTreeMap, sync::Arc};
+use std::{
+    collections::{BTreeMap, BTreeSet},
+    sync::Arc,
+};
 
 use assert_matches::assert_matches;
 use assert_matches2::assert_let;
 use matrix_sdk_common::{
     deserialized_responses::{
         AlgorithmInfo, DecryptedRoomEvent, EncryptionInfo, TimelineEvent, TimelineEventKind,
-        VerificationState,
+        UnableToDecryptInfo, UnableToDecryptReason, VerificationState,
     },
     linked_chunk::{
         ChunkContent, ChunkIdentifier as CId, LinkedChunkId, Position, Update, lazy_loader,
@@ -47,6 +50,24 @@ use crate::event_cache::{Gap, store::DEFAULT_CHUNK_CAPACITY};
 /// Keep in sync with [`check_test_event`].
 pub fn make_test_event(room_id: &RoomId, content: &str) -> TimelineEvent {
     make_test_event_with_event_id(room_id, content, None)
+}
+
+/// Create a `m.room.encrypted` test event with all data filled, for testing
+/// that linked chunk correctly stores event data for encrypted events.
+pub fn make_encrypted_test_event(room_id: &RoomId, session_id: &str) -> TimelineEvent {
+    let device_id = "DEVICEID";
+    let builder = EventFactory::new()
+        .encrypted("", "curve_key", device_id, session_id)
+        .room(room_id)
+        .sender(*ALICE);
+
+    let event = builder.into_raw();
+    let utd_info = UnableToDecryptInfo {
+        session_id: Some(session_id.to_owned()),
+        reason: UnableToDecryptReason::MissingMegolmSession { withheld_code: None },
+    };
+
+    TimelineEvent::from_utd(event, utd_info)
 }
 
 /// Same as [`make_test_event`], with an extra event id.
@@ -143,6 +164,9 @@ pub trait EventCacheStoreIntegrationTests {
 
     /// Test that getting all events in a room works as expected.
     async fn test_get_room_events(&self);
+
+    /// Test that getting events in a room of a certain type works as expected.
+    async fn test_get_room_events_filtered(&self);
 
     /// Test that saving an event works as expected.
     async fn test_save_event(&self);
@@ -962,7 +986,10 @@ impl EventCacheStoreIntegrationTests for DynEventCacheStore {
         .unwrap();
 
         // Now let's find the events.
-        let events = self.get_room_events(room_id).await.expect("failed to query for room events");
+        let events = self
+            .get_room_events(room_id, None, None)
+            .await
+            .expect("failed to query for room events");
 
         assert_eq!(events.len(), 2);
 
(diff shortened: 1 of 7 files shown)


crates/matrix-sdk-sqlite/migrations/event_cache_store/012_store_event_type.sql
```diff
@@ -0,0 +1,57 @@
...
+
+DELETE from linked_chunks;
+DELETE from event_chunks; -- should be done by cascading
+DELETE from gap_chunks; -- should be done by cascading
+DELETE from events;
+
+DROP TABLE events;
+
+-- Events and their content.
+CREATE TABLE "events" (
```
Hywan, line 15: Can't we use [an `ALTER TABLE` command](https://sqlite.org/lang_altertable.html) instead of dropping and re-creating the table entirely?

crates/matrix-sdk-sqlite/migrations/event_cache_store/012_store_event_type.sql
```diff
@@ -0,0 +1,57 @@
...
+    "rel_type" BLOB,
+
+    -- Primary key is the event ID.
+    PRIMARY KEY (event_id)
+)
+WITHOUT ROWID;
+
+-- Add an index to speed up queries that look for related events in a room.
+CREATE INDEX "relates_to_idx"
+    ON "events" ("room_id", "relates_to");
```
Hywan, line 48: This index already exists, crated in `009_related_event_index.sql`? Why re-creating it here?

crates/matrix-sdk-sqlite/migrations/event_cache_store/012_store_event_type.sql
```diff
@@ -0,0 +1,57 @@
...
+
+-- Add an index to speed up queries that look for related events in a room, with an additional
+-- filter.
+CREATE INDEX "relates_to_rel_type_idx"
+    ON "events" ("room_id", "relates_to", "rel_type");
```
Hywan, line 53: Same: this index already exists, crated in `009_related_event_index.sql`? Why re-creating it here?

crates/matrix-sdk-base/src/event_cache/store/integration_tests.rs
```diff
@@ -49,6 +52,24 @@ pub fn make_test_event(room_id: &RoomId, content: &str) -> TimelineEvent {
     make_test_event_with_event_id(room_id, content, None)
 }
 
+/// Create a m.room.encrypted test  event with all data filled, for testing that
```
Hywan: Double space + missing markdown.

```diff
- /// Create a m.room.encrypted test  event
+ /// Create a `m.room.encrypted` test event
```

crates/matrix-sdk-base/src/event_cache/store/memory_store.rs
```diff
@@ -217,11 +217,28 @@ impl EventCacheStore for MemoryStore {
...
+            .map(|(event, _pos)| event.clone())
+            .filter(|e| match (event_type, session_id) {
+                (None, None) => true,
+                (None, Some(session_id)) => e.kind.session_id() == Some(session_id),
+                (Some(event_type), None) => e.kind.event_type().as_deref() == Some(event_type),
+                (Some(event_type), Some(session_id)) => {
+                    e.kind.session_id() == Some(session_id)
+                        && e.kind.event_type().as_deref() == Some(event_type)
+                }
+            })
```
Hywan: Funny, I would have written that in 2 `filter`s:

```rust
.map(|(event, _pos)| event.clone())
.filter(|event| event_type.is_some_and(|ty| ty == event.kind.event_type().as_deref()))
.filter(|event| session_id.is_some_and(|s| s == event.kind.session_id()))
```

~~I'm lazy to compare the generated code, but I suspect it should be smaller.~~

Turns out, I wasn't lazy: https://godbolt.org/z/q6b1r5565. The 2 `filter`s approach generates less code and less jumps.

crates/matrix-sdk-indexeddb/src/event_cache_store/mod.rs
```diff
@@ -479,14 +479,34 @@ impl EventCacheStore for IndexeddbEventCacheStore {
...
+                        (None, None) => true,
+                        (None, Some(session_id)) => e.kind.session_id() == Some(session_id),
+                        (Some(event_type), None) => {
+                            e.kind.event_type().as_deref() == Some(event_type)
+                        }
+                        (Some(event_type), Some(session_id)) => {
+                            e.kind.session_id() == Some(session_id)
+                                && e.kind.event_type().as_deref() == Some(event_type)
+                        }
+                    })
```
Hywan: Same as above.

Hywan (changes requested):
Great! I left a couple of comments, I think we must resolve them before merging :-).

crates/matrix-sdk-sqlite/migrations/event_cache_store/012_store_event_type.sql
poljar: How would I get to the event ID while inside a `ALTER TABLE` call?

crates/matrix-sdk-sqlite/migrations/event_cache_store/012_store_event_type.sql
poljar: Because it got deleted while the table got dropped.

crates/matrix-sdk-sqlite/migrations/event_cache_store/012_store_event_type.sql
poljar: Same argument, table got deleted, this deletes the index.

crates/matrix-sdk-base/src/event_cache/store/memory_store.rs
poljar: Hmm, I don't mind one way or the other, but the closures you posted don't do the same thing as the one closure I wrote originally. Tests start to fail, confirming this as well.

The equivalent closure would be as follows:

```rust
if let Some(event_type) = event_type {
    Some(event_type) == event.kind.event_type().as_deref()
} else {
    true
}
```

That is, only if the passed in event type is `Some` do we do the comparison, otherwise we return `true`. On the other hand `is_some_and()` returns false on the `None` case. This likely avoids a bit of jumping in the generated code.



Taking a look at the generated code, now it seems that the two filters produce slightly more code: https://godbolt.org/z/917be19e3.

Looking at some benchmarks, these two approaches, once they behave the same, produce the same performance:

```
Timer precision: 20 ns
example            fastest       │ slowest       │ median        │ mean          │ samples │ iters
├─ test_one_bench                │               │               │               │         │
│  ├─ 100          19.06 ns      │ 43.44 ns      │ 21.44 ns      │ 21.83 ns      │ 100     │ 12800
│  ├─ 200          51.33 ns      │ 97.42 ns      │ 52.42 ns      │ 52.85 ns      │ 100     │ 6400
│  ├─ 400          89.11 ns      │ 115 ns        │ 89.77 ns      │ 90.02 ns      │ 100     │ 3200
│  ├─ 8000         1.484 µs      │ 1.684 µs      │ 1.489 µs      │ 1.491 µs      │ 100     │ 200
│  ├─ 16000        2.939 µs      │ 3.059 µs      │ 2.949 µs      │ 2.956 µs      │ 100     │ 100
│  ╰─ 32000        5.849 µs      │ 6.159 µs      │ 5.879 µs      │ 5.89 µs       │ 100     │ 100
╰─ test_two_bench                │               │               │               │         │
   ├─ 100          19.22 ns      │ 20.16 ns      │ 19.45 ns      │ 19.44 ns      │ 100     │ 12800
   ├─ 200          51.8 ns       │ 58.05 ns      │ 52.27 ns      │ 52.32 ns      │ 100     │ 6400
   ├─ 400          89.14 ns      │ 100.7 ns      │ 89.77 ns      │ 89.91 ns      │ 100     │ 3200
   ├─ 8000         1.484 µs      │ 4.299 µs      │ 1.494 µs      │ 1.532 µs      │ 100     │ 200
   ├─ 16000        2.949 µs      │ 3.069 µs      │ 2.959 µs      │ 2.966 µs      │ 100     │ 100
   ╰─ 32000        5.859 µs      │ 6.279 µs      │ 5.898 µs      │ 5.901 µs      │ 100     │ 100
```

For completion sake, here's the benchmakred code as well as the benchmark.

```rust
use arbitrary::Arbitrary;

#[derive(Debug, Arbitrary)]
pub struct S {
    a: Option<u32>,
    b: Option<u32>,
}

pub fn test_one(vec: Vec<S>, e_a: Option<u32>, e_b: Option<u32>) -> Vec<S> {
    vec.into_iter()
        .filter(|S { a, b }| match (e_a, e_b) {
            (None, None) => true,
            (Some(e_a), None) => *a == Some(e_a),
            (None, Some(e_b)) => *b == Some(e_b),
            (Some(e_a), Some(e_b)) => *a == Some(e_a) && *b == Some(e_b),
        })
        .collect()
}

pub fn test_two(vec: Vec<S>, e_a: Option<u32>, e_b: Option<u32>) -> Vec<S> {
    vec.into_iter()
        .filter(|S { a, .. }| {
            if let Some(ea) = e_a {
                Some(ea) == *a
            } else {
                true
            }
        })
        .filter(|S { b, .. }| {
            if let Some(ea) = e_b {
                Some(ea) == *b
            } else {
                true
            }
        })
        .collect()
}
```

```rust
use arbitrary::{Arbitrary, Unstructured};
use divan::Bencher;
use foo::{S, test_one, test_two};

fn main() {
    // Run registered benchmarks.
    divan::main();
}

fn generate_inputs(n: usize) -> Vec<S> {
    let mut rng = fastrand::Rng::with_seed(42);
 
    (0..n)
        .map(|_| {
            let mut bytes = vec![0u8; 64];

            rng.fill(&mut bytes[..]);
            let mut u = Unstructured::new(&bytes);
            S::arbitrary(&mut u).unwrap()
        })
        .collect()
}

#[divan::bench(args = [100, 200, 400, 8000, 16000, 32000])]
fn test_one_bench(bencher: Bencher, n: usize) {
    let e_a = None;
    let e_b = Some(1);

    bencher
        .with_inputs(|| generate_inputs(n))
        .bench_values(|input| test_one(input, e_a, e_b));
}

#[divan::bench(args = [100, 200, 400, 8000, 16000, 32000])]
fn test_two_bench(bencher: Bencher, n: usize) {
    let e_a = None;
    let e_b = Some(1);

    bencher
        .with_inputs(|| generate_inputs(n))
        .bench_values(|input| test_two(input, e_a, e_b));
}
```

crates/matrix-sdk-sqlite/migrations/event_cache_store/012_store_event_type.sql
Hywan: Oh, neat!

crates/matrix-sdk-base/src/event_cache/store/memory_store.rs
Hywan: Alright! Thanks for checking.

Hywan (approved):

crates/matrix-sdk-base/src/event_cache/store/memory_store.rs
poljar: I noticed that there's `Option::is_none_or()` we can switch to that if you prefer the two filter approaches.

The performance is the same.

crates/matrix-sdk-base/src/event_cache/store/integration_tests.rs
poljar: Fixed.

crates/matrix-sdk-base/src/event_cache/store/memory_store.rs
poljar: Changed it to `is_none_or()` as this is in line with the original request in the comment.

PR #5819, opened 2025-10-30
crypto: add new `CryptoStore` method `get_withheld_sessions_by_room_id`

richvdh:
Add a new method to `CryptoStore` which allows retrieving all the withheld sessions in a given room.

Unfortunately, for indexeddb, that means a data migration, because currently the withheld sessions are indexed by `(session_id, room)` rather than `(room, session_id)`.

This is groundwork for https://github.com/element-hq/element-meta/issues/2876.

### DIFF
--- a/crates/matrix-sdk-crypto/CHANGELOG.md
+++ b/crates/matrix-sdk-crypto/CHANGELOG.md
@@ -8,6 +8,8 @@ All notable changes to this project will be documented in this file.
 
 ### Features
 
+- Expose new method `CryptoStore::get_withheld_sessions_by_room_id`.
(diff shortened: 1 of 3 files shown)


crates/matrix-sdk-sqlite/migrations/crypto_store/012_withheld_code_by_room.sql
```diff
@@ -0,0 +1,2 @@
+CREATE INDEX "direct_withheld_info_room_id_idx"
+    ON "direct_withheld_info" ("room_id");
```
poljar, line 2: Reminder to myself, this will be a merge conflict with #5817.

PR #5822, opened 2025-10-31
Don't send a auth header when fetching the versions during a token refresh

### LINKED ISSUE #5816: Calling `/versions` no longer works when the access token requires a refresh (via MAS).
We started noticing an issue recently on Element X (both platforms) which goes something along these lines:
- Launch the app when both the access token and the cached `ServerInfo` (which contains `/versions` data) have expired.
- The SDK attempts to make a `/versions` request.
- As of #5789, Ruma now includes the access token on `/versions` meaning that the following error is returned
    - `Api(Server(ClientApi(Error { status_code: 401, body: Standard { kind: UnknownToken { soft_logout: false }, message: "Token is not active" } })))`
- The SDK sees that error and [attempts to refresh the access token](https://github.com/matrix-org/matrix-rust-sdk/blob/f78f1795ebf243347ea441ef136faa05cf0e1172/crates/matrix-sdk/src/client/futures.rs#L112).
- Refreshing the access token via MAS requires knowledge of the authentication server so a [`get_authorization_server_metadata`](https://github.com/matrix-org/matrix-rust-sdk/blob/e8fb133cbf7b182098c6046a7de5b7360a906aaa/crates/matrix-sdk/src/authentication/oauth/mod.rs#L502-L503) request is sent.
- A `get_authorization_server_metadata` request requires knowledge of the supported `/versions` in order to [build the correct path](https://github.com/ruma/ruma/blob/15307f0eb361d7e04cfa33602fdffbf10c54a173/crates/ruma-client-api/src/discovery/get_authorization_server_metadata.rs#L29-L30).
- Bad things happen™️

In practice we've seen this manifest in 2 different ways:
- The session is restored and navigation throughout the app works, but the SDK only ever sends `/versions` requests no matter what you attempts to do.
- Whilst setting up the session, [`SyncServiceBuilder::finish`](https://github.com/matrix-org/matrix-rust-sdk/blob/b1c28f4bc1788ebf2633b5f5486513e2540457d9/bindings/matrix-sdk-ffi/src/sync_service.rs#L137-L143) throws the following error which (on EXI at least) we handle with a fatal error
    - `Generic(msg: "the server returned an error: [401 / M_UNKNOWN_TOKEN] Token is not active", details: Optional("RoomList(SlidingSync(Http(Api(Server(ClientApi(Error { status_code: 401, body: Standard { kind: UnknownToken { soft_logout: false }, message: \"Token is not active\" } }))))))"))`

poljar:
This closes #5816.

### DIFF
--- a/crates/matrix-sdk/src/client/mod.rs
+++ b/crates/matrix-sdk/src/client/mod.rs
@@ -1956,7 +1956,8 @@ impl Client {
         request_config: Option<RequestConfig>,
     ) -> HttpResult<get_supported_versions::Response> {
         let server_versions = self
-            .send_inner(get_supported_versions::Request::new(), request_config, Default::default())
+            .send(get_supported_versions::Request::new())
+            .with_request_config(request_config)
             .await?;
 
         Ok(server_versions)


crates/matrix-sdk/src/client/mod.rs
```diff
@@ -1956,7 +1956,8 @@ impl Client {
         request_config: Option<RequestConfig>,
     ) -> HttpResult<get_supported_versions::Response> {
         let server_versions = self
-            .send_inner(get_supported_versions::Request::new(), request_config, Default::default())
+            .send(get_supported_versions::Request::new())
+            .with_request_config(request_config)
```
Hywan, line 1960: What does it fix? The code looks pretty similar.

poljar: Like the commit message says, this enables token refresh for this request.

`Client::send_inner()` does not contain the token refresh logic, while `Client::send()` does.

PR #5826, opened 2025-11-04
xtask: Workaround UniFFI's noHandle generation for Swift.

pixlwave:
Replaces https://github.com/element-hq/matrix-rust-components-swift/commit/3f0075fb8acaf202755512dfbafc28024e5125b2, not really sure why I didn't just do it here in the first place…

It shouldn't need to stay for long, there's already a [PR](https://github.com/mozilla/uniffi-rs/pull/2720) open but this unblocks Swift SDK users for now.

### DIFF
--- a/xtask/src/swift.rs
+++ b/xtask/src/swift.rs
@@ -250,6 +250,7 @@ fn build_xcframework(
     consolidate_modulemap_files(&generated_dir, &headers_module_dir)?;
 
     move_files("swift", &generated_dir, &swift_dir)?;
+    apply_uniffi_mocks_workaround(&swift_dir)?;
 
     println!("-- Generating MatrixSDKFFI.xcframework framework");
     let xcframework_path = generated_dir.join("MatrixSDKFFI.xcframework");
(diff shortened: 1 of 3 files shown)


xtask/src/swift.rs
```diff
@@ -431,3 +432,29 @@ fn consolidate_modulemap_files(source: &Utf8Path, destination: &Utf8Path) -> Res
...
+        let entry = entry?;
+        if entry.file_type()?.is_file() {
+            let path = entry.path();
+            if path.extension() == Some("swift") {
+                let mut contents = std::fs::read_to_string(path)?;
+                let new_contents = regex.replace_all(
+                    &contents,
+                    r#"$1        guard handle != 0 else { return }
+$2$3"#,
+                );
```
stefanceriu: This is super fugly but it's gonna be gone in a week and ain't nobody got time to faff with it.

stefanceriu (approved):
👌

xtask/src/swift.rs
pixlwave: [Updated](https://github.com/matrix-org/matrix-rust-sdk/compare/2b1d072a6a9f936010868b49f8e28f70378f5b84..2919244b8b4d811e60f2b78d36b34ef7c91b7575)

PR #5827, opened 2025-11-04
refactor: Add profile cache for handle_remote_event

jmartinesp:
This should slightly improve the performance for loading timeline items since we might skip loading the sender profiles for some/most items.

In my local tests on Android, this can save up to 50ms for each timeline initialization, although the benchmark only shows a 1.7% improvement 🥲 . That said, it also says the the function that populates the cache is not called when it definitely is, 
so it doesn't seem too reliable.

Signed-off-by:


Hywan (commented):
Thank you for the contribution!
How do you manage the user profile invalidation? If a user has updated its profile, how is it updated from the cache (the `HashMap`)?

jmartinesp:
That's the neat part, you don't! Jokes aside, when I discussed it with @poljar (and I think I confirmed it by looking at the code), by the time we reach this part of the flow we'd have already applied any changes caused by the state events, including the room member ones, so it *should* be fine to not invalidate the cache, since it'll be used just for the current vector diffs.

I may be wrong though.

Hywan (approved):
The cache is local to `handle_remote_events_with_diffs`. I think it's correct. Let's go :-].

PR #5828, opened 2025-11-04
Fix up changelogs incorrectly updated since 0.14.0

richvdh:
All of these entries have been incorrectly added to the changelogs *since* 0.14.0 was released :(


poljar (approved):
Well that's very annoying that this keeps happening.

Sorry about that.

Looks good.

PR #5829, opened 2025-11-04
fix(sqlite): Fix the database version

Hywan:
The database has been updated but the version hasn't been bumped.

---

* Follow up of https://github.com/matrix-org/matrix-rust-sdk/pull/5819/.


Hywan:
I'll open a PR to address that. Thanks for the suggestion.

PR #5833, opened 2025-11-05
refactor(sdk): Fetch member data concurrently

jmartinesp:
Creating a `RoomMember` takes a lot of store queries, and previously all of them were done sequentially. I've tried to make this process run as much in parallel as I can so we can benefit from having a DB connection pool.

This *should* improve timeline loading times quite a bit, especially on Android: local tests suggested almost 50% faster loading times, since a lot of it is spent loading the room members associated with the initial timeline items.

Signed-off-by:

### DIFF
--- a/crates/matrix-sdk-base/src/room/members.rs
+++ b/crates/matrix-sdk-base/src/room/members.rs
@@ -176,26 +176,28 @@ impl Room {
         display_names: &'a [DisplayName],
     ) -> StoreResult<MemberRoomInfo<'a>> {
         let max_power_level = self.max_power_level();
-        let power_levels = self.power_levels_or_default();
+        let power_levels = async { Ok(self.power_levels_or_default().await) };
 
         let users_display_names =
             self.store.get_users_with_display_names(self.room_id(), display_names);
 
-        let raw_ignored_users =
-            self.store.get_account_data_event_static::<IgnoredUserListEventContent>();
+        let ignored_users = async {
+            Ok(self
+                .store
+                .get_account_data_event_static::<IgnoredUserListEventContent>()
+                .await?
+                .map(|c| c.deserialize())
+                .transpose()?
+                .map(|e| e.content.ignored_users.into_keys().collect()))
+        };
 
-        let (power_levels, users_display_names, raw_ignored_users) =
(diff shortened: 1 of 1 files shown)


jmartinesp:
> This should improve timeline loading times quite a bit, especially on Android: local tests suggested almost 50% faster loading times, since a lot of it is spent loading the room members associated with the initial timeline items.

Well, according to the benchmark it's now almost 10% slower... which doesn't match my tests at all 😅.

jmartinesp:
> Well, according to the benchmark it's now almost 10% slower... which doesn't match my tests at all 😅.

Turns out the benchmark is using an in-memory store, so that would explain why it's not benefitting from using the connection pool... since there is no pool 😬 . I guess we need to test this on real devices, or create a benchmark that uses the sqlite store.

crates/matrix-sdk-base/src/room/members.rs
```diff
@@ -181,7 +181,7 @@ impl Room {
         let users_display_names =
             self.store.get_users_with_display_names(self.room_id(), display_names);
 
-        let ignored_users = async move {
+        let ignored_users = async {
```
Hywan, line 184: Could be a fixup commit.

poljar:
> > Well, according to the benchmark it's now almost 10% slower... which doesn't match my tests at all 😅.
> 
> Turns out the benchmark is using an in-memory store, so that would explain why it's not benefitting from using the connection pool... since there is no pool 😬 . I guess we need to test this on real devices, or create a benchmark that uses the sqlite store.

If you do add a benchmark, do note that you'll need to run it locally. The benchmarks on the CI don't measure any sort of system call. The benchmarks on the CI run in [instrumentation mode](https://codspeed.io/docs/instruments/cpu/overview) and as such don't measure wall-clock time directly, the also just skip the measuring of system calls.

jmartinesp:
I'm waiting for some feedback on performance changes on iOS from @stefanceriu before merging.

jmartinesp:
> I'm waiting for some feedback on performance changes on iOS from @stefanceriu before merging.

Apparently this improvement is not noticeable on iOS, since the timeline loading times don't change that much. It seems like we can merge it safely.

PR #5836, opened 2025-11-05
feat(ffi): add bindings for granting login with a QR code

Johennes:
This adds FFI bindings for the two flows of granting login to a new device by way of a QR code.

### DIFF
--- a/bindings/matrix-sdk-ffi/src/client.rs
+++ b/bindings/matrix-sdk-ffi/src/client.rs
@@ -104,7 +104,7 @@ use crate::{
     encryption::Encryption,
     notification::{NotificationClient, NotificationEvent},
     notification_settings::NotificationSettings,
-    qr_code::LoginWithQrCodeHandler,
+    qr_code::{GrantLoginWithQrCodeHandler, LoginWithQrCodeHandler},
     room::{RoomHistoryVisibility, RoomInfoListener, RoomSendQueueUpdate},
     room_directory_search::RoomDirectorySearch,
     room_preview::RoomPreview,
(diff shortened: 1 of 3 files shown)


bindings/matrix-sdk-ffi/src/client.rs
```diff
@@ -576,19 +576,26 @@ impl Client {
...
     ) -> LoginWithQrCodeHandler {
         LoginWithQrCodeHandler::new(self.inner.oauth(), oidc_configuration)
     }
 
+    /// Create a handler for granting login from this device to a new device by
+    /// way of a QR code.
+    pub fn new_grant_login_with_qr_code_handler(self: Arc<Self>) -> GrantLoginWithQrCodeHandler {
+        GrantLoginWithQrCodeHandler::new(self.inner.oauth())
+    }
+
```
Johennes, line 598: CC @pixlwave for thoughts on the API shape.

pixlwave: Definitely clearer about whats going on from my perspective, thanks for the update 👍

bindings/matrix-sdk-ffi/src/qr_code.rs
```diff
@@ -123,6 +123,100 @@ impl LoginWithQrCodeHandler {
...
+#[matrix_sdk_ffi_macros::export]
+impl GrantLoginWithQrCodeHandler {
+    /// This method allows you to grant login with a scanned QR code.
+    ///
+    /// The new device needs to display the QR code which this device can
+    /// scan, call this method and handle its progress updates to grant the
+    /// login.
+    ///
+    /// This method uses the login mechanism described in [MSC4108]. As such,
+    /// it requires OAuth 2.0 support as well as Sliding Sync support.
```
poljar: Huh? Why would the grant login require sliding sync? Or do you mean that the bindings in general assume sliding sync support?

poljar (changes requested):
Looks good, except maybe the thing about sliding sync. Not sure if sync at all is relevant here.

bindings/matrix-sdk-ffi/src/qr_code.rs
Johennes: Hm, I just copied this part from the docs on `LoginWithQrCodeHandler` which were previously contained in the docs of the `login_with_qr_code` function. Does either of the two sides depend on sync actually?

poljar: I don't think so, certainly the new device can't sync before it logs in.

On the other hand, no messages are transferred over the sync, and we poll manually the `/devices` endpoint on the existing device.

I don't think the sync mechanism is relevant here at all.

Johennes: Fixed up in all 4 spots with 2e858ba6ca01b89c52643f883ba860f2b6b4235b.

I allowed myself to auto-squash this tiny fix-up directly to spare you another round trip on this PR.

PR #5837, opened 2025-11-05
refactor(common): Provide a unified, solid API between `try_lock_once` and `spin_lock` on `CrossProcessLock`

Hywan:
This patch changes the signature of `CrossProcessLock::try_lock_once`. It was returning a:

```rust
Result<CrossProcessLockResult, CrossProcessLockError>
```

Now it returns a:

```rust
Result<Result<CrossProcessLockKind, CrossProcessLockUnobtained>, L::LockError>
```

We will explain these new types in a moment.

This patch also changes the signature of `CrossProcessLock::spin_lock`. It was returning a:

```rust
Result<CrossProcessLockGuard, CrossProcessLockError>
```

Now it returns a:

```rust
Result<Result<CrossProcessLockKind, CrossProcessLockUnobtained>, L::LockError>
```

First off, we notice that the returned types are now unified. The `CrossProcessLockResult` type has been renamed `CrossProcessLockKind` and lives in a `Result::Ok`. The `CrossProcessLockResult::Unobtained` variant has been removed, but `CrossProcessLockUnobtainedReason` has been renamed to `CrossProcessLockUnobtained` and lives in a `Result::Err`.

Second, the `CrossProcessLockError` now is a union type between `CrossProcessLockUnobtained` and `TryLock::LockError`. It's not used by `try_lock_once` or `spin_lock`, but only by the code using the cross-process lock to provide a unified error type.

The ideas behind these changes are:

- it's easy to forward an error from the `TryLock`,
- it's difficult to ignore the `Clean` vs. `Dirty` state of the lock guard,
- unified API with clearly separated responsibility (the first `Result` vs. the second `Result`).

Note: the `CrossProcessLockKind::into_guard` method aims at being removed. It's useful now to maintain compatibility but it's “dangerous” as it makes trivial to skip `Clean` vs. `Dirty` states. We ultimately don't want that.

---

### DIFF
--- a/crates/matrix-sdk-base/src/event_cache/store/mod.rs
+++ b/crates/matrix-sdk-base/src/event_cache/store/mod.rs
@@ -84,7 +84,7 @@ impl EventCacheStoreLock {
 
     /// Acquire a spin lock (see [`CrossProcessLock::spin_lock`]).
     pub async fn lock(&self) -> Result<EventCacheStoreLockGuard<'_>, CrossProcessLockError> {
-        let cross_process_lock_guard = self.cross_process_lock.spin_lock(None).await?;
+        let cross_process_lock_guard = self.cross_process_lock.spin_lock(None).await??.into_guard();
 
         Ok(EventCacheStoreLockGuard { cross_process_lock_guard, store: self.store.deref() })
     }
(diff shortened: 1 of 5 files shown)


poljar (approved):
Thanks, looking good.

PR #5838, opened 2025-11-06
WASM Improvements

Marcel-Nordeck:
This brings a few issues I noticed while trying to port https://gitlab.opencode.de/fitko/matrix-g2x/fit-neo-web/-/merge_requests/3 from the filament Wasm mega PR over to the main branch of the SDK. Without these (and downgrading uniffi 0.29.4 due to https://github.com/jhugman/uniffi-bindgen-react-native/issues/316 ) the bindings cannot compile.

Note that also the inline stubs seem to be required as uniffi does checksums over the actual function signature _including_ any attributes. This seems to happen _before_ evaluation of them, however. Which caused checksum mismatches if I used nicer looking cfg attributes (splitting Wasm and non Wasm world into 2 functions instead of inline blocks). Ideally, one day uniffi gets smarter about that, but it doesn't seem to be a thing yet. https://mozilla.github.io/uniffi-rs/latest/proc_macro/index.html#conditional-compilation seems to also not help us here.

One more thing I noticed is that while the indexeddb support was added, the `bundled-sqlite` feature still stayed as a default feature. This causes issues on Wasm as SQLite can't be compiled on that platform using the `sqlite-sys` package which rust-sdk utilizes. (It will complain about missing headers). Therefore, I think it's easiest to just remove it from the defaults and have it active conditionally by the consumer.

Finally, WASM threading support currently lacks Send, so uniffi has to be told to be less strict about their requirements to support this. This might change in the future, however, if Wasm gets proper threading support in rust land. Additionally, there was some code using futures-executor, but the dependency was not added to the wasm platform.

---

---

(This PR is made with my Nordeck :tophat: for the FITKO Matrix-G2X project over at https://gitlab.opencode.de/fitko/matrix-g2x/gitlab-profile )

### DIFF
--- a/bindings/matrix-sdk-ffi/Cargo.toml
+++ b/bindings/matrix-sdk-ffi/Cargo.toml
@@ -24,7 +24,7 @@ crate-type = [
 ]
 
 [features]
-default = ["unstable-msc4274", "experimental-element-recent-emojis"]
+default = ["bundled-sqlite", "unstable-msc4274", "experimental-element-recent-emojis"]
 # Use SQLite for the session storage.
 sqlite = ["matrix-sdk/sqlite"]
 # Use an embedded version of SQLite.


Marcel-Nordeck:
Hm not entirely sure why it fails on the feature flag for uniffi as the workspace Cargo.toml claims 0.30.0. I know it works with 0.29.4 upwards. I am guessing it may just need a new lockfile right?

Marcel-Nordeck:
Also sorry for the force pushing. Realised the conventional commit policy too late. From now on, I will instead use fixup commits when needed.

poljar:
> Hm not entirely sure why it fails on the feature flag for uniffi as the workspace Cargo.toml claims 0.30.0. I know it works with 0.29.4 upwards. I am guessing it may just need a new lockfile right?

complement-crypto, which is used for one of the CI runs, downgrades to `0.28.x` since `uniffi-bindgen-go` doesn't support `0.29` or `0.30` yet.

I think only enabling this feature under Wasm, which makes sense regardless of complement-crypto, would be one way to solve this problem.

poljar:
> > Hm not entirely sure why it fails on the feature flag for uniffi as the workspace Cargo.toml claims 0.30.0. I know it works with 0.29.4 upwards. I am guessing it may just need a new lockfile right?
> 
> complement-crypto, which is used for one of the CI runs, downgrades to `0.28.x` since `uniffi-bindgen-go` doesn't support `0.29` or `0.30` yet.
> 
> I think only enabling this feature under Wasm, which makes sense regardless of complement-crypto, would be one way to solve this problem.

Well ok, I updated complement-crypto here https://github.com/matrix-org/complement-crypto/pull/215.

bindings/matrix-sdk-ffi/Cargo.toml
```diff
@@ -24,7 +24,7 @@ crate-type = [
 ]
 
 [features]
-default = ["bundled-sqlite", "unstable-msc4274", "experimental-element-recent-emojis"]
+default = ["unstable-msc4274", "experimental-element-recent-emojis"]
```
poljar: Is this necessary? Can't you use `--no-default-features` to get the feature set you'd want/need on Wasm?

> Therefor we probably
shouldn't have it as the default since indexeddb support was added.

Why not? The defaults are here to serve the most common use-case. Having only the memory store by default seems like a worse situation.

The `matrix-sdk` crate also enables SQLite support by default, even though it ~always supported IndexedDB.

bindings/matrix-sdk-ffi/src/client.rs
```diff
@@ -942,22 +941,54 @@ impl Client {
...
+            return Ok(());
+        }
+
+        #[cfg(target_family = "wasm")]
+        {
+            Err(ClientError::Generic {
+                msg: "set_display_name is not supported on wasm platforms".to_owned(),
+                details: None,
+            })
+        }
```
poljar, line 995: Huh, why isn't this supported on Wasm?

bindings/matrix-sdk-ffi/src/client.rs
```diff
@@ -942,22 +941,54 @@ impl Client {
...
+                    &MediaRequestParameters {
+                        source: source.media_source,
+                        format: MediaFormat::File,
+                    },
+                    filename,
+                    &mime_type,
+                    use_cache,
+                    temp_dir,
+                )
+                .await?;
```
poljar, line 962: A comment explaining why this doesn't work would be nice.

I guess it's because of the usage of the temp dir.

Marcel-Nordeck: Yes thats exactly why. I added a comment here now. I guess for the future it would be cool to find a way to support this without the temporary file though. But SdkMediaFileHandle (the MediaFileHandle in the main sdk) isn't currently designed to allow for a solution without a filesystem.

bindings/matrix-sdk-ffi/src/client.rs
Marcel-Nordeck: So refreshing my own memory the issue is simply that .context doesnt work on wasm due to:

```
error[E0599]: the method `context` exists for enum `std::result::Result<(), matrix_sdk::Error>`, but its trait bounds were not satisfied
   --> bindings/matrix-sdk-ffi/src/client.rs:979:14
    |
975 | /         self.inner
976 | |             .account()
977 | |             .set_display_name(Some(name.as_str()))
978 | |             .await
979 | |             .context("Unable to set display name")?;
    | |             -^^^^^^^ method cannot be called on `std::result::Result<(), matrix_sdk::Error>` due to unsatisfied trait bounds
    | |_____________|
    |
    |
   ::: /run/media/marcel/1ac60e61-e5e2-4157-9a7a-830801a3fb0b/projects/fitko/fit-neo-web-v2/rust_modules/matrix-rust-sdk/crates/matrix-sdk/src/error.rs:265:1
    |
265 |   pub enum Error {
    |   -------------- doesn't satisfy `matrix_sdk::Error: anyhow::context::ext::StdError`, `matrix_sdk::Error: std::marker::Send` or `matrix_sdk::Error: std::marker::Sync`
    |
   ::: /home/marcel/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/result.rs:557:1
    |
557 |   pub enum Result<T, E> {
    |   --------------------- doesn't satisfy `_: Context<(), Error>`
    |
    = note: the following trait bounds were not satisfied:
            `matrix_sdk::Error: anyhow::context::ext::StdError`
            which is required by `std::result::Result<(), matrix_sdk::Error>: anyhow::Context<(), matrix_sdk::Error>`
            `matrix_sdk::Error: std::marker::Send`
            which is required by `std::result::Result<(), matrix_sdk::Error>: anyhow::Context<(), matrix_sdk::Error>`
            `matrix_sdk::Error: std::marker::Sync`
            which is required by `std::result::Result<(), matrix_sdk::Error>: anyhow::Context<(), matrix_sdk::Error>`
```

However in hindsight this is not a good way to solve it. I will have another stab at doing this in a nicer way that doesnt loose the actual function.

poljar: I think removing the context on Wasm is better than removing the whole function.

bindings/matrix-sdk-ffi/Cargo.toml
Marcel-Nordeck: Right, that's fixed on my downstream side now. I am guessing reverting 463589e69cdf2ea84daa4454c8a4c03bb8811cd7 is fine for the process, or is a fixup! commit preferred?

poljar: Doesn't really matter, we'll squash merge whichever you produce. I guess a revert is slightly less work on your side.

bindings/matrix-sdk-ffi/src/client.rs
```diff
@@ -942,22 +941,60 @@ impl Client {
...
+    }
+
+    pub async fn set_display_name(&self, name: String) -> Result<(), ClientError> {
+        #[cfg(not(target_family = "wasm"))]
+        {
+            self.inner
+                .account()
+                .set_display_name(Some(name.as_str()))
+                .await
+                .context("Unable to set display name")?;
```
poljar, line 984: Didn't we want to just get rid of the `context()` call here?

poljar (changes requested):
The `set_display_name()` method is still unnecessarily disable on Wasm.

I think the rest looks good now.

bindings/matrix-sdk-ffi/src/client.rs
Marcel-Nordeck: The context call still works for non wasm platforms.

So what it now does is use map_err on wasm and context everywhere else. That way it should be (almost) the same for both cases in behaviour. Of course map_err isn't identical to context (especially due to just using the to_string() message on wasm) but it still generally ensures the information is the same on both.

But I can also just remove it in general I guess. I just thought it might make sense to preserve the context on existing bindings basically.

Marcel-Nordeck: Also its not disbled on wasm anymore. It does actually execute the action on wasm now. It just does different error handling.

poljar: Ah, sorry I misread this part. Thanks for pointing this out.

PR #5840, opened 2025-11-07
Fix a panic in the qr login granting tests

poljar:
The QR login granting tests often move the `MatrixMockServer` to a different tokio task while some `MockGuard`s stay in the main task of the test.

This can lead to the mock server being dropped before the `MockGuard`s thus leading to a panic.

This PR moves the `MockGuard`s contained in the `MockedRendezvousServer` to the task as well, avoiding this separation of lifetimes.

I attempted to tie the lifetimes together by letting `MockedRendezvousServer` hold on to a reference of the `MatrixMockServer` but that doesn't seem to have helped.

A review commit by commit is the way to go here, as there are two semi-related commits before the main patch.

cc: @Johennes not sure if you have a better idea on how to fix this.

### DIFF
--- a/crates/matrix-sdk/src/authentication/oauth/qrcode/grant.rs
+++ b/crates/matrix-sdk/src/authentication/oauth/qrcode/grant.rs
@@ -403,11 +403,16 @@ mod test {
         DeviceNotCreated,
     }
 
+    #[allow(clippy::too_many_arguments)]
     async fn request_login_with_scanned_qr_code(
         behaviour: BobBehaviour,
         qr_code_rx: oneshot::Receiver<QrCodeData>,
         check_code_tx: oneshot::Sender<u8>,
         server: MatrixMockServer,
+        // The rendezvous server is here because it contains MockGuards that are tied to the
+        // lifetime of the MatrixMockServer. Otherwise we might attempt to drop the
+        // MatrixMockServer before the MockGuards.
+        _rendezvous_server: MockedRendezvousServer,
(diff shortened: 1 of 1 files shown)


crates/matrix-sdk/src/authentication/oauth/qrcode/grant.rs
```diff
@@ -797,11 +797,9 @@ mod test {
...
 
         // Wait for all tasks to finish.
-        join!(
-            async { updates_task.await.expect("Alice should run through all progress states") },
-            async { grant.await.expect("Alice should be able to grant the login") },
-            async { bob_task.await.expect("Bob's task should finish") }
-        );
+        grant.await.expect("Alice should be able to grant the login");
+        updates_task.await.expect("Alice should run through all progress states");
+        bob_task.await.expect("Bob's task should finish");
```
Johennes, line 813: Ohhh, right! 🤦‍♂️ 

I guess the `join` is superfluous in the tests in `login.rs`, too, then because two of the futures were already `spawn`ed there as well?

Johennes:
Hm, I cannot think of another way from the top of my head. Given the existing complexity of these tests, this seems like a small tweak and, thus, quite acceptable to me though. Thanks a lot for catching and fixing this!

crates/matrix-sdk/src/authentication/oauth/qrcode/grant.rs
poljar: Yeah, the join doesn't hurt much but makes the test execution even less predictable.

PR #4741, opened 2025-03-03
feat(ui): `RoomListService` re-enables `share_pos`

Hywan:
WIP

The idea is the following: if we keep the sliding sync `pos`, first off, the homeservers will be happier because they don't have to re-compute a session every time, and second, the overall user experience may be better as we receive much less `limited` responses from the homeservers.

PR #5852, opened 2025-11-11
fix(ui): Undo an optimisation to start at `SettingUp`

Hywan:
This patch undo an optimisation that was initialising the `RoomListService` at the `SettingUp` state if a `pos` value was recovered successfully (see bbf9bf2c0be026c2bcf237f48953918057a1e15c from https://github.com/matrix-org/matrix-rust-sdk/pull/4741). The problem is that it starts with a range of 0..99 instead of 0..19, which can slow things done in particular cases. Whilst a good idea on paper, it's not in practise. So let's continue to recover the `pos`, but let's keep starting at the `Init` state.


pixlwave:
Related: https://github.com/element-hq/element-x-ios/issues/4728

PR #5853, opened 2025-11-11
Don't do an authenticated `/versions` request in the `RoomListService` builder

### LINKED ISSUE #5845: Building the `SyncService` can fail if a `/versions` request fails
Since https://github.com/matrix-org/matrix-rust-sdk/pull/5618 building a `SyncService` might fail to build because a `/versions` requests fails.

The builder attempts to check if support for [MSC4306](https://github.com/matrix-org/matrix-spec-proposals/pull/4306) is enabled on the homeserver.

https://github.com/matrix-org/matrix-rust-sdk/blob/f9584f5b2af7d71e031196071e199cacda1e7815/crates/matrix-sdk-ui/src/room_list_service/mod.rs#L160-L180

Critically, `/versions` nowadays can also fail due to token expiration, for more info see https://github.com/matrix-org/matrix-rust-sdk/issues/5816 and https://github.com/matrix-org/matrix-rust-sdk/pull/5822. This combination means that building a `SyncService` can simply fail due to a network request failing. Just enabling token refresh on the `/versions` request results in a deadlock: https://github.com/matrix-org/matrix-rust-sdk/pull/5825.

Two things that would be nice to achieve:
1. Remove the possibility of a network failure to fail the `SyncService` builder.
2. Allow token refres for the `/versions` request.

poljar:
This closes #5845

This rather avoids the issue instead of fixing it. As the `/versions` call in the `RoomListService` is only used to figure out if a certain feature the user has enabled should actually be enabled and this is a release blocker for EX, I deemed that it's fine to avoid tackling the problem that `/versions` calls can't do token refresh.

### DIFF
--- a/crates/matrix-sdk-ui/src/room_list_service/mod.rs
+++ b/crates/matrix-sdk-ui/src/room_list_service/mod.rs
@@ -63,7 +63,7 @@ use eyeball::Subscriber;
 use futures_util::{Stream, StreamExt, pin_mut};
 use matrix_sdk::{
     Client, Error as SlidingSyncError, Room, SlidingSync, SlidingSyncList, SlidingSyncMode,
-    event_cache::EventCacheError, timeout::timeout,
+    config::RequestConfig, event_cache::EventCacheError, timeout::timeout,
 };
 pub use room_list::*;
 use ruma::{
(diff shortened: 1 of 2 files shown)


crates/matrix-sdk-ui/tests/integration/room_list_service.rs
```diff
@@ -2906,7 +2907,8 @@ async fn test_thread_subscriptions_extension_enabled_only_if_server_advertises_i
             .mock_versions()
             .ok_custom(&["v1.11"], &features_map)
             .named("/versions, first time")
-            .mock_once()
+            .up_to_n_times(3)
+            .expect(3..)
```
poljar, line 2915: The `/versions` call in the `RoomListService` builder won't fill the cache for the `/versions` call anymore. So we're doing a couple more of those.

Hywan: Can we add a comment to explain that please?

crates/matrix-sdk-ui/src/room_list_service/mod.rs
```diff
@@ -158,11 +158,20 @@ impl RoomListService {
...
+                .supported_versions_cached()
                 .await
-                .map_err(|err| Error::SlidingSync(err.into()))?
-                .features;
+                .map_err(|e| Error::SlidingSync(e.into()))?
+            {
+                cached.features
+            } else {
+                client
+                    .fetch_server_versions(Some(RequestConfig::new().skip_auth()))
```
Hywan, line 175: Could we explain why we call `skip_auth` here please?

poljar: Added now https://github.com/matrix-org/matrix-rust-sdk/pull/5853/commits/4ed853a84cc74463f6de8f08406167242120ae98#diff-d16dcf09afbab7ae166212f775a19928d6631642a910e1b098407aeea658b084R168-R173

crates/matrix-sdk-ui/tests/integration/room_list_service.rs
poljar: Sure.

PR #5854, opened 2025-11-11
Add better default target-feature values for Android in ARM64 devices

jmartinesp:
What the title says. These are the same flags used for iOS ARM64, but for some reason they're not enabled by default for `aarch64-linux-android`. When testing locally, this improved performance quite a bit.

Signed-off-by:


.cargo/config.toml
```diff
@@ -7,3 +7,7 @@ crates-io = "https://docs.rs/"
 
 [unstable]
 rustdoc-map = true
+
+[target.aarch64-linux-android]
+# These rust flags improve the performance on Android on arm64
+rustflags = ["-C", "target-feature=+neon,+aes,+sha2,+sha3,+pmuv3"]
```
poljar: Missing newline at the end of the file.

JadedBlueEyes:
For context: https://developer.android.com/ndk/guides/cpu-arm-neon

> All ARMv8-based ("arm64") Android devices support Neon. Almost all ARMv7-based ("32-bit") Android devices support Neon, including all devices that shipped with API level 21 or later. The NDK enables Neon by default for both Arm ABIs.

All devices since android 5 / 2014 support this. I didn't check the other flags, but I woudn't be too worried about older devices here

PR #5855, opened 2025-11-11
feat(base): Explicitly handle the `CrossProcessLockKind::Dirty` case in `MediaStore`

Hywan:
This patch replaces the `into_guard()` call by a `match` over `CrossProcessLockKind` so that the `Dirty` case is explicitly handled.

The mid-term idea is to remove the `into_guard()` method because it is “dangerous” as it hides the `Dirty` case.

---

### DIFF
--- a/crates/matrix-sdk-base/src/media/store/mod.rs
+++ b/crates/matrix-sdk-base/src/media/store/mod.rs
@@ -33,7 +33,7 @@ use std::{ops::Deref, sync::Arc};
 
 use matrix_sdk_common::cross_process_lock::{
     CrossProcessLock, CrossProcessLockError, CrossProcessLockGeneration, CrossProcessLockGuard,
-    TryLock,
+    CrossProcessLockKind, TryLock,
 };
 use matrix_sdk_store_encryption::Error as StoreEncryptionError;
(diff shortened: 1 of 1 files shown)


crates/matrix-sdk-common/src/cross_process_lock.rs
```diff
@@ -478,6 +478,12 @@ pub enum CrossProcessLockKind {
     /// obtained it again. The lock is marked as dirty. It means the value
     /// protected by the cross-process lock may need to be reloaded if
     /// synchronisation is important.
+    ///
+    /// Until [`CrossProcessLock::clear_dirty`] is called,
+    /// [`CrossProcessLock::is_dirty`], [`CrossProcessLock::try_lock_once`] and
+    /// [`CrossProcessLock::spin_lock`] will express the lock is dirty. Put it
```
stefanceriu:
```suggestion
    /// [`CrossProcessLock::spin_lock`] will report the lock as dirty. Put it
```

Perhaps that's clearer?

Hywan: Much better, thanks!

PR #5856, opened 2025-11-11
feat(base,sdk): Support for dirty cross-process lock in the `EventCache`

Hywan:
The `EventCache` stores and organises all events. A `RoomEventCache` represents a subset of the `EventCache`, be all events associated to a particular room. That's the type used by each `Room` to store and organise their events. The `EventCacheStore` is the type representing the store for the `EventCache`. It is used by `RoomEventCache` to represent the _stored data_. `RoomEventCache` also has in-memory data. In-memory data are always in-sync with data in the store. The `EventCacheStore` is behind a `CrossProcessLock`, a special lock that protects data from being acquired across several processes. So far, so good.

We've recently introduced the possibility for a cross-process to be invalidated/to be dirty, see #4874. When a `CrossProcessLock` is obtained, it is now possible to know if another process obtained it before, thus invalidating the loaded data from the store we might have in-memory. It is a sign we need to refresh the in-memory data.

This is what this set of patches is doing: we are reacting to a dirty cross-process lock and refreshing the in-memory state of `RoomEventCache` accordingly.

This set of patches must be reviewed one-by-one to understand the modifications and their impact. The tl;dr is the following:

- `RoomEventCacheState` becomes `RoomEventCacheStateLock`, representing 2 locks at once: the per-thread lock over the state, and the cross-process lock over the store,
- `RoomEventCacheStateInner` becomes `RoomEventCacheStateLockInner`,
- `RoomEventCacheInner::state` holds a `RoomEventCacheLock` instead of a `RwLock<RoomEventCacheState>`,
- `RoomEventCacheLock` implements a `read` and a `write` methods, to respectively obtain a read and a write lock. These operations are no longer infallible. It impacts a couple of callers, but nothing fancy.

The impacted code is mostly internal code.

## Todo

- [x] Ensure `VectorDiff`s are broadcasted to the subcribers. This isn't done by `shrink_to_last_chunk` because it's supposed to run when no more subscribers are listening. We must handle this, I forgot!

---

### DIFF
--- a/crates/matrix-sdk/src/event_cache/room/mod.rs
+++ b/crates/matrix-sdk/src/event_cache/room/mod.rs
@@ -418,7 +418,14 @@ impl RoomEventCache {
     /// Return a nice debug string (a vector of lines) for the linked chunk of
     /// events for this room.
     pub async fn debug_string(&self) -> Vec<String> {
-        self.inner.state.read().await.unwrap().room_linked_chunk().debug_string()
+        match self.inner.state.read().await {
+            Ok(read_guard) => read_guard.room_linked_chunk().debug_string(),
+            Err(err) => {
+                warn!(?err, "Failed to obtain the read guard for the `RoomEventCache`");
+
+                vec![]
+            }
+        }
     }
 }
 


crates/matrix-sdk/src/event_cache/room/mod.rs
```diff
@@ -708,7 +734,18 @@ mod private {
...
         ) -> Result<Self, EventCacheError> {
-            let store_lock = store.lock().await?;
+            let store_guard = match store.lock().await? {
+                // Lock is clean: all good!
+                EventCacheStoreLockState::Clean(guard) => guard,
+
+                // Lock is dirty, not a problem, it's the first time we are creating this state, no
+                // need to refresh.
+                EventCacheStoreLockState::Dirty(guard) => {
+                    EventCacheStoreLockGuard::clear_dirty(&guard);
```
Hywan, line 771: Note: I must add a test for this case, and see the dirty marker is reset.

crates/matrix-sdk/src/event_cache/room/mod.rs
```diff
@@ -749,182 +786,341 @@ mod private {
...
-                return Err(EventCacheError::InvalidLinkedChunkMetadata {
-                    details: format!(
-                        "chunks {} and {} both claim to be last chunks",
-                        last.identifier.index(),
-                        other_last.identifier.index()
-                    ),
-                });
-            }
+                    // All good now, mark the cross-process lock as non-dirty.
+                    EventCacheStoreLockGuard::clear_dirty(&guard.store);
```
Hywan, line 933: Note: I must add an assertion to see if the dirty marker is reset.

crates/matrix-sdk/src/event_cache/room/mod.rs
```diff
@@ -749,182 +786,341 @@ mod private {
...
+                    let mut guard = RoomEventCacheStateLockWriteGuard {
+                        state: state_guard,
+                        store: store_guard,
+                    };
+                    guard.shrink_to_last_chunk().await?;
 
-                current = *pred_meta;
-            }
+                    // All good now, mark the cross-process lock as non-dirty.
+                    EventCacheStoreLockGuard::clear_dirty(&guard.store);
```
Hywan, line 989: Note: I must add an assertion to see if the dirty marker is reset.

crates/matrix-sdk/src/event_cache/room/mod.rs
```diff
@@ -812,6 +814,15 @@ mod private {
             })
         }
 
+        /// Lock this [`RoomEventCacheStateLock`] with per-thread shared access.
```
andybalaam, line 852: Does this really have per-thread shared access given it calls `let state_guard = self.locked_state.write().await;` inside?

crates/matrix-sdk/src/event_cache/room/mod.rs
```diff
@@ -909,6 +904,19 @@ mod private {
         store: EventCacheStoreLockGuard,
     }
 
+    impl<'a> RoomEventCacheStateLockWriteGuard<'a> {
+        /// Synchronously downgrades a write lock into a read lock.
+        ///
+        /// The per-thread/state lock is downgraded atomically, without allowing
+        /// any writers to take exclusive access of the lock in the meantime.
+        ///
+        /// It returns an RAII guard which will drop the write access to the
```
andybalaam, line 1040: Since it's been downgraded, won't the RAII guard drop the _read_ access to state, not the _write_ access?

crates/matrix-sdk/src/event_cache/room/mod.rs
```diff
@@ -824,17 +830,78 @@ mod private {
...
-                        state: state_guard.downgrade(),
-                        store: store_guard,
-                    })
+                    Ok(RoomEventCacheStateLockReadGuard { state: state_guard, store: store_guard })
                 }
                 EventCacheStoreLockState::Dirty(store_guard) => {
+                    // Drop the read lock, and take a write lock to modify the state.
+                    // This is safe because only one semaphore permit exists.
+                    drop(state_guard);
+                    let state_guard = self.locked_state.write().await;
```
andybalaam, line 922: Ah, I see you addressed my earlier comment!

crates/matrix-sdk/src/event_cache/room/mod.rs
Hywan: Done.

crates/matrix-sdk/src/event_cache/room/mod.rs
Hywan: Done.

crates/matrix-sdk/src/event_cache/room/mod.rs
Hywan: Done.

crates/matrix-sdk-ui/src/timeline/builder.rs
```diff
@@ -164,7 +164,7 @@ impl TimelineBuilder {
         room.client().event_cache().subscribe()?;
 
         let (room_event_cache, event_cache_drop) = room.event_cache().await?;
-        let (_, event_subscriber) = room_event_cache.subscribe().await;
+        let (_, event_subscriber) = room_event_cache.subscribe().await.unwrap();
```
poljar: Naked unwrap?

crates/matrix-sdk/src/event_cache/room/mod.rs
Hywan: The first write lock is transformed to a read lock as soon as possible. The result for the caller is indeed a per-thread shared access.

crates/matrix-sdk/src/event_cache/room/mod.rs
Hywan: Hmm no, it drops the write lock and obtains a read lock, so the RAII guard drops the write access.

crates/matrix-sdk-ui/src/timeline/builder.rs
```diff
@@ -209,36 +209,36 @@ impl TimelineBuilder {
...
+                        room_id = ?room.room_id(),
+                        focus = focus.debug_string(),
+                        prefix = internal_id_prefix
+                    );
+                    span.follows_from(Span::current());
+
+                    // Note: must be done here *before* spawning the task, to avoid race conditions
+                    // with event cache updates happening in the background.
+                    let (_events, receiver) =
+                        room_event_cache.subscribe_to_thread(root.clone()).await.unwrap();
```
poljar: Another naked unwrap.

labs/multiverse/src/widgets/room_view/details/events.rs
```diff
@@ -28,7 +28,7 @@ impl Widget for &mut EventsView<'_> {
                 let events = tokio::task::block_in_place(|| {
                     Handle::current().block_on(async {
                         let (room_event_cache, _drop_handles) = room.event_cache().await.unwrap();
-                        room_event_cache.events().await
+                        room_event_cache.events().await.unwrap()
```
poljar: I know this is multiverse, but please `expect()` here as well.

crates/matrix-sdk/src/event_cache/redecryptor.rs
```diff
@@ -280,7 +289,7 @@ impl EventCache {
         // Get the cache for this particular room and lock the state for the duration of
         // the decryption.
         let (room_cache, _drop_handles) = self.for_room(room_id).await?;
-        let mut state = room_cache.inner.state.write().await;
+        let mut state = room_cache.inner.state.write().await?;
```
poljar, line 305: This means that we won't replace the events that got decrypted, shouldn't we attempt to clean up the cache in this case and replace the events?

crates/matrix-sdk/src/event_cache/pagination.rs
```diff
@@ -182,11 +185,13 @@ impl RoomPagination {
         // there's no previous events chunk to load.
 
         loop {
-            let mut state_guard = self.inner.state.write().await;
+            let mut state_guard = self.inner.state.write().await?;
```
poljar, line 188: I guess it's fine to fail the pagination her, it's a user requested pagination so we can tell them "oh it didn't work, please try again later".

crates/matrix-sdk/src/event_cache/room/mod.rs
```diff
@@ -397,15 +402,23 @@ impl RoomEventCache {
...
+                warn!("couldn't save event in the event cache: {err}");
+            }
         }
     }
 
     /// Return a nice debug string (a vector of lines) for the linked chunk of
     /// events for this room.
     pub async fn debug_string(&self) -> Vec<String> {
-        self.inner.state.read().await.room_linked_chunk().debug_string()
+        self.inner.state.read().await.unwrap().room_linked_chunk().debug_string()
```
poljar: Naked unwrap. I think it would be much better to return an empty `Vec` and log an error.

It is debug functionality, but it is `pub` and apparently we're even exposing this over the FFI.

crates/matrix-sdk/src/event_cache/room/mod.rs
```diff
@@ -749,182 +806,389 @@ mod private {
...
+                self.read_lock_acquisition.available_permits(),
+                0,
+                "The semaphore must have only one permit"
+            );
 
-            // Transform the vector into a hashmap, for quick lookup of the predecessors.
-            let chunk_map: HashMap<_, _> =
-                all_chunks.iter().map(|meta| (meta.identifier, meta)).collect();
+            // Obtain a read lock.
+            let state_guard = self.locked_state.read().await;
```
poljar, line 911: Since the semaphore has only a single permit, you could have used a Mutex as well.

A Mutex to protect the RWLock, which binds the two a bit better together, but it does look silly. Semaphore is the more common choice for such critical section protections.

crates/matrix-sdk/src/event_cache/room/mod.rs
```diff
@@ -749,182 +806,389 @@ mod private {
...
+    impl<'a> RoomEventCacheStateLockWriteGuard<'a> {
+        /// Synchronously downgrades a write lock into a read lock.
+        ///
+        /// The per-thread/state lock is downgraded atomically, without allowing
+        /// any writers to take exclusive access of the lock in the meantime.
+        ///
+        /// It returns an RAII guard which will drop the write access to the
+        /// state and to the store when dropped.
+        fn downgrade(self) -> RoomEventCacheStateLockReadGuard<'a> {
+            RoomEventCacheStateLockReadGuard { state: self.state.downgrade(), store: self.store }
```
poljar, line 1043: Hey, neat that you can downgrade, didn't notice that till now.

crates/matrix-sdk/src/event_cache/redecryptor.rs
poljar: Ah no, the `write()` method already does so, this is a true storage error.

If that's correct ignore my previous comment and just mark this as resolved.

crates/matrix-sdk-ui/src/timeline/builder.rs
Hywan: Fixed.

crates/matrix-sdk-ui/src/timeline/builder.rs
Hywan: Fixed, thanks.

labs/multiverse/src/widgets/room_view/details/events.rs
Hywan: Done, thanks!

crates/matrix-sdk/src/event_cache/room/mod.rs
Hywan: ❤️

crates/matrix-sdk/src/event_cache/room/mod.rs
Hywan: You're pointing something important. The semantics between `Semaphore(permit=1)` and `Mutex` is the same, but with `Mutex`, we don't have to guarantee there is a single permit. I'm going to change for a `Mutex`.

Hywan: Done.

crates/matrix-sdk/src/event_cache/room/mod.rs
Hywan: Done, thanks!

crates/matrix-sdk-ui/src/timeline/builder.rs
```diff
@@ -164,7 +164,10 @@ impl TimelineBuilder {
         room.client().event_cache().subscribe()?;
 
         let (room_event_cache, event_cache_drop) = room.event_cache().await?;
-        let (_, event_subscriber) = room_event_cache.subscribe().await;
+        let (_, event_subscriber) = room_event_cache
+            .subscribe()
+            .await
+            .expect("Failed to subscribe to the `RoomEventCache` because of a store error");
```
poljar: But the `build()` method returns an error, why wouldn't we return the error instead of panicking?

poljar (changes requested):
Still one `expect()` that feels like it shouldn't be there.

crates/matrix-sdk-ui/src/timeline/builder.rs
Hywan: True! Resolved.

poljar (approved):
Alright, I trust you that you fixed the last expect.

crates/matrix-sdk-ui/src/timeline/builder.rs
```diff
@@ -224,7 +224,11 @@ impl TimelineBuilder {
 
                 // Note: must be done here *before* spawning the task, to avoid race conditions
                 // with event cache updates happening in the background.
-                let (_events, receiver) = room_event_cache.subscribe_to_thread(root.clone()).await;
+                let (_events, receiver) =
+                        room_event_cache
+                            .subscribe_to_thread(root.clone())
+                            .await
+                            .expect("Failed to thread-subscribe to the `RoomEventCache` because of a store error");
```
poljar: Wait, but this one's still here? Can't we return the error as well?

Hywan: Fixed!

Issue #5839, opened 2025-11-06
Expose m.space.child permission value from RoomPowerLevelsValues


ganfra:
Expose the `m.space.child` from the `RoomPowerLevelsValues`  

https://github.com/matrix-org/matrix-rust-sdk/blob/c064ca8b18137335856d6a5656590e51392ba182/bindings/matrix-sdk-ffi/src/room/power_levels.rs#L188

PR #5857, opened 2025-11-11
feat(sdk): Add a power level value field for `StateEventType::SpaceChild`

jmartinesp:
Closes https://github.com/matrix-org/matrix-rust-sdk/issues/5839

Signed-off-by:


stefanceriu (approved):
We agreed to rename `manage_rooms_in_space` to `space_child` but otherwise lgtm 👍

PR #5858, opened 2025-11-12
feat(timeline): utilize the cache and include common relations when focusing on an event without context

Johennes:
At the moment, `TimelineController::init_focus` doesn't use the event cache at all in the `TimelineFocus::Event` arm. I believe this is because the cache currently cannot load surrounding context for an event. However, when no context is requested, meaning `num_context_events = 0`, this result in a potentially superfluous `/context` request.

This pull request, differentiates this special case and attempts to load the event from the cache before falling back to the server.

Additionally, common relations (reactions & edits) are included, if possible. This is similar to the behavior in the `TimelineFocus::PinnedEvents` arm and allows using focused timelines to power event details views.

Open questions:

- Given the similarity to the `TimelineFocus::PinnedEvents` behavior, I'm not sure if it might be better to generalize `TimelineFocus::PinnedEvents` to support both cases.
- It would be great if `TimelineFocus::Event` would make use of the event cache regardless of whether context was requested or not. I'm not sure if that's feasibly possible though.

### DIFF
--- a/crates/matrix-sdk-ui/src/timeline/controller/mod.rs
+++ b/crates/matrix-sdk-ui/src/timeline/controller/mod.rs
@@ -22,6 +22,7 @@ use imbl::Vector;
 #[cfg(test)]
 use matrix_sdk::Result;
 use matrix_sdk::{
+    config::RequestConfig,
     deserialized_responses::TimelineEvent,
     event_cache::{DecryptionRetryRequest, RoomEventCache, RoomPaginationStatus},
     paginators::{PaginationResult, PaginationToken, Paginator},
@@ -40,7 +41,7 @@ use ruma::{
         poll::unstable_start::UnstablePollStartEventContent,
         reaction::ReactionEventContent,
         receipt::{Receipt, ReceiptThread, ReceiptType},
-        relation::Annotation,
+        relation::{Annotation, RelationType},
         room::message::{MessageType, Relation},
     },
     room_version_rules::RoomVersionRules,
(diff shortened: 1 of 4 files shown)


crates/matrix-sdk-ui/src/timeline/controller/mod.rs
```diff
@@ -474,16 +475,40 @@ impl<P: RoomDataProvider> TimelineController<P> {
...
+                        }
+                    }
+                } else {
+                    // Start a /context request to load the focussed event and surrounding events.
+                    let start_from_result = event_paginator
+                        .start_from(event_id, (*num_context_events).into())
+                        .await
+                        .map_err(PaginationError::Paginator)?;
+                    start_from_result.events
+                };
```
Johennes, line 509: A potential issue I'm now seeing is that the `if` branch will leave the paginator in initial state meaning you cannot use it to extend the timeline later. This isn't really an issue for the intended event details use case but might still be problematic. Maybe this is further reason to extend `TimelineFocus::PinnedEvents` rather than `TimelineFocus::Event`? We could just add a different `loader` type that loads a single event only based on a specified ID.

crates/matrix-sdk-ui/src/timeline/controller/mod.rs
```diff
@@ -474,16 +475,40 @@ impl<P: RoomDataProvider> TimelineController<P> {
...
+                        .load_event_with_relations(event_id, request_config, relations_filter)
+                        .await
+                    {
+                        Ok((event, related_events)) => {
+                            let mut events = vec![event];
+                            events.extend(related_events);
+                            events
+                        }
+                        Err(err) => {
+                            warn!("error when loading focussed event: {err}");
```
Hywan: If it's an error, we must have an `error!`, not a `warn!`.

Hywan: Also, if it fails, we want to reach `/context` to get a result. Please revisit this flow to handle this situation :-).

crates/matrix-sdk-ui/src/timeline/controller/mod.rs
Hywan: Is it related to my feedback about falling back to `/context` if the `if` branch fails?

Johennes: I think partially. If the `if` branch does _not_ fail, we'd still leave the paginator in initial state with the current implementation though. This might be ok, however, since pagination doesn't seem like a likely use case when loading no context at all.

crates/matrix-sdk-ui/src/timeline/controller/mod.rs
Johennes: Good points. Have addressed both with b4f5e1acee796a77749117c104701e04c3d9ab71.

Johennes:
> Can you add a test when num_context_events = 0 to test the new code please?

Have added two tests with e27d0d5b7e189c2176d44753dbaefcb457427951 / 37d56be98fcdca5d6f4d2728e02697cbcbd7309e.

crates/matrix-sdk-ui/src/timeline/controller/mod.rs
```diff
@@ -468,6 +468,15 @@ impl<P: RoomDataProvider> TimelineController<P> {
 
                 let event_paginator = Paginator::new(self.room_data_provider.clone());
 
+                let load_events_with_context = || async {
+                    // Start a /context request to load the focussed event and surrounding events.
```
Hywan: focussed -> focused

crates/matrix-sdk-ui/src/timeline/controller/mod.rs
```diff
@@ -487,17 +496,14 @@ impl<P: RoomDataProvider> TimelineController<P> {
                             events
                         }
                         Err(err) => {
-                            warn!("error when loading focussed event: {err}");
-                            vec![]
+                            error!("error when loading focussed event: {err}");
+                            // Fall back to load the focussed event using /context.
```
Hywan: focussed -> focused

crates/matrix-sdk-ui/src/timeline/controller/mod.rs
Hywan: Ok.

crates/matrix-sdk-ui/src/timeline/controller/mod.rs
Johennes:
```suggestion
                    // Start a /context request to load the focused event and surrounding events.
```

crates/matrix-sdk-ui/src/timeline/controller/mod.rs
Johennes:
```suggestion
                            // Fall back to load the focused event using /context.
```

Johennes:
Hm, I will have to look into why the tests suddenly fail now. I'm out on vacation though and will probably only be able to get back to this after the holidays, apologies.

Johennes:
@Hywan HNY and many apologies for the delay. I fixed the tests and rebased on current `main`. So this should be ready now.

Hywan (approved):

Hywan:
Thanks!

PR #5860, opened 2025-11-12
Use MSRV-aware resolver

jplatte:
After this, `cargo update` will take `rust-version` into account.

Signed-off-by: Jonas Platte


poljar:
Is that PR a draft for a reason?

PR #5861, opened 2025-11-12
Improve the wording of error messages when redecryption fails

andybalaam:
The previous message implied that we had received a session for this message, but that is only one of the several reasons we might encounter this situation. If redecryption failed, it is more likely we got here because we'd been asked to attempt redecryption for all UTDs e.g. when we build a new timeline.

Additionally, having similar wording for the error case and the unable to decrypt case could also cause confusion, so I adjusted the wording to make clear which situation is happening.


andybalaam:
Inspired by this rageshake https://github.com/element-hq/element-x-android-rageshakes/issues/8495

poljar (approved):

PR #5864, opened 2025-11-13
fix(ui): Sending read receipt in live timeline uses threaded event id instead

jmartinesp:
## Changes:

- In the live timeline, don't send read receipts for threaded events if `hide_thread_events` is enabled. `EventMeta` has now a `thread_root_id` field so we can know if the event is part of a thread, and `TimelineController::latest_event_id` filters those out if `hide_thread_events` is enabled.
  - Why do this instead of just filtering out non `visible` items? Because reactions, edits, replacements, etc. are also non-visible items that have to be taken into account for the read receipts. So it seems like we have 2 types of 'not visible' items in the timeline, those that aren't visible because we don't want to display them as separate items (aggregations), and those that aren't visible because they shouldn't be displayed at all.
  - Then why not avoid adding the items at all? Because some logic in the SDK depends on the threaded events being added to the remote events in the timeline, even if they aren't displayed because of `hide_threaded_events`, so some tests were broken.
- Fix test that was consistently failing locally on `main` - not really related to this PR, but it also made it fail in CI. It looks like a race condition of some kind.
- Modify `ffi::Timeline::latest_event_id` to use `ui::Timeline::latest_event_id`, which will return the value from `TimelineController::latest_event_id`.

Fixes https://github.com/element-hq/element-x-android/issues/5639.

Signed-off-by:

### DIFF
--- a/bindings/matrix-sdk-ffi/CHANGELOG.md
+++ b/bindings/matrix-sdk-ffi/CHANGELOG.md
@@ -85,6 +85,7 @@ All notable changes to this project will be documented in this file.
 
 ### Changes
 
+- `Timeline::latest_event_id` now uses its `ui::Timeline::latest_event_id` counterpart, instead of getting the latest event from the timeline and then its id.([#5864](https://github.com/matrix-org/matrix-rust-sdk/pull/5864))
 - Build Android ARM64 bindings using better default RUSTFLAGS (the same used for iOS ARM64). This should improve performance. [(#5854)](https://github.com/matrix-org/matrix-rust-sdk/pull/5854)
 
 ## [0.14.0] - 2025-09-04
--- a/crates/matrix-sdk-ui/CHANGELOG.md
+++ b/crates/matrix-sdk-ui/CHANGELOG.md
@@ -22,6 +22,7 @@ All notable changes to this project will be documented in this file.
 
 ### Bug Fixes
 
+- `Timeline::latest_event_id` won't take threaded events into account on live/event focused timelines if `hide_threaded_events` is enabled. This fixes a bug in `Timeline::mark_as_read` that incorrectly tried to send a read receipt for threaded events that aren't really part of those timelines. ([#5864](https://github.com/matrix-org/matrix-rust-sdk/pull/5864/))
 - Avoid replacing timeline items when the encryption info is unchanged.
   ([#5660](https://github.com/matrix-org/matrix-rust-sdk/pull/5660))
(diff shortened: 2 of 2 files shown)


crates/matrix-sdk-ui/tests/integration/room_list_service.rs
```diff
@@ -2993,6 +2993,8 @@ async fn test_thread_subscriptions_extension_enabled_only_if_server_advertises_i
                 },
             },
         };
+
+        mock_server.reset().await;
```
jmartinesp: This plus the change below fix the issue I was seeing locally, consistently making this test fail, and sometimes fail in CI. It seems like there is some weird race condition.

crates/matrix-sdk-ui/src/timeline/controller/metadata.rs
```diff
@@ -577,10 +577,16 @@ pub(in crate::timeline) struct EventMeta {
     /// Note that the #2 timeline item (the day divider) doesn't map to any
     /// remote event, but if it moves, it has an impact on this mapping.
     pub timeline_item_index: Option<usize>,
+
+    pub thread_root_id: Option<OwnedEventId>,
```
Hywan: Please add documentation for this field. Also —just a personal taste— I would put this field after `event_id`, not mandatory.

crates/matrix-sdk-ui/src/timeline/controller/mod.rs
```diff
@@ -1761,7 +1761,27 @@ impl TimelineController {
...
+            }
+            _ => true,
+        };
+        state
+            .items
+            .all_remote_events()
+            .iter()
+            .rev()
+            .filter_map(|item| {
+                if !filter_out_thread_events || item.thread_root_id.is_none() {
```
Hywan, line 1783: Can you explain with an inline comment here why we want to filter out event with a thread root?

crates/matrix-sdk-ui/src/timeline/mod.rs
```diff
@@ -268,6 +268,11 @@ impl Timeline {
         }
     }
 
+    /// Get the latest of the timeline's event ids.
```
Hywan: You may want to precise that, contrary to `latest_event`, this method only returns the ID of the last **remote** event (which is obvious for us because a local event doesn't have an event ID, only a transaction ID, but it's better to make it clear.

Hywan (approved):
Looks good to me, modulo the small feedback. Thank you for the new test and for the fix!

PR #5866, opened 2025-11-13
sdk: improve logging for received history bundles

richvdh:
We had an instance where a user joined a room on Element X but did not download the key bundle, so let's add some logging to help figure out what was going on.

### DIFF
--- a/crates/matrix-sdk/src/client/mod.rs
+++ b/crates/matrix-sdk/src/client/mod.rs
@@ -81,7 +81,7 @@ use ruma::{
 };
 use serde::de::DeserializeOwned;
 use tokio::sync::{Mutex, OnceCell, RwLock, RwLockReadGuard, broadcast};
-use tracing::{Instrument, Span, debug, error, instrument, trace, warn};
+use tracing::{Instrument, Span, debug, error, info, instrument, trace, warn};
 use url::Url;
 
 use self::futures::SendRequest;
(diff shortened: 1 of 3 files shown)


crates/matrix-sdk/src/client/mod.rs
```diff
@@ -1581,6 +1581,7 @@ impl Client {
         room_id: &RoomId,
         pre_join_room_info: Option<PreJoinRoomInfo>,
     ) -> Result<Room> {
+        info!(?pre_join_room_info, "Completing join to room {}", room_id);
```
Hywan:
```suggestion
        info!(?pre_join_room_info, ?room_id, "Completing joining the room");
```

I'm not sure about my English, but it's better to put `room_id` as an argument of the log, instead of formatting a string with the `room_id` inside it.

crates/matrix-sdk/src/room/shared_room_history.rs
```diff
@@ -128,6 +128,7 @@ pub(crate) async fn maybe_accept_key_bundle(room: &Room, inviter: &UserId) -> Re
         olm_machine.store().get_received_room_key_bundle_data(room.room_id(), inviter).await?
     else {
         // No bundle received (yet).
+        info!("No room key bundle from inviter found.");
```
Hywan:
```suggestion
        info!("No room key bundle from inviter found");
```

I think we don't put a “.” at the end of log sentences. Just a small details.

Hywan (approved):
Thanks!

A small suggestion, but I'm approving the patch ahead of time.

crates/matrix-sdk/src/client/mod.rs
richvdh: thanks, applied with light editing.

crates/matrix-sdk/src/room/shared_room_history.rs
richvdh: thanks, done.

PR #5868, opened 2025-11-14
crypto: Refer to MSC4362 when we are talking about encrypted state

andybalaam:
MSC4362 is now a grown-up MSC all on its own, not based on MSC3414, and what we have implemented is for MSC4362, so rename everything to refer to that.

Related: https://github.com/ruma/ruma/pull/2287


stefanceriu (approved):
👌

PR #5871, opened 2025-11-14
Spaces: Add methods to add/remove space children.

### LINKED ISSUE #5821: [Task] FFI: Add the ability to query and set the parents of a `Room`.
Unsure at this stage what querying should return (probably just room IDs) and whether or not it should be based on the graph within the `SpaceService` (which breaks any cycles) or should be the raw `m.space.parent` values.

pixlwave:
This PR makes 2 changes
- Add `SpaceService::add_child_to_space` and `SpaceService::remove_child_from_space` to allow for spaces to be edited.
- Add `SpaceService::joined_parents_of_child` to allow the editing of spaces from a room/subspace context (and for setting join rules).

Closes #5821

### DIFF
--- a/crates/matrix-sdk-ui/src/spaces/graph.rs
+++ b/crates/matrix-sdk-ui/src/spaces/graph.rs
@@ -67,6 +67,14 @@ impl SpaceGraph {
             .map_or(vec![], |node| node.children.iter().map(|id| id.as_ref()).collect())
     }
 
+    /// Returns the parents of a given node. If the node does not exist, it
+    /// returns an empty vector.
+    pub(super) fn parents_of(&self, node_id: &RoomId) -> Vec<&RoomId> {
+        self.nodes
+            .get(node_id)
+            .map_or(vec![], |node| node.parents.iter().map(|id| id.as_ref()).collect())
+    }
+
     /// Adds a node to the graph. If the node already exists, it does nothing.
     pub(super) fn add_node(&mut self, node_id: OwnedRoomId) {
         self.nodes.entry(node_id.clone()).or_insert(SpaceGraphNode::new(node_id));
@@ -184,8 +192,8 @@ mod tests {
 
         assert_eq!(graph.root_nodes(), vec![&a]);
 
-        assert!(graph.nodes[&b].parents.contains(&a));
-        assert!(graph.nodes[&c].parents.contains(&a));
+        assert_eq!(graph.parents_of(&b), vec![&a]);
+        assert_eq!(graph.parents_of(&c), vec![&a]);
 
         assert_eq!(graph.children_of(&a), vec![&b, &c]);
     }
--- a/crates/matrix-sdk-ui/src/spaces/mod.rs
+++ b/crates/matrix-sdk-ui/src/spaces/mod.rs
@@ -216,6 +216,20 @@ impl SpaceService {
         SpaceRoomList::new(self.client.clone(), space_id).await
     }
 
+    /// Returns all known direct-parents of a given space room ID.
+    pub async fn joined_parents_of_child(&self, child_id: &RoomId) -> Vec<SpaceRoom> {
+        let graph = &self.space_state.lock().await.graph;
+
+        graph
+            .parents_of(child_id)
+            .into_iter()
+            .filter_map(|parent_id| self.client.get_room(parent_id))
+            .map(|room| {
+                SpaceRoom::new_from_known(&room, graph.children_of(room.room_id()).len() as u64)
+            })
+            .collect()
+    }
+
     pub async fn add_child_to_space(
         &self,
         child_id: OwnedRoomId,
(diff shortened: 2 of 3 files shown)


crates/matrix-sdk-ui/src/spaces/mod.rs
```diff
@@ -204,6 +212,98 @@ impl SpaceService {
...
     }
 
+    /// Returns all known direct-parents of a given space room ID.
+    pub async fn joined_parents_of_child(&self, child_id: &RoomId) -> Vec<SpaceRoom> {
+        let graph = &self.space_state.lock().await.graph;
+
+        graph
+            .parents_of(child_id)
+            .into_iter()
+            .filter_map(|parent_id| self.client.get_room(parent_id)) // The graph is built from joined rooms so nothing should be filtered out here.
```
pixlwave: I'm like 90% confident in this assumption, happy to make this throw if that would be a better solution.

crates/matrix-sdk-ui/src/spaces/mod.rs
```diff
@@ -643,4 +939,27 @@ mod tests {
...
+                    factory.create(user_id!("@creator:example.com"), RoomVersionId::V1),
+                )
+                .add_state_event(
+                    factory.power_levels(&mut power_levels).state_key("").sender(user_id),
+                );
+
+            server.sync_room(client, builder).await;
+        }
+    }
 }
```
stefanceriu, line 973: I wonder if we should move the tests to a different file 🤔

crates/matrix-sdk-ui/src/spaces/graph.rs
```diff
@@ -184,8 +192,8 @@ mod tests {
 
         assert_eq!(graph.root_nodes(), vec![&a]);
 
-        assert!(graph.nodes[&b].parents.contains(&a));
-        assert!(graph.nodes[&c].parents.contains(&a));
+        assert_eq!(graph.parents_of(&b), vec![&a]);
+        assert_eq!(graph.parents_of(&c), vec![&a]);
```
stefanceriu, line 196: Why not make this change throughout the rest of the file? 🤔

crates/matrix-sdk-ui/src/spaces/mod.rs
stefanceriu: I think this is perfectly fine, would remove the comment.

crates/matrix-sdk-ui/src/spaces/mod.rs
```diff
@@ -204,6 +212,98 @@ impl SpaceService {
         SpaceRoomList::new(self.client.clone(), space_id).await
     }
 
+    /// Returns all known direct-parents of a given space room ID.
+    pub async fn joined_parents_of_child(&self, child_id: &RoomId) -> Vec<SpaceRoom> {
```
stefanceriu, line 220: I'm not sure whether `child` is the right word here, instead of maybe `room`. I wonder if `child` will get conflated with `m.space.child` which is not really the case as the graph holds reverse relations as well 🤔

crates/matrix-sdk-ui/src/spaces/mod.rs
```diff
@@ -204,6 +212,98 @@ impl SpaceService {
...
+            .parents_of(child_id)
+            .into_iter()
+            .filter_map(|parent_id| self.client.get_room(parent_id)) // The graph is built from joined rooms so nothing should be filtered out here.
+            .map(|room| {
+                SpaceRoom::new_from_known(&room, graph.children_of(room.room_id()).len() as u64)
+            })
+            .collect()
+    }
+
+    pub async fn add_child_to_space(
```
stefanceriu, line 233: Same `child` vs. `room` as above here too.

crates/matrix-sdk-ui/src/spaces/mod.rs
```diff
@@ -204,6 +212,98 @@ impl SpaceService {
...
+            && let Ok(power_levels) = child_room.power_levels().await
+            && power_levels.user_can_send_state(user_id, StateEventType::SpaceParent)
+        {
+            let parent_route = space_room.route().await.map_err(Error::UpdateRelationship)?;
+            child_room
+                .send_state_event_for_key(&space_id, SpaceParentEventContent::new(parent_route))
+                .await
+                .map_err(Error::UpdateRelationship)?;
+        } else {
+            warn!("The current user doesn't have permission to set the child's parent.");
```
stefanceriu, line 263: We should really bubble this error up or maybe or maybe add a parameter to have the client tell us what they expect?

stefanceriu, line 263: Also, this else branch doesn't necessarily mean the user doesn't have permissions, it might've failed to fetch the power levels.

crates/matrix-sdk-ui/src/spaces/mod.rs
```diff
@@ -204,6 +212,98 @@ impl SpaceService {
...
+            .await
+            .is_err()
+        {
+            return Err(Error::MissingState(StateEventType::SpaceChild, child_id));
+        }
+        // Redacting state is a "weird" thing to do, so send {} instead.
+        //
+        // Specifically, "The redaction of the state doesn't participate in state
+        // resolution so behaves quite differently from e.g. sending an empty form of
+        // that state events".
```
stefanceriu, line 292: Any spec docs or msc/draft/wtvr you can point this at?

crates/matrix-sdk-ui/src/spaces/mod.rs
```diff
@@ -204,6 +212,98 @@ impl SpaceService {
...
+            .is_err()
+        {
+            warn!("A space parent event wasn't found on the child room, ignoring.");
+            return Ok(());
+        }
+        // Same as the comment above.
+        child_room
+            .send_state_event_raw("m.space.parent", space_id.as_str(), serde_json::json!({}))
+            .await
+            .map_err(Error::UpdateRelationship)?;
```
stefanceriu, line 310: Why is one branch fallible but other not? Isn't it valid to have an `m.space.parent` and no `m.space.child`?

crates/matrix-sdk-ui/src/spaces/mod.rs
```diff
@@ -620,6 +722,200 @@ mod tests {
...
+                            .space_child(parent_space_id_2.to_owned(), child_space_id.to_owned())
+                            .sender(user_id),
+                    ),
+            )
+            .await;
+
+        let space_service = SpaceService::new(client.clone());
+
+        // Wait for the space hierarchy to register.
+        _ = space_service.joined_spaces().await;
```
stefanceriu, line 803: Should we just move the subscription logic to a separate `setup` method? That's what I did for the new GroupedRoomList service and honestly I don't have a better idea at the moment 🤔

crates/matrix-sdk-ui/src/spaces/graph.rs
pixlwave: This was the only test that was using `graph.children_of` (all the others use `nodes.children`) so I thought you had a specific reason and followed the pattern 🤷‍♂️

crates/matrix-sdk-ui/src/spaces/mod.rs
pixlwave: I went with `parent` and `child` here because it could be a space and a subspace (or even a room as the parent of a space because they're just IDs).

crates/matrix-sdk-ui/src/spaces/mod.rs
pixlwave: Here I assumed that you would only call it with an actual space as adding a child to a room is stupid.

crates/matrix-sdk-ui/src/spaces/mod.rs
pixlwave: That sounds fair, but I think I'll leave that exercise to you (or me in the next PR if I need it for sharing the graph logic).

crates/matrix-sdk-ui/src/spaces/mod.rs
pixlwave: We discussed this. It's hard to know what the caller would expect with a single method, so for now we'll leave it as is. In the future we'll potentially add some parameters and can throw based on those.

crates/matrix-sdk-ui/src/spaces/mod.rs
pixlwave: My bad, will update to `warn!` on an unset event for both directions of the relationship.

PR #5872, opened 2025-11-17
xtask: Add support for building watchOS targets.

pixlwave:
This PR makes the following changes to the Swift xtask:

- Add support for building Tier 3 targets (by using nightly and `-Zbuild-std`).
- Add the [watchos targets](https://doc.rust-lang.org/nightly/rustc/platform-support/apple-watchos.html) (I ignored `armv7k-apple-watchos` as the last supported device with that was the Series 3 which is became unsupported as of watchOS 9).
- Add a CLI flag to include the tier 3 targets when building all targets (which are ignored by default).
- Add a linker workaround for watchOS simulator builds.

### DIFF
--- a/bindings/matrix-sdk-ffi/build.rs
+++ b/bindings/matrix-sdk-ffi/build.rs
@@ -43,6 +43,23 @@ fn setup_x86_64_android_workaround() {
     }
 }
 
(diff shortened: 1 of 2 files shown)


poljar (approved):
Alright, can't spot anything wrong with it and I trust you didn't break your own build 😅

PR #5873, opened 2025-11-17
base: Bump ruma to 91424b1fc

andybalaam:
And update to reflect the new feature name `unstable-msc4362`, which provides the new unstable prefix `io.element.msc4362.encrypt_state_events`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4805,7 +4805,7 @@ dependencies = [
 [[package]]
 name = "ruma"
 version = "0.13.0"
-source = "git+https://github.com/ruma/ruma?rev=15307f0eb361d7e04cfa33602fdffbf10c54a173#15307f0eb361d7e04cfa33602fdffbf10c54a173"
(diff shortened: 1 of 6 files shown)


crates/matrix-sdk-ui/src/timeline/tests/encryption.rs
```diff
@@ -928,7 +928,9 @@ async fn test_retry_decryption_updates_reply() {
 
     // We receive the UTD.
     {
-        let event = assert_next_matches!(stream, VectorDiff::PushBack { value } => value);
+        let event =
+            assert_next_matches_with_timeout!(stream, VectorDiff::PushBack { value } => value);
```
stefanceriu, line 932: This feels somewhat unrelated, is it intentional?

andybalaam: It is intentional - it was failing for me when I was testing this. I included it in a separate commit - can be a separate PR if you like.

stefanceriu: No no, it's fine, just checking 👍

PR #5874, opened 2025-11-18
refactor: Make `ClientBuilder` safer to use when restoring a session

jmartinesp:
Changes:

- Add some docs to discourage using the `ClientBuilder` methods that modify the server or homeserver URLs when restoring an existing session.
- Removes the unused `ffi::Session::homeserver_url` field, since it can lead to clients believing that modifying that value may have an effect in the SDK behaviour.

- [ ] Public API changes documented in changelogs (optional)

Signed-off-by:

PR #5876, opened 2025-11-18
Revert "refactor(ffi): Remove unused Session::homeserver_url value"

jmartinesp:
This reverts commit 4eb3cc9.Reverts matrix-org/matrix-rust-sdk#5874

Turns out this didn't make the `ClientBuilder` safer but completely broke it when the server is not reachable and we try to restore a session. The complement crypto PR is: https://github.com/matrix-org/complement-crypto/pull/217.

### DIFF
--- a/bindings/matrix-sdk-ffi/src/client_builder.rs
+++ b/bindings/matrix-sdk-ffi/src/client_builder.rs
@@ -221,54 +221,18 @@ impl ClientBuilder {
         Arc::new(builder)
     }
 
-    /// Set the server name to discover the homeserver from.
-    ///
-    /// We assume we can connect in HTTPS to that server.
-    ///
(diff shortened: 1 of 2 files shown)


jmartinesp:
We use it in the clients, to be able to save it locally to our own DB, which we will later read from and use this info to restore the client.

I thought we didn't need this anymore, but it seems like I completely misunderstood how this is used. The problem is now we need something that allows us to restore the server name and the HS url and not trigger a HS discovery request if we already had some server info.

However, I'm not sure how feasible this is at the moment since we don't have a way to check the `StateStore` before the parent `Client` is created if I'm not mistsaken. Maybe we could create an instance of it, use it just for this check, then drop it before creating the `Client`.

---

Well, that isn't possible because `BaseStateStore` is only present in the `matrix-sdk-base` crate...

Issue #5862, opened 2025-11-12
SlidingSync pos restoration and intial request timeouts


stefanceriu:
The introduction of storage for the SlidingSync position makes it more obvious that the intial request needs to go out without a timeout (i.e. zero, the default), as opposed to the 30 seconds that are used at the moment. 

If the backend has no new data when restoring a session from a given position then the RoomListService state machine will be stuck between Initial and SettingUp until the connection times out and as a result publish unexpected values on the SyncIndicator stream.

The current timeout is based on the whether any of the inner sliding sync lists are in Selective mode (default config) which wasn't visible until https://github.com/matrix-org/matrix-rust-sdk/pull/5852 made the room list not go directly into Growing anymore.

One idea would be to take out the timeout from the lists and instead let SlidingSync users (e.g. the RoomListService) configure them based on their own state.

PR #5878, opened 2025-11-18
fix(ui): Change how `SlidingSync` enables long-polling

Hywan:
This patch changes how `SlidingSync` enables long-polling. Long-polling with sliding sync is based on a `timeout` query parameter. If absent or 0, the server is expected to return immediately, otherwise the server waits up to _timeout_ milliseconds before returning a response.

Before this patch, to define whether long-polling was necessary, we were relying on this simple algorithm (applied onto each sliding sync list):

- if the list is `Selective` (i.e. it has a constant range), or if the list is fully loaded, then do long-polling,
- otherwise, do not do long-polling.

This behaviour has been introduced in https://github.com/matrix-org/matrix-rust-sdk/pull/3853.

As pointed in https://github.com/matrix-org/matrix-rust-sdk/issues/5862, this algorithm is a bit annoying with our today's use of `SlidingSync`. Why? Because sometimes the server has nothing new to return despite the list is not `Selective` and is not fully-loaded. Thus, we need a more customizable mechanism, hence this set of patches.

First off, we introduce a new `SlidingSyncListBuilder::requires_timeout` method that accepts a user-defined function controlling whether a timeout is required or not, i.e. if the sliding sync client must do long-polling or not. The default function mimics the existing behaviour for compatibility purposes.

Second, `RoomListService` defines its own function, and uses it.

Finally, the test suite has been updated to test this new behaviour.

---

* Fixes https://github.com/matrix-org/matrix-rust-sdk/issues/5862

### DIFF
--- a/crates/matrix-sdk-ui/tests/integration/room_list_service.rs
+++ b/crates/matrix-sdk-ui/tests/integration/room_list_service.rs
@@ -73,6 +73,7 @@ macro_rules! sync_then_assert_request_and_fake_response {
         [$server:ident, $room_list:ident, $stream:ident]
         $( states = $pre_state:pat => $post_state:pat, )?
         $( assert pos $pos:expr, )?
+        $( assert timeout $timeout:expr, )?
         assert request $assert_request:tt { $( $request_json:tt )* },
         respond with = $( ( code $code:expr ) )? { $( $response_json:tt )* }
         $( , after delay = $response_delay:expr )?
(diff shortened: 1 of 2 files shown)


crates/matrix-sdk-ui/src/room_list_service/mod.rs
```diff
@@ -222,7 +225,23 @@ impl RoomListService {
...
+                            // server, with no long-polling.
+                            State::Init
+                            | State::SettingUp
+                            | State::Recovering
+                            | State::Error { .. }
+                            | State::Terminated { .. } => false,
+
+                            // Otherwise we want long-polling if the list is fully-loaded.
+                            State::Running => request_generator.is_fully_loaded(),
+                        }
```
stefanceriu, line 243: This ended up being so nice and clean 👏

Hywan: Yup!

PR #5881, opened 2025-11-20
Fix a race condition in the redecryptor leading to missed decryption attempts

### LINKED ISSUE #5474: UTDs not retried when key arrives at almost the same time

poljar:
This PR fixes a race condition in the redecryptor.

Namely, events and room keys are received over two different sync streams from the homeserver. When events are received over the sync, we first try to decrypt them, this might fail because the room key hasn't yet arrived over the other sync stream. The event cache will then persist the event as a UTD.

At the same time, the redecryptor will listen to room keys that arrive on the other sync stream. Once the redecryptor gets notified about a room key, it will attempt to fetch the event from the event cache to decrypt the event and replace it.

Crucially if the key arrives before the event gets persisted but after the initial decryption attempt we might never attempt to redecrypt such an event.

The fix is relatively straightforward, we'd need a synchronization point for the two different tasks, the event cache which adds events and the redecryptor which listens to room keys to decrypt events.

A lock could have been used, so the storing and redecrypting of events becomes synchronized via the storage layer. This approach could have degraded performance since the event cache needs to handle a lot of events.

The approach that was chosen here is to let the redecryptor listen to updates coming from the event cache itself. If the event cache tells us that it persisted a UTD, we will attempt to decrypt. Upon a successful decryption we will replace the event in the cache as well.

A review commit by commit is probably the easiest way to handle this.

This closes #5474.

### DIFF
--- a/crates/matrix-sdk/src/event_cache/redecryptor.rs
+++ b/crates/matrix-sdk/src/event_cache/redecryptor.rs
@@ -46,6 +46,11 @@
 //! received and R2D2 won't get notified about it. To work around this
 //! decryption requests can be explicitly sent to R2D2.
 //!
+//! The final gotcha is that a room key might be received just in between the
+//! time the event was initially tried to be decrypted and the time it took to
+//! persist it in the event cache. To handle this race condition R2D2 listens to
+//! the event cache and attempts to decrypt any UTDs the event cache persists.
+//!
 //! In the graph bellow the Timeline block is meant to be the `Timeline` from
 //! the `matrix-sdk-ui` crate, but it could be any other listener that
 //! subscribes to [`RedecryptorReport`] stream.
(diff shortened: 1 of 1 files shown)


andybalaam (approved):
Looks good. Excited about this fix.

crates/matrix-sdk-common/src/linked_chunk/updates.rs
```diff
@@ -114,6 +114,31 @@ pub enum Update<Item, Gap> {
...
+impl<Item, Gap> Update<Item, Gap> {
+    /// Get the items from the [`Update`] if any.
+    ///
+    /// This function is useful if you only care about the items from the
+    /// [`Update`] and not what kind of update it was and where the items
+    /// should be placed.
+    ///
+    /// [`Update`] variants which don't contain any items will return an empty
+    /// [`Vec`].
+    pub fn into_items(self) -> Vec<Item> {
```
Hywan, line 126: Makes sense to me.

Hywan (approved):
Looks good to me. Well done! Finding the race condition was non-trivial…

PR #5885, opened 2025-11-21
feat(ui,ffi): Add `LatestEventValue::Local { sender, profile, .. }`.

Hywan:
This patch adds a `sender: OwnedUserId` and a `profile: TimelineDetails<Profile>` fields to `LatestEventValue::Local` in `matrix_sdk_ui::timeline` (and the corresponding `matrix_sdk_ffi` type).

---

### DIFF
--- a/crates/matrix-sdk-ui/CHANGELOG.md
+++ b/crates/matrix-sdk-ui/CHANGELOG.md
@@ -22,6 +22,9 @@ All notable changes to this project will be documented in this file.
   an event.
 - `TimelineFocusKind::Event` can now handle both the existing event pagination and thread pagination if the focused 
   event is part of a thread ([#5678](https://github.com/matrix-org/matrix-rust-sdk/pull/5678)).
+- [**breaking**] The `Room` type in `room_list_service` is renamed to
+  `RoomListItem`.
+  ([#5684](https://github.com/matrix-org/matrix-rust-sdk/pull/5684))
 
 ### Bug Fixes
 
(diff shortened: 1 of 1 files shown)


bindings/matrix-sdk-ffi/src/timeline/mod.rs
```diff
@@ -1304,6 +1304,7 @@ pub enum LatestEventValue {
     },
     Local {
         timestamp: Timestamp,
+        profile: ProfileDetails,
```
poljar, line 1308: That's a breaking change isn't it? Don't we want to advertise this in the changelog of the FFI crate?

crates/matrix-sdk-ui/src/timeline/latest_event.rs
```diff
@@ -54,6 +54,9 @@ pub enum LatestEventValue {
         /// The timestamp of the local event.
         timestamp: MilliSecondsSinceUnixEpoch,
 
+        /// The sender's profile.
+        profile: TimelineDetails<Profile>,
+
```
poljar, line 62: Same for the UI crate, that's a breaking change?

crates/matrix-sdk-ui/src/timeline/latest_event.rs
```diff
@@ -226,8 +235,9 @@ mod tests {
         let value =
             LatestEventValue::from_base_latest_event_value(base_value, &room, &client).await;
 
-        assert_matches!(value, LatestEventValue::Local { timestamp, content, is_sending } => {
+        assert_matches!(value, LatestEventValue::Local { timestamp, profile, content, is_sending } => {
```
poljar: Not sure how much effort it would be, but don't we want to have test with an available profile?

poljar (approved):
Looks good, there are some missing changelogs. Approving ahead of time.

bindings/matrix-sdk-ffi/src/timeline/mod.rs
Hywan: Yup, added, thanks!

crates/matrix-sdk-ui/src/timeline/latest_event.rs
Hywan: Correct, added!

crates/matrix-sdk-ui/src/timeline/latest_event.rs
Hywan: The `profile_from_user_id` method is tested elsewhere, and it's not setup in these test cases. However, I've added the `sender` field too, and this one is tested, which increases the trust in the code (except if someone computes a different `UserId` in the meantime, but it's pretty unlikely).

PR #5890, opened 2025-11-24
Spaces: Fix removing a space child when the relationship is only 1-way.

pixlwave:
Seconds before #5871 auto-merged, I noticed that removing a child from a space where only the m.space.parent was set would never work as I copy & pasted an early return to fix https://github.com/matrix-org/matrix-rust-sdk/pull/5871#discussion_r2546373677 instead of using an `if { } else { }`.

Then I noticed there were no tests for removing the child…

Then I noticed that the removal was checking if the `m.space.child`/`parent` result was `Ok` without checking if the event was `Some`.

Long story short, this PR fixes it all 🤦‍♂️

### DIFF
--- a/crates/matrix-sdk-ui/src/spaces/mod.rs
+++ b/crates/matrix-sdk-ui/src/spaces/mod.rs
@@ -276,10 +276,8 @@ impl SpaceService {
         let child_room =
             self.client.get_room(&child_id).ok_or(Error::RoomNotFound(child_id.to_owned()))?;
 
-        if space_room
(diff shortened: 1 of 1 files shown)


stefanceriu:
Benchmarks are failing with a 503 on `curl -fsSL https://github.com/CodSpeedHQ/runner/releases/download/v$RUNNER_VERSION/codspeed-runner-installer.sh | bash -s -- --quiet` (duh) so I'll just go ahead and merge it.

PR #5895, opened 2025-11-25
misc: Add support for bridge spans

jmartinesp:
Bridge spans will use `bridge_trace_id` to map an existing client transactions/spans to Rust ones so they'll be displayed as a single one in Sentry.

This is done through the `sentry.trace` field, which will be used by `sentry-tracing` to differentiate these kinds of special spans (see [docs](https://github.com/getsentry/sentry-rust/tree/master/sentry-tracing#special-span-fields)).

The special fields need to be added on the Span creation, that's why we do it in the constructor instead of just using `span.record(...)` later.

Also bumps Sentry SDK to the latest version.

---

Note this doesn't seem to work 100% of the time: 
- <del>It might be some bug in the Sentry SDK.</del>
- <del>Ot a race condition that happens when the SDK spans are received before the client ones.</del>
- <del>Or maybe our Sentry instance has a bug since it's a bit out of date.</del>

Until all the code that runs using that span has finished (including spawned long running tasks) Sentry doesn't consider the span as ready to upload. We can't force sentry to send the span ahead of time (at least, there is no way that I know given how `sentry-tracing` works), so in those cases the spans just won't be uploaded.

The result is sometimes you get flawless results, everything is uploaded and works smoothly, and other times no matter how many times you try, the SDK spans won't be uploaded/attached. However, the time metrics for the spans won't change, so not having the SDK spans, while not ideal, it's not mandatory to have consistent metrics.

Signed-off-by:

### DIFF
--- a/bindings/matrix-sdk-ffi/src/platform.rs
+++ b/bindings/matrix-sdk-ffi/src/platform.rs
@@ -3,8 +3,6 @@ use std::sync::OnceLock;
 use std::sync::{atomic::AtomicBool, Arc};
 
 #[cfg(feature = "sentry")]
-use sentry::MaxRequestBodySize;
-#[cfg(feature = "sentry")]
 use tracing::warn;
 use tracing_appender::rolling::{RollingFileAppender, Rotation};
 #[cfg(feature = "sentry")]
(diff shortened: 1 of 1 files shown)


Hywan (approved):
LGTM

PR #5897, opened 2025-11-25
Spaces: Add support for getting a flattened list of editable spaces.

pixlwave:
This PR makes 2 changes:
- Refactors the tests for Spaces to use a common `add_space_rooms` function instead of having 3 variants.
- Adds a new method to the `SpaceService` to get all of the user's joined spaces where they have the power to edit the spaces' children.

### DIFF
--- a/bindings/matrix-sdk-ffi/src/spaces.rs
+++ b/bindings/matrix-sdk-ffi/src/spaces.rs
@@ -78,6 +78,15 @@ impl SpaceService {
         })))
     }
 
+    /// Returns a flattened list containing all the spaces where the user has
+    /// permission to send `m.space.child` state events.
+    ///
+    /// Note: Unlike [`Self::joined_spaces()`], this method does not recompute
+    /// the space graph, nor does it notify subscribers about changes.
+    pub async fn editable_spaces(&self) -> Vec<SpaceRoom> {
+        self.inner.editable_spaces().await.into_iter().map(Into::into).collect()
+    }
+
     /// Returns a `SpaceRoomList` for the given space ID.
     pub async fn space_room_list(
         &self,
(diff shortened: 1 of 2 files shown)


bindings/matrix-sdk-ffi/src/spaces.rs
pixlwave, line 1: Note: I'm open to naming suggestions as I wasn't massively confident on what to call this (and the matrix_sdk) function.

crates/matrix-sdk-ui/src/spaces/mod.rs
```diff
@@ -211,6 +211,33 @@ impl SpaceService {
         spaces
     }
 
+    /// Returns a flattened list containing all joined spaces where the
+    /// user has permission to send `m.space.child` state events.
+    ///
+    /// Note: Unlike [`Self::joined_spaces()`], this method does not recompute
+    /// graph, nor does it notify subscribers about changes.
+    pub async fn all_editable_joined_spaces(&self) -> Vec<SpaceRoom> {
```
poljar: One way to make this name a bit more pleasant is to remove the `all` prefix, and the joined nominator.

I think it's a given that you can only edit joined spaces or rooms for that matter.

So `editable_spaces()` sounds more palatable to me.

poljar (changes requested):
Looks mostly good, I think I found a bit of a better name for this function.

Let me know what you think.

crates/matrix-sdk-ui/src/spaces/mod.rs
pixlwave: > I think it's a given that you can only edit joined spaces or rooms for that matter.

Oh yeah good point, totally missed that fact 😆

pixlwave: [Updated](https://github.com/matrix-org/matrix-rust-sdk/compare/e4b17ada000df5857de874fb834a19195215a48b..61dbc07144d5e63fa4b4ffb24134bbc30c440e88)

poljar (approved):

Issue #3065, opened 2024-01-29
Filter RRs on specific event types.


Velin92:
On some clients, like Element X we would like to still keep state events, but not show read receipts on them, however this would create an issue where RRs would magically disappear. Since we are working on a filter that *hides* the state event, but is capable of moving the RR on the closest event, would be possible to improve this API further (or create a similar one) were the filtered event *does not hide but has its RR move to the closest non excluded event* so that we can have RRs only show for Messages  if we want on a specific client.

bnjbvr:
Thanks for opening an issue, I'll take a stab at it.

bnjbvr:
Fixed with Doug's PR

PR #5900, opened 2025-11-26
Read Receipts: Allow Timelines to be configured to hide read receipts on state events.

pixlwave:
This is an mvp attempt at #3065 to fix https://github.com/element-hq/element-x-ios/issues/2338.

The PR adds a new `TimelineReadReceiptTracking` enum which is used in place of the `track_read_receipts` boolean to decide whether to enable tracking for all events, just for message-like events or to disable tracking.

When tracking is configured for message-like only, any read receipts on state events are bundled up into the previous event, using the same logic that receipts from filtered events are handled with.

| Before | After |
| - | - |
| Screenshot 2025-11-26 at 9 53 19 am | Simulator Screenshot - iPhone 17 - 2025-11-26 at 09 53 22 |

### DIFF
--- a/crates/matrix-sdk-ui/src/timeline/controller/state_transaction.rs
+++ b/crates/matrix-sdk-ui/src/timeline/controller/state_transaction.rs
@@ -470,10 +470,10 @@ impl<'a, P: RoomDataProvider> TimelineStateTransaction<'a, P> {
     ) -> bool {
         match event {
             AnySyncTimelineEvent::State(_) => {
-                settings.track_read_receipts == TimelineReadReceiptTracking::AllEvents
+                matches!(settings.track_read_receipts, TimelineReadReceiptTracking::AllEvents)
             }
             AnySyncTimelineEvent::MessageLike(_) => {
-                settings.track_read_receipts != TimelineReadReceiptTracking::Disabled
+                !matches!(settings.track_read_receipts, TimelineReadReceiptTracking::Disabled)
             }
         }
     }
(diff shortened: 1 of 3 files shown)


crates/matrix-sdk-ui/src/timeline/controller/state_transaction.rs
```diff
@@ -468,8 +469,12 @@ impl<'a, P: RoomDataProvider> TimelineStateTransaction<'a, P> {
         event: &AnySyncTimelineEvent,
     ) -> bool {
         match event {
-            AnySyncTimelineEvent::State(_) => settings.state_events_can_show_read_receipts,
-            AnySyncTimelineEvent::MessageLike(_) => true,
+            AnySyncTimelineEvent::State(_) => {
+                settings.track_read_receipts == TimelineReadReceiptTracking::AllEvents
```
Hywan: It's probably better to use `matches!(settings.track_read_receipts, TimelineReadReceiptTracking::AllEvents)` here, so that you can remove the `PartialEq` implementation on this enum. Less code.

pixlwave: Ah nice, updated here: https://github.com/matrix-org/matrix-rust-sdk/compare/4b903d041a6d4ddcc864a42eb258851c58d383a7..fdc747962df908135935f6cfb8b071628dfc46fa

pixlwave:
Force pushed to fix a conflict on the changelog.

PR #5905, opened 2025-11-27
fix: The search crate doesn't actually depend on the main crate

poljar:
This removes a circular dependency we had resulting in a semi-broken release process.


Hywan (approved):

poljar:
So that's where the dependency lies, in the docs.

PR #5911, opened 2025-11-28
feat(sdk): Add Client::optimize_stores

jmartinesp:
Added methods to perform cleanup/maintenance/optimization actions in the stores, so we can manually trigger them from the clients. This is a first attempt to understand if running these optimizations fixes an observed issue in the clients where the DB performance degrades severely as the cache grows.

Also added another set of methods to calculate the store sizes, when possible (`Client::get_store_sizes` + `*Store::get_size`).

cc @Hywan 

Signed-off-by:

### DIFF
--- a/bindings/matrix-sdk-ffi/CHANGELOG.md
+++ b/bindings/matrix-sdk-ffi/CHANGELOG.md
@@ -6,6 +6,10 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased] - ReleaseDate
 
+### Features
+
+- Add `Client::get_store_sizes()` so to query the size of the existing stores, if available. ([#5911](https://github.com/matrix-org/matrix-rust-sdk/pull/5911))
+
 ### Breaking changes
 
 - `TimelineConfiguration::track_read_receipts`'s type is now an enum to allow tracking to be enabled for all events
--- a/crates/matrix-sdk/CHANGELOG.md
+++ b/crates/matrix-sdk/CHANGELOG.md
@@ -6,6 +6,10 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased] - ReleaseDate
 
+### Features
+
+- Add `Client::get_store_sizes()` so to query the size of the existing stores, if available. ([#5911](https://github.com/matrix-org/matrix-rust-sdk/pull/5911))
+
 ### Refactor
 
 - [**breaking**]: `Client::server_vendor_info()` requires to enable the


jmartinesp:
I kept the `doc_hidden` annotations in the `optimize_stores` methods, let me know if we still want to do that or just having the warning (maybe an improved version of it?) is good enough.

crates/matrix-sdk/src/client/mod.rs
```diff
@@ -3218,6 +3074,73 @@ impl Client {
...
+    ///
+    /// **Warning:** this was added to check if SQLite fragmentation was the
+    /// source of performance issues, **DO NOT use in production**.
+    #[doc(hidden)]
+    pub async fn optimize_stores(&self) -> Result<()> {
+        trace!("Optimizing state store...");
+        self.state_store().optimize().await?;
+
+        trace!("Optimizing event cache store...");
+        self.event_cache_store().lock().await?.as_clean().unwrap().optimize().await?;
```
poljar: Won't this panic in case the event cache lock is dirty and we try to optimize the store?

poljar (changes requested):
I think there's a panic in there.

poljar:
> I kept the `doc_hidden` annotations in the `optimize_stores` methods, let me know if we still want to do that or just having the warning (maybe an improved version of it?) is good enough.

I think it's fine to keep it hidden for now. Not that it matters one way or the other since only people that implement the traits will be interested in this and they will have to know about it.

crates/matrix-sdk/src/client/mod.rs
jmartinesp: Oops. Fixed in https://github.com/matrix-org/matrix-rust-sdk/pull/5911/commits/9eb04ed31e29a3fd137ad7a278cdd762c9a0cd44 using the same check than for the `get_store_sizes` function instead.

Hywan: The `as_clean()` is supposed to be used only for testing. It's better to pattern-match the lock state.

bindings/matrix-sdk-ffi/src/client.rs
```diff
@@ -359,6 +359,17 @@ impl Client {
...
 #[matrix_sdk_ffi_macros::export]
 impl Client {
+    /// Perform database optimizations if any are available, i.e. vacuuming in
+    /// SQLite.
+    pub async fn optimize_stores(&self) -> Result<(), ClientError> {
+        Ok(self.inner.optimize_stores().await?)
+    }
+
+    /// Returns the sizes of the existing stores, if known.
+    pub async fn get_store_sizes(&self) -> Result<StoreSizes, ClientError> {
```
Hywan, line 369: We usually don't have a `get_` prefix, only a `set_` prefix. It's not the end of the world, but so you know, it's more Rust idiomatic :-).

PR #5913, opened 2025-11-28
feat(sdk,ui): Support edits as `LatestEventValue`

Hywan:
This patch adds the ability for an edit (`m.room.message` with a `m.relates_to` + `m.replace`) to be considered a `LatestEventValue` candidate if and only if it edits the immediate previous event. Otherwise it's easy to edit an old message and create a “broken” `LatestEventValue` because it points to an older message that the user may not be able to find easily.

---

* Fix https://github.com/element-hq/element-x-android/issues/5633
* Fix https://github.com/element-hq/element-x-ios/issues/2683

### DIFF
--- a/crates/matrix-sdk/src/event_cache/room/mod.rs
+++ b/crates/matrix-sdk/src/event_cache/room/mod.rs
@@ -322,11 +322,14 @@ impl RoomEventCache {
     /// Try to find a single event in this room, starting from the most recent
     /// event.
     ///
+    /// The `predicate` receives two arguments: the current event, and the
+    /// ID of the _previous_ (older) event.
+    ///
     /// **Warning**! It looks into the loaded events from the in-memory linked
     /// chunk **only**. It doesn't look inside the storage.
     pub async fn rfind_map_event_in_memory_by<O, P>(&self, predicate: P) -> Result<Option<O>>
     where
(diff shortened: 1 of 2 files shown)


pixlwave:
Tested on EXI, this works and I love it! 🫶

pixlwave:
tenor

crates/matrix-sdk-ui/src/timeline/latest_event.rs
```diff
@@ -102,11 +108,50 @@ impl LatestEventValue {
...
+                        ..
+                    }) => {
+                        // Let's map the edit into a regular message.
+                        match TimelineAction::from_content(
+                            AnyMessageLikeEventContent::RoomMessage(RoomMessageEventContent::new(
+                                new_content.msgtype,
+                            )),
+                            // We don't care about the `InReplyToDetails` in the context of a
+                            // `LatestEventValue`.
+                            None,
```
networkException, line 132: Setting `in_reply_to` to `None` here disables reply fallback sanitization in https://github.com/matrix-org/matrix-rust-sdk/blob/238e4e8a87baad51bcfd44c619f0caa985472cc3/crates/matrix-sdk-ui/src/timeline/event_item/content/mod.rs#L201-L202

This currently makes EXA render reply fallbacks in the room list

PR #5916, opened 2025-12-02
feat(sdk): Support refreshing the access token in Client::fetch_server_versions

zecakeh:
We need to handle 2 possible deadlocks for this:

1. We cannot try to refresh an expired access token if this call happens
   while we are currently trying to refresh the token. The easiest way
   to handle this is to never try to refresh the token when making this
   call inside `get_path_builder_input()` so we implement a "failsafe"
   mode that disables refreshing the access token in case it expired.
   However it attempts the GET /versions request again without the token.
2. We cannot access the cached supported versions if we are in the
   process of refreshing that cache because the RwLock has a write lock.
   So if the access token has expired and we try to refresh it, the
   possible calls to `get_path_builder_input()` must not wait for a read
   lock to be available. So the solution is to never wait for a read
   lock, and skip the cache if a read lock is not available.

This also gets rid of workarounds in other functions.


andybalaam (approved):
Very clear, thanks.

PR #5917, opened 2025-12-02
feat(sdk): Add support for the stable `m.oauth` UIAA type

zecakeh:
By replacing the custom implementation with the one in Ruma.

It landed recently in the spec: https://github.com/matrix-org/matrix-spec/pull/2234.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -66,7 +66,7 @@ rand = "0.8.5"
 regex = "1.12.2"
 reqwest = { version = "0.12.24", default-features = false }
 rmp-serde = "1.3.0"
-ruma = { version = "0.14.0", features = [
+ruma = { git = "https://github.com/ruma/ruma", rev = "a67081e402dce14365089b34f50489dacc9c53b5", features = [
     "client-api-c",
     "compat-upload-signatures",
     "compat-arbitrary-length-ids",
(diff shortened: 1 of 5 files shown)


Cargo.toml
```diff
@@ -66,7 +66,7 @@ rand = "0.8.5"
 regex = "1.12.2"
 reqwest = { version = "0.12.24", default-features = false }
 rmp-serde = "1.3.0"
-ruma = { version = "0.14.0", features = [
+ruma = { git = "https://github.com/ruma/ruma", rev = "a67081e402dce14365089b34f50489dacc9c53b5", features = [
```
poljar, line 69: We still didn't get the release out because of https://github.com/matrix-org/matrix-rust-sdk/pull/5923. So this will need to wait for a bit.

zecakeh: How long will it take? Because we could do a "patch" release of ruma if necessary.

poljar: It should happen this week, I still need to write a small patch before we can release though.

zecakeh: Okay, let's wait then.

jmartinesp:
AFAICT this worked fine when tested on Element X Android, although I'm not sure I tested the right thing. Is logging using OIDC enought?

zecakeh:
To test this code you need to reset the cross-signing keys. In the bindings it is exposed as `Encryption::reset_identity()` I believe.

jmartinesp:
Ok, I re-tested this by resetting the identity in one of my test accounts and it seemed to work fine.

Issue #5883, opened 2025-11-20
`SyncService` becomes offline if it fails processing the sync response


jmartinesp:
Reported on Element X Android: https://github.com/element-hq/element-x-android/issues/5567#issuecomment-3559791646

```
2025-11-20T19:16:04.014485Z DEBUG matrix_sdk::http_client: Got response | crates/matrix-sdk/src/http_client/mod.rs:210 | spans: next_sync_with_lock > sync_once{conn_id="encryption"} > send{request_id="REQ-72" method=POST uri="https://matrix.cervoi.se/_matrix/client/unstable/org.matrix.simplified_msc3575/sync" request_size="153B" status=200 response_size="311B" request_duration=682.703125ms}
2025-11-20T19:16:04.023573Z DEBUG matrix_sdk_base::response_processors::account_data::global: Found direct room data in the Store, applying it | crates/matrix-sdk-base/src/response_processors/account_data/global.rs:147 | spans: sync_once{conn_id="room-list"} > handle_response
2025-11-20T19:16:04.035039Z ERROR matrix_sdk_ui::sync_service: Error while processing room list in sync service: Some(
    RoomList(
        SlidingSync(
            StateStore(
                Json(
                    Error("the enum variant JoinRule::_Custom cannot be serialized", line: 0, column: 0),
                ),
            ),
        ),
    ),
) | crates/matrix-sdk-ui/src/sync_service.rs:426
2025-11-20T19:16:04.035238Z  INFO matrix_sdk_ui::sync_service: Entering the offline mode | crates/matrix-sdk-ui/src/sync_service.rs:160 | spans: supervisor task
```

We should probably just set if as offline for errors related to not reaching the server.

Hywan:
Possibly yes. What do you think @poljar?

poljar:
I mean, in this case we'll perpetually fail to sync. So I don't think it matters much that we're "online" if we never manage to process this sync response.

jmartinesp:
So any syncs including rooms with custom join rules will just fail indefinitely 🫤 . I guess it could make sense since we wouldn't know how to treat them, and probably just ignoring the room/marking it as invalid somehow would be difficult, but I feel like we should at least have some specific error so we can tell the user something like 'the app won't sync because one of your rooms ({room_id/room_name}) has incompatible join rules" instead of just being 'offline' without further feedback.

It seems like this issue comes from this decision: https://github.com/ruma/ruma/pull/719#discussion_r717480552

[...]

Issue #2949, opened 2023-12-14
Not robust against invalid state event


bmarty:
I have a room with this `m.room.join_rules` state event:
```
{
  "content": {
    "join_rule": "publictest"
  },
  "origin_server_ts": 1613508609851,
  "sender": "@benoitx:matrix.org",
  "state_key": "",
  "type": "m.room.join_rules",
  "unsigned": {
    "replaces_state": "$FZOkUjJvHm8mTuUkJBDhTOS86xfq17rZuxo0MRKBwH8",
    "age": 83365036681
  }
```
which is not valid regarding the [Matrix specs](https://spec.matrix.org/v1.9/client-server-api/#mroomjoin_rules), but the server did not perform any validation when I sent it, so it's there and the SDK must accept it and ignore any invalid value.

When I send a message to this room, I have got this error in loop:
> matrix_sdk_ui::sync_service: Error while processing room list in sync service: SlidingSync failed: the enum variant JoinRule::_Custom cannot be serialized | crates/matrix-sdk-ui/src/sync_service.rs:283.

Then the app becomes unusable, because the sync is restarted automatically in case of error. This is another issue that we will have to handle on the app side.

Rageshake: https://github.com/element-hq/element-x-android-rageshakes/issues/1210

This is how we mitigated this on the Kotlin SDK: https://github.com/element-hq/element-android/pull/3062

bnjbvr:
My wild guess would be the "hacky/temporary" caching of the events that's happening in sliding sync, failing because there's an invalid event. We could try to filter out custom events before serializing them, would be fun to see if Ruma allows to do this without going too further into details of the event types.

poljar:
Closed by #5924.

PR #5924, opened 2025-12-03
fix: Don't attempt to serialize custom join rules

poljar:
This is not supported by Ruma. The join_rule field, despite being defined as a pure string, can have associated data to it based on the join rule variant.

This means that custom and unknown enum variants might lose data when reserializing.

Let's just skip the serialization of custom join rules in the RoomInfo, the concrete value is still available in the state store, it's just not kept at hand in the RoomInfo.

This closes https://github.com/matrix-org/matrix-rust-sdk/issues/5883 and https://github.com/matrix-org/matrix-rust-sdk/issues/2949.

I have checked, the other enums that can encounter custom variants, i.e. `HistoryVisibilty` are not problematic as they have a working `Serialize` implementation for all variants, even the custom one.

### DIFF
--- a/crates/matrix-sdk-base/src/room/room_info.rs
+++ b/crates/matrix-sdk-base/src/room/room_info.rs
@@ -224,7 +224,7 @@ impl BaseRoomInfo {
                 | JoinRule::Restricted(_)
                 | JoinRule::KnockRestricted(_)
                 | JoinRule::Public => self.join_rules = Some(c.into()),
-                r => warn!("Encountered a custom joine rule {}, skipping", r.as_str()),
+                r => warn!("Encountered a custom join rule {}, skipping", r.as_str()),
             },
             AnySyncStateEvent::RoomCanonicalAlias(a) => {
                 self.canonical_alias = Some(a.into());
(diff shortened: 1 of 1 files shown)


crates/matrix-sdk-base/src/room/room_info.rs
```diff
@@ -217,9 +217,15 @@ impl BaseRoomInfo {
...
-                self.join_rules = Some(c.into());
-            }
+            AnySyncStateEvent::RoomJoinRules(c) => match c.join_rule() {
+                JoinRule::Invite
+                | JoinRule::Knock
+                | JoinRule::Private
+                | JoinRule::Restricted(_)
+                | JoinRule::KnockRestricted(_)
+                | JoinRule::Public => self.join_rules = Some(c.into()),
+                r => warn!("Encountered a custom joine rule {}, skipping", r.as_str()),
```
Hywan:
```suggestion
                r => warn!("Encountered a custom join rule {}, skipping", r.as_str()),
```

crates/matrix-sdk-base/src/room/room_info.rs
```diff
@@ -294,9 +300,15 @@ impl BaseRoomInfo {
...
-                self.join_rules = Some(c.into());
-            }
+            AnyStrippedStateEvent::RoomJoinRules(c) => match &c.content.join_rule {
+                JoinRule::Invite
+                | JoinRule::Knock
+                | JoinRule::Private
+                | JoinRule::Restricted(_)
+                | JoinRule::KnockRestricted(_)
+                | JoinRule::Public => self.join_rules = Some(c.into()),
+                r => warn!("Encountered a custom joine rule {}, skipping", r.as_str()),
```
Hywan:
```suggestion
                r => warn!("Encountered a custom join rule {}, skipping", r.as_str()),
```

PR #5929, opened 2025-12-04
refactor: Upgrade uniffi to the latest upstream

jmartinesp:
This is needed to fix an issue with the Kotlin bindings crashing the unit tests that reference any classes from the FFI bindings, even if they're instantiated with the `NoHandle` value to ensure it doesn't use the actual native code.

Signed-off-by:

### DIFF
--- a/.deny.toml
+++ b/.deny.toml
@@ -53,4 +53,6 @@ allow-git = [
     "https://github.com/matrix-org/vodozemac",
     # A patch override for the bindings: https://github.com/Alorel/rust-indexed-db/pull/72
     "https://github.com/matrix-org/rust-indexed-db",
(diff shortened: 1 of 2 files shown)


jmartinesp:
> Merging https://github.com/matrix-org/matrix-rust-sdk/pull/5929 will improve performances by 46.62%

image

poljar:
> > Merging #5929 will improve performances by 46.62%
> 
> image

Sometimes it behaves like this, on one of the simplests benchmarks we have. One of these days we need to investigate.

jmartinesp:
I knew it was too good to be true! (also, there are no bindings involved in the benchmark so... 😅 )

PR #5930, opened 2025-12-04
feat(room): make load_event_with_relations also load relations when falling back to the network

Johennes:
`Room::load_event_with_relations` currently tries to load the event and any known relations from the cache first. If that fails, `/event` is called to retrieve the event itself. Nothing is done to also load the relations, however.

This PR extends `load_event_with_relations` to also issue `/relations` requests to load any relations when the cache look-up failed.

### DIFF
--- a/crates/matrix-sdk-ui/src/timeline/pinned_events_loader.rs
+++ b/crates/matrix-sdk-ui/src/timeline/pinned_events_loader.rs
@@ -177,17 +177,30 @@ impl PinnedEventsRoom for Room {
         related_event_filters: Option<Vec<RelationType>>,
     ) -> BoxFuture<'a, Result<(TimelineEvent, Vec<TimelineEvent>), matrix_sdk::Error>> {
         Box::pin(async move {
+            // First try to load both the event and its relations from the cache.
             if let Ok((cache, _handles)) = self.event_cache().await
                 && let Some(ret) =
-                    cache.find_event_with_relations(event_id, related_event_filters).await?
+                    cache.find_event_with_relations(event_id, related_event_filters.clone()).await?
             {
                 debug!("Loaded pinned event {event_id} and related events from cache");
                 return Ok(ret);
             }
 
+            // The event was not in the cache. Try loading it from the homeserver.
             debug!("Loading pinned event {event_id} from HS");
             let event = self.event(event_id, request_config).await?;
 
+            // Maybe at least some relations are in the cache?
+            if let Ok((cache, _handles)) = self.event_cache().await {
+                let related_events =
+                    cache.find_event_relations(event_id, related_event_filters).await?;
+                if !related_events.is_empty() {
+                    debug!("Loaded relations for pinned event {event_id} from cache");
+                    return Ok((event, related_events));
+                }
+            }
+
+            // No relations were found in the cache. Try loading them from the homeserver.
             debug!("Loading relations of pinned event {event_id} from HS");
             let mut related_events = Vec::new();
             let mut opts = RelationsOptions {
--- a/crates/matrix-sdk/CHANGELOG.md
+++ b/crates/matrix-sdk/CHANGELOG.md
@@ -17,6 +17,9 @@ All notable changes to this project will be documented in this file.
   `LatestEvent` (replaced by `LatestEventValue`). See the documentation of
   `matrix_sdk::latest_event` to learn about the new API.
   [#5624](https://github.com/matrix-org/matrix-rust-sdk/pull/5624/)
+- Expose a new method `RoomEventCache::find_event_relations` for loading
+  events relating to a specific event ID from the cache.
+  [#5930](https://github.com/matrix-org/matrix-rust-sdk/pull/5930/)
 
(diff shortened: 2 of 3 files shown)


crates/matrix-sdk-ui/src/timeline/pinned_events_loader.rs
```diff
@@ -180,7 +184,27 @@ impl PinnedEventsRoom for Room {
...
-            self.event(event_id, request_config).await.map(|e| (e, Vec::new()))
+            let event = self.event(event_id, request_config).await?;
+
+            debug!("Loading relations of pinned event {event_id} from HS");
+            let mut related_events = Vec::new();
+            let mut opts = RelationsOptions {
+                include_relations: IncludeRelations::AllRelations,
+                recurse: true,
+                ..Default::default()
+            };
```
Johennes, line 211: I wasn't sure if maybe we should set a limit for the events to fetch?

andybalaam: Recommendation from the Rust team is to add a limit, just in case. I would expect ~1000 items seems about right?

Johennes: Sounds good, have made this change. I force pushed it because it was so small and I wanted to spare you another cycle. Hope that's ok.

crates/matrix-sdk-ui/src/timeline/pinned_events_loader.rs
```diff
@@ -180,7 +186,28 @@ impl PinnedEventsRoom for Room {
...
+            let mut related_events = Vec::new();
+            let mut opts = RelationsOptions {
+                include_relations: IncludeRelations::AllRelations,
+                recurse: true,
+                limit: Some(uint!(1000)),
+                ..Default::default()
+            };
+
+            loop {
+                let relations = self.relations(event_id.to_owned(), opts.clone()).await?;
```
andybalaam, line 214: Hmm, does this fetch all relations because the pinned event was not in the cache? Maybe the relations are already in the cache? Or is that handled somehow?

crates/matrix-sdk-ui/src/timeline/pinned_events_loader.rs
andybalaam: We are settling on about 256 items as a reasonable guess.

andybalaam: Oh sorry, we crossed over on this. Fine by me to force-push with such a small change.

crates/matrix-sdk-ui/src/timeline/pinned_events_loader.rs
```diff
@@ -180,7 +186,28 @@ impl PinnedEventsRoom for Room {
...
+                related_events.extend(relations.chunk);
+                if let Some(next_from) = relations.next_batch_token {
+                    opts.from = Some(next_from);
+                } else {
+                    break;
+                }
+            }
```
andybalaam, line 221: One more question. We think it would be good to store these events in the cache after downloading them. We think something like this might work:

```
if let Ok((cache, _handles)) = self.event_cache().await {
            cache.save_events(related_events).await;
        }
```

If that seems like too much of a rabbit-hole let me know and we can make it separate.

jmartinesp: If we wanted to make this take into account all cases, I think the flow should probably be something like:

1. Fetch everything from the cache if possible.
2. If not, load the event from the HS.
3. Try to fetch the relations from the cache using `EventCacheStore::find_event_relations`.
4. If they're not there, use the new loop.
5. If using this new loop, also use `cache.save_events(related_events)` for the events we just loaded.

`EventCacheStore::find_event_relations` is not public though, so maybe we'd have to create some wrapper function to expose it to the pinned events loader code. I'm not saying this has to be done now, just that I think it's the most correct flow.

crates/matrix-sdk-ui/src/timeline/pinned_events_loader.rs
Johennes: Sorry, I shot too quickly there. 🙈 

Have tuned the limit down to 256 in 265c9188661c5bffe3b09fa7da57532eb9d9be6f.

crates/matrix-sdk-ui/src/timeline/pinned_events_loader.rs
Johennes: Sounds good on both. I have added related events to the cache with f6dc2f1089fc32af16fed260526d4c8494839c99 and added a step to load the events from the cache in 2e6b214c633c574ee658e70eea373dfb702fc30d.

crates/matrix-sdk-ui/src/timeline/pinned_events_loader.rs
Johennes: I think we're covering this in the thread at https://github.com/matrix-org/matrix-rust-sdk/pull/5930#discussion_r2592234985 now.

PR #5934, opened 2025-12-08
change(notification_client): use cross process locking event and media stores instead of in-memory ones

stefanceriu:
This will allow the NSE process to populate the event cache store with even data after retrieving it from sync, as opposed to dropping after rendering the notification it as it did previously.


crates/matrix-sdk-base/src/client.rs
```diff
@@ -221,14 +221,16 @@ impl BaseClient {
...
+                .crypto_store(self.crypto_store.clone());
 
         let copy = Self {
-            state_store: BaseStateStore::new(config.state_store),
-            event_cache_store: config.event_cache_store,
-            media_store: config.media_store,
+            state_store: BaseStateStore::new(in_memory_store_config.state_store),
+            // The event and media stores both have cross process locking support.
+            event_cache_store: self.event_cache_store.clone(),
+            media_store: self.media_store.clone(),
```
Hywan: ⚠️  This is wrong!

You're cloning the `EventCacheStoreLock` and `MediaStoreLock`, with the main cross-process holder name. It doesn't use the new `cross_process_store_locks_holder_name` received by argument here.

Hywan (changes requested):

Hywan:
@stefanceriu if you want to give another try, have fun!

stefanceriu:
Turns out this doesn't actually work and we probably **do** need the state store cross process lock. Will raise again once we're ready to do this properly.

PR #5936, opened 2025-12-08
fix(event_cache): store correct encryption information for local echoes

stefanceriu:
This patch exposes one's underlying `EncryptionInfo` used when sending raw `Room` events, uses it to create a `DecryptedRoomEvent` and corresponding `TimelineEvent`, and correctly populates the EventCache with local echoes.

### DIFF
--- a/crates/matrix-sdk-crypto/CHANGELOG.md
+++ b/crates/matrix-sdk-crypto/CHANGELOG.md
@@ -6,6 +6,12 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased] - ReleaseDate
 
+### Features
+
+- The `OutboundGroupSession` and `OlmMachine` now return the `EncryptionInfo` 
+  used when encrypting raw events.
+  ([#5936](https://github.com/matrix-org/matrix-rust-sdk/pull/5936))
+
 ## [0.16.0] - 2025-12-04
 
 ### Features
--- a/crates/matrix-sdk/CHANGELOG.md
+++ b/crates/matrix-sdk/CHANGELOG.md
@@ -8,6 +8,9 @@ All notable changes to this project will be documented in this file.
 
 ### Features
 
+- Sending `MessageLike` and `RawMessageLike` events through a `Room` now returns
+  the used `EncryptionInfo`, if any.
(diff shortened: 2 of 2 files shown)


crates/matrix-sdk/src/room/futures.rs
```diff
@@ -95,7 +96,7 @@ impl<'a> SendMessageLikeEvent<'a> {
 }
 
 impl<'a> IntoFuture for SendMessageLikeEvent<'a> {
-    type Output = Result<send_message_event::v3::Response>;
+    type Output = Result<(send_message_event::v3::Response, Option<EncryptionInfo>)>;
```
poljar: A proper custom type for this instead of a tuple would be preferred.

If we ever have to modify this again, it would be far less painful that way.

crates/matrix-sdk-crypto/src/machine/mod.rs
```diff
@@ -1984,6 +1984,25 @@ impl OlmMachine {
...
     }
 
+    /// Retrieve information about our own encryption setup.
+    pub fn get_own_encryption_info(&self) -> EncryptionInfo {
+        let identity_keys = self.identity_keys();
+
+        EncryptionInfo {
+            sender: self.inner.user_id.clone(),
+            sender_device: Some(self.inner.device_id.clone()),
+            algorithm_info: AlgorithmInfo::MegolmV1AesSha2 {
```
poljar: Hmm, this isn't completely correct, we assume `MegolmV1AesSha2` but this might not be the algorithm that will be used in `room_send()`.

crates/matrix-sdk/src/room/futures.rs
poljar: If `encrypt_room_event_raw()` returns the `EncryptionInfo`, then we can populate it with the correct algorithm and the session ID as well.

Basically `encrypt_room_event_raw()` would call `olm.get_own_encryption_info()` for us.

Sorry, this is a bit different from my original suggestion but it's more correct™.

stefanceriu: No no, this makes way more sense 👍

stefanceriu:
Okay, I went ahead and rewrote most of this based on the last idea.

PR #5931, opened 2025-12-05
feat[bindings]: expose power level thresholds in corresponding timeline event

JoFrost:
Hello! This PR exposes power-level thresholds in OtherState::RoomPowerLevels for the FFI timeline.

It includes action requirements such as ban, kick, invite, redact, state_default, events_default, per-event overrides and notifications. This gives FFI clients direct access to the effective power levels of a room, which were previously unavailable.

- [x] Public API changes documented in changelogs (optional)

Signed-off-by:

PR #5937, opened 2025-12-09
feat(ffi): Add fn RoomPowerLevels::events

jmartinesp:
## Changes:

- Rename `ffi::TimelineEventType` to `ffi::TimelineEventContent` since it had the contents inside the enum, not just the type.
- Create a new `ffi::TimelineEventType` enum with just the type info. Add missing cases like the multimedia ones. For this I had to add new features to the Ruma dependency.
- Rename `ffi::TimelineEvent::event_type` to `ffi::TimelineEvent::content` to match the renaming done above.
- Add `ffi::RoomPowerLevels::events` function, which returns a `HashMap<TimelineEventType, i64>`. All the previous changes are so we can use something type-safe for the keys, and also to fix the event type not containing only the type issue.

Signed-off-by:

### DIFF
--- a/bindings/matrix-sdk-ffi/CHANGELOG.md
+++ b/bindings/matrix-sdk-ffi/CHANGELOG.md
@@ -23,9 +23,16 @@ All notable changes to this project will be documented in this file.
   `Room::new_latest_event` overwrites the `Room::latest_event` method. See the
   documentation of `matrix_sdk::latest_event` to learn about the new API.
   [#5624](https://github.com/matrix-org/matrix-rust-sdk/pull/5624/)
+- Created `RoomPowerLevels::events` function which returns a `HashMap<TimelineEventType, i64>` with all the power 
+  levels per event type. ([#5937](https://github.com/matrix-org/matrix-rust-sdk/pull/5937))
   
 ### Refactor
 
+- [**breaking**] The existing `TimelineEventType` was renamed to `TimelineEventContent`, because it contained the 
+  actual contents of the event. Then, we created a new `TimelineEventType` enum that actually contains *just* the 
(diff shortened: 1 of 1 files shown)


jmartinesp:
FWIW all of the refactoring was done so we don't pass a raw string around for the event types in the FFI layer. If we think this is an overkill we can just keep the first commit and ignore the rest, although I think the new changes make more sense.

jmartinesp:
It seems like https://github.com/matrix-org/matrix-rust-sdk/pull/5931 it's doing something quite similar, even providing extra functionality on some parts 🤔 . However, there we now have `ffi::TimelineEventType` and `ffi::FFITimelineEventType`, which seems quite confusing. I'm wondering if we should try to unify both PRs.

JoFrost:
Hey, author of #5931 here.

Indeed, it looks like our PRs are complementary. I had planned to make a second PR to expose the power level for any given event, and you’ve essentially covered that here. Your approach is cleaner, and I can use your changes as the base for #5931.

bindings/matrix-sdk-ffi/src/room/power_levels.rs
```diff
@@ -29,6 +29,10 @@ impl RoomPowerLevels {
         self.inner.clone().into()
     }
 
+    fn events(&self) -> HashMap<crate::event::TimelineEventType, i64> {
+        self.inner.events.iter().map(|(key, value)| (key.clone().into(), (*value).into())).collect()
+    }
```
poljar, line 34: Docs please, also it's a bit weird that this isn't `pub` when all of our other exported methods are.

jmartinesp:
@poljar I finally got the CC tests to pass in https://github.com/matrix-org/complement-crypto/pull/219. Could you also approve those?

PR #5940, opened 2025-12-09
fix(sdk): support granting login with QR to client with device IDs that isn't a base64 encoded Curve25519 public key

hughns:
MSC4108 does not mandate that the device ID is a base64 encoded Curve25519 public key. Therefore we should support granting a login from a client that is using some other form of device ID.


hughns:
Unless I am misreading, the CI failure in the wasm tests appears to be unrelated to the changes in this PR.

poljar:
> Unless I am misreading, the CI failure in the wasm tests appears to be unrelated to the changes in this PR.

Yes sometimes this test doesn't want to behave. We don't know why as of yet. Kicking it gets it going...

PR #5941, opened 2025-12-10
Upgrade matrix-sdk-crypto to Rust edition 2024

jplatte:
I see some extra warnings locally that don't seem to be related to the upgrade, hope they don't block CI.

Signed-off-by: Jonas Platte

### DIFF
--- a/crates/matrix-sdk-crypto/src/backups/mod.rs
+++ b/crates/matrix-sdk-crypto/src/backups/mod.rs
@@ -357,14 +357,10 @@ impl BackupMachine {
             let private_identity = self.store.private_identity();
             let identity = private_identity.lock().await;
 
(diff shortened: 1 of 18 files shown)


jplatte:
Is this a known-flaky benchmark? If not, probably got faster from some sort of lock being dropped earlier (i.e. before an `else` branch) => lower contention. `cargo fix --edition` logged _many_ warnings about types dropping earlier (some locks, but many more types that just zeroize on drop).

poljar:
> Is this a known-flaky benchmark? If not, probably got faster from some sort of lock being dropped earlier (i.e. before an `else` branch) => lower contention. `cargo fix --edition` logged _many_ warnings about types dropping earlier (some locks, but many more types that just zeroize on drop).

Sadly known-flaky.

PR #5943, opened 2025-12-10
feat(crypto): Add `forwarder_data` to `InboundGroupSession` and pickle.

kaylendog:
Issue: https://github.com/matrix-org/matrix-rust-sdk/issues/5109

- [ ] Public API changes documented in changelogs (optional)

PR #5945, opened 2025-12-10
feat: Add `forwarder: ForwarderInfo` to `EncryptionInfo`.

kaylendog:
Depends on #5943, #5980

### DIFF
--- a/crates/matrix-sdk-common/CHANGELOG.md
+++ b/crates/matrix-sdk-common/CHANGELOG.md
@@ -14,6 +14,8 @@ All notable changes to this project will be documented in this file.
 
 ### Features
 
+- Add field `forwarder` of type `ForwarderInfo` to `EncryptionInfo`, which which exposes information about the forwarder of the  keys with which an event was encrypted if they were shared as part of an [MSC4268](https://github.com/matrix-org/matrix-spec-proposals/pull/4268) room key bundle.
+  ([#5945](https://github.com/matrix-org/matrix-rust-sdk/pull/5945)).
 - [**breaking**] Cross-process lock can be dirty. The `CrossProcess::try_lock_once` now returns a new type `CrossProcessResult`, which is an enum with `Clean`, `Dirty` or `Unobtained` variants. When the lock is dirty it means it's been acquired once, then acquired another time from another holder, so the current holder may want to refresh its internal state.
   ([#5672](https://github.com/matrix-org/matrix-rust-sdk/pull/5672)).
 
--- a/crates/matrix-sdk-crypto/CHANGELOG.md
+++ b/crates/matrix-sdk-crypto/CHANGELOG.md
@@ -8,9 +8,7 @@ All notable changes to this project will be documented in this file.
 
 ### Features
 
-- Added a new field `forwarder` to `InboundGroupSession` of type `ForwarderData`, which stores 
-  information about the forwarder of a session shared in a room key bundle under 
-  [MSC4268](https://github.com/matrix-org/matrix-spec-proposals/pull/4268).
+- Added a new field `forwarder` to `InboundGroupSession` of type `ForwarderData`, which stores information about the forwarder of a session shared in a room key bundle under [MSC4268](https://github.com/matrix-org/matrix-spec-proposals/pull/4268).
   ([#5980])(https://github.com/matrix-org/matrix-rust-sdk/pull/5980)
 - The `OutboundGroupSession` and `OlmMachine` now return the `EncryptionInfo` 
   used when encrypting raw events.


crates/matrix-sdk-common/src/deserialized_responses.rs
```diff
@@ -328,6 +328,17 @@ pub struct EncryptionInfo {
     /// The device ID of the device that sent us the event, note this is
     /// untrusted data unless `verification_state` is `Verified` as well.
     pub sender_device: Option<OwnedDeviceId>,
+    /// If the keys for this message shared-on-invite as part of an
```
andybalaam:
```suggestion
    /// If the keys for this message were shared-on-invite as part of an
```

crates/matrix-sdk-common/src/deserialized_responses.rs
```diff
@@ -328,6 +328,17 @@ pub struct EncryptionInfo {
...
+    /// [MSC4268] key bundle, the ID of the user who sent them to us.
+    ///
+    /// [MSC4268]: https://github.com/matrix-org/matrix-spec-proposals/pull/4268
+    pub forwarder: Option<OwnedUserId>,
+    /// If the keys for this message were shared-on-invite as part of an
+    /// [MSC4268] key bundle, the ID of the device from which they were
+    /// sent.
+    ///
+    /// [MSC4268]: https://github.com/matrix-org/matrix-spec-proposals/pull/4268
+    pub forwarder_device: Option<OwnedDeviceId>,
```
andybalaam: I think I would prefer a single Option that contains a struct containing these 2 values, since it sounds like they will both be either None or Some in lockstep. I am persuadable though.

crates/matrix-sdk-crypto/src/machine/mod.rs
```diff
@@ -2016,11 +2018,26 @@ impl OlmMachine {
...
+                }
+                _ => {
+                    // TODO: Should this return an error?
+                    warn!(
+                        sender_id = ?sender,
+                        device_id = ?device_id,
+                        session_id = session.session_id(),
+                        "Invalid sender verification state for forwarded session"
+                    );
+                    None
```
kaylendog: This is an interesting case since it *should* never happen, since we only ever import sessions with sender data matching the above variants.

crates/matrix-sdk-crypto/src/machine/mod.rs
```diff
@@ -2016,11 +2018,26 @@ impl OlmMachine {
...
+            sender: sender.to_owned(),
+            sender_device: device_id.clone(),
+            forwarder: session.forwarder_data.as_ref().and_then(|data| data.user_id()),
+            forwarder_device: session.forwarder_data.as_ref().and_then(|data| match data {
+                SenderData::SenderUnverified(known_sender_data)
+                | SenderData::SenderVerified(known_sender_data) => {
+                    known_sender_data.device_id.clone()
+                }
+                _ => {
+                    // TODO: Should this return an error?
```
andybalaam: I don't know. Should this never happen? If so, printing and continuing is probably OK. I think it should be an error rather than a warning though, if we're confident it shouldn't happen.

crates/matrix-sdk-common/src/deserialized_responses.rs
kaylendog: Changed in https://github.com/matrix-org/matrix-rust-sdk/pull/5980

crates/matrix-sdk-crypto/src/machine/mod.rs
kaylendog: The new changes in https://github.com/matrix-org/matrix-rust-sdk/pull/5980 let us avoid this, but we still have some optional chaining shenanigans to get around `KnownSenderData::device_id` being `Optional` for backwards compatibility.

crates/matrix-sdk-common/src/deserialized_responses.rs
kaylendog: Addressed in https://github.com/matrix-org/matrix-rust-sdk/pull/5945/commits/0ee914e9a366f5db51adf73f0f2d579a7c6a10c0

crates/matrix-sdk-common/CHANGELOG.md
```diff
@@ -14,6 +14,8 @@ All notable changes to this project will be documented in this file.
 
 ### Features
 
+- Add field `forwarder` of type `ForwarderInfo` to `EncryptionInfo`, which which exposes information about the forwarder of the  keys with which an event was encrypted if they were shared as part of an [MSC4268](https://github.com/matrix-org/matrix-spec-proposals/pull/4268) room key bundle.
+  ([#5945](https://github.com/matrix-org/matrix-rust-sdk/pull/5945)).
```
richvdh, line 18: This was not released as part of 0.16.0: it should be in the "unreleased" section.

richvdh: Fix in https://github.com/matrix-org/matrix-rust-sdk/pull/6007

PR #5933, opened 2025-12-05
IndexedDB:  initialize `StoreCipher` in encrypted store tests

mgoldenberg:
## Background

This pull request is part of a series of pull requests to add a full IndexedDB implementation of the `EventCacheStore` and  `MediaStore` (see #4617, #4996, #5090, #5138, #5226, #5274, #5343, #5384, #5406, #5414, #5497, #5506, #5540, #5574, #5603, #5676, #5682, #5749, #5795). This particular pull request fixes a number of bugs related to encrypted stores and their tests.

The primary issue is that the encrypted tests were being run without a `StoreCipher` for both the `EventCacheStore` and the `MediaStore`. Furthermore, once the `StoreCipher` is initialized for these tests, some of the `EventCacheStore` tests were failing.

### Initialize `StoreCipher` in encrypted tests

This is pretty simple, just add a `StoreCipher` to `event_cache_store::tests::encrypted::get_event_cache_store` and `media_store::tests::encrypted::get_media_store`.

### Fix failing tests in `EventCacheStore`

A number of tests began to fail when introducing encryption to the `EventCacheStore`.

- `event_cache_store_integration_tests::test_get_room_events`
- `event_cache_store_integration_tests::test_get_room_events_filtered`
- `event_cache_store_integration_tests::test_remove_room`

These tests failed because the constants representing the uppermost and lowermost possible bounds of an `EventId` were being encrypted when constructing a query for all events in a room. Once encrypted, these constants no longer represented the uppermost and lowermost possible bounds of an `EventId`. 

The solution is to not encrypt these constant values, as they lose their purpose once encrypted.

## Future Work

- Expose `EventCacheStore` and `MediaStore` outside of the `matrix-sdk-indexeddb` 

---

- [x] Public API changes documented in changelogs (optional)

Signed-off-by: Michael Goldenberg <m@mgoldenberg.net>

PR #5946, opened 2025-12-11
IndexedDB: use implementations of `EventCacheStore` and `MediaStore` in client

mgoldenberg:
## Background

This is the final pull request in the series of pull requests to add a full IndexedDB implementation of the `EventCacheStore` and  `MediaStore` (see #4617, #4996, #5090, #5138, #5226, #5274, #5343, #5384, #5406, #5414, #5497, #5506, #5540, #5574, #5603, #5676, #5682, #5749, #5795, #5933). This particular pull request adds the IndexedDB implementations of `EventCacheStore` and `MediaStore` into the `matrix_sdk::Client` when it is configured to use IndexedDB as its data store.

### Expose `EventCacheStore` and `MediaStore` functions outside `matrix-sdk-indexeddb`

This involved re-exporting some of the types internal to the crate - e.g., `IndexeddbEventCacheStore` and `IndexeddbMediaStore`.

Additionally, a composite type was added to hold all stores in a single structure - i.e., `IndexeddbStores`. This structure is also equipped with a set of functions for opening all the stores with a single function call.

### Refactor feature flags

There were some issues with enabling individual feature flags - e.g., enabling `event-cache-store` or `media-store` without `e2e-encryption` failed to compile. These issues surfaced once trying to use them via `matrix-sdk`. 

The flags have been refactored so that each of the different stores may be used independent of and in combination with the others.

### Use implementations of `EventCacheStore` and `MediaStore` in `matrix-sdk::Client`

[`matrix_sdk::client::builder::build_indexeddb_store_config`](https://github.com/matrix-org/matrix-rust-sdk/blob/f2ba338e1202db55268d9f8e0b845f4c6036e55e/crates/matrix-sdk/src/client/builder/mod.rs#L702) has been updated so that it initializes the `StoreConfig` with IndexedDB implementations of all of the stores. And, of course, it includes the crypto store when `e2e-encryption` is enabled.

---

### DIFF
--- a/crates/matrix-sdk/CHANGELOG.md
+++ b/crates/matrix-sdk/CHANGELOG.md
@@ -23,6 +23,9 @@ All notable changes to this project will be documented in this file.
 - Expose a new method `RoomEventCache::find_event_relations` for loading
   events relating to a specific event ID from the cache.
   [#5930](https://github.com/matrix-org/matrix-rust-sdk/pull/5930/)
+- Replace in-memory stores with IndexedDB implementations when initializing
+  `Client` with `BuilderStoreConfig::IndexedDb`.
+  [#5946](https://github.com/matrix-org/matrix-rust-sdk/pull/5946)
 
 ### Bugfix
 


crates/matrix-sdk-indexeddb/Cargo.toml
```diff
@@ -14,11 +14,14 @@ default-target = "wasm32-unknown-unknown"
...
-media-store = ["dep:matrix-sdk-base"]
-state-store = ["dep:matrix-sdk-base", "growable-bloom-filter"]
-e2e-encryption = ["dep:matrix-sdk-base", "dep:matrix-sdk-crypto"]
+default = ["e2e-encryption", "state-store", "event-cache-store", "media-store"]
+event-cache-store = ["dep:matrix-sdk-base", "media-store", "indexed-type-serializer"]
+media-store = ["dep:matrix-sdk-base", "indexed-type-serializer"]
+state-store = ["dep:matrix-sdk-base", "growable-bloom-filter", "safe-encode-traits"]
+e2e-encryption = ["dep:matrix-sdk-base", "dep:matrix-sdk-crypto", "safe-encode-serializer"]
+safe-encode-traits = []
+safe-encode-serializer = ["dep:matrix-sdk-crypto", "safe-encode-traits"]
```
Hywan, line 23: What `safe` means here?

crates/matrix-sdk/CHANGELOG.md
```diff
@@ -797,24 +792,22 @@ simpler methods:
...
+  `cross_process_store_locks_holder_name` argument.
+- `StoreConfig` no longer implements `Default`.
+- `BaseClient::new()` has been removed.
+- `BaseClient::clone_with_in_memory_state_store()` now takes a
+  `cross_process_store_locks_holder_name` argument.
+- `BaseClient` no longer implements `Default`.
+- `EventCacheStoreLock::new()` no longer takes a `key` argument.
+- `BuilderStoreConfig` no longer has
+  `cross_process_store_locks_holder_name` field for `Sqlite` and
+  `IndexedDb`.
```
Hywan: Nop, those are a sublist of the previous item.

crates/matrix-sdk/CHANGELOG.md
```diff
@@ -828,20 +821,20 @@ simpler methods:
...
- let mut subscriber = observer.subscribe();
+// Subscribe to the observer.
+let mut subscriber = observer.subscribe();
 
- // Use the subscriber as a `Stream`.
- let (message_event, (room, push_actions)) = subscriber.next().await.unwrap();
- ```
+// Use the subscriber as a `Stream`.
+let (message_event, (room, push_actions)) = subscriber.next().await.unwrap();
+```
```
Hywan: Nop.

mgoldenberg:
> I wonder why we would need a struct holding all the stores. We don't have that for SQLite nor in-memory stores. Why is it important here?

I was mostly just trying to stick to the pattern that was already established in [`matrix-sdk-indexeddb::open_stores_with_name`](https://github.com/matrix-org/matrix-rust-sdk/blob/4d3125e58e6c285407f3b3920a6eadf2c910e420/crates/matrix-sdk-indexeddb/src/lib.rs#L31). I don't know that it is super important, but my guess is that the original author was trying to help ensure that all the names of all the databases would have the same prefix.

I am happy to re-think this and, instead, expose functions that enable users of the crate to construct each store individually. However, there's one consideration to note about going that route.

The original author initialized a single `StoreCipher` through the `IndexeddbStateStore` and passed this around to subsequent stores.

```rust
pub async fn open_stores_with_name(
    name: &str,
    passphrase: Option<&str>,
) -> Result<(IndexeddbStateStore, IndexeddbCryptoStore), OpenStoreError> {
    // -- snip --

    let crypto_store =
        IndexeddbCryptoStore::open_with_store_cipher(name, state_store.store_cipher.clone())
            .await?;

    // -- snip --
}
```

I'm not sure why this was done, but it requires access to `IndexeddbStateStore::store_cipher`, which is currently `pub(crate)`. 

In order to allow users of the crate to construct each store individually, we would have to do one of two things.

1. Make `IndexeddbStateStore::store_cipher` accessible in some way to users of the crate 
2. Add functionality to each store so that it can initialize its own `StoreCipher` from a passphrase.
    - I believe `matrix-sdk-sqlite` does something like this 

Let me know what you think, happy to go in any direction! And also happy to explain more if this isn't making sense.

crates/matrix-sdk-indexeddb/Cargo.toml
mgoldenberg: `safe` is part of the name of an encoder and serializer - i.e., [`trait SafeEncode`](https://github.com/matrix-org/matrix-rust-sdk/blob/ad58607013ebc4064e39d46777155c9859da398a/crates/matrix-sdk-indexeddb/src/serializer/safe_encode/traits.rs#L37).

I agree that this isn't a good name for a feature, but it helps a lot to have more granular feature flags when adding `[#cfg(feature = ...)]` statements in the code. I thought about prefixing it with `_` or `internal-` to show that it really isn't for an end user, but for internal organization.

mgoldenberg: Open to other directions with this!

crates/matrix-sdk/CHANGELOG.md
mgoldenberg: Oh... these must be formatting changes... would you like me to try to revert them? I didn't realize those were in there!

crates/matrix-sdk/CHANGELOG.md
mgoldenberg: Same as above!

crates/matrix-sdk/CHANGELOG.md
Hywan: Please, yes, revert these modifications :-).

crates/matrix-sdk-indexeddb/Cargo.toml
Hywan: Let's keep as they are, and see how it goes in the future.

Hywan:
I prefer to stick with the original vision for the moment. We can still change in the future! Thank you for the explanations.

crates/matrix-sdk/CHANGELOG.md
mgoldenberg: fixup: [`4d21c7f603898469af4c12be78f1d87ff9ea1d98`](https://github.com/matrix-org/matrix-rust-sdk/pull/5946/commits/4d21c7f603898469af4c12be78f1d87ff9ea1d98)

crates/matrix-sdk/CHANGELOG.md
mgoldenberg: fixup: [`4d21c7f603898469af4c12be78f1d87ff9ea1d98`](https://github.com/matrix-org/matrix-rust-sdk/pull/5946/commits/4d21c7f603898469af4c12be78f1d87ff9ea1d98)

PR #5947, opened 2025-12-11
fix(sdk): Latest Event lazy-loads the Room Event Cache

Hywan:
This patch improves Latest Events in multiple ways.

First off, the system is lazier. Before, a `LatestEventValue` was initialized by computing a `Remote(_)` with the help of the `RoomEventCache`. Now, it is initialized by copying the value in the `RoomInfo`. It solves a bug where a latest event was registered but its corresponding `RoomEventCache` was not created yet.

Second, in addition to that, if the restored `LatestEventValue` is `None`, a lazy computation is asked from the Event Cache. It fixes a bug where the `RoomInfo` has no `LatestEventValue` but it's possible to compute one from the Event Cache.

Finally, replacing a `LatestEventValue::None` by itself is ignored, thus reducing the number of updates in the system.

---

### DIFF
--- a/crates/matrix-sdk/src/latest_events/mod.rs
+++ b/crates/matrix-sdk/src/latest_events/mod.rs
@@ -624,7 +624,6 @@ mod tests {
     }
 
     #[async_test]
-
     async fn test_latest_events_are_lazy() {
         let room_id_0 = room_id!("!r0");
         let room_id_1 = room_id!("!r1");


jmartinesp:
I can approve this from the client side perspective, but someone else should probably take a deeper look at the code.

crates/matrix-sdk/src/latest_events/mod.rs
```diff
@@ -1017,9 +1017,9 @@ mod tests {
         // The event cache has received its update from the sync. It has emitted a
         // generic update, which has been received by `LatestEvents` tasks, up to the
         // `compute_latest_events` which has updated the latest event value.
-        assert_next_matches!(
-            latest_event_stream,
-            LatestEventValue::Remote(RemoteLatestEventValue { kind: TimelineEventKind::PlainText { event }, .. }) => {
+        assert_matches!(
+            latest_event_stream.next().await,
+            Some(LatestEventValue::Remote(RemoteLatestEventValue { kind: TimelineEventKind::PlainText { event }, .. })) => {
```
poljar, line 1021: Will this fix things on slow systems? Don't you need `assert_next_matches_with_timeout!()` instead?

crates/matrix-sdk/src/latest_events/latest_event.rs
```diff
@@ -180,8 +203,48 @@ impl LatestEvent {
     }
 }
 
+pub(super) struct With<T, W> {
+    result: T,
+    with: W,
+}
```
poljar, line 214: Err can we document this. I'm not sure what the purpose of this is.

It seems to me like it's replacing a tuple with a struct for a result. We have this in many places, but I don't think we ever made a generic thing like this.

crates/matrix-sdk/src/latest_events/mod.rs
```diff
@@ -627,30 +584,47 @@ async fn compute_latest_events(
...
+            timestamp: MilliSecondsSinceUnixEpoch::now(),
+            content: SerializableEventContent::new(&AnyMessageLikeEventContent::RoomMessage(
+                RoomMessageEventContent::text_plain(body),
+            ))
+            .unwrap(),
+        }
+    }
 
     #[async_test]
+
```
poljar: Useless whitespace.

poljar (approved):
Alright, there are couple of nits but this looks good otherwise.

Approving ahead of time.

crates/matrix-sdk/src/latest_events/mod.rs
Hywan: I can use a timeout but it's unlikely to be useful here I think, it's really a small unit test.

crates/matrix-sdk/src/latest_events/latest_event.rs
Hywan: Just had a bit of fun for the sake of semantics. I will add documentation :-).

PR #5948, opened 2025-12-11
test: Allow setting a HTTP proxy using an env variable for the integretaion tests

poljar:


zecakeh:
Don't the default environment variables supported by reqwest work out of the box: https://docs.rs/reqwest/latest/reqwest/#proxies?

poljar:
> Don't the default environment variables supported by reqwest work out of the box: https://docs.rs/reqwest/latest/reqwest/#proxies?

Oh, duh. Indeed they do.

PR #5951, opened 2025-12-11
refactor(r2d2): Create common report_lag method to reduce duplicated code

poljar:


crates/matrix-sdk/src/event_cache/redecryptor.rs
```diff
@@ -638,6 +640,22 @@ impl EventCache {
     }
 }
 
+#[inline(always)]
+fn upgrade_event_cache(cache: &Weak<EventCacheInner>) -> Option<EventCache> {
+    cache.upgrade().map(|inner| EventCache { inner })
```
stefanceriu, line 645: Heh, I  believe we (iOS land) call this to "strongify" something.

stefanceriu (approved):
lgtm 👍

crates/matrix-sdk/src/event_cache/redecryptor.rs
poljar: Well the stdlib method is called `upgrade()` so I reused that terminology.

stefanceriu: Yeah no, don't get me wrong, wasn't commenting on your choices here, just naming differences between platforms.

Issue #5798, opened 2025-10-23
UTDs on pinned messages are not retried after key is received


richvdh:
* Pinned message is recieved but cannot be decrypted
* key is downloaded from key backup
* no further attempt at redecrypting pinned message

Related to https://github.com/matrix-org/matrix-rust-sdk/issues/5474, and may well be fixed by https://github.com/matrix-org/matrix-rust-sdk/issues/3872

poljar:
This is due to https://github.com/matrix-org/matrix-rust-sdk/issues/5954, workaround is being implemented in https://github.com/matrix-org/matrix-rust-sdk/pull/5955.

richvdh:
UTD Pinned events will now be retried if they are in the event cache, due to #5955.

However, pinned events may not be in the event cache, due to #5954, and so still may not get retried.

richvdh:
I think we ought to reopen this, until #5954 is resolved. It's not clear from that issue that it can cause UTDs.

poljar:
Will be closed by https://github.com/matrix-org/matrix-rust-sdk/pull/6085.

Issue #5954, opened 2025-12-11
Focused timelines bypass the event cache: no reads, no writes, limited to some updates only


poljar:
The issue consists of multiple problems:

1. Focused timelines fetch from the homeserver instead of the event cache
2. Focused timelines never populate the event cache with events
3. The event cache sends out updates using indices

As of 2025.12.11. we have the following variants in the `TimelineFocus` enum:

* `TimelineFocus::Live` - The normal timeline, showing all events.
* `TimelineFocus::Event` - A timeline focusing on a specific event and a number of contextual events around it.
* `TimelineFocus::Thread` - A timeline focusing on a specific thread.
* `TimelineFocus::PinnedEvents` - A timeline showing only the pinned events of a room. 

The latter three are what are considered to be focused timelines and work differently from the normal `Live` timeline. The event cache works on all events in a specific order, while focused timelines work on a relatively small subset of events.

The event cache gets populated by the sync and by backpagination requests of the live timeline. This makes it possible to know where to insert a new chunk of events and each event gets assigned a certain order and index.

These indices are used by the event cache to notify listeners, i.e. the live timeline, to update its view. 

So if a new event comes around or an event replaces another in the event cache, you might get something like "Set this event at index 0" or "Replace event at index 7 with this event".

Now a focused timeline only shows a subset of all events and is initially populated by getting events from the homeserver. The events are never put into the event cache and thus the event cache never assigned an index to those events.

Meaning, the focused timelines can't really react to the event cache providing live updates.

This issue is briefly mentioned in https://github.com/matrix-org/matrix-rust-sdk/issues/5053#issuecomment-2891250594

I'll copy the relevant snippet over:

> For the other timelines, though, it's a lie: we're still accumulating metadata from the events loaded in the context of those timelines, and putting them into all_remote_events, but they're not modeled after the live updates from the event cache. As a result, we can't apply the live updates in real-time (because the indices from the event cache updates won't match any of the events backing the non-live timeline), and have to complicate the implementation of handling live events related to a non-live timeline (see last commit in https://github.com/matrix-org/matrix-rust-sdk/pull/5060).

This has been the reason of at least the following workarounds and issues:

* https://github.com/matrix-org/matrix-rust-sdk/pull/5060
* https://github.com/matrix-org/matrix-rust-sdk/pull/4645
* https://github.com/matrix-org/matrix-rust-sdk/issues/5703
[...]

Issue #5703, opened 2025-09-24
thread summary: UTD not updated after being correctly decrypted


bnjbvr:
Might need #3872 first, for sanity.

richvdh:
@poljar: Is this still a problem? #5955 claims it might have fixed this

(Though presumably it is subject to the same caveats as #5798, and dependent on #5954 ?)

poljar:
> [@poljar](https://github.com/poljar): Is this still a problem? [#5955](https://github.com/matrix-org/matrix-rust-sdk/pull/5955) claims it might have fixed this
> 
> (Though presumably it is subject to the same caveats as [#5798](https://github.com/matrix-org/matrix-rust-sdk/issues/5798), and dependent on [#5954](https://github.com/matrix-org/matrix-rust-sdk/issues/5954) ?)

Your presumption is correct, threads are one of the focused timeline types.

richvdh:
Right. So it sounds like this doesn't just affect the thread *summary* (as implied by the issue title), but in fact the whole thread?

poljar:
Yes I think that's correct. This issue might be about the fact that #5954 isn't a sufficient fix for thread summaries to be handled as well.

Though I would hope that it is.

@bnjbvr would thread summaries now be updated if they are decrypted later on under the assumption that #5954 was fixed or do they need further special handling?

bnjbvr:
I think I could see why this wouldn't be working, so I'll write a test to double-check for it here, but TLDR: the thread summary might not get updated because r2d2 post processing pretends it's a sync, which may have the bad side-effect that it's skipping thread processing mostly. If I'm correct, though, then the patch is easy and I'm happy to write it.

bnjbvr:
Kind of unrelated to #5954, so removing the blocking relationship here.

PR #5955, opened 2025-12-11
Allow focused timelines to replace UTDs after redecryption

poljar:
This partially and very hackishly fixes a problem where focused timelines don't replace UTDs after the event cache redecrypts the event.

The problem is more closely described in https://github.com/matrix-org/matrix-rust-sdk/issues/5954.

This closes: #5798. Probably https://github.com/matrix-org/matrix-rust-sdk/issues/5703 as well but I still need to write a test for that one.

### DIFF
--- a/testing/matrix-sdk-integration-testing/src/tests/timeline.rs
+++ b/testing/matrix-sdk-integration-testing/src/tests/timeline.rs
@@ -1213,7 +1213,7 @@ async fn test_pinned_events_are_decrypted_after_recovering_with_event_count(
     // Still a UTD.
     assert!(
         item.content().is_unable_to_decrypt(),
-        "The pinned event should be an UTD as we didn't recover yet"
+        "The pinned event should be a UTD as we didn't recover yet"
     );
     assert_eq!(pinned_timeline.items().await.len(), 2);
 
@@ -1228,7 +1228,7 @@ async fn test_pinned_events_are_decrypted_after_recovering_with_event_count(
     assert_let!(VectorDiff::Set { index: 0, value } = next_item);
     let content = value.content();
 
-    // And we're not an UTD anymore.
+    // And we're not a UTD anymore.
     assert!(!content.is_unable_to_decrypt());
     let message = content.as_message().expect("The pinned event should be a message");
     assert_eq!(message.body(), "It's a secret to everybody");


poljar:
@jmartinesp or @stefanceriu this would probably be a good candidate for live testing with EX.

I don't think I changed anything except the UTD -> decrypted codepath, and there are some tests confirming this, but it would be wise to double check that this didn't break something accidentally.

A good test would be to login and open a pinned timeline and see if UTDs get resolved correctly and if nothing else is broken with the pinned timeline.

stefanceriu:
I don't know what to say, it's not looking good


https://github.com/user-attachments/assets/8cd0345c-f0d5-4310-a4ba-c4b0eca72971


1) It's not auto decrypting, app's been running for 5 mins before I started recording
2) Reloading the list doesn't do it
3) Triggering a backpagination does although not all items are from the first backpaginated page. Some are much older.
4) Even when decrypted it's showing me a super old version of my pinned items
5) Leaving the room and coming back in makes it show me the right pinned items (which I just added this morning)

poljar:
Alright, I added another test which confirms what we suspected.

If the pinned events aren't part of the main timeline and thus never get put into the event cache, then R2D2 can't find them to decrypt them. This is why the events in the video aren't decrypted.

I also found another issue where we would replace events with the incorrect thing because the mapping from event cache index to timeline index is broken.

testing/matrix-sdk-integration-testing/src/tests/timeline.rs
```diff
@@ -1081,3 +1082,152 @@ async fn test_local_echo_to_send_event_has_encryption_info() -> TestResult {
...
+    assert_eq!(another_alice.encryption().recovery().state(), RecoveryState::Enabled);
+
+    // The next update for the timeline should replace the UTD item with a decrypted
+    // value.
+    let next_item = assert_next_with_timeout!(stream, 5000);
+
+    assert_let!(VectorDiff::Set { index: 0, value } = next_item);
+    let content = value.content();
+
+    // And we're not an UTD anymore.
```
Hywan: “a UTD”, but ultra nit.

poljar: Fixed in https://github.com/matrix-org/matrix-rust-sdk/pull/5955/commits/d1b095bcda114903d9dc06244dd3a1d7548545bd.

PR #5957, opened 2025-12-12
feat(sdk,ffi): add `CheckCodeSender::validate()`

hughns:
The purpose of this is to allow the UI to validate the entered check code and provide user with the ability to retry entering the code.

The existing `send()` function is one shot and does not allow for retries.

This is for use in a scenario where the user has mistyped the code rather than where a MITM attack has been detected.

I'm not wedded to the `validate()` name, so if something else is preferable I can change it.

### DIFF
--- a/crates/matrix-sdk/src/authentication/oauth/qrcode/grant.rs
+++ b/crates/matrix-sdk/src/authentication/oauth/qrcode/grant.rs
@@ -756,14 +756,8 @@ mod test {
                             .expect("The checkcode should only be forwarded once")
                             .await
                             .expect("Alice should receive the checkcode");
-                        assert!(
-                            checkcode_sender.validate(checkcode),
-                            "The checkcode should validate"
-                        );
-                        assert!(
-                            !checkcode_sender.validate((checkcode + 1) % 100),
-                            "An incorrect checkcode should not validate"
-                        );
+                        assert!(checkcode_sender.validate(checkcode));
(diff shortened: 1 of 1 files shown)


hughns:
I will add some test coverage to this.

bindings/matrix-sdk-ffi/src/qr_code.rs
```diff
@@ -633,4 +633,16 @@ impl CheckCodeSender {
...
+    ///
+    /// This can be used to provide immediate feedback to the user.
+    ///
+    /// # Arguments
+    /// * `check_code` - The check code in digits representation.
+    ///
+    /// Returns `true` if the code is valid, `false` otherwise.
+    pub fn validate(&self, check_code: u8) -> bool {
+        self.inner.validate(check_code)
+    }
```
bmarty, line 647: Why introducing a new method? The existing `send` could fail fast if the code is not correct after a local check.

bmarty: Actually it's probably not a good idea to never send the code. So maybe add a parameter to the `send` method like `validate: Boolean`?
This is quite equivalent, so feel free to ignore my comments ;)

hughns:
This PR originated out of working on https://github.com/element-hq/element-meta/issues/3048. The conclusion for that project is that we don't need a way for the user to attempt re-entering the code and instead the user is prompted to restart the whole process. As such I no longer need this PR, so closing.

In case some future application does need this then a new PR can be made.

PR #5959, opened 2025-12-12
Remove unused `ShieldStateCode::SentInClear`

richvdh:
`VerificationState::to_shield_state_{strict,lax}` return a type `ShieldState`, whose inner type `ShieldStateCode` currently includes a variant `SentInClear` for "unencrypted event". This is very misleading, because those functions never actually set that variant.

Rather, there is a separate method in `matrix-sdk-ui`, `EventTimelineItem::get_shield`, which uses the same type, but *does* set that variant where appropriate.

As a user of matrix-sdk-common, without the matrix-sdk-ui layer, this is dangerously misleading: it gives the impression that we are checking for unencrypted events, when in fact we are not.

The solution seems to be to use different types for the different levels of the stack.

While we're at it, we fix up some of the confusion of methods that return an `Option` of an enum type which itself has a `None` variant.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2906,6 +2906,9 @@ name = "language-tags"
 version = "0.3.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "d4345964bb142484797b161f473a503a434de77149dd8c7427788c6e13379388"
+dependencies = [
+ "serde",
+]
 
 [[package]]
 name = "lazy_static"
@@ -4812,6 +4815,7 @@ dependencies = [
  "assign",
  "js_int",
  "js_option",
+ "language-tags",
  "ruma-client-api",
  "ruma-common",
  "ruma-events",
@@ -4886,6 +4890,7 @@ dependencies = [
  "indexmap",
  "js_int",
  "js_option",
+ "language-tags",
  "percent-encoding",
  "pulldown-cmark",
  "regex",
--- a/bindings/matrix-sdk-ffi/CHANGELOG.md
+++ b/bindings/matrix-sdk-ffi/CHANGELOG.md
@@ -28,9 +28,16 @@ All notable changes to this project will be documented in this file.
   `Room::new_latest_event` overwrites the `Room::latest_event` method. See the
   documentation of `matrix_sdk::latest_event` to learn about the new API.
   [#5624](https://github.com/matrix-org/matrix-rust-sdk/pull/5624/)
+- Created `RoomPowerLevels::events` function which returns a `HashMap<TimelineEventType, i64>` with all the power 
+  levels per event type. ([#5937](https://github.com/matrix-org/matrix-rust-sdk/pull/5937))
   
 ### Refactor
 
+- [**breaking**] The existing `TimelineEventType` was renamed to `TimelineEventContent`, because it contained the 
+  actual contents of the event. Then, we created a new `TimelineEventType` enum that actually contains *just* the 
+  event type. ([#5937](https://github.com/matrix-org/matrix-rust-sdk/pull/5937))
+- [**breaking**] The function `TimelineEvent::event_type` is now `TimelineEvent::content`. 
+  ([#5937](https://github.com/matrix-org/matrix-rust-sdk/pull/5937))
 - [**breaking**] The `SpaceService` will no longer auto-subscribe to required
   client events when invoking the `subscribe_to_joined_spaces` but instead do it
   through its, now async, constructor.
--- a/bindings/matrix-sdk-ffi/Cargo.toml
+++ b/bindings/matrix-sdk-ffi/Cargo.toml
@@ -65,7 +65,30 @@ matrix-sdk-ffi-macros.workspace = true
 matrix-sdk-ui = { workspace = true, features = ["uniffi"] }
 mime = "0.3.17"
 once_cell.workspace = true
-ruma = { workspace = true, features = ["html", "unstable-msc3488", "compat-unset-avatar", "unstable-msc3245-v1-compat", "unstable-msc4278"] }
+ruma = { workspace = true, features = [
+    "html",
+    "unstable-msc3488",
+    "compat-unset-avatar",
+    "unstable-msc3245-v1-compat",
+    "unstable-msc4278",
+    "unstable-msc3230",
+    # Audio event type
+    "unstable-msc3927",
+    # File event type
+    "unstable-msc3551",
+    # Image event type
+    "unstable-msc3552",
+    # Video event type
+    "unstable-msc3553",
+    # Voice event type
+    "unstable-msc3245",
+    # Emote event type
+    "unstable-msc3954",
+    # Image pack event type
+    "unstable-msc2545",
+    # Room language event type
+    "unstable-msc4334",
+] }
 serde.workspace = true
 serde_json.workspace = true
 sentry = { workspace = true, optional = true, default-features = false, features = [
--- a/bindings/matrix-sdk-ffi/src/event.rs
+++ b/bindings/matrix-sdk-ffi/src/event.rs
@@ -11,6 +11,7 @@ use ruma::{
         AnySyncMessageLikeEvent, AnySyncStateEvent, AnySyncTimelineEvent, AnyTimelineEvent,
         MessageLikeEventContent as RumaMessageLikeEventContent, RedactContent,
         RedactedStateEventContent, StaticStateEventContent, SyncMessageLikeEvent, SyncStateEvent,
+        TimelineEventType as RumaTimelineEventType,
     },
     EventId,
 };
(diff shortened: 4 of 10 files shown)


richvdh:
This will need complement-crypto updates, but maybe someone could give it a review for sanity before I invest time in that.

crates/matrix-sdk-common/CHANGELOG.md
```diff
@@ -6,6 +6,11 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased] - ReleaseDate
 
+- [**breaking**] `ShieldStateCode` no longer includes
+  `SentInClear`. `VeificationState::to_shield_state_{lax,strict}` never
+  returned that code, ans so having it in the enum was somewhat misleading.
+  ([#5959](https://github.com/matrix-org/matrix-rust-sdk/pull/5936))
```
richvdh:
```suggestion
  ([#5959](https://github.com/matrix-org/matrix-rust-sdk/pull/5959))
```

crates/matrix-sdk-ui/CHANGELOG.md
```diff
@@ -8,6 +8,10 @@ All notable changes to this project will be documented in this file.
 
 ### Features
 
+- [**breaking**] `EventTimelineItem::get_shield` now returns a new type,
+  `TimelineEventShieldState`, which extends the old `ShieldState` with a code
+  for `SentInCliear`, now that the latter has been removed from `ShieldState`.
+  ([#5959](https://github.com/matrix-org/matrix-rust-sdk/pull/5936))
```
richvdh:
```suggestion
  ([#5959](https://github.com/matrix-org/matrix-rust-sdk/pull/5959))
```

crates/matrix-sdk-ui/src/timeline/event_item/mod.rs
```diff
@@ -693,3 +693,76 @@ impl ReactionsByKeyBySender {
...
+        /// A human readable description.
+        message: &'static str,
+    },
+    /// A grey shield with a tooltip containing the associated message should be
+    /// presented.
+    Grey {
+        /// A machine-readable representation.
+        code: TimelineEventShieldStateCode,
+        /// A human readable description.
+        message: &'static str,
```
stefanceriu: Seeing as we're on the UI level I'm questioning how useful the message actually is without localization support. I'd say the code is good enough for both debugging and localization purposes and we can entirely drop the message.

crates/matrix-sdk-ui/CHANGELOG.md
```diff
@@ -8,6 +8,10 @@ All notable changes to this project will be documented in this file.
 
 ### Features
 
+- [**breaking**] `EventTimelineItem::get_shield` now returns a new type,
+  `TimelineEventShieldState`, which extends the old `ShieldState` with a code
+  for `SentInCliear`, now that the latter has been removed from `ShieldState`.
```
stefanceriu:
```suggestion
  for `SentInClear`, now that the latter has been removed from `ShieldState`.
```

bindings/matrix-sdk-ffi/src/timeline/mod.rs
```diff
@@ -1262,13 +1262,13 @@ impl LazyTimelineItemProvider {
     /// Returns the shields for this event timeline item.
     fn get_shields(&self, strict: bool) -> Option<ShieldState> {
```
stefanceriu: Why did you decide to keep the option on the FFI layer but the non-optional None variant on the UI level? 🤔

crates/matrix-sdk-ui/src/timeline/event_item/mod.rs
```diff
@@ -693,3 +693,75 @@ impl ReactionsByKeyBySender {
...
+    },
+    /// No shield should be presented.
+    None,
+}
+
+impl From<ShieldState> for TimelineEventShieldState {
+    fn from(value: ShieldState) -> Self {
+        match value {
+            ShieldState::Red { code, message } => {
```
stefanceriu: Random point: I've said this before and I'll keep saying it every time I see these things: Red and Gray have no intrinsic meaning and depending on the culture you come from might mean the exact opposite (e.g. China). Naming them after a color does by no means guarantee that they will actually be colored that way in the UI (being in the UI crate makes it worse). I would much rather see them call for what they are BigProblem, LegacyProblem, NoProblem or whatever. Thanks for coming to my TED Talk.

crates/matrix-sdk-ui/src/timeline/event_item/mod.rs
```diff
@@ -693,3 +693,75 @@ impl ReactionsByKeyBySender {
...
+            }
+            ShieldState::None => TimelineEventShieldState::None,
+        }
+    }
+}
+
+/// Extends [`ShieldStateCode`] to allow for a `SentInClear` code.
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+#[cfg_attr(feature = "uniffi", derive(uniffi::Enum))]
+pub enum TimelineEventShieldStateCode {
```
stefanceriu, line 732: I quite like how [errors derive their display implementation](https://github.com/dtolnay/thiserror), I wonder if there's anything like that for normal enums. Then again, any reason why `TimelineEventShieldStateCode` can't be an actual `thiserror::Error`?

stefanceriu (approved):
Makes sense to me 👍 I'll leave it to you to deal with my comments in whatever way you see fit.

crates/matrix-sdk-ui/src/timeline/event_item/mod.rs
richvdh: Yup. To be honest my goal here was to make the code slightly less confusing without having to spend too long digging into (or changing) the way that element-x-android or element-x-ios use the shield state. If neither app uses the message, then of course I'm happy to drop it, but I'm reluctant to start refactoring the higher-level code.

bindings/matrix-sdk-ffi/src/timeline/mod.rs
richvdh: yeah, good question. Again it comes down to not wanting to have to change the application layers too much. Removing the `Option` in favour or the `None` variant felt like a cleaner solution, so I went with that for matrix-sdk-ui, but it felt like a more invasive change than just removing the `None` variant.

crates/matrix-sdk-ui/src/timeline/event_item/mod.rs
richvdh: yup, absolutely. It's also worth noting that we don't actually use *shields* in the UI for this (rather, padlocks, /!\ symbols, anything other than a shield), so the name is crap twice over.

But again, I didn't really want to do a big refactor here.

crates/matrix-sdk-ui/src/timeline/event_item/mod.rs
richvdh: I'm not entirely following you. Do we need to implement display here? What don't you like about what's there today?

crates/matrix-sdk-ui/src/timeline/event_item/mod.rs
stefanceriu: omg, didn't even think of that 🤦

bindings/matrix-sdk-ffi/src/timeline/mod.rs
stefanceriu: As a general rule we tend to be very lax on accepting breaking ffi changes. It's the price to pay to move things along without tripping on ceremony and we're fine with it.

crates/matrix-sdk-ui/src/timeline/event_item/mod.rs
stefanceriu: Yup okay, makes sense 👍

crates/matrix-sdk-ui/src/timeline/event_item/mod.rs
stefanceriu: Ah sorry, this was coming as an addition to removing the msg. We have very well definied messages in the documentatioe here and also a good way of exposing them to the final user through that macro. I figured it fits better.

bindings/matrix-sdk-ffi/src/timeline/mod.rs
richvdh: We discussed this elsewhere, and decided it would make sense to do the same thing in the bindings as in matrix-sdk-ui.

crates/matrix-sdk-ui/src/timeline/event_item/mod.rs
richvdh: we discussed this elsewhere, and decided to remove `message`.

richvdh:
https://github.com/matrix-org/complement-crypto/pull/220 contains the fixes to complement-crypto

poljar:
This will need a matching complement-crypto PR.

richvdh:
> This will need a matching complement-crypto PR.

It has one, at https://github.com/matrix-org/complement-crypto/pull/220.

PR #5961, opened 2025-12-15
Let R2D2 listen to backup state changes and report when a backup becomes available

poljar:


poljar:
> Is this already tested? If not, some tests would be appreciated.

The test that got modified, `test_enabling_backups_retries_decryption`, does test this.

The test fails if commit no. 3 gets applied before commit no. 2.

PR #5962, opened 2025-12-15
test: Disable the lease lock tests for the memory store on wasm

poljar:
This test for one reason or the other sporadically panics with an:
    > RuntimeError: unreachable

Let's disable this test on Wasm for now since the memory store isn't that relevant anyways, especially not on Wasm.


poljar:
To clarify, this is because we're seeing these kinds of failures: https://github.com/matrix-org/matrix-rust-sdk/actions/runs/20232260036/job/58077951474?pr=5961.

PR #5968, opened 2025-12-15
fix(sdk): Fix sticky local latest event

Hywan:
Based on top of https://github.com/matrix-org/matrix-rust-sdk/pull/5947. Must be merged before this one.

The problem we are trying to solve is the following:

- a local event is being sent,
- the `LatestEventValue` is `LocalIsSending`,
- the local event is finally sent,
- the `LatestEventValue` is still `LocalIsSending` purposely, with the hope that an update from the Event Cache will replace it, thus becoming a `Remote`.

But sometimes, this update from the Event Cache comes **before** the update from the Send Queue because the whole mechanism is async with no ordering guarantee. Why is it problem? Because updates from the Event Cache are ignored until the buffer of local `LatestEventValue`s aren't empty (so that local latest event cannot be overwritten and has the priority). It means that if the update from the Event Cache is received before `RoomSendQueueUpdate::SentEvent`, it is ignored, and the `LatestEventValue` stays in the `LocalIsSending` state. That's annoying.

The idea is to introduce a new state: `LocalHasBeenSent` which mimics `Remote`, but for a local event. It clarifies the state of a sent event, **without relying on the Event Cache**.

The patch is twofold:

1. It introduces the new `LatestEventValue::LocalHasBeenSent` variant,
2. It updates the `latest_event` Room List sorter to handle this new case.

---

### DIFF
--- a/bindings/matrix-sdk-ffi/CHANGELOG.md
+++ b/bindings/matrix-sdk-ffi/CHANGELOG.md
@@ -6,6 +6,13 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased] - ReleaseDate
 
+### Bug Fixes
+
+- [**breaking**] `LatestEventValue::Local { is_sending: bool }` is replaced
+  by [`state: LatestEventValueLocalState`] to represent 3 states: `IsSending`,
+  `HasBeenSent` and `CannotBeSent`.
+  ([#5968](https://github.com/matrix-org/matrix-rust-sdk/pull/5968/))
+
 ### Features
 
 - Add `SpaceService::get_space_room` to get a space given its id from the space graph if available.
--- a/crates/matrix-sdk-base/CHANGELOG.md
+++ b/crates/matrix-sdk-base/CHANGELOG.md
@@ -6,6 +6,12 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased] - ReleaseDate
 
+### Bug Fixes
+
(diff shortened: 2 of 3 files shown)


crates/matrix-sdk-base/src/latest_event.rs
```diff
@@ -65,6 +65,19 @@ impl LatestEventValue {
...
+    /// [`LocalIsSending`] nor [`LocalCannotBeSent`].
+    ///
+    /// [`LocalIsSending`]: LatestEventValue::LocalIsSending
+    /// [`LocalCannotBeSent`]: LatestEventValue::LocalCannotBeSent
+    pub fn is_remote_like(&self) -> bool {
+        match self {
+            Self::LocalIsSending(_) | Self::LocalCannotBeSent(_) => false,
+            Self::LocalHasBeenSent(_) | Self::Remote(_) | Self::None => true,
+        }
+    }
```
poljar: Perhaps a better name would be `has_been_sent()` though not sure how `LatestEventValue::None` fits into this.

Hywan: That was my initial choice, but I thought it was weird with remote events. They have been sent, for sure, but not by the current client.

Hywan: Changing for `is_unsent` as agreed together in the Matrix room.

PR #5970, opened 2025-12-16
fix(sdk_common): `TimelineEvent::from_bundled_latest_event` can remove `session_id`

jmartinesp:
What's more, this is saved into the event cache and sometimes it overrides another instance of the same event that actually contains the right info. This results in unresolvables UTDs.

This change tries to fetch the session id from the existing event content. It fixed these kind of UTDs when tested in a real client.

Signed-off-by:


crates/matrix-sdk-common/src/deserialized_responses.rs
```diff
@@ -713,14 +713,22 @@ impl TimelineEvent {
...
-                // UTD info, since we can't really do much better.
+                // information about it in the unsigned map. Try to fetch the information from
+                // the content instead.
+                let session_id = if let Some(content) =
+                    latest_event.get_field::<EncryptedEventScheme>("content").ok().flatten()
+                {
+                    match content {
+                        EncryptedEventScheme::MegolmV1AesSha2(content) => Some(content.session_id),
+                        EncryptedEventScheme::OlmV1Curve25519AesSha2(_) => None,
+                        _ => None,
```
Hywan: You can use `_ => None` only.

PR #5971, opened 2025-12-16
feat(sdk): Re-decrypt Room `latest_event` values after enabling key backup or the re-decryptor lagging

jmartinesp:
What the title says.

Signed-off-by:

### DIFF
--- a/crates/matrix-sdk/src/latest_events/mod.rs
+++ b/crates/matrix-sdk/src/latest_events/mod.rs
@@ -51,28 +51,35 @@ mod latest_event;
 mod room_latest_events;
 
 use std::{
-    collections::HashMap,
+    collections::{BTreeSet, HashMap},
     ops::{ControlFlow, DerefMut, Not},
     sync::Arc,
 };
 
 pub use error::LatestEventsError;
 use eyeball::{AsyncLock, Subscriber};
+use futures_util::pin_mut;
 use latest_event::{LatestEvent, With};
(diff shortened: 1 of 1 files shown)


crates/matrix-sdk/src/latest_events/mod.rs
```diff
@@ -582,6 +595,100 @@ async fn compute_latest_events(
     }
 }
 
+async fn handle_utds_task(registered_rooms: Arc<RegisteredRooms>, event_cache: EventCache) {
+    let reports_stream = event_cache.subscribe_to_decryption_reports();
+    pin_mut!(reports_stream);
+
```
Hywan, line 601: I think we could remove `LatestEventValue::Utd` by doing so:

- this task handles 2 streams:
  1. `EventCache::subscribe_to_decryption_reports`
  2. Another internal channel/stream, let's call it: `latest_utd_events`
- when a `LatestEventValue` is computed and a UTD is found, it sends a message over `latest_utd_events`
- the task keeps doing what it does now, but instead of using `RoomLatestEvents` to find out the `LatestEvent`, it should receive them via `latest_utd_events` or something.

The idea is to avoid having `LatestEventValue::Utd`, and also to reduce the pressure on the locks inside `LatestEvents` and `RoomLatestEvents`, with less `await` point.

Do you see what I mean?

jmartinesp:
Closing this as we found a better alternative.

Hywan:
Replaced by https://github.com/matrix-org/matrix-rust-sdk/pull/5976.

PR #5972, opened 2025-12-16
chore(spaces): make subscriptions explicit and rename certain fields.

stefanceriu:
This series of patches extracts the client events subscription from the `subscribe_to_joined_spaces` and moves it to the (now async) constructor, as well as rename certain fields and methods to clarify what each one of them does e.g. `joined_spaces` is actually a list of `top_level_spaces`, not all the joined ones.

Best reviewed commit by commit.

### DIFF
--- a/bindings/matrix-sdk-ffi/src/spaces.rs
+++ b/bindings/matrix-sdk-ffi/src/spaces.rs
@@ -52,11 +52,6 @@ impl SpaceService {
 
 #[matrix_sdk_ffi_macros::export]
 impl SpaceService {
-    /// Sets up the `SpaceService` by subscribing to necessary client events.
-    pub async fn setup(&self) {
-        self.inner.setup().await;
-    }
-
     /// Returns a list of all the top-level joined spaces. It will eagerly
     /// compute the latest version and also notify subscribers if there were
     /// any changes.
--- a/crates/matrix-sdk-ui/src/spaces/mod.rs
+++ b/crates/matrix-sdk-ui/src/spaces/mod.rs
@@ -105,8 +105,7 @@ struct SpaceState {
 ///
 /// # async {
 /// # let client: Client = todo!();
-/// let space_service = SpaceService::new(client.clone());
-/// space_service.setup().await;
+/// let space_service = SpaceService::new(client.clone()).await;
 ///
 /// // Get a list of all the joined spaces
 /// let joined_spaces = space_service.top_level_joined_spaces().await;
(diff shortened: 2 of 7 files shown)


crates/matrix-sdk-ui/src/spaces/mod.rs
```diff
@@ -835,7 +845,7 @@ mod tests {
         let space_service = SpaceService::new(client.clone());
 
         // Wait for the space hierarchy to register.
-        _ = space_service.joined_spaces().await;
+        _ = space_service.top_level_joined_spaces().await;
```
pixlwave: This can also call `setup` now instead

bindings/matrix-sdk-ffi/src/spaces.rs
```diff
@@ -56,7 +56,7 @@ impl SpaceService {
     /// compute the latest version and also notify subscribers if there were
     /// any changes.
     pub async fn joined_spaces(&self) -> Vec<SpaceRoom> {
```
pixlwave: Presumably you're going to rename the FFI too, but commenting in case :)

crates/matrix-sdk-ui/src/spaces/mod.rs
stefanceriu: Nice, thanks 👍

crates/matrix-sdk-ui/src/spaces/mod.rs
```diff
@@ -149,11 +150,8 @@ impl SpaceService {
         }
     }
 
-    /// Subscribes to updates on the joined spaces list. If space rooms are
-    /// joined or left, the stream will yield diffs that reflect the changes.
-    pub async fn subscribe_to_joined_spaces(
-        &self,
-    ) -> (Vector<SpaceRoom>, VectorSubscriberBatchedStream<SpaceRoom>) {
+    /// Sets up the `SpaceService` by subscribing to necessary client events.
+    pub async fn setup(&self) {
```
Hywan: Why not moving the `setup` logic inside `new`?

crates/matrix-sdk-ui/src/spaces/mod.rs
```diff
@@ -389,7 +389,7 @@ impl SpaceService {
         space_state.graph = new_graph;
     }
 
-    async fn joined_spaces_for(client: &Client) -> (Vec<SpaceRoom>, SpaceGraph) {
+    async fn build_space_state(client: &Client) -> (Vec<SpaceRoom>, SpaceGraph) {
```
Hywan, line 392: Nitpicking: Since we are already on a type named `SpaceService`, I would name this method `build_state` directly.

crates/matrix-sdk-ui/src/spaces/mod.rs
```diff
@@ -374,7 +370,7 @@ impl SpaceService {
         Ok(handle)
     }
 
-    async fn update_joined_spaces_if_needed(
+    async fn update_space_state_if_needed(
```
Hywan, line 377: Nitpicking: Same logic would apply here, with `update_state_if_needed`. Just an idea, don't feel forced to adopt it!

crates/matrix-sdk-ui/src/spaces/mod.rs
stefanceriu: That's a good point actually. I initially didn't want to break client side setup but I think we must.

crates/matrix-sdk-ui/src/spaces/mod.rs
stefanceriu: I figured since we call the internal state struct a `SpaceState` it's best to just make it super explicit.

Hywan: Ok.

PR #5976, opened 2025-12-17
Let R2D2 redecrypt and update encryption info of in-memory events automatically

poljar:
This PR improves R2D2 a bit more. It will now automatically redecrypt events kept in memory in the event cache if it may have missed some room keys due to lag in the room key stream or if backups get enabled.

This removes the need for components which have a one-to-one mapping between the events the event cache keeps in memory and the component itself.

The timeline is such a case, thus the last patch of this PR removes the explicit redecryption requesting of the timeline.

There's a bit of refactoring to be done here in the first couple of commits, so a review commit by commit should be the easiest.

The behavior of the timeline with regards to backups being enabled should not change, as the `test_enabling_backups_retries_decryption` confirms.


Hywan (approved):
Looks great, thanks!

PR #5980, opened 2025-12-19
feat(crypto): Introduce `ForwarderData` for session forwader info.

kaylendog:
- Introduces a new enum `ForwarderData` as a wrapper for valid variants of `SenderData` for which we can accept key bundles under MSC4268.
- Converts `forwarder_data` in `InboundGroupSession` and `PickledInboundGroupSession` to use `Option<ForwarderData>` over `Option<SenderData>`.

### DIFF
--- a/crates/matrix-sdk-crypto/src/olm/group_sessions/forwarder_data.rs
+++ b/crates/matrix-sdk-crypto/src/olm/group_sessions/forwarder_data.rs
@@ -18,7 +18,7 @@ use serde::{Deserialize, Serialize};
 use crate::olm::{KnownSenderData, SenderData};
 
 /// Represents information about the user who forwarded a megolm session under
-/// [MSC4268]. This is similar to [SenderData], but it is limited to variants
+/// [MSC4268]. This is similar to [`SenderData`], but it is limited to variants
 /// where a cross-signing key has been received from the forwarder, as specified
 /// in the MSC.
 ///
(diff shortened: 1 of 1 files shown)


crates/matrix-sdk-crypto/src/olm/group_sessions/forwarder_data.rs
```diff
@@ -0,0 +1,159 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use ruma::{DeviceId, UserId};
+use serde::{Deserialize, Serialize};
+
+use crate::olm::{KnownSenderData, SenderData};
+
+/// Represents information about the user who forwarded a megolm session under
+/// [MSC4268]. This is similar to [SenderData], but it is limited to variants
```
richvdh: I think our normal style is to use a fixed-width font for identifier names:

```suggestion
/// [MSC4268]. This is similar to [`SenderData`], but it is limited to variants
```

(likewise below)

richvdh (approved):
LGTM otherwise

crates/matrix-sdk-crypto/src/olm/group_sessions/forwarder_data.rs
kaylendog: Addressed in https://github.com/matrix-org/matrix-rust-sdk/pull/5980/commits/cce65335f8b4bf17ddd6e41c5934116cf43171b7

PR #5981, opened 2025-12-19
Room: Add a method to set your own user's display name within a room

pixlwave:
I've wanted to add this for ages now… This PR lets clients implement e.g. the `/myroomnick` slash command.

🎄🎁

### DIFF
--- a/crates/matrix-sdk/src/room/mod.rs
+++ b/crates/matrix-sdk/src/room/mod.rs
@@ -114,7 +114,7 @@ use ruma::{
             avatar::{self, RoomAvatarEventContent},
             encryption::RoomEncryptionEventContent,
             history_visibility::HistoryVisibility,
-            member::{MembershipChange, SyncRoomMemberEvent},
+            member::{MembershipChange, RoomMemberEventContent, SyncRoomMemberEvent},
             message::{
                 AudioInfo, AudioMessageEventContent, FileInfo, FileMessageEventContent,
                 ImageMessageEventContent, MessageType, RoomMessageEventContent,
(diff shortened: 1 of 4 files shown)


crates/matrix-sdk/src/room/mod.rs
```diff
@@ -1109,6 +1109,37 @@ impl Room {
...
+        let member_event =
+            self.get_state_event_static_for_key::<RoomMemberEventContent, _>(user_id).await?;
+
+        let Some(member_event) = member_event else {
+            return Err(Error::InsufficientData);
+        };
+
+        let RawSyncOrStrippedState::Sync(raw_event) = member_event else {
+            return Err(Error::InsufficientData);
+        };
```
zecakeh: These two checks can be collapsed into a single one.

crates/matrix-sdk/src/room/mod.rs
```diff
@@ -1109,6 +1109,37 @@ impl Room {
...
+
+        let event = raw_event.deserialize()?;
+
+        let SyncStateEvent::Original(ev) = event else {
```
zecakeh: Having a redacted event shouldn't prevent from changing the display name. We can create a new `RoomMemberEventContent` with the same membership.

pixlwave: Ah nice, that makes sense. [65e19c9...5ca3fde](https://github.com/matrix-org/matrix-rust-sdk/compare/65e19c9caf95a245bd0115e6aa6c67ce0c6650ec..5ca3fde4c11421f358358f0a9de93fa16ba6ee0c)

crates/matrix-sdk/src/room/mod.rs
pixlwave: [65e19c9...5ca3fde](https://github.com/matrix-org/matrix-rust-sdk/compare/65e19c9caf95a245bd0115e6aa6c67ce0c6650ec..5ca3fde4c11421f358358f0a9de93fa16ba6ee0c)

PR #5988, opened 2025-12-26
feat(sdk): Add mTLS client certificate support

lazard36:
Add support for mutual TLS (mTLS) authentication by allowing clients to provide a client certificate (identity) that will be presented to the server during the TLS handshake.

Changes:
- Add `client_identity` field to `HttpSettings` in `native.rs`
- Add `client_certificate()` method to `ClientBuilder`
- Expose `client_certificate()` via FFI for Swift bindings

The feature accepts PKCS#12/PFX certificate data that can be used with reqwest::Identity for mutual TLS authentication.

Signed-off-by:

### DIFF
--- a/bindings/matrix-sdk-ffi/src/client_builder.rs
+++ b/bindings/matrix-sdk-ffi/src/client_builder.rs
@@ -4,7 +4,7 @@
 use std::{num::NonZeroUsize, sync::Arc, time::Duration};
 
 #[cfg(not(target_family = "wasm"))]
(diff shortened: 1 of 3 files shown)


poljar (changes requested):
This PR mostly makes sense but it seems that there are some compilation issues depending on the TLS version that is enabled.

Please fix those before we continue with the review.

poljar:
If you're going to open a new PR instead of updating this one please close the replaced PR in the future.

PR #5992, opened 2025-12-31
fix(search): Create key dirs if they don't exist.

mlm-games:
The issue is related to the encrypt_store_dir fn where, when creating the key file, it doesn't ensure that the parent directory exists first. It might be not optimal for the user of the crate to ensure in an non hacky manner, as the sdk iterates through most of its directories internally. Have a test to verify it, which can be removed later (if being merged) 

Needs a review, might not be the optimal solution as this is my first pr with the crate and am not that familiar with it (although do use it in one of my apps).

### DIFF
--- a/crates/matrix-sdk-search/src/encrypted/encrypted_dir.rs
+++ b/crates/matrix-sdk-search/src/encrypted/encrypted_dir.rs
@@ -415,7 +415,7 @@ impl EncryptedMmapDirectory {
     ) -> Result<KeyBuffer, OpenDirectoryError> {
         let dir_path = key_path.parent().unwrap_or(key_path);
 
-        let _ = create_dir_all(dir_path);
+        create_dir_all(dir_path).map_err(|err| err.into_tv_err(dir_path))?;
         // Derive a AES key from our passphrase using a randomly generated salt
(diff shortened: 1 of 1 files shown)


crates/matrix-sdk-search/src/encrypted/encrypted_dir.rs
```diff
@@ -413,10 +413,13 @@ impl EncryptedMmapDirectory {
         passphrase: &str,
         pbkdf_count: u32,
     ) -> Result<KeyBuffer, OpenDirectoryError> {
+        let dir_path = key_path.parent().unwrap_or(key_path);
+
+        let _ = create_dir_all(dir_path);
```
Hywan: I think we must **not** ignore the error coming from `create_dir_all`. It's important to forward this error to the caller.

mlm-games: True, used the same tv_forward_call to forward it (left it while testing, should have definitely done it before PRing)

Should it be force pushed, or is the second commit fine (would be squashed?)

Hywan: I can squash myself.

Issue #5877, opened 2025-11-18
Prevent `HomeserverConfig::discover` from performing the `discover` request if previous server info exists


jmartinesp:
In the clients we restore the session using the `fn ClientBuilder::homeserver_url` to restore the previous homeserver URL. However, this modifies the `HomeserverConfig` the client is built with, so if we had used `HomeserverConfig::ServerNameOrHomeserverUrl` or `HomeserverConfig::ServerName` when we logged in we'll now use `HomeserverConfig::HomeserverUrl` and that will override the existing homeserver config, so when we try to reset the server info we'll try to fetch the server name for that, which is no longer there.

We'd need to use `ClientBuilder::server_name_or_homeserver_url` instead, or a variant that allows passing both values. But then, this forces a server info fetch operation when we build the `Client` and if this fails, it won't be built, so no more offline mode in any implementing clients. 

So what we really need is some way to allow that request to fail when restoring an existing session, while also making sure we can retrieve the existing server info from the `StateStore`. Which isn't accessible before the `Client` is created.

zecakeh:
The problem is for the `.well-known` info right? Why not use the server name of the user ID as first fallback before attempting to use the homeserver URL? When restoring a session, the user ID should be restored too and in theory that should be the proper location for the `.well-known`.

Also the only data from `.well-known` that is used after logging in is the `rtc_foci` if I'm not mistaken, and MSC4143 changed the location where this info is found to [a new `GET /rtc/transports` endpoint](https://github.com/matrix-org/matrix-spec-proposals/blob/680ef7dd65bbfbc275cc3177f36286e94c021fa9/proposals/4143-matrix-rtc.md?plain=1#L671-L724) which is [implemented in Synapse](https://github.com/element-hq/synapse/pull/18967), so it seems like updating the MSC implementation would actually fix any issue with that?

PR #5996, opened 2026-01-05
fix(sdk): Use the server name from the user id as a fallback URL for fetching the well-known info

jmartinesp:
When using `Client::fetch_client_well_known`, use the server name in the user id as a fallback value if the server name is missing. If it's not there or this first attempt fails, use the homeserver url instead as we were doing so far.

Fixes https://github.com/matrix-org/matrix-rust-sdk/issues/5877 (or workarounds it).

Signed-off-by:

### DIFF
--- a/crates/matrix-sdk/CHANGELOG.md
+++ b/crates/matrix-sdk/CHANGELOG.md
@@ -29,6 +29,7 @@ All notable changes to this project will be documented in this file.
 
 ### Bugfix
 
+- Use the server name extracted from the user id in `Client::fetch_client_well_known` as a fallback value. Otherwise, sometimes the server name is not available and we can't reload the well-known contents. ([#5996](https://github.com/matrix-org/matrix-rust-sdk/pull/5996))
 - Latest Event is lazier: a `RoomLatestEvents` can be registered even if its
   associated `RoomEventCache` isn't created yet.
   ([#5947](https://github.com/matrix-org/matrix-rust-sdk/pull/5947))


crates/matrix-sdk/src/client/mod.rs
```diff
@@ -2046,22 +2046,43 @@ impl Client {
...
             .unwrap_or(
                 // Sometimes people configure their well-known directly on the homeserver so use
                 // this as a fallback when the server name is unknown.
                 &self.homeserver(),
             )
-            .to_string();
+            .to_owned();
+
+        // First try using the server name or homeserver url
+        match self.fetch_client_well_known_with_url(server_url.to_string()).await {
```
bmarty: Does it make sense to try with the server name first? We know that this can fail, and when this is failing, this is precisely why we need a .well-known file (to discover the homeserver API endpoint).

Maybe this is useful for the first attempt, when login in, and the `userId` is not known yet? In this case `server_url` is not resolved yet to the CS endpoint?

The .well-known URL should always be computed from the userId domain part, and this should never fail if the file exists.

jmartinesp: That's the alternative I mentioned in the PR description. I kept the previous structure adding this 2nd attempt, but maybe it's better to always try first with the server name, explicit or implicit, then the homeserver url as the fallback 🤔 .

jmartinesp: That said, if we extract the server name from the user id and we have no other url to check... which scheme should we add for the constructed URL? Should we enforce https, or just default to plain http, just in case the server does not use HTTPS?

bmarty: I'd go for https only

jmartinesp: I'd agree with that, but I think I'll try to reuse the scheme in the `homeserver` url, since otherwise we won't be able to test this behaviour (the network mocking layer in the tests doesn't work with HTTPS if I'm not mistaken).

crates/matrix-sdk/src/client/mod.rs
```diff
@@ -4524,4 +4552,72 @@ pub(crate) mod tests {
 
         assert_matches!(client.device_exists(owned_device_id!("ABCDEF")).await, Err(_));
     }
+
+    #[async_test]
+    async fn test_fetching_well_known_with_homeserver_url() {
+        let server = MatrixMockServer::new().await;
+        let client = server.client_builder().build().await;
+
+        Mock::given(method("GET"))
```
zecakeh: nit: you shouldn't need to use wiremock types, there is `server.mock_well_known().ok()` (or `ok_with_homeserver_url()` if you want to change the URL). Same for the other tests.

jmartinesp: Thanks, done in https://github.com/matrix-org/matrix-rust-sdk/pull/5996/commits/5535b41bf3fa1c791ec5b548125c90c484f6a6a9.

crates/matrix-sdk/src/client/mod.rs
```diff
@@ -2046,22 +2046,46 @@ impl Client {
...
+        let server_url = self
             .server()
-            .unwrap_or(
-                // Sometimes people configure their well-known directly on the homeserver so use
-                // this as a fallback when the server name is unknown.
-                &self.homeserver(),
-            )
-            .to_string();
+            .map(|server| server.to_string())
+            .or_else(|| self.user_id().map(|id| format!("{}://{}", scheme, id.server_name())));
```
Hywan, line 2069: I believe this part deserves a bit more explanations about why we are doing this, otherwise it looks good!

Hywan (changes requested):
Looks good to me. I thought we were already doing that… anyway. Thanks for the tests too!

Can you explain improve the comments to explain why we are doing this? I believe the explanations must live in the documentation of the method itself rather than an inline comment.

Finally, the `CHANGELOG.md` expects a new entry 😃.

jmartinesp:
Done, thanks: I added some more inline comments as well as doc comments. I also added the changelog entry, I usually do that once the approach used in the PR has been validated, so I don't have to also change it every time the logic in the PR does 😅 .

andybalaam:
For reference, the spec for this behaviour is at https://spec.matrix.org/v1.17/client-server-api/#well-known-uris I think. If you haven't, it would be worth reading through and checking we comply with it.

I looked it up because I thought we weren't allowed to guess hostname from Matrix ID, but it looks like in this case we are.

Hywan (approved):
Great!

PR #6004, opened 2026-01-07
fix(sqlite): Add WAL checkpoints when creating DB pools and vacuuming

jmartinesp:
We've noticed the WAL files growing larger and larger, not shrinking as expected. This may be due to connections being busy and not allowing the automatic WAL checkpoints to perform this shrinking operation or some other causes, we're still investigating the issue.

In the meantime we can manually add WAL checkpoints:

1. When creating the DB pools, before any read/write operations are done.
2. Before vacuuming, so the WAL has room to grow to match the DB.
3. After vacuuming, since the copied data in the WAL file has been stored in the DB.

Signed-off-by:

### DIFF
--- a/crates/matrix-sdk-sqlite/src/utils.rs
+++ b/crates/matrix-sdk-sqlite/src/utils.rs
@@ -187,6 +187,7 @@ pub(crate) trait SqliteAsyncConnExt {
     ///
     /// Only returns an error in tests, otherwise the error is only logged.
     async fn vacuum(&self) -> Result<()> {
+        // Truncate the WAL file before vacuuming so it has room to grow.
         self.wal_checkpoint().await;
         if let Err(error) = self.execute_batch("VACUUM").await {
             // Since this is an optimisation step, do not propagate the error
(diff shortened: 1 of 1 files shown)


crates/matrix-sdk-sqlite/src/utils.rs
```diff
@@ -187,6 +187,7 @@ pub(crate) trait SqliteAsyncConnExt {
     ///
     /// Only returns an error in tests, otherwise the error is only logged.
     async fn vacuum(&self) -> Result<()> {
+        self.wal_checkpoint().await;
```
Hywan, line 191: Please, add an inline comments to explain the WAL checkpoints here and below.

crates/matrix-sdk-sqlite/src/utils.rs
```diff
@@ -198,11 +199,19 @@ pub(crate) trait SqliteAsyncConnExt {
             return Err(error.into());
         } else {
             trace!("VACUUM complete");
+            self.wal_checkpoint().await;
         }
 
         Ok(())
     }
 
+    async fn wal_checkpoint(&self) {
```
Hywan, line 214: Can you add a documentation here explaining what it does, and why it might be useful please?

Hywan (approved):
Okay to try this :-). Thanks!

jmartinesp:
Added doc and inline comments in https://github.com/matrix-org/matrix-rust-sdk/pull/6004/commits/305d742fb5b334bc464f1090e27fb620a6527be8

PR #6005, opened 2026-01-07
feat(sdk): Introduce the `PollTimeout` type

Hywan:
This patch introduces the `PollTimeout` type to represent either no timeout with `PollTimeout::None`, some timeout with `PollTimeout::Some(_)`, or a default timeout with `PollTimeout::Default`. It's finer than the previous `bool` that was used, where `false` meant `PollTimeout::None`, and `true` meant `PollTimeout::Default`. It's now possible to pass a precise timeout value.

The underlying idea of this patch is to allow passing `timeout=0` instead of no `timeout` for initial sync requests in `RoomListService`.

Note: [The specs says](https://github.com/matrix-org/matrix-spec-proposals/blob/erikj/sss/proposals/4186-simplified-sliding-sync.md#top-level):

> How long to wait for new events in milliseconds. If omitted the response is always returned immediately, even if there are no changes. Ignored when no `pos` is set.

So no `pos` is equivalent to `pos=0`. The changes in `RoomListService` should be invisible, but at least it's explicit.

### DIFF
--- a/crates/matrix-sdk/src/sliding_sync/mod.rs
+++ b/crates/matrix-sdk/src/sliding_sync/mod.rs
@@ -891,12 +891,20 @@ pub struct UpdateSummary {
 }
 
 /// Define what kind of poll timeout [`SlidingSync`] must use.
+///
+/// [The spec says about `timeout`][spec]:
+///
+/// > How long to wait for new events […] If omitted the response is always
+/// > returned immediately, even if there are no changes.
+///
+/// [spec]: https://github.com/matrix-org/matrix-spec-proposals/blob/erikj/sss/proposals/4186-simplified-sliding-sync.md#top-level
 #[derive(Debug)]
 pub enum PollTimeout {
     /// No `timeout` must be present.
     None,
 
-    /// A `timeout=X` must be present, where `X` is the value inside `Some`.
+    /// A `timeout=X` must be present, where `X` is in seconds and
+    /// represents how long to wait for new events.
     Some(u32),
 
     /// A `timeout=X` must be present, where `X` is the default value passed to
@@ -905,6 +913,16 @@ pub enum PollTimeout {
 }
 
 impl PollTimeout {
+    /// Computes the smallest `PollTimeout` between two of them.
+    ///
+    /// The rules are the following:
+    ///
+    /// * `None` < `Some`,
+    /// * `Some(x) < Some(y)` if and only if `x < y`,
+    /// * `Some < Default`.
+    ///
+    /// The `Default` value is unknown at this step but is assumed to be the
+    /// largest.
     fn min(self, left: Self) -> Self {
         match (self, left) {
             (Self::None, _) => Self::None,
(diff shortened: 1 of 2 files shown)


crates/matrix-sdk/src/sliding_sync/list/builder.rs
```diff
@@ -71,7 +71,13 @@ impl SlidingSyncListBuilder {
     pub(super) fn new(name: impl Into<String>) -> Self {
         Self {
             sync_mode: SlidingSyncMode::default(),
-            requires_timeout: Arc::new(|request_generator| request_generator.is_fully_loaded()),
+            requires_timeout: Arc::new(|request_generator| {
+                if request_generator.is_fully_loaded() {
+                    PollTimeout::Default
+                } else {
+                    PollTimeout::None
```
andybalaam, line 78: What does `None` do? I.e. what is the server behaviour when we don't provide a timeout? Do we ever want this?

crates/matrix-sdk/src/sliding_sync/mod.rs
```diff
@@ -873,6 +877,36 @@ enum RoomSubscriptionState {
...
+    /// A `timeout=X` must be present, where `X` is the value inside `Some`.
+    Some(u32),
+
+    /// A `timeout=X` must be present, where `X` is the default value passed to
+    /// [`SlidingSyncBuilder::poll_timeout`].
+    Default,
+}
+
+impl PollTimeout {
+    fn min(self, left: Self) -> Self {
```
andybalaam, line 926: What is the meaning of this function? Please can we have a doc comment? Also, please can you explain why None is < Default (assuming I read this right?).

Is this function asymmetrical? (I.e. if I swap the inputs do I get different answers sometimes?) If so, please can you explain why?

crates/matrix-sdk/src/sliding_sync/list/builder.rs
Hywan: [The specs says](https://github.com/matrix-org/matrix-spec-proposals/blob/erikj/sss/proposals/4186-simplified-sliding-sync.md#top-level):

> How long to wait for new events in milliseconds. If omitted the response is always returned immediately, even if there are no changes. Ignored when no `pos` is set.

So no `pos` is equivalent to `pos=0`. The changes in `RoomListService` should be invisible, but at least it's explicit.

crates/matrix-sdk/src/sliding_sync/mod.rs
Hywan: I've added the following comments:

```rust
/// Computes the smallest `PollTimeout` between two of them.
///
/// The rules are the following:
///
/// * `None` < `Some`,
/// * `Some(x) < Some(y)` if and only if `x < y`,
/// * `Some < Default`.
///
/// The `Default` value is unknown at this step but is assumed to be the
/// largest.
```

This function is symmetrical.

andybalaam: Thanks!

PR #6009, opened 2026-01-08
chore(sdk): Sliding Sync `room_subscriptions` is definitely not sticky

Hywan:
Follow up of https://github.com/matrix-org/matrix-rust-sdk/pull/6002.

Our information were wrong. `room_subscriptions` is definitely not sticky, we must send them for each request.

This patch removes `RoomSubscriptionState`. Subscriptions are not marked as “applied” anymore, they are always sent.


Hywan:
No problem! Thanks!

PR #6010, opened 2026-01-08
feat(ffi): Add `ffi::CreateRoomParameters::is_space`

jmartinesp:
This allows us to create spaces - rooms with a `RoomType::Space` - from the clients.

Signed-off-by:


jmartinesp:
All the code is in the FFI layer, so I'm not sure if we can actually test it.

---

Oh, we can. We just don't bother most of the time because it's FFI <-> SDK mappings.

PR #6016, opened 2026-01-12
fix(latest event): mark a recoverable failure to send as a sending item

bnjbvr:
In the send queue, failures to send can be classified into two
categories:

- permanent failures (e.g. invalid parameters)
- recoverable failures (e.g. network is down; server responded with a
  transient error code)

The latest event system would classify all the failures as "cannot be
sent", which is slightly incorrect if the failure was recoverable. In
this case, we should still consider the local event as being sent, as
the system should try to send it some time soon.

I think this should fix https://github.com/element-hq/element-x-android/issues/5885, I'll ask for some confirmation from one of the people who could repro the issue.

### DIFF
--- a/crates/matrix-sdk/src/latest_events/latest_event.rs
+++ b/crates/matrix-sdk/src/latest_events/latest_event.rs
@@ -881,9 +881,18 @@ impl LatestEventValueBuilder {
             // An error has occurred.
             //
             // Mark the latest event value matching `transaction_id`, and all its following values,
-            // as “cannot be sent”.
-            RoomSendQueueUpdate::SendError { transaction_id, .. } => {
-                buffer_of_values_for_local_events.mark_cannot_be_sent_from(transaction_id);
+            // as “cannot be sent” if the error isn't recoverable, and as "sending" if the error
+            // was recoverable.
+            RoomSendQueueUpdate::SendError { transaction_id, is_recoverable, .. } => {
+                if *is_recoverable {
+                    // If the room send queue error is recoverable, the send queue may retry to send
+                    // it in a short while, so the event should still be considered
+                    // sending. Leave it to that, at this point.
+                    buffer_of_values_for_local_events.mark_is_sending_from(transaction_id);
+                } else {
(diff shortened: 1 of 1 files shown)


crates/matrix-sdk/src/latest_events/latest_event.rs
```diff
@@ -892,9 +881,18 @@ impl LatestEventValueBuilder {
...
+            RoomSendQueueUpdate::SendError { transaction_id, is_recoverable, .. } => {
+                if *is_recoverable {
+                    // If the room send queue error is recoverable, the send queue may retry to send
+                    // it in a short while, so the event should still be considered
+                    // sending. Leave it to that, at this point.
+                    buffer_of_values_for_local_events.mark_is_sending_from(transaction_id);
+                } else {
+                    // If the error isn't recoverable, mark as a true "cannot be sent".
+                    buffer_of_values_for_local_events.mark_cannot_be_sent_from(transaction_id);
+                }
```
bnjbvr, line 895: This is the fix.

crates/matrix-sdk/src/latest_events/latest_event.rs
```diff
@@ -3112,6 +3123,84 @@ mod tests_latest_event_value_builder {
...
+        // (and thus will be retried soon, or as soon as network comes back).
+        let previous_value = {
+            let update = RoomSendQueueUpdate::SendError {
+                transaction_id: transaction_id_0.clone(),
+                error: Arc::new(Error::UnknownError("no more network".to_owned().into())),
+                is_recoverable: true,
+            };
+
+            // The `LatestEventValue` has changed, it still matches the latest local
+            // event but it's marked as “cannot be sent”.
```
Hywan: It's not marked as “cannot be sent”, I think it's a copy-paste issue :-].

bnjbvr: Good catch!

bnjbvr:
@jmartinesp confirmed (thanks!) this fixes the symptoms observable in https://github.com/element-hq/element-x-android/issues/5885. Merging now!

PR #6017, opened 2026-01-12
feat: Download room keys from backup before building key bundle.

kaylendog:
- Modifies `share_room_history` to fetch all room keys from backup if this has not done before.
- Introduces a new store in IndexedDB and a new table in SQLite to track whether the above has been done b efore.

Signed-off-by: Skye Elliot <actuallyori@gmail.com>

### DIFF
--- a/crates/matrix-sdk-crypto/CHANGELOG.md
+++ b/crates/matrix-sdk-crypto/CHANGELOG.md
@@ -13,6 +13,12 @@ All notable changes to this project will be documented in this file.
 - The `OutboundGroupSession` and `OlmMachine` now return the `EncryptionInfo` 
   used when encrypting raw events.
   ([#5936](https://github.com/matrix-org/matrix-rust-sdk/pull/5936))
+- Expose a new method `CryptoStore::has_downloaded_all_room_keys`, used to track whether the
+  client has previously downloaded historical room keys for a given room from key backup prior
+  to building an [MSC4268](https://github.com/matrix-org/matrix-spec-proposals/pull/4268) room
+  key bundle.
+  ([#6017](https://github.com/matrix-org/matrix-rust-sdk/pull/6017))
+
 
 ### Refactor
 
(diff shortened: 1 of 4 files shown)


crates/matrix-sdk-crypto/src/store/traits.rs
```diff
@@ -356,6 +356,10 @@ pub trait CryptoStore: AsyncTraitDeps {
         user_id: &UserId,
     ) -> Result<Option<StoredRoomKeyBundleData>, Self::Error>;
 
+    /// Get whether we have previously downloaded all room keys for a particular
```
andybalaam, line 359:
```suggestion
    /// Get whether we have previously downloaded all room keys for a particular room
```

crates/matrix-sdk-sqlite/src/crypto_store.rs
```diff
@@ -813,6 +823,16 @@ trait SqliteObjectCryptoStoreExt: SqliteAsyncConnExt {
...
     }
+
+    async fn has_downloaded_all_room_keys(&self, room_id: Key) -> Result<bool> {
+        self.query_row(
+            "SELECT EXISTS (SELECT 1 FROM room_key_backups_fully_downloaded WHERE room_id = ?)",
+            (room_id,),
+            |row| row.get(0),
+        )
+        .await
+        .map_err(Into::into)
```
andybalaam: I think I'd prefer `.await?` and then wrapping in `Ok()`, unless you really like this?

kaylendog:
> Before we merge I'd like to see how this is rendered in clients. Presumably this download could be really slow so they will need to display a spinner or equivalent.

[We do indeed](https://github.com/element-hq/element-x-ios/blob/9131c0976c69cd2528cb7ba25123a88f82037946/ElementX/Sources/Screens/InviteUsersScreen/InviteUsersScreenViewModel.swift#L81), tested and works!

PR #6021, opened 2026-01-13
feat(timeline): allow threaded timelines to be initialised with events from the server in both directions

Johennes:
This adds a enum for controlling how a timeline using `TimelineFocus::Thread` is initialised with events. It supports the following modes:

- events from the cache only (the only mode existing today)
- events loaded from `/relations` starting from the thread root up to a specified limit (for later paginating forwards)
- events loaded from `/relations` starting from the end of the thread root up to a specified limit (for later paginating backwards)

### DIFF
--- a/bindings/matrix-sdk-ffi/src/ruma.rs
+++ b/bindings/matrix-sdk-ffi/src/ruma.rs
@@ -1757,6 +1757,21 @@ impl From<RumaRoomAccountDataEvent<UnstableMarkedUnreadEventContent>> for RoomAc
     }
 }
 
+#[derive(Debug, Clone, Copy, uniffi::Enum)]
+pub enum Direction {
+    Backward,
+    Forward,
+}
+
+impl From<Direction> for ruma::api::Direction {
+    fn from(value: Direction) -> Self {
+        match value {
+            Direction::Backward => ruma::api::Direction::Backward,
+            Direction::Forward => ruma::api::Direction::Forward,
+        }
+    }
+}
+
 #[cfg(feature = "unstable-msc4274")]
 pub use galleries::*;
 
--- a/bindings/matrix-sdk-ffi/src/timeline/configuration.rs
+++ b/bindings/matrix-sdk-ffi/src/timeline/configuration.rs
@@ -13,6 +13,7 @@ use super::FocusEventError;
 use crate::{
     error::ClientError,
     event::{MessageLikeEventType, RoomMessageEventMessageType, StateEventType},
+    ruma::Direction,
 };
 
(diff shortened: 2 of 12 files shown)


bnjbvr:
Thanks for this draft PR; out of curiosity, what are the intended use cases for the new modes?

Johennes:
> Thanks for this draft PR; out of curiosity, what are the intended use cases for the new modes?

We're specifically interested in this case (and I think I'll remove the other variant from the PR):

> - events loaded from `/relations` starting from the end of the thread root up to a specified limit (for later paginating backwards)

The product idea is that when opening a thread, the client would display the bottom of the thread timeline and allow you to scroll up. Currently the SDK only loads cached events into the timeline. This would also allow to display the bottom of the thread timeline but only if all events in the thread are already in the cache.

Does that make sense?

bnjbvr:
> > Thanks for this draft PR; out of curiosity, what are the intended use cases for the new modes?
> 
> We're specifically interested in this case (and I think I'll remove the other variant from the PR):
> 
> > * events loaded from `/relations` starting from the end of the thread root up to a specified limit (for later paginating backwards)
> 
> The product idea is that when opening a thread, the client would display the bottom of the thread timeline and allow you to scroll up. Currently the SDK only loads cached events into the timeline. This would also allow to display the bottom of the thread timeline but only if all events in the thread are already in the cache.
> 
> Does that make sense?

I'm confused: currently, the implementation will load from the event cache, but any subsequent attempt to back-paginate the timeline will load from the network, if the in-memory event cache is empty or when it hits the start gap. Wouldn't that be sufficient for this use case?

Johennes:
> I'm confused: currently, the implementation will load from the event cache, but any subsequent attempt to back-paginate the timeline will load from the network, if the in-memory event cache is empty or when it hits the start gap. Wouldn't that be sufficient for this use case?

I think this is correct but it depends on actually having the latest thread event in the cache which I thought isn't guaranteed?

Johennes:
Turns out there was an issue in the code we use to integrate the SDK which (a) made me believe there is a problem in the SDK and (b) disguised the actual effect of my changes in this PR. 🤦‍♂️

Thanks for the help here. 👍

bnjbvr:
Sure thing, thanks for commenting back here! Is there anything in the documentation that could be improved, and that would've avoided the integration issue in the SDK?

Johennes:
No, I think I should've just started investigating the issue at the application layer rather than at the SDK layer. That would've prevented me working off of wrong assumptions. So in essence, the issue is on me mostly. 🙈

PR #6025, opened 2026-01-14
feat(spaces): Space filtering API

stefanceriu:
This PR introduces the initial version of the components required to filter a RoomList by Space data as defined in https://github.com/element-hq/element-meta/issues/2966

Space filters provide access to a custom subset of the space graph that can be used in tandem with the [`crate::RoomListService`] to narrow down the presented rooms.

They are limited to the first 2 levels of the graph, with the first level only containing direct descendants while the second holds the rest of them recursively.

Because Spaces need to be ordered by a special mechanism this PR also extracts that out of the SpaceRoomList implementation and makes it available to subspaces retrieved outside /hierarchy.


crates/matrix-sdk-ui/src/spaces/room.rs
```diff
@@ -132,4 +138,203 @@ impl SpaceRoom {
             via: vec![],
         }
     }
+
+    /// Sorts spare rooms by various criteria as defined in
```
poljar: I could use some spare rooms, but what are they? You meant space rooms?

crates/matrix-sdk-ui/src/spaces/mod.rs
```diff
@@ -475,15 +531,89 @@ impl SpaceService {
...
+
+        filters
     }
 }
 
+#[derive(Debug, Clone, PartialEq)]
+pub struct SpaceFilter {
+    /// The underlying [`SpaceRoom`]
+    pub space_room: SpaceRoom,
+
```
poljar, line 655: Some missing docs for public fields.

crates/matrix-sdk-ui/src/spaces/mod.rs
```diff
@@ -215,6 +226,34 @@ impl SpaceService {
         top_level_joined_spaces
     }
 
+    /// Space filters provide access to a custom subset of the space graph that
+    /// can be used in tandem with the [`crate::RoomListService`] to narrow
+    /// down the presented rooms.
```
poljar: That's a non-trivial interaction between these two components.

Can we point to the exact method which takes these space filters?

An example would do wonders here as well.

crates/matrix-sdk-ui/src/spaces/mod.rs
stefanceriu: 06652c8cb8b3e9e62a0d61cc9d7f5176f69c5119

crates/matrix-sdk-ui/src/spaces/mod.rs
stefanceriu: ecd05d50e3e9bd77884877936fb92127a27e780d

crates/matrix-sdk-ui/src/spaces/room.rs
stefanceriu: 076d523c75a5f2c391fcb1834bc391bdd1ccaa83

stefanceriu:
Thanks for the review! Bar some unfortunate styling and typos on my side that pollute the fixup commits it should be good to go for another review now!

PR #6031, opened 2026-01-15
feat(timeline): expose the user's threaded receipt on each thread summary

bnjbvr:
This is half of the work: this will load the threaded receipt for each thread, every time we add/update a timeline item for an event that had a thread summary. Since we don't know which of the private or the public receipt is the most advanced, we simply pass both, to start with; it's expected that this code dies later, when we fold it in into the event cache.

The second half consists in updating the thread summaries when a new read receipt event happens, and it contains receipts updates for threads.

### DIFF
--- a/crates/matrix-sdk-ui/src/timeline/controller/state_transaction.rs
+++ b/crates/matrix-sdk-ui/src/timeline/controller/state_transaction.rs
@@ -697,13 +697,13 @@ impl<'a, P: RoomDataProvider> TimelineStateTransaction<'a, P> {
         // As an exception to handle the latest "implicit" read receipt (which is the
         // latest event sent by the user): if the latest event has been sent by
         // the current user, then we consider that as a read receipt.
+        #[allow(clippy::collapsible_if)] // clippy has poor taste
         if let Some(ref latest_reply) = summary.latest_reply {
-            if let Some(event) = RoomDataProvider::load_event(room_data_provider, latest_reply)
+            if let Ok(event) = RoomDataProvider::load_event(room_data_provider, latest_reply)
                 .await
                 .inspect_err(|err| {
                     warn!("Failed to load thread latest event: {err}");
                 })
-                .ok()
             {
                 // Parse the sender.
                 if let Ok(Some(sender)) = event.raw().get_field::<OwnedUserId>("sender")
--- a/crates/matrix-sdk-ui/src/timeline/event_item/content/mod.rs
+++ b/crates/matrix-sdk-ui/src/timeline/event_item/content/mod.rs
@@ -74,6 +74,7 @@ pub use self::{
 use super::ReactionsByKeyBySender;
 
 /// The content of an [`EventTimelineItem`][super::EventTimelineItem].
+#[allow(clippy::large_enum_variant)]
 #[derive(Clone, Debug)]
 pub enum TimelineItemContent {
     MsgLike(MsgLikeContent),


crates/matrix-sdk-ui/src/timeline/controller/state_transaction.rs
```diff
@@ -692,9 +694,47 @@ impl<'a, P: RoomDataProvider> TimelineStateTransaction<'a, P> {
             } else {
                 None
             };
+
+            // Load the public and private read receipts for the user, in the thread. In the
+            // future, we might move this code in the event cache, so that read receipt
+            // handling happens there instead.
+
+            // As an exception to handle the "implicit" read receipt (which is the latest event
+            // sent by the user): if the latest event has been sent by the current user, then we
```
andybalaam: Is it OK that we are only looking at the latest event to see whether it was sent by us? I am guessing this information is only used to decide whether the specific latest event is read or unread, in which case this check is adequate (instead of searching back through the timeline to find the actual latest event we sent).

crates/matrix-sdk-ui/src/timeline/controller/read_receipts.rs
```diff
@@ -538,11 +538,68 @@ impl<P: RoomDataProvider> TimelineStateTransaction<'_, P> {
...
+                                            thread_summary.private_read_receipt_event_id =
+                                                Some(event_id.clone());
+                                        } else {
+                                            continue;
+                                        }
+
+                                        let mut new_content = new_item
+                                            .content
+                                            .as_msglike()
+                                            .expect("only msglike has a thread summary")
```
andybalaam: Is there no way to match on MsgLike earlier and avoid this expect?

crates/matrix-sdk-ui/src/timeline/controller/read_receipts.rs
```diff
@@ -538,11 +538,68 @@ impl<P: RoomDataProvider> TimelineStateTransaction<'_, P> {
...
+                                            thread_summary.public_read_receipt_event_id =
+                                                Some(event_id.clone());
+                                        } else if receipt_type == ReceiptType::ReadPrivate {
+                                            thread_summary.private_read_receipt_event_id =
+                                                Some(event_id.clone());
+                                        } else {
+                                            continue;
+                                        }
+
+                                        let mut new_content = new_item
```
andybalaam: Why do we need `new_content`? Haven't we updated the `thread_summary` of `new_item.content` already?

crates/matrix-sdk-ui/src/timeline/controller/state_transaction.rs
bnjbvr: Yeah, this is a simplified version of what the final feature would be (where we'd go through the timeline to find indeed the latest implicit read receipt). The better version will be done in #4113.

crates/matrix-sdk-ui/src/timeline/controller/read_receipts.rs
bnjbvr: We can indeed reuse it by mutably borrowing its fields. Good catch!

crates/matrix-sdk-ui/src/timeline/controller/read_receipts.rs
bnjbvr: Indeed, that implies cloning unconditionally, but that should be fine…

PR #6032, opened 2026-01-15
fix: `LeaveSpaceRoom` `is_last_admin` check

### LINKED ISSUE #6028: `LeaveSpaceHandle` `is_last_admin` returns a false negative if 1 or more admins have already left the room
## Steps to reproduce

- Create a space
- Then invite 2 new people
- Make on of them an admin and leave the other as a normal user
- Ban the admin or have them leave the room
- Try to leave the space

## Expected result
- Since you are the last joined admin of the space, the value of `isLastAdmin` should be true and the client should prevent you from leaving the space

## What happens instead
- You can leave the space since `isLastAdmin` is false.

The reason behind the issue is that the code below, only checks your membership state, and the Power Levels, but does not check the membership state of the other members that have the Admin PL, which means that if an admin was banned or had left, since the PLs remember members that left a room, the check will result in a false negative.

Another issue with this check is that this returns a false positive (or more than a false positive, it' s not enough to determine if a user can leave the room) when you are the only member left in the room, since you are indeed the last admin, but leaving a room as the only member is totally fine regardless of your PL: https://github.com/matrix-org/matrix-rust-sdk/issues/6030

Considering also the issue with v12 rooms: https://github.com/matrix-org/matrix-rust-sdk/issues/5832 is likely that this check would need to improved/refactored to also account for the membership state of the room ids that have been found, and not just the PLs (which wouldn't be enough in v12 rooms where creators have an infinite PL that doesn't appear in the PL)
https://github.com/matrix-org/matrix-rust-sdk/blob/fbff1ee99a5122e22384e823e0f6ad08910bff55/crates/matrix-sdk-ui/src/spaces/leave.rs#L44-L79

### LINKED ISSUE #6030: Include `joined_members_count` in the `LeaveSpaceRoom`
This is an improvement required since `isLastAdmin` while does not allow to leave a room/space when you are indeed the last admin, but you should still be allowed to leave the room if you are the last admin and also the only joined member of the room/space.

Velin92:
- added the `joined_members_count` since if the count is only 1  the user should still be able to leave the room even if they are the last owner, since they will be also the last member.
- Added a check for the membership state, so that we are sure that the last admin check is only operated on admins that are in the joined state, to avoid triggering false negatives due to other admins having left the room
- Since for the v12 fix we need a product decision will fix it in a follow up PR

fixes #6030 
fixes #6028

### DIFF
--- a/crates/matrix-sdk-ui/CHANGELOG.md
+++ b/crates/matrix-sdk-ui/CHANGELOG.md
@@ -9,8 +9,8 @@ All notable changes to this project will be documented in this file.
 ### Bug Fixes
 
 - Fix the `is_last_admin` check in `LeaveSpaceRoom` since it was not 
-  accounting for the membership state. Also added the joined_members_count.
-  [#6032](https://github.com/matrix-org/matrix-rust-sdk/pull/6032/commits)
+  accounting for the membership state. Also added the `joined_members_count`.
+  [#6032](https://github.com/matrix-org/matrix-rust-sdk/pull/6032)
 - [**breaking**] `LatestEventValue::Local { is_sending: bool }` is replaced
   by [`state: LatestEventValueLocalState`] to represent 3 states: `IsSending`,
   `HasBeenSent` and `CannotBeSent`.
--- a/crates/matrix-sdk-ui/src/spaces/leave.rs
+++ b/crates/matrix-sdk-ui/src/spaces/leave.rs
@@ -14,6 +14,7 @@
 
 use matrix_sdk::{Client, RoomState, room::RoomMemberRole};
 use ruma::{Int, OwnedRoomId, events::room::member::MembershipState};
+use tracing::info;
 
 use crate::spaces::{Error, SpaceRoom};
 
@@ -56,32 +57,33 @@ impl LeaveSpaceHandle {
             }
 
             if !room.are_members_synced() {
+                info!("Syncing members for room {} to check if can leave", room.room_id());
                 _ = room.sync_members().await.ok();
             }
 
-            let users_to_power_levels = room.users_with_power_levels().await;
-            let admin_ids = users_to_power_levels
-                .iter()
+            let admin_ids = room
+                .users_with_power_levels()
+                .await
+                .into_iter()
                 .filter(|(_, power_level)| {
-                    let Some(power_level) = Int::new(**power_level) else {
+                    let Some(power_level) = Int::new(*power_level) else {
                         return false;
                     };
 
                     RoomMemberRole::suggested_role_for_power_level(power_level.into())
                         == RoomMemberRole::Administrator
                 })
-                .map(|p| p.0)
-                .collect::<Vec<_>>();
+                .map(|p: (ruma::OwnedUserId, i64)| p.0);
 
             let mut joined_admin_ids = Vec::new();
             for admin_id in admin_ids {
-                if let Ok(Some(member)) = room.get_member_no_sync(admin_id).await
+                if let Ok(Some(member)) = room.get_member_no_sync(&admin_id).await
                     && *member.membership() == MembershipState::Join
                 {
                     joined_admin_ids.push(admin_id);
                 }
             }
-            let is_last_admin = joined_admin_ids == vec![room.own_user_id()];
+            let is_last_admin = joined_admin_ids == [room.own_user_id()];
 
             rooms.push(LeaveSpaceRoom {
                 space_room: SpaceRoom::new_from_known(&room, 0),


crates/matrix-sdk-ui/src/spaces/leave.rs
```diff
@@ -53,25 +59,49 @@ impl LeaveSpaceHandle {
...
+            if !room.are_members_synced() {
+                _ = room.sync_members().await.ok();
+            }
+
             let users_to_power_levels = room.users_with_power_levels().await;
+            let power_levels = room.power_levels_or_default().await;
+            let mut creator_ids = Vec::new();
+            if let Some(create) = room.create_content() {
+                // There must be a better way to be sure about room version 12 creators
+                if create.room_version.as_str() == "12" {
```
Velin92: Is there a better way to perform this check? I just want to be sure that now creators have infinite PL (which means they won't be normally included in the users_to_power_level vector)

crates/matrix-sdk-ui/src/spaces/leave.rs
```diff
@@ -53,25 +55,51 @@ impl LeaveSpaceHandle {
...
                     let Some(power_level) = Int::new(**power_level) else {
                         return false;
                     };
 
-                    RoomMemberRole::suggested_role_for_power_level(power_level.into())
-                        == RoomMemberRole::Administrator
+                    // We should consider owners as those users with power level equal to
+                    // or greater than the power level required to send a tombstone event.
+                    power_level
+                        >= power_levels.for_state(ruma::events::StateEventType::RoomTombstone)
```
Velin92: @mxandreas I think this is the best way to determine if a user should or not leave the room, if they are able to upgrade it or not, but let me know if we want to for an admin check or a >= PL 150 for v12 room.

crates/matrix-sdk-ui/src/spaces/leave.rs
Hywan: Room's version can be any strings, so I think this is correct. You may prefer to use [`RoomVersionId::V12`](https://docs.rs/ruma/latest/ruma/enum.RoomVersionId.html#variant.V12) though:

```suggestion
                if create.room_version == RoomVersionId::V12 {
```

crates/matrix-sdk-ui/src/spaces/leave.rs
mxandreas: I need a bit more context to answer - e.g. what's the reason for this PR, what problem are we solving?

crates/matrix-sdk-ui/src/spaces/leave.rs
Velin92: decided to drop this for now since we probably need to decide what to check... if the user has a PL of over 150 (which is not clearly defined in the spec and in the SDK) or if their PL is above the tombstoned check. So I reduced the scope of the PR of only fixing the other 2 bugs

crates/matrix-sdk-ui/src/spaces/leave.rs
```diff
@@ -53,9 +55,12 @@ impl LeaveSpaceHandle {
                 continue;
             }
 
-            let users_to_power_levels = room.users_with_power_levels().await;
+            if !room.are_members_synced() {
+                _ = room.sync_members().await.ok();
+            }
```
Velin92, line 62: Not sure what is better in terms of performance:
- Check if the members are synced and sync if not
- Then get each member through the `no_sync` function

OR
- Just get each member through a synced version of the function

@Hywan

bindings/matrix-sdk-ffi/CHANGELOG.md
```diff
@@ -13,6 +13,9 @@ All notable changes to this project will be documented in this file.
   ignored if they matched the default power level values specified by the spec: these may not be 
   the same in the homeserver and result in rooms with incorrect power levels being created.
   ([#6034](https://github.com/matrix-org/matrix-rust-sdk/pull/6034))
+- Fix the `is_last_admin` check in `LeaveSpaceRoom` since it was not 
+  accounting for the membership state. Also added the joined_members_count.
```
Hywan, line 17:
```suggestion
  accounting for the membership state. Also added the `joined_members_count`.
```

crates/matrix-sdk-ui/src/spaces/leave.rs
```diff
@@ -53,9 +55,12 @@ impl LeaveSpaceHandle {
                 continue;
             }
 
-            let users_to_power_levels = room.users_with_power_levels().await;
+            if !room.are_members_synced() {
+                _ = room.sync_members().await.ok();
```
Hywan, line 61: Please add a log (`info!`) here, I'm having an intuition it could be useful in a close future.

crates/matrix-sdk-ui/src/spaces/leave.rs
```diff
@@ -53,9 +55,12 @@ impl LeaveSpaceHandle {
...
+            }
 
-            let is_last_admin = users_to_power_levels
+            let users_to_power_levels = room.users_with_power_levels().await;
+            let admin_ids = users_to_power_levels
```
Hywan: Nit: The `users_to_power_levels` is used in a single place: we can inline it.

```suggestion
            let admin_ids = room.users_with_power_levels().await
```

crates/matrix-sdk-ui/src/spaces/leave.rs
```diff
@@ -66,12 +71,22 @@ impl LeaveSpaceHandle {
                         == RoomMemberRole::Administrator
                 })
                 .map(|p| p.0)
-                .collect::<Vec<_>>()
-                == vec![room.own_user_id()];
+                .collect::<Vec<_>>();
```
Hywan: No need to collect into a `Vec` here. It's going to create an allocation and many `memcpy` for no reasons. You can simply use the iterator (`.map(…)`) inside the `for` loop:

```rust
let admin_ids = ….iter().filter(…).map(…);

for admin_id in admin_ids {
    …
}
```

crates/matrix-sdk-ui/src/spaces/leave.rs
```diff
@@ -66,12 +71,22 @@ impl LeaveSpaceHandle {
...
+
+            let mut joined_admin_ids = Vec::new();
+            for admin_id in admin_ids {
+                if let Ok(Some(member)) = room.get_member_no_sync(admin_id).await
+                    && *member.membership() == MembershipState::Join
+                {
+                    joined_admin_ids.push(admin_id);
+                }
+            }
+            let is_last_admin = joined_admin_ids == vec![room.own_user_id()];
```
Hywan: I believe [you can compare a `Vec<T>` with a `&[T]` (a slice)](https://doc.rust-lang.org/std/vec/struct.Vec.html#impl-PartialEq%3C%26%5BU%5D%3E-for-Vec%3CT,+A%3E):

```rust
           let is_last_admin = joined_admin_ids == &[room.own_user_id()];
```

That way you're saving one `Vec` allocation.

crates/matrix-sdk-ui/CHANGELOG.md
```diff
@@ -8,6 +8,9 @@ All notable changes to this project will be documented in this file.
 
 ### Bug Fixes
 
+- Fix the `is_last_admin` check in `LeaveSpaceRoom` since it was not 
+  accounting for the membership state. Also added the joined_members_count.
```
Hywan:
```suggestion
  accounting for the membership state. Also added the `joined_members_count`.
```

crates/matrix-sdk-ui/CHANGELOG.md
```diff
@@ -8,6 +8,9 @@ All notable changes to this project will be documented in this file.
...
+  [#6032](https://github.com/matrix-org/matrix-rust-sdk/pull/6032/commits)
```
Hywan:
```suggestion
  [#6032](https://github.com/matrix-org/matrix-rust-sdk/pull/6032/)
```

crates/matrix-sdk-ui/src/spaces/leave.rs
Velin92: done, buy had to use into_iter() because the compiler was complaining that I was borrowing a variable that was immediately going to be dropped

crates/matrix-sdk-ui/src/spaces/leave.rs
```diff
@@ -56,32 +57,33 @@ impl LeaveSpaceHandle {
             }
 
             if !room.are_members_synced() {
+                info!("Syncing members for room {} to check if can leave", room.room_id());
```
Hywan, line 60: It's preferable to do a:

```
info!(room_id = room.room_id(), "Syncing …");
```

PR #6034, opened 2026-01-16
fix(ffi): Use `RoomPowerLevelsContentOverride` when creating rooms

jmartinesp:
This fixes an issue that prevented default values coming from the SDK from being uploaded in the create room request, which could mean rooms created with the wrong power levels if the default values in the homeserver didn't match those of the SDK.

i.e. I create a room with power level `invite: 0`, which is the default in the spec, but the homeserver has overridden this default value to `50`.

1. The SDK won't add the `invite: 0` value to the request.
2. The server will set the default value since none was provided.
3. In the end you get a room with `invite: 50` power level even though you specified `0` in the client.

Signed-off-by:

### DIFF
--- a/bindings/matrix-sdk-ffi/CHANGELOG.md
+++ b/bindings/matrix-sdk-ffi/CHANGELOG.md
@@ -8,6 +8,11 @@ All notable changes to this project will be documented in this file.
 
 ### Bug Fixes
 
+- `Client::create_room` now uses `RoomPowerLevelsContentOverride` under the hood instead of 
+  `RoomPowerLevelsEventContent` to be able to explicitly set values which would previously be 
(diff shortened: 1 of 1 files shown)


bindings/matrix-sdk-ffi/src/client.rs
```diff
@@ -2280,7 +2259,7 @@ impl From<PowerLevels> for RoomPowerLevelsEventContent {
                 (event_type, (*power_level).into())
             })
             .collect();
-
+        power_levels.notifications = value.notifications.map(Into::into).unwrap_or_default();
```
stefanceriu: Nit: move these next to the other ones not requiring iteration.

Issue #5317, opened 2025-07-01
Ignoring a user doesn't remove their messages from the room list


pixlwave:
When ignoring a user with `Client::ignore_user` (in the FFI), this user and all of their messages are hidden from the timeline. However… If they sent the last message in one or more rooms, then these messages remain visible in the room list until newer messages come in for each room.

Hywan:
Thanks for opening the issue.
You're saying that the message doesn't disappear from the latest event, is that correct?

pixlwave:
Yep you’re right, and my bad I forgot to mention I was talking about the room list. Updated.

Hywan:
Fixed!

PR #6035, opened 2026-01-16
fix(sdk): The `LatestEventValue` is erased when a room has been emptied

Hywan:
This patch is twofold:

1. It ensures that a `LatestEventValue` is erased when a room has been emptied,
2. It implements a test for #5317: the `LatestEventValue` is re-computed when a user is ignored.

Best to review one patch at a time.

---

- Fix https://github.com/matrix-org/matrix-rust-sdk/issues/5317

### DIFF
--- a/crates/matrix-sdk/tests/integration/latest_event.rs
+++ b/crates/matrix-sdk/tests/integration/latest_event.rs
@@ -1,0 +1,81 @@
+use assert_matches::assert_matches;
+use matrix_sdk::{
+    latest_events::LatestEventValue,
+    linked_chunk::{ChunkIdentifier, LinkedChunkId, Position, Update},
+    test_utils::mocks::MatrixMockServer,
+};
+use matrix_sdk_test::{async_test, event_factory::EventFactory};
+use ruma::{event_id, room_id, user_id};
+use tokio::task::yield_now;
+
+#[async_test]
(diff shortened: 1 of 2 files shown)


crates/matrix-sdk/src/latest_events/latest_event/builder.rs
```diff
@@ -659,10 +659,20 @@ impl LatestEventValueBuilder {
         own_user_id: &UserId,
         power_levels: Option<&RoomPowerLevels>,
     ) -> Option<LatestEventValue> {
+        // If we are computing a value from the Event Cache, it's because we have
+        // received an update from the Event Cache. This update falls in two categories:
+        // either an event has been added or updated, or the room has been emptied.
+        //
+        // We consider the room has been emptied by default. If we are able to scan at
+        // least one in-memory event, we consider the room has not been emptied.
```
poljar, line 56: What does it mean that a room has been emptied? How do you empty a room?

edit: Oh, by calling `clear_all_rooms()`? Can we reuse the terminology of this function and perhaps even mention it in the comment?

poljar (approved):
Makes sense, please just clarify the comment about what's happening bit.

crates/matrix-sdk/src/latest_events/latest_event/builder.rs
Hywan: Sure!

PR #6036, opened 2026-01-16
refactor `is_last_admin` to `is_last_owner `to account for v12 rooms

### LINKED ISSUE #5832: `LeaveSpaceHandle` doesn't check if user is last with PL150 for v12
### The problem

According to the following code:

https://github.com/matrix-org/matrix-rust-sdk/blob/main/crates/matrix-sdk-ui/src/spaces/leave.rs#L66

We check if the user that is leaving a room is the last admin, and if we so we prevent them to leave so that client can prompt them to promote another admin in their place, and then leave.

This would be fine for any room version below v12, because since v12 being an admin is not sufficient anymore, since a new role called Creator (with infinite power level) has been introduced and the `upgrade` room state event can now only be sent by people that have PL >= 150 as stated [here](https://github.com/matrix-org/matrix-spec-proposals/pull/4289/files#diff-2232eb7f3d7cadcb7c7b305c7c1c816208f2a2afb7870cdf2bfcbafec530475aR131-R155)

### Short term solution

The simplest short term solution is to have the Leave Handle check if the last user has PL >= 150 if we are on a v12 room (for previous versions we can leave it as it right now), and then have clients prompt the user to promote someone to that specific power level

### Long term solution

I think to avoid this issue from happening again we should introduce a new role level, called Owner (as EX and EC already do client side) to differentiate an admin with PL 100 from one with PL 150. Sadly as of right now there is no spec for such role, but I think this would make things more clear, and avoid mistakes like this in the future.

Velin92:
fixes #5832 

The v12 rules for privileged creators is now included in the check, this refactors the `is_last_admin` into an `is_last_owner` check, that will work just as before on v11 rooms where the owner is just the admin, but will work differently on v12 rooms where an owner is considered either a creator or someone with PL 150 (which is the default PL for tombstoning a room on v12 that is always set on room creation).

Also we are returning a `are_creators_privileged` bool, so that the client knows if the owner is referring to admins or to creators and whoever has PL 150 and above.

### DIFF
--- a/crates/matrix-sdk-ui/src/spaces/leave.rs
+++ b/crates/matrix-sdk-ui/src/spaces/leave.rs
@@ -82,7 +82,7 @@ impl LeaveSpaceHandle {
                     };
 
                     if are_creators_privileged {
-                        i64::from(power_level) >= 150
+                        power_level >= ruma::int!(150)
                     } else {
                         RoomMemberRole::suggested_role_for_power_level(power_level.into())
                             == RoomMemberRole::Administrator


crates/matrix-sdk-ui/src/spaces/leave.rs
```diff
@@ -70,25 +83,31 @@ impl LeaveSpaceHandle {
                         return false;
                     };
 
-                    RoomMemberRole::suggested_role_for_power_level(power_level.into())
-                        == RoomMemberRole::Administrator
+                    if are_creators_privileged {
+                        i64::from(power_level) >= 150
```
Velin92: Don't really like this.
I remember @jmartinesp mentioning that we force tombstone event on v12 to alway be 150, so I imagine there is some kind of constant stored somewhere for this? But couldn't find it.

crates/matrix-sdk-ui/src/spaces/leave.rs
```diff
@@ -70,25 +83,31 @@ impl LeaveSpaceHandle {
...
+                        RoomMemberRole::suggested_role_for_power_level(power_level.into())
+                            == RoomMemberRole::Administrator
+                    }
                 })
                 .map(|p: (ruma::OwnedUserId, i64)| p.0);
 
-            let mut joined_admin_ids = Vec::new();
-            for admin_id in admin_ids {
-                if let Ok(Some(member)) = room.get_member_no_sync(&admin_id).await
+            owner_ids.extend(non_creator_owner_ids);
```
Velin92: @Hywan is this okay in terms of performance or there is a better solution to join the creators with this iterator?

crates/matrix-sdk-ui/src/spaces/leave.rs
jmartinesp: It's actually in Ruma, and there is no constant 😓 : https://github.com/ruma/ruma/blob/d8c4d11875c0811df4c46f18ffc4405196b1399f/crates/ruma-events/src/room/power_levels.rs#L137

crates/matrix-sdk-ui/src/spaces/leave.rs
```diff
@@ -59,7 +61,18 @@ impl LeaveSpaceHandle {
...
 
-            let admin_ids = room
+            let mut owner_ids = Vec::new();
+            let mut are_creators_privileged = false;
+            if let Some(create) = room.create_content() {
+                let rules = create.room_version.rules().unwrap_or(ROOM_VERSION_RULES_FALLBACK);
+                if rules.authorization.explicitly_privilege_room_creators {
+                    are_creators_privileged = true;
+                    owner_ids.push(create.creator);
+                    owner_ids.append(&mut create.additional_creators.clone());
```
Hywan, line 71: You may want to use `owner_ids.extend(create.additional_creators);`, not tested but I think it can save you one `.clone()` here. It's also closer to your intent.

crates/matrix-sdk-ui/src/spaces/leave.rs
```diff
@@ -81,8 +81,12 @@ impl LeaveSpaceHandle {
                         return false;
                     };
 
-                    RoomMemberRole::suggested_role_for_power_level(power_level.into())
-                        == RoomMemberRole::Administrator
+                    if are_creators_privileged {
+                        i64::from(power_level) >= 150
```
Hywan, line 85: Please use a `u64`, I don't think we can have negative power levels, do we?
Also, maybe add an inline comment to explain where this 150 comes from?

PR #6038, opened 2026-01-16
feat(sdk): New state event candidate for `LatestEventValue`: `m.room.member` with `join`

Hywan:
This patch adds a new state event candidate for `LatestEventValue`: `m.room.member` when the `membership` is `join` and the `state_key` is the current user ID. Put differently: when the current user joins a room, we are able to compute a `LatestEventValue`.

---

- Address https://github.com/matrix-org/matrix-rust-sdk/issues/5726


Hywan:
No problem!
Thanks!

PR #6039, opened 2026-01-16
Bump `async-stream` to `0.3.6`

### LINKED ISSUE #4599: Bumping async_stream causes test failures
As discovered in #4596, updating `async-stream` cause some tests to fail.

The reason the `identity_status_changes` test fail is because the event handler for the `m.room.member` event registered over here:

https://github.com/matrix-org/matrix-rust-sdk/blob/8d27b0c811c96d00673eb43b0a813bb3bcf26542/crates/matrix-sdk/src/room/identity_status_changes.rs#L178-L193

isn't called anymore. As we can see, we create a drop guard for the event handler. So once the guard is dropped the handler is removed. The drop guard gets moved into a stream over here:

https://github.com/matrix-org/matrix-rust-sdk/blob/8d27b0c811c96d00673eb43b0a813bb3bcf26542/crates/matrix-sdk/src/room/identity_status_changes.rs#L93-L123

This means that the handler should be alive as long as the stream is alive, but as it seems something changed between `async_stream` version `0.3.5` and `0.3.6` and the drop guard gets dropped before the stream is.

Downgrading `async_stream` fixes the issue. The only relevant change seems to be in this PR: https://github.com/tokio-rs/async-stream/commit/97d1d3e6de46ef768567aecceb6235847bb65e35.

mgoldenberg:
## Overview

When bumping `async-stream` from `0.3.5` to `0.3.6`, a number of tests associated with `IdentityStatusChanges` begin to fail. This pull request bumps the dependency and also introduces a fix to ensure the tests continue to pass.

## Problem

When bumping the dependency, a subtle issue is introduced where the stream returned by [`IdentityStatusChanges::create_stream`](https://github.com/matrix-org/matrix-rust-sdk/blob/90390d7488b5f9d27d995dcdfcc97039c6753e43/crates/matrix-sdk/src/room/identity_status_changes.rs#L89-L123) no longer takes ownership of an associated `EventHandlerDropGuard` which is created in the scope of the function. Consequentially, the `EventHandlerDropGuard` is dropped at the end of the function, and the stream returned does not receive any `RoomIdentityChange` events.

```rust
impl IdentityStatusChanges {
   pub async fn create_stream(
        room: Room,
    ) -> Result<impl Stream<Item = Vec<IdentityStatusChange>>> {
        // -- snip --

        let state = IdentityStatusChanges {
            room_identity_state: RoomIdentityState::new(room).await,
            // This is being dropped at the end of the function
            _drop_guard: drop_guard,
        };

        // Previous versions of `async-stream` seemed to cause the call
        // to `stream!` below to take ownership of `state`, but in the
        // latest version it does not
        Ok(stream!({            
            // -- snip --
        }))
    }
}
```

This problem is a result of fixing some soundness bugs in `async-stream` (see tokio-rs/async-stream#109).

## Solution

The easiest way to fix this issue was to force the returned stream to take ownership of the `EventHandlerDropGuard` with a trivial assignment statement.

```rust
impl IdentityStatusChanges {
   pub async fn create_stream(
        room: Room,
    ) -> Result<impl Stream<Item = Vec<IdentityStatusChange>>> {
        // -- snip --

        let state = IdentityStatusChanges {
            room_identity_state: RoomIdentityState::new(room).await,
            _drop_guard: drop_guard,
        };

        Ok(stream!({
            // Force enclosing stream to take ownership of state, so that
            // _drop_guard does not get dropped.
            let mut state = state;
            
            // -- snip --
        }))
    }
}
```

## Other Thoughts

While the solution above works, it does seem like there might be subtler things going on here. For instance, adding an empty `Drop` implementation for `IdentityStatusChanges` prevents the `_drop_guard` from being dropped - I'm really not sure why that would be the case...

---
Fixes #4599.

### DIFF
--- a/crates/matrix-sdk/src/room/identity_status_changes.rs
+++ b/crates/matrix-sdk/src/room/identity_status_changes.rs
@@ -94,12 +94,16 @@ impl IdentityStatusChanges {
         let mut unprocessed_stream = combine_streams(identity_updates, room_member_events);
         let own_user_id = room.client.user_id().ok_or(Error::InsufficientData)?.to_owned();
 
-        let mut state = IdentityStatusChanges {
+        let state = IdentityStatusChanges {
             room_identity_state: RoomIdentityState::new(room).await,
             _drop_guard: drop_guard,
         };
 
         Ok(stream!({
+            // Force enclosing stream to take ownership of state, so that
+            // _drop_guard does not get dropped.
+            let mut state = state;
+
             let mut current_state =
                 filter_for_initial_update(state.room_identity_state.current_state(), &own_user_id);
 


mgoldenberg:
By the way, I didn't add a line in any of the change logs - was that correct? The dependency was at the workspace level, but there doesn't seem to be a change log at that level, so I wasn't sure where it might be recorded.

poljar:
> By the way, I didn't add a line in any of the change logs - was that correct? The dependency was at the workspace level, but there doesn't seem to be a change log at that level, so I wasn't sure where it might be recorded.

Bumping a dependency isn't that important, if we do consider it important enough I would say the `matrix-sdk` changelog would be the most appropriate place for it as this dep is mostly used there.

Alternatively you could put it into the changelog of every crate which is affected.

But again, I don't think this warrants an entry.

PR #3741, opened 2024-07-19
sdk: basic support for sending and stopping live location shares

torrybr:
This MR introduces basic functionality for sending and stopping live location shares, as outlined in the [Matrix Spec Proposal #3489](https://github.com/matrix-org/matrix-spec-proposals/pull/3489). This foundational support allows users to share their live location, setting the stage for further enhancements.

**Key Changes**
- Live Location Start: Saving and sending an initial `beacon_info` state event.
- Live Location Stop: Stopping a users existing `beacon_info` state event from the state store.

These changes were adapted from a previous MR discussion, available for reference [here](https://github.com/matrix-org/matrix-rust-sdk/pull/3621).

**Follow-up**

Future MRs will expand on this initial implementation, using a layered approach to fully realize the functionality described in the spec above and align with ongoing community discussions.

This does not send any `beacon` message events. This MR simply introduces `beacon_info` support with the intention of follow on MRs to build on this basic functionality.

Signed-off-by: Torry Brelsford <tb@brehl.dev>

PR #6041, opened 2026-01-17
refactor(base): Remove unused `beacons` from `BaseRoomInfo`

zecakeh:
This field was added already unused in the initial PR https://github.com/matrix-org/matrix-rust-sdk/pull/3741 for live location sharing. By "unused" I mean that it is written to but never read.

The follow up live location PRs didn't make use of it either:

- https://github.com/matrix-org/matrix-rust-sdk/pull/3771
- https://github.com/matrix-org/matrix-rust-sdk/pull/3794
- https://github.com/matrix-org/matrix-rust-sdk/pull/4025

Note that it is not exposed in the public API so no one can use it outside of the crate.

Maybe ping @torrybr which added this field, to clarify what it was intended to be used for in the first place, or if it is still intended to be used but the feature hasn't been implemented yet?

### DIFF
--- a/crates/matrix-sdk-base/src/room/room_info.rs
+++ b/crates/matrix-sdk-base/src/room/room_info.rs
@@ -28,7 +28,6 @@ use ruma::{
     events::{
         AnyStrippedStateEvent, AnySyncStateEvent, AnySyncTimelineEvent, StateEventType,
         SyncStateEvent,
-        beacon_info::BeaconInfoEventContent,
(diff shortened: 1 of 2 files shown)


Hywan (approved):
I think we should merge this PR. We can still revert it in case this field was necessary in the future.

Waiting on another review from @stefanceriu.

stefanceriu:
I'm just waiting for an answer from the original creator, there's no rush here.

stefanceriu:
Guess it's been long enough 🚀 

Thank your keeping an eye out, Kevin 🙏

PR #6043, opened 2026-01-19
Spaces: Add a method to reset a SpaceRoomList.

pixlwave:
We now have the ability to add/remove space children in #5871, however as the `SpaceRoomList` is based on `/hierarchy` it isn't aware of any changes made by the space service. This PR helps by making a way to reset the list so that the next `paginate()` call starts from the beginning again.

Whilst we could listen for `m.space.child`/`parent` events in the sync loop, this wouldn't help if you removed a room where you aren't a member.

### DIFF
--- a/bindings/matrix-sdk-ffi/src/spaces.rs
+++ b/bindings/matrix-sdk-ffi/src/spaces.rs
@@ -156,7 +156,7 @@ impl SpaceService {
     }
 }
 
-/// The `SpaceRoomList`represents a paginated list of direct rooms
+/// The `SpaceRoomList` represents a paginated list of direct rooms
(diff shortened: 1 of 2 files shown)


bindings/matrix-sdk-ffi/src/spaces.rs
```diff
@@ -156,7 +156,7 @@ impl SpaceService {
     }
 }
 
-/// The `SpaceRoomList`represents a paginated list of direct rooms
+/// The `SpaceRoomList` represents a paginated list of direct rooms
```
stefanceriu, line 159: Thanks 👍

PR #6044, opened 2026-01-19
fix: Correctly store rooms with downloaded keys in SQLite and IndexedDB.

kaylendog:
While https://github.com/matrix-org/matrix-rust-sdk/pull/6017 is mostly functional, there are two issues:

- I did not process `changes.room_key_bundles_fully_downloaded` in `matrix-sdk-sqlite`, meaning any updates made via `Changes` would not be persisted;
- I used a non-encrypting `JsValue` serialisation for the same field in `matrix-sdk-indexeddb`, which causes errors when passed to the decryption-enabled deserializer.

Solutions:

- Process the aforementioned changes such that keys are added to SQLite;
- Use a non-encrypting deserialiser, since this is effectively a hash-set, and the contents aren't sensitive.

Signed-off-by: Skye Elliot <actuallyori@gmail.com>

### DIFF
--- a/crates/matrix-sdk-crypto/CHANGELOG.md
+++ b/crates/matrix-sdk-crypto/CHANGELOG.md
@@ -18,6 +18,7 @@ All notable changes to this project will be documented in this file.
   to building an [MSC4268](https://github.com/matrix-org/matrix-spec-proposals/pull/4268) room
   key bundle.
   ([#6017](https://github.com/matrix-org/matrix-rust-sdk/pull/6017))
+  ([#6044](https://github.com/matrix-org/matrix-rust-sdk/pull/6044))
 
 
 ### Refactor
--- a/crates/matrix-sdk-indexeddb/CHANGELOG.md
+++ b/crates/matrix-sdk-indexeddb/CHANGELOG.md
@@ -13,8 +13,10 @@ All notable changes to this project will be documented in this file.
   `IndexeddbStores::open`. Additionally, allow feature flags for each of the
   stores to be used independent of and in combination with the others.
   ([#5946](https://github.com/matrix-org/matrix-rust-sdk/pull/5946))
-- Implement new method `CyptoStore::has_downloaded_all_room_keys`.
+- Implement new method `CyptoStore::has_downloaded_all_room_keys`, and process
+  `room_key_backups_fully_downloaded` field in `Changes`.
   ([#6017](https://github.com/matrix-org/matrix-rust-sdk/pull/6017))
+  ([#6044](https://github.com/matrix-org/matrix-rust-sdk/pull/6044))
 
 ### Bug Fixes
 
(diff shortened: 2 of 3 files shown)


stefanceriu:
Can we get a bit more in terms of description here? What issue does it fix, what was the problem, why did you make the changes you made etc.?

kaylendog:
> Can we get a bit more in terms of description here? What issue does it fix, what was the problem, why did you make the changes you made etc.?

Yes!

While https://github.com/matrix-org/matrix-rust-sdk/pull/6017 is mostly functional, there are two issues:
- I did not process `changes.room_key_bundles_fully_downloaded` in `matrix-sdk-sqlite`, meaning any updates made via `Changes` would not be persisted;
- I used a non-encrypting `JsValue` serialisation for the same field in `matrix-sdk-indexeddb`, which causes errors when passed to the decryption-enabled deserializer.

I have copied this to the PR description.

crates/matrix-sdk-indexeddb/src/crypto_store/mod.rs
```diff
@@ -745,7 +745,8 @@ impl IndexeddbCryptoStore {
             for room_id in &changes.room_key_backups_fully_downloaded {
                 let key =
                     self.serializer.encode_key(keys::ROOM_KEY_BACKUPS_FULLY_DOWNLOADED, room_id);
-                room_store.put(key, JsValue::from_bool(true));
+                let value = self.serializer.serialize_value(&true)?;
+                room_store.put(key, value);
```
richvdh: I do wonder if a more sensible solution might be to fix the deserializing code not to attempt decryption. It feels like the encryption isn't buying anything here except CPU overhead. Note that `OLM_HASHES` is an existing store that just stores a placeholder in the `value` field.

kaylendog: That seems like a better idea, I'll go for that.

kaylendog: Addressed in https://github.com/matrix-org/matrix-rust-sdk/pull/6044/commits/7ea05b973747e7bc7abb7466c5df28032f808ec6

richvdh: Thanks. Could you update the PR description?

crates/matrix-sdk-sqlite/src/crypto_store.rs
richvdh, line 1: Could you update the changelogs to reference this PR? It's probably fine just to add this PR to the existing lines for #6017.

crates/matrix-sdk-indexeddb/src/crypto_store/mod.rs
richvdh: oh you did already. I'm a crank.

PR #6048, opened 2026-01-20
feat(timeline): add filter for membership and profile changes

Johennes:
Use case: Display membership changes (join, leave, etc.) in the timeline but suppress profile changes (display name or avatar URL). This is currently not possible with `TimelineEventTypeFilter` because both types of changes have the same event type (`m.room.member`).

This pull request introduces a new `TimelineEventFilter` for filtering on either the event type or parts of its content. Content filters are only added for membership and profile changes but more enum variants can be added in future.

### DIFF
--- a/bindings/matrix-sdk-ffi/src/timeline/configuration.rs
+++ b/bindings/matrix-sdk-ffi/src/timeline/configuration.rs
@@ -46,7 +46,7 @@ impl TimelineEventTypeFilter {
     }
 }
 
-/// A timeline filter that in- or excludes events based on their type or
+/// A timeline filter that includes or excludes events based on their type or
 /// content.
 #[derive(uniffi::Object)]
 pub struct TimelineEventFilter {
--- a/crates/matrix-sdk-ui/src/timeline/event_filter.rs
+++ b/crates/matrix-sdk-ui/src/timeline/event_filter.rs
@@ -68,7 +68,7 @@ impl TimelineEventCondition {
     /// `true` if the condition matches or `false` otherwise.
     fn matches(&self, event: &AnySyncTimelineEvent) -> bool {
         match self {
-            Self::EventType(event_type) => &event.event_type() == event_type,
+            Self::EventType(event_type) => event.event_type() == *event_type,
             Self::MembershipChange => match event {
                 AnySyncTimelineEvent::State(AnySyncStateEvent::RoomMember(
                     SyncStateEvent::Original(ev),


bindings/matrix-sdk-ffi/CHANGELOG.md
```diff
@@ -8,6 +8,10 @@ All notable changes to this project will be documented in this file.
 
 ### Bug Fixes
 
+- Add `TimelineFilter::EventFilter` for filtering events based on their type or
+  content. For content filtering, only membership and profile change filters
+  are available as of now.
+  ([#6048](https://github.com/matrix-org/matrix-rust-sdk/pull/6048/))
```
bnjbvr: Would be a feature rather than a bug fix, right?

crates/matrix-sdk-ui/src/timeline/tests/event_filter.rs
```diff
@@ -261,6 +264,287 @@ async fn test_event_type_filter_exclude_messages() {
...
+    timeline.handle_live_event(f.room_name("A new room name").sender(&ALICE)).await;
+    timeline.handle_live_event(f.room_name("A new room name (again)").sender(&ALICE)).await;
+    // And a different state event
+    timeline.handle_live_event(f.room_topic("A new room topic").sender(&ALICE)).await;
+
+    // The timeline should contain only the room name events
+    let event_items: Vec<Arc<TimelineItem>> = timeline.get_event_items().await;
+    let text_message_items_count = event_items.iter().filter(is_text_message_item).count();
+    let room_name_items_count = event_items.iter().filter(is_room_name_item).count();
+    let room_topic_items_count = event_items.iter().filter(is_room_topic_item).count();
```
bnjbvr: nit, for all these variables, can you name them `num_text_message_items` rather than using the `_count` suffix, please?

crates/matrix-sdk-ui/src/timeline/tests/event_filter.rs
```diff
@@ -261,6 +264,287 @@ async fn test_event_type_filter_exclude_messages() {
...
+        .await;
+
+    // The timeline should contain only the invite and join events
+    let event_items: Vec<Arc<TimelineItem>> = timeline.get_event_items().await;
+    let text_message_items_count = event_items.iter().filter(is_text_message_item).count();
+    let room_name_items_count = event_items.iter().filter(is_room_name_item).count();
+    let room_topic_items_count = event_items.iter().filter(is_room_topic_item).count();
+    let membership_change_items_count =
+        event_items.iter().filter(is_membership_change_item).count();
+    let profile_change_items_count = event_items.iter().filter(is_profile_change_item).count();
```
bnjbvr: ditto for `num_` vs `_count`

crates/matrix-sdk-ui/src/timeline/tests/event_filter.rs
```diff
@@ -261,6 +264,287 @@ async fn test_event_type_filter_exclude_messages() {
...
+        .await;
+
+    // The timeline should contain only the display name and avatar URL changes
+    let event_items: Vec<Arc<TimelineItem>> = timeline.get_event_items().await;
+    let text_message_items_count = event_items.iter().filter(is_text_message_item).count();
+    let room_name_items_count = event_items.iter().filter(is_room_name_item).count();
+    let room_topic_items_count = event_items.iter().filter(is_room_topic_item).count();
+    let membership_change_items_count =
+        event_items.iter().filter(is_membership_change_item).count();
+    let profile_change_items_count = event_items.iter().filter(is_profile_change_item).count();
```
bnjbvr: ditto

crates/matrix-sdk-ui/src/timeline/tests/event_filter.rs
```diff
@@ -261,6 +264,287 @@ async fn test_event_type_filter_exclude_messages() {
...
+        .await;
+
+    // The timeline should contain everything except for the invite and join events
+    let event_items: Vec<Arc<TimelineItem>> = timeline.get_event_items().await;
+    let text_message_items_count = event_items.iter().filter(is_text_message_item).count();
+    let room_name_items_count = event_items.iter().filter(is_room_name_item).count();
+    let room_topic_items_count = event_items.iter().filter(is_room_topic_item).count();
+    let membership_change_items_count =
+        event_items.iter().filter(is_membership_change_item).count();
+    let profile_change_items_count = event_items.iter().filter(is_profile_change_item).count();
```
bnjbvr: ditto

crates/matrix-sdk-ui/src/timeline/tests/event_filter.rs
```diff
@@ -261,6 +264,287 @@ async fn test_event_type_filter_exclude_messages() {
...
+
+    // The timeline should contain everything except for the display name and avatar
+    // URL changes
+    let event_items: Vec<Arc<TimelineItem>> = timeline.get_event_items().await;
+    let text_message_items_count = event_items.iter().filter(is_text_message_item).count();
+    let room_name_items_count = event_items.iter().filter(is_room_name_item).count();
+    let room_topic_items_count = event_items.iter().filter(is_room_topic_item).count();
+    let membership_change_items_count =
+        event_items.iter().filter(is_membership_change_item).count();
+    let profile_change_items_count = event_items.iter().filter(is_profile_change_item).count();
```
bnjbvr: you know the drill :P

crates/matrix-sdk-ui/CHANGELOG.md
```diff
@@ -8,6 +8,10 @@ All notable changes to this project will be documented in this file.
 
 ### Bug Fixes
 
+- Add `TimelineEventFilter` for filtering events based on their type or
+  content. For content filtering, only membership and profile change filters
+  are available as of now.
+  ([#6048](https://github.com/matrix-org/matrix-rust-sdk/pull/6048/))
```
bnjbvr: ditto, not a bugfix but a new feature

bnjbvr (approved):
Thanks, looks good in principle. Tiny nits to address, and this is good to land.

Out of curiosity: could we merge the `TimelineFilter::EventTypeFilter` into `TimelineFilter::EventFilter`, as the latter seems to be a nice generalization of the former?

crates/matrix-sdk-ui/src/timeline/tests/event_filter.rs
```diff
@@ -261,6 +264,287 @@ async fn test_event_type_filter_exclude_messages() {
     assert_eq!(room_topic_items_count, 1);
 }
 
+#[async_test]
+async fn test_event_filter_include_only_room_names() {
+    // Only return room name events
+    let event_filter = TimelineEventFilter::Include(vec![TimelineEventCondition::EventType(
+        TimelineEventType::RoomName,
+    )]);
```
bnjbvr, line 272: Can you add (or tweak) a test, so that there are 2 conditions in the `Include` vector? I was surprised that the specification said "would match *any* condition" and not *all* conditions, so having a test expliciting this behavior would be nice.

And ditto, I'd like a test with 2 conditions in the `Exclude` vector please :)

bindings/matrix-sdk-ffi/CHANGELOG.md
Johennes: Err, of course. 🤦‍♂️ 

Fixed in 89f364ea5c06b12527840dfbd66c7a559323ab18.

crates/matrix-sdk-ui/CHANGELOG.md
Johennes: Fixed in 89f364ea5c06b12527840dfbd66c7a559323ab18.

crates/matrix-sdk-ui/src/timeline/tests/event_filter.rs
Johennes: These were basically copy/pasted from the existing tests in this file. I have applied the renaming in both the old and new tests with 2a5eb36b27d4803b0263cd8f701ff133938d972f.

crates/matrix-sdk-ui/src/timeline/tests/event_filter.rs
Johennes: Fixed in 2a5eb36b27d4803b0263cd8f701ff133938d972f.

crates/matrix-sdk-ui/src/timeline/tests/event_filter.rs
Johennes: Fixed in 2a5eb36b27d4803b0263cd8f701ff133938d972f.

crates/matrix-sdk-ui/src/timeline/tests/event_filter.rs
Johennes: Fixed in 2a5eb36b27d4803b0263cd8f701ff133938d972f.

crates/matrix-sdk-ui/src/timeline/tests/event_filter.rs
Johennes: Fixed in 2a5eb36b27d4803b0263cd8f701ff133938d972f.

crates/matrix-sdk-ui/src/timeline/tests/event_filter.rs
Johennes: I tried to mimic `TimelineEventTypeFilter` there where it uses `contains` on the specified event types (which is equivalent to "any match").

Have added two new tests in a9a9f61663232abe31969fe80c4d42db7350f600.

Johennes:
> Out of curiosity: could we merge the `TimelineFilter::EventTypeFilter` into `TimelineFilter::EventFilter`, as the latter seems to be a nice generalization of the former?

Yes, I think functionally the event type filter is obsolete now and could be replaced by `TimelineFilter::EventFilter`. The only downside is that the new filter requires an extra level of enum wrapping because of the intermediate `TimelineEventCondition` type. I can remove the old filter if that's not an issue though.

Otherwise, I've addressed all comments and have squashed the three small fixups.

bnjbvr:
> Yes, I think functionally the event type filter is obsolete now and could be replaced by TimelineFilter::EventFilter. The only downside is that the new filter requires an extra level of enum wrapping because of the intermediate TimelineEventCondition type. I can remove the old filter if that's not an issue though.

Sweet, maybe we can do this as a followup, to not make this PR grow overly large.

bindings/matrix-sdk-ffi/src/timeline/configuration.rs
```diff
@@ -45,6 +46,37 @@ impl TimelineEventTypeFilter {
     }
 }
 
+/// A timeline filter that in- or excludes events based on their type or
```
bnjbvr:
```suggestion
/// A timeline filter that includes or excludes events based on their type or
```

crates/matrix-sdk-ui/src/timeline/event_filter.rs
```diff
@@ -0,0 +1,100 @@
...
+    ///
+    /// # Arguments
+    ///
+    /// * `event` - The event to test the condition against.
+    ///
+    /// # Returns
+    /// `true` if the condition matches or `false` otherwise.
+    fn matches(&self, event: &AnySyncTimelineEvent) -> bool {
+        match self {
+            Self::EventType(event_type) => &event.event_type() == event_type,
```
bnjbvr: It's slightly more idiomatic to dereference rather than creating a new ref:

```suggestion
            Self::EventType(event_type) => event.event_type() == *event_type,
```

PR #6049, opened 2026-01-20
Spaces: Remove the redundant LeaveSpaceRoom.joined_members_count property

pixlwave:
#6032 added this property, but turns out its already available via `space_room.num_joined_members`.

I haven't included a changelog as that PR was only merged a few hours ago so it doesn't seem at all worthwhile to say that something is being removed that has only just appeared.


Velin92:
I think we should at least update the changelog to remove the part about mentioning the `joined_members_count`

pixlwave:
> I think we should at least update the changelog to remove the part about mentioning the `joined_members_count`

Good spot, done [55a6c94..3bc3284](https://github.com/matrix-org/matrix-rust-sdk/compare/55a6c94479395767d6bd7111abe47d84dbfab96f..3bc32840be47f7d61596e89f6f836da6b2f4cca1)

PR #6050, opened 2026-01-20
feat(timeline): enable focusing a thread root using `TimelineFocus::Event`

Johennes:
Use case: Focus a timeline on a thread, initialize it with the thread root and let the timeline paginate _forward_ from there as the user scrolls. In other words, display a thread starting from the oldest messages.

This isn't currently possible with `TimelineFocus::Thread` because it only supports backwards pagination. It's also not possible with `TimelineFocus::Event` because the latter doesn't recognize that it's on a thread if the focused event is the thread root.

This PR extends `TimelineFocus::Event` to allow controlling the initialization behavior for threads, making it possible to mark the focused event as being the root of a thread.

CC @bnjbvr

### DIFF
--- a/bindings/matrix-sdk-ffi/CHANGELOG.md
+++ b/bindings/matrix-sdk-ffi/CHANGELOG.md
@@ -23,6 +23,9 @@ All notable changes to this project will be documented in this file.
 
 ### Features
 
+- [**breaking**] Extend `TimelineFocus::Event` to allow marking the target
+  event as the root of a thread.
+  [#6050](https://github.com/matrix-org/matrix-rust-sdk/pull/6050)
 - [**breaking**] Remove `TimelineFilter::EventTypeFilter` which has been replaced by
   the more generic `TimelineFilter::EventFilter`. Users of `TimelineEventTypeFilter::include`
   and `TimelineEventTypeFilter::exclude` can switch to `TimelineEventFilter::include_event_types`
--- a/bindings/matrix-sdk-ffi/src/timeline/configuration.rs
+++ b/bindings/matrix-sdk-ffi/src/timeline/configuration.rs
@@ -2,7 +2,7 @@ use std::sync::Arc;
 
 use matrix_sdk_ui::timeline::{
     event_filter::{TimelineEventCondition, TimelineEventFilter as InnerTimelineEventFilter},
-    TimelineReadReceiptTracking,
+    TimelineEventFocusThreadMode, TimelineReadReceiptTracking,
 };
 use ruma::{
     events::{AnySyncTimelineEvent, TimelineEventType},
@@ -120,8 +120,8 @@ pub enum TimelineFocus {
         event_id: String,
         /// The number of context events to load around the focused event.
         num_context_events: u16,
-        /// Whether to hide in-thread replies from the live timeline.
-        hide_threaded_events: bool,
+        /// How to handle threaded events.
+        thread_mode: TimelineEventFocusThreadMode,
     },
     Thread {
         /// The thread root event ID to focus on.
@@ -141,18 +141,14 @@ impl TryFrom<TimelineFocus> for matrix_sdk_ui::timeline::TimelineFocus {
     ) -> Result<matrix_sdk_ui::timeline::TimelineFocus, Self::Error> {
         match value {
             TimelineFocus::Live { hide_threaded_events } => Ok(Self::Live { hide_threaded_events }),
-            TimelineFocus::Event { event_id, num_context_events, hide_threaded_events } => {
+            TimelineFocus::Event { event_id, num_context_events, thread_mode } => {
                 let parsed_event_id =
                     EventId::parse(&event_id).map_err(|err| FocusEventError::InvalidEventId {
                         event_id: event_id.clone(),
                         err: err.to_string(),
                     })?;
 
-                Ok(Self::Event {
-                    target: parsed_event_id,
-                    num_context_events,
-                    hide_threaded_events,
-                })
+                Ok(Self::Event { target: parsed_event_id, num_context_events, thread_mode })
             }
             TimelineFocus::Thread { root_event_id } => {
                 let parsed_root_event_id = EventId::parse(&root_event_id).map_err(|err| {
--- a/crates/matrix-sdk-ui/CHANGELOG.md
+++ b/crates/matrix-sdk-ui/CHANGELOG.md
@@ -21,6 +21,9 @@ All notable changes to this project will be documented in this file.
 
 ### Features
 
+- [**breaking**] Extend `TimelineFocus::Event` to allow marking the target
+  event as the root of a thread.
+  ([#6050](https://github.com/matrix-org/matrix-rust-sdk/pull/6050))
 - [**breaking**] Remove `TimelineEventTypeFilter` which has been replaced by
   the more generic `TimelineEventFilter`.
   ([#6070](https://github.com/matrix-org/matrix-rust-sdk/pull/6070/))
--- a/crates/matrix-sdk-ui/src/timeline/controller/mod.rs
+++ b/crates/matrix-sdk-ui/src/timeline/controller/mod.rs
@@ -72,7 +72,7 @@ use super::{
 };
 use crate::{
     timeline::{
-        MsgLikeContent, MsgLikeKind, Room, TimelineEventFilterFn,
+        MsgLikeContent, MsgLikeKind, Room, TimelineEventFilterFn, TimelineEventFocusThreadMode,
         algorithms::rfind_event_by_item_id,
         controller::decryption_retry_task::compute_redecryption_candidates,
         date_dividers::DateDividerAdjuster,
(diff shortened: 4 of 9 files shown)


bindings/matrix-sdk-ffi/src/timeline/configuration.rs
```diff
@@ -127,6 +127,38 @@ impl TryFrom<TimelineFocus> for matrix_sdk_ui::timeline::TimelineFocus {
...
+        /// When the target event is not part of a thread, whether to
+        /// hide in-thread replies from the live timeline. Has no effect
+        /// when the target event is part of a thread.
+        ///
+        /// This should be set to true when the client can create
+        /// [`TimelineFocus::Thread`]-focused timelines from the thread roots
+        /// themselves and doesn't use the [`Self::ThreadRoot`] mode.
+        hide_threaded_events: bool,
+    },
+}
```
Johennes: I'm not super happy with the naming and shape of this but was struggling to come up with something better. Happy to hear alternative suggestions.

bnjbvr:
Thanks! I'll be happy to do the review whenever you think it's in a good shape, so feel free to assign the review to me at that time :)

Johennes:
@bnjbvr I think I cannot directly assign you myself but it's basically ready for input from my side. I was just going to wait on the CI to pass before undrafting it.

Johennes:
The complement crypto test failures seem unrelated to my change here:

```
❌ TestOnRejoinBobCanSeeButNotDecryptHistoryInPublicRoom (2.41s)
❌ TestOnRejoinBobCanSeeButNotDecryptHistoryInPublicRoom/{rust_hs1}|{rust_hs1} (2.41s)
```

crates/matrix-sdk-ui/src/timeline/controller/mod.rs
```diff
@@ -509,14 +509,22 @@ impl<P: RoomDataProvider> TimelineController<P> {
...
-                let thread_root_event_id = events
-                    .iter()
-                    .find(
-                        |event| {
-                            if let Some(id) = event.event_id() { id == *event_id } else { false }
-                        },
-                    )
-                    .and_then(|event| extract_thread_root(event.raw()));
+                let (thread_root_event_id, hide_threaded_events) = match thread_mode {
+                    TimelineEventFocusThreadMode::ThreadRoot => (Some(event_id.clone()), false),
```
bnjbvr: There's one failing mode here: if the focussed-on event id is actually an event that's part of a thread, this will try to open a thread which root is that in-thread-but-not-root event. We should probably handle this, and add a test case making sure that we don't fail this case.

bindings/matrix-sdk-ffi/CHANGELOG.md
```diff
@@ -8,6 +8,9 @@ All notable changes to this project will be documented in this file.
 
 ### Bug Fixes
 
+- [**breaking**] Extend `TimelineFocus::Event` to allow marking the target
+  event as the root of a thread.
+  [#6050](https://github.com/matrix-org/matrix-rust-sdk/pull/6050)
```
bnjbvr: I would put this in the new features section

crates/matrix-sdk-ui/src/timeline/mod.rs
```diff
@@ -156,6 +153,27 @@ pub enum TimelineFocus {
     PinnedEvents { max_events_to_load: u16, max_concurrent_requests: u16 },
 }
 
+/// Options for controlling the behaviour of [`TimelineFocus::Event`]
+/// for threaded events.
+#[derive(Clone, Debug, PartialEq)]
```
bnjbvr, line 159: Could we put the `uniffi` derive here, making it optional based on the presence of the `uniffi` feature? That way, we could avoid the duplication of this struct in the FFI layer.

crates/matrix-sdk-ui/CHANGELOG.md
```diff
@@ -8,6 +8,9 @@ All notable changes to this project will be documented in this file.
 
 ### Bug Fixes
 
+- [**breaking**] Extend `TimelineFocus::Event` to allow marking the target
+  event as the root of a thread.
+  [#6050](https://github.com/matrix-org/matrix-rust-sdk/pull/6050)
```
bnjbvr: ditto (new feature)

crates/matrix-sdk-ui/src/timeline/controller/mod.rs
Johennes: Do you want it to fail in this case or to refocus on the actual thread root?

bindings/matrix-sdk-ffi/CHANGELOG.md
Johennes: Yes, sorry. Same story as in the other PR. I completely missed the section. 🤦‍♂️ 

Fixed in fccc7d3af3e48fd79f01fc0a09b1e7b6718c1c1d.

crates/matrix-sdk-ui/CHANGELOG.md
Johennes: Fixed in fccc7d3af3e48fd79f01fc0a09b1e7b6718c1c1d.

crates/matrix-sdk-ui/src/timeline/mod.rs
Johennes: Holy cow! I didn't know that was possible. 🤯 

Have made the change in cc7b303dddf9f8fbaa7b54d47e80e40891ebd9c7.

crates/matrix-sdk-ui/src/timeline/controller/mod.rs
bnjbvr: Good question. My intuition is that if a user called this function with `ThreadRoot`, they're expecting a threaded timeline with the backwards and forwards navigation capabilities. And, if we turned it into an error, what would happen is that the user would have to do some extra detection that the event is part of a thread or not… which is exactly what we're doing a few lines below, so that would be wasted.

So I think we should make it focus on the actual thread root, in this case. Maybe we could rename this variant `EnsureThreaded` or `AlwaysThreaded` or `ForceThreaded`, so that this intent is clearer? If the event is a thread root (or equivalently, not in a thread), this becomes a threaded timeline focused on the start; if the event is in a thread, this makes sure we end up in a threaded timeline.

What do you think?

Johennes: Ok. I've renamed the variant to `ForceThread` and made it so that when the target event is itself a threaded event, it'll focus on the thread but still remain at the position of the target event initially. This makes the behaviors the same as with `Automatic` and `hide_threaded_events = false` which seemed sensible to me.

crates/matrix-sdk-ui/src/timeline/controller/mod.rs
```diff
@@ -518,6 +518,18 @@ impl<P: RoomDataProvider> TimelineController<P> {
                     )
                     .and_then(|event| extract_thread_root(event.raw()));
 
+                // Determine the timeline's threading behavior.
+                let (thread_root_event_id, hide_threaded_events) = match thread_mode {
+                    TimelineEventFocusThreadMode::ForceThread => {
+                        // If the event is part of a thread, use its thread root. Otherwise,
+                        // assume the event itself is the thread root.
+                        (extracted_thread_root.or(Some(event_id.clone())), false)
```
bnjbvr: I think you can use `.or_else(|| event_id.clone())` here, which is better in terms of performance, as it'll do the cloning only when the option is `None`, instead of cloning eagerly and maybe wasting the cloned event id.

crates/matrix-sdk-ui/CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ All notable changes to this project will be documented in this file.
 
 ### Bug Fixes
 
+  [#6050](https://github.com/matrix-org/matrix-rust-sdk/pull/6050)
```
bnjbvr: nit: remnant of a bad merge conflict?

crates/matrix-sdk-ui/src/timeline/controller/mod.rs
Johennes: Oh, good point. Fixed in 2b6a7da1debabae587ff17104dc186d48659fddb.

crates/matrix-sdk-ui/CHANGELOG.md
Johennes: Oops, yes. 🤦 

Fixed in 97b60d94aca1290687ae1550d8f4e278443e851c.

Johennes:
Have applied the two small fixes, merged in `main` and then rebased with auto-squash. I pushed in-between so hopefully it should still be possible to follow the changes for the review.

crates/matrix-sdk-ui/tests/integration/timeline/mod.rs
```diff
@@ -217,7 +252,36 @@ async fn test_timeline_is_threaded() {
...
+                thread_mode: TimelineEventFocusThreadMode::Automatic { hide_threaded_events: true },
+            })
+            .build()
+            .await
+            .unwrap();
+        assert!(timeline.is_threaded());
+    }
+
+    {
+        // An event-focused timeline, focused on a thread root, is also threaded \o/
```
bnjbvr, line 264: Out of curiosity: is that test different from the previous one (at line 171), and if so, how so?

Johennes: This one uses `num_context_events: 2`, the other `num_context_events: 0`. In fact, I think most / all of the test are just duplicated with and without context events included.

bnjbvr: I see, thanks!

PR #6053, opened 2026-01-20
Support TLS v1.3 on all platforms

### LINKED ISSUE #5800: Add TLS 1.3 support
`rustls-tls` by default uses the `native-tls` backend which currently doesn't have support for TLS 1.3 (see https://github.com/sfackler/rust-native-tls/issues/140). This currently breaks compatibility with home-servers that only support TLS 1.3 or Ciphers which are only available on TLS 1.3.

The `rustls-tls` backend option uses the [rustls](https://crates.io/crates/rustls) crate in the background which does have support for TLS 1.3, therefore enabling this backend in favor of the default `native-tls` backend should fix the compatibility issues with homeserver that only support TLS 1.3.

mgoldenberg:
**Note:** _this pull request has a companion pull request in the [`complement-crypto`](https://github.com/matrix-org/complement-crypto/pull/229) repository, which must be merged in conjunction with this one._

_Before merging, this should be tested in conjunction with the Element X iOS client to ensure that TLS v1.3 is working properly._ @stefanceriu has agreed to work on this.

## Overview

The primary change in this pull request upgrades the `reqwest` dependency to its latest version, which defaults to using `rustls` with support for `rustls-platform-verifier` instead of `native-tls` (see [`reqwest@0.13.0`](https://github.com/seanmonstar/reqwest/releases/tag/v0.13.0)). The benefit here is that `rustls` supports TLS v1.3 on all platforms, whereas [`native-tls` does not](https://github.com/sfackler/rust-native-tls/pull/278).

Additionally, this pull request makes `rustls` the default TLS implementation in all the crates in this repository.

This will be particularly helpful with element-hq/element-x-ios#786.

## Changes

- `reqwest` bumped to `0.13.1`
    - The API for adding/replacing certificates has changed a bit, so this required some updating in `HttpSettings::make_client`
- `oauth2-reqwest` added in favor of `oauth2/reqwest` 
    - This is required in order to be compatible with `reqwest^0.13`
    - _**`oauth2-reqwest` is currently in alpha release, so it probably makes sense to let this stabilize a bit.**_ For details, see https://github.com/ramosbugs/oauth2-rs/issues/333#issuecomment-3906712203. 
- `getrandom` bumped to `0.3.4`
    - This is required in order to be compatible with `oauth2@5.1.0`
- `proptest` bumped to `1.9.0`
    - This is required in order to be compatible with `getrandom@0.3.4` 
- Make `rustls` the default TLS implementation

### Mirror feature flag names?

A number of feature flags have been replaced in the dependencies above.

1. _**`reqwest/rustls-tls` => `reqwest/rustls`**_ - this is simply a name change, but is semantically identical (see [`reqwest@0.13.0`](https://github.com/seanmonstar/reqwest/releases/tag/v0.13.0)).
2. _**`getrandom/js` => `getrandom/wasm_js`**_ - the semantics here have changed slightly, but it seems to just make it easier to enable the `wasm_js` backend (see [`getrandom@0.3.4`](https://github.com/rust-random/getrandom/blob/master/CHANGELOG.md#major-change-to-wasm_js-backend)).

At any rate, I have updated references to these flags in each of the various `Cargo.toml` files, but have not changed the names of our exposed features to mimic those in the dependencies.

Any thoughts or preferences on whether to mirror those names? That would, of course, result in a breaking change.

### Default to using `rustls`? Deprecate `native-tls`?

Now that the dependencies have all been bumped, we can use `rustls` on all platforms. Should this be the new default given that `native-tls` will very likely never support TLS v1.3 on Apple devices? And should `native-tls` be deprecated as a result?

**UPDATE:** _The consensus here seems to be that we should default to using `rustls`, but that `native-tls` should still be available._

---
Fixes #5800.

### DIFF
--- a/crates/matrix-sdk/src/http_client/native.rs
+++ b/crates/matrix-sdk/src/http_client/native.rs
@@ -191,13 +191,6 @@ impl HttpSettings {
             http_client = http_client.danger_accept_invalid_certs(true)
         }
 
-        if !self.additional_root_certificates.is_empty() {
-            info!(
-                "Adding {} additional root certificates to the HTTP client",
-                self.additional_root_certificates.len()
-            );
-        }
-
         if self.disable_built_in_root_certificates {
             info!("Built-in root certificates disabled in the HTTP client.");
             http_client = http_client.tls_certs_only(self.additional_root_certificates.clone());


ShadowRZ:
@mgoldenberg FYI for the current situation on oauth2-rs: https://github.com/ramosbugs/oauth2-rs/pull/334#issuecomment-3869594380

mgoldenberg:
> @mgoldenberg FYI for the current situation on oauth2-rs: [ramosbugs/oauth2-rs#334 (comment)](https://github.com/ramosbugs/oauth2-rs/pull/334#issuecomment-3869594380)

Thank you, I should have put an update here! The Rust team discussed this comment yesterday and decided we'd wait a little to see if any information is released about the new crate.

> I'm planning to release a separate crate that provides a `reqwest 0.13` client for `oauth2` [^1]
 
If movement is slow, we'll follow the author's prescription of implementing the traits ourselves.

> as mentioned in [the docs](https://docs.rs/oauth2/latest/oauth2/#http-clients), any user of this crate can bring their own HTTP client by providing either a closure or an implementation of the `AsyncHttpClient`/`SyncHttpClient` traits. this solution can also be used to support a different version of `reqwest`/`ureq`/etc. if you need to upgrade to `reqwest 0.13` urgently, I suggest taking that route. [^1]

[^1]: From [comment](https://github.com/ramosbugs/oauth2-rs/pull/334#issuecomment-3869594380)

Neustradamus:
Good news rust-native-tls has now the TLS 1.3 support!

Linked to:
- https://github.com/rust-native-tls/rust-native-tls/pull/353
- https://github.com/rust-native-tls/rust-native-tls/pull/291
- https://github.com/rust-native-tls/rust-native-tls/pull/278
- https://github.com/rust-native-tls/rust-native-tls/pull/235
- https://github.com/rust-native-tls/rust-native-tls/pull/159
- https://github.com/rust-native-tls/rust-native-tls/issues/140
- https://github.com/retep998/winapi-rs/pull/963
- https://github.com/retep998/winapi-rs/pull/859
- https://github.com/olback/barium/issues/14
- https://github.com/element-hq/element-x-ios/issues/786
- https://github.com/matrix-org/matrix-rust-sdk/pull/6053
- https://github.com/deltachat/deltachat-ios/issues/1693

frebib:
> > @mgoldenberg FYI for the current situation on oauth2-rs: [ramosbugs/oauth2-rs#334 (comment)](https://github.com/ramosbugs/oauth2-rs/pull/334#issuecomment-3869594380)
> 
> Thank you, I should have put an update here! The Rust team discussed this comment yesterday and decided we'd wait a little to see if any information is released about the new crate.
> 
> > I'm planning to release a separate crate that provides a `reqwest 0.13` client for `oauth2` [1](#user-content-fn-1-68daded367a137865ba25ad0b16f32af)

There's an alpha implementation now https://github.com/ramosbugs/oauth2-rs/issues/333#issuecomment-3906712203

.deny.toml
```diff
@@ -29,6 +29,11 @@ allow = [
     "Zlib",
 ]
 
+[[licenses.clarify]]
```
ShadowRZ: I think it's incorrect to say that aws-lc-sys doesn't contain OpenSSL license, as aws-lc-sys binds to aws-lc, which currently still contains OpenSSL licensed code.

Also https://github.com/aws/aws-lc/issues/2203

ShadowRZ:
@mgoldenberg As I explained in review, I don't think it's right to specify that aws-lc-sys doesn't use the OpenSSL license.

In fact it does have parts covered by the OpenSSL license via aws-lc project.

Could you please clarify why you did https://github.com/matrix-org/matrix-rust-sdk/pull/6053/changes/1b88edc8c6cc1e8470d09a165979af533a27d909..ec4f6af3acbcf827bde8b83e6e4e7d9ef0d9050e in the review?

mgoldenberg:
@ShadowRZ, I'm not sure I see a review from you?

But at any rate, the Rust team recommended I use `cargo deny`'s license clarification feature, instead of allowing OpenSSL licenses.

My understanding from the conversation was that the license could be recast based on the following statement from [OpenSSL](https://openssl-library.org/source/license/).
> For the 3.0 release, and later releases derived from that, [the Apache License v2](https://openssl-library.org/source/license/apache-license-2.0.txt) applies. This also applies to the git “master” branch.

Presumably this is preferable because adding a new license might involve some legal hurdles? I'm not terribly familiar with licensing issues, etc. so I'm mostly just following guidance here. I'd be happy to change the `clarify` clause into an `allow` clause, but would want @poljar to weigh in here.

crates/matrix-sdk-common/Cargo.toml
```diff
@@ -65,7 +65,7 @@ wasm-bindgen-test.workspace = true
 
 [target.'cfg(target_family = "wasm")'.dev-dependencies]
 # Enable the JS feature for getrandom.
-getrandom = { workspace = true, default-features = false, features = ["js"] }
+getrandom = { workspace = true, default-features = false, features = ["wasm_js"] }
```
Hywan, line 68: Good to me.

crates/matrix-sdk/src/http_client/native.rs
Hywan: This log is no longer relevant here as the additional root certificates are no longer added here. Either move it done (merge it with `info!("Built-in root certificates…")`), or remove it.

Cargo.lock
```diff
@@ -4660,6 +4843,47 @@ dependencies = [
...
+ "url",
+ "wasm-bindgen",
+ "wasm-bindgen-futures",
+ "web-sys",
+ "webpki-roots 1.0.2",
+]
+
+[[package]]
+name = "reqwest"
+version = "0.13.1"
```
Hywan: We have 2 versions of `reqwest` now. Is it normal?

Cargo.lock
```diff
@@ -7103,6 +7380,15 @@ dependencies = [
  "windows-link 0.1.1",
 ]
 
+[[package]]
+name = "windows-sys"
+version = "0.45.0"
```
Hywan, line 7313: We have 2 versions of `windows-sys` now. Is it fixable?

Cargo.lock
```diff
@@ -7139,6 +7425,21 @@ dependencies = [
  "windows-link 0.2.1",
 ]
 
+[[package]]
+name = "windows-targets"
+version = "0.42.2"
```
Hywan, line 7358: Two versions of `windows-targets` now. I didn't check, but it may be caused by two versions of `reqwest` probably.

Cargo.toml
```diff
@@ -59,7 +59,8 @@ insta = { version = "1.44.1", features = ["json", "redactions"] }
 itertools = "0.14.0"
 js-sys = "0.3.82"
 mime = "0.3.17"
-oauth2 = { version = "5.1.0", default-features = false, features = ["reqwest", "timing-resistant-secret-traits"] }
+oauth2 = { version = "5.0.0", default-features = false, features = ["timing-resistant-secret-traits"] }
+oauth2-reqwest = { version = "0.1.0-alpha.1" }
```
Hywan: Can you add `default-features = false` too here, in case of. We try to have as few deps as possible.

niclashoyer:
@mgoldenberg 

> My understanding from the conversation was that the license could be recast based on the following statement from [OpenSSL](https://openssl-library.org/source/license/).
> 
> > For the 3.0 release, and later releases derived from that, [the Apache License v2](https://openssl-library.org/source/license/apache-license-2.0.txt) applies. This also applies to the git “master” branch.

that is correct regarding OpenSSL, but reqwest 0.13 is using rustls with aws-lc by default and aws-lc is based on a specific version of BoringSSL, which was based on OpenSSL < 3.0.

I raised the issue regarding reqwest / aws-lc again here: https://github.com/aws/aws-lc/issues/2203#issuecomment-3925152892

My understanding is, that as aws-lc is largely a fork/extension of BoringSSL which was licensed with OpenSSL this still applies to aws-lc *until* the maintainers of aws-lc change their licensing or clarify on this matter (which https://github.com/aws/aws-lc/issues/2203 is all about).

AFAIK the better option would be to not use reqwest with the default features, but try to use OpenSSL instead of rustls. I don't know if using `native-tls` is a viable option. After all it should be possible to check this using `cargo deny`.

Or just leave it as it is or set `rustls-no-provider` and any user of `matrix-rust-sdk` has to decide for themselves what kind of rustls backend should be used. E.g. using `rustls-ring` would be an Apache 2.0 based provider for rustls.

poljar:
> that is correct regarding OpenSSL, but reqwest 0.13 is using rustls with aws-lc by default and aws-lc is based on a specific version of BoringSSL, which was based on OpenSSL < 3.0.

That's information that is easy to miss, thanks for raising this. In that case we can't allow this license without becoming GPL-incompatible.

We'll have to pick a different default TLS backend.

crates/matrix-sdk/src/http_client/native.rs
mgoldenberg: Which log? This comment doesn't seem to be attached to a line of code...

Cargo.lock
mgoldenberg: Hm... I don't think this should be the case, I will investigate...

mgoldenberg: This seems to be because `sentry` uses an old version of `reqwest`

```bash
> cargo tree --all-features --workspace --target all --invert reqwest@0.12.24
reqwest v0.12.24
└── sentry v0.46.0
    └── matrix-sdk-ffi v0.16.0 (/home/mgoldenberg/Repositories/matrix/matrix-rust-sdk/bindings/matrix-sdk-ffi)
```

mgoldenberg: We are relatively up-to-date with our `sentry` dependency, but even the [latest version uses `reqwest@0.12.25`](https://docs.rs/crate/sentry/latest/source/Cargo.toml.orig#77). So, I don't think we could reduce the number of `reqwest` versions, unless `sentry` updates.

I could make an issue about this in the `sentry` repository, if you'd like - just let me know!

Cargo.lock
mgoldenberg: It turns out we have five versions of `windows-sys` in this branch.

```bash
git:(support-tls-1.3)> cargo tree --all-features --workspace --target all --invert windows-sys
error: There are multiple `windows-sys` packages in your project, and the specification `windows-sys` is ambiguous.
Please re-run this command with one of the following specifications:
  windows-sys@0.45.0
  windows-sys@0.48.0
  windows-sys@0.52.0
  windows-sys@0.59.0
  windows-sys@0.61.2
```

However, this is only one more than `main`.

```bash
git:(main)> cargo tree --all-features --workspace --target all --invert windows-sys
error: There are multiple `windows-sys` packages in your project, and the specification `windows-sys` is ambiguous.
Please re-run this command with one of the following specifications:
  windows-sys@0.48.0
  windows-sys@0.52.0
  windows-sys@0.59.0
  windows-sys@0.61.2
```

I think this number increased because of having two versions of `reqwest`, but as mentioned above, I'm not sure we can do anything about that.

mgoldenberg: Interestingly, adding a newer version of `reqwest` adds an older version of `windows-sys` - i.e., `windows-sys@0.45.0`. It turns out this is because `rust-platform-verifier` is using an old version of `jni`.

```bash
git:(support-tls-1.3)> cargo tree --all-features --workspace --target all --invert windows-sys@0.45.0
windows-sys v0.45.0
└── jni v0.21.1
    └── rustls-platform-verifier v0.6.2
        └── reqwest v0.13.1
```

We don't have control over this either, but I could make an issue in `rustls-platform-verifier` to update to `jni@0.22.1` which would bring `windows-sys` to `0.61`, reducing the number of versions we have.

Cargo.lock
mgoldenberg: Similar to above. Additional `windows-targets` version added, but it is older than existing ones because of `rustls-platform-verifier`.

`main` has two versions.

```bash
git:(main)> cargo tree --all-features --workspace --target all --invert windows-targets
error: There are multiple `windows-targets` packages in your project, and the specification `windows-targets` is ambiguous.
Please re-run this command with one of the following specifications:
  windows-targets@0.48.5
  windows-targets@0.52.6
```

This branch has three versions.

```bash
git:(support-tls-1.3)> cargo tree --all-features --workspace --target all --invert windows-targets
error: There are multiple `windows-targets` packages in your project, and the specification `windows-targets` is ambiguous.
Please re-run this command with one of the following specifications:
  windows-targets@0.42.2
  windows-targets@0.48.5
  windows-targets@0.52.6
```

The extra version is because of old versions of `jni` and, consequently, `windows-sys` in `rustls-platform-verifier`.

```
git:(support-tls-1.3)> cargo tree --all-features --workspace --target all --invert windows-targets@0.42.2
windows-targets v0.42.2
└── windows-sys v0.45.0
    └── jni v0.21.1
        └── rustls-platform-verifier v0.6.2
            └── reqwest v0.13.1
```

mgoldenberg:
The Kotlin build failure in CI is due to aws/aws-lc-rs#1006. Fixes have been released, and also some workarounds are mentioned. But, it seems the licensing issue might mean we won't depend on `aws-lc-rs`, so I will wait for that conversation to progress before pursuing a fix here.

crates/matrix-sdk/src/http_client/native.rs
Hywan: What the… how Github did that?

Here https://github.com/matrix-org/matrix-rust-sdk/pull/6053/changes#diff-f832b9c871b7a2cc3ace691506c7bc796292359685d508ff64e27d1182ebfa96R194

Cargo.lock
Hywan: Yes, I think it's important here. It's duplicating a lot of dependencies, not super great :-/.

Cargo.lock
Hywan: Yes please. Let's start with that.

crates/matrix-sdk/src/http_client/native.rs
mgoldenberg: Ah, I see now. 

So, the reason I left it up here is because I would have to add the conditional and the log to both branches below. This is because if the `Vec` is not empty, additional certs will always be added below, it's just that they will either replace or be merged in with the built-in root certificates.

Does that make sense? 

Happy to duplicate the condition and log, if you prefer. Or remove it altogether. I'm not feeling strongly about either option, so will defer to you.

Cargo.lock
mgoldenberg: https://github.com/getsentry/sentry-rust/issues/1000

Cargo.lock
mgoldenberg: https://github.com/rustls/rustls-platform-verifier/issues/222

mgoldenberg: Looks like this is already underway, actually!

https://github.com/rustls/rustls-platform-verifier/pull/193

Cargo.lock
mgoldenberg: Turns out this is already underway!

https://github.com/getsentry/sentry-rust/pull/998

mgoldenberg: The pull request in `sentry` got merged, so `master` now uses `reqwest@0.13.1`. When it gets released, I will update the dependency!

Hywan: Great!

mgoldenberg: I have created an issue for this, so that we can resolve this conversation (see #6237).

Cargo.lock
mgoldenberg: I have created an issue for this, so that we can resolve this conversation (see #6238).

Cargo.lock
mgoldenberg: I have created an issue for this, so that we can resolve this conversation (see #6238).

crates/matrix-sdk/src/http_client/native.rs
Hywan: I guess it's fine to remove it.

mgoldenberg: commit: [`5e75b4f172400ad445399791a55b9c08a865b221`](https://github.com/matrix-org/matrix-rust-sdk/pull/6053/commits/5e75b4f172400ad445399791a55b9c08a865b221)

mgoldenberg:
> There we go. Can you rebase your fixup commits please?

Rebasing the entire branch is a bit of a pain because of all the merges that have happened to keep this branch up to date, so I just reworded the fixup commit and rebased from there. If the commit history is feeling too messy, I can create a new branch with a cleaner history - just let me know!

Issue #2778, opened 2023-10-25
State events that fail to deserialize should not be ignored


zecakeh:
This is a problem similar to #607: state events that fail to deserialize are ignored.

But usually in the spec a state event that has invalid fields in the content still replaces the previous state event. It is just treated as if there was no state event of that type to begin with.

I believe that's due to servers not having to validate the content of state events outside of state res, so they always treat any accepted state event as the current state.

Examples from the spec:

- For [`m.space.child`](https://spec.matrix.org/v1.8/client-server-api/#mspacechild-relationship):
    > Child rooms can be removed from a space by omitting the `via` key of `content` on the relevant state event, such as through redaction or otherwise clearing the content.
    
    Although the`via` field is now marked as required: https://github.com/matrix-org/matrix-spec/pull/1618.
- For [`m.room.name`](https://spec.matrix.org/v1.8/client-server-api/#mroomname):
    > A room with an `m.room.name` event with an absent, null, or empty name field should be treated the same as a room with no `m.room.name` event.

PR #6055, opened 2026-01-21
fix(base): Handle sync state events that fail to deserialize

zecakeh:
If the homeserver provides a state event to a client, it means that it considers the event to be valid. An event considered valid by the homeserver might fail to deserialize because a homeserver MUST accept any event from another server that passes the authorization rules, which check only a handful of fields and not the full content expected in the spec.

If a state event is valid, it always updates the state map of the room. So ignoring events that fail to deserialize in a sync response means that the local state map is different than the one from the server.

In some cases the Matrix spec even explicitly says that if a required field from the content of a state event is missing or invalid, it should be treated as if the event is missing from the state map. And if a field is invalid, or a required field is missing, the event will fail to deserialize.

So this handles state events very closely to how a server would: we only deserialize the `type` and `state_key` first to make sure that a valid state event always updates the local state map. Then we only deserialize the events lazily when we encounter an event type that updates the `RoomInfo`. Because we deserialize the event lazily and some methods might edit parts of an event before passing it to `RoomInfo`, the (possibly edited) deserialized event is cached alongside the raw event and its keys to be able to pass it further down the chain.

Part of #2778. We need to do the same thing for stripped state (although it should be less of a problem in practice because it rarely needs to be updated), and there might be other places where only state events that deserialize successfully are handled.

### DIFF
--- a/crates/matrix-sdk-base/src/response_processors/state_events.rs
+++ b/crates/matrix-sdk-base/src/response_processors/state_events.rs
@@ -465,7 +465,7 @@ async fn decrypt_state_event(
             return None;
         }
         Err(error) => {
-            warn!("Couldn't deserialize encrypted state event's ID: {error}");
+            warn!(?error, "Couldn't deserialize encrypted state event's ID");
             return None;
         }
     };
--- a/crates/matrix-sdk-base/src/utils.rs
+++ b/crates/matrix-sdk-base/src/utils.rs
@@ -344,7 +344,7 @@ impl RawSyncStateEventWithKeys {
             match raw.deserialize_as_unchecked() {
                 Ok(fields) => fields,
                 Err(error) => {
-                    warn!("Couldn't deserialize type and state key of state event: {error}");
+                    warn!(?error, "Couldn't deserialize type and state key of state event");
                     return None;
                 }
             };
@@ -372,9 +372,7 @@ impl RawSyncStateEventWithKeys {
         {
             Ok(fields) => fields,
             Err(error) => {
-                warn!(
-                    "Couldn't deserialize type and optional state key of timeline event: {error}"
-                );
+                warn!(?error, "Couldn't deserialize type and optional state key of timeline event");
                 return None;
             }
         };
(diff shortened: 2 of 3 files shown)


crates/matrix-sdk-base/src/response_processors/state_events.rs
```diff
@@ -438,23 +447,39 @@ pub fn is_tombstone_event_valid(
...
     use tracing::{trace, warn};
 
+    let event_id = match raw_event.raw.get_field::<OwnedEventId>("event_id") {
+        Ok(Some(event_id)) => event_id,
+        Ok(None) => {
+            warn!("Couldn't deserialize encrypted state event's ID: missing `event_id` field");
+            return None;
+        }
+        Err(error) => {
+            warn!("Couldn't deserialize encrypted state event's ID: {error}");
```
Hywan: I think it's more idiomatic to write `warn!(?error, "Couldn't …");`.

crates/matrix-sdk-base/src/utils.rs
```diff
@@ -313,3 +316,138 @@ impl From<&StrippedRoomTombstoneEvent> for MinimalStateEvent<RoomTombstoneEventC
...
+
+        let event = as_variant_fn(any_event);
+
+        if event.is_none() {
+            // This should be a developer error, or an upstream error.
+            error!(
+                "Couldn't deserialize `{}` state event: got `{}` event type",
+                C::TYPE,
+                any_event.event_type()
+            );
```
Hywan, line 426: Could we use log field instead, something like `error!(expected_type = C::TYPE, actual_type = any_event.event_type(), "Couldn't deserialize state event: unexpected type");` for example.

crates/matrix-sdk-base/src/response_processors/state_events.rs
zecakeh: Done.

crates/matrix-sdk-base/src/utils.rs
zecakeh: Done

Hywan:
Lovely, thanks!

PR #6056, opened 2026-01-22
feat(sdk): Add invite support in `LatestEventValue`

Hywan:
This patch adds `LatestEventValue::RemoteInvite`.

Basically, when the user receives an invite for a room, the `LatestEventValue` couldn't be computed because:

1. the Event Cache doesn't handle invites (because invite rooms are composed of stripped state-events, whilst the Event Cache only manipulates non stripped (state) events),
2. the Send Queue cannot send events in an invite room.

**The `LatestEvents` API needs to listen to a new channel: `RoomInfo` updates, aka `RoomInfoNotableUpdates`**.

The idea is to generate a new `LatestEventValue::RemoteInvite` variant. It is composed of:

- an optional event ID,
- a timestamp, and
- an inviter (user) ID.

Stripped state events don't contain an `origin_server_ts` where the timestamp is derived from. As [the spec](https://spec.matrix.org/latest/client-server-api/#stripped-state) says:

> Stripped state events can only have the `sender`, `type`, `state_key` and `content` properties present.

So by default, we use `now()` for the timestamp. Once a `LatestEventValue` is computed for an invite, it won't be re-computed except if the cache is cleared, or if the user logs out and logs in again. Concretely, it means that, in a cache clear or fresh logs in, the user will get all the invites at the top of the room list. There is no better alternative for the moment. If a user wants to ignore an invite, we need to wait on the Matrix spec to improve.

The `RoomInfoNotableUpdateReasons::MEMBERSHIP` reason is triggered for invites and knocks. The `LatestEvents` task listen to `RoomInfo` updates and acts accordingly.

It's best to review this contribution one patch at a time for clarity.

- [x] I'll update the `CHANGELOG.md`s once the review is done, to avoid conflicts.

---

* Fix https://github.com/matrix-org/matrix-rust-sdk/issues/5726

### DIFF
--- a/crates/matrix-sdk/CHANGELOG.md
+++ b/crates/matrix-sdk/CHANGELOG.md
@@ -8,6 +8,8 @@ All notable changes to this project will be documented in this file.
 
 ### Features
 
+- `LatestEventValue::RemoteInvite` is added to handle a Latest Event for invite room.
+  ([#6056](https://github.com/matrix-org/matrix-rust-sdk/pull/6056))
 - Add `Room::set_own_member_display_name` to set the current user's display name
   within only the one single room (can be used for /myroomnick functionality).
   [#5981](https://github.com/matrix-org/matrix-rust-sdk/pull/5981)


crates/matrix-sdk/src/event_cache/mod.rs
```diff
@@ -985,7 +985,10 @@ impl EventCacheInner {
 
         // Left rooms.
         for (room_id, left_room_update) in updates.left {
-            let room = self.for_room(&room_id).await?;
+            let Ok(room) = self.for_room(&room_id).await else {
+                error!(?room_id, "Room must exist");
+                continue;
+            };
```
poljar, line 991: Is this fixing an issue or bug?

Why was this added?

crates/matrix-sdk-base/src/response_processors/state_events.rs
```diff
@@ -332,6 +372,35 @@ pub mod stripped {
...
+        state_events: &[AnyStrippedStateEvent],
+    ) {
+        // Start from the last event; the first membership event we see in that order is
+        // the last in the regular order, so that's the only one we need to
+        // consider.
+        for event in state_events.iter().rev() {
+            if let AnyStrippedStateEvent::RoomMember(member) = &event {
+                // If this event updates the current user's membership, record that in the
+                // room_info.
+                if member.state_key.as_str() == user_id.as_str() {
```
poljar: Isn't this a `state_events.iter().find()` call? Might be nicer to express this using a `find()`.

poljar:
Oh, another thing, you might want to consider advertising this feature in the changelog.

crates/matrix-sdk/src/event_cache/mod.rs
Hywan: It's a potential bug, yes.

poljar: Alright, but which bug? How does it manifest. Do we have a test for this bug?

crates/matrix-sdk-base/src/response_processors/state_events.rs
Hywan: Done.

crates/matrix-sdk/src/event_cache/mod.rs
Hywan: I'm not sure it's possible to trigger this bug with the public API. I'll check how to make a relevant test case in a follow up PR.

- Imagine the `handle_room_updates` method receives a `RoomUpdates` about a left room
- If this room isn't known by the SDK, all the other next `RoomUpdates` will be skipped

This is annoying. We don't want to miss updates because of an error.

Technically, the rooms are always created by the sync `response_processors`, this is why it's impossible to trigger this error with the public API.

Hywan:
> Oh, another thing, you might want to consider advertising this feature in the changelog.

Yes, it was part of the todo list of this PR. Thanks for the reminder anyway.

PR #6062, opened 2026-01-26
refactor(crypto-ffi): Serialize the secrets bundle when exporting it

poljar:
This is necessary because by default the secrets bundle can't be serialized over the FFI without more uniffi exports and functions.

Since the apps just want to pass the secrets bundle around, we can just serialize the bundle for them.


andybalaam:
Does this need a changelog entry?

poljar:
> Does this need a changelog entry?

The crypto-ffi bindings are a bit on their way out and don't have a changelog at all. :see_no_evil:

PR #6065, opened 2026-01-26
feat(sqlite): allow storing the same event in multiple linked chunks of the same room

bnjbvr:
Previously, there was a `primary key` constraint on the `event_id` used in `event_chunks`, which made it impossible for the same event to be stored in multiple linked chunks. This would be an issue, as soon as we tried to store a thread event in both the unthreaded room linked chunk, and as part of a thread linked chunk. In addition to this, we may need to store duplicates in the context of #5954: an event could be present in the unthreaded timeline *and* one pinned events linked chunk.

So, we get rid of the primary key on `event_id`, and we add an index there instead, to not hurt performance.

But we do still need a primary key, because the table was created with `WITHOUT ROWID`. I've promoted the other existing index into a primary key, as it makes sense to do so: the tuple `(linked_chunk_id, chunk_id, position)` uniquely identifies an item's position in a linked chunk. Also, the new index is a nice speedup for the use case of loading only a linked chunk's metadata.

A test has been added, that would fail on main and succeed with this branch.

Part of #5954.

### DIFF
--- a/crates/matrix-sdk-sqlite/src/event_cache_store.rs
+++ b/crates/matrix-sdk-sqlite/src/event_cache_store.rs
@@ -2531,6 +2531,84 @@ mod tests {
             });
         }
     }
+
+    #[async_test]
+    async fn test_event_chunks_allows_same_event_in_room_and_thread() {
+        // This test verifies that the same event can appear in both a room's linked
+        // chunk and a thread's linked chunk. This is the real-world use case:
+        // a thread reply appears in both the main room timeline and the thread.
+
+        let store = get_event_cache_store().await.expect("creating cache store failed");
+
+        let room_id = *DEFAULT_TEST_ROOM_ID;
+        let thread_root = event_id!("$thread_root");
+
+        // Create an event that will be inserted into both the room and thread linked
+        // chunks.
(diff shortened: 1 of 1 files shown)


crates/matrix-sdk-sqlite/migrations/event_cache_store/014_event_chunks_event_id_index.sql
```diff
@@ -0,0 +1,34 @@
...
+    -- improves the performance. Also, we don't have a ROWID, so we must use a PRIMARY KEY, hence
+    -- we use this composite key as the primary key.
+    PRIMARY KEY (linked_chunk_id, chunk_id, position),
+
+    -- If the owning chunk gets deleted, delete the entry too.
+    FOREIGN KEY (linked_chunk_id, chunk_id) REFERENCES linked_chunks(linked_chunk_id, id) ON DELETE CASCADE
+)
+WITHOUT ROWID;
+
+-- Create a non-unique index on event_id for query performance.
```
Hywan:
```suggestion
-- Create a non-unique index on `event_id` for query performance.
```

PR #6066, opened 2026-01-26
test: make the `retry_limit_http_requests` test more robust

bnjbvr:
See the commits' descriptions. No changes in functionality.

### DIFF
--- a/crates/matrix-sdk/src/client/mod.rs
+++ b/crates/matrix-sdk/src/client/mod.rs
@@ -3542,15 +3542,15 @@ pub(crate) mod tests {
         let server = MatrixMockServer::new().await;
         let client = server
             .client_builder()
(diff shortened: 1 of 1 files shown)


crates/matrix-sdk/src/http_client/native.rs
```diff
@@ -111,17 +111,18 @@ impl HttpClient {
 
         send_request
             .retry(backoff)
-            .adjust(|err, default_timeout| {
+            .adjust(|err, suggested_timeout| {
```
poljar: Hmm, perhaps we should be explicit and call this `backon` timeout?

Suggested timeout makes me think of the timeout suggested by the homeserver which we call `retry_after`.

bnjbvr: Sure, even called it `backon_suggested_timeout` :+1:

PR #6068, opened 2026-01-27
fix(sdk): Fix latest event erasing `RoomInfo`

jmartinesp:
## Changes:

- Move sync lock locking in `LatestEvent::store` up a bit to avoid a race condition where we copied the room info there, while in parallel modified and saved the room info during the sync, then we modified the now outdated room info and saved it, overwriting the previously saved one and losing data, such as the room's display name.
- Make `m.room.avatar` part of the required state for the notification's sliding sync.
- Simplify a convoluted and unnecessary check in the event handler inside `NotificationClient::try_sliding_sync`.

Thanks @bnjbvr and @Hywan for your help on debugging and fixing the issue!

Signed-off-by:

### DIFF
--- a/crates/matrix-sdk-ui/src/notification_client.rs
+++ b/crates/matrix-sdk-ui/src/notification_client.rs
@@ -393,24 +393,21 @@ impl NotificationClient {
             move |raw: Raw<AnySyncTimelineEvent>| async move {
                 match &raw.get_field::<OwnedEventId>("event_id") {
                     Ok(Some(event_id)) => {
-                        let request =
-                            &requests.iter().find(|request| request.event_ids.contains(event_id));
-                        if request.is_none() {
+                        let Some(request) =
+                            &requests.iter().find(|request| request.event_ids.contains(event_id))
+                        else {
                             return;
-                        }
-                        let room_id = request.unwrap().room_id.clone();
-                        for request in requests.iter() {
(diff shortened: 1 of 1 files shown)


crates/matrix-sdk-ui/src/notification_client.rs
```diff
@@ -393,24 +393,21 @@ impl NotificationClient {
             move |raw: Raw<AnySyncTimelineEvent>| async move {
                 match &raw.get_field::<OwnedEventId>("event_id") {
                     Ok(Some(event_id)) => {
-                        let request =
-                            &requests.iter().find(|request| request.event_ids.contains(event_id));
-                        if request.is_none() {
+                        let Some(request) =
+                            &requests.iter().find(|request| request.event_ids.contains(event_id))
+                        else {
                             return;
```
stefanceriu, line 399: 👌

crates/matrix-sdk/src/latest_events/latest_event/mod.rs
stefanceriu, line 1: Huh, this is so easy to mess up, nice find!

jmartinesp: It was actually @Hywan who found what the root cause was. I just found *where* it was happening 😅 .

PR #6070, opened 2026-01-27
feat(timeline): remove the obsolete event type filter

Johennes:
This is a follow-up on https://github.com/matrix-org/matrix-rust-sdk/pull/6048#issuecomment-3799862659 which removes the event type filter that has been superseded in #6048 by a new event filter


bindings/matrix-sdk-ffi/src/timeline/configuration.rs
```diff
@@ -62,12 +31,30 @@ impl TimelineEventFilter {
...
     }
+
+    #[uniffi::constructor]
+    pub fn exclude_event_types(event_types: Vec<FilterTimelineEventType>) -> Arc<Self> {
+        let conditions = event_types
+            .iter()
+            .map(|t| TimelineEventCondition::EventType(t.clone().into()))
+            .collect();
+        Arc::new(Self { inner: InnerTimelineEventFilter::Exclude(conditions) })
+    }
```
Johennes, line 57: I've added these convenience constructors so that current FFI users of the old event type filter don't have to wrap their event types in another enum layer.

bnjbvr: That's a nice touch. Can you mention them in the CHANGELOG, as an alternative to the existing `TimelineEventTypeFilter`?

Johennes: Good idea. Have added them in the changelog.

Issue #6058, opened 2026-01-23
Improve logging around sending and receiving key gossip requests


andybalaam:
Several recent bug reports have suffered from a lack of information about the sending and receiving of `m.secret.send` requests. We would like to add:

* A request ID when we send an `m.secret.send` so we can see whether requests are being retried or additional new ones being made.
* A link between request ID and secret name in the logging so we can track which request is for which secret
* Logging of the `OlmMachine.cross_signing_status` after the call to `GossipMachine.receive_secret_event` so we can confirm the state of the Olm machine immediately after we have imported a secret.

andybalaam:
It turns out we have a request ID/transaction ID on outgoing gossip requests. in `matrix_sdk_crypto::gossiping::machine::GossipMachine::share_secret` we do:

```
        let request = ToDeviceRequest::new(
            device.user_id(),
            device.device_id().to_owned(),
            &encrypted_event_type,
            content.cast(),
        );

        let request = OutgoingRequest {
            request_id: request.txn_id.clone(),
            request: Arc::new(request.into()),
        };
        self.inner.outgoing_requests.write().insert(request.request_id.clone(), request);
```

so we can see that `request.request_id` is a `TransactionId` that we could log.

We don't store this request into the crypto store as far as I can tell, but I'm not sure we need to here.

PR #6074, opened 2026-01-27
Log more information about outgoing `m.secret.send` messages

andybalaam:
Log more information about gossip requests, so we can track which `m.secret.send` messages were successfully sent or retried, and which secrets were contained in them.

Part of #6058

This makes the debug logging more verbose, but since these secret exchanges only happen in response to a UI action, and I'm only adding one or two more messages for each secret that is sent, and there are only ~4 secrets to send, I think that's OK.

After this change, after a verification exchange we see logs like this in the existing device (which sends secrets):

```
14:53:11.379 DEBUG matrix_sdk_crypto::gossiping::machine: Creating outgoing `m.secret.send` to-device request
    recipient="@andyblocal:localhost:8008" event_type="m.secret.send" request_id="0525e7bb0f1e406395b913de386a47c0" secret_name="m.cross_signing.self_signing" message_id="9ad21c513c9f4f6db3a5495f7ccfe1c4"
    at /home/andy/code/public/matrix-rust/matrix-rust-sdk/crates/matrix-sdk-crypto/src/gossiping/machine.rs:566
    in matrix_sdk_crypto::machine::receive_sync_changes rageshake.ts:69:27

14:53:45.442 DEBUG matrix_sdk_crypto::gossiping::machine: Marking outgoing request as sent
    request_id="0525e7bb0f1e406395b913de386a47c0"
    at /home/andy/code/public/matrix-rust/matrix-rust-sdk/crates/matrix-sdk-crypto/src/gossiping/machine.rs:848 rageshake.ts:69:27
```

Importantly, the `secret_name` can be matched up with a `request_id`, and then later we can see that the same `request_id` was successfully sent, and the `message_id` is included with the event, so can be matched up when the receiver gets it.

On the new device (which receives secrets):

```
12:59:45.135 DEBUG matrix_sdk_crypto::gossiping::machine: Marking outgoing secret request as sent
    recipient="@andyblocal:localhost:8008" request_type="m.cross_signing.user_signing" request_id="c5be84ead4e64e5d966959b56118bab5"
    at /home/andy/code/public/matrix-rust/matrix-rust-sdk/crates/matrix-sdk-crypto/src/gossiping/machine.rs:838 rageshake.ts:69:27
```

we can see that a request for the secret was sent, and then we see it being received:

```
12:59:45.370 DEBUG matrix_sdk_crypto::gossiping::machine: Received a m.secret.send event with a matching request
    at /home/andy/code/public/matrix-rust/matrix-rust-sdk/crates/matrix-sdk-crypto/src/gossiping/machine.rs:923
    in matrix_sdk_crypto::gossiping::machine::receive_secret_event with sender="@andyblocal:localhost:8008" request_id="c5be84ead4e64e5d966959b56118bab5" secret_name="m.cross_signing.user_signing"
    in matrix_sdk_crypto::machine::handle_decrypted_to_device_event with sender_key="curve25519:iH/T0d+dHVA4PmQbcI+GonWxEOE9UN/6IlUYH62urXY" event_type="m.secret.send"
    in matrix_sdk_crypto::machine::receive_to_device_event with sender="@andyblocal:localhost:8008" event_type="m.room.encrypted" message_id="f5401aa69464445d92789c7a6c039ece"
    in matrix_sdk_crypto::machine::receive_sync_changes rageshake.ts:69:27


12:59:45.373 DEBUG matrix_sdk_crypto::gossiping::machine: Successfully received a secret, removing the request
    recipient="@andyblocal:localhost:8008" request_type="m.cross_signing.user_signing" request_id="c5be84ead4e64e5d966959b56118bab5"
    at /home/andy/code/public/matrix-rust/matrix-rust-sdk/crates/matrix-sdk-crypto/src/gossiping/machine.rs:864
    in matrix_sdk_crypto::gossiping::machine::receive_secret_event with sender="@andyblocal:localhost:8008" request_id="c5be84ead4e64e5d966959b56118bab5" secret_name="m.cross_signing.user_signing"
    in matrix_sdk_crypto::machine::handle_decrypted_to_device_event with sender_key="curve25519:iH/T0d+dHVA4PmQbcI+GonWxEOE9UN/6IlUYH62urXY" event_type="m.secret.send"
    in matrix_sdk_crypto::machine::receive_to_device_event with sender="@andyblocal:localhost:8008" event_type="m.room.encrypted" message_id="f5401aa69464445d92789c7a6c039ece"
    in matrix_sdk_crypto::machine::receive_sync_changes rageshake.ts:69:27
```

### DIFF
--- a/crates/matrix-sdk-crypto/CHANGELOG.md
+++ b/crates/matrix-sdk-crypto/CHANGELOG.md
@@ -8,6 +8,8 @@ All notable changes to this project will be documented in this file.
 
 ### Features
 
+- Improved logging when we are sending secrets in `GossipMachine`.
+  ([#6074](https://github.com/matrix-org/matrix-rust-sdk/pull/6074))
 - Added a new field `forwarder` to `InboundGroupSession` of type `ForwarderData`, which stores information about the forwarder of a session shared in a room key bundle under [MSC4268](https://github.com/matrix-org/matrix-spec-proposals/pull/4268).
   ([#5980])(https://github.com/matrix-org/matrix-rust-sdk/pull/5980)
 - The `OutboundGroupSession` and `OlmMachine` now return the `EncryptionInfo` 
@@ -19,7 +21,6 @@ All notable changes to this project will be documented in this file.
   key bundle.
   ([#6017](https://github.com/matrix-org/matrix-rust-sdk/pull/6017))
   ([#6044](https://github.com/matrix-org/matrix-rust-sdk/pull/6044))
-
 
 ### Refactor
 
--- a/crates/matrix-sdk-crypto/src/gossiping/machine.rs
+++ b/crates/matrix-sdk-crypto/src/gossiping/machine.rs
@@ -323,7 +323,7 @@ impl GossipMachine {
                         "Sharing a secret with a device",
                     );
 
-                    match self.share_secret(&device, content).await {
+                    match self.share_secret(&device, content, secret_name).await {
                         Ok(s) => Ok(Some(s)),
                         Err(OlmError::MissingSession) => {
                             info!(
@@ -534,13 +534,22 @@ impl GossipMachine {
         }
     }
 
+    /// Add an `m.secret.send` request to the `outgoing_requests` queue.
+    ///
+    /// # Arguments
+    ///
+    /// * `device` is the device to send the request to.
+    /// * `content` is the actual event content, containing the secret to send.
+    /// * `secret_name` is the name of the secret e.g. `m.megolm_backup.v1`
+    ///   (used for logging).
     async fn share_secret(
         &self,
         device: &Device,
         content: SecretSendContent,
+        secret_name: &SecretName,
     ) -> OlmResult<Session> {
         let event_type = content.event_type().to_owned();
-        let (used_session, content) = device.encrypt(&event_type, content).await?;
+        let (used_session, content, message_id) = device.encrypt(&event_type, content).await?;
 
         let encrypted_event_type = content.event_type().to_owned();
 
@@ -555,6 +564,16 @@ impl GossipMachine {
             request_id: request.txn_id.clone(),
             request: Arc::new(request.into()),
         };
+
+        debug!(
+            recipient = ?device.user_id(),
+            event_type,
+            request_id = ?request.request_id,
+            ?secret_name,
+            ?message_id,
+            "Creating outgoing `m.secret.send` to-device request"
+        );
+
         self.inner.outgoing_requests.write().insert(request.request_id.clone(), request);
 
         Ok(used_session)
@@ -583,6 +602,15 @@ impl GossipMachine {
             request_id: request.txn_id.clone(),
             request: Arc::new(request.into()),
         };
+
+        debug!(
+            recipient = ?device.user_id(),
+            event_type = event_type,
+            request_id = ?request.request_id,
+            session_id = session.session_id(),
+            "Creating outgoing `m.forwarded_room_key` to-device request"
+        );
+
         self.inner.outgoing_requests.write().insert(request.request_id.clone(), request);
 
         Ok(used_session)
@@ -816,7 +844,7 @@ impl GossipMachine {
         let info = self.inner.store.get_outgoing_secret_requests(id).await?;
 
         if let Some(mut info) = info {
-            trace!(
+            debug!(
                 recipient = ?info.request_recipient,
                 request_type = info.request_type(),
                 request_id = ?info.request_id,
@@ -824,6 +852,14 @@ impl GossipMachine {
             );
             info.sent_out = true;
             self.save_outgoing_key_info(info).await?;
+        } else if let Some(req) = self.inner.outgoing_requests.read().get(id) {
+            // This outgoing event was not saved into the store. This is
+            // expected: for example `m.secret.send` events are only stored in
+            // `outgoing_requests`, not in the store.
+            debug!(
+                request_id = ?req.request_id,
+                "Marking outgoing request as sent"
+            );
         }
 
         self.inner.outgoing_requests.write().remove(id);
@@ -835,7 +871,7 @@ impl GossipMachine {
     ///
     /// This will queue up a request cancellation.
     async fn mark_as_done(&self, key_info: &GossipRequest) -> Result<(), CryptoStoreError> {
-        trace!(
+        debug!(
             recipient = ?key_info.request_recipient,
             request_type = key_info.request_type(),
             request_id = ?key_info.request_id,
--- a/crates/matrix-sdk-crypto/src/identities/device.rs
+++ b/crates/matrix-sdk-crypto/src/identities/device.rs
@@ -412,12 +412,23 @@ impl Device {
     ///
     /// # Arguments
     ///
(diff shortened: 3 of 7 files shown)


crates/matrix-sdk-crypto/src/gossiping/machine.rs
```diff
@@ -555,6 +556,15 @@ impl GossipMachine {
             request_id: request.txn_id.clone(),
             request: Arc::new(request.into()),
         };
+
+        debug!(
+            recipient = ?device.user_id(),
+            event_type,
+            request_id = ?request.request_id,
+            ?secret_name,
+            "Creating outgoing secret sharing request"
```
richvdh: As a general note (also applying to the title of the PR), I found this terminology a little confusing. "Gossip request" and "secret sharing request" sound to me like `m.secret.request` messages, but I think this an `m.secret.send`?

```suggestion
            "Creating outgoing `m.secret.send` to-device request"
```

That said, I can see we use the existing terminology elsewhere, and consistency is important.

crates/matrix-sdk-crypto/src/gossiping/machine.rs
```diff
@@ -538,6 +538,7 @@ impl GossipMachine {
         &self,
         device: &Device,
         content: SecretSendContent,
+        secret_name: &SecretName,
     ) -> OlmResult<Session> {
         let event_type = content.event_type().to_owned();
         let (used_session, content) = device.encrypt(&event_type, content).await?;
```
richvdh: `Device::encrypt` will create a unique message_id, which is added to the content, and then logged on the server and the recipient. Can we log it on the sender side too?

(It is slightly different to the txnid on a ToDeviceRequest because (a) it stays attached to the message from one end to the other; (b) there can in theory be multiple to-device messages, each with a different msgid, in a single ToDeviceRequest.)

crates/matrix-sdk-crypto/src/gossiping/machine.rs
```diff
@@ -816,14 +835,19 @@ impl GossipMachine {
...
                 request_id = ?info.request_id,
                 "Marking outgoing secret request as sent"
             );
             info.sent_out = true;
             self.save_outgoing_key_info(info).await?;
+        } else if let Some(req) = self.inner.outgoing_requests.read().get(id) {
+            debug!(
+                request_id = ?req.request_id,
+                "Marking outgoing request as sent"
+            );
```
richvdh, line 862: this might be a bit spammy?

crates/matrix-sdk-crypto/src/gossiping/machine.rs
```diff
@@ -538,6 +538,7 @@ impl GossipMachine {
         &self,
         device: &Device,
         content: SecretSendContent,
+        secret_name: &SecretName,
```
richvdh, line 549: it would be good to add some documentation, noting particularly that `secret_name` is only used for logging.

crates/matrix-sdk-crypto/src/gossiping/machine.rs
```diff
@@ -583,6 +593,15 @@ impl GossipMachine {
             request_id: request.txn_id.clone(),
             request: Arc::new(request.into()),
         };
+
+        debug!(
+            recipient = ?device.user_id(),
+            event_type = event_type,
+            request_id = ?request.request_id,
+            session_id = session.session_id(),
+            "Creating outgoing forwarded room key request"
```
richvdh:
```suggestion
            "Creating outgoing `m.forwarded_room_key` to-device request"
```

crates/matrix-sdk-crypto/src/gossiping/machine.rs
andybalaam: I think your suggestion preserves existing terminology while still making it clearer.

crates/matrix-sdk-crypto/src/gossiping/machine.rs
andybalaam: It only logs when we have successfully sent a gossip request, which I think only happens with a user-triggered action, right?

crates/matrix-sdk-crypto/src/gossiping/machine.rs
andybalaam: Done in 8e44fad2a1

crates/matrix-sdk-crypto/src/gossiping/machine.rs
richvdh: Ah, yes, I missed that this was in `GossipMachine` rather than `OlmMachine`.

Do we ever hit this line?

andybalaam: Yes - some outgoing events are not saved into the store, including `m.secret.send`. It seems like we only save to the store if we're expecting a reply e.g. `m.secret.request`.

(An example of us hitting this line is in the second log line of the first code block above.)

crates/matrix-sdk-crypto/src/gossiping/machine.rs
andybalaam: Done in f1f934a1e6

crates/matrix-sdk-crypto/src/identities/device.rs
```diff
@@ -417,7 +417,7 @@ impl Device {
         &self,
         event_type: &str,
         content: impl Serialize,
-    ) -> OlmResult<(Session, Raw<ToDeviceEncryptedEventContent>)> {
+    ) -> OlmResult<(Session, Raw<ToDeviceEncryptedEventContent>, String)> {
```
richvdh, line 431: should probably document the result type.

crates/matrix-sdk-crypto/CHANGELOG.md
```diff
@@ -19,7 +19,8 @@ All notable changes to this project will be documented in this file.
   key bundle.
   ([#6017](https://github.com/matrix-org/matrix-rust-sdk/pull/6017))
   ([#6044](https://github.com/matrix-org/matrix-rust-sdk/pull/6044))
-
+- Improved logging when we are sending secrets in `GossipMachine`.
+  ([#6074](https://github.com/matrix-org/matrix-rust-sdk/pull/6074))
```
richvdh: this should really be at the top of the list, not the bottom, since the most recent entries are at the top of the file.

crates/matrix-sdk-crypto/src/gossiping/machine.rs
richvdh: might be worth writing a comment to that effect.

richvdh (approved):
a few nits, but lgtm

crates/matrix-sdk-crypto/CHANGELOG.md
andybalaam: Done in 5af4cb3078

crates/matrix-sdk-crypto/src/identities/device.rs
andybalaam: Done in d7ff87bb26

crates/matrix-sdk-crypto/src/gossiping/machine.rs
andybalaam: Done in 42896ae0e1

PR #6080, opened 2026-01-29
feat(xtask): Introduce `xtask log sync` to visualise logs about sync

Hywan:
This patch introduces a new family of commands: `xtask log`. The goal is to manipulate logs, to extract the right amount of data we need to solve specific problems. The first member of this family is `sync` to visualise logs about the sync process. It presents the sync requests and responses in a table, with a "timeline" _à la_ network profiler graph. The code is rather simple, on purpose. The generated HTML reports are lightweight, and fully standalone: no JavaScript, pure HTML and CSS, no external resources. These reports can be shared or archived super easily.

Features:

- requests/responses are grouped by connection ID
- permalink to specific request ID
- status have colours
- time is displayed in a human form
- duration is calculated from the log timestamps
- view syncs in a "tree-like" flavor, a "time graph", super quick to spot long requests
- each line can be "opened" to see details, so far only log line numbers to get more context manually

Thanks to @manuroe for nerd sniping me on this!

Example:

- [report.html](https://github.com/user-attachments/files/24967655/report.html)
- or [as a screenshot](https://github.com/user-attachments/assets/697462c3-81aa-44cb-b04c-5f7865e20bd9):

Screenshot 2026-01-29 at 12-11-32 Network viewer

### DIFF
--- a/xtask/src/log/mod.rs
+++ b/xtask/src/log/mod.rs
@@ -11,9 +11,11 @@ pub struct LogArgs {
     cmd: LogCommand,
 }
 
+/// Analysis around logs (formatted by `matrix-sdk-ffi`).
 #[derive(Subcommand)]
 enum LogCommand {
-    /// Visualise the sync requests and responses.
+    /// Visualise the sync requests and responses with a table and a duration
+    /// graph.
     Sync {
         #[clap(long)]
         log_file: PathBuf,


xtask/src/log/sync.rs
```diff
@@ -0,0 +1,260 @@
...
+      <td>{time}</td>
+      <td>
+        <div class=\"span\" style=\"--start-at: {start_at}; --duration: {duration}\"><span>{duration_label}</span></div>
+        <details>
+          <summary><span class=\"hidden\">information</span></summary>
+          <ul>
+            <li>Request log line number: {request_log_line}</li>
+            <li>Response log line number: {response_log_line}</li>
+          </ul>
+        </dialog>
```
Hywan: Must be a `</details>`.

Hywan: Done.

xtask/src/log/mod.rs
```diff
@@ -0,0 +1,39 @@
...
+
+#[derive(Args)]
+pub struct LogArgs {
+    #[clap(subcommand)]
+    cmd: LogCommand,
+}
+
+#[derive(Subcommand)]
+enum LogCommand {
+    /// Visualise the sync requests and responses.
```
poljar: Can we document this a bit better? It doesn't tell you at all which format the log file should be.

I think you could easily set up tracing a bit differently and have the regex not work.

poljar (approved):

PR #6081, opened 2026-01-29
Refactor the  QRCodeData struct to support multiple data types in preparation of MSC4388 support

poljar:
This PR contains quite a lot of commits but does not contain any functional changes.

The main gist of it is that the `QrCodeData` struct underneath now has an enum with a MSC4108 variant.

In the future we'll add a MSC4388 variant to support both version of the QR code format we're using for QR login.

### DIFF
--- a/crates/matrix-sdk-crypto/src/types/qr_login/mod.rs
+++ b/crates/matrix-sdk-crypto/src/types/qr_login/mod.rs
@@ -121,7 +121,9 @@ impl From<msc_4108::QrCodeIntent> for QrCodeIntent {
 /// The [`QrCodeData`] can be serialized and encoded as a QR code or it can be
 /// decoded from a QR code.
 #[derive(Debug, Clone, PartialEq, Eq)]
-pub struct QrCodeData(QrCodeDataInner);
+pub struct QrCodeData {
+    inner: QrCodeDataInner,
+}
 
 impl QrCodeData {
     /// Create a new [`QrCodeData`] object which conforms to the data format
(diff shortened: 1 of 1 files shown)


crates/matrix-sdk-crypto/src/types/qr_login/mod.rs
```diff
@@ -62,3 +64,97 @@ pub enum LoginQrCodeDecodeError {
...
+    /// Encode the [`QrCodeData`] into a base64 encoded string.
+    pub fn to_base64(&self) -> String {
+        let bytes = self.to_bytes();
+        base64_encode(bytes)
+    }
+
+    /// The ephemeral Curve25519 public key. Can be used to establish a shared
+    /// secret using the Diffie-Hellman key agreement.
+    pub fn public_key(&self) -> Curve25519PublicKey {
+        match &self.0 {
```
stefanceriu: I find these pretty jarring, can't we just name that field something appropriate or is that not rusty enough?

poljar: `inner` is another common idiom for this if you don't like tuple structs.

Otherwise, suggestions welcome.

stefanceriu: `underlying_data` maybe?

poljar: That's a mouthful, let's stick to `inner` then.

crates/matrix-sdk-crypto/src/types/qr_login/msc_4108.rs
```diff
@@ -21,8 +21,10 @@ use byteorder::{BigEndian, ReadBytesExt};
 use url::Url;
 use vodozemac::Curve25519PublicKey;
 
-use super::{LoginQrCodeDecodeError, VERSION};
+use super::LoginQrCodeDecodeError;
 
+/// The version of the QR code data, currently only one version is specified.
+const VERSION: u8 = 0x02;
```
andybalaam, line 27: Should this be called TYPE?

andybalaam: Or even MSC4108_TYPE?

poljar: Well for MSC4108 it was still called a version, though the publicly accessible error variant now calls everything a type.

This isn't publicly accessible anymore.

:shrug:

PR #6084, opened 2026-01-29
chore(spaces): reduce `m.space.child` and `m.space.parent` deserialisation failure log levels as they are expected when removing either from the hierarchy

stefanceriu:
What is says up there ^


bnjbvr (approved):
Thanks! (Even `trace` would be OK IMO in this case, since it'd be so frequent.)

stefanceriu:
> Thanks! (Even `trace` would be OK IMO in this case, since it'd be so frequent.)

Let it be trace then!

PR #6085, opened 2026-01-29
feat(pinned events): persist the pinned events timeline

bnjbvr:
This brings some nice benefits:

- reloading the previous list of pinned events from storage, which gives us…
- …offline access for pinned events
- better support for automatic redecryption of pinned events

Some important design points:

- the pinned events linked chunk *does* include relations to pinned events, i.e. they're appended to the chunks as they show up from other sources (sync/backpaginations). This is different from the thread linked chunks, which don't, and should likely be uniformized towards that, in the future.
- pinned events are first reloaded from storage, upon creation of the pinned event cache (lazily, when subscribing to pinned events for the first time); then, reloaded from network if needs be; then, updated over time, by a background task in the pinned event cache.
- The task is never killed, at the moment. Ideally, it would be killed whenever there are no more listeners to the pinned events cache for this room. We should implement this in a soon future.
- Lots of code duplication with the storage of linked chunks in the case of the room. To be deduplicated soon™.

Part of https://github.com/matrix-org/matrix-rust-sdk/issues/5954.

### DIFF
--- a/crates/matrix-sdk/src/event_cache/room/pinned_events.rs
+++ b/crates/matrix-sdk/src/event_cache/room/pinned_events.rs
@@ -24,7 +24,7 @@ use matrix_sdk_base::{
         store::{EventCacheStoreLock, EventCacheStoreLockGuard, EventCacheStoreLockState},
     },
     linked_chunk::{LinkedChunkId, OwnedLinkedChunkId, Update},
-    serde_helpers::{RelationsType, extract_relation},
+    serde_helpers::extract_relation,
     task_monitor::BackgroundTaskHandle,
 };
 #[cfg(feature = "e2e-encryption")]
@@ -557,10 +557,8 @@ impl PinnedEventCache {
 
         // Don't include thread responses in the pinned event chunk.
         match rel_type {
-            RelationsType::Thread => None,
-            RelationsType::Edit | RelationsType::Annotation | RelationsType::Reference => {
-                Some(event_id)
-            }
+            RelationType::Thread => None,
+            _ => Some(event_id),
         }
     }
 
--- a/crates/matrix-sdk-common/src/serde_helpers.rs
+++ b/crates/matrix-sdk-common/src/serde_helpers.rs
@@ -19,7 +19,7 @@ use ruma::{
     MilliSecondsSinceUnixEpoch, OwnedEventId,
     events::{
         AnyMessageLikeEventContent, AnySyncMessageLikeEvent, AnySyncTimelineEvent,
-        relation::BundledThread,
+        relation::{BundledThread, RelationType},
     },
     serde::Raw,
 };
@@ -27,30 +27,10 @@ use serde::Deserialize;
 
 use crate::deserialized_responses::{ThreadSummary, ThreadSummaryStatus};
 
-/// The type of relation an event has to another one, if any.
-#[derive(Clone, Copy, Debug, Deserialize, PartialEq, Eq)]
-pub enum RelationsType {
-    /// The event is part of a thread, and the related event is the thread root.
-    #[serde(rename = "m.thread")]
-    Thread,
-    /// The event is an edit of another event, and the related event is the one
-    /// being edited.
-    #[serde(rename = "m.replace")]
-    Edit,
-    /// The event is an annotation of (reaction to) another event, and the
-    /// related event is the one being annotated.
-    #[serde(rename = "m.annotation")]
-    Annotation,
-    /// The event is referencing another event, and the related event is the one
-    /// being referenced.
-    #[serde(rename = "m.reference")]
-    Reference,
-}
-
 #[derive(Deserialize)]
 struct RelatesTo {
     #[serde(rename = "rel_type")]
-    rel_type: RelationsType,
+    rel_type: RelationType,
     #[serde(rename = "event_id")]
     event_id: Option<OwnedEventId>,
 }
@@ -74,8 +54,8 @@ pub fn extract_thread_root_from_content(
 ) -> Option<OwnedEventId> {
     let relates_to = content.deserialize_as_unchecked::<SimplifiedContent>().ok()?.relates_to?;
     match relates_to.rel_type {
-        RelationsType::Thread => relates_to.event_id,
-        RelationsType::Edit | RelationsType::Reference | RelationsType::Annotation => None,
+        RelationType::Thread => relates_to.event_id,
+        _ => None,
     }
 }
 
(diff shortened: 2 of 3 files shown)


crates/matrix-sdk/src/event_cache/room/mod.rs
```diff
@@ -235,6 +236,25 @@ impl RoomEventCache {
         Ok(state.subscribe_to_thread(thread_root))
     }
 
+    /// Get the pinned event cache for this room.
+    ///
```
poljar: A copy paste issue or hasn't been updated properly?

crates/matrix-sdk/src/event_cache/room/pinned_events.rs
```diff
@@ -0,0 +1,626 @@
...
+struct PinnedEventCacheState {
+    /// The ID of the room owning this list of pinned events.
+    room_id: OwnedRoomId,
+
+    /// A sender for live events updates in this room's pinned events list.
+    sender: Sender<RoomEventCacheUpdate>,
+
+    /// The linked chunk representing this room's pinned events.
+    ///
+    /// Does not contain related events by default.
```
poljar: I think this isn't quite accurate, we can see someplace above that related events are added.

crates/matrix-sdk/src/event_cache/room/pinned_events.rs
```diff
@@ -0,0 +1,626 @@
...
+    }
+
+    /// Propagate the changes in this linked chunk to observers, and save the
+    /// changes on disk.
+    async fn propagate_changes(&mut self) -> Result<()> {
+        let updates = self.chunk.store_updates().take();
+        self.send_updates_to_store(updates).await
+    }
+
+    // NOTE: copy/paste
```
poljar: Maybe increase the verbosity here a bit, what should be copy pasted where or from where it was copy pasted.

crates/matrix-sdk/src/event_cache/room/pinned_events.rs
```diff
@@ -0,0 +1,626 @@
...
+    }
+}
+
+/// All the information related to a room's pinned events cache.
+pub struct PinnedEventCache {
+    state: Arc<RwLock<PinnedEventCacheState>>,
+
+    /// The task handling the refreshing of pinned events for this specific
+    /// room.
+    // TODO(bnjbvr): use the background job handle for this, when
```
poljar: It's available. DO IT.

crates/matrix-sdk/src/event_cache/room/pinned_events.rs
```diff
@@ -0,0 +1,626 @@
...
+                let event_cache = event_cache.clone();
+                Self::load_event_with_relations(event_id, room, event_cache)
+            }))
+            .buffer_unordered(MAX_CONCURRENT_REQUESTS)
+            // Flatten all the vectors.
+            .flat_map(stream::iter)
+            .collect()
+            .await;
+
+        if loaded_events.is_empty() {
```
poljar, line 812: Let's add a warning here if we couldn't load all the pinned events, i.e. if the number of pinned event IDs is smaller than the number of loaded events.

crates/matrix-sdk-ui/src/timeline/tasks.rs
```diff
@@ -32,6 +32,79 @@ use tracing::{error, instrument, trace, warn};
...
+                    EventsOrigin::Pagination => RemoteEventOrigin::Pagination,
+                    EventsOrigin::Cache => RemoteEventOrigin::Cache,
+                };
+
+                let has_diffs = !diffs.is_empty();
+
+                timeline_controller.handle_remote_events_with_diffs(diffs, origin).await;
+
+                if has_diffs && matches!(origin, RemoteEventOrigin::Cache) {
+                    timeline_controller.retry_event_decryption(None).await;
```
poljar: Is this needed anymore? R2D2 listens to the updates which I think might include things where we load from storage. This should the automatically trigger redecryption attempts for UTDs we persisted in the event cache when we load them.

crates/matrix-sdk-ui/src/timeline/tasks.rs
```diff
@@ -189,8 +262,9 @@ pub(in crate::timeline) async fn room_event_cache_updates_task(
 
                 if matches!(timeline_focus, TimelineFocus::Live { .. }) {
                     timeline_controller.handle_remote_events_with_diffs(diffs, origin).await;
-                } else {
-                    // Only handle the remote aggregation for a non-live timeline.
+                } else if !matches!(timeline_focus, TimelineFocus::PinnedEvents) {
+                    // Only handle the remote aggregation for a non-live timeline, that's not the
+                    // pinned events one (since this one handles its own remote events diffs).
```
poljar: I'm not sure that this is needed for pinned events.

crates/matrix-sdk-ui/tests/integration/timeline/pinned_event.rs
```diff
@@ -476,10 +513,17 @@ async fn test_pinned_timeline_with_pinned_event_ids_and_empty_result_fails() {
...
-    // The timeline couldn't load any events so it fails to initialise
-    assert!(ret.is_err());
+    // While the timeline gets initialized, it will fail to be updated by the
+    // background pinned cache task, after one second.
+    let timeline =
+        TimelineBuilder::new(&room).with_focus(pinned_events_focus(1)).build().await.unwrap();
+
+    let (_, mut timeline_stream) = timeline.subscribe().await;
+
+    let result = timeout(timeline_stream.next(), Duration::from_secs(1)).await;
```
poljar: 1s might be a bit long for something we'll always have to wait on. But I guess it's hard to know when the timeout is due to nothing coming in vs the timeout expired because the CI was slow to run things.

crates/matrix-sdk/src/event_cache/room/pinned_events.rs
```diff
@@ -123,29 +123,18 @@ impl PinnedEventCacheStateLock {
                 drop(state_guard);
                 let state_guard = self.locked_state.write().await;
 
-                let guard =
+                let mut guard =
                     PinnedEventCacheStateLockWriteGuard { state: state_guard, store: store_guard };
 
                 // Force to reload by shrinking to the last chunk.
```
poljar: We're not shrinking the chunk here, no?

crates/matrix-sdk/tests/integration/room/pinned_events.rs
```diff
@@ -120,3 +129,93 @@ async fn test_unpin_event_is_returning_an_error() {
...
+                Update::NewItemsChunk { previous: None, new: ChunkIdentifier::new(0), next: None },
+                Update::PushItems {
+                    at: Position::new(ChunkIdentifier::new(0), 0),
+                    items: vec![pinned_event.clone()],
+                },
+            ],
+        )
+        .await
+        .unwrap();
+
```
poljar: We should probably test the case where we need to fetch from network as well.

poljar (approved):
Alright, I left some comments. Most of them are minor so I'll approve ahead of time.

bnjbvr:
Clearing the other review requests, since Damir has reviewed this. Thanks!

crates/matrix-sdk-ui/src/timeline/tasks.rs
bnjbvr: Indeed :partying_face:

crates/matrix-sdk-ui/src/timeline/tasks.rs
bnjbvr: Hmm, I thought the live update task wouldn't be needed for the pinned events timeline, but that's actually unclear; it might be that the member profiles updates are still useful for such a timeline. Will defer this to a subsequent PR, and fix the comment in the meanwhile.

jmartinesp:
We might want to delay merging this, I found an issue while testing it with EXA where after clearing the cache, some rooms with just 1 pinned event incorrectly displays several of them.

jmartinesp:
The latest changes fix the previously found issue 🥳 !

PR #6088, opened 2026-01-30
refactor(base): Use `PossiblyRedactedStateEventContent` bound in `MinimalStateEvent`

zecakeh:
We usually don't care if the event was redacted or not, we usually want to know whether a field is set or not, so we don't need `Original` and `Redacted` variants.

This simplifies several parts of the code since we don't have to handle the intermediate enum to access the content now. Due to new APIs in Ruma we can also just convert original and redacted event contents to possibly redacted event contents.

This should allow further code removal in a follow-up PR, by allowing to use the same code to handle `Sync*` and `Stripped*` events in `RoomInfo`.

The backwards-compatible deserialization of `MinimalStateEvent` was verified in a client with Fractal.

### DIFF
--- a/crates/matrix-sdk-base/src/utils.rs
+++ b/crates/matrix-sdk-base/src/utils.rs
@@ -1,44 +1,21 @@
 use ruma::{
-    EventId, OwnedEventId, assign,
+    OwnedEventId,
     events::{
-        AnySyncStateEvent, AnySyncTimelineEvent, RedactContent, RedactedStateEventContent,
-        StateEventContent, StateEventType, StaticEventContent, StaticStateEventContent,
-        SyncStateEvent,
+        AnySyncStateEvent, AnySyncTimelineEvent, PossiblyRedactedStateEventContent, RedactContent,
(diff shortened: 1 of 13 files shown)


crates/matrix-sdk-base/src/utils.rs
```diff
@@ -1,236 +1,191 @@
...
-/// For more details see [`MinimalStateEvent`].
-#[derive(Clone, Debug, Deserialize, Serialize)]
-pub struct OriginalMinimalStateEvent<C>
-where
-    C: StateEventContent,
-{
+#[serde(
+    bound(serialize = "C: Serialize + Clone"),
+    from = "MinimalStateEventSerdeHelper<C>",
+    into = "MinimalStateEventSerdeHelper<C>"
```
zecakeh, line 29: Note: I would have preferred to be able to serialize this directly without a tag, but for some reason applying `#[serde(untagged)]` on a variant of the deserialization helper makes the deserialization of the join rules content fail.

poljar (approved):
As far as I understand this looks good.

PR #6089, opened 2026-01-30
feat(crypto): Add support for the QR code data type defined in MSC4388

poljar:
This PR is mostly a copy of the MSC4018 support in the `QrCodeData` type. We keep support for these MSCs separate to ease the cleanup once we're able to deprecate MSC4018 support. 

### DIFF
--- a/bindings/matrix-sdk-ffi/src/qr_code.rs
+++ b/bindings/matrix-sdk-ffi/src/qr_code.rs
@@ -256,6 +256,7 @@ impl QrCodeData {
                     Some(server_name.to_owned())
                 }
             },
+            qr_login::QrCodeIntentData::Msc4388 { .. } => None,
         }
     }
 }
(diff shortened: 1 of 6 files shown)


bindings/matrix-sdk-ffi/src/qr_code.rs
```diff
@@ -328,6 +329,7 @@ impl From<qrcode::QRCodeLoginError> for HumanQrLoginError {
                 | SecureChannelError::MessageDecode(_)
                 | SecureChannelError::Json(_)
                 | SecureChannelError::RendezvousChannel(_) => HumanQrLoginError::Unknown,
+                SecureChannelError::UnsupportedQrCodeType => HumanQrLoginError::Unknown,
```
andybalaam: Any reason this is not combined with the other values `|`'d together above it?

crates/matrix-sdk-crypto/src/types/qr_login/msc_4388.rs
```diff
@@ -0,0 +1,341 @@
...
+            // 4. Let's get the public key and convert it to our strongly typed
+            // Curve25519PublicKey type.
+            let mut public_key = [0u8; Curve25519PublicKey::LENGTH];
+            reader.read_exact(&mut public_key)?;
+            let public_key = Curve25519PublicKey::from_bytes(public_key);
+
+            // 5. We read two bytes for the length of the rendezvous ID.
+            let rendezvous_id_len = reader.read_u16::<BigEndian>()?;
+            // 6. We read the rendezvous ID itself.
+            let mut rendezvous_id = vec![0u8; rendezvous_id_len.into()];
```
andybalaam, line 134: No idea, but should we be zeroizing any of these things e.g. the `Vec<u8>` form of `rendezvous_id`?

poljar: They aren't really secrets, so if we have some additional copies of this in memory it doesn't really matter.

bindings/matrix-sdk-ffi/src/qr_code.rs
poljar: Err, did this in a hurry and forgot to put them together. Will change it.

PR #6091, opened 2026-01-30
fix(sqlite): Replace `unwrap` when using `interact`

Hywan:
This patch replaces the `interact(…).unwrap()` by a proper error.

So far, `interact()` was only returning `InteractError::Panic` despites `InteractError::Aborted` exists. With https://github.com/deadpool-rs/deadpool/pull/461, we now get `InteractError::Aborted` when the SDK is shutdown, sometimes. This results in hitting the `unwrap` and having a panic again. This patch solves the problem by changing the `unwrap` to a proper error. Note: in case of `InteractError::Panic`, we continue to panic.

This patch makes sense with or without the merge of the PR on `deadpool`.

### DIFF
--- a/crates/matrix-sdk-sqlite/CHANGELOG.md
+++ b/crates/matrix-sdk-sqlite/CHANGELOG.md
@@ -13,6 +13,11 @@ All notable changes to this project will be documented in this file.
   ([#6017](https://github.com/matrix-org/matrix-rust-sdk/pull/6017))
   ([#6044](https://github.com/matrix-org/matrix-rust-sdk/pull/6044))
 
+### Bug Fixes
+
+- Fix a panic when the SQLite connection is aborted.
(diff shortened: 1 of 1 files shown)


Hywan:
Removed the use of our own fork.

crates/matrix-sdk/CHANGELOG.md
```diff
@@ -53,6 +53,8 @@ All notable changes to this project will be documented in this file.
 - Allow granting of QR login to a new client whose device ID is not a base64
   encoded Curve25519 public key.
   ([#5940](https://github.com/matrix-org/matrix-rust-sdk/pull/5940))
+- Fix a panic when the SQLite connection is aborted.
+  ([#6091](https://github.com/matrix-org/matrix-rust-sdk/pull/6091))
```
poljar: New entries go to the top.

Hywan: it's also the wrong changelog. Fixed!

PR #6096, opened 2026-02-02
fix(latest_event): handle edits when the target event is not directly preceding

Johennes:
**Problem:** The latest event calculation currently only considers edits when the replacement event directly follows the replaced event with no other events in-between. This leads to incorrect latest event values in certain situations such as when an event is edited twice in a row.

This PR changes the calculation to remember the latest edit per event ID, continue iterating backwards until an actual matching event is found and then return that event's latest edit or the event itself if it hasn't been edited.

### DIFF
--- a/crates/matrix-sdk/CHANGELOG.md
+++ b/crates/matrix-sdk/CHANGELOG.md
@@ -53,6 +53,8 @@ All notable changes to this project will be documented in this file.
 
 ### Bugfix
 
+- Latest Event is correctly computed when multiple edits exist for the same event candidate.
+  ([#6096](https://github.com/matrix-org/matrix-rust-sdk/pull/6096))
 - Restrict which `m.room.member` can be a `LatestEventValue` candidate by relying on `MembershipChange` for more control.
   ([#6143](https://github.com/matrix-org/matrix-rust-sdk/pull/6143))
 - Add manual WAL checkpoints when opening Sqlite DBs and when vacuuming them, since the WAL files aren't automatically shrinking. ([#6004](https://github.com/matrix-org/matrix-rust-sdk/pull/6004))
--- a/crates/matrix-sdk/src/latest_events/latest_event/builder.rs
+++ b/crates/matrix-sdk/src/latest_events/latest_event/builder.rs
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 use std::{
+    collections::HashMap,
     iter::once,
     ops::{ControlFlow, Deref},
 };
@@ -57,14 +58,16 @@ impl Builder {
         let mut room_has_been_emptied = true;
         let mut current_value_must_be_erased = false;
 
+        // Track the most recent edit for each event.
+        let mut latest_edit_for_event: HashMap<OwnedEventId, TimelineEvent> = HashMap::new();
+
         if let Ok(Some(event)) = room_event_cache
-            .rfind_map_event_in_memory_by(|event, previous_event| {
+            .rfind_map_event_in_memory_by(|event| {
                 // At least one event lives in-memory: we consider the room is not empty.
                 room_has_been_emptied = false;
 
                 match filter_timeline_event(
                     event,
-                    previous_event,
                     current_value_event_id.as_ref(),
                     own_user_id,
                     power_levels,
@@ -72,14 +75,29 @@ impl Builder {
                     // Let's continue, event is not suitable.
                     ControlFlow::Continue(FilterContinue {
                         current_value_must_be_erased: erased,
+                        edited_event_id,
                     }) => {
                         current_value_must_be_erased = erased;
+
+                        if let Some(edited_event_id) = edited_event_id {
+                            // This is an edit. Store it if we haven't seen an edit for the
+                            // targeted event yet.
+                            latest_edit_for_event.entry(edited_event_id).or_insert(event.clone());
+                        }
 
                         None
                     }
 
                     // Stop! We found a suitable event!
-                    ControlFlow::Break(()) => Some(event.clone()),
+                    ControlFlow::Break(()) => {
+                        // Return the latest known edit of the event or the event itself if it
+                        // hasn't been replaced.
+                        event
+                            .event_id()
+                            .and_then(|event_id| latest_edit_for_event.get(&event_id))
+                            .cloned()
+                            .or_else(|| Some(event.clone()))
+                    }
                 }
             })
             .await
@@ -150,7 +168,6 @@ impl Builder {
                         Ok(content) => {
                             if filter_any_message_like_event_content(
                                 content,
-                                None,
                                 current_value_event_id.as_ref(),
                             )
                             .is_break()
@@ -277,7 +294,6 @@ impl Builder {
                         Ok(content) => {
                             if filter_any_message_like_event_content(
                                 content,
-                                None,
                                 current_value_event_id.as_ref(),
                             )
                             .is_break()
@@ -562,6 +578,8 @@ impl BufferOfValuesForLocalEvents {
 struct FilterContinue {
     /// Whether the current [`LatestEventValue`] must be erased or not.
     current_value_must_be_erased: bool,
+    /// When the event is a replacement, this is the targeted event ID.
+    edited_event_id: Option<OwnedEventId>,
 }
 
 /// Build the [`ControlFlow::Break`] for the filters.
@@ -571,12 +589,27 @@ fn filter_break() -> ControlFlow<(), FilterContinue> {
 
 /// Build the [`ControlFlow::Continue`] for the filters.
 fn filter_continue() -> ControlFlow<(), FilterContinue> {
-    ControlFlow::Continue(FilterContinue { current_value_must_be_erased: false })
+    ControlFlow::Continue(FilterContinue {
+        current_value_must_be_erased: false,
+        edited_event_id: None,
+    })
 }
 
 /// Build the [`ControlFlow::Continue`] with erasing, for the filters.
 fn filter_continue_with_erasing() -> ControlFlow<(), FilterContinue> {
-    ControlFlow::Continue(FilterContinue { current_value_must_be_erased: true })
+    ControlFlow::Continue(FilterContinue {
+        current_value_must_be_erased: true,
+        edited_event_id: None,
+    })
+}
+
+/// Build the [`ControlFlow::Continue`] with an edited event ID, for the
+/// filters.
+fn filter_continue_with_edit(edited_event_id: OwnedEventId) -> ControlFlow<(), FilterContinue> {
+    ControlFlow::Continue(FilterContinue {
+        current_value_must_be_erased: false,
+        edited_event_id: Some(edited_event_id),
+    })
 }
 
 /// Filter a [`TimelineEvent`].
@@ -584,14 +617,10 @@ fn filter_continue_with_erasing() -> ControlFlow<(), FilterContinue> {
 /// Be careful:
 ///
 /// - `event` is the current event in the collection of events that is scanned.
-/// - `previous_event` is the event sitting next to `event` in this collection,
-///   it's the event that comes before `event` (`previous_event` is older than
-///   `event`).
 /// - `current_value_event_id` is the event ID of the current
 ///   [`LatestEventValue`].
 fn filter_timeline_event(
     event: &TimelineEvent,
-    previous_event: Option<&TimelineEvent>,
     current_value_event_id: Option<&OwnedEventId>,
     own_user_id: &UserId,
     power_levels: Option<&RoomPowerLevels>,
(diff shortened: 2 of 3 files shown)


Johennes:
> But, we need to be very aware of the rule: an edit must be ignored if another event is inserted between the edit and the related event. And that's not what your code does as far as I understand it, right?

If I understood correctly, it should actually do just that. The new logic is to go backwards, remembering edits until you hit a target event. Then you return the latest edit of that event or the event itself if there are no edits.

In your example, the chain is: `X (edits A) > D (edits C) > C > B > A`

The code should skip over the two edits until it hits `C` and then take the latest edit for `C` which is `D`.

I've added a test case for it in a92ae5a44bd275c01d36af70dabcd32d2b41a558.

Hywan:
Thanks, I'll take a look.

crates/matrix-sdk/src/latest_events/latest_event/builder.rs
```diff
@@ -57,29 +57,48 @@ impl Builder {
         let mut room_has_been_emptied = true;
         let mut current_value_must_be_erased = false;
 
+        // Track the most recent edit for each event.
+        let mut latest_edit_for_event: std::collections::HashMap<OwnedEventId, TimelineEvent> =
```
Hywan: Please transform to `HashMap`, and add `use std::collections::HashMap` at the top of the module.

crates/matrix-sdk/src/latest_events/latest_event/builder.rs
```diff
@@ -57,29 +57,48 @@ impl Builder {
...
-                    ControlFlow::Break(()) => Some(event.clone()),
+                    ControlFlow::Break(()) => {
+                        // Return the latest known edit of the event or the event itself if it
+                        // hasn't been replaced.
+                        if let Some(event_id) = event.event_id()
+                            && let Some(edit) = latest_edit_for_event.get(&event_id)
+                        {
+                            return Some(edit.clone());
+                        }
+                        Some(event.clone())
```
Hywan: I would prefer a more Rust idiomatic way of writing this:


```suggestion
                        event
                            .event_id()
                            .and(|event_id| latest_edit_for_event.get(&event_id).clone())
                            .unwrap_or_else(|| event.clone())
```

crates/matrix-sdk/src/latest_events/latest_event/builder.rs
```diff
@@ -559,6 +576,9 @@ impl BufferOfValuesForLocalEvents {
 struct FilterContinue {
     /// Whether the current [`LatestEventValue`] must be erased or not.
     current_value_must_be_erased: bool,
+    /// When the event is a replacement, the edited event ID for tracking
+    /// in the outer iteration.
```
Hywan:
```suggestion
    /// When the event is a replacement, this is the targeted event ID.
```

No need to mention the outer iteration here.

crates/matrix-sdk/src/latest_events/latest_event/builder.rs
```diff
@@ -57,29 +57,48 @@ impl Builder {
         let mut room_has_been_emptied = true;
         let mut current_value_must_be_erased = false;
 
+        // Track the most recent edit for each event.
+        let mut latest_edit_for_event: std::collections::HashMap<OwnedEventId, TimelineEvent> =
+            std::collections::HashMap::new();
+
         if let Ok(Some(event)) = room_event_cache
-            .rfind_map_event_in_memory_by(|event, previous_event| {
+            .rfind_map_event_in_memory_by(|event, _| {
```
Hywan: You can clean up the `rfind_map_event_in_memory_by` method to not return the `previous_event` as part of the `predicate`, since it was the only place where it was used.

The bottom-most method should then look like this:

```rust
pub fn rfind_map_event_in_memory_by<O, P>(&self, mut predicate: P) -> Option<O>
where
    P: FnMut(&Event) -> Option<O>,
{
    self.state
        .room_linked_chunk
        .revents()
        .find_map(|event| predicate(event))
}
```

I reckon the lifetime `'i` can be removed from the original method (which I did here) without too much trouble.

Be careful to also update the documentation of this method.

crates/matrix-sdk/src/latest_events/latest_event/builder.rs
```diff
@@ -568,27 +588,37 @@ fn filter_break() -> ControlFlow<(), FilterContinue> {
...
+        edited_event_id: None,
+    })
+}
+
+/// Build the [`ControlFlow::Continue`] with an edited event ID, for the
+/// filters.
+fn filter_continue_with_edit(
+    edited_event_id: Option<OwnedEventId>,
+) -> ControlFlow<(), FilterContinue> {
+    ControlFlow::Continue(FilterContinue { current_value_must_be_erased: false, edited_event_id })
```
Hywan: Having an option for `edited_event_id` doesn't make sense here as we always have it.

```suggestion
fn filter_continue_with_edit(
    edited_event_id: OwnedEventId,
) -> ControlFlow<(), FilterContinue> {
    ControlFlow::Continue(FilterContinue { current_value_must_be_erased: false, edited_event_id: Some(edited_event_id) })
```

crates/matrix-sdk/src/latest_events/latest_event/builder.rs
```diff
@@ -647,13 +675,10 @@ fn filter_any_message_like_event_content(
...
-                    // latest event candidate, otherwise let's ignore it.
-                    if Some(event_id) == previous_event.and_then(|event| event.event_id()) {
-                        filter_break()
-                    } else {
-                        filter_continue()
-                    }
+                    // Edits are only suitable as latest events when the replaced event would
+                    // otherwise be the latest event. We pass the target event ID up from here
+                    // so that it can be tracked in the outer loop.
+                    filter_continue_with_edit(Some(event_id))
```
Hywan:
```suggestion
                    // Edits are suitable as latest events if and only if the targeted event is also suitable.
                    // Let's remember this edit, and the associated targeted event ID.
                    filter_continue_with_edit(event_id)
```

crates/matrix-sdk/src/latest_events/latest_event/builder.rs
```diff
@@ -812,46 +837,17 @@ mod filter_tests {
...
-        // With a previous event, and that's the one being replaced!
-        {
-            let previous_event =
-                Some(event_factory.text_msg("hello").event_id(event_id!("$ev0")).into_event());
-
-            assert!(
-                filter_timeline_event(&event, previous_event.as_ref(), None, user_id, None)
-                    .is_break()
-            );
-        }
```
Hywan, line 849: We need to test all these 3 cases. Please update them instead of removing them.

crates/matrix-sdk/src/latest_events/latest_event/builder.rs
```diff
@@ -2056,6 +2052,231 @@ mod builder_tests {
...
+                .unwrap();
+        }
+
+        let event_cache = client.event_cache();
+        event_cache.subscribe().unwrap();
+
+        let (room_event_cache, _) = event_cache.for_room(room_id).await.unwrap();
+
+        assert_remote_value_matches_room_message_with_body!(
+            // We get `event_id_1` because `event_id_2` isn't a candidate,
```
Hywan: `event_id_2` doesn't exist. Please fix the comment.

crates/matrix-sdk/CHANGELOG.md
```diff
@@ -46,7 +46,8 @@ All notable changes to this project will be documented in this file.
   ([#6017](https://github.com/matrix-org/matrix-rust-sdk/pull/6017))
 
 ### Bugfix
-
+- Handle edits when the target event is not directly preceding.
```
Hywan: A mention to the Latest Event API would be great.

```suggestion
- Latest Event is correctly computed when multiple edits exist for the same event candidate.
```

crates/matrix-sdk/src/latest_events/latest_event/builder.rs
Johennes: Hm, the `filter_timeline_event` function doesn't take a `previous_event` argument anymore. It just looks at the current event in isolation. So I think these tests don't apply anymore?

crates/matrix-sdk/CHANGELOG.md
Johennes: Good point. Applied in df421b591b6d92ead67233859cb544b591e78093.

crates/matrix-sdk/src/latest_events/latest_event/builder.rs
Johennes: My bad. I fixed the comment in 5a067cce2a993fdb59d8c45206829e5b6534ece5.

crates/matrix-sdk/src/latest_events/latest_event/builder.rs
Johennes: Applied in 283b8d71af6e5042e1400742f64529701fc52803.

crates/matrix-sdk/src/latest_events/latest_event/builder.rs
Johennes: Of course, yes. 🤦‍♂️ Fixed in e50bed249530711aed0f36698ed8146ee8ce78e4.

crates/matrix-sdk/src/latest_events/latest_event/builder.rs
Johennes: Applied in 91335f5de49f9bce1e9de7d5e54bbe56121c1c68.

crates/matrix-sdk/src/latest_events/latest_event/builder.rs
Johennes: Good point. I applied a slightly modified version of this in 7cf6d95ce864b112e9468bb7db9655dd30241570.

crates/matrix-sdk/src/latest_events/latest_event/builder.rs
Johennes: Fixed in bb2cccac80976e060e0adaee802c9aec2af90aa1.

crates/matrix-sdk/src/latest_events/latest_event/builder.rs
Johennes: Oh, I didn't realize this was the only usage. Have removed the parameter in 499dae95ba7bc3c0e9510bc70cbf9718dd7204d7 and the lifetime in c03930212f6997d132105f0b24df385a9f3f771f.

crates/matrix-sdk/src/latest_events/latest_event/builder.rs
Hywan: I think it's still relevant to test when an edit is found without the original event.

Johennes: Oh, you meant cover these cases in the other tests. Got it now, sorry. I have added two more tests in 70db95b7dbf930c1de18716ea40398cc7eb4477b.

crates/matrix-sdk/src/latest_events/latest_event/builder.rs
Hywan: Thanks!

crates/matrix-sdk/src/latest_events/latest_event/builder.rs
Hywan: Good enough to me, thanks.

crates/matrix-sdk/src/latest_events/latest_event/builder.rs
Hywan: Thanks.

crates/matrix-sdk/src/latest_events/latest_event/builder.rs
Hywan: Thanks!

crates/matrix-sdk/src/latest_events/latest_event/builder.rs
Hywan: Thanks.

crates/matrix-sdk/src/latest_events/latest_event/builder.rs
Hywan: 👍

crates/matrix-sdk/src/latest_events/latest_event/builder.rs
Hywan: Thanks!

crates/matrix-sdk/src/latest_events/latest_event/builder.rs
Hywan: Great, thanks!

Hywan:
We are good to go, thank you!
Can you rebase your fixups, and fix the conflict, so that we can merge this contribution?

Johennes:
Have cleaned up the history. Thanks for the review! ❤️

Hywan (approved):

Hywan:
Thank you for your patience and for your work!

Issue #6026, opened 2026-01-14
Events lose sender profile after member leaves DM room


JoFrost:
While testing reactions in a DM room, I ran into an unexpected behavior around sender profile resolution.

When the other participant leaves the room, previously received MessageLike events lose their resolved sender profile. The timeline item still references the sender MXID, but the associated profile data becomes empty:

```
EventTimelineItem {
    sender: "@user:matrix.org",
    sender_profile: Ready(
        Profile {
            display_name: None,
            display_name_ambiguous: false,
            avatar_url: None,
        },
    ),
...
}
```

If the same user later rejoins the room, the SDK immediately rehydrates the profile information for the existing event, without the event itself changing:

```
EventTimelineItem {
    sender: "@user:matrix.org",
    sender_profile: Ready(
        Profile {
            display_name: Some("Display name"),
            display_name_ambiguous: false,
            avatar_url: None,
        },
    ),
}
```

The same phenomenon was observed for state events. However, in that case, the sender’s display name is still available inside the event payload itself:

```
    TimelineItem {
        kind: Event(
            EventTimelineItem {
                sender: "@26uwjdyxbamf96wxty:...",
                sender_profile: Ready(
                    Profile {
                        display_name: None,
                        display_name_ambiguous: false,
                        avatar_url: None,
                    },
                ),
                timestamp: 2026-01-12T16:32:01.301,
                content: MembershipChange(
                    RoomMembershipChange {
                        user_id: "@26uwjdyxbamf96wxty:...",
                        content: Original {
                            content: RoomMemberEventContent {
                                avatar_url: None,
                                displayname: Some(
                                    "My user name...",
                                ),
                                is_direct: None,
                                membership: "join",
                                third_party_invite: None,
                                blurhash: None,
                                reason: None,
                                join_authorized_via_users_server: None,
                            },
                            prev_content: Some(
                                RoomMemberEventContent {
                                    avatar_url: None,
                                    displayname: Some(
                                        "My user name...",
                                    ),
                                    is_direct: Some(
                                        true,
                                    ),
                                    membership: "invite",
                                    third_party_invite: None,
                                    blurhash: None,
                                    reason: None,
                                    join_authorized_via_users_server: None,
                                },
                            ),
                        },
                        change: Some(
                            InvitationAccepted,
                        ),
                    },
                ),
                kind: Remote(
                    RemoteEventTimelineItem {
[...]

PR #6097, opened 2026-02-02
feat(base): improve profile updates for left/banned users

JoFrost:
Hello! This PR implements a potential fix for the issue #6026, following the investigation of the issue.

To briefly restate the context, during member sync, the SDK stores the full `m.room.member` event, but display name resolution relies only on `RoomMemberEventContent`. In some cases (for example when a user leaves a room), the current content no longer contains the profile fields needed to compute a display name, even though they are still available in `prev_content`.

This PR updates `RoomMember::display_name` to fall back to `prev_content` when the current content lacks the required fields, allowing the last known display name to be used.

This may not be the most idiomatic solution, but it seems to behave reliably in practice. The intent of this PR is to open discussion around whether this behavior is desirable and/or where it should be handled.

Signed-off-by:

### DIFF
--- a/crates/matrix-sdk-base/src/response_processors/profiles.rs
+++ b/crates/matrix-sdk-base/src/response_processors/profiles.rs
@@ -32,6 +32,9 @@ pub fn upsert_or_delete(
     // Senders can fake the profile easily so we keep track of profiles that the
     // member set themselves to avoid having confusing profile changes when a
     // member gets kicked/banned.
+    // We don't want to update the profile if the member is leaving the room, as the
+    // server may return a dummy empty profile along the leave event. We want to
+    // keep the last known profile in that case.
     if event.state_key() == event.sender() && *event.membership() != MembershipState::Leave {
         context
             .state_changes
@@ -54,5 +57,12 @@ pub fn upsert_or_delete(
             .entry(room_id.to_owned())
             .or_default()
             .push(event.state_key().clone());
+
+        // Address the edge case of "in-flight" profiles. If an earlier event in
+        // this same sync batch inserted a profile (for example user joined then got
+        // banned by an admin of a room), we MUST NOT reinsert it.
+        if let Some(room_profiles) = context.state_changes.profiles.get_mut(room_id) {
+            room_profiles.remove(event.state_key());
+        }
     }
 }
--- a/crates/matrix-sdk-ui/tests/integration/timeline/profiles.rs
+++ b/crates/matrix-sdk-ui/tests/integration/timeline/profiles.rs
@@ -54,21 +54,12 @@ async fn test_user_profile_after_being_banned() {
     let room = client.get_room(&DEFAULT_TEST_ROOM_ID).unwrap();
     let timeline = Arc::new(room.timeline().await.unwrap());
 
-    // Previous content of Alice's leave event
-    let prev_content_alice = PreviousMembership::new(MembershipState::Join)
-        .display_name("*profanities*")
-        .avatar_url("mxc://456".into());
-
-    // Same, but for Carol.
-    let prev_content_carol = PreviousMembership::new(MembershipState::Join)
-        .display_name("Carol")
-        .avatar_url("mxc://789".into());
-
-    // Build a simple timeline with Bob creating the room, Alice accepting an invite
-    // and sending a message, Bob sending a message, and Alice leaving
+    // Build a simple timeline with Bob joining the room, Alice accepting an invite,
+    // sending some messages, and then getting banned by Bob. Alice's profile should
+    // be unavailable after the ban, while Bob's profile should be unaffected.
     sync_builder.add_joined_room(
         JoinedRoomBuilder::new(&DEFAULT_TEST_ROOM_ID)
-            // Bob create the room
+            // Bob joins the room
             .add_timeline_event(
                 f.member(&BOB)
                     .display_name("Bob")
@@ -89,21 +80,11 @@ async fn test_user_profile_after_being_banned() {
             .add_timeline_event(f.text_msg("*insults*").sender(&ALICE))
             .add_timeline_event(f.text_msg("Nope. You are out.").sender(&BOB))
             .add_timeline_event(
-                f.member(&ALICE).membership(MembershipState::Ban).previous(prev_content_alice),
-            )
-            // Carol accepts an invite into the room
-            .add_timeline_event(
-                f.member(&CAROL)
-                    .display_name("Carol")
-                    .avatar_url("mxc://789".into())
-                    .membership(MembershipState::Join)
-                    .previous(MembershipState::Invite),
-            )
-            // Carol sends a text message
-            .add_timeline_event(f.text_msg("hi!").sender(&CAROL))
-            // Carol leaves the room
-            .add_timeline_event(
-                f.member(&CAROL).membership(MembershipState::Leave).previous(prev_content_carol),
+                f.member(&ALICE).membership(MembershipState::Ban).sender(&BOB).previous(
+                    PreviousMembership::new(MembershipState::Join)
+                        .display_name("*profanities*")
+                        .avatar_url("mxc://456".into()),
+                ),
             ),
     );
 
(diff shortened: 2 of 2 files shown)


crates/matrix-sdk-base/src/room/members.rs
```diff
@@ -258,11 +258,29 @@ impl RoomMember {
 
     /// Get the display name of the member if there is one.
     pub fn display_name(&self) -> Option<&str> {
-        if let Some(p) = self.profile.as_ref() {
-            p.as_original().and_then(|e| e.content.displayname.as_deref())
-        } else {
-            self.event.displayname_value()
+        // Profile cache first
+        // Not to fond of this double as_ref() chain, but it avoids an extra clone.
```
bnjbvr: I don't think the second `.as_ref()` should be necessary; the compiler ought to call `.as_original(&self)`, then the `name` is a deref so already a reference. Not sure what this `.as_ref()` call does?

crates/matrix-sdk-base/src/room/members.rs
```diff
@@ -258,11 +258,29 @@ impl RoomMember {
 
     /// Get the display name of the member if there is one.
     pub fn display_name(&self) -> Option<&str> {
-        if let Some(p) = self.profile.as_ref() {
-            p.as_original().and_then(|e| e.content.displayname.as_deref())
-        } else {
-            self.event.displayname_value()
+        // Profile cache first
```
bnjbvr:
```suggestion
        // Try from the cached profile first.
```

crates/matrix-sdk-base/src/room/members.rs
```diff
@@ -258,11 +258,29 @@ impl RoomMember {
...
+            .profile
+            .as_ref()
+            .as_ref()
+            .and_then(|p| p.as_original())
+            .and_then(|ev| ev.content.displayname.as_deref())
+        {
+            return Some(name);
+        }
+
+        // Then current event content
```
bnjbvr:
```suggestion
        // Then from the current event's content.
```

crates/matrix-sdk-base/src/room/members.rs
```diff
@@ -258,11 +258,29 @@ impl RoomMember {
...
+        if let Some(name) = self.event.displayname_value() {
+            return Some(name);
         }
+
+        // Last resort: prev_content
```
bnjbvr:
```suggestion
        // As a last resort, try the previous content.
```

crates/matrix-sdk-base/src/room/members.rs
```diff
@@ -275,11 +293,25 @@ impl RoomMember {
 
     /// Get the avatar url of the member, if there is one.
     pub fn avatar_url(&self) -> Option<&MxcUri> {
-        if let Some(p) = self.profile.as_ref() {
-            p.as_original().and_then(|e| e.content.avatar_url.as_deref())
-        } else {
-            self.event.avatar_url()
+        if let Some(url) = self
```
bnjbvr: It would be sweet to:

1. have the same kind of comments around here
2. remove the extraneous `.as_ref()` call below

crates/matrix-sdk-base/src/room/members.rs
JoFrost: Because self.profile is an `Arc<Option<_>>`, I needed to borrow through both the `Arc` and the `Option` to access the inner value without moving it. However, I realized I overcomplicated this, so I reverted to the way it was before.

JoFrost:
All done, thanks for your feedbacks! @bnjbvr

bnjbvr:
I'm putting this on hold at the moment, because we've discovered this might be intentional, after all (sorry about that!). If a user has been kicked of a room (which materializes by a `leave` `m.room.member` event, that will have no avatar or displayname set), because for example, their name contained profanities, then we wouldn't want to keep displaying their name all the time.

An alternative would be to *not* update timeline items predating the membership change, but this wouldn't address the moderation use case above. I'll chat with our internal UX folks on their opinions about this and will get back to this PR then.

JoFrost:
No worries! Thanks for letting me know.

JoFrost:
Actually that's cool, I can take care of it! It looks indeed very easy to handle, and I'll write an unit test for it!

JoFrost:
I think this is good now.

Banned users no longer resolve a profile in the timeline, while users who leave or get kicked still keep their last known profile. I was also able to remove the prev_content fallback entirely, and everything continues to behave correctly.

JoFrost:
Hey @bnjbvr! Following up on this PR. Please let me know if any changes are required :)

crates/matrix-sdk-base/src/response_processors/profiles.rs
```diff
@@ -32,7 +32,7 @@ pub fn upsert_or_delete(
     // Senders can fake the profile easily so we keep track of profiles that the
     // member set themselves to avoid having confusing profile changes when a
     // member gets kicked/banned.
-    if event.state_key() == event.sender() {
+    if event.state_key() == event.sender() && *event.membership() != MembershipState::Leave {
```
bnjbvr, line 38: Let's update the code comment above, to indicate that the profile would be empty upon leave, and that we want to keep the previous profile in this case.

crates/matrix-sdk-ui/tests/integration/timeline/profiles.rs
```diff
@@ -34,6 +35,199 @@ use wiremock::{
...
+    // and sending a message, Bob sending a message, and Alice leaving
+    sync_builder.add_joined_room(
+        JoinedRoomBuilder::new(&DEFAULT_TEST_ROOM_ID)
+            // Bob create the room
+            .add_timeline_event(
+                f.member(&BOB)
+                    .display_name("Bob")
+                    .avatar_url("mxc://123".into())
+                    .membership(MembershipState::Join),
+            )
```
bnjbvr: The comment is wrong: this is Bob joining the room 👀

crates/matrix-sdk-ui/tests/integration/timeline/profiles.rs
```diff
@@ -34,6 +35,199 @@ use wiremock::{
...
+                    .membership(MembershipState::Join)
+                    .previous(MembershipState::Invite),
+            )
+            // Bob sends a text message
+            .add_timeline_event(f.text_msg("hey!").sender(&BOB))
+            // Alice send some insults and gets banned
+            .add_timeline_event(f.text_msg("*insults*").sender(&ALICE))
+            .add_timeline_event(f.text_msg("Nope. You are out.").sender(&BOB))
+            .add_timeline_event(
+                f.member(&ALICE).membership(MembershipState::Ban).previous(prev_content_alice),
```
bnjbvr: I think you'd need to modify the sender, to make it realistic: it's likely Bob sending this ban event.

crates/matrix-sdk-ui/tests/integration/timeline/profiles.rs
```diff
@@ -34,6 +35,199 @@ use wiremock::{
...
+
+    // Previous content of Alice's leave event
+    let prev_content_alice = PreviousMembership::new(MembershipState::Join)
+        .display_name("*profanities*")
+        .avatar_url("mxc://456".into());
+
+    // Same, but for Carol.
+    let prev_content_carol = PreviousMembership::new(MembershipState::Join)
+        .display_name("Carol")
+        .avatar_url("mxc://789".into());
```
bnjbvr: Can these two variables be inlined into their one use? It's a bit hard to follow what's going on by defining them ahead of time here.

crates/matrix-sdk-ui/tests/integration/timeline/profiles.rs
```diff
@@ -34,6 +35,199 @@ use wiremock::{
...
+    // as she has been banned. The SDK **should** return an empty
+    // profile for banned users.
+    let alice_profile = timeline_items[4].as_event().unwrap().sender_profile();
+    match alice_profile {
+        TimelineDetails::Ready(profile) => {
+            assert_eq!(profile.display_name, None);
+            assert_eq!(profile.avatar_url, None);
+        }
+        _ => panic!("Expected Alice's profile to be ready"),
+    }
```
bnjbvr: I'd like more checks on this timeline item, notably that it's been sent by Alice indeed. Looking at this check, I have absolutely no clues if this is the correct event or not.

crates/matrix-sdk-ui/tests/integration/timeline/profiles.rs
```diff
@@ -34,6 +35,199 @@ use wiremock::{
...
+    let _response = client.sync_once(sync_settings.clone()).await.unwrap();
+    server.reset().await;
+
+    mock_encryption_state(&server, false).await;
+
+    let room = client.get_room(&DEFAULT_TEST_ROOM_ID).unwrap();
+    let timeline = Arc::new(room.timeline().await.unwrap());
+
+    // Previous content of Alice's leave event
+    let prev_content = PreviousMembership::new(MembershipState::Join)
```
bnjbvr: same comment about inlining this into its one use

crates/matrix-sdk-ui/tests/integration/timeline/profiles.rs
```diff
@@ -34,6 +35,199 @@ use wiremock::{
...
+        .display_name("Alice")
+        .avatar_url("mxc://...".into());
+
+    // Build a simple timeline with Bob creating the room, Alice accepting an invite
+    // and sending a message, Bob sending a message, and Alice leaving
+    sync_builder.add_joined_room(
+        JoinedRoomBuilder::new(&DEFAULT_TEST_ROOM_ID)
+            // Bob create the room
```
bnjbvr: same comment about joining, not creating

crates/matrix-sdk-ui/tests/integration/timeline/profiles.rs
```diff
@@ -34,6 +35,199 @@ use wiremock::{
...
+
+    // Carol has left the room, but her profile should still be available.
+    let carol_profile = timeline_items[8].as_event().unwrap().sender_profile();
+    match carol_profile {
+        TimelineDetails::Ready(profile) => {
+            assert_eq!(profile.display_name, Some("Carol".to_owned()));
+            assert_eq!(profile.avatar_url, Some("mxc://789".into()));
+        }
+        _ => panic!("Expected Carol's profile to be ready"),
+    }
```
bnjbvr: Since the next test makes sure that the profile of a left member is still available, can we reduce the scope of this one test, and only test the ban use case? In particular, we could remove all the events related to Carole, as these events are meant to reproduce the same tests as in the next test function.

crates/matrix-sdk-base/src/response_processors/profiles.rs
```diff
@@ -41,8 +41,8 @@ pub fn upsert_or_delete(
             .insert(event.sender().to_owned(), event.into());
     }
 
-    if *event.membership() == MembershipState::Invite {
-        // Remove any profile previously stored for the invited user.
+    if matches!(*event.membership(), MembershipState::Invite | MembershipState::Ban) {
+        // Remove any profile previously stored for the invited/banned user.
```
bnjbvr, line 48: Thanks for updating the comment here :)

crates/matrix-sdk-ui/tests/integration/timeline/profiles.rs
JoFrost: Regarding that, if I actually specify the sender as Bob, Alice's profile end up showing up again for some reasons, although the response processor does delete Alice's profile... Not sure why yet, but it's most likely coming from my test.

JoFrost:
Hey @bnjbvr! 
Sorry for coming back on this PR so late. I did the changes. The changes showed up an interesting bug that took me a small while to investigate and fix. 

When we insert multiple events at once, and have at the same time joining AND banning (and when the ban event is from a different sender), it could happens that the profile was not wiped out properly. I made a specific check for this case, which solved the issue and allowed the test to pass (without breaking anything else I hope!). 

Lmk what you think! :)

PR #6101, opened 2026-02-02
ci: Add a `dist` profile with optimised binary size

jmartinesp:
## Changes:

- Adds a new `dist` profile that can be used by the bindings to generate library versions with optimised binary size. This was specially a pain point for Android: with this config and some debug symbol stripping, the library size is now 50-70MB smaller *per architecture*, resulting in very significant file size reduction for universal APKs, which have 4 archs.
- Also create new `reldev` profile for FFI bindings consumers that allows having way smaller binaries while still being able to have proper backtraces: `reldbg` is great for iOS because it allows inline debugging using LLDB in Xcode, but it produces enormous binaries (last I checked was 1.25GB), while for Android we can't use that properly and we'd only be interested in having symbolicated backtraces, which this profile achieves with binaries an order of magnitude smaller.

I'm aware `reldev` is quite easy to mistake wirh `reldbg`, so I'm open to finding a better name for it 😅 .

Signed-off-by:

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -169,6 +169,19 @@ unused_async = "warn"
 # Saves a lot of disk space. If symbols are needed, use the dbg profile.
 debug = 0
 
+# Profile for debug builds with full optimization and minimal debug symbols.
+# This should be just enough to have proper backtraces, having way smaller binaries
+# (10% of the size with full debug symbols profile, like `reldbg`).
+# This profile differs from `reldbg` in not containing the debug symbols needed for
+# debugging with LLDB/GDB, trading that for binary size, allowing quick iterations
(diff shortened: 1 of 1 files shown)


Cargo.toml
```diff
@@ -169,6 +169,14 @@ unused_async = "warn"
 # Saves a lot of disk space. If symbols are needed, use the dbg profile.
 debug = 0
 
+# Profile for debug builds with full optimization and minimal debug symbols.
+# This should be just enough to have proper backtraces, having way smaller binaries.
```
Hywan: Can you explain the difference between `reldev` and `reldbg` from a developer point of view please?

Hywan:
Great comment, thanks!

Issue #5989, opened 2025-12-26
[meta] Reducing the number of dependencies


Hywan:
Hello,

This project has too many dependencies according to me.

The number of dependencies (regardless of their versions, features etc.)

```shell
$ cargo tree --all-features --edges all --prefix none | rg -v '^$' | cut -d' ' -f 1 | sort | uniq | wc -l
539
```

This is a meta issue aiming at reducing the number of dependencies, thus reducing the compile-time, and the resulting binary size.

## Removing duplicated dependencies

We have 35 duplicated dependencies:

```shell
$ cargo tree --all-features --duplicates | rg '^[a-z]' | datamash -W groupby 1 collapse 2 | tr '\t,' ' ' | sed -E -e 's/^([a-z0-9_\-]+)/* [ ] `\1`:/' | sed -E -e 's/(v.*) v/\1, v/' | sed -E -e 's/v([0-9]+)/\1/g'
```

* [ ] `bitflags`: 2.10.0, 2.10.0
* [ ] `camino`: 1.2.1, 1.2.1
* [ ] `crypto-common`: 0.1.6, 0.1.6
* [ ] `darling`: 0.20.10, 0.21.1
* [ ] `darling_core`: 0.20.10, 0.21.1
* [ ] `darling_macro`: 0.20.10, 0.21.1
* [ ] `derive_more`: 1.0.0, 2.0.1
* [ ] `derive_more-impl`: 1.0.0, 2.0.1
* [ ] `digest`: 0.10.7, 0.10.7
* [ ] `getrandom`: 0.2.15, 0.3.1
* [ ] `hashbrown`: 0.14.5 0.15.2, 0.16.1
* [ ] `indexmap`: 2.12.1, 2.12.1
* [ ] `itertools`: 0.10.5 0.13.0, 0.14.0
* [ ] `js_int`: 0.2.2, 0.2.2
* [ ] `libc`: 0.2.175, 0.2.175
* [ ] `phf`: 0.11.2, 0.13.1
* [ ] `phf_shared`: 0.11.2, 0.13.1
* [ ] `rand`: 0.8.5, 0.8.5
* [ ] `rand_core`: 0.6.4 0.6.4, 0.9.3
* [ ] `ruma-identifiers-validation`: 0.12.0, 0.12.0
* [ ] `rustix`: 0.38.41, 1.0.8
* [ ] `semver`: 1.0.27, 1.0.27
[...]

PR #6104, opened 2026-02-03
chore: Use `default-features = false` for most dependencies

Hywan:
This patch adds `default-features = false` to all dependencies to avoid fetching useless dependencies by default.

One dependency is also removed, and a couple are updated.

Numbers:

- `cargo check --workspace --tests` jumps from 725 to 710,
- `cargo tree --all-features --edges all --prefix none | rg -v '^$' | cut -d' ' -f 1 | sort | uniq | wc -l` jumps from 538 to 529.

There is a bit more work for the `matrix-sdk` crate, but I want to tackle them separately.

---

- Address https://github.com/matrix-org/matrix-rust-sdk/issues/5989

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -601,6 +601,29 @@ dependencies = [
 ]
 
 [[package]]
+name = "cargo-platform"
+version = "0.1.9"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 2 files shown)


andybalaam (approved):
Looks good, thank you!

PR #6109, opened 2026-02-03
refactor(event cache): a few renamings around the `RoomEventCacheState` and locks

bnjbvr:
I've also shortened the comment in the `read` method, as talking about the previous implementations that didn't work in such detail doesn't help that much IMO. See the description commit by commit.

### DIFF
--- a/crates/matrix-sdk/src/event_cache/room/mod.rs
+++ b/crates/matrix-sdk/src/event_cache/room/mod.rs
@@ -723,9 +723,12 @@ mod private {
         /// The per-thread lock around the real state.
         locked_state: RwLock<RoomEventCacheState>,
 
+        /// A lock taken to avoid multiple attempts to upgrade from a read lock
+        /// to a write lock.
+        ///
         /// Please see inline comment of [`Self::read`] to understand why it
         /// exists.
-        read_lock_acquisition: Mutex<()>,
+        state_lock_upgrade_mutex: Mutex<()>,
     }
 
     struct RoomEventCacheState {
@@ -885,7 +888,7 @@ mod private {
                     waited_for_initial_prev_token,
                     subscriber_count: Default::default(),
                 }),
-                read_lock_acquisition: Mutex::new(()),
+                state_lock_upgrade_mutex: Mutex::new(()),
             })
         }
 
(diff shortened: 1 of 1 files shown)


crates/matrix-sdk/src/event_cache/room/mod.rs
```diff
@@ -902,49 +902,16 @@ mod private {
...
-            // Note that it doesn't conflict with the `write` method because this later
-            // immediately obtains a write lock, which avoids any conflict with this method.
-            //
-            // [^1]: https://docs.rs/lock_api/0.4.14/lock_api/struct.RwLock.html#method.upgradable_read
-            // [^2]: https://docs.rs/async-lock/3.4.1/async_lock/struct.RwLock.html#method.upgradable_read
+            // Other attempts have been done in the past, including but not limited to
+            // using: an upgradeable read lock, a downgrading write lock, atomic
+            // read and write locks, and a semaphore with one permit. Each of
+            // these attempts had their own problems, so they've been ditched in
+            // favor of this solution.
```
Hywan: I'm okay with this summary, but the details are important. Please restore the previous explanations.

crates/matrix-sdk/src/event_cache/room/mod.rs
```diff
@@ -723,9 +723,12 @@ mod private {
         /// The per-thread lock around the real state.
         locked_state: RwLock<RoomEventCacheState>,
 
+        /// A lock taken to avoid multiple attempts to upgrade from a read lock
+        /// to a write lock.
+        ///
         /// Please see inline comment of [`Self::read`] to understand why it
         /// exists.
-        read_lock_acquisition: Mutex<()>,
+        state_lock_upgrade_mutex: Mutex<()>,
```
Hywan, line 731: This lock is only used for the `read` operation. Please revert the renaming, as it now creates confusion about where it's used. An “upgrade from a read lock to a write lock” is supposedly only happening in the `read` operation, but it might not be the case in the future. Let's not try to be generic here.

crates/matrix-sdk/src/event_cache/room/mod.rs
bnjbvr: Alright, reverted this commit.

PR #6112, opened 2026-02-03
fix(ffi): Remove checksums in all crates using UniFFI

jmartinesp:
[Previously](https://github.com/matrix-org/matrix-rust-sdk/pull/6069), this was only applied to the FFI bindings crate, but it's not the only one affected by the issue with 32bit ARM checksum validations.

See https://github.com/element-hq/element-x-android/issues/6028#issuecomment-3843942410.

Signed-off-by:


jmartinesp:
Actually, I just realised I forgot to upload the most important `uniffi.toml` file, the one for the `matrix-sdk` crate... I'll create a new PR with that.

PR #6120, opened 2026-02-04
feat(xtask): Display `pos` and `timeout` for sync report

Hywan:
Follow up of https://github.com/matrix-org/matrix-rust-sdk/pull/6118.

This patch updates the `cargo xtask log sync` command to extract the `pos` and `timeout` fields so that we can display them. The `timeout` is displayed in its own column, while the `pos` is displayed in the line details.

Screenshot: see the new “Req. timeout” column, and the “Sliding Sync `pos`” list item in the log line details.

Screenshot 2026-02-04 at 17-09-41 Network viewer

🆕  After the feature requested by @poljar to display the timeout in the duration graph, it looks like this:

Screenshot 2026-02-05 at 10-44-35 Network viewer

- When the duration is smaller than the timeout, the timeout is not displayed in the graph (e.g. with REQ-7 and REQ-9).
- When the duration is longer than the duration (it can happen if the server doesn't return because it is calculating a response), the total duration is larger than the timeout, and then the timeout is displayed (e.g. with REQ-6).
- When the timeout is reached with no response, the bar is filled by the timeout only (e.g. with REQ-8).

### DIFF
--- a/xtask/src/log/sync.rs
+++ b/xtask/src/log/sync.rs
@@ -189,8 +189,8 @@ pub(super) fn run(log_path: path::PathBuf, output_path: path::PathBuf) -> Result
                     },
                 )| {
                     let uri = Url::parse(uri);
-                    let duration = duration.num_milliseconds();
-                    let timeout = timeout.map(|timeout| timeout.as_millis());
+                    let duration = duration.num_milliseconds().abs() as u64;
+                    let timeout: Option<u64> = timeout.map(|timeout| timeout.as_millis().try_into().expect("Failed to cast a u128 to u64"));
 
                     format!(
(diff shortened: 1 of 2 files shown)


xtask/src/log/sync.rs
poljar: Shouldn't we substract the timeout from the duration? This will let us know how long a request was long-polling and how long we actually had to wait for the server to respond.

poljar (approved):

xtask/src/log/sync.rs
Hywan: Done Sir! It's not subtracted, but both information are displayed, as discussed in private. See the PR's description to see the details.

PR #6121, opened 2026-02-04
doc: clarify that IO_ELEMENT_MSC4388 is the unstable prefix

hughns:
This is a minor clarification to the inline comments that were part of https://github.com/matrix-org/matrix-rust-sdk/pull/6089

Given that https://github.com/matrix-org/matrix-rust-sdk/pull/6089 has not been included in a release maybe this doesn't need a separate changelog entry?


poljar:
> Given that #6089 has not been included in a release maybe this doesn't need a separate changelog entry?

That's correct.

PR #6126, opened 2026-02-05
fix(sliding-sync): re-dispatch invited/knocked rooms after reinvite

VEL0C1TY22:
## Problem

When a room goes through:

join -> leave/kick -> re-invite
Sliding Sync responses still include the room, but the SDK does not emit an
Invited/Knocked update. Because of this, RoomListService is never notified and
the room disappears from the room list until the application restarts.

## Root cause
"update_any_room()" only returned a RoomUpdateKind when one was explicitly
produced. For re-invites it could return "None", causing the room to be skipped.

## Fix
Always emit a default:
- "RoomUpdateKind::Invited"
- "RoomUpdateKind::Knocked"
when the room state is Invited/Knocked but no update kind was generated.

## Tests
Added a regression test covering:
join -> leave -> re-invite to ensure the room is surfaced again in the invited list.

### DIFF
--- a/crates/matrix-sdk-base/src/sliding_sync.rs
+++ b/crates/matrix-sdk-base/src/sliding_sync.rs
@@ -358,9 +358,7 @@ mod tests {
         assert!(sync_resp.rooms.invited.contains_key(room_id));
     }
 
-    use ruma::{
-        events::AnyStrippedStateEvent,
-    };
+    use ruma::events::AnyStrippedStateEvent;
 
     fn invite_state_for(
         user_id: &UserId,


crates/matrix-sdk-base/src/sliding_sync.rs
```diff
@@ -339,6 +339,44 @@ mod tests {
...
+        // leave/kick
+        let mut room = http::response::Room::new();
+        set_room_left(&mut room, &user_id);
+
+        let response = response_with_room(room_id, room);
+        client.process_sliding_sync(&response, &RequestedRequiredStates::default()).await.unwrap();
+
+        // reinvite 
+        let mut room = http::response::Room::new();
+        set_room_invited(&mut room, &user_id, &user_id);
```
stefanceriu: We should do the same flow for `set_room_knocked` to cover both cases.

stefanceriu (commented):
Very nice, thank you for contributing! 👏 

Will merge as soon as you add a test path for knocking.

crates/matrix-sdk-base/src/sliding_sync.rs
VEL0C1TY22: Sure I will be working on that !

VEL0C1TY22:
Hello @stefanceriu , 
I have made the changes mentioned above so kindly review them and let me know if more changes are needed 
Thank you!

stefanceriu (approved):
Beautiful, thank you!

PR #6128, opened 2026-02-05
refactor(event cache): common out some persistence related helpers

bnjbvr:
This starts commoning out some persistence utils for the linked chunk:

- first, starts with the function helpers related to stripping bundled relations
- then the `send_updates_to_store` function, which is quite similar in both cases

Also, beefed up the comment explaining *why* we're stripping bundled relations from events before storing them in the event cache.

A next step would be to generalize the read/write state guards, using a trait with a generic parameter for the state type, and with an associated function `on_dirty_store`. This could remove a lot of similarly looking code!

Part of #5954.

### DIFF
--- a/crates/matrix-sdk/src/event_cache/persistence.rs
+++ b/crates/matrix-sdk/src/event_cache/persistence.rs
@@ -37,6 +37,17 @@ pub(super) async fn send_updates_to_store(
     }
 
     // Strip relations from updates which insert or replace items.
+    //
+    // The reason we're doing this, is that consumers of the event cache might look
+    // into bundled relations, and assume they're up to date. If we were to keep
+    // the relations in the events, when storing them, then it could be that
+    // they become outdated (as soon as a new relation comes over sync), so we'd
+    // need to update the bundled relations in this case, which would
+    // have a non-negligible cost, as we'd need to look up related events for each
+    // forwarded to a listener.
+    //
+    // As a result, we choose to strip bundled relations from events when we forward
+    // them to the store, and consumers have to explicitly ask for relations.
     for update in updates.iter_mut() {
         match update {
             Update::PushItems { items, .. } => strip_relations_from_events(items),


crates/matrix-sdk/src/event_cache/room/pinned_events.rs
```diff
@@ -16,9 +16,10 @@ use std::{collections::BTreeSet, sync::Arc};
 
 use futures_util::{StreamExt as _, stream};
 #[cfg(feature = "e2e-encryption")]
+use matrix_sdk_base::deserialized_responses::TimelineEventKind;
```
Hywan: Could be merged with the next `use`:

```rust
use matrix_sdk_base::{deserialized_response::…, linked_chunk::…};
```

crates/matrix-sdk/src/event_cache/persistence.rs
```diff
@@ -0,0 +1,61 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use matrix_sdk_base::{deserialized_responses::TimelineEventKind, event_cache::Event};
+use ruma::serde::Raw;
+
+/// Strips the bundled relations from a collection of events.
+pub(in crate::event_cache) fn strip_relations_from_events(items: &mut [Event]) {
```
Hywan:
```rust
pub(in super)
```

I think it's better to have relative scope here, in case of a later refactoring.

crates/matrix-sdk/src/event_cache/persistence.rs
```diff
@@ -0,0 +1,61 @@
...
+    for ev in items.iter_mut() {
+        strip_relations_from_event(ev);
+    }
+}
+
+/// Strips the bundled relations from an event, if they were present.
+pub(in crate::event_cache) fn strip_relations_from_event(ev: &mut Event) {
```
Hywan:
```rust
pub(in super)
```

PR #6132, opened 2026-02-06
Spaces: Check the power levels before removing an m.space.parent event.

pixlwave:
Small bug fix that makes `remove_child_from_parent` behave like `add_child_to_parent`.

Additionally introduces a new Error case so that clients can decide to ignore any failures updating the child → parent relationship.

### DIFF
--- a/crates/matrix-sdk-ui/src/spaces/mod.rs
+++ b/crates/matrix-sdk-ui/src/spaces/mod.rs
@@ -1731,13 +1731,15 @@ mod tests {
                 );
             }
 
-            if let Some(power_level) = parameters.power_level {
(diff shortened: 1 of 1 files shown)


stefanceriu (approved):
lgtm 👍

stefanceriu:
... you're going to have to update the test though or even improve them.

pixlwave:
> ... you're going to have to update the test though or even improve them.

My bad, it's because the power levels weren't set. There's already tests for specific power levels but the failures were the ones using the default power levels and so were no longer compatible with the changes this PR made.

Please feel free to squash-merge this PR if the history isn't clean enough 👍

PR #6134, opened 2026-02-06
Minor refactoring for the rendezvous channel

poljar:
I discovered another place where we'll soon need to account for MSC4388.


andybalaam (approved):
Looks good, thanks!

PR #6135, opened 2026-02-06
feat(client): make it possible to subscribe to key upload errors

Johennes:
We are apparently hitting this error in our project for .. currently unknown reasons. As a client, it's helpful to be notified about this because the issue doesn't fix itself otherwise.

### DIFF
--- a/crates/matrix-sdk/src/client/mod.rs
+++ b/crates/matrix-sdk/src/client/mod.rs
@@ -100,7 +100,6 @@ use crate::{
     client::thread_subscriptions::ThreadSubscriptionCatchup,
     config::{RequestConfig, SyncToken},
     deduplicating_handler::DeduplicatingHandler,
-    encryption::DuplicateOneTimeKeyErrorMessage,
     error::HttpResult,
     event_cache::EventCache,
     event_handler::{
@@ -120,7 +119,10 @@ use crate::{
 #[cfg(feature = "e2e-encryption")]
 use crate::{
     cross_process_lock::CrossProcessLock,
-    encryption::{Encryption, EncryptionData, EncryptionSettings, VerificationState},
+    encryption::{
+        DuplicateOneTimeKeyErrorMessage, Encryption, EncryptionData, EncryptionSettings,
+        VerificationState,
+    },
 };
 
 mod builder;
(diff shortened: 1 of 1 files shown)


crates/matrix-sdk/src/encryption/mod.rs
```diff
@@ -700,6 +700,15 @@ impl Client {
                                                 StateStoreDataValue::OneTimeKeyAlreadyUploaded,
                                             )
                                             .await?;
+
+                                        if let Err(e) =
+                                            self.inner.duplicate_key_upload_error_sender.send(())
```
poljar: Don't you want the `DuplicateOneTimeKeyErrorMessage` be sent out to listeners?

Johennes: Hm, yes that might be helpful for further debugging actually. Maybe the error as string is enough though? Otherwise, I think, we'll have to send `Option<DuplicateOneTimeKeyErrorMessage>`.

Johennes:
> Any more info on why you think you might be seeing this?
> 
> Do not it's quite easy to trigger this failure if you create two `Client` objects for the same user objects backed by the same crypto store, unless you use cross-process locking.
> 
> Element X on Android used to do this 1-2 years ago.

No further insights so far, sadly. It's most probably an issue on the integration side but we haven't pinned it down yet. I'll double check the cross-process locking though, thanks.

crates/matrix-sdk/src/encryption/mod.rs
poljar: Any problem with sending that? I think we can make that easily `Clone`.

That way you could see on the homeserver when the old key was uploaded to pinpoint when/where things started to go bad.

Johennes: > Any problem with sending that? 

No, I guess not really. Have made the change in 580b8f08441c02450a6ba510cb407c3d5294b2e7.

Johennes: Sorry, I forgot to add a feature guard. Have made another fixup in 19e3bd5853adb87607fe6f941109f443135f8a85.

PR #6136, opened 2026-02-06
Exclude service members from room names when using sliding sync to get notifications

### LINKED ISSUE #6107: Functional members are not ignored when the NotificationClient is computing the room displayName
Functional members are not ignored when the `NotificationClient` is computing the room `displayName`

For Alice to reproduce the issue:
- Alice has a dm with Bob
- Alice invite a new member, Charlie
- Let Charlie be a functional member
- The room name is correct in the room list of Alice: "Bob", i.e. the functional member is not taken into account. This is expected.
- Alice receives a message (from Bob or Charlie) in this room
- The Notification Client is providing a `NotificationRoomInfo.displayName` which contains the name of the functional member: "Charlie, Bob"

This is not correct, the `displayName` should be "Bob".

mgoldenberg:
# Overview

When a room's display name is computed, it is meant to exclude service members from the name, as per #4335. While this had been implemented effectively in most scenarios, a room's display name still includes service members when it is encapsulated in a notification (see #6107). 

The behavior in #6107 has been replicated in a unit test, see [`test_notification_room_display_name_excludes_service_members`](https://github.com/mgoldenberg/matrix-rust-sdk/blob/d0dec3f471aea7b229a922f06bc3cdba696ac093/crates/matrix-sdk-ui/src/notification_client.rs#L1084).

# Changes

Interestingly, the behavior only occurs in the test above when the notification is retrieved using sliding sync - i.e., [`NotificationClient::get_notification_with_sliding_sync`](https://docs.rs/matrix-sdk-ui/latest/matrix_sdk_ui/notification_client/struct.NotificationClient.html#method.get_notification_with_sliding_sync) - but not when it is retrieved using the `/context` endpoint - i.e., [`NotificationClient::get_notification_with_context`](https://docs.rs/matrix-sdk-ui/latest/matrix_sdk_ui/notification_client/struct.NotificationClient.html#method.get_notification_with_context).

This was fixed by adding `StateEventType::MemberHints` to the required state for the `SlidingSyncList` created in the `NotificationClient` - thanks to @jmartinesp for pointing this out!

---
Closes #6107.

### DIFF
--- a/crates/matrix-sdk-ui/CHANGELOG.md
+++ b/crates/matrix-sdk-ui/CHANGELOG.md
@@ -8,6 +8,9 @@ All notable changes to this project will be documented in this file.
 
 ### Bug Fixes
 
+- Ensure that the display name of a `Room` in a `NotificationStatus` coming 
+  from a `NotificationClient` excludes service members.
+  ([#6136](https://github.com/matrix-org/matrix-rust-sdk/pull/6136))
 - Fix the `is_last_admin` check in `LeaveSpaceRoom` since it was not
   accounting for the membership state.
   [#6032](https://github.com/matrix-org/matrix-rust-sdk/pull/6032)
@@ -18,9 +21,6 @@ All notable changes to this project will be documented in this file.
 - Fix the redecryption of events in timelines built using the
   `TimelineFocus` of `PinnedEvents`, `Thread`, `Event`.
   ([#5955](https://github.com/matrix-org/matrix-rust-sdk/pull/5955))
-- Ensure that the display name of a `Room` in a `NotificationStatus` coming 
-  from a `NotificationClient` excludes service members.
-  ([#6136](https://github.com/matrix-org/matrix-rust-sdk/pull/6136))
 
 ### Features
 


jmartinesp:
> So, I tried to query the state store (provided as an argument to Global::process_direct_rooms) for MemberHintsEvents, but no events were found - i.e., no service members were identified.

The state store in the `NotificationClient` is an in-memory one, so it might not contain the info you need 🫤 . Also, it's not part of the required state of the custom sliding sync instance we create in the `NotificationClient` either, so maybe that's why this in-memory store lacks the needed info:

https://github.com/matrix-org/matrix-rust-sdk/blob/db56d14177034ff8f7a1e19b81907305d520abd4/crates/matrix-sdk-ui/src/notification_client.rs#L488-L499

mgoldenberg:
Thank you @jmartinesp, very helpful!

crates/matrix-sdk-ui/CHANGELOG.md
```diff
@@ -18,6 +18,9 @@ All notable changes to this project will be documented in this file.
 - Fix the redecryption of events in timelines built using the
   `TimelineFocus` of `PinnedEvents`, `Thread`, `Event`.
   ([#5955](https://github.com/matrix-org/matrix-rust-sdk/pull/5955))
+- Ensure that the display name of a `Room` in a `NotificationStatus` coming
```
poljar: New entries to the top please.

mgoldenberg: Oops! Will do!

mgoldenberg: Also, sorry - realizing I probably should have made these extra change log commits fixups! I can rebase, squash them together, and force push, if you'd like - up to you!

poljar: Not that important I would say.

Issue #6137, opened 2026-02-06
Can't restore my session after upgrading Ruma


jmartinesp:
After running the Element X Android app with a bindings version containing [the latest Ruma version](https://github.com/matrix-org/matrix-rust-sdk/pull/6131), I can no longer restore my session. I see this log instead:

```
matrix_sdk_sqlite::utils: Failed to deserialize matrix_sdk_base::room::room_info::RoomInfo in a store: base_info.canonical_alias sentry=true err=base_info.canonical_alias: unknown variant `PossiblyRedacted`, expected `Original` or `Redacted` at line 1 column 1204 | crates/matrix-sdk-sqlite/src/utils.rs:636 | spans: restore_session_with
```

poljar:
False alarm, see this comment for more info: https://github.com/matrix-org/matrix-rust-sdk/pull/6138#issuecomment-3861244025.

PR #6138, opened 2026-02-06
Revert "refactor(base): Use PossiblyRedactedStateEventContent bound in MinimalStateEvent

poljar:
Seems that https://github.com/matrix-org/matrix-rust-sdk/pull/6088 causes some breakage when deserializing the room info.

Issue can be found here: #6137.

This reverts commit 9ad7cf9662f3c0df4441e30b52fd4aa6bfa262a1.


stefanceriu:
🐑 ノ( º _ ºノ)

poljar:
Turns out, this was a false alarm due to updating the SDK version and then going back to an older commit.

PR #6140, opened 2026-02-09
feat(xtask): Add `log overview`

Hywan:
In the same spirit of #6080.

This patch adds the `cargo xtask log overview` task that generates a standalone HTML report representing the logs as a tree where each node is a target, and each leaf is a log location with occurrences, spans and fields.

Each node displays the sum of errors and warnings for this node and its children. It helps to quickly spot the problematic targets, and it guides to exploration of the node without taking the reader's hand by trying to draw conclusions. We don't want to guide the reader to a mistake: we just want to guide the reader to draw its own conclusions.

Each line can be highlighted. When a node is folded/closed, it is also highlighted if at least one of its child is highlighted.

The occurrences of logs are displayed on a timeline.

#### Example

Here is a generated report from a rageshake: [test.html](https://github.com/user-attachments/files/25180681/test.html).

#### Screenshots

<figure>

Screenshot 2026-02-09 at 11-12-25 Overview

<figcaption>

The page when opened for the first time. We can see the top nodes.

</figcaption>
</figure>

<figure>

Screenshot 2026-02-09 at 11-12-55 Overview

<figcaption>

The `matrix_sdk` contains 500 errors. When opened, we see that the `latest_events` contains all these 500 errors. When opened, we see that 1 error is in `latest_event`, and the other 499 are in `room_latest_events`. The exploration is guided.

</figcaption>
</figure>

<figure>

Screenshot 2026-02-09 at 11-13-50 Overview

<figcaption>

We can see the occurrences of logs for `latest_events::latest_event::builder` and `latest::events::room_latest_events` in a timeline, with the fields represented as a table.

</figcaption>
</figure>

<figure>

Screenshot 2026-02-09 at 11-16-34 Overview

<figcaption>

We can highlight every line in this tree. It helps to keep track of data, helping with the cognitive burden of debugging via logs.

</figcaption>
</figure>

<figure>

Screenshot 2026-02-09 at 11-17-31 Overview

<figcaption>

Highlights are "remembered" when the bottom-most node is closed. It helps to navigate in the tree without loosing context.

Pro tip: this is done with a single CSS selector: `details:not([open]):has(.highlight > input:checked)::before`, CSS is so great.

</figcaption>
</figure>

### DIFF
--- a/xtask/src/log/overview.rs
+++ b/xtask/src/log/overview.rs
@@ -230,7 +230,7 @@ pub(super) fn run(log_path: path::PathBuf, output_path: path::PathBuf) -> Result
         writeln!(
             buffer,
             "<li>\n\
-            <details><summary>(root)\
+            <details open><summary>(root)\
             <span class=\"aside\">\
             <span class=\"highlight\"><input type=\"checkbox\" title=\"Highlight\" /></span>\
             </span>\


xtask/src/log/overview.rs
```diff
@@ -0,0 +1,481 @@
...
+                .entry(source_file.to_owned())
+                .or_insert_with(Spans::default)
+                // `Spans`
+                .entry(source_line)
+                .or_insert_with(BTreeMap::default)
+                // `Span`!
+                .insert(date_time, span);
+        }
+    }
+
```
Hywan, line 183: 👆  end of parsing (super small)

👇 start of rendering (not so small)

PR #6141, opened 2026-02-09
feat(qr-login): add additional QRCodeGrantLoginError enum values

hughns:
On top of https://github.com/matrix-org/matrix-rust-sdk/pull/6178

This adds additional `QRCodeGrantLoginError` enum values in the SDK crate and additional `HumanQrGrantLoginError` enum values in the FFI crate.

Most of the new values are assigned from what previous went to `Unknown` enum values.

The is one bugfix for where `::DeviceIDAlreadyInUse` was erroneously being returned when the actually the device was not returned by the homeserver. This is now mapped to `::DeviceNotFound` values.

Signed-off-by:

### DIFF
--- a/crates/matrix-sdk/CHANGELOG.md
+++ b/crates/matrix-sdk/CHANGELOG.md
@@ -79,6 +79,8 @@ All notable changes to this project will be documented in this file.
 
 ### Refactor
 
+- [**breaking**] `QRCodeGrantLoginError::UnableToCreateDevice` has been removed
+  ([#6141](https://github.com/matrix-org/matrix-rust-sdk/pull/6141)
 - The `RoomEventCache::paginate_thread_backwards` method is replaced by `RoomEventCache::thread_pagination` which returns a new `ThreadPagination` type, similar to `RoomPagination`.
   ([#6174](https://github.com/matrix-org/matrix-rust-sdk/pull/6174))
 
--- a/crates/matrix-sdk/src/authentication/oauth/qrcode/grant.rs
+++ b/crates/matrix-sdk/src/authentication/oauth/qrcode/grant.rs
@@ -115,7 +115,7 @@ async fn finish_login_grant<Q>(
             .unwrap_or(device_authorization_grant.verification_uri.to_string())
             .as_str(),
     )
-    .map_err(|_| QRCodeGrantLoginError::UnableToCreateDevice)?;
+    .map_err(|e| QRCodeGrantLoginError::Unknown(e.to_string()))?;
     state.set(GrantLoginProgress::WaitingForAuth { verification_uri });
 
     // We send the new device the m.login.protocol_accepted message to let it know
--- a/bindings/matrix-sdk-ffi/CHANGELOG.md
+++ b/bindings/matrix-sdk-ffi/CHANGELOG.md
@@ -90,6 +90,8 @@ All notable changes to this project will be documented in this file.
 
 ### Refactor
 
+- [**breaking**] `HumanQrGrantLoginError::UnableToCreateDevice` has been removed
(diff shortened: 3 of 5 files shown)


crates/matrix-sdk/src/authentication/oauth/qrcode/grant.rs
hughns: I *think* that this value might be misnamed/inappropriate too.

The error that is mapped is that the verification URI from the Device Authorization Grant cannot be converted into a URL.

What I am thinking is that `UnableToCreateDevice` should be removed from the enums (SDK+FFI) and instead map to `Unknown` (rather than adding a `InvalidDeviceAuthorizationGrantUri`).

@Johennes do you have any recollection from originally implementing this?

Johennes: Hm, yeah, I think you're right. The semantics don't make sense here. I have to admit I don't recall how we landed on this error but replacing it with `Unknown` seems fair to me.

hughns: 👍 I've done this in 180a6943a7413e6433bff4ab05c217107bf01f04

crates/matrix-sdk/CHANGELOG.md
```diff
@@ -50,6 +50,14 @@ All notable changes to this project will be documented in this file.
...
   question from the client's key backup.
   ([#6017](https://github.com/matrix-org/matrix-rust-sdk/pull/6017))
+- Add `QRCodeGrantLoginError::SecureChannel` for secure channel errors
+  ([#6141](https://github.com/matrix-org/matrix-rust-sdk/pull/6141)
+- Add `QRCodeGrantLoginError::UnexpectedMessage` for protocol message errors
+  ([#6141](https://github.com/matrix-org/matrix-rust-sdk/pull/6141)
+- Add `QRCodeGrantLoginError::LoginFailure` for login failure errors received from the other device
+  ([#6141](https://github.com/matrix-org/matrix-rust-sdk/pull/6141)
+- Add `QRCodeGrantLoginError::DeviceNotFound` for when the requested device was not returned by the homeserver
+  ([#6141](https://github.com/matrix-org/matrix-rust-sdk/pull/6141)
```
poljar: New entries to the top please.

crates/matrix-sdk/CHANGELOG.md
```diff
@@ -65,6 +73,14 @@ All notable changes to this project will be documented in this file.
...
   encoded Curve25519 public key.
   ([#5940](https://github.com/matrix-org/matrix-rust-sdk/pull/5940))
+- Return `QRCodeGrantLoginError::DeviceNotFound` instead of `QRCodeGrantLoginError::DeviceIDAlreadyInUse` for when
+  the new device is not returned by the homeserver.
+  ([#6141](https://github.com/matrix-org/matrix-rust-sdk/pull/6141)
+
+### Refactor
+
+- [**breaking**] `QRCodeGrantLoginError::UnableToCreateDevice` has been removed
+  ([#6141](https://github.com/matrix-org/matrix-rust-sdk/pull/6141)
```
poljar: Same, to the top please.

crates/matrix-sdk/CHANGELOG.md
hughns: Done in d65e27b87ff991756f44f0e4bc8617cf1c177131

crates/matrix-sdk/CHANGELOG.md
hughns: Ditto in d65e27b87ff991756f44f0e4bc8617cf1c177131

poljar:
Oh, seems that there are some conflicts now, could you resolve them?

hughns:
I missed one of the changelog entries. I'll fix that now.

PR #6143, opened 2026-02-09
fix(sdk): Restrict when `m.room.member` represents a `LatestEvent` candidate

Hywan:
This patch relies on `MembershipChange` to decide when a `m.room.member` represents a `LatestEvent` candidate. It was a mistale to rely on the `membership` strictly, because the `prev_content` must be taken into account.

Thus, this patch adds the following cases to `Knocked`, `Joined` and `Invited`: `InvitationAccepted` and `KnockedAccepted`. Moreover, this patch excludes other cases, including `ProfileChanged`, which was a bug previously! When the user had a new display name, it was considered as a `LatestEvent` candidate.

### DIFF
--- a/crates/matrix-sdk/CHANGELOG.md
+++ b/crates/matrix-sdk/CHANGELOG.md
@@ -50,6 +50,8 @@ All notable changes to this project will be documented in this file.
 
 ### Bugfix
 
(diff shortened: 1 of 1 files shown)


pixlwave:
Works great, thanks for the quick fix @Hywan 🙌

PR #6144, opened 2026-02-09
refactor(sdk): `waited_for_initial_prev_token` is no more an `Arc<AtomicBool>`

Hywan:
This patch changes `RoomEventCacheState::waited_for_initial_prev_token` from `Arc<AtomicBool>` to `bool`. First off, the `Arc` wasn't used in any useful way (never cloned for example). Second, the `AtomicBool` was always used as a regular bool, no atomicity was really used. Lastly, this patch adds the `assume_has_waited_for_initial_prev_token` method to replace the `= true` operation to make code a bit more readable.

### DIFF
--- a/crates/matrix-sdk/src/event_cache/room/mod.rs
+++ b/crates/matrix-sdk/src/event_cache/room/mod.rs
@@ -1230,13 +1230,8 @@ mod private {
         }
 
         /// Get the `waited_for_initial_prev_token` value.
-        pub fn waited_for_initial_prev_token(&self) -> bool {
-            self.state.waited_for_initial_prev_token
-        }
-
-        /// Assume the system has already waited for the initial `prev_token`.
-        pub fn assume_has_waited_for_initial_prev_token(&mut self) {
-            self.state.waited_for_initial_prev_token = true;
+        pub fn waited_for_initial_prev_token(&mut self) -> &mut bool {
+            &mut self.state.waited_for_initial_prev_token
         }
 
         /// Find a single event in this room.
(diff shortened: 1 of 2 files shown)


crates/matrix-sdk/src/event_cache/room/mod.rs
```diff
@@ -1233,9 +1229,14 @@ mod private {
...
-            &self.state.waited_for_initial_prev_token
+        /// Get the `waited_for_initial_prev_token` value.
+        pub fn waited_for_initial_prev_token(&self) -> bool {
+            self.state.waited_for_initial_prev_token
+        }
+
+        /// Assume the system has already waited for the initial `prev_token`.
+        pub fn assume_has_waited_for_initial_prev_token(&mut self) {
+            self.state.waited_for_initial_prev_token = true;
         }
```
bnjbvr: I don't agree with the new function: it's extra cognitive overhead, plus it's not "assuming" (which would make it hypothetical), it's indeed _confirming_ we've waited for the initial previous token. Plus, the runtime overhead of a function call (despite high chances it would be inlined) doesn't justify the one-liner. I'd inline it at both call sites, since it's rather clear what it's doing (setting the boolean value to true).

Hywan: I agree. We at least need a setter to avoid exposing the `state` to the `pagination` module. I propose to rename `assume_…` to `set_waited_for_initial_prev_token(&mut self, waited: bool)`. Thoughts?

Hywan: or we can make `waited_for_initial_prev_token` to return a mutable reference?

Hywan: Proposition here https://github.com/matrix-org/matrix-rust-sdk/pull/6144/commits/61cce73c3bbad59b753d9d1055d10e0523b15277.

PR #6148, opened 2026-02-09
Account: Fix a deserialisation failure when the `avatar_url` is `null`.

pixlwave:
As seen in Element X, Ruma's newer `get_profile_field` doesn't decode `{"avatar_url":null}` as `None` and instead throws an error, so this PR uses the older request which does the right thing.

### DIFF
--- a/crates/matrix-sdk/src/account.rs
+++ b/crates/matrix-sdk/src/account.rs
@@ -42,9 +42,9 @@ use ruma::{
             config::{get_global_account_data, set_global_account_data},
             error::ErrorKind,
             profile::{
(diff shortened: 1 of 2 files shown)


pixlwave:
> Error: fatal: unable to access 'https://github.com/matrix-org/matrix-rust-sdk/': The requested URL returned error: 500

CI is looking happy right now then 😄

pixlwave:
> could you please open an upstream issue?

Sure: https://github.com/ruma/ruma/issues/2360

poljar:
> > could you please open an upstream issue?
> 
> Sure: [ruma/ruma#2360](https://github.com/ruma/ruma/issues/2360)

Thanks.

PR #6150, opened 2026-02-09
doc(sdk): Fix a typo in `subscribe_to_pinned_events`

Hywan:
Yup.


crates/matrix-sdk/src/event_cache/room/mod.rs
```diff
@@ -1826,7 +1826,7 @@ mod private {
         /// be kept up-to-date as new events are pinned, and new related
         /// events show up from sync or backpagination.
         ///
-        /// This requires that the room's event cache be initialized.
```
bnjbvr, line 1829: That was the [subjonctive form](https://en.wikipedia.org/wiki/English_subjunctive) 😛

PR #6157, opened 2026-02-10
fix(event cache): make sure the `ThreadSummary`'s latest event is recomputed after a successful redecryption

### LINKED ISSUE #5703: thread summary: UTD not updated after being correctly decrypted

bnjbvr:
When an in-thread event is redecrypted, the related thread root's summary would NOT be updated, before the fix commit in this PR. The most of the PR is the test, the fix is twofold:

- first, when post-processing happens from r2d2, make sure to force a recompute of the thread summary, for the related thread root (if the redecrypted event was part of a thread)
- second, when recomputing this summary, force an update, even if the summary hasn't changed; for an event that transitions from UTD to decrypted, the event ID remains the same, and *not* updating if the summary hasn't changed wouldn't trigger an update for observers (and thus the timeline could never update its code).

The added test is a regression test that would showcase a failure mode of #5703.

Fixes #5703.

### DIFF
--- a/crates/matrix-sdk/src/event_cache/room/mod.rs
+++ b/crates/matrix-sdk/src/event_cache/room/mod.rs
@@ -1707,7 +1707,7 @@ mod private {
             let mut new_events_by_thread: BTreeMap<_, Vec<_>> = BTreeMap::new();
 
             for event in events {
-                self.maybe_apply_new_redaction(&event).await?;
+                self.maybe_apply_new_redaction(&event, post_processing_origin).await?;
 
                 if self.state.enabled_thread_support {
                     // Only add the event to a thread if:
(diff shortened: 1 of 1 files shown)


crates/matrix-sdk/src/event_cache/room/mod.rs
```diff
@@ -1839,10 +1848,10 @@ mod private {
                 None
             };
 
-            if prev_summary == new_summary.as_ref() {
-                trace!(%thread_root, "thread summary is already up-to-date");
-                return Ok(());
-            }
+            // Note: we don't check whether the summary has actually changed before
```
poljar: Not sure how easy this would be, but can't we trigger the update only if the source of the update is a redecryption and leave the logic as is for all other sources of updates?

bnjbvr: Yep, could implement it by piping down the `post_processing_origin` parameter! There's one cargo feature set  in which it isn't used (e2ee disabled), so that lead to a bit of ugly code, but that's manageable and alright here.

PR #6159, opened 2026-02-11
feat: Expose MSC4171 service members

frebib:
Part of https://github.com/element-hq/element-x-android/issues/4034

As per title. Useful for client applications to appropriately filter out service members.

(I've never really written rust before, so if I did something stupid/suboptimal, please be kind. I also wildly guessed at the API shape here. Suggestions welcome)


andybalaam (approved):
This looks great to me - thanks!

PR #6161, opened 2026-02-11
chore(ffi): use ruma's built-in uniffi support to eliminate some event type enums

Johennes:
This is the smallest possible way that I could find to start using https://github.com/ruma/ruma/pull/2338 in the FFI bindings. It eliminates the types `StateEventType`, `MessageLikeEventType` and `RoomAccountDataEventType` from the FFI crate and instead uses the Ruma types directly .

The only difference of the two types in the foreign language should be the use of `Custom` vs. `Other`.

### DIFF
--- a/.deny.toml
+++ b/.deny.toml
@@ -47,7 +47,5 @@ allow-git = [
     "https://github.com/jplatte/const_panic",
     # A patch override for the bindings: https://github.com/smol-rs/async-compat/pull/22
     "https://github.com/element-hq/async-compat",
-    # We can release vodozemac whenever we need but let's not block development
-    # on releases.
-    "https://github.com/matrix-org/vodozemac",
+    "https://github.com/mozilla/uniffi-rs"
 ]
--- a/.github/workflows/bindings_ci.yml
+++ b/.github/workflows/bindings_ci.yml
@@ -194,7 +194,7 @@ jobs:
 
   complement-crypto:
     name: "Run Complement Crypto tests"
-    uses: matrix-org/complement-crypto/.github/workflows/single_sdk_tests.yml@399a1deeab0d7e4fa9604cbe83b1df6058c40193 # main
+    uses: matrix-org/complement-crypto/.github/workflows/single_sdk_tests.yml@1b1f86348079608fbc5276fe77c54dd66638a622 # main
     with:
         use_rust_sdk: "." # use local checkout
         use_complement_crypto: "MATCHING_BRANCH"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -70,7 +70,7 @@ rand = { version = "0.10.1", default-features = false, features = ["std", "std_r
 regex = { version = "1.12.2", default-features = false }
 reqwest = { version = "0.13.1", default-features = false }
 rmp-serde = { version = "1.3.0", default-features = false }
-ruma = { version = "0.16.0", features = [
+ruma = { git = "https://github.com/ruma/ruma", rev = "8beb557b466a43b057c9c155277c77fe8766fadf", features = [
     "client-api-c",
     "compat-unset-avatar",
     "compat-upload-signatures",
@@ -114,8 +114,8 @@ tracing-core = { version = "0.1.34", default-features = false }
 tracing-subscriber = { version = "0.3.20", default-features = false, features = ["std", "smallvec", "fmt"] }
 unicode-normalization = { version = "0.1.25", default-features = false }
 unicode-segmentation = { version = "1.12.0", default-features = false }
-uniffi = { version = "0.31.0", default-features = false, features = ["cargo-metadata"] }
-uniffi_bindgen = { version = "0.31.0", default-features = false, features = ["cargo-metadata"] }
+uniffi = { git = "https://github.com/mozilla/uniffi-rs.git", rev = "e5f4821410bea19e71984ea5e06a7bc8b11ed9e5", version = "0.31.0", default-features = false, features = ["cargo-metadata"] }
+uniffi_bindgen = { git = "https://github.com/mozilla/uniffi-rs.git", rev = "e5f4821410bea19e71984ea5e06a7bc8b11ed9e5", version = "0.31.0", default-features = false, features = ["cargo-metadata"] }
 url = { version = "2.5.7", default-features = false }
 uuid = { version = "1.18.1", default-features = false }
 vergen-gitcl = { version = "1.0.8", default-features = false }
--- a/bindings/matrix-sdk-ffi/src/event.rs
+++ b/bindings/matrix-sdk-ffi/src/event.rs
@@ -19,9 +19,9 @@ use ruma::{
     EventId,
     events::{
         AnySyncMessageLikeEvent, AnySyncStateEvent, AnySyncTimelineEvent, AnyTimelineEvent,
-        MessageLikeEventContent as RumaMessageLikeEventContent, RedactContent,
-        RedactedStateEventContent, StaticStateEventContent, SyncMessageLikeEvent, SyncStateEvent,
-        TimelineEventType as RumaTimelineEventType,
+        MessageLikeEventContent as RumaMessageLikeEventContent, MessageLikeEventType,
+        RedactContent, RedactedStateEventContent, StateEventType, StaticStateEventContent,
+        SyncMessageLikeEvent, SyncStateEvent, TimelineEventType as RumaTimelineEventType,
         room::{
             encrypted,
             message::{MessageType as RumaMessageType, Relation},
@@ -90,7 +90,7 @@ impl From<AnyTimelineEvent> for TimelineEvent {
 /// The timeline event type.
 #[derive(Clone, uniffi::Enum, PartialEq, Eq, Hash)]
 #[uniffi::export(Eq, Hash)]
-pub enum TimelineEventType {
+pub enum FfiTimelineEventType {
     /// The event is a message-like one and should be displayed as such.
     MessageLike { value: MessageLikeEventType },
     /// The event is a state event, and may or may not be displayed in the
@@ -98,7 +98,7 @@ pub enum TimelineEventType {
     State { value: StateEventType },
 }
 
-impl From<RumaTimelineEventType> for TimelineEventType {
+impl From<RumaTimelineEventType> for FfiTimelineEventType {
     fn from(value: RumaTimelineEventType) -> Self {
         match value {
             RumaTimelineEventType::Audio => {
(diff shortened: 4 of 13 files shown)


Johennes:
This is failing the complement-crypto tests now. These [downgrade](https://github.com/matrix-org/complement-crypto/blob/main/rebuild_rust_sdk.sh#L55) uniffi to 0.28.3. This no longer works because ruma's new uniffi support depends on changes from uniffi 0.31.0.

In particular:

- `doc(hidden)` doesn't work in 0.28.3
- `uniffi::export` only works on functions and `impl` blocks in 0.28.3 (ruma uses `uniffi::export(Display, Eq, Hash)` on an enum)
- In 0.28.3 `custom_type!` only takes two type parameters and then you need to implement `UniffiCustomTypeConverter`. In 0.31.0 the macro takes the converters directly in a third argument

I can think of the following ways to resolve this:

1. Split the uniffi implementation in ruma into two features, one that works with 0.28.3 and one that uses 0.31.0
2. Change the `rebuild_rust_sdk.sh` script from complement to hack the changes into a local ruma checkout

I'm not sure which of these is worse but maybe I've missed a better option?

jplatte:
The better option is to help move https://github.com/NordSecurity/uniffi-bindgen-go/pull/86 forward, IMHO.

Johennes:
> The better option is to help move [NordSecurity/uniffi-bindgen-go#86](https://github.com/NordSecurity/uniffi-bindgen-go/pull/86) forward, IMHO.

That's still two minor versions behind what ruma needs but, yes, ideally the complement crypto tests wouldn't have to downgrade uniffi in the first place.

jplatte:
Yeah, but it would unblock additional upgrades (which may be much easier than that one, if you're lucky ^^).

poljar:
This should be unblocked as https://github.com/matrix-org/complement-crypto/pull/233 is now merged.

Johennes:
Looks like this still fails the complement-crypto tests. Will have to investigate what's wrong there.

```
 Error: Failed to extract data from archive member `ruma_events-2c1d40fe02f50a4b.ruma_events.21d0e9ca7cef90a2-cgu.09.rcgu.o`

Caused by:
    0: extracting metadata for 'UNIFFI_META_RUMA_EVENTS_CONSTRUCTOR_EPHEMERALROOMEVENTTYPE_FROM_STRING'
    1: Constructor return type must be Self or Arc<Self>
```

.github/workflows/bindings_ci.yml
```diff
@@ -194,7 +194,7 @@ jobs:
 
   complement-crypto:
     name: "Run Complement Crypto tests"
-    uses: matrix-org/complement-crypto/.github/workflows/single_sdk_tests.yml@399a1deeab0d7e4fa9604cbe83b1df6058c40193 # main
+    uses: matrix-org/complement-crypto/.github/workflows/single_sdk_tests.yml@1b1f86348079608fbc5276fe77c54dd66638a622 # main
```
Hywan, line 197: Why this change?

Hywan (commented):
Thank you for your contributions!

.github/workflows/bindings_ci.yml
Johennes: This is the commit that updated complement-crypto to uniffi 0.31.0 (which is the version the uniffi support in Ruma [requires](https://github.com/ruma/ruma/blob/14f35298b45544cb93ce184427017ab4d5c301b4/Cargo.toml#L44)) and it is slightly ahead of the currently used SHA: https://github.com/matrix-org/complement-crypto/commits/main/

Johennes:
> Looks like this still fails the complement-crypto tests. Will have to investigate what's wrong there.
> 
> ```
>  Error: Failed to extract data from archive member ``ruma_events-2c1d40fe02f50a4b.ruma_events.21d0e9ca7cef90a2-cgu.09.rcgu.o``
> 
> Caused by:
>     0: extracting metadata for 'UNIFFI_META_RUMA_EVENTS_CONSTRUCTOR_EPHEMERALROOMEVENTTYPE_FROM_STRING'
>     1: Constructor return type must be Self or Arc<Self>
> ```

The error appears to refer to [this method](https://github.com/ruma/ruma/blob/14f35298b45544cb93ce184427017ab4d5c301b4/crates/ruma-macros/src/events/event_enum/event_type.rs#L301). But confusingly that does already return `Self`. 🤷‍♂️

Johennes:
Sigh. I think `#[uniffi::constructor]` may only work on objects but not on enumerations. So we'll probably need to fix this in Ruma.

Cargo.toml
```diff
@@ -71,7 +71,7 @@ rand = { version = "0.10.1", default-features = false, features = ["std", "std_r
 regex = { version = "1.12.2", default-features = false }
 reqwest = { version = "0.13.1", default-features = false }
 rmp-serde = { version = "1.3.0", default-features = false }
-ruma = { version = "0.16.0", features = [
+ruma = { git = "https://github.com/Johennes/ruma", rev = "aaf6737173deaf50630f794573a1cace17d0c9c3", features = [
```
Johennes: Using this temporarily to test out https://github.com/ruma/ruma/pull/2501.

Johennes:
The cargo-deny error seems to also exist on `main`. Apart from that, I believe this should finally be ready for review now.

.deny.toml
```diff
@@ -47,7 +47,5 @@ allow-git = [
     "https://github.com/jplatte/const_panic",
     # A patch override for the bindings: https://github.com/smol-rs/async-compat/pull/22
     "https://github.com/element-hq/async-compat",
-    # We can release vodozemac whenever we need but let's not block development
-    # on releases.
-    "https://github.com/matrix-org/vodozemac",
+    "https://github.com/mozilla/uniffi-rs"
```
andybalaam, line 50: Are you deliberately denying vodozemac? If so, how is this related to this PR?

andybalaam (commented):
This looks reasonable to me, without much context on the complement-crypto changes etc. Waiting for Ivan to comment, and asked one question about what I think is a merge mistake.

.deny.toml
Johennes: I _think_ I removed it because it produced a warning.

```
❯ cargo deny check
warning[unmatched-source]: allowed source was not encountered
   ┌─ /Users/jm/code/matrix-rust-sdk/.deny.toml:52:6
   │
52 │     "https://github.com/matrix-org/vodozemac",
   │      ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ no crate source matched these criteria
```

andybalaam: I suggest leaving it in in this PR and making a new PR or issue where we can discuss why this warning is appearing and whether it indicates a problem we should fix.

poljar: The warning is normal, vodozemac was left as an allowed source for convenience reasons since we depended on a git branch of vodozemac relatively often.

Though this didn't happen for a while so maybe it's best to just remove it.

andybalaam: OK, fine by me then

andybalaam (approved):
Good with me but I think there are unresolved conversations with other reviewers.

Johennes:
Sorry, I'm now realizing the bindings tests have uncovered a few more name clashes such as:

> /Users/runner/work/matrix-rust-sdk/matrix-rust-sdk/bindings/apple/generated/swift/matrix_sdk_ffi.swift:40996:30: error: 'TimelineEventType' is ambiguous for type lookup in this context

I'm not sure why these didn't show up before. I think this will require renaming the exported FFI type names or trying to eliminate the FFI types in favour of the Ruma types.

Johennes:
Moving back to draft as this seems to require still further work. 😕

bindings/matrix-sdk-ffi/src/event.rs
```diff
@@ -90,15 +90,15 @@ impl From<AnyTimelineEvent> for TimelineEvent {
 /// The timeline event type.
 #[derive(Clone, uniffi::Enum, PartialEq, Eq, Hash)]
 #[uniffi::export(Eq, Hash)]
-pub enum TimelineEventType {
+pub enum FfiTimelineEventType {
```
Johennes, line 93: This required renaming because it caused a name-clash with the identically named Ruma type which is now also exported. Ruma's type is flat. So this wouldn't have been a drop-in replacement.

I tried renaming just the exported type name but that broke auto-generating the `Eq` and `Hash` implementations.

jplatte: > I tried renaming just the exported type name but that broke auto-generating the `Eq` and `Hash` implementations.

Not sure what this means exactly, but sounds like an upstream bug? Did you file an issue already?

Johennes: I first tried adding `#[uniffi::export(name = "FfiTimelineEventType")]` but that then failed when generating bindings claiming that it cannot locate `TimelineEventType`. I have a _hunch_ this might be because of the exported `Eq` and `Hash` implementations but haven't verified that. Reporting upstream sounds sensible assuming we can reproduce it in a minimal example.

Johennes:
It looks like this is finally ready. The remaining lint error also exists on `main`.

bindings/matrix-sdk-ffi/src/event.rs
Johennes: Tried to reproduce it in a minimal example but couldn't. So I may have hit something else or maybe it's already been fixed.

poljar, line 93: That's a bit of an ugly name, hopefully the users of the bindings won't mind too much.

The more important bit here is that this is a breaking change though. We should mention this in the changelog.

poljar (changes requested):
Looks good to me. We should just mention the rename of the exported type in the changelog as well.

bindings/matrix-sdk-ffi/src/event.rs
Johennes: Yeah, it's not pretty. Eventually this should just be replaced with the ruma type. That one has a different structure though. So doing that would have been a bigger breaking change than the renaming.

Maybe CC @stefanceriu for visibility on this.

poljar: And @jmartinesp as well.

Beware there's an annoying rename in this PR.

poljar (approved):

bindings/matrix-sdk-ffi/src/event.rs
Johennes: Argh, sorry for forgetting the Android side. I'm lacking coffee. 🙈 ☕

jmartinesp:
I feel like I've said this quite a few times, but next time we do something breaking in the FFI side, could we have some time to test it before merging something 😅 ?

This broke a few unit tests we have on Android, since now `StateEventType` and `MessageLikeEvenType` have one case which is an `Object` (`Custom`) and that means it'll automatically try to load the native library (which isn't even built for the host's architecture) and the test won't even start. It's not like they're important tests so we can just remove them, but it would have been better to know about this in advance.

poljar:
> This broke a few unit tests we have on Android, since now StateEventType and MessageLikeEvenType have one case which is an Object (Custom) and that means it'll automatically try to load the native library (which isn't even built for the host's architecture) and the test won't even start. It's not like they're important tests so we can just remove them, but it would have been better to know about this in advance.

Oh, I thought about requesting a real life test for it but underestimated how much this changed things.

Sorry.

If it's problematic we can revert this and reaply later, as it's a single commit.

jmartinesp:
No, I think it's fine, it's just a few unit tests we'll disable.

Johennes:
Apologies. I probably also should have pinged you about this much earlier than today. Sorry for the extra hassle.

PR #6162, opened 2026-02-11
feat(ffi): replace tracing's RollingFileAppender with SizeAndDateRollingWriter

stefanceriu:
This new file writer supports both total size and age based rolling, and will take care of a whole class of edge cases that make rageshake uploading fail.

* It will rotate log files when the configured time period elapses (e.g., hourly, daily).
* When the total size of all log files exceeds the configured limit, the oldest files are removed.
* Logs older than the configured max age are automatically removed during cleanup. 
* Only files matching the configured prefix and suffix are managed by any given writer instance.

**Disclosure**: I have built this by guiding an LLM (Claude Sonnet 4.5 in Zed) through the requirements, successive refactoring steps and necessary changes. I have also reviewed the results at every point and tested them against Element X iOS. I chose to do so here because it's a standalone component and the perfect testing grounds.

### DIFF
--- a/bindings/matrix-sdk-ffi/src/platform/rolling_writer.rs
+++ b/bindings/matrix-sdk-ffi/src/platform/rolling_writer.rs
@@ -587,7 +587,9 @@ mod tests {
         // Manually create several log files with different timestamps (alphabetically
         // sorted = oldest first) Total of 240 bytes
         std::fs::write(log_path.join("total.2024-01-01-10-00.log"), "x".repeat(80)).unwrap();
+        std::thread::sleep(std::time::Duration::from_millis(10));
         std::fs::write(log_path.join("total.2024-01-01-10-01.log"), "y".repeat(80)).unwrap();
+        std::thread::sleep(std::time::Duration::from_millis(10));
         std::fs::write(log_path.join("total.2024-01-01-10-02.log"), "z".repeat(80)).unwrap();
 
         let count_files = || {


bindings/matrix-sdk-ffi/src/platform/mod.rs
```diff
@@ -218,21 +226,17 @@ fn make_file_layer(
...
-
-    let writer =
-        builder.build(&file_configuration.path).expect("Failed to create a rolling file appender.");
+    let writer = SizeAndDateRollingWriter::new(
+        &file_configuration.path,
+        file_configuration.file_prefix,
+        file_configuration.file_suffix.unwrap_or_else(|| String::from(".log")),
+        Rotation::HOURLY,
+        file_configuration.max_total_size_bytes.unwrap_or(10 * 1024 * 1024), // Default: 10MB
+        file_configuration.max_age_seconds.unwrap_or(7 * 24 * 60 * 60),      // Default: 1 week
```
bnjbvr: Can you define these two constant values as `const` values, at the top level? Makes it simpler to identify them.

bindings/matrix-sdk-ffi/src/platform/rolling_writer.rs
```diff
@@ -0,0 +1,781 @@
+use std::{
```
bnjbvr, line 15: Can you add the copyright notice?

bindings/matrix-sdk-ffi/src/platform/rolling_writer.rs
```diff
@@ -0,0 +1,781 @@
...
+/// This writer will rotate log files when the configured time period elapses
+/// (e.g., hourly, daily).
+///
+/// Files are named with a timestamp. When the total size of all log files
+/// exceeds the configured limit, the oldest files are removed.
+///
+/// Logs older than the configured max age are automatically removed during
+/// cleanup. Only files matching the configured prefix and suffix are managed.
+pub(super) struct SizeAndDateRollingWriter {
+    inner: Arc<Mutex<SizeAndDateRollingWriterInner>>,
```
bnjbvr: Out of curiosity: do you need the Arc? do you need the mutex? Maybe I'm missing it, but I don't see it cloned and shared with other data types…

bindings/matrix-sdk-ffi/src/platform/rolling_writer.rs
```diff
@@ -0,0 +1,781 @@
...
+struct SizeAndDateRollingWriterInner {
+    current_file: File,
+    current_path: PathBuf,
+    base_path: PathBuf,
+    file_prefix: String,
+    file_suffix: String,
+    rotation: Rotation,
+    max_total_size_bytes: u64,
+    max_age_seconds: u64,
+    last_rotation_time: String,
```
bnjbvr: If this is a duration, can it be stored in a type that better represents a time/date?

bindings/matrix-sdk-ffi/src/platform/rolling_writer.rs
```diff
@@ -0,0 +1,781 @@
...
+}
+
+impl SizeAndDateRollingWriter {
+    pub(super) fn new(
```
bnjbvr, line 162: Can you add a doc comment with the main side effects of constructing this:

- it does create missing paths
- it does rotate logs at startup
- it does trim the old logs

bindings/matrix-sdk-ffi/src/platform/rolling_writer.rs
```diff
@@ -0,0 +1,781 @@
...
+        inner.rotate(false)?;
+        inner.trim_old_logs()?;
+
+        Ok(Self { inner: Arc::new(Mutex::new(inner)) })
+    }
+
+    /// Manually trigger log rotation if conditions are met.
+    ///
+    /// This will rotate the current log file if the time period has changed.
+    #[allow(dead_code)]
```
bnjbvr: Is this dead code? If so: can it be removed; if not, can you remove the `dead_code` annotation?

bindings/matrix-sdk-ffi/src/platform/rolling_writer.rs
```diff
@@ -0,0 +1,781 @@
...
+    fn roll(&self) -> io::Result<()> {
+        let mut inner = self.inner.lock().unwrap();
+        inner.rotate(true)
+    }
+
+    /// Manually trigger cleanup of old log files.
+    ///
+    /// This removes all log files older than the configured max age, keeping
+    /// only files that match the configured prefix and suffix.
+    #[allow(dead_code)]
```
bnjbvr: ditto

bindings/matrix-sdk-ffi/src/platform/rolling_writer.rs
```diff
@@ -0,0 +1,781 @@
...
+impl SizeAndDateRollingWriterInner {
+    /// Check if rotation is needed based on time period change.
+    fn should_rotate_by_time(&self) -> bool {
+        let current_time = self.get_rotation_time();
+        current_time != self.last_rotation_time
+    }
+
+    /// Get the current rotation time string based on the configured rotation
+    /// period.
+    fn get_rotation_time(&self) -> String {
```
bnjbvr: Could this also return a type better representing a time/date?

bindings/matrix-sdk-ffi/src/platform/rolling_writer.rs
```diff
@@ -0,0 +1,781 @@
...
+///
+/// Logs older than the configured max age are automatically removed during
+/// cleanup. Only files matching the configured prefix and suffix are managed.
+pub(super) struct SizeAndDateRollingWriter {
+    inner: Arc<Mutex<SizeAndDateRollingWriterInner>>,
+}
+
+struct SizeAndDateRollingWriterInner {
+    current_file: File,
+    current_path: PathBuf,
```
bnjbvr: I think this one shouldn't be "global" state in the struct, but passed with a function parameter to the users, since this is used only temporarily.

bindings/matrix-sdk-ffi/src/platform/rolling_writer.rs
```diff
@@ -0,0 +1,781 @@
...
+    /// Get the current rotation time string based on the configured rotation
+    /// period.
+    fn get_rotation_time(&self) -> String {
+        let now = chrono::Local::now();
+        match self.rotation {
+            Rotation::MINUTELY => now.format("%Y-%m-%d-%H-%M").to_string(),
+            Rotation::HOURLY => now.format("%Y-%m-%d-%H").to_string(),
+            Rotation::DAILY => now.format("%Y-%m-%d").to_string(),
+            Rotation::NEVER => String::from("never"),
+            _ => now.format("%Y-%m-%d").to_string(), // Handle WEEKLY and any future variants
```
bnjbvr: Maybe use a `TODO` in the comment? Or handle weekly here?

bindings/matrix-sdk-ffi/src/platform/rolling_writer.rs
```diff
@@ -0,0 +1,781 @@
...
+        rotation: Rotation,
+        max_total_size_bytes: u64,
+        max_age_seconds: u64,
+    ) -> io::Result<Self> {
+        let base_path = path.as_ref().to_path_buf();
+        fs::create_dir_all(&base_path)?;
+
+        let mut inner = SizeAndDateRollingWriterInner {
+            current_file: File::create("/dev/null")?, // Temporary placeholder
+            current_path: PathBuf::new(),
```
bnjbvr: (These two lines are a bad code smell, that the two fields should probably not live in the struct)

bindings/matrix-sdk-ffi/src/platform/rolling_writer.rs
```diff
@@ -0,0 +1,781 @@
...
+                    break;
+                }
+            }
+        }
+
+        Ok(())
+    }
+}
+
+impl Write for SizeAndDateRollingWriter {
```
bnjbvr: Are the two `Write` impl needed? My expectation would be that this one implements only `MakeWriter`, and the other type it returns would implement `Write`?

bindings/matrix-sdk-ffi/src/platform/rolling_writer.rs
```diff
@@ -0,0 +1,781 @@
...
+    inner: Arc<Mutex<SizeAndDateRollingWriterInner>>,
+    _phantom: std::marker::PhantomData<&'a ()>,
+}
+
+impl<'a> Write for SizeAndDateRollingWriterHandle<'a> {
+    fn write(&mut self, buf: &[u8]) -> io::Result<usize> {
+        let mut inner = self.inner.lock().unwrap();
+
+        // Check if rotation is needed
+        inner.rotate(true)?;
```
bnjbvr: Do we need to check for rotation *and* maybe rotate for every single write to the file? Seems a bit heavy, in terms of load, and this might be super slow.

I wonder if publicly exposing the top-level functions for rotating and discarding old logs would be sufficient, so embedders could call them at the right time? (for instance, just before uploading logs to a remote server, when these become rageshakes)

bindings/matrix-sdk-ffi/src/platform/rolling_writer.rs
```diff
@@ -0,0 +1,781 @@
...
+    fn test_total_size_limit_removes_oldest_files() {
+        // This test verifies that when the total size of all log files exceeds
+        // max_total_size_bytes, the oldest files are removed
+        let temp_dir = tempdir().unwrap();
+        let log_path = temp_dir.path();
+
+        // Manually create several log files with different timestamps (alphabetically
+        // sorted = oldest first) Total of 240 bytes
+        std::fs::write(log_path.join("total.2024-01-01-10-00.log"), "x".repeat(80)).unwrap();
+        std::thread::sleep(std::time::Duration::from_millis(10));
```
bnjbvr, line 590: Can we remove the sleep statements in the tests?

bindings/matrix-sdk-ffi/src/platform/rolling_writer.rs
```diff
@@ -0,0 +1,781 @@
...
+/// exceeds the configured limit, the oldest files are removed.
+///
+/// Logs older than the configured max age are automatically removed during
+/// cleanup. Only files matching the configured prefix and suffix are managed.
+pub(super) struct SizeAndDateRollingWriter {
+    inner: Arc<Mutex<SizeAndDateRollingWriterInner>>,
+}
+
+struct SizeAndDateRollingWriterInner {
+    current_file: File,
```
bnjbvr: Can you add a doc comment that this one is required in the struct because it's passed to the result of the `MakeWriter` impl? (or something along those lines)

bindings/matrix-sdk-ffi/src/platform/rolling_writer.rs
stefanceriu: I'm surprised by how inconsistent these headers are in the FFI layer. I'm gonna go update them all.

bindings/matrix-sdk-ffi/src/platform/rolling_writer.rs
stefanceriu: That was used to store the current file's timestamp string (excluding the prefix and suffix) to compare it for rolling. I asked it to derive it from the `current_path` instead.

bindings/matrix-sdk-ffi/src/platform/rolling_writer.rs
stefanceriu: Nop, the Arc is definitely not needed but we do need a mutex for acessing the File itself. I had it refactor the code and split it up into a `WriterConfig` and a separate mutex protected `WriterState`

bindings/matrix-sdk-ffi/src/platform/rolling_writer.rs
stefanceriu: Told it it's confusing, it came up with `format_rotation_timestamp` and even put an example result in the doc comment 😅

bindings/matrix-sdk-ffi/src/platform/rolling_writer.rs
stefanceriu: Yeah.. took quite a bit of prodding but it finally agreed that the first one's not needed.

bindings/matrix-sdk-ffi/src/platform/rolling_writer.rs
stefanceriu: We (me and doug) did think about that as another option but as long as we want hourly rolling that's not really an option. It's also a tad unclear to me how one would keep a reference to the writer over ffi do to so.

fwiw the tracing rolling writer seems to check every time too https://github.com/tokio-rs/tracing/blob/74fc079d2308a959d3a17907b3ba592ffd0a98f7/tracing-appender/src/rolling.rs#L226

bindings/matrix-sdk-ffi/src/platform/rolling_writer.rs
stefanceriu: They're used for the tests but I had it move them to a `#[cfg(test)]` implementation block.

stefanceriu:
In keeping with the exercise I went ahead and put Benji's review comment into the LLM again and had it come up with solutions. It mostly went well but there were situations in which I had to double and triple check it to have it agree certain things aren't needed.

bindings/matrix-sdk-ffi/src/platform/mod.rs
```diff
@@ -1,10 +1,25 @@
+// Copyright 2025 The Matrix.org Foundation C.I.C.
```
bnjbvr, line 1: I don't especially care, but all those new notices read 2025… 😇

bindings/matrix-sdk-ffi/src/platform/rolling_writer.rs
```diff
@@ -0,0 +1,923 @@
+// Copyright 2025 The Matrix.org Foundation C.I.C.
```
bnjbvr: This one, though…

```suggestion
// Copyright 2026 The Matrix.org Foundation C.I.C.
```

bindings/matrix-sdk-ffi/src/platform/mod.rs
stefanceriu: Yeh, I figured it's *moar* correct than putting 2026 on all of them 😅

bindings/matrix-sdk-ffi/src/platform/rolling_writer.rs
stefanceriu: Bah, of course I forgot this one

Issue #6040, opened 2026-01-16
add support for pushing secrets


uhoreg:
implement https://github.com/matrix-org/matrix-spec-proposals/pull/4385

- [x] add function to push secret to verified devices
- [x] handle pushed secrets
- [ ] ~add bindings to ffi~ (will split into separate issue)

PR #6164, opened 2026-02-12
Add support for pushing secrets and receiving secret pushes

uhoreg:
see [MSC4385](https://github.com/matrix-org/matrix-spec-proposals/pull/4385)

Pushing secrets allow devices to send secrets to other devices without waiting for a secret request.

closes https://github.com/matrix-org/matrix-rust-sdk/issues/6040

Signed-off-by:

### DIFF
--- a/crates/matrix-sdk-indexeddb/src/crypto_store/migrations/mod.rs
+++ b/crates/matrix-sdk-indexeddb/src/crypto_store/migrations/mod.rs
@@ -917,28 +917,28 @@ mod tests {
         assert_eq!(withheld_entry.content.withheld_code(), WithheldCode::Blacklisted)
     }
 
-    /// Test migrating `withheld_sessions` data from store v14 to latest,
+    /// Test migrating `secrets_inbox` data from store v105 to latest,
     /// on a store with encryption disabled.
     #[async_test]
-    async fn test_v103_v104_migration_unencrypted() {
-        test_v103_v104_migration_with_cipher("test_v101_migration_unencrypted", None).await
+    async fn test_v105_v107_migration_unencrypted() {
+        test_v105_v107_migration_with_cipher("test_v107_migration_unencrypted", None).await
     }
 
-    /// Test migrating `withheld_sessions` data from store v14 to latest,
+    /// Test migrating `secrets_inbox` data from store v105 to latest,
     /// on a store with encryption enabled.
     #[async_test]
-    async fn test_v103_v104_migration_encrypted() {
+    async fn test_v105_v107_migration_encrypted() {
         let cipher = StoreCipher::new().unwrap();
-        test_v103_v104_migration_with_cipher(
-            "test_v101_migration_encrypted",
+        test_v105_v107_migration_with_cipher(
+            "test_v107_migration_encrypted",
             Some(Arc::new(cipher)),
         )
         .await;
     }
 
-    /// Helper function for `test_v14_v101_migration_{un,}encrypted`: test
-    /// migrating `withheld_sessions` data from store v14 to store v101.
-    async fn test_v103_v104_migration_with_cipher(
+    /// Helper function for `test_v105_v107_migration_{un,}encrypted`: test
+    /// migrating `secrets_inbox` data from store v105 to store v107.
+    async fn test_v105_v107_migration_with_cipher(
         db_prefix: &str,
         store_cipher: Option<Arc<StoreCipher>>,
     ) {


crates/matrix-sdk-crypto/src/store/traits.rs
```diff
@@ -337,6 +339,22 @@ pub trait CryptoStore: AsyncTraitDeps {
     /// stored.
     async fn delete_secrets_from_inbox(&self, secret_name: &SecretName) -> Result<(), Self::Error>;
 
+    /// Get all the pushed secrets with the given [`SecretName`] we have
+    /// currently stored.
+    #[cfg(feature = "experimental-push-secrets")]
+    async fn get_pushed_secrets_from_inbox(
+        &self,
+        secret_name: &SecretName,
+    ) -> Result<Vec<SecretPushContent>, Self::Error>;
```
richvdh: do we not need to call this in `matrix-sdk` somewhere? Is that planned for a future PR?

crates/matrix-sdk-crypto/src/store/traits.rs
```diff
@@ -337,6 +339,22 @@ pub trait CryptoStore: AsyncTraitDeps {
...
+
+    /// Delete all the pushed secrets with the given [`SecretName`] we have
+    /// currently stored.
+    #[cfg(feature = "experimental-push-secrets")]
+    async fn delete_pushed_secrets_from_inbox(
+        &self,
+        secret_name: &SecretName,
+    ) -> Result<(), Self::Error>;
```
richvdh: Is there a reason to give pushed secrets and requested secrets separate APIs? Could we not expose a single API that returns both, thus simplifying application logic? (and maybe even use the same backing store in sqlite/indexeddb, though that might involve some irritating backwards support)?

I notice that, although the current API returns a `GossipedSecret` which is specific to `m.secret.send`, the callers of said API don't actually make any use of any fields other than `secret` and `name`.

crates/matrix-sdk-crypto/src/store/traits.rs
uhoreg: Yes, that is a future PR.  I wrote this PR first, then worked on the js-sdk parts, and then worked on the `matrix-sdk` parts.  I could add the `matrix-sdk` parts to this PR, but I kind of feel like it's cleaner keeping them separate.  But I'm not entirely sure, and am willing to be convinced otherwise.

crates/matrix-sdk-crypto/src/store/traits.rs
uhoreg: Yes, I considered trying to combine them.  It would be a breaking change.  I don't know if anything else uses `matrix-sdk-crypto` directly, and these functions in particular, but it seems unlikely that anything would require anything other than `secret` or `name`.

crates/matrix-sdk-crypto/src/store/traits.rs
richvdh: ah, no worries. Could I ask you to try and make that sort of thing clear (e.g. in the description) for future PRs?

crates/matrix-sdk-crypto/src/store/traits.rs
richvdh: We've not been shy with breaking changes of matrix-sdk-crypto in the past.

uhoreg: I've combined the pushed secrets and requested secrets in the crypto store.

uhoreg:
I believe this is ready for re-review

crates/matrix-sdk-indexeddb/src/crypto_store/mod.rs
```diff
@@ -717,14 +717,15 @@ impl IndexeddbCryptoStore {
...
-            let mut secret_store = indexeddb_changes.get(keys::SECRETS_INBOX);
+            let mut secret_store = indexeddb_changes.get(keys::SECRETS_INBOX_V2);
 
             for secret in &changes.secrets {
+                use std::ops::Deref;
                 let key = self.serializer.encode_key(
-                    keys::SECRETS_INBOX,
-                    (secret.secret_name.as_str(), secret.event.content.request_id.as_str()),
+                    keys::SECRETS_INBOX_V2,
+                    (secret.secret_name.as_str(), secret.secret.as_str()),
```
richvdh, line 730: why is the secret value used as part of the key?

crates/matrix-sdk-indexeddb/src/crypto_store/migrations/v104_to_v106.rs
```diff
@@ -0,0 +1,79 @@
...
+        Ok(())
+    })
+    .await
+}
+
+/// Migrate the data from different versions of the secrets inbox
+pub(crate) async fn data_migrate(name: &str, serializer: &SafeEncodeSerializer) -> Result<()> {
+    let db = MigrationDb::new(name, 11).await?;
+
+    // The new store has been made for inbound group sessions; time to populate it.
```
richvdh: not inbound group sessions

crates/matrix-sdk-crypto/src/store/types.rs
```diff
@@ -96,6 +100,47 @@ pub struct Changes {
...
+/// A secret that was received via an `m.secret.send`.
+#[derive(Clone, Serialize, Deserialize)]
+pub struct SecretsInboxItem {
+    /// The name of the secret.
+    pub secret_name: SecretName,
+    /// The contents of the secret.
+    #[serde(
+        serialize_with = "zeroizing_string_serializer",
+        deserialize_with = "zeroizing_string_deserializer"
+    )]
```
richvdh: this is a bit of a pain. I'm not really seeing why we need to implement Serialize for `SecretsInboxItem`?

Assuming we *do* need to implement it, it might be cleaner instead to make `secret` a regular `String`  and then make `SecretsInboxItem` `[#derive(Zeroize,ZeroizeOnDrop)]`

crates/matrix-sdk-sqlite/migrations/crypto_store/017_add_new_secrets_inbox.sql
```diff
@@ -0,0 +1,4 @@
+CREATE TABLE "secrets_inbox" (
+    "secret_name" BLOB NOT NULL,
+    "secret" BLOB NOT NULL
+);
```
richvdh, line 4: probably needs an index on `secret_name`

crates/matrix-sdk-sqlite/src/crypto_store.rs
```diff
@@ -190,16 +206,29 @@ impl SqliteCryptoStore {
...
 /// key for the dehydrated device pickle key in the key/value table.
 const DEHYDRATED_DEVICE_PICKLE_KEY: &str = "dehydrated_device_pickle_key";
 
-/// Run migrations for the given version of the database.
-async fn run_migrations(conn: &SqliteAsyncConn, version: u8) -> Result<()> {
+/// Initialize the database to version 1
+///
+/// This must be done before creating the store cipher, because the store cipher
+/// requires the `kv` table.
+pub(crate) async fn initialize_store(conn: &SqliteAsyncConn, version: u8) -> Result<u8> {
```
richvdh, line 238: what is the `version` argument?

crates/matrix-sdk-sqlite/src/crypto_store.rs
```diff
@@ -135,23 +158,16 @@ impl SqliteCryptoStore {
 
         let version = conn.db_version().await?;
         debug!("Opened sqlite store with version {}", version);
-        run_migrations(&conn, version).await?;
 
         conn.wal_checkpoint().await;
```
richvdh: the wal checkpoint now happens before we initialize the store? Why is that?

crates/matrix-sdk-sqlite/src/crypto_store.rs
```diff
@@ -95,6 +97,27 @@ impl EncryptableStore for SqliteCryptoStore {
 }
 
 impl SqliteCryptoStore {
+    /// Create an `SqliteCryptoStore` struct without trying to do any
+    /// initialization or migration
```
richvdh: Please could you define what this means a bit better? What is involved in "initialization" in this context? Does it involve opening files on the disk? Why is this a useful method? When would I want to call it? Relatedly, please document the parameters.

crates/matrix-sdk-sqlite/src/crypto_store.rs
```diff
@@ -210,8 +239,24 @@ async fn run_migrations(conn: &SqliteAsyncConn, version: u8) -> Result<()> {
...
     }
 
+    Ok(version)
+}
+
+/// Run migrations for the given version of the database.
+///
+/// max_version is the maximum version that the database will be migrated to,
+/// and is only used for testing, so will only be checked for the versions that
+/// are needed for tests.
```
richvdh: Given this description, I don't understand how `version` and `max_version` are different.

crates/matrix-sdk-sqlite/src/crypto_store.rs
```diff
@@ -210,8 +239,24 @@ async fn run_migrations(conn: &SqliteAsyncConn, version: u8) -> Result<()> {
...
         .await?;
+        return Ok(1);
     }
 
+    Ok(version)
+}
+
+/// Run migrations for the given version of the database.
+///
+/// max_version is the maximum version that the database will be migrated to,
```
richvdh:
```suggestion
/// `max_version` is the maximum version that the database will be migrated to,
```

crates/matrix-sdk-sqlite/src/crypto_store.rs
```diff
@@ -2098,6 +2182,74 @@ mod tests {
         assert!(backup_keys.decryption_key.is_some());
     }
 
+    /// Test that we migrate the secrets inbox properly
```
richvdh: could you mention something about which schema version changed the format, and how.

crates/matrix-sdk-sqlite/src/crypto_store.rs
```diff
@@ -2098,6 +2182,74 @@ mod tests {
...
+        };
+        let value = old_data_store.serialize_json(&secret).unwrap();
+        old_data_store
+            .write()
+            .await
+            .prepare("INSERT INTO secrets (secret_name, data) VALUES (?1, ?2)", |mut stmt| {
+                stmt.execute((SecretName::CrossSigningMasterKey.to_string(), value))
+            })
+            .await
+            .unwrap();
```
richvdh: maybe some more comments? This isn't just "Create a database" imho

crates/matrix-sdk-sqlite/src/crypto_store.rs
```diff
@@ -2098,6 +2182,74 @@ mod tests {
...
+
+        // After we open the store, which will migrate the data
```
richvdh: this sentence doesn't seem to make sense

crates/matrix-sdk-crypto/CHANGELOG.md
```diff
@@ -49,6 +49,10 @@ All notable changes to this project will be documented in this file.
   been eliminated entirely.
   ([#5963](https://github.com/matrix-org/matrix-rust-sdk/pull/5963))
 
+- [**breaking**] `CryptoStore::get_secrets_from_inbox` now returns a `Vec` of
+  the secrets as strings, rather than a `Vec` of `GossippedSecret` structs.
+  ([#6164](https://github.com/matrix-org/matrix-rust-sdk/pull/6164))
```
richvdh: and `Changes` has, uh, changed.

richvdh: Also: newest entries go at the top, in the same way that newest releases go at the top.

crates/matrix-sdk-crypto/src/gossiping/machine.rs
```diff
@@ -288,6 +292,60 @@ impl GossipMachine {
...
+                            content.cast(),
+                        );
+                        let request = OutgoingRequest {
+                            request_id: request.txn_id.clone(),
+                            request: Arc::new(request.into()),
+                        };
+                        self.inner
+                            .outgoing_requests
+                            .write()
+                            .insert(request.request_id.clone(), request);
```
richvdh, line 347: do the messages and their msgids get logged anywhere, so we can see what is going on?

crates/matrix-sdk-crypto/src/types/events/secret_push.rs
```diff
@@ -0,0 +1,104 @@
...
+    fn zeroize(&mut self) {
+        self.secret.zeroize();
+    }
+}
+
+impl Drop for SecretPushContent {
+    fn drop(&mut self) {
+        self.zeroize()
+    }
+}
```
richvdh: you could use `derive(ZeroizeOnDrop)` for this

crates/matrix-sdk-crypto/CHANGELOG.md
```diff
@@ -33,6 +33,8 @@ All notable changes to this project will be documented in this file.
   key bundle.
   ([#6017](https://github.com/matrix-org/matrix-rust-sdk/pull/6017))
   ([#6044](https://github.com/matrix-org/matrix-rust-sdk/pull/6044))
+- Add support for MSC4385.
+  - Add new method `OlmMachine::push_secret_to_verified_devices`.
```
richvdh: again, wrong place. Also, needs a PR link.

crates/matrix-sdk-crypto/src/gossiping/machine.rs
```diff
@@ -346,6 +351,36 @@ impl GossipMachine {
...
+        }
+        let Some(device) = self.inner.store.get_device_from_curve_key(sender, *sender_key).await?
+        else {
+            error!(?sender, ?sender_key, "Received secret push from unknown device");
+            return Ok(());
+        };
+        if !device.is_verified() {
+            error!(?sender, device_id = ?device.device_id(), "Received secret push from unverified device");
+            return Ok(());
+        }
```
richvdh, line 385: I think these should be warnings rather than errors, since they can be caused by other users or devices doing unexpected things, rather than necessarily reflecting an internal error on our side?

crates/matrix-sdk-indexeddb/src/crypto_store/mod.rs
uhoreg: If we get sent two different secrets for the same name, we need to keep both of them, so that we can check which one is correct.

crates/matrix-sdk-sqlite/migrations/crypto_store/017_add_new_secrets_inbox.sql
uhoreg: Yeah, probably.  I just copied the old schema and forgot about that.  The table will likely be small enough that it won't make a difference in practice, but I should add one anyways.

crates/matrix-sdk-indexeddb/src/crypto_store/mod.rs
uhoreg: Speaking of which, I need to add a test to make sure that things don't explode when you receive the same secret twice.

richvdh: > If we get sent two different secrets for the same name, we need to keep both of them, so that we can check which one is correct.

Ah ok. It would be good to note that as a comment somewhere.

crates/matrix-sdk-crypto/src/store/types.rs
uhoreg: Right.  I had originally planned to serialize this struct in the data store, before deciding to just write it out as a string, and then forgot to remove this serialization code.  Remove now.

uhoreg:
I messed up the latest commits, because I forgot to enable the feature flag when trying to build it, so it failed to build without me noticing. :(  So I'll have to redo some of the latest commits.

crates/matrix-sdk-indexeddb/src/crypto_store/mod.rs
uhoreg: Added comment, and added the test that I mentioned in https://github.com/matrix-org/matrix-rust-sdk/pull/6164#discussion_r2977357105

uhoreg:
Should be ready for re-review, though I'm going to have to work out the merge conflict (merge conflicts with db migrations are fun).

crates/matrix-sdk-indexeddb/src/crypto_store/mod.rs
```diff
@@ -717,14 +717,18 @@ impl IndexeddbCryptoStore {
         }
 
         if !changes.secrets.is_empty() {
-            let mut secret_store = indexeddb_changes.get(keys::SECRETS_INBOX);
+            let mut secret_store = indexeddb_changes.get(keys::SECRETS_INBOX_V2);
 
             for secret in &changes.secrets {
+                use std::ops::Deref;
+                // The secret is included in the key to allow us to receive
```
richvdh:
```suggestion
                // The (hashed) secret value is included in the key to allow us to receive
```

crates/matrix-sdk-indexeddb/src/crypto_store/mod.rs
```diff
@@ -717,14 +717,18 @@ impl IndexeddbCryptoStore {
...
+                // multiple secrets of the same name, and allow the client to
```
richvdh:
```suggestion
                // multiple secrets of the same name (indexeddb store entries must have a unique key), and allow the client to
```

crates/matrix-sdk-sqlite/migrations/crypto_store/016_add_new_secrets_inbox.sql
```diff
@@ -0,0 +1,6 @@
+CREATE TABLE "secrets_inbox" (
+    "secret_name" BLOB NOT NULL,
+    "secret" BLOB NOT NULL
+);
+CREATE INDEX "secrets_inbox_secret_name_idx"
+    ON "secrets_inbox" ("secret_name", "secret");
```
richvdh: If you're going so far as to put the index on both columns, might as well use it to enforce uniqueness


```suggestion
CREATE UNIQUE INDEX "secrets_inbox_secret_name_secret_idx"
    ON "secrets_inbox" ("secret_name", "secret");
```

crates/matrix-sdk-sqlite/src/crypto_store.rs
```diff
@@ -95,6 +97,36 @@ impl EncryptableStore for SqliteCryptoStore {
 }
 
 impl SqliteCryptoStore {
+    /// Create an `SqliteCryptoStore` struct without trying to do create the
```
richvdh:
```suggestion
    /// Create an `SqliteCryptoStore` struct without trying to create the
```

crates/matrix-sdk-sqlite/src/crypto_store.rs
richvdh: bump

crates/matrix-sdk-sqlite/src/crypto_store.rs
```diff
@@ -210,8 +248,30 @@ async fn run_migrations(conn: &SqliteAsyncConn, version: u8) -> Result<()> {
...
+    Ok(version)
+}
+
+/// Run migrations for the given version of the database.
+///
+/// # Arguments
+///
+/// * `store` - The store to run the migrations on
+///
+/// * `version` - the current version of the database.
```
richvdh:
```suggestion
/// * `version` - The current version of the database.
```

crates/matrix-sdk-sqlite/src/crypto_store.rs
```diff
@@ -210,8 +248,30 @@ async fn run_migrations(conn: &SqliteAsyncConn, version: u8) -> Result<()> {
...
+///
+/// * `max_version` - the maximum version that the database will be migrated to,
+///   and is only used for testing, so will only be checked for the versions
+///   that are needed for tests.
```
richvdh:
```suggestion
/// * `max_version` - The maximum version that the database will be migrated to.
///   Only used for testing, so will only be checked for the versions
///   that are needed for tests.
```

crates/matrix-sdk-crypto/src/gossiping/machine.rs
```diff
@@ -288,6 +295,98 @@ impl GossipMachine {
...
+                            request_id: request.txn_id.clone(),
+                            request: Arc::new(request.into()),
+                        };
+                        debug!(
+                            device = ?device.device_id(),
+                            event_type,
+                            request_id = ?request.request_id,
+                            ?secret_name,
+                            ?message_id,
+                            "Creating outgoing `io.element.msc4385.secret.push` to-device request"
```
richvdh: We're going to forget to update this when we switch to stable identifiers, and anyway it's just unnecessary coupling.

We're already recording the event_type. I don't think we need to repeat it here.

```suggestion
                            "Creating outgoing secret push to-device request",
```

crates/matrix-sdk-crypto/src/store/integration_tests.rs
```diff
@@ -1134,6 +1134,18 @@ macro_rules! cryptostore_integration_tests {
...
                 assert!(restored.is_empty(), "We should not have secrets of a different type stored");
 
+                // if we push a secret with the same name and secret, it should
+                // succeed
+                let mut changes = Changes::default();
+                changes.secrets.push(another_secret.into());
+                store.save_changes(changes).await.unwrap();
+
+                let restored = store.get_secrets_from_inbox(&SecretName::RecoveryKey).await.unwrap();
+                tracing::debug!(length = ?restored.len(), "length");
```
richvdh: Either remove this or give it a more helpful message, please

uhoreg: Oops.  I added that to debug the test, and then forgot to remove it afterwards.


```suggestion
```

crates/matrix-sdk-sqlite/src/crypto_store.rs
uhoreg: sorry for missing.  fixed now.

crates/matrix-sdk-indexeddb/src/crypto_store/migrations/mod.rs
```diff
@@ -907,6 +917,105 @@ mod tests {
         assert_eq!(withheld_entry.content.withheld_code(), WithheldCode::Blacklisted)
     }
 
+    /// Test migrating `withheld_sessions` data from store v14 to latest,
+    /// on a store with encryption disabled.
+    #[async_test]
+    async fn test_v103_v104_migration_unencrypted() {
+        test_v103_v104_migration_with_cipher("test_v101_migration_unencrypted", None).await
```
richvdh: confusing db prefix name here? Likewise below

crates/matrix-sdk-indexeddb/src/crypto_store/migrations/mod.rs
```diff
@@ -907,6 +917,105 @@ mod tests {
...
+    async fn test_v103_v104_migration_encrypted() {
+        let cipher = StoreCipher::new().unwrap();
+        test_v103_v104_migration_with_cipher(
+            "test_v101_migration_encrypted",
+            Some(Arc::new(cipher)),
+        )
+        .await;
+    }
+
+    /// Helper function for `test_v14_v101_migration_{un,}encrypted`: test
```
richvdh:
```suggestion
    /// Helper function for `test_v103_v104_migration_{un,}encrypted`: test
```

I think?

crates/matrix-sdk-indexeddb/src/crypto_store/migrations/mod.rs
```diff
@@ -907,6 +917,105 @@ mod tests {
...
+    /// migrating `withheld_sessions` data from store v14 to store v101.
+    async fn test_v103_v104_migration_with_cipher(
+        db_prefix: &str,
+        store_cipher: Option<Arc<StoreCipher>>,
+    ) {
```
richvdh: I think all of these functions need a rename, actually?

crates/matrix-sdk-indexeddb/src/crypto_store/migrations/mod.rs
```diff
@@ -907,6 +917,105 @@ mod tests {
...
+
+        let serializer = SafeEncodeSerializer::new(store_cipher.clone());
+
+        let _ = make_tracing_subscriber(None).try_init();
+        let db_name = format!("{db_prefix:0}::matrix-sdk-crypto");
+
+        // delete the db in case it was used in a previous run
+        let _ = Database::delete_by_name(&db_name).unwrap().await.unwrap();
+
+        // Given a DB with data in it as it was at v5
```
richvdh, line 964: why v5, of all things?

uhoreg: because there was already a helper function for creating a v5 db

uhoreg:
@richvdh can you mash the "Merge" button?  I don't seem to have permissions to do so.

PR #6165, opened 2026-02-12
chore: do not build `multiverse` by default

bnjbvr:
This introduces a new `xtask` to `cargo check` or `cargo run` `multiverse`, so that we can disable checking and building it the rest of the time, on dev machines. It's still tested when running `clippy` in CI.

### DIFF
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -94,6 +94,5 @@ jobs:
       - name: Run the benchmarks
         uses: CodSpeedHQ/action@4deb3275dd364fb96fb074c953133d29ec96f80f
         with:
-          run: cargo codspeed run
-          mode: "instrumentation"
+          mode: "simulation"
           token: ${{ secrets.CODSPEED_TOKEN }}


bnjbvr:
Turns out that the Clippy task still builds `multiverse`, so maybe we don't need a new job.

Cargo.toml
```diff
@@ -12,7 +12,7 @@ members = [
 ]
 exclude = ["testing/data"]
 # xtask, testing and the bindings should only be built when invoked explicitly.
-default-members = ["benchmarks", "crates/*", "labs/*"]
+default-members = ["benchmarks", "crates/*"]
```
poljar: Now remove benchmarks as well, those are already run as part of the CI.

bnjbvr:
Can I keep the xtask? I kind of like it, even for developer usage it's nice to not have to figure what's the required proper cargo line to run multiverse.

poljar:
> Can I keep the xtask? I kind of like it, even for developer usage it's nice to not have to figure what's the required proper cargo line to run multiverse.

Err, sure, though isn't it just `cargo run -p multiverse` and `cargo check -p multiverse`?

bnjbvr:
Yeah, maybe not worth an xtask, even CI could use that honestly. Alrighty then!

PR #6167, opened 2026-02-12
refactor(event cache): tiny refactorings around the pinned event cache

bnjbvr:
Some small refactorings following #6085. See each commit for details, they're rather simple and self-contained.

Part of #5954.

### DIFF
--- a/crates/matrix-sdk/src/event_cache/mod.rs
+++ b/crates/matrix-sdk/src/event_cache/mod.rs
@@ -79,7 +79,7 @@ mod persistence;
 mod redecryptor;
 mod room;
 
-pub use caches::TimelineVectorUpdate;
+pub use caches::TimelineVectorDiffs;
 pub use pagination::{RoomPagination, RoomPaginationStatus};
 #[cfg(feature = "e2e-encryption")]
 pub use redecryptor::{DecryptionRetryRequest, RedecryptorReport};
@@ -455,7 +455,7 @@ impl EventCache {
                         // so let's do it!
                         if !diffs.is_empty() {
                             let _ = room.inner.update_sender.send(
-                                RoomEventCacheUpdate::UpdateTimelineEvents(TimelineVectorUpdate {
+                                RoomEventCacheUpdate::UpdateTimelineEvents(TimelineVectorDiffs {
                                     diffs,
                                     origin: EventsOrigin::Cache,
                                 }),
@@ -1016,10 +1016,7 @@ impl EventCacheInner {
             let updates_as_vector_diffs = state_guard.reset().await?;
 
             let _ = room.inner.update_sender.send(RoomEventCacheUpdate::UpdateTimelineEvents(
-                TimelineVectorUpdate {
-                    diffs: updates_as_vector_diffs,
-                    origin: EventsOrigin::Cache,
-                },
+                TimelineVectorDiffs { diffs: updates_as_vector_diffs, origin: EventsOrigin::Cache },
             ));
 
             let _ = room
@@ -1254,7 +1251,7 @@ pub enum RoomEventCacheUpdate {
     },
 
     /// The room has received updates for the timeline as _diffs_.
-    UpdateTimelineEvents(TimelineVectorUpdate),
+    UpdateTimelineEvents(TimelineVectorDiffs),
 
     /// The room has received new ephemeral events.
     AddEphemeralEvents {
--- a/crates/matrix-sdk/src/event_cache/room/mod.rs
+++ b/crates/matrix-sdk/src/event_cache/room/mod.rs
@@ -52,7 +52,7 @@ use super::{
 };
 use crate::{
     client::WeakClient,
-    event_cache::{EventCacheError, caches::TimelineVectorUpdate},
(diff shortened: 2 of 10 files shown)


crates/matrix-sdk/src/event_cache/room/pinned_events.rs
```diff
@@ -254,10 +254,15 @@ impl PinnedEventCache {
...
-            Self::pinned_event_listener_task(room, state.clone()),
-        ));
+        let task = Arc::new(
+            room.client()
+                .task_monitor()
+                .spawn_background_task(
+                    "pinned_event_listener_task",
+                    Self::pinned_event_listener_task(room, state.clone()),
+                )
+                .abort_on_drop(),
```
Hywan, line 263: It's annoying it doesn't change the type of the task. I would have expect to see an `AbortOnDrop<T>` somewhere 🤔.

crates/matrix-sdk/src/event_cache/room/threads.rs
```diff
@@ -31,9 +31,9 @@ use crate::event_cache::{
     room::{LoadMoreEventsBackwardsOutcome, events::EventLinkedChunk},
 };
 
-/// An update coming from a thread event cache.
+/// A diff update for an event cache timeline represented as a vector.
 #[derive(Clone, Debug)]
-pub struct ThreadEventCacheUpdate {
+pub struct TimelineVectorUpdate {
```
Hywan: The sibling type for `RoomEventCache` is `RoomEventCacheUpdate::UpdateTimelineEvents`. Then, I suggest to you the `TimelineEventsUpdate` name here, instead of `TimelineVectorUpdate`. Thoughts?

crates/matrix-sdk/src/event_cache/room/mod.rs
```diff
@@ -731,7 +729,10 @@ mod private {
         events::EventLinkedChunk,
         sort_positions_descending,
     };
-    use crate::{Room, event_cache::room::PostProcessingOrigin};
+    use crate::{
+        Room,
+        event_cache::{TimelineVectorUpdate, room::PostProcessingOrigin},
```
Hywan: We have `use super::super` a couple lines above to access the `event_cache` module. Please group imports in the same path prefix.

crates/matrix-sdk/src/event_cache/room/pinned_events.rs
```diff
@@ -39,17 +39,20 @@ use tracing::{debug, instrument, trace, warn};
...
 };
-use crate::{Room, client::WeakClient, config::RequestConfig, room::WeakRoom};
 
 pub(in super::super) struct PinnedEventCacheState {
     /// The ID of the room owning this list of pinned events.
     room_id: OwnedRoomId,
 
     /// A sender for live events updates in this room's pinned events list.
-    sender: Sender<RoomEventCacheUpdate>,
+    sender: Sender<TimelineVectorUpdate>,
```
Hywan: Why don't we want to keep yielding `RoomEventCacheUpdate`? I'm okay with the change, but the commit message doesn't motivate this change.

crates/matrix-sdk/src/event_cache/mod.rs
```diff
@@ -1253,13 +1254,7 @@ pub enum RoomEventCacheUpdate {
...
 
     /// The room has received updates for the timeline as _diffs_.
-    UpdateTimelineEvents {
-        /// Diffs to apply to the timeline.
-        diffs: Vec<VectorDiff<TimelineEvent>>,
-
-        /// Where the diffs are coming from.
-        origin: EventsOrigin,
-    },
+    UpdateTimelineEvents(TimelineVectorUpdate),
```
Hywan: At this point, I wonder if the name of `TimelineVectorUpdate` should simply be `VectorDiffs` (plural), or `TimelineVectorDiffs` (if we want to make the `Timeline` part explicit), thoughts?

crates/matrix-sdk/src/event_cache/room/pinned_events.rs
bnjbvr: It's not using the `AbortOnDrop` helper, since this one requires a `JoinHandle`, and we don't have one with the `BackgroundJobHandle`.

crates/matrix-sdk/src/event_cache/room/mod.rs
bnjbvr: Honestly I don't really care what rust-analyzer decides is good enough for imports, and I don't imagine people stare at use statements that much in the wild, so I'll let it as is.

crates/matrix-sdk/src/event_cache/room/pinned_events.rs
bnjbvr: See the TODO that has been removed; I'll expand a bit in the commit message. The pinned events timeline is _a priori_ only interested in vector diff updates, likely not the other two kinds of changes, at this point.

crates/matrix-sdk/src/event_cache/mod.rs
bnjbvr: Yeah, `TimelineVectorDiffs` works ok too.

PR #6168, opened 2026-02-12
test: remove `StateTestEvent` and use the `EventFactory` instead

bnjbvr:
The `StateTestEvent` enum was redirecting to fixed uses of state events that were encoded in raw JSON in the code. It's actually possible to use the `EventFactory` for all these use cases, by introducing new methods in there if needs be. Works like a charm, and results in net LOC loss, so that's a win!?

Part of #3716.


stefanceriu (approved):
Woo, this is amazing, nicely done! 👏

PR #6170, opened 2026-02-12
test: remove more test event types and use the `EventFactory` instead

bnjbvr:
More simple opportunities to remove test types again.

Part of #3716.


Hywan:
👌

Issue #5112, opened 2025-05-27
History sharing: support restarting import of keys after a crash or being stopped


richvdh:
Imagine the keys and invite have already arrived (and we have joined the room), and we have started importing, and then we crashed. We want to re-do the import when the app restarts.

On matrix-js-sdk, as part of the "notice keys arriving after we joined" work, we have a list of recently-joined-and-not-processed-key-bundle rooms in the RustCrypto class. We should:

* Persist this info so it is available on startup. Make this persistence cross-platform.
* On startup, re-check this list and continue rooms importing rooms that we have not finished importing.

On matrix-rust-sdk, we don't handle late-arriving keys quite correctly, so we should:

* Fix late-arriving key bundles
* On startup, use the same mechanism as js-sdk to continue rooms importing rooms that we have not finished importing.

Part of https://github.com/element-hq/element-meta/issues/2829

kaylendog:
@poljar Andy and I have been discussing the best way to go about this, and I was wondering if there was a particular reason you opted for a duration-based import policy in https://github.com/matrix-org/matrix-rust-sdk/pull/5322? I do like it a bit more than my approach in the JS SDK where I store the room IDs we joined while the feature flag is enabled, but the latter is more easily converted into a persistent solution.

I'm wondering if we can consolidate this logic into the Rust SDK + crypto store which would let us remove the JS logic entirely. I think this would need us to move the "rooms we recently joined" introduced in https://github.com/matrix-org/matrix-rust-sdk/pull/5333 into `matrix-sdk-crypto` which seems a little silly.

poljar:
Hmm, I think that we discussed this in one of our meetings, i.e. when is it fine to accept room key bundles. It somewhat attempts to protect against the ability for the inviter to inject room keys into your client arbitrarily. 

I think that this is a bit orthogonal from the "oh I need to resume the download of the bundle" feature. Thus I think the js-sdk should have such an acceptance limitation as well.

As for consolidating the logic, this peace of information could live in the crypto store, though the `RoomInfo` in the SDK seemed like a more natural place for this as the `invite_accepted` field could be useful outside of a crypto context as well.

kaylendog:
> Hmm, I think that we discussed this in one of our meetings, i.e. when is it fine to accept room key bundles. It somewhat attempts to protect against the ability for the inviter to inject room keys into your client arbitrarily.
[...]

PR #6172, opened 2026-02-13
feat: Store invite acceptance details in the crypto store

kaylendog:
Per discussion under https://github.com/matrix-org/matrix-rust-sdk/issues/5112, it might be a good idea to move invite acceptance details into the crypto store, allowing us to consolidate the history sharing acceptance logic in both the Rust and JS SDK into one place.

Signed-off-by:


richvdh:
Replaced by https://github.com/matrix-org/matrix-rust-sdk/pull/6199

Issue #6152, opened 2026-02-10
[meta] Event Cache layers refactoring


Hywan:
<figure role="presentation">
<p align="center">
<img src="https://github.com/user-attachments/assets/f0bc8c69-a32f-4e14-b78c-9067029dcf05" width="50%" />
</p>
<figcaption><p align="center"><a href="https://en.wikipedia.org/wiki/Blinking_Sam">Blinking Sam (Samuel Johnson)</a></p></figcaption>
</figure>

<hr />

I would like to revisit the Event Cache design now it has grown a bit more and has more responsibilities. See the Problems Section to understand the motivation.

### Architecture

The current architecture of the Event Cache is the following:

- `EventCache`
  - contains one `RoomEventCache` per room
    - contains all events for the _main timeline_ in an `EventLinkedChunk `
    - contains one `ThreadEventCache` per thread
      - contains all events for the _thread timeline_ in an `EventLinkedChunk`
    - contains one `PinnedEventCache` for the pinned events in an `EventLinkedChunk`

### Sync

The _sync flow_ is the following (super simplified):

- For each updated room, fetch the appropriate `RoomEventCache`
- Push all events in it
- Post-process the events
  - Copy and dispatch the threaded events
  - Copy and dispatch the pinned events

### Pagination

The _pagination flow_ is the following (super simplified):

- Use a dedicated method on a `RoomEventCache` to paginate:
  - One method for the _main timeline_
  - One method for threads
  - None for pinned events
  - Push the events in their own buckets

## Problems

The major problems with this architecture are the following:

1. Threaded events and pinned events are first and foremost pushed in the `RoomEventCache`, before being copied over `ThreadEventCache` and `PinnedEventCache`.
    - It forces us to filter the threaded events from `RoomEventCache`, and possibly other events in the future
    - This is a waste of memory and disk usage
2. There is a non-negligible amount of duplicated code between `RoomEventCache`, `ThreadEventCache` and `PinnedEventCache`. Some are using a very similar read-write lock around their state (`RoomEventCache` and `PinnedEventCache`), some are sharing some parts of the pagination mechanism (`RoomEventCache` and `ThreadEventCache`).
    - The code is really confusing
    - Bugs can be duplicated
    - Fixing a bug can create a new bug somewhere else

### Pipelines

I propose to improve the _sync flow_ by introducing _pipelines_. It basically reverses the _post process_ to a _pre process_:

- Each event is dispatched in its own pipeline
[...]

PR #6174, opened 2026-02-13
refactor(sdk): Introduce the shared `Pagination` type and `PaginatedCache` trait

Hywan:
This patch implements a new `Pagination` type that is used by `RoomPagination`, and allows us to implement `ThreadPagination`.

The main problem is that the state of `ThreadEventCache` lives in `RoomEventCache`. See #6152 to see the full plan. We have to be creative a little bit, but nothing bad.

---

* Address https://github.com/matrix-org/matrix-rust-sdk/issues/6152

### DIFF
--- a/bindings/matrix-sdk-ffi/CHANGELOG.md
+++ b/bindings/matrix-sdk-ffi/CHANGELOG.md
@@ -80,6 +80,8 @@ All notable changes to this project will be documented in this file.
 
 ### Refactor
 
+- `RoomPaginationStatus` is renamed to `PaginationStatus`.
+  ([#6174](https://github.com/matrix-org/matrix-rust-sdk/pull/6174/))
 - Replaced `ClientBuilder::cross_process_store_locks_holder_name` with `ClientBuilder::cross_process_lock_config`, 
   which accepts a `CrossProcessLockConfig` value to specify whether the resulting `Client` will be used in a single 
   process or multiple processes. ([#6160](https://github.com/matrix-org/matrix-rust-sdk/pull/6160))
--- a/crates/matrix-sdk/CHANGELOG.md
+++ b/crates/matrix-sdk/CHANGELOG.md
@@ -82,6 +82,8 @@ All notable changes to this project will be documented in this file.
   ```rust
   room_event_cache.thread_pagination(thread_id).run_backwards_once(42).await
   ```
+- `RoomPaginationStatus` is renamed to `PaginationStatus`.
+  ([#6174](https://github.com/matrix-org/matrix-rust-sdk/pull/6174/))
 - [**breaking**] Replaced `ClientBuilder::cross_process_store_locks_holder_name` with 
   `ClientBuilder::cross_process_store_config` to allow specifying the configuration for the cross-process lock and 
   whether it should act as a no-op (client used in a single process) or we should keep the previous behavior (client 


crates/matrix-sdk/src/event_cache/room/threads.rs
```diff
@@ -173,7 +173,11 @@ impl ThreadEventCache {
         // resolve the gap.
         if let Some(prev_token) = self.chunk.rgap().map(|gap| gap.prev_token) {
             trace!(%prev_token, "thread chunk has at least a gap");
-            return LoadMoreEventsBackwardsOutcome::Gap { prev_token: Some(prev_token) };
+            return LoadMoreEventsBackwardsOutcome::Gap {
+                prev_token: Some(prev_token),
+                // No `prev_token` for threads, let's assume it's been waited.
+                waited_for_initial_prev_token: true,
```
bnjbvr, line 180: An even better comment would be that since you have `Some` token just above, the value of this field is irrelevant. (Making me wonder if the couple `prev_token` / `waited_for_initial_prev_token` should belong to an enum, so that it's clearer what's the relationship is between those two fields)

crates/matrix-sdk/src/event_cache/caches/pagination.rs
```diff
@@ -0,0 +1,336 @@
...
+use eyeball_im::VectorDiff;
+use matrix_sdk_base::{event_cache::Event, timeout::timeout};
+use tracing::{debug, instrument, trace};
+
+use super::super::{EventCacheError, Result};
+
+/// Type to run paginations.
+#[derive(Clone, Debug)]
+pub(in super::super) struct Pagination<C> {
+    pub(in super::super) cache: C,
```
bnjbvr: nit: this can just be `pub` (since the struct itself has `pub(in super::super)`, this means this field can be only `pub(in super::super)` even if it's marked `pub`)

crates/matrix-sdk/src/event_cache/caches/pagination.rs
```diff
@@ -0,0 +1,336 @@
...
+                        .await;
+
+                        trace!("done waiting");
+
+                        self.cache.mark_has_waited_for_initial_prev_token().await?;
+
+                        // Retry!
+                        //
+                        // Note: the next call to `load_more_events_backwards` should not return
+                        // `WaitForInitialPrevToken` because we've just mark we've waited for the
```
bnjbvr: nit: we've just marked*

crates/matrix-sdk/src/event_cache/caches/pagination.rs
```diff
@@ -0,0 +1,336 @@
...
+    ) -> Result<Option<BackPaginationOutcome>>;
+}
+
+/// Status for the pagination on a cache.
+#[derive(Debug, PartialEq, Clone, Copy)]
+#[cfg_attr(feature = "uniffi", derive(uniffi::Enum))]
+pub enum PaginationStatus {
+    /// No pagination is happening right now.
+    Idle {
+        /// Have we hit the star of the timeline, i.e. paginating wouldn't
```
bnjbvr: nit: start

(I'm starting to wonder if I'm reviewing my own code that's just moved…)

crates/matrix-sdk/src/event_cache/caches/pagination.rs
```diff
@@ -0,0 +1,336 @@
...
+    /// All the events that have been returned in the back-pagination
+    /// request.
+    ///
+    /// Events are presented in reverse order: the first element of the vec,
+    /// if present, is the most "recent" event from the chunk (or
+    /// technically, the last one in the topological ordering).
+    pub events: Vec<Event>,
+}
+
+/// TODO
```
bnjbvr: TODO here

bnjbvr (approved):
sweet, thanks!

bnjbvr (changes requested):
Actually marking request changes so we can discuss the organization of files in the `event_cache` folder; I'll open a PR with suggestions.

crates/matrix-sdk/src/event_cache/caches/pagination.rs
Hywan: Indeed. Solved.

crates/matrix-sdk/src/event_cache/caches/pagination.rs
Hywan: Yup. Solved.

crates/matrix-sdk/src/event_cache/caches/pagination.rs
Hywan: Solved.

crates/matrix-sdk/src/event_cache/caches/pagination.rs
Hywan: Solved.

crates/matrix-sdk/src/event_cache/room/threads.rs
Hywan: Done. Let's do one refactoring at a time. I wanted to go easy to simplify the review process.

PR #6178, opened 2026-02-16
test(qr-login): assert expect error codes in grant test cases

hughns:
This PR asserts the type of error that is returned when a grant login test case is expected to fail.

In adding these I've identified that for two of the test cases the error given doesn't make sense and should be fixed. I've added a TODO for these and will fix in a subsequent PR.

Signed-off-by:  Hugh Nimmo-Smith <hughns@element.io>

### DIFF
--- a/crates/matrix-sdk/src/authentication/oauth/qrcode/grant.rs
+++ b/crates/matrix-sdk/src/authentication/oauth/qrcode/grant.rs
@@ -1175,7 +1175,11 @@ mod test {
         });
 
         // Wait for all tasks to finish / fail.
-        grant.await.expect_err("Alice should abort the login");
+        assert_matches!(
+            grant.await,
+            Err(QRCodeGrantLoginError::Unknown(_)),
+            "Alice should abort the login with expected error"
+        );
         updates_task.await.expect("Alice should run through all progress states");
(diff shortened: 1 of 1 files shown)


crates/matrix-sdk/src/authentication/oauth/qrcode/grant.rs
```diff
@@ -1651,7 +1666,13 @@ mod test {
             .await;
         });
 
-        grant.await.expect_err("Alice should abort the login");
+        // TODO: the DeviceIDAlreadyInUse error here doesn't make much sense, instead
```
poljar, line 1669: Do you want to address those right away? I don't mind adding a couple more finer grained error variants.

hughns: At the moment I've addressed them in https://github.com/matrix-org/matrix-rust-sdk/pull/6141 where more variants are indeed added. I think I'm almost ready to undraft the PR so you could review as one if that makes more sense?

poljar: Ah ok, no let's just merge this then and we can look at #6141 after that.

poljar (approved):

PR #6184, opened 2026-02-17
feat(sqlite): `SqliteCryptoStore` has 1 write connection

Hywan:
Similarly to #5382 and #5744, this patch introduces a write-only connection in `SqliteCryptoStore`. The idea is to get many read-only connections, and a single write-only connection behind a lock, so that there is a single writer at a time.

This patch renames the `acquire` method to `read`, and it introduces a new `write` connection.


poljar (approved):
Thanks, looks sensible.

